PR #5989, opened 2014-02-13
Work to allow AWS security tokens and profiles to work with Ansible

willthames:
##### Issue Type:

Feature Pull Request
##### Ansible Version:

Based on:
ansible 1.5 (devel 6f405c8970) last updated 2014/02/13 18:12:39 (GMT +1000)
##### Environment:

N/A
##### Summary:

Allow security tokens and profiles to be used as arguments
to the 'common' ec2 modules

Mostly refactoring to provide two new methods,
`get_aws_connection_info`, which results in a dict that can be
passed through to the boto `connect_to_region` calls, and
`connect_to_aws` that can pass that dict through to the
`connect_to_region` method of the appropriate module.

Tidied up some variable names

Works around boto/boto#2100

profiles don't work with boto < 2.24, but this detects for that
and fails with an appropriate message. It is designed to work
if profile is not passed but boto < 2.24 is installed.

Modifications to allow empty aws auth variables to be passed
(this is useful if wanting to have the keys as an optional
parameter in ec2 calls - if set, use this value, if not set,
use boto config or env variables)

Further documentation enhancements and refactorings
will likely result from incorporating this.

While tested with `ec2`, `ec2_tag`, `ec2_vol` would be good
to have more testers (/cc @jctanner)
##### Expected Results:

profile and security_token can be passed through to AWS calls

### DIFF
--- a/lib/ansible/module_utils/ec2.py
+++ b/lib/ansible/module_utils/ec2.py
@@ -14,33 +14,44 @@ AWS_REGIONS = ['ap-northeast-1',
                'us-west-2']
 
 
-def ec2_argument_keys_spec():
+def aws_common_argument_spec():
     return dict(
+        ec2_url=dict(),
         aws_secret_key=dict(aliases=['ec2_secret_key', 'secret_key'], no_log=True),
         aws_access_key=dict(aliases=['ec2_access_key', 'access_key']),
+        validate_certs=dict(default=True, type='bool'),
+        security_token=dict(no_log=True),
+        profile=dict(),
     )
+    return spec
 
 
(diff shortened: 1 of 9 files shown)


willthames:
Updated to refactor to incorporate validate_certs changes from #5998

pas256:
This is great.

And thanks for removing `ec2_` from all of that too.

mpdehaan:
cc @jimi-c can you merge this one and update all the 1.5's to 1.6's in the version_added ?

jimi-c:
Merged, thanks!

Issue #5741, opened 2014-01-23
uncompress doesn't like FreeBSD


g--:
Uncompress doesn't work with FreeBSD, even after bsdtar has been replaced with gnutar.. it returns the generic error:

```
TASK: [unarchive test] ******************************************************** 
<127.0.0.1> ESTABLISH CONNECTION FOR USER: vagrant
<127.0.0.1> EXEC ['ssh', '-tt', '-vvv', '-o', 'ControlMaster=auto', '-o', 'ControlPersist=60s', '-o', 'ControlPath=/home/geoff/.ansible/cp/ansible-ssh-%h-%p-%r', '-o', 'Port=2200', '-o', 'IdentityFile=/home/geoff/.vagrant.d/insecure_private_key', '-o', 'KbdInteractiveAuthentication=no', '-o', 'PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey', '-o', 'PasswordAuthentication=no', '-o', 'User=vagrant', '-o', 'ConnectTimeout=10', '127.0.0.1', "/bin/sh -c 'mkdir -p $HOME/.ansible/tmp/ansible-1390500933.84-273596065993280 && chmod a+rx $HOME/.ansible/tmp/ansible-1390500933.84-273596065993280 && echo $HOME/.ansible/tmp/ansible-1390500933.84-273596065993280'"]
[...]

Issue #5659, opened 2014-01-17
unarchive module fails on FreeBSD hosts


ryanschulze:
Ansible version 1.4.3

```
- name: upload and extract file
  unarchive: src=foobar.tar.gz dest=/usr/local/src/
```

Works on Linux servers, but fails on FreeBSD servers with the error "dest must be an existing dir" (yes, the directory exists). 

Unfortunately I have no experience with python, but as far as I can tell the root problem is that the _remote _md5() function is returning a md5 back to the unarchive module (because "/sbin/md5 <directory>" on FreeBSD will give you a md5 for directories).

Unfortunately I have no good suggestion for a solution. The best thing I can think of offhand is a dirty workaround to skip /sbin/md5 if it is a directory on FreeBSD, no idea how this would affect other modules though.

```
( [ ! -d %s ] && /sbin/md5 -q %s 2>/dev/null)
```

bcoca:
There is a common md5 function that works with *BSDs that other modules use
(file,copy,etc) so this should be easy to fix.

## 

Brian Coca
Stultorum infinitus est numerus
0110000101110010011001010110111000100111011101000010000001111001011011110111010100100000011100110110110101100001011100100111010000100001
Pedo mellon a minno

bcoca:
Sorry, I misunderstood the issue. The _remote_md5 functions is already used
and it has it's own test to figure out the type:

"rc=0; [ -r \"%s\" ] || rc=2; [ -f \"%s\" ] || rc=1; [ -d \"%s\" ] && rc=3"

^ %s gets subsituted with the path

unarchive tests for != 3 to return   "dest must be an existing dir".

This should work on FreeBSD and would not depend on the output of the md5
command (unless I'm reading the code incorrectly).

ryanschulze:
the "3" is initially correctly set, but get's overwritten because /sbin/md5 on FreeBSD returns true (and prints a md5sum) for directories and the both are connected via || 

Other Linux/*NIX flavors return false and no md5sum for directories.

bcoca:
so AC (After Coffee) and reading more code, it seems that your initial
assessment was on the nose.

The rc var i mentioned in previous posts, only gets returned when md5
fails, directories do so on linux, but not on BSDs. For once this is not a
POSIX ambiguity as the standard says nothing about md5 command line
utilities, so just different programs/implementations.

BTW, this bug also affects the copy module in specific cases (content=).

I can probably submit a patch this weekend that solves this.

[...]

PR #5993, opened 2014-02-13
directory detection should now work on BSDs, added AIX alternate path

### LINKED ISSUE #5970: runner/__init__.py:_remote_md5() fails on FreeBSD when target is a directory.
Instead of returning a '3' indicating a directory it will calculate the MD5 hash of the directory. This breaks the behavior of a few modules including unarchive. Tested on FreeBSD 9 and 10.

bcoca:
This should fix #5970, #5741 and #5659

Small changes but I don't have AIX nor all the BSDs available for testing so I ask those that do to test this patch.

### DIFF
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -944,15 +944,16 @@ class Runner(object):
         path = pipes.quote(path)
         # The following test needs to be SH-compliant.  BASH-isms will
         # not work if /bin/sh points to a non-BASH shell.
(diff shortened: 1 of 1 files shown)


mpdehaan:
merged, thanks!

g--:
This patch works for me with FreeBSD. (I manually patched 1:1.4.3).

Looking forward to seeing this in an official release!

bcoca:
it already is, 1.5 has this

PR #5999, opened 2014-02-13
Insert newline when lineinfile is inserting at EOF, fixes #5825.

### LINKED ISSUE #5825: lineinfile doesn't play well with files that don't end with a newline
The following task...

```
-  lineinfile: dest=/etc/ssh/sshd_config regexp="^AllowGroups" line="AllowGroups root projects vagrant" state=present
```

... is intended to make sure the line exists in the default sshd configuration file.  However, apparently when lineinfile doesn't find the regexp in the file at all, it attempts to insert the line beginning at the `dest` file's EOF.  If that EOF is not a newline, as is the case in Ubuntu 13.10's default sshd_config, it results in the line value being merged with the last line in the file.  It looks like this:

```
UsePAM yesAllowGroups root projects vagrant
```

Which is a syntax error that prevents sshd from starting up.  If your goal is to automate the provisioning of a server inaccessible via ssh, perhaps this is what you want.  I just can't help but think it's not the expected behavior for most folks.

Oh, and I saw this mentioned [here](https://www.mail-archive.com/ansible-project@googlegroups.com/msg00205.html) but I don't see where an issue was ever opened.

donspaulding:
In the case that the `regexp` is not found in the `dest` file, lineinfile defaults to adding the `line` at the EOF.  If this EOF is not a newline character already, the effect is that lineinfile appends `line` to the last line of the file, which is surprising behavior.  This (untested!) change should fix #5825, if you consider #5825 a bug, as I do.

### DIFF
--- a/library/files/lineinfile
+++ b/library/files/lineinfile
@@ -248,7 +248,7 @@ def present(module, dest, regexp, line, insertafter, insertbefore, create,
     # if insertafter=/insertbefore didn't match anything
     # (so default behaviour is to add at the end)
     elif insertafter == 'EOF':
-        lines.append(line + os.linesep)
+        lines.append(os.linesep + line + os.linesep)
         msg = 'line added'
         changed = True
     # Do nothing if insert* didn't match


mpdehaan:
untested changes are pretty well discouraged -- not sure why this was submitted untested?

This seems like it also needs to check to see if the last line was already  newline so as to not add two newlines in a row.

PR #6000, opened 2014-02-13
service: use `systemctl list-units` in addition to `list-unit-files`

andreasf:
The service module checks for the presence of a service before issuing
a command on it, e.g. before starting the service. On systemd systems, the list
of available services is currently retrieved with `systemctl list-unit-files`.
Because LSB-defined services in /etc/init.d don't have unit files, they are not
part of this list. This commit adds an additional call to `systemctl
list-units`, to add support for LSB services.


andreasf:
This only appeared to work... the correct way is to just don't add ".service" to LSB service names and the service module will choose the right tool to enable it.

PR #6002, opened 2014-02-13
Small fix for ec2 validate_certs option

chrishoffman:
This is just a small bug fix to the new validate_certs option in the ec2 module.


jimi-c:
Oops, thanks for catching that!

PR #6005, opened 2014-02-14
Adding better failure output for chkconfig failures 

pgehres:
I spent a couple of hours trying to debug why a new init script worked great on the box but didn't work when I used Ansible.  The output of the failure led me in the wrong direction:

FAILED >> {
    "failed": true,
    "msg": "unknown service name"
}

It turned out enabled=true was what what breaking this because the init script didn't have any chkconfig settings. Hopefully this message will save people from the same hair pulling in the future.


mpdehaan:
merged, thanks!

PR #6011, opened 2014-02-14
Bug fix: Digital Ocean cannot delete server with unique_name

stevearc:
The unique_name parameter works great for state=present, but the call to `getkeyordie` causes it to fail when state=absent.


mpdehaan:
Think we just closed a duplicate of this from someone else just now, let me know if you still have problems!

Commit is https://github.com/ansible/ansible/commit/003a405182d2a6dc23719ad71ac34d6d8f7e860d

Thanks!

PR #6014, opened 2014-02-14
Remove superfluous/invalid quote

amirhhz:
Allow the example to be copy-pasted with minimal confusion about the error.


jimi-c:
Merged, thanks!

Issue #5652, opened 2014-01-16
memtotal_mb fact on HP-UX produces None type error on HP-UX B.11.11 OS instances


Logos01:
The current method of gathering memtotal_mb for HP-UX is causing a None type error on multiple HP-UX systems within my environment. Upon reviewing the command used, it appears there is no output.  I have modified my local library/system/setup file to reflect the below, which is successfully averting this error (changes beginning from line 1305).

```
             try:
                 rc, out, err = module.run_command("grep Physical /var/adm/syslog/syslog.log")
                 data = re.search('.*Physical: ([0-9]*) Kbytes.*',out).groups()[0].strip()
                 self.facts['memtotal_mb'] = int(data) / 1024
             except:
                 rc, out, err = module.run_command(
                     "echo 'phys_mem_pages/D '| adb -k /stand/vmunix /dev/kmem | tail -1 | awk '{print $1}'"
                 )
                 data = out
                 self.facts['memtotal_mb'] = int(data) / 1024
```

mpdehaan:
If you would like to supply a patch, please submit it as a pull request and be sure to catch the exception, unguarded "excepts" will not be accepted.

Thanks!

mpdehaan:
Seems it would be better to test what was available prior to running the command too.

Logos01:
My example code has a lot of cleanup before it's pull-worthy. But at least
I narrowed down the cause and have a solution.  I'll take a look at
cleaning it up for pull-request later.

On Thu, Jan 16, 2014 at 10:50 AM, Michael DeHaan
notifications@github.comwrote:

> Seems it would be better to test what was available prior to running the
> command too.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/ansible/ansible/issues/5652#issuecomment-32493053
> .

## 

##

mpdehaan:
Hi!

Thanks very much for your interest in Ansible.  It sincerely means a lot to us. 

On September 26, 2014, due to enormous levels of contribution to the project Ansible decided to reorganize module repos, making it easier
for developers to work on the project and for us to more easily manage new contributions and tickets.

We split modules from the main project off into two repos, http://github.com/ansible/ansible-modules-core and http://github.com/ansible/ansible-modules-extras

If you would still like this ticket attended to, we will need your help in having it reopened in one of the two new repos, and instructions are provided below.

We apologize that we are not able to make this transition happen seamlessly, though this is a one-time change and your help is greatly appreciated -- 
[...]

PR #6018, opened 2014-02-14
gather_memory_facts on HP-UX without syslog

Logos01:
Depending on the system configuration of HP-UX systems, Memory information may not be delivered to syslog, causing the current setup module to return an AttributeError in the get_memofy_facts function.  In this case, use the absolute debugger to grab the information directly from the system.  (Closes Issue #5652)

### DIFF
--- a/library/system/setup
+++ b/library/system/setup
@@ -1302,9 +1302,17 @@ class HPUX(Hardware):
         data = int(re.sub(' +',' ',out).split(' ')[5].strip())
         self.facts['memfree_mb'] = pagesize * data / 1024 / 1024
         if self.facts['architecture'] == '9000/800':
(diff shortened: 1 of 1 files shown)


uucico:
/dev/kmem cannot be read by non-root users.

mpdehaan:
Sorry for the delay in getting to this, we have a large backlog we're working through.

What @uucico says this is true, and also seems to have some indent errors.

Also in latest devel, run_command requires a use_unsafe_shell=True parameter for shell operations, so this wouldn't work verbatim.

I'm closing this due to the kmem question as this must work as non-root.  Checking the UUID and skipping the test would be permissable, as would checking the readability of /dev/kmem.

If you would like to update this, that would be appreciated!

Thanks!

PR #6020, opened 2014-02-14
cloud/ec2_elb module: Added wait_timeout parameter

bpennypacker:
As mentioned in the Ansible project mailing list (https://groups.google.com/forum/?fromgroups#!topic/ansible-project/qG9Thrx100A) this module has an issue whereby a transient Amazon error can result in the module failing when in fact the command sent to the elb succeeds.  The problem arises where, when polling an ELB, you can occasionally see an instance enter an error state very briefly prior to it reaching the InService state.

This pull request adds an optional wait_timeout parameter.  If wait_timeout is 0 (the default) then the behavior of the module is unchanged.  If wait_timeout is nonzero then any errors are ignored for wait_timeout seconds, ensuring that any transient errors don't trigger a failure prematurely.

### DIFF
--- a/library/cloud/ec2_elb
+++ b/library/cloud/ec2_elb
@@ -82,6 +82,12 @@ options:
     choices: ["yes", "no"]
     aliases: []
     version_added: "1.5"
+  wait_timeout:
+    description:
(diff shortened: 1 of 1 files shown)


mpdehaan:
merged, thanks!

PR #6028, opened 2014-02-16
yum module downgrade support

cove:
Allows the yum module to downgrade an RPM if it's older than the currently installed version. 

From the ref thread below, we discuses adding an optional version= param and mentioning that you can specify the version in the name. I didn't end up doing it that way because version is a bit ambiguous due to RPM's complex notion of version (i.e. epoch, rel, ver). Using the regular spec notation (it's an expression actually) would be more intuitive I believe, and there's examples in the module already. But am happy to adjust as needed. This is prob more of a bug fix than a feature add. 

Ref thread:
https://groups.google.com/forum/#!searchin/ansible-project/yum$20downgrade/ansible-project/P_Rp4fVJYHk/ZettkjxKBlcJ

### DIFF
--- a/library/database/mongodb_user
+++ b/library/database/mongodb_user
@@ -46,6 +46,11 @@ options:
             - The port to connect to
         required: false
         default: 27017
+    replicaset:
+        description:
+            - Replicaset to connect to (automatically connects to primary for writes)
+        required: false
+        default: null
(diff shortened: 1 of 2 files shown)


cove:
going to resubmit with correct changes listed.

PR #6029, opened 2014-02-16
Proposal for a new parameter to the command module ("maybe_changes")

b6d:
I find myself writing more and more playbooks in which I use the following sequence of tasks to notify a handler in case a call to a command (`smbpasswd` for example) changes some file (`secrets.tdb` in this example):
- a `command`-task calling `md5sum` for the file with `changed_when=false` and `register=state_before`
- a `command`-task to actually do the work (`smbpasswd` here) with `changed_when=false`
- a `command`-task calling `md5sum` for a second time with `register=state_after`, `changed_when="state_before != state_after"` and `notify=some_handler`

It is clear to me that it is practically impossible to determine reliably whether a `command`-task changed something. But I think there are use cases, where it makes sense to check if the task only changed a particular file. For these cases I'd like to propose a new parameter for the command module to specify the file one expects to change.

I've tried to make a sample implementation of this (the new parameter is called `maybe_changes` in it) which I'd like to leave here for your consideration/comments.

### DIFF
--- a/library/commands/command
+++ b/library/commands/command
@@ -24,6 +24,8 @@ import traceback
 import re
 import shlex
 import os
+import sha
+import errno
 
 DOCUMENTATION = '''
 ---
@@ -54,6 +56,13 @@ options:
     version_added: "0.8"
     required: no
     default: null
+  maybe_changes:
+    description:
+      - A filename. When this file's contents have not changed after the
+        command has been run, return C(changed=false).
+    version_added: "1.5"
+    required: no
+    default: null
   chdir:
     description:
       - cd into this directory before running the command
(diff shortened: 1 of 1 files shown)


mpdehaan:
So generally we have "changed_when" here for this sort of thing, though it's not something you could use trivially for file contents.

What I would probably do here would be to sample the md5 before and after and then do something like:
- name: mark changed
  shell: /bin/false
  when: md5one != md5two

Basically I have a bit of an issue with a field that is essentially single purpose for a particular class of command like "mark_changed_when_file_changes", which is what this is, where in other cases, change may be driven by different criteria.

I'd like to have a wider discussion though, perhaps you would like to stop by ansible-devel ?

PR #6033, opened 2014-02-16
add yum module downgrade support

cove:
Allows the yum module to downgrade an RPM if it's older than the currently installed version.

From the ref thread below, we discuses adding an optional version= param and mentioning that you can specify the version in the name. I didn't end up doing it that way because version is a bit ambiguous due to RPM's complex notion of version (i.e. epoch, rel, ver). Using the regular spec notation (it's an expression actually) would be more intuitive I believe, and there's examples in the module already. But am happy to adjust as needed. This is prob more of a bug fix than a feature add.

Ref thread:
https://groups.google.com/forum/#!searchin/ansible-project/yum$20downgrade/ansible-project/P_Rp4fVJYHk/ZettkjxKBlcJ

### DIFF
--- a/library/packaging/yum
+++ b/library/packaging/yum
@@ -1,6 +1,7 @@
 #!/usr/bin/python -tt
 # -*- coding: utf-8 -*-
 
+# (c) 2014, Epic Games, Inc.
 # (c) 2012, Red Hat, Inc
 # Written by Seth Vidal <skvidal at fedoraproject.org>
 #
--- a/library/packaging/yum
+++ b/library/packaging/yum
@@ -1,9 +1,9 @@
 #!/usr/bin/python -tt
 # -*- coding: utf-8 -*-
 
-# (c) 2014, Epic Games, Inc.
 # (c) 2012, Red Hat, Inc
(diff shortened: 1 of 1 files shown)


mpdehaan:
Thanks for this.

Would it be possible to call the is_installed only on the downgrade operation?   Just thinking this might be a hair expensive in some cases, or at least, not needed in the other cases.

risaacson:
Using the playbook here: https://gist.github.com/risaacson/99af626953ae2b76e2a0
I tested that this does work.

@mpdehaan I added in the logic to not run the downgrade logic if a version is not defined.

Merging this through another PR after cherry-picking the commits.

jimi-c:
@cove, this had to be reverted. The splitFilename() function does not operate correctly without a full RPM spec, so things like 'perl-foo-bar' without a version/release/arch were causing it to return some mangled results that in-turn prevented the packages from being installed. 

If you can rewrite this to account for the above situation, we'd be open to re-including it. Thanks!

cove:
@jimi-c yep will do, thanks for the heads up!

PR #5062, opened 2013-11-26
Module zabbix_maintenance to create maintenance windows in zabbix

abulimov:
Hello. I've wrote simple module to create and remove maintenance windows in Zabbix. It uses zabbix-api python module.

With this module you can, for example, set your hosts to maintenance mode before rolling update, and delete maintenance window after update.

PR #6034, opened 2014-02-16
new modules: zabbix_host, zabbix_group, zabbix_screen, zabbix_webcheck

cove:
These modules allow you to add/update hosts, put them in the correct groups and link in the monitoring templates, the basic requirements for bring a new host online. Been in production use since late last year, appear to be working fine against Zabbix 2.0.

### DIFF
--- a/library/monitoring/zabbix_group
+++ b/library/monitoring/zabbix_group
@@ -17,21 +17,19 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with Ansible. If not, see <http://www.gnu.org/licenses/>.
-# This is a port of the ruby zabbix api found here:
-# http://trac.red-tux.net/browser/ruby/api/zbx_api.rb
 #
 
 DOCUMENTATION = '''
 ---
 module: zabbix_group
-short_description: Zabbix host groups creates/deletes.
+short_description: Zabbix host groups creates/deletes
 description:
-   - Create Zabbix host groups if they don't exist.
-   - Delete the existing host groups in Zabbix.
+   - Create host groups if they don't exist.
+   - Delete existing host groups if they exist.
 version_added: "1.6"
 author: Tony Minfei Ding, Harrison Gu
 requirements:
-    - zabbix-api python module from https://github.com/gnetsman/zabbix_api (the 1.0 version in pip is too old unfortunately)
+    - zabbix-api python module
 options:
     server_url:
         description:
@@ -52,7 +50,7 @@ options:
         default: null
     state:
         description:
-            - create or delete host group.
+            - Create or delete host group.
             - Possible values are: present and absent.
         required: false
         default: "present"
@@ -60,11 +58,13 @@ options:
         description:
             - List of host groups to create or delete.
         required: true
+notes:
+    - Too many concurrent updates to the same group may cause Zabbix to return errors, see examples for a workaround if needed.
 '''
 
 EXAMPLES = '''
 # Base create host groups example
-- name: Create host groups in zabbix
+- name: Create host groups
   local_action:
     module: zabbix_group
     server_url: http://monitor.example.com
@@ -76,7 +76,7 @@ EXAMPLES = '''
       - Example group2
 
 # Limit the Zabbix group creations to one host since Zabbix can return an error when doing concurent updates
-- name: Create host groups in zabbix
+- name: Create host groups
   local_action:
     module: zabbix_group
     server_url: http://monitor.example.com
@@ -92,7 +92,7 @@ EXAMPLES = '''
 from ansible.module_utils.basic import *
 
 try:
-    from zabbix_api import ZabbixAPI
+    from zabbix_api import ZabbixAPI, ZabbixAPISubClass
     from zabbix_api import Already_Exists
 
     HAS_ZABBIX_API = True
@@ -156,7 +156,7 @@ def main():
     )
 
     if not HAS_ZABBIX_API:
-        module.fail_json(msg="Missing requried zabbix-api module, see this module's requirements for which version to install")
+        module.fail_json(msg="Missing requried zabbix-api module (check docs or install with: pip install zabbix-api)")
 
     server_url = module.params['server_url']
     login_user = module.params['login_user']
@@ -181,7 +181,7 @@ def main():
         group_ids, group_list = hostGroup.get_group_ids(host_groups)
 
     if state == "absent":
-   		# create host groups 	
+        # delete host groups
         if group_ids:
             delete_group_names = []
             hostGroup.delete_host_group(group_ids)
@@ -190,9 +190,9 @@ def main():
             module.exit_json(changed=True,
                              result="Successfully deleted host group(s): %s." % ",".join(delete_group_names))
         else:
-            module.exit_json(changed=False, result="Not found available host group(s) to delete.")
+            module.exit_json(changed=False, result="No host group(s) to delete.")
     else:
-    	# delete host groups
+        # create host groups
         group_add_list = hostGroup.create_host_group(host_groups)
         if len(group_add_list) > 0:
             module.exit_json(changed=True, result="Successfully created host group(s): %s" % group_add_list)
--- a/library/monitoring/zabbix_group
+++ b/library/monitoring/zabbix_group
@@ -1,7 +1,7 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
 
-# (c) 2013, Epic Games, Inc.
+# (c) 2013-2014, Epic Games, Inc.
 #
 # This file is part of Ansible
 #
@@ -54,6 +54,10 @@ options:
             - Possible values are: present and absent.
         required: false
         default: "present"
+    timeout:
+        description:
+            - The timeout of API request(seconds).
+        default: 10
     host_groups:
         description:
             - List of host groups to create or delete.
@@ -150,7 +154,8 @@ def main():
             login_user=dict(required=True),
             login_password=dict(required=True),
             host_groups=dict(required=True),
-            state=dict(default="present")
+            state=dict(default="present"),
+            timeout=dict(default=10)
         ),
         supports_check_mode=True
     )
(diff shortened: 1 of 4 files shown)


abulimov:
Looks great. Why didn't you used any existing Zabbix API library? Maintaining own api realization coud become a problem in the future... And i think, it would be better to have a unified parameter names in your modules and my zabbix_maintenance module (#5062). For example, 'zabbix_login_user' in my module is called just 'user', maybe it would be better to call it 'login_user' in all zabbix modules?

cove:
There were a few other issues in our env which is why the lib was in the module, however it looks like we can switch over to using the regular lib, and possibly submit a patch to for the lib instead. Fully agree on the login_user, yep.

On a side note, we also have a module for creating screens at this point, so we'll be submitting that as well, we're a bit busy right now, but should be in before the end of the month if not sooner.

jimi-c:
I think we're going to hold off on this, and let you guys work the common param names out. 

@cove, if you do stick with the urllib method, you'll need to start using the new `fetch_url` code that is in `lib/ansible/module_utils/urls.py` (see some of the other new modules that use it for examples). The reason for this is that by default, urllib\* doesn't validate SSL certs, so it's a bit unsafe to use directly. The new fetch_url() method handles cert validation, but also allows you to set a boolean to ignore validation for internal systems using self-signed certs.

Thanks!

abulimov:
@cove, I've updated my zabbix_maintenance module to use unified parameter names. Please, take a look and feel free to comment if you don't want to use some particular parameter name.

cove:
@abulimov these look good to me, we'll update to match those.

cove:
@abulimov can you try these updated modules out and see if they work for you?

library/monitoring/zabbix_group
```diff
@@ -0,0 +1,154 @@
...
+    host_groups:
+        description:
+            - List of host groups to create.
+        required: true
+'''
+
+EXAMPLES = '''
+- name: Create host groups in zabbix
+  local_action:
+    module: zabbix_host
```
abulimov: should be "module: zabbix_group"

library/monitoring/zabbix_host
```diff
@@ -0,0 +1,295 @@
...
+        default: None
+    status:
+        description:
+            - Status and function of the host.
+            - Possible values are: 0 - (default) monitored host; 1 - unmonitored host.
+        required: false
+        default: "0"
+    interfaces:
+        description:
+            - List of interfaces to be created for the host (see example).
```
abulimov, line 84: Maybe we should explain this a little more? At least give a link to Zabbix API documentation (https://www.zabbix.com/documentation/2.0/manual/appendix/api/hostinterface/definitions#host_interface). Example does not explain what "type: 1" ot "type: 4" means.

library/monitoring/zabbix_host
```diff
@@ -0,0 +1,295 @@
...
+    template_ids = host.get_template_ids(link_templates)
+    group_ids = host.get_group_ids(host_groups)
+    # check if host exist
+    isHostExist = host.is_host_exist(host_name)
+
+    if isHostExist:
+        # get host id by host name
+        host_id = host.get_host_id(host_name)
+        # get exist interface
+        exist_interface_list = host._zapi.hostinterface.get({'output': 'extend', 'filter': {'hostid': host_id}})
```
abulimov: As of version 1.0 of zabbix_api library (installed with 'pip install zabbix-api'), there is an error on host update: 'ZabbixAPI' object has no attribute 'hostinterface'.

srgvg:
Bumped in the same problems. API has changed, Examples talk about "host_names" whilst the parameter is "host_name", ..

Also:
- some parameters set as required in the documentation, but this is notr enforced in the AnsibleModule declaration.
- No support for check mode
- values for status: 0 and 1 should better be replaced by enabled/disabled
- no support to remove/delete a host?

cove:
we have a few more fixes to submit here still.

cove:
@abulimov @sergevanginderachter let me know how these updates work for you. thx.

library/monitoring/zabbix_host
```diff
@@ -0,0 +1,366 @@
...
+                group_id = group['groupid']
+                group_ids.append({'groupid': group_id})
+        return group_ids
+
+    # get host's templates
+    def get_templates_by_host_id(self, host_id):
+        template_ids = []
+        template_list = self._zapi.template.get({'output': 'extend', 'hostids': host_id})
+        for template in template_list:
+            template_ids.append(template['hostid'])
```
abulimov: Here should be template_ids.append(template['templateid'])

abulimov:
I've tested **zabbix_host, zabbix_group and zabbix_screen** modules, and everything seemed fine except the one error i've mentioned above. But in **zabbix_webcheck** module i was unable to create application with error 'Failed to get the application example app id: list index out of range'. And error message was also 'list index out of range' when I tried to add a webcheck for non-existent host.

cove:
@harrisongu @abulimov 
1. Fix zabbix_webcheck bug from abulimov's report.
2. Refactor zabbix_webcheck module to support more args passed.
3. Add "update" function support to zabbix_webcheck module.
4. Add timeout supports to all modules.

library/monitoring/zabbix_webcheck
```diff
@@ -0,0 +1,384 @@
...
+                    new_webcheck_step['httptestid'] = httptestid
+                    relace_webcheck_steps.append(new_webcheck_step)
+        return relace_webcheck_steps
+
+    # get webcheck params
+    def get_web_check_params(self, web_check_name, host_id, application_id, web_check_steps, optional_scenario_values,
+                             web_check_obj=None, httptestid=None):
+        steps = self.get_steps(web_check_steps)
+        if httptestid:
+            old_steps = list(web_check_obj['steps'].values())
```
abulimov: AttributeError: 'list' object has no attribute 'values', must be "old_steps = list(web_check_obj['steps'])"

abulimov:
@cove error in zabbix_host module that i've commented on line 263 is still not fixed. And i've commented another error in zabbix_webcheck module, line 239. Both fixes are trivial.

library/monitoring/zabbix_webcheck
cove: this was a doc error, see updated docs

library/monitoring/zabbix_host
cove: this should be fixed now

abulimov:
@cove error in webcheck module on line 239 is not fixed. No check for missing "steps" parameter, so module crashes when no "steps" specified.

cove:
@abulimov try now lmk

abulimov:
@cove check for missing "steps" works as expected, but I still have error when updating check:
on line 243 

```
old_steps = list(web_check_obj['steps'].values())
AttributeError: 'list' object has no attribute 'values'
```

And when I simply remove .values() call, so this line looks like 

```
old_steps = list(web_check_obj['steps'])
```

everything works fine. I'm testing this modules against Zabbix server 2.2.2, with Python 2.7.6.

cove:
@abulimov this appears to be due to a difference between 2.0,6 (the version we're on) and 2.2.2. We're looking into it.

cove:
@abulimov let us know how this works.

abulimov:
@cove now everything works great, thanks! Hope it will soon be merged - you have made awesome work!

cove:
@abulimov ah great, thanks for all the testing! thanks goes to @harrisongu @TonyMinfeiDing and @Damonchenchengkong for writing the modules.

srgvg:
I tested the zabbix_host module, and AFAICS this works well, except it seems to always return a 'changed' state. looking at the code, this is confirmed as the return dict always gets `changed=True`.
So the feedback is never "idempotent".

library/monitoring/zabbix_host
```diff
@@ -0,0 +1,372 @@
...
+        else:
+            if not group_ids:
+                module.fail_json(msg="Specify at least one group for updating host '%s'." % host_name)
+
+            # get exist interface
+            exist_interface_list = host._zapi.hostinterface.get({'output': 'extend', 'filter': {'hostid': host_id}})
+
+            # update host
+            interfaces_len = len(interfaces) if interfaces else 0
+            if len(exist_interface_list) > interfaces_len:
```
srgvg: What is the point of this conditional? I don't see any functional difference between the if and the else clause, unless I'm overlooking something?

library/monitoring/zabbix_host
```diff
@@ -0,0 +1,372 @@
...
+    def link_or_clear_template(self, host_name, template_ids_list):
+
+        host_id = self.get_host_id(host_name)
+
+        # get host's exist templates
+        exist_template_ids_list = self.get_templates_by_host_id(host_id)
+        exist_templates_ids = set(exist_template_ids_list)
+        template_ids = set(template_ids_list)
+        # get unlink and clear templates
+        templates_clear = exist_templates_ids.difference(template_ids)
```
srgvg: I also had to modify this line into 
`templates_clear = exist_templates_ids.difference(template_ids)`
and add an extra line
`template_ids_list = list(template_ids)` (to use the set)
to avoid issues where duplicate template id's where passed in

library/monitoring/zabbix_host
harrisongu: @sergevanginderachter,
This’s the specific requirement for the JMX monitor. For example, we must be specify the JMX interface when we need to link a “JMX template”(the template include some of JVM monitor items), otherwise errors. And If you want to remove the JMX interface, you’d must be unlink the JMX template at first.

cove:
@abulimov @sergevanginderachter can you validate @harrisongu's changes address your issues?

srgvg:
I re-imported those modules (_group and _host) and ran some tests, this seems to work well, without modifications.

Keep in mind I only tested some functionalities of some scripts, not everrything.

cove:
@abulimov  @sergevanginderachter ok, we've been using these scripts for quite a while at this point. Any objections to having it committed to the devel branch at this point?

jimi-c:
@cove we'll try and get these into testing as soon as possible, however we do have quite a backlog at the moment, so having more +1's always helps.

cove:
@resmo Definitely send a PR, thanks! ;-)

mpdehaan:
Hi!

Thank you very much!

Issue #3887, opened 2013-08-19
Expose realtime output from 'shell'


rektide:
Would love to see each 'shell' (&other) command creathave an option that creates a pipe for the duration of the process and pipes it's output into the pipe real time.

Doing some long lived load generation and Ansible is crucial, but I don't know how to monitor process, short of adding more layers.

mpdehaan:
This is the "async should return partial output".   I'm not sure this would make sense for many operations with a large number of hosts but it it's been discussed...

fdhenard:
+1

dwt:
+1

Please let me discuss the use case I would like to tackle with this.

Right now we are using puppet to deploy our systems. We are currently evaluating the switchover to ansible for at least part of the workload.

Our current process has been that we develop the puppet recipes against a vm and then we run it against a subset of the target hosts only showing what it would change (not doing it yet). This we evaluate and if it looks good we run it a second time actually doing the changes.

Since actually getting this to run is a multi step process, getting it completely automated with ansible was our first goal. This works, but it's hard to impossible to get the complete output of a remote tool (but I'm still investigating if it may be possible using the it's api). More importantly though it only arrives completely after the remote tool has run, which means that we lose the advantage of being able to monitor what it is doing while it is doing stuff which gives a better continuous workflow and keeps one in the zone instead of first concentrating on something else and then trying to come back to the output.

Also important to us, when the tool actually runs if we only get the output after the fact, there is no way to stop it halfway through if we realize that it does something horrible. (We luckily didn't need this ability yet - but defense in depth is better than believing one is always right).

I believe that substituting 'puppet' with any other long running tool generating a lot of output and doing important work and I believe the use case should be valid - even should you believe that we could just substitute puppet completely with ansible to achieve most of what I'm asking of here.

candlerb:
Here is another real-world use case: copying very large files (1TB+) from an NFS server to a target host. A single copy takes upwards of an hour.

While the playbook is running, I want to be sure that it is making progress (not stalled) and to get some feedback on how far this step has progressed.

I can use ansible to run a pipeline on the target host like:

```
pv /path/to/src | cat >/path/to/dest
```

which will give progress output to stderr. This could be captured into a buffer and passed back periodically to the client. I want to drive this from the API, so I'd also like to have periodic callbacks when stderr is being written to. Using callbacks also relieves ansible from having to maintain an infinite buffer.

[...]

PR #6035, opened 2014-02-16
Add support for realtime output

lorin:
This patch adds a new "show_output" option that can be enabled on
individual tasks, plays, or globally via a config option.(It defaults to false).

When enabled, whenever a running module calls AnsibleModule.run_command,
the output will appear in the user's console as the task is running.

Implements #3887

For example, on a task:

```
- name: upgrade packages
  apt: upgrade=full
  show_output: true
```

When running, output would appear in the console, like this:

```
TASK: [upgrade packages] ******************************************************
Reading package lists...
Building dependency tree...
Reading state information...
Reading extended state information...
Initializing package states...
The following NEW packages will be installed:
  linux-image-3.2.0-30-generic
...
```
## Overview of how it works

This change only shows output for modules that use the AnsibleModule.run_command method to invoke commands on the server. It modifies the run_command method so that when it invokes a command on the server, stdout and stderr are written to files instead of to a pipe. 

The client starts up a new process and runs "tail -f" on the stdout and stderr files so that the output appears on the client as the task is running. When the task completes, it kills the process.
### Creating temporary files

In `lib/ansible/runner/__init__.py`, the new `Runner._make_live_output_files` method will create two temporary files on the server for stdout and stderr. The Runner._executor_internal_inner method calls calls `_make_live_output_files` if show_output is enabled.
### Passing the filenames to module

The Runner._executor_internal_inner method adds  `ansible_stdout_file` and `ansible_stderr_file` entries to the local `inject` dictionary, which it passes to the ModuleReplacer.modify_module method.

The ModuleReplacer.modify_module method looks for the `ansible_stdout_file` and `ansible_stderr_file` variables. If they are present, it inserts them into lib/ansible/module_utils/basic.py as the new MODULE_STDOUT_STDERR_FILES dynamically inserted variable.
### Writing to files in run_command

In lib/ansible/module_utils/basic.py, the AnsibleModule.run_command method now checks if stdout/stderr files have been specified by checking the new MODULE_STDOUT_STDERR_FILES dynamically inserted variable.

If MODULE_STDOUT_STDERR_FILES contains a pair of filenames ('/path/to/temp/stdoutfile', '/path/to/temp/stderrfile'), then run_command will open those file in read+append mode, record the initial location of the files,
and pass them to subprocess.Popen. When done, it will read the contents
of those 
### Writing the files into the module

The names of the files are automatically inserted into lib/ansible/module_common.py via the new REPLACER_STDOUT_STDERR_FILES option as a pair of filenames ('/path/to/temp/stdoutfile', '/path/to/temp/stderrfile').

The ModuleReplacer will read the names of the files from the 'ansible_stdout_file' and 'ansible_stderr_file' variables if they are there, and replacing them.
## Watching the temporary files

The new lib/ansible/runner/live_output.py file contains a class called LiveOutput that is responsible for displaying the output of the temporary stdout and stderr file on the remote server.
This class will start a new process using multiprocessing library, connect to the remote server by calling `Runner._low_level_exec_command`, and then invoke "tail -f stdoutfile & tail -f stderrfile", passing `capture_output=False` as an argument so that the output would be sent directly to the terminal instead of to a pipe or file.
## Cleaning up the files

The `Runner._executor_internal_inner` method stops the LiveOutput object (terminating the process) and deletes the remote stdout and stderr file once it finishes running the module.
## Outputting directly to console

The proposed change adds the "capture_output" argument to the Connection.exec_command method of all connection plugins. The default value is set to `True`, which is the original behavior of this method (capture the output to a pipe).

If `capture_output=False`, then the output of the command will be sent directly to the user's terminal instead of being sent to a pipe or a file. This was necessary to implement the behavior of the LiveOutput class.

All connection plugins were updated so that they could handle the capture_output argument. However, this functionality was only implemented for the plugins:
- ssh
- local
- jail
- chroot

For the other connection plugins: (accelerate, fireball, funcd, paramiko_alt, paramiko_ssh, ssh_old), setting `capture_output=False` will raise an exception.
## Adding support for "show_output" option

Adding the new "show_output" option on plays, task, and globally required changes to:
- example/ansible.cfg
- lib/ansible/constants.py
- lib/ansible/playbook/play.py
- lib/ansible/playbook/task.py


mpdehaan:
So this is a rather large pull request, can you start a thread on ansible-devel to discuss?

Some initial preferences would be to make it a per-task option rather than a global, and it will need some more comments, though I'd like to discuss implementation details first.

lorin:
Will do. Note that it currently is per-task, but you can also specify it per-play or globally.

utapyngo:
Could you please share a link to the discussion?

lorin:
@utapyngo https://groups.google.com/d/topic/ansible-devel/cvbyUMspqE0/discussion

utapyngo:
Thank you @lorin. It does not seem to be discussed anymore.
Why is not this pull request still merged?

lorin:
@utapyngo See https://github.com/ansible/ansible/issues/3887#issuecomment-39005404

jimsmith:
Can this merge conflict be resolved as this realtime output I really do need for dev team and for us engineers, as presently I need to wait a few minutes for tasks to run to determine if they have completed or error  out and some tasks take more than a few minutes to run.

lorin:
@jimsmith I don't think this particular pull request is ever going to be merged upstream (see https://github.com/ansible/ansible/issues/3887#issuecomment-39005404). Are you still interested in it?

utapyngo:
Not in this particular pull request, but in the feature it was supposed to implement.

jimsmith:
@lorin I am very much interested in this feature as having realtime output is vital for my needs.

Thanks :)

lorin:
@jimsmith @utapyngo I'm afraid that I don't have the cycles to work on an implementation that would be consistent with the [suggested approach](https://groups.google.com/d/msg/ansible-devel/cvbyUMspqE0/DMLaYR8DfUoJ). I don't think the code in its current form will get merged, so I'm going to close this out for now.

PR #6036, opened 2014-02-16
docs_pull_request: Add parameters highlighting

mxxcon:
Added highlighting to parameter references in docs so that they standout a bit better.


mpdehaan:

PR #6037, opened 2014-02-17
Add troubleshooting page to docs

holyjak:
A start of a page with few troubleshooting tips.
Hopefully others will help to grow this.

### DIFF
--- a/docsite/rst/playbooks.rst
+++ b/docsite/rst/playbooks.rst
@@ -24,5 +24,6 @@ It is recommended to look at `Example Playbooks <https://github.com/ansible/ansi
    playbooks_conditionals
    playbooks_loops
    playbooks_best_practices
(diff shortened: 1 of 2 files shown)


mpdehaan:
While I'm not against a troubleshooting section, I do have opinions about what it contains.

In particular, the way to troubleshoot modules is not KEEP_REMOTE_FILES but ./hacking/test-module on the remote system.

I'll keep this in mind but this is really something we should incorporate into the FAQ instead.

holyjak:
Well, this was just a proposal. I think it is better to come with something than just open a improvement request that nobody will have time to work on.

I really hope that some kind of troubleshooting page will be added _in a near future_, it will help new users like me immensely.

BTW It was not about troubleshooting modules but playbooks; but perhaps that is the same thing?

PR #6044, opened 2014-02-17
Ansible vault: a framework for encrypting any playbook or var file.

jctanner:
```
$ ansible-vault --help
Usage: ansible-vault [create|decrypt|edit|encrypt|rekey] [--help] [options] file_name

Options:
  -h, --help  show this help message and exit

See 'ansible-vault <command> --help' for more information on a specific command.
```
- create: start a new encrypted file
- decrypt: unencrypt file(s) in place
- edit: open an encrypted file in the default editor
- encrypt: encrypt file(s) in place
- rekey: decrypt with previous password, re-encrypt with new password

```
$ ansible-playbook -i inventory site.yml --ask-vault-pass
Vault password:
```

### DIFF
--- a/bin/ansible-playbook
+++ b/bin/ansible-playbook
@@ -102,6 +102,7 @@ def main(args):
     su_pass = None
     if not options.listhosts and not options.syntax and not options.listtasks:
         options.ask_pass = options.ask_pass or C.DEFAULT_ASK_PASS
+        options.ask_vault_pass = options.ask_vault_pass or C.DEFAULT_ASK_VAULT_PASS
         # Never ask for an SSH password when we run with local connection
         if options.connection == "local":
             options.ask_pass = False
--- a/bin/ansible-playbook
+++ b/bin/ansible-playbook
@@ -62,6 +62,8 @@ def main(args):
         check_opts=True,
         diff_opts=True
     )
+    #parser.add_option('--vault-password', dest="vault_password",
+    #    help="password for vault encrypted files")
     parser.add_option('-e', '--extra-vars', dest="extra_vars", action="append",
         help="set additional variables as key=value or YAML/JSON", default=[])
     parser.add_option('-t', '--tags', dest='tags', default='all',
@@ -105,7 +107,8 @@ def main(args):
             options.ask_pass = False
         options.ask_sudo_pass = options.ask_sudo_pass or C.DEFAULT_ASK_SUDO_PASS
         options.ask_su_pass = options.ask_su_pass or C.DEFAULT_ASK_SU_PASS
-        (sshpass, sudopass, su_pass) = utils.ask_passwords(ask_pass=options.ask_pass, ask_sudo_pass=options.ask_sudo_pass, ask_su_pass=options.ask_su_pass)
+        #import epdb; epdb.st()
+        (sshpass, sudopass, su_pass, vault_pass, __) = utils.ask_passwords(ask_pass=options.ask_pass, ask_sudo_pass=options.ask_sudo_pass, ask_su_pass=options.ask_su_pass, ask_vault_pass=options.ask_vault_pass)
         options.sudo_user = options.sudo_user or C.DEFAULT_SUDO_USER
         options.su_user = options.su_user or C.DEFAULT_SU_USER
 
@@ -170,7 +173,8 @@ def main(args):
             diff=options.diff,
             su=options.su,
             su_pass=su_pass,
-            su_user=options.su_user
+            su_user=options.su_user,
+            vault_password=vault_pass
         )
 
         if options.listhosts or options.listtasks or options.syntax:
--- a/bin/ansible-playbook
+++ b/bin/ansible-playbook
@@ -108,7 +108,7 @@ def main(args):
             options.ask_pass = False
         options.ask_sudo_pass = options.ask_sudo_pass or C.DEFAULT_ASK_SUDO_PASS
         options.ask_su_pass = options.ask_su_pass or C.DEFAULT_ASK_SU_PASS
-        (sshpass, sudopass, su_pass, vault_pass, __) = utils.ask_passwords(ask_pass=options.ask_pass, ask_sudo_pass=options.ask_sudo_pass, ask_su_pass=options.ask_su_pass, ask_vault_pass=options.ask_vault_pass)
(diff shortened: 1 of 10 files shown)


sivel:
I may have missed it here, but will there be an ansible.cfg configuration option for ask_vault_pass ?

jctanner:
@sivel I'll look into adding that. For now, it produces an error if any file is encrypted and no password is given.

**FIXED**

mpdehaan:
Hi everybody -- There will be a ansible-project mailing list thread about this shortly -- we can answer questions on that thread once we get it up there.

mpdehaan:
We've talked about adding "soft imports" so the crypto libs aren't brought in unless used, and will be adding this as a package dependency

mpdehaan:
Possibly some things going on with pycrypto from pip and RHEL underlying libs....

mpdehaan:
If I don't have the library installed I get:

python Crypto.Foo.Bar.Something.AES is not installed

This should probably say something like "pycrypto is not installed.  Fix this with your package manager, for instance, yum-install python-crypto OR (apt equivalent)"

**FIXED**

mpdehaan:
Installing from pip seems to be able to generate a version that is incompatible with the EL6 guest.  We may have to intercept that error and warn about it.   (testing this shortly)

mpdehaan:
ansible-vault should probably print "encryption successful" or "decryption successful" for commands that don't spawn an editor.

**FIXED**

The message "add more entropy" should probably say "add more entropy (suggest using mouse inputs".  I'm guessing some people won't know what entropy sources are.

**FIXED**

mpdehaan:
The message on an invalid decryption should mention the decryption failed

[root@localhost ansible]# ansible-playbook /tmp/test.yml -i /tmp/hosts -c local --ask-vault-pass
Vault password: 
ERROR: A vault password must be specified to decrypt /tmp/test2.yml

**FIXED**

mpdehaan:
Just got that from entering the password correctly with vars_files.

mpdehaan:
When specifying an invalid password with vars_files, I get the message:

fatal: [localhost] => A vault password must be specified to decrypt /tmp/group_vars/region

This should be changed to be more clear

"Vault password incorrect for /tmp/group_vars/region"

As above, it's unclear if a password was supplied or not, and we do know.

**FIXED**

mpdehaan:
I'm also unable to get group-vars working with the correct password so this seems generic on my platform

mpdehaan:
password on encrypt or create should probably be a double prompt

**FIXED**

mpdehaan:
and also rekey

mpdehaan:
We also want to add pycrypto to the dependencies for at least the RPM and dpkg builds.

mpdehaan:
Can remove "if 1" here:

https://github.com/jctanner/ansible/commit/98f079f62a7ba6f0325977252ff01604e29eaf31#diff-b270ccfb9473766a7e7790c7ff58f277R169

**FIXED**

mpdehaan:
This check should probably be explicitly against None as empty string is also False:
-    if not password:

**FIXED**

mpdehaan:
(Note: earlier comment about pycrypto from pip is not a thing, correctly refused to install, which is good)

mpdehaan:
ansible-vault create should not the start the file with "---" and nothing should assume "---" must be in the file as we may want to also support JSON files.

**FIXED**

bcoca:
testing on Ubuntu, works nicely so far, aside from already made comments
for password verification (you've seen how often i mysytpe).

mpdehaan:
For rekey, should not ask for new password if decrypt on old password fails, should also double up on the new password prompt.

mpdehaan:
ERROR:  global name 'out path' is not defined

this is what you get when you try to rekey an unencrypted file.  Should instead error out that the file was unencrypted

**FIXED**

mpdehaan:
The "-n" parameter for rekey should be removed (was removed elsewhere so this is just a straggler)

**FIXED**

bcoca:
When using edit, I would try delete/unlink the tempfile, even though
permissions are already restrictive (stdin to vim?).

Also is the tmp dir configurable? I am partial to ~/.cache.

mpdehaan:
in utils/vault.py this can be simplified:

```
# read first line of the file
with open(filename) as f:
    head=[f.next() for x in xrange(1)]
```

To:

```
head = f.next()
```

and then later just use head vs head[0]

Easier might just be to use .read with a byte limit and use ".startswith"

**FIXED**

mpdehaan:
per code standards, the instance Vault should be assigned to a variable "V", should be named "vault = Vault(...)" or "my_vault", etc.

**FIXED**

Also if Vault is a useful class we should not be calling private methods on it.   (V._strip_sha).   Should this funciton be named "decrypt?" or something similar?

mpdehaan:
encrypt seems to assume writing to a file.    There should be an in-memory version for unit test purposes and a seperate function/abstraction/layer that writes to files.

(consider maybe using a cStringIO)

mpdehaan:
various "f = open" ... "f.close" logic could be written in utils/vault.py to use the "with" operator.

mpdehaan:
In "_strip_sha", instead of "del_lines[0]" do 

'\n'.join(lines[1:])

**FIXED**

mpdehaan:
There is support for a version in the header string but the code doesn't have a way to know if the version of a future vault is "too new".

The code should have a constant "VAULT_VERSION_MAX" or similar and raise an error if the header exceeds it.  Similarly probably also a "VAULT_VERSION_MIN".   For the initial release both of these could be 1.0

**FIXED**

bcoca:
also, could there be a choice of encryption methods? AES is nice but
sometimes policy requires us to use specific suites and lengths .

mpdehaan:
We just noticed gather_facts: no affects it a little.  WIP.

(Need to plumb through to Runner in ansible-playbook)

**FIXED**

bcoca:
ansible-vault create file.test.yml, creates the encrypted file even if I
quit out of the editor, I would expect it not to.

jctanner:
@bcoca the cipher is "pluggable" so that contributors can submit others. Check how the "AES" class is loaded.

jctanner:
@bcoca "creates the encrypted file even if I quit out of the editor," ...

How are you "quitting"?

bcoca:
On Mon, Feb 17, 2014 at 5:52 PM, jctanner notifications@github.com wrote:

> the encrypted file even if I quit out of the editor," ...
> 
> How are you "quitting"?

:q! <= not a smiley

## 

Brian Coca
Pedo mellon a minno

jctanner:
Superseded by https://github.com/ansible/ansible/pull/6058

PR #6045, opened 2014-02-17
Set use_mirrors default to "no" for the pip module

TallGirlVanessa:
Also deprecate `use_mirrors`, if that's ok! I hope I'm not overstepping my bounds; it just seems like a good idea.

Anyway, PyPI moved to a CDN on 2013-05-26, so mirrors are now largely unnecessary and pip removed the functionality in 1.5. More importantly (and why I wrote this request): on 2014-02-15 the mirror directory was taken offline, so mirrors may not work anymore even for pip versions that support them. Somehow one of my machines was able to install things anyway (using pip 1.0), but another one of my machines hung indefinitely when trying to install a new version of a package.

### DIFF
--- a/library/packaging/pip
+++ b/library/packaging/pip
@@ -287,8 +287,6 @@ def main():
     if env:
         path_prefix="/".join(pip.split('/')[:-1])
 
-    if extra_args:
-        cmd += ' %s' % extra_args
     if name:
         # pip can accept a path to a local project or a VCS url beginning
         # with svn+, git+, hg+, or bz+ and these sources usually do not qualify
@@ -326,6 +324,10 @@ def main():
 
         if not is_package and state != 'absent' and use_mirrors and '--use-mirrors' in cmd_opts:
             cmd += ' --use-mirrors'
+
+    if extra_args:
+        cmd += ' %s' % extra_args
+    if name:
         cmd += ' %s' % _get_full_name(name, version)
     elif requirements:
         cmd += ' -r %s' % requirements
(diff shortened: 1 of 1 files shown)


mpdehaan:
What should be done here IMHO is to keep the argument to "eat" the parameter and remove the support for mirrors from the code.

TallGirlVanessa:
I think I've fixed it. I also think I fixed an unrelated problem: it looks like -e is automatically added to extra_args when pip installing a VCS URL, but by that time extra_args has already been added to cmd and isn't read again.

mpdehaan:
I think we should also remove use_mirrors from the comments, as it's not deprecated, it's gone.

Can you update this? 

Thanks!

TallGirlVanessa:
Yeah, I was wondering while I was writing it if so much documentation for a feature you guys were removing was a good idea :smiley: I've removed it.

mpdehaan:
Merged, thanks!

TallGirlVanessa:
NP, thanks for the feedback!

dhozac:
Adding -e breaks existing deployment playbooks that expect it to not be set, especially as there is no way to turn it off.

mpdehaan:
Hi @dhozac !!!

TallGirlVanessa:
Sorry about that; I noticed the examples for the module say VCS URLs automatically get -e and that somewhere along the line that logic broke, so I "fixed" it. But yeah, not being able to turn it off kinda sucks.

What's the protocol for when a merged PR has a bug? Should I open a new one?

PR #6046, opened 2014-02-17
Avoid breaking on unicode input when logging to syslog in modules

mtr:
After commit 254f87e, non-ascii input broke logging to syslog.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -814,10 +814,10 @@ class AnsibleModule(object):
             except IOError, e:
                 # fall back to syslog since logging to journal failed
                 syslog.openlog(str(module), 0, syslog.LOG_USER)
-                syslog.syslog(syslog.LOG_NOTICE, msg.encode('utf8'))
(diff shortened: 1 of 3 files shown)


mpdehaan:
Thanks for this.

Out of curiosity can you translate the playbook for those of us whom Google Translate does not work so well?  :)

mpdehaan:
cc @jimi-c can you take a look at this?

jimi-c:
Merged, thanks!

mpdehaan:
Forgive me for not inputing the right source language!  :)

mtr:
Thanks for accepting the pull request.  

I didn't put much effort into coming up with meaningful test strings; just ensured that the strings included some non-ASCII characters. I'll translate the different strings here:

## Norwegian

"Loggføring fungerer" = "Logging works"
"Feilsøking" = "Debugging"

## German

"Die Süßigkeit" = "the candy/sweet"

If you want me to, I can of course create a pull request with similar comments in the playbook.

jimi-c:
Nope, that won't be necessary. Thanks again!

PR #6051, opened 2014-02-18
Fix zfs create command for properties with spaces

illenseer:
If a property value contains one or more spaces, the zfs command will fail.
With value quoted this behavior is fixed.

The problem occurrs while creating a new ZFS dataset, e.g.:

```
zfs: name=tank/test state=present sharenfs="-ro 10.0.0.0/13"
```


mpdehaan:
merged, thanks!

PR #6053, opened 2014-02-18
corrections for Arch PKGBUILD

cgtx:
Use correct permission mode when installing files and directories.  Also, create the group_vars and host_vars directories, since the ansible documentation states that these are best practice.

http://docs.ansible.com/intro_inventory.html#splitting-out-host-and-group-specific-data

### DIFF
--- a/packaging/arch/PKGBUILD
+++ b/packaging/arch/PKGBUILD
@@ -39,8 +39,6 @@ package() {
   python2 setup.py install -O1 --root="$pkgdir"
 
   install -Dm644 examples/ansible.cfg $pkgdir/etc/ansible/ansible.cfg
-  install -dm755 $pkgdir/etc/ansible/group_vars
-  install -dm755 $pkgdir/etc/ansible/host_vars
 
   install -Dm644 README.md $pkgdir/usr/share/doc/ansible/README.md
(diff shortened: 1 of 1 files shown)


mpdehaan:
As I'm not using Arch, can you please explain permissions before and after and the logic behind the changes above?

On the second point, we do not want to pre-create group_vars or host_vars directories as this would be inconsistent with other package managers, and also the inventory file is relocatable.   Some users also keep group_vars and host_vars alongside their playbooks.

leucos:
Perms changes looks good to me :+1:

Agreed with @mpdehaan on {host,group}_vars. I don't think many people use these under /etc anyway. It would require to sudo all the time for edits. I suppose almost everyone relocates and uses -i or `hostfile` in ansible.cfg.

cgtx:
Thanks for the quick feedback!

@mpdehaan Without explicit permissions, /etc/ansible/ansible.cfg was getting created as executable, which is unnecessary.  I replaced the mkdir's with install's just as a best practice, since mkdir would be affected by a user's custom umask.  Again, better to be explicit.

I'll yank out the {host,group}_vars and add another commit to this pull request.

mpdehaan:
Went ahead and merged this in, thanks!

Issue #5914, opened 2014-02-08
copy with contents fails if contents are json with variable substitutions.


kvdveer:
```
  - copy:
      content: '{"a": "{{ inventory_hostname }}"}'
      dest: /root/it.json
```

Raises an error, even before the module is executed:

```
<213.187.240.154> ESTABLISH CONNECTION FOR USER: root
failed: [services-1.foo] => {"failed": true}
msg: could not write content temp file: expected a character buffer object
```

I can trigger this issue with any valid json that also contains variables, making it invalid json (by prepending some garbage) or removing the variable substitution works as expected.

Versions: 1.4.4 and latest head

ansibot:
Thanks for filing a ticket! I am the friendly GitHub Ansibot. I see you did not fill out your issue description based on our new issue template. Please copy the contents of this [template](https://raw2.github.com/ansible/ansible/devel/examples/issues/ISSUE_TEMPLATE.md) and **paste it into the description** of your ticket. 

Our system will automatically close tickets that do not have an issue template filled out within 7 days. Please note that due to large interest in Ansible, humans may not comment on your ticket if you ask them questions. Don't worry, you're still in the queue and the robots are looking after you.

mpdehaan:
cc @risaacson

risaacson:
@kvdveer thanks for submitting this. I'll dig into it and get it fixed quickly.

risaacson:
@kvdveer Is this something that you have had working before? I am testing on a Ubuntu 13.10 desktop to a CentOS 5 server. I have gone back to 1.4.0 and I don't see that this worked. Is this your first time trying this or have you seen it work before?

kvdveer:
I have never seen this working, but I only stepped in during 1.4.3, and wote my first json-based copy command yesterday. 

Is there anything I can do thelp diagnose/resolve this? I'm deploying to Ubuntu 12.04, running from Ubuntu 13.10.

kvdveer:
v1.3.4: seems to work as expected
v1.4.0: show the aforementioned symptoms

I've tried bisecting the issue, but it doesn't seem to be entirely determinisic as bisect yields nonsense results. I'll try to work with clean venvs, possiby that helps, bit for now it's bedtime... ;-)

kvdveer:
Sleeping helped... ;-)
According to my git bisect, the culprit is this: 90cce35919e6b93c0b904078c5285e1dbaeb0fdf

risaacson:
[...]

PR #6057, opened 2014-02-18
Copy action_plugin: encode content when dict.

risaacson:
When content is processed and found to be valid JSON it is decoded into a dict. To write it out to a file we need to encode it back into a string.

Addresses GH-5914.


mpdehaan:
This works for me.

PR #6062, opened 2014-02-19
module update: mqtt notification now uses Paho as mosquitto.py being deprecated

jpmens:
The Python mosquitto module [is being deprecated in favor of Paho](http://mosquitto.org/2013/12/paho-mqtt-python-client/), so I've converted the `mqtt` notification module to use that.

### DIFF
--- a/library/notification/mqtt
+++ b/library/notification/mqtt
@@ -1,7 +1,7 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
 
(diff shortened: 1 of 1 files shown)


jpmens:
For testers: as mentioned in the module DOCUMENTATION, requires https://pypi.python.org/pypi/paho-mqtt

mpdehaan:
Merged,thanks!

PR #6065, opened 2014-02-19
`docker pull' image instead of building if no path is given

hkariti:
Instead of failing, give the option to pull the image from index.docker.io.

### DIFF
--- a/lib/ansible/runner/action_plugins/assert.py
+++ b/lib/ansible/runner/action_plugins/assert.py
@@ -17,7 +17,7 @@
 
 import ansible
 
-from ansible import utils
+from ansible import utils, errors
 from ansible.runner.return_data import ReturnData
 
 class ActionModule(object):


sivel:
This is great...I was working on an almost identical pull request earlier this morning.

One thing though.  Don't import json.  json is already made available by module_utils/basic.py and handles the case of importing simplejson if needed.

Also, you have mixed use of tabbed indentation and space indentation.  Make sure to use all spaces.  Lines with tabs are 134 and 136.

mpdehaan:
Yep, please fix the tabs above and we can look at merging this in.

hkariti:
I hope I used json right this time..

sivel:
That import will not function in most cases.

There is no need for any import at all. JSON is already available to you without any explicit import. You can just remove that whole import line.

hkariti:
Hm, it worked for me with the import just now. And I'm a bit uncomfortable with the json-is-available-using-magic. But if you say that's the way, alright.

sivel:
Another issue I see here.  It looks like you added a file called `plugins/callbacks/log_json.py` to this branch and pushed it.  Please ensure that you use separate branches for separate features.

The "magic" is due to how modules are built, which I'll keep out of the conversation now, but the import at the bottom of the file enables this.

hkariti:
Whoops. Sorry about that.

sivel:
I've tested, and can confirm things are working as I would expect.

sivel:
I'd like to see this for 1.5, just a heads up that there appears to be pulls to update all modules to require docker-py 0.3.0 so, this pull could require updating depending on which makes it in first.

hkariti:
What kind of updates? Did they change the API?

softzilla:
I have already added 0.3.0 support at #6196.

hkariti:
Had a mess with branches, sent this again with a seperate branch as #6247.

PR #6069, opened 2014-02-19
Update homebrew module.

danieljaouen:
Flesh out the homebrew packaging module.


mpdehaan:
Removing install_options would be a backward incompatible change, can you take care of this one?

Also we favor present and absent and should not remove working aliases.

danieljaouen:
@mpdehaan These issues have been fixed.

mpdehaan:
Merged, thank you!

Issue #6061, opened 2014-02-19
Modules "unarchive" and "copy" do not expand pathes correctly


muffl0n:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.4.5 
##### Environment:

Gentoo Base System release 2.2
##### Summary:

This simple playbook:

```
- hosts: localhost
  tasks:

  - name: ensure group test is present
    group: name=test gid=1111

  - name: ensure user test is present
    user: name=test uid=1111 home=/tmp/test

  - file: path=~test/test owner=test group=test mode=0644 state=directory

  - copy: src=test.txt dest=~test/test.txt owner=test group=test mode=0600

  - unarchive: src=test.tar.gz dest=~test
```

produces an error when being executed:

```
$ ansible-playbook test.yml 

PLAY [localhost] ************************************************************** 

GATHERING FACTS *************************************************************** 
ok: [localhost]

TASK: [ensure group test is present] ****************************************** 
ok: [localhost]

TASK: [ensure user test is present] ******************************************* 
changed: [localhost]

TASK: [file path=~test/test owner=test group=test mode=0644 state=directory] *** 
changed: [localhost]

TASK: [copy src=test.txt dest=~test/test.txt owner=test group=test mode=0600] *** 
changed: [localhost]

TASK: [unarchive src=test.tar.gz dest=~test/] ********************************* 
failed: [localhost] => {"failed": true}
msg: dest must be an existing dir

FATAL: all hosts have already failed -- aborting

PLAY RECAP ******************************************************************** 
           to retry, use: --limit @/home/svs/test.retry

localhost                  : ok=5    changed=3    unreachable=0    failed=1
```

Apparently unarchive does not expand the home directory "~test/" as file and copy do. Despite the fact that the directory is expanded in https://github.com/ansible/ansible/blob/devel/library/files/unarchive#L176 the corresponding action_plugin let's it fail before the expansion: https://github.com/ansible/ansible/blob/devel/lib/ansible/runner/action_plugins/unarchive.py#L66

I'm quite new to ansible and maybe I mixed up my conclusion. But the error exists and can only be fixed if I change 

```
- unarchive: src=test.tar.gz dest=~test
```

to

```
- unarchive: src=test.tar.gz dest=/tmp/test
```

[...]

PR #6073, opened 2014-02-19
Fixes "unarchive" and "copy" do not expand pathes correctly

muffl0n:
Proposed fix for #6061. 
I'm new to Ansible and Python, so please be gentle. :)

### DIFF
--- a/lib/ansible/runner/action_plugins/copy.py
+++ b/lib/ansible/runner/action_plugins/copy.py
@@ -48,9 +48,9 @@ class ActionModule(object):
         if complex_args:
             options.update(complex_args)
         options.update(utils.parse_kv(module_args))
-        source  = options.get('src', None)
+        source  = os.path.expanduser(options.get('src', None))
         content = options.get('content', None)
-        dest    = options.get('dest', None)
+        dest    = os.path.expanduser(options.get('dest', None))
         raw     = utils.boolean(options.get('raw', 'no'))
         force   = utils.boolean(options.get('force', 'yes'))
 
--- a/lib/ansible/runner/action_plugins/unarchive.py
+++ b/lib/ansible/runner/action_plugins/unarchive.py
@@ -46,8 +46,8 @@ class ActionModule(object):
         if complex_args:
             options.update(complex_args)
         options.update(utils.parse_kv(module_args))
-        source  = options.get('src', None)
-        dest    = options.get('dest', None)
+        source  = os.path.expanduser(options.get('src', None))
+        dest    = os.path.expanduser(options.get('dest', None))
         copy    = utils.boolean(options.get('copy', 'yes'))
 
         if source is None or dest is None:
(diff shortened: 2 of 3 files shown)


bcoca:
you really don't want to expanduser locally for dest, in case the user
doesn't exist or has a different home path.

muffl0n:
Sounds reasonable. Unfortunately I have too little experience with Ansible to really understand how to fix it correctly.

jimi-c:
`expanduser` actually correctly expands based on the correct home path, and if the user doesn't exist it just returns the string as-is (for example, `~foo` doesn't change if the user foo doesn't exist). So I think this is fine and doesn't need to be changed.

jimi-c:
Merged, with one small change - I moved the expanduser calls down as otherwise they were sometimes being passed in `None` when the parameter wasn't set, which caused a traceback. Thanks!

muffl0n:
Thanks for merging!

bcoca:
you still have the problem when the user exists and the homes are different across machines. It is a rare corner case, but I've had it happen.

muffl0n:
So maybe one should expand the src locally and the dest on the target machine?
Today I learned that the action_plugins are called locally while "library/files/copy", for example, is copied to the target machine and executed there. Is this correct?

jimi-c:
Yes, I just had to fix it so that the expanduser call was done in the module, so it should be removed from the action_plugin. I'll take care of it.

muffl0n:
@jimi-c Did you have a chance to take a look at this? I tried to create a bugfix on my own but unfortunately didn't succeed. 

Though I found something to care about when fixing this:
homedir of local user "test": /tmp/test
homedir of remote user "test": /tmp/testuser
Playbook "test-unarchive.yml"

```
- hosts: remote
  tasks:
    - unarchive: src=~test/test.tar.gz dest=~test copy=no
      ignore_errors: True
    - unarchive: src=~test/test.tar.gz dest=~test copy=yes
      ignore_errors: True
```

```
$ ansible-playbook -i test test-unarchive.yml 

PLAY [remote] ***************************************************************** 

ok: [remote]

TASK: [unarchive src=~test/test.tar.gz dest=~test copy=no] ******************** 
failed: [remote] => {"failed": true}
msg: dest '/tmp/test' must be an existing dir, src: /tmp/test/test.tar.gz
...ignoring

TASK: [unarchive src=~test/test.tar.gz dest=~test copy=yes] ******************* 
failed: [remote] => {"failed": true}
msg: dest '/tmp/test' must be an existing dir, src: /tmp/test/test.tar.gz
...ignoring

remote                     : ok=3    changed=0    unreachable=0    failed=0
```

As one can see, dest is always expanded locally.
Whether or not you have to expand the src on the locale machine depends on the parameter copy. This parameter is described by "Should the file be copied from the local to the remote machine?".
So if the file is to be copied from the local machine the source has to be expanded locally. Otherwise it has to be expanded on the remote machine.
Currently it is always expanded locally which is wrong if "copy=yes".

PS: I changed unarchive.py to show the debug output above in 958ff640ccc77d092e982520f37bc7a44a93c928

PR #6081, opened 2014-02-20
Added support for tagging subnets in ec2_vpc module

ap1kenobi:
Tagging a subnet allows us to reference it later without using/knowing its id.  For example, if we tag subnets as tiers such as Web, App and DB we can later use this metadata to provision VMs into appropriate subnets:

``` YAML
# roles/ec2-provisioner/tasks/main.yml

---
- name: "Create Virtual Private Cloud (VPC)"
  local_action:
    module: ec2_vpc
    state: present
    cidr_block: 10.0.0.0/16
    subnets:
      - cidr: 10.0.1.0/24 
        az: us-east-1a   
        resource_tags: { "Environment":"DEV", "Tier" : "Web" }
      - cidr: 10.0.2.0/24
        az: us-east-1a       
        resource_tags: { "Environment":"DEV", "Tier" : "App" }
      - cidr: 10.0.3.0/24
        az: us-east-1a    
        resource_tags: { "Environment":"DEV", "Tier" : "DB" }
        ...
  register: env_vpc

```

### DIFF
--- a/library/cloud/ec2_vpc
+++ b/library/cloud/ec2_vpc
@@ -277,7 +277,10 @@ def create_vpc(module, vpc_conn):
                 add_subnet = False
         if add_subnet:
             try:
-                vpc_conn.create_subnet(vpc.id, subnet['cidr'], subnet.get('az', None))
+                created_subnet = vpc_conn.create_subnet(vpc.id, subnet['cidr'], subnet.get('az', None))
+                subnet_tags = subnet.get('tags', None)
+                if subnet_tags:
+                    vpc_conn.create_tags(created_subnet.id, subnet_tags)
                 changed = True
             except EC2ResponseError, e:
                 module.fail_json(msg='Unable to create subnet {0}, error: {1}'.format(subnet['cidr'], e))
--- a/library/cloud/ec2_vpc
+++ b/library/cloud/ec2_vpc
@@ -280,10 +280,18 @@ def create_vpc(module, vpc_conn):
                 add_subnet = False
         if add_subnet:
             try:
-                created_subnet = vpc_conn.create_subnet(vpc.id, subnet['cidr'], subnet.get('az', None))
-                subnet_tags = subnet.get('tags', None)
-                if subnet_tags:
-                    vpc_conn.create_tags(created_subnet.id, subnet_tags)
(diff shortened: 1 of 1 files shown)


deoxxa:
This would be immediately useful for me, and would make #6127 irrelevant, as this is a much better approach to solving the same problem.

pas256:
This addition makes sense.

I don't like the option called `instance_tags` though because you are tagging the subnet, not the instances running inside the subnet. Perhaps `resources_tags` or specifically `subnet_tags` is better.

Again, this is in some sense opening a can of worms, because tagging route tables, ACLs, IGWs, DHCP options should also be supported by this module.

ap1kenobi:
I'm not attached to the name - just wanted to stay consistent (a la ec2 module).  I like `resource_tags` because it's more generic sounding than `subnet_tags`, `vpc_tags`, `dhcp_tags`, etc. - less syntax/config names to learn.  I'll change the name tomorrow.

I use tags obsessively in AWS for various reasons and actually would like built-in tagging capability in all ec2\* modules as opposed to using ec2_tag.  I don't mind getting this propagated across the rest of the modules.  In fact, I asked if anyone would mind if I did on ansible-project google group...

pas256:
I completely agree!

ap1kenobi:
grr...  this was not intentional

mpdehaan:
This seems to be related to #6085, can (or should) these pull requests perhaps be unified?

I'm a bit unclear given the similarities.

Thanks!

ap1kenobi:
@mpdehaan, #6085 and this could be combined if desired but I intentionally kept them separate to avoid dependencies.  
This PR adds "resource_tags" to allow "inline" tagging of VPC subnets, while #6085 adds "resource_tags" to tag the VPC itself, among other things.  
In other words, they are related by being in the same module but their respective effects are different and independent of one another.

mpdehaan:
If you can unify these and present these as a squashed commit that might be easiest for review.

Thanks!

mpdehaan:
Actually looks like we can merge this as is, doing this now, thanks!

ap1kenobi:
Sorry, I just noticed you earlier request to squash...  Thanks for merging in.

PR #6084, opened 2014-02-20
Fix mistaken double backslash

dsedivec:
The module helper function run_command was appending a literal backslash
followed by 'n' to the stdin of a command it runs unless you called it
with binary_data=True (not the default).  I have changed it to what
I expect was the intent, to append an actual line feed to stdin.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -1039,7 +1039,7 @@ class AnsibleModule(object):
             
             if data:
                 if not binary_data:
-                    data += '\\n'
+                    data += '\n'
             out, err = cmd.communicate(input=data)
             rc = cmd.returncode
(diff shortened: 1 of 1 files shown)


mpdehaan:
I think there's a chance that this is there such that output from long command streams like yum are easily presentable in valid JSON coming out the other end, though this system may also escape.  We should be careful in testing.

dsedivec:
In case this helps I thought I should reiterate that this `\n` is getting appended to the standard _input_ of the program being run.  I don't immediately see that this has any bearing on the program's output--but mpdehaan's point may have gone right over my head.

mpdehaan:
We were going through this code the other day and discovered nothing is apparently using this, which makes sense, given the original did not make any sense :)

mpdehaan:
merged, thanks!

PR #6085, opened 2014-02-20
Added VPC tagging and ability to provision multiple VPCs with the same CIDR

ap1kenobi:
Built-in tagging support a la `ec2`'s `instance_tags`.

This change also allows provisioning of multiple VPCs with the same CIDR blocks provided unique tags are used to differentiate between them.

The following two tasks will apply to two different VPCs:  

``` YAML
- name: "Create Virtual Private Cloud (VPC) - DEV"
  local_action:
    module: ec2_vpc
    resource_tags: { "Environment":"DEV" }
    state: present
    region: us-east-1
    cidr_block: 10.0.0.0/16
    ...
```

``` YAML
- name: "Create Virtual Private Cloud (VPC) - DEMO"
  local_action:
    module: ec2_vpc
    resource_tags: { "Environment":"DEMO", "BillingCode":"XXM1234"}
    state: present
    region: us-east-1
    cidr_block: 10.0.0.0/16
    ...
```

In the above example, if CIDR/Tag combination does not exist new VPC will be created.  

~~_Note, that newly created VPCs will then need to be tagged in order to avoid creating a new VPC every time this task is played.  This module does not track or enforce uniqueness of tags across VPCs._~~  

Newly created VPC will be tagged using tags from the `resource_tags` list.  Tags not on `resource_tags` list will be ignored.

Deleting a VPC by supplying a list of tags is also possible as long as its CIDR block is also provided:

``` YAML
- name: "Delete Virtual Private Cloud (VPC) - DEV"
  local_action:
    module: ec2_vpc
    resource_tags: {"Environment": "DEV"}
    cidr_block: 10.0.0.0/16
    region: us-east-1
    state: absent
  register: env_vpc
```

### DIFF
--- a/library/cloud/ec2_vpc
+++ b/library/cloud/ec2_vpc
@@ -56,6 +56,12 @@ options:
     required: false
     default: null
     aliases: []
+  vpc_id_tags:
+    description:
+      - A list of tags uniquely identifying a VPC in the form of: {Tag1: Value1, Tag2: Value2, ...}. This list works in conjunction with CIDR (cidr_block) and is gnored when VPC id (vpc_id) is specified.
+    required: false
+    default: null
+    aliases: []
   internet_gateway:
     description:
       - Toggle whether there should be an Internet gateway attached to the VPC
@@ -127,6 +133,7 @@ EXAMPLES = '''
         module: ec2_vpc
         state: present
         cidr_block: 172.23.0.0/16
+        vpc_id_tags: { "Environment":"Development" }
         region: us-west-2
 # Full creation example with subnets and optional availability zones.
 # The absence or presense of subnets deletes or creates them respectively.
@@ -134,6 +141,7 @@ EXAMPLES = '''
         module: ec2_vpc
         state: present
         cidr_block: 172.22.0.0/16
+        vpc_id_tags: { "Environment":"Development" }
         subnets: 
           - cidr: 172.22.1.0/24
             az: us-west-2c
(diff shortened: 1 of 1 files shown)


pas256:
`vpc_id_tags`? Not just `vpc_tags`? What was your reasoning behind that?

ap1kenobi:
Short answer: because it's only a subset of tags actually assigned to the object, intended to be a composite ID of sorts.  In other words: it's used for identification not for assignment.

I use VPCs as containers for various stages of the product I'm working on: Dev (main and discreet branches), Test (system, functional, performance, etc.), Demo, and others, intentionally keeping environment configuration the same, including the CIDR block.

Unfortunately, Ansible uniquely identifies VPCs by their CIDR so you can't have two VPCs with 10.0.0.0/16 CIDR.  So, I added a list of tags to be used in conjunction with CIDR block to find the right VPC.  The code will settle on a VPC when it finds a subset of tags matching the supplied list.

My thinking has changed since I began using this myself and I'd actually prefer to be able to tag a VPC in the same task that creates it.

However, if I use `resource_tags` as you suggested in a comment to #6081 for assigning tags to a VPC, do I use the same list to ID it?  This would imply that this list must remain immutable, otherwise a new VPC will be created every time one changes list values.

pas256:
I see. Thanks for the explanation.

In your use case, the tag is only an ID because you are using it that way. Its your convention (and a very reasonable one at that). However, `tags` is a plural, and the technology supports having multiple key/value pairs, so what is considered an ID for you may not be for the next person using it.

The term `resource_tags` is specific to what the tags are describing, and can be used for all levels of options in the VPC module (subnets, ACLs, etc) since I saw a PR for that yesterday too.

ap1kenobi:
Good point.
I'll change this to `resource_tags` and use it to tag a new instance or locate an existing one.

mpdehaan:
@pas256 thanks for review, do you have thoughts on the updates?

@spectrumx can you please add a "version_added: '1.6'" on the new attribute in the DOCUMENTATION section?

pas256:
LGTM

ap1kenobi:
Added "version_added" attribute.

mpdehaan:
merged, thank you!

PR #6086, opened 2014-02-20
Match directory permission with 755 instead of 644

alvin-chang:
Directories need the execution bit being set therefore instead of 644, it needs 755.


mpdehaan:
Merged docfix, thanks!  This will show up the next time docs are pushed.

PR #6090, opened 2014-02-20
Make missing known hosts file message a verbose message

tyll:
Since ansible does not completely implement a OpenSSH option parser it
does not really know which known hosts file needs to be used. Therefore
a faulty message appears if eg a GlobalKnownHosts file is used or a
non-standard UserKnownHosts file as specified by ~/.ssh/config or eg
/etc/ssh/ssh_config. Therefore the message a about the known hosts files
not being found should be only shown if VERBOSITY is at least 1.

### DIFF
--- a/lib/ansible/runner/connection_plugins/ssh.py
+++ b/lib/ansible/runner/connection_plugins/ssh.py
@@ -157,7 +157,8 @@ class Connection(object):
                         return False
 
         if (hfiles_not_found == len(host_file_list)):
-            print "previous known host file not found"
+            if utils.VERBOSITY > 0:
+                print "INFO: No previous known host file found"
         return True
 
     def exec_command(self, cmd, tmp_path, sudo_user=None, sudoable=False, executable='/bin/sh', in_data=None, su_user=None, su=False):


mpdehaan:
I disagree with the above assertion that we should hide the message unless verbosity is set, and think it would be more interesting to get at the root of the underlying problem.

Further, this patch breaks attribution of numerous lines by cleaning up trailing whitespace, so it should not do this.

tyll:
Why does someone need this message? If ansible does not find the host key but the ssh client does, why does the user need to know? If the ssh client does not find the key as well, then ssh will ask the user for confirmation and notify the user about it as well. So there is not much use for normal users unless they want to debug something and then increasing the verbosity is reasonable IMHO.

I removed the whitespace changes, not sure if this really matters if the patch will not be accepted anyhow.

tyll:
Why does someone need this message? If ansible does not find the host key but the ssh client does, why does the user need to know? If the ssh client does not find the key as well, then ssh will ask the user for confirmation and notify the user about it as well. So there is not much use for normal users unless they want to debug something and then increasing the verbosity is reasonable IMHO.

I removed the whitespace changes, not sure if this really matters if the patch will not be accepted anyhow.

PR #6097, opened 2014-02-20
Normalise the module name when comparing against the module names in /proc/modules

mattjeffery:
When using the modprobe command , '_' and '-' in the name of the module are interchangeable. However, in /proc/module they are always listed with '_' in the name. When 'modprobe' Ansible module if the name is given with a '-' the module is enabled regardless of whether it was enabled previously or not, thus marking the task as changed. 

eg. 

``` yaml
- modprobe: name=snd-aloop state=present
```

Will always be marked as changed. The work-around is to use name=snd_aloop.

### DIFF
--- a/library/system/modprobe
+++ b/library/system/modprobe
@@ -60,8 +60,9 @@ def main():
     try:
         modules = open('/proc/modules')
         present = False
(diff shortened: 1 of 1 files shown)


mpdehaan:
merged, thanks!

PR #6104, opened 2014-02-21
Make sure droplet deletion works even when idempotency is achieved using 'name' and 'unique_name' rather than 'id'

gianlucaborello:
So this fixes a bug that makes it impossible to delete a droplet that is specified with just a combination of name and unique_names (for idempotency).

In particular, without the fix, this code: 

```
digital_ocean:
  image_id=1646467
  name=mydroplet
  ssh_key_ids=54036
  size_id=63
  region_id=4
  state=absent
  unique_name=yes
```

will always fail with:

```
TASK: [deploy droplets] *******************************************************
failed: [127.0.0.1] => (item=mydroplet) => {"failed": true, "item": "mydroplet"}
msg: Unable to load id

```

because the module will fail, not finding 'id' (even though, according to the manual and to what I need to do, seems like a completely legit case)

### DIFF
--- a/library/cloud/digital_ocean
+++ b/library/cloud/digital_ocean
@@ -348,7 +348,7 @@ def core(module):
 
         elif state in ('absent', 'deleted'):
             # First, try to find a droplet by id.
-            droplet = Droplet.find(id=getkeyordie('id'))
+            droplet = Droplet.find(module.params['id'])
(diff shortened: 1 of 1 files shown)


mpdehaan:
Merged, thanks!

PR #6105, opened 2014-02-21
Added ansible-vault to the installer

tonk:
The `ansible-vault` command is not installed. Fixed this in `setup.py`.


mpdehaan:
Excellent catch, thanks!

PR #6116, opened 2014-02-22
In the route53 module, documenting that TXT and SPF values must be surrounded by quotes

spatten:
Otherwise you get a cryptic error message from Route 53.


mpdehaan:
Merged, thanks!  Will appear in next docs push.

PR #6118, opened 2014-02-22
Expand templates inside of notify

tdfischer:
This patch expands templates inside of the notify block, to allow for implementing 'hooks' in roles.

Example:
roles/hooked/main.yml:

```

---
- debug: Frobbing "{{thing}}"
  notify: $onUpdate
```

main.yml:

```

---
hosts: all
roles:
  - role: hooked
    thing: "Something"
    onUpdate:
      - reload
  - role: hooked
    thing: "Something Else"
    onUpdate:
      - reload
      - cleanup
handlers:
  - name: reload
    action: debug Reloaded
  - name: cleanup
    action: debug Cleaned up!
```

This results in the reload handler being executed once for each task, and the cleanup handler executing once for the second application of the hooked role.

### DIFF
--- a/lib/ansible/playbook/__init__.py
+++ b/lib/ansible/playbook/__init__.py
@@ -451,21 +451,27 @@ class PlayBook(object):
 
     # *****************************************************
 
-    def _flag_handler(self, play, handler_name, host):
+    def _flag_handler(self, play, handlers_or_handler, host):
         '''
         if a task has any notify elements, flag handlers for run
         at end of execution cycle for hosts that have indicated
         changes have been made
         '''
 
-        found = False
-        for x in play.handlers():
-            if handler_name == template(play.basedir, x.name, x.module_vars):
-                found = True
-                self.callbacks.on_notify(host, x.name)
-                x.notified_by.append(host)
-        if not found:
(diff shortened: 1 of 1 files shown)


mpdehaan:
Code details of this pull request aside (to be reviewed and considered later), this needs to be updated to use four space indent.

One thing to address is that the request variable should likely be templated earlier.  The way you have it now it will call template excessively through the loop process, and I think it can be called when the handler is first loaded.

(FYI, public service announcement -- You're also using outdated variable syntax above, which will generate warnings and most definitely not work in 1.6, and has issued warnings since 1.4.  If you have any playbooks doing this, time to update them!)

tdfischer:
Thanks for your feedback.

Any immediate suggestion where I could start looking to template the variable earlier?

mpdehaan:
Sorry for the late response, it's track all the stream of github comments.

Basically when it gets parsed by task.py in the playbook/ code.

amenonsen:
I think this PR should be closed now. The feature was all right, but the code in devel has changed significantly enough to require a rewrite, and the original revision request has gone unanswered for over a year.

bcoca:
this is already working in current devel, so I'm closing the ticket.

PR #6120, opened 2014-02-22
Don't require pycurl in apt_repository when it's not actually needed

lalinsky:
The `python-pycurl` package is only required for adding PPA repositories, which I expect is not even the main use case of `apt_repository`. I think it makes sense to only require if it's actually needed.

### DIFF
--- a/library/packaging/apt_repository
+++ b/library/packaging/apt_repository
@@ -28,9 +28,10 @@ short_description: Add and remove APT repositores
 description:
     - Add or remove an APT repositories in Ubuntu and Debian.
 notes:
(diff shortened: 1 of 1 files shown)


mpdehaan:
This seems reasonable.

I'm also wondering if the usage of pycurl can be reduced down to using urllib to remove even that requirement though.

lalinsky:
As far as I know, `urllib2` still can't do SSL certificate validations, so that would involve some custom code to see if the certificate matches the domain.

mpdehaan:
merged, thanks!

PR #6122, opened 2014-02-22
Add env variables as default in the ConfigParser

rach:
This allow to do substitution base on environment  variables. The change is so tiny that I did the PR directly but you may have some reason to refuse of exposing the env var in the ansible.cfg and I will be happy to discuss them on the mailing list.
#### Some Background about this change:

Our team is currently sharing generic roles between project via git submodule.
Loads of us are not familiar with git submodules so we've been talking about changing and keeping our role in a specific path to look at. But this create a problem as we are loosing the link between roles version and project version. 
I recently thought that I could simply make our roles installable via setuptools and then having a requirement file specific for our generic roles version. This will allow use to install our generic roles via `pip`.
But then I would like to able to set the role_path to $VIRTUAL_ENV/lib/python2.7/site-packages/ansible-roles/
so for being able to do so I would need access to the env variables from ansible.cfg 

Now my change is maybe redundant and you may have some better suggestions to handle role versioning other than `git submodule` or `subtree` .

R.

### DIFF
--- a/lib/ansible/constants.py
+++ b/lib/ansible/constants.py
@@ -58,7 +58,7 @@ def _get_config(p, section, key, env_var, default):
 def load_config_file():
     ''' Load Config File order(first found is used): ENV, CWD, HOME, /etc/ansible '''
 
-    p = ConfigParser.ConfigParser()
+    p = ConfigParser.ConfigParser(os.environ)
 
     path0 = os.getenv("ANSIBLE_CONFIG", None)
     if path0 is not None:


mpdehaan:
Typically we have been doing expandenv type things on parameters instead.  Need to investigate how this would be different.

rach:
Thanks, Let me know if you need anymore details.
I would have much prefer a being able to use substitution with dotted name but from what I aware configparser doesn't surport it. So something like below won't be possible

```
 [default]
 roles_path = %(env_vars.virtual_env)s/roles
```

but only direct substitution is supported:

```
 [default]
 roles_path = %(virtual_env)s/roles
```

mpdehaan:
Sorry for the late response, we have a large queue here.

I think we need to continue to proceed with expanding os.path.expandvars explicitly in cases where we want to expand the path.

I'm not following the above python formatting magic, and as we are not catering to an exclusively python developing audience I don't think many folks would grok what that variables they could use in format strings there.

rach:
Thanks for your answer and your time.
If we use `os.path.expandvars` , that's perfect and much cleaner.
To be honest I was not aware of this method 5min ago.

PR #6124, opened 2014-02-22
Add packaging module composer

dmtrs:
[Composer](https://getcomposer.org/) is a tool for dependency management in PHP. It allows you to declare the dependent libraries your project needs and it will install them in your project for you.

### DIFF
--- a/library/packaging/composer
+++ b/library/packaging/composer
@@ -1,0 +1,153 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2014, Dimitrios Tydeas Mengidis <tydeas.dr@gmail.com>
+
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+DOCUMENTATION = '''
+---
+module: composer
+author: Dimitrios Tydeas Mengidis
+short_description: Dependency Manager for PHP
+version_added: "1.4"
+description:
+    - Composer is a tool for dependency management in PHP. It allows you to declare the dependent libraries your project needs and it will install them in your project for you
+options:
+    working_dir:
+        description:
+            - Directory of your project ( see --working-dir )
+        required: true
+        default: null
+        aliases: [ "working-dir" ]
+    prefer_source:
+        description:
+            - Forces installation from package sources when possible ( see --prefer-source )
+        required: false
+        default: "no"
+        choices: [ "yes", "no" ]
+        aliases: [ "prefer-source" ]
+    prefer_dist:
+        description:
+            - Forces installation from package dist even for de versions ( see --prefer-dist )
+        required: false
+        default: "no"
+        choices: [ "yes", "no" ]
+        aliases: [ "prefer-dist" ]
+    no_dev:
+        description:
+            - Disables installation of require-dev packages ( see --no-dev )
+        required: false
+        default: "yes"
+        choices: [ "yes", "no" ]
+        aliases: [ "no-dev" ]
+    no_scripts:
+        description:
+            - Skips the execution of all scripts defined in composer.json ( see --no-scripts )
+        required: false
+        default: "no"
+        choices: [ "yes", "no" ]
+        aliases: [ "no-scripts" ]
+    no_plugins:
+        description:
+            - Disables all plugins ( see --no-plugins )
+        required: false
+        default: "no"
+        choices: [ "yes", "no" ]
+        aliases: [ "no-plugins" ]
+    optimize_autoloader:
+        description:
+            - Optimize autoloader during autoloader dump ( see --optimize-autoloader ). Convert PSR-0/4 autoloading to classmap to get a faster autoloader. This is recommended especially for production, but can take a bit of time to run so it is currently not done by default.
+        required: false
+        default: "yes"
+        choices: [ "yes", "no" ]
+        aliases: [ "optimize-autoloader" ]
+requirements:
+    - php
+    - composer installed in bin path (recommended /usr/local/bin)
+notes:
+    - Default options that are always appended in each execution are --no-ansi, --no-progress, and --no-interaction
+'''
+
+EXAMPLES = '''
+# Downloads and installs all the libs and dependencies outlined in the /path/to/project/composer.lock
+- composer: command=install working_dir=/path/to/project
+'''
+
+import os
+import re
+
+def parse_out(string):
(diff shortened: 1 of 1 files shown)


stof:
I would not allow running `composer update` here. You should _never_ run `update` during your deployment, as it means you will deploy code that you never tested before that (as you will get newer versions of your dependencies). A deployment should always run `composer install` to install the locked dependencies (this also implies that your `composer.lock` should be versionned)

library/packaging/composer
```diff
@@ -0,0 +1,208 @@
...
+    pckgs = " ".join(packages) if len(packages) > 0 else ""
+
+    before_lock = read_lock(module)
+    rc, out, err = composer_exec(module, command, "%s %s" % (opts, pckgs))
+
+    if rc != 0:
+        module.fail_json(msg=parse_out(err))
+    else:
+        after_lock = read_lock(module)
+        module.exit_json(changed=(before_lock!=after_lock), before_lock=before_lock, after_lock=after_lock, msg=parse_out(out))
```
stof: an update can change the installed dependencies without having a different hash for the `composer.json` (you can get newer deps matching the same constraints)

dmtrs:
@stof (on update) That's true, I will remove the update functionality.

(on hash) Was not aware of this and since I will remove the update functionality there won't be an issue.

What is really tricky, is that it's not easy/safe to parse the output of composer to determine if something changed.

For ansible modules there is also the [Check Mode](http://docs.ansible.com/developing_modules.html#check-mode) that can be used along with the `composer --dry-run`, so considering hat the dry-run will write to lock file its not easy to determine if there will be change.

Could be a work around?

stof:
@dmtrs `composer install` will never write anything into the lock file. It reads it to determine what should be installed (and this is precisely why a deployment should use `install` rather than `update`)

dmtrs:
@stof Thanks for your feedback. I will alter the module code and update the pr, so you will get noticed.

dmtrs:
@stof Removed the update and read from lock functionality. Let me know if there is anything else

library/packaging/composer
```diff
@@ -0,0 +1,154 @@
...
+            - Forces installation from package dist even for de versions ( see --prefer-dist )
+        required: false
+        default: "no"
+        choices: [ "yes", "no" ]
+        aliases: [ "prefer-dist" ]
+    no_dev:
+        description:
+            - Disables installation of require-dev packages ( see --no-dev )
+        required: false
+        default: "no"
```
stof: for a prod deployment, the sensible default is to use `--no-dev`.

However, I'm not a user of Ansible (I went here because you posted about the support of composer in Ansible, and I went giving my feedback as a core contributor to Composer), so it might depend on the way Ansible considers its main use case

library/packaging/composer
```diff
@@ -0,0 +1,154 @@
...
+            - Disables all plugins ( see --no-plugins )
+        required: false
+        default: "no"
+        choices: [ "yes", "no" ]
+        aliases: [ "no-plugins" ]
+    optimize_autoloader:
+        description:
+            - Optimize autoloader during autoloader dump ( see --optimize-autoloader ). Convert PSR-0/4 autoloading to classmap to get a faster autoloader. This is recommended especially for production, but can take a bit of time to run so it is currently not done by default.
+        required: false
+        default: "no"
```
stof: same than for `--no-dev`. For a prod deployment, it makes sense to turn it on by default

library/packaging/composer
```diff
@@ -0,0 +1,154 @@
...
+        aliases: [ "optimize-autoloader" ]
+requirements:
+    - php
+    - composer installed in bin path (recommended /usr/local/bin)
+notes:
+    - Default options that are always appended in each execution are --no-ansi, --no-progress, and --no-interaction
+'''
+
+EXAMPLES = '''
+# Downloads and installs all the libs and dependencies outlined in the /path/to/project/composer.json
```
stof: I would rather say `composer.lock` than `composer.json`. If you change your composer.json without updating the lock file, the locked dependencies will be installed here, not the new requirements (as you haven't run `composer update` to update the locked versions)

library/packaging/composer
```diff
@@ -0,0 +1,154 @@
...
+
+    # Prepare options
+    for i in module.params:
+        opt = "--%s" % i.replace("_","-")
+        p = module.params[i]
+        if isinstance(p, (bool)) and p:
+            options.add(opt)
+        elif isinstance(p, (str)):
+            print opt, p
+            options.add("%s=%s" % (opt, p))
```
stof: I'm not experienced with Python, but shoudln't you escape the values here to avoid issues with shell special chars ?

stof:
@dmtrs For what concerns Ansible, I cannot give you any review (you are more experienced than me on it, as my experience with Ansible does not exist). You will need to wait for a review of the Ansible maintainers

dmtrs:
@stof Your feedback the kind of feedback I was asking for. Thanks for your time

mpdehaan:
This looks pretty reasonable, I'm going to merge this and update the version added field to say "1.6".

mpdehaan:
This is now merged in, thank you!

rdohms:
A bit late to the party, but i have a few concerns on this and how it would be used in an ansible setting.

I see 2 opportunities for using composer in a provision setting:
1. Building a CI/Build server, where tools like PHPUnit should be globally available
2. Running composer install upon project deploy

This seems to only address scenario 2, which is ok, but is also something most people try to avoid.

Scenario 1 would be the most used scenario, but this lacks support for composer commands like the `global` command that would allow this to provision that scenario.

I think it would be more productive to have a complete mapping of composer that is more similar to `gem` support with `state` and such, allowing you to manage packages with composer, not just use it to install project dependencies.

Should we look into adding this? it may represent BC breaks for future releases.

dmtrs:
@rdohms For the first case scenario (CI/Build server), the most common scenario (imho) is that even phpunit is declared in developers dependencies and this for reasons like:
- Individual projects may rely on diff versions of phpunit package
- This way the project come more "packaged" in terms that anyone can get and test locally without setting up anything globally to the machine.

Can you elaborate with more details on your scenario 1 case?

rdohms:
@dmtrs let me point out a few points
- CI is about more then just PHPunit, its about PHPCS, PDepend, PHPLoc, PHPDoc, etc. These are usually only executed at CI time for stats, so people usually don't add these into the dev dependencies because they are not required in dev environments. (apart from testing and code style stuff).
- This would be useful in provisioning your CI server, or a build server especially, in dev environments composer install would indeed be the frequent case.

As someone noted on Twitter, the module as-is only replaces:
`command: composer install -a -b` with `composer: command=install a=x b=x`
Allowing it to be used to manage packages on a gem-like level (more common then it looks) would give it more purpose and effectively replace PEAR for global stuff, which for some people is still the goto.

At least those are the scenarios i can pick out from my ansible playbooks now.

geerlingguy:
In agreement with rdohms; global install is useful for things like Jenkins and SonarQube. In those cases, composer can basically function like PEAR would. I would like to see the functionality expanded a little, as outlined above.

dmtrs:
@geerlingguy , @rdohms That's fine. Could you share some examples in the following format, so we all are in the same page?

```
# Downloads and installs all the libs and dependencies outlined in the /path/to/project/composer.lock
- composer: command=install working_dir=/path/to/project
```

rdohms:
Here is what i can think of:

```
# Install PHPCS at version * globally (composer global require 'squizlabs/php_codesniffer=*')
- composer: package=squizlabs/php_codesniffer version=* global=true state=present
```

I'll see if i can look at composer and find other actions to extract.

PR #6125, opened 2014-02-23
Modules for managing Rackspace Cloud Databases instances, databases and users.

jails:
Theses modules allow to manage Rackspace Cloud Databases instances , databases and users.
### rax_cdb

Supported operations are:
- creation (when state == present)
- deletion (when state == absent)
- resizing (when flavor is modified)
- resizing the db volume (when the volume is increased, i.e. increase only)
### rax_cdb_database

Supported operations are:
- creation (when state == present)
- deletion (when state == absent)
### rax_cdb_user

Supported operations are:
- creation (when state == present)
- deletion (when state == absent)
- grant/revoke privileges (according the the databases option)
- always update password according the db_password option (this action doesn't generate a "change event" since the previous state can't be known).
- Note: do not use an host option != '%'. It make the user "invisible" from the API (maybe a bug in pyrax)

Example of configuration

``` yaml
  tasks:
  - name: build webserver
    local_action:
      module: rax_cdb
      credentials: ../.rax.conf
      region: IAD
      name: db_server
      flavor: 1
      volume: 1
      state: present
      wait: yes
    register: rax_db_server

  - name: build database
    local_action:
      module: rax_cdb_database
      credentials: ../.rax.conf
      region: IAD
      cdb_id: "{{ rax_db_server.cdb.id }}"
      name: db
      state: present
    register: rax_db_database

  - name: build database user
    local_action:
      module: rax_cdb_user
      credentials: ../.rax.conf
      region: IAD
      cdb_id: "{{ rax_db_server.cdb.id }}"
      db_username: user
      db_password: user
      databases: ['db']
      state: present
    register: rax_db_user
```

### DIFF
--- a/library/cloud/rax_cdb
+++ b/library/cloud/rax_cdb
@@ -1,0 +1,293 @@
+#!/usr/bin/python -tt
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: rax_cdb
+short_description: create / delete or resize a Rackspace Cloud Databases instance
+description:
+  - creates / deletes or resize a Rackspace Cloud Databases instance
+    and optionally waits for it to be 'running'. The name option needs to be unique since
+    it's used to identify the instance.
+version_added: "1.6"
+options:
+  api_key:
+    description:
+      - Rackspace API key (overrides I(credentials))
+    aliases:
+      - password
+  credentials:
+    description:
+      - File to find the Rackspace credentials in (ignored if I(api_key) and
+        I(username) are provided)
+    default: null
+    aliases:
+      - creds_file
+  region:
+    description:
+      - Region to create an instance in
+    default: DFW
+  username:
+    description:
+      - Rackspace username (overrides I(credentials))
+  name:
+    description:
+      - Name of the databases server instance
+    default: null
+  flavor:
+    description:
+      - flavor to use for the instance 1 to 6 (i.e. 512MB to 16GB)
+    default: 1
+  volume:
+    description:
+      - Volume size of the database 1-150GB
+    default: 2
+  state:
+    description:
+      - Indicate desired state of the resource
+    choices: ['present', 'absent']
+    default: present
+  wait:
+    description:
+      - wait for the instance to be in state 'running' before returning
+    default: "no"
+    choices: [ "yes", "no" ]
+  wait_timeout:
+    description:
+      - how long before wait gives up, in seconds
+    default: 300
+requirements: [ "pyrax" ]
+author: Simon JAILLET
+notes:
+  - The following environment variables can be used, C(RAX_USERNAME),
+    C(RAX_API_KEY), C(RAX_CREDS_FILE), C(RAX_CREDENTIALS), C(RAX_REGION).
+  - C(RAX_CREDENTIALS) and C(RAX_CREDS_FILE) points to a credentials file
+    appropriate for pyrax. See U(https://github.com/rackspace/pyrax/blob/master/docs/getting_started.md#authenticating)
+  - C(RAX_USERNAME) and C(RAX_API_KEY) obviate the use of a credentials file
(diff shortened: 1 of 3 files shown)


sivel:
I'll try to look over this on Monday, and provide feedback.

Thanks for this!

mpdehaan:
Thanks for the PR @jails, I'll probably be deferring to @sivel for review things.

@sivel off topic but I'm thinking we _do_ want to start finding a better way to load in common args documentation, I'll see if I can bump the mailing list thread.   This won't hold back the above of course.

jails:
Thanks guys. Just update the PR to provide a full support of the Rackspace Cloud Databases with databases & users management. Hope you'll get a chance to review it.

sivel:
I haven't had an opportunity to look at this yet, but it is still on my to-do list.  I haven't forgotten you.  It may be several days before I can get back to this.

jails:
No worries, feel free to review it when you get a chance. ;-)

mpdehaan:
@sivel -- if you get a chance to take a look please let us know.  Thanks!

sivel:
### all
1. The cloud databases implementation is rackspace specific, so we do not need the following module attributes documented:
   - `auth_endpoint`
   - `env`
   - `identity_type`
   - `tenant_id`
   - `tenant_name`
   - `verify_ssl`
2. `version_added` should be bumped to `1.6`
3. I know it isn't really a requirement for ansible, but we are trying to maintain pep8 in the rax modules
   1. You have a number of lines that end with colons
   2. Other than the DOCUMENTATION section we try to limit to 79 char line length
   3. 2 blank lines between 
4. `rax_slugify` should only be used in compute plugins where we don't want to stomp on other variables derermined during setup.  Maybe look at one of the other rax plugins such as `rax_dns` for a non compute `to_dict` function

#### Coding standards type stuff
1. Don't do `**` expansion on `{}` invocations of dictionaries.  Create a dict before or use `dict()` instead.  Using `dict()` is actually preferred over `{}`. See https://github.com/ansible/ansible/blob/devel/CODING_GUIDELINES.md#misc-preferences
2. Try to use '%s' printf like string formatting instead of `+` string concatentation.
3. Avoid using generic function names like `find()`.  I am worried that in the future if rax.py or basic.py are extended we could run into naming conflicts.
4. I recommend with `module.fail_json` to use something like `module.fail_json(msg='%s' % e)` or `module.fail_json(msg='%s' % e.message)`. We have run into issues where e.message was a value other than a string and it caused an exception to be raised.

### rax_cdb
1. Use 'GB' instead of 'Go' or 'GO'. Most people are likely to not understand this translation
2. Under flavor you reference `512ko to 16Go`, this should be `512MB to 16GB`
3. in `delete` the `for arg, value in dict(name=name).iteritems():` feels over complicated for the need, should probably just remvoe that, and just check `if not name:`

jails:
Hi @sivel, thanks for your comments. Pushed some updates for this PR. Btw I didn't squash it yet so you can keep focus on changes only.

Ps:
Since it's my first Pyton code, don't hesistate to ping me if I missunderstood something.

sivel:
I think the changes look good!  One more small thing.  Can you rename your functions such as `saveDatabase` to use lower case with underscores such as `save_database`?

jails:
Yeah sorry I mixed conventions, updated, squashed & rebased.

sivel:
Great. I'll do another quick look, some testing and update in the next few days.

sivel:
Sorry for the delay.  I've had some time to play around with these and I have 1 more small request and then I think we are good to go.  `rax_db_user` should set `no_log=True` to for `db_password`:

``` diff
diff --git a/library/cloud/rax_cdb_user b/library/cloud/rax_cdb_user
index 54942be..e12a64e 100644
--- a/library/cloud/rax_cdb_user
+++ b/library/cloud/rax_cdb_user
@@ -224,7 +224,7 @@ def main():
         dict(
             cdb_id=dict(type='str', required=True),
             db_username=dict(type='str', required=True),
-            db_password=dict(type='str', required=True),
+            db_password=dict(type='str', required=True, no_log=True),
             databases=dict(type='list', default=[]),
             host=dict(type='str', default='%'),
             state=dict(default='present', choices=['present', 'absent'])
```

jails:
Thanks for your comments, just updated the PR.

sivel:
Great, I think this is good to merge.

As a note to myself and the folks from Ansible, once this is merged I'll submit another pull request for the shared module docs stuff.

jails:
You need something more @mpdehaan ?

jails:
Anything I can do to make it real ?

jails:
Would you mind keeping me updated on this PR @mpdehaan ? It's fine if you don't want to merge it, just keep me updated (or close this PR) so I can work on my own. Thanks.

mpdehaan:
Busy project, we don't read all github notifications except when we're back to looking at what to next triage in the queue.

Taking a look now.

mpdehaan:
Merged, thank you!

mpdehaan:
Sorry about the delay on this one, our prior "queue attack metrics" were prioritizing some other things.  Glad to see these in and they will be a part of 1.8 and on the docs site very shortly.

jails:
Awesome, thanks !

PR #6126, opened 2014-02-23
allow instance or interface targets for routes in ec2_vpc

deoxxa:


pas256:
Please add documentation for this.

mpdehaan:
I'm going to close this one as it is missing an EXAMPLES section that shows how this is used.  If you would like to update this with usage, we will look into merging it, thanks!

PR #6127, opened 2014-02-23
sort returned subnets as specified in config in ec2_vpc

deoxxa:


mpdehaan:
Can you provide some information about the results of this one?

I'm not entirely following the current_subnets logic and was expecting to maybe see a "sort" in there.

Thanks!

mpdehaan:
I'm going to close this one due to lack of info.
If you would like to provide more information in this pull request and resubmit that would be welcome.

Thanks!

PR #6138, opened 2014-02-23
Off by one error

do3cc:
When len(probline) == column then probline[column] is outside of the string

### DIFF
--- a/lib/ansible/utils/__init__.py
+++ b/lib/ansible/utils/__init__.py
@@ -374,7 +374,7 @@ It should be written as:
 """
         return msg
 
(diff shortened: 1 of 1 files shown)


mpdehaan:
Can you share how to reproduce the problem for testing purposes?

Thanks!

do3cc:
yes.
I'd love to provide a test but I would only be able to do this in a week or so. Here is an actual minimal test case on the console:

``` shell
echo "- test: *" > test.yml
ansible-playbook test.yml
```

mpdehaan:
Yeah I just meant a reproducing playbook or CLI call, that's great.  Thanks!

risaacson:
Closed with #6199 which had a duplicate effort for the commit.

PR #6139, opened 2014-02-23
Assemble decode delimiter strings

jiv-e:
Without this it seems impossible to add new lines to prefix, suffix or delimiter.

### DIFF
--- a/library/files/assemble
+++ b/library/files/assemble
@@ -57,6 +57,18 @@ options:
     required: false
     choices: [ "yes", "no" ]
     default: "no"
+  prefix:
+    description:
+      - A prefix prepended at the start of the assembled file.
+    version_added: "1.5"
+    required: false
+    default: null
+  suffix:
+    description:
+      - A suffix appended at the end of the assembled file.
+    version_added: "1.5"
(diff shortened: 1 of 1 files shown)


jimi-c:
I'm going to close this for a couple of reasons. Primarily, this is something we've actually noticed for some other parameters and needs to probably be fixed at a higher level, rather than doing it in each module.

I've been working on a couple of patches for assemble already that will be merged soon, one of which is detecting if the delimiter ends with a newline character - if not it will insert one. If/when we merge in the suffix/prefix I can add the same checks for those.

Let us know if you have any questions regarding this.

jiv-e:
It is definitely better to fix this on the higher level. I tried to find if this was caused by some encode function, but I couldn't find any. Maybe strings from YAML files are encoded by default. This also works: string.replace("\\n", "\n"). But tabs etc. are left out of course.

In my opinion it's not a good idea to add a new line by default. I can imagine some use cases where the new line is not wanted behavior.

bcoca:
you might want to move some of the common code to module.utils, to avoid
x2 work in the future.

PR #6143, opened 2014-02-24
Files exists

trbs:
Add exists command, simple command to test existence or files, directories, links, etc on filesystem.

While you can do the same thing with the `stats` module, this is far cleaner and more elegant for simple cases where one wants to check existence of some path on the filesystem


mpdehaan:
Thanks for this, though the exists command is not needed as you can use the stat module.   Since one of our goals is to be a somewhat minimal application, we don't want to duplicate code that is also in the stat module.
- stat: path=/some/path
  register: stat
- shell: echo hi
  when: stat.stat.exists

I see you have added that you view this to be more cleaner and elegant, etc, and that's fine, though we're going to have a difference of opinion here about code duplication.

Also in the future, please submit pull requests without merge commits, as we cannot include them.  

It also looks like you have an unrelated commit in there in the form of b7df767

Thanks again!

PR #6144, opened 2014-02-24
Update playbooks_lookups.rst add example with using encrypt

lied:
Add a sample for using password lookup with a system user account and option encrypt

### DIFF
--- a/docsite/rst/playbooks_lookups.rst
+++ b/docsite/rst/playbooks_lookups.rst
@@ -58,6 +58,11 @@ Starting in version 1.4, password accepts a "chars" parameter to allow defining 
     - hosts: all
 
       tasks:
+        # create a user account with encryption sha256 and read the password from the file password if it exists
+        # if password file does not exist, create a random password and store it in this file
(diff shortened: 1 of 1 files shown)


mpdehaan:
Thanks for this!

(Note to self - probably want to move this down a bit and tweak the language some after merging)

mpdehaan:
Somewhat concerned we should just stress usage of vault here.

This will generate a SHA and use the SHA everywhere, but doesn't really get the original back out.   Maybe something we should deprecate (having shoud not have merged originally)

lied:
For me and my team it is working quite good :-)

amenonsen:
I think this PR should just be closed now. It's a trivial five-line patch, but really all it adds to the other existing examples of lookup('password', ...) is the encrypt= parameter. It's not worth keeping around for a year or more, especially given the concerns @mpdehaan expressed (which I agree with, FWIW).

gregdek:
Agreed, we'll pass on this patch. Thanks @lied for the submission, and sorry it's taken so long for a decision here.

PR #6150, opened 2014-02-24
Fix incorrect use of .join on list causing a traceback

hkariti:
tmpcond is a list, so .join won't work.

While investigating this behavior I also noticed that a role is evaluated several times (and so its `when' condition added to tmpcond several times) during parsing - is this ok?

### DIFF
--- a/lib/ansible/playbook/play.py
+++ b/lib/ansible/playbook/play.py
@@ -275,7 +275,7 @@ class Play(object):
                                     if type(passed_vars['when']) is str:
                                         tmpcond.append(passed_vars['when'])
                                     elif type(passed_vars['when']) is list:
-                                        tmpcond.join(passed_vars['when'])
+                                        tmpcond += passed_vars['when']
 
                                     if type(dep_vars['when']) is str:
                                         tmpcond.append(dep_vars['when'])


mpdehaan:
Can you share an example playbook snippet that reproduces this problem?  Curious how things might have been working before or we've got a test we need to add.

Thanks!!!

hkariti:
Not really a playbook, but roles. If you have the following dependency chain, each with a condition:
role1 -`when '2'`-> role2 -`when '3'`-> role3 -`when '4'`-> role4
Problems occur after the 3rd role.

Adding some print statements in `_build_role_dependencies` produces the following output:

```
Role: {'role': 'role2', 'when': '2'}
Passed: {'when': '2'} dep: {'when': '3'}
Role: {'role': 'role3', 'when': '3'}
Passed: {'when': ['2', '3']} dep: {'when': '4'}
  File "/usr/bin/ansible-playbook", line 269, in <module>
    sys.exit(main(sys.argv[1:]))
  File "/usr/bin/ansible-playbook", line 209, in main
    pb.run()
  File "/usr/lib/pymodules/python2.7/ansible/playbook/__init__.py", line 229, in run
    play = Play(self, play_ds, play_basedir)
  File "/usr/lib/pymodules/python2.7/ansible/playbook/play.py", line 83, in __init__
    ds = self._load_roles(self.roles, ds)
  File "/usr/lib/pymodules/python2.7/ansible/playbook/play.py", line 329, in _load_roles
    roles = self._build_role_dependencies(roles, [], self.vars)
  File "/usr/lib/pymodules/python2.7/ansible/playbook/play.py", line 278, in _build_role_dependencies
    self._build_role_dependencies([dep], dep_stack, passed_vars=dep_vars, level=level+1)
  File "/usr/lib/pymodules/python2.7/ansible/playbook/play.py", line 278, in _build_role_dependencies
    self._build_role_dependencies([dep], dep_stack, passed_vars=dep_vars, level=level+1)
  File "/usr/lib/pymodules/python2.7/ansible/playbook/play.py", line 266, in _build_role_dependencies
    tmpcond.join(passed_vars['when'])
AttributeError: 'list' object has no attribute 'join'
```

jimi-c:
Merged, thanks!

PR #6151, opened 2014-02-24
Vault rewrite, pass 1

jctanner:
Refactor vault into VaultLib, VaultEditor, and VaultAES classes.

### DIFF
--- a/bin/ansible-vault
+++ b/bin/ansible-vault
@@ -20,13 +20,13 @@
 # example playbook to bootstrap this script in the examples/ dir which
 # installs ansible and sets it up to run on cron.
 
+import os
 import sys
 import traceback
 
 from ansible import utils
 from ansible import errors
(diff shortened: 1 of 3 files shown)


mpdehaan:
Checking this out on my VM shortly... thanks!

mpdehaan:
This is a big upgrade to the architecture, as mentioned, ideally we'd want some test/units/TestVault.py for VaultLib's class and we're solid.

mpdehaan:
Here's a small buglet:

[root@localhost ansible]# ansible-vault edit /tmp/test.yml  # valid filename
Vault password: invalid_password
ERROR: global name 'filename' is not defined

edit: might have been the same

mpdehaan:
Possible usability tweak:  Technically the "is already encrypted / is not encrypted" checks could happen before asking for the passwords.

PR #6154, opened 2014-02-24
Add support for shared module documentation fragments

sivel:
See https://groups.google.com/forum/#!topic/ansible-devel/B-Md0nYB0s4

This pull request adds support for shared module documentation fragments.

Modules can share documentation, by extending a fragment located in `ansible/utils/module_docs_fragments.py`

The code that handles the merging, is done in `ansible/utils/module_docs.py`

The `notes` section is appended to.  All other sections are looped over and either set if they don't exist in the fragment, or updated.

I've added the fragments for the `rax` modules thus far.  Since I was touching the DOCUMENTATION strings for all `rax` modules, I also made some cleanup there as well.

Sample DOCUMENTATION from a module:

``` yaml

---
module: rax_queue
short_description: create / delete a queue in Rackspace Public Cloud
description:
     - creates / deletes a Rackspace Public Cloud queue.
version_added: "1.5"
options:
  name:
    description:
      - Name to give the queue
    default: null
  state:
    description:
      - Indicate desired state of the resource
    choices:
      - present
      - absent
    default: present
author: Christopher H. Laco, Matt Martz
extends_documentation_fragment: RACKSPACE
```

The `extends_documentation_fragment` key references a variable from `ansible/utils/module_docs_fragments.py`.  The provided value is always uppercased.

### DIFF
--- a/lib/ansible/utils/module_docs.py
+++ b/lib/ansible/utils/module_docs.py
@@ -23,8 +23,7 @@ import ast
 import yaml
 import traceback
 
-from ansible.utils import module_docs_fragments as fragments
-
+from ansible import utils
 
 # modules that are ok that they do not have documentation strings
 BLACKLIST_MODULES = [
@@ -37,6 +36,10 @@ def get_docstring(filename, verbose=False):
     in the given file.
     Parse DOCUMENTATION from YAML and return the YAML doc or None
     together with EXAMPLES, as plain text.
+
+    DOCUMENTATION can be extended using documentation fragments
+    loaded by the PluginLoader from the module_docs_fragments
+    directory.
     """
 
     doc = None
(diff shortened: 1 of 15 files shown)


mpdehaan:
Thanks very much!

This looks pretty solid and it's pretty understandable.   My first thought would be to make the fragments directory work on a plugin system though, so we could keep the fragments in different files.  Can you tweak this?   utils/docs_fragments/rackspace.py (etc) ?

I also believe you had mentioned the fragments taking a list, is this something in the above or you think we might need?

sivel:
I'll give that a go.

As far as the list stuff, I basically just allowed it to be supplied via different fragments, so I could have a RACKSPACE and RACKSPACE_AND_OPENSTACK fragment.

I'll have to tweak this a bit.

When you mention plugins, are you wanting to use PluginLoader?  Or something a little more simplistic?

mpdehaan:
I was thinking the existing loader, yeah.  It probably doesn't need to be configurable throughout the whole stack yet as I doubt we want to expose this to userland (it would be confusing and usually not needed).

We may include a comment or something in the rax modules that additional common documentation lives in <FILENAME> and is mixed in dynamically, IDK?

sivel:
Ok, I'll play around and see what I can come up with.  Likely no more updates for today.

sivel:
The pull request has been updated to use the PluginLoader.  A few comments:
1. The newly added `fragment_loader`, currently looks in a python package path, and a file system path.  The path ends up being the same, which prevents configurability currently.  We can eventually move the path to a constant in the future if we want this to be configurable down the road.
2. The class for fragments is ModuleDocFragment
3. The files live in `ansible/utils/module_docs_fragments/`
4. the `extends_documentation_fragment` key, is now lowercase, and refers to the .py file located in the above path
5. `extends_documentation_fragment` can use dot notation, to specify a fragment variable within the class to use.  The default is 'documentation'.  An example would be just specifying `rackspace` for the standard docs, and `rackspace.openstack` to use the docs that allow use with other OpenStack clouds.  So basically this gives us the ability for multiple doc fragments, per "provider"
6. The `rax` modules updated in this pull also add reference that the DOCUMENTATION is a specific doc stub for the module, extended by this framework.

I believe everything else remains the same.

mpdehaan:
Merging this now, for real!

sivel:
I haven't looked at the merge status of this in a while.  Let me know if you need any help or whatnot with merging.

mpdehaan:
This is now merged in, and conflicts have been resolved.

I've additionally used this for the file/copy/template shared args.

You may wish to audit things for potentially shared parameters added since, like region, as I think region was the only conflict.

Thanks!

PR #6156, opened 2014-02-24
Add support for checking host against global known host files.

iiordanov:
Hi Michael,

I've added support for the standard SSH known hosts files as well as the user known hosts file. Please note there are two possible locations for the global known_hosts file and I check both.

Also, please note I've added a check for existence into the new loop which iterates over the list of hosts files, and if none of the files were found (i.e. the number of files not found is equal to the length of the list), I output the error that you used to output before ("previous known host file not found"). Therefore the original check for existence became obsolete and I've removed it.

Thanks!
iordan

### DIFF
--- a/lib/ansible/runner/connection_plugins/ssh.py
+++ b/lib/ansible/runner/connection_plugins/ssh.py
@@ -118,35 +118,46 @@ class Connection(object):
 
     def not_in_host_file(self, host):
         if 'USER' in os.environ:
-            host_file = os.path.expandvars("~${USER}/.ssh/known_hosts")
+            user_host_file = os.path.expandvars("~${USER}/.ssh/known_hosts")
         else:
-            host_file = "~/.ssh/known_hosts"
-        host_file = os.path.expanduser(host_file)
-        if not os.path.exists(host_file):
-            print "previous known host file not found"
-            return True
-        host_fh = open(host_file)
-        data = host_fh.read()
-        host_fh.close()
-        for line in data.split("\n"):
-            if line is None or line.find(" ") == -1:
+            user_host_file = "~/.ssh/known_hosts"
+        user_host_file = os.path.expanduser(user_host_file)
+        
+        host_file_list = []
+        host_file_list.append(user_host_file)
+        host_file_list.append("/etc/ssh/ssh_known_hosts")
+        host_file_list.append("/etc/ssh/ssh_known_hosts2")
+        
+        hfiles_not_found = 0
+        for hf in host_file_list:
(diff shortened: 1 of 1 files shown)


risaacson:
@iiordanov What behavior are you trying to change for this? I have done some testing but I am not seeing a change in the ssh connection plugin execution.

iiordanov:
If machines _are_ in /etc/ssh/ssh_known_hosts file, but _are not_ in the
user used to run ansible's ${HOME}/.ssh/known_hosts file, then ansible does
not "trust" them and parallel execution is broken.

On Mon, Mar 3, 2014 at 11:26 PM, Richard Isaacson
notifications@github.comwrote:

> @iiordanov https://github.com/iiordanov What behavior are you trying to
> change for this? I have done some testing but I am not seeing a change in
> the ssh connection plugin execution.
> 
> ## 
> 
> Reply to this email directly or view it on GitHubhttps://github.com/ansible/ansible/pull/6156#issuecomment-36591124
> .

## 

The conscious mind has only one thread of execution.

risaacson:
@iiordanov That helps out a bit and I don't think I would have figured this out right away without that comment.

I was able to do a|b testing between devel and a branch with this change cherry-picked. The files I used for testing are in this gist: https://gist.github.com/risaacson/8097b8635ee7353ac2a1

Next time it would be helpful if you apply your changes to a branch that has a descriptive name. The CONTRIBUTING.md document.

Thanks for the contribution! This will definitely help speed up runs in some instances.

mpdehaan:
It should be noted that this doesn't cover paramiko, so if someone wants paramiko to have the same behavior (about knowing when to lock to ask questions) it would need to be made more intelligent.

I'm ok with this, for the most part, but in case someone finds this later, heads up that "paramiko_ssh.py" was not yet modified.

iiordanov:
Thanks guys! I don't think I'll have a chance to work on paramiko soon, but
if I do, I'll let you know.

Cheers!
iordan

On Tue, Mar 4, 2014 at 10:04 AM, Michael DeHaan notifications@github.comwrote:

> It should be noted that this doesn't cover paramiko, so if someone wants
> paramiko to have the same behavior (about knowing when to lock to ask
> questions) it would need to be made more intelligent.
> 
> I'm ok with this, for the most part, but in case someone finds this later,
> heads up that "paramiko_ssh.py" was not yet modified.
> 
> ## 
> 
> Reply to this email directly or view it on GitHubhttps://github.com/ansible/ansible/pull/6156#issuecomment-36632925
> .

## 

The conscious mind has only one thread of execution.

PR #6157, opened 2014-02-24
Fix expires

rgbkrk:
String interpolation typo causes the `expires` option not to work. This fixes it!


mpdehaan:
But this breaks exipres!  Just kidding, much appreciated!

rgbkrk:
Your humor is one of my favorite features of Ansible.

PR #5997, opened 2014-02-13
Elasticache vpc security groups

jonasi:
A couple of changes to `ec2_elb_lb` and `elasticache` to add better ec2 security group support.

For `elasticache`, only cache security groups which allow public access to elasticache instances.  I added a `security_group_ids` option to allow for internal ec2 applications access to elasticache (like inside a vpc).

For `ec2_elb_lb`, I added a `security_group_ids` options to allow for optionally setting one or more security groups to your elb.

PR #6162, opened 2014-02-25
Add options to ec2_elb_lb action to create ELBs inside of a VPC, and attach security groups

tdfischer:
This patch adds two new options to the elb action to support creating ELBs within a VPC, and for attaching security groups to it.


pattu777:
I was waiting for ELB creation inside of a VPC. Any idea about the next release date..??

pas256:
Is this a (partial) duplication of work with this and #5997?

mpdehaan:
Thanks, let's continue this over at #5997, sorry for duplicate efforts!

PR #6163, opened 2014-02-25
Add Ansible module for Atlassian JIRA

tarka:
This adds a module to create and manipulate issues in a JIRA instance.


mpdehaan:
Hi,

This is interesting!  

Can you submit this as a squashed commit to make it easier for pull request view purposes?

(Curious how you are also using it in your workflow... are you closing tickets after deploys and such?)

Thanks!

srgvg:
Also curious on this!

tarka:
Done: https://github.com/ansible/ansible/pull/6168

Re. workflows, we're not currently using this, it's something I've thought would be useful for a while but never had a chance to implement. But we have an occasional work hackathon called [ShipIt](https://www.atlassian.com/company/about/shipit), so I banged this out in a few hours.

I suspect the primary use-case would be to pull information from tickets to implement repetitive sysadmin tasks (e.g. the create-user example in the docs); we actually have plugins to our internal JIRA instance that does some of this stuff currently, but this feels like a better separation of duties. The more advanced stuff (i.e. transitions) is mostly just gravy, or the result of too much caffeine.

Cheers,
Steve

PR #5480, opened 2014-01-03
ec2_group support for egress rules and referring to security groups by name in rules

esp1:
This patch allows security groups to be referred to by name in rules by using the group_name parameter. If a named security group does not exist, it will be automatically created.

Rules now also support a type parameter to specify whether they are ingress or egress rules. If the type parameter is omitted the rule will be considered an ingress rule.

PR #6164, opened 2014-02-25
Egress rules for ec2 security groups

codezerk:
This will have the ability of adding egress rules, here is an example:

```
 - name: example ec2 group
   local_action:
     module: ec2_group
     name: example
     description: an example EC2 group
     vpc_id: 12345
     region: eu-west-1a
     ec2_secret_key: SECRET
     ec2_access_key: ACCESS
     rules:
       - proto: tcp
         from_port: 80
         to_port: 80
         cidr_ip: 0.0.0.0/0
         egress: yes <------------------ this will mark it as egress
       - proto: tcp
         from_port: 22
         to_port: 22
         cidr_ip: 10.0.0.0/8
```

When creating a security group on EC2, a default egress rule is always created, this rule gives unrestricted access to all protocols and all ports. This is unrelated to Ansible nor Boto.
When creating a security group using my changes, if no egress rules are configured then it will not change any egress rules that already exist on the group. If there are rules configured then similar to ingress rules it will remove the ones that are not configured.
I'm not sure if this is behavior is better for most users, but after contemplating I thought it is more reasonable.

### DIFF
--- a/library/cloud/ec2_group
+++ b/library/cloud/ec2_group
@@ -91,6 +91,12 @@ EXAMPLES = '''
       - proto: all
         # the containing group name may be specified here
         group_name: example
+      - proto: tcp
+        from_port: 0
+        to_port: 65535
+        cidr_ip: 0.0.0.0/0
+        # egress means that the rule will be assigned to outbound traffic
+        egress: yes
 '''
 
 try:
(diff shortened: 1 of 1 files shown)


pas256:
I believe this is a duplicate of #5480, and prefer #5480 as is uses `type: egress`.

mpdehaan:
Thanks guys, let's continue this ticket over at #5480 -- sorry for duplicate efforts!

PR #6168, opened 2014-02-25
Add Ansible module for Atlassian JIRA (squashed version)

tarka:
Initial version of an Ansible JIRA module to create, edit and transition JIRA issues.

### DIFF
--- a/library/web_infrastructure/jira
+++ b/library/web_infrastructure/jira
@@ -1,0 +1,344 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2014, Steve Smith <ssmith@atlassian.com>
+# Atlassian open-source approval reference OSR-76.
+#
+# This file is part of Ansible.
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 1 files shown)


jsmartin:
Just curious, could you explain the typical scenario when someone would use this module?  One example I thought of was opening/closing issues as some code was deployed with ansible.  Also, as someone mentioned on the mailing list, you'll want to use the freebie fetch_url method for your http interactions.

tarka:
Your scenario is exactly what I was thinking when I started; we use Ansible to deploy releases, and there is usually a ticket associated with a release, so automating the closing was the natural case. Others I've considered are:
- We raise tickets to create users for new hires; use Ansible to pull the information from the ticket (username, team etc) and create the user, then close the ticket automatically.
- Rather than closing the ticket, add a comment that the relevant work has been done (when other manual steps need to be performed).
- Raise tickets when deployments fail.
- Our sysadmin team are using ansible.runner as part of a Bottle app to automate taking a thread-dump of misbehaving JVM apps. They're looking at using the JIRA module to automatically raise a ticket and attach the thread-dump.

Ideally I'd also like to add JIRA management operations too (i.e. to deploy and configure a JIRA instance), but I figured issue manipulation was the low-hanging fruit.

tarka:
Superceded by https://github.com/ansible/ansible/pull/6512

PR #5639, opened 2014-01-16
Docker: names, update port mapping support, add `publish_all_ports` flag, add container links

joshuaconner:
This builds on #5301, but adds updated port-mapping support and the `publish_all_ports` flag. It also supports container links. LMK if you have any questions!

PR #6171, opened 2014-02-25
docker: update to docker-py 0.3.0, plus full port binding support etc. from #5639

joshuaconner:
This rebases the changes against #5639 from the current devel branch, plus adds (well, requires) support for `docker-py` 0.3.0, which is the latest version. (See discussion from `ansible-devel` list here: https://groups.google.com/forum/#!topic/ansible-devel/L57eChJ93Tk)

### DIFF
--- a/library/cloud/docker
+++ b/library/cloud/docker
@@ -1,6 +1,7 @@
 #!/usr/bin/python
 
 # (c) 2013, Cove Schneider
+# (c) 2014, Joshua Conner <joshua.conner@gmail.com>
 # (c) 2014, Pavel Antonov <antonov@adwz.ru>
 #
 # This file is part of Ansible,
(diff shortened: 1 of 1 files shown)


mpdehaan:
This is in queue, thanks! 

Can you add the "version_added" attribute on the new parameters so folks can tell they weren't available in 1.4.X?

joshuaconner:
Done!

jctanner:
@joshuaconner thanks for the patch!

I did have trouble getting the module to set the nat table rules for the forwarding, but that appears to be docker-py at fault rather than the module code.

If anyone has issues getting the port forwarding to work, check your iptables rules to make sure docker-py is functioning ...

```
$ docker run -d -p 8080:80 jt/httpd /usr/sbin/httpd -DFOREGROUND
$ iptables -L -n -t nat
Chain DOCKER (2 references)
target     prot opt source               destination
DNAT       tcp  --  0.0.0.0/0            0.0.0.0/0            tcp dpt:8080 to:172.17.0.2:80
```

EDIT:

Narrowed it down further to the docker daemon. Docker-py sends the right json post data to the daemon, but no rules are added to the docker iptables chain. strace'ing the daemon reveals that no calls to the iptables command happen when going through the docker-py path, but calls are made when using "docker run"

joshuaconner:
Did you add an [EXPOSE](http://docs.docker.io/en/latest/reference/builder/#expose) directive to your Dockerfile? You can now expose through the 1.8 or 1.9 API also but the module doesn't do it right now.

PR #6175, opened 2014-02-26
filesystem: Ignore blkid cache

AdmiralNemo:
Sometimes, `blkid` will incorrectly return no information about a block
device, even if it exists and has a valid filesystem. This causes the
_filesystem_ module to fail if _force=no_. Instructing `blkid` to use
`/dev/null` as a cache file will force it to rescan the block device on
each run, making results more consistent.

Signed-off-by: Dustin C. Hatch admiralnemo@gmail.com

### DIFF
--- a/library/system/filesystem
+++ b/library/system/filesystem
@@ -79,7 +79,7 @@ def main():
 
     cmd = module.get_bin_path('blkid', required=True)
 
-    rc,raw_fs,err = module.run_command("%s -o value -s TYPE %s" % (cmd, dev))
+    rc,raw_fs,err = module.run_command("%s -c /dev/null -o value -s TYPE %s" % (cmd, dev))
     fs = raw_fs.strip()
 
 


mpdehaan:
Can you share information about how to reproduce this problem?

Also OS you are using, etc?

AdmiralNemo:
Without my patch, the following playbook fails 9 out of 10 times:

```
- name: 'Test for issue #6175'
  hosts: '!localhost'
  tasks:
    - name: create lvm volume
      lvol: lv=test6175
            vg=vmachines
            size=2G
    - name: create filesystem
      filesystem: dev=/dev/vmachines/test6175
                  fstype=xfs
```

This is the error:

```
failed: [managed.host.fqdn] => {"err": "mkfs.xfs: /dev/vmachines/test6175 appears to contain an existing filesystem (xfs).\nmkfs.xfs: Use the -f option to force overwrite.\n", "failed": true, "rc": 1}
msg: Creating filesystem xfs on device '/dev/vmachines/test6175' failed
```

If I set _fstype_ to `ext4` instead of `xfs`, instead of failing with an error, the task is reported as changed every time, since `mkfs.ext4` does not care if there is an existing filesystem or not.

Playbook was invoked with:

```
ansible-playbook -i hosts -l managed.host.fqdn 6175.yml -sK
```

Managed operating system is Gentoo amd64
- sys-apps/util-linux-2.22.2
- sys-fs/xfsprogs-3.1.10

Management machine is Gentoo ~amd64
- app-admin/ansible-1.4.3

mpdehaan:
merged, thanks!

Issue #6080, opened 2014-02-19
stat module returns mode 420 for a file that is 0644 


jctanner:
##### Issue Type:

Bug Report
##### Ansible Version:

```
ansible 1.5 (devel ee6a0359f5) last updated 2014/02/19 16:45:25 (GMT -400)
```
##### Environment:

ubuntu 12.04 on both hosts
##### Summary:

A file with 0644 permissions is reported as 420 by the stat module.
##### Steps To Reproduce:

```
- name: make a new file
  copy: dest=/tmp/foo.txt mode=0644 content="hello world"

- name: check stat of file
  stat: path=/tmp/foo.txt
  register: stat_result

- assert:
    that:
        - "stat_result.stat.mode == 644"
```
##### Expected Results:

Assert task does not fail.
##### Actual Results:

Assert task fails.

nilsbunger:
The mode is correct. 644 is an octal representation, which is 420 in decimal.

If you change the assertion to
    - "{{ '%o' % stat_result.stat.mode }} == 644"

then your assertion passes.

jimi-c:
I think it would actually probably be better to fix the mode to correctly show the octal display in the stat module as follows:

<pre>
diff --git a/library/files/stat b/library/files/stat
index 6e7dcd6..abece4a 100644
--- a/library/files/stat
+++ b/library/files/stat
@@ -88,7 +88,7 @@ def main():
     # back to ansible
     d = {
         'exists'   : True,
-        'mode'     : S_IMODE(mode),
+        'mode'     : "%04o" % S_IMODE(mode),
         'isdir'    : S_ISDIR(mode),
         'ischr'    : S_ISCHR(mode),
         'isblk'    : S_ISBLK(mode),
</pre>

I think users would expect to see the octal representation there.

nilsbunger:
Maybe you're right. But it would be backwards-incompatible to just change it, and there are advantages to the way it is now (easier to do a test like "is this file readable" with bit masking). Maybe a project lead can weigh in on what they want?

bcoca:
add another field returned 'octal_perms'?

mpdehaan:
I think I agree with jimi-c here in that we should fix the glitch.  Everybody thinks about these things in octal.

risaacson:
Closed as a part of #6221.

PR #6176, opened 2014-02-26
added octal representation of mode and made md5 checksumming optional

### LINKED ISSUE #6174: Enhancement: stat module - allow skipping of md5sum
I'd like to be able to skip the md5 check sum when running the stat module on a file.

My scenario is that I have a dvd.iso file located on one of my ansible_managed machines.  When I run stat, it takes a long time for the module to return since the file is about 4GB.  In this case, I would like just to see that the file exists at the remote location and don't really care about the md5sum.  Would it be possible to implement an option for this module to skip performing the md5 check sum?

Thanks,
-Razi

bcoca:
this should close #6174 and #6080

Signed-off-by: Brian Coca <briancoca+dev@gmail.com>

### DIFF
--- a/library/files/stat
+++ b/library/files/stat
@@ -34,6 +34,12 @@ options:
     required: false
     default: no
     aliases: []
+  get_md5:
+    description:
+      - Whether to return the md5 sum o fthe file
+    required: false
+    default: yes
+    aliases: []
 author: Bruce Pennypacker
 '''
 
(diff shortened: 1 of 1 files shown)


rkhaja:
Wow that was super fast. Thank you!

mpdehaan:
This looks pretty clean, +1

mpdehaan:
See my comments on the other ticket, BTW, when we merge this we may want to just fix the input.

PR #6178, opened 2014-02-26
docker_image: added support for docker_py >= 0.3.0, Docker API >= 1.8. Merged with #6171

softzilla:
Fixes problems with new versions of Docker (Docker API, docker-py).
Also merge with @joshuaconner docker module implementation #6171.


mpdehaan:
Please don't merge other's pull requests in like the above.   It would be ok to cherry-pick things but we don't allow merge requests in PRs since they complicate review.

In this case there are some unrelated changes to the chnagelog in that merge commit, so we can't accept this as is.

Thanks!

softzilla:
Got it.  Replace pull request with #6196

PR #6183, opened 2014-02-26
mysql_db: Emphasize target is on remote host. List supported file types.

gkoller:
From the documentation it is not immediately clear that the 'target'
option refers to a location on the remote host. This change emphasizes that.
In addition to .sql files, .bz2 and .gz files are supported for dumps and
restores. This is now documented.

### DIFF
--- a/library/database/mysql_db
+++ b/library/database/mysql_db
@@ -76,7 +76,8 @@ options:
     default: null
   target:
     description:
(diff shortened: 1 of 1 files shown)


mpdehaan:
Merged, thank you!  This will appear the next time we push docs.

PR #6184, opened 2014-02-26
Properly match existing security group

patrys:
If we don't care about `vpc_id` then neither should the code.


mpdehaan:
Merged, thanks!

PR #6198, opened 2014-02-26
First pass at ec2 module tests

jlaska:
Includes a general framework for designing amazon module tests, along with tests for ec2_key and ec2_group.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,4 @@ debian/
 # Vim swap files
 *.swp
 *.swo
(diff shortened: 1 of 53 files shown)


mpdehaan:
Great!  Let's talk about this tomorrow in actual people space and we can share details w/ the list and such!

mpdehaan:
Merged, thanks!  We'll post to the list in a bit and share details about where this is going (and also about other clouds too)

PR #6200, opened 2014-02-27
EC2 VPC "instance_tenancy"

risaacson:
This pull allows you at creation time to specify that the VPC is default or dedicated. The default is that instance_tenancy="default".


dfederlein:
Tested and can confirm working as intended.

jimi-c:
This was merged, but as a squashed commit so github isn't picking it up, so closing manually.

PR #5279, opened 2013-12-13
Adding 'export' option to subversion module (to export instead of co)

magnetik:
Our deploy process requires to export to a directory to have a full "static" version of each deployed version.

PR #6201, opened 2014-02-27
Escape subversion parameters for running svn commands

cchurch:
The subversion module breaks when its parameters contain single quotes.  In particular, you cannot use a password that contains any number of single quote (`'`) characters.

Due to the way the svn command line arguments are assembled, it is also possible to run arbitrary commands by setting certain parameters to carefully crafted strings (e.g. `username="'; rm -rf /"`).

This patch passes all svn arguments as a list of strings to allow subprocess.Popen to escape them properly and adds an integration test to verify single quotes are handled correctly.

### DIFF
--- a/library/source_control/subversion
+++ b/library/source_control/subversion
@@ -94,47 +94,46 @@ class Subversion(object):
 
     def _exec(self, args):
         bits = [
-            'LANG=C',
             self.svn_path,
             '--non-interactive',
             '--trust-server-cert',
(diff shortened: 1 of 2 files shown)


cchurch:
This pull request will break the changes for #5279 and #5347 -- maybe accept those first and I can rebase and update this one?

mpdehaan:
Merged, thanks!

Issue #5621, opened 2014-01-14
rpm_key doesn't work for el5


WramblinWreck:
Using Ansible 1.4.3
ansible running on Mac OSX Mavricks
ansible target machine CentOS 5.8

In trying to get the rpm_key module to run against a centos 5.8 box, I noticed a few issues with the module:
1. neither gpg nor gpg2 (gnupg and gnupg2 packages) for centos 5.8 return output that contains a line that starts "keyid:".  The module then throws an exception saying gpg didn't return the expected output and fails.
1. on CentOS 6.\* the output does contain keyid: on two of it's output lines (for the key I'm trying to import...http://pkg.jenkins-ci.org/redhat/jenkins-ci.org.key), but from what I can read in the ansible source, the module uses the keyid returned from gpg to stay idempotent by comparing that output to the output of "rpm -q gpg-pubkey".  However, in both centos 5 and centos 6, these two things are not the same in my case and I would guess in general. 
2. Another slightly less important issue is with the documentation itself.  Other modules docs state external package dependencies so they are clear, this module doesn't list that it depends on gpg you have to look at the source to determine it.

mpdehaan:
This should be resolved by the above merge commit.

mpdehaan:
(please reopen if not).

PR #6203, opened 2014-02-27
Fix issue #5621, rpm_key doesn't work for el5

garetharmstronghp:
This small patch fixes issue #5621 and enables the rpm_key to work on Rhel 5 and derivative distros while still ensuring compatibility with more recent Red Hat ones.  The first part of the patch just adds Python 2.4 support and the second part grabs the keyid from the signature packet listing of the gpg command.


mpdehaan:
merged, thanks!

PR #6207, opened 2014-02-27
A module for interacting with Linux kernel capabilities through getcap/setcap

natefoo:
Various privileges that are normally only available to root can be granted to unprivileged users with the `capabilities(7)` feature in the Linux kernel. This module allows managing those capabilities on a per-file basis.


mpdehaan:
Merged, thank you!

mpdehaan:
github seemed to not auto-close this.  Bad github.  Doing this now.

PR #6218, opened 2014-02-28
Copy Module: Add the ability to set directory attributes on recursive copy.

### LINKED ISSUE #6194: Copy module does not apply owner and group to directories
##### Issue Type:

Bug Report
##### Ansible Version:

```
ansible 1.5 (devel 33857855ad) last updated 2014/02/26 17:10:07 (GMT +100)
```
##### Environment:

Running Ansible from: Debian 7 (Wheezy)
Managing: CentOS 6.5
##### Summary:

When using the copy module with an source value that is a directory, and attempting to use the `owner` and `group` options of the file module, the options are applied correctly to any files within the copied directory, but they are not applied to the directory itself or further directories within this directory.
##### Steps To Reproduce:

Run this playbook task against the host.

```
- copy: src=/etc/emacs dest=/tmp/test-copy owner=daemon group=daemon mode=500
```
##### Expected Results:

I expect `/tmp/test-copy` to have ownership `daemon` and group `daemon`.  I expect `/tmp/test-copy/emacs` and any subdirectories to have that ownership also.
##### Actual Results:

```
# ls -ld /tmp/test-copy/
drwxr-xr-x. 3 root root 4096 Feb 26 17:17 /tmp/test-copy/
# ls -ld /tmp/test-copy/emacs
drwxr-xr-x. 3 root root 4096 Feb 26 17:17 /tmp/test-copy/emacs
# ls -ld /tmp/test-copy/emacs/site-start.d
drwxr-xr-x. 2 root root 4096 Feb 26 17:17 /tmp/test-copy/emacs/site-start.d
# ls -ld /tmp/test-copy/emacs/site-start.el 
-r-x------. 1 daemon daemon 412 Feb 26 17:17 /tmp/test-copy/emacs/site-start.el
```

risaacson:
Closes GH-6194.

These changes pass attribute modifications through to new directories during a recursive copy with an addition allowing for the directory mode to be set independently.

Tested with: https://gist.github.com/risaacson/4a8a1e8e344f2327b6de

### DIFF
--- a/library/files/copy
+++ b/library/files/copy
@@ -78,8 +78,8 @@ options:
     version_added: "1.2"
   directory_mode:
     description:
-      - When doing a recursive copy set the mode for the directories. If this is not set we will default to the file
-        permissions if set.
+      - When doing a recursive copy set the mode for the directories. If this is not set we will default the system
+        defaults.
     required: false
     version_added: "1.5"
   others:
--- a/library/files/copy
+++ b/library/files/copy
@@ -76,6 +76,12 @@ options:
     required: false
     default: ""
     version_added: "1.2"
+  directory_mode:
+    description:
+      - When doing a recursive copy set the mode for the directories. If this is not set we will default to the file
+        permissions if set.
+    required: false
+    version_added: "1.5"
   others:
     description:
       - all arguments accepted by the M(file) module also work here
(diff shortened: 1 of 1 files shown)


mpdehaan:
One possible thing.

When setting permissions on directories (for instance the file would be 644) most likely, but not always, we'd want the directory to be 755

Also a mode flag being a boolean is a little strange.   Might be better to call this set_directory_permissions=True (default False) and then use the "required_together" macro in module_utils to make that parameter only valid when used with recursive=yes

mpdehaan:
Looks like I totally misread the above patch and you do have directory mode as octal, which makes more sense.  Let's keep that the way you have it.

(FYI: You appear to have a tab on line 135, this can be fixed by configuring your editor for py files to insert 4 spaces for a tab)

risaacson:
Fixed the behavior for a directory when directory_mode is not defined. It will now default to the system default. (i.e. umask).

It came up in chat that maybe if directory_mode is defined and it is not recursive that we should throw an error. With the copy module it ends up being cleaner to not do anything as no changes will be made. I augmented the documentation a little bit to make this more clear.

PR #4617, opened 2013-10-21
apt: autoinstall python-apt if apt or apt_pkg is not available

resmo:
This PR fixes #4079.

Issue #4079, opened 2013-09-10
option to auto-install python-apt module


iivvoo:
If the 'apt' module is used but python-apt is not installed on the target host, ansible bails out with:

```
TASK: [Install sudo] ********************************************************** 
failed: [hoppi] => {"failed": true}
msg: Could not import python modules: apt, apt_pkg. Please install python-apt package.

FATAL: all hosts have already failed -- aborting
```

which means you're always forced to do some workaround to make sure python-apt is installed.

An option to somehow auto-install python-apt if necessary would be nice.

mpdehaan:
This would actually be a pretty good idea, and I don't think it needs to be an option.

The module is a packaging module and has a non-stock-OS dependency, which is a problem for people with stock images and makes the initial user experience on Debian/Ubuntu rougher than on Red Hat.

We should test if it's not there and then feed in the command line to auto-install it.

resmo:
We should considering to add this to the documentation. 

If you are doing a minimal installation, you may not have python as well. So bootstrapping ansible dependencies is not only an problem using the apt module.

So better doing less magic, print an good error message so users just can do:

```
ansible -m raw -a "apt-get install python-apt -y"
```

mpdehaan:
No, it should not be documentation, we should auto install the module.

If someone left out Python they ignored most of the install instructions for using Ansible, so I'm less concerned with that.

jctanner:
@resmo can you make a pull request for https://github.com/resmo/ansible/commit/ba561dbefd4d8e89d4d75eee8d641703185cc64d ?

marcusramberg:
I still see this issue with Ansible 1.5.5. Which release did this change go into?

jimi-c:
@marcusramberg this is definitely in 1.6.0.

Wilfred:
Similarly, apt_repository depends on python-curl. Should this be autoinstalled too?

jimi-c:
@Wilfred the installation is done via apt, so deps should be installed along with that package.

d-mo:
I'm getting this error with Ansible version 1.8.2. Downgrading to 1.7.2 fixes it.

hwmrocker:
I get this error with ansible 1.9.1, and 1.7.2 will not install in a virtual env because it cannot delete stuff in /usr/share WTF

anton-rudeshko:
On Ubuntu Hardy 8.04 I always get: 

```
Some packages could not be installed. This may mean that you have
[...]

PR #6220, opened 2014-02-28
apt_repository: autoinstall python-apt if not available

trevorah:
This is almost identical to the apt issue (#4079) and its corresponding fix (#4617)

### DIFF
--- a/library/packaging/apt_repository
+++ b/library/packaging/apt_repository
@@ -366,7 +366,13 @@ def main():
     )
 
     if not HAVE_PYTHON_APT:
-        module.fail_json(msg='Could not import python modules: apt_pkg. Please install python-apt package.')
+        try:
+            module.run_command('apt-get update && apt-get install python-apt -y -q')
+            global apt, apt_pkg
+            import apt
+            import apt_pkg
+        except:
+            module.fail_json(msg='Could not import python modules: apt, apt_pkg. Please install python-apt package.')
 
     if not HAVE_PYCURL:
         module.fail_json(msg='Could not import python modules: pycurl. Please install python-pycurl package.')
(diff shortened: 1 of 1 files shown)


lorin:
+1, this would be a great enhancement.

resmo:
+1

jctanner:
@trevorah can you rework this based on the way it was implemented in the yum module?

https://github.com/ansible/ansible/blob/devel/library/packaging/yum#L158-L167

I prefer explicit calls and return code checks over wrapping everything in a try/except block.

trevorah:
@jctanner sure, give me a sec...

resmo:
@trevorah and also in the `apt`-module as well. plz :)

trevorah:
@jctanner do you mean something like this?

mpdehaan:
Pretty much.

Now that module.run_command does not use the shell by default this should be modified to split this up into two commands instead of using "&&", or alternatively pass use_unsafe_shell=True to run command.

Also since module_utils/urls has the "fetch_url" code we can also remove the dependency on pycurl at the same time.

jctanner:
@trevorah thank you!

I remove the pycurl functions since they're no longer needed per: 7af8a33def34519de5ca390c47bdbd432d7b7ee2

If you'd like to do the same for the apt module, I would be happy to review.

trevorah:
@jctanner no problem, glad I could help! Sorry for the delay btw, python isn't my first language.

gdevenyi:
This change disagrees with http://docs.ansible.com/intro_installation.html#managed-node-requirements

Was this going to be documented at some point? I ran an ansible check on some systems only to find it install packages _WITHOUT ASKING_

PR #6223, opened 2014-02-28
Support ssh-ed25519 keys

lvh:
The newest version of OpenSSH supports a new, wonderful key type. authorized_key incorrectly discards pubkeys of this type as busted because it doesn't recognize type signature.


lvh:
I think @bradobro might be interested in this.

bcoca:
+elliptical curve

mpdehaan:
merged, thanks!

PR #6224, opened 2014-02-28
Fully delete 'ssh_old' and 'paramiko_alt' options

gildegoma:
These transports have been removed in 33857855ad6d358d56d5fe60bd49be653b247e0b.


mpdehaan:
Good catch on the leftover code, thanks!

PR #6230, opened 2014-03-01
nova_compute: fix for partial match b/w params['name'] and an existing name

joshuaconner:
When there is an Openstack instance that has a name that's a partial match for module.params['name'], but a server with name module.params['name'] doesn't yet exist*, this module would fail with a list index out of bounds error. This fixes that by filtering by exact name and only **then** getting the server from the list if the list is still not empty.

\* For example, a server with name 'nginx-prod' exists, and module.params['name'] = 'nginx'.

### DIFF
--- a/library/cloud/nova_compute
+++ b/library/cloud/nova_compute
@@ -197,7 +197,8 @@ def _get_server_state(module, nova):
             # with names that partially match the server name, so we have to
             # strictly filter here
             servers = [x for x in servers if x.name == module.params['name']]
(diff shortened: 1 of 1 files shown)


mpdehaan:
Thanks!  Per the code guidelines doc in root can  you split the ternary into multiple lines?

Then this looks safe to merge.

Much appreciated!

joshuaconner:
Fixed!

mpdehaan:
merged, thanks!

PR #6233, opened 2014-03-01
Added modules to manage ec2 autoscaling groups

garethr:
Includes management of Launch Configuration and related Autoscaling
Groups.

Based on a complete working example here: https://github.com/garethr/ansible-provisioner-aws

### DIFF
--- a/library/cloud/ec2_asg
+++ b/library/cloud/ec2_asg
@@ -21,7 +21,7 @@ short_description: Create or delete AWS Autoscaling Groups
 description:
   - Can create or delete AWS Autoscaling Groups
   - Works with the ec2_lc module to manage Launch Configurations
-version_added: "1.0"
+version_added: "1.6"
 requirements: [ "boto" ]
 author: Gareth Rushgrove
 options:
--- a/library/cloud/ec2_asg
+++ b/library/cloud/ec2_asg
@@ -30,7 +30,7 @@ options:
       - register or deregister the instance
     required: true
     choices: ['present', 'absent']
-  group:
+  name:
     description:
       - Unique name for group to be created or deleted
     required: true
@@ -75,7 +75,7 @@ options:
 
 EXAMPLES = '''
 - ec2_asg: >
-    group_name: special
+    name: special
     load_balancers: 'lb1,lb2'
     availability_zones: 'eu-west-1a,eu-west-1b'
     launch_config_name: 'lc-1'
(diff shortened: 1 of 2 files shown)


mpdehaan:
Excellent, glad to see this.  We're going to do a push to merge a _LOT_ of the cloud stuff in a little over 1-2 weeks.   This will be a great addition.

garethr:
Excellent. Any feedback/changes/etc. just let me know

library/cloud/ec2_asg
```diff
@@ -0,0 +1,190 @@
...
+version_added: "1.0"
+requirements: [ "boto" ]
+author: Gareth Rushgrove
+options:
+  state:
+    description:
+      - register or deregister the instance
+    required: true
+    choices: ['present', 'absent']
+  group:
```
pas256: `group` does not match `group_name` on line 78, and does not match `name` on line 111 and 137.

My preference is to change all of them to simple `name`.

pas256:
I would like to see this support 2 more things
- `resource_tags` similar to issue #6085 
- support for VPCs

Otherwise, great work.

library/cloud/ec2_asg
garethr: Thanks, I've corrected the documentation and the example. I agree, `name` provides a better interface.

I've left the internal variable name as `group_name` however as that maps to the actual AWS name. Making this more explicit if you're in the code feels useful to me.

library/cloud/ec2_asg
```diff
@@ -0,0 +1,190 @@
...
+
+try:
+    import boto.ec2.autoscale
+    from boto.ec2.autoscale import AutoScaleConnection, AutoScalingGroup
+    from boto.exception import BotoServerError
+except ImportError:
+    print "failed=True msg='boto required for this module'"
+    sys.exit(1)
+
+AWS_REGIONS = ['ap-northeast-1',
```
willthames, line 100: No need to define AWS_REGIONS in modules that include module_utils/ec2.py

willthames:
Would be good to handle existing launchconfigurations in the create_autoscaling_groups (same as how you check that it exists before you delete it) rather than catch it in the same exception that will handle connection/authentication failures.

The more information a module returns, the better, in general.

timperrett:
Yes!! Love to see this merged!

mpdehaan:
Merged, thank you!  We would be also welcome to entertain future upgrades if folks have suggestions/additions.

willthames:
Further improvements in #6419.

PR #6238, opened 2014-03-02
allow for ssh-specific arguments

jeremyherbert:
Hi,

scp and scp are not necessarily argument compatible. For example, the -R switch (which allows you to create a reverse tunnel) is usable on SSH only and not SCP. I am trying to use this in an ansible script but unfortunately there is no configuration option to pass arguments to just the ssh command and not scp/sftp.

This code allows you to specify commands that will only be passed to ssh via ssh_specific_args in ansible.cfg.

### DIFF
--- a/lib/ansible/constants.py
+++ b/lib/ansible/constants.py
@@ -152,6 +152,7 @@ DEPRECATION_WARNINGS           = get_config(p, DEFAULTS, 'deprecation_warnings',
 
 # CONNECTION RELATED
 ANSIBLE_SSH_ARGS               = get_config(p, 'ssh_connection', 'ssh_args', 'ANSIBLE_SSH_ARGS', None)
+ANSIBLE_SSH_SPECIFIC_ARGS      = get_config(p, 'ssh_connection', 'ssh_specific_args', 'ANSIBLE_SSH_SPECIFIC_ARGS', None)
 ANSIBLE_SSH_CONTROL_PATH       = get_config(p, 'ssh_connection', 'control_path', 'ANSIBLE_SSH_CONTROL_PATH', "%(directory)s/ansible-ssh-%%h-%%p-%%r")
 ANSIBLE_SSH_PIPELINING         = get_config(p, 'ssh_connection', 'pipelining', 'ANSIBLE_SSH_PIPELINING', False, boolean=True)
 PARAMIKO_RECORD_HOST_KEYS      = get_config(p, 'paramiko_connection', 'record_host_keys', 'ANSIBLE_PARAMIKO_RECORD_HOST_KEYS', True, boolean=True)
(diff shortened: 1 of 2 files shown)


mpdehaan:
Are you actually passing -R with ansible?  I'm not sure that entirely makes sense with Ansible's calling architecture and perhaps is something that should be done explicitly instead?

Basically I'm wanting to understand more about the underlying use case prior to adding this.

Thanks!

jeremyherbert:
Hi again,

I am using the -R flag successfully now with this change. My use case is this: I have hosts which have extremely limited internet access; they have 443 forwarded via a proxy on the frontend, and I can SSH to them. That's about it and it is not under my control. Unfortunately, I need to do things like pull code off github, download packages and various other internet related tasks. 

The way I have solved this is to run tinyproxy locally on my machine, and forward a port back to it using SSH arguments. Then, simply by adding an extra environment attribute to a task, I can selectively give tasks access to the internet for things like wget. For example:

ansible.cfg:

```
[ssh_connection]
ssh_specific_args = -R 12346:localhost:12345
```

vars.yml:

```
--
proxy_environment:
  http_proxy: http://localhost:12346
  https_proxy: http://localhost:12346
```

playbook.yml:

```
---
- hosts: all
  vars_files:
    - vars.yml
  sudo: True
  user: deploy

  tasks:

    - name: Download neo4j
      shell: "wget \"http://download.neo4j.org/artifact?edition=community&version=2.0.1&distribution=tarball&dlid=3646018\" -O ~/neo4j-community-2.0.1-unix.tar.gz creates=~/neo4j-community-2.0.1-unix.tar.gz"
      environment: proxy_environment
```

I have just used this over the weekend to do a bunch of deploying and I have not encountered any problems.

jimi-c:
Isn't this something that could be passed with the ANSIBLE_SSH_ARGS environment variable or the ssh_args option already in the ansible.cfg file? This seems to duplicate that functionality.

mpdehaan:
@jimi-c I believe the key part of the issue was " I am trying to use this in an ansible script but unfortunately there is no configuration option to pass arguments to just the ssh command and not scp/sftp."

sigio:
It would be nice to have these options be configurable in the inventory on a per-host or per-group method. I have a group of hosts where I need to specify '-s /usr/lib/sftp-server' to use sftp.

jimi-c:
@sigio please open a new issue for that, if you have not already, so we can keep track of the request. Thanks!

sigio:
I had... it was merged into this one by mpdehaan. (#8637)

jimi-c:
@sigio my apologies, I didn't see that. Thanks!

amenonsen:
This PR can be closed. I've submitted a rebased version as #11681.

gregdek:
Closed because #11908 now appears to cover this case.

PR #6240, opened 2014-03-02
get rid of newline chars when reading password file

fesplugas:
To be consistent with: https://github.com/ansible/ansible/blob/devel/bin/ansible-playbook#L130-L131


jctanner:
@fesplugas thanks!

PR #6248, opened 2014-03-03
Fix incorrect use of copy() on list, causing a traceback

hkariti:
Similar issue to  #6150, this time with tag handling.


mpdehaan:
Merged, thanks!

PR #6251, opened 2014-03-03
Fix regression causing setup to run for every play

dagwieers:
This patch makes sure setup only runs once, unless it is enforced on a play.

### DIFF
--- a/lib/ansible/playbook/__init__.py
+++ b/lib/ansible/playbook/__init__.py
@@ -479,6 +479,10 @@ class PlayBook(object):
             return {}
 
         host_list = self._trim_unavailable_hosts(play._play_hosts)
+        if play.gather_facts is None:
+            host_list = [h for h in host_list if h not in self.SETUP_CACHE or 'module_setup' not in self.SETUP_CACHE[h]]
+            if len(host_list) == 0:
+                return {}
 
         self.callbacks.on_setup()
         self.inventory.restrict_to(host_list)
--- a/lib/ansible/playbook/play.py
+++ b/lib/ansible/playbook/play.py
@@ -117,7 +117,7 @@ class Play(object):
         self.sudo             = ds.get('sudo', self.playbook.sudo)
         self.sudo_user        = ds.get('sudo_user', self.playbook.sudo_user)
         self.transport        = ds.get('connection', self.playbook.transport)
-        self.gather_facts     = ds.get('gather_facts', True)
+        self.gather_facts     = ds.get('gather_facts', None)
         self.remote_port      = self.remote_port
         self.any_errors_fatal = utils.boolean(ds.get('any_errors_fatal', 'false'))
         self.accelerate       = utils.boolean(ds.get('accelerate', 'false'))


jctanner:
looks like this was intended behavior: c039e276a2089cac804db4b0ea9072d8a27fee86

jimi-c:
@dagwieers one thing to add, I noticed this has actually been wrong for a while, but the code should use the booleans utility call: `utils.boolean(ds.get('gather_facts', 'false'))`

Can you fix that in this patch as well?

dagwieers:
@jimi-c Well, internally gather_facts is NOT a boolean, it has three states True, False and None.

mpdehaan:
It used to have those three states and I remember the patch that added it, however, it was never documented.

The code above was removed by me because it was causing some problems elsewhere and wasn't backed by any tests or documented, and I believed the behavior was rather circumspect anyway, and it felt reasonable to revisit the logic.

gather_facts: True/False is controllable every play, but the code in question would not re-evaluate if it already talked to a host, unless someone remembered to force gather_facts: True, which also seemed unintuitive.

As such, I don't believe this behavior should be reinstated.

mpdehaan:
Here is the link we meant to link above BTW:

https://github.com/dagwieers/ansible/commit/c039e276a2089cac804db4b0ea9072d8a27fee86

dagwieers:
With the following simple playbook, you get all hosts to repopulate facts at the start of _every_ play. It means running setup 3 times:

``` yaml
- name: First play
  hosts: all
  tasks:
  - action: ping

- name: Second play
  hosts: all
  tasks:
  - action: ping

- name: Third play
  hosts: all
  tasks:                     
  - action: ping
```

If that is the intended behaviour I am not going to argue with you. But it seems rather strange you are changing something that worked as expected (run only setup for every host the very first time) to something that runs setup at the start of every play, unless you disable it for a play.

So in case your first play does not cover all hosts, you cannot disable gather_facts in the second play because you risk having no facts for some hosts. I do not understand why you think that is a good thing.

And I am talking about this: https://github.com/dagwieers/ansible/commit/87258f6d428a74c5da0eeda76f797981a9418984

bcoca:
would this be an acceptable compromise? (also adding it to docs of course!)

```
   diff --git a/lib/ansible/playbook/__init__.py b/lib/ansible/playbook/__init__.py
   index 6596552..e6e2a06 100644
   --- a/lib/ansible/playbook/__init__.py
   +++ b/lib/ansible/playbook/__init__.py
   @@ -479,6 +479,10 @@ class PlayBook(object):
                return {}

            host_list = self._trim_unavailable_hosts(play._play_hosts)
   +        if isinstance(play.gather_facts, basestring) and  play.gather_facts.lower() == 'cached':
   +            host_list = [h for h in host_list if h not in self.SETUP_CACHE or 'module_setup' not in    self.SETUP_CACHE[h]]
   +            if len(host_list) == 0:
   +                return {}

            self.callbacks.on_setup()
            self.inventory.restrict_to(host_list)
```

mpdehaan:
Brian submitted https://github.com/ansible/ansible/pull/6508 which I think we're going to go with, with the possible addition of making smart not the default.  (Need to think about this).

So, given list conversations, going to go with that one, which is currently open as a ticket.

PR #6252, opened 2014-03-03
`docker pull' image instead of building if no path is given

hkariti:
Resending #6065

### DIFF
--- a/library/cloud/docker_image
+++ b/library/cloud/docker_image
@@ -152,6 +152,14 @@ class DockerImageManager:
 
         return image_id
 
+    def pull(self):
+        stream = self.client.pull(repository=self.name, tag=self.tag, stream=True)
+        for res in stream:
+            pass
+        res = json.loads(res)
+        self.changed = True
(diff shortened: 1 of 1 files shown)


jctanner:
@hkariti can you please rebase and fix the merge conflicts from b1a3e367b6b4b88acef677f24cb6e483380f335d

hkariti:
Tried to add it nicely alongside build.
I also noticed a problem - pulling can download several images, and in that case the id returned will be chosen practically at random. Don't know how to tackle this, as I'm pretty sure build returns a single value.

jctanner:
@hkariti can you explain the new workflow here? I'm a bit confused about what this would do or what the intended use case is.

mpdehaan:
You may wish to coordinate this with #6340 ?

hkariti:
@mpdehaan I was about to mention that request as well. He seems to have done a more thorough job, so I think it's better to use that.

mpdehaan:
Ok, we will close this one out and track the other.  Thank you!

PR #6262, opened 2014-03-04
Allow ec2_tag module to print the tags of an instance

willthames:
##### Issue Type:

Feature Pull Request
##### Ansible Version:

ansible 1.6 (devel 435104065e) last updated 2014/03/04 11:07:56 (GMT +1000)
##### Environment:

N/A
##### Summary:

Use the print argument to state to just collect the tags of a resource through the AWS API. This allows the extraction of tags information without connecting to the resource itself.

The intention is then to use this information in further tasks.

### DIFF
--- a/library/cloud/ec2_tag
+++ b/library/cloud/ec2_tag
@@ -19,7 +19,7 @@ DOCUMENTATION = '''
 module: ec2_tag 
 short_description: create and remove tag(s) to ec2 resources.
 description:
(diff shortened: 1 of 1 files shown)


mpdehaan:
Very reasonable.

We don't really have a greatly established convention here, but perhaps, "list" instead of "print" as it's not actually printing them.

We could then adopt that convention for other modules that just wished to interrogate data.

willthames:
Sounds reasonable. I'll tweak it to that effect shortly.

willthames:
@mpdehaan fixed as suggested.

mpdehaan:
merged, thank you!

PR #6264, opened 2014-03-04
Add new cloud/rds DB instance types (db.m3 and db.cr1 families)

sqs:
This PR adds the new AWS RDS DB instance types (the db.m3 and db.cr1 families). The list now matches that on http://aws.amazon.com/rds/details/.

These new DB instance types were relesed on 2014 Feb 20: http://aws.amazon.com/about-aws/whats-new/2014/02/20/amazon-rds-now-offers-new-faster-and-cheaper-db-instances/.

Incidentally, it would be nice if this list were not hardcoded. Would you accept a PR to that effect?

### DIFF
--- a/library/cloud/rds
+++ b/library/cloud/rds
@@ -60,7 +60,7 @@ options:
     required: false
     default: null
     aliases: []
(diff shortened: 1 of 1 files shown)


mpdehaan:
Merged, thank you!

PR #6265, opened 2014-03-04
Add credential parameters to the GCE modules.

### LINKED ISSUE #5583: gce modules need a better way to handle invalid key files
If the user specifies an unsupported Service Account private key, the cloud/gce\* modules do not catch the error and produce a traceback like:

failed: [127.0.0.1] => {"failed": true, "parsed": false}
  File "/root/.ansible/tmp/ansible-1389289770.6-124236068572659/gce", line 1446, in <module>
    main()
  File "/root/.ansible/tmp/ansible-1389289770.6-124236068572659/gce", line 376, in main
    module.fail_json(msg=unexpected_error_msg(e), changed=False)
  File "/root/.ansible/tmp/ansible-1389289770.6-124236068572659/gce", line 190, in unexpected_error_msg
    error.http_code, error.code, str(error.value))
AttributeError: 'exceptions.ValueError' object has no attribute 'http_code'

@erjohnso will work on a fix and submit a pull-request.

fcuny:
In order to simplify the workflow with the GCE modules, it's now
possible to add the parameters and project name as arguments to the
various GCE modules.

The inventory plugin also returns the IP of the host in
`ansible_ssh_host` so that you don't have to specify IPs into the
inventory file.

Some update to the documentation are also added.

Closes #5583.

### DIFF
--- a/library/cloud/gce_lb
+++ b/library/cloud/gce_lb
@@ -110,6 +110,27 @@ options:
     default: "present"
     choices: ["active", "present", "absent", "deleted"]
     aliases: []
+  service_account_email:
+    version_added: 1.5.1
+    description:
+      - service account email
+    required: false
+    default: null
+    aliases: []
+  pem_file:
+    version_added: 1.5.1
+    description:
+      - path to the pem file associated with the service account email
+    required: false
+    default: null
+    aliases: []
+  project_id:
+    version_added: 1.5.1
+    description:
+      - your GCE project ID
+    required: false
+    default: null
+    aliases: []
 
 requirements: [ "libcloud" ]
 author: Eric Johnson <erjohnso@google.com>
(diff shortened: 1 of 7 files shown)


fcuny:
Sorry for the previous mess, @jctanner and @erjohnso.

library/cloud/gce_lb
```diff
@@ -130,40 +151,21 @@ EXAMPLES = '''
 import sys
 
 USER_AGENT_PRODUCT="Ansible-gce_lb"
```
erjohnso: Since you moved these USER_AGENT_\* variables off to `module_utils.gce`, I think that's sufficient and we don't need to override them in each GCE sub-module. If you don't mind, how about removing these lines and remove them (if they exist) in the other modules?

plugins/inventory/gce.py
```diff
@@ -73,7 +73,7 @@
 '''
 
 USER_AGENT_PRODUCT="Ansible-gce_inventory_plugin"
```
erjohnso, line 75: You can remove these USER_AGENT_ variables too if you don't mind.

erjohnso:
@franckcuny - this is really great stuff, thanks for making this MUCH better! I didn't have any big feedback but I did have a little request to remove the per-module `USER_AGENT_` variables. I think it's sufficient to have them defined in the `module_utils.gce`.

@jctanner - this looks good to me.  I also did some basic testing for adding/deleting instances and disks.

fcuny:
I actually have an issue with the inventory plugin. If you can hold on, I'll submit an update tonight.

plugins/inventory/gce.py
fcuny, line 75: I'm not going to remove this one since the inventory plugin doesn't use the authentication method like the other modules.
I'll try to fix that in a future PR.

fcuny:
Thanks @erjohnso for reviewing!

@jctanner it's good to go for me.

plugins/inventory/gce.py
erjohnso, line 75: Sure, that sounds great.

erjohnso:
Thank you @franckcuny, this was really great to see!

If you're so inclined, drop me a note at erjohnso at google dot com and I can see about sending you a t-shirt or something!

jctanner:
@franckcuny thanks!

PR #6266, opened 2014-03-04
Synchronize .deb package version with VERSION file

drybjed:
`make deb` at the moment builds `ansible_1.5_all.deb` package which cannot be automatically installed using information from `VERSION` file ("1.6"). This patch fixes that issue.


mpdehaan:
merged, thank you!

PR #6271, opened 2014-03-04
Removes pluralization of manager tag attribute.

jimkytola:
Fix to docker_image get_images method for Docker-py version >= 0.3 (Docker API >= 1.8)


jctanner:
@jimkytola thanks!

PR #6275, opened 2014-03-04
Fix wrong module name for exception in nova compute

mscherer:
failed: [127.0.0.1] => {"failed": true, "parsed": false}
    File "/tmp/ansible-tmp-1393950384.39-102240090845592/nova_compute", line 1328, in <module>
      main()
    File "/tmp/ansible-tmp-1393950384.39-102240090845592/nova_compute", line 241, in main
      except exc.Unauthorized, e:
  NameError: global name 'exc' is not defined

### DIFF
--- a/library/cloud/nova_compute
+++ b/library/cloud/nova_compute
@@ -238,9 +238,9 @@ def main():
                               service_type='compute')
     try:
         nova.authenticate()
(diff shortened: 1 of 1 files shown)


mpdehaan:
merged, thanks!   This error path does not seem well tested.   Much appreciated.

Issue #1206, opened 2012-10-02
Change in gather_facts not working in all cases


dagwieers:
The change from #1201 does not work in all cases because the test to see if the setup module was already run (once) is making incorrect assumptions which fail in our use-case.

What we do for provisioning is summarized as:
- Disable fact_gathering while 'locally' running:
  - Running network_facts (from DNS/yaml)
  - Running ilo or esx facts (from iLO device/ESX server)
  - Create boot-media using the above facts
  - Boot the physical/virtual system using this boot-media
- Wait for the system to come online in post-install mode
- Start the actual provisioning

This fails for the current implementation because when we start the actual provisioning, the setup module is not run for the first time because our other facts-modules already added facts to the **SETUP_CACHE**. What I would propose is that each facts-module in fact adds it's own fact to track whether it has already run. This is useful even outside the core usage, because I currently already check whether one of the facts-modules has run by testing one of the facts the module provides. Of course it would be better to not test a generic fact that we "know" has been set, but rather make this part of the standard.

So I propose every facts-module (setup included) to set the:

```
module_name: true
```

So in our case, I would have the following set:

```
module_ilo: true
module_esx: true
module_network: true
```

and the setup module would check if **module_setup** is set in order to decide whether it should run.

dagwieers:
Including @dhozac into the discussion

PR #6278, opened 2014-03-04
Populate module_setup from the setup module rather than special code elsewhere

dagwieers:
This small change allows for individual setup actions to populate the SETUP_CACHE and not cause a subsequent facts-gathering when not needed. This follows the standard of other facts modules as laid out in #1206 and implemented in fedfd187749654105a22be20c27e0050bc722d0a. It allows to test of the setup module has already been run even when gather_facts was explicitely disabled.


mpdehaan:
merged, thanks!

PR #6281, opened 2014-03-04
Fix broken Ansible Tower link.

alimony:
This will fix a broken link to Ansible Tower link; it was missing an initial colon. Also, removes some trailing whitespace, which I put in a separate commit.


mpdehaan:
Thanks!  This will appear the next time we push docs, which should be soon!

(The whitespace commit can probably be left last time, it breaks attribution, less of an issue here, but more so in the sources)

Issue #5146, opened 2013-12-03
ValueError: I/O operation on closed file


rajasaur:
This has been reported in #3518 by @hngkr. 

When running a playbook through celery, the fix for #3518 results in a ValueError: I/O Operation on closed file. 

```
fatal: [centos] => Traceback (most recent call last):
  File "/sources/projects/virtualenvs/customdeploy/ansible/lib/ansible/runner/__init__.py", line 390, in _executor
    self._new_stdin = os.fdopen(os.dup(sys.stdin.fileno()))
ValueError: I/O operation on closed file
```

If I comment out the fix related to #3518, all works fine(through celery task as well as command line).  

Copying the test playbook referenced in the other issue:

```

---
# file: site1.yml
- hosts: 127.0.0.1
  connection: local
  gather_facts: False
  tasks:
  - copy: src="/tmp/blarg" dest="/tmp/blob"
```

mpdehaan:
This appears unrelated to the linked ticket.

mpdehaan:
edit: ah, I see you are saying this line of code is causing the problem.   Interesting, we don't get this when people are running Ansible through Jenkins.   We can take a look.

jctanner:
@rajasaur how are you calling ansible-playbook from celery? Is it importing the ansible libraries or is it running shell commands?

jctanner:
I'm not able to reproduce the error by calling ansible-playbook with subprocess.Popen from a celery task.

hngkr:
I'm seeing the error when I'm creating an ansible.playbook.PlayBook and
running it - I looked at the code in ansible-playbook and mimicked that in
a Celery task.

When that playbook contains copy tasks, then I'm hitting this error (unless
I uncomment the reload(sys) function call). It's just the copy
module/command that fails - everything else executes as expected.

2014/1/2 jctanner notifications@github.com

jctanner:
@hngkr I'm going to need more information, such as an example of your celery task. I just replicated ansible-playbook as a celery task and did not see anything related to the error you and @rajasaur have mentioned.

This is my testcase: https://gist.github.com/jctanner/8229044

rajasaur:
@jctanner , I needed to do it by importing the ansible playbook module and invoking run on it. Here is how I did it.

```
                       pb = ansible.playbook.PlayBook(
                                    playbook=playbook_file,
                                    inventory=inventory,
                                    callbacks=playbook_cb,
                                    stats=stats,
[...]

PR #6283, opened 2014-03-04
fix missing stdin in _parallel_runner

### LINKED ISSUE #5617: ValueError: I/O operation on closed file (Multiple hosts)
Very much like this issue: https://github.com/ansible/ansible/issues/5146

i applied the patch from this issue and it work when ansible has only one host to deal with, but i still get the problem on multiple hosts.

i use an external inventory which has 4 hosts, and pattern= to narrow it down.
When i say that it's working on one host, i mean that the pattern match only one host.

```
import ansible.runner
runner = ansible.runner.Runner(
   module_name  = "copy",
   module_args  = "src=/tmp/ansibletest dest=/tmp/ansibletest",
   pattern      = "myhost0*"
)
results = runner.run()
```

```
    results = runner.run()
  File "/usr/lib/python2.7/site-packages/ansible/runner/__init__.py", line 989, in run
    results = self._parallel_exec(hosts)
  File "/usr/lib/python2.7/site-packages/ansible/runner/__init__.py", line 902, in _parallel_exec
    new_stdin = os.fdopen(os.dup(sys.stdin.fileno()))
ValueError: I/O operation on closed file
```

Thank you very much

magicrobotmonkey:
Fixes #5617

### DIFF
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -1075,9 +1075,17 @@ class Runner(object):
             job_queue.put(host)
         result_queue = manager.Queue()
 
+        try:
+            fileno = sys.stdin.fileno()
+        except ValueError:
+            fileno = None
+
         workers = []
         for i in range(self.forks):
-            new_stdin = os.fdopen(os.dup(sys.stdin.fileno()))
+            if fileno is not None:
+                new_stdin = os.fdopen(os.dup(fileno))
+            else:
+                new_stdin = None
             prc = multiprocessing.Process(target=_executor_hook,
                 args=(job_queue, result_queue, new_stdin))
             prc.start()


risaacson:
@magicrobotmonkey I am trying to test this so that it can be merged. I am unable to reproduce the problem at this point. I am using the script in the below gist based off of the API documentation and I am not having the same problem.

https://gist.github.com/risaacson/7290d30a612e0c70ea1b

Can you provide more information on what you did to duplicate this issue so that it is documented and that I can confirm and merge if it makes sense?

magicrobotmonkey:
@risaacson Sorry for the delay, I had a long weekend... 
Anyways, an easy way to recreate it is to add sys.stdin.close() to the top of your gist. Now when running with "hosts: test-*" it will work, #5146 fixed that. In order to replicate the multiple hosts version of it, you need to run "hosts: test-1:test2" or something similar. That's the case I'm addressing here. 

Thanks, Aaron

PR #6284, opened 2014-03-04
Automatically install apt_repository dependencies

omgjlk:
This follows the example of (and borrows code from) the apt module,
which will automatically install python-apt. This will make working with
repositories a bit easier just as the apt module is now easier.

### DIFF
--- a/library/packaging/apt_repository
+++ b/library/packaging/apt_repository
@@ -366,10 +366,21 @@ def main():
     )
 
     if not HAVE_PYTHON_APT:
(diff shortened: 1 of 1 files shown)


mpdehaan:
Thanks for this -- we may _possibly_ have a dup of this, I can't recall, but we should include one or the other regardless :)

omgjlk:
Ah, I forgot to search. There is #6220 , but that only covers python-apt, not python-pycurl. #6120 is relevant, and would have impact on my PR, and probably by #6220 doesn't automatically install it.

So with that searching, I think #6220 is the one to take, rather than this one. I'll close this one down.

PR #5837, opened 2014-01-31
Test hosts defined in multiple groups in different files

g-k-r:
The test now works identical to calling ansible directly as mentioned in #5749

Issue #5749, opened 2014-01-24
hosts defined in two groups in different inventory files each is used multiple times


g-k-r:
when using multiple inventory files defining different groups for the same hosts, ansible adds the hosts multiple times to the inventory list.

version: 1.4.4+dfsg-1 (from debian jessie)
same problem with current devel 00b3f627ebb2b8e9cf743671a9ddd2d8d8c9fe45

reproduce: 
- use files from https://gist.github.com/g-k-r/8599442
- run **ansible all --list**
- run **ansible group2 --list** and compare with **ansible type2 --list**

actual output:

``` shell
$ ansible all --list
    www.google.com
    www.yahoo.com
    www.yahoo.com
$ ansible group2 --list
    www.yahoo.com
$ ansible type2 --list
    www.yahoo.com
    www.yahoo.com
```

note when the group is defined in both files like [type1] this does not happen.

``` shell
$ ansible type1 --list
    www.google.com
```

expected output: www.yahoo.com shows up once only i.e.

``` shell
$ ansible all --list
    www.google.com
    www.yahoo.com
$ ansible group2 --list
    www.yahoo.com
$ ansible type2 --list
    www.yahoo.com
```

additional information:

to verify that tasks would actually be executed multiple times use

``` shell
$ ansible all -m debug -a '"msg=TEST {{inventory_hostname}}"'
www.google.com | success >> {
    "msg": "TEST www.google.com"
}

www.yahoo.com | success >> {
    "msg": "TEST www.yahoo.com"
}

www.yahoo.com | success >> {
    "msg": "TEST www.yahoo.com"
}
```

sivel:
Per contributing.md can you please provide your ansible version?  (`ansible --version`)

If I remember correctly, this has been resolved in devel (1.5)

jctanner:
Yes, this was resolved by https://github.com/ansible/ansible/commit/559e89036b98127510794125beeaeebc82e0c141

Please rebase and try again if you are running an older version.

g-k-r:
@sivel: amended the report

jctanner:
@g-k-r this is resolved in devel, and will be in the 1.5 release.

g-k-r:
@jctanner: tested with devel, not solved, will provide unit test in PR based on devel

mpdehaan:
Reopening for testing.

jctanner:
@g-k-r unable to reproduce ...

```
[...]

PR #6287, opened 2014-03-05
Fix inventory for test_dir_inventory

risaacson:
It came up that fixing this unit test may relate to another ticket that is open. This work allows us to uncomment this unit test by fixing how we pars variables allowing a quoted variable to contain a '#'.

This relates directly to GH-5837 and GH-5749.

Work also went into cleaning up some of the test data to clarify what was working.

Lastly work went into cleaning up formatting so that the code is easily read.

### DIFF
--- a/lib/ansible/inventory/ini.py
+++ b/lib/ansible/inventory/ini.py
@@ -23,6 +23,7 @@ from ansible.inventory.group import Group
 from ansible.inventory.expand_hosts import detect_range
 from ansible.inventory.expand_hosts import expand_hostname_range
 from ansible import errors
+import ansible.utils as utils
 import shlex
 import re
 import ast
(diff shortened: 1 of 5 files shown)


mpdehaan:
Per IRC discussion, we can (edit: should) probably make this shorter.

I think this could be accomplished by a regex, per line, that captures anything before a # that is not escaped, or at least, replaces all unescaped # with a different comment marker like ##HEYHEYIMACOMMENT## and then splits on _that_ string instead.

(Also anything of this complexity would usually warrant creation of some private functions)

mpdehaan:
Closing so this can be shortened up a bit.

risaacson:
Reopened.

@mpdehaan I have spent a bit of time digging into the regex path and ran into a problem with the negating look-behind which does not have support in python for variable width. To continue with the regex path without this feature makes it just as complex as the steps that I had implemented. Unfortunately I have not found a workaround via searching or questions in #python I did take your advice and move the comment stripping code off into a method which is under utils.

This PR merges in everything outstanding from GH-5749 which needed to be cherry-picked.

Passes all make tests.

PR #6288, opened 2014-03-05
Document that lookups are evaluated on server side

techtonik:


techtonik:
From here https://groups.google.com/forum/#!topic/ansible-project/96XWmqlCtSc

mpdehaan:
While this is true, "outside sources" can also mean web services.

I'd prefer another sentence be added here.

Lookup plugins allow access of data in Ansible from outside sources.   These lookups are evaluated on the control machine and can include file contents on the control machine, or even remote data services such as values in a datastore such as Redis.

Something like that?

techtonik:
Something is very wrong with my typing abilities today. PTAL.

mpdehaan:
I'll tweak this one.  Thanks for the heads up.

mpdehaan:
Done, thanks!

PR #6289, opened 2014-03-05
Make irc module accept the nick being shortened by the server.

rawtaz:
This can happen if the server has a NICKLEN setting which is less
than the length of the specified nick. With this patch we now grab
that nick and use it moving forward, instead of failing because we
didn't get back the one we specified, in the connection response.

### DIFF
--- a/library/notification/irc
+++ b/library/notification/irc
@@ -39,7 +39,7 @@ options:
     default: 6667
   nick:
     description:
-      - Nickname
+      - Nickname. May be shortened, depending on server's NICKLEN setting.
     required: false
     default: ansible
   msg:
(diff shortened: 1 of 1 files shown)


rawtaz:
I should add that I ran the regular tests (as per CONTRIBUTING.md) and also tested this change manually with Freenode, GIMPnet and IRCnet, all having different NICKLEN. I think it's fine.

The only regression I can think of right now is if someone specifies the nick to use in their play, and then also use that same nick somewhere else in a way such that they rely on it not being changed by the IRC server. That seems like a rather special and unlikely case though.

mpdehaan:
merged, thanks!

PR #6291, opened 2014-03-05
Make user module create accounts with temporary passwords

kalininsn:
This Pull Request adds new option passchangeonfirstlogin to user module, if it set to 'yes', after user creation procedure command 'chage -d 0' is being called, so after first login user should change his password

### DIFF
--- a/library/system/user
+++ b/library/system/user
@@ -393,6 +393,7 @@ class User(object):
         cmd.append(self.name)
         return self.execute_command(cmd)
 
+
     def modify_user_usermod(self):
         cmd = [self.module.get_bin_path('usermod', True)]
         info = self.user_info()
--- a/library/system/user
+++ b/library/system/user
@@ -86,13 +86,28 @@ options:
         description:
             - Whether the account should exist.  When C(absent), removes
               the user account.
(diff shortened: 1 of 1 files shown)


mpdehaan:
Thanks for this.

There are a few things with this pull request that should be changed.

The parameter should not be allruntogetherlikethis, but should be something like "change_on_first_login", or it seems even better to just expose the various aspects of chage  directly, allowing more flexibility, which may also warrant additional parameters.

This is a little gross in that it would require most of the parameters to chage, such as "chage_min", "chage_max", etc.   However I think once added, people will soon want the other knobs and buttons to be exposed.

There's a stray print that makes the return of the module not completely JSON, that may work if it appears after the JSON, but must be removed.

Finally, since this is a new option, in the DOCUMENTATION block under the parameter, add "version_added: '1.6'".

I do note that it doesn't detect when the password age has been changed with the "change=True", which should probably

kalininsn:
Thanks a lot for feedback, i'll try to fix it all tomorrow

kalininsn:
Here are some fixes for PR

mpdehaan:
Hi!

Thank you very much!

PR #6292, opened 2014-03-05
Add "checksum" option to 'synchronize' module

amandolo:
Activate the "checksum" option, already provided by rsync.

### DIFF
--- a/library/files/synchronize
+++ b/library/files/synchronize
@@ -53,10 +53,11 @@ options:
     required: false
   checksum:
     description:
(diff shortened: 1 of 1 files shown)


mpdehaan:
This looks pretty reasonable, though I'd like the "may not work as well as you would like" comment clarified, there's a small typo up there too.

Also be sure to add the "version_added" attribute to the DOCUMENTATION blurb so it's clear will only be available in 1.6.

Thanks!

amandolo:
I fixed the documentation blurb as you said.

Happy to help you!
Bye!

mpdehaan:
merged, thank you!

Issue #4454, opened 2013-10-11
mysql_user can't handle all syntax features of .my.cnf


jhagg:
When mysql_user tries to read my ~/.my.cnf it chokes on two things:
1. My TAB-indentation
   I like pretty files. :-)
   But I can live without it.
2. Flags without values
   Like 'silent'.
   This I can't live whithout. :-)

Example: .my.cnf

[client]
    password=XXXXXXXXXX
    default-character-set=utf8

[mysql]
    silent

I suspect that the real problem here is the ConfigParser class so
fixing this might be difficult.

A simple workaround would be to be able set the filename instead of
forcing it to ~/.my.cnf.
That could even make this module more flexible. :-)

jctanner:
The cnf file format as you have seen is not quite compliant with the ini spec that configparser wants. 

https://github.com/ansible/ansible/commit/3c4dd618e4a41a116a1043249077c6ca377eeddb is a workaround. Long term, the mysql connection parameter "read_default_file" should be used instead.

jctanner:
Test playbook:

<pre>
- hosts: localhost
  connection: local
  gather_facts: False
  vars:
     DB_DNS : localhost
     DB_USER : "root"
     DB_USER_PASSWORD : "redhat"
  tasks:
    - name: check connection with user+pass
      mysql_user: login_user={{ DB_USER }} 
                  login_password={{ DB_USER_PASSWORD }}
                  name={{ DB_USER }} 
                  password={{ DB_USER_PASSWORD }} 
                  host=% 
                  priv=*.*:ALL 
                  state=present 
                  login_host={{ DB_DNS }}
      ignore_errors: yes
    - name: create a good cnf file
      shell: echo -e "[client]\nuser=root\npassword=redhat" > ~/.my.cnf
      ignore_errors: yes
    - name: check connection with no user ... good cnf
      mysql_user: login_user={{ DB_USER }} 
                  name={{ DB_USER }} 
                  password={{ DB_USER_PASSWORD }} 
                  host=% 
                  priv=*.*:ALL 
                  state=present 
                  login_host={{ DB_DNS }}
      ignore_errors: yes
    - name: check connection with no user or password ... good cnf
      mysql_user: name={{ DB_USER }} 
                  password={{ DB_USER_PASSWORD }} 
                  host=% 
                  priv=*.*:ALL 
                  state=present 
                  login_host={{ DB_DNS }}
      ignore_errors: yes
    - name: create a bad cnf file
[...]

PR #6295, opened 2014-03-05
Use MySQLdbs read_default_file

muffl0n:
Removed custom config parsing and some unused code. Let's you specify a config_file that should be read instead of ~/.my.cnf

These changes are quite extensive, it will break config-parsing where non-mysql-compliant formats where used.

### DIFF
--- a/lib/ansible/module_utils/mysql.py
+++ b/lib/ansible/module_utils/mysql.py
@@ -1,0 +1,23 @@
+def connect(module, login_user, login_password, config_file):
+    default_file = '~/.my.cnf'
+    if config_file is not None:
+        default_file = config_file
+    
+    config = {
+        'host': module.params['login_host'],
+        'db': 'mysql'
+    }
+
+    if module.params['login_unix_socket']:
+        config['unix_socket'] = module.params['login_unix_socket']
+    else:
+        config['port'] = int(module.params['login_port'])
+
+    if os.path.exists(default_file):
+        config['read_default_file'] = default_file
+    else:
+        config['user'] = login_user
+        config['passwd'] = login_password
+        
+    db_connection = MySQLdb.connect(**config)
+    return db_connection.cursor()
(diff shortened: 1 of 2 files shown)


mpdehaan:
Much cleaner.

Can you elaborate on that last part?

muffl0n:
In the current version there is some custom handling for config files that can't be parsed with ConfigParser because it doesn't support key-only-entrys:

```
[mysql]
silent
```

This was added in #4454. It was also added that a config file could have this format:

```
user=foo
pass=bar
```

instead of mysql's format

```
[client]
user=foo
password=bar
```

Surely this will no longer work as "pass" is not supported in [mysql_options](http://dev.mysql.com/doc/refman/5.5/en/mysql-options.html)

Despite the fact that letting MySQLdb do all the parsing work I don't think ansible should need to support these crippled config files. Either the config file has the correct format or ansible should not accept it.

muffl0n:
@mpdehaan: Is there anything more I can do to get this merged?

muffl0n:
Could I please get a yes/no on this? I'm thinking about fixing the other mysql-modules as well. But first I'd like to know if these fixes are wanted at all.

tyll:
I cannot do anything to get this merged, but I hope it will. However I would like to suggest to use a dictionary to prepare the keyword arguments for the mysql connection and call it with e.g. `**mysql_args` to reduce duplicate code.

muffl0n:
You are absolutely right, thanks for the hint! I will work on this once a decision about accepting this PR is made.

muffl0n:
I was in the right mood and gave it a quick try. Unfortunately, I couldn't test it but I hope I'm not completely wrong. Gonna check it on Monday at the office.

muffl0n:
Fixed. Thanks!

muffl0n:
I just moved connect() to ansible.module_utils.mysql to open the option for other mysql-modules to use the same functionality.

mpdehaan:
Hi!

Thank you very much!

PR #6300, opened 2014-03-05
Update debian install documentation to include dependency resolution

l0neranger:
Just make the install process a tad bit easier.


mpdehaan:
Merged, thanks!

PR #6301, opened 2014-03-05
Pagerduty improvements

suprememoocow:
This PR consists of two simple improvements to the pagerduty monitoring module:
- Added support for using tokens instead of usernames and passwords.
- Added support for setting up maintenance windows of under an hour by allowing users to specify a `minutes` value.


mpdehaan:
Thanks!  Can you add the "version_added: '1.6'" attribute to each of the new fields?

suprememoocow:
Ah! Missed that. Thanks for the quick response. I've added the fields and also put some examples in that use the new fields.

mpdehaan:
Hi sorry for the trouble here and delay in getting back to you -- there has been a bit of a backlog to power through! ... can you also look at rebasing this commit?   We've since introduced module.fetch_url which we want to use, though this may require some new testing.

We'd do this ourselves but don't have a pagerduty account handy

Hopefully this change will be small.

Thanks!

suprememoocow:
Hi, thanks for getting back to me and sorry for the delay in implementing your request! I've tested this and it's working nicely with the fetch_url. 

Thanks for a great product!
Andrew

PR #6302, opened 2014-03-06
Fix respecting remote_tmp when su is used

aresch:
It looks like there is a typo in the evaluation of su_user for the purposes of obeying remote_tmp or not.

Also, su_user is always being set to None in Task.**init** since su is None.  I've removed the check and have it just set sudo/su_user every time but I am not sure if this is the correct fix.

### DIFF
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -415,7 +415,7 @@ class Runner(object):
 
         environment_string = self._compute_environment_string(inject)
 
-        if tmp.find("tmp") != -1 and (self.sudo or self.su) and (self.sudo_user != 'root' or self.su_user != 'root'):
+        if tmp.find("tmp") != -1 and (self.sudo and self.sudo_user != 'root') or (self.su and self.su_user != 'root'):
             # deal with possible umask issues once sudo'ed to other user
             cmd_chmod = "chmod a+r %s" % remote_module_path
             self._low_level_exec_command(conn, cmd_chmod, tmp, sudoable=False)
@@ -444,7 +444,7 @@ class Runner(object):
             else:
                 argsfile = self._transfer_str(conn, tmp, 'arguments', args)
 
-            if (self.sudo or self.su) and (self.sudo_user != 'root' or self.su_user != 'root'):
+            if (self.sudo and self.sudo_user != 'root') or (self.su and self.su_user != 'root'):
                 # deal with possible umask issues once sudo'ed to other user
                 cmd_args_chmod = "chmod a+r %s" % argsfile
                 self._low_level_exec_command(conn, cmd_args_chmod, tmp, sudoable=False)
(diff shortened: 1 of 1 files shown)


mpdehaan:
The first part definitely looks correct.

The second part we'll need to test.  I'm flagging this as a priority item so we can look at it soon and get in the next dot release.

cc @angstwad

Thanks!

aresch:
I may have made some other changes in my branch or added su_user to my ansible.cfg which may also have contributed to the fix.  I can take a look tomorrow morning and make sure this is all that is necessary to fix the bug.

angstwad:
I'll look at this first thing in the morning. Thanks!

aresch:
I checked my local branch and configuration, and these two commits do appear to fix the problem on their own.

angstwad:
@aresch @mpdehaan So the first commit definitely looks correct. 

But, on commit 8de412e, I'm not sure what you're solving for.  This appears to be function normally when I use the following playbook as a test.  Also, the remote_tmp appears to be set appropriately on both su and sudo.

```
---
- name: Test su connecting as current user
  hosts: etcd
  gather_facts: no
  tasks:
    - name: "No sudo or su"
      command: /usr/bin/whoami

    - name: "su to root"
      command: /usr/bin/whoami
      su: yes
      su_pass: password

    - name: "su to testuser"
      command: /usr/bin/whoami
      su: yes
      su_user: testuser
      su_pass: password

    - name: "sudo to root"
      command: /usr/bin/whoami
      sudo: yes
      sudo_pass: password

    - name: "sudo to testuser"
      command: /usr/bin/whoami
      sudo: yes
      sudo_user: testuser
      sudo_pass: password
```

angstwad:
@aresch Yeah, your second commit definitely looks like a no-op to me -- I'm not sure we should change this area at all.  Perhaps I'm missing something here, so can you please provide more information about what issues that you were seeing that the second commit fixes?

aresch:
When only sudo is yes, then su_user is set to None.

This causes a problem here:

```
if (self.sudo or self.su) and (self.sudo_user != 'root' or self.su_user != 'root') and basetmp.startswith('$HOME'):
             basetmp = os.path.join('/tmp', basefile)
```

Specifically:

```
(self.sudo_user != 'root' or self.su_user != 'root')
```

This evaluates to True when self.su_user is None and False when self.su_user is 'root'.

aresch:
Maybe the logic should be:

(sudo and sudo_user != 'root') or (su and su_user != 'root')

aresch:
Without the second patch, this playbook fails for me since my /tmp is noexec.

```
---
- name: testing
  hosts: all
  gather_facts: no

  tasks:
    - name: script
      script: foo.sh
      sudo: yes
```

```
PLAY [testing] **************************************************************** 

TASK: [script] **************************************************************** 
root
failed: [10.223.224.244] => {"changed": true, "rc": 126}
stdout: /bin/bash: /tmp/ansible-tmp-1394127986.33-116041719412878/foo.sh: Permission denied


           to retry, use: --limit @/home/aresch/test.retry

10.223.224.244             : ok=0    changed=0    unreachable=0    failed=1  
```

angstwad:
@aresch Hey, this last commit looks good now -- there were some logic errors there that I had overlooked.  Thanks!

@mpdehaan I think this is good.

jimi-c:
I ran through the logic on this, and in the old case there are two situations where permissions would be incorrect: when su is true and sudo_user != root, or when sudo is true and su_user != root. 

In queue for testing.

jimi-c:
Merged, thanks!

Issue #2211, opened 2013-02-27
Add on user module to lock account.


cocoy:
From mailing list. http://bit.ly/YySA29

mpdehaan:
I think we were saying

state=locked

and if the user didn't exist, locking it wouldn't do any damage.

state=unlocked

should unlock IF possible, but if not possible, that's ok, we wouldn't implement it.

state=present

if locked or unlocked, works as normal, no changes

rrtj3:
I'm planning on working on this. Just a bit busy at the moment. Might be a
week or two. If someone else wants to work on it before me though, please
go ahead. Just update here so we don't both work on it at the same time.

But yes, I was thinking adding additional states of locked and unlocked to
the user module would be the cleanest. We should be able to do this cleanly
by detecting if a account is locked or not with 'passwd -S <username>'. The
second field indicates if the account is locked, contains no password or
has a password set.

Romeo

mpdehaan:
Ok I'll take patches, closing the ticket to clean up the queue.  Thanks!

PR #6303, opened 2014-03-06
Add locked state for user module, beginnings of integration test

billwanjohi:
This branch adds implementations for locking user accounts across systems, but returns an error for all but Generic (Linux) systems, since I'm not able to test the others. Guided by #2211.

It also has basic tests for creating and deleting a user.

Lastly, there's a skipped test that would check for a locked password, but I'm unconvinced that that is necessary once you've expired the account.

### DIFF
--- a/library/system/user
+++ b/library/system/user
@@ -82,11 +82,12 @@ options:
     state:
         required: false
         default: "present"
-        choices: [ present, absent, locked ]
+        choices: [ present, absent, expired ]
         description:
-            - Whether the account should exist, and whether it is locked.
+            - Whether the account should exist, and whether it is expired.
               When C(absent), removes the user account.
-              Locked state is only implemented for Linux.
+              When C(expired), the user will not be able to login through any means.
+              Expired state is only implemented for Linux.
     createhome:
         required: false
         default: "yes"
@@ -314,7 +315,7 @@ class User(object):
             cmd.append('-s')
             cmd.append(self.shell)
 
-        if self.state == 'locked':
+        if self.state == 'expired':
             cmd.append('--expiredate')
             cmd.append('1')
 
@@ -418,7 +419,7 @@ class User(object):
             cmd.append('-s')
             cmd.append(self.shell)
 
-        if self.state == 'locked':
+        if self.state == 'expired':
             cmd.append('--expiredate')
             cmd.append('1')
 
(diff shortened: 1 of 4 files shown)


mpdehaan:
Thanks for this.

(Since this makes users on the system where the tests are run, this probably needs to be moved to the "destructive" integration test category.  Can you update this?)

I'll leave testing for others when we get to this in the queue.

Thanks again!

billwanjohi:
That makes sense, moved it to destructive.

quiver:
It looks to me what @billwanjohi  implementted here is setting expire date for user accounts(`$ useradd --expiredate`), rather than making it locked.
I think #2211 is to implement `$ usermod --lock/--unlocked` in ansible, doesn't it?

billwanjohi:
You are correct about what's implemented here. The problem with --lock is that it only affects the password. Specifically, it prepends an "!" in front of the user's stored password, thus breaking it. Any other type of access is permitted.

I think that that is inadequate and deceptive, and so the account should be expired.

quiver:
> The problem with --lock is that it only affects the password. 

That's right. Locked users still can log in with key  authentications.

> I think that that is inadequate and deceptive, and so the account should be expired

`lock` and `expire` are two different thing. `lock`-ed account may behave counterintuitive but this is lock's specification. If you want to lock an account, you can make the account locked state. If you want to expire an account, you can make the account expired state. Locking an account ends up with expiring it seems odd to me.

FYI, chef's user lock is `usermod --lock` internally.  http://docs.opscode.com/resource_user.html

Anyway, I have no objection to adding `expire` state to user module.

billwanjohi:
I am not opposed to the state being named expired.

mpdehaan:
Sounds like the argument is for adding "expired" in addition to lock, please make this so and I'll see about merging this.  

Meanwhile, we should also make very sure the description for each field makes this clear, so users don't think they have locked something completely when key access is still available.

billwanjohi:
OK, I renamed the state to expired, and added an explanatory sentence to the user state documentation. Tests still pass for me.

I did not implement the locked or unlocked state, because it seems either redundant or insufficient in almost all cases. But I did leave a (when: no) task in the user integration test, if someone wanted to add that later. If that were implemented, I would suggest it be a separate user option altogether, maybe "password_lock" with yes and no choices.

All that said, does this look good to merge?

mpdehaan:
Hi!

Thank you very much!

PR #6306, opened 2014-03-06
Make it easier to determine whether instance creation or tagging fails

willthames:
If there is an AWS permissions issue with instance creation or tagging, it can be difficult to see whether the issue is with instance creation or with tagging. This just clarifies that situation.


mpdehaan:
merged, thank you!  Much better error message!

PR #6308, opened 2014-03-06
Module "file": Fix link to "path" in documentation

muffl0n:
c(dest) -> c(path)

dest is just an alias to path so the linking fails.


mpdehaan:
merged, thank you!

PR #6313, opened 2014-03-06
Don't use keyword `error` in encode

phenomenes:
This is not supported in Python 2.6. Just use positional arguments.


phenomenes:
Fixes error on Python 2.6

mpdehaan:
Merged, thank you!

PR #6323, opened 2014-03-07
Fix rpm_key on system with no gpg keys imported

aresch:
Without the -a option, rpm command will fail (exit code 1) and execute_command() will fail causing an initial key import to not work.

[root@test ~]# rpm -q gpg-pubkey
package gpg-pubkey is not installed
[root@test ~]# echo $?
1
[root@test ~]# rpm -qa gpg-pubkey
[root@test ~]# echo $?
0


mpdehaan:
merged, thanks!

PR #6324, opened 2014-03-07
BUGFIX: route53 module won't round-trip * and @ in records

scottanderson42:
Boto/AWS return asterisks and at symbols as octal codes, which prevents the change detection of the route53 module from functioning correctly with strings like *.name.domain.com or @.

This bugfix replaces octal representations of such characters with the actual character.


mpdehaan:
merged, thanks!

PR #4375, opened 2013-10-04
copy: Implement recursive copying if src is a directory.

pfalcon:
If src param to copy is a directory, all files under it are collected
and pushed one by one to target. Source dir path handled in a way
simalar to rsync: if it ends with slash, only inside contents of
directory are copied to destination, otherwise the dir itself is
copied (with all contents of course). Original idea and implementation
by https://github.com/ansible/ansible/pull/1809 . Rewritten to address
review comments and simplify/correct logic.

PR #6325, opened 2014-03-07
Documentation: FAQ is wrong about copying files recursively

muffl0n:
The copy module supports recursive copying since #4375.


mpdehaan:
I think we can probably just remove this section now, or alternatively we should mention synchronize as well.  Leaving this open as a note that I need to do this.

mpdehaan:
Ok this is now done and we mention both of the above options, thanks for the heads up!

PR #6333, opened 2014-03-07
Merge pull_6033.

risaacson:
Closes GH-6033

Merges the ability to define a version of a rpm so that you can downgrade. Adds not running the extra code if a version is not defined.

Test playbook used at: https://gist.github.com/risaacson/99af626953ae2b76e2a0


cove:
@Rissacson Thank you!

PR #6335, opened 2014-03-07
Rename private_key_file to private_key on example ansible.cfg

jmonteiro:
Example configuration is wrong, it's private_key instead of private_key_file. This patch simply renames it.


mpdehaan:
Actually not the case.

See constants.py

DEFAULT_PRIVATE_KEY_FILE  = shell_expand_path(get_config(p, DEFAULTS, 'private_key_file', 'ANSIBLE_PRIVATE_KEY_FILE', None))

PR #6339, opened 2014-03-08
systemd template services must not discard template part

charles-dyfis-net:
Right now, trying to enable the service `dhcpcd@eth0` instead checks, and tries to start, `dhcpcd@`.

While the end of the service name needs to be trimmed for determining whether this service exists in systemd, it should not be discarded for any other purpose.


mpdehaan:
merged, thanks!

PR #6340, opened 2014-03-08
docker_image: add support for pulling an image from a registry

joshuaconner:
This allows the user to omit the `path` argument from the module with `state=present` and the module will attempt to pull the image from a Docker registry instead of building it from source. Even though the `docker` module will already attempt to pull an image when it needs it to run a container, being able to ensure an image already exists on a set of machines is helpful when instances on different hosts need to be created at roughly the same time.

I also changed the `build` option for the `state` module parameter to `force` to better match the semantics of what it would do when building or when pulling.

### DIFF
--- a/library/cloud/docker_image
+++ b/library/cloud/docker_image
@@ -259,7 +259,7 @@ def main():
             state           = dict(default='present', choices=['absent', 'present', 'build', 'force']),
             docker_url      = dict(default='unix://var/run/docker.sock'),
             timeout         = dict(default=600, type='int'),
-            source          = dict(default='local', choices=['local', 'repository'])
+            source          = dict(default='local', choices=['local', 'repository']),
             username        = dict(required=False, default=None),
             password        = dict(required=False, default=None),
         ),
--- a/library/cloud/docker_image
+++ b/library/cloud/docker_image
@@ -193,6 +193,14 @@ class DockerImageManager:
                 self.error_msg = chunk_json['error']
                 return None
             else:
+                # The stream from client.pull() is SUPER noisy.
+                #
+                # Rather than returning hundreds of lines where, for many, the
+                # only change from the previous line is a slight difference in
+                # an ASCII progress bar, we just keep a log of each line where
+                # either the id or status changed from the previous line. This
+                # should give the user a good idea of what's going on without
+                # overwhelming them with unnecessary information.
                 id = chunk_json['id']
                 status = chunk_json['status']
                 if status != last_status or id != last_id:
--- a/library/cloud/docker_image
+++ b/library/cloud/docker_image
@@ -31,19 +31,19 @@ description:
 options:
   path:
     description:
-       - Path to directory with Dockerfile
+       - Path to directory with Dockerfile, if building from source
     required: false
     default: null
     aliases: []
   name:
     description:
-       - Image name to work with
+       - Name to apply to the resulting image if building from source, or repository to pull from otherwise
     required: true
     default: null
     aliases: []
   tag:
     description:
-       - Image tag to work with
+       - Image tag to apply/pull
     required: false
     default: ""
     aliases: []
@@ -61,16 +61,28 @@ options:
     aliases: []
   state:
     description:
-      - Set the state of the image
+      - Set the state of the image ("force" will always re-pull/re-build image)
     required: false
     default: present
-    choices: [ "present", "absent", "build" ]
+    choices: [ "present", "absent", "force" ]
     aliases: []
   timeout:
     description:
       - Set image operation timeout
     required: false
     default: 600
+    aliases: []
+  username:
+    description:
+      - Registry username (if pulling from a private registry)
+    required: false
+    default: null
+    aliases: []
+  password:
+    description:
+      - Registry password (if pulling from a private registry)
+    required: false
+    default: null
     aliases: []
 requirements: [ "docker-py" ]
 '''
@@ -90,7 +102,7 @@ Build new version of image:
   sudo: yes
   tasks:
   - name: check or build image
-    docker_image: path="/path/to/build/dir" name="my/app" state=build
+    docker_image: path="/path/to/build/dir" name="my/app" state=force
 
 Remove image from local docker storage:
 
@@ -99,6 +111,14 @@ Remove image from local docker storage:
   tasks:
   - name: run tomcat servers
     docker_image: name="my/app" state=absent
+
+Ensure the coreos/etcd has been pulled from the public Docker registry
+
+- hosts: web
+  sudo: yes
+  tasks
+  - name: Ensure coreos/etcd image exists
+    docker_image: name=coreos/etcd
 
 '''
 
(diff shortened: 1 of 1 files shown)


joshuaconner:
(ping @softzilla)

softzilla:
Great!
I agree with state-options changes: "force"-state looks better.

mpdehaan:
This adds new parameters and does not add a "version_added" attribute, so we can't quite take this one as is.

Also we can't remove choices from existing parameters as this will break playbooks.  

If you would like to address these things we would like to have them, though it might be better to have an explicit mode flag about whether it is building or pulling rather than trying to figure it out?

joshuaconner:
`version_added` tags added, plus I put `build` back in as a possible `state`. 

Additionally, I added a `source` argument as a mode flag - it defaults to `local` for backwards-compatibility (in which case it will attempt to build an image), but you can set it to `repository` to pull from an index as well.

joshuaconner:
(Hold off on this for a bit - this PR will be broken until https://github.com/dotcloud/docker-py/issues/176 is fixed.)

ruudk:
Any news on this?

joshuaconner:
The upstream issue was fixed in docker-py, then Docker 0.9 had some API changes that broke it again. When https://github.com/dotcloud/docker-py/pull/210 gets merged in and a 0.3.2 version is released I'll update this PR. In the mean time, we've been using these two commands as a substitute/workaround: 

```
- name: Check if image has already been built using some fancy grep
  shell: if [ -z "$(docker images | grep '{{ docker_image }}[[:blank:]]*{{ docker_image_tag }}')" ]; then echo image not built; else echo image built; fi
  register: cmd_docker_images_grepped
  changed_when: cmd_docker_images_grepped.stdout != 'image built'

- name: Ensure Docker image has been built
  command: docker pull {{ docker_image }}:{{ docker_image_tag }}
  when: cmd_docker_images_grepped.changed
```

If you're not interested in a particular tag (though you should be!) something like this should work:

```
- name: Check if image has already been built using some fancy grep
  shell: if [ -z "$(docker images | grep '{{ docker_image }}')" ]; then echo image not built; else echo image built; fi
  register: cmd_docker_images_grepped
  changed_when: cmd_docker_images_grepped.stdout != 'image built'

- name: Ensure Docker image has been built
  command: docker pull {{ docker_image }}
  when: cmd_docker_images_grepped.changed
```

sivel:
I've been playing with this a little also.  Seems that stream=True causes the the iteration of the response to be a giant concatenated string of json hashes.

I wrote a function to try and handle that gracefully.

https://github.com/sivel/ansible/compare/docker-image-pull

That is a completely different set of work to add pull functionality, that I haven't played a lot with in the context of the other parameters that the module provides.

joshuaconner:
I had considered this approach as well (and that _is_ a clever hack) but my feeling was that docker-py will eventually fix this issue, and if Ansible had merged in changes to work around docker-py's broken-ness, Ansible itself would then be broken. The workaround I posted above seems simple enough that merging code into Ansible that could potentially break things for people in the future seems unnecessarily risky to me?

sivel:
The `concated_hashes_to_list` function is "smart" enough to handle just a single non concatenated hash as well, so it should be backwards and forwards compatible.  It just turns every hash into a list item, so a single non concatenated hash, would become a list with 1 element that could still be iterated over.

I think by docker v1 we should have a clearer picture, but right now, I would like to see this in the module.  I am maintaining a separate module for docker pull that is distributed with a role now.

joshuaconner:
Unless that hash contains the string "}{"?

sivel:
That is true.  I did have it as `"}{"` to make the matching a _little_ safer.  But yes, there is the potential for it being fragile if we are splitting on string matches.

fangpenlin:
Just tried to use this module, spotted some issues:

### email is missing for login

```
msg: Docker API error: Registration: "Wrong email format (it has to match \"[^@]+@[^@]+\\.[^@]+\")"
```

the installed docker-py version is 0.3.1, not pretty sure is that issue fixed in a newer version. However, as far as I know, it seems the email is a  required parameter for login the server (stupid design, but it acts like that)

### There should be an extra parameter for specifying docker image repository

I noticed that repository url is obtained like this:

``` python
        image = self.module.params.get('name')
        # ... skip ...
        namespace = image.split('/')[0]
```

then you login

``` python
        if username:
            self.client.login(username, password=self.module.params.get('password'), registry=namespace)
```

I don't understand design rationale behind this. Reference to [docker-py source code](https://github.com/dotcloud/docker-py/blob/master/docker/client.py#L589), the `registry` is the URL to docker image repo. Which means iI want to specify the repo URL, I need to set the image name to something like this?

```
index.docker.io/v1/foo/bar
```

okay, you got the `index.docker.io` as the registry, then what about the `/v1`? What if my own private repo is not mounted on root? And it looks like then `index.docker.io/v1/foo/bar` will be feed to `pull` method as the image name, I don't think that's a correct way to pull an image. IMHO, there should be an extra parameter dedicated for repp URL.

fangpenlin:
https://github.com/dotcloud/docker-py/blob/master/docker/client.py#L660

okay, I see, so it actually parse the given image name

``` python
    def pull(self, repository, tag=None, stream=False):
        if not tag:
            repository, tag = utils.parse_repository_tag(repository)
        registry, repo_name = auth.resolve_repository_name(repository)
```

but still, not sure this is working for the login part

library/cloud/docker_image
```diff
@@ -187,48 +256,72 @@ def main():
...
+                    action = ACTION_BUILD
+                else:
+                    action = ACTION_PULL
         # remove image or images
         elif state == "absent":
             manager.remove_images()
 
-        if do_build:
-            image_id = manager.build()
+        if action:
```
fangpenlin, line 293: `action` is not defined when `state` is `absent`

library/cloud/docker_image
```diff
@@ -187,48 +256,72 @@ def main():
...
-                msg = "Error: %s\nLog:%s" % (manager.error_msg, manager.get_log())
+                result = dict(
+                    failed      = True,
+                    changed     = manager.has_changed(),
+                    msg = "Error: %s" % manager.error_msg,
+                    log = manager.get_log(),
+                )
 
-        module.exit_json(failed=failed, changed=manager.has_changed(), msg=msg, image_id=image_id)
+        module.exit_json(**result)
```
fangpenlin, line 317: `result` is also not defined when `action` is false

joshuaconner:
Hi Victor! Apologies, this sort of fell off my radar and nowI'm pretty busy with some other stuff at the moment. If you want to put together your own PR based off of this, you have my blessing! But I'm going to close this for now as I don't have time to get it working reasonably soon.

fangpenlin:
@joshuaconner 

I was actually thinking about continue your work on this pull request. However, I just wondered why don't you guys simply create a new module `docker_pull`? For docker command, `build` and `pull` are different subcommands, won't it be odd to put these two functions into single command `image`?

fangpenlin:
I found there is a repo out there is for pulling image: https://github.com/modcloth-labs/ansible-docker-pull

joshuaconner:
The way I thought of it is that though the subcommands for Docker are different, the result – that an image is in a given state on the machine you're managing – is the same.

fangpenlin:
@joshuaconner yeah, but what about `file` and `get_url`? they are all have the same result in file, should they share the same module?

My thinking is we can have multiple means to get the same result, but it doesn't necessary mean you need to put all these functionalities into the same module. When having different modules leads to simpler module design and more obvious propose, I would definitely choose to have different modules. 

However, I am actually pretty new to ansible, I have no idea what's the design philosophy for modules here.  @mpdehaan thoughts?

joshuaconner:
(That is a good point! Best of luck, whichever approach you take.)

PR #6347, opened 2014-03-08
Update playbooks_variables.rst

dwheaton:
incorrect/confusing wording in Jinja2 note?


mpdehaan:
good catch, will appear in next docs push, thanks!

PR #6349, opened 2014-03-08
Adds ec2_instance_facts module

scottanderson42:
A module that can retrieve information about particular ec2 instances filtered and/or sorted by name, state, or tag.

### DIFF
--- a/library/cloud/ec2_instance_facts
+++ b/library/cloud/ec2_instance_facts
@@ -1,0 +1,211 @@
+#!/usr/bin/python
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: ec2_instance_facts
+version_added: "1.5"
+short_description: Return a list of ec2 instances, possibly filtered by tag
+description:
+     - Return a list of ec2 instances, possibly filtered by tag. This module has a dependency on python-boto >= 2.5
+options:
+  name:
+    description:
+      - Filter returned instances by this name.
+    required: false
+    default: null
+    aliases: []
+  instance_ids:
+    description:
+      - Filter returned instances by membership in a list of ids.
+    required: false
+    default: null
+    aliases: []
+  states:
+    description:
+      - Filter returned instances by instance state.
+    required: false
+    default: null
+    aliases: []
+  tags:
+    description:
+      - Filter returned instances by existence of the given tags.
+    required: false
+    default: null
+    aliases: []
+  ec2_url:
+    description:
+      - Url to use to connect to EC2 or your Eucalyptus cloud (by default the module will use EC2 endpoints).  Must be specified if region is not used. If not set then the value of the EC2_URL environment variable, if any, is used
+    required: false
+    default: null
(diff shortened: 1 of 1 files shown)


mpdehaan:
So I generally feel the inventory script for ec2 should be responsible for returning most information here.

If there are variables you find missing, could they be extended there instead?

Jinja operations like "with" should be sortable and so forth.

scottanderson42:
On Mar 8, 2014, at 7:29 PM, Michael DeHaan notifications@github.com wrote:

> So I generally feel the inventory script for ec2 should be responsible for returning most information here.
> 
> The inventory has a few issues here:

1) It doesn’t know about hosts that don’t exist. In order to work with the inventory script, I would have to a) have a local play to check if the host is available in the results of the script, and if not create it and inject it. I guess it’s not tremendously different from using a module, but it does feel clunkier to me.

2) The inventory script is a blunt tool. There’s no way to say “instances that have _both_ an environment tag of 'production' and a type tag of 'maint’”. I could add filtering to the command line args of the script, but at that point I have to build a bash script around the ansible play in order to set the arguments properly for the inventory script. In my use case, having this logic outside of the ansible script is the wrong place. I don’t want to have to edit two different scripts to modify behavior, nor do I want to maintain an external bash script for every play invocation.

> If there are variables you find missing, could they be extended there instead?
> 
> No doubt, but it’s more the filtering that I’m missing than any information.
> Jinja operations like "with" should be sortable and so forth.
> 
> Understood; I had actually added that functionality for another purpose (too involved to go in here) and I didn’t realize I could do that via with.

Regards,
-scott

mpdehaan:
Seems like we'd want to add filtering here to the inventory script, by an environment variable like AWS_FILTER_BY_TAG or something to cover your use case.  I'm still not sure we need this one.

scottanderson42:
I haven’t had time to reply to your other thoughtful email, but another consideration to keep in mind: if you have 100 servers, the inventory script will make over 100 AWS API calls to get the information. The facts module is capable of getting that information with a single call. This plays into API rate limiting.

Summing up:

From an architectural standpoint:

1) Dealing with resources (inventory) only via an external script is not a consistent design.
2) If instances are treated as inventory, RDS, ElastiCache, AMIs, everything should be treated as such.
3) If everything from #2 has to go through the inventory script, it will quickly become extraordinarily unwieldy, violating the Law of Demeter.
4) If some resources allow information gathering from within a playbook (the rds module, ec2_facts, rax_facts for example), then why not this?
5) Given #4, if the intent is that all inventory/resource facts be gathered externally, then you’ve already lost. :-) And if that is the intent, then this inventory information gathering should be a) supported and b) a lot more structured than the current “write a script to return whatever you want” implementation.
6) If everything has to go in the inventory script, the performance is going to be atrocious and API call intensive. I don’t need to figure out all of the AMIs (replying to the comment you just left on ec2_ami_facts :-), RDS instances, and ec2 instances if all I’m trying to do is manage my ElastiCaches.
7) Managing the incantation of the ec2.py script is going to involve external scripting (in bash or what not) once it gets complicated enough. I’m really not interested in having two separate scripts to do one thing.
8) If you’re going to suggest putting all inventory in a single inventory script, you’re going to want a module system to fight the unwieldy. You already have a module system, why not leverage that?

From an IT management standpoint:

1) I have to deal with concerns other than just “is this technically feasible”. I can’t allow my organization to rely on something that is unsupported example code, nor can I allow us to just internally maintain something as a result. The ec2 inventory script, as an unsupported example, isn’t terribly “batteries included” from a maintenance standpoint.

My last attempt. :-) If the answer is still no I’ll just resign myself to maintaining the module for our own use.

yoshiwaan:
http://docs.ansible.com/ansible/ec2_facts_module.html

For anyone passing through, this achieves essential the point of this case

scottanderson42:
@yoshiwaan No, it does not. :-) From ec2_facts:

"The module must be called from within the EC2 instance itself."

This use case is asking for this information from a machine that is NOT the EC2 instance.

bcoca:
ec2_remote_facts is what you want

scottanderson42:
@bcoca Well, for my own purposes I'll just keeping using the internal module that kicked off this whole thread. :-) I'm very glad that Ansible seems to be on the facts module train now, though.

PR #6355, opened 2014-03-09
Add a package to missing list if npm list reported it as invalid

hkariti:
When bumping a dependency's version in packge.json npm list will list it as "invalid" and not "missing", causing the npm module to not mark it for installation.


mpdehaan:
merged, thanks!

Issue #6353, opened 2014-03-09
Assemble module should put a newline between each file


pgehres:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.5 / devel @ 04a6dc6d12
##### Environment:

Unrelated but
Host: OS 10.9.1, 
Target: Centos 6.5
##### Summary:

The assemble module (as of 3b2d409) ensures that a newline is placed after any delimiter usage but the module does not put a newline between each of the files. Without putting a blank line in each file (which can easily get accidentally removed) the first line of each file is concatenated with the last line of the preceding file.  This breaks a lot of configuration files that are "assembled" (e.g. iptables in my case).
##### Expected Results:

I would expect a newline between fragments.
##### Actual Results:

The first line of each file is concatenated with the last line of the preceding file 
##### Steps To Reproduce:

Try the assemble module

mpdehaan:
I'm a bit confused by this one as if the delimiter always adds a newline, there should be a newline between assembled files.

Can you help explain further how to reproduce this by showing some file snippets, task lines, and output?

pgehres:
Sure. I'll work on that. To clarify, the delimiter is an optional parameter. Simply adding a " " as the delimiter doesn't work as python evaluates that to False (if delimeter: ... ).

pgehres:
Also, since you agree that assemble should put a newline between files, I'll submit a pull request as well.

pgehres:
Test playbooks at https://github.com/pgehres/ansible-playbooks-test

The output I get is:

```
*filter
:INPUT ACCEPT [0:0]
:FORWARD ACCEPT [0:0]
:OUTPUT ACCEPT [0:0]
-A INPUT -m state --state ESTABLISHED,RELATED -j ACCEPT
-A INPUT -p icmp -j ACCEPT
-A INPUT -i lo -j ACCEPT-A INPUT -m state --state NEW -m tcp -p tcp --dport 22 -j ACCEPT-A INPUT -m state --state NEW -m udp -p udp --dport 123 -j ACCEPT
-A INPUT -m state --state NEW -m tcp -p tcp --dport 4848 -j ACCEPT
-A INPUT -m state --state NEW -m tcp -p tcp --dport 5099 -j ACCEPT
-A INPUT -m state --state NEW -m tcp -p tcp --dport 5666 -j ACCEPT-A INPUT -j REJECT --reject-with icmp-host-prohibited
[...]

PR #6359, opened 2014-03-09
Fix for #6353 adding a newline between assembled files

pgehres:


mpdehaan:
This looks like this might result in extra newlines when the delimiter is set:

```
   if delimit_me and delimiter:
        # un-escape anything like newlines
        delimiter = delimiter.decode('unicode-escape')
        tmp.write(delimiter)
        # always make sure there's a newline after the 
        # delimiter, so lines don't run together
        if delimiter[-1] != '\n':
            tmp.write('\n')
    if os.path.isfile(fragment):
        tmp.write(file(fragment).read())
        # always put a newline between fragments
        tmp.write('\n')
```

mpdehaan:
This otherwise looks like it will fix the problem.   Can you add a "if not delimiter" in there before the extra \n write at the bottom?

pgehres:
So, the only way I see that happening is if delimiter is "\nDELIMITER" otherwise you just end up with the newline between the delimiter start and the previous fragment.

On a similar note, this would result in an extra "\n" at the end of the file. I could reuse "delimit_me" to add the "\n" between files at the top of the loop, similar to the way the delimiter works. Thoughts?

mpdehaan:
So I was just testing this as follows:

```
root@llamacube:~/ansible# ansible -i "127.0.0.1," 127.0.0.1 -m "assemble" -a "src=/tmp/inputs dest=/tmp/dest delimiter='---'" -c local
```

/tmp/inputs/bar
    bar
    bar
    bar 
    <EOF>

/tmp/inputs/baz

```
baz baz no newline<EOF>
```

/tmp/inputs/foo

```
foo new newline<EOF>
```

/tmp/inputs/glorp

```
 glorp
 glorp
 glorp
 <EOF>
```

And when I run this before your change, I get the following output:

```
bar
bar
bar with newline

---
baz baz no newline
---
foo no newline
---
glorp?
glorp?
glorp?
```

And when I run it with your change, I get what seems to be redundant newlines:

```
bar 
bar
bar with newline


---
baz baz no newline

---
foo no newline

---
glorp?
glorp?
glorp?
```

From the above, the original behavior seems to make more sense to me.   I also tested this without the delimiter on the development branch...

The devel branch HEAD produces:

```
bar
bar
bar with newline

baz baz no newline
foo no newline
glorp?
glorp?
glorp?
```

The version you have produces the following:

```
bar
bar
bar with newline


baz baz no newline

foo no newline

glorp?
glorp?
glorp?
```

Apologies on the crazy test examples.

From the above, it seems like the development branch head version does not result in extra newlines, but also that with the above patch applied the newlines are inserted too frequently.

pgehres:
Weird.  I'll test again against 1.5.3 and devel and see what I get. Thanks.

mpdehaan:
See my comment on the other ticket, there's a commit on March 6th that may have addressed this (to devel only, most likely).

If that works, we can slate for inclusion in the next dot release.

mpdehaan:
This fix does appear to exist on the 1.5.3 branch.

jimi-c:
Closing, as the other PRs above were merged instead (which included this work). Thanks!

PR #6361, opened 2014-03-09
Identify Arch Linux as 'Arch' for consistency with Python 3 platform module

charles-dyfis-net:
As of current Python 2.x, the platform module is not aware of Arch Linux. Appropriately, Ansible detects this distribution itself, mapping `/etc/arch-release` to a distribution identity of `Archlinux`.

In Python 3, the output of `platform.linux_distribution()` on Arch Linux is `('arch', 'Arch', 'Linux')`. Correspondingly, on this platform, the output of `lsb_release -s -i` (to get the distribution id in short form) is `Arch`.

For these reasons, the detection performed by the setup module should be normalized; otherwise, this will introduce an extra level of user-visible incompatibility (or need for mapping) on migration to Python 3.0.

### DIFF
--- a/library/system/setup
+++ b/library/system/setup
@@ -117,7 +117,7 @@ class Facts(object):
                     '/etc/system-release': 'OtherLinux',
                     '/etc/alpine-release': 'Alpine',
                     '/etc/release': 'Solaris',
-                    '/etc/arch-release': 'Archlinux',
+                    '/etc/arch-release': 'Arch',
                     '/etc/SuSE-release': 'SuSE',
                     '/etc/os-release': 'Debian' }
     SELINUX_MODE_DICT = { 1: 'enforcing', 0: 'permissive', -1: 'disabled' }


mpdehaan:
Ansible isn't going to be supporting Python 3.0 anytime soon, because so much of our userbase has 2.x as native -- including all Enterprise Linux variants, Ubuntu releases, etc.   Basically we look at Python 3 as a seperate language.

So, python 2.4 or higher is a requirement on all nodes.

The above change would also break any playbooks keying off that value.

Our users are not looking at Ansible as a "Python" application per se -- but a generic management one (many people are deploying Java, Scala, Perl, C, Rails, or other applications too), so as such the values coming back also need not align with what a python developer might expect.

Thus I don't think this is really going to cause any confusion.

If you'd like to continue this discussion further feel free to stop by the development list.

Thanks!

PR #6362, opened 2014-03-09
Modular fact finding

charles-dyfis-net:
At present, the Facts class (and its children) are only available within the setup module.

This is unfortunate, as the Facts class contains a substantial amount of generically useful code. As an example, the operating system detection implemented here is deep and robust, including a great deal of fallback logic for the case where the Python standard library's `platform` module is not up to task.

Correspondingly, there are presently cases where facts determined by the setup module have more accurate and useful information than is available to other modules, whether built into Ansible or otherwise.

This branch moves the Facts class and its children and helpers into common code, and makes portions individually invokable to avoid the expense of needing to determine the entire set of facts from any single user.


mpdehaan:
Thanks!

Note to team -- approved this idea previously.  Because of the size of the change and the likelihood of drift in the fact code as we merge other things, we should move the functions over manually, and then cherry-pick the other two.

mpdehaan:
I've pushed a similar commit, because I wanted to make sure any recent updates to the setup module (of which there were several) made it in.  I've kept the same filenames.

Thanks!

Issue #6365, opened 2014-03-10
Cloudformation module fails to update already existing stack


cloudify:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.5 
##### Environment:

Darwin 13.1.0 Darwin Kernel Version 13.1.0: Thu Jan 16 19:40:37 PST 2014; root:xnu-2422.90.20~2/RELEASE_X86_64 x86_64

Python 2.7.5

boto (2.27.0)
##### Summary:

When cloudformation stack state is 'present', Ansible creates the stack on the first run but fails on subsequent runs with the message `msg: Stack [STACKNAME] already exists` where `STACKNAME` is the name of the stack.
##### Steps To Reproduce:

Run a simple playbook that creates a cloudformation stack (state should be "present").
Run the playbooks two times.
##### Expected Results:

On first run the stack gets created. On second run no change is made and no error is generated.
##### Actual Results:

On first run the stack gets created. On second run the playbook fails with `msg: Stack [...] already exists`

mpdehaan:
Should be fixed by the above, so closing, thanks!

PR #6366, opened 2014-03-10
Fix for #6365: cloudformation module fails to update if stack exists

cloudify:
It looks like the error message returned by the AWS API has changed, this pull request updated the code that checks whether the stack already exists.

See #6365


mpdehaan:
Merged, thanks!

cloudify:
You're welcome :smile:

PR #4982, opened 2013-11-20
Allow with_nested to loop through data structures

mivok:
This change allows nested lists to refer to the current item in an outer list
using item[0] and item[1] and so on inside the with_nested clause. It does
this by performing the expansion variables on each iteration instead of just
once at the beginning, subsituting in the current values of each.

For example:

Variable data:

```

---
users:
    foo:
        ssh_keys:
            - "key1"
            - "key2"
    bar:
        ssh_keys:
            - "key3"
            - "key4"
```

and a task of:

```
    - debug: msg="{{item[0]}} - {{item[1]}}"
      with_nested:
        - users.keys()
        - users[item[0]]['ssh_keys']
```

Will give:

```
ok: [default] => (item=['foo', 'key1']) => {"item": ["foo", "key1"], "msg":
"foo - key1"}
ok: [default] => (item=['foo', 'key2']) => {"item": ["foo", "key2"], "msg":
"foo - key2"}
ok: [default] => (item=['bar', 'key3']) => {"item": ["bar", "key3"], "msg":
"bar - key3"}
ok: [default] => (item=['bar', 'key4']) => {"item": ["bar", "key4"], "msg":
"bar - key4"}
```

PR #6370, opened 2014-03-10
New lookup_plugin for nested dictionary (based on subelements but it will keep the key)

exarv:
New lookup_plugin for nested dictionary (based on subelements but it will keep the key)

The nested list can also be a list in stead of a dictionary, Then the item.1.0 contains 0, 1, 2 and item.1.1 contains the string value

Example:
- name: test
  action: debug msg="Device {{ item.0.0 }} with details: {{ item.0.1 }} has partition {{ item.1.0 }} with details: {{ item.1.1}}"
  with_nested_dict:
  - ansible_devices
  - partitions

### DIFF
--- a/lib/ansible/runner/lookup_plugins/nested_dict.py
+++ b/lib/ansible/runner/lookup_plugins/nested_dict.py
@@ -1,0 +1,74 @@
+# (c) 2013, Robert Verspuy <robert@exa.nl>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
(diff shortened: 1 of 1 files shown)


Peetz0r:
Awesome! I think I will definately use it someday :)

mpdehaan:
I'm unlikely to want to expand the iterators for this given this can be solved in a different way, and find the idea of having to say item.1.1 confusing -- with_subelements is confusing enough to explain, and "with_details" is a rather generic name too.

You could solve this by using with_subelements and having a key in the dict that has the value "name".

```
 mylist:
    - name: bob
      foo: bar
      baz: [ 1,2,3 ]
    - name: alice
      foo: glorp
      baz: [4,5,6]
```

with_subelements: mylist

That being said there is nothing wrong with you having a customer iterator of your own, we're just not going to want to include it in core.

exarv:
Ok, you are right. There is a workaround,

But what about the case where you want to iterate over the ansible_devices / partitions? the "sda" and "sda1" values are only available in the keys of the dictionary and not as a separate "name" hashkey.

So this means that the ansible devices / partitions can not be properly used in a task with the standard looping mechanisms of ansible.

I'm happy to change the code to something thats easier to use and not using item.0, item.1 etc.

Or maybe you can have a look at https://github.com/ansible/ansible/pull/4982 again.
because that also gives the posibility to use the keys in action, something not available with with_subelements.

Thnx!
Regards,
Robert

PR #6371, opened 2014-03-10
Fix for #6341: Use ''show'', not ''status'', for systemd service state

### LINKED ISSUE #6341: systemd service backend can fail to detect services not running
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.4.5 + PR #6339 
##### Environment:

Arch (Rolling Release / No Version Number).

systemd 208-11
##### Summary:

The `service` module should be able to correctly determine that systemd services reported as dead are not running.
##### Steps To Reproduce:

...on a system wherein the `eth0` Ethernet interface does not have an associated dhcpcd service running, apply:

```
- service: name=dhcpcd@eth0 state: started enabled:yes
```
##### Expected Results:

The relevant service should be started and enabled, or an error should be detected.
##### Actual Results:

Even when the service is down, if it is enabled, this task responds with:

```
ok: [localhost]
```

Inspection with `strace` shows that `servicectl status dhcpcd@eth0` is being invoked after an attempt to start the service, with the following output:

```
dhcpcd@ens3.service - dhcpcd on ens3
  Loaded: loaded (/usr/lib/systemd/system/dhcpcd@.service: enabled)
  Active: inactive (dead)
```

It is expected that this `inactive` result would cause the operation to be detected as a failure.

charles-dyfis-net:
`systemctl show` is built to be easily parsed by machine. `systemctl status` is built for human consumption only.

Use the `show` command from `get_status_dict()` when systemd is in use. Closes #6341.

### DIFF
--- a/library/system/service
+++ b/library/system/service
@@ -473,7 +473,27 @@ class LinuxService(Service):
         if location.get('initctl', None):
             self.svc_initctl = location['initctl']
 
+    def get_systemd_service_status(self):
+        (rc, out, err) = self.execute_command("%s show %s" % (self.enable_cmd, self.__systemd_unit))
+        d = dict(line.split('=', 1) for line in out.splitlines())
+        if d['ActiveState'] == 'active':
+            # run-once services (for which a single successful exit indicates
+            # that they are running as designed) should not be restarted here.
+            # Thus, we are not checking d['SubState'].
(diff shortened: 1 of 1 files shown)


jimi-c:
Merged, thanks!

PR #6374, opened 2014-03-10
Add the ability to run binary modules

nemunaire:
Compiled modules from a non-interpretable language, such as C, C++, Go, ... are indicated in [the documentation](http://docs.ansible.com/developing_modules.html#tutorial) as acceptable; but in practice, they trigger two errors:

First, `failure encoding into utf-8` when trying to transfer UTF-8 encoded text whereas this is binary content.

Then, `module is missing interpreter line` while looking for a shebang into binary content.

For example, [here](https://gist.github.com/nemunaire/9465051) is minimal module written in C and [here](https://gist.github.com/shirou/5444078) a Go module inspired from [the tutorial](http://docs.ansible.com/developing_modules.html). They both successfully pass the `hacking/test-module`.

This patch adds binary detection by looking for the ELF magic number in module content.

### DIFF
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -399,6 +399,7 @@ class Runner(object):
         (
         module_style,
         shebang,
+        binary_module,
         module_data
         ) = self._configure_module(conn, module_name, args, inject, complex_args)
 
@@ -408,7 +409,13 @@ class Runner(object):
 
         remote_module_path = os.path.join(tmp, module_name)
 
-        if (module_style != 'new'
+        if binary_module:
+            conn.put_file(module_data, remote_module_path)
+
+            cmd_chmod = "chmod a+x %s" % remote_module_path
+            self._low_level_exec_command(conn, cmd_chmod, tmp, sudoable=False)
+
+        elif (module_style != 'new'
            or async_jid is not None
            or not conn.has_pipelining
            or not C.ANSIBLE_SSH_PIPELINING
@@ -466,10 +473,12 @@ class Runner(object):
                 cmd = " ".join([str(x) for x in [remote_module_path, async_jid, async_limit, async_module]])
 
         if not shebang:
-            raise errors.AnsibleError("module is missing interpreter line")
-
-
-        cmd = " ".join([environment_string.strip(), shebang.replace("#!","").strip(), cmd])
+            if binary_module:
+                cmd = " ".join([environment_string.strip(), cmd])
+            else:
+                raise errors.AnsibleError("module is missing interpreter line")
+        else:
+            cmd = " ".join([environment_string.strip(), shebang.replace("#!","").strip(), cmd])
         cmd = cmd.strip()
 
         if "tmp" in tmp and not C.DEFAULT_KEEP_REMOTE_FILES and not persist_files and delete_remote_tmp:
@@ -1067,6 +1076,7 @@ class Runner(object):
         (
         module_style,
         module_shebang,
+        binary_module,
         module_data
         ) = self._configure_module(conn, module_name, module_args, inject, complex_args)
         module_remote_path = os.path.join(tmp, module_name)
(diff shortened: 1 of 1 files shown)


mpdehaan:
This looks pretty short and as previously discussed on the list a few times I'm definitely for the ability to have modules written in things like C/Go/etc.

In queue for testing.

We would want to make sure binary modules assumed they were recieving JSON argument-file  inputs, I think, to keep things maximally simple.

jctanner:
@nemunaire this appears to have broken async tasks:

```
$ pwd
/root/ansible.http/test/integration
$ TEST_FLAGS='-t test_async' make non_destructive
TASK: [test_async | run a 2 second loop] **************************************
fatal: [testhost] => Traceback (most recent call last):
  File "/root/ansible.http/lib/ansible/runner/__init__.py", line 529, in _executor
  File "/root/ansible.http/lib/ansible/runner/__init__.py", line 621, in _executor_internal
  File "/root/ansible.http/lib/ansible/runner/__init__.py", line 815, in _executor_internal_inner
  File "/root/ansible.http/lib/ansible/runner/action_plugins/async.py", line 39, in run
    (module_path, is_new_style, shebang) = self.runner._copy_module(conn, tmp, module_name, module_args, inject, complex_args=complex_args)
  File "/root/ansible.http/lib/ansible/runner/__init__.py", line 1066, in _copy_module
    ) = self._configure_module(conn, module_name, module_args, inject, complex_args)
ValueError: too many values to unpack


```

nemunaire:
Oops, my latest changes hasn't been added. That is now fixed!

Tatsh:
Can we add in not just checking for ELF but anything that might be a binary file the remote system can execute if given a proper shebang? For one thing, OS X uses Mach-O anyway. Check for `\b'FEEDFACE'` using the `struct.unpack()` function because endian-ness matters.

But regarding other languages which can be part binary, I get this error when using a PHP PHAR file:

```
fatal: [localhost] => failure encoding into utf-8
```

Just to note: A PHP PHAR file looks like a PHP file due to its shebang and second line `<?php` but actually has binary data. If these could be used as plugins, then it would be a lot easier to use the PHAR directly as a module. Completely removes the need to push dependencies to the remote machine before executing a module written in PHP.

Similar things could be done for other languages.

I have [https://github.com/Appdynamics/ansible-wordpress](this project) in wait until Ansible can use any language at all, in particular ones that an interpreter knows about but might contain binary. This can even include Bash scripts (there have been plenty of these, as simple self-contained files, that manage to include binary content in the file, for simplicity for the end-user).

Until the issue with binary in modules is fixed, I'm stuck with using a stub function to look for the code I depend on. And I have to push the dependencies to the machine first, where they are not used for anything else.

hoffoo:
So as of right now is there a way to write a module in compiled languages? Or is there some other way to get some JSON output from a binary and use it as vars in ansible?

headmin:
Any news? This issues is discussed for quite a while. https://groups.google.com/forum/#!topic/ansible-project/wsWutaQVcYc 

there's also go-lang demo code by WAKAYAMA Shirou : https://gist.github.com/shirou/5444078

amenonsen:
So this change needs to be rebased to apply to `lib/ansible/plugins/action/__init__.py` in current devel, but it doesn't look too difficult to do. But I think someone on the core team needs to look at it and decide what, if anything, we want to do about supporting binary modules in v2.

allen13:
Is this still an expected feature? I wrote an influxdb module in golang and I am getting that same utf8 error. I'm on ansible 1.9.2

amenonsen:
@allen13 It's still a relevant and desirable feature, but nobody seems to be working on it at the moment.

tom025:
This would also be very useful for me too.

I am currently writing a module to manipulate plists in go. Go has good support for manipulating plists in various formats and since it is statically compiled it requires no dependencies to be installed on the hosts.

tom025:
BTW As a workaround you can embed your binary executable into a bash script.

http://stackoverflow.com/questions/10491704/embed-a-executable-binary-in-a-shell-script

ddbenson:
FWIW -- we have the same issue with Python PEX files -- unsurprisingly I guess.

ddbenson:
We were able to get around with the workaround @tom025 provided.

sivel:
I'm going to go ahead and close this in favor of #13771

ddbenson:
I think a separate issue needs to be filed that covers the PEX / PHAR scenarios.

PR #6384, opened 2014-03-11
authorized_key: Be more specific about manage_dir

tyll:
- Explain which properties are managed when manage_dir is True.


mpdehaan:
I'm wondering if re: "Make sure to
-        set C(manage_dir=no) if you are using an alternate directory for
-        authorized_keys set with C(path)"

We should force manage_dir=no in this case.

mpdehaan:
Merged, thank you!  This will appear the next time we push docs.

PR #6390, opened 2014-03-11
group_vars plugin: do not parse hidden files in subfolders

srgvg:
I recently reorganised my group_vars files as directories, that allow multiple vars files to be combined, which is an new functionality that was introduced, I'm not sure when.
The group_vars plugin will parse all files en subdirectories of group_vars/$groupname.

I believe previous version of the group_vars plugin used to ignore hidden files here, hower, 
As we however use svn as VCS, I noticed here that ansible tries to parse the .svn/ files...

This patch fixes that.


mpdehaan:
merged, thanks!

PR #6391, opened 2014-03-11
Allow hash_behaviour=merge to be respected in core inventory

srgvg:
Seems like no one bumped into this, but having hash_behaviour=merge, was not respected when using inventory scripts.

ATM, only inventory scripts go through group calculation in Host.get_variables() as the group_vars plugins handles this directly for yaml files.


mpdehaan:
merged, thanks!

PR #6392, opened 2014-03-11
Add check for virtual packages during install process. If package is virtual - raise an error.

mightydok:
If apt module dont raise an error, then "provided by" package will be installed and on every ansible playbook run ther will be message that configuration was changed.
Also ansible will install provider package that not listed in playbook, without any information to playbook owner, so i add little check for virtual packages. 

As example, try to install php5-mhash package on Ubuntu 12.04:
ansible all -i production -m apt -a "pkg=php5-mhash state=latest" -k -K

### DIFF
--- a/library/packaging/apt
+++ b/library/packaging/apt
@@ -169,6 +169,10 @@ def package_status(m, pkgname, version, cache, state):
         ll_pkg = cache._cache[pkgname] # the low-level package object
     except KeyError:
         if state == 'install':
+            # Lets check pkgname is virtual package, if True - raise an error
+            if cache.is_virtual_package(pkgname):
+                m.fail_json(msg="Package '%s' is virtual and not available for install, please choose not virtual package" % pkgname)
+                return False, False, False
             if cache.get_providing_packages(pkgname):
                 return False, True, False
             m.fail_json(msg="No package matching '%s' is available" % pkgname)


mpdehaan:
It seems to me that we should be able to install virtual packages and we shouldn't take away this ability because it is returning "changed" on every run.

Can you explain a bit more about the underlying problem?

bcoca:
a fix, would be to identify that any provider for a virtual package is
already installed.

Also some feedback on which package was installed when it happens would be
nice.

mightydok:
mpdehaan - my fix solve two my problems:
1) Apt module always install provided package for virtual.package and always return Changed state for host. Thats so annoying for me =).
2) I loose some kind of control over my server, because in playbook type php5-mhash module but apt module install php5-common.

bcoca - i dont think so, because sysadmin must make a decision about which package must be installed, not ansible + apt. And sysadmin must have information about difference between playbook and server installed packages, so when sysadmin made this error ansible can give some feedback.

bcoca:
it is not always an error, many times a sysadmin doesn't care and will take
the default. may a suggest a virtuals=default|error setting?

mightydok:
bcoca - hm, maybe its more right way to solve this problem, and i can rewrite my code.
mpdehaan - any suggestions?

mpdehaan:
I'm not entirely sure what the correct behavior is here, but I don't want to do anything that would change the defaults in a way that would break any existing playbooks.   The "virtuals=default|error" thing seems non-intuitive at first look.

It seems it would be better to detect that it was already installed/fulfilled and return the proper changed state instead.

mightydok:
mpdehaan - ok, can u fix this error? Ansible apt module must return Change status False when provided package for virtual package already installed.
As Example: ansible all -i production -m apt -a 'pkg=php5-mhash state=latest'

mpdehaan:
Hi!

Thank you very much!

PR #6405, opened 2014-03-11
Add support for new pkgng repository format and annotations

fabianfreyer:
PACKAGESITE is deprecated and should no longer be used. Installing from different repositories is now done using the -r flag.

Also added support for annotations using a simple +/-/:<key>=<value> syntax to add, remove and change annotations, respectively.


mpdehaan:
merged, thanks!

PR #6416, opened 2014-03-12
Added support for egress rules and referring to security groups by name....

esp1:
... Will also create security groups referred to in rules if they do not already exist.

### DIFF
--- a/library/cloud/ec2_group
+++ b/library/cloud/ec2_group
@@ -24,7 +24,7 @@ options:
     required: false
   rules:
     description:
-      - List of firewall rules to enforce in this group (see example).
+      - List of firewall rules to enforce in this group (see example). Rules are assumed to be ingress rules unless 'type: egress' is specified.
     required: true
   region:
     description:
@@ -113,6 +113,11 @@ EXAMPLES = '''
       - proto: all
         # the containing group name may be specified here
         group_name: example
+      - proto: tcp
+        from_port: 123
+        to_port: 123
+        group_name: My_Awesome_Security_Group
+        type: egress
 '''
 
 try:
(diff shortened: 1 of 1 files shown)


esp1:
Squashed all commits from #5480 and resubmitting

codezerk:
I did something similar at #6164 
The difference was that when adding egress rules I made sure to remove the default egress rule which opens all ports for all addresses. It also had the ability to remove other unused egress rules that are not specified. This pull request doesn't address these important issues so I sure hope it is reconsidered to alter my solution to use type: [egress/ingress] like proposed here.

Right now this is useless as it requires manual intervention or the use of external scripts to have a fully automated solution.

esp1:
Would you be up for creating a patch to this pull request to address the concerns you mentioned?

mpdehaan:
So we have three of these now, #6602 #6416 and #6598 

Obviously there's a lot of duplicate effort here, but can you all please get together and sort out which one of these you prefer and define any changes that need to be made?

mmoya:
@esp1 care to give some feedback on issues mentioned in https://github.com/ansible/ansible/pull/6598#issuecomment-38197254 ? Thanks

codezerk:
There are several ways to go about this, In my (now closed) pull request
#6164 https://github.com/ansible/ansible/pull/6164 I addressed an issue:

When creating a security group on EC2, a default egress rule is always

> created, this rule gives unrestricted access to all protocols and all ports.
> In my pull request, if no egress rules are provisioned in the rules
> property then it will not change any egress rules that already exist on the
> group. But If there are rules provisioned then similar to ingress rules it
> will remove the ones that are not provisioned.

You can either choose this or make egress rules provisioning required, else
no egress rules will be configured on the group.
I think the choice should be based on this issue.

mecha1:
Sorry for the late reply. I'm happy with whatever folks decide, whether you want to use my stuff as a base or not. I think any of the offered solutions will be a nice improvement over the current functionality. I'm not doing anything wrt default egress rule, so fixing that behavior on this pull request would require a patch.

mpdehaan:
We've got one vote for @mmoya so I'm going to go with that after we merge we can take pull requests on top, thanks!!!

PR #6418, opened 2014-03-12
templates ignore_errors

### LINKED ISSUE #4892: want to be able to use a variable for the value of ignore_errors
AFAICT it's impossible to use a **variable** for _ignore_errors_ 
### Given

```
vars:
    ignore_err: no
```
### Solution Attempts

Interpreted as a string, so always true:

```
ignore_errors: $ignore_err
```

Dito:

```
ignore_errors: {{ ignore_err }}
```

Illegal YAML syntax:

```
ignore_errors: {{ ignore_err | bool }}
```

String again...:

```
ignore_errors: "{{ ignore_err | bool }}"
```

Maybe I'm stupid, but I can't see how to do this. I had to duplicate the task with one ignore_errors: yes and one no and control it with a when...

bcoca:
now you can use a variable for ignore_errors, this fixes #4892

### DIFF
--- a/lib/ansible/playbook/__init__.py
+++ b/lib/ansible/playbook/__init__.py
@@ -402,6 +402,10 @@ class PlayBook(object):
             ansible.callbacks.set_task(self.runner_callbacks, None)
             return True
 
+        # template ignore_errors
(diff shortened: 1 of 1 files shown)


mpdehaan:
merged, thanks!

PR #6420, opened 2014-03-12
added new role_names variable that exposes current roles list

### LINKED ISSUE #4392: role_name variable
It would be nice to be able test against roles as variables.  I would like to have a template where I can test against what roles a particular playbook is including.

Particularly in the iptables space. Being able to say open ports 80,443 because I included httpd role. Or any other grouping of ports based on inclusion of a role would be awesome.

bcoca:
fixes #4392


mpdehaan:
merged, thank you!

PR #6433, opened 2014-03-12
Always run cleanup after running cloud_integration tests

jlaska:
The proposed change allows the cloud tests to _always_ run cleanup, even if they fail.  Should a failure occur during a cloud target, the error code is saved, cleanup is performed, and the saved error code is returned.


mpdehaan:
merged, thanks!

PR #6434, opened 2014-03-12
Output tags while listing tasks

muffl0n:
Output tags for each listed task. This is primarily for debugging reasons but could be handy apart from that.

### DIFF
--- a/bin/ansible-playbook
+++ b/bin/ansible-playbook
@@ -230,7 +230,7 @@ def main(args):
                             set(task.tags).intersection(pb.skip_tags)):
                             if getattr(task, 'name', None) is not None:
                                 # meta tasks have no names
-                                print '    %s' % task.name
+                                print '    %s (tags: %s)' % (task.name, ", ".join(set(task.tags)))
                 print ''
             continue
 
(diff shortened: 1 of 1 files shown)


mpdehaan:
If there are no tags the output here would be "(tags: )" ?

Can we fix that part so it leaves off the (tags) part in that case?

Thanks!

muffl0n:
I'm not sure, can this ever be empty? Wouldn't it at least contain "all"? Maybe this should be filtered out as this has no benefit. I'm giving it a try.

muffl0n:
No tags specified:

```
  play #1 (localhost):
    debug msg="foo"
```

Tag "bar" specified:

```
  play #1 (localhost):
    debug msg="bar" (tags: bar)
```

The implicit tag 'all' is hidden.

muffl0n:
Is there anything I should change to get this merged?

mpdehaan:
I'm needing to think about this one if I would want it in the output.   Probably can be included but still need to consider.

bcoca:
already implemented as part as tag revamping

muffl0n:
@bcoca: Could you tell me when this was implemented? I just upgraded to 1.8.4 and was wondering if this is already included in this release and where to find it.

bcoca:
implemented in  1.9

Issue #6430, opened 2014-03-12
Getting errors during GATHERING FACTS


tonyjchong:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.5.2
##### Environment:

OSX on my laptop
vagrant 1.5.0
Ansible 
Ubuntu 12.04 for the VMs
##### Summary:

After updating to ansible 1.5.2, starting getting errors on the GATHERING FACTS portion of the play
##### Steps To Reproduce:

Fresh install vagrant 1.5.0, virtualbox 4.3.8, ansible 1.5.2
run an ansible play that worked prior to the update
##### Expected Results:

GATHERING FACTS should say localhost [OK]
##### Actual Results:

```
GATHERING FACTS ***************************************************************
failed: [localhost] => {"failed": true, "parsed": false}
invalid output was: Traceback (most recent call last):
  File "/Users/tchong/.ansible/tmp/ansible-tmp-1394605176.37-234489011526252/setup", line 3442, in <module>
    main()
  File "/Users/tchong/.ansible/tmp/ansible-tmp-1394605176.37-234489011526252/setup", line 2359, in main
    data = run_setup(module)
  File "/Users/tchong/.ansible/tmp/ansible-tmp-1394605176.37-234489011526252/setup", line 2300, in run_setup
    facts = ansible_facts(module)
  File "/Users/tchong/.ansible/tmp/ansible-tmp-1394605176.37-234489011526252/setup", line 2291, in ansible_facts
    facts.update(Network(module).populate())
TypeError: __init__() takes exactly 1 argument (2 given)
```

My first play is for cloudformation. Here's what it looks like

```

---

- name: bot AWS services that are shared between developer and prod

  hosts: localhost

  tasks:

  - name: create developer aws services with cloudformation
    cloudformation: >
      stack_name="{{ app_stack_name }}" state=present
      region={{ app_default_ec2_region }}
      template=cloudformation/stack_shared.json
    args:
      template_parameters:
        S3BucketName: "{{ stack_s3 }}"
```

Google searches turn up nothing. Thanks for your help.

mpdehaan:
Can you verify that you have 1.5.2 installed?

I am curious here as if we were having gathering facts issues on Ubuntu I think we would have had more reports.

In queue for investigation.

mpdehaan:
[...]

PR #6436, opened 2014-03-12
Fix issue#6430 by instantiating Network() with expected arguments

jlaska:
This resolves the problem reported in issue #6430


mpdehaan:
merged, thanks!

PR #6445, opened 2014-03-12
Improve openstack modules

kvdveer:
- Al modules now get their default authentication information from the environment, just like the openstack cli tools do.
- Where possible, objects (volumes, networks, routers, etc.) can be referenced both by name and by ID.
- nova_compute has received a general cleanup, and now accepts its parameters in much more intuitive forms.
- nova_compute now allows vm rebuild
- nova_compute now allows volume attachments.
- cinder_volume allows creation and deletion of volumes.
- quantum is now usable for non-admins (e.g. public cloud users)

This fork request isn't quite ready for merging yet. This merge request is mostly intended to solicit some reviews.


mpdehaan:
This can be checked in the review but quick questions:

"Al modules now get their authentication information from the environment, just like the openstack cli tools do."

Want to make sure this is in addition to existing parameters and not removing them and also using a module_utils/openstack file for common code.

kvdveer:
Authentication now just retrieves the defaults from the environment; one is still allowed to override using the parameters. Environment variables (usually set through .bashrc) typically aren't available when working remote; so in practice these defaults work only when working via localhost.

mpdehaan:
Thanks for this, as this is a rather large pull request, can you ask on ansible-devel and get some other OpenStack users to help test and offer feedback?

Thanks!

mpdehaan:
Ok, so talked about this with some folks.   

We generally think this is too large to accept and review, but would be open to a series of smaller focused pull requests.  

The merge commits in the above series are things we can't accept -- as a rule -- because they are impossible to audit.

The shared module stuff is great and I know some others are also working with this in parallel so I think a post to coordinate on ansible-devel may be easiest.

Hope this makes sense -- very interested in upgrades, but I'm worried at what hidden behavior may lurk in something of this size, however unintentional.

Sound reasonable?

Thanks!

kvdveer:
This sounds reasonable; I'm having auditing problems myself, hence the 
lack of an actual merge request.

I'm willing to cut things into smaller (module-sized) chunks. But 
still, the first module would include (most of?) the shared module. 
Would that be acceptable?

mpdehaan:
If it's moving things mostly to an openstack shared code file that is totally reasonable.

mpdehaan:
Closing this per the above split.

Thanks!

Issue #6431, opened 2014-03-12
Cloning git repository fails with ansible 1.5.2


fago:
##### Summary:

Please summarize your request in this space.  You will earn bonus points for being succinct, but please add enough detail so we can understand the request.
##### Steps To Reproduce:

Since the update from 1.4.4 to 1.5.2 (via the ubuntu PPAs) cloning a Git repository started failing. It turns out the directory gets cloned correctly, but ls-remote command fails now. When executing the command manually in the cloned Git directory it works, but when executing it manually elsewhere it produces the same error. 
Maybe ansible executes the command without chaing the working directory?

That's the configuration which worked previously as is: (drush_version defaults to "6.x")

```
- name: Drush | Cloning drush Git repo
  git: repo=https://github.com/drush-ops/drush.git
       dest=/opt/drush.git
       version={{ drush_version }}
```
##### Expected Results:

Task runs through as it should ;-)
##### Actual Results:

Error message:

```
failed: host] => {"cmd": ["/usr/bin/git", "ls-remote", "origin", "-h", "refs/heads/6.x"], "failed": true, "item": "", "rc": 128}
stderr: fatal: 'origin' does not appear to be a git repository
fatal: The remote end hung up unexpectedly

msg: fatal: 'origin' does not appear to be a git repository
fatal: The remote end hung up unexpectedly
```

fago:
Updated description to match the template.

mpdehaan:
This doesn't appear to be running through the new get_url code or anything so _may_ be a problem with github.

We'll see if we can reproduce this.

mpdehaan:
This works for me on both 1.5.2 and the devel branch.

As we were seeing issues with github yesterday (lots) and some intermittent this morning, I expect this was github failing.  If you are still seeing problems, let me know.

This seems to align with what I'm seeing on https://status.github.com/messages

jlaska:
I wonder if this may still be a problem.  Github appears to be operating normally, and I'm hitting the reported problem.

```
TASK: [update project using git and accept hostkey] ***************************
failed: [localhost] => {"cmd": "/usr/bin/git ls-remote origin -h refs/heads/master", "failed": true, "rc": 128}
stderr: fatal: 'origin' does not appear to be a git repository
fatal: The remote end hung up unexpectedly

msg: fatal: 'origin' does not appear to be a git repository
fatal: The remote end hung up unexpectedly

FATAL: all hosts have already failed -- aborting
```

[...]

PR #6446, opened 2014-03-12
git: Execute git commands in `dest` directory

patricksmith:
Fixes bug reported in https://github.com/ansible/ansible/issues/6431.

If a `version` other than "HEAD" is supplied to the git command, ansible will check to see if it is a git branch or git tag. Both of these commands were failing because they were not run in the correct directory (the destination directory), with output like:

```
failed: [localhost] => {"cmd": ["/usr/bin/git", "ls-remote", "origin", "-t", "refs/tags/v1.0.0"], "failed": true, "item": "", "rc": 128}
stderr: fatal: 'origin' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

msg: fatal: 'origin' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

```

I've tested this locally and this allows me to properly clone a git repo with a specified version, like:

```
- name: Checkout repo from git
  sudo_user: our_user
  git: repo=git@bitbucket.org:{{ org }}/{{ repo }}.git
       version=v1.0.0
       dest={{ dest_code_dir }}
```

### DIFF
--- a/library/source_control/git
+++ b/library/source_control/git
@@ -263,7 +263,7 @@ def get_remote_head(git_path, module, dest, version, remote, bare):
 
 def is_remote_tag(git_path, module, dest, remote, version):
     cmd = '%s ls-remote %s -t refs/tags/%s' % (git_path, remote, version)
-    (rc, out, err) = module.run_command(cmd, check_rc=True)
+    (rc, out, err) = module.run_command(cmd, check_rc=True, cwd=dest)
     if version in out:
         return True
     else:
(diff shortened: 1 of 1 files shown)


mpdehaan:
Yep, this is a result of security changes on an private branch which was unfortunate.

@jimi-c just developed an identical patch around this and we are going to include this in a 1.5.3 release likely tomorrow.   Give us a few to merge things in.

jimi-c:
Merged, thanks!

PR #6448, opened 2014-03-12
Fix #5679 again after recent changes in core

### LINKED ISSUE #5679: lineinfile ignores newline “\n” in the line argument
I need to insert a multiline string with the _lineinfile_ module. According to [this article](http://fabian-affolter.ch/blog/the-lineinfile-module-of-ansible/), the special char `\n` should work, but in the current version it doesn’t.

```
lineinfile:
    dest=/tmp/file
    regexp='^MY_CONF='
    line='#some comment\nMY_CONF=value'
```

Result:

```
#some comment\nMY_CONF=value
```

Instead of:

```
#some comment
MY_CONF=value
```

Neither an actual new line doesn’t work:

```
lineinfile:
    dest=/tmp/file
    regexp='^MY_CONF='
    line='#some comment
          MY_CONF=value'
```

Result:

```
#some comment MY_CONF=value
```

jirutka:
My previous fix for #5679 stop working after some recent changes in Ansible and the test case I’ve wrote wasn’t ported to the new test suite. There’s a fix that works on the current version and also a new test.

### DIFF
--- a/library/files/lineinfile
+++ b/library/files/lineinfile
@@ -350,9 +350,8 @@ def main():
         if ins_bef is None and ins_aft is None:
             ins_aft = 'EOF'
 
-        # Replace the newline character with an actual newline. Don't replace
-		# escaped \\n, hence sub and not str.replace.
-        line = re.sub(r'\n', os.linesep, params['line'])
+        # Replace the newline character with an actual newline.
+        line = params['line'].decode('string_escape')
 
         present(module, dest, params['regexp'], line,
                 ins_aft, ins_bef, create, backup, backrefs)
--- a/test/integration/roles/test_lineinfile/tasks/main.yml
+++ b/test/integration/roles/test_lineinfile/tasks/main.yml
@@ -209,3 +209,21 @@
     that:
     - "result.stat.md5 == 'fef1d487711facfd7aa2c87d788c19d9'"
 
+
+- name: insert a multiple lines at the end of the file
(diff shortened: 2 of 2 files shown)


spasche:
I gave it a try and found out that it is not idempotent when you try to add a line with a new line (thus two lines). If the action is run more than once, the two lines are appended at the end even if they are already there.

jirutka:
Yeah, that’s true. This module is designed to work with just single lines and it’ll not be trivial and clean to make this work correctly with multiple lines IMHO.

spasche:
Right, I understand. Maybe we need a "linesinfile" module ;-)

jirutka:
I’m actually working on a module specifically for managing “key=value” configuration files (shell-like, Java properties, ini, …) that will handle most of the problems I have with lineinfile. :)

spasche:
Great, looking forward to it :)

While a key value manipulation module would be nice to have, a use case I encountered several times which might not be covered using a key value mechanism is simply to add block of lines after a specific position in a file (section in a samba smb.conf file for instance). It would be nice to have a way of doing that. But sorry, I digress ;-)

mpdehaan:
FYI there is the ini_file module.  I'd probably welcome "API" compatible upgrades versus bifurcating it if possible.

Always better to just use template when you can!

mpdehaan:
merged, thanks!

Issue #6397, opened 2014-03-11
mysql_db module does not import database schema from gzipped files


drybjed:
##### Issue Type:

Bug Report
##### Ansible Version:

Ansible 1.6 (devel)
##### Environment:

Debian
##### Summary:

After latest changes in 1.5.1, mysql_db module does not import provided schema file.
##### Steps To Reproduce:

```

---
- hosts: all
  sudo: True

  tasks:
    - name: Install PHPMyAdmin packages
      apt: pkg={{ item }} state=latest install_recommends=no
      with_items: [ 'mysql-server', 'phpmyadmin', 'python-mysqldb' ]

    - name: Create database for PHPMyAdmin
      mysql_db: name=phpmyadmin state=present
      register: phpmyadmin_database

    - name: Import PHPMyAdmin schema
      mysql_db: name=phpmyadmin state=import
                target=/usr/share/doc/phpmyadmin/examples/create_tables.sql.gz
      when: phpmyadmin_database is defined and phpmyadmin_database.changed == True
```
##### Expected Results:

Schema is imported correctly.
##### Actual Results:

```
PLAY [all] ******************************************************************** 

GATHERING FACTS *************************************************************** 
ok: [olaf]

TASK: [Install PHPMyAdmin packages] ******************************************* 
ok: [olaf] => (item=mysql-server,phpmyadmin,python-mysqldb)

TASK: [Create database for PHPMyAdmin] **************************************** 
changed: [olaf]

TASK: [Import PHPMyAdmin schema] ********************************************** 
failed: [olaf] => {"failed": true}
msg: gzip: unrecognized option '--user=root'
Try `gzip --help' for more information.

FATAL: all hosts have already failed -- aborting

PLAY RECAP ******************************************************************** 
           to retry, use: --limit @/home/drybjed/phpmyadmin.retry

olaf                       : ok=3    changed=1    unreachable=0    failed=1
```

drybjed:
via `git bisect`, bug was introduced in ba0fec4f4238e381adc5f5a2f286bcd5b32f0586

mpdehaan:
is this something you would like to work on a fix for?

I'd rather not revert that commit.

Thanks!

drybjed:
[...]

PR #6450, opened 2014-03-12
Fix for #6397 and potentially more: call run_command with "use_unsafe_shell = True" where needed.

renaudguerin:
Since commit ba0fec4, run_command() requires use_unsafe_shell=True for shell pipelines to work.
This fixes it for hopefully all modules, including mysql_db (issue #6397)

### DIFF
--- a/library/database/mysql_db
+++ b/library/database/mysql_db
@@ -135,7 +135,7 @@ def db_dump(module, host, user, password, db_name, target, port, socket=None):
         cmd = cmd + ' | bzip2 > ' + target
     else:
         cmd += " > %s" % target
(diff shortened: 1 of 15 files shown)


mpdehaan:
We are not going to apply this because it additionally needs to add pipes.quote() around the args, @risaacson is working on a patch and should have this soon.

This will then be applied to devel and included in a 1.5.3 branch for release that is slated for tomorrow.

Thanks!!!  Let me know if there are questions.

mpdehaan:
You can also see the note I added to coding guidelines to explain this here (just now)

https://github.com/ansible/ansible/commit/29c32890d5ca262eb05f9d1b72d5f8c5f5467d26

Issue #6443, opened 2014-03-12
git module fails when using sudo on Ansible 1.5.1/1.5.2


dmkoch:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.5.2
##### Environment:

Ubuntu 13.04 x86_64 on Ansible system and managed host
##### Summary:

git module broken as of Ansible 1.5.1/1.5.2 when running with sudo
##### Steps To Reproduce:

Run a git task over ssh as root, using sudo to an unprivileged user

```
- name: checkout project from git
  git: repo={{ my_repo }} dest=/srv/myproject accept_hostkey=yes
  sudo: yes
  sudo_user: myuser
```

/srv/myproject is owned by myuser.
##### Expected Results:

A copy of myproject from my_repo would be checked out in /srv/myproject
##### Actual Results:

```
failed: [host1.example.net] => {"cmd": ["/usr/bin/git", "ls-remote", "git@github.com:example/example.git", "-t", "refs/tags/HEAD"], "failed": true, "item": "", "rc": 128}
stderr: fatal: failed to stat '.': Permission denied

msg: fatal: failed to stat '.': Permission denied

FATAL: all hosts have already failed -- aborting

```

The task runs as expected when I downgrade Ansible to 1.5.0.

mpdehaan:
Can you check this on the devel branch?  We are working on a 1.5.3 candidate using some cherry-picks off there and this seems ok, despite some minor issues with adding host keys multiple times that we are chasing.

There was an issue with the way the host key commands were being passed not through the shell in that version.

Thanks!

yuriks:
I can confirm and have the same issue. What seems to be happening is that sudo doesn't change the cwd, it stays in root's cwd, then git tries to read the cwd and fails.

dmkoch:
The same error occurs with the devel branch at https://github.com/ansible/ansible/commit/ebfc776

packetcollision:
It's probably an ubuntu-specific issue. Ubuntu does not reset the HOME env when doing sudo, whereas Debian (and probably others) do. You can always override HOME by using "sudo -H" so maybe that should be the default if it isn't already.

yuriks:
`ansible -vvvv` indicates that `-H` is already being passed to sudo:

```
<provision> ESTABLISH CONNECTION FOR USER: root
<provision> REMOTE_MODULE git repo=git@github.com:digEcor/glide-cm.git dest=/home/tas-cm/glide-cm version=develop
[...]

PR #6451, opened 2014-03-12
Work around for cwd stat problems with run_command and sudo

dmkoch:
This fixes issue #6443


jimi-c:
Merged with one small change, just added a check to make sure the cwd exists and is a directory before trying to switch to it. Thanks!

PR #6456, opened 2014-03-12
split in get selections should only split line at first :, so answers ca...

marklee77:
split needs to be told to only split on first ':', otherwise there's an error if you have an answer that already has a : in it. I had this problem configuring ldap-auth-config/ldapns/ldap-server, which has a default value of ldapi:///

also, need to pass use_unsafe_shell=True when setting selections.

### DIFF
--- a/lib/ansible/playbook/play.py
+++ b/lib/ansible/playbook/play.py
@@ -229,11 +229,11 @@ class Play(object):
                                 allow_dupes = utils.boolean(meta_data.get('allow_duplicates',''))
 
                         # if any tags were specified as role/dep variables, merge
-                        # them into the passed_vars so they're passed on to any 
+                        # them into the current dep_vars so they're passed on to any 
                         # further dependencies too, and so we only have one place
-                        # (passed_vars) to look for tags going forward
+                        # (dep_vars) to look for tags going forward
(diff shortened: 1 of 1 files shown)


mpdehaan:
The "use_unsafe" thing requires some usage of pipes.quote elsewhere I think, looking shortly.

(note to us: this need not go on 1.5.3 as this is a 1.6 only module)

Thanks!

mpdehaan:
Here:

```
    cmd = ["echo '%s %s' |" % (pkg, data), setsel]
```

I would expect the pipes.quote to surround pkg and also perhaps data

marklee77:
That problem is in the original code that currently exists in upstream. You
do have a good point, so I've fixed it in my branch as requested.

On Thu, Mar 13, 2014 at 1:31 AM, Michael DeHaan notifications@github.comwrote:

> Here:
> 
> ```
> cmd = ["echo '%s %s' |" % (pkg, data), setsel]
> ```
> 
> I would expect the pipes.quote to surround pkg and also perhaps data
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/ansible/ansible/pull/6456#issuecomment-37490875
> .

## 

Mark Lee Stillwell
marklee@fortawesome.org

PR #6475, opened 2014-03-13
Bypass warning to allow MySQL replication through SSH tunnel or VPN.

jazmit:
The mysql_replication module is set to fail on warnings.  If MASTER_PASSWORD is set without enabling MySQL SSL, a security warning is printed which will cause the module to fail, but this warning can be safely ignored if an SSH tunnel or VPN is used for replication.  This change allows the playbook to continue to run when this warning is given.

### DIFF
--- a/library/database/mysql_replication
+++ b/library/database/mysql_replication
@@ -161,7 +161,13 @@ def start_slave(cursor):
 
 def changemaster(cursor, chm):
     SQLPARAM = ",".join(chm)
-    cursor.execute("CHANGE MASTER TO " + SQLPARAM)
+    with warnings.catch_warnings(record=True):
+        # If master_password is set but SSL parameters are not,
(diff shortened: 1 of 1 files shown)


mpdehaan:
So the "with warnings" stuff is python 2.6, so we'd have to probably add some guard code here so this can still work on 2.4 remote hosts.

jazmit:
Ok, I'll add this and re-submit.

James Smith
james@schoolshape.com

mpdehaan:
That would be great, thanks!

jgornick:
@mpdehaan Was this PR accidentally closed?

jgornick:
Another approach to ignore the warning can be found @ #6040

jazmit:
I prefer the approach in #6040, this one should be merged rather than my patch.

PR #6479, opened 2014-03-13
[GH-6449] Apply --limit to localhost and 127.0.0.1

### LINKED ISSUE #6449: ansible-playbook ignores --limit option
##### Issue Type:

BugReport
##### Ansible Version:

ansible 1.6 (d37f0c6d12)
##### Environment:

N/A
##### Summary:

ansible-playbook ignores --limit option.
##### Steps To Reproduce:

ansible-playbook hosts.yaml --limit=boo
##### Expected Results:

PLAY [Localhost] **************************************************************
skipping: no hosts matched

PLAY [All servers] ************************************************************
##### Actual Results:

PLAY [Localhost] **************************************************************

TASK: [setup] *****************************************************************
ok: [localhost]

gildegoma:
This is a first proposal to fix #6449. It was still a work in progress, but I think it is worth to start the discussion on this base.

Note that I am currently not fully satisfied with the proposed fix, because `!localhost`will also skip `127.0.0.1`. Many thanks in advance for your review!

I use following playbook for my tests (e.g. `--limit=localhost`, `--limit='group1:!127.0.0.1'`, etc)

```
- name: mix local and remote
  hosts: 'localhost:all'
  gather_facts: no

  tasks:
    - name: 'remote AND localhost task, but not 127.0.0.1'
      command: echo 'localhost and remote'

- name: 127.0.0.1

  hosts: 127.0.0.1
  gather_facts: no

  tasks:
    - name: 127.0.0.1 task 1
      command: echo 'local task 1'

    - name: 127.0.0.1 task 2
      command: echo 'local task 2'

- name: localhost

  hosts: localhost
  gather_facts: no

  tasks:
    - name: localhost task 1
      command: echo 'local task 1'

    - name: localhost task 2
      command: echo 'local task 2'

- name: group1

  hosts: group1
  gather_facts: no

  tasks:
    - name: remote task 1
      command: echo 'remote task 1'
    - name: local action
      local_action: command echo 'local action'

- name: group2

  hosts: group2
  gather_facts: no

  tasks:
    - name: remote task 2
      command: echo 'remote task 2'
```

### DIFF
--- a/bin/ansible-playbook
+++ b/bin/ansible-playbook
@@ -97,7 +97,7 @@ def main(args):
 
     inventory = ansible.inventory.Inventory(options.inventory)
     inventory.subset(options.subset)
-    if len(inventory.list_hosts()) == 0:
+    if len(inventory.list_hosts('all:localhost:127.0.0.1')) == 0:
         raise errors.AnsibleError("provided hosts list is empty")
 
     sshpass = None
--- a/lib/ansible/inventory/__init__.py
+++ b/lib/ansible/inventory/__init__.py
@@ -384,8 +384,6 @@ class Inventory(object):
         """ return a list of hostnames for a pattern """
 
         result = [ h.name for h in self.get_hosts(pattern) ]
-        if len(result) == 0 and pattern in ["localhost", "127.0.0.1"]:
-            result = [pattern]
         return result
 
     def list_groups(self):


mpdehaan:
While there may be a bug on the first line 100, the code is somewhat idiomatic and I'd want to understand the why.

list hosts with no args should return 0 just as that expression, though the other might _NOT_ want to.

There's also the details of implicit localhost being their mainly for delegate_to/local_action so new code for handling implict local in the host spec needs to be checked somewhat carefully.

Mostly thinking out loud.

mpdehaan:
Thanks for the efforts on this -- probably going to defer this in favor of something like the above.

gildegoma:
@risaacson @mpdehaan Thanks for taking care of this.

As already mentioned, this patch was first provided to illustrate my understanding about how limit on implicit localhost/127.0.0.1 should work, to easier agree on the final behavior to obtain.

I tested fc5d0d8, but for me it doesn't solve yet the problem: `FATAL: all hosts have already failed -- aborting` is still wrongly raised.

In general, I guess that we agree that all `--limit` rules should apply the same to these implicit hosts, and in particular:
- `localhost` and `127.0.0.1` should be treated as distinct host names (no aliasing/synonyms). 
- Subset that don't mention implicit hosts should work. For instance `!group1` with the above playbook should give:

```
PLAY [mix local and remote] *************************************************** 

TASK: [remote AND localhost task, but not 127.0.0.1] ************************** 
changed: [machine2]
changed: [localhost]

PLAY [127.0.0.1] ************************************************************** 

TASK: [127.0.0.1 task 1] ****************************************************** 
changed: [127.0.0.1]

TASK: [127.0.0.1 task 2] ****************************************************** 
changed: [127.0.0.1]

TASK: [localhost task 1] ****************************************************** 
changed: [localhost]

TASK: [localhost task 2] ****************************************************** 
changed: [localhost]

PLAY [group1] ***************************************************************** 
skipping: no hosts matched

PLAY [group2] ***************************************************************** 

TASK: [remote task 2] ********************************************************* 
changed: [machine2]

127.0.0.1                  : ok=2    changed=2    unreachable=0    failed=0   
localhost                  : ok=3    changed=3    unreachable=0    failed=0   
machine2                   : ok=2    changed=2    unreachable=0    failed=0   
```

Voilà for my 2 cents! Hope it helps :)

risaacson:
@gildegoma There really ends up being two paths that come into ansible-playbook that trigger the implicit addition of localhost and 127.0.0.1. First there is the explicit inclusion in the playbook in which it was the first effort for implicit localhosts. The second which you called out here which was not covered earlier is using --limit to subset the inventory. The commit that you tested was the first revision that I had put out after diving into this and seeing your good points.

I think that what you are pointing out in your most recent comment is a new feature that I think may be handy in some situations but if set by default breaks one of the contracts that ansible has made about hosts or groups specified in playbooks. I'll defer to @mpdehaan on this though as it has to deal with a core behavior. If you want to pursue this I would suggest creating a feature issue that clarified the situation, like the comment above, and we can evaluate it for work or if you want to put together a pull request we will welcome that also.

Otherwise I think that we are golden for the original issue here.

Thanks!

-r

gildegoma:
Thanks @risaacson for your work and your clarifications.

I tested your bugfix branch, but I'm still confused about the obtained results... I reported them to the original issue (#6449). As the things are getting harder, I would suggest to add some unit tests to specify and cover these cases (e.g. `--limit=localhost`, `--limit='!localhost'`, etc), when using implicit (or explicit) localhost/127.0.0.1.

> I think that what you are pointing out in your most recent comment is a new feature that I think may be handy in some situations but if set by default breaks one of the contracts that ansible has made about hosts or groups specified in playbooks.

Sorry but with my current understanding, I cannot figure out the broken contract that you mention. You might have good reason to not include this point in the current bug fix, but I still do think that it will be misleading (and possibly frustrating) if `--limit` option differently affects implicit localhost/127.0.0.1 compared to the inventoried hosts. There is also the case where `localhost` is explicitly specified in the ansible inventory... I am afraid that it ends up with complicated documentation on these special cases, and some user that requests for this change. Don't you think?

PR #6480, opened 2014-03-14
Properly catch import errors in apt

do3cc:
When one accidentally tries to run this module as a user, he gets the error message that python-apt must be installed, no matter what. Because importing apt will trigger an exception as a regular user. Explicitly catching the ImportError will let the exception bubble. The exception clearly says Permission denied somewhere, and the user has a better idea, what he must fix.

### DIFF
--- a/library/packaging/apt
+++ b/library/packaging/apt
@@ -153,7 +153,7 @@ HAS_PYTHON_APT = True
 try:
     import apt
     import apt_pkg
-except:
+except ImportError:
     HAS_PYTHON_APT = False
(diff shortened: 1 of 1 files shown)


do3cc:
Here is a quick way to reproduce it:
ansible -m apt ANY -u USER

mpdehaan:
Hmm, it's curious that python-apt would do things on init prior to construction of objects, but I can see it doing that.   

Good catch on the ImportError, merging this now, thanks!

PR #6481, opened 2014-03-14
File fixes

bcoca:
This PR is NOT currently ready for merge, but I wanted to publish for comments and refactoring.

I noticed that now dangling symlinks get created w/o force=yes (contrary to docs) so I took a peek into the file module and noticed it had an indentation issue ... but also lots of complicated conditionals, in many cases useless. This PR is now a refactor that will fix the 1 issue and hopefully bring back some sanity into the module, its gotten very complicated to figure when it does what it does.

I also saw that 2 identical functions were created for changing fs permissions, making them one but keeping old names until I get confirmation and then I'll fix all references from other modules.

I'll also be adding test cases to ensure all functions as expected.

please test, comment, correct, etc ...

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -954,7 +954,7 @@ class AnsibleModule(object):
                 context = self.selinux_default_context(dest)
 
         try:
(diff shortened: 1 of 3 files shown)


bcoca:
cleaned up and resubmitted as #6506

PR #6486, opened 2014-03-14
Rename present to running, add new present state

### LINKED ISSUE #6395: docker module not idempotent
##### Issue Type:

Feature Request
##### Ansible Version:

ansible 1.6 (devel c7d1c1a306) last updated 2014/03/07 16:45:29 (GMT +200)
##### Environment:

Ubuntu saucy & precise
##### Summary:

I expected the docker module to run a new container only if the container doesn't exist yet.
##### Steps To Reproduce:

Given this task:

```
- name: run some container
  docker:
    image: fish/some-container
    name: some-containers-name
```
##### Expected Results:

I expected the docker module to run a new container only if the container doesn't exist yet.
##### Actual Results:

It's starting a new container each time. 
##### Notes:

That should be easy to implement if the name attribute is provided:

If there is no container with that name: run (docker run) a new one
If there is one with that name, and it's stopped: start it (docker start)
If there is one with that name and it's running: noop
Please summarize your request in this space.  You will earn bonus points for being succinct, but please add enough detail so we can understand the request.

discordianfish:
The new present state just makes sure that a container exists, not that
it's running, although it get started one creation.
This is very useful for data volumes. This also changes the old
present, now running (default) state to only create the container if
it's not found, otherwise it just get started.

See also discussion on mailinglist:
https://groups.google.com/forum/#!topic/ansible-devel/jB84gdhPzLQ

This closes #6395

### DIFF
--- a/library/cloud/docker
+++ b/library/cloud/docker
@@ -148,7 +148,7 @@ options:
       - Set the state of the container
     required: false
     default: present
-    choices: [ "present", "stopped", "absent", "killed", "restarted" ]
+    choices: [ "present", "running", "stopped", "absent", "killed", "restarted" ]
     aliases: []
   privileged:
     description:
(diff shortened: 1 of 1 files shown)


discordianfish:
Here is a simple example how to use docker volume containers with this change:

```
---
- name: create mysql-data container
  docker: image=busybox name=mysql-data volumes=/var/lib/mysql/ command=/bin/true state=present

- name: run mysql-server
  docker:
    image: jonwood/mysql-server
    name: mysql-server
    volumes_from: mysql-data
```

discordianfish:
Ok, this can be merged from my POV.

mpdehaan:
Merged, thank you!

PR #6487, opened 2014-03-14
add with_product for nested loop

ant31:
I had the need to do nested loop (product): 

```
for x in A:
   for y in B 
     (x,y)
```

i.e: 
Execute a action over a list and delegate_to all host in a group:

``` yaml
- name:  My Role | Send configs
  template: src="{{item.0}}" dest=/etc/mysoft/conf.d/ mode=0644
  delegate_to: "{{item.1}}"
  with_product:
    - mytemplates
    - groups['logstash-indexers']
```


mpdehaan:
So there's already "with_nested" for this, which is also documented.

Since we are unlikely to read comments on closed tickets, please stop by the development list if you have further questions!

PR #6488, opened 2014-03-14
add register_append keyword

ant31:
Hello,

register_append is exactly the same as register but instead of overwrite previous 'key' it will append result to it:

``` python
hash[register] = result
hash[register_append].append(result)
```

We needed this feature to allow DRY/shared handlers:

Use case
Each time a new template is generated, backup the old one to a third server:
The backup task is done/write only once in the handler. 
Without register_append, the 'new_tpl' variable would contains the last result only.

``` yaml
tasks:
- name: template
  template: src=mytpl.conf.j2 dest=/root/mytpl.conf backup=yes
  register_append: new_tpl
  notify:
    - backup conf
  tags:
    - backup

- name: template 2
  template: src=mytpl2.conf.j2 dest=/root/mytpl2.conf backup=yes
  register_append: new_tpl
  notify:
    - backup conf
  tags:
    - backup

handlers:
- name: backup conf
  synchronize: mode=pull src={{item.backup_file}} dest=/archive/{{inventory_hostname}}/{{item.backup_file | basename}}
  delegate_to: backup-server-1
  when: '{{item.backup_file is defined}}'
  with_items: new_tpl
  register: backup
```

Thank you

Antoine Legrand


mpdehaan:
I can see a use for this -- We'll think about this one. 

My first thoughts is that someone might want to set a hash key, in which case we might want to support that too, and also that it might just want to be called "append".

ant31:
Thx for your fast reply, 

What do you mean by "set a hash key"?

Antoine Legrand

mpdehaan:
with_dict is already a thing for iterators.

I mean being able to do

dict_variable[new_key] = new_value

So "insert into a hash" as well as append to a list.

ant31:
I have updated code:
added 2 new keywords:

``` yaml
register_as_list: boolean
register_as_dict: hashkey
```

They both require the 'register' keyword to be present.
Its define the data structure, 
- register_as_list: yes   =>  appends the result to a list
- register_as_dict: key => adds the result to a hash[key]

mpdehaan:
The merge commits in this one will make it impossible to merge.

I would definitely recommend looking into git topic branches, and using git pull --rebase vs git pull and git rebase vs git merge, which should help.

If you can resubmit we can get this back in queue.  Thanks!

PR #6491, opened 2014-03-14
[apt_repository] Correct check_mode conditional logic

jlaska:
The desired behavior is to _not_ add the ppa signing key when check_mode is
enabled.  This fix corrects the conditional logic to comply with the stated
behavior.

### DIFF
--- a/library/packaging/apt_repository
+++ b/library/packaging/apt_repository
@@ -353,9 +353,9 @@ def get_add_ppa_signing_key_callback(module):
         module.run_command(command, check_rc=True)
 
     if module.check_mode:
-        return _run_command 
-    else:
         return None
+    else:
+        return _run_command
 
 
 def main():


mpdehaan:
Merged, thanks!

sthen:
It looks like the release tarball was re-rolled for this fix without bumping the version number - this causes a number of problems for OS packagers; could I request that if something is important enough to re-roll a tar.gz, that the version number is bumped too? Thanks very much!

phantomwhale:
Sorry to ask what might be obvious - would this change mean apt_add_respository starts adding keys again when pulling in PPAs - and also mean that this is the wrong way around on version 1.5.3 ?

If so, then this would address my issue on the mailing list https://groups.google.com/forum/#!topic/ansible-project/qaMcLpyuJis and I'll avoid opening another issue.

jlaska:
@phantomwhale you are correct, this fix addresses the problem you reported on the list.  The fix was included in 1.5.3, please verify you are using the the [correct 1.5.3 tarball](https://pypi.python.org/pypi/ansible). Distribution packagers have been alerted and have rolled out updates with this version.

phantomwhale:
Interesting, I am using a 1.5.3 tarball (compiled from source as apt only had 1.5.2 at that stage).

I'll try reinstalling in case I'm using the "wrong" 1.5.3 tarball (although surely there should be only one tarball for 1.5.3 ?)

PR #6492, opened 2014-03-14
netctl service support

charles-dyfis-net:
Arch Linux's neworking support uses a wrapper for systemctl services called netctl. netctl services need to be enabled via a wrapper to systemd rather than invoked directly.

Some other enhancements are made in the course of supporting this robustly: run_command() is given to accept tuples as well as the already-supported lists and strings, and Service.execute_command() is granted the ability to pass keyword arguments through to run_command(), which is used to pass `check_rc=True` through for systemd services. (This is kept off for other service backends due to inability to test, though it probably should be turned on in at least some subset of them if not on-by-default).

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -1006,9 +1006,8 @@ class AnsibleModule(object):
         - executable (string) See documentation for subprocess.Popen().
                               Default is None.
         '''
-
         shell = False
-        if isinstance(args, list):
+        if isinstance(args, list) or isinstance(args, tuple):
             if use_unsafe_shell:
                 args = " ".join([pipes.quote(x) for x in args])
                 shell = True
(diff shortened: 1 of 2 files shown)


mpdehaan:
This looks decent, we need to make sure we test on other OSes to make sure it doesn't cause any side effects there obviously.

Thanks!

charles-dyfis-net:
Consider this PR temporarily withdrawn. There's a brown-paper-bag bug in here, but I want to test the fix before publishing it.

charles-dyfis-net:
Consider this no longer withdrawn (as of bdb5af4).

If it's going to be held back for testing, I'd be tempted to be a bit more aggressive and make check_rc true for all enable/disable calls, rather than systemd only (as done here to reduce chance of breakage on other systems).

charles-dyfis-net:
The mechanism by which netctl services are invoked by systemd is not reliably compatible with the mechanism Ansible currently uses to check whether systemd services are enabled; enabled-and-started services are correctly detected as enabled when this patch is applied, but enabled-but-not-started services are not. Withdrawing this PR until a more robust solution is available (which may involve a patch to netctl).

PR #6496, opened 2014-03-14
Imply +x permissions on readable directories by deafult

hkariti:
This enabled recursively changing the mode of a directory without making all the files executable. Also added an `implied_direxec` option to the common_file_option that is on by default, if one wishes to disable this behavior.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -109,6 +109,7 @@ FILE_COMMON_ARGUMENTS=dict(
     serole = dict(),
     selevel = dict(),
     setype = dict(),
+    implied_direxec = dict(default='true', choices=BOOLEANS),
     # not taken by the file module, but other modules call file so it must ignore them.
     content = dict(),
     backup = dict(),
@@ -423,7 +424,7 @@ class AnsibleModule(object):
             changed = True
         return changed
 
-    def set_mode_if_different(self, path, mode, changed):
+    def set_mode_if_different(self, path, mode, changed, exec_read=False):
         path = os.path.expanduser(path)
         if mode is None:
             return changed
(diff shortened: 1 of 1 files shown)


amenonsen:
I don't really understand this patch. I get the idea that it copies r bits to x bits in the mode unless you set "implied_direxec" (whose name I don't like), but I don't understand where/how it's used, or what the consequences are. Even if it's not a user-visible option, but only a module-visible one, given its effect on permissions, I think a few extra comments in the source are justified.

The merge conflicts should be easy enough to resolve.

hkariti:
don't like the name either, but didn't have a better idea. Ideas are welcome.

regarding comments - i didn't understand what you didn't understand. Can you clarify so I can add some comments about that?

amenonsen:
"This enabled recursively changing the mode of a directory without making all the files executable": so, how do I recursively change the mode of a directory without making all the files executable? Who is this feature for, and how can they use it?

hkariti:
Currently, you can't set the mode of a directory recursively so it's accessible without marking all of its files as executable.

If you do this:

```
- file: path=/some/directory mode=444 recurse=yes
```

The directory isn't accessible as it's missing the `x` permission.

If you do this:

```
- file: path=/some/directory mode=555 recurse=yes
```

All of the files are marked executable.

This patch automatically adds the `x` permission to a directory if the mode given has the `r` permission, so doing the first line (`mode=444`) will now correctly also add `x` to directories (i.e. dirs will have mode `555` and files `444`).

The added parameter is used to disable this behavior:

```
- fie: path=/some/directory mode=444 recurse=yes implied_direxec=false
```

Will set the mode of the directory to `444` instead of `555`.

amenonsen:
@hkariti thanks for the explanation.

After thinking about it for a couple of days, this feels a bit too magic to me. What would you think of leaving the current behaviour alone and adding a `dirmode=NNN` parameter that lets you specify a separate mode for any directories, and which can be different from the file mode? I think that would be easier to explain than the implicit r→x duplication, and more usable (easier to understand the effect of, more flexible). In any case, this change definitely needs a documentation update.

hkariti:
The concept of dirmode is also used in other places as well (smbmount and
stuff) so yeah it is better.

gregdek:
Ping @hkariti -- just wondering how you intend to proceed here, given the discussion with @amenonsen above. New PR? Refactor and rebase of this one? Let us know.

hkariti:
I'll send a new PR

PR #6497, opened 2014-03-14
Feature/multi version import

abadger:
multi-version import that lets us use a newer-than-the-default pycrypto package.

This implementation replaces the distutils setup function in setup.py with the one from setuptools.  The advantage is that we're then able to specify which version of pycrypto we need in setup.py and that information is then pushed out to the egg-info in a manner that setuptools can read.  This means that we can then set each script to require ansible and setuptools' (its pkg_rsources module) will look up the deps for ansible and make sure that those are installed and the correct version.

If desired, this implementation could be further enhanced so that different scripts could require different dependencies: http://pythonhosted.org/setuptools/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies

### DIFF
--- a/bin/ansible-playbook
+++ b/bin/ansible-playbook
@@ -18,6 +18,9 @@
 
 #######################################################
 
+__requires__ = ['ansible']
+import pkg_resources
+
 import sys
 import os
 import stat
--- a/bin/ansible-vault
+++ b/bin/ansible-vault
@@ -20,6 +20,9 @@
 # example playbook to bootstrap this script in the examples/ dir which
 # installs ansible and sets it up to run on cron.
 
+__requires__ = ['ansible']
+import pkg_resources
+
 import os
 import sys
 import traceback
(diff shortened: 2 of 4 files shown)


abadger:
https://github.com/ansible/ansible/pull/6498 is an alternative to this one.  Pick the one that works best for you.

mpdehaan:
Yeah this is fine -- there is a contigency that wants setuptools anyway.  All good.

As we haven't required setuptools to build setup.py things in the past it looks like we'll want to try/except importerror around setuptools and tell people how to install it, and probably also add some BuildRequires to the RPM spec bits.

abadger:
Couple commits to start addressing those issues.  Not sure what more you might want the message about setuptools not being available at build time to say, though.

jimi-c:
@abadger so is this pull request mutually exclusive with #6498, since it has been merged in already?

mpdehaan:
@jimi-c we are taking this one, not the other one, the other has not been merged.

mpdehaan:
Looks like the branches on the other have changed, we do want this, neither has been merged so this still needs to go in for 1.6

mpdehaan:
The message about setuptools being required to run should probably be modified slightly to tell non-technical users how to install setuptools, i.e. "pip install setuptools"

jimi-c:
Merged, thanks!

PR #6498, opened 2014-03-14
Use setuptools to get a recent enough version of python-crypto2.6

abadger:
This is an alternative means of doing multi-versions to what's in pull request https://github.com/ansible/ansible/pull/6497

This implementation does not require changing from distutils setup to setuptools setup in the setup.py file.  The tradeoff is that in this implementation, any script that ansible ships that needs a non-default version will need to have those dependencies listed in the `__requires__` in the script (instead of being able to specify a package-wise "ansible" dep as in PR:6497)

### DIFF
--- a/bin/ansible-playbook
+++ b/bin/ansible-playbook
@@ -18,6 +18,9 @@
 
 #######################################################
 
+__requires__ = ['pycrypto >= 2.6']
+import pkg_resources
+
 import sys
 import os
 import stat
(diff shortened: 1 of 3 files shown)


mpdehaan:
+1 (pending a bit more testing)

we'll probably need to update docs about install reqs.

mpdehaan:
Thanks for this, I think we'll go with the setuptools one.

PR #6501, opened 2014-03-15
fixed bad detection of change on default and premature break

bcoca:
when using 'default' option module always flagged as 'changed', fixed the detection as per
https://groups.google.com/forum/#!msg/ansible-project/WVD371ocufo/pH2FD5aVCdsJ

### DIFF
--- a/library/files/acl
+++ b/library/files/acl
@@ -115,6 +115,9 @@ def split_entry(entry):
         print "wtf?? %s => %s" % (entry,a)
         raise e
 
+    if d:
+        d = True
+
     if t.startswith("u"):
         t = "user"
     elif t.startswith("g"):
@@ -248,7 +251,6 @@ def main():
                         if not old_permissions == permissions:
                             changed = True
                         break
-                break
         if not matched:
             changed=True
 


mpdehaan:
merged, thanks!

astorije:
Hi @mpdehaan, has this fix been shipped?

I am running Ansible 1.6.3 (host being Ubuntu 13.10, guest being Debian 7.5 under Vagrant) and the following is always marked as `changed`:

```
- acl: >
    name=/home/vagrant/test
    entity=jeremie
    etype=user
    permissions=r
    state=present
```

Note that `/home/vagrant/test` is a basic, empty file created using `touch test` and that my playbook actually only runs the task described above.

I manually checked that `getfacl /home/vagrant/test` outputs the same thing between 2 executions of the task.

Let me know if this look like something unrelated and I'll open a new issue at your will.

astorije:
Actually, I noticed that using `permissions=r--` rather than `permissions=r` fixes this. Looking at `man setfacl` (which accepts the `r` form), the documentation of the module and the code of the module, I think this is actually a bug.

@mpdehaan can you confirm this please? It seems to me that the permissions lacks normalization in the code, let me know if you want me to come up with a fix.

bcoca:
i never intended for you to need the --, if you look at the examples
permissions="r" is in there, I'll check the code.

bcoca:
when using the entity parameter, the module adds - when needed for the
comparison, but does nothing with permissions parameter, should be easy to
solve.​

astorije:
Hey @bcoca thanks for the quick answer!
I'll have a look to fix that over the weekend then.

astorije:
You meant entry parameter rather than entity parameter right?
Anyway, that's what I gather when looking at the code too, so I'll come up with the fix.

bcoca:
yes, that is what I meant, I tested with your example and both reproduced the issue and it's absence after the patch.

astorije:
Hey well, you beat me to it! I was not on the laptop so I could not submit the patch but I essentially figured out the same thing. Thanks for being so responsive!

PR #6503, opened 2014-03-15
Fix docs for playbooks_intro

quiver:
remove unnecessary period.

Corresponding url is here:
http://docs.ansible.com/playbooks_intro.html#tips-and-tricks


mpdehaan:
merged, thanks!  This will appear next time we push docs!

Issue #5004, opened 2013-11-22
A file: state=directory that tolerates links


rektide:
state=directory dies if the thing that's there is a link. it's possible to add control flow to playbooks to tip toe around this failure, but it would be great to have a means baked in that will permit a symlink (if it points to a directory).

mpdehaan:
Can you please try to explain this better?

Perhaps an example playbook or rewording the above, or perhaps steps to reproduce this problem?

Thanks!

bcoca:
It looks like a follow=true option that tests the target to see if its a
symlink, then follows it until a non symlynk source and then apply the file
command to that (default follow=false, would keep module working as is).

rektide:
Rolling with what bcoca has put forth, this would pass:

```
file: path=/tmp/directory state=directory
file: src=/tmp/directory dest=/tmp/dirlink state=link
file: path=/tmp/dirlink state=directory follow=true
```

mpdehaan:
Understood, pull requests for this seem fine to me, but I worry that people would expect consistency for the copy and template (and other) modules as well, so probably needs to leverage some intelligence in module_common code.

mpdehaan:
@bcoca is working on some consistency cleanup for state=link and so forth right now, which should take care of this, so I'm going to close out this ticket.

PR #6506, opened 2014-03-15
Fixes and cleanup to file functions and module

bcoca:
- unified set attribute functions ... not sure why 2 identical functions
  exist with diff names, now there are 3 while i repoint all modules to 1
- fixed issue with symlinks being created w/o existing src when force=no
- refactored conditionals, simplified where possible
- added tests for symlink to nonexistant source, with both force options
- made symlink on existing attomic (force)

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -464,7 +464,7 @@ class AnsibleModule(object):
                 changed = True
         return changed
 
-    def set_file_attributes_if_different(self, file_args, changed):
+    def set_fs_attributes_if_different(self, file_args, changed):
         # set modes owners and context as needed
         changed = self.set_context_if_different(
             file_args['path'], file_args['secontext'], changed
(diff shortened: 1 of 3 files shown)


bcoca:
this does not fix #5004, which is really a feature request to add 'follow' parameter so that state=file|directory|hard|absent are applied to the target of a symlink instead of the symlink if that is what 'path' points at.

jimi-c:
Merged, thanks!

muffl0n:
I think this change broke the symlink creation in the file module. Strangely this works when issuing the playbook for "localhost" but not for a remote host.

test-link.yml:

```
- hosts: remote
  tasks:
    - file: 
        src=/tmp/testuser
        path=/tmp/bar 
        state=link
```

/tmp/testuser exists

Output:

```
$ ansible-playbook -i test test-link.yml 

PLAY [remote] ***************************************************************** 

ok: [remote]

TASK: [file src=/tmp/testuser path=/tmp/bar state=link] *********************** 
failed: [remote] => {"failed": true, "parsed": false}
  File "/root/.ansible/tmp/ansible-tmp-1395311657.23-178706096573480/file", line 1490, in <module>
    main()
  File "/root/.ansible/tmp/ansible-tmp-1395311657.23-178706096573480/file", line 308, in main
    os.unlink(tmppath)
OSError: [Errno 2] No such file or directory: './tmp/bar.7317.1395311651.71.tmp'


           to retry, use: --limit @/home/svs/test-link.retry

remote                     : ok=1    changed=0    unreachable=0    failed=1
```

The same for localhost:

```
$ ansible-playbook -i test test-link.yml 

ok: [localhost]

TASK: [file src=/tmp/testuser path=/tmp/bar state=link] *********************** 
changed: [localhost]

```

PR #6507, opened 2014-03-15
Set keepalive to 5 seconds in paramiko_ssh.py

### LINKED ISSUE #6476: Request to set_keepalive in paramiko_ssh
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.5
##### Environment:

Running ansible on Ubuntu 10.04 64-bit 2.6.32-48-generic
managing client on Ubuntu 10.04 64-bit 2.6.38-16-server
##### Summary:

During a package install via ansible's apt module the package I'm installing reconfigures iptables which momentarily causes a disruption in ansible's SSH session  The result of this causes ansible to hang indefinitely and never finish the apt command. My client I'm trying to configure is Ubuntu 10.04. In order to fix this I added line 188 below to runner/connection_plugins/paramiko_ssh.py

187:            chan = self.ssh.get_transport().open_session()
188-            self.ssh.get_transport().set_keepalive( 5 )

Is there another way to do this using ansible.cfg or ssh config, or would ansible care to incorporate this change?
##### Steps To Reproduce:

The package I'm installing is software I cannot share, however I believe the root of the issue is iptables reconfiguration during an SSH session.
##### Expected Results:

I expect the ansible apt command to exit promptly after completing the action.  For some reason the session doesn't close unless I set_keepalive.
##### Actual Results:

The ansible apt command hung indefinitely and never exited.  NOTE: This is not a problem for every other package I've installed, so its definitely unique the package I'm installing.

brdo:
This update resolves https://github.com/ansible/ansible/issues/6476.

### DIFF
--- a/lib/ansible/runner/connection_plugins/paramiko_ssh.py
+++ b/lib/ansible/runner/connection_plugins/paramiko_ssh.py
@@ -185,6 +185,7 @@ class Connection(object):
         bufsize = 4096
         try:
             chan = self.ssh.get_transport().open_session()
+            self.ssh.get_transport().set_keepalive(5)
         except Exception, e:
             msg = "Failed to open session"
             if len(str(e)) > 0:


mpdehaan:
In queue for testing, thanks!  (note: we'll want to test carefully to make sure this doesn't affect anything, seeing transport is somewhat critical)

risaacson:
I have started looking into this. I don't see the similar behavior in how the ssh transport works. I will dig a bit deeper if this is the best way to implement the behavior that you are looking for.

jimi-c:
Merged, thanks!

PR #6508, opened 2014-03-15
added gathering control to ansible, 

bcoca:
As talked here:
https://groups.google.com/forum/#!msg/ansible-project/lGrvGwKuENU/-MNfmPBUHY4J

Before 1.5 fact gathering was 'magically smart' if you had gathered facts and you did not explicitly force fact gathering ansible would reuse the cache and avoid running fact gathering again. Striving for correctness and 'less magic' and unintended consequences, this was removed in 1.5.

This adds it back making it more visible and controllable by users through configuration.

### DIFF
--- a/examples/ansible.cfg
+++ b/examples/ansible.cfg
@@ -24,6 +24,11 @@ transport      = smart
 remote_port    = 22
 module_lang    = C
 
+# controls implicit fact gathering (always, never or smart).
+# smart gathers only if not currently in memory.
+# does NOT affect explicit 'gather_facts' entries.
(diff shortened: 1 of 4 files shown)


mpdehaan:
Some discussion towards later end of thread from me on naming and assignment of policy to the variable, but general concept is good.

mpdehaan:
Looking at this now.

mpdehaan:
merged, thanks!

I changed the default policy to 'implicit' per the mailing list thread, though the old behavior is now accessible by choosing 'smart'.

PR #6512, opened 2014-03-16
Add a JIRA control module for Ansible. 

tarka:
This adds a module to create and manipulate issues in a JIRA instance. This version uses fetch_url rather than urllib2.


mpdehaan:
merged, thanks!

PR #6513, opened 2014-03-16
Fix acl module doc and  error messages

lotia:
- Fix typos in module doc
- Provide clearer error message when entry and etype are both missing


lotia:
Closing because an executable bit was incorrectly set due to editor issues.

PR #6514, opened 2014-03-16
Make parsing variables for hosts and groups more consistent

dmage:
This commit allows to specify arrays in [group:vars] block.

```
[group1]
example1.com targets="['aaa', 'bbb']"  # this is array

[group2]
example2.com
[group2:vars]
targets=['aaa', 'bbb']  # for now, this is string
```


mpdehaan:
I'm not crazy about these being inline vs using group_vars/ and host_vars/ but this is a reasonable cleanup as well, and seems enough folks do want something like this.

So, I won't stand in the way of progress!

mpdehaan:
merged, thank you!

PR #6515, opened 2014-03-16
Fix acl doc and error

lotia:
Small doc and error message fixes in acl module:
- Fix typo in docstring
- Correct error message displayed when entry and etype are both set
- Remove extra whitespace in check to determine if etype or entity or 
  permissions are set when entry is set.


mpdehaan:
merged, thanks!  This will appear the next time docs are pushed.

PR #5222, opened 2013-12-09
apt: update index before trying to autoinstall python-apt

resmo:
Minor fix to make sure index is updated.

PR #6518, opened 2014-03-17
fix: bug in apt module when python-apt is not available

spiroid:
##### Summary:

When python-apt is not installed the apt module tries to use apt-get, update apt-cache and install python-apt. But due to pull request #5222 modifications and use of the unsafe string '&&' the proper command was never executed.

This fix is minimal and uses the use_unsafe_shell=True parameter to make it finally work.
Any input appreciated, 

Regards, 

@Spiroid & @Ssaboum

### DIFF
--- a/library/packaging/apt
+++ b/library/packaging/apt
@@ -377,7 +377,7 @@ def main():
 
     if not HAS_PYTHON_APT:
         try:
-            module.run_command('apt-get update && apt-get install python-apt -y -q')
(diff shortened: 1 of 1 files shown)


mpdehaan:
merged, thank you!

jlaska:
I've also added an integration test to assert that automatic installation of python dependencies works properly for both the `apt` and `yum` modules (4ef2603a6226e68db4ff50709abd5bb7cf92e064).

PR #6521, opened 2014-03-17
fetch_url: Avoid credential stripping for FTP-scheme URLs

gdvalle:
This fixes the get_url module to work with FTP URLs that contain credentials.

The patch avoids entering logic that strips credentials to place into urllib2's HTTPPasswordMgr store if the URL scheme is FTP.


mpdehaan:
merged, thanks!

PR #6523, opened 2014-03-17
Pass through vault_password when parsing host/group vars as directories.

srgvg:
Fixes a bug where vault_password parameter was not passed through in
`_load_vars_from_folder()`

Encrypted files in 
- group_vars/$groupname/some.yml
- host_vars/$hostname/other.yml
  could not be decrypted because of this missing parameter.

modified:   lib/ansible/inventory/vars_plugins/group_vars.py


jctanner:
@sergevanginderachter thank you!

PR #6524, opened 2014-03-17
service module: Properly disable Debian services

tyll:
Services on Debian need to be disabled with 'disable' instead of 'remove'
to avoid them being enabled again when 'update-rc.d $service defaults' is run,
e.g. as part of a postinst script.


mpdehaan:
merged, thanks!

Issue #6499, opened 2014-03-14
git module fails with "failed to stat '.': Permission denied"


yuriks:
##### Issue Type:

Bug report
##### Ansible Version:

`ansible-playbook 1.5.3`
##### Environment:

Both host and slave are running Ubuntu Server 12.04LTS
##### Summary:

This is the same issue as #6443 which doesn't seem to be fixed as of 1.5.3.

```
yuriks@glide-dev:~/src/glide-server/ansible$ ansible-playbook --version
ansible-playbook 1.5.3
yuriks@glide-dev:~/src/glide-server/ansible$ ansible-playbook -i yuriks.inventory site.yml

PLAY [glide-servers] **********************************************************

GATHERING FACTS ***************************************************************
ok: [provision]

[~~~snip~~~]

TASK: [cm | check out glide-cm] ***********************************************
failed: [provision] => {"cmd": ["/usr/bin/git", "ls-remote", "git@github.com:digEcor/glide-cm.git", "-h", "refs/heads/develop"], "failed": true, "item": "", "rc": 128}
stderr: fatal: failed to stat '.': Permission denied

msg: fatal: failed to stat '.': Permission denied

FATAL: all hosts have already failed -- aborting

PLAY RECAP ********************************************************************
           to retry, use: --limit @/home/yuriks/site.retry

provision                  : ok=9    changed=0    unreachable=0    failed=1
```

The playbook is being run as root, but the git step has sudo and sudo_user fields to run as an unprivileged user. What seems to be happening is that the `git` command is being run with a cwd of `/root`, which isn't readable by the unprivileged user. Git tries to stat it and fails.
##### Steps To Reproduce:

Try to clone a new git repository using sudo.
##### Expected Results:

Cloning to succeed to the unprivileged user-owned directory.
##### Actual Results:

Git tries to read a directory it doens't have access to and aborts.

mpdehaan:
Thanks, we'll take a look this very soon!

jctanner:
@yuriks Can you show me the task from the playbook? 

I'm unable to reproduce based on my current understanding of the description:

```
- hosts: el6.lab.net
  #connection: local
  gather_facts: False
  user: jtanner
  tasks:
    - shell: whoami
    - shell: whoami
      sudo: True
    - shell: whoami
      sudo: True
      sudo_user: jtanner

    - git: repo=https://github.com/jctanner/ansible-pull-test  dest=/tmp/1
[...]

PR #6527, opened 2014-03-17
Fix for hostkey directory

patrickheeney:
I am using the new `accept_hostkey=yes` on the `git` module and I am getting the following error: `msg: /root/.ssh/ does not exist`. I tracked this down to the `add_host_key` method. 

The environment I am running this in looks like this:

```
        "SUDO_USER": "root",
        "USER": "vagrant",
        "USERNAME": "vagrant"
```

So it should not be looking into the `/root/.ssh` because I do not use `sudo` on the playbook or the play. I checked the `not_in_host_file` method and it looks within the `${USER}` known_hosts but when adding it, it does not. 

This solution will first check the USER environment variable or fallback to the old way, similar to the other methods. I have not been able to fully test this yet. Also, this could use some regression tests.

### DIFF
--- a/lib/ansible/module_utils/known_hosts.py
+++ b/lib/ansible/module_utils/known_hosts.py
@@ -119,13 +119,21 @@ def add_host_key(module, fqdn, key_type="rsa"):
     result = False
     keyscan_cmd = module.get_bin_path('ssh-keyscan', True)
 
-    if not os.path.exists(os.path.expanduser("~/.ssh/")):
-        module.fail_json(msg="%s does not exist" % os.path.expanduser("~/.ssh/"))
+    if 'USER' in os.environ:
+        user_ssh_dir = os.path.expandvars("~${USER}/.ssh/")
+        user_host_file = os.path.expandvars("~${USER}/.ssh/known_hosts")
+    else:
+        user_ssh_dir = "~/.ssh/"
+        user_host_file = "~/.ssh/known_hosts"
+    user_ssh_dir = os.path.expanduser(user_ssh_dir)
+
+    if not os.path.exists(user_ssh_dir):
+        module.fail_json(msg="%s does not exist" % user_ssh_dir)
 
     this_cmd = "%s -t %s %s" % (keyscan_cmd, key_type, fqdn)
 
     rc, out, err = module.run_command(this_cmd)
-    module.append_to_file("~/.ssh/known_hosts", out)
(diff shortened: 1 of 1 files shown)


patrickheeney:
I made an attempt at the integration tests, but unable to test them because of #6528 so I do not know if they work. I attempted to change the user to root during a play so that it would error `/root/.ssh/ does not exist`. Then attempt to use a regular user and check that the host was added to `~/.ssh/known_hosts` and not `/root/.ssh`.

```
- name: checkout ssh://git@github.com repo with accept_hostkey and root user (expected fail)
  git:
    repo: '{{ repo_format2 }}'
    dest: '{{ checkout_dir }}'
    accept_hostkey: true
    key_file: '{{ github_ssh_private_key }}'
  remote_user: root
  register: git_result
  when: github_ssh_private_key is defined
  ignore_errors: true

- assert:
    that:
      - 'git_result.failed'
      - 'git_result.msg == "/root/.ssh/ does not exist"'

- name: checkout git@github.com repo with accept_hostkey and regular user (expected pass)
  git:
    repo: '{{ repo_format2 }}'
    dest: '{{ checkout_dir }}'
    accept_hostkey: true
    key_file: '{{ github_ssh_private_key }}'
  remote_user:  test
  register: git_result
  when: github_ssh_private_key is defined

- assert:
    that:
      - 'git_result.changed'
  when: not git_result|skipped
```

mpdehaan:
So this should be thoroughly cross-referenced with https://github.com/ansible/ansible/issues/6499

mpdehaan:
FWIW, I'm  uncomfortable with https://github.com/patrickheeney/ansible/commit/a9aab23f1132d14d12abb1adf4e7148f254c630f due to past experience with USER being defined unpredictably between distros/shells.  What works on Ubuntu cannot be assumed elsewhere, etc.

patrickheeney:
@mpdehaan Fair enough, however that line was taken from https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/known_hosts.py#L73-L77 which runs right before the method I modified, so the logic should probably be updated here as well.

If you have an alternative suggestion on how to fix I would be happy to experiment and try to modify the PR.

mpdehaan:
Yeah on second thought that's not different and is probably ok.   Looking again...

mpdehaan:
Merged, thanks!

PR #6529, opened 2014-03-17
Remove dependency on pycurl in apt_repository

sivel:
This pull request removes the dependency on `python-pycurl` for `apt_repository`, by replacing the pycurl calls with `fetch_url` from `module_utils/urls.py`

pycurl was historically used to verify SSL certs, and now that fetch_url exists and can handle this, there probably is no reason to keep the dependency on pycurl.

I've not added the `validate_certs` attribute, since we were forcing SSL cert verification anyway previously.

I've run the test_apt_repository integration tests on an ubuntu 12.04 box, and everything appears to be passing.

### DIFF
--- a/library/packaging/apt_repository
+++ b/library/packaging/apt_repository
@@ -31,7 +31,6 @@ notes:
     - This module works on Debian and Ubuntu and requires C(python-apt).
     - This module supports Debian Squeeze (version 6) as well as its successors.
     - This module treats Debian and Ubuntu distributions separately. So PPA could be installed only on Ubuntu machines.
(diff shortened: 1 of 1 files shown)


mpdehaan:
Excellent, I now must construct an extra statue of @sivel

sivel:
A calvary of sivels!

mpdehaan:
I am thinking several different types of mounts, including perhaps a gryffon, ostrich, and walrus.   Perhaps in a Merry-Go-Round formation.

mpdehaan:
Merged, thank you!

Issue #5206, opened 2013-12-06
shell module: Shell echo is including arguments


patrickheeney:
Task:

```
- shell: "echo -e -n \"test\n\ttest2\" >> test creates=test"
```

Results:

```
<localhost> EXEC ['/bin/sh', '-c', 'mkdir -p $HOME/.ansible/tmp/ansible-1386356406.38-102382384091680 && echo $HOME/.ansible/tmp/ansible-1386356406.38-102382384091680']
<localhost> REMOTE_MODULE command echo -e -n "test
    test2" >> test creates=test #USE_SHELL
<localhost> PUT /tmp/tmp188keY TO /home/vagrant/.ansible/tmp/ansible-1386356406.38-102382384091680/command
<localhost> EXEC ['/bin/sh', '-c', '/usr/bin/python /home/vagrant/.ansible/tmp/ansible-1386356406.38-102382384091680/command; rm -rf /home/vagrant/.ansible/tmp/ansible-1386356406.38-102382384091680/ >/dev/null 2>&1']
```

Outputs:

```
-e -n test
    test2
```

Expected Result:
The arguments -e or -n should not be in the file.

mpdehaan:
We can take a look to see if we think there is something wrong with the escaping.

First time I've heard of it though.

janeznemanic:
I will have a look at this one and try to figure out something.

patrickheeney:
@janeznemanic It seems to be a bigger issue then just shell and echo:

```
- name: main | check for db existance
  command: >
    mysql -e "show databases" | fgrep {{ item }}
  register: mysql_exist
  with_items:
    - app
```

From this I get the response: `Usage: mysql [OPTIONS] [database]`. So it is not receiving the arguments. It is passing the right information to the remote, something is happening after that.

janeznemanic:
It looks like I can't reproduce the bug. Here is my sample playbook:

 hosts: localhost
  gather_facts: false
  connection: local
  tasks:
- name: debbuging
  
  shell: "echo -e -n \"test\n\ttest2\" >> test creates=test"

The command that I run is: ansible-playbook -i inventory -vv play.yml
Console output is:

PLAY [localhost] *************************************************************\* 

TASK: [debbuging] ************************************************************\* 
<localhost> REMOTE_MODULE command echo -e -n "test
    test2" >> test creates=test #USE_SHELL
[...]

PR #6535, opened 2014-03-17
Add test to demonstrate issue

patrickheeney:
This test demonstrates issue #5206 in my environment. If you have vagrant installed, I can share my configuration with you so that you can easily bring up an environment to see and test this in. Let me know.

It also happens on other commands with arguments, but I opted for `echo` since its easiest to test versus something like `mysql -e "show databases" | fgrep {{ item }}` which ignores the `-e`.

### DIFF
--- a/test/integration/roles/test_command_shell/tasks/main.yml
+++ b/test/integration/roles/test_command_shell/tasks/main.yml
@@ -168,3 +168,25 @@
   assert:
     that:
       - "shell_result4.changed == False"
+
+# shell arguments
+
+- name: verify that test is absent
+  file: path={{output_dir_test}}/test_argument state=absent
+
+- name: echo to file to verify arguments are not passed
+  shell: "echo -e -n \"test\n\ttest2\" >> {{output_dir_test}}/test_argument creates={{output_dir_test}}/test_argument"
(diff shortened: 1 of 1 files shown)


sivel:
I believe we have this covered in the unit tests as well.

```
FAIL: test_run_command_string_unsafe_with_pipe (TestModuleUtilsBasic.TestModuleUtilsBasic)
  File "/Users/matt/python_venvs/ansibledev/ansible/test/units/TestModuleUtilsBasic.py", line 100, in test_run_command_string_unsafe_with_pipe
    self.assertEqual(out, 'foo bar')
AssertionError: '-n foo bar\n' != 'foo bar'

```

See also https://github.com/ansible/ansible/issues/6539#issuecomment-37885996

This seems to be an Mac issue.

patrickheeney:
I am running this within a VM with ubuntu 12.04. I don't have ansible installed on my mac.

sivel:
Ok, ignore me then, I may have misunderstood, and these may be completely unrelated issues :)

jctanner:
@patrickheeney Let's close this and try to sort out what is going on with the mysql command in a new bug.

Issue #6490, opened 2014-03-14
[Feature Idea] Ansible Galaxy - GalaxyFile


patrickheeney:
##### Issue Type:

Feature Idea
##### Ansible Version:

1.6+
##### Environment:

N/A
##### Summary:

This feature was originally proposed on [google groups](https://groups.google.com/forum/#!topic/ansible-devel/7AVJJL7_LEU) and will be summed up here for tracking purposes. 

At a high level this will provide the same functionality that a puppetfile does to librarian-puppet. In ansible terms this would be a pure YAML based document that contains instructions for ansible-galaxy to download, install, and update roles. This would allow for an familiar and convenient way to manage your galaxy roles, private / public / forked git roles, or even ones located on the file system. The galaxyfile could look something like this:

```
roles:
   - { source: galaxy, name: "foo.bar_role", version: "1234" }
   - { source: git, repo: "...", version: "1234" }
   - { source: file, path: "..." }
```

Upon running a command like `ansible-galaxy install -f galaxyfile.yml`, ansible galaxy would run through the galaxyfile and install the relevant versions from galaxy, git, or on your local file system. The `ansible-galaxy update` command would re-run this process and update any roles that are not tied to specific version.
##### Notes on Implementation
- Ansible GIT module should have all support functions moved to `module_utils/git.py` so that galaxy can use them. Look at basic.py or ec2.py for implementation on existing modules.
- Ansible galaxy when using a galaxy based role can hit the galaxy API to first check if an update is needed before downloading and installing. 
- The GIT based version would require git to be installed on the local machine in order to clone and checkout the version we want. Updates would be a simple git pull. 
- The file based version should could verify the checksum and cache is locally so on future updates it does not need to be re-installed. Look to leverage some functionality from the unarchive module.
- Ansible Galaxy `-f` option should remain backwards compatible with the comma delimitated format. It could just detect a `.yml` extension to load the new role parser.
##### Steps To Reproduce:

N/A
##### Expected Results:

N/A
##### Actual Results:

N/A

mpdehaan:
So yes, this was posted at Google Groups and had assumed you were building -- as this should be the case and I don't want to fragment the discussion, I don't think we need the feature ticket here in the issue tracker -- let's continue discussion on the list and if you want to submit a pull request we can discuss any details on that too.

Thanks!

patrickheeney:
@mpdehaan, The last reply on the google groups discussion was from James Cammarata that said:

[...]

PR #6537, opened 2014-03-17
Galaxy CLI requirements file improvements

patrickheeney:
This is the PR for #6490. **This is not ready and strictly interested in feedback and implementation suggestions**. 

This is my first time programming in python so please enlightenment me on better ways to do things. This may not work right away, I have been rebasing with different versions so it could be in a different state. I have made a note about why I have done some things below:
1. All the GIT functions were necessary since the only other place that has them is the GIT module and they require the module to be passed in as arguments, which does not work in this instance.
2. I had to modify `basic.py` because I needed access to the `get_bin_path` functions. I moved them to what seems like global state and use the class function to call the global ones.
3. This should be backwards compatible because I convert the old format which is `name,value` into the new format before processing it. Need to test it.

Usage:

```
ansible-galaxy -r test.yml
```

test.yml (need to add a real world test file here, this is a sample to show implementation):

```
# galaxy

- source: galaxy
  name: test/module
  version: 1234

- name: test/module

- { source: galaxy, name: test/module }

# git

- source: git
  name: git_role
  repo: 'git@github.com:jctanner/ansible-pull-test.git'
  dest: ansible/roles
  remote: origin
  depth: 1
  version: 1234
  executable: /path/to/git

- { source: git, repo: 'git@github.com:jctanner/ansible-pull-test.git' }

# file

- source: file
  name: file_role
  path: file_role.tar.gz

- name: test/module.tar.gz

- source: file
  name: file_role
  path: https://github.com/ansible/ansible/archive/v1.5.3.tar.gz

```

Todo:
- [ ] Testing / Fixing bugs. This was a rough go at it, I did not test a lot of variations, just merely got it git cloning. 
- [ ] Better errors. I was not sure how to handle the error messages when not within a module. I have seen some scripts `print >>sys.stderr`, others `print "..." sys.exit(1)`, and finally `exit_without_ignore(options)`. Once I know how to handle this I will clean this up.
- [ ] Git improvements. If the `name` option in git is blank it should pull it from the repo name.
- [ ] File improvements. Add support for `zip` and some other file extensions if it does not have this already. Add md5_checksum verification. 
- [ ] Remote file. Right now `file:` I believe only works locally. Would be nice to fetch this from a remote and install it locally. For example if you wanted to link to a tar.gz hosted by github. 
- [ ] Unit tests? I did not see anywhere that galaxy is tested. I thought it could use some tests to make sure its parsing the files correctly. Maybe this is not needed right now.

I am open to any and all suggestions. Thanks for reading!

### DIFF
--- a/bin/ansible-galaxy
+++ b/bin/ansible-galaxy
@@ -3,6 +3,7 @@
 ########################################################################
 #
 # (C) 2013, James Cammarata <jcammarata@ansible.com>
+# (C) 2013, Patrick Heeney <patrickheeney@gmail.com>
 #
 # This file is part of Ansible
 #
@@ -37,6 +38,7 @@ from collections import defaultdict
 from distutils.version import LooseVersion
 from jinja2 import Environment
 from optparse import OptionParser
+from ansible.utils import cmd_functions
 
 import ansible.constants as C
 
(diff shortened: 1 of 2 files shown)


mpdehaan:
So this part:
    - source: git
      name: git_role
      repo: 'git@github.com:jctanner/ansible-pull-test.git'
      dest: ansible/roles
      remote: origin
      depth: 1
      version: 1234
      executable: /path/to/git

Strikes me as a little verbose.

Seems source/name/repo is mostly what's needed most of the time, plus sometimes version.

It should not need to specify "dest" for any of these at this comes from the role name, which should be based on the repo name?

bcoca:
won't this break most modules?

mpdehaan:
So this is about the requirements file for Galaxy supporting an additional YAML option that allows for git sources as well as galaxy sources, not sure specifically what part you mean would break

mpdehaan:
I see that's mostly available above, so ignore my first comment

patrickheeney:
@mpdehaan I made that particular one verbose to show all the options. You can accomplish the same with `- { source: git, repo: 'git@github.com:jctanner/ansible-pull-test.git' }`. The destination is only needed if you need a custom folder (In my code base I have a secondary roles folder that galaxy would not know about otherwise). It will figure it out for you if left blank, but that option overrides it. 

@bcoca This is my first time working with python. If something would break most modules can you please point out where and how? These changes should be contained strictly to ansible-galaxy so I don't understand how this would do anything like that.

bcoca:
the break will come moving the get_bin_path outside of the module object,
all core modules subclass this and those that access that function/method
do so through the object they subclass.

patrickheeney:
I moved it outside the module object but the class still has the methods, they just forward on the call. See [here](https://github.com/ansible/ansible/pull/6537/files#diff-90085fdcec6ed8b273ba885eaee60328R865) and [here](https://github.com/ansible/ansible/pull/6537/files#diff-90085fdcec6ed8b273ba885eaee60328R912).  I was attempting to make those functions global and have the class functions pass the call onto the global one. From my tests and running integration tests, everything appeared to be fine. But I have been working with python for only 5 hours so I don't know for sure.

I am also open to reverting that change completely but I need some guidance on where the functions can be duplicated to so that i can query for a bin path.

bcoca:
sorry, i had not looked at patch, just read your comment, that should work
just fine.

geerlingguy:
+1

amenonsen:
This PR also needs to wait for the Galaxy code in v2 to stabilise before it can be revisited.

amenonsen:
I'm not familiar with galaxy, but 2.0 is out now. Is this patch still relevant? cc @chouseknecht 

@patrickheeney Sorry it's taken so long, but are you still interested in getting this polished up and merged?

patrickheeney:
It looks like galaxy may have this functionality now: http://docs.ansible.com/ansible/galaxy.html#advanced-control-over-role-requirements-files . It doesn't look to support all of the functionality that I was hoping for, but perhaps those can be feature requests elsewhere. I left a few feature requests last year on the galaxy-issues repo, but have not heard anything there either.

I don't have the time to pick up v2 or clean this up at the moment if it is relevant, so I will close this.

Issue #6327, opened 2014-03-07
docker module ports option error


lenfree:
##### Issue Type:

Bug
##### Ansible version:

1.5
##### Environment:

*BSD
##### Summary:

can't use ports option to bind an interface with multiple ports if they both tcp or udp
e.g. 
ports="10.0.2.15:80:80, 10.0.2.15:443:443" - this doesn't work
but ports="10.0.2.15:80:80, 10.0.2.15:443:443/udp" - this works
##### Steps To Reproduce:

 docker: image="name/nginx-reverse-proxy:1.0" name="nginx"     state=present hostname="nginx" l ports="10.0.2.15:80:80, 10.0.2.15:443:443"
##### Expected Result:

should be able to bind multiple ports on the same interface such as
10.0.2.15:80->80/tcp
10.0.2.15:443->443/tcp
##### Actual result:

Error

joshuaconner:
Hmm, I'm not able to reproduce this issue. (Here's a [gist](https://gist.github.com/joshuaconner/8bbea5b0a1cc02d5fcad) of the playbook, hosts file and Dockerfile I used, as well as the output I got from running `docker ps` on the host after the playbook had run.)

Maybe the problem is the space after the comma in the port declaration?

lenfree:
oh there shouldn't be a space after the comma to prevent error. Thanks! :)

PR #6541, opened 2014-03-18
docker bugfixes surrounding parsing a comma-separated string to a list

joshuaconner:
1) A single number param (3333 being passed to 'ports', for example) will cause the module to error - this PR explicitly casts the value to string before splitting if it's not a list already.
2) In #6327, someone had an issue involving spaces between commas causing issues. This calls `strip()` on each list param to keep that from happening in the future.

### DIFF
--- a/library/cloud/docker
+++ b/library/cloud/docker
@@ -381,8 +381,10 @@ class DockerManager:
         """
         param_list = self.module.params.get(param_name)
         if not isinstance(param_list, list):
-            param_list = param_list.split(delimiter)
-        return param_list
+            # if param_list is a number, like 3333, this will fail, so we coerce to a str first
+            param_list = str(param_list).split(delimiter)
+        # whitespace in between commas will cause problems if we don't strip each param
+        return [param.strip() for param in param_list]
 
 
     def get_exposed_ports(self, expose_list):
--- a/library/cloud/docker
+++ b/library/cloud/docker
@@ -331,7 +331,7 @@ class DockerManager:
         if self.module.params.get('volumes'):
             self.binds = {}
             self.volumes = {}
-            vols = self.parse_list_from_param('volumes')
+            vols = self.module.params.get('volumes')
             for vol in vols:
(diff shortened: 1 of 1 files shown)


mpdehaan:
The original module is missing an example of "param_list" being used, I would hope that type=list (or dict) could perhaps be used and avoid the string cast of a list, which seems potentially quirky. 

Can you share an example?

joshuaconner:
I locally changed the module to use the `list` type rather than casting a string to a list, but it breaks a lot of the existing examples. For instance:

```
- hosts: web
  sudo: yes
  tasks:
  - name: run tomcat server
    docker: image=centos name=tomcat command="service tomcat6 start" ports=8080
```

gives the following error:

```
failed: [IP_ADDRESS] => {"failed": true}
msg: argument ports is of invalid type: <type 'int'>, required: list
```

Options I see are:
- Leave it as-is. It's not super pretty but it does work.
- Use a `list` type for everything BUT `ports`, and manually attempt to cast that argument to a list if it's not already. 
- Ansible starts casting `8080` to `[8080]` when the argument is of `list` type. This is probably a bad idea that will have some unforeseen consequences.
- ??

I'm definitely open to ideas here. What would you like to see happen?

joshuaconner:
Whoops, just kidding! That example works fine! It gets mad when you do it like this:

```
- hosts: web
  sudo: yes
  tasks:
  - name: run hello world server
    docker:
      image: tutum/hello-world
      ports: 80
```

Since that is actually incorrect (`80` should be the first element of the `ports` array) this is a-okay!

joshuaconner:
(This should be good to test whenever. I changed the module to use the `list` type instead of doing the wonky manual cast.)

mpdehaan:
This looks good to me, so merging, we can make the list stuff accept non-string scalar types to get around what you mentioned above.

mpdehaan:
The above is also done, all merged in, thanks!

PR #6544, opened 2014-03-18
A module for notifications through the Slack.com platform

ramondelafuente:
This module can send messages through Slack.com, and supports all options
available in the "incoming webhook integration".
- text
- channel
- username
- icon_url / icon_emoji
- link_names
- parse

### DIFF
--- a/library/notification/slack
+++ b/library/notification/slack
@@ -20,10 +20,9 @@
 
 DOCUMENTATION = """
 module: slack
-short_description: Send notifications to U(Slack.com)
+short_description: Send Slack notifications
 description:
-    - The M(slack) module sends notifications to U(Slack.com) via the incoming WebHook
-      See the Slack documentation U()
+    - The M(slack) module sends notifications to U(http://slack.com) via the Incoming WebHook integration
 version_added: 1.6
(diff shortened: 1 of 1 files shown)


mpdehaan:
Does Slack have a users page somewhere so we could gauge the level of updake on this service?

Thanks!

dirtyharrycallahan:
https://slack.com/love

ramondelafuente:
Honestly, I just signed up two days ago. Loving the **\* out of it. Took 4 hours to get everything integrated for the team (5) - then about twice that to write the module because, well... Ansible all the things.

alirayl:
Here's our latest growth chart, which includes usage numbers: 
image.

(from https://twitter.com/SlackHQ/status/443823101491425280)

mpdehaan:
Thanks, that's more than sufficient, I think.

We want to avoid adding modules for services that might not have enough Ansible users, but when the code is simple, these things are easy to keep in there, and it lets people browse what has integration so they can choose what services to pick.

mpdehaan:
merged, thank you!

alirayl:
Thank you very much! We'll figure out the appropriate way to link this up on our site as well.

PR #6556, opened 2014-03-18
add param to note example for ssh-keyscan

andreaskern:
hostnames in the known hosts file are typically stored as Hashed values, calling 'ssh-keyscan' with '-H' changes to output to the Hashed format so that the known_hosts file looks more consistent


mpdehaan:
It's actually better now with accept_hostkey=yes, I'll make sure that note is in there too.

mpdehaan:
merged, thanks!

PR #6540, opened 2014-03-18
Add cloud modules ec2_scaling_policy and ec2_metricalarm

Zeekin:
Added AWS modules ec2_scaling_policy and ec2_metricalarm for configuring scaling policies for autoscaling groups, and metric alarms.

PR #6419, opened 2014-03-12
Improvements to ec2 autoscaling modules

willthames:
- Added desired_capacity and vpc_zone_identifier to ec2_asg
- Use ec2_argument_spec() method and then remove unnecessary
  declarations from argument_spec
- Remove AWS_REGIONS declaration
- Rename block_device_mappings to volumes to be consistent with ec2
- Remove all pep8 warnings except line length and continuation indent
- Use updated module_utils/ec2.py to add profile and security_token
  support
- Remove mandatory arguments for delete to make launchconfig deletion
  work
- Handle existing launch configurations better
- Improve output information
- Improve documentation

PR #6558, opened 2014-03-18
Added ec2_cloudwatch module to create and destroy CloudWatch alarms

brafales:
As discussed on the mailing list given that none of the current ec2 modules support it

### DIFF
--- a/library/cloud/ec2_cloudwatch
+++ b/library/cloud/ec2_cloudwatch
@@ -1,0 +1,280 @@
+#!/usr/bin/python
+DOCUMENTATION = '''
+---
+module: ec2_cloudwatch
+short_description: create and destroy cloudwatch alarms.
+description:
+    - This module creates and destroys AWS CloudWatch alarms for AWS.
+version_added: 1.4
+options:
+  dimensions:
+    description:
+      - A dictionary of dimension key/values where the key is the dimension name and the value is either a scalar value or an iterator of values to be associated with that dimension.
+    required: false
+  name:
+    description:
+      - Name of alarm.
+    required: true
+  metric:
+    description:
+      - Name of alarm's associated metric.
+    required: true
+  namespace:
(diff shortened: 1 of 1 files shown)


mpdehaan:
@pas256 any chance I can get you to take a look at this real quick?  Thank you!

mpdehaan:
Also note there was an alarm module for AWS that was just merged, so I think this is probalby a more of a question of consolidation with #6540?  Let me know your thoughts @brafales and @willthames and @Zeekin

Thanks!

brafales:
Both modules are very similar. The one that was merged in fact does some more stuff in checking and updating existing alarms and only reporting the status of `changed` when the alarm has actually been changed.
Only thing I'm not sure about is the parameters required attributes. Haven't played much with `boto` but although some of the parameters there are optional, missing them will not let you create the alarm (i.e, `statistic`, `namespace`, `metric` and so on). @Zeekin what are your thoughts about this?
Happy to drop this request in favour of the other merge as it was done before mine and is more complete.

willthames:
@brafales - the reason that the parameters are optional is that the same module copes with creation and deletion - and the parameters aren't required for deletion. 

In #6419 we deal with a similar issue for `ec2_asg` by having an `enforce_required_arguments` method that is only used for creation of auto scaling groups - something similar could be done here if required.

brafales:
I haven't thought about this scenario, it makes sense.

pas256:
Yeah, I think #6540 has it covered - I love a module that can clean up after itself.

Zeekin:
@brafales with regards to the optional parameters, I think @willthames has explained the reasoning - also worth noting if parameters change between creates on an existing alarm, it will apply any 'updates' to the alarm's parameters; not all parameters are required to perform an 'update'.

brafales:
Yes, feel free to close this one :)

PR #6562, opened 2014-03-18
get_distribution_facts called twice

hostmaster:
On Linux `get_distribution_facts` called twice form `get_platform_facts`  and `__init__` then


mpdehaan:
This code has since moved to a common utils class in module_utils and seems to be resolved now.

If you still see this problem in the new shared file (I don't seem to see it) feel free to submit another.

Thanks!

PR #6564, opened 2014-03-18
Update example to include delete on termination

joehack3r:
Benefit of creating EBS volumes on EC2 instance launch is to have the volumes terminate when the instance is terminated. Updated the examples to show this is possible.


mpdehaan:
merged, thanks!  This will appear in the next docs update.

PR #6565, opened 2014-03-18
Fixed bug that occurred when pkgsite not specified

fabianfreyer:
Fixes a bug introduced in #6405 where installing a package would fail if pkgsite was not specified.


mpdehaan:
merged, thanks!

PR #6569, opened 2014-03-19
File fixes relsymlink fix

bcoca:
fixes issue with relative symlinks produced by my last PR.

they are now correctly detected as existing/missing sources when they are relative to path.

```
    - name: make relative link
      file: path=/tmp/www src=../var/www state=link
```


bcoca:
sneak commit, closing

PR #6570, opened 2014-03-19
now correctly checks absolute path for src= existance for links

bcoca:
also updated docs to be a bit clearer on symlink behaviour

```
   - name: make relative link
      file: path=/tmp/www src=../var/www state=link
```


bcoca:
merged into #6595

PR #6572, opened 2014-03-19
unified calls to fs function vs duplicate functions for file and dir

bcoca:
The functions were identical and already merged in previous commit, now all modules that used them call the unified function so there is no need for the placeholders.


jimi-c:
As discussed on IRC, this may break non-core modules so we'll continue to keep these function stubs for backwards compatiblity. Thanks!

PR #6574, opened 2014-03-19
Improve yum module perfomance

MiLk:
The result of the `what_provides` function is filtered to use only the most recent version of each package.
When a package has a huge list of versions available, it will only test if the most recent is installed.

I tested with the following task:

```
- name: "Update yum and plugins"
  yum-devel:
    name={{ item }}
    state=latest
    enablerepo=epel-yum-rawhide
  with_items:
    - yum
    - yum-plugin-fastestmirror
```

Repo file here: http://repos.fedorapeople.org/repos/james/yum-rawhide/epel-yum-rawhide.repo

Before this change without repoquery: 3m57s
After this change without repoquery: 34s

Before this change with repoquery: 4m20s
After this change with repoquery: 20s

### DIFF
--- a/library/packaging/yum
+++ b/library/packaging/yum
@@ -25,8 +25,6 @@
 import traceback
 import os
 import yum
-import re
-from distutils.version import LooseVersion
 
 try:
     from yum.misc import find_unfinished_transactions, find_ts_remaining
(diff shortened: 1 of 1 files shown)


mpdehaan:
I'm a little wary of the regex in the middle of this code making this non-intuitive to read.   Can you add some comments to this?  (And possibly consider using string operations if possible instead?)

MiLk:
I use the regex to extract the name of the package.
The format is `%{name}-%{version}-%{release}.%{arch}`.
I'll try to find a better way to do that.

MiLk:
The new implementation seems to be better.
It's faster too.

It doesn't use anymore `what_provides`.
It queries only the latest available version of the package.

MiLk:
Any news?

mpdehaan:
Hi!

Thank you very much!

PR #6575, opened 2014-03-19
Hipchat callback: Make "notify" flag optional (default off)

mahemoff:
The "notify" flag means everyone in the room sees the title change color and possible visual/audio alert, which I found too much for most situations, especially as it happens on start and end of the playbook run. This makes it optional by extending the environment variable configuration. It also defaults it to off.

### DIFF
--- a/plugins/callbacks/hipchat.py
+++ b/plugins/callbacks/hipchat.py
@@ -54,7 +54,6 @@ class CallbackModule(object):
         self.room = os.getenv('HIPCHAT_ROOM', 'ansible')
         self.from_name = os.getenv('HIPCHAT_FROM', 'ansible')
         self.allow_notify = (os.getenv('HIPCHAT_NOTIFY') != 'false')
-        print(self.allow_notify)
 
         if self.token is None:
             self.disabled = True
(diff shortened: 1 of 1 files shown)


mpdehaan:
This breaks the current behavior of the script @sivel is using by requiring a new flag.  I'd accept a parameter to disallow notification, however.

mahemoff:
I'd rather not use a negative name for a variable (it leads to double-negative logic, which is confusing). I could make the default true, which should serve the same purpose.

mahemoff:
Just to add, it's a great plugin which I've been using all day, but I do think the current notification default will not be what most people want. It means everyone gets a notification at least twice per playbook run. Notification flag should only be used for critical errors, PMs, and so on IMO.

sivel:
Also note that files in the 'plugins' directory are just examples.  However, our requirement was that everyone should be notified that a play is running and when it ends.

I think just changing the default to true instead of false sounds reasonable.

mahemoff:
Makes sense, that's done.

mpdehaan:
merged, thanks!

PR #6580, opened 2014-03-19
Changed call to EDITOR to allow for parameters

ramondelafuente:
The EDITOR environment variable is used to create and edit files in the vault.
But if the EDITOR variable contains parameters, subprocess.call() breaks.

```
export EDITOR="subl -w"
```

This fixes the EDITOR environment variable to be safely split into a list.
It adds a dependency on shlex.

-- please review, I am not a python developer by day. Or by night.

### DIFF
--- a/lib/ansible/utils/vault.py
+++ b/lib/ansible/utils/vault.py
@@ -19,6 +19,7 @@
 # installs ansible and sets it up to run on cron.
 
 import os
+import shlex
 import shutil
 import tempfile
 from io import BytesIO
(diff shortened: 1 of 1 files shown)


mpdehaan:
Thanks for this, it looks pretty reasonable.

per code conventions in CODING_GUIDELINES.md (this is minor),I would definitely rename toCall to "to_call".

It should also be possible to move the needed logic inside of "call" instead so the logic isn't repeated.

ramondelafuente:
rename the var: will do.

As for moving the logic to inside of call.. isn't that a subprocess method? Don't know how to proceed on that one.

ramondelafuente:
I removed the duplication; not sure if this is what you meant.

jctanner:
@ramondelafuente can you squash the two commits? Looks good right now, and passes my basic tests so I'll merge it once squashed.

ramondelafuente:
@jctanner Should be good to go.

jctanner:
@ramondelafuente thanks!

ramondelafuente:
@jctanner No, good sir, thank you for a fine product that makes us very happy users :-)

PR #6584, opened 2014-03-19
Expose ansible_version to playbooks, add version_compare filter. Fixes #6583

### LINKED ISSUE #6583: Expose Ansible version to the playbooks
##### Issue Type:

Feature Idea
##### Ansible Version:

`ansible 1.5.3`
##### Environment:

N/A
##### Summary:

It would be needed to have the Ansible version available from the playbooks in a variable, like `ansible_version` that evaluates to `1.5.3`.

There have been some issues recently that make my roles compatible only with the latest Ansible, and I'd like to enforce this.
##### Steps To Reproduce:

N/A
##### Expected Results:

N/A
##### Actual Results:

N/A

sivel:
A number of people have asked for the ansible version to be exposed to playbooks.

This pull request exposes it as `ansible_version`.

Additionally, on it's own, it is complicated to compare that version to another value.

This pull request additionally adds a `version_compare` filter for easy comparison of versions.

Unit tests have been added for the new filter, and here is an example playbook to see this in action:

```

---
- hosts: all
  tasks:
    - debug:
        var: ansible_version

    - debug:
        msg: "{{ ansible_version|version_compare(1.6) }}"

    - debug:
        msg: "{{ ansible_version|version_compare(1.4, 'gt') }}"

    - debug:
        msg: "{{ ansible_version|version_compare(1.7, 'lt') }}"

    - debug:
        msg: "{{ ansible_version|version_compare(1.6, '>=') }}"

    - debug:
        msg: "{{ ansible_version|version_compare(1.6, '<') }}"
```

The `version_compare` accepts an 'operator' argument, which supports: <, lt, <=, le, >, gt, >=, ge, ==, =, eq, !=, <>, ne

### DIFF
--- a/docsite/rst/playbooks_variables.rst
+++ b/docsite/rst/playbooks_variables.rst
@@ -208,6 +208,28 @@ To get the symmetric difference of 2 lists (items exclusive to each list)::
 
     {{ list1 | symmetric_difference(list2) }}
 
+.. _version_comparison_filters:
+
+Version Comparison Filters
+--------------------------
+
+.. versionadded:: 1.6
+
+To compare a version number, such as checking if the running Ansible version is greater than or equal to
+1.6, you can use the ``version_compare`` filter::
+
+    {{ ansible_version | version_compare(1.6, operator='ge') }}
+
+If ``ansible_version`` is 1.6 or greater, this filter will return True, otherwise it will return False.
+
+The ``version_compare`` filter can also be used to evaluate the ``ansible_distribution_version``::
+
+    {{ ansible_distribution_version | version_compare(12, '>') }}
+
+The ``version_compare`` filter accepts the following operators::
+
+    <, lt, <=, le, >, gt, >=, ge, ==, =, eq, !=, <>, ne
+
 .. _other_useful_filters:
 
 Other Useful Filters
@@ -677,6 +699,8 @@ period, without the rest of the domain.
 Don't worry about any of this unless you think you need it.  You'll know when you do.
 
 Also available, *inventory_dir* is the pathname of the directory holding Ansible's inventory host file, *inventory_file* is the pathname and the filename pointing to the Ansible's inventory host file.
+
+The running version of Ansible is also available via the *ansible_version* variable.
 
 .. _variable_file_seperation_details:
 
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -34,6 +34,7 @@ import subprocess
 
 import ansible.constants as C
 import ansible.inventory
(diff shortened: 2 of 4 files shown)


mpdehaan:
This looks fine to me, though when adding a new filter plugin we also need to add it to the docs section about filters.

sivel:
I'll get the docs added shortly.

sivel:
I've added docs about `version_compare` and `ansible_version` and squashed the commits.

srgvg:
Appears I oversaw this PR when making my own :) See #6619 which however only introduces the ansible_version variable, and no Jinja filter, albeit:
- in runner, not in playbook, so it's also available outside of playbooks
- as a dict, where I split out the different components of the version string
- I used the code that generates the --version string used by the cli tools, which for a git checkout also includes extra information.

Maybe we could combine features?

sivel:
@sergevanginderachter  I'm not too convinced on the implementation in the other pull.  Also can you provide a use case for needing ansible_version outside of a playbook?

Also, I think that with a useful version_compare filter, a dictionary/hash value for ansible_version is potentially overcomplicated and unneeded.  If we think it is ultimately useful, a more simplistic approach would be something like:

```
version_string = utils.version('')
version = version_string.split()[0]
version_parts = map(int, ('%s.0' % version).split('.'))
inject['ansible_version'] = dict(release=version,
                                 major=version_parts[0],
                                 minor=version_parts[1],
                                 patch=version_parts[2],
                                 string=version_string)
```

@mpdehaan do you have any feedback?

srgvg:
@sivel Disregard my remark about availability outside playbook, I misread yours, you actually implemented at the same place as I did.

Your handing of integer conversion is obviously cleaner, but what if a version has a non-integer component? I was just making sure nothing would break for people that maintain a private branch, and might do versions like "2.3.0b1".

E.g. at one customer, I myself have this:

```
~# ansible --version
ansible 1.6.0-acd-20140312-0-custombugfixes
```

If your compare filter handles that too, fine by me.

mpdehaan:
I like the idea of using StrictVersion, while I don't agree it is quite necessary to have it available in Runner, that's consistent for most of Ansible provided it does not do calculuation each time.  Which I've commented in the other PR about.  So going to work on that one.

Sorry for duplicate efforts!

sivel:
Do you want a version_compare filter?  And are you saying that it should use StrictVersion instead?

I chose LooseVersion as that is what ansible-galaxy uses, and it is more flexible, but I can make that change.  I am more interested in the version_compare filter over exposing ansible_version anyway.

If you aren't interested, I won't waste time submitting another pull request.

PR #6586, opened 2014-03-19
Fix logic in process_common_errors for unbalanced quotes

sivel:
This pull addresses a logic issue in `process_common_errors` where we were always only looking for unbalanced single quotes instead of unbalanced single and double quotes.

The matching unit test has been uncommented now as well.


mpdehaan:
Merged, thank you!

PR #6593, opened 2014-03-20
Atomic move restore owner group rebase

tyll:
If src and dest in atomic_move() are not on the same file system, the owner/group of dest are not restored. This can be tested with the lineinffile module:

```
touch /root/lineinfile-test.txt
chown nobody:  /root/lineinfile-test.txt
ansible --sudo host -m lineinfile -a "dest=/root/lineinfile-test.txt regexp=^test line=test state=present"
```

Afterwards the /root/lineinfile-test.txt does not belong to nobody anymore. For this happen /tmp needs to be on a separate filesystem.

Also note that the tempfile in atomic_move() is created insecurely, this is fixed by using the tempfile module.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -55,6 +55,7 @@ import types
 import time
 import shutil
 import stat
+import tempfile
 import traceback
 import grp
 import pwd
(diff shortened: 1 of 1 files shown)


mpdehaan:
cc @bcoca -- thoughts as this was your baileywick?

bcoca:
update looks good to me, I was actually thinking of making similar patch
;-)

bcoca:
just one comment, tempfile module relies on randomness to provide uniqueness, this is not good on systems that have high concurrency and randomness use, adding os.getpid(),time.time() gives you a very high likelyhood of uniqueness, which is what is really needed.

I was also thinking of adding a acl/xattr detection/preservation as it is becoming an issue for me when replacing files, but that gets lots more complicated. ;-)

tyll:
On Thu, Mar 20, 2014 at 06:17:55AM -0700, Brian Coca wrote:

> just one comment, tempfile module relies on randomness to provide
> uniqueness, this is not good on systems that have high concurrency and
> randomness use, adding os.getpid(),time.time() gives you a very high
> likelyhood of uniqueness, which is what is really needed.

Actually tmpfile relies on  os.O_EXCL to be properly implemented to
ensure uniqueness. It if it works it will never create a file that is
already there. Also it uses /dev/urandom by default, which falls to to a
CSPRNG if there is not enough entropy in the entropy pool. I strongly
doubt that it will be easier to guess the next 16 bytes from
/dev/urandom. However, if you insist to add os.getpid()/time.time() to
the file name I will not fight it, as long as the tempfile module is
properly used.

> I was also thinking of adding a acl/xattr detection/preservation as it
> is becoming an issue for me when replacing files, but that gets lots
> more complicated. ;-)

It sound like a good idea to handle acl/xattr eventually.

tyll:
Btw. ansible uses the tempfile module several times already, why should it be only a problem here?

bcoca:
@tyll, it really is not, I doubt ansible will ever run with the high concurrency update issues I describe above, I might just bee a bit to paranoid since this has bitten me in the past.

bcoca:
@tyll, it really is not, I doubt ansible will ever run with the high concurrency update issues I describe above, I might just be a bit to paranoid since this has bitten me in the past.

jimi-c:
In queue for testing.

jimi-c:
Merged, thanks!

tyll:
The insecure tempfile usage is now tracked as [CVE-2014-2912](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2014-2912) (it might take some time till the URL is available).

PR #6597, opened 2014-03-20
Handle missing symlink tempfile 

tarka:
Only unlink the symlink tempfile on error if it was created in the previous operation.

### DIFF
--- a/library/files/file
+++ b/library/files/file
@@ -305,7 +305,8 @@ def main():
                         os.symlink(src, tmppath)
                     os.rename(tmppath, path)
                 except OSError, e:
-                    os.unlink(tmppath)
+                    if os.path.exists(tmppath):
+                        os.unlink(tmppath)
                     module.fail_json(path=path, msg='Error while replacing: %s' % str(e))
             try:
                 if state == 'hard':


mpdehaan:
Can you share the steps that lead you to experience this error? (for test purposes)

Thanks!

bcoca:
should be fixed in #6595 (which includes ##6578)

tarka:
I've got a weird error where the symlink is not created (I'm still working on why that is, but it may be related to path handling @bcoca fixed above). In that case the exception handler will throw an error on the unlink as the tmpfile doesn't exist.

tarka:
@bcoca I believe your PR fixes the underlying issue (bad paths for symlink tempfile) but I think we should still check if the tmp link was created correctly before attempting to remove it.

bcoca:
this patch might still be useful for corner cases in which module does not
check permissions and tmpfile creation fails, but I was thinking of adding
another try/except to also capture the even more unlikely external
permission change between tmpfile creation and deletion on rename fail.

mpdehaan:
merged for now, I'm ok with upgrading this to a try/except later as well.

PR #6599, opened 2014-03-20
use region_name in nova_compute

dmage:
Currently region_name is not used. This commit fixes it.

### DIFF
--- a/library/cloud/nova_compute
+++ b/library/cloud/nova_compute
@@ -249,6 +249,7 @@ def main():
                               module.params['login_password'],
                               module.params['login_tenant_name'],
                               module.params['auth_url'],
+                              region_name=module.params['region_name'],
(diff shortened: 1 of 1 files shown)


kvdveer:
+1
Looks good to me

tryggth:
+1 lgtm.
What happens when the region is left out? For Essex I imagine it drops the parameter. And I guess it just fails on Havana.

dmage:
Eventually it will be here as filter_value: https://github.com/openstack/python-novaclient/blob/master/novaclient/service_catalog.py#L34
So it shouldn't depend on OpenStack release.
If you have only one region, then you may omit the region (filter_value is None). Otherwise you have to specify region or you'll get AmbiguousEndpoints error.

jimi-c:
Merged, thanks!

PR #6600, opened 2014-03-20
assign_public_ip was actually added in 1.5

jjshoe:
Checked 1.4.5 and this option doesn't exist. Will check/update/pull docs next.


mpdehaan:
Thanks!

PR #6598, opened 2014-03-20
library/cloud/ec2_group: Enable egress rules

randomvariable:
This pull request enables egress rules in similar fashion to https://github.com/ansible/ansible/pull/6416 . However, this includes a new parameter for egress rules, 'rules_egress' rather than having to declare the type.
In addition, default and non-declared rules will be removed as is currently the case for ingress rules.

PR #6602, opened 2014-03-20
Add support for ec2 security group egress rules

mmoya:
This patchset add support for egress rules in ec2 security groups. Patchset is split for easy reviewing. There are some minor differences with respect to #6416 and #6598.

Feedback welcome.


mpdehaan:
So we have three of these now, #6602 #6416 and #6598 

Obviously there's a lot of duplicate effort here, but can you all please get together and sort out which one of these you prefer and define any changes that need to be made?

PR #6603, opened 2014-03-20
rax.py inventory script improvements

sivel:
This pull adds a number of requested features:
- Support other OpenStack clouds via `~/.pyrax.cfg` and the `RAX_ENV` environment variable
- Add flavor based groups (instance size) (eg flavor-4 or flavor-performance2-15)
- Add image based groups (eg image-2ab974de-9fe5-4f5b-9d58-766a59f3de61 or image-debian-7-wheezy)
- Add instance UUID based groups (eg instance-eba77620-be18-4abd-b19d-f67c054a04f0)
- Add meta key_value groups (eg meta_foo_bar) (this is configurable to "match" ec2.py by setting `RAX_META_PREFIX=tag`)
- Support for secondary groups in addition to a primary group.  When creating a server, use a comma separated list of groups like the following when creating a server using the rax module:

```
    meta:
        groups: foo,bar,baz
```
- When 'region' is set in an environment in `~/.pyrax.cfg`, only that region is inspected
- Support for keyring storage of password/api_key is achievable via ~/.pyrax.cfg by utilizing the `keyring_username` configuration

No existing functionality was changed.  This maintains backwards compatibility, just adds the above new functionality.

### DIFF
--- a/plugins/inventory/rax.py
+++ b/plugins/inventory/rax.py
@@ -22,9 +22,11 @@ DOCUMENTATION = '''
 inventory: rax
 short_description: Rackspace Public Cloud external inventory script
 description:
-  - Generates inventory that Ansible can understand by making API request to Rackspace Public Cloud API
+  - Generates inventory that Ansible can understand by making API request to
+    Rackspace Public Cloud API
   - |
-    When run against a specific host, this script returns the following variables:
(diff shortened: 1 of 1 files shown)


mpdehaan:
Merged, thank you!

PR #6607, opened 2014-03-20
Added all_databases option to mysql_db. This works with import and dump

Jmainguy:
This works like --all-databases when using mysqldump. It works like not naming a db at all when doing import, so it will overwrite all db's.


mpdehaan:
version_added should be an attribute on the the task

choices should just be yes or no in the documentation list

I'm not sure how I feel about name= being required for when the option "all databases is used".   Should perhaps maybe "ALL_DATABASES" be a special value of name?   Not sure.

Jmainguy:
I will focus on coding it so name is not required when using all_databases. Not sure how to word the documentation when this is done since name= is required for all other uses when all_databases is not specified.

Jmainguy:
Should be good to go now.

Jmainguy:
Well, I bunched this one, closing it out and will link new pull request to it.

PR #6609, opened 2014-03-20
etcd lookup: use $ANSIBLE_ETCD_URL if it exists in the env

joshuaconner:
Pretty much does what it says on the tin! LMK if you have questions.


mpdehaan:
Merged, thanks!

PR #6612, opened 2014-03-21
Allow npm to install to a directory that doesn't yet exist

willthames:
If path is specified but does not exist, create it.
Fail if path is specified but is not a directory


mpdehaan:
merged, thanks!

PR #6613, opened 2014-03-21
Allow modules to be installed as independent modules

willthames:
The toplevel parameter allows a module to be installed independent
of other modules

This matters when a module should have the binary exist in
$NODE_PATH/node_modules/.bin, even though it might already be
installed as a dependency of another module

### DIFF
--- a/library/packaging/npm
+++ b/library/packaging/npm
@@ -67,6 +67,13 @@ options:
     required: false
     default: present
     choices: [ "present", "absent", "latest" ]
+  toplevel:
+    description:
+      - if toplevel is true, the package will be installed even if already present as a dependency of another node
+    required: false
+    default: no
+    choices: [ "yes", "no" ]
+    version_added: "1.6"
 '''
 
 EXAMPLES = '''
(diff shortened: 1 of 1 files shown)


mpdehaan:
This new logic does not appear to pass any new flags and may appear to always return changed=True, can you explain a bit more about the code changes?

willthames:
@mpdehaan A full demonstration gist is here https://gist.github.com/willthames/9699529

It doesn't need to pass more flags - it just runs install more often.

Previously, if hubot was installed, and coffee was a dependency of hubot, npm.list() on coffee would return ['hubot'] for installed and [] for missing. If toplevel is true and the module to be installed is coffee, then the install method with coffee will be called because 'coffee' is not in installed. 

Running the test playbook twice shows that the expected changed/ok behaviour occurs.

However, I realised while testing this that my test playbook does depend on my other npm change (#6612) so I'll rebase to incorporate that.

jimi-c:
Hi!

Thanks very much for your submission to Ansible. It sincerely means a lot to us that you've taken time to contribute. Unfortunately, we've decided to merge in the alternate commit above.

Since this is a really busy project, we don't always see comments on closed tickets, but want to encourage open dialog. You can stop by the development list, and we'd be glad to talk about it - and we might even be persuaded otherwise!

https://groups.google.com/forum/#!forum/ansible-devel - for user questions, tips, and tricks

PR #6614, opened 2014-03-21
fix delegate_to user

jimhe:
this fixes delegate_to if you have defined ansible_ssh_user to be different than your remote_user.

### DIFF
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -357,27 +357,18 @@ class Runner(object):
 
         # inventory > playbook > original_host
 
-        actual_user = inject.get('ansible_ssh_user', self.remote_user)
-        thisuser = None
+        return_user = None
 
         if host in inject['hostvars']:
             if inject['hostvars'][host].get('ansible_ssh_user'):
                 # user for delegate host in inventory
-                thisuser = inject['hostvars'][host].get('ansible_ssh_user')
+                return_user = inject['hostvars'][host].get('ansible_ssh_user')
 
-        if thisuser is None and self.remote_user:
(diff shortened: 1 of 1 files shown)


mpdehaan:
Can you please paste some example that illustrates the problem this fixes?

The removal of the above code seems to imply it served no useful function, which I'd have a hard time believing it did not :)

jimhe:
Actually, I did not realize that `inject['hostvars'][host].get('ansible_ssh_user')` already encompasses all the other ansible variables to give the correct `ansible_ssh_user` for the delegate host. What I am having trouble with is the key checks in:

``` python
if host in inject['hostvars']:
    if inject['hostvars'][host].get('ansible_ssh_user'):
        # user for delegate host in inventory
        thisuser = inject['hostvars'][host].get('ansible_ssh_user')
```

Even though the value exists, it's not getting through. I'll open another PR when I figure this out.

jimhe:
Ok, I've narrowed the problem down. What I am doing is issuing a `delegate_to` to a host that is in a different group than the current host's group. My use case is that I need to perform some actions on a `db_server` which are dependent on looping through some elements defined in an `app_server`. So the playbook executes for all hosts in `app_server`, and delegates some actions to a host in `db_server`. What ends up happening is that the `HostVars` dict wrapper hides all hosts that are not app_servers, so 

``` python
if host in inject['hostvars']:
```

fails. 

I am not sure if my use case is very strange, but ideally we should be allowing for global use of `hostvars` for delegate_to hosts?

hkariti:
I was bitten by this as well, when using --limit. This `if` statement is the devil :P

PR #6616, opened 2014-03-21
Allow state=absent for ec2_vol

willthames:
Add methods to the module to make it more readable and allow reuse

Tested as best I could (create volume, delete volume, create volume and attach it to instance). Had a look at the ec2_vol integration tests but they aren't ready yet. So if @jctanner has more tests, please let me know.


mpdehaan:
merged, thank you!

Issue #4357, opened 2013-10-03
Allow playbooks to require a specific version of Ansible


magnus919:
I would like to be able to require a specific release or do a comparison test for minimum/maximum version of Ansible to make sure plays will run as expected. Ansible versions outside of that specified in the playbook should ideally fail to run with a useful human readable error.

mpdehaan:
Thanks for the feature request.

This is easy to do and I'll probably look at doing this soon as it's been a frequent request.

It probably looks like this

```
- hosts: blarg
  ansible_required: 1.4
```

We should also do the same for roles, where this could be stored in the "meta" file already being used to control dependencies for roles.

candlerb:
A more flexible way would be to expose it as a variable, then you could do things like:

```
- local_action: apt pkg=ansible state=latest
  when: ansible_version < "01.04.00"
```

or

```
- action: some_future_module foo=bar
  when: ansible_version >= "01.05.00"
```

(suggest "xx.yy.zz" to avoid problems with "1.4.10" < "1.4.9")

gregbuehler:
@candlerb +1

you can just start roles/plays with something like

```
 - fail: msg="This playbook requires Ansible 1.0.5.00 or greater."
   when: {{ ansible_version }} <= "01.05.00"
```

mpdehaan:
We've keeping track of this list but haven't seen any traction here, so are closing this ticket out.  If someone would like to submit a pull request, that would be welcome, and we're still open to it.

alimony:
For anyone finding their way here, I solved this by adding this to the top of my playbook:

``` ruby
- hosts: all
  tasks:
    - name: Get current Ansible version on local host
      local_action: command ansible --version
      register: local_ansible_version
    - assert:
        that:
          - "'1.5.4' in local_ansible_version.stdout"
```

srgvg:
FYI #6619 is in queue to have an version var available, and there is a filter in 1.6 that allows to compare "versions".

axelabs:
This doesn't work when comparing to version 2+ since ansible_version is a dictionary but this seems to work:

```
  tasks:
  - debug: msg="You'll need ansible > 1.9.3 to update all packages (https://goo.gl/4J8NXy) {{ ansible_version }}"
    when: update_all_packages == True and ansible_version.major < "2"
```

srvg points out [Jinja2's version_compare filter](http://docs.ansible.com/ansible/playbooks_filters.html#version-comparison-filters), which I'll try and update this post with results...

[...]

PR #6619, opened 2014-03-21
Introduce an ansible_version dict as runner variable

srgvg:
Given the version:
    "1.6 (ansible_version_var 14499e8bf3) last updated 2014/03/21 17:07:50 (GMT +200)"

  We get the special variable:
    "ansible_version": {
        "full": "1.6",
        "major": 1,
        "minor": 6,
        "revision": 0,
        "string": "1.6 (ansible_version_var 14499e8bf3) last updated 2014/03/21 17:07:50 (GMT +200)"
       }

  modified:   lib/ansible/runner/**init**.py

also in plain bin/ansible:
$ ansible -m debug -a var=ansible_version all
localhost | success >> {
    "ansible_version": {
        "full": "1.6", 
        "major": 1, 
        "minor": 6, 
        "revision": 0, 
        "string": "1.6 (ansible_version_var 69ee37f7a0) last updated 2014/03/21 18:12:46 (GMT +200)"
    }
}

### DIFF
--- a/lib/ansible/playbook/__init__.py
+++ b/lib/ansible/playbook/__init__.py
@@ -177,6 +177,8 @@ class PlayBook(object):
         ansible.callbacks.load_callback_plugins()
         ansible.callbacks.set_playbook(self.callbacks, self)
 
+        self._ansible_version = utils.version_info(gitinfo=True)
+
     # *****************************************************
 
     def _load_playbook_from_file(self, path, vars={}):
@@ -371,6 +373,7 @@ class PlayBook(object):
         )
 
         runner.module_vars.update({'play_hosts': hosts})
+        runner.module_vars.update({'ansible_version': self._ansible_version})
 
         if task.async_seconds == 0:
             results = runner.run()
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -578,6 +578,9 @@ class Runner(object):
         if self.inventory.src() is not None:
             inject['inventory_file'] = self.inventory.src()
 
+        # could be already set by playbook code
+        inject.setdefault('ansible_version', utils.version_info(gitinfo=False))
+
         # allow with_foo to work in playbooks...
         items = None
         items_plugin = self.module_vars.get('items_lookup_plugin', None)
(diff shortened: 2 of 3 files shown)


mpdehaan:
So there are some issues here in that the way the full version including the git sha is calculated would be calculated once per fork, which would be abusive towards running git on the remote system.  I'd accept this just using the **VERSION** variable in lib/ansible/**init**.py though I would like to see a function in utils to return the dictionary rather than this going straight-inline in runner.

A simple dict.update(utils.get_version_info()) would be ok.

Thanks!

srgvg:
Will update, just gimme some time.

srgvg:
Updated per your request, dict is set in runner without calling the git version retrieval code.
If playbook code is involved, then the git info version code is retrieved, but this code is run only once, and cached.

briceburg:
+1 
would be nice to use ansible_version in `when` conditionals; as sometimes there's an improved way to accomplish a task that you'd like to keep in the playbook as reference.

moeffju:
:+1: on this and #6584 (and by extension, #4357). I’m even wondering if it might not make sense to allow pinning playbooks to ansible versions, similar to what Gemfiles and Vagrantfiles allow. I was bitten twice by Ansible modules subtly changing their behavior or their output, and it took a while to figure out that some devs in the team were running different ansible versions. If I could have an attribute like:
- ansible_version: 1.6

optionally with version matchers like bundler/pip (`~> 1.6`, `>=1.6.7`, …). I could have pinned my playbooks and Ansible could’ve told people using other versions that the pin doesn’t match, making the problem obvious.

srgvg:
Or you could

```
- assert:
    that:
    - ansible_version.full = "1.6"
```

moeffju:
@sergevanginderachter Yes, I could, but I think it might be more obvious, more explicit, and allow for more sensible error messages if version pinning was a first-class citizen.

mpdehaan:
Merged, thanks!

drzraf:
Why isn't there an ANSIBLE_VERSION_ID (similar to PHP_VERSION_ID) which just increases as newer releases are published.
Then testing that version is > 1.9.4 is just as simple as doing:
`when: ANSIBLE_VERSION_ID > 109040`

srgvg:
@drzraf there is now, I typically do something like

```
  - name:       Validate we run with the correct Ansible version                
    assert:                                                                     
      that:                                                                     
                - ansible_version is defined                                    
                - ansible_version.full|version_compare('2.0.1.0', '=')          
    run_once:   True                                                            
```

PR #6620, opened 2014-03-21
Clarify documentation for apt pkg, state, install_recommends

cmattson:
Minor changes to the apt packaging documentation to be clearer and better match other packaging modules.


mpdehaan:
Merged, thank you!  Good updates.

These will appear the next time we push the docs build.

PR #6624, opened 2014-03-22
rax: add security_groups param (for OpenStack private cloud support)

joshuaconner:
This makes it so that you can add security groups to your instances if you're using the `rax` module to deploy servers to your own private OpenStack-backed cloud.

(cc @sivel)

### DIFF
--- a/library/cloud/rax
+++ b/library/cloud/rax
@@ -154,6 +154,11 @@ options:
     description:
       - how long before wait gives up, in seconds
     default: 300
+  security_groups:
+    description:
+      - For OpenStack private clouds, the security groups to add your instance to.
+    default: []
+    version_added: 1.6
 requirements: [ "pyrax" ]
 author: Jesse Keating, Matt Martz
 notes:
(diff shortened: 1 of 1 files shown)


sivel:
Thanks for this.  I'll look it over in more depth soon.  I have a branch where I am working on something that may support this in a way that may not expose this option this way, but that still may suit your needs.  So far my approach allows defining just about any option pyrax accepts that this module doesn't explicitly expose.

I'll run these options by a few folks to see what we feel is the right path forward.

joshuaconner:
Cool, definitely no pressure. I just got it working for us and figured I should send in a PR since the work is already done.

sivel:
At this time the Rackspace Public Cloud does not support security_groups.  I've discussed this with a few people, and we want to keep the attributes available to the `rax` module, in alignment with public cloud support at this time.

We have however, just gotten functionality merged to pass arbitrary arguments to the create request, which would allow for security_groups to be configured.

See #6707 

security_group configuration should look like:

```
    - name: Server create request
      local_action:
        module: rax
        credentials: ~/.raxpub
        name: test
        flavor: 2
        image: b8db83d3-59d3-4d02-809f-6be2d41521c2
        disk_config: manual
        networks:
          - public
          - private
        extra_create_args:
          security_groups:
            - name: default
            - name: other
        region: RegionOne
        state: present
        wait: yes
      register: rax
```

Let me know if I can help further.

joshuaconner:
That looks perfect. Thanks!

PR #6626, opened 2014-03-22
Allow custom hipchat urls

brentley:
This allows custom hipchat urls, mainly for use with self-hosted hipchat servers.

I've tested this with my self-hosted hipchat service, and against the normal hipchat.com.  This should be a no-op change for current users of the existing module.

### DIFF
--- a/library/notification/hipchat
+++ b/library/notification/hipchat
@@ -54,6 +54,13 @@ options:
     default: 'yes'
     choices: ['yes', 'no']
     version_added: 1.5.1
+  api:
+    description:
+      - API url if using a self-hosted hipchat server
+    required: false
+    default: 'https://api.hipchat.com/v1/rooms/message'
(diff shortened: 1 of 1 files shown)


mpdehaan:
This looks pretty good.

Looks like the default on line 79 is not used and should probably be set as a constant at the top of the file anyway, like DEFAULT_API_SERVER if it is to be repeated in two places in the code.
Can you update?  Thanks!

brentley:
Okay, cool. It was a constant prior to me messing with it, so this commit https://github.com/brentley/ansible/commit/ca14df47908bde30bd9d173a15e9e00395761238 just goes back to that style.  

I _did_ move the trailing "?" in the original url to the beginning of the auth-token, where I think it makes slightly more sense, once custom urls are used.

mpdehaan:
Merged, thank you!

PR #6629, opened 2014-03-22
work around environment problems in sudo mode

iambocai:
1. work around [environment problems in sudo mode](https://groups.google.com/forum/#!topic/ansible-project/rKHzovQYSLY), by add a sudo_preexec hook and run source command in it by default, this hook can do any other things.
2. host-patterns like ~cache0[1-9], should regard as a full regex-string, not range(~cache0[1:9])


mpdehaan:
Can you please provide an example that shows how to reproduce the problems you encountered?

Also, can you please update your history?   I'm sorry I can't understand the comments in those other commits.

The host pattern commits should also be submitted seperately as they are unrelated to any sudo changes above.

Thanks!

mpdehaan:
I'm also not sure we want to be executing the user's environment.   This is something as a centrally managed system we're explicitly trying to _NOT_ rely on.

iambocai:
@mpdehaan  sorry for my non English comment, I'll pull another request insead :)

The problem can reproduce like this：
on some of my hosts(take 'fe08' as an example here), I have some tools that _only_ can be used by user 'work'(for some security reason), so as a result, I register it in work's .bashrc:

> [work@lg-com-xcache09 ~]$ which ipaddr
> <strong>~/bin/ipaddr</strong>
> [work@lg-com-xcache09 ~]$ cat .bashrc
> 
> ``` bash
> # .bashrc
> # ------ skip unrelated -----
> # User specific environment and startup programs
> export PATH=$HOME/bin:$PATH
> ```

But ansible seems can't find them at all:

> [bob@cs00 ~]$ ansible 'fe08' -m shell -a 'ipaddr' -s -U work
>    fe08 | FAILED | rc=127 >>
>    /bin/sh: ipaddr: command not found

Later I have found that ansibles sudo mode is in fact a non-login terminal to run the command：

>    [bob@cs00 ~]$ ansible 'fe08' -m shell -a 'shopt login_shell' -s -U work  
>    fe08 | FAILED | rc=1 >>
>    login_shell     off

But according to the [norms](http://linux.die.net/man/1/bash), even non-login shell also need to source ~ /.bashrc, that's which we did not do recently:

>    [bob@cs00 ~]$ ansible 'fe08' -m shell -a 'echo $PATH' -s -U work  
>    fe08 | success | rc=0 >>
>    /usr/lib64/qt-3.3/bin:/usr/local/bin:/bin:/usr/bin

In fact, we run a lot of programs in production environments which is in need for environment variables like PYTHONPATH JAVA_HOME and so on.

there is a '-i' option in sudo can do these things, but when I add it to sudo_flags(in ansible.cfg), ansible crashed:

>    [bob@cs00 ~]$ ansible 'fe08' -m shell -a 'whoami' -s -U work  
>    fe08 | FAILED >> {
>        "failed": true, 
>        "msg": "\r\n", 
>        "parsed": false
>    }

To work around this problem, I [add a shell hook](https://github.com/iambocai/ansible/commit/e4a03e4ad82f9ba440c7d2cba5dcd49fdce2fe35), and [run souce commands in it](https://github.com/iambocai/ansible/commit/32163da7f00b11e808a74925373f6f4b3f0db663),that's all

thank you

iambocai:
new pull(#6636) is avaliable, @mpdehaan

PR #6630, opened 2014-03-22
Modified library/system/user so that password change date is set on Sola...

bigl0af:
Modified library/system/user so that password change date is set on Solaris when updating password. Variable fields[2] now contains days since epoch before the join() and ultimate writing out of /etc/shadow on SunOS machines.


mpdehaan:
merged, thank you!

PR #6631, opened 2014-03-22
Add tesing deps installation into hacking/README

kavu:
Micro Pull Request. 

During the hacking on Ansible I tried to run `make tests` and test didn't execute properly until I installed `nose`, `passlib` and `pycrypto`. So, I think it's reasonable to add those deps into [hacking/README](/ansible/ansible/blob/devel/hacking/README.md).


mpdehaan:
merged, thanks!

PR #6632, opened 2014-03-22
funny mistype

peroksid:
lies => lays


mpdehaan:
Pretty sure this is correct as written.

Source: I looked it up on Yahoo Answers

PR #6634, opened 2014-03-23
Remove print statements

pilou-:
"print item" statement raises an exception when type of item is jinja2.runtime.StrictUndefined.


mpdehaan:
merged, thanks!

PR #6635, opened 2014-03-23
Fix ValueError when selection contains a colon

pilou-:
For example ValueError exception occurs with `postfix/mynetworks` parameter because value contains colons:

```
$ debconf-show postfix |grep networks
  postfix/mynetworks: 127.0.0.0/8 [::ffff:127.0.0.0]/104 [::1]/128
```


mpdehaan:
merged, thanks!

Issue #6466, opened 2014-03-13
Add 'update' option to ansible-galaxy


geerlingguy:
##### Issue Type:

Feature Idea
##### Ansible Version:

ansible 1.5
##### Environment:

N/A
##### Summary:

As [mentioned](https://groups.google.com/forum/#!topic/ansible-devel/o_-WIDQWKqo) on the mailing list, the `ansible-galaxy` command has `install` and `remove` options, but no `update` option.

The solution is currently to use `ansible-galaxy install [role],[version] --force`, but that feels a little strange, and isn't necessarily obvious.

Could we add an update command that works like `ansible-galaxy update [role],[version]` (where `,[version]` is optional, and specifies a particular version to download), and simply updates to the latest version of the module available—or, if it's already the latest, no update is performed?
##### Steps To Reproduce:

N/A
##### Expected Results:

N/A
##### Actual Results:

N/A

geerlingguy:
Thank you, ansibot :)

ramondelafuente:
:+1: 

Also, I think it would be nice if ansbile-galaxy install didn't download the zipfile before finding out that the folder already exists.

If you add a role to your dependency list you can update -i (ignore errors), but all roles are downloaded.

nickjj:
Sounds good but I think update would also need an optional version. I have some tooling that generates a rails application along with a play book/inventory to go with it.

I would like to eventually have it update the roles automatically but I wouldn't want to always goto the latest version. If I goto the latest version then I'm forced to update my tool whenever I release a new role or someone releases a breaking change in a role they release.

steenzout:
@nickjj why not use a requirements.yml and specify your dependencies with version there?
you'll need to tag your role's repo.

```
- src: git+ssh://git@host:port/repo/myrole.git
  version: v1.0.0-1  # git tag that you want to download
  name: myrole
  scm: git
```

```
$ ansible-galaxy install -r requirements.yml
```

to update you may need to create a script to remove the roles since I didn't manage to do a force install using the requirements file.

```
#!/bin/bash
# update-roles.sh
grep 'name:' requirements.yml | awk -F: '{print $2}' | xargs -I {} ansible-galaxy remove {}
ansible-galaxy install -r requirements.yml
```

damko:
I totally support this feature request, even without versioning at first.

mchelen:
Update would be very helpful. When using a role with many dependencies they all have to be `install --force` manually.

jloh:
+1

[...]

PR #6637, opened 2014-03-23
Implement basic `galaxy info` command

kavu:
As a part of my investigation on #6466 I found out that `galaxy info` function consists just of one `pass`, so I decided to implement basic `info` command. 

I can call it Proof on Concept, so if you'll want me to fix something or to expand output of the command, I'll be glad to help.

### DIFF
--- a/bin/ansible-galaxy
+++ b/bin/ansible-galaxy
@@ -251,7 +251,7 @@ def api_get_config(api_server):
     except:
         return None
 
-def api_lookup_role_by_name(api_server, role_name):
+def api_lookup_role_by_name(api_server, role_name, quiet=False):
     """
     Uses the Galaxy API to do a lookup on the role owner/name.
     """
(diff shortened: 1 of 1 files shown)


bin/ansible-galaxy
```diff
@@ -262,9 +262,9 @@ def api_lookup_role_by_name(api_server, role_name):
         parts = role_name.split(".")
         user_name = ".".join(parts[0:-1])
         role_name = parts[-1]
-        print " downloading role '%s', owned by %s" % (role_name, user_name)
+        if not quiet:
```
kavu, line 265: To avoid copy-pasting I am using `api_lookup_role_by_name` with additional parameter, quieting output during API call.

bin/ansible-galaxy
```diff
@@ -262,9 +262,9 @@ def api_lookup_role_by_name(api_server, role_name):
...
+            print " downloading role '%s', owned by %s" % (role_name, user_name)
     except:
-        parser.print_help()
```
kavu, line 267: `parser` is undefined in this function, I think we can delete this statement safely.

bin/ansible-galaxy
```diff
@@ -604,7 +604,33 @@ def execute_info(args, options, parser):
...
-    pass
+    api_server = get_opt(options, 'api_server', 'galaxy.ansible.com')
+
+    for i, role_name in enumerate(args):
+        if get_role_metadata(role_name, options):
+            info       = get_galaxy_install_info(role_name, options)
+            role_data  = api_lookup_role_by_name(api_server, role_name, quiet=True)
+
+            if info and role_data:
+                print 'Role: %s' % role_name
```
kavu, line 615: I am not sure if it's ok to use `print print print`. Should I define a multiline-string template and interpolate values with it?

kavu:
Closing in favour of existing `galaxy info`.

PR #6638, opened 2014-03-23
Fix relative symlink support

ef4:
A recent change (I think 2d25577e1104101c67bb5ac41790693780a6d51b) broke support for creating relative symlinks. This should fix it.

### DIFF
--- a/library/files/file
+++ b/library/files/file
@@ -266,8 +266,13 @@ def main():
 
     elif state in ['link','hard']:
 
-        if not os.path.exists(src) and not force:
-            module.fail_json(path=path, src=src, msg='src file does not exist')
+        if not force:
+            if os.path.isabs(src):
+                abs_src = src
+            else:
+                abs_src = os.path.normpath(os.path.join(os.path.dirname(path), src))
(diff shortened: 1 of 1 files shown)


mpdehaan:
Can you please illustrate how you reproduced the breakage?

ef4:
It's very easy to reproduce. This:

```
---
- hosts: bug_test
  tasks:
  - file: state=directory dest=/tmp/example
  - file: state=link src=example dest=/tmp/link

```

fails with "src file does not exist". Whereas it used to work, and with this patch it works again.

muffl0n:
Confirmed, creating symlinks doesn't seem to work at all for some days. See #6506

bcoca:
#6595 should already take care of this

mpdehaan:
Yes, we'll have to compare the two.

Thanks!

jimi-c:
Confirmed, #6595 fixes this issue, and they appear to be pretty functionally equivalent. However, since that PR also includes another fix I'll close this one as a dupe. Thanks!

PR #6639, opened 2014-03-23
Add missing sha1 import on known_hosts.py

fvieira:
Due to a missing import of the `sha1` module in the `known_hosts.py` file coupled with a catch all `except` (line 105), the git module silently failed to validate hashed fingerprints in the `.ssh/known_hosts`.
This makes it quite hard to use a pre-made `known_hosts` file without using `accept_hostkey`, since the module always complains that the hostkey is unknown even though it is in the `known_hosts` file.


mpdehaan:
merged, thank you!

PR #6640, opened 2014-03-23
ec2: Adding type int for count (in line with other cloud modules)

falloutdurham:
Was having some trouble writing a playbook that span up a specified count of spot instances. A simple fix to make count be an int seems to be a good idea (and is done in other cloud modules).


mpdehaan:
This will break my ability to provision decimal numbers of systems but ...

Just kidding, thanks for this!  :)

PR #2677, opened 2013-04-14
Use platform.node to find out the hostname (#2663)

ciupicri:
Use `platform.node()` instead of `socket.getfqdn()` or
`socket.gethostname` as discussed on bug #2663 [1]

[1] https://github.com/ansible/ansible/issues/2663#issuecomment-16341359

PR #6643, opened 2014-03-24
Add a fact providing the full hostname, without the domain portion

herbygillot:
truncated.

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -108,7 +108,7 @@ class Facts(object):
         self.facts['python_version'] = platform.python_version()
         self.facts['fqdn'] = socket.getfqdn()
         self.facts['hostname'] = platform.node().split('.')[0]
-        self.facts['full_hostname'] = platform.node()
+        self.facts['nodename'] = platform.node()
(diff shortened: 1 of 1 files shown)


mpdehaan:
This was brought up before apparently, with no clear conclusion:

https://github.com/ansible/ansible/pull/2677

https://groups.google.com/forum/#!topic/ansible-project/vpkLqyXFMYE

The worst part of this though is that I don't think many of our users would be able to tell which one to use by looking at the different variable names, which makes this pull request a bit confusing -- and I'm not certain what damage would be done by changing the fqdn fact to return the full platform.node()

This may be very well a case where a local facts module is warranted.

herbygillot:
@mmoya @mpdehaan I am indeed running into the same issue as Anand B. in the referenced thread.

Since socket.get_fqdn and platform.node are taking slightly different code paths, the results aren't the same.

I still think that the full contents of platform.node should be made available as a standard fact.

Since the issue is one of confusion/naming, perhaps the right thing to do here is simply name this fact as `nodename` instead?

mpdehaan:
Works for me, merged, thanks!

herbygillot:
@mpdehaan Thank you.

PR #6646, opened 2014-03-24
Add elasticsearch plugin manager

ada-u:
Hello. At first, I want to say thank you for this awesome app and wait for new release.

I developed a plugin manager of elasticsearch and implemented the feature to setup plugins.
Would you like to merge this module?

Thank you.
##### About elasticsearch plugin

http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/modules-plugins.html

### DIFF
--- a/library/packaging/elasticsearch_plugin
+++ b/library/packaging/elasticsearch_plugin
@@ -1,0 +1,144 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2014, Yuta Adachi <ada-u@ada-u.com>
+#
+# This file is part of Ansible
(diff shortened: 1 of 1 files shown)


mpdehaan:
This may be around the barrier at which we say no to this one, as it may be easier for most people to just shell out to the command.   That being said, this doesn't mean you or others couldn't continue to use this one -- and it's not a bad idea.

I find the need to specify both name and package to be a little confusing in terms of interface.

In ansible, the path to a program is usually called "executable" which may also help.

ada-u:
Thank you for nice help. I'll improve this module, and send better one.

PR #6649, opened 2014-03-24
Accept status code from a list of valid codes.

ghost:
As discussed on the mailing list, it is desirable that that uri module can accept any one of a list of http status codes, rather than just one. 

https://groups.google.com/forum/#!msg/ansible-project/ksesTAi2LEM/rNDneEKBQIkJ

### DIFF
--- a/library/network/uri
+++ b/library/network/uri
@@ -101,7 +101,7 @@ options:
     required: false
   status_code:
     description:
-      - A valid, numeric, HTTP status code that signifies success of the request.
+      - A valid, numeric, HTTP status code that signifies success of the request. Can also be comma separated list of status codes.
     required: false
     default: 200
   timeout:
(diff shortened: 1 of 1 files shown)


ghost:
And here is a usage example:

```
---
- hosts: all
  sudo: yes
  pre_tasks:
  tasks:
   - name: create the admin user (can repeat multiple times)
     action: uri
         url=http://localhost:8080/riak-cs/user
         method=POST
         body='{"email":"admin12@admin.com","name":"admin12"}'
         HEADER_Content-Type=application/json
         status_code=201,409
     register: create_admin_result
     ignore_errors: True
```

mpdehaan:
So this has some debug stderr in here that needs to be removed.

Romeo Theirault was also going to add a new type called "int_list" to do casting, if not added, I would at least be sure to cast all members to ints.

ghost:
Well spotted, and good points. How does this look now?

rrtj3:
Hi @picsolvebryan, thanks for making a go at this. @mpdehaan, this may be a good initial commit until I can get around to sending in the pull request for the int_list type. This should work, but will obviously error if someone passes in something that's not a number.

binarytemple:
Thanks guys, yes @romeotheriault I agree. It will probably be better to have this working now, and then add the stronger type guarantee later.

binarytemple:
@romeotheriault @mpdehaan Is it possible to merge this pull request? I'd like to get some Ansible Galaxy roles out there.

mpdehaan:
Merged, thanks!

Issue #6621, opened 2014-03-21
"[Errno 2] No such file or directory" error when 'validate' parameter on template module sets ENV variable


rwehner:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.5.3
##### Environment:

Running ansible from CentOS 6.5
Managing CentOS 6.5
##### Summary:

When using the 'validate' parameter on the template module it produces an "[Errno 2] No such file or directory" error if the 'validate' argument includes setting a variable.
I suspect this is related to the recent security work around shell variable handling. Perhaps similar to https://github.com/ansible/ansible/issues/6385?
##### Steps To Reproduce:

Install Ansible 1.5.3
Run ansible-playbook with a template task that uses a 'validate' that sets a shell variable. The playbook [here](https://gist.github.com/rwehner/9696857) produces the error. We set the PATH in this way since visudo is at different locations on Linux and Solaris.
##### Expected Results:

No error produced and the shell command in the 'validate' parameter is run as it was in the 1.4 releases.
##### Actual Results:

```
$ ansible-playbook test_sudo.yml

PLAY [localhost] **************************************************************

TASK: [Install sudoers file] **************************************************
failed: [localhost] => {"cmd": ["PATH=/usr/sbin:/usr/local/sbin", "visudo", "-cf", "/usr/local/rpcm/.ansible/tmp/ansible-tmp-1395438209.82-186195729410123/source"], "failed": true, "rc": 2}
msg: [Errno 2] No such file or directory

FATAL: all hosts have already failed -- aborting

PLAY RECAP ********************************************************************
           to retry, use: --limit @/usr/local/rpcm/test_sudo.retry

localhost                  : ok=0    changed=0    unreachable=0    failed=1


```

rwehner:
Updated description with appropriate template format.

mpdehaan:
What is happening here is the validate command here is not being passed through the shell such that any arguments supplied are treated securely (for instance, such that if you referenced {{ user_input }} it couldn't invoke rm -rf, and so forth).

What you can do here, however, is path visudo explicitly, or also set the 'environment' keyword in Ansible.

We should likely indicate in the docs that this command cannot rely on shell functions.

rwehner:
OK. I'll change break my tasks into two (one for Solaris and one for Linux with different explicit paths to visudo).

I've created a PR that hopefully updates the docs to include this information in an acceptable way. https://github.com/ansible/ansible/pull/6652

Thanks for the quick response.

[...]

PR #6652, opened 2014-03-24
Update 'validate=' parameter documentation 

rwehner:
Make clear that the command passed via this parameter is not passed via shell and so all shell features
will not work.

### DIFF
--- a/library/files/copy
+++ b/library/files/copy
@@ -73,6 +73,7 @@ options:
     description:
       - The validation command to run before copying into place.  The path to the file to
         validate is passed in via '%s' which must be present as in the visudo example below.
+        The command is passed securely so shell features like expansion and pipes won't work.
     required: false
     default: ""
     version_added: "1.2"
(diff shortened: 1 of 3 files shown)


rwehner:
Intended to address https://github.com/ansible/ansible/issues/6621

mpdehaan:
merged, thank you!  This will appear in the next docs push.

skyler:
It should be noted that the `lineinfile` documentation specifically uses `visudo` validation as an example, which results in this issue's error.

http://docs.ansible.com/lineinfile_module.html

The example is

```
# Validate a the sudoers file before saving
- lineinfile: dest=/etc/sudoers state=present regexp='^%ADMIN ALL\=' line='%ADMIN ALL=(ALL) NOPASSWD:ALL' validate='visudo -cf %s'
```

```
TASK: [core | enable the wheel group to give users sudo] **********************
failed: [192.168.33.12] => {"cmd": "visudo -cf /tmp/tmpdWvOX6", "failed": true, "rc": 2}
msg: [Errno 2] No such file or directory
```

PR #6654, opened 2014-03-24
Replace absolute imports by relative

davidfischer-ch:


bcoca:
what does this fix?

davidfischer-ch:
Make it more easy to switch to Python 3.

Fix a recursive import:

```
lib/ansible/utils/__init__.py -> import of display() is moved into the methods using it.
```

Exanple code that failed before:

```
from ansible.callbacks import AggregateStats, PlaybookCallbacks, PlaybookRunnerCallbacks
from ansible.playbook import PlayBook
from ansible.utils import VERBOSITY
```

davidfischer-ch:
[WARNING] I found some mistakes not covered by unit-tests. Work in progress!

davidfischer-ch:
Still faulty because I import callbacks and it is also a callbacks variable ... Work in progress.

mpdehaan:
We're not interested in moving to python 3 anytime soon.

RHEL is quite the priority and it is still possible to access python2 on systems where /usr/bin/python is python3 by using the ansible_python_interpreter variable.

davidfischer-ch:
Why do you close this ?

Because it does break nothing in the unit-tests and it does make the code more Python 3 ready at no cost for you ...

Issue #6591, opened 2014-03-20
template module shows diff even if "force=no" was passed


8191:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.6 (devel 991399edf0) last updated 2014/03/20 08:41:39 (GMT +200)
##### Environment:

Host: Fedora 20
Target: OpenBSD 5.4
##### Summary:

When calling ansible-playbook with --diff, a diff is even shown by the template module, if the copy parameter "force=no" was passed and the dest file exists. Actually everything behaves correctly (since copy obeys the force and does not alter the file), but it is quite confusing to see a diff for a file which was not altered.

The template module should suppress the diff if force=no was passed and the dest file exists.
##### Steps To Reproduce:

Create a playbook with the _template_ module and specify an existing dest file and the _force=no_ parameter. Then call the playbook with the _--diff_ parameter of ansible-playbook.

``` yaml
- host: somehost
- tasks:
  - name: create new file but don't touch existing
    template:
      src: foo.j2
      dest: /tmp/existing.file
      force: no
```
##### Expected Results:

No diff should be shown and the dest file should not be touched.
##### Actual Results:

The template module shows a diff between the existing destination file and the generated template file. Nevertheless the file is not being touched (as expected).

mpdehaan:
Hi!

Thanks very much for your interest in Ansible.  It sincerely means a lot to us. 

On September 26, 2014, due to enormous levels of contribution to the project Ansible decided to reorganize module repos, making it easier
for developers to work on the project and for us to more easily manage new contributions and tickets.

We split modules from the main project off into two repos, http://github.com/ansible/ansible-modules-core and http://github.com/ansible/ansible-modules-extras

If you would still like this ticket attended to, and believe this problem or idea is still present in the latest version of Ansible (1.7.2) or the development branch, we will need your help in having it reopened in one of the two new repos, and instructions are provided below.

We apologize that we are not able to make this transition happen seamlessly, though this is a one-time change and your help is greatly appreciated -- 
this will greatly improve velocity going forward.

Both sets of modules will ship with Ansible, though they'll receive slightly different ticket handling.

To locate where a module lives between 'core' and 'extras'
- Find the module at http://docs.ansible.com/list_of_all_modules.html
- Open the documentation page for that module
- If the bottom of the docs say "This is an extras module", submit your ticket to https://github.com/ansible/ansible-modules-extras
[...]

PR #6658, opened 2014-03-24
'changed' key is not defined when task fails

pilou-:
fix KeyError introduced by fix of #6591.

For example when a parameter is misspelled (here `ower` instead of `owner`) , a `KeyError` exception is raised:

```
- name: Generate /etc/hosts
  template: src=hosts dest=/etc/hosts ower=root
```

### DIFF
--- a/lib/ansible/runner/action_plugins/template.py
+++ b/lib/ansible/runner/action_plugins/template.py
@@ -123,7 +123,7 @@ class ActionModule(object):
                 return ReturnData(conn=conn, comm_ok=True, result=dict(changed=True), diff=dict(before_header=dest, after_header=source, before=dest_contents, after=resultant))
             else:
                 res = self.runner._execute_module(conn, tmp, 'copy', module_args, inject=inject, complex_args=complex_args)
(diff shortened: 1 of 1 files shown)


srgvg:
Are you sure nothing changed for every possible occurrence where 'changed' isn't defined in the result?

pilou-:
Using this task:

```
- name: Generate /etc/hosts
  template: src=hosts dest=/etc/hosts ower=root
```

and the following script with `git bisect start HEAD HEAD~250 -- && git bisect run /tmp/script.sh`

```
#!/bin/bash
. /tmp/ansible/hacking/env-setup &> /dev/null
cd /ansible-conf
echo -e "# test" >> roles/test/templates/etc/hosts
ansible-playbook site.yml 2>&1 |grep "ower"
exit ${PIPESTATUS[1]}
```

b8158f5f6a3fc832a1d6e4ba318090e23f3432ac is the first bad commit.

mpdehaan:
The .get() here seems fine, merged, thanks!

PR #6659, opened 2014-03-24
Added category option, to install / remove packages by large category. T...

Jmainguy:
...his is exactly the same as the -Y option to pkgadd.


mpdehaan:
This looks fine but is currently having some conflicts, can you rebase this commit and test?

Seems better that you do this since you're more likely to be using these flags :)

Thanks!

PR #6662, opened 2014-03-24
Introduce set_fact module in var docs

gblahaerath:
When I was trying to solve the problem discussed in the thread https://groups.google.com/forum/#!topic/ansible-project/wqiclLkqmGU I looked at the variable documentation first.  I added this bread crumb since it seemed like a good place for it.


mpdehaan:
While set_fact can be useful for various shorthanding of things like foo.stdout, I'm not clear on the pre_tasks example referenced above, and I'd like to discourage usage in _MOST_ cases, leaving set_fact around for more niche use cases.

I'm inclined to not mention this one here as in most cases Ansible shouldn't need usage of set_fact (and I'm not a great fan of the module -- though it has it's uses in those particular cases!)

PR #6664, opened 2014-03-24
intro_adhoc: some improvements on readibility

chikamichi:
When one reads the Ansible doc in order, this is the first occurence of `state=` in adhoc commands. It may be a bit confusing without any explanations.

There are a few other things in the same vein, covered in this copy-editing PR.


mpdehaan:
This pull request introduces some gramatical errors, so I'm not going to include this one.

PR #6665, opened 2014-03-25
intro_installation.rst: some copy-editing

chikamichi:
Just a few things to make it more readable and consistent.


mpdehaan:
Some of these are ok, but some of these I disagree with stylistically.

As writing is somewhat of a personal thing, I'm not going to include this one.

chikamichi:
Fair enough :)

Issue #6538, opened 2014-03-17
Group vars overwritten when a host is in multiple inventories


darkone23:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible-playbook 1.5.3
##### Environment:

Debian, but shouldn't matter
##### Summary:

Ansible seems to be wrecking inventory vars when a host appears in multiple groups. It looks like it is impossible to figure out which vars it will use if you have a single host in multiple hostgroups.
##### Steps To Reproduce:

Run this playbook with the provided inventory:

Inventory file:

``` ini
[a_local]
localhost

[b_local]
localhost

[c_local]
localhost

[a_local:vars]
foo=one

[b_local:vars]
foo=two

[c_local:vars]
foo=three
```

Playbook:

``` yaml
- hosts: a_local
  gather_facts: no
  tasks: [{debug: var=foo}]

- hosts: b_local
  gather_facts: no
  tasks: [{debug: var=foo}]

- hosts: c_local
  gather_facts: no
  tasks: [{debug: var=foo}]
```
##### Expected Results:

The debug prompt should show 'one', then 'two', then 'three'.
##### Actual Results:

```
$ ansible-playbook -i inventory.ini -c local site.yaml                                                                                                             +[master]

PLAY [a_local] **************************************************************** 

TASK: [debug var=foo] ********************************************************* 
ok: [localhost] => {
    "foo": "three"
}

PLAY [b_local] **************************************************************** 

TASK: [debug var=foo] ********************************************************* 
ok: [localhost] => {
    "foo": "three"
}

PLAY [c_local] **************************************************************** 

TASK: [debug var=foo] ********************************************************* 
ok: [localhost] => {
    "foo": "three"
}

PLAY RECAP ******************************************************************** 
localhost                  : ok=3    changed=0    unreachable=0    failed=0   
```

darkone23:
Filled out with a template

mpdehaan:
This is a common misconception.

Groups can describe geography, purpose, or environment.    When variables are all at the same level, there is no defined order.

As an example of this, if you select machines that are webservers, the variables from machines in "phoenix" should still be loaded in -- not just the variables for the "webservers" group.

bcoca:
your expectation is incorrect, expected behavior is that the var will have
the value of the last group loaded.

darkone23:
[...]

PR #6666, opened 2014-03-25
Use host-pattern for inventory var resolution

darkone23:
Fixes unintuitive behavior described in https://github.com/ansible/ansible/issues/6538

Prior behavior was to flatten variables by hostname. This meant that if a var was declared in multiple groups on a single host you would not be able to know which value of the var ansible would use.

This pull request utilizes the host pattern to prioritize the vars in matching hostgroups over non-matching hostgroups. This reduces ambiguous/undefined behavior in favor of an intuitive resolution order.

Prior resolution order was:
1. flattened vars from host class
2. vars from plugins
3. host vars from parser

This commit changes the resolution order:
1. flattened vars from host class
2. group vars from hosts pattern :sparkles: **_new step_** :sparkles:
3. vars from plugins
4. host vars from parser

In my local tests everything is completely backwards compatible, and we are using this fix in production currently.  The use case in #6538 also works.

Thanks, let me know what I can do to help to get this merged,

Thomas Omans
Commission Junction

### DIFF
--- a/lib/ansible/inventory/__init__.py
+++ b/lib/ansible/inventory/__init__.py
@@ -37,9 +37,9 @@ class Inventory(object):
     Host inventory for ansible.
     """
 
-    __slots__ = [ 'host_list', 'groups', '_restriction', '_also_restriction', '_subset', 
-                  'parser', '_vars_per_host', '_vars_per_group', '_hosts_cache', '_groups_list',
-                  '_pattern_cache', '_vars_plugins', '_playbook_basedir']
+    __slots__ = [ 'host_list', 'groups', '_restriction', '_also_restriction', '_subset', 'parser',
+                  '_vars_per_host', '_vars_per_group', '_vars_per_pattern', '_hosts_cache',
+                  '_groups_list', '_pattern_cache', '_vars_plugins', '_playbook_basedir']
 
     def __init__(self, host_list=C.DEFAULT_HOST_LIST):
 
@@ -50,11 +50,12 @@ class Inventory(object):
         # caching to avoid repeated calculations, particularly with
         # external inventory scripts.
 
-        self._vars_per_host  = {}
-        self._vars_per_group = {}
-        self._hosts_cache    = {}
-        self._groups_list    = {} 
-        self._pattern_cache  = {}
+        self._vars_per_host    = {}
+        self._vars_per_group   = {}
+        self._vars_per_pattern = {}
+        self._hosts_cache      = {}
+        self._groups_list      = {}
+        self._pattern_cache    = {}
 
         # to be set by calling set_playbook_basedir by ansible-playbook
         self._playbook_basedir = None
@@ -353,12 +354,42 @@ class Inventory(object):
             raise Exception("group not found: %s" % groupname)
         return group.get_variables()
 
-    def get_variables(self, hostname, vault_password=None):
-        if hostname not in self._vars_per_host:
-            self._vars_per_host[hostname] = self._get_variables(hostname, vault_password=vault_password)
-        return self._vars_per_host[hostname]
+    def get_variables(self, hostname, vault_password=None, pattern=None):
+        vars = {}
+
+        host_vars = self._get_variables(hostname, vault_password=vault_password)
+        pattern_vars = self._get_variables_from_pattern(pattern)
+
+        vars.update(host_vars)
+        vars.update(pattern_vars)
+
+        return vars
+
+    def _get_variables_from_pattern(self, pattern=None):
+        if pattern in self._vars_per_pattern:
+            return self._vars_per_pattern[pattern]
+
+        vars = {}
+
+        if not pattern:
+            return vars
+
+        patterns = pattern.replace(";",":").split(":")
+        patterns = filter(lambda x: not x.startswith("!"), patterns)
+        patterns = map(lambda x: x.replace("&", ""), patterns)
+
+        for group in map(lambda x: self.get_group(x), patterns):
+            if group:
+                vars.update(group.get_variables())
+
+        self._vars_per_pattern[pattern] = vars
+
+        return vars
 
     def _get_variables(self, hostname, vault_password=None):
+
+        if hostname in self._vars_per_host:
+            return self._vars_per_host[hostname]
 
         host = self.get_host(hostname)
         if host is None:
@@ -373,6 +404,8 @@ class Inventory(object):
         vars = utils.combine_vars(vars, host.get_variables())
         if self.parser is not None:
             vars = utils.combine_vars(vars, self.parser.get_host_variables(host))
+
+        self._vars_per_host[hostname] = vars
         return vars
 
     def add_group(self, group):
--- a/lib/ansible/inventory/__init__.py
+++ b/lib/ansible/inventory/__init__.py
@@ -354,22 +354,15 @@ class Inventory(object):
             raise Exception("group not found: %s" % groupname)
         return group.get_variables()
 
-    def get_variables(self, hostname, vault_password=None, pattern=None):
-        vars = {}
-
-        host_vars = self._get_variables(hostname, vault_password=vault_password)
-        pattern_vars = self._get_variables_from_pattern(pattern)
-
-        vars.update(host_vars)
-        vars.update(pattern_vars)
-
-        return vars
-
-    def _get_variables_from_pattern(self, pattern=None):
+    def _get_pattern_variables(self, pattern=None):
         if pattern in self._vars_per_pattern:
             return self._vars_per_pattern[pattern]
 
-        vars = {}
+        group = self.get_group('all')
+        if group:
+            vars = group.get_variables()
+        else:
+            vars = {}
 
         if not pattern:
             return vars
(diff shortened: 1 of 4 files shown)


darkone23:
Here's an example of the problem this patch solves,

Consider the following inventory:

```
[web]
my-only-server.example.com

[db]
my-only-server.example.com

[web:vars]
ansible_sudo=yes
ansible_sudo_user=web

[db:vars]
ansible_sudo=yes
ansible_sudo_user=db
```

And playbook:

``` yaml
- hosts: web
  roles: [ web-stuff ]  

- hosts: db
  roles: [ db-stuff ]
```

Prior to the patch the first play would run as the db user.
After the patch it runs as the web user.

All previous behavior is intact,
see: https://github.com/ansible/ansible/pull/6666/files#diff-b7ecd023ab82518ff60c8713109dd9f6R393

mpdehaan:
Not sure what the IRC comment is about, but this proposed behavior is not one which is desired in Ansible - and is what I was already mentioning in the previous ticket.

For instance, a host group can represent the purpose of the system, the hardware type, or the geography, and the system could be in all three groups.

There's nothing saying that when targetting by geography the geography variables should win over the role variables, etc, and we should not be inserting extra semantics here.

(I believe on IRC (or the list) I said I _MIGHT_ be interested in seeing a patch that emitted a warning when a variable is sourced to the host via multiple groups, but this warning itself might be too spammy to include in the system, and the code to support this might be too involved)

darkone23:
@mpdehaan I think perhaps we are talking about different things?

_"There's nothing saying that when targetting by geography the geography variables should win over the role variables, etc, and we should not be inserting extra semantics here."_

There's nothing in this pull request that violates this statement so I'm not sure why this point is relevant. The role variables still take precedence over the inventory variables, exactly like they did before. The only change is host groups matching your host pattern take precedence over groups not matching your pattern.

Futher, the `hosts` key is not _geography_ it is _hostgroup_. While you are free to store geography information, the official examples show it as application groups: http://docs.ansible.com/intro_inventory.html#hosts-and-groups

It is certainly reasonable for applications to run on the same host. The current behavior of the inventory makes this impractical and unintutive.

Again, this is pull request is completely backward compatible. It doesn't change the way roles behave, it doesn't change the way playbooks behave, it changes the order in which inventory variables resolve.

Can you maybe explain why the above pull request is incompatible with ansibles design?

darkone23:
Here is why inventory var scoping is important:

Consider this inventory:

```
[web]
funny.example.com ansible_python_interpreter=/funny/bin/python
regular.example.com

[web:vars]
ansible_python_interpreter=/usr/bin/python
```

Without inventory scoping `funny.example.com` uses the incorrect value for `ansible_python_interpreter` (it uses the copy from web:vars instead of host vars).

This scoping fix resolves this issue.

mpdehaan:
"The only change is host groups matching your host pattern take precedence over groups not matching your pattern."

This is what I was posting above.

Don't think of roles exactly, but the idea of groups that are used to assign role-ishness

The problem you just illustrated above (host variable obscured by group variable) is a bug -- and  different than what this ticket is about.  A seperate pull request that deals with fixing just that would be fine, but this ticket conflates those two questions.

oppianmatt:
It's suggested in the best practices guide to assign roles via groups:

http://docs.ansible.com/playbooks_best_practices.html

> It is suggested that you define groups based on purpose of the host (roles) and also geography or datacenter location (if applicable)

But if I put the same host in multiple groups since it does multiple roles that might share the same variable. For example logrotate from ansible galaxy. If the host is in the dbservers group I want to create a variable for the dbservers group to rotate the database logs. If a host is in the web group, then rotate the web logs. But if a host is in both groups how are we meant to deal with this? 

Now I got to merge all my `logrotate_scripts` variable into one and it will contain entries for hosts that might not even have the log files to rotate. 

Otherwise can you tell me how to solve the problem? I'm only going by the best practices guide.

Some code:

```
# site.yaml
- hosts: webservers
  roles:
    - { role: nginx, sudo: yes, tags: [ 'nginx' ] }
    - role: nickhammond.logrotate
      sudo: yes
      tags: logrotate

- hosts: dbservers
  roles:
    - { role: postgresql, sudo: yes, tags: [ 'postgresql', 'postgres' ] }
    - role: nickhammond.logrotate
      sudo: yes
      tags:
        - logrotate
        - postgresql
        - postgres

```

Now where do I put these?

```
logrotate_scripts:
  - name: nginx
    path: /var/log/nginx/*.log
    options:
      - daily
      - rotate 7
      - missingok
      - compress
      - delaycompress
      - create 640 nginx adm
      - sharedscripts
    scripts:
      postrotate: "[ -f /var/run/nginx.pid ] && kill -USR1 `cat /var/run/nginx.pid`"
```

and

```
logrotate_scripts:
  - name: postgresql-common
    path: /var/log/pg.log
    options:
      - rotate 7
      - size 10G
      - missingok
      - compress
      - delaycompress
    scripts:
      postrotate: "reload rsyslog >/dev/null 2>&1 || true"
```

If I put them in group files like dbservers.yaml and webservers.yamls and the host is in both groups then it won't work. 

Sorry if this is rehashing old information but it is counterintuitive and if you follow the advice on the best practices it doesn't work. It becomes a struggle to manage.

It should be the group specified in the play has precedent when variables clash, like the PR proposes which shouldn't break anything as people wouldn't be using the same host in multiple groups with the same variable now since it doesn't work, the behaviour is unknown presently.

Also could the best practices or faq be updated with this very scenario of group variables clashing and how to solve it please?

sebastianreloaded:
Jep, i would like to know a best practice too. I have a frontend server group and a api server group. it happens that both are handled on the same nginx webserver. I cant add both nginx configs to the webserver through the group_vars because they are overwritten.

eik3:
Thanks for this patch Thomas!

We are facing the exact same problem when trying to provision more multiple apps (which include a shared nginx role) onto one host. We tried the CNAME approach discussed in https://github.com/ansible/ansible/issues/9065 but that multiplies the Ansible runtime and leads to conflict errors resulting from Ansible connecting to the same host (through CNAMES) in parallel.

Unfortunately, Ansible has changed quiet a bit in the last 1.5 years and together with my limited Python knowledge I wasn't able to apply the patch to 1.9.2. 

Does anyone use this patch at the moment? I would love to try it out on the current Ansible stable.

If you're not using the patch, how did you get around this problem?

bcoca:
When i supported multiple apps on nginx with the same role, i just made
each app write it's own file using the appname, which was a var passed to
the role. If configured this way it avoids conflicts and overwrites.
- {role: nginx, appname=myapp, appsettings=appsettings_dict}

...

in role:
- template: src=generic_app_template.j2
  dest=/etc/nginx/conf.d/{{appname}}.conf

## 

Brian Coca

tcosta84:
Sorry for bringing back such an old issue but I really would like to understand it better and maybe learn to do this in the best way.

Here is what I'm trying to accomplish:

I want to configure Monit (a monitoring tool) to create two checks, one for the Nginx service and the other one for the Postgres service. However, both services are currently in the same host and that's probably why I keep seeing this odd behaviour where Ansible is using the `group_vars/dbservices` for both the `webservers.yml` and the `dbservers.yml` instead of using `group_vars/webservers` for `webservers.yml` and `group_vars/dbservers` for `dbservers.yml`.

Here is my inventory file:

```
# hosts/local
[webservers]
192.168.88.10

[dbservers]
192.168.88.10

[local:children]
webservers
dbservers
```

And here are my playbooks:

```
# webservers.yml
---
- hosts: webservers
  remote_user: root
  roles:
    - acotel.monit

# dbservers.yml
---
- hosts: dbservers
  remote_user: root
  roles:
    - acotel.monit
```

And here are my group_vars:

```
# groups_vars/webservers
---
monit_services:
 'nginx':
    - 'check process nginx with pidfile /var/run/nginx.pid'
    - 'start program = "/etc/init.d/nginx start"'
    - 'stop program  = "/etc/init.d/nginx stop"'

# groups_vars/dbservers
---
monit_services:
 'postgres':
    - 'check process pgsql with pidfile /var/lib/pgsql/data-9.4.4/postmaster.pid'
    - 'start program = "/etc/init.d/pgsql start"'
    - 'stop  program = "/etc/init.d/pgsql stop"'
```

I really would like to keep this structure (multiple playbooks and group_vars) even if it seems a little complex for such a small installation but this is considered a Best Practice by the Ansible Team and I'm planning to move the Postgres service to a separate host in the near future anyway.

Can you guys help me? I know I can use role variables as described by @oppianmatt:

```
# webservers.yml
---
- hosts: webservers
  remote_user: root
  roles:
    - { role: acotel.monit, monit_services: monit_webservers }

# dbservers.yml
---
- hosts: dbservers
  remote_user: root
  roles:
    - { role: acotel.monit, monit_services: monit_dbservers }
```

but the group_vars/webservers solution seems more "correct".

Thanks!

jagibson:
@bcoca 

I think merging this would be valuable.  Many of the cloud modules are run against localhost (or your deploy server) with different variable sets.  If this was added I could deploy my changes in parallel.  Without it I have to resort to include_vars and multiple plays which just isn't as elegant and takes longer to run.

How I do this now: (ugly)
playbook:

```
- hosts: localhost
  gather_facts: false
  pre_tasks: 
    - include_vars: "group_vars/node1.yml"
  roles:
    - role: aws_deploy_instance

- hosts: localhost
  gather_facts: false
  pre_tasks: 
    - include_vars: "group_vars/node2.yml"
  roles:
    - role: aws_deploy_instance
```

With the group_vars/nodeX.yml files containing the instance-specific settings.

What I would like to do:
inventory/hosts:

```
[node1]
localhost

[node2]
localhost
```

playbook:

```
- hosts: node1:node2
  gather_facts: false
  roles:
    - role: aws_deploy_instance
```

I find it counter-intuitive that this does not work.  As others have said- the best practices guide suggests assigning variables via groups.

Please reconsider adding this to Ansible.

alex-sherwin:
A bit surprising to run across this, one would just assume variables have scope/hierarchy for resolution precedence

It's pretty disappointing that Ansible has no plans to remedy it; the comments in the past seem to have an undertone of "I think it's fine to flatten everything and your use case doesn't matter"

Would be nice to see this fixed

jagibson:
I think I've found a better pattern for doing this by using delegate_to:

```
- hosts: node1:node2
  gather_facts: false
  roles:
    - role: aws_deploy_instance
      delegate_to: localhost
      delegate_facts: yes
```

Then put host-specific variables in `host_vars/node1.yml` and `host_vars/node2.yml`

Idanatcox:
Any chance to reconsider merging this?
I got the same issue where same hosts serves 2 different purposes and should logically be different group.
Thanks

pcsmits:
Same issues. Please merge this

cloudnull:
+1 I'd love to see this become a reality in Ansible

darkone23:
This code can no longer be merged -- the ansible codebase has changed too much, and the patch was only used temporarily at CJ.

The solution I used years ago was to write an [alternate config server](https://github.com/eggsby/oroboros) and use ansible's dynamic inventory functionality, always assuming everything was a global variable -- since ansible only likes the vars to be global and not scoped in any way beyond the server hostname.

I hope a solution can be found since the existing behavior is ambiguous and non-obvious, but I don't think re-opening this issue is the way to go.

Hopefully someone can provide a patch in the future now that @mpdehaan has passed the torch on the ansible project, since as I understand it he was the only one strongly against this functionality. 

I believe [what I posted earlier](https://github.com/ansible/ansible/pull/6666#issuecomment-38528086) is a perfect summary of the 'bug' using only ansible behavior, and if someone is invested in following this issue to resolution I recommend they open that statement as a new bug against the project.

Cheers 
-- Thomas

cloudnull:
+1, I've been hacking on the inventory libs recently trying to re-implement the changes for the ansible 2.x code base (nothing yet as like you said the code has changed greatly). Additionally my colleague has added this feature idea https://github.com/ansible/ansible/issues/17236 which is within the same vain as this issue/PR. Thanks again @eggsby.

ReSearchITEng:
I've summarized the options here: https://github.com/ansible/ansible/issues/9065#issuecomment-379246209

PR #6671, opened 2014-03-25
svr4pkg supports the -G switch for pkgadd

dermute:
By standard, pkgadd adds an package in the global zones and all non-global zones. But with an configuration management, you maybe dont want to have this behavior. pkgadd supports the -G switch, which triggers an installation only in the current zone. This pullrequest enables this switch for the package svr4pkg, but doesn't set is as default (it would break Solaris 9 support)

This pull request is nearly identical to the request ansible/5873 (with fixed default behaviour)

### DIFF
--- a/library/packaging/svr4pkg
+++ b/library/packaging/svr4pkg
@@ -64,6 +64,7 @@ options:
     required: false
     default: "all"
     choices: ["current", "all"]
+    version_added: "1.6"
 '''
 
 EXAMPLES = '''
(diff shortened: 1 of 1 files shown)


mpdehaan:
This needs to add the "version_added" field for the new parameter in the documentation block.

dermute:
Okay, I added the field.

mpdehaan:
merged, thank you!

PR #6674, opened 2014-03-25
Add register_as_list/hash

ant31:
cleaner version of: https://github.com/ansible/ansible/pull/6488

Added 2 new keywords:

register_as_list: boolean
register_as_dict: hashkey
They both require the 'register' keyword to be present.
Its define the data structure,

register_as_list: yes => appends the result to a list
register_as_dict: key => adds the result to a hash[key]

exemple:

``` yaml
- name: Append result in a list
  template: src= a.conf.j2 dest=/root/a.conf backup=yes
  register: foo_list
  register_as_list: yes

- name: template1
  template: src=mytpl1.conf.j2 dest=/root/mytpl1.conf backup=yes
  register: bar_dict
  register_as_dict: key1

- name: template2
  template: src=mytpl2.conf.j2 dest=/root/mytpl2.conf backup=yes
  register: bar_dict
  register_as_dict: key2
```

use the registred results:

``` yaml
- name: display results (register_as_list)
  debug: msg={{item}}
  with_items: foo_list

- name: display result template1
  debug: msg={{bar_dict['key1']}}
```

### DIFF
--- a/lib/ansible/playbook/__init__.py
+++ b/lib/ansible/playbook/__init__.py
@@ -489,6 +489,18 @@ class PlayBook(object):
             # rather than the setup cache - vars should be transient between playbook executions
             if 'stdout' in result and 'stdout_lines' not in result:
                 result['stdout_lines'] = result['stdout'].splitlines()
+            if task.append_to_list:
+                if task.register not in self.VARS_CACHE[host]:
+                    self.VARS_CACHE[host][task.register] = []
+                result_list = self.VARS_CACHE[host][task.register]
+                result_list.append(result)
+                result = result_list
+            elif task.insert_into_hash:
+                if task.register not in self.VARS_CACHE[host]:
+                    self.VARS_CACHE[host][task.register] = {}
+                result_hash = self.VARS_CACHE[host][task.register]
+                result_hash[task.insert_into_hash] = result
+                result = result_hash
             utils.update_hash(self.VARS_CACHE, host, {task.register: result})
 
         # add facts to the global setup cache
--- a/lib/ansible/playbook/task.py
+++ b/lib/ansible/playbook/task.py
@@ -27,8 +27,8 @@ class Task(object):
     __slots__ = [
         'name', 'meta', 'action', 'when', 'async_seconds', 'async_poll_interval',
         'notify', 'module_name', 'module_args', 'module_vars', 'default_vars',
-        'play', 'notified_by', 'tags', 'register', 'role_name',
-        'delegate_to', 'first_available_file', 'ignore_errors',
+        'play', 'notified_by', 'tags', 'register', 'append_to_list', 'insert_into_hash',
+        'role_name', 'delegate_to', 'first_available_file', 'ignore_errors',
         'local_action', 'transport', 'sudo', 'remote_user', 'sudo_user', 'sudo_pass',
         'items_lookup_plugin', 'items_lookup_terms', 'environment', 'args',
         'any_errors_fatal', 'changed_when', 'failed_when', 'always_run', 'delay', 'retries', 'until',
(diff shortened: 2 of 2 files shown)


mpdehaan:
I'm not sold on the proposed syntax here and will need to think about this.

mpdehaan:
Perhaps "append_to_list" and "insert_into_hash" ?

ant31:
The version proposed use the register variable and change directly its type to list or hash. 
With this behavior, I Think it is more clear with the keywords 'register_as_list / register_as_hash', like this you know it's related to the register var. 

I'm fine with append_to_list, insert_into_hash want me to update the code?

mpdehaan:
yes, please, that sounds good.  Looks like this could also use a rebase so we could more easily merge it in.

Thanks!

ant31:
Rebased and Commited

test exemple:

``` yaml
- hosts: server
  tasks:
    - name: Append first result
      shell: "echo a"
      register: foo_list
      append_to_list: yes

    - name: Append second result
      shell: 'echo b'
      register: foo_list
      append_to_list: yes

    - name: Add result to dict with key 'c'
      shell: 'echo key is c'
      register: bar_dict
      insert_into_hash: 'c'

    - name: Add result to dict with key 'd'
      shell: 'echo key is d'
      register: bar_dict
      insert_into_hash: 'd'

    - name: Append third result
      shell: 'echo c'
      register: foo_list
      append_to_list: yes

    - name: display list element index 2
      debug: msg="{{foo_list[2]}}"
# ok: {"msg": "c"}

    - name: display dict, key 'd'
      debug: msg="{{bar_dict['d'].stdout}}"
# ok: {"msg": "key is d"}

    - name: display all values in list
      debug: msg="item= {{item}}"
      with_items: foo_list
#  ok: {"msg": "a"}
#  ok: {"msg": "b"}
#  ok: {"msg": "c"}

    - name: display all values in dict
      debug: msg="bar_dict[{{item.key}}] = {{item.value.stdout}}"
      with_dict: bar_dict
# ok: { "msg": "bar_dict[c] = key is c"}
# ok: { "msg": "bar_dict[d] = key is d"}
```

ant31:
We have currently forked ansible to use this feature in my company because we needed it.
Do you have any eta or version 1.8.x, 1.9?

The use case is:
- after any file change, backup to oldfile to a storage server.

tasks/main.yml

``` yaml
- name: send projectA configuration                                                            
  template: src=projectA.conf.j2 dest=/etc/projectA/conf backup=yes                
  register: new_conf                                                            
  append_to_list: yes                                                        
  notify:                                                                       
     - backup conf    

- name: send projectB configuration                                                            
  template: src=projectB.conf.j2 dest=/etc/projectB/conf backup=yes                
  register: new_conf                                                             
  append_to_list: yes                                                        
  notify:                                                                       
     - backup conf                                                                                                                    
```

handlers/main.yml

``` yaml
- name: backup conf
  synchronize: mode=pull src={{item.backup_file}} dest=/archive/{{inventory_hostname}}/{{item.backup_file | basename}}
  delegate_to: backup-server-1
  when: '{{item.backup_file is defined}}'
  with_items: new_conf
  register: backup
```

This way we write only one unique handler. 
All other solutions with the current ansible version require to write one task or one handler per file to backup. Since we have a lot of templates, it would add too much noise in roles.

Thank you

cemo:
@mpdehaan any chance to include this in next release? By the way could someone please give a workaround for currently?

ant31:
Hi again, 

The PR have now soon one year, I still don't see how to do DRY handler without it. 

I have already rebased already once in septembre to fix 'new' conflicts. 
It is now in merge conflct a second time. 
I can fix it again but only do you have any intention to merge it? Otherwise it's probably better to close the PR. 

Thank you.

cemo:
I am really concerned about future of Ansible.

antlegrand:
They are too busy trying to sell Tower :)

bcoca:
so, sorry for the delay, there are many tickets and this one is labled p5 (low priority)

We discussed this in the core team and think a better approach is allowing to register to specific variable sub references (this does not currently work):

```
- hosts: localhost
  gather_facts: false
  connection: local
  vars:
        my_output:
                ls: { }
         list_stuff: []
  tasks:
        - shell:  ls
          register: my_output['ls']

        - shell:  ls
          register: list_stuff[]
```

which seems much cleaner than adding the new directives

ant31:
Hi bcoca, 

I may spend time on it again but will you raise the priority? Moreover there is real use cases on this features. 

About your comment, I agree the syntax is cleaner, but I would prefere to not have to define the vars before using it.  I propose to try to find references to the var name and if not exists to create it with the right type (hash or list). 

In short there would be 2 new naming conventions to register results: 
- varname[]         : list
- varname['key']  : hash

ant31:
Ok, I'm reopening the PR with the syntax updated. 

Comment: 
- The regexp match only hash key without quote/doublequote/space, is that ok?
- I overwrite previous register variables in case of type mismatch, I'm doing because it's the current behavior. Otherwise I can raise an error, let me know.

## 

Antoine

ant31:
reopened in https://github.com/ansible/ansible/pull/10325

PR #6675, opened 2014-03-25
add limiting Inventory to a list of 1 or more VPCs

danjenkins:
Due to limitations in the EC2 API you can't limit DescribeInstance

However, with the addition of VPCs you may want to limit yourself
to a particular VPC for a particular environment

This allows you to specify which VPC(s) you want to describe

My Python skills aren't great so there may be easier ways to do what I've done.

### DIFF
--- a/plugins/inventory/ec2.ini
+++ b/plugins/inventory/ec2.ini
@@ -42,6 +42,10 @@ route53 = False
 # 'route53_excluded_zones' as a comma-separated list.
 # route53_excluded_zones = samplezone1.com, samplezone2.com
 
+# You can also specify a list of VPCs you want to limit your instance lookup to
+# If you switch between VPCs then you'll need to refresh your cache too
+# vpcs = vpc-<id>
+
 # API calls to EC2 are slow. For this reason, we cache the results of an API
 # call. Set this to the path you want cache files to be written to. Two files
 # will be written to this directory:
--- a/plugins/inventory/ec2.py
+++ b/plugins/inventory/ec2.py
@@ -222,6 +222,11 @@ class Ec2Inventory(object):
             self.route53_excluded_zones.extend(
                 config.get('ec2', 'route53_excluded_zones', '').split(','))
 
+        if config.has_option('ec2', 'vpcs'):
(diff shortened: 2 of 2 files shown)


pas256:
While this approach will work, I would prefer using the Patterns approach Ansible has had since the beginning. 
http://docs.ansible.com/intro_patterns.html

Where in your playbook, you can use set operations to target only machines in a specific VPC.

Right now, the EC2 inventory plugin doesn't group instance into VPCs, but it could, and that is actually a great feature to add.

Take a look at how the plugin creates groups or regions, availability zones, security groups, etc here:
https://github.com/danjenkins/ansible/blob/add-vpc-limitation/plugins/inventory/ec2.py#L426

jjshoe:
@pas256 @mpdehaan - What would it take to get this merged?

pas256:
Nothing on my end.

I do believe #6676 solves the same problem as this, but in a cleaner way, so I would opt to merge that, and then close this. 

To active the same result, use:

```
ansible-playbook -l vpc_id_vpc-1234abcd my-playbook.yml
```

pas256:
```
$ ansible-playbook -h

Options:
  ... 
  -l SUBSET, --limit=SUBSET
                        further limit selected hosts to an additional pattern
```

You pass in any pattern you like to it:
http://docs.ansible.com/intro_patterns.html

So if you have tagged in instance with `Name=Test 1`, you could use:

```
ansible-playbook -l tag_Name_Test_1 my-playbook.yml
```

cchurch:
# Closing This Ticket

[#8822](https://github.com/ansible/ansible/pull/8822) added support for using the filters available to [DescribeInstances](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/ApiReference-query-DescribeInstances.html#query-DescribeInstances-filters).

You can now use [`instance_filters` in ec2.ini](https://github.com/ansible/ansible/blob/devel/plugins/inventory/ec2.ini#L77) to filter by VPC, e.g.:

```
instance_filters = vpc-id=vpc-ab456cde
```

We believe the change described above should resolve this problem for you.

issue is resolved.

Thank you!

Issue #6579, opened 2014-03-19
ansible-vault 1.5.3 cannot decrypt files from 1.5.0


zytek:
##### Issue Type:

Bug Report
##### Ansible Version:

1.5.3
1.6
##### Environment:

OS X
##### Summary:

ansible-vault 1.5.3 (tested also HEAD) cannot decrypt files encrypted with ansible-vault 1.5.0.
##### Steps To Reproduce:

Try to decrypt file encrypted with ansible 1.5.0
##### Expected Results:

File decrypted.
##### Actual Results:

```
ERROR: Decryption failed
```

mpdehaan:
We have unit tests for this, so this is interesting. 

Is there any chance you might have specified the wrong password?  (i.e. can you check decryption with the older version, etc?)

mpdehaan:
also, can you test 'edit' as well as 'decrypt' ?

zytek:
I have checked both edit, decrypt and also using password from file. 
Same issue on my friend's box.

Bonus issue: even when we encrypt files again with new vault (1.1, AES256) then ansible-playbook with --vault-password-file fails. Running more tests now, will update with command output.

jctanner:
@zytek how are you installing ansible and testing the various versions?

Can we also see the output of these two commands?

```
python -c "import site; print site.getsitepackages()"
python -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())"
```

zytek:
Seems like it does not work only with long passwords.
Password in file has 50 chars. Works fine with 3 letter password.

```
bash-3.2$ brew switch ansible 1.5.0
Cleaning /usr/local/Cellar/ansible/1.5.0
Cleaning /usr/local/Cellar/ansible/HEAD
101 links created for /usr/local/Cellar/ansible/1.5.0
bash-3.2$ ansible --version
ansible 1.5
bash-3.2$ ansible-vault --vault-password-file=/Users/zytek/work/secure/ansible/vault_main_pass create test
bash-3.2$ cat test 
$ANSIBLE_VAULT;1.0;AES
53616c7465645f5f4390a0ca2764a1bedecb4304b6d7f528e95f6a20af89038941c37ce2ac970cc5
bfbef6aa0d4a63b799452df478ad58a72259851738bbff2ed45729c6231e4219d879dd0b1eb02f0f
e64931037c68cb469b2e17d7e3a987fc
bash-3.2$ brew switch ansible HEAD
[...]

PR #6681, opened 2014-03-25
Properly revert changes introduced in #6579 

zytek:
While reverting changes @jctanner forgot to revert strip() removal which made ansible-vault and ansible{,-playbook} incompatible. This fixes it.


jctanner:
@zytek thank you!

PR #6686, opened 2014-03-25
Adds groups example for user module

sethwoodworth:
The expected syntax for the groups argument was unclear to me. I found the answer on [serverfault](http://serverfault.com/questions/566756/how-can-i-add-a-user-to-multiple-groups-in-ansible) and thought folks would rather find it here.


mpdehaan:
I'll see about adding something without the space and long form version.  Thanks for heads up!

PR #6689, opened 2014-03-25
Funny grammar fixed plus added detail

jjshoe:
Found funny grammar while trying to find out if the cache update happens before or after you add the repo. I believe lines 377 and 378 confirm that the source is added before a cache update.


mpdehaan:
I have some preferences on wording here that I'd like to apply, but leaving this open for now.  Thanks!

mpdehaan:
Fixes applied, thanks for the report!

PR #6690, opened 2014-03-25
Adds to rax: bypass url and scheduler hints.

nicholaskuechler:
Adds the following to the rax module:
- Ability to specify bypass_url feature of python-novaclient
- Ability to specify scheduler_hints feature of python-novaclient


sivel:
I am actually working on an alternate solution for this.  I'm not sure we want to directly expose these options to the world as is.

You can see that work at:

https://github.com/sivel/ansible/compare/rax-extra-args

nicholaskuechler:
Closing in favor of sivel's implementation.

PR #6695, opened 2014-03-26
New module to trigger build in Jenkins

muffl0n:
##### Issue Type:

New Module Pull Request
##### Ansible Version:

N/A
##### Environment:

N/A
##### Summary:

This module allows the user to trigger a job in Jenkins. Nothing more, nothing less. :)
##### Steps To Reproduce:

N/A
##### Expected Results:

N/A
##### Actual Results:

N/A

### DIFF
--- a/library/web_infrastructure/jenkins
+++ b/library/web_infrastructure/jenkins
@@ -1,0 +1,101 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2014, Sven Schliesing <s.schliesing@ndr.de>
+#
+# This file is part of Ansible.
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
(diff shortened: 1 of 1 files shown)


mpdehaan:
I've seen a lot of Jenkins triggering ansible, but the other way around is a bit unexpected.

Can you share a bit about the use case and how you are using this in your environment?

muffl0n:
We are using SVN to store the config files for our tomcats. When committing to this SVN project a deployment to our staging servers get's triggered in Jenkins. The job in Jenkins does some basic syntax checking and then uses rsync to mirror these files to all known servers.

When installing a new staging-server with ansible we start the tomcat at the very end. But before the tomcat can be started the config files need to be mirrored to the newly installed server. So we trigger this job in Jenkins before starting the tomcat.

PS: We don't want to do an explicit checkout via the subversion module because we don't want to skip the logic used in the Jenkins job.

mpdehaan:
So using Jenkins for a way of rigging up job dependencies.  Interesting.

mpdehaan:
Ok, so this is a cool idea, but I want to contain module sprawl.   Can you perhaps add an addition to the documentation of the uri module to show how to do this with the uri module?  If so, we don't need a module for Jenkins specifically, as it contains much of the logic in common.

muffl0n:
Sound's reasonable! I mistook the description of uri "Interacts with webservices" and didn't think about using it to just call an url.
I will test and report on monday when I'm back at the office.

mpdehaan:
Excellent, thank you!

muffl0n:
Works with force_basic_auth=yes and status_code=201. See #6775

muffl0n:
I'm closing this because the same goal can be achieved with module "uri".

PR #6697, opened 2014-03-26
Made ec2_snapshot more useful

hkariti:
- Added wait + wait_timeout options to wait until the snapshot is complete
- Added tagging on creation
- Return more data: original volume's id and size and the snapshot's tags

### DIFF
--- a/library/cloud/ec2_snapshot
+++ b/library/cloud/ec2_snapshot
@@ -124,6 +124,8 @@ def main():
             ec2_url = dict(),
             ec2_secret_key = dict(aliases=['aws_secret_key', 'secret_key'], no_log=True),
             ec2_access_key = dict(aliases=['aws_access_key', 'access_key']),
+            wait = dict(choices=BOOLEANS, default='true'),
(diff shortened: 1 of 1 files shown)


mpdehaan:
merged, thank you!

Osterjour:
Do I understand correctly: There is an undocumented wait-param now, which changes the (usual) behavior of the API command from AWS and waits until the snapshot is complete?
I understand that some people wanted to get sure the snapshot is completed (which isn't necessary by the way), but please don't change the known behavior and set wait to false on default.

Osterjour:
This fixes the change of the behavior to the way it SHOULD be (and was till this change): https://github.com/ansible/ansible/pull/7438

mariow:
Second this, the default should be not to wait for the completion of the snapshot as this may take a very long time to complete and as long as the process has been started I can be sufficiently sure that  the snapshot will eventually be created.

hkariti:
Yeah, my bad. Sorry about that.

PR #6699, opened 2014-03-26
Fixes minor typos in the GCE guide, one of which causes a code block to ...

clconway:
...fail to render.


mpdehaan:
merged, thanks!  This will appear in the next docs push.

PR #6701, opened 2014-03-26
Make pip dry run work better for the common case

carlanton:
This change allows the pip module to figure out if something needs to be done during
a dry run using 'pip freeze'. This is implemented for the most basic case: name,
version and state=(present|absent). In other cases it will fall back to the old
behavior.


mpdehaan:
Merged, thank you!

PR #6702, opened 2014-03-26
Allow you to pass in arbitrary rsync options

jjshoe:


jjshoe:
Wasn't sure what to put in for the default for an empty array.

jjshoe:
Probably wouldn't hurt to add an example for this in the future. Happy to address both doc issues as another pull request if anyone has input.

jctanner:
@jjshoe please add at least one example in the docstrings.

jctanner:
Sorry to nitpick, but please make the example use complex args format ...

```
foo:
  - bar
  - baz
```

jjshoe:
Per irc, use https://github.com/ansible/ansible/blob/devel/library/cloud/ec2#L354-L365 as an example

jctanner:
@jjshoe while testing this, I realized that the synchronize action plugin doesn't actually handle complex args properly in the case where rsync_opts is a list. I've merged your PR as is, because we can't actually show a complex args example. I'll file a bug to fixup synchronize's ability to use complex args at some point in the future.

jctanner:
https://github.com/ansible/ansible/issues/6951

PR #6703, opened 2014-03-26
rax* modules improvements

sivel:
- Catch issues with invalid regions
- Ensure we send string only data as meta values in the rax module
- Add public_key/lookup example for rax_keypair
- Clean up import statements

### DIFF
--- a/library/cloud/rax
+++ b/library/cloud/rax
@@ -1,4 +1,4 @@
-#!/usr/bin/python -tt
+#!/usr/bin/python
 # This file is part of Ansible
 #
 # Ansible is free software: you can redistribute it and/or modify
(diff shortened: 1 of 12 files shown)


mpdehaan:
Looks like this pull request may need to be rebased.

Also, in the future, I'd recommend moving functions with long argument signatures to use named arguments where possible, as it makes the code easier to trace.   If you can update these, we would appreciate it.

Thanks!

sivel:
Working on this now.

sivel:
I've rebased.  I agree about using named arguments, however I'd like to manage that in a separate pull request.  This will give me more time to test, and a smaller number of changes to test on.

After this pull is merged, I'll begin on that work.

mpdehaan:
merged, thanks!

PR #6706, opened 2014-03-26
Fix template validation docs and example.

geerlingguy:
Please accept this request to update the template module docs and example for validation to match that in the copy module.


mpdehaan:
merged, thanks.  will appear in next docs push.

PR #6707, opened 2014-03-26
rax module: Support arbitrary extra args

sivel:
Add `extra_create_args` and `extra_client_args` to rax module to support passing
advanced configuration options to client instantiation and server create calls.

This allows the module to be used a little more fully with other OpenStack clouds without exposing new individual attributes to users for options not necessarily supported in the Rackspace Public Cloud.

### DIFF
--- a/library/cloud/rax
+++ b/library/cloud/rax
@@ -98,6 +98,17 @@ options:
         state=active/present
     default: no
     version_added: 1.4
+  extra_client_args:
+    description:
+      - A hash of key/value pairs to be used when creating the cloudservers
+        client. This is considered an advanced option, use it wisely and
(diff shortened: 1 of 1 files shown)


nicholaskuechler:
I've just tested this out with the extra args (specifically bypass_url and scheduler_hints) and all tests were successful. Also tested without any extra create or client args and tests were also successful.

LGTM!

ahill00:
Same here. +1

omgjlk:
:+1:  from me, I need this functionality to replace a home-built module. I'd much rather use upstream modules.

mpdehaan:
merged, thank you!  Docs will show up on next docs build.

PR #6709, opened 2014-03-26
Additional documentation for with_subelements

randomvariable:
This improves the ability to work on EC2 related tasks, where you may have run an EC2 task that creates many items with another list of items.

Documentation explaining what happens has been added to the relevant wiki page https://github.com/randomvariable/ansible/blob/with_dict_lists/docsite/rst/playbooks_loops.rst#iterating-over-a-nested-list-as-the-output-of-register .

A better plugin name is welcome.

### DIFF
--- a/docsite/rst/playbooks_loops.rst
+++ b/docsite/rst/playbooks_loops.rst
@@ -133,14 +133,14 @@ Looping over Subelements
 ````````````````````````
 
 Suppose you want to do something like loop over a list of users, creating them, and allowing them to login by a certain set of
-SSH keys. 
+SSH keys.
 
 How might that be accomplished?  Let's assume you had the following defined and loaded in via "vars_files" or maybe a "group_vars/all" file::
 
     ---
     users:
       - name: alice
-        authorized: 
+        authorized:
           - /tmp/alice/onekey.pub
           - /tmp/alice/twokey.pub
       - name: bob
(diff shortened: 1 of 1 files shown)


mpdehaan:
This looks like it seems to be the same as "with_subelements", but I could be wrong.  Could you explain a bit about how this is different?

mpdehaan:
(I'm also wondering if it might be possible to teach with_subelements to cover both cases if not).

randomvariable:
Oh yeah, haha.
The item.0 and item.1 syntax threw me. Maybe the example in this updated commit would help?

mpdehaan:
Think you may mean "{{ item.1.id }}"

Otherwise looks good

randomvariable:
Ooof. Fixed now.

amenonsen:
This PR is still relevant, and I found the added example quite useful. As far as I can tell, the commit no longer applies only because of the extraneous whitespaces changes included originally (removing whitespace at the end of various lines, which were subsequently fixed in other commits).

@randomvariable, any chance you could update the PR to just include the new example?

amenonsen:
I rebased this PR and, in the process, changed my mind. I think the existing example is sufficient, and this PR should be closed. (This adds a lengthy example of using ec2 and iterating over its JSON results, but in terms of _subelement_ usage, it adds nothing new or explanatory.)

bcoca:
closing as per above

PR #6715, opened 2014-03-27
Added list state to file module

willthames:
Useful for testing purposes so that the state of a file can
be interrogated without fear of changing it

Also fixed some minor typos in the documentation

### DIFF
--- a/library/files/file
+++ b/library/files/file
@@ -46,18 +46,19 @@ options:
     aliases: ['dest', 'name'] 
   state:
     description:
-      - If C(directory), all immediate subdirectories will be created if they
+      - If C(directory), all intermediate subdirectories will be created if they
         do not exist. If C(file), the file will NOT be created if it does not
         exist, see the M(copy) or M(template) module if you want that behavior.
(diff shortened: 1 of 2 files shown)


bcoca:
isn' t that what the stat module is for?

willthames:
It's really more a simple information gathering tactic. Given the extra lines of code required for this change, it's almost free. I think stat is a lot more sophisticated, but when you only care that a directory exists, this is a simpler approach. 

I think if more modules had `list` state, we'd be a lot better off - we could use them to gather info for integration testing side-effect free. I've made a start with `ec2_tag` and this could progress with other modules.

I see this as a replacement for the need for my willthames/ansible-testing repo, as we can use the `assert` module to validate infrastructure and applications without the need for additional ansible modules.

mpdehaan:
So re more modules having list state, we've discussed data gathering before and most likely this would be done as a seperate module namespace for those kind of query tasks.    I'd rather not see this continue too much further as this might make this harder or more inconsistent later.

I agree that the stat module already does this and we shouldn't duplicate functionality also.

PR #6717, opened 2014-03-27
If .my.cnf doesn't specify a user, default to user login name

pilou-:
mysql_db should follow the behavior of mysql_user module.

Is there a way to share a function between two module ?

### DIFF
--- a/library/database/mysql_db
+++ b/library/database/mysql_db
@@ -100,6 +100,7 @@ EXAMPLES = '''
 '''
 
 import ConfigParser
+import getpass
 import os
 import pipes
 try:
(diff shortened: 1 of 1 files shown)


mpdehaan:
There's the module_utils system, which we could start a file there for mysql...

muffl0n:
Like described in #6295 the config parsing in the mysql modules should be fixed in general. ConfigParser is not 100% compatible with mysql config files and should therefore not be used.

pilou-:
I close this request: a cleaner solution is proposed in #6295 for `mysql_user` and should be used for `mysql_db` too :-)

PR #6718, opened 2014-03-27
Performance optimization in resolving host patterns

srgvg:
Avoid resolving a pattern that is a plain host. When matching a hostname in the
hosts_cache, just use the host object from there.

When running a task on say 750 hosts, this yields a huge improvement.

### DIFF
--- a/lib/ansible/inventory/__init__.py
+++ b/lib/ansible/inventory/__init__.py
@@ -202,15 +202,18 @@ class Inventory(object):
         hosts = []
 
         for p in patterns:
-            that = self.__get_hosts(p)
-            if p.startswith("!"):
-                hosts = [ h for h in hosts if h not in that ]
-            elif p.startswith("&"):
-                hosts = [ h for h in hosts if h in that ]
+            # avoid resolving a pattern that is a plain host
+            if p in self._hosts_cache:
+                hosts.append(self.get_host(p))
             else:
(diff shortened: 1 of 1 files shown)


mpdehaan:
Does this take globs like "web*.example.com" into account?

srgvg:
Yes. The optimization only shortcuts pattern resolving when (pseudo code) "pattern == inventory_hostname", implemented by retrieving the host object in the host cache when pattern is a key in that cache.

Running with this patch:

```
[trnk] serge@goldorak:~/acd$ ansible msb --list-hosts
    msb-on-1
    msb-on-2
    msb-oe-1
    msb-oe-2
    msb-pr-1
    msb-pr-2
[trnk] serge@goldorak:~/acd$ ansible 'msb-o*' --list-hosts
    msb-on-1
    msb-on-2
    msb-oe-1
    msb-oe-2
[trnk] serge@goldorak:~/acd$ ansible 'msb-o*1' --list-hosts
    msb-on-1
    msb-oe-1
```

srgvg:
This was actually a follow-up to #6734, when looking into thing to optimize a playbook run on my large inventory. I filed it separately, as it stands on its own.

jimi-c:
This has been merged in, but was done as a cherry-pick into another branch, so github's not picking it up correctly. Closing manually, thanks!

PR #6719, opened 2014-03-27
ufw module state value unification

ahtik:
Current ufw module uses values for state as "enabled", "disabled", "reloaded", "reseted". This adds confusion to users who already know ufw and ufw is using "enable", "disable", "reload", "reset". Keeping values the same avoids documentation lookup. Also module doc specified "reseted" while module itself expected "reset". With this PR both of these issues are fixed (no longer "..ed").

Additionally, because logging is a separate ufw command I split up the first line of EXAMPLE usage to have logging as a separate example line. Previously this command returned Invalid Syntax error for ufw v0.33.

As this module seems to be there since 1.6 (unreleased) hopefully this module API change is still alright.

### DIFF
--- a/library/system/ufw
+++ b/library/system/ufw
@@ -27,7 +27,7 @@ short_description: Manage firewall with UFW
 description:
     - Manage firewall with UFW.
 version_added: 1.6
-author: Aleksey Ovcharenko, Jarno Keskikangas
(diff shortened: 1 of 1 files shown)


mpdehaan:
enabled and disabled (with usual aliases for present and absent) are conventions in ansible.

These must remain consistent throughout all modules, so I'm not going to accept this pull request and have this one module be different.

PR #6720, opened 2014-03-27
Fixed ufw module doc reset vs reseted

ahtik:
As a result of closing https://github.com/ansible/ansible/pull/6719 without merge, here is a new PR that only fixes the "reset"/"reseted" issue (current code is internally expecting "reset" but generated doc states "reseted").

I'm not that fluent with American English but "reset" sounds more correct than "reseted" for past tense.

I didn't add my name to module authors tag but if it makes sense to record all contributors there, let me know, will adjust the PR.


mpdehaan:
reset is definitely correct.

thanks!

PR #6721, opened 2014-03-27
EC2 inventory uses AWS *Name* tag if available.

ingmar:
This changes the dynamic EC2 inventory to make use of the AWS _Name_ tag where available. This has the effect that instead of seeing a series of gibberish `ec2-123-123-123-123.compute-15.amazonaws.com` hosts you see the content of the name tag (e.g. `db4`) when running playbooks.

You can also select by these names, e.g. `ansible -m ping db4` (previously you had to do this using `tag_Name_db4`, which of course also still works).

Hosts that do not have a _Name_ tag keep showing up with the ec2 names as before.

### DIFF
--- a/plugins/inventory/ec2.py
+++ b/plugins/inventory/ec2.py
@@ -340,8 +340,14 @@ class Ec2Inventory(object):
             # Skip instances we cannot address (e.g. private VPC subnet)
             return
 
+        # If the instance has a *Name* tag, use that for the display name and put `dest` into ansible_ssh_host
+        ssh_host = None
+        if 'Name' in instance.tags:
+            ssh_host = dest
+            dest = instance.tags['Name']
+
         # Add to index
-        self.index[dest] = [region, instance.id]
+        self.index[dest] = [region, instance.id, ssh_host]
 
         # Inventory: Group by instance ID (always a group of 1)
         self.inventory[instance.id] = [dest]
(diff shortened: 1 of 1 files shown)


mpdehaan:
So with cloud instances you aren't supposed to care about the names of the instances.

i.e. "treat instances like cattle, not like pets"

Because the hostnames given above are often externally useful, changing this may break people who are using {{ inventory_hostname }} in their playbooks.

As a result of the above, we are not going to accept this change.

Note you do have some controls in terms of ec2.ini:

# This is the normal destination variable to use. If you are running Ansible

# from outside EC2, then 'public_dns_name' makes the most sense. If you are

# running Ansible from within EC2, then perhaps you want to use the internal

# address, and should set this to 'private_dns_name'.

destination_variable = public_dns_name

It seems you may be able to set the Name attribute here.

ingmar:
The `destination_variable` setting does not work for this. While it would use the **Name** tag as inventory hostname, Ansible would then no longer be able to connect to that host without the `ansible_ssh_host` that is added by this patch.

Basically, this PR does for dynamic inventories what can be done with a static inventory like so:

```
db4 ansible_ssh_hostname=ec2-123-123-123-123.compute-1.amazonaws.com
```

And most cattle keepers still want to know if a dead animal was kept for the meat or the milk (or less analogous, if a failed host was a http endpoint or a background worker). Currently when applying a playbook to several groups at once, the only way to find this out is by looking up the hostname in EC2, which is what annoyed me into writing this patch. And most cattle herders will still have a few pets too :)

pas256:
This PR requires the value of the tag 'Name' to be unique per host. That is a false assumption.

PR #6722, opened 2014-03-27
EC2 inventory toggle to exclude/include RDS instances.

ingmar:
This simply adds a toggle to choose inclusion of RDS instances in the EC2 dynamic inventory or not. This is because these instances are not SSH manageable and therefore may not be desirable to be included in the **all** hosts selection.

### DIFF
--- a/plugins/inventory/ec2.ini
+++ b/plugins/inventory/ec2.ini
@@ -38,8 +38,8 @@ vpc_destination_variable = ip_address
 # Route53, uncomment and set 'route53' to True.
 route53 = False
 
-# To include RDS instances in inventory, set to True.
-rds = False
+# To exclude RDS instances from the inventory, uncomment and set to False.
+#rds = False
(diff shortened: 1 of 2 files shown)


mpdehaan:
I have no problem with this, though it should default to True to not break any existing users.

If you can amend this to include True in the file and as the default, I'll take this one.

Thanks!

pas256:
This is good. I agreed with @mpdehaan suggestion of default setting.

Why RDS made it into this I will never know: You cannot SSH to an RDS instance.

ingmar:
Any reason not to merge this?

I've changed the default to include RDS instances in https://github.com/ingmar/ansible/commit/cc01f7b7f5e9a02a7e6c7ddfbdff1f37c0c5838e, so it doesn't change existing behaviour.

jimi-c:
Merged, thanks!

Issue #6648, opened 2014-03-24
Issue passing arbitrary data structures to roles


Synforge:
##### Issue Type:

Bug Report
##### Ansible Version:

1.5.2
##### Summary:

There seems to be an issue when passing complex data structures created in host_vars and group_vars to roles.

Please note that this occurs when using both hash_behaviour merge and hash_behaviour replace.
##### Steps To Reproduce:

I've created a minimal project that demonstrates this:

https://github.com/Synforge/ansible-example
##### Expected Results:

I'd expect the provided hash to merge with the defaults included in the role.
##### Actual Results:

``` python
fatal: [mysql-master-1.example.org] => Traceback (most recent call last):
  File "/home/paul/dev/ansible/lib/ansible/runner/__init__.py", line 530, in _executor
    exec_rc = self._executor_internal(host, new_stdin)
  File "/home/paul/dev/ansible/lib/ansible/runner/__init__.py", line 565, in _executor_internal
    inject = utils.combine_vars(inject, self.module_vars)
  File "/home/paul/dev/ansible/lib/ansible/utils/__init__.py", line 1061, in combine_vars
    return merge_hash(a, b)
  File "/home/paul/dev/ansible/lib/ansible/utils/__init__.py", line 545, in merge_hash
    result[k] = merge_hash(a[k], v)
  File "/home/paul/dev/ansible/lib/ansible/utils/__init__.py", line 540, in merge_hash
    for k, v in b.iteritems():
AttributeError: 'unicode' object has no attribute 'iteritems'
```

mpdehaan:
Thanks for the ticket.

Arbitrary datastructures should work, but perhaps not in this format.  We can investigate.

Synforge:
I've submitted a pull request now that fixes this issue:
https://github.com/ansible/ansible/pull/6725

jimi-c:
The above PR has been merged in, so closing this. Thanks!

PR #6725, opened 2014-03-27
Fix for issue #6648

Synforge:
Added template call to _executor_internal to allow the passing of arbitrary data to modules.


jimi-c:
Merged, thanks!

PR #6732, opened 2014-03-27
Micro-optimizations

ciupicri:
A couple of micro-optimizations related to string operations.


mpdehaan:
Not sure these optimize anything of note, but they are more readable, so this is totally fine with me.

Thanks!

PR #6733, opened 2014-03-27
Add meta to the directory layout for roles

jjshoe:


mpdehaan:
thanks!  This will appear when we next push docs.

PR #6734, opened 2014-03-27
vars_plugins refactor and move into core inventory

srgvg:
This patch set does a major set of things, with a big performance benefit in the case of very large inventories (tested on one with 750 hosts)
- Split out parsing of vars files from per host, triggered by runner, to per host and per group parsing, instead of (re)parsing all groups each time for every host. Introduce separate methods to parse a particular group or host and inject this into the proper Inventory objects, instead of parsing all vars for 1 host. This enhances performance.
- Introduce a noop vars plugin, to prepare to move the group_vars plugin into core.
- Move inventory.set_playbook_basedir from ansible-playbook to playbook constructor, to enable it to trigger inventory events.
- Move group/host_vars parsing from vars_plugin code into core inventory. this unifies parsing of all variables from all sources into the Inventory constructor, instead of being called for various sources from various ansible modules  at various times. Also move methods to read files from the plugin, to util (including the vault specific code.)
- This patch until here has the downside to parse  the _whole_ inventory with _all_ its vars at Inventory.init() time, which takes some more time due to the group/host_vars files loading. This was however especially a problem (as in noticeable extra time) in the case of a (very) large inventory with `hash_behaviour=merge`, which is optimized by replacing the `deepcopy` function, by manual parsing in `utils.def merge_hash(a, b):`

This also integrates the vault_password as a parameter of the Inventory, which makes a lot of sense :)


mpdehaan:
"This patch until here has the downside to parse the whole inventory with all its vars at Inventory.init() time, which takes some more time due to the group/host_vars files loading. "

This seems like it might not be a disadvantage, as inventory is parsed once and not per host.   Still reading, but can you elaborate on what the downside might be to this in your view?   Just wanted to understand.   Thanks!

mpdehaan:
Also curious if you can share before/after performance info with (if you have it) links to any test inventory dummy-data you might have?

I recall you had a very large number of groups, maybe exceeding the number of hosts, though I'm not sure how deeply nested it would be, etc.   Just curious to be able to run the same benchmarks on this end.

srgvg:
@mpdehaan I should have stressed this was a (small) disadvantage in _my_ large inventory, as the inventory initialization takes a noticeable extra time here. A time noticed before playbook code kicks in, which is - in my case - always noticed: both when running a play on the full inventory ('all'), as well as in a `--limit`'ed run on say just a couple of hosts, where before this patch, running on just a couple of hosts rarely showed a noticeable initialization time compared to a full run (which took easily several minutes in my case, as opposed to a couple of seconds now.)

People running on smallish inventories will probably not see big differences.
For ther record, my inventory is about 750 hosts with 725 groups, of which about 480 have vars assigned to. host_vars are negleecteable.

I gave @jctanner a json file with my (exported) inventory of hosts and groups (not vars though). Should be easy to generate some vars on each group, or contact me, and I'll look into obfuscating my vars.

Looking at #6734 (this one) and #6718 combined, I got a 50% gain in time to run (10' instead of 20') for a playbook doing some tasks on about 750 hosts.

Be sure to test the use case of hash_behaviour=merge which is my default (particularly of interest with #6718).

mpdehaan:
Yep, I think it's fair to say we optimized (and handled) for large hosts but didn't really handle very large numbers of groups so much.

We'll check this out.

gplasky:
Using the code merged into devel, I tested `ansible-playbook` with a very simple play, operating on ~500 hosts in a single group. With a vault-encrypted file, when `group_vars` is used, playbook execution time is as expected however when used with `vars_files`, execution time skyrockets.

``` yml
---
# ansible-test.yml
- hosts: foo_group
  sudo: no
  gather_facts: no

  vars_files:
    - foo.yml

  tasks:
    - name: Noop
      command: uptime
```

Decrypted file with `vars_files`:

```
$ time ansible-playbook ansible-test.yml 2>&1 | grep real

real    0m23.729s
```

Encrypted file with `vars_files`:

```
$ time ansible-playbook --ask-vault-pass 2>&1 | grep real
Vault password:

real    6m15.755s
```

Remove `vars_files` and `group_vars` within the playbook directory instead:

```
$ time ansible-playbook ansible-test.yml --ask-vault-pass 2>&1 | grep real
Vault password:

real    0m26.614s
```

```
$ ansible-playbook --version
ansible-playbook 1.7 (devel d51e10a3f4) last updated 2014/07/26 13:13:39 (GMT -700)
```

srgvg:
@gplasky so this is an issue with vars_files which is unrelated to this inventory stuff.
You'd better raise a separate issue for this, as closed tickets don't get more attention from ansible devs.

brian-brazil:
Hi,

I've discovered a regression in this change via git bisect, group_vars/all now overrides inventory file variables.

With an inventory of:

```
localhost

[all:vars]
a=bar
```

A site.yml of

```
---
- name: play
  hosts: all
  gather_facts: false
  tasks:
    - name: task
      debug: var=a
```

and a group_vars/all of

```
---
a: foo
```

`make && ansible-playbook -i inv site.yml` now has foo instead of bar.

brian-brazil:
PR #9620 put in to fix this.

PR #6739, opened 2014-03-28
mysql_db: Adding target as a valid param for state=present.

PierrePaul:
Will import the file specified by "target" when using the state=present.

Use case : 
I want to create the basic schema for the database right after creating it. I don't want to run the .sql more than one time and only when creating the database the first time. 

Right now I had to check if 
- the database was already created and use "register" to save the output
- use state=present to create it
- use state=import + target=schema.sql when my_register=False

The overhead was a bit of a pain and changing the module was pretty easy.
Let me know what you think!

Note : the changes at line 298 and 299 are only removing the empty space at the end of the line.

### DIFF
--- a/library/database/mysql_db
+++ b/library/database/mysql_db
@@ -295,8 +295,8 @@ def main():
             else:
                 module.exit_json(changed=True, db=db, msg=stdout)
         elif state == "import":
-            rc, stdout, stderr = db_import(module, login_host, login_user, 
-                                        login_password, db, target, 
+            rc, stdout, stderr = db_import(module, login_host, login_user,
+                                        login_password, db, target,
                                         port=module.params['login_port'],
                                         socket=module.params['login_unix_socket'])
             if rc != 0:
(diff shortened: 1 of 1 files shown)


mpdehaan:
I'm not sure I'm entirely understanding this one.

Can you show your playbook line before and after (using the "target") parameter?

This may help and can also help in testing.

Thanks!

PierrePaul:
While re-doing I thought of something else, so now I can do it in 2 steps instead of 3.

``` yml
---
- hosts: all
  user: vagrant
  sudo: true
  pre_tasks:
    - name: install the required python-mysqldb for the ansible module mysql_db
      apt: name=python-mysqldb state=present

  tasks:
    - name: mysql check if db is created
      mysql_db: name=testdatabase state=present
      register: result

    - name: import the schema if we just created the database
      mysql_db: name=testdatabase state=import target=/tmp/test.sql
      when: result|changed
```

And the tasks becomes :

``` yaml
    - name: mysql create the database and import the schema
      mysql_db: name=testdatabase state=present target=/tmp/test.sql 
```

Can't wait to see your talk at PyCon. Thanks :)

Jmainguy:
Hey PierrePaul: Can you resolve the merge conflicts, I am sure this will speed up the merge. The code should only import the database if the database does not previously exist. Is that what this is doing? Or will it import the database everytime. Would it be easier to read if we passed target -> db_create and it imports the sql if db_create is called and target is defined? Just some thoughts, what do you think PierrePaul?

mpdehaan:
Hi!

Thank you very much!

PR #6740, opened 2014-03-28
Add Amazon SNS notification plugin

mjschultz:
- Supports a default message or a special message for any of SNS's
  protocols (email, sqs, sms, http, https)
  - http://docs.aws.amazon.com/sns/latest/dg/PublishTopic.html

Uses [boto's SNS API](http://boto.readthedocs.org/en/latest/ref/sns.html). For reference, [Amazon's API](http://docs.aws.amazon.com/sns/latest/api/API_Publish.html)

### DIFF
--- a/library/notification/sns
+++ b/library/notification/sns
@@ -1,0 +1,185 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2014, Michael J. Schultz <mjschultz@gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


mpdehaan:
Thanks for the submission!

mpdehaan:
Looks like there maybe should be some error handling around this part to call "fail_json" if the connection fails, so it doesn't traceback:
-    connection.publish(topic=arn_topic, subject=subject,
- ```
                    message_structure='json', message=json_msg)
  ```

Can you update this one? 

Otherwise looks good to me.  Thanks again!

mjschultz:
No problem, I added a BotoServerError handling that just spits out the exception as a string (this seems to be what other aws modules do for the connection).

I also added a failure if the `region` param isn't supplied by the call to `get_aws_connection_info` since if `region=None` I ended up with an AttributeError on the `connection.publish()`. Let me know if there is a better/more appropriate way.

mpdehaan:
Merged, thank you!

mjschultz:
No problem, thanks for ansible!

PR #6747, opened 2014-03-28
Bugfix for gc_storage and s3

ciupicri:
Make keysum return None if not key_check (this case wasn't covered).


mpdehaan:
merged, thanks!

PR #6752, opened 2014-03-29
New attribute "submodules" to select which sub-module to update.

eroldan:
Replaces "recursive" functionality.
Changes error handling when doing submodule operations

### DIFF
--- a/library/source_control/git
+++ b/library/source_control/git
@@ -528,7 +528,7 @@ def main():
                 module.exit_json(changed=False, before=before, after=remote_head)
         if module.check_mode:
             module.exit_json(changed=True, before=before, after=remote_head)
-        fetch(git_path, module, repo, dest, version, remote, bare)
+        fetch(git_path, module, repo, dest, version, remote, bare, submodules)
 
     # switch to version specified regardless of whether
     # we cloned or pulled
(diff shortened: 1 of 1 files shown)


mpdehaan:
This behavior should probably be tweaked because we don't like giving meaningful behavior to empty strings or contrasting None with "", which both have a "truthiness" in Python as False.

```
+            - Which sub-modules will be updated. Empty string skips all.
+              If this argument is not supplied all sub-modules will be
+              updated. 
```

Can we change this to "none" skips all submodules?

(Be sure to do a case insensitive match also)

Thanks!

mpdehaan:
(Note that will have to be applied in 1.6 timeframe as the recursive module is new in 1.6, and otherwise can't be removed)

eroldan:
Your recommendations were done. Thanks!

leth:
FYI: the recursive flag is currently broken -- setting to to 'no' does not prevent all sending of "--recursive" flags to git

eroldan:
Bummer. Ansible 1.6 is released making the 'recursive' parameter written in stone :( 
Maybe I can adapt the "submodule" parameter to coexist with recursive, but I'm not very interested right now.
As others stated, the "recursive" flag is broken in 1.6 release, this fix seems ok https://github.com/ansible/ansible/pull/7033/commits

jimi-c:
@eroldan based on the above, I'm going to go ahead and close this. If you do have further ideas to improve the functionality of the git module please let us know. Thanks!

PR #6753, opened 2014-03-29
fixed/extended module package/zypper

berendt:
- Introduced new parameter recommends to be able to install
  and updated packages without installing recommend packages.
- Made disable_gpg_check usable with updating packages.
- Changed SuSE to SUSE in the description, because that's the
  correct name of the company. Have a look at
  http://en.wikipedia.org/wiki/SUSE for verification.

### DIFF
--- a/library/packaging/zypper
+++ b/library/packaging/zypper
@@ -33,9 +33,9 @@ DOCUMENTATION = '''
 module: zypper
 author: Patrick Callahan
 version_added: "1.2"
-short_description: Manage packages on SuSE and openSuSE
+short_description: Manage packages on SUSE and openSUSE
 description:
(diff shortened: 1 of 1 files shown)


mpdehaan:
So with this patch the "version_added" attribute needs to be added in the documentation for the new parameter.

Also, is the default for recommends the way it worked before this change?

Thanks!

berendt:
@mpdehaan Recommends works the way it worked before. Recommends is used by default by zypper and the default in the Ansible module is yes.

berendt:
Any news here?

alxgu:
The feature to disable "install recommends" is already in 1.8 devel https://github.com/ansible/ansible/pull/7379
Did not see your pr :-(

berendt:
Ok. Then this pull request can be abandoned.

PR #6754, opened 2014-03-29
Bugfix to ufw module to properly support "logging" (and docfix for choises, choices):

ahtik:
"logging" is a separate parameter so updated docs and set it as mutually exclusive param.

Also due to an array construction typo logging command was not working in any situation (ufw LOGLEVEL was passed to cmd instead of ufw logging LOGLEVEL).

Also fixed doc and parameters parsing typo ("choises" should be "choices")

### DIFF
--- a/library/system/ufw
+++ b/library/system/ufw
@@ -1,6 +1,7 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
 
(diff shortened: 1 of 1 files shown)


mpdehaan:
merged, thank you!

PR #6758, opened 2014-03-29
backreferences should be double escaped

cgtx:
The **DOCUMENTATION** string mentions that the line parameter should have backreferences double escaped, but the **EXAMPLES** string contains an example where it is not.


mpdehaan:
merged, thank you -- will appear when docs are next pushed

PR #6759, opened 2014-03-30
More descriptive error message if init system is unknown

Malvineous:
I just spent a long time debugging a problem where I thought Ansible could not find an initscript to start a service, but it turned out the problem was that Ansible did not know about the init system on my target machine.  I was mislead by an inaccurate error message, so I would like to submit this trivial patch to change the error message to be more helpful.

### DIFF
--- a/library/system/service
+++ b/library/system/service
@@ -565,7 +565,7 @@ class LinuxService(Service):
     def service_enable(self):
 
         if self.enable_cmd is None:
-            self.module.fail_json(msg='service name not recognized')
+            self.module.fail_json(msg='unknown init system, cannot enable service')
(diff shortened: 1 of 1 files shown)


jimi-c:
I would think some combination of the two would be better, since it could be that the init script was not found OR the init system itself wasn't found?

Malvineous:
I've just tested this to confirm, but if the initscript is not found that error is picked up earlier in the sequence, leading to this message:

```
msg: cannot find 'service' binary or init script for service,  possible typo in service name?, aborting
```

jimi-c:
Merged, thanks!

Issue #6765, opened 2014-03-30
Bug report: mysql_db does not fail when using import and bz2 or gz


Jmainguy:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.6

Bug was introduced https://github.com/ansible/ansible/pull/4307
##### Environment:

N/A applies to all 
##### Summary:

When using state=import, and the target= ends with .bz2 or .gz, it will succeed even when the bunzip2 or gunzip command fails. If the target does not exist, it succeeds. If the target exists but is not actually a zipped up file, it still succeeds. The module should fail if the bunzip2 or gunzip commands fail.
##### Steps To Reproduce:

ansible -i hosts realhostname -m mysql_db -a "name=test target=/backup/test.sql.gz state=import"

and the target does not exist, or is not really zipped up.
##### Expected Results:

I expect the module to return back fail with the stderr of the bunzip2 or gunzip command. 
##### Actual Results:

It returns ok as in the entire thing succeeded (when indeed it did not)

Jmainguy:
See bug fix above. It does not change dump it all, only import. I can alter the code to use the same kind of logic for dump if that is deemed beneficial.

mpdehaan:
Thanks for this, given there is the other pull request ticket talking about this one, I'm going to close this ticket to keep the database organized.

I commented a bit about a slight tweak to the above code in your pull request but otherwise this looks pretty good to me.

Thank you!

PR #6766, opened 2014-03-30
mysql_db module now correctly fails on import if bz2 or gz fails.

Jmainguy:
This is a bug fix for the bug report

https://github.com/ansible/ansible/issues/6765


mpdehaan:
This part looks a little dangerous:
-            cmd += " < %s" % pipes.quote(target[0:-3])

As it assumes the ending of the filename is significant (which it doesn't have to be).

Can this be updated to use os.path.splitext(path) instead?

Thanks!

Jmainguy:
Wise suggestion. Added.

jimi-c:
Merged, thanks!

PR #6774, opened 2014-03-31
ssh: Properly check for wrong su password

tyll:
There is an obvious typo.


mpdehaan:
While this is true, your correction is not fixing it:

Take a look at this line closely:

```
 +            if self.runner.su and su and self.runner.su:
```

tyll:
On Mon, Mar 31, 2014 at 06:42:57AM -0700, Michael DeHaan wrote:

> While this is true, your correction is not fixing it:
> 
> Take a look at this line closely:
> 
> ```
>  +            if self.runner.su and su and self.runner.su:
> ```

Hooray for patch reviews. I changed this.

jimi-c:
Merged, thanks!

PR #6775, opened 2014-03-31
Module "uri": Example for queueing build in Jenkins

muffl0n:


mpdehaan:
I'll probably amend this to show this used in a local play, as in most cases this wouldn't make sense inside the host loop, but this is a nice example to include.

Thanks!

mpdehaan:
merged w/ some tweaks, thanks.  This will appear next time we push docs.

PR #5324, opened 2013-12-17
setup module: Additional facts when using ansible with Linux on System z

GuiGui2:
- Improved parsing of /proc/cpuinfo to take the specifics of Linux on System z. ansible_processor and ansible_processor_cores are now completed.
- Also improved virtualization facts in the same environment, parsing the output of /usr/bin/lscpu to populate virtualization_type and virtualization_role more precisely. If lscpu is not available, then stick to the previous behavior.

PR #6777, opened 2014-03-31
Setup module: Additional facts when using ansible with Linux on System z

GuiGui2:
- Improved parsing of /proc/cpuinfo to take the specifics of Linux on
  System z. ansible_processor and ansible_processor_cores are now
  completed.
- Also improved virtualization facts in the same environment, parsing
  the output of /usr/bin/lscpu to populate virtualization_type and
  virtualization_role more precisely. If lscpu is not available, then
  stick to the previous behavior.

This is a follow-up of pull-request 5324 (https://github.com/ansible/ansible/pull/5324)


mpdehaan:
Merged, thank you!

PR #6785, opened 2014-04-01
Added region name to client connection as it is needed for some cloud pr...

CaptTofu:
I could not get this to work with HP Cloud because there are two endpoints and if you don't specify the region, you will get this error:

"Found more than one valid endpoint. Use a more restrictive filter"

This would happen in any setup with multiple regions.

### DIFF
--- a/library/cloud/nova_compute
+++ b/library/cloud/nova_compute
@@ -249,6 +249,7 @@ def main():
                               module.params['login_password'],
                               module.params['login_tenant_name'],
                               module.params['auth_url'],
+                              region_name=module.params['region_name'],
                               service_type='compute')
     try:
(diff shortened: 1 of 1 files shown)


CaptTofu:
Hi all! Just checking in on this and wondering if anyone had questions, thoughts.

mpdehaan:
Can you please update this to remove the merge commit?  We don't allow those.

(tip:  always use git pull --rebase vs git pull, and git rebase vs git merge, and that will fix that problem)

Thanks!

CaptTofu:
Thanks! Sorry about that. I just did what you said and pushed that up to my fork.

jimi-c:
Merged, thanks!

PR #6788, opened 2014-04-01
ssh connection plugin: Make comments more verbose

tyll:
Explain _communicate() more.


mpdehaan:
Note when testing -- this pull request says "explain communicate more" but also adds a ".close()"--  so it's not just docs, despite the description.

tyll:
It only changes the comment to be before stdin.close() instead of being in the same line

mpdehaan:
Oh ok, good deal, thanks for pointing out.

mpdehaan:
merged, thank you!

PR #6791, opened 2014-04-01
Fix cron module and memtotal_mb fact on HP-UX

tgolly:
On HP-UX, the crontab command works like this, because there is no "-u" it's picky about the order of the arguments:

```
 crontab(1)                                                       crontab(1)
 NAME
      crontab - user job file scheduler
 SYNOPSIS
      crontab [file]
      crontab -e [username]
      crontab -l [username]
      crontab -r [username]
```

The first patch ensures "crontab -l [username]" for reading, and "crontab [file]" for writing (same as SunOS).

The second patch attempts to address Michael's recommendations in pull request #6018 .

This was something we have been manually fixing each release of Ansible (we rotate syslog.log on HP-UX).  This will ensure setup gets the **memtotal_mb** fact when running witih --sudo.


jimi-c:
Merged, thanks!

PR #6794, opened 2014-04-01
Make `vpc_security_group` docs more accurate

follower:
Without the `subnet` parameter supplied there's an error `msg: Parameter vpc_security_groups invalid for create command`. (This might be a bug?)

If the VPC security group name rather than ID is supplied there's an error: `msg: Invalid security group , groupId= <some group name>, groupName=.` (Accepting a group name might be a feature enhancement.)

In my case I set the subnet as `default` and used `register` to get the result of the security group creation section and just referred to its `group_id` property.

### DIFF
--- a/library/cloud/rds
+++ b/library/cloud/rds
@@ -125,7 +125,7 @@ options:
     aliases: []
   vpc_security_groups:
     description:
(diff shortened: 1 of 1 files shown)


mpdehaan:
Merged, thanks!

PR #6795, opened 2014-04-01
module_retention isn't defined anywhere, including the help.

jjshoe:


follower:
It appears that currently a non-default retention period is silently dropped due to the use of "module_retention" which seems to be a typo.

mpdehaan:
merged, thanks!

PR #6796, opened 2014-04-01
Remove extra quote from the mysqldump password argument

mistaka0s:
For complex passwords, the mysqldb Ansible module will fail for database imports or dumps with a '1045: Access Denied' mysql error even though the credentials work properly and the mysqldump command can be run manually.

This is caused by the extra quote around the '--password' argument to mysqldump, as pipes.quotes already quotes the password string.

> > > "--password='%s'" % pipes.quote('simple')
> > > "--password='simple'"
> > > 
> > > "--password='%s'" % pipes.quote('c0mplexp@ssword!')
> > > "--password=''c0mplexp@ssword!''"
> > > 
> > > "--password='%s'" % pipes.quote('password with space')
> > > "--password=''password with space''"

### DIFF
--- a/library/database/mysql_db
+++ b/library/database/mysql_db
@@ -124,7 +124,7 @@ def db_delete(cursor, db):
 
 def db_dump(module, host, user, password, db_name, target, port, socket=None):
     cmd = module.get_bin_path('mysqldump', True)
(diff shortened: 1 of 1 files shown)


mpdehaan:
merged, thank you!

PR #6802, opened 2014-04-01
The open_iscsi module fails if no active sessions.

marker:
The open_iscsi module does a test for "logged on" status prior to logging in a configured target.  The target_loggedon test will always fail if there is not at least one existing iSCSI connection active, because in the absence of any sessions, iscsiadm returns error 21 and the module fails.  On newly-built machines, it is therefore impossible to connect to iSCSI targets because there aren't any sessions (a catch-22).

### DIFF
--- a/library/system/open_iscsi
+++ b/library/system/open_iscsi
@@ -162,6 +162,8 @@ def target_loggedon(module, target):
     
     if rc == 0:
         return target in out
+    elif rc == 21 and err.find("No active sessions") != -1:
+        return False
     else:
         module.fail_json(cmd=cmd, rc=rc, msg=err)
 


srgvg:
Wondering why I never encountered this issue, using this almost daily. 

I recall I had to implement something similar (see https://github.com/marker/ansible/blob/patch-1/library/system/open_iscsi#L141) which IIRC was about a recent version of iscsiadm on Fedora (I use the version om Ubuntu 12.04 atm, can't check which exact version that is.

Must be something similar for a use case I never encountered.

marker:
I suspect you're correct and it's just a difference between versions.  I should have included in my notes that I'm specifically dealing with RHEL/CentOS 6.5 hosts, with iscsi-initiator-utils-6.2.0.873-10.el6.x86_64.

srgvg:
Hm, Ubuntu 12.04 appears to have these:

Version: 2.0.873-3ubuntu5~ubuntu12.04.1
Version: 2.0.871-0ubuntu9.12.04.2
Version: 2.0.871-0ubuntu9

I'll check tomorrow the version reported by iscsiadm on the deployed hosts.

marker:
To follow up, I've had to rebuild a host today and got a little time to experiment on this with a freshly installed and updated CentOS 6.5 host.

On the target host:
    # rpm -qa 'iscsi*'
    iscsi-initiator-utils-6.2.0.873-10.el6.x86_64
    # 
    # iscsiadm -m node
    10.10.10.5:3260,1 iqn.2001-05.com.equallogic:0-8a0906-10ae99a0a-819d2f34f5853064-mysql-data
    10.10.10.5:3260,1 iqn.2001-05.com.equallogic:0-8a0906-10ae99a0a-819d2f34f5853064-mysql-data
    # 
    # iscsiadm -m session
    iscsiadm: No active sessions.
    # echo $?
    21

Trying an ansible run:
$ ansible dbtest -u root -m open_iscsi -a "target=iqn.2001-05.com.equallogic:0-8a0906-10ae99a0a-819d2f34f5853064-mysql-data login=yes"
dbtest | FAILED >> {
    "cmd": "/sbin/iscsiadm --mode session", 
    "failed": true, 
    "msg": "iscsiadm: No active sessions.\n", 
    "rc": 21
}

After patching open_iscsi:
$ ansible dbtest -u root -m open_iscsi -a "target=iqn.2001-05.com.equallogic:0-8a0906-10ae99a0a-819d2f34f5853064-mysql-data login=yes"
dbtest | success >> {
    "changed": true, 
    "connection_changed": true, 
    "devicenodes": [
        "/dev/sdd"
    ]
}

Unfortunately I don't have any Debian or Ubuntu hosts handy to compare post-patch results at this time.

srgvg:
@marker can you confirm the version of iscsiadm?
I have on Ubuntu 12.04:

```
 iscsiadm --version
 iscsiadm version 2.0-871
 iscsiadm -m session
 iscsiadm: No active sessions.
 echo $?
 0
```

So you're patch seems correct to me.

marker:
[root@dbtest ~]# iscsiadm --version
iscsiadm version 6.2.0-873.10.el6
[root@dbtest ~]# iscsiadm -m session
iscsiadm: No active sessions.
[root@dbtest ~]# echo $?
21

The open-iscsi.org page shows their current version as 2.0-873 so it looks like RH/CentOS are on the current stable version.  I'm curious as to whether this difference in return code comes from changes in upstream open-iscsi between 2.0-871 and 2.0-873, or if they're distro-specific patches.  Unfortunately I don't have anything older than CentOS 6.3 so I can't compare how their older versions behave.

EDIT:  Grabbed the source and poked around, they return 21 for no active sessions in 2.0-873:
    include/iscsi_err.h:    ISCSI_ERR_NO_OBJS_FOUND     = 21,

In 2.0-871 there's no such item, which probably accounts for the difference in behavior we're seeing.  When Debian/Ubuntu hit 2.0-873 you'll probably see the same thing.

jimi-c:
It looks like 592a5b6fb1f357bf388d80a231e94c0b803e38f9 was already merged in to address this, though it doesn't check the error message (which I do not think is important). So I'll go ahead and close this, but if it turns out that other things may trigger a rc of 21 then feel free to resubmit that as an additional patch. Thanks!

PR #6803, opened 2014-04-01
redhat_subscription calls AnsibleModule() without argument_spec

aresch:
The Rhsm object requires an AnsibleModule but it isn't created with an argument_spec and fails.  Since the rhn.module is set directly after, setting None for the required argument of Rhsm fixes the module.

### DIFF
--- a/library/packaging/redhat_subscription
+++ b/library/packaging/redhat_subscription
@@ -327,7 +327,7 @@ class RhsmPools(object):
 def main():
 
     # Load RHSM configuration from file
-    rhn = Rhsm(AnsibleModule())
+    rhn = Rhsm(None)
 
     module = AnsibleModule(
                 argument_spec = dict(


jctanner:
@aresch what was the failure message/traceback?

aresch:
@jctanner 

```
  File "/home/ansible/.ansible/tmp/ansible-tmp-1396375983.71-124437843062785/redhat_subscription", line 1564, in <module>
    main()
  File "/home/ansible/.ansible/tmp/ansible-tmp-1396375983.71-124437843062785/redhat_subscription", line 330, in main
    rhn = Rhsm(AnsibleModule())
TypeError: __init__() takes at least 2 arguments (1 given)


```

mpdehaan:
What version of ansible was this?

mpdehaan:
yeah this is obviously the way it should be from the constructor, likely damaged in the 1.5.X fixes in the past and didn't have enough coverage.

Merging, thanks!

PR #6805, opened 2014-04-01
Fix interpreter lines in a couple of modules

sivel:
The `riak` and `docker_image` modules, both use `#!/usr/bin/env python` instead of `#!/usr/bin/python`

This pull request updates them to the proper values so that `ansible_python_interpreter` will work as expected


jctanner:
@sivel thanks!

PR #6808, opened 2014-04-01
debconf may always reassert passwords

gundalow:
On some versions of Ubuntu debconf-get-selections hides the passwords, even when run as root. This means the debconf module believes the password isn't set correctly.

Make this clear in the documentation

@bcoca FYI as the original author

### DIFF
--- a/library/system/debconf
+++ b/library/system/debconf
@@ -34,6 +34,8 @@ notes:
     - A number of questions have to be answered (depending on the package).
       Use 'debconf-show <package>' on any Debian or derivative with the package
       installed to see questions/settings available.
+    - One some distros passwords will also be reasserted. This is due to
+      debconf-get-selections masking passwords.
 options:
   name:
     description:
--- a/library/system/debconf
+++ b/library/system/debconf
@@ -34,8 +34,7 @@ notes:
     - A number of questions have to be answered (depending on the package).
       Use 'debconf-show <package>' on any Debian or derivative with the package
       installed to see questions/settings available.
-    - One some distros passwords will also be reasserted. This is due to
-      debconf-get-selections masking passwords.
+    - Some distros will always record tasks involving the setting of passwords as changed. This is due to debconf-get-selections masking passwords.
 options:
   name:
     description:


mpdehaan:
I'm not sure if the end user will know what "reasserted" means.   Perhaps this should explain a bit more (maybe when to not use things?) and switch up the terminology?

gundalow:
@mpdehaan by "reassert" I mean the value will always change. I'm not sure what the Ansible term for that is.

If you run most playbooks a second time you will just get "ok", this isn't the case with the password vtype, it will be listed as "changed".

kiranos:
I second this, this would be a problem as often when automating mysql-server installs, a good way is to set the mysql root password with debconf and then run the installer in noninteractive mode.

gundalow:
@mpdehaan Do you have a suggestion for the term for when Ansible will also trigger a "changed" event, rather than an "OK" (unchanged)?

mpdehaan:
I'd probably say "will always record the task as having changed" or something similar ...

kiranos:
if https://github.com/ansible/ansible/issues/6925 (ability to set when: based upon if package is installed) a solution could be to set this password when mysql-server is not installed and remove the debconf settings once the package is installed, default is:

mysql-server-5.5        mysql-server/root_password_again        password
mysql-server-5.5        mysql-server/root_password      password

which would not be "always changed"

gundalow:
Wording updated

xezpeleta:
Hi!

I had the same problem. I've solved it with:

```
- name: check if foo is already installed
  shell: dpkg -l foo > /dev/null 2> /dev/null
  sudo: yes
  changed_when: false
  register: foo_installed
  ignore_errors: true

- name: preconfigure foo
  debconf: name=foo question='foo/password' value{{ foo_password }} vtype=password
  sudo: yes
  when: foo_installed|failed
```

Thanks

gundalow:
This can be closed
https://github.com/ansible/ansible-modules-extras/pull/178 is the fix for ansible-modules-core

PR #6824, opened 2014-04-02
Allow 'make rpm' work from a release tarball

sivel:
`make rpm` cannot be run from a release tarball because the release tarball is missing the `VERSION` and `MANIFEST.in` files.

I am unsure if this is intended, but this is a somewhat simple pull request to have `MANIFEST.in` include `VERSION` and itself, so that `make rpm` will work.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -6,3 +6,5 @@ include packaging/distutils/setup.py
 recursive-include docs *
 recursive-include library *
 include Makefile
+include VERSION
+include MANIFEST.in


mpdehaan:
This may be worthwhile consulting with Nirik in IRC to see how he does this, but we in fact run from source checkouts and do "make rpm", which leverages the results of "make sdist" too.

So I'm a bit confused.

As the RPM should not probably not include MANIFEST.in and VERSION on the filesystem, I'm a little unclear about this one.

sivel:
Yeah, it is fine when you build from a git clone.  But the Makefile requires VERSION and MANIFEST.in in order to build the RPMs.  So if they are not included in the release tarball, you can't build RPMs from the release.

I'm fine either way, and I can ping Nirik.  I'm not too concerned about the functionality but was helping someone else in IRC earlier who was trying to do this.

sivel:
I believe from our conversation today that this should be safe.  I've verified that neither MANIFEST.in nor VERSION end up in the RPM.

jimi-c:
Merged, thanks!

PR #6829, opened 2014-04-02
Update cron module documentation. The 'name' parameter is required.

jlund:


mpdehaan:
merged, thanks!  This will appear when we next push docs.

dirtyharrycallahan:
I refactored this module last year and have not taken a close look at the work since then but I wonder is the claim made is true?

```
# Removes a cron file from under /etc/cron.d
- cron: cron_file=ansible_yum-autoupdate state=absent
```

PR #6833, opened 2014-04-03
Add '--list-groups' flag and support to ansible & ansible_playbook

angstwad:
We implemented this in our own scripts, turned out to be really useful for _our_ dynamic inventory scripts -- not so useful for really flat orgs, with maybe one/two groups, but, you know, it does what it says it does.  It lists groups.  :)

### DIFF
--- a/bin/ansible
+++ b/bin/ansible
@@ -61,6 +61,8 @@ class Cli(object):
         parser.add_option('-m', '--module-name', dest='module_name',
             help="module name to execute (default=%s)" % C.DEFAULT_MODULE_NAME,
             default=C.DEFAULT_MODULE_NAME)
+        parser.add_option('--list-groups', dest='listgroups', action='store_true',
+            help='outputs a list of groups; does not execute anything else')
 
         options, args = parser.parse_args()
         self.callbacks.options = options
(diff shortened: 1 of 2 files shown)


mpdehaan:
List groups that the playbook would run against?  Or all groups?

This would make it perhaps inconsistent with --list-hosts, which is why I'm asking.

mpdehaan:
This might be fixable by changing the name of the option.   Or it might warrant a seperate CLI.

Eventually I wouldn't mind having a unified CLI for ad-hoc and main programs, which could have an "inventory" subcommand, IDK.

angstwad:
It lists all groups returned by the inventory.

angstwad:
What name were you thinking for the option?

mpdehaan:
I'm thinking --list-all-groups would probably avoid most user confusion, along with a comment that details the same.

srgvg:
I'd vote for a separate utility, that can query all things inventory.
`bin/ansible-inventory`
Could perhaps also integrate the graphviz dot generation stuff?

angstwad:
@mpdehaan And it is done.

```
14:26 $ ansible -i hosts --list-groups -m ping all
    all
    database
    etcd
    local
    proxy
    staging
    test
    ungrouped
    webapp

14:26 $ ansible-playbook -i hosts --list-all-groups test-groups.yml
    all
    database
    etcd
    local
    proxy
    staging
    test
    ungrouped
    webapp
```

mpdehaan:
I think we should go with Serge's idea here, seeing it doesn't bear on the idea of not having relevancy to the playbook.

It could also have an indented tree option maybe.

Additionally I think the --list-all-groups vs --list-groups could be somewhat confusing being different between applications.

The proposed utility may need to take a --limit.

Anyway, I greatly appreciate all the time invested in this one, and this is perhaps a good discussion for ansible-devel

Sorry for the delay, our queue attack algorithms were prioritized towards bugs and cloud things for a while.  I hope you understand and thanks again!

PR #6835, opened 2014-04-03
Add version_compare filter

sivel:
Based on the filter proposed in #6584, this pull request adds only the `version_compare` filter.

Changed in this pull request, is that by default it will use LooseVersion, with an argument (`strict`) to `version_control` to allow use of StrictVersion instead.

LooseVersion was chosen as the default, since I think people will attempt to do something like:

```
{{ ansible_distribution_version | version_compare(12, '>') }}
```

With StrictVersion `12` would be an invalid version number.

### DIFF
--- a/docsite/rst/playbooks_variables.rst
+++ b/docsite/rst/playbooks_variables.rst
@@ -208,6 +208,32 @@ To get the symmetric difference of 2 lists (items exclusive to each list)::
 
     {{ list1 | symmetric_difference(list2) }}
 
+.. _version_comparison_filters:
+
+Version Comparison Filters
+--------------------------
+
+.. versionadded:: 1.6
+
+To compare a version number, such as checking if the ``ansible_distribution_version``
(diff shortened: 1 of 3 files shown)


mpdehaan:
So we also have SVG's #6619 which I'm hoping he'll revise to not include the dynamic git lookup (which is a little inefficient in a runner call) which would allow the same.   So this is kind of a duplicate for now so you guys need to sort this out.

If we can revise #6619 that one should be able to go, or do you find differences in the implementation where you would prefer this one?

Thanks!

sivel:
#6619 only introduces ansible_version. As far as I can tell there is no version comparison filter added in his pull request.

I've removed the ansible_version functionality from this pull.  It is strictly related to adding a new filter to perform version comparisons.

srgvg:
I'll rework #6619, just needs some time :)

mpdehaan:
I think I was confused by another PR then.  Thanks, looking at this now.

mpdehaan:
Merged, thank you!

PR #6836, opened 2014-04-03
adding ternary operator to filters

odo:
This would allow things such as: `success(result) | ternary("it worked", "it didn't")`.


mpdehaan:
While I appreciate the brevity of this option, IMHO, this shouldn't be too neccessary as Jinja2 can already do if/else.

We generally want to avoid duplicate approaches, and while I know what a ternary operator is, many of our users come from admin backgrounds and would not, which may decrease readability of the playbook.

It's still fine of course if you continue to use this by dropping this in the filter_plugins path.

Jinja2 looks like:

```
{% if foo %}
    foo
{% else %}
    bar
{% endif %}
```

odo:
Fair enough.

PR #6837, opened 2014-04-03
Fix several changed tests in ec2 auto scaling related modules

sayap:
Currently, the ec2_asg and ec2_metric_alarm modules do not handle `"changed": true/false` correctly in certain situations:
- ec2_asg: the availability zones returned by AWS/boto can have different order than the availability zones passed in, e.g. `['ap-southeast-1b', 'ap-southeast-1a']` versus `['ap-southeast-1a', 'ap-southeast-1b']`
- ec2_metric_alarm: if not set, the description returned by AWS/boto would somehow be empty string instead of null
- ec2_metric_alarm: if the existing dimensions and the new dimensions have different keys, ansible can crash
- ec2_asg: if desired_capacity is not passed in, we shouldn't set changed to true

### DIFF
--- a/library/cloud/ec2_asg
+++ b/library/cloud/ec2_asg
@@ -157,11 +157,18 @@ def create_autoscaling_group(connection, module):
     else:
         as_group = as_groups[0]
         changed = False
-        for attr in ('launch_config_name', 'max_size', 'min_size', 'desired_capacity',
+        for attr in ('launch_config_name', 'max_size', 'min_size',
                      'vpc_zone_identifier'):
-            if getattr(as_group, attr) != module.params.get(attr):
+            if getattr(as_group, attr) != module.params[attr]:
                 changed = True
                 setattr(as_group, attr, module.params.get(attr))
+        # handle desired_capacity separately, so that we can leave the option
+        # unset, and thus do not disrupt the auto scaling group
+        desired_capacity = module.params.get('desired_capacity')
+        if desired_capacity is not None:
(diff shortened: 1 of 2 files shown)


mpdehaan:
This pull request does not supply a description.

Can you please detail the various problems that were corrected by these additions so that folks can use these to help validate the code changes and logic?

Thank you!

sayap:
I have added some description, but struggled a bit with the wordings...

jimi-c:
This PR now has several conflicts due to another PR that was merged in for ec2_asg. Could you please rework this based on the current state of the module? Thanks!

sayap:
@jimi-c While trying to rebase my changes, I notice several issues with 3e875c786729:

> allows setting of tags, changes upon which mark the task changed

This doesn't allow tag(s) removal, i.e. if the tags param is a subset of the existing tags, nothing happens.

It also doesn't convert the tag value to string first before doing the comparison, so `"changed": true` can be wrongly returned.

> doesn't mark changed if the parameter wasn't specified

This prevents setting min_size, max_size, and desired_capacity to 0.

> Availability Zones are pulled from the region

This is only done during asg creation, but not during modification. So, if a ec2_asg `state=present` task doesn't have availability_zones and vpc_zone_identifier set, it will always return `"changed": true`

I will see if I can put up a PR that also includes fixes for the issues.

jimi-c:
@sayap sounds good, thanks!

PR #6838, opened 2014-04-03
ec2_scaling_policy: Always return ARN when creating/updating a policy.

sayap:
This includes scenario where the policy remains unchanged, as the ARN could be
needed by subsequent tasks (e.g. putting metric alarm).

The other return values can be inferred by the caller, so I have removed them.

Also change to call `fail_json` when there is an error with policy delete.


jimi-c:
Similar to your other PR, the underlying module has unfortunately changed, so your PR will need to be rebased against it. Please resubmit this when it's ready to go. Thanks!

PR #6839, opened 2014-04-03
Enhancements to EC2 functionality

ghost:
Mostly minor additions extracted from work we've done in our playbooks.

### DIFF
--- a/library/cloud/ec2
+++ b/library/cloud/ec2
@@ -217,6 +217,11 @@ options:
     required: false
     default: null
     aliases: []
+  ebs_optimized:
+    description:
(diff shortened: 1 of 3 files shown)


mpdehaan:
Thanks very  much for the submission!

Can you please explain the additions in the comment of the ticket?

Also please split up these pull requests into seperate submissions?   This allows us to merge things seperately in the case where we want different things, or there is already a duplicate submission for various tweaks and we prefer one submission over another.

Also also, when adding new fields, please add the "version_added:" attribute on each new field, so it's clear that the attribute is new in a particular version, in this case 1.6, so people running previous versions won't try the options and be confused.

Much appreciated!

atlashealth:
No prob! We'll re-submit today.

PR #6841, opened 2014-04-03
Fix a typo in the subversion module documentation

gawel:
Just added a missing letter


mpdehaan:
Merged, thanks!  This will appear when we next push docs!

PR #6847, opened 2014-04-03
ec2_vol: adds ability to list existing volumes

ghost:
Extends ec2_vol to support the listing of volumes for a given instance id. 

Why is this needed? The built-in facts (and ec_facts module) don't expose EC2 volume id's, which may be needed for other tasks. This feature was modeled on the "list" functionality in https://github.com/ansible/ansible/blob/devel/library/cloud/ec2_tag.


mpdehaan:
Merged, thanks!

PR #6848, opened 2014-04-03
ec2_snapshot: updated docs

ghost:
Included snapshot_tags in the module's documentation. The actual functionality was previously added by another contributor per c900254ed0700d1183c667b883df16795873440b.

### DIFF
--- a/library/cloud/ec2_snapshot
+++ b/library/cloud/ec2_snapshot
@@ -86,7 +86,7 @@ options:
     version_added: "1.6"
   snapshot_tags:
     description:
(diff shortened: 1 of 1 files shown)


mpdehaan:
Just reading through this -- we usually don't require inline JSON as it's somewhat hostile to the user, so we may want to adapt this into an example where we can show it expanded out as a dictionary in the YAML.

atlashealth:
How's this look?

mpdehaan:
Perfect, looking at merging now, thanks!

mpdehaan:
Merged, thanks!  Apologies on the delay, this will be live in the next docs push.

PR #6850, opened 2014-04-03
ec2: added ability to set EBS optimized instances

ghost:
Adds support for EBS optimized instances, which has been part of Boto for about a year now.

### DIFF
--- a/library/cloud/ec2
+++ b/library/cloud/ec2
@@ -217,6 +217,12 @@ options:
     required: false
     default: null
     aliases: []
+  ebs_optimized:
+    version_added: "1.6"
+    description:
+      - whether instance is using optimized EBS volumes, see U(http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSOptimized.html)
+    required: false
+    default: false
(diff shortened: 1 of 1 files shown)


lwade:
It looks like this would potentially break compatibility with EC2-compatible implementations.  Whereas other "advanced" parameters in the module are added by additional steps to construct the parameters, this goes into the base set:

``` python
    if count_remaining == 0:
        changed = False
    else:
        changed = True
        try:
            params = {'image_id': image,
                      'key_name': key_name,
                      'monitoring_enabled': monitoring,
                      'placement': zone,
                      'instance_type': instance_type,
                      'kernel_id': kernel,
                      'ramdisk_id': ramdisk,
                      'ebs_optimized': ebs_optimized,
                      'user_data': user_data}
```

So even if the parameter is not used the ebs_optimised=false option is supplied to the request?

atlashealth:
@lwade - makes sense. Here's a commit that handles it optionally.

lwade:
Thanks @atlashealth, I think that's good.

PR #6852, opened 2014-04-04
cloudformation's disable_rollback must be a bool

tedder:
passing "yes" to boto causes severe problems.


mpdehaan:
The "type=bool" is great but this commit also breaks the aliases line.   Can you please force push to this branch with a corrected commit?

Thanks!

tedder:
wow- I don't know how I did that, and don't know how I missed it when I diffed. I fixed it.

mpdehaan:
merged, thank you!

PR #5910, opened 2014-02-08
apt module: Add support for installing .deb packages

lorin:
Support installing .deb packages from the local filesystem.

```
apt: deb=/tmp/mypackage.deb
```

This change modifies the install() function so that it returns results instead of calling
exit_json() and fail_json directly, so that it can be reused by the new deb_install 
function to install package dependencies.

PR #6856, opened 2014-04-04
Update apt:deb with correct version

drewblas:
The docs site says this option is available in 1.5, but it is not.  
#5910 added the deb option.  The PR was generated two months ago (before 1.5) but was not included until after 1.5 was released.  This fixes the docs.


mpdehaan:
merged, thanks!

PR #6858, opened 2014-04-04
Improve roles intro

techtonik:
This makes it more straightforward for those of us who weak in abstractions. =)


mpdehaan:
Thanks for the submission.

I disagree with the section being moved above, and there's also a grammar error introduced from "roles are" to "roles is".

I think we're going to leave this as is.

techtonik:
Fixed error in a commit, but not updated here, because GitHub doesn't do this for closed PRs. I am not an English native, so I expect smb. who is more sensitive to fix that.

I still don't understand your argument and it would be nice if smb. could explain. It is hard to sent docs enhancements without a quality feedback why they do not fit.

PR #6859, opened 2014-04-04
lvg module bugfix on vg_options patch

srgvg:
When no vg_options are passed to the module, 'vg_options' still exists
in the module.params dict with a value of None, so the default empty string in
the get method is never used. None cannot be "splitted", which backtraced.

### DIFF
--- a/library/system/lvg
+++ b/library/system/lvg
@@ -105,7 +105,7 @@ def main():
             vg=dict(required=True),
             pvs=dict(type='list'),
             pesize=dict(type='int', default=4),
-            vg_options=dict(),
+            vg_options=dict(default=''),
             state=dict(choices=["absent", "present"], default='present'),
             force=dict(type='bool', default='no'),
         ),
@@ -116,7 +116,7 @@ def main():
     state = module.params['state']
     force = module.boolean(module.params['force'])
     pesize = module.params['pesize']
-    vgoptions = module.params.get('vg_options', '').split()
+    vgoptions = module.params['vg_options'].split()
 
     if module.params['pvs']:
         dev_string = ' '.join(module.params['pvs'])


mpdehaan:
Thanks for the submission.

The description talks about the code works, can you please share the example of how the user encounters the problem (and the expected vs actual output) instead?   This helps us much more in testing.

Thanks!

srgvg:
This is about a task with the updated lvg module that doesn't specify the new 'vg_options' parameter. When running my playbook with this updated code, I get following trace:

```
  File "<stdin>", line 1411, in <module>
  File "<stdin>", line 119, in main
AttributeError: 'NoneType' object has no attribute 'split'
```

The bug is about the module not providing a default for this parameter.

mpdehaan:
Sure, so can you please share what commands you shared to trigger this scenario?

srgvg:
It's really straightforward. Bug can be replicated with this playbook:

```
- hosts: localhost                                                              
  gather_facts: false                                                           
  connection: local                                                             
  tasks:                                                                        
    - command: dd if=/dev/zero of=./disk.img bs=1024 count=1024                                                                   
    - command: losetup /dev/loop7 ./disk.img                                    
    - lvg: vg=test pvs=/dev/loop7 state=present                                 
      ignore_errors: True                                                       
    - command: losetup -d /dev/loop7                                            
    - command: rm ./disk.img                                                    
```

Change the lvg task to e.g.  `- lvg: vg=test pvs=/dev/loop7 state=present vg_options=--force` and this bug is not triggered anymore.

neh:
I have the same issue, exactly as described.

fermayo:
+1

xals:
Bump.

echo-devnull:
Ditto. Bump.

srgvg:
Seems another recent PR ( #7810 )  for this was merged without the fuzz we got here :)
I leave this one open, as there's still a line from my patch not in that other one, that I find useful.

jimi-c:
Merged, thanks!

PR #6861, opened 2014-04-04
add distribution_major_version to fact gathering

ericlake:
This will set a fact for ansible_distribution_major_version without the need to install the redhat-lsb and required dependencies.


mpdehaan:
merged, thanks!

PR #6862, opened 2014-04-04
Module rax_clb_nodes: Allow respecification of a node without requiring node_id. Fixes #5401

### LINKED ISSUE #5401: rax_clb_nodes module: state=present errors out on repeated run
When adding a node to a load balance with state=present, it throws an error on subsequent ansible runs with:

"Duplicate nodes detected. One or more nodes already configured on load balancer."

I would assume "present" should just be a changed=False noop if it's already in the LB.

sivel:
This pull request addresses #5401

Previously not specifying `node_id` would cause an error of "Duplicate nodes detected. One or more nodes already configured on load balancer."

This pull request allows node matching by any of node_id, address and port.

Test play:

```
- hosts: localhost
  gather_facts: False
  tasks:
    - name: Server build requests
      local_action:
        module: rax
        credentials: ~/.raxpub
        name: test
        flavor: performance1-1
        image: ubuntu-1204-lts-precise-pangolin
        region: ORD
        state: present
        wait: yes
      register: rax

    - name: Provision Load Balancer
      local_action:
        module: rax_clb
        credentials: ~/.raxpub
        name: my-lb
        port: 8080
        protocol: HTTP
        type: PUBLIC
        timeout: 30
        region: ORD
        wait: yes
        state: present
        meta:
          foo: baz
      register: clb

    - name: Add servers to Load balancer
      local_action:
        module: rax_clb_nodes
        credentials: ~/.raxpub
        load_balancer_id: "{{ clb.balancer.id }}"
        address: "{{ item.rax_networks.private|first }}"
        port: 80
        condition: enabled
        type: primary
        wait: yes
        region: ORD
      with_items: rax.instances
```

### DIFF
--- a/library/cloud/rax_clb_nodes
+++ b/library/cloud/rax_clb_nodes
@@ -137,19 +137,18 @@ def _activate_virtualenv(path):
 
 def _get_node(lb, node_id=None, address=None, port=None):
     """Return a matching node"""
-    searches = {
-        'id': node_id,
-        'address': address,
-        'port': port
-    }
-
     for node in getattr(lb, 'nodes', []):
-        try:
-            if all(getattr(node, attr) == value
-                   for (attr, value) in searches.items() if value is not None):
-                return node
-        except AttributeError:
-            continue
+        match_list = []
+        if node_id is not None:
+            match_list.append(getattr(node, 'id', None) == node_id)
+        if address is not None:
+            match_list.append(getattr(node, 'address', None) == address)
+        if port is not None:
+            match_list.append(getattr(node, 'port', None) == port)
+
+        if match_list and all(match_list):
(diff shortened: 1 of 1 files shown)


mpdehaan:
Thanks for this:

this line is probably too clever:

  if all(getattr(node, attr) == value for (attr, value) in searches.items() if value is not None):

should really be simplified, even if it means breaking up into more statements, as it poses a barrier to reading comprehension and maintaince.

Can you take care of this one?

Thanks!

sivel:
I can...however this logic was taken largely as is from the pyrax findall method.

I'll see what I can do to maybe simplify it.

jimi-c:
Merged, thanks!

Issue #6778, opened 2014-03-31
[rax] it'd be nice if the module returned info about servers that are already running, in addition to info about the servers it starts


joshuaconner:
##### Issue Type:

Bug report 
##### Ansible Version:

ansible 1.6 (devel efba8b4771) last updated 2014/03/31 10:51:21 (GMT -700)
##### Environment:

Running from OSX 10.9
##### Summary:

Basically if you run the `rax` module with the `exact_count` option set to `yes`, the module will only return info about servers it has started. It would be nice if, in addition to that, it returned info about all of the matched servers, including ones that are already running.

That way, you can create in-memory host groups _a la_ https://github.com/ansible/ansible-examples/blob/master/language_features/eucalyptus-ec2.yml
##### Steps To Reproduce:

I've created a gist with a simple playbook and output from running said playbook here: https://gist.github.com/joshuaconner/dca192736971d74e80a8
##### Expected Results:

`rax` would return info about instances that are already running in addition to info about instances it has created.
##### Actual Results:

`rax` only returns info about instances it creates. See again the gist here:
https://gist.github.com/joshuaconner/dca192736971d74e80a8, Note that 
- in the first task, the length of the "instances" array is 1 (this is correct) 
- in the second task the "instances" array is empty (even though one instance in the group is running)
- in the third task the length of the "instances" array is 1 (even though **two** instances in the group are now running)

If you'd like, I'd be happy to either a) take this to the mailing list or b) put together a PR. Let me know, thanks!

joshuaconner:
(cc @sivel, who I think has done most of the work on this module and was super helpful in helping me to get it working with our OpenStack install, though this particular issue isn't OpenStack-specific)

sivel:
This will refactoring to work. Currently, existing servers don't get passed into the 'create' function, only a list of names of servers to create and the details required to create them.

Due to this only newly created instances are available.

I'll look into making these changes.  However, you might want to look into using the rax.py inventory script if you are not already using it.  The inventory script will query the API and return host groups based on the 'group' attribute you used during creation.

bmcsheehy:
Speaking of the rax.py inventory script. It would be nice if the inventory script returned host data for systems in Gen1, NextGen / Perf Tier. As of now it appears to exclude anything in Gen1.

joshuaconner:
@bmcsheehy Open an issue about it? :)

bmcsheehy:
Will do :)

mpdehaan:
[...]

PR #6863, opened 2014-04-04
rax module: Return untouhced servers with exact_count

sivel:
This pull request addresses #6778

With this pull request existing servers that are untouched by the run will be returned in the response from the module.

As describe in a comment on the other ticket:

`instances` will only contain servers that are:
1. successfully created during the run
2. servers matching the group
3. servers that were not just deleted

So when servers are deleted, they will not show up in `instances`, they will only be available via `success`, `timeout` or `error` in such a case.

There is also a slight behavior change with `create`, in that it will not include servers who entered an error state or timed out while waiting. instances in this case will only include already created matching servers and servers just created successfully.

### DIFF
--- a/library/cloud/rax
+++ b/library/cloud/rax
@@ -200,7 +200,7 @@ def rax_slugify(value):
     return 'rax_%s' % (re.sub('[^\w-]', '_', value).lower().lstrip('_'))
 
 
-def pyrax_object_to_dict(obj):
+def server_to_dict(obj):
     instance = {}
     for key in dir(obj):
         value = getattr(obj, key)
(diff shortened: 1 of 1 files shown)


joshuaconner:
Sorry, I just got around to looking at this. It looks great and worked perfectly in my tests! 

`instances` now always contains the currently running instances, `action` is the action that occurred (i.e. if you set `count: 1` and `exact_count: yes` and there had previously been 2 instances running, the action would be `delete`) and `success` contains instances to which `action` has been successfully applied. Thanks for this!

mpdehaan:
merged, thanks!

PR #6865, opened 2014-04-04
Release1.5.4

hansbaer:
I suggest to change the error message for failed MySQL connects in all relevant modules to 

module.fail_json(msg="unable to connect. Check login_user and login_password are correct (alternatively if ~/.my.cnf contains correct credential) and the path to the Unix domain socket (use parameter login_unix_socket if not in default path).")

or something along that line. That makes it clearer that it is not necessarily the credentials which are wrong, but could also be the unix socket. See issue 6860.


mpdehaan:
So first off, thanks for the interest.

Unfortunately, this is a giant mess of version control history changes :)

We don't accept merge requests to release branches as a matter of project management policy -- rather, any problems should be fixed on the development branch, and if we decide to do a maintaince release, we will cherry pick various issues that we want to move over.

So, what I'd suggest here is on the development branch, create a simple topic branch and submit a pull request for just that change -- and most likely this will go out in the next major release.

Also be sure to use git pull --rebase vs git pull, and git rebase vs git merge, to avoid merge commits in the pull request.

If you have further questions, you can also ask for help/details on ansible-devel@googlegroups.com

Thanks!

hansbaer:
I was not aware that I was doing the request on the release branch. Of course it's fine if it just helps someone in the future.

However, my understanding of git is rudimentary, and I was not able to add the changes. Sorry.

PR #6866, opened 2014-04-05
Add check mode support for non-python modules

guysab:
Allows non-python modules to run in check mode.

If a module contains '#SUPPORTS_CHECK_MODE' and check mode is enabled, ansible will call the module even if it's not new-style python. The module will then receive the argument CHECKMODE=True.

Please note this pull request was developed at the same time as my next pull request 'Helper for bash modules'. There's a small reference in the doc to the helper module, so if the two pull requests are not merged at more or less the same time, please ask me to remove the reference. Thanks!

### DIFF
--- a/docsite/rst/developing_modules.rst
+++ b/docsite/rst/developing_modules.rst
@@ -296,10 +296,12 @@ mode, your module will simply be skipped.
 To support check mode in non-python modules, insert a '#SUPPORTS_CHECK_MODE=True' in your source.
 The module argument CHECKMODE will be 'True' when check mode is enabled. For example::
 
+    #!/bin/bash
     #SUPPORTS_CHECK_MODE=True
-    #INCLUDE_BASH_HELPER --args 'somearg1 somearg2 CHECKMODE'
+    ... Parse args file, will contain 'CHECKMODE=True' if in check mode ...
     if [ "$CHECKMODE" = 'True' ]; then
-        ...
+       # Test and flag if changes are needed, but *don't* modify anything!
+       echo "ok=true changed=true"
 
 .. _module_dev_pitfalls:
 
--- a/docsite/rst/developing_modules.rst
+++ b/docsite/rst/developing_modules.rst
@@ -291,6 +291,16 @@ system state is altered when the user enables check mode.
 If your module does not support check mode, when the user runs Ansible in check
 mode, your module will simply be skipped.
 
+.. versionadded:: 1.6
+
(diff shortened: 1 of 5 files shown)


mpdehaan:
Thanks for this.

Unless I'm mistaken, this references the "bash helper magic" prior to it being merged in.

This is more of a note for us, but in places where we are now returning tuples of 3-4 items these should really be moved over to using NamedTuples.

guysab:
As per my other pull request, I'll leave the bash helper out of the picture.

Would you like me to clean this pull request and change it so it uses NamedTuples? This is a feature I'd really like to see in core, as it's a bit messy to maintain on the side with the devel branch. Thanks!

mpdehaan:
Yes, please.

guysab:
Pull request has been updated. Please make sure my namedtuple style is appropriate for you, it seems it's the first time they'd be used in ansible. Thanks!

bcoca:
named tuple was added in 2.6, which is ok for code that executes on the
'master', but not for what executes on target hosts which must support
earlier versions. Since you are using it in module_common this is probably
going to break on older targets.​

guysab:
module_common appears to be used only in runner and hacking, so I suspect it only runs on ansible server. I tested my mod with a python 2.4.3 target which does does not have collections/namedtuple and it worked fine. Are you sure module_common runs on targets?

jimi-c:
Yes, `module_common.py` is master side only - it's things in `module_utils/` that need to be 2.4 compatible.

guysab:
Great. Namedtuples will stay in there then. Please comment on namedtuple style if need be.

bcoca:
my bad, confused the 2, ignore me, move along, nothing to see here.​

mpdehaan:
# Rebase Request

Hi!

which is very active:

amenonsen:
This PR seems reasonable enough, but the rebase request has not been responded to for nearly a year, so it should be closed now.

guysab:
This PR was designed to be merged at the same time as #6867, which was rejected by Michael. Without the #6867 enhancements, this pull request had very minimal value. In the end, I had to fork a private branch of Ansible to implement full Bash support, but my client (City of Montreal) did not like this very much, so the project fell through. Do you see value in this PR without #6867? If so, I'll rebase it. Thanks!

amenonsen:
@guysab sorry to hear about the project falling through. Without #6867 I agree that this isn't much practical use. I don't think it's worth rebasing on its own.

guysab:
Agreed. If anyone is interested in the full bash enhancements, they can always use my fork guysab/ansible_bash_support. Thanks!

PR #6867, opened 2014-04-05
Add helper for modules written in bash

guysab:
Adds bash helper functions for ansible bash modules to:
- Automatically parse the module's arguments directly into environment variables.
- Capture stdout/stderr output and return the output as JSON.
- Intercept script errors and return a meaningful messages to ansible when they occur.
- Build structured JSON responses.
- Assist in debugging.

The entire process is described in more details in the included doc, sample and helper module. This pull request is designed to be implemented at the same time or after my other pull request 'Add check mode support to non-python modules'.

Please feel free to make any suggestions as how to make this contribution better and easier for you to review. Thanks!

### DIFF
--- a/docsite/rst/developing_modules.rst
+++ b/docsite/rst/developing_modules.rst
@@ -347,7 +347,7 @@ and guidelines:
 Shorthand Vs JSON
 `````````````````
 
-To make it easier to write modules in bash and in cases where a JSON
+In cases where a JSON
 module might not be available, it is acceptable for a module to return
 key=value output all on one line, like this.   The Ansible parser
 will know what to do::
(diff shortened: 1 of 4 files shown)


mpdehaan:
I think I said I supported this originally but I'm rather uncomfortable with supporting the density and complexity of the bash code here, and am starting to worry a little bit that this may encourage people to submit modules in bash, or write modules in bash and therefore not be able to submit them to core.

Bash modules can of course output simple key=value data and perhaps there are nice CLI tools for working with JSON, but I don't know... I'm not sure we want to absorb this now.

It's an interesting idea and I don't want to kill it, so I hope you can kind of follow where my thought process is at.

guysab:
I understand, I don't want to mess with ansible's direction :-). Plus you guys are really loaded.

I thought for a very long time of how I could reduce or split the helper to make it more palatable, but in the end I feel its functionality is really most useful when fully integrated as it is. If there's a subset of the helper would like to see in core let me know!  Otherwise, unless you have other thoughts of how this could be adapted in a way that fits better with ansible's near future, I'm thinking the best would probably for me to simply publish it on the side, which was my original intention anyhow. This will also give me the opportunity to fully support it directly, so you guys don't have to worry about bash stuff.

What do you think?

mpdehaan:
I think we've had some discussion on this on the -devel list in the past, but we need to move to a point where any helper function is more pluggable before we get to this.

Ultimately, I am also not sure about bash long term, given the 'script' module is there for basic bash (non-modules) and so on.

I'm still open to disucssion on ansible-devel should you wish to, though it's likely we are going to want to force pure JSON to all modules soon, in which case, it may become harder to do bash in the future.

PR #5929, opened 2014-02-09
add mongodb replicaset support

cove:
user add will fail if you're not connected to the current primary (the primary can change), specifying the replicaset allows pymongo to automatically determine which host is the primary and add the user to the correct host.

PR #6869, opened 2014-04-05
add replica_set attr to mongodb_user module

cove:
This is an updated re-submission of #5929 to add a replica_set attr, along with error handling improvements so the underlining causes for errors are reported up.

Changes:
- merged in 66b2d55
- renamed replicaset to replica_set
- added version_added to replica_set attr
- removed unrelated changes from pull request


mpdehaan:
merged, thank you!

PR #6870, opened 2014-04-06
Modified Cron to add AIX Crontab support

MorrisA:
changed according to suggestions from mpdehaan and redone entirely to make it a single commit to the latest version.


MorrisA:
This leaves two calls to platform.system() in the cron module if self.user is set and the OS is not Solaris.  I could pull that out so that a variable is set for platform.system() but I'm not sure it's worth the extra trouble.

mpdehaan:
merged, thank you!

PR #6871, opened 2014-04-06
Add S3 lookup plugin

dragon3:
This is a new lookup plugin for S3.

Example playbook:

```

---
- hosts: localhost
  gather_facts: false
  vars:
    - myname: "{{ lookup lookup('s3', 'bucket=sample-bucket object=/foo/myname.txt') }}"
  tasks:
    - debug: msg="Hello, {{ myname }}"
```

### DIFF
--- a/lib/ansible/runner/lookup_plugins/s3.py
+++ b/lib/ansible/runner/lookup_plugins/s3.py
@@ -1,0 +1,75 @@
+# (c) 2014, Ryuzo Yamamoto <ryuzo.yamamoto(at)gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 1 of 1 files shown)


mpdehaan:
If someone is doing this in a host loop across 500 servers to get a value, this is going to be very inefficient.

Hmmm...

This is interesting. I'm inclined to say this isn't a great fit for a lookup plugin, and perhaps someone should use the s3 module to download a file instead, and then use the contents value with register.

The reason being is that variables in "vars", (and everywhere) are lazy evaluated in the individual host loops.

This of course doesn't mean you can't use this, though I think it would be too hammering on s3 to include in core ansible.

If you'd like to discuss on ansible-devel please stop by.

Thanks!

dragon3:
@mpdehaan Thank you for your review, I understood.

I'd like to get sensitive variables from S3 and set template, put file on servers.
Sensitive variables are for example password, API key, and so on.

As you said, I can use s3 module to download a file and then use the contents value with register,
but if I can use this s3 lookup plugin, I think it is very simple.

Thank you for your consideration.

PR #6877, opened 2014-04-06
Sudo template parameter

### LINKED ISSUE #5942: allow sudo variable to be set by host_vars variables
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.5 (devel 36250c6b86)
##### Environment:

I was running Ansible from Mac OSX 10.9.1 and managing CentOS release 5.7 (Final)
##### Summary:

sudo variable in my playbook can not be set by host_vars files variables
##### Steps To Reproduce:

1) I have a playbook with the following:

```
tasks:
 name: test
 command: do something
 sudo: "{{item.use_sudo}}"
 with_items: somelist
```

2) I have a host_var file with the following:

```
somelist:
- {name: jim, use_sudo: yes}
- {name: jim2, use_sudo: no}
```

3) When I run my playbook which references "somelist" from the host_vars file, I get a "Permission Denied" error

Looks like it was not able to set the sudo variable properly.

However, if I use "sudo: yes" in my playbook, then everything works fine
##### Expected Results:

The expected behavior should be that I can set the sudo variable with whatever value I pass from the host_vars variables
##### Actual Results:

I get "Permission Denied" error because sudo was not set

noriordan:
Attempting to resolve: https://github.com/ansible/ansible/issues/5942

### DIFF
--- a/lib/ansible/playbook/play.py
+++ b/lib/ansible/playbook/play.py
@@ -27,6 +27,7 @@ import shlex
 import os
 import sys
 import uuid
+from ansible.callbacks import vvv
 
 class Play(object):
 
@@ -156,9 +157,7 @@ class Play(object):
         # apply any missing tags to role tasks
         self._late_merge_role_tags()
 
-        if self.sudo_user != 'root':
-            self.sudo = True
-
+    
         # place holder for the discovered hosts to be used in this play
         self._play_hosts = None
 
--- a/lib/ansible/playbook/play.py
+++ b/lib/ansible/playbook/play.py
@@ -27,7 +27,6 @@ import shlex
 import os
 import sys
 import uuid
-from ansible.callbacks import vvv
 
 class Play(object):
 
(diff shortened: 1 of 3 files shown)


mpdehaan:
I'm a bit worried about the "attempting" part :)

From list discussions, we had suggested making the 'avoid duplicate sudo' behavior configurable, which needs to be the case, rather than changing this wholesale -- we can't change the default behavior, but would accept a config option.

Can you please resubmit this in a way that makes this behavior configurable?

mpdehaan:
This change is somewhat intrusive and includes some spurious commits.  I believe this could be solved instead by just moving the boolean function down and including the following lines:

```
   if self.sudo is not None:
         self.sudo = utils.boolean(template.template(self.basedir, self.sudo, inject))

     if self.su is not None:
         self.su = utils.boolean(template.template(self.basedir, self.su, inject))
```

Which would also avoid the new variable "sudo_var"

PR #6878, opened 2014-04-07
ec2: Fix bug with running/stopped state and wait=yes.

sayap:
If `get_all_instances` returns multiple reservations, the old wait loop only
dealt with the first reservation. Thus, the wait loop may end before all
instances get to be running/stopped.

Also clean up the code a little.

### DIFF
--- a/library/cloud/ec2
+++ b/library/cloud/ec2
@@ -193,7 +193,7 @@ options:
   instance_ids:
     version_added: "1.3"
     description:
-      - list of instance ids, currently only used when state='absent'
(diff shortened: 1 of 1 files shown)


jimi-c:
Beyond the above issue, this looks good. Just get that cleaned up and tested and I'll get this in queue for testing. Thanks!

sayap:
Ah, good spot on the `+=` bug. I will do a rebase with the bug fix and the list comprehension change, plus a little clean up.

jimi-c:
Merged, thanks!

PR #6880, opened 2014-04-07
Storm module for Ansible

sgargan:
Add storm module that can be used to deploy and manage topologies within an Apache Storm realtime compute cluster


sgargan:
there are errors in this pull request. I will resubmit one with them removed.

PR #6883, opened 2014-04-07
Adding bold to service doc to call up state and enabled.

ericlake:
The docs for the service module do not show state and enabled as being required. It is easy to miss that one or both are required in the docs. Making this bold to try to make it stand out more.


mpdehaan:
merged, thank you, will appear in next docs push.

PR #6890, opened 2014-04-08
docker: added support for read-only volume access. Fix #6884

### LINKED ISSUE #6884: docker module doesn't support volume access rights
##### Issue Type:

"Feature Idea” and “Documentation Report”
##### Ansible Version:

ansible 1.6 (devel 8d3e22ccc4) last updated 2014/04/01 17:31:20 (GMT +100)
##### Environment:

Linux Ubuntu server 13.10 x64
##### Summary:

I've realised that docker module doesn't support the access rights specification.

Although without specify them works properly mapping them as rw, it is quite confusing how to map them, so I spent some time trying to figure out why the directory wasn't mapped and the problem was when I specified the volumes as e.g. volumes="/host_machine:/tmp:rw" it was running the container with the expected mapping, basically because the module skips as container-path has been specified when the string split parts is different of 2 (https://github.com/ansible/ansible/blob/devel/library/cloud/docker#L352)
##### Steps To Reproduce:

Just run any task with `docker module` and use `volumes` property providing a string as `host_port:container_port:rw|ro`
##### Expected Results:

Map the directory between host machine and docker container as expected by using [docker run](http://docs.docker.io/en/latest/reference/commandline/cli/#run)
##### Actual Results:

Not sure, but probably the result is the same than only providing a `volumes=host_port`.

softzilla:
Allow specify volumes with access protection: "/var/cloud:/tmp/cloud:ro"

Mount /var/cloud with read-only access to /tmp/cloud. Try write file "echo" into write-protected directory.

```
    - name: "TEST"
      docker:
        image: ubuntu
        volumes:
          - "/var/cloud:/tmp/cloud:ro"
        command: /bin/bash -c 'echo 1 > /tmp/cloud/echo'
      tags:
        - deploy
```

Result:

```
docker ps -a
CONTAINER ID        IMAGE                  COMMAND                CREATED             STATUS              PORTS                                      NAMES
472c1e055df9        ubuntu:latest          /bin/bash -c echo 1    2 seconds ago       Exit 1

docker logs 472c1e055df9
/bin/bash: /tmp/cloud/echo: Read-only file system
```


mpdehaan:
Can you please submit this without merge commits?    We don't allow these in the code as they mess with history and are difficult to audit.

Use "git pull --rebase" vs "git pull" and "git rebase" vs "git merge" and this will avoid most problems.

I'd also highly recommend using a feature branch for each new feature to keep things separate.

Thanks!

softzilla:
Fixed.
Replace with #6945

PR #6892, opened 2014-04-08
modules.rst: be consistent with service module documentation

jaseg:
See valid values for state in http://docs.ansible.com/service_module.html#options


mpdehaan:
merged, thanks!  This will appear when we next push docs.

PR #6899, opened 2014-04-08
Add portability fix for hacking/env-setup

mgrubb:
Change `` usage to $()
Add check for KSH93 to set HACKING_DIR
Pure shell fix for FULL_PATH


bcoca:
closing as these changes were done as part of another PR

abadger:
I noticed we didn't have the KSH stuff so I cherrypicked that.

bcoca:
good catch

PR #6902, opened 2014-04-08
Have paramiko use /etc/ssh_known_hosts

vegitron:
Fixes an issue with a confusing error: "paramiko: The authenticity of host '[host]' can't be established" when ssh on the command line doesn't complain


vegitron:
I'm going to resubmit this pull request on a branch, so it doesn't get mixed in with the other pull request.

PR #6907, opened 2014-04-08
Let's also drop usernames and passwords.

jjshoe:


jjshoe:
previously: /etc/apt/sources.list.d/bob_qwerty_s3_amazonws_com_my_repo.list
now: /etc/apt/sources.list.d/s3_amazonaws_com_my_repo.list

bob/qwerty being the user pass.

jjshoe:
WARNING: I should note that this fix WILL result in changes being reported for anyone who:

1) Upgrades to the version of ansible this gets released in

AND

2) Runs ansible

AND

3) Uses username and passwords with apt

This change will NOT notify the users in any way that they have a file that anyone can read on disk named with their apt credentials.

jjshoe:
I wouldn't merge this until some discussion happens over here: https://github.com/ansible/ansible/issues/6908

mpdehaan:
So this is good but the ticket in question mentions a bit more that needs to be dealt with -- apologies on duplicate efforts -- though there's a bit more to be done.

Let's continue this discussion on that ticket.

PR #6913, opened 2014-04-09
Use common module document fragments for AWS modules

willthames:
Created common module doc fragment, and applied to all
modules that use ec2_connect or connect_to_aws as
they definitely share the common doc fragments

Thanks to the fantastic work by @sivel that this is based upon.


jctanner:
@willthames thanks!

PR #6914, opened 2014-04-09
Use common code for ec2_elb and ec2_elb_lb

willthames:
Uses the new `get_aws_connection_info` and `connect_to_aws` common methods to reuse code

Now complains if region is not set in one of the three possible methods

Also moved over to common documentation code so this is actually based on #6913


jctanner:
@willthames thanks!

PR #6915, opened 2014-04-09
Random filter

bcoca:
returns a random int using the filtered input as the top/end/stop.

```
{{ 59|random}} * * * * /cron/script
```

optionally it takes a start (0 is default) and a step (1 is default)

```
{{ 100|random(10, 2)
```

only even numbers from 10-100


bcoca:
wth??? another random commit sneaked in

PR #6917, opened 2014-04-09
simple random filter

bcoca:
returns a random int using the filtered input as the top/end/stop.

```
{{ 59|random}} * * * * /cron/script
```

optionally it takes a start (0 is default) and a step (1 is default)

```
{{ 100|random(10, 2)}}
```

only even numbers from 10-100

### DIFF
--- a/docsite/rst/playbooks_variables.rst
+++ b/docsite/rst/playbooks_variables.rst
@@ -234,6 +234,27 @@ be used.  The default is ``False``, and if set as ``True`` will use more strict 
 
     {{ sample_version_var | version_compare('1.0', operator='lt', strict=True) }}
 
+.. _random_filter
+
+Random Number Filter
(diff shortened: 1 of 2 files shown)


jctanner:
@bcoca thanks!

fredjean:
There was already a random filter that we were using in combination with the cron module to spread some activity across a largish cluster of system:

```
{{ range(0,59) | random }}
```

This was already available in 1.5.4. This also introduces a breaking change when trying to use the code in the devel branch.

I like the new syntax, but this is causing us a little bit of heartburn.

AnneTheAgile:
Note that the code here was not used in the end - #7113 code was used instead. The diff is here; https://github.com/psa/ansible/commit/83f93ba6eae4a4d59a5858b16ab48ec0215e95d1

PR #6918, opened 2014-04-09
Add Nagios NDO inventory plugin

jle64:
Add an inventory plugin that uses hosts and hostgroups from Nagios NDO database.
Only tested with NDO on MySQL but uses SQLAlchemy so should work with postgres too.


amenonsen:
Anyone using NDO who can test this PR?

bcoca:
@gentledevil also can you change PR to target new directory in contrib/inventory?

jle64:
@bcoca : done. I also made little changes so that Travis build passes for all Python versions.

PR #6921, opened 2014-04-09
remove "No hosts matched" from --list-hosts

bugi:
Ansible --list-hosts says "No hosts matched" on stdout when it matches
no hosts.  Emitting "No hosts matched" on stdout interferes with a
script doing something useful with the lack of matching hosts.

```
% ansible -i hosts --list-hosts fred
No hosts matched
% ansible -i hosts --list-hosts fred >/dev/null
% for h in $( ansible -i hosts --list-hosts fred ) ; do echo $h ; done
No
hosts
matched
% ansible --version
ansible 1.5.3
```

My preference is to simply emit nothing in this case, not even on
stderr.  That no hosts were matched is evident from the lack of
hosts emitted.

Note that this behavior is new in 1.5.  It didn't do this in 1.4.3.

### DIFF
--- a/bin/ansible
+++ b/bin/ansible
@@ -94,7 +94,6 @@ class Cli(object):
             inventory_manager.subset(options.subset)
         hosts = inventory_manager.list_hosts(pattern)
         if len(hosts) == 0:
-            callbacks.display("No hosts matched")
             sys.exit(0)
 
         if options.listhosts:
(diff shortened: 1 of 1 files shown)


jimi-c:
So it looks like 46554bbc30d00a91e96fc0b7e051d5c33cf9d0d3 changed this, with the most important part being that the message now goes to stdout instead of stderr. So I think the correct approach here would be to partially-revert that so that the error again goes to stderr, which would make it easier to split off with `2>/dev/null` if you don't care about it.

cc @j2sol

omgjlk:
Hrm. So in ansible-playbook, the message "skipping: no hosts matched" is printed to stdout, which is why I think I made ansible do that as well. I'm not opposed to having it go to stderr instead, where it used to go.

jimi-c:
I can make it consistent in both places.

mpdehaan:
Fixed on devel (messages go to stderr).  Thanks.

PR #6923, opened 2014-04-09
Added region to nova client connection. 

CaptTofu:
Added the argument to nova client instantiation. Without it, it will fail if the cloud provider you are using uses multiple regions because the driver doesn't know how to disambiguate.

I made this pull request on an earlier branch. I'm re-doing so as to make it simple.

### DIFF
--- a/library/cloud/nova_compute
+++ b/library/cloud/nova_compute
@@ -249,6 +249,7 @@ def main():
                               module.params['login_password'],
                               module.params['login_tenant_name'],
                               module.params['auth_url'],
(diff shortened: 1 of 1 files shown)


mpdehaan:
region_name as a parameter takes a scalar type right now.  Should it be made "type='list'" in the argument spec?

Ansible is smart enough to coerce strings to lists in that case.

dmage:
Not, it should be scalar. Duplicate: #6599

jimi-c:
Since this is a duplicate of #6599, I will close this since they are identical. Thanks!

PR #6926, opened 2014-04-09
[bugfix] AWS VPC security group membership fix.

ebrevdo:
- facts now provides a list of existing vpc security group memberships
- adding a comma separated list of sg-\* tags properly modifies membership (previously: threw an exception)

boto has been using this modified interface since I believe version 2.12.0.

### DIFF
--- a/library/cloud/rds
+++ b/library/cloud/rds
@@ -465,7 +465,9 @@ def main():
         params["security_groups"] = security_groups.split(',')
 
     if vpc_security_groups:
-        params["vpc_security_groups"] = vpc_security_groups.split(',')
+        params["vpc_security_groups"] = [
+            boto.rds.VPCSecurityGroupMembership(vpc_group=x) if x else None
+            for x in vpc_security_groups.split(',')]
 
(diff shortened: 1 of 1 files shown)


jimi-c:
Two notes:
1. You're using list comprehensions in this code, however those do not work with python 2.4. We try and keep all module code 2.4 compliant, so those need to be fixed.
2. If this syntax was introduced in 2.12.0, you'll need to add a boto version check to make sure this won't cause a stack trace on older versions of boto. A clean fail_json with a clear message should be used to fail cleanly if the user's boto version is too old.

Address those two issues and we can get this merged in. Thanks!

mpdehaan:
There are list comprehensions in 2.4 but they aren't as advanced as what you can do in later versions -- still might be an issue.

jimi-c:
Merged, after making a few other small improvements and adding in the list comprehension fix I mentioned above. Thanks!

ebrevdo:
Thank you for following up as I've been on vacation!

PR #6931, opened 2014-04-10
Filter hosts based on ec2 variables in inventory script

hkariti:
This adds a host_filter option to ec2.ini that adds only hosts that pass the filter to the inventory. Can be used to separate stage and prod envs in the same region via tags, for example.

### DIFF
--- a/plugins/inventory/ec2.ini
+++ b/plugins/inventory/ec2.ini
@@ -14,6 +14,10 @@
 regions = all
 regions_exclude = us-gov-west-1,cn-north-1
 
+# Use only hosts that pass though this filter, based on ec2 variables (including tags).
+# Can be variable name or var=value.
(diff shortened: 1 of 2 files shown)


mpdehaan:
Can you please provide some examples in the comments on this ticket about how you use this one?

I'm also confused by the trailing = in "+#host_filter = ec2_tag_env="

Thank you!

hkariti:
I added an example as comments in ec2.ini. Basically you put either a variable name to check for the presence of this variable, or a variable and value:

host_filter = ec2_tag_tagName
host_filter = ec2_tag_tagName = some_value
host_filter = ec2_placement = us-west-1c

Adding a "=" at the end is simply an empty value:
host_filter = ec2_tag_should_be_empty=

hkariti:
Closing this in favor of #7480

Issue #6932, opened 2014-04-10
when changing repo address git module fails


pkit:
##### Issue Type:

Bug report
##### Ansible Version:

ansible 1.6 (master c2886e0295) last updated 2014/04/10 11:59:10 (GMT +300)
##### Environment:

DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=12.04
DISTRIB_CODENAME=precise
DISTRIB_DESCRIPTION="Ubuntu 12.04.4 LTS"
##### Summary:

when changing repo URL address git module fails to fetch the needed version
##### Steps To Reproduce:

```
- name: get foo from git
  action: git repo={{ foo_repo }} dest={{ foo_home }} version={{ foo_version }} update=yes force=yes
```

If I change "foo_repo" and "foo_version" git module fails to fetch the new version, because it seems to ignore that the repo url has changed and tries to find it in the existing repo.
##### Expected Results:

should fetch the version from the new remote that was supplied in the git command
##### Actual Results:

tries to fetch it from the old remote

mpdehaan:
This is a known issue about what happens when changing the repo to a different location, so I'm closing this as a duplicate.

PR #6933, opened 2014-04-10
fix issue #6932

pkit:
add `git remote set-url` before fetch to set the remote repo to the new url, if provided


jimi-c:
Merged, thanks!

PR #6937, opened 2014-04-10
Fix hosts file location

fbiville:
The tutorial wrongly mentions /etc/ansible/hosts.
This at least does not work with the version of Ansible I just installed (ansible 1.6) where the active configuration file is located under $ANSIBLE_HOSTS.

### DIFF
--- a/docsite/rst/intro_getting_started.rst
+++ b/docsite/rst/intro_getting_started.rst
@@ -47,7 +47,7 @@ Your first commands
 
 Now that you've installed Ansible, it's time to get started with some basics.
 
-Edit (or create) /etc/ansible/hosts and put one or more remote systems in it, for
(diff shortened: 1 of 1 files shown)


mpdehaan:
/etc/ansible/hosts is in fact the default location for the inventory file so it makes sense to mention it here.

It may be you have an ansible.cfg that picked another location, but I think this is clearer than referencing an env variable that people might not know about yet.

Of course the host location can also be specified with -i.

fbiville:
It is a fresh first install of Ansible in my case and it cost me ~20 minutes to figure out that the default location file might be somewhere overridden. At this point in the tutorial, nothing helps for troubleshooting this, that's just frustrating for newcomers to Ansible like me.

fbiville:
I am not saying the change I pushed into the docs is the one to merge, but there might be a needed improvement to make here, in my opinion.

jeremyajohnson:
I ran into this as well this weekend.

PR #6938, opened 2014-04-10
wait_for module: Allow search_regex to be used with port

sivel:
It is possible that someone may want to look for a string in a port connection with `wait_for` to validate that things are indeed ready and not just open.

For example the following could verify that "OpenSSH" is in the response on port 22:

```
$ ansible all -i localhost, -c local -m wait_for -a 'host=test.example.org port=22 search_regex=SSH'

localhost | success >> {
    "changed": false,
    "elapsed": 1,
    "path": null,
    "port": 22,
    "search_regex": "OpenSSH",
    "state": "started"
}
```

### DIFF
--- a/library/utilities/wait_for
+++ b/library/utilities/wait_for
@@ -76,9 +76,9 @@ options:
     version_added: "1.4"
     required: false
     description:
-      - with the path option can be used match a string in the file that must match before continuing.  Defaults to a multiline regex.
-   
-notes: []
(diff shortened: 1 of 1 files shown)


mpdehaan:
This needs to be updated to indicate version_added is 1.7, and because this code is a little low level, should really be moved to a function and given some comments.

Thanks!

jimi-c:
Merged, thanks!

PR #6943, opened 2014-04-10
Make test-module interpret --args='{...' as yaml.

fxkr:
I found this pretty useful when testing modules with complex args.

Of course reusing --args for this is somewhat hacky. But then
there is already --args='@...' which loads a YAML file, so...


mpdehaan:
Sounds good to me, merged thanks!

PR #6944, opened 2014-04-11
Explain how to do multiple paths in roles_path

ex-nerd:
Document that roles_path can support multiple colon-separated paths


mpdehaan:
Merged, thank you!

PR #6952, opened 2014-04-11
Added example demonstrating other functionality

juliedavila:
Simple addition to the examples list in the users module. Demonstrates uses of 'groups', 'append', and 'shell'.


mpdehaan:
merged, thank you!

PR #6956, opened 2014-04-11
docker: from API 1.10 dns and volumes_from should be passed to start()

mattjbray:
See https://github.com/dotcloud/docker-py/pull/200


mpdehaan:
Merged, thanks!

PR #6958, opened 2014-04-11
Add python bools with casing

ngoyal:
Ansible 1.5.3 required me to do `choices=BOOLEANS + ['True',True,'False',False]` to keep the vars clean.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -33,8 +33,8 @@ MODULE_ARGS = "<<INCLUDE_ANSIBLE_MODULE_ARGS>>"
 MODULE_LANG = "<<INCLUDE_ANSIBLE_MODULE_LANG>>"
 MODULE_COMPLEX_ARGS = "<<INCLUDE_ANSIBLE_MODULE_COMPLEX_ARGS>>"
 
-BOOLEANS_TRUE = ['yes', 'on', '1', 'true', 1]
-BOOLEANS_FALSE = ['no', 'off', '0', 'false', 0]
+BOOLEANS_TRUE = ['yes', 'on', '1', 'true', 1, 'True', True]
+BOOLEANS_FALSE = ['no', 'off', '0', 'false', 0, 'False', False]
 BOOLEANS = BOOLEANS_TRUE + BOOLEANS_FALSE
 
 # ansible modules can be written in any language.  To simplify


mpdehaan:
Can you share where it would not accept True/False as strings in the system?

Wanted to know so we can attempt to reproduce.  Thanks!

ngoyal:
Please see https://github.com/ngoyal/ansible-bool-sample - this may be a bug in the way the module is written, in which case I apologize.

jimi-c:
Calling boolean() on the value should not be required, since the `type='bool'` option already calls that on the value. I'm also confused as to why you'd need this, as the boolean() function checks if the argument is a boolean already, and if so returns it as-is. I believe you just need to remove the extra call to boolean(), and this will work as expected. Thanks!

ngoyal:
Are you talking about the call to module.boolean in my module?  I removed it and a quick test shows it still fails with the following output:

```
failed: [localhost] => (item={'sample_val': True, 'sample_path': '/tmp/yes'}) => {"failed": true, "item": {"sample_path": "/tmp/yes", "sample_val": true}}
msg: value of sample_val must be one of: yes,on,1,true,1,no,off,0,false,0, got: True
failed: [localhost] => (item={'sample_val': False, 'sample_path': '/tmp/no'}) => {"failed": true, "item": {"sample_path": "/tmp/no", "sample_val": false}}
msg: value of sample_val must be one of: yes,on,1,true,1,no,off,0,false,0, got: False
failed: [localhost] => (item={'sample_val': True, 'sample_path': '/tmp/true'}) => {"failed": true, "item": {"sample_path": "/tmp/true", "sample_val": true}}
msg: value of sample_val must be one of: yes,on,1,true,1,no,off,0,false,0, got: True
failed: [localhost] => (item={'sample_val': False, 'sample_path': '/tmp/false'}) => {"failed": true, "item": {"sample_path": "/tmp/false", "sample_val": false}}
msg: value of sample_val must be one of: yes,on,1,true,1,no,off,0,false,0, got: False

```

I think there is a bigger issue with the way values get converted via with_items.  Note the _msg: value of sample_val must be one of: yes,on,1,true,1,no,off,0,false,0, got: True_

jimi-c:
Remove the choices=BOOLEANS too, that is old syntax which we don't use anymore. It will automatically be converted and clipped to a boolean value.

ngoyal:
Excellent - looks like that did the trick.  At least on 1.6.2.  Thanks a lot!

PR #6959, opened 2014-04-11
Add probe functionality to modprobe

dlundquist:
Often times you want to use modprobe to load a module if a hardware
device is present. Adding state=probe that will attempt to load a
module and report changed if the module was loaded, no change if the
a hardware device was not present and error if another error occurred.

### DIFF
--- a/library/system/modprobe
+++ b/library/system/modprobe
@@ -31,7 +31,7 @@ options:
     state:
         required: false
         default: "present"
-        choices: [ present, absent ]
+        choices: [ present, absent, probe ]
         description:
             - Whether the module should be present or absent.
     params:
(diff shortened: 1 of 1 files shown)


mpdehaan:
So just to be clear, this is the same as "present" in your eyes, but intended to do nothing if the HW does not exist, correct?

dlundquist:
Correct, this is the same as "present" except in the case that loading the module returns ENODEV ("No such device") in which case it is a no op.

amenonsen:
This commit looks like it should still apply (or be trivially easy to modify to apply), but should now be targeted to system/modprobe.py in the ansible-modules-extra repository.

PR #6969, opened 2014-04-12
Fix empty group name in rax.py output

sivel:
This pull request addresses an issue where the output would contain a group that was an empty string.

The invalid output looked like:

```
{
    "": [
        "ubuntu",
        ... 
    ],
    ...
}
```


mpdehaan:
Merged, thank you!

PR #6970, opened 2014-04-13
Add reference to the join filter.

nZac:
Simply a reference to the `join` filter in Jinja that can be "Useful" from time to time.


mpdehaan:
Merged, thanks!  This will appear in our next docs push.

PR #6973, opened 2014-04-14
Dynamically loadable modules

jjshoe:
Short: Allow you to specify a non-existent module at compile time like you could with action. 

Why:

You want to try a module from 1.6, but 1.5.x is the latest release
You want to develop a module, and you want to keep track of your changes on git
You developed a module, and you're waiting for a pull request to be accepted but you need your co-workers to use it
You developed a module and never intend to share it with the world

For all of the above cases, you no longer need to maintain your own ansible fork.

### DIFF
--- a/lib/ansible/playbook/task.py
+++ b/lib/ansible/playbook/task.py
@@ -31,7 +31,7 @@ class Task(object):
         'local_action', 'transport', 'sudo', 'remote_user', 'sudo_user', 'sudo_pass',
         'items_lookup_plugin', 'items_lookup_terms', 'environment', 'args',
         'any_errors_fatal', 'changed_when', 'failed_when', 'always_run', 'delay', 'retries', 'until',
-        'su', 'su_user', 'su_pass', 'no_log',
+        'su', 'su_user', 'su_pass', 'no_log', 'dynamic_module',
     ]
 
     # to prevent typos and such
@@ -41,7 +41,7 @@ class Task(object):
          'delegate_to', 'local_action', 'transport', 'remote_user', 'sudo', 'sudo_user',
          'sudo_pass', 'when', 'connection', 'environment', 'args',
          'any_errors_fatal', 'changed_when', 'failed_when', 'always_run', 'delay', 'retries', 'until',
-         'su', 'su_user', 'su_pass', 'no_log',
+         'su', 'su_user', 'su_pass', 'no_log', 'dynamic_module',
     ]
 
     def __init__(self, play, ds, module_vars=None, default_vars=None, additional_conditions=None, role_name=None):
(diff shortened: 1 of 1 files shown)


jjshoe:
To test I dropped a module in place after I kicked off ansible, worked like a charm. 

`- name: Testing dynamic module
  dynamic_module: yes
  joel: msg="oh hai"`

lib/ansible/playbook/task.py
```diff
@@ -66,20 +66,11 @@ def __init__(self, play, ds, module_vars=None, default_vars=None, additional_con
 
             # code to allow for saying "modulename: args" versus "action: modulename args"
             if x in utils.plugins.module_finder:
```
jjshoe, line 68: I could have done:

`if x in utils.plugins.module_finder or ('dynamic_module' in ds and x != 'dynamic_module' and x not in Task.VALID_KEYS and not x.startswith("with_")):`

But felt that was a hideous if statement to grok

srgvg:
Where exactly are you going to drop those extra modules?

You can already drop custom modules in a ./library directory, relative to your playbooks.
Same with pretty much all plugins actually, so this includes ./action_plugins.

jjshoe:
@sergevanginderachter - Which sounds great until you look at all but one of the uses cases where the module becomes part of core. Sure, I could just delete the role, and hunt out everywhere the role is used, but I would also have to make sure no one did anything custom in the task for that role as well.

I would use a local_action to put them in library, of course.

srgvg:
(You are talking about roles now, I presume you meant modules.)

I'm afraid I don't understand the case around this. Where do you keep your custom modules to make them available? What does 'dynamic_module' exactly do? Can you show a more elaborate example?

jjshoe:
You keep them anywhere you want, because you use an ansible task with local_action to bring them onto each person who needs to use them. Think librarian/berkshelf/galaxy, except for modules. 

This isn't elabtorate.

my_module: msg="hi"
dynamic_module: 1

is the same as:

action: my_module msg="hi"

jjshoe:
And no, I was talking about roles, as you said you could just do this with roles, which you can, but when something gets accepted upstream it's far more work to un-spin a role.

srgvg:
not role specic, you can have just 1 directory next to your playbooks where you group all custom modules

srgvg:
So the idea is to distribute custom modules, with ansible, to hosts where ansible is run from?

jjshoe:
Yes.

srgvg:
I still don't see how ansible is then supposed to pick up those modules and run them, and how that would be different from putting them in ./library?

jjshoe:
Would you prefer to:

1) Automate installing a custom module for your co-workers
2) Letting them know via e-mail/wiki/im/sms/carrier pidgin

jjshoe:
I highly recommend patching your ansible and trying this, or at the very least mimicking it by specifying action: some_module for you to better understand how this works. 

Edit: As stated above ALL I'm doing, is implementing the same behavior as action when you specify my option:
- action: cow foo=bar
- cow foo=bar

srgvg:
custom modules are part of my playbook repository, so this is totally transparent

jjshoe:
And when you want to use it with a different playbook do you include the role?

mpdehaan:
I really don't like how this introduces a new keyword that is hard to explain to people here.

There's already a module path in ansible that can be configured for looking in different places.

Let's start a thread on ansible-devel to discuss.

jjshoe:
I'll take this as not interested in implementing this. I'll just use action on modules then that I want to load dynamically. I simply used the new keyword so you could still implement typo checking on existing distributed modules.

Completely disinterested in joining a mailing list. This isn't 1995 and I don't have time to sort through crap that's not relevant.

jjshoe:
Another possible way to do this would be use to galaxy for more than roles. However I imagine that'll take far longer to accomplish.

jjshoe:
"ANSIBLE_LIBRARY - Override the default ansible module library path"

Unless it's an append operation, I then have to maintain an entire directory of modules, as well, not just the one I want to change.

jjshoe:
"Ansible knows how to look in multiple locations if you feed it a colon separated path, and it also will look for modules in the ”./library” directory alongside a playbook."

Ok, so you can feed in more than one path via ansible.cfg, which is good, but I want to use ansible to grab the library for me. I'm still not able to do so, because I'll fail the module name validation stuff my PR gets you around by using the keyword.

I'm happy to do the work to make this work, just let me know here or on IRC what you think it should be.

PR #6975, opened 2014-04-14
Adds a new module, ec2_iam_role.

scottanderson42:
This module can create and delete IAM roles, policies, and instance profiles for association with an instance, either via instance creation or with a launch configuration.

### DIFF
--- a/library/cloud/ec2_iam_role
+++ b/library/cloud/ec2_iam_role
@@ -74,6 +74,37 @@ EXAMPLES = '''
           policy: "{{ lookup('file', 'access_s3_file.json') }}"
         - name: policy_name2
           policy: "{{ lookup('file', 'access_policy2_file.json') }}"
+
+- ec2_iam_role:
+    state: present
+    name: some-role
+    instance_profile_name: some-role-permissions
+    trust_policy:
+      Version: "2012-10-17"
+      Statement:
+        - Effect: "Allow"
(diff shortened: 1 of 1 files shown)


mpdehaan:
Can you expand out the JSON in the example rather than using the lookup plugin?  (i.e. inline data structures?)

We can show both but I think it would be useful to show ways of doing this without the JSON.

Thanks!

scottanderson42:
I need to change how the parms are read due to the issue with JSON being parsed in complex args vs. non-complex; I'll get back to this.

scottanderson42:
All set, thanks for the feedback.

jimbydamonk:
This looks good. +1, I could really use this.

terretta:
In documentation section, this has "module: ec2_asg", presume "ec2_iam_role" instead?

If missing region, instead of asking to supply region, fails in get_existing_role
    existing_role = connection.get_role(role_name)
AttributeError: 'NoneType' object has no attribute 'get_role'

mpdehaan:
cc @jsmartin - is this something you might be interested in testing out?

I'm also a bit curious if we can edit things as well as create/delete them.

We would want to change the name of the module in the docs stub, but we can take care of that, as @terretta mentions above.

mpdehaan:
Hi!

Thank you very much!

PR #6977, opened 2014-04-14
Adds return information to ec2_asg.

scottanderson42:
Returns the following information for the new/changed autoscaling group:

```
    name
    launch_config
    availability_zones
    cooldown
    default_cooldown
    desired_capacity
    enabled_metrics
    health_check_period
    health_check_type
    instance_id
    launch_config_name
    load_balancers
    max_size
    min_size
    placement_group
    termination_policies
    vpc_zone_identifier
```

### DIFF
--- a/library/cloud/ec2_asg
+++ b/library/cloud/ec2_asg
@@ -22,6 +22,7 @@ description:
   - Can create or delete AWS Autoscaling Groups
   - Works with the ec2_lc module to manage Launch Configurations
 version_added: "1.6"
+requirements: [ "boto" ]
 author: Gareth Rushgrove
 options:
   state:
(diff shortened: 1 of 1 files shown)


mpdehaan:
Most of our modules don't return things in sub-hashes, so can you modify this to return these values top level?   This would make it operate better with register.

Thanks!

scottanderson42:
Done, thanks for the feedback.

mpdehaan:
# Possible Duplicated Efforts

Hi!

It looks like the work from this pull request was duplicated by a recent change in the module.

For some additional information:
- in the source of the module on the devel branch there is a get_properties function that appears to do the same thing as this pull request, though possibly not for a complete set of items
- we are definitely open to additions of new parameters to the get_properties function
- we're aware this addition may have been submitted AFTER yours, for which, we apologize, though I hope the needs are met by this other PR as well.  If not, this is a good topic for the devel list.

We're absolutely always up for discussion.  Since this is a really busy project, we don't always see comments on closed tickets, but want to encourage

PR #6978, opened 2014-04-14
Adds regex example to "Playbooks: Conditionals" doc chapter.

clconway:
It took me a long time to figure out this was possible, so I thought it would be helpful to add it to the main docs...


mpdehaan:
Thanks for this, though this really should be applied to the "filters" section of the docs.  I believe it's there -- if not, please send us a pull request for that section.

Thanks again!

PR #6981, opened 2014-04-14
Fix localhost tasks with complex executables

aliles:
Enable the use of executable commands that use command line options with
the localhost command runner. These commands require parsing out the
base executable from the command string to pass to subprocess.

### DIFF
--- a/lib/ansible/runner/connection_plugins/local.py
+++ b/lib/ansible/runner/connection_plugins/local.py
@@ -53,15 +53,16 @@ class Connection(object):
 
         if not self.runner.sudo or not sudoable:
             if executable:
-                local_cmd = [executable, '-c', cmd]
+                local_cmd = executable.split() + ['-c', cmd]
             else:
                 local_cmd = cmd
         else:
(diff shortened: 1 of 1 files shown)


mpdehaan:
Can you elaborate a bit more about what this patch allows the user to achieve?

Apologies for not following.  Thanks!!!

aliles:
Yes, after reading that commit message on a different day it's not very clear at all what this is fixing.

On some unorthodox systems (we have one of these) it may be necessary to define an `executable` that uses command line flags. For example, forcing bash to be a login shell with an alternate init file. This would be configured in `ansible.cfg` like:

```
[defaults]
executable = /bin/bash --login --init-file /opt/app/bash.rc
```

This works successfully for remote connections where the complete command string is passed over the wire. However, for `localhost` connections, the `subprocess` module is used. This module will fail with an exception about being unable to find the executable command. This is caused by attempting to pass the complete `executable` line without removing the command line flags first.

jimi-c:
Merged, thanks!

PR #6983, opened 2014-04-14
Update wait_for module to handle directories as path parameter.

bibliotechy:
Currently, wait_for module only handles files passed as the path parameter. This pull request allows directories to be passed as the path parameter so that wait_for can check their absence or presence. 

Updates the docs to indicate that directories can be handled.


mpdehaan:
Thanks for this -- can you update the description to indicate the directory support is new in 1.6?

bibliotechy:
Oh yeah, that _would_ be helpful to know.

PR #6984, opened 2014-04-14
Adds search/match examples to "Playbooks > Variables > Jinja2 Filters" d...

clconway:
...oc section.

Following up on https://github.com/ansible/ansible/pull/6978 by moving the docs to a more appropriate place.

### DIFF
--- a/docsite/rst/playbooks_variables.rst
+++ b/docsite/rst/playbooks_variables.rst
@@ -291,6 +291,18 @@ doesn't know it is a boolean value::
    - debug: msg=test
      when: some_string_value | bool
 
+To match strings against a regex, use the "match" or "search" filter::
+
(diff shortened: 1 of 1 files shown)


mpdehaan:
I'm wondering thoughts on using this versus:

```
when: "'bar' in str"
```

With the given examples.

Can these be updated to use a more complicated regex that is obviously a regex?  This may help folks more and make it more obvious.

Thanks!!!

and

```
when: "str == 'bar'"
```

?

clconway:
OK, changed the example to be less trivial.

> and
> 
> when: "str == 'bar'"
> ?

Not sure what you mean here...

mpdehaan:
Yep, this is what I'm looking for, thanks!

PR #6991, opened 2014-04-14
Docker: don't send wrong parameters to client.start()

joshuaconner:
As part of being updated for the 1.10 API, a couple of parameters were added to the kwargs that are passed to the docker.client.start() command that it doesn't accept. This caused the module to error out if it tried to start any Docker containers. This removes those parameters so the module works again.

For reference, here is the method signature for `docker.client.start`:

```
    def start(self, container, binds=None, port_bindings=None, lxc_conf=None,
              publish_all_ports=False, links=None, privileged=False):
```

Note that neither `dns` nor `volumes_from` are kwargs.

### DIFF
--- a/library/cloud/docker
+++ b/library/cloud/docker
@@ -565,6 +565,9 @@ class DockerManager:
             'privileged':   self.module.params.get('privileged'),
             'links': self.links,
         }
+        if docker.utils.compare_version('1.10', self.client.version()['ApiVersion']) >= 0 and hasattr(docker, '__version__') and docker.__version__ > '0.3.0':
+            params['dns'] = self.module.params.get('dns')
+            params['volumes_from'] = self.module.params.get('volumes_from')
 
         for i in containers:
             self.client.start(i['Id'], **params)
--- a/library/cloud/docker
+++ b/library/cloud/docker
@@ -566,10 +566,6 @@ class DockerManager:
             'links': self.links,
         }
 
-        if docker.utils.compare_version('1.10', self.client.version()['ApiVersion']) >= 0:
-            params['dns'] = self.module.params.get('dns')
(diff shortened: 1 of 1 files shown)


mattjbray:
Sorry, I was probably a bit too eager in submitting this: https://github.com/ansible/ansible/pull/6956

In the 1.10 API, passing `dns` and `volumes_from` to create() has no effect - they need to be passed to start().  Unfortunately docker-py hasn't been updated for this yet.  The PR is here: https://github.com/dotcloud/docker-py/pull/200

In the meantime I'm doing `pip install --upgrade git+git://github.com/ticosax/docker-py.git@move-dns-and-volumes-from` to be able to use `dns` and `volumes_from`.

library/cloud/docker
```diff
@@ -566,10 +566,6 @@ class DockerManager:
             'links': self.links,
         }
 
-        if docker.utils.compare_version('1.10', self.client.version()['ApiVersion']) >= 0:
```
mattjbray, line 569: something like `... and hasattr(docker, '__version__') and docker.__version__ > '0.3.0'`?

joshuaconner, line 569: The current version of `docker-py` is 0.3.1, but they didn't update the `__version__` attribute of `__init__.py`. Assuming they update it for the next release (there's an open issue https://github.com/dotcloud/docker-py/issues/201), that should be good. BUT I'm sort of hesitant to do that until we know for sure what `docker-py` is going to do?

joshuaconner:
I put the lines back in with an additional version check as @mattjbray suggested. It works now, and should continue to work whenever `docker-py` releases an update for the 1.10 API.

jimi-c:
Merged, thanks!

PR #6992, opened 2014-04-14
Support hosts without private ip without errors

joefiorini:
Not all Linode hosts have a private ip. This fixes an actual error that was happening because the generated list of private ips is empty when there isn't one.


jimi-c:
Merged, thanks!

PR #6995, opened 2014-04-15
Adding support for spot instances in ansible_lc.

fdv:
AWS allows to define launch configuration with spot instances. This is also natively supported by boto (see http://boto.readthedocs.org/en/latest/ref/autoscale.html)


colstrom:
This would be very useful to have. Curious why it's tagged P5.

tboeghk:
+1 on this one, would be very useful

PR #6996, opened 2014-04-15
module replace - report changed only if contents is changed

orgoj:
Substitution report change even when contents is the same. This patch compare contents for change detection.


jimi-c:
@orgoj it appears that a second commit has snuck in with this, which is not related to the first. Could you please create separate feature branches for each of these?  Thanks!

orgoj:
I make two separate branches https://github.com/ansible/ansible/pull/7013 and https://github.com/ansible/ansible/pull/7012, but they are doing conflict.

PR #7002, opened 2014-04-15
Added FILE_COMMON_DOCSTRINGS for common options.

cr3:
Multiple modules share the same common arguments that are expressed
in code by the add_file_common_args argument in the AnsibleModule
class. The modules that set this argument to True don't always update the
documentation accordingly though. For example, the owner is sometimes
assumed to exist only from the examples. This branch proposes a way to
share documentation across modules by using a FILE_COMMON_DOCSTRINGS
instead of copying and pasting the same documentation across modules.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -119,6 +119,43 @@ FILE_COMMON_ARGUMENTS=dict(
     directory_mode = dict(), # used by copy
 )
 
+FILE_COMMON_DOCSTRINGS=dict(
+    # Prepend option name and append common attributes.
+    (key, '''{}:
+    {}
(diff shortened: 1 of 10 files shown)


mpdehaan:
I'm not in favor of the the implementation here for readability reasons (strings resulting in data structures is a little confusing), but thankfully there's a nicer way to dot his already implemented -- which is why I'd like this to use the same mechanism.

See the "module fragments" work done in commits like this -- https://github.com/ansible/ansible/commit/509561f658c20a4fdb20b2f2bc8fd984459530d2

the same can be applied to the file modules.

I'm going to close this one, though updates to use this mechanism would be quite welcome.

Thanks!!!

Issue #6987, opened 2014-04-14
Deploying EC2 Instance in VPC with Private IP set, get public ip and subnet throws an error


evanccnyc:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.5.3
##### Environment:

Mac OS X
##### Summary:

When I use ansible to setup an ec2 instance that is also in a vpc while specifying a private local ip, getting a public ip and giving it the correct subnet, I get the following output:
msg: InvalidParameterCombination: Network interfaces and an instance-level private IP address may not be specified on the same request.

Based on this site I think ansible isnt using the correct boto call:

http://drewblas.com/2012/08/24/aws-vpc-error-client-invalidparametercombination/
##### Steps To Reproduce:

Here is the ansible code I am using:
- name: Provision EC2 instances - frontend zone 1
  ec2:
  
    region: "{{ aws_region }}"
    keypair: "{{ aws_keypair }}"
    group: ansibleDeployedFrontend
    instance_type: "{{ aws_InstanceType }}"
    image: "{{ aws_image }}"
    vpc_subnet_id: subnet-b27c62c6
    wait: yes
    private_ip: 10.1.10.10
    assign_public_ip: true
    instance_tags:
      type: "frontend"
      Name: "{{ aws_zone1 }}.{{ aws_InstanceType }}.frontend.01"
    volumes:
  - device_name: /dev/sda1
     volume_size: 150
    retries: 5
    register: ec2_instances_zone1
##### Expected Results:

Deploy the instance
##### Actual Results:

msg: InvalidParameterCombination: Network interfaces and an instance-level private IP address may not be specified on the same request.

PR #7003, opened 2014-04-15
Fixing bug where if both private_ip and assign_public_ip in ec2 are set ansible fails out

evanccnyc:
Ansible fails out.

This fixes this bug:

https://github.com/ansible/ansible/issues/6987

I tested all 3 variations after creating this patch: Private_ip alone, assign_public_ip and both set. All work. It also passes the "make tests".

This is my first patch so let me know if I did something wrong.

### DIFF
--- a/library/cloud/ec2
+++ b/library/cloud/ec2
@@ -820,15 +820,13 @@ def create_instances(module, ec2, override_count=None):
                             private_ip_address=private_ip,
                             groups=group_id,
                             associate_public_ip_address=assign_public_ip)
-                        interfaces = boto.ec2.networkinterface.NetworkInterfaceCollection(interface)
-                        params['network_interfaces'] = interfaces
                     else:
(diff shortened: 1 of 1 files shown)


evanccnyc:
Here is a playbook to test against, needs a vpc :

```
- hosts: local
  connection: local
  gather_facts: False
  tasks:
  - name: Provision EC2 instances - frontend zone 1
    ec2:
        aws_access_key: XXXXX
        aws_secret_key: XXXXXX
        region: "us-east-1"
        keypair: "ansible"
        group: ansibleDeployedFrontend
        instance_type: "m3.medium"
        image: "ami-59a4a230"
        vpc_subnet_id: "subnet-XXXXX"
        wait: yes
        private_ip: 10.1.10.10
        assign_public_ip: true

```

jimi-c:
Merged, thanks!

PR #7004, opened 2014-04-15
Added dynamic inventory plugin for Abiquo

danfaizer:
- Dynamic inventory plugin for Abiquo cloud software

Abiquo cloud software ( http://www.abiquo.com ) is used by several and huge service providers with thousands of customers in many countries.

We would like to share this plugin to ease integration with Ansible.


Marc-Morata-Fite:
I've tested it and works like a charm. Good job!

jimi-c:
Merged, thanks!

PR #7011, opened 2014-04-16
Add missing info for the append option for user module

hernandesbsousa:
Both 'default' and 'choices' options were missing at the documentation.


jimi-c:
Merged, thanks!

PR #1864, opened 2013-01-13
Prevent permission denied when cloning

eval:
Without read permission to the current working directory, git-clone will
fail:

```
root@host:~$ sudo -u git -H git clone \
  git://github.com/ansible/ansible.git /home/git/ansible
fatal: Could not change back to '/root': Permission denied
```

With ansible:

```
$ ansible host -u root --sudo-user git -m git -a 'repo=git://github.com/ansible/ansible.git dest=/home/git/ansible
127.0.0.1 | FAILED >> {
    "failed": true, 
    "msg": "fatal: Could not change back to '/root': Permission denied\n"
}
```

This commit ensures that the working directory is changed to the
parent-folder of the destination before doing the clone.

PR #7014, opened 2014-04-16
prevent permission denied failure

yteraoka:
avoid accessing login user's home directory
#1864
# issuecomment-40570064

### DIFF
--- a/library/source_control/git
+++ b/library/source_control/git
@@ -524,6 +524,7 @@ def main():
     before = None
     local_mods = False
     if not os.path.exists(gitconfig):
+        os.chdir('/') # prevent permission denied (avoid accessing login user's home directory)
         if module.check_mode:
             remote_head = get_remote_head(git_path, module, dest, version, repo, bare)
             module.exit_json(changed=True, before=before, after=remote_head)


jimi-c:
I believe we have fixed this in `module_utils/basic.py`, where `run_command()` will now chdir appropriately. What version are you running this on, as the issue you reference is extremely old?

yteraoka:
hmm..
This issue occurred with Ansible 1.5.4.

$ ansible --version
ansible 1.5.4

yteraoka:
oh,

in module_utils/basic.py

```
        try:
            # make sure we're in the right working directory
            if cwd and os.path.isdir(cwd):
                os.chdir(cwd)
```

When destination directory does not exist (first time), this issue occurs.

jimi-c:
Ok thanks, we'll look into it.

jimi-c:
Looking back at this, and I'm unclear where the problem would be. The clone() call should create the directory and then uses that as the cwd= argument to run_command(), so I'm unclear how this would still be failing. Regardless, a better solution would probably be to check the os.access() rights to the cwd rather than doing a blanket switch to `/`. I'm going to go ahead and close this for now, if you'd like to resubmit it based on what I said above feel free. Thanks!

PR #7020, opened 2014-04-16
apt-key list only returns keys in uppercase

jjshoe:
More user friendly if someone is silly enough to enter a lower case key.

Issue #7018, opened 2014-04-16
apt_key fails if key id is too long


berle:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.5.4 (Also applies to head)
##### Environment:

Ubuntu 12.04
##### Summary:

apt_key will fail to find a key in the keychain if the id specified is too long, due to apt-key list not printing keys with their full fingerprint.
##### Steps To Reproduce:

```

---
- name: Add MariaDB key
  apt_key: id="0xcbcb082a1bb943db" url="http://keyserver.ubuntu.com/pks/lookup?op=get&search=0xcbcb082a1bb943db"
```
##### Expected Results:

Key added and verified.
##### Actual Results:

```
msg: key does not seem to have been added
```
##### Suggested Fix:

Perhaps a better way to check for the presence of the key would be apt-key export, which supports using up to the full fingerprint?

jjshoe:
Unable to reproduce when passing in the correct id:

```
---
- name: Add MariaDB key
  sudo: yes
  apt_key: id="1BB943DB" url="http://keyserver.ubuntu.com/pks/lookup?op=get&search=0xcbcb082a1bb943db"
```

jjshoe:
https://github.com/ansible/ansible/pull/7022

berle:
How is that the "correct id" when the apt-key utility itself happily accepts the longer version? If the apt_key module requires a very specific key format, it should give an error if the user does not provide this format. It's not a user-error, nor is it intuitively wrong in any way to specify the key id in a format that apt-key actually allows.

jjshoe:
apt-key list shows 8 digit, not 16 digit. It's possible to skip apt-key, and go straight to gpg, then do logic based on 8 digit vs 16 digit keys, but that would require a large re-write when the 8 digit key will do.

berle:
Yes, I've already explained that in the initial bug report, so there's no need to explain it again. I also suggested how to work around the problem in a way that don't even require making any significant changes, still using apt-key, so I'm not really sure why you're suggesting the only way to fix this is to use gpg directly. Sigh...

andreasf:
I just ran into the same issue. Since only the last 8 digits seem to be interesting, how about:

```
diff --git a/library/packaging/apt_key b/library/packaging/apt_key
index 2308d34..77bbac4 100644
--- a/library/packaging/apt_key
+++ b/library/packaging/apt_key
@@ -210,6 +210,7 @@ def main():
[...]

PR #7022, opened 2014-04-16
apt_key lists all keys in uppercase

jjshoe:
allow for a little user error

### DIFF
--- a/library/packaging/apt_key
+++ b/library/packaging/apt_key
@@ -210,6 +210,7 @@ def main():
             _ = int(key_id, 16)
             if key_id.startswith('0x'):
                 key_id = key_id[2:]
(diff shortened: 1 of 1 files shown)


jimi-c:
Just for future reference, you've created a feature branch for this, so you could just add your fix using `git commit --amend` on that branch and then `git push --force` to overwrite the previous commit on github. That will automatically update the PR without having to create a second one.

I'm assuming #7020 can be closed?

jjshoe:
Yes, it can be, I was trying to update the PR but made a giant mess.

andreasf:
Please also address #7018, as the fix for that would be in the same line

jimi-c:
Merged, thanks!

PR #7025, opened 2014-04-16
Add Docker inventory plugin

angstwad:
This adds more Docker to Ansible.  It also fixes the feedback/issues from mpdehaan that we discusses via the devel mailing list.

Thanks, guys.

### DIFF
--- a/plugins/inventory/docker.py
+++ b/plugins/inventory/docker.py
@@ -1,0 +1,359 @@
+#!/usr/bin/env python
+
+# (c) 2013, Paul Durivage <paul.durivage@gmail.com>
+#
+# This file is part of Ansible.
+#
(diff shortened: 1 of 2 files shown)


mpdehaan:
Hi Paul,

I recall there being a previous ticket about this but I can't find it -- can you link the number here so I can see any reply to comments?  I recall having a question about multiple host support that I was unsure about.

Thanks!!!

mpdehaan:
nm ... I do see multiple hosts I believe, nice!

angstwad:
Yes sir, multiple host support is there.

angstwad:
There wasn't a ticket, it was an email on the devel list.

jimi-c:
Merged, thanks!

PR #7026, opened 2014-04-16
Fix race conditions in monit module, and general monit cleanup

brian-brazil:
Environment: recent head on Ubuntu 12.04

The monit module has a number of race conditions that'll trigger if monit gets into the correct state too quickly. For example, if you set state=started and by the time the summary command is run and it's already in state 'Running'. This is causing failures on about 50% of my ansible runs that touch this code.

I also took care of the fixme by refactoring the status checking code, and exit codes are now also checked.
I've manually tested all code paths.

As a general module style question, should this module be checking that it worked or should it adopt a more fire and forget approach? There seems to be quite a bit of complexity added by this in this module, and I'm wondering if removing it would be better overall.

### DIFF
--- a/library/monitoring/monit
+++ b/library/monitoring/monit
@@ -47,8 +47,6 @@ EXAMPLES = '''
 - monit: name=httpd state=started
 '''
 
-import pipes
-
 def main():
     arg_spec = dict(
         name=dict(required=True),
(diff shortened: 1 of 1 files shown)


jimi-c:
Overall, I would say all of the fail_json calls you're making need to have better error messages, as they are pretty much just reprinting the status as returned by the command but no cause is given. Pretty much a minor fix, if you get that changed I'll get this merged in ASAP. Thanks!

jimi-c:
And one other thing I just noticed: if status() returns None, your use of `"whatever" in status()` is going to cause a TypeError.

brian-brazil:
Thanks for the review, I think I've addressed your comments. I don't think I can do much better on the error messages without a lot more complication unfortunately, this should be slightly more useful than what's already there.

jimi-c:
Merged, thanks!

PR #7031, opened 2014-04-17
apt_key: fix key lookup for 64 bit key ids

### LINKED ISSUE #7018: apt_key fails if key id is too long

andreasf:
Fixes #7018 by extending `all_keys` to work with either short or long keys.

Also removes the ununsed method `key_present`.

### DIFF
--- a/library/packaging/apt_key
+++ b/library/packaging/apt_key
@@ -120,11 +120,11 @@ def check_missing_binaries(module):
     if len(missing):
         module.fail_json(msg="binaries are missing", names=missing)
 
-def all_keys(module, keyring):
+def all_keys(module, keyring, short_format):
     if keyring:
-        cmd = "apt-key --keyring %s list" % keyring
+        cmd = "apt-key --keyring %s adv --list-public-keys --keyid-format=long" % keyring
     else:
-        cmd = "apt-key list"
+        cmd = "apt-key adv --list-public-keys --keyid-format=long"
     (rc, out, err) = module.run_command(cmd)
     results = []
     lines = out.split('\n')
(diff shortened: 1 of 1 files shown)


jimi-c:
Why take the key characters from the end of the string? I don't think that makes things any less likely to collide, and seems counter-intuitive?

andreasf:
It needs to work with both short and long keys: long keys make collisions less likely, and short keys are required for backwards compatibility. Both short and long key ids are derived from the fingerprint, which is the SHA-1 of the signing certificate. The short key id consists of the last 8 chars/32 bits, and the long key id of the last 16 chars/64 bits of the fingerprint (see `man gpg` for reference). If a short key id is supplied, I convert the long key ids I get from apt-key to the short format.

First, I check if the key id is short or long:

```
short_format = (key_id is not None and len(key_id) == 8)
```

Then, I _could_ call apt-key with parameters depending on `short_format`. Instead, I opted to make the conversion in Python, because it's so simple. Depending on the value of `short_format`, my version of `all_keys()` returns all key ids either in short or long format.

jimi-c:
Oh ok, I didn't realize that was the apt-key convention, I'm used to git which uses the first 7+ characters of the SHA.

jimi-c:
Merged, thanks!

PR #7035, opened 2014-04-17
Add pre_tasks and post_tasks for roles.

benosman:
This is something i've wanted for a while; the ability to have pre_tasks and post_tasks for each role.

My primary use case is apt (or yum) sources, to avoid unnecessarily calling apt-get update many times.

The idea would be that in the pre_tasks of a role, all the keys and repos would be set - if there are changes then apt-update is run once, otherwise it is not run. This would help me in re-running the playbook periodically for configuration updates with out the delay each time.

I tried to keep the changes as simple and as consistent as possible with that is already there, even though this area of code could probably do with some refactoring to reduce duplication.

### DIFF
--- a/lib/ansible/playbook/play.py
+++ b/lib/ansible/playbook/play.py
@@ -350,19 +350,12 @@ class Play(object):
         if type(roles) != list:
             raise errors.AnsibleError("value of 'roles:' must be a list")
 
+        new_pre_tasks = []
         new_tasks = []
+        new_post_tasks = []
         new_handlers = []
         new_vars_files = []
         defaults_files = []
-
-        pre_tasks = ds.get('pre_tasks', None)
-        if type(pre_tasks) != list:
-            pre_tasks = []
-        for x in pre_tasks:
-            new_tasks.append(x)
-
-        # flush handlers after pre_tasks
-        new_tasks.append(dict(meta='flush_handlers'))
 
         roles = self._build_role_dependencies(roles, [], self.vars)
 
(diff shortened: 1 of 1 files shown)


mpdehaan:
Hmm, this could be interesting for having a role like "monitoring_window_wrapped" or "load_balanced".

I like the idea.

zellski:
@mpdehaan - does this feature still have the blessing of the powers that be, do you think? If someone (possibly even me) fixed the pull request from Apr 17 2014, would it be merged?

I'm forced to move away from using roles as often as I'd like, because I have to jump through so many ugly hoops to do setup before meta/main.yml dependencies are evaluated.

akkie:
I would also be interested in this feature :+1:

amenonsen:
I notice that devel has pre_tasks and post_tasks, documented in docsite/rst/guide_rolling_upgrade.rst:

```
The ``pre_tasks`` keyword just lets you list tasks to run before the roles are called.
```

I can't quite tell if that's the same thing that this PR intends (though I think it is), because there's no documentation at all included in the PR. Anyway, the code in devel is significantly different from the code at the time the PR was created, it's not just a matter of resolving a merge conflict. But I think v2 already does what was intended.

I think this PR should be closed now.

gregdek:
Seems like v2 solves this, and even if it doesn't, a major rewrite of this patch would now be necessary in any case.

@benosman thanks for the PR, but we're closing now. If you think v2 doesn't solve your case here, happy to have you submit a new PR.

PR #7039, opened 2014-04-17
Add support for a tag that lets a task be always run.

### LINKED ISSUE #3157: Add a special 'all' tag that always matches
The problem is that once you have a playbook that uses `group_by`, or possibly some other modules that define hosts or groups, you can't use `--tags` on the command line, because there is no way you can get the `group_by` play match (as far as I can see, maybe I'm wrong).

Let's say I have a playbook like this:

```
- hosts: all
  tasks:
    - name: Create a group of all hosts by operating system
      action: group_by key={{ansible_distribution}}

- hosts: Ubuntu
  tasks:
    - name: Install nginx
      action: apt name=nginx state=installed
      tags: nginx

- hosts: CentOS
  tasks:
    - name: Install nginx
      action: yum name=nginx state=installed
      tags: nginx
```

If I run this command, nothing actually gets executed, because the first play does not match the `nginx` tag:

```
ansible-playbook -i hosts --tags nginx
```

I'd suggest introducing a special syntax `tags: all` that would match any tag specified on the command line:

```
- hosts: all
  tasks:
    - name: Create a group of all hosts by operating system
      action: group_by key={{ansible_distribution}}
      tags: all
```

The only problem is that this is a backward-incompatible change, because people already might be using `all` as a regular tag.

brian-brazil:
Fixes #3157

'all' already has special meaning, so I went with the next suggestion of *. That requires quoting, so I'm open to other options if you think that'll be a problem.

This change also adds tests for this and basic tag functionality.

### DIFF
--- a/bin/ansible-playbook
+++ b/bin/ansible-playbook
@@ -77,7 +77,7 @@ def main(args):
     #    help="password for vault encrypted files")
     parser.add_option('-e', '--extra-vars', dest="extra_vars", action="append",
         help="set additional variables as key=value or YAML/JSON", default=[])
-    parser.add_option('-t', '--tags', dest='tags', default='all',
+    parser.add_option('-t', '--tags', dest='tags',
         help="only run plays and tasks tagged with these values")
     parser.add_option('--skip-tags', dest='skip_tags',
         help="only run plays and tasks whose tags do not match these values")
@@ -146,7 +146,9 @@ def main(args):
             # Arguments as Key-value
             extra_vars = utils.combine_vars(extra_vars, utils.parse_kv(extra_vars_opt))
 
-    only_tags = options.tags.split(",")
+    only_tags = options.tags
+    if options.tags is not None:
+        only_tags = options.tags.split(",")
     skip_tags = options.skip_tags
     if options.skip_tags is not None:
         skip_tags = options.skip_tags.split(",")
@@ -210,28 +212,10 @@ def main(args):
             print 'playbook: %s' % playbook
             print ''
             playnum = 0
-            for (play_ds, play_basedir) in zip(pb.playbook, pb.play_basedirs):
+            for playnum, play in enumerate(pb.plays_to_run()):
                 playnum += 1
-                play = ansible.playbook.Play(pb, play_ds, play_basedir,
-                                              vault_password=pb.vault_password)
                 label = play.name
                 hosts = pb.inventory.list_hosts(play.hosts)
-
-                # Filter all tasks by given tags
-                if pb.only_tags != 'all':
-                    if options.subset and not hosts:
-                        continue
-                    matched_tags, unmatched_tags = play.compare_tags(pb.only_tags)
-
-                    # Remove skipped tasks
-                    matched_tags = matched_tags - set(pb.skip_tags)
-
-                    unmatched_tags.discard('all')
-                    unknown_tags = ((set(pb.only_tags) | set(pb.skip_tags)) -
-                                    (matched_tags | unmatched_tags))
-
-                    if unknown_tags:
-                        continue
 
                 if options.listhosts:
                     print '  play #%d (%s): host count=%d' % (playnum, label, len(hosts))
@@ -240,15 +224,11 @@ def main(args):
 
                 if options.listtasks:
                     print '  play #%d (%s):' % (playnum, label)
-
-                    for task in play.tasks():
-                        if (set(task.tags).intersection(pb.only_tags) and not
-                            set(task.tags).intersection(pb.skip_tags)):
-                            if getattr(task, 'name', None) is not None:
-                                # meta tasks have no names
-                                print '    %s' % task.name
-                if options.listhosts or options.listtasks:
-                    print ''
+                    for task in pb.tasks_to_run_in_play(play):
+                        if getattr(task, 'name', None) is not None:
+                            # meta tasks have no names
+                            print '    %s' % task.name
+                print ''
             continue
 
         if options.syntax:
--- a/docsite/rst/playbooks_tags.rst
+++ b/docsite/rst/playbooks_tags.rst
@@ -40,6 +40,10 @@ And you may also tag basic include statements::
 
 Both of these have the function of tagging every single task inside the include statement.
 
+New in version 1.7, the tag "all"  will cause the task to run even if not specified in ``--tags``, or
+before ``--start-at-task``. ``--skip-tags`` still applies.  This is useful for 
+modules such as "group_by" and "ec2_facts".
+
 .. seealso::
 
    :doc:`playbooks`
--- a/lib/ansible/callbacks.py
+++ b/lib/ansible/callbacks.py
@@ -619,7 +619,8 @@ class PlaybookCallbacks(object):
                 if actual_name == self.start_at or fnmatch.fnmatch(actual_name, self.start_at):
                     del self.start_at
 
-        if hasattr(self, 'start_at'): # we still have start_at so skip the task
+        if hasattr(self, 'start_at') and 'all' not in self.task.tags:
+            # we still have start_at so skip the task
             self.skip_task = True
         elif hasattr(self, 'step') and self.step:
             msg = ('Perform task: %s (y/n/c): ' % name).encode(sys.stdout.encoding)
--- a/lib/ansible/playbook/__init__.py
+++ b/lib/ansible/playbook/__init__.py
@@ -122,7 +122,7 @@ class PlayBook(object):
         if extra_vars is None:
             extra_vars = {}
         if only_tags is None:
-            only_tags = [ 'all' ]
+            only_tags = []
         if skip_tags is None:
             skip_tags = []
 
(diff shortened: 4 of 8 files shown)


brian-brazil:
Using this in the wild indicated that this doesn't interact in the way you'd want with --start-at-task, namely that if you've something like a group_by you always want it to run so that whichever later tasks are running have the groups/variables/inventory they need. Pull request and tests updated accordingly.

mpdehaan:
In queue for review, though I should note, it will need to be "all".

mpdehaan:
Every task gets tagged with all, but this could be internally implemented such that if the tag is "all" it gets tagged with something else the software looks for instead, i.e. the interface and the implementation can probably be kept seperate, or the existing "all" references in the code might be changeable, since they are not publically surfaced.

brian-brazil:
I've refactored and updated the code to use 'all', with the existing functionality of 'all' no longer depending on a tag.

This also cleans up some edge cases, making --list-tasks more correct.

LeonB:
Any ETA on when this is going to be merged?

jaimegago:
+1 on this.
FTR my use case is a safe guard "pre_tasks" aimed at preventing running a playbook in the wrong AWS account with AWS keys loaded in environment. That worked until I decided to limit the scope of a run with --tags QQ.

Also I got slightly confused with the "always_run" option, when I first started using it I did it right (i.e. for a tasks I needed to run even in "dry mode") but then I had a mental switch and started assigning "always_run" to tasks I, well, always wanted to run...My point being since this PR is not about extending the "always_run" to also work when specifying tags, then IMHO the "always_run" should be renamed to something that contains a reminder to its specific "run even in dry run" use case (e.g. override_dry_run_and_execute, or something even longer =P)

arnuschky:
I'd love to see this in ansible. It's a functionality I quite naturally expected to be there, especially using the 'all' keyword

aabdnn:
I was looking for this same functionality, and I'm happy to see a pull request. Please pull it as soon as you can :)

jgornick:
If I understand correctly, this PR provides the ability to tag specific resources with a "all" tag and when a resource is tagged with "all" it will always run regardless of your `--tags` and `--skip-tags` options. 

It seems a little unclear to me why you'd want that specific functionality. I say that because with this PR you are required to tag a resource with a tag, but why implement specific functionality for an "all" tag where the user should just use the `--tags` and `--skip-tags` options.

For example, let's say instead of overriding the existing "all" tag, let's use a custom tag of "always":

```
---
- name: verify tags work as expected
  hosts: local
  gather_facts: False
  connection: local
  tasks:
    - name: Task_with_tag
      debug: msg=
      tags: tag
    - name: Task_with_always_tag
      debug: msg=
      tags: always
    - name: Task_without_tag
      debug: msg=
```

Then, when running the playbook if you want to run the "always" task, you include it with the `--tags` option: `--tags "always,tag"`.

brian-brazil:
@jgornick #3157 #9010 and #9020 go into why this would be very useful. 

This PR obeys --skip-tags, as I figured there'd be edge cases where you didn't want this behaviour (mostly around disaster recovery) and none of the use cases indicated otherwise.

This PR currently has merge conflicts due to drift in the codebase in the 4 months since I last had feedback from the committers. If a commiter can let me know when they'll have a chance to look at it, I'll clean it up.

jgornick:
@brian-brazil Yeah, #9010 is my issue, but this isn't the functionality I was looking for. It doesn't appear that I described my proposal with enough detail, but let me see if I can summarize here:

In my proposal, every resource would automatically be tagged with "all", as it does right now. However, when a tag is specified on a resource, the automatic "all" tag is removed from the resource. This approach **overrides** the default tag assigned to the resource with the specified tag. Since the "all" tag is just a tag, you could still include the "all" tag in your overridden set of tags.

For example, if I have a set of resources tagged as:

```
  tasks:
    - name: "Task (default all)"
      command: "echo 'task:default-all'"

    - name: "Task (tagged all)"
      command: "echo 'task:tagged-all'"
      tags:
        - all

    - name: "Task (foo)"
      command: "echo 'task:foo'"
      tags:
        - foo

    - name: "Task (all, foo)"
      command: "echo 'task:all,foo'"
      tags:
        - all
        - foo

    - name: "Task (bar)"
      command: "echo 'task:bar'"
      tags:
        - bar

    - name: "Task (all, bar)"
      command: "echo 'task:all,bar'"
      tags:
        - all
        - bar

    - name: "Task (foo, bar)"
      command: "echo 'task:foo,bar'"
      tags:
        - foo
        - bar
```

The following is a list of acceptance criteria I would expect based the implementation I'm proposing:

**Command:** 

`ansible-playbook -i "127.0.0.1," -c local tag-test.yml`

**Expected Results:**

```
PLAY [Test Play for All Tags] ************************************************* 

TASK: [Task (default all)] **************************************************** 
changed: [127.0.0.1]

TASK: [Task (tagged all)] ***************************************************** 
changed: [127.0.0.1]

TASK: [Task (foo)] ************************************************************ 
changed: [127.0.0.1]

TASK: [Task (all, foo)] ******************************************************* 
changed: [127.0.0.1]

TASK: [Task (bar)] ************************************************************ 
changed: [127.0.0.1]

TASK: [Task (all, bar)] ******************************************************* 
changed: [127.0.0.1]

TASK: [Task (foo, bar)] ******************************************************* 
changed: [127.0.0.1]
```

**Command:** 

`ansible-playbook -i "127.0.0.1," -c local --tags="all" tag-test.yml`

**Expected Results:**

```
PLAY [Test Play for All Tags] ************************************************* 

TASK: [Task (default all)] **************************************************** 
changed: [127.0.0.1]

TASK: [Task (tagged all)] ***************************************************** 
changed: [127.0.0.1]

TASK: [Task (all, foo)] ******************************************************* 
changed: [127.0.0.1]

TASK: [Task (all, bar)] ******************************************************* 
changed: [127.0.0.1] 
```

**Command:** 

`ansible-playbook -i "127.0.0.1," -c local --tags="foo" tag-test.yml`

**Expected Results:**

```
PLAY [Test Play for All Tags] ************************************************* 

TASK: [Task (foo)] ************************************************************ 
changed: [127.0.0.1]

TASK: [Task (all, foo)] ******************************************************* 
changed: [127.0.0.1]

TASK: [Task (foo, bar)] ******************************************************* 
changed: [127.0.0.1]
```

**Command:** 

`ansible-playbook -i "127.0.0.1," -c local --tags="all,foo" tag-test.yml`

**Expected Results:**

```
PLAY [Test Play for All Tags] ************************************************* 

TASK: [Task (default all)] **************************************************** 
changed: [127.0.0.1]

TASK: [Task (tagged all)] ***************************************************** 
changed: [127.0.0.1]

TASK: [Task (foo)] ************************************************************ 
changed: [127.0.0.1]

TASK: [Task (all, foo)] ******************************************************* 
changed: [127.0.0.1]

TASK: [Task (all, bar)] ******************************************************* 
changed: [127.0.0.1]

TASK: [Task (foo, bar)] ******************************************************* 
changed: [127.0.0.1]
```

**Command:** 

`ansible-playbook -i "127.0.0.1," -c local --tags="foo" --skip-tags="all" tag-test.yml`

**Expected Results:**

```
PLAY [Test Play for All Tags] ************************************************* 

TASK: [Task (foo)] ************************************************************ 
changed: [127.0.0.1]

TASK: [Task (foo, bar)] ******************************************************* 
changed: [127.0.0.1]
```

**Command:** 

`ansible-playbook -i "127.0.0.1," -c local --tags="all" --skip-tags="foo" tag-test.yml`

**Expected Results:**

```
PLAY [Test Play for All Tags] ************************************************* 

TASK: [Task (default all)] **************************************************** 
changed: [127.0.0.1]

TASK: [Task (tagged all)] ***************************************************** 
changed: [127.0.0.1]

TASK: [Task (all, bar)] ******************************************************* 
changed: [127.0.0.1]
```

This approach plays well with the current implementation of tagging. It also supports the scenarios described here and in the other issues. The only difference is you probably wouldn't tag a resource that you always want to run with "all", but rather use a tag like "always" and include it in your list of tags to include. Some may also argue, "I don't want to have to specify the 'always' tag every time I run my playbook!". I'd write a wrapper around ansible-playbook that modifies the tags argument and adds the "always" tag.

brian-brazil:
@jgornick Ah, that's a different use case then. I don't think your proposal supports the use case this PR is aiming for as the goal is not to have to remember to specify --tag=always every time, so I'd suggest getting #9010 reopened to avoid confusion.

jaimegago:
@brian-brazil @jgornick  you might want to consider our benevolent dictator for life @mpdehaan latest answer on this topic.

https://groups.google.com/forum/?hl=fr#!searchin/ansible-project/how$20to$20always$20run$20code$2C$20regardless$20of$20tag/ansible-project/-tVlPzRGavA/r6TBAUYMTKwJ

jgornick:
@brian-brazil Sounds good! Will do.

Not that my opinion matters much, but I still have a couple of issues with this proposal. 

By implementing logic in `ansible-playbook` to always execute resources tagged as "all" is confusing. It's confusing because, currently, all resources are tagged with "all". I'd much rather see the tag be "always" or something that doesn't conflict with an already existing tag. This new tag can then be turned into a "reserved tag" similar to the "all" tag. With all that being said, I'm still not a fan of this implementation.

I feel that `ansible-playbook` should not **assume** what tagged resources I want to execute. I should have to explicitly specify them.

Like I said, if a user doesn't want to include the tag to always run in the `--tags` option, then they can write a wrapper script.

jgornick:
I wrote more words about this feature: https://github.com/ansible/ansible/issues/9010#issuecomment-56470105

I'll stop now :smile:

rocco-muscaritolo:
My 2cents,

I have a production playbook that uses 6 different tags based on roles. Each role is given a tag for both testing and production use cases.

I've now switched to a best practices method of using "group_by" to set system variables.

---
- hosts: all
  user: root
  
  tasks:
  - group_by: key={{ ansible_os_family }}
    tags: [ 'ius-repos', 'apache2', 'php5', 'mysql', 'phpmyadmin', 'holland' ]

Since this pre-task must _always_ be run I have to manually put in each tag I create. And its only going to get uglier as time goes on.

An "all_tags" or "always_tagged" option is a good feature and IMO should have been added already.

brian-brazil:
I've resolved the conflicts, this PR will once again merge cleanly.

redbaron:
Is there any way we can help to speedup merging of this PR? Testing, documenting, anything. It's emarassing that such a nice feature is lingering for so long.

bobbyno:
+1

Just ran into a need for this when including a common playbook that doesn't require tags into another playbook that does.

tkinz27:
Out of curiosity, any chance this pull request would make it so tasks tagged with the "all" or whatever keyword was decided on would also run during a --start-at-task where the task to start at is after the set of tasks that would be tagged with "all"?

brian-brazil:
@tkinz27 I made that change back in May, see the second comment on the thread.

bcoca:
see if this covers all the use cases #10183

Signum:
@bcoca: Good. Looking forward to it. Thanks.

bcoca:
@brian-brazil copied your tests (I hope you don't mind) and copied your use of a method to reuse task selection in ansible-playbook

I also changes 'always' tag to be skippable

I'm not adding star-at-task functionality nor tag pattern matching in this round, leaving those for future PRs

brian-brazil:
#10183 was merged and contained most of this PR, so this is now obsolete.

PR #7044, opened 2014-04-17
hint to install python-ssl on redhat systems

42wim:
Give the user some hints about how to fix the SSL validation error. By installing python-ssl from EPEL on RHEL5.x systems.


mpdehaan:
merged, thank you!  very good upgrade to this error message!

PR #7046, opened 2014-04-17
docker: correctly compare 'command' param against existing container command'

joshuaconner:
If a container has an "entrypoint" in addition to a "command", the Docker API will return "{} {}".format(entrypoint, command) for the value of 'Command' in the container summary JSON. This checks if a container's 'Command' value **ends with** module.params('command') instead of equals it so that containers with entrypoints are correctly classified as match the module's argument spec. (I think this is what caused #6395, and should fix it)

### DIFF
--- a/library/cloud/docker
+++ b/library/cloud/docker
@@ -489,8 +489,7 @@ class DockerManager:
         return inspect
 
     def get_deployed_containers(self):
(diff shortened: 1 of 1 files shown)


joshuaconner:
whoops, a stray change made it in here, going to close and make a new PR

PR #7047, opened 2014-04-17
docker: exclude 'entrypoint' from comparing 'command' param with containers

joshuaconner:
The JSON the Docker API returns includes the container's entrypoint (if it has one) prepended to its `Command` value. So instead of checking if `container['Command'] == module.params['command']` – which will be wrong if the container has an entrypoint – we just check that `container['Command'].endswith(module.params['command'])` so the entrypoint won't affect a container being properly classified as matching the module params or not.

Also I refactored a super-long `if` statement into some temporary variables - I did it to help me figure out what was going wrong, and then it makes the code more readable so I kept it.

(Note that this won't work with the currently-released `docker-py` until #6991 is merged in, because the `docker` module is broken until that happens)

### DIFF
--- a/library/cloud/docker
+++ b/library/cloud/docker
@@ -489,8 +489,7 @@ class DockerManager:
         return inspect
 
     def get_deployed_containers(self):
-        # determine which images/commands are running already
-        containers = self.client.containers(all=True)
+        """determine which images/commands are running already"""
         image = self.module.params.get('image')
         command = self.module.params.get('command')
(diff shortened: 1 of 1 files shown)


joshuaconner:
Without knowing if the container mentioned in #6395 had an entrypoint I can't say for sure, but I suspect this will fix that as well.

jimi-c:
Merged, thanks!

PR #7052, opened 2014-04-18
module docker_image - fix default tag latest

orgoj:
Fix constant rebuild of image if no tag specified.


jimi-c:
Merged, thanks!

PR #7059, opened 2014-04-18
ec2_eip integration tests.

jsmartin:
First attempt at an integration test.  Comments welcome!

### DIFF
--- a/test/integration/Makefile
+++ b/test/integration/Makefile
@@ -48,7 +48,7 @@ $(CREDENTIALS_FILE):
 	@exit 1
 
 amazon: $(CREDENTIALS_FILE)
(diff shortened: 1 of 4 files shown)


jlaska:
Thanks for the changes!  I suggest the following adjustments ...
- Don't `filter` for only `running` instances, other instance states support terminate
- Make the eip_log file more general purpose.  I wager this will be something we'll want to re-use for future ec2 modules.  I'd suggest adding optparse support for the eip_log file (with an appropriate default).  Additionally, it's not a log, perhaps this should be a YAML data file instead? 
- Don't store the eip_log in `/tmp/...`   Perhaps use `{{output_dir}}` for this?  Or maybe a custom variable is needed for caching cloud resources?

jsmartin:
For the EIP log, I wanted something I can append/stream to quickly -- YAML doesn't really seem to be the right tool for that.  It is a log--it's a log of EIPs that were created during the test.  I'll fix the filter and the output_dir stuff.

jimi-c:
Merged, thanks!

PR #7063, opened 2014-04-18
Fix what looks like a bad copy/pasta

jjshoe:


jjshoe:
Going to review all places this is used.

PR #7067, opened 2014-04-19
Update validate to match copy module

jjshoe:


jjshoe:
messed up indentation.

PR #7070, opened 2014-04-19
Allow %s to be optional in validate

jjshoe:
In the case where some external program doesn't need the filename passed in.

### DIFF
--- a/library/files/copy
+++ b/library/files/copy
@@ -204,7 +204,10 @@ def main():
                 os.unlink(dest)
                 open(dest, 'w').close()
             if validate:
-                (rc,out,err) = module.run_command(validate % src)
+                if "%s" in validate:
+                    (rc,out,err) = module.run_command(validate % src)
+                else:
+                    (rc,out,err) = module.run_command(validate)
                 if rc != 0:
                     module.fail_json(msg="failed to validate: rc:%s error:%s" % (rc,err))
             module.atomic_move(src, dest)
--- a/library/files/lineinfile
+++ b/library/files/lineinfile
@@ -154,7 +154,10 @@ def write_changes(module,lines,dest):
     validate = module.params.get('validate', None)
     valid = not validate
     if validate:
-        (rc, out, err) = module.run_command(validate % tmpfile)
+        if "%s" in validate:
+            (rc,out,err) = module.run_command(validate % src)
+        else:
+            (rc,out,err) = module.run_command(validate)
         valid = rc == 0
         if rc != 0:
             module.fail_json(msg='failed to validate: '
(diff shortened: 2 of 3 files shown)


jjshoe:
Whoops, was looking at validate wrong, nukey nukey.

php-coder:
In 1.6.1 I got the following trace:

```
  File "/home/vagrant/.ansible/tmp/ansible-tmp-1400701232.68-201098736271552/copy", line 1401, in <module>
    main()
  File "/home/vagrant/.ansible/tmp/ansible-tmp-1400701232.68-201098736271552/copy", line 207, in main
    (rc,out,err) = module.run_command(validate % src)
TypeError: not all arguments converted during string formatting
```

Looks like it was broken or changed intentionally in commit a34478d6f38324346da543523a5c844cb394f1eb.

Could someone explain what happened?
Thanks in advance!

jjshoe:
@php-coder can you paste the failing task?

php-coder:
@jjshoe yes, sure:

```
- name: don't show version on error page
  copy: >
    dest=/etc/nginx/conf.d/server_tokens.conf
    content="server_tokens off;\n"
    validate="nginx -t"
  notify: reload nginx
```

jjshoe:
You MUST specify %s. I'm not sure why the error didn't print out though.

jjshoe:
Open an issue, I'll take a peak at fixing.

php-coder:
But why this restriction exists? I don't understand why this is mandatory. In my example I don't need it.

(For a now I used the following workaround: `validate="echo '%s' >/dev/null; nginx -t"`

bcoca:
that does not do what you think it does. validate runs on the temporary file BEFORE it puts it into place, %s has the location of the temp file. Your workaround checks that the previous configuration was correct, not the new one you are installing.

this is done to prevent pushing into place a bad configuration.

php-coder:
@bcoca Thanks for explanation, now I fully understand a problem.

@jjshoe I've just created #7483 Thank you!

jjshoe:
@php-coder pass in nginx -t -c %s

php-coder:
@jjshoe Yes, thank you, I will do that.

php-coder:
@jjshoe BTW, unfortunately it doesn't work:

```
msg: failed to validate: rc:1 error:nginx: [emerg] "server_tokens" directive is not allowed here in /home/vagrant/.ansible/tmp/ansible-tmp-1400703853.53-42334331702762/source:1
nginx: configuration file /home/vagrant/.ansible/tmp/ansible-tmp-1400703853.53-42334331702762/source test failed
```

`-c` option must points to full nginx configuration file, not a part of it (as in my case). I don't know how to do what I need, probably there is no way.

bcoca:
Confirming, validate does not work for multiparty configs. In my case nginx
init script does full config check on restart ( before stopping )

Brian Coca

PR #7076, opened 2014-04-19
Add a better error message for service

mscherer:
While migrating my playbook to a newer ansible version, I faced
the error message "unknown init system, cannot enable service". It turned
out to be caused by a wrong service name that was not expanded anymore.

So by giving the name of the service that cannot be enabled and a more precise
reason, i think people will be able to diagnose their issue more easily.

### DIFF
--- a/library/system/service
+++ b/library/system/service
@@ -565,7 +565,7 @@ class LinuxService(Service):
     def service_enable(self):
 
         if self.enable_cmd is None:
-            self.module.fail_json(msg='unknown init system, cannot enable service')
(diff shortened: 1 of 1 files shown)


jimi-c:
Could you come up with a combination of those two messages? It seems that it could be one or the other, and this had been changed in https://github.com/ansible/ansible/pull/6759 already for that reason.

mscherer:
Is this one better ?

jimi-c:
Yep, merged. Thanks!

PR #7077, opened 2014-04-19
Added 'clear_keys' parameter to the 'authorized_keys' module

kustodian:
Added a new `clear_keys` parameter to the `authorized_keys` module which if it is set to yes (default: no), clears all keys from the authorized_keys file before adding a new key.

This parameter could be useful if you would like to be sure that the user's authorized_keys file only has the key(s) you would like it to have and no other. Because the `authorized_keys` module appends new keys and only deletes those whose state is put as `absent` it was very hard of deleting other keys which were not set by Ansible. `clear_keys=yes` could be an easy and fast solution for clearing all unneeded keys.

### DIFF
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -600,7 +600,7 @@ class Runner(object):
             if type(items) != list:
                 raise errors.AnsibleError("lookup plugins have to return a list: %r" % items)
 
-            if len(items) and utils.is_list_of_strings(items) and self.module_name in [ 'apt', 'yum', 'pkgng' ]:
+            if len(items) and utils.is_list_of_strings(items) and self.module_name in [ 'apt', 'yum', 'pkgng', 'authorized_key' ]:
                 # hack for apt, yum, and pkgng so that with_items maps back into a single module call
                 use_these_items = []
                 for x in items:
--- a/library/system/authorized_key
+++ b/library/system/authorized_key
@@ -342,8 +342,12 @@ def enforce_state(module, params):
     key_options = params.get("key_options", None)
     clear_keys  = params.get("clear_keys")
 
-    # extract indivial keys into an array, skipping blank lines and comments
-    key = [s for s in key.splitlines() if s and not s.startswith('#')]
+    # an array of passed SSH keys
+    keys = []
+    # 'key' is a string of all keys separated by a comma
+    for split_key in key.split(','):
+        # extract indivial keys into an array, skipping blank lines and comments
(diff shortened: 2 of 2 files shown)


juliedavila:
+1

jimi-c:
This seems somewhat counter-intuitive in regards to the example above, wouldn't it be better to use the file module first to clear it out if that was the desired action? At the very least, the items in the with_\* list should be combined into one to prevent issues (see how we do it for the packaging modules like yum and apt in `lib/ansible/runner/__init__.py`).

bcoca:
You can use template or assemble module to do this, it ensures that only
they keys you want are there.

## 

Brian Coca
Pedo mellon a minno

kustodian:
@jimi-c the idea behind this is if you want to always be sure that only the keys you setup with Ansible are in the authorized_keys file, you just set `clear_keys=yes` and you don't have to worry about anything. The problem with using file module is because it is not idempotent, since if you always clear the file and than set it with authorized_keys module, it will always report as changed. I agree about the the current problems with a list of keys and will check how it works with yum and apt modules.

@bcoca using template and assemble is a workaround, but why use a workaround when it can easily be done with the module which created for working with that?

I also thought about changing the `clear_keys` to `mode` which would be set to `append` by default and it could be set to `replace`, but I'm not sure if it is more intuitive?

jimi-c:
@kustodian I agree, state=replace would be confusing, as the keys are always replaced if the content or options are different. As I said above, if this were to function the way you have it, I would just account for the with_\* usage as we do the package modules.

kustodian:
@jimi-c thanks for the suggestion about with_*. I looked at the code and from the looks of it, it would be really easy to do, I would just need to add the authorized_key module into the list in `lib/ansible/runner/__init__.py` and in the module just split the keys with a comma. I will try it out later today or tomorrow and add an additional commit.

Since we are on subject of passing a list of with_\* items to the module and since it is more of a hack (there is even a comment saying it is a hack), maybe we should think about having this somehow built into the core. Modules could just pass to the AnsibleModule constructor something like `merge_with_items=True`, or whatever and it would pass the list of items to the module, instead of each item separately. Currently only 3 modules use this, so we would only need to modify 3 modules and this one (once I make the change). This would make it easier for all modules to support this in the future. What do you think about the idea?

kustodian:
Sorry for not getting back on this sooner, but I saw it's P4, so I guess it is not high priority.

I made the changes as @jimi-c suggested. Now the module works `with_*` the same as `yum` module.

P.S. I removed the previous limitation in the first comment, since the limitation doesn't exist anymore.

mpdehaan:
Hi!

Thank you very much!

PR #7078, opened 2014-04-19
Enhance error message about missing ControlPersist

mscherer:
ansible_ssh_args is good for playbook, but in the ~/.ansible.cfg
file, it should be ssh_args in a specific section.


jimi-c:
Merged, thanks!

PR #7080, opened 2014-04-20
An attempt to allow role directories to be flatter

etanol:
The amount of directory nesting inside the roles can be unconfortable when editing the files from the command line.  With a very small change it is possible to provide a choice when creating the role directory structure.

### DIFF
--- a/lib/ansible/playbook/play.py
+++ b/lib/ansible/playbook/play.py
@@ -469,6 +469,10 @@ class Play(object):
                  os.path.join(basepath, 'main.yml'),
                  os.path.join(basepath, 'main.yaml'),
                  os.path.join(basepath, 'main.json'),
+                 basepath,
+                 basepath + '.yml',
+                 basepath + '.yaml',
+                 basepath + '.json',
                 )
         if sum([os.path.isfile(x) for x in mains]) > 1:
             raise errors.AnsibleError("found multiple main files at %s, only one allowed" % (basepath))


brent-hoover:
I don't know the ramifications of the change, but I definitely like the idea of being able to eliminate so many role directories.

jimi-c:
Thank you for the contribution, however we have previously stated that we do not wish to flatten role directories in this manner. Also, doing it this way would restrict the role to only having tasks and would prevent it from using vars, meta, defaults, or any of the other benefits of the role directory structure we currently use. As always, we recommend users execute `ansible-galaxy init <role_name>` to create the directory structure and yaml files so they don't have to do it manually. Thanks again.

etanol:
I'm okay with a rejection due to policy or design.  However, the technical restrictions mentioned are just not true.  Please correct me if I'm wrong.

The proposed change does not restrict anything.  Apart from being fully backwards comptaible, it allows a bit more flexibility in the role directory tree.  For example, it would be possible to mix a `tasks` subdirectory (with a `main.yml` file in it) along with a `meta` file (or `meta.yml`).

etanol:
For the record, the link to [the relevant discussion](https://groups.google.com/d/topic/ansible-project/az_8fV0iBXM/discussion); just in case anyone is trying to implement the same and stumbles upon this.

jimi-c:
@etanol yes I see what you mean, but yes it's more of a policy thing as well. Thanks!

PR #7082, opened 2014-04-20
fixed issue with FreeBSD network facts

### LINKED ISSUE #5951: Not getting proper mtu on FreeBSD 10
##### Summary:

setup module not picking up proper mtu from FreeBSD ifconfig
##### Steps To Reproduce:
1. Run 'setup' module against FreeBSD 10.0-RELEASE machine
2. Look at ansible_lo0.mtu or any other interface
3. Compare to ifconfig output for the selected interface. See difference.
##### Expected Results:

ansible_vmx0.mtu should equal tue actual mtu from ifconfig
##### Actual Results:

ansible_<iface>.mtu == 0

bcoca:
ifconfig changed on FreeBSD so the parsing was off.

tested this on 8,9 and 10 latest stableversions and it now works, this fixes #5951

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -1748,8 +1748,9 @@ class GenericBsdIfconfigNetwork(Network):
     def parse_interface_line(self, words):
         device = words[0][0:-1]
         current_if = {'device': device, 'ipv4': [], 'ipv6': [], 'type': 'unknown'}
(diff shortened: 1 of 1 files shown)


bcoca:
fyi, fact gathering on FreeBSD is broken in current dev, but is fixed by #7081

jimi-c:
Merged, thanks!

PR #7083, opened 2014-04-20
Improve error message when multiple action are given

mscherer:
This can be illustrated using this playbook:

```
- command: id
  delegate_to: "{{ remote_server }}"
  user: "{{ remote_user }}"
```

The error is to use 'user' instead of 'remote_user', but the error message
do not really mention it, so it can be a bit hard to spot.


jimi-c:
Merged, thanks!

PR #7087, opened 2014-04-21
Create roles_file configurartion

EHER:
Create configuration roles_file to define the file used by ansible-galaxy to install roles. 

Example

```
# ansible.cfg
[defaults]
roles_file=ROLES_FILE
```

```
# ROLES_FILE
EHER.git
```

With that configurations into ansible.cfg and a [roles file](https://galaxy.ansible.com/intro) named ROLES_FILE, we are able to run `ansible-galaxy install` without parameters.

With that configuration we can create more flexible workflows into applications projects that holds ansible playbooks to configure environment.

That should not cause problems to other PR witch pretend make yml file available to configure roles.

---

It also fix random_string used into constants tests, that was ignoring length parameter and aways using 6.

### DIFF
--- a/bin/ansible-galaxy
+++ b/bin/ansible-galaxy
@@ -181,7 +181,7 @@ def build_option_parser(action):
             '-n', '--no-deps', dest='no_deps', action='store_true', default=False,
             help='Don\'t download roles listed as dependencies')
         parser.add_option(
-            '-r', '--role-file', dest='role_file',
+            '-r', '--role-file', dest='role_file', default=C.DEFAULT_ROLES_FILE,
             help='A file containing a list of roles to be imported')
     elif action == "remove":
         parser.set_usage("usage: %prog remove role1 role2 ...")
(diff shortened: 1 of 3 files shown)


jimi-c:
I don't quite see how this would be very useful, as it would always default to using a single roles file?

EHER:
With that configuration, we can have a ansible.cfg in each project with independent ROLES_FILE.

Some project like [PHP Amsterdam Website](https://github.com/AmsterdamPHP/amsterdamphp.nl) have his own playbook.

Would be great have a default configuration on project root (ansible.cfg), just like Vagrantfile. And then developers can use ansible-galaxy as other dependencies managers.

```
ansible-galaxy install
```

Other guys that use same pattern

```
npm install # noje.js
composer install # php
bundle install # ruby
```

The difference is that with ansible you do not need have a default file name to define roles list:
- specify via command line parameter -r (like pip)
- define file name into machine configuration (main ansible.cfg)
- define into a local ansible.cfg (project root or some other place of the project)

mpdehaan:
# Question About This One

Hi!

- ansible-galaxy takes a -r argument that is easy to use already
- it would be easy to alias this command if someone thought it was hard to use

PR #7089, opened 2014-04-21
Add an example to show all hostvars using debug

dagwieers:
This is a useful example to help debug how facts and vars are being collated.


jimi-c:
Merged, thanks!

PR #7094, opened 2014-04-21
Add check of return code for autoinstallation of python-apt

radeksimko:
When you try to use `apt` role without having `python-apt` installed and **without passing `--sudo` as an option to `ansible-playbook`**, it will fail with quite vague error message:

```
msg: Could not import python modules: apt, apt_pkg. Please install python-apt package.
```

Requested change will pass the stderr from `apt-get install` command further to the user, so the message will be actually helpful:

```
install python-apt -y -q", "failed": true, "item": "git,git-svn", "rc": 100}
stderr: E: Could not open lock file /var/lib/apt/lists/lock - open (13: Permission denied)
E: Unable to lock directory /var/lib/apt/lists/
E: Could not open lock file /var/lib/dpkg/lock - open (13: Permission denied)
E: Unable to lock the administration directory (/var/lib/dpkg/), are you root?

msg: E: Could not open lock file /var/lib/apt/lists/lock - open (13: Permission denied)
E: Unable to lock directory /var/lib/apt/lists/
E: Could not open lock file /var/lib/dpkg/lock - open (13: Permission denied)
E: Unable to lock the administration directory (/var/lib/dpkg/), are you root?
```

In fact current message will look like this until #7095  will gets merged.

```
fatal: [localhost] => failed to parse: {"stdout": "", "cmd": "apt-get update && apt-get install python-apt -y -q", "failed": true, "stderr": "E: Could not open lock file /var/lib/apt/lists/lock - open (13: Permission denied)\nE: Unable to lock directory /var/lib/apt/lists/\nE: Could not open lock file /var/lib/dpkg/lock - open (13: Permission denied)\nE: Unable to lock the administration directory (/var/lib/dpkg/), are you root?\n", "rc": 100, "msg": "E: Could not open lock file /var/lib/apt/lists/lock - open (13: Permission denied)\nE: Unable to lock directory /var/lib/apt/lists/\nE: Could not open lock file /var/lib/dpkg/lock - open (13: Permission denied)\nE: Unable to lock the administration directory (/var/lib/dpkg/), are you root?"}
{"msg": "Could not import python modules: apt, apt_pkg. Please install python-apt package.", "failed": true}
```

PR #7095, opened 2014-04-21
Prevent catching unrelated exceptions

radeksimko:
There are many places in the code where ansible simply catches all exceptions, which is a bad practice also called [_Pokemon exception handling_](http://www.dodgycoder.net/2011/11/yoda-conditions-pokemon-exception.html).

I believe we can fine-grain most of them (as it's actually mentioned in [the coding guidelines](https://github.com/ansible/ansible/blob/devel/CODING_GUIDELINES.md#exceptions)) and this is the first place where it's actually already causing issues.

If the `module.run_command(...` will trigger `sys.exit()`, then [`SystemExit`](https://docs.python.org/2/library/exceptions.html#exceptions.SystemExit) will be catched by this statement too and if there was already a JSON output (`fail_json()` has been already called before exiting), it will end up in parse error on the output as two `fail_json()` calls will generate two unparsable concatenated JSON strings.

The erroneous situation can be simulated within #7094

### DIFF
--- a/library/packaging/apt
+++ b/library/packaging/apt
@@ -436,7 +436,7 @@ def main():
             global apt, apt_pkg
             import apt
             import apt_pkg
-        except:
+        except ImportError:
             module.fail_json(msg="Could not import python modules: apt, apt_pkg. Please install python-apt package.")
 
     global APTITUDE_CMD


jimi-c:
Merged, thanks!

Issue #6947, opened 2014-04-11
[ansible.cfg] Add vault_password_file option


ruudk:
##### Issue Type:

Feature Idea
##### Ansible Version:

ansible 1.5.4
##### Environment:

Ubuntu 14.04
##### Summary:

I'm using vaults and have the password stored inside a vault-password-file. I don't like to add the `--vault-password-file=~/.ansible_vault_password.txt` to it every time I run commands (like `ansible` and `ansible-vault`). Would be great if it would read this setting out of `ansible.cfg`
##### Steps To Reproduce:

Add `vault_password_file=file.txt` to `ansible.cfg`

gerhard:
Not sure if this is the best place to post this, but I've put together a quick bash wrapper script around `ansible-vault` that leverages OS X's built-in keychain to store the vault password:

``` sh
#!/usr/bin/env bash

tmpfile=$(mktemp -t vault)

ensure_tmpfile_gets_removed() {
  rm -f $tmpfile
}

trap ensure_tmpfile_gets_removed EXIT

echo $(security find-generic-password -a ansible_vault -w) > $tmpfile

command="$1"
shift
args=$@

ansible-vault $command $args --vault-password-file $tmpfile
```

I have a single ansible_vault password, but it would be trivial to add a password per-file for the more paranoid.

ruudk:
That's awesome :)

resmo:
+1

mpdehaan:
(I do disagree with this needing to be a config option since different projects may have multiple vault password files)

Rather, this can be easily solved with bash aliases.

I do like the above script idea a lot, how about contributing a PR to add it to examples

resmo:
ok, I see, makes sense.

ljsommer:
This would be an incredibly useful feature. Storing this value in a config file is useful on its own but adding vault password files to inventory groups, roles, etc would save a lot of time.

adrianlzt:
+1

sivel:
Also see #7649

Petrox:
```
(I do disagree with this needing to be a config option since different projects may have multiple vault password files)
Rather, this can be easily solved with bash aliases.
```

Different project may have multiple vault password files: true
That's why we've got different ansible.cfg for each project (thus making versioning easier).

So +1 for "Add vault_password_file=file.txt to ansible.cfg"

(Nearly all commandline options should have a cfg alternative. If you don't want to use it, you can, but if someone needs that, it's his/her decision to do.)

PR #7097, opened 2014-04-21
Bash wrapper that integrates OS X Keychain with ansible-vault

gerhard:
Initially posted as an alternative to making the ansible vault password file an `ansible.cfg` option #6947.

I have a similar approach for `ansible-playbook` and `ansible` commands, I find the whole typing distracting from the end goal. Happy to share them if anyone's interested.

cc @mpdehaan

### DIFF
--- a/examples/scripts/ansible-vault-os-x-keychain.bash
+++ b/examples/scripts/ansible-vault-os-x-keychain.bash
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+#
+# This bash wrapper allows me to use ansible-vault in combination with OS X's
+# Keychain for securely handling the encryption password with no input. No
+# prompts, to typos, same security as SSH key passphrases.
+#
+# I am using the "Ansible Vault" application password entry for the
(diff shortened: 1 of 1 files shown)


sivel:
What are your thoughts on handling this with python instead, and make use of the keyring module? This would allow this to work on more than just OS X.

gerhard:
That does sound like a better solution indeed. As my Python skills are 0 and the rather opinionated bash solution solves my problem, I'm happy to leave this here for others until a more robust implementation.

paulnicholson:
I used an anonymous fifo to eliminate the temp file.

```
exec ansible-playbook $args --vault-password-file <(echo "$PASSWORD")
```

gerhard:
Very nice tip, thanks ; )

renaudguerin:
Would love this to be merged

paulnicholson:
+1

sivel:
With #7649 already merged, and pending #8561 and #8560 I believe this PR can be closed.

gerhard:
Happy to close as I would personally like to use your solution long-term.

PR #7099, opened 2014-04-21
Update facts.py

### LINKED ISSUE #7093: openbsd interfaces detection
##### Actual Results: list of interfaces plus some out-of-nowhere nic

Here is a sample patch for system/setup (get_interfaces_info from GenericBsdIfconfigNetwork):

1735 was:
if re.match('^\S', line) and len(words) > 3:
1735 becomes:
if words[0] == 'pass':
   continue
elif re.match('^\S', line) and len(words) > 3:

Regards.

faust64:
Fixes #7093 --devel


jimi-c:
Thanks!

Issue #7103, opened 2014-04-21
template module's error messages too terse and confusing; generic solution - add --traceback switch?


pfalcon:
##### Issue Type:

Feature Idea
##### Ansible Version:

ansible 1.6 (devel f1ce9abe0d) last updated 2014/04/21 21:05:18 (GMT +300)
##### Environment:

Ubuntu 12.04
##### Summary:

I'm suffering following error message when working with "template" action:

```
fatal: [tcwg.ci.linaro.org] => {'msg': 'files/private/jenkins_ssh_publishers.xml.tcwg-ci', 'failed': True}
```

Supposedly, that gives me, a developer, some hints, but will definitely drive a user (== someone who got a playbook to execute, but didn't write it) nuts.
##### Steps To Reproduce:

In this specific case it happens because jinja2 cannot find file included within template.
##### Expected Results:

More detailed error message.
##### Actual Results:

Reported behavior is caused by the following line:

```
result = dict(failed=True, msg=str(e))
```

The problem, default Python's Exception.**str**() just prints exception "args" attribute (a tuple), try "print AssertionError(1,2,3)", so you don't have any context of what actual error is.

I'm going to submit a patch to fix (or at least improve) that specific case, but I would like to raise a more general question: can we have a mode in Ansible which causes all (or most) tracebacks happening during playbook execution to be fully printed, instead of just error messages which may be too terse. 
This would be similar to django-admin's switch --traceback.

Having tracebacks would help debugging playbook much.

mpdehaan:
Template includes are known to be a little off right now, there are a few tickets on this.

pfalcon:
What about idea for --traceback switch, is it viable? I may imagine it would be a bit of chore to get complete coverage for it, but can be done gradually, just need to know it's otherwise fits Ansible paradigm.

mpdehaan:
I'd rather not have a traceback switch.

pfalcon:
Ok, closing. Hope that pull request #7105 is still useful.

PR #7105, opened 2014-04-21
template: When rendering exception from Jinja, include exception name.

pfalcon:
This will give basic context or _what_ the error is, otherwise there're
messages like:

fatal: [tcwg.ci.linaro.org] => {'msg': 'files/private/jenkins_ssh_publishers.xml.tcwg-ci', 'failed': True}

Which give a filename related to error, but not an idea what error is.

Addresses #7103.

### DIFF
--- a/lib/ansible/runner/action_plugins/template.py
+++ b/lib/ansible/runner/action_plugins/template.py
@@ -87,7 +87,7 @@ class ActionModule(object):
         try:
             resultant = template.template_from_file(self.runner.basedir, source, inject, vault_password=self.runner.vault_pass)
         except Exception, e:
-            result = dict(failed=True, msg=str(e))
+            result = dict(failed=True, msg=type(e).__name__ + ": " + str(e))
             return ReturnData(conn=conn, comm_ok=False, result=result)
 
         local_md5 = utils.md5s(resultant)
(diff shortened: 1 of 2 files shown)


mpdehaan:
I think a better solution might be to detect and catch this type of error message and then return a different string message that explains the error in greater depth.   If possible.

pfalcon:
Yes, makes sense for "template not found" error, but the problem is that who knows how many more exceptions Jinja2 can throw, and for what reasons. All of them would be hard to predict... What about idea in #7103 - to provide CLI switch to show exception tracebacks to user?

Otherwise, should I give this another try or leave it to you guys? My guess that such check would belong to lib/ansible/utils/template.py , which appears to be proxy between Jinja2 and rest of Ansible, is that good place or do I miss something?

pfalcon:
Ok, I added user-friendly handling of TemplateNotFound, while still left previous generic handling of jinja2 exceptions in case something else might pop up in there.

Example of reporting:

```
fatal: [tcwg.ci.linaro.org] => {'msg': 'AnsibleError: file: /home/pfalcon/devel/linaro/linaro-jenkins-tools/ansible/roles/jenkins-ssh-publish/templates/jenkins.plugins.publish_over_ssh.BapSshPublisherPlugin.xml, error: Cannot find/not allowed to load (include) template ../../ansible-secrets-jenkins/jenkins_ssh_publishers.xml.tcwg-ci', 'failed': True}
```

Let me know if this needs further tweaks.

pfalcon:
ping

jimi-c:
I merged this in while squashing the commits, so it was merged despite github's UI indicating it was simply closed. Thanks!

PR #7123, opened 2014-04-23
Add a salt when creating the password hash.

doanerock:
Salts should be used when creating a password hash.


mpdehaan:
I'm not quite following what the user is meant to do with this command, it seems we would want to update the above line if we didn't like how it was generating things, and only show one python one-liner.

Can you elaborate further on the intent?

Thanks!

doanerock:
I was saying that a salt should also be in the password hash, but on further inspection it appears that the python one liner that was already in the documentation generates a random salt for the password hash.  So the one liner that I added would only be necessary if you want to add a custom salt instead of random salt which is not really any more secure.  

My bad.  Thanks

Thanks

PR #7126, opened 2014-04-23
Give more information when a task is empty

mscherer:
I made a typo in a playbook and was great by:

```
ERROR: expecting dict; got: None
```

The issue was a single - on the last line of a playbook.
With the name of the file, I was able to see right away where the
error was.


jimi-c:
Merged, thanks!

PR #7128, opened 2014-04-23
update ec2_tags documentation to include missing option

tedder:
'list' was added more recently, it was omitted from the documentation.


jimi-c:
Merged, thanks! It would probably also be good to add a note in the description there that 'list' was added in the current version.

PR #7129, opened 2014-04-23
Possible docs error: 'path' instead of 'src'

erikaheidi:
I came to this example today (line 50), and when I tried it, I got an error:

`msg: src and dest are required for creating links`

So, I changed **path** to **src** and it worked fine. Looks like the example is wrong, since the docs say "dest" is already an alias for "path", in which case we would have 2 "path" definitions.


jimi-c:
Merged, thanks! The src= param is required only for links, so that example was definitely incorrect.

erikaheidi:
awesome \o/

PR #7131, opened 2014-04-23
cloud/ec2_elb module - Deregister an instance even if its already OutOfService

bpennypacker:
##### Issue Type:

Bugfix Pull Request
##### Ansible Version:

ansible 1.6 (devel 2be8feebce) last updated 2014/04/22 18:45:33 (GMT +000)
##### Environment:

N/A
##### Summary:

The current behavior of the ec2_elb module does not match the documentation.  If an EC2 instance is registered with the ELB, but it's current state is OutOfService, then "state=absent" does nothing.  The instance remains registered with the ELB.
##### Steps To Reproduce:

Create an AWS EC2 instance.  Create an ELB instance and associate the EC2 instance with it.  Ensure the instance is listed as OutOfService in the ELB (no need for anything to be running on the EC2 instance).  Use the ec2_elb module as follows:

```
ec2_elbs=<name of ELB>
instance_id=<id of EC2 instance>
state=absent
```
##### Expected Results:

The instance should be removed from the ELB
##### Actual Results:

The instance is ignored because it is in the OutOfService state.  This pull request changes the behavior to be what is actually expected.  This change basically boils down to:
- If the instance is not found in the ELB then simply return with changed=false
- If the instance is found in the ELB, no matter the state (InService or OutOfService), deregister it from the ELB and return changed=true

### DIFF
--- a/library/cloud/ec2_elb
+++ b/library/cloud/ec2_elb
@@ -129,18 +129,20 @@ class ElbManager:
 
         for lb in self.lbs:
             initial_state = self._get_instance_health(lb) if wait else None
-
-            if initial_state and initial_state.state == 'InService':
-                lb.deregister_instances([self.instance_id])
-            else:
+            if initial_state is None:
+                # The instance isn't registered with this ELB so just 
+                # return unchanged
                 return
+
+            lb.deregister_instances([self.instance_id])
+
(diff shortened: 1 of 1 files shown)


jimi-c:
Merged, thanks!

PR #7133, opened 2014-04-23
file multiple new dirs permissoins

### LINKED ISSUE #7130: File doesn't create sub-directories with the same permissions as the top level directory
##### Issue Type:

“Bug Report”
##### Ansible Version:

ansible 1.5.4
##### Environment:

Ubuntu 14.04
##### Summary:

The file module doesn't create sub-directories with the same permissions as the top level directory. 
##### Steps To Reproduce:

``` yml
- name: Configure the instance
  hosts: myhosts
  user: ec2-user
  sudo: true
  gather_facts: true
  tasks:
    - name: Create project directory
      file:
        path: "/home/ec2-user/apps/myapp"
        owner: ec2-user
        group: ec2-user
        state: directory
```
##### Expected Results:

The `apps` directory should have the same owner and group as the `myapp` directory.
##### Actual Results:

The `apps` directory is owned by root.

bcoca:
now when making multiple dirs in path to make directory, same permissions are assigned (but only for NEW dirs). Not that I think this is a good idea but its been asked for soo many times that I'm willing to do it just to remove the noise (fixes #7130)

Signed-off-by: Brian Coca <briancoca+ansible@gmail.com>

### DIFF
--- a/lib/ansible/utils/module_docs_fragments/files.py
+++ b/lib/ansible/utils/module_docs_fragments/files.py
@@ -30,7 +30,7 @@ options:
   state:
     description:
       - If C(directory), all immediate subdirectories will be created if they
-        do not exist, since 1.6 they will be created with the supplied permissions.
+        do not exist, since 1.7 they will be created with the supplied permissions.
         If C(file), the file will NOT be created if it does not exist, see the M(copy)
         or M(template) module if you want that behavior.  If C(link), the symbolic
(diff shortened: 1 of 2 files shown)


jimi-c:
I'm ok with this, but a note on the behavior change should be made in the module docs.

jimi-c:
Merged, thanks!

PR #7134, opened 2014-04-23
Show how to optionally include vault playbook

toddwells:
Excluding a vault playbook still requires the vault password using most include methods.  Added description shows how include_vars can be used to optionally include a vault playbook and not require the vault password if it is not included.


mpdehaan:
I'm generally not in the mode of including tricks and workarounds in mainline docs as they impede the flow a bit, but we'll think about where we can put this one.

I think a better solution _MAY_ be to detect if it's going to encounter any vault files, though that's tricky to do not in runtime.

mpdehaan:
I think we're going to leave this out of main docs, as per the above.  Thanks for showing the tip and if it comes up this is a good solution.  Thanks!

toddwells:
Yes, a solution that does not require this kind of trick would be nice, but
I understand the potential difficulties.

Is there any place were solutions like this are documented?  I frequently
have to spend a lot of time trying to find solutions to problems and I have
not yet found any good location to get help on this.

PR #7142, opened 2014-04-23
Vmware core

rhoop:
This module provides a way to create/delete/manage a guest VM through VMware vSphere.

Discussed with michael@ansible.com.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,8 @@ credentials.yml
 results.xml
 coverage.xml
 /test/units/cover-html
+# Development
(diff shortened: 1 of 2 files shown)


jimi-c:
Merged, thanks! I would add cloning to the list of todos for this as well, as I know in past jobs where we used vsphere we typically kept gold images to deploy machines more quickly (even if they were just standard hardware configs that did a PXE boot to kickstart).

rhoop:
Can do. I will try and have something in next week.

PR #7148, opened 2014-04-24
 add back sudo_output to stdout when the sudo needn't passwd

qus-jiawei:
when ansible running sudo command on a remote machine that needn't enter passwd, it would lost some output ,as it throught them was the sudo passwd hit.

### DIFF
--- a/lib/ansible/runner/connection_plugins/paramiko_ssh.py
+++ b/lib/ansible/runner/connection_plugins/paramiko_ssh.py
@@ -191,6 +191,7 @@ class Connection(object):
                 msg += ": %s" % str(e)
             raise errors.AnsibleConnectionFailed(msg)
 
+        stdout = ""
         if not (self.runner.sudo and sudoable) and not (self.runner.su and su):
             if executable:
                 quoted_command = executable + ' -c ' + pipes.quote(cmd)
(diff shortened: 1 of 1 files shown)


jimi-c:
Could you provide an example command that shows this behavior, so we can verify this patch corrects it? Thanks!

qus-jiawei:
ansible-playbook:

```
 - hosts:
   - host1
   tasks:
   - name: copy java to machine
     sudo: yes
     unarchive: src=/xxx/jdk.16.jar dest=/usr/java/
```

and the host1 should set the user could sudo without passwd.

/etc/sudoers

```
user1   ALL=(ALL)   ALL
user1  ALL=(ALL)       NOPASSWD: ALL
```

And I have try this on ansible 1.4.3 and it work fine.
Maybe, it is only effect on ansible 1.5+

jimi-c:
Looking back at this, with and without your patch, I'm not seeing any difference in output when running a simple command. Could you also include the difference in stdout for your use-case above? I'm unclear as to why the unarchive output would be different.

mpdehaan:
Closing this due to lack of info on the above - please feel free to submit with information if you like.

Thanks!

PR #7149, opened 2014-04-24
Adjust text for validate. Put all the text in one spot.

jjshoe:


jimi-c:
Two minor notes:
1. comment from old code: '# AWS only documentation fragment'
2. would it also be better to move the required and default fields into this? Making them consistent across the different modules would be nice as well.

jjshoe:
@jimi-c I looked at version_added, I knew that changed. I'll look into the required/default fields and get them moved in if appropriate. Whoops on copying in the aws only document fragment piece.

jjshoe:
@jimi-c All cleaned up and good to go!

mpdehaan:
This diff appears to remove the body of the "validate" documentation but doesn't remove the actual keys from each entry or the "version_added" parts of the doc.

Can you update this?

Thanks!

jjshoe:
@mpdehaan "version_added" can not be removed, as the version in which validate was added is different among modules.

Issue #7146, opened 2014-04-24
Cannot install ansible with pip on OSX


brendanfalkowski:
##### Issue Type:

Documentation Report
##### Ansible Version:

Cannot install Ansible yet
##### Environment:

Mac OS X Mavericks (10.9.2)
##### Summary:

Trying to install Ansible on Mac OS X Mavericks (10.9.2) so I can setup a testing environment in Vagrant. I'm a frontend developer — brace yourself. I don't have Homebrew or MacPorts installed. I read pip is the best way to install Ansible, and to avoid other package managers. Have been Googling/reading for about 3 hours, but no luck.
##### Steps To Reproduce:

Here is exactly what I did, as per:
http://docs.ansible.com/intro_installation.html#latest-releases-via-pip

```
sudo easy_install pip
```

This executed fine.

```
pip --version
pip 1.5.4 from /Library/Python/2.7/site-packages/pip-1.5.4-py2.7.egg (python 2.7)
```

FYI: Python is also bone-stock OSX.

```
python --version
Python 2.7.5
```

Then came the problem:

```
sudo pip install ansible
```

Which leads to this error output:
https://gist.github.com/brendanf/11240002

Is there some assumed dependency that I'm missing? This is basically an unmodified Mac except for NodeJS and git being installed. I apologize up front, but the StackOverflow answers I found about gcc, cflags, and clang flew right over my head.

The docs suggest after two commands it works so I must be doing something obvious incorrectly. Appreciate any help.
##### Expected Results:

Ansible installs.
##### Actual Results:

Ansible does not install.

joshuaswarren:
When you get to the step 'sudo pip install ansible', do this instead:

``` shell
sudo su -
export CFLAGS=-Qunused-arguments
export CPPFLAGS=-Qunused-arguments 
pip install ansible
exit
```

That last line is to just make sure you log out of the root account before you break anything accidentally. 

This doesn't appear to be ansible-specific, and instead appears to be caused by the version of clang that Mavericks is using and the default settings in it. So if you encounter a similar problem with a 'pip install' command in the future, this _should_ resolve it.

brendanfalkowski:
Well, that totally worked. Thanks @joshuaswarren! I did read that Mavericks has an issue with clang, but didn't know what to do with that info.

Ironically, I did try executing those export commands (as my user account not as root... d'oh). Reference: http://stackoverflow.com/a/22427311/1497746

[...]

PR #7150, opened 2014-04-24
Play nicely with osx mavericks

jjshoe:
Make it easier for pip installs on osx mavericks as per #7146. Should have no side affects for linux or older versions of osx.

### DIFF
--- a/docsite/rst/intro_installation.rst
+++ b/docsite/rst/intro_installation.rst
@@ -228,7 +228,7 @@ your version of Python, you can get pip by::
 
 Then install Ansible with::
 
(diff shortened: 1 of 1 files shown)


jimi-c:
@jjshoe I think this needs to be a note for Mavericks only, rather than for all OSX versions.

jjshoe:
@jimi-c I'm just following what @mpdehaan said: https://github.com/ansible/ansible/issues/7146#issuecomment-41273138

mpdehaan:
Added some pip install notes that will be live soon, should be good to go, thanks!

Issue #7139, opened 2014-04-23
(Doc fix) "Inventory": ports listed in ssh config can be used


jf:
##### Issue Type:

Documentation Report
##### Ansible Version:

1.6 but doesnt matter!
##### Environment:

N/A
##### Summary:

hi, first up sorry if this has been raised already. I'm guessing probably not, because if it had, the fix would have been in already (it's not like fixing a typo is a terribly difficult task?). I'm just working my way through the documentation right now... so do expect future issue reports if I spot any.

http://docs.ansible.com/intro_inventory.html says:

```
... Ports listed in your SSH config file won’t be used, so it is important that you set them if things are not running on the default port:
```

Actually this isnt true anymore if OpenSSH is used instead of paramiko. As http://docs.ansible.com/intro_getting_started.html states:

```
By default, Ansible 1.3 and later will try to use native OpenSSH for remote communication when possible. This enables both ControlPersist (a performance feature), Kerberos, and options in ~/.ssh/config such as Jump Host setup.
```

So yes, even if this used to be the case (that ports in the SSH config file are _not_ used), this is now no longer 100% true all of the time; so it would be good to preface that note in "Inventory" about ports with a condition so that it's doesnt conflict with the earlier note ("Getting Started"), and the documentation as a whole is inherently consistent and not confusing.

Would also be good to state what happens when there's a conflict between the port stated in the ansible config, vs the port stated in the SSH config file.

Otherwise, good job and thank you!
##### Steps To Reproduce:

N/A
##### Expected Results:

Note should be fixed.
##### Actual Results:

See Summary.

jimi-c:
Thanks, we'll take a look at this. The docs are contained within the repo itself, so if you want to send a PR to fix the above feel free.

mpdehaan:
Yep, we're happy that these are used, stale docs must be fixxored.

Thanks!

jf:
Got it, and thanks. I'll try to work on a PR hopefully by the end of the week.

mpdehaan:
I applied some slightly different edits on the devel branch -- sorry for confusion and pickiness there, will be live soon.

Thanks!

PR #7152, opened 2014-04-24
Documentation fix for #7139: correct information on port numbers, ...

jf:
How about this? I struggled a bit with the exact wording, but finally settled on this.


jf:
The chief question I have about submitting a change would be about the language (the restructured format is not a problem for what I need to do here). Will this work (language- and accuracy-wise)?

mpdehaan:
Merged some updates for this just now so should be good to go.  Thanks!

PR #7154, opened 2014-04-24
documentation fix availability_zones should be required for ec2_asg

budlight:


budlight:
Right now this attribute is required, but it might be desired to actually fix the module so that it isn't required.

jimi-c:
The params list does not show this as being required, does it throw an error if it is not set?

budlight:
yes it does

jimi-c:
Ok, then go ahead and modify the param definition as well, and I'll get this merged in. Thanks!

budlight:
It appears this is technically correct in that availability_zones OR vpc_zone_identifier is required in the boto api.  I still think that if both a vpc and availability_zones aren't set it should pass a list of all the availability_zones for that region, this is kinda messy as you have to create another boto connection object.

I'm personally not aware of how the test setup on this project works, so I haven't ran or modified the test suite at all. 

 If this pasts the existing tests it should be ready to merge.

mpdehaan:
@jimi-c noticed this contains a code change and not just a docs update, so I'm retagging this one.

jimi-c:
Merged, thanks!

Issue #5841, opened 2014-01-31
Allow async command to work with_<lookup> loops


nathanhruby:
Using ansible 1.4.3 or later,  I would like to iterate over a hash of items using with_items and run a action (git clone, pip install, etc..) with each of those items using async.  This is so I don't need to worry about ssh timing out during some of the commands, which can take upwards of 30 minutes the first run.

Presently this is not allowed and ansible errors with:

```
fatal: [default] => lookup plugins (with_*) cannot be used with async tasks
```

It seems as if there are two action items:
- The documentation should provide a warning/caveat about this on the playbook_async page.
- async should work with lookup / with_*

mpdehaan:
Yeah this comment is in there because it doesn't work.

I think it's more likely this should just be in the docs for async, since the code is basically saying it's not going to be implemented.

nathanhruby:
Hi,

I can make a patch to the docs later this weekend.

I was unsure if this is also a issue that should be fixed or this was working as designed.  This post makes it seem like there's a possibility in the code to make it happen (though I have not looked at said code to determine feasibility / viability):
https://groups.google.com/forum/#!msg/ansible-project/3gtO9c__E6I/y_DipDKRbgUJ

If this is the design as intended, I can live with that, modulo the doc update. 

Thanks!

-n

nathanhruby:
made PR for doc issue

neersighted:
Any progress on this? I spawn a bunch of processes with `with_nested` and could really use fire-and-forget.

garlandkr:
I like it. I need it. I want it.

``` yaml
- name: Delete the ASGs which have LBs
  ec2_asg:
    name: "{{ vpc_prefix }}-{{ item.key }}"
    region: "{{ region }}"
    state: "absent"
  async: 1000
  poll: 0
  with_dict: "{{ auto_scaling_group_lb | default({}) }}"
  register: asg_delete_results

# Wait for all asynchronous tasks to complete
- name: Check ASG delete task
  async_status: jid={{ item.ansible_job_id }}
  register: asg_delete_result
  until: asg_delete_result.finished
  retries: 30
  with_items: echo.asg_delete_results
```

mmahut:
Would love to see this too.

dkoppstein:
Yes, this is super important for long-running apt-get processes. 

Also, why not add functionality to run a whole playbook as an async command?

leafpeak:
[...]

PR #7155, opened 2014-04-24
#5841: update playbook_async to note that iterators are not compatible w...

nathanhruby:
...ith async


mpdehaan:
I'm going to decide to not include this right now, the reason being it does generate an error, but also that we'd like to fix the underlying issue instead and this would likely leave a stale entry in the docs.

Once can also use lookup plugins, just not the "with_" stuff.   

Runner refactoring should make this possible, but I'll admit it may be a ways out.

Thanks!

nathanhruby:
Sounds good to me, thanks!

PR #7156, opened 2014-04-24
Missing 'base64' import

ricard33:


mpdehaan:
Merged, thanks!  Not good to see this missing obviously, I'm wondering if this was lost from module common imports at one point in time, it's definitely down a side-path.

This will be in 1.6.

PR #7162, opened 2014-04-25
Pagerduty module additions

jasonarewhy:
Tested with test_module and in a production environment. 

Added:
- ability to end maintenance windows with the 'stopped' state
- token auth (requester_id defaults to user)
- set minutes when creating a maintenance
- lookups for service ID and user ID
- implemented 'changed' logic
- added error handling to PD API calls

Changed:
- 'user' can now be name, email or ID
- 'service' can now be name or ID
- deprecated redundant 'running' state, merely an alias to 'started'

### DIFF
--- a/library/monitoring/pagerduty
+++ b/library/monitoring/pagerduty
@@ -13,11 +13,11 @@ requirements:
 options:
     state:
         description:
-            - 'started' will create a new maintenance window using
-            the service and desc fields. 'stopped' will end all
-            maintenance windows that match the desc and service.
-            Use 'ongoing' to retreive a JSON list of ongoing
-            maintenance windows.
+            - started will create a new maintenance window using
+              the service and desc fields. 'stopped' will end all
+              maintenance windows that match the desc and service.
+              Use 'ongoing' to retreive a JSON list of ongoing
+              maintenance windows.
         required: true
         default: null
         choices: [ "started", "stopped", ongoing" ]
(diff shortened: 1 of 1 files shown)


jimi-c:
The ID fields are numeric, aren't they? It should be possible to detect that they are IDs and if not do the lookups instead, that way we don't need to break backwards compatibility.

jasonarewhy:
They're alphanumeric, unfortunately (ex: "P96Z8XC"). The only defining characteristics are that they start with "P" and are 7 characters long. Using those, a service name like "PINGDOM" would slip through. Would love to make this backwards compatible somehow, though.

jasonarewhy:
@jimi-c I'm going to try adding a second lookup by ID if the first one for the user/service fails to provide any results.

jimi-c:
Sounds good.

jasonarewhy:
@jimi-c added and tested that it works as expected. The only remaining backwards compatibility issue is the 'running' state that I removed, which was redundant to 'started', which was confusing. If it needs to stay in, I could add it to the code but leave it out of the docs perhaps? Not sure how you prefer to handle deprecation.

jimi-c:
Yes, internally it would just be an alias. You could put a note in the description field for that if you wanted.

jasonarewhy:
@jimi-c should be all set here, by the way, let me know if there's anything else I should do to prepare this for merging.

mpdehaan:
Hi!

Thanks you very much!

PR #7163, opened 2014-04-25
Update playbooks_roles.rst

ngealy:
I was trying to put the -include: directive under task: but I kept getting and ERROR: tasks is not a legal parameter in an Ansible task or handler.  I had to look at sample playbook to see how to properly use -include, because these instructions were confusing.

### DIFF
--- a/docsite/rst/playbooks_roles.rst
+++ b/docsite/rst/playbooks_roles.rst
@@ -51,23 +51,20 @@ A task include file simply contains a flat list of tasks, like so::
 
 Include directives look like this, and can be mixed in with regular tasks in a playbook::
 
-   tasks:
-
      - include: tasks/foo.yml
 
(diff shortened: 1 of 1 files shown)


mpdehaan:
an include can happen under a tasks section.   It can also be used in a playbook to include plays.

As such, I'm not ok with this change as stated, but want to understand how you got into an error condition.

Can you paste the playbook you had?

What you include must be a list of tasks, and not a play, so I think that's the problem -- you included a playbook file, not a list of tasks.  Which is what we should correct rather than changing what the example is talking about.

(As such, we won't merge this, but we should add a clarification)

ngealy:
Yeah, you are correct.  I was trying to include a playbook instead of a task.  I didn't realize you could create a .yml file that contains just tasks. I already fixed my error, so I can't paste it here.  Sorry for the confusion.

PR #7170, opened 2014-04-27
always set user in SSH connection

thomas-mangin:
therefore there is no reason to not be explicit about the user.

### DIFF
--- a/lib/ansible/runner/connection_plugins/ssh.py
+++ b/lib/ansible/runner/connection_plugins/ssh.py
@@ -93,8 +93,7 @@ class Connection(object):
             self.common_args += ["-o", "KbdInteractiveAuthentication=no",
                                  "-o", "PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey",
                                  "-o", "PasswordAuthentication=no"]
-        if self.user != pwd.getpwuid(os.geteuid())[0]:
-            self.common_args += ["-o", "User="+self.user]
+        self.common_args += ["-o", "User="+self.user]
         self.common_args += ["-o", "ConnectTimeout=%d" % self.runner.timeout]
 
         return self


mpdehaan:
There's not enough here for me to understand what you mean to do with this change.

Can you please provide more info about how you ran into a problem with this code, and expected vs actual results?

Thanks!

thomas-mangin:
Hello Michael,

I want to be able to setup SSH on new server using ansible, changing the port and installing the key under for root, on newly installed Ubunty 14.04 servers to the use ansible "normally".

On ubuntu, root login with password is disabled, therefore the only way to perform this first installation with ansible is by using the normal user account created at setup and its password.

In order to perform such installation I have ansible setup like follow 

```
$ cat ~/ansible/ssh.yml
---
- hosts: setup
  remote_user: thomas
  sudo: True
  sudo_user: root
  vars:
    ansible_ssh_user: thomas
    ansible_remote_user: thomas
    ansible_ssh_port: 22
    host_key_checking: false
  tasks:
    - name: Check that the server's alive
      action: ping
  #  - name: update ssh
  #   apt: pkg=openssh-server state=latest update_cache=yes
  # - name: change the ssh port
  #   lineinfile: dest=/etc/ssh/sshd_config regexp="^Port " insertafter="^#Port " line="Port {{ new_ssh_port }}"
  #   notify:
  #   - restart sshd
  # - name: ensure sshd is running
  #   service: name=sshd state=started
  handlers:
    - name: restart ssh
      service: name=sshd state=restarted

$ cat ~/ansible/hosts.ini
[setup]
10.0.0.10

$ ansible-playbook --ask-pass --inventory ~/ansible/hosts.ini ~/ansible/ssh.yml -vvvv
SSH password:

PLAY [setup] ******************************************************************

<10.0.0.10> ESTABLISH CONNECTION FOR USER: thomas
<10.0.0.10> REMOTE_MODULE setup
<10.0.0.10> EXEC ['sshpass', '-d4', 'ssh', '-C', '-tt', '-vvv', '-o', 'ControlMaster=auto', '-o', 'ControlPersist=60s', '-o', 'ControlPath=/Users/thomas/.ansible/cp/ansible-ssh-%h-%p-%r', '-o', 'Port=22', '-o', 'GSSAPIAuthentication=no', '-o', 'PubkeyAuthentication=no', '-o', 'ConnectTimeout=10', '10.0.0.10', "/bin/sh -c 'mkdir -p /tmp/ansible-tmp-1398632669.83-205011020778813 && chmod a+rx /tmp/ansible-tmp-1398632669.83-205011020778813 && echo /tmp/ansible-tmp-1398632669.83-205011020778813'"]
fatal: [10.0.0.10] => Authentication failure.

TASK: [Check that the server's alive] *****************************************
FATAL: no hosts matched or all hosts have already failed -- aborting

           to retry, use: --limit @/Users/thomas/ssh.retry

10.0.0.10                  : ok=0    changed=0    unreachable=1    failed=0
```

As you can see from the sshpass command, the "User=thomas" line is not included, and the ssh connection fails. Adding the "User" parameter fixes the issue. 

I am running this command on OS X 10.9.2 with sshpass installed with the following brew command

```
brew install https://raw.github.com/eugeneoden/homebrew/eca9de1/Library/Formula/sshpass.rb
```

Sorry for the initial first short commit message, but when using virtualenv with the patched version of ansible, I reliably crash my mac !  It seems I found a way to trigger a bug in parallels which runs my ubuntu host. After two crashs I gave up on trying to record the whole output, and went for "here it is".

Hopefully this is the information you wanted to understand the change. It may be that the patch need to be : 

```
self.common_args += ["-o", "User="+ self.user or pwd.getpwuid(os.geteuid())[0]]
```

Looking at the code I believe self.user is always initialised, but I am sure you know better. Feel free to just change the code the way you know is best :-)

Thank you for your time.

thomas-mangin:
Michael, the reason for the patch is that on my test system my local user name was the same as the remote user on the server, which is not going to be the case in our development environment.
Therefore I do not really need this patch, so if you do not intend to pull it, could you please let me know and I will delete my tree as I have nothing more to contribute. Thank you for writing ansible. Thomas.

jimi-c:
I have merged this in, but squashed it with a commit to switch the user to what you noted above to catch the situation where `self.user` may not have been set. Since the github UI doesn't deal with squashes, this issue was closed as part of the commit message. Thanks!

fly-me:
This modification breaks compatibility. A specific user name may be unwanted if it is set up in ssh_config file.

thomas-mangin:
Could you please be more specific, I am as well failing to see the problem you are reporting and would rather not see this patch removed but if anything improved.

fly-me:
My ~./ssh/config file contains the following lines
Host 10.0.1.*
    User ubuntu
    IdentityFile ~/.ssh/somekey
My user name on the local machine is different. With the above mentioned ssh_config file I was able to write playbooks without need to specify ansible_ssh_user. Now it is impossible.

thomas-mangin:
I now understand. I believe the patch as I wrote it would work for this use case ... 
@jimi-c - could you please see apply the patch as I proposed it as I believe it not display the side behaviour reported by @fly-me

jimi-c:
@thomas-mangin no, I believe it would, since specifically setting `-o User` would override the config. I believe we will need to roll this one back.

bcoca:
yes, I did the reverse patch a looong time ago to enable using .ssh/config
by not specifying the user when it matched the run user.​

Issue #6809, opened 2014-04-01
rsync error going from Ubuntu < 14.04 to 14.04


simonjgreen:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.6
##### Environment:

Ubuntu 12.04 and Ubuntu 14.04
##### Summary:

Synchronize module does not work going TO Ubuntu 14.04 from anything older than Ubuntu 14.04
##### Steps To Reproduce:

templates/dotfiles contains around 350 files. This does not reproduce when copying, say, 1 file.

simon@ubuntu:~/BuildTools/ansible$ ansible all -m synchronize -a "src=templates/dotfiles dest=/tmp/test" -i hosts/simonTest -u wirehive -k
SSH password:
wirehive@10.10.10.159's password:
10.10.10.159 | FAILED >> {
    "cmd": "rsync --delay-updates -FF --compress --timeout=10 --archive --rsh 'ssh  -o StrictHostKeyChecking=no' --out-format='<<CHANGED>>%i %n%L' templates/dotfiles wirehive@10.10.10.159:/tmp/test",
    "failed": true,
    "msg": "unknown message 0:0 [sender]\nrsync error: error in rsync protocol data stream (code 12) at io.c(441) [sender=3.0.9]\n",
    "rc": 12
}
##### Expected Results:

Should copy!
##### Actual Results:

rsync error: error in rsync protocol data stream (code 12) at io.c(441) [sender=3.0.9]

mpdehaan:
See http://rsync.samba.org/issues.html and search for io.c

Ultimately I'm not sure this is our ticket since the system is calling out to rsync.

shiinee:
I'm running into the same issue and the culprit seems to be the `--timeout=10` option added by Ansible.

**Run rsync command using playbook:**

```
$ ansible-playbook -vv systemd.yml
PLAY [all] ********************************************************************

GATHERING FACTS ***************************************************************
<10.252.105.160> REMOTE_MODULE setup
ok: [10.252.105.160]

TASK: [copy source files] *****************************************************
failed: [10.252.105.160] => {"cmd": "rsync --delay-updates -FF --compress --timeout=10 --archive --rsh 'ssh  -o StrictHostKeyChecking=no' --out-format='<<CHANGED>>%i %n%L' /Users/skyler/systemdtest/../le_codez/systemd/ root@10.252.105.160:/root/systemd/", "failed": true, "rc": 12}
msg: unknown message 0:0 [sender]
[...]

PR #7173, opened 2014-04-28
Remove --timeout option to rsync from synchronize defaults

shiinee:
The current 10-second default timeout for rsync seems to be behind issue #6809.

As a workaround for the underlying issue in rsync when different versions are
used on the source and destination sides, don't include the timeout option
unless the user specifies a timeout > 0.


jimi-c:
Merged, thanks!

PR #7175, opened 2014-04-28
Add play_hosts to on_play_start callback

hkariti:
Saves the duplicity of calling inventory.list_hosts() twice (once in the callback, once outside)

### DIFF
--- a/lib/ansible/playbook/__init__.py
+++ b/lib/ansible/playbook/__init__.py
@@ -621,13 +621,14 @@ class PlayBook(object):
     def _run_play(self, play):
         ''' run a list of tasks for a given pattern, in order '''
         
-        self.callbacks.on_play_start(play.name)
         # Get the hosts for this play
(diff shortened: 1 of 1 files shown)


mpdehaan:
# Revision Request

Hi!

- this looks solid, but it appears that if no hosts are matched the "on play start" callback will happen after, which gets our events out of order
- be sure to call the on play start" callback before signalling "no hosts matched"

which is very active:

hkariti:
on_plat_start is now called before the on_no_hosts_matched test

amenonsen:
I looked at these commits. The code in devel (a) has changed enough that the diff would need a rewrite, not just a revision, but (b) the v2 code already seems to do what the patch intended (i.e. call list_hosts via get_serialized_batches before sending the on_play_start callback).

Therefore this PR should be closed now.

bcoca:
agreed, current devel already has this behavior, closing

PR #7176, opened 2014-04-28
Arch PKGBUILD fixes : removes man3 section and adds contributor list

leucos:
Seems that there are no man for section 3 anymore. Removed install snippet from PKGBUILD.
Added list of PKGBUILD contributors in header.


jimi-c:
Merged, thanks!

PR #7177, opened 2014-04-28
Do not fail if we try to generate a ssh key for a user without being root

mscherer:
If we try to use the user module without being root, it fail on RHEL/Fedora
because usermod --help cannot be run. The root cause is lack of permission
due to EAL4+ certification, as seen in shadow-utils changelo.

So if we cannot run it, assume there is no append. It doesn't matter
much since we will not be able to run usermod at all with or without the
option.


jimi-c:
Merged, thanks!

PR #7183, opened 2014-04-28
use get_bin_path to get the path to a2dismod

berendt:


berendt:
This one can be closed. Not longer necessary after merge of #7191.

PR #7187, opened 2014-04-28
use a2query instead of a2enmod to check if a module is enabled

berendt:


berendt:
Revoking. a2query is not available with all Apache distributions.

PR #7191, opened 2014-04-28
made enabled/disabled checks of apache2_module workable

berendt:
At the moment the check if a module is enabled or disabled is broken because a2enmod always enables a module and returns 0 if the module is already enabled or it got successfully enabled. This patch should solve this issue.

### DIFF
--- a/library/web_infrastructure/apache2_module
+++ b/library/web_infrastructure/apache2_module
@@ -1,7 +1,7 @@
 #!/usr/bin/python
 #coding: utf-8 -*-
 
-# (c) 2013, Christian Berendt <berendt@b1-systems.de>
+# (c) 2013-2014, Christian Berendt <berendt@b1-systems.de>
 #
 # This module is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


kiranos:
can this be merged before 1.6 is freezed? the module is already included.

jimi-c:
@kiranos have you tested this in your environment? +1's from other testers are always helpful in moving things forward.

geerlingguy:
I don't have the time to manually test the patch right now, but I just installed 1.6 and found the inability for the apache2_module module to show changed/unchanged status slightly jarring (though I can confirm the basic functionality works for me).

My task:

``` yaml
- name: "Apache: Enable mod_rewrite."
  apache2_module: name=rewrite state=present
  notify: restart apache
```

It seems to work, but won't `restart apache` because it never knows if the module was changed. (It always reports `ok: [default]`).

jimi-c:
Yes sorry, I was hoping to get this in early in the morning since we had planned to do the release in the afternoon, so we just ran out of time before the release. We'll look to get this included if/when we release a 1.6.1.

kiranos:
+1 I've done some more testing now and this works:

changed: [IP] => (item=rewrite)
changed: [IP] => (item=headers)
changed: [IP] => (item=expires)

---

< NOTIFIED: restart apache2 >

so when changed both from present and from absent it does trigger the notify command.

jimi-c:
Merged, thanks!

PR #7192, opened 2014-04-28
Add validate_certs option to apt_repository

hosiawak:
apt_repository is missing validate_certs option

It's failing for me for the following example:

```
apt_repository: repo='ppa:brightbox/ruby-ng' update_cache=yes
```

### DIFF
--- a/library/packaging/apt_repository
+++ b/library/packaging/apt_repository
@@ -55,6 +55,13 @@ options:
         required: false
         default: "yes"
         choices: [ "yes", "no" ]
+    validate_certs:
+        description:
+            - If C(no), SSL certificates for the target repo will not be validated. This should only be used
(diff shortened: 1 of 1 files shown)


sivel:
This was actually omitted on purpose.  Originally this module used pycurl, and had certificate validation forced.  Since we were forcing SSL cert validation in previous versions, we had left it as forced without the option to disable.

Not that this couldn't be changed, just providing background as to why is was not included.

jimi-c:
I'm fine with adding this in, for those who may be running an internal mirror without a valid SSL cert.

mpdehaan:
Merged, thank you!

I added the 'version_added' attribute so it was clear this is now available as an option in 1.8 and later.

PR #7196, opened 2014-04-28
New files inaccessible (0600)

whiskybar:
##### Issue Type:

Bugfix Pull Request
##### Ansible Version:

ansible 1.5.4
##### Environment:

Ubuntu 12.04, all
##### Summary:

Following 898a38b074f2561d19059315030b2b5da68a4f34, all newly created files are 0600 but should be 0644 as before (i.e. by umask)
##### Steps To Reproduce:

```
echo HOSTNAME >/tmp/hosts
ansible HOSTNAME -i /tmp/hosts -m copy -a 'content=hello dest=/tmp/1'
```
##### Expected Results:

```
"mode": "0644"
```

usually and driven by umask. This was in Ansible before the commit.
##### Actual Results:

```
 "mode": "0600"
```

driven by the `tempfile.NamedTemporaryFile` behavior: making a temporary file private for the creator. This behavior after the commit did and will break previous playbooks for many users.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -1056,6 +1056,11 @@ class AnsibleModule(object):
         if creating and os.getenv("SUDO_USER"):
             os.chown(dest, os.getuid(), os.getgid())
 
+        if creating:
+            umask = os.umask(0)
+            os.umask(umask)
+            os.chmod(dest, 0666 ^ umask)
+
         if self.selinux_enabled():
             # rename might not preserve context
             self.set_context_if_different(dest, context, False)


bcoca:
doesn't this defeat the purpose (security?) of using tmpfile?​

jimi-c:
Are all files affected by this (ie. using the file/copy/template modules) or is this something solely affecting temporary files?

bcoca:
No, I misread what the patch was doing. Looking at the full code, it just
makes sure that the permissions on the final file follow the expected mask,
only when the file did not exist.

Ignore my previous comments

bcoca:
looking at it a bit longer, a better approach might be setting dest_stat
when the file doesn't exist, it would make the function shorter and more
readable.

If no one beats me to it I'll probably have a PR by the weekend.

## 

Brian Coca
Pedo mellon a minno

jimi-c:
I've gone ahead and implemented this in a slightly different way, though I still gave credit for the patch to @whiskybar since it was his code (just slightly rearanged). Doing it via the dest_stat seemed just as complicated, so I went with this instead. Thanks!

PR #7208, opened 2014-04-29
Fix "no_log=True" option for copy module (fixes #7193)

### LINKED ISSUE #7193: 'copy' module logs despite no_log: true
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.6 (devel a482e1583d) last updated 2014/04/29 01:53:48 (GMT +800)
##### Environment:

Ubuntu 12.04.4
##### Summary:

'copy' module logs on the remote despite no_log: true. This is not the case for other modules (service, command, shell). I would submit a patch, except that I'm still figuring out the source.
##### Steps To Reproduce:

Create a play with a copy task, and set no_log: true. Run the play, and you'll see that the copy task is still logged on the remote.
##### Expected Results:

No log on the remote for the copy task.
##### Actual Results:

The copy task is logged on the remote.

jf:


jimi-c:
Merged, thanks!

PR #7212, opened 2014-04-29
Make sure a default allow out rule exists if no other egress rules do

### LINKED ISSUE #7027: Default to sane egress policy in ec2_group (change since 1.5.3->1.6)
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.6
##### Environment:

Mac OS X 10.7
##### Summary:

rules_egress was added to ec2_group in 1.6 (previous there was no rules_egress).  Unfortunately it defaults to nothing in it, meaning that previous ansible scripts will not allow anything out of a security group.  This breaks many security groups.
##### Steps To Reproduce:

Create a ec2_group (security group) with no rules_egress.  Watch everything get blocked going outbound.
##### Expected Results:

Should default to allowing everything outbound since thats what ansible did previously.
##### Actual Results:

Unable to access going outbound.

jimi-c:
Fixes #7027

### DIFF
--- a/library/cloud/ec2_group
+++ b/library/cloud/ec2_group
@@ -313,6 +313,20 @@ def main():
                                 src_group_id=grantGroup,
                                 cidr_ip=ip)
                     changed = True
+        elif vpc_id and not module.check_mode:
+            # when using a vpc, but no egress rules are specified, 
+            # we add in a default allow all out rule, which was the
+            # default behavior before egress rules were added
+            if 'out--1-None-None-None-0.0.0.0/0' not in groupRules:
+                ec2.authorize_security_group_egress(
+                    group_id=group.id,
+                    ip_protocol=-1,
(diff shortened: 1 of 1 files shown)


mmoya:
I **strongly disagree** with this. See explanation [here](https://github.com/ansible/ansible/issues/7027#issuecomment-41704388).

jimi-c:
@mmoya this patch should only be hit if the egress rules are not specified at all, it will not add a default 0/0 out rule if other egress rules are specified. From my understanding, this was how it worked before but with the addition of the egress rules it's preventing any outbound traffic by default.

mmoya:
You're right. I didn't put enough attention when reviewing this PR.

jimi-c:
@mmoya no problem, let us know if you have any further questions.

PR #7215, opened 2014-04-29
Document how to turn off ec2 cache

lorin:
Document how to disable cache lookup with the ec2 inventory script


jimi-c:
Merged, thanks!

PR #7217, opened 2014-04-29
If overwrite is set, don't bother doing the checksum.

feanil:


amenonsen:
I looked at the diff, and it should be trivial to fix the conflicts. But this PR should be re-targeted against the ansible-modules-core repository (cloud/amazon/s3.py).

feanil:
Closing this: I've opened it against the ansible-modules-core repo: https://github.com/ansible/ansible-modules-core/pull/1795

PR #7220, opened 2014-04-29
Improve error messages of mysql modules when connection fails

abeverley:
Mysql connections can fail for a variety of reasons, not just bad credentials. In my case, the database didn't exist. This patch displays the actual mysql error message to aid debugging.

### DIFF
--- a/library/database/mysql_db
+++ b/library/database/mysql_db
@@ -279,7 +279,7 @@ def main():
             db_connection = MySQLdb.connect(host=module.params["login_host"], port=int(module.params["login_port"]), user=login_user, passwd=login_password, db=connect_to_db)
         cursor = db_connection.cursor()
     except Exception, e:
-        module.fail_json(msg="unable to connect, check login_user and login_password are correct, or alternatively check ~/.my.cnf contains credentials")
+        module.fail_json(msg="error %d connecting to database: %s, check login_user and login_password are correct, or alternatively check ~/.my.cnf contains credentials" % (e.args[0],e.args[1]))
 
     changed = False
     if db_exists(cursor, db):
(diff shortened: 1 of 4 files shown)


hoxu:
I ended up doing a similar change too. The current error message is very misleading.

abeverley:
I see that the source code has now been updated to include a warning in the event of an unknown database, and therefore this patch no longer merges cleanly.

However, the new changes still don't account for any _other_ error, which will still be reported without the actual error message, but instead with the same generic message about checking the username and password.

I am happy to provide another patch, but it would be useful for the maintainer to comment on this before I spend time doing so. Thanks.

Jmainguy:
I personally think the more error messages the better, to help isolate the issue for the end user.

abeverley:
Passing the actual error message from the SQL server to the user gives the exact reason as to the problem. As it is, the very specific error message from the database is translated to "please check your username and password" or "database does not exist" - not very helpful if the error doesn't happen to be either of those problems. There seems very little point in taking a very accurate error message from a server and then translating it to something vague. Why not just pass the details straight through?

Jmainguy:
I have been focusing on mysql issue reports lately and a few of them were about error messages. Check this one out for if socket does not exist. #8542. I believe there already is a PR approved and in the module for unknown database. I believe passing the mysql error message is beneficial in some cases, but of course it does not cover all cases.

abeverley:
The mysql error message should be displayed if it has been possible to talk to mysql and it has returned an error message. Of course, if it's not possible to even get that far, then an error message stating as much should be given instead.

In short: if mysql gives an error message, then please display that error message.

mpdehaan:
Hi!

Thanks you very much!

Issue #6025, opened 2014-02-15
Feature request: Linode add private ip address


bsdlp:
##### Issue Type:

Feature Idea
##### Ansible Version:

ansible 1.4.5
##### Environment:

N/A
##### Summary:

The Linode cloud module is currently missing the ability to add private IP addresses to an instance. This is important for the NodeBalancer product, which requires Linodes to be added using the Linode's private IPv4 address.

https://www.linode.com/api/linode/linode.ip.addprivate
##### Steps To Reproduce:

N/A
##### Expected Results:

Adds private IP address after Linode is created, but before it is booted.
##### Actual Results:

N/A

mpdehaan:
Sorry for the late response on this as we've been working our way through the queue.

We've been mostly concentrating on extending things like EC2 and Rackspace, but obviously Linode is good stuff too -- is this something you would like to work on?

If so, we are open to pull requests.  

Thanks!

pilwon:
@mpdehaan @fly Created PR #7223 for this issue. Please review & merge.

mpdehaan:
Closing this ticket since pull request above was submitted, thank you very much!

Issue #5992, opened 2014-02-13
Linode module: Add private IP


mahemoff:
##### Issue Type:

Feature request
##### Ansible Version:

1.4.4
##### Environment:

Ubuntu 12.04
##### Summary:

An option to add private IP would be great (it's possible in Linode's API - https://www.linode.com/api/linode/linode.ip.addprivate). This is rather cumbersome to do manually as it requires a reboot.
##### Steps To Reproduce:

ansible localhost -m linode -a "private_ip=true (other-args)"
##### Expected Results:

Linode has a private IP and has been rebooted to ensure it's active.
##### Actual Results:

Linode has no private IP. It must be created manually in Linode console.

mpdehaan:
This sounds like it would be a nice idea, but it is not likely something the project admins are going to implement.

Is this something you would like to work on?

If not, perhaps it is something you should bring up on the mailing list to see if someone would like to implement.

Thanks!

mahemoff:
Understand. As I'm time-constrained now, here's a simple Gist of how I got this and #5991 working - https://gist.github.com/mahemoff/8983846

I might mail the list or put a proper patch in later.

pilwon:
@mpdehaan @mahemoff Created PR #7223 for this issue. Please review & merge.

aleemb:
+1

arturmartins:
I actually like the prefix Linode ID. I would recommend that both (Prefix and Private IP) should be active by configuration (example: add_prefix: false, add_private_ip: true)

Issue #5991, opened 2014-02-13
Linode module: Don't prepend Linode ID


mahemoff:
##### Issue Type:

Feature request
##### Ansible Version:

1.4.4
##### Environment:

Ubuntu 12.04
##### Summary:

The [doc](http://docs.ansible.com/linode_module.html) says "To keep sanity on the Linode Web Console, name is prepended with LinodeID"., but I have a logical naming scheme which is interrupted by the arbitrary Linode IDs.

I think this feature should at very least be optional.
##### Steps To Reproduce:

ansible localhost -m linode -a "name=web (other-ags ..."
##### Expected Results:

Linode created is called "web" in Linode console.
##### Actual Results:

Linode created is called "1234_web" in Linode console, where "1234" is the Linode node ID.

mahemoff:
The ID's prepended at https://github.com/ansible/ansible/blob/devel/library/cloud/linode#L252

mpdehaan:
Can we start a discussion on this on ansible-project?

I'm open to the idea but wouldn't want to break anyone or remove soemthing some folks were liking.

Thanks!

mpdehaan:
Also, this is is unlikely to be something we are going to personally work on. Should you wish to do this, the best thing to do would be to submit a pull request.

For discussion topics, we prefer the list as there are more people reading the list (several thousands) versus github history (a couple).

Thanks!

pilwon:
@mpdehaan @mahemoff Created PR #7223 for this issue. Please review & merge.

PR #7223, opened 2014-04-30
Linode: Don't prepend Linode ID & Add private IP

pilwon:
@mahemoff and @fly suggested improvements for `Linode` module.
- Don't prepend Linode ID (#5991)
- Add private IP (#5992, #6025)

### DIFF
--- a/library/cloud/linode
+++ b/library/cloud/linode
@@ -34,7 +34,6 @@ options:
   name:
     description:
      - Name to give the instance (alphanumeric, dashes, underscore)
-     - To keep sanity on the Linode Web Console, name is prepended with LinodeID_
     default: null
     type: string
   linode_id:
@@ -249,7 +248,8 @@ def linodeServers(module, api, state, name, plan, distribution, datacenter, lino
                                         PaymentTerm=payment_term)
                 linode_id = res['LinodeID']
                 # Update linode Label to match name
-                api.linode_update(LinodeId=linode_id, Label='%s_%s' % (linode_id, name))
+                api.linode_update(LinodeId=linode_id, Label=name)
+                api.linode_ip_addprivate(LinodeId=linode_id)
                 # Save server
                 servers = api.linode_list(LinodeId=linode_id)
             except Exception, e:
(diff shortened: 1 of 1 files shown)


library/cloud/linode
```diff
@@ -249,7 +248,8 @@ def linodeServers(module, api, state, name, plan, distribution, datacenter, lino
                                         PaymentTerm=payment_term)
                 linode_id = res['LinodeID']
                 # Update linode Label to match name
-                api.linode_update(LinodeId=linode_id, Label='%s_%s' % (linode_id, name))
+                api.linode_update(LinodeId=linode_id, Label=name)
+                api.linode_ip_addprivate(LinodeId=linode_id)
```
bsdlp: If I'm reading this correctly, this adds a private IP address to Linodes by default, which is a bad idea.

Linodes with only the public IPv4 + auto-assigned IPv6 addresses will have working network connectivity out of the box via DHCP and RA, but once you have a private IP address, you will need to use static network configuration.

IMO, it would be optimal for this to be an option, but disabled by default.

pilwon: I agree, and updated PR #7223 accordingly.

pilwon:
In response to @fly's [comment](https://github.com/ansible/ansible/pull/7223#discussion-diff-12126253), I updated PR to add `private_networking` option & disable private IP by default.

dvic:
Any updates on when/if this is going to be merged? Thanks.

mpdehaan:
Hi!

Thank you very much!

meshy:
Are there any plans for updating this so that it can be merged?

qrkourier:
resurrecting this in https://github.com/ansible/ansible/pull/28947/

Routhinator:
Did this make it in? I would prefer to prevent the module from adding the ID to the name so I don't have to write ID's to the inventory after the creation stage. The names should already be unique without adding IDs.

qrkourier:
@Routhinator Doesn't look like it, and in the intervening months quite a few conflicting changes *were* accepted which makes this laborious to reconcile.

Issue #7218, opened 2014-04-29
Error validating launchpad.net SSL certificate


jcassee:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.6
(commit 71a870a6be16c7d94debd114f60804ea2ba9ca15)
##### Environment:

Running Ansible from Ubuntu 12.04. Managed system is Ubuntu 14.04.
##### Summary:

Module `apt_repository` fails to validate launchpad.net SSL certificate.

(See also: https://groups.google.com/forum/#!topic/ansible-project/lmAp8ui0JEc)
##### Steps To Reproduce:

Run this playbook with any inventory:

```

---
- hosts: all
  gather_facts: no
  tasks:
  - apt_repository: repo='ppa:webupd8team/java'
```
##### Expected Results:

The Apt repository should have been added to `/etc/apt/sources.list.d`.
##### Actual Results:

Running the playbook gives the following error:

```
PLAY [all] ******************************************************************** 

TASK: [apt_repository repo='ppa:webupd8team/java'] **************************** 
failed: [XXX] => {"failed": true}
msg: Failed to validate the SSL certificate for launchpad.net:443. Use validate_certs=no or make sure your managed systems have a valid CA certificate installed. Paths checked for this platform: /etc/ssl/certs, /etc/pki/ca-trust/extracted/pem, /etc/pki/tls/certs, /usr/share/ca-certificates/cacert.org, /etc/ansible

FATAL: all hosts have already failed -- aborting

PLAY RECAP ******************************************************************** 
           to retry, use: --limit @/home/joost/test.retry

XXX        : ok=0    changed=0    unreachable=0    failed=1
```
##### Extra Information:

I extracted the temporary ca_certs file that `urls.py` feeds to `ssl.wrap_socket` and used `gnutls-cli` to check whether it contains the root certificate for launchpad.net. It verified correctly:

```
$ gnutls-cli --x509cafile tmp_ca_certs.pem launchpad.net
Processed 332 CA certificate(s).
[...]
 - Certificate[0] info:
  - subject `OU=Domain Control Validated,CN=launchpad.net', issuer `C=US,ST=Arizona,L=Scottsdale,O=GoDaddy.com\, Inc.,OU=http://cer
ts.godaddy.com/repository/,CN=Go Daddy Secure Certificate Authority - G2', RSA key 2048 bits, signed using RSA-SHA256, activated `2
[...]

PR #7236, opened 2014-04-30
Make concatenating certs robust in urls.py

jcassee:
Add a newline after each certificate file explicitly to avoid problems with files that do not end with a newline themselves.

Fixes issue #7218.


jimi-c:
Merged, thanks!

PR #7238, opened 2014-05-01
Improve the ec2_snapshot argument spec and docs

willthames:
ec2_snapshot got missed when moving to a common argument spec.
It could already make use of the capabilities (as it uses
ec2_connect) and the documentation suggested it supported the
common argument spec (thanks to the documentation fragment work)
so it was just a matter of fixing the argument spec.

Removed unnecessary documentation for profile and security_token
that is covered by documentation fragment


jimi-c:
Merged, thanks!

PR #7249, opened 2014-05-01
Allow group module to handle check_mode properly when group.state=present and group already exists

jfchevrette:
The Group.group_mod function always returns (True, '', '') [1] when check_mode is True which is in turn failing the task when requesting group.state = present and group already exists [2]

I compared both user and group modules which are working in a similar way and the proposed fix handles this case much like the user module does.

[1] https://github.com/ansible/ansible/blob/devel/library/system/group#L123-L124
[2] https://github.com/ansible/ansible/blob/devel/library/system/group#L123-L124

### DIFF
--- a/library/system/group
+++ b/library/system/group
@@ -121,7 +121,7 @@ class Group(object):
         if len(cmd) == 1:
             return (None, '', '')
         if self.module.check_mode:
-            return (True, '', '')
+	    return (0, '', '')
         cmd.append(self.name)
         return self.execute_command(cmd)
 


jfchevrette:
I don't have corresponding systems to test but it appears as if the other classes group_mod function should be changed in the same fashion (i.e.: returning 0 instead of True) when check_mode is set.
- https://github.com/ansible/ansible/blob/devel/library/system/group#L209
- https://github.com/ansible/ansible/blob/devel/library/system/group#L248
- https://github.com/ansible/ansible/blob/devel/library/system/group#L290
- https://github.com/ansible/ansible/blob/devel/library/system/group#L332

siboulet:
+1
Without this patch, running a playbook in Check mode will fail if the group is not present:

TASK: [some-task | ensure some-group group is present] *******************\* 
failed: [some-host1] => {"failed": true, "item": "", "name": "some-group"}
failed: [some-host2] => {"failed": true, "item": "", "name": "some-group"}

siboulet:
Any chance this can make it for the next release? Appears to be a very simple fix.

jimi-c:
Merged, thanks!

jimi-c:
FYI, I've also made the similar change for the other group class objects.

jfchevrette:
Awesome thanks James!

PR #7255, opened 2014-05-02
simple fix for allowing zero length sequences

darKoram:
Currently, there is no way to have a zero count with_sequence

```
 - name: Test count = 0
   debug: msg='hello'
   with_sequence: count=0
```

This fails with "can't count backwards"
     - name: Test start = end 
       debug: msg='hello'
       with_sequence: start=1 end=1
This runs the action once

see gist for tests: https://gist.github.com/darKoram/baa4c6e2a025b48d67ae

Simplest fix is the easiest to code and makes the start, end semantics consistent with python 'range' command
range(start,end) produces [start,end), that is, from start to, but not including, end.

The downside is that this will break the most code.

Behavior changes
1. with_sequence: start=1 end=1 previously ran the action once, now it skips
2. with_sequence: start=1 end=2 previously ran the action twice, now it runs once
3. with_sequence: count=1 previously ran the action once, now it still runs once
4. with_sequence: count=0 previously errored with "can't count backwards", now it skips

The other options include leaving the semantics of start,end as is, but changing count to accept 0. This will require less clean code but would leave 1 and 2 unchanged breaking less code (but keeping the inconsistency with python semantics).  If we leave start, end as is then the next simplest fix would allow end = start-1 to represent a zero count.  Kinda hooky.

### DIFF
--- a/lib/ansible/runner/lookup_plugins/sequence.py
+++ b/lib/ansible/runner/lookup_plugins/sequence.py
@@ -151,7 +151,7 @@ class LookupModule(object):
             )
         elif self.count is not None:
             # convert count to end
-            self.end = self.start + self.count * self.stride - 1
+            self.end = self.start + self.count * self.stride
             del self.count
         if self.end < self.start:
             raise AnsibleError("can't count backwards")
@@ -159,7 +159,7 @@ class LookupModule(object):
             raise AnsibleError("bad formatting string: %s" % self.format)
 
     def generate_sequence(self):
-        numbers = xrange(self.start, self.end + 1, self.stride)
+        numbers = xrange(self.start, self.end, self.stride)
 
         for i in numbers:
             try:


jimi-c:
What useful purpose does this serve? I don't understand why you'd want to iterate over a zero-length sequence. Also, I do not like that this changes the behavior of certain ranges, which will almost certainly cause problems for others playbooks.

darKoram:
Apologies for not giving more context with the pull request.
It comes from this thread
https://groups.google.com/forum/#!msg/ansible-project/swkbKGLio1g/uHmckDsmff0J

At the end, Michael suggests he would consider a pull request for this feature.

The use case is whenever whith_sequence is controlling an action with side-effects.
In my case, ensuring there are N vms or floating ips when some may already exist.
On a second run, if we already have N items, there is no way to skip the action without creating a minimum of 1 new thing.

kesten

stibi:
Hi, thanks for this PR, I have the same problem. I have to set count=1 as a workaround for tasks which are skipped, but otherwise would fail because count=0.

amenonsen:
I looked at this PR and the original thread where it was discussed. I understand the need for some way to address the original use case, but I don't really like the changed behaviour in other cases, and it's certainly completely unacceptable without any sort of documentation.

bcoca:
This is not problem anymore in current devel, as sequence is a lot more flexible now.

here is the output of the included tests for this case:

```
skipping: [testhost] => {"changed": false, "results": [], "skipped": true, "skipped_reason": "No items in the list"}
```

and the test case:

```
- name: test with_sequence not failing on count == 0
  debug: msg='previously failed with backward counting error'
  with_sequence: count=0

```

PR #7260, opened 2014-05-02
Extend wait_for module to support draining of active tcp connections

rackergs:
note: I know it needs to be rebased and squashed, but i'd rather only do that once any potential changes are made post review 

In our environment when we are doing updates we typically pull a node out of the load balancer pool but can't actually continue without at least a reasonable wait for active connections to close.  To this end I implemented a state of 'drained' that checks for tcp connections (by parsing /proc/net/tcp) that match the defined host:port combination.

The load balancers do continue performing health checks, so the need to ignore certain connections was needed, thus adding the parameter exclude_hosts.

Example:

```
# wait 300 seconds for port 8000 of any IP to close active connections,
# don't start checking for 10 seconds
- wait_for: host=0.0.0.0 port=8000 delay=10 state=drained

# Same as above but excluding the firewalls
- wait_for: host=0.0.0.0 port=8000 delay=10 state=drained
      exclude_hosts:10.10.10.15,10.10.10.16
```

### DIFF
--- a/library/utilities/wait_for
+++ b/library/utilities/wait_for
@@ -23,6 +23,15 @@ import datetime
 import time
 import sys
 import re
+import binascii
+
+HAS_PSUTIL = False
+try:
+    import psutil
+    HAS_PSUTIL = True
+    # just because we can import it on Linux doesn't mean we will use it
+except ImportError:
+    pass
 
 DOCUMENTATION = '''
 ---
@@ -38,6 +47,9 @@ description:
      - In 1.6 and later, this module can 
        also be used to wait for a file to be available or absent on the 
        filesystem.
+     - In X and later, this module can also be used to wait for active
+       connections to be closed before continuing, useful if a node
+       is being rotated out of a load balancer pool.
 version_added: "0.7"
 options:
   host:
@@ -62,10 +74,10 @@ options:
     required: false
   state:
     description:
-      - either C(present), C(started), or C(stopped), C(absent)
-      - When checking a port C(started) will ensure the port is open, C(stopped) will check that it is closed
+      - either C(present), C(started), or C(stopped), C(absent), or C(drained)
+      - When checking a port C(started) will ensure the port is open, C(stopped) will check that it is closed, C(drained) will check for active connections
       - When checking for a file or a search string C(present) or C(started) will ensure that the file or string is present before continuing, C(absent) will check that file is absent or removed
-    choices: [ "present", "started", "stopped", "absent" ]
+    choices: [ "present", "started", "stopped", "absent", "drained" ]
     default: "started"
   path:
     version_added: "1.4"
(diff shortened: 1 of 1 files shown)


jimi-c:
I like this idea, however I'd like to make sure it has some cross-platform support (BSDs, etc).

rackergs:
Any pointers of supporting this for BSD? I'm not versed in its ways, and from poking at the freebsd box I just built I see there is no /proc/net/tcp\* equivalent

rackergs:
Found: http://psutil.googlecode.com/hg-history/ab4a1f9dc57cabef4834ab4fb00c5e267acd3112/examples/netstat.py

And then noticed how user and group implement platform strategies... i'll poke at this but am in the middle of a move this coming week so we'll see.

gregswift:
so i got kinda far along in implementing the platform strategy concept (didn't push it yet) where linux uses /proc/net/tcp directly and freebsd uses the psutil library.

@jimi-c should I just require the psutil library on linux too to (which should in theory resimplify the extension)? or leave linux without the additional library dependency?

jimi-c:
I think leaving it without a dependency is more desirable for sure. It just needs to be split out into the platform-style as you noted (see service, user, etc. for examples) and use fail_json on BSD if psutil isn't installed.

rackergs:
okay. pushed some genreal bits. Don't think its functional yet (could be wrong) but i'd be interested in some 'you are doing it wrong' feedback about the xplatform bits.

rackergs:
@jimi-c okay.. so i set it up so that the generic implementation uses psutil, which should in theory work everywhere.  Then I placed the procfs in the Linux specific implementation.  Hostname seems overly complex in comparison to Service, which I had looked at it first.

jimi-c:
@rackergs great, this will definitely be in-queue for testing. Thanks!

gregswift:
@jimi-c i was about to poke at it a bit more just to make sure i was happy and i noticed that in LinuxTCPConnectionInfo.get_active_connection_count() it returns a 0 if the file can't be read.  I had moved this logic from the original while loop down on 418.

The two approaches I can think of:

1: remove the try catch from the class and put it back in the while loop
2: return a number >0 when the exception is caught

thoughts?

gregswift:
Went with the try catch in the while loop.  seems to match other behaviors more after i thought about it.

Rebased/squashed.

rackergs:
@jimi-c do i need to rebase/squash this again? is there a specific time when i should do that?

jimi-c:
Nah, don't worry about it, I can sort it out.

gregswift:
Found a bug in the exclude_hosts handling. Should be better now.

jimi-c:
@gregswift could you rebase this to get rid of the merge commit?

gregswift:
@jimi-c done

jimi-c:
Thanks!

gregswift:
@jimi-c so, generally I'm trying to be patient about this PR, but its currently delaying our ability to move to ansible for a bunch of things.  I'm currently working on a work around, which I'd rather not do obviously.  Any thing in particular that could be done to get a target ?

jimi-c:
# Merged

Hi!

Thank you!

PR #7263, opened 2014-05-03
[lvg] Switch to mapper device name if pvs returns the wrong one.

ajsalminen:
The lvg module fails after the initial run if the pvs parameter includes device-mapper logical volumes on some older distributions such as Debian Squeeze and Ubuntu 12.04. This issue is encountered when using LVM on LUKS for example.

The issue is caused because on these distro versions "pvs" returns the non-persistent device names like /dev/dm-0 for device-mapper logical volumes. This makes lvg to think the device has not yet been added to the volume group/initialized while in reality it has.

The proposed patch would use dmsetup to find out what the device name is when encountering /dev/dm- prefixed device names in the list from "pvs". This will allow using /dev/mapper/name  but not /dev/vg/name in the pvs parameter.

### DIFF
--- a/library/system/lvg
+++ b/library/system/lvg
@@ -67,7 +67,7 @@ EXAMPLES = '''
 # Create a volume group on top of /dev/sda1 with physical extent size = 32MB.
 - lvg:  vg=vg.services pvs=/dev/sda1 pesize=32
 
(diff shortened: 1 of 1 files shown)


jimi-c:
Merged, thanks!

Issue #6467, opened 2014-03-13
ask_sudo_pass in the config is causing --sudo 


makimotoz:
##### Issue Type:

Bug report 
##### Ansible Version:

ansible 1.5.2
##### Environment:

Master centos 6.4. Manages rhel 5.5
##### Summary:

When using ask_sudo_pass in the config ("ask_sudo_pass  = True") I get auth issues in _some_ boxes.
Once the directive is set to false in the cfg file, that works without issues.
##### Steps To Reproduce:

set ask_sudo_pass  = True in the .cfg file and do any task that does not require actual sudo access on the remote box
##### Expected Results:

no auth issues
##### Actual Results:

```
$ ansible commando -m ping -K
sudo password:
previous known host file not found
commando | FAILED => Authentication or permission failure.  In some cases, you may have been able to authenticate and did not have permissions on the remote directory. Consider changing the remote temp path in ansible.cfg to a path rooted in "/tmp". Failed command was: mkdir -p /tmp/ansible-1392202978.96-158385720428034 && chmod a+rx /tmp/ansible-1392202978.96-158385720428034 && echo /tmp/ansible-1392202978.96-158385720428034, exited with result 1: mkdir: cannot create directory `/tmp/ansible-1392202978.96-158385720428034':
```

jctanner:
Confirming this is the wrong behavior per 506178542d04a15647467e4f055bfbf8ade1c70e

Reproduced on devel:

```
jtanner@u1204:~/issues/6467-sudo-broken$ cat inventory 
el6.lab.net
jtanner@u1204:~/issues/6467-sudo-broken$ ansible -vvvv el6.lab.net -i inventory -m ping
sudo password: 
<el6.lab.net> ESTABLISH CONNECTION FOR USER: jtanner
<el6.lab.net> REMOTE_MODULE ping
[...]

PR #7264, opened 2014-05-03
Make --ask-sudo-pass/-K imply --sudo for `ansible`

jf:
As per subject: make the `--ask-sudo-pass` / `-K` imply `--sudo` for `ansible`. I think this is only natural for `ansible`; and having it not be so would be kind of meaningless when you ask for a password, but never act on it as in the case when you give `--ask-sudo-pass`, but miss out on `--sudo`.

### DIFF
--- a/bin/ansible
+++ b/bin/ansible
@@ -63,6 +63,8 @@ class Cli(object):
             default=C.DEFAULT_MODULE_NAME)
 
         options, args = parser.parse_args()
+        if options.ask_sudo_pass:
+            options.sudo = True
         self.callbacks.options = options
 
         if len(args) == 0 or len(args) > 1:
--- a/bin/ansible
+++ b/bin/ansible
@@ -63,8 +63,6 @@ class Cli(object):
             default=C.DEFAULT_MODULE_NAME)
 
         options, args = parser.parse_args()
-        if options.ask_sudo_pass:
-            options.sudo = True
         self.callbacks.options = options
 
         if len(args) == 0 or len(args) > 1:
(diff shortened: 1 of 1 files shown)


jf:
(And yes, I'm aware that the documentation will have to be changed if this patch makes it in. I will work on the manpage if this patch/change is accepted.)

sivel:
This was actually changed somewhat recently to NOT imply sudo based on supplying --ask-sudo-pass/-K

See #6467

jf:
Thanks @sivel; you can check out my comment there at #6467 .

jf:
And btw, thanks for being alert and chiming in, and for alerting me to that other issue! I'd otherwise not have found the correct fix (see my 2nd commit) for this problem; thanks...

mpdehaan:
What @sivel said.

This should not be a thing because not all steps should neccessarily be sudoed.   We explicitly _removed_ this behavior before.

We're not going to change this, but if you'd like to discuss further, perhaps bring up a topic on the development list.

jf:
@mpdehaan yes, I make that same case at https://github.com/ansible/ansible/issues/6467#issuecomment-42105692: not all steps should be sudo-ed for **ansible-playbook**. To be clear, I am not asking to change this for ansible-playbook which runs multiple steps. I am asking to change this for **ansible** ad-hoc commands - _which only have one step_. There is a distinction here.

sivel:
So that my response is here too.  The referenced ticket (#6467) was to solve an issue where having 'ask_sudo_pass = True' in ~/.ansible.cfg caused the ansible command to always run ad-hoc commands as sudo.

Many people will have 'ask_sudo_pass = True' in ~/.ansible.cfg to prevent needing to type 'ansible-playbook -K' every time.

So adding this functionality back, would prevent people with 'ask_sudo_pass = True' in ~/.ansible.cfg from being able to easily run ad-hoc commands without sudo.

The other options is decoupling 'ask_sudo_pass = True' from '--ask-sudo-pass', so that only when you supply '--ask-sudo-pass' via the command line, does it imply '--sudo'.  My feeling then, is that it is inconsistent use of configuration.

jf:
Thanks, @sivel. To leave a link for readers (and we can then carry on the discussion on the mailing list): https://groups.google.com/forum/#!topic/ansible-project/YuXMmn0PG_M

PR #7268, opened 2014-05-03
[PEP 3105] use function print instead of statement print

berendt:
Changed all print statements to the print function in bin/ansible-galaxy.

http://legacy.python.org/dev/peps/pep-3105/


mpdehaan:
This patch adds no value as we are not supporting Python 3 yet (much more important to support all the python that is actively in use),  and also only affects a single file.

While your concern is appreciated, do not send us any more of these.  If you would like to discuss this topic, please bring up a thread on the development list.

PR #7271, opened 2014-05-04
Module for creating/updating/deleting a domain record in Digital Ocean

manicminer:
Decided to use python-digitalocean (https://github.com/koalalorenzo/python-digitalocean) as dopy (https://github.com/devo-ps/dopy) seems unmaintained

### DIFF
--- a/library/cloud/digital_ocean_domain_record
+++ b/library/cloud/digital_ocean_domain_record
@@ -1,0 +1,204 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+DOCUMENTATION = '''
+---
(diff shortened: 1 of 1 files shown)


jimi-c:
We already have a module for this, though it does appear that yours adds more options. Why not extend the functionality of the existing module?

Also, while it appears it's been a few months since the last commit, but I wouldn't call it abandoned quite yet. Our other digital ocean modules require dopy, so at this point I don't think we'd want to switch the underlying libraries without migrating all of them.

manicminer:
Hi jimi-c, thanks very much for your comments. There is an existing module 'digital_ocean_domain' which can essentially create zones in Digital Ocean's DNS platform, however this module extends on that to create individual DNS records within a zone.

Usage might look something like this:

``` YAML
- digital_ocean_domain: state=present name=example.com ip=1.2.3.4
                        client_id={{ digital_ocean_client_id }}
                        api_key={{ digital_ocean_api_key }}
- digital_ocean_domain_record: state=present domain=example.com
                               name=blah
                               record_type=A
                               data=1.2.3.5
                               client_id={{ digital_ocean_client_id }}
                               api_key={{ digital_ocean_api_key }}
```

I had thought about combining the functionality into the existing module, but thought that the invocation would get overcomplicated with many records?

manicminer:
Oh, and I had thought about changing the other DO modules to use python-digitalocean as well - dopy has some long outstanding pull requests and is currently incompatible with the DO API for the DomainRecord entity, making it unfeasible to use it for my new module.

jimi-c:
Ok, I guess I misunderstood what the current module does? It seemed like it created an A record for a given system but yours allowed other record types whereas yours allowed other record types.

Regarding dopy, I think it would be best to standardize on a single python library, having to install multiple libraries for similar functions is not a good user experience, but maybe that means re-writing the existing DO modules? Definitely not a trivial task.

mpdehaan:
Hi!

Thanks you very much!

davisonio:
@tombamford Are you going to submit this into the core or extras repository? I think that this module would be a great addition.

PR #7273, opened 2014-05-04
Used module_args passed to function instead of from runner

denisphillips:
The module_args passed to the run(...) function is the proper one to use as with other plugins.


jimi-c:
Merged, thanks!

PR #7275, opened 2014-05-05
updated to quantum_network for non-admin users.

ephur:
Some clouds allow regular tenants to create their own networks. In order for this module to accept that it must not try and list all tenants from keystone, and must allow users to exclude some variables from being included as part of the create network payload. The module should also support insecure connections, since some dev/test environments may not use properly resolving SSL certificates.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,4 @@ coverage.xml
 venv
 Vagrantfile
 .vagrant
+ansible.egg-info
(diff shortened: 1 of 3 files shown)


jimi-c:
@ephur could you please rebase and squash these commits down into a single commit? If you have any questions on how to accomplish that, let us know. Once you do that, you can just "git push --force" to this same feature branch to update the PR, or just close this and send us another. Thanks!

ephur:
Will fix that and submit a new PR, also going to update the other quantum modules even though I'm not using them at the moment, to keep them consistent.

jimi-c:
Great, thanks!

PR #7278, opened 2014-05-05
Evaluate item[*]/item.* in list expressions for with_nested lookup

Ernest0x:
Support for the evaluation of list expressions that contain `item[*]`
or `item.*`. This allows to access the value that was in effect at 
any of the previous (upper) levels at the current loop path/branch
in the nested loop.

For example, consider the following 3-level nested loop:

```
with_nested:
   - ['a', 'b', 'c']
   - ['d', 'e', 'f']
   - item.0
```

item.0 will be evaluated to the value that is in effect at the first
level of the loop for each possible loop path. So, the following
loop paths (combinations) will be produced:

```
['a', 'd', 'a'], ['a', 'e', 'a'], ['a', 'f', 'a'], ['b', 'd', 'b'], ['b', 'e', 'b'], ['b', 'f', 'b'], ['c', 'd', 'c'], ['c', 'e', 'c'], ['c', 'f', 'c']
```

For a more practical example, consider the following setup:

At the hosts file:

```
[somegroup]
host1.example.org
host2.example.org

[anothergroup]
host3.example.org
```

at the host_vars of host1.example.org:

```
network_interfaces:
  - name: eth0
    address: 1.2.3.4
  - name: eth1
    address: 1.2.4.5
```

at the hosts vars of host2.example.org:

```
network_interfaces:
  - name: eth0
    address: 1.2.3.5
  - name: eth1
    address: 1.2.4.6
```

Now, let's say you want from host "host3.example.org" to ping all
the network interfaces for each host in "somegroup". This patch
allows you to do that like this:

```
hosts: host3.example.org
command: /bin/ping {{ item.1.address }}
with_nested:
  - groups["somegroup"]
  - hostvars[item.0].network_interfaces
```

### DIFF
--- a/lib/ansible/runner/lookup_plugins/nested.py
+++ b/lib/ansible/runner/lookup_plugins/nested.py
@@ -16,8 +16,12 @@
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 
 import ansible.utils as utils
-from ansible.utils import safe_eval
 import ansible.errors as errors
+import re
+
+# regular expression to match jinja expressions that contain
+# "item[*]" or "item.*" subexpressions
+item_regex = re.compile(r"[^a-zA-Z0-9_]*item(\[\d+\]|\.\d+)")
 
 def flatten(terms):
     ret = []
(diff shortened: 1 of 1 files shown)


mpdehaan:
"Support the evaluation of "item[]" or "item." subexpressions
in list expressions for loops constructed with with_nested lookup.
This allows to use the values that were in effect at previous
(less deep) loop levels of a nested loop. E.g. an expression can
use item[0], which is evaluated to the value that was in effect at
level 0 (outside/nesting loop) for each possible loop path/branch."

Any chance of getting this dumbed down a bit :)   This language is a bit inpenetrable for me.

Please also fix the formatting in the Markdown above.  Indent 4 spaces and things should render ok.

Thanks!

Ernest0x:
I have added another example that makes it more clear. Also fixed format for markdown.

mpdehaan:
An example is welcomed, but I don't want to see it sprawling in the commit message, and I still find this hard to understand.

As such I think this does not fit our complexity threshold criteria.

michaelkrupp:
Actually IMHO this is much easier to understand than the current implementation of `with_subelements` or `with_nested` as it resembles simple nested loops more closely. This is what I (just judging by the name) would have expected `with_subelements` to behave like.

MannerMan:
Seconded, this would be very useful to have!

ghost:
i've used this as an alternative to vanilla `with_nested` but this doesn't work anymore in ansible2+ after the changes to the plugin API, i've tried with different plugins like action or filter but i can seem to make this work without too much complexity, more that it was required before. Any tips/ideas about how to implement this in ansible2.x+ with the new plugin API @bcoca ?

PR #7280, opened 2014-05-05
Docker inventory bugfix for potential traceback

angstwad:
Docker client occasionally returns a NoneType, so we have to catch the TypeError.  Fixes possible Traceback opportunity.


jimi-c:
Merged, thanks!

PR #7282, opened 2014-05-05
ignore remnants from doing a setup.py install

ephur:
Makes it so all the git toolbars/status are clear and show no changes when doing test installs from the repo during development.


jimi-c:
Merged, thanks!

PR #7051, opened 2014-04-18
Adds support for VPC subnets to ELB

jimbydamonk:
Adds support through boto for VPC subnets.
- name: Create ELB  
  local_action:
    module: ec2_elb_lb
    state: present
    name: 'New ELB'
    security_group_ids: 'sg-123456, sg-67890'
    region: us-west-2
    subnets: 'subnet-123456, subnet-67890'
    listeners:
      - protocol: http
        load_balancer_port: 80
        instance_port: 80
  Or of you have created the through ec2-vpc and register a vpc variable
  - name: Create ELB  
    local_action:
    module: ec2_elb_lb
    state: present
    name: 'New ELB'
    security_group_ids: 'sg-123456, sg-67890'
    region: us-west-2
    subnets: "{{ vpc.subnets | join(',',  attribute='id') }}"  
    listeners:
    - protocol: http
      load_balancer_port: 80
      instance_port: 80

PR #7288, opened 2014-05-06
Improve VPC support

psa:
This change improves VPC support (mostly around subnets).
The group editing change is required for those of us who have networks complex enough to require circular dependencies and can't waste security groups working around it otherwise which is particularly an issue with VPCs and the extra limitations they incur around groups.

I did a little white space cleanup (mostly removing trailing white spaces) while I was here.


mpdehaan:
Can you please provide more information about what is changed and can you elaborate about circular dependencies?

psa:
- ec2_elb_lb: Add VPC subnet support (required for launching an ELB in a VPC). I made it work like zones currently do as it's the VPC equivalent.
- ec2_lc:
  - Add the ability to force the assignment of public IP addresses when instances are launched.
  - Add the ability to specify a spot price (and thus launch spot instances) when declaring a launch config
- ec2_vpc:
  - make resource_tags required as not having them causes a new VPC to be launched every time the playbook is run
- ec2_group:
  - add `create_group_only` flag which restricts the operations to only creating the group (rules are not touched). This is useful where you have two groups that need to include each other in their rules so it's possible to create the groups and then reference the group IDs of the already created groups when updating rules. Without this, if the group is declared twice (so that the group ID can be referenced), all the rules from the group are removed during the invocation for creating the group.

e.g.

```

- name: "Initialize group_a"
  ec2_group:
    name: "group_a"
    description: "group_a"
    vpc_id: "{{ test_vpc_id }}"
    region: "us-east-1"
    state: "present"
    create_group_only: true
  register: "group_a"

- name: "Initialize group_b"
  ec2_group:
    name: "group_b"
    description: "group_b"
    vpc_id: "{{ test_vpc_id }}"
    region: "us-east-1"
    state: "present"
    create_group_only: true
  register: "group_b"

- name: "Create rules for group_a"
  ec2_group:
    name: "group_a"
    description: "group_a"
    vpc_id: "{{ test_vpc_id }}"
    region: "us-east-1"
    state: "present"
    rules:
      - proto: tcp
        group_id: "{{ group_b.group_id }}"
        from_port: 80
        to_port: 80
    rules_egress:
      - proto: tcp
        group_id: "{{ group_b.group_id }}"
        from_port: 443
        to_port: 443

- name: "Create rules for group_b"
  ec2_group:
    name: "group_b"
    description: "group_b"
    vpc_id: "{{ test_vpc_id }}"
    region: "us-east-1"
    state: "present"
    rules:
      - proto: tcp
        group_id: "{{ group_a.group_id }}"
        from_port: 443
        to_port: 443
    rules_egress:
      - proto: tcp
        group_id: "{{ group_a.group_id }}"
        from_port: 80
        to_port: 80
```

juliedavila:
Subnet support was recently merged:
https://github.com/ansible/ansible/pull/7051

psa:
I'm abandoning this pull request and breaking it into a number of smaller ones.

PR #7292, opened 2014-05-06
Interpret comma seperated values in AWS tags as seperate values

nickpack:
In my current environment, I rely heavily on AWS tags to describe the purpose of boxes, and roles are applied specifically to them via tag values. In instances where a machine is filling more than one role, I needed the flexibility to target specific instances by tag grouping. This commit aims to prevent a significant amount of tags being required by expanding a CSV value for a tag into a list of values.

The hostvars ec2_tag_foo value becomes a list of tag values from splitting the field by comma, and an additional group tag_foo_value is added to the inventory for each of the CSV values.

An example of one of the machine tags is: 
Name: role, Value: webserver,dns,ssh,logger

Without this commit, the value is interpreted as: ec2_tag_role = 'webserver_dns_ssh_logger' and in the inventory there is a key [tag_role_webserver_dns_ssh_logger]

With this commit, the value is interpreted as: ec2_tag_role = [
'webserver', 'dns', 'ssh', 'logger'
]

and in the inventory there are groups: [tag_role_webserver], [tag_role_dns], [tag_role_ssh], [tag_role_logger]

Therefore buying the flexibility of being able to target instances by specific role, and avoiding having to create a tonne of additional tags/hostgroup handlers

### DIFF
--- a/plugins/inventory/ec2.ini
+++ b/plugins/inventory/ec2.ini
@@ -53,3 +53,7 @@ cache_path = ~/.ansible/tmp
 # seconds, a new API call will be made, and the cache file will be updated.
 # To disable the cache, set this value to 0
 cache_max_age = 300
+
+# If set to true, if a comma seperated value is found in a tag, the value is
+# exploded into a list of values to provide additional groups to the inventory
+treat_csv_in_tags_as_multiple = False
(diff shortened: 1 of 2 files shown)


mpdehaan:
I'm wondering if this might break some other users who did think the comma was allowed in a value. 

As such, this might need to be a configurable option.

nickpack:
@mpdehaan No worries, I'll add it to ec2.ini and check the settings before splitting the strings - On a more large scale note, is it worth implementing a high level option into the main options parser for collecting inventory options? Obviously in this case it doesnt matter as I have ec2.ini to work with, but just thinking in terms of other inventory plugins - might make things a bit more flexible.

nickpack:
@mpdehaan Configurable option added, default False to not break BC.

nickpack:
Please see #8676

PR #7293, opened 2014-05-06
Update user mailing list link in faq.rst

dhobbs:
Change user mailing list link to ansible-project from ansible-devel


mpdehaan:
Merged, thank you!  This will show up as soon as we run another docs build.

PR #7294, opened 2014-05-06
Fix sync_binlog name in mysql_variables example

adamchainz:
The previous sync_binary_log is not a real MySQL variable.


jimi-c:
Merged, thanks!

PR #7306, opened 2014-05-06
Feature/egg info

abadger:
pkg_resources does its work by reading in metadata about a python package saved in egg-info files.  Regular installs of ansible (python setup.py install) will create the egg-info but for hacking on ansible from a checkout we need to generate egg-info so that the pkg_resources calls in ansible-vault and ansible-playbook can get the information.  hacking/env-setup is the way that we tell people to setup their dev environment so script creation of the egg-info there.

### DIFF
--- a/bin/ansible-playbook
+++ b/bin/ansible-playbook
@@ -18,8 +18,8 @@
 
 #######################################################
 
-#__requires__ = ['ansible']
-#import pkg_resources
+__requires__ = ['ansible']
+import pkg_resources
 
 import sys
 import os
--- a/bin/ansible-vault
+++ b/bin/ansible-vault
@@ -20,8 +20,8 @@
 # example playbook to bootstrap this script in the examples/ dir which
 # installs ansible and sets it up to run on cron.
 
-#__requires__ = ['ansible']
-#import pkg_resources
+__requires__ = ['ansible']
+import pkg_resources
 
 import os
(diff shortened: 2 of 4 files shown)


abadger:
@jimi-c Here's the pull request that lets ansible-vault and ansible-playbook work from a checkout I promised you yesterday.

jimi-c:
@abadger thanks, we'll take a look at this as soon as possible.

mpdehaan:
Hi @abadger sorry for the delay on this one (large queue as you're aware!), one of the ideas behind hacking/env-setup was that it should not modify your system, i.e. you should be able to do this (or python setup.py develop now, on devel) and then stop using the new version to go back to using your old ansible.

Would this cause trouble by over-writing a file from a previous install?

Thanks!

abadger:
@mpdehaan If I understand your question, this should be okay.  It modifies the egg-info files in the checkout directory, not any files on the system. So, for instance, you should be able to run "source ./hacking/env-setup" in the checkout in one terminal and use the the ansible from the checkout.  In a separate terminal, using ansible will use the system installed ansible.

The files on the system (module, scripts, and egg-info) are modified if you do a python setup.py install as per normal.

abadger:
Just a note -- also have to patch /sur/bin/ansible similarly so that ansible --vault-password works.

PR #7312, opened 2014-05-07
reworded copy option docs

bcoca:
recent thread shows that the docs were not clear enough to let people know this toggles between using a 'local' archive and a existing remote one

### DIFF
--- a/library/files/unarchive
+++ b/library/files/unarchive
@@ -39,7 +39,8 @@ options:
     default: null
   copy:
     description:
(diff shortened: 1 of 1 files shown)


bcoca:
including this now might prevent more duplicate updates to these docs, as people seem to be pretty confused by it. Or just include one of the other competing PRs for the same thing.

mpdehaan:
Merged, thank you!

PR #7326, opened 2014-05-07
ec2_tag: Make it possible to search for EC2 resources by specified tags

herbygillot:
On many an occasion, I've wanted the ability to be able to specify from combination of tags and values, and be able to get a list of resource IDs back that match that criteria.  This adds that feature to the **ec2_tag** module.  There is additionally a **resource_type** parameter that can be used to constrain the search down to one particular type of resource for the given tags.

### DIFF
--- a/library/cloud/ec2_tag_search
+++ b/library/cloud/ec2_tag_search
@@ -1,0 +1,110 @@
+#!/usr/bin/python
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: ec2_tag_search
+short_description: search for EC2 resources by tag
+version_added: 1.6.2
+description:
+    - Search EC2 for resources that match the specified tags and returns their resource-specific IDs.  If no explicit resource type is specified, results are returned in a dictionary where results are mapped to their particular resource type.  Otherwise, just the list of resource IDs is returned.  If results are found, found_items is to True.  This module has a dependency on python-boto.
+options:
+  resource_type:
+    description:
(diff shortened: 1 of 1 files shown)


jimbydamonk:
This looks very useful... I am going to give it a try today!

scottanderson42:
Looks similar to a module I wrote: https://github.com/ansible/ansible/pull/6349

If Michael accepts this I will be surprised, given his opposition to inventory-ish modules. :-)

herbygillot:
@scottanderson42 , there are some excellent points for discussion there.

I’ve definitely felt some of the annoyances you mention.

Stepping back, here are the overall themes I see here:
- The concept of what should be inventory and what shouldn’t be
- Exposing information as facts, vs targeted queries through modules (and the ease of working with both methods)

I don’t think I’d be wrong to say that only things you can “log in to” should be considered inventory.  That's why I think in your case in #6349, @mpdehaan wasn't wrong in wanting to keep that kind of functionality within Ansible’s current stratum of inventory.  The way that I actually handled that issue was by having the instance tags be host groups in the form of ‘key=value’, and then performing filtering using group patterns.  If individual attributes about the hosts in the subset were needed, they’d be pulled out of hostvars.  But that was just what worked for me.  I still feel that some of the points you mentioned in #6349 (performance and others) still stand.

That being said, I don’t think this module overlaps with any of the inventory concepts (except for if you were to search for instances, but that’s just incidental).

There are many other resource types that you need to work with other than instances, and the semi-annoying thing is that not every Ansible module for each resource type offers a way to search them by some set of criteria, or for many of the modules, even a way to search at all, for that matter.

Though this change adds a fraction of that functionality in a very rough and unsophisticated way, the right thing in retrospect would probably be to properly add search and query behavior to each module for each resource type.  The searching should be implemented in a consistent way, so that you can use the different criteria types across all of the resource types. 

The big reason for this (and I think can that you agree here) is that we (rightly) do not consider these resources to qualify as inventory, and I think that it's the correct consideration.

It’s often a lot easier and more performant to directly query for what you want intra-play (which you mentioned in #6349 and I agree with).  I do feel that it’s a lot easier and versatile many times to just be like “hey module, just give me foos and bars by this set of criteria” and work with the results intra play.  If things are changing intra-play, it’s way faster and saner to just use this “direct query” approach through modules, rather than having to re-gather facts and figure out what you need from those, which can definitely be a lot more awkward.  Though this plays back into the eternal battle within Ansible of trying to be more declarative vs. more of a programming language, I still feel that this is the correct approach.

So even though this change doesn’t do all this in the most elegant way, I think that it’s a step in the right direction, and its scope is for the most part focused on resources that don't fit in with Ansible’s current inventory concepts.

herbygillot:
Also thinking that this would be better implemented as its own module, perhaps named `ec2_tag_search` or something, as there is definitely functionality that would make this a lot more useful, such as being able to sort by creation time or the like.

herbygillot:
Alright, I've pulled it out into its own module.  It's doing nothing special aside from simply returning the list of resource IDs that match the specified tags.

mpdehaan:
I agree in that this module doesn't declare or enforce state, and in fact should be something that comes out of inventory, or is otherwise fine for others to use but not something I'd like to include in core.

We can discuss further on the mailing list.

PR #7334, opened 2014-05-08
Fail if monit reload failed; don't mask the error.

lamby:
This is necessary for the scenario when you push a new, broken monit
config out, and then set a state=reloaded handler - the error was
previously swallowed so you could end up with successful play but missing
monitoring (!).

Signed-off-by: Chris Lamb chris@chris-lamb.co.uk


jimi-c:
Merged, thanks!

PR #7335, opened 2014-05-08
Route53 module handling of domains with "*" is broken

napkindrawing:
##### Issue Type

Bugfix Pull Request
##### Ansible Version:

ansible 1.6.1

But affects devel branch as well
##### Environment:

OSX 10.9.2
##### Summary:

The escaping to get around boto/ec2 wildcard domains mistakenly discards the escaping performed for `*`.

The initial attempt to fix this problem ( #6324 ) mistakenly discarded the handling of `*`

`decoded_name` was created twice, each from `rset.name.replace(...)`

So, the second call to `.replace(r'\100', '@')` overwrites `decoded_name`, discarding the result of the call to `.replace(r'\052', '*')`
##### Steps To Reproduce:

Specify a `record` for the route53 module containing a `*` using `overwrite=yes`
##### Expected Results:

The domain should be handled correctly and updated, just as domains without `*` are.
##### Actual Results:

Use of the route53 module to update a domain containing `*` results in the following error:

```
failed: [ZZZ.com] => (item=*.XXXX.com) => {"failed": true, "item": "*.XXXX.com"}                                       │YYY.com : ok=3    changed=0    unreachable=0    failed=0
msg: Tried to create resource record set [name='\052.XXXX.com.', type='CNAME'] but it already exists
```

### DIFF
--- a/library/cloud/route53
+++ b/library/cloud/route53
@@ -223,7 +223,7 @@ def main():
         # Due to a bug in either AWS or Boto, "special" characters are returned as octals, preventing round
         # tripping of things like * and @.
         decoded_name = rset.name.replace(r'\052', '*')
-        decoded_name = rset.name.replace(r'\100', '@')
+        decoded_name = decoded_name.replace(r'\100', '@')
 
         if rset.type == type_in and decoded_name == record_in:
             found_record = True


jimi-c:
Merged, thanks!

napkindrawing:
That was fast — thanks! 

I'm lovin' ansible btw =)

jimi-c:
I try and immediately merge in fixes for obvious mistakes like that. And thanks!

PR #7338, opened 2014-05-08
Remove check for Ghost state

discordianfish:
This doesn't exist anymore and causes ansible to throw a KeyError


jimi-c:
Most likely that should be changed to `i['State'].get('Ghost', False)` instead of dropped completely, as some users might still be running a version of docker that includes it.

discordianfish:
@jimi-c Fair enough, fixed!

jimi-c:
Merged, thanks!

PR #7341, opened 2014-05-09
rabbitmq module: Don't hardcode sbin in the prefix

phenomenes:
This can be passed in prefix and some distributions install rabbitmq-plugins in a different location.

### DIFF
--- a/library/messaging/rabbitmq_plugin
+++ b/library/messaging/rabbitmq_plugin
@@ -64,7 +64,7 @@ class RabbitMqPlugins(object):
         self.module = module
 
         if module.params['prefix']:
-            self._rabbitmq_plugins = module.params['prefix'] + "/sbin/rabbitmq-plugins"
+            self._rabbitmq_plugins = module.params['prefix'] + '/rabbitmq-plugins'
         else:
             self._rabbitmq_plugins = module.get_bin_path('rabbitmq-plugins', True)
 


mpdehaan:
While I agree, this may break some existing users passing in values, we may wish to consider some soft intelligence here.

jimi-c:
I agree, why not use `os.path.exists()` to try a few of those possible variations?

<pre>
for sub_path in ("", "bin", "sbin", "and", "some", "others"):
    found = False
    tmp_sub_path = os.path.join(module.params['prefix'], sub_path, "rabbitmq-plugins")
    if os.path.exists(tmp_sub_path):
        self._rabbitmq_plugins = tmp_sub_path
        found = True
    if not found:
        module.fail_json(msg="no rabbitmq-plugins found at the specified prefix: %s" % module.params['prefix'])
</pre>

phenomenes:
I think that the idea of having 'prefix' is to be able to run rabbitmq-pluggins from any location. With the above code the user is tied to always install rabbitmq-plugins in either 'bin', 'sbin', etc.

This is might not be very elegant, but it give us backward compatibility:

```
            sbin_path = module.params['prefix'] + '/sbin/rabbitmq-plugins'
            if os.path.exists(sbin_path):
                self._rabbitmq_plugins = sbin_path
            else:
                self._rabbitmq_plugins = module.params['prefix'] + '/rabbitmq-plugins'
```

phenomenes:
Any comments? I'd like to get this solved so if there are any preferences I'd be happy to change the PR

jimi-c:
@phenomenes sorry for the delay, we'll get to this as soon as possible, there are just a few higher priority items in the queue. Thanks!

phenomenes:
any news on this?

jimi-c:
@phenomenes sorry for the delay, there's nothing wrong with this PR, we're just working through quite a backlog at the moment.

mpdehaan:
Sorry for the handling of this @phenomenes - we should have answered the easy question rather than just saying there's a bit of a backlog so you have a chance to dive in here.

Your proposal seems quite good to me about trying both.  If you want to amend this patch to use the above, I'll gladly take it.

Because the queue handling of this is long, if you give me a prod on ansible-devel or email me once complete I'll push it in.

Thanks!

phenomenes:
After further analysing this I realised that the rpm creates the following symlink: 

```
 /usr/lib/rabbitmq/bin/rabbitmq-plugins -> ../lib/rabbitmq_server-3.1.5/sbin/rabbitmq-plugins
```

therefore it gets installed in /sbin, so I'm closing this PR, sorry for the noise and thanks.

PR #7342, opened 2014-05-09
rds module: Fix undeclared reference to "params" when setting password in modify mode

herbygillot:
While modifying an RDS instance with the 'rds' module, one of the actions was to change the password.

While doing so, this happened:

```
TASK: [Set password replica] ******************************************
  File "/Users/herby/.ansible/tmp/ansible-tmp-1399628204.52-67026896752847/rds", line 1865, in <module>
    main()
  File "/Users/herby/.ansible/tmp/ansible-tmp-1399628204.52-67026896752847/rds", line 393, in main
    params["master_password"] = password
UnboundLocalError: local variable 'params' referenced before assignment
```

The proposed change fixes this issue.

### DIFF
--- a/library/cloud/rds
+++ b/library/cloud/rds
@@ -370,6 +370,9 @@ def main():
         else:
             return 'vpc_security_groups'
 
+    # Package up the optional parameters
+    params = {}
+
     # Validate parameters for each command
     if command == 'create':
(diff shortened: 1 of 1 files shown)


jimi-c:
Merged, thanks!

PR #7344, opened 2014-05-09
Changing removed to absent

majksner:


mpdehaan:
Yeah so, installed/removed are pretty much normal pairs to use.  They aliased to present/absent.

We should be consistent if changing "removed" we should also changing "installed".

Installed seems to make a bit more sense for packages so I'm ok with leaving this for now.

PR #7350, opened 2014-05-09
Fix: Remove check for Ghost state

discordianfish:
This fixed the just committed change handling non existent Ghost state.

@jimi-c: Sorry about that


jimi-c:
Merged, thanks!

PR #7351, opened 2014-05-09
Make checks work with older Docker version

discordianfish:
ApiVersion isn't available on older Docker versions. Beside that, 0.3.1 doesn't support 0.10 api commands, so I've changed that check as well.

### DIFF
--- a/library/cloud/docker
+++ b/library/cloud/docker
@@ -543,7 +543,7 @@ class DockerManager:
                   'tty':          self.module.params.get('tty'),
                   }
 
-        if docker.utils.compare_version('1.10', self.client.version()['ApiVersion']) < 0:
+        if docker.utils.compare_version('0.10', self.client.version()['Version']) < 0:
             params['dns'] = self.module.params.get('dns')
             params['volumes_from'] = self.module.params.get('volumes_from')
 
@@ -574,7 +574,7 @@ class DockerManager:
             'privileged':   self.module.params.get('privileged'),
             'links': self.links,
         }
-        if docker.utils.compare_version('1.10', self.client.version()['ApiVersion']) >= 0 and hasattr(docker, '__version__') and docker.__version__ > '0.3.0':
+        if docker.utils.compare_version('0.10', self.client.version()['Version']) >= 0 and hasattr(docker, '__version__') and docker.__version__ > '0.3.1':
             params['dns'] = self.module.params.get('dns')
             params['volumes_from'] = self.module.params.get('volumes_from')
 


discordianfish:
@mpdehaan Any reasons thats not yet merged?

davisein:
I have tested and it works for me. I was getting this error this days without solution. :+1:

CharlyP:
It fixed the problem for me too (Docker version 0.9.1, build 3600720 and Ansible 1.7) --> :+1:

disposable-ksa98:
How can I apply this patch to my ansible installation? @davisein 

Someone please merge it! I'm stuck! I have the same setup as @CharlyP 

Edit: NVM, got it: `pip install -Iv https://github.com/discordianfish/ansible/archive/fix-version-checks.zip`. I haven't had time to test it yet, I'll be back after that.

Edit2: Didn't work. I'm still getting `KeyError: 'ApiVersion'`

davisein:
Hi there

I have just tried the code in the pull request.

Just googling I found this:

https://gist.github.com/piscisaureus/3342247

But I have just copied the PR repo and checkout the right branch.

The follow the docs of Ansible in the docs to use the last version, but in
that directory instead of in a fresh one.

Tell me if that helps!

discordianfish:
After some discussions I came to the conclusion that using ApiVersion is better. This makes it impossible to use ansible with docker <= 0.10 but it makes it easier to maintain this in the future.

discordianfish:
I don't need this anymore because I've upgraded all our systems to >= 0.10 (and you should do!).
If anyone needs to support older versions I would recommend to check if ApiVersion exists and assume < 0.10 if not. This would fix the issue as well but still use the correct 'ApiVersion'.

disposable-ksa98:
@davisein Thanks! It was about time I learned how to do that.

@discordianfish Yep, the problem I was having was that Ubuntu 14.04 comes with 0.9.1 (why?!) and there wasn't a nice way to upgrade it. What I ended up doing was installing it like that (official instructions), and then simply overwriting the binary with the newest version downloaded from the website.

balboah:
I'm also having the issue with Ubuntu 14.04 LTS, which I would believe is a good thing to support even if it's 0.9.1 by default since this will be a very common environment.

The version comparisons is a bit confusing to me.
In one place we have this:

``` python
if docker.utils.compare_version('1.10', self.client.version()['ApiVersion']) < 0:
            params['dns'] = self.module.params.get('dns')
            params['volumes_from'] = self.module.params.get('volumes_from')
```

In the next we have this:

``` python
if docker.utils.compare_version('1.10', self.client.version()['ApiVersion']) >= 0 and hasattr(docker, '__version__') and docker.__version__ > '0.3.0':
            params['dns'] = self.module.params.get('dns')
            params['volumes_from'] = self.module.params.get('volumes_from')
```

So the first case sets these extra params if the version is OLDER than 1.10, while the other sets it if it's equal to or newer than 1.10 with the additional check of the docker module version.

Shouldn't both be the same logic where they are setting the same extra parameters **dns** and **volumes_from**?
Also in the docker-py module code itself, there are a couple of checks against the API version that was set upon initialising the client, which is 1.9 by default. If we want 1.10, shouldn't the client be initialised with this version? Are these bugs or just me not understanding the code?

balboah:
Also mentioned in #7619

disposable-ksa98:
@balboah I had issues with Ubuntu 14.04 too. Just install Docker normally (0.9.1) and then update it to the latest version manually, by overwriting the binary with the one you can download from the website. Also make sure to switch to AUFS.

typehorror:
Having the issue with 14.04 too. Unclear how to get it fixed.

```
TASK: [Make sure image is running] ******************************************** 
failed: [localhost] => (item=ubuntu/logstash1) => {"failed": true, "item": "ubuntu/logstash1", "parsed": false}
invalid output was: SUDO-SUCCESS-ygrosbmxvftnpsrwvtgpncvaqzweazux
  File "/home/brice/.ansible/tmp/ansible-tmp-1403046155.99-7077217954574/docker", line 1971, in <module>
    main()
  File "/home/brice/.ansible/tmp/ansible-tmp-1403046155.99-7077217954574/docker", line 693, in main
    containers = manager.create_containers(1)
  File "/home/brice/.ansible/tmp/ansible-tmp-1403046155.99-7077217954574/docker", line 548, in create_containers
    if docker.utils.compare_version('1.10', self.client.version()['ApiVersion']) < 0:
KeyError: 'ApiVersion'


```

versions:

```
brice@vm1:~/playbooks$ ansible --version
ansible 1.6.3
brice@vm1:~/playbooks$ docker --help
docker - version 1.4
...
```

PR #7355, opened 2014-05-10
Add dots to valid path characters

itiut:
Recognize paths containing dots as valid paths.
Homebrew module can be used for linuxbrew and linuxbrew is usually installed in `~/.linuxbrew`.


jimi-c:
Merged, thanks!

Issue #7258, opened 2014-05-02
unarchive module doesn't check for previous upload (edited description)


alexanderjardim-zz:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.6 (devel b753625dbf) last updated 2014/05/02 17:13:20 (GMT -200)
##### Environment:

Darwin Kernel Version 13.1.0: Wed Apr  2 23:52:02 PDT 2014; root:xnu-422.92.1~2/RELEASE_X86_64 x86_64
##### Summary:

When I use unarchive on a task, it takes too long to execute even if the file was already unarchived.
##### Steps To Reproduce:

Setup {{java_dir}}, {{download_dir}} properly
Using this tasks in a playbook.

```

---
- name: download Oracle JDK7 to host
  local_action: >
    command wget -N -P "{{download_dir}}" --no-check-certificate --no-cookies --header "Cookie: oraclelicense=accept-securebackup-cookie" http://download.oracle.com/otn-pub/java/jdk/7u51-b13/jdk-7u51-linux-x64.tar.gz
    creates="{{download_dir}}/jdk-7u51-linux-x64.tar.gz"
    sha256sum=77367c3ef36e0930bf3089fb41824f4b8cf55dcc8f43cce0868f7687a474f55c

- name: "{{java_dir}} must exist"
  file: path={{java_dir}} state=directory

- name: unpacks Jdk7
  unarchive: 
    src="{{download_dir}}/jdk-7u51-linux-x64.tar.gz"
    dest="{{java_dir}}"
    creates="{{java_dir}}/{{java_dist}}/bin/java"
    force=no

- name: setup JAVA_HOME
  lineinfile:
    dest=~/.bashrc
    line="export JAVA_HOME={{java_dir}}/jdk1.7.0_51"
    state=present
    insertafter=EOF
    create=true

- name: append JAVA_HOME to PATH
  lineinfile:
    dest=~/.bashrc
    line="export PATH=$JAVA_HOME/bin:$PATH"
    state=present
    insertafter=EOF
    create=true
```

Run the following command twice.
$ time ansible-playbook -i hostfile playbook.yml
##### Expected Results:

Task "unpack jdk7" should take less time on the second execution as it doesn't need to change the files again at the hosts.
##### Actual Results:

Even as the files don't need to change, I see ansible PUTting the "tgz"package again at the hosts.

```
TASK: [oracle-jdk | append JAVA_HOME to PATH] ********************************* 
[...]

PR #7357, opened 2014-05-10
Skip unarchive if creates is specified and file exists.

jonhadfield:
Unarchive fails to skip the step if the file specified in creates exists. This change makes ansible behave in the way the documentation suggests.


jimi-c:
I don't believe this is the correct resolution for #7258, though it may also be a useful feature. The issue with #7258 is the file copy always occurs, which may be the case even with this patch. As I said, this may still be a useful feature with unarchive, as some people may not want to actually extract the file even if the file upload previously existed but had already been extracted.

PR #7359, opened 2014-05-11
Add an option to ec2.ini to filter instances based on various things

tgerla:
This lets you filter instances right at the EC2 API level using the EC2 inventory plugin. You can use this to retrieve only instances with certain tags, or only certain instance types, etc, etc. See ec2.ini for more info and examples.


mpdehaan:
Nice!

Quick question -- Does this support multiple tags, suppose I want to import tags named "datacenter1" and "datacenter2" ?

Basically I'm wondering if we should do the "OR" thing before we do the "AND" thing.  Obviously doing both (and arbitrary levels of mixing and or) could get complicated.

tgerla:
So, this filtering option is "and". The EC2 API doesn't do "or", so if we wanted "or", we'd have to separate the filters, loop over them, and call get_all_instances multiple times with each filter. A little messier, but doable, and perhaps more intuitive...

tylerhenthorn:
Awesome.

mpdehaan:
This now has a merge commit in here, which we normally don't accept.

Be sure to do "git pull --rebase" vs "git pull" or "git rebase" vs "git merge".

Still needs to change to "OR".

tgerla:
I messed up this PR; will re-do on a feature branch. Closing.

Issue #7367, opened 2014-05-12
docker module: image build fails on first pass with ValueError decoding json


hjwp:
Full traceback:

```
failed: [54.82.88.127] => {"failed": true, "parsed": false}
invalid output was: Traceback (most recent call last):
  File "/root/.ansible/tmp/ansible-tmp-1399891318.63-48689284481872/docker_image", line 1327, in <module>
    main()
  File "/root/.ansible/tmp/ansible-tmp-1399891318.63-48689284481872/docker_image", line 217, in main
    image_id = manager.build()
  File "/root/.ansible/tmp/ansible-tmp-1399891318.63-48689284481872/docker_image", line 140, in build
    chunk_json = json.loads(chunk)
  File "/usr/lib/python2.7/json/__init__.py", line 338, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python2.7/json/decoder.py", line 368, in decode
    raise ValueError(errmsg("Extra data", s, end, len(s)))
ValueError: Extra data: line 1 column 87 - line 1 column 274067 (char 86 - 274066)
```

The operation completes if run a second time...

hjwp:
a print-based debug in decoder.py suggests the json may be malformed -- it looks like there are two
objects rather than one:

```
{"status":"Pulling image (trusty) from ubuntu","progressDetail":{},"id":"99ec81b80c55"}{"status":"Pu (rest of line truncated)
```

column 87 seems to be the point at which the first object ends and the second one begins, `}{`.

sivel:
It seems that the pull for the  base image shows up in a single 'chunk' from build with `stream=True`.

Not sure I terribly enjoy this solution, but here is something I tested:

```
diff --git a/library/cloud/docker_image b/library/cloud/docker_image
index 2f5a02b..f960aca 100644
--- a/library/cloud/docker_image
+++ b/library/cloud/docker_image
@@ -140,6 +140,9 @@ class DockerImageManager:
             if not chunk:
                 continue

+            if '"}{"' in chunk:
+                chunk = '{"%s' % chunk.split('"}{"')[-1]
+
             chunk_json = json.loads(chunk)

             if 'error' in chunk_json:
```

or

```
diff --git a/library/cloud/docker_image b/library/cloud/docker_image
index 2f5a02b..f1cba45 100644
--- a/library/cloud/docker_image
+++ b/library/cloud/docker_image
@@ -140,7 +140,10 @@ class DockerImageManager:
             if not chunk:
                 continue

[...]

PR #7373, opened 2014-05-12
Handle ValueError during json.loads of json data from build

sivel:
See #7367

When building an image from a Dockerfile, the first time the base image is downloaded from the index, that portion of the build step returns invalid JSON data that causes `json.loads` to raise a ValueError.

This pull request catches that exception and continues on, ignoring the parsing failure.

If for some reason we skip every iteration of looping through the stream and no `image_id` was set, indicate that an unknown error was encountered.

### DIFF
--- a/library/cloud/docker_image
+++ b/library/cloud/docker_image
@@ -140,7 +140,10 @@ class DockerImageManager:
             if not chunk:
                 continue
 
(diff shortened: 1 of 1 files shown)


jimi-c:
Merged, thanks!

PR #7374, opened 2014-05-13
Allow wait_for to wait on non-traditional files

### LINKED ISSUE #6710: wait_for hangs on UNIX pipes
When trying to `wait_for` a pipe, it might block forever. This is because the process blocks until the other end of the pipe gets opened. A solution to this would be to use `stat` before trying to open the file.

hfinucane:
Use os.path.exists to check for file existence, instead of "can we open
this file for reading".

Fixes #6710

### DIFF
--- a/library/utilities/wait_for
+++ b/library/utilities/wait_for
@@ -176,7 +176,8 @@ def main():
         end = start + datetime.timedelta(seconds=timeout)
         while datetime.datetime.now() < end:
             if path:
-                if os.path.exists(path):
+                try:
+                    os.stat(path)
                     if search_regex:
                         try:
                             f = open(path)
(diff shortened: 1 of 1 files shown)


jimi-c:
Edited, as it was not accurate. However, using os.path.exists() on a path that does exist but is not accessible returns False as opposed to giving some other error. If you use os.stat() instead, a proper error is raised as follows:

<pre>
>>> os.path.exists('/var/log/httpd/error.log')
False
>>> os.stat('/var/log/httpd/error.log')
OSError: [Errno 13] Permission denied: '/var/log/httpd/error.log'
</pre>

That error should be caught and fail_json used to properly alert the user as opposed to returning a false-positive/false-negative.

hfinucane:
The semantics of the code _before_ this patch was, if it cannot access a file because of permissions, it will retry until it can (or time out). I preserved that.

Do you still want me to change it?

jimi-c:
Yes, I think failing fast is more appropriate than waiting for it to timeout. The modifications to your patch should be fairly minimal.

hfinucane:
Indeed. New version should return errors like
`"msg": "Failed to stat /proc/1/exe, Permission denied"`
when failing.

jimi-c:
Looks good, we'll get it in queue for testing as soon as possible. Thanks!

jimi-c:
Merged, thanks!

PR #7376, opened 2014-05-13
When ansible_ssh_user is a number ssh fail to connect due to string concat

marcore:
Convert to string to fix runtime error due to string concat in self.common_args += ["-o", "User="+self.user] when ansible_ssh_user is a number


jimi-c:
Merged, thanks!

PR #7379, opened 2014-05-13
Added --no-recommends option to zypper module

alxgu:
Now it is possible to use the "--no-recommends" parameter for the install command of zypper. 
zypper's default is to install recommended packages, but for the ansible zypper module is no-recommends the default.


mpdehaan:
Needs to add the "version_added" attribute in the module docs, FWIW...

alxgu:
Any further issues with the pull request?

mpdehaan:
Merged in for 1.8 devel, thank you!

PR #7380, opened 2014-05-13
Fix missing variable initialization so an unspecified internal_network works

carsongee:
Without this line the module fails when leaving internal_network unspecified from calling an if on an undeclared variable.


jimi-c:
Merged, thanks!

PR #7394, opened 2014-05-14
Hostinfo dynamicinventory

AZaugg:
Created dynamic inventory for CMDB tool hostinfo

### DIFF
--- a/plugins/inventory/hostinfo.ini
+++ b/plugins/inventory/hostinfo.ini
@@ -1,0 +1,21 @@
+# EXAMPLE
+# [SOLARIS]
+# hosts=hostinfo os=solaris
+# ignorehosts=builder10
+
+
+[SOLARIS]
+hosts=hostinfo os=solaris
+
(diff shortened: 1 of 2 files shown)


mpdehaan:
I'm not entirely sure of the rate of deployment of hostinfo so am unsure if we want to include this in core.  More information about how widely it is deployed would be useful to make the case for this.

Also, a few cleanup things - Can you please add copyright header information as per the other inventory scripts, and also modify functionNamesInCamelCase to be with_underscores, as suggested in our coding guidelines document https://github.com/ansible/ansible/blob/devel/CODING_GUIDELINES.md

Thank you!

AZaugg:
I agree, hostinfo does not have a large user foot print at present

BigAl:
This would be a handy feature. any chance to get it merged?

dwagon:
As one of the hostinfo developers this would be a really cool feature :)

PR #7395, opened 2014-05-14
Add hostname_variable to ec2 inventory script

hkariti:
Use it to override the inventory name set for ec2 hosts with an ec2 variable.

### DIFF
--- a/contrib/inventory/ec2.ini
+++ b/contrib/inventory/ec2.ini
@@ -29,6 +29,11 @@ regions_exclude = us-gov-west-1,cn-north-1
 # in the event of a collision.
 destination_variable = public_dns_name
 
+# This allows you to override the inventory_name with an ec2 variable, instead
+# of using the destination_variable above. Addressing (aka ansible_ssh_host)
+# will still use destination_variable. Tags should be written as 'tag_TAGNAME'.
+#hostname_variable = tag_Name
+
 # For server inside a VPC, using DNS names may not make sense. When an instance
 # has 'subnet_id' set, this variable is used. If the subnet is public, setting
 # this to 'ip_address' will return the public IP address. For instances in a
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -236,7 +236,11 @@ class Ec2Inventory(object):
         # Destination addresses
         self.destination_variable = config.get('ec2', 'destination_variable')
         self.vpc_destination_variable = config.get('ec2', 'vpc_destination_variable')
-
+        if config.has_option('ec2', 'hostname_variable'):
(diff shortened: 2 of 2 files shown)


willthames:
Looks good to me.

I don't think there's any reason why the same approach taken for `hostname_variable` couldn't also be applied to `destination_variable` and `vpc_destination_variable`, so might be worth making a method out of the tag extraction and applying to all three.

hkariti:
Not sure, because these options should be addressable values

willthames:
If the tag_Name happens to be in DNS, then it will be addressable. If I understand your meaning correctly.

mpdehaan:
# Rebase Request

Hi!

which is very active:

~  
~  
~

hkariti:
Rebased. Also added the patch to add_rds_instance, but I don't have rds instances to didn't test

micxer:
Is there any progress in merging this? I would have good use for this feature as well, as I have an upcoming use case, where I need to set variables based on the host, it would be nice to be able to use the name of an instance as hostname in Ansible and not the IP address. I fiddled with `destination_variable` and `vpc_destination_variable` but didn't get there. AFAIK these are only intended to let Ansible find out how to connect the instances.

bennojoy:
@micxer  this capabilty has been added via 9825 , are you testing with latest devel ?

hkariti:
@bennojoy #9825 requires that the tag be also the value that's given to ssh. So, for example you have to have the hostname in /etc/hosts, or have the full dns resolvable name in the aws tag.

micxer:
@bennojoy just try it and it works. However that's not, what I tried to achieve. I want ansible to connect to the machine using the IP oder public DNS or whatever that works, but the ansible_hostname should match a given tag, i.e. the Name tag.

So I can have `server1` and `server2` for file names in `host_vars` and not the IP addresses and run `ansible --limit server1`.

amenonsen:
@bennojoy did you get a chance to look at this again in light of the comments above?

davidvasandani:
+1. Would love to have this.

amenonsen:
ping @bennojoy

magnetik:
Same use case as others. With current implementation we are using the tag name as destination variable, and running a custom script to populate the ssh config with the right "fake" hostname to the real DNS hostname

amenonsen:
@hkariti I read your patch, and it seems simple enough, and several people have said they would find it useful. Do you feel like rebasing once more, by any chance?

hkariti:
Rebased. I noticed that there's also support for elasticcache clusters in the inventory script. I don't know how these work, so I didn't try to add support for them (didn't know if that even made sense).

amenonsen:
Looks good, thanks. Merged.

PR #7401, opened 2014-05-14
ec2: set type and default for ebs_optimized flag

ghost:
Improves upon a change we made about a month ago by setting type and value, resulting in more predictable behavior.


analytically:
+1

mpdehaan:
Merged, thank you!

PR #7406, opened 2014-05-14
Fix known_hosts.py to work on older versions of python

specnazzz:
This pull request https://github.com/ansible/ansible/pull/6639 fixed the validation of fingerprint hashes in `known_hosts.py` but broke the git module on managed nodes where the default version of python is older then 2.5 (e.g. rhel5) and would require to also install `python-hashlib` package instead of only `python-simplejson`.


mpdehaan:
merged, thank you!

PR #7407, opened 2014-05-14
Give reasonable error when symlink fails

telsacolton:
When a symlink fails because the path already exists and force=no, we should output path in the error message instead of source.


jimi-c:
Merged, thanks!

PR #7409, opened 2014-05-15
add ldap module

grossws:
When I tried to configure ldap with ansible, I have to do it in three phases: copy `ldif` file to target system, perform `ldapsearch` to check if entry exists and call to `ldapadd` or `ldapmodify` after that (conditionally). It's uncomfortable and error-prone. So I implemented ldap module using python ldap library, present in base repos (CentOS base in my case). 

Another advantage over calling `ldapmodify` (and other utils from `openldap-clients`) that module supports fully idempotent ldap operation and check mode. 

Supported operation modes are equal to `ldapadd`, `ldapmodify` and `ldapdelete` openldap utils but will not apply any operations if LDAP entry wouldn't change after these operations (like deleting non-existent attribute, replace attribute with values already present in it etc).

Supported features:
- connect using `ldap`, `ldaps` and `ldapi` protocols;
- _STARTTLS_ over `ldap` protocol;
- anonymous and authenticated operation (using `simple` auth);
- SASL _EXTERNAL_ auth method on `ldapi` protocol (usually used to configure OpenLDAP 2.4+ on-line configs in `cn=config` subtree);
- adding and deleting entries (via `state=present` and `state=absent`);
- adding, deleting and replacing individual attributes, attribute-value pairs or their combinations;
- `simple_args` and `complex_args` format for entry modification command;
- many examples to explain different args format and operation modes;
- simple DSL for one-line ldap entry modification command (with EBNF-like `simple_args` format definition);
- _UTF-8_ for attribute values;
- grouping (by attribute) and collapsing ldap modification operations to get true idempotent behavior;
- parsed and final ldap modifications (as they will be sent to ldap server) are logged to simplify module learing and operations debugging.

Module from #7073 was useless for me.

Target system requirement:
- python-ldap

### DIFF
--- a/library/system/ldap
+++ b/library/system/ldap
@@ -46,6 +46,13 @@ options:
         required: false
         default: null
         aliases: []
+    auth:
+        description:
+            - Authentication method. C(sasl_ext) is supported only on U(ldapi://) (unix socket connection)
+        required: false
+        aliases: []
+        choices: ["simple", "sasl_ext"]
+        default: "simple"
     dn:
         description:
             - Operation-specific distinguished name (dn)
@@ -185,6 +192,13 @@ objectClass: dcObject
 delete: dc
 
 
+# Use SASL EXTERNAL auth method on unix socket (ldapi://).
+# It's often used on initial OpenLDAP 2.4+ configuration on cn=config tree.
+
+- ldap: uri=ldapi:// auth=sasl_ext
+        dn=ou=ou1,dc=my-domain,dc=com
+        data=a/objectClass=organizationalUnit;a/ou=ou1
+
 
 # `data` one-line format BNF-like grammar.
 
@@ -206,6 +220,7 @@ safe-utf8-init-char ::= SAFE-INIT-CHAR | UTF8-2 | UTF8-3 | UTF8-4 | UTF8-5 | UTF
 
 try:
     import ldap
+    import ldap.sasl
     import ldap.modlist
     import ldapurl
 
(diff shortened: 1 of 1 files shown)


mpdehaan:
Possibly might wish to be a role -- on our list to dig in.  Thanks!

grossws:
@mpdehaan, my English is not so good to understand what did you mean. Will you or other commiter accept or reject this pull request after reviewing in a while? Or something else?

mpdehaan:
We need to think about it.

jirutka:
@grossws Can I ask you why my module was useless for you? o.O

grossws:
@jirutka, there are several unsupported features required for me:
- SASL bind operation with EXTERNAL method (useful for modifying cn=config subtree under local root);
- partial updates (e. g. I want to add or replace one attr and don't want to write all LDIF for entry like `olcDatabaseConfig` or something similar);
- selective attrs delete (e.g you can't delete some attr-value pairs or all attr-value pairs for one attr, but only full entry itself);
- UTF-8 (`python-ldap` requires special utf-8 handling as you can see in my code).

Also there are implementation bugs:
- `changed=True` in dry_run even if ldap entry not present.

These are differences at first glance. I haven't dig into your implementation to find more of them.

Also, I have found your implementation only when I was preparing pull request after my version was finished. My first attemp to write ldap module was like yours (but it had two modules: one for entry, one for attr). When I tried to find any ldap module I found nothing.

mpdehaan:
Hi!

Thanks you very much!

grossws:
Created new PR: ansible/ansible-modules-extras#30

PR #7410, opened 2014-05-15
Fix parameters check for ec2_vol

willthames:
Split the error check into two error checks and delay the
checks so that listing existing volumes works more nicely.

The error check should check that:
- One and only one of volume_size id or name is set

This fix adds the 'only one' part of that check and provides
more useful error messages.


mpdehaan:
Merged, thank you!

PR #7412, opened 2014-05-15
Fix skip dir inventory extensions

bellkev:
I discovered this after I upgraded from 1.6 to the current devel branch. My /etc/ansible looks like this:

```
├── ansible.cfg
└── hosts
    ├── ec2
    ├── ec2.ini
    └── local
```

(where ec2 is really ec2.py)

And doing e.g. `ansible --list-hosts all` started returning ini errors. I discovered that it was because the extension filtering broke in a recent refactoring of ansible.inventory.dir.InventoryDirectory.

This pull request includes a new test that demonstrates the issue as well as a fix.

### DIFF
--- a/lib/ansible/inventory/dir.py
+++ b/lib/ansible/inventory/dir.py
@@ -36,13 +36,12 @@ class InventoryDirectory(object):
         self.parsers = []
         self.hosts = {}
         self.groups = {}
- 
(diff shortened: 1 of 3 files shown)


jimi-c:
Merged, thanks!

PR #7421, opened 2014-05-15
Add run_once task keyword to cause a task to bypass the host loop

sivel:
This pull request adds a `bypass` task key, to allow a task to funciton like a bypass host loop plugin similar to add_host/pause/etc...

This was inspired by:
https://groups.google.com/forum/#!topic/ansible-project/3-LaugNHe0M

Example:
###### bypass.yml

```

---
- hosts: all
  tasks:
    - debug:
        var: inventory_hostname
      bypass: True
```
###### output

```

ok: [localhost]
ok: [srv1.example.org]
ok: [srv2.exampe.org]
ok: [srv3.example.org]

TASK: [debug ] ****************************************************************
ok: [localhost] => {
    "inventory_hostname": "localhost"
}

srv1.example.org           : ok=2    changed=0    unreachable=0    failed=0
srv2.example.org           : ok=2    changed=0    unreachable=0    failed=0
srv3.example.org           : ok=2    changed=0    unreachable=0    failed=0
```

### DIFF
--- a/docsite/rst/playbooks_delegation.rst
+++ b/docsite/rst/playbooks_delegation.rst
@@ -118,6 +118,40 @@ Here is an example::
 Note that you must have passphrase-less SSH keys or an ssh-agent configured for this to work, otherwise rsync
 will need to ask for a passphrase.
 
+.. _run_once:
+
+Run Once
+````````
+
+In some cases there may be a need to only run a task one time and only on one host. This can be achieved
+by configuring "run_once" on a task::
+
+    ---
+    # ...
+
+      tasks:
+
+        # ...
+
+        - command: /opt/application/upgrade_db.py
+          run_once: true
+
+        # ...
+
+This can be optionally paired with "delegate_to" to specify an individual host to execute on::
+
+        - command: /opt/application/upgrade_db.py
+          run_once: true
+          delegate_to: web01.example.org
+
+When "run_once" is not used with "delegate_to" it will execute on the first host, as defined by inventory,
+in the group(s) of hosts targeted by the play. e.g. webservers[0] if the play targeted "hosts: webservers".
+
+This aproach is similar, although more concise and cleaner than applying a conditional to a task such as::
+
+        - command: /opt/application/upgrade_db.py
+          when: inventory_hostname == webservers[0]
+
 .. _local_playbooks:
 
 Local Playbooks
--- a/lib/ansible/playbook/__init__.py
+++ b/lib/ansible/playbook/__init__.py
@@ -396,6 +396,7 @@ class PlayBook(object):
             vault_pass = self.vault_password,
             run_hosts=hosts,
             no_log=task.no_log,
+            run_once=task.run_once,
         )
 
         runner.module_vars.update({'play_hosts': hosts})
--- a/lib/ansible/playbook/task.py
+++ b/lib/ansible/playbook/task.py
@@ -31,7 +31,7 @@ class Task(object):
         'local_action', 'transport', 'sudo', 'remote_user', 'sudo_user', 'sudo_pass',
         'items_lookup_plugin', 'items_lookup_terms', 'environment', 'args',
         'any_errors_fatal', 'changed_when', 'failed_when', 'always_run', 'delay', 'retries', 'until',
-        'su', 'su_user', 'su_pass', 'no_log',
+        'su', 'su_user', 'su_pass', 'no_log', 'run_once',
     ]
 
     # to prevent typos and such
@@ -41,7 +41,7 @@ class Task(object):
          'delegate_to', 'local_action', 'transport', 'remote_user', 'sudo', 'sudo_user',
          'sudo_pass', 'when', 'connection', 'environment', 'args',
          'any_errors_fatal', 'changed_when', 'failed_when', 'always_run', 'delay', 'retries', 'until',
-         'su', 'su_user', 'su_pass', 'no_log',
+         'su', 'su_user', 'su_pass', 'no_log', 'run_once',
     ]
 
     def __init__(self, play, ds, module_vars=None, default_vars=None, additional_conditions=None, role_name=None):
@@ -122,6 +122,7 @@ class Task(object):
         self.environment  = ds.get('environment', {})
         self.role_name    = role_name
         self.no_log       = utils.boolean(ds.get('no_log', "false"))
+        self.run_once     = utils.boolean(ds.get('run_once', 'false'))
 
         #Code to allow do until feature in a Task 
         if 'until' in ds:
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -144,6 +144,7 @@ class Runner(object):
         vault_pass=None,
         run_hosts=None,                     # an optional list of pre-calculated hosts to run on
         no_log=False,                       # option to enable/disable logging for a given task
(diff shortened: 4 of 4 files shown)


mpdehaan:
I can definitely see value for this though most of my question is in how we present it.

Should this maybe be something like "just_once: True" or even "delegate_once: localhost" or even "run_only: N" (times) (I thik I like this last one less)

I'm not entirely sure, though in either case we'll need to add this to an appropriate docs section -- it's a bit of an advanced concept and I don't want to clutter the main paths with the complex workarounds, but again, I can see this being useful.

Thoughts?

sivel:
I did look for an existing docs section and did not find one.  So I think we might need to add one.

I'm open to modification.  I was just sticking with existing (internal) terms already in use, for a true/false configuration.

But looking at your suggestion I do like "delegate_once" and specifying the host to delegate to.

sivel:
I am wondering, if we could still leave it as 'bypass=true/false' and pair it with 'delegate_to' to achieve 'delegate_once' as opposed to a actually creating a delegate_once option.

But I do think that 'bypass' might be misleading, so something along the lines of 'just_once' sounds good.  I just am not sure I like the name 'just_once'.

mpeters:
I'm definitely in favor of this feature, but I'm not sold on the "bypass" keyword. That makes it seems like it's going to be skipped completely. I also don't like "delegate_once" because then you have to specify the server to run on it. If we had a "just_once" flag then it could be combined with a "delegate" option to achieve that. So I like the semantics that just_once implies even though that name seems a little weak.

BTW, as far as where in the documentation this would live, I'd suggest the docs on rolling upgrades/deployment as that's where serial is discussed as well and they seem similar in "advancedness". Plus I think this feature is useful for deployment.

mpeters:
Some other options for the name:
- first_host
- single_host
- run_once
- execute_once
- host_limit (more like run_only)

jimi-c:
My $0.02 is for `single_host: <hostname>`, with optional syntax for specifying the first or last host in the group.

mpeters:
I was just thinking `single_host: true` not with an actual hostname. If you wanted that you could do `single_host: true` and `delegate_to: <hostname>`. I don't think `single_host` should do what `delegate_to` does as well.

sivel:
My only "concern" with `single_host: <hostname>` is needing to then handle also specifying "delegate_to", or even "local_action".

Just some things that need to be taken into account.

I think if we leverage existing local_action and delegate_to functionality paired with some type of "only run me once" option, that it reduces points of confusion.

```
- debug: var=inventory_hostname
  single_host: true
  delegate_to: foobar
```

If delegate_to is not present, then it functions like add_host, and just does hosts[0].

However, I think the wording of "single_host" then seems strange when using "delegate_to".

mpeters:
But I could see allowing `single_host: true`, and `single_host: first` or `single_host: last`.

mpeters:
I agree with @sivel. Having composable features is good. So the parameter should probably not have `host` in the name. So something like `run_once`, etc.

jimi-c:
They could be made mutually exclusive, as su and sudo are now, but I'm fine with making them composable as you say. In that case it should be just `single_host: yes` (or whatever boolean string you prefer) and let delegate_to/local_action define the target host - the default being to act on the first host in the list as BYPASS does now.

bcoca:
do we really need extra semantics for what can already be done ?

when: inventory_hostname == play_hosts[0]

it is not as pretty,  but I find that adding semantics that don't simplify
things by a lot just complicates the tool needlessly.

mpeters:
I can see your point @bcoca if that's all it did. But I was hoping that it would also be an option available to whole plays so you didn't have to litter multiple tasks with it.

Also, if you have dozens or hundreds of machines in a group, then that's a lot of skips to wait on and fill the screen with. If the pause module can do it, it would be nice it other tasks could too without having to write a whole custom module for it.

But I also understand that littering a big generic project like ansible with lots of custom nice-to-haves is problematic. I'm just hoping this is useful enough to overcome that threshold.

jimi-c:
@bcoca we had discussed this, where you could achieve this using "with" and "delegate_to", however that's a lot of boilerplate that might need to be repeated endlessly, so we thought it'd be worth-while to have a language feature for it specifically.

sivel:
I'll start working on the improvements for this soon and post updates as I progress.

sivel:
I've renamed `bypass` to `run_once` and squashed and pushed.

This functionality will already allow use with `delegate_to` without any additional modifications.

```
---
- hosts: all
  gather_facts: False
  tasks:
    - command: hostname
      run_once: True

    - command: hostname
      run_once: True
      delegate_to: srv3.example.org
```

```

TASK: [command hostname] ******************************************************
changed: [localhost] => {"changed": true, "cmd": ["hostname"], "delta": "0:00:00.003770", "end": "2014-05-19 11:10:27.091348", "rc": 0, "start": "2014-05-19 11:10:27.087578", "stderr": "", "stdout": "localhost.localdomain"}

TASK: [command hostname] ******************************************************
changed: [localhost] => {"changed": true, "cmd": ["hostname"], "delta": "0:00:00.008851", "end": "2014-05-19 11:10:27.446403", "rc": 0, "start": "2014-05-19 11:10:27.437552", "stderr": "", "stdout": "srv3.example.org"}

localhost                  : ok=2    changed=2    unreachable=0    failed=0
srv1.example.org           : ok=2    changed=2    unreachable=0    failed=0
srv2.example.org           : ok=2    changed=2    unreachable=0    failed=0
srv3.example.org           : ok=2    changed=2    unreachable=0    failed=0
```

The delegation portion of the playbooks documentation has been updated to include a 'run_once' section.

This isn't supported at the play level (currently).  I'm not sure it's needed there, but if someone thinks it should be, I can always look into it.  I figure the "hosts" specification should just be specified appropriately to only target the host you want to execute on.

iamjvn:
I have a scenario where this will be useful in a rolling upgrade playbook. 
- I currently use "when: inventory_hostname == play_hosts[0]" on every task in a "pre play" to run those cluster pre-configuration tasks only once from only one of the servers in the group. 
- This "pre play" is then followed by the actual rolling upgrade play touching every server in the group.
- Finally, I use "when: inventory_hostname == play_hosts[0]" again, on every task in a "post play" to run those cluster post-configuration tasks only once from only one of the servers in the group. 

Unless there's a better way..

hoettle:
+1 re the use of pre / post play tasks..especially for singleton actions such as DB schema upgrades or some other non-repeated activity. A run_once keyword would definitely improve readability in this case.

danillos:
+1

jimi-c:
Merged, thanks!

jimsmith:
Hello there I have this running ansible 1.7 and run_once does not work.

I have a DB schema update and other one time only tasks that need to run, I am also using delegate_to and serial as well so it and the playbook honors the delegate_to and serial set as 1 but the run_once tasks are still being ran more than 'once' on the delegate_to host that I've assigned.

```
 - name: run django migrations
  command: "/path/to/db_migrations.sh"
  # run_once: true -- this action will be availabe in ansible version 1.7, which is dev version now.
  # run_once doesn't work in current release, so 'delegate_to' should be specified to be executed for particular host
  run_once: True
  delegate_to: "192.168.1.2"      
  register: run_migrations
  # Print all contents of the shell task's output.
  - debug: var=run_migrations.stdout_lines
```

So the above is not being honored fully, as I have 2 hosts and the command gets ran more than once.

?

PR #7426, opened 2014-05-15
adding disable_plugins options to yum module

mpeters:
This branch adds yum options for `noplugins` and `disableplugin`. `disableplugin` behaves very similar to `disablerepo` and accepts a comma separated list of plugins to disable.

### DIFF
--- a/library/packaging/yum
+++ b/library/packaging/yum
@@ -120,8 +120,8 @@ EXAMPLES = '''
 - name: install the latest version of Apache from the testing repo
   yum: name=httpd enablerepo=testing state=installed
 
-- name: upgrade all packages
-  yum: name=* state=latest
+- name: upgrade all packages without the fastestmirror plugin
+  yum: name=* state=latest disable_plugin=fastestmirror
 
 - name: install the nginx rpm from a remote repo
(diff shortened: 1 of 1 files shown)


mpdehaan:
I'd prefer to see this as disable_plugins with a special value of "all".

We require variables to have "underscores_between_things" basically.

Various functions are also getting a little long on the argument signature and may wish to move to key-value arguments

If you'd like to update this first part, we'd appreciate it, and we can probably take care of the latter at a later date.

mpeters:
> I'd prefer to see this as disable_plugins with a special value of "all".

That makes sense.

> We require variables to have "underscores_between_things" basically.

I actually prefer this too, but I was trying to follow the original conventions of this module (and the yum command line tool itself) which uses `disablerepo`, `enablerepo`.

mpeters:
Changes made.

amenonsen:
@mpeters any chance you could rebase this and submit it to the `ansible-modules-core` repositorry? This PR (against the `ansible` repository) should be closed.

gregdek:
Yep, this should be against ansible-core-modules. Closing.

PR #7438, opened 2014-05-17
Setting the default of wait to false

Osterjour:
The known (and usual) behavior of this is a callback right after the AWS API returns and not wait until the snapshot is complete. So the default of the (undocumented) wait param should be false (which is the recommended way of doing snapshots on EBS).

### DIFF
--- a/library/cloud/ec2_snapshot
+++ b/library/cloud/ec2_snapshot
@@ -122,7 +122,7 @@ def main():
             ec2_url = dict(),
             ec2_secret_key = dict(aliases=['aws_secret_key', 'secret_key'], no_log=True),
             ec2_access_key = dict(aliases=['aws_access_key', 'access_key']),
-            wait = dict(type='bool', default='true'),
+            wait = dict(type='bool', default='false'),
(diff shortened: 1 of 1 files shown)


mpdehaan:
This shouldn't be accepted as it changes default behavior, rather the wait parameter should be documented.

The reason for this is many people may already be depending on the snapshot.

Please submit a ticket about the missing documentation and we can take care of this.

Thanks!

Osterjour:
Sorry but YOU changed the default behavior with merging this pull request in the first place: https://github.com/ansible/ansible/pull/6697
This change just fixes it back to the normal behavior (since the module was created) which is expected by all users of the AWS API. Amazon itself recommends not to wait because it isn't necessary.
I struggled with this after a update from Ansible 1.5 to 1.6 and had to find out why my backup-script now need more than 4 hours instead of 30 seconds!

mariow:
My comment from #6697 applies here as well. The default was and still should be not to wait for the snapshot to finish because this may take a very long time. Even if the parameter wasn't there before, the behavior was as if wait had been ==false.

PR #7439, opened 2014-05-17
Implement a per-task check_mode setting

amenonsen:
This allows an individual task to be run in check mode, so that other
tasks can be made conditional on its result.

Author: Carlos Chapi carlos@2ndQuadrant.com

### DIFF
--- a/lib/ansible/playbook/__init__.py
+++ b/lib/ansible/playbook/__init__.py
@@ -358,6 +358,11 @@ class PlayBook(object):
         hosts = self._trim_unavailable_hosts(task.play._play_hosts)
         self.inventory.restrict_to(hosts)
 
+        if task.check_mode:
+            check_mode = True
+        else:
+            check_mode = self.check
+
         runner = ansible.runner.Runner(
             pattern=task.play.hosts,
             inventory=self.inventory,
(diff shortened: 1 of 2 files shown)


bcoca:
isn't this redundant with always_run?​

amenonsen:
No, the purpose is not to run the task when --check is specified (which is what always_run does), but to run (only) one task in check mode when --check is NOT specified.

This feature was actually suggested by @mpdehaan in a long-ago thread on the mailing list (2013-08-13!) in response to my proposal for an "apt_wouldinstall" module to find out if certain modules were installed or not, without actually installing them (something for which there is unfortunately no simple command line invocation of apt-get).

bcoca:
thanks for clarifying, nice new feature!

## 

Brian Coca
Pedo mellon a minno

mpdehaan:
I don't recall suggesting this actually, in either case, I don't believe this should be exposed as a language keyword as I find it a little confusing.

I do recall something along the lines of saying there shouldn't be a seperate module for this.

But I'm trying to control language sprawl and this seems to be a complicated thing without a lot of applications.

Feel free to bring up a thread on ansible-devel if you'd like to discuss more.

Issue #7265, opened 2014-05-03
apt: upgrade=[safe | full | dist] ignores "default_release" parameter


bfqrst:
##### Issue Type:

Bug Report
##### Ansible Version:

Ansible 1.5.4, 1.5.5, 1.6
##### Environment:

Managing machine OS: OS X Mavericks
Target machine OS: Debian Wheezy
##### Summary:

Suppose you add the Wheezy backports repo via apt_repository to the target system and want to perform any of the safe-upgrade, dist-upgrade, full-upgrade INCLUDING the backports repo, the latter will be ignored. This is due to the fact that in the upgrade function of the apt module lacks a check for the default_release parameter. The install function on the other hand has such a check which will correctly install packages from alternative repos when passed.
##### Steps To Reproduce:

In my playbook I add the backports repo and want to perform the safe-upgrade:

```
    - name: Add Wheezy Backports
      apt_repository: repo='deb http://ftp.de.debian.org/debian wheezy-backports main' state=present update_cache=yes
    - name: Initial safe-upgrade of the system (including wheezy-backports)
      apt: upgrade=safe default_release=wheezy-backports update_cache=yes dpkg_options=force-confold
```
##### Expected Results:

The expected result would be that all upgradable packages (including those from backports, since default_release was passed) are installed and the second run of the task would leave me with a "green" OK result. In apt-get as well as in aptitude terms that would translate to the "-t alternative_repo" switch on the target machine.
##### Actual Results:

Only the upgradable packages from the default repo are installed and whatever repo is passed to the default_release parameter is ignored. Shelling to the target machine and manually firing the apt-get or aptitude commands, proves that there are still upgradable packages.

jimi-c:
The above PR has been merged, which should address this issue. If you continue seeing any problems related to this, please let us know. Thanks!

PR #7440, opened 2014-05-17
Allow '-t ' to be passed to apt: upgrade

amenonsen:
Without this, the following task:

```
- name: Try to run "apt-get dist-upgrade -t wheezy-backports"
  apt: upgrade=dist default_release=wheezy-backports
```

would run "apt-get dist-upgrade", because the default_release parameter is ignored for upgrade().

### DIFF
--- a/library/packaging/apt
+++ b/library/packaging/apt
@@ -367,7 +367,7 @@ def remove(m, pkgspec, cache, purge=False,
             m.fail_json(msg="'apt-get remove %s' failed: %s" % (packages, err), stdout=out, stderr=err)
         m.exit_json(changed=True, stdout=out, stderr=err)
 
-def upgrade(m, mode="yes", force=False,
+def upgrade(m, mode="yes", force=False, default_release=None,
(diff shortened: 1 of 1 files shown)


mpdehaan:
p['default_release'] as a syntax seems to imply this will have a value which looks like it will cause problems (tracebacks) when not specified.

amenonsen:
I've tested an "apt: upgrade=dist" task without specifying default_release, and there's no traceback. As I understand it, p['default_release'] is referring to something setup by the module's argument_spec:

```
default_release = dict(default=None, aliases=['default-release']),
```

There's other code that accesses p['default_release'] assuming a default of None too.

bfqrst:
...seems to me this is a duplicate of #7265 , right?

amenonsen:
Well, it's the same underlying problem, but I've included a patch to fix it. I didn't notice your bug report, though. Thanks for the pointer.

jimi-c:
Merged, thanks!

PR #7444, opened 2014-05-18
Improvements to OpenStack inventory script

carsongee:
This adds the option to use common environment variables instead of the just the INI for common things like password and auth URL.  It creates inventory groups for each key/value in the metadata dictionary similar to the ec2.py scripts (but keeps the special group metadata tag as well for backwards compatibility), and adds the _meta host var dictionary added in 1.3 that speeds things up quite a bit. It also adds a configurable flag as to whether to prefer the public or private IP (fixed or floating) of each instance, and prints the json in a more human friendly format.

### DIFF
--- a/plugins/inventory/nova.ini
+++ b/plugins/inventory/nova.ini
@@ -14,13 +14,20 @@ api_key       =
 auth_url      =
 
 # Authentication system
-auth_system =
+auth_system = keystone
 
 # OpenStack nova project_id
 project_id    = 
 
 # Serverarm region name to use
 region_name   =
+
+# Specify a preference for public or private IPs (public is default)
+prefer_private = False
+
+# What service type (required for newer nova client)
+service_type = compute
+
 
 # TODO: Some other options
 # insecure      =
--- a/plugins/inventory/nova.ini
+++ b/plugins/inventory/nova.ini
@@ -1,37 +1,45 @@
 # Ansible OpenStack external inventory script
 
 [openstack]
+
+#-------------------------------------------------------------------------
+#  Required settings
+#-------------------------------------------------------------------------
+
 # API version
 version       = 2
 
 # OpenStack nova username
 username      =
 
-# OpenStack nova api_key
+# OpenStack nova api_key or password
 api_key       =
 
 # OpenStack nova auth_url
 auth_url      =
 
-# Authentication system
-auth_system = keystone
+# OpenStack nova project_id or tenant name
+project_id    =
 
-# OpenStack nova project_id
-project_id    = 
+#-------------------------------------------------------------------------
+#  Optional settings
+#-------------------------------------------------------------------------
+
+# Authentication system
+# auth_system = keystone
 
 # Serverarm region name to use
-region_name   =
+# region_name   =
 
 # Specify a preference for public or private IPs (public is default)
-prefer_private = False
+# prefer_private = False
 
 # What service type (required for newer nova client)
-service_type = compute
+# service_type = compute
 
 
 # TODO: Some other options
 # insecure      =
 # endpoint_type =
 # extensions    =
-# service_type  =
 # service_name  =
--- a/plugins/inventory/nova.py
+++ b/plugins/inventory/nova.py
@@ -39,6 +39,7 @@ NOVA_CONFIG_FILES = [os.getcwd() + "/nova.ini",
 NOVA_DEFAULTS = {
     'auth_system': None,
     'region_name': None,
+    'service_type': 'compute',
 }
 
 
--- a/plugins/inventory/nova.py
+++ b/plugins/inventory/nova.py
@@ -25,10 +25,8 @@ from novaclient import client as nova_client
 
 try:
     import json
-except:
+except ImportError:
     import simplejson as json
-
-from ansible.module_utils.openstack import *
 
 ###################################################
 # executed with no parameters, return the list of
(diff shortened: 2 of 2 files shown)


mpdehaan:
The documentation on the new parameters is definitley a lot on the sparse side and will need to be improved if we are to consider this.

_meta is of course a great addition - perhaps these should be submitted seperately?

carsongee:
I've gone ahead and added documentation for the prefer_private option.  Oddly enough there was already doc for the service_type even though that parameter wasn't used at all in the script. The existing doc there is accurate though.  

Most of the changes other than that one were just refactors and additional error checking (it was raising in my environment).  It will be a bit hard to tease out the different changes since a lot of them are interdependent on the refactors.

msabramo:
This seems like a few different things in one PR.
1. Typo fix
2. Use `OS_*` environment variables that python-novaclient uses
3. Some new parameters and logic
4. Indented JSON output

I didn't look in depth yet but I know that we wanted 2 and I believe that someone forked nova.py internally so we could use those environment variables. So definitely looking forward to that one!

plugins/inventory/nova.py
```diff
@@ -111,12 +111,17 @@
       - In RackSpace use 'cloudServersOpenStack'
     required: false
     default: null
-  insicure:
+  insecure:
```
msabramo: AFAICT, this parameter is not used (even when spelled correctly :-)), along with a few others. See https://github.com/ansible/ansible/pull/7988

plugins/inventory/nova.py
```diff
@@ -160,46 +165,148 @@ def nova_load_config_file():
...
-    region_name = config.get('openstack', 'region_name'),
-    project_id  = config.get('openstack', 'project_id'),
-    auth_system = config.get('openstack', 'auth_system')
+    version=config.get('openstack', 'version'),
+    username=username, 
+    api_key=api_key, 
+    auth_url=auth_url,
+    region_name=config.get('openstack', 'region_name'),
+    project_id=project_id,
+    service_type=config.get('openstack', 'service_type'),
```
msabramo: It looks like you added support for `service_type`, but removed `auth_system`? Was that intentional?

plugins/inventory/nova.py
```diff
@@ -160,46 +165,148 @@ def nova_load_config_file():
...
 )
 
-if len(sys.argv) == 2 and (sys.argv[1] == '--list'):
-    groups = {}
-
+# Default or added list option
+if (len(sys.argv)==2 and sys.argv[1]=='--list') or len(sys.argv)==1:
```
msabramo: I suspect (from looking at https://github.com/ansible/ansible/blob/devel/CODING_GUIDELINES.md#misc-preferences) that @mpdehaan would prefer spaces around `==` a la PEP8.

plugins/inventory/nova.py
```diff
@@ -160,46 +165,148 @@ def nova_load_config_file():
...
+        group = server.metadata['group'] if server.metadata.has_key('group') else 'undefined'
+        push(groups, group, access_ip)
+
+        # Add vars to _meta key for performance optimization in
+        # Ansible 1.3+
+        groups['_meta']['hostvars'][access_ip] = get_metadata(server)
 
     # Return server list
-    print json.dumps(groups)
+    print(json.dumps(groups, sort_keys=True, indent=2))
```
msabramo: Cool. I like the Python 3 compatible `print` and the nicer JSON output. :+1:

plugins/inventory/nova.py
carsongee: I went ahead and removed this to try and avoid the merge conflict

plugins/inventory/nova.py
carsongee: No, it looks like it got missed in the shuffle. I added it back in

plugins/inventory/nova.py
carsongee: I fixed this and several other pep8/pylint violations

plugins/inventory/nova.ini
```diff
@@ -14,7 +14,7 @@ api_key       =
 auth_url      =
 
 # Authentication system
-auth_system =
+auth_system = keystone
```
msabramo: Just curious why you set `keystone` here when it used to be blank?

msabramo:
So internally we have a version of `nova.py` where we also used the `OS_*` python-novaclient environment variables but my colleague did this by changing it to use exclusively the variables and he eliminated the .ini file entirely. I wonder if it's preferable to simply eliminate the .ini file; we didn't really see the point of the .ini file when the OpenStack tooling already uses the environment variables.

What say you, @mpdehaan?

plugins/inventory/nova.py
```diff
@@ -158,48 +164,151 @@ def nova_load_config_file():
...
-    project_id  = config.get('openstack', 'project_id'),
-    auth_system = config.get('openstack', 'auth_system')
+    version=config.get('openstack', 'version'),
+    username=username,
+    api_key=api_key,
+    auth_url=auth_url,
+    region_name=config.get('openstack', 'region_name'),
+    project_id=project_id,
+    auth_system=config.get('openstack', 'auth_system'),
+    service_type=config.get('openstack', 'service_type'),
```
msabramo, line 176: So I tested your PR and adding `service_type` and not creating a default for it causes it to fail with an older .ini file that doesn't specify `service_type`. I submitted the following PR to fix this issue: https://github.com/carsongee/ansible/pull/1

msabramo:
@carsongee: @jimi-c merged 2 PRs of mine that touched the same file, so this no longer merges cleanly; can you rebase? (sorry for the trouble!)

carsongee:
No problem. Rebased

msabramo:
@carsongee: @jimi-c merged another PR of mine so, sorry, looks like you have to rebase again.

carsongee:
Rebased again

msabramo:
Cool, thanks @carsongee!

Wondering what @mpdehaan and @jimi-c think now; it's evolved a little bit since the initial comments.

msabramo:
@jimi-c: Is this something you'd consider merging? We'd like to use the feature where it looks in the OpenStack environment variables instead of the .ini file. That's just much more convenient since we already have the environment variables.

Cc: @sontek

jimi-c:
@msabramo yes we're interested in this, we're just continuing to work through our backlog of issues at the moment.

msabramo:
@jimi-c: Cool, thanks!

msabramo:
@carsongee: This doesn't merge cleanly anymore and needs a little love.

carsongee:
Rebased and fixed a small bug I'd ran into when instances have multiple addresses.

t0mk:
Hi, I did another rework of the Nova inventory, I tried to cover your features and add some more, and clean up the code. If you are interested, please see:
https://github.com/ansible/ansible/pull/9378

msabramo:
I'm surprised this is still open. Ansible folks, can you review?

Issue #7330, opened 2014-05-08
docker module traces with KeyError: 'Ghost'


ghost:
##### Issue Type:

Bug report
##### Ansible Version:

1.6
##### Environment:

Workstation: Archlinux
Managed node: Ubuntu Precise (12.04)
##### Summary:

Trying to install a private docker registry which results in a Python stack trace thrown as "KeyError: 'Ghost'".
##### Steps To Reproduce:

```
- hosts: all
  tasks:
    - name: Install private docker registry
      docker: image='samalba/docker-registry' name=registry detach=true ports=5000:5000
```
##### Expected Results:

Running docker container.
##### Actual Results:

Ansible breaks with Python stacktrace. In fact the container is running properly but the Ansible run breaks.

```
Traceback (most recent call last):
File "/tmp/ansible/kusch/ansible-tmp-1399540829.6-154372885089505/docker", line 1922, in <module>
main()
File "/tmp/ansible/kusch/ansible-tmp-1399540829.6-154372885089505/docker", line 713, in main
facts = manager.get_running_containers()
File "/tmp/ansible/kusch/ansible-tmp-1399540829.6-154372885089505/docker", line 527, in get_running_containers
  if i['State']['Running'] == True and i['State']['Ghost'] == False:
KeyError: 'Ghost'
```

etehtsea:
I have similar problem after upgrade to ansible 1.6.

```
failed: [docker-01.vbox] => {"failed": true, "parsed": false}
invalid output was: SUDO-SUCCESS-imajjsytebmqcqkoxmgplauttreznkcq
Traceback (most recent call last):
  File "/home/vagrant/.ansible/tmp/ansible-tmp-1399546185.36-39970679925453/docker", line 1922, in <module>
    main()
  File "/home/vagrant/.ansible/tmp/ansible-tmp-1399546185.36-39970679925453/docker", line 704, in main
    containers = manager.create_containers(delta)
  File "/home/vagrant/.ansible/tmp/ansible-tmp-1399546185.36-39970679925453/docker", line 546, in create_containers
    if docker.utils.compare_version('1.10', self.client.version()['ApiVersion']) < 0:
KeyError: 'ApiVersion'
```

mpdehaan:
What version is your docker python module?  Is it at the latest version?

etehtsea:
I've got first error with ansible 1.5.5/docker 0.11.1 and latter with ansible 1.6.1/docker 0.9.1.

[...]

PR #7445, opened 2014-05-18
Fix for https://github.com/ansible/ansible/issues/7330

CaptTofu:
i['State']['Ghost'] no longer seems to be defined with newer versions of docker-py. Added logic to handle this.


CaptTofu:
closing, Must've had an old clone somehow, though I was sure I had rebased.

Issue #7368, opened 2014-05-12
PR to resolve the issue with ec2_vpc module not allowing to use instance id in route_tables


romanrev:
##### Issue Type: Bugfix Pull Request

See [pull request to allow the use of an instance id in the route creation in ec2_vpc (for NAT private networks)](https://github.com/romanrev/ansible/pull/1) for a possible solution
##### Environment:

Running on OS X 10.9.2 managing AWS VPC
##### Summary:

When I am trying to add a route table to a VPC that uses a NAT instance id as a gateway, I get the error like follows:

```
failed: [127.0.0.1] => {"failed": true, "item": ""}
msg: Unable to create and associate route table {u'routes': [{u'dest': u'0.0.0.0/0', u'gw': u'i-09d47501'}], u'subnets': [u'10.1.10.0/24', u'10.1.12.0/24', u'10.1.20.0/24']}, error: EC2ResponseError: 400 Bad Request
<?xml version="1.0" encoding="UTF-8"?>
<Response><Errors><Error?<Code>InvalidGatewayID.NotFound</Code <Message>The gateway ID 'i-09d47501' does not exist</Message></Error></Errors><RequestID>fa95104e-37ab-44ee-b8fd-2c5033179585</RequestID></Response>
```
##### Steps To Reproduce:
1. create a VPC using a `ec2_vpc` Ansible module
2. launch a NAT instance into a subnet with the internet gateway in it
3. obtain NAT instance id 
4. re-run `ec2_vpc` Ansible module with a record in the `routes_table` assigning the NAT instance ID as a gateway in the route to '0.0.0.0/0' network destination. 

I.e. example

```
route_tables:
  - subnets:
      - '10.1.0.0/24'
      - '10.1.1.0/24'
    routes:
      - dest: 0.0.0.0/0
        gw: igw
  - subnets:
     - '10.1.10.0/24'
     - '10.1.12.0/24'
     - '10.1.20.0/24'
    routes:
      - dest: 0.0.0.0/0
        gw: 'i-09d47501'
```
##### Expected Results:

The VPC should be updated with the new route table defined that directs all internet traffic from the designated private subnets to the NAT instance.
##### Actual Results:

Getting the following error:

```
failed: [127.0.0.1] => {"failed": true, "item": ""}
[...]

PR #7447, opened 2014-05-19
Fix to resolve VPC NAT routing issue https://github.com/ansible/ansible/issues/7368

romanrev:
See https://github.com/ansible/ansible/issues/7368 for a detailed description

According to [boto.vpc.VPCCOnnection.create_route](http://bit.ly/1sku9W6) there is a need to use named parameters in the method call to differentiate between a gateway ID and an instance ID when trying to create a route.

there is a need to use named parameters in the method call
to differentiate between a gateway ID and an instance ID when
trying to create a route

Otherwise upon trying to register a route with an existing instance ID
instead of the gateway ID one gets an error like follows:

```
failed: [127.0.0.1] => {"failed": true, "item": ""}
msg: Unable to create and associate route table {u'routes': [{u'dest': u'0.0.0.0/0', u'gw': u'i-09d47501'}], u'subnets': [u'10.1.10.0/24', u'10.1.12.0/24', u'10.1.20.0/24']}, error: EC2ResponseError: 400 Bad Request
<?xml version="1.0" encoding="UTF-8"?>
<Response><Errors><Error><Code>InvalidGatewayID.NotFound</Code><Message>The gateway ID 'i-09d47501' does not exist</Message></Error></Errors><RequestID>fa95104e-37ab-44ee-b8fd-2c5033179585</RequestID></Response>
```

### DIFF
--- a/library/cloud/ec2_vpc
+++ b/library/cloud/ec2_vpc
@@ -422,7 +422,10 @@ def create_vpc(module, vpc_conn):
                             '(igw) route, but you have no Internet Gateway'
                         )
                     r_gateway = igw.id
-                vpc_conn.create_route(new_rt.id, route['dest'], r_gateway)
+                if r_gateway[:3] == 'igw':
+                    vpc_conn.create_route(new_rt.id, route['dest'], gateway_id=r_gateway)
+                else:
+                    vpc_conn.create_route(new_rt.id, route['dest'], instance_id=r_gateway)
 
             # Associate with subnets
             for sn in rt['subnets']:


mpdehaan:
Thanks for this submission.

FYI, in the future, it's ok to just say #7368 and github will link things up automatically, versus pasting the URL.

romanrev:
Somehow I missed the notification for the reply in this thread. 
@mpdehaan: Could you tell how long it will take to review and merge the fix? I understand you've got stacks of PRs, but this one seems to me to be a fairly simple and straightforward one.
Thanks!

jimi-c:
@romanrev I've gone ahead and merged in @sayap's fix to this, so this issue should be resolved and I will close this pull request. If you continue seeing any problems related to #7368, please let us know. Thanks!

romanrev:
Thanks, @jimi-c, much appreciated!

PR #7464, opened 2014-05-20
Enable use of vault password for --syntax-check, --list-hosts and --list...

carlanton:
This patch makes it possible to use --syntax-check, --list-hosts and --list-tasks on a playbook that uses encrypted vars_files.

The old behavior looks like this:

```
$ cat playbook.yml
- hosts: some-host
  vars_files:
    - vars.yml # <- encrypted with ansible-vault

  tasks:
    - debug: msg={{ test_var }}

$ ansible-playbook --vault-password-file my-password-file --syntax-check playbook.yml

playbook: playbook.yml

ERROR: A vault password must be specified to decrypt data
```


jimi-c:
Merged, thanks!

PR #7467, opened 2014-05-20
Support internal-only ELBs in EC2.

asnare:
Pull request #7051 recently added support for attaching ELBs to VPC subnets. This pull request extends this with support for making internal ELBs. This is via the 'scheme' attribute when creating the ELB: normally ELBs use the default of 'internet-facing', but 'internal' means the ELB will only be reachable from within the VPC. (There doesn't appear to be any way to update it after it has been created.)

Tested with the following playbook:

```
- hosts: 127.0.0.1
  connection: local
  tasks:
  - name: VPC load balancer
    local_action:
      module: ec2_elb_lb
      name: "demo-vpc-elb-{{ item }}"
      region: "eu-west-1"
      scheme: "{{ item }}"
      subnets:
        - subnet-XXXXXXXX
      listeners:
        - protocol: http
          load_balancer_port: 80
          instance_port: 80
      state: present
    with_items:
      - internet-facing
      - internal
```

### DIFF
--- a/library/cloud/ec2_elb_lb
+++ b/library/cloud/ec2_elb_lb
@@ -78,6 +78,11 @@ options:
       - Purge existing subnet on ELB that are not found in subnets
     required: false
     default: false  
+  scheme:
+    description:
+      - The scheme to use when creating the ELB. For a private VPC-visible ELB use 'internal'.
+    required: false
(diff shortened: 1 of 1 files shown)


jimi-c:
Merged, thanks!

Issue #7465, opened 2014-05-20
alternatives: /usr/sbin/update-alternatives is not valid on all distributions


ThomasNunninger:
##### Issue Type:

“Bug Report”
##### Ansible Version:

ansible 1.6
##### Environment:

Running: Ubuntu 12.04.4 LTS
Managed: Ubuntu 14.04 LTS
##### Summary:

Module alternatives uses wrong binary path on Ubuntu 14.04.
##### Steps To Reproduce:

Sorry, I'm new to ansible and I don't know how to write a reproducible script in a short time.
##### Expected Results:

/usr/bin/update-alternatives is used instead of /usr/sbin/update-alternatives
##### Actual Results:

```
failed: [server.local] => {"cmd": "/usr/sbin/update-alternatives --query editor", "failed": true, "item": "", "rc": 2}
msg: [Errno 2] No such file or directory
```

jimi-c:
I have merged in the patch above, which should resolve this issue. Please let us know if you continue having any problems related to this. Thanks!

PR #7470, opened 2014-05-21
now module searches for update-alternatives vs hardcoding it

bcoca:
should solve #7465, tested against ubuntu 12.04 and 14.04


jimi-c:
Merged, thanks! There are actually a few other issues with alternatives, if you're interested in working on that module. I've noticed the output from that command differs on EL/Fedora systems vs. Ubuntu (it seems the RH-based distros are using a much older version).

bcoca:
I'll check the backlog, I'm replacing a lot of shell with this module

Brian Coca

PR #7472, opened 2014-05-21
EC2: stopped instances can also be terminated

ghost:
The EC2 API and dashboard allow one to terminate "stopped" instances too. This simple commit ensures Ansible can do the same thing.


jimi-c:
Merged, thanks!

PR #7477, opened 2014-05-21
rax module: Support config_drive and user_data

sivel:
This pull request adds attributes for `config_drive` and `user_data`.  This is primarily useful for utilizing cloudinit.

The `user_data` attribute can be a string or a file path.  If `user_data` is supplied on it's own, it implies `config_drive`.  `config_drive` can be used on it's own.

Here is an example task:

``` yaml
    - name: Build CoreOS Servers
      rax:
        credentials: ~/.raxpub
        name: coreos%02d
        flavor: performance1-1
        image: 70210a69-081e-4a5e-b402-3cdbabfdcddd
        region: ORD
        state: present
        count: 3
        exact_count: yes
        group: coreos
        wait: yes
        user_data: files/coreos-cloud-config.yml
```

### DIFF
--- a/library/cloud/rax
+++ b/library/cloud/rax
@@ -136,7 +136,7 @@ options:
   user_data:
     description:
       - Data to be uploaded to the servers config drive. This option implies
-        I(config_drive)
+        I(config_drive). Can be a file path or a string
     version_added: 1.7
   wait:
     description:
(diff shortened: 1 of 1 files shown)


sivel:
I believe this will need rebased. I'll look tomorrow.

mpdehaan:
The create function here is starting to sprawl with positional arguments.  This really needs to be updated to use named arguments at least as the order is getting difficult to follow.

Thanks!

sivel:
That is actually part of what this pull request does, switch from positional args to keyword args.  This is one reason the rebase is required, since a pull that was just merged added another positional arg that will need converted to a keyword arg.

sivel:
Rebase complete. Was a little more simple than I thought. The 2nd commit in the PR switches to keyword based args.

jimi-c:
Merged, thanks!

PR #7478, opened 2014-05-21
rax_keypair module: Allow public_key to be a path to a file

sivel:
This pull request allows you to pass a file path to as the public_key parameter.

This simplifies the workflow of having to potentially use `lookup('file', ...)` to pass a key into the module from a file.

This change should be backwards compatible.

### DIFF
--- a/library/cloud/rax_keypair
+++ b/library/cloud/rax_keypair
@@ -30,7 +30,7 @@ options:
     required: true
   public_key:
     description:
-      - Public Key string to upload
+      - Public Key string to upload. Can be a file path or string
     default: null
   state:
     description:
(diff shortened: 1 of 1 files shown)


mpdehaan:
This might be a little confusing if I named my SSH keypair "/etc/motd" but that's grasping at straws a bit.

Curious if it should be a different parameter like "key_file" here (guessing no), or just show using the lookup plugin in the example.

I may be a bit worried about consistency issues.

sivel:
If you specify a file that exists but not containing a public key, the API will respond with an error.  A string with "/etc/motd" would fail as well, since that is not a public key.

jimi-c:
Merged, thanks!

PR #7479, opened 2014-05-21
Update django_manage `command` argument description

mwarkentin:
The docs were a little bit out of date with what commands are available to be run. They also didn't explain that you could pass custom commands - I almost went down the path of trying to run our custom management commands with the generic Ansible `command` module.


jimi-c:
Merged, thanks!

PR #7480, opened 2014-05-21
Add an option to ec2.ini to filter instances based on various things

tgerla:
(This is try two for the PR, replacing #7359 which ended up with merge commits. See the previous PR for a bit of commentary.)

This lets you filter instances right at the EC2 API level using the EC2 inventory plugin. You can use this to retrieve only instances with certain tags, or only certain instance types, etc, etc. See ec2.ini for more info and examples.

### DIFF
--- a/plugins/inventory/ec2.ini
+++ b/plugins/inventory/ec2.ini
@@ -53,3 +53,21 @@ cache_path = ~/.ansible/tmp
 # seconds, a new API call will be made, and the cache file will be updated.
 # To disable the cache, set this value to 0
 cache_max_age = 300
+
+# Instance filters can be used to control which instances are retrieved for
+# inventory. For the full list of possible filters, please read the EC2 API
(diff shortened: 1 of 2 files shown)


mpdehaan:
Just to confirm, this is the "AND" filtering on multiple tags, yes?

Thanks!

mpdehaan:
sorry, I mean "OR".

Explanation:  I am speculating users would need to suck down hosts tagged "team1" or "team2" at the same time.

tgerla:
This only does "AND" right now. "OR" will take a bit of work to specify an "or" syntax and then make multiple calls to the EC2 API, since the API only supports "AND" for filters in a single call.

mpdehaan:
Fun.

My thinking is we can only accept this if it does "OR", which means we may have to combine results in memory, while people could combine multiple groups in Tower, it would be too hard for the core application if they needed that behavior.

I think it would be fine if it only did tags and not some of the other features, so this should be ok.

Thus, we don't need a new syntax, we could just provide a comma separated list of tags.

mpdehaan:
Closing this one given we have the other, thanks!

PR #7487, opened 2014-05-22
Clean non-printable chars from stdout instead of dropping the whole thin...

### LINKED ISSUE #7486: script module drops stdout if it contains any non-printable characters
##### Summary:

When using the script module, stdout gets dropped (replaced with '') if it contains non-printable characters.  They should be stripped instead.
##### Steps To Reproduce:

``` bash
    $ cat script.sh
    echo -e \\a critical information

    $ ansible -i $target, -m script -a script.sh $target
```
##### Expected Results:

``` bash
$ ansible -i $target, -m script -a script.sh $target
$target | success >> {
    "changed": true,
    "rc": 0,
    "stderr": "",
    "stdout": " critical information\r\n"
}
```
##### Actual Results:

``` bash
    $ cat script.sh
    echo -e \\a critical information

    $ ansible -i $target, -m script -a script.sh $target
    $target | success >> {
        "changed": true,
        "rc": 0,
        "stderr": "",
        "stdout": ""
    }
```

I believe the problem lies in the following line (879 in lib/ansible/runner/**init**.py):

``` python
                    if not string_functions.isprintable(data['stdout']):
                                            data['stdout'] = ''
```

The correct action is probably to clean the string instead.  I'll whip up a pull request, but please review thoroughly.

nirvinehh:
Fixes #7486

### DIFF
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -877,7 +877,7 @@ class Runner(object):
             if hasattr(sys.stdout, "isatty"):
                 if "stdout" in data and sys.stdout.isatty():
                     if not string_functions.isprintable(data['stdout']):
-                        data['stdout'] = ''
+                        data['stdout'] = ''.join(c for c in data['stdout'] if string_functions.isprintable(c))
 
             if 'item' in inject:
                 result.result['item'] = inject['item']


nirvinehh:
A better fix might be to tell the terminal it can't do escape codes, but I'm not sure how to accomplish that.

PR #7491, opened 2014-05-22
Update docs to include directories inside group_vars

hkariti:
Related ansible-project thread: https://groups.google.com/forum/#!topic/ansible-project/VgGIyBhFtZM

Thought I shouldn't just complain about lack of docs...

### DIFF
--- a/docsite/rst/intro_inventory.rst
+++ b/docsite/rst/intro_inventory.rst
@@ -171,6 +171,17 @@ the 'raleigh' group might look like::
 
 It is ok if these files do not exist, as this is an optional feature.
 
+As an advanced use-case, you can create *directories* named after your groups or hosts, and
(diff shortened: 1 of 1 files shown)


jimi-c:
Could you add a note regarding when the directory feature was added? I'm not sure when it was, but we like to make sure those are noted, just in case. Thanks!

hkariti:
Ok, some `git checkout`-ing, and apparently it works since 1.4.0. Added a
note at the end.

mpdehaan:
Thanks...

note:  I may want to tweak wording a bit, so not merging this immediately.  Much appreciated!

mpdehaan:
Merged, thank you.

Apologies on the delay on this one, this will appear when we next rebuild and push docs, which should be soon.

Thanks again!

Issue #7492, opened 2014-05-22
[rax_*] Deal with rate limits


smola:
##### Issue Type:

Feature Idea
##### Ansible Version:

ansible 1.6
##### Environment:

N/A
##### Summary:

rax_\* should deal with rate limits.

I'm hitting rate limits on rax Cloud DNS API calls. That makes the whole playbook fail instead of retrying. It would be nice to have some throttling settings or respect the "Reply-After" response header and retry after the specified amount of time (or both).

Rate limits example:

http://docs.rackspace.com/cdns/api/v1.0/cdns-devguide/content/Rate_Limits-d1e1222.html
##### Steps To Reproduce:

Just go over rate limits playbook fails.
##### Expected Results:

When rate limit is found, it should retry after the amount of time specified by the API response. API calls might be throttled to mitigate the issue.
##### Actual Results:

Playbook fails. Example:

```
failed: [bran] => (item=bitsnbrains.com) => {"failed": true, "item": "bitsnbrains.com"}
msg: OverLimit Retry...
```

smola:
Note that this can be mitigated by using `serial: 1` on the playbook.

mpdehaan:
cc @sivel - any thoughts on this one?

I'm perhaps just suggesting a quick docs note and an acceptable serial: N suggestion.

sivel:
@mpdehaan yes, it is our recommendation to use serial with rax_dns.  I will add notes to rax_dns and rax_dns_record and submit a pull request soon.

mpdehaan:
Closing per sivel's docs change.

Thanks!

PR #7501, opened 2014-05-22
Add notes to rax_dns and rax_dns_record to recommend using serial: 1

sivel:
This pull request addresses #7492 in that the API limit for the Rackspace CloudDNS API can easily be exceeded due to the way that API works.

The recommendation is to use `serial: 1` in plays utilizing the `rax_dns` and `rax_dns_record` to avoid exceeding the API limits.


mpdehaan:
Merged, thank you.

Will appear in next docs push.

PR #7504, opened 2014-05-22
module rax_dns_record: Add PTR record support

sivel:
(was: New module: rax_dns_ptr)

This pull request adds PTR record support to rax_dns_record

Sample play:

```

---
- hosts: localhost
  gather_facts: False
  tasks:
    - name: Create server
      local_action:
        module: rax
        credentials: ~/.raxpub
        name: ptr.example.org
        flavor: performance1-1
        image: ubuntu-1204-lts-precise-pangolin
        disk_config: manual
        region: ORD
        state: present
        wait: yes
        wait_timeout: 360
      register: rax

    - name: Create PTR record
      local_action:
        module: rax_dns_record
        credentials: ~/.raxpub
        name: ptr.example.org
        type: PTR
        server: "{{ (rax.instances|first).rax_id }}"
        address: "{{ (rax.instances|first).rax_accessipv4 }}"
        region: ORD
        ttl: 3600
      when: rax.action != 'delete'

    - name: Modify PTR record
      local_action:
        module: rax_dns_ptr
        credentials: ~/.raxpub
        name: ptr.example.org
        type: PTR
        server: "{{ (rax.instances|first).rax_id }}"
        address: "{{ (rax.instances|first).rax_accessipv4 }}"
        region: ORD
        ttl: 900
      when: rax.action != 'delete'

    - name: Delete PTR record
      local_action:
        module: rax_dns_ptr
        credentials: ~/.raxpub
        name: ptr.example.org
        type: PTR
        server: "{{ (rax.instances|first).rax_id }}"
        address: "{{ (rax.instances|first).rax_accessipv4 }}"
        region: ORD
        state: absent
```

### DIFF
--- a/library/cloud/rax_dns
+++ b/library/cloud/rax_dns
@@ -45,9 +45,9 @@ options:
       - Time to live of domain in seconds
     default: 3600
 notes:
-  - "It is recommended that plays utilizing this module be run with C(serial: 1)
-    to avoid exceeding the API request limit imposed by the Rackspace CloudDNS
-    API"
+  - "It is recommended that plays utilizing this module be run with
+    C(serial: 1) to avoid exceeding the API request limit imposed by
+    the Rackspace CloudDNS API"
 author: Matt Martz
 extends_documentation_fragment: rackspace
 '''
--- a/library/cloud/rax_dns_record
+++ b/library/cloud/rax_dns_record
@@ -34,8 +34,12 @@ options:
     required: True
   domain:
     description:
-      - Domain name to create the record in
-    required: True
+      - Domain name to create the record in. This is an invalid option when
+        type=PTR
+  loadbalancer:
+    description:
+      - Load Balancer ID to create a PTR record for. Only used with type=PTR
+    version_added: 1.7
   name:
     description:
       - FQDN record name to create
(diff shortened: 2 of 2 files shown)


jimi-c:
Is there any reason this wasn't added to the rax_dns_record module? That one already supports multiple record types.

sivel:
Unfortunately the way PTR record creation works is significantly different than the way a regular record works, in that it has to specifically be tied to an actual device.

It's easier to manage them independently due to the different work flows and to make documentation more simple due to different required options.

mpdehaan:
Would the signature of methods be THAT different?

I too would like to contain the module sprawl here.

We have things like mutually_exclusive in argument spec that can help, and if the code needs to be made more OO or segmented, that could be done.

jimi-c:
I agree, this module is only ~200 lines, a lot of which is boilerplate. I'm thinking these modules can and should probably be merged.

sivel:
I'll see if I can figure it out.  I'd have to create a new function and call either the existing one or the old one depending on the record type which is not a huge issue.

The big thing is losing the ability to use `required_one_of` and needing to implement that functionality on my own.

Additionally, PTR records differ from the other record types in that they only allow a subset of attributes to be updated and being able to properly articulate that in the documentation as well as when certain configurations are needed and when they aren't, and not making it too confusing to use.

I think in general the options are very conflicting in requirements.  If only from a docs perspective it makes sense to leave them separated to me.

tl;dr

My biggest concern is that the documentation will become too confusing.  I can see a situation where users of the module would have to just play with it to figure out how it works and what options they can use together.  But I'll see what I can do.

sivel:
Ok, I've updated the title and initial comment to reflect the change in direction for this pull request.

I've merged the functionality for PTR records into rax_dns_record.  Please let me know if the documentation is clear enough.

jimi-c:
Merged, thanks!

PR #7516, opened 2014-05-23
Relax suds module requirement

evgkrsk:
Please relax this strict requirement for suds module (like for bigsuds).


jimi-c:
This has been merged in, even though github is not reflecting it correctly. Thanks!

PR #7517, opened 2014-05-23
Don't double (or triple) up common error messages

### LINKED ISSUE #7498: Too many calls to process_common_errors in utils/__init__.py
While investigating an unrelated problem, I left an unquoted {{ expression }} as a hash value in one of my YAML files, and was greeted with the following error message (quoted in its entirety):

```
ERROR: Syntax Error while loading YAML script, /home/ams/blah.yml
Note: The error may actually appear before this position: line 14, column 12

  authorized_key:
    user: {{ item.name }}
           ^
This one looks easy to fix.  YAML thought it was looking for the start of a
hash/dictionary and was confused to see a second "{".  Most likely this was
meant to be an ansible template evaluation instead, so we have to give the
parser a small hint that we wanted a string instead. The solution here is to
just quote the entire value.

For instance, if the original line was:

    app_path: {{ base_path }}/foo

It should be written as:

    app_path: "{{ base_path }}/foo"

We could be wrong, but this one looks like it might be an issue with
missing quotes.  Always quote template expression brackets when they
start a value. For instance:         

    with_items:
      - {{ foo }}

Should be written as:

    with_items:
      - "{{ foo }}"   


This one looks easy to fix.  YAML thought it was looking for the start of a
hash/dictionary and was confused to see a second "{".  Most likely this was
meant to be an ansible template evaluation instead, so we have to give the
parser a small hint that we wanted a string instead. The solution here is to
just quote the entire value.

For instance, if the original line was:

    app_path: {{ base_path }}/foo

It should be written as:

    app_path: "{{ base_path }}/foo"
```

Note that it says X, adds Y, and repeats X. This is thanks to the following code in process_yaml_error (from lib/ansible/utils/**init**.py:504 onwards as of commit 104141, but edited here to remove multiline strings for clarity):

```
        msg = process_common_errors(msg, probline, mark.column)
        if not unquoted_var:
            msg = process_common_errors(msg, probline, mark.column)
        else:
            msg = msg + "We could be wrong, but ..."
            msg = process_common_errors(msg, probline, mark.column)
```

Each call to process_common_errors appends to msg, which results in the monstrosity quoted above, since we're in the branch where unquoted_var is true. But note that the doubling happens in the other branch too.

To me, it seems that two of the calls should be eliminated, leaving only:

```
if not unquoted_var:
    msg = process_common_errors(msg, probline, mark.column)
else:
    msg = "We could be wrong but ..."
```

Or perhaps it should instead be:

```
msg = process_common_errors(msg, probline, mark.column)
if unquoted_var:
    msg = msg + "We could be wrong, but..."
```

I've made the former change (since I see no reason to even say X+Y), and I get sensible error messages. If you like, I can submit a pull request.

amenonsen:
process_common_errors() was called thrice, each time appending to the
existing error message, and leading to confusing repetition in the
message that was finally displayed.

Fixes #7498

### DIFF
--- a/lib/ansible/utils/__init__.py
+++ b/lib/ansible/utils/__init__.py
@@ -501,7 +501,6 @@ Note: The error may actually appear before this position: line %s, column %s
                 if '"{{' not in probline or "'{{" not in probline:
                     unquoted_var = True
 
-            msg = process_common_errors(msg, probline, mark.column)
             if not unquoted_var:
                 msg = process_common_errors(msg, probline, mark.column)
             else:
@@ -519,7 +518,6 @@ Should be written as:
       - "{{ foo }}"      
 
 """
-                msg = process_common_errors(msg, probline, mark.column)
         else:
             # most likely displaying a file with sensitive content,
             # so don't show any of the actual lines of yaml just the


jimi-c:
Merged, thanks!

PR #7520, opened 2014-05-23
apache2_module module fixes

jjneely:
Handle the output of a2enmod when it deals with apache module dependencies.  Report the correct changed status.  Also whitespace cleanup.


mpdehaan:
Merged, thank you!

PR #6606, opened 2014-03-20
Support multiple AWS accounts in the ec2.py inventory

jjneely:
This allows one to list IAM credentials in the ec2.ini file under a
section for each AWS account.  The ec2.py will then present a combined
view of the inventory to Ansible, rather than having to run Ansible
seperately for each AWS account.

PR #7523, opened 2014-05-23
Support multiple AWS accounts in the ec2.py inventory

jjneely:
This allows one to list IAM credentials in the ec2.ini file under a
section for each AWS account.  The ec2.py will then present a combined
view of the inventory to Ansible, rather than having to run Ansible
seperately for each AWS account.

This is a re-submission of pull request #6606

### DIFF
--- a/plugins/inventory/ec2.ini
+++ b/plugins/inventory/ec2.ini
@@ -53,3 +53,31 @@ cache_path = ~/.ansible/tmp
 # seconds, a new API call will be made, and the cache file will be updated.
 # To disable the cache, set this value to 0
 cache_max_age = 300
+
+# The following sections describe specific AWS account profiles and can
+# be used to support multiple AWS accounts.  If present these profiles
+# will be used rather than boto searching for your AWS creds.  Create an IAM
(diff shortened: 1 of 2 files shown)


mpdehaan:
Thanks!

Note for review:  We would want to verify whether this change would affect tower and whether any keys starting with [aws] from existing upgrades would still be read and same env vars are still read - which is basically the same thing here.

amenonsen:
@jjneely any chance you could rebase this (it needs to target `contrib/inventory/ec2.py` now) to apply to current devel and test that it still works? Sorry it's taken so long to comment on this.

amenonsen:
I guess this should be closed in favour of #9130 now?

amenonsen:
The above comment is wrong: this is different from #9130—where that PR allowed you to specify the AWS profile to use, this one allows you to use _multiple_ profiles at the same time. That is a new feature. However, there has been no activity on this PR (either from the original author, or from anyone else who wants the feature), and the code is by now quite out of date after the profile-related changes were merged.

So I'll close this PR with apologies now. @jjneely if you feel like rebasing and resubmitting, we'll try to at least review it more quickly this time.

PR #7524, opened 2014-05-23
Replace CommandException with Exception

aresch:


srgvg:
Why?

aresch:
CommandException is not a valid Exception type.  I suspect it was removed at some point from ansible.module.basic and this module was not changed to reflect that.

jimi-c:
Merged, thanks!

PR #7529, opened 2014-05-24
A plugin for task timings

jlafon:
I've had this plugin in [another repo](https://github.com/jlafon/ansible-profile) for a while, but I thought you might want to include it in Ansible. 

Installing this plugin will cause task timings to be printed at the end of the playbook, as in this example:

``` bash
ansible <args here>
   <normal output here>
   really slow task  | Download project packages-----------------------------11.61s
   security | Really slow security policies-----------------------------------7.03s
   common-base | Install core system dependencies-----------------------------3.62s
   common | Install pip-------------------------------------------------------3.60s
   common | Install boto------------------------------------------------------3.57s
   nginx | Install nginx------------------------------------------------------3.41s
   serf | Install system dependencies-----------------------------------------3.38s
   duo_security | Install Duo Unix SSH Integration----------------------------3.37s
   loggly | Install TLS version-----------------------------------------------3.36s
```

### DIFF
--- a/plugins/callbacks/profile_tasks.py
+++ b/plugins/callbacks/profile_tasks.py
@@ -1,0 +1,57 @@
+# (C) 2014, Jharrod LaFon <jlafon@eyesopen.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


mpdehaan:
This is not something I really want to maintain in core because it could output a LOT of output in one big giant chunk, but rather something we should include at the bottom of each task I think.

This may require some modifications but if we could get to this:

Task:  foo ***************

started:  <time>

host result
host result
host result

stopped:   <time>
elapsed:   15 seconds

That seems a bit more useful, it's not an optimization report or anything, but I suspect most people wouldn't need that and might prefer the other.   Something to discuss on ansible-devel perhaps?

(I'd also like it configured off by default to keep the output concise)

Thanks!!!

PR #7530, opened 2014-05-24
Ec2_lc instance_monitoring option

imothee:
We are migrating to Ansible for all our AWS tool management. One of the things we need for creating Launch Configurations is the ability to enable detailed monitoring.

I have checked similar functions and pull requests and hope that this is acceptable.

I would also be happy if a core maintainer were to add the feature in a better way.

Thanks


imothee:
This is also addressed in #7588

imothee:
Closed as addressed in #7896

PR #7531, opened 2014-05-24
Make one fs function

bcoca:
sending this again, to simplify/dedupe the code, keeping the old functions for backwards compatibility with non core modules.

### DIFF
--- a/library/files/assemble
+++ b/library/files/assemble
@@ -192,7 +192,7 @@ def main():
     os.remove(path)
 
     file_args = module.load_file_common_arguments(module.params)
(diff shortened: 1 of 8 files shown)


mpdehaan:
I'm going ahead and merging this as the typo clearly causes problems, though ordinarily I'd like to see the typo separate so we didn't have to force the other in with it.  Still, makes sense to go ahead.

Thanks!

bcoca:
now i see that typo commit x2 in devel ... sorry, had not noticed it sneaking into my PR, I blame the github gremlin!

Issue #7495, opened 2014-05-22
Union of lists of dicts vars doesn't work


gitaarik:
##### Issue Type:

Bug Report
##### Ansible Version:

1.5.5
##### Environment:

OS/X Mavericks
##### Summary:

Using the jinja2 `union` filter to combine two lists doesn't work if the two lists contain dictionaries. 
##### Steps To Reproduce:

You can use the jinja2 `union` filter to combine two lists like this:

```
list1:
  - item1
  - item2

list2:
  - item3
  - item4

list3: "{{ list1 | union(list2) }}"
```

http://docs.ansible.com/playbooks_variables.html#set-theory-filters

However, when you try this with lists of dicts, it doesn't work, the `list3` variable seems to become a list of string objects and not dict objects:

```
list1:
  - name: item1
    value: test1
  - name: item2
    value: test2

list2:
  - name: item3
    value: test3
  - name: item4
    value: test4

list3: "{{ list1 | union(list2) }}"
```
##### Expected Results:

A list of dictionaries that have the items of the two other lists.
##### Actual Results:

A list of strings. Not sure what's in these strings to be honest, Ansible just gave the error: `One or more undefined variables: 'str object' has no attribute 'value'`.

mpdehaan:
Seems like you should assign this to a variable.

If you use it straight in a file like you seem to be doing, I'm not sure because of the weird "=" syntax, it's going to wind up a string in that file.

You probably need to show more context about how you are using this variable, but Jinja2 union is Jinja2, not ansible.

Perhaps stop by ansible-project to ask the "how do I do this" question, vs filing the bug?

Thanks!

bcoca:
union is a ansible jinja2 filter, it was never meant to work with
dictionaries, it works with lists/sets.

gitaarik:
@mpdehaan Sorry, the `=` was a mistake in the example, corrected it.
@bcoca Ok, union is meant for lists only, is there a way to union dictionaries? If no, are there plans to add this feature? If no, could I maybe make an attempt to add it and make a pull request?

bcoca:
doing set theory on sets is easy, doing it on stuff that is not sets ....

you actually want a dictionary merge, which is not something python/jinja
have built in but should not be hard to code.

## 

Brian Coca
Stultorum infinitus est numerus
0110000101110010011001010110111000100111011101000010000001111001011011110111010100100000011100110110110101100001011100100111010000100001
Pedo mellon a minno

[...]

PR #7534, opened 2014-05-25
set theory v2

bcoca:
Now it can handle non hashable items like dicts (which used to produce a TypeError)
Also it now makes each list unique before operating on them

### DIFF
--- a/lib/ansible/runner/filter_plugins/core.py
+++ b/lib/ansible/runner/filter_plugins/core.py
@@ -140,19 +140,40 @@ def regex_replace(value='', pattern='', replacement='', ignorecase=False):
     return _re.sub(replacement, value)
 
 def unique(a):
-    return set(a)
+    try:
+        c = set(a)
+    except TypeError, e:
+        c = []
+        c = filter(lambda x: x not in c, a)
+    return c
 
 def intersect(a, b):
-    return set(a).intersection(b)
+    try:
+        c = set(a) & set(b)
+    except TypeError, e:
+        c = filter(lambda x: x in b, a)
+    return c
 
 def difference(a, b):
-    return set(a).difference(b)
+    try:
+        c = set(a) - set(b)
+    except TypeError, e:
+        c = filter(lambda x: x not in b, a)
+    return c
 
 def symmetric_difference(a, b):
-    return set(a).symmetric_difference(b)
+    try:
+        c = set(a) ^ set(b)
+    except TypeError, e:
+        c = filter(lambda x: x not in intersect(a,b), union(a,b))
+    return c
 
 def union(a, b):
-    return set(a).union(b)
+    try:
+        c = set(a) | set(b)
+    except TypeError, e:
+        c = a + b
+    return c
 
 def version_compare(value, version, operator='eq', strict=False):
     ''' Perform a version comparison on a value '''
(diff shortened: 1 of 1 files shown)


bcoca:
this should solve issues presented in #7495, without a need for a workaround

mpdehaan:
Would it be possible to check the types of the arguments instead of the try/except?

bcoca:
I can check for dicts, but other types can be passed that are not
'hashable', the try/except covers all possible ones.

But considering that most of this should be comming from yaml, it should be
limited to dicts.

## 

Brian Coca
Pedo mellon a minno

bcoca:
actually, after thinking for 10s:

if  isinstance(a, collections.Hashable) : ...
​

lib/ansible/runner/filter_plugins/core.py
```diff
@@ -140,19 +141,40 @@ def regex_replace(value='', pattern='', replacement='', ignorecase=False):
...
+    else:
+        c = filter(lambda x: x not in intersect(a,b), union(a,b))
+    return c
 
 def union(a, b):
-    return set(a).union(b)
+    if isinstance(a,collections.Hashable) and isinstance(b,collections.Hashable):
+        c = set(a) | set(b)
+    else:
+        c = a + b
```
kilburn: Notice that, conceptually, the `set(a) | set(b)` operation is different than `a + b`. The former performs set union meaning that duplicates are removed, whereas the latter is a simple concatenation with the possibility for the same item to appear twice.

If these are considered "set operations", then the else part should be replaced by something like:

```
c = a + filter(lambda x: x not in a, b)
```

bcoca:
thanks, I had changed the rest, I'm not sure why I missed this one

kilburn:
Actually @mpdehaan pointed out that, by turning inputs into sets, we are removing duplicates in them too. Hence, the fallback operations are all wrong here.

Example with hashable items:

```
a,b = [x,x,y,y], [x,z,z]
unique(a) # set([x,y])
intersect(a, b) # set([x])
difference(a, b) # set([y])
symmetric_difference(a, b) # set([y,z])
union(a, b) # set([x,y,z])
```

Example with non-hashable items:

```
a,b = [x,x,y,y], [x,z,z]
unique(a) # set([x,x,y,y]) - Wrong: the not in c check is useless because c is populated only
                             after finishing the whole filter operation
intersect(a, b) # set([x,x]) - Wrong: duplicates in a are not removed
difference(a, b) # set([y,y]) - Wrong: duplicates in a are not removed
symmetric_difference(a, b) # set([y,y,z,z]) - Wrong: duplicates not removed from neither a nor b
union(a, b) # set([x,y,z]) # set([x,x,y,y,x,z,z]) - Wrong: duplicates not removed from neither a nor b
```

Also, the _exact_ same result (than the set operations) is impossible to achieve because sets are implicitly unordered (meaning that you can get varying orders when iterating it) whereas lists are ordered by definition. Anyway, I guess this is an acceptable trade-off...

bcoca:
I'll resubmit tonight, also making changes to try to coerce always
returning a list/set when used in templated fields (with_items:).​

s0x:
This even fixes an issue when using filters in with_items (e.g. https://github.com/ansible/ansible/issues/8055)

srgvg:
This patch fixes #8141 for me. +1

bcoca:
+1 as we keep seeing people bitten by the use of non hashable objects with the set theory filters

jimi-c:
Merged, thanks!

PR #7536, opened 2014-05-25
warn if remote SSL cert is not checked in linode

bsdlp:
`linode-python` optionally requires `pycurl` to do [SSL certificate verification](https://github.com/tjfontaine/linode-python/blob/master/linode/api.py#L81). Without `pycurl`, `linode-python` will use `urllib` and skips checking the SSL cert. It's probably not a good idea (?) to suppress this warning.

### DIFF
--- a/library/cloud/linode
+++ b/library/cloud/linode
@@ -88,7 +88,7 @@ options:
     description:
      - how long before wait gives up, in seconds
     default: 300
-requirements: [ "linode-python" ]
+requirements: [ "linode-python", "pycurl" ]
 author: Vincent Viallet
 notes:
   - LINODE_API_KEY env variable can be used instead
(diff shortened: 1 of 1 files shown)


mpdehaan:
So the warning in mainline output will break the JSON output of Ansible, which is not what you want.

It seems it would be better if removing this line to also make sure pycurl is explicitly included as a dependency with a "try/except ImportError" block that calls fail_json if it's not present.

bsdlp:
thanks @mpdehaan, I addressed your suggestion in 61e6da6.

jimi-c:
Your fix there will not work, two things need to be corrected:
1) You're missing "import" before pycurl
2) The module item is not available at that point in the code. You need to fail in the same way as the import of linode below it:

<pre>
     print("failed=True msg='linode-python required for this module'")
      sys.exit(1)
</pre>

bsdlp:
ugh, sorry guys.

fixed in be7ff82e2de56621fcd3809e61ee26f3f71acbbe

thanks @jimi-c

jimi-c:
Merged, thanks!

Issue #7515, opened 2014-05-23
file: state=link should provide ability to create symlinks relative target dir


edrozenberg:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.7
git 2014-05-23
##### Environment:

Running on Mac OSX 10.9.3 managing Slackware64 14.1
##### Summary:

Currently it appears there is no way to create a symlink relative to the directory of the target.

For ex, would like to be able to achieve the following:

```
cd /usr/src    &&    ln  -s  linux-3.10.40  linux
```

but Ansible only provides the ability to create fully specified links (i.e. ln -s /usr/src/linux-3.10.40  /usr/src/linux

The functionality is important to have because relative symlinks continue to function when an entire directory containing them is moved, while fully specified symlinks would break in such a scenario.
##### Steps To Reproduce:

Create task:

```
- file: src=linux-3.14.4 dest=/usr/src/linux owner=foo group=foo state=link
```

The symlink creation will fail, it is targeting the wrong 'src'
##### Expected Results:

Would expect a new link /usr/src/linux with target "linux-3.14.4"
##### Actual Results:

Task error, attempted to create a symlink targeting /root/linux-3.14.4 (wrong)
but the target obviously doesn't exist.

ansibot:
Thanks for filing a ticket! I am the friendly GitHub Ansibot. I see you did not fill out your issue description based on our new issue template. Please copy the contents of this [template](https://raw2.github.com/ansible/ansible/devel/examples/issues/ISSUE_TEMPLATE.md) and **paste it into the description** of your ticket. 

Our system will automatically close tickets that do not have an issue template filled out within 7 days. Please note that due to large interest in Ansible, humans may not comment on your ticket if you ask them questions. Don't worry, you're still in the queue and the robots are looking after you.

DSpeichert:
I believe it is a regression, this feature worked a few days ago and just stopped working now.
In fact, it works in release1.6.1 but not in release1.6.2

edrozenberg:
It must be a regression, I remember using the same syntax earlier and it used to work fine:

```
file: src=linux-3.14.4 dest=/usr/src/linux owner=foo group=foo state=link
```

bcoca:
this works for me

ansible -m file -a "state=link src=../test.yml
path=/home/user/tmp/test2.yml force=yes" -i 'localhost,' all -c local
​
with existing /home/user/test.yml

edrozenberg:
[...]

PR #7537, opened 2014-05-25
now correctly examines path for relative links

bcoca:
fixes regression recently introduced w/o breaking the fix it applied

### DIFF
--- a/library/files/file
+++ b/library/files/file
@@ -189,10 +189,11 @@ def main():
 
     elif state in ['link','hard']:
 
-        absrc = src
-        if not os.path.isabs(absrc):
-            absrc = os.path.abspath(src)
-
+        if os.path.is_dir(path): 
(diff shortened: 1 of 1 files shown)


bcoca:
should fix #7515

jimi-c:
This does not pass the file integration test:

<pre>
TASK: [test_file | create soft link to file] ********************************** 
failed: [testhost] => {"failed": true, "item": "", "path": "/root/ansible_testing/soft.txt", "src": "/root/ansible_testing/foo.txt", "state": "absent"}
msg: src file does not exist, use "force=yes" if you really want to create the link: /root/ansible_testing/root/ansible_testing/foo.txt
</pre>

Looking into why.

jimi-c:
This was merged in squashed with the two later fixes, so it's not reflected here properly.

PR #7538, opened 2014-05-25
CONTRIBUTING.md cleanup

bsdlp:
No new content was added. This is mainly a housekeeping PR.
- clean up markdown
- remove random extra spaces
- remove trailing spaces
- restrict to 78 chars per line wherever possible
- minor grammar edits
- capitalize Ansible and GitHub for uniformity
- wrap commands in inline code spans
- replace quotes with typographical equivalent, i.e. `'` for `’`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,154 +1,198 @@
-Ansible Community  Information
+Ansible Community Information
 ==============================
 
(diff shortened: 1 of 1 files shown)


mpdehaan:
We do not take commits to markdown files that remove spaces as this breaks attribution on the lines and also makes the request harder to review.

Markdown is also usually rendered in printable form, so wordwrap changes are not neccessary.

Other changes are probably ok,

PR #7542, opened 2014-05-26
rax_cbs module: Don't fail when the volume is not found

sivel:
If the volume is not found, make sure we don't fail the module.  It is expected that the volume may not exist, which is the reason for creating it.

See https://groups.google.com/forum/#!topic/ansible-project/_Ad_Sh8iqsU


jimi-c:
Merged, thanks!

PR #7543, opened 2014-05-26
Fix portability issue when checking local facts file permission.

yangliping:
Under Solaris, os.access(fn, os.X_OK) return True for user root even if the file is not executable by user root. In that way, ansible always try to execute non-executable local facts files instead of reading it.

I suggest use os.stat() to check the executable bit for portability reason.

Please refer the following link for more info.

http://bugs.python.org/issue14706


mpdehaan:
merged, thanks!

PR #7544, opened 2014-05-26
Differentiate machinfo output key/val separator

rik2803:
get_hw_facts for HP-UX: machinfo in HP-UX B.11.23 uses '=' as separator between key and value, while B.11.31 uses ':'.


mpdehaan:
Could this be made more generic so it splits if "=" is found otherwise ":"?  It seems like this might not be the only affected version and the code is a bit specific.

rik2803:
It is the only case, since only 11.23 and 11.31 run on ia64, and the machinfo command only exists in the ia64 version of HP-UX.

jimi-c:
There are other direct checks for the "B.11.23" version in the HPUX class already, so I'm fine with this. Merged, thanks!

PR #7545, opened 2014-05-26
Allow multiple and second-level 'filter' statement in module 'setup'.

jtru:
This small patch enhances the "setup" module's "filter" capabilities: It can now filter second-level elements in the ansible_facts dict (by using "-&gt;" like so: 'filter=ansible_eth0->macaddress') and also take into account multiple filter statements separated by ";" at once ('filter=ansible_eth0;ansible_user_id').

### DIFF
--- a/library/system/setup
+++ b/library/system/setup
@@ -28,6 +28,10 @@ options:
         version_added: "1.1"
         description:
             - if supplied, only return facts that match this shell-style (fnmatch) wildcard.
+              Separate multiple wildcards (which will be ORed) using ";".
+              To match a second-level filter, use the "first_level->second_level" form.
+              Note that second-level filter results overwrite those of first-level filters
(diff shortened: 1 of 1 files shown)


mpdehaan:
I'm really not a fan of the syntax here.

I would accept something Pythonic, perhaps, if there weren't too much code behind it.

Though in this case this really only would be of use to /usr/bin/ansible, in which case, I think we should err on the side of keeping things minimal here, which is something we always strive to do.

Hope this makes sense.

Thanks!

jtru:
To clarify: By "syntax", you mean the syntax of extending the 'filter' argument the way I came up with, or my take at implementing the desired functionality?

That said, do I understand correctly that you wouldn't merge this feature even if it were to meet your expectations referred to above?

Thanks for clearing that up/letting me know.

PR #7549, opened 2014-05-26
add module for rake command

nicolai86:
much like the django_manage rake is a utility for a web framework and
used often when deploying rails applications.

this module adds the rake command and enables the exeution of arbitrary rake commands, even for non
rails projects.

I'd like to get some feedback about adjustments necessary to get this upstream. 
Cheers!

### DIFF
--- a/library/web_infrastructure/rake
+++ b/library/web_infrastructure/rake
@@ -176,4 +176,6 @@ def main():
 
     module.exit_json(**result)
 
+from ansible.module_utils.basic import *
+
 main()
--- a/library/web_infrastructure/rake
+++ b/library/web_infrastructure/rake
@@ -1,0 +1,179 @@
+#!/usr/bin/python
+#coding: utf-8 -*-
+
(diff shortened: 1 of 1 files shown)


mpdehaan:
As with comments in https://github.com/ansible/ansible/pull/7550 I'm a little worried that this is just trivally wrapping a command tool and the environment logic needs to be removed.

nicolai86:
This command actually is more than just a wrapper as it incooperates an often found problem when working with rake commands: executing them only if something changed.

Rails startup time is slow and you'd only want to execute asset precompilation or database migrations when files changed between deployments. Of course it's a problem of rails but it needs to be solved for proper deployment scenarios, and waiting till rails fixes itself is no option.

That being said the point about wrapping a established command from the ruby world & adding the convenience & in general ansible-level user friendliness on top is valuable. 
Hope this makes sense to you :)

nicolai86:
this one also goes away as soon as you introduce a module to check files/ directories for differences.
closed as well, thanks @mpdehaan for your concerns :)

PR #7550, opened 2014-05-26
add bundle module

nicolai86:
Since bundler is a utility used very often for rails deployments the addition of this module would make it even easier to use ansible instead of other tools like capistrano, chef, etc.

I'd like to get some feedback to know what needs to be changed for this PR to be accepted as a new module.

### DIFF
--- a/library/packaging/bundle
+++ b/library/packaging/bundle
@@ -1,0 +1,165 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2013-2014, Raphael Randschau <nicolai86@me.com>
+# Written by Matthew Williams <matthew@flowroute.com>
+# Based on yum module written by Seth Vidal <skvidal at fedoraproject.org>
+#
+# This module is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This software is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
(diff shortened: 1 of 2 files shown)


mpdehaan:
Seems like this should be renamed "bundler" ?

I'd also remove the comments about where this module originated (i.e. the re: Seth and Matt parts) as this is pretty removed from it.

Several parameters like gemfile should be like "gem_file" per Ansible conventions on not running words together.

Do you think people will know what "binstubs" are ?

In general I'm on the fence about this.

It's mostly trivially wrapping a command that could be called with the shell command pretty cleanly.

I'm also not a fan of the environment handling, generally Ansible errs on not trusting the user shell (as it should) and then using the "environment" keyword to set parameters down explicitly.
In this case, GEM_HOME could be set this way or have a reasonable default, though it seems it might also want to be a CLI option.

nicolai86:
While it is true that this is "just" a wrapper around bundler the same could be said about most modules, since they also just wrap regular binaries. 

For me the value of this module comes from two things: 
first, it properly interprets stdout from bundler, marking the step as changed as necessary. And secondly it's mostly convenience as I do not have to care about the details when throwing it at my rails deployments. 

From my point of view the convenience is key here, as some flags (e.g. binstubs) are set which rails-newcomers might not know, but which ease the working if you have to manually ssh to your server and fire up the application at some point.

That being said I'd be perfectly fine to add more value here e.g. by adding support for the common ruby version managers to play along nicely. E.g. to auto-detect rvm, rbenv or packaged rubies and to just do the right thing.

Especially sind that is the major pain point when working with ruby based applications.

Maybe @kiranos want's to chime in as well.

nicolai86:
then again the options covered here a quiet rails deployment specific and might not make the best sense in every scenario. 

I'll check to see if I can obsolete this wrapper as it's mostly a dependeny for the rails-deployment role.

nicolai86:
looks like this can easily be dropped using a combination of `command` alongside with `environment`, at least for packaged & rbenv based scenarios. 

So thanks for your valuable feedback, @mpdehaan :+1:

PR #7552, opened 2014-05-26
Option for keeping the environment untouched when calling the service module on Linux

Ernest0x:
On Linux, if the 'service' binary is installed, it is used by default
to control services by stripping all environmental variables except
TERM, PATH and LANG before calling the init scripts.
This patch provides an option that allows to keep the environment
untouched by controlling the service directly through its init.d
script.

### DIFF
--- a/library/system/service
+++ b/library/system/service
@@ -67,6 +67,16 @@ options:
         default: 'default'
         description:
         - "For OpenRC init scripts (ex: Gentoo) only.  The runlevel that this service belongs to."
+    keep_environment:
+        required: false
+        choices: ["yes", "no"]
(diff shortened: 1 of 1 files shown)


mpdehaan:
Ansible is designed to not trust the user's environment and instead have it dictated via the 'environment' keyword.

That being said from the patch above doesn't really seem to change the way the environment is handled, which I find a little confusing -- but rather runs the init script instead of /sbin/service.

Can't see what that makes sense on a system where the service binary is present.

Sounds like this is a case of a bad init scripts really -- please stop by ansible-devel if you'd like to discuss further.  Thanks!

PR #7555, opened 2014-05-26
SoftLayer Inventory script

icebourg:
Adds a script based on the SoftLayer Python API bindings to get an inventory of servers from SoftLayer.

Returns groups based on:
- Hardware vs virtual
- Memory size
- Number of cores
- FQDN
- Hostname
- Domain
- SoftLayer Datacenter

This is my first contribution to Ansible, so please let me know if you have any questions or suggestions for improvements.


jimi-c:
Merged, thanks!

PR #7556, opened 2014-05-27
Remove duplicate table of contents in Conditionals

jturner314:
Duplicate tables of contents were defined on line 4 and line 14. This removes the one defined on line 14.


jimi-c:
Merged, thanks!

PR #6522, opened 2014-03-17
Implemented pkg for Solaris 11 package manager

dermute:
Hello,
Solaris 11 included a new package manager "pkg". Its Solaris first management with a central repository (called publisher). Even with backward compatibility to svr4 packages (with the module "svr4pkg"), you should use the new feature.

I implemented the 3 standard-features 'present', 'latest' and 'absent' to work with the setup repositories.

PR #7558, opened 2014-05-27
added solaris fact detection for ansible_pkg_mgr

dermute:
The fact for ansible_pkg_mgr is not detected for Solaris and it is not possible to use ansible_pkg_mgr in any playbook (overwriting with group_vars is not possible).

Thats why I added the detection of the Solaris package manager. It is svr4pkg for solaris 10 (and older) and pkg for Solaris 11. The pkg module is still pending in pull request #6522


jimi-c:
Merged, thanks!

dermute:
Merging this pull request completly is not correct without looking at #6522. Until now, there is no module called 'pkg'.

PR #7562, opened 2014-05-27
Additional modules for interacting with RabbitMQ

jaconza:
Modules to allow users to add and remove queues, bindings and
exchanges using the RabbitMQ management plugin's http based API.

Since the default rabbitmq control script does not allow for adding queues, exchanges or bindings we opted to make use of the management plugin's http based api.

### DIFF
--- a/library/messaging/rabbitmq_binding
+++ b/library/messaging/rabbitmq_binding
@@ -195,7 +195,7 @@ class RabbitMQExchange(object):
         list_bindings = json.loads(content)
         for binding in list_bindings:
             # does the routing keys match?
-            routing_key_match = ('routing_key' in self.binding) and (self.routing_key == binding['routing_key']) if self.routing_key is not None else True
+            routing_key_match = ('routing_key' in binding) and (self.routing_key == binding['routing_key']) if self.routing_key is not None else True
 
             # same amount of arguments?
             args_length_match = (len(self.arguments) == len(binding['arguments']))
--- a/library/messaging/rabbitmq_binding
+++ b/library/messaging/rabbitmq_binding
@@ -100,7 +100,7 @@ options:
 
 EXAMPLES = '''
 # Bind an exchange to a queue.
-- rabbitmq_binding: host='http://localhost:15672/api'
+- rabbitmq_binding: host='localhost'
                     user='guest'
                     password='guest'
                     vhost='/'
@@ -110,7 +110,7 @@ EXAMPLES = '''
                     routing_key='test'
 
 # Bind an exchange to another exchange.
-- rabbitmq_binding: host='http://localhost:15672/api'
+- rabbitmq_binding: host='localhost'
                     user='guest'
                     password='guest'
                     vhost='/'
@@ -120,7 +120,7 @@ EXAMPLES = '''
                     routing_key='test'
 
 # To delete a binding between an exchange and a queue
-- rabbitmq_binding: host='http://localhost:15672/api'
+- rabbitmq_binding: host='localhost'
                     user='guest'
                     password='guest'
                     vhost='/'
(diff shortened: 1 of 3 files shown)


jimbydamonk:
Very nice. I was about to start working on this.

jimbydamonk:
I can't seem to get your examples to work when using arguments.
I copied your example, and get the following:

---

 ERROR: Syntax Error while loading YAML script, test.yml
Note: The error may actually appear before this position: line 10, column 55

```
              queue=test_queue
              arguments='{"x-dead-letter-exchange": "dead"}'
                                                  ^
```

This one looks easy to fix.  It seems that there is a value started
with a quote, and the YAML parser is expecting to see the line ended
with the same kind of quote.  For instance:

```
when: "ok" in result.stdout
```

Could be written as:

   when: '"ok" in result.stdout'

or equivalently:

   when: "'ok' in result.stdout"

jaconza:
@jimbydamonk It seem like yaml does not like the space in the arguments string. I have update the example to remove this space. 

The following yields an error:

```
arguments='{"x-dead-letter-exchange": "dead"}'
```

But removing the space solves the error:

```
arguments='{"x-dead-letter-exchange":"dead"}'
```

While doing this I also realised that the host argument was wrong in most of the example. This to have been update. 

Here is the example playbook I used for testing:

``` yaml
- hosts: local
  user: root 
  tasks:
    - rabbitmq_queue: host='localhost'
                      user=guest
                      password=guest
                      vhost=/
                      queue=test_queue
                      arguments='{"x-dead-letter-exchange":"dead"}'

    - rabbitmq_exchange: host='localhost'
                         user=guest
                         password=guest
                         vhost=/
                         name=test_exchange
                         type=direct
                         durable=no
                         internal=no
                         auto_delete=no
                         arguments='{"some-additional-arg":true}'

    - rabbitmq_binding: host='localhost'
                        user='guest'
                        password='guest'
                        vhost='/'
                        source='test_exchange'
                        destination='test_queue'
                        binding_type='queue'
                        routing_key='test_routing_key'
```

jimbydamonk:
@jaconel That worked. Removing the space works. 

Now on to the next error. If I store the args in a variable it seems to be hitting a json parse error. I assume this is something to do with the way jinja is messing with the var. 
Here is my play book:

```
---
- hosts: localhost
  vars:
    product_queues: 
      - { name: 'ha.reject', arguments: '{"x-dead-letter-exchange":"dead"}' }
  tasks:
    - name: Debug
      debug: msg=" args = {{ item.arguments }}"
      with_items: product_queues 

    - name: Test rabbit no var 
      rabbitmq_queue: user=guest password=guest vhost=/ queue=test_queue 
        arguments='{"x-dead-letter-exchange":"dead"}'    
      ignore_errors: yes

    - name: Test rabbit single var  
      rabbitmq_queue: user=guest password=guest vhost=/ queue=test_queue 
        arguments=" {{ product_queues[0].arguments }}"   
      ignore_errors: yes

    - name: Test Rabbit in a loop 
      rabbitmq_queue: user=guest password=guest vhost=/ queue=test_queue 
        arguments="{{ item.arguments }}"
      with_items: product_queues
```

I put the out put in a gist https://gist.github.com/jimbydamonk/5700cc06b8db20743bdb

You can see that the tasks with a variable die with:
`ValueError: Expecting property name: line 1 column 3 (char 2)`

jimbydamonk:
Oh, and I hit a bit with the binding:
on line 198 of rabbit_binding you have self.bindings instead of just bindings. 
Removing self. makes it work.

jaconza:
So managed to fix the scope error in rabbitmq_binding. Thanks for that. 

Still looking into the "args in variable" error. It would appear that Jinja is stripping the `"` from the string, thus causing the json parser in the module to fail. Will confirm shortly. Have already tried using the `safe` filter, no love.

jaconza:
Ok, gave it a quick test. It seems to be that Jinja is stripping the quotes from the strings. Test can be seen here: https://gist.github.com/jaconel/1dfddf21f2467d520ee5

jaconza:
Broke away from passing the 'arguments' parameter in as a json string. Instead used a complex argument and altered examples.

mpdehaan:
Hi!

Thanks you very much!

PR #7563, opened 2014-05-27
Recent BSD su work introduced a failing test. Removed leading space from...

tima:
... the assertion that seems to be the culprit.

### DIFF
--- a/test/units/TestUtils.py
+++ b/test/units/TestUtils.py
@@ -512,7 +512,7 @@ class TestUtils(unittest.TestCase):
         cmd = ansible.utils.make_su_cmd('root', '/bin/sh', '/bin/ls')
         self.assertTrue(isinstance(cmd, tuple))
         self.assertEqual(len(cmd), 3)
-        self.assertTrue(' root -c "/bin/sh' in cmd[0])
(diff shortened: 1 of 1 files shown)


jimi-c:
That's odd, I do not see this error when running `make tests` on my devel branch.

jimi-c:
Merged, thanks!

tima:
I thought this had corrected the problem I was having here but I was mistaken. The same unit test is failing. I slipped a sys.stderr.write of cmd[0] in at line 515 of TestUtils. This is what cmd[0] on my RHEL6 box reports:

```
/bin/sh -c 'su  root /bin/sh -c '"'"'echo SUDO-SUCCESS-xjpueejhpdfifwmcdfielkzthmnmwtmq; /bin/ls'"'"'' 
```

Since the failing test is looking for 'root -c "/bin/sh' that fails. Is this some sort of quirk between RHEL vs. BSD? I can fix it to work on my system, but have some concern that I will break others.

jimi-c:
No, the command should be consistent across both platforms, so fixing it in the test should be consistent. Send us a PR whenever you're ready to have us look at it. Thanks!

tima:
I've submitted my PR. See  #7652.

PR #7567, opened 2014-05-27
Fixed documentation for etype parameter on acl module.

kitwestneat:
The etype parameter just had a copy of the default parameter documentation.
Changed it to reflect what it really is for.


jimi-c:
Merged, thanks!

PR #7576, opened 2014-05-27
add elb health/status module.

tedder:
Returns state of a given ELB in AWS. Has pretty docu and examples.

works with test-module, and works on my machine(tm).

Somewhat wonder if I should return it as site_facts. That would remove the 'register' var, right? Let me know.

### DIFF
--- a/library/cloud/elb_health
+++ b/library/cloud/elb_health
@@ -1,0 +1,102 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+
+DOCUMENTATION = '''
+module: elb_health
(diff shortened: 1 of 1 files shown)


mpdehaan:
I would rather not see another module for this, rather, it should be information that can be returned by the elb module.

Would this be possible to add a different state that doesn't create, but just returns info, and also return this info when created?  

Even so, this tends to imply "elb_facts", and I'm still not wanting to see extra module sprawl for this if there's a better place to put this.

tedder:
yeah, I'll put it in ec2_elb with a state. Closing.

PR #7579, opened 2014-05-28
Fix ec2_asg delete (and default availability zones)

bellkev:
The main issue here is that the asg delete function was not polling the instance count correctly, so it would hang indefinitely.

Along the way (while trying to simplify the integration test) I also fixed the default availability zone logic (sorry for not making a separate pull request for this, but it was another very simple change).

This pull request includes fixes for both bugs as well as a new integration test for the ec2_asg module. Note that, for practical reasons, I only ever ran this test in isolation as

```
TEST_FLAGS=" --tags test_ec2_asg" make amazon
```

 but I don't think that should be a problem...

Also note that I used the async param to enforce a 5 min timeout on the delete, which consistently failed for me with the old/buggy code and passed with my fix, but because of the general non-determinism of all this stuff, you may want to up that a bit. (I'm not sure what your workflow for running the longer-running integration tests is...)

### DIFF
--- a/library/cloud/ec2_asg
+++ b/library/cloud/ec2_asg
@@ -131,7 +131,7 @@ def create_autoscaling_group(connection, module):
             ec2_connection = connect_to_aws(boto.ec2, region, **aws_connect_params)
         except boto.exception.NoAuthHandlerFound, e:
             module.fail_json(msg=str(e))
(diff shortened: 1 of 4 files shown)


jimi-c:
Merged, thanks! (I rebased after merging your code in, which messes up github's auto-detection that this was merged in, so closing manually).

PR #7583, opened 2014-05-28
Adds support to ec2_group for purging group rules

jimbydamonk:
Similar to ec2_elb_lb, this adds support to dictate wether or not to purge the group rules. This is very useful for updating security groups later on after creation. Default behavior is not changed.


mpdehaan:
Merged, thank you!

PR #7593, opened 2014-05-30
New module: Windows Azure

jwhitbeck:
Use this module to provision or terminate virtual machines on windows azure.

We have been happily using this in prod at our company for 5 months now so we thought we would share it with the community.

### DIFF
--- a/library/cloud/azure
+++ b/library/cloud/azure
@@ -1,0 +1,399 @@
+#!/usr/bin/python
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: azure
+short_description: create or terminate a virtual machine in azure
+description:
+     - Creates or terminates azure instances. When created optionally waits for it to be 'running'. This module has a dependency on python-azure >= 0.7.1
+version_added: "1.5"
+options:
+  name:
+    description:
+      - name of the virtual machine and associated cloud service.
+    required: true
+    default: null
+  location:
+    description:
+      - the azure location to use (e.g. 'East US')
+    required: true
+    default: null
+  subscription_id:
+    description:
+      - azure subscription id. Overrides the AZURE_SUBSCRIPTION_ID environement variable.
+    required: false
+    default: null
+  management_cert_path:
+    description:
+      - path to an azure management certificate associated with the subscription id. Overrides the AZURE_MANAGEMENT_CERT_PATH environement variable.
+    required: false
+    default: null
+  storage_account:
+    description:
+      - the azure storage account in which to store the data disks.
+    required: true
+  image:
+    description:
+      - system image for creating the virtual machine (e.g., b39f27a8b8c64d52b05eac6a62ebad85__Ubuntu_DAILY_BUILD-precise-12_04_3-LTS-amd64-server-20131205-en-us-30GB)
+    required: true
+    default: null
+  role_size:
+    description:
+      - azure role size for the new virtual machine (e.g., Small, ExtraLarge, A6)
+    required: false
+    default: Small
+  endpoints:
+    description:
+      - a comma-separated list of TCP ports to expose on the virtual machine (e.g., "22,80")
+    required: false
+    default: 22
+  user:
+    description:
+      - the unix username for the new virtual machine.
+    required: false
+    default: null
+  password:
+    description:
+      - the unix password for the new virtual machine.
+    required: false
+    default: null
+  ssh_cert_path:
+    description:
+      - path to an X509 certificate containing the public ssh key to install in the virtual machine. See http://www.windowsazure.com/en-us/manage/linux/tutorials/intro-to-linux/ for more details.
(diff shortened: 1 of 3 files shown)


mpdehaan:
This probably should just be shorted to "azure" in the module name, as we similarly don't have "amazon_aws" or "rackspace_rax" and this could be used to spin up Linux machines.

mpdehaan:
In queue for review, we have some other parties interested in the same thing so this is quite timely.  Thanks!

mpdehaan:
note to us (or if you'd like to fix that would be great), version_added stuff should be updated to say 1.7 (current development release).

Current shipping release is 1.6.X

matlockx:
I'm working on something similar but have a bit more functionality in it.pull request should follow tomorrow. Maybe you wait and then compare what to take or take this one and I try to merge mine later on.

matlockx:
Just some points from my side,correct me if I'm wrong:
With this pr it's not possible to create multiple vms to a deployment, cause you need to call add_role from the api a far I know. 
Terminate does not clean up os vhds and not ready for multiple roles per deployment. 
Since I already implemented that I would be happy to merge that into this pr after you have accepted it.

jwhitbeck:
@matlockx All your points are correct. We did not implement these as we did not need them. It sounds like you have a more complete azure implementation, so we are looking forward to your pr.

matlockx:
Well, complete is relative. I don't have that neat cache and config stuff so I'd be happy to merge my changes into yours. Maybe it makes sense to do a pull request into your repo first and you merge the parts that make sense and do a new pull request to the main repo. What do you think?

jwhitbeck:
Let's see what the ansible devs prefer. I think it would be easier to first merge our module into the main repo, and then let you merge your improvements on top of that when they are ready.

matlockx:
k, i’m also fine with that

mpdehaan:
We're going to merge this one after some (pending) testing review and take some modifications on top of that.  We can then explore other additions in smaller steps.

Thanks all!

jimi-c:
While trying to create a vm using the documented example (with my own storage account name there), I'm getting the following error:

<pre>
&lt;Error xmlns="http://schemas.microsoft.com/windowsazure" xmlns:i="http://www.w3.org/2001/XMLSchema-instance"&gt;&lt;Code&gt;BadRequest&lt;/Code&gt;&lt;Message&gt;The hosted service name is invalid.&lt;/Message&gt;&lt;/Error&gt;
</pre>

Any ideas what's wrong there?

jimi-c:
Nevermind, turns out azure does not like VM names with an underscore in it. However, now I'm getting a different error... working through it.

One thing I'm definitely noticing, the error messages are not descriptive enough. For example, when a service name already exists, all that is printed is this:

<pre>
failed: [127.0.0.1] => {"failed": true}
msg: Conflict (Conflict)
</pre>

I'll be improving those before merging this in.

jwhitbeck:
Thanks for all the work. If you would like me to chip in, just let me know.

jimi-c:
So the issue I'm running into right now is that after deleting a VM, it does not allow me to recreate it using the same name. I see the service appear, and it says it's deploying, however after that no new VM's start up. So I tried deleting the entire cloud service - which is now recreated, but without certificates apparently. When I try and re-run the playbook I see this:

<pre>
failed: [127.0.0.1] =&gt; {"failed": true}
msg: failed to create the new virtual machine, error was: Unknown error (Bad Request)
&lt;Error xmlns="http://schemas.microsoft.com/windowsazure" xmlns:i="http://www.w3.org/2001/XMLSchema-instance"&gt;&lt;Code&gt;BadRequest&lt;/Code&gt;&lt;Message&gt;The certificate with thumbprint XXXXXXXXXXXXXXXXX was not found.&lt;/Message&gt;&lt;/Error&gt;
</pre>

So I'm not sure if this is something going on with Azure's API or if it's something the code isn't doing correctly. The thumbprint above shows up in my Azure console under my management certificates and matches exactly, so I'm not sure what's going on there yet.

jimi-c:
Merged! You can see my changes in the committed version. One of the areas of improvement after using this module a bit could be that rather than fail when certain errors are caught (ie, the redirect error), that the operation be retried for a short timeout, which might make it a bit more resilient. I noted a lot of these errors during testing, and just had to rerun the playbook.

matlockx:
Hi,

Thumbprint error is because it is an async request to add the certificate
to the cloud service/deployment. We need to wait for it. I'll do a pull
request.
The conflict error comes with no information from the sdk.I'm not sure
whether it's related to the sdk or the azure api directly. Will check that.
Temporary redirects are a big problem as I already told. We can wrap it in
a retry loop but I tried that and it doesn't solve it always. Meanwhile the
guys from the sdk merged my pr to fix the problem in the sdk but I don't
know when they will do a release.
https://github.com/Azure/azure-sdk-for-python/pull/157

So what do you think shall we do:
A)wrap each sdk request into a try catch loop,  do a search in the error
message whether it's a 'temporary redirect' and do the request again then.
But this will not fix all situations. I had really long retries which
exceeded the given timeout....
B) override the method in the sdk when the sdk version is <=0.8.0 to follow
redirects.

Best
Martin

jimi-c:
That's fine if the retry loop exceeds a timeout, we can make that configurable, but it's definitely better than the spurious failures I was seeing (somewhat consistently).

I don't think we need to override the method, we can just require the SDK version to be > 0.8.0 after they do the release to fix the issue.

matlockx:
k, I think something like that should do the trick:

```
   def handle_temporary_redirects(f):
    count = 0
    while True:
        try:
            return f()
        except WindowsAzureError as e:
            if str(e).lower().find("temporary redirect") and count < 10:
                count += 1
                time.sleep(2)
                pass
            else:
                raise e
```

result = handle_temporary_redirects(lambda : azure_sdk_method(...))

Do you really think that it is necessary that the timeout is configurable? I would do it hard coded  for 300 seconds.

jimi-c:
It doesn't have to be.

matlockx:
i created a pr: #7694 which should fix most of the problems that you saw. I will add another entry in the google group to discuss how we proceed with adding new feature.

PR #7595, opened 2014-05-30
corrected doc example, non existing fact to the 2 it had 'merged'

bcoca:
example was wrong, this might make more sense going from specific to general


mpdehaan:
merged, thanks!

PR #7597, opened 2014-05-30
Merge and intersect lists without using sets.

### LINKED ISSUE #7596: Bug when combining vars_files from main playbook and included playbooks
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.7 (devel 188ab7b952) last updated 2014/05/30 06:53:46 (GMT +200)
##### Environment:

OS X Mavericks host - Debian wheezy guests (but this should be irrelevant)
##### Summary:

I have a playbook which uses vars_files to load some variables. Then it includes another playbook which uses "conditional imports" (http://docs.ansible.com/playbooks_variables.html#conditional-imports) to load more variables from external files. Ansible fails when trying to combine both these vars_files definitions with a `TypeError: unhashable type: 'list'`.
##### Steps To Reproduce:

Create empty files `vars/default`, `vars/a`, and `vars/b`. Then define this two playbooks:
**play.yml**:

```

---
- hosts: all
  vars_files: 
    - vars/default
- include: play-included.yml
```

**play-included.yml**:

```

---
- hosts: all
  vars_files: 
    - [ "vars/a", "vars/b" ]
  tasks:
    - fail: msg="This is actually success."
```

Now try to run `playbook.yml` with `ansible-playbook` and you get the error.
##### Expected Results:

Something like the following:

```

ok: [web.dev.local]

ok: [web.dev.local]

TASK: [fail msg="This is actually success."] ********************************** 
failed: [web.dev.local] => {"failed": true}
msg: This is actually success.

           to retry, use: --limit @/Users/marc/play.retry

web.dev.local              : ok=2    changed=0    unreachable=0    failed=1   
```
##### Actual Results:

Ansible fails with the following error:

```
  File "/Users/marc/Documents/Projects/ansible/ansible/bin/ansible-playbook", line 317, in <module>
    sys.exit(main(sys.argv[1:]))
  File "/Users/marc/Documents/Projects/ansible/ansible/bin/ansible-playbook", line 200, in main
    force_handlers=options.force_handlers
  File "/Users/marc/Documents/Projects/ansible/ansible/lib/ansible/playbook/__init__.py", line 176, in __init__
    (self.playbook, self.play_basedirs) = self._load_playbook_from_file(playbook, vars)
  File "/Users/marc/Documents/Projects/ansible/ansible/lib/ansible/playbook/__init__.py", line 261, in _load_playbook_from_file
    p['vars_files'] = utils.list_union(p.get('vars_files', []), play_vars_files)
  File "/Users/marc/Documents/Projects/ansible/ansible/lib/ansible/utils/__init__.py", line 1014, in list_union
    set_a = set(_listify(a))
TypeError: unhashable type: 'list'
```
##### Reporter's comments:

The error is pretty clear in itself. The list_union function (introduced in 85bd6810) is defined as:

```
def list_union(a, b):
    set_a = set(_listify(a))
    set_b = set(_listify(b))
    return list(set_a.union(set_b))
```

... but this is conceptually wrong. You cannot insert lists into sets because they are not hashable. Moreover, sets have no guaranteed order, so using set_union to combine lists may break the variable load ordering as intended by the user.

kilburn:
Using sets for these operations is dangerous because sets cannot contain
certain object types (such as lists) and their iteration order is
undefined.

Fixes #7596

### DIFF
--- a/lib/ansible/utils/__init__.py
+++ b/lib/ansible/utils/__init__.py
@@ -1003,21 +1003,19 @@ def is_list_of_strings(items):
             return False
     return True
 
-def _listify(a):
-    if not isinstance(a, (list, tuple)):
-        return [a,]
-    else:
-        return a
-
 def list_union(a, b):
-    set_a = set(_listify(a))
-    set_b = set(_listify(b))
-    return list(set_a.union(set_b))
+    result = list(a)
+    for i in b:
+        if i not in result:
+            result.append(i)
+    return result
 
 def list_intersection(a, b):
-    set_a = set(_listify(a))
-    set_b = set(_listify(b))
-    return list(set_a.intersection(set_b))
+    result = []
+    for i in a:
+        if i in b:
+            result.append(i)
+    return result
 
 def safe_eval(expr, locals={}, include_exceptions=False):
     '''
--- a/lib/ansible/utils/__init__.py
+++ b/lib/ansible/utils/__init__.py
@@ -1004,17 +1004,20 @@ def is_list_of_strings(items):
     return True
 
 def list_union(a, b):
-    result = list(a)
-    for i in b:
-        if i not in result:
-            result.append(i)
+    result = []
+    for x in a:
+        if x not in result:
+            result.append(x)
+    for x in b:
+        if x not in result:
+            result.append(x)
     return result
 
 def list_intersection(a, b):
     result = []
-    for i in a:
-        if i in b:
-            result.append(i)
+    for x in a:
+        if x in b and x not in result:
+            result.append(x)
     return result
 
 def safe_eval(expr, locals={}, include_exceptions=False):


bcoca:
any advantages over #7534 ?

kilburn:
Maybe you can reuse those new functions, but the affected part of the code is very different. 

Bug #7534 concerns jinja filters, whereas the affected code here is not in any template-related code but in the playbook loading code. Hence that fix doesn't solve the issue presented here.

bcoca:
I was looking at the approach/style, they both are similar issues and I
wanted to know if this one had any aspects that were better than the one I
used.
​

kilburn:
No it doesn't, except for the minor issue about union that I already commented on your commit :)

mpdehaan:
Possible lurking bug in this PR:

```
for i in a:
+        if i in b:
 +            result.append(i)
```

If "a" had duplicates, it's going to have duplicates in the result.

We'd have to make sure the behavior was the same as before.

kilburn:
I have updated the code to avoid duplicates in the output. It is a bit more verbose but I don't see any way to make it more concise...

kilburn:
Also, I can't find any usage of the `list_intersection` function. Maybe we should remove it?

jimi-c:
Merged, thanks!

PR #7601, opened 2014-05-30
Tag Based EC2 Inventory Names

bubba-h57:
This pull request implements a feature to allow an EC2 tag to override the vpc_destination_variable which already overrides the default destination_variable. When using the inventory script, I found it much more meaningful to see a friendly, recognizable host name in the Ansible output than the options available.

### DIFF
--- a/plugins/inventory/ec2.ini
+++ b/plugins/inventory/ec2.ini
@@ -34,6 +34,11 @@ destination_variable = public_dns_name
 # be run from with EC2.
 vpc_destination_variable = ip_address
 
+# if you would prefer to use an EC2 Instance Tag for the destination, set this
+# to the tag name to override any previous choices above; however, it is your
+# responsibility to ensure the tag will resolve to an IP address.
+# tag_destination_variable = Name
+
 # To tag instances on EC2 with the resource records that point to them from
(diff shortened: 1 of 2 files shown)


mpdehaan:
So I find the above language somewhat confusing, but perhaps that's just me.   Perhaps an example would help?

I guess what I'm thinking of here is what would this do if the tag didn't exist and wasn't applied consistently, etc.

bubba-h57:
The try construct attempts to load the a defined tags value. 

It could fail due to the tag not existing or due to the configuration not having the tag defined. 

Either case should fall to Exception that simply loads the vpc_destination_variable or the destination_variable the same as it would previously.

jmehnle:
I've been successfully using this patch on my company's infrastructure. What does it take to get this merged?

bcoca:
it needs rebasing as it does not merge cleanly anymore

jmehnle:
@bubba-h57 Can you rebase and perhaps omit the gratuitous whitespace changes, so this can be merged?

bcoca:
actually, I merged a competing simpler implementation in #9825, I'm closing this ticket in favor of that one

PR #7604, opened 2014-05-30
Document all config options and env vars

sivel:
This pull request adds a section to intro_configuration.rst to add a list of all config sections, config options, corresponding env vars and defaults.

This was a result of discussions about lack of documentation about the available env vars.


mpdehaan:
We don't want to feature the env based configuration - it's legacy only, and while I think this may do as an index, this is not an index, and it also hard codes defaults in the docs that may change.

As such, I'd like to keep the section as is, but add documentation for any missing config options instead.

mpdehaan:
Could link to the example.cfg in the repo for the complete list, that way we don't repeat the settings.

Issue #5059, opened 2013-11-26
Error message from lookup plugin


oxyc:
playbook:

```
- host: development
  vars:
    hostname_v1: 'precise64'
    hostname_v2: '{{ ansible_hostname }}'
```

role task:

```
- debug: var=ansible_hostname
- debug: msg="{{ lookup('password', inventory_dir + '/credentials/' + ansible_hostname + '/root/mariadb_password length=15') }}"
- debug: var=hostname_v1
- debug: msg="{{ lookup('password', inventory_dir + '/credentials/' + hostname_v1 + '/root/mariadb_password length=15') }}"
- debug: var=hostname_v2
- debug: msg="{{ lookup('password', inventory_dir + '/credentials/' + hostname_v2 + '/root/mariadb_password length=15') }}"
```

Output:

```
TASK: [mariadb | debug var=ansible_hostname] ********************************** 
ok: [localhost] => {
    "ansible_hostname": "precise64", 
    "item": ""
}

TASK: [mariadb | debug msg="{{lookup('password', inventory_dir + '/credentials/' + ansible_hostname + '/root/mariadb_password length=15')}}"] *** 
ok: [localhost] => {
    "item": "", 
    "msg": "8_47E5O9AZO29aS"
}

TASK: [mariadb | debug var=hostname_v1] *************************************** 
ok: [localhost] => {
    "hostname_v1": "precise64", 
    "item": ""
}

TASK: [mariadb | debug msg="8_47E5O9AZO29aS"] ********************************* 
ok: [localhost] => {
    "item": "", 
    "msg": "8_47E5O9AZO29aS"
}

TASK: [mariadb | debug var=hostname_v2] *************************************** 
ok: [localhost] => {
    "hostname_v2": "precise64", 
    "item": ""
}
Traceback (most recent call last):
  File "/usr/bin/ansible-playbook", line 269, in <module>
    sys.exit(main(sys.argv[1:]))
  File "/usr/bin/ansible-playbook", line 209, in main
    pb.run()
  File "/usr/lib/python2.7/site-packages/ansible/playbook/__init__.py", line 263, in run
    if not self._run_play(play):
  File "/usr/lib/python2.7/site-packages/ansible/playbook/__init__.py", line 592, in _run_play
    if not self._run_task(play, task, False):
  File "/usr/lib/python2.7/site-packages/ansible/playbook/__init__.py", line 357, in _run_task
[...]

PR #7608, opened 2014-05-30
template.py: Handle purposely raised exceptions in lookup()

jjneely:
If a lookup plugin is run by the lookup() template method it should pass
along any AnsibleError (or child exception classes) rather than just eat
them.  These exceptions are purposely raised by the plugin.

This was broken by: https://github.com/ansible/ansible/issues/5059

### DIFF
--- a/lib/ansible/utils/template.py
+++ b/lib/ansible/utils/template.py
@@ -91,6 +91,9 @@ def lookup(name, *args, **kwargs):
         # safely catch run failures per #5059
         try:
             ran = instance.run(*args, inject=vars, **kwargs)
+        except errors.AnsibleError:
+            # Plugin raised this on purpose
+            raise
         except Exception, e:
             ran = None
         if ran:


jjneely:
Usage case here is if a lookup plugin wants to halt the Ansible run for some, hopefully important, reason.  This worked prior to #5059.

I also feel uncomfortable that we are /dev/null'ing exception that come out of plugins this way.  I didn't see a logger handy or I would have made sure that all exceptions at least get a log message.

mpdehaan:
merged, thanks!

nathanhruby:
This bug has hit me twice in the past two weeks.  Any thoughts about backporting it to 1.6.x or will 1.7.x be out soon enough?

jimi-c:
I have reverted this commit, as it was causing a more serious issue (#8297). Essentially the issue boils down to the fact that when a lookup is done in the `vars/main.yml` of a role, the fact gathering step will fail since the lookup plugin does not have any information about the role's path. For this to work, there must be some way to signal the lookup function that errors should not be fatal during the setup module execution.

nathanhruby:
Hi Jimi,

I'm unsure why errors should be ignored in setup, wouldn't that initialize variables in the play that would then pass a defined test later in execution even though they were not correctly looked up? 

-n

jimi-c:
@nathanhruby no, failure to template a variable leaves the string as-is. When re-run through the template engine later, the flag would be disabled so exceptions would be generated. This is how we handle variables that are defined later (such as inventory variables).

PR #7610, opened 2014-05-30
updating ec2_group example, to be more consistent with parameter names

CorbanR:
updating ec2_group example, to be more consistent with parameter names


mpdehaan:
merged, thanks!  Will appear in next docs build!

PR #7616, opened 2014-05-31
Add role templates path to jinja2 loader paths.

wkf:
Lately, I've found myself wanting to include 'base' templates in a role, and then extend them with more specfic templates outside the roles. This makes the roles more reusable, and can be accomplished now by using a path like `roles/nginx/templates/vhost.j2`. Where it starts to break down is when you have altered `roles_path`, since now the nginx role might not be in the `roles` directory. So, in order to avoid coupling the path to the layout of your directory, I've made a small change to search the role's templates directory when using `extends`.

### DIFF
--- a/lib/ansible/runner/action_plugins/template.py
+++ b/lib/ansible/runner/action_plugins/template.py
@@ -78,6 +78,10 @@ class ActionModule(object):
             else:
                 source = utils.path_dwim(self.runner.basedir, source)
 
+        if '_original_file' in inject:
+            templates_paths = [utils.path_dwim_relative(inject['_original_file'], 'templates', '', self.runner.basedir, check=False)]
+        else:
+            templates_paths = None
 
         if dest.endswith("/"):
             base = os.path.basename(source)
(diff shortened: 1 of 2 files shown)


mpdehaan:
Thanks for this, I don't see a problem with the roles/template path being able to be used easily from template includes.

There are some things here I'd like to see corrected.

A variable like "templates", if it exists, seems to imply an array.

Rather than "template_from_file" taking a role_paths parameter, perhaps it should take a "search_paths" parameter.

Variables should default to None if not set, not empty string.

I'm also a little bit unclear on this part and what this does:

```
role_templates != load_paths[1]:
```

Also, in the future, please do not submit commits that only affect whitespace.   This breaks line attribution and really doesn't have any effects we are seeking in commit history.

Thanks!

wkf:
Thanks for the feedback @mpdehaan... I've dropped the whitespace commit, and made the changes you requested. `role_templates != load_paths[1]:` was an attempt to only add the item to the list if it wasn't there already. I've updated to code to make the intent more clear. Thanks!

amenonsen:
@wkf any chance you could rebase this to apply to current devel?

sivel:
@wkf,

Hi!

amenonsen:
Closing per the comment above.

PR #7617, opened 2014-05-31
Azure module Number 2

matlockx:
Hi,

in addition to https://github.com/ansible/ansible/pull/7593 this is another pull request for an azure module. Basically it still lacks:
- creating a virtual network
- directly restart or stop an instance

It has the following issues:
- It still might cause public port conflicts for clusters but not for ssh
- There is an open issue with the azure sdk (https://github.com/Azure/azure-sdk-for-python/issues/129) which forced me to monkey patch a method to follow a temporary redirect. Otherwise that bites the cluster functionality hard.
- There is also some work to do for the facts which the module returns (and also the inventory plugin). Currently it works (just publishes the structures from the sdk) but I will change that the next days.

In conclusion: this is a starting point and i'm working currently on the above mentioned points. They should follow the next days. 

Open questions from my side:

I'm not quite sure whether i have completely understood the whole azure thing so I assumed some use cases in the implementation of the module, e.g. creating some stuff when not defined by command line. If most of the people think this is bad, let me know and I change that. I just implemented it in a way which is easy to use and fulfills my requirements.


matlockx:
closed for cleaned up duplicate

PR #7618, opened 2014-05-31
Azure module Number 2

matlockx:
Hi,

in addition to #7593 this is another pull request for an azure module. Basically it still lacks:

creating a virtual network
directly restart or stop an instance
It has the following issues:

It still might cause public port conflicts for clusters but not for ssh
There is an open issue with the azure sdk (Azure/azure-sdk-for-python#129) which forced me to monkey patch a method to follow a temporary redirect. Otherwise that bites the cluster functionality hard.
There is also some work to do for the facts which the module returns (and also the inventory plugin). Currently it works (just publishes the structures from the sdk) but I will change that the next days.
In conclusion: this is a starting point and i'm working currently on the above mentioned points. They should follow the next days.

Open questions from my side:

I'm not quite sure whether i have completely understood the whole azure thing so I assumed some use cases in the implementation of the module, e.g. creating some stuff when not defined by command line. If most of the people think this is bad, let me know and I change that. I just implemented it in a way which is easy to use and fulfills my requirements.

### DIFF
--- a/lib/ansible/module_utils/azure.py
+++ b/lib/ansible/module_utils/azure.py
@@ -1,0 +1,49 @@
+import os, sys
+
+
+def azure_common_argument_spec():
+    return dict(
+        certificate_path=dict(aliases=['CERTIFICATE_PATH'], type='str'),
+        subscription_id=dict(aliases=['SUBSCRIPTION_ID'], no_log=True, type='str'),
+        cert_thumbprint=dict(aliases=['CERT_THUMBPRINT'], no_log=True, type='str'),
+        cloud_service_cert=dict(aliases=['CLOUD_SERVICE_CERT'], type='str'))
+
+
+def azure_props(x):
+    return dict((key, getattr(x, key)) for key in dir(x) if key not in dir(x.__class__))
+
+
+def create_azure_conform_name(name):
+    if name:
+        return name.lower().replace("_", "-").replace("--", "-")
+    else:
+        return name
+
+
+class AzureConfig(object):
+    '''
+    To use azure from ansible you need at least:
+    * an azure account
+    * working pem/cer pair (https://github.com/Azure/azure-sdk-for-python)
+    * azure python sdk (pip install azure)
+    '''
+
+    def __init__(self, module):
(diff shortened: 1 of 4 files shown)


matlockx:
sorry for closing/reopening. the old pr was too messy...

mpdehaan:
So let's not merge this one, but rather take the other one first and you can apply changes on top, this way we don't have warring pull requests.

I'm also a bit worried by the comment "I'm not quite sure whether i have completely understood the whole azure thing" from the above.

I'm also not sure how monkey patching even _applies_ to Ansible modules, so this is not making sense to me and it sounds like "but I will change that the next days" in that you are not done with things, so this isn't ready for submission yet.

We definitely dont' want something that just fulfills local requirements, but are wanting something that can work for everyone.

matlockx:
ok i’m fine with that, but let me comment that:
1. "I'm not quite sure whether i have completely understood the whole azure thing” was not related to the api and azure itself (i worked a lot with ec2, maybe this is the reason), but i’m still a bit confused on how to use the api. You need always a hosted service, you need a deployment and in that deployment you can start roles (vms). when creating a deployment it also starts one role but if you want to add another one you need to do a completely other api call. That seems strange to me and let me to the comment "I'm not quite sure whether i have completely understood the whole azure thing”. What also confuses me is that you need to provide different port mappings for roles under the same deployment because they are all only available with one public ip. That means that you need to add several different ‘ssh’ port mappings otherwise you get a conflict error. So maybe this was a bit misleading.
2. i did the monkey patch for a method in the azure sdk. There is an open issue out there for months and they haven’t fixed it yet. It’s related to temporary redirects and the azure sdk does not follow them. So you always get an error. This is really frustrating when you are waiting for the running state of a machine. There is the option of catching the exception and retry the api call by yourself, but this is not really better because it will mostly end in timeouts since the temporary redirect does not disappear quickly. I’m working with this patch and I think it’s ready for submission and I don’t know why this should be something which only fulfills local requirements.

just my 2 cents

matlockx:
I tried a pull request for the azure-sdk: https://github.com/Azure/azure-sdk-for-python/pull/157 maybe that would make the fix easier since i'm also not satisfied with a monkey patch.
At the end: thanks for your feedback. I'm really looking forward for the merge so that I can provide some additions on that.

PR #7619, opened 2014-05-31
fix for docker backwards compatibility

jgreene:
This fixes an issue when using the ansible docker module with an older version of docker.  Specifically 0.9.1 which is the default in ubuntu 14.04.

### DIFF
--- a/library/cloud/docker
+++ b/library/cloud/docker
@@ -545,7 +545,10 @@ class DockerManager:
                   'tty':          self.module.params.get('tty'),
                   }
 
-        if docker.utils.compare_version('1.10', self.client.version()['ApiVersion']) < 0:
+        version = self.client.version()
+        api_version = version['ApiVersion'] if 'ApiVersion' in version else version['Version']
+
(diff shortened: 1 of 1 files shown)


mpdehaan:
"+        api_version = version['ApiVersion'] if 'ApiVersion' in version else version['Version']"

Please don't do one liners like this in submissions, as per the coding guidelines document in the github root.

If you can amend this we can consider application.

library/cloud/docker
```diff
@@ -545,7 +545,10 @@ class DockerManager:
                   'tty':          self.module.params.get('tty'),
                   }
 
-        if docker.utils.compare_version('1.10', self.client.version()['ApiVersion']) < 0:
+        version = self.client.version()
+        api_version = version['ApiVersion'] if 'ApiVersion' in version else version['Version']
+
+        if docker.utils.compare_version('1.10', api_version) < 0:
```
discordianfish, line 551: This doesn't work: 1.10 is an API version for docker 0.10. So if you take ApiVersion you need to compare it to 1.10, but if you take Version you need to compare it to 0.10.

Issue #1171, opened 2012-10-01
ansible is not using .ssh/config


bluszcz:
I am trying to use ansible, unfortunately It does not work with my list of hosts defined in .ssh/config.

In paramiko mode - my hosts are throwing "Name of service not known" - my /etc/ansible/hosts file refers to aliases defined in .ssh/config.

In -c ssh - it tries connect to default 22 port, instead of the ones defines in .ssh/config

jpmens:
I think you have to specify the port in your inventory file:

```
[somegroup]
foo ansible_ssh_port=1234
bar ansible_ssh_port=1235
```

If you're using host aliases, you'll probably need to set those up in `/etc/resolv.conf`.

bluszcz:
This patch make it possible to work with .ssh/config: https://gist.github.com/3811126

mpdehaan:
yeah, but that patch ignores any inventory settings you might have had set.

In cases of connecting to a tunnel where that's not normally in your config
file, that would break some folks.

So, basically some of the reasons we don't get everything from that file is
because Ansible is a systems management application that happens to use
SSH, but it's not, say, EXACTLY just ssh.

See my note about -c ssh and let me know if you have a ansible config file
anywhere, because I think that may be messing up your SSH config port
setting.  The code adds it only when the port is set, and it seems to be
set.
We may also need to change the -c ssh port default to None which may
require a bit more tweaking.

mpdehaan:
Patches, BTW, must be submitted as pull requests.   Though this one is not quite applicable for reasons given above.

bluszcz:
Sorry, just noticed the duplicated tickets. I paste back my reply here:

I don't want to copy .ssh/config and paste it in different version in ansible/hosts file. I have my hosts definition defined in ssh way - and since ansible is using ssh I expect it to honor my settings. Copying and duplication of configuration (which is my hosts definition in .ssh/config) is against the SCM idea.

I would like to have possibility use .ssh/config file - and when they are overwrites from command - just overwrite options from the .ssh/config.

mpdehaan:
So there are reasons to keep your ansible SSH port info in your inventory file -- particularly when sharing ansible among different groups of people and letting them all log in from their user accounts -- has advantages.

While yes, it does use SSH as transport, it's better to think of SSH as just a transport protocol, as to ansible being an extension of /usr/bin/ssh.

[...]

PR #7623, opened 2014-06-01
Synchronize uses ~/.ssh/config like Ansible core does

damianmoore:
If the user has an ~/.ssh/config file with parameters for the current host, Ansible uses those to connect. Before this change, the synchronize module would fail as it would just fall back to user running the command, provided hostname, default port and no key. I believe this makes the module on par with what the user expects running the rest of Ansible.

Related issue:
https://github.com/ansible/ansible/issues/1171

### DIFF
--- a/lib/ansible/runner/action_plugins/synchronize.py
+++ b/lib/ansible/runner/action_plugins/synchronize.py
@@ -17,6 +17,7 @@
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 
 import os.path
+from paramiko import SSHConfig
 
 from ansible import utils
 from ansible.runner.return_data import ReturnData
@@ -73,9 +74,9 @@ class ActionModule(object):
 
     def setup(self, module_name, inject):
         ''' Always default to localhost as delegate if None defined '''
-   
+
         self.inject = inject
-    
+
         # Store original transport and sudo values.
         self.original_transport = inject.get('ansible_connection', self.runner.transport)
         self.original_sudo = self.runner.sudo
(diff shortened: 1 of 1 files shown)


mpdehaan:
Ansible uses those when using the SSH transport, but we should NOT be using paramiko code when not using paramiko, as it creates a dependency we don't wish to have.

I'm also a bit concerned about the "fall back to..." comment above and the size of this patch, , as well as what various uncommented lines added to the synchronize transport here do.   This doesn't mean I think they are wrong, it's just their purpose is not immediately obvious.

Other lines like this:

```
 if 'hostname' in ssh_cfg:
 +            dest_host = ssh_cfg['hostname']
```

Should be written:

```
dest_host = ssh_config.get('hostname', dest_host)
```

If you are talking about synchronize with the OpenSSH connection type, that seems more like this could be solved without the paramiko bits.  

Really, synchronize is intended as a "friendly" wrapper around local_action: shell rsync and in many cases if you want something very specific, that may be a good answer.

Please stop by ansible-devel if you'd like to discuss further.   Thanks!

mpdehaan:
It's also worthwhile to mention we've had people report things before where they want the global SSH config file also read

damianmoore:
Paramiko is already a dependency of Ansible. I can see some related code reading ~/.ssh/config with paramiko in /plugins/inventory/ssh_config.py. I'll see if I can use that function instead of the slight duplication.

I understand that synchronize should be simple but I think it is currently confusing for the user as it behaves differently from the rest of Ansible.

Lets say I am a newish user to Ansible or am just not interested in reading through the source code. I have all my hosts set up in my ~/.ssh/config and Ansible works fine. As time goes on I add more deployment steps, then I add synchronize and if fails with weird errors. I'm not trying to do anything complicated with synchronize, just a basic synchronisation. I end up trying all sorts of options before realising I have to duplicate my settings in an ansible hosts file. Using the synchronize plugin is a nicer syntax in a list of tasks compared to running custom commands.

mpdehaan:
Sounds like we're at an impassse with this one, so I'm going to close the ticket.  I'm still up for discussion though, so stop by ansible-devel if you like.

Thanks!

PR #7626, opened 2014-06-01
Added encryption param to ec2_vol

lowwalker:
Added option to be able to encrypt volumes on creation.

AWS Ref - http://aws.amazon.com/blogs/aws/protect-your-data-with-new-ebs-encryption/

Boto Ref - http://boto.readthedocs.org/en/latest/ref/ec2.html#boto.ec2.connection.EC2Connection.create_volume

### DIFF
--- a/library/cloud/ec2_vol
+++ b/library/cloud/ec2_vol
@@ -24,9 +24,9 @@ version_added: "1.1"
 options:
   instance:
     description:
-      - instance ID if you wish to attach the volume.
+      - instance ID if you wish to attach the volume. 
     required: false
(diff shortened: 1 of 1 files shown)


mpdehaan:
A few quick notes:

```
"+    require: false"
```

This element in the YAML should be "required", not "require".

Also please turn off the automatic whitespace removal bits in your editor, as this makes the diff difficult to review and also breaks line attribution for "git blame" about who last touched each line.

Resubmitting without the end-of-line alterations is probably best.

Thanks!

PR #7629, opened 2014-06-01
Added merge_dicts Jinja2 filter

AndreiPashkin:
Added a new simple filter for merging dicts manually and not in-place.
Ther was [the discussion on mailing list](https://groups.google.com/forum/#!msg/ansible-project/XS1FNBKVUCw/giQwl8QMw_QJ).

ps.
Should I have to add the docs section?

### DIFF
--- a/docsite/rst/playbooks_variables.rst
+++ b/docsite/rst/playbooks_variables.rst
@@ -212,6 +212,9 @@ To get the symmetric difference of 2 lists (items exclusive to each list)::
 
 Dictionary filters
 ------------------
+
+.. versionadded:: 1.7
+
 Merges two dictionaries and returns a third. Original ones remain unchanged::
 
     {{ dict_one | merge_dicts(dict_two) }}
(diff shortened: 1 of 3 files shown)


jirutka:
+1

jimi-c:
Yes, this also needs to be added to the docs. Be sure to include the version added information (see other entries there for examples). Thanks!

AndreiPashkin:
How can I know in which version this feature will be released?

sivel:
Assume for the time being that it is 1.7.  If it doesn't make it into 1.7 you will just need to bump it at a later date.

jimi-c:
Or we will fix it when we merge it in.

AndreiPashkin:
I've added 1.7

darkk:
Is the merge intentionally shallow?

`hash_behaviour=merge` produced deep merge, so [POLA](http://en.wikipedia.org/wiki/Principle_of_least_astonishment) implies that filter `merge_dicts` should provide deep merge too instead of shallow one.

Actually, I've already implemented `merge_hash` filter, so I wonder what is the correct way to contribute it:
- create another pull request with merge_dict filter
- rename my `merge_hash` to `merge_deep`, but it leads to two different merge filters in ansible that may be confusing

jirutka:
@darkk The best solution would be IMHO merge these two filters and use a parameter to choose between deep (default), or shallow merge. Maybe something like `foo | merge_hash(deep=false)` for shallow merge.

PR #7632, opened 2014-06-02
Don't use deprecated BaseException.message in keystone_user

cdwertmann:
fixes error "failed to parse: <attribute 'message' of 'exceptions.BaseException' objects>
TypeError: <attribute 'message' of 'exceptions.BaseException' objects> is not JSON serializable"


jimi-c:
Merged, thanks!

PR #7642, opened 2014-06-02
Option to hide commands from stderr and ansible log

rameshraithatha:
**Syntax:** 
_hide: true_

**Description:** 
Hide commands from stderr and ansible log. This will hide command arguments as well when used in verbose mode. This option would be helpful when you are running some command which requires passwords and you don't wont ansible to display it on your screen or ansible log.

**NOTE:**
 Ansible already has a way to hide passwords, but that works only when you pass something like "password=mypass". It will scan your command and search for a 'password' string, which might not be the case for all use cases.

*_Example: *_ 

```
- name: Display facts
  command: facter -p
  hide: true
```

**Output:**

```
<HOST> REMOTE_MODULE command ARGUMENTS_HIDDEN
changed: [HOST] => {"changed": true, "cmd": "COMMAND_HIDDEN", "delta": "0:00:01.433558", "end": "2014-06-02 16:26:58.353121", "rc": 0, "start": "2014-06-02 16:26:56.919563", "stderr": "", "stdout": ""}
```


bcoca:
look at the existing no_log option, it works at task level, there is a PR
to add it to play level.

## 

Brian Coca
Pedo mellon a minno

rameshraithatha:
no_log is only for syslog, doesn't hide it from stdout/stderr and ansible logs.

jimi-c:
I would think then that no_log should be expanded to also hide it from the JSON output, rather than adding another option for that.

rameshraithatha:
Okay, expanded no_log to hide command from the JSON output and ansible.log

mpdehaan:
Bumping priority to P3 as this is more of what we want vs another option.

Haven't read code yet though.

amenonsen:
Quick review:

Extending no_log to hide module arguments/output per @jimi-c's suggestion seems reasonable, but what this series of commits actually does is to make no_log unconditionally hide them. That may not be what was originally intended (I have no strong opinion either way). 

The code is long out of date, however, and since there's work ongoing to make no_log work in v2, I suggest this PR be closed now.

jimi-c:
@amenonsen I agree, this PR is no longer applicable. We're going to be tweaking no_log a bit more in the near future as well, so we'll close this PR now.

@rameshraithatha thanks for the submission, and our apologies that it has been superseded by current work.

PR #7649, opened 2014-06-03
Allow --vault-password-file to work with a script as well as a flat file

sivel:
This pull request updates `--vault-password-file` to work much like `-i/--inventory` in that it will check if the file is executable and if so it will execute it.

The password provided by the script should be sent to STDOUT (`\r` and `\n` are stripped from the output).

STDERR is not captured to make it easier for users who may need to prompt for input via the script.

Additionally, this pull request adds an `ANSIBLE_VAULT_PASSWORD_FILE` environment variable, and a config file option of `vault_password_file`.

This would allow for things such as #7097 to be implemented in a way more standard with how ansible works with other scripts.


jimi-c:
Merged, thanks!

PR #7652, opened 2014-06-03
Fixed the expected result string in the su command test to reflect what ...

tima:
...actually comes back when it succeeds. Take 2.

### DIFF
--- a/test/units/TestUtils.py
+++ b/test/units/TestUtils.py
@@ -512,7 +512,7 @@ class TestUtils(unittest.TestCase):
         cmd = ansible.utils.make_su_cmd('root', '/bin/sh', '/bin/ls')
         self.assertTrue(isinstance(cmd, tuple))
         self.assertEqual(len(cmd), 3)
-        self.assertTrue('root -c "/bin/sh' in cmd[0])
+        self.assertTrue(' root /bin/sh -c ' in cmd[0])
         self.assertTrue(re.compile(cmd[1]))
         self.assertTrue('echo SUDO-SUCCESS-' in cmd[0] and cmd[2].startswith('SUDO-SUCCESS-'))
 


jimi-c:
I'm seeing this test fail using this, which is very odd. Can you verify the content of the "make_su_cmd" function in `lib/ansible/utils/__init__.py`? There is no logic there to change the format of the command, and it appears that you're getting the old format of the command (before the original BSD su fix).

tima:
I wiped out the Jenkins workspace and re-ran the build. It still fails. Here is that method:

```
def make_su_cmd(su_user, executable, cmd):
    """
    Helper function for connection plugins to create direct su commands
    """
    # TODO: work on this function
    randbits = ''.join(chr(random.randint(ord('a'), ord('z'))) for x in xrange(32))
    prompt = '[Pp]assword: ?$'
    success_key = 'SUDO-SUCCESS-%s' % randbits
    sudocmd = '%s %s %s -c "%s -c %s"' % (
        C.DEFAULT_SU_EXE, C.DEFAULT_SU_FLAGS, su_user, executable or '$SHELL',
        pipes.quote('echo %s; %s' % (success_key, cmd))
    )
    return ('/bin/sh -c ' + pipes.quote(sudocmd), prompt, success_key)
```

We do not have access to su here. The tool exists on ours systems. I just get a "incorrect password" error. I don't see why that would have any effect. These jobs are not running with an ansible.cfg nor am I setting any of the params that method uses via environmental variables.

jimi-c:
That's very odd, I'm not seeing why you're getting a different output string from that. The test looks for `'root -c "/bin/sh'`, which based on the code snippet above is what you should be seeing. Can you stick a "print cmd" into the TestUtils.py test to show what it actually is?

tima:
I found it! My Jenkins job was running python 2.6.5 (we're just moving to 2.7 now -- don't ask.) and building a virtualenv before running nosetest. When I switched to 2.7.6 everything passed. We also had 2.7(.0) available on our machines so I ran that and got a green signal. Go back to 2.6.5 and that test fails. It could even be specific to how the admins installed that version of python -- I'm not sure but it wouldn't be the first time I've seen that happen here.

jimi-c:
Interesting, I'm betting maybe it has to do with the pipes.quote() call there perhaps returning something different. Give this patch a try:

<pre>
diff --git a/test/units/TestUtils.py b/test/units/TestUtils.py
index 241f9cd..f95955d 100644
--- a/test/units/TestUtils.py
+++ b/test/units/TestUtils.py
@@ -512,7 +512,7 @@ class TestUtils(unittest.TestCase):
         cmd = ansible.utils.make_su_cmd('root', '/bin/sh', '/bin/ls')
         self.assertTrue(isinstance(cmd, tuple))
         self.assertEqual(len(cmd), 3)
-        self.assertTrue('root -c "/bin/sh' in cmd[0])
+        self.assertTrue('root -c "/bin/sh' in cmd[0] or ' root -c /bin/sh' in cmd[0])
         self.assertTrue(re.compile(cmd[1]))
         self.assertTrue('echo SUDO-SUCCESS-' in cmd[0] and cmd[2].startswith('SUDO-SUCCESS-'))
 
</pre>

tima:
That seems to fix it for 2.6.5. 

I added a sys.stderr.write(str(cmd)) right before that line and got this from 2.6.5:

```
('/bin/sh -c \'su  root -c "/bin/sh -c \'"\'"\'echo SUDO-SUCCESS-fogjzshulmtrlydlqdkdmyqgmfwsfuir; /bin/ls\'"\'"\'"\'', '[Pp]assword: ?$', 'SUDO-SUCCESS-fogjzshulmtrlydlqdkdmyqgmfwsfuir')
```

With 2.7 I get essentially the same thing though:

```
('/bin/sh -c \'su  root -c "/bin/sh -c \'"\'"\'echo SUDO-SUCCESS-cpfgeieeasjnwecepgflbfngpidzdtzr; /bin/ls\'"\'"\'"\'', '[Pp]assword: ?$', 'SUDO-SUCCESS-cpfgeieeasjnwecepgflbfngpidzdtzr')
```

Also 2.7.6 same thing.

```
('/bin/sh -c \'su  root -c "/bin/sh -c \'"\'"\'echo SUDO-SUCCESS-ypeuwrnfeddjdmilpjavfhwdbdcowpgw; /bin/ls\'"\'"\'"\'', '[Pp]assword: ?$', 'SUDO-SUCCESS-ypeuwrnfeddjdmilpjavfhwdbdcowpgw')
```

I don't know. Let it go since things are working now?

tima:
Having trouble letting this one go. I'm wondering if it's assertTrue evaluating that statement differently from 2.6.5 to 2.7+? There was substantial changes made to the unittest library between those two that its hard for me to compare. That could be something.

jimi-c:
I went ahead and merged my fix in. I'm not sure about the assert, it seems unlikely that such a core thing would have such different behavior between the two versions. But for now, that alternate check should make everyone happy.

tima:
Agreed. Thanks for your assistance.

PR #7653, opened 2014-06-03
ec2_group: Fix for 500 error when creating new security groups

evanccnyc:
Amazon will sometimes not propagate the creation of the new security group fast enough, causing the ansible module to error out with a 500.

This fixes that problem by checking until the security group has propagated and then running:

group = ec2.get_all_security_groups(group_ids=(group.id,))[0]

### DIFF
--- a/library/cloud/ec2_group
+++ b/library/cloud/ec2_group
@@ -230,6 +230,10 @@ def main():
                 # to 0.0.0.0/0 is added automatically but it's not
                 # reflected in the object returned by the AWS API
                 # call. We re-read the group for getting an updated object
+                # amazon sometimes takes a couple seconds to update the security group so wait till it exists
+                while len(ec2.get_all_security_groups(filters={ 'group_id': group.id, })) == 0:
+                    time.sleep(0.1)
+
                 group = ec2.get_all_security_groups(group_ids=(group.id,))[0]
             changed = True
     else:


ap1kenobi:
This seems to be working for me.

jimi-c:
Merged, thanks!

coop:
I am sporadically seeing errors since upgrading to 1.6.5 and using the stacktrace I'm arriving near this change:

``` shell
TASK: [aws | Create vpc security groups] **************************************
changed: [localhost] => (item={'rules': [{'to_port': 22, 'from_port': 22, 'cidr_ip': '0.0.0.0/0', 'proto': 'tcp'}], 'name': 'nat', 'description': 'nat'})
failed: [localhost] => (item={'rules': [], 'name': 'app', 'description': 'app'}) => {"failed": true, "item": {"description": "app", "name": "app", "rules": []}, "parsed": false}
  File "/Users/troll/.ansible/tmp/ansible-tmp-1403945257.6-10932493619541/ec2_group", line 1770, in <module>
    main()
  File "/Users/troll/.ansible/tmp/ansible-tmp-1403945257.6-10932493619541/ec2_group", line 237, in main
    group = ec2.get_all_security_groups(group_ids=(group.id,))[0]
  File "/Library/Python/2.7/site-packages/boto/ec2/connection.py", line 2935, in get_all_security_groups
    [('item', SecurityGroup)], verb='POST')
  File "/Library/Python/2.7/site-packages/boto/connection.py", line 1157, in get_list
    raise self.ResponseError(response.status, response.reason, body)
boto.exception.EC2ResponseError: EC2ResponseError: 400 Bad Request
<?xml version="1.0" encoding="UTF-8"?>
<Response><Errors><Error><Code>InvalidGroup.NotFound</Code><Message>The security group 'sg-94b9d0f1' does not exist</Message></Error></Errors><RequestID>b30294fc-56f8-48b4-8513-0eb236fe6702</RequestID></Response>
```

I wasn't having this issue before the change but I'm not certain it is causing the problem.

coop:
On further investigation making the following change solves the problem:

``` diff
-                while len(ec2.get_all_security_groups(filters={ 'group_id': group.id, })) == 0:
-                    time.sleep(0.1)
+                time.sleep(10)
```

I'm guessing the change was introduced because `ec2.get_all_security_groups(group_ids=(group.id,))[0]` raises an exception where `== 0` doesn't thus allowing the retry functionality. Couldn't this problem be fixed with something like the following:

``` ruby
group = nil

loop do
  result = ec2.get_all_security_groups(filters={ 'group_id': group.id, }))
  if result.any?
    group = result.first
    break
  end
  sleep 0.1
end

group
```

Someone needs to translate my ruby to python but at least this guarantees that the "found" result is the one returned from the method.

evanccnyc:
This wasnt checked in for 1.6.5 only the 1.7 branch.

coop:
cb4b196288ce25bb21501c8c6da88e93ae4a493a is tagged as the 1.6.5 release [as you can see this code](https://github.com/ansible/ansible/blob/cb4b196288ce25bb21501c8c6da88e93ae4a493a/library/cloud/ec2_group#L233-L235) is present.

evanccnyc:
Ah you are correct. Ill see if I can code for the exception on Monday.

PR #7656, opened 2014-06-04
made nested lists diff lengths - caused confusion

tsemerad:
As a relatively-new user to Ansible, the Nested Loop example made me think it was looping through both lists together (`alice` would go with `clientdb`, `bob` with `employeedb`) and not actually nested. I realize now I was just being stupid, but I feel making the example lists different lengths would prevent such confusion in the future.


jimi-c:
Merged, thanks!

PR #7659, opened 2014-06-04
Fix for libcloud 0.14.1

sebgoa:
The extra arguments in libcloud nodes changed in 0.14.0.
The latest stable release on pypi is 0.14.1
This makes the script compatible with the latest libcloud

### DIFF
--- a/plugins/inventory/apache-libcloud.py
+++ b/plugins/inventory/apache-libcloud.py
@@ -222,12 +222,12 @@ class LibcloudInventory(object):
         self.push(self.inventory, self.to_safe('type_' + node.instance_type), dest)
         '''
         # Inventory: Group by key pair
-        if node.extra['keyname']:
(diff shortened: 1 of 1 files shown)


jimi-c:
Ideally, this code should check for both, since some users may still be using the older version of libcloud.

sebgoa:
Yes, ideally. But I don't have time right now to send you a more proper fix. Sorry.
I use the latest libcloud (I am a libcloud dev) and I wanted to test the dynamic inventory (which I wrote),  and realized there was this problem.

I will get back to it ultimately, this script could be improved quite a bit

jimi-c:
No problem, when you update it just let us know. Thanks!

sebgoa:
libcloud is now on 0.16.0 and we provide upgrade for all users. Everyone should have switched to the new version. Can you disregard this and merge a pr I just sent.
thanks

PR #7667, opened 2014-06-04
ansible-pull should respect inventory file settings because, say, Python...

alexclear:
... interpreter path can be non-default on a host

### DIFF
--- a/bin/ansible-pull
+++ b/bin/ansible-pull
@@ -150,7 +150,10 @@ def main(args):
     now = datetime.datetime.now()
     print >>sys.stderr, now.strftime("Starting ansible-pull at %F %T")
 
-    inv_opts = 'localhost,'
+    if not options.inventory:
+        inv_opts = 'localhost,'
+    else:
+        inv_opts = options.inventory
     limit_opts = 'localhost:%s:127.0.0.1' % hostname
     repo_opts = "name=%s dest=%s" % (options.url, options.dest)
 


jimi-c:
Merged, thanks!

willthames:
This breaks ansible-pull when the inventory does not exist until the repository is checked out. See #8324. 

I'd like to fix #8324 while not breaking whatever this fixed.

Without Actual vs Expected Behaviour information in this PR, it's difficult to understand intent of this PR.

sfromm:
I'm confused why this commit uses the inventory file on the checkout operation.  Because all the options needed for the checkout are provided on the cli, using  `'localhost,'` should be used for inventory.  @alexclear, can you provide more information about what this commit fixed for you?

jimi-c:
@willthames the purpose of the PR was to allow the inventory to be specified, if so desired, whereas before the inventory always defaulted to localhost only. I'm curious why this is just now coming up, as it was merged in almost 2 months ago.

@sfromm the point was, that on a system like Solaris, the python interpreter option will need to be set, thus allowing the -i option to specify the inventory where options are set ansible-pull can be used where otherwise it would fail.

sfromm:
Would an environmental variable (eg _ANSIBLE_PYTHON_INTREPRETER_) be a sufficient workaround in case python is at a non-default location?  I realize that doesn't exist at this time.

willthames:
@jimi-c we use an inhouse version of Ansible (because there are patches in Ansible that I raised months ago that are well down your priority lists) into which I merge from upstream on a semi-regular basis (i.e. when I remember or need to). My last update to inhouse Ansible must have brought in this "fix". 

The inventory only defaults to localhost for pulling down the repo - the inventory gets passed in full to the ansible-playbook run (https://github.com/ansible/ansible/blob/bb1941a906e3716a003dff729d9db385fb75b4fc/bin/ansible-pull#L208)

What actually uses the python interpreter though? I guess it's the ansible command that does the checkout? 

It seems that not everyone stores their inventory in the repo being pulled down, but we do, and this worked for us for months.

jimi-c:
@willthames ahh I see. Yes the interpreter would be used for any modules the playbook executed after checkout.

willthames:
My question is whether it's needed for the checkout itself. If it isn't we could just move the change until after the checkout. If it is, we have a conflict that I'm really not sure how to resolve (unless we could somehow try to read the inventory file and ignore it if need be until after checkout)

jimi-c:
It is not, and I believe that would work. Just be sure to test it with inventories both ways and we'll get it merged in ASAP.

PR #7682, opened 2014-06-06
bigip_monitor bugfix for check mode

srgvg:
Fixes case where properties are retrieved for a non existent monitor, which happens in check mode.


jimi-c:
Merged, thanks!

PR #7684, opened 2014-06-06
Fix documentation of deb option in apt module

JeanFred:
The apt module may install a .deb package on the remote machine, not on local.
Fix the misleading documentation.


jimi-c:
Merged, thanks!

PR #7687, opened 2014-06-06
Fix neutron floating IP allocation for networks w/ multiple subnets

ryanpetrello:
For networks that have both a v4 and a v6 subnet, the floating IP plugin
currently has two problems:
- When determining the subnet for the provided `internal_network_name`, it
  assumes that the first item in the list of subnets is the one you want.
  Instead, it should pick the first v4 subnet.
- When multiple fixed IP's exist for a given port (as is the case in a network
  a v4 and a v6 subnet), neutron needs a hint as to which fixed IP to associate
  to the floating IP address (the v4 one).


jimi-c:
Merged, thanks!

PR #7694, opened 2014-06-07
azure - fix for: temporary redirect, not deleted vhds, missing locations...

matlockx:
... and role types, async api calls.

this addresses pr #7593 
- temporary redirects will cause errors in the azure sdk, the workaround here is to try for 5 minutes the original api call again and again (with sleeps) until it succeeded
- missing thumbprint error is related to async api calls which haven been finished yet. the solution is to wait for the response as the windows documentation points out
- added missing locations and role types
- delete vhds when roles (vms) are deleted

### DIFF
--- a/library/cloud/azure
+++ b/library/cloud/azure
@@ -223,7 +223,6 @@ def create_virtual_machine(module, azure):
     Returns:
         True if a new virtual machine was created, false otherwise
     """
-
     name = module.params.get('name')
     hostname = module.params.get('hostname') or name + ".cloudapp.net"
     endpoints = module.params.get('endpoints').split(',')
@@ -302,8 +301,8 @@ def create_virtual_machine(module, azure):
             module.fail_json(msg="failed to create the new virtual machine, error was: %s" % str(e))
 
 
-        deployment = azure.get_deployment_by_name(service_name=name, deployment_name=name)
-        return (changed, urlparse(deployment.url).hostname, deployment)
+    deployment = azure.get_deployment_by_name(service_name=name, deployment_name=name)
+    return (changed, urlparse(deployment.url).hostname, deployment)
 
 
 def terminate_virtual_machine(module, azure):
--- a/library/cloud/azure
+++ b/library/cloud/azure
@@ -92,6 +92,11 @@ options:
   wait_timeout:
     description:
       - how long before wait gives up, in seconds
+    default: 600
+    aliases: []
+  wait_timeout_redirects:
+    description:
+      - how long before wait gives up for redirects, in seconds
     default: 300
     aliases: []
   state:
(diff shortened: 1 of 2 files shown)


library/cloud/azure
```diff
@@ -320,25 +340,27 @@ def terminate_virtual_machine(module, azure):
...
-                    # Azure api error, wait a second and retry
-                    time.sleep(1)
-                    continue
-                time.sleep(5)
+            # gather disk info
+            results = []
+            for role in deployment.role_list:
+                role_props = azure.get_role(name, deployment.name, role.role_name)
+                if role_props.os_virtual_hard_disk.disk_name not in disk_names:
+                    disk_names.append(role_props.os_virtual_hard_disk.disk_name)
```
matlockx, line 347: remember disks

library/cloud/azure
```diff
@@ -320,25 +340,27 @@ def terminate_virtual_machine(module, azure):
...
+
+            result = azure.delete_deployment(name, deployment.name)
+            _wait_for_completion(azure, result, wait_timeout, "delete_deployment")
+
+            for disk_name in disk_names:
+                azure.delete_disk(disk_name, True)
```
matlockx, line 353: delete disks

library/cloud/azure
```diff
@@ -351,34 +373,41 @@ def get_azure_creds(module):
...
-
     # create azure ServiceManagementService object
     subscription_id, management_cert_path = get_azure_creds(module)
-    azure = ServiceManagementService(subscription_id, management_cert_path)
 
-    if module.params.get('state') == 'absent':
-        changed = terminate_virtual_machine(module, azure)
+    wait_timeout_redirects = int(module.params.get('wait_timeout_redirects'))
+    if LooseVersion(windows_azure.__version__) <= "0.8.0":
+        #wrapper for handling redirects which the sdk <= 0.8.0 is not following
```
matlockx, line 402: handle temporary redirects in a wrapper. only needed for version <= 0.8.0

library/cloud/azure
```diff
@@ -255,36 +288,22 @@ def create_virtual_machine(module, azure):
...
-            except WindowsAzureMissingResourceError as e:
-                pass # deployment still not available
-            except WindowsAzureError as e:
-                # got a bad response from azure, wait a second and then try again
-                time.sleep(1)
-                continue
-            if deployment:
-                break
-            else:
-                time.sleep(5)
```
matlockx, line 285: we wait in the _wait_for_completion block by using the request ID, so this is not needed anymore.

library/cloud/azure
```diff
@@ -227,7 +258,9 @@ def create_virtual_machine(module, azure):
         if ssh_cert_path:
             fingerprint, pkcs12_base64 = get_ssh_certificate_tokens(module, ssh_cert_path)
             # Add certificate to cloud service
-            azure.add_service_certificate(name, pkcs12_base64, 'pfx', '')
+            result = azure.add_service_certificate(name, pkcs12_base64, 'pfx', '')
+            _wait_for_completion(azure, result, wait_timeout, "add_service_certificate")
```
matlockx, line 261: wait for the certificate to be added to the cloud servers, otherwise the create_deployment below might fail

library/cloud/azure
```diff
@@ -198,24 +238,15 @@ def create_virtual_machine(module, azure):
...
-    deployment = None
-    try:
-        deployment = azure.get_deployment_by_name(service_name=name, deployment_name=name)
-    except WindowsAzureMissingResourceError as e:
-        pass # no such deployment
-    except WindowsAzureError as e:
-        module.fail_json(msg="failed to create the new deployment, error was: %s" % str(e))
-
-    if deployment:
+    cloud_service_name_available = azure.check_hosted_service_name_availability(name)
```
matlockx, line 240: check hosted service availability can be used here

library/cloud/azure
```diff
@@ -198,24 +238,15 @@ def create_virtual_machine(module, azure):
...
         changed = False
     else:
         changed = True
-
         # Create cloud service if necessary
         try:
-            existing_service_names = [service.service_name for service in azure.list_hosted_services()]
-            if not name in existing_service_names:
-                azure.create_hosted_service(service_name=name, label=name, location=location)
+            result = azure.create_hosted_service(service_name=name, label=name, location=location)
```
matlockx, line 247: we here know that the service name is available

jimi-c:
Merged, thanks!

PR #7698, opened 2014-06-08
Update intro_installation.rst

gsker:
This is different wording for using the PPA to install.  I also added the installation of apt-add-repository package which isn't always installed.

(I'm a newbie but while running these I wished whatever style was being used on the docs pages would have let me  cut and paste the commands without having to edit out the $ characters manually....)

### DIFF
--- a/docsite/rst/intro_installation.rst
+++ b/docsite/rst/intro_installation.rst
@@ -173,10 +173,11 @@ Latest Releases Via Apt (Ubuntu)
 
 Ubuntu builds are available `in a PPA here <https://launchpad.net/~rquillo/+archive/ansible>`_.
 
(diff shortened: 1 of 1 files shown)


gsker:
This is different wording for using the PPA to install. I also added the installation of apt-add-repository package which isn't always installed.

mpdehaan:
Merged, thanks!

Not entirely sure what you mean about the "$" parts.

gsker:
With the "$" characters in the code-block:: bash segments, it is not possible to cut-n-paste the three lines of code. One has to carefully select the lines one by one without the $ characters.  If the blocks did not have the $, one could just cut and paste the whole block.

Does that clarify?

PR #7700, opened 2014-06-09
support hash merging for ansible_facts

bennojoy:
This PR enables hash merging for ansible_facts returned by set_fact or include_vars modules.

### DIFF
--- a/lib/ansible/playbook/__init__.py
+++ b/lib/ansible/playbook/__init__.py
@@ -469,10 +469,10 @@ class PlayBook(object):
                 for res in result['results']:
                     if type(res) == dict:
                         facts = res.get('ansible_facts', {})
-                        self.SETUP_CACHE[host].update(facts)
+			self.SETUP_CACHE[host] = utils.combine_vars(self.SETUP_CACHE[host], facts)
             else:
                 facts = result.get('ansible_facts', {})
-                self.SETUP_CACHE[host].update(facts)
+		self.SETUP_CACHE[host] = utils.combine_vars(self.SETUP_CACHE[host], facts)
             if task.register:
                 if 'stdout' in result and 'stdout_lines' not in result:
                     result['stdout_lines'] = result['stdout'].splitlines()


mpdehaan:
This patch looks like it may contain indentation problems, can you investigate?

Also, can you provide an example of what the before and after result of this patch is?

mpdehaan:
I'm not entirely positive we want the values of facts to be blended with facts from previous runs, so this may also be a question.

For instance, if facts return

```
{
    "color" : "red"
}
```

And later:

```
{
    "number": 12
}
```

It would return both color and number in future runs, with no way to remove the first entry.

This may break future plans around fact expiration where we would like this to be based only on the key.

bennojoy:
Hi Michael,

The indentation looks good in my branch.. i am not exactly sure why the commit shows it slighty misaligned.

Here's a gist which shows the before and after for the patch.

https://gist.github.com/bennojoy/076e5d6e5d59e260f8e7

and yes, we wont be able to remove color: "red"  from a hash but that would be only applicable if the user has  specified hash_behaviour=merge in ansible.cfg else the hash would be replaced by
"number: 12 "

redbaron:
changing merging behavior globally looks too restrictive as different roles can expect one or another. Plus there is a question of should it be shallow or deep merge.

What to you think about approach, where you "inherit" from parent at the level where you override. 

```
---
# group_vars
- groupvar:
    nested:
      a: 42
      b: 69
    nomerge:
      x: 10

# host_vars
- @groupvar:
    addme: 99
    nomerge: "no merge for this key, replacing instead"
    @nested:
      c: "merged with groupvar.nested"
```

mpdehaan:
yeah I'm thinking we may not want this one.  Perhaps a subject for discussion on ansible-devel

Meanwhile, @redbaron, we generally try to push back on new syntax and that strikes me as rather super complex - so if there are use cases that are not covered, I'm open to talking about solutions on the -project list - there may be a better way to model your problem so this isn't neccessary that we could work through.

Let us know!

erikdw:
@bennojoy: FYI, the reason the indentation was off is that you are using tab characters in your commit.  Tabs are evil.  Don't use tabs to indent code unless you absolutely must (e.g., in Makefiles).

Issue #6881, opened 2014-04-07
lineinfile should detect missing EOF newline


FiloSottile:
##### Issue Type:

Bug Report
##### Ansible Version:

`ansible 1.5.3`
##### Environment:

Linux
##### Summary:

When lineinfile adds a line at the end of a file, it should detect if the last character of the file is not a `\n`, and add it in case.
##### Steps To Reproduce:

```
- name: add ClientAliveInterval to sshd_config
  lineinfile: dest=/etc/ssh/sshd_config regexp="^ClientAliveInterval "
    line="ClientAliveInterval 15"
  sudo_user: root
```
##### Expected Results:

```
[...]
UsePAM yes
UseDNS no
ClientAliveInterval 15
ClientAliveCountMax 4
```
##### Actual Results:

```
UsePAM yes
UseDNS noClientAliveInterval 15
ClientAliveCountMax 4
```

mpdehaan:
This ticket seems to be rather that it didn't add the line, not that it had left off a \n.

Can you confirm the intent of this bug report?  Thanks!

FiloSottile:
Please focus on the `ClientAliveInterval 15` line, the rest is for context.

As you can see it's there in both Expected and Actual, but in Expected it's concatenated to the last line.

jabclab:
I've come across this bug too, e.g. default `/etc/ssh/sshd_config` on vanilla Centos 6.5. I want to disable root login via SSH so have this in a playbook:

```
- name: disable root login over SSH
  lineinfile: line='PermitRootLogin no'
              regexp='^PermitRootLogin'
              dest='/etc/ssh/sshd_config'
              insertafter='EOF'
              validate='sshd -T -f %s'
  sudo: yes
```

The issue is that the config file has no trailing `\n` at `EOF` so we see this:

#### before

```
# Example of overriding settings on a per-user basis
#Match User anoncvs
#   X11Forwarding no
#   AllowTcpForwarding no
#   ForceCommand cvs server
```

#### after

```
# Example of overriding settings on a per-user basis
#Match User anoncvs
#   X11Forwarding no
#   AllowTcpForwarding no
#   ForceCommand cvs serverPermitRootLogin no
```

akkie:
I have the same problem with the same file on Amazon Linux. Is there a workaround for this issue?

rafaltrojniak:
Hello, 
We also were hit by this bug.
[...]

PR #7702, opened 2014-06-09
Fix lineinfile module to properly handle files with no newline at EOF

ahtik:
Currently when using lineinfile with insertafter=EOF for a file that has no newline at EOF then the added line is not prefixed with a newline. Instead it's joined with the last line.

As a concrete example, Ubuntu distro default /etc/ssh/sshd_config has no newline so lineinfile module cannot be used to add a new line to the end of the file.

Also found a related bug: #6881

This fix prefixes insertafter=EOF line inserts with a newline when needed (file is not empty and EOF has no newline).

### DIFF
--- a/library/files/lineinfile
+++ b/library/files/lineinfile
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 
 # (c) 2012, Daniel Hokka Zakrisson <daniel@hozac.com>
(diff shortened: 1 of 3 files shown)


jimi-c:
Merged, thanks!

ahtik:
Thank you!

PR #7705, opened 2014-06-09
Ensure random jinja filter doesn't get cached in loops.

benosman:
When using the random filter in a template, I noticed that when used in a loop, jinja2 caches the result of the random filter when the parameters are constant.

For instance the following code for me always produces the same random number.

```
{% for item  in [0,1,2] %}
{{ 24 | random }}
{% endfor %}
```

I ran it through a debugger and found it wasn't reaching the random filter method for the second iteration, so Jinja is caching it. In Jinja's own random filter it uses the environmentfilter decorator to disable this.

This patch imitates that and for me it fixes the issue.

Info on the decorator:
http://jinja.pocoo.org/docs/api/#utilities

Jinja's random filter:
https://github.com/mitsuhiko/jinja2/blob/master/jinja2/filters.py#L362

### DIFF
--- a/lib/ansible/runner/filter_plugins/core.py
+++ b/lib/ansible/runner/filter_plugins/core.py
@@ -28,6 +28,7 @@ from ansible import errors
 from ansible.utils import md5s
 from distutils.version import LooseVersion, StrictVersion
 from random import SystemRandom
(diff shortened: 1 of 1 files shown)


jimi-c:
Merged, thanks!

PR #7710, opened 2014-06-10
use the full list of available campfire sounds

brad:
From the list found [here](http://www.emoji-cheat-sheet.com/)

I haven't tested every last one of these, but I have tested most of them directly in Campfire and I've tested several of the previously missing sounds using the Ansible module.

### DIFF
--- a/library/notification/campfire
+++ b/library/notification/campfire
@@ -31,11 +31,15 @@ options:
     description:
       - Send a notification sound before the message.
     required: false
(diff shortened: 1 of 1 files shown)


mpdehaan:
Merged thanks!

Apologies to everyone with volume turned on in advance :)

mpdehaan:
Users using older versions running this may have some module errors about those being invalid choices, but 1.7 will be out in about a month, and I suspect those using sounds and hitting errors can figure it out.

PR #7718, opened 2014-06-10
Don't skip printing on unmatched tags

omgjlk:
The way ansible-playbook is using the API for list-tags, each play is
processed/printed as it is read, is different from how plays are
actually executed, where all the plays are read and only if a tag was
unmatched in /all/ the plays would the execution error out.

Without this change, tasks from a play that has unmatched tags would not
print, even if that tag is later matched in a different play. In the
scenario of --skip-tags, those tasks would actually run, because they
don't match the tag to be skipped.

This change makes listing of tasks more inline with how playbooks are
actually executed.

### DIFF
--- a/bin/ansible-playbook
+++ b/bin/ansible-playbook
@@ -227,9 +227,6 @@ def main(args):
                     unknown_tags = ((set(pb.only_tags) | set(pb.skip_tags)) -
                                     (matched_tags | unmatched_tags))
 
-                    if unknown_tags:
-                        continue
-
                 if options.listhosts:
                     print '  play #%d (%s): host count=%d' % (playnum, label, len(hosts))
                     for host in hosts:


mpdehaan:
Having a tiny bit of trouble following.

Can you please show expected vs actual (or before vs after) results and an example playbook or CLI invocation to help understand what you mean here?

Thanks!

omgjlk:
Sure. Consider this playbook:

```
- name: derp
  hosts: localhost

  tasks:
    - name: task1
      debug: msg="task1"

    - name: task2
      debug: msg="task2"

- name: hurp
  hosts: localhost

  tasks:
    - name: task3
      debug: msg="task3"
      tags: taggy

    - name: task4
      debug: msg="task4"
```

There is one task with a tag, "task3".
If I run ansible-playbook test.yaml --list-tasks I should see them all

```
playbook: test.yaml

  play #1 (derp):
    task1
    task2

  play #2 (hurp):
    task3
    task4
```

And if I do --list-tasks and --tags taggy
I should see just task3, and I do:

```
playbook: test.yaml

  play #2 (hurp):
    task3
```

But if I do --skip-tags taggy I would expect to see task1, task2, and task4 listed. But without my patch, this is what you see:

```
playbook: test.yaml

  play #2 (hurp):
    task4
```

All the tasks from play derp have been skipped. This does not match what happens if these were to actually run without just list-tasks:

```
$ ansible-playbook -i localhost, test.yaml  --skip-tags taggy

PLAY [derp] *******************************************************************

ok: [localhost]

TASK: [task1] *****************************************************************
ok: [localhost] => {
    "msg": "task1"
}

TASK: [task2] *****************************************************************
ok: [localhost] => {
    "msg": "task2"
}

PLAY [hurp] *******************************************************************

ok: [localhost]

TASK: [task4] *****************************************************************
ok: [localhost] => {
    "msg": "task4"
}

localhost                  : ok=5    changed=0    unreachable=0    failed=0
```

As you can see, task1 and task2 fire, along with task4.

With my patch, --list-tasks now looks like:

```
$ ansible-playbook -i localhost, test.yaml --list-tasks --skip-tags taggy
playbook: test.yaml

  play #1 (derp):
    task1
    task2

  play #2 (hurp):
    task4

$ ansible-playbook -i localhost, test.yaml --list-tasks -t taggy

playbook: test.yaml

  play #1 (derp):

  play #2 (hurp):
    task3
```

Does this help?

One could argue that printing of tasks without playing them (--list-tasks) could in fact just be a callback (much like --step is) instead of duplicate code for handling playbooks residing in ansible-playbook, but I went the less invasive route of fixing up ansible-playbook.

bcoca:
this is fixed in current developent

PR #7723, opened 2014-06-11
Added module to trim old snapshots in ec2.

Surgo:
This module trimming old snapshots as same as boto's trim_snapshots.


magnetik:
I need a module doing this sort of things.

Is there any reason why did you closed the pull request?

PR #7729, opened 2014-06-11
Error checking and reporting on interface ufw rule

rgbkrk:
If someone fails to specify the direction on a rule for an interface, they are met with this message:

```
failed: [my_super_cool_server] => {"failed": true}
msg: ERROR: Invalid token 'on'
```

This just provides a check and a more informative message.


jimi-c:
Merged, thanks!

PR #7730, opened 2014-06-11
 Have paramiko use /etc/ssh_known_hosts

vegitron:
Fixes an issue with a confusing error: "paramiko: The authenticity of host '[host]' can't be established" when ssh on the command line doesn't complain

### DIFF
--- a/lib/ansible/runner/connection_plugins/paramiko_ssh.py
+++ b/lib/ansible/runner/connection_plugins/paramiko_ssh.py
@@ -147,6 +147,7 @@ class Connection(object):
         self.keyfile = os.path.expanduser("~/.ssh/known_hosts")
 
         if C.HOST_KEY_CHECKING:
+            ssh.load_system_host_keys("/etc/ssh/ssh_known_hosts")
             ssh.load_system_host_keys()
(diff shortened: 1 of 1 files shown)


vegitron:
This was pull request #6902, before I accidentally added another commit to it.

jimi-c:
I'm assuming this is done when the playbook is using ssh to connect back to localhost?

vegitron:
My servers (including the deployment host) are managed by another group in the organization.  They have host keys for all other servers in the system known_hosts file.  Without this diff, sshing to the host works, but ansible fails, and it isn't clear why.

Since hosts are frequently added, just copying the system known_hosts file to my account (and having every other deployment account do the same thing) isn't a viable option.

amenonsen:
This PR (a) still seems relevant, (b) needs to be targeted to lib/ansible/plugins/connections/paramiko_ssh.py in devel (where it should apply without trouble).

amenonsen:
I've created a new PR #11730 that targets the (trivial, one-line) change to the correct file. This PR can be closed now.

bcoca:
closing by merged rebased PR, attribution was kept

PR #7731, opened 2014-06-11
 Adds an async flag to django_manage

vegitron:
This flag makes it possible to start long-running/daemon processes with ansible.


jimi-c:
This should be done through the normal async method available to plays. If you wish to fire-and-forget, simply set the async polling time to 0. If you have any further questions, please let us know. Thanks!

vegitron:
OK, thanks - the documentation made it look like async was connected to the commands module.  I can work with that, though I'd be happier if I could use async with with_items :)

PR #7733, opened 2014-06-12
Update intro_installation.rst

gsker:
Corrected an error about which package needed to be installed to get the apt-add-repository command in Via Apt (Ubuntu).

Remove $ characters from bash code blocks. They make cutting and pasting multiple lines into a bash shell prompt impossible. Does that make sense?


jimi-c:
We keep those `$` there to indicate it is a shell command. If there are other typos to correct, please submit just those as a PR. Thanks!

PR #7735, opened 2014-06-12
pkgutil: exception when 'site' parameter is missing

toddmowen:
The first of these commits fixes a bug in the pkgutil module. Although the 'site' parameter is intended to be optional, omitting it would cause the following exception if the requested package was already installed:

```
  File "/tmp/ansible-tmp-1402553652.18-229268856436124/pkgutil", line 1398, in <module>
    main()
  File "/tmp/ansible-tmp-1402553652.18-229268856436124/pkgutil", line 153, in main
    if not package_latest(module, name, site):
  File "/tmp/ansible-tmp-1402553652.18-229268856436124/pkgutil", line 78, in package_latest
    site = pipes.quote(site)
  File "/usr/lib/python2.6/pipes.py", line 271, in quote
    for c in file:
TypeError: 'NoneType' object is not iterable
```

The fix was simply not to call `pipes.quote(site)` until after checking whether the argument is Null.

The other two commits in this PR are general cleanup in the pkgutil module, including removal of use_unsafe_shell.

### DIFF
--- a/library/packaging/pkgutil
+++ b/library/packaging/pkgutil
@@ -74,12 +74,10 @@ def package_installed(module, name):
 
 def package_latest(module, name, site):
     # Only supports one package
-    name = pipes.quote(name)
-    site = pipes.quote(site)
     cmd = [ 'pkgutil', '--single', '-c' ]
     if site is not None:
-        cmd += [ '-t', site ]
-    cmd.append(name)
+        cmd += [ '-t', pipes.quote(site) ]
+    cmd.append(pipes.quote(name))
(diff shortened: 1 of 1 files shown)


jimi-c:
This appears to be two un-related patches (fixing the traceback vs. run_command changes). Could you split this up and resubmit it as two separate PRs? Thanks!

toddmowen:
I will do that, but will wait until the first (more critical) PR is accepted before submitting the second -- just to make sure it merges cleanly.

Issue #7696, opened 2014-06-08
zfs module incompatible with new "share.nfs" and "share.smb" properties


toddmowen:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.6.2 
##### Environment:

Managing Solaris 11.1.
##### Summary:

The version of ZFS released with Solaris 11.1 replaced the "sharenfs" and "sharesmb" properties with a hierarchy of share.\* properties (see [docs](http://docs.oracle.com/cd/E26502_01/html/E29007/gayne.html)). This breaks Ansible's zfs module.
##### Steps To Reproduce:

```
tasks:
- name: Setup /export/opencws as ZFS filesystem
  zfs: name=rpool/export/opencsw state=present sharenfs=on
```
##### Expected Results:

First run: create the specified ZFS filesystem, with NFS sharing enabled.

Second run: do nothing.
##### Actual Results:

First run: as expected.

Second run:

```
invalid output was: Traceback (most recent call last):
  File "/tmp/ansible-tmp-1402197494.77-156175913445546/zfs", line 1617, in <module>
    main()
  File "/tmp/ansible-tmp-1402197494.77-156175913445546/zfs", line 398, in main
    zfs.set_properties_if_changed()
  File "/tmp/ansible-tmp-1402197494.77-156175913445546/zfs", line 306, in set_properties_if_changed
    if current_properties[prop] != value:
KeyError: 'sharenfs'
```
##### Explanation:

With the new version of ZFS (specificially, pools running pool version 34 or higher), filesystem creation continues to work correctly because "sharenfs" and "sharesmb" are aliased to "share.nfs" and "share.smb" respectively.

But rerunning the task (expected to make no changes) fails because the Ansible module expects to find the property "sharenfs" in the output of "zfs get -H all", which actually looks like this:

```
...
rpool/export/opencsw    shadow  none    -
rpool/export/opencsw    share.* ...     local
rpool/export/opencsw    snapdir hidden  default
...
```

Notice that the actual value of share.nfs isn't even displayed. It can be obtained by running "zfs get -H share.all".

toddmowen:
A possible fix would be to look for "share.*" in the output of `zfs get all`. If found, run `zfs get share.all` and merge the output into the dictionary.

jimi-c:
@toddmowen would you like to work on that? Just send us a PR if so, and reference this issue in it.

jimi-c:
[...]

PR #7737, opened 2014-06-12
zfs: fix incompatibility - share.nfs vs sharenfs (issue #7696)

toddmowen:
Implement the fix proposed on the [ticket](https://github.com/ansible/ansible/issues/7696).

Note that zfs itself accepts "sharenfs" as an alias for "share.nfs", so there is no problem setting the property. The new name only causes issues because we dump all properties to a dict and then expect it to contain "sharenfs".


jimi-c:
Merged, thanks!

Issue #5761, opened 2014-01-25
Incorrect sector size for 4K (4096) disk


ciupicri:
Ansible reports that a [Hitachi HDS724040ALE640](http://www.hgst.com/tech/techlib.nsf/techdocs/3C330019E6ED04A48825797700626F0D/$file/DS7K4000_ds.pdf) disk has 512 bytes sectors, even they're really 4096 bytes (4K).

The output of the **setup** module is:

```
# ansible -v -i <(echo localhost) -c local -m setup '*'
    ...
    "sdd": {
        "holders": [],
        "host": "SATA controller: Intel Corporation 7 Series/C210 Series Chipset Family 6-port SATA Controller [AHCI mode] (rev 04)",
        "model": "Hitachi HDS72404",
        "partitions": {
            "sdd1": {
                "sectors": "7814035087",
                "sectorsize": 512,
                "size": "3.64 TB",
                "start": "2048"
            }
        }, 
        "removable": "0",
        "rotational": "1",
        "scheduler_mode": "cfq",
        "sectors": "7814037168",
        "sectorsize": "512",
        "size": "3.64 TB",
        "support_discard": "0",
        "vendor": "ATA"
    }
```

**fdisk** says something else:

```
Disk /dev/sdd: 3.7 TiB, 4000787030016 bytes, 7814037168 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 4096 bytes
I/O size (minimum/optimal): 4096 bytes / 4096 bytes
```

I'm running kernel-3.12.8-300.fc20.x86_64 and `/sys/block/sdd/queue` looks like this:
- hw_sector_size = 512 (this is used by Ansible right now)
- logical_block_size = 512
- physical_block_size = 4096
- minimum_io_size = 4096
- optimal_io_size = 0

The files are explained in the [kernel documentation](https://www.kernel.org/doc/Documentation/block/queue-sysfs.txt):

> hw_sector_size (RO)
> This is the hardware sector size of the device, in bytes.
> 
> logical_block_size (RO)
> This is the logical block size of the device, in bytes.
> 
> physical_block_size (RO)
> This is the physical block size of device, in bytes.
> 
> minimum_io_size (RO)
> This is the smallest preferred io size reported by the device.
> 
> optimal_io_size (RO)
> This is the optimal io size reported by the device.

[...]

PR #7740, opened 2014-06-12
changed sector size fact source 

jkramarz:
changed sector size fact source from hw_sector_size to physical_block_size, fixes issue #5761


jimi-c:
Merged, thanks!

PR #7744, opened 2014-06-12
mysql_user module: add option to require SSL to GRANT queries

egorchakov:
Adding an option to append 'REQUIRE SSL' to GRANT queries, which is necessary for establishing SSL connections to a MySQL database.


mpdehaan:
New attributes need to have "version_added" in the DOCUMENTATION section.

egorchakov:
@mpdehaan Done, version_added set to "1.7"

PR #7745, opened 2014-06-12
Update playbooks_intro.rst

danrue:
Remove reference to learning about modules in a previous chapter, since the module chapter comes later.


mpdehaan:
Seems this would be better to link to the appropriate section rather than removing it, since docs are a bit more random access.

As a result, I'm going to close this one out.   a PR to link the chapter/section would be welcome!

PR #7748, opened 2014-06-13
fix for dnssec-tools install, fixes #7452

### LINKED ISSUE #7452: Installation of 'dnssec-tools' fails on Ubuntu 14.04 using apt
##### Issue Type:

Bug report
##### Ansible Version:

Ansible 1.6.1
##### Environment:

Running on OSX 10.9.2. Connecting to an Ubuntu 14.04 server
##### Summary:

Running a task such as 

```
- name: bind | Install packages/s
  apt: pkg={{ item }} state=installed
  with_items:
    - dnssec-tools
```

Will cause the ansible process to hang. Sivel on IRC ran a strace on the apt process and found it was hanging on the 

```
'Processing triggers for ureadahead (0.100.0-16) ...'
```

part. Running the command 

```
apt-get -y -o "Dpkg::Options::=--force-confdef" -o "Dpkg::Options::=--force-confold" install dnssec-tools
```

on the server causes no issues and the process completes successfully.
##### Steps To Reproduce:

Install dnssec-tools using a task like

```
- name: bind | Install packages/s
  apt: pkg={{ item }} state=installed
  with_items:
    - dnssec-tools
```
##### Expected Results:

Dnssec-tools should be installed successfully and the ansbile task should continue to run.
##### Actual Results:

The process hangs and completes no further tasks. The packages are however installed just fine if you SSH in and look at the server yourself.

kcghost:
fixes #7452

problem with using subprocess.PIPE for for large data described here:
http://thraxil.org/users/anders/posts/2008/03/13/Subprocess-Hanging-PIPE-is-your-enemy/

subprocess can hang if PIPE is used, change uses temporary files instead

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -1161,13 +1161,16 @@ class AnsibleModule(object):
         if data:
             st_in = subprocess.PIPE
 
+        temp_out = tempfile.TemporaryFile()
+        temp_err = tempfile.TemporaryFile()
+
         kwargs = dict(
             executable=executable,
             shell=shell,
             close_fds=close_fds,
             stdin= st_in,
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE 
+            stdout=temp_out,
+            stderr=temp_err 
         )
 
         if path_prefix:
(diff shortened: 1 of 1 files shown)


jimi-c:
I've merged in the above commit, which resolves this in a different way. Thanks!

PR #7755, opened 2014-06-13
module rax_scaling_group: safe normalization of nics

sivel:
pyrax currently returns a `net-id` key in it's 'nic' dictionaries.  We can't use `net-id` in asg requests, it must be uuid.

If pyrax updates to return `uuid` this code will prevent the module from breaking.

I will be submitting a pull request to pyrax to get this change made.


jimi-c:
Merged, thanks!

PR #7758, opened 2014-06-13
Add optional docker_api_version parameter to Docker Module

jmoretti:
When using the Ansible Docker module, I found myself needing to target a later version of the Docker remote API then is defaulted by Docker-py library.  I added an optional parameter to the Ansible Docker module that allows you to specify the desired docker_api_version.  If not provided, the module will default to Docker-py library default version.

### DIFF
--- a/library/cloud/docker
+++ b/library/cloud/docker
@@ -107,6 +107,13 @@ options:
     required: false
     default: unix://var/run/docker.sock
     aliases: []
+  docker_api_version:
+    description:
+      - Remote API version to use
+    required: false
+    default: docker-py default remote API version
+    aliases: []
+    version_added: "1.7"
   username:
     description:
       - Set remote API username
@@ -388,7 +395,8 @@ class DockerManager:
 
         # connect to docker server
         docker_url = urlparse(module.params.get('docker_url'))
-        self.client = docker.Client(base_url=docker_url.geturl())
+        docker_api_version = module.params.get('docker_api_version')
+        self.client = docker.Client(base_url=docker_url.geturl(), version=docker_api_version)
 
 
     def get_links(self, links):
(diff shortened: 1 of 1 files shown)


mpdehaan:
This is adding a "version_added" attribute in the docstring of the new attribute, though ultimately I'm thinking a better fix is for the docker folks to start understanding API compatibility.

What happens if this is left off?  The functions we are calling against the python libraries are not changing.

Ultimately I'm kind of inclined to think the important part of our docker support is being able to build images in the docker file, and this module will eventually fall by the wayside as better placement engines start to arise.

jmoretti:
This isn't so much a "version_added", more-so a "version_desired".  I agree for the need for better API management, and hopefully now that Docker has reached 1.0, we'll start to see that, but given the rapid evolution of Docker over the last year, I am inclined to be more lenient of API issues.

As it stands, if a version is not supplied, the Python library will use the same default API version, just as it was previously.  This just gives the Ansible user an option, if they so desire, to target a specific version of the Docker API.

We're using Ansible currently for our orchestration, but we will be moving to a more purpose-built engine in the future.   That being said, I can definitely see value of using this particular Docker module for standing up development environments and for automated testing of Docker images after a CI build process.

sivel:
@jmoretti "version_added" is part of the docs specification.  You added a new option to the module, so you need to document in which version of ansible the option was added.  The value in this case would be "1.7".

On another note, I am inclined to say that the module should only work with Docker 1.0 and the most recent API version included with Docker at 1.0 release time which I believe is 1.12.

I do not believe, without a lot of logic, that the docker modules can reasonably be able to work with all versions of docker and the docker API.  They experienced huge flux during development to the 1.0 release.

jmoretti:
@sivel Sorry for the delay, I've been under the weather and am just getting caught back up.  I've added the version_added: "1.7" to the docstring as requested and rebased.

I agree that we should probably target Docker >= 1.0.  Also, as of 15 days ago, Docker has released docker-py 0.3.2 which now defaults to API version 1.12.  This effectively gets me around the problem I was trying to solve in the short term; however, since Docker does give you an option at targeting different API versions, I still see value in providing Ansible users with the ability to optionally target a non-docker-py-default version of the Docker API.

jimi-c:
# Merged

Hi!

issue is resolved.

Thank you!

PR #7762, opened 2014-06-13
Do variable expansion for 'register' variable names

redbaron:
Allows to use {{}} expansion in 'register:' variable names. Goes very well with parametrized roles

### DIFF
--- a/lib/ansible/playbook/__init__.py
+++ b/lib/ansible/playbook/__init__.py
@@ -461,6 +461,8 @@ class PlayBook(object):
         contacted = results.get('contacted', {})
         self.stats.compute(results, ignore_errors=task.ignore_errors)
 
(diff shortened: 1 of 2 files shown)


mpdehaan:
This is not something I'm really wanting to include because the destination variable name could change based on various inputs, and can lead to some very hard to understand playbooks.

Ansible does not wish to be a programming language.

Parameterized roles would still be ok because they could just not rely on a variable after they were done.

If you still think this should be a thing, stop by ansible-devel, and we can discuss details.

redbaron:
The use case for that is to have conditional handlers.  Because handlers are executed at the end of the play, without parametrizing registered variables blend and only last one wins

``` yaml
---
- name: restart mydb-{{realm}} service
  service: name=mydb-{{realm}} state=restarted
  when: is_formatted_{{realm}}|success
```

Issue #7763, opened 2014-06-14
Incorrect hostname setting for RHEL7


rmarchei:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.7
##### Environment:

Managing RHEL 7
##### Summary:

To set the hostname on RHEL7 Ansible edits the /etc/sysconfig/network file (RedHatStrategy) but it should use the hostnamectl command, like the FedoraStrategy
##### Steps To Reproduce:

Use the hostname module to set hostname on a RHEL7
##### Expected Results:

Hostname is changed using hostnamectl and nothing is written on /etc/sysconfig/network
##### Actual Results:

Ansible writes the HOSTNAME variable on /etc/sysconfig/network but this doesn't change the "static hostname" as shown by the hostnamectl command.

jimi-c:
Merged, thanks!

PR #7770, opened 2014-06-14
Use different hostname strategy for RHEL >= 7. Fixes 7763

rmarchei:
fixes issue #7763

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -142,6 +142,18 @@ def get_distribution():
         distribution = None
     return distribution
 
+def get_distribution_version():
+    ''' return the distribution version '''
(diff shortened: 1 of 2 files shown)


jimi-c:
Why have you duplicated the `get_distribution()` into this module? That should be removed and the one from module_utils used instead.

rmarchei:
It is `get_distribution_version()`, similar to `get_distribution()` (that's why I copied part of the code) but it returns the version of the OS (like 6.5, 7, etc)

jimi-c:
I see, however that should still be in `module_utils/basic.py` and not in the module itself.

rmarchei:
perfect! I'm changing it right now

jimi-c:
Merged, thanks!

PR #7773, opened 2014-06-14
documents group, owner, mode as they're referenced in the examples

tristanfisher:
This require simply adds descriptions for the group, owner, and mode parameters for the copy_module.html page, as they are referenced in the example usage section.  The description text is from the file_module page for the purposes of uniformity.

### DIFF
--- a/library/files/copy
+++ b/library/files/copy
@@ -69,6 +69,21 @@ options:
     choices: [ "yes", "no" ]
     default: "yes"
     aliases: [ "thirsty" ]
+  mode:
+    description:
+      - mode the file or directory should be, such as 0644 as would be fed to chmod
+    required: false
+    default: null
+  owner:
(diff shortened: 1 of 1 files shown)


tristanfisher:
@halberom pointed out over IRC that ['FILE_COMMON_ARGUMENTS'](https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L104) gets called by `copy`, so perhaps this PR should be extended to include the outstanding parameters? 

I still see this PR as a minor improvement on its own, but if it's better to close this one and create a more comprehensive one, I'm okay with that.

jimi-c:
It appears that this module is just lacking the syntax to include the files documentation (`extends_documentation_fragment: files`). I'll add that rather than merge this in. Thanks!

PR #7774, opened 2014-06-15
[openbsd-pkg] Add note regarding package flavors

mwpher:
This adds a note to the openbsd-pkg module which explains how to specify
package flavors as well as how to avoid ambiguity errors on a package
with multiple flavors.

Technically, this is an Ansible issue, but it's pretty likely the user will
encounter this problem.

--Matt

### DIFF
--- a/library/packaging/openbsd_pkg
+++ b/library/packaging/openbsd_pkg
@@ -53,6 +53,13 @@ EXAMPLES = '''
 
 # Make sure nmap is not installed
 - openbsd_pkg: name=nmap state=absent
+
+# Specify a pkg flavour with '--'
+- openbsd_pkg: name=vim--nox11 state=present
+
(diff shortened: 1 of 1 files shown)


mwpher:
Something I'm wondering about: whenever I execute a playbook that specifies a package flavor, the resource for the package always reports itself as "changed", even though the package is already installed.

```
changed: [127.0.0.1] => (item=sshguard--) => {"changed": true, "item": "sshguard--", "name": "sshguard--", "state": "present"}
```

jimi-c:
In regard to your question, it would appear that the logic to determine if something is already installed does not properly account for the flavor specified, or it does not handle the 'default' flavor syntax correctly. Please open another issue regarding that.

jimi-c:
Merged, thanks!

PR #7783, opened 2014-06-15
Find IPs on OpenStack clouds without floating-ips

emonty:
The floating-ip extension, while pretty ubiquitous, is not a
foregone conclusion. Specifically, Rackspace, while also
served by the rax module, is a valid OpenStack cloud and can
be interacted with directly via nova interfaces.

Add support for determining public and private IPs for
OpenStack clouds that don't use floating ips by reading
the public and private keys from the addresses dict.


emonty:
Closing this, as it was included in a different PR.

PR #7784, opened 2014-06-15
Implement a static global module alias registry

emonty:
In order to allow for non-code-copying module renames that
allow for both sides of the rename to exist concurrently. It
should really be used sparingly, because at the end of the
day renames are trouble. Nevertheless, if they do eventually
need to happen, mitigating their pain is friendlier than not.


mpdehaan:
Closing this as the linked PR seems to double-include this, so this is redundant.

PR #7785, opened 2014-06-15
Swap neutron alias

emonty:
Includes https://github.com/ansible/ansible/pull/7784 as well, but additionally takes the next step by making the base modules neutron_ \* and making quantum_\* aliases to that.


mpdehaan:
I'm not ok with the idea of global aliases being hard coded here.

We'll actually rename the modules at some point in the future, most likely, as part of a regular release cycle.

PR #7789, opened 2014-06-16
fix for repo urls like ssh://git@github.com/ansible/ansible.git

mattjbray:
Fixes the following error when using `accept_hostkey=yes` in the git module:

```
failed to add git@github.com hostkey: getaddrinfo git@github.com: Name or service not known
```


jimi-c:
Merged, thanks!

PR #7798, opened 2014-06-16
Allow npm executable to contain spaces

koenpunt:
NVM has a special script which loads the correct node version before executing a command.
The syntax for this is `/usr/local/nvm/nvm-exec npm list --json` for example.
But previously when specifying `executable='/usr/local/nvm/nvm-exec nvm'` this would not work because the string was treated as one executable. This PR fixes that.


jimi-c:
Merged, thanks!

PR #7801, opened 2014-06-17
Allow gem executable to contain spaces

koenpunt:
rbenv has a special exec function which loads the correct ruby version before executing a command.
The syntax for this is `/usr/local/rbenv/bin/rbenv exec gem ...` for example.
But previously when specifying `executable='/usr/local/rbenv/bin/rbenv exec gem'` this would not work because the string was treated as one executable. This PR fixes that by splitting the executable on spaces.

Relates to #7798

### DIFF
--- a/library/packaging/gem
+++ b/library/packaging/gem
@@ -91,12 +91,12 @@ import re
 
 def get_rubygems_path(module):
     if module.params['executable']:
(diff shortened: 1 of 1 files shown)


jimi-c:
Merged, thanks!

PR #7802, opened 2014-06-17
Fixes password lookup docs

vcastellm:
Fixes some string module naming mistakes that were driving me crazy.


jimi-c:
Merged, thanks!

PR #7807, opened 2014-06-17
Add EC2 instance "termination protection" support and "source_check_dest" changeability

amandolo:
Another EC2 option useful during provisioning.

### DIFF
--- a/library/cloud/ec2
+++ b/library/cloud/ec2
@@ -952,13 +952,13 @@ def create_instances(module, ec2, override_count=None):
         for res in res_list:
             running_instances.extend(res.instances)
 
-        # Enabled by default by Amazon
-        if not source_dest_check:
+        # Enabled by default by AWS
+        if source_dest_check is False:
             for inst in res.instances:
                 inst.modify_attribute('sourceDestCheck', False)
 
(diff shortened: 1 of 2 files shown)


jimi-c:
Can you modify the termination protection on an instance after its created? If so, it looks like your patch only modifies the setting when they're created.

amandolo:
I modified the module to permit changes to "termination_protection" and "source_dest_check" after instance creation (when STATE is "running" or "stopped").

The EC2 module has these behaviours:
- state "present/absent" only create/terminate instances
- state "running/stopped" change instances state and change some options (for now only termination_protection, source_dest_check).

analytically:
+1

zytek:
+1

szibis:
+1

jimi-c:
When testing this, I got the following API error:

<pre>
You may only modify the sourceDestCheck attribute for VPC instances
</pre>

So it appears this can only be True when the instance is on a VPC. Also, it appears that some additional commits have snuck into this branch, so we cannot merge this as it stands. Could you please correct the above issue and flatten the relevant commits into a single commit? Thanks!

PR #7808, opened 2014-06-17
Updates django_manage so it populates VIRTUAL_ENV

joelcrocker:
Virtualenv's activate script sets the VIRTUAL_ENV environment variable to the path of the virtualenv. Checking this variable is a reasonably common way to verify that execution is happening in a virtualenv. It would be convenient if this module's virtualenv handling set this environment variable.


jimi-c:
Merged, thanks!

PR #7810, opened 2014-06-17
Set default vg_options to empty string

rivik:


jimi-c:
Merged, thanks!

srgvg:
That was a quick one. Actually #6859 has been a trivial patch in queue for about 6 weeks now, whilst it should have been trivial enough to merge in 1.6 window :)

@jimi-c my patch also tweaked another line, you might want to look at that too.

PR #7813, opened 2014-06-17
host-pattern as dynamic inventory script argument

lorencarvalho:
### Hello!

Very interested in taking Ansible out for a test drive in my current position, however I ran into a problem with the dynamic script implementation that didn't quite work for our current inventory system ([range](https://github.com/ytoolshed/range)).

The way range works (for us) is to provide a cluster name to expand, and you recieve back a list of hosts, for example:

```
$ range --expand "%datacenter.cluster.instance"
host1.localdomain
host2.localdomain
host3.localdomain
```

You can also get information about a single host by various means (akin to the --host implementation expected to be returned by dynamic inventory scripts)

While it is possible to return every cluster and subsequent members of each cluster by querying range, it's a somewhat expensive and potentially slow thing to do (especially when dealing with 10s of thousands of hosts).

The cleanest and most obvious way to deal with this would be to provide a clustername to the dynamic inventory query in order to only get back cluster members from specific clusters, akin to:

```
$ /etc/ansible/range.py --list %datacenter.cluster.instance
{
  '%datacenter.cluster.instance': [
    'host1.localdomain',
    'host2.localdomain',
    'host3.localdomain'
    ],
  '_meta': {
    'hostvars': {
      'host1.localdomain': {},
      'host2.localdomain': {},
      'host3.localdomain': {}
      }
    }
  }
```

In order to accomplish this without having to do anything drastic to ansible.inventory.Inventory() I simply added a boolean '-r' argument to bin/ansible that indicates that the host-pattern should be passed to the dynamic inventory script as an argument. This limits the scope of the query to only the cluster I care about, while keeping the usage syntax the same. 

I recognize that there are probably better ways to do this than what is represented in my pull request, but think of it more as a proof of concept. All tests pass, all default syntax remains the same, merely adding the option to pass the host-pattern to the dynamic inventory script.

Thanks for looking!

P.S. Sorry about the whitespaces in the diff, I have a VIM plugin that automatically cleans them up :(

### DIFF
--- a/bin/ansible
+++ b/bin/ansible
@@ -45,12 +45,12 @@ class Cli(object):
         ''' create an options parser for bin/ansible '''
 
         parser = utils.base_parser(
-            constants=C, 
-            runas_opts=True, 
-            subset_opts=True, 
+            constants=C,
+            runas_opts=True,
+            subset_opts=True,
             async_opts=True,
-            output_opts=True, 
-            connect_opts=True, 
+            output_opts=True,
+            connect_opts=True,
             check_opts=True,
             diff_opts=False,
             usage='%prog <host-pattern> [options]'
@@ -135,7 +135,8 @@ class Cli(object):
 
             if not options.ask_vault_pass:
                 vault_pass = tmp_vault_pass
-
+        if options.inventory_arguments is True:
+            options.inventory = tuple([options.inventory, pattern])
         inventory_manager = inventory.Inventory(options.inventory)
         if options.subset:
             inventory_manager.subset(options.subset)
--- a/bin/ansible-playbook
+++ b/bin/ansible-playbook
@@ -100,6 +100,9 @@ def main(args):
     if (options.ask_vault_pass and options.vault_password_file):
             parser.error("--ask-vault-pass and --vault-password-file are mutually exclusive")
 
+    if options.inventory_arguments is True:
+        options.inventory = tuple([options.inventory, ' '.join(args)])
+    print options.inventory
     inventory = ansible.inventory.Inventory(options.inventory)
     inventory.subset(options.subset)
     if len(inventory.list_hosts()) == 0:
--- a/bin/ansible-playbook
+++ b/bin/ansible-playbook
@@ -102,7 +102,6 @@ def main(args):
 
     if options.inventory_arguments is True:
         options.inventory = tuple([options.inventory, ' '.join(args)])
-    print options.inventory
     inventory = ansible.inventory.Inventory(options.inventory)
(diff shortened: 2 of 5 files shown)


jimi-c:
So typically we have avoided making any changes to the options the inventory plugins take or pass, and rather prefer that configuration files or environment variables be used to modify the dynamic scripts behaviors. Also, a lot of the plugins we use cache results to prevent needless querying, so you may want to add that to your inventory script as well.

lorencarvalho:
Yeah I've been fumbling around with the ansible-playbooks workflow today and it seems that offloading  as much of the work as possible to the dynamic inventory script is optimal. (right now in ansible-playbook I am doing the same tuple packing and just having the dynamic inventory script check if the arg is a file/playbook or clustername before building the json response). 

```
# bin/ansible:
$ /etc/ansible/range.py --list %datacenter.cluster.instance
{
  '%datacenter.cluster.instance': [
    'host1.localdomain',
    'host2.localdomain',
    'host3.localdomain'
    ],
  '_meta': {
    'hostvars': {
      'host1.localdomain': {},
      'host2.localdomain': {},
      'host3.localdomain': {}
      }
    }
  }

# bin/ansible-playbook
/etc/ansible/range.py --list ~/test-playbook.yaml ~/test-playbook2.yaml
{
  '%datacenter.cluster1.instance': [
    'host1.localdomain',
    'host2.localdomain',
    'host3.localdomain'
    ],
  '%datacenter.cluster2.instance': [
    'host4.localdomain',
    'host5.localdomain'
   ], 
  }    
```

Something along those lines.

While I totally get the mindset of using env vars or config files and caching results, it's kind of counter-intuitive to do so with range, it's sole purpose is serving inventory-like metadata so it has a lot of caching built in. It's intended to be queried as needed, having two layers of cache could introduce some inconsistencies or unexpected behavior. 

My only beef with not having the script args come from ansible itself is that I'd have to have some sort of wrapper or something on top of ansible, which I suppose would work but it doesn't feel as concise. A 'hack' instead of just extending the tool itself. That being said, if you want to close this PR, I have no problem with that, I can simply patch my changes in locally, considering it's a very specific use case anyway :)

mpdehaan:
Yes typically this might be done by setting an inventory script like "INVENTORY_CUSTOMER_ID=asdf1234" when working with large inventory databases and wishing to return a subset of it.   Your inventory script could return something like RANGE_PATTERN=foo and constrict the "all" group to such a subset, possibly.

I'm afraid we can't change the API of the inventory scripts at this time.

lorencarvalho:
No worries -- thanks for looking!

PR #7817, opened 2014-06-18
Issue #7358 Create non-unique parameter for group module

hahnicity:
I created a non_unique parameter for all platforms except AIX which requires a chsec call to change group ID security. Furthermore the chsec call has three separate options which is not conducive to being represented with a boolean.

### DIFF
--- a/library/network/get_url
+++ b/library/network/get_url
@@ -76,6 +76,14 @@ options:
     version_added: "1.3"
     required: false
     default: null
+  md5sum:
+    description:
+      - Performs the same service as sha256sum but using a MD5 checksum. This option
+        cannot be set at the same time as sha256sum. MD5 checksum verification is 
+        available for usage on systems with Python 2.4, whereas SHA-256 checksum
+        verification is not.
+    required: false
+    default: null
   use_proxy:
     description:
       - if C(no), it will not use a proxy, even if one is defined in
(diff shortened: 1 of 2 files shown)


jimi-c:
So one thing I noticed in a quick test, when trying to remove a non-unique group the module fails because it's finding the first group (which belongs to a user) and tries to remove that:

<pre>
$ ansible -m group -a "name=testing2 state=absent gid=1001 non_unique=yes" ubuntu1204
192.168.122.110 | FAILED >> {
    "failed": true, 
    "msg": "groupdel: cannot remove the primary group of user 'testing'\n", 
    "name": "testing2"
}
</pre>

That will need to be addressed before this can be merged in.

hahnicity:
Weird, I have had success with the operation

```
(ansible)greg@gregr:~/workspace/python/ansible/hacking$ ansible targets -m group -a "name=testing2 non_unique=yes gid=1005 state=absent" -i inventory -s
localhost | success >> {
    "changed": true, 
    "name": "testing2", 
    "state": "absent"
}
```

I also am using ubuntu 12.04. 

My steps were:
- Create testing1 with gid 1005
- Create testing2 with gid 1005
- Remove testing2 with gid 1005

Perhaps I am missing some of your steps or I am not testing on the correct env. I will try to play around with this to reproduce

hahnicity:
Nvm I reproduced when I looked more closely at your comment

hahnicity:
Blah... I did some pretty bad branching on this one...

hahnicity:
jimi-c I'm not sure your concern is easily addressed: I would have to do some perl mucking which would not go through groupdel 

```
[ ! -f /etc/group.lock ] && perl -ne 'next if /^<group name>:/; print' /etc/group > /etc/group.lock && mv /etc/group.lock /etc/group && grpconv
```

If this still blocks the feature from being merged then I guess the feature itself could be closed. If we can not resolve this issue I will just close the pull and place my other commits in a separate pull.

jimi-c:
Yes, I think this would be a deal-breaker, since it results in a somewhat unpredictable removal. So we'll go ahead and close this, and you can resubmit other things as needed. Thanks!

PR #7818, opened 2014-06-18
adding selective rule addition & removal support to ec2_group while keep...

praddy:
Hi,

I am adding support for selectively adding or removing rules to/from security groups. The default behaviour is still the same old, "exact: no" should be used to trigger the selective behaviour. I have also refactored some of the code here as part of the this exercise.

Thanks,
- praddy

### DIFF
--- a/library/cloud/ec2_group
+++ b/library/cloud/ec2_group
@@ -44,6 +44,17 @@ options:
     required: false
     default: 'present'
     aliases: []
+  exact:
+    version_added: "1.7"
+    description:
+      - |
+        If 'yes' & state is 'present' only the specified rules will be added/kept & all others will be
+        removed, if state is 'absent' the whole group will be removed. If 'no' & state is 'present' the specified
+        rules will be added & the existing ones won't be touched, if state is 'absent' then only the specified
+        rules will be removed & not the whole group
+    required: false
+    default: 'yes'
+    aliases: []
 
 extends_documentation_fragment: aws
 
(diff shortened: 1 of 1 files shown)


jimi-c:
Very nice, I'll get this into queue for testing as soon as possible.

praddy:
Awesome, thanks.

praddy:
Added another commit which fixes another existing issue.

mpdehaan:
I find the exact parameter here pretty confusing in particular WRT the explanation.   Typically we've done something like "append=yes|no".

praddy:
Rebased the changes. @mpdehaan I chose exact because `exact=yes` means make the respective security group exactly what I wish to (in the module args), in case it already exists. That, in my opinion, is a destructive change and should be explicitly called out. Not sure if `append=no` sends as strong a message as `exact=yes`. In case of `exact=no`, `state` reflects what exactly would be done: `state=present` means respective rules will be added & `state=absent` means respective rules will be removed. I think using `append=yes` & `state=absent` together will be more confusing.

mpdehaan:
# Possible Duplicated Efforts

Hi!

It looks like the work from this pull request was duplicated by a recent change in the module.

For some additional information:

```
* we've recently accepted parameters that allowed two new parameters "purge_rules" and "purge_egress_rules" that control these items seperately
* these seem to offer the same capabilities
```

praddy:
I think the patch that has been accepted doesn't provide a way to selectively remove rules. This patch provides that functionality too.

PR #7819, opened 2014-06-18
fix group by tag so that equal sign is preserved

twalling:
Unless I'm mistaken, I don't think the equal sign should be passed along in the to_safe function call because its getting converted to an underscore. I updated my own local copy of ec2.py and wanted to submit a pull request to get this updated.

### DIFF
--- a/plugins/inventory/ec2.py
+++ b/plugins/inventory/ec2.py
@@ -371,7 +371,7 @@ class Ec2Inventory(object):
 
         # Inventory: Group by tag keys
         for k, v in instance.tags.iteritems():
-            key = self.to_safe("tag_" + k + "=" + v)
+            key = self.to_safe("tag_" + k) + "=" + self.to_safe(v)
             self.push(self.inventory, key, dest)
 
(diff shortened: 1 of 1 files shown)


jimi-c:
This is not a mistake - we currently convert all non-text, numbers or dash characters to underscores. Unfortunately, if your patch were merged in, I believe it would break a considerable number of users plays which reference tagged groups as they're currently created via to_safe().

We may be open to considering an option for ec2.ini which would make the character set in to_safe() configurable, so feel free to work on that approach. Also, this may be a good topic for the ansible-devel mailing list. If you have any further questions, please let us know. Thanks!

twalling:
Thanks. I think I misunderstood how those group names are intended to be used and jumped the gun as I haven't even written my playbooks yet. I was just getting the dynamic inventory working first. I can see now how the final string doesn't even really matter since the string is less a name value pair and more just an identifier for the inventory.

jimi-c:
Yep, and no problem!

Issue #7728, opened 2014-06-11
Synchronize ssh user variable templating error


berdario:

jimi-c:
Have you tried this on a recent version of 1.6.x or devel? I believe this issue may have been fixed recently.

berdario:
Yes, I tried it on 1.6.3 and it's affected as well

tima:
I can confirm this problem is happening for us also. Today we upgraded a system that was using 1.4.4 to 1.6.3 and synchronize stopped working. After a lot of playing around we discovered that the asterisk in the src is the problem. Remove that and the error (response code 23/ rsync: link_stat [etc.] failed) goes away. That can be worked around though it is suboptimal for us as removing the asterisk brings the parent directory with it and requires us to adjust the path through out our setup. I did a diff of the synchronize code (both action and module files) and don't see why this is happening in there making me think it's somewhere away from the synchronize specific code. I will try to come back to finding where this is going wrong, but I make no promises at the moment.

tima:
BTW: When we run the rsync command as the Ansible verbose output indicates, the command runs without error.This has us questioning if what we see in that output is what is being executed. Not sure though.

tima:
I found the problem, but I don't have a solution. See PR #7820 for details.

mpdehaan:
Assuming "the variable" above is like ~/ or $HOME?

jimi-c:
Hi, this situation is no longer present in the devel branch (which will be ansible 2.0). 

If you continue seeing any problems related to this issue, or if you have any further questions, please let us know by stopping by one of the two mailing lists, as appropriate:
- https://groups.google.com/forum/#!forum/ansible-project - for user questions, tips, and tricks
- https://groups.google.com/forum/#!forum/ansible-devel - for strategy, future planning, and questions about writing code

Because this project is very active, we're unlikely to see comments made on closed tickets, but the mailing list is a great way to ask questions, or post if you don't think this particular issue is resolved.

Thank you!

PR #7820, opened 2014-06-18
turned on unsafe shell handling so synchronize can work with unix wildcards

tima:
This pull request addresses the problem reported in #7728.

### DIFF
--- a/library/files/synchronize
+++ b/library/files/synchronize
@@ -314,7 +314,7 @@ def main():
 
     cmd = ' '.join([cmd, source, dest])
     cmdstr = cmd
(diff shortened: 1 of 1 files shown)


jimi-c:
For security reasons, we cannot turn on unsafe shell processing when there are any inputs which are user-specified. Doing so would open up the possibility of injecting shell scripting into commands modules execute, so unfortunately we cannot merge this. Instead, you could use `with_fileglob` (http://docs.ansible.com/playbooks_loops.html#looping-over-fileglobs) to achieve the same effect.

If you have any further questions, please let us know. Thanks!

PR #7823, opened 2014-06-18
Add mysql_index module

caphrim007:
This module provides the ability for ansible to add and remove indices
from a table in MySQL without needing to use other, more raw, modules to
accomplish similar goals.

### DIFF
--- a/library/database/mysql_index
+++ b/library/database/mysql_index
@@ -1,0 +1,331 @@
+#!/usr/bin/python
+
+# Ansible module to manage mysql indexes
+#
+# Certain parts are taken from Mark Theunissen's mysqldb module
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


library/database/mysql_index
```diff
@@ -0,0 +1,331 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: mysql_index
+short_description: Adds or removes an index from a MySQL table.
+description:
+   - Adds or removes an index from a MySQL table.
+version_added: "1.6"
```
omgjlk, line 28: This will probably have to be 1.7, as 1.6 is already released.

mpdehaan:
Hi, 

While this is interesting I think most people will apply this through a database schema upgrade or execute a SQL script.

You are welcome to continue to use this module though we are unlikely to include it in Ansible core.

Feel free to make a post about it to ansible-project list to let people know about it.

Thanks!

caphrim007:
@mpdehaan hey michael, actually, we were using a SQL script before we developed this module. We created it because re-using the SQL script caused duplicate indices to be created (non idempotent). We also were able to wrap this module in a with_items that resulted in a cleaner-looking playbook and elimination of said SQL script altogether.

srgvg:
FYI, @caphrim007 if you are interested, I'm collecting various plugins and modules not in core here: https://github.com/ginsys/ansible-plugins

PR #7825, opened 2014-06-18
Added new module to support LXC containers

cloudnull:
The new module will allow users to control LXC containers from ansible. 
The module was built for use in LXC 1.0 or greater and implements most 
of what can be done using the `lxc-*` clients.

The module provides:
- build containers
- destroy containers
- archive containers
- list all containers
- info from a single container
- clone a container locally
- start / stop / restart containers
- run commands within containers
- add/modify lxc config for a container
- module supports multiple backends including LVM

### DIFF
--- a/library/cloud/lxc
+++ b/library/cloud/lxc
@@ -1,0 +1,1587 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2014, Kevin Carter <kevin.carter@rackspace.com>
+# (c) 2014, Hugh Saunders <hugh.saunders@rackspace.co.uk>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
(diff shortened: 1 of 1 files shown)


mpdehaan:
Hmm, this one is interesting as while raw LXC is good stuff, increasingly people are going through docker and other layers on top of that, and I think investment in those areas may be better longer term, than leaving people confused as to which route to go.  That being said, I understand the purpose.

his module will likely need to be split out as we are de-emphasizing modules that support a lot of "command" pattern type stuff and wishing things to be declarative, and also trying to get more inventory stuff routed down towards inventory scripts versus fact modules.

cloudnull:
Hey @mpdehaan,

Thanks for reviewing the module. I would be happy to split the module up. Any guidance into how you would like to see the module carved up or consolidated I would appreciate.

Question: would this module be more suitable if it were making better use of states? Something like, state=[present, absent, running, stopped, frozen] and I ripped out the command syntax? I think the only methods that would not fit within that ideology would be `create_tar` and `config` which both could be done outside this module using other core ansible modules. Additionally if I moved to using states I think I could drop the `info` and just return data based on the state of a container after the run.  However, I'd have to ponder how to best implement the `list` method as that would have no set state. 

We'd love to see raw lxc hit core ansible as we've found it to be an excellent containerization system that provides flexibility and stability without a lot of additional overhead. Like I said I'd love any feedback that you or anyone else may have. Thanks again.

mark-casey:
+1
Regarding Docker et al., imho the latest buzz is logically always going to be a bit behind a sizable chunk of what is in production now, so I think raw lxc had purpose previously and will continue to do so. I'd definitely find this useful with some of my own environments.

DavidWittman:
:+1: for raw LXC support

mpdehaan:
Hi!

Thanks you very much!

cloudnull:
@mark-casey @DavidWittman - This PR has been updated in the "ansible-modules-extras" repo as found here: "https://github.com/ansible/ansible-modules-extras/pull/123" If you have a moment to review the new PR I would greatly appreciate it.

PR #7826, opened 2014-06-19
add ssh dir, user group creation

tedder:
- the ssh directory was already being created if generate_ssh_key=yes.
  This genericizes it slightly. (more to come along these lines?)
- group creation fails if the group doesn't exist. Sure, one could call
  the 'group' module, but that gets clumsy (for instance, when iterating
  through lists). useradd can autocreate the group, which is kinda slick.


tedder:
killing this. the -U option doesn't work (consistently) as I expected. The .ssh creation can be done through the authorized_key module.

PR #7827, opened 2014-06-19
ec2 module: Delay instance tagging to later in the instance creation process

willthames:
##### Issue Type:

Bugfix Pull Request
##### Ansible Version:

ansible 1.7 (devel aac194e639) last updated 2014/06/19 11:13:25 (GMT +1000)
##### Environment:

N/A
##### Summary:

When creating instances with tags using the ec2 module, the tagging bit can result in
InvalidInstanceID.NotFound errors.

By delaying the tagging until the last part of instance creation, we should be typically more fortunate (handling all such race conditions might need more work such as retry with backoffs)
##### Steps To Reproduce:

This appears to be sporadic but has been reported elsewhere
http://blog.picloud.com/2013/02/17/dealing-with-the-inconsistent-ec2-api/

Until today we'd never hit it (or at least not enough to concern us) but today all instance creation has been failing. 

Essentially create an ec2 instance with the ec2 module using the tags argument. However, a successful instance creation just suggests that the timing glitch was avoided.
##### Expected Results:

Instance gets created and tagged appropriately
##### Actual Results:

failed: [instance123] => {"failed": true}
msg: Instance tagging failed => InvalidInstanceID.NotFound: The instance ID 'i-abcd1234' does not exist

### DIFF
--- a/library/cloud/ec2
+++ b/library/cloud/ec2
@@ -913,12 +913,6 @@ def create_instances(module, ec2, override_count=None):
         except boto.exception.BotoServerError, e:
             module.fail_json(msg = "Instance creation failed => %s: %s" % (e.error_code, e.error_message))
 
-        if instance_tags:
-            try:
-                ec2.create_tags(instids, instance_tags)
-            except boto.exception.EC2ResponseError, e:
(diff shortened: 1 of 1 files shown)


jimi-c:
Merged, thanks!

PR #7829, opened 2014-06-19
return properties of the newly created scaling policy

bennojoy:
This PR enable's the module  to return the properties of the created scaling policy. This is especially useful
when creating cloud watch alarms which requires the arn of the policy.


jimi-c:
Merged, thanks!

PR #7831, opened 2014-06-19
Use isinstance() to type check dicts and lists.

benosman:
Here is a pull-request in response to the post on ansible-devel.

https://groups.google.com/forum/#!topic/ansible-devel/d1eUS46H-ZM

I went a bit beyond the original thinking for consistency, I've converted all type checking for dicts AND lists to use is_instance.

All changes are straightforward, except for one in  lib/ansible/utils/template.py, which should be checking whether the class is HostVars. 

https://github.com/smoothify/ansible/commit/b677ec33ff5e5867b1f03c5b44a916ca53b18f58
Here, I had to add a scoped import to avoid a circular dependency error. Rather than importing, and using isinstance, an alternative method would be to do the following:

`type(var).__name__ == 'HostVars'`

### DIFF
--- a/examples/scripts/yaml_to_ini.py
+++ b/examples/scripts/yaml_to_ini.py
@@ -64,7 +64,7 @@ class InventoryParserYaml(object):
 
         # first add all groups
         for item in yaml:
-            if type(item) == dict and 'group' in item:
+            if isinstance(item, dict) and 'group' in item:
                 group = Group(item['group'])
 
                 for subresult in item.get('hosts',[]):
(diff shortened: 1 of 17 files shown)


mpdehaan:
Need to understand the use cases here rather than the technical points before we consider this. 

Let's continue discussion on the devel list.

mpdehaan:
# Question About This One

Hi!

- the changes above seem correct, however, we don't know what instances are specifically broken at this time requring the above change
- we're open to applying something like the above given 1.8 has just started but things have shifted underfoot and this will need to be reapplied
- trying to rebase this ourselves was not clearcut, and given we didn't know of any bugs this solved we decided to not apply this at this time

PR #7842, opened 2014-06-19
adding min() and max() filters for use in templates

mpeters:
In the process of trying to make some of my configuration file templates more dynamic to handle different configurations for different environments I ran into a couple of cases where a `min()` and `max()` filter would have been extremely useful. Basically I wanted to be able to say "derive this value from other host specific values, but never have it go below X (or above Y). I was able to work around it by having some ugly nested if/else stuff. But having a `min()` and `max()` would have made it much simpler.

It basically works just like the python builtins operating only on lists:

``` yaml
- hosts: localhost
  connection: local
  gather_facts: false
  tasks:
    - debug: msg="min 1 - {{ [2,3,1] | min }}"
    - debug: msg="min 3 - {{ [9, 8, 12, 3, 4] | min }}"
    - debug: msg="max 5 - {{ [2,5,1] | max }}"
    - debug: msg="max 12 - {{ [9, 8, 12, 3, 4] | max }}"
```

### DIFF
--- a/docsite/rst/playbooks_variables.rst
+++ b/docsite/rst/playbooks_variables.rst
@@ -180,6 +180,20 @@ Jinja2 provides a useful 'default' filter, that is often a better approach to fa
 In the above example, if the variable 'some_variable' is not defined, the value used will be 5, rather than an error
 being raised.
 
+.. _list_filters:
+
+List Filters
+--------------------
+.. versionadded:: 1.7
+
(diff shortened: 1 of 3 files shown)


mpdehaan:
Manually merged.  Updated this to say added in 1.8 since that's when it was applied.

Much appreciated, thanks!

Issue #7848, opened 2014-06-19
ec2_vpc occasionally fails creation with InvalidVpcID.NotFound


mpeters:
##### Issue Type:

Bug Report
##### Ansible Version:

1.6.2
##### Environment:

Ansible is running on Fedora 20 with AWS API as the target.
##### Summary:

Occasionally calls to ec2_vpc that create a new VPC will fail with a `InvalidVpcID.NotFound` error. This seems to be because the boto `vpc_conn.create_vpc` will return a vpc_id to soon such that a call to `vpc_conn.get_all_vpcs` won't be able to return. So instead of returning a VPC object with out a state of "available" it instead throws a `boto.execption.BotoServerError` which bails out. But if you look at the AWS console a few seconds later you can see that the VPC does indeed exist (minus any tags since those didn't get added before things failed).
##### Steps To Reproduce:

It happens randomly, but if you try to create a new vpc (with a task like the following) you will eventually hit it. You might need to delete the VPC from the AWS console each time it's created if it doesn't fail or (depending on how you created it) the ec2_vpc module won't create a new one, but will instead just update the existing one.

``` yaml
    - name: Create VPC for new environment
      local_action:
        module: ec2_vpc
        state: present
        cidr_block: 10.0.0.0/16
        resource_tags: { "Name": "blah" }
        region: "us-east-1"
        dns_hostnames: true
        dns_support: true
        internet_gateway: true
        validate_certs: true
        instance_tenancy: default
        wait: true
        subnets: []
        route_tables: []
```
##### Expected Results:

It should be able to create a new VPC every time.
##### Actual Results:

It fails randomly without any changes to the task with an error like:

```
failed: [localhost] => {"failed": true}
msg: InvalidVpcID.NotFound: The vpc ID 'vpc-XXXXXXXX' does not exist
```

ghost:
I'm having the exact same issue.

jimi-c:
Closing, as the above fix was merged in. Thanks!

jaytaylor:
FWIW, I also got this error after I'd specified an invalid `vpc_id` in my VPC creation playbook.

PR #7849, opened 2014-06-19
This should fix issue #7848

mpeters:
We need to catch any InvalidVpcID.NotFound errors and treat them
just like getting back a vpc that's not "available"


jimi-c:
Merged, thanks!

PR #7851, opened 2014-06-19
Module to manage RabbitMQ clusters

ghost:
Added a module to join or leave a [RabbitMQ cluster](https://www.rabbitmq.com/clustering.html), based on the `rabbitmq_vhosts` module.


ghost:
Withdrawn. This is better managed through [rabbitmq.config](https://www.rabbitmq.com/configure.html), not the command line.

Issue #7853, opened 2014-06-19
ec2_vpc confuses "id" and "vpc_id"


mpeters:
##### Issue Type:

Bug Report
##### Ansible Version:

1.6.2
##### Environment:

Ansible is running on Fedora 20 targetting the AWS API
##### Summary:

When updating a VPC using the ec2_vpc module the documentation says to use the `vpc_id` parameter to specify which VPC to update. The argument spec also expects the `vpc_id`, but the code actually tries to use a parameter called `id` which can't exist.
##### Steps To Reproduce:

Try to update an existing VPC with something like this (make sure to not have a cidr and resource_tags attached to the VPC directly as it will use those to identify the VPC if no id is provided).

``` yaml
  -  ec2_vpc:
        state: present
        vpc_id: vpc-XXXXXXXX
        region: us-east-1
        wait: true
        subnets:
          - cidr: 10.0.0.0/24
```
##### Expected Results:

It should update the VPC.
##### Actual Results:

You get an error like this:

```
failed: [localhost] => {"failed": true}
msg: You must specify either a vpc id or a cidr block + list of unique tags, aborting
```

jimi-c:
The above patch has been merged in. Thanks!

PR #7854, opened 2014-06-19
fixing issue #7853 by using the right name for the vpc_id parameter

mpeters:
This change uses `vpc_id` name for the parameter instead of the wrong `id`.


jimi-c:
I cherry-picked your commit directly to tweak the commit message, so closing this. Thanks!

PR #7856, opened 2014-06-19
Lineinfile now supports symlinks.

njharman:
One line change (adding os.path.realpath) so the atomic_move of lineinfile does not clobber symlinks.


jimi-c:
Merged, thanks!

PR #7869, opened 2014-06-20
DOCFIX add another Unix passwd creation solution using OpenSSL

scottbrown:
... most people have it installed by default.

It's a simple fix but some people do not want to yak-shave getting pip installed, and mkpasswd doesn't exist in all operating systems.  OpenSSL is typically there by default.

### DIFF
--- a/docsite/rst/faq.rst
+++ b/docsite/rst/faq.rst
@@ -209,21 +209,22 @@ The mkpasswd utility that is available on most Linux systems is a great option::
 
     mkpasswd --method=SHA-512
 
-If this utility is not installed on your system (e.g. you are using OS X) then you can still easily
-generate these passwords using Python. First, ensure that the `Passlib <https://code.google.com/p/passlib/>`_
-password hashing library is installed.
+``mkpasswd`` is often bundled with the ``whois`` package (On Ubuntu, you can do ``sudo apt-get install whois``).  If this utility is not installed on your system (e.g. you are using OS X) then you can still easily
+generate these passwords using Python standard library.  The only requirement is that this is run on the same operating where the encrypted password will be stored (this is due to the way the ``crypt`` implementation is system-dependent).
 
-    pip install passlib
+SHA512 password values can be generated with Python as follows::
 
-Once the library is ready, SHA512 password values can then be generated as follows::
+    python -c 'import crypt; print crypt.crypt("<password>", "$6$<salt>")'
 
-    python -c "from passlib.hash import sha512_crypt; print sha512_crypt.encrypt('<password>')"
+where ``<password>`` is the password to generate a hash, and ``<salt>`` is a set of random characters provided by you.
 
-If you have ``openssl`` installed on your machine, you can generate a password with the following command::
+If you wish Python to automatically generate a randomized salt for you, use this slightly longer one-liner::
 
-    openssl passwd -1 -salt <salt> <password>
+    python -c 'import crypt,random,string; print crypt.crypt("<password>", "$6$" + "".join([ random.choice(string.digits + string.letters + string.punctuation) for x in range(12)]))'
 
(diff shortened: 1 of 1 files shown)


quiver:
> openssl passwd -1 -salt <salt> <password>

This command  hashes passwords with MD5 while existing docs use SHA512  hashing.
Most modern Linux distros no longer use MD5 for password hashing for security reasons.

Following is the output of (recent)Amazon Linux instance:

```
$ sudo authconfig --test | grep password
 shadow passwords are enabled
 password hashing algorithm is sha512
```

At the moment, `open-ssl  passwd`does not provide ways to compute password hashes w/  SHA-variant, so I'm -1 for this Pull Request.

quiver:
> It's a simple fix but some people do not want to yak-shave getting pip installed, and mkpasswd doesn't exist in all operating systems. OpenSSL is typically there by default.

After some googling, I discovered that Python standard library itselft  can generate SHA512 based password hashing(although you need to take care of salt). 

```
$ python -c 'import crypt; print crypt.crypt("test", "$6$random_salt")'
```

http://serverfault.com/a/330072

So we can rewrite "How do I generate crypted passwords for the user module?" section. No `pip install passlib` is needed, yay!

bcoca:
This little script actually creates a nice random salt, you can also create
random passwords the same way

import crypt
import random
import string

random_list = [ random.choice(string.digits + string.letters +
string.punctuation) for x in range(12)]
print crypt.crypt("mysecretpassword", "$6$" + ''.join(random_list))

​

scottbrown:
Good points @quiver.  Ansible (and everyone) should be promoting secure systems, so I shouldn't be showing outdated password techniques.

I'll rewrite it to use the python stdlib version.  I may also add a note about the md5 technique I used, saying that it is outdated and is not recommended, in the hopes that it prevents some future security mistakes by others.

@bcoca Thank you for the nice, readable script.  I'm reluctant to use a script since this should really be a one-liner for someone.  If there's a good way to incorporate the script into the wording of the docs, let me know.  For now, I'm +1 to leave it out mainly because it requires the user to know how to write/save/execute a python script, increasing the cognitive load of creating a Unix shadow password.  One-liners are a copy/paste activity.

bcoca:
one liner version:

python -c "import crypt,random,string; print
crypt.crypt('mypassword', "$6$" + ''.join([ random.choice(string.digits +
string.letters + string.punctuation) for x in range(12)])"​

scottbrown:
heh.  I guess I asked for that one.  ;)  Thanks @bcoca, I'll look at adding it in.

sneakyice69:
hello there Mr Scott.  .. I believe you are in charge of my system....  so why do you get to moniter me..   and I can do a system moniter with -htop.... just wondering. .... 

Sent from my T-Mobile 4G LTE Device

-------- Original message --------

scottbrown:
Alrighty... I have rewritten the entire section.  Please review.

@quiver I have included your stdlib one-liner.
@bcoca I have included your longer one-liner in case we want a automatic randomized salt.

quiver:
> The mkpasswd utility that is available on most Linux systems is a great option:
> mkpasswd --method=SHA-512

This is not your fault, but there are two kinds of `mkpasswd` in linux
1. debian mkpasswd -> front end to `crypt(3)` as in this FAQ. part of `whois` package
2. redhat mkpasswd -> tool  to generate random password. part of `expect` package

So "most Linux systems" would be "most Debian like Linux systems" or something.
So is  "(e.g. you are using OS X) " phrase.

scottbrown:
Done and done.  Thanks @quiver

quiver:
:+1:

mpdehaan:
Sorry about the delay in response on this topic.

I'm uncomfortable with the level of complexity here and the phrase "Debian-like" which doesn't cover RHEL/CentOS/Fedora solutions.   How about stopping by ansible-project to discuss?

mpdehaan:
(This may also be helpful given the long comment change and revisions on this thread)

PR #7815, opened 2014-06-17
Fix/improve speed of filtering long non-JSON module output.

gdvalle:
I'm not sure how expensive/definitive the check on key=value existence needs to be.  I've reordered the conditionals to check the k/v condition last since it is the most expensive.  Also switched to using a cheaper substring search instead of regex.  With this change, output could (still) be invalid key=value pairs.

Larger transfers using the copy module would peg the CPU and slow to a crawl if an "=" existed toward the end of the base64 data in "content" key.  test_ansible_slow_copy.sh will reproduce this behavior.

Script to reproduce:
https://gist.github.com/GDvalle/80c838757e2b48b88334

Script to test speed of filter_leading_non_json_lines (place in ansible.git/lib/ansible/utils):
https://gist.github.com/GDvalle/bee94926983b65fa5508

________Before fix________
$ ./test_ansible_slow_copy.sh
System info:
Linux wksgreg 3.11.0-22-generic #38-Ubuntu SMP Thu May 15 20:47:00 UTC 2014 x86_64 x86_64 x86_64 GNU/Linux
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 13.10
Release:        13.10
Codename:       saucy
OpenSSH_6.2p2 Ubuntu-6ubuntu0.4, OpenSSL 1.0.1e 11 Feb 2013
ansible 1.7 (fix_copy_module_slow_regex 2c93410e7f) last updated 2014/06/17 15:01:59 (GMT -500)
ansible-playbook 1.7 (fix_copy_module_slow_regex 2c93410e7f) last updated 2014/06/17 15:01:59 (GMT -500)
Python 2.7.5+

60000+0 records in
60000+0 records out
60000 bytes (60 kB) copied, 0.0881104 s, 681 kB/s
Generated /tmp/testfile_src
First run
elasticsearch_logger disabled.  Set env var ANSIBLE_ES_SERVER to enable.

PLAY [localhost] *************************************************************\* 

TASK: [copy src=/tmp/testfile_src dest=/tmp/testfile_dest] *******************\* 

changed: [localhost] => {"changed": true, "dest": "/tmp/testfile_dest", "gid": 1997, "group": "greg", "md5sum": "58a5fddb03502742c03bb2554ed4d368", "mode": "0644", "owner": "greg", "size": 81053, "src": "/home/greg/.ansible/tmp/ansible-tmp-1403036853.88-229266854979868/source", "state": "file", "uid": 1997}                                  

[...]

PR #7871, opened 2014-06-20
Fixed inefficient regex.

lonerr:


jimi-c:
Your new regex has an error in it, you're missing a `\` before the first `w`.

mpdehaan:
This appears to be a bugfix, not a feature request, so changing tags @jimi-c 

can you please share instructions how to get into this scenario where it becomes a problem?

Thanks!

lonerr:
@jimi-c you are right, sorry for my carelessness :(
What i suppose to do? Close this pull request and create the new one?

lonerr:
@mpdehaan , my real case:
Using diff mode with 'template' module on file /etc/services (~2500 lines, ~87KB) ansible-playbook performed about 40 seconds on my laptop. After this patch applied - about 1 second.

Simple test case for regex:
[lonerr@neon ~]% time python -c "import re; re.match(r'._\w+=\w+._',''.join('x' for x in xrange(65536)))"
real    0m23.890s
user    0m23.798s
sys 0m0.043s

[lonerr@neon ~]% time python -c "import re; re.search(r'\w=\w',''.join('x' for x in xrange(65536)))"
real    0m0.039s
user    0m0.027s
sys 0m0.010s

jimi-c:
@lonerr you can just update the patch with `git commit --amend` and then re-push the branch with `--force`. Github will then update the PR in place and I can merge it in (after testing of course).

lonerr:
Thanks @jimi-c ! Done.

jimi-c:
Now that I look at this, there are a additional problems. Using `\w=\w` will probably catch any = in the middle of a string, and isn't much better than doing `if '=' in ...`. I've tested the following out and it seems to be just as quick while properly identifying things that will most likely be a k=v pair:

<pre>
$ time python -c "import re; test='A'*65536+' foo=bar '+'A'*65536; print len(test); print re.match(r'^.*\b(\w+=\w+)\b', test).groups()"131081
('foo=bar',)
real    0m0.011s
user    0m0.010s
sys 0m0.000s
$ time python -c "import re; test='A'*65536+' foo=bar'; print len(test); print re.match(r'^.*\b(\w+=\w+)\b', test).groups()"
65544
('foo=bar',)
real    0m0.009s
user    0m0.005s
sys 0m0.003s
$ time python -c "import re; test='foo=bar '+'A'*65536; print len(test); print re.match(r'^.*\b(\w+=\w+)\b', test).groups()"
65544
('foo=bar',)
real    0m0.011s
user    0m0.007s
sys 0m0.003s
</pre>

So go ahead and update your PR with what I have above: `r'^.*\b\w+=\w+\b'` and we can test that.

lonerr:
Hmm, @jimi-c , could you provide an example of string that mistakenly matched with r'\w=\w' and not matched with 'r'^.*\b\w+=\w+\b'  ?

jimi-c:
As it stands, `\w=\w` won't match anything except for a simple string like `a=b`. For example:

<pre>
python -c "import re; test='foo=bar'; print len(test); print re.match(r'(\w=\w)', test).groups()"
7
  File "<string>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'groups'
</pre>

So you need to use the `+` to capture more characters, however that's what was making it slow, since it had to traverse the whole string. Anchoring it to the start of the line as well as requiring the word boundary around it seems to help with that though as shown by my tests above.

lonerr:
Hmm, do you feel the difference between re.match and re.search?

[lonerr@neon ~]% python -c "import re; test='foo=bar'; print len(test); print re.search(r'(\w=\w)', test).groups()"
7
('o=b',)

I will clarify my question: could you provide an example of string that mistakenly matched with re.search(r'\w=\w') and not matched with re.match('r'^.*\b\w+=\w+\b') ?

lonerr:
Hi, @jimi-c !
I'll be glad to hear your answer to the last question.

jimi-c:
@lonerr sorry, been a busy week :)

Hrm, I wasn't aware that .search() would treat that differently than .match(), I guess it implies a word boundary for a match? In that case, the original is fine when using just search().

lonerr:
re.match implies '^' at line beginning, in other words re.match('foo') is roughly equivalent to re.search('^foo').
And re.search('\w=\w') is "minimum one alphanumeric character followed by equivalence sign and minimum one alphanumeric character in any position of line".
In my opinion it is a necessary and sufficient condition.

amenonsen:
This PR can be closed, there's no regex in _filter_leading_non_json_lines any more.

jimi-c:
@amenonsen I believe we need to add it back in, so I'll leave this open for now.

amenonsen:
OK, in that case, @lonerr, do you feel like revising this PR to add the regex test to devel?

amenonsen:
Note that the function has moved to `lib/ansible/plugins/action/__init__.py` now; see also PR #7815.

lonerr:
This pull request can be closed since the refactored version of filter_leading_non_json_lines() has no more regex at all.

PR #7872, opened 2014-06-20
New Jinja2 filters: hash_merge & hash_replace

darkk:
Link to intro_configuration.html#hash-behaviour is coded without :doc: prefix as I've not managed to create proper link with anchor for sphinx.

This code was inspired by PR #7629

### DIFF
--- a/docsite/rst/playbooks_variables.rst
+++ b/docsite/rst/playbooks_variables.rst
@@ -260,6 +260,33 @@ Get a random number from 1 to 100 but in steps of 10::
     {{ 100 |random(1, 10) }}    => 31
     {{ 100 |random(start=1, step=10) }}    => 51
 
+.. _hash_filters:
+
+Hash Merge Filters
+------------------
+
+.. versionadded:: 1.7
+
+These two filters allow fine-grained control over `hash_behaviour <intro_configuration.html#hash-behaviour>`_ at template level.
+``hash_merge`` is deep version that mimics ``hash_behaviour=merge`` and
+``hash_replace`` is shallow version that mimics ``hash_behaviour=replace``.
+Both filters do not modify existing dictionary and produce new one::
+
+    {{ dict_one | hash_merge(dict_two, dict_three) }}
+    {{ dict_one | hash_replace(dict_two, dict_so_on) }}
+
+These filters can be especially useful while constructing `complex-args <https://github.com/ansible/ansible-examples/blob/master/language_features/complex_args.yml>`_ to pass them to module. Attempt to ``hash_merge`` two hashes with different schema may lead to template rendering failure::
+
+    vars:
+      fatality: { foo: { bar: 42 } }
+      brutality: { foo: }
+    tasks:
+      - name: Replace wins!
+        debug:
+          msg: '{{ fatality | hash_replace(brutality) }}'
+      - name: NoneType fatality!
+        debug:
+          msg: '{{ fatality | hash_merge(brutality) }}'
 
 .. _other_useful_filters:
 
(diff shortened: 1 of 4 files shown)


mpdehaan:
Thanks.

I'd prefer to see this documented without reference to the ansible.cfg setting at all, but still need to think about these.

mpdehaan:
I'd also like to see the examples to be a bit more real world rather than the Mortal Kombat references, so it's clear when they might be used.

jirutka:
I like Mortal Kombat references! :smiley:

darkk:
@mpdehaan I referenced settings value as it's sort of "compatibility" filter that inherits all bugs of dict merging implementation. I'm unsure what is better: fix bugs in merge_hash that leads to NoneType dereference or leave warning sign about possible `hash_merge` hazard - that's what mortal example is about.

I try to use (abuse?) ansible as scripting language to generate monitoring system configuration as ansible is already used as deploy tool and having single DSL to learn is a bit better than two different alike DSLs. I use dict merging to provide different sets of default values for different checks and nested structures are used for syntax sugar when

```
http:
  path: /ping
  disable_ipv6: yes
```

is used as shortcut for

```
active: http
active_args:
  path: /ping
  disable_ipv6: yes
```

In this case `hash_merge` is used to provide cluster-wide variable that controls `disable_ipv6` setting and leaves `/path` service-specific. I'm unsure if this real-world example is useful as juggler (monitoring system I am speaking about) is not released yet.

I dive into ansible-project@ mailing list a bit and found several requests for different `merge` behavior, but none of them used complex-args as an example… So the point of this PR is to move global hash_merge settings to template and make it explicit.

jhriggs:
+1 to this functionality. My use case is slightly different in that I just need a way to combine/merge dictionaries. I would like to see these filters accept keyword arguments too, though, as a simpler way of passing in values. My use case:

``` yaml
vars:
  default_ec2_tags:
    environment: 'dev'
    application: 'doSomething'

tasks:
  - ec2_tag:
      resource: 'vpc-1234567'
      tags: '{{ default_ec2_tags | hash_replace(Name="MyVPC") }}'
```

I think that the keyword syntax `hash_replace(Name="MyVPC")` is simpler than `hash_replace({Name: "MyVPC"})` in this case. It may just be my personal preference, but at least it can be possible with a quick change (the keyword args dict is just added as an item to the terms tuple):

``` diff
--- hash_merge.py   2014-09-17 09:46:10.000000000 -0500
+++ library/plugins/filter_plugins/hash_merge.py    2014-09-17 09:31:19.000000000 -0500
@@ -22,19 +22,19 @@

 # These are two possible behaviours expressed by `ansible.utils.combine_vars`.

-def hash_merge(*terms):
+def hash_merge(*terms, **kwterms):
     for t in terms:
         if not isinstance(t, dict):
             raise errors.AnsibleFilterError("|hash_merge expects dictionaries, got " + repr(t))

-    return reduce(utils.merge_hash, terms)
+    return reduce(utils.merge_hash, terms + (kwterms,))

-def hash_replace(*terms):
+def hash_replace(*terms, **kwterms):
     for t in terms:
         if not isinstance(t, dict):
             raise errors.AnsibleFilterError("|hash_replace expects dictionaries, got " + repr(t))

-    return dict(itertools.chain(*map(dict.iteritems, terms)))
+    return dict(itertools.chain(*map(dict.iteritems, terms + (kwterms,))))

 class FilterModule(object):
     ''' Ansible jinja2 filter to avoid global hash_behaviour=merge when it's needed '''
```

gimoh:
+1

My use case is wrapping a builtin module (`docker`) to provide extra optional features, and I need to be able to inject settings into the parameters for the `docker` module.  See [the role](https://galaxy.ansible.com/list#/roles/3101) for [details](https://github.com/gimoh/ansible_docker_container/blob/master/tasks/main.yml#L26).

amenonsen:
I could also use the hash_merge and hash_replace functionality, in fact I have a custom filter that does the former and would be happy to get rid of it. @darkk any chance you could rebase this to apply to current devel?

tersmitten:
Nice

PR #7877, opened 2014-06-21
Give option on how git tracks submodules.

### LINKED ISSUE #7030: Git submodule update only tracks branches
##### Issue Type:

“Bug Report” /  “Feature Idea”
##### Ansible Version:

ansible 1.6 (devel a5e7492c4f) last updated 2014/04/15 21:12:13 (GMT -400)
##### Environment:

Ubuntu 12.04
##### Summary:

Git submodule update is hard coded to track branches (always adds the `--remote` flag, assuming the git version is late enough to support tracking branches.

https://github.com/ansible/ansible/blob/1c9950678a8a226020c6a1a32d6a60f942495072/library/source_control/git#L423-424

I have a use case where we definitely want to track the sha1 recorded in the superproject, not a branch or tag of the submodule. I'm willing to put in a PR on this, but wonder about the best way to add in a choice on this. A submodule_remote parameter? Something else?
##### Steps To Reproduce:

In a git repo with submodules that are tracked by commit hash stored in the superproject rather than branch or tag, run 

`git: repo=reponame dest=place/to/put/it`
##### Expected Results:

I'd expect all of my submodule repos to be checkout out on the commit stored in the the superproject.
##### Actual Results:

All of my submodule repos are be checkout out to the commit at the tip of the master branch.

If the submodules also do not use the label "master" for the master branch, it throws an error

`fatal: Needed a single revision`

`Unable to find current origin/master revision in submodule path`

bibliotechy:
Fixes https://github.com/ansible/ansible/issues/7030

Allows user to decide if git submodule should track branches/tags or track commit hashes defined in the superproject.

Add track_branches parameter to the git module.

Defaults to track branches behavior.

### DIFF
--- a/library/source_control/git
+++ b/library/source_control/git
@@ -128,6 +128,16 @@ options:
         description:
             - if C(no), repository will be cloned without the --recursive
               option, skipping sub-modules.
+
+    track_branches:
+        required: false
+        default: "yes"
+        choices: ["yes", "no"]
+        version_added: "1.7"
+        description:
+            - if C(no), repository will be cloned without the --recursive
+              option, allowing submodules to be tracked by commit hash
+              instead of branch name.
 notes:
     - "If the task seems to be hanging, first verify remote host is in C(known_hosts).
       SSH will prompt user to authorize the first contact with a remote host.  To avoid this prompt, 
@@ -399,7 +409,7 @@ def get_head_branch(git_path, module, dest, remote, bare=False):
     f.close()
     return branch
 
-def fetch(git_path, module, repo, dest, version, remote, bare):
+def fetch(git_path, module, repo, dest, version, remote, bare, track_branches):
     ''' updates repo from remote sources '''
     (rc, out0, err0) = module.run_command([git_path, 'remote', 'set-url', remote, repo], cwd=dest)
     if rc != 0:
@@ -417,10 +427,10 @@ def fetch(git_path, module, repo, dest, version, remote, bare):
         (rc, out2, err2) = module.run_command("%s fetch --tags %s" % (git_path, remote), cwd=dest)
     if rc != 0:
         module.fail_json(msg="Failed to download remote objects and refs")
-    (rc, out3, err3) = submodule_update(git_path, module, dest)
+    (rc, out3, err3) = submodule_update(git_path, module, dest, track_branches )
     return (rc, out1 + out2 + out3, err1 + err2 + err3)
 
-def submodule_update(git_path, module, dest):
+def submodule_update(git_path, module, dest, track_branches):
     ''' init and update any submodules '''
 
     # get the valid submodule params
@@ -431,7 +441,7 @@ def submodule_update(git_path, module, dest):
         return (0, '', '')
     cmd = [ git_path, 'submodule', 'sync' ]
     (rc, out, err) = module.run_command(cmd, check_rc=True, cwd=dest)
-    if 'remote' in params:
+    if 'remote' in params and track_branches:
         cmd = [ git_path, 'submodule', 'update', '--init', '--recursive' ,'--remote' ]
     else:
         cmd = [ git_path, 'submodule', 'update', '--init', '--recursive' ]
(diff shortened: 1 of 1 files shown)


wizonesolutions:
+1 based on my limited understanding of the issue here and how to solve it.

bugweis:
+1
Landed here when trying to resolve in fact the same issue. When git submodule is called with `force=no` option I ended up with following error with current behaviour:

```
failed: [devel] => {"failed": true}
msg: Local modifications exist in repository (force=no).
```

mpdehaan:
This option seems really confusing to me.   We already have the recursive option, is it not possible to just make it do the right thing?   If not, can this option and documentation be named/written differently?

bugweis:
It's hard to decide what _"just make it do the right thing"_ means. Current option "recursive" tracks the submodule branches so the most up-to-date version is always cloned. This leads to less control of what version of git submodule is actually being checked out.
Default submodule policy for `git clone --recursive` is to fetch version that is recorded in the "super-repository". So if the submodule contains new commits they are not included which is _"safer"_. This would be possible if this commit was merged.

christophebiocca:
+1 We're running into problems with this unexpected `--remote` argument as well. We're just executing a command to properly reset the submodules, but it would be better not to break them in the first place.

@mpdehaan: I'd say that the "right thing" is to not include `--remote` by default, but that's been shipped already and people may be relying on that being the default behaviour.

bibliotechy:
Sorry about the long delay in response. Been away from reliable internet for a few weeks. 

So, my description in the documentation was...just wrong. I'm certain that added to the confusion. I read through it and thought...WTF? So, yeah, sorry. :(

I've added another commit that _correctly_ explains the effect of the option and updates the name of the option to `track_submodule_branches` which I hope makes it clearer what this option does.

library/source_control/git
```diff
@@ -431,7 +441,7 @@ def submodule_update(git_path, module, dest):
         return (0, '', '')
     cmd = [ git_path, 'submodule', 'sync' ]
     (rc, out, err) = module.run_command(cmd, check_rc=True, cwd=dest)
-    if 'remote' in params:
+    if 'remote' in params and track_submodule_branches:
```
josef-spak, line 444: I understand the pull request author was trying not to break existing behavior.

In my opinion, the default 'right' thing for the recursive option is to do what git does by default, which is **not to use --remote** and (quoting from the manual here) **use the superproject’s recorded SHA-1 to update the submodule**:

```
--remote
           This option is only valid for the update command. Instead of using the superproject’s recorded SHA-1 to update the submodule, use the status of the submodule’s remote-tracking branch. The remote used is branch’s remote
           (branch.<name>.remote), defaulting to origin. The remote branch used defaults to master, but the branch name may be overridden by setting the submodule.<name>.branch option in either .gitmodules or .git/config (with
           .git/config taking precedence).

           This works for any of the supported update procedures (--checkout, --rebase, etc.). The only change is the source of the target SHA-1. For example, submodule update --remote --merge will merge upstream submodule changes
           into the submodules, while submodule update --merge will merge superproject gitlink changes into the submodules.

           In order to ensure a current tracking branch state, update --remote fetches the submodule’s remote repository before calculating the SHA-1. If you don’t want to fetch, you should use submodule update --remote --no-fetch.
```

What should rather be explained (and documented) is why Ansible's git module uses --remote whenever it is detected to be available, in get_submodule_update_params()

mpdehaan:
Hi!

Thanks you very much!

pztrick:
This commit worked wonderfully for me -- at least after I added the `github.com` public key.

@bibliotechy do you have the free time to make this pull request against `ansible-modules-core`? See this gist: [https://gist.github.com/willthames/afbaaab0c9681ed45619](https://gist.github.com/willthames/afbaaab0c9681ed45619)

:+1:

thanks!

bibliotechy:
@pztrick https://github.com/ansible/ansible-modules-core/pull/55

library/source_control/git
mrsfixit, line 444: Sorry, but this sounds like 5 monkeys and a ladder.
(http://www.wisdompills.com/2014/05/28/the-famous-social-experiment-5-monkeys-a-ladder/)

Yer its a breaking change and should, by semantic versioning, require a version bump, but if this is broken, isn't it best just to fix it, so it makes more sense for future new comers to ansible?

I'd hazard a guess that more people are being tripped up by this (as they're most likely new and inexperienced with ansible), than the number of people that would get tripped up by fixing it.
- A. because existing users would be more experienced in ansible
- B. because it can be a well documented breaking change that existing users knowingly bring upon themselves by upgrading.
- C. this is the submodule behaviour that happens on eg. travis

PR #7886, opened 2014-06-22
Bugfix and better error propagation for win_service module

chrishoffman:
Fixing a bug with the stopped status always displaying changed and adding better error propagation.


mpdehaan:
Merged, thanks!

PR #7893, opened 2014-06-23
Adding support for spot instances in ansible_lc.

fdv:
AWS allows to define launch configuration with spot instances. This is also natively supported by boto (see http://boto.readthedocs.org/en/latest/ref/autoscale.html)


jimi-c:
Merged, thanks!

Issue #7898, opened 2014-06-23
health_check_period being set to 0 by default makes ASG instances being terminated immediately


fdv:
##### Issue Type:

Bug report
##### Ansible Version:

`ansible 1.7` and `ansible 1.6`
##### Environment:

N/A
##### Summary:

When creating an autoscaling group using the `ec2_asg` module, it is impossible to specify `health_check_period`, which is then set to 0.

According to Boto documentation, `health_check_period` refers to:

> Length of time in seconds after a new EC2 instance comes into service that Auto Scaling starts checking its health.

When set to 0, AWS health check will actually wait 0 seconds, which is way too fast for the instance to launch. 
##### Steps To Reproduce:

Just create an ASG and try to have the instance survive for more than 5 minutes.
##### Expected Results:

Instance should not spawn and be killed immmediately.
##### Actual Results:

Every launched instance within the autoscaling group is immediately killed after boot.

jimi-c:
The above PR has been merged in, which should address this issue. Please let us know if you continue seeing any problems related to this. Thanks!

PR #7899, opened 2014-06-23
Adds support for `health_check_period` and `health_check_type` in `ec2_a...

fdv:
...sg` module.

Default is set to 300 seconds, which is AWS default.

This PR fixes a bug (#7898) where instances created within an autoscaling group using the `ec2_asg` module gets immediately terminated because the `health_check_period` is set to 0, which causes the instance to be checked without having the time to actually boot.

Adding `health_check_type` is needed because you may want to check your instance health against an ELB instead of just EC2 default cloudwatch.

### DIFF
--- a/library/cloud/ec2_asg
+++ b/library/cloud/ec2_asg
@@ -73,6 +73,18 @@ options:
     required: false
     default: None
     version_added: "1.7"
+  health_check_period:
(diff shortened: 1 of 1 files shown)


jimi-c:
Merged, thanks!

PR #7904, opened 2014-06-23
force evaluation of vars memory_mb, num_cpus to be int fixes #7903

### LINKED ISSUE #7903: vsphere_guest: values are being evaluated as strings, not int
##### Issue Type:

Bug Report 
##### Ansible Version:

ansible 1.7 (devel 65a8ffef6e) last updated 2014/06/23 08:42:27 (GMT -400)
##### Environment:

OSX 10.9 / RHEL 6.5
##### Summary:

when specifying variables for memory, num_cpus in vsphere_guest, they are being evaluated as strings, and not integers
##### Steps To Reproduce:

specify memory_mb, num_cpus in vsphere_guest:
    vm_hardware:
      memory_mb: "{{ memory_mb }}"
      num_cpus: "{{ num_cpus }}

try reducing the amount of memory on a VM, the if statemets to see if the playbook value is less than the vm value don't evaluate to true, thus the configuration doesn't get changed
##### Expected Results:

the configuration would get changed 
##### Actual Results:

the module fails

ghost:
change if statements to evaluate memory_mb and num_cpus as an int.

### DIFF
--- a/library/cloud/vsphere_guest
+++ b/library/cloud/vsphere_guest
@@ -496,7 +496,7 @@ def reconfigure_vm(vsphere_client, vm, module, esxi, resource_pool, cluster_name
     # Change Memory
     if vm_hardware['memory_mb']:
 
-        if vm_hardware['memory_mb'] != vm.properties.config.hardware.memoryMB:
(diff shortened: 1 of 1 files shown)


jimi-c:
Merged, thanks!

PR #7915, opened 2014-06-24
Add examples for `shell` and `command` modules.

gitaarik:
I added examples that make use of the `args` form (https://github.com/ansible/ansible-examples/blob/master/language_features/complex_args.yml). This allows people to make their `shell` or `command` actions more clear if they use the options.

I didn't know about this language feature so for me it was quite a search before I found it out. It would be nice to add it to the examples of these modules because you would probably use this language feature faster with these modules than others.

### DIFF
--- a/library/commands/command
+++ b/library/commands/command
@@ -77,11 +77,19 @@ author: Michael DeHaan
 '''
 
 EXAMPLES = '''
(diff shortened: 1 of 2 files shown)


jimi-c:
Merged, thanks!

PR #7921, opened 2014-06-24
azure - added storage account module

matlockx:
Added an windows azure storage module + inventory command to list storages.
A storage account is needed for the normal azure module to create a vm. The name of the storage account must therefore be provided in the azure module. This module can be combined with the azure module to create at first a storage account and then create a vm.

### DIFF
--- a/lib/ansible/module_utils/azure.py
+++ b/lib/ansible/module_utils/azure.py
@@ -1,0 +1,99 @@
+# This code is part of Ansible, but is an independent component.
+# This particular file snippet, and this file snippet only, is BSD licensed.
+# Modules you write using this snippet, which is embedded dynamically by Ansible
+# still belong to the author of the module, and may assign their own license
+# to the complete work.
+#
+# Copyright (c), Michael DeHaan <michael.dehaan@gmail.com>, 2012-2013
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification,
+# are permitted provided that the following conditions are met:
+#
+#    * Redistributions of source code must retain the above copyright
+#      notice, this list of conditions and the following disclaimer.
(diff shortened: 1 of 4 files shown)


jimi-c:
Thanks, I did notice the lack of that when testing the original azure module. We'll get this into testing as soon as possible.

mpdehaan:
"--list-storages" should be "list storage", as storages is not a word.

Also it appears there are options in the inventory script already that ansible can't use, like --list-images that probably do not belong.

If these are returned as group_vars variables, then they should be returned by default, as ansible can't read them otherwise as it will never pass those flags.

mpdehaan:
In all, this is a relatively large patch that also makes some changes like removing the LOCATIONS array, and I'd be interested in why.

Rather than take this one wholesale, I would ask that you break this up into smaller commits with individual commit messages explaining each step.   

Changes to the azure module should probably be in a seperate pull request than the inventory module, as well as the change to add the storage module.

The name "azure_s" for the storage module is also not sufficient, I'd expect "azure_storage", and similarly, in docs, Azure should be capitalized.   "deletes a storage" also doesn't quite seem complete, does it mean an account, a volume, etc?

matlockx:
Location array is not gone I moved it to module utils like the ec2 approach because it is needed in more than one module. 
Azure_s is A typo. Don't know how I missed that since everything else is called azure_storage a you proposed. 
So based on your comment I will:
Split it up into the parts  azure_storage module, azure module and the inventory. 
Luckily I also have the good news that the guys from the azure SDK made a new release which fixes a critical http redirect problem so I will add that version dependency to the azure module too.

matlockx:
New pull request based on mentioned improvements created: #7966

Issue #7691, opened 2014-06-07
FreeBSD ports fail to install


francisco1844:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.7 (devel aaa16acdf1) last updated 2014/06/07 16:34:23 (GMT +000)
##### Environment:

Running ansible from ubuntu and managing FreeBSD 10 machine.
##### Summary:

Installing FreeBSD ports causes the program to crash.
So far has failed while trying to install nginx and apache 2.4
##### Steps To Reproduce:

portinstall: name=www/nginx
##### Expected Results:

Port should be installed.
##### Actual Results:

failed: [t1] => {"failed": true, "parsed": false}
invalid output was: Traceback (most recent call last):
  File "/root/.ansible/tmp/ansible-tmp-1402158888.34-20696508689981/portinstall", line 1425, in <module>
    main()
  File "/root/.ansible/tmp/ansible-tmp-1402158888.34-20696508689981/portinstall", line 200, in main
    install_packages(module, pkgs, p["use_packages"])
  File "/root/.ansible/tmp/ansible-tmp-1402158888.34-20696508689981/portinstall", line 170, in install_packages
    matches = matching_packages(module, package)
  File "/root/.ansible/tmp/ansible-tmp-1402158888.34-20696508689981/portinstall", line 104, in matching_packages
    occurrences = int(parts[0])
ValueError: invalid literal for int() with base 10: 'www/nginx'

FATAL: all hosts have already failed -- aborting

francisco1844:
Issue confirmed now with a third port.. postfix.
So far every port I have tried to install has failed: apache 2.4, nginx and postfix

Installation from pkg works though.

weirdbricks:
I believe I am running into the same issue on Ansible 1.6.3 - in my case I am running it on FreeBSD 10 and managing another FreeBSD 10 host

**Command to repro:**
 ansible all -m portinstall -a "name=sysutils/moosefs-chunkserver state=present"

**Error:**
192.168.2.16 | FAILED >> {
    "failed": true,
[...]

PR #7929, opened 2014-06-25
Fixed issue with portinstall on FreeBSD

sidred:
Piping shell commands needs use_unsafe_shell set to True. Fixes issue #7691

### DIFF
--- a/library/packaging/portinstall
+++ b/library/packaging/portinstall
@@ -99,7 +99,7 @@ def query_package(module, name):
 def matching_packages(module, name):
 
     ports_glob_path = module.get_bin_path('ports_glob', True)
-    rc, out, err = module.run_command("%s %s | wc" % (ports_glob_path, name))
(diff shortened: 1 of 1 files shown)


jimi-c:
We can't enable the use_unsafe_shell anywhere where user input is executed as a shell command, for security reasons. It looks like the correct approach here would be to remove the `| wc` and parse the output in python directly. If you'd like to rework this patch to do that, we can get it merged in. Thanks!

sidred:
Can I pipe the input using pipes.quote as per guidelines in module_security
https://github.com/ansible/ansible/blob/0cea4d8a32ba815152b0e8974e3c101b3a840099/CODING_GUIDELINES.md#module-security
?

Thanks,
Sid

jimi-c:
That may work, though we prefer to avoid the use of unsafe shell whenever possible, so really I think it would be better to execute it as I described above. The piping through wc seems to be a simple thing to replicate in pure python on the stdout from the command.

sidred:
Sure. I submitted #7940 to resolve it using string count in python.

PR #7931, opened 2014-06-25
In yum module example, use present/absent instead of installed/removed

quiver:
In yum module code, `installed`/`removed` is defined as aliases for `present`/`absent`. But both of `installed` and `removed` are not listed as choices for `state` parameter in the documentation.


mpdehaan:
Merged, thanks.  Will appear in next docs push.

PR #7934, opened 2014-06-25
Fix missing code block in playbook variables doc

chrishepner:
A code sample was not being rendered as a code block. This adds the preceding '::' reStructuredText markup.


mpdehaan:
Merged, thanks!  Will appear in next docs push.

PR #7940, opened 2014-06-26
Fix FreeBSD ports install issue

sidred:
Fixes the port install issue (#7691). Current method uses shell pipe and wc to count the number of packages found. This errors out because piping requires use_unsafe_shell set to true.

Changed to use python's string count to count for newline '\n'.

Tested on FreeBSD 10.0-Release.


jimi-c:
Looks good, but you still have `, use_unsafe_shell=True` in the run_command call. Just remove that I can get this merged in.

PR #7945, opened 2014-06-26
Add soft requirement of python-hashlib

mrbanzai:
Note that python-hashlib is required on Python < 2.5 for certain modules

### DIFF
--- a/docsite/rst/intro_installation.rst
+++ b/docsite/rst/intro_installation.rst
@@ -58,6 +58,7 @@ Managed Node Requirements
 On the managed nodes, you only need Python 2.4 or later, but if you are are running less than Python 2.5 on the remotes, you will also need:
 
 * ``python-simplejson`` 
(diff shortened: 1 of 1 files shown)


mrbanzai:
I'll note that perhaps it isn't desirable to expand requirements for Python < 2.5 on managed nodes, but it would take some work to remove that requirement from some of the modules.  I'm having to pre-install this package on all of my managed nodes running Python 2.4 to do git checkouts, for example (due to its reliance on the known_hosts module).

mpdehaan:
Lots of modules require dependencies in different places.   I'm thinking this would be better as a note in the particular modules, such as adding it to the dependency list "python-hashlib (if Python < 2.5)" or somesuch, since we have a system for this already in the automated docs generator.

Thanks!

mrbanzai:
Sounds good, I'll visit that instead as I come across those (as I seem to be one of very few users running Python < 2.5 on managed nodes :) ... oh, XenServer ...).

PR #7949, opened 2014-06-27
Check at most 16 characters of the key. Fixes #7936

### LINKED ISSUE #7936: problem when using apt_key module with keyserver
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.6.3
##### Environment:

Ubuntu 14.04
##### Summary:

Trying to use apt_key module to get the equivalent of `sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 36A1D7869245C8950F966E92D8576A8BA88D21E9` fails with an error.

I've put what goes into recv-keys above into the id attribute, because there is no recv-keys attribute of the module. Should there be?
##### Steps To Reproduce:

Add key for Docker repository

```
- name: Add Docker key
  sudo: true
  apt_key: keyserver=hkp://keyserver.ubuntu.com:80 id=36A1D7869245C8950F966E92D8576A8BA88D21E9
```
##### Expected Results:

No error message in Ansible run and apt key for Docker has been added
##### Actual Results:

```
failed: [localhost] => {"failed": true, "id": "36A1D7869245C8950F966E92D8576A8BA88D21E9", "item": ""}
msg: key does not seem to have been added

```

hacosta:
Signed-off-by: Hector Acosta hector.acosta@gmail.com

### DIFF
--- a/library/packaging/apt_key
+++ b/library/packaging/apt_key
@@ -254,7 +254,7 @@ def main():
                 keys2 = all_keys(module, keyring, short_format)
                 if len(keys) != len(keys2):
                     changed=True
-                if key_id and not key_id in keys2:
+                if key_id and not key_id[-16:] in keys2:
                     module.fail_json(msg="key does not seem to have been added", id=key_id)
                 module.exit_json(changed=changed)
     elif state == 'absent':


mpdehaan:
Can you provide some background on the change in logic for this one?   (Why just the last 16 characters, etc?)

hacosta:
Sure, this module uses:
apt-key adv --list-public-keys --keyid-format=long
To get a listing of the currently installed keys.
key IDs. "short" is the traditional 8-character key ID. "long" is the more accurate (but less convenient) 16-character key ID. [1]
However, fingerprints (which is what was used here are 20 bytes (longer than the longest possible output of list-public-keys).  Not to fear! since we can just get the last characters of the fingerprint to get the keyid.

1.- http://linux.die.net/man/1/gpg2

mpdehaan:
Merged, thanks!

Issue #7713, opened 2014-06-10
Spaces in SSH key path break Ansible


skoop:
##### Issue Type:

Bug Report
##### Ansible Version:

1.6.3
##### Environment:

OSX
##### Summary:

It seems that when the path to the SSH key contains spaces, these are not correctly escaped, effectively breaking Ansible. This issue was reported before: https://github.com/ansible/ansible/issues/6559 but was closed by the reporter for some reason. 
##### Steps To Reproduce:

I have my project set up in an OSX disk image for security reasons. This is automatically mounted by OSX into /Volumes/Disk Image/. In our project, an SSH key is defined in the base box that I need to specify to Ansible when running ansible-playbook. My command is:

```
ansible-playbook --private-key="/Volumes/Disk Image/project/key/id_rsa" --user=vagrant --limit="web" --inventory-file=ansible/inventory/my-dev ansible/vagrant-server.yml -vvvv
```

I've extracted this command from vagrant provision, since vagrant provision failed to report to me what the exist issue is. The vagrant provision command output:

```
==> web: Running provisioner: ansible...
ANSIBLE_FORCE_COLOR=true ANSIBLE_HOST_KEY_CHECKING=false PYTHONUNBUFFERED=1 ansible-playbook --private-key=/Volumes/Disk Image/project/key/id_rsa --user=vagrant --limit='web' --inventory-file=ansible/inventory/my-dev -v ansible/vagrant-server.yml

PLAY [web] ********************************************************************

GATHERING FACTS ***************************************************************
fatal: [172.31.222.222] => SSH encountered an unknown error during the connection. We recommend you re-run the command using -vvvv, which will enable SSH debugging output to help diagnose the issue

TASK: [default | updating] ****************************************************
FATAL: no hosts matched or all hosts have already failed -- aborting


PLAY RECAP ********************************************************************
       to retry, use: --limit @/Users/stefan/vagrant-server.retry

172.31.222.222             : ok=0    changed=0    unreachable=1    failed=0

Ansible failed to complete successfully. Any error output should be
visible above. Please fix these errors and try again.
```

Running the ansible-playbook directly with -vvvv gives me an error:

```
PLAY [web] ********************************************************************

GATHERING FACTS ***************************************************************
<172.31.222.222> ESTABLISH CONNECTION FOR USER: vagrant
<172.31.222.222> REMOTE_MODULE setup
[...]

PR #7951, opened 2014-06-27
spaces in host directory name will break ssh

earnubs:
Same as in #7713, except without ssh key


earnubs:
Actually, this is a bit more broken as exec_command also barfs on the space

PR #7954, opened 2014-06-27
cron module: Use correct parameter name in error message.

alimony:
The parameter is called `cron_file` but the error message uses `file` which is confusing.


AnneTheAgile:
+1
//AnneTheAgile

mpdehaan:
Merged, thanks!

PR #7955, opened 2014-06-27
Add support for relative paths in the template lookup plugin for roles

jdauphant:
Now the template lookup can be use in role. (add search in the role/templates/ and the playbook_dir)
I have add an error message if the template doesn't exist ( before the error was a empty string).

Same as https://github.com/ansible/ansible/commit/ad97c618cf82a1db2e242b8af039b4612294adde#diff-b4b03eaeefc82fb143509b97102a08ce but for the template lookup.

### DIFF
--- a/lib/ansible/runner/lookup_plugins/template.py
+++ b/lib/ansible/runner/lookup_plugins/template.py
@@ -16,7 +16,9 @@
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 
 from ansible.utils import template
+from ansible import errors
 import ansible.utils as utils
+import os
 
 class LookupModule(object):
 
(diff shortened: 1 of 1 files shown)


mpdehaan:
I'm rather confused what this code change does and why the code in question couldn't be applied specifically in the "template" lookup plugin.

It should already know to look in "files/" as it should, relative to the role, so I'm guessing this may not be needed.

Can you give examples of what you were trying to do before with playbook examples included, and expected vs actual results?

mpdehaan:
nm on part of the above, I see this is applied on the "template" module not a level up.  Ok.

Previous code makes sense a bit, but really should be abstracted so we don't repeat the code here.  

Comments above can be disregarded.  Thanks!

jdauphant:
You have right a simplification of the code could be better.
I am not enough confident with the Ansible API and python in general to know where to place it.
Have you a suggestion ?

Also for the record, a example of usage is to use this sort of task in a role :

```
  - name: define vm
    virt: name=foo
          command=define
          xml="{{ lookup('template', 'container-template.xml.j2') }}"
          uri=lxc:///
```

PR #7408, opened 2014-05-14
Add networking mode parameter (net) to docker

discordianfish:
Since 0.11 Docker supports different networking modes (see docker run --net). This commit add support for that to the ansible docker module.

This depends on dotcloud/docker-py@2d58351

PR #7956, opened 2014-06-27
Docker network mode selection

### LINKED ISSUE #7324: Support "--net" option in docker.io module
##### Issue Type:

Feature Idea
##### Ansible Version:

Latest
##### Environment:

N/A
##### Summary:

Docker v0.11 is released and it supports powerful `--net` option for advanced networking. Please add support of "--net" option to the `docker` module.
- Host networking (from the [announcement](http://blog.docker.io/2014/05/docker-0-11-release-candidate-for-1-0/))

For those wanting their containers to directly access the host system’s network interfaces, Docker 0.11 features a “host networking” mode. A container in host networking mode is sandboxed as usual, except for its network stack which is shared with the host system.

By default, when you start a container, it’ll only have a loopback interface `lo` and a virtual interface `eth1` connected to the docker bridge:

```
$> docker run busybox ip a
```

You can now directly use the interfaces of you host (it’s much faster as it doesn’t go through the bridge) via --net=host:

```
$> docker run --net=host busybox ip a
```
##### Steps To Reproduce:

N/A
##### Expected Results:

N/A
##### Actual Results:

N/A

pmyjavec:
Enables selection of the 'NetworkMode' option present in Docker

Granted this is nearly identical to of https://github.com/ansible/ansible/pull/7408 except after some testing the 0.3.2 checking seems important and I believe setting a default is required.

I believe adding might be important as without it, it doesn't seem docker-py sets it as the default, though I could be wrong. It was just another observation from testing.

This would close https://github.com/ansible/ansible/issues/7324

Cheers,

Pauly

### DIFF
--- a/library/cloud/docker
+++ b/library/cloud/docker
@@ -169,6 +169,13 @@ options:
     default: null
     aliases: []
     version_added: "1.5"
+  network_mode:
+    description:
+      - Set network mode for the container, e.g. network_mode=host
+    required: false
+    default: "bridge"
+    aliases: []
+    version_added: "1.6"
   stdin_open:
     description:
       - Keep stdin open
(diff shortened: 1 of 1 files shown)


jgmchan:
+1 we really need this fix.

pmyjavec:
Yup, would be good to get either one merged :)

pmyjavec:
Think I was wrong about my assumptions, argh!

PR #7963, opened 2014-06-27
Be able to pass in tag names from variables.

feanil:
When the module only accepted a dictionary for the tags, the names
could not use Jinja and so had to be statically defined.  This change
is backwards compatible but if you pass in a list of dictionaries, you
get more flexibility for the names you give your instances.

### DIFF
--- a/library/cloud/ec2_tag
+++ b/library/cloud/ec2_tag
@@ -69,6 +69,16 @@ tasks:
     tags:
       Name: webserver
       env: prod
+
+# Adding tags with dynamic names as well as dynamic values.
+tasks:
+- name: tag my instance
+  local_action: ec2_ntag resource={{ item.id }} region=us-east-1 state=present
+  with_items: ec2.instances
+  args:
+    tags:
+      - Name: "{{ some_variable }}"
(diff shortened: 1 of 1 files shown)


mpdehaan:
I'm a little unclear how the code above changes what is done, or why Jinja2 didn't apply before.  Can you go into a bit more detail?

feanil:
Previously tags would be passed in like this:

```
 - name: tag my launched instances
   local_action: ec2_tag resource={{ item.id }} region=eu-west-1 state=present
   with_items: ec2.instances
   args:
     tags:
       Name: webserver
```

In this case the key would be `Name` and the value would be `webserver`.  We could easily change this to have the value of webserver be dynamic or built using multiple vars using jinja because the values get rendered as jinja templates but the keys do not.

The change I made allows for a list of Name, Value dictionaries so that you can build up the key as well as the value using Jinja templates.  I used the words `Name` and `Value` because to be consistent with what AWS uses in their documentation.

feanil:
This was reopened as https://github.com/ansible/ansible-modules-core/pull/880

PR #7967, opened 2014-06-27
Add port knocking

urbaniak:
Hi,

I'm using port knocking which opens ssh port, to avoid some garbage in logs and "protect" from some bots trying to access ssh.

It's done by those simple iptables rules:

```
iptables -A INPUT -p tcp -m state --state NEW -m tcp --dport 22 -m recent --update --seconds 300 --name SSH --rsource -j ACCEPT
iptables -I INPUT -p tcp -m state --state NEW -m tcp --dport 1234 -m recent --remove --name SSH --rsource
iptables -I INPUT -p tcp -m state --state NEW -m tcp --dport 1235 -m recent --set --name SSH --rsource
iptables -I INPUT -p tcp -m state --state NEW -m tcp --dport 1236 -m recent --remove --name SSH --rsource
```

It will be quite good to have a variable which defines a port on which ansible will "knock" before trying to ssh there.

### DIFF
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -796,6 +796,15 @@ class Runner(object):
             return ReturnData(host=host, comm_ok=False, result=result)
 
         try:
+	    actual_port_knock = inject.get('ansible_ssh_port_knock')
+	    if actual_port_knock:
+		s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+		s.settimeout(0.0)
+		try:
+		    s.connect((actual_host, actual_port_knock))
+		except:
+		    pass
+
             conn = self.connector.connect(actual_host, actual_port, actual_user, actual_pass, actual_transport, actual_private_key_file)
             if delegate_to or host != actual_host:
                 conn.delegate = host


mpdehaan:
Hmm, I don't know about this.

I'm not convinced of the "real" security of port knocking, regardless, it is not assumed that only someone would be requiring socket activity on a given port.   What if the port requires "shave and a haircut, two bits, etc?" - this wouldn't cover but the most simplest implementations.  If it's something else, is that something else more than a plaintext password?   Wouldn't someone just knock _all_ the ports?   Would there be links to recommended practices?

You should absolutely control _where_ you accept login attempts, or limit login attempts to bastion hosts versus your entire cloud infrastructure, use VPCs, and things like this.   Ultimately, security in the connection layer would require security even if this feature were defeated.   I understand the log concern - you can definitely also change the default SSH port in ansible already, and for safety reasons I'd probably consider only allowing keys.

I don't see this as something we're likely to include, but I'm persuadable - this would be a good discussion to start on ansible-devel, perhaps.

This shouldn't hold you up though.

You could of course "pre-knock" all of your servers with a simple playbook that called "local_action" with netcat or something similar:

```
- hosts: all
  gather_facts: False
  tasks:
     - local_action: nc <options> {{ inventory_hostname }}
```

Let's start a thread on ansible-devel and see what comes of it.

PR #7968, opened 2014-06-28
acl module: now compares permissions correctly

bcoca:
should fix comparison bug reported in closed #6501


bcoca:
any chance this makes it in 1.7?  it fixes an annoying existing bug for users of the acl module

mpdehaan:
Merged, thanks!

PR #7971, opened 2014-06-28
Add docker connexion plugin using nsenter

mscherer:
This plugin permit to orchestrate a docker container from the
outside, using nsenter to enter the namespaced container without
needing to run ssh inside or anything.

See http://jpetazzo.github.io/2014/06/23/docker-ssh-considered-evil/ for
motivation of such a module

### DIFF
--- a/lib/ansible/runner/connection_plugins/docker-nsenter.py
+++ b/lib/ansible/runner/connection_plugins/docker-nsenter.py
@@ -1,0 +1,121 @@
+# Based on local.py (c) 2012, Michael DeHaan <michael.dehaan@gmail.com>
+# Based on chroot.py (c) 2013, Maykel Moya <mmoya@speedyrails.com>
+# (c) 2014, Michael Scherer <misc@zarb.org>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+import distutils.spawn
+import os
+import subprocess
+from ansible import errors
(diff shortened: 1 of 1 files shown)


mpdehaan:
How would hosts be referenced in inventory in this case?

I'd also wonder if this should be more simply renamed "local_docker" or something like that.

The real question is should this be in core if it can't remote to multiple hypervisors, in which case I think it's possible we may decide this is fine for people to use, but we might not wish to include it, as another example is people would build containers (hopefully using Ansible in the dockerfile) and then deploy to many different hosts and then be a bit lost as to why they would need to run ansible on each indivdiual host.

Does this make sense?  Perhaps a good topic for ansible-devel list.

mscherer:
My idea was more to use the docker module to create the docker container, then use a variable to orchestrate the container that was just created using this module.

abadger:
@mscherer I'm interested in merging something along this lines.  When  I brought this up at the ansible sprints at pycon, people said that nsenter might be obsoleted by docker exec.

Are there advantages to nsenter?  (It would work with other containers besides docker?)

Is there a chance that nsenter would become unmaintained or stops working with docker while docker exec continues to work?

abadger:
oh, this also would go into v2 rather than the current v1 tree.  We can sort that out once we figure out whether it needs to be nsenter or docker exec though.

abadger:
http://blog.docker.com/2014/10/docker-1-3-signed-images-process-injection-security-options-mac-shared-directories/

mscherer:
nsenter is part of utils-linux:

```
$ rpm -qf =nsenter
util-linux-2.23.2-21.el7.x86_64
```

And nsenter would indeed work on more than docker, but the current code only work with docker ( https://github.com/mscherer/ansible/commit/b82e121ae1610ea0166b99ab4cb94352b97913b0#diff-d760cbbc859c8133215387f7b4d570a4R50 ). So in the current state, nope, i do not think there is much advantage. We could switch to a model where we use the PID as a target, but it doesn't seem very user friendly IMHO. 

Maybe we could encode some information like docker/$dockername lxc/$lxcname , etc and then get the PID and enter the containers, but I am not sure it belong to the connexion plugin. I can come tomorow to the sprint to discuss ( and rebase a few of my PR )

abadger:
@mscherer sounds good.  I'm currently in 512a.  I may move if there's only a few people today but I'll be sure to let you know.

amenonsen:
@abadger @mscherer any thoughts on whether you want to proceed with this, now that we have #11650? Otherwise this PR can be closed.

mscherer:
I do not have a strong need for the connexion plugin, so I close this one.

PR #7985, opened 2014-06-30
Added OpenBSD install instructions

Mischi:


mpdehaan:
Thanks for this, though I think we've decided to suggest pip for most OSes as it allows us to get updates out a bit faster.

It's fine for people to continue to use the OS methods, however!

PR #7988, opened 2014-06-30
nova.py: Remove docs for unused config settings

msabramo:
AFAICT, these settings are not used. I wonder if @mavimo meant to implement these later, but never got around to it? I'm not the only one who noticed this and was confused by it (see https://github.com/ansible/ansible/pull/7444#issuecomment-43503710).

The other way to go is to implement these settings, but I'm not familiar with these, don't think that my company needs them, and I'm not sure if they require a newer version of python-novaclient or something. In short, I don't feel qualified to implement support, so I'm just removing the docs for the unimplemented feature. Someone else that needs these settings could perhaps add them in at a later date.

BTW, one of settings, `insecure` was misspelled as `insicure` (sic).

Cc: @mavimo, @carsongee (because if this gets merged, #7444 will probably need to be rebased, because it probably touches a lot of the same lines.

### DIFF
--- a/plugins/inventory/nova.ini
+++ b/plugins/inventory/nova.ini
@@ -21,10 +21,3 @@ project_id    =
 
 # Serverarm region name to use
 region_name   =
-
-# TODO: Some other options
-# insecure      =
-# endpoint_type =
-# extensions    =
-# service_type  =
-# service_name  =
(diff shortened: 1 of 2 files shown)


mpdehaan:
Actually an inventory script shouldn't have a DOCUMENTATION style block at all - nothing in Ansible will load it.

msabramo:
OK, see https://github.com/ansible/ansible/pull/7993

PR #7989, opened 2014-06-30
Explicit absolute import of ansible.utils.vault

msabramo:
in `ansible.utils`

This is more explicit, more readable, and recommended by PEP8 (http://legacy.python.org/dev/peps/pep-0008/#imports).

It's also Python 3 compatible, whereas implicit relative imports are not.


mpdehaan:
No thanks, change doesn't do anything.

PR #7990, opened 2014-06-30
Clean up imports ansible utils

msabramo:
Sorry, the contributing guide seems to sort of discourage changes for PEP8, whitespace, etc. but I couldn't help myself. It's just so much prettier now. And the old version had `import sys` twice so clearly having stuff be neat and tidy has a benefit. :smile: 

If you're worried about this breaking attribution (I guess you mean `git blame`?), then I might suggest that you take my diff and commit it yourself. I'm totally OK with that.

### DIFF
--- a/lib/ansible/utils/__init__.py
+++ b/lib/ansible/utils/__init__.py
@@ -45,8 +45,7 @@ import getpass
 import sys
 import json
 
-#import vault
-from vault import VaultLib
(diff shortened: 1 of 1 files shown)


mpdehaan:
Yep, we also don't break these as it's easy for someone to miss something on the diff.

We will correct them when there are actual problems.

PR #7992, opened 2014-06-30
fix swap translation on 4.x openbsd with 2.5.4 python

dehylton:
string.translate() operation in this version of python does not appear to match documentation. specifically, None is not accepted for the translation table.

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -31,6 +31,8 @@ import getpass
 import ConfigParser
 import StringIO
 
+from string import maketrans
+
 try:
     import selinux
     HAVE_SELINUX=True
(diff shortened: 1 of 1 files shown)


mpdehaan:
Sorry, I'm having some trouble following the above, can you give an example where the code above gets into trouble?

dehylton:
absolutely. my assertion that the code didn't follow the documentation was incorrect. the problem is centered on the version of python, and only related to OpenBSD in that this particular method is called for OpenBSD in this particular module. so instead of showing you actual OpenBSD output that triggers the problem, i'll strip it down to its lowest level ...

docs for 2.5.4:

```
translate(  table[, deletechars])
Return a copy of the string where all characters occurring in the optional argument deletechars are removed, and the remaining characters have been mapped through the given translation table, which must be a string of length 256.

You can use the maketrans() helper function in the string module to create a translation table.
```

this is different in the newer python implementations; here are the docs for 2.6:

```
str.translate(table[, deletechars])

    Return a copy of the string where all characters occurring in the optional argument deletechars are removed, and the remaining characters have been mapped through the given translation table, which must be a string of length 256.

    You can use the maketrans() helper function in the string module to create a translation table. For string objects, set the table argument to None for translations that only delete characters:

    >>> 'read this short text'.translate(None, 'aeiou')
    'rd ths shrt txt'

    New in version 2.6: Support for a None table argument.
```

now, some code to show it in action:

```
$ uname -sr
OpenBSD 4.6
$ python
Python 2.5.4 (r254:67916, Jul  1 2009, 11:37:21)
[GCC 3.3.5 (propolice)] on openbsd4
>>> '1234k'.translate(None, 'k')
TypeError: expected a character buffer object
>>> from string import maketrans
>>> '1234k'.translate(maketrans(' ',' '), 'k')
'1234'

...

$ uname -sr
OpenBSD 5.1
$ python
Python 2.7.1 (r271:86832, Feb  7 2012, 06:26:14)
[GCC 4.2.1 20070719 ] on openbsd5
>>> '1234k'.translate(None, 'k')
'1234'
```

mpdehaan:
merged, thanks!

dehylton:
excellent, thanks! sorry for the badly-written description; i'll try to do better next time. 

----- Original Message -----

PR #7994, opened 2014-07-01
Correct typo: 'additioanl' -> 'additional'

adamnbowen:


jimi-c:
Merged, thanks! There's actually another typo in that paragraph: `like any other pice` if you'd like to fix that.

PR #8001, opened 2014-07-01
added the missing -f option to mkfs if force used

prune998:
the -f flag is not specified in the mkfs command line, so the fs will not be created even if the force=yes option is set in the playbook.
I added the -f in any case as the test case to stop the format if the fs already exists is done before in the 

```
elif fs and not force:
        module.fail_json(msg="'%s' is already used as %s, use force=yes to overwrite"%(dev,fs), rc=rc, err=err)
```


prune998:
I canceled as this code is not workign either... going with a new pull request soon

PR #8002, opened 2014-07-01
added missing force flag on mkfs commandline

prune998:
even if the option `force=yes` is used in the playbook, it is not reflected in the mkfs command line.
As force option is dependent of the fs type, a "if-then-else" case have been added. Also, some FS types does not have a force option.


mpdehaan:
It appears around line 98 things need to be updated to use four space indent.

Can you take care of this?  Thanks!

mpdehaan:
Not important, can deal with that later.

Merged, thanks!

PR #8005, opened 2014-07-01
Make resource_tags required and do whitespace cleanup

psa:
Originally in #7288
- ec2_vpc:
  - make resource_tags required as not having them causes a new VPC to be launched every time the playbook is run


jimi-c:
Merged, thanks!

PR #8007, opened 2014-07-01
nova.py: Set defaults for OpenStack settings

msabramo:
`nova.py`: Set defaults for OpenStack settings
- `auth_system`
- `region_name`

These are config settings that could be left out in many scenarios, but the current code is requiring them. With this in place, I can get by with this minimal .ini file:

``` ini
[openstack]
version=1.1
auth_url=http://{keystone_host}:5000/v2.0
project_id=web
username={username}
api_key={password}
```

Cc: @jimi-c, @sontek, @aconrad


jimi-c:
Merged, thanks!

PR #8009, opened 2014-07-01
Add support for nested inventory groups with EC2

cchurch:
Without this change (which is still the default setting), all EC2 inventory groups are created in a flat namespace.  This change adds additional parent groups to contain regions, instance types, key pairs, security groups and tags.

This change has already been incorporated into Tower 1.4.11 and later.

Example without nested groups:
- i-12345678
- us-east-1
- us-east-1c
- type_t1_micro
- key_foo
- security_group_default
- tag_name_webservers

Example with nested groups:
- instances
  - i-12345678
- regions
  - us-east-1
    - us-east-1c
- types
  - type_t1_micro
- keys
  - key_foo
- security_groups
  - security_group_default
- tags
  - tag_name
    - tag_name_webservers

Nested groups are also supported for RDS instances with similar parent groups as above, along with "rds_engines" and "rds_parameter_groups" parent groups.

I've also added RDS instance variables to ["_meta"]["hostvars"] which were previously missing.

### DIFF
--- a/plugins/inventory/ec2.ini
+++ b/plugins/inventory/ec2.ini
@@ -53,3 +53,6 @@ cache_path = ~/.ansible/tmp
 # seconds, a new API call will be made, and the cache file will be updated.
 # To disable the cache, set this value to 0
 cache_max_age = 300
+
+# Organize groups into a nested/hierarchy instead of a flat namespace.
+nested_groups = False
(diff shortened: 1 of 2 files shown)


mpdehaan:
merged, thanks very much!

jimi-c:
Merged, thanks!

PR #8010, opened 2014-07-01
Add option to return all EC2/RDS instances regardless of state.

cchurch:
By default, the inventory plugin only returns instances that are "running" (EC2) or "available" (RDS).  It may be desirable in certain cases (e.g. Tower inventory sync) to retrieve all instances and then filter or group hosts based on their state.  This pull request preserves the default behavior, but adds configurable options for retrieving all instances.


mpdehaan:
Merged, thanks!

PR #8012, opened 2014-07-01
Callback plugin cleanup

cchurch:
on_error / runner_on_error is never used by runner, and I can't find any indication of it being used a least as far back as v1.0, so I've removed it from built-in callbacks and example callback plugins (955dadf7431f184f249e7b1903c05f661c232908).

The playbook_on_play_start callback receives the name of the play as its argument (https://github.com/ansible/ansible/blob/devel/lib/ansible/playbook/__init__.py#L615) instead of the host pattern.  The two will be the same only if the play didn't explicitly specify a name.  I've changed the argument from 'pattern' to 'name' (30fae95efeec959be880800470459e0e58311a4b and afbc7f8a0f344c0f82d07709a0da9f353c47f54a).

When a play fails, we break out of the loop and don't clear the 'play' attribute stored on each callback plugin instance.  I've changed it so that the play is cleared whenever we finish running all plays in the playbook (4e98e3785acee725b6f70f9f30d179b62662dd10).

### DIFF
--- a/lib/ansible/callback_plugins/noop.py
+++ b/lib/ansible/callback_plugins/noop.py
@@ -41,9 +41,6 @@ class CallbackModule(object):
     def runner_on_ok(self, host, res):
         pass
 
-    def runner_on_error(self, host, msg):
-        pass
-
     def runner_on_skipped(self, host, item=None):
         pass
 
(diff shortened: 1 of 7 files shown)


mpdehaan:
merged, thanks!

PR #8015, opened 2014-07-02
win_service module updates and tests

cchurch:
Update the win_service module by @chrishoffman to return additional data in the module result (name, display_name, start_mode and state).  Add integration tests to exercise module.


chrishoffman:
I like these integration tests.  I am going to have to integration them into the rest of the modules.

mpdehaan:
merged, thanks!

PR #8018, opened 2014-07-02
Juniper Junos OS Modules

shermdog:


mpdehaan:
Per a recent thread on ansible-project we are going to be moving networking modules to Ansible galaxy (Arista modules are also moving out of core).   This allows vendor updates at any time (as I believe you had expressed interest in) as well as easy expansion.

https://www.mail-archive.com/ansible-project@googlegroups.com/msg06627.html

I would suggest hosting these on Galaxy under a "juniper" account, and in the role have a "./library" directory - include modules in there and anyone can include the juniper role and use them in their playbooks.

I hope this makes sense, for any other questions please ask on ansible-project list.

We believe this is a better approach than being the gatekeeper to vendor commits and limiting how people may choose to use things, and this also (technically) allows modules to be written in non-Python languages to be hosted on Galaxy as well (though I might not encourage it, it would be fine).

https://galaxy.ansible.com/intro

As I'm closing this ticket, replies will not be monitored, so follow up on ansible-project if you have questions.

PR #8020, opened 2014-07-02
Error if private_key_file is group/world readable

msabramo:
Currently, if you specify an SSH private key that is group or world readable, then ansible fails with a generic error and suggests running again with `-vvvv`. This isn't bad but pinpointing the specific problem immediately is even more user-friendly.

```
$ ls -l devops.pem
-rw-r--r--+ 1 marca  staff  1679 Jul  2 11:25 devops.pem

$ ansible -m ping --private-key=devops.pem mt3-pyweb01
mt3-pyweb01 | FAILED => private_key_file (devops.pem) is group-readable or world-readable and thus insecure - you will probably get an SSH failure
```

Cc: @sontek, @aconrad

### DIFF
--- a/lib/ansible/runner/connection.py
+++ b/lib/ansible/runner/connection.py
@@ -18,6 +18,9 @@
 
 ################################################
 
+import os
+import stat
+
 from ansible import utils
 from ansible.errors import AnsibleError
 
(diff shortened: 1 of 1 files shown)


lib/ansible/runner/connection.py
```diff
@@ -18,6 +18,10 @@
 
 ################################################
 
+import os
+import stat
+import warnings
```
sontek: You never use warnings here

sontek:
+1

lib/ansible/runner/connection.py
msabramo: Ah yes, as you probably guessed, I was originally going to use `warnings.warn` instead of `raise AnsibleError` and then I changed my mind.

Thanks!

mpdehaan:
Merged, thank you!

PR #8022, opened 2014-07-02
Fix for #7368 breaks peering connections. Unreak them.

psa:
#7368  breaks peering by making instance routing the default instead of gateway routing. This patch makes gateway routing the default again.


jimi-c:
Merged, thanks!

PR #8023, opened 2014-07-02
ansible_ec2_instance-id key incorrect

btashton:
ansible_ec2_instance-id should be ansible_ec2_instance_id

fatal: [_IP_] => One or more undefined variables: 'dict object' has no attribute 'ansible_ec2_instance-id'


jimi-c:
Merged, thanks!

PR #5869, opened 2014-02-04
Fixed UnicodeDecodeError: 'utf8' codec can't decode byte 0xb7 ...

dagwieers:
Without this fix I get an ugly traceback when using the uri module using return_content=yes on my own website. Have not tested beyond that.

```
[dag@moria ~]$ ansible localhost -m uri -a 'url=http://dag.wieers.com/ return_content=yes'
localhost | FAILED >> {
    "failed": true, 
    "msg": "Traceback (most recent call last):\n  File \"/home/dag/.ansible/tmp/ansible-1391523728.61-242369951155415/uri\", line 1450, in <module>\n    main()\n  File \"/home/dag/.ansible/tmp/ansible-1391523728.61-242369951155415/uri\", line 419, in main\n    module.exit_json(changed=changed, content=content, **uresp)\n  File \"/home/dag/.ansible/tmp/ansible-1391523728.61-242369951155415/uri\", line 1258, in exit_json\n    print self.jsonify(kwargs)\n  File \"/home/dag/.ansible/tmp/ansible-1391523728.61-242369951155415/uri\", line 1248, in jsonify\n    return json.dumps(data)\n  File \"/usr/lib64/python2.6/json/__init__.py\", line 230, in dumps\n    return _default_encoder.encode(obj)\n  File \"/usr/lib64/python2.6/json/encoder.py\", line 367, in encode\n    chunks = list(self.iterencode(o))\n  File \"/usr/lib64/python2.6/json/encoder.py\", line 309, in _iterencode\n    for chunk in self._iterencode_dict(o, markers):\n  File \"/usr/lib64/python2.6/json/encoder.py\", line 275, in _iterencode_dict\n    for chunk in self._iterencode(value, markers):\n  File \"/usr/lib64/python2.6/json/encoder.py\", line 294, in _iterencode\n    yield encoder(o)\nUnicodeDecodeError: 'utf8' codec can't decode byte 0xb7 in position 2804: invalid start byte\n", 
    "parsed": false
}
```

After adding that snippet:

```
[dag@moria ~]$ ansible localhost -m uri -a 'url=http://dag.wieers.com/ return_content=yes'
localhost | success >> {
    "_content_encoding": "gzip", 
    "cache_control": "no-cache; must-revalidate", 
    "changed": false, 
    "connection": "close", 
    "content": "<HTML lang=\"en\">\n\t<head>\n\t\t<title>DAG: Field Commander Wieers</title>....</body>\n</html>\n", 
    "content_length": "11293", 
[...]

PR #8027, opened 2014-07-03
uri: provide raw_content, parse json without double-decoding

### LINKED ISSUE #7586: uri module doesn't properly handle escape sequences in JSON
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.5.4
##### Environment:

Ubuntu 12.04
##### Summary:

uri module calls `content.decode('unicode_escape')` on response, which means escape sequences in JSON responses aren't handled properly
##### Steps To Reproduce:

You can test this by running this ansible command line against a test URL I uploaded:

```
ansible localhost -m uri -a "url=https://s3.amazonaws.com/oresource/python_version.json return_content=true"
```
##### Expected Results:

Output should have a "json" key with the PythonVersion string decoded properly.
##### Actual Results:

```
localhost | success >> {
    "accept_ranges": "bytes",
    "changed": false,
    "content": "{\"PythonVersion\": \"2.7.6 (default, May 16 2014, 16:48:38) \n[GCC 4.4.7 20120313 (Red Hat 4.4.7-4)]\"}",
    "content_length": "100",
    "content_location": "https://s3.amazonaws.com/oresource/python_version.json",
    "content_type": "application/json",
    "date": "Wed, 28 May 2014 21:00:47 GMT",
    "etag": "\"2ad455693e539d42df070a367a8f16a3\"",
    "last_modified": "Wed, 28 May 2014 20:52:45 GMT",
    "redirected": false,
    "server": "AmazonS3",
    "status": 200,
    "x_amz_id_2": "Li22a6y8Vb+mYo8AZTN04QolNK+jlJRD8O3nS/rsmUxTpQMAT0ie8K6rbXOf6sDE",
    "x_amz_request_id": "E077ABA55DC7E255"
}
```
##### Further comments

Note that the output does not include the "json" key due to the above bug.

I've been trying to use the "uri" module and hitting an HTTP endpoint which returns some JSON with header `Content-Type: application/json`. However, one of the strings in my JSON has a line feed (`\n` control character) in it. This is correctly escaped, and is valid JSON. However, the "uri" module does the following on the response content (see [here](https://github.com/ansible/ansible/blob/devel/library/network/uri#L307)):

```
unicode(content.decode('unicode_escape'))
```

This decodes the `\n` string to a literal line feed, and causes the `json.loads()` further down to fail with an exception like:

```
ValueError: Invalid control character at: line 1 column 58 (char 58)
```

There's a catch-all exception around the json.loads() that silently ignores this error without setting the `.json` key.

So two things:

1) Why does the uri module do this extra (un)escaping, effectively breaking the content in this case?

2) This issue was hardish to debug as the `json.loads()` has a catch-all except around it. (Catch-all except clauses are bad practice for this reason.) I recommend changing it to a "except ValueError, error" and then call `module.fail_json()` with a string representation of the error in the response.

darkk:
Fixes #7586 
Regression potential:
- `raw_content` is written to `dest` file instead of decoded `content`
- `raw_content` doubles module reply


srgvg:
+1 for me, as this solves an old bug where I download zip files, and sometimes get the dreaded 

`msg: failed to create temporary content file: 'ascii' codec can't encode character u'\x85' in position 10: ordinal not in range(128)`

(That old bug started after 97eeded and cb79947, didn't have the time to mention it, and then forgot, working around it with the old version of uri in my personal library/)

darkk:
This branch does not introduce regression regarding #5869 (1) for me (2) using ubuntu 14.04 & python 2.7.

@jimi-c do you have a test-case for #5869 ?

jimi-c:
@darkk I believe at the time we did not yet have integration tests for modules, so my test cases were usually sample playbooks provided with issues. Unfortunately, your unit test uses "start-stop-daemon", which is not cross-platform and will need to be changed.

darkk:
@jimi-c is `async` with hardcoded timeout good enough replacement for `start-stop-daemon` in case of tests?

jimi-c:
Ideally I would not like to add any deps for running the integration tests, so if you can create a simple python wrapper to daemonize the code and deploy/remote that as part of the integration test it would be the best approach.

darkk:
@jimi-c I've already replaced `start-stop-daemon` with `async` in the branch and it works.

I see following issues with the implementation:
- SimpleHTTPServer is terminated with delay and I'm not sure if it worth fixing. The process is eventually terminated using timeout.
- role with tests assumes that it's applied to `localhost` as files/\* are not transfered. This assumption is valid at the moment and I don't know if tests should be future-proof.

There are no external dependencies: SimpleHTTPServer is part of python/2.5+, `async` is ansibe feature.

jimi-c:
My apologies, I thought you were referencing some other CLI utility and not an ansible feature... that should work. You could have the script launch it and report the PID, then kill it at the end of the integration test too.

darkk:
Are there any open questions regarding the code?
Is some improvement required or is long review queue the only remaining issue to the moment? :)

jimi-c:
@darkk no this is fine, we're just continuing to work through our queue and we'll get to this as soon as possible.

jlaska:
@darkk thanks for including integration tests!  I have some feedback on the integration tests ...
1. Use make use of `start-stop-daemon` which is only available on Ubuntu.  Can you either support other Fedora-based distributions, or only conditionally run the tests for Ubuntu?
2. If the integration tests fail, they'll leave the SimpleHTTPServer still running.  I'd suggest some cleanup handling in the `Makefile`.  A similar technique is used for cleaning up [stale amazon test resources](https://github.com/ansible/ansible/blob/devel/test/integration/Makefile#L89).

test/integration/roles/test_uri/tasks/main.yml
```diff
@@ -0,0 +1,121 @@
...
+    checkout_dir: '{{ output_dir }}/git'
+
+- name: verify that python2 is installed so this test can continue
+  shell: which python2
+  register: py2
+
+- name: start SimpleHTTPServer
+  shell: cd {{ files_dir }} && {{ py2.stdout }} -m SimpleHTTPServer {{ http_port }}
+  async: 15 # this test set takes ~8 seconds to run
+  poll: 0
```
darkk, line 32: > `start-stop-daemon`

@jlaska  I've already removed start-stop-daemon and replaced it with `async` ansible task.
`async` also handles cleanup, although in an ugly way (timeout). I think it's good enough for tests. Am I wrong?

darkk:
Should I resolve merge conflicts or will you do it while merging?

jimi-c:
@darkk If you could do it, that would be helpful. Thanks!

mpdehaan:
Hi!

Thanks you very much!

keronsen:
@darkk @jimi-c Have this PR been reopened in ansible-modules-core? I was trying to fix a json issue in the uri module https://github.com/ansible/ansible-modules-core/issues/121 when I came across this (by the looks of it) existing fix.

abadger:
If anyone is interested in this issue still, please have a look at: https://github.com/ansible/ansible-modules-core/pull/1331 to see if it fixes the issues for you (or at least, does not make things any worse).

PR #8030, opened 2014-07-03
ansible-shell

dominis:
ansible-shell is an interactive shell for ansible with built-in tab completion for all the documented modules

Available commands:

```
 cd - change host/group (you can use host patterns eg.: app*.dc*:!app01*)
 list - list available hosts in the current path
 serial - change serial
 sudo - sudo
 ! - forces shell module instead of the ansible module (!yum update -y)
```

Configuration:
 create a [ansible-shell] section in your ansbile.cfg

```
 cwd=app-servers - you can set your default host pattern
 serial=100 - default serial
```


msabramo:
:+1:

neersighted:
:+1:, awesome!

resmo:
:+1:

jkordish:
+1

maxamillion:
+1 (fwiw)

ghost:
I have to say, the code is very nice. There's only one or two spots where I would have liked a comment to explain what's going on but it looked self-explanatory (especially with a deeper reading of the code).

Awesome, very awesome. Hope it gets merged in and soon!

mpdehaan:
Sorry for the delay in getting to this one.

Making some notes now, which I may personally hack on.

(A)  running without an inventory file gives a traceback about a missing inventory file, when this message should be not a traceback.

mpdehaan:
bash-3.2# ./bin/ansible-shell -i /tmp/hosts
  File "./bin/ansible-shell", line 349, in <module>
    AnsibleShell(options, args).cmdloop()
  File "./bin/ansible-shell", line 76, in **init**
    self.read_config_file()
  File "./bin/ansible-shell", line 92, in read_config_file
    if config.has_section('ansible-shell'):
AttributeError: 'NoneType' object has no attribute 'has_section'

These values should have defaults so it does not traceback, and should also load parameters via constants.py versus explicit command parser usage.

mpdehaan:
This is very nice -- thanks for submitting and sorry about the delay.

Here's some initial review comments:

Help cd should give an example

Should probably be a subcommand "connection" since -c is not supported.

Should probably have a subcommand "forks"

Typing "ping" after logging in with 5 groups gives "No host found" until you cd into "all", possibly a consequence of my hacking around the config file issue.

Typing "ping" appears to have done "shell ping", so I think we might make specifying "shell" required.

Helptext says "shell" is undocumented

Callback output is a little different from /usr/bin/ansible and should probably use the ansible callback classes.

I'm unclear what "dump" does and "EOF" seems to be an alias for "exit"

control-C appears to be intercepted -- when commands are not running I think control-C should exit the program.

any CLI flags to /usr/bin/ansible should be used in common to the shell.  There's a utility function that can be used for this.

It would be nice if there was a way to execute other modules than shell, for instance, if "shell echo hi" is the way to run "echo hi", I should expect running "modulename key=value key=value" to call that module.

If those things above can be cleaned up I can see merging this definitely - this may be something we get to ourselves, though if others want to take this and run with it that would be appreciated.

It's definitely a very nice feature, just has enough stumbling blocks at the moment where it could cause a series of bug reports and want to get them stomped before merging.

dominis:
Hey Michael,

thanks for the comments! I'm moving to a new flat in the next few days but i'll defenitely look into them next week.

dominis:
> Help cd should give an example

done in 6576129

> Should probably be a subcommand "connection" since -c is not supported.

d9b9f0e

> Should probably have a subcommand "forks"

there is a serial subcommand but I've created a fork as well in 83f4e19

> Typing "ping" after logging in with 5 groups gives "No host found" until you cd into "all", possibly a consequence of my hacking around the config file issue.

fixed in 2b40323

> Typing "ping" appears to have done "shell ping", so I think we might make specifying "shell" required.

so by default if there is a module name matched to the command it will use the module:

```
⌂29% [:~/src/ansible] [venv] devel(3) ± ./bin/ansible-shell -i /tmp/hosts
Welcome to the ansible-shell.
Type help or ? to list commands.

dominis@* (50)[s:100]$ cd app01.bfc
dominis@app01.bfc (1)[s:100]$ ping
=============== app01.bfc ================
{
    "changed": false,
    "invocation": {
        "module_args": "",
        "module_name": "ping"
    },
    "ping": "pong"
}
```

if you want to run a shell ping you have to run this:

```
dominis@app01.bfc (1)[s:100]$ !ping -c1 192.168.2.1
=============== app01.bfc ================
PING 192.168.2.1 (192.168.2.1) 56(84) bytes of data.
64 bytes from 192.168.2.1: icmp_seq=1 ttl=255 time=1.61 ms

--- 192.168.2.1 ping statistics ---
1 packets transmitted, 1 received, 0% packet loss, time 1ms
rtt min/avg/max/mdev = 1.618/1.618/1.618/0.000 ms
```

> Helptext says "shell" is undocumented

added 3f88c2f

> Callback output is a little different from /usr/bin/ansible and should probably use the ansible callback classes.

```
dominis@app01.bfc (1)[s:100]$ ping
=============== app01.bfc ================
{
    "changed": false,
    "invocation": {
        "module_args": "",
        "module_name": "ping"
    },
    "ping": "pong"
}
dominis@app01.bfc (1)[s:100]$ ^D
⌂38% [:~/src/ansible] [venv] devel(4) 1m41s ± ./bin/ansible app01.bfc* -i ~/src/puppet/files/gm-cobbler/etc_ansible_hosts -m ping
app01.bfc.kinja-ops.com | success >> {
    "changed": false,
    "ping": "pong"
}
```

> It would be nice if there was a way to execute other modules than shell, for instance, if "shell echo hi" is the way to run "echo hi", I should expect running "modulename key=value key=value" to call that module.

this is one of the main purpose of this project. I was always forget the argument names for the modules so the whole thing started as a wrapper for that. I don't know why it's not working for you

```
dominis@app01.bfc (1)[s:100]$ ? yum
Manages packages with the I(yum) package manager
Parameters:
  state Whether to install (C(present), C(latest)), or remove (C(absent)) a package.
  disablerepo I(Repoid) of repositories to disable for the install/update operation. These repos will not persist beyond the transaction. When specifying multiple repos, separate them with a ",".
  name Package name, or package specifier with version, like C(name-1.0). When using state=latest, this can be '*' which means run: yum -y update. You can also pass a url or a local path to a rpm file.
  enablerepo I(Repoid) of repositories to enable for the install/update operation. These repos will not persist beyond the transaction. When specifying multiple repos, separate them with a ",".
  list Various (non-idempotent) commands for usage with C(/usr/bin/ansible) and I(not) playbooks. See examples.
  disable_gpg_check Whether to disable the GPG checking of signatures of packages being installed. Has an effect only if state is I(present) or I(latest).
  conf_file The remote yum configuration file to use for the transaction.

dominis@app01.bfc (1)[s:100]$ yum name=ntpd state=present
```

to run yum via the shell module you have to do this:

```
dominis@app01.bfc (1)[s:100]$ !yum install ntpd
```

dominis:
> control-C appears to be intercepted -- when commands are not running I think control-C should exit the program.
> any CLI flags to /usr/bin/ansible should be used in common to the shell. There's a utility function that 
> can be used for this.

both fixed. 

@mpdehaan could you please check your list? thanks

bcoca:
This breaks with non standard module paths, it should be using "utils.plugins.module_finder". Also it assumes that the library path is a single path string, it can be multiple ':' separated.

```
  File "/home/bcoca/work/ansible/bin/ansible-shell", line 351, in <module>
    AnsibleShell(options, args).cmdloop()
  File "/home/bcoca/work/ansible/bin/ansible-shell", line 78, in __init__
    self.modules = self.list_modules()
  File "/home/bcoca/work/ansible/bin/ansible-shell", line 133, in list_modules
    for root, dirs, files in os.walk(C.DEFAULT_MODULE_PATH):
  File "/usr/lib64/python2.7/os.py", line 278, in walk
    names = listdir(top)
TypeError: coercing to Unicode: need string or buffer, NoneType found
```

dominis:
I'm closing this. It was opened ~7 months ago and I don't see it will be merged ever

resmo:
please reopen. I know, things not always gets fast in this project but this may change soon.

bcoca:
reopen, still needs some work to be merged but we definitely want this, probably once v2 is out

dominis:
Let me know when you're ready to merge it, until then I'll continue development on its own repo https://github.com/dominis/ansible-shell/ it's easier for me to review and merge PR over there.

PR #8034, opened 2014-07-03
Updated tag behavior for AutoScale Groups

jsmartin:
This pull request is intended to make the ec2_asg behave similar to the ec2 module with regards to tagging:

Currently the module expects tags in this form:

```
tags:
  - key: environment
    value: production
    propagate_at_launch: no
```

With this patch, it can be streamlined to:

```
tags:
  - environment: production
    propagate_at_launch: no
```

And makes it in line with the ec2 module's behavior:

```
instance_tags:
    db: postgres
```

### DIFF
--- a/library/cloud/ec2_asg
+++ b/library/cloud/ec2_asg
@@ -69,7 +69,7 @@ options:
     default: None
   tags:
     description:
-      - List of tag dictionaries to use. Required keys are 'key', 'value'. Optional key is 'propagate_at_launch', which defaults to true.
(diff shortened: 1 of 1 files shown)


mpdehaan:
Is this intended to be backward compatible with the old format for people who might have already deployed this?

jsmartin:
No.  The support for tags in autoscale groups was added in 1.7 on June 9th.

terretta:
I agree with jsmartin the different (recently added) tag approach in ASG was annoying.  Much prefer this updated behavior.

jsmartin:
Added backwards compatibility with old format, and marked it as deprecated in the example.

PR #8038, opened 2014-07-04
Fix missing msg argument

caphrim007:
The following patch adds a missing 'msg=' syntax. An exception is raised
in ansible if this block is reached during the execution of the module

```
TypeError: fail_json() takes exactly 1 argument (2 given)
```

With the 'msg=' added, you get a more informative error. For example

```
msg: No settings provided to update_domain().
```


jimi-c:
Merged, thanks!

PR #8039, opened 2014-07-04
Suppor for joyent inventory script

ahelal:
Hi,

We are using joyent cloud http://www.joyent.com/ I developed an invtory script I hope that might help other users.


amenonsen:
Anyone interested in reviewing/testing this?

jimi-c:
As there has been no follow up on this, and since we have since relocated the location of the inventory files, we'll go ahead and close this with the caveat that the original submitter can re-submit it at a later date if they so choose.

Thanks!

PR #8044, opened 2014-07-06
Added module win_domainjoin.

trondhindenes:
win_domainjoin allows Ansible to join a windows computer to a domain. Parameters include the ability to auto-restart after a successfull join (this is turned off by default but should possily be enabled) and which OU path (by DistinguishedName) to put the computer into.

### DIFF
--- a/library/windows/win_domainjoin
+++ b/library/windows/win_domainjoin
@@ -1,0 +1,86 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2014, Trond Hindenes <trond@hindenes.com>, and others
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+# this is a windows documentation stub.  actual code lives in the .ps1
+# file of the same name
+
+DOCUMENTATION = '''
+---
+module: win_domainjoin
+version_added: "1.7"
+short_description: Adds a computer to an Active Directory domain
+description:
+     - Adds a computer to an Active Directory domain
+options:
+  domainname:
+    description:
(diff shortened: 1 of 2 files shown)


library/windows/win_domainjoin
```diff
@@ -0,0 +1,86 @@
...
+    required: true
+    default: null
+    aliases: []
+  domainpassword:
+    description:
+      - password for the domainusername account    
+	required: true
+    default: null
+    aliases: []
+  OUPath:
```
cchurch, line 50: I would lean toward this parameter being lower case to be consistent with the others (and Ansible modules in general); just my 2 cents.

library/windows/win_domainjoin
```diff
@@ -0,0 +1,86 @@
...
+    description:
+      - The OU in which to place the computer
+    default: null
+    aliases: []
+  Restart:
```
cchurch, line 55: Parameter name is capitalized here, but not when used below.  I know PowerShell variables aren't case-sensitive, but code and docs should at least match.

library/windows/win_domainjoin.ps1
```diff
@@ -0,0 +1,123 @@
...
+
+$result = New-Object psobject @{
+    changed = $false
+}
+
+If ($params.domainname) {
+    $domainname = $params.domainname
+}
+Else {
+    Fail-Json $result "mising required argument: domainname"
```
cchurch, line 33: s/mising/missing

library/windows/win_domainjoin.ps1
```diff
@@ -0,0 +1,123 @@
...
+{
+    $JoinDomain = $true
+
+}
+
+
+if ($JoinDomain)
+{
+    #Construct credobject from username/password
+    #Test parameters
```
cchurch, line 79: I don't see any need to keep your test lines around...

library/windows/win_domainjoin.ps1
```diff
@@ -0,0 +1,123 @@
...
+            $ErrorActionPreference = "Stop"
+            Add-Computer -DomainName $domainname -Credential $credential -OUPath $OUPath -WarningVariable WarningVar -restart:$restart
+        }
+        catch
+        {
+            Fail-Json $result $_.Exception.Message
+        }
+    }
+
+    Set-Attr $result "domainjoin_result" $success
```
cchurch, line 114: Where is $success defined?

library/windows/win_domainjoin.ps1
```diff
@@ -0,0 +1,123 @@
...
+    $result.changed = $true
+}
+Else
```
cchurch, line 117: This Else block is not really necessary since you already set $result.change = $false above.

trondhindenes:
Lots of good comments here Chris, thanks. I'm closing this PR and will open a new one with revised code.

PR #8045, opened 2014-07-06
Added script for configuring winrm for Ansible

trondhindenes:
The script can be used to set up a windows host with WinRM with the least possible effort.


jimi-c:
Merged, thanks!

PR #8048, opened 2014-07-06
Win msi improvements

trondhindenes:
I basically rewrote the win_msi module. 
-This version actually checks the exit code of the install, and reports error if its anything other than zero (sucess)

-Added a few new parameters to make it easier to skip already installed software, using WMI/CIM classes rather than just a "check for existing file" like the previous version did.

### DIFF
--- a/lib/ansible/inventory/__init__.py
+++ b/lib/ansible/inventory/__init__.py
@@ -146,7 +146,7 @@ class Inventory(object):
 
         # get host vars from host_vars/ files and vars plugins
         for host in self.get_hosts():
-            host.vars = utils.combine_vars(host.vars, self.get_variables(host.name, vault_password=self._vault_password))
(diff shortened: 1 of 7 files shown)


mpdehaan:
Thoughts @sivel ?

sivel:
There are some spelling errors such as "exit_reasion" instead of "exit_reason".

Other than that, there are a number of places where using `Get-Attr` would shorten some code and allow for a safer way of accessing members of the custom psobject.

mpdehaan:
Thoughts on updates @trondhindenes ?

trondhindenes:
Sorry, been on vacation. I'll get these fixes in during the weekend.

trondhindenes:
I have no idea what those other commits are doing here. Closing this.

PR #8052, opened 2014-07-07
assert: add optional `msg` argument

darkk:
Sample playbook when it may be useful:

```
- uri: return_content=yes url={{ api_base }}/action/{{ inventory_hostname }}
  register: output
- assert:
    msg: '{{ output.json | to_nice_json }}'
    that:
      - output.json.foo != "bar"
      - output.json.boo == 42
```


jimi-c:
Merged, though I tweaked it in one regard so that the msg= will only be shown if it was set specifically, rather than defaulting it to a generic message.

jimi-c:
Thanks!

PR #8053, opened 2014-07-07
ec2 module accepts user_data_file param

mcclurmc:
While the boto library doesn't support it, most other ec2 libraries and command line tools provide an option to specify a file which contains user data, instead of providing it as a string. This patch adds this capability to the ec2 module.

I've set the user_data_file's version_added to "1.7". I wasn't sure what it should be set to, so can change this if I guessed wrongly.

### DIFF
--- a/library/cloud/ec2
+++ b/library/cloud/ec2
@@ -130,6 +130,13 @@ options:
     required: false
     default: null
     aliases: []
+  user_data_file:
+    version_added: "1.7"
+    description:
+      - file containg opaque blob of data which is made available to the ec2 instance
+    required: false
+    default: null
+    aliases: []
   instance_tags:
     version_added: "1.0"
     description:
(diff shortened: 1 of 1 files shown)


jsmartin:
Does this patch make possible to use jinja2 evaluation in the user_data_file?

mcclurmc:
@jsmartin Right now, the patch simply reads the file specified by `user_data_file`, and does no preprocessing of that file. I suppose that since it's retrieving user data from a separate file, it's expected that the file has been automatically generated or otherwise preprocessed (as it is in my case).

Do we want Ansible/Jinja2 to be able to template-expand the external `user_data_file`? It seems to me that this is unnecessary, since allowing one to use an external file for user data gives you the ability to do even more robust YAML generation than Jinja2 templates can give you.

jsmartin:
I was more thinking of the case where my user_data is a bit more dynamic, and I want to substitute the content of some variables into the user_data.  By specifying a pre-rendered file I don't have that option.  I suppose it could just be a two-step thing though:
1. generate the user_data file from a template, using the template module,
2. launching the instance with the user_data_file=/path/to/userdata.  

This would be equally useful for the ec2_lc module.

mpdehaan:
Hi!

Thanks you very much!

Issue #8057, opened 2014-07-08
Support specifying an endpoint type in glance_image


ericwb:
##### Issue Type:

Feature Idea
##### Ansible Version:

1.6.6
##### Environment:

Ubuntu 12.04.4
##### Summary:

In our environment, the publicURL for glance is https with a self-signed certificate and the internalURL is plain http. So we require an option to specify the endpoint type to use for importing the image.
##### Steps To Reproduce:

n/a
##### Expected Results:

n/a
##### Actual Results:

n/a

jimi-c:
Merged in the above, so closing this. Thanks!

PR #8059, opened 2014-07-08
Adds an option to specify the glance endpoint type

ericwb:
Some environments that utilize an SSL terminator with a self-signed
certificate can use the publicURL without getting certificate
verify errors.  This allows using the internalURL with in my case
is HTTP and not HTTPS.

Closes issue: #8057


jimi-c:
Merged, thanks!

Issue #8070, opened 2014-07-08
mysql_db module ignoring the login_port 


wrosario:
##### Issue Type:

“Bug Report”
##### Ansible Version:

1.7
##### Environment:

Centos 6.5
##### Summary:

Running module mysql_db with login_port parameter connect to database regardless of the port being used. mysql_db is running on port 3306
##### Steps To Reproduce:

The following test:

```
- name: test mysql_db login_port param
  mysql_db: name=dbname state=present login_port=99999 
  register: result
```
##### Expected Results:

It should fail to connect to port 99999, since mysql is not running on port 99999.
##### Actual Results:

`result` give the following message: `ok: [testhost] => {
    "result": {
        "changed": true, 
        "db": "dbnmae", 
        "invocation": {
            "module_args": "name=dbname state=present login_port=99999", 
            "module_name": "mysql_db"
        }
    }
}``

jimi-c:
The above pull request has been merged in, which should resolve this issue. Thanks!

PR #8079, opened 2014-07-09
Makes ansible fail if login_port is defined and login_host is either not...

Jmainguy:
... defined, or defined as localhost. This is because if login_host is localhost then MySQLdb will use the socket instead of TCP. This leads to confusion for ansible users as, when a port is defined it gets ignored if login_host is localhost. This is to fix the bug reported by wrosario #8070. Info on MysqlDb can be read here http://mysql-python.sourceforge.net/MySQLdb.html

### DIFF
--- a/library/database/mysql_db
+++ b/library/database/mysql_db
@@ -50,7 +50,7 @@ options:
     default: localhost
   login_port:
     description:
-      - Port of the MySQL server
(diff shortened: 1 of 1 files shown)


jimi-c:
Merged, thanks!

PR #8087, opened 2014-07-10
Fix hostname when get_distribution_version() returns a string.

neersighted:
The distro version is a string ('jessie/sid') on Debian unstable and testing.
Because load_platform_subclass() evaluates every subclass as it loops,
comparing numbers when get_distribution_version() does not return a number will
break.

This patch fixes that by returning a number (`0`) when we have an invalid
version, instead of returning `None` or a string.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -188,8 +188,11 @@ def get_distribution_version():
         except:
             # FIXME: MethodMissing, I assume?
             distribution_version = platform.dist()[1]
+        if type(distribution_version) == str:
+            distribution_version = 0
     else:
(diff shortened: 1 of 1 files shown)


Jmainguy:
Nice find neersighted, was definitely a confusing issue.

mpdehaan:
I was at first a bit worried that this was used in the facts classes and we'd stop returning the actual distro version strings, but this appears to be only used in the hostname module now, so this does in fact seem reasonable.

Merging, thanks!

mpdehaan:
Per discussion on IRC, I've backed this out, we can include a more specific fix in the hostname module in case other code is using this or wants the string value.

PR #8091, opened 2014-07-10
Add 'stripext' filter plugin.

cmhamill:
Removes file extension, if any, from input string.

### DIFF
--- a/docsite/rst/playbooks_variables.rst
+++ b/docsite/rst/playbooks_variables.rst
@@ -278,6 +278,11 @@ To get the directory from a path::
 
     {{ path | dirname }}
 
+To split the file extension from a path::
+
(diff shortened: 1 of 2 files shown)


jimi-c:
You're missing a comma at the end of that line.

mpdehaan:
I'm guessing we'd rather surface splitext (the native function) and you could get the last attribute off of it.

Further, when adding a new filter plugin it does need to be documented so people can find it :)

cmhamill:
D'oh.  All better.

amenonsen:
This PR is now obsolete and can be closed (splitext is included in filters/core.py in devel and also documented).

bcoca:
closing as per above

PR #8072, opened 2014-07-08
To fix error message produced in bug report 8065

Jmainguy:

PR #8093, opened 2014-07-11
Switched primary and alias names of the name/db parameter. Fixes #8065.

### LINKED ISSUE #8065: Module mysql_db:  missing db message does not match param in docs
##### Issue Type:

“Docs Pull Request” or "“Bugfix Pull Request”
##### Ansible Version:

1.7
##### Environment:

Centos 6.5 
##### Summary:

Running module mysql_db with missing db name return error message that does not match the module documentation.
##### Steps To Reproduce:

The following test 

```
- name: test with no parameters 
  mysql_db:
  register: result
  ignore_errors: true
```

`result` the following message: `missing required arguments: db`

Looking at the documentation for mysql_db module, there is no db parameter on the docs http://docs.ansible.com/mysql_db_module.html
##### Expected Results:

Module parameters message should match the documentation online. One of the following suggestions below:
1. `result` should return:  `missing required arguments: name` 
2. update the mysql_db_module.html parameter from `name` to `db` or make it more clear that name is a db alias.
##### Actual Results:

`result` the following message:   `missing required arguments: db`

ghjm:
In the documentation, the name of the mysql database to create is called "name," with an alias of "db." However, in the code, "db" is the primary name and "name" is the alias. This causes a confusing error message. This patch updates the code to match the documentation.

### DIFF
--- a/library/database/mysql_db
+++ b/library/database/mysql_db
@@ -265,7 +265,7 @@ def main():
             login_host=dict(default="localhost"),
             login_port=dict(default="3306"),
             login_unix_socket=dict(default=None),
-            db=dict(required=True, aliases=['name']),
+            name=dict(required=True, aliases=['db']),
             encoding=dict(default=""),
             collation=dict(default=""),
             target=dict(default=None),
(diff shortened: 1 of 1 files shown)


ghjm:
This is a near duplicate of #8072.

Jmainguy:
This fix should work, for #8065 . The module now matches the doc's in so far as db is an alias of name now, and not the other way around. Error messages will be less confusing once this is merged in.

jimi-c:
Merged, thanks!

PR #8094, opened 2014-07-11
Fixed tracebacks during VM creation. Fixes #8067.

### LINKED ISSUE #8067: vsphere module not finding datacenter
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.7 (devel 366b39cfc1) last updated 2014/07/07 10:46:37 (GMT -700)
##### Environment:

OSX 10.9.3 
##### Summary:

The goal is to create a VM on a large vSphere cluster, in a specific datacenter, in a specific folder.
##### Steps To Reproduce:
- using the Playbook below -
- manually create a VM on vSphere called “ansibletestvm”
- run playbook
- expected results = VM already exists & ansible playbook reports no changes and sucessfully completes the run
- edit playbook - set “guest: ansiblecreatevm”
- run playbook
##### Expected Results:
1. Ansible verifies VM does not exists
2. Ansible creates VM on vSphere cluster
##### Actual Results:
1. Ansible errors out with following:

<localhost> REMOTE_MODULE vsphere_guest password=VALUE_HIDDEN state=present guest=ansiblecreatevm vcenter_hostname=prdasvmw02.gsm1900.org username=Cliverm
<localhost> EXEC ['/bin/sh', '-c', 'mkdir -p $HOME/.ansible/tmp/ansible-tmp-1404760410.46-75812875617325 && chmod a+rx $HOME/.ansible/tmp/ansible-tmp-1404760410.46-75812875617325 && echo $HOME/.ansible/tmp/ansible-tmp-1404760410.46-75812875617325']
<localhost> PUT /var/folders/3c/lxfy48s918zdv2dk70qpgtlm0000gp/T/tmp887Hcg TO /Users/chrisl/.ansible/tmp/ansible-tmp-1404760410.46-75812875617325/vsphere_guest
<localhost> EXEC ['/bin/sh', '-c', u'LANG=en_US.UTF-8 LC_CTYPE=en_US.UTF-8 /usr/bin/python /Users/chrisl/.ansible/tmp/ansible-tmp-1404760410.46-75812875617325/vsphere_guest; rm -rf /Users/chrisl/.ansible/tmp/ansible-tmp-1404760410.46-75812875617325/ >/dev/null 2>&1']
  File "/Users/chrisl/.ansible/tmp/ansible-tmp-1404760410.46-75812875617325/vsphere_guest", line 2515, in <module>
    main()
  File "/Users/chrisl/.ansible/tmp/ansible-tmp-1404760410.46-75812875617325/vsphere_guest", line 1203, in main
    state=state
  File "/Users/chrisl/.ansible/tmp/ansible-tmp-1404760410.46-75812875617325/vsphere_guest", line 602, in create_vm
    v in vsphere_client.get_datacenters().items() if v == datacenter][0]
IndexError: list index out of range

ghjm:
This patch fixes three tracebacks that can occur when creating a vSphere virtual machine:
1. If the datacenter name in the playbook does not exist on the vCenter server, the list comprehension gives an empty result and dereferencing to [0] fails. Instead of checking if [0]==None, we check if the list has any members. If it does, then we can use [0] safely.
2. If the playbook does not supply a value for the notes parameter in vm_extra_config, this causes a traceback when we try to compare it to None. Instead, we should check if the "notes" key is present in the dictionary. If it is present and also has the value None, then we wind up setting the annotation to None, which seems like reasonable behavior.
3. If vm_extra_config is not supplied and the power state is not requested to be on or off, then we are not retrieving the vm from pysphere, which causes a traceback when we try to gather facts about it at the end of the module. Since we always want to gather these facts, we always need the vm reference.

### DIFF
--- a/library/cloud/vsphere_guest
+++ b/library/cloud/vsphere_guest
@@ -604,10 +604,11 @@ def create_vm(vsphere_client, module, esxi, resource_pool, cluster_name, guest, 
     datacenter = esxi['datacenter']
     esxi_hostname = esxi['hostname']
     # Datacenter managed object reference
-    dcmor = [k for k,
-             v in vsphere_client.get_datacenters().items() if v == datacenter][0]
-
-    if dcmor is None:
+    dclist = [k for k,
+             v in vsphere_client.get_datacenters().items() if v == datacenter]
+    if dclist:
+        dcmor=dclist[0]
+    else:
         vsphere_client.disconnect()
         module.fail_json(msg="Cannot find datacenter named: %s" % datacenter)
 
@@ -710,7 +711,7 @@ def create_vm(vsphere_client, module, esxi, resource_pool, cluster_name, guest, 
     vmfiles.set_element_vmPathName(datastore_name)
     config.set_element_files(vmfiles)
     config.set_element_name(guest)
-    if vm_extra_config['notes'] is not None:
+    if 'notes' in vm_extra_config:
         config.set_element_annotation(vm_extra_config['notes'])
     config.set_element_memoryMB(int(vm_hardware['memory_mb']))
     config.set_element_numCPUs(int(vm_hardware['num_cpus']))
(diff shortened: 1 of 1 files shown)


jimi-c:
Merged, thanks!

Issue #6214, opened 2014-02-27
mysql_db import action fails with confusing error when database does not exist


packetcollision:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.4.4
##### Environment:

N/A (but I'm on a mac, trying to send commands to a machine running Ubuntu 12.04)
##### Summary:

If you try to run "mysql_db: name=nonexistantdb state=import target=/foo.sql" and the database doesn't exist, the command fails with "unable to connect, check login_user and login_password are correct, or alternatively check ~/.my.cnf contains credentials"
##### Steps To Reproduce:

run "mysql_db: name=nonexistantdb state=import target=/foo.sql" when the database does not exist. Make sure to use the right credentials (or have them specified in .my.cnf)
##### Expected Results:

It really should output the actual SQL error, which is: 1049, "Unknown database 'nonexistantdb'"
##### Actual Results:

the command fails with "unable to connect, check login_user and login_password are correct, or alternatively check ~/.my.cnf contains credentials"

DSpeichert:
I just had the same problem, I think the database should be created automatically.
Ubuntu 14.04 / Ansible 1.5.3-trusty-unstable-ppa2

josef-spak:
I can reproduce this exactly, with the latest dev version as well.

ansible-playbook 1.6 (devel 3b285d736c)

Jmainguy:
The reason it is failing is because it uses the database name to connect with initially, so if it doesn't exist, it cant work. 

That being said the code could be changed to connect to to db=mysql in the event of state=import and database does not exist. If this seems like a good idea I can take a stab at adding it in.

varas:
Same issue: 
- Version: 1.5.5
- Modules: mysql_user, mysql_db
- Target: ubuntu 14.04

Msg:

```
unable to connect to database, check login_user and login_password are correct or ~/.my.cnf has the credentials
```

Jmainguy:
Added error message for when db does not exist to import. #8103

jimi-c:
The above PR has been merged in, which should correct this. Thanks!

PR #8103, opened 2014-07-11
mysql_db: Fix error message for #6214

Jmainguy:
Errors out with stderr 1049, unknown database: when importing a database that does not exist


jimi-c:
Merged, thanks!

PR #8104, opened 2014-07-11
Correct regex in is_keyid

AdmiralNemo:
The `rpm_key` module claims to take a key ID as an argument, however it fails if one is actually given:

```
msg: Not a valid key DEADBEEF
```

The reason for this is the regular expression in the `RpmKey.is_keyid` method is incorrect; it will only match the literal string "0-9a-f" exactly 8 times, rather than exactly 8 characters in the set 0-9 or a-f.


jimi-c:
Merged, thanks!

Issue #8108, opened 2014-07-12
Pip module: problem with float version


izidormatusov:
This task:

``` yaml
- name: Install supervisord
  pip:
    name: supervisord
    version: 3.0
```

fails:

```
invalid output was: SUDO-SUCCESS-apaeftrvqreqvqairirxfsgrskgujukv
Traceback (most recent call last):
  File "/home/ubuntu/.ansible/tmp/ansible-tmp-1405194864.4-214406963454158/pip", line 1541, in <module>
    main()
  File "/home/ubuntu/.ansible/tmp/ansible-tmp-1405194864.4-214406963454158/pip", line 295, in main
    cmd += ' %s' % _get_full_name(name, version)
  File "/home/ubuntu/.ansible/tmp/ansible-tmp-1405194864.4-214406963454158/pip", line 153, in _get_full_name
    resp = name + '==' + version
TypeError: cannot concatenate 'str' and 'float' objects
```

Apparently version is treated as float instead of string. Surrounding `3.0` with quotes `"3.0"` solves this.

I think ansible should convert version to string although yaml interprets it as a float.

ansibot:
Thanks for filing a ticket! I am the friendly GitHub Ansibot. I see you did not fill out your issue description based on our new issue template. Please copy the contents of this [template](https://raw2.github.com/ansible/ansible/devel/examples/issues/ISSUE_TEMPLATE.md) and **paste it into the description** of your ticket. 

Our system will automatically close tickets that do not have an issue template filled out within 7 days. Please note that due to large interest in Ansible, humans may not comment on your ticket if you ask them questions. Don't worry, you're still in the queue and the robots are looking after you.

mpdehaan:
closing per above fix, thank you!

PR #8111, opened 2014-07-13
Fix pip failure when version is parsed as a float

hfinucane:
##### Summary:

See #8108 for more details. Only applies if you're using YAML to specify module arguments instead of the more customary `k=v` format.
##### Steps To Reproduce:

```
- name: Install supervisor
  pip:
    name: supervisor
    version: 3.0
```
##### Expected Results:

`supervisord` should get installed
##### Actual Results:

```
line 153, in _get_full_name
    resp = name + '==' + version
TypeError: cannot concatenate 'str' and 'float' objects
```

### DIFF
--- a/library/packaging/pip
+++ b/library/packaging/pip
@@ -206,7 +206,7 @@ def main():
         argument_spec=dict(
             state=dict(default='present', choices=state_map.keys()),
             name=dict(default=None, required=False),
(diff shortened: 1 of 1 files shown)


mpdehaan:
Merged, thanks!

PR #8113, opened 2014-07-13
Enhance zypper repo module: Changed the repo_exists function to respect .repo endings.

xenji:
Changed the `repo_exists` function to respect `.repo` endings.
This is important, as a repo url is valid when ending on .repo instead of a slash.
The check of the existance of the .repo url in the output of
`zypper lr --uri`
would always fail, as `--uri` just shows the url variant with the trailing slash.

### DIFF
--- a/library/packaging/zypper_repository
+++ b/library/packaging/zypper_repository
@@ -95,7 +95,7 @@ def add_repo(module, repo, alias, description, disable_gpg_check):
 
     rc, stdout, stderr = module.run_command(cmd, check_rc=False)
     changed = rc == 0
-    return (rc, stdout, stderr, changed)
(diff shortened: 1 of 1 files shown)


mpdehaan:
A few things

Contributions very welcome, but this is probably not a large enough change to change the module copyright.

Change list() to []

replacing name with local_name seems not needed since there wasn't a scope conflict.

Otherwise looks good, so safe to cherry-pick...

xenji:
Code changes are on their way. Regarding the copyright: I personally don't care if my name is in there or not, I just wanted to state the change made to the original file. If this is not wanted, I'll remove that.

xenji:
I'll close those two requests and open new ones with the squashed commits

PR #8114, opened 2014-07-13
Enhance zypper module: two new options (refresh and without_recommens)

xenji:
Added refresh option for refreshing the repo meta data before ensuring the state.
Added option to skip recommends from zypper.
Added corresponding documentation and some examples.

Changed some of the variable names to be more self explanatory, made some changes according to PEP8.

### DIFF
--- a/library/packaging/zypper
+++ b/library/packaging/zypper
@@ -26,8 +26,6 @@
 # You should have received a copy of the GNU General Public License
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 
-import re
-
 DOCUMENTATION = '''
 ---
 module: zypper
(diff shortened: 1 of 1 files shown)


mpdehaan:
Can you please submit these without the variable renames and also add the "version_added" attribute to indicate these options are new in 1.7?

Squash commit would also make review easier if you don't mind.

Thanks!

xenji:
Sure I'll add that to the docs and squash it down to one commit.

xenji:
I'll close those two requests and open new ones with the squashed commits

alxgu:
For information, there is already a pull request to skip recommends: https://github.com/ansible/ansible/pull/7379

PR #8119, opened 2014-07-14
Fix indentation in ec2.py

adamchainz:
Just tried to use this file and it didn't work due to a space missing at the start of `get_rds_instances_by_region`.


cchurch:
It's really a tab instead of 8 spaces; it only gave me an error if I ran `python -tt ec2.py`.  

It appears to have been there since the original rds commit: dd2d0dd175906624f14efedaa1eeb5012c7d0655.

Thanks for the fix!

PR #8122, opened 2014-07-14
Docs: fix the name of the package that provides apt-add-repository on Ubuntu

tgerla:
Quick docs fix. The name of the package that provides "apt-add-respository" is not "apt-add-repository", but instead either "software-properties-common" or "python-software-properties", depending on what version of Ubuntu you're on.


jimi-c:
Merged, thanks!

PR #8124, opened 2014-07-14
Added setting of the health_check options on an existing ASG

ferrants:
The module will return with these values and will set them if they change.


ferrants:
`health_check_period` and `health_check_type` were added recently, this is more of a bug fix for them than anything else

jimi-c:
Merged, thanks!

PR #8126, opened 2014-07-14
Update lineinfile validate help

dingus9:
It wasn't clear that validate requires %s in your command string. I spent a couple minutes debugging before I noticed the %s in the example validate command.


jimi-c:
Merged, thanks!

PR #8127, opened 2014-07-14
Allow relative paths for jinja2 template extend statements

DavidMikeSimon:
A template cannot extend from another template that in turn extends from _another_ template using a relative path; ansible will assume all relative paths are relative to the first template, breaking the second reference.

This PR adds a test which exhibits this bug and a small fix (based on http://stackoverflow.com/a/8530761/351149) that corrects it.

### DIFF
--- a/lib/ansible/plugins/action/template.py
+++ b/lib/ansible/plugins/action/template.py
@@ -25,7 +25,8 @@ import time
 from ansible import constants as C
 from ansible.plugins.action import ActionBase
 from ansible.utils.hashing import checksum_s
-from ansible.utils.unicode import to_bytes, to_unicode
+from ansible.utils.unicode import to_bytes
+from ansible.template import Templar
 
 class ActionModule(ActionBase):
 
@@ -84,9 +85,6 @@ class ActionModule(ActionBase):
 
         # template the source data locally & get ready to transfer
         try:
-            with open(source, 'r') as f:
-                template_data = to_unicode(f.read())
-
             try:
                 template_uid = pwd.getpwuid(os.stat(source).st_uid).pw_name
             except:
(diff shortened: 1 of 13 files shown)


amenonsen:
@DavidMikeSimon, do you feel like rebasing this PR onto devel? Your change looks reasonable.

DavidMikeSimon:
@amenonsen Sure, I should be able to get to it this week.

DavidMikeSimon:
@amenonsen Okay, it's working again. The changes had to go a little deeper this time around; it looks like at some point during the transition to v2, ansible lost the ability to do j2 relative extends even _one_ level deep. To correct this, I had to change Templar to allow callers to provide a file path as an alternative to the raw template data.

DavidMikeSimon:
@amenonsen Bump. Anything else I need to do on this PR?

jimi-c:
Hi @DavidMikeSimon, not sure if we want to merge your PR in exactly as it is, however I am interested in fixing the bug you've demonstrated at the very least so we'll hold this open. Also, don't worry about rebasing this again, as we've already asked for it once without merging it.

Thanks!

jimi-c:
Also also, sorry it's taken us so long to follow up on this!

bcoca:
so in current code the extends works with relative paths as long as they are relative to the searchpath.

I can fix the tests to work with just adding `{% extends 'base/grandparent.j2' %}` to parent.j2

bcoca:
@DavidMikeSimon can you resubmit just the tests (modified as i pointed out above)?

mattclay:
The integration tests should be moved from `test/integration/roles` to `test/integration/targets`.

Also, create a `test/integration/targets/test_template_relative_paths/aliases` file with the contents:

```
posix/ci/group2
``` 

That will cause the tests to run as part of the 2nd CI group on Shippable.

DavidMikeSimon:
@bcoca @mattclay No problem, I'll be able to get to this tomorrow.

DavidMikeSimon:
@bcoca On closer inspection, this is a little tricky because changing the line in `parent.j2` to `{% extends 'base/grandparent.j2' %}` would also have caused the test to pass even under the effect of the bug I was attempting to fix in the first place. :-\

If nowadays extends statements in jinja2 templates in ansible are always based on the search_path and ignore the location of the template file itself, then my original bug report isn't relevant anymore. This behavior also lines up with the jinja2 spec, which allows the application to determine the meaning of the argument passed to `extends`. I'll just go ahead and close this issue.

DavidMikeSimon:
However, see #17960

PR #8130, opened 2014-07-14
New option for route53: retry_interval

napkindrawing:
##### Issue Type

Feature Pull Request
##### Ansible Version:

ansible 1.6.6

But affects devel branch as well
##### Environment:

OSX 10.9.2
##### Summary:

The route53 module has a default, hard-coded, retry interval of 500 seconds (8 min 20sec).

If either Route53 is having a bad day, or you have many domains to manage, you're almost guaranteed to experience one or more 8+min retry intervals, which can really add up and make ansible take forever to run.
##### Steps To Reproduce:

Use the route53 module with more than a dozen or so domains and you're more than likely to experience one or more 8+ minute retry interval. 

I'm currently booting up a cluster of 15 machines, each with 2 Route53 records, and runtime is usually 20+ minutes just for the route53 part.
##### Expected Results:

I'd like to be able to customize the retry interval when managing many domains. For this patch, 15 seconds worked quite nicely.
##### Actual Results:

A very long wait. My beard got longer. I should have gone and made a snack or called my mom.

### DIFF
--- a/library/cloud/route53
+++ b/library/cloud/route53
@@ -78,6 +78,12 @@ options:
     required: false
     default: null
     aliases: []
+  retry_interval:
+    description:
+      - In the case that route53 is still servicing a prior request, this module will wait and try again after this many seconds. If you have many domain names, the default of 500 seconds may be too long.
+    required: false
+    default: 500
+    aliases: []
 requirements: [ "boto" ]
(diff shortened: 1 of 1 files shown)


jimi-c:
Merged, thanks!

PR #8134, opened 2014-07-14
Set Jinja2's keep_trailing_newline=True in template_from_string.

leverich:
Without this setting, the newline idempotence of scalar strings changes depending on whether
or not they contain any Jinja2 template blocks.

I've included a patch to the test_template integration test to demonstrate the behavior, but to be explicit, consider the following vars:

```
vars:
  x: "Hello\n"
  y: "{{'Hello'}}\n"
  z: "{{'Hello'}}\n\n"
```

Without keep_trailing_newline, x != y, but instead x == z, as Jinja2 strips one trailing newline on y and z. This only affects string templates, as file templates already have code to ensure newline parity.

I expect that newlines should be handled identically for strings, regardless of whether or not a template is used.

Not limited to vars, this affects other places where templates may be used; for example, copy:contents.

### DIFF
--- a/lib/ansible/utils/template.py
+++ b/lib/ansible/utils/template.py
@@ -322,7 +322,7 @@ def template_from_string(basedir, data, vars, fail_on_undefined=False):
         def my_finalize(thing):
             return thing if thing is not None else ''
 
-        environment = jinja2.Environment(trim_blocks=True, undefined=StrictUndefined, extensions=_get_extensions(), finalize=my_finalize)
+        environment = jinja2.Environment(trim_blocks=True, keep_trailing_newline=True, undefined=StrictUndefined, extensions=_get_extensions(), finalize=my_finalize)
         environment.filters.update(_get_filters())
         environment.template_class = J2Template
 
--- a/test/integration/roles/test_template/tasks/main.yml
+++ b/test/integration/roles/test_template/tasks/main.yml
@@ -63,4 +63,10 @@
   assert: 
     that: 
       - "file_result.changed != True"
-        
+
+# VERIFY STRING TEMPLATE PRESERVES NEWLINE
+
+- name: ensure template strings preserve newline
+  assert:
+    that:
+      - "newline_no_template_string == newline_template_string"
--- a/test/integration/roles/test_template/vars/main.yml
+++ b/test/integration/roles/test_template/vars/main.yml
@@ -1,1 +1,3 @@
 templated_var: templated_var_loaded
+newline_template_string: "Hello {{ 'World' }}\n"
+newline_no_template_string: "Hello World\n"


jimi-c:
Merged, thanks!

jimi-c:
@leverich hey, we had to revert this, as apparently the keep_trailing_newline option is only available in certain versions of jinja2 (not backported to older Enterprise Linuxes apparently). The traceback we saw reported was:

<pre>
  File "/usr/bin/ansible-playbook", line 309, in <module>
    sys.exit(main(sys.argv[1:]))
  File "/usr/bin/ansible-playbook", line 249, in main
    pb.run()
  File "/usr/lib/pymodules/python2.7/ansible/playbook/__init__.py", line 293, in run
    play = Play(self, play_ds, play_basedir, vault_password=self.vault_password)
  File "/usr/lib/pymodules/python2.7/ansible/playbook/play.py", line 152, in __init__
    self._tasks      = self._load_tasks(self._ds.get('tasks', []), load_vars)
  File "/usr/lib/pymodules/python2.7/ansible/playbook/play.py", line 595, in _load_tasks
    role_name=role_name
  File "/usr/lib/pymodules/python2.7/ansible/playbook/task.py", line 211, in __init__
    self.async_seconds = template.template_from_string(play.basedir, self.async_seconds, self.module_vars)
  File "/usr/lib/pymodules/python2.7/ansible/utils/template.py", line 325, in template_from_string
    environment = jinja2.Environment(trim_blocks=True, keep_trailing_newline=True, undefined=StrictUndefined, extensions=_get_extensions(), finalize=my_finalize)
TypeError: __init__() got an unexpected keyword argument 'keep_trailing_newline'
</pre>

So if you can resubmit this with a check for the jinja2 version to ensure no backwards compatibility breakages, we'd be happy to review it again. Thanks!

leverich:
Bummer! I can add a version check. The downside is that Jinja2 <= 2.6 will still have the wrong newline behavior, and the integration test will need a version check as well.

Another option is to hack around the problem by appending newlines if any are stripped from the end by Jinja2 (as done in template_from_file):

```
# Preserve newlines. Copied from template_from_file.
res_newlines  = count_newlines_from_end(res)
data_newlines = count_newlines_from_end(data)
if data_newlines > res_newlines:
    res += '\n' * (data_newlines - res_newlines)
```

This passes all non_destructive integration tests with new and old versions of Jinja2.

Any preference?

bcoca:
Can't this just be added to ansible.cfg jinja defaults entry?  (for those
that can use it)

## 

Brian Coca
Pedo mellon a minno

jimi-c:
@bcoca that is a possibility, however a version check would still be required to prevent errors for those that enable it without having the current version.

leverich:
AFAIK, the existing ansible.cfg setting for jinja specifies which
extensions to load.

I suppose the real decision is whether or not the current behavior (trim
newlines for templated strings but not for plain strings) should be
preserved.

If so, perhaps there could be a new "jinja2_string_settings" in
ansible.cfg, with functionality similar to the jinja2 overrides in
template_from_file.

If, instead, the current behavior is a bug (it's unexpected, at the very
least), we should make sure the newline doesn't get trimmed, and maybe add
a configuration option to preserve the bad behavior.

A third option is to keep adding '\n\n' to the end of template strings :-)

bennojoy:
if i am not mistaken we do allow to load jinja environrment overrides per template

so adding the below header 
#jinja2:keep_trailing_newline=True

in the template should set the jinja env variable.

leverich:
@bennojoy indeed, but that's only for templates loaded from files. The issue here pertains to newlines in YAML strings processed by template_from_string, for which there is no way to control jinja2's behavior.

Just to reiterate, here's a playbook that exhibits the problem. The first assertion succeeds and the second fails.

```
---
- hosts: all
  vars:
    x: "Hello World\n"
    y: "{{'Hello'}} World\n"
    z: "{{'Hello'}} World\n\n"
  tasks:
    - assert:
        that: "x == z"
    - assert:
        that: "x == y"
```

@jimi-c I think the most straight-forward solution is for template_from_string to exactly match template_from_file's newline handling, which explicitly counts trailing newlines and adds any back if necessary. This gets around the issue of keep_trailing_newline not being available in all versions of Jinja2. I'm happy to submit a new pull request with that strategy.

Cheers

manuelmeurer:
Would love to see this included as well! The current behavior is not intuitive IMO.
How can `keep_trailing_newline` be enabled in `ansible.cfg` for now?

abadger:
Just a note -- we still don't have keep_trailing_newline due to library versions but we have implemented the workaround and (in devel which will become 2.0) we think we are keeping trailing newlines in all instances.

Issue #8102, opened 2014-07-11
hostname module failing on CentOS 7 (released)


phips:
##### Issue Type:

Bug Report
##### Ansible Version:

```
$ ansible --version
ansible 1.7 (devel fbdb10d8a5) last updated 2014/07/11 17:14:11 (GMT +100)
```
##### Environment:

Control host: OSX 10.9.4
Target host: CentOS 7 VM

```
$ ansible -ki hosts -m setup centos7 -a 'filter=ansible_distribution_\*'
centos7 | success >> {
    "ansible_facts": {
        "ansible_distribution_major_version": "7", 
        "ansible_distribution_release": "Core", 
        "ansible_distribution_version": "7.0.1406"
    }, 
    "changed": false
}
```
##### Summary:

Hostname module fails with a traceback (see below).
##### Steps To Reproduce:

See _Actual Results_.
##### Expected Results:

```
TASK: [hosts | Ensure hostname set] ******************************************* 
changed: [default]
```
##### Actual Results:

```
TASK: [hosts | Ensure hostname set] ******************************************* 
failed: [centos7] => {"failed": true, "parsed": false}
invalid output was: Traceback (most recent call last):
  File "<stdin>", line 1597, in <module>
  File "<stdin>", line 1600, in RedHatServerHostname
ValueError: invalid literal for float(): 7.0.1406


FATAL: all hosts have already failed -- aborting
```

phips:
OK, it would appear the version number of 7.0 is 7.0.1406, which is returned from platform.linux_distribution()[1]

'float' doesn't handle that so well. Could it be fixed by stripping off the last part of the version number, giving 7.0? Would that have lots of bad side effects?

jimi-c:
I think that'd be fine, since we're primarily concerned with just the major/minor number - the revision or whatever the rest of that would be considered does not need to be preserved.

vincentvdk:
Not sure if it's a good idea to completely drop it. Maybe you want to target machines at a certain patch level. Until now we had point release (6.4, 6.5, ..), but now you can be more fine grained if needed.

phips:
Looking through the code though @vincentvdk this routine (in basic.py) isn't used in any other module, currently. Which implies nobody cares about a sub release number. Plus, if we do take it account, then the code needs changing considerably as float() isn't going to work.

I started to look at patching it yesterday. Got the 7.0 captured just fine, but for some reason it still wasn't working (different error). Am working through it.

phips:
[...]

PR #8135, opened 2014-07-14
fix issue #8102

rmarchei:
fix issue #8102

### DIFF
--- a/library/system/hostname
+++ b/library/system/hostname
@@ -39,6 +39,8 @@ EXAMPLES = '''
 - hostname: name=web01
 '''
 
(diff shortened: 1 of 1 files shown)


jimi-c:
This looks fine, the only thing I dislike is the fact that we're having to create distro and strategy classes for every variation in each distro's name. We should have some kind of mapping table to clean this up and reduce it to only having one class per hostname management method. If you'd like to work on that, I'd appreciate it :)

jimi-c:
Merged, thanks!

rmarchei:
I agree with you, I'll work on that!

PR #8138, opened 2014-07-15
added environment variables as config overrides for vmware inventory

bcoca:
so now this can be configured through exporting env variables like other inventory scripts.

### DIFF
--- a/plugins/inventory/ec2.py
+++ b/plugins/inventory/ec2.py
@@ -240,7 +240,7 @@ class Ec2Inventory(object):
         self.cache_path_cache = cache_dir + "/ansible-ec2.cache"
         self.cache_path_index = cache_dir + "/ansible-ec2.index"
         self.cache_max_age = config.getint('ec2', 'cache_max_age')
-        
+
 
 
     def parse_cli_args(self):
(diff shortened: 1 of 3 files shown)


mpdehaan:
We do need env var configuration but it is very important that this be a predictable variable name (particularly for tower).  Could this be VMWARE_USER, etc?

mpdehaan:
cc @lukesneeringer

bcoca:
I can add both, since i need multiple configs with diff users/servers I
need the 'variety'. I'll update PR tonight

## 

Brian Coca
Pedo mellon a minno

lukesneeringer:
Hi @bcoca,
First of all, thank you for your patch! :-)

I'm thinking through the thing with the environment variable names, and something still doesn't sit right with me about the current implementation. Having the script name on disk be used for the environment variables, while this is done sometimes (e.g. for bash autocomplete), is unusual. I guess I don't mind that much, since 99% of users will just use `VMWARE_*`, but as it stands I'm worried about getting in a weird situation where a variable "magically" takes and the user doesn't know why.

My original thought was to ask you to change the precedence order, so we check `VMWARE_*` first, then we check `{scriptname}-*` second.

After thinking about it more, though, I think what I'd prefer would be that if you need the multiple environment variables, that they be faux-namespaced one more level.

So:
- first, check `VMWARE_{scriptname}_HOST`
- second, check `VMWARE_HOST`
- third, check the INI file

Essentially, I'm asking you to change `{scriptname}_HOST` to `VMWARE_{scriptname}_HOST` (and same for `USER` and `PASSWORD`), and leave everything else about your implementation the same. Essentially that makes it so that there's really no way that anyone stumbles on that by accident. It also makes it very clear if we have a situation where we apply similar functionality to other dynamic inventory scripts. (I'm not sure whether we will do that, but I want to be thinking as if we might.)

Would that be okay with you?

bcoca:
no problem, I'm just a lazy typist, I'll push and update tonight.​

PR #8139, opened 2014-07-15
Fix "TypeError: group id must be integer" if groupid passed in for group key

peterkh:
Hi,
Came across this small issue today.

Triggered by:
- user: name='peterkh' comment='Peter Hall' shell='/bin/bash' uid=600 group=600 home=/home/peterkh createhome=False

Cheers
Peter


mpdehaan:
merged, thanks!

PR #8143, opened 2014-07-15
portage: default exit (not fail) if sync=yes in check mode

PinGwynn:
Exit by default in check mode, if sync=yes


mpdehaan:
Merged, thanks, this should probably also return changed: True.

PR #8145, opened 2014-07-15
Fix EC2_TAG module output structure when listing tags

amandolo:
Fix EC2_TAG module output structure when listing tags


mpdehaan:
Not many people using state == list so I imagine playbook changes would be minimal.

I agree these should be structured.

Merging, thanks!

PR #8146, opened 2014-07-15
fix get_distribution_version() on amazon linux

rmarchei:
fixes hostname change on amazon linux (issue #8102)


jimi-c:
Merged, thanks!

Issue #8011, opened 2014-07-01
docker_image module object has no attribute APIError


zjeraar:
##### Issue Type:

Bug Report
##### Ansible Version:

1.6.5
##### Environment:

control OS: Ubuntu 14.04
managed OS: Ubuntu 14.04
##### Summary:

The docker_image module gives me an error every time I try to run it, no matter what state I use.
##### Steps To Reproduce:

docker_image:
    name='user/app'
    tag='1.0'
    path='/build/path'
    state=build
##### Expected Results:

A user/app image, with no error reported back.
##### Actual Results:

The build itself actually works (image is present afterwards), but I get the following error:

fatal: [my.server.com] => failed to parse: SUDO-SUCCESS-zifpeodwifbudvwfyirskzschingrdvb
{"msg": "Image builded: 8b63a1d631f0", "failed": false, "changed": true, "image_id": "8b63a1d631f0"}
Traceback (most recent call last):
  File "/home/tds/.ansible/tmp/ansible-tmp-1404244300.34-22987635386790/docker_image", line 1455, in <module>
    main()
  File "/home/tds/.ansible/tmp/ansible-tmp-1404244300.34-22987635386790/docker_image", line 229, in main
    except docker.client.APIError as e:
AttributeError: 'module' object has no attribute 'APIError'

myhro:
Experienced something like this, except that I couldn't build the image. Downgrading to `docker-py==0.3.1` solved the problem.

jimi-c:
We've had a couple of these, due to the docker API changing. Have you tested this against devel to see if this was one of the errors we've fixed already?

zjeraar:
No I did not test it on devel, will try asap.

fangpenlin:
got the same issue here

prune998:
same issue. While the APIError exception not existing is an error, the exception is triggered in the create_containers() function.
digging in it right now.

jimi-c:
I believe this was fixed here: https://github.com/ansible/ansible/pull/7029

prune998:
docker version

```
Client version: 1.0.1
Client API version: 1.12
Go version (client): go1.2.1
Git commit (client): 990021a
Server version: 1.0.1
Server API version: 1.12
Go version (server): go1.2.1
Git commit (server): 990021a
```

Docker-py version : 0.3.2
Server version : ubuntu 14.04 :

```
[...]

PR #8150, opened 2014-07-15
removed privileged option on container creation

prune998:
the privilege option is not present in Docker-py code so this ansible module is not working anymore.

FYI, in docker-py, as of version 0.3.2 : 

```
    def create_container(self, image, command=None, hostname=None, user=None,
                         detach=False, stdin_open=False, tty=False,
                         mem_limit=0, ports=None, environment=None, dns=None,
                         volumes=None, volumes_from=None,
                         network_disabled=False, name=None, entrypoint=None,
                         cpu_shares=None, working_dir=None, domainname=None,
                         memswap_limit=0):
```

### DIFF
--- a/library/cloud/docker
+++ b/library/cloud/docker
@@ -588,7 +588,6 @@ class DockerManager:
             'binds': self.binds,
             'port_bindings': self.port_bindings,
             'publish_all_ports': self.module.params.get('publish_all_ports'),
-            'privileged':   self.module.params.get('privileged'),
             'links': self.links,
         }
         if docker.utils.compare_version('1.10', self.client.version()['ApiVersion']) >= 0 and hasattr(docker, '__version__') and docker.__version__ > '0.3.0':


prune998:
As this is a breaking functionality bug, I suggest merging it to dev AND to latest release, or schedule a new release soon.
this is related to issue https://github.com/ansible/ansible/issues/8011

jimi-c:
The above code params are not used in create_container() - they're used in client.start(), which looks like this in docker-py 0.3.2:

<pre>
    def start(self, container, binds=None, port_bindings=None, lxc_conf=None,
              publish_all_ports=False, links=None, privileged=False,
              dns=None, dns_search=None, volumes_from=None, network_mode=None):
</pre>

So I'm unclear as to why this would be throwing an error for you.

Finally, if this is a param that was introduced or removed at a specific version of docker-py, you will need to check that as other params are in that function, rather than omitting it completely so that users on other versions do not have their functionality broken. Thanks!

PR #8154, opened 2014-07-15
Add new module postgresql_exec

jirutka:
This module is intended for initializing database schema and maybe populating with some seed data from a SQL script. However, it can be used to execute any SQL commands. It is up to the user to maintain idempotence.

### DIFF
--- a/lib/ansible/runner/action_plugins/postgresql_exec.py
+++ b/lib/ansible/runner/action_plugins/postgresql_exec.py
@@ -1,0 +1,84 @@
+# -*- coding: utf-8 -*-
+# (c) 2014, Jakub Jirutka <jakub@jirutka.cz>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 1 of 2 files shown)


mpdehaan:
Thanks for the submission.

Let's discuss this one on ansible-devel.   I'm unclear why someone wouldn't just shell into the psql command (or use the script module) and I'm not wanting to expand action_plugin coverage if possible.

We may be able to come to a reasonable conclusion on an alternative approach, or it may make sense still as just a local module with a preceeding 'copy'.

jirutka:
I’ve used to do this with the `shell` module and `psql`, but it’s quite ugly and doesn’t support check mode (rollback). When the script file is on a local machine (in a role), you must also copy it to a remote machine and then delete it (i.e. two extra tasks).

``` yml
- name: initialize database
  shell: >
    PGPASSWORD='{{ roundcube_db_password }}'
    psql -h {{ roundcube_db_host }}
    -p {{ roundcube_db_port }}
    -U {{ roundcube_db_username }}
    -d {{ roundcube_db_database }}
    -f {{ roundcube_www_dir }}/SQL/postgres.initial.sql
    --set ON_ERROR_STOP=1
```

jnv:
`mysql_db` solves the initial schema import through the `import` state and there is already an outstanding PR to add the same functionality to `postgresql_db`: #6825.

PR #8157, opened 2014-07-16
Added a vm_extra_config option for the folder the VM should be created in

stewrutledge:
Added a section in vm_extra_config (folder) which allows you to specify which vm folder the new guest should be created in.

First there is a check that the folder exists on the system, failing if it doesn't.

If it does exist, the vmfmor variable is set to the first key (I am not sure how this behaves if there are multiple folders with the same name) found in the dictionary returned by _get_managed_objects.

If it is not set, it just goes to the current default (the root folder on the datacenter/host).

```
    args:
        vm_extra_config:
          vcpu.hotadd: yes
          mem.hotadd:  yes
          notes: This is a test VM
          folder: AutoDeploy
```

### DIFF
--- a/lib/ansible/module_utils/splitter.py
+++ b/lib/ansible/module_utils/splitter.py
@@ -76,7 +76,7 @@ def split_args(args):
         do_decode = True
     except UnicodeDecodeError:
         do_decode = False
-    tokens = args.split(' ')
+    items = args.split(' ')
 
     # iterate over the tokens, and reassemble any that may have been
     # split on a space inside a jinja2 block.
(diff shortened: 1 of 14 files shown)


mpdehaan:
This seems reasonable but I'd like to see the new option referenced in examples in the DOCUMENTATION string, else folks aren't going to find out that this option exists.

Can you update the DOCUMENTATION part above in the same file to show an example?

stewrutledge:
Updated the documentation to show the option under extra_config.

I think this module could use a bit more documentation in general, that'll have to be another pull request in the future.

stewrutledge:
Closing as this was made on my local devel branch.

Re-opened with own branch here:
https://github.com/ansible/ansible/pull/8451

PR #8165, opened 2014-07-16
Moved to absolute path for playbook_path... and playbook.basedir

dingus9:
The goal is to mirror the absolute path like what is in {{inventory_dir}} with {{playbook_dir}} instead of the relative path usually ".". I just wrapped the two primary places based_dir is used in os.path.realpath().

I tested the code with a few ansible runs however I didn't see any unit tests for playbooks specifically... If you point me to the test sweet I should run for this, I'll run it and update this PR.

Thanks

### DIFF
--- a/lib/ansible/playbook/__init__.py
+++ b/lib/ansible/playbook/__init__.py
@@ -162,7 +162,7 @@ class PlayBook(object):
         if self.module_path is not None:
             utils.plugins.module_finder.add_directory(self.module_path)
 
-        self.basedir     = os.path.dirname(playbook) or '.'
+        self.basedir     = os.path.realpath(os.path.dirname(playbook or '.'))
         utils.plugins.push_basedir(self.basedir)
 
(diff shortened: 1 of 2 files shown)


mpdehaan:
I'm ok with calling the realpath stuff here but I'm very unclear about the "or" additions.

Can you explain the intent of these in greater detail?

dingus9:
I added the or in play.py but the or was already in the **init**.py. Actually looking at it I'm thinking it should just ignore the basedir parameter and trust what is provided by the playbook object. I changed it to just accepting the internal property playbook.basedir since that should always be the basedirectory of the playbook.

How does that look

mpdehaan:
# Revision Request

Hi!

- this code change currently breaks "make tests", if you can update the code such that it passes, or fix the tests we would appreciate it

which is very active:

mpdehaan:
Closing this due to lack of activity int he above, please feel free to resubmit.  Thanks!

PR #8166, opened 2014-07-16
Fixing improperly closed file handle in ansible vault

scottbrown:
The parentheses are left off the call, telling Python to return a method but not call the method.  You can see how this works in a REPL by calling `close` with and without the parentheses, noting how the file object changes.

``` python
>>> f = open("/etc/passwd", "rb")
>>> f
<open file '/etc/passwd', mode 'rb' at 0x103ce26f0>
>>> f.close
<built-in method close of file object at 0x103ce26f0>
>>> f
<open file '/etc/passwd', mode 'rb' at 0x103ce26f0>
>>> f.close()
>>> f
<closed file '/Users/scott/.ansible/stow_vault_password', mode 'rb' at 0x103ce26f0>
```

### DIFF
--- a/bin/ansible-vault
+++ b/bin/ansible-vault
@@ -107,7 +107,7 @@ def get_opt(options, k, defval=""):
 def _read_password(filename):
     f = open(filename, "rb")
     data = f.read()
-    f.close
+    f.close()
     data = data.strip()
     return data
 


mpdehaan:
Merged, thanks!

We're aware of first class functions obviously.  Nice catch.

scottbrown:
Thanks @mpdehaan. I figured you guys knew about it, and I didn't intent it to mean I was questioning your python knowledge.  It was more of a correct vs. complete argument for the bug/PR report so that any dewy-eyed pythonista can read the PR and understand how the error happens.  Sort of a "need-and-not-have versus have-and-not-need" situation.

PR #8167, opened 2014-07-16
Fix sanitization of sudo success_key from stdout

debfx:
Commit 7e3dd10 caused a regression with sanitizing the sudo success_key from stdout.
There is no newline in front of SUDO-SUCCESS when using sudo.
This change adapts the regex to make the line break optional.

Tested on Ubuntu 14.04.


jimi-c:
Merged, thanks!

PR #8168, opened 2014-07-16
Fixing a bug where the password file option fails with file not found if...

scottbrown:
... given a tilde character, because the tilde is not expanded before being given to the open() function.

Adding a tilde (`~`) in front of the filename is a shell/bash helper, but Python passes it straight through to the `open()` method without first expanding it to the user's home directory.  This causes the `open()` method to fail because the file/directory doesn't exist.  The solution is to first check for the existence of the tilde prefix and, if it exists, expand it before proceeding with opening the file.

before fix:

``` bash
$ ansible-vault encrypt --vault-password-file=~/.ansible/common_vault_password roles/app-user/files/bitbucket_deploy_key
ERROR: [Errno 2] No such file or directory: '~/.ansible/common_vault_password'
```

after fix:

``` bash
$ ansible-vault encrypt --vault-password-file=~/.ansible/common_vault_password roles/app-user/files/bitbucket_deploy_key
Encryption successful
```

### DIFF
--- a/bin/ansible-vault
+++ b/bin/ansible-vault
@@ -105,6 +105,9 @@ def get_opt(options, k, defval=""):
 #-------------------------------------------------------------------------------------
 
 def _read_password(filename):
+    if filename.startswith("~"):
+        filename = os.path.expanduser(filename)
+
     f = open(filename, "rb")
     data = f.read()
     f.close
--- a/bin/ansible-vault
+++ b/bin/ansible-vault
@@ -105,8 +105,7 @@ def get_opt(options, k, defval=""):
 #-------------------------------------------------------------------------------------
 
 def _read_password(filename):
-    if filename.startswith("~"):
-        filename = os.path.expanduser(filename)
+    filename = os.path.expanduser(filename)
 
     f = open(filename, "rb")
     data = f.read()


bin/ansible-vault
```diff
@@ -105,6 +105,9 @@ def get_opt(options, k, defval=""):
 #-------------------------------------------------------------------------------------
 
 def _read_password(filename):
+    if filename.startswith("~"):
+        filename = os.path.expanduser(filename)
```
scottbrown: normally I don't like modifying a function argument (in this case, `filename`) in place, but I wanted to keep the code as simple and short as possible.

mpdehaan:
os.path.expanduser won't cause any harm if "~" is not in the filename, so can you please update this to just:

```
 filename = os.path.expanduser(filename)
```

If so, I'll gladly merge this.

Thanks!

scottbrown:
Thanks for the review @mpdehaan.  I was being defensive because I didn't know if there were weird edge cases when using `expanduser`, so I wanted to limit any potential damage to just input with tilde prefixes.

New code pushed and is ready for review, please.

amenonsen:
This PR is superseded by changes to read_vault_password_file (which does call os.path.expanduser) in devel and can now be closed.

bcoca:
closing as per comment above

PR #8170, opened 2014-07-16
Load existing group_vars file for new group_by group

### LINKED ISSUE #8159: group_by is reading group_vars/key_group only if such key_gorup is defined in inventory
Hello,
Using Ansible 1.7 latest. found issue using group_by.
play fragment : 
- group_by: key={{ ansible-distribution }}
- group_vars/CentOS is exists
- some variable is defined in group_vars/CentOS
  When inventory does not have such dynamic group, ansible does not try to load any info from group_vars/CentOS , and playbook can not use dynamicaly loaded variables.

When inventory file has such dynamic groups ( without hosts added ) defined, variable has correct value from corresponding key_group variable file.

I've checked _ansible-playbook_ run with _strace_ and noticed that _group_vars_ directory is accessed only for files that correspond to all groups defined in inventory file

On Ansible 1.6.2 group_by loads group_vars/files correctly , without requiring defining such groups in inventory. 

Maybe it related to some changes in the way ansible loading group_vars files

srgvg:
if it was not loaded yet, which happens if that group
didn't exist yet in the inventory.

fixes #8159

@dshvedchenko can you test this patch?

### DIFF
--- a/lib/ansible/runner/action_plugins/group_by.py
+++ b/lib/ansible/runner/action_plugins/group_by.py
@@ -93,6 +93,7 @@ class ActionModule(object):
             if not inv_group:
                 inv_group = ansible.inventory.Group(name=group)
                 inventory.add_group(inv_group)
+                inv_group.vars = inventory.get_group_variables(group, update_cached=False, vault_password=inventory._vault_password)
             for host in hosts:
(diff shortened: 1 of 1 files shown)


jimi-c:
Merged, thanks!

PR #8171, opened 2014-07-16
Support for OS X 10.10 DP3.

jbenden:
The attached commit fixes two known problems with Yosemite 10.10 OS X.

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -1397,7 +1397,7 @@ class Darwin(Hardware):
 
         rc, out, err = module.run_command("sysctl hw.usermem")
         if rc == 0:
-            self.facts['memfree_mb'] = long(out.splitlines()[-1].split()[1]) / 1024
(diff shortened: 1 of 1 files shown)


mpdehaan:
Appreciate the additions.

May I ask that you share information about what the known problems were?   (Errors?  Incorrect values?)

Also, what behaviors exist on prior versions with this change.

Thanks for elaboration.

jbenden:
For some reason, the general reading of sysctl -a fails to read the two properties in question under OS X 10.10. They are simply missing from the output. However, directly querying them works as required. This is what the proposed patch changes.

Because the same variables are read under any version of OS X, this change will not break previous versions.

ericduran:
Quick FYI: I just tested the patch on 10.10 & it fixed the issue for me.

mpdehaan:
This now contains a change in the memfree math.  Is this saying memfree was wrong in all previous OS X versions as well or does this change need to be OS specific?

(In general, it's a good idea to keep these kinds of changes on seperate branches and therefore as seperate branch submissions, so one thing doesn't hold up another).

In interest of not having OS X have fact gathering issues, I'm going to merge this regardless.

Thanks!

PR #8172, opened 2014-07-16
Adds channel key parameter

phenomenes:
I realise you could pass the key in the channel, eg `channel="#foo my_key"` but I considered this a hack and I'll much prefer a separate parameter.


mpdehaan:
Merged, thank you!

PR #8176, opened 2014-07-17
Retr0h/align markdown headers

retr0h:


mpdehaan:
I'm sorry but we don't take whitespace altering commits as they break line attribution.

Occasionally we may run them as overseers of the project, along with fixes for a selection of styling issues, however.

If you'd like to submit just the markdown cleanup, we'd be happy to merge it.

Thank you!

PR #8184, opened 2014-07-17
Enable puppet facts

yasn77:
Currently facter facts omit facts that a distributed via Puppet. This
commit adds the `--puppet` option.

In cases where puppet is not installed, the command sends a warning to
STDERR _but_ completes successfully. So should not cause any issues.

The benefit is, filtering can be done based on facts set by Puppet.

### DIFF
--- a/library/system/setup
+++ b/library/system/setup
@@ -91,7 +91,7 @@ def run_setup(module):
     # if facter is installed, and we can use --json because
     # ruby-json is ALSO installed, include facter data in the JSON
     if facter_path is not None:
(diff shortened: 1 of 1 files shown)


mpdehaan:
So the warning to stderr _MAY_ be messing with the JSON in some cases, but I think in most cases it does not.   

That all being said, I would expect very very very very very very very very (is that enough veries) people to be using facter if they didn't have Puppet introducing them to it, since we provide our own facts too.

Sounds good!

mpdehaan:
merged, thanks!

PR #8188, opened 2014-07-18
Allow role paths to be appended on the command line

willthames:
##### Issue Type:

Feature Pull Request
##### Ansible Version:

ansible 1.7 (devel d4548fdd01) last updated 2014/07/18 11:25:37 (GMT +1000)
##### Environment:

N/A
##### Summary:

Allow role paths to be appended on the command line

Move the DEFAULT_ROLE_PATH handling into bin/ansible-playbook and just pass roles_paths to playbooks (the plays pick up the path from the playbook)

Usage:

```
ansible-playbook -r /opt/local/ansible/common/roles -r /opt/local/ansible/application/roles playbook.yml
```

Still respects the defaults and settings in ansible configuration file.
##### Steps To Reproduce:

N/A
##### Expected Results:

N/A
##### Actual Results:

N/A

### DIFF
--- a/bin/ansible-playbook
+++ b/bin/ansible-playbook
@@ -83,6 +83,9 @@ def main(args):
         help="start the playbook at the task matching this name")
     parser.add_option('--force-handlers', dest='force_handlers', action='store_true',
         help="run handlers even if a task fails")
+    parser.add_option('-r', '--roles_path', dest='roles_paths', action='append',
+        default=map(os.path.expanduser, C.DEFAULT_ROLES_PATH.split(os.pathsep)),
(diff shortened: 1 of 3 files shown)


mpdehaan:
Normally, we generally try to limit command line option sprawl (it really clutters up --help and makes it confusing to new users), so I'm disinclined to add this as a -r, but depending on use cases, this might be useful.  I need to think about it.

I kind of don't see a problem with setting this as the environment variable, which is already surfaced.

mpdehaan:
I'm going to say no to the command line sprawl on this particular one, but I'm open to making it surfaced via an environment variable, so I'm going to close this one.

PR #8189, opened 2014-07-18
vmware inventory fixes and features

bcoca:
added environment variables as config overrides for vmware inventory
fixed vmware properties, was getting parent's properties
cache now actually works
removed tags, they were not the ones i was looking for
renamed cache defaults fixed ini file to match
avoid several tracebacks you got if misconfigured

this is a resubmit of #8138 (incorporating feedback), trying to update that branch brought in other unwanted commits

### DIFF
--- a/plugins/inventory/vmware.ini
+++ b/plugins/inventory/vmware.ini
@@ -6,8 +6,8 @@ guests_only = True
 #hw_group =
 
 [cache]
-cache_max_age = 3600
-cache_dir = /var/tmp
+max_age = 3600
+dir = ~/.cache/ansible
 
 [auth]
 host = vcenter.example.com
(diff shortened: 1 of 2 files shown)


kleppy:
Made an edit to line 91 to get it to work for me. 
                'vmware_name' : host.name,

# 'vmware_parent': host.parent.name,

```
            'vmware_parent': host.parent_name,
           }
```

Tested against VCenter Version 5.1.0 Build 1465747 .

mpdehaan:
cc @lukesneeringer @cchurch

mpdehaan:
(note to team:  we need to make sure cache controls are effectively leveraged/disabled like ec2 in Tower's usage)

bcoca:
@kleppy, weird, it used to work with my 5.1 (we just decommissioned it so I
cannot test).

I'll add a test and use one or the other on success, probably update it
tonight.
​

jimi-c:
Merged, thanks!

PR #8175, opened 2014-07-16
Added basic support for SSL when using IRC

ahamilton55:
I've added basic support for connecting to an IRC server with SSL. Nothing fancy but it works.

PR #8190, opened 2014-07-18
Support SSL IRC servers

phenomenes:
Adds SSL support


mpdehaan:
Duplicate of #8175 so I'm linking tickets (this will appear closed but can be found from the other), we can decide which we like better :)

Thank you!

PR #8193, opened 2014-07-18
Re-factor to allow importing of ec2 inventory plugin

lgtml:
# Purpose

Update to allow import of inventory as library in other ansible modules. This save me from having to shell execute the ec2.py and `json.loads` the stdout.
# Tests
- [x] Tested normal execution from shell
- [x] Tested `--host` execute from shell
- [x] Test importation from other ansible library

### DIFF
--- a/plugins/inventory/ec2.py
+++ b/plugins/inventory/ec2.py
@@ -154,22 +154,29 @@ class Ec2Inventory(object):
         elif not self.is_cache_valid():
             self.do_api_calls_update_cache()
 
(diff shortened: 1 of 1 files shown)


mpdehaan:
Hi, 

We don't want to see other modules importing inventory because you may be running ec2 from an intermediate host inside a region and that region should not need ansible installed on it.

It seems weird to want to access inventory from a module, can you stop by ansible-project and perhaps we can understand your use case more?

Thanks!

PR #8195, opened 2014-07-18
Adding Tests Integration for mysql db

wrosario:
Adding the following Test Coverage for RedHat and Debian OS:
- Use mysql_db module to create, delete databases using different encoding.
- Backup/Restore databases using different file format: sql, gz and bz2.
- Create and delete databases using different user privilege.
- Assert database creation, deleting and content using system commands.


jimi-c:
Merged, thanks!

Issue #6131, opened 2014-02-23
Unarchive should support unpacking remote archives


Bilge:
If I wish to download an archive on the remote host using get_url then the unarchive module is unable to help me extract it since unarchive only extracts local archives. In its current implementation, the only way to achieve this is: 
- get_url the file on the remote host
- fetch it back to the local host
- unarchive it back to the remote host

This is incredibly wasteful and nigh on ridiculous. Unarchive should be able to extract remote archives.

ansibot:
Thanks for filing a ticket! I am the friendly GitHub Ansibot. I see you did not fill out your issue description based on our new issue template. Please copy the contents of this [template](https://raw2.github.com/ansible/ansible/devel/examples/issues/ISSUE_TEMPLATE.md) and **paste it into the description** of your ticket. 

Our system will automatically close tickets that do not have an issue template filled out within 7 days. Please note that due to large interest in Ansible, humans may not comment on your ticket if you ask them questions. Don't worry, you're still in the queue and the robots are looking after you.

mpdehaan:
First request -- Please don't call contributors contributions ridiculous -- I'm very thankful for unarchive, and it's true it could do more.   However, because a personal itch hasn't been scratched doesn't mean there isn't a lot of value in a new module, and I'm thankful for it.

This feature idea is also slightly based on an assumption that the unarchive module must be handled in three step process.  It's two.  Download to the local host, push it to the remote.  This requires an extra line or two, 
- get_url: http://foo dest=/path/to/file.tgz
- unarchive: src=/path/to/file.tgz ...

Which is a little longer than normal, but it's not any more wasteful on bandwidth.  It wastes an additional line in the playbook, but that's all.

Bilge:
get_url does not download to the local host, it downloads to the remote host, nor do I want to download to the local host so your steps would not be functionally equivalent even if it did work that way. Which it doesn't.

mpdehaan:
The get_url module will download to any host you run the play on, actually.  You can do that in a local play, or use "delegate_to/local_action".

The play is a better approach.

Bilge:
I don't know what you're talking about but I thought I made it clear I don't want to download to the local host. The problem is that get_url works on the remote side and unarchive works on the local side. Therefore they cannot be used together directly.

mpdehaan:
Not directly, but they can in fact be used in conjunction.

```
- hosts: localhost
  tasks:
     - get_url: ...

- hosts: managed
  tasks:
[...]

PR #8200, opened 2014-07-19
[Docs] Added another example to demonstrate unarchive module working entirely on a remote machine

geetotes:
Like many people in [this thread](https://github.com/ansible/ansible/issues/6131) I missed the part in the unarchive documentation about it being able to work entirely on a remote machine. I've added an example to the docs to remedy this.

### DIFF
--- a/library/files/unarchive
+++ b/library/files/unarchive
@@ -71,7 +71,7 @@ EXAMPLES = '''
 - unarchive: src=foo.tgz dest=/var/lib/foo
 
 # Unarchive a file on a remote machine
-- unarchive: src=/tmp/foo.zip dest=/usr/local/bin
+- unarchive: src=/tmp/foo.zip dest=/usr/local/bin copy=false
 '''
 
 import os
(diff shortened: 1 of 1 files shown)


mpdehaan:
I'm confused by this one since it looks just like the example above:

```
 - unarchive: src=foo.tgz dest=/var/lib/foo
 +
 +# Unarchive a file on a remote machine
 +- unarchive: src=/tmp/foo.zip dest=/usr/local/bin
```

If you're suggesting local versus remote is governed by whether our not the src is absolute, that's a bug in the module that we need to chase, as this should not be automagic as such.

Let me know.

bcoca:
I think this PR clarifies usage of copy
https://github.com/ansible/ansible/pull/7312

flynfish:
:+1: This is a much better explanation and reduces confusion.

geetotes:
Lol, sorry about that bad example; that's what I get for too little sleep. I've pushed up an example showing the `copy=false` flag in action. Should it be changed to `copy=no` to reflect the language in #7312 ?

mpdehaan:
false or no are both valid, but no is more consistent with the rest of ansible, I'll merge this and make the docs tweak.  Thanks!

PR #5534, opened 2014-01-07
FactCache updates

sivel:
- The get method of a dict does not internally call `__getitem__`, so add a get method that does

Currently `utils.update_hash` uses `FactCache.get()` which does not internally utilize the `__getitem__` method, so it is always returning `{}`
- When trying to update subkeys of self.SETUP_CACHE make sure to use utils.update_hash

These are just a few missed from 3ba0ea064df78387f0898a680c6936bd14a20d9c

PR #8203, opened 2014-07-19
Fact Caching Revisited

joshdrake:
Discussed initially on the project mailing list here: (https://groups.google.com/d/topic/ansible-project/bMWWRPxRQqg/discussion). Also discussed at length in other pulls and mailing list topics:
- [Fact caching (work in progress) infrastructure now on development branch](https://groups.google.com/d/topic/ansible-devel/aC-BjeBTMho/discussion)
- [Pull Request #5534](https://github.com/ansible/ansible/pull/5534)

Reiterating some bits specific to this pull request for posterity:
1. Only SETUP_CACHE is leveraging caching backends. VARS_CACHE is untouched as I'm not quite sure I understand the use-case behind caching play variables between playbook executions.
2. Caching backends have a base class they should extend to ensure the API is properly implemented. All the heavy lifting is done by each caching backend.
3. Given the existing usage of SETUP_CACHE (eg: dictionary based access), caching backends must be able to return the keys that are being held in cache. There are various ways of doing that can be seen in the diff. Redis is perhaps the most interesting and optimal since it allows usage of sorted sets.
4. All unit tests pass and the sample playbooks noted as issues in the previous threads are not present. I haven't had time recently to do so, but I'll work on running the integration tests as well.

As I mentioned on the mailing list, I think redis (and probably memcached) would handle large-scale use-cases fine, but other potential backends (and probably the _file_ backend I've written as well, it flushes writes to disk far to frequently) may not. If someone could advise on how best to set up a testing harness to run this against a large quantity of hosts, I would be happy to do so.

Finally, the commits will be squashed or reworded before merging to fit commit guidelines. The diff should also be viewed ignoring whitespace ([as so](https://github.com/joshdrake/ansible/compare/feature/fact_caching?expand=1&w=1)), but I can fix that immediately if needed.

### DIFF
--- a/bin/ansible-playbook
+++ b/bin/ansible-playbook
@@ -83,6 +83,8 @@ def main(args):
         help="start the playbook at the task matching this name")
     parser.add_option('--force-handlers', dest='force_handlers', action='store_true',
         help="run handlers even if a task fails")
+    parser.add_option('--flush-cache', dest='flush_cache', action='store_true',
+        help="flush to fact cache")
 
     options, args = parser.parse_args(args)
 
@@ -201,6 +203,10 @@ def main(args):
             vault_password=vault_pass,
             force_handlers=options.force_handlers
         )
+
+        if options.flush_cache:
+            display(callbacks.banner("FLUSHING FACT CACHE"))
+            pb.SETUP_CACHE.flush()
 
         if options.listhosts or options.listtasks or options.syntax:
             print ''
@@ -324,4 +330,3 @@ if __name__ == "__main__":
     except KeyboardInterrupt, ke:
         display("ERROR: interrupted", color='red', stderr=True)
         sys.exit(1)
-
--- a/lib/ansible/cache/__init__.py
+++ b/lib/ansible/cache/__init__.py
@@ -50,10 +50,12 @@ class FactCache(MutableMapping):
         return len(self._plugin.keys())
 
     def copy(self):
-        """
-        Return a primitive copy of the keys and values from the cache.
-        """
+        """ Return a primitive copy of the keys and values from the cache. """
         return dict([(k, v) for (k, v) in self.iteritems()])
 
     def keys(self):
         return self._plugin.keys()
+
+    def flush(self):
+        """ Flush the fact cache of all keys. """
+        self._plugin.flush()
--- a/lib/ansible/cache/__init__.py
+++ b/lib/ansible/cache/__init__.py
@@ -1,0 +1,59 @@
+# (c) 2014, Michael DeHaan <michael.dehaan@gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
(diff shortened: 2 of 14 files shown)


bcoca:
very cool, I'll probably test it this week.

mpdehaan:
Excellent.

On my list here is to do a "time ansible-playbook" run with a somewhat meaningful playbook, 10k hosts, with fact caching turned off vs using Redis.  That performance differential will let us know how we want to tune this.

joshdrake:
Forgot to push up a commit I had sitting locally for `cPickle` over `pickle` in the redis module. It drastically speeds up set operations. It's also worth mentioning that using the `hiredis` parser can also provide some performance benefit as well. The Python `redis` package will automatically use the `hiredis` parser if possible.

I'll also push up an **etcd** backend in the next couple of days if performance from testing it is decent. I was playing around with it lately for other purposes and I can conceivably see how it might be useful for persisting facts from Ansible.

joshdrake:
I had previously been using this branch in testing without letting Ansible fork (eg: forks=1, I was invoking playbook execution programmatically in another application's child process), and I found a large bug today when testing under normal circumstances where Ansible leverages concurrency. 

Basically, `python-memcached` has zero connection pooling capability and thus isn't safe under concurrent access where the connection is shared between threads/forks. `redis-py` includes a connection pool and handles this gracefully; `pylibmc` (the other major Python library which requires `libmemcached`) also has connection pooling available. 

Since introducing another dependency on top of the memcached library itself is not great, using `pylibmc` is probably out of the question. Thus, I [implemented a simple connection pool](https://github.com/joshdrake/ansible/commit/8049eb00a441a75453604ae35f1372f41f541433) as a transparent client proxy in the memcached plugin. 

This complicates the implementation but avoids having to make changes in the Ansible runner itself. Such an alternative would involve delaying cache plugin instantiation until after forking so that each child process creates its own connection. The connection pool implementation aligns with how the redis cache backend already operates and obviates the need for potentially introducing complications elsewhere in the codebase.

mpdehaan:
Yeah probably means dropping the memcache part :)

Deps can be soft and only included if the plugin is used.

I would think we'd probably suggest Redis?

mpdehaan:
All merged in, with comments, see my thread on https://github.com/ansible/ansible/pull/8203 and feel free to weigh in!

Super awesome.  Made a few tweaks to not use cPickle (compatibility) and ended up working out nicely once I figured out I needed the redis python library from pip, not EPEL.

Docs pending, anyway, see the mailing list and wanted to say thanks again to you at @bcoca for doing this!

Very excited.

joshdrake:
Looks great, thanks for the review and merge effort @mpdehaan! Just also wanted to confirm that I think the recommendation for using redis is valid here. It was the most performant in my testing and is the most robust and idiomatic backend implementation. I'm busy this week but I'll keep an eye on the mailing list and help out with docs when I can if no one else gets to it first. 

Cheers!

geerlingguy:
For anyone else led here through Google Search (until the docs section rises in search ranking to supersede this issue), here's a link to the docs explaining how you can use fact caching: http://docs.ansible.com/playbooks_variables.html#fact-caching

PR #8204, opened 2014-07-19
Add quotes around variables and substitution in hacking/env-setup

martin-ueding:
In Bash, the variables need to be quoted. Fixed strings are fine. See
http://mywiki.wooledge.org/BashPitfalls#A.5B_.24foo_.3D_.22bar.22_.5D to
see more of this.

Every variable could possibly contain spaces, and the output from any
process can contain spaces as well. Therefore it is important to put
quotes around everything.

I also upgraded the tests from `[` to `[[` which is a bash builtin, has
more features and is faster since it does not require a fork. For this
script, it will not matter, but since this is using bash already, one
can use its features. See http://mywiki.wooledge.org/BashFAQ/031 for
more info on `[`, `[[` and `test`.

The backticks are legacy, one can use `$( ... )` in bash that even
allows nesting, see http://mywiki.wooledge.org/CommandSubstitution for
more on this.

### DIFF
--- a/hacking/env-setup
+++ b/hacking/env-setup
@@ -3,31 +3,31 @@
 #    modifies environment for running Ansible from checkout
 
 # When run using source as directed, $0 gets set to bash, so we must use $BASH_SOURCE
-if [ -n "$BASH_SOURCE" ] ; then
-    HACKING_DIR=`dirname $BASH_SOURCE`
-elif [ $(basename $0) = "env-setup" ]; then
-    HACKING_DIR=`dirname $0`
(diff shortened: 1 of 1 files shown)


amenonsen:
This PR can be closed. There's been a lot of churn in the code since it was opened, and many quotes have been added. Anyway A=$(foo) is safe even if foo returns something with spaces. But the cases where it matters, e.g. PYTHONPATH=$PREFIX_PYTHONPATH:$PYTHONPATH are already quoted in devel.

bcoca:
closing as per comment above

PR #8210, opened 2014-07-19
Add documentation regarding testing with support_check_mode=True

hacosta:
Signed-off-by: Hector Acosta hector.acosta@gmail.com


hacosta:
On second thought, this may be an implementation detail that shouldn't be exposed, mind if i add a --check argument to test_module.py?

mpdehaan:
Yep!  I'd much rather see this as a parameter to the hacking script that says "--check".

That would be great, thanks!

PR #8215, opened 2014-07-20
Make code more PEP8 compliant & Pythonic

ciupicri:


mpdehaan:
We generally do not take these commits from outside and make passes occasionally, it's ok to refactor something while in the process of working on it, but in general, we do not want to have our pull request queue cluttered up with these types of submissions since we're going to de-prioritize them in favor of other work.

Bug fixes would be great!

ciupicri:
I was afraid of this, although the [CODING_GUIDELINES.md](https://github.com/ansible/ansible/blob/devel/CODING_GUIDELINES.md) mentions only whitespace<sup>*</sup> and my changes were a bit more consistent. Anyway, thanks for having a look at it so fast.

> <sup>*</sup>Do not submit pull requests that simply adjust whitespace in the code

PR #8217, opened 2014-07-21
Register boolean jinja filters as jinja tests

ghost:
### Summary

This Pull Request ensures that all Ansible-provided boolean Jinja2 filters are also available as Jinja2 tests. Two main changes make this possible:
1. I moved boolean-valued filters from `runner/filter_plugins/core.py` to `runner/test_plugins/core.py`.
2. In `utils/template.py`, both filters and tests are registered as Jinja2 filters. Only tests are registered as tests.

I also updated some related documentation, and did some minor refactoring in `utils/template.py` to avoid excessive repetition.
### Motivation

Many Ansible filters return boolean values, but they cannot be used in tests. For example:

``` yaml
# test.yml
- hosts: localhost
  gather_facts: no
  vars:
    sample:
      - { a: hello, b: world }
      - { a: goodbye, b: world }
      - { a: foo, b: bar }
  tasks:
    - debug: var="{{ sample | rejectattr('b', 'match', 'bar') | list }}"
```

Running this (with `ansible-playbook -i localhost, -e connection=local test.yml`) results in:

```
jinja2.exceptions.TemplateRuntimeError: no test named 'match'
```

After these commits, the output is what you probably expect:

```
ok: [localhost] => {
    "[{'a': 'hello', 'b': 'world'}, {'a': 'goodbye', 'b': 'world'}]": [
        {
            "a": "hello", 
            "b": "world"
        }, 
        {
            "a": "goodbye", 
            "b": "world"
        }
    ]
}
```

Of course, this is a trivial, silly example. Because this change also allows users to define their own test_plugins, it should allow for some interesting ways to pluck data from list variables that result from calls to other modules.
### Followup work
- In order to get the tests to pass, I hacked in an import in `filter_plugins/core.py`.
  - We could obviously move some tests, but I didn't want to do a lot of boilerplate work if no one is interested in this change, and
  - It seemed like it might be more suitable to test at a higher level (i.e., try to actually use the templates) instead of baking a specific `filter_plugin/core.py` API into the tests.
- It might be worth considering a refactor of some of the plugin loading, which has quite a bit of boilerplate (I assume this is a result of minor copy-paste-edits like I used here to add the test plugins).

### DIFF
--- a/docsite/rst/developing_plugins.rst
+++ b/docsite/rst/developing_plugins.rst
@@ -56,6 +56,15 @@ If you want more Jinja2 filters available in a Jinja2 template (filters like to_
 
 Jump into `lib/ansible/runner/filter_plugins/ <https://github.com/ansible/ansible/tree/devel/lib/ansible/runner/filter_plugins>`_ for details.
 
+.. _developing_test_plugins:
+
+Test Plugins
+--------------
+
+If you want more Jinja2 tests available in a Jinja2 template (tests like match and version_compare are provided by default), they can be extended by writing a test plugin.  Test plugins are also automatically available as Jinja2 filters. Most of the time, when someone comes up with an idea for a new test they would like to make available in a playbook, we'll just include them in 'core.py' instead.
+
+Jump into `lib/ansible/runner/test_plugins/ <https://github.com/ansible/ansible/tree/devel/lib/ansible/runner/test_plugins>`_ for details.
+
 .. _developing_callbacks:
 
 Callbacks
--- a/lib/ansible/runner/filter_plugins/core.py
+++ b/lib/ansible/runner/filter_plugins/core.py
@@ -19,14 +19,13 @@ import base64
 import json
 import os.path
 import yaml
-import types
 import pipes
 import glob
 import re
-import operator as py_operator
 from ansible import errors
 from ansible.utils import md5s
-from distutils.version import LooseVersion, StrictVersion
+# TODO: hack to make tests pass; update tests instead
+from ansible.runner.test_plugins.core import *
 from random import SystemRandom
 from jinja2.filters import environmentfilter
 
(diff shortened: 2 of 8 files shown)


docsite/rst/developing_plugins.rst
```diff
@@ -56,6 +56,15 @@ If you want more Jinja2 filters available in a Jinja2 template (filters like to_
...
 
+.. _developing_test_plugins:
+
+Test Plugins
+--------------
+
+If you want more Jinja2 tests available in a Jinja2 template (tests like match and version_compare are provided by default), they can be extended by writing a test plugin.  Test plugins are also automatically available as Jinja2 filters. Most of the time, when someone comes up with an idea for a new test they would like to make available in a playbook, we'll just include them in 'core.py' instead.
+
+Jump into `lib/ansible/runner/test_plugins/ <https://github.com/ansible/ansible/tree/devel/lib/ansible/runner/test_plugins>`_ for details.
+
```
ghost, line 67: I did not test build the documentation, since the Contributing docs didn't indicate to do that. I did edit carefully, though, basing everything on existing examples.

lib/ansible/runner/test_plugins/core.py
```diff
@@ -0,0 +1,151 @@
+# (c) 2012, Jeroen Hoekx <jeroen@hoekx.be>
+# (c) 2014, Brandon Mintern <brandon@mintern.net>
```
ghost, line 2: Am I supposed to add a copyright entry like this?

lib/ansible/runner/filter_plugins/core.py
```diff
@@ -42,46 +41,6 @@ def to_nice_json(a, *args, **kw):
     '''Make verbose, human readable JSON'''
     return json.dumps(a, indent=4, sort_keys=True, *args, **kw)
```
ghost, line 43: All deletions below moved verbatim to `test_plugins/core.py`.

lib/ansible/utils/template.py
```diff
@@ -73,6 +89,16 @@ def _get_extensions():
 
     return jinja_exts
```
ghost, line 91: I added the following function to remove some of the existing repetition and to avoid exacerbating it with my changes

mpdehaan:
There's something rather ugly with this diff, first a large scale removal and then what appears to be the replacement of the whole file.  Would it be possible to clean this up?

mpdehaan:
I'm also unclear about the idea of "test plugins".   This feels like a new topic and really needs to be brought up on the development list.

I'm generally not up for introducing new types of plugins.

mpdehaan:
I'm also unclear why this is moving various filters to a new class.

It would seem that exposing more jinja2 internals could be done without moving this to a different namespace or introducing a new plugin structure.

Stop by ansible-devel if you would like to discuss.

mintern:
I answered a lot of these questions in my Pull Request summary.

Jinja seems to have two separate notions: _tests_ transform variables into booleans (used by `is`, `is not`, and some filters); all other transformations are _filters_ (which use the `|` syntax).

Ansible conflates these two notions, referring to all of them as filters; thus we have, e.g., `| success` instead of `is success`, and then `is defined` looks inconsistent as a result.

Furthermore, Ansible "filters" like `match` and `version_compare` are logically tests, but they can't be used in jinja filters like `reject` since they are not registered as such.

Finally, there is currently no way for users to define simple, helpful new tests (e.g., `startswith`) to be registered with Ansible's jinja instance.

This Pull Request addresses all of those problems with pretty minimal changes. I don't really understand the confusion. Do you consider this a low quality Pull Request? Were my comments unhelpful?

ghost:
I reopened this using a much cleaner approach in #8235.

PR #8227, opened 2014-07-21
win_setup improvements: Correcly outputs ansible_totalmem

trondhindenes:
The original setup.ps1 didn't report ansible_totalmem on computers with more than one physcial ram chip, as this causes an array to be returned which the script didn't cater for. This version will report the correct amount of ram regardless of the number of ram chips installed on the system. (tested on both physical multi-chip systems as well as virtual ones)


mpdehaan:
Merged, thank you!

PR #8229, opened 2014-07-22
Update intro_getting_started.rst

enwood:
New to ansible, but installing via HomeBrew on Mac OS X (Mountain Lion), I discovered from a note on http://lampros.chaidas.com/index.php/2014/06/22/beginning-with-ansible-on-freebsd-10/ that the inventory file (`hosts`) needs to be in `/usr/local/etc/ansible` not `/etc/ansible`. Will submit change for Inventory doc as well. With a hosts file in `/etc/ansible`, ansible did not detect it, and continued to prompt me to specify one with the "-i" parameter. Am I wrong, or is this an OS X anomaly?

### DIFF
--- a/docsite/rst/intro_getting_started.rst
+++ b/docsite/rst/intro_getting_started.rst
@@ -47,8 +47,9 @@ Your first commands
 
 Now that you've installed Ansible, it's time to get started with some basics.
 
-Edit (or create) /usr/local/etc/ansible/hosts and put one or more remote systems in it, for
-which you have your SSH key in ``authorized_keys``::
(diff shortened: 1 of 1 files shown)


srgvg:
Same remark as #8230

mpdehaan:
This is a packaging error in homebrew deviating from upstream without discussing it first, it would be better to correct homebrew to match upstream ansible.

It's very terrible that they would patch the code to look in different default locations.

Until this is resolved, I'm much inclined to just suggest installation via pip, which also works on OS X.

PR #8230, opened 2014-07-22
Update intro_inventory.rst

enwood:
New to ansible, but installing via HomeBrew on Mac OS X (Mountain Lion), I discovered from a note on http://lampros.chaidas.com/index.php/2014/06/22/beginning-with-ansible-on-freebsd-10/ that the inventory file (`hosts`) needs to be in `/usr/local/etc/ansible` not `/etc/ansible`. Submitted change for Intro doc as well. With a hosts file in `/etc/ansible`, ansible did not detect it, and continued to prompt me to specify one with the "-i" parameter. Am I wrong, or is this an OS X anomaly?

### DIFF
--- a/docsite/rst/intro_inventory.rst
+++ b/docsite/rst/intro_inventory.rst
@@ -8,7 +8,7 @@ Inventory
 Ansible works against multiple systems in your infrastructure at the
 same time.  It does this by selecting portions of systems listed in
 Ansible's inventory file, which defaults to being saved in 
-the location /etc/ansible/hosts.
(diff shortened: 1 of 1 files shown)


srgvg:
hi @enwood ; I believe your updates are specific for HomeBrew, whilst the docs are correct for most other platforms, at least Linux. I think your updates should leave current docs as is, and only mentioning a special remark for OSX?

mpdehaan:
Homebrew has made a change I don't agree with, and we don't control this packaging.

let's get homebrew to fix the problem with homebrew.

PR #8236, opened 2014-07-22
digital ocean module:  Fail if given an unknown state

zupo:
I just lost hours debugging what the hell is going wrong with this module. In the end I found that I mis-spelled the state: "preset" instead of "present". The code just skipped everything, no return, no error report. 

It should fail if given an invalid state.

### DIFF
--- a/library/cloud/digital_ocean_domain
+++ b/library/cloud/digital_ocean_domain
@@ -212,6 +212,9 @@ def core(module):
 
         event_json = domain.destroy()
         module.exit_json(changed=True, event=event_json)
+        
+    else:
+        raise ValueError('Unknown state: {}'.format(state))
 
 
 def main():


mpdehaan:
We generally like to have things 2.4 compliant, can you replace the "format" with the "classic %s"?

mpdehaan:
So the module has a facility for describing valid choices for the state.  This is the best way to assert the states are required.  Let me look at this really quick.

mpdehaan:
Ok I'm a bit confused by this one, see this line in the module:

```
        state = dict(choices=['active', 'present', 'absent', 'deleted'], default='present'),
```

May I ask what version of ansible you might be using?

If it's recent, I'd like to see the line from your playbook with the mispelling so I can see what's up.

zupo:
Sorry, semi-incorrect bug report: the culprit was that my state was "active", which is listed as a valid state, but it does have a case for it in the if-elif tree in the module.

So if state is present or absent, there is code that is ran. If your state is active or deleted, the if-elif does not have a case for it, and nothing is ran.

mpdehaan:
Ok so sometimes the states are invalid choices then, I believe that's what you are saying.

Perhaps the message shouldn't be "unknown state" but instead describe what the valid states are in those cases?

zupo:
Or, even better, remove 'active' and 'deleted' from the choices list.

mpdehaan:
Hi!

Thanks you very much!

PR #8241, opened 2014-07-22
windows copy module

t-v:
this needs pysmbc

The purpose of this module is to clone the default ansible module as much as possible and transfer files through smb


mpdehaan:
The win_copy module looks like it's executing client side with local_action which might be confusing.

Can you see about moving this to an action plugin?

t-v:
yes, it's running on the ansible machine, and not the managed server,
but the main reason for that is that I can use it for transfering items from my management server to the managed server.
If I did this the other way around, I would need to have a managed SMB share on the machine where my ansible is installed, which doesn't really make any sense to me.
My thought was, if you make a role, you have a tasks, files, templates,.. section.
With this module, I could keep my file in my default location, and get it pushed over smb towards the needed machine.
If I would have made it a pull action, my playbooks/roles/templates on my management machine need to be published over smb or http if I'd use the fetch command.

Or I'm just looking at it all wrong.

t-v:
The base framework is already there. It's not up to par with the original copy module yet, but that'll come in time. I also added a _remote.md5 function to the winrm.py to be able to do an md5 check through powershell from a serverside action.

cchurch:
Remote md5 is already available through runner.  For example, here's how the fetch action plugin uses it:  https://github.com/ansible/ansible/blob/devel/lib/ansible/runner/action_plugins/fetch.py#L82

t-v:
but that function makes an ssh connection and does the md5sum with python, this was meant to be for a windows system, so either ssh and python wont work natively.
This is the reason I made the md5 module in powershell over winrm.

cchurch:
When using a winrm connection, runner uses this PowerShell code for remote md5:
https://github.com/ansible/ansible/blob/devel/lib/ansible/runner/shell_plugins/powershell.py#L87

mpdehaan:
note to make sure this import is try/excepted to not require a new ansible dep for non-windows users

t-v:
Luckily, that already happened.

ptman:
Is it impossible to implement the copy module without allowing SMB/CIFS from the managing Linux host to the managed Windows host? Would it be possible to pass the file contents as base64 encoded data in the control stream? I admit that I'm not quite sure how the copy module works on unix hosts, but at least you don't have to run a fileserver on the target

bcoca:
you kind of do, it uses ssh, which has scp and sftp as clients that can
copy streams or files over to the ssh server.​

cchurch:
As @bcoca said, SSH connections provide a mechanism for transferring files to/from *nix hosts, whereas the Windows management protocols don't provide anything quite like that.

The current winrm connection plugin already sends files (i.e. modules and scripts) as PowerShell commands with base64 encoded arguments (https://github.com/ansible/ansible/blob/devel/lib/ansible/runner/connection_plugins/winrm.py#L155).  It's inefficient, but functional for pushing modules.  **Your action plugin could use conn.put_file() as an alternative to SMB.**

I'm looking into sending file content as stdin to PowerShell (via http://msdn.microsoft.com/en-us/library/cc251742.aspx once I add support to my fork of pywinrm) instead of a bas64-encoded argument, which would allow for pushing files in larger chunks.

trondhindenes:
We really need copy/template modules on Windows, but introducing SMB/CIFS is a step in the wrong direction IMHO. For now, put_file() should be all that's needed right?

PR #8242, opened 2014-07-22
kickstart script for Windows to fully enable ansible

t-v:
credit to: Matt Martz and trondhindenes

This script will ask for a password for the ansible user and at the end it will ask if you wish to reboot
(if it has done some actions that require it)
It won't do any actions twice, so it doesn't generate an problem if you run it mutliple times after eachothe

### DIFF
--- a/examples/scripts/kickstart.ps1
+++ b/examples/scripts/kickstart.ps1
@@ -1,0 +1,448 @@
+# Powershell script to upgrade a PowerShell 2.0 system to PowerShell 3.0, configure WinRM and set your firewall rules and create an ansible user
+# based on http://occasionalutility.blogspot.com/2013/11/everyday-powershell-part-7-powershell.html
+# credit to: Matt Martz and trondhindenes
+#
+#   PS > kickstart.ps1
+#   it will ask for the ansible password
+#   it will ask for a reboot
+#   
+
+# Get version of OS
+
+# 6.0 is 2008
+# 6.1 is 2008 R2
+# 6.2 is 2012
+# 6.3 is 2012 R2
+
+[CmdletBinding()]
+
+Param (
+    [string]$SubjectName = $env:COMPUTERNAME,
+    [int]$CertValidityDays = 365,
+    $CreateSelfSignedCert = $true,
+    [string]$user = "ansible"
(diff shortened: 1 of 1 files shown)


mpdehaan:
Can you please add the documentation stub for this module?

mpdehaan:
Can you share how this is different from https://github.com/ansible/ansible/pull/8045 ?

t-v:
This upgrades your .NET framework, upgrades to PowerShell 3, creates an Ansible account, creates a selfsigned certificate, sets up WinRM over HHTPS and configures your required firwall rule.
The script is also fully idempotent, so you can keep running it on servers where you aren't sure they are decently enlisted already.
This makes it an All-in-one ansible-client setup script. and was the reason why I assumed it would be usefull to have it under the examples in the Ansible repo.

mpdehaan:
so that's what it does, can you relate how it was different from #8045?  I'm looking to merge these two ideas, basically.

t-v:
it does identically the same, since I used most of his code for the winrm configuration and firewall settings.
(also the reason I credited him in the script, it was a great help.)
The addition I made, was let it also create a dedicated ansible user (for which it asks the password)
And more importantly, it installs the required .NET framework (prereq for powershell 3.0) if it wasn't installed yet, and upgrades powershell to v3.0 as it is required for ansibe 1.7.
And it is idempotent, so there is no problem running the script multiple times.

trondhindenes:
We should consider combining these into one, and then do variable-controlled configurations, e.q createansibleuser=$true and so on. I don't think I want my name in a script which disables UAC on the system without the user explicitly telling it to do so (and frankly I don't understand why you want to disable UAC anyway, winrm does isn't bothered by it anyways). My suggestion is to move the discussion to the google groups, and come up with a script that does everything we need it to in the best (and most secure) way possible.

jimi-c:
Based on the above, we are going to merge in #8045 and close this. If there are any differences you'd like to add to the setup script as it stands, please make PRs agaist that, but as @trondhindenes mentions any major actions like disabling UAC or any other security features must be a configurable option (with the default set to leaving the security feature on). Thanks!

PR #8244, opened 2014-07-22
Update mysql_user doc and example

wrosario:
Update to mysql_user comments and adding an example 

Updated mysql_user doc to add an example of revoking user permissions.  Also remind users where to find acceptable values for the 'priv' parameter".

### DIFF
--- a/library/database/mysql_user
+++ b/library/database/mysql_user
@@ -123,6 +123,9 @@ EXAMPLES = """
 # Specify grants composed of more than one word
 - mysql_user: name=replication password=12345 priv=*.*:"REPLICATION CLIENT" state=present
 
+# Revoke all privileges for user 'bob' and password '12345' 
+- mysql_user: name=bob password=12345 priv=*.*:USAGE state=present
+
 # Example privileges string format
 mydb.*:INSERT,UPDATE/anotherdb.*:SELECT/yetanotherdb.*:ALL
 


mpdehaan:
I'm a bit unclear on the above, is this the actual syntax for the REVOKE or does USAGE get replaced with something?

```
+# Revoke all priveleges for user 'bob' and password '12345' 
+- mysql_user: name=bob password=12345 priv=*.*:USAGE state=present
+
```

Also it appears there is a slight inconsistency in indentation on line 73, compared to other lines, though I'm unclear if this causes a problem.

I'd probably leave out the "For more information search for privileges provided by MySQL" unless we intended to tell them where to search, which might not hurt.

wrosario:
In order to revoke we are using the usage.
USAGE does not get expand or replace.  USAGE Synonym for “no privileges”.
The actual usage is explain by this table: https://dev.mysql.com/doc/refman/5.1/en/grant.html#idm47715201608896 

Couldn't  find a permanent link for the privileges on mysql website.  Will update the indentation on line 73.

mpdehaan:
Got it!

wrosario:
Updated code. Please review.

mpdehaan:
Spelling error on "privileges" needs a quick update, otherwise +1

wrosario:
good catch, I updated the spelling.

mpdehaan:
Merged, thanks!  This will appear in the next docs rebuild.

PR #8246, opened 2014-07-22
synchronize: use a single -F instead of -FF

Ernest0x:
This is needed when one uses an .rsync-filter file to exclude some paths from both being transferred and being deleted (because they are handled with other tasks). Deletions are done on the side of the receiver, so the .rsync-filter file must be transferred to the receiver, so that the receiver knows what files to delete and what not to delete.

### DIFF
--- a/library/files/synchronize
+++ b/library/files/synchronize
@@ -252,7 +252,7 @@ def main():
     group = module.params['group']
     rsync_opts = module.params['rsync_opts']
 
-    cmd = '%s --delay-updates -FF' % rsync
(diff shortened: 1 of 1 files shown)


Ernest0x:
Hello,

Is it planned to merge this for 1.7 which is soon to be released? It's a very small change but very needed for correct behavior on deletions.

Ernest0x:
I have just resubmitted this pull request at the new 'ansible-modules-core' branch. Please, merge this for 1.8.

PR #8248, opened 2014-07-22
Fix concatenated json chunk parsing issue for docker_image module

fangpenlin:
It might be useful to see log when something went wrong in result. Current implementation is to drop json if something goes wrong while parsing it. My approach is to feed the json char by char and parse it. So that we won't drop any json chunk

### DIFF
--- a/library/cloud/docker_image
+++ b/library/cloud/docker_image
@@ -118,6 +118,31 @@ try:
 except ImportError:
     from docker.client import APIError as DockerAPIError
 
+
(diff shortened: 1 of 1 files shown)


mpdehaan:
I think this code is a little heavy for dealing with JSON parse errors.

I would be ok with at try/catch and then returning the string in another attribute.

Also, are JSON parse errors common in Docker?

I'd suggest stopping by the ansible-project mailing list so we can discuss further.

Thanks!

PR #8256, opened 2014-07-23
FreeBSD does not like spaces in /etc/sysctl.conf.

berenddeboer:
Fix so sysctl works on FreeBSD. FreeBSD does not want to see spaces around the ' = ' character.

### DIFF
--- a/library/system/sysctl
+++ b/library/system/sysctl
@@ -293,7 +293,7 @@ class SysctlModule(object):
                 new_line = "%s=%s\n" % (self.args['name'], self.args['value'])
             else:
                 new_line = "%s = %s\n" % (self.args['name'], self.args['value'])
-            self.fixed_lines.append(new_line)
+            self.fixed_lines.append(new_line)                    
 
     # Completely rewrite the sysctl file
     def write_sysctl(self):
--- a/library/system/sysctl
+++ b/library/system/sysctl
@@ -100,7 +100,7 @@ import re
 class SysctlModule(object):
 
     def __init__(self, module):
-        self.module = module
+        self.module = module 
         self.args = self.module.params
 
         self.sysctl_cmd = self.module.get_bin_path('sysctl', required=True)
@@ -151,11 +151,11 @@ class SysctlModule(object):
             self.write_file = True
 
         # use the sysctl command or not?
-        if self.args['sysctl_set']:
+        if self.args['sysctl_set']:            
             if self.proc_value is None:
                 self.changed = True
             elif not self._values_is_equal(self.proc_value, self.args['value']):
-                self.changed = True
+                self.changed = True 
                 self.set_proc = True
 
         # Do the work
@@ -196,10 +196,10 @@ class SysctlModule(object):
     #   SYSCTL COMMAND MANAGEMENT
     # ==============================================================
 
-    # Use the sysctl command to find the current value
+    # Use the sysctl command to find the current value 
     def get_token_curr_value(self, token):
         thiscmd = "%s -e -n %s" % (self.sysctl_cmd, token)
-        rc,out,err = self.module.run_command(thiscmd)
+        rc,out,err = self.module.run_command(thiscmd)    
         if rc != 0:
             return None
         else:
@@ -230,7 +230,7 @@ class SysctlModule(object):
 
             rc,out,err = self.module.run_command(sysctl_args)
 
-        if rc != 0:
+        if rc != 0:            
             self.module.fail_json(msg="Failed to reload sysctl: %s" % str(out) + str(err))
 
     # ==============================================================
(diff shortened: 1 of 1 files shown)


mpdehaan:
when you say "works" can you share the previous behavior?

berenddeboer:
Example:

This is the error message on FreeBSD without this fix:

```
msg: Failed to reload sysctl: sysctl: unknown oid 'debug.trace_on_panic ' at line 12: No such file or directory
sysctl: unknown oid 'debug.debugger_on_panic ' at line 13: No such file or directory
sysctl: unknown oid 'vfs.nfsd.server_min_nfsvers ' at line 14: No such file or directory
sysctl: unknown oid 'kern.maxvnodes ' at line 15: No such file or directory
sysctl: unknown oid 'vfs.nfsd.tcphighwater ' at line 16: No such file or directory
```

dch:
TL;DR +1 to merge if _only_ the `"%s = %s\n"` fixes go in. Thanks @berenddeboer for tracking this down, it saved me some time.

This PR contains 3 separate changes, though:
1. correcting EOL whitespace
2. switching from the rc.d reload to the sysctl -f ... approach
3. removing the additional whitespace inside `"%s = %s\n"`

AFAICT only 3 is appropriate, can you re-submit with just that, optionally with 1 as a separate commit in the same PR if @mpdehaan cares.

The change in 2 doesn't work on less than FreeBSD 10, (see http://www.freebsd.org/cgi/man.cgi?query=sysctl&apropos=0&sektion=0&manpath=FreeBSD+9.3-stable&arch=default&format=html for example) as the `-f` flag is new.

If you test for /bsd/ instead of just freebsd then this will also work AFAICT (not tested)  on openbsd which also doesn't have the `-f` flag, nor the `-p` flag, see http://www.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man8/sysctl.8?&manpath=OpenBSD-current&sec=8&query=sysctl

berenddeboer:
Is EOL really important? I would say no file should have them. Sorry to complain about that, but it's a bit hard to turn off with my editor. So I would say that's an improvement.

On 2, the reload there doesn't work, when I have that, I get:

```
failed: [a.b.rootbsd.net] => {"failed": true}
msg: Failed to reload sysctl: 
```

The reason is that /etc/rc.d/sysctl reload always returns error code 1:

```
/etc/rc.d/sysctl reload ; setenv res $? ; echo $res
```

So we can either ignore the error code (not good), or do it my way.

berenddeboer:
People on #ansible explained the white space policy to me, so have re-added that. Obviously the git history looks a bit squashy, and it seems we have a merge conflict now. If you are happy with the current state, will clean this up in a new branch, as I don't know how to rebase properly.

Everytime I rebase I seem to bring in all the changes in devel in my new branch and I think they will show up as part of my commit (any tips really appreciated, in #ansible people suggest my sledgehammer approach (create new branch)).

mpdehaan:
The commits to re-instate whitespace are a little excessive in the end game of things, I'd really prefer to see here a commit that only fixed the problem in question, that way attribution on previous line numbers was not affected.

(removing and re-adding is the same problem as just removing when it comes to source code history, except made more so by extra commits in history to sift through...)

mpdehaan:
"switching from the rc.d reload to the sysctl -f ... approach"

This part seems unrelated to the commit and I think warrants some discussion, if this doesn't work on older BSDs, and is not required for newer, is a change required here?

berenddeboer:
Hi Michael, just will supply a clean patch without these white space changes. But will have to do that on a new branch, as it seems I can't rebase without introducing all new commits since the rebase into the history. for some reason.

berenddeboer:
It is _required_, else it will always fail with exit code one.

This may have changed since FreeBSD 10. The reason is that the reload script looks for two files: /etc/sysctl.conf /etc/sysctl.conf.local.

This is the portion of the script that does that:

```
    for _f in /etc/sysctl.conf /etc/sysctl.conf.local; do
            if [ -r ${_f} ]; then
                    ${command} ${command_args} ${_f} > /dev/null
            fi
    done
```

If I create /etc/sysctl.conf.local, the script does not return exit code 1.

But seeing the other platforms just reload a single file, we can do that on FreeBSD as well, and that will fix the issue too. So that's what I have converted it to.

If you want to, I'm happy to split these changes in two: 
1. first just the sysctl fixes to make sure it at least writes something FreeBSD accepts.
2. And then a second pull request, which changes the reload behaviour to just a single file.

Please let me know

mpdehaan:
From comments on ansible-project list:

I had first requested some whitespace commits not go in, and this was fixed by a revert - which doesn't solve the history problem.

It appears there are now about 12 other commits in there, so the best thing to do would be to resubmit this as a squashed patch containing just the one commit.

Commits that take spaces in and add them back completely miss the point we had originally - which was about not breaking attribution on those lines in source control history, when there was no need to change those lines.

So yeah, please submit a fix that only fixes this problem, as minimally as possible, ideally in one commit and we should be good to go.

PR #8259, opened 2014-07-23
Couchbase Server module

brianshumate:
This is an initial module for Couchbase Server. Presently, it provides the ability to initialize a Couchbase Server cluster, rebalance the cluster, and create data buckets.

### DIFF
--- a/library/database/couchbase_server
+++ b/library/database/couchbase_server
@@ -1,0 +1,268 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2013, Brian Shumate <brian@couchbase.com>
+
+DOCUMENTATION = '''
+---
+module: couchbase-server
+short_description: Couchbase Server administration module
+description:
+     - This module can execute Couchbase Server administrative commands,
+       change cluster configuration, and query various operational statistics
+version_added: '2.2.0'
+options:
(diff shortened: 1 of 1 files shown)


mpdehaan:
+version_added: '2.2.0'

Current version of ansible is 1.7 (which is the development branch), can you update this?

Also it looks like some commands like 'server-add' should _maybe_ be modelled as a resource (I'm not too high on miscellaneous command-bucket modules) and return changed: True only when changes are made.

I'm not positive what the right modelling for the various resources might be, some are definitely command oriented, but those that are simple wrappers to shell commands may be better served by just using the shell module, and we generally do not include modules that mostly wrap shell commands unless they provide extra logic to check changed=True/False and not run commands unless they need to be run, etc.

Can you perhaps stop by ansible-devel list if you'd like to discuss?

brianshumate:
Thanks for the great feedback!

Looks like I'd accidentally specified our product version vs. the Ansible version there... oops!

I am going to try and work with a colleague to improve this based on your suggestions, but will also try to get on the ansible-devel list to discuss further as well.

brianshumate:
After some discussion, we're going to remove the module from the Galaxy role, just use the `shell` module for now, and revisit this later.

Thanks!

Issue #8257, opened 2014-07-23
unarchive broken in 1.6.8 and later


frikkelfuzz:
##### Issue Type:

Bug Report
##### Ansible Version:

1.6.8
##### Environment:

debian wheezy
##### Summary:

unarchive fails with "duplicate parameter src"
##### Steps To Reproduce:

ansible wheezy1 -m unarchive -a "src=roles/tomcat/files/tomcat7-pocket.tar dest=/var/lib"
##### Expected Results:

success
##### Actual Results:

wheezy1 | FAILED >> {
    "failed": true, 
    "msg": "duplicate parameter: src (value=/root/.ansible/tmp/ansible-tmp-1406098204.14-81045059317860/source)"
}

mpdehaan:
This is already fixed on the development branch.  It only breaks in certain modes, and this fix will be applied in a future 1.6.9 likely out today.

See commit in #8263 which is applied.

PR #8263, opened 2014-07-23
fix for unarchive 8257

bennojoy:
This commit fixes the issue in #8257


mpdehaan:
Merged for unarchive, thanks!

PR #8270, opened 2014-07-24
Default for fstab missing, causing KeyError: 'fstab'

berenddeboer:
fstab misses default I think, get this:

failed: [a.b.rootbsd.net] => {"failed": true, "parsed": false}
  File "/root/.ansible/tmp/ansible-tmp-1406163834.52-79827425762471/mount", line 1641, in <module>
    main()
  File "/root/.ansible/tmp/ansible-tmp-1406163834.52-79827425762471/mount", line 272, in main
    if not os.path.exists(args['fstab']):
KeyError: 'fstab'

### DIFF
--- a/library/system/mount
+++ b/library/system/mount
@@ -244,7 +244,7 @@ def main():
             dump   = dict(default=None),
             src    = dict(required=True),
             fstype = dict(required=True),
-            fstab  = dict(default=None)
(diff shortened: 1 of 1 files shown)


mpdehaan:
Merged, thanks!

PR #8271, opened 2014-07-24
fix an issue that copying directory will fail when trying to change file attributes

tksmd:
As I already discussed with @jimi-c on the https://github.com/ansible/ansible/commit/84759faa0950146a6bae8452580b4a4cede6d871#commitcomment-7109780, copying directory will fail when trying to change file attributes and the target file already exists on remote.

I added brief testing and you can run the test as follows

```
$ cd test/integration
$ TEST_FLAGS="--tags=test_copy" make non_destructive
```

If you need more info, kindly let me know. Thanks!

### DIFF
--- a/lib/ansible/runner/action_plugins/copy.py
+++ b/lib/ansible/runner/action_plugins/copy.py
@@ -268,6 +268,7 @@ class ActionModule(object):
                 new_module_args = dict(
                     src=tmp_src,
                     dest=dest,
+                    original_basename=source_rel
                 )
                 if self.runner.noop_on_check(inject):
                     new_module_args['CHECKMODE'] = True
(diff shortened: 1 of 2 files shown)


bstinsonmhk:
Confirming the bug behavior and that the fix resolves for my use-case. Thanks @tksmd!

nathanhruby:
Is there a chance that this can be merged and quickly rolled into a release?  Since the only workaround is to disable permissions / ownership setting, which is a semi-security issue (or re-factoring job to do it in two passes) folks relying on this feature are in a pickle.

Thanks!

jimi-c:
Merged, thanks!

@nathanhruby we're going to be doing a 1.7 release soon (currently aiming for next week or the week after that, at the latest), so this will be included in that.

tksmd:
Thanks!

PR #8273, opened 2014-07-24
Add fstype arg to the mount examples

ps-jay:
Needed fstype in the examples, since fstype is a required argument for
mount.

### DIFF
--- a/library/system/lvol
+++ b/library/system/lvol
@@ -133,7 +133,7 @@ def main():
         # LVCREATE(8) -L --size option unit
         elif size[-1].isalpha():
             if size[-1] in 'bBsSkKmMgGtTpPeE':
-                size_unit = size[-1]
+                size_unit = size[-1].lower()
                 if size[0:-1].isdigit():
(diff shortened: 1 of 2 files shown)


mpdehaan:
Please split these into two pull requests as they are for different modules.  Usage of a topic branch for each pull request versus a common devel branch should allow us to merge things seperately.

When adding new modules also please add the new parameter to the DOCUMENTATION block at the top of the module so this can appear in the generated documentation.

Thanks!

ps-jay:
Hi Michael,
sorry about that - newbie mistake, didn't realise I needed multiple forks for multiple pull requests.
Is this okay now? Or would you like a brand-new pull request with just one commit in it?

Regards,
  Phil

ps-jay:
Actually, now that I've re-read CONTRIBUTING.md again, I realise I should have been using branches.
I'll resubmit.

ps-jay:
I think I've done it better this time... see: https://github.com/ansible/ansible/pull/8283

PR #8275, opened 2014-07-24
file: protect links from removal in state=absent

PinGwynn:
Added new optional parameter "protect=yes|<no>"
If protect option set in yes and state=absent, hard/symlinks will be protected from removal
Default value = no (if state=absent, dirs,files,links will be removed)

### DIFF
--- a/library/files/file
+++ b/library/files/file
@@ -62,6 +62,7 @@ def main():
             original_basename = dict(required=False), # Internal use only, for recursive ops
             recurse  = dict(default='no', type='bool'),
             force = dict(required=False,default=False,type='bool'),
+            protect = dict(required=False,default=False,type='bool'),
             diff_peek = dict(default=None),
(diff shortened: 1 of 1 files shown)


mpdehaan:
Hmm, I'm not clear I want this feature as this can be guarded against with the usage of the stat module.

I was originally thinking you were talking about recursive deletion protection, but this appears to not be it.

Also when adding a new module parameter be sure to always add it to the DOCUMENTATION block as well.

Can you stop by ansible-devel to discuss the nature of this one and how it might be best surfaced in module arguments, perhaps?

PinGwynn:
Thank you for your response, i'm already up this feature discussion in ansible-devel.

mpdehaan:
I think we came to the conclusion from ansible-devel that we would not do this one.

If anyone has questions on this feel free to open a new thread.

Thank you!

PR #8276, opened 2014-07-24
use recurse=yes for directory removal

btisdall:
The [documentation](http://docs.ansible.com/file_module.html) says that recursive directory removal should work, but the existing logic only works for directory creation because an exception is thrown unless state=directory. This PR fixes that while ensuring that recursive removals fail when the previous state is anything other than "directory".

NOTE: sorry, no tests yet. I'm at work and don;t have time to grok playbook testing.

### DIFF
--- a/library/files/file
+++ b/library/files/file
@@ -133,8 +133,11 @@ def main():
     changed = False
 
     recurse = params['recurse']
-    if recurse and state != 'directory':
-        module.fail_json(path=path, msg="recurse option requires state to be 'directory'")
+    if recurse:
+        if state not in ["directory", "absent"]:
+            module.fail_json(path=path, msg="recurse option requires state to be 'directory' or 'absent'")
+        if state == 'absent' and prev_state != 'directory':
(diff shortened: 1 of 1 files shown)


mpdehaan:
I believe directory removal is already recursive by nature. In which case this parameter just becomes something to be ignored.

Just wanted to make sure this was the intent of the code change, which it seems to be, which seems reasonable. 

Please confirm your thoughts.  Thanks!

btisdall:
Ah yes you're right, it looks as if the code will work just fine if recurse=no and indeed re-reading that is what the documentation says.

So it's just that current behaviour when recurse=yes confounds reasonable expectations (IMO) -  either the behaviour should change or the documentation should be more explicit.

mpdehaan:
I'm unclear on whether the above should still follow directories when they are symlinks.

In this case, most problems I think could be avoided by just clarifing in the docs the recurse=no parameter is not needed to be specified for directory removal, or better, adding code that unsets recurse for state=absent if accidentally set.

mpdehaan:
Hi!

Thank you very much!

PR #8279, opened 2014-07-24
Added nova_facts

CaptTofu:
Hi there! So, I needed nova_facts as well as docker_facts (another pull request). works perfectly and pep8-a-steptic

### DIFF
--- a/library/cloud/nova_facts
+++ b/library/cloud/nova_facts
@@ -1,0 +1,265 @@
+#!/usr/bin/python
+
+# vim: ts=4:expandtab:au BufWritePost
+# (c) 2014 Patrick "CaptTofu" Galbraith <patg@patg.net> 
+# Code also from rax_facts, docker_facts and the primary nova module
+#
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


mpdehaan:
So my usual question when all fact modules come up that I must ask - can we source these variables easily from the inventory script and avoid needing the facts module?

CaptTofu:
Michael,

Me and others on Monty’s team discussed this a few weeks ago and as useful as the dynamic inventory plugin is, it’s so much easier to have access to them using a facts module from within a playbook so as to have a means to do some useful things. Also, we may not always run plays to manage a fleet of instances using the inventory plugin hence where this facts module comes in handy. The others I’ve CC’d can also explain where it’s useful to them.

Thanks,

Patrick

emonty:
I agree with @mpdehaan for the most part - doing things as host_vars in the inventory is 1000000 times better, and if you are using nova_inventory, then this is completely and utterly pointless. However, there are times when you aren't using nova_inventory but the facts that nova_inventory would have injected into host_vars would still be useful for you. As a specific example, in our Infra systems, we're still using puppet (tons of existing code, migrating off all at once would be super painful) We're stepping one piece at a time towards using nova_inventory, but at the moment, we're using a local inventory module that queries the puppet cert list for machines. For us there, being able to call nova_facts without having to migrate fully to nova_inventory could be helpful.

I'd _actually_ like to take a pass through and make the  metadata reading that's used in nova_inventory librified so that the host_vars and the facts module produce exactly the same thing - but I haven't gotten around to that - I figured I'd get to that once the inventory patch landed.

mpdehaan:
Hi!

Thanks you very much!

PR #8282, opened 2014-07-24
Improve GCE dynamic inventory configuration handling

kevinoid:
Hey Ansible Developers,

Thanks for working on the GCE support, now that I've got it figured out it's really slick!  I had a bit of difficulty figuring out the authentication initially, so I've written this pull request to address a couple of the issues that complicated it for me.  I hope you find it useful:

The previous handling of configuration information for the GCE dynamic inventory plugin was sub-optimal in a couple ways:
- `gce.ini` was required even when `secrets.py` was on `PYTHONPATH`.
  The error message would be `"ConfigParser.NoSectionError: No section: 'gce'"`.
- Several classes of errors would be silently ignored, resulting in
  unrelated errors later in the script or empty values being used.
  (e.g. syntax errors in `secrets.py`, incorrect `$GCE_INI_PATH`)
- If `gce.ini` and `secrets.py` were missing, the error message was
  `"ConfigParser.NoSectionError: No section: 'gce'"`

This commit attempts to improve the error handling to address these
issues:
- `gce.ini` is no longer required if secrets.py can be imported
  without it.
- An error is raised if `$GCE_INI_PATH` is non-empty and can not be read.
- If neither `gce.ini` nor `secrets.py` could be loaded, an error message
  with a link to the documentation is printed.
- The logic for reading the preferences was moved into a new
  `_get_gce_driver_args` method to avoid deep nesting by returning early.

Let me know if you think any of these changes are unwarranted or if you'd prefer a different implementation.

Thanks,
Kevin

### DIFF
--- a/plugins/inventory/gce.py
+++ b/plugins/inventory/gce.py
@@ -111,49 +111,84 @@ class GceInventory(object):
         sys.exit(0)
 
 
-    def get_gce_driver(self):
-        '''Determine GCE authorization settings and return libcloud driver.'''
+    def _get_gce_driver_args(self):
+        '''Determine GCE authorization settings for the libcloud driver.'''
 
-        gce_ini_default_path = os.path.join(
-            os.path.dirname(os.path.realpath(__file__)), "gce.ini")
-        gce_ini_path = os.environ.get('GCE_INI_PATH', gce_ini_default_path)
-
-        config = ConfigParser.SafeConfigParser()
-        config.read(gce_ini_path)
-
-        # the GCE params in 'secrets.py' will override these
-        secrets_path = config.get('gce', 'libcloud_secrets')
-
-        secrets_found = False
+        # First, try secrets.py in PYTHONPATH
         try:
             import secrets
             args = getattr(secrets, 'GCE_PARAMS', ())
             kwargs = getattr(secrets, 'GCE_KEYWORD_PARAMS', {})
-            secrets_found = True
+            return (args, kwargs)
         except:
             pass
 
-        if not secrets_found and secrets_path:
+        config = None
+        gce_ini_path = os.environ.get('GCE_INI_PATH')
+        if gce_ini_path:
+            config = ConfigParser.SafeConfigParser()
+            try:
+                config.readfp(open(gce_ini_path))
+            except Exception, e:
+                err = "Unable to parse "
+                err += gce_ini_path
+                err += " from $GCE_INI_PATH: "
+                err += str(e)
+                print(err)
+                sys.exit(1)
+
+        gce_ini_default_path = os.path.join(
+            os.path.dirname(os.path.realpath(__file__)), "gce.ini")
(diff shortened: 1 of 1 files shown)


mpdehaan:
Hmm - So i think the weirdness is that there IS a secrets.py, and we should be able to put nearly everything in the INI file.  The idea of secrets.py itself is somewhat unusual compared to the other inventory plugin.   Other inventory scripts use the INI file, so that seems consistent.

Thoughts?

cc @lukesneeringer

erjohnso:
Have any other cloud modules started using libcloud? If so, keeping secrets.py might be useful, otherwise I think it's fine to dump it in favor of INI only.

kevinoid:
I'd be fine with dropping `secrets.py` as I don't have anything else which makes use of it.  I'm not sure how common it might be for users to want to share `secrets.py` between Ansible and the other `libcloud` applications.  Let me know if you'd like me to resubmit without `secrets.py`.

erjohnso:
Totally up to @lukesneeringer and ansible folks.

lukesneeringer:
> Hmm - So i think the weirdness is that there _is_ a `secrets.py`, and we should be able to put nearly everything in the INI file. The idea of `secrets.py` itself is somewhat unusual compared to the other inventory plugin. Other inventory scripts use the INI file, so that seems consistent.

I made some changes (not yet pushed) a couple of days ago to make it such that you don't have to use `secrets.py`, but the alternative I implemented is to use environment variables. I agree that the INI file is the preferred approach.

> Have any other cloud modules started using libcloud? If so, keeping `secrets.py` might be useful, otherwise I think it's fine to dump it in favor of INI only.

Not that I'm aware of. libcloud itself is really nice, but the `secrets.py` file is a very odd red herring; it's doubly difficult, I think, because it's not documented well.

#### A few general thoughts

A problem with dumping `secrets.py` entirely is that it's _there now_, which means that it's a backwards incompatibility to drop it. Is that something we're okay with, @mpdehaan?

My own feeling is that we should get the GCE plugin to work as much like the other plugin modules as possible, without breaking the way it works now, if that is feasible. That would mean, in my view, that the precedence order should be something like:
- environment variables
- INI file
- `secrets.py`

That would match the "environment variables trump INI file" precedence order that the other dynamic inventory plugins have, and leave `secrets.py` as an option for the people using it now.

amenonsen:
Any updates on this PR in light of the earlier discussion?

amenonsen:
@lukesneeringer @erjohnso @kevinoid Any updates here? Otherwise this PR should be closed now.

kevinoid:
@amenonsen I wouldn't mind rebasing the changes and adjusting it to whatever behavior is chosen for environment vs INI vs secrets.py, but I haven't used Ansible or GCE for a while, so it's not a big deal to me.  Feel free to close if you guys aren't concerned about it.

amenonsen:
I'm closing this per the earlier discussion.

PR #8201, opened 2014-07-19
Issue #7816 adding md5 checksum verification for get_url module

hahnicity:

PR #7639, opened 2014-06-02
get_url: Allow additional hashing algorithms

muffl0n:
Allow all hashing algorithms that hashlib provides (e.g. 'md5', 'sha1', 'sha224', 'sha256', 'sha384', 'sha512'). Introduces a new parameter "checksum" with format: `<algorithm>:<checksum>`, e.g. `checksum="md5:66dffb5228a211e61d6d7ef4a86f5758"`
Deprecates parameter `sha256sum`. Uses hashlib directly instead of methods in module_utils/basic.py.

Also gives helpful error messages:
`msg: Could not hash with algorithm 'md6'. Available algorithms: ('md5', 'sha1', 'sha224', 'sha256', 'sha384', 'sha512')`
`msg: The checksum parameter has to be in format <algorithm>:<checksum>`

See also #6304

PR #8294, opened 2014-07-25
Adds a md5sum option for get_url module

ewollesen:
Just a helper, many sites provide MD5, but not SHA256. This just makes it a little easier, so I don't have to download the file manually to get generate a SHA256 for future use.

### DIFF
--- a/library/network/get_url
+++ b/library/network/get_url
@@ -76,6 +76,14 @@ options:
     version_added: "1.3"
     required: false
     default: null
+  md5sum:
+    description:
+      - If a MD5 checksum is passed to this parameter, the digest of the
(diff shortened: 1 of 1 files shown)


flynfish:
:+1: awesome, I came here looking to see if this was an option. Hope they merge it soon!

muffl0n:
I already added an option to support all algorithms that are supported by hashlib. See PR #7639.

ewollesen:
So you did, looks great!

mpdehaan:
Currently preferring #8201 so let's work from there (I'm not seeing where this calculates the sum in particular), apologies on duplicate efforts!

PR #8295, opened 2014-07-25
Adding the following Test Coverage:

wrosario:
Use mysql_user module to create, delete users.
Update user password and ensure new password was updated for the correct user.
Assert user has access to multiple databases
Assert user creation, deleting using different user privilege and ensure privilege work correctly.


wrosario:
Made additional changes to mysql_user tests. Please review.  Thanks

jimi-c:
Merged, thanks!

PR #8300, opened 2014-07-26
Document regression in 1.6.9

techtonik:


jimi-c:
This has most likely already made its way out to packagers, and we generally only accept patches on devel. I'll be sure to add this note to the CHANGELOG there once I backport the updates to devel. Thanks!

PR #7476, opened 2014-05-21
implemented the missing "command" option, defaults to "install"...

geschke:
Hello!

I've written the previously documented, but missing "command" option in the composer module, because I needed the ability to do a composer update command. I'd be happy if you could review and add this feature to your repository. 

Thanks in advance & Kind regards,
  Ralf

PS. Updated with some documentation lines...

PR #8304, opened 2014-07-28
Fix composer module checkmode and change detection

komuta:
Hi,

While writing and testing playbooks using the composer module, I ran into two issues:
- check mode didn't work, even if it seems supported
- the changed status detection test was not valid

I propose this simple pull request to address those issues.

Regarding my fix proposal for check mode, I think it is suboptimal, because to me the root cause of the bug resides in the **for** loop which treats each module parameter. It might be a cleaner and safer way to build the options for composer by accessing explicitly each module parameters, as the _gem_ module does. Nevertheless, I thought this quicker proposal might be better suited for reviewing purpose.

### DIFF
--- a/library/packaging/composer
+++ b/library/packaging/composer
@@ -101,7 +101,7 @@ def parse_out(string):
     return re.sub("\s+", " ", string).strip()
 
 def has_changed(string):
-    return (re.match("Nothing to install or update", string) != None)
+    return 'Nothing to install or update' not in string
 
 def composer_install(module, command, options):
     php_path      = module.get_bin_path("php", True, ["/usr/local/bin"])
@@ -142,6 +142,8 @@ def main():
 
     # Prepare options
     for i in module.params:
+        if (i == 'CHECKMODE'):
+            continue
         opt = "--%s" % i.replace("_","-")
         p = module.params[i]
         if isinstance(p, (bool)) and p:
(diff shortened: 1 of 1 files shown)


mpdehaan:
The way that checkmode is being detected here is different from other modules, can you please update this to be consistent?

Thanks!

komuta:
Yes, I'll update the module.

komuta:
Hi,

Sorry for the delay. I've updated the source code of the module to be more consistent.

komuta:
I noticed you merged #7476 meanwhile, so I rebased my branch to the latest devel, to include the command parameter.

komuta:
Hi,

I rebased my branch again today. Could you merge it, so it can be available in next release 1.8? Thanks!

ghost:
composer has been moved to https://github.com/ansible/ansible-modules-extras/

komuta:
Thanks for the info. I've created a new request in the extras repository: https://github.com/ansible/ansible-modules-extras/pull/19.
If everything is ok, should I close this pull request?

techtonik:
Yes. This PR can be closed.

komuta:
Request split to https://github.com/ansible/ansible-modules-extras/pull/19 and https://github.com/ansible/ansible-modules-extras/pull/94.

PR #8307, opened 2014-07-28
Fix extra newline at end of hostname for Fedora

robinro:
This fixes a bug introduced by 138b45e3.
The hostname has an additional newline at the end which leads to the
state always being 'changed: true' even if the hostname is unchanged.


jimi-c:
Merged, thanks!

PR #8311, opened 2014-07-28
Wrap get_distribution_version() in the hostname module

neersighted:
We wrap get_distribution_version() with a new function,
_get_distribution_version(), that returns `0` when the result is a string or
`None`.

This accounts for the case when get_distribution_version() returns a string,
and we try to compare it to a float. We do this in the hostname module instead
of the module snippets because other modules may want the real string
version.module snippets because other modules may want the real string version.

### DIFF
--- a/library/system/hostname
+++ b/library/system/hostname
@@ -45,6 +45,18 @@ from distutils.version import LooseVersion
 from ansible.module_utils.basic import *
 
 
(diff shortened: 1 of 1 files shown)


neersighted:
As per IRC discussion.

/cc @mpdehaan

mpdehaan:
Merged, thank you!

PR #8322, opened 2014-07-29
Print is a function, Lets show some respect

lathama:
In an effort to understand Ansible code better I started to work on adding Python3 support. Many projects like Apache's libcloud have found the conversion to be much easier than they had feared. This pull request only changes the print to be a function. Many stderr calls were also addressed in a proper way and all tests pass fine. This is me jumping into the development cycle to swallow yet another OSS project process so point out my failures as they happen. I may have softened at least one "wtf" found in print output, sorry about that.
"""
Ran 165 tests in 6.667s

OK
"""

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ ASCII2MAN = @echo "ERROR: AsciiDoc 'a2x' command is not installed but is require
 endif
 
 PYTHON=python
-SITELIB = $(shell $(PYTHON) -c "from distutils.sysconfig import get_python_lib; print get_python_lib()")
+SITELIB = $(shell $(PYTHON) -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())")
 
 # VERSION file provides one place to update the software version
 VERSION := $(shell cat VERSION)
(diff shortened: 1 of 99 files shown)


mpdehaan:
We're very interested in Python 3 support but are going to approach this at a later time in the future.

Right now it's more important to deal with the pending queue of submissions.

There has been some nice work using "six" for ansible-tower-cli, so we expect that would be used for joint 2.6+ and 3+ support when we decide to do this though naturally module coverage is going to be something that is not vital now.

Naturally when distros support only python 3 or choose to default to python 3 (if they do, though this is really a completely different language due to compatibility breaks and such) we'll be ready for it.

In all thanks for the effort, though I wish this was discussed more on ansible-devel first to spare efforts, but we simply can't deal with all the forced rebasing this would cause at this time.  We will probably make the change when the time comes, but I do not feel that has to be now, as python 2.X will be supported for a very very long time, and again, we're cognizant of those dates, so no one should be concerned.

I suspect things will be massively refactored and architecturally different at this time and we'd naturally add a tox.ini to make sure the test battery ran on both versions, and configure Jenkins to run the integration test battery against both versions.

PR #8325, opened 2014-07-29
added option for associate_public_ip_address to ec2_lc module

mschurenko:
I need this option for using launch configs in VPC.  More information can be found here:

http://boto.readthedocs.org/en/latest/ref/autoscale.html#module-boto.ec2.autoscale.launchconfig


mschurenko:
I didn't do the params option correctly. Will fix and submit another pull request.

PR #8326, opened 2014-07-29
adding option to associate_public_ip_address to ec2_lc

mschurenko:
I needed this option for VPC. More info can be found here:

http://boto.readthedocs.org/en/latest/ref/autoscale.html#module-boto.ec2.autoscale.launchconfig

### DIFF
--- a/library/cloud/ec2_asg
+++ b/library/cloud/ec2_asg
@@ -264,14 +264,15 @@ def delete_autoscaling_group(connection, module):
 
         instances = True
         while instances:
-            groups = connection.get_all_groups(names=[group_name])
+            groups = connection.get_all_groups()
             for group in groups:
                 if group.name == group_name:
                     if not group.instances:
+                        target_group = group
                         instances = False
             time.sleep(10)
 
-        group.delete()
+        target_group.delete()
         module.exit_json(changed=True)
     else:
         module.exit_json(changed=False)
--- a/library/cloud/ec2_asg
+++ b/library/cloud/ec2_asg
@@ -264,7 +264,7 @@ def delete_autoscaling_group(connection, module):
 
         instances = True
         while instances:
-            groups = connection.get_all_groups()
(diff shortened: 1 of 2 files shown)


mschurenko:
I don't know if there's a way for me to create another pull request or if this one has to be close first? Anyways my last commit is for the ec2_asg module. Currently it does not delete asg properly because when it makes a call with get_all_groups() it forgets to add the name of the asg so sometimes you get a response from a different asg and then it fails because it tries to delete that asg. This could be a very serious bug if it were able to delete that wrong asg. Luckily it will fail if there are running ec2 instances in the group.

mschurenko:
Actually I was incorrect with my last commit. get_all_groups() is called with no arguments; however only the result that contains the asg name is used. So there is still a bug where the asg cannot be deleted even after all the instances have been terminated. This could be related to some kind of scaling activity that is occurring at the time delete is being called. This is a very hard bug to reproduce. It will work like 20 times in a row then suddenly will fail 5 times in a row.

jimi-c:
The new param needs a `version_added: "1.7"` field.

mschurenko:
Is that correct?

jimi-c:
@mschurenko no, the field you modified is the one which shows version in which the entire module was added. Each parameter can have a similar field, so yours would be:

<pre>
 +  associate_public_ip_address:
 +    description:
 +      - whether to assign a public IP address to each instance launched in a Amazon VPC
 +    required: false
 +    default: false
 +    aliases: []
 +    version_added: "1.7"
</pre>

So you'll need to remove the previous change and add the field above.

mschurenko:
Is that better?

mschurenko:
This fixes the issue where ec2_asg sometimes tries to delete the wrong asg when state=absent. I'm not sure if I am doing all of this properly as I don't have much experience contributing to github projects. Right now this pull request is for two separate issues with two separate ansible modules. Let me know if you want this done in a different manner. I just didn't know how.

Thanks

PR #8327, opened 2014-07-29
Fixes a regression in 1453f7b

jsmartin:
Fixes a regression in 1453f7b013393678b8a616d7cfbda1b68b2c31db that causes unnecessary wait_timeouts in certain module operations when a given IP address could not be found.  Now wait_timeouts are only active during address associations of newly provisioned EIPs.

Cleanup and comment of main().

### DIFF
--- a/library/cloud/ec2_eip
+++ b/library/cloud/ec2_eip
@@ -41,6 +41,11 @@ options:
     required: false
     default: false
     version_added: "1.6"
+  wait_timeout:
+    description:
+      - how long to wait in seconds for newly provisioned EIPs to become available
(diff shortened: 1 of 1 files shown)


jsmartin:
Added support for setting wait_timeout for newly provisioned EIPs.  Default is 300 seconds.

mpdehaan:
This is missing the "version_added" attribute on the new parameter.

jsmartin:
Latest push adds the version_added attribute in the docs.

jimi-c:
Merged, thanks!

Issue #7254, opened 2014-05-02
file module doesnt remove existing directory even if force is set


landryb:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.6 (devel c6f3a0a4cc) last updated 2014/05/02 09:59:57 (GMT +200)
##### Environment:

Debian/OpenBSD
##### Summary:

file module doesnt remove existing directory even if force is set
##### Steps To Reproduce:

Usecase: i have a /var/lib/lxc directory, that i want to make a symlink pointing to /srv/lxc/root.

```
file: src=/srv/lxc/root dest=/var/lib/lxc state=link force=yes
```
##### Expected Results:

/var/lib/lxc dir is removed, and is made a symlink to /srv/lxc/root
##### Actual Results:

doesnt work, because the file module only removes the destination if it's a file, not if it's a dir.

```
failed: [tiger] => {"failed": true, "gid": 0, "group": "root", "item": "", "mode": "0755", "owner": "root", "path": "/var/lib/lxc", "size": 4096, "state": "directory", "uid": 0}
msg: Error while replacing: [Errno 21] Is a directory
```

jimi-c:
@bcoca I believe this used to explicitly fail in this situation, as we did not want users accidentally wiping out directories of files.

landryb:
well, if the user used force=yes...

jimi-c:
When using force with links it signifies that we want to rewrite the link, not that we'll force a state transition say from a directory to a link, as that's a very fast way to damage a system.

jimi-c:
The above commit should address this in the following way: If the directory is not empty, the module will refuse to replace it even if force=yes. However, if the directory is empty, the module will replace it when force=yes is specified. If you have any further questions regarding this, please let us know. Thanks!

landryb:
Agreed, i think it makes sense to solve it this way. Thanks!

cdosborn:
As a result of this decision, If I want to create a symlink I must first check if a non-empty directory exists. `force: True` should respect this.

PR #8329, opened 2014-07-29
file: prevent replace failure when overwriting empty directory with hard or soft link on force=yes

rdolejsi:
Issue #7254 finalization (reopen) - empty directory replacement with a link was unable to be performed (Error while replacing: [Errno 21] Is a directory).


jimi-c:
Merged, thanks!

PR #8334, opened 2014-07-29
Expand variables in configuration files, take two

veeti:
Let's try this again.

The `remote_tmp` variable shouldn't be expanded in the first place, right? It can lead to some confusing behavior. For example, I might have put `remote_tmp = ~/.ansible/bla` in my configuration. If I use a user called `veeti` on my own computer and execute an ansible module on a server as `root` it'll try to write to `/home/veeti/` on the server.

Is there something else failing after this? My tests pass.


mpdehaan:
I think this has nailed the question of why that was so confusing a few times before, thanks very much!

mpdehaan:
Merged, thanks!

PR #8345, opened 2014-07-30
Suppport for kerberos\domain authentication for winrm

bennojoy:
This patch allows ansible to authenticate to winrm server via kerberos for Domain users.

Pre-requisite steps on Ansible host:

1) kinit user@domain.com  [The user gets a valid kerberos ticket to authenticate]

2) Set the windows variables in ansible. 
 ansible_ssh_user: ansible_kerberos
 ansible_ssh_pass: ansible_kerberos
 ansible_ssh_port: 5986
 ansible_connection: winrm 
Please make sure the username and password is set to 'ansible_kerberos' which tells ansible to use the kerberos ticket for authentication.

Pre-requisite steps on Windows host:

1) Follow the steps on http://docs.ansible.com/intro_windows.html under the section "Window sys prep"

2) Make sure that Kerberos and Negotiate authentication is enabled on WinRm service. 
'winrm g winrm/config/service/auth'

3) Verify connection  with WinRM service using Kerberos.
'winrm id -r:hostname.somedomain.com -a:Kerberos -u:domainusername@somedomain.com -p:'

4) Make sure there are no firewall blocking access to winrm service from ansible host.
on ansible host: 'telnet <host> 5986/5'

Notes:
kinit is provided with the package 'krb5-workstation' in centos/Rhel

An example /etc/krb5.conf file

```
[logging]
 default = FILE:/var/log/krb5libs.log
 kdc = FILE:/var/log/krb5kdc.log
 admin_server = FILE:/var/log/kadmind.log

[libdefaults]
 default_realm = BENNO.COM
 dns_lookup_kdc = false
 ticket_lifetime = 24h
 renew_lifetime = 7d
 forwardable = true
 udp_preference_limit = 1

[realms]
 BENNO.COM = {
  kdc = WIN-EUO6PCEBU37.benno.com
  admin_server = WIN-EUO6PCEBU37.benno.com
  default_domain = benno.com
 }

[appdefaults]
validate=false

[domain_realm]
 .benno.com = BENNO.COM
 benno.com = BENNO.COM

```

Please also make sure the hostnames given in ansible are available in both FORWARD and REVERSE DNS lookups.

### DIFF
--- a/lib/ansible/runner/connection_plugins/winrm.py
+++ b/lib/ansible/runner/connection_plugins/winrm.py
@@ -72,7 +72,10 @@ class Connection(object):
         if cache_key in _winrm_cache:
             vvvv('WINRM REUSE EXISTING CONNECTION: %s' % cache_key, host=self.host)
             return _winrm_cache[cache_key]
-        transport_schemes = [('plaintext', 'https'), ('plaintext', 'http')] # FIXME: ssl/kerberos
+        if self.user == 'ansible_kerberos' and self.password == 'ansible_kerberos':
+            transport_schemes = [('kerberos', 'https'), ('kerberos', 'http')]
+        else:
+            transport_schemes = [('plaintext', 'https'), ('plaintext', 'http')] # FIXME: ssl/kerberos
         if port == 5985:
             transport_schemes = reversed(transport_schemes)
         exc = None


trondhindenes:
+1

mpdehaan:
Somebody on the list was also asking about this, I'll share with them so we can get some joint review.

I think the PR should also add the above to the Windows setup docs in a "Kerberos" section, ideally, as this may be hard for people to guess.

Thanks!!!

walthowd:
I tested this with ansible 1.7 (devel 189824dd76) under Ubuntu 14.04.1 and it worked against a large selection of Windows machines over the type of SSL endpoint currently recommended in the documentation. 

The only issue I saw is that I receive 20 to 25 lines of the following every time I run a win_ping:
User username@DOMAIN.COM authenticated successfully using Kerberos authentication (EDIT -- Looks like that is coming from pywinrm here - https://github.com/diyan/pywinrm/blob/master/winrm/transport.py#L164 )

If the PR includes the docsite changes, you should include the fact that you need python-dev and libkrb5-dev installed on Debian based systems, and that you need the Python Kerberos module (pip install kerberos).

bennojoy:
yeah.. those lines are due to the print statement in transport.py of winrm, i will try to send in a patch upsteam.

ianclegg:
Great see this is being worked on! All, a number of engineers will be operating in regulated sectors where they may need different credentials for different hosts. One thing I would like to see in this feature is a way to specify the identity of the user to be used.

I believe that it we 'shore up' pywinrm it should be possible without any changes whatsoever to ansible. Of course this will depend on the pywinrm's appetite for enhancement.

Ultimately, thinking about desired behaviour, may I suggest the following:

**Kerberos single sign on**
`ansible_ssh_user: <not specified>`
`ansible_ssh_pass: <not specified>`

Its safe to say the only sensible thing we can do here is try kerberos with the current users UPN aka Single Sign On (this what you were doing with 'ansible_kerberos' / 'ansible_kerberos' combination).

**Kerberos with a specific user (UPN)**
`ansible_ssh_user: user@realm`
`ansible_ssh_pass:`

This is not a valid local windows username, but is a valid UPN, so its kerb. There is no password so we cant acquire a TGT. So check the ticket cache. If do not have a valid TGT for that identify we should prompt or fail.

**Kerberos with a specific username password**
`ansible_ssh_user: user@realm`
`ansible_ssh_pass: password`

We have a password this time so we can silently acquire a TGT if a valid one does not already exist.

pywinrm doe not yet distinguish very well between transports and authentication, if we can address this there it would be a more elegant solution. I will look into this now.

bennojoy:
Totally agree, we should support all the three options also as said the logic should rather be in pywinrm  than Ansible. 
The reason for choosing 'ansible_kerberos/ansible_kerberos' as username/password  rather than null was becuase there is a caching mechanism in our winm connection plugin which caches the transport/authentication object based on the username/password and i didnt want to bypass that.

ianclegg:
@bennojoy I hadn't seen that caching, but it shouldn't be hard to address

cchurch:
@bennojoy It wouldn't be a big deal to remove the winrm caching.  It only works for a single host (i.e. forks=1) anyways at the moment.

I've updated my fork of pywinrm (https://github.com/cchurch/pywinrm) to comment out the print statement.  Upstream pywinrm hasn't yet accepted my pull request for Python 2.6 compatibility either, so I'm not sure how actively he's maintaining that library.

trondhindenes:
I'm not sure I agree with the logic around deciding whether to use Basic or Kerberos.
A local (e.g. non-domain) user in Windows has the Netbios form "REALM\UserName" (where realm is the local computer), for example 'SERVER1\Administrator'. 
A Domain user in windows can be specified using either netbios (DOMAIN\UserName) or UPN (username@subdomain.domain.tld)

A UPN-form domain user can use either kerberos or basic auth. If I was designing this I would attempt kerberos first (given that the username specified is UPN-style) (or possible check whether we have a kerb ticket), and then failing that fallback to basic and attempt that.

The USER@REALM form proposed above doesn't really belong anywhere in windows, as it's a mix between netbios and upn-style usernames. I doesn't feel right to me.

ianclegg:
@trondhindenes 
Your reasoning is sound, but I'm not so sure about the facts and I hope to dissuade you :)

> A UPN-form domain user can use either kerberos or basic auth.

This is not the case, I have tried. WinRM will not authenticate a domain user with basic auth. This the main reason for implementing support for kerberos. See the constant 'WSManFlagUseBasic' on
http://msdn.microsoft.com/en-us/library/aa384293(v=vs.85).aspx

> The USER@REALM form proposed above doesn't really belong anywhere in windows

Having worked at Microsoft and being MSCE certified, this is news to me. The UPN form is the preferred format for logon's since the adoption of Kerberos in Windows 2000. See 'userPrincipalName' description (2nd paragraph) at http://msdn.microsoft.com/en-us/library/ms677605(v=vs.85).aspx. See MSCE Training _'MCSE Planning, Implementing, and Maintaining a Microsoft Windows Server 2003 Active Directory Infrastructure'_. It is a fact that the SAM based Domain/User form is retained for mainly for compatibility.

Does this change anything?

trondhindenes:
Bad explanation on my part. Given AD domain "xyz.com" which has a netbios domain name of "xyz", I would consider the following username forms valid:
XYZ\administrator
administrator@xyz.com

and the following username form not valid (or at least very weird):
administrator@xyz

ianclegg:
@trondhindenes 
I was proposing
`user@realm`

where realm would naturally be the kerberos realm (the FQDN) rather than a netbios name. So don't worry: administrator@xyz would not be valid anyway.

The salient point here is to implement support for netbios domains, as in your example:
`netbios domain\user`

There are a lot of Windows Administrators who are used to this form and I agree it would make sense. However, performing kerberos with netbios names is a Microsoft extension. I would not like to 'assume' the netbios name is the same as the realm since there are many situations (after a business merges) where this may not be the case.

Do you know if netbios naming is supported by the underlying python kerberos package (which basically wraps the MIT krb5 library) or how we would go about this?

trondhindenes:
I'm trying to test this on my system. I'm guessing I need the Kerberos python package, which fails to install on my (ubuntu) node. Has anyone seen this error?
Command /usr/bin/python -c "import setuptools, tokenize;**file**='/tmp/pip_build_root/kerberos/setup.py';exec(compile(getattr(tokenize, 'open', open)(**file**).read().replace('\r\n', '\n'), **file**, 'exec'))" install --record /tmp/pip-UmWh_C-record/install-record.txt --single-version-externally-managed --compile failed with error code 1 in /tmp/pip_build_root/kerberos
Storing debug log for failure in /home/azureuser/.pip/pip.log
azureuser@ansible-tower:~$

bennojoy:
@trondhindenes  did you try installing it via apt instead of pip ?

apt-get update
apt-get install python-kerberos

ianclegg:
Hmm...just looking into this - so pywinrm uses: https://pypi.python.org/pypi/kerberos
Except its not really a kerberos library, but a GSS-API implementation (SPNEGO) - looks like its built on an some 2010 source's from Apple. The sources indicate it is not possible to set a specific UPN.

PyKerberos (https://pypi.python.org/pypi/pykerberos)
Looks a little more promising, the sources look more up to date (2013 Apple sources) and it has support for setting a specific the principal.

`maj_stat = gss_import_name(&min_stat, &principal_token, GSS_C_NT_USER_NAME, &name);`

There is a mention of an NT username, I guess we need to check that out

trondhindenes:
Yup, got it working. So, one problem I'm seing straight away in the as-is code is that ansible tries to use https even if the port specified is 5985 (http).

trondhindenes:
@ianclegg 
I completely agree that for Kerberos support, UPN-style usernames (user@domain.com) would be the only supported form. Windows folks won't have a problem with that.

bennojoy:
@trondhindenes  According to code logic if port is 4985 it will first try http and kerberos and if it fails it would try https.
can you please run it with -vvvv and check if a connection with http is tried ?

jhawkesworth:
Just been testing this out against a win 2008 r2 sp1 box and hit System.OutOfMemory error when running win_ping.  Turns out this is a winrm issue - there's a hotfix for it here: http://support.microsoft.com/kb/2842230

I am still getting to know ansible but I wondered if instead of the 'magic' ansible_kerberos username and password it would be possible to add a new ansible_connection type, something like 'winrm-domain' which would leave the ansible_ssh_user and ansible_ssh_pass settings free to specify relevant credentials.  I imagine in big setups you might need to use different domains and users for different environments, so I think it would be necessary to have a way of selecting user/domain/realm to be used for a given host group.

I pulled ansible from source this morning and manually applied this patch and found if port=4985 then http is tried and if port=4986 then https was tried.

trondhindenes:
Maybe some retry logic should be done here as well, similar to the ssh connection types. On Windows, PowerShell defaults to Kerberos, but lets you override that by specifying the "-Authentication" parameter. The downside would be (I'm guessing here) an even longer time to set up the WinRM connection (which is already more than slow enough compared to the blazing speed of ssh - that's WIndows' fault and not Ansible's tho).

nicodeslandes:
Just to let everyone know, I started a topic on this subject on ansible-devel: https://groups.google.com/forum/#!topic/ansible-devel/TYMdAoN9IK8

themalkolm:
So.. is it possible to use kerberos/domain authentication in ansible?

jhawkesworth:
I have not tested this yet, but it appears this has recently been added to
latest dev version of ansible now via this pull request:

https://github.com/ansible/ansible/pull/10202

Jon

bcoca:
closing this in favor of the merged #10202, we can add features on top of that

PR #8040, opened 2014-07-05
template overrides with cleansing

bcoca:
This should allow overriding template interpolation/execution with data sanitation based on the overriden values

PR #8347, opened 2014-07-30
add 'variable_start_string', 'variable_end_string' to jinja2 specail header parameters

dave-re:
I will use to "variable_start_string" and "variable_end_string".


bcoca:
this defeats the whole purpose of the security patch that disabled it in
the first place​, PR #8040 adds this capability back but still trying to
keep the security intact

mpdehaan:
Yep, we'll need to go with the above.   Closing this for the reasons @bcoca stated above.

Thanks!

PR #8348, opened 2014-07-30
adding stat parameters to stat documentation file

baddev:
Hi, 

I thought that it would be better to find all stat parameters for its variable on the Ansible docs page. So here it is.

Please tell me if needed improvements.

Best regards

### DIFF
--- a/docsite/.gitignore
+++ b/docsite/.gitignore
@@ -11,5 +11,5 @@ ansible*.xml
 .buildinfo
 objects.inv
 .doctrees
-rst/modules/*.rst
+#rst/modules/*.rst
 *.min.css
--- a/docsite/rst/modules/stat.rst
+++ b/docsite/rst/modules/stat.rst
@@ -24,7 +24,7 @@ Options
 
 .. csv-table::
    :header: "parameter", "required", "default choices", "comments"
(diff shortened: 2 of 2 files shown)


mpdehaan:
Thanks for this, but the modules dir is for code generated files only.  Can you update the DOCUMENTATION block for the stat module and add some of this into the EXAMPLES free form section, or the notes block?

Thanks very much, just need to keep this organized.  Much appreciated!!

baddev:
On 07/30/2014 09:59 PM, Michael DeHaan wrote:

> Thanks for this, but the modules dir is for code generated files only.
> Can you update the DOCUMENTATION block for the stat module and add some
> of this into the EXAMPLES free form section, or the notes block?

Ok I've done.. Something. I'm not sure if this what you asked but the
way ansible-doc is organized I couldn't fit the parameter list properly.
So I added a little function.

I did the most clear I could. You even have a unittest script to see
what it does with random values.

If it's really too awful : just say and I will return to something less
over my head. :)

## Cheers,

“One original thought is worth a thousand mindless quotings.”
“Le vrai n'est pas plus sûr que le probable.”

```
                                          Diogene Laerce
```

PR #8341, opened 2014-07-29
ssh - make known_hosts check compatible with entire format

AlexRiedler:
# Fix Known Hosts SSH Issues
## Existing issues
- Currently the implementation works by looking up the `known_hosts` file (by assuming it is in the default location) this is not always the case.
- It does not check the `ssh_config` which may have HostAlias's or specially set identities
- It does not support ? and \* operators
- It assumes the host is in the file even if it is a prefix of another host
- It does not support different ports with different host keys (e.g. forwarded ports with different end machines)
## Solution

This fix works by using `ssh-keygen` and its `-F` parameter to determine if the set of potential hosts is available. See code for better details.

Also Fixes: https://github.com/ansible/ansible/issues/8333

Always open for debate and changes :smiley_cat:

PR #8349, opened 2014-07-30
ssh - Support for other known_hosts formats in not_in_host_file()

### LINKED ISSUE #8333: Tasks run in serial when using same ansible_ssh_host with different ansible_ssh_port's.
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.6.10
##### Environment:

Debian Sid managing Centos 6
##### Summary:

A standard environment when running using Vagrant is that it creates an inventory file at .vagrant/provisioners/ansible/inventory/vagrant_ansible_inventory with something like the following:

> $ cat example-hosts-broken
> www1 ansible_ssh_host=127.0.0.1 ansible_ssh_port=2222 ansible_ssh_user=root
> www2 ansible_ssh_host=127.0.0.1 ansible_ssh_port=2200 ansible_ssh_user=root
> db1 ansible_ssh_host=127.0.0.1 ansible_ssh_port=2201 ansible_ssh_user=root
> db2 ansible_ssh_host=127.0.0.1 ansible_ssh_port=2202 ansible_ssh_user=root`

When using this inventory file, all actions seem to take place in a serial fashion. i.e. tasks take 4x as long as they should!
##### Steps To Reproduce:

Using the hosts file above, run something like `ansible all -m shell -a "sleep 5 && uptime" -f 10 -i example-hosts-broken` and note that the output is delayed by at least 5 seconds between each report despite `-f 10` and no serial keyword. Note that it takes 5 seconds for each vm to report.
##### Expected Results:

If you run it against dedicated addresses, without reusing them, you get expected behaviour returning after 5 seconds:

> $ cat example-hosts
> www1 ansible_ssh_host=192.168.33.21 ansible_ssh_user=root
> www2 ansible_ssh_host=192.168.33.22 ansible_ssh_user=root
> db1 ansible_ssh_host=192.168.33.41 ansible_ssh_user=root
> db2 ansible_ssh_host=192.168.33.42 ansible_ssh_user=root
> $ ansible all -m shell -a "sleep 5 && uptime" -f 10 -i example-hosts
> db1 | success | rc=0 >>
>  12:17:09 up 59 min,  0 users,  load average: 0.00, 0.01, 0.00
> 
> db2 | success | rc=0 >>
>  12:17:17 up 59 min,  0 users,  load average: 0.04, 0.03, 0.00
> 
> www2 | success | rc=0 >>
>  12:17:09 up  1:00,  0 users,  load average: 0.00, 0.00, 0.00
> 
> www1 | success | rc=0 >>
>  12:17:09 up  1:01,  0 users,  load average: 0.00, 0.00, 0.00

(Again, vm's clocks aren't perfect but they all returned at the same time, at 5 seconds)

I would expect this behaviour in the above example too.
##### Actual Results:

> $ ansible all -m shell -a "sleep 5 && uptime" -f 10 -i example-hosts-broken 
> db1 | success | rc=0 >>
>  12:15:40 up 58 min,  0 users,  load average: 0.01, 0.02, 0.00
> 
> www2 | success | rc=0 >>
>  12:15:45 up 59 min,  0 users,  load average: 0.00, 0.00, 0.00
> 
> www1 | success | rc=0 >>
>  12:15:51 up 59 min,  0 users,  load average: 0.00, 0.00, 0.00
> 
> db2 | success | rc=0 >>
>  12:16:03 up 58 min,  0 users,  load average: 0.01, 0.02, 0.00

(VM clocks aren't perfect above, but there are 5 seconds between each report.)

PriceChild:
Several other formats in known_hosts files are valid including [host]:port which aren't taken into account currently.

Uses pull request #8341 (46e31a66a3325ccbbe5d48402ffa217e75160218) as inspiration for this simpler implementation.

Fixes #8333

### DIFF
--- a/lib/ansible/runner/connection_plugins/ssh.py
+++ b/lib/ansible/runner/connection_plugins/ssh.py
@@ -203,54 +203,19 @@ class Connection(object):
         stdin.close()
         return (p.returncode, stdout, stderr)
 
-    def not_in_host_file(self, host):
-        if 'USER' in os.environ:
-            user_host_file = os.path.expandvars("~${USER}/.ssh/known_hosts")
-        else:
-            user_host_file = "~/.ssh/known_hosts"
-        user_host_file = os.path.expanduser(user_host_file)
-        
-        host_file_list = []
-        host_file_list.append(user_host_file)
-        host_file_list.append("/etc/ssh/ssh_known_hosts")
-        host_file_list.append("/etc/ssh/ssh_known_hosts2")
-        
-        hfiles_not_found = 0
-        for hf in host_file_list:
-            if not os.path.exists(hf):
-                hfiles_not_found += 1
-                continue
-            try:
-                host_fh = open(hf)
-            except IOError, e:
-                hfiles_not_found += 1
-                continue
-            else:
-                data = host_fh.read()
-                host_fh.close()
-                
-            for line in data.split("\n"):
-                if line is None or " " not in line:
-                    continue
-                tokens = line.split()
-                if tokens[0].find(self.HASHED_KEY_MAGIC) == 0:
-                    # this is a hashed known host entry
-                    try:
-                        (kn_salt,kn_host) = tokens[0][len(self.HASHED_KEY_MAGIC):].split("|",2)
-                        hash = hmac.new(kn_salt.decode('base64'), digestmod=sha1)
-                        hash.update(host)
-                        if hash.digest() == kn_host.decode('base64'):
-                            return False
-                    except:
-                        # invalid hashed host key, skip it
-                        continue
-                else:
-                    # standard host file entry
-                    if host in tokens[0]:
-                        return False
-
-        if (hfiles_not_found == len(host_file_list)):
-            vvv("EXEC previous known host file not found for %s" % host)
+    def not_in_host_file(self, host, port=22):
+        if port != 22:
+            # Although the key for host may be present and valid for [host]:port and cause this check
+            # to fail, you have to ask why we've been asked to connect to a random other port. There's
+            # no easy way around this and the user should 'ssh-keygen -f ~/.ssh/known_hosts -R $host'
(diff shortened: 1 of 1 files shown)


AlexRiedler:
I should probably note that the reason I used certain functions (e.g. didn't use with, etc) was because of python 2.4 support in ansible.

PriceChild:
Thanks @AlexRiedler . Guess I'd misread contributing! Please close this pull request.

mpdehaan:
Ansible actually needs 2.6 or higher for the control machine, not 2.4 or higher.

Do we have competing pull requests on this?  I'm not sure I should close this one for the 2.4 compliance reason as it does not apply.

mpdehaan:
Also there appears to be no PR for the linked commit above, if there is one, can you please supply the ticket number?  

If not, I think this is ok and fine, and is a simple enough patch to test.   We want to make sure the performance is not adversely affected for large host counts that may be calling this 500 times in parallel though, which may make this solution not a good one (or one where it would need a boolean in ansible.cfg to enable this option).

PriceChild:
@mpdehaan subprocess.check_output() doesn't appear until 2.7. According to @AlexRiedler ssh-keygen's return code cannot be relied upon over other platforms to denote success/failure in finding a key either. This means 2.5's subprocess.check_call() couldn't be used to make things even simpler instead.

In short, though I was hoping to simplify the process it doesn't look possible so we should give up on this PR.

As for 500 times in parallel... yes that is a worry and I was initially hoping to just lightly modify the existing code. An even simpler PR would just add the post arguments to the method and then the `if port != 22: host = "[%s]:%i" % (host, port)` line rather than delegating to subprocess/ssh-keygen. There seemed to be some good arguments from @AlexRiedler over why ssh-keygen would be a better solution though, catching more odd environments.

The rest of the logic 'could' remain as is.

And just noticed I could close it myself...

PR #8353, opened 2014-07-30
return the most up-to-date info about an instance when starting/stopping it

gdamjan:
otherwise the module will return the info about the instance that it got prior to the action taken
So if you had a task to start an instance:

```
  ec2:
     instance_ids: ...
     state: running
  register: ec2_info
```

the registered data would have empty public_dns_name, public_ip, private_dns_name, private_ip


mpdehaan:
Merged, thanks!

PR #8354, opened 2014-07-30
More info in "not a legal parameter" error

msabramo:
Without this, I was getting:

```
ERROR: apt is not a legal parameter in an Ansible task or handler
```

which doesn't give me any idea what task ansible is having trouble with
(I reference "apt" many times in many roles)

With this, I get:

```
ERROR: "apt" is not a legal parameter in an Ansible task or handler
    task name: "Update apt-get cache"
    line: "apt: update_cache=yes cache_valid_time=86400"
    role_name: "common"
```

which points me right at the task that ansible is having trouble with.

### DIFF
--- a/lib/ansible/playbook/task.py
+++ b/lib/ansible/playbook/task.py
@@ -108,7 +108,15 @@ class Task(object):
                 ds['when'] = "%s %s" % (when_name, ds[x])
                 ds.pop(x)
             elif not x in Task.VALID_KEYS:
-                raise errors.AnsibleError("%s is not a legal parameter in an Ansible task or handler" % x)
+                if 'name' in ds:
+                    task_name = ds['name']
+                else:
+                    task_name = '(No name)'
+                line = "%s: %s" % (x, ds[x])
+                raise errors.AnsibleError(
+                    '"%s" is not a legal parameter in an Ansible task or handler\n'
+                    '    task name: "%s"\n    line: "%s"\n    role_name: "%s"'
+                    % (x, task_name, line, role_name))
 
         self.module_vars  = module_vars
         self.default_vars = default_vars


mpdehaan:
I like this idea, but am not sure I want to display the hash data, because most people are thinking in YAML and that's effectively a python datastructure (nor would I want to show YAML).   Perhaps we could show the task name if set, otherwise show the task line.   Though that might require a bit more logic.

Ultimately though, this is an indentation error, so this is a module search path issue.

msabramo:
Thanks. That's valid I suppose about showing the python data structure. I am a Python programmer, so I'm totally cool with it and didn't give it a second though, but I guess a lot of folks can use ansible without knowing a lick of Python, so I can see your point that the messaging should preferably be Python-agnostic.

My original idea was to show the file and line #, but I didn't see how to get at that. I'll take a stab at changing it to do as you say - that's just as informative and friendlier to a wider variety of people.

This isn't actually an indentation error. The playbook is fine. What's happening in this specific case is that ansible is having trouble finding modules because I'm hacking around with the `pip install -e` thing and I didn't run the `hacking/env-setup` script. But that's probably immaterial because this error will probably be displayed when ansible is installed correctly and the playbook is messed up, which is surely the more common case.

msabramo:
Rebased this to display the info a little more structured and not just spitting out the raw Python dict.
New output looks like this:

```
ERROR: "apt" is not a legal parameter in an Ansible task or handler
    task name: "Update apt-get cache"
    line: "apt: update_cache=yes cache_valid_time=86400"
    role_name: "common"
```

msabramo:
@mpdehaan: Can I do anything else to make this better?

resmo:
+1

msabramo:
:sleeping:

msabramo:
Ping @jimi-c, @bcoca

cemo:
+1

msabramo:
Ping

msabramo:
My general experience with Ansible has been that it's great when it works, but it is very hard to debug when it doesn't work. I think more information about errors would be a welcome addition.

amenonsen:
I think this PR should be closed now, the underlying code has changed completely. Pity, it was a useful idea to begin with, modulo the comments about displaying the hash data above.

msabramo:
Agreed that this PR should be closed (note: I was the submitter). I tried to resolve the merge conflict a few weeks ago and it was non-trivial.

I'm not even sure this is needed anymore, as error reporting seems to be a lot better in 2.0.

PR #8360, opened 2014-07-30
Work on nova inventory plugin- pep8 plus connects to recent client

CaptTofu:
Fixed this plugin so that it works with current nova client library. I also cleaned up cruft and made it pep8 and pyflakes happy.

I intend to add features to it next and make it even more useful.

### DIFF
--- a/plugins/inventory/nova.py
+++ b/plugins/inventory/nova.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 # (c) 2012, Marco Vito Moscaritolo <marco@agavee.com>
+# (c) 2014 Patrick "CaptTofu" Galbraith <patg@patg.net> 
 #
 # This file is part of Ansible,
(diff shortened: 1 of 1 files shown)


mpdehaan:
As a project policy, we only apply pep8 fixes ourselves, because this preserves attribution which is important for git blame and makes code review easier.   (even if we did, we'd require this as a seperate commit, ideally broken up into lots of commits...but we don't).

Can you please submit this with only the "current nova version" fixes?

Thanks very much!

CaptTofu:
Hi there!

So, I realized that the rax inventory plugin had the features I wanted and completely reworked this to be mostly fixes to make it work the way the rax pluging does and the way people would want it to work. There isn't much code left of the original version and pep8 is coincidental at this point as I have to make it so otherwise editing I get a lot of errors from vim :)

PR #8363, opened 2014-07-30
Fix Docker image name parsing

### LINKED ISSUE #8357: docker module parses image names incorrectly
##### Issue Type:

Bugfix Pull Request
##### Ansible Version:

ansible 1.7 (devel 189824dd76) last updated 2014/07/30 12:12:50 (GMT -400)
##### Environment:

N/A
##### Summary:

The docker module parses image names incorrectly in `get_split_image_tag`.  Docker images can be of the form "registry/username/image", so `image` should be set to `image.split('/')[-1]`.  It may also be worth keeping the username, since it may be a meaningful namespace distinction.
##### Steps To Reproduce:

```
- name: Create a few of my_image
  docker: image=my_registry.example.com/myself/my_image
              count=3

- name: Create a few of my_other_image
  docker: image=my_registry.example.com/myself/my_other_image
              count=3
```
##### Expected Results:

There should be 6 containers, 3 of each type.
##### Actual Results:

Since both actions interpret the image name as "myself", only 3 containers total are created, of image "my_image".

greenmoss:
Fixes #8357

### DIFF
--- a/library/cloud/docker
+++ b/library/cloud/docker
@@ -470,7 +470,7 @@ class DockerManager:
 
     def get_split_image_tag(self, image):
         if '/' in image:
-            image = image.split('/')[1]
+            image = image.split('/')[-1]
         tag = None
         if image.find(':') > 0:
             return image.split(':')


jimi-c:
Merged, thanks! If you'd like to follow-up with another PR to preserve the username portion somehow, I do think that would be a good addition.

PR #8364, opened 2014-07-30
Logging of module invocation for huge datasets to ansible module is very slow

piyush0101:
TL;DR: 
Ansible is inefficient in logging the invocation of a module if input data to the module is pretty huge. This is because of a very general regex expression.

Full Story:

I was collecting some data from several thousand hosts and passing it on to an AnsibleModule. Module was very slow. I did some profiling with cProfile and pstats and found that invocation of __log_invocation() method call in the class AnsibleModule is taking most of the time. It is doing expensive regular expression matching trying to obfuscate sensitive data like passwords from ansible syslogs. For complex input data which in this case is a dictionary of all hosts for a region it does not scale. The problem is the regular expression which is very general and open which results in a lot of potential matches and for a large dataset it becomes very slow. Here's the output from the profiler.

update_regions.prof% sort cumulative
update_regions.prof% stats 10
Tue Jul 29 17:50:26 2014 update_regions.prof

```
 54670 function calls (53455 primitive calls) in 215.633 seconds
```

Ordered by: cumulative time
List reduced from 959 to 10 due to restriction

ncalls tottime percall cumtime percall filename:lineno(function)
1 0.002 0.002 215.633 215.633 update_looksee_dfw:32()
1 0.000 0.000 215.579 215.579 update_looksee_dfw:67(main)
1 0.000 0.000 213.086 213.086 update_looksee_dfw:254(**init**)
1 0.045 0.045 213.055 213.055 update_looksee_dfw:848(_log_invocation)
131 212.851 1.625 212.851 1.625 {method 'match' of '_sre.SRE_Pattern' objects}

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -873,7 +873,7 @@ class AnsibleModule(object):
         filter_re = [
             # filter out things like user:pass@foo/whatever
             # and http://username:pass@wherever/foo
-            re.compile('^(?P<before>.*:)(?P<password>.*)(?P<after>\@.*)$'), 
+            re.compile('^(?P<before>.*user(name)?:)(?P<password>.*)(?P<after>\@.*)$'), 
         ]
 
         for param in self.params:


mpdehaan:
I'm not sure the above regex is accurate, but this may be skimming too fast, does this look for the explicit string "user"?  This seems incorrect.

mpdehaan:
I do agree the regex could be smarter, definitely.

mpdehaan:
Also curious if you could supply sample input data, if possible.

piyush0101:
This looks for texts like these anywhere in the string. Would that be good enough?
user:pass
username:pass

piyush0101:
Let me see if I can get the data.

piyush0101:
So, I've tested with these patterns. These are matched by the above regex.

"user:pass@foo/whatever"
"http://username:pass@wherever/foo"

piyush0101:
Data is a hostvars dictionary with an entry like this. Consider a dictionary with about 7000 of host entries.
{"hostvars": {"host1": {"pstack": {"running": "875.4", "symlinked": "885.0", "tars": [], "versions": ["885.0"]}} }}

piyush0101:
I cannot provide you the exact data but here is a script that would generate almost the same. This data goes as input to the module.

``` python
hostvars = {'hostvars': {}} 

for i in range(1, 7000, 1): 
    host_facts = {'host%s' % i:
                  {'pstack':
                   {'running': '875.1',
                    'symlinked': '880.0',
                    'tars': [], 
                    'versions': ['885.0']}}}
    hostvars['hostvars'].update(host_facts)

print hostvars
```

abadger:
@piyush0101 I think what mpdehaan is wondering is whether the new regex will catch things like 

```
abadger:password123@github.com
https://randomperson:randomchars@wherever/foo
```

It appears that the new regex is specifically looking for the string "user" but this code needs to be more generic and able to catch people's real choices of username and password.

Looking at the current code, I also think it will miss things like:

```
urls= ['https://abadger:password@wherever/datasource1', 'https://abadger:password@wherever/datasource2']
```

My testing finds it letting the first password through because the regex engine obeys the greediness rule and so it finds the last ":" that is followed by an @.

This may be a case where a regex is too slow but using the linear search methods on a string object is acceptable.

abadger:
Preliminary testing shows that using str.partition() can be made extremely fast on the sample data.  Checking if performance holds up when there's several username:password entries in the dataset before implementing for real.

abadger:
@piyush0101 if you'd like to try out that commit, I think that it will address your issue.  I'm just checking on whether the unittests are desirable before merging.

PR #8365, opened 2014-07-30
Parameters to disable_rollback are "true" or "false"

thisdavejohnson:


mpdehaan:

PR #8368, opened 2014-07-30
Proposing a Debugging section for intro_windows

DamonOverboe:
I'm thinking something pretty simple like this to handle a few of the issues I hit.
1. The Error 401 code is discussed [here](https://groups.google.com/forum/#!topic/ansible-project/dyOC1cWX90w). If that's something that everyone is seeing (I did with Server 2012 VM edition) then it might need to be moved up in the document.
2. The `-vvvv` is a bit obscure I think if you're new to Ansible. I'm assuming most will have some experience with Linux and will try just `-v` which is what I tried. I then remembered from a demo that `-vvvv` was what I wanted.
3. The local admin account is something I glossed over when reading the instructions. I figured I could substitute my account for testing, since I'm an admin on that box. So calling that out could help.

### DIFF
--- a/docsite/rst/intro_windows.rst
+++ b/docsite/rst/intro_windows.rst
@@ -254,16 +254,18 @@ Debugging
 Start with the win_ping test. If that fails, pass -vvvv to get verbose output.
 
 
-If you get a 401 error code, log into your Windows client and run the following:
+If you get a 401 error code, log into your Windows client and run the following::
+
 
     winrm set winrm/config/service/auth @{Basic="true"}
+
 
 Then test the ping from your Linux Controller again.
 
 
-Note that currently, you must use a local administrator account on the Windows
+*Note that currently, you must use a local administrator account on the Windows
 client. Support for Kerberos and domain accounts has received a pull request
-but isn't integrated quite yet.
+but isn't integrated quite yet.*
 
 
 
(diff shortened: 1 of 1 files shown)


mpdehaan:
Possibly good as a start, though I wouldn't want to merge anything that refers to pull requests right now.  Let me think about this one.

We'll have part of this soon, and I've also modified the docs recently to mention a new setup script that should remove the need for that first command.  Maybe.

DamonOverboe:
Yeah, I didn't expect that this would be how you would want it to go, but thought it might spur some thoughts or discussion.

I also thought of one other error; I encountered $EDITOR not being set in my ~/.bash_rc, and that was throwing the ErrorNo2 if I recall correctly.

Also it might make more sense to name it Troubleshooting instead of Debugging. Ideally, I would think you'd want to grow that into a document / page of it's own.

mpdehaan:
I'm still open to a troubleshooting section, though I think we can close this one _for now_ because the stuff about auth setup should be dealt with by the new setup script linked in the docs.

If you'd like to discuss, hop on ansible-devel and I'd be glad to.

Thanks!

DamonOverboe:
I had to put this down for a few days, but will do on the discussion.

I'm going through getting a new target set up, using the script, and I'm hitting issues. I'll bring those up in ansible-devel.

PR #8028, opened 2014-07-03
Added support for general purpose SSD EBS volume_type

iainlbc:
This adds support for AWS general purpose SSD volumes: http://aws.amazon.com/ebs/details/ 
- volume_type module parameter added which defaults to standard if one is not specified
- if standard and iops are both specified, volume_type is changed to io1 for backwards compatibility (as opposed to requiring the user to now specify volume_type in plays) .
- if gp2 volume_type is specified with an iops value, the play will error. Defaulting this to io1 as well seemed assuming to much since it's reasonable that one could mix up he ssd types, and there are performance/cost ramifications between the two types. 

Not a huge fan of assuming if iops are specified the volume_type should be silently changed to io1 from standard, but it seemed best for maintaining backwards compatibility. 

Thanks!
iain

PR #8369, opened 2014-07-31
Modified ec2_vol to use the new SSD based storage

phillipholmes:
The older magnetic storage was referenced as volume_type = 'standard'
The new SSD based general purpose storage is now referenced as volume_type = 'gp2'
The change proposed will leverage the new storage without needing to retrofit the module to have volume_type added as a required type and expect people to know its moved from 'storage' to 'gp2'.


jimi-c:
This is a duplicate of #8028, so closing. Thanks!

PR #7983, opened 2014-06-29
ec2_ami_search: support for SSD and IOPS provisioned EBS images

lorin:
In ec2_ami_search module, add support for the new SSD-backed EBS (ebs-ssd)
volume as well as images with provisioned IOPS (ebs-io1) that Ubuntu images now support.

For details, see this [mailing list announcement](https://lists.ubuntu.com/archives/ubuntu-cloud-announce/2014-June/000055.html).

PR #8371, opened 2014-07-31
ec2_ami_search: Added support for ssd and io1 AMI images

shintonik:
Until now only ebs and instance-store as story types were supported. There are another two store types for AMI images (ssd, io1). Added those options.


lorin:
I think this is a dupe of #7983

mpdehaan:
yeah I think so, so closing this one, please comment on the other ticket if there are concerns.  Thanks!

PR #8382, opened 2014-07-31
add doc to stat module + function to display doc

baddev:
Hi, 

I had to make a function in ansible-doc to display parameters list properly.

Best regards

PS: tmp_stat is a working directory


mpdehaan:
I really would not like to see a seperate doc page for the stat module (the new stat.rst), and would like to see this accounted for in the inline documentation for the stat module if possible.  

If you'd like to discuss, feel free to stop by one of the mailing lists.

Keeping this open to take a look at the ansible-doc parts and possibly cherry-pick some things,but have not had a chance to look yet.

Thanks!

baddev:
- consolidated the code to prevent any unexpected freeze

mpdehaan:
Oh nice, this is on the right track I think.  Still need to test but attacking some other priorities at the moment.  

I'm thinking a better name for parameters might be "returns"?

mpdehaan:
The output could also be structured, so really just being able to nest a JSON example in the YAML and show it as free form text may help other modules more.

The other problem is for some modules, the returns differ based on the parameters.

baddev:
Hi,

On 08/11/2014 09:52 PM, Michael DeHaan wrote:

> Oh nice, this is on the right track I think. Still need to test but
> attacking some other priorities at the moment.

Thanks, no worries. :)

> I'm thinking a better name for parameters might be "returns"?

That's ok for me. May I ask why ?

> The output could also be structured, so really just being able to nest a
> JSON example in the YAML and show it as free form text may help other
> modules more.

I'm not sure to understand what you mean by this, is it :

```
- substitute the (list of lists) in the
  old:parameters//new:returns DOC definition, by a JSON
  formatted list of parameters
```

or
    - instead of the (line of a table) as an output, that the
      format_param function should return a JSON formatted (line or
      whole table) ?

> The other problem is for some modules, the returns differ based on the
> parameters.

Could you please give me an example ?

## Thanks for your time,

“One original thought is worth a thousand mindless quotings.”
“Le vrai n'est pas plus sûr que le probable.”

```
                                          Diogene Laerce
```

baddev:
Hi,

So I kept it simple as I thought I heard you whispered. ;)
YAML has been used as you suggested (thanks for the module dev doc I finally understood :) )
Cheers

bcoca:
had not seen this PR before, currently resolved through a similar approach in #10295

PR #8385, opened 2014-07-31
ec2 module: document valid states

lorin:
Document the valid values for the state parameter in the ec2 module.

Also, fix a typo.


mpdehaan:

PR #8389, opened 2014-08-01
ansible-pull: Improved inventory handling when inventory does not yet exist

### LINKED ISSUE #8324: Ansible-pull fails if inventory has not been downloaded
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.7 (devel aee940aaca) last updated 2014/07/29 12:05:46 (GMT +1000)
##### Environment:

N/A
##### Summary:

ansible-pull does not work if the inventory has not yet been downloaded. 
##### Steps To Reproduce:

You want to run ansible-pull to download a remote repository and then run a playbook using inventory in that playbook

```
ansible-pull -d /var/tmp/ansible-playbooks -i /var/tmp/ansible-playbooks/inventory/hosts -U http://git.internal/ansible-playbooks -m git path/to/playbook.yml
```
##### Expected Results:

A successful ansible-pull run
##### Actual Results:

Starting ansible-pull at 2014-07-29 12:09:48
ERROR: Unable to find an inventory file, specify one with -i ?

willthames:
Fixes #8324
Improves #7667

As discussed with @mpdehaan on ansible-devel.

### DIFF
--- a/bin/ansible-pull
+++ b/bin/ansible-pull
@@ -49,6 +49,7 @@ import time
 from ansible import utils
 from ansible.utils import cmd_functions
 from ansible import errors
+from ansible import inventory
 
 DEFAULT_REPO_TYPE = 'git'
 DEFAULT_PLAYBOOK = 'local.yml'
@@ -150,7 +151,9 @@ def main(args):
     now = datetime.datetime.now()
     print >>sys.stderr, now.strftime("Starting ansible-pull at %F %T")
 
-    if not options.inventory:
+    # Attempt to use the inventory passed in as an argument
+    # It might not yet have been downloaded so use localhost if note
+    if not options.inventory or not os.path.exists(options.inventory):
         inv_opts = 'localhost,'
     else:
         inv_opts = options.inventory


mpdehaan:
This isn't the behavior I had discussed on ansible-devel actually:

What I had suggested was:

-i is specified:
    pass the -i along to the git checkout command only if the file exists
    otherwise pass "localhost," to the git checkout command

always pass -i to the playbook (the ansible-playbook execution) as given if specified
otherwise pass "localhost,"

In general, we prefer things to be more explicit than relying on try/except logic.

I think I'd like that better

willthames:
Ok, updated as suggested.

willthames:
@mpdehaan can the needs_revision label be removed please?

mpdehaan:
I'll do one better than that and merge it in.

Thanks!

mpdehaan:
Hmm, might not have pushed this one yet... will investigate.

mpdehaan:
Corrected - thanks!

PR #8393, opened 2014-08-01
Fix exponential regex performance issue

gnosek:
`filter_leading_non_json_lines` effectively does

```
re.match(".*\w+=\w+.*", line)
```

for every line of output. This has abysmal performance in case of large
Base64-encoded data (which ultimately does not match the regex but does
match the .*\w+= part) as returned e.g. by the template module (diffs).

Replacing the match with

```
re.search("\w=\w", line)
```

drops the complexity back to linear, and actually usable with large
diffs from the template module (a 150 KB Base64 diff kept Ansible
spinning at 100% cpu for minutes).

Also, check the easy cases (line.startswith) first while we're here.

Closes: #8932

### DIFF
--- a/lib/ansible/utils/__init__.py
+++ b/lib/ansible/utils/__init__.py
@@ -1041,11 +1041,11 @@ def filter_leading_non_json_lines(buf):
     filter only leading lines since multiline JSON is valid.
     '''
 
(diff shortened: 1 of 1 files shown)


mpdehaan:
Looks reasonable, in queue for testing, thank you!

mpdehaan:
Merged, thanks!

PR #8396, opened 2014-08-01
Load callback plugins in the ansible script

Glandos:
This is an open discussion, since I'm far from an expert on ansible.
However, since I often run some commands on a lot of host, I've written a small callback plugin that summarize output on stdout and stderr.
I need this callback to be called when running ansible (not the playbook), since I only run once a command, and forget it after that.
Maybe the modified line is not enough, too much, or at a wrong location. I just copied it from the ansible-playbook script.

### DIFF
--- a/bin/ansible
+++ b/bin/ansible
@@ -38,6 +38,7 @@ class Cli(object):
     def __init__(self):
         self.stats = callbacks.AggregateStats()
         self.callbacks = callbacks.CliRunnerCallbacks()
+        callbacks.load_callback_plugins()
 
     # ----------------------------------------------
 


mpdehaan:
I'm not positive this may have the intended effect for all users, enabling callbacks that may make assumptions for users who just run them with playbooks for non-playbook invocatons.  But it might be ok.  Can you perhaps start a thread on ansible-devel list?

Glandos:
See https://groups.google.com/forum/#!topic/ansible-devel/8TaWL05aw0M for discussion.

alxgu:
I guess this PR can be closed. It should be solved by https://github.com/ansible/ansible/pull/8911

Glandos:
You're right. In fact, the final commit was in https://github.com/ansible/ansible/commit/2064d260853ae001d0c8c1ccd2d8cf30e36e55df because it change the setting name. But it now fulfills my needs :+1:

PR #8398, opened 2014-08-01
Adding test Coverage for mysql_variable module

wrosario:
Use mysql_variable to query, set and update variables.
Assert using user and password to query, set and update variables.
Assert using single quotes, double quotes and no quotes when using variables


mpdehaan:
Merged, thanks!

PR #8400, opened 2014-08-01
"allow_no_value=True" for ini_file module

nicocesar:
I use init_file to edit my.cnf that has values that are only keys. 

Could you please update this? 

Thanks,
Nico

### DIFF
--- a/library/files/ini_file
+++ b/library/files/ini_file
@@ -94,7 +94,7 @@ import ConfigParser
 def do_ini(module, filename, section=None, option=None, value=None, state='present', backup=False):
 
     changed = False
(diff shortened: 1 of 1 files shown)


mpdehaan:
Merged, thanks!

mpeters:
This seems to break ini_file on python < 2.7. See #8430

nicocesar:
Maybe checking  sys.version_info  will make the trick... I'll do a pull request in short

mpdehaan:
revert applied.  I'll happily take updates to allow this to work on newer versions.  Thanks @nicocesar

PR #8402, opened 2014-08-01
Envirionment variable support in GCE inventory plugin.

lukesneeringer:
This pull-request adds support for environment variables in the GCE inventory plugin.

It is somewhat orthogonal to the yet-to-be-accepted #8282. We will need to figure out how we want to handle them both.


mpdehaan:
Merged, thanks!

PR #8403, opened 2014-08-01
Azure inventory improvements.

lukesneeringer:
This PR adds several improvements to the Azure inventory plugin:
- Makes it more resilient against cache files not being specified.
- Adds `_meta` to output.
- Adds `--host` as a flag.

### DIFF
--- a/plugins/inventory/windows_azure.py
+++ b/plugins/inventory/windows_azure.py
@@ -71,6 +71,11 @@ class AzureInventory(object):
         self.read_environment()
         self.parse_cli_args()
 
+        # Cache setting defaults.
+        self.cache_path_cache = '/tmp/ansible-azure.cache'
+        self.cache_path_index = '/tmp/ansible-azure.index'
+        self.cache_max_age = 0
+        
         # Initialize Azure ServiceManagementService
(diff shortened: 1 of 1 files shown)


mpdehaan:
I'm noticing that above the cache path seem hard coded in /tmp whereas with the ec2 module this will use a location in the user's homedir.  This seems like what we'd want to mirror here.

Note in usage with Tower, we specifiy a different CWD each time and the cache won't take effect, where as in this usage with Tower, it's likely the cache WOULD take effect, since there's not a way to override the cache location from a config variable.

Anway, this is a note that we need a way to override the cache dir, and it should probably not pick /tmp, but something along the lines of what the ec2 module does.

lukesneeringer:
The override method already exists (see the `read_settings` method), and the cache is disabled unless a max age is given (since it defaults to 0).

I'm pushing a commit to move the default from `/tmp` to `~`, however. :)

amenonsen:
@cchurch are you already looking into this?

cchurch:
@amenonsen #11446 includes these changes and more -- I just need to address @abadger's comments on that PR.

bcoca:
closing via merged #11446

PR #8405, opened 2014-08-01
Update version.

wenottingham:


mpdehaan:
Good catch, I really should take the version out of there if I forget to update it that often :)

Thanks!

PR #5590, opened 2014-01-12
tests: added travis.yml to run unit test of pull requests on travis-ci.org

resmo:
I would like to do another try to integrate the free for open source projects unit test service travis-ci.org into ansible.
## Why?
- It makes people more sensible implementing unit test, if they see, each and every pull request will be tested against regression errors.
- does not cost a dime.
- transparency as travis is well integrated into github.
- easy and simple, like ansible.
## How?
1. sign in on travis-ci.org using your github account (oauth)
2. go to https://travis-ci.org/profile.
3. activate the ansible/ansible repo.
4. merge this PR.
## Preview how it would look:

https://github.com/resmo/ansible/pull/1

PR #8407, opened 2014-08-02
Setup continuous testing with travis, do coding style check

fangpenlin:
I really like to contribute more on ansible, however, it's kinda painful to read ansible's source code, so many [PEP8](http://legacy.python.org/dev/peps/pep-0008/) violations, unused imports, different spacing, line way way way too long and so on, since there is basically no coding style check.

When this project comes bigger and bigger, I believe these bad coding style will soon become heavy technical debts that all ansible developers need to pay. I am not saying we should stick to all PEP8 rules, but at least, many reasonable basic rules should be ensured.

Hence, I wrote a `.travis.yml` file, for ensuring tests are ran against every single commit. It also checks the coding style.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,3 +12,4 @@ script:
   - 'nosetests test/units -sv --with-coverage'
   - 'pip install flake8 --use-mirrors'
   - 'flake8 lib/ansible --ignore=E501,W293'
+  - 'flake8 test/units --ignore=E501,W293'
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,4 +11,4 @@ install:
 script: 
   - 'nosetests test/units -sv --with-coverage'
   - 'pip install flake8 --use-mirrors'
-  - 'flake8 ansible --ignore=E501,W293'
+  - 'flake8 lib/ansible --ignore=E501,W293'
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,5 +11,5 @@ install:
 script: 
   - 'nosetests test/units -sv --with-coverage'
   - 'pip install flake8 --use-mirrors'
-  - 'flake8 lib/ansible --ignore=E501,W293'
-  - 'flake8 test/units --ignore=E501,W293'
+  - 'flake8 lib/ansible --ignore=E501,E221,W293'
+  - 'flake8 test/units --ignore=E501,E221,W293'
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,15 @@
 language: python
 python:
   - '2.7'
+env:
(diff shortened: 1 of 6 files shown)


fangpenlin:
After cleaning runner module up, I have some thoughts.

## Enable pep8 checking in your editor

By enabling [SublimeLint](https://github.com/lunixbochs/sublimelint), I can spot style errors easily. Here I found two undefined variables.

https://github.com/ansible/ansible/pull/8407/files#diff-af869caf5f6f46706dbf35970e39c686R1094
https://github.com/ansible/ansible/pull/8407/files#diff-af869caf5f6f46706dbf35970e39c686R1137

I strongly suggest to let your editor do the syntax checking for you.

## Breaks too long dict, lis, function calls and function definitions

It's very common in ansible to have so many arguments for a function, I suggest they should all arguments should be broken into individual lines if they exceed 80 chars. For example

``` python
def foobar(a, b, c, d, ....):
    pass
```

Since the line def foobar exceeds 80 chars, so it should be written in

``` python
def foobar(
    a,
    b,
    c,
    d,
    ....
):
    pass
```

not just for function, either list, dict should follow the same rule. This makes it easier not just to read and also easier to write.

fangpenlin:
will add some more ideas about coding style later when I have spare time

sivel:
There are a few things to be aware of:
1. travis support is very specifically not enabled.

You may want to check out https://github.com/ansible/ansible/search?q=travis&ref=cmdform&type=Issues

There are at least 5 closed issues/PRs for this. I believe the following best sums the responses up: https://github.com/ansible/ansible/pull/5590#issuecomment-32266958
1. pep8 PRs are not to be submitted, see the following URL for more info about pep8 as it pertains to ansible:

https://github.com/ansible/ansible/blob/devel/CODING_GUIDELINES.md#pep8-and-basic-style-checks

mpdehaan:
If you think Ansible's coding style is "bad", you really haven't read much bad code.  It's quite good and we are quite big on code review.  An automated style checker is only relevant at the most basic level, concerns of architecture, intent, and variable naming are not things checked by the basic "pep8" tool, and religious following of it doesn't really do any good in the end run.

Adherance to 80 character line widths in the age of widescreen monitors in particular is a silly thing.

As mentioned above, we do remove unused imports and run a very large subset of PEP8 and pyflakes replacements on an occasional basis, but preserving attribution is vitally important and we would never accept an untrusted widescale diff to replace wide portions of the codebase.  That would be irresponsible to trust and not do ourselves.

We do integration tests through Jenkins and many of our integration tests spin up new cloud nodes and integrating these with Travis with passing OS credentials to them would not be appropriate.

PR #8410, opened 2014-08-02
fail module: Add an "all" argument to trigger a fatal error

anoutsider:
Adds an optional "all" parameter to the "fail" module, which modifies it's behavior to trigger a fatal error for the whole playbook.

this was implemented without raising an ansible error, we only set the remaining_hosts to False and stop the rest of the task execution, and it's triggered by passing a "Fatal" keyword in the module return data.

This PR originated here : https://groups.google.com/forum/#!topic/ansible-project/gDzrZVQ3AMw

An integration test is included under test/integration/test_fail.yml 

```

---
- hosts: local
  vars:
    fail_run: "YES"

  tasks:
    - name: test fail with all
      fail: msg="Failed" all=yes
      when: fail_run == "YES"
      tags: ["fail_with_all"]    

    - name: test fail with all
      debug: msg="Dummy run"
      tags: ["fail_with_all"]  

    - name: test fail without all
      debug: msg="Dummy run"
      tags: ["fail_without_all"]  

    - name: test fail without all
      fail: msg="Failed"
      when: fail_run == "YES" 
      tags: ["fail_without_all"]

    - name:
      debug: msg="Still running"
      tags: ["fail_with_all", "fail_without_all"]
```

### DIFF
--- a/lib/ansible/runner/action_plugins/fail.py
+++ b/lib/ansible/runner/action_plugins/fail.py
@@ -18,6 +18,7 @@
 import ansible
 
 from ansible import utils
+from ansible import errors
 from ansible.runner.return_data import ReturnData
 
 class ActionModule(object):
@@ -39,6 +40,7 @@ class ActionModule(object):
         args.update(utils.parse_kv(module_args))
         if not 'msg' in args:
             args['msg'] = 'Failed as requested from task'
-
         result = dict(failed=True, msg=args['msg'])
+        if 'all' in args:
+            result = dict(failed=True, msg=args['msg'], fatal=True)        
         return ReturnData(conn=conn, result=result)
--- a/lib/ansible/runner/action_plugins/fail.py
+++ b/lib/ansible/runner/action_plugins/fail.py
@@ -42,5 +42,5 @@ class ActionModule(object):
             args['msg'] = 'Failed as requested from task'
         result = dict(failed=True, msg=args['msg'])
         if 'all' in args:
-            result = dict(failed=True, msg=args['msg'], fatal=True)        
+            result['fatal'] = True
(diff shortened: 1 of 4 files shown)


mpdehaan:
This appears to insert a situation where returning the string 'all' in a module will fail on all hosts.

This has some possible security usage, in that one host could choose to fail a playbook for all hosts.

Plus I'm not sure it should be called "all" in the return.  Ideally I think we need a way for just the fail module to leverage this.

Can you stop by ansible-devel perhaps and we can discuss?

Thanks!!!

lib/ansible/runner/action_plugins/fail.py
```diff
@@ -39,6 +40,7 @@ def run(self, conn, tmp, module_name, module_args, inject, complex_args=None, **
         args.update(utils.parse_kv(module_args))
         if not 'msg' in args:
             args['msg'] = 'Failed as requested from task'
-
         result = dict(failed=True, msg=args['msg'])
+        if 'all' in args:
+            result = dict(failed=True, msg=args['msg'], fatal=True)
```
anoutsider: i think "result['fatal'] = True"  should be enough/better here?

mpdehaan:
Copied from mailing list:

Hi Amr!

Thanks for the clarification, I think I misread the code that was reading the result and the explicit check for the module name seems quite reasonable to me.

I'd definitely add the "version_added" attribute to the new parameter so the docs can indicate what version this is added in (probably will be 1.8), which is minor.

Much appreciated!

anoutsider:
Thanks,
 i'v added the "version_added" attribute.

mpdehaan:
# Module Repo Information

Hi!

Thanks you very much!

PR #8412, opened 2014-08-03
Nova compute enhancements

emonty:
There's a bunch of updates here, but most of them center around enhancements to telling Nova to actually provide a working IP address at boot, and on ensuring that one has such a thing on subsequent checks. There are a few things around image and flavor selection by name which make config expression much cleaner to read ("Ubuntu Server 14.04.1 LTS" rather than "169d484a-dde2-44c8-8f15-daaa1ba69e94") In general, this stack should get us to feature parity with the OpenStack provisioning that the OpenStack infra team uses in production so that we can transition to using ansible instead of custom scripts (whee!) It's all in one PR because each of the different features in here touch the same bits, so making independent pull requests that don't conflict was kinda impossible - but I'm happy to split it up if that's better.

### DIFF
--- a/lib/ansible/module_utils/openstack.py
+++ b/lib/ansible/module_utils/openstack.py
@@ -54,3 +54,16 @@ def openstack_argument_spec():
     else:
         spec['login_tenant_name'] = dict(required=True)
     return spec
+
+
+def openstack_find_nova_addresses(addresses, ext_tag, key_name=None):
+
+    ret = []
+    for (k, v) in addresses.iteritems():
+        if key_name and k == key_name:
+            ret.extend([addrs['addr'] for addrs in v])
+        else:
+            for interface_spec in v:
+                if 'OS-EXT-IPS:type' in interface_spec and interface_spec['OS-EXT-IPS:type'] == ext_tag:
+                    ret.append(interface_spec['addr'])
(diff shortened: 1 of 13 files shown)


mpdehaan:
Hi Monty,

Can you please split this up into seperate pull requests (i.e. per module where relevant)?

This would make things easier to review on a case by case basis.

Thanks!

emonty:
Sure thing!

emonty:
@mpdehaan Added 5 new pull requests ( #8445 #8446 #8447 #8448 #8449 )  that are _most_ of this split out. There will be some merge conflicts between them, but I can take care of those as they come depending on if and how you want to sequence landing any of these. There are two changes I did not break out because the merge conflicts would be stupendous - I'll make a new patch and rebase those once these badboys are in. #

CaptTofu:
I finally had a minute to try this out on the plane. Wondering what would be the best way for me to sanity check - I reckon each pull request at a time...

mpdehaan:
Ok the deal with this set is that the merge conflicts did get pretty fun, so here's what I'm going to do.

The seperate patches were GREAT for review, and I'm ok with all of them.  I'm going to merge this _together_, this one, and then update the version_added strings for each.

PR #8416, opened 2014-08-04
Improved handling of already terminated instances

willthames:
It is possible to create an instance, terminate the instance and then
attempt to recreate the instance with the same parameters. In this case
`ec2.run_instances` returns a reservation list containing the instance ids
but the logic gets stuck waiting for the instance to exist in the call to
`ec2.get_all_instances`, even if wait is no).


mpdehaan:
Merged, thanks!

PR #8426, opened 2014-08-04
Include unicode and var_precedence tests in 'all' target

jlaska:
Previously, the `unicode` and `test_var_precedence` targets were not run by the `all` `Makefile` target.  I'm not sure if this is intentional, but I've updated the `Makefile` to include these targets.


jlaska:
Merged with review from @jimi-c  Thanks!

PR #8432, opened 2014-08-04
runner + callbacks: allow callbacks to template the delegates

srgvg:
Testing the latest 1.7 devel, I noticed the new feature where delegates are mentioned via the playbook callbacks. Given I often use templates/variables in `delegate_to:` I noticed that wasn't being templated in the callback, which has no access to host variables.

I converted the creation of the runner inject to a method, so callbacks can retrieve it and do the necessary templating, as discussed with @jimi-c.

### DIFF
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -611,6 +611,11 @@ class Runner(object):
         inject['environment'] = self.environment
         inject['playbook_dir'] = self.basedir
 
+        # template this one is available, callbacks use this
+        delegate_to = self.module_vars.get('delegate_to')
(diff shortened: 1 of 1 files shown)


mpdehaan:
Not sure I'm comfortable with all the calls to "get_inject" or understand all that's going on here.

Can we discuss this on ansible-devel ?

Thanks!

srgvg:
I mailed the list. If so possible, please consider this as a bug to current 1.7 devel, as a recent patch exposes not templated strings to the end-user.

srgvg:
new force updated commit with alternative fix

jimi-c:
Merged, thanks!

PR #8435, opened 2014-08-04
Send file

stewrutledge:
If the guest machine has vmware tools installed ansible can now send files to the guest directly via vsphere.

```
- vsphere_guest:
    vcenter_hostname: vcenter.mydomain.local
    username: myuser
    password: mypass
    guest: newvm001
    send: yes
    src_file: /path/to/local/file
    dst_file: /path/on/guest
    overwrite: no
    guest_user: guestadmin
    guest_passwd: supersecret
```

There are a few commits here that don't actually apply to the pull request (and aren't reflected in the file, it's from a rebased version from the current devel branch, they are in another pull request I've made).

If this is still too much in here then I can just refork the repo and start over. I'm not really used to doing pull requests!

### DIFF
--- a/library/cloud/vsphere_guest
+++ b/library/cloud/vsphere_guest
@@ -130,6 +130,7 @@ EXAMPLES = '''
       vcpu.hotadd: yes
       mem.hotadd:  yes
       notes: This is a test VM
+      folder: AutoDeploys
     vm_disk:
       disk1:
         size_gb: 10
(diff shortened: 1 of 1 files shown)


mpdehaan:
Hi @stewrutledge 

Yeah, can you please submit it without those stray commits and the merge commit?

Creating a different branch per topic is highly recommended, and also using "git pull --rebase" vs "git merge" and "git rebase" vs "git pull" will help with problems with pulling in merge commits.

Apologies in that we must require this for submission, but with the volume of things going in it would be very ineffective to have to manually cherry-pick and filter out each pull request.  

You should be able to just create another branch yourself and cherry-pick these without having to create a new repo, which should make it somewhat easy.

Thanks for the submission!

mpdehaan:
An example of how this works or how it would be tested would also be welcome!

PR #8441, opened 2014-08-05
Unarchive should work when parent directory is not writable

willthames:
Correct unarchive so that the checks for writeability are
sensible.

Added a test for when parent directory is not writable

### DIFF
--- a/library/files/unarchive
+++ b/library/files/unarchive
@@ -115,8 +115,6 @@ class TgzFile(object):
         self.zipflag = 'z'
 
     def is_unarchived(self):
-        dirof = os.path.dirname(self.dest)
-        destbase = os.path.basename(self.dest)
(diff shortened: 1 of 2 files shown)


mpdehaan:
Thanks!  Flagging this for P2 so hopefully can get in 1.7 if time.   Much appreciated.

mpdehaan:
Note this appears to be just a test and not the fix, per discussion on ansible-project.  Lowering priority for now.

willthames:
No, it has both. Two files changed - one is the test, one is the unarchive module itself.

willthames:
@Tenzer good point :) I was more worried about ensuring that it was a directory, I don't know why I kept the existence check.

willthames:
@Tenzer updated the fix to incorporate your observation. (of course now the line has changed your line level comment has gone!)

Tenzer:
@willthames It's more important the code is correct than keeping my comments ;)

mpdehaan:
Merged, thanks!

PR #8442, opened 2014-08-05
Removed max 8192 bytes, fixes #8424

### LINKED ISSUE #8424: rpm_key doesn't import key with size > 8192 bytes
This doesn't work in ansible 1.6.10:

```
- name: install mariadb public key
  action: rpm_key state=present key=https://yum.mariadb.org/RPM-GPG-KEY-MariaDB
```

This key has a size of 8353 bytes. In module rpm_key is limit of 8192 bytes for download.

Jmainguy:
Fixes #8424. Tested on Centos 6.5 with rpm_key state=present key=https://yum.mariadb.org/RPM-GPG-KEY-MariaDB

### DIFF
--- a/library/packaging/rpm_key
+++ b/library/packaging/rpm_key
@@ -65,10 +65,6 @@ import os.path
 import re
 import tempfile
 
-# Attempt to download at most 8192 bytes.
-# Should be more than enough for all keys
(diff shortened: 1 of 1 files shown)


mpdehaan:
Merged, thanks!

PR #8445, opened 2014-08-05
Add support for two common nova parameters

emonty:
Availability zones and config drive are present on OpenStack public clouds in the wild. Support them.

### DIFF
--- a/library/cloud/nova_compute
+++ b/library/cloud/nova_compute
@@ -92,6 +92,11 @@ options:
         - A list of network id's to which the VM's interface should be attached
      required: false
      default: None
+   availability_zone:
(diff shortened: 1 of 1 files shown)


mpdehaan:
Thanks very much for splitting this up and we'll dive on merging these as soon as 1.8 starts (this week).

Can you add the version_added attribute (should be a value of 1.8) to the DOCUMENTATION string in each of the new parameters you have added?   This will allow people reading the code generated docs to understand they can't use these arguments on older versions.

Thanks very much!

mpdehaan:

mpdehaan:

PR #8446, opened 2014-08-05
Pass region_name to nova_keypair

emonty:
Again, on things with regions, when we're configuring it, we should
actually pass it in to the compute constructor.


mpdehaan:

mpdehaan:
Merged, thanks!

PR #8447, opened 2014-08-05
Support specifying images and flavors by name

emonty:
Config files fully of uuids are hard to read and reason about - and sometimes cloud providers to things with uuids that you would not expect (true story) Give people a choice as to whether they want to specify uuids or general descriptions of images and flavors. (name in the case of image, amount of ram in the case of flavors) Additionally, provide filtering functions to deal with common situations that arise in the current OpenStack public clouds.

### DIFF
--- a/library/cloud/nova_compute
+++ b/library/cloud/nova_compute
@@ -69,12 +69,22 @@ options:
      default: None
    image_id:
      description:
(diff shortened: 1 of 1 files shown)


mpdehaan:

mpdehaan:

PR #8448, opened 2014-08-05
Add support for floating ips in nova

emonty:
Handling floating ips correctly across various openstack clouds is non-trivial, but it is an understandable problem space. Generally speaking, a user wants either a specific known floating ip, wants a public ip but does not care what ip - or wants one from a pool. This handles all three of those cases, as well as adding code to clean up during deletion, and to verify state on inspection so that if a server was created but did not get an IP, a subsequent run will fix the situation. The last case is also useful in the case that the config for a server is changed - such as promoting a replacement server to be the real server by moving the floating ip.

### DIFF
--- a/library/cloud/nova_compute
+++ b/library/cloud/nova_compute
@@ -193,21 +193,116 @@ def _delete_server(module, nova):
     module.fail_json(msg = "Timed out waiting for server to get deleted, please check manually")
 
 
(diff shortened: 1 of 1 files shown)


mpdehaan:
I will make a list of parameters to bump the "version_added" of, so looking at merging this patch series now.

Thanks!

mpdehaan:

PR #8449, opened 2014-08-05
Handle auth arguments across openstack modules consistently

emonty:
Introduce an argument_spec function that can be used in all of the openstack related modules for consistency. Also, add support in that function for setting defaults in the case that the standard OpenStack environment variables are set, to make ad-hoc ansible command line operation do what an openstack admin would expect.


mpdehaan:
Merged, thanks!

PR #8453, opened 2014-08-05
Update ec2_lc to match ec2 module options

colindunn:
Adding instance_profile_name, ebs_optimized and assign_public_ip to match the current ec2 module.
Essentially a copy of the code from that module.


colindunn:
Has been implemented by someone else. Closing.

PR #8459, opened 2014-08-05
Adds support for public_ip boolean for launch configurations.

jsmartin:


mpdehaan:
I'm thinking this may be associate_public_ip in other modules and we may wish to be consistent.   If this is inconsistent in other modules, we should probably provide the most common one and alias.

jsmartin:
Reworded it to "assign_public_ip", same wording that the ec2 module uses.

mpdehaan:
Merged, thanks!

PR #8462, opened 2014-08-06
Improve playbook role name error reporting

mdaniel:
Previously, the documentation had a `{ name: apache, http_port: 8080 }` example in it, but if one were to implement a `roles` dictionary in that manner, it would error out with the message:

> ERROR: expected a role name in dictionary: {'name': 'java'}

Aside from fixing the documentation, this PR updates the message so that it now dies like this:

> ERROR: expected a role name in key 'role' of dictionary: {'name': 'java'}

I feel this is clearer.


amenonsen:
So I tested this with a playbook like:

```
---
- hosts: localhost
  roles:
    - { xyzzy: bar }
```

In devel, this dies with "ERROR! role definitions must contain a role name" (which is an improvement, but I agree with @mdaniel that the role:/name: keys (both now accepted) should be mentioned here), but that's followed by the whole "It seems that there is a value started with a quote, and the YAML parser is expecting to see the line ended with the same kind of quote" message pointing at the opening brace, which is completely and utterly misleading.

It seems that the AnsibleError that's raised is being handled by some code that doesn't expect the possibility that the error was already diagnosed and reported, but I haven't dug into this in more detail.

mdaniel:
I'm going to close the PR since it no longer applies (in both senses of the verb) and the docs have been fixed. Too bad, we almost made it to the one year anniversary of this being open. I would have brought :cake:

PR #8463, opened 2014-08-06
Show the task dict with erroneous key

mdaniel:
This is very similar to #8462, so if I should merge them into one please let me know.

When a task has an erroneous key, it dies with the message:

> ERROR: path is not a legal parameter in an Ansible task or handler

After this patch, it dies with the more verbose but also more illustrative:

> ERROR: 'path' is not a legal parameter in task or handler: {'path': 'foo.sh', 'stat': None, 'register': 'my_var', 'name': 'foo bar', 'ignore_errors': True}

Observe that the offending key is now quoted, to set it apart from the rest of the text

### DIFF
--- a/lib/ansible/playbook/task.py
+++ b/lib/ansible/playbook/task.py
@@ -20,7 +20,7 @@ from ansible import utils
 from ansible.module_utils.splitter import split_args
 import os
 import ansible.utils.template as template
-import sys
+
 
 class Task(object):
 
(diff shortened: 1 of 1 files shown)


mpdehaan:
Thanks for this.

Previously I had opened up the idea of reporting the line number and file instead.  I'm probably interested in exploring this concept instead as the data in the role may be quite large and possibly repeated between roles but it was somewhat difficult to "get at" at the time.

mdaniel:
Makes sense; there are so many places in Ansible where the error reporting is sub-optimal, I thought it just required someone taking the time to address them.

In the interim until the file+line is finalized, do you think just showing the keys in the dict would be an acceptable compromise?

amenonsen:
I guess this PR should be closed now, the underlying code has again changed and the "is not a legal parameter" error no longer exists in devel.

gregdek:
Yes, this PR is now very outdated. Thanks @mdaniel for the effort, and sorry it took so long to make a decision on it.

PR #8465, opened 2014-08-06
Changed the issue template link to point to the web view instead of the raw output

andreausu:
Before: https://raw2.github.com/ansible/ansible/devel/examples/issues/ISSUE_TEMPLATE.md
After: https://github.com/ansible/ansible/blob/devel/examples/issues/ISSUE_TEMPLATE.md


mpdehaan:
Thanks for this, though I somewhat disagree with this change.  The raw output was intentionally chosen because it could be copy and pasted as markdown and dropped in the template.  Copying and pasting the web view results in incorrectly formatted data.

So yes, it's easier to read, but results in a few copy/paste problems.

I hope you understand... a bit hard to get it both ways!  :)

Thanks!

PR #8469, opened 2014-08-06
packaging: add short hash and branch name in package version for unoffici...

srgvg:
...al builds

/cc @jlaska with whom I discussed this.

Next step would be to show the same git info one gets when running ansible --version from git (hacking script), when running it from a package - at least for unofficial builds.

[Review on Reviewable](https://reviewable.io/reviews/ansible/ansible/8469)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,11 @@ VERSION := $(shell cat VERSION)
 # Get the branch information from git
 ifneq ($(shell which git),)
 GIT_DATE := $(shell git log -n 1 --format="%ai")
+GIT_HASH := $(shell git log -n 1 --format="%h")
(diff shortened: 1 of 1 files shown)


mpdehaan:
Thanks for this!

cc @jlaska for review and whether this makes our PPA/other things happy.

jlaska:
Including the hash in the DEB and RPM release string makes sense for unofficial builds.  I'm not 100% sure we need the branch name included, but it certainly doesn't hurt.

amenonsen:
What's the verdict on this one? It seems like it's a no-brainer and should be merged.

srgvg:
@jlaska the idea for the branch name is usefull for testing feature branches separately, or maintaining private patches, etc..

PR #8470, opened 2014-08-06
Added option tot install only security updates eq "yum --security update...

DJMuggs:
Hi i added the option to only install security updates with yum package manager.


mpdehaan:
This looks good, can you please update the version_added to say "1.8" which is likely the release in which this will be merged in.

Much appreciated, thank you!

DJMuggs:
Moved this update to my yum branch and will create a new pull request from that branch.

Issue #6399, opened 2014-03-11
npm module won't install a package if it is installed as another package's dependency


davidminor:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.5
##### Environment:

N/A
##### Summary:

npm module won't install a package if it is installed as another package's dependency
##### Steps To Reproduce:

``` yml
- name: Install less compiler, css minifier
  npm: name={{ item }}
  with_items:
    - less
    - clean-css
```
##### Expected Results:

clean-css should be installed, as it will be if you run:

``` shell
# npm install less clean-css
```
##### Actual Results:

clean-css is not installed.

This is because the npm module is misinterpreting the results of npm list. If the less package is installed, `npm list --json clean-css` returns:

``` json
{
  "dependencies": {
    "less": {
      "version": "1.7.0",
      "from": "less@",
      "dependencies": {
        "clean-css": {
          "version": "2.1.6",
          "from": "clean-css@2.1.x"
        }
      }
    }
  }
}
```

The npm module interprets this to mean that clean-css is installed.

bartdag:
Looks like a duplicate of #4659

jimi-c:
Hi!

The above change has been applied for this ticket, which should resolve this item for you.

If you believe this ticket is not resolved, or have further questions, please let us know by stopping by one of the two mailing lists, as appropriate:

https://groups.google.com/forum/#!forum/ansible-project - for user questions, tips, and tricks
https://groups.google.com/forum/#!forum/ansible-devel - for strategy, future planning, and questions about writing code

Because this project is very active, we're unlikely to see comments on closed tickets, though the mailing list is a great way to get involved or discuss this one further.

Thanks!

PR #8475, opened 2014-08-06
Ensure npm packages actually are installed in a reachable way.

### LINKED ISSUE #4659: Allow 'npm' to install top-level modules
**Scenario**:

Trying to install `uglify-js` (globally) because I need the _uglifyjs_ executable. Unfortunately the `npm` Ansible module won't install it because it's already listed as a dependency of another nodejs package (this other package is installed earlier in my playbook and I can't really move it). But when installed as dependency, it didn't get installed with the executable.

**Possible solutions**:
- A 'force' option to install regardless. But that's against idempotency I suppose.
- Make the `npm` Ansible module aware if a package is installed at top-level or not and allow installation is the package to be installed is strictly a dependency.

christophebiocca:
Fixes #4659 and #6399.
Alternative to #6613

I have a task as follows:

```
- name: install npm dependencies
  npm: name={{item}} state=latest
  sudo: yes
  with_items:
    - stitch
    - underscore
```

Then I can:

```
> require('stitch')
{ compilers: 
   { js: [Function],
     coffee: [Function],
     eco: [Function] },
  Package: [Function: Package],
  createPackage: [Function] }
```

but 

```
> require('underscore')
Error: Cannot find module 'underscore'
```

This is because underscore was installed as a dependency of stitch, and the current behaviour of the `npm` command is to consider a package installed even if it's hidden away as a dependency of some other module (and hence unusable).

Most workarounds involve installing packages in a very specific order (but that's incredibly brittle, and won't work with circular dependencies), or using npm install manually using `command` (which is slow).

This patch ensures that `state=present` or `state=latest` actually mean that the module can be `require`'d without an error.
#6613 does a similar thing, but requires a flag to do so and I'd argue that `toplevel=no` is never what the end user wants (it pretty much is equivalent to [RANDOMLY-LOSE](http://tools.ietf.org/html/rfc748)).

This pull won't break pre-existing code, unless they relied on non-deterministic errors in order to work.

### DIFF
--- a/library/packaging/npm
+++ b/library/packaging/npm
@@ -163,6 +163,8 @@ class Npm(object):
                     missing.append(dep)
                 else:
                     installed.append(dep)
+            if self.name and self.name not in installed:
+                missing.append(self.name)
         #Named dependency not installed
         else:
             missing.append(self.name)


jimi-c:
Merged, thanks!

PR #8478, opened 2014-08-06
EC2 ELB AWS Rate Limit Issue

Lazyshot:
When dealing with large enough number of ELBs registration will throw a rate limit exceeded failure.
Boto doesn't seem to retry, which leads to the failure of the entire task.

My particular situation involved 7 ec2 instances and 34 ELBs.

To combat this I stopped registration calls for instances already "InService" and added a rate limit sleep of 1 second.

### DIFF
--- a/library/cloud/ec2_elb
+++ b/library/cloud/ec2_elb
@@ -152,7 +152,7 @@ class ElbManager:
 
             initial_state = self._get_instance_health(lb)
 
-            if initial_state.state == 'InService':
+            if initial_state is not None and initial_state.state == 'InService':
                 continue
 
             if enable_availability_zone:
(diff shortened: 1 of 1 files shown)


mpdehaan:
Thoughts on this on @jsmartin or @jimi-c ?

jimi-c:
I would move the sleep to later in the loop, as there's no point in sleeping if the host is going to be skipped. Beyond that, this seems like it would be helpful if you had a large number of ELBs, thought it also depends on whether the overall API limits are set per product or per region (or something else). If it's per region, this would not be very helpful as the number of API calls for other environment objects would be much higher than those for the ELBs.

Lazyshot:
I would initially agree with you on the placement of the sleep, but gathering instance health requires an API hit. 

So if a host is already registered to a largish quantity of ELBs, then it will skip every registration, but still hit the API each time for getting instance health on that ELB.

At least that was my rationale for placing it where it is. Otherwise I would agree with you.

mpdehaan:
# Module Repo Information

Hi!

Thanks you very much!

PR #8479, opened 2014-08-06
Adding route53_included_zones parameter to limit zones queried to build inventory

jerm:
inventory script.

Rationale: Having many zones can severly slow down the inventory script and it's
conceivable if not likely that one may only want/need a few zones in their inventory.
This is a simple enough addition that it was a no-brainer and is completely opt-in.

Also, now stripping spaces from route53_included_zones and route53_excluded_zones
because spaces around commas would break the list and the example had spaces.
Figured it made more sense to make the parameter more flexible than just fix the
example.

### DIFF
--- a/plugins/inventory/ec2.ini
+++ b/plugins/inventory/ec2.ini
@@ -45,6 +45,11 @@ route53 = False
 # 'route53_excluded_zones' as a comma-separated list.
 # route53_excluded_zones = samplezone1.com, samplezone2.com
 
+# If you have lots of zones the inventory script can take a long time to run
+# If you only want/need a few of the zones to build your inventory you can 
+# limit the included zones here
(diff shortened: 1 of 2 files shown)


amenonsen:
@jerm, do you feel like rebasing to apply to devel? Your change seems reasonable.

amenonsen:
@jerm I'd like to merge this if you want to rebase to apply to v2 (which might not involve much more than targeting ec2.py in contrib/inventory instead of plugins/inventory).

amenonsen:
I'm closing this PR now, but please feel free to open a new one if you get a chance to update your patch.

Since this is a really busy project, we don't always see comments on closed tickets, but if you want to reopen discussion on this subject, please feel free to do so on the development mailing list:

```
```

Issue #5748, opened 2014-01-24
hostname module detect FQDN and update /etc/hosts


twouters:
The hostname module currently sets the hostname wether it's an FQDN or not.

```
# hostname
testhostname.example.tld
# hostname -f
hostname: Unknown host
```

Ideally it should set the hostname to the first part of the FQDN (until the first dot).
`/etc/hosts` should be updated to include a rule like this:

```
127.0.0.1    {{ hostname_fqdn }} {{ hostname }} localhost localhost.localdomain ....
```

Or if the provided hostname is not an FQDN:

```
127.0.0.1    {{ hostname }} localhost localhost.localdomain ...
```

Another option would be to insert a new line above all other entries using the primary IP address of the host, leaving the existing localhost lines as they are.
But this could be less trivial to do, since detecting what the primary IP is might be difficult.

```
192.168.0.2    {{ hostname_fqdn }} {{ hostname }}
```

I'm not entirely sure this is exactly how it should be done, it's just how I usually do it.
I haven't been able to think of cornercases which would make this an invalid procedure.

kiranos:
This is how fqdn should be set if for example hostname -f should output:  server.example.com 

/etc/hosts:
192.168.5.9     server.example.com server

/etc/hostname
server

and /etc/hosts should also have 127.0.0.1 localhost etc. but the part about fqdn is as above.

```
#/etc/hosts:
127.0.0.1       localhost
192.168.5.9     server.example.com server

# The following lines are desirable for IPv6 capable hosts
::1     ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
```

```
#/etc/hostname
server
```

Imo the hostname module should be able to set the FQDN and not just the partial hostname, maybe I should write a separate feature request for that?

filipenf:
@kiranos What if we have 2 NICs? Just pick the first one?

kiranos:
you should pick the one that host the ip which the domain name is pointing to

filipenf:
It wouldn't work. The domain name will point to the public IP and in most of the cases we just have the private IP.

I think the sugestion from @twouters is easier to implement and works well.

filipenf:
Actually I just sent a PR (#8482) to add an alias to /etc/hosts.

mpdehaan:
Hi!

[...]

PR #8482, opened 2014-08-07
Set /etc/hosts alias to match hostname

filipenf:
Issue: #5748

On debian based systems we need to change the /etc/hosts file to reflect the host name alias

### DIFF
--- a/library/system/hostname
+++ b/library/system/hostname
@@ -44,6 +44,8 @@ from distutils.version import LooseVersion
 # import module snippets
 from ansible.module_utils.basic import *
 
+import tempfile
+import shutil
 
 # wrap get_distribution_version in case it returns a string
 def _get_distribution_version():
@@ -182,12 +184,28 @@ class DebianStrategy(GenericStrategy):
                 str(err))
 
     def set_permanent_hostname(self, name):
-        try:
-            f = open(self.HOSTNAME_FILE, 'w+')
-            try:
+        def set_hosts_alias():
+            temp = tempfile.NamedTemporaryFile()
+            hosts = open('/etc/hosts', 'r')
+            for line in hosts:
+                if line.startswith('127.0.0.1'):
+                    temp.write("127.0.0.1 %s localhost "
+                        "localhost.localdomain" % name)
+                else:
+                    temp.write(line)
+
+            hosts.close()
+            temp.flush()
+            shutil.copyfile(temp.name, "/etc/hosts")
+            temp.close()
+
+        def set_hostname():
+            with open(self.HOSTNAME_FILE, 'w+') as f:
                 f.write("%s\n" % name)
-            finally:
-                f.close()
+
+        try:
+            set_hostname()
+            set_hosts_alias()
         except Exception, err:
             self.module.fail_json(msg="failed to update hostname: %s" %
                 str(err))
--- a/library/system/hostname
+++ b/library/system/hostname
@@ -185,23 +185,26 @@ class DebianStrategy(GenericStrategy):
 
     def set_permanent_hostname(self, name):
         def set_hosts_alias():
-            temp = tempfile.NamedTemporaryFile()
-            hosts = open('/etc/hosts', 'r')
-            for line in hosts:
-                if line.startswith('127.0.0.1'):
-                    temp.write("127.0.0.1 %s localhost "
-                        "localhost.localdomain" % name)
-                else:
-                    temp.write(line)
-
-            hosts.close()
-            temp.flush()
-            shutil.copyfile(temp.name, "/etc/hosts")
-            temp.close()
+            try:
+                temp = tempfile.NamedTemporaryFile(delete=False)
+                hosts = open('/etc/hosts', 'r')
+                for line in hosts:
+                    if line.startswith('127.0.0.1'):
+                        temp.write("127.0.0.1 %s localhost "
+                            "localhost.localdomain" % name)
+                    else:
+                        temp.write(line)
+            finally:
+                hosts.close()
+                temp.close()
+                shutil.move(temp.name, "/etc/hosts")
 
         def set_hostname():
-            with open(self.HOSTNAME_FILE, 'w+') as f:
+            try:
+                f = open(self.HOSTNAME_FILE, 'w+')
                 f.write("%s\n" % name)
+            finally:
+                f.close()
 
         try:
             set_hostname()


filipenf:
If hostname changes are not reflected to /etc/hosts one may start getting errors like this:  http://stackoverflow.com/questions/21003910/linux-and-unknownhostexception-for-a-java-program

mpdehaan:
# Revision Request

Hi!

- (A) the code using the with statement in python needs to be changed, as it's not 2.4 complaint.  We still have to support python 2.4 on some older legacy platforms for modules, even though Ansible's main program does take 2.6+ and the "with" is fine there.
- (B) Should it be that the NamedTemporaryFile is unlinked after the copyfile, or that the copyfile is replaced with a rename to not leave the named temporary file around?

which is very active:
- https://groups.google.com/forum/#!forum/ansible-devel - for user questions, tips, and tricks
- https://groups.google.com/forum/#!forum/ansible-project - for strategy, future planning, and questions about writing code

filipenf:
I rolled back the use of with clause.

I don't believe we would have any problem with NamedTemporaryFile because by default it automatically deletes the file. Anyway I changed the intent becomes explicit.

If there's something else please let me know.

mmoya:
Hi @filipenf. I'm not sure if you considered idempotence here. AFAICS _/etc/hosts_ gets rewritten everytime the module is run.

My solution so far have been this (note it's 127.0.**1**.1):

```
---
- name: set hostname facts
  set_fact:
    full_hostname="{{ inventory_hostname}}"
    short_hostname="{{ inventory_hostname.split('.', 1)[0] }}"

# " lovely lonely double-quote for fixing vim highlighting

- name: set hostname
  hostname:
    name={{ short_hostname }}

- name: add hostname to /etc/hosts
  lineinfile:
    dest=/etc/hosts
    line='127.0.1.1 {{ full_hostname }} {{ short_hostname }}'
    regexp='^127\.0\.1\.1'
    insertafter='^127\.0\.0\.1'
    state=present
```

and it works as expected, ie:

```
$ hostname
foo
$ hostname -f
foo.bar.guz
```

filipenf:
@mmoya Your suggestion is somewhat equal to what I'm using currently. The thing is that this shouldn't be necessary because it is only required for debian systems. I believe modules should behave the same way in different systems.

About being idempotent: The file is rewritten every time modifying only the 127.0.0.1 line. It is possible to check if the line is equal the value we want before updating it, do you want me to change?

dooblem:
This is what I used to change the fqdn AND hostnames of my hosts:

```
cat /etc/ansible/hosts
1.2.3.4 fqdn=test.example.com
```

The playbook:

```
- hosts: 1.2.3.4
  vars:
    ansible_fqdn_new: "{{ fqdn | default(ansible_fqdn) }}"
    ansible_hostname_new: "{{ ansible_fqdn_new.split('.', 1)[0] }}"
  tasks:
  - fail: msg="Bad FQDN"
    when: ansible_fqdn == ansible_hostname
  - replace: dest=/etc/hosts regexp='(\s+){{ ansible_fqdn }}(\s+.*)?$' replace='\1{{ ansible_fqdn_new }}\2' backup=yes
  - replace: dest=/etc/hosts regexp='(\s+){{ ansible_hostname }}(\s+.*)?$' replace='\1{{ ansible_hostname_new }}\2'
  - hostname: name={{ ansible_hostname_new }}

```

mmoya:
@filipenf In this cases I used to generate a tempfile, check md5/whatever sums and only update the target files if sums differs. Seems like a big deal for just a tiny change but writing to a temp file is more reliable in face of write errors (you don't break the target file). Besides, access and modification times are preserved. My 2¢.

mpdehaan:
# Module Repo Information

Hi!

Thanks you very much!

Jmainguy:
Can you reopen this PR in the new repo? Looks like there is still a bit of interest into this. One idea raised on irc today was to make a edit_hosts= option to the module so it only edits /etc/hosts of edit_hosts=true

akatch:
+1

kiranos:
+1

craigmarvelley:
+1

dk-dkopen:
-1 as it is - see below + although I myself use ubuntu and have to do this, i'm fine doing it myself with a short lineinfile (unlike /etc/hostname where it's clear and simple) rather than something rushed for such an important file.

Something like @mmoya suggested - note that Ubuntu by default structures it as:
127.0.0.1 localhost
127.0.1.1   myhostname

I would say this feature should:
- match the way it's set up by a fresh default Ubuntu (or otherwise) system deployment
- be careful not to change anything else in /etc/hosts or cause any unexpected side effects
- should be optional and disabled by default
- should only write file when actually changed
- ensure that ansible_fqdn ends up picking up the hostname thus specified

Issues with current pull request:
- replaces ALL lines that .startswith('127.0.0.1') which could corrupt it if you have further aliases to 127.0.0.1 down the file
- rewrites regardless of whether or not file changed

But, given all the trouble of specifying, implementing, testing and vetting, I'm not sure I'd consider this worthwhile as such.

Maybe a separate "hosts" module to fully express /etc/hosts would be better?
(as an alternative / in general)
More in line with keeping it simple, specific and modular.

akatch:
My interest in making this change to the hostname module is simply that,
in my opinion, a module designed to change the hostname should make ALL
changes needed to update the hostname.  Updating the hostname should not
require a separate lineinfile task (as it does now, for certain types of
systems) to ensure all applications are able to resolve the new
hostname.  For instance, without editing the hosts file, Tomcat will
throw a fit and refuse to start because it can't resolve the local
hostname.

I would agree that changing all instances of 127.0.0.1 might not be a
great idea.  The module could instead change all instances of 127.0.1.1
or use a different address in the 127/8 subnet.  No matter how it is
done, though, fully updating the hostname shouldn't require an
additional task.  Just my $.02.

romanr:
What is the status of this change/pull request? From conversation I understand it went to another old repo or something? This pull request appears to be merged to ansible "develop" branch and tags 1.7.2.
Yet when using 1.7.2 my /etc/hosts is not updated when setting hostname.
So it's not in 1.7.2, but ticket is Closed?

Jmainguy:
@romanr I dont see a reference to a new PR, so as far as I can tell this is still closed. If someone else wants to open a new PR against the new repo's, it would gather some attention for sure.

er1c:
+1

PR #8495, opened 2014-08-07
adding fix where sometimes state=absent tries to delete wrong asg

mschurenko:
This was a pretty bad bug where sometimes, after all of the instances of the asg where deleted ec2_asg would try to delete the wrong asg. The good thing is that since the wrong group has running instances it would just fail. This was hard to repro as it didn't happen all the time. This is a small fix which solves the problem.

### DIFF
--- a/library/cloud/ec2_asg
+++ b/library/cloud/ec2_asg
@@ -291,10 +291,11 @@ def delete_autoscaling_group(connection, module):
             for group in groups:
                 if group.name == group_name:
                     if not group.instances:
+                        target_group = group
                         instances = False
             time.sleep(10)
 
-        group.delete()
+        target_group.delete()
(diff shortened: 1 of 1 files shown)


jimi-c:
It appears commit 7fd4c2f5e1f60ff63ce6cc89fa499552bddd111d conflicts with this, and may resolve this issue. Can you confirm if you are still seeing this issue with the current devel branch (or specifically with the above commit applied)?

mschurenko:
I've tested it out and it seems like it's been fixed.

jimi-c:
@mschurenko great, in that case I'm going to go ahead and close this PR. If you find that the issue is still occuring for you, just rebase your patch and send it to us in a new PR. Thanks!

PR #7962, opened 2014-06-27
ELB enhancements

ferrants:
- connection draining
- cross AZ load balancing

PR #8496, opened 2014-08-07
adding option to enable connection draining and crosszone load balancing

mschurenko:
I needed these options for running in our VPC. I added cross zone load balancing and connection draining. I didn't add accesslog as I don't need that ATM. More information can be found here:

http://docs.aws.amazon.com/ElasticLoadBalancing/latest/APIReference/API_ModifyLoadBalancerAttributes.html

### DIFF
--- a/library/cloud/ec2_elb_lb
+++ b/library/cloud/ec2_elb_lb
@@ -93,6 +93,12 @@ options:
     choices: ["yes", "no"]
     aliases: []
     version_added: "1.5"
+  attributes:
+    description:
+      - An associative array of attribute settings. Only crosszone_load_balancing and connection_draining are implemented right now. See EXAMPLES.
+    required: false
+    default: None
+    version_added: "1.7"
 
 extends_documentation_fragment: aws
 """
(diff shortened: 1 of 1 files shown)


jsmartin:
This implementation seems like it is a bit more complicated than https://github.com/ansible/ansible/pull/7962, with the tradeoff of being less explicit.

mschurenko:
I don't know. I find mine more explicit in some ways, especially the way it has options for both enabling connection draining and for setting the time out. These are two separate parameters; however #7962 assumes enabled when a timeout is set. What if you want to disable connection draining? It doesn't seem like there is a way to do that with this implementation. In mine you can set enabled to no and then it will disable it. Looking at my code it looks like it would still require a timeout to be set; however it should be ignored (I'll have to double check that), or I can change that. Also I have cross zone and connection draining under "attributes" which is more in line with the ELB documentation.

jimi-c:
# Possible Duplicated Efforts

Hi!

It looks like the work from this pull request was duplicated by another PR, which we feel is more appropriate for the following reasons:
- Where appropriate, we try and add parameters for specific features, rather than requiring them to be placed under a complex data structure.
- Regarding your concern about the timeout being required in #7962, upon review of the code it is clear to us that not specifying the option means the feature will be disabled.

However, we're absolutely always up for discussion.  Since this is a really busy project, we don't always see comments on closed tickets, but want to encourage open dialog.  You can stop by the development list, and we'd be glad to talk about it - and we might even be persuaded otherwise! If you feel that your PR offers additional features, we would be happy to review any future enhancements if/when this PR has been merged in.

PR #8500, opened 2014-08-07
rax: Move additional shared code into module utils

sivel:
This pull request moves additional duplicated code for the rax\* modules into `ansible.module_utils.rax`

I've tested this quite a bit, but would be happy for it to make it into 1.8 early in the dev cycle to potentially catch  any bugs that I have not been able to find.

One thing that this PR changes outside of this is to add an ANSIBLE_VERSION constant into basic.py that is replaced using the module replacer code.  This was added so that we could manipulate the user agent of pyrax to include ansible, so that we could make the distinction between pure pyrax requests and those made with ansible.


mpdehaan:
Just started 1.8 so a great time to merge this, thanks!

PR #8515, opened 2014-08-08
Small doc addition to ec2_group to show the various values for "state"

mpeters:
It wasn't clear from the docs what the values of "state" could be, so this small fix should clear that up.


mpdehaan:
Merged, thanks!

PR #8516, opened 2014-08-08
adding include_pattern and exclude_pattern options to ec2.py

mpeters:
Sometimes it's useful to be able to restrict the hosts returned from ec2.py using a pattern. For instance, if you need to make sure you only run against hosts in a specific project, or you are outside of a vpc and and some servers are inside of it, so there's no reason to even include them since you can't ever reach them.

This adds 2 optional parameters to ec2.ini (include_pattern and exclude_pattern) that are just regular expressions that are matched against whatever key you're using for the destination (destination_variable or vpc_destination_variable)


mpdehaan:
Merged, thank you!

PR #8517, opened 2014-08-08
gce.py: Add hosts to group "all" 

icoloma:
`gce.py` is not populating the group `all` with the instances retrieved from GCE. As a result, vars specified for the "all" group are not being applied to these hosts. This commit fixes that.

### DIFF
--- a/plugins/inventory/gce.py
+++ b/plugins/inventory/gce.py
@@ -259,6 +259,9 @@ class GceInventory(object):
             stat = 'status_%s' % status.lower()
             if groups.has_key(stat): groups[stat].append(name)
             else: groups[stat] = [name]
+            
+            if groups.has_key('all'): groups['all'].append(name)
+            else: groups['all'] = [name]
         return groups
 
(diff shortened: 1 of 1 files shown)


mpdehaan:
# Code Modification Request

Hi!

We have noticed some things in the code that we need to be modified before we can consider merging this one:
- ansible requires that python conditionals not be expressed as one line statements, so please break these up into multiple lines
- the "all" group in ansible should be implicit, we're curious what version of Ansible you are testing against but this should not be something that needs to be fixed in the GCE module, but should be handled by general inventory code.  If you believe there is a problem with this (we're not positive there is yet), we'd like to have it fixed generically.

which is very active:

In particular, ansible-devel is quite appropriate for asking questions related to the code behind this change.

icoloma:
You are right, I was developing against 1.6.7 (from 18 days ago). I have upgraded to 1.7.0 and this behavior seems to be fixed. I will drop the issue.

PR #8526, opened 2014-08-08
Fix issue with relative symlinks

### LINKED ISSUE #7711: File module fails to make relative symlinks
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.7 (devel 0073b34653) last updated 2014/06/10 02:55:50 (GMT -500)
##### Environment:

Running from OSX 10.7
Remote is Debian stable
##### Summary:

The file module errors on relative links.
##### Steps To Reproduce:

Make /tmp/foo/foo
Try to symlink /tmp/bar/bar to ../foo/foo
##### Expected Results:

For the link to succeed.
##### Actual Results:

```
failed: [default] => {"failed": true, "gid": 1000, "group": "vagrant", "mode": "0777", "owner": "vagrant", "path": "/tmp/bar/bar", "size": 10, "src": "../foo/foo", "state": "link", "uid": 1000}
msg: src file does not exist, use "force=yes" if you really want to create the link: /tmp/bar/foo
```

phy1729:
Fixes #7711

### DIFF
--- a/library/files/file
+++ b/library/files/file
@@ -237,7 +237,7 @@ def main():
         else:
             relpath = os.path.dirname(path)
 
-        absrc = os.path.normpath('%s/%s' % (relpath, os.path.basename(src)))
+        absrc = os.path.normpath('%s/%s' % (relpath, src))
         if not os.path.exists(src) and not os.path.exists(absrc) and not force:
             module.fail_json(path=path, src=src, msg='src file does not exist, use "force=yes" if you really want to create the link: %s' % absrc)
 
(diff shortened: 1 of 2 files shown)


jimi-c:
cc @bcoca for review.

phy1729:
@bcoca, what about the above patch?

bcoca:
https://github.com/bcoca/ansible/tree/symlink_tests <== added tests, seems like all other cases were already covered, will pull fix when i get the chance and run all

phy1729:
Issue #7711 was closed by #8845

Issue #8511, opened 2014-08-08
utils._clean_data extremely slow when handling large strings


leth:
##### Issue Type:

Bug Report
##### Ansible Version:

devel branch HEAD
##### Environment:

Ubuntu 14.04 and OSX
##### Summary:

Unarchive task seems to hang.
Looking at a process listing, the tar command finishes, and then the ansible task process stays pegged at 100% CPU
##### Steps To Reproduce:
1. Create large tar archive.
   (Listing of our tar file contents: https://gist.github.com/leth/bc059d52be90b1ef9e14 )
2. unarchive tar using ansible
3. Watch
##### Expected Results:

I'd expect this to complete within reasonable time
##### Actual Results:

This task does not ever seem to complete

mpdehaan:
# Information Request

Hi!

Thanks very much for your submission to Ansible.  It sincerely means a lot to us. 

We have some questions we'd like to know about before we can get this request queued up.   If you can help answer them, we'd greatly appreciate it:
- when you say devel branch head, please include the output of ansible --version
- how long did you wait?
- the code in question moves the file and then is executing the GNU toolchain tools here, so this seems curious to us that it would hang.  See https://github.com/ansible/ansible/blob/devel/library/files/unarchive
  - If you have any python development skills, some help debugging the problem in your environment may be useful, as our integration tests for the unarchive module are not detecting problems at this time.  

Just as a quick reminder of things, this is a really busy project.  We have over 800 contributors and to manage the queue effectively
we assign things a priority between P1 (highest) and P5.  We'd like to thank you very much for your time!  
We'll work things in priority order, so just wanted you to be aware of the queue and know we haven't forgot about you!

We will definitely see your comments on this issue when reading this ticket, but may not be able to reply promptly.  You may also wish to join one of our two mailing lists
which is very active:
- https://groups.google.com/forum/#!forum/ansible-project - for user questions, tips, and tricks
- https://groups.google.com/forum/#!forum/ansible-devel - for strategy, future planning, and questions about writing code

Thank you once again for this and your interest in Ansible!

leth:
> when you say devel branch head, please include the output of ansible --version

`ansible 1.8 (devel d4cedfc7f1) last updated 2014/08/08 14:54:24 (GMT +100)`

> how long did you wait?

Our jenkins server has hung on this step for hours. I've not waited that long locally.

[...]

PR #8536, opened 2014-08-09
Optimise string handling in ansible.utils._clean_data

leth:
As discussed in #8511, [here is the benchmarking info](https://gist.github.com/leth/1f2ba2d6b961fabe5d60).

### DIFF
--- a/lib/ansible/utils/__init__.py
+++ b/lib/ansible/utils/__init__.py
@@ -33,7 +33,7 @@ from ansible.module_utils.splitter import split_args, unquote
 import ansible.constants as C
 import ast
 import time
-import cStringIO
+import StringIO
 import stat
 import termios
 import tty
@@ -366,7 +366,7 @@ def _clean_data(orig_data, from_remote=False, from_inventory=False):
 
     regex = PRINT_CODE_REGEX if replace_prints else CODE_REGEX
 
-    with contextlib.closing(cStringIO.StringIO()) as data:
+    with contextlib.closing(StringIO.StringIO()) as data:
         # these variables keep track of opening block locations, as we only
         # want to replace matched pairs of print/block tags
         last_pos = 0
--- a/lib/ansible/utils/__init__.py
+++ b/lib/ansible/utils/__init__.py
@@ -33,7 +33,7 @@ from ansible.module_utils.splitter import split_args, unquote
 import ansible.constants as C
 import ast
 import time
-import StringIO
+import cStringIO
 import stat
 import termios
 import tty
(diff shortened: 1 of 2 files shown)


jimi-c:
So part of the problem with cStringIO is that it does not handle unicode. If you run the unicode integration test with this patch `(cd test/integration; make unicode)`, you'll see it fail. I believe using io.StringIO() is just as quick, though running your benchmark it doesn't seem to matter much which library you use, the _regex method is quicker leading me to suspect that the underlying buffer is not that important. For example, just replacing cStringIO with StringIO again in your _regex method gives the following benchmark results:

<pre>
length: 2000 density: 0.0
regex   none      0.03549504
orig cs none      1.33576512
orig    none      4.06493807
regex   inventory 0.03513479
orig cs inventory 1.32415199
orig    inventory 4.05717015
regex   both      0.03523111
orig cs both      1.32435417
orig    both      4.04388309
regex   remote    0.03511596
orig cs remote    1.32523990
orig    remote    4.04796195
</pre>

This is slightly slower than cStringIO, but is still much faster than the other methods.

You have at least one typo in the code:

<pre>
-    regex = PRINT_CODE_REGEX if replace_prints else ONLY_CODE_REGEX
+    regex = PRINT_CODE_REGEX if replace_prints else CODE_REGEX
</pre>

which just looks like a cut/paste error from your benchmark code.

leth:
Thanks; sorry about that.

Fortunately swapping back to normal stringio doesn't have much of a performance impact:

```
length: 2000 density: 0.0
regex cs     none      0.02231717
regex non-cs none      0.02372718
regex cs     inventory 0.02257013
regex non-cs inventory 0.02402496
regex cs     both      0.02252388
regex non-cs both      0.02398491
regex cs     remote    0.02258110
regex non-cs remote    0.02402186

length: 20000 density: 0.0
regex cs     none      0.20524287
regex non-cs none      0.20422101
regex cs     inventory 0.20501399
regex non-cs inventory 0.20466089
regex cs     both      0.20503783
regex non-cs both      0.20483708
regex cs     remote    0.21779609
regex non-cs remote    0.21423101
```

I'll update the PR appropriately

mpdehaan:
In your previous comment: 

"Our jenkins server has hung on this step for hours. I've not waited that long locally."

I'm wondering if Jenkins was just waiting for a password and needed to be fed one, or a task went interactive.  Or otherwise, what might be unique about your playbooks to cause trouble here.

leth:
I remember checking once and ansible had no subprocesses and was using 100% CPU.

jimi-c:
# Merged

Hi!

issue is resolved.

Thank you!

Issue #6663, opened 2014-03-24
mysql_replication module does not recognize 'login_port=' parameter


diclophis:
##### Issue Type:

What kind of ticket is this?  `Bug Report`
##### Ansible Version:

```
ansible 1.4.5
```
##### Environment:

```
Darwin wesson.local 12.4.0 Darwin Kernel Version 12.4.0: Wed May  1 17:57:12 PDT 2013; root:xnu-2050.24.15~1/RELEASE_X86_64 x86_64
Linux mavenlink.vagrant 3.2.0-23-generic #36-Ubuntu SMP Tue Apr 10 20:39:51 UTC 2012 x86_64 x86_64 x86_64 GNU/Linux
```
##### Summary:

I was attempting to establish a master/slave pair on my vagrant host and ran into a problem where the `mysql_replication:` module does not understand `login_port=` ... I was able to work around the problem by using the `login_unix_socket=` parameter.
##### Steps To Reproduce:

Create a task with the following:

```
mysql_replication: login_port=3308 mode=stopslave
```
##### Expected Results:

The module will use the provided login_port: 3308 to communicate with the database
##### Actual Results:

```
msg: unsupported parameter for module: login_port
```

Jmainguy:
login_port was never included in this module so this is a feature request instead of a bug report. I have opened a PR which addresses adding login_port. Thank you for reporting this @diclophis

mpdehaan:
Hi!

Thanks very much for your interest in Ansible.  It sincerely means a lot to us. 

On September 26, 2014, due to enormous levels of contribution to the project Ansible decided to reorganize module repos, making it easier
for developers to work on the project and for us to more easily manage new contributions and tickets.

We split modules from the main project off into two repos, http://github.com/ansible/ansible-modules-core and http://github.com/ansible/ansible-modules-extras

If you would still like this ticket attended to, and believe this problem or idea is still present in the latest version of Ansible (1.7.2) or the development branch, we will need your help in having it reopened in one of the two new repos, and instructions are provided below.

We apologize that we are not able to make this transition happen seamlessly, though this is a one-time change and your help is greatly appreciated -- 
this will greatly improve velocity going forward.

Both sets of modules will ship with Ansible, though they'll receive slightly different ticket handling.

To locate where a module lives between 'core' and 'extras'
- Find the module at http://docs.ansible.com/list_of_all_modules.html
- Open the documentation page for that module
[...]

PR #8541, opened 2014-08-10
mysql_replication: Add feature - login_port, in response to #6663

Jmainguy:
This adds the feature login_port which was requested in #6663.

I used the same logic from mysql_db.

Tested this on Centos 6.5

### DIFF
--- a/library/database/mysql_replication
+++ b/library/database/mysql_replication
@@ -54,6 +54,12 @@ options:
         description:
             - mysql host to connect
         required: False
+    login_port:
+        description:
+            - Port of the MySQL server. Requires login_host be defined as other then localhost if login_port is used
+        required: False
+        default: 3306
+        version_added: "1.8"
     login_unix_socket:
         description:
             - unix socket to connect mysql server
@@ -115,6 +121,9 @@ EXAMPLES = '''
 
 # Change master to master server 192.168.1.1 and use binary log 'mysql-bin.000009' with position 4578
 - mysql_replication: mode=changemaster master_host=192.168.1.1 master_log_file=mysql-bin.000009 master_log_pos=4578
+
+# Check slave status using port 3308
+- mysql_replication: mode=getslave login_host=ansible.example.com login_port=3308
 '''
 
 import ConfigParser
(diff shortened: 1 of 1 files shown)


diclophis:
Thanks, I do have a question regarding this requirement:

```
 Requires login_host be defined as other then localhost if login_port is used
```

Please explain the requirement for this validation rule, I was thinking maybe it was because file sockets should be preferred over network sockets when doing localhost->localhost communication? But I can imagine this might cause more headache than anything else given that mysql is usually configured to listen to a single location by default. Or is there something about the underlying mysql bridge module that prevents `-h localhost -p 3307` ?

Jmainguy:
This is because of the way python connects to mysql. If login_host = localhost it will just use socket and ignore port. So if someone is defining a port other then 3306, but does not specify host other then localhost, then they will be very confused on why it wont connect (or will connect if they put the wrong port in). By making the module fail when port is defined but not host then we give the user a clear explanation of what needs to get fixed.

mpdehaan:
# Information Request

Hi!

We have some questions we'd like to know about before we can get this request queued up.   If you can help answer them, we'd greatly appreciate it:
- Can you please explain the hardcode of the MySQL port to the bottom of the commit?   I'm wondering if that can/should be pulled dynamically but did not exactly understand the details from it.  Thanks!

which is very active:

Jmainguy:
@mpdehaan I believe you are talking about this section of the code

db_connection = MySQLdb.connect(host=module.params["login_host"], port=int(module.params["login_port"]), user=login_user, passwd=login_password, db="mysql")

This is pulling the variable login_port which the user passes, so I believe it is dynamic in that sense and not hardcoded. This allows the port to be used when using tcp/ip connection to mysql server (before this was added, the only tcp/ip connection would be on default 3306)

If you are talking about this section

elif module.params["login_port"] != "3306" and module.params["login_host"] == "localhost":

This causes the module to fail if port is not 3306 (someone defined it other then 3306 ) and login_host is  localhost ( which is the default, meaning very likely someone did not define it, but they could have defined it as localhost if they love typing) 

MySQLdb.connect will not use tcp/ip connection if login_host == localhost, it will instead use the socket connection directly. This leads to confusion for end users who define a port and are surprised to see it connect (when using a port they know is bad), or see it not connect when they know it is listening on that port. 

This also matches what has been pulled into the mysql_db module. If you are ok with this, I would like to expand this PR to add login_port following the same logic to mysql_users and mysql_variable modules as I belive all 4 modules should use the same connection parameters and arguments.

mpdehaan:
# Module Repo Information

Hi!

Thanks you very much!

PR #8544, opened 2014-08-11
apt_repository: Check if key already exists before adding it

Tenzer:
According to my testing, adding a PPA repository with the `add_repository` module takes around 11-12 seconds, where around half of that time is spent on getting the key from Ubuntu's key servers, even if the key is already added. This commit optimizes that so it will only try to get the key from Ubuntu's servers if they key doesn't already exist. This takes the normal execution time down to around 6-7 seconds.


mpdehaan:
Merged, thank you!

PR #7614, opened 2014-05-31
Added json file module.

miguelcnf:
As written in the doc section this module will add, update or delete a specific key on a given json file.
It allows to manage json files without having to use templates, much like the ini_file module.

Miguel

PR #8554, opened 2014-08-11
New json file module.

miguelcnf:
Same as #7614 only moved to a proper branch.

As written in the doc section this module will add, update or delete a specific key on a given json file.
It allows to manage json files without having to use templates, much like the ini_file module.

Miguel

### DIFF
--- a/library/files/json
+++ b/library/files/json
@@ -1,0 +1,187 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2014, Miguel Fonseca <miguel.cnf () gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+import os
+
+try:
+    import json
+except ImportError:
+    import simplejson as json
+
+DOCUMENTATION = '''
+---
+module: json
+author: Miguel Fonseca
+short_description: Manage a json file
+description:
+  - This module will add, update or delete a specific key on a given json file
+version_added: "1.7"
+options:
+  json_file:
+    description:
+      - json file to be updated
+    required: true
+    aliases: [ name, file ]
+  key:
+    description:
+      - key to add/update/delete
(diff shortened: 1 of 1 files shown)


svisser:
Some thoughts:
- It may be good to add an action to validate a JSON file. I want to prevent deployment of an invalid JSON file (which is manually edited, I know) so I now have the following but that's admittedly ugly:

``` yaml
- name: Check whether file is JSON
  connection: local
  command: python -c "import json; json.loads(open('{{ playbook_dir }}/path/to/file.json').read());"
  register: json_validation
  failed_when: "'Traceback' in json_validation.stderr"
```

miguelcnf:
@svisser added a validate action:

```
- name: validate json
  json: name="/path/to/file.json" validate="yes"
```

You'll be able to run it as the above and if needed register the proper return code.

svisser:
This looks great!

One thing: the file object in `validate_json_file` isn't closed as in the other functions. How about reusing `read_json_file` so that the file object is closed?

svisser:
Actually reusing `read_json_file` is not such a good idea as you no longer have access to `f` then. So it would need to be solved by adding a `finally:` in `validate_json_file`.

miguelcnf:
You're right, as I was passing a string with `.read()` into `json.loads` unlike in the other methods where I had the file object passed into `json.load` I missed closing up the file.

Let me fix that in a minute.

svisser:
This is better but if the file fails to be opened then `f` is undefined when running the `finally` block. That'll lead to a `NameError` being raised.

miguelcnf:
I think that should cover all the cases.

Changed `finally` to an `else` block to only execute if no exceptions are raised.

svisser:
We're getting there but I just noticed you're not calling the `close` function, you're only accessing it.

miguelcnf:
That has bitten me in the back before. Thanks for noticing!

svisser:
Although I can't really use this module yet, I was playing around with the `get_key` method and it seems I can make it fail by doing:

``` python
>>> def get_key(my_dict, key):
...     return reduce(lambda d, k: d.get(k, {}), key.split('.'), my_dict)
...
>>>
>>> d = {}
>>> d['a'] = {}
>>> d['a']['b'] = 'c'
>>> d
{'a': {'b': 'c'}}
>>> get_key(d, 'a.b.c')
  File "<stdin>", line 2, in get_key
  File "<stdin>", line 2, in <lambda>
AttributeError: 'str' object has no attribute 'get'
```

I think the behaviour of `get_key` also has implications for empty dictionaries (which are valid in JSON). If I do:

``` python
>>> d['d'] = {}
>>> d
{'a': {'b': 'c'}, 'd': {}}
>>> get_key(d, 'd')
{}
```

then the `delete_key` function won't allow me to delete the key `'d'` because the value is considered False. It's made a bit worse by the fact that JSON also allows `null` (`None`) so `get_key` needs to distinguish between a value that is there (some of which can be falsey, such as `None`, `{}` and `""`) and a value that is really not there.

svisser:
Interesting project that may be relevant here: https://github.com/stefankoegl/python-json-patch

mpdehaan:
# Module Repo Information

Hi!

Thanks you very much!

PR #8560, opened 2014-08-11
ansible-vault should support script for --vault-password-file too

sivel:
Somehow I completely missed adding support to `ansible-vault` to allow `--vault-password-file` to be a script.

This PR adds this functionality to match the now existing functionality of `ansible` and `ansible-playbook`.

### DIFF
--- a/bin/ansible-vault
+++ b/bin/ansible-vault
@@ -27,6 +27,8 @@ import os
 import sys
 import traceback
 
+import ansible.constants as C
+
 from ansible import utils
 from ansible import errors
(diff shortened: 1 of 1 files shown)


mpdehaan:
Looks like this may need to also apply to the new "view" command on ansible-vault as well:

https://github.com/ansible/ansible/blob/devel/bin/ansible-vault

sivel:
I've updated this as requested. I have rebased again, and updated the view command as well.

I must have rebased last before the view command was added.

mpdehaan:
Merged, thank you!

PR #8564, opened 2014-08-12
Fix UnicodeDecodeError when replace interpreter in non-ascii file

Mic92:
I got this error, when using locale_gen module and set interpreter to "/usr/bin/python2" . I'm not quite sure whether assuming unicode is the best solution, but as far as I see at least modules from ansible itself all use utf-8.

### DIFF
--- a/lib/ansible/module_common.py
+++ b/lib/ansible/module_common.py
@@ -168,6 +168,10 @@ class ModuleReplacer(object):
                     facility = inject['ansible_syslog_facility']
                 module_data = module_data.replace('syslog.LOG_USER', "syslog.%s" % facility)
 
+            try:
+                module_data = module_data.decode('utf-8')
+            except UnicodeDecodeError:
+                pass
             lines = module_data.split("\n")
             shebang = None
             if lines[0].startswith("#!"):


mpdehaan:
Typically this code has needed a try/except around any .decode()'s but can test certainly.

Thanks!

Mic92:
- Add try block

Mic92:
Hi. What is actually missing for this small change to merge?

abadger:
For being a oneline change, this has far reaching consequences.  And I'm not sure if it is definitely the right thing to do.  When we read the file in, we get it as bytes.  Then we add the module parameters as other byte strings to the text.  We do other manipulations and then at some point we're going to write the file back out as bytes to the disk.

The two parts that I'm not sure about are whether we can safely convert the file to unicode once we've substituted the parameters in as byte strings (my gut says yes at the moment because there is no way to get non-decodable bytes through the jinja templating step.  But that's something of an anti-feature so it may change in the future.)  and whether we need the file to be bytes again by the time we return the value.  My gut says maybe here -- we'd want it to be bytes when we write the file but perhaps we can delay keep the data as unicode until just before we write the file and then transform it there.  That certainly seems like a better way to go about it for v2 in any case.

It may be better to see the traceback here so we can know what to do.  I have a feeling that for v1 it might be a lot less dangerous to turn something in here (possibly the value of ansible_pyhton_interpreter?) into a utf-8 byte str than to turn the file data into unicode type.  For v2, where we want to use unicode internally, it may depend on how much we're doing to the data before it gets written to again.

abadger:
I'm not able to replicate this either (although I see what combination of things could make it happen).

(1) locale_gen.py is using "smart quotes" ( “” ) instead of standard double quotes (").  This is not needed in the module so it should arguably be replaced with standard double quotes.

(2)

``` python
lines[0] = shebang = "#!%s %s" % (inject[interpreter_config], " ".join(args[1:]))
module_data = "\n".join(lines)
```

If inject[interpreter_config] contains a unicode string then lines[0] will be unicode.  When we attempt to join that with the byte strings that are in lines (of which one contains non-ascii characters) we get a UnicodeError traceback.  We need to make the types match here.  It's just a question of whether it needs to be unicode or byte str.

For v1, I'm going to fix both of those things.  For v2, the module change will be picked up automatically but I'm not sure yet what we want to do with the module_replacer code.

abadger:
Okay, those commits should fix this issue for you.  Like I say, I'm not sure if in v2 we'd want to continue to keep the data in a str or transform it into unicode.  If unicode we may want to transform it early (right after the f.read())  rather than later.  If you just wanted this error fixed, you probably don't need to think further about it.  If you wanted it to use unicode then you might want to check back once the v2 tree becomes the new devel branch.

# Closing This Ticket

Hi!

issue is resolved.

Thank you!

PR #8569, opened 2014-08-12
Fix typo in CHANGELOG.md

drewblas:
This change occured int he ec2 module, not ec2_vol

I was searching for this change, thinking it related to setting "gp2" on EBS.  But this is actually a different change and is a typo in the CHANGELOG.


mpdehaan:
merged, thanks!

PR #5987, opened 2014-02-13
Allow ec2 inventory to use a boto profile

willthames:
##### Issue Type:

Feature Pull Request
##### Ansible Version:

Based on:
ansible 1.5 (devel 6f405c8970) last updated 2014/02/13 18:12:39 (GMT +1000)
##### Environment:

N/A
##### Summary:

This allows the EC2 inventory plugin to be used with
the same configuration against different EC2 accounts

Added documentation on profiles to EC2 dynamic inventory doc

Only tries to use profiles if --profile argument is given
to maintain compatibility will boto < 2.24.

Works around a minor bug in boto where if you try and use
a security token with a profile it fails (boto/boto#2100)

PR #8582, opened 2014-08-12
ec2.py boto_profile option for multiple AWS account credentials in boto file

ghost:
Add an ec2 inventory option "boto_profile" that allows
the use of boto profiles for separating credentials as
specified in

http://boto.readthedocs.org/en/latest/boto_config_tut.html

### DIFF
--- a/plugins/inventory/ec2.ini
+++ b/plugins/inventory/ec2.ini
@@ -73,3 +73,8 @@ nested_groups = False
 
 # If you want to exclude any hosts that match a certain regular expression
 # pattern_exclude = stage-*
+
+# A boto configuration profile may be used to separate out credentials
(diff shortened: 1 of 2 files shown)


willthames:
#5987 allows an EC2_PROFILE environment variable to be set to configure the profile.

ghost:
Yes it does. I prefer not encoding static information in environment variables. With a setting in the .ini file, you can have inventory directories with ec2.py and ec2.ini for multiple AWS accounts where hosts might live.

Not to say the approach of #5987 is wrong, ideally both would be supported. That is the common paradigm with boto/aws CLI utilities.

amenonsen:
@jeffb-bt @jalaziz I guess this PR should be closed in favour of #9130 now?

gregdek:
Closing in favor of #9130 (which I hope gets merged soon!)

Issue #8299, opened 2014-07-26
keystone_user documentation lists invalid option


ericwb:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.6.3
##### Environment:

Ubuntu 12.04.4
##### Summary:

The keystone_user documentation on the ansible website lists "description" as an option.  It should be "tenant_description" after looking at the code on git.
##### Steps To Reproduce:

Just try to use the keystone_user module with description.
##### Expected Results:

I'd expect description to work.
##### Actual Results:

Ansible run fails with:

failed: [controller01] => {"failed": true}
msg: unsupported parameter for module: description

mpdehaan:
Something you would like to submit a pull request for?

(Not required, just asking, because you've already been through the code in git...  it's the DOCUMENTATION block in that file that needs to change, so it's pretty straightforward)

Thanks!

ericwb:
Sure, I'll submit a patch.

PR #8587, opened 2014-08-12
Update keystone_user doc for tenant_description

ericwb:
The description var should be tenant_description.

Closes-bug: #8299


mpdehaan:
Merged, thank you!

PR #8597, opened 2014-08-13
add implict --sudo (edit: for /bin/ansible)

hudecof:
add implict --sudo when --ask-sudo-pass or --sudo-user=SUDO_USER is specified
This code was removed between version 1.5.6 and 1.6.0

### DIFF
--- a/bin/ansible
+++ b/bin/ansible
@@ -145,9 +145,10 @@ class Cli(object):
             callbacks.display("No argument passed to %s module" % options.module_name, color='red', stderr=True)
             sys.exit(1)
 
-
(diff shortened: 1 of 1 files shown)


bcoca:
+1, the distinction made sense for ansible-playbook but not for bin/ansible

teriyakichild:
+1.  Doesn't make sense to prompt for a sudo password, unless you want to run with sudo.

bcoca:
actually it does make sense, when you are going against several hosts with different connection settings in the inventory, with ansible_sudo=true vs others that do not and would break by forcing sudo.

PR #8600, opened 2014-08-13
Allow ansible-galaxy to install roles from URLs

willthames:
ansible-galaxy can now refer to SCM URLs (git and hg at this point)
for role names
Dependencies need to use the full SCM URLs too.
Otherwise all seems to work well

Test rolesfile

```
http://bitbucket.org/willthames/git-ansible-galaxy,v1.1
https://bitbucket.org/willthames/hg-ansible-galaxy
```

(works with ssh too)

### DIFF
--- a/bin/ansible-galaxy
+++ b/bin/ansible-galaxy
@@ -696,10 +696,12 @@ def execute_install(args, options, parser):
 
     roles_done = []
     if role_file:
-        # roles listed in a file, one per line
-        # so we'll go through and grab them all
         f = open(role_file, 'r')
-        roles_left = f.readlines()
+        if role_file.endswith('.yaml') or role_file.endswith('.yml'):
+            roles_left = map(ansible.utils.role_yaml_parse, yaml.safe_load(f))
+        else:
+            # roles listed in a file, one per line
+            roles_left = map(ansible.utils.role_spec_parse, f.readlines())
         f.close()
     else:
         # roles were specified directly, so we'll just go out grab them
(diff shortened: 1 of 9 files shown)


mpdehaan:
As posted to the ML, how does the code know a http:// and https:// URL is git or hg or svn?

I'm wondering if maybe we need to do git+https:// or something.

willthames:
@mpdehaan see lines 334 onwards. It uses `git ls-remote` and `hg identity` - whichever has an exit status of 0 is the right SCM to use. 

No real need for the git+https although might want to extend it to svn, bzr etc.

willthames:
Changed to use git+http as it does make determining SCM simpler and more explicit. 

I can squash commits before this is accepted but wanted to preserve history for now in case this needs further revision.

mpdehaan:
Checking this out now...

mpdehaan:
Awesome, I've merged this, also adding a docs section to the ansible-galaxy docs page.

I've also upgraded error handling in a few minor places, and made the system understand that github.com URLs are git+ protocol in a few cases, as I originally found the "non-tarball" message a little confusing and wanted to fix that up.

This works solidly and I greatly appreciate the tests.

Docs should appear soon and I'll also write the list about this.

We also need to update the galaxy website eventually to cross-reference docs about these options.

Thank you very much this will be a very very popular feature!

mpdehaan:
Closing this, I'll reply to the list discussion in a bit.

Thanks!

ibratoev:
it already is ;)

willthames:
Awesome, that's great news.

I haven't had a look yet but on your note regarding github, it should be
able to handle downloading from a github URL pointing to an archive rather
than a repo. I think there's an integration test for that.

Will

willthames:
Created #8714 for the github archive problem.

PR #8602, opened 2014-08-13
Encrypts the vault file after edition only if the contents changed

pfigue:
Whenever I edit a vault-protected file, it is re-encrypted even if I discard the changes or don't do any change, so git thinks that the file was actually modified. It forces me to git-checkout it to the previous version.

To avoid this issue I wrote this little patch.


direvus:
+1, this has been bothering me too.

amenonsen:
This PR does something I've _really_ wanted. I've rebased and resubmitted as #11740. This PR can now be closed.

Issue #8413, opened 2014-08-03
Install of bonnie++ with homebrew module fails


mjaschen:
##### Issue Type:

Bug Report
##### Ansible Version:

1.6.10
##### Environment:

OS X 10.9.4
##### Summary:

The installation of _bonnie++_ with the homebrew module fails.
##### Steps To Reproduce:

Write a playbook task:

```
    - name: Install homebrew applications
      homebrew: >
        name="{{ item }}"
        state=latest
      with_items:
        - bonnie++
```

Run the playbook.
##### Expected Results:

`bonnie++` should be installed.
##### Actual Results:

Ansible exists with an error message "msg: Invalid package: bonnie++."

mpdehaan:
I'm not sure from the bugreport where there's a generic issue with things without alphanumeric package names or the issue is because of something else.

Something you might like to help investigate or work on, perhaps?

mjaschen:
Sure, I'll take a look :-)

mjaschen:
Hi Michael,

after some tests I'm pretty sure it's related to the `+` signs in the package name. I tried other Homebrew packages with `+` in the package name (`vera++`, `png++`, etc.) and all failed to install with Ansible's homebrew module.

I looked for other unusual package names and found out that installing packages with a dot in the package name fails as well (e. g. `virtualhost.sh`, `shell.fm`, etc.)

Installing other Homebrew packages with alphanumeric package names (as well as packages containing a `-` sign) works fine.

mpdehaan:
Closing per above bugfix, thanks!

PR #8612, opened 2014-08-13
Homebrew module: Added `+` and `.` characters to the package name regex.

mjaschen:
Solves installation failures for packages like `bonnie++` or
`virtualhost.sh` as described in #8413.


mpdehaan:
Merged, thank you!

Issue #7432, opened 2014-05-16
async task with poll set to 0 fails when it is skipped


jasonarewhy:
##### Ansible Version:

1.6.1, devel
##### Environment:

Centos 6.5, OS X 10.9.2
##### Summary:

An async task with 'poll: 0' fails only when the 'when' condition causes it to be skipped. If I remove the poll option or set it to a number higher than 0, the task is successfully skipped. If I remove the 'when' condition, the async task is successful with 'poll' set to 0.
##### Steps To Reproduce:

Create an async task with 'poll: 0' and a when condition that causes it to be skipped.

```
- name: test
  hosts: localhost

  tasks:
    - name: a local task
      shell: echo hello
      async: 30
      poll: 0
      when: a_variable is defined
```
##### Expected Results:

The async task should be skipped.
##### Actual Results:

```
TASK: [a local task] **********************************************************
skipping: [localhost]
Traceback (most recent call last):
  File "/Users/jyoung/lib/ansible-main/bin/ansible-playbook", line 317, in <module>
    sys.exit(main(sys.argv[1:]))
  File "/Users/jyoung/lib/ansible-main/bin/ansible-playbook", line 257, in main
    pb.run()
  File "/Users/jyoung/lib/ansible-main/lib/ansible/playbook/__init__.py", line 319, in run
    if not self._run_play(play):
  File "/Users/jyoung/lib/ansible-main/lib/ansible/playbook/__init__.py", line 673, in _run_play
    if not self._run_task(play, task, False):
  File "/Users/jyoung/lib/ansible-main/lib/ansible/playbook/__init__.py", line 449, in _run_task
    results = self._run_task_internal(task)
  File "/Users/jyoung/lib/ansible-main/lib/ansible/playbook/__init__.py", line 413, in _run_task_internal
    self.runner_callbacks.on_async_ok(host, res, poller.runner.vars_cache[host]['ansible_job_id'])
KeyError: 'ansible_job_id'
```

mpdehaan:
"An async task with 'poll: 0' fails only when the 'when' condition causes it to be skipped"

I think you mean something totally different here, that it fails with a traceback and the when condition isn't evaluated.

jasonarewhy:
You are correct in that by 'fail' I meant ansible exits with a traceback. However, it does appear the when condition is being evaluated (at least partially) since I am seeing 'skipping: [localhost]' just before the traceback.

The traceback only occurs when poll is 0 and the when condition would cause the task to be skipped.

[...]

PR #8616, opened 2014-08-14
fix the 'KeyError' when skipping asynchronous tasks with poll: 0

mountkin:
This is related to #7432.
I ran into this issue just now, so I wrote this patch. Hopefully it'll fix the issue. 

The following playbook will reproduce the exception.

``` yaml

---
- hosts: localhost
  tasks:
  - name: test conditional "fire and forget" task (when skipped)
    command: sleep 5
    poll: 0
    async: 10
    when: False
```

### DIFF
--- a/lib/ansible/playbook/__init__.py
+++ b/lib/ansible/playbook/__init__.py
@@ -437,7 +437,7 @@ class PlayBook(object):
                 results = self._async_poll(poller, task.async_seconds, task.async_poll_interval)
             else:
                 for (host, res) in results.get('contacted', {}).iteritems():
-                    if not ('skipped' in res and res['skipped']):
+                    if not res.get('skipped'):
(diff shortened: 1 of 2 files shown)


lib/ansible/playbook/__init__.py
```diff
@@ -437,7 +437,8 @@ def _run_task_internal(self, task):
                 results = self._async_poll(poller, task.async_seconds, task.async_poll_interval)
             else:
                 for (host, res) in results.get('contacted', {}).iteritems():
-                    self.runner_callbacks.on_async_ok(host, res, poller.runner.vars_cache[host]['ansible_job_id'])
+                    if not ('skipped' in res and res['skipped']):
```
ciupicri: It could be replaced with `if not res.get('skipped'):`

mountkin: Thanks for reminding me of that.

amenonsen:
This PR can be closed, the example playbook no longer causes any error when run on devel.

PR #8622, opened 2014-08-14
Fix typo: & -> and

ciupicri:
`and` is the boolean operator for AND in [Python](https://docs.python.org/2/reference/expressions.html#boolean-operations).
`&` is the bitwise operator for AND in [Python](https://docs.python.org/2/reference/expressions.html#binary-bitwise-operations).


mpdehaan:
Merged, thanks!

Issue #7372, opened 2014-05-12
Since 1.6.0 `template` always does `chown`, causing 'Operation not permitted'


TallGirlVanessa:
After upgrading to 1.6.0, I can no longer use the `template` module to write to folders I can't `chown`. I isolated this with `git bisect` to commit 8f778a83dfaaea142663b878bafd84be3afee3f6: looks like the `template` module now does a `chown` whether or not the play specifies the destination's owner/group.

This is a problem since our deploy script is sudoed to a non-root user: I first noticed this issue when I tried to deploy to one of our integration environments. Surprisingly, it broke on my local machine as well: we develop using Parallels VMs managed by Vagrant, which keep our code in a Parallels shared folder which is very lenient with permissions (the files appear to be owned by whoever is trying to access them at the time :smiley:). Turns out this arrangement (a mount of type `prl_fs`) does not permit `chown`, even by root. Could you please prevent `template` from using `chown` when owner/group is not specified?

Example plays:

```
- name: Deploy | Create Flask instance folder
  file: path={{ code_home }}/instance state=directory

- name: Deploy | Write Ansible-customized config
  template: src=config_from_ansible.py.jinja2 dest={{ code_home}}/instance/config_from_ansible.py
```

Result:

```
__________________________________________________________
< TASK: deploy_web | Deploy | Create Flask instance folder >
 ----------------------------------------------------------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||


ok: [33.33.33.50] => {"changed": false, "gid": 1000, "group": "vagrant", "item": "", "mode": "0755", "owner": "vagrant", "path": "/var/www/bridge-it/instance", "size": 102, "state": "directory", "uid": 1000}
 _____________________________________________________________
< TASK: deploy_web | Deploy | Write Ansible-customized config >
 -------------------------------------------------------------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||


failed: [33.33.33.50] => {"failed": true, "item": ""}
msg: Could not replace file: /home/vagrant/.ansible/tmp/ansible-tmp-1399918533.3-108641875157900/source to /var/www/bridge-it/instance/config_from_ansible.py: [Errno 1] Operation not permitted: '/var/www/bridge-it/instance/.config_from_ansible.py.11003.1399918533.48'

FATAL: all hosts have already failed -- aborting
```

[...]

PR #8624, opened 2014-08-14
atomic_move() only works as root (edit:  "sometimes atomic_move will chown things when it cannot")

l1k:
Linux and BSD derivatives do not allow unprivileged users to "give away" files to others for security reasons. (System V derivatives allow that but they're rare nowadays.)

Thus, the additional chown in atomic_move() which was apparently introduced with v1.6 and which was already objected to in issue #7372 breaks usage of Ansible by non-root users. An example use case is a group-writable directory on a remote machine where users of the same group may drop off configuration files, possibly overwriting each other's files with newer versions. This used to work with versions prior to v1.6 and is now broken.

It used to be a key selling point of Ansible that it allows for non-root operation, in fact this is still mentioned in glossary.rst ("Ansible does not require root logins"), suggesting that the behaviour of v1.6 and onwards is in fact a bug.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -1145,9 +1145,13 @@ class AnsibleModule(object):
                 if self.selinux_enabled():
                     self.set_context_if_different(
                         tmp_dest.name, context, False)
-                tmp_stat = os.stat(tmp_dest.name)
-                if dest_stat and (tmp_stat.st_uid != dest_stat.st_uid or tmp_stat.st_gid != dest_stat.st_gid) and os.getuid() == 0:
-                    os.chown(tmp_dest.name, dest_stat.st_uid, dest_stat.st_gid)
+		try:
+		    tmp_stat = os.stat(tmp_dest.name)
+		    if dest_stat and (tmp_stat.st_uid != dest_stat.st_uid or tmp_stat.st_gid != dest_stat.st_gid):
+			os.chown(tmp_dest.name, dest_stat.st_uid, dest_stat.st_gid)
+		except OSError, e:
+		    if e.errno != errno.EPERM:
+			raise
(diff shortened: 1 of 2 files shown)


bcoca:
It is meant to work with root or when you sudo, ansible cannot give you
more privileges than the system would.

The additional chown was added because people complained about the files
being overwritten with different permissions than expected (they expected
them to be preserved).

But I agree with you that this should not error out when you CAN overwrite
the file and you CANNOT change permissions due to not using root/sudo/su,
so maybe a change to make this an 'optimistic' update and continue with
just a warning/feedback/docs that permissions changed/can change in certain
circumstances.

## 

Brian Coca
Pedo mellon a minno

mpdehaan:
I agree with @bcoca above about the permissions system and it trying to be as atomic as possible when it can.  These are good things.

I also agree about error avoidance being a good thing, but would need to see an example of how you are making it catch that scenario as it hasn't been a common report.

Ansible continues to work without root, so I do contest the assertion that it does not.

mpdehaan:
@bcoca, so +1 for the idea of catching the exception around the chown and in that case using permissions as available?

bcoca:
also, a better patch would use os.geteuid() as getuid will only work if you
login as root, will not catch sudo privs to root​

l1k:
Hm, there's already code right at the beginning of atomic_move() for an optimistic chown (for the case when src and dest reside on the same filesystem so that os.rename() works). Duplicate that code further down to optimistically chown the tmp file?

In line 1139 (also in atomic_move()), os.getuid() is used instead of os.geteuid(). Is that correct?

l1k:
Et voilà, optimistic chown plus unit test.

jimi-c:
# Merged

Hi!

Thank you!

PR #8627, opened 2014-08-14
fixes #8626 by reverting 0d8b81cd

### LINKED ISSUE #8626: urpmi module behaves differently when python-rpm is installed
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.6.8
##### Environment:

Mageia 4
##### Summary:

from 0d8b81cd95a2f847f083fe03d769e7206415542f, urpmi module uses python-rpm, if present, instead of rpm -q to speed up package searches. Unfortunately search results aren't the same.
The two method are equivalent only when requesting packages with basename (e.g. 'mariadb'), not when specifying a revision (e.g. 'mariadb-5.8.3')
With plain rpm the latter is handled correctly, while using python rpm no match is detected.
Requesting a specific package version imho is an essential feature, thus the mentioned change should be reversed
##### Steps To Reproduce:

create an urpmi task requesting a specific package version, e.g.
    ---
    - hosts: all
      user: root
      tasks:
        - name: mariadb
           urpmi: pkg=mariadb-5.5.38

The requested package will be installed correctly. On subsequent runs the task will fail because the package is already installed, but the urpmi module tries to install again aborting when installation fails.
##### Expected Results:

urpmi module should detect if mariadb-5.5.38 is already installed and skip the installation
##### Actual Results:

```
failed: [host] => (item=mariadb-5.5.38) => {"failed": true, "item": "mariadb-5.5.38"}
msg: 'urpmi mariadb-5.5.38' failed: 
```

rghe:


mpdehaan:
Merged, thanks!

I'm ok with this given my general feelings on the rpm python module.   If anyone else has problems with this, please let us know on one of the mailing lists.

PR #8632, opened 2014-08-14
fixed ignore_errors during copy files for relative paths

jeviolle:
ignore_errors would never be evaluated for roles using copy. This however works fine for standard play books. output examples below.

https://gist.github.com/jeviolle/202274c74247c12efe4f


mpdehaan:
Can you please share the error before that was not being ignored and the playbook that reproduces it?

Some things with ignore_errors we do want to continue to be fatal, such as missing sources.

jeviolle:
This gist shows an example playbook/role and the output

https://gist.github.com/jeviolle/202274c74247c12efe4f

PR #8640, opened 2014-08-15
File Module: Allow path to be relative with state=directory

sivel:
Currently, when you specify a relative path in `path` with `state=directory`, ansible tries to create the directory in `/`:

```

---
- hosts: all
  tasks:
    - file:
        path: foo
        state: directory
```

```
TASK: [file ] *****************************************************************
  File "/Users/matt/.ansible/tmp/ansible-tmp-1408123613.88-251928671518108/file", line 1677, in <module>
    main()
  File "/Users/matt/.ansible/tmp/ansible-tmp-1408123613.88-251928671518108/file", line 216, in main
    os.mkdir(curpath)
OSError: [Errno 13] Permission denied: '/foo'
```

This pull request will run `os.path.realpath` on the path before trying to split the path using the `/` separator.

However, I am wondering if it wouldn't actually be better to do this in a more global way such as:

(Although I am not sure what other consequences that may have)

``` diff
diff --git a/library/files/file b/library/files/file
index 6e663e5..67f2c5b 100644
--- a/library/files/file
+++ b/library/files/file
@@ -120,7 +120,7 @@ def main():
     src = params['src']

     # modify source as we later reload and pass, specially relevant when used by other modules.
-    params['path'] = path = os.path.expanduser(params['path'])
+    params['path'] = path = os.path.realpath(os.path.expanduser(params['path']))

     # short-circuit for diff_peek
     if diff_peek is not None:
```

### DIFF
--- a/library/files/file
+++ b/library/files/file
@@ -210,7 +210,7 @@ def main():
                 module.exit_json(changed=True)
             changed = True
             curpath = ''
-            for dirname in path.split('/'):
+            for dirname in os.path.realpath(path).split('/'):
                 curpath = '/'.join([curpath, dirname])
                 if not os.path.exists(curpath):
                     os.mkdir(curpath)


major:
@sivel - This isn't the behavior I'd expect:

```
>>> import os
>>> path='/var/www/html'
>>> os.path.realpath(path).split('/')
['', 'var', 'www', 'html']
>>> path='relative/to/pwd'
>>> os.path.realpath(path).split('/')
['', 'root', 'ansible', 'relative', 'to', 'pwd']
```

If I'm reading the code correctly, it seems like directories under your current working directory could be altered when specifying a relative path.

mpdehaan:
Working directory in ansible is a little confusing to me, because you're logging in as different users and using various temp dirs.

Is this mostly about local_action type plays?

major:
@mpdehaan - I think `local_action` tasks or `ansible_connection=local` hosts would benefit from this, but I'm not sure about @sivel's use case.  My examples are in #8642.

jimi-c:
After some discussion and analysis of the issue, we've decided to merge in @major's version of this patch. Thanks!

PR #8641, opened 2014-08-15
Tense correction

ianmiell:


mpdehaan:
Merged, thanks!  This will show up in the next docs build/push.

PR #8642, opened 2014-08-15
File module cannot create relative paths

major:
When trying to create a directory relative to the current working
directory, a directory is created at the root of the filesystem
instead.  This patch ensures that directories specified with relative
paths will be created in the current working directory.  Fully
qualified paths aren't affected

### DIFF
--- a/library/files/file
+++ b/library/files/file
@@ -210,8 +210,15 @@ def main():
                 module.exit_json(changed=True)
             changed = True
             curpath = ''
-            for dirname in path.split('/'):
+            # Split the path so we can apply filesystem attributes recursively
+            # from the root (/) directory for absolute paths or the base path
+            # of a relative path.  We can then walk the appropriate directory
+            # path to apply attributes.
+            for dirname in path.strip('/').split('/'):
                 curpath = '/'.join([curpath, dirname])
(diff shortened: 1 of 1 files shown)


mpdehaan:
Can you share a play that replicates the problem this addresses?  I'm curious if this is about local_action plays because if doing something on remotes, I'm not sure what base path a user would expect would be sensible.  Thanks!

major:
Here's an example: https://gist.github.com/major/db1ecc28a5644169cc69

It's nice to be able to use the working directory when using `ansible_connection=local`. It was working great in 1.6.10 and earlier.

mpdehaan:
can we determine consensus between this and #8640 as to which we want ?

major:
When I patched in the code from #8640, I didn't get the desired output from pre-1.7.0 ansible versions.  The patch that I added here seems to work well.
- If you provide a fully qualified path, it walks the directory tree from the root of the filesystem to the directory requested.
- If you provide a relative path, it walks the directory tree from your current working directory to the final directory requested.

major:
Testing both: https://gist.github.com/major/2de0053313bc515b1e30

jimi-c:
# Merged

Hi!

Thank you!

PR #8648, opened 2014-08-16
Include tip for getting the 'walking a group' to work

veegee:
The example for 'walking a group to find all IP addresses' does not work if facts for all hosts are not gathered.
This example to add an empty tasks list for all hosts forces ansible to populate its facts array.


mpdehaan:
Thanks for this addition, I believe this is well covered elsewhere in the doc and we'd have to repeat this tip in many other sections, so I don't think we need to include this in this particular section.  We hope you understand our thoughts on organization and definitely would welcome additional docs additions in the future.

Thank you!

Issue #8387, opened 2014-08-01
locale_gen does not generate locales on Arch Linux


tummychow:
##### Issue Type:

Bug report
##### Ansible Version:

1.6.10
##### Environment:

Arch Linux (both master and slave)
##### Summary:

locale_gen doesn't generate locales on Arch Linux.

The [locale_gen](https://github.com/ansible/ansible/blob/devel/library/system/locale_gen#L66) module looks for lines starting with a hash and a space, and removes the hash to enable locale generation for those lines. But Arch's locale.gen doesn't have spaces (only the hash) so no lines get uncommented. Since no locales are enabled, locale-gen does nothing.
##### Steps To Reproduce:

Create a fresh Arch Linux system and install openssh/python2 on it. Add it to your Ansible hosts.

```
$ ansible -m locale_gen -a "name=en_US.UTF-8 state=present" --extra-vars "ansible_python_interpreter=/usr/bin/python2" <your Arch Linux target here>
$ ssh <your arch linux target here> locale -a
```
##### Expected Results:

The Ansible ad-hoc command should succeed and indicate "changed"
The ssh command `locale -a` should list all the locales on the system. `en_US.UTF-8` should be on the list.
##### Actual Results:

The Ansible command does succeed, but C and POSIX are the only locales listed aftewards. `en_US.UTF-8` does not appear.

ghost:
can you post a snippet of ArchLinux's locale.gen for comparison? I'm guessing a regex that looks for the hash and an optional space should be used. The `replace_line` is using String.replace and that would have to be replaced.

tummychow:
Sure:

```
#en_IE@euro ISO-8859-15  
#en_IN UTF-8  
#en_NG UTF-8  
#en_NZ.UTF-8 UTF-8  
#en_NZ ISO-8859-1  
#en_PH.UTF-8 UTF-8  
#en_PH ISO-8859-1  
#en_SG.UTF-8 UTF-8  
#en_SG ISO-8859-1  
en_US.UTF-8 UTF-8  
#en_US ISO-8859-1  
#en_ZA.UTF-8 UTF-8  
#en_ZA ISO-8859-1  
#en_ZM UTF-8  
#en_ZW.UTF-8 UTF-8  
#en_ZW ISO-8859-1  
#es_AR.UTF-8 UTF-8  
```

As you noted, the fix for the issue would be to make the spaces optional.

mpdehaan:
Hi!

[...]

PR #8656, opened 2014-08-17
Fix for issue #8387. Also: Add a new line to /etc/locale.gen if the spec...

ercpe:
...ified locale wasn't listed.


ercpe:
Sorry, messed things up. Will open another pr.

PR #8388, opened 2014-08-01
Complete re-work to work with current nova, other features

CaptTofu:
Hi there! Another shot - and I had decided to completely revamp it to be like the rax dynamic inventory plugin (and gave attribution)

PR #8657, opened 2014-08-17
Fix up nova inventory to be modern

emonty:
Update the nova inventory module to be more like the ec2 and rax modules. Notably:
- added support for nova.ini to express multiple clouds that the inventory should span
- added caching similar to the ec2 module
- aligned authentication and floating ip handling with the other nova modules
- added "--private" flag so that one can specify "please use the private IP (fixed IP) of the instance as ansible_ssh_host

### DIFF
--- a/plugins/inventory/nova.py
+++ b/plugins/inventory/nova.py
@@ -43,10 +43,12 @@ NOVA_CONFIG_FILES = [
 ]
 
 NON_CALLABLES = (basestring, bool, dict, int, list, NoneType)
-flavor_cache = None
 
 
 class OpenStackCloud(object):
+
+    flavor_cache = None
+    image_cache = dict()
 
     def __init__(self, name, username, password, project_id, auth_url,
                  region_name, service_type, insecure):
@@ -66,9 +68,14 @@ class OpenStackCloud(object):
     def get_region(self):
         return self.region_name
 
+    def get_flavor_name(self, flavor_id):
+        if not self.flavor_cache:
+            self.flavor_cache = dict([(flavor.id, flavor.name) for flavor in self.client.flavors.list()])
+        return self.flavor_cache.get(flavor_id, None)
+
     def connect(self):
         # Make the connection
-        client = nova_client.Client(
+        self.client = nova_client.Client(
             self.username,
             self.password,
             self.project_id,
@@ -79,7 +86,7 @@ class OpenStackCloud(object):
         )
 
         try:
-            client.authenticate()
+            self.client.authenticate()
         except exceptions.Unauthorized, e:
             print("Invalid OpenStack Nova credentials.: %s" %
                   e.message)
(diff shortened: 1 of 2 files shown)


emonty:
This includes and builds off of @CaptTofu work in #8388 I've tested that this works on both HP and Rackspace clouds - and that the multi-cloud support works with both of those clouds in the same config. I've also tested that it works both with and without config files, and that the config file overrides values found in the environment.

mpdehaan:
Looks reasonable to me, can you post to ansible-devel and see if we can get some +1s from some other openstack users perhaps?

My only other comment was some fields now have regions set in the config, if it worked for all regions before we _may_ want to see that region parameter commented out and have a default of all regions, or does that not make sense?

Thinking about people who upgrade that have a previous INI file, though since inventory plugins are not really part of a release, no one should really upgrade and get any surprises.

CaptTofu:
Michael,

Indeed, thank you. I emailed the dev list asking for reviewers. I think that my work and then Monty’s great work on top (I merged his pull request) will make this plugin top notch!

Regards,

Patrick

emonty:
@mpdehaan - region is actually a required parameter for python-novaclient. The previous version filled in a fairly bogus region param if you didn't specify one - so I think it's fairly unlikely that someone was relying on that for a production usage - unless they actually have a production cloud with some very strange config ... I'll follow up on the mailing list thread.

plugins/inventory/nova.py
```diff
@@ -30,103 +37,316 @@
...
+                flavor_name = cloud.get_flavor_name(server.flavor['id'])
+                if flavor_name:
+                    groups['flavor-%s' % flavor_name].append(server.name)
+
+                groups['image-%s' % server.image['id']].append(server.name)
+                image_name = cloud.get_image_name(server.image['id'])
+                if image_name:
+                    groups['image-%s' % image_name].append(server.name)
+
+                # And finally, add an IP address
```
ibratoev: This is duplicated 15 lines above that ;)

plugins/inventory/nova.py
```diff
@@ -30,103 +37,316 @@
...
+                if (self.private is True):
+                    hostvars[server.name]['ansible_ssh_host'] = private[0]
```
ibratoev: This will break the whole inventory if a single host has no network (for whatever reason).
Maybe ignore the host if it has no IP address?

CaptTofu: Thank you very much for your review of this! Will fix.

emonty:
Oops on the duplication. Great catch on the breakage. Patch incoming...

ibratoev:
What do you think about moving the _private_ option to the ini config file?
Currently, for one reason or another, I need to use _private_ ip addresses exclusively and to accomplish that I just changed the code itself.
Other than that the inventory is a great improvement - faster and cleaner than the current one. I have been using it extensively the last couple of weeks.
+1 for merging it sooner.

emonty:
@ibratoev I think that's a great idea! I'll do that right now.

emonty:
@ibratoev There you go - I tested this locally that it still works on the command line and that it also works setting it in the config file. Can you let me know if it works for you how you expect?

t0mk:
Hi, I did another rework of the Nova inventory, I tried to cover your features and add some more, and clean up the code. If you are interested, please see:
https://github.com/ansible/ansible/pull/9378

PR #8659, opened 2014-08-18
rds_module: Fix AWS connection

willthames:
Use the common AWS connection improvements so that security_token,
boto_profile etc. are taken into account


mpdehaan:
Merged, thanks!

Issue #8206, opened 2014-07-19
Windows RM commands timeout at 60 seconds, failing long running commands


uxp:
##### Issue Type:

Bug Report
##### Ansible Version:

1.7 (f0aab11)
##### Environment:

Host: Mac OS X 10.9.4 (13E28)

Client: Microsoft Windows NT 6.3.9600.0 (Server 2012-r2)
##### Summary:

Command connections timeout at 60 seconds, making some commands unusable (win_get_url with large files, for example)
##### Steps To Reproduce:

This should be a sufficiently basic `Vagrantfile` (for vmware fusion, customize as needed). Vagrant 1.6+ needed for Vagrant Cloud support. Simply run `vagrant up`, specifying the provider as needed, while in the directory containing the following two files.

```
# -*- mode: ruby -*-
# vi: set ft=ruby :

Vagrant.configure("2") do |config|
  config.vm.define "win-server-2012-r2-std-sql"

  # https://vagrantcloud.com/hplogsdon/win-server-2012-r2-std
  config.vm.box = "hplogsdon/win-server-2012-r2-std"

  config.vm.provider :vmware_fusion do |v, override|
    v.gui = false
    v.vmx["memsize"] = "2048"
    v.vmx["numvcpus"] = "2"
    v.vmx["cpuid.coresPerSocket"] = "1"
    v.vmx["ethernet0.virtualDev"] = "vmxnet3"
    v.vmx["RemoteDisplay.vnc.enabled"] = "false"
    v.vmx["RemoteDisplay.vnc.port"] = "5900"
    v.vmx["scsi0.virtualDev"] = "lsisas1068"
  end

  config.vm.provision :ansible do |ansible|
    ansible.playbook = "vagrant.yml"
    ansible.limit = 'all'
    ansible.verbose = "vvvv"
  end
```

The most basic example playbook: save as `vagrant.yml` in the same directory as the above `Vagrantfile`.

```

---
- hosts: all
  vars:
    ansible_ssh_user: vagrant
    ansible_ssh_pass: vagrant
    ansible_ssh_port: 5985
    ansible_connection: winrm
  tasks:
    - name: Fetch SQL Server installer
      win_get_url:
        url: "http://download.microsoft.com/download/8/D/D/8DD7BDBA-CEF7-4D8E-8C16-D9F69527F909/ENU/x64/SQLEXPRWT_x64_ENU.exe"
[...]

PR #8665, opened 2014-08-18
WinRM test for long-running command.

cchurch:
pywinrm accepted my pull request (https://github.com/diyan/pywinrm/pull/25) to fix the issue when commands take longer than 60 seconds (https://github.com/ansible/ansible/issues/8206).  I've added a test to verify it works (when using the latest pywinrm).

https://github.com/ansible/ansible/issues/8206 should be fixed if users update to the latest pywinrm (`pip install -U http://github.com/diyan/pywinrm/archive/master.zip#egg=pywinrm`).


mpdehaan:
Merged, thanks!  Presumably this may wish to include a commen about what winrm version is required or something if folks run this on older versions, but I suspect many users wouldn't be running the tests in this case so that's only a minor thing.  Thanks very much!

PR #8673, opened 2014-08-19
Minor fixes to the locale_gen module.

0x44:
-   locale_gen module: make 'state' parameter optional.
  
  The 'state' parameter is optional in doc but required in actual implementation. As it has 'present' as it's default value, it should be safe to make it optional as most modules do.
-   locale_gen module: remove the dot check in locale names.
  
  In fact, most of the locales don't have dot in their names:
  
  ```
  $ lsb_release -ds
  Debian GNU/Linux testing (jessie)
  $ wc -l /usr/share/i18n/SUPPORTED
  471 /usr/share/i18n/SUPPORTED
  $ grep -v \\. /usr/share/i18n/SUPPORTED | wc -l
  308
  ```

### DIFF
--- a/library/system/locale_gen
+++ b/library/system/locale_gen
@@ -113,8 +113,6 @@ def main():
     )
 
     name = module.params['name']
-    if not "." in name:
-        module.fail_json(msg="Locale does not match pattern. Did you specify the encoding?")
     state = module.params['state']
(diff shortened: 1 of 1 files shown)


0x44:
See #8674.

PR #8679, opened 2014-08-19
Add support for Arch to module_utils.basic.py

### LINKED ISSUE #8653: hostname module: python platform.linux_distribution does not support Arch
#### Issue Type:

Bug
#### Ansible Version:

```
ansible 1.7.1
```
#### Environment:

```
From: Arch Linux
To: Arch Linux
```
#### Summary:

Ansible module `hostname` fails because python stdlib function `platform.linux_distribution` does not recognize Arch.
#### Steps To Reproduce:

Try to run hostname module on Arch virtualbox:

```
env/bin/ansible arch-vbox -m hostname -a name=foo
```
#### Expected Results:

```
arch-vbox | success >> {
    "changed": true, 
    "name": "foo"
}
```
#### Actual Results:

```
arch-vbox | FAILED >> {
    "failed": true, 
    "msg": "hostname module cannot be used on platform Linux (OtherLinux)"
}
```

j0057:
Fixes ansible/ansible#8653

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -181,7 +181,8 @@ def get_distribution():
     ''' return the distribution name '''
     if platform.system() == 'Linux':
         try:
-            distribution = platform.linux_distribution()[0].capitalize()
+            supported_dists = platform._supported_dists + ('arch',)
+            distribution = platform.linux_distribution(supported_dists=supported_dists)[0].capitalize()
             if not distribution and os.path.isfile('/etc/system-release'):
                 distribution = platform.linux_distribution(supported_dists=['system'])[0].capitalize()
                 if 'Amazon' in distribution:


rmorgans:
Hi - this seems to fail the CI build for python 2.6, however the failure appears to be unrelated to the functionality ([log](https://travis-ci.org/ansible/ansible/jobs/57313994#L388)) in  core/cloud/google/gc_storage.py and related to docstrings, rather than anything to do with hostnames for arch.

How would I go about getting this to run and getting this merged?

abadger:
I think the reason that it was failing in travis was that it was based on an old version, not the current version of the tree.  Merged it and it passed so that seems fine.

Sorry it took so long to get this applied.  The code looked a bit hacky but taking a look at how python-3.3 fixed it made it seem like the right way to go about it.  The one thing I hate about it is that we're doing the same sorts of contortions in two places -- the fact gathering has to be able to figure out that the system is arch and this module does as well.  And there may be other modules that do this as well :-(  Would be great if we could pull all of these pieces of code into one place... But that's a larger refactor than just this bug.

# Merged

Hi!

issue is resolved.

Thank you!

PR #8691, opened 2014-08-20
Release1.7.1

uwehanskruse:


jimi-c:
If there is a commit in here that you're trying to submit, please make sure that you're basing it off the development branch only, and do not include other commits. Thanks!

PR #8693, opened 2014-08-20
Make template's relative path available to ansible_managed

mmoya:
Given that `{file}` in _ansible_managed_ gets substituted to an absolute path, it can change in situations where an username is part of that path no matter if the template is the same file with respect to the playbook using it. For example _/home/foo/ansible-site/path/to/template.j2_ vs. _/home/bar/work/tests/ansible-site/path/to/template.j2_)

This changeset exports the template's relative path as `template_relpath` var and also substitutes `relfile` in _ansible_managed_.

### DIFF
--- a/lib/ansible/utils/template.py
+++ b/lib/ansible/utils/template.py
@@ -268,7 +268,8 @@ def template_from_file(basedir, path, vars, vault_password=None):
     managed_str = managed_default.format(
         host = vars['template_host'],
         uid  = vars['template_uid'],
-        file = vars['template_path']
+        file = vars['template_path'],
+        relfile = vars['template_relpath']
     )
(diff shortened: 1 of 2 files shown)


ernestob:
:+1:

amenonsen:
@mmoya if you plan to rebase this to apply to current devel, please also add a quick note to the documentation about the changes.

amenonsen:
The original case for the change still seems valid, but this PR hasn't been updated for months, so I'll close this now. Anyone who would find this feature useful is welcome to rebase and resubmit it (which can be done while retaining attribution to the original author, with additional fixup commits if necessary). If you do so, please note the request above for a note in the documentation.

Issue #8664, opened 2014-08-18
group_vars from group_by do not override group_vars/all since f6a55a3


direvus:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.8 (devel 6180d9a693) last updated 2014/08/18 17:03:49 (GMT +1100)
##### Environment:

Bug discovered with Debian x86_64, confirmed and bisected with Gentoo x86_64.
##### Summary:

When using group_by to create a group on the fly (e.g. `group_by: key="{{ ansible_os_family }}"`), the group_vars fail to override vars in `group_vars/all`.

I bisected this to f6a55a35525a6ca0b56ad856f2ab11bbd226c1b4.
##### Steps To Reproduce:

`site.yml`:

``` yaml

---
- hosts: all
  tasks:
  - group_by: key={{ ansible_os_family }}
  - fail: msg="testvar = {{ testvar }}"
    when: testvar != "override OK"
```

`group_vars/all`:

``` yml

---
testvar: override FAIL
```

`group_vars/Gentoo` (where "Gentoo" is whatever your `ansible_os_family` actually resolves to):

``` yaml

---
testvar: override OK
```
##### Expected Results:

```
PLAY [all] ******************************************************************** 

GATHERING FACTS *************************************************************** 
ok: [127.0.0.1]

TASK: [group_by key={{ ansible_os_family }}] ********************************** 
changed: [127.0.0.1]

TASK: [fail msg="testvar = {{ testvar }}"] ************************************ 
skipping: [127.0.0.1]

PLAY RECAP ******************************************************************** 
127.0.0.1                  : ok=2    changed=1    unreachable=0    failed=0   
```
##### Actual Results:

```
PLAY [all] ******************************************************************** 

GATHERING FACTS *************************************************************** 
ok: [127.0.0.1]

TASK: [group_by key={{ansible_os_family}}] ************************************ 
changed: [127.0.0.1]

TASK: [fail msg="testvar = {{testvar}}"] ************************************** 
failed: [127.0.0.1] => {"failed": true}
msg: testvar = override FAIL

FATAL: all hosts have already failed -- aborting

PLAY RECAP ******************************************************************** 
           to retry, use: --limit @/home/direvus/site.retry

127.0.0.1                  : ok=2    changed=1    unreachable=0    failed=1   

```

srgvg:
This looks the same as #8159 - I'll need to re-investigate.

jimi-c:
[...]

PR #8702, opened 2014-08-21
Add integration test for group_by variable precedence.

direvus:
Test that, when a group is created with group_by, variables from the
corresponding group_vars/ file are correctly pulled in, and override
variables specified in group_vars/all.

See ansible/ansible#8664.


mpdehaan:
Hi, thanks for this.

One problem I see is this says it is checking group_by, but I don't see a call to the group_by module in the output.  Is it possible that maybe there's a file that forgot to be checked in for this one?

Thanks!

mpdehaan:
Also I believe this is already fixed on the development branch, I'll let @jimi-c confirm.

The workaround would be to:

```
 - group_by: key=grp-{{ foo }}
 - include_vars: path/to/group_vars/grp-{{ foo }}.yml
```

direvus:
Hi @mpdehaan,

Not sure what you mean about it not running `group_by`.  The test_group_by.yml playbook runs group_by four times, on different variations of the variable `genus`, at the top of the file.  I left those tasks alone, except to make them run group_by on all hosts, not just those in the `lamani` group, so as to provide testing coverage on hosts that are initially ungrouped in inventory.  

All my new tests do, is check that once group_by is run, the hosts have properly picked up vars for the new group from `group_vars`, and that they properly override vars picked up from `group_vars/all`.

`test_group_by` seemed like the most natural place to put the new tests, but if you'd prefer I move them to `test_var_precedence` or somewhere else, let me know.

The bug I reported in ansible/ansible#8664 is indeed already fixed in devel.  The purpose of this PR is solely to add integration testing coverage, following @sergevanginderachter's suggestion.

Thanks for posting that workaround.

amenonsen:
@direvus any chance you could rebase this to apply to current devel? I think it's a useful test to have.

direvus:
Hi @amenonsen, I have created a new PR against current devel as requested, see #11752.  Unfortunately I couldn't just update the original PR, as that fork no longer exists.

PR #8706, opened 2014-08-21
Force pseudo-tty allocation on sync (untested)

Firefishy:
Force the allocation of a pseudo-tty so that sudo on the remote connection can run.
The sudoers on redhat 6+ and other distros have requiretty enabled by default, requiring the ssh to allocate a tty.

### DIFF
--- a/library/files/synchronize
+++ b/library/files/synchronize
@@ -303,7 +303,7 @@ def main():
     else:
         private_key = '-i '+ private_key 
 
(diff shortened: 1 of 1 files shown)


sivel:
Is this in fact untested?  If so, why would you submit a PR for untested code?

Firefishy:
Sorry, I'm being an idiot. Pull is broken and doesn't work. rsync on remote uses stdin and allocating a tty breaks it.

PR #8714, opened 2014-08-22
Allow roles to be installed from archives on github

willthames:
Ensure that the automated scm determination for github.com
repos still copes with .tar.gz archive files.

Handling .zip archives is left as an exercise for the interested reader

Removed some debug print statements
Corrected a FIXME


mpdehaan:
Merged in, thanks!

PR #8725, opened 2014-08-23
[get_url module]parse content-disposition filename parameter even if it's not quoted

quiver:
`get_url` module currently parses `content-disposition` response header field if `filename` parameter is double quoted.
But some servers do not quote `filename` parameter. One notable example is github's download server.

```
$ curl --head https://codeload.github.com/antirez/redis/tar.gz/3.0.0-beta8
HTTP/1.1 200 OK
Content-Length: 1327483
Access-Control-Allow-Origin: https://render.githubusercontent.com
Content-Security-Policy: default-src 'none'
X-XSS-Protection: 1; mode=block
X-Frame-Options: deny
X-Content-Type-Options: nosniff
Strict-Transport-Security: max-age=31536000
Vary: Authorization,Accept-Encoding
ETag: "f3b0cbd15c2e5afb1d309ed4de548d334ef85900"
Content-Type: application/x-gzip
Content-Disposition: attachment; filename=redis-3.0.0-beta8.tar.gz
Date: Sat, 23 Aug 2014 02:59:29 GMT
```

This PR tries to parse this parameter even if it's not double quoted.

For RFC specification wise, `Content-Disposition` has a lot more varieties(See RFC 6266).
But the spec is too complex and this simple PR should solve more than 99.99% cases.

OK pattern : Content-Disposition: attachment; filename="foo.tar.gz"
OK pattern : Content-Disposition: attachment; filename=foo.tar.gz <- previously NG
OK pattern : Content-Disposition: attachment;filename="foo.tar.gz"
OK pattern : Content-Disposition: attachment;filename=foo.tar.gz <- previously NG

### DIFF
--- a/library/network/get_url
+++ b/library/network/get_url
@@ -169,7 +169,7 @@ def extract_filename_from_headers(headers):
 
     Looks for the content-disposition header and applies a regex.
     Returns the filename if successful, else None."""
-    cont_disp_regex = 'attachment; ?filename="(.+)"'
+    cont_disp_regex = 'attachment; ?filename="?([^"]+)'
     res = None
 
     if 'content-disposition' in headers:


jimi-c:
@quiver the odd thing in looking at this, is that it doesn't look like the old regex should have worked? It's using a match() instead of a search() but isn't anchoring against anything in the string. When I was testing this using match(), neither regex returned a re object:

<pre>
>>> regex = r'attachment; ?filename="?([^"]+)'
>>> re.match(regex, r'Content-Disposition: attachment; filename="fname.ext"')
>>>
</pre>

But when using search(), it worked as expected:

<pre>
>>> re.search(regex, 'Content-Disposition: attachment; filename=foo').groups()
('foo',)
>>> re.search(regex, 'Content-Disposition: attachment; filename="foo"').groups()
('foo',)
</pre>

Should this be updated, or am I missing something?

quiver:
>  the odd thing in looking at this, is that it doesn't look like the old regex should have worked?

As the doc says https://docs.python.org/2/library/re.html#re.match, `match` matches "the beginning of string", while `search` searches the entire string.

```
>>> re.match(regex, 'Content-Disposition: attachment; filename="fname.ext"')
>>> re.match(regex, 'attachment; filename="fname.ext"')
<_sre.SRE_Match object at 0xb7255f20>
```

In the code

```
cont_disp = headers['content-disposition'] # <- 
```

cont_disp is `"attachment; filename=foo"` or something and not the entire header field like `Content-Disposition: attachment; filename="fname.ext"`, so the old code worked.
No need to update.

jimi-c:
Ahh, yeah that explains it, I was testing against what the full header would have been.

jimi-c:
# Merged

Hi!

Thank you!

PR #8676, opened 2014-08-19
Updated version of my original PR #7292 to incorporate nested group changes

nickpack:
My previous PR for the expanding of CSV values in ec2 tags did not merge cleanly with the current devel HEAD, so have updated the patches to resolve this.

PR #8726, opened 2014-08-23
Add support for list of tag values for ec2 plugin

codebynumbers:
Detects commas in ec2 tags and treats each one as a seperate tag group.
For example if you have tag key of "class" and value of "web, cron", server is is added to tag_class_web and tag_class_cron.


jimi-c:
# Possible Duplicated Efforts

Hi!

It looks like the work from this pull request is a duplicate of the following PR:
- #8676, which also adds the functionality to disable the splitting of tags for backwards-compatibility purposes.

PR #8743, opened 2014-08-25
Handling space in paths / Reversing order or role_path selection

y-nk:
I think the commit [1b4eb84](https://github.com/ynk/ansible/commit/1b4eb84b696e47ab0c66ac1ca19e63dc7301b2b5) will be very handy for mac users as i am. Thing is, if you have a space in your `given` path, then python will output the path within single quotes, which results in a false state of the `if given.startswith("/"):` test.

The second [298f52a](https://github.com/ynk/ansible/commit/298f52a4f839b06630783b1fd07d0440e02858c2) looks more pratical to me. It looked like the `DEFAULT_ROLE_PATH` constant was added to the end of the `possible_paths` list, resulting in a default-first/configuration-last order, which i think should be reversed, as configuration should be prior to default.

### DIFF
--- a/lib/ansible/playbook/play.py
+++ b/lib/ansible/playbook/play.py
@@ -191,16 +191,16 @@ class Play(object):
 
         role_path = None
 
-        possible_paths = [
-            utils.path_dwim(self.basedir, os.path.join('roles', role_name)),
-            utils.path_dwim(self.basedir, role_name)
-        ]
+        possible_paths = []
 
         if C.DEFAULT_ROLES_PATH:
             search_locations = C.DEFAULT_ROLES_PATH.split(os.pathsep)
             for loc in search_locations:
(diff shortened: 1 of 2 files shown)


jimi-c:
Hi @ynk, I understand the purpose behind the 1b4eb84 patch (though we do have a utility function in `module_utils/splitter.py` which may be more appropriate to use there), however I do not understand the purpose behind the other commit, and why that's specific to OSX/Mac users?

y-nk:
My bad, actually i should have done 2 pulls requests for such separate things. Let me clean that PR and make a 2nd.

jimi-c:
No problem, thanks!

y-nk:
Those have been reopened as [#8744](https://github.com/ansible/ansible/pull/8744) and [#8745](https://github.com/ansible/ansible/pull/8745)

PR #8745, opened 2014-08-25
fixing role path sorting in possible_paths list

y-nk:
I changed the order of `possible_paths`list. _Before_, **default** values were first appended then the `DEFAULT_ROLE_PATH` constant. That would cause the next for..in [line 205](https://github.com/ynk/ansible/compare/fix/role_path_order?expand=1#diff-30c8c86d294b2f7b782122bc2eb3e4aeR205) to choose first default, then user-defined role path (and break at first occurence). I think should be reversed, as configuration should be prior to default.

### DIFF
--- a/lib/ansible/playbook/play.py
+++ b/lib/ansible/playbook/play.py
@@ -191,16 +191,16 @@ class Play(object):
 
         role_path = None
 
-        possible_paths = [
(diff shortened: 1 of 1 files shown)


amenonsen:
This PR can be closed, the code has been completely reorganised in devel (and I believe the problem is also fixed, though I haven't tested).

bcoca:
closing as per above

PR #8749, opened 2014-08-25
Allow modules to "invoke" add_host

sivel:
Currently when cloud modules create new hosts, it is typical that people will run `add_host` afterwards to make each new host available to ansible during it's run time.

This pull request will enable the module to return a top level key called `add_host` that is a list of dictionaries.  The dictionaries returned will be used in populating hosts and groups just as the `add_host` module would.

Assuming this is accepted, modules such as `rax` and `ec2` could be updated to utilize this functionality.

Example module (called "foo" in test playbook below):

``` python
#!/usr/bin/python

import json

foo = {
    "add_host": [
        {
            "name": "foo",
            "groups": "bar",
            "baz": "qux",
        }
    ]
}

print json.dumps(foo, indent=4, sort_keys=True)
```

Test playbook:

```

---
- hosts: localhost
  connection: local
  gather_facts: false
  tasks:
    - action: foo
    - add_host: name=test groups=testgroup something=else
    - assert:
        that: groups.keys() == ['bar', 'all', 'testgroup']
    - assert:
        that: hostvars['foo'] is mapping
    - assert:
        that: hostvars['test'] is mapping
    - assert:
        that: hostvars['foo']['baz'] == 'qux'
    - assert:
        that: hostvars['test']['something'] == 'else'
```

Output (with -vv):

```

TASK: [foo] ******************************************************************* 
<localhost> REMOTE_MODULE foo
ok: [localhost] => {"add_host": [{"baz": "qux", "groups": "bar", "name": "foo"}]}
creating host via 'add_host': hostname=foo
added host to group via add_host module: bar

TASK: [add_host name=test groups=testgroup something=else] ******************** 
creating host via 'add_host': hostname=test
added host to group via add_host module: testgroup
ok: [localhost] => {"new_groups": ["testgroup"], "new_host": "test"}

TASK: [assert ] *************************************************************** 
ok: [localhost] => {"msg": "all assertions passed"}

TASK: [assert ] *************************************************************** 
ok: [localhost] => {"msg": "all assertions passed"}

TASK: [assert ] *************************************************************** 
ok: [localhost] => {"msg": "all assertions passed"}

TASK: [assert ] *************************************************************** 
ok: [localhost] => {"msg": "all assertions passed"}

TASK: [assert ] *************************************************************** 
ok: [localhost] => {"msg": "all assertions passed"}

localhost                  : ok=7    changed=0    unreachable=0    failed=0   
```

### DIFF
--- a/lib/ansible/inventory/__init__.py
+++ b/lib/ansible/inventory/__init__.py
@@ -30,6 +30,7 @@ from ansible.inventory.group import Group
 from ansible.inventory.host import Host
 from ansible import errors
 from ansible import utils
+from ansible.callbacks import vv
 
 class Inventory(object):
     """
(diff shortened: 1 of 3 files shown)


mpdehaan:
Discussed this was James and have some concerns about this one, namely that other modules may abuse it and the add_host idiom is somewhat in place.

Further, it would require changes and testing on all cloud types, which is somewhat aggressive, and may confuse playbooks from one version or another.

We would suggest instead we allow:
- add_host: result={{ registered_result_of_any_cloud_module }} group=group_name

This also avoids adding a new parameter or two to each cloud module.

In particular, I'm a bit concerned about a hacked module adding hosts to other groups, when the playbook had nothing to do with provisioning, resulting in ansible contacting other hosts it wasn't programmed to contact.  It's a bit "out there", but it's not something I'd want to happen.

To simplify invocation, though this can't change the returns of any of the cloud modules for compatibility reasons, it can be made and tested to work with all of them.

Rather than first being proposed by a code example, in the future, I would like to move more of these discussions to ansible-devel prior to implementation, to avoid possible extra work.   This would make it easier to talk about what we want any possible concerns.

Hope this is understandable, and if we can start a thread on ansible-devel that would be great.

Thanks!

sivel:
My apologies.  This was something that James Tanner had approached me about several months ago in which we had a number of conversations about implementation.  I had been working on this but never had the time to finish up or submit the PR while he was still working on the project.

I'll send out a -devel email later today.

mpdehaan:
Yeah Tanner should have moved that conversation up to ansible-devel, but all good.  Thanks!

sivel:
I'm going to close this out for now.  I may not be able to start working on it for a little while.  If things work out I will submit another PR.

PR #8750, opened 2014-08-25
mysql_user module: adding support for various REQUIRE options.

egorchakov:
This is meant to be a replacement for the incomplete and now closed PR #7744 .

This PR introduces 5 new parameters:
- require_ssl
- require_x509
- require_subject
- require_issuer
- require_cipher

More information about their purpose can be found [here](http://dev.mysql.com/doc/refman/5.1/en/grant.html).

Personally, I've only ever used 'REQUIRE SSL' in MySQL queries, but some might want more fine-grained control, hence the other 4 parameters.

### DIFF
--- a/library/database/mysql_user
+++ b/library/database/mysql_user
@@ -472,6 +472,10 @@ def main():
     require_subject = module.params["require_subject"]
     require_issuer = module.params["require_issuer"]
     require_cipher = module.params["require_cipher"]
+    requirements = requirements_unpack(module, require_ssl, require_x509,
+                                               require_subject, require_issuer,
+                                               require_cipher)
+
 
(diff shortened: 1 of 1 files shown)


jimi-c:
# Cleanup Request

Hi!

In general, we don't mind too much if a PR includes a few commits, however the large number of commits you have here makes this more difficult to review, and could represent major challenges if we had to revert the change.

To resolve this problem, it may be helpful to do the following:
- Create a new branch and use `git merge --squash` to squash all of these down to a single commit.
- You can then send us the PR from that branch instead.

We will definitely see your comments on this issue when reading this ticket, but may not be able to reply promptly.  You may also wish to join one of our two mailing lists which is very active:

PR #8754, opened 2014-08-26
Customizing groups_vars and host_vars directory structure

y-nk:
As i wanted to create a working directory that would fit my needs, i coded this small patch that allows 3 new vars in the `ansible.cfg` file, named `vars_path`, `group_vars_name` and `host_vars_name`.

I setup those regarding the current logic in `constants.py` with a default values so nothing would break if not customized. The actual customization i've been doing in my `ansible.cfg` file is : 

```
[defaults]
log_path          = ./logs/ansible.log

hostfile          = ./inventories/hosts
roles_path        = ./roles

vars_path         = ./vars
group_vars_name   = group
host_vars_name    = host
```

so my working directory would look like : 

```
.
├── README.md
├── ansible.cfg
├── inventories
│   └── hosts
├── logs
│   └── ansible.log
├── playbooks
│   └── default.yml
├── requirements.yml
├── roles
└── vars
    ├── group
    └── host
```

Mostly, as i think about playbooks as objects, this enhancement would be also effective for playbooks as simple folders having a similar _but lighter_ directory structure as roles :

```
└── playbooks
    └── myplaybook
        ├── main.yml
        └── vars
            ├── group
            └── host
```

### DIFF
--- a/lib/ansible/constants.py
+++ b/lib/ansible/constants.py
@@ -156,6 +156,10 @@ DEFAULT_VARS_PLUGIN_PATH       = get_config(p, DEFAULTS, 'vars_plugins',       '
 DEFAULT_FILTER_PLUGIN_PATH     = get_config(p, DEFAULTS, 'filter_plugins',     'ANSIBLE_FILTER_PLUGINS', '/usr/share/ansible_plugins/filter_plugins')
 DEFAULT_LOG_PATH               = shell_expand_path(get_config(p, DEFAULTS, 'log_path',           'ANSIBLE_LOG_PATH', ''))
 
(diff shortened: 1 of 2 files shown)


jimi-c:
# Question About This One

Hi!

Thanks very much for your submission to Ansible.  It sincerely means a lot to us that you've taken time to contribute. Unfortunately, we're not sure if this is a feature that would benefit many users. We don't want to seem confrontational, and we're absolutely always up for discussion.

Since this is a really busy project, we don't always see comments on closed tickets, but want to encourage open dialog.  You can stop by the development list, and we'd be glad to talk about it - and we might even be persuaded otherwise!

PR #8773, opened 2014-08-27
Always return a list even if its empty.

emil-palm:
This is not more a fix then it is a standardization of the returns from the module.
All other cases it will return a list except when it doesnt add anything then it returns None.

This messes up a few things when using: "with_subelements".


jimi-c:
# Merged

Hi!

Thank you!

PR #8777, opened 2014-08-27
Adds percent filter.

jsmartin:
```
- hosts: localhost
  tasks:
  - debug: msg="{{ 99 |percent(15) }}"
```

```
TASK: [debug msg="14.85"] *****************************************************
ok: [localhost] => {
    "msg": "14.85"
}
```

### DIFF
--- a/docsite/rst/playbooks_variables.rst
+++ b/docsite/rst/playbooks_variables.rst
@@ -277,6 +277,26 @@ Get a random number from 1 to 100 but in steps of 10::
     {{ 100 |random(1, 10) }}    => 31
     {{ 100 |random(start=1, step=10) }}    => 51
 
+.. _percent_filter:
+
+Perecent Filter
+--------------------
+
+.. versionadded:: 1.8
+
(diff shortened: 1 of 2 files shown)


mpdehaan:
When adding a new filter, it's needed to also add documentation about the new filter and the version_added information for the filters section.  Can you add this?

Thanks!

jsmartin:
Added docs.

sivel:
Slightly curious as to why a filter is actually needed here.

Something like the following requires no additional code added to ansible and is actually shorter to type than invoking a filter:

```
- debug: msg="{{ 99 * 0.15 }}"
```

The above produces the same results.

jsmartin:
Absolutely, that works as well.  Maybe the percent filter is redundant.  Probably easier to find in a search, but we could just add some docs show the above example.

jimi-c:
Based on the above discussion I'm going to go ahead and close this. @jsmartin if you want, you can send a docs PR to add any documentation you feel would be helpful. Thanks!

PR #8805, opened 2014-08-28
docker: add pull_latest option to pull latest image when creating container

discordianfish:
This option will cause ansible to pull the latest image before running a new container.

### DIFF
--- a/library/cloud/docker
+++ b/library/cloud/docker
@@ -204,6 +204,12 @@ options:
     default: ''
     aliases: []
     version_added: "1.8"
+  pull_latest:
+    description:
+      - Pull the latest image before running a container
+    required: false
+    default: false
+    aliases: []
 
 author: Cove Schneider, Joshua Conner, Pavel Antonov
 requirements: [ "docker-py >= 0.3.0", "docker >= 0.10.0" ]
@@ -610,11 +616,7 @@ class DockerManager:
 
             return results
 
-        try:
-            containers = do_create(count, params)
-        except:
-            resource = self.module.params.get('image')
-            image, tag = self.get_split_image_tag(resource)
+        def do_pull(resource, tag):
             if self.module.params.get('username'):
                 try:
                     self.client.login(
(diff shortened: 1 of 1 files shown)


iragsdale:
I have a question about the intended behavior of this module when updating to the latest version of an image. I would probably consider this a bug, but perhaps others may not.

If you pull the latest version of a given named image, the docker image that name references will change, but any existing containers launched using that name are still using the old image. If you do a "docker inspect" on those containers, they still list the image name in Config->Image, but the top level Image property correctly references the original Image ID.

The existing code compares the name given in the Ansible configuration to the Config->Image property. So, if that image is updated, the Docker module will fail to relaunch the instance with the newly updated image, because the name has not changed. If it were to instead look up the ID that image name refers to and compare it to the top-level Image property, it would correctly recreate the container.

I think that behavior is likely to be very confusing - you'll have a newly downloaded image, but no running containers will have restarted to use it. This will happen whether the image is updated via the module or updated manually. I think this COULD be considered a Docker bug, but I think they would likely call that intended behavior.

I believe the correct thing to do would be to look up the current ID for a given image name and compare that to the top-level Image property for the container when deciding whether to recreate the container.

discordianfish:
@iragsdale Maybe I missed something, but this is not suppose to touch a running container. It will only make sure to pull the latest image _before_ creating/running a new container. If you want to upgrade your container, you still need to remove it first.

iragsdale:
Hmm, the current code _does_ recreate the container _if_ the image it was started with is named differently than the image specified in Ansible. Seems to me that it would be confusing if it didn't do that when the tag changed - otherwise you could intend to have a set of containers running the same image but actually running different images.

Perhaps it would make sense to have a separate flag that enables/disables the restart behavior? I'm really not sure what makes the most sense semantically, and maybe this is a feature request and not a bug.

iragsdale:
Either way the issue is not really in your PR, just potentially exposed by it. I should probably open a new issue with Ansible about the issue. :)

iragsdale:
New issue opened - I'd consider this to be at least somewhat related: #8905.

mpdehaan:
# Module Repo Information

Hi!

Thanks you very much!

PR #8808, opened 2014-08-28
Added include_defaults

dvarrazzo:
Please find my first iteration at providing a command "include_defaults".

include_defaults works like include_vars but gives the variables a lower priority such that their value can be overridden in other places, such as the inventory or a group_vars file.

The feature is tested in test_var_blending and is verified that the imported variables are known to the play and the value gets correctly overridden.

### DIFF
--- a/lib/ansible/runner/action_plugins/include_defaults.py
+++ b/lib/ansible/runner/action_plugins/include_defaults.py
@@ -1,0 +1,62 @@
+# (c) 2014 Daniele Varrazzo <daniele.varrazzo@gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+# Why is this not working?
+# import ansible.runner.action_plugins.include_vars as iv
+
+# class ActionModule(iv.ActionModule):
+#     _target = 'defaults'
+
+import os
+from ansible.utils import template
+from ansible import utils
+from ansible import errors
+from ansible.runner.return_data import ReturnData
+
+class ActionModule(object):
+
+    TRANSFERS_FILES = False
+
+    def __init__(self, runner):
+        self.runner = runner
+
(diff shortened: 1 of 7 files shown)


mpdehaan:
Hi!

- we've discussed this topic on ansible-devel and cannot pin down a use case where this can't be modelled more idiomatically through other ansible-means
- we believe introducing extra syntax for this feature would add to complexity in learning the application that we would like to solve through more idiomatic means

dvarrazzo:
No problem Michael. As stated in the forum I'm happy the feature can be used as an external plugin: it would allow me and other people to test with the idea and see if it's valid.

For reference here is the [include defaults plugin](https://gist.github.com/dvarrazzo/7418a89b7278ff69267c) implementation.

ThomasSteinbach:
Great, this action plugin is exactly doing for what I am struggeling since hours. I wrote me now an Ansible role, patching the local Ansible installation with the 'load_defaults' plugin. Nevertheless it would be nice to see this plugin as a core module.

**What is my use-case?**

I wrote a couple of roles, each differing in behavior based on a variable called 'install_type'. This variable could carry the values 'server' and 'client'. Depending on this variable I load in my tasks/main.yml either the tasks/server.yml or the tasks/client.yml. Those files again should load their corresponding vars- and default-files. The 'include_var' module is already part of Ansible, the 'include_defaults' not. Different defaults for client and server makes sense, if you think about log settings or hardware priority settings.

First I wrote all default settings in one defaults/main.yml and tried to differ between client and server settings with Jinja2-If statements. But Jinja2 templating does not work within variable files (vars, defaults, group_vars, vault file).

Loading specific variable files is now my solution, thanks to the plugin.

**Why would I see the plugin as core module?**
1. Because using `include_defaults` is obvious when there is a `include_vars`.
2. Because there is a file called _main_.yml inside the _defaults_ folder, there must be a right to exist for other *.yml files inside that folder, that work exactly the same like the main defaults file.
3. There is no other way to define a larger group of defaults in subject to a selection. Using 'set_facts' for this purpose is just an ugly looking workaround.

I hope there will be a second chance for this pull request :smiley:

dvarrazzo:
@ThomasSteinbach unfortunately this implementation of `include_defaults` has an issue: because it changes some data structures in-place it doesn't work when ansible runs in parallel on many hosts, because the process forks and the modified variables get lost.

In my case, I have solved the issue by adding a `vars_files` list to the `group_vars` files: the list usually contains the name of the file in `vars_files` itself but it can contain further customization (can create configurations "inheriting" from a base one and tweaking a few things). I've then converted every task file where I was using `include_defaults` into `include_vars`, with a final include of `vars_files` too, in order to have the values specified in `vars_files` to override the "defaults" (which become too powerful otherwise). E.g.
- file `roles/web_server/defaults/main.yml`:

```
http_port: 80
```
- file `group_files/testsite`:

```
http_port: 8080
vars_files:
  - group_files/testsite
```
- file `roles/web_client/tasks/main.yml`:

```
- include_vars: "{{item}}"
  with_items:
    - roles/web_server/defaults/main.yml
    - {{vars_files}}
```

PR #8824, opened 2014-08-30
Add ansible_powershell_version and ansible_winrm_certificate_expires to windows facts

jhawkesworth:
This changes the major revision number of powershell on the windows host, and the date and time of expiry of the self-signed cert created when the ConfigureRemotingForAnsible.ps1 script is run.

### DIFF
--- a/library/windows/setup.ps1
+++ b/library/windows/setup.ps1
@@ -66,4 +66,11 @@ $ips = @()
 Foreach ($ip in $netcfg.IPAddress) { If ($ip) { $ips += $ip } }
 Set-Attr $result.ansible_facts "ansible_ip_addresses" $ips
 
+$psversion = $PSVersionTable.PSVersion.Major
+$winrm_cert_expiry = Get-ChildItem -Path Cert:\LocalMachine\My | where Subject -EQ "CN=$env:COMPUTERNAME" | select NotAfter
+
+Set-Attr $result.ansible_facts "ansible_powershell_version" $psversion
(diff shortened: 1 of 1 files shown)


cchurch:
While we'd recommend connecting using HTTPS, if I do connect to a server using HTTP instead, I get the following error via stderr, and `ansible_winrm_certificate_expires` is not set in the result.

> You cannot call a method on a null-valued expression. At C:\Users\Administrator\AppData\Local\Temp\ansible-tmp-1409430220.31-17616293 7321427\setup.ps1:216 char:1 + Set-Attr $result.ansible_facts "ansible_winrm_certificate_expires" $winrm_cert_e ... + ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ~~~ + CategoryInfo : InvalidOperation: (:) [], RuntimeException + FullyQualifiedErrorId : InvokeMethodOnNull

I'd recommend only setting `ansible_winrm_certificate_expires` if `$winrm_cert_expiry` is not null.

jhawkesworth:
Very good point, I'll fix it up and create a new PR

trondhindenes:
@jhawkesworth you should look at the cert thumbprint in the wsman listener, and then use that to enumerate the correct "ansible winrm certificate" (if any). I wouldn't just assume that the first cert macthing computername is the right one.

jhawkesworth:
Thanks, @trondhindenes you are right, I will create a new PR which will compare with the thumbprint from wsman.

PR #8827, opened 2014-08-31
TTL must be provided with deletion request

yarons:
I spent hours trying to figure what's wrong with my request, then I added the TTL value from the recorded "get" data and it finally worked, must change in all official documentation, really frustrating.

### DIFF
--- a/library/cloud/route53
+++ b/library/cloud/route53
@@ -113,6 +113,7 @@ EXAMPLES = '''
       record={{ rec.set.record }}
       type={{ rec.set.type }}
       value={{ rec.set.value }}
+      ttl={{ rec.set.value }}
 
 # Add an AAAA record.  Note that because there are colons in the value
 # that the entire parameter list must be quoted:


jimi-c:
Is this the case with `get` requests too? If so, the ttl parameter should be marked as required=True, otherwise a check should be made if the command is create or delete and a failure should be issued if the TTL is not specified.

yarons:
Nope, let me describe the process, BTW don't close the other issue as it's fixing another part of the text.

The problem is this:
You can't make a "delete" request without specifying the TTL, the manual says: not required which is true for get but false for "create" and "delete".

jimi-c:
@yarons regarding the other PR, just include that here so they can be merged in together rather than piecemeal.

So for the TTL being required, there should also be a check added such that if the command is specified, the ttl must also be set otherwise the task fails. Right now, we have this in the code (line 204 of `library/cloud/route53`):

<pre>
    if command_in == 'create' or command_in == 'delete':
        if not value_in:
            module.fail_json(msg = "parameter 'value' required for create/delete")
</pre>

A second check should be added as follows:

<pre>
    if command_in == 'create' or command_in == 'delete':
        if not value_in:
            module.fail_json(msg = "parameter 'value' required for create/delete")
        if not ttl_in:
            module.fail_json(msg = "parameter 'ttl' required for create/delete")
</pre>

yarons:
That's correct, I think it should be changed as well, so far I got a vague message about missing values each time I tried removing a DNS record without specifying the TTL.

I can't tell why though, if you think about it then the other R53 tools doesn't require such parameter, maybe we should conduct a more thorough test for the necessity of this value and why the request fails without it.

Not sure how to move the changes here, if we'll figure out a way of action I'll just create another pull request with both adjustments (surely they'll differ as you suggested).

jimi-c:
@yarons use `git cherry-pick <commit sha>` to move any commits over to the feature branch you used for this pull request, and then use `git push --force` to push them to github. GitHub will update this PR with the new version.

yarons:
@jimi-c I'm not using git, I was using the visual editor, this is why the commits are separated.

jimi-c:
@yarons In that case don't worry about it, I will correct this. Thanks!

yarons:
Awesome! Thanks!

mpdehaan:
Hi!

Thanks you very much!

PR #8829, opened 2014-08-31
Add slack callback plugin

quiver:
This is a callback plugin to post messages to Slack(https://slack.com/)

Due to service similarities, most codes are copied from `plugins/callbacks/hipchat.py` and my addition
modification is API handling.

As commented in the code, two environment variables(`SLACK_TOKEN` and `SLACK_DOMAIN`) are required.

### DIFF
--- a/plugins/callbacks/slack.py
+++ b/plugins/callbacks/slack.py
@@ -1,0 +1,207 @@
+# (C) 2014, George Yoshida <dynkin+dev@gmail.com>
+
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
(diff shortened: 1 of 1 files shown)


jimi-c:
This looks really good, the only thing I would potentially change is the direct use of urllib2. We wrote our fetch_url helper (in module_utils/urls.py) to get around some deficiencies in python 2.x's handling of SSL certs, so it might be better to use that instead. If you'd like to work on that, we can get this in queue for testing later, otherwise we can modify it when we merge it in.

quiver:
> We wrote our fetch_url helper (in module_utils/urls.py) to get around some deficiencies in python 2.x's handling of SSL certs

I looked at this code and have one question. Isn't `module_utils/urls.py` designed to be used inside modules?
For example, when you start a playbook, `self.callbacks.on_play_start(play.name)` is called from `ansible/playbook/__init__.py` and only the playbook name is passed to the callback handler.

But `module_utils/urls.py` function requires module object as its first argument and there's no way to pass module info from playbook_on_play_start callback handler.

Am i missing something?

jimi-c:
Ahh you're correct, that won't work. We have other code that is also in module_utils/ which is generic enough to be used both in core and modules (splitter.py, for example). So yeah, don't worry about using fetch_url for this then, as I forgot it's tied too closely to modules at the moment (a lot of it was moved from get_url to that snippet so it could be shared by other modules).

Sorry for the confusion.

amenonsen:
Probably should be closed in favour of #8955 now?

gregdek:
Yep, closing in favor of #8955. Thanks @quiver for submitting this one, and sorry it took so long to come to a decision on it.

PR #8830, opened 2014-08-31
Modifications to ignored files in inventory directories

quentinsf:
When specifying a directory as an inventory, README\* files are added to the list of files ignored by default.

In addition, a directory can include an '.inventoryignore' file (analagous to .gitignore) which lists patterns of files to be ignored (in place of the defaults).

Hope this is useful!


acm1:
+1

aioue:
+1 just hit this issue with an inventory script that has a module folder @quentinsf considered rebasing?

bcoca:
so we like the idea, but would prefer to make this a global config override from ansible.cfg, putting into the directory can add some complexity specially with nested directories that require different handling and don't specify an ignore file.

quentinsf:
OK Brian - that makes sense... I'll have to do some more exploration to see how that would be done.
If anyone gets there before me - feel free!

Enjoyed your NYC talk video, BTW :-)

bcoca:
in the same constants.py file you updated, just use something similar to:
`DEFAULT_SELINUX_SPECIAL_FS = get_config(p, 'selinux', 'special_context_filesystems', None, 'fuse, nfs, vboxsf', islist=True)`
which will allo for setting it in ansible.cfg or a env var and allow you to set the default setting

quentinsf:
Hello all - I've re-done this as a config option, but the rebasing became messy.  I'm going to delete my fork and recreate a clean new pull request.

quentinsf:
Moved to #11642.

PR #8833, opened 2014-08-31
Revised additions to windows facts

jhawkesworth:
This PR adds 2 new facts to windows hosts.
Powershell version (major number only)
If found, the expiry date of the certificate used by WinRM when the ConfigureRemotingForAnsible.ps1 script is run.
Added a null check as suggested by Chris Church to handle http connections (or certs not found).  It is perfectly valid to use non-self signed certs.  The code looks for certs in the local system and filters based on the hostname (which should match due to the need for hostnames to be correct when using https).


cchurch:
# Merged

Hi!

issue is resolved.

Thank you!

PR #8838, opened 2014-09-01
Include cloud loadbalancers in rax inventory

grahamrhay:


sivel:
I have a few concerns here, the first and biggest being that I don't believe load balancers should be part of inventory.

You cannot log into them and the rax_clb module offers an appropriate way of selecting a LB by both name or ID.

Since the plugins directory is essentially a collection of examples, I would recommend that if you need this functionality that you should maintain a fork that suits your needs in this regards.

grahamrhay:
I need the id to use with the rax_clb_node module, is there a better way of getting it?

grahamrhay:
I'm happy to delete the PR, and keep the changes in our codebase, just thought it might be useful for others.

sivel:
The better way of getting it would be to have a rax_clb task with a register before you use rax_clb_nodes.

Not only will this ensure the configuration of your load balancer, but the registered variable will also contain all of the information about the load balancer that can be used in later tasks.

An example of this can be found at:

http://docs.ansible.com/guide_rax.html#complete-environment

grahamrhay:
interesting, thanks

PR #8841, opened 2014-09-01
zabbix_maintenance improvements: env vars option, default state to present

resmo:
I made 2 commits which helped me with the daily work with this module. As you may know, the module is new in 1.8 and I started to use it. The first commit defaults the `state` option to `present` which is the default for many (all?) modules I saw having this argument. 

The second change helps me saving time by providing the login infos once in environment variables, but you can still provide them with args like before the change, or mix them. 

Regarding in the future, I also wanted to have more zabbix modules, e.g. zabbix_host for adding new hosts,  which can make use of those variables too.

### DIFF
--- a/library/monitoring/zabbix_maintenance
+++ b/library/monitoring/zabbix_maintenance
@@ -33,8 +33,8 @@ options:
     state:
         description:
             - Create or remove a maintenance window.
-        required: true
-        default: null
+        required: false
+        default: present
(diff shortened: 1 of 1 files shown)


resmo:
PR moved to:
- https://github.com/ansible/ansible-modules-extras/pull/5
- https://github.com/ansible/ansible-modules-extras/pull/4

PR #8862, opened 2014-09-02
zabbix_group: new module for adding and removing host groups to Zabbix

resmo:
Zabbix module for managing the host groups in Zabbix. This module makes sense together with the `zabbix_host`module.


resmo:
PR moved to https://github.com/ansible/ansible-modules-extras/pull/6

PR #8866, opened 2014-09-03
Document Environment in tasks as well

Fresheyeball:
Googling around for environment variables in Ansible yields this page. It would be useful to have the task version documented here as well.


Fresheyeball:
Good catch. Updated

PR #8870, opened 2014-09-03
Add support for multiple organizations in spacewalk.

photoninger:
I added some code to limit the inventory only to one organization within a spacewalk installation. The option -o takes the organization number. 
Also added the -p option to prefix the group names with the organization id, in case you use the script without -o and there are groups with the same name in more than one organization. Otherwise the groups would be merged to one big group in the inventory.

I hope you like the changes.

### DIFF
--- a/contrib/inventory/spacewalk.ini
+++ b/contrib/inventory/spacewalk.ini
@@ -1,0 +1,16 @@
+# Put this ini-file in the same directory as spacewalk.py
+# Command line options have precedence over options defined in here.
+
+[spacewalk]
+# To limit the script on one organization in spacewalk, uncomment org_number
+# and fill in the organization ID:
+# org_number=2
+
+# To prefix the group names with the organization ID set prefix_org_name=true.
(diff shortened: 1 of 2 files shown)


amenonsen:
@photoninger I believe the general policy is to not add options other than `--list` and `--host` because there's no way to invoke these through the Ansible inventory API. Instead, the best practice is to add them as options to the corresponding `.ini` file. Any chance you could rework your changes along these lines and refresh this PR with git push -f? It would also be good to have confirmation that your changes still work, since I'm not in a position to test.

photoninger:
@amenonsen You are right. I will try to add an .ini file, but I will also keep the extra command line options. IMHO they are still usefull in combination with "-H --list" to create a static inventory file with the script. Which is my current use case ;).

photoninger:
@amenonsen I added support for a ini-file and I had some fun implementing the _meta dictionary for hostvars in the json output for --list.

Issue #8872, opened 2014-09-03
FAQ recommends typing plaintext passwords into shell


RichardBarrell:
`docsite/rst/faq.rst` has a line recommending running the following command to generate password hashes using passlib: `python -c "from passlib.hash import sha512_crypt; print sha512_crypt.encrypt('<password>')"`

If you type that into a terminal, you'll almost certainly end up with the plaintext of your password in `~/.bash_history` (or fish or zsh's equivalents).

It also doesn't work without caveats if anyone wants to use a password which has characters in it such as '\'; you'd have to escape these characters from the shell, and then a second time for the string literal.

ansibot:
# Can You Help Us Out?

Thanks for filing a ticket! I am the friendly GitHub Ansibot. 

It looks like you might not have filled out the issue description based on our standard issue template.  You might not have known about that, and that's ok too, we'll tell you how to do it.

We have a standard template because Ansible is a really busy project and it helps to have some standard information in each ticket, and GitHub doesn't yet provide a standard facility to do this like some other bug trackers.  We hope you understand as this is really valuable to us!. 

Solving this is simple: please copy the contents of this [template](https://raw2.github.com/jctanner/issuetests/master/ISSUE_TEMPLATE.md) and **paste it into the description** of your ticket.  That's it!

# If You Had A Question To Ask Instead

If you happened to have a "how do I do this in Ansible" type of question, that's probably more of a user-list question than a bug report, and you should probably ask this question on the project mailing list instead.  

However, if you think you have a bug, the report is the way to go!  We definitely want all the bugs filed :)  Just trying to help!

# About Priority Tags

Since you're here, we'll also share some useful information at this time.

In general tickets will be assigned a priority between P1 (highest) and P5, and then worked in priority order. We may also have some follow up questions along the way, so keeping up with follow up comments via GitHub notifications is a good idea.

Due to large interest in Ansible, humans may not comment on your ticket immediately.  

# Mailing Lists

If you have concerns or questions, you're welcome to stop by the ansible-project or ansible-development mailing lists, as appropriate.  Here are the links:
- https://groups.google.com/forum/#!forum/ansible-project - for discussion of bugs and how-to type questions
- https://groups.google.com/forum/#!forum/ansible-devel - for discussion on how to implement a code change, or feature brainstorming among developers

Thanks again for the interest in Ansible!

RichardBarrell:
Pull request here: https://github.com/ansible/ansible/pull/8873

[...]

PR #8873, opened 2014-09-03
Issue 8872: use getpass.getpass() to read password.

RichardBarrell:
This avoids passing the plaintext raw password into the script as a
command-line argument, which shells such as bash and zsh will
typically log into a file like ~/bash_history.

As a small bonus, this works a LOT more smoothly if the user's
password has characters in it that either the shell or Python syntax
might like to interpret as metacharacters, like '\' or '"'.

getpass.getpass() is part of Python's standard library in all versions
between 2.4 and 3.3 at least.

### DIFF
--- a/docsite/rst/faq.rst
+++ b/docsite/rst/faq.rst
@@ -217,7 +217,7 @@ password hashing library is installed.
 
 Once the library is ready, SHA512 password values can then be generated as follows::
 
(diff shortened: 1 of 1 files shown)


RichardBarrell:
Issue: #8872

mpdehaan:
Merged, thanks!... this will appear in our next docs rebuild/push.

PR #8878, opened 2014-09-03
Add environment to plays and roles.

jhriggs:
(See https://groups.google.com/d/msg/ansible-project/w3aqRqZ1hIw/SCMV3qwaE-wJ). Play-level environment is used for every task in the play unless task's environment is set. Task environment overrides play environment (i.e. they are not merged).

Future thoughts:
- Merge play + task environment rather than override or make it optional?
- Add environment settings to roles.

### DIFF
--- a/lib/ansible/playbook/play.py
+++ b/lib/ansible/playbook/play.py
@@ -38,7 +38,7 @@ class Play(object):
        'accelerate_port', 'accelerate_ipv6', 'sudo', 'sudo_user', 'transport', 'playbook',
        'tags', 'gather_facts', 'serial', '_ds', '_handlers', '_tasks',
        'basedir', 'any_errors_fatal', 'roles', 'max_fail_pct', '_play_hosts', 'su', 'su_user',
-       'vault_password', 'no_log',
+       'vault_password', 'no_log', 'environment',
     ]
 
     # to catch typos and so forth -- these are userland names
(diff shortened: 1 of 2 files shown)


jimi-c:
Yes, the environments should be merged rather than overriding, otherwise this could lead to some confusing situations. In task.py, I believe it should be:

<pre>
self.environment  = ds.get('environment', {}).update(play.environment)
</pre>

I think a straight update is the correct way to go (as opposed to a dictionary merge), as the task would be an override.

jimi-c:
For safety, it'd probably be best NOT to do it as I show above (just in case someone specifies an environment that is not a dictionary, which would lead to a traceback).

jhriggs:
My concern with merging/updating play + task is that the user then has no way to unset or not use the play's environment, if needed for a particular task. Or, do you think that is an exception case not really worth worrying about?

WRT "NOT to do it as I show above", are you just saying that some sanity checks need to be done on the values first to ensure that they are dictionaries?

jhriggs:
See if this latest commit addresses your concerns/recommendations.

jhriggs:
Any thoughts on this one, @jimi-c?

bcoca:
sorry, an alternate implementation was already merged, plays already support environment in current devel

jhriggs:
@bcoca: What about environment at the role level which is also part of this PR? Does that exist also, or should I open a new PR for role environment?

bcoca:
@jhriggs i missed that, yes a new PR would be required

PR #8885, opened 2014-09-04
added postgresql_lang

jensdepuydt:
Since I regurarly was missing this type of functionality and was always using an ugly workaround, I decided to create a module to add, remove or change Postgresql procedural languages to db's myself.

Tested the module with all options.


jensdepuydt:
Replaced with new pull request: add postgresql_lang #8931
(had some issues with branching on GitHub)

PR #8900, opened 2014-09-05
Don't throw away useful stdout and stderr. Fixes #8418

### LINKED ISSUE #8418: _remote_md5 fails when passing -K but no sudo password is required.
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.7 (devel b9a5242f03) last updated 2014/08/04 09:22:22 (GMT +1000)
##### Environment:

N/A
##### Summary:

When running e.g. the unarchive module in an ansible-playbook in sudo mode against a server and passing -K (ask sudo password) to a server that doesn't require a password to sudo, it fails with the message `dest '/target/directory'  must be an existing dir`. 

The source of this message the unarchive action_plugin, and that plugin believes that the directory does not exist because the `runner._remote_md5` call returns `INVALIDMD5SUM`. 

It would be good to get the error message improved. 
##### Steps To Reproduce:

Have a server with sudo (-u user) -i not require a password. 
Use unarchive to unarchive against the target server as the sudo user

```
  hosts: target

  tasks:
  - get_url: url="https://github.com/ansible/ansible-examples/archive/master.zip" dest=/var/tmp/ansible-examples.zip

  - unarchive: src="/var/tmp/ansible-examples.zip" dest="/var/tmp/" copy=no
```

Then run that playbook with the -K option and enter a password. 
##### Expected Results:

A message along the lines of 'You sent a password to a system that didn't ask for one'
Or ignoring the password if it is not required. 
##### Actual Results:

Failure with the message `dest '/target/directory' must be an existing dir`

sivel:
When supplying a sudo password to a server that uses passwordless sudo,
we should not throw away useful stdout and stderr. This is particularly
important for modules that perform md5 checks as part of the pre module
execution.

I have tested these changes with several playbooks and adhoc tasks and have not encountered any negative side effects.

See #8418

### DIFF
--- a/lib/ansible/runner/connection_plugins/ssh.py
+++ b/lib/ansible/runner/connection_plugins/ssh.py
@@ -301,6 +301,8 @@ class Connection(object):
 
         self._send_password()
 
+        no_prompt_out = ''
+        no_prompt_err = ''
         if (self.runner.sudo and sudoable and self.runner.sudo_pass) or \
                 (self.runner.su and su and self.runner.su_pass):
             # several cases are handled for sudo privileges with password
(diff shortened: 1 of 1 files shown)


jimi-c:
#9014 includes and builds upon this patch, so noting here that if it is accepted this PR can simply be closed.

abadger:
#9014 which includes these changes have been merged.

Issue #4265, opened 2013-09-26
PDF documentation


rfleschenberg:
It would be nice to have a pre-built PDF of the docs. Building them yourself with Sphinx is somewhat cumbersome.

mpdehaan:
This is not a priority for us at this time as it's important for people to see what is new and available and we can share news this way.

I would suggest slurping down the website or using offline browser modes, but ansible-doc may be an even better solution.

php-coder:
I agree, that is was a good idea to publish generated documentation in PDF format.

cooniur:
PDF documentation is really a good idea, especially when you don't have internet connection, for example, on a commute train. Please make one!

felipythondev:
It would be really nice, I would like to read about it on train going to work.

kikicarbonell:
+1

jeremypruitt:
+1

mpasternak:
I'd gladly grab ANY form of offline docs. I've built myself, with Sphinx, only to find out, that search function still redirects to docs.ansible.com . PDFs would be great.  :+1:

lqueryvg:
+1

lqueryvg:
A very quick and dirty DIY solution is as follows (I'm on CentOS):
- `yum install pandoc pandoc-pdf`
- `git clone` ansible from github
- `cd ansible/docsite/rst`
- `for file in *.rst; do pandoc -f rst -o ${file%.rst}.pdf $file; done`

You now have lots of pdf files, one for each rst. You could read them offline like this, or concatenate them all into one pdf with `pdfunite *.pdf ansible_docsite.pdf`.

Of course, the ordering is a bit messed up (you could spend some time fixing this by specifying the exact order you want with the `pdfunite` command line), but it's good enough for me to be able to read on the train!

mpasternak:
Hi,

I'm on Mac OS X. I found this wonderful application for offline docs, with great integration with system and my IDE (PyCharm). Yes, this sounds like spam. No, I'm no way affiliated or endorsed by authors. It's called "Dash".

xenithorb:
https://github.com/bonndan/ansible-userguide-pdf

pythononwheels:
I know that this is old but I looked for myself for a way to carry the docs around and found no pdf.
But I think this one is a really simple workaround: (You need python 3.x)

**Prereqs:**
1. get the ansible repo: `git clone https://github.com/ansible/ansible.git`
2. install sphinx:  `pip install sphinx`

[...]

PR #8906, opened 2014-09-06
Add support for generating a PDF of the documentation

sivel:
In every project I have ever worked in people always seem to want PDF versions of the docs (#4265).

This PR adds support for building a PDF via sphinx.  `make webdocs` from the root will build both html and pdf.  `make docs` from the docsite dir will build both.  To build just the PDF `make pdf` from docsite.

This utilizes `rst2pdf` which provides PDF support for sphinx.

The PDF is built as `docsite/pdfout/ansible.pdf`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ docsite/_static/*.png
 docsite/_static/websupport.js
 docsite/searchindex.js
 docsite/htmlout
+docsite/pdfout
 # deb building stuff...
 debian/
 deb-build
--- a/docsite/Makefile
+++ b/docsite/Makefile
@@ -10,6 +10,9 @@ docs: clean modules staticmin
 	-(cp *.jpg htmlout/)
 	-(cp *.png htmlout/)
 
+pdf:
+	./build-site.py pdf
+
 variables:
 	(mkdir -p htmlout/)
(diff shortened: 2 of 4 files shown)


mpdehaan:
Hi @sivel,

This is a reasonably clean addition

However, we believe the docs being online and live are very important, because people can get corrections, updates, as well as news about Ansible (like AnsibleFests).  

For instance, if we fix a bug in the documentation, that fix is available to everyone instantly, and we don't have a couple thousand people potentially emailing us about errors and typos that are already corrected.   This greatly increases our bandwidth to have it this way, and at our scale relative to number of people answering questions, we deem this necessary.

Another example is when someone reads up on say, Galaxy CLI, and finds out that 1.8 contains a feature they want, they can know they need to upgrade, whereas if reading the 1.7 doc snapshot, would be more apt to file a feature request, not knowing the feature is already there.   

Keeping people on the latest version, by being aware of the new things, gives everyone the best experience.

These are actually one of the greatest assets to the project -- and we believe having a download option where we lose that communication vehicle would be detrimental to us.

At our huge project size, the docs really need to be online and live, so I hope you understand our reasoning here.

This may still change in the future.

cjw296:
It's a shame, as it makes it much harder for me to learn Ansible when I'm commuting and have no internet connection :-(

PR #8907, opened 2014-09-06
Fixed result|success must return 'false' if result.state is 'skipped', fixes #8753

### LINKED ISSUE #8753: check mode causes register: var to be successful
##### Issue Type:

Bug Report
##### Ansible Version:

devel 1.8, 1.7.1
##### Environment:

What OS are you running Ansible from? Ubuntu 14.04 

What OS are you managing? CentOS / Ubuntu 14.04
##### Summary:

Having a playbook where the condition becomes true in check_mode and causes the task to run in check mode while in "live mode" this task is skipped.
##### Steps To Reproduce:

Playbook:

```

---
- hosts: localhost
  connection: local
  gather_facts: no
  tasks:
  - command: /bin/false
    register: test_var
    changed_when: false
    ignore_errors: true

  - debug: msg="should be skipped in check_mode"
    when: test_var|success
```

Live mode:

```
ansible-playbook test.yml 

TASK: [command /bin/false] **************************************************** 
failed: [localhost] => {"changed": false, "cmd": ["/bin/false"], "delta": "0:00:00.003600", "end": "2014-08-26 11:55:17.597110", "rc": 1, "start": "2014-08-26 11:55:17.593510", "stdout_lines": [], "warnings": []}
...ignoring

TASK: [debug msg="should be skipped in check_mode"] *************************** 
skipping: [localhost]

```

Check mode:

```
ansible-playbook test.yml --check

TASK: [command /bin/false] **************************************************** 
skipping: [localhost]
ok: [localhost]

TASK: [debug msg="should be skipped in check_mode"] *************************** 
ok: [localhost] => {
    "msg": "should be skipped in check_mode"
}

```
##### Expected Results:

debug task should be skipped using --check-mode. 
##### Actual Results:

debug task is going to be executed in check mode.

resmo:
Added bug fix and integration tests. This fixes GH-8753 as discussed on the mailing list.

### DIFF
--- a/lib/ansible/runner/filter_plugins/core.py
+++ b/lib/ansible/runner/filter_plugins/core.py
@@ -58,7 +58,7 @@ def failed(*a, **kw):
 
 def success(*a, **kw):
     ''' Test if task result yields success '''
-    return not failed(*a, **kw)
+    return not failed(*a, **kw) and not skipped(*a, **kw)
 
 def changed(*a, **kw):
     ''' Test if task result yields changed '''
--- a/test/integration/Makefile
+++ b/test/integration/Makefile
@@ -19,7 +19,7 @@ TMPDIR = $(shell mktemp -d 2>/dev/null || mktemp -d -t 'mytmpdir')
 
 VAULT_PASSWORD_FILE = vault-password
 
-all: parsing test_var_precedence unicode non_destructive destructive includes check_mode test_hash test_handlers test_group_by test_vault
+all: parsing test_var_precedence unicode non_destructive destructive includes check_mode test_hash test_handlers test_group_by test_vault test_var_results
 
 parsing:
 	ansible-playbook bad_parsing.yml -i $(INVENTORY) -e @$(VARS_FILE) $(CREDENTIALS_ARG) -vvv $(TEST_FLAGS) --tags prepare,common,scenario1; [ $$? -eq 3 ]
@@ -58,6 +58,9 @@ test_hash:
 
 test_var_precedence:
 	ansible-playbook test_var_precedence.yml -i $(INVENTORY) $(CREDENTIALS_ARG) -v -e 'extra_var=extra_var' -e 'extra_var_override=extra_var_override'
+
+test_var_results:
+	ansible-playbook test_var_results.yml -i $(INVENTORY)
 
 test_vault:
 	ansible-playbook test_vault.yml -i $(INVENTORY) $(CREDENTIALS_ARG) -v $(TEST_FLAGS) --vault-password-file $(VAULT_PASSWORD_FILE) --list-tasks
--- a/test/integration/roles/test_var_results/tasks/changed.yml
+++ b/test/integration/roles/test_var_results/tasks/changed.yml
@@ -1,0 +1,42 @@
+# (c) 2014, René Moser <mail@renemoser.net>
+
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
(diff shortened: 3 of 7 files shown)


sivel:
I am concerned about making this change.

I wonder how many people this will cause issues for that were previously relying on this "feature".

cmhamill:
I have concerns also, especially considering that:
- One can already check any combination of success/failure and skipped, because we have filters for each individually.
- By default I expect failure to mean "this host is dropped from the rest of the play," and this would break that consistency.

resmo:
@sivel
They probably implemented the workaround while they experienced the problem in the first run. So it is more a matter of time point. I didn't expect this behaviour and was hit hard, when I ran it in check mode.

@chamill 
To me, it was quite surprising that a result can be successful if it was skipped, the result should be actually unknown.

sivel:
In a way, I think I agree with the current behavior.  Maybe it is cultural, but I would not consider "skipped" an alternative to "failed" or "success".  I would be more likely to say that "changed" and "skipped" are part of "success".

resmo:
... what points to my [question](https://groups.google.com/forum/#!topic/ansible-devel/gy1jv2QcgPA) on the mailing list: "... So the question is, is a task successful when it was skipped?"

Maybe a pragmatic solution would be to have another filter for `|success_skipped` which implements the current behaviour implemented.

jimi-c:
A skipped task is not failed, so by that definition I believe `success` should not exclude it. You can achieve this test with `when: result|success and not result|skipped` without needing a new filter for it.

resmo:
@jimi-c I am totally agree that a skipped task is not failed. Again, can we assume then it is succeed? IMHO we can't.

There are a few workarounds for this behaviour, like `always_run` for a command or your example, (which I actually use). However, the problem I see is more like, you **will not use** `result|success` the **right way** in the **first place**.

I would like to show you some examples: Think about you run this in a **RedHat environment** or in a **Debian environment in check mode**:

```
- command: debianway 
  register: debian_command
  when: ansible_os_family == "Debian"

- service: name=app-debian state=restarted
  when: debian_command|success
```

Another one (more or less out of the docs) running in check mode. uri task will be skipped, **fail will be executed anyway with errors**.

```
- local_action: uri url=http://www.example.com return_content=yes
  register: webpage

- action: fail
  when: webpage|success and 'AWESOME' not in webpage.content
```

Or an example out of my work, only install open-vm-tools if vmware-tools-core is not installed:

```
- name: check if VMware-tools are already installed
  command: rpm -q vmware-tools-core
  register: vmware_tools_installed
  ignore_errors: yes

- name: install open-vm-tools if vmware-tools are not installed
  yum: pkg=open-vm-tools state=installed
  when: vmware_tools_installed_redhat|failed

- name: ensure vmtoolsd is running on RedHat
  service: name=vmtoolsd state=started enabled=yes
  when: vmware_tools_installed_redhat|failed

- name: ensure VMware tools services are running on Redhat
  service: name=vmware-tools-services state=started enabled=yes
  when: vmware_tools_installed_redhat|success
```

Very simple and easy right? In **live mode**, it **works as expected**, **in check mode** the last task will always be **executed and fails**. Especially the circumstance that something will be executed in check mode when it does not in live mode.

Again, **_if_ you know** the behaviour, it is **easy to fix**  in a way , it does what you expect. But this is not simple anymore. It would be simplier, if `|success` would return `false` if skipped. It is not **the principe of least surprise**.

resmo:
any chance to get this in? otherwise feel free to close it.

bcoca:
I'm closing this as I don't see us merging it, for the reasons discussed above:
- it breaks current expectations
- you have ways to work around this, you can test for skipped or if you really want it failed, use failed_when on that task to change the result

PR #8912, opened 2014-09-06
Documentation edit for vsphere_guest module

nathansoz:
I initially couldn't get the vpshere_guest module to work with a standard esxi host. After using pysphere to try to create a vm by hand, I realized that every standalone ESXi host has a default datacenter called ha-datacenter. This isn't displayed anywhere in the web-gui, and I couldn't find it in vmware's documentation. My initial attempt was leaving the datacenter as an empty string (I suspect this would be what others would try as well). I think it would be nice if the module documentation mentioned that standalone connections should use ha-datacenter. If I should make the change somewhere else in the docstring, let me know!

### DIFF
--- a/library/cloud/vsphere_guest
+++ b/library/cloud/vsphere_guest
@@ -60,7 +60,7 @@ options:
     default: None
   esxi:
     description:
-      - Dictionary which includes datacenter and hostname on which the VM should be created.
+      - Dictionary which includes datacenter and hostname on which the VM should be created. For standalone ESXi hosts, ha-datacenter should be used as the datacenter name
     required: false
     default: null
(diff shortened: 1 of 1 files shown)


mpdehaan:
Merged,thanks!  This will appear in our next docs push.

nmartinez-nimeops:
Hi, 
i'm not sure to really understand... 
Is vsphere_guest can create VM on a standalone ESXi without vcenter (free version)?

I've tried to put ESXi hostname in vcenter_hostname... but i've got an [Errno 110] Connection timed out.
So i think that ansible tries to connect to an vCenter :443 

Thank you.

PR #8914, opened 2014-09-07
Add support for using windows domain or active directory users for WinRM connections

jhawkesworth:
Hi

I have a need to support allowing active directory (also known as domain) users on windows hosts with Ansible.  I've been following https://github.com/ansible/ansible/pull/8345 but it has gone a bit quiet and so I've had a go at adding support and tackling what I can of the comments on the above PR.

In summary this PR adds a new parameter, ansible_authorization_type, which can either be omitted or set to 'plaintext' or 'kerberos'.  If it is set to 'kerberos', ansible will attempt to use kinit to set up the connection, silently (and raise AnsibleError if it isn't installed or fails to execute).

This PR includes documentation changes to intro_windows.rst - feedback on setup instructions would be greatfully appreciated especially as I have not been able to test from an ubuntu controller, and I'm not 100% certain I have capture the correct packages required for either centos or ubuntu.

Please note there is a very small change to the configureAnsibleForRemoting.ps1 so if this has been run previously it will either need running again or the firewall rule called "Allow Winr HTTPS" needs to be changed to Profile=all ( the previous version of the rule excluded domain connections).

(new rule is as follows):

netsh advfirewall firewall add rule Profile=all name="Allow WinRM HTTPS" dir=in localport=5986 protocol=TCP action=allow

Currently the connection credentials are not cached, but need to establish if it is functional before attempting to speed things up.

One other concern I have is the setup instructions for windows now look more complicated.  If that's a concern then I could create a separate page for those needing to use domain accounts - its hard to get a feel for how important this feature is to others.

### DIFF
--- a/docsite/rst/intro_windows.rst
+++ b/docsite/rst/intro_windows.rst
@@ -26,7 +26,48 @@ Installing on the Control Machine
 
 On a Linux control machine::
 
-   pip install http://github.com/diyan/pywinrm/archive/master.zip#egg=pywinrm
+   $ pip install http://github.com/diyan/pywinrm/archive/master.zip#egg=pywinrm
+
+If you want to use windows domain user accounts, rather than local users, also install the following on the Linux control machine::
+
+   $ yum install krb5-workstation krb5-libs
+   $ pip install kerberos
+
+or:: 
+
+   $ sudo apt-get install krb5-user python-kerberos
+
+.. note::
+
+    What's a domain user?  Do I need this?
+    
+    Domain users, also known as Active Directory users can be set up on windows computers.
+    This lets allows a user to log in on any machine that is part of the same group of computers 
+    using the same username and password, regardless of which computer they are using.
+    The group of computers is known as a domain or a realm.
+    
+    To make this work, a couple of things need to happen:
+    1/ special machines are added to the network which store the usernames and password
+    of anyone who is allowed to log into the realm.  These machines are called domain controllers and typically there is 
+    a primary domain controller and a secondary one that keeps things going if the primary one fails.
+    2/ a configuration change is made on each machine that needs to be a part of the realm, so that it knows 
+    which machine to ask when someone attempts to log in.   Its worth knowing that a conversation happens between 
+    machines on the domain and the domain controllers which results in something called a ticket being stored.
+    A bit like a concert ticket, the ticket says that you, the ticket holder, are allowed in until a certain time, after which you either 
+    need to leave or get another ticket.
+    
+    Fortunately, domain users were implemented using an existing technology called kerberos.  Since kerberos 
+    is freely available, linux machines can also talk to domain controllers and make use of domain users.
+    
+    So, if you happen to have an existing windows domain you can store a single username and password in Ansible inventory
+    and use it to control many computers.  
+
+    For the purposes of controlling windows computers, Ansible only needs to be able to get hold of a ticket and present
+    it when it starts talking to the windows machine, so the setup instructions below only cover this.  
+    You can do other nice things, such as allowing users to log in to your Ansible controller using the domain password, 
+    caching tickets so that you can log in when away from the network, creating home directories and 
+    mapping network drives when users first log in, or even turning your linux host into a domain controller but 
+    since none of this is necessary for Ansible it is not described here.
 
 .. _windows_inventory:
 
@@ -41,6 +82,7 @@ Ansible's windows support relies on a few standard variables to indicate the use
 
 In group_vars/windows.yml, define the following inventory variables::
 
+    # local user example
     # it is suggested that these be encrypted with ansible-vault:
     # ansible-vault edit group_vars/windows.yml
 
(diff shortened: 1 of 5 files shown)


mpdehaan:
Thanks, this is something we can review.

(Note to James/Chris/Toshio/etc:  please do not merge directly, if we like the implementation I'd like to make a pass over the docs section, though this is a good starting point)

jhawkesworth:
Thanks, please let me know if there is anything I can do to help getting this functionality into Ansible.

AdmiralNemo:
I don't think it's a good idea to unconditionally call kinit. You should look to see if a ticket-granting ticket is available first, and only try to obtain a new one if not, or just expect a ticket to already exist and fail loudly if not. Forcing the user to get a new tgt will break things like winbind's auto-renewal.

Furthermore, you probably shouldn't explicitly give the path to the kinit executable. The Popen object will find it on $PATH if you just call kinit.

jhawkesworth:
Thanks for this.  Good points.

Do you think expecting the ticket to be there, catching the exception if it
is not and then calling kinit in the exception handler would be acceptable?
 Thinking about it, I think that might be preferable to looking for the tgt
each time (which I imagine would be a case of searching the klist output
for a string like "krbtgt/EXAMPLE.COM@EXAMPLE.COM" which strikes me as a
bit fragile).

Definitely better not to have a hardcoded path to kinit, I'll fix that.

AdmiralNemo:
Yes, trying to parse klist output is not a good idea either; I didn't mean to imply that. I'm not convinced that calling kinit is necessary. It may be enough to handle the error and print a helpful message to the user. That would eliminate the need to store the password, since the whole point of Kerberos is to avoid that. I can think of a few cases where calling kinit like that wouldn't work anyway. For example, if pkinit (smart cards) is used, or a keytab. There's also the edge case where the user principal name is different than the login name (rare, I know, but I've seen it).

I think you'll find that most software that uses Kerberos (ssh, smbclient, etc.) will require that the user already have a tgt before it will work. I don't know why Ansible would need to be any different.

AdmiralNemo:
Also, pywinrm may end up having native support for requesting the tgt given a principal and password[1], so having Ansible call kinit directly would not be necessary in that case.

[1] https://github.com/diyan/pywinrm/issues/6#issuecomment-54955244

jhawkesworth:
Thanks for this.
I think I've gone down this route because I didn't want to assume that just
because I was logged into my Ansible controller machine I would (at that
time) necessarily have relevant kerberos TGT tickets for every domain that
has machines in that I want to control.

Correct me if/when I am wrong...
Its my impression that if I logged in to my ansible controller using
kerberos, then I would have a ticket for whatever is defined as default
domain only, lets call that domain A.
If I needed to be able to do things on domains A B and C (perhaps unlikely)
in a single play, then I'd still need to acquire tickets for B and C.
 Calling kinit is not exactly arduous but if you are not running
interactively - say you have a build pipeline that is calling a play then
I'd want to be able to give it everything it needs before I started the
play.

My thinking is that you would not want to force Ansible users to log in to
their controller with kerberos credentials.  In the course of my testing I
have done this - logged into Ansible controller as one user, then run plays
which require a different login name on a specific domain.  I guess I think
Ansible might need to be different from the other apps you mention because
the point of it is to be 'in charge' - its the conductor rather than just
one of the violins (to take a slightly literal interpretation of
'Orchestration').

So I think assuming the TGT is present but requesting it (via kinit or
pywinrm ) if it isn't would let you use domain logins (or indeed a smart
card or keytab) but wouldn't force you to have acquired all the TGTs you
need before you start running your plays.

If I have this all wrong please let me know.

Many thanks,

Jon

nicodeslandes:
I took a different approach and updated pywinrm to support username and password, which are currently ignored if you're using Kerberos with it.
The changes are available there: https://github.com/nicodeslandes/pywinrm/commits/alt_credentials
That unfortunately also required some changes to the underlying pykerberos library, so that it would accept a password. That change is here: https://github.com/nicodeslandes/pykerberos/commits/kerberos_passwords_spike.

With these changes, there's no need to call kinit. pywinrm/pykerberos take care of requesting a TGT for the username/password provided.
This was necessary for me for 2 reasons:
- We plan to use Ansible Tower to automate Ansible deployment. So we cannot call kinit ahead of a deployement
- For some reason the kerberos usernames in my client's production environment start with a hyphen; so kinit doesnt' work there, as it takes the username for an option!

I also added an option to allow the delegation of the user's credentials on the remote server. Useful if for instance you need to access a network share from the Windows server you're accessing.

Now both the proposed solution and the one I implemented in pywinrm/pykerberos have a big security problem in my scenario (using Tower). If user A provides a password for a Kerberos principal, the TGT for this princiap gets added to the credential cache of whichever user executes ansible ('awx' in that case). That means that user B may start a deployment for the same kerberos principal without a password and still have access to a TGT for it from the credential cache.
I haven't found a solution to this issue yet, but we would need a sort of transient credential cache that would only be used for a specific ansible playbook run.

mpdehaan:
Just a quick note, if you want to discuss Windows things, this might make a good topic to start on ansible-devel -- there are some additional folks there that would be interested, and a bit easier to see that there are new replies.

Just saying that as I don't think I've seen you on that list before.

Thanks!

nicodeslandes:
Thanks Michael; I wasn't quite sure what the best way to discuss this was.
I'll create a post about this on ansible-devel.

nicodeslandes:
Here's a link to the topic I started on ansible-devel: https://groups.google.com/forum/#!topic/ansible-devel/TYMdAoN9IK8

jhawkesworth:
I'm closing this PR in favour of Chris Church's implementation mentioned here:
https://groups.google.com/forum/#!searchin/ansible-devel/domain/ansible-devel/TYMdAoN9IK8/7oaNzpHj5UAJ

cchurch:
@jhawkesworth I'll still refer back to this PR for documentation changes; thanks for those!

https://github.com/ansible/ansible/pull/9372/files addresses the setup script changes.

I've found a simpler way to allow a connection plugin to access additional host variables if needed: https://github.com/cchurch/ansible/commit/c0c1cd762d186913ac1cdc3f2612bc9a640f4b47#diff-0

jhawkesworth:
@cchurch  you're welcome to the documentation, I've undeleted the branch in case you need it.

PR #8924, opened 2014-09-08
Further refactoring of OpenStack code

emonty:
This is a follow on to #8388 and should really be ignored probably until that merges since there's no real way here to submit a dependent PR and not show the contents of the first one. BUT ...

I wanted to put this up there because I've been doing a lot of hacking which is getting long in the tooth. The end goal here is to get all of the code which groks OpenStack internal logic, things like "how do I authenticate" or "how do I get a list of images" into the openstack module_utils library so that the things in library/ are more concerned with being expressions of _ansible_ logic, not tons of repeated cantrips that shore up deficiencies in the upstream libraries. Then, the next step from us at OpenStack is to take the code in the module_utils library and get it into upstream libraries, because most of it is not even partially ansible specific, but ansible has been an excellent case study in the deficiencies of our SDK world. Once that's done, then we can start deleting code in module_utils and replacing with cleaner calls to upstream library.

I've got a POC of the end state of that all on my laptop, but the upstream library changes aren't landed yet, so it's not possible to present that here - but I think that showing the movement of the code in general to module_utils should be clear enough, so that the slimming of module_utils and replacement with a better upstream library should similarly be an easy patch to review.

In the process, and because I needed more features in the OpenStack Infra project, I sucked in some of tylernorth's code to support cinder volumes. I took two of his patches, fixed them up to work with module_utils and then squashed them into a single patch.

### DIFF
--- a/lib/ansible/module_utils/openstack.py
+++ b/lib/ansible/module_utils/openstack.py
@@ -28,6 +28,13 @@
 
 import os
 
+HAVE_NOVACLIENT = True
+try:
+    from novaclient.v1_1 import client as nova_client
+    from novaclient import exceptions
+except:
+    HAVE_NOVACLIENT = False
+
 
 def openstack_argument_spec():
     # Consume standard OpenStack environment variables.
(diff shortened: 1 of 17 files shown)


sivel:
I just took a quick gander at things.  There are some improvements you can make around shared module documentation.  Ansible includes plugin functionality for doing exactly this

See an example at:
https://github.com/ansible/ansible/blob/release1.7.1/lib/ansible/utils/module_docs_fragments/files.py

and how it is used at:

https://github.com/ansible/ansible/blob/release1.7.1/library/files/file#L36

I haven't looked at much else yet, but figured that functionality would be of interest.

stenstad:
Shouldn't auth be using keystoneclient instead of the old/deprecated code in novaclient? Look at how auth is implemented in openstackclient.

The current code doesn't work with identity v3 and domains.

emonty:
@sivel AWESOME I will go learn that and use it
@stenstad yeah - I've got it trying to do that and then falling back if it can't.... but next up is to rework how all of those things work. I definiltey want it to all start using keystoneclient directly. Sadly, cinderclient does no support pre-authing with keystoneclient yet. _headdesk_

stenstad:
@emonty Are you sure Cinder can't support that? My python-foo isn't that strong, but it sure seems that is happening in openstackclient/volume/client.py?

mpdehaan:
Hi!

Thanks you very much!

PR #4929, opened 2013-11-17
Test module improvements

hayeah:
Allows the module test runner to:
1. dump the generated module file
2. take json string as complex args to module

PR #8935, opened 2014-09-09
Add options to control output and execution of test-module

willthames:
test-module is useful but sometimes you want to edit the
result before running it to e.g. set a debug point.

Added a noexecute option (i.e. just create the module script, don't
run it) and an output option to choose the filename of the result.


amenonsen:
Looks good, should be merged in preference to #4929 (which has accumulated merge conflicts, and whose complex-args feature has been implemented already).

amenonsen:
#11688 applies without fuzz.

bcoca:
closing via rebased #11688 which kept attribution

PR #8936, opened 2014-09-09
Allow InventoryScript JSON with childgroups only

srgvg:
and without hosts and vars

Without this patch, the simplified syntax is triggered when a group
is defined like this:

```
"platforms": {
    "children": [
        "cloudstack"
    ]
}
```

Which results in a group 'platforms' with 1 host 'platforms'.

```
modified:   lib/ansible/inventory/script.py
```

[Review on Reviewable](https://reviewable.io/reviews/ansible/ansible/8936)


mpeters:
There are no tests with this PR, but I can verify that it passes all the unit and integration tests when merged into the latest devel (ansible 1.8 (devel 91ca8d42e7) last updated 2014/09/18 15:03:04).

PR #8951, opened 2014-09-09
ansible-galaxy should work when roles directory does not exist

### LINKED ISSUE #8950: ansible-galaxy should not need roles directory to exist
Commit aa46fcf09 introduces a check to ensure that `roles_path` exists. 

However, previously, if roles_path did not exist, it just created the directories anyway. 
https://github.com/ansible/ansible/blob/devel/bin/ansible-galaxy#L547
On that basis I'm not sure that the check for whether the directory exists is in the right place (it's nice to just blow away the roles directory entirely sometimes and reinstall from rolesfile)

I would like to understand the reason for this commit (it doesn't seem to be linked to an issue or PR)

cc @jimi-c

willthames:
Roles should be installed when roles directory does not exist (fixes #8950)

This change also fixes installing mercurial roles from ansible-galaxy.

Both of these two changes are codependent because the galaxy tests (make test_galaxy) require both fixes to succeed. 

cc @jimi-c as this reverts a commit of his.

### DIFF
--- a/bin/ansible-galaxy
+++ b/bin/ansible-galaxy
@@ -704,14 +704,6 @@ def execute_install(args, options, parser):
         print "- please specify a user/role name, or a roles file, but not both"
         sys.exit(1)
 
-    # error checking to ensure the specified roles path exists and is a directory
-    if not os.path.exists(roles_path):
-        print "- the specified role path %s does not exist" % roles_path
-        sys.exit(1)
-    elif not os.path.isdir(roles_path):
-        print "- the specified role path %s is not a directory" % roles_path
-        sys.exit(1)
-
     roles_done = []
     if role_file:
         f = open(role_file, 'r')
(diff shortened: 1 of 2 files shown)


abadger:
@willthames, confirmed the first part is fine; is the second part meant to be part of #8959 or is it necessary for this?

willthames:
It's part of making ansible-galaxy work as expected in that it makes ansible-galaxy work with mercurial (and thus fixes all the broken tests) - it's not that it's essential for this change but more for testing that this change works as expected.

mpdehaan:
Yeah can you please update the description to reflect the nature of the change?

(Such as "make ansible-galaxy CLI work with mercurial repos when..." or something?)

It was not clear what "expected" meant.  Thanks!

willthames:
@mpdehaan I've updated title and description. Fixing hg roles is really a side effect of ensuring all tests pass to confirm the fix for missing roles directory.

PR #8955, opened 2014-09-10
Add new Slack callback plugin

sivel:
This PR addresses 2 things:
1. Adds a new callback plugin called slack.py for slack.com
2. Updates the hipchat.py callback plugin to move some work around code into it's more proper method

There is another slack callback plugin proposed in #8829, but unfortunately when I was attempting to use it earlier it was missing some "niceties" that the hipchat.py plugin offered such as color coded messages indicating success or failure, and some additional benefits that the slack API provides.  Additionally, by using some of the enhanced slack APIs the number of messages sent have been reduced by utilizing message "attachments".

This work is based off of the original hipchat.py and the slack.py callback that we are using internally.

It also replaces the slack incoming webhook bot icon with a cool Ansible icon!

### DIFF
--- a/plugins/callbacks/slack.py
+++ b/plugins/callbacks/slack.py
@@ -1,0 +1,249 @@
+# (C) 2014, Matt Martz <matt@sivel.net>
+
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import json
+import uuid
+import urllib2
+
+from ansible import utils
+
+try:
+    import prettytable
+    HAS_PRETTYTABLE = True
+except ImportError:
+    HAS_PRETTYTABLE = False
+
+
+class CallbackModule(object):
+    """This is an example ansible callback plugin that sends status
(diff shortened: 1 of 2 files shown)


sivel:
Here is an example of what the messages look like:

example

sivel:
@mpdehaan would you mind re-opening this?  This is a callback plugin that still lives in the core repo.

Thanks!

baldwinSPC:
@mpdehaan Will this be merged in eventually or should users pull from @sivel 's work branch? I was _just_ about to code this up when I found this PR. Thanks @sivel

sivel:
This will actually need rewritten for Ansible 2.0.  I'll look into doing so soon.  If you are on 1.x I recommend just pulling this file down.

bcoca:
we can merge now, there is new api for callbacks, but the old one is till there and will work.

sivel:
IIRC, this plugin does some deep inspection of the inventory, play and playbook attributes. Since those underlying classes have changed I don't believe it will be compatible.

I have to rewrite for our internal use, so I might as well address this as well.

bcoca:
ah, yes, had not looked at code, if it was doing reference climbing into the internals to slurp other stuff, that won't work anymore

amenonsen:
Is the plan to leave this PR open until the code is rewritten to work with v2?

plugins/callbacks/slack.py
```diff
@@ -0,0 +1,249 @@
...
+
+        data = json.dumps(payload)
+        url = self.webhook_url % (self.team, self.token)
+        req = urllib2.Request(url, data)
+        try:
+            urllib2.urlopen(req)
+        except:
+            utils.warning('Could not submit message to Slack')
+
+    def on_any(self, *args, **kwargs):
```
bcoca, line 102: no need for these if you inherit from new callback base class

jpooler:
I wasn't able to get the v2 version to work(yet), but I was able to get @sivel's script working by updating the webhook url to the line below (looks like slack updated this at some point). I also modified the url arguments that get passed as we only expect 1 value now instead of 2.

self.webhook_url = ('https://hooks.slack.com/services/%s')

url = self.webhook_url % ( self.token)

bcoca:
closing in favor of #13408

Issue #8937, opened 2014-09-09
Ansible Galaxy (requirements file feature): Cloning git repositories via SSH failed


pschirch:
##### Issue Type:

Bug Report
##### Ansible Version:

1.8 (devel)

Last check with: _ansible 1.8 (devel dc1f233e8d) last updated 2014/09/09 11:05:00 (GMT +200)_
##### Environment:

**Host:** Ubuntu 14.04 LTS
##### Summary:

We want to use the new **[role requirements file feature](http://docs.ansible.com/galaxy.html#advanced-control-over-role-requirements-files)** to share internal organization ansible roles over our gitlab server. Cloning git repositories via SSH currently doesn't work. The [Ansible Project Google Group](https://groups.google.com/forum/?hl=de#!topic/ansible-project/Bz-Zy4aK4tU) advised to define such repositories as follows:

``` yaml

---
- src: ssh://git@github.com:geerlingguy/ansible-role-php.git
  path: roles
```
##### Steps To Reproduce:
- create an `requirements.yml` file

``` yaml

---
- src: ssh://git@github.com:geerlingguy/ansible-role-php.git
  path: roles

```
- install requirements

``` bash
ansible-galaxy install -r requirements.yml
```
##### Expected Results:

Cloning repositories via SSH succesfully.
##### Actual Results:

``` bash
- downloading role from ssh://git@github.com:geerlingguy/ansible-role-php.git
- error: failed to download the file.
- ansible-role-php was NOT installed successfully.
- you can use --ignore-errors to skip failed roles.
```

mpdehaan:
I'd probably use http:// for anonymous github downloads, not SSH.

Though in this case we can check and see what's up, that would be the ideal solution.

pschirch:
My steps to reproduce the issue using SSH with Github - correct, but only to simplify matters. In my organization we use [Gitlab](https://about.gitlab.com/) (_open source software to collaborate on code_) to share our internal ansible roles. Every colleague can clone this repositories via SSH (_we do not  provide anonymous [Gitlab](https://about.gitlab.com/) downloads, because our security policy_).

Certainly I can use HTTP authentication:

``` yaml
---
- src: git+https://ourgitlabhost/example-group/example-role.git
  path: roles
```

But the `ansible-galaxy install -r requirements.yml` command requires username and password for every defined repository. Using SSH you add your private key to the ssh agent **only once**. 

[...]

PR #8958, opened 2014-09-10
Autodetect more than github repository, using a more generic heuristic (...

mscherer:
... see #8937 )

### DIFF
--- a/lib/ansible/utils/__init__.py
+++ b/lib/ansible/utils/__init__.py
@@ -401,6 +401,10 @@ def role_spec_parse(role_spec):
     if 'github.com/' in tokens[0] and not tokens[0].startswith("git+") and not tokens[0].endswith('.tar.gz'):
         tokens[0] = 'git+' + tokens[0]
 
+    # automatically detect git repository on hosted gitlab or similar forge
+    if not tokens[0].startswith("git+") and tokens[0].endswith('.git'):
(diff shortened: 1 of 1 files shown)


willthames:
I'm really not a fan of the magic URLs as it is (I originally used git and hg autodetection to work out the type of the repo, but @mpdehaan said that was too magic - I think this is too much magic - and I would have said so about the github.com special case too).

If you are going to do this, my recommendation is to make it DRY and have a function called `guess_scm_from_url` or similar and use it in all code paths that rely on this detection.

mscherer:
Yeah, I was also not fan of having duplicated code path, especially since it took me 5 minutes to see why it was not working as intended by doing half of the modification. I will try to provides a better PR, (with tests for example) if that's ok on the feature point of view.

willthames:
Inferring that repos that end in `.git` are git repos seems reasonable to me. 

Definitely room for improvement in abstracting common code from `role_spec_parse` and `role_yaml_parse`, especially as they're both accumulating similar magic. 

Whether this gets in is not a decision I can make - it is often worth discussing new features in ansible-devel mailing list to avoid spending too much effort on something that isn't wanted.

amenonsen:
@mscherer are updates here following the discussing with @willthames?

mscherer:
Not really. I guess i can close this PR since I totally forgot the context, and the original bug is closed.

PR #8959, opened 2014-09-10
Allow fairly common ssh repo clone path to work

willthames:
Allows github-style ssh repo clone paths (i.e. starting `git@...` rather than `ssh://git@...`)

Addresses #8937. 

Brings in #8951 too (as tests rely on it). Recommend merging #8951 first and then this if desired.

### DIFF
--- a/lib/ansible/utils/__init__.py
+++ b/lib/ansible/utils/__init__.py
@@ -387,15 +387,12 @@ def role_spec_parse(role_spec):
   
     role_spec = role_spec.strip()
     role_version = ''
+    default_role_versions = dict(git='master', hg='tip')
     if role_spec == "" or role_spec.startswith("#"):
         return (None, None, None, None)
 
     tokens = [s.strip() for s in role_spec.split(',')]
     
-    if not tokens[0].endswith('.tar.gz'): 
-        # pick a reasonable default branch
-        role_version = 'master'
-
     # assume https://github.com URLs are git+https:// URLs and not
     # tarballs unless they end in '.zip'
     if 'github.com/' in tokens[0] and not tokens[0].startswith("git+") and not tokens[0].endswith('.tar.gz'):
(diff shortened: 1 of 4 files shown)


lib/ansible/utils/__init__.py
```diff
@@ -412,6 +409,8 @@ def role_spec_parse(role_spec):
         role_name = tokens[2]
     else:
         role_name = repo_url_to_role_name(tokens[0])
+    if scm and not role_version:
+        role_version = default_role_versions.get(scm, '')
```
abadger, line 413: We want role_version to be defined by the end of the function but this conditional leaves out some cases. What about something like:

``` python
if not role_version:
    if scm:
        role_version = default_role_versions.get(scm, 'master')
    else:
        role_version = 'master'
```

willthames:
@abadger does `role_version` need to be `master` if `scm` is unset (i.e. it's an archive)? `role_version` is set to `''` at the start of the function

abadger:
@willthames It doesn't if we detect that it's an archive but jimi remembers some cases where roles stored in git weren't having role_version set to master and so were throwing errors. So it seems safest to set to 'master' by default and change it to something else if we detect that type of role storage. Does that make sense?

willthames:
It would be good if @jimi-c could add such a test case to `test/integration/galaxy_roles.yml` or similar. I don't have much objection to changing `role_version=''` to `role_version='master'` but it seems that is basically hiding some implicit logic somewhere.

jimi-c:
@willthames we can and will definitely add more integration tests for this.

jimi-c:
I'm going to go ahead and merge this in as-is, and I'll retest. It's possible I was doing something incorrectly on my system at the time, or it may have been related to a different bug that was fixed.

jimi-c:
# Merged

Hi!

Thank you!

Issue #8720, opened 2014-08-22
winrm.py: protocol.send_message('') crashes with https


madema:
when winrm port is set to https port (default 5986), winrm.py crashes on protocol.send_message('') statement:

madema@xmgtansible:~/playbooks$ ansible pmgtnix01 -m setup
sudo password:
pmgtnix01 | FAILED => Traceback (most recent call last):
  File "/usr/lib/pymodules/python2.7/ansible/runner/**init**.py", line 561, in _executor
    exec_rc = self._executor_internal(host, new_stdin)
  File "/usr/lib/pymodules/python2.7/ansible/runner/__init__.py", line 666, in _executor_internal
    return self._executor_internal_inner(host, self.module_name, self.module_args, inject, port, complex_args=complex_args)
  File "/usr/lib/pymodules/python2.7/ansible/runner/__init__.py", line 837, in _executor_internal_inner
    conn = self.connector.connect(actual_host, actual_port, actual_user, actual_pass, actual_transport, actual_private_key_file)
  File "/usr/lib/pymodules/python2.7/ansible/runner/connection.py", line 34, in connect
    self.active = conn.connect()
  File "/usr/lib/pymodules/python2.7/ansible/runner/connection_plugins/winrm.py", line 132, in connect
    self.protocol = self._winrm_connect()
  File "/usr/lib/pymodules/python2.7/ansible/runner/connection_plugins/winrm.py", line 90, in _winrm_connect
    err_msg = str(exc.args[0])
IndexError: tuple index out of range

ansible 1.7.1
pywinrm 0.0.2dev

sivel:
I cannot reproduce this.  Any chance you could provide the full python traceback that you get?

sivel:
Perhaps this is the same as what is mentioned at https://groups.google.com/d/msg/ansible-project/XCKgXFIOras/bTz6GbE0eU8J

mpdehaan:
For completeness and not having to click the link, copying here:

```
 I am running Ansible 1.8-dev from source (Tried 1.7.1 from pip as well)

I wanted to control windows machine So I setuped the workstation (Centos 6.3) followed steps on http://docs.ansible.com/intro_windows.html

Below is what I got:

[root@ansible-test-1 ~]# ansible windows  -i ~/hosts  -m win_ping --ask-vault-pass -vvvv
Vault password: 
[...]

PR #8961, opened 2014-09-10
another way of checking winrm hosts

kworr:
While protocol.send_message('') crashes for https why we can't try executing some no-op shell command instead (shamelessly ripped from winrm example)?

### DIFF
--- a/lib/ansible/runner/connection_plugins/winrm.py
+++ b/lib/ansible/runner/connection_plugins/winrm.py
@@ -83,7 +83,11 @@ class Connection(object):
             protocol = Protocol(endpoint, transport=transport,
                                 username=self.user, password=self.password)
             try:
-                protocol.send_message('')
+                shell_id = protocol.open_shell()
+                command_id = protocol.run_command(shell_id, 'rem')
+                print (protocol.get_command_output(shell_id, command_id)[0])
+                protocol.cleanup_command(shell_id, command_id)
+                protocol.close_shell(shell_id)
                 _winrm_cache[cache_key] = protocol
                 return protocol
             except WinRMTransportError, exc:


sivel:
I do believe that this issue has already been resolved in #8720, although perhaps this is a good approach as well.

Additionally, the use of a branch called 'patch-1' generally leads me to believe that the patch was created using the github.com in browser code editor, which generally indicates that the code has not actually been tested.

This is not to say that the approach does not have merit. I am just "concerned".

kworr:
I favor SVN so I'm not into installing git just to submit a patch. It works for me locally for 1.7.1 including one fix for `err_msg = str(exc.args[0])` but that line was superseded by #8720. I'm using locally `repr(exc.args)` instead of `str(exc)`.

I don't know how can I prove it working...

```
> ansible windows -m setup

10.30.20.50 | success >> {
    "ansible_facts": {
        "ansible_distribution": "Microsoft Windows NT 6.1.7601 Service Pack 1",
        "ansible_distribution_version": "6.1.7601.65536",
        "ansible_fqdn": "########",
        "ansible_hostname": "########",
        "ansible_ip_addresses": [
            "10.30.20.50"
        ],
        "ansible_os_family": "Windows",
        "ansible_system": "Win32NT",
        "ansible_totalmem": "4294967296"
    },
    "changed": false
}
```

cchurch:
As @sivel said, the original traceback you had seen has been fixed.

Sending an empty message verifies that ansible can connect and authenticate successfully.  It's roughly equivalent to the following command:

```
curl -vk -d "" -u "user:pass" https://windows-host:5986/wsman"
```

A 411 status code indicates that we connected and authenticated, but provided no request body.  Even a simple shell command requires at least 5 HTTP requests vs. 1 request for the empty message.  Another option I have considered would be to send a request for the winrm configuration (e.g. http://msdn.microsoft.com/en-us/library/cc251718.aspx).

We're not interested in this particular change, but would be open to discussing it further on the development list:

PR #8977, opened 2014-09-11
add distutils package classifiers

billwanjohi:
I was particularly interested in the programming language ones,
but the others might be useful to others browsing PyPI.


amenonsen:
@billwanjohi seems reasonable enough, feel like rebasing to apply to `setup.py` in current devel? Should be trivial.

billwanjohi:
:+1: rebase complete

abadger:
- License classifier should be: License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)
- Not 100% sure about the python language version... It's accurate for the controller (where you run /usr/bin/ansible*).  However, most of the modules are usable on python2 from python2.4 onwards which I bet is the quesetion a lot of sysadmins are actually thinking of when they look at this.  No tsure what to do about that mismatch in expectations....  Any other @ansible/core people want to weigh in on this?
- Topic :: Utilities might be another one to add.

bcoca:
also:

```
'Intended Audience :: Developers',
'Intended Audience :: Release Managers',
'Intended Audience :: General Automaters',
```

bcoca:
QA , auditors ... is there a 'only not end user/desktop'?

abadger:
There's only a limited set of classifiers:
https://pypi.python.org/pypi?%3Aaction=list_classifiers

abadger:
we talked about python version and decided that what you have (2.6 and 2.7) is fine.  It's the best we can do and think that most sys admins won't read the classifiers alone to decide whether this will work for them.

billwanjohi:
OK, I updated the license and added the Utilities topic. As @abadger says, classifiers is a strict set, so I left those alone.

PR #8982, opened 2014-09-12
Extra vars precedence issue with includes

sgargan:
Issue Type:

Bugfix Pull Request

Ansible Version:

1.7.1+ 

Environment:

Mac OSX 10.9

Summary:

Fixes variable precedence issue with include files where extra vars were not overriding default values as expected.

Steps To Reproduce:

Expected Results:

- include: myinclude.yml some_var='{{extra_var}}'
  
Actual Results:


sgargan:
realized this is on devel branch, moving to a bugfix_branch

PR #8984, opened 2014-09-12
Extra vars precedence issue with includes  

sgargan:
Issue Type:

Bugfix Pull Request

Ansible Version:

1.7.1+

Environment:

Mac OSX 10.9, bug is not os specific.

Summary:

Fixes variable precedence issue with include files where extra vars were not overriding default values as expected.

Steps To Reproduce:

Expected Results:

include: myinclude.yml some_var='{{extra_var}}'

Actual Results:

### DIFF
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -605,10 +605,10 @@ class Runner(object):
         # use combined_cache and host_variables to template the module_vars
         # we update the inject variables with the data we're about to template
         # since some of the variables we'll be replacing may be contained there too
-        module_vars_inject = utils.combine_vars(host_variables, combined_cache.get(host, {}))
+        module_vars_inject = utils.combine_vars(self.extra_vars, combined_cache.get(host, {}))
+        module_vars_inject = utils.combine_vars(host_variables, module_vars_inject)
         module_vars_inject = utils.combine_vars(self.module_vars, module_vars_inject)
         module_vars = template.template(self.basedir, self.module_vars, module_vars_inject)
-
         inject = {}
 
         inject = utils.combine_vars(inject, self.default_vars)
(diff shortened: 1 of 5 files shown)


jimi-c:
Hi, thanks for reporting this! 

I'm not quite sure I understand why you've got the extra vars being merged in first, with the host variables second, meaning the extra vars would be overwritten by inventory variables? Can you explain your logic here?

sgargan:
The behavior I was looking for is that extras alway trump vars configured
anywhere else, So that if I have a variable 'cluster' that defaults to dev
then I can override this with an extra at the command line -e cluster=qa.
This was the behavior prior to the 1.7.0 variables refactor.

> extra vars being merged in first, with the host variables second, meaning
> the extra vars would be overwritten by inventory variables? Can you explain
> your logic here?

I played around with the positioning of the merge so that the expected
'extras win' behavior is what is observed. I know it seems counterintuitive
that extras is the first set processed, but that is what works.

All the tests pass, including the ones I added. I've also been running this
patch on my deployment servers and haven't seen any issues with host
variables getting inadvertently clobbered. Are you seeing an issue? If
there is anything else you'd like to see tested here I'd be happy to add
it, just let me know.

Thanks,

Steve.

jimi-c:
@sgargan the order of merging determines their variable precedence, where the first ones merged in essentially have a lower precedence than those merged in later. Since "extra vars" in Ansible are supposed to have the highest precedence, it seems counter-intuitive to me that they were being merged in first and yet "winning" when it came to showing the final variable value.

You said "I've also been running this patch on my deployment servers and haven't seen any issues with host variables getting inadvertently clobbered", however it is the "extra vars" that I'm concerned about getting clobbered in this scenario.

I will definitely test this, I just want to make sure that extra vars are not being trumped by inventory variables.

sgargan:
Closed in favor of #9106.

PR #8988, opened 2014-09-12
Fix a bug causing yum module to not update to latest in some cases

bbasleeper:
No updates when using state=latest, enablerepo=<something> and yum-utils package is not installed on target server. Because of a typo, the list of disabled repos was set to the list of enabled repos.


mpdehaan:
Merged, thank you!

PR #8994, opened 2014-09-12
Noting version when regex_replace was added.

ycombinator:
Noting on docs.ansible.com that the `regex_replace` filter was added in Ansible version 1.6.

### DIFF
--- a/docsite/rst/playbooks_variables.rst
+++ b/docsite/rst/playbooks_variables.rst
@@ -349,6 +349,8 @@ To match strings against a regex, use the "match" or "search" filter::
 
 To replace text in a string with regex, use the "regex_replace" filter::
 
+.. versionadded:: 1.6
+
     # convert "ansible" to "able"    
     {{ 'ansible' | regex_replace('^a.*i(.*)$', 'a\\1') }}         
 


jimi-c:
Hi @ycombinator, thanks for the contribution.

One issue here is that the new line needs to be added outside of the block started on the previous line with `::`. As is, the `version_added` line is entered into the pre/code block rather than the indented section below it. If you can get that fixed we can get this merged in . Thanks!

amenonsen:
Here's an updated diff against current devel:

```
diff --git a/docsite/rst/playbooks_filters.rst b/docsite/rst/playbooks_filters.rst
index 10ea62f..5be04e0 100644
--- a/docsite/rst/playbooks_filters.rst
+++ b/docsite/rst/playbooks_filters.rst
@@ -384,6 +384,8 @@ To match strings against a regex, use the "match" or "search" filter::

 'match' will require a complete match in the string, while 'search' will require a match inside of the string.

+.. versionadded:: 1.6
+
 To replace text in a string with regex, use the "regex_replace" filter::

     # convert "ansible" to "able"    
```

bcoca:
fixed via bb1a907823712f8d25f22bfef4ff51309a10178b

PR #8995, opened 2014-09-13
Do not remove local role file if given on the commandline

mscherer:
using this (for testing purpose):

```
$ ansible-galaxy install COPYING
- error: the file downloaded was not a tar.gz
- COPYING was NOT installed successfully.
- you can use --ignore-errors to skip failed roles.
```

this result in COPYING being erased, which is surprising for the user.
This also prevent erasing requirements.yml if someone use the wrong flag.


abadger:
Looks good.  I looked further down the method and saw another os.unlink(tmpfile) which probably was added before seeing that this one existed.  I'll remove that when I merge this.

PR #8997, opened 2014-09-13
Fix hostname update on CentOS 6

dooblem:
The actual code fails with the following error on centos 6:

```
$ ansible xxxxxx -m hostname -a 'name=toto'
195.154.117.15 | FAILED >> {
    "failed": true, 
    "msg": "Command failed rc=127, out=, err=/bin/sh: hostnamectl: command not found\n"
}
```

### DIFF
--- a/library/system/hostname
+++ b/library/system/hostname
@@ -372,7 +372,7 @@ class CentOSHostname(Hostname):
 class CentOSLinuxHostname(Hostname):
     platform = 'Linux'
     distribution = 'Centos linux'
-    strategy_class = FedoraStrategy
+    strategy_class = RedHatStrategy
 
 class ScientificHostname(Hostname):
     platform = 'Linux'


abadger:
@dooblem Which version of Centos6 are you running on?  I'm looking at a centos 6.5 box where distribution is "Centos" and a centos 7.0 box where distribution is "Centos linux".    So for those boxes, the code is setting the strategy correctly.  If there's a centos 6 variant that uses "Centos linux" then we'll probably need to add a version check as well as matching the distribution string.

( You can check which distribution your box will return by doing:  `python -c 'import platform; print(platform.linux_distribution()[0])'` )

dooblem:
Hello,
Here it is:

```
# python -c 'import platform; print(platform.linux_distribution()[0])'
CentOS Linux
[root@atome ~]# cat /etc/issue
CentOS Linux release 6.0 (Final)
Kernel \r on an \m

# cat /etc/redhat-release 
CentOS Linux release 6.0 (Final)
# cat /etc/sysconfig/network
NETWORKING=yes
NETWORKING_IPV6=yes
HOSTNAME=atome
```

abadger:
Thanks, @dooblem !

Because the strings identifying centos6.x don't appear to be consistent I took your PR as a basis and changed it to take version into account.  I think that will resolve the issue for you.  If not feel free to open up a new issue or otherwise get my attention and we can figure out what the right approach  is.

# Closing This Ticket

issue is resolved.

Thank you!

dooblem:
Works like a charm (at least on my centos 6).
Thanks a lot.

Issue #8875, opened 2014-09-03
Ansible fails to install correctly into a virtualenv when built as a wheel


evansd:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.7.1
##### Environment:

Ubuntu 14.04
##### Summary:

When built as a Python wheel and then installed from that wheel into a virtualenv, ansible fails to populate the `share/ansible` directory, and so breaks.

Note that other that this issue, ansible and its dependencies install fine from wheels.
##### Steps To Reproduce:

``` bash
virtualenv wheel-test
source wheel-test/bin/activate
pip install wheel
pip wheel ansible==1.7.1
pip install --use-wheel --find-links wheelhouse ansible==1.7.1
ls wheel-test/share/ansible
```
##### Expected Results:

```
cloud     internal    net_infrastructure  source_control      windows
commands  inventory   network             system
database  messaging   notification        utilities
files     monitoring  packaging           web_infrastructure
```
##### Actual Results:

```
ls: cannot access wheel-test/share/ansible: No such file or directory
```

mpdehaan:
# Possible Misunderstanding

Hi!

Thanks very much for your submission to Ansible.  It sincerely means a lot to us.

We believe the ticket you have filed is being somewhat misunderstood, as one thing works a little differently than stated.

In particular, we believe you have said:
- that ansible tries to install into LSB specific locations is a bug

However, however what is happening is that:
- ansible is installing into LSB applications because various contents are not python modules and it's a systems management application, not just a basic python library.  There are other options though...

We would suggest:
- you refer to instructions in http://docs.ansible.com/intro_installation.html -- you could use pip in editable mode, or simply do a checkout and run things with "source ./hacking/env-setup" - no installation is required.

In the future, this might be a topic more well suited for the user list, which you can also post here if you'd like some more help with the above.
- https://groups.google.com/forum/#!forum/ansible-project

Thank you once again for this and your interest in Ansible!

evansd:
I completely understand that this isn't a priority for you (and I'll try to track down the issue and submit a patch myself) but I don't think I've misunderstood the issue here.

Just to clarify, installing in a virtualenv without building a wheel first works perfectly:

```
virtualenv wheel-test
source wheel-test/bin/activate
pip install ansible==1.7.1
ls wheel-test/share/ansible
# displays correct directories
```

Installation should work identically whether using the package or the wheel. I think that's true regardless of whether it's a "basic python library" or not.

[...]

PR #8999, opened 2014-09-14
Add doc warning about pip install from wheel

msabramo:
When pip installing from a wheel into a virtualenv, this creates a broken installation (https://github.com/ansible/ansible/issues/8875)

So I added a warning to the installation docs about this, because this is a tricky problem to debug.

Related to: #8875

Cc: @evansd, @aconrad, @sontek, @georges

### DIFF
--- a/docsite/rst/intro_installation.rst
+++ b/docsite/rst/intro_installation.rst
@@ -239,6 +239,15 @@ If you are installing on OS X Mavericks, you may encounter some noise from your 
 
 Readers that use virtualenv can also install Ansible under virtualenv, though we'd recommend to not worry about it and just install Ansible globally.  Do not use easy_install to install ansible directly.
 
+.. note::
+
+   Ansible is reported to not install correctly into a virtualenv when built
+   from a `wheel <http://wheel.readthedocs.org/>` (see `this issue`
+   <https://github.com/ansible/ansible/issues/8875>). So if you tend to build
+   wheels for your Python packages, you may want to skip building a wheel for
+   Ansible and/or pass the ``--no-use-wheel`` flag when doing a ``pip
+   install``.
+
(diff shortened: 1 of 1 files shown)


mpdehaan:
I suspect lots of people won't know anything about wheel, so we probably don't need to explicitly document it.

We do document suggested install and build steps, so we believe this should not occur.

Hope this is understandable.

Thanks!

sontek:
I disagree that people wont know about wheels, it is the recommended way to create binary packages in the python packaging ecosystem.  It is mentioned in the first paragraph in https://packaging.python.org/en/latest/  and is a first class command when you run `pip help`.  I think it is very important to warn against using wheel's if they are not supported with ansible.

sivel:
2 things to note:
1. ansible is not a binary package
2. installing in a virtual environment is not the recommended way to install ansible

sontek:
@sivel Wheels are not tied to a virtualenv at all.  You can use system python + pip with a wheel with no issues at all.

aconrad:
@sivel It doesn't have to be binary. For large automated infrastructure, turning pure python packages as wheels will speed up installs.

msabramo:
> We do document suggested install and build steps, so we believe this should not occur.

I think that this can still occur.

The documentation says that you can do:

```
sudo pip install ansible
```

For most users, this will be absolutely fine of course, because PyPI doesn't have a wheel for ansible. However, some users like myself may use an internal package index (e.g.: devpi) and they might build wheels for their packages to make provisioning faster. Pip will default to using a wheel if it's available, so in this case, the above command will lead to breakage.

In other words, the Ansible docs don't know for sure that a wheel won't be used. So it is possible than an external factor could change what the above command does and cause it to not work.

PR #9002, opened 2014-09-15
Fix ssh_config.py bug with IdentityFile

lorin:
When IdentityFile is specified in an ssh config file, using the ssh_config.py inventory plugin would yield the error:

```
AttributeError: 'list' object has no attribute 'startswith'
```

This is because Paramiko returns a list of private key paths when parsing the
ssh config, and Ansible expects a string.

This commit fixes the issue by taking the first element in the list returned by Paramiko.


t0mk:
Hey, I fixed this and tested with a lot of paramiko versions in
https://github.com/ansible/ansible/pull/9365

lorin:
@t0mk Cool, I'll close this pull request since your solution works across multiple paramiko versions.

PR #9003, opened 2014-09-15
fix TestModuleUtilsBasic.py to run in python 2.6

mpeters:
The unittest library in python 2.6 does not have the `assertIn()` or `assertNotIn()` methods. This patch checks for their existence and falls back to an `assertTrue()` or `assertFalse()` equivalent.

Without this patch running the unit tests in 2.6 fails with an error like this:

```
ERROR: test_log_sanitize_correctness (TestModuleUtilsBasic.TestModuleUtilsBasicHelpers)
  File "/root/ansible/test/units/TestModuleUtilsBasic.py", line 297, in test_log_sanitize_correctness
    self.assertNotIn('pas:word', url_output)
AttributeError: 'TestModuleUtilsBasicHelpers' object has no attribute 'assertNotIn'
```

### DIFF
--- a/test/units/TestModuleUtilsBasic.py
+++ b/test/units/TestModuleUtilsBasic.py
@@ -294,12 +294,20 @@ class TestModuleUtilsBasicHelpers(unittest.TestCase):
         ssh_output = self.module._heuristic_log_sanitize(ssh_data)
 
         # Basic functionality: Successfully hid the password
-        self.assertNotIn('pas:word', url_output)
-        self.assertNotIn('pas:word', ssh_output)
+        if 'assertNotIn' in dir(self):
+            self.assertNotIn('pas:word', url_output)
+            self.assertNotIn('pas:word', ssh_output)
+        else:
+            self.assertFalse('pas:word' in url_output)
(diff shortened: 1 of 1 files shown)


amenonsen:
This PR should be targeted to `test/units/module_utils/test_basic.py` now (which does still use assertIn and assertNotIn). But it's not clear to me how far we want to go to preserve Python 2.6 compatibility here, and even if we did want it, something that emulated assertIn/assertNotIn for 2.6 would be better than adding this kind of code to all potential callers. @jimi-c, @bcoca, perhaps you could comment?

abadger:
We have a compat library that uses the unittest2 Library if unittest
doesn't support what we need.
https://github.com/ansible/ansible/blob/devel/lib/ansible/compat/tests/unittest.py

To maintain compat we should use that rather than fallback to assertTrue

amenonsen:
@mpeters, how about revising your PR to use the compat unittest stuff?

abadger:
Just checked and in devel we're using the compat.tests.unittest import so this should be working.  (Also, travis is now running unittests on python2.6 and those are passing so I think we're good to go here.)

PR #9005, opened 2014-09-15
Update `with_subelements` to handle sub-lists located deeper in dict

jumoel:
Before, `with_subelements` could only handle that the nested list was one element 'deep' compared to the first argument.

Now, `with_subelements` accepts the second parameter in the form of `foo.bar.baz`, which will get the list at `foo.bar.baz` instead of complaining that the literal key `foo.bar.baz` doesn't exist.

### DIFF
--- a/lib/ansible/runner/lookup_plugins/subelements.py
+++ b/lib/ansible/runner/lookup_plugins/subelements.py
@@ -55,12 +55,19 @@ class LookupModule(object):
             if item0.get('skipped',False) != False:
                 # this particular item is to be skipped
                 continue 
-            if not subelement in item0:
+            
+            item = item0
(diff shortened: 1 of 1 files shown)


dmkoch:
+1 This would be extremely useful.

podung:
+1 We are utilizing the change from the PR.  Would love to see this pull request accepted.

cadetstar:
+1 Would be handy.

johnalla0:
+1, we need this

alanflorendo:
+1 yes please

dylanjw:
+1

jumoel:
Something similar seems to have been merged into what will be Ansible v2: #10995.

PR #9016, opened 2014-09-16
docker: fix using images without a namespace

mantiz:
I wasn't able to run images without a namespace like `busybox`, I always got:

```
  File "/home/mantiz/.ansible/tmp/ansible-tmp-1410827919.81-129122313988220/docker", line 2086, in <module>
    main()
  File "/home/mantiz/.ansible/tmp/ansible-tmp-1410827919.81-129122313988220/docker", line 775, in main
    running_containers = manager.get_running_containers()
  File "/home/mantiz/.ansible/tmp/ansible-tmp-1410827919.81-129122313988220/docker", line 597, in get_running_containers
    for i in self.get_deployed_containers():
  File "/home/mantiz/.ansible/tmp/ansible-tmp-1410827919.81-129122313988220/docker", line 573, in get_deployed_containers
    image, tag = self.get_split_image_tag(image)
  File "/home/mantiz/.ansible/tmp/ansible-tmp-1410827919.81-129122313988220/docker", line 524, in get_split_image_tag
    registry, resource = image.rsplit('/', 1)
ValueError: need more than 1 value to unpack
```

I think this was introduced by 1c3039f0 where the check for `/` was removed.

I hope I didn't break anything.

### DIFF
--- a/library/cloud/docker
+++ b/library/cloud/docker
@@ -507,15 +507,22 @@ class DockerManager:
 
     def get_split_image_tag(self, image):
         # If image contains a host or org name, omit that from our check
-        registry, resource = image.rsplit('/', 1)
+        if '/' in image:
+            registry, resource = image.rsplit('/', 1)
+        else:
+            registry = None
+            resource = image
 
         # now we can determine if image has a tag
         if resource.find(':') > 0:
             resource, tag = resource.split(':', 1)
-            return '/'.join((registry, resource)), tag
         else:
             tag = "latest"
-            return image, tag
+
+        if registry:
+            resource = registry + '/' + resource
+
+        return resource, tag
 
     def get_summary_counters_msg(self):
         msg = ""


jimi-c:
Hi @mantiz, thanks for the contribution! I believe @abadger fixed this yesterday, if you'd like to re-pull and test the latest devel branch. If there is still an issue there, please let us know.

abadger:
@mantiz I think I may have broken that with yesterday's fix, actually.  Let me check.

abadger:
@mantiz, sorry - I didn't see that this was a pull request rather than a simple bug report.  And a good thing that it was a pull request too!  After I committed I saw that your change took care of another corner case that mine did not.  I've updated the change to be closer to yours: 9f708305696e685718496b11aa846fa8c6a655fb

Thanks for the fix!

mantiz:
@abadger, glad I could help. But I think there is still an issue with the current solution. If I am correct, the current code does not (re)add an eventually found registry if the given image has no tag.

I think this part

``` python
if registry:
    resource = '/'.join((registry, resource))
```

has to be placed directly before the return statement in any case, wether or not the given image parameter contained a tag or not. Am I right?

abadger:
@mantiz  Thanks for checking on this.  If there's no tag then we set the resource variable to the original string we were given as a parameter (`resource = image`).  That way the return at the end contains the registry without us having to recombine anything.

mantiz:
@abadger oh dear, you're totally right. Sorry for this and thanks for the explaination. ;)

PR #9017, opened 2014-09-16
defining Top-level env for main(), modules easily extensible via import

abhishekkr:
The way core modules are written, they are not extensible (import and build upon).
Suppose if I want to import ansible file module for reusing existing pieces of "library/files/lineinfile" level functions, currently... the module is like

```
# import module snippets
from ansible.module_utils.splitter import *

main()
```

which will call main() even if it is imported.

This Pull Request modifies, all such module definitions (all python, except one in Powershell for Windows), to be like

```
if __name__ == "__main__":

    # import module snippets
    from ansible.module_utils.splitter import *

    main()
```

which will make sure, "main()" only runs when it get called as a script and not when it gets imported to be extended.

### DIFF
--- a/library/cloud/azure
+++ b/library/cloud/azure
@@ -478,7 +478,10 @@ class Wrapper(object):
                     raise e
 
 
-# import module snippets
-from ansible.module_utils.basic import *
+if __name__ == "__main__":
 
-main()
+
+    # import module snippets
+    from ansible.module_utils.basic import *
+
+    main()
--- a/library/cloud/cloudformation
+++ b/library/cloud/cloudformation
@@ -228,7 +228,7 @@ def main():
     try:
         cf_region = Region(region)
         cfn = boto.cloudformation.connection.CloudFormationConnection(
-                  aws_access_key_id=aws_access_key, 
+                  aws_access_key_id=aws_access_key,
                   aws_secret_access_key=aws_secret_key,
                   region=cf_region,
               )
@@ -306,8 +306,11 @@ def main():
 
     module.exit_json(**result)
 
-# import module snippets
-from ansible.module_utils.basic import *
-from ansible.module_utils.ec2 import *
 
-main()
+if __name__ == "__main__":
+
+    # import module snippets
+    from ansible.module_utils.basic import *
+    from ansible.module_utils.ec2 import *
+
+    main()
--- a/library/cloud/digital_ocean
+++ b/library/cloud/digital_ocean
@@ -428,7 +428,10 @@ def main():
     except (DoError, Exception), e:
         module.fail_json(msg=str(e))
 
-# import module snippets
-from ansible.module_utils.basic import *
 
-main()
+if __name__ == "__main__":
+
+    # import module snippets
+    from ansible.module_utils.basic import *
+
+    main()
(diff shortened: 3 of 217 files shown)


abhishekkr:
used a script to edit files, so made it remove trailing whitespaces as well
forgot to mention in commit message

jimi-c:
# Question About This One

Hi!

Unfortunately, we're not sure if we want this feature in the program. We don't want this to seem confrontational, however our reasons for this are:
- Many of the modules rely on the AnsibleModule code brought in via the `from ansible.module_utils.basic import *` line, so without that the vast majority of the module code would not work and would be useless outside of a few utility functions. If these utility functions are generic or common to other modules, they should be moved to a new file within `lib/ansible/module_utils/` instead.

mpdehaan:
I appreciate the effort that went into this.

I think I'm ok with the **name** == 'main' stuff everywhere, but this would need to be submitted without (A) moving the import, and (B) without doing anything else to the other modules.

We're unlikely to accept this just because of the danger of something going wrong in one untested module, and needing to see this applied to 235+ of them.

The whitespace changes are problematic as they break attribution in the source history that we'd like to preserve.

I'd be open to this if it only makes the "**main**" stuff be indented, but I prefer to see imports remain at the top of the file, especially the way they are handled automagically not being actual real imports (due to some behind the scenes voodoo that is useful to the 'no-bootstrapping' nature of ansible)

As the above says, I'm willing to entertain a discussion on the devel list as well, in case we've missed an important use case.

abhishekkr:
@jimi-c
I should have noted the crippling w/o imports, my mistake.
What if the change was just

```
if __name__ == "__main__":
    main()
```

it will import all required modules, and still not call main() if imported.
Getting changed PR would just take 5minutes of mine, but the thing is do you guys want that. Otherwise it's one more PR for you folks to go through among so many. Let me know.

@mpdehaan
I understand the risk of large commits. At my side it was all single style change and single task that made the change, so I went ahead with one commit.
I've not changed action logic of any module and just the minimalist part where it get called, so it'is actually review-able from Diffs (though would take some time).
As @jimi-c did found the flaw.
And if we don't have test around core modules, dev-practices say let's built those first... instead of fearing to change them. Let me know if need support on that.

jimi-c:
@abhishekkr the module templating code takes those special lines and turns them into inserted snippets. Without that module templating, those imports will not work and the modules will still be crippled. Short of creating a stub in those modules that did the same thing, I don't see any other way for these to work.

abhishekkr:
@jimi-c hey, thanks for the info... I noticed the temporary python files pushed to remotes machines being appended core module logic and pieces it depends upon (I suppose).
Are you mentioning the same behavior being done by module templating?
Can you please guide me to section of that module templating logic, I'd like to have a look at it please.

Also had a query, if my one custom module imports some other custom python libraries (or a custom module)... is the templating engine currently capable of tracing those as well and appending them? Is there a specific convention around it, like right before calling main or so?

jimi-c:
The module templating code is done via `lib/ansible/module_common.py`. Examples of its use can be found also in the unittest code `test/units/TestModuleUtilsBasic.py`.

jimi-c:
Regarding your second question, the templating code does not insert code from standard python libraries, since those should be available on the target system. It only inserts code snippets from `lib/ansible/module_utils/`, where the form of the import is `from ansible.module_utils import XXXX`.

PR #9019, opened 2014-09-16
don't re-encode file if it wasn't edited

lsiudut:
This patch is introducing intelligent editing - if file was NOT edited it won't be re encoded, but instead it will be safely removed (it's required that shred is installed in system and available in $PATH).


mpdehaan:
Closing this per user request.

PR #9021, opened 2014-09-16
Default debug message is misleading.

jedops:
Printing Hello world can be misleading and to a newbie can imply that you got some kind of test code hanging around. I think a more specific message would be a better way of doing this.


jimi-c:
# Question About This One

Hi!

- The message is supposed to be generic, whereas the version you've supplied could also be confused by a novice user as error-related. The module documentation pretty clearly spells out that this message will be shown if none is provided, so the chances of this being confused for other output is minimal.

PR #9022, opened 2014-09-16
Remove xterm escape sequences from output

cyberj:
When working with a server which use xterm escape squences (ex: dynamic title, see http://www.faqs.org/docs/Linux-mini/Xterm-Title.html#s3 ) I have the following error:

```
$ ssh root@myserver tail -n 2 /root/.bashrc
echo -en "\033]0;${USER}@$(hostname)\007"

$ ansible -i hosts -u root -m setup myserver    
myserver | FAILED >> {
    "failed": true, 
    "msg": "\u001b]0;root@myserver\u0007/usr/bin/python: can't open file '\u001b]0': [Errno 2] No such file or directory\r\n/bin/sh: 1: root@myserver\u0007/root/.ansible/tmp/ansible-tmp-1401869953.18-104973309409260/setup: not found\r\n", 
    "parsed": false
}
```

This patch fix this issue.

### DIFF
--- a/lib/ansible/executor/process/result.py
+++ b/lib/ansible/executor/process/result.py
@@ -131,16 +131,17 @@ class ResultProcess(multiprocessing.Process):
 
                     for result_item in result_items:
                         # if this task is notifying a handler, do it now
-                        if 'ansible_notify' in result_item and result.is_changed():
-                            # The shared dictionary for notified handlers is a proxy, which
-                            # does not detect when sub-objects within the proxy are modified.
-                            # So, per the docs, we reassign the list so the proxy picks up and
-                            # notifies all other threads
-                            for notify in result_item['ansible_notify']:
-                                if result._task._role:
-                                    role_name = result._task._role.get_name()
-                                    notify = "%s : %s" % (role_name, notify)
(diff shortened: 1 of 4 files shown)


mpdehaan:
Yeah short term workaround here is to not have a bashrc that is outputting things - this can cause problems with various applications.

I can somewhat understand setting the prompt.

Can you share the logic for making it output this?   Thanks!

cyberj:
On all our servers we have a `.bashrc` line like : 

```
if [ "$TERM" != dumb ]; then echo -en "\033]0;${USER}@$(hostname)\007"; fi
```

So, when we connect to one of them, our client terminal can display user and hostname on each tab/windows : usefull when you have 10+ ssh connections and you want to navigate between them.

Usually, when an application can't handle specific term behavior (like xterm ones) or in a non-interactive mode, it set `$TERM` to `dumb`. But doing this with ansible will break some features which need an interactive terminal.

http://unix.stackexchange.com/questions/43945/whats-the-difference-between-various-term-variables

Btw, the second short-term workaround is to use paramiko.

athoune:
Paramiko is far slower than the SSH driver.

PR #9028, opened 2014-09-16
updated git module documentation

dhoss:
Clarified key_file option for the git module, since it requires the user to specify a private key, not a public key.


mpdehaan:
Merged, thank you!  This will appear the next time we build/push docs, which should be soon.

PR #9043, opened 2014-09-17
Find modules when ansible is installed from a wheel

### LINKED ISSUE #8875: Ansible fails to install correctly into a virtualenv when built as a wheel

msabramo:
This is simple as adding `<site_packages>/usr/share/ansible` to the list of paths to look for modules (`DIST_MODULE_PATH`).

Fixes https://github.com/ansible/ansible/issues/8875

Cc: @evansd, @sontek, @aconrad, @mcdonc

### DIFF
--- a/lib/ansible/constants.py
+++ b/lib/ansible/constants.py
@@ -15,6 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 
+import distutils
 import os
 import pwd
 import sys
@@ -104,7 +105,10 @@ else:
 local_module_path = os.path.abspath(
     os.path.join(os.path.dirname(__file__), '..', '..', 'library')
 )
-DIST_MODULE_PATH = os.pathsep.join([DIST_MODULE_PATH, local_module_path])
+site_packages_usr_share_ansible = os.path.join(
+    distutils.sysconfig.get_python_lib(),
+    'usr', 'share', 'ansible')
+DIST_MODULE_PATH = os.pathsep.join([DIST_MODULE_PATH, site_packages_usr_share_ansible, local_module_path])
 
 # check all of these extensions when looking for yaml files for things like
 # group variables -- really anything we can load


msabramo:
@evansd: Have you gotten a chance to try this? Just curious.

evansd:
Hi @msabramo, I haven't had a chance yet I'm afraid! I can totally believe it works though, but I don't think it fixes the root problem. Ideally, installing from source and installing from a wheel should result in exactly the same set of files on disk. If we can get that right then the extra search path won't be necessary.

msabramo:
@evansd: I agree that it would be nice if ansible were consistent about where stuff got installed, but IIRC, it has to do with the fact that ansible likes to install stuff in `/usr/share` when not installed in a virtualenv. I'm not sure why it does this, but @mpdehaan seems to believe pretty strongly that it should be done. I suspect he won't accept a PR that changes substantially the way ansible does things, because he likes to be very conservative and not risk breaking things if he doesn't have to. So that's why I went with a very small, focused change that keeps the status quo.

evansd:
Ah, I see. Is the problem that, when building the wheel, pip doesn't know whether it's going to be installed in a virtualenv or not, so there's no way of putting the files in the correct place?

I don't really know enough about the wheel format to know if this is the case, and if so whether it's fixable. But I'm planning to look into it further when I get them time.

If there really is no way of getting the ansible wheel to do the right thing then your patch seems like a good workaround.

msabramo:
Amazingly, there have been huge changes to ansible's packaging lately (looking much cleaner now) and this doesn't merge cleanly anymore.

Hopefully, wheels work now but I haven't tried it.

PR #9044, opened 2014-09-17
Fix bdist_wheel

msabramo:
Without this, I get:

```
$ python setup.py bdist_wheel
...
copying ./library/cloud/azure ->
build/bdist.macosx-10.4-x86_64/wheel/private/tmp/ansible.venv/bin/../share/ansible/cloud
error:
build/bdist.macosx-10.4-x86_64/wheel/private/tmp/ansible.venv/bin/../share/ansible/cloud:
No such file or directory
```


msabramo:
Well, this doesn't merge anymore, because there has been some major refactoring of `setup.py` and how the packaging works (yay! it looks nicer now!).

Hopefully installing from a wheel now works and this PR isn't need anymore.

PR #9058, opened 2014-09-18
minor tweak to documentation of required modules for running tests

mpeters:
This minor doc change mentions that python-nosetests is sometimes packaged as python-nose. Also mentions python-passlib which is needed to get the tests to pass.


jimi-c:
Merged, thanks!

PR #9059, opened 2014-09-18
documenting extra packages that need to be installed to run the full integration tests

mpeters:
Minor doc addition to mention packages that are needed by the full integration tests (svn, hg, python-pip and gem).


jimi-c:
Merged, thanks!

PR #9060, opened 2014-09-18
add LANG=C to apt module so the string matches on the output always matc...

jamespharaoh:
Ansible's apt module is checking the output of the command in order to determine the result. As a result, we need to set the language so that the output can be matched correctly. This change hard codes the value of the LANG environment variable, which is similar to what the "subversion" module does.


abadger:
# Merged

Hi!

issue is resolved.

Thank you!

Issue #7816, opened 2014-06-18
get_url should support md5 verification


pdeva:
##### Issue Type:

Feature Idea
##### Ansible Version:

ansible 1.6.2
##### Environment:

N/A
##### Summary:

get_url supports only sha256 verification right now.
it is weird that md5 verification was left out of this task.
please allow get_url to do md5 verification
##### Steps To Reproduce:

If this is a bug ticket, please enter the steps you use to reproduce the problem in the space below.  If this is a feature request, please enter the steps you would use to use the feature.  If an example playbook is useful, please include a short reproducer inline, indented by four spaces.  If a longer one is necessary, please link one uploaded to gist.github.com. 
##### Expected Results:

Please enter your expected results in this space.  When running the steps supplied above, what would you expect to happen?  If showing example output, indent your output by four spaces so it will render correctly in GitHub. 
##### Actual Results:

Please enter your actual results in this space.  When running the steps supplied above, what actually happened?  If showing example output, indent your output by four spaces so it will render correctly in GitHub.

hahnicity:
I'm going to take this on

jirutka:
+1

RealSalmon:
+1

mpdehaan:
Hi!

Thanks very much for your interest in Ansible.  It sincerely means a lot to us. 

On September 26, 2014, due to enormous levels of contribution to the project Ansible decided to reorganize module repos, making it easier
for developers to work on the project and for us to more easily manage new contributions and tickets.

We split modules from the main project off into two repos, http://github.com/ansible/ansible-modules-core and http://github.com/ansible/ansible-modules-extras

If you would still like this ticket attended to, and believe this problem or idea is still present in the latest version of Ansible (1.7.2) or the development branch, we will need your help in having it reopened in one of the two new repos, and instructions are provided below.

We apologize that we are not able to make this transition happen seamlessly, though this is a one-time change and your help is greatly appreciated -- 
this will greatly improve velocity going forward.

Both sets of modules will ship with Ansible, though they'll receive slightly different ticket handling.

To locate where a module lives between 'core' and 'extras'
- Find the module at http://docs.ansible.com/list_of_all_modules.html
- Open the documentation page for that module
- If the bottom of the docs say "This is an extras module", submit your ticket to https://github.com/ansible/ansible-modules-extras
- Otherwise, submit your pull request to update the existing module to https://github.com/ansible/ansible-modules-core
[...]

PR #9062, opened 2014-09-18
Issue #7816

hahnicity:
Adding MD5 checksum verification for the get_url module

This is in response to https://github.com/ansible/ansible/pull/8201. It was easier to just create a new branch and then cherry-pick the changes from the old branch

### DIFF
--- a/library/network/get_url
+++ b/library/network/get_url
@@ -76,6 +76,14 @@ options:
     version_added: "1.3"
     required: false
     default: null
+  md5sum:
+    description:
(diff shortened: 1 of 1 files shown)


jimi-c:
@hahnicity I had actually gotten your branch resolved myself, but in looking at the other PRs I had some questions as to whether or not they were a better long term solution. I was going to run my thoughts by @mpdehaan, so for now I'm going to close this. If we do go with your solution, I've got the branch from #8201 ready to go.

Thanks!

PR #9063, opened 2014-09-18
Allow ansible (ad-hoc) to support --extra-vars

sivel:
Currently the `ansible` command does not allow the use of `--extra-vars`.  This has been a frequent request in IRC recently.  I see no technical reasons why this wasn't included.

This PR adds `--extra-vars` support to the `ansible` command and moves the `--extra-vars` option to `ansible.utils.base_parser` and moves the now common `extra_vars` parsing into `ansible.utils.parse_extra_args`

A common use case is to override `ansible_python_interpreter`.

### DIFF
--- a/bin/ansible
+++ b/bin/ansible
@@ -120,6 +120,8 @@ class Cli(object):
         if not options.ask_vault_pass and options.vault_password_file:
             vault_pass = utils.read_vault_file(options.vault_password_file)
 
+        extra_vars = utils.parse_extra_vars(options.extra_vars, vault_pass)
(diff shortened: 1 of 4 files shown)


bcoca:
~~not sure why people need this for ad-hoc as it cannot template, what use are the extra vars?~~

never mind, i just saw from related ticket, it allows for inventory connection vars override, please rebase

sivel:
Will do and I will update once I have done so.

sivel:
Rebase complete.

PR #9064, opened 2014-09-18
support more version for pip module

### LINKED ISSUE #9011: pip module to support version targeting, greater than, less than, etc: >=, >, <=, <
##### Issue Type:

Feature Idea / Limitation
##### Ansible Version:

ansible 1.7.1
##### Environment:

N/A
##### Summary:

pip itself supports something like this:
pip install ansible>=1.71

but the pip module in Ansible only support (http://docs.ansible.com/pip_module.html)
pip: name=ansible version=1.7.1
##### Steps To Reproduce:
- name: install Ansible
  action: pip name=ansible version>=1.7.1
##### Actual Results:

Error message:
msg: unsupported parameter for module: version>

kpx-dev:
Fix #9011 

@jimi-c can you help CR please.

Thanks,
Kien

### DIFF
--- a/library/packaging/pip
+++ b/library/packaging/pip
@@ -147,10 +147,18 @@ def _get_cmd_options(module, cmd):
     
 
 def _get_full_name(name, version=None):
-    if version is None:
-        resp = name
-    else:
+    resp = name
+
+    if version is not None:
+        # this will support version='>1.0.1'
+        for op in ('>', '>=', '<', '<='):
+            # remove all white space so it can support version=' >1.0.1'
+            version = str(version).strip()
+            if version.startswith(op):
+                return name + version
+
         resp = name + '==' + version
+
     return resp
 
 def _is_present(name, version, installed_pkgs):
@@ -219,7 +227,7 @@ def main():
         argument_spec=dict(
             state=dict(default='present', choices=state_map.keys()),
             name=dict(default=None, required=False),
-            version=dict(default=None, required=False, type='str', aliases=['version>', 'version<']),
+            version=dict(default=None, required=False, type='str'),
             requirements=dict(default=None, required=False),
             virtualenv=dict(default=None, required=False),
             virtualenv_site_packages=dict(default='no', type='bool'),
--- a/library/packaging/pip
+++ b/library/packaging/pip
@@ -219,7 +219,7 @@ def main():
         argument_spec=dict(
             state=dict(default='present', choices=state_map.keys()),
             name=dict(default=None, required=False),
-            version=dict(default=None, required=False, type='str'),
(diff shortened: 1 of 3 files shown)


jimi-c:
Hi @kienpham2000, I'm not sure if putting the </> in the param name vs. the value is the correct way to go here, and I'm not clear that it would even work like this being set in the aliases for the `version` param? Right now, you can actually do something like `name="foo>=x.y.z"` without specifying the version directly and it works as expected, so I think this probably just needs an overhaul in the pip module.

kpx-dev:
@jimi-c yea right now it only work with >=, <= but not >, < since it doesn't have equal sign. Do you know if any other modules in ansible have support for something that is not assignment? Ex: version>1.0

Right now it fails everytime here if we don't have =
https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L967

Do you allow module writer to overwrite this file or modify it to support better assignment?

jimi-c:
@kienpham2000 right, I would not expect that to work as you have it, since the parameters need the '=' to be properly parsed. Instead, if you do `name=foo>x.y.z` it should work as you expect.

My proposal to make this work with version would be to use this patch instead:

<pre>
diff --git a/library/packaging/pip b/library/packaging/pip
index 613717a..b49a797 100644
--- a/library/packaging/pip
+++ b/library/packaging/pip
@@ -150,7 +150,12 @@ def _get_full_name(name, version=None):
     if version is None:
         resp = name
     else:
-        resp = name + '==' + version
+        for op in ('>', '>=', '&lt;', '&lt;='):
+            if str(version).startswith(op):
+                resp = name + version
+                break
+        else:
+            resp = name + '==' + version
     return resp 
 </pre>

kpx-dev:
@jimi-c I actually had similar implementation but it was failed by this:
https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L967

Maybe I'll have to patch that base module to support = assignment for key

jimi-c:
@kienpham2000 that line you highlighted is why `version<x.y.z` will not work. But with my patch above, you would instead do `version='>x.y.z'`, which would work.

jimi-c:
To make it clearer, you could instead use the "complex arguments" syntax with my patch:

<pre>
- pip:
    name: foo
    version: ">x.y.z"
</pre>

or:

<pre>
- pip:
    name: foo
    version: ">=x.y.z"
</pre>

kpx-dev:
@jimi-c I see now. Ok I'll patch it up with your suggestion.

kpx-dev:
@jimi-c pushed new changes based on your feedback. I also have to create a new dummy_test package so that I can test multiple versions.

mpdehaan:
Hi!

Thanks you very much!

Issue #6294, opened 2014-03-05
Ansible 1.5 file lookup plugin strips double quotes


aabdnn:
##### Issue Type:

Bug Report
##### Ansible Version:

1.5
##### Environment:

N/A
##### Summary:

One of either the file lookup plugin or the authorized_key module, is removing double quotes from the file being read.
##### Steps To Reproduce:

Create a play like this:

authorized_key: user=monitoring key="{{lookup('file', 'monitoring.pub')}}"

In the monitoring.pub file, insert the content:

from="addr1,addr2,addr3",no-pty ssh-rsa AAAA....

Then run the play.
##### Expected Results:

I expect the contents of the file monitoring.pub to be inserted into the monitoring user's authorized_keys file.
##### Actual Results:

Ansible gives me the error:

msg: invalid option string: from=addr1,addr2,addr3,no-pty

mpdehaan:
So what is happening here is the lookup plugin is evaluated at Jinja2 level and then inserted directly into the YAML.

To preserve the double quotes, you will need to escape them, because the first quote around the "{{" will match up with the quote around the ="addr1, and close off the pair.

So in your file:

```
from=\"addr1,addr2,addr3\",no-pty ssh-rsa AAAA....
```

And the inline expansion won't break things.

I also imagine you might be able to leave out the quotes.  

This isn't going to be fixable but is a consequence of how the lookup plugin is evaluating your line.

A cleaner way to get around this is also this:

authorized_key: 
    user: monitoring 
    key: "{{lookup('file', 'monitoring.pub')}}"

Notice the passing of arguments as hash members versus key=value strings.

In which case the string won't be "shlexed" and will just work, without modifying the file contents.

Hope this helps!

aabdnn:
Hi Michael,

I understand your comments. However, in 1.4.3, my setup was working. So are you saying that the file lookup plugin in 1.4.3 was buggy, and that you have fixed it in 1.5?

PR #9068, opened 2014-09-19
Recommend use of 'key: value' syntax instead of content= and key=

amenonsen:
Using key="{{ value }}" syntax with copy's content parameter and authorized_key's key parameter can cause problems when the value contains quotes or newlines. This is especially noticeable when a lookup('file', …) returns a key with quoted key options (issue #6294). There have also been multiple regressions in this area between 1.6.8, 1.6.9, and devel.

The use of "key: value" syntax makes all these problems go away, so this commit recommends this form in the examples for the respective modules. (I feel a more general warning and guidance about when to use key=val and when to use key: val is also needed, but I'll leave it to the maintainers to determine a place for it.)

### DIFF
--- a/docsite/rst/playbooks_loops.rst
+++ b/docsite/rst/playbooks_loops.rst
@@ -154,10 +154,12 @@ It might happen like so::
     - user: name={{ item.name }} state=present generate_ssh_key=yes
       with_items: users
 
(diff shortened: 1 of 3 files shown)


mpdehaan:
Hi!

Thanks you very much!

PR #9106, opened 2014-09-23
Templating of injected vars

sgargan:
Issue Type:

Bugfix Pull Request

Ansible Version:

1.7.1+

Environment:

Mac OSX 10.9, bug is not os specific.

Summary:
Only substitute templates in variables once the order of all injected variables has been decided.

So based on comments by James for the previous pull request for this issue #8984, I took another look at what was going on here and it seems that the templates in variables are being resolved before all of the variables sets are combined into the the injected set. Moving the templating to a point where the full set has been gathered removes this issue, with existing and added tests passing. I will close the other pull request.

Steps To Reproduce:

Expected Results:

include: myinclude.yml some_var='{{extra_var}}'

Actual Results:

### DIFF
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -587,14 +587,7 @@ class Runner(object):
         host_variables = self.inventory.get_variables(host, vault_password=self.vault_pass)
         combined_cache = self.get_combined_cache()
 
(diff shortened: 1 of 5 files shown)


amenonsen:
This PR can be closed, the problem has already been fixed in devel. (I tested.)

bcoca:
closing as per comment above

PR #9115, opened 2014-09-24
Add a new option immediate= to immediately apply a permanent change

mscherer:
Currently, either you apply the change in the configuration
of firewalld ( without permanent=True ), or you apply it live.
I most of the time want to do the 2 at the same time, ie open the
port ( so I can use the service ) and make sure it stay open on reboot.

I discussed this with @maxamillion about this, and tought this would be a good addition to the firewalld module

### DIFF
--- a/library/system/firewalld
+++ b/library/system/firewalld
@@ -52,6 +52,11 @@ options:
       - "Should this configuration be in the running firewalld configuration or persist across reboots"
     required: true
     default: true
+  immediate:
+    description:
+      - "Should this configuration be applied immediately, if set as permanent"
+    required: false
+    default: false
   state:
     description:
       - "Should this port accept(enabled) or reject(disabled) connections"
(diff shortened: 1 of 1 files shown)


maxamillion:
While I'm not the biggest fan of the fact that upstream 'firewall-cmd --add-service=ssh --permanent' (or $other) doesn't result in the service port being immediately open, it is how upstream administrative utilities function so I wanted to try and keep the module consistent with that. @mscherer suggested this approach and I think it's a good solution.

+1 LGTM

jimi-c:
This looks good, however I would tweak the message `msgs.append('Non-permanent operation')` to more accurately reflect what's going on there, as it could be a permanent operation now.

mscherer:
So since the repo was changed, let's continue the discussion on the new PR : 
https://github.com/ansible/ansible-modules-extras/pull/1

jimi-c:
I agree, closing this one since you re-submitted on the new repo.

Thanks!

PR #9117, opened 2014-09-24
module: wordpress

dz0ny:
This module enables installing wordpress core, it's themes and plugins from ansible.

### DIFF
--- a/library/web_infrastructure/wordpress
+++ b/library/web_infrastructure/wordpress
@@ -1,0 +1,238 @@
+#!/usr/bin/python -tt
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2014, Janez Troha <dz0ny@ubuntu.si>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


sivel:
A few things to note here based on a quick glance at the module:
1. You shouldn't preferably be using urllib3 and requiring it as a dependency.  You should be using `fetch_url` as provided by `ansible.module_utils.urls` and it's associated functions/functionality
2. Perhaps it might be a better requirement to instead rely on `wp-cli` since this seems to be the community standard for performing these types of tasks
3. `version_added` needs to be bumped to `1.8`, since that would be the next upcoming release, 1.6 was released about 6 months or so ago
4. There is an API for themes too, although this could be alleviated by using `wp-cli`: `http://api.wordpress.org/themes/info/1.0/`
5. Generally speaking, unit tests should not be included in the module itself, or not even within ansible.  It is preferable to add integration tests instead

jimi-c:
Hi @dz0ny. Thanks for the submission, though the issues @sivel has pointed out above are all definite issues. Since we're about to split the modules out from the main Ansible repo (see the ansible-project/devel groups on Google for details), I'm going to go ahead and close this, since the PR will need to be pointed at the new repo.

So, as soon as we have the new repos in place, we'd be very happy if you'd resubmit this then, with the above issues addressed.

Thank you!

dz0ny:
@sivel && @jimi-c thx for info

tersmitten:
Great work. I like the fuctionality.

PR #9118, opened 2014-09-24
Add Chef inventory plugin

stucki:
Provide inventory data from a Chef environment.

This script uses the "knife" command to talk to the Chef server, so it must run within a Chef project directory.

### DIFF
--- a/plugins/inventory/chef.py
+++ b/plugins/inventory/chef.py
@@ -1,0 +1,149 @@
+#!/usr/bin/env python
+
+"""
+External inventory script for Chef
+===================================
+
+Ansible has a feature where instead of reading from /etc/ansible/hosts
+as a text file, it can query external programs to obtain the list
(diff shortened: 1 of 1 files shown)


mpdehaan:
Hi!

- we do not want to maintain code related to other configuration management systems in our core project since we would like to devote resources to working on Ansible instead, and don't want to keep up with changes to Chef and knife.
- while we do recognize lots of people are in transitional states with their infrastructure and use more than one config tool, or are large orgs and have different departments using different tools, this sends some mixed messages about using ansible for configuration management to new users, who are not aware of that

My suggestion would be that you continue to maintain this one on GitHub, and even write a blog about it, though we're just not going to include it in core.

Hopefully this is understandable.

stucki:
Hi Michael, thanks for your feedback. That is ok with me, I'll publish it in a separate repository and add the link here when it's ready...

PR #9120, opened 2014-09-24
defining top-level env 'just' for main(), import extensible modules from new lib

abhishekkr:
the call to main() in python-based core modules gets changed to, just

> if **name** == "**main**":
>     main()

it will import all required modules, and still not call main() if imported into any newly created module by users who just wanna extend some functionality over existing modules

### DIFF
--- a/library/cloud/azure
+++ b/library/cloud/azure
@@ -481,4 +481,6 @@ class Wrapper(object):
 # import module snippets
 from ansible.module_utils.basic import *
 
-main()
+
+if __name__ == "__main__":
+    main()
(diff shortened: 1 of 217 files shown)


jimi-c:
Hi @abhishekkr, due to the fact that we're about to split modules out into separate repos, we're going to go ahead and close this PR. Please note however, that we still have concerns regarding the use of this, as the module code was never designed to be run independently of the core code, and will not function in most instances without the templating of the extra code from module_utils/basic.py (at the very least).

Thanks!

PR #9127, opened 2014-09-24
Limit what is returned by the cobbler inventory plugin based on the filter_status attribute

etrikp:
This pull requests adds the ability to limit what the cobbler plugin returns, based on the --status attribute of a system in cobbler. The attribute is configurable via cobbler.ini and supports either a single string, or a list of strings as a value. The attribute defaults to production when its not present in the ini file.

### DIFF
--- a/plugins/inventory/cobbler.ini
+++ b/plugins/inventory/cobbler.ini
@@ -12,9 +12,13 @@ host = http://PATH_TO_COBBLER_SERVER/cobbler_api
 #   - ansible-cobbler.index
 cache_path = /tmp
 
+# We need to support the ability to filter results from cobbler so inventory
+# can be isolated per environment. For example, if we want staging (acceptance)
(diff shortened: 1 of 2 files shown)


jimi-c:
It looks like there are some commented out `print` statements, which should be cleaned up. Also, if the filter_status is not set, it appears that this may filter out everything, since there is no check to see if the filter_status is not None before comparing it to the host['status'].

etrikp:
Thanks for the feedback! I've removed the old print statements and added a test for filter_status.

amenonsen:
@etrikp any chance you could rebase this to apply to current devel?

amenonsen:
I'll close this PR now, since it hasn't been updated in months, but you're welcome to open a new PR if you get a chance to rebase and test this change. @etrikp Thanks for your contribution to Ansible, it is appreciated.

PR #9130, opened 2014-09-24
EC2 Inventory Boto Profile Support

jalaziz:
This PR effectively combines #5987 and #8582. It also fixes the merge conflicts in the two PRs and standardizes the option names.

I also removed the environment variable support as boto has since added AWS_PROFILE (starting with version 2.29) which accomplishes the same thing and is consistent with AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY.

There is a minor inconsistency right now with the cache directory names. If you use the AWS_PROFILE environment variable, your cache directory name will not change. However, if you use the CLI option or the config option, the cache directory will be named after the profile. We could check the AWS_PROFILE for naming, but the behavior is debatable and I've started a discussion on the ansible-devel mailing list.

### DIFF
--- a/contrib/inventory/ec2.ini
+++ b/contrib/inventory/ec2.ini
@@ -134,3 +134,7 @@ group_by_elasticache_replication_group = True
 # tag Name value matches webservers1*
 # (ex. webservers15, webservers1a, webservers123 etc) 
 # instance_filters = tag:Name=webservers1*
+
+# A boto configuration profile may be used to separate out credentials
+# see http://boto.readthedocs.org/en/latest/boto_config_tut.html
+# boto_profile = some-boto-profile-name
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -165,8 +165,7 @@ class Ec2Inventory(object):
         # as pre 2.24 boto will fall over otherwise
         if self.boto_profile:
             if not hasattr(boto.ec2.EC2Connection, 'profile_name'):
-                sys.stderr.write("boto version must be >= 2.24 to use profile\n")
-                sys.exit(1)
+                self.fail_with_error("boto version must be >= 2.24 to use profile")
 
         # Cache
         if self.args.refresh_cache:
@@ -364,7 +363,6 @@ class Ec2Inventory(object):
                     continue
                 self.ec2_instance_filters[filter_key].append(filter_value)
 
-
     def parse_cli_args(self):
         ''' Command line argument processing '''
 
@@ -403,10 +401,7 @@ class Ec2Inventory(object):
             conn = boto.connect_euca(host=self.eucalyptus_host)
             conn.APIVersion = '2010-08-31'
         else:
-            conn = ec2.connect_to_region(region)
-        # connect_to_region will fail "silently" by returning None if the region name is wrong or not supported
-        if conn is None:
-            self.fail_with_error("region name: %s likely not supported, or AWS is down.  connection to region failed." % region)
+            conn = self.connect_to_aws(ec2, region)
         return conn
 
     def boto_fix_security_token_in_profile(self, connect_args):
@@ -415,7 +410,6 @@ class Ec2Inventory(object):
         if boto.config.has_option(profile, 'aws_security_token'):
             connect_args['security_token'] = boto.config.get(profile, 'aws_security_token')
         return connect_args
-
 
     def connect_to_aws(self, module, region):
         connect_args = {}
@@ -426,8 +420,10 @@ class Ec2Inventory(object):
             self.boto_fix_security_token_in_profile(connect_args)
 
         conn = module.connect_to_region(region, **connect_args)
+        # connect_to_region will fail "silently" by returning None if the region name is wrong or not supported
+        if conn is None:
+            self.fail_with_error("region name: %s likely not supported, or AWS is down.  connection to region failed." % region)
         return conn
-
 
     def get_instances_by_region(self, region):
         ''' Makes an AWS EC2 API call to the list of instances in a particular
@@ -440,14 +436,8 @@ class Ec2Inventory(object):
                 for filter_key, filter_values in self.ec2_instance_filters.items():
                     reservations.extend(conn.get_all_instances(filters = { filter_key : filter_values }))
             else:
-                conn = ec2.connect_to_region(region, profile_name=self.boto_profile)
+                reservations = conn.get_all_instances()
 
-            # connect_to_region will fail "silently" by returning None if the region name is wrong or not supported
-            if conn is None:
-                print("region name: %s likely not supported, or AWS is down.  connection to region failed." % region)
-                sys.exit(1)
-
-            reservations = conn.get_all_instances()
             for reservation in reservations:
                 for instance in reservation.instances:
                     self.add_instance(instance, region)
@@ -459,7 +449,6 @@ class Ec2Inventory(object):
                 backend = 'Eucalyptus' if self.eucalyptus else 'AWS' 
                 error = "Error connecting to %s backend.\n%s" % (backend, e.message)
             self.fail_with_error(error)
-
 
     def get_rds_instances_by_region(self, region):
         ''' Makes an AWS API call to the list of RDS instances in a particular
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -296,6 +296,11 @@ class Ec2Inventory(object):
         self.cache_path_index = cache_dir + "/ansible-ec2.index"
         self.cache_max_age = config.getint('ec2', 'cache_max_age')
 
+        # boto configuration profile
+        self.boto_profile = None
+        if config.has_option('ec2', 'boto_profile'):
+            self.boto_profile = config.get('ec2', 'boto_profile')
+
         # Configure nested groups instead of flat namespace.
         if config.has_option('ec2', 'nested_groups'):
             self.nested_groups = config.getboolean('ec2', 'nested_groups')
@@ -430,7 +435,7 @@ class Ec2Inventory(object):
                 for filter_key, filter_values in self.ec2_instance_filters.items():
                     reservations.extend(conn.get_all_instances(filters = { filter_key : filter_values }))
             else:
-                conn = self.connect_to_aws(ec2, region)
+                conn = ec2.connect_to_region(region, profile_name=self.boto_profile)
 
             # connect_to_region will fail "silently" by returning None if the region name is wrong or not supported
             if conn is None:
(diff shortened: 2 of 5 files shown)


plugins/inventory/ec2.py
```diff
@@ -308,6 +334,26 @@ def do_api_calls_update_cache(self):
         self.write_to_cache(self.index, self.cache_path_index)
 
 
+    def boto_fix_security_token_in_profile(self, connect_args):
+        ''' monkey patch for boto issue boto/boto#2100 '''
+        profile = 'profile ' + self.boto_profile
+        if boto.config.has_option(profile, 'aws_security_token'):
+            connect_args['secuirty_token'] = boto.config.get(profile, 'aws_security_token')
```
dwhjames: typo?

``` python
connect_args['secuirty_token']
```

mtotheikle:
I'm new to Ansible and the work-flow that the Ansible team uses for merging pull requests but is there any chance this will be available soon in master? This is an important feature for my team as we deal with multiple AWS accounts.

plugins/inventory/ec2.py
jalaziz: Nice catch. I'll fix this and update the PR.

stackmagic:
I just had a look at the code and I'm a bit confused about the `boto_profile`, `EC2_PROFILE` and `AWS_PROFILE` (env)variables... Are they always the same?

I suspect the `EC2` one in `intro_dynamic_inventory.rst` should be named `AWS` too. This might yet be a bug in there?

Will this PR also work when I 'only' have profiles in my `~/.aws/credentials` file or does it require the use of `~/.boto`?

jalaziz:
@stackmagic Thanks for catching that typo.

To answers your other question, they aren't necessarily the same. AWS_PROFILE comes from environment, while boto_profile should override it. However, taking a look at the boto code, it looks like there may be a bug where profile_name on the connection is not honored.

Profiles defined anywhere that boto accepts should work fine. This PR is mainly for overriding these "default" profile names.

There is an issue with this PR currently, where only boto_pfoile will result in the inventory cache being saved separately per profile. I'm not really sure what the correct behavior should be, since AWS_PROFILE and other boto config methods are part of the calling environment. On the other hand, it can be argued that the inventory should still be smart to handle profiles regardless.

I'll do some testing later this week to confirm my suspicions about boto_profile and update the PR accordingly.

willthames:
@jalaziz, I don't think you should assume that everyone can upgrade to 2.29. 

However, centralising on one environment variable whether or not supported by boto would be a sensible approach.

willthames:
The inventory cache should definitely be per-profile (having needed to connect to two different profiles within five minutes, I can say this is an essential feature). 

The documentation should also be updated to reflect these changes.

jalaziz:
@willthames, profiles were introduced in 2.24 and the environment variable was introduced in 2.29. I could use the environment variable to set the parameter between 2.24 and 2.29.

However, support for profiles in general will require people to upgrade to at least 2.24.

jalaziz:
@willthames, Yeah, the more I think about it, the more I believe we should make the extra effort to figure out the current profile and namespace the cache. The part that I believe is debatable is how do we handle the default profile? You are able to use an environment variable to set a default profile and it would be nice to namespace in all cases. However, last I checked, extracting the current profile from boto wasn't trivial as it wasn't been stored correctly (this may have been fixed, but will restrict the supported version further).

willthames:
@jalaziz, I know profiles were added in 2.24 (https://github.com/jalaziz/ansible/blob/e0c997184bb453bfa956234050033d360fd8ab37/plugins/inventory/ec2.py#L157-L159) - I added the checks to ensure that at least attempting to use profiles with a pre-2.24 boto would fail with a useful message. And that's fine.

I'm not sure I really understand the complexity of determining the default profile. If it's in an environment variable, that's easy. If it's in ~/.boto, that should be relatively easy to extract too. So is is set differently to either of those (and even boto.config can't handle it, it's just an ini file). Is there a reference to a bug/doc about this problem?

gmalouf:
This feature is critical for anyone who has to work with multiple AWS accounts.  What is needed to get it merged?

bcoca:
first it needs rebasing, it would also help  to solve the issues others have brought up

willthames:
ok, rebase looks reasonable at https://github.com/willthames/ansible/compare/pr-9130?expand=1 (I genuinely don't know what better way to update this PR, other than create a new one, which will likely cause yet more delays)

I don't think any of the outstanding questions are that important, compared to actually allowing profiles to work in Ansible, a feature I first implemented over a year ago in the original #5987. We could have had a dozen pull requests against the original implementation by now to fix any problems identified rather than just sitting on it.

bcoca:
@willthames since most of the questions outstanding came from you, I'm ok then with just a rebase.

I'm not sure why there would be a need to create a new PR?

willthames:
@bcoca I've rebased all the commits against latest devel on my repo, but this PR is from @jalaziz's repo, and I can't push to that - so not sure anyone can fix up this PR other than @jalaziz

jalaziz:
@willthames Sorry for being completely out of touch recently. Been swamped at work and haven't had much time to fix this PR. I'd be happy to work on cleaning it up (there were a few issues I wanted to fix) later tonight/tomorrow. However, in the meantime, I've add you as a collaborator on my fork so you can push.

willthames:
@jalaziz no worries - thanks for the push access, the history is looking a little tidier after the rebase too (it eliminated the merge commits).

amenonsen:
@jalaziz @willthames any plans to rebase this to apply to current devel?

jalaziz:
@amenonsen I've been meaning to, but I'm going to have very limited internet access over the next week (vacation). I'll have plenty of free time after, so I'll work on it as soon as I'm back on land.

willthames:
@IanVS I think that case is covered in https://github.com/ansible/ansible/pull/9130/files#diff-c2e371537339bbab5d9d0cc5d6cb7adcR424

@amenonsen, @jalaziz, I've rebased from devel and pushed.

Once this is accepted I also have a fix for when a user doesn't have access to ElastiCache or RDS but I don't think that should be part of this PR

IanVS:
@willthames Yes that handles the case in `connect_to_aws`, where you are testing for presence of `self.boto_profile` (https://github.com/jalaziz/ansible/blob/ec2-inventory-profiles/contrib/inventory/ec2.py#L420).  Why would you not also need to do the same in `get_instances_by_region`, which is still called directly by `do_api_calls_update_cache`?  Or you could change the call in `do_api_calls_update_cache` to `connect_to_aws`, which is maybe what you intended.

willthames:
@IanVS - good spot, thanks. Updated.

bcoca:
cc @defionscode @bennojoy

willthames:
@bcoca, hasn't this one gone around the houses enough? It keeps needing to be rebased because other stuff gets merged in in between every time you ask for further confirmation that this is ok.

halberom:
bump @defionscode @bennojoy

juliedavila:
Oof, let's get one more rebase, I have cycles to test this week.

amenonsen:
@defionscode @willthames @jalaziz I had a quick look, and the part that needs rebasing is really the changes to `inventory/__init__.py` in order to cope with the recent (fairly major) changes there.

A quick summary: this PR supports `ansible-playbook -i 'ec2.py --profile prod' myplaybook.yml` as well as passing in the profile name by setting `AWS_PROFILE=xxx` in the environment (or a setting in `ec2.ini`). The latter forms are straightforward (and I prefer them for this reason, but note that I don't use profiles, so I may be missing something). The `ec2.py --profile prod` form, however, requires special support in the inventory parser. If I understand the code correctly (with `git fetch origin pull/9130/head:pr-9130 && git diff $(git merge-base devel pr-9130)..pr-9130`), this would also allow `hosts: ec2.py --profile prod --foo bar`, and indeed the same with any other inventory plugin. Or any other executable.

Now, I think it's worthwhile to have some mechanism to pass in more options to inventory plugins, but I think there has to be more to it than what this patch does. I'm willing to be convinced otherwise, but at the moment I'm leaning towards thinking it would be best to allow profile selection only through the `ini` file or the environment. The extra command-line option can be safely added later as part of a more well-thought-out change.

amenonsen:
https://github.com/amenonsen/ansible/tree/pr-9130-rebase has everything but the inventory changes.

amenonsen:
@defionscode could you please confirm that the above tree is sufficient for your testing? (On IRC, @willthames said he didn't object to simplifying the patch by dropping the inventory changes.) I would like to see this PR merged without yet another round of delays (especially given how small the patch is, not to mention how many times it's been reinvented because people need it), so please let me know if there's anything else I can do to help.

amenonsen:
ping @defionscode

juliedavila:
I agree that it should all be driven by environment variables or the ini file.

amenonsen:
@defionscode sure, but is my tree sufficient for you to test this change and finally get it merged? That's what I was really pinging you about.

amenonsen:
OK, so I guess this has been abandoned again. Anyone else interested in looking at it?

juliedavila:
It's not abandoned and it's unfortunate that it comes across that way. This is a big project with a million moving pieces, even though the official roles of @gregdek and I at Ansible are not as maintainers, we try to help our open-source/core team as much as possible, but we also have to balance our activity here with our primary areas of responsibility. I hope that helps understand why sometimes things get looked at quicker/faster at times and much slower at other times; it is really isn't even indicative of abandonment or anything like that, again, apologies. Your branch looks fine to me, if you could make a PR and reference it here, that'd be great, I have some cycles to test today if that's possible for you.

amenonsen:
It comes across that way because you said thirteen days ago that you had cycles to test "this week", and two weeks went by without any response to my question of "is this enough to test". Given that this PR is a year old, has only a couple of dozen lines of code, and has been reinvented in various forms multiple times, I hope that helps understand why it feels like (just the latest) abandonment. (I'm not even trying to be sarcastic; I am taking it for granted that you have nothing but the best intentions, and I sincerely appreciate your offering to take a look at this PR. I am frustrated, of course, but I hope you will recognise that I'm trying to deal with it constructively.)

Resubmitted as requested: #12548.

juliedavila:
I totally get it, and I'm sure things will get better. @bcoca @gregdek this should be closed in favor of @amenonsen PR #12548

gregdek:
Yep, this is one of those cases where a PR touches multiple files, all of which are active and undergoing constant change, which complicates our ability to coordinate. I'll defer to @bcoca to close this one, as I'm trying not to touch active PRs in ansible/ansible directly at present.

abadger:
Thanks everyone who's been working on this both for working on it and for persevering through the frustrations with getting it reviewed and tested while the underlying code changes.  PR #12548 has been merged so I'm closing this one now.

PR #9136, opened 2014-09-25
Adding ip address to ansible-playbook RECAP

arxcruz:

### DIFF
--- a/bin/ansible-playbook
+++ b/bin/ansible-playbook
@@ -283,6 +283,9 @@ def main(args):
 
             for h in hosts:
                 t = pb.stats.summarize(h)
+                h_ip = pb.inventory.get_host(h)
+                if h_ip.vars.get('ansible_ssh_host'):
+                    h = '%s - %s' % (h, h_ip.vars['ansible_ssh_host'])
 
                 display("%s : %s %s %s %s" % (
                     hostcolor(h, t),


amenonsen:
This might've been a good idea, but the underlying code has changed considerably. @arxcruz if you are interested in making a similar change to current devel, please feel free to open a new PR; note that the changes should probably go in `lib/ansible/executor/playbook_executor.py` in the `run` method, though note the `FIXME` comment.

I think this PR should be closed now.

arxcruz:
Hi,

I'm not sure if the best place is in lib/ansible/executor/playbook_executor.py (actually, I'm not seeing this file). The host is just an string, and you can only get the ip address calling the pb.inventory.get_host(h) for that particular host. Which will return another string, I'm just concatenating here. If you want another PR because of the several changes in ansible, I can do it, but I'm not seeing a better place for this code.

gregdek:
@arxcruz it will need to be rebased, at the very least. So long as the PR applies cleanly, how you want to do it is up to you. Thanks!

amenonsen:
@arxcruz if you don't see `lib/ansible/playbook/playbook_executor.py`, you're looking in the wrong place. You should be looking at the v2 code, i.e. the `devel` branch of https://github.com/ansible/ansible (where there is no `bin/ansible-playbook`).

amenonsen:
@arxcruz Any updates here? Given that your changes no longer apply to the current code, I tend to think this PR should be closed now, but I'll leave it open for a while longer in case you're interested in updating it.

arxcruz:
@amenonsen I think you can close, I've been checking the new code, and now, only stats is being passed to callbacks and there I have only the host, so I don't have access to ip address from the inventory. Unless there's a method that I can call to get the ip address, like Inventory.get_stats('ssh_connect_host', host).

arxcruz:
Well, I decided to create a new PR on devel branch at https://github.com/ansible/ansible/pull/15152

amenonsen:
I'm closing this PR now, since it's been superseded by #15152.

PR #9138, opened 2014-09-25
Swap out rspec for vagrant in gem module example

meatballhat:
as installing vagrant from gem is very much deprecated.


meatballhat:
Looks like I need to submit this elsewhere now :smiley_cat:

PR #9145, opened 2014-09-25
Allow configuration of which etcd API version to use for lookups

jszwedko:
v1 of the etcd API is currently deprecated so we should allow the use of v2. I kept the default at v1 to preserve compatibility.

### DIFF
--- a/lib/ansible/runner/lookup_plugins/etcd.py
+++ b/lib/ansible/runner/lookup_plugins/etcd.py
@@ -28,10 +28,14 @@ ANSIBLE_ETCD_URL = 'http://127.0.0.1:4001'
 if os.getenv('ANSIBLE_ETCD_URL') is not None:
     ANSIBLE_ETCD_URL = os.environ['ANSIBLE_ETCD_URL']
 
+ANSIBLE_ETCD_API_VERSION = 'v1'
+if os.getenv('ANSIBLE_ETCD_API_VERSION') is not None:
+    ANSIBLE_ETCD_API_VERSION  = os.environ['ANSIBLE_ETCD_API_VERSION']
+
 class etcd():
     def __init__(self, url=ANSIBLE_ETCD_URL):
         self.url = url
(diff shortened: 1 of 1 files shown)


lib/ansible/runner/lookup_plugins/etcd.py
```diff
@@ -45,12 +49,16 @@ def get(self, key):
...
+            # v2: {"action":"get","node":{"key":"/name","value":"Jane Jolie","modifiedIndex":5,"createdIndex":5}}
             item = json.loads(data)
-            if 'value' in item:
-                value = item['value']
+
             if 'errorCode' in item:
                 value = "ENOENT"
+            elif ANSIBLE_ETCD_API_VERSION = 'v1'
+                value = item['value']
+            elif ANSIBLE_ETCD_API_VERSION = 'v2'
```
meatballhat: These `elif`s each need a `:` at eol.  Does this module have tests?  This should be a `SyntaxError` :scream_cat:

jszwedko: Oops! Note to self to not edit after testing. This module doesn't currently have any tests, but I'd be happy to try to add some.

jszwedko:
Anything else I need to do here to get this in?

jszwedko:
We may also want to start defaulting to v2 at this point rather than v1 (but still allow the override).

amenonsen:
@jszwedko what's the current situation with the API versions? Is it safe to default to v2 now? Is a v1 override still needed? Do you have any plans to rebase this PR to apply to current devel?

jszwedko:
@amenonsen I think it is safe to default to v2 now -- it's been awhile since I've been in this world though so it's probably better if someone takes this work over.

sivel:
@jszwedko,

Hi!

jszwedko:
Hi @sivel -- I'll close this since I no longer use etcd and ansible together.

PR #9155, opened 2014-09-25
Skip setting up delegate if using local transport

ahamilton55:
##### Issue Type:

Feature Pull Request
##### Ansible Version:

ansible 1.8 (devel 1970361) last updated 2014/09/25 10:53:31 (GMT -700)
##### Environment:

Running from: Mac OS X 10.9, Fedora 20
Running against: Debian 7, CentOS 6/7
##### Summary:

If using the "local" transport then there is no need to setup a delegation host. This "feature" skips the delegation setup when `self.transport` is set to "local." The local transport connection plugin sets itself up and does not require any information that is inside of the delegation variable.

This should should not be noticed by the user as nothing really changes, less work is just done.

### DIFF
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -840,7 +840,7 @@ class Runner(object):
         if self.delegate_to:
             self.delegate_to = template.template(self.basedir, self.delegate_to, inject)
 
-        if self.delegate_to is not None:
+        if self.delegate_to is not None and self.transport != 'local':
             delegate = self._compute_delegate(actual_pass, inject)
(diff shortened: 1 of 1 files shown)


bcoca:
i can see that i have a play with connection: local but want a task delegated to a monitoring machine over ssh, if I use delegate_to I would not want it to stay local

amenonsen:
I think this should be closed now. As @bcoca says, there are valid reasons to delegate to another host from a playbook that runs with «connection: local», and the underlying code has changed as well.

bcoca:
closing as per above

PR #9163, opened 2014-09-26
Add support for STS Tokens (aws_security_token) to all cloud modules

semifocused:
Extends get_ec2_creds() in lib/ansible/module_utils/ec2.py to include security token.

Modifies all cloud modules to respect security token and provide to service-specific boto connections as required.

These changes enable ansible's cloud modules to use temporary credentials provided as a result of assuming an IAM Role, rather than using IAM user credentials.

### DIFF
--- a/lib/ansible/module_utils/ec2.py
+++ b/lib/ansible/module_utils/ec2.py
@@ -150,7 +150,7 @@ def get_ec2_creds(module):
     ''' for compatibility mode with old modules that don't/can't yet
         use ec2_connect method '''
     region, ec2_url, boto_params = get_aws_connection_info(module)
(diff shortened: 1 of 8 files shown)


jimi-c:
Hi @semifocused, first off - this looks great.

One small change required is, if this is all ec2 cloud modules, the documentation for the new param should be added to the common document fragment, located in `lib/ansible/utils/module_docs_fragments/aws.py`. 

However, due to the fact that we're going to be moving the modules to new repos, you may have to split this up into two PRs - one for the changes to the core code (module_utils, etc.) and one for the modules themselves. So, I'm going to close this for now so that you can resubmit it later after the split.

If you have any questions related to the split, I'll be happy to answer them on the thread you started on ansible-project.

Thanks!

PR #9170, opened 2014-09-26
yum: add update_cache feature

resmo:
Yum does not always update to latest package version unless metadata cache has expired. Using makecache we ensure the metadata cache has been updated.


resmo:
PR moved to https://github.com/ansible/ansible-modules-core/pull/9

Issue #9176, opened 2014-09-26
ec2_asg: boto objects don't JSON stringify


tphummel:
##### Issue Type:

“Bug Report”
##### Ansible Version:

ansible 1.8 (devel 6af5455edc) last updated 2014/09/26 13:56:36 (GMT -700)
##### Environment:

running from osx 10.9.4. controlling ubuntu 14.04
##### Summary:

boto Tag objects do not JSON encode. 
##### Steps To Reproduce:

https://gist.github.com/tphummel/7206f1842e4b26d8d02b
##### Expected Results:

successful run of play and output json result without error
##### Actual Results:

LC, ASG action complete but play fails trying to encode `asg_properties` as JSON, specifically the `tags` list in this example. 

`TypeError: Tag(Name=my-custom-tag-value) is not JSON serializable`

here is a dump of the entire `asg_properties` prior to failure: 

```
{'availability_zones': [u'us-west-2c'],
 'default_cooldown': 300,
 'desired_capacity': 1,
 'health_check_period': 60,
 'health_check_type': u'ELB',
 'healthy_instances': 1,
 'in_service_instances': 1,
 'instance_facts': {u'i-87eb6a88': {'health_status': u'Healthy',
                                    'launch_config_name': u'lc-111111',
                                    'lifecycle_state': u'InService'}},
 'instances': [u'i-87eb6a88'],
 'launch_config_name': u'lc-111111',
 'load_balancers': [u'lb-test'],
 'max_size': 1,
 'min_size': 1,
 'name': u'health-server-group',
 'pending_instances': 0,
 'placement_group': None,
 'tags': [Tag(Name=test-server)],
 'terminating_instances': 0,
 'termination_policies': [u'Default'],
 'unhealthy_instances': 0,
 'viable_instances': 1,
 'vpc_zone_identifier': u'subnet-1111111'}
```

the `tags` key is type `<class 'boto.resultset.ResultSet'>`. and the value at index 0 is type `<class 'boto.ec2.autoscale.tag.Tag'>`

I can skirt the error by doing `asg_properties.pop('tags', None)` on the line prior to `module.exit_json` at the bottom of `library/cloud/ec2_asg`

mpdehaan:
Can you please include the full traceback from ansible/boto if available?

Thanks!

mpdehaan:
Also: can you share the ansible output in context so it's clear exactly what task failed?  (like the TASK header above the traceback)

Also also thanks!

[...]

PR #9177, opened 2014-09-26
ec2_asg: remove ‘tags’ key from asg_properties before exit_json

tphummel:
I am _NOT_ proposing this as the solution to #9176, merely as a reference of the stopgap I'm using


mpdehaan:
Thanks for this reference.

As I understand this is _not_ the fix we should apply here, I'm going to close this ticket so we don't accidentally apply it, but it will remain linked in the ticket.

Hope that makes sense and feel free to comment on the other ticket if you disagree.

Much appreciated!

PR #9190, opened 2014-09-28
Fix link in reST template

martin-ueding:
There was a missing trailing underscore (`_`) that would have marked the
content in the backticks as a link. This adds it and fixes the link on
every core module page.


mpdehaan:
Merged, thanks!  This will show up when I next trigger the docs build, which will be very soon.

Much appreciated!

PR #9191, opened 2014-09-28
Expand central consistency in openstack modules

emonty:
Module docstrings and argument specs have things in common. Make sure
that the ansible repo has the things we want so that the modules
can consume them.


mpdehaan:
Thanks for this!

Just to confirm, do you agree that these are already supported by all the modules and we don't need to merge anything else PR-related first?

emonty:
Yes. I confirm that these are already supported by the mdules.

bcoca:
all of the quantum_\* and the glance_image modules seem to have the endpoint_type, but the nova_ and keystone_user do not.

PR #9192, opened 2014-09-28
docs: Fix code block in Installation

Nowaker:
Broken at http://docs.ansible.com/intro_installation.html#running-from-source.


mpdehaan:
Thanks for this one, looks like we got two PRs at the same time, so I've already merged in a fix for that one.

Sorry about duplicate efforts - contribution is much appreciated!

PR #9206, opened 2014-09-29
with_items maps back into a single module call for authorized_key module

kustodian:
This pull request is needed for the `clear_keys` parameter pull request for the `authorized_key` module https://github.com/ansible/ansible-modules-core/pull/42

### DIFF
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -672,7 +672,7 @@ class Runner(object):
             if type(items) != list:
                 raise errors.AnsibleError("lookup plugins have to return a list: %r" % items)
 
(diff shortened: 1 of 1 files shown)


kustodian:
We should probably think about making this possible for any module by passing a variable or something, so that future modules don't need to modify core to be able to receive `with_items` elements as a map.

jimi-c:
@kustodian that has been discussed, and we'd be open to reviewing that for sure. Thanks!

kustodian:
That is great news. If you are up for it, I would be interested to make a pull request.

BitOne:
:+1: 
Very much needed for security reason (guy leaving your company, you want to make sure that the present keys are the one valid).

Thanks !

amenonsen:
I think this should be closed now. The PR that depended on it was closed months ago.

kustodian:
Closing this, since it's not needed anymore.

PR #9210, opened 2014-09-30
Allow some ansible libraries to be imported on Windows

willthames:
The intention is to allow ansible-inventory-grapher to run on
windows. Windows-specific issues with other parts of the codebase
definitely still exist

Remove unix specific dependencies from ansible.utils to ansible.utils.unix
Removed unwanted pwd import from ansible.constants
Import display_function in ansible.utils only when required
Fixed broken path construction for loading host and group vars
Handling filenames that windows doesn't like (typically globs
for hosts in inventory - e.g. web*)

There's a weird copyright conundrum as a result of this change, in terms
of attribution for the additional python module. git blame suggests that
@tbielawa contributed the functionality and so I've credited him in the
copyright for the file.

### DIFF
--- a/lib/ansible/utils/unix.py
+++ b/lib/ansible/utils/unix.py
@@ -1,0 +1,32 @@
+# (c) 2012-4, Tim Bielawa <tbielawa@redhat.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+import sys
+import termios
+import tty
+
+
+def getch():
+    ''' read in a single character '''
+    fd = sys.stdin.fileno()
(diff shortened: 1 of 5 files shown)


lib/ansible/utils/unix.py
```diff
@@ -0,0 +1,32 @@
+# (c) 2012-4, Tim Bielawa <tbielawa@redhat.com>
```
timlnx, line 1: I did indeed contribute this code for the pause plugin. It is very very *nix specific.

Thanks for thinking of me!

Just a bit of nit-picking about the copyright header (feel free to ignore this, I won't be offended)

http://www.gnu.org/licenses/gpl-howto.html - "How to use GNU licenses for your own software"

"For software with several releases over multiple years, it's okay to use a range (“2008-2010”) instead of listing individual years (“2008, 2009, 2010”) if and only if every year in the range, inclusive, really is a “copyrightable” year that would be listed individually; and you make an explicit statement in your documentation about this usage.

Always use the English word “Copyright”; by international convention, this is used worldwide, even for material in other languages. The copyright symbol “©” can be included if you wish (and your character set supports it), but it's not necessary. There is no legal significance to using the three-character sequence “(C)”, although it does no harm."

See also: http://www.law.cornell.edu/uscode/text/17/401 "17 U.S. Code § 401 - Notice of copyright: Visually perceptible copies"

amenonsen:
This PR is, of course, outdated now by the devel→v2 migration (though the comments about copyright notices still apply, though I disagree vehemently with the whole "multiple years" thing; note that 17 USC §401 doesn't say anything about this). Ahem.

I think this PR should be closed, and if necessary a new issue/PR opened for windows compatibility problems in the current codebase.

timlnx:
@amenonsen @willthames is there anything I need to/can do there? I'm not sure what's going on at this point anymore :)

PR #9212, opened 2014-09-30
Now to copy content must define a filename not a directory

ethervoid:
To copy content you must define the filename instead of create a temp file.


amenonsen:
This can be closed, I've submitted an updated version.

bcoca:
closed as per merged #11732

PR #9213, opened 2014-09-30
Adding readlink to filters ?

i5513:
Hello,

It is useful at least for installing software to /opt/xxx, where /opt/xxx-version is the real directory and /opt/xxx is a symlink

At ansible host:
ls -l
2.4/
latest -> 2.4

So you can play with latest and readlink in your playbook

Thanks

### DIFF
--- a/lib/ansible/runner/filter_plugins/core.py
+++ b/lib/ansible/runner/filter_plugins/core.py
@@ -30,6 +30,7 @@ from ansible.utils import md5s
 from distutils.version import LooseVersion, StrictVersion
 from random import SystemRandom
 from jinja2.filters import environmentfilter
+from os import readlink
 
 
 def to_nice_yaml(*a, **kw):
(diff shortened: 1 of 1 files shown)


jimi-c:
Hi @i5513, the addition of new filters also requires an update to the corresponding docs page. Could you please add that as well?

abadger:
@i5513 Hey, taking a look at disentangling this... It looks like between the time you opened this and now we added realpath as a filter: https://github.com/ansible/ansible/blob/devel/lib/ansible/runner/filter_plugins/core.py#L299  (I remember that because 1.8.x has realpath but not readlink and someone asked that the documentation be changed to mention realpath... I should have explored this further when I did that :-(

Does realpath suit your needs or is there a difference between realpath and readlink that means we want to have both?

amenonsen:
@abadger, I looked at the realpath documentation, and it uses readlink to follow symlinks. There's no reason to include both, and this PR can be closed.

i5513:
Sorry for do not reply on time, and thanks for incluye this feature.

PR #9215, opened 2014-09-30
Update intro_configuration.rst

### LINKED ISSUE #7418: Wrong sudo_flags used (or error in docs)
##### Issue Type:

Documentation Report
##### Ansible Version:

ansible 1.7 (ec2_hostname_variable 172fdf9e6a) last updated 2014/05/14 12:57:21 (GMT +300)
##### Environment:

Ubuntu 13.10
##### Summary:

The [docs](http://docs.ansible.com/intro_configuration.html#sudo-flags) say that `sudo_flags` defaults to preserving the environment, but the flag actually passed is `-H`, which doesn't do that at all. The correct flag is `-E`.

Is this an error in the docs or a bug?
##### Steps To Reproduce:

`cat docsite/rst/intro_configuration.rst  | grep environment | grep preserve`

`man sudo | grep environment | grep preserve`
##### Expected Results:

The two docs agree with each other.
##### Actual Results:

They do not.

nmadhok:
Fixes https://github.com/ansible/ansible/issues/7418

### DIFF
--- a/docsite/rst/intro_configuration.rst
+++ b/docsite/rst/intro_configuration.rst
@@ -495,9 +495,10 @@ the sudo implementation is matching CLI flags with the standard sudo::
 sudo_flags
 ==========
 
-Additional flags to pass to sudo when engaging sudo support.  The default is '-H' which preserves the home directory
-of the original user.  In some situations you may wish to add or remote flags, but in general most users
-will not need to change this setting::
+Request that the security policy set the HOME environment variable to the home directory specified by the
+target user's password database entry. The default is '-H' which preserves the home directory of the original
+user.  In some situations you may wish to add or remove flags, but in general most users would not need to
+change this setting::
 
    sudo_flags=-H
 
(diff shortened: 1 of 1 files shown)


jimi-c:
Hi @nmadhok, could we change this to correspond to what's in the sudo man page more closely? For example, the man page on my system says: 

<pre>
Request that the security policy set the HOME environment variable to 
the home directory specified by the target user's password database entry.  
Depending on the policy, this may be the default behavior.
</pre>

nmadhok:
@jimi-c This can be merged now. Made the requested changes.

nmadhok:
@jimi-c This is another one!

jimi-c:
Hi @nmadhok , sorry for being unclear on this one. For this section, the topic is sudo flags in general, and not just `-H`. I'm fine if you make the language regarding `-H` line up with the man page, however some of the previous text needs to be preserved as well.

nmadhok:
@jimi-c It's not just `-H` but for sudo flags in general. The only text that was removed is `Additional flags to pass to sudo when engaging sudo support`. Do you want me to include that as well?

amenonsen:
This PR can be closed, the wording in devel has changed and mentions $HOME rather than "the environment". (But note that the wording in devel is not what @jimi-c asked for above.)

PR #9220, opened 2014-09-30
Resolve issue where repo_path contains multiple '.git' strings

jlaska:
If the repo_path contained multiple '.git' strings, the _git_repo_info()
call resulted in a traceback.  This change removes the trailing '.git'
and resolves the traceback.


jimi-c:
Merged, thanks!

PR #9225, opened 2014-09-30
AIX unprivileged facts collection

qvicksilver:
Allows to collect facts on an AIX machine as an unprivileged user and also fetches some more network interface facts.

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -186,15 +186,20 @@ class Facts(object):
         if self.facts['system'] == 'Linux':
             self.get_distribution_facts()
         elif self.facts['system'] == 'AIX':
-            try:
-                rc, out, err = module.run_command("/usr/sbin/bootinfo -p")
+            # Attempt to use getconf to figure out architechture
+            # fall back to bootinfo if needed
+            if module.get_bin_path('getconf'):
+                rc, out, err = module.run_command([module.get_bin_path('getconf'),
+                                                   'MACHINE_ARCHITECTURE'])
                 data = out.split('\n')
                 self.facts['architecture'] = data[0]
-            except:
-                self.facts['architecture'] = 'Not Available'
+            else:
+                rc, out, err = module.run_command([module.get_bin_path('bootinfo'),
+                                                   '-p'])
+                data = out.split('\n')
+                self.facts['architecture'] = data[0]
         elif self.facts['system'] == 'OpenBSD':
             self.facts['architecture'] = platform.uname()[5]
-
 
     def get_local_facts(self):
 
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -2278,6 +2278,26 @@ class AIXNetwork(GenericBsdIfconfigNetwork, Network):
     """
     platform = 'AIX'
 
+    def get_default_interfaces(self, route_path):
+        netstat_path = module.get_bin_path('netstat')
+
+        rc, out, err = module.run_command([netstat_path, '-nr'])
+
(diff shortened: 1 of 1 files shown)


zitterbacke:
:+1:

amenonsen:
@qvicksilver any chance you could rebase this to apply to current devel? It looks easy-ish, but I'm not in a position to test on AIX, so it would be best if you could do it.

qvicksilver:
I could to the rebase but I have no way of testing it. I've cancelled the
co-location of the AIX box I had. Tried getting someone from ##aix on
freenode to help me out but got zero response.

On Mon, Jul 27, 2015 at 4:42 AM, Abhijit Menon-Sen <notifications@github.com

amenonsen:
@qvicksilver thanks for making the effort to find testers. @bmanojlovic any chance you could help with testing this on AIX if it's rebased? (I ask only because you have the other open PR for AIX support.)

Otherwise I guess this PR should be closed.

bmanojlovic:
i did not know about this PR, so i have few network patches applied already in meantime
https://github.com/ansible/ansible/pull/10420/files
here is output on AIX of few mentioned commands (and failed ones because of no permission)

```
$ /usr/sbin/bootinfo -p
ksh: /usr/sbin/bootinfo: 0403-006 Execute permission denied.
$ getconf MACHINE_ARCHITECTURE
chrp
$ netstat -nr
Routing tables
Destination        Gateway           Flags   Refs     Use  If   Exp  Groups

Route Tree for Protocol Family 2 (Internet):
default            10.222.222.190    UGS      11 2099310322 en0      -      -   
10.222.222.160     10.222.222.181    UHSb      0         0 en0      -      -   =>
10.222.222.160/27  10.222.222.181    U         1         1 en0      -      -   
10.222.222.181     127.0.0.1         UGHS      0     14620 lo0      -      -   
10.222.222.191     10.222.222.181    UHSb      2     26656 en0      -      -   
127/8              127.0.0.1         U        22   3216712 lo0      -      -   

Route Tree for Protocol Family 24 (Internet v6):
::1%1              ::1%1             UH        2       596 lo0      -      -   
$ route -n  get 8.8.8.8
ksh: route: 0403-006 Execute permission denied.
$ 
```

amenonsen:
OK, so we know that `getconf MACHINE_ARCHITECTURE` works, and `netstat -nr` works. @qvicksilver, please do rebase your patches to apply to current devel and refresh this PR with "git push -f". @bmanojlovic thanks very much for your help with testing, it's greatly appreciated.

Note: in the second hunk of the second patch, it says `/usr/bin/entstat`. I'm guessing that's a typo?

amenonsen:
Hm, no. `entstat` seems to not be a typo, but as mentioned earlier, #10240 already adds some code to do network facts discovery on AIX. So the second patch in this PR is only partially relevant and will need to be adapted.

qvicksilver:
There, now its rebased

amenonsen:
A couple of quick comments:
1. Before doing `out.split`, I think you should check that the command has not failed. Either test that `rc == 0` or (perhaps better, since you can't test directly) that `out is not None`. I know the `bootinfo -p` code didn't do it, and plenty of other code in facts.py doesn't do it, but still. Better to be defensive.
2. Note misspelling of architecture in the comment in the first commit.

amenonsen:
OK, it looks like `out` can't be None. `stdin` and `stdout` start out as empty strings, and only ever have data appended to them in `run_command`. So it should be safe to split straightaway (though there's plenty of code that checks `if out:` too). So you can ignore that comment or add an extra guard, as you see fit.

qvicksilver:
I remember checking if out could be None and decided that splitting the empty string was OK so I'll leave that be. The typo is fixed.

PR #9229, opened 2014-10-01
broken link in developing_modules.rst

wjimenez5271:
link to core modules source is broken since restructuring of public ansible repos


jimi-c:
Merged, thanks!

PR #9230, opened 2014-10-01
add 'description' parameter.

shirou:
PROPOSAL: 'description' parameter

I'd like to add a "description" parameter.
Ansible does nothing about on it. it just describing what is it.

I know YAML can have a comment and a name is good to know what this task
do. But a comment is not structured and a name can not be long because
handlers use it.

If ansible defines "description", a documentation tool such as
Sphinx can read it and produces a good document.

```

---
- hosts: all
  description: This playbook is an example for describe directive
  tasks:
    - name: create empty file
      file: path=/tmp/empty state=touch
      description: >
        This empty file is needed to later tasks,
        such as A, B, and C.
    - name: do some script
      script: /some/local/script.sh
      description:
        - a
        - b
        - c
    - script: /some/local/script.sh
      description:
        can: use map
        some: thing
        other: thing

- include: other.yml
  description:
    - include can have description also
```

I will send this proposal to the ansible-project mailing list.

### DIFF
--- a/lib/ansible/playbook/play.py
+++ b/lib/ansible/playbook/play.py
@@ -38,7 +38,7 @@ class Play(object):
        'accelerate_port', 'accelerate_ipv6', 'sudo', 'sudo_user', 'transport', 'playbook',
        'tags', 'gather_facts', 'serial', '_ds', '_handlers', '_tasks',
        'basedir', 'any_errors_fatal', 'roles', 'max_fail_pct', '_play_hosts', 'su', 'su_user',
-       'vault_password', 'no_log',
+       'vault_password', 'no_log', 'description',
     ]
 
     # to catch typos and so forth -- these are userland names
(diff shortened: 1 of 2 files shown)


bcoca:
you already have 'name' at play level, why not use that?

## 

Brian Coca
Pedo mellon a minno

shirou:
When you play, 'name' is always printed. If name is very long, output becomes too messy.

amenonsen:
I think this should be closed now. There doesn't seem to have been much support for the directive-as-comment proposal, and I don't think it's a great precedent either.

bcoca:
closing, as stated above a comment will do the same thing as a noop directive

shirou:
I still think some kind of text notation is requred to generate a document automatically. But also know this idea is not supported widely. So, I agree to close. 
Thank you.

PR #9232, opened 2014-10-01
Unarchive no longer transfers archive if creates= check succeeds

rdolejsi:
Unarchive no longer transfers the archive in case the creates= parameter yields a successful match, thus preventing lenghty operations over the network (especially with large archive distributions).

### DIFF
--- a/lib/ansible/runner/action_plugins/unarchive.py
+++ b/lib/ansible/runner/action_plugins/unarchive.py
@@ -49,6 +49,7 @@ class ActionModule(object):
         source  = options.get('src', None)
         dest    = options.get('dest', None)
         copy    = utils.boolean(options.get('copy', 'yes'))
+        creates = options.get('creates', None)
 
         if source is None or dest is None:
             result = dict(failed=True, msg="src (or content) and dest are required")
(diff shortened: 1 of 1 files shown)


rdolejsi:
Please note: the unpack module on the client is actually called twice to achieve this - with extra signalling using precheck=True/False parameter. Please suggest other form of signalling if this is not the correct way of doing it.

rdolejsi:
This is a replacement request for the closed pullreq #9139 (no changes, update to current HEAD).

This is a half of the patch, complementary to modules-core pullreq http://github.com/ansible/ansible-modules-core/pull/100. Both pullreqs are required.

abadger:
Oops, I think we merged a similar request in November: https://github.com/ansible/ansible-modules-core/pull/124

Note: I just fixed a bug in that code https://github.com/ansible/ansible/commit/f1fdddb640e1dfcd8fdc930031db267947d8cb70 that causes creates= to be ignored if unarchive was fed its arguments in complex args format instead of key=value format.

Do those commits seem to address everything that you had or is there something further we should add?

abadger:
Commit that implemented the alternative to this: https://github.com/ansible/ansible/commit/e79acf3f882a2f9d46a44df500ce9cd9955ca227

# Possible Duplicated Efforts

Hi!

- https://github.com/ansible/ansible/pull/9261

PR #9234, opened 2014-10-01
Update developing_modules.rst

gamename:
The existing description does not make 2 things clear.
1) This is an environment variable
2) It must be set on the _controlling_ host, not the remote host.


jimi-c:
Merged, thanks!

PR #9237, opened 2014-10-02
Add doc of @filename in intro_patterns.rst

jkleckner:
I noticed that the @filename notation for the --limit argument was not
conveniently described.

This patch adds a simple example to the intro_patterns.rst file.


mpdehaan:
Merged, thank you!

Issue #7251, opened 2014-05-01
Setting ssh_args disables pipelining mode incorrectly


axelson:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.5.5
##### Environment:

Running Ansible on Mac OS X Mavericks and managing SmartOS (Solaris).
##### Summary:

SSH Pipelining [1] doesn't work if sudo_user is set. If I comment out the sudo_user and sudo then the pipelining works.

[1] http://docs.ansible.com/intro_configuration.html#pipelining
##### Steps To Reproduce:

Here's my simple.yml file:

```

---
- hosts: all
  sudo: yes
  sudo_user: jason
  tasks:
    - name: git
      git: repo=https://github.com/ansible/ansible.git dest=~/ansible update=yes force=no
      environment: env_var
```

ansible-playbook -e "pipelining=True" simple.yml

With sudo_user and sudo and running with -vvv I can see that Ansible is running separate mkdir and chmod commands over ssh.

env_var is needed to set PATH for git
##### Expected Results:

I expect ansible to not send separate SSH commands for mkdir and chmod. 
##### Actual Results:

SSH Pipelining is working so the ssh command for mkdir and chmod are combined.

mpdehaan:
Can you please show the output from -vvv ?

Thanks!

axelson:
Here you go (I've only changed the hostname):

```
PLAY [middle_tier] ************************************************************ 

GATHERING FACTS *************************************************************** 
<example.com> ESTABLISH CONNECTION FOR USER: jason
<example.com> REMOTE_MODULE setup
<example.com> EXEC ['ssh', '-C', '-tt', '-q', '-o', 'ControlMaster=auto', '-o', 'ControlPersist=60s', '-o', 'ControlPath=/Users/jason/.ansible/cp/ansible-ssh-%h-%p-%r', '-o', 'KbdInteractiveAuthentication=no', '-o', 'PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey', '-o', 'PasswordAuthentication=no', '-o', 'ConnectTimeout=10', 'example.com', "/bin/sh -c 'mkdir -p /tmp/ansible-tmp-1398987879.46-122273948422799 && chmod a+rx /tmp/ansible-tmp-1398987879.46-122273948422799 && echo /tmp/ansible-tmp-1398987879.46-122273948422799'"]
[...]

PR #9246, opened 2014-10-02
Warn that ssh_args is incompatible with pipelining, note such in documentation

dpkirchner:
Documents the behavior reported in https://github.com/ansible/ansible/issues/7251 (which appears to be deliberate based on a read of ssh.py).


mpdehaan:
Hi,

I would rather we attempt to fix this bug instead of updating the docs, so I'm going to close this ticket.

PR #9249, opened 2014-10-02
Added a note about escaping backreferences when using 'regex_replace' filter

tjanez:
Users will often be puzzled why 'regex_replace' is not working as intended when used inside
YAML arguments. This note explains what they have to do to get it working.

### DIFF
--- a/docsite/rst/playbooks_variables.rst
+++ b/docsite/rst/playbooks_variables.rst
@@ -355,6 +355,9 @@ To replace text in a string with regex, use the "regex_replace" filter::
     # convert "foobar" to "bar"
     {{ 'foobar' | regex_replace('^f.*o(.*)$', '\\1') }}
 
+.. note:: If "regex_replace" filter is used with variables inside YAML arguments (as opposed to simpler 'key=value' arguments),
+   then you need to escape backreferences (e.g. ``\\1``) with 4 backslashes (``\\\\``) instead of 2 (``\\``).
+
 A few useful filters are typically added with each new Ansible release.  The development documentation shows
 how to extend Ansible filters by writing your own as plugins, though in general, we encourage new ones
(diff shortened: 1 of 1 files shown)


mpdehaan:
I'm having trouble understanding this example, though I also also never use regex_replace.  I think this might be better served by a positive example, i.e. showing how to do it maybe?

tjanez:
Sure, I can give a positive example.

Currently, the documentation under subsection [Other Userful Filters](http://docs.ansible.com/playbooks_variables.html#other-useful-filters), doesn't list full examples (e.g. concrete task definitions), but rather just shows uses of Jinja2 variables with certain filters. That's why I didn't list a full example in the Note.

Anyway, here is an example that shows a task that uses the `regex_replace` filter.
1. When using `key=value` arguments, backreferences need to be escaped with two backspaces (`\\`):
   
   ```
   - name: create directories for myapps
     file: path=/path/to/myapps/{{ item | regex_replace('^myapp-(.*)$', '\\1') }} owner=root group=root mode=0755 state=directory
     with_items:
       - myapp-dashboard
       - myapp-analysis
       - myapp-database
       - myapp-results
   ```
2. When using so called _YAML_ arguments, backreferences need to be escaped with four backspaces (`\\\\`):
   
   ```
   - name: create directories for myapps
     file:
       path: /path/to/myapps/{{ item | regex_replace('^myapp-(.*)$', '\\\\1') }}
       owner: root
       group: root
       mode: 0755
       state: directory
     with_items:
       - myapp-dashboard
       - myapp-analysis
       - myapp-database
       - myapp-results
   ```

Do you prefer such an example in the Note? Or in the short paragraph about `regex_replace`?

tjanez:
Ping?

PR #9250, opened 2014-10-02
Add readlink filter doc

i5513:
As requested at request pull #9213, I'm proposing updathing this doc. Do you have man pages where I can update this too ?

Thanks


mpdehaan:
Merged, thanks!

(We don't have manpages for filters, so this should cover it)

PR #9283, opened 2014-10-07
Allow modules to provide useful information to --diff

kvdveer:
`--diff` is a very useful feature, but right now it is only available for templates and copies. However, lots of modules would benefit from being allowed to report their differences. Imagine a check/diff run telling you which packages would be installed by apt/yum etc.

This patch allows modules to return a diff value, which gets passed to the relevant callbacks.

### DIFF
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -543,7 +543,7 @@ class Runner(object):
         data = utils.parse_json(res['stdout'], from_remote=True, no_exceptions=True)
         if 'parsed' in data and data['parsed'] == False:
             data['msg'] += res['stderr']
-        return ReturnData(conn=conn, result=data)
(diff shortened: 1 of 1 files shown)


amenonsen:
This is pretty neat, and should be revisited now that diff is being implemented in v2.

mgedmin:
My experiments with current `devel` show that this is no longer necessary: modules can use, e.g.

``` python
module.exit_json(
    ...
    changed=True,
    diff=dict(
        before_header='currently installed packages',
        after_header='packages to be installed',
        before='old\nvalue\n',
        after='new\nvalue\n',
    ),
    ...
)
```

and invoking `ansible-playbook -C --diff` will show the appropriate diff.

bcoca:
closing as per above, this is implemented in current devel

PR #9290, opened 2014-10-08
Do not use 'remote_user' in the variable passing with include statements example.

tjanez:
Previous documentation might erroneously lead readers into assuming that
passing a variable named 'remote_user' with an include statement will set the
'remote_user' parameter to the given value on all the included tasks.
This is not the case, so it is better to use another variable name in the
example.
Also, using 'user' for the variable name is consistent with the examples
showing passing variables with a different syntax.

### DIFF
--- a/docsite/rst/playbooks_roles.rst
+++ b/docsite/rst/playbooks_roles.rst
@@ -85,7 +85,7 @@ which also supports structured variables::
 
       - include: wordpress.yml
         vars:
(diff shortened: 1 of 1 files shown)


tjanez:
This PR is obsolete, since the problem was fixed (in a similar way) in 8e2a8c92ac7c4134f822ec77de08a1106da36fc6.

PR #9303, opened 2014-10-10
Use https for git submodules

willthames:
It's much more likely that people can use the https protocol
than the git protocol (many firewalls block the latter)


mpdehaan:
Merged, thanks!

PR #9304, opened 2014-10-10
Give useful errors when rsync or ssh are not installed on the system.

ShawnMilo:
I wasted way too much time today trying to figure out "file not found" errors. Especially with synchronize, since there is no traceback provided.

The problem was discovered running Ansible from within Docker. Docker images tend to be fairly stripped down, and their use is increasing. Ansible also seems to be the preferred configuration tool to be paired with Docker.

### DIFF
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -224,7 +224,13 @@ class Runner(object):
                 self.transport = "paramiko"
             else:
                 # see if SSH can support ControlPersist if not use paramiko
-                cmd = subprocess.Popen(['ssh','-o','ControlPersist'], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+                try:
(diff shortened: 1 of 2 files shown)


amenonsen:
Although I'm sympathetic to the idea of this PR (I'm always in favour of "useful errors when ..."), the first part seems clearly _wrong_: the code that runs ssh is trying to decide whether to use ssh or paramiko, and making it exit if it can't use ssh is obviously counterproductive (nor is it the only place where we try to run ssh). I also think the no-rsync-installed error should be handled inside the synchronize module, not in the caller. I think this PR should be closed now.

@ShawnMilo, if you feel like rebasing to devel and submitting a new more complete/correct PR, that would be welcome.

bcoca:
closing this PR as it does not apply anymore and as stated above 1/2 of it was making a wrong assumption. As for syncrnize it states that it depends on rysnc already in the docs.

PR #9307, opened 2014-10-10
Rackspace inventory: add support for new environment variable RAX_PRIVATE_IP

angstwad:
Add support for new environment variable, RAX_PRIVATE_IP which will the ansible_ssh_host value to the private IP of the instances.  Additionally removes redundant parenthesis and is a redo of the original documentation.

```
RAX_PRIVATE_IP=yes rax.py --list
```

Additionally, some changes occurred at Rackspace where new customers may not have all endpoints for servers in all data centers, despite pyrax reporting that they do.  This is part of a larger issue that is resolved by the second commit, bd323d0

### DIFF
--- a/plugins/inventory/rax.py
+++ b/plugins/inventory/rax.py
@@ -104,6 +104,7 @@ import os
 import re
 import sys
 import argparse
+import warnings
 import collections
 
 from types import NoneType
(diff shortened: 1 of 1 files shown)


sivel:
I'm thinking that I might prefer seeing this functionality in a slightly different form.

Perhaps where we can have the user provide what network name, and what IP version.

An example might be:
    RAX_ACCESS_NET=private
    RAX_IP_VER=4

We use this info to do something like

``` python
network = os.getenv('RAX_ACCESS_NET', 'public')
ip_version = int(os.getenv('RAX_IP_VER', 4))
addresses = pyrax.addresses.get(network, [])
address = [a['addr'] for a in addresses if a['version'] == ip_version][0]
```

sivel:
I'm also wondering if it isn't time to support a config file for the inventory plugin. We are starting to build quite a number of env vars. env vars should stay, but a conf file might be useful.

sivel:
This PR can be closed now that #9530 was merged.

PR #4098, opened 2013-09-13
Check for link in path in utils.path_dwim_relative()

jimi-c:
This allows an included path to be relative to a directory, if the
basedir of the original path is a symlink. In that case, the path
is normalized and the file is searched for in the normalized path.

Fixes #3312

PR #9309, opened 2014-10-10
If included file is a symlink, set basedir in utils.path_dwim_relative() to the real path

iljabauer:
If a role has following code in the main.yml:

```
- include: test.yml
```

and test.yml is a symlink and has following content:

```
- template: src=test.j2 dest=tmp
```

The file will not be found because the file will be searched in the folder of the main.yml.

With this change the file will be searched in the real folder of the file - like this is done with directories in https://github.com/ansible/ansible/pull/4098

### DIFF
--- a/lib/ansible/utils/__init__.py
+++ b/lib/ansible/utils/__init__.py
@@ -347,6 +347,10 @@ def path_dwim_relative(original, dirname, source, playbook_base, check=True):
     if os.path.islink(basedir):
         basedir = unfrackpath(basedir)
         template2 = os.path.join(basedir, dirname, source)
+    elif os.path.islink(original):
+        linked_file = unfrackpath(original)
+        basedir = os.path.dirname(linked_file)
+        template2 = os.path.join(basedir, "..", dirname, source)
     else:
         template2 = os.path.join(basedir, '..', dirname, source)
     source2 = path_dwim(basedir, template2)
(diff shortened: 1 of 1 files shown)


amenonsen:
I can confirm that this behaviour exists in current devel, but the patch no longer applies. `path_dwim_relative` has moved from `utils/__init__.py` to `parsing/__init__.py`, but the code has changed significantly.

@iljabauer any chance you could revise this to apply to current devel?

iljabauer:
Yes, I will have a look in to the new code

iljabauer:
I looked at the new code. The fix is not so easy anymore, because path_dwim_relative doesn't get the original file path anymore. This means that inside the method, we don't know that the file is imported from a symlinked file.

This information is now stored in ActionModule._task._task_include. I can pass this information inside the path_dwim_relative as optional named parameter, but then we would have to do this in every place where we need this behaviour.

It would be easier if we pass this information to the DataLoader class, than its accessible for every module which uses path_dwim_relative.

I can do this, but we have to decide, which solution is more preferable.

jimi-c:
Any follow-ups on this? There were recent changes to dwim functions to make improvements, so wondering if this is still an issue.

cc @bcoca

bcoca:
sounds like we just need to set basedir to the original path instead of the resolved one?

iljabauer:
I implemented a solution. It is working for static and dynamic includes. Tests are not completely ready yet.
I can't update the pull request, because I deleted my fork a while ago. Should I open a new pull request?

bcoca:
so we seem to have fixed this at one point, it is working in current devel. So i'm closing this issue.

PR #9310, opened 2014-10-10
Fixed hostname matching with pattern having '~' at the beginning

sh-vi:
**Issue Type:**
Bugfix Pull Request

**Ansible Version:**
- ansible 1.7.1
- ansible 1.7.2

**Environment:**
Fedora release 20 (Heisenbug)

**Summary:**
After using match() instead of search() for pattern matching ansible doesn't match hostnames with pattern that started with '~'.

**Steps to reproduce**:

```
$ cat inventory 
foo
bar
foobar
barfoo
```

**Expected Results:**

```
$ ansible '~bar' -i inventory --list-hosts
    bar
    foobar
    barfoo
```

**Actual Results:**

```
$ ansible '~bar' -i inventory --list-hosts
    bar
    barfoo
```

### DIFF
--- a/lib/ansible/inventory/__init__.py
+++ b/lib/ansible/inventory/__init__.py
@@ -162,14 +162,14 @@ class Inventory(object):
         results = []
         try:
             if not pattern_str.startswith('~'):
-                pattern = re.compile(fnmatch.translate(pattern_str))
+                pattern = re.compile('^' + fnmatch.translate(pattern_str))
             else:
                 pattern = re.compile(pattern_str[1:])
         except Exception, e:
             raise errors.AnsibleError('invalid host pattern: %s' % pattern_str)
 
         for item in items:
-            if pattern.match(getattr(item, item_attr)):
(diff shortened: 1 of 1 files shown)


mpdehaan:
We want to tread carefully here as reverting this would seemingly open up other bug reports linked to those commits.

Perhaps another fix is in order.

amenonsen:
I can confirm that devel still lists only bar and barfoo for a pattern like '~bar', and that seems wrong according to the documentation: «Most people don’t specify patterns as regular expressions, but you can. Just start the pattern with a ‘~’».

But I don't understand the match vs. search discussion and context, nor do I have any idea what "another fix" might look like in this case. Anyone?

jimi-c:
This PR has been open for a year and a half with no movement, and per the above discussion I do not believe we are going to pursue it any further. 

If you or anyone else would like to continue discussing the merits of this PR, please use the devel mailing list or IRC channel:
- #ansible-devel on Freenode

Thank you!

PR #9312, opened 2014-10-11
Inventory Plugin to get hostgroups from FreeIPA

admiyo:
Only handles the default:  ignore the command line params and assumes --list was called, but uses _meta to handle host data.


mpdehaan:
Merged!

Thanks Adam!

PR #9329, opened 2014-10-13
python2.4 compatibility issue with urlparse

mcodd:
On python2.4 clients, urlparse does not actually parse urls correctly when they contain something that's not "http" (e.g. "git://" or "ssh://").  This causes parts[1] in lib/ansible/module_utils/known_hosts.py's get_fqdn() function to end up as an empty string.  Commit 578e881, which fixes urls that contain @ characters, was not indented below the check for an empty parts[1] so it ends up causing a failure/stacktrace on python2.4 clients that looks exactly like the issue described in #9045 (that issue was closed during the recent modules repo reorg).


abadger:
Looks good, thanks!

# Merged

Hi!

issue is resolved.

Thank you!

PR #9331, opened 2014-10-13
Fix detecting distribution release on OpenSuSE

marcinpraczko:
Ansible raised exception during parsering /etc/SuSE-release file.
Regular expresion should use string instead of list.
Fix tested on OpenSuse 13.1


marcinpraczko:
This is a Fix for issue: #9331 

Result after applying patch:

```
GATHERING FACTS **************
ok: [localhost]
```

mpdehaan:
Merged, thank you!

Issue #9339, opened 2014-10-14
docker images can't be pulled from repo


prune998:
it seems there are many error in the docker module. Maybe something changed in docker-py or docker API itself, but I see many error...

Call in ansible playbook is : 

```
  - name: "start container"
    docker: image="peach:5000/my-docker-image-0.1.6"
            count='1'
            state=running
```

What the module does is first, dirty, then not working.

```
        try:
            containers = do_create(count, params)
        except:
            image, tag = self.get_split_full_image_tag(params['image'])
            self.client.pull(image, tag = tag)
            self.increment_counter('pull')
            containers = do_create(count, params)
```

It first calls do_create to create the container. As this will fail, as container does not exist localy first, it will run the exception, parse the image name and do a pull : 

This leads to the error : 

```
failed: [dogwood] => {"changed": true, "failed": true}
msg: Docker API error: No such image: peach:5000/my-docker-image-0.1.6 (tag: latest)
```

Moreover, if I change the image name to add the version (latest tag) like : 

```
  - name: "start container"
    docker: image="peach:5000/my-docker-image-0.1.6:latest"
            count='1'
            state=running
```

The error is then : 

```
failed: [dogwood] => {"failed": true, "parsed": false}
invalid output was: Traceback (most recent call last):
  File "/home/belisarius/.ansible/tmp/ansible-tmp-1413310524.41-110309152753058/docker", line 2130, in <module>
    main()
  File "/home/belisarius/.ansible/tmp/ansible-tmp-1413310524.41-110309152753058/docker", line 736, in main
    containers = manager.create_containers(delta)
  File "/home/belisarius/.ansible/tmp/ansible-tmp-1413310524.41-110309152753058/docker", line 593, in create_containers
    image, tag = self.get_split_full_image_tag(params['image'])
ValueError: too many values to unpack
```

So there are 2 problems here...
As seen on the "next" module release, one solution seems to replace the 

```
image, tag = self.get_split_full_image_tag(params['image'])
```

by 

```
image, tag = self.get_split_image_tag(params['image'])
```

Once done, the image variable contains my-docker-image-0.1.6 while the tag contains 'latest'

Still, this is not correct, as the pull() function from docker-py needs two more things : 

```
    def pull(self, repository, tag=None, stream=False, insecure_registry=False):
```

you HAVE TO set : 
- repository = full image URL, like "peach:5000/my-docker-image-0.1.6"
-  insecure_registry=True 

[...]

PR #9340, opened 2014-10-14
docker module can't pull a remote image anymore

prune998:
all the details in the issue https://github.com/ansible/ansible/issues/9339

### DIFF
--- a/library/cloud/docker
+++ b/library/cloud/docker
@@ -590,8 +590,8 @@ class DockerManager:
         try:
             containers = do_create(count, params)
         except:
-            image, tag = self.get_split_full_image_tag(params['image'])
-            self.client.pull(image, tag = tag)
+            image, tag = self.get_split_image_tag(params['image'])
+            self.client.pull(https://github.com/ansible/ansible/issues/9339, tag = tag, insecure_registry=True)
             self.increment_counter('pull')
             containers = do_create(count, params)
 


jimi-c:
Hi @prune998, there are a couple of issues with this:

1) there appears to be a cut/paste error in this (the URL is hard-coded). 
2) this pull request is against release1.7.2 - all PRs must be sent to the devel branch.

Please get the patch corrected, and re-submit this PR against the development branch.

Thanks!

prune998:
Well, I tried to help, but you're not taking it.
I'd rather let everybody trying to use the docker module report a bug until you correct it. I'm not going to install the devel branch on my productions servers to see if there is a regression with the docker module. Sorry.
There is one, I pointed it out, you can either correct it (the patch is not that hard to read/understand) or wait until the new release is out and hope it correct the bug...

Regards,

jimi-c:
@prune998 we definitely appreciate the time you've taken to send us a PR, and the comments I gave above were aimed at making sure you received credit for your fix as opposed to us tweaking your patch.

However, we do clearly note in our community/contributing document (http://docs.ansible.com/community.html#contributing-code-features-or-bugfixes) that all PRs should be made against the `devel` branch. I only closed this issue because GitHub does not support changing the source of a PR, so it would have to be changed via a new request.

If you do not currently run devel and do not wish to modify existing servers it is quite easy to do a git checkout on another machine/vm/docker container so you have a clean environment on which to work, so we hope you do resubmit this based on our guidelines.

Thank you again!

prune998:
sorry beeing rude. This is a production regression, and I find it funny to have to wait until next release to correct it. 
I'll try to test the devel version when I can and see if the bad behaviour is corrected, as it seems some part of the code greatly changed.

jimi-c:
@prune998 we will most likely be doing a 1.7.3 sooner rather than later, so if you find that it is fixed in devel (or if you do send us an updated PR) we'll get this included in that (as well as several other bugs we've fixed in that module).

Issue #9343, opened 2014-10-14
Make --one-line truly one line per host


edgan:
I want to be able to pipe the output of ansible to grep and grep out hosts. To do this I really need hosts to be one per line. The current implementation of --one-line is ineffective.

I have already figured out a very easy way to do this.

mpdehaan:
I'm going to close this one as the pull request tracks the same concept, which makes it easier to sort through the GitHub.

Thanks!

PR #9344, opened 2014-10-14
Making --one-line truly one line per host by converting newlines in output to the string \n

edgan:
See issue:
https://github.com/ansible/ansible/issues/9343


amenonsen:
I think this can probably be closed now. The PR itself is outdated and in devel, `plugins/callback/oneline.py` does `replace('\n','')` on messages in various places (but not in `_command_generic_msg`, which is why I say "probably" above).

bcoca:
fixed via a56ff7ae54d93d2c930d2deb59362a7913830d54

Issue #9351, opened 2014-10-15
Improve error message for invalid key in task


hughsaunders:
##### Summary:

When trying to use a module that is not available ansible responds with "ERROR: %s is not a legal parameter in an Ansible task or handler". This is not a helpful messages as custom modules are legal, except this one hasn't been found. It would be better to note that it may be a missing module and show the current search path. 
##### Steps To Reproduce:
1. Call a missing module from a task list
2. Note confusing error message
##### Expected Results:

Helpful error message, for example:
ERROR: Ansible does not recognise debuug as an available module or legal parameter. Current module search paths: /home/hugh/git/ansible/.venv/local/lib/python2.7/site-packages/ansible/modules
##### Actual Results:

Less helpful message (see summary).

jimi-c:
# Closing This Ticket

Hi @hughsaunders, this should be resolved with a much better error in 2.0+.

If you continue seeing any problems related to this issue, or if you have any further questions, please let us know by stopping by one of the two mailing lists, as appropriate:
- https://groups.google.com/forum/#!forum/ansible-project - for user questions, tips, and tricks
- https://groups.google.com/forum/#!forum/ansible-devel - for strategy, future planning, and questions about writing code

Because this project is very active, we're unlikely to see comments made on closed tickets, but the mailing list is a great way to ask questions, or post if you don't think this particular issue is resolved.

Thank you!

PR #9352, opened 2014-10-15
Improve error message for missing module

hughsaunders:
See Issue #9351


amenonsen:
This can be closed now, the underlying code has changed completely and the fix would now have to be applied to `lib/ansible/parsing/mod_args.py` (and look very different).

For the record, devel says `ERROR! no action detected in task`, which is also not as useful as it could be.

PR #9358, opened 2014-10-16
Print tasks grouped by role when doing --list-tasks

pescobar:
When one playbook include different roles and option --list-tasks is used the different tasks are printed grouped by the role they belong to.

this is the output when doing "ansible-playbook --list-tasks"  

```
$> ansible-playbook web-balancers.yml --list-tasks

playbook: web-balancers.yml

  play #1 (Configure web balancers):
    role: common
        install resolv.conf file {{ rolename }} 
        install every yum repos keys 
        disable selinux 
        Install ntp 
        Copy ntp conf file 
        Start the ntp service 
        Install postfix 
        Copy postfix conf file 
        Start the postfix service 
        Install vim 
        Copy .vimrc for root 
        Copy .vim dir for root 
        Install check_mk monitoring agent 
        Copy config file for check_mk monitoring agent 
        Install munin-node package 
        Copy munin-node config file 
        disable unwanted services 
    role: apache
        Open required IPTables ports 
        Check if apache SSL cert already exists 
        Create self-signed SSL cert for Apache (valid for 10 years) 
        Enable the new SSL cert in Apache conf (line 1) 
        Enable the new SSL cert in Apache conf (line 2) 
```

This is my first look at ansible's code so maybe I am missing something or there is a better way to achieve this but it seems to be working fine for me

### DIFF
--- a/bin/ansible-playbook
+++ b/bin/ansible-playbook
@@ -251,6 +251,8 @@ def main(args):
 
                     if play.roles:
                         for rol in play.roles:
+			    if isinstance(rol, dict):
+			        rol = rol.get('role')
                             print '    role: %s' % (rol)
                             for task in taskslist:
                                 if task.role_name is rol:
--- a/bin/ansible-playbook
+++ b/bin/ansible-playbook
@@ -241,25 +241,24 @@ def main(args):
                 if options.listtasks:
                     print '  play #%d (%s):' % (playnum, label)
 
+                    taskslist = []
+                    for task in play.tasks():
+                        if (set(task.tags).intersection(pb.only_tags) and not
+                            set(task.tags).intersection(pb.skip_tags)):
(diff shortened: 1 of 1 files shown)


pescobar:
I have added a new commit to correctly handle when the roles are included as a dict. 

Ex:

```
  roles:
    - common
    - { role: infiniband, when: "infiniband == 'TRUE'", tags: ['infiniband'] }
```

amenonsen:
This particular PR should probably be closed because the code has been completely overhauled, but note the open issue #11737 about `--list-tasks` not working in devel.

bcoca:
list-tasks now works in devel

amenonsen:
I'm closing this now because it no longer applies to the current code, and --list-tasks has been fixed per the above comment. @pescobar Thanks for your contribution.

PR #9360, opened 2014-10-16
Read AWS key id and secret from ec2.ini

jessedhillon:
If you have multiple projects which use AWS, it's not convenient to use global- or user-wide boto configs. You could use environment variables, but since we have ec2.ini right there, it seems like an easy place to have your project's AWS credentials.


jessedhillon:
Really sorry, but I have to reissue this PR. I didn't do my development in a branch, but rather in `devel` and now these commits cannot be isolated from the commits related to another feature I want to merge upstream as well. Sorry again.

PR #9365, opened 2014-10-17
fix ssh_config inventory plugin to work with paramiko 1.7.2 till 1.15.2

t0mk:
The parsed dict from ~/.ssh/config is different across various versions on Paramiko. I made it to work with versoins from 1.7.2 to 1.15.1 (current newest in pip). Ubuntu 12.04 has now 1.7.7 and CentOS 6.5. has 1.7.5, so I think that's compatible enough.

### DIFF
--- a/contrib/inventory/ssh_config.py
+++ b/contrib/inventory/ssh_config.py
@@ -19,6 +19,10 @@
 
 # Dynamic inventory script which lets you use aliases from ~/.ssh/config.
 #
+# There were some issues with various Paramiko versions. I took a deeper look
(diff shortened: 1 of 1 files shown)


amenonsen:
@t0mk the list of Paramiko versions in the comment has a lot of duplications. Do you feel like fixing that and rebasing this to apply onto devel? This seems like a reasonable change, since you already did the hard work of testing each version.

t0mk:
I'll take a look. thank for pointing out the duplications.

t0mk:
checked and rebased

PR #9366, opened 2014-10-17
Split comma separated tags and include in groups

anperez:
Split tags that are comma separated and add the host to those different groups.

E.g.: myhost: env = "dev,staging"

Instead of having the actual:
"tag_env_dev_staging": [ ..., "myhost" ]

this would result in:
"tag_env_dev": [ ..., "myhost" ],
"tag_env_staging": [ ..., "myhost" ]

I think it's more flexible and allows more grouping options based in tags.


amenonsen:
This is a duplicate of #8676.

bcoca:
closing in favor of #8676 which also makes this configurable

PR #9368, opened 2014-10-17
adding for loop on list to to handle the return of none when the list is empty

silvax:
adding for loop on list to to handle the return of none when the list is empty. With the previous method if the list was empty the script died. See http://stackoverflow.com/questions/18852324/typeerror-sequence-item-0-expected-string-nonetype-found.

This was causing the inventory script to fail on our account until we made the change. Works good now

### DIFF
--- a/plugins/inventory/ec2.py
+++ b/plugins/inventory/ec2.py
@@ -622,8 +622,8 @@ class Ec2Inventory(object):
                 for group in value:
                     group_ids.append(group.id)
                     group_names.append(group.name)
-                instance_vars["ec2_security_group_ids"] = ','.join(group_ids)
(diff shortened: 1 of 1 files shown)


mpdehaan:
Thanks for this - this seems reasonable, though _might_ bite it on unicode group names.  I'm not sure that's a thing.

We might want to make a simple function that only stringifies ints, maybe?  Or maybe default group_names to an empty list and add some error code instead.

Not sure...

jimi-c:
The names should be strings already (and the ids too), so the use of str() there probably isn't required.

silvax:
Awesome. Thanks

PR #9372, opened 2014-10-17
Improved ConfigureRemotingForAnsible.ps1

trondhindenes:
```
-now checks for an existing fw rule before creating a new
-fw rule applies to all connection profiles, not just public ones (important in domain-joined scenarios)
```


cchurch:
@trondhindenes I've included these changes and more in https://github.com/ansible/ansible/pull/9537

cchurch:
I've merged https://github.com/ansible/ansible/commit/a798ecb0d48bf9ddb36183a7b2b8f1fb627a3803 which includes these fixes/changes.

PR #9373, opened 2014-10-17
Add 'ipaddr' filter

drybjed:
'ipaddr' is an Ansible filter plugin which recognizes IPv4 and IPv6
addresses and can manipulate them to extract useful information or
reformat the output.


drybjed:
I'm working on some fixes and new features in the filter, I will create new pull request later with improved plugin.

PR #9375, opened 2014-10-18
Fix os_family and distribution on archlinux

mantiz:
Since `1.7.0` the facts `os_family` and `distribution` are set to `NA` on Archlinux.
I think this was introduced by https://github.com/ansible/ansible/commit/f4fd9d36.
The file `/etc/arch-release` is empty on arch which is the reason why it is ignored.

This was originally pullrequested in #9012 which was closed because repositories were reorganized.

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -273,84 +273,85 @@ class Facts(object):
             self.facts['distribution_release'] = dist[2] or 'NA'
             # Try to handle the exceptions now ...
             for (path, name) in Facts.OSDIST_LIST:
-                if os.path.exists(path) and os.path.getsize(path) > 0:
-                    if self.facts['distribution'] in ('Fedora', ):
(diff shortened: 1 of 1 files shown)


Jmainguy:
I can confirm this fix works for ArchLinux

mantiz:
Are there any plans to merge this at some time? :)
Arch ships v1.8.1 now and this is still not fixed even though a fix exists for 2,5 months now.
No offense, I just wonder why it is not merged. Is there anything wrong with the PR?

bcoca:
sorry for the delay, but there are many contributions to ansible and as much as we want to include them all we just physically cannot, we will get to this as soon as we can.

mantiz:
Thanks for the quick response, I totally understand that you guys have a lot of work and I really enjoy the results. Just wanted to know if it still on the radar or if something is wrong. So, take your time and I will just wait for it. :)

mantiz:
Great, thx for merging this. :+1:

PR #9376, opened 2014-10-18
Updated documentation to consistently use yes/no instead of True/False

kustodian:
Updated documentation to consistently use `yes`/`no`, instead of `True`/`False` and `true`/`false`.


bcoca:
actually they are all valid as booleans, closing this PR as this should work as is.

PR #9377, opened 2014-10-18
Implement a playbook gather_facts_force flag

craigtracey:
Right now when executing a play that makes use of the subsets (via
--limit) or tags, facts are only gathered from the set of hosts that
will comprise the play. This is problematic when those playbooks or
their associated roles attempt to utilize facts gathered from hosts
that are not in the set.

Fact caching has recently been implemented. This would be an alternative
to gather_facts_force, but this requires key-value store solutions (ie.
redis) that might not fit into some environments. Similarly, there may
be some scenarios where the potentially stale facts are problematic.

This change allows subsets to be conditionally applied. The default case
is to apply subsets, but this may be overriden with the apply_subset
argumement to both inventory.list_hosts and inventory.get_hosts.

This is discussed here: https://groups.google.com/forum/#!topic/ansible-project/f90Y4T4SJfQ

### DIFF
--- a/lib/ansible/inventory/__init__.py
+++ b/lib/ansible/inventory/__init__.py
@@ -36,7 +36,7 @@ class Inventory(object):
     Host inventory for ansible.
     """
 
-    __slots__ = [ 'host_list', 'groups', '_restriction', '_also_restriction', '_subset', 
+    __slots__ = [ 'host_list', 'groups', '_restriction', '_also_restriction', '_subset',
                   'parser', '_vars_per_host', '_vars_per_group', '_hosts_cache', '_groups_list',
                   '_pattern_cache', '_vault_password', '_vars_plugins', '_playbook_basedir']
 
@@ -53,7 +53,7 @@ class Inventory(object):
         self._vars_per_host  = {}
         self._vars_per_group = {}
         self._hosts_cache    = {}
-        self._groups_list    = {} 
+        self._groups_list    = {}
         self._pattern_cache  = {}
 
         # to be set by calling set_playbook_basedir by playbook code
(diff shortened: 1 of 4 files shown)


thijsc:
This would very useful for us :+1:

jimi-c:
Hi @craigtracey. I believe a cleaner way to accomplish this would be to add a new 4th option for `gather_facts` (something like `all` or `force`) which would accomplish the same thing, as opposed to adding a new language option for this. 

I would think this should be a bit easier to do too, as we already have the check for `smart` mode there, so the new option would simply allow the list of hosts touched to be modified, similarly to the way smart mode works.

mpdehaan:
Fact caching is going to be our solution to this one.

jimi-c:
One final note here, the way we normally recommend working around this issue is to have an empty play at the top of the playbook (or one you include at the top of your playbook) which looks like this:

<pre>
- hosts: all
</pre>

This will use the fact_caching now, as well as the smart fact-gathering mode to efficiently gather facts about hosts that have not been touched (or for which the cache may have expired).

craigtracey:
@jimi-c correct me if I am wrong, but this method does not work with the --limit flag.

craigtracey:
@jimi-c I was between 'gather_facts_force: True' and 'gather_facts: force'. I was waiting for comments to see which was preferred, but I guess this is a moot point as this is not going to be considered.

bcoca:
I was thinking of an alternate method that would always work even with
limit, using delegate to.
- setup
  delegate_to: item
  with_items: groups['groupineedfactson']

the current issue is that delegate_to will insert the facts in the current
host, I think we should change it to insert the facts into the delegated
host or add a flag that allows for this and leaves the current behavior by
default.

On Mon, Oct 20, 2014 at 8:59 AM, craigtracey notifications@github.com
wrote:

> @jimi-c https://github.com/jimi-c I was between gather_facts_force and
> gather_facts: force. I was waiting for comments to see which was preferred,
> but I guess this is a moot point as this is not going to be considered.
> 
> —
> https://github.com/ansible/ansible/pull/9377#issuecomment-59785141.

## 

Brian Coca
Pedo mellon a minno

zelenkovsky:
This one seems to be exactly what I want: https://github.com/ansible/ansible/pull/9387

PR #9378, opened 2014-10-19
rework of the nova inventory

t0mk:
- you can refer to instances by their nova name, DNS name, or IP address
- you can refer to single tenants, regions and openstack environments in ansible{-playbook} calls
- you can refer to a hostgroup when you pass the arbitrary --meta group= in "nova boot" [1]
- it caches the state of the cloud => consequent calls are fast
- it tries to guess ansible_ssh_user based on name of image ('\cubuntu' -> 'ubuntu', '\ccentos' -> 'cloud-user', ...)
- allows to access machines by their private ip [1]
- it will work with no additional configuration, just handling single tenant from set OS_\* environment variables (just like python-novaclient).
- you can choose to heavy-configure it for multiple environments
- configured from simple YAML (I dislike ConfigParser). See nova.yml in the diff

[1] I took few ideas and some code from other pull requests
- https://github.com/ansible/ansible/pull/8657 by Monty Taylor
- https://github.com/ansible/ansible/pull/7444 by Carson Gee


t0mk:
thread:

https://groups.google.com/forum/#!topic/ansible-devel/MI2sUZTF8GE

emonty:
Hey, we should for sure figure out how to combine forces - I like a bunch of the things you've done here. My most recent stuff is now up in #9421 and uses the os-client-config library. One of the main changes there is moving most of the logic into the shade library so that the same stuff can be used in a facts module and in other places. I've got a long flight tomorrow, so I'll try to take a stab at a consolidation and see what you think?

t0mk:
Hi Monty, I am happy that someone else is also looking on OpenStack in Ansible. I checked your PRs
https://github.com/ansible/ansible/pull/9421
and
https://github.com/ansible/ansible-modules-extras/pull/92
It's a nice code, but I have a few comments:
- too much search paths - that's confusing to read. Keep it simple and just expect the config in a place and if a value is not there gracefully fall to OS_\* env var.
- dependency on two random (although yours) pip packages: os-client-config and shade. Can you please justify? I am not convinced that the logic of converting a few env vars and/or a yaml file, deserves a pip package (os-client-config). shade is a good idea for sure, I will use it for my wrappers probably (so thanks for putting this one up!), but I am not sure if it's necessary between Ansible and OpenStack.
- the current OpenStack modules work quite OK, I actually had a demo last week about those. What does your PR to extras bring on top of them?

At this point, I am not interested in joining efforts, because I don't agree with the dependencies and I find stuff too complicated in your code. However, I would like to hear an opinion of an Ansible dev. I am ready to yield if the maintaners are on your side :).

abadger:
Hey, I just saw this -- I haven't looked at comparisons of the code or any of that.  Just wanted to respond to the "ansible dev" portion.  We're trying (slowly but hopefully we're making progress in this area) to push some of the decisions about specific areas out to people who have knowledge in that area.  This is something like the lieutenant model that is used with the Linux kernel.  @emonty is one of the lieutenants for openstack.  So it would be great if you could work with him to address any simplifications, bugfixes, or feature additions to openstack portions of the code.

emonty:
Oh  - I totally thought I'd responded to this a while back ... sorry. We want to move forward with the openstack code being based on the shade library which is also seeing use in OpenStack's nodepool and which is now going in to upstream gate testing - which should make it pretty rock solid. I'd love to get any features you've got here that are not in the newly landed openstack.py script. I'll put it on the list of things to circle back on.

emonty:
@gregdek we already merged the new inventory - we should go ahead and close this one.

PR #9379, opened 2014-10-19
Preserve backup files

wfranzini:
This pull-request introduce two distinct changes:
- an existing backup file is never overwritten;
- the backup extension is the same for all backups in a given run.

Without this change if a file is modified multiple times in the same
execution only the last edit can be rolled back.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -244,6 +244,7 @@ def json_dict_unicode_to_bytes(d):
 
 
 class AnsibleModule(object):
+    backup_ext = None
 
     def __init__(self, argument_spec, bypass_checks=False, no_log=False,
         check_invalid_arguments=True, mutually_exclusive=None, required_together=None,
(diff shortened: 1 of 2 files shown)


mpdehaan:
Note we are planning to overhaul the way 'backup' works - so we may instead decide to replace this with the new way.

wfranzini:
On Mon, Nov 17 2014, Michael DeHaan wrote:

> Note we are planning to overhaul the way 'backup' works - so we may
> instead decide to replace this with the new way.

I'd like to have "correct" backup files, the way it's implemented is
less important to me :-)

## ciao 

walter franzini

amenonsen:
Looking at `backup_local` in `module_utils/basic.py` now, I don't see any signs that the "new way" has been implemented yet. Is this still on the cards? If not, this change seems reasonable. @wfranzini, do you feel like rebasing your PR onto devel? I also think this could use some explanation in the docs.

wfranzini:
On Sat, Jul 25 2015, Abhijit Menon-Sen wrote:

> Looking at `backup_local` in `module_utils/basic.py` now, I don't
> see any signs that the "new way" has been implemented yet. Is this
> still on the cards? If not, this change seems reasonable. @wfranzini,
> do you feel like rebasing your PR onto devel? 

No problem, I'll rebase it in the next days.

## 

walter franzini

amenonsen:
I'll close this PR since it hasn't been updated for months, but please feel free to open a new one if you have a chance to rebase and test this change. @wfranzini Thanks for your contribution to Ansible, it is appreciated.

PR #9381, opened 2014-10-19
Detect kvm guests through sys_vendor.

debfx:
The two detection methods currently implemented do not cover all cases.

For example qemu guests invoked like this:
qemu -machine pc-i440fx-1.4,accel=kvm -cpu SandyBridge

return this information:
product_name: Standard PC (i440FX + PIIX, 1996)
cpuinfo: model name      : Intel Xeon E312xx (Sandy Bridge)


jimi-c:
Merged, thanks!

Issue #9349, opened 2014-10-15
Docker containers aren't detected as such


gquintard:
On more recent docker versions, libcontainer is used instead of lxc, and so ansible_virtualization_type is not filled correctly.

Here's a patch proposition, checking /proc/1/cgroup for '/docker/' as well as '/lxc/'. It may be a bit brutal though.

http://pastebin.com/mWXGWGEp

mpdehaan:
Can you please submit this as a pull request instead of a gist?

We'd be glad to have it.

Thanks!

PR #9382, opened 2014-10-20
Detect docker as virtualization_type

gquintard:
New docker versions use "docker" in /proc/1/cgroup, which disturbs the
virtualization_type detection. So, grep for docker, in addition of
"lxc".

This answers #9349


jimi-c:
Merged, thank you!

PR #6846, opened 2014-04-03
Don't disable GSSAPIAuthentication or PubkeyAuthentication when using --ask-pass

kolbyjack:
As requested on irc, here's the output of ansible-playbook -vvvv with one echo on a delegate_to host and one on the main host.  The password on the delegated host is different from the one I provided on the command line, so it had to use the pubkey, and the main host doesn't have a pubkey on it yet, so it had to use the password.

TASK: [createvm | echo on delegated host] ****************************************
<delegate.example.net> ESTABLISH CONNECTION FOR USER: ansible
<delegate.example.net> REMOTE_MODULE command echo "This is on the delegated host"
<delegate.example.net> EXEC ['sshpass', '-d14', 'ssh', '-C', '-vvv', '-o', 'ControlMaster=auto', '-o', 'ControlPersist=60s', '-o', 'ForwardAgent=yes', '-o', 'ControlPath=/home/jon/.ansible/cp/ansible-ssh-%h-%p-%r', '-o', 'User=ansible', '-o', 'ConnectTimeout=10', u'delegate.example.net', '/bin/sh -c /usr/bin/python']
changed: [main.example.net] => {"changed": true, "cmd": ["echo", "This is on the delegated host"], "delta": "0:00:00.001584", "end": "2014-04-03 13:19:35.184026", "item": "", "rc": 0, "start": "2014-04-03 13:19:35.182442", "stderr": "", "stdout": "This is on the delegated host"}

TASK: [bootstrap | echo on main host] *****************************************
<main.example.net> ESTABLISH CONNECTION FOR USER: root
<main.example.net> REMOTE_MODULE command echo "This is on the main host"
<main.example.net> EXEC ['sshpass', '-d9', 'ssh', '-C', '-vvv', '-o', 'ControlMaster=auto', '-o', 'ControlPersist=60s', '-o', 'ForwardAgent=yes', '-o', 'ControlPath=/home/jon/.ansible/cp/ansible-ssh-%h-%p-%r', '-o', 'User=root', '-o', 'ConnectTimeout=10', u'main.example.net', '/bin/sh -c /usr/bin/python']
changed: [main.example.net] => {"changed": true, "cmd": ["echo", "This is on the main host"], "delta": "0:00:00.001749", "end": "2014-04-03 13:19:42.459113", "item": "", "rc": 0, "start": "2014-04-03 13:19:42.457364", "stderr": "", "stdout": "This is on the main host"}

PR #9386, opened 2014-10-20
Add support for multiple required auth for openssh v6.2 onwards.

wfranzini:


jimi-c:
Hi @wfranzini, shouldn't this include a check of the SSH version, so that this doesn't break for older versions still?

wfranzini:
Hi @jimi-c I don't think a check against ssh version is required, since my commit just _drop_ GSSApiAuthentication=no and PubKeyAuthentication=no when a password is used.  

This feature is enable on the server side not on the client side (ie. I'm using ossh 6.0 client against ossh 6.2 server).

Let me know if more info are needed.

amenonsen:
This PR the same as #6846, where I've posted an updated patch with the merge conflicts resolved. This PR can be closed. There are no backwards-compatibility concerns.

bcoca:
closing this in favor of dupe #6846

PR #9387, opened 2014-10-20
Add 'gather_facts: force' capability

craigtracey:
This change adds the 'force' configuration value to the gather_flags
play attribute. With 'gather_facts: force' set, all hosts will have
facts gathered in accordance with the default fact gathering policy
(ie. smart, implicit, etc.).

This is useful in cases where --limit or --tags constrains the subset
of hosts that will be executed against, thereby limiting the number of
hosts that will be fact-gathered in the traditional sense.

### DIFF
--- a/docsite/rst/intro_configuration.rst
+++ b/docsite/rst/intro_configuration.rst
@@ -266,9 +266,9 @@ gathering
 
 New in 1.6, the 'gathering' setting controls the default policy of facts gathering (variables discovered about remote systems).
 
-The value 'implicit' is the default, meaning facts will be gathered per play unless 'gather_facts: False' is set in the play.  The value 'explicit' is the inverse, facts will not be gathered unless directly requested in the play.
+The value 'implicit' is the default, meaning facts will be gathered per play unless 'gather_facts: False' is set in the play.  The value 'explicit' is the inverse, facts will not be gathered unless directly requested in the play. The value 'smart' means each new host that has no facts discovered will be scanned, but if the same host is addressed in multiple plays it will not be contacted again in the playbook run.  This option can be useful for those wishing to save fact gathering time.
 
-The value 'smart' means each new host that has no facts discovered will be scanned, but if the same host is addressed in multiple plays it will not be contacted again in the playbook run.  This option can be useful for those wishing to save fact gathering time.
+New in 1.8, 'gather_facts' supports True, False, and 'force'. Fact gathering is governed by both the default gathering configuration (mentioned immediately above) as well as any play-specific 'gather_facts' declarations. In the case of 'gather_facts: True' (the default), facts will be gathered according to the default gathering policy. Inversely, 'gather_facts: False' will disable fact gathering for the play.  'gather_facts: force' is useful in cases where --limit or --tags are applied to a play. In this case, facts will be gathered for all hosts (again, in accordance with the default gathering policy), not just the subset of hosts limited by those two command line arguments.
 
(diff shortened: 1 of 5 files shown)


zelenkovsky:
Great feature, I need it a lot! How to vote for it?

zelenkovsky:
This patch does not pickup variables from hosts_vars folder. What is the proper way to access them?
To be more precise: How can I access variables located in hosts_vars folder for another host, when I use --limit or --tag?

This code does not work:
{% for host in groups['database'] %}
    Description    {{ hostvars[host]['my_deskcription']  }}
{% endfor %}

omgjlk:
That's not likely to work. This is entirely about host gathered facts, essentially forcing a run of the 'setup' module on hosts outside of the limit. Delegate_to wouldn't work as it would associate all the gathered facts with the host in the loop, rather than with the delegated host. Adding the desired hosts back into a run-time inventory with add_host is too onerous for this kind of action as well.

I believe this should be merged as is, as it does just what it advertises, forces the gathering of facts for all the matched hosts ignoring the limit specification.

igrek00:
Is this going to merge? This feature will add a lot of flexibility to playbook play process.

wenottingham:
This would need a port to v2 - if there are a lot of people that can make use of this, let us know and we can prioritize it.

abadger:
@bcoca mentioned on chat that he has an alternate idea (delegated fact gathering) that would enable the same functionality.  @bcoca, @craigtracey maybe you guys should hash that out on the mailing list to see if that will satisfy the requirements and what it would look like.

craigtracey:
@bcoca if you could describe on ML and/or point me at a PR, I'd love to take a look. Thanks!

bcoca:
basically allow fact gathering to affect delegated hosts:

setup: update={{item}}
with items: "{{groups['dbservers']}}
delegate to: item

currently facts that are delegated are incorporated to the inventory
hostname

zelenkovsky:
I was able to do workaround like this:
    deploy.yml
    ---
    # Deploy web client & blog servers
    - hosts: all:&{{ myinventory }}
    sudo: yes
    roles:
      - web
      - blog

ansible-playbook deploy.yml --extra-vars="myinventory=$SERVER"

But it does not mean that it's a good practice.

bcoca:
I'm going to close this in favor of adding the ability to delegate fact gathering which would allow a much more fine grained control and provide the same functionality. The force feature is a bit too all or nothing.

omgjlk:
Doing it as a task means adding a bunch of tasks all over our playbooks. That's not ideal.

bcoca:
forcing it to gather all facts every time can become too onerous, specially when dealing with large inventories or hosts on which fact gathering is slow. If you frequently use cross host facts I would suggest fact caching as a better alternative.

omgjlk:
So our setup is such that we can run roles independently, with tags. However some roles DO require facts from all hosts, even if the role nominally only operates on a specific set of hosts. That's why we have gather_facts set to force on those particular plays that call those roles. Yes, we could just add the task to force the facts, and probably what we'd do just to get out of having forked ansible, but given that the end result is basically the same, gathering facts for those hosts, what's the net gain on manually calling the setup task?

bcoca:
to your particular case, not much, but as a general feature it is a big difference

bcoca:
at this point if you want to bring this up in ansible-devel and get more of the community involved, there might be other arguments to be made neither of us are seeing.

PR #9396, opened 2014-10-21
Update dict.py

hughdbrown:
Write with a list comprehension.


amenonsen:
As far as I can tell, there's no compelling reason to apply this patch. Given that it's accumulated merge conflicts, I think it should just be closed now.

bcoca:
closing as this is a style change and we don't normally accept them as they break attribution

PR #9401, opened 2014-10-22
Use entry_points console_scripts

tomoh1r:
It's enable you to test tools just call the function.

see also http://www.scotttorborg.com/python-packaging/command-line-scripts.html

Best regards.


mpdehaan:
Hi,

Thanks for the idea and submission.

We're going to consider this for the v2/ tree in the future, but not now and not for the main tree.

Should be a release or two down the line.

Thanks!

PR #9408, opened 2014-10-23
Update playbooks_variables.rst

juriglx:
There is some kind of duplicated content with http://docs.ansible.com/faq.html#how-do-i-loop-over-a-list-of-hosts-in-a-group-inside-of-a-template and this gotcha isn't listed here.


jimi-c:
Hi @juriglx, could you modify this slightly to mention fact caching in 1.8? For example, `You need to make sure that the facts of those hosts have been populated before though, for example by running a play against them if the facts have not been cached recently (fact caching was added in Ansible 1.8).`

Thanks!

juriglx:
Hi @jimi-c, it's done.

jimi-c:
Merged, thank you!

PR #9418, opened 2014-10-24
Ignore missing tags when using --skip-tags

sudarkoff:
The rationale behind this change is that I don't really care if a tag is
missing when I want to skip it.

One potential reason to prefer to throw an error is to prevent a user
from executing a tag that they did not intend to execute as theoretically
that could cause issues. But in practice, one should not rely on
--skip-tags for avoiding disastrous results. There are better ways to
not break your hosts.

Alternatively, I can add an option, say --ignore-missing-tags, that will
instruct --skip-tags not to throw an error.

### DIFF
--- a/lib/ansible/playbook/__init__.py
+++ b/lib/ansible/playbook/__init__.py
@@ -330,7 +330,7 @@ class PlayBook(object):
         # if the playbook is invoked with --tags or --skip-tags that don't
         # exist at all in the playbooks then we need to raise an error so that
         # the user can correct the arguments.
-        unknown_tags = ((set(self.only_tags) | set(self.skip_tags)) -
+        unknown_tags = (set(self.only_tags) -
                         (matched_tags_all | unmatched_tags_all))
         unknown_tags.discard('all')
 


hyfather:
+1 for this behavior.
Dupe here: https://github.com/ansible/ansible/issues/10510

bcoca:
the reason it errors out now is because you could have misspelled a tag and would then run parts of the play you did not intend

sudarkoff:
@bcoca Yes, as I noted in the description of this PR, that's one possible reason for throwing an error. On one hand, if running a tag can cause a serious issue, --skip-tags is a poor protection. (Because if it's easy to misspell the tag name, it is just as easy to forget to specify that option to begin with.) On the other hand, I can add a `--ignore-missing-tags` option to switch the behavior.

amenonsen:
I don't feel strongly about ignoring `--skip-tags` tags if they don't exist (though I have a weak preference for the status quo, because whenever I've used that option with a tag that didn't exist, it was a misunderstanding on my part). If someone really wants this behaviour, though, I think a separate `--ignore-missing-tags` option would be a better way to do it.

boukeversteegh:
@bcoca

i think avoiding the dangers of misspellings should not be a goal for a project like ansible which is built for users that take responsibility for the deployment and maintenance of production servers and such, and who need to be careful anyway to avoid making hundreds of different mistakes.

My suggestion is to remove this error so that users who should be careful anyway can create playbooks without this kind of limitation.

sudarkoff:
I do not subscribe to this point of view, @boukeversteegh. "Being careful" is not an adequate protection from screw ups. I no longer feel this is a good PR, so going to withdraw it.

Issue #9414, opened 2014-10-24
Add 'eu-central-1' to AWS_REGIONS in ec2 module


adonig:
##### Issue Type:

“Bug Report”
##### Ansible Version:

ansible 1.7.2
##### Environment:

“N/A”
##### Summary:

Ansible misses support for the new AWS Frankfurt region. You can fix it by adding 'eu-central-1' to AWS_REGIONS in lib/ansible/module_utils/ec2.py
##### Steps To Reproduce:

Try to create a stack in region 'eu-central-1' with the cloudformation module.
##### Expected Results:

It should create the stack.
##### Actual Results:

It fails with the following error message:

```
msg: value of region must be one of: ap-northeast-1,ap-southeast-1,ap-southeast-2,eu-west-1,sa-east-1,us-east-1,us-west-1,us-west-2, got: eu-central-1
```

cchurch:
This will only work if you've also upgraded to boto 2.34.0.  Just my $0.02, but it may make more sense to use `boto.regioninfo.get_regions("service_name")` instead of maintaining a separate list of regions?

mpdehaan:
Module tickets should be filed on the modules repo.  Can you please open this here:

https://github.com/ansible/ansible-modules-core

willthames:
@mpdehaan - you might think this is a modules issue but the code for this is almost entirely in ansible core, as the report suggests (the fix is in `lib/ansible/module_utils/ec2.py`)
@cchurch - wow, it's incredible that boto needs updating too (I guess not too incredible otherwise ansible wouldn't need a fix either). I wonder how best to expose that to the user who wishes to use eu-central-1 but boto fails. It's not very handy that `connect_to_region` silently fails and returns `None` rather than throwing an exception

PR #9419, opened 2014-10-25
Add Frankfurt AWS region

willthames:
AWS recently announced the Frankfurt region.

http://aws.amazon.com/blogs/aws/aws-region-germany

This fixes all AWS modules that rely on the module_utils/ec2.py (the vast majority). The exceptions are `ec2_facts` and `ec2_ami_search`. A fix against ansible/ansible-modules-core will be required to address those two.

### DIFF
--- a/lib/ansible/module_utils/ec2.py
+++ b/lib/ansible/module_utils/ec2.py
@@ -37,6 +37,7 @@ AWS_REGIONS = [
     'ap-southeast-1',
     'ap-southeast-2',
     'eu-west-1',
+    'eu-central-1',
     'sa-east-1',
     'us-east-1',
     'us-west-1',
--- a/lib/ansible/module_utils/ec2.py
+++ b/lib/ansible/module_utils/ec2.py
@@ -164,6 +164,11 @@ def boto_fix_security_token_in_profile(conn, profile_name):
 
 def connect_to_aws(aws_module, region, **params):
     conn = aws_module.connect_to_region(region, **params)
+    if not conn:
+        if region not in [aws_module_region.name for aws_module_region in aws_module.regions()]:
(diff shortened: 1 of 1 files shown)


jimi-c:
Hi @willthames, I believe there are a few modules that still duplicate the hard-coded list. Could you be sure to update those as well (or even better, modify them as well to use the shared list)?

Thanks!

willthames:
@jimi-c I explicitly mentioned those in my pull request, but they would need a separate PR as they're in ansible/ansible-modules-core

Are you saying you'll only accept this PR once there exists a separate PR in a separate project?

willthames:
@cchurch made a useful comment in the now closed #9414 that we could programatically determine the region choices. 

That would involve updating `ec2_argument_spec` so that modules could pass it the servicename to check the regions against and then updating all the modules that use it. But the results would be more accurate (if a service wasn't available in a region it would not be in the list of choices). Making the failure message more useful (this service does not appear to be available in your chosen region, or your region doesn't exist, or you need to upgrade boto!) could be tricky too (as it would fail at module call time rather than during the run of the module itself). I guess one possibility would be to make the choices as wide as possible (as per this PR) but also further check the region being passed at the time of running `connect_to_aws` (just needs a map of boto module, which `connect_to_aws` uses, to servicename)

jimi-c:
@willthames sorry for missing that, this is fine then.

willthames:
Added code to handle case where boto doesn't yet know about the region. Note that modules that use `connect_to_aws` rather than `ec2_connect` should be updated to handle the `StandardError` that `connect_to_aws` throws. 

@jimi-c can you remove the needs_revision label please.

willthames:
ansible/ansible-modules-core#260 has the module related changes for this pull request.

PR #9421, opened 2014-10-26
Add support for new OpenStack modules

emonty:
New module_utils functions, and a new inventory plugin needed by ansible/ansible-modules-extras#92


jimi-c:
@emonty scratch that, as I realized you guys only have access to commit on the module repos.

PR #9423, opened 2014-10-26
Add required_if to AnsibleModule

emonty:
There is a common pattern in modules where some parameters are required
only if another parameter is present AND set to a particular value. For
instance, if a cloud server state is "present" it's important to
indicate the image to be used, but if it's "absent", the image that was
used to launch it is not necessary. Provide a check that takes as an
input a list of 3-element tuples containing parameter to depend on, the
value it should be set to, and a list of parameters which are required
if the required parameter is set to the required value.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -247,7 +247,8 @@ class AnsibleModule(object):
 
     def __init__(self, argument_spec, bypass_checks=False, no_log=False,
         check_invalid_arguments=True, mutually_exclusive=None, required_together=None,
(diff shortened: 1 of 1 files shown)


omgjlk:
I've wanted something like this before as well, :+1:

abadger:
Talked about this with @j2sol on the live meeting today.  Looks good.  I'll merge and make sure it applies to the v2_final branch as well.

bcoca:
+1

abadger:
Applied cleanly to v2_final as well.  Thanks!  Now we just need to have someone clean up all the modules where this logic is written out manually ;-)

# Merged

Hi!

issue is resolved.

Thank you!

PR #9427, opened 2014-10-27
Add a connection backend to manage remote LXC containers

Gu1:
This pull request introduces a new connection backend to manage remote LXC containers by ssh'ing into the host machine and using lxc-attach to get a shell on the container.

The rationale for it is to allow to manage LXC containers without having to run a publicly accessible ssh daemon on every single one of them, with all the hassle it causes (having to assign a public ipv4 on each container or having to do port forwarding for each containers's ssh daemon on the host, or having ipv6 assigned to each container but requiring ipv6 to access them...).

While the code in this pull request should work, it definitely still needs more testing and some polishing (the CommunicateCallbacks class still feels a bit clunky).

Please review.

### DIFF
--- a/lib/ansible/runner/connection_plugins/lxc_remote.py
+++ b/lib/ansible/runner/connection_plugins/lxc_remote.py
@@ -1,0 +1,171 @@
+# (c) 2014, Gu1 <gu1@aeroxteam.fr>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible. If not, see <http://www.gnu.org/licenses/>.
+
+import functools
+import pipes
+import os.path
+
+from ansible.runner.connection_plugins import ssh
+from ansible import errors
+from ansible import utils
+from ansible.callbacks import vvv
+
+
+def lxc_check(func):
+    @functools.wraps(func)
+    def inner(self, *args, **kwargs):
+        if self.lxc_name is None:
+            raise errors.AnsibleError("to use the lxc_remote connection mode, you must specify a container name")
+        return func(self, *args, **kwargs)
+    return inner
+
+
+class Connection(ssh.Connection):
+    """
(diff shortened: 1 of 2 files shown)


Gu1:
Pull request updated, I fixed a few bugs (mostly typos).
I can confirm this patch works fine with privileged and unprivileged containers

Gu1:
Since this probably [won't get merged](https://groups.google.com/d/topic/ansible-devel/4S7asT5rVOg) soon, I made a standalone version, in case anyone is interested:
https://github.com/Gu1/ansible-lxc-remote

amenonsen:
This is pretty interesting. First, a quick summary from the mailing list thread of why this wasn't merged: (a) large changes to `ssh.py`, (b) don't want to maintain both docker and lxc support separately, (c) no strategy for connection chaining.

@Gu1 note that the current devel branch is the v2 code, and there have been significant changes to `lib/ansible/plugins/connections/ssh.py` meanwhile. Are you still interested in working on this patch?

As far as I know, there isn't a policy yet on how to handle chained connections, but there have been recent discussions about it, and this might be a good reason to get it restarted.

Gu1:
Hi @amenonsen,
I've tried ported my code to ansible 1.9.x recently (as you can see on the standalone repo). I haven't had time to test it much, but it appears to work fine.
I must say I haven't looked at ansible 2.0 yet, but I'm still interested in maintaining the code, yeah.

Until now, the most problematic part of the patch is the fetch_file() method.
The only reliable way that I found to read a file from a container without temporary files is to just `cat` the file to stdout.
My custom connection handler is implemented on top of the default ssh connection plugin, which reads the whole stdout to memory before giving it back.
I had to implement a callback system for I/O in the ssh plugin. It allows me to process the incoming data while it's read from ssh's stdout, without buffering everything in memory.

bcoca:
@Gu1 if you look at the current jails/chroot plugins we implemented the copy using dd, as this also sidesteps several security issues. cc @abadger

Gu1:
@bcoca But in case the chroot plugin, you are working on the local machine, correct ?
The goal of my connection plugin is to manage lxc containers located on a remote server, by first connecting to the lxc host using ssh, and then attaching to the container to execute commands. In this case it's not possible to just dd from the container directly to the local machine.
Hoever, it's true that in my case `dd` might be a good alternative to `cat` to read the file to stdout..

jody-frankowski:
This is a feature I needed before. I really hope this can be merged before 2.0. @Gu1 Thank you for your awesome work.

sivel:
@Gu1,

Hi!

sivel:
@Gu1,

Hi!

amenonsen:
I'll close this PR now, since it hasn't been updated; but I do so reluctantly, because I really think it should be possible for Ansible to do this cleanly. @Gu1 if you feel like bringing this up on the development mailing list to discuss this further, that would be most welcome.

```
```

PR #9428, opened 2014-10-27
parse /etc/fstab on OpenBSD to get mount facts

landryb:
Pretty straightforward, copy/paste the code used by NetBSD and FreeBSD.. works here, but i question the usefulness of parsing `/etc/fstab` vs getting the information live from the system, parsing 'mount' command output ? I see that linux parses `/etc/mtab`, and keep only some entries (more or less `grep ^/ /etc/mtab | grep -v none` output), which should be more live/accurate data, no ?

Should the other operating systems (ie BSDs) also try to gather information from the running system and reflect more the actual 'facts', rather than 'what it should be according to fstab' ? Or facts should also mention/know about say, noauto mounts.. right now, parsing fstab here gives me a lot more entries than mount output. Should we also report NFS mounts ?

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -926,6 +926,7 @@ class OpenBSDHardware(Hardware):
         self.get_memory_facts()
         self.get_processor_facts()
         self.get_device_facts()
+        self.get_mount_facts()
         return self.facts
 
     def get_sysctl(self):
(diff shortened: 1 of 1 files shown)


landryb:
thinking a bit more, if we want to stay on the 'lets parse /etc/fstab', we should also skip some "special" entries such as

```
f756d7eccb8369dd.b none swap sw
f756d7eccb8369dd.d /altroot ffs xx 0 0
```

bcoca:
I did the original fstab parsing on FreeBSD, it was short sighted of me as
it always matched the output of mount on my machines.

I think facts should gather both, intended state and actual state, though
parsing the output of mount can get messy.

As for NFS, this is something that was recently requested of the main fact
parsing, I left it out virtual/remote filesystems on purpose but I see that
people might need this, but to avoid changing expected behavior I would add
it as a new key.

Just my 2c

On Tue, Oct 28, 2014 at 2:06 AM, Landry Breuil notifications@github.com
wrote:

> thinking a bit more, if we want to stay on the 'lets parse /etc/fstab', we
> should also skip some "special" entries such as
> 
> f756d7eccb8369dd.b none swap sw
> f756d7eccb8369dd.d /altroot ffs xx 0 0
> 
> —
> https://github.com/ansible/ansible/pull/9428#issuecomment-60726548.

## 

Brian Coca
Pedo mellon a minno

landryb:
But in that case, what should be done to stay backwards-compatible ? add a pair of boolean attributes with 'mounted:true/false' and 'remote:true/false' ?

landryb:
Can this be merged eventually ? leaving the fstab parsing vs live info question for later...

PR #9432, opened 2014-10-28
Fix chroot connection type.

sean-:
For reasons not fully understood, it is possible for os.path.normpath()
to return a path that contains more than one leading slash.  There is
an assumption in connection_plugins/chroot's put_file() that there is only one leading slash after
os.path.normpath(), however the code for os.path.normpath() makes it
clear that two slashes are possible.  To address this, change the order
in which the chroot(2)'ed path is constructed and normalize the result,
not the inputs to os.path.normpath().  Before:

</mnt> PUT /tmp/tmpjNYchB TO /.ansible/tmp/ansible-tmp-1414452814.97-43623097496280/setup

with additional debugging:

</mnt> out_path: //.ansible/tmp/ansible-tmp-1414453077.31-250891897990627/setup
</mnt> normpath: //.ansible/tmp/ansible-tmp-1414453077.31-250891897990627/setup
</mnt> in_path: /tmp/tmpagDDBL
</mnt> self.chroot: '/mnt'
</mnt> PUT /tmp/tmpagDDBL TO /.ansible/tmp/ansible-tmp-1414453077.31-250891897990627/setup

and after:

</mnt> PUT /tmp/tmpheJAEY TO /mnt/.ansible/tmp/ansible-tmp-1414454933.62-231262699030276/setup


sean-:
Bump.  Any chance this is going to get merged?

amenonsen:
This change looks good, with clearly described rationale, and I think it should have been merged. But looking at lib/ansible/plugins/connections/chroot.py now, I can't figure out where the change can be applied, or if it's even needed any more. @sean-, any chance you could have a look at this? In any case your PR needs to be cleaned up to include only the commit you actually want to merge.

bcoca:
this PR is against the v1 tree which is dead code ttps://github.com/ansible/ansible/blob/devel/v1/README.md and the new plugin has departed enough that this does not apply anymore, so I'm closing this PR.

PR #9433, opened 2014-10-28
Cache the results for the RackSpace dynamic inventory.

intellectronica:
This patch to the RackSpace dynamic inventory plugin makes it cache results for 10m (similar to the EC2 plugin). Running the complete API call for large sites every time can be very slow. Like the EC2 inventory you can force-refresh the cache by running with --refresh-cache, or by deleting the cache file from the temp dir.

@angstwad @j2sol @sivel

### DIFF
--- a/plugins/inventory/rax.py
+++ b/plugins/inventory/rax.py
@@ -167,7 +167,6 @@ except ImportError:
     print('pyrax is required for this module')
     sys.exit(1)
 
-from tempfile import gettempdir
 from time import time
 
 
@@ -345,7 +344,10 @@ def _list_into_cache(regions):
 
 def get_cache_file_path(regions):
     regions_str = '.'.join([reg.strip().lower() for reg in regions])
-    return os.path.join(gettempdir(),
+    ansible_tmp_path = os.path.join(os.path.expanduser("~"), '.ansible', 'tmp')
+    if not os.path.exists(ansible_tmp_path):
+        os.makedirs(ansible_tmp_path)
+    return os.path.join(ansible_tmp_path,
                         'ansible-rax-%s-%s.cache' % (
                             pyrax.identity.username, regions_str))
 
--- a/plugins/inventory/rax.py
+++ b/plugins/inventory/rax.py
@@ -339,23 +339,25 @@ def _list_into_cache(regions):
     if hostvars:
         groups['_meta'] = {'hostvars': hostvars}
 
-    with open(get_cache_file_path(), 'w') as cache_file:
+    with open(get_cache_file_path(regions), 'w') as cache_file:
         json.dump(groups, cache_file)
 
 
-def get_cache_file_path():
+def get_cache_file_path(regions):
+    regions_str = '.'.join([reg.strip().lower() for reg in regions])
     return os.path.join(gettempdir(),
-                        'ansible-rax-{}.cache'.format(pyrax.identity.username))
+                        'ansible-rax-%s-%s.cache' % (
+                            pyrax.identity.username, regions_str))
 
 
 def _list(regions, refresh_cache=True):
-    if (not os.path.exists(get_cache_file_path()) or
+    if (not os.path.exists(get_cache_file_path(regions)) or
         refresh_cache or
-        (time() - os.stat(get_cache_file_path())[-1]) > 600):
+        (time() - os.stat(get_cache_file_path(regions))[-1]) > 600):
         # Cache file doesn't exist or older than 10m or refresh cache requested
         _list_into_cache(regions)
 
-    with open(get_cache_file_path(), 'r') as cache_file:
+    with open(get_cache_file_path(regions), 'r') as cache_file:
         groups = json.load(cache_file)
         print(json.dumps(groups, sort_keys=True, indent=4))
 
--- a/plugins/inventory/rax.py
+++ b/plugins/inventory/rax.py
@@ -167,6 +167,10 @@ except ImportError:
     print('pyrax is required for this module')
     sys.exit(1)
 
+from tempfile import gettempdir
+from time import time
+
+
 NON_CALLABLES = (basestring, bool, dict, int, list, type(None))
 
 
@@ -214,7 +218,7 @@ def host(regions, hostname):
     print(json.dumps(hostvars, sort_keys=True, indent=4))
 
 
-def _list(regions):
+def _list_into_cache(regions):
     groups = collections.defaultdict(list)
     hostvars = collections.defaultdict(dict)
     images = {}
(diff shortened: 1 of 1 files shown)


cchurch:
:-1: as implemented.

Ansible itself will not be able to pass the --refresh-cache parameter when it runs the inventory script; the only argument it will use is --list.  Additional inventory script options should probably be passed via environment variables or a separate config file.

For Tower usage, it is imperative that we can explicitly specify a path for the cache or disable it entirely, so that multiple inventories accessing different Rackspace accounts will each use their own cache file.

intellectronica:
@cchurch thanks for looking at the patch.

I think it would be good to keep the interface (--refresh-cache) since it's identical to how the EC2 plugin behaves, and that's arguably one of the most popular dynamic inventory plugins. It does make sense to keep the cache separated for different account. Would such a change be sufficient for landing this?

intellectronica:
b.t.w the EC2 plugin does not separate the cache for different accounts.

intellectronica:
Updated the plugin to save the cache in a file specific to rackspace account being used.

sivel:
I will review this soon and get back to you with additional feedback.

plugins/inventory/rax.py
```diff
@@ -202,7 +205,26 @@ def _list(regions):
 
     if hostvars:
         groups['_meta'] = {'hostvars': hostvars}
-    print(json.dumps(groups, sort_keys=True, indent=4))
+
+    with open(get_cache_file_path(), 'w') as cache_file:
+        json.dump(groups, cache_file)
+
+
+def get_cache_file_path():
```
sivel: I think I would prefer that this file didn't get dumped into a tmp directory.

plugins/inventory/rax.py
```diff
@@ -202,7 +205,26 @@ def _list(regions):
...
+    return os.path.join(gettempdir(),
+                        'ansible-rax-{}.cache'.format(pyrax.identity.username))
```
sivel: This should also support whether a regions were limited via `RAX_REGIONS` or `~/.pyrax.cfg`

sivel: This line is also not python 2.6 compatible.  In general printf like formatting would be preferred over `.format()`

sivel:
In addition to the above comments, I'm not a huge fan of file based caches like this.  I know it is somewhat the standard now that a few inventory plugins support file based caches, however I would much rather prefer a cache that is implemented using memcached or redis.  Maybe this is a discussion that should be held on ansible-devel.

plugins/inventory/rax.py
intellectronica: Where would be a better place to put this?

Keep in mind that this is highly ephemeral - the file is no longer relevant after 10m. Also it's optional, if the file isn't there everything will still work, only slower. Finally, this is exactly what equivalent inventory plugins (like the EC2 and DO ones) do, so users will be used to it.

plugins/inventory/rax.py
intellectronica: Good points. Addressed in a new commit.

intellectronica:
@sivel thanks a lot for your review and suggestions.

I have modified the cache path so that it includes the regions and made it use old-school string interpolation so that it's compatible with older Python versions.

As for your other concerns, I'm not sure there are things I think should be addressed as part of this iteration. It makes sense to cache the results locally to where Ansible is being run. The cache is a mere optimization, everything will still work if it's not available, so there's really no problem with other machines or users not having access to it. Also, it would be quite burdensome if users had to run Redis or Memcache to be able to run quicker - part of the magic of Ansible is that it's so easy to get it to work with only SSH from your local machine.

plugins/inventory/rax.py
```diff
@@ -202,7 +205,28 @@ def _list(regions):
...
         groups['_meta'] = {'hostvars': hostvars}
-    print(json.dumps(groups, sort_keys=True, indent=4))
+
+    with open(get_cache_file_path(regions), 'w') as cache_file:
+        json.dump(groups, cache_file)
+
+
+def get_cache_file_path(regions):
+    regions_str = '.'.join([reg.strip().lower() for reg in regions])
+    return os.path.join(gettempdir(),
```
sivel: The cache path used by DO and ec2 is configurable.  I am in the process of adding a config file for `rax.py`.  The default in `ec2.ini` is `~/.ansible/tmp` in `digital_ocean.ini` it is `/tmp`.  I'm still not a huge fan of the default being `/tmp`

Perhaps for now we can just hard code it to  `~/.ansible/tmp` and make it configurable later.

intellectronica: OK, I'll modify it to write to `~/.ansible/tmp` and once you introduce a config file we can make that configurable.

intellectronica: Moved the cache to ~/.ansible/tmp

intellectronica:
@sivel @cchurch wdyt of the patch in its current form? I think I have incorporated most of your suggestions. I realise this patch might not be the best solution to the problem, but since it is in line with equivalent plugins and works well (we've been using it for a while to manage our RS servers) I think it will be good to land it. I'm happy to be involved in helping with improving on it later.

sivel:
I will look over this more and get feedback from others soon.

Keep in mind that while this hasn't been prioritized yet, it is likely a P3 or P4.  Due to the priority queue and working higher priority issues, it could be a significant wait before this is merged.

intellectronica:
Cheers @sivel. No rush, we don't depend on it in any way since we anyway copy the plugin. Just wanted to make sure this makes it in eventually to benefit others.

sivel:
@intellectronica A PR was recently merged that added a config file for rax.py.  Additionally this PR now contains merge conflicts that will need to be resolved.  Let me know if you have any questions.

intellectronica:
Thanks for the heads up @sivel. Conflicts resolved.

amenonsen:
@sivel any news about this? It's languished for 8+ months after @intellectronica made all the requested changes. I don't use RAX, so I can't test/review, but the patch still seems relevant in devel, and though it doesn't apply any more, it doesn't look like it would be too hard to merge.

bcoca:
@intellectronica if you can rebase against the new location contrib/inventory/rax.py i might be able to merge this soon

amenonsen:
@bcoca since (a) @intellectronica has not replied (and after 8+ months, that's not exactly surprising), (b) the rebase was trivial and could be automatically handled by git's 3-way merge, (c) you indicated that you could merge this soon, I've rebased and pushed this as #11923 (with original attribution retained). This PR can be closed.

intellectronica:
@amenonsen sorry for missing it earlier and thanks a lot for taking this on. exciting to see this finally making it in!

sivel:
Closing in favor of https://github.com/ansible/ansible/pull/11923

PR #9436, opened 2014-10-28
vars_files haven't been discussed yet

bobobox:
I think this is a chapter ordering thing... vars_files have not actually be explained yet (aside from a reference to them in the note right above, which also might need reconsidering?). I believe they do get introduced in the next section, 'Variables'.


jimi-c:
Merged, thank you!

PR #9444, opened 2014-10-29
As per yum module documents it state only support 'present, absent and latest'

praveenkumar:
Current intro doc is pointing older way of using yum module.
Yum module documents -> http://docs.ansible.com/yum_module.html


abadger:
Merged, thanks!

PR #9445, opened 2014-10-29
Read AWS key and secret from `ec2.ini`

jessedhillon:
If you have multiple projects which use AWS, it's not convenient to use global- or user-wide boto configs. You could use environment variables, but since we have ec2.ini right there, it seems like an easy place to have your project's AWS credentials.

I had to resubmit this PR, previously it was https://github.com/ansible/ansible/pull/9360


amenonsen:
I think this PR should be closed in favour of the others that make it possible to use multiple Boto profiles. I don't think it's a good idea to allow authentication credentials to be stored in `ec2.ini` anyway (I at least tend to keep it in source control along with the playbook).

amenonsen:
Closing per the above comment, especially since the functionality to explicitly specify a profile to use was merged in the interim.

PR #9446, opened 2014-10-29
Adds localhost to ec2 inventory

jessedhillon:
`localhost` is important to have in your inventory when running a deployment on AWS: it's the host on which your provisioning plays typically run. So if `ec2.py` doesn't provide a description of `localhost` then one has to create a trivial inventory file just to have localhost. This patch adds `localhost` to the EC2 inventory script's output.

### DIFF
--- a/plugins/inventory/ec2.py
+++ b/plugins/inventory/ec2.py
@@ -133,7 +133,17 @@ except ImportError:
 
 class Ec2Inventory(object):
     def _empty_inventory(self):
-        return {"_meta" : {"hostvars" : {}}}
+        return {
+            "_meta": {
+                "hostvars": {}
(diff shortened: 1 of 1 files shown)


jimi-c:
Hi @jessedhillon, as of a few versions ago, the `localhost` entry is implicit in inventory, and should not need to be specified. Are you continuing to see this as an issue in 1.7 or the devel branch?

amenonsen:
This PR can be closed. I've verified that «ansible -i ec2.py localhost -m ping» works, and there's been no response to the request for more info for ~8 months.

bcoca:
closing as this is implemented in the inventory class and there is no need to add to the scripts.

PR #9450, opened 2014-10-29
removes extension from ansible-doc output

bcoca:
it also fixes issue with mismatch on (doc) blacklisted module matching.


bcoca:
folding this into load_aliases changes, no need for separate request

PR #9451, opened 2014-10-29
Load aliases

bcoca:
This introduces the capability of creating module aliases and/or hiding modules as 'deprecated'

### DIFF
--- a/docsite/rst/developing_modules.rst
+++ b/docsite/rst/developing_modules.rst
@@ -466,7 +466,7 @@ Included modules will ship with ansible, and also have a change to be promoted t
 gives them slightly higher development priority (though they'll work in exactly the same way).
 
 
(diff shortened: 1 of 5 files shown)


docsite/rst/developing_modules.rst
```diff
@@ -465,6 +465,23 @@ a github pull request to the `extras <https://github.com/ansible/ansible-modules
 Included modules will ship with ansible, and also have a change to be promoted to 'core' status, which
 gives them slightly higher development priority (though they'll work in exactly the same way).
 
+
+Deprecating and makingm module aliases
```
erjohnso: s/makingm/making/

erjohnso:
:dancers: Love it, thanks Brian!

PR #9454, opened 2014-10-29
Added HDD serial numbers.

ghost:
Added HDD serial numbers to facts.py.

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -774,7 +774,7 @@ class LinuxHardware(Hardware):
                 return
             serial = re.search("Unit serial number:\s+(\w+)", drivedata)
             if serial:
-                d[key] = serial.group(1)
+                d['serial'] = serial.group(1)
 
             for key in ['vendor', 'model']:
                 d[key] = get_file_content(sysdir + "/device/" + key)
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -765,6 +765,18 @@ class LinuxHardware(Hardware):
                     continue
             d = {}
             diskname = os.path.basename(sysdir)
+
+            sg_inq = module.get_bin_path('sg_inq') 
+            for key in ['serial']:
+                device = "/dev/%s" % (block)
+                try:
+                    rc, drivedata, err = module.run_command([sg_inq, device])
+                except:
+                    return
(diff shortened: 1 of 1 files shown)


Jmainguy:
I can confirm this does work and find the serial. It requires sg_inq to be installed on the target to gather HDD specific facts, if it is not installed it just doesn't provide them, does not spit out any ugly errors.

Good job @jlstanton  keep it up.

jimi-c:
@jlstanton I would go ahead and remove the for loop, as it doesn't do anything in this case. Beyond that the patch looks fine.

ghost:
@jimi-c, good point. I removed the for loop.

jfchevrette:
Hmm I could actually have used that feature yesterday when I had to identify disks on about ~100 servers.
LGTM

ghost:
.

amenonsen:
Looks fine to me.

lib/ansible/module_utils/facts.py
```diff
@@ -765,6 +765,17 @@ def get_device_facts(self):
                     continue
             d = {}
             diskname = os.path.basename(sysdir)
+
+            sg_inq = module.get_bin_path('sg_inq') 
+            device = "/dev/%s" % (block)
+            try:
+                rc, drivedata, err = module.run_command([sg_inq, device])
+            except:
+                return
```
bcoca, line 774: this should not be a return as it breaks the whole for loop

lib/ansible/module_utils/facts.py
```diff
@@ -765,6 +765,17 @@ def get_device_facts(self):
                     continue
             d = {}
             diskname = os.path.basename(sysdir)
+
+            sg_inq = module.get_bin_path('sg_inq') 
+            device = "/dev/%s" % (block)
+            try:
```
bcoca, line 771: add if that skips it all if sq_inq is not found

amenonsen:
I'd like to get this merged. @bcoca's comment about the return was correct, but no longer applies because the for loop has been removed, but his other comment about not trying to execute sg_inq if get_bin_path returns None still applies. @stantonnet sorry it's taken so long to get to this, but do you feel like making that change?

amenonsen:
Anyone else using sg_inq interested in updating this patch?

Jmainguy:
Sure, I will take a stab at updating this, will update here once done, if someone beats me to it, that is also very cool

mattclay:
Closing since this was rebased and merged in #22852.

PR #9461, opened 2014-11-01
Simpler source dev env bootstrap script

pauloppenheim:
When the tutorial suggested

```
$ sudo pip install paramiko PyYAML Jinja2 httplib2
```

And

```
$ source ./hacking/env-setup
```

I thought
1. I'm a py dev, I don't want that in my system python
2. I use long-lived shells, I don't want to hose my shell

So I wrote a higher level dev wrapper before continuing the tutorials. From a clean checkout and a clean ubuntu VM with only virtualenv installed (or otherwise available locally!), I can:

```
$ alias ansible='~/src/ansible/hacking/ansible-dev -i ~/ansible-tutorial/hosts'
$ ansible all -m ping
<.... lots of virtualenv output on first run ...>
fork.nodes.foostache.com | success >> {
    "changed": false, 
    "ping": "pong"
}
```

No installing deps to try ansible. Virtualenv is a single file, so could be put into `hacking/` if that is too much dep.


pauloppenheim:
Less verbose, so that actual ansible output isn't hidden by the output of setup commands

pauloppenheim:
I needed to use this again today. For easier merging with less thinking, this should be split into two parts - an automated virtualenv part, and an alternative to sourcing part. I will do that.

PR #9462, opened 2014-11-01
Setting keepalive option before opening session

tkinz27:
Setting the keepalive option before opening up the paramiko session
in order to avoid the slim chance that the connection is stalled
in the short time between opening the session and setting up the
keepalive. I described the issue I am solving at
https://groups.google.com/forum/#!topic/ansible-project/rkwvz3vyvLk


jimi-c:
Merged, thank you!

PR #9463, opened 2014-11-01
Adds --force-read option to force ansible to read instead of execute executable files.

stephenwithav:
This option short-circuits the is_executable check by adding a `force_read` parameter to the `Inventory` and `InventoryDirectory` constructors.

If --force-read is given as a command-line option to either `ansible` or `ansible-playbook`, skips the `is_executable` check.


stephenwithav:
Use case:

`vagrant up` automatically creates an inventory file when ansible provisioners are used.  If your vagrant VMs are on an NTFS drive, the inventory file can't be read.

(I have a patch to vagrant that adds support for this option, too.)

mpdehaan:
Thanks for this submission.

Unfortunately, this isn't something we want as we do not want to clutter up ansible with a lot of confusing command line flags.

It seems the solution here would be for Vagrant to write out inventory (INI file, etc) and not rely on a python dynamic inventory file.

--force-read is confusing because what it means is really "--interpret-as-script-even-if-no-chmod-plus-x"

An alternate approach might be to instead note the shebang line at the top of the file, pull the line out of the file, and execute it - however, I worry that people would file security tickets on that implementation.

Definitely can continue discussion on the mailing list if you are interested.

stephenwithav:
Michael, thank you for taking a look at this.

If I may, --force-read actually means --interpret-as-static-even-if-chmod-plus-x.

If you would like, I can record an animated gif showing where the existing workflow fails on an NTFS drive and how --force-read solves the problem.

Apologies for continuing the discussion here instead of on the mailing list.  My thread there hasn't seen any activity since my last post a few days ago, so I didn't want it to get lost.

bcoca:
This seems like a very corner case, I don't think most people use inventory from an NTFS formatted USB drive from which they also use executables.

stephenwithav:
Fair enough, I suppose.

I'll try to think of some other solutions.

stephenwithav:
One last question, if I may: if I rewrote this as an inventory plugin (e.g., static.py), would you want me to send a pull request so you can include it or would it not be of interest to you?

PR #9465, opened 2014-11-01
Fix TypeError exception if sudo password is integer

yoojinl:
Hi, I have a virtual machine for testing where sudo password is integer

Inventory file:

```
[staging]
192.168.0.100 ansible_ssh_port=2222 ansible_ssh_user=ubuntu ansible_ssh_pass="123456" ansible_sudo_pass="123456"
```

When I run anisble, it fails with the next exception

```
fatal: [192.168.0.100] => Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/ansible/runner/__init__.py", line 561, in _executor
  File "/usr/local/lib/python2.7/dist-packages/ansible/runner/__init__.py", line 666, in _executor_internal
  File "/usr/local/lib/python2.7/dist-packages/ansible/runner/__init__.py", line 884, in _executor_internal_inner
  File "/usr/local/lib/python2.7/dist-packages/ansible/runner/action_plugins/normal.py", line 57, in run
    return self.runner._execute_module(conn, tmp, module_name, module_args, inject=inject, complex_args=complex_args)
  File "/usr/local/lib/python2.7/dist-packages/ansible/runner/__init__.py", line 526, in _execute_module
    res = self._low_level_exec_command(conn, cmd, tmp, sudoable=sudoable, in_data=in_data)
  File "/usr/local/lib/python2.7/dist-packages/ansible/runner/__init__.py", line 1025, in _low_level_exec_command
    in_data=in_data)
  File "/usr/local/lib/python2.7/dist-packages/ansible/runner/connection_plugins/ssh.py", line 352, in exec_command
    stdin.write(self.runner.sudo_pass + '\n')
TypeError: unsupported operand type(s) for +: 'int' and 'str'
```

### DIFF
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -846,7 +846,7 @@ class Runner(object):
         actual_private_key_file = template.template(self.basedir, actual_private_key_file, inject, fail_on_undefined=True)
         self.sudo = utils.boolean(inject.get('ansible_sudo', self.sudo))
         self.sudo_user = inject.get('ansible_sudo_user', self.sudo_user)
-        self.sudo_pass = inject.get('ansible_sudo_pass', self.sudo_pass)
+        self.sudo_pass = str(inject.get('ansible_sudo_pass', self.sudo_pass))
         self.su = inject.get('ansible_su', self.su)
         self.su_pass = inject.get('ansible_su_pass', self.su_pass)
         self.sudo_exe = inject.get('ansible_sudo_exe', self.sudo_exe)


jimi-c:
Hi @rustyrobot, I believe a simpler fix for this may be to simply force the sudo_pass variable to a str, rather than changing the formatting in multiple places. I would look at trying to do this when the sudo_pass member variable is set in the runner class.

yoojinl:
Thanks @jimi-c for your comment, updated and tested the patch.

yoojinl:
@jimi-c let me know if there is anything I can improve in this patch

jimi-c:
@rustyrobot it looks good, it has been prioritized and we will get to it as soon as possible. In the meantime, if this is causing a problem for you now, just be sure to put quotes around values in your playbook and the YAML interpretter will automatically bring them in as strings:

<pre>
   sudo_password: "your password here"
</pre>

Thanks!

amenonsen:
The code in devel has changed quite a bit since this PR was opened, and now presumably the change should be applied in lib/ansible/executor/connection_info.py, but it's not really clear to me how it should be done. @jimi-c, any thoughts?

jimi-c:
@amenonsen thinking about this, we should probably move ConnectionInformation from executor -> playbook and make it a subclass of the playbook Base class, so that it does proper type-checking on the attributes in `post_validate`. 

Right now, the Base class does that, so any fields that come from the Play/Task will be strings, however things coming from Host or Group in inventory won't. Also, since the ConnectionInformation `post_validate` simply does templating, if those fields are set to variables the resulting template could result in any type.

We have had discussions in the past about renaming the ConnectionInformation object to something like PlayContext (because that's really what it has become), so maybe this would be the time to do it.

amenonsen:
I think this PR can be closed now. @jimi-c committed the PlayContext change, the _post_validate function in playbook/base.py does the required type checking/conversion, and I can no longer reproduce a TypeError when setting ansible_sudo_pass to an integer.

bcoca:
in current devel type check is now built into the basic classes so this PR is not needed anymore.

PR #8528, opened 2014-08-08
'copy' write new line at end of file if 'content' passed without newline...

adamchainz:
Fixes #7855. Checked it by using hacking/env-setup and creating a temporary file with ad-hoc ansible.

Issue #7855, opened 2014-06-19
content argument of copy module doesn't add end of line character


ryan-lane:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.6.3
##### Environment:

Ubuntu 14.04
##### Summary:

When using content='blah' for a file, the file will have 'blah' with no end of line character, which is generally an invalid file.
##### Steps To Reproduce:

``` yaml
- name: Add world to /tmp/hello
  copy: dest=/tmp/hello content='world'
```
##### Expected Results:

/tmp/hello:

```
world<end-of-line>
```
##### Actual Results:

/tmp/hello:

```
world
```

jimi-c:
The content is verbatum, so just add a '\n' to it if you want a newline. Files without newlines are not "generally invalid", as there are plenty of config files that do not end with a newline. If you have any further questions regarding this, please let us know. Thanks!

ryan-lane:
http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_205

POSIX states that text files should end with a newline to specify end of file. It might be a problem for this to be the default for people writing binary into their files via content (do people actually do this?) but for the vast majority of cases the module is likely writing POSIX in-compliant files.

jimi-c:
Feel free to send in a PR to check for a newline at the end of content, and to add one if it doesn't exist.

PR #9468, opened 2014-11-03
 'copy' write new line at end of file if 'content' passed without newlin...

landryb:
...e.

POSIX states that a file should end by a newline.
http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206

For previous discussion/github issue, refer to https://github.com/ansible/ansible/pull/8528 / https://github.com/adamchainz/ansible/commit/842a1b976d650b2c3117a6e0ed1329e5d0d62873 and https://github.com/ansible/ansible/issues/7855

### DIFF
--- a/lib/ansible/runner/action_plugins/copy.py
+++ b/lib/ansible/runner/action_plugins/copy.py
@@ -316,6 +316,8 @@ class ActionModule(object):
         f = os.fdopen(fd, 'w')
         try:
             f.write(content)
+            if not content.endswith('\n'):
+                f.write('\n')
         except Exception, err:
             os.remove(content_tempfile)
             raise Exception(err)


jfchevrette:
I'm not sure about that one. 

If a file I copy to a remote server does not end with a newline it's no one's business and especially not my automation tool which I expect will copy the file as is and nothing more unless explicitly specified as a parameter. 

If a server admin care that much about POSIX compliance, why not add the newline to the source file? Am I missing something?

Again ansible should not modify my file content without my explicit intent. If enough people care about this, maybe it could be added as module parameter?

landryb:
This is _only_ in the case where content argument is passed. It's not the server admins that care about POSIX compliance, its the underlying daemons behind, which often fail in surprising ways if their config file (which can be a oneliner filled by copy+content...) doesnt end with a newline.

jfchevrette:
@landryb ah! I completely missed the 'content' part :+1:

adamchainz:
+1 was just looking at reopening this. The OpenBSD package maintainer emailed me, sounds like he's been bitten by this in the past and applies this patch to the OpenBSD package.

landryb:
Yes, that patch is applied to our port, see http://cvsweb.openbsd.org/cgi-bin/cvsweb/ports/sysutils/ansible/patches/patch-lib_ansible_runner_action_plugins_copy_py?rev=1.1&content-type=text/x-cvsweb-markup

adamchainz:
Was actually just bitten by this. Writing out some `/etc/cron.d/` files with a two-line micro template with content, `cron` refuses to load them.

Task (simplified):

``` yaml
- name: cron.d items
  copy:
    dest: /etc/cron.d/{{ item.name }}
    content: |
      MAILTO=me@adamj.eu
      DJANGO_SETTINGS_MODULE=settings.production
      {{ item.schedule }} app /opt/app/manage.py {{ item.name }} 2>&1 >>/var/log/app/{{ item.name }}.log
  with_items:
    - {name: clearsessions, schedule: '0 * * * *'}
```

Complaint from cron in `/var/log/syslog`:

```
Nov 21 13:33:32 ip-10-209-21-179 cron[16950]: (*system*clearsessions) ERROR (Missing newline before EOF, this crontab file will be ignored)
```

Inserting a new line at the end of the content block doesn't work (yaml ignores it), and you can't write a newline inside the content block. The workaround I found is to add `{{ '\n' }}` to the end of the content block which is just jinja2 being helpfully flexible.

amenonsen:
I submitted a rebased PR #11749, so this can be closed now.

bcoca:
closing as per above

Issue #9464, opened 2014-11-01
[bug report] Possible issue with ignore_errors and variables


zachsnow:
### Summary:

Difference in ability to use variables with `ignore_errors` from version 1.5.5 vs. successive versions (including 1.7.2) on OSX.
### Expected:

In Ansible 1.5.5 one could use variable names in `ignore_errors`, which we used to dynamically allow ignoring or not ignoring errors.  For example:

```
# inventory.hosts
localhost ansible_connection=local

# group_vars/all.yml
test_variable: no

# playbook.yml
- name: Issue (playbook)
  hosts: all
  tasks:
    - name: Show variable
      debug:
        var: test_variable
    - name: Use variable in 'when'
      debug:
        msg: Testing...
      when: test_variable
    - name: Use variable in 'ignore_errors'
      fail:
        msg: Failing...
      ignore_errors: test_variable

```

Behavior in 1.5.5 is as expected: the first task shows the variable's value, the second is skipped, and the third ignores errors and so succeeds despite the use of `fail`.

(Note that in my actual scenario I have a "real" inventory; this simple inventory is solely for purposes of illustrating the issue).
### Actual:

In versions later than 1.5.5 (I believe I tested them all, and certainly including 1.7.2) this results in an error:

`ERROR: error while evaluating conditional: test_variable`.

However, the first task correctly displays the contents of `test_variable`, and the second is correctly skipped:

```
TASK: [Show variable] *********************************************************
ok: [localhost] => {
    "test_variable": "False"
}

TASK: [Use variable in 'when'] ************************************************
skipping: [localhost]
```

I tried various different syntaxes and concluded there's something amiss; `{{ test_variable|bool }}` yields `jinja2.exceptions.UndefinedError: 'test_variable' is undefined` which is seemingly false. Finally I dropped `import ipdb; ipdb.set_trace()` [around here](https://github.com/ansible/ansible/blob/release1.7.2/lib/ansible/playbook/__init__.py#L467) and found that the variable isn't available in the set of those being used when evaluating the conditional after all, and that we are using different logic to process `when` as compared to `ignore_errors` entirely.

mpdehaan:
First off, an undefined variable is undefined, not false.

However, since you have it defined in "all", it's most likely defined, and inventory variables aren't being pulled in here, which does make this a bug, most likely.

Thanks for the report!

zachsnow:
"Seemingly false" means "apparently not the case". Note the lack of backticks.

[...]

PR #9476, opened 2014-11-03
Address ignore_errors issue

zachsnow:
Compute the value of `ignore_errors` at the right moment (the same as `when`) so that all the necessary variables are available. One upshot is that `ignore_errors` becomes per-host/per-iteration as opposed to affecting the task as a whole; this might not be desirable but I believe that it matches the old (1.5.5) behavior at least (see #9464).

### DIFF
--- a/lib/ansible/callbacks.py
+++ b/lib/ansible/callbacks.py
@@ -212,10 +212,12 @@ class AggregateStats(object):
         prev = (getattr(self, what)).get(host, 0)
         getattr(self, what)[host] = prev+1
 
-    def compute(self, runner_results, setup=False, poll=False, ignore_errors=False):
+    def compute(self, runner_results, setup=False, poll=False):
         ''' walk through all results and increment stats '''
 
         for (host, value) in runner_results.get('contacted', {}).iteritems():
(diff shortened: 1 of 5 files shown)


zachsnow:
Is there any point in me updating this pull request? It languished for months while it was ready to be merged (or at least reviewed).

bcoca:
sorry for the delay, we have improved on the merging process and added more bodies, but we have a huge backlog and are working to get through it.

that said, yes, we would like this feature and a rebase would be needed before we can merge it.

zachsnow:
All right I'll have a look, cheers.

bcoca:
actually, we are in the middle of releasing 1.9, you might want to wait a bit until we put in the v2 as the main development tree

amenonsen:
@zachsnow devel is now the v2 codebase. Feel like having a go at rebasing your change?

sivel:
@zachsnow,

Hi!

amenonsen:
Closing per above comment. @zachsnow you're welcome to open a new PR if you get a chance to rework the feature. Sorry again it didn't get reviewed in time originally.

PR #9122, opened 2014-09-24
Initial ansible_ssh_proxy support for ssh transport

sivel:
This pull request adds functionality to supply a host to use with the SSH ProxyCommand as requested in #8963

Currently this only:
1. Adds support for the `ssh` transport
2. Only accepts an IP or hostname, not the full ProxyCommand (The ProxyCommand utilizes `ssh -W`)
3. Only keys off of `ansible_ssh_proxy` as a "global" var (--extra-vars/vars_files/etc...) or from inventory (group_vars and host_vars included)

PR #9477, opened 2014-11-04
Inventory-based SSH proxy/bastion host support, ssh_args + docs

nitzmahone:
Supersedes #9122 

First-class inventory var support for proxy/bastion/jump hosts seems to be increasingly requested. I've tried to make it reasonably comprehensive and well-documented, though I intentionally omitted proxy_password support (had some trouble stacking the file descriptors for the two-level-deep sshpass).

(CC @sivel)


jimsmith:
+1 for this, it's a request that I rarely get asked but it's becoming more frequently.

mpdehaan:
So this has been discussed a bit recently on the devel list, and I believe there may be some competing implementations, can you stop by ansible-devel to discuss?

Thanks!

tzz:
I've needed this so definitely happy to see an implementation!

quinot:
I'm just starting to use Ansible and was wondering whether this feature existed, or how to implement it if not. Would love to have this merged!

rmetzler:
I really want this feature in Ansible, but after looking at the code, I would think you shouldn't add that many arguments to `connect`. Maybe use `*args` and `**kwargs`.

nitzmahone:
Funny, that was my first thought when I looked at it too- "why are they passing all these args positionally?" I looked at the history, and the last couple of args were added later. I assumed there was some reason to do with the abstraction of the Connect stuff, and went for the "when in Rome" approach, but if reviewers feel the same way, I'm happy to convert it to use named args instead. It _is_ much cleaner, IMHO.

rmetzler:
I want to say, I just said this as a user who needs this feature and searched through the different threads / PRs. I really don't have any authority regarding Ansible.

igoratencompass:
I've been waiting for this kind of module definitely needed for dynamic aws inventory of mixed public and private instances! Can we please have this released asap???

christophgysin:
I'm also looking forward to this!

lib/ansible/runner/connection_plugins/ssh.py
```diff
@@ -58,10 +63,35 @@ def connect(self):
...
+            proxy_args += ["-o", "KbdInteractiveAuthentication=no",
+                           "-o", "PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey",
+                           "-o", "PasswordAuthentication=no"
+            ]
+
+            if self.proxy_user is not None and self.proxy_user != pwd.getpwuid(os.geteuid())[0]:
+                proxy_args += ["-o", "User="+self.proxy_user]
+            proxy_args += ["-o", "ConnectTimeout=%d" % self.runner.timeout]
+
+            self.common_args += ["-o", "ProxyCommand=ssh {proxy_host} {proxy_args} -W %h:%p".format(proxy_host=self.proxy_host, proxy_args=" ".join(proxy_args))]
```
christophgysin, line 92: ssh -W asumes openssh-5.4 or later. Older versions of openssh need netcat to achieve this.

nitzmahone, line 92: Yeah, I considered using netcat as a lowest-common-denominator approach. However, I saw some complaints around the openssh community about instability with various versions of netcat that the -W mode attempts to address (don't have a reference to share off the top of my head- it was several months ago). If a bastion host can't be upgraded to a reasonably recent version of openssh, the ssh_args escape hatch is still there- I think that beats the alternatives of:
- forcing everyone to use a potentially unstable approach that has another dependency (netcat)
- adding yet another mode-changing variable (ansible_ssh_proxy_mode=netcat or some such blah)
- version-sniffing ssh (rife with issues)

anthcourtney:
Is there a timeline for this feature progressing / being integrated?

rmetzler:
For everyone needing a workaround fast, this is a great tutorial: http://alexbilbie.com/2014/07/using-ansible-with-a-bastion-host/

But I also would prefer first class support in Ansible.

mpeters:
That approach only works when you have a single bastion server for all your environments. This won't work if you have a separate bastion for say production, staging and integration.

bcoca:
@mpeters it does, you just need to specify the diff bastions per host or host pattern (this is where naming conventions can help a lot).

mpeters:
That's assuming you are naming your servers. How would you handle just IP addresses (since these are livestock, not pets) in different VPCs? The IPs can be repeated for a different environment. 

Also, how would you hook this up with Tower? (You'd need multiple Tower installs right?) Multiple Tower installs? 

What about a single environment that uses multiple VPCs for PCI, etc?

bcoca:
no need for multiple ansible/tower installs, one thing you can do is set ssh args per group/host/inventory, passing specific ssh config for each.

In any case, this is a workaround, the patch is a good step into a more integrated solution.

mpeters:
Can you really set ansible_ssh_args per group/host/inventory (that would be great)? The documentation doesn't list it with the rest of the ansible_ssh_\* args (http://docs.ansible.com/intro_inventory.html#list-of-behavioral-inventory-parameters) and the only place I see you can set it is in the ansible config file, which is global.

rmetzler:
I had a quick look into the source and it seems like `ssh_args` is respected in 1.8.2. Not sure since when, though.

https://github.com/ansible/ansible/blob/v1.8.2/lib/ansible/constants.py#L165
https://github.com/ansible/ansible/blob/v1.8.2/lib/ansible/runner/connection_plugins/ssh.py#L55-L64

mpeters:
But that just shows that it's configurable globally from the ansible configuration file. Not that it's possible to specify per group/host/inventory, right?

snackycracky:
I also need the 'Ansible EC2 external inventory script' to detect the internal hosts. http://docs.ansible.com/intro_dynamic_inventory.html#example-aws-ec2-external-inventory-script
it only finds the jump-hosts for each environment but not the internal servers.

dopry:
@bcoca,  Are you comfortable with this approach to bastion host support, or is there a competing implementation you prefer? I'm currently working on a VPC deployment and can put some cycles in testing and getting the PR up to date if you guys are interested in getting it merged.

eriweb:
Thanks @nitzmahone, I'm testing this, but I seem to have trouble getting Ansible to wait for the proxycommand to finish.
I get this error message blazingly fast: "fatal: [r-4-VM] => SSH Error: data could not be sent to the remote host. Make sure this host can be reached over ssh"
but if I try the same SSH command manually it works just fine..

bcoca:
I was thinking about integrating this soon now that v2 is almost 'there', but needs massive rebase, should probably wait till v2 is in place in devel

bcoca:
the other thing I wanted to do is move away from having ansiible_ssh and go for ansible_connection (or at least conn) as the variable naming, so we can actually use the same vars for other protocols w/o asking 'why does winrm use ansible_ssh_user?'

nitzmahone:
Yeah, this will probably just end up being a whole new PR for v2 (adding support for proxy chaining, new arg naming, etc), but now that I know cchurch's trick for smuggling vars in v1, I might backport the chaining stuff to this one as a standalone connection plugin. Let's leave this one open for now, and I'll try to prioritize the v2 work and backport.

sivel:
This PR was superseded by the already merged https://github.com/ansible/ansible/pull/11908

PR #9480, opened 2014-11-04
Add support for google-authenticator time-based OTP (TOTP)

bauruine:
This PR adds support for google-authenticator style OTPs.
It requires a patched version of sshpass (https://github.com/bauruine/sshpass) to work. I'm not sure if there is a demand for this and the requirement for a patched version of an external programm is another downside.

It's not yet completed and there are still some parts missing (e.g. documentation) but i would like to get some feedback before i spend too much time on it.

Is there a chance that this feature gets merged?

### DIFF
--- a/bin/ansible
+++ b/bin/ansible
@@ -88,6 +88,9 @@ class Cli(object):
         if (options.ask_vault_pass and options.vault_password_file):
             parser.error("--ask-vault-pass and --vault-password-file are mutually exclusive")
 
+        if (options.ask_pass and options.ask_otp_code):
+            parser.error("--ask-pass and --ask-otpcode are mutually exclusive")
+
         return (options, args)
 
     # ----------------------------------------------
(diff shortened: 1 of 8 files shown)


bcoca:
Sorry for the delay on feedback, this would be a great feature to add, but I really loathe the requirement of a patched sshpass. If there is any way to bypass this requirement I'm sure this is something we would want to include.

jmMeessen:
I would also be interested in such feature. +1

amenonsen:
I think this should probably be closed now. It would have been a nice feature, but the code is outdated with respect to current devel, and there's been no further suggestion on how to bypass the requirement to use a patched sshpass. (Also note that it's pretty fragile to authenticate with the OTP and set `ControlPersist=10m` and just hope the playbook will finish before a new OTP will be needed, since there's no way to ask for another OTP while the playbook is running.)

PR #9481, opened 2014-11-04
fixes powershell upgrade script to work on different System architectures.

almartino:
On x86 systems doesn't work so, starting by the line 63 we check if the architecture is x86 or x64.


cchurch:
Merged, thanks!

PR #9484, opened 2014-11-05
Fix for --list-tasks with multiple tags and plays

gplasky:
This is a cosmetic fix for the following scenario:
- Playbook with multiple plays
- Multiple tags passed to either `--skip-tags` or `--tags` 
- `--list-tasks` passed to ansible-playbook

Previously the loop which iterated over plays in "no-op" mode (`--list-tasks`, `--list-hosts`, or `--syntax`) would return an empty set of tasks when the above conditions were met. The logic was constructed to break out of the loop when an "unknown" tag is encountered. Breaking out here is inappropriate considering this scope doesn't cover all playbook tags and only sees those inside the current play.

I've elected to duplicate the logic from the main playbook library to throw an error at the user when an unknown tag is encountered. Since `ansible-playbook` catches this under a normal playbook run, the unknown tag logic could be removed in these no-op modes, but my thinking is that it would be nice to catch in check mode vs a live run.

I've ran this change through the full battery of unit and integration tests without any issues. I've also constructed a few sample playbooks to test the fix and it works as expected (would be happy to provide output if desired).

### DIFF
--- a/bin/ansible-playbook
+++ b/bin/ansible-playbook
@@ -210,6 +210,8 @@ def main(args):
             print 'playbook: %s' % playbook
             print ''
             playnum = 0
+            matched_tags_all = set()
+            unmatched_tags_all = set()
             for (play_ds, play_basedir) in zip(pb.playbook, pb.play_basedirs):
                 playnum += 1
                 play = ansible.playbook.Play(pb, play_ds, play_basedir,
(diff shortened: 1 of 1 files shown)


gplasky:
IMHO this is really more of a bugfix since under the given scenario, `--list-tasks` does not work as designed.

mhite:
Once this merge conflict is resolved and it is verified, can we possibly consider accepting this PR? Thanks!

bcoca:
at this point it is probably better to look at v2_final branch that devel

gplasky:
When I get some time, I'll look at re-working this on `v2_final`.

amenonsen:
Since `--list-tasks` isn't done in devel yet, this should probably be closed now.

bcoca:
check against current devel, --list-tags has been implemented

amenonsen:
I can confirm that it works in current devel.

gplasky:
Cool, I'll double-check against my test case and if so, I'll close this bugger out.

gplasky:
v2 devel appears to be broken at the moment, so I'll take your word for it. :)

PR #7287, opened 2014-05-06
(edited) remove manpage output from source control

hlieberman:
##### Issue Type:

Bugfix Pull Request
##### Ansible Version:

ansible 1.6 (master 812851cb8b) last updated 2014/05/06 01:23:59 (GMT +000)
##### Environment:

N/A
##### Summary:

There is a minor manpage bug in ansible-pull.1 due to a missing macro definition.
##### Steps To Reproduce:

Running:

``` bash
LC_ALL=en_US.UTF-8 MANROFFSEQ='' MANWIDTH=80  \
  man --warnings -E UTF-8 -l -Tutf8 -Z ansible-pull.1 >/dev/null
```

will produce the following error:

```
<standard input>:28: warning: macro `Aq' not defined
```
##### Expected Results:

N/A
##### Actual Results:

N/A

PR #9490, opened 2014-11-05
removed generated man pages and added them to gitignore

bcoca:
just to avoid 'changed' status in git repo after running make docs


sivel:
Also see #7287

bcoca:
@sivel, so if I'm reading this correctly, the other request becomes necessary  once this is merged in?

sivel:
I believe #7287 becomes UNnecessary once this is merged in.  #7287 modifies a file that this PR removes.

PR #9496, opened 2014-11-06
Run keys through Jinja

dstillman:
I ran into the same problem as in #7963 — I wanted to dynamically set the names for EC2 tags, in my case to record role info. (Since an instance might have multiple roles and EC2 tag names are unique, the role has to go in the name.) But trying to use a variable in the key didn't work:

```
- name: Set role state tag
  local_action: ec2_tag resource={{ ec2_id }} region=us-east-1
  args:
    tags:
      role-{{ service }}: "{{ 'started' if result.rc == 1 else 'stopped' }}"
```

Since only values are passed through Jinja, the literal "role-{{ service }}" would be set as the EC2 tag name.

Before discovering #7963, I fixed this by passing keys through Jinja as well. I'm not sure if there are any other core modules where it'd be helpful to dynamically set keys, but this seems like it could be a more natural approach in these situations than requiring the use of two separate variables. (If dynamic variables do pose a problem — performance? overriding other keys? — then the approach in #7963 would be fine.)

### DIFF
--- a/lib/ansible/utils/template.py
+++ b/lib/ansible/utils/template.py
@@ -126,6 +126,7 @@ def template(basedir, varname, vars, lookup_fatal=True, depth=0, expand_lists=Tr
         elif isinstance(varname, dict):
             d = {}
             for (k, v) in varname.iteritems():
+                k = template(basedir, k, vars, lookup_fatal, depth, expand_lists, fail_on_undefined=fail_on_undefined)
                 d[k] = template(basedir, v, vars, lookup_fatal, depth, expand_lists, fail_on_undefined=fail_on_undefined)
             return d
         else:


sivel:
I'm not sure how I feel about this.

I think it opens up the ability to basically use variable variables which are not really a good thing.

Historically from other issues and pull requests, variable variables are functionality that we wanted to stay away from.

dstillman:
Totally fair. And it's possible this problem is rare enough that it should just be dealt with on a per-module basis.

bbradbury:
My use case (and others, too) would benefit from this feature. I want to tag multiple roles on an instance, like 

role_rabbitmq=yes role_worker=yes

where the value after role_ is a variable driven by a list. I'll then drive a dynamic inventory with those role tags, allowing me to have instances in multiple buckets.  Eli from StackOverflow is doing something similar: http://stackoverflow.com/questions/27805976/resolve-dictionary-key-or-parameter-variable-in-ansible

Particularly for instance tags, I think "variable variables" make complete sense, as that's the very nature of tags.

amenonsen:
The code needs updating here, but before that, can we have an authoritative opinion about whether this feature is desirable/acceptable? (I don't personally use enough different tags like this that it's a burden to set the names statically, but I can certainly see the attraction and potential use cases of being able to do this.) One possible easier fix would be to make the `ec2_tag` module also take a list of `tag: x, value: y`-style arguments, but that's a much narrower fix than the implementation in this PR.

amenonsen:
Since this PR has neither been updated in months, nor has anyone else commented on the feature or its implementation, I'll close this now.

PR #9505, opened 2014-11-06
Added support for AWS credentials in ec2.ini

RealSalmon:
Adds support for specifiying AWS credentials in ec2.ini that is used by the ec2.py dynamic inventory.


amenonsen:
Duplicate of #9445, same comments: I don't think this is a good idea.

PR #9522, opened 2014-11-09
Fixed 404-link for Michael DeHaan profile page.

rndmh3ro:


cchurch:
Thanks, I'm sure @mpdehaan will sleep much better tonight knowing that he's not generating a 404.

PR #9524, opened 2014-11-09
guide_rax.rst: fix add_host invocations

ktosiek:
change `groupname` to `groups`, as per add_host documentation


amenonsen:
This looks fine to merge, the rax modules in devel don't mention groupname at all.

Issue #9415, opened 2014-10-24
fetch: should use hostname instead of ip for directory name


mwarkentin:
##### Summary:

As per the [Ansible Project discussion](https://groups.google.com/forum/?nomobile=true#!msg/ansible-project/ky01x1PY_H8/70c5jGGJolgJ) with @jimi-c and @mpdehaan, `fetch` should be using `inventory_hostname` rather than `conn.host` to name the download directory.

In the meantime, you can use this as a workaround:

`- fetch: src=/etc/hosts dest="/tmp/{{inventory_hostname}}/" flat=yes`
##### Steps To Reproduce:
- Create a task which uses the `fetch` module to pull down a file from one or more remote hosts
##### Expected Results:

File system tree is built using the hostname: 
![](https://s3.amazonaws.com/snaps.michaelwarkentin.com/scan-fs.yml__waveapps_2014-10-24_11-35-04.png)
##### Actual Results:

File system tree is built using the host ip: 
![](https://s3.amazonaws.com/snaps.michaelwarkentin.com/scan-fs.yml__waveapps_2014-10-24_11-36-32.png)

bcoca:
closed via merge

PR #9525, opened 2014-11-09
fix bug 9415 - fetch uses conn.host instead of inventory_hostname

srri-zz:
Use inventory_hostname instead of conn.host to avoid IP addresses as folder names when fetching from multiple servers. See: #9415


mwarkentin:
Nice, thanks!

srri-zz:
I made a mistake with my code actually, will reopen once fixed

PR #9530, opened 2014-11-10
rax.py inventory: improvements

sivel:
This PR adds the following improvements to the `rax.py` inventory script:
1. Adds support for RackConnect v3 by allowing the user to configure which network to use for `ansible_ssh_host` as well as whether to use IPv4 or IPv6
2. Adds support for reading configurations from a config file as well as environment vars (example config file is provided)
3. Supports servers booted from volume

The ability to specify the access network and access IP version, also supports fallbacks. Specifying `RAX_ACCESS_NETWORK=public,private` and `RAX_ACCESS_IP_VERSION=6,4` will look for public IPv6, public IPv4, private IPv6, and finally private IPv4. If no matching networks are found, `ansible_ssh_host` will not be set, ultimately falling back to the `inventory_hostname`.

Additionally, the in code "documentation" has been cleaned up. 

This PR supersedes #9307 

Extra testing appreciated: @angstwad @claco

### DIFF
--- a/plugins/inventory/rax.ini
+++ b/plugins/inventory/rax.ini
@@ -1,0 +1,55 @@
+# Ansible Rackspace external inventory script settings
+#
+
+[rax]
+
+# Environment Variable: RAX_CREDS_FILE
+#
+# An optional configuration that points to a pyrax-compatible credentials
+# file.
+#
+# If not supplied, rax.py will look for a credentials file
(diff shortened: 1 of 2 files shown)


claco:
My testing isn't very extensive, but I dropped it into my inventory and all the things I usually do still work. :+1:

sivel:
@claco thanks, I appreciate your testing!

mpdehaan:
Merged, thanks!

PR #9541, opened 2014-11-12
Allow multiple interpreters to be specified

quinot:
It may be inconvenient to set a specific ansible_*_interpreter
explicitly for each host in inventory. Instead allow the specification
of a comma-separated list of interpreters. The first time an interpreter
is required for a given host, try them in turn until an executable is
found, and cache the value for future use.

### DIFF
--- a/docsite/rst/intro_inventory.rst
+++ b/docsite/rst/intro_inventory.rst
@@ -222,7 +222,9 @@ mentioned::
       than one Python or not located at "/usr/bin/python" such as \*BSD, or where /usr/bin/python
       is not a 2.X series Python.  We do not use the "/usr/bin/env" mechanism as that requires the remote user's
       path to be set right and also assumes the "python" executable is named python, where the executable might
-      be named something like "python26".
+      be named something like "python26". This can be a comma-separated list of paths that will be tried
+      in turn until a valid executable is found. For example when managing a mix of Linux and \*BSD machines,
+      you can globally set "ansible_python_interpreter=/usr/local/bin/python,/usr/bin/python".
     ansible\_\*\_interpreter
       Works for anything such as ruby or perl and works just like ansible_python_interpreter.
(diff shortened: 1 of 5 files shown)


quinot:
Hi, any target date for integration of this proposed enhancement?

bcoca:
I'm not sure this is needed as you can set interpreters by group also, even in group_vars/all as a default

quinot:
@bcoca The point is to avoid the significant maintainance burden of having to maintain an explicit mapping of machine -> interpreter.

quinot:
I have updated this pull request to the current "devel" branch. (Note that the Travis build failure is unrelated, it's the recent change to gc_storage.py).

quinot:
@bcoca other thing: you might not even _know_ beforehand the correct interpreter path for each target machine. For example, if the targets are a farm of machines you are hosting for third parties, it might well happen that each is administered with different OSs and policies, and that might even change over time, and in that case it may be impractical to initially collate, and subsequently track and maintain, the correct interperter for each target machine.

quinot:
Hello, any feedback on this proposed enhancement? I would very much appreciate some visibility on a possible target date for integration. Thanks!

bcoca:
sorry for the delay, we have a large backlog and are in the middle of replacing the main engine (v2), this will be back into consideration soon, once the new engine is in place.

quinot:
Thanks for the update Brian. I do look forward to the v2 engine!

amenonsen:
@quinot, do you have the time and inclination to update this to apply to devel? It's a useful feature that I would like to use myself. The merge conflicts aren't entirely trivial to resolve (i.e. it's not just renamed files, but changed code too: there's no `inject`, for example), but not terrible either.

quinot:
@amenonsen sure, I definitely intend to port this feature to devel.

quinot:
Superseded by PR #11810 for v2.

PR #9542, opened 2014-11-12
Docker inventory script tls connection support

gpardikis:
This PR adds TLS connection support for the docker inventory script.

I have come across the problem of the docker-inventory script not working any more with boot2docker 1.3+.
The issue is that boot2docker creates a server that wants to use a full TLS mutually authenticated connection (e.g. server cert+client cert).
The current docker inventory script does not support TLS connection settings setup.

Notable changes:
- Now you can use the script to inventorize one docker instance by having the configuration variables stored in shell environment variables.
  
  boot2docker example:
  
  (By default boot2docker will use tls as of version 1.3.0)
  
  ```
  boot2docker init
  eval "$(boot2docker shellinit)"
  export DOCKER_ASSERT_HOSTNAME="boot2docker"
  ./docker.py --list
  ```
- Now the connection settings stored in shell environment variables are imported using docker-py's docker.utils.kwargs_from_env() function. That properly sets up the kwargs (including the docker.tls.TLSConfig) needed by the Connection constructor.
- The configuration data structures -and so the YAML confguration file- were changed, so that the server connection configuration sits in its own dict.
  
  Now the script supports multiple connection methods (plain, simple unauthenticated https://, mutually authenticated tls). This means that the docker Client connection **init** now can be called with may different signatures. Cherrypicking the options can lead to problems: either leaving something out, or calling with an incompatible constellation of options. It's cleaner and more maintainable to keep this settings in a seperate dict, that can be used as kwargs for the Connection constructor.
  
  Now the variable names in the 'server' configuration stanza correspond to the arguments of the constructor of the docker.Client class, the exception being the tls object. The TLSConfig class is instantiated with the kwargs in the tls_config object.
  
  For reference, the signature of the docker.tls.TLSConfig **init**:
  
  ```
  def __init__(self, client_cert=None, ca_cert=None, verify=None,
               ssl_version=None, assert_hostname=None):
  ```
  
  The signature of the docker.Client **init**:
  
  ```
  def __init__(self, base_url=None, version=DEFAULT_DOCKER_API_VERSION,
               timeout=DEFAULT_TIMEOUT_SECONDS, tls=False)
  ```
  
  Configuration example:
  
  ```
  ---
  - defaults:
      server:
        base_url: https://192.168.59.103:2376
        tls_config:
          client_cert:
            - /home/me/.boot2docker/certs/boot2docker-vm-docker-dev/cert.pem
            - /home/me/.boot2docker/certs/boot2docker-vm-docker-dev/key.pem
          ca_cert: /home/me/.boot2docker/certs/boot2docker-vm-docker-dev/ca.pem
          verify: true
          assert_hostname: false
          ssl_version: TLSv1
        version: 1.9
        timeout: 60
  ```
- The old-style settings are imported, for the script needs to be backward-compatible. For imported settings, TLS is disabled (as the old configuration had no support for tls settings.) If you have an old-style configuration file and you want tls connections, you will have to convert your settings. If you don't want tls, you don't need to do anything.

### DIFF
--- a/plugins/inventory/docker.py
+++ b/plugins/inventory/docker.py
@@ -197,60 +197,45 @@ def setup():
                 sys.exit(1)
 
     # Environment Variables
-    env_base_url = os.environ.get('DOCKER_HOST')
-    env_version = os.environ.get('DOCKER_VERSION')
-    env_timeout = os.environ.get('DOCKER_TIMEOUT')
-    env_ssh_port = os.environ.get('DOCKER_PRIVATE_SSH_PORT', '22')
-    env_default_ip = os.environ.get('DOCKER_DEFAULT_IP', '127.0.0.1')
+    env_vars = dict()
+    env_vars['server'] = docker.utils.kwargs_from_env(assert_hostname=False)
+    env_vars['ssh_port'] = os.environ.get('DOCKER_PRIVATE_SSH_PORT', '22')
+    env_vars['default_ip'] = os.environ.get('DOCKER_DEFAULT_IP', '127.0.0.1')
     # Config file defaults
     defaults = config.get('defaults', dict())
-    def_host = defaults.get('host')
-    def_version = defaults.get('version')
-    def_timeout = defaults.get('timeout')
-    def_default_ip = defaults.get('default_ip')
-    def_ssh_port = defaults.get('private_ssh_port')
 
     hosts = list()
 
     if config:
+        env_server = env_vars.pop('server', dict())
+        default_server = defaults.pop('server', dict())
         hosts_list = config.get('hosts', list())
         # Look to the config file's defined hosts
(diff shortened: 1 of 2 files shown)


amenonsen:
This seems a pretty comprehensive series of patches, but I'm not in a position to test them. @gpardikis is this still a problem, and is yours still the best approach? Any interest in rebasing this to apply to `contrib/inventory/docker.py` in current devel? (Ignore the code under `v1/`, it's dead.)

amenonsen:
I'm closing this PR now, since it hasn't been updated, but please feel free to open a new one if you get a chance to update your patch.

Since this is a really busy project, we don't always see comments on closed tickets, but if you want to reopen discussion on this subject, please feel free to do so on the development mailing list:

```
```

Issue #9544, opened 2014-11-12
ssh + sudo password can hang indefinitely


stanhu:
##### Issue Type:

Bug Report
##### Ansible Version:

$ ansible --version
ansible 1.7.2
##### Environment:

MacOS (also occurs on Ubuntu 14.04)
##### Summary:

ssh logins that require sudo passwords can hang indefinitely.
##### Steps To Reproduce:

I think the bug could be be reproduced if you enable sudo passwords and change sudo to a program that runs 'sleep 10' before it executes. Then run a playbook with the "--ask-sudo" parameter with that machine in the inventory.

Here is what I think is happening:
1. Ansible attempts to ssh into a machine.
2. Ansible successfully logs in and attempts to run 'sudo'.
3. sudo takes a while and requires a password to be entered.
4. Ansible does a select() on the read socket.
5. The select() times out after 10 seconds. Ansible waits for the SSH process to terminate (from ansible/runner/connection_plugins/ssh.py):

```
         if not rfd:
                    # timeout. wrap up process communication                                                                                                                                 
                    stdout = p.communicate()
                    raise errors.AnsibleError('ssh connection error waiting for sudo or su password prompt')
```

However, since Ansible is logged in and sudo is running, the process never terminates. communicate() gets stuck.

Essentially, I think this is a corner case where the ssh login succeeds, but the time it takes for sudo to execute and ask for a password exceeds the timeout. Since the ssh session was successful, the ssh process will never return, causing Ansible to hang indefinitely.

I verified this was happening by adding a 'continue', here:

```
   if not rfd:
                    continue
                    # timeout. wrap up process communication                                                                                                                                 
                    stdout = p.communicate()
```

This worked--essentially a short time later, the select() returned successfully, and Ansible functioned well. To confirm this, I added some debug statements using time.time(). Here is what I saw:
1. t = 0: start of first select()
2. t = 10.000742: timeout occurs
3. t = 10.000874: second select() returns successfully

How to fix this? Ideally, we would have some way of knowing that the ssh session successfully logged in but was waiting for sudo. I don't think there is a foolproof way of knowing this. 

In any case, communicate() should have a timeout, so that if ssh fails to terminate, we either terminate the ssh process or run select() again just to be sure. There are some ideas here for how to implement a timeout:

http://stackoverflow.com/questions/1191374/subprocess-with-timeout 

For now, I've worked around the problem by changing timeout = 20, but I think the code still needs to prevent communicate() from getting stuck.
[...]

PR #9545, opened 2014-11-12
Fix ssh sudo hang

stanhu:
Use Python 3.2 backport of subprocess to prevent ssh processes from hanging indefinitely. See #9544.


bcoca:
first, we don't take white space only commits as they break attribution, we tend to clean the white space when needing to modify the lines that have it.

The fix is interesting but I'm not sure if we want to introduce a new dependency for core ansible, specially one that seems only available through pip, which makes it another dependency for non pip installation methods.

Closing this for now, please start a discussion on the ansible-devel list about the python subprocess32 module as a solution to the timeout issues, I think we need to weigh in the concerns above and  maybe others before using it as a solution.

PR #9560, opened 2014-11-14
Add ipaddr() filter plugin

drybjed:
With `ipaddr()` filter plugin you can test and manipulate IP addresses and network ranges inside Jinja2 templates. `python-netaddr` library is used as the backend and is required on Ansible Controller, but not on remote hosts.

More information and usage examples can be found here: https://github.com/drybjed/ansible-ipaddr-filter

### DIFF
--- a/lib/ansible/runner/filter_plugins/ipaddr.py
+++ b/lib/ansible/runner/filter_plugins/ipaddr.py
@@ -1,0 +1,531 @@
+# (c) 2014, Maciej Delmanowski <drybjed@gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+from ansible import errors
+
+try:
+    import netaddr
+except Exception, e:
+    raise errors.AnsibleFilterError('python-netaddr package is not installed')
+
+
+# ---- IP address and network filters ----
+
+def ipaddr(value, query = '', version = False, alias = 'ipaddr'):
+    ''' Check if string is an IP address or network and filter it '''
+
+    query_types = [ 'type', 'bool', 'int', 'version', 'size', 'address', 'ip', 'host', \
+                    'network', 'subnet', 'prefix', 'broadcast', 'netmask', 'hostmask', \
+                    'unicast', 'multicast', 'private', 'public', 'loopback', 'lo', \
+                    'revdns', 'wrap', 'ipv6', 'v6', 'ipv4', 'v4', 'cidr', 'net', \
+                    'hostnet', 'router', 'gateway', 'gw', 'host/prefix', 'address/prefix' ]
+
+    if not value:
+        return False
+
+    elif value == True:
+        return False
+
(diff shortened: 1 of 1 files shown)


bmoar:
Beats parsing strings w/ python. public/private query is neat. :+1:

kiranos:
Worked great for me, I have it in filter_plugins/ until merged

resmo:
+1 use it, works for me

clickfreak:
+1, thanx @drybjed !

surpr1ze:
+1

redrampage:
+1

magicrobotmonkey:
+++1 using ansible to configure switches, this is essential!

triplem:
+1

abadger:
Basically looks good :-)  Things to modify:
- Instead of failing on load if the ipaddr library is unavailable, set a flag and fail with the message to install python-netaddr when the filter is used. 

Refactorings (could be done after merge):
- Exception catching would be better if it was targeted at specific exceptions rather than catching almost anything
- Experience with large if-elif-else chains in service and facts make me leary of their long-term maintainability.  Perhaps we should use the dict-dispatching idiom instead:

``` python
def bool_query(v):
    if v:
        return True
[...]
query_actions = {'bool': bool_query, [...]}
[...]
try:
    return query_actions[query]()
except KeyError:
    [...]
    raise errors.AnsibleFilterError(alias + ': unknown filter type: %s' % query)
```

abadger:
@drybjed Okay, here's most of the cleanups to the code: https://github.com/ansible/ansible/compare/pr-9560?expand=1

I smoketested with a few examples from your ipaddr repo but I did not try out all of the functionality.  If you have some playbooks to run further tests to make sure I didn't break things, that would be great.

Two things that aren't in here:
- Specifying exceptions.  I was afraid of changing the behaviour of the code.  I think that most of your bare exceptions are meant to act like "If any error occurred: return False" so that seems like it would be a valid use of an unspecified exception.  If there are exceptions which aren't meant to be used like that, though, they should be specified instead.
- The meaning of this piece of code changed but I'm not sure if it was a bug before:

``` python
        ### ?? 6to4 and link-local were True here before.  Should they still?
        if query and (query not in query_func_map or query == 'cidr_lookup') and ipaddr(query, 'network'):
            iplist = netaddr.IPSet([netaddr.IPNetwork(query)])
            query = 'cidr_lookup'
```

6to4 and link-local weren't in the query_types list.  This conditional was checking for things to be in there.  The logic looks like those queries should have been in query_types before (otherwise they would have become a cidr_lookup when they encountered this conditional) so I think the new code fixes a bug.  What do you  think?

drybjed:
@abadger Thanks for the rewrite! As for `6to4` and `link-local` not being in `query_types`, you're right - looks like I've forgot to add them to that list when I added the code later, I wonder why they still work...

`ipaddr()` is designed to filter both strings and lists, like:

```
'{{ (ansible_all_ipv4_addresses + ansible_all_ipv6_addreses) | ipaddr("public") }}'
```

Most of the filters either return a correct value or `False`, otherwise list functionality would choke on first incorrect value - that way you can just filter out incorrect ones and discard (or just pass in case of `wrap`) the rest.

I'm going to check your version with my playbook and report back shortly if any errors show up.

drybjed:
@abadger: I filtered my example README script through your version and `git` didn't show any changes, so I think we're good to go. :+1:

abadger:
Yay!  Excellent.  Looks like we're good for code then.  We also want to document the new filters.  Do you want to take a stab at consolidating the relevant portions of your README into playbooks_variables.rst ?  We could do it as a separate PR if you want; we just have to make sure it gets documented.

drybjed:
@abadger sure, I'll see what I can do. I wrote that README with inclustion into `playbooks_variables.rst` in mind (it's even in reStructuredText, what a coincidence!), so I think most of it could go there. Will do a separate PR.

abadger:
Thanks @drybjed !  I've merged the modified version.  Feel free to ping me on IRC when you have the documentation PR ready.

# Merged

Hi!

issue is resolved.

Thank you!

PR #9565, opened 2014-11-16
Fix file mode calculation

pilou-:
Since 9da0627, file mode calculation is incorrect. For example when `umask` is `0077`, file mode is `0611` instead of `0600`.

See 898a38b / #7196.


pilou-:
It seems to me that the label should be `bugfix_pull_request` instead of `feature_pull_request`.

abadger:
Looks like this was fixed via: https://github.com/ansible/ansible/pull/10258

You're right, this should have been categorized a bugfix_pull_request; probably why we didn't see this earlier :-(  Apologies for the chain of mistakes.

PR #9567, opened 2014-11-16
WIP added deploy_helper module tests

ramondelafuente:
I did not know if this was required, but it seemed prudent to add tests for the deplot_helper module.
The pull request for the module is on the ansible-modules-extra's repository: https://github.com/ansible/ansible-modules-extras/pull/110

I consider this a WIP. The tests all run down here, and they cover all the module features/parameters, but I'd still like to have a second opinion.

### DIFF
--- a/test/integration/non_destructive.yml
+++ b/test/integration/non_destructive.yml
@@ -43,4 +43,5 @@
     - { role: test_embedded_module, tags: test_embedded_module }
     - { role: test_add_host, tags: test_add_host }
     - { role: test_binary, tags: test_binary }
+    - { role: test_deploy_helper, tags: test_deploy_helper }
 
--- a/test/integration/roles/test_deploy_helper/meta/main.yml
+++ b/test/integration/roles/test_deploy_helper/meta/main.yml
@@ -1,0 +1,2 @@
+dependencies:
+  - prepare_tests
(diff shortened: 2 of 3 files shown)


amenonsen:
This commit needs to be revised to apply to current devel, but I see that the module itself is under discussion in the `ansible-modules-extra` repository, and @gregdek said «I think we're going to try to have some explicit review-a-thons soon» so I'll assume this will be handled as part of that process.

ramondelafuente:
Having the tests and the module in separate repositories is a hassle. I have updated the PR for the module - if that is finally OK'ed then I'll update these as well. The module has not changed so the tests should still be valid.

ramondelafuente:
The `deploy_helper` module was merged. So this PR might still be required, right?
(I'm unclear about the policy on tests for modules in extra's)

amenonsen:
@gregdek @robynbergeron _is_ there a policy on tests for modules in -extra?

bcoca:
@amenonsen tests are fine (we do plan to put extra modules tests in extras repo, but not there yet)

gregdek:
The policy, I would say, is that tests are "encouraged". The mechanism will be much clearer when the tests are all moved to the extras repo, but as @bcoca notes, that hasn't happened yet.

ramondelafuente:
I haven't made any attempt to keep this PR up to date since the policy around it was not clear. It's almost got a beard now ;-)

amenonsen:
@ramondelafuente any chance I could convince you to neaten up its beard :-) a bit and resubmit?

mattclay:
Tests have been rebased, updated and merged in https://github.com/ansible/ansible/pull/23245

PR #9570, opened 2014-11-17
Updated shebang from #!/bin/bash to #!/usr/bin/env bash (more flexible)

Maroloccio:


bcoca:
I get the $() vs `` and most bashism changes, but why remove all the double quotes?

Maroloccio:
Which shells exactly are meant to be supported by this script?

bcoca:
bash in general, primarily linux/OS X/BSD, but probably solaris and AIX variants also

Maroloccio:
Bash "in general"? Explain. For Bash, the removal of quotes is correct. Solaris and AIX are OSs, not shells.. What do you mean?

abadger:
I've merged the commits that I am comfortable with:
- shebang to env
- backticks to $()
- Adding more quoting

I do not like the other changes as they could potentially affect portability.  Since there was a complaint that there are other non-posix constructs in env-setup already I created an alternate PR to get rid of the one that was noted on IRC: https://github.com/ansible/ansible/pull/9975

abadger:
As discussed on IRC, removed shebang line: https://github.com/ansible/ansible/commit/245f934629250ff071d419d45ed670fb79490789

abadger:
Okay, I think there's general agreement among the committers that our ideal target for this script is posix compatibility.  If we can feasibly make changes to increase posix compatibility we'll add those.  If there's changes that decrease posix compatibility that can feasibly be done in a different manner we'll ask that they be done in that other manner.

I think that we've cherry-picked all of the changes that make things clearer and do not affect the target of being posix compatible so I'll close this issue now.

Thanks for pointing out these issues and looking for a statement on what we're trying to target.

# Partially Merged

Hi!

We've merged portions of this that we thought were inline with what we're trying to achieve with the code.  the changes will also be included in the next major release.

issue is resolved.

Thank you!

PR #9579, opened 2014-11-19
Detect bhyve hosts and guests

jpte:


bcoca:
this adds to virtualization facts

amenonsen:
Quick summary: bhyve is the (Free)BSD hypervisor. This patch adds code to detect Linux guests, and FreeBSD and OpenBSD specific subclasses of Virtual, to detect the host (based on the existence of `/dev/vmm` on FreeBSD) and OpenBSD guests (`/sbin/sysctl -n hw.product`) and FreeBSD guests (based on the output of `dmesg`, which seems rather fragile; is there no better way?)

The patch seems reasonable, but needs to be rebased to apply to current devel (which has stub `FreeBSDVirtual` and `OpenBSDVirtual` subclasses already). @jpte any interest in revising this PR? Is there any better way to detect FreeBSD guests?

amenonsen:
Since this PR hasn't been updated in months, I'll close it now, but anyone is welcome to open a new one if they have a chance to rebase and test this change.

PR #9583, opened 2014-11-20
change location of ec2.ini example to moved location

electroniceagle:
If you right-click on the `ec2.ini` file in the docs it downloads a file with "301PermMove".


jimi-c:
Merged, thanks!

PR #9585, opened 2014-11-20
Fixing wrong make command in docsite/README

Sgoettschkes:
Fixed the documentation make command for the html output from `webdocs` to
`htmldocs` in the docsite/README.md

### DIFF
--- a/docsite/README.md
+++ b/docsite/README.md
@@ -11,7 +11,7 @@ If you wish to verify output from the markup
 such as link references, you may install sphinx and build the documentation by running
 `make viewdocs` from the `ansible/docsite` directory.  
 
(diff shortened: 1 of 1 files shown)


Sgoettschkes:
Any input? I'm fine with throwing my changes away and doing something different, but I'd need input on **what** to do in order to not waste any time.

bcoca:
the original text is correct, 

```
you'll need to run `make webdocs` at the top level of the repository
```

the change proposed, describes is how it is run from inside the docsite directory, not from the top. Also it is not equivalent, the first once runs `make docs` in the directory, which does more work than `make htmldocs`.

So if you change the method, you need to change the wording also as it won't work from the top level.

PR #9588, opened 2014-11-20
Add missing behavioral inventory parameters in doc

jody-frankowski:


amenonsen:
This PR needs revision to fix the merge conflicts, but also to fix the following error: it says "ansible_su" when it means "ansible_su_user", and should also include "ansible_su" separately. It's a separate question whether we should still be documenting these in preference to ansible_become*.

jody-frankowski:
We certainly should precise ansible_{su,sudo} are deprecated but still document them since ansible can still use them.

jody-frankowski:
See updated pr #11670

PR #9589, opened 2014-11-21
test_git: don't delete ~/.ssh/known_hosts during test

rohanpm:
This test is included in the non_destructive suite, but when I ran it I
found it had deleted my known_hosts, which is certainly destructive.

Avoid that by using custom ssh configuration for those tests which need
to run with an empty known_hosts file.


rohanpm:
Sorry, seems to be broken right now, it needs an update.

rohanpm:
It looks like this approach won't work as the git module already prepares a GIT_SSH wrapper (I had no idea).  I also had a look at using the ssh_opts to override the known_hosts path but that seems to interfere with the result somehow (it causes "checkout ssh://git@github.com repo without accept_hostkey (expected fail)" to start prompting if the host key should be accepted, rather than being non-interactive only).

I'll leave it alone for now, sorry for the noise.

PR #9591, opened 2014-11-21
Use full timestamp so backups during the same minute/DST changes are unique

sarahn:
When testing playbooks on a single host, it's easily possible to make multiple changes to a file within a single minute, and it's possible for there to be confusion during daylight savings rollover.

This patch includes the time zone without resorting to dateutil.tz.

I tested by applying the patch to my installed copy of 1.7.2, as I lack all the dependencies for a full test.

Example output:
grub.conf.2014-11-20@21:22:46.393099-0800~
grub.conf.2014-11-21@06:36:32.014887+0100~

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -1268,8 +1268,16 @@ class AnsibleModule(object):
 
     def backup_local(self, fn):
         '''make a date-marked backup of the specified file, return True or False on success or failure'''
-        # backups named basename-YYYY-MM-DD@HH:MM~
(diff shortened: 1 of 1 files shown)


bcoca:
this seems to already be the case, it was updated in another PR

sarahn:
On 06/18/2015 07:47 AM, Brian Coca wrote:

> this seems to already be the case, it was updated in another PR

Not completely. I suppose nobody cares about daylight savings time, but my PR included the current time zone offset. I can add that if it would be
accepted.

bcoca:
missed that part!, yes please resubmit

PR #9592, opened 2014-11-21
Change to the new syntax

adrianlzt:


bcoca:
Actually, both are valid (even a 3rd "action: module=shell") and will continue to be valid, no real need to change this.

PR #9594, opened 2014-11-22
Jinja rand filter now supports repeatable randomness

gusnuf:
By passing in a seed, which allows for repeatable randomness
- in conjunction with a variable that's unique per-host, this allows for repeatable 'random' values per host that, most importantly, don't change for each run of the playbook
- ex:  ansible all -m debug -a "msg={{ 60|random(seed=inventory_hostname) }}"
- personally, I would like to use this to repeatably splay a set of cron jobs across a cluster of hosts

### DIFF
--- a/lib/ansible/runner/filter_plugins/core.py
+++ b/lib/ansible/runner/filter_plugins/core.py
@@ -23,12 +23,12 @@ import types
 import pipes
 import glob
 import re
+import random
 import collections
 import operator as py_operator
 from ansible import errors
 from ansible.utils import md5s, checksum_s
 from distutils.version import LooseVersion, StrictVersion
-from random import SystemRandom, shuffle
 from jinja2.filters import environmentfilter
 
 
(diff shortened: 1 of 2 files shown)


lib/ansible/runner/filter_plugins/core.py
```diff
@@ -220,25 +220,33 @@ def version_compare(value, version, operator='eq', strict=False):
...
         if start or step:
             raise errors.AnsibleFilterError('start and step can only be used with integer values')
-        return r.choice(end)
+        if seed is not None:
+            return random.choice(end)
+        return random.SystemRandom().choice(end)
     else:
         raise errors.AnsibleFilterError('random can only be used on sequences and integers')
 
 def randomize_list(mylist):
```
anguslees, line 246: It would be neat if `randomize_list` also accepted a similar `seed` arg for repeatable (pseudo-random) ordering.

test/units/TestFilters.py
```diff
@@ -183,3 +184,23 @@ def test_min(self):
     def test_max(self):
         a = ansible.runner.filter_plugins.core.max([3, 2, 5, 4])
         assert a == 5
+
```
bcoca, line 187: this is a very old version of this file, which does not even exist in current devel

amenonsen:
This seems like a potentially useful feature. @gusnuf any plans to update it to apply to current devel? This also definitely needs a description and example in `playbooks_filters.rst` for inclusion.

sivel:
@gusnuf,

Hi!

gusnuf:
The code my change was in was abandoned by the mainline project.  :-/

jon-d:
Hi, the code has not been removed, the original file is now in lib/ansible/plugins/filter/core.py.

PR #9595, opened 2014-11-22
Small fix of code's syntax in Markdown

svetlyak40wt:
In markdown it is enough to use single backquotes for inline code blocks. So I replaced double backquotes with single backquotes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,7 +59,7 @@ Some other notable changes:
 * can set ansible_sudo_exe as an inventory variable which allows specifying
   a different sudo (or equivalent) command
 * git module: Submodule handling has changed.  Previously if you used the
-  ``recursive`` parameter to handle submodules, ansible would track the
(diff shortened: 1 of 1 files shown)


amenonsen:
This change is harmless, but entirely unnecessary (single and double backtick quotes are equivalent in Markdown, but the latter allows single backticks to occur in the quoted content). This PR can be merged or closed, but there's no point keeping it open any more.

svetlyak40wt:
Ok, seems that is just lack of Python Common Mark, which I use in allmychanges.com for rendering markdown. It doesn't support multiple backticks for inline code.

bcoca:
closing as this is rst and not markdown, it is interpreted by sphinx and it requires double backticks

svetlyak40wt:
@bcoca Brian, if this is "rst" file, then probably you need to change it's extention from ".md" to ".rst"???

bcoca:
sorry, mixed up documents, it is markdown, still the change should not make any difference

PR #9596, opened 2014-11-22
Add auto accept_hostkey to ansible-pull

micaelmalta:


amenonsen:
Why is it a good idea to always accept the host key silently? That seems very much unlike the policy followed in others parts of the code. I do not think this should be merged.

bcoca:
there is already a command line switch for this in devel, so I'm going to close this as done.

PR #9601, opened 2014-11-24
Set delegate on connection plugin before calling connect()

cchurch:
The `host` argument currently passed to the connection plugin reflects the value of `ansible_ssh_host` or `delegate_to`, whereas inventory variables are looked up by hostname.  Since `conn.delegate` is currently set after the `connect()` method is called on the plugin, it is not always possible to access host variables before connecting.

With this change, a connection plugin can use the following code to read host variables prior to making a connection:

```
def connect(self):
    hostvars = self.runner.inventory.get_variables(self.delegate or self.host)
    # do stuff with hostvars and connect
    return self
```

I'm planning to use this approach to support additional host variables for authentication with a winrm connection, without having to handle these host variables as a special case within runner.

### DIFF
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -931,9 +931,11 @@ class Runner(object):
             return ReturnData(host=host, comm_ok=False, result=result)
 
         try:
-            conn = self.connector.connect(actual_host, actual_port, actual_user, actual_pass, actual_transport, actual_private_key_file)
             if self.delegate_to or host != actual_host:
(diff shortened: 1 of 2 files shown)


abadger:
This code does look more sensible to me than the previous code.  Since delegate wasn't previously available when calling connect it shouldn't cause any backwards compatibility issues.

After merging you should also look at whether this needs to be applied to v2 as well (if v2 hasn't yet ported the delegate_to feature you don't have to do anything.)

+1 to merge.

PR #9611, opened 2014-11-24
Add support for win_copy, win_file and win_template modules.

jhawkesworth:
This PR is for the supporting action plugins and integration tests needed to support the core windows modules available from https://github.com/ansible/ansible-modules-core/pull/384

Both this PR and the one referenced above would be needed to add win_copy, win_file and win_template modules.


jhawkesworth:
This PR is based on Chris Church's suggestions of 14  Oct 2014 here:

https://groups.google.com/forum/#!topic/ansible-devel/Jl5qP73CiKo

I have moved the md5 checksum-calculating logic into lib/ansible/module_utils/powershell.ps1 and out of win_stat.ps1 so its available to all modules using WANT_POWERSHELL

# Observations.

The integration tests look messy now as I have left in commented out tests (which are all just copied from non-windows versions), really so others can review.

I would be grateful for a review of the module documentation as there aren't currently many examples.

I have not been able run the integration tests against window server 2008 or 2008 r2.

In use, care has to be taken when specifying windows paths, especially if you are using a combination of vars and hardcoded paths as 
you can easily run into a problem where the generated path 
either does not parse as a single quoted string, or if you use double quoted strings, some path names can become mangled because they look like escape sequences.

Using the -v option with ansible-playbook can be very helpful here as it will show the generated paths.

One thing that makes this much easier to handle is that recent versions of windows now accept / or \ for a path separator on windows systems calls.
Anywhere where a path is being passed to powershell, a / seems to be accepted, so in paths for windows modules / will be accepted.
If paths are being passed to older executables this is unlikely to work, however.

For example in the integration tests, the win_output_dir var is specified as

'C:/temp/'

The following variations all fail because some of the tasks specify a directory to copy to without a trailing / or \

# don't use:

win_output_dir: 'C:\temp\'
win_output_dir: 'C:\temp'
win_output_dir : C:\temp
win_output_dir: C:/temp

And this also fails seeming because the \t part of the name is unescaped to tab character.

# don't use either:

win_output_dir: "C:\temp"

ansible-playbook -v output demonstrates this:

  TASK: [test_win_file | verify that we are checking a file and it is present] ***
   <192.168.99.99> REMOTE_MODULE win_file path=C: emp\foo.txt state=file
   failed: [WinTestHost] => {"failed": true}
   msg: path will not be created

Currently if you hit this, you will get a stacktrace from (the excellently Pythonically-named) splitter.py 

  fatal: [WinTestHost] => Traceback (most recent call last):
  File "/home/jon/ansible/lib/ansible/runner/**init**.py", line 582, in _executor
  File "/home/jon/ansible/lib/ansible/runner/__init__.py", line 755, in _executor_internal
  File "/home/jon/ansible/lib/ansible/runner/__init__.py", line 988, in _executor_internal_inner
  File "/home/jon/ansible/lib/ansible/runner/action_plugins/win_copy.py", line 253, in run
    module_return = self.runner._execute_module(conn, tmp_path, 'win_copy', module_args_tmp, inject=inject, complex_args=complex_args, delete_remote_tmp=delete_remote_tmp)
  File "/home/jon/ansible/lib/ansible/runner/__init__.py", line 504, in _execute_module
    argsfile = self._transfer_str(conn, tmp, 'arguments', utils.jsonify(utils.parse_kv(args)))
  File "/home/jon/ansible/lib/ansible/utils/__init__.py", line 781, in parse_kv
    vargs = split_args(args)
  File "/home/jon/ansible/lib/ansible/module_utils/splitter.py", line 185, in split_args
    raise Exception("error while splitting arguments, either an unbalanced jinja2 block or quotes")
Exception: error while splitting arguments, either an unbalanced jinja2 block or quotes

Using -v it is relatively straightforward to avoid however.  

Throwing an ansible exception under the circumstances might be sufficient, 
although handling the various ways windows paths can be specified would be best.

The initial version of this also had the following notes, which are no longer true:

"I have not tackled moving the checksum logic to sha1 hashes so these modules just use md5 checksums.  
Given the large amount of change in these PRs already I felt this was too big a change.

I had to modify the windows version of fetch.py action plugin to get the existing integration tests to pass."

jhawkesworth:
argh, just realised I have merge conflicts.  Off to work out how to resolve them.

jhawkesworth:
rebased, after a fair bit of git magic.

cchurch:
@jhawkesworth From looking at some of your changes, I noticed that the md5 to sha1 updates hadn't been applied for PowerShell, so I'd fixed those in https://github.com/ansible/ansible/pull/9688.  

My changes there conflicted with some of yours here, so you'll need to rebase this PR.

jhawkesworth:
Thanks for this, once I've rebased the change will be more focused.

lib/ansible/runner/action_plugins/win_template.py
```diff
@@ -0,0 +1,147 @@
...
+
+            if '_original_file' in inject:
+                source = utils.path_dwim_relative(inject['_original_file'], 'templates', source, self.runner.basedir)
+            else:
+                source = utils.path_dwim(self.runner.basedir, source)
+
+
+        if dest.endswith("\\"): # TODO: Check that this fixes the path for Windows hosts.
+            base = os.path.basename(source)
+            dest = os.path.join(dest, base)
```
cchurch: I might suggest use `conn.shell.has_trailing_slash` and `conn.shell.join_path` like the copy module for any path manipulations on `dest`.

jhawkesworth:
I have switched to using  conn.shell.has_trailing_slash and conn.shell.join_path in win_template.py.
I have removed the changes I made to fetch.py which were still attempting to work with md5 checksums - fetch is better now anyway following the merge of https://github.com/ansible/ansible/pull/9688
I have updated the integration tests (as some contained md5 checksums in asserts).

jhawkesworth:
Rebased again.  In the process I noticed that I hadn't actually switched the powershell modules over to using SHA1 checksums, so I have fixed that and pushed against https://ansible/ansible-modules-core#384

bcoca:
# Rebase Request

Hi!

We'll work things in priority order, so just wanted you to be aware of the queue and know we haven't forgotten about you!

which are very active:

jhawkesworth:
Rebased, as requested.

For what its worth, I had a failure running the new win_feature integration
tests against windows server 2012 R2.  I'll gather some more evidence and
see if I can track down the cause.

Jon

jhawkesworth:
Thanks for this.  How on earth you guys get through so much stuff I will never know!
All the best,
Jon

*\* please reply to my unity.demon.co.uk address, not googlemail.  Thanks! **

Issue #9606, opened 2014-11-24
Synchronize module doesn't prepend path to src when run locally


mishari:
##### Issue Type:

Bug Report
##### Ansible Version:

1.7.2
##### Environment:

Vagrant Ubuntu 14.04
##### Summary:

synchronize module doesn't prepend ansible role/xxx/files directory to src
##### Steps To Reproduce:

Have the following role:
- name: Copy solr directory to tomcat
  synchronize: src=solr dest="{{tomcat_dir}}"

Then run as normal from localhost:
ansible-playbook ansible/site-vagrant.yml -i ansible/stage --connection=local
##### Expected Results:

in the generated python file:
MODULE_COMPLEX_ARGS = '{"dest": "/home/vagrant/webapps/tomcat", "src": "/home/vagrant/ansible/roles/solr/files/solr"}'
##### Actual Results:

in the generated python file:
MODULE_COMPLEX_ARGS = '{"dest": "/home/vagrant/webapps/tomcat", "src": "solr"}'

acorbi:
+1

mishari:
I have a preliminary patch based on what I found in template.py but I believe it may still cause problems due to rsync's different remote/local syntax:

```
diff --git a/lib/ansible/runner/action_plugins/synchronize.py b/lib/ansible/runner/action_plugins/synchronize.py
old mode 100644
new mode 100755
index 8a8555a..252dcdc
--- a/lib/ansible/runner/action_plugins/synchronize.py
+++ b/lib/ansible/runner/action_plugins/synchronize.py
@@ -108,6 +108,11 @@ class ActionModule(object):
         src = template.template(self.runner.basedir, src, inject)
         dest = template.template(self.runner.basedir, dest, inject)

+        if '_original_file' in inject:
+            src = utils.path_dwim_relative(inject['_original_file'], 'files', src, self.runner.basedir)
+        else:
+            src = utils.path_dwim(self.runner.basedir, src)
+
         try:
             options['local_rsync_path'] = inject['ansible_rsync_path']
         except KeyError:
--
1.9.3 (Apple Git-50)
```

mishari:
In regarding to pull request #9614 I'm not sure how valid this is since synchronize.py is in a submodule. But the code seems to make the assumption that one side of the equation has to be a remote host, so I added an extra condition that is a catchall and prepends full path to the source.

jimi-c:
Hi, this situation is no longer present in the devel branch (which will be ansible 2.0)

[...]

PR #9614, opened 2014-11-25
Fix synchronize local

mishari:
Fixes issue #9606


amenonsen:
This PR should be closed now. In devel, _process_origin does call _get_absolute_path when the source path does not contain ':' or start with '/'. So the commits in this PR are no longer applicable. The original issue #9606 can still be reproduced on devel, but that's a separate matter, and this PR is not the required fix.

bcoca:
closing as per comment above

PR #9617, opened 2014-11-25
Allow to select a subset of tasks with --only

tyll:
Add --only to alow to easily select a subset of tasks to run. This makes development of playbooks easier/faster, because one can select only the tasks that one modified to run instead of starting a full ansible run each time.


bcoca:
since tags already provide this feature I'm going to close this ticket without merging as there is too much overlap.

tyll:
On Thu, Feb 26, 2015 at 08:24:00PM -0800, Brian Coca wrote:

> since tags already provide this feature I'm going to close this ticket without merging as there is too much overlap.

Tags do not properly provide this features, since they require to add a
unique tag first to the tasks one wants to run to be able to select
them. This is contrary to make it easy to test-run individual tasks like
it is possible with --only.

PR #9619, opened 2014-11-25
Fix regression in #6734

brian-brazil:
Group variables defined in inventory should override
entries in the group_vars directories.

Add regression test.


brian-brazil:
Wrong branch...

PR #9625, opened 2014-11-25
Prevent empty diffs being written between task messages

jonhadfield:
When running a playbook with the --check and --diff options, empty diffs are still passed to the display method, resulting in an empty line being output between each task message.
This PR simply checks a non-empty diff was returned before passing to display.


amenonsen:
There's no diff mode in devel (yet?), so this PR can be closed.

bcoca:
diff will be implemented for 2.0, this might just be built in, but keeping open as a reminder

bcoca:
closed via a8b2b5ba

Issue #9622, opened 2014-11-25
A particular jinja construct doesn't evaluate to a list anymore in 1.8-devel


srgvg:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.8 (devel 8e7447962e) last updated 2014/11/25 11:12:10 (GMT +200)
  lib/ansible/modules/core: (detached HEAD 10ebcccedb) last updated 2014/11/25 11:12:10 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD 317654dba5) last updated 2014/11/25 11:12:10 (GMT +200)
  v2/ansible/modules/core: (detached HEAD cb69744bce) last updated 2014/10/07 09:23:36 (GMT +200)
  v2/ansible/modules/extras: (detached HEAD 8a4f07eecd) last updated 2014/10/07 09:23:40 (GMT +200)
  configured module search path = /usr/share/ansible

Worked fine in 1.7.2
##### Environment:

Ubuntu 14.04 on management host
##### Summary:

I'm using a the result of a difference between two lists in a loop:

`java_all_packages|difference([java_pkg])`
##### Steps To Reproduce:

This debug output shows the values of each variable:

```
TASK: [packages/java | debug var=java_pkg] ************************************ 
Tuesday 25 November 2014  15:52:18 +0100 (0:00:00.025)       0:00:02.722 ****** 
ok: [tomcathost] => {                                                           
    "java_pkg": "oracle-java7"                                                  
}                                                                               

TASK: [packages/java | debug var=java_all_packages] *************************** 
Tuesday 25 November 2014  15:52:18 +0100 (0:00:00.115)       0:00:02.838 ****** 
ok: [tomcathost] => {                                                           
    "java_all_packages": [                                                      
    ¦   "sun-java6-bin",                                                        
    ¦   "sun-java6-jdk",                                                        
    ¦   "sun-java6-jre",                                                        
    ¦   "sun-java5",                                                            
    ¦   "sun-java6",                                                            
    ¦   "sun-java7",                                                            
    ¦   "oracle-java7",                                                         
    ¦   "oracle-java8"                                                          
    ]                                                                           
}                                                                               

```
##### Expected Results:

A list with all items except `oracle-java7`
##### Actual Results:

Ansible yields an error on the result not being a list when used in a with_items_loop:

`fatal: [tomcathost] => with_items expects a list or a set`

Please note that adding the `list` filter like this, doesn't help here:

[...]

PR #9626, opened 2014-11-25
fixes to template function

bcoca:
- fix 'listification' by partially reverting a change to how terms are passed (noted in #9622)
- added missing parameters to recursive calls to function to keep calls consistent
- replaced var named vars (reserved python) for templatevars


srgvg:
Part of this fix has a regression in 1.8.1 due to 5b3c82f4ba550eaf3fafbc95334ffb0e8b02bdf5

PR #9630, opened 2014-11-26
EC2 Inventory Enhancement - Adds ability to address dynamic servers by local DNS or /etc/hosts.

isegal:
If you run a local DNS server and/or have your EC2 hosts stored in /etc/hosts using friendlier names than the default amazon FQDN, it is convenient to be able to address them. This also allows for nice tricks such as defining groups and variables in a static inventory/hosts file and merging them with the dynamic inventory such that all the variables become available.

Example: assuming that a host web0 is defined in /etc/hosts, 
Default behavior:

```
ansible tag_Name_web0 -m ping
ec2-xx-xx-xx-xx.us-west-1.compute.amazonaws.com | success >> {
    "changed": false, 
    "ping": "pong"
}
```

New behavior when enabled via ec2.ini: use_local_dns = True

```
ansible tag_Name_web0 -m ping
web0 | success >> {
    "changed": false, 
    "ping": "pong"
}
```

As a bonus - it allows merging with groups defined in inventory/hosts.
For example:

```
[my_group]
web0

[my_group:vars]
server_type=development
```

In the above example assuming that web0 exists in /etc/hosts, one can address it via tag_Name or via web0, and the server_type variable will properly be pulled. This helps avoid accidents in certain configurations when one invokes a playbook via tag and another via a hostname and causes differences in configured variables.

### DIFF
--- a/library/database/mysql_user
+++ b/library/database/mysql_user
@@ -40,7 +40,7 @@ options:
     description:
       - the hash of the user's password to be passed directly to mysql. This is mutually exclusive with the plain text password.
 		required: false
-    default: no
+    default: null
     version_added: '1.7'
   host:
     description:
--- a/library/database/mysql_user
+++ b/library/database/mysql_user
@@ -36,6 +36,12 @@ options:
       - set the user's password
     required: false
     default: null
+  password_is_hashed:
+    description:
+      - password value is specified as a hash that is used by mysql
+		required: false
+    default: no
+    version_added: '1.7'
   host:
     description:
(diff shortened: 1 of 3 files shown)


amenonsen:
Since #7395 has been merged and provides an alternative solution to this problem (and one which I think is more scalable; keeping /etc/hosts up-to-date with my extant AWS hosts became painful very quickly), I am going to close this PR (which is now outdated, and contains some unrelated commits anyway).

Issue #9635, opened 2014-11-26
LC_MESSAGES should also be forced: Ansible may not behave the same depending on your ssh_config SendEnv values


batmat:
Hi,
Just spent some hours yesterday digging into an issue.
To sum up, we (4 people) all run our playbooks from the same control machine we connect with ssh to.

The tested playbook was working fine for everyone but me.

We finally found out it has to do with /etc/ssh/ssh_config 's (which seems to be a default setting nowadays on servers, a mistake imo, but that's another story):
    SendEnv LC_*

And then some yum messages get displayed in my native language, instead of english. Which makes Ansible unable to correctly parse out the "Nothing to do" string, as it actually becomes "Rien à faire".

I was the only one (using Linux as my box) having this settings, since the others were connecting from Windows machines... A pure waste as you migh imagine :-).

Anyway, IMO, the value of Ansible is that you expect it to kind of behave the same not depending on who's running the playbook. And in this case, Ansible fails in some cases because ssh sub-layers screws up.

One might say that the solution is just to comment out the line above in the ssh_config, but though we indeed already did it, I don't feel this is a durable/solid solution. Hence I think we should enforce this in the Ansible code setting LC_MESSAGES as is already done with setting LC_CTYPE and LANG.

ansibot:
# Can You Help Us Out?

Thanks for filing a ticket! I am the friendly GitHub Ansibot. 

It looks like you might not have filled out the issue description based on our standard issue template.  You might not have known about that, and that's ok too, we'll tell you how to do it.

We have a standard template because Ansible is a really busy project and it helps to have some standard information in each ticket, and GitHub doesn't yet provide a standard facility to do this like some other bug trackers.  We hope you understand as this is really valuable to us!. 

Solving this is simple: please copy the contents of this [template](https://raw.githubusercontent.com/ansible/ansible/devel/ISSUE_TEMPLATE.md) and **paste it into the description** of your ticket.  That's it!

# If You Had A Question To Ask Instead

If you happened to have a "how do I do this in Ansible" type of question, that's probably more of a user-list question than a bug report, and you should probably ask this question on the project mailing list instead.  

However, if you think you have a bug, the report is the way to go!  We definitely want all the bugs filed :)  Just trying to help!

# About Priority Tags

Since you're here, we'll also share some useful information at this time.

[...]

PR #9636, opened 2014-11-26
Setting LC_MESSAGES: prevent unparseable messages (fixes issue #9635)

batmat:
This locale variable defines how tools should display their messages.
This is for example gonna change the yum message from "Nothing to do" to
"Rien a faire" in my case (french).

As the yum module parses that string in err, if the message is not
enforced in english this is gonna fail.

So this commits just enriches a bit more the code that's already written
for that enforcement.

This commit fixes issue #9635.


batmat:
Any chance to get this merged/considered some day? Thanks

PR #9641, opened 2014-11-26
Fix issue with ssl.wrap_socket not using TLSv1 version by default

swimlappy:
Fix a small issue when doing a URL request to a remote that forces TLSv1 and does not support SSLv2 or v3.  Seems we are doing TLSv1 by default anyway in CustomHTTPSConnection.

### DIFF
--- a/examples/ansible.cfg
+++ b/examples/ansible.cfg
@@ -212,3 +212,12 @@ accelerate_daemon_timeout = 30
 # is "no".
 #accelerate_multi_key = yes
 
+[ssl]
+# SSL/TLS Protocol
+# Configure the default protocol strength of any SSL/TLS connections
+# made by Ansible.  Valid values are
+#  SSLv2  - 0
+#  SSLv3  - 1
+#  SSLv23 - 2
+#  TLSv1  - 3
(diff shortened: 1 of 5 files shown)


bcoca:
I'm thinking that we might want to make this configurable in ansible.cfg as SSL settings WILL change and different environments might be forced to use more insecure ones than the general public.

swimlappy:
@bcoca That sounds OK to me, it adds more flexibility for future changes.

I've also added in code to have ansible respect the "no_proxy" environment variable.  It was completely ignoring that up to this point.  I work in a heavy proxy environment, so this would pretty much be a requirement for me to get this working.

swimlappy:
@bcoca Is this what you had in mind for an ansible.cfg parameter to set the SSL protocol version? Thanks

bcoca:
yes, thanks a lot

bcoca:
this produces syntax errors, I think you left a few parens ')' out

swimlappy:
Bah my github fork is completely messed up. I'll have to redo this.  Not sure what exactly has happened (lack of git experience probably).

PR #9642, opened 2014-11-26
Permit instances with no public_ip address

foozlevazquez:
Currently, the `gce` dynamic inventory script dies when an instance has no public IP address.  It shouldn't given that an instance may be accessible through a VPN to the instance's private IP address.

This patch leaves an instance with no public IP address with a `gce_public_ip` of null, and uses the private ip for `ansible_ssh_host`.


amenonsen:
This can be closed, commit 16f66a39 fixed the problem in devel already.

bcoca:
closing as per above

PR #9644, opened 2014-11-26
Added facts for OpenIndiana

basictheprogram:
OpenIndina report i86pc so changed the regex to detect this arch
Used the Oracle Solaris as a template for OpenIndiana

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -82,7 +82,7 @@ class Facts(object):
     subclass Facts.
     """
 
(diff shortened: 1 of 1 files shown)


basictheprogram:
New to git, first pull request.

abadger:
I think that I fixed this in a different way with the following commit: https://github.com/ansible/ansible/commit/04635333d0b423ee0e2c10a957a7063a67cd080f

(Adding support for multiple Solaris-family distributions).

Let me know if that change misses anything.

PR #9651, opened 2014-11-27
Update test_strategies.rst

astorije:
Switch from old format "action: module [...]" to newer format "module: [...]"

### DIFF
--- a/docsite/rst/test_strategies.rst
+++ b/docsite/rst/test_strategies.rst
@@ -64,7 +64,7 @@ Here's an example of using the URI module to make sure a web service returns::
 
    tasks:
 
-     - action: uri url=http://www.example.com return_content=yes
+     - uri: url=http://www.example.com return_content=yes
        register: webpage
 
      - fail: msg='service is not happy'


astorije:
`P5` and 3-month delay for a single word changed in the docs? :-)

amenonsen:
What's even more surprising is that the (reasonable) change still applies without any trouble. This PR can be merged safely.

astorije:
Agreed, @amenonsen...

@bcoca, I know this is labeled `P5`, but it's the simplest change ever and kind of an obvious change, 8 month old PR...
Any chance this can be merged discretely? :-)

bcoca:
actually its not 'old' format, action is not being removed or deprecated anytime soon

astorije:
I beg to differ @bcoca. Taken from [the docs themselves](http://docs.ansible.com/ansible/playbooks_intro.html#tasks-list):

> Tasks can be declared using the legacy “action: module options” format, but it is recommended that you use the more conventional “module: options” format. This recommended format is used throughout the documentation, but you may encounter the older format in some playbooks.

So in my opinion, either my change still applies, or the paragraph above should be updated.

astorije:
Not sure this comment will ever be seen/responded now that the PR is closed. A shame, but at least I tried...

PR #9652, opened 2014-11-27
Run sftp in in batch mode.

davidfritzsche:
Normally, the return code of sftp is always 0, independent of failure or success of the interactive commands. As a consequence put or get errors in the ssh connection plugin are not detected right now if sftp is used for file transport (which is the default). By switching to batch mode the return code of sftp is nonzero if the put or get fails.


amenonsen:
This PR can be closed, devel supports running sftp in batch mode already.

bcoca:
closing as per comment above

Issue #9607, opened 2014-11-24
Syntax-check executes vars_prompt


hughsaunders:
Issue Type: Bug Report
Ansible Version:  ansible 1.7.2
Environment: N/A

Summary: 
When syntax checking a playbook with --syntax-check, the user is prompted for input when vars_prompt is processed. This prevents automated syntax checking. 

Steps To Reproduce:
1) Create a playbook using vars_prompt
2) Check syntax with ansible-playbook --syntax-check

Expected Results:

Ansible-playbook non-interactively checks the syntax of the supplied playbooks. No part of the playbook is executed, it is only checked for syntax. 

Actual Results:

Ansible-playbook prompts the user for input during the syntax checking processes whenever vars_prompt occurs.

ansibot:
# Can You Help Us Out?

Thanks for filing a ticket! I am the friendly GitHub Ansibot. 

It looks like you might not have filled out the issue description based on our standard issue template.  You might not have known about that, and that's ok too, we'll tell you how to do it.

We have a standard template because Ansible is a really busy project and it helps to have some standard information in each ticket, and GitHub doesn't yet provide a standard facility to do this like some other bug trackers.  We hope you understand as this is really valuable to us!. 

Solving this is simple: please copy the contents of this [template](https://raw.githubusercontent.com/ansible/ansible/devel/ISSUE_TEMPLATE.md) and **paste it into the description** of your ticket.  That's it!

# If You Had A Question To Ask Instead

If you happened to have a "how do I do this in Ansible" type of question, that's probably more of a user-list question than a bug report, and you should probably ask this question on the project mailing list instead.  

However, if you think you have a bug, the report is the way to go!  We definitely want all the bugs filed :)  Just trying to help!

# About Priority Tags

Since you're here, we'll also share some useful information at this time.

In general tickets will be assigned a priority between P1 (highest) and P5, and then worked in priority order. We may also have some follow up questions along the way, so keeping up with follow up comments via GitHub notifications is a good idea.

Due to large interest in Ansible, humans may not comment on your ticket immediately.  

# Mailing Lists

If you have concerns or questions, you're welcome to stop by the ansible-project or ansible-development mailing lists, as appropriate.  Here are the links:
- https://groups.google.com/forum/#!forum/ansible-project - for discussion of bugs and how-to type questions
- https://groups.google.com/forum/#!forum/ansible-devel - for discussion on how to implement a code change, or feature brainstorming among developers

Thanks again for the interest in Ansible!

[...]

PR #9657, opened 2014-11-28
FIX for #9607 - Syntax-check executes vars_prompt

rameshraithatha:
Summary: 
When syntax checking a playbook with --syntax-check, the user is prompted for input when vars_prompt is processed. This prevents automated syntax checking.

Steps To Reproduce:
1) Create a playbook using vars_prompt
2) Check syntax with ansible-playbook --syntax-check

Expected Results:

Ansible-playbook non-interactively checks the syntax of the supplied playbooks. No part of the playbook is executed, it is only checked for syntax.

Actual Results:

Ansible-playbook prompts the user for input during the syntax checking processes whenever vars_prompt occurs.


bcoca:
there seems to be a lot going on in your branch unrelated to the feature the ticket states, could you rebase into something cleaner?

amenonsen:
This PR is long out of date with the code in devel, but running `--syntax-check` on a playbook that uses `vars_prompt` in devel gives a traceback with the following:

```
File "/home/ams/extern/ansible/ansible/lib/ansible/executor/playbook_executor.py", line 109, in run
self._tqm.send_callback('v2_playbook_on_vars_prompt', vname, private, prompt, encrypt, confirm, salt_size, salt, default)
AttributeError: 'NoneType' object has no attribute 'send_callback'
```

amenonsen:
The problem I noted above has been fixed now, and this PR was never updated, so I'll close it now.

PR #9658, opened 2014-11-28
Give a nice message when trying to use undefined variable within jinja2 template

cinek810:
Instead of write:
{'msg': "TypeError: argument of type 'StrictUndefined' is not iterable", 'failed': True}
Give a nice message similar to the one defined in:
template_from_string function.


bcoca:
Instead of an if in the general exception handler can you reformat to 
from jinja2.runtime import StrictUndefined
...
  except StrictUndefined,e:
...
  except Exception, e:

amenonsen:
I think this PR can be closed now, devel handles undefined variables better.

bcoca:
closing as per comment above

PR #9665, opened 2014-11-29
Correct typo of "actively"

follower:
The word "actively" seems more appropriate than "activity" here.


bcoca:
sorry, but i misspelled 2 words on that line, previous correction makes this an 'unclean' merge

PR #9050, opened 2014-09-17
Refactor: vault_password is an inventory parameter

srgvg:
Use inventory.vault_password instead of passing it through every function.

```
modified:   ../../lib/ansible/inventory/__init__.py
modified:   ../../lib/ansible/inventory/vars_plugins/noop.py
modified:   ../../lib/ansible/playbook/__init__.py
modified:   ../../lib/ansible/playbook/play.py
modified:   ../../lib/ansible/runner/__init__.py
modified:   ../../lib/ansible/runner/action_plugins/add_host.py
modified:   ../../lib/ansible/runner/action_plugins/group_by.py
modified:   ../units/TestPlayVarsFiles.py
```

[Review on Reviewable](https://reviewable.io/reviews/ansible/ansible/9050)

PR #9676, opened 2014-11-30
Allow multiple vault password files

ktosiek:
This PR adds support for passing multiple --vault-password-file. When decrypting, all the files will be used in turn until one matches.
I tried not to break vars_plugins (which are passed the current vault password) in case there's 0 or 1 password, but they will break for >1 (they'll get a list in `vault_password`).

I tried to change v2/\* too. `make newtests` passes, but I don't really know how to make sure the v2 changes work.

### DIFF
--- a/bin/ansible
+++ b/bin/ansible
@@ -78,7 +78,7 @@ class Cli(object):
                          "and su arguments ('-su', '--su-user', and '--ask-su-pass') are "
                          "mutually exclusive")
 
-        if (options.ask_vault_pass and options.vault_password_file):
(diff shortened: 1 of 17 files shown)


amenonsen:
This PR is long out of date with the code in devel, and should probably be closed now. If the feature is desired, then PR #9050 is relevant; there will be fewer places that need to be changed once that refactoring is done.

But I don't really understand the use case for this feature.

ktosiek:
The use case was restricting access to some passwords - only some specific persons would hold keys for the more important vault files. I'm not interested in working on this anymore, so I'm closing the PR.

Issue #9549, opened 2014-11-13
Wrong disk size, 21TB instead of 3TB


agrrh:
This is a bug report (attaching a label currently unavailable for me)

Hello!

I've got an issue using ansible 1.7.2 from OS X 10.9.5 to Debian 7.7:

```
$ ansible --version
ansible 1.7.2
```

See the root size of 21+ TB after executing `$ ansible groupname -i hosts -m setup`:

```
"sdb": {
                "holders": [], 
                "host": "SATA controller: Intel Corporation 7 Series Chipset Family 6-port SATA Controller [AHCI mode] (rev 04)", 
                "model": "WDC WD30EFRX-68A", 
                "partitions": {
                    "sdb1": {
                        "sectors": "5860530176", 
                        "sectorsize": 512, 
                        "size": "2.73 TB", 
                        "start": "2048"
                    }
                }, 
                "removable": "0", 
                "rotational": "1", 
                "scheduler_mode": "cfq", 
                "sectors": "5860533168", 
                "sectorsize": "4096", 
                "size": "21.83 TB", 
                "support_discard": "0", 
                "vendor": "ATA"
            }
```

but this is actually the 3 TB disk:

```
$ df -h | grep sdb
/dev/sdb1                                               2.7T  124G  2.6T   5% /mnt/mountpoint

# hdparm -I /dev/sdb | grep 'device size'
    device size with M = 1024*1024:     2861588 MBytes
    device size with M = 1000*1000:     3000592 MBytes (3000 GB)
```

`/dev/sda` is detected correctly.

Please tell me if I should provide something more.

Jmainguy:
I can confirm I have seen this as well

justinwyer:
Can't reproduce this in either release1.7.2 branch or on devel running against a debian 7.7 vagrant box

agrrh:
Reproduced from Debian to same Debian machine with Ansible 1.7.2 from stable repos:

```
root@core:~# cat /etc/debian_version
7.7
root@core:~# ansible --version
ansible 1.7.2
root@core:~# ansible all -i hosts -m setup | grep \"size\" | grep 21
                "size": "21.83 TB",
```

Looks like this is the following formula result:

```
size=(5860533168*4096)/(1024**4) # ~= 21.8
```

where 5860533168 is `/sys/block/sdb/size` contents and 4096 is from `"sectorsize": "4096",`, not sure where Ansible gets this value.

[...]

PR #9681, opened 2014-12-01
prefer hw_sector_size over physical_block_size

agrrh:
Wrong disk size, 21TB instead of 3TB
https://github.com/ansible/ansible/issues/9549


justinwyer:
This undoes earlier work around https://github.com/ansible/ansible/issues/5761 and https://github.com/ansible/ansible/pull/7771

PR #9682, opened 2014-12-01
Make sure the port obtained from parsing a URL is an integer.

wulczer:
This playbook demonstrates the problem:

``` yaml
- hosts: localhost
  connection: local
  gather_facts: False
  tasks:
    ze- get_url: url="https://wulczer.org:443/index.html" dest=/tmp/i
```

It fails for me with:

```
TypeError: an integer is required
```


amenonsen:
This can be closed, it was fixed in devel by (unrelated) commit afc19894.

wulczer:
Sigh, thanks for the heads up.

PR #9683, opened 2014-12-01
Disk size for advanced drives over 2 TB is incorrect

justinwyer:
/sys/block/sdX/queue/physical_block_size does not correlate with /sys/block/sdX/size for advanced drives larger than 2TB, /sys/block/sdX/queue/logical_block_size correlates with both see #9549 

This undoes work done by #5761

We may want to rather keep displaying the physical_block_size for sector size but then the math of "sectors" \* "sectorsize" will no longer hold true. Perhaps we want to report "logicalsectorsize" and "physicalsectorsize". Using logical_sector_size seems like the best option for now?


jimi-c:
Merged, thanks!

PR #9688, opened 2014-12-02
Fixes for fetch module when connecting to Windows hosts

cchurch:
@abadger made changes in https://github.com/ansible/ansible/commit/f1267c0b053e5975dc08c151530c802015902242 and https://github.com/ansible/ansible/commit/bc4272d2a26e47418c7d588208482d05a34a34cd to shell plugins that weren't applied for PowerShell.

This PR adds PowerShell implementations of the `expand_user` and `checksum` methods.

I also reduced the buffer size for fetching files to 0.5MB; I was getting an IOError with some EC2 Windows hosts when using a 1MB buffer.

### DIFF
--- a/lib/ansible/runner/shell_plugins/powershell.py
+++ b/lib/ansible/runner/shell_plugins/powershell.py
@@ -84,12 +84,24 @@ class ShellModule(object):
         # FIXME: Support system temp path!
         return _encode_script('''(New-Item -Type Directory -Path $env:temp -Name "%s").FullName | Write-Host -Separator '';''' % basefile)
 
-    def md5(self, path):
+    def expand_user(self, user_home_path):
+        # PowerShell only supports "~" (not "~username").  Resolve-Path ~ does
+        # not seem to work remotely, though by default we are always starting
+        # in the user's home directory.
+        if user_home_path == '~':
+            script = 'Write-Host (Get-Location).Path'
+        elif user_home_path.startswith('~\\'):
+            script = 'Write-Host ((Get-Location).Path + "%s")' % _escape(user_home_path[1:])
+        else:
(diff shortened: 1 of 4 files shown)


lib/ansible/runner/shell_plugins/powershell.py
```diff
@@ -84,12 +84,24 @@ def mkdtemp(self, basefile, system=False, mode=None):
...
+        # PowerShell only supports "~" (not "~username").  Resolve-Path ~ does
+        # not seem to work remotely, though by default we are always starting
+        # in the user's home directory.
+        if user_home_path == '~':
+            script = 'Write-Host (Get-Location).Path'
+        elif user_home_path.startswith('~\\'):
+            script = 'Write-Host ((Get-Location).Path + "%s")' % _escape(user_home_path[1:])
+        else:
+            script = 'Write-Host "%s"' % _escape(user_home_path)
+        return _encode_script(script)
```
abadger, line 97: Note -- the sh.py version of this does not handle multiple path elements.  The calling function must have reduced it to a single directory/filename.  So when the calling function gives you good values they'd be "~" or "~username".  If the calling function were to give the sh.py module any other value it will return a command that will return the string itself.

I don't know if that matters since your comment says that powershell already differs due to not being able to handle ~username but thought I'd mention it.

cchurch, line 97: In my test below (https://github.com/ansible/ansible/pull/9688/files#diff-f9c7ca43474a10fad9eac81b060d8a95R148) I end up getting the full path because backslashes get used as the path separator instead of forward slashes. If it starts with ~\ I handle it as a special case of a path relative to the home directory, otherwise it is just returned verbatim.

PR #9689, opened 2014-12-02
Make copy fail when source is empty

willthames:
An odd combination of events can cause copy to have some strange
behaviour.

If `with_files_found` finds zero files, and copy is used under
a role with files or templates beneath (at most one file need
exist under either of those two), then the copy tries to
copy the entire data structure under `dest`.

Copy should fail when source is empty - and that should include
the empty string, not just the `None` sentinel.

Other approaches might include failing when `with_files_found`
finds zero files

### DIFF
--- a/lib/ansible/runner/action_plugins/copy.py
+++ b/lib/ansible/runner/action_plugins/copy.py
@@ -64,10 +64,10 @@ class ActionModule(object):
                 except UnicodeDecodeError:
                     pass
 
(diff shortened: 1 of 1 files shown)


willthames:
See https://github.com/willthames/copy_with_null_items for a playbook and role that demonstrate this behaviour.

willthames:
I'm going to close this one because I think it's symptomatic of a wider issue (what should be done with empty `with_` groups) and further discussion on ansible-devel is merited.

PR #9694, opened 2014-12-02
Request: Honor the no_proxy environment variable

swimlappy:
I would like to submit this feature request.  The no_proxy variable is not being honored by Ansible.  This adds support for that in module_utils/urls.py.  Thanks

### DIFF
--- a/lib/ansible/module_utils/urls.py
+++ b/lib/ansible/module_utils/urls.py
@@ -252,6 +252,7 @@ class SSLValidationHandler(urllib2.BaseHandler):
         except:
             self.module.fail_json(msg='Connection to proxy failed')
 
+<<<<<<< HEAD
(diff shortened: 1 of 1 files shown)


abadger:
Double negatives like `if not no_proxy` are a bit hard to understand.  Maybe rename the function so we know what a True value in no_proxy means.  Could rename the variable too but it's used close enough to where it's set right now that just renaming the function would probably work.

Also, the code is currently going to always satisfy the conditional because the function returns either None or an empty dict, both of which evaluate to False.  Better to explicitly return True and False from the function.

swimlappy:
@abadger Thanks for the suggestions.  Does this code look a little better?

swimlappy:
I'm having trouble with my fork.  Will have to resend this later.

PR #9697, opened 2014-12-02
Allow .gitkeep in the jsonfile fact cache dir

quixoten:
##### Issue Type:

“Feature Pull Request”
##### Ansible Version:

devel
##### Environment:

Ubuntu 12.04 managed from Arch
##### Summary:

I would like to add a `.gitkeep` file to the `cache_dir` used by the `jsonfile` fact cache so I can run ansible from a fresh checkout of my repo without any additional setup. This PR modifies the `jsonfile` fact cache so that it ignores hidden files in the `cache_dir`.

RFC @bcoca

### DIFF
--- a/lib/ansible/cache/jsonfile.py
+++ b/lib/ansible/cache/jsonfile.py
@@ -98,7 +98,7 @@ class CacheModule(BaseCacheModule):
     def keys(self):
         keys = []
         for k in os.listdir(self._cache_dir):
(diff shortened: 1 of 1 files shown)


bcoca:
I think a simpler patch would be:

```
if not (k.startswith('.')  or self.has_expired(k))
```

quixoten:
I agree! Amended!

PR #9699, opened 2014-12-03
Re-add line that allows for complete playbook templating

ahamilton55:
##### Issue Type:

Bugfix Pull Request
##### Ansible Version:

``` bash
ansible 1.9 (devel 2bf269568b) last updated 2014/12/02 15:46:49 (GMT -700)
  lib/ansible/modules/core: (detached HEAD 3a80b734e6) last updated 2014/12/02 09:45:53 (GMT -700)
  lib/ansible/modules/extras: (detached HEAD 317654dba5) last updated 2014/12/01 10:08:51 (GMT -700)
  v2/ansible/modules/core: (detached HEAD cb69744bce) last updated 2014/11/08 19:22:54 (GMT -700)
  v2/ansible/modules/extras: (detached HEAD 8a4f07eecd) last updated 2014/11/08 19:22:55 (GMT -700)
  configured module search path = <>
```
##### Environment:

Running from Mac OS X 10.10.1.
Running against Debian Squeeze.
##### Summary:

We currently have a set of playbooks that I want to run on a set of groups of hosts. So, we've added a variable to the "hosts" part of a playbook. We have the group name built by a variable in a group_vars file which also uses some info from --extra-vars and have the group added dynamically.

A line was deleted in commit 9a0f8f0 that breaks this and this commit adds this back in. The tests for `test_var_precedence` have been run without any issues showing up after this revert.
##### Steps To Reproduce:

We use AWS and tagging but you should be able to reproduce this without using AWS.

group_vars/test

``` bash
my_group: "tag_group_{{ group }}"
```

test.yml

``` yaml

---
# Test
- hosts: "{{ my_group }}"
  vars_files:
    - group_vars/test
  tasks:
    - debug: msg="Test message"
```

Run something like the following:

``` bash
$ ansible-playbook -i ec2.py -e "group=test" test.yml
```
##### Expected Results:

``` bash
PLAY [tag_group_test] ***********************************************

ok: [ip-10-0-16-40...]
ok: [ip-10-0-16-236...]

TASK: [debug msg="Test message"] **********************************************
ok: [ip-10-0-16-40...] => {
    "msg": "Test message"
}
ok: [ip-10-0-16-236...] => {
    "msg": "Test message"
}

ip-10-0-16-236... : ok=2    changed=0    unreachable=0    failed=0
ip-10-0-16-40... : ok=2    changed=0    unreachable=0    failed=0
```
##### Actual Results:

``` bash
PLAY [{{ my_group }}] ****************************************************
skipping: no hosts matched

```

### DIFF
--- a/lib/ansible/playbook/play.py
+++ b/lib/ansible/playbook/play.py
@@ -859,6 +859,7 @@ class Play(object):
                             target_filename = filename4
                     update_vars_cache(host, data, target_filename=target_filename)
                 else:
+                    self.vars = utils.combine_vars(self.vars, data)
                     self.vars_file_vars = utils.combine_vars(self.vars_file_vars, data)
                 # we did process this file
                 return True


mpdehaan:
Is this about using an inventory-scoped variable from group_vars in the hosts specification?

This doesn't feel like it _SHOULD_ work to me, though perhaps you can give another example.

I do agree however that "-e group=test" absolutely MUST work in the hosts specification.

jimi-c:
HI @ahamilton55, this patch will introduce a regression due to the fact that it stores data back into `self.vars` which in terms of variable precedence is not in the correct merge order. I have created an alternative patch which resolves the issue using your example playbook above:

<pre>
diff --git a/lib/ansible/playbook/play.py b/lib/ansible/playbook/play.py
index 882d174..6e7cc0f 100644
--- a/lib/ansible/playbook/play.py
+++ b/lib/ansible/playbook/play.py
@@ -108,10 +108,16 @@ class Play(object):
         self._update_vars_files_for_host(None)
 
         # template everything to be efficient, but do not pre-mature template
-        # tasks/handlers as they may have inventory scope overrides
+        # tasks/handlers as they may have inventory scope overrides. We also
+        # create a set of temporary variables for templating, so we don't
+        # trample on the existing vars structures
         _tasks    = ds.pop('tasks', [])
         _handlers = ds.pop('handlers', [])
-        ds = template(basedir, ds, self.vars)
+
+        temp_vars = utils.merge_hash(self.vars, self.vars_file_vars)
+        temp_vars = utils.merge_hash(temp_vars, self.playbook.extra_vars)
+
+        ds = template(basedir, ds, temp_vars)
         ds['tasks'] = _tasks
         ds['handlers'] = _handlers
 
</pre>

If you can confirm that resolves your issue using your full playbook I can get that merged in today.

Thanks!

ahamilton55:
@mpdehaan I'm mainly doing this so that I can keep things on our end simple. Being able to do things this way allows me to change the tags in one place if I need to instead of tens of playbooks. I'm also able to change multiple tags by only changing the base tag. For example:

``` yaml
old_host_group: tag_group_{{ group_tag_value }}
new_host_group: "{{ old_host_group }}_new"
```

Then you can use the --extra-vars argument to pass in the "group_tag_value"

This to me seems much easier than having to spell both of these tags out and then change them.

@jimi-c: Yeah that seems to have worked. Thanks. I was trying to figure out a different way to do this but the code in this part is rather confusing to me.

jimi-c:
# Closing This Ticket

Hi!

Excellent, I've gone ahead and merged that in now. This will also be included in the next release.

Thank you!

PR #9701, opened 2014-12-03
Adds filter allowing user to ensure that a variable ends with a path separator.

jwilner:
Allows playbooks to control path_separator presence for prompted directories without resorting to regexes and relying on os.path to do it the right way for us.

"/some/user/provided/path" -> "/some/user/provided/path/"
"/some/user/provided/path/" -> "/some/user/provided/path/"

### DIFF
--- a/docsite/rst/playbooks_variables.rst
+++ b/docsite/rst/playbooks_variables.rst
@@ -336,6 +336,10 @@ To get the real path of a link (new in version 1.8)::
 
    {{ path | readlink }}
 
+To ensure a path ends with a path separator (useful with directories provided as prompt vars)::
+
+    {{ path | ensure_separator }}
+
 To work with Base64 encoded strings::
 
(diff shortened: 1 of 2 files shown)


jhawkesworth:
Will this work with windows paths?

jwilner:
Yes, the Python OS module is implemented specifically to handle that eventuality.

amenonsen:
This seems like it could be useful (though I haven't needed it myself), but I don't think `ensure_separator` is a good name. I would have preferred something like `trailing_slash(0)` or `trailing_slash(1)`, but that isn't quite natural on Windows. Perhaps `ending_separator(0)` or `ending_separator(1)` (the argument to determine whether to add or remove the separator).

Of course, the patch needs a rebase anyway.

sivel:
@jwilner,

Hi!

amenonsen:
Closing per above comment.

PR #9698, opened 2014-12-03
Exception when variable name and delegate_to hostname match

ahamilton55:
##### Issue Type:

Bugfix Pull Request
##### Ansible Version:

``` bash
ansible 1.9 (devel 2bf269568b) last updated 2014/12/02 15:46:49 (GMT -700)
  lib/ansible/modules/core: (detached HEAD 3a80b734e6) last updated 2014/12/02 09:45:53 (GMT -700)
  lib/ansible/modules/extras: (detached HEAD 317654dba5) last updated 2014/12/01 10:08:51 (GMT -700)
  v2/ansible/modules/core: (detached HEAD cb69744bce) last updated 2014/11/08 19:22:54 (GMT -700)
  v2/ansible/modules/extras: (detached HEAD 8a4f07eecd) last updated 2014/11/08 19:22:55 (GMT -700)
  configured module search path = <>
```
##### Environment:

Running on Mac OS X 10.10.1.
Running against Debian Squeeze.
##### Summary:

I am using a variable to designate my delegate host. The variable name I was using was "bastion" and the hostname returned was also "bastion" (it's in my SSH config file under that name). When these two values are the same, looking up the variable in the hostvars comes back successful but fails when trying to lookup a value inside of the dict it believes is available.
##### Steps To Reproduce:

Have a host named something such as "bastion". Currently I have one setup in my SSH config.

Playbook to reproduce this bug:

``` bash

---
# Test
- hosts: localhost
  vars:
    - bastion: bastion
  tasks:
    - debug: msg="Test message"
      delegate_to: "{{ bastion }}"
```
##### Expected Results:

``` bash
PLAY [localhost] **************************************************************

GATHERING FACTS ***************************************************************
ok: [localhost]

TASK: [debug msg="Test message"] **********************************************
ok: [localhost -> bastion] => {
    "msg": "Test message"
}

PLAY RECAP ********************************************************************
localhost                  : ok=2    changed=0    unreachable=0    failed=0
```
##### Actual Results:

``` bash
PLAY [localhost] **************************************************************

GATHERING FACTS ***************************************************************
ok: [localhost]

TASK: [debug msg="Test message"] **********************************************
fatal: [localhost -> bastion] => Traceback (most recent call last):
  File "/Users/andrew/workspace/ansible/lib/ansible/runner/__init__.py", line 590, in _executor
    exec_rc = self._executor_internal(host, new_stdin)
[...]

PR #9702, opened 2014-12-03
better exception handling with delegated hosts

bcoca:
this should fix  #9698


ahamilton55:
@bcoca: This change works. Added a comment to the commit with an error that's in there right now for the exception name in the "except" block.

Thanks.

ahamilton55:
Will this get merged in for the next release. It would be nice to move off of my branch that fixes this.

PR #9707, opened 2014-12-03
Issue-9704 Better handling of missing python

jimktrains:
When they python interpreter is set incorrectly for the machine the file
is being checked for (e.g. for the local or the remote), the error
manifests as a readability or directory missing error which can be very
misleading.


abadger:
Thanks for the pach and the time spent debugging the issue with me!

PR #9273, opened 2014-10-06
add a proxmox inventory plugin

gauthierl:
I'd like to add a dynamic inventory plugin for Proxmox (http://www.proxmox.com).

There is a closed issue for this feature (#2784).

The code is actualy not well tested and probably too soon for any merge. For example, I like to add useful variables in the inventory.

Any tests, comments or wish are welcome (even tips to make the code clean enough to be merged) !

PR #9714, opened 2014-12-04
Added dynamic inventory plugin for ProxmoxVE.

UnderGreen:
Hosts grouped by host servers. It is possible to get information
about host (-- host parameter) and get list of grouped hosts (--list
parameter, default=True). Connection credentials are readed from
.ini file.


amenonsen:
Note that #9273 also adds a proxmox plugin.

bcoca:
added the inventory plugin from #9273 making this PR obsolete, but feel free to update the existing one and add options it does not have (like using ini config file).

Issue #9637, opened 2014-11-26
Relative file lookup inside roles no longer works


jheino:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.8
##### Environment:

Debian 7.0 "wheezy"
##### Summary:

I have a file stored as roles/mariadb/files/mariadb.gpg. Using the lookup plugin to read its contents no longer works when using a relative path inside the role. It used to work in 1.7.2 but not anymore in 1.8.0.
##### Steps To Reproduce:

Put a file in roles/mariadb/files/mariadb.gpg. Then run a playbook that utilizes the mariadb role.

The contents of roles/mariadb/tasks/main.yml:

```

---
- debug: msg="{{ lookup('file', 'mariadb.gpg') }}"
```

The contents of the playbook:

```

---
- hosts: all
  roles:
  - mariadb
```
##### Expected Results:

The contents of the mariadb.gpg file.
##### Actual Results:

```
TASK: [mariadb | debug msg=""] ************************************************
ok: [192.168.100.39] => {
    "msg": ""
}
```

mpdehaan:
We have tests for exactly this but will validate.  Thanks for the report!

jheino:
After further inspection, it seems that relative paths fail to work anywhere. The only way I can get the file lookup plugin to work is to use an absolute path:

```
jukka@atom:~/ansible-bug$ echo 'Testing, 1 2 3' > helloworld.txt
jukka@atom:~/ansible-bug$ /home/virtualenv/bin/ansible -m debug -a "msg=\"{{ lookup('file', 'helloworld.txt') }}\"" -i ansible_hosts all
192.168.100.39 | success >> {
    "msg": ""
}

jukka@atom:~/ansible-bug$ /home/virtualenv/bin/ansible -m debug -a "msg=\"{{ lookup('file', '/home/jukka/ansible-bug/helloworld.txt') }}\"" -i ansible_hosts all
192.168.100.39 | success >> {
    "msg": "Testing, 1 2 3"
}
```

davidfritzsche:
I think the problem originates in the change in `utils.listify_lookup_plugin_terms()`. It tries to expand the content (as before in Ansible 1.7.2), but it now throws (with default settings) an exception if this does not work and I guess you do not have a variable `mariadb.gpg`. The exception is then silenced in `template.lookup()`.

[...]

Issue #9369, opened 2014-10-17
Setting a boolean variable default


MatthewMi11er:
# Issue Type:

Bug Report
# Ansible Version:

ansible 1.7.2
# Environment:

 OS X testing against localhost
# Summary:

When I try to use the default filter to set a variable to a boolean value, it appears to get set correctly but if it is evaluated in a `when` clause it does not get evaluated to the expected value.
This appears to be a bug, unless i'm doing something incorrectly.
# Steps To Reproduce:

``` yaml

---
- name: True/False Test
  hosts: Local
  gather_facts: False
  vars:
    t: true
    t1: '{{ t | default(true) }}'
    t2: '{{ t | default (true, true ) }}'
    t3: '{{ n | default (true, true ) }}'
    f: false
    f1: '{{ f | default(false) }}'
    f2: '{{ f | default (false, true ) }}'
    f3: '{{ n | default ("False") }}'
  tasks:
    ####################
    #shows correct value
    - debug: var=t
    - debug: var=t1
    - debug: var=t2
    - debug: var=t3
    - debug: var=f
    - debug: var=f1
    - debug: var=f2
    - debug: var=f3
    #
    #####################

    ####################
    #works as expected
    - debug: var=t
      when: t    
    - debug: var=t1
      when: t1
    - debug: var=t2
      when: t2
    - debug: var=t3
      when: t3
    - debug: var=f
      when: not f
    #
    ####################

    ######################  
    #should not be skipped
    - debug: var=f1
      when: not f1
    - debug: var=f2
      when: not f2
    - debug: var=f3
      when: not f3
    #
    ###################### 

```
# Expected Results:

```
...

TASK: [debug var=f1] ********************************************************** 
ok: [localhost] => {
    "f1": "False"
}

TASK: [debug var=f2] ********************************************************** 
ok: [localhost] => {
    "f2": "False"
}

TASK: [debug var=f3] ********************************************************** 
ok: [localhost] => {
    "f3": "False"
}

```

Actual Results:

```
...
TASK: [debug var=f1] ********************************************************** 
skipping: [localhost]

TASK: [debug var=f2] ********************************************************** 
skipping: [localhost]

TASK: [debug var=f3] ********************************************************** 
skipping: [localhost]

```

ansibot:
# Can You Help Us Out?

Thanks for filing a ticket! I am the friendly GitHub Ansibot. 

[...]

Issue #9609, opened 2014-11-24
with_ is not evaluating jinja2 expressions withouth '{{ }}'


Ernest0x:
_Issue Type:_
Bug Report

_Ansible Version:_
ansible 1.8 (devel 116109468c) last updated 2014/11/24 21:06:08 (GMT +300)
  lib/ansible/modules/core: (detached HEAD 19b328c4df) last updated 2014/11/24 21:05:39 (GMT +300)
  lib/ansible/modules/extras: (detached HEAD e64751b0eb) last updated 2014/11/24 21:05:40 (GMT +300)
  v2/ansible/modules/core: (detached HEAD cb69744bce) last updated 2014/11/24 21:06:56 (GMT +300)
  v2/ansible/modules/extras: (detached HEAD 8a4f07eecd) last updated 2014/11/24 21:07:01 (GMT +300)
  configured module search path = /usr/share/ansible

_Environment:_
Archlinux

_Summary:_
with_subelements fails when giving it a list of dicts passed through some jinja2 filter.

_Steps To Reproduce:_

```
 # inventory
 users:
   - name: someuser
     password: somepass
     authorized_keys:
       - ssh_public_keys/key1.pub
       - ssh_public_keys/key2.pub
   - name: anotheruser
     password: anotherpass

 # playbook task
 - name: Add authorized keys for users
   authorized_key: user="{{ item.0.name }}" key="{{ lookup('file', item.1) }}"
   with_subelements:
     - users|rejectattr("authorized_keys", "undefined")|list
     - authorized_keys
   tags:
     - users
```

_Expected Results:_
I expect it to work as it works on 1.7.x.

_Actual Results:_
Gives the error: 

```
 subelements lookup expects a dictionary, got 'users|rejectattr("authorized_keys", "undefined")|list'
```

ansibot:
# Can You Help Us Out?

Thanks for filing a ticket! I am the friendly GitHub Ansibot. 

It looks like you might not have filled out the issue description based on our standard issue template.  You might not have known about that, and that's ok too, we'll tell you how to do it.

We have a standard template because Ansible is a really busy project and it helps to have some standard information in each ticket, and GitHub doesn't yet provide a standard facility to do this like some other bug trackers.  We hope you understand as this is really valuable to us!. 

Solving this is simple: please copy the contents of this [template](https://raw.githubusercontent.com/ansible/ansible/devel/ISSUE_TEMPLATE.md) and **paste it into the description** of your ticket.  That's it!

[...]

Issue #9335, opened 2014-10-14
with_items doesnt work with default filter


podnov:
**Issue Type:**

Bug Report

**Ansible Version:**

```

[deploy@localhost vagrant]$ ansible --version
ansible 1.8 (devel 617352a38e) last updated 2014/10/13 19:54:37 (GMT +000)
  lib/ansible/modules/core: (detached HEAD cb69744bce) last updated 2014/10/13 19:54:50 (GMT +000)
  lib/ansible/modules/extras: (detached HEAD 8a4f07eecd) last updated 2014/10/13 19:55:02 (GMT +000)
  v2/ansible/modules/core: (detached HEAD cb69744bce) last updated 2014/10/13 19:55:22 (GMT +000)
  v2/ansible/modules/extras: (detached HEAD 8a4f07eecd) last updated 2014/10/13 19:55:40 (GMT +000)
  configured module search path = /usr/local/ansible/library

```

**Environment:**

CentOS release 6.4 (Final)

**Summary:**
I have a lot of tasks that make use of the default filter.  with_items doesn't appear to work with defaulted variables anymore.

With 'users' defined as:

```

- hosts: all:!control
  remote_user: deploy
  sudo: yes
  vars:
    users: [
      {
        "username": "team", 
        "name": "Team User"
      }
    ]

```

the tasks:

```

- name: debug users
  debug: var=users

- name: debug users with items
  debug: var=item
  with_items: users

- name: debug defaulted users with items
  debug: var=item
  with_items: users | default([])

```

yield the output:

```

TASK: [users | debug users] ***************************************************
<192.168.33.60> ESTABLISH CONNECTION FOR USER: deploy on PORT 22 TO 192.168.33.60
ok: [192.168.33.60] => {
    "users": [
        {
           "name": "Team User",
           "username": "team"
        }
    ]
}

TASK: [users | debug users with items] ****************************************
<192.168.33.60> ESTABLISH CONNECTION FOR USER: deploy on PORT 22 TO 192.168.33.60
ok: [192.168.33.60] => (item={'username': 'custapps', 'uid': 1001, 'name': 'Custom Apps Team User', 'directories': ['/home/custapps/bin'], 'createhome': 'yes', 'authorized': ['../files/deploy.pub']}) => {
    "item": {
        "name": "Team User",
        "username": "team"
    }
}

TASK: [users | debug defaulted users with items] ******************************
fatal: [192.168.33.60] => with_items expects a list or a set

```

[...]

PR #9715, opened 2014-12-04
Listify revert errorhandling

bcoca:
this should fix most templating issues we've been having with_<lookups>


bcoca:
should fix #9637, #9609 and #9335, still does not fix #9369

PR #9716, opened 2014-12-04
Add more info to error messages

msabramo:
Instead of:

```
fatal: [vagrant] => One or more undefined variables: 'start_command' is undefined
```

we now get:

```
fatal: [vagrant] => One or more undefined variables: 'start_command' is undefined
(module_name: debug; module_args: msg="start_command = {{ start_command }}";
file: /Users/marca/ansible/roles/pythonapp/tasks/main.yml)
```

which immediately points me to the error in a large library of roles.

Cc: @sontek, @sudarkoff

### DIFF
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -594,12 +594,16 @@ class Runner(object):
             if not exec_rc.comm_ok:
                 self.callbacks.on_unreachable(host, exec_rc.result)
             return exec_rc
-        except errors.AnsibleError, ae:
-            msg = str(ae)
+        except errors.AnsibleError as ae:
+            original_file = self.get_inject_vars(host).get('_original_file')
+            msg = '%s (module_name: %s; module_args: %s; file: %s)' % (
+                ae, self.module_name, self.module_args, original_file)
             self.callbacks.on_unreachable(host, msg)
             return ReturnData(host=host, comm_ok=False, result=dict(failed=True, msg=msg))
         except Exception:
-            msg = traceback.format_exc()
+            original_file = self.get_inject_vars(host).get('_original_file')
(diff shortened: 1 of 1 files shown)


msabramo:
Ping @jimi-c

msabramo:
Ping @bcoca

cemo:
+1

msabramo:
Ping

msabramo:
Ping

msabramo:
My general experience with Ansible has been that it's great when it works, but it is very hard to debug when it doesn't work. I think more information about errors would be a welcome addition.

jimi-c:
@msabramo the new v2 code base (now live in devel) handles errors like this much better. I believe that would mostly supersede this work, however if there's anything additional in this PR that isn't covered by what the newer code does please let us know. Otherwise, we can go ahead and close this, as it is no longer applicable to the code base.

amenonsen:
I looked at this commit, and devel (which is significantly different, to the point that I couldn't even figure out where such code might be applied at all) doesn't actually do much better. Rather than saying «One or more undefined variables: 'foo' is undefined», it says «ERROR! the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'foo' is undefined».

If you ask me, that's actually _worse_ than the old output in terms of being more wishy-washy and verbose. It also doesn't display the module arguments in the way this PR does.

So yes, this PR had some good ideas and could have been helpful, but it should be closed now because there's no way it can be easily reworked to fit the present code. Pity.

bcoca:
so i'm closing this as current devel does a much better job of showing errors in this and many other cases. As explained by @jimi-c above, if you find anything that is still unclear, feel free to submit PRs for these cases.

PR #9720, opened 2014-12-04
typofixes - https://github.com/vlajos/misspell_fixer

vlajos:


bcoca:
I'm just happy that only 1/2 of the misspellings are m I ne

mpdehaan:
merged, thank you!  Docs will be updated on next push.

Issue #8513, opened 2014-08-08
TestUtils.TestUtils.test_split_args unit test fails (edit: python version issue in units)


dirtyharrycallahan:
TestUtils.TestUtils.test_split_args unit test has failed since July 24/25.

Failed

TestUtils.TestUtils.test_split_args (from nosetests)

Failing for the past 7 builds (Since Failed#433 )
Took 2 ms.
Error Message

failUnlessRaises() takes at least 3 arguments (2 given)
-------------------- >> begin captured stdout << ---------------------
SENT:  a b=c d=f
WANT:  ['a', 'b=c', 'd=f']
GOT:  [u'a', u'b=c', u'd=f']
SENT:  a b='c' d="e" f='g'
WANT:  ['a', "b='c'", 'd="e"', "f='g'"]
GOT:  [u'a', u"b='c'", u'd="e"', u"f='g'"]
SENT:  a {{ y }} z
WANT:  ['a', '{{ y }}', 'z']
GOT:  [u'a', u'{{ y }}', u'z']
SENT:  a {{ x | filter('moo', 'param') }} z {{ chicken }} "waffles"
WANT:  ['a', "{{ x | filter('moo', 'param') }}", 'z', '{{ chicken }}', '"waffles"']
GOT:  [u'a', u"{{ x | filter('moo', 'param') }}", u'z', u'{{ chicken }}', u'"waffles"']

--------------------- >> end captured stdout << ----------------------
Stacktrace

  File "/usr/lib64/python2.6/unittest.py", line 278, in run
    testMethod()
  File "/var/lib/jenkins/jobs/ansible-devel/workspace/test/units/TestUtils.py", line 700, in test_split_args
    with self.assertRaises(Exception):
failUnlessRaises() takes at least 3 arguments (2 given)
-------------------- >> begin captured stdout << ---------------------
SENT:  a b=c d=f
WANT:  ['a', 'b=c', 'd=f']
GOT:  [u'a', u'b=c', u'd=f']
SENT:  a b='c' d="e" f='g'
WANT:  ['a', "b='c'", 'd="e"', "f='g'"]
GOT:  [u'a', u"b='c'", u'd="e"', u"f='g'"]
SENT:  a {{ y }} z
WANT:  ['a', '{{ y }}', 'z']
GOT:  [u'a', u'{{ y }}', u'z']
SENT:  a {{ x | filter('moo', 'param') }} z {{ chicken }} "waffles"
[...]

PR #9722, opened 2014-12-05
Add support for tox

msabramo:
Tox allows you to run the `nosetests` test suite in multiple Python versions (py26 and py27) and helps to ensure that both work.

Speaking of which, a few tests are broken on py26, because they're using unittest assertions that don't exist in py26 (see https://gist.github.com/msabramo/5e82d78ad9687ddf2cbc)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,4 @@ venv
 Vagrantfile
 .vagrant
 ansible.egg-info/
+.tox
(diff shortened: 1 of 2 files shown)


dirtyharrycallahan:
Also see #8513 for tests that fail with python26

msabramo:
Thanks for the link, @dirtyharrycallahan -- wow, it's been failing on python2.6 for a while.

mpdehaan:
We're going to be doing python 3 support eventually in v2/ tree and will consider doing something with tox then.

I am fine with the units only running in 2.7 for now.

PR #9726, opened 2014-12-05
Add info to assert for periodic test failure

msabramo:
Added more info to an assert statement that is intermittently failing for me when I run the test suite.


mpdehaan:
I'd rather discuss why this might be failing on ansible-devel first before I see code pushed for this that I don't understand.

Please stop by.  Thanks!

msabramo:
I don't think I know enough about ssh output and what it does with curly braces to phrase a useful email. I was hoping that the additional info that this gives would help diagnose the issue, but if it's not useful to you, that's fine.

PR #9730, opened 2014-12-05
fileglob_to_dict lookup plugin for returning dictionaries from fileglobs

ghost:
I would like to contribute a module that allows targeting a directory with a fileglob and getting back a list of dictionaries that can be used as items in a loop.

I've been using it to a target a directory of user files for passing to the user module, but it's generally useful for any directory full of files that have key/attribute pairs.

### DIFF
--- a/lib/ansible/runner/lookup_plugins/fileglob_to_dict.py
+++ b/lib/ansible/runner/lookup_plugins/fileglob_to_dict.py
@@ -1,0 +1,80 @@
+# (c) 2014, Tristan Fisher <tfisher@singleplatform.com>; <code@tristanfisher.com>
+
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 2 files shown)


mpdehaan:
We're trying to control lookup plugin sprawl so I don't think we are likely to include this in core.

Further DOCUMENTATION blocks don't get read by lookup plugins.

Stop by ansible-project if you would like to discuss, thanks!!!

ghost:
I understand the motivations to make Ansible more support-ready.  Is there a place that I could commit it in non-core?  (ansible-modules-extras doesn't include plugins)

bcoca:
lookups are for querying data, not a good place to put data tranformations,
you might want to use a filter for that

On Tue, Dec 9, 2014 at 10:32 AM, tristan notifications@github.com wrote:

> I understand the motivations to make Ansible more support-ready. Is there
> a place that I could commit it in non-core? (ansible-modules-extras doesn't
> include plugins)
> 
> —
> https://github.com/ansible/ansible/pull/9730#issuecomment-66300596.

## 

Brian Coca

ghost:
Sounds good to me, thanks for the response.

PR #9735, opened 2014-12-06
Add with_allocate: allocate a list items over a list of "slots"

herbyg-axial:
Given some list of hosts for example, it would be useful to have a simple way to distribute them into a given set of networks or zones.  with_allocate provides Ansible a loop feature for doing this.  It takes the list of things to distribute into as the `slots` parameter, and allocates the given items into each slot in round-robin fashion.

Returns results in the format:

```
" {{ item.current }} {{ item.slot }}"
```

Example:

```
    - debug: msg="{{ item.current }} is in {{ item.slot }}"
      with_allocate:
         slots:
         - "zone1"
         - "zone2"
         items:
         - "my_server1"
         - "my_server2"
         - "my_server3"
         - "my_server4"
         - "my_server5"
```

### DIFF
--- a/docsite/rst/playbooks_loops.rst
+++ b/docsite/rst/playbooks_loops.rst
@@ -218,6 +218,27 @@ One of the provided strings will be selected at random.
 
 At a more basic level, they can be used to add chaos and excitement to otherwise predictable automation environments.
 
+.. _allocating_loop:
+
+Allocating Loop
+```````````````
+
+The 'allocate' feature can be used to allocate some list of things over another list of "slots" in round-robin fashion.  You could use it to distribute a list of resources into some subset.  This returns each item and its assigned slot in the following format::
+
+    - debug: msg="{{ item.current }} is in {{ item.slot }}"
+      with_allocate:
+         slots:
+         - "zone1"
+         - "zone2"
+         items:
+         - "my_server1"
+         - "my_server2"
+         - "my_server3"
+         - "my_server4"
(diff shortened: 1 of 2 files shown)


mpdehaan:
I think this might be too confusing for a lot of users, so I don't think we want to take this one.  However, you're definitely free to use it by including it in a "./filter_plugins" directory alongside your playbook or configuring in your filter plugins path in ansible.cfg.

Thank you!

herbyg-axial:
OK.  But out of curiosity, is there a purely Ansible-official way of getting this exact functionality, or is there a way that it could have been structured that you think would have been simpler or more understandable?

Ernest0x:
@herbyg-axial I think you are complicating things with your terminology about 'slots' etc. Instead, you could rename the lookup to 'with_distribute', make it take two lists as its terms and just say that the items of the second term are distributed evenly to the items of the first term, so that you can access the pairs as `{{ item.0 }}` and `{{ item.1 }}` in a similar fashion to other lookups (with_together, with_subelements, etc.).

So, your example would look like this:

```
with_distribute:
   - [ 'zone1', 'zone2' ]
   - [ 'myserver_1', 'myserver_2', 'myserver_3', 'myserver_4', 'myserver_5' ]
```

bcoca:
Lookups are intended to get information, not manipulate it, this is a
complex interface for that. Look at using filters to combine lists.

## 

Brian Coca

PR #9736, opened 2014-12-06
Docs: install on Fedora using repo 'updates-testing'

funollet:
Added site docs for Fedora users to install Ansible from repo 'updates-testing'. This provides the latest release of Ansible while still using 'yum'.


mpdehaan:
It's mostly behind because the package might need votes - I think it might be better to just encourage voting.

While I appreciate this, we also want to keep the doc install path small and limited - I think more engaged users would know about updates-testing, but I would not want them getting everything from updates-testing.

I hope this is understandable.

Thanks!

funollet:
It's understandable and I agree; that's why the 'yum' command I suggested installs only the 'ansible' package from this repo. The repo is not permanently enabled.

PR #9737, opened 2014-12-07
Docfix: running background tasks without polling

lorin:
Fix the example in the doc to explicitly set the poll interval to zero so the job doesn't poll.

To run a background task without polling, you need to set the poll interval to zero. However, Ansible's default poll setting is 15 seconds, so not specifying the poll interval will cause a background job to poll.


mpdehaan:
Merged, thanks!

PR #460, opened 2012-06-10
Search multiple paths for modules.

atomatt:
Minimal change to allow a list of paths (separated by the typical path
separator) to be searched in sequence for the named module.

PR #9739, opened 2014-12-07
Add integration test for ansible-modules-core #460

jbinto:
Please see https://github.com/ansible/ansible-modules-core/pull/460

This integration test uses the `file` module to set `state=touch` on a hard link. Currently in Ansible, that causes an error. [ansible-modules-core/pull/460](https://github.com/ansible/ansible-modules-core/pull/460) fixes that.

n.b. If there is generally a better way to coordinate PRs for changes that span across submodules of the project, please let me know.

### DIFF
--- a/test/integration/roles/test_file/tasks/main.yml
+++ b/test/integration/roles/test_file/tasks/main.yml
@@ -106,6 +106,15 @@
     that:
       - "file6_result.changed == true"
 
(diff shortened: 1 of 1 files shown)


mpdehaan:
Leaving this open until the other is merged.  Thanks!

Issue #9703, opened 2014-12-03
Syntax-check does not print file name if it fails to parse


hughsaunders:
##### Summary:

If you specify a list of playbooks to ansible-playbook --syntax-check and one fails on parsing, ansible-playbook will halt but not tell you which file caused the failure. This is annoying if you are checking a large set of playbooks. 
##### Steps To Reproduce:

```
$ tail *.yml
==> a.yml <==

---
- hosts: localhost
  tasks:
    - ping:


==> b.yml <==

---
hosts: localhost
ping:


$ ansible-playbook -i ../local_only_inventory --syntax-check *.yml

playbook: a.yml


ERROR: parse error: playbooks must be formatted as a YAML list, got <type 'dict'>
```

The above example show two playbooks, the syntax of a.yml is fine, b.yml is invalid. Syntax check does not print that the error is with b.yml. 
##### Expected Results:

```
$ tail *.yml
==> a.yml <==

---
- hosts: localhost
  tasks:
    - ping:


==> b.yml <==

---
hosts: localhost
ping:


$ ansible-playbook -i ../local_only_inventory --syntax-check *.yml

playbook: a.yml

playbook: b.yml
ERROR: parse error: playbooks must be formatted as a YAML list, got <type 'dict'>
```
##### Actual Results:

See steps to reproduce

srri-zz:
Fixed in #9741, new output is:

```
playbook: a.yml

ERROR: b.yml
parse error: playbooks must be formatted as a YAML list, got <type 'dict'>
```

resmo:
could not reproduce with 2.1.1, foo.yml is correct, bla.yml not

```
 $ ansible-playbook --syntax-check foo.yml bla.yml 
 [WARNING]: Host file not found: /etc/ansible/hosts

 [WARNING]: provided hosts list is empty, only localhost is available

ERROR! 'debug' is not a valid attribute for a Play

The error appears to have been in '/home/resmo/tmp/xyz/bla.yml': line 2, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

---
- hosts: bla
  ^ here
```

closing

PR #9741, opened 2014-12-07
fix bug #9703 parse error

srri-zz:
Outputs the name of the file with a parsing error (more details in #9703)

```
$ ansible-playbook -i ../local_only_inventory --syntax-check *.yml
```

old output:

```
playbook: a.yml

ERROR: parse error: playbooks must be formatted as a YAML list, got <type 'dict'>
```

New output:

```
playbook: a.yml

ERROR: b.yml
parse error: playbooks must be formatted as a YAML list, got <type 'dict'>
```

### DIFF
--- a/lib/ansible/playbook/__init__.py
+++ b/lib/ansible/playbook/__init__.py
@@ -1,4 +1,5 @@
 # (c) 2012-2014, Michael DeHaan <michael.dehaan@gmail.com>
+# (c) 2014, Steven Richards <sbrichards@mit.edu>
 #
 # This file is part of Ansible
 #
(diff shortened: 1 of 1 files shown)


mpdehaan:
I don't really care for how this appears on multiple lines, so much, but do appreciate the spirit of this pull request.

We're going to overhaul this (and have done so already) on the in-progress 'v2' tree, where we show paths to filenames.  As such, I think we're not going to merge this is now, as the next release won't likely ship core code changes out of this directory.

We hope this is understandable and thanks for the submission.

Thanks!

hootsuitesteven:
Understood, thanks!

PR #9748, opened 2014-12-08
Add an option to allow an inventory that does not define any host

mguillaume:
The use case for this is when hosts are first dynamically detected/provisionned on a local play and added to the inventory with add_host.

Currently, we can use the "comma hack", but:
1) it's a hack and may disappear
2) it does not allow the use of a group_vars directory at the same level as the inventory file (the comma makes is_file return false, so no basedir is set, and then no basedir is searched)

### DIFF
--- a/bin/ansible-playbook
+++ b/bin/ansible-playbook
@@ -159,7 +159,7 @@ def main(args):
 
     inventory = ansible.inventory.Inventory(options.inventory, vault_password=vault_pass)
     inventory.subset(options.subset)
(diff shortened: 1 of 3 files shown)


mpdehaan:
I'd rather see this done as something explicit like "--non-inventory-hosts" or something that makes them all implicit or something, though as we are doing the v2/ tree right now, this is likely going to be obsoleted by that work.  As such, I propose we wait on this until that tree is live.

Please stop by ansible-devel if you would like to discuss, thanks!

PR #9749, opened 2014-12-08
Improve error reporting when parsing an ini file.

sontek:
If an executable bit is removed from a hosts file that is a python script
the inventory will be parsed as ini and the output isn't very helpful
in debugging why this would be the case.

This changes the error:

```
ansible.errors.AnsibleError: Invalid ini entry: sys - need more than 1 value to unpack
```

to this:

```
ansible.errors.AnsibleError: Invalid ini entry in /path/to/scripts/hosts/nova.py:20. Line was: 'import sys'. Expected format is 'variable=value'
```

### DIFF
--- a/lib/ansible/inventory/ini.py
+++ b/lib/ansible/inventory/ini.py
@@ -34,6 +34,7 @@ class InventoryParser(object):
     """
 
     def __init__(self, filename=C.DEFAULT_HOST_LIST):
+        self.filename = filename
 
         with open(filename) as fh:
             self.lines = fh.readlines()
@@ -87,7 +88,7 @@ class InventoryParser(object):
         self.groups = dict(all=all, ungrouped=ungrouped)
         active_group_name = 'ungrouped'
 
-        for line in self.lines:
+        for line_number, line in enumerate(self.lines):
             line = utils.before_comment(line).strip()
             if line.startswith("[") and line.endswith("]"):
                 active_group_name = line.replace("[","").replace("]","")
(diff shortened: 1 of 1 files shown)


msabramo:
This is a much, much better error. :+1:

```
❯ ansible-playbook -i hosts/ playbook.yml
ERROR: Invalid ini entry: sys - need more than 1 value to unpack

❯ git checkout improve_error_reporting
M   lib/ansible/modules/core
Switched to branch 'improve_error_reporting'
Your branch is up-to-date with 'sontek/improve_error_reporting'.

❯ ansible-playbook -i hosts/ playbook.yml
ERROR: Invalid ini entry in hosts/nova.py:20. Line was: 'import sys'. Expected format is 'variable=value'
```

:+1:

mpdehaan:
I'm not positive that's the only way this error could be raised though, which is I think why we shouldn't assume it's the case.

sontek:
@mpdehaan Are you sure?  The code is this:

```
try:
    (k,v) = t.split("=", 1)
except ValueError, e:
```

You think there will be a time a `ValueError` is raised when calling `split('=', 1)` on a string that might not be because of the string not being `key=value`?

msabramo:
The `try/except` is round a single line that does a `split`. It looks pretty clear that if `ValueError` is raised then something was wrong with that line.

Printing the file and line number is way better than `need more than 1 value to unpack`.

sontek:
@mpdehaan Could I get more details on what other things might cause this `ValueError` that would prevent this PR from being accepted?  I'm willing to improve the code if you can provide more details but with my knowledge of Python and the string split command I can't think of any other reasons a `ValueError` may be raised.

msabramo:
I can only think of one other way to get a `ValueError` from `split` -- using an empty separator:

``` python
>>> (k,v) = t.split("", 1)
ValueError: empty separator
```

but that doesn't apply here of course, because the separator is hard-coded to `"="`.

The original `ValueError` comes from https://github.com/python/cpython/blob/master/Python/ceval.c#L3806. The "empty separator" case comes from https://github.com/python/cpython/blob/master/Objects/stringlib/split.h#L154.

I don't see any other way that a `ValueError` could be raised.

sontek:
@mpdehaan Ping, just want to follow up on this. Any more details on why this close would greatly help me modify the code to meet your requirements.

cc @jimi-c @abadger

msabramo:
:+1:

This seems like a great change. Not sure if it still applies in 2.0, but the error message is much more helpful.

msabramo:
Please reopen

jimi-c:
@msabramo, @amenonsen has been working on this quite a bit, so it probably doesn't apply anymore. You might want to check out the latest devel and see if it's still an issue, and if so ping him about it (his nick is `crab` in #ansibe-devel on Freenode).

PR #9767, opened 2014-12-10
added a vars_plugin

whiskybar:
I have written and used a vars_plugin which other people might find useful, both as an example as well as a variable plugin for the real world use.

I'm adding a new example plugin which uses python files to add extra variables to the current set of variables. The directory `python_vars` is scanned for python files; only the ones named as the current group are taken into account. Each such file is scanned for the function `run(host)` and its result is merged into the current variable set.

For example, let's assume the current host is in the group `dns`. Then, `python_vars/dns.py` can contain

```
def run(host):
    return {'zones': ['example.com']}
```

and the variable `zones` will be defined as the variable for the current host.

### DIFF
--- a/lib/ansible/inventory/vars_plugins/python_vars.py
+++ b/lib/ansible/inventory/vars_plugins/python_vars.py
@@ -1,0 +1,28 @@
+import sys
+import os.path
+import ansible.utils
+
+
+class VarsModule(object):
+    def __init__(self, inventory):
+        self.inventory = inventory
+
+    def run(self, host, vault_password=None):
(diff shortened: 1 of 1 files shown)


jwerak:
+1

amenonsen:
I agree that this is useful as an example, though I personally have no need to actually use it. @whiskybar any interest in rebasing to apply to current devel? It needs to live in `lib/ansible/plugins/vars` now. It shouldn't be too hard to make it work with v2, but if it's meant to be an example, then it really needs some more explanatory comments.

amenonsen:
Since this PR was never updated to apply to devel, I'll close it now.

PR #9777, opened 2014-12-11
Augment sys.path to find modules relative to bin/ansible

vaporos:
This avoids the need to source hacking/env-setup when using bin/ansible
from a local git repository clone of the ansible repository.

A similar but slightly different augmentation already exists in bin/ansible-playbook.

### DIFF
--- a/bin/ansible
+++ b/bin/ansible
@@ -22,6 +22,11 @@
 import os
 import sys
 
+# Augment sys.path (PYTHONPATH) to find Python modules relative to the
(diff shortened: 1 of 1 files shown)


bcoca:
we are going to pass on this feature as it can lead to accidental use of modules, the -M option already allows for the inclusion of extra paths in the command line.

ansible-playbook does something a bit different as it is not related to it's path or working directory, but with the play/role definitions.

PR #9780, opened 2014-12-11
Fix the comment about arguments, as this was changed a long time ago

mscherer:


amenonsen:
This PR can be closed (bin/ansible-pull no longer exists, wording doesn't occur elsewhere).

bcoca:
this info should be in the usage function and the man page, closing this issue as the comments are in a script that does not exist anymore.

Issue #7959, opened 2014-06-27
`vars_prompt` default value when shown in console doesn't evaluate jinja2 tags 


MaikuMori:
##### Summary:

When `default` value in `vars_prompt` contains Jinja2 expression it is printed out as is. When the default value is later used it works as expected.
##### Steps To Reproduce:

A `vars_prompt` section which shows the problem:

```
vars_prompt:
  - name: some_user
    prompt: 'Enter username'
    private: no
    default: '{{ lookup("env", "USER") }}'
```
##### Expected Results:

```
Enter username  [MyUsername]:
```
##### Actual Results:

```
Enter username  [{{lookup("env", "USER")}}]:
```

jimi-c:
Hi, this situation is no longer present in the devel branch (which will be ansible 2.0). Here is sample output using your example above:

<pre>
$ ansible-playbook -vv 7959.yml 
1 plays in 7959.yml
Enter username [{{ lookup("env", "USER") }}]: 
PLAY: ***************************************************************************
TASK [debug var=some_user] ******************************************************
ok: [localhost] => {
    "changed": false, 
    "some_user": "root"
}
PLAY RECAP **********************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=0   
$ ansible-playbook -vv 7959.yml 
1 plays in 7959.yml
Enter username [{{ lookup("env", "USER") }}]: foo
PLAY: ***************************************************************************
TASK [debug var=some_user] ******************************************************
ok: [localhost] => {
    "changed": false, 
    "some_user": "foo"
}
PLAY RECAP **********************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=0   
</pre>

If you continue seeing any problems related to this issue, or if you have any further questions, please let us know by stopping by one of the two mailing lists, as appropriate:
- https://groups.google.com/forum/#!forum/ansible-project - for user questions, tips, and tricks
- https://groups.google.com/forum/#!forum/ansible-devel - for strategy, future planning, and questions about writing code

Because this project is very active, we're unlikely to see comments made on closed tickets, but the mailing list is a great way to ask questions, or post if you don't think this particular issue is resolved.

Thank you!

PR #9789, opened 2014-12-12
Add support for templating in vars_prompt

nathancahill:


sudarkoff:
:+1: Thanks, @nathancahill, I was just about to do the same.

I'd add a few tests for this.

jeromedoyle:
Any update on this or alternate method to accomplish the same effect?

bcoca:
this probably needs to be redone as the whole system has drastically changed.

amenonsen:
As @bcoca says, the code in devel is very different now, and achieving the object of this PR (which is slightly unclear; some examples/tests would be welcome) would need a rewrite. I don't think there's any point keeping this PR open.

bcoca:
closing this as current 2.0 should fix  #7959 (as mentioned in that ticket).

PR #9793, opened 2014-12-12
Variables lookup in a template should handle properly the undefined case

nremond:
Hi,

If I create a role that is only doing this action :

``` ansible
  - authorized_key: user=deploy key="{{ lookup('file', item) }}"
    with_fileglob:
    - ../keys/*
```

I end up with the following error: `ERROR: Unexpected error in during lookup: 'item' is undefined`.

With the provided patch, Ansible is successfully executing the role.
I'm not familiar with the code base, let me know if there is a better solution to this issue.

Thanks


abadger:
Discussed with @bcoca and decided that this would indeed be fine.

Thanks!

# Merged

Hi!

issue is resolved.

Thank you!

PR #9801, opened 2014-12-13
Fix the note to reference the good file

mscherer:
intro_patterns do not have anything on the per-host
and per-group variables, but this is in the best practices
AFAIK.


abadger:
I think it's actually here: http://docs.ansible.com/intro_inventory.html#splitting-out-host-and-group-specific-data

abadger:
Tested and committed a fix that links to the inventory section mentioned above.  Thanks for pointing out this was linking to the wrong section!

https://github.com/ansible/ansible/commit/b74a02301bd65264c681c267fa8396585541d327

PR #9802, opened 2014-12-13
Document the behavior when a directory is used

mscherer:
For group_vars and host_vars, you can use a directory instead of a flat
file. This permit to have for example 1 private file coming from 1 git repo
containing per host variables, and 1 file coming from a public repository
( or less restricted internal repository ).


abadger:
Since this already exists here: http://docs.ansible.com/intro_inventory.html#splitting-out-host-and-group-specific-data we probably don't want to repeat ourselves.  At most maybe a short explanation and link to the inventory section?

bcoca:
closing this as the feature is already documented elsewhere, as mentioned above

PR #9803, opened 2014-12-13
Use command when we intended to

fgsch:
While here sort register variables and add a comment to signal multiline testing.


fgsch:
I will submit this in a separate branch.

PR #9804, opened 2014-12-13
Add a etcd backend for caching facts

mscherer:
This backend is modeled on the redis one, but using the builtin
expiration system of etcd. After looking at various options for
the etcd module, I settled on etcd because it was the easiest to
install with pip, but it should be trivial to change the client.

### DIFF
--- a/lib/ansible/cache/etcd.py
+++ b/lib/ansible/cache/etcd.py
@@ -1,0 +1,82 @@
+# (c) 2014, Michael Scherer
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+
+from __future__ import absolute_import
+import json
+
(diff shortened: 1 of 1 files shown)


lib/ansible/cache/etcd.py
```diff
@@ -0,0 +1,82 @@
...
+import json
+
+from ansible import constants as C
+from ansible.cache.base import BaseCacheModule
+
+try:
+    from etcd.client import Client
+except ImportError:
+    import sys
+    print "The 'etcd' python module is required, 'pip install etcd'"
```
apollo13, line 29: Should print to stderr

lib/ansible/cache/etcd.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def delete(self, key):
+        self._cache.node.delete(self._make_key(key), value)
+
+    def flush(self):
+        for key in self.keys():
+            self.delete(key)
+
+    def copy(self):
+        ret = dict()
```
apollo13, line 79: return dict((key, self.get(key)) for key in self.keys())

lib/ansible/cache/etcd.py
mscherer, line 29: Ideally, yes, but I tried to keep consistency with others modules ( like the redis one ). I will look if there is a consesnus accross the ansible source code.

bcoca, line 29: actually, it should raise an AnsiblError exception (I have fixing the other cache plugins to do same on my list)

bcoca:
these files have been moved, they now live under lib/ansible/plugins/cache

amenonsen:
@mscherer any plans to update this?

mscherer:
Not much for today. I should recheck which etcd python api did survive since when I first wrote that 1 year ago and redo accordingly, but I do not have much need for it, so not a priority for me.

amenonsen:
OK, thanks for the update. Then I think it's best to close this PR now (especially since no other potential users have commented), we can always revisit this if you need it again.

PR #302, opened 2012-05-02
Set default selinux context with '_default'

sfromm:
This removes the 'context' option and replaces it with checks for
'_default' value for seuser, serole, setype, or (maybe) selevel.
If '_default' is provided _and_ there is a default context for the given
file, this will set the file context to the available default.

Related to pull request #66 and #88 on ansible.github.com repo.  Tested on debian, centos5, and centos6, and with selinux disabled and enforcing on the latter two distributions.

PR #9806, opened 2014-12-13
Fixes #302 - Copy fails when source file is identical to the destination...

sysadmin75:
... and destination is a hard link.

### DIFF
--- a/lib/ansible/runner/action_plugins/copy.py
+++ b/lib/ansible/runner/action_plugins/copy.py
@@ -270,7 +270,10 @@ class ActionModule(object):
                     # self.runner._remove_tmp_path(conn, tmp_path)
                     continue
 
-                tmp_src = tmp_path + source_rel
+                if tmp_path:
+                    tmp_src = tmp_path + source_rel
+                else:
+                    tmp_src = source_full
 
                 # Build temporary module_args.
                 new_module_args = dict(


bcoca:
sorry, can you clarify the error case? an example would be helpful

sysadmin75:
@bcoca I'm not sure I understand the question. Could you explain more please?

bcoca:
how can i reproduce the error exactly, source and destination being identical can mean several things, checksums are same? file is the same inode? path is the same but on diff machines?

sysadmin75:
The original bug is filed under https://github.com/ansible/ansible-modules-core/issues/302

I followed the origina posters directions.
On the destination I hardlinked these two files:
# ll /etc/sysconfig/networking/profiles/default/hosts /etc/hosts
-rw-r--r-- 2 root root 158 Dec 14 10:29 /etc/hosts
-rw-r--r-- 2 root root 158 Dec 14 10:29 /etc/sysconfig/networking/profiles/default/hosts

This is the playbook I used.
- hosts: apache
  gather_facts: no
  tasks:
  - name: Setting perms on /etc/hosts (hardlinked)
    file: path=/etc/ansible/files/hosts owner=root group=root mode=0666

This is the error when I run the playbook.
TASK: [Setting perms on /etc/hosts (hardlinked)] ****************************** 
failed: [apache] => {"failed": true}
msg: src and dest are required for creating links

bcoca:
so the error is with the file module, not with copy as stated in the title?

sysadmin75:
I'm glad you noticed that. I provided the wrong task.

On my remote server, I have two files hardlinked.
[root@apache-server ~]# ll /etc/sysconfig/networking/profiles/default/hosts /etc/hosts
-rw-r--r-- 2 root root 158 Dec 14 10:29 /etc/hosts
-rw-r--r-- 2 root root 158 Dec 14 10:29 /etc/sysconfig/networking/profiles/default/hosts

I run this playbook on my manage machine.
  tasks:
    - copy: src=/etc/hosts dest=/etc/hosts

[root@centos6-server ansible]# ll /etc/hosts
-rw-r--r-- 1 root root 264 Jan 25 12:31 /etc/hosts

The results are the file is copied to destination but it unlinks the hardlink.

Remote server:
[root@apache-server ~]# ll /etc/sysconfig/networking/profiles/default/hosts /etc/hosts
-rw-r--r-- 1 root root 264 Feb 16 19:47 /etc/hosts
-rw-r--r-- 1 root root 158 Dec 14 10:29 /etc/sysconfig/networking/profiles/default/hosts

bcoca:
that is expected behavior as it replaces the file, which breaks the hard link a normal copy or scp would result in the same.

sysadmin75:
I just tested a normal cp command and scp and both do not break the hard link.

bcoca:
it does when different partitions are involved

sysadmin75:
I didn't think hard links worked across partitions anyway.

abadger:
@sysadmin75, With some research, I found that you are correct about standard behaviour of cp.  cp reuses the inode of the file it's overwriting so  hard links are preserved.

However, I think that atomicity is a better design for ansible than preserving the file's inode (From documentation ( http://www.delorie.com/gnu/docs/cfengine/cfengine-Reference_85.html ) it looks like cfengine may be doing something similar).  Maybe preserving inode vs atomicity would be valid as a parameter to the copy module but I haven't thought it through.

OTOH, from reading https://github.com/ansible/ansible-modules-core/issues/302 I think preserving the inode may be a tangent?  302 seems to show that ansible has an error that needs to be fixed.  I think the following playbook would be a reproducer for that error:

```
---
- hosts: localhost
  vars:
    - output_dir: /var/tmp
  tasks:
    - lineinfile:
        line: "this is a test"
        dest: "{{output_dir}}/source"
    - file:
        state: hard
        src: "{{output_dir}}/source"
        dest: "{{output_dir}}/hardlink"
    - fetch:
        src: "{{output_dir}}/source"
        dest: "{{output_dir}}/local-copy"
    - copy:
        src: "{{output_dir}}/local-copy/localhost/var/tmp/source"
        dest: "{{output_dir}}/hardlink"
```

sysadmin75:
@abadger  From your response, I think you're saying the breaking of hardlinks is an issue but it seems you're not convicnced it needs fixed. I do believe preserving hardlinks with the copy command is relevant since the cp command does it. Otherwise you're wanting users of ansible to use the cp command in cases like this instead of using your product.

bcoca:
first of all, i was wrong above, i had some systems with cp aliased which led me to confusion, second, I think we can detect this corner case and work around it. It will require thorough testing though.

rostchri:
I have similar problems using ansible version 1.9.1.

I'm using the template-module to write the redhat interface-files in /etc/sysconfig/network-sripts/ifcfg-ethX.
The following code-fragments are used:

```
name: Writing network-interface configuration
template: src=ifcfg-rhel-std.j2  dest=/etc/sysconfig/network-sripts/ifcfg-{{ item.device }} 
with_items: net_cfg.eth
when: net_cfg.eth is defined
```

```
ifcfg-rhel-std.j2 :
DEVICE={{ item.device }}
MTU={{ item.mtu|default(1500) }}
ONBOOT=yes
IPADDR={{ item.ipaddr }}
NETMASK={{ item.netmask }}
```

```
net_cfg:
  eth:
    - purpose: admin-network
      device:  "{{adm_interfacename}}"
      ipaddr:  "{{adm-ip}}"
      netmask: "{{networks[location]['adm'][admvlan]['netmask']}}"
    - purpose: backup-network
      device:  "{{backup_interfacename}}"
      ipaddr:  "{{backup-ip}}"
      netmask: "{{networks[location]['backup'][backupvlan]['netmask']}}"
```

The actual variables are set somewhere else individually for each host.

Everything works fine but on some installations somebody is frequently using the system-config-network-commands, which generates hard-links in /etc/sysconfig/networking/profiles/default/ and /etc/sysconfig/networking/devices/ to /etc/sysconfig/network-sripts/ifcfg-ethX

This really drives me mad! Because after that it simply breaks my ansible-task: 
template: src=ifcfg-rhel-std.j2  dest=/etc/sysconfig/network-sripts/ifcfg-{{ item.device }} 
with the confusing error-message: "msg: src and dest are required for creating links"

I tried to fix that by doing 
template: src=ifcfg-rhel-std.j2  dest=/etc/sysconfig/network-sripts/ifcfg-{{ item.device }}  state=file

it helps partly but it does'nt work in all cases (First case: No /etc/sysconfig/network-sripts/ifcfg-ethX-file exists at all, Second case: /etc/sysconfig/network-sripts/ifcfg-ethX-file exists without hardlinks, Third case: /etc/sysconfig/network-sripts/ifcfg-ethX-file exists with hardlinks attached to it). The template-module does'nt really know/interpret/accept the state-option ...

I would be really glad if this can be fixed in the ansible-core-modules. 
Finding a simple workaround for this bug seems to be really hard, and breaks the simplicity of my task.

Regards
Christian

jhawkesworth:
Sometimes I use

File module state=absent with a list of items to remove if they are present
before replacing them with my templates. I know it's not a single line fix
but I think might do what you need.

Hope this helps.

Jon

rostchri:
@jhawkesworth The problem is that I loose the functionality to see possible changes in the ifcfg-files using --diff --check options in ansible-playbook-command, when I implement a workaround with state=absent

dago:
@abadger Your snippet still shows the issue cleanly: the last copy task to a hardlink fails with the cryptic error message `absolute paths are required`:

```
TASK: [copy ] ***************************************************************** 
failed: [localhost] => {"checksum": "6476df3aac780622368173fe6e768a2edc3932c8", "failed": true}
msg: absolute paths are required
```

So to summarize we all agree that this is a bug that needs fixing.

Regarding the copy task we must define the intended behavior:
1. Keep the hardlink and replace the contents of the file. This conforms to standard behavior of `cp`.
2. Replace only the file and break the hardlinkage. This conforms to the behavior of `cfengine`.

Implementing (1) means breaking atomic changes with `link`, nonetheless I think it is the correct default behavior.

To gain (2) I suggest to use the `state=file` parameter (instead of the default in this case `state=hard`) from the `file` module to change the behavior.

There is one more related issue with the `template` module and hardlinks. Imagine this additional task:

```
    - template: src=sourcefile dest="{{output_dir}}/hardlink"
```

The issues are:
1. When the contents is changed the hardline is not preserved. When (1) is used above this should also change.
2, When the contents is not changed a cryptic error message is returned:

```
TASK: [template src=sourcefile dest="{{output_dir}}/hardlink"] **************** 
failed: [localhost] => {"failed": true}
msg: src and dest are required for creating links
```

I think this is definitely an error and should silently return no change instead.

dago:
The following playbook can be used to reproduce the error:

```
---
- hosts: localhost
  vars:
    - output_dir: /var/tmp
  tasks:
    - file: dest="{{output_dir}}/source" state=absent
    - file: dest="{{output_dir}}/hardlink" state=absent
    - copy: src=sourcefile dest="{{output_dir}}/source"
    - file: state=hard src="{{output_dir}}/source" dest="{{output_dir}}/hardlink"
    - template: src=sourcefile dest="{{output_dir}}/hardlink"
```

With the current code the output contains

```
TASK: [template src=sourcefile dest="{{output_dir}}/hardlink"] **************** 
failed: [localhost] => {"failed": true}
msg: src and dest are required for creating links
```

whereas the correct output with the patch applied is

```
TASK: [template src=sourcefile dest="{{output_dir}}/hardlink"] **************** 
ok: [localhost]
```

gregdek:
Looking at this PR and the PR proposed in https://github.com/ansible/ansible-modules-core/issues/1591...

Are they mutually exclusive? Is the fix proposed in core/1591 sufficient to solve the issue that this PR addresses?

dago:
I don't fully understand the other proposed patch, but I tried it and it does not fix the unchanged template issue. I just noticed that my PR has an issue with directories that already exist:

```
TASK: [file path=/var/tmp/blubb state=directory mode=0755 owner=root group=other] *** 
failed: [localhost] => {"failed": true, "gid": 1, "group": "other", "mode": "0755", "owner": "root", "path": "/var/tmp/blubb", "size": 2, "state": "directory", "uid": 0}
msg: file (/var/tmp/blubb) is directory, cannot continue
```

I made the condition more strict in the PR update.

rostchri:
@dago Your patch works for me. My template-code is now working. Great :+1:

amenonsen:
This patch no longer applies, but it's not just because of the action_plugins → plugins/actions rename, and it's not clear to me how to make the corresponding change (i.e. it seems non-trivial). Someone who actually understands the problem better should take a closer look at this.

sivel:
@sysadmin75,

Hi!

jctanner:
Closing due to submitter timeout.

PR #9808, opened 2014-12-14
Fix: Add support for SSL protocol version configuration option

swimlappy:
Hello, I am re-submitting this request for @bcoca.  This fix allows you to set the SSL protocol version in the ansible.cfg file.  It also fixes a couple of places where Ansible was not setting the SSL protocol version properly for URL calls.  So connecting to a host where only TLSv1 was supported would not work.

### DIFF
--- a/examples/ansible.cfg
+++ b/examples/ansible.cfg
@@ -212,3 +212,12 @@ accelerate_daemon_timeout = 30
 # is "no".
 #accelerate_multi_key = yes
 
+[ssl]
+# SSL/TLS Protocol
+# Configure the default protocol strength of any SSL/TLS connections
+# made by Ansible.  Valid values are
+#  SSLv2  - 0
+#  SSLv3  - 1
+#  SSLv23 - 2
+#  TLSv1  - 3
+ssl_protocol = 3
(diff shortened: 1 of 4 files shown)


swimlappy:
Any update on this possibly getting tested and merged in?  Thanks

swimlappy:
Great thank you

bcoca:
sooo .... modules cannot use configuration resources as it would require ansible to be installed on the target systems, need to rethink the approach, probably need to add a ssl_version= parameter to every module that uses this and pass it on, defaulting to TLSv1 as currently hardcoded.

sivel:
Could we use the REPLACER code in basic.py to store the SSL version in the generated module such as what is done for `ANSIBLE_VERSION`?

https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L32
https://github.com/ansible/ansible/blob/devel/lib/ansible/module_common.py#L34
https://github.com/ansible/ansible/blob/devel/lib/ansible/module_common.py#L168

Just a thought.

bcoca:
@sivel: that looks like the simplest fix, I still think we might wan't to add the SSL version setting per module as different machines/OS versions might require a different setting, but that can be addressed as it's own PR.

bcoca:
@swimlappy do you want to fix PR as per @sivel 's suggestion?

swimlappy:
I'm not familiar with the replacer code, so I will have to figure that out first.  Someone else might be able to get it done faster though.  I can look at it tonight.

bcoca:
check #10053, it does this for the v2 path

Issue #7732, opened 2014-06-11
debug: var=ansible_facts traceback


ktosiek:
##### Summary:

I just created a task to check if there's an ansible_facts variable, and it breaks Ansible.
##### Steps To Reproduce:

Running this playbook:

```
- hosts: localhost
  tasks:
    - debug: var=ansible_facts
```
##### Expected Results:

message from debug and normal exit
##### Actual Results:

```
$ ansible-playbook -i localhost_inv ansible_facts.yml

PLAY [localhost] ************************************************************** 

GATHERING FACTS *************************************************************** 
ok: [localhost]

TASK: [debug var=ansible_facts] *********************************************** 
ok: [localhost] => {
    "ansible_facts": "{{ ansible_facts }}"
}
Traceback (most recent call last):
  File "/home/tomek/ansible/ansible/bin/ansible-playbook", line 320, in <module>
    sys.exit(main(sys.argv[1:]))
  File "/home/tomek/ansible/ansible/bin/ansible-playbook", line 260, in main
    pb.run()
  File "/home/tomek/ansible/ansible/lib/ansible/playbook/__init__.py", line 323, in run
    if not self._run_play(play):
  File "/home/tomek/ansible/ansible/lib/ansible/playbook/__init__.py", line 677, in _run_play
    if not self._run_task(play, task, False):
  File "/home/tomek/ansible/ansible/lib/ansible/playbook/__init__.py", line 475, in _run_task
    self.SETUP_CACHE[host].update(facts)
ValueError: dictionary update sequence element #0 has length 1; 2 is required
```

mscherer:
Indeed, i can reproduce on git HEAD. The problem is there:

https://github.com/ansible/ansible/blob/devel/lib/ansible/runner/action_plugins/debug.py#L55

Since ansible_facts is used to communicate facts from modules ( http://jpmens.net/2012/07/15/ansible-it-s-a-fact/ ), the debug module should use a different way ( ie, not foo=value ). But doing this would break the interface. This is however not documented ( http://docs.ansible.com/debug_module.html ).

abadger:
# Closing This Ticket

Hi!

We believe recent commits (likely detailed above) should resolve this question or problem for you.

This will also be included in the next major release.

If you continue seeing any problems related to this issue, or if you have any further questions, please let us know by stopping by one of the two mailing lists, as appropriate:
- https://groups.google.com/forum/#!forum/ansible-project - for user questions, tips, and tricks
[...]

PR #9810, opened 2014-12-14
Do not use the variable name as a key for the result of the module

mscherer:
Using the variable name can interfere with various systems
used to communicate between modules and ansible ( as
reported in #7732 , where ansible_facts is a reserved key for
updating the fact cache, for example ).


abadger:
I see from the bug that you noted this will change the value returned.  Discussed with @bcoca and we came to the conclusion that if someone was using register with debug they were likely doing the wrong thing.  Thanks for the fix and the analysis!

# Merged

Hi!

issue is resolved.

Thank you!

PR #9812, opened 2014-12-14
Allow globbing in creates= and removes= directives

fgsch:
Fixes 1904

See https://github.com/ansible/ansible-modules-core/pull/513


gildegoma:
:+1: see also https://github.com/ansible/ansible-modules-core/pull/513#issuecomment-68891643.

fgsch:
@abadger this should be merged as well as part of https://github.com/ansible/ansible-modules-core/pull/513.

Issue #9784, opened 2014-12-11
run_once tasks are skipped without checking the delegate_to host


ccciudatu:
##### Issue Type:

Bug report
##### Ansible Version:

1.8
##### Environment:

Any
##### Summary:

As `run_once` will always target the first host in the list, if that host is supposed to skip a set of tasks (due to some condition, for instance), it will also skip any `run_once` tasks even though they may `delegate_to` other hosts which are willing to run that task.
##### Steps To Reproduce:
- A playbook targets a group of hosts (say `host1`, `host2`, `host3`)
- A role in that playbook is applied conditionally (`when: some_flag`).
- `host1` does not have the flag set, while `host2` and `host3` do
- The "regular" tasks within that role are therefore only applied on `host2` and `host3` and are skipped on `host1` (as expected)
##### Expected Results:

In the above scenario, for a task marked with `run_once: yes` and `delegate_to: host2` I would expect it should run (on `host2`), since it does have the flag set
##### Actual Results:

However, the above-mentioned `run_once` task _will also be skipped_ (by `host1`, as it's the first one in the list).

I don't expect this to be the intended behavior (but merely a consequence of the current implementation), as that task which should run only on `host2` has no reason to be skipped.

abadger:
PR merged.  Thanks!

weiyu10:
This bug not fixed, I have a comment in https://github.com/ansible/ansible/pull/9814

weiyu10:
## This is my site.yaml
- hosts: all
  
  tasks:
  - name: debug for run_once and delegate_to
    debug: msg=success
    run_once: True
    delegate_to: "{{ groups['compute'][0] }}"
    when: inventory_hostname in groups['compute']

This is my host
[compute]
compute01

[control]
control01
control02
control03

This is result 

TASK: [debug for run_once and delegate_to] ************************************
skipping: [control01]

PLAY RECAP ********************************************************************
compute01                  : ok=1    changed=0    unreachable=0    failed=0
control01                  : ok=1    changed=0    unreachable=0    failed=0
control02                  : ok=1    changed=0    unreachable=0    failed=0
control03                  : ok=1    changed=0    unreachable=0    failed=0

jimi-c:
@spunkzwy this has been fixed in devel and will be included in 2.0.1. Which version of Ansible are you running?

weiyu10:
@jimi-c thanks
[...]

PR #9814, opened 2014-12-15
run_once tasks are skipped without checking the delegate_to host #9784

ccciudatu:
The fix is to choose the `delegate_to` host (when defined and included in the target host list) instead of the first one in the list for running the tasks marked with `run_once`.

### DIFF
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -1461,9 +1461,15 @@ class Runner(object):
             # Expose the current hostgroup to the bypassing plugins
             self.host_set = hosts
             # We aren't iterating over all the hosts in this
-            # group. So, just pick the first host in our group to
+            # group. So, just choose the "delegate_to" host if that is defined and is
+            # one of the targeted hosts, otherwise pick the first host in our group to
             # construct the conn object with.
-            result_data = self._executor(hosts[0], None).result
+            if self.delegate_to is not None and self.delegate_to in hosts:
+                host = self.delegate_to
+            else:
+                host = hosts[0]
+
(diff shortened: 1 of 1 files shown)


lib/ansible/runner/__init__.py
```diff
@@ -1461,9 +1461,11 @@ def run(self):
             # Expose the current hostgroup to the bypassing plugins
             self.host_set = hosts
             # We aren't iterating over all the hosts in this
-            # group. So, just pick the first host in our group to
+            # group. So, just choose the "delegate_to" host if that is defined and is
+            # one of the targeted hosts, otherwise pick the first host in our group to
             # construct the conn object with.
-            result_data = self._executor(hosts[0], None).result
+            host = self.delegate_to if self.delegate_to is not None and self.delegate_to in hosts else hosts[0]
```
abadger: In general, unless the condition and the values are very simple, don't use a ternary.  it's easier to read if it's in long form.

abadger:
Can you give an example of where this is failing?  delegate_to and run_once are used together very frequently so I'm not sure under what conditions this change would be necessary.

ccciudatu:
I described a failure scenario within the issue referenced by this commit: https://github.com/ansible/ansible/issues/9784.

Basically, a `when` test defined for such a task will be evaluated on the first host in the list, disregarding the `delegate_to` setting. So if I want to run something once on `hosts[1]` (using `delegate_to`) I would not expect the command to be skipped just because `hosts[0]` does not qualify to run that task.

abadger:
Thanks, Had a chance to review the code and it looks good.

# Merged

Hi!

issue is resolved.

Thank you!

Issue #6346, opened 2014-03-08
ansible-pull should support key_file or accept_hostkey


HelloGrayson:
This makes it more difficult than needed to bootstrap instances... Is there no way to pass  arguments to the git module while using ansible-pull? I'm on Ansible 1.5.

ansibot:
Thanks for filing a ticket! I am the friendly GitHub Ansibot. I see you did not fill out your issue description based on our new issue template. Please copy the contents of this [template](https://raw2.github.com/ansible/ansible/devel/examples/issues/ISSUE_TEMPLATE.md) and **paste it into the description** of your ticket. 

Our system will automatically close tickets that do not have an issue template filled out within 7 days. Please note that due to large interest in Ansible, humans may not comment on your ticket if you ask them questions. Don't worry, you're still in the queue and the robots are looking after you.

mpdehaan:
So both of these options are new in 1.5, which is why they are not currently present.

Additions would be welcome!

kavu:
@mpdehaan may I try this out? Assuming only `source_control/git` supports `key_file` and `accept_hostkey` I think it won't be difficult to add two new options to the command invocation.

jaydrogers:
I was able to get a workaround in **Ansible 1.5.4** by adding `key_file=/path/to/my/key.pub` to the `/etc/ansible/ansible.cfg` file.

When I ran `ansible-pull -d /path/to/my/repo -U git@github.com:username/my-repo.git` I was able to get access to a private repository.

PR #9821, opened 2014-12-15
Add --private-key opt

nawadanp:
Add parameter --private-key on the ansible-pull script, to specify the ssh private key to use on the repository.
It check if the key file exists before running the checkout.


nawadanp:
Duplicate to #6346 
Already solved here https://github.com/ansible/ansible/commit/44dfda855a02292c6f02b4b582f21bf003b011c9

PR #9824, opened 2014-12-16
Fixed error handling for the enabling of PS Remoting

woop:
Power Shell Remoting will throw an exception if you attempt to enable it while one of your network interfaces is set to "public". Currently this error is hidden and the script continues to execute until it attempts to test PS Remoting to localhost. At that point the code fails with the following error:

"Unable to establish an HTTP or HTTPS remoting session."

This happens because PS Remoting didn't start. In actual fact the error that should have been thrown was the one that occurred when an attempt was made to enable PS Remoting.

I have removed the try/catch and set the ErrorAction to Stop so that the script doesn't continue and the correct error is displayed. It will also now be clear why the script failed and how to rectify the problem.


cchurch:
Makes sense to me.  Thanks!

PR #9825, opened 2014-12-16
Allow ec2 tags to be used to address servers in ec2 dynamic inventory.

kalefranz:
My use case for this PR is addressing instances using domain names separate from EC2's public and private dns names. The PR is completely backward-compatible.

### DIFF
--- a/plugins/inventory/ec2.ini
+++ b/plugins/inventory/ec2.ini
@@ -24,14 +24,17 @@ regions_exclude = us-gov-west-1,cn-north-1
 # This is the normal destination variable to use. If you are running Ansible
 # from outside EC2, then 'public_dns_name' makes the most sense. If you are
 # running Ansible from within EC2, then perhaps you want to use the internal
-# address, and should set this to 'private_dns_name'.
+# address, and should set this to 'private_dns_name'. The key of an EC2 tag
+# may optionally be used; however the boto instance variables hold precedence
+# in the event of a collision.
 destination_variable = public_dns_name
 
 # For server inside a VPC, using DNS names may not make sense. When an instance
 # has 'subnet_id' set, this variable is used. If the subnet is public, setting
 # this to 'ip_address' will return the public IP address. For instances in a
 # private subnet, this should be set to 'private_ip_address', and Ansible must
-# be run from with EC2.
+# be run from with EC2. The key of an EC2 tag may optionally be used; however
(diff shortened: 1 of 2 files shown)


kalefranz:
After submitting this PR, and then reading the forum post [here](https://groups.google.com/forum/#!topic/ansible-devel/qANPvLNyL8Y), I realized #7601, #7395, and #9630.  Of the three PRs, #7395 and #9630 actually serve different purposes.

This PR and #7601 have a functionally similar purpose, with slightly different implementations.

kalefranz:
I guess I should also add specifically my example use case.  On each EC2 instance I want in my inventory, I've added an `FQDN` tag.  Instances without an `FQDN` tag do not show up in my inventory, and ansible is able to communicate with the instance via the value in the `FQDN` tag.  Also, the value of the `FQDN` tag is displayed in the Tower inventory list (instead of the private_ip, for example).

cchurch:
My 2 cents... I may be interested in this change for the next Tower release.

I prefer your implementation over https://github.com/ansible/ansible/pull/7601, as it is a little simpler and doesn't define yet another option in `ec2.ini`.

https://github.com/ansible/ansible/pull/7395 allows for setting the hostname from a variable or tag, but keeps the EC2 DNS name via `ansible_ssh_host`.  I think that PR and yours can coexist.  I do prefer its  convention of only checking tags if the value starts with `tag_`.

sankalp-khare:
I don't get it. When I do this :

``` ini
vpc_destination_variable = Name
```

in place of

``` ini
vpc_destination_variable = private_ip_address
```

How, then, does ansible ssh into the instances for doing its work?
I'm geting errors like:

```
TASK: [Get instance ec2 facts] ************************************************
fatal: [mvue1-01-uat-worker-01] => SSH encountered an unknown error during the connection. We recommend you re-run the command using -vvvv, which will enable SSH debugging output to help diagnose the issue
```

where `mvue1-01-uat-worker-01` is my instance's name

sankalp-khare:
Nevermind, I figured it out. I'd also then have to have a value inside the tag that makes sense to ssh. Interesting feature, but not what I thought it was.

How about an option to show the `tag:Name` in ansible output but connect using the `destination_variable`?

sankalp-khare:
But that would need to go outside the scope of the ec2 inventory plugin

cchurch:
@sankalp-khare #7395 should address your issue once it is merged.

Issue #9831, opened 2014-12-16
Wishlist: Please add additional math filters to jinja2 (logarithm)


xicesky:
##### Summary:

Ansible is missing a logarithm function.
##### Steps To Reproduce:

Example:

```
- name: Calculate the logarithm of 8 to the base of 2
  debug:
    msg: "{{ 8 | log(2) }}"
```

... or something similar.
##### Expected Results:

Example output:
3
##### Actual Results:

By itself, jinja2 only provides a few math functions. As the documentation in http://jinja.pocoo.org/docs/dev/templates/#math states: "This is rarely useful in templates but exists for completeness’ sake."
In Ansible, however, jinja2 is just the only way to do math calculations.

In my case, i was looking for a way to round up to the next power of 2, which is commonly done by taking the logarithm, rounding upwards, then taking the power of 2 of the result. Sadly jinja2 is missing the logarithm part. I suppose logarithms are a fairly common operation and should not require an additional addon.

abadger:
# Closing This Ticket

Hi!

We believe recent commits (likely detailed above) should resolve this question or problem for you.

This will also be included in the next major release.

If you continue seeing any problems related to this issue, or if you have any further questions, please let us know by stopping by one of the two mailing lists, as appropriate:
- https://groups.google.com/forum/#!forum/ansible-project - for user questions, tips, and tricks
- https://groups.google.com/forum/#!forum/ansible-devel - for strategy, future planning, and questions about writing code

Because this project is very active, we're unlikely to see comments made on closed tickets, but the mailing list is a great way to ask questions, or post if you don't think this particular
issue is resolved.

Thank you!

PR #9832, opened 2014-12-16
math filters!

bcoca:
provides a bit more than wished for in #9831


xicesky:
Awesome, that was fast, ty!

abadger:
Thanks bcoca!

xicesky:
Small typo though:

```
return = math.log(x, base)
```

should probably be:

```
return math.log(x, base)
```

Issue #4104, opened 2013-09-13
When statement errors when variable is used alone


josephtate:
Given the variable
central_log_server set to either "some-host-name-10" or "off". I believe the dashes in the hostname are important because of the error message.

And the task:
- Central logging
  copy: ...
  when: central_log_server
- Turn off central logging
  file: state=absent...
  when: not central_log_server

I get a traceback:
TASK: [Central logging] ********************************
fatal: [test01.crunch.io] => Traceback (most recent call last):
  File "/home/jtate/.virtualenvs/pow/local/lib/python2.7/site-packages/ansible/runner/**init**.py", line 382, in _executor
    exec_rc = self._executor_internal(host, new_stdin)
  File "/home/jtate/.virtualenvs/pow/local/lib/python2.7/site-packages/ansible/runner/__init__.py", line 471, in _executor_internal
    return self._executor_internal_inner(host, self.module_name, self.module_args, inject, port, complex_args=complex_args)
  File "/home/jtate/.virtualenvs/pow/local/lib/python2.7/site-packages/ansible/runner/__init__.py", line 555, in _executor_internal_inner
    if not utils.check_conditional(cond, self.basedir, inject, fail_on_undefined=self.error_on_undefined_vars):
  File "/home/jtate/.virtualenvs/pow/local/lib/python2.7/site-packages/ansible/utils/__init__.py", line 172, in check_conditional
    conditional = template.template(basedir, presented, inject)
  File "/home/jtate/.virtualenvs/pow/local/lib/python2.7/site-packages/ansible/utils/template.py", line 310, in template
    varname = template_from_string(basedir, varname, vars, fail_on_undefined)
  File "/home/jtate/.virtualenvs/pow/local/lib/python2.7/site-packages/ansible/utils/template.py", line 512, in template_from_string
    res = jinja2.utils.concat(t.root_render_func(t.new_context(_jinja2_vars(basedir, vars, t.globals, fail_on_undefined), shared=True)))
  File "<template>", line 9, in root
TypeError: unsupported operand type(s) for -: 'StrictUndefined' and 'int'

If I change the when: block to "not not central_log_server" or "central_log_server is defined and central_log_server" it works.

Is the variable name being replaced with the value before being fed to jinja2 so that it interprets "some" "host" and 10 as tokens with minus operators between them?

jimi-c:
What version of ansible are you running?

josephtate:
I just pulled HEAD on the devel branch.

[...]

PR #9834, opened 2014-12-17
Support variable values with dashes

nathancahill:
Undoes 35d138a since it's not needed anymore. Playbook that tests the original bug and the new dashes bug:

```
- name: Test play
  hosts: localhost
  vars:
    with_dash: "a-value-with-dashes"
    with_dash_and_number: "some-host-name-10"
  tasks:
    - debug: msg="with_dash"
      when: with_dash
    - debug: msg="with_dash_and_number"
      when: with_dash_and_number
```

### DIFF
--- a/lib/ansible/utils/__init__.py
+++ b/lib/ansible/utils/__init__.py
@@ -269,7 +269,7 @@ def check_conditional(conditional, basedir, inject, fail_on_undefined=False):
 
     conditional = conditional.replace("jinja2_compare ","")
     # allow variable names
-    if conditional in inject and '-' not in str(inject[conditional]):
+    if conditional in inject:
         conditional = inject[conditional]
     conditional = template.template(basedir, conditional, inject, fail_on_undefined=fail_on_undefined)
     original = str(conditional).replace("jinja2_compare ","")


juliedavila:
Putting this into my source, I'm still unable to evaluate a variable with multiple dashes. 

Try this:

``` yaml
vars: 
   dict_by_num:
       10: "-SOME -VAR"
tasks:
  - debug: msg={{ dict_by_num.10 }}
```

This currently returns

``` shell
ok: [localhost] => {
    "msg": "-SOME"
}
```

sivel:
@nousdefions your example works for me, with one modification, to handle lack of quoting:

```
  - debug: msg="{{ dict_by_num.10 }}"
```

Otherwise your `-VAR` is seen as a string not part of the `msg` argument since you also have spaces in the string, and as such is not printed.

juliedavila:
ah you're completely right, good catch.

bcoca:
Soo, I run the following without problems, not sure what error this fixes

```
- name: Test play
  hosts: localhost
  vars:
    with_dash: "a-value-with-dashes"
    with_dash_and_number: "some-host-name-10"
    with_dash_and_space: "some -host -name-10"
    with_dict_num:
        10: '-test -var'
  tasks:
    - debug: msg="with_dash {{with_dash}}"
      when: with_dash

    - debug: msg="with_dash_and_number {{with_dash_and_number}}"
      when: with_dash_and_number

    - debug: msg="with_dash_and_space {{with_dash_and_space}}"
      when: with_dash_and_space

    - debug: msg="with_dict_num.10 {{with_dict_num.10}}"
      when: with_dict_num.10
```

nathancahill:
Yes, the error is that it runs without exceptions, because a special case was added (35d138a) when evaluating variable values with dashes to fix #4104. However, it's a hack, and it's not needed anymore since the underlying Jinja templating has changed, which fixed the bug it was patching.

`"a-value-with-dashes"` should behave the same as `"random string value"`. The error for all of the tasks in @bcoca's example should be: `error while evaluating conditional`

bcoca:
reverting this broke plugin loading

bcoca:
strings in general evaluate to 'true', so that they have dashes or not should not matter, in all these cases it should NOT be a exception.

martijnvermaat:
I agree, but currently it _only_ works for strings with dashes.

For example, if you remove the dashes from any value in @bcoca's example playbook, it will fail.

```
- name: Test play
  hosts: localhost
  vars:
    with_dash: "a-value-with-dashes"
    with_dash_and_number: "some-host-name-10"
    with_dash_and_space: "some -host -name-10"
    with_dict_num:
        10: '-test -var'
    without_dash: "a value without dashes"
  tasks:
    - debug: msg="with_dash {{with_dash}}"
      when: with_dash

    - debug: msg="with_dash_and_number {{with_dash_and_number}}"
      when: with_dash_and_number

    - debug: msg="with_dash_and_space {{with_dash_and_space}}"
      when: with_dash_and_space

    - debug: msg="with_dict_num.10 {{with_dict_num.10}}"
      when: with_dict_num.10

    - debug: msg="without_dash {{without_dash}}"
      when: without_dash
```

```
PLAY [Test play] ************************************************************** 

TASK: [debug msg="with_dash {{with_dash}}"] *********************************** 
ok: [default] => {
    "msg": "with_dash a-value-with-dashes"
}

TASK: [debug msg="with_dash_and_number {{with_dash_and_number}}"] ************* 
ok: [default] => {
    "msg": "with_dash_and_number some-host-name-10"
}

TASK: [debug msg="with_dash_and_space {{with_dash_and_space}}"] *************** 
ok: [default] => {
    "msg": "with_dash_and_space some -host -name-10"
}

TASK: [debug msg="with_dict_num.10 {{with_dict_num.10}}"] ********************* 
ok: [default] => {
    "msg": "with_dict_num.10 -test -var"
}

TASK: [debug msg="without_dash {{without_dash}}"] ***************************** 
fatal: [default] => Failed to template {% if a value without dashes %} True {% else %} False {% endif %}: template error while templating string: expected token 'end of statement block', got 'value'

           to retry, use: --limit @/home/martijn/playbook.retry

default                    : ok=4    changed=0    unreachable=1    failed=0   

```

I tried many things, but couldn't get a conditional to work on a variable without a dash in its value.

martijnvermaat:
Actually I now finally did find a workaround:

```
when: not not without_dash
```

Yes, it's ugly, I think it should just work without this.

PR #9835, opened 2014-12-17
cloudstack: add dynamic inventory

resmo:


plugins/inventory/cloudstack.py
```diff
@@ -0,0 +1,232 @@
...
+        if projects:
+            for p in projects['project']:
+                if p['name'] == project or p['id'] == project:
+                    return p['id']
+        print >> sys.stderr, "Error: Project %s not found." % project
+        sys.exit(1)
+
+
+    def get_host(self, name, project_id=''):
+        hosts = self.cs.listVirtualMachines(projectid=project_id)
```
TomaszKlosinski, line 131: This doesn't work when list of projects is empty (example: exoscale.ch). It results with following error:

`Traceback (most recent call last):
  File "./cloudstack.py", line 232, in <module>
    CloudStackInventory()
  File "./cloudstack.py", line 113, in __init__
    data = self.get_list()
  File "./cloudstack.py", line 178, in get_list
    groups = self.cs.listInstanceGroups(projectid=project_id)
  File "/Library/Python/2.7/site-packages/cs/client.py", line 105, in handler
    return self._request(command, **kwargs)
  File "/Library/Python/2.7/site-packages/cs/client.py", line 157, in _request
    response.status_code), response, data)
cs.client.CloudStackException: ('HTTP 431 response from CloudStack', <Response [431]>, {u'errorcode': 431, u'uuidList': [], u'cserrorcode': 9999, u'errortext': u'Unable to execute API command listinstancegroups due to invalid value. Invalid parameter projectid value= due to incorrect long value format, or entity does not exist or due to incorrect parameter annotation for the field in api cmd class.'})`

It works, when I remove `projectid=project_id` form the API calls.

TomaszKlosinski:
Additionally, it doesn't use IP of the instance to connect, but instead is trying to use its name as a hostname:
```
$ ansible -i cloudstack.py -m ping server-1                                                                                                                                                                                                                       server-1 | UNREACHABLE! => {
    "changed": false,
    "msg": "Failed to connect to the host via ssh: ssh: Could not resolve hostname server-1: nodename nor servname provided, or not known\r\n",
    "unreachable": true
}
```
But it works fine when I run:
```
./cloudstack.py --host server-1
```

Is there a way to force it to use IP instead?

TomaszKlosinski:
Ok, it works when I replaced:
```
data['_meta']['hostvars'][host_name]['default_ip'] = nic['ipaddress']
```
with:
```
data['_meta']['hostvars'][host_name]['ansible_host'] = nic['ipaddress']
```

TomaszKlosinski (commented):
@bcoca , @resmo ,  this script doesn't work with Exoscale (CloudStack-based public cloud in Switzerland). I can help testing if you guys plan to fix it.

resmo:
@tklo please do not comment on 4 old PRs. create an issue. I'll fix it. I have a exoscale account but thanks for the offer!

PR #9766, opened 2014-12-10
Expand Memory Facts in facts.py

juliedavila:
This fixes problem originally discussed here: https://github.com/ansible/ansible/issues/8035

As discussed in the issue tracker the output now looks like the following:

```
        "ansible_memory": {
            "nocache": {
                "free": 395,
                "used": 197
            },
            "real": {
                "free": 37,
                "total": 592,
                "used": 555
            },
            "swap": {
                "cached": 0,
                "free": 2047,
                "total": 2047,
                "used": 0
            }
        }
```

And it includes the old top-level facts to preserve backwards compatibility.

Issue #8035, opened 2014-07-03
ansible_memfree_mb fact combines disk-cache use of memory


juliedavila:
##### Summary:

I setup a play to alert whenever memory utilization is dangerously high. I immediately received a handful of alerts and was a bit concerned until I realized it takes into account memory being used by the disk cache and not 'real' memory utilization.
##### Steps To Reproduce:

To reproduce just compare the ansible_memfree_mb fact, which in my example is 

```
        "ansible_memfree_mb": 10550,
        "ansible_memtotal_mb": 15042,
```

with the return value of running

```
free -m
```
##### Actual Results:

In my case returns

```
             total       used       free     shared    buffers     cached
Mem:         15042       4483      10559        139        231       2859
-/+ buffers/cache:       1393      13649
Swap:            0          0          0
```

As you can see, the ansible_facts are correct but misleading, in this case saying only 10gb are free when in fact I have 13gb that are useable.

Perhaps this could be added as a ansible_nocache_memfree_mb fact?

mpdehaan:
Sounds like an RFE for another fact yes.   I might have preferences on a shorter name.

juliedavila:
What's RFE? and yeah, agreed on the name.

jimi-c:
@nousdefions RFE = "request for enhancement". And I see nothing wrong with `ansible_memfree_minus_buffers_and_cache_mb`... :) Really, it almost sounds like this might be better as a dictionary-type fact, like:

<pre>
'ansible_memory': {
   'real': {
      'total': 15042,
      'free': 10559,
      ...
   },
   'nocache': {
      'used': 1393,
      'free': 13649
   },
   'swap': {
      'total': 0,
      'used': 0,
      'free': 0
   }
}
</pre>

Just a thought.

juliedavila:
Ha! Hilarious and yes the dictionary-type makes sense.

juliedavila:
Fixed issue here https://github.com/ansible/ansible/pull/8071

jimi-c:
Yes, I'm going to go ahead and close this, and we'll track the fix in the PR. Thanks!

PR #9839, opened 2014-12-17
Memfacts

juliedavila:
As opened originally in #9766 this is the same thing just in it's own branch (as it should have been from the start) to avoid pr conflicts. Original text below:

This fixes problem originally discussed here: https://github.com/ansible/ansible/issues/8035

As discussed in the issue tracker the output now looks like the following:

```
        "ansible_memory": {
            "nocache": {
                "free": 395,
                "used": 197
            },
            "real": {
                "free": 37,
                "total": 592,
                "used": 555
            },
            "swap": {
                "cached": 0,
                "free": 2047,
                "total": 2047,
                "used": 0
            }
        }
```

And it includes the old top-level facts to preserve backwards compatibility.

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -571,6 +571,8 @@ class LinuxHardware(Hardware):
 
     platform = 'Linux'
     MEMORY_FACTS = ['MemTotal', 'SwapTotal', 'MemFree', 'SwapFree']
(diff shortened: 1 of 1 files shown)


mscherer:
Could you rewrite git history to merge the fix and various clean up ?

juliedavila:
@mscherer yeah, good idea. Done.

PR #9841, opened 2014-12-17
Testing Pull Requests, submodule testing addition to docs

juliedavila:
Simple but helpful addition to the documentation for testing submodule pull requests


amenonsen:
This looks OK, but again, I'd prefer to see a single sentence saying you need to cd into the right module directory before doing the fetch/checkout rather than duplicating the block.

bcoca:
actually for testing i would use a local 'library' directory adjacent to play, so as not to have to deal with root permissions nor backing out changes for the ansible installation that might even be shared.

juliedavila:
Ha! Completely forgot about this PR. I'll update it later on.

amenonsen:
ping @defionscode

PR #9853, opened 2014-12-18
Make issue rypes as an enumeration

axos88:
Easier to copy&paste, and delete all except the correct line.


amenonsen:
Looks good.

PR #9866, opened 2014-12-19
Improve jsonfile

mscherer:
This make the plugin use a internal format more resilient against someone fiddling directly with it, by using a extension.

### DIFF
--- a/lib/ansible/cache/jsonfile.py
+++ b/lib/ansible/cache/jsonfile.py
@@ -42,6 +42,9 @@ class CacheModule(BaseCacheModule):
                 utils.warning("error while trying to create cache dir %s : %s" % (self._cache_dir, str(e)))
                 return None
 
+    def _get_cache_file(self, key):
+        return "%s/%s" % (self._cache_dir, key)
+
     def get(self, key):
 
         if key in self._cache:
@@ -50,7 +53,7 @@ class CacheModule(BaseCacheModule):
         if self.has_expired(key):
            raise KeyError
 
-        cachefile = "%s/%s" % (self._cache_dir, key)
+        cachefile = self._get_cache_file(key)
         try:
             f = open( cachefile, 'r')
         except (OSError,IOError), e:
@@ -66,7 +69,7 @@ class CacheModule(BaseCacheModule):
 
         self._cache[key] = value
 
-        cachefile = "%s/%s" % (self._cache_dir, key)
+        cachefile = self._get_cache_file(key)
         try:
             #TODO: check if valid keys can have invalid FS chars, base32?
             f = open(cachefile, 'w')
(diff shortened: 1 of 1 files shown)


lib/ansible/cache/jsonfile.py
```diff
@@ -98,7 +101,9 @@ def has_expired(self, key):
     def keys(self):
         keys = []
         for k in os.listdir(self._cache_dir):
-            if not (k.startswith('.') or self.has_expired(k)):
+            if not (k.startswith('.') or self.has_expired(k)) and k.endswith('.json'):
+                # remove .json
+                k = k[:-5]
```
resmo, line 106: This seems to be quite a hack. Why not just keep it simple without `.json`?

mscherer, line 106: Well, the whole point is to have a more robust naming, in case of direct manipulation, so removing the .json would defeat the purpose ( now, i can aslo accept that this is maybe not so useful as it seemed when i did the PR )

bcoca:
I'm not sure this is needed, if someone is directly manipulating cache files ... they really shouldn't. Adding an extension does not seem to help much with that  either, can you give a usage example?

mscherer:
It was a long time ago, but I think what I did was just "open the file to see its content with vim", and this broke ansible, because the .swp was read as one of the json file ( while it wasn't ).

amenonsen:
I think this PR's description is misleading: the changes aren't so much about resilience to fiddling. The first patch factors out repeated cachefile assignments, and the second makes it use and require a `.json` extension for the cache files (so that it doesn't pick up stray files in the `keys` method). I think both changes are perfectly sensible, and are still relevant to `lib/ansible/plugins/cache/jsonfile.py`.

amenonsen:
Unfortunately, if I understand correctly, applying the second patch breaks backwards compatibility with existing cached facts. We could apply the patch so that we create only `.json` files, but read in only files that end in `.json` or have no extension… but the problem is that the keys are hostnames, so they may well look like they have extensions other than .json. So this might have been a good idea to begin with, but it's hard to implement in retrospect without more complexity than seems justified. And without the second patch, the first patch is pretty academic: a good idea, but not compelling.

@mscherer any thoughts on this?

mscherer:
I think in the end, we have to decide how much compatibility we want to give for a extension that wasn't even documented in the first place. I can see how breaking the cache would be breaking some people setup, but we can also figure that would go in a major release, and could be acceptable there.

amenonsen:
@mscherer any further plans for this?

mscherer:
We still didn't decide of the plan or anything. I guess i am the only one who care, but I also do not have much time to rebase, so we can drop it if it doesn't interest anyone but me. I can trivially resubmit if there is a need later or anything.

amenonsen:
Closing per above comment.

PR #9869, opened 2014-12-19
Display Cyan error message instead of Red when ignoring errors.

the-gigi:
Some tasks are designed to fail under certain conditions and the play should continue as described here:
http://docs.ansible.com/playbooks_error_handling.html

Ansible provides the 'ignore_errors' field you can set to 'yes'. The problem is that Ansible currently still displays a Red error message in this case, which is visually distracting and causes the impression to a casual observer that something went wrong. This change in the on_failed() callback checks if ignore_errors is True and if so displays the error message in Cyan, which is the color of the 'ignoring...' message.

### DIFF
--- a/lib/ansible/callbacks.py
+++ b/lib/ansible/callbacks.py
@@ -80,13 +80,13 @@ def log_lockfile():
     uid = os.getuid()
     path = os.path.join(tempdir, ".ansible-lock.%s" % uid)
     lockfile = open(path, 'w')
-    # use fcntl to set FD_CLOEXEC on the file descriptor, 
+    # use fcntl to set FD_CLOEXEC on the file descriptor,
     # so that we don't leak the file descriptor later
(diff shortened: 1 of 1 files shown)


amenonsen:
As colour changes go, the rationale for this seems more sound than most. But (a) the code has changed and this change would have to be made in multiple places in `lib/ansible/plugins/callback` now and (b) note that we display expanded output on failure, so maybe a giant block of cyan is not the best idea either.

(Disclaimer: my intense dislike of cyan may make me an impartial observer in this case.)

bcoca:
there is a cyan message in this case that shows 'ignoring errors', we really don't want to confused the user making the task looked as it was skipped. If you really want this it is now in v2, the control of the callback plugins so it should be easy to create w/o needing to patch main ansible code.

Closing this request as we won't make this the default.

PR #9878, opened 2014-12-22
Handle quoting of values in dict parameters

abadger:
In https://github.com/ansible/ansible-modules-core/issues/543 the user is trying to set a dictionary parameter via a string (non-complex args syntax).  This is failing because the value contains a comma inside of quotes and the naive parsing of the parameter thinks that the comma is a dictionary field separator.

This will fix that.  As a side effect it also allows space to be a separator, not just comma.  For example the new code should do this:

```
in a playbook:
ex_dictionary: a='test' b='test,test_b',c='foo bar',    d='yep, "this" too'
==>
inside the module:
ex_dictionary == {'a': 'test', 'b': 'test,test_b', 'c': 'foo bar', 'd': 'yep, "this" too'}
```

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -1070,7 +1070,32 @@ class AnsibleModule(object):
                         raise TypeError('unable to evaluate string as dictionary')
                     return result
             elif '=' in value:
-                return dict([x.strip().split("=", 1) for x in value.split(",")])
+                fields = []
+                field_buffer = []
+                in_quote = False
+                in_escape = False
+                for c in value.strip():
+                    if in_escape:
+                        field_buffer.append(c)
+                        in_escape = False
+                    elif c == '\\':
+                        in_escape = True
+                    elif not in_quote and c in ('\'', '"'):
+                        in_quote = c
+                    elif in_quote and in_quote == c:
+                        in_quote = False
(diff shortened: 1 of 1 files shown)


abadger:
Comments on the PR:
- Doesn't handle escaping characters.  For instance var: '\"air quotes\"' results in literal string: air quotes  (double quotes are omitted)
  - Looks like current code handles escaping differently but also incorrectly.  In current code, we end up with the literal string: \"air quotes\" (backslash is included in the literal string)
- Always troubled by having to write a new parser.  On the other hand, the only parser currently available to modules is splitter.parse_args.  That parser doesn't handle commas as a separator so we can't use it verbatim.  It's also going away in v2 so we'll need a different parser for that tree.

abadger:
New code now handles escaped quotes.  Still a new parser but v2 won't include splitter.parse_args in the module code, the splitter.parse_args parser doesn't handle commas so it wouldn't handle this case anyway (at least, in a backwards compatible manner), and really, we're replacing an existing parser rather than adding a new one (this parser is just more complete than the old one.)

amenonsen:
@abadger what's the status of this now?

abadger:
Rebased and ready to be reviewed.

abadger:
bcoca reviewed.  Merging.

PR #9881, opened 2014-12-23
facts caching: fix missing json in jsonfile caching

resmo:


bcoca:
json should be available from utils import, no need to do it again here

resmo:
this is what I thought, but json.load failed without this import. weird.

bcoca:
rechecking/retesting as I tested this when i took the json line out

bcoca:
hmm, I cannot reproduce the error on current devel, what ansible version? OS? exact error?

abadger:
json.load() is being used in both redis.py and jsonfile.py inside of get() so it should be throwing an error without a json import inside of those modules.  In addition to merging, I switched the preferred json implementation to simplejson because of the large speed boost and also restored the input to the redis caching.

https://github.com/ansible/ansible/commit/5ed7a55990e446b0f4a214e3e7228e3483390635

If you wanted to make use of the json module imported ansible.utils you'd have to make use of it similar to this:

``` python
from ansible.utils import json
json.load(f)
```

But it's not good style to do that unless you're making a choice of a default for your whole codebase.  If that's the intent, then we should move the import to a new file which is just used for these choices instead of the very cluttered utils and also make sure that all of our json usage is importing from the new module instead of import json directly.  These changes would make it obvious to others what we're doing and what the right choice to import from are.

abadger:
# Merged

Hi!

issue is resolved.

Thank you!

resmo:
@abadger thx for the explanation!

PR #9885, opened 2014-12-24
Use svg instead of png to get better image quality

PeterDaveHello:


amenonsen:
These badges are displayed when you read the `README.md` by going to `https://github.com/ansible/ansible` (which took me quite some time to figure out). Since these are tiny little text badges that say things like "pypi package 1.9.2", enhanced graphics quality doesn't seem an extraordinary priority. Summary: I think this should be closed.

PR #9892, opened 2014-12-25
Add Solaris enhancements and SmartOS fixes

xen0l:
This pull request fixes annoyances on SmartOS hosts (wrong pkg_mgr is detected, not SSH public key fingerprints are printed and distribution release/version is misisng).

I also added two functions, which print DMI and disk facts. I am using smbios utility, which is shipped with Solaris by default. For gathering disk facts, I am gathering information for sderr kstat module, which prints information such as disk vendor, disk product, serial number, size, numerous error metrics etc.

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -46,7 +46,7 @@ except ImportError:
     import simplejson as json
 
 # --------------------------------------------------------------
-# timeout function to make sure some fact gathering 
+# timeout function to make sure some fact gathering
 # steps do not exceed a time limit
 
 class TimeoutError(Exception):
@@ -320,6 +320,11 @@ class Facts(object):
                                 self.facts['distribution_version'] = data.split()[1]
                                 self.facts['distribution_release'] = ora_prefix + data
                                 break
+                            elif 'SmartOS' in data:
+                                self.facts['distribution'] = data.split()[0]
+                                self.facts['distribution_version'] = data.split()[1]
+                                self.facts['distribution_release'] = data.split()[1]
+                                break
                         elif name == 'SuSE':
                             data = get_file_content(path)
                             if 'suse' in data.lower():
@@ -387,6 +392,9 @@ class Facts(object):
             dsa_filename = '/etc/ssh_host_dsa_key.pub'
             rsa_filename = '/etc/ssh_host_rsa_key.pub'
             ecdsa_filename = '/etc/ssh_host_ecdsa_key.pub'
+        elif self.facts['distribution'] == 'SmartOS':
+            dsa_filename = '/var/ssh/ssh_host_dsa_key.pub'
+            rsa_filename = '/var/ssh/ssh_host_rsa_key.pub'
         dsa = get_file_content(dsa_filename)
         rsa = get_file_content(rsa_filename)
         ecdsa = get_file_content(ecdsa_filename)
@@ -410,6 +418,8 @@ class Facts(object):
                 self.facts['pkg_mgr'] = pkg['name']
         if self.facts['system'] == 'OpenBSD':
                 self.facts['pkg_mgr'] = 'openbsd_pkg'
+        if self.facts['distribution'] == 'SmartOS':
+                self.facts['pkg_mgr'] = 'pkgin'
 
     def get_lsb_facts(self):
         lsb_path = module.get_bin_path('lsb_release')
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -865,6 +865,7 @@ class SunOSHardware(Hardware):
     def populate(self):
         self.get_cpu_facts()
         self.get_memory_facts()
+        self.get_dmi_facts()
         return self.facts
 
     def get_cpu_facts(self):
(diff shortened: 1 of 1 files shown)


abadger:
Some notes:
- I've merged https://github.com/ansible/ansible/commit/04635333d0b423ee0e2c10a957a7063a67cd080f which should hopefully take care of the distribution/release missing portion.
- I don't have any SmartOS (or Solaris-family) boxes at the moment to test any of these changes.  If there's someone else who's running SmartOS that would care to give a quick once over to the pkg_mgr and ssh key changes, that would be great.
- Similarly, for the dmi and disk facts if someone who's running a solaris family box would verify that those changes are good/won't break on other solaris-family machines I'd be more comfortable merging them.

I'll leave some comments on the code itself inline as well.

lib/ansible/module_utils/facts.py
```diff
@@ -915,6 +927,104 @@ def get_memory_facts(self):
...
+
+        SMB_TYPE_CHASSIS = dict(
+            form_factor='Chassis Type',
+        )
+
+        for (k, v) in SMB_TYPE_BIOS.items():
+            if smbios_bin is not None:
+                (rc, out, err) = module.run_command('%s -t SMB_TYPE_BIOS' % (smbios_bin))
+                if rc == 0:
+                    smbios_dict = dict((line.split(':')[0].strip(), line.split(':')[1].strip()) for line in out.splitlines() if ':' in line )
```
abadger, line 953: Does the check module.run_command() through the transformation into smbios_dict have to be inside of the loop?  It seems liks we could do that outside of the loop to be much more efficient.

lib/ansible/module_utils/facts.py
```diff
@@ -915,6 +927,104 @@ def get_memory_facts(self):
...
+                    thisvalue = smbios_dict[v]
+                    try:
+                        json.dumps(thisvalue)
+                    except UnicodeDecodeError:
+                        thisvalue = "NA"
```
abadger, line 958: I see that other get_dmi_decode() methods are using this pattern as well but it seems very inefficient as well.  json.dumps() is not cheap and here we're just throwing away the value.  Perhaps it would be better to do something like:

``` python
if isinstance(str, thisvalue):
    thisvalue = unicode(thisvalue, encoding='utf-8', errors='replace').encode('utf-8')
```

xen0l:
I have tested these changes on even more machines and I found some issues.

My notes:
- I found a bug in get_dmi_facts, which fails on broken HW system, which don't report all DMI variables.
- Problem with https://github.com/ansible/ansible/commit/04635333d0b423ee0e2c10a957a7063a67cd080f and with my change is that we expect SmartOS, OpenIndiana, OmniOS and NexentaStor to be a Solaris, which is pretty false assumption. They are based on illumos, which is a fork of Solaris, but newer Solaris features are not compatible with it. As there are problems determining if we are running on illumos ('uname -a' reports SunOS, other platform-detection will also detect is as Solaris), I'd propose to hardcode OS type to illumos. 
- My changeset, which detects SmartOS hosts distribution can be dropped because https://github.com/ansible/ansible/commit/04635333d0b423ee0e2c10a957a7063a67cd080f is a better approach.
- SmartOS is using pkgsrc, so it's enough to set it here. Problem with autodection was that it reported SmartOS as "macports", because pkgin was distributing other commands such as port, yum, apt-get etc. Once run, those commands outputed that one should use pkgin instead. It's seems that the newer pkgin packages stopped to distribute those commands, but there still might be dozens of older SmartOS hosts with older packages. So hardcoding pkgin is needed here.

Expect fixes for these bugs some time this week.

lib/ansible/module_utils/facts.py
xen0l, line 958: This code was taken from LinuxHardware and adapted to Solaris. If we fix this, we should fix everywhere.

lib/ansible/module_utils/facts.py
xen0l, line 953: Nope, this can be run before the loop. But if we do so, then I need to fix the condition so if the 'smbios' binary is not found then it won't return anything. smbios is Solaris 10+ and it could break compatibility will older releases (I don't have access to them, so I can't test).

lib/ansible/module_utils/facts.py
```diff
@@ -915,6 +927,104 @@ def get_memory_facts(self):
...
+                      "Hard Errors", "Soft Errors", "Transport Errors", "Media Error",
+                      "Predictive Failure Analysis", "Illegal Request",
+                     ]
+
+        d = {}
+        rc, out, err = module.run_command("/usr/bin/kstat -C sderr")
+        if rc != 0:
+            return dict()
+
+        sd_instances = set(sorted([line.split(':')[1] for line in out.split('\n') if line.startswith('sderr')]))
```
abadger, line 1013: Is there a reason to sort this?  Since we're just putting it into a set (to uniquify the list I assume) it doesn't seem necessary to sort.  Since we're just iterating over sd_instances, a frozenset is more appropriate than a set.

xen0l, line 1013: Sadly, I don't remember why I put it there. I guess that I had some troubles with duplicated fields and used set to remove them. Sorted was put there, so it was printed in the sorted order. I didn't realize that set will do the same. I will test with frozenset and change it.

lib/ansible/module_utils/facts.py
```diff
@@ -915,6 +927,104 @@ def get_memory_facts(self):
...
+                else:
+                    self.facts[k] = 'NA'
+            else:
+                self.facts[k] = 'NA'
+
+        for (k, v) in SMB_TYPE_SYSTEM.items():
+            if smbios_bin is not None:
+                (rc, out, err) = module.run_command('%s -t SMB_TYPE_SYSTEM' % (smbios_bin))
+                if rc == 0:
+                    smbios_dict = dict((line.split(':')[0].strip(), line.split(':')[1].strip()) for line in out.splitlines() if ':' in line )
```
abadger, line 970: If the values can contain a colon we'll want to change the second line.split() so that we don't truncate the value.  Something like this:

``` python
smbios_dict = dict((line.split(':')[0].strip(), line.split(':', 1)[1].strip()) for line in out.splitlines() if ':' in line )
```

lib/ansible/module_utils/facts.py
abadger, line 958: <nod>  I'd be happy to see fixing it everywhere added here if you want but don't demand it :-)

PR #9902, opened 2014-12-27
fixed facts gathering for strict remote umasks

lpirl:
**Changes:**
Instead of calling `chmod` to change the mode for the directory _after_ its creation, the umask is temporarily altered _before_ calling `mkdir` to ensure the parent directories that are possibly created have the desired mode too.

**Detailed explanation:**

If the remote umask is very strict (e.g. 077) the facts gathering fails for non-root users with the following message:

```
/usr/bin/python: can't open file '/tmp/ansible/root/ansible-tmp-…/setup': [Errno 13] Permission denied
```

The reason is the following:
In `lib/ansible/runner/shell_plugins/sh.py` a temporary directory is created with `mkdir -p`.
Subsequently, the desired mode is set with `chmod`.
Possibly created parent directories do not get `chmod`ed and the non-root user won't be able to execute the setup module.

### DIFF
--- a/v1/ansible/runner/__init__.py
+++ b/v1/ansible/runner/__init__.py
@@ -1282,7 +1282,7 @@ class Runner(object):
 
         tmp_mode = None
         if self.remote_user != 'root' or (self.become and self.become_user != 'root'):
-            tmp_mode = 'a+rx'
+            tmp_mode = 0755
 
         cmd = conn.shell.mkdtemp(basefile, use_system_tmp, tmp_mode)
         result = self._low_level_exec_command(conn, cmd, None, sudoable=False)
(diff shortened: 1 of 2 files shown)


lpirl:
And for the sake…

**Type:** Bugfix Pull Request

**Ansible Version:** ansible 1.7.2

**Environment:** unix-like

**Summary:** facts gathering crashes with strict remote umasks

**Reproduce:**
- change remote umask to 077
- reboot remote machine
- run a Playbook using su (and probably sudo too), including facts gathering

**Expected Results:** facts gathering does not crash with "permission denied"

**Actual Results:**

Facts gathering crashes with `/usr/bin/python: can't open file '/tmp/ansible/root/ansible-tmp-…/setup': [Errno 13] Permission denied
`

amenonsen:
@lpirl do you feel like rebasing (and squashing) your commits onto devel? I would suggest including some of your explanation in the commit message too.

lpirl:
Thanks for considering this change! I did the rebase.

bcoca:
so the v1 tree is dead code and we don't take PRs to update it

we would consider it if you resubmit a PR vs the actual devel code

lpirl:
Ups sorry. Filed as #11764.

PR #9912, opened 2014-12-29
Added steps for installing Ansible from source

jfrappier:


bcoca:
the original is meant to stay generic and assumes you know how to install packages in your own distro, this is a bit too specific to what you were using at the time.

jfrappier:
Agree to a point, wasn't quite clear where to find sshpass. Also, you show how to install other packages such as jinja2. I can reword if you like - maybe a more specific prereq section is needed? For example you need Python, but I can install Python without python-setuptools. Maybe list the specific packages needed and if they were available via yum/apt-get or rpm?

docsite/rst/intro_installation.rst
```diff
@@ -103,7 +103,15 @@ when they are implemented, and also easily contribute to the project. Because th
 nothing to install, following the development version is significantly easier than most
 open source projects.
 
-To install from source.
+Prior to installing Ansible, there are a few packages that need to be installed. Replace yum with apt-get as needed. First, if you do not have git installed, do so now::
+
+	$ yum install git
+
+Next, you need Phyton and Python Setup Tools::
```
rnelson0: Typo "Phyton"

docsite/rst/intro_installation.rst
```diff
@@ -103,7 +103,15 @@ when they are implemented, and also easily contribute to the project. Because th
 nothing to install, following the development version is significantly easier than most
 open source projects.
 
-To install from source.
+Prior to installing Ansible, there are a few packages that need to be installed. Replace yum with apt-get as needed. First, if you do not have git installed, do so now::
```
rnelson0: Perhaps just list the software needed. "Ensure that git, python, and python-setuptools are installed."

docsite/rst/intro_installation.rst
```diff
@@ -138,7 +146,19 @@ other than /etc/ansible/hosts:
 
 You can read more about the inventory file in later parts of the manual.
 
-Now let's test things with a ping command:
+Now let's test things with a ping command. In order to perform this test you will need sshPass. If wget is not installed, install it now::
```
rnelson0: A link to sshpass's package listings should be sufficient. rpm can install from URL, apt-get cannot, but the user can figure that part out.

amenonsen:
This PR is a bit difficult to review because of the various generations of changes included. Anyway, it's accumulated merge conflicts, and you need sshpass only if you aren't using public key authentication. This should probably be closed now, and we can revisit the subject in a new PR if someone feels the need to update the instructions.

Issue #9712, opened 2014-12-04
Dynamic Inventory should have implicit localhost and/or Empty inventory files shouldn't be an error


follower:
##### Issue Type:

Bug Report / Feature Idea
##### Ansible Version:

ansible 1.8.1
##### Environment:

N/A
##### Summary:

Since the introduction of implicit localhost with `local_action` it is legitimate to want to run a playbook with an empty inventory file.

At present an empty inventory file or an empty dynamically generated inventory (e.g. from `ec2.py` in the situation where no EC2 instances exist/running) triggers an error--even if the only plays are for `local_action`.

The "accepted" workaround when not using dynamic inventory seems to be to supply `-i 'localhost,'` but this is apparently not recommended practice. Additionally, this also requires the use of `ansible_connection=local` to work around an "issue" with `synchronize` trying to log onto `localhost` via ssh.

It appears a "better" workaround is to supply a completely invalid IP in the inventory list as this case is handled explicitly (via https://github.com/ansible/ansible/issues/5820, https://github.com/ansible/ansible/commit/f955d9db6a01c6d5c2cd761b3bdfd76ae20de86b and https://github.com/ansible/ansible/blob/devel/lib/ansible/inventory/__init__.py#L360) and does not require a workaround for `synchronize`. e.g.

```
ansible-playbook  -i '999.999.999.999,' playbook.yml
```

There appears to be no workaround to this situation when using dynamic inventory and even if an explicit `localhost` entry was added to the inventory it would complicate the use of the `all` parameter for `hosts:`.

It seems like with the addition of implicit localhost, the commit https://github.com/ansible/ansible/pull/1628/files (see https://github.com/ansible/ansible/blob/devel/bin/ansible-playbook#L162) should now either be reversed and/or have the same special implicit localhost handling that the "undocumented" approach gained with https://github.com/ansible/ansible/commit/f955d9db6a01c6d5c2cd761b3bdfd76ae20de86b.

It is possible that in some situations it would still be appropriate to issue a warning in the case of an empty inventory (e.g. if there are no `local_actions` in the playbook).

As it stands both these situations are a hinderance to "green fields" deployment situations which as a "first use" situation seems like an important thing to get right. (Equally if for some reason all your VMs have been terminated...by a bad `exact_count` bug for example. :) )

The issue specifically in relation to dynamic inventory is mentioned in https://github.com/ansible/ansible/issues/6442 but this seems to have been buried.

Thanks!
##### Steps To Reproduce:
[...]

Issue #6442, opened 2014-03-12
Implicit localhost does not work with this simple example.


HelloGrayson:
##### Issue Type:

Bug  Report
##### Ansible Version:

ansible 1.5.2
##### Environment:

OSX 10.9.1
##### Summary:

Playbook errors out on this simple implicit localhost example.
##### Steps To Reproduce:

Create a playbook called `test.yml` with these contents:

```

---
- name: implicit localhost test...
  hosts: localhost
  tasks:
  - debug: msg="hello world"
```

Then run the file like so:

```
ansible-playbook test.yml
```
##### Expected Results:

```
TASK: [debug msg="hello world"] ***********************************************
ok: [localhost] => {
    "msg": "hello world"
}
```
##### Actual Results:

```
ERROR: provided hosts list is empty
```

mpdehaan:
This works for me on the devel branch.

Can you share what is in your host file (/etc/ansible/hosts unless specified by an environment variable?)

HelloGrayson:
`/etc/ansible/hosts` is empty. here is my ansible.cfg

```
[defaults]
remote_user = ubuntu
hostfile = ./inventory/ec2.py
host_key_checking = False

[ssh_connection]
pipelining = True
```

HelloGrayson:
@mpdehaan were you able to reproduce using my `ansible.cfg`?

HelloGrayson:
Just verified that this still does not work with 1.5.3

mpdehaan:
Empty or missing?

HelloGrayson:
@mpdehaan `/etc/ansible/hosts` does not exist. I installed ansible with pip on OSX.

vadimkuznetsov:
If the hosts is empty, i.e. size 0 or  nova.py --list returns {}, then ansible-playbook fails with 
ERROR: provided hosts list is empty.
if hosts file has even garbage line, for example 300.300.300.300, then implicit localhost is present.

jorgebastida:
Similar situation running a simple playbook which creates a vpc on aws using the `ec2_vpc` module. It works fine until I terminate the only server running on this account. I get `ERROR: provided hosts list is empty` thereafter. Booting a new empty server solves the problem.

ScottSturdivant:
[...]

PR #9913, opened 2014-12-30
Removed early check for empty inventory

dedalusj:
The `ansible-playbook` script throws an error for empty inventory files before creating a playbook object and running its plays.

This is sometimes unwanted since some playbooks may contain only action to be performed on localhost (which is implicit), see #6442 and #9712 .

### DIFF
--- a/bin/ansible-playbook
+++ b/bin/ansible-playbook
@@ -167,6 +167,8 @@ def main(args):
 
     inventory = ansible.inventory.Inventory(options.inventory, vault_password=vault_pass)
     inventory.subset(options.subset)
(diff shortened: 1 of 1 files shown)


gmalouf:
Any timeline on getting this merged?  The current bug causes a lot of pain for me personally.

bcoca:
so I am reluctant to remove this completely as it will silently accept empty inventories for people that don't intend that, changing it from an error to a warning might be a better solution.

dedalusj:
That's fair. I converted the error message to a warning. I hope I chose the right way to show the warning.

PR #9923, opened 2014-12-31
Correct URL to github so links work when testing locally

gundalow:
Previously if you did "make webdocs" and we would link to 
http://localhost/ansible-dev/github.com/ansible/ansible-modules-core


amenonsen:
Already fixed, this PR can be closed.

PR #9924, opened 2015-01-01
Strip formatting from Module lists

gundalow:
http://docs.ansible.com/list_of_all_modules.html

**BEFORE**
dnf (E) - Manages packages with the I(dnf) package manager

**AFTER**
dnf (E) - Manages packages with the dnf package manager


amenonsen:
OK, so (a) the problem still exists, as can be seen on the page listing all modules, (b) the patch still applies, even after 6+ months, and (c) it even looks reasonable (except I don't really understand why it calls `cgi.escape()`; that seems to be new). Merge ahoy?

bcoca:
so we do one better already in the  def rst_ify(text): function we change it to be rst style markup.

so I applied it to the category description in the list pages, it was already applied in the module doc page. Closing this via e5d618a254dd6c994f4047e068e5a6892f1f8cf9

PR #9927, opened 2015-01-02
Retry exec command via ssh

hughsaunders:
##### Summary:

This PR adds the option to retry failed ssh executions, if the failure is caused by ssh itself, not the remote command. This can be helpful if there are transient network issues. Retries are only implemented in the openssh connection plugin and are disabled by default. Retries are enabled by setting ssh_connection > exec_retry to an integer greater than 1.
##### Steps To Reproduce:

Running a long series of playbooks, or a short playbook against a large cluster may result in transient ssh failures, some examples logged [here](https://trello.com/c/1yh6csEQ/13-ssh-errors).
##### Expected Results:

Ansible should be able to retry an ssh connection in order to survive transient failures.
##### Actual Results:

Ansible marks a host as failed the first time it fails to contact it.


sigmavirus24:
This looks good to me and seems to work as expected. I'd love to have this in ansible.

RubenKelevra:
This patch is GREAT! I often had this issues, thanks!

jimi-c:
Cleanly cherry-picked and applied to v2_final.

jimi-c:
As this has been applied to the v2 branch, I'm going to go ahead and close it now.

Thanks!

PR #9928, opened 2015-01-02
make a mirror clone rather than a regular clone

mscherer:
This permit to give a git branch as a version for ansible-galaxy,
which in turn open interesting way of managing the roles in a seamless
fashion.


amenonsen:
I think this should probably be closed, since #10620 is a more complete approach, and @semifocused is working on rebasing it to apply to devel already.

mscherer:
Yup, indeed

Issue #7106, opened 2014-04-21
Jinja2 doesn't allow to use template includes outside playbook dir


pfalcon:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.6 (devel f1ce9abe0d) last updated 2014/04/21 21:05:18 (GMT +300)
##### Environment:

Ubuntu 12.04
##### Summary:

We have our playbook's data (group/host vars, templates, etc.) split in 2 parts: public, which goes alongside playbook, and private, which is held in completely different repo. Both repos are checkout out into a common parent dir, so from public dir, private can be accessed as ../ansible-secrets-jenkins/ . That works for accessing most data, except for templates - Jinja2 by default doesn't want to access templates with ".." in path.
##### Steps To Reproduce:

In this specific case, we have:

```
secret_dir: ../ansible-secrets-jenkins/
jenkins_ssh_publishers_file: "{{secrets_dir}}/jenkins_ssh_publishers.xml.tcwg-ci"
```

and within a template:

```
{% if jenkins_ssh_publishers_file is defined %}
{% include jenkins_ssh_publishers_file %}
{% endif %}
```
##### Expected Results:

I'd expect to be able to use template located outside playbook dir,
just as possible for var files, copy file, etc.
##### Actual Results:

With #7105 applied:

```
fatal: [tcwg.ci.linaro.org] => {'msg': 'TemplateNotFound: ../../ansible-secrets-jenkins/jenkins_ssh_publishers.xml.tcwg-ci', 'failed': True}
```

That's because:

jinja2/loaders.py:

```
def split_template_path(template):
    """Split a path into segments and perform a sanity check.  If it detects
    '..' in the path it will raise a `TemplateNotFound` error.
    """
```

mpdehaan:
Note this is just about template includes, not templates in general - I've edited the description to reflect this.

As I've mentioned before, template includes in Ansible aren't really a thing presently and need some hammering to be made to work.   (I agree it would be nice -- but not too many people use them).

pfalcon:
Ah, thanks for correction - indeed, I didn't try to reference direct template form outside a playbook dir, that indeed works. Well, per our usecase, really need to keep parts of template separate, and including subparts seems like most straightforward way. Currently working the issue around by having a symlink within playbook dir pointing to outside dir.

rsyring:
I could really use them.  There is a discussion about this here:

[...]

PR #9933, opened 2015-01-03
add role directory to template lookup locations #7106

rsyring:
This addresses:  Jinja2 doesn't allow to use template includes outside playbook dir #7106

### DIFF
--- a/lib/ansible/utils/template.py
+++ b/lib/ansible/utils/template.py
@@ -210,7 +210,11 @@ def template_from_file(basedir, path, vars, vault_password=None):
 
     from ansible import utils
     realpath = utils.path_dwim(basedir, path)
-    loader=jinja2.FileSystemLoader([basedir,os.path.dirname(realpath)])
+    template_lookup_paths = [basedir, os.path.dirname(realpath)]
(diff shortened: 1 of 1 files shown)


rsyring:
Michael DeHaan said this is not likely to get accepted due to a preference for changes to be against the v2 codebase.  For future reference:

> v2 is not open for submissions just yet, plugins are in progress -- we anticipate this should be ready in about a month or so for helping testing and development.   There will be an analogous file in the action_plugins tree, called template.py, and be pretty similar, just a bit nicer.

dagwieers:
Thanks. This is something I would like to see improved as well.

bcoca:
v2 will make it into main devel soon after 1.9 is released, this will probably make it in then

amenonsen:
@rsyring note that the current devel branch is the v2 codebase. I _think_ this change is still relevant. Any interest in rebasing it to apply to devel?

jimi-c:
I've merged in the above commit, which should address this in the current devel branch.

Thanks!

rsyring:
@jimi-c thanks!

PR #9938, opened 2015-01-05
Improve docs in regards to custom module creation

Meai1:
It's necessary to give people instructions on how to use all these praised features, like custom language modules.


amenonsen:
I agree with the premise of this PR, but the wording needs to be tightened up. It's not OK to say things like "the most reliable way seems to be" in Ansible documentation.

PR #9949, opened 2015-01-07
Expand ~ in filename given as option for ansible and ansible-playbook

### LINKED ISSUE #9937: Cant use ~ in --module-path
Works:
ansible-playbook --ask-pass ~/myplaybook.txt --module-path=/home/pc/ansible_library

Does not work:
ansible-playbook --ask-pass ~/myplaybook.txt --module-path=~/ansible_library

So in my opinion this doesnt make sense because I'm launching this in a terminal as "pc" user, so there should theoretically be no confusion as to which computer is meant and which user is meant by the --module-path.

mscherer:
Fix https://github.com/ansible/ansible/issues/9937

### DIFF
--- a/lib/ansible/utils/__init__.py
+++ b/lib/ansible/utils/__init__.py
@@ -1042,6 +1042,9 @@ def increment_debug(option, opt, value, parser):
     global VERBOSITY
     VERBOSITY += 1
 
+def expand_tilde(option, opt, value, parser):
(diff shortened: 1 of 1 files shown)


jimi-c:
Hi @mscherer, I've tweaked this patch and merged it in with some fixes as noted above.

Thanks!

PR #9950, opened 2015-01-07
Fix various inventory scripts to use env python

### LINKED ISSUE #9948: gce.py uses #!/usr/bin/python, causes errors for virtual environments
The gce.py script uses the os install /usr/bin/python executable instead of the executable located by /usr/bin/env.  This causes problems for users who run virtual environments because they will likely not have apache-libcloud installed.

mscherer:
Fix https://github.com/ansible/ansible/issues/9948


mpdehaan:
Merged, thank you!

PR #9954, opened 2015-01-08
Set ansible_distribution_version differently for Solaris variants.

natefoo:
Just a small improvement to the Solaris distribution fact changes that were just made by @abadger. And fix a bug that was introduced with those.


abadger:
Thanks for the improvements, the raw data, _and_ fixing my mistakes ;-)

PR #9959, opened 2015-01-08
Added before & after module info to error message

grahamrhay:
It's not always very obvious what you've done wrong when you get the error below, having the before & after lists of module args makes it a lot easier.


amenonsen:
This PR can be closed now (the underlying code has changed completely, v2 no longer even has this error).

bcoca:
error messages in v2 are a lot more informative now, closing as per above

PR #9974, opened 2015-01-10
Naming item variable in item lookups

shawnsi:
I regularly find myself looping over registered variables in chains.  This results in variable names like `item.item.item...` in my playbooks.  The attached patch allows naming the `item` variable within an item lookup for clarity.

See the attached example.

### DIFF
--- a/lib/ansible/plugins/filter/core.py
+++ b/lib/ansible/plugins/filter/core.py
@@ -34,6 +34,7 @@ from functools import partial
 import operator as py_operator
 from random import SystemRandom, shuffle
 import uuid
+from collections import defaultdict
 
 import yaml
 from jinja2.filters import environmentfilter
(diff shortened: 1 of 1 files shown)


shawnsi:
I opted to copy the item key, rather than moving, to limit any collateral damage with this change.  This was my first foray into the ansible code so this was the quickest and safest way to communicate the feature addition.

omgjlk:
I'd like to see an example which leads to item.item.item chains to really understand what problem this is trying to fix.

shawnsi:
Here is an example I cherry picked.  I'm fully aware the docker module exists but I'm calling the CLI directly for the sake of showing a meaningful chain of variable registrations here.  I've also found some limitations with the docker module in real world cases that have forced me back to the CLI so this isn't a superficial need either.

``` yaml
- hosts: localhost
  connection: local
  vars:
    ansible_python_interpreter: python2
    services:
      - image: nginx
        name: web1
      - image: nginx
        name: web2
  gather_facts: no
  tasks:
  - name: Timestamp
    command: /usr/bin/date +%s
    register: timestamp

  - name: Start containers
    shell: /usr/bin/docker run -p 80 --name {{ service.name }}_{{ timestamp.stdout }} -d {{ service.image }}
    item_name: service
    register: containers
    with_items: services

  - name: Discover exposed ports
    command: /usr/bin/docker port {{ container.stdout }} 80/tcp
    register: ports
    item_name: container
    with_items: containers.results

  - name: Publish services (without item naming)
    debug: msg="Service {{ item.item.item.name }} is published on port {{ item.stdout }}"
    with_items: ports.results

  - name: Publish services (with item naming)
    debug: msg="Service {{ port.container.service.name }} is published on port {{ port.stdout }}"
    item_name: port
    with_items: ports.results
```

apotek:
A. +1 for being a minimally invasive approach to fixing the issue and allowing complex references to registered results variables to be more easily read.  However, we are also, if I'm understanding the code correctly, adding some memory overhead. Looking at the implementation, it does seem like that memory is only being gobbled when item_name: [name] is set, so that mollifies my concern here. I will only be using that extra memory when I have explicitly asked that I need the readability of using item names, by setting the item_name property.

B.

>  I've also found some limitations with the docker module in real world cases that have forced me back to the CLI

If the situations that have led to item.item.item... chains are all localizable to work you have been doing with docker, I would suggest the fix/improvement should be to the docker module, if possible. This patch and the sample indicate a much higher level of "code-yness" than most people would want to encapsulate in ansible playbooks and roles. So the preferred solution would be to make the docker module more capable.

On the other hand, if this pattern is generalized among many modules, this is probably the best approach to fixing it. I certainly can't think of a better way. Nice work.

Thanks!

amenonsen:
The proliferation of things named `item` has troubled me before (e.g. when using `with_together`, it's not easy to review which item is which), so I'm receptive to this idea. But the PR is outdated, doesn't include any documentation, and the commits incorrectly include submodule pointer updates. Those definitely need to go.

I think this PR can be closed, but I would like the core team's opinion on implementing this feature in 2.x. Is there a better/safer/more comprehensive way to do this?

shawnsi:
I'm closing this PR as I've dropped the work from devel branch and reset to upstream.

PR #9975, opened 2015-01-11
Get rid of bash compound command to make more portable

abadger:
The compound command [[ is available on many modern shells as an enhanced substitute for /bin/test.  But it is not part of the posix standard and not available on all posix compliant shells.  Implement in a different manner to enhance portability.


abadger:
Okay, there were some questions of whether there are gotchas in the portability of expr.  I took a look at https://www.gnu.org/software/autoconf/manual/autoconf-2.65/html_node/Limitations-of-Usual-Tools.html and went through the section on expr use.  I think that the limited amount of expr functionality that we're using is safe.  If env-setup starts failing on other platforms we can revert this change.

PR #9989, opened 2015-01-13
Apply template to task (handler) name on initialization.

ashorin:
Apply template to task (handler) name on initialization. Saves lots of time when having tens of handlers. It saves hundreds of calls to template().


amenonsen:
This can be closed, the underlying code has changed completely.

bcoca:
this is already the case in current devel, closing this ticket

PR #9993, opened 2015-01-13
Use current directory in preference to all other paths when looking for host/group vars.

patch0:
This allows playbook reuse with different variables in the current
directory structure.

e.g.

  ansible-playbook -i ~/shared/inventories ~/shared/playbooks/authenticate.yml

with vars in the current working directory in group_vars/ and
host_vars/.

### DIFF
--- a/lib/ansible/inventory/__init__.py
+++ b/lib/ansible/inventory/__init__.py
@@ -607,8 +607,9 @@ class Inventory(object):
     def _get_hostgroup_vars(self, host=None, group=None, new_pb_basedir=False):
         """
         Loads variables from group_vars/<groupname> and host_vars/<hostname> in directories parallel
-        to the inventory base directory or in the same directory as the playbook.  Variables in the playbook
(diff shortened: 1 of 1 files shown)


abadger:
I'm not sure cwd is a good idea.  It could be a security risk to load different values depending on the current directory you're in when you run ansible.

patch0:
The benefit is being able to apply the same playbook many times with different variables set, depending on which directory you're in.  In our usage case we have many clients, each with their own ansible tree, for whom we want to run a single "authenticate" playbook on, with different users/passwords/keys being added depending on which client tree you're in.

I think the security risk would be very small.

Actually, thinking about it, maybe this is just better done with tags and roles.

bcoca:
this can also be done with inventory vars (I imagine the clients have different inventories).

closing this ticket as I don't think this is something we want to implement, it would create a lot of confusion if you for example run ansible from the inventory dir and override the play group_vars when it normally works the other way around.

PR #10022, opened 2015-01-16
New Jinja2 'resolve' filter.

DenKoren:
Translates list of domain names into list of IP addresses.
You can select IPs by address family.
By default filter lists IPv4 and IPv6 addresses.

Usage (in Jinja2 template): 
- Get list of IPv4 addresses for list of domains:
  - `{{ domains | resolve('IPv4') }}`
  - `{{ domains | resolve('v4') }}`
  - `{{ domains | resolver(4) }}`
- Get list of IPv6 addresses for list of domains:
  - `{{ domains | resolve('IPv6') }}`
  - `{{ domains | resolve('v6') }}`
  - `{{ domains | resolve(6) }}`

### DIFF
--- a/lib/ansible/runner/filter_plugins/core.py
+++ b/lib/ansible/runner/filter_plugins/core.py
@@ -19,6 +19,7 @@ import sys
 import base64
 import json
 import os.path
+import socket
 import types
 import pipes
 import glob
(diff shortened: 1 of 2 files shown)


bcoca:
so this makes more sense as a lookup plugin, filters are for altering data and are not expected to make external/network connections. There is already a dnstxt lookup plugin, this seems to be a nice companion to that.

bcoca:
closing in favor of #10311 which implements the dig lookup plugin

PR #10024, opened 2015-01-16
Handle CP437 encoding returned from pywinrm.

cchurch:
Per https://github.com/diyan/pywinrm/blob/master/winrm/protocol.py#L62, pywinrm opens a shell with WINRS_CODEPAGE=437.  The stdout/stderr in the response we get back is a Python `str` instance encoded as 'cp437', so decode that into a `unicode` instance before passing back to ansible.

Should fix the issue described in https://groups.google.com/d/topic/ansible-project/uR9kL1o_P7U

### DIFF
--- a/lib/ansible/runner/connection_plugins/winrm.py
+++ b/lib/ansible/runner/connection_plugins/winrm.py
@@ -150,7 +150,9 @@ class Connection(object):
         except Exception, e:
             traceback.print_exc()
             raise errors.AnsibleError("failed to exec cmd %s" % cmd)
-        return (result.status_code, '', result.std_out.encode('utf-8'), result.std_err.encode('utf-8'))
+        std_out = result.std_out.decode('cp437')
+        std_err = result.std_err.decode('cp437')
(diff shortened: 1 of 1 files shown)


bitmax1:
Sorry to tell you that doesn't fix the problem, at least in my environment.
Decoding it to unicode does an implizit conversion to ascii, which throws the decoding error.
I changed line 153 in 
from: 
    return (result.status_code, '', result.std_out.encode('utf-8'), result.std_err.encode('utf-8')
to:
    return (result.status_code, '', result.std_out.decode('cp437').encode('utf-8'), result.std_err.decode('cp437').encode('utf-8'))
solved the problem for me.
I'm not a python expert, but a german, so used to those superficious character above 128, so I don't know wether that is the best solution.

bitmax1:
Sorry, I'm new to github, didnt see your commit.
You solved the problem.

cchurch:
Figured out how to use UTF8 codepage instead of CP437 for #11719, so this PR is no longer relevant.

PR #10025, opened 2015-01-16
This will add the uuid of the device in ansible_mounts variable

digi691:
This will return 'NA' if the command is unsuccessful as older OSes may not have lsblk command.


bcoca:
instead of just calling lsblk, could you use get_bin_path?

digi691:
Not a problem...

digi691:
I messed up the rebase back into my fork.  Let me know if you want to close this and i'll issue another PR.

bcoca:
if you have a 'clean' local branch, you can push --force to your github branch and that should fix it

PR #10026, opened 2015-01-16
adds to v2 the ability to run with empty inventory (implicit localhost)

bcoca:
which is now a warning and not an error, this brings v2 into parity with v1 change


abadger:
+1

PR #10027, opened 2015-01-16
By default, don't hide undefined variable errors

pilou-:
Besides, when the variables below are used inside a playbook:

> dir1: "{{ nonexistent }}"
> var1: "{{ lookup('password', 'file' + dir1 + '/filename length=10') }}"

the patch allows to replace the "ERROR: Unexpected error in during lookup: list
index out of range" error message with "jinja2.exceptions.UndefinedError: 'var'
is undefined'."


bcoca:
Sadly we cannot accept this patch. We had applied the same one and released with 1.8 but this broke many playbooks as lookups are implicitly used in with_ constructs, which also has implied templating. It makes it impossible to distinguish between a literal and a variable, getting a lot of false positives on 'undefined' when they were really literals.

PR #10028, opened 2015-01-16
V2 ssl configurable

bcoca:
another patch to keep parity with v1


bcoca:
we should also look in v2 to allow modules to override default ssl as different systems will have different versions of the libraries and sadly many older systems will be incompatible with the new settings.

Though the SA in me protests and thinks they should upgrade as the SSL issues that exist in those versions are serious enough to consider the traffic as non encrypted.

bcoca:
this actually breaks modules, closing until it can be rethought

PR #10029, opened 2015-01-17
Add filter to turn a string into a UUID

gaqzi:
This filter was made because I needed to create idempotent UUIDs when
installing the agent for Go (http://go.cd), which uses UUIds to
distinguish the agents from each other.

The filter doesn't do anything special when the string is empty, would this be a good idea to add an assert for?

I hope this would be useful for inclusion in core. :)

### DIFF
--- a/docsite/rst/playbooks_variables.rst
+++ b/docsite/rst/playbooks_variables.rst
@@ -409,6 +409,10 @@ To work with Base64 encoded strings::
     {{ encoded | b64decode }}
     {{ decoded | b64encode }}
 
+To create a UUID from a string (new in version 1.9)::
+
+    {{ hostname | to_uuid }}
+
 To cast values as certain types, such as when you input a string as "True" from a vars_prompt and the system
 doesn't know it is a boolean value::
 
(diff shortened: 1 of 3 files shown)


sivel:
I'd recommend a few changes. Such as not naming the function or the filter `uuid`, but perhaps name it something more specific to its function since the uuid provided functions all do something different and the namespace you chose is specifically for FQDN hostnames.

gaqzi:
@sivel: What do you think of me doing this:

1) Create an Ansible namespace and use that (it's just a UUID anyway)
2) Rename `uuid` to `uuid_from_string` ?

bcoca:
to_uuid?

gaqzi:
`to_uuid` is a lot clearer, and to the point, thanks @bcoca. I've updated the pull request to use `to_uuid` instead. I've also created an Ansible UUID namespace that is now used instead of the DNS namespace.

docsite/rst/playbooks_variables.rst
```diff
@@ -409,6 +409,16 @@ To work with Base64 encoded strings::
     {{ encoded | b64decode }}
     {{ decoded | b64encode }}
```
bcoca, line 411: this is adding the recently remove sha1 doc

gaqzi, line 411: oops, sorry about that, unsuccessful merge rebase on my end. Fixed now!

Also noticed the format for when a filter was introduced is added in the docs, so updated to match that. :)

PR #10030, opened 2015-01-17
aks: added to_list filter & tests

aks:
Added a new filter: to_list.

This causes values to be coerced into a list, so the subsequent recipe can iterate on it -- even if it's just a single item.

A use case was the `debops.ansible_users` role, which makes use of a `forward` attribute which is one or more addresses.  The 95% use-case for `~/.forward` is a single address: the average user of this role will leave the forward attribute empty, or put a single address in it.  If that value is not a list, the `debops.ansible_users` role appears to work but it creates a broken address of a list of single-characters, separated by commas, because it's trying to create a list of addresses, separated by commas.  

I've filed an issue on that role, but in examining how to fix it in my own repo, there didn't seem to be an obvious way to coerce a scalar value into a list, and filters seemed to be the Right Way to address this, so here's a new `to_list` filter, used something like this:

```
forward = {{ item.forward | to_list }}
```

This pull modifies `filter_plugins/core.py` and `TestFilters.py`.  The tests passed.

I also added the same filter to the corresponding `core.py` in the `v2` directory, but the tests there don't seem to be as well developed, so I didn't pursue the test implementation there.

### DIFF
--- a/lib/ansible/runner/filter_plugins/core.py
+++ b/lib/ansible/runner/filter_plugins/core.py
@@ -203,6 +203,9 @@ def union(a, b):
         c = unique(a + b)
     return c
 
+def to_list(a):
+    return type(a) == type([]) and a or [a]
+
 def min(a):
     _min = __builtins__.get('min')
     return _min(a);
@@ -330,6 +333,7 @@ class FilterModule(object):
             'difference': difference,
             'symmetric_difference': symmetric_difference,
             'union': union,
+            'to_list': to_list,
             'min' : min,
             'max' : max,
 
--- a/test/units/TestFilters.py
+++ b/test/units/TestFilters.py
@@ -131,6 +131,18 @@ class TestFilters(unittest.TestCase):
                                                       'a\\1')
         assert a == 'ansible'
 
+    def test_to_list_on_string(self):
+        a = ansible.runner.filter_plugins.core.to_list('ansible')
+        assert type(a) == type([])
+        assert a[0] == 'ansible'
+
+    def test_to_list_on_list(self):
+        a = ansible.runner.filter_plugins.core.to_list(['ansible', 'more stuff'])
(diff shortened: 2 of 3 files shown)


bcoca:
how is this different than the 'list' filter?

```
forward = {{ item.forward | list }}
```

aks:
Hi Brian,

I appreciate your attention on this.

If `item.forward` is a string, then `item.forward | list` results in a list
of the characters in item.forward.

For example, if item.forward is "aks@stebbens.org", then the result is:

```
[ 'a', 'k', 's', '@', 's', 't', 'b', 'b', 'e', 'n', 's', '.', 'o', 'r',
```

'g' ]

which is not a desired result.  In fact, I can't think of a useful
application for this feature.

However, perhaps there is a use case for this, so the patch I created for
`to_list` simply coerces scalar values into a list, if they are not already
a list.

Given the previous example, the result of `item.forward | to_list` would be:

```
[ 'aks@stebbens.org' ]
```

Alan

On Thu, Feb 26, 2015 at 7:29 PM, Brian Coca notifications@github.com
wrote:

> how is this different than the 'list' filter?
> 
> forward = {{ item.forward | list }}
> 
> —
> https://github.com/ansible/ansible/pull/10030#issuecomment-76324951.

## 

Best regards,
Alan

bcoca:
ah, then you want:

```
forward: "{{ lookup('items', item.forward, wantlist=True) }}"
```

aks:
It's possible that this will work too:

```
forward: "{{ [ item.forward ] | flatten }}"
```

amenonsen:
So it looks like this PR can be closed now, since there are multiple alternative ways to accomplish this behaviour. (It certainly did surprise me that `'xyz'|list` results in `['x','y','z']` though.)

bcoca:
closing this as per above

PR #9437, opened 2014-10-28
changed get_group to run in O(n*log(n)) instead of O(n^2) by using _grou...

hrld:
...ps_cache_hash

TODO: check if cache needs to be invalidated in more places (someone who knows the code should check for that...)

ansible startup time goes down by 12 seconds with 12k groups in the inventory. Searching in a list is just slow.

See also: https://groups.google.com/forum/#!topic/ansible-devel/YB-6SRvS5MU

PR #10034, opened 2015-01-18
start-up performance boost for bigger inventories

will0:
Starting somewhere between 1k-10k hosts there is some user noticeable lag between hitting enter and output starting. After profiling (& taking) the easy wins, `--list-hosts` on a 10k host inventory has a much more satisfying response time.

This PR adds a very simplistic `bob[0000:9999]` inventory and improves the performance in the top two places called out by running `python -m cProfile -s time ./bin/ansible all -i test/units/inventory_test_data/huge_range --list-hosts`. The top profiler identified time spenders are called out with before & after in each commit.


amenonsen:
This is very similar to #9437. Both are outdated, but the basic optimisation (using a cache) still seems applicable. @will0 any interest in bringing this up to date with the current devel branch?

amenonsen:
For what it's worth, this situation still exists in devel. An inventory file containing `bob[0:9999]` takes 46s for `bin/ansible --list-hosts`. The top of the profile looks like this:

```
     104411688 function calls (104398055 primitive calls) in 46.653 seconds

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
 99990000   20.181    0.000   20.181    0.000 host.py:38(__eq__)
    20000   19.597    0.001   39.781    0.002 __init__.py:353(__append_host_to_results)
    10000    3.728    0.000    3.731    0.000 __init__.py:410(_get_host)
```

With this patch applied, it takes less than half that time and only `_get_host` shows up in the profile:

```
     4431688 function calls (4418055 primitive calls) in 21.034 seconds

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
   10000   17.191    0.002   17.202    0.002 __init__.py:410(_get_host)
```

It's a small, self-contained change that made sense in January and still makes sense now. (Note that the similarity to #9437 noted above is superficial, it addresses a different problem in the same area.)

amenonsen:
And the merge conflict was trivially resolved by the three-way merge, so I've gone ahead and resubmitted this as #12001 with attribution retained.

bcoca:
closing vi rebased #12001 which kept attribution

will0:
@amenonsen @bcoca thanks! Just got back from a long trip and this gave me an extra smile.

PR #10040, opened 2015-01-18
POSIX shell scripting improvements and bug corrections

Maroloccio:


abadger:
Made comments on each commit since there's things here we definitely don't want and a few bugs that need to be fixed.

In the future, separate pull requests are definitely better -- at least, separate the bug fixes from the style changes.

Since I've gone through them this time, do you want me to cherry-pick the commits we want like with your last pull request?

Maroloccio:
You can cherry-pick, no problem with that. Please consider my code examples in the answers to your comments and see on which technical grounds I made the changes. Can you, for example, prove a case in which `dirname "$0"` is correct? I can't..

bcoca:
dirname $0 was used when people executed it as a script and not sourced it

Maroloccio:
So, Brian, you agree that it serves no purpose now. It is vestigial. If a process is executed, the fork/exec paradigm allows setenv() to only alter the child process' space, not the parent's. Let's merge this in.

abadger:
Cherry picked.  If you want to submit a new PR that keeps just a single instance of the ansible paths in the path variables I'll look at it.  Just be careful of what I was saying about building up too much complexity in a script that is at its heart, provided for convenience.

PR #10041, opened 2015-01-18
Add server mode

skorokithakis:
Thus is a preliminary commit, for review, of an implementation of server mode. This spawns an HTTP server and runs the playbook whenever a specified URL is hit.

Please don't merge yet, there's some cleanup that needs doing, as well as the addition of a separate thread. Please do provide feedback, though.

### DIFF
--- a/bin/ansible-pull
+++ b/bin/ansible-pull
@@ -193,9 +193,10 @@ def run(parser, options, args):
 
 def start_server(interface, port, url, parser, options, args):
     from BaseHTTPServer import BaseHTTPRequestHandler, HTTPServer
+    import threading
 
     class PullHandler(BaseHTTPRequestHandler):
-        last_pull_time = 0
+        pull_in_progress = False
 
         def _constant_time_compare(self, val1, val2):
             """
@@ -210,8 +211,9 @@ def start_server(interface, port, url, parser, options, args):
             return result == 0
 
         def _pull(self):
-            PullHandler.last_pull_time = time.time()
+            PullHandler.pull_in_progress = True
             run(parser, options, args)
+            PullHandler.pull_in_progress = False
 
         def do_GET(self):
             if self._constant_time_compare(self.path, url):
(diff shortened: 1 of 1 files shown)


skorokithakis:
This should be good to merge now.

skorokithakis:
Bump?

skorokithakis:
I've been using this on production and it works well, by the way.

colthreepv:
ansible devs does not seem interested in that or the issue, but... +1 :+1:

skorokithakis:
That's too bad, it's been a lifesaver for me. Literally every source code tool has webhooks, and this stops Ansible from polling all day.

jimi-c:
Hi! Thanks very much for your submission to Ansible.  It sincerely means a lot to us that you've taken time to contribute.

Unfortunately, we're not sure if we want this feature in the program. Without trying to be overly confrontational, our primary reason for this is the lack of security in this mode of operation. Since this would most likely be running as root, or as a user with privilege escalation capability for most users, the potential for abuse is way too high. Beyond that, as you note it is quite easy to add in the capability to execute a hook action with source control, which could be purposed to run Ansible from the system hosting the source code repository. We believe this would be a better solution, which would continue to leverage the security of the transport modes we use in Ansible today.

skorokithakis:
You are, indeed, correct. I've spend a few minutes refactoring the code into its own package, which I've called "Captain Webhook":

https://pypi.python.org/pypi/captainwebhook

Instructions are:

```
pip install captainwebhook
cptwebhook -k SomeLongRandomKey "ansible-pull yourrepo"
```

And it does pretty much exactly the same thing as this PR, only it can be generalized to any task.

bcoca:
i was using tcpserver to do the same thing (part of ucspi)

## 

Brian Coca

skorokithakis:
@bcoca, that definitely works, but it runs the program on any connection, so it doesn't have authentication, or returning a valid HTTP response, or not running the command while the previous one is still running (unless I'm mistaken).

bcoca:
all that can be easily layered on top, i did not use http but a simple
key/value protocol on the ucscpi-ssl version, it is normally a lot less
overhead than http.

## 

Brian Coca

Issue #9888, opened 2014-12-24
`groups['groupname']|length` in group_vars evaluates incorrectly in some contexts


hfinucane:
Given this inventory:

```
# inv
[group]
localhost
```

And this `group_vars/all`:

```

---
intermediate_value: "{{ groups['group']|length }}"
shard_size: "{{ 256 // intermediate_value|int }}"
```

I would expect this to happen (and it does!)

```
$ ansible all -i inv -m debug -a var=shard_size
localhost | success >> {
    "var": {
        "shard_size": "256"
    }
}
```

However, if I do this (which, does not really involve templating):

```
ansible all -i inv -m copy -a "content=foo dest=/tmp/bar"
```

I get this beauty:

```
localhost | FAILED => Traceback (most recent call last):
  File "/home/zzz/src/ansible/lib/ansible/runner/__init__.py", line 590, in _executor
    exec_rc = self._executor_internal(host, new_stdin)
  File "/home/zzz/src/ansible/lib/ansible/runner/__init__.py", line 792, in _executor_internal
    return self._executor_internal_inner(host, self.module_name, self.module_args, inject, port, complex_args=complex_args)
  File "/home/zzz/src/ansible/lib/ansible/runner/__init__.py", line 1025, in _executor_internal_inner
    result = handler.run(conn, tmp, module_name, module_args, inject, complex_args)
  File "/home/zzz/src/ansible/lib/ansible/runner/action_plugins/copy.py", line 181, in run
    remote_checksum = self.runner._remote_checksum(conn, tmp_path, dest_file, inject)
  File "/home/zzz/src/ansible/lib/ansible/runner/__init__.py", line 1248, in _remote_checksum
    python_interp = inject['hostvars'][host].get('ansible_python_interpreter', 'python')
  File "/home/zzz/src/ansible/lib/ansible/runner/__init__.py", line 105, in __getitem__
    self.lookup[host] = template.template('.', result, self.vars_cache)
  File "/home/zzz/src/ansible/lib/ansible/utils/template.py", line 132, in template
    d[k] = template(basedir, v, templatevars, lookup_fatal, depth, expand_lists, convert_bare, fail_on_undefined, filter_fatal)
  File "/home/zzz/src/ansible/lib/ansible/utils/template.py", line 118, in template
    varname = template_from_string(basedir, varname, templatevars, fail_on_undefined)
  File "/home/zzz/src/ansible/lib/ansible/utils/template.py", line 364, in template_from_string
[...]

PR #10046, opened 2015-01-19
Make the source of template errors visible

hfinucane:
This fixes the mysterious-traceback issue in #9888. Instead of template evaluation failure in vars files causing tracebacks like this:

```
  File "/home/zzz/src/ansible/lib/ansible/utils/template.py", line 364, in template_from_string
    res = jinja2.utils.concat(rf)
  File "<template>", line 9, in root
ZeroDivisionError: integer division or modulo by zero
```

with no hint of the source of the error, it will print out the snippet that caused the problem. Instead, you get this:

```
  File "/home/zzz/src/ansible/lib/ansible/utils/template.py", line 367, in template_from_string
    res = jinja2.utils.concat(rf)
  File "<template>", line 9, in root
ZeroDivisionError: integer division or modulo by zero

{{ 256 // intermediate_value|int }}
```

 I think this is actually fairly reasonable, given that most inline templates are pretty short.

The full path to the file would be better, but I couldn't figure out the plumbing.

I looked at v2, but it seems too green right now.

### DIFF
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -590,8 +590,10 @@ class Runner(object):
             msg = str(ae)
             self.callbacks.on_unreachable(host, msg)
             return ReturnData(host=host, comm_ok=False, result=dict(failed=True, msg=msg))
-        except Exception:
+        except Exception, e:
             msg = traceback.format_exc()
+            if hasattr(e, "_ansible_srcinfo"):
+                msg = ("%(varname)s from %(basedir)s\n" % e._ansible_srcinfo) + msg
             self.callbacks.on_unreachable(host, msg)
             return ReturnData(host=host, comm_ok=False, result=dict(failed=True, msg=msg))
 
--- a/lib/ansible/utils/template.py
+++ b/lib/ansible/utils/template.py
@@ -122,19 +122,23 @@ def template(basedir, varname, templatevars, lookup_fatal=True, depth=0, expand_
                 varname = "{{%s}}" % varname
 
         if isinstance(varname, basestring):
-            if '{{' in varname or '{%' in varname:
-                try:
-                    varname = template_from_string(basedir, varname, templatevars, fail_on_undefined)
-                except errors.AnsibleError, e:
-                    raise errors.AnsibleError("Failed to template %s: %s" % (varname, str(e)))
+            try:
+                if '{{' in varname or '{%' in varname:
+                    try:
+                        varname = template_from_string(basedir, varname, templatevars, fail_on_undefined)
+                    except errors.AnsibleError, e:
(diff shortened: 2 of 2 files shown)


bcoca:
we cannot capture all exceptions here as it creates problems later on, as we discovered with similar patch in 1.8. We can look at something more targeted.

hfinucane:
What if I didn't change the type of the exception, and instead just tunneled the extra information somehow?

lib/ansible/utils/template.py
```diff
@@ -122,19 +122,23 @@ def template(basedir, varname, templatevars, lookup_fatal=True, depth=0, expand_
...
+                    # template_from_string may return non strings for the case where the var is just
+                    # a reference to a single variable, so we should re_check before we do further evals
+                    if isinstance(varname, basestring):
+                        if (varname.startswith("{") and not varname.startswith("{{")) or varname.startswith("["):
+                            eval_results = utils.safe_eval(varname, locals=templatevars, include_exceptions=True)
+                            if eval_results[1] is None:
+                                varname = eval_results[0]
+            except Exception, e:
+                e._ansible_srcinfo = { "basedir": basedir, "varname": varname }
+                raise
```
hfinucane, line 141: `raise`-with-no-arguments means the traceback, and the original exception, are unaffected.

lib/ansible/runner/__init__.py
```diff
@@ -590,8 +590,10 @@ def _executor(self, host, new_stdin):
             msg = str(ae)
             self.callbacks.on_unreachable(host, msg)
             return ReturnData(host=host, comm_ok=False, result=dict(failed=True, msg=msg))
-        except Exception:
+        except Exception, e:
             msg = traceback.format_exc()
+            if hasattr(e, "_ansible_srcinfo"):
+                msg = ("%(varname)s from %(basedir)s\n" % e._ansible_srcinfo) + msg
```
hfinucane, line 596: Ideally, there would be more information available, but this will at least let you know what is failing to evaluate, instead of just that... something somewhere is invalid.

hfinucane:
Can we talk about this again, now that it's more targeted?

bcoca:
yes, this is on my list to revisit in v2, also looking at deprecating implicit templating in with_

amenonsen:
@hfinucane meanwhile could you please rebase your PR to apply onto current devel?

sivel:
@hfinucane,

Hi!

hfinucane:
The fix in this PR is in mainline. You get an error like this-

```
the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'intermediate_value' is undefined
```

when trying to use something undefined, which is more informative than what I did. Sorry I kind of dropped this for two years.

PR #10053, opened 2015-01-20
V2 ssl configurable

bcoca:
now uses replacer code to set defaults, depends on config changes included in #9808


swimlappy:
Very nice, thanks!  Shouldn't you also fix lines 271 and 276 in v2/ansible/module_utils/urls.py
?  Those are also socket calls that require the SSL version to be set.

abadger:
Since the previous implementation was reverted, we need to re-add the config file and constants.py changes here.

Might be nice to generalize this into a way to add arbitrary constants to the modules but that would be a separate issue.

bcoca:
REJECTED!

PR #10055, opened 2015-01-20
Add the uuid of the device in ansible_mounts variable

digi691:
This will return 'NA' if the command is unsuccessful as older OSes may not have the lsblk command.

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -795,6 +795,13 @@ class LinuxHardware(Hardware):
                         size_available = statvfs_result.f_bsize * (statvfs_result.f_bavail)
                     except OSError, e:
                         continue
+                    lsblkPath = module.get_bin_path("lsblk")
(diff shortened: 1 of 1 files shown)


bcoca:
doesn't it make more sense to add the uuid at the block device listing? not all blocks may be mounted.

digi691:
The main reason i'm putting it in ansible_mounts is because ansible_devices does not show any information about logical volumes mounted or not.  As a side note I also think the UUID belongs in both places as the UUID can usually be synonymously used in place of the device name.

Out of scope of this commit:

I also noticed neither function get_device_facts or get_mount_facts returned info about my mounted SWAP partition.  get_device_facts does not seem to support LVM and this SWAP information is not available for get_mount_facts in /etc/mtab.  I haven't looked yet, but I bet things like ISO Loop Mounts, and possibly some BTRFS items that should appear in ansible_devices or ansible_mounts are not showing.  Just an FYI...

bcoca:
point taken, leaving uuid on block level for future PRs.

btrfs, zfs and other filesystems don't use traditional blocks as they are 1/2 volume managers (like lvm).
There are specific modules for managing these, requests to add this info to standard facts and/or specific fact modules will be considered.

PR #10064, opened 2015-01-22
Clarify order in which handlers are run

takis:
When reading the documentation I found that note not explicit enough

Although one could deduce the behavior because alternative run orders less likely, I still think it is better to be explicit about this in documentation.

### DIFF
--- a/docsite/rst/playbooks_intro.rst
+++ b/docsite/rst/playbooks_intro.rst
@@ -309,7 +309,7 @@ Handlers are best used to restart services and trigger reboots.  You probably
 won't need them for much else.
 
 .. note::
-   Notify handlers are always run in the order written.
+   Notify handlers are always run in the order written in the handlers list.
 
 Roles are described later on.  It's worthwhile to point out that handlers are
 automatically processed between 'pre_tasks', 'roles', 'tasks', and 'post_tasks'


amenonsen:
I'm not sure I understand the original problem. Would "Notify handlers are always run in the order in which they are defined" be better?

amenonsen:

bcoca:
yes, also it is a bit inaccurate as 'the handlers list' is not the only place where you can write handlers, so closing this.

takis:
My point was that it was not clear to me in which order they would run, because:
"run in the order written" immediately caused me to think, "written where?". Written in the notifiers list or in the file with a list of handlers.

At least I do not seem to be the only one who was confused :-)
http://typicalrunt.me/2014/11/09/be-careful-with-the-order-of-ansible-handlers/

dagwieers:
Yes, this is confusing (even without the ambiguous documentation, re-enforcing that Ansible's behaviour is erratic).

I wonder what the rationale was to use the order in the handlers-file(s) and not the order that was explicitly requested from the notify-statement ? This seems to contradict the design goals. Lists are always ordered, aren't they ?

I must have been lucky to never have stumbled over this before.

PR #10066, opened 2015-01-22
Document checksum error codes

hughsaunders:

### DIFF
--- a/docsite/rst/playbooks_error_handling.rst
+++ b/docsite/rst/playbooks_error_handling.rst
@@ -84,6 +84,20 @@ does not cause handlers to fire::
         changed_when: False
 
 
+Error Codes
+```````````
+
+The Checksum function used by the template module and probably others returns
+the following error codes:
+
+    1: File not found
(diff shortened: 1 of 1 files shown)


amenonsen:
I know it was meant in a lighthearted way, but "get a checksum error" really doesn't help someone who's reviewing this PR without being involved in any earlier discussion. So I had to dig through the code to find out what @hughsaunders meant. Well, perhaps it was an appropriate punishment.

Anyway, I don't think this PR should be applied in its present form. First, saying "probably others" just won't do in the documentation. Yes, the checksum code is used by several other modules, including assemble, copy, fetch. But template seems to be the only one that will dump the remote_checksum value in an error message. Second, the template module seems to print 0/2/3/4, but the patch documents only 1/2/3 (and the code says 1 is fine). Third, the real problem is that template dumps the value when it encounters an error, and the proper fix would be to make it stop doing that, not to enshrine its mildly antisocial behaviour in documentation.

PR #10069, opened 2015-01-22
Fix ansible-doc --list on Solaris

mavit:
On Solaris, `/usr/bin/stty` has no `size` option.  However, `/usr/ucb/stty` does have this option.

Install package `compatibility/ucb` to get `/usr/ucb/stty` on Solaris 11.


bcoca:
closing in favor of #10174

Issue #10074, opened 2015-01-23
a tasks"sudo" directive is not templatable


ptisserand:
Issue Type: Bug Report
Ansible version: ansible 1.7.2
Environment: Debian testing amd64

Summary:
Using ansible 1.7.2, I'm not able to use a variable in order to change "sudo" parameter for a task.
I have try with the following playbook:

```
- hosts: all
  vars:
    prod: true

  tasks:
    - name: Show me
      shell: "whoami && echo {{ prod }}"
      sudo: "{{ prod }}"
```

And the result is the following:
`changed: [**EDITED**] => {"changed": true, "cmd": "whoami && echo True", "delta": "0:00:00.009227", "end": "2015-01-23 15:21:18.719197", "rc": 0, "start": "2015-01-23 15:21:18.709970", "stderr": "", "stdout": "robin\nTrue"}`

Using current HEAD (53a3644ecd49dc7e9ee1a5ba5131a9298055dd33) of devel branch, I still have this issue.

Steps to reproduce: 
1- create a file name /tmp/foo.yml with the following contents:

```
- hosts: all
  vars:
    prod: true

  tasks:
    - name: Show me
      shell: "whoami && echo {{ prod }}"
      sudo: "{{ prod }}"
```

2- run `ansible-playbook -v /tmp/foo.yml`

Expected results: 
Output should contain:
`changed: [**EDITED**] => {"changed": true, "cmd": "whoami && echo True", "delta": "0:00:00.009227", "end": "2015-01-23 15:21:18.719197", "rc": 0, "start": "2015-01-23 15:21:18.709970", "stderr": "", "stdout": "root\nTrue"}`

Actual results:
`changed: [**EDITED**] => {"changed": true, "cmd": "whoami && echo True", "delta": "0:00:00.009227", "end": "2015-01-23 15:21:18.719197", "rc": 0, "start": "2015-01-23 15:21:18.709970", "stderr": "", "stdout": "robin\nTrue"}`

udondan:
Confirmed, having the same issue.

jimi-c:
Hi, this situation is no longer present in the devel branch (which will be ansible 2.0).

If you continue seeing any problems related to this issue, or if you have any further questions, please let us know by stopping by one of the two mailing lists, as appropriate:
- https://groups.google.com/forum/#!forum/ansible-project - for user questions, tips, and tricks
- https://groups.google.com/forum/#!forum/ansible-devel - for strategy, future planning, and questions about writing code

[...]

PR #10075, opened 2015-01-23
makes tasks' sudo directive templatable

ptisserand:
This pull request is a fix for issue #10074 

Use template expansion to evaluate value before setting sudo/su parameters.
Reorder source code to get all variable before evaluation for sudo and su parameters.


amenonsen:
This PR should be closed now, there's no problem in devel using «sudo: "{{ variable }}"» (I tested). The corresponding issue #10074 has already been closed.

bcoca:
closing as per comment above

PR #4664, opened 2013-10-24
Lookup plugin for etcd

jpmens:
This is a new lookup plugin for CoreOS' [etcd](https://github.com/coreos/etcd).

Example playbook (including deprecated `$ETCD()` invocation):

``` yaml

- hosts: localhost
  gather_facts: false
  vars:
  - myname: "{{ lookup('etcd', 'name') }}"
  - mynum: "{{ $ETCD(nr) }}"

  tasks:
  - action: debug msg="the value of name is {{ myname }} (and {{mynum}})"
```

PR #10078, opened 2015-01-24
Ansible support for Consul

sgargan:
##### Issue Type:

Feature Pull Request
##### Ansible Version:

1.9
##### Environment:

OSX 10.10 (Yosemite)
Linux (Ubuntu 14.04 Trusty Tahr) 3.13.0-37-generic
##### Summary:

Initial commit of Ansible support for the Consul clustering framework (http://consul.io).

Submission includes support for
- creating and registering services and checks
- reading, writing and lookup for values in consul's kv store
- creating and manipulating sessions for distributed locking on values in the kv
- creating and manipulating ACLs for restricting access to the kv store
- inventory support that reads the Consul catalog and group nodes according to
  - datacenters
  - exposed services
  - service availability
  - arbitrary groupings from the kv store

This submission makes extensive use of the python-consul library and this is required
as a dependency and can be installed from pip.
##### Steps To Reproduce:

The tests were written to target a vagrant cluster which can be setup by following the
instructions here http://github.com/sgargan/consul-vagrant

Note that this submission contains only the inventory & lookup plugins and tests, the modules are contained in a commit to the ansible-modules-extras project.

### DIFF
--- a/lib/ansible/runner/lookup_plugins/consul_kv.py
+++ b/lib/ansible/runner/lookup_plugins/consul_kv.py
@@ -1,0 +1,128 @@
+# (c) 2015, Steve Gargan <steve.gargan@gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+'''
+Lookup plugin to grab metadata from a consul key value store.
+============================================================
+
+Plugin will lookup metadata for a playbook from the key value store in a
+consul cluster. Values can be easily set in the kv store with simple rest
+commands e.g.
+
+curl -X PUT -d 'some-value' http://localhost:8500/v1/kv/ansible/somedata
+
+this can then be looked up in a playbook as follows
+
+- debug: msg='key contains {{item}}'
+  with_consul_kv:
+    - 'key/to/retrieve'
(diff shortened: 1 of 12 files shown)


davidvasandani:
+1 can't wait for this. @sgargan Does this support lookup expressions?

`secret: "{{ lookup('consul', 'key/to/retrieve') }}"`

vs

```
with_consul_kv:
    - 'key/to/retrieve'
```

This would be inline with the other lookup plugins like etcd

```
- debug: msg="{{ lookup('etcd', 'foo') }} is a value from a locally running etcd"
```

Documented here: http://docs.ansible.com/playbooks_lookups.html#more-lookups

xakraz:
+1: Can't wait either !

I have already written and am using a lookup_plugin for consul
I was thinking to submit it but yours is great !

davidvasandani:
@bcoca Now that this has been merged where is the best place to have discussion regarding consul lookup expressions?

sgargan:
Here is good for on the record stuff, the irc channel is better for a quick
response. What would you like the lookup plugins to do?

davidvasandani:
I was hoping that this plugin would use similar formatting to etcd.

etcd using a lookup expression that looks like:

`- debug: msg="{{ lookup('etcd', 'foo') }} is a value from a locally running etcd"`

I'd like to do something like this with the consul plugin:

`debug: msg="{{ lookup('consul', 'key/to/retrieve') }}"`

Thanks @sgargan

sgargan:
David,

The behavior you're looking for is there, its just named consul_kv not
consul; naming it consul caused a namespace collision with the consul
library and caused it to load the wrong file. You can see examples of using
it toward the end of this integration script

https://github.com/ansible/ansible/blob/devel/test/integration/roles/test_consul_kv/tasks/main.yml

look for with_consul_kv.

Let me know if this works for you.

Steve.

xakraz:
@davidneudorfer , Yes my lookup was inspired by the one written by @jpmens : #4664

It was just a basic substitution between etcd / Consul key path + B64 decoding support ...

I don't know how I can submit the contribution somewhere

tzz:
It's great that this was merged.  Will there be docs for the inventory and metadata modules?  I couldn't find them, if they exist already.

davidbirdsong:
+1 was about to write a simpler version of this and was lucky to turn up this PR after a few searches.

PR #10082, opened 2015-01-24
add dictmap function

jamespharaoh:
Provides useful function to map a list with keys to values via a dictionary.

Sample usage:

```
[3, 1] | dictmap ({ 1: 'one', 2: 'two', 3: 'three' })
```

Result:

```
['three', 'one']
```

### DIFF
--- a/docsite/rst/playbooks_variables.rst
+++ b/docsite/rst/playbooks_variables.rst
@@ -216,6 +216,11 @@ To get the maximum value from a list of numbers::
 
     {{ [3, 4, 2] | max }}
 
+To map the items in a list using a dictionary:
+
+    {{ [3, 1] | dictmap ({ 1: "one", 2: "two", 3: "three" }) }}
+
+
 .. _set_theory_filters:
 
 Set Theory Filters
--- a/lib/ansible/runner/filter_plugins/core.py
+++ b/lib/ansible/runner/filter_plugins/core.py
@@ -221,6 +221,7 @@ def max(a):
     return _max(a);
 
 def dictmap(l, d):
+    ''' Map the elements of a list using a dictionary '''
     return [d[k] for k in l]
 
 def version_compare(value, version, operator='eq', strict=False):
--- a/lib/ansible/runner/filter_plugins/core.py
+++ b/lib/ansible/runner/filter_plugins/core.py
@@ -220,6 +220,9 @@ def max(a):
     _max = __builtins__.get('max')
     return _max(a);
 
+def dictmap(l, d):
+    return [d[k] for k in l]
+
 def version_compare(value, version, operator='eq', strict=False):
     ''' Perform a version comparison on a value '''
     op_map = {
@@ -382,6 +385,7 @@ class FilterModule(object):
             'union': union,
             'min' : min,
             'max' : max,
+            'dictmap' : dictmap,
 
             # version comparison
             'version_compare': version_compare,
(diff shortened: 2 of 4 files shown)


ghost:
you'd also have to add documentation for this, and perhaps a test.

jamespharaoh:
Ok, I can add some tests, and a doc string in the function, but I'm not really sure where ansible's custom jinja2 filters are documented, if anywhere. Is there a specific place for these?

bcoca:
http://docs.ansible.com/playbooks_variables.html#jinja2-filters

in repo docsite/rst/playbooks_variables.rst

Gerst20051:
:+1:

cemo:
+1

jamespharaoh:
ok, i think this is what you've asked for. please let me know any other amends etc.

amenonsen:
Can't this be done with the builtin `map` in current jinja2?

jamespharaoh:
As i understand it, the map function applies a jinja2 function to a list, but i am not aware of any way to make it perform a dictionary lookup.

"Applies a filter on a sequence of objects or looks up an attribute. This is useful when dealing with lists of objects but you are really only interested in a certain value of it."

So the map function can do x -> y(x) or x -> x[y] but not x -> y[x], which is basically what my function does.

amenonsen:
> So the map function can do x -> y(x) or x -> x[y] but not x -> y[x], which is basically what my function does.

Ah, but `y(x)` is really just `y[x]` in disguise. You could do something like this:

```
- debug: msg="{{ ['a','b'] | map("dictlookup", {'a': 31, 'b': 42, 'c': 44}) | list }}"
```

For a very simple test, I created a `dictlookup` filter that just does `return b[a]`, and this works fine whether you pass in a literal dict or a dict that's defined elsewhere (e.g. in `vars`). I could imagine a slightly more clever `lookup` filter that detected the type of b and looked up either a dict key or a list index. (I'm a bit surprised there isn't one already.)

I really don't mean to argue for the sake of arguing. My approach is a one-line filter, and so is yours. I just think writing something that works with `map` is a more general approach that fits the templating language better than to add a variety of special-case `xxxmap` filters. The example above makes my approach look more clumsy than yours, but I think in most cases it'll look like this, which really isn't too bad:

```
- debug: msg="{{ somelist|map('lookup',somedict)|list }}"
```

Thoughts?

jamespharaoh:
I prefer lots of simpler functions, since it makes the code a little easier to read, but I really don't have a strong opinion. Certainly what you suggest should be possible as well.

amenonsen:
@bcoca can we have an opinion from core on which approach would be preferred? A special-case `listvar|dictmap(dictvar)` filter (already implemented) or an equally simple but (very slightly) more general `listvar|map("lookup", dictvar)` that could also be made to work with lists? (See discussion above.)

halberom:
agree with @amenonsen, given map already exists, it can easily be reused for iterating, so just create a filter that works on the individual items.

i.e. I would use something like dictvalue. then I can do

```
'foo' | dictvalue({"foo": "bar"})   # = 'bar'
['foo', 'abc'] | map('dictvalue', {"foo": "bar", "abc": "xyz"}) | list   # = ['bar', 'xyz']
```

obviously dictvalue isn't actually useful for anything else because you can just do `somedict[somekey]` or `somedict.somekey`, but the principle is inline with other things in jinja - see [equalto()](http://jinja.pocoo.org/docs/dev/templates/#equalto).

it could also be extended to give a default value in the event a key doesn't exist

amenonsen:
@jamespharaoh do you have any plans to submit an alternative implementation according to the earlier suggestions from @halberom and me? If not, I would like to, since I have now found a use case for it myself.

sivel:
@jamespharaoh,

Hi!

amenonsen:
Closing per above comment, since `extract` is now merged.

PR #10085, opened 2015-01-25
allow missing subelement keywords to be optionally ignored

aks:
This request is a simple change to `subelements.py` and `constants.py`, which allows the user to optionally ignore missing keywords on a `with_subelements` loop.  The change is a new configuration item: `ignore_missing_subelements`, and the logic to use the variable in `subelements.py`.

The default for this config variable is false, which causes the code in `subelements.py` to behave as usual and raise an error when subelement keywords are missing.  The default behavior is less flexible, but since some users may have written their playbooks or roles to depend on this, we leave it as the default.

However, the more flexible -- and typical -- usage of this feature will be to have `ignore_missing_subelements` set to true, which allows for flexible data patterns in the application of tasks looping with `with_subelements`.  The items for which there is no corresponding key are simply ignored.

At our site, this feature is important (and in use) for using ansible in managing keys in our user's `authorized_keys` file, in that (a) not all of our users have the same number of ssh keys, and (b) some of our users have no ssh key.

### DIFF
--- a/lib/ansible/constants.py
+++ b/lib/ansible/constants.py
@@ -160,6 +160,7 @@ DEPRECATION_WARNINGS           = get_config(p, DEFAULTS, 'deprecation_warnings',
 DEFAULT_CALLABLE_WHITELIST     = get_config(p, DEFAULTS, 'callable_whitelist', 'ANSIBLE_CALLABLE_WHITELIST', [], islist=True)
 COMMAND_WARNINGS               = get_config(p, DEFAULTS, 'command_warnings', 'ANSIBLE_COMMAND_WARNINGS', False, boolean=True)
 DEFAULT_LOAD_CALLBACK_PLUGINS  = get_config(p, DEFAULTS, 'bin_ansible_callbacks', 'ANSIBLE_LOAD_CALLBACK_PLUGINS', False, boolean=True)
+IGNORE_MISSING_SUBELEMENTS     = get_config(p, DEFAULTS, 'ignore_missing_subelements', 'IGNORE_MISSING_SUBELEMENTS', False, boolean=True)
 
 # CONNECTION RELATED
 ANSIBLE_SSH_ARGS               = get_config(p, 'ssh_connection', 'ssh_args', 'ANSIBLE_SSH_ARGS', None)
--- a/lib/ansible/runner/lookup_plugins/subelements.py
+++ b/lib/ansible/runner/lookup_plugins/subelements.py
@@ -17,6 +17,7 @@
 
 import ansible.utils as utils
 import ansible.errors as errors
+import ansible.constants as constants
 
 
 class LookupModule(object):
@@ -56,6 +57,8 @@ class LookupModule(object):
                 # this particular item is to be skipped
                 continue 
             if not subelement in item0:
+                if constants.IGNORE_MISSING_SUBELEMENTS:
+                    continue      # skip this item because it has no subelement
                 raise errors.AnsibleError("could not find '%s' key in iterated item '%s'" % (subelement, item0))
             if not isinstance(item0[subelement], list):
                 raise errors.AnsibleError("the key %s should point to a list, got '%s'" % (subelement, item0[subelement]))


podung:
+1 Would love to see this PR accepted.

cadetstar:
+1 Would love to see this as well.

johnalla0:
+1, we need this

alanflorendo:
+1 yes please

amenonsen:
I don't have any strong feeling about this change, but it's obvious that some people like it. The code itself is trivial and should be easy to rebase to apply to current devel. It needs documentation updates, of course. Before spending energy on rebasing it, however, can we have an opinion on whether this is desirable at all?

aks:
This feature is still very desirable in general, and specifically for my use case.

I have a role called `aks.ansible-users` (forked and modified from `devops.ansible-users`) which takes one or more lists of input and manages almost all aspects of a user account on many systems, including SSH key administration.  This role also removes users, when the `state` is set to `absent`.

Without this patch, this kind of code is needed:

```
# gone users
#
- { name: 'bob',    state: absent , ssh_keyfiles: []}
- { name: 'jane',   state: absent , ssh_keyfiles: []}
- { name: 'frank',  state: absent , ssh_keyfiles: []}
- { name: 'suzy',   state: absent , ssh_keyfiles: []}
- { name: 'tracy',  state: absent , ssh_keyfiles: []}
- { name: 'james',  state: absent , ssh_keyfiles: []}
- { name: 'amir',   state: absent , ssh_keyfiles: []}
- { name: 'adjit',  state: absent , ssh_keyfiles: []}
- { name: 'sunil',  state: absent , ssh_keyfiles: []}
```

These empty `ssh_keyfiles` are also needed on new users without ssh keys.

Without those empty `ssh_keyfiles` (and without this patch), the `with_subelements` phrase will complain.

Of course, I can rewrite the tasks so that there are separate list of users: those with `ssh_keyfiles`, and those without, and, additionally, a separate list of "gone" users.

There is no use-case (for me) where a missing subelement error _should_ be detected.  But, if I _had_ to do so, a simple test for an empty `item.1` would suffice.

So, in all my usage of `with_subelements`, I have had to work harder to ensure that missing subelements are absent.  Multiply that effort by all the other ansible developers working under same constraint.

If `with_subelements` were more flexible, and skipped records for which there was no corresponding `item.1`, there would be a little less work involved with setting up and using lists of subelements.  In short, all the ansible developers would be doing a little bit less work than before.

amenonsen:
@aks I actually meant an opinion from the core team about whether this would be merged, I pretty much assumed your own opinion would be positive. :-) But I must say you make your case well.

bcoca:
so instead of a global config option it would be better to just pass it directly to the lookup, there are several that take additional options already you can use as a guide.

amenonsen:
@aks are you planning to work on revising your PR as suggested by @bcoca?

aks:
So, I'll rework the PR to incorporate a config option, but if my 99% use case involves always passing this option, it would also be better to have a global default value to apply in the absence of the passed option.

Otherwise, on each instance of `with_subelements`, I'm working harder to get what I believe should be the default behavior.

In fact, I'd really appreciate someone to speak up and share the use case of `with_subelements` that benefited from an error on a missing subelement.  If the current default is benefiting no one, why is it the default?

sivel:
@aks,

Hi!

amenonsen:
I have to say I find @aks's logic pretty compelling. I can't think of a reason for with_subelements to have to be configured to not break (but then I don't use with_subelements much).

bcoca:
The main reason this should fail is to signal to the user that their data is incomplete (how it currently works), that is why an explicit parameter makes more sense to me to 'ignore missing', it is not a default behaviour i would expect of any system.

aks:
@bcoca, the case for exactly conforming data is apparently much less frequent than the case for flexibly matching the data that is available.  So, all things being equal, it would seem logical to make the default case be the common and more flexible case of skipping missing sub-elements silently.

We can, as suggested, create an option to `with-subelements` that can cause either failures on missing sub-elements, or silence on them.  If the default were configurable, then both options would be required, because otherwise, how would a playbook be written to operate consistently, regardless of configuration?

Another, less-verbose way than using options on the existing iterator, is to create new, more specific iterators: `with-optional-subelements` and `with-required-subelements`.  This avoids the use of options, and makes the playbook design explicit with consistent behavior.

If there is general agreement, I'll volunteer some time to code up both of these new iterators for ansible2, and submit for review.

bcoca:
First, I don't agree that this is the common case, second, this would be true for a NEW development, but since this is how it has been working until now, backwards compatibility comes into play and pushes for the default to be the existing behaviour.

bcoca:
another way to do this is just using select/selectattr/reject filters on the list before subelements gets it.

jimi-c:
Hi all, going back through old PRs and noticed this. My main object here is that it's adding a config option for a lookup plugin. In general, that's not how we've done things. The direction we want to go is to standardize the way in which args are passed into lookup plugins, as right now there are several disparate mechanisms (none of which are actually documented I think).

So, for now we'll close this, though we should really come up with a new proposal for the way we handle args to lookup plugins.

Any further discussion of this should take place on the devel mailing list:

Thank you!

PR #10087, opened 2015-01-25
fix checksum calculation for large files

### LINKED ISSUE #10086: Shell checksum validation fails on large files
Current##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.8.2
##### Environment:

Linux
##### Summary:

Current implementation reads whole file into memory. For files larger than available memory, it fails.
##### Steps To Reproduce:

Try fetching a large file from a remote server
##### Expected Results:

It should work
##### Actual Results:

 {"changed": false, "file": "/root/foobar-folyondar-sport.tar.gz", "msg": "unable to calculate the checksum of the remote file"}

(and the file does not get created locally)

axos88:
Fixes https://github.com/ansible/ansible/issues/10086 - now reading files in chunks

### DIFF
--- a/lib/ansible/runner/shell_plugins/sh.py
+++ b/lib/ansible/runner/shell_plugins/sh.py
@@ -98,7 +98,7 @@ class ShellModule(object):
 
         test = "rc=flag; [ -r \'%(p)s\' ] || rc=2; [ -f \'%(p)s\' ] || rc=1; [ -d \'%(p)s\' ] && rc=3; %(i)s -V 2>/dev/null || rc=4; [ x\"$rc\" != \"xflag\" ] && echo \"${rc}\"\'  %(p)s\' && exit 0" % dict(p=path, i=python_interp)
         csums = [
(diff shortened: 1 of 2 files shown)


abadger:
Hmm...  Two comments that are tied together:
- Python2.4 doesn't have the with statement.  (It doesn't have hashlib in the stdlib either but there is a hashlib backport on pypi that may be installed on the machines).  If we can avoid the with statementthen we should.
- If we can make a version that doesn't use the with statement then we should also apply the fix to the import sha version of the code.

Also -- I wasn't able to get the \n syntax to work with python -c from my shell.  Are you sure that it works?  I found an answer on stackoverflow that might let things work via exec: http://stackoverflow.com/questions/2043453/executing-python-multi-line-statements-in-the-one-line-command-line

axos88:
Yeah - i wasn't able to get it working from my shell either, but I am now using this patch, so yeah it works.
Probably what happens is bash does not actuall translate the \n-s into newlines, but the python interpolator does.

axos88:
I'll try to rewrite this without the with statement, sorry, i'm not very familiar with python.

abadger:
I think something similar to this will work:

python -c 'exec("import hashlib; BLOCKSIZE = 65536; hasher = hashlib.sha1();afile = open(\"%s\", \"rb\", BLOCKSIZE)\nbuf = afile.read(BLOCKSIZE)\nwhile buf:\n\tbuf = afile.read(BLOCKSIZE)\n\thasher.update(buf)\nprint(hasher.hexdigest())")'

abadger:
Thanks @axos88! 

# Merged

Hi!

issue is resolved.

Thank you!

PR #10089, opened 2015-01-25
Ld library path

sirkubax:
1.
Following changes would  allow for defining custom LD_LIBRARY_PATH 'per host' in inventory (or in  ansible.config - default for all hosts)

example:
[myhosts]
somehostname ld_library_path=/home/user/mycustomlib:/opt/openssl/lib:/usr/local/lib

2.
Allow for skipping SELINUX - important when you use custom ansible_python_interpreter that does not work with selinux (that is installed on host)
usage:
skip_selinux=True


sirkubax:
LAter

PR #10095, opened 2015-01-26
Expand reties: parameter from template

ashorin:
... along with delay:


amenonsen:
I'm not sure what this is supposed to do, but the code has changed completely, and this can be closed now.

bcoca:
all task parameters are now templated by default (devel), so there should be no need of specific patches to add templating anymore. So I'm closing this PR

Issue #10094, opened 2015-01-26
setup.py not python3 friendly


gilou:
Hi,

When installing via pip with python3, it fails:

```
File "setup.py", line 12
    print "Ansible now needs setuptools in order to build. " \
                                                           ^
SyntaxError: Missing parentheses in call to 'print'
```

I'd assume this is to be backward compatible, but using print() in setup.py will probably fix it, then using `__future__` or any other 2to3 magic if compatibility with Python < 2.6 is a requirement there.

Not doing a PR, as I'm not sure what the policy is in this regard, but this is trivial, unless there are python3 incompatibilities somewhere else.

(This was tested on devel, 53a3644ecd49dc7e9ee1a5ba5131a9298055dd33)

abadger:
Thanks!  I've merged your fix.

PR #10096, opened 2015-01-26
Make setup.py work with Python 3 (and requiring 2.6)

gilou:
Linked to https://github.com/ansible/ansible/issues/10094

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,8 +9,8 @@ from ansible import __version__, __author__
 try:
     from setuptools import setup, find_packages
 except ImportError:
-    print "Ansible now needs setuptools in order to build. " \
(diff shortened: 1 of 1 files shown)


abadger:
I'm going to merge this because it is backwards compatible (older python will treat the parenthesis as simply grouping or line continuation.)

However, ansible itself is likely pretty far from python3 compatibility.  We're hoping to make that better in the v2 branch but it's likely to be a hard enough problem that we won't make official dual python2-python3 compatibility.  Instead we'll switch from python2 to python3 as the Linux distributions get to the point where they're shipping python3 and not python2.

# Merged

Hi!

issue is resolved.

Thank you!

PR #10097, opened 2015-01-26
Add a ansible_fw_manager fact

mscherer:
Main use will be to know when to use firewalld module or
ufw module to open ports, but more can be surely found.

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -119,6 +119,13 @@ class Facts(object):
                  { 'path' : '/usr/bin/pkg',         'name' : 'pkg' },
     ]
 
+    # Like PKG_MGRS, place default firewall manager last. And if there is a
+    # ansible module, use this as the value for 'name' key.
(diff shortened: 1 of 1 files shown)


amenonsen:
Just a quick thought: I'm not sure how sound it is to set `ansible_fw_manager` based on the existence of the `ufw` binary or so on. If more than one is installed, the order of definition of the entries in the list will determine which one we think is in use, regardless of which is actually in use (if any at all). I'm not sure how much that matters.

mscherer:
I am not sure if more than one will be installed at a time. But I assume we would have sooner or later the same problem as we had with ansible_pkg_mgr ( like dfn/yum transition ). What kind of heuristic could we use to find what would/should be the prefered firewall interface ?

( for firewalld, I assume having it running is a clear signal, but for the others... )

bcoca:
I've had ufw, firewalld and others installed on a system i managed with just iptables ..., the binaries don't mean much.

bcoca:
I'm going to close this, reopen if we find away around the impasse.

PR #10113, opened 2015-01-27
Url lookup

bcoca:
now you can loop over what a url returns, with the same usability and security concerns of piping curl/wget into your shells!

This does nothing fancy, can easily be expanded to handle specific needs of different URLS (ftp authentication? https? etc).

### DIFF
--- a/docsite/rst/playbooks_lookups.rst
+++ b/docsite/rst/playbooks_lookups.rst
@@ -128,7 +128,7 @@ Here are some examples::
 
          - debug: msg="{{ lookup('etcd', 'foo') }} is a value from a locally running etcd"
 
-         - debug: msg="{{item}}
+         - debug: msg="{{item}}"
            with_url:
                 - 'http://github.com/gremlin.keys'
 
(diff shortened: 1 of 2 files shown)


docsite/rst/playbooks_lookups.rst
```diff
@@ -128,6 +128,10 @@ Here are some examples::
 
          - debug: msg="{{ lookup('etcd', 'foo') }} is a value from a locally running etcd"
 
+         - debug: msg="{{item}}
```
resmo: there is a " missing

resmo:
Would this also work with other protocols? e.g. FTP?

bcoca:
yes, tested with   'ftp://ftp.us.debian.org/debian/README', for example

bcoca:
these were my tests:

```
- hosts: localhost
  gather_facts: false
  connection: local
  tasks:
    - debug: msg="{{item}}"
      with_url:
        - 'http://github.com/otherguy.keys'
        - 'ftp://ftp.us.debian.org/debian/README'
        - 'file:///tmp/test1'
```

resmo:
nice!

cemo:
nice

PR #10122, opened 2015-01-29
Leveraging setuptools automatic script creation

lorencarvalho:
Hi there,

Not a huge change here, mostly cosmetically moving the /bin directory into the library and substitutiing 'scripts' in setup.py to use setuptools automatic script creation instead. 

Conceptually, it's a bit of a departure, so I understand if this pull is turned down. I mostly did this for my own needs, as I want to use ansible as a library and then import the Cli() object from bin/ansible. Using automatic script creation makes that super simple, but it also has other benefits.

https://pythonhosted.org/setuptools/setuptools.html#automatic-script-creation

Thanks for your consideration!

### DIFF
--- a/lib/ansible/cli/ansible_.py
+++ b/lib/ansible/cli/ansible_.py
@@ -1,5 +1,3 @@
-#!/usr/bin/env python
-
 # (c) 2012, Michael DeHaan <michael.dehaan@gmail.com>
 #
 # This file is part of Ansible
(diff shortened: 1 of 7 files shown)


cchurch:
It has been discussed before in https://groups.google.com/d/topic/ansible-devel/c4rf3pR6zHk/discussion and https://github.com/ansible/ansible/pull/9401.

lorencarvalho:
fair enough, I suppose I can just patch locally until V2

bcoca:
closing this as per comments above

lorencarvalho:
@bcoca @cchurch 

In both the devel list and the previously closed ticket there didn't seem to be a commitment to moving to console_scripts -- 
Is there any new authoritative decision or is it still just being considered?

abadger:
There hasn't been any decision that I've heard of.  I believe it's still just a possible option.  Are there reasons that an entrypoint would be good?  My experiences with them have been mostly "It either works or fails horribly in certain situations."  But I haven't seen much benefit.

I could see benefits to moving most of the code out of the bin/ansible\* script itself and into the module without using setuptools entry-points to create the script itself... perhaps we should compare costs and benefits for both of those?

abadger:
Also -- this is a closed ticket ;; idscussion is probably best had on the devel mailing list.  And then if there's consnsus reached we can open a ticket or PR to work on implementing it.

PR #10129, opened 2015-01-29
Fix for memory fact gathering

RadishTheHut:
I have a host which started to fail while gathering facts after the addition
of expanded memory facts in PR #9839:

  File "/home/ansible/.ansible/tmp/ansible-tmp-1422536976.05-133253824703289/setup", line 4278, in <module>
    main()
  File "/home/ansible/.ansible/tmp/ansible-tmp-1422536976.05-133253824703289/setup", line 137, in main
    data = run_setup(module)
  File "/home/ansible/.ansible/tmp/ansible-tmp-1422536976.05-133253824703289/setup", line 81, in run_setup
    facts = ansible_facts(module)
  File "/home/ansible/.ansible/tmp/ansible-tmp-1422536976.05-133253824703289/setup", line 4217, in ansible_facts
    facts.update(Hardware().populate())
  File "/home/ansible/.ansible/tmp/ansible-tmp-1422536976.05-133253824703289/setup", line 2339, in populate
    self.get_memory_facts()
  File "/home/ansible/.ansible/tmp/ansible-tmp-1422536976.05-133253824703289/setup", line 2375, in get_memory_facts
    'cached': memstats['swapcached']
KeyError: 'swapcached'

My problem host doesn't have SwapCached in /proc/meminfo. It may be better to
set defaults for these keys, since the values provided by /proc/meminfo can
change from version to version.

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -635,19 +635,19 @@ class LinuxHardware(Hardware):
                  memstats[key.lower()] = long(val) / 1024
         self.facts['memory_mb'] = {
                      'real' : {
-                         'total': memstats['memtotal'],
-                         'used': (memstats['memtotal'] - memstats['memfree']),
-                         'free': memstats['memfree']
+                         'total': memstats.get('memtotal', 0),
+                         'used': (memstats.get('memtotal', 0) - memstats.get('memfree', 0)),
+                         'free': memstats.get('memfree', 0)
                      },
                      'nocache' : {
-                         'free': memstats['cached'] + memstats['memfree'] + memstats['buffers'],
-                         'used': memstats['memtotal'] - (memstats['cached'] + memstats['memfree'] + memstats['buffers'])
+                         'free': memstats.get('cached', 0) + memstats.get('memfree', 0) + memstats.get('buffers', 0),
+                         'used': memstats.get('memtotal', 0) - (memstats.get('cached', 0) + memstats.get('memfree', 0) + memstats.get('buffers', 0))
                      },
                      'swap' : {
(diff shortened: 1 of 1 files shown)


bcoca:
I agree we should have defaults, but 0 seems to be a deceptive one, shouldn't it be None or N/A?

RadishTheHut:
True, None would be better, but then there would need to be more logic around the arithmetic. Maybe we could have the direct values default to None, and the derived values default to 0. Or add some if blocks to set the derived values to None if any of the values are not present.

abadger:
+1 to None.  if we're going to shield against any of these values being left out it's probably easier to first create the memstats dict with with all of the keys we are going to use and values set to None and then overwrite the values if we find relevant data in /proc/meminfo.

For the calculated values I'd do as you're suggesting in your last sentence: a few conditional blocks to check whether any of the needed values are None and set the calculated value to None if so (otherwise do the calculation).

abadger:
@RadishTheHut what do you think?  If that set of changes ( https://github.com/ansible/ansible/commit/f20967078ebaa5c5bcb88e6ad12d50da28f761b5 ) looks good to you I'll push it to devel.

RadishTheHut:
@abadger I still got a KeyError for memstats['swapcached'], because that key hadn't been initialized. I've updated my pull request with an adjustment to your commit that uses memfacts.get(), setting any missing values to None. Please see what you think.

abadger:
<nod>  I see the problem with my change -- I forgot to lower the keys when I pre-filled the dict at the top.

Your revised PR looks like it will do the job too.  So merged, thanks!

# Merged

Hi!

issue is resolved.

Thank you!

Issue #10104, opened 2015-01-26
inventory is -i/--inventory-file on command line, "hostfile" in ansible.cfg


stenwt:
This seems arbitrary and confusing. Specifying "inventory" in ansible.cfg does nothing. The parameter, per docs at http://docs.ansible.com/intro_configuration.html is "hostfile".

bcoca:
# List Information

Hi!

Thanks very much for your interest in Ansible.  It sincerely means a lot to us. 

This appears to be a user question, and we'd like to direct these kinds of things to either the mailing list or the IRC channel.
- IRC: #ansible on irc.freenode.net  
- mailing list: https://groups.google.com/forum/#!forum/ansible-project

If you can stop by there, we'd appreciate it.  This allows us to keep the issue tracker for bugs, pull requests, RFEs and the like.

Thank you once again and we look forward to seeing you on the list or IRC.  Thanks!

maxamillion:
@bcoca I don't think this is a user question, is this really not considered to be an issue? 

There's inconsistent vocabulary which makes documentation unclear. If anything I would think this is an RFE to add consistency.

PR #10130, opened 2015-01-29
made inventory consistent in config file, deprecated old config hostfile

bcoca:
should bring naming consistency as per #10104

### DIFF
--- a/examples/ansible.cfg
+++ b/examples/ansible.cfg
@@ -11,7 +11,7 @@
 
 # some basic default values...
 
-hostfile       = /etc/ansible/hosts
+inventory      = /etc/ansible/hosts
 #library        = /usr/share/my_modules/
 remote_tmp     = $HOME/.ansible/tmp
 pattern        = *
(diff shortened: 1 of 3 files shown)


tyll:
This beaks overriding the inventory with ANSIBLE_HOSTS, becaus it introduces ANSIBLE_INVENTORY. ANSIBLE_INVENTORY is also not added to the documentation, only ANSIBLE_HOSTS.

bcoca:
ANSIBLE_HOSTS should still work to override the hostfile entry, but made note to update docs with ANSIBLE_INVENTORY

tyll:
ANSIBLE_HOSTS does not work if there is a "inventory" section in the config file. This seems to be silently addressed by you for the v2 branch in commit cdefeb6d84499d86bf6fef8352b06d626c1bf4ae - this will make ANSIBLE_INVENTORY not work if there is a "hosts" section in the config file, which might cause less trouble.

bcoca:
I thought that was case in both trees, v2 is about to be merged in so leaving it as is

PR #10132, opened 2015-01-29
Client side validation for winrm connections  

sgargan:
Issue Type:

Bugfix Pull Request

Ansible Version:

ansible 1.9 (detached HEAD 16f9ebd) last updated 2015/01/29 22:28:40 (GMT +100)

Environment:

OSX 10.10.1 Yosemite
Windows Server 2012 (Azure)
Python 2.7.9 (default, Dec 15 2014, 10:01:34) [GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.56)] on darwin

Summary:

Since python 2.7.9 ssl certificates are rejected by default if invalid and as a result python will object to self-signed certificates (see https://www.python.org/dev/peps/pep-0476/).

The current configuration for Windows clients sets up a self signed cert for for the HTTPS connector. If a 2.7.9 runtime is used to connect to this server then HTTPS connection will be rejected.

This patch allows you to provide a certificate to validate the ssl connection either via the ansible config property 'ansible_client_validation_cert' or as cli arg --client-validation-cert.

Steps To Reproduce:

Running the window example with a 2.7.9 python install and verbose logging the connection fails with a certificate error.

ESTABLISH WINRM CONNECTION FOR USER: sgargan on PORT 5986 TO sgargan.cloudapp.net
WINRM CONNECT: transport=plaintext endpoint=https://sgargan.cloudapp.net:5986/wsman
WINRM CONNECTION ERROR: 500 WinRMTransport. [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:581)
WINRM CONNECT: transport=plaintext endpoint=http://sgargan.cloudapp.net:5986/wsman

Expected Results:

Running with the certificate specified in the ansible.cfg or from the commandline, the connection succeeds as follows

WINRM CONNECT Context: verification cert=/Users/sgargan/Downloads/ansible.pem
ESTABLISH WINRM CONNECTION FOR USER: sgargan on PORT 5986 TO sgargan.cloudapp.net
WINRM CONNECT: transport=plaintext endpoint=https://sgargan.cloudapp.net:5986/wsman
REMOTE_MODULE win_ping
...

### DIFF
--- a/bin/ansible
+++ b/bin/ansible
@@ -155,6 +155,7 @@ class Cli(object):
             inventory=inventory_manager,
             timeout=options.timeout,
             private_key_file=options.private_key_file,
+            client_validation_cert=options.client_validation_cert,
             forks=options.forks,
             pattern=pattern,
             callbacks=self.callbacks,
@@ -212,4 +213,3 @@ if __name__ == '__main__':
         # Generic handler for ansible specific errors
         callbacks.display("ERROR: %s" % str(e), stderr=True, color='red')
         sys.exit(1)
-
--- a/lib/ansible/constants.py
+++ b/lib/ansible/constants.py
@@ -112,6 +112,7 @@ DEFAULT_POLL_INTERVAL     = get_config(p, DEFAULTS, 'poll_interval',    'ANSIBLE
 DEFAULT_REMOTE_USER       = get_config(p, DEFAULTS, 'remote_user',      'ANSIBLE_REMOTE_USER',      active_user)
 DEFAULT_ASK_PASS          = get_config(p, DEFAULTS, 'ask_pass',  'ANSIBLE_ASK_PASS',    False, boolean=True)
 DEFAULT_PRIVATE_KEY_FILE  = shell_expand_path(get_config(p, DEFAULTS, 'private_key_file', 'ANSIBLE_PRIVATE_KEY_FILE', None))
+DEFAULT_VALIDATION_CERT   = shell_expand_path(get_config(p, DEFAULTS, 'client_validation_cert', 'ANSIBLE_CLIENT_VALIDATION_CERT', None))
 DEFAULT_SUDO_USER         = get_config(p, DEFAULTS, 'sudo_user',        'ANSIBLE_SUDO_USER',        'root')
 DEFAULT_ASK_SUDO_PASS     = get_config(p, DEFAULTS, 'ask_sudo_pass',    'ANSIBLE_ASK_SUDO_PASS',    False, boolean=True)
 DEFAULT_REMOTE_PORT       = get_config(p, DEFAULTS, 'remote_port',      'ANSIBLE_REMOTE_PORT',      None, integer=True)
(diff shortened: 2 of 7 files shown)


kstrauser:
I was doing something similar, except using `ssl._create_unverified_context()` to disable verification (see [PEP 476 - Opting out](https://www.python.org/dev/peps/pep-0476/#opting-out) for context). Could you add a sentinel value like `<disabled>` to use the unverified context instead of a cert? Then users could just as easily disable verification altogether using the same mechanism if that suited their needs.

jvantuyl:
Just upgraded Python and hit this.  +1

sgargan:
This was my first approach too, though its completely insecure, a feature
like that is a lot of rope with which to hang yourself. I agree in a caveat
emptor approach so an option to disable might be useful so long as it
warned that it was insecure. On the other hand If you want to use it
insecurely, you'd be better off using the http endpoint rather than
unverified https giving the illusion of security.

kstrauser:
Perfect is the enemy of good: the current situation discourages use of SSL for anyone who doesn't want to spring for a CA-signed cert or set up their own CA. I'd bet that's a sizable portion of first-time Ansible users, and it's not a requirement that's imposed on someone managing servers via SSH (or Python 2.7.8- users).

Long-term, we could add caching of certs and reporting if one changed. That would get us roughly the same identity assurances that SSH provides in most cases; while it's possible to predistribute SSH host keys, in practice I've not seen this in the wild.

Also, https has a use case even here: basic auth sends usernames and passwords in cleartext. I can live with the risks of self-signed certs while in early development mode, but I'd rather not be sending readable passwords over the wire.

sgargan:
In no way am I advocating for an insecure channel. The PR allows you to use
a self signed cert, you just have to have access to it on both the server
and client side. IMHO this is better than disabling the varification, but
if that's something you'd like it would be easy to add.

jvantuyl:
I think disabling verification will help enough people who are "cert-impaired" that it is worth having.  In particular, it knocks down a roadblock during prototyping and testing; which is a critical part of the "adopting ansible" process.  It also gives us a chance to expose them to the right solution when they find it in the help text, docs, or Google.

Basically, choices are good, even when sometimes hazardous ones are on the menu.  -1 for enabling people who don't respect security to inflict the consequences on their users.  +1 for helping people test and prototype.  +1 for helping people adopt Ansible.  And +1 for all of those situations that we haven't thought of.

juliedavila:
This hits me too also on 2.7.9 @sgargan can you rebase?

cchurch:
See https://github.com/ansible/ansible/issues/10294#issuecomment-93629047 for a workaround.

klj613:
Assuming I understand this correctly. This would allow me to use a single self-signed cert per ansible-playbook play? If I had more than 1 windows machine I guess I'd have to use the same cert across them?

dreamcat4:
OK. So assuming that I know absolutely nothing. And havve followed the official instructions at: http://docs.ansible.com/ansible/intro_windows.html

That means running this powershell script on the target machine: http://docs.ansible.com/ansible/intro_windows.html#windows-system-prep

Then how do I find the self-signed cert on that machine? What's it's name, and how do I specify it to this new argument on the control machine? I.e. is it the fingerprint id, or the `.cert` file, etc? Can I have a real-world example to follow please? Including any steps on the windows machine. Many thanks.

dreamcat4:
So will this PR still be needed, when there is this other new PR out?

https://github.com/ansible/ansible/pull/12687

I mean, what's the difference?

sivel:
@sgargan,

Hi!

sivel:
@sgargan,

Hi!

sivel:
@sgargan,

Hi!

PR #10138, opened 2015-01-30
initial draft virtualbox inventory script

bcoca:
works for me, short n simple, no cache as it is normally local


tersmitten:
Great work. Is it problem that check_output is not available in python 2.6?

bcoca:
yes it is, I'll go back to popen, I've just been waiting to find an excuse to use any other of subproces' methods.

PR #10143, opened 2015-01-30
Adding --list-vars option to ansible-playbook

risingmoon:
I wanted a way to list out variables without running any plays.

First commit is the implementation of the code.
Second commit is my efforts to refactor and implement some PEP8 standards to this file.


bcoca:
we don't normally take formatting patches as they obscure the origins, also the 80 column limit is something we don't agree with much, at this point with screen size evolution it is actually counter productive.

risingmoon:
If I omit the second commit and PEP8 style changes then it should be fine?

risingmoon:
@bcoca I've removed the styling formats and just moved my imports to the top.

risingmoon:
Fixed merge conflicts.

Issue #10080, opened 2015-01-24
For fast install from git source, can we using git clone --depth=1?


ghost:
For fast install from git source, can we using git clone --depth=1?

The main repository growing fast , it will take more disk space to clone the full ansible repository

can we save our diskspace and getting faster to finish the clone process?

How about?

```
git clone git://github.com/ansible/ansible.git --recursive --depth=1
```

bcoca:
Closing this in favor of the pull request

PR #10144, opened 2015-01-31
Faster than normal clone, using git clone depth 1

ghost:
See https://github.com/ansible/ansible/issues/10080 #10080 
By default, git will clone full repository, it will eat your disk space.
For example, the linux kernel git repository larger than 1GB.
If you clone full repository, you may take long times and need more
diskspace.

We can make things getting easier, just clone git repo with depth.

it will working as normal, but save your time and diskspace.

### DIFF
--- a/docsite/rst/intro_installation.rst
+++ b/docsite/rst/intro_installation.rst
@@ -27,10 +27,10 @@ What Version To Pick?
 `````````````````````
 
 Because it runs so easily from source and does not require any installation of software on remote
-machines, many users will actually track the development version.  
+machines, many users will actually track the development version.
 
 Ansible's release cycles are usually about two months long.  Due to this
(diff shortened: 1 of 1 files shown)


ghost:
Any problem with this pull request? can it be merged into master branch now?

bcoca:
we go merging by priority and age, please be patient.

ghost:
Two weeks past, any new updates?

amenonsen:
I don't think this PR should be applied in its present form. At most, we want to add a sentence under the clone/cd/source example block that says "For a faster clone without full history, you can add the `--depth 1` option to the `git clone` command above". Repeating the block with one changed command is entirely unnecessary.

The patch also contains ten unrelated whitespace changes, so it'll inevitably accumulate merge conflicts over time.

ghost:
Due to long time no active , i will close it .thanks

PR #10145, opened 2015-01-31
filter plugins for IP arithmetic

alpo:
Can be used for traversing long IP ranges in playbooks and templates.


abadger:
I've just merged an ipaddr plugin: https://github.com/ansible/ansible/pull/9560  (documentation pending but there's an extensive README: https://github.com/drybjed/ansible-ipaddr-filter )

I think the ipaddr('int') and ipaddr('address') filters from that have the same functionality as you're making here.  Additionally, some of the other filters in there may make your job even easier.

# Possible Duplicated Efforts

Hi!

- https://github.com/ansible/ansible/pull/9560

PR #10154, opened 2015-02-02
Add missing dependency to deb package.

matt2000:
I ran into this while testing a playbook in a Docker container using ubuntu/utoptic as the base.

### DIFF
--- a/packaging/debian/control
+++ b/packaging/debian/control
@@ -3,7 +3,7 @@ Section: admin
 Priority: optional
 Standards-Version: 3.9.3
 Maintainer: Michael DeHaan <michael.dehaan@gmail.com>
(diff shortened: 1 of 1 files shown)


matt2000:
This is actually a bug-fix. The deb package will fail to install on some systems without this change.

bcoca:
which systems? how can i reproduce? it is weird to need ssh-client as a build depenency

matt2000:
I can't reproduce this with a 'hello world' playbook, so it must be specific to the original playbook I was using, and so not a problem with ansible itself.

bcoca:
well, the file updated is part of the deb package build source, it would not affect playbooks at all, only the building of the .deb from the ansible source

PR #10176, opened 2015-02-06
add scaffolding for serverspec integration testing

dhutty:
This adds an option, --serverspec, to create scaffolding to support integration testing with kitchen/serverspec as described here: https://github.com/neillturner/kitchen-ansible.

### DIFF
--- a/bin/ansible-galaxy
+++ b/bin/ansible-galaxy
@@ -130,6 +130,40 @@ Author Information
 An optional section for the role authors to include contact information, or a website (HTML is not allowed).
 """
 
+default_wrapper_playbook = """
+---
+- name: "{{ rolename }} wrapper playbook for integration testing with kitchen/serverspec"
+  hosts: localhost
+  roles:
+    - {{ rolename }}
+
+"""
+
+default_spec = """
+require 'spec_helper'
+
+# if os[:family] == 'ubuntu'
+#   describe '/etc/lsb-release' do
+#     it "exists" do
+#         expect(file('/etc/lsb-release')).to be_file
+#     end
+#   end
+# end
+# if os[:family] == 'redhat'
+#   describe '/etc/redhat-release' do
+#     it "exists" do
+#         expect(file('/etc/redhat-release')).to be_file
+#     end
+#   end
+# end
+"""
+
+spec_helper = """
+require 'serverspec'
+
+set :backend, :exec
+"""
+
 #-------------------------------------------------------------------------------------
 # Utility functions for parsing actions/options
 #-------------------------------------------------------------------------------------
@@ -176,6 +210,9 @@ def build_option_parser(action):
             '-p', '--init-path', dest='init_path', default="./",
             help='The path in which the skeleton role will be created. '
                  'The default is the current working directory.')
+        parser.add_option(
+            '--serverspec', dest='serverspec', default=False, action='store_true',
+            help="Create scaffolding for serverspec integration testing")
         parser.add_option(
             '--offline', dest='offline', default=False, action='store_true',
             help="Don't query the galaxy API when creating roles")
@@ -587,6 +624,7 @@ def execute_init(args, options, parser):
     api_server = get_opt(options, "api_server", "galaxy.ansible.com")
     force      = get_opt(options, 'force', False)
     offline    = get_opt(options, 'offline', False)
+    serverspec = get_opt(options, 'serverspec', False)
 
     if not offline:
         api_config = api_get_config(api_server)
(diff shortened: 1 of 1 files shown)


sivel:
Since serverspec is not really the recommended way of testing ansible, I'm going to have to say -1 on this.

http://docs.ansible.com/test_strategies.html

dhutty:
This is an _option_, off by default, that allows people to take advantage of large bodies of existing code/experience when migrating from other CMs. One of the key advantages of tools as flexible as Ansible is that they allow various different approaches to meeting needs, rather than highly opinionated tools that insist on 'my way or the highway'.

mscherer:
The option do not seems to do much that would requires from being in ansible main script, and could be done by a external script/playbook in contribs, no ?

dhutty:
1) It's not for the ansible main script, it's for ansible-galaxy.
2) Yes, it could be done an external script or playbook, of course - almost anything could; that's the nature of general purpose tools. 

Creating scaffolding for new roles (as referred to in http://docs.ansible.com/galaxy.html#building-out-role-scaffolding) is the job of ansible-galaxy(1).

Since this patch is to optionally modify the new role created by that command, it makes sense to put it into the code which creates new roles: `ansible-galaxy init`. Yes, it would be possible to create another script or playbook to create scaffolding, but that would be repeating functionality that already exists.

sgargan:
+1 for test-kitchen. Its almost essential toolkit for me, in the absence of something more ansible oriented it is my go-to for role testing. There are many roles in galaxy that could benefit from the robustness test-kitchen and serverspec can bring to them, having the support to help folks along the right path from the outset would not be a bad thing. It does not need to replace current testing strategies as it is complimentary. @sivel all this needs to be recommended is for someone to update that page, I think that should happen.

sivel:
I'm still a big -1

sgargan:
@sivel why so adamant? Would you agree there is a space for an integration platform that test-kitchen currently fills. Serverspec I'm ambivalent about but automation that test-kitchen allows is excellent, I highly recommend it. Maybe outline your objections?

bcoca:
These seem to be more for code testing than anything else, I find server spec a lot more work than the actual plays or roles that it means to test.

mscherer:
So the question is more "do we want to support different templates for a role generated by ansible-galaxy". On one hand, I can see why people want to have the tool of their choice to be supported, but on the other hand, I can see also why we want to have 1 single way that we focus on, rather than going in every direction.

bcoca:
I'm also torn here, I would like to make it easy for anyone to use the test framework of their choice, but I'm not sure this code belongs in ansible.

dhutty:
Sidenote: Serverspec is actually offered as an alternative by "Learning Ansible" book. I realise that this is not Ansible, Inc.'s recommendation (nor even the Ansible community's, but still).

@bcoca _You_ might find it more difficult. But some people already have large existing bodies of serverspec code. Large enterprises will often be using multiple CM systems, this just makes it easier to benefit and reduces the requirement to repeat effort. Not to mention that for those with QA teams and CI processes based in Ruby/Rspec/Serverspec.

@sivel perhaps you're coming from the position that there should be one specific, preferred approach to testing, which sounds very PEP-20, but it's rather at odds with the extreme flexibility that Ansible provides.

As @sgargan noted, there are reasons for easing the process of integration testing and while it would be nice if there were something in the python world that would be more suitable than serverspec, I'm not aware of anything.

But the **_main point**_ is that the workflow of "kitchen create && kitchen converge && kitchen verify" is extremely powerful/convenient. This patch just makes role-writing a little easier for those taking that approach.

@bcoca and as to whether the scaffolding belongs in Ansible itself, this is a patch against ansible-galaxy(1) not ansible(1). It seems slightly odd to me that `ansible-galaxy init` behaviour is built-in. But given that galaxy creates scaffolding, it seems obvious that this is the correct place to modify that scaffolding.

mscherer:
After thinking a while, i would rather see done outside of ansible. If the testing tools best practice evolve, it has to evolve has well, so to me, this belong on the testing tool side more than ansible side. Ansible ( ie, using ansible to speak of the project, we got the point of "modify ansible-galaxy script" ) will take care to generate a working initial role for the current ansible version, while the testing tool will generate proper file for himself.

dhutty:
I think perhaps I should have named the topic branch 'kitchen support':)

Would you be more interested if I were to add default functionality that allows the test-kitchen workflow, but instead of creating serverspec scaffolding, created assert-based testing in an Ansible playbook?

bcoca:
soo, after thinking for a while let me shoot this out to see if it makes sense. Instead of hardcoding all this in side ansible-galaxy, have it use external templates, with configurable overrides/additions.

This would keep the actual test scaffolding outside of ansible, but make it easy for people to drop it in their installations and have it integrated into their roles. 

Does this make sense?

dhutty:
You mean something like: `ansible-galaxy init --boilerplate_dir <path>` where the directory is a hierarchy of jinja2 templates that create custom role boilerplate?

Perhaps:
- without the option it behaves as now
- with the option it looks in ~/.ansible/role_template
- with the option and a value, it looks in the directory specified

mscherer:
I think it should definitely combine the regular directory from ansible, with the addition of another boilerplatedir. How can we express to not use a file ?

bcoca:
So in v2 I've made the scaffolding templates into external files, I'm open to adding an option to galaxy to add an external template dir to use when executing 'init'.

That said, this PR does not apply at all, I'll keep it open for another week to take in feedback but I'm marking it for closing as the code has changed drastically and we still would not want the specs directly included in ansible.

@mscherer not sure what you mean by 'not use a file'

mscherer:
I have no idea, this was 4 months ago, I can hardly remember what I did had in my plate at lunch :(

But I think that what I meant is that if we have the possibility of adding/overriding a file from the scaffolding template, we should also have a option to suppress a file. Ie, adding is easy, we could just take all files from all directory with 1 rule to decide of the priority ( kinda like systemd ). Now, combining different directories and having a file be removed is not as easy.

dhutty:
This is obsoleted by molecule

PR #10178, opened 2015-02-06
Obfuscate passwords in a variety of output

### LINKED ISSUE #10101: Subversion password shows in failure message
When using password="This is a test" and a checkout fails, following error message is shown

failed: [webserver -> localhost] => {"cmd": "/usr/bin/svn --non-interactive --trust-server-cert --no-auth-cache --username testuser --password *******\* is a test' info /tmp/checkout/", "failed": true, "rc": 1}
stderr: svn: E155007: '/tmp/checkout' is not a working copy

abadger:
Fixes #10101 

Did some testing with the following playbook:

```
- hosts: localhost
  tasks:
    - git:
        repo: https://abadger:asdksadksadd@examppppp.com/faassad/fake/
        dest: /var/tmp/checkout
        update: yes
      ignore_errors: True
    - subversion:
        password: "this is a test"
        dest: '/var/tmp/checkout'
        repo: '/var/tmp/nonexist'
      ignore_errors: True
```

and the following commandline:

```
ansible-playbook svn-passwd.yml -vvvv
```

Found several sources of output that did not identify and hide password strings.  This Pull request should fix that for the v1 tree.  Some questions to answer before merging:
- Should I be obfuscating passwords in stdout and stderr output (when calling fail_json() from basic.py AnsibleModule._run_command())?  obfuscation can be overzealous in some cases and so I don 't know if this would end up making debugging hard in some cases.  no_log=True will hide the stdout and stderr output if we decide not to hide stdout/stderr.
- I couldn't find password obfuscation code in the v2 tree.  Inspecting the output shows that there is a little bit of password obfuscation occurring in v2 but it isn't complete.  We probably need to at least modify basicpy run_command() in v2 as well and there may be other places.  There may also be more logging/displaying code yet to go into v2 and those would have to be checked for unobfuscated passwords afterwards as well.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -171,6 +171,7 @@ FILE_COMMON_ARGUMENTS=dict(
     directory_mode = dict(), # used by copy
 )
 
+PASSWD_ARG_RE = re.compile(r'^[-]{0,2}pass[-]?(word|wd)?')
 
 def get_platform():
     ''' what's the platform?  example: Linux is a platform. '''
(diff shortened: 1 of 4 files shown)


abadger:
- Decided that stdout and stderr should not be included in password obfuscation.  Making debugging harder is not desirable and stdout/stderr output could be very large and scanning for passwords could slow this down too much.  if there's places where stdout/stderr are a concern, then using no_log: true is the right thing to do.
- Password obfuscation in v2 is currently only implemented in basic.py so I'll port that over now.  The other places in v1 where password obfuscation occurs will happen when we port those pieces over to v2.

PR #10183, opened 2015-02-09
adds complex tag management

bcoca:
Adds a special tag:
- always: always runs no matter what --tags/--skip-tags says

Adds 4 special keywords for --tags
- all: all tagged + untagged tasks
- tagged: only tagged tasks
- untagged: only untagged tasks
- always: only run tasks tagged 'always'

I think this covers all cases described in various tickets asking for more fine control over tags

### DIFF
--- a/lib/ansible/playbook/__init__.py
+++ b/lib/ansible/playbook/__init__.py
@@ -753,7 +753,7 @@ class PlayBook(object):
                             if task_set == u:
                                 should_run = True
                         else:
-                            if len(task_set.intersection(self.only_tags)) > 0:
+                            if task_set.intersection(self.only_tags):
                                 should_run = True
 
                     # Check for tags that we need to skip
@@ -768,7 +768,7 @@ class PlayBook(object):
                                 should_run = False
                         else:
                             if should_run:
-                                if len(task_set.intersection(self.skip_tags)) > 0:
+                                if task_set.intersection(self.skip_tags):
                                     should_run = False
 
                 if should_run:
--- a/lib/ansible/playbook/__init__.py
+++ b/lib/ansible/playbook/__init__.py
@@ -672,7 +672,53 @@ class PlayBook(object):
         return filename
 
     # *****************************************************
+    def tasks_to_run_in_play(self, play):
 
+        tasks = []
+
+        for task in play.tasks():
+            # only run the task if the requested tags match or has 'always' tag
+            u = set(['untagged'])
+            task_set = set(task.tags)
+
+            if 'always' in task.tags:
+                should_run = True
+            else:
+                if 'all' in self.only_tags:
+                    should_run = True
+                else:
+                    should_run = False
+                    if  'tagged' in self.only_tags:
+                        if task_set != u:
+                            should_run = True
+                    elif 'untagged' in self.only_tags:
+                        if task_set == u:
+                            should_run = True
+                    else:
+                        if task_set.intersection(self.only_tags):
+                            should_run = True
+
+            # Check for tags that we need to skip
+            if 'all' in self.skip_tags:
+                should_run = False
+            else:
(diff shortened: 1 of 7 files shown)


Ernest0x:
> always: only run tasks tagged 'always'

I think it should be:
- always: always run tasks tagged with "always" whatever the '--tags' command-line option is set (or not set) to

brian-brazil:
Taking a quick look based on by work on #7039:
- The tag code generally needs a refactor to correctly handle --list-tasks, as the code is currently duplicated incorrectly
- `always` should also bypass --start-at
-  There should be checks that tasks aren't using reserved tags
-  Unittests should be added

I'd suggest taking my PR, and adding the additional functionality on top of it as it handles the above already.

lib/ansible/playbook/__init__.py
```diff
@@ -730,19 +735,41 @@ def _run_play(self, play):
...
+                    else:
+                        if 'tagged' in self.skip_tags:
+                            if task_set != u:
+                                should_run = False
+                        elif 'untagged' in self.skip_tags:
+                            if task_set == u:
+                                should_run = False
+                        else:
+                            if should_run:
+                                if len(task_set.intersection(self.skip_tags)) > 0:
```
brian-brazil: You can do:

```
if task_set.intersection(self.skip_tags):
```

arnuschky:
This covers my use case, but the `always` keyword/tag would suffice. 

What does the `all` keyword to over omitting `--tags` altogether?

For me, `tagged` and `untagged` are only marginally interesting: I always tags everything, so `untagged` might only help in debugging.

hkariti:
One thing I see missing here that has been requested in the past is the ability to AND multiple tags. Something like: `--tags conf:&app1`

mscherer:
So, what would "--skip-tags always" do, and what would --skip-tags "all" do ? 

The second seems like a no-op, but the first is more puzzling. If it skip always, then always is not always. If if doesn't skip anything, then that's a exception to the rule.

brian-brazil:
`--skip-tags always` should not run the always tags. There are disaster recovery cases where you may not wish some `always` tasks to run (e.g. some backend you need is down, causing the task to fail), and this is one way to handle that.

lib/ansible/playbook/__init__.py
bcoca: good point, I'll update

bcoca:
current state of this request:
- "--tags all" is the actual default if you omit --tags, so it is also a noop when specified (this is true w/o  patch)
- "--skip-tags always" doesn't do anything, things tagged 'always' cannot be skipped
- "--skip-tags all" will only execute tasks with 'always'
- "--tags always" is basically the same as "--skip-tags all"
- --start-at-task is not tag aware, imho it should not be
- --skip-tags overrides --tags (this is also true w/o patch)

I left out tests,list-tasks and task expressions (matching host expressions) as I wanted to nail down the main functionality first. The 2 points of discord I still see are that some don't want 'always' to be always and the inclusion of --start-at-task in tag management.

brian-brazil:
> the inclusion of --start-at-task in tag management.

I don't think there's discord here. Only @tkinz27 and I have mentioned this on #7039, and both of us appear to want `always` to override --start-at-task.

bcoca:
@brian-brazil copied your tests (I hope you don't mind) and copied your use of a method to reuse task selection in ansible-playbook

I also changed 'always' tag to be skippable

I'm not adding star-at-task functionality nor tag pattern matching in this round, leaving those for future PRs

brian-brazil:
I'm happy to get this feature in, so please take whatever is useful from my PR. 

You may want to pull in the deletion of `_is_valid_tag` in lib/ansible/playbook/play.py, as that's dead code that would cause confusion given the new code.

sevein:
Nice!

PR #10198, opened 2015-02-10
Add issue_tracker_url to the galaxy template

emonty:
There is an optional issue_tracker_url setting that can be set in
meta/main.yml:galaxy_info. Include it in the generated template
commented out so that it can be filled out if desired.


jimi-c:
Merged, thanks!

PR #10161, opened 2015-02-03
Adds Kerberos authentication to winrm if available

kstrauser:
If the `kerberos` module is available, winrm will attempt to establish a
Kerberized connection to a Windows server. This allows use of Windows
domain accounts, which are quite often the only kinds of accounts
enabled on enterprise networks.

This also pulls the `transport_schemes` variable up into the
`winrm.Connection` class. This lets tests or future modifications alter
the list of available schemas without reaching into
`Connection.__init__`.

PR #10202, opened 2015-02-10
Adds Kerberos authentication to winrm if available

kstrauser:
If the kerberos module is available, winrm will attempt to establish a Kerberized connection to a Windows server. This allows use of Windows domain accounts, which are quite often the only kinds of accounts enabled on enterprise networks.

This also pulls the `transport_schemes` variable up into the `winrm.Connection` class. This lets tests or future modifications alter the list of available schemas without reaching into `Connection.__init__`.

(Identical to #10161, but now with moar devel.)

### DIFF
--- a/lib/ansible/runner/connection_plugins/winrm.py
+++ b/lib/ansible/runner/connection_plugins/winrm.py
@@ -37,6 +37,13 @@ try:
 except ImportError:
     raise errors.AnsibleError("winrm is not installed")
 
+HAVE_KERBEROS = False
+try:
+    import kerberos
+    HAVE_KERBEROS = True
+except ImportError:
+    pass
+
 _winrm_cache = {
     # 'user:pwhash@host:port': <protocol instance>
 }
@@ -46,6 +53,11 @@ def vvvvv(msg, host=None):
 
 class Connection(object):
     '''WinRM connections over HTTP/HTTPS.'''
+
+    transport_schemes = {
+        'http': [('kerberos', 'http'), ('plaintext', 'http'), ('plaintext', 'https')],
+        'https': [('kerberos', 'https'), ('plaintext', 'http'), ('plaintext', 'https')],
+        }
 
     def __init__(self,  runner, host, port, user, password, *args, **kwargs):
         self.runner = runner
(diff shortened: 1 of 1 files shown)


jhawkesworth:
Very glad to see this merged but concerned its not documented that you'd need to add kerberos module (this is based on previous experience where lots of people are familiar with the notion of a windows domain login and don't necessarily know that underneath its actually based on kerberos).
Happy to tackle a documentation PR but thought I'd ask in case it was pending.

-- cheers,
Jon

kstrauser:
That’s a great point, and I confess that I hadn’t started the doc PR yet. Let me know if you’d rather you or I do it.

bcoca:
I was thinking that if the transport is kerberos, instead of a pass when
the library is missing it should raise an AnsibleError describing the issue.

## 

Brian Coca

lib/ansible/runner/connection_plugins/winrm.py
```diff
@@ -72,11 +84,10 @@ def _winrm_connect(self):
...
             vvvv('WINRM REUSE EXISTING CONNECTION: %s' % cache_key, host=self.host)
             return _winrm_cache[cache_key]
-        transport_schemes = [('plaintext', 'https'), ('plaintext', 'http')] # FIXME: ssl/kerberos
-        if port == 5985:
-            transport_schemes = reversed(transport_schemes)
         exc = None
-        for transport, scheme in transport_schemes:
+        for transport, scheme in self.transport_schemes['http' if port == 5985 else 'https']:
+            if transport == 'kerberos' and not HAVE_KERBEROS:
+                continue
```
bcoca, line 90: this should raise an exception telling the user they need the required library

kstrauser, line 90: The thinking behind not doing that earlier was that it would retain backward compatibility with existing installations that don't have kerberos.py installed. With that change, since the `kerberos` is in both transport schemes and will always be tried, all Windows admins would start getting exceptions until they install kerberos.

kstrauser:
As an alternative, what if `HAVE_KERBEROS` is False and all transport schemes fail, then raise an exception suggesting that they install kerberos if they need domain account authentication?

bcoca:
that sounds as the best alternative

kstrauser:
Something like 2587c9cdc865218340bfcc3e1bf19f44fd08ddd5?

jhawkesworth:
Please go ahead and document it...
This looks like a good place to mention installing kerberos 
http://docs.ansible.com/intro_windows.html#installing-on-the-control-machine

Well worth reading through this thread as there's some useful info about domain / kerberos support in it too:

https://groups.google.com/forum/#!searchin/ansible-devel/kerberos/ansible-devel/TYMdAoN9IK8/7oaNzpHj5UAJ

kstrauser:
@jhawkesworth Will do.

kstrauser:
@jhawkesworth See #10256.

PR #10203, opened 2015-02-10
Add HP-UX network facts

pdelared:
Added support for HPUX network facts

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -2048,6 +2048,57 @@ class GenericBsdIfconfigNetwork(Network):
             for item in ifinfo[ip_type][0].keys():
                 defaults[item] = ifinfo[ip_type][0][item]
 
+class HPUX(Network):
+    """
+    HP-UX-specifig subclass of Network. Defines networking facts:
+    - default_interface
+    - interfaces (a list of interface names)
+    - interface_<name> dictionary of ipv4 address information.
+    """
+    platform = 'HP-UX'
+
+    def __init__(self, module):
+        Network.__init__(self, module)
+
+    def populate(self):
+        netstat_path = self.module.get_bin_path('netstat')
+        if netstat_path is None:
+            return self.facts
+        self.get_default_interfaces()
+        interfaces = self.get_interfaces_info()
+        self.facts['interfaces'] = interfaces.keys()
+        for iface in interfaces:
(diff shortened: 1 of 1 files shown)


lib/ansible/module_utils/facts.py
```diff
@@ -2048,6 +2048,57 @@ def merge_default_interface(self, defaults, interfaces, ip_type):
...
+            return self.facts
+        self.get_default_interfaces()
+        interfaces = self.get_interfaces_info()
+        self.facts['interfaces'] = interfaces.keys()
+        for iface in interfaces:
+                self.facts[iface] = interfaces[iface]
+        return self.facts
+
+    def get_default_interfaces(self):
+        rc, out, err = module.run_command("/usr/bin/netstat -nr", use_unsafe_shell=True)
```
abadger, line 2075: What's the reason for using the shell here?  There's no redirection, pipes, or other needs for the shell that I can see.

lib/ansible/module_utils/facts.py
```diff
@@ -2048,6 +2048,57 @@ def merge_default_interface(self, defaults, interfaces, ip_type):
...
+        for line in lines:
+                words = line.split()
+                if len(words) > 1:
+                    if words[0] == 'default':
+                        self.facts['default_interface'] = words[4]
+                        self.facts['default_gateway'] = words[1]
+
+    def get_interfaces_info(self):
+        interfaces = {}
+        rc, out, err = module.run_command("/usr/bin/netstat -ni", use_unsafe_shell=True)
```
abadger, line 2086: Same question here about unsafe_shell as in get_default_interfaces()

abadger:
I don't have any HP-UX machines so I'll have to trust that the code works :-)  I just have a few questions about the idioms in you're using in the code.

lib/ansible/module_utils/facts.py
pdelared, line 2075: I based my code on class HP UX(Hardware). I'm not a python expert. Feel free to correct my bad idioms an i will do the test on HP UX box

lib/ansible/module_utils/facts.py
pdelared, line 2086: same as above

amenonsen:
AFAICS, the needs_info request has been answered, and the PR is ready to merge with use_unsafe_shell=False in two places and the spelling correction noted above.

PR #10217, opened 2015-02-11
Tasks use do-until by default

git-harry:
By default if a task fails ansible does not retry the task. Ansible
tasks can be dependent on many moving parts and so transient errors are
common. Do-until loops can be used to retry tasks and must be specified
for each task individually.

Retrying a task is such a commonly used feature it would be useful to be
to have this enabled by default.

This patch adds the ability to enable do-until loops on all tasks by
default. This feature is enabled with the constant
DEFAULT_DO_UNTIL_RETRIES_ENABLED.

Three additional constants are added:

```
DEFAULT_DO_UNTIL_RETRIES
DEFAULT_DO_UNTIL_DELAY
DEFAULT_DO_UNTIL_CONDITION
```

DEFAULT_DO_UNTIL_CONDITION is set to '{register}|success' by default
where '{register}' is used as a placeholder for the tasks registered
variable.

A do-until loop can be disabled on a task with 'until: null'.


amenonsen:
I don't like the idea of having _such_ a dramatic change to control-flow be configurable; that it comes with no documentation changes makes it completely unacceptable, IMHO. Besides which, the code has changed and the patch doesn't apply. I think this should be closed.

bcoca:
actually in 2.0 you can make this a strategy plugin, much cleaner and definable by play.

closing this PR as it does not apply to the current code which has changed drastically.

PR #10222, opened 2015-02-11
made sure passwords are strings

### LINKED ISSUE #10166: ansible_sudo_pass for numeric passwords
Hi, there is a bug with python types, when we use numeric passwords in inventory file:

```
  File "/usr/lib/pymodules/python2.7/ansible/runner/connection_plugins/ssh.py", line 352, in exec_command
    stdin.write(self.runner.sudo_pass + '\n')
TypeError: unsupported operand type(s) for +: 'int' and 'str'
```

`self.runner.sudo_pass` interpreted as integer and concatenation fails. :)

bcoca:
Also strip vault input for usage as it is for encryption
should fix #10166

### DIFF
--- a/lib/ansible/utils/__init__.py
+++ b/lib/ansible/utils/__init__.py
@@ -33,6 +33,7 @@ from ansible.utils.hashing import secure_hash, secure_hash_s, checksum, checksum
 from ansible.callbacks import display
 from ansible.module_utils.splitter import split_args, unquote
 from ansible.module_utils.basic import heuristic_log_sanitize
+from ansible.utils.unicode import to_bytes, to_unicode
 import ansible.constants as C
 import ast
 import time
@@ -1088,9 +1089,9 @@ def ask_vault_passwords(ask_vault_pass=False, ask_new_vault_pass=False, confirm_
 
     # enforce no newline chars at the end of passwords
     if vault_pass:
-        vault_pass = vault_pass.strip()
+        vault_pass = to_bytes(vault_pass, errors='strict', nonstring='simplerepr').strip()
     if new_vault_pass:
-        new_vault_pass = new_vault_pass.strip()
+        new_vault_pass = to_bytes(new_vault_pass, errors='strict', nonstring='simplerepr').strip()
 
     return vault_pass, new_vault_pass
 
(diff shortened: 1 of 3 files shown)


abadger:
- Have to import to_bytes from ansible.utils.unicode
- Probably want to raise an exception if we aren't able to convert the password into a byte string for some reason -- otherwise we could end up with the user encrypting a file and not being able to decrypt it.
- May want to be explicit about nonstring=simplerepr.  It is the default of the function but it will help future programmers who read the code to realize that one of the purposes of the code is to handle values that aren't strings.

So the function call will look like this:

``` python
to_bytes(vault_pass, errors='strict', nonstring='simplerepr').strip()
```

bcoca:
i was referencing the to_bytes function in the same file, I was not aware of the same named function in unicode.py, do we want to make both functions the same?

abadger:
yes, it would be great to have gotten rid of all other to_bytes() and to_unicode() functions by the time we release the v2 codebase.  Taking a look, the to_bytes() inside of that file won't be able to handle int values like the to_bytes() in ansible/utils/unicode.py can.

Sorry for the confusion.

test/units/TestUtils.py
```diff
@@ -512,15 +512,15 @@ def test_make_su_cmd(self):
...
 
     def test_to_unicode(self):
-        uni = ansible.utils.to_unicode(u'ansible')
+        uni = ansible.utils.unicode.to_unicode(u'ansible')
         self.assertTrue(isinstance(uni, unicode))
         self.assertEqual(uni, u'ansible')
 
-        none = ansible.utils.to_unicode(None)
+        none = ansible.utils.unicode.to_unicode(None)
         self.assertTrue(isinstance(none, type(None)))
```
abadger: This one will fail with the new to_unicode.   With no arguments, this will return `"None"``(the equivalent of``str(none)`` ).  If we want the test to return None, then update like this:

``` python
none = ansible.utils.unicode.to_unicode(None, nonstring=passthru)
```

lib/ansible/utils/__init__.py
```diff
@@ -1104,18 +1105,26 @@ def ask_passwords(ask_pass=False, ask_sudo_pass=False, ask_su_pass=False, ask_va
...
         sudopass = getpass.getpass(prompt=sudo_prompt)
         if ask_pass and sudopass == '':
             sudopass = sshpass
+        if sudopass:
+            sudopass = to_bytes(sudopass, errors='strict', nonstring='simplerepr')
 
     if ask_su_pass:
         su_pass = getpass.getpass(prompt=su_prompt)
+        if supass:
+            supass = to_bytes(supass, errors='strict', nonstring='simplerepr')
```
abadger: `su_pass` is used in the getpass but `supass` without the underscore is used here.  Probably needs to be `su_pass` in both places.

test/units/TestUtils.py
bcoca: changed, though i think it should default to that

lib/ansible/utils/__init__.py
bcoca: made them all match same convention ... having 2 is what creates typoes!

abadger:
+1 to merge

renard:
this returns an error:

```
  File "/home/renard/Developer/ansible/ansible/bin/ansible-playbook", line 334, in <module>
    sys.exit(main(sys.argv[1:]))
  File "/home/renard/Developer/ansible/ansible/bin/ansible-playbook", line 135, in main
    (sshpass, sudopass, su_pass, vault_pass) = utils.ask_passwords(ask_pass=options.ask_pass, ask_sudo_pass=options.ask_sudo_pass, ask_su_pass=options.ask_su_pass, ask_vault_pass=options.ask_vault_pass)
  File "/home/renard/Developer/ansible/ansible/lib/ansible/utils/__init__.py", line 1132, in ask_passwords
    return (sshpass, sudopass, supass, vaultpass)
UnboundLocalError: local variable 'supass' referenced before assignment
```

Please replace:
- `supass` by `su_pass`
- `vaultpass` by `vault_pass`

in `ask_passwords` from `lib/ansible/utils/__init__.py`

Did you test the patch before merge?

renard:
Ok both 932fd2e and 0087816 fixed the issue. Thanks

PR #10229, opened 2015-02-12
patch action plugin does not require dest

### LINKED ISSUE #258: patch module is requiring both src and dest
`runner/action_plugins/patch.py` requires both `src` and `dest`, but `modules/extras/files/patch.py` declares `dest` as not required.

bcoca:
dest is not required as patchfile can have that info and the module did not require it
should fix https://github.com/ansible/ansible-modules-extras/issues/258

### DIFF
--- a/lib/ansible/runner/action_plugins/patch.py
+++ b/lib/ansible/runner/action_plugins/patch.py
@@ -34,8 +34,8 @@ class ActionModule(object):
         dest = options.get('dest', None)
         remote_src = utils.boolean(options.get('remote_src', 'yes'))
 
-        if src is None or dest is None:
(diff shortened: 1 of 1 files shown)


lib/ansible/runner/action_plugins/patch.py
```diff
@@ -34,8 +34,8 @@ def run(self, conn, tmp, module_name, module_args, inject, complex_args=None, **
         dest = options.get('dest', None)
```
abadger, line 34: Note: Looks like setting dest here can be removed as well.

abadger:
Looks good.  +1 to merge.

PR #10231, opened 2015-02-13
'aformat' Jinja2 filter

tavlima:
This is a simple 'alternative format' filter, which comes in hand
when mapping a data formatting function over a list.

eg. suppose we have a list like

``` yaml
userlist:
  - firstname: John
    surname: Foo
  - firstname: Paul
    surname: Bar
```

we could format + join their names, in a template or playbook, using
the following construct:

``` jinja
{{ userlist|map('aformat', '%(firstname)s %(surname)s')|join(', ') }}
```

### DIFF
--- a/lib/ansible/runner/filter_plugins/core.py
+++ b/lib/ansible/runner/filter_plugins/core.py
@@ -306,6 +306,9 @@ def get_encrypted_password(password, hashtype='sha512', salt=None):
 def to_uuid(string):
     return str(uuid.uuid5(UUID_NAMESPACE_ANSIBLE, str(string)))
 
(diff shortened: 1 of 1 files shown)


amenonsen:
I suppose this might be useful, but (a) what is alternative about this format? (b) it needs documentation changes. @tavlima do you have any plans to update to apply to current devel? It shouldn't be too hard, it just needs to target `plugins/filters/core.py` and documentation in `playbooks_filters.rst`.

sivel:
@tavlima,

Hi!

amenonsen:
Closing per above comment.

PR #10232, opened 2015-02-13
Support SCM URLs containing spaces (for branch specifiers)

skitt:
It's become common practice to specify branches to clone from git repositories using -b, as in

```
git://my.git.server/repository.git -b branch
```

ansible-galaxy doesn't support this as is, but enabling it is easy: all it takes is to split the role URL. This comes at the expense of supporting URLs containing unescaped spaces, but that's probably not an issue...

### DIFF
--- a/bin/ansible-galaxy
+++ b/bin/ansible-galaxy
@@ -347,7 +347,7 @@ def scm_archive_role(scm, role_url, role_version, role_name):
         print "- scm %s is not currently supported" % scm
         return False
     tempdir = tempfile.mkdtemp()
-    clone_cmd = [scm, 'clone', role_url, role_name]
+    clone_cmd = [scm, 'clone'] + role_url.split() + [role_name]
     with open('/dev/null', 'w') as devnull:
         try:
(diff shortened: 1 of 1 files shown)


sivel:
Isn't this already supported by using the branch as the version?

skitt:
Unfortunately not, because the git repository is freshly cloned, so `git archive` can't find the branch that it's given if it's not already been checked out. For `git archive` to work in this scenario you need to do

```
git clone ...
git checkout branch
```

Once this is done, `git archive` will work from any branch; so for example

```
git checkout master
git archive -o ... branch
```

works fine. But simply doing

```
git clone ...
git archive -o ... branch
```

fails if `branch` isn't the default branch:

```
fatal: Not a valid object name
```

sivel:
My recommendation would be to fix it so the current functionality works for this use case, instead of supporting something different.

skitt:
OK, thanks for taking the time to discuss this.

BTW it turns out there is a workaround for this: specifying `remotes/origin/branch` instead of `branch` allows `git archive` to find the branch.

PR #10237, opened 2015-02-13
Remove auth token and port openstack module_utils changes to v2 tree

emonty:


emonty:
@bcoca we should pop this one in around the same time as the ansible-modules-extras one. I just read through the outstanding PRs on the old modules, and there is one feature someone wanted that's really good that will need this and the corresponding change there (which is already pushed up) After that, it's a piece of cake to support.

bcoca:
my worry here is that this will break existing modules, even if deprecated they should still work

emonty:
@bcoca the existing deprecated modules do not use this function, they use the other one.

lib/ansible/utils/module_docs_fragments/openstack.py
```diff
@@ -69,12 +60,35 @@ class ModuleDocFragment(object):
       - How long should ansible wait for the requested resource.
     required: false
     default: 180
+  api_timeout:
+    description:
+      - How long should the socket layer wait before timing out for API calls.
+        If this is omitted, nothing will be passed to the requests library.
+    required: false
+  verify:
```
bcoca: is this the same as "validate_certs" in other modules? if so I would prefer to standardize the naming to that

lib/ansible/utils/module_docs_fragments/openstack.py
```diff
@@ -69,12 +60,35 @@ class ModuleDocFragment(object):
       - How long should ansible wait for the requested resource.
     required: false
     default: 180
+  api_timeout:
+    description:
+      - How long should the socket layer wait before timing out for API calls.
+        If this is omitted, nothing will be passed to the requests library.
+    required: false
```
bcoca, line 67: non required fields should show default, even if just null

bcoca:
aside from the 2 comments everything else looks good

PR #10238, opened 2015-02-13
make ssh plugin resilient against invalid entries in hosts file

bcoca:
should mitigate issue discussed here:
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/ansible-project/0yAgEcva9jI/2ePHw3l48MQJ


amenonsen:
@bcoca this patch still seems relevant and desirable, but needs to be rebased. Do you want me to revise and resubmit, or will you take care of it?

PR #10242, opened 2015-02-14
Enable sudo mode if user explicitely ask for sudo password

mscherer:
Since using ansible -K without --sudo will not result into anything
useful and since we are doing this already for the --ask-su-pass
option, we should as well be consistent on it. A user came on the
irc today with the exact problem of having -K without the required
options.

### DIFF
--- a/bin/ansible
+++ b/bin/ansible
@@ -143,6 +143,10 @@ class Cli(object):
 
         if options.su_user or options.ask_su_pass:
             options.su = True
+
+        if options.ask_sudo_pass:
+            options.sudo = True
+
         options.sudo_user = options.sudo_user or C.DEFAULT_SUDO_USER
         options.su_user = options.su_user or C.DEFAULT_SU_USER
         if options.tree:


sivel:
The change to not imply sudo when supplying a password was done as a result of https://github.com/ansible/ansible/issues/6467 so -1 from me

mscherer:
I cannot reproduce the bug ( at least not by using ping ).

Now indeed, there is a issue of properly processing the intent of the user, and I do not think the cnfiguration file should result in forcing --sudo. Would having only -K ( ie, explicit command line option ) be acceptable ?

bcoca:
@mscherer closing this as sudo should not be forced on all hosts implicitly when have the program prompt for the password. 

The current implementation allows for the case in which you set ansible_sudo by host but not on all hosts, your change will remove this functionality. I understand this is a convenience for those that have homogeneous setups on their hosts, but we have to support those that have heterogeneous sudo setups also.

PR #10250, opened 2015-02-16
improve yaml parsing performance by ~25% by using PyYAML's CParser loader

cove:
This is with PyYAML-3.11, the CParser loader is a good bit faster than the python loader in my tests.

With SafeParser:
`% time ansible-playbook -l dummyNonConnectingHost site.yml --vault-password-file=~/.vault_pass > /dev/null
613.76s user 86.28s system 97% cpu 11:54.65 total`

With CSafeParser:
`% time ansible-playbook -l dummyNonConnectingHost site.yml --vault-password-file=~/.vault_pass > /dev/null
439.21s user 82.81s system 98% cpu 8:52.45 total`

### DIFF
--- a/lib/ansible/utils/__init__.py
+++ b/lib/ansible/utils/__init__.py
@@ -70,6 +70,11 @@ try:
 except ImportError:
     import json
 
+try:
+    from yaml import CSafeLoader as Loader
+except ImportError:
+    from yaml import SafeLoader as Loader
+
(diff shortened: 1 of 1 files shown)


abadger:
Thanks!   I think we had assumed that the yaml library was using the C extension if that was available but your benchmarks and a review of the code show that we have to be explicit about using it.

cove:
Great thanks. ;)

PR #10256, opened 2015-02-16
Added winrm/kerberos configuration information

kstrauser:
This documents using the `kerberos` module to use the control machine's TGT to authenticate to Windows domain accounts.

### DIFF
--- a/docsite/rst/intro_windows.rst
+++ b/docsite/rst/intro_windows.rst
@@ -28,6 +28,12 @@ On a Linux control machine::
 
    pip install http://github.com/diyan/pywinrm/archive/master.zip#egg=pywinrm
 
+If you wish to connect to domain accounts published through Active Directory (as opposed to local accounts created on the remote host)::
+
+   pip install kerberos
+
+Kerberos is installed and configured by default on OS X and many Linux distributions. If your control machine has not already done this for you, you will need to.
+
 .. _windows_inventory:
 
 Inventory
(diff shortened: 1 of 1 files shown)


trondhindenes:
I would argue that this can be clarified better. This: "This uses the principal you are authenticated to Kerberos with on the control machine and not the `ansible_ssh_user` specified above*. " The Windows doc is very step-by-step and this section should be too. If the user needs to get a Kerb ticket, the doc should indicate how that's done in order to keep the detailed-ness of the rest of the page.

kstrauser:
@trondhindenes I dither on this. On one hand, I understand and agree with your opinion that we should be very clear and give details. On the other, Kerberos administration is an _enormous_ subject and I'm not sure where to start or end.

If it were as simple as saying "type `kinit me@EXAMPLE.COM`" then I'd be 100% in favor of your comment (and would have already added it), but I suspect that users are largely going to fall into two camps:

1) Those who've already used Kerberos and know to run kinit with the appropriate principal, and therefore don't need this spelled out; and
2) Those who haven't used Kerberos and need to install it, set up their krb5.conf to point to their AD controller or KDC, etc. I think helping these people out is beyond the scope of what we want to document here.

If you have a favorite Kerberos guide, perhaps a link to it with "for more information, read..." might be appropriate?

trondhindenes:
I haven't done any work on Kerb on Linux at all, so I don't have any favorite guide I'm afraid. I'm just thinking that the primary audience for this section will be Windows admins wanting to use Kerb authentication, and anything we can do to at least help them in the right direction would be a big plus.

Issue #10253, opened 2015-02-16
ansible-vault edit does not properly retain file permissions


flibby-jibbit:
Issue Type:

Bug Report

Ansible Version:

$ ansible --version
ansible 1.9 (devel 2da35a55a5) last updated 2015/02/16 10:38:29 (GMT -500)
  lib/ansible/modules/core: (detached HEAD aa6c73f11b) last updated 2015/02/16 10:38:34 (GMT -500)
  lib/ansible/modules/extras: (detached HEAD 2ffd7235e5) last updated 2015/02/16 10:38:34 (GMT -500)
  v2/ansible/modules/core: (detached HEAD 34784b7a61) last updated 2015/02/12 16:14:10 (GMT -500)
  v2/ansible/modules/extras: (detached HEAD 312d59061c) last updated 2015/02/12 16:14:15 (GMT -500)
  configured module search path = None

Environment:

OS is Mac OS X 10.10.2. The user account is managed in an Active Directory domain. Not managing any remote hosts.

Summary:

`ansible-vault edit` does not retain the permissions of the file.

Steps To Reproduce:

``` bash
$ vim some-file
$ ls -l
total 4.0K
-rw-r--r-- 1 pkoebbe staff 14 Feb 16 10:39 some-file

$ ansible-vault encrypt some-file
Vault password:
Confirm Vault password:
Encryption successful

$ ls -l
total 4.0K
-rw-r--r-- 1 pkoebbe staff 355 Feb 16 10:39 some-file

$ ansible-vault edit some-file
Vault password:
```

Expected Results:

``` bash
$ ls -l
total 4.0K
-rw-r--r-- 1 pkoebbe staff 355 Feb 16 10:39 some-file
```

Actual Results:

``` bash
$ ls -l
total 4.0K
-rw------- 1 pkoebbe FLOAK\Domain Users 355 Feb 16 10:39 some-file
```

abadger:
# Is This A Bug?

Hi!

Thanks very much for your submission to Ansible.  It sincerely means a lot to us.

We're not sure this is a bug, and we don't mean for this to be confrontational.  Let's explain what we're thinking:
- The usage of vault files is more like a secure opaque data store rather than encrypting a file.  In that model ansible is doing the right thing to make the file permissions more secure.
  
[...]

PR #10260, opened 2015-02-16
Set proper file permissions (umask 077) in write_data

gtirloni:
So the file permissions remain consistent across all ansible-vault operations that use write_data().

Addresses issue #10253

### DIFF
--- a/lib/ansible/utils/vault.py
+++ b/lib/ansible/utils/vault.py
@@ -336,9 +336,16 @@ class VaultEditor(object):
     def write_data(self, data, filename):
         if os.path.isfile(filename): 
             os.remove(filename)
+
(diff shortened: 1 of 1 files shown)


amenonsen:
@gtirloni, the code has changed in devel and a umask save/restore such as you have suggested has been added to the `_edit_file_helper` method. I agree with you that this seems better placed in `write_data`, but if so your change should be rebased to target `lib/ansible/parsing/vault/__init__.py` and move the umask save/restore code from one place to another. But even if you don't do that, the original #10253 report has been addressed.

Either way, this particular PR can be closed.

gtirloni:
Thanks for the analysis. Closing PR.

PR #10265, opened 2015-02-17
Addition of LD_LIBRARY_PATH for custom python/openssl environments

sirkubax:
Following changes would allow for defining custom LD_LIBRARY_PATH 'per host' in inventory (or in ansible.config - default for all hosts)

That is important for environments with custom python/openssl settings

example:
[myhosts]
somehostname ld_library_path=/home/user/mycustomlib:/opt/openssl/lib:/usr/local/lib


bcoca:
why not just use the environment directive?

```
environment:
   LD_LIBRARY_PATH: "{{ldpath}}"
```

where ldpath can be defined globally/by group/by host

amenonsen:
I think this PR should be closed now. As @bcoca said, this can already be accomplished using the environment: mechanism.

bcoca:
closing as per above

PR #10266, opened 2015-02-17
rax_cbs: Remove explicit msg match for invalid size in integration tests

sivel:
This (small) PR, just removes a now invalid test assertion to validate the msg of a failed task when an invalid size was provided.  It's enough to just validate the failure.


abadger:
Thanks @sivel!

PR #10267, opened 2015-02-17
Add play-level environment support

craigtracey:
Right now ansible only provides task-level support for environment
variables.  This is quite useful for things like installing via
proxies, but a bit onerous when it needs to be applied to individual
tasks.

This change allows for play-level environment variable assignments.
The behavior is to accept environment variables from most specific
(tasks) over least specific (plays).


bcoca:
already done:
https://github.com/craigtracey/ansible/commit/ac4dc1f7

craigtracey:
Good catch @bcoca. Was working against an old branch and missed this. Thanks!

Issue #10165, opened 2015-02-05
Long delay in ansible-playbook, strace shows loop that stats non-existing module-files


jensdepuydt:
Issue type: bug report
Ansible version: 

```
[jendep@be-plw-ans-0001 ansible]$ ansible --version
ansible 1.8.2
  configured module search path = /etc/ansible/library
```

Environment:

```
[jendep@be-plw-ans-0001 ansible]$ cat /etc/centos-release
CentOS Linux release 7.0.1406 (Core)
[jendep@be-plw-ans-0001 ansible]$ uname -r
3.10.0-123.13.2.el7.x86_64
```

Summary:
We're using Ansible for quite a while now and for now we're very happy about it. 
Only one small thing is bothering me and it's that there's a delay between launching ansible-playbook and the actual process of deploying. For some reason, I suspected something related to SSH but finally I've made some time free to investigate the "issue" and I see with strace that there is a loop, searching for non-existing module files. After some debugging, I found out that the problem should be in "/usr/lib/python2.7/site-packages/ansible/playbook/play.py" but I can't seem to find what the exact problem is.

Steps to reproduce:
Execute ansible-playbook
For example:

```
[jendep@be-plw-ans-0001 ansible]$ /usr/bin/ansible-playbook -i /etc/ansible/inventory/test/hosts /etc/ansible/playbooks/test.yml -l be-plw-ans-0007 -K
```

Expected results:
no delay between prompt for password and start of playbook-deployment

Actual results:
long (=-15s) delay between prompt for password and start of playbook-deployment

Fragment from the strace-output when the delay appears:

```
open("/usr/lib64/python2.7/distutils/version.py", O_RDONLY) = 4
fstat(4, {st_mode=S_IFREG|0644, st_size=11433, ...}) = 0
open("/usr/lib64/python2.7/distutils/version.pyc", O_RDONLY) = 8
fstat(8, {st_mode=S_IFREG|0644, st_size=7208, ...}) = 0
mmap(NULL, 4096, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f7704212000
read(8, "\3\363\r\n\357\204\240Sc\0\0\0\0\0\0\0\0\3\0\0\0@\0\0\0sq\0\0\0d\0"..., 4096) = 4096
fstat(8, {st_mode=S_IFREG|0644, st_size=7208, ...}) = 0
[...]

PR #10268, opened 2015-02-17
Optimize the plugin loader.

abadger:
We have been caching the paths to the plugins but before we would only
cache the specific plugin that we were looking for.  This meant that we
might search through all of the plugin directories before finding the
specific module we were interested in.  The next plugin we needed we
might again search through all the plugin directories before finding the
plugin we wanted.

This new code will cache all the potential plugins in each directory
searched.  For a play that only uses one plugin we'll only search
through directories until we encounter the directory that has the proper
plugin.  For a large playbook with many plugins used we'll search
through each directory at most once each.

This should alleviate #10165


abadger:
# Possible Duplicated Efforts

Hi!

- INSERT PR(S) HERE

PR #10275, opened 2015-02-18
Optimize plugin loader update

jensdepuydt:
After testing the change requested in pull request #10268, I added a minor fix to not search for plugins in non-existing directories. 

If this is not the correct way to fix an ongoing pull request, I'm sorry. I'm not aware of a possibility to do an update to an existing pull request so I created a new pull request.

### DIFF
--- a/lib/ansible/utils/plugins.py
+++ b/lib/ansible/utils/plugins.py
@@ -64,6 +64,7 @@ class PluginLoader(object):
         self._plugin_path_cache = PLUGIN_PATH_CACHE[class_name]
 
         self._extra_dirs = []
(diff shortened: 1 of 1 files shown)


abadger:
Yep, I think you've done the right thing, both in creating the pull request and in the code :-)

Another way to do this would be to use glob.glob() instead of os.listdir().  I had code that did that but decided  it was probably faster to use os.listdir() (with glob.glob I had to have another loop for the file suffixes).

You confirmed that there was a significant speedup for some use cases using this code so I'm going to go ahead and merge it.  Thanks!

# Merged

Hi!

issue is resolved.

Thank you!

PR #10285, opened 2015-02-19
rearranged math filters

bcoca:
should also help with avoiding collisions with the math lib

### DIFF
--- a/lib/ansible/runner/filter_plugins/core.py
+++ b/lib/ansible/runner/filter_plugins/core.py
@@ -23,7 +23,6 @@ import types
 import pipes
 import glob
 import re
-import collections
 import crypt
 import hashlib
 import string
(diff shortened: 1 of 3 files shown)


abadger:
Couple notes: The `from __future__` bit should stay in... filters are shared with v2 and really we want to add the rest of our standard boilerplate for increased python3 compat to the files, not make less of it.

There's no need to rename the file.  `from __future__ import absolute_import` is eventually going to be present in all the files in the ansible codebase and after a while it will become an idiom in other people's files as well.  Problems such as this have no general solution except absolute_import so as the world slowly migrates to python2.6+ we'll see more and more use of absolute_import.

Which is not to say that you can't rename the file.  Just that it isn't necessary to do so.

bcoca:
re added absolute_import, kept the rename as a courtesy to those not using it yet

PR #10292, opened 2015-02-19
Add support for mounts in Solaris facts

dago:


bcoca:
why the timeout?

dago:
`/etc/mnttab` is a presented by a special filesystem driver in Solaris (`mntfs`). Under rare circumstances this may hang, but I want to avoid a hang during fact gathering even in rare situations.

PR #10295, opened 2015-02-19
ansible-doc now shows return value docs

bcoca:
modified ansible-doc to show documented return values (registered var structure) if a module has this data embeded.

TODO: 
- add to docsite/rst (docs.ansible.com)
- create common docs for common data returned by all modules (msg,changed,failed,skipped, etc)


L2G:
Sweet! Thanks.

PR #10298, opened 2015-02-19
Allow requesting a list to lookup plugins instead of , delimited string

### LINKED ISSUE #10291: Lookup plugins cannot return list to Jinja2 template
##### Issue Type:

Bug
##### Ansible Version:

ansible 1.8.2
(Also tested by @bcoca on current devel)
##### Environment:

N/A
##### Summary:

If a lookup plugin returns a list `['one', 'two', 'three']` this is stringified and returned to a Jinja2 template as a string `one,two,three`.
###### environment

```
export ANSIBLE_LOOKUP_PLUGINS=`pwd`
```
###### lookup mini.py

``` python
from ansible import utils, errors
import os

class LookupModule(object):

    def __init__(self, basedir=None, **kwargs):
        self.basedir = basedir

    def run(self, terms, inject=None, **kwargs):

        ret = ['gato', 'chat', 'katze', 'miau' ]

        return ret
```
###### playbook

``` yaml
- hosts:
   - localhost
  gather_facts: False
  connection: local
  tasks:
  - action: template src=t.in dest=/tmp/t.out
```
###### template

```
#-- result from lookup()
{% for k in lookup('mini', 'blah')  %}
        {{ k }}
{% endfor %}

#-- result from lookup() | list
{% for k in lookup('mini', 'blah') | list  %}
        {{ k }}
{% endfor %}

#-- lookup().split()
{{ lookup('mini', 'blah').split(',')  }}

#-- Jinja2 list
{% for k in  ['uno', 'dos', 'tres' ] %}
        {{ k }}
{% endfor %}
```
##### Expected Results:

I expected the following from `lookup('mini', 'blah')`

```
    gato
    chat
    katze
    miau
```
##### Actual Results:

Results from running above playbook:

```
#-- result from lookup()
        g
        a
        t
        o
        ,
        c
        h
        a
        t
        ,
        k
        a
        t
        z
        e
        ,
        m
        i
        a
        u

#-- result from lookup() | list
        g
        a
        t
        o
        ,
        c
        h
        a
        t
        ,
        k
        a
        t
        z
        e
        ,
        m
        i
        a
        u

#-- lookup().split()
['gato', 'chat', 'katze', 'miau']

#-- Jinja2 list
        uno
        dos
        tres
```

bcoca:
so now you can do something like:

```
{% for k in lookup('items',"['one','two','three']",wantlist=True) %}
{{k}}
{% endfor %}
```

should fix #10291

### DIFF
--- a/lib/ansible/utils/template.py
+++ b/lib/ansible/utils/template.py
@@ -89,6 +89,8 @@ def lookup(name, *args, **kwargs):
     instance = utils.plugins.lookup_loader.get(name.lower(), basedir=kwargs.get('basedir',None))
     tvars = kwargs.get('vars', None)
 
+    wantlist = kwargs.pop('wantlist', False)
+
     if instance is not None:
         try:
             ran = instance.run(*args, inject=tvars, **kwargs)
@@ -98,7 +100,7 @@ def lookup(name, *args, **kwargs):
             raise errors.AnsibleUndefinedVariable("One or more undefined variables: %s" % str(e))
         except Exception, e:
             raise errors.AnsibleError('Unexpected error in during lookup: %s' % e)
-        if ran:
+        if ran and not wantlist:
             ran = ",".join(ran)
         return ran
     else:
--- a/docsite/rst/playbooks_lookups.rst
+++ b/docsite/rst/playbooks_lookups.rst
@@ -2,13 +2,15 @@ Using Lookups
 =============
 
 Lookup plugins allow access of data in Ansible from outside sources.  These plugins are evaluated on the Ansible control
-machine, and can include reading the filesystem but also contacting external datastores and services.  
+machine, and can include reading the filesystem but also contacting external datastores and services.
 These values are then made available using the standard templating system
 in Ansible, and are typically used to load variables or templates with information from those systems.
 
-.. note:: This is considered an advanced feature, and many users will probably not rely on these features.  
+.. note:: This is considered an advanced feature, and many users will probably not rely on these features.
 
 .. note:: Lookups occur on the local computer, not on the remote computer.
+
+.. note:: Since 1.9 you can pass wantlist=True to lookups to use in jinja2 template "for" loops.
 
 .. contents:: Topics
 


lib/ansible/utils/template.py
```diff
@@ -217,7 +218,7 @@ def template_from_file(basedir, path, vars, vault_password=None):
 
     def my_lookup(*args, **kwargs):
         kwargs['vars'] = vars
-        return lookup(*args, basedir=basedir, **kwargs)
+        return lookup( *args, basedir=basedir, **kwargs)
```
sivel: Is the added whitespace here on purpose?

lib/ansible/utils/template.py
```diff
@@ -87,6 +87,7 @@ class Flags:
 def lookup(name, *args, **kwargs):
     from ansible import utils
     instance = utils.plugins.lookup_loader.get(name.lower(), basedir=kwargs.get('basedir',None))
+    wantlist = kwargs.get('wantlist', False)
```
sivel: Would it be beneficial to pop this off as to not pass it into `.run()`?

sivel:
+1, I like this approach better than some of the other recommendations.

lib/ansible/utils/template.py
bcoca: nope

lib/ansible/utils/template.py
bcoca: yes

jpmens:
@bcoca `'msg': "UnboundLocalError: local variable 'wantlist' referenced before assignment", 'failed': True}`

jpmens:
Other than 

`wantlist = False`
before

``` python
if 'wantlist' in kwargs:
```

SHIP IT!

;-)

sivel:
might be easier to use:

```
wantlist = kwargs.pop('wantlist', False)
```

bcoca:
i did not know pop could also use the same default values as get ... something new every day

jpmens:
Thank you, Brian!

PR #10304, opened 2015-02-20
Add LDom support

dago:
This patch adds LDom support for T-series Sparc CPUs.

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -2498,6 +2498,27 @@ class SunOSVirtual(Virtual):
                 if 'VirtualBox' in line:
                     self.facts['virtualization_type'] = 'virtualbox'
                     self.facts['virtualization_role'] = 'guest'
+        # Detect domaining on Sparc hardware
+        if os.path.exists("/usr/sbin/virtinfo"):
+            # The output of virtinfo is different whether we are on a machine with logical
+            # domains ('LDoms') on a T-series or domains ('Domains') on a M-series. Try LDoms first.
+            rc, out, err = module.run_command("/usr/sbin/virtinfo -p")
+            # The output contains multiple lines with different keys like this:
+            #   DOMAINROLE|impl=LDoms|control=false|io=false|service=false|root=false
+            # The output may also be not formated and the returncode is set to 0 regardless of the error condition:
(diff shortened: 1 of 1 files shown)


lib/ansible/module_utils/facts.py
```diff
@@ -2498,6 +2498,27 @@ def get_virtual_facts(self):
                 if 'VirtualBox' in line:
                     self.facts['virtualization_type'] = 'virtualbox'
                     self.facts['virtualization_role'] = 'guest'
+        # Detect domaining on Sparc hardware
+        if os.path.exists("/usr/sbin/virtinfo"):
```
bcoca, line 2502: it might be better to use the module method to find binaries, in case someone moves this to another path

lib/ansible/module_utils/facts.py
```diff
@@ -2498,6 +2498,27 @@ def get_virtual_facts(self):
...
+            # The output of virtinfo is different whether we are on a machine with logical
+            # domains ('LDoms') on a T-series or domains ('Domains') on a M-series. Try LDoms first.
+            rc, out, err = module.run_command("/usr/sbin/virtinfo -p")
+            # The output contains multiple lines with different keys like this:
+            #   DOMAINROLE|impl=LDoms|control=false|io=false|service=false|root=false
+            # The output may also be not formated and the returncode is set to 0 regardless of the error condition:
+            #   virtinfo can only be run from the global zone
+            for line in out.split('\n'):
+                fields = line.split('|')
```
bcoca: before continuing here, check that fields is actually populated as a list, or try/except the following code

lib/ansible/module_utils/facts.py
dago, line 2502: I took that test from the above section 15 lines up testing for `/usr/bin/zonename`, so when I adjust the pull request we should consistently switch the other locations in the code also. I can assure you that you cannot safely move stuff around in Solaris without breaking the packaging system and workflows.

lib/ansible/module_utils/facts.py
dago: Sorry for my suboptimal Python foo, I just started coding Python.

bcoca:
don't sweat it, I learn something new about python every day :-)

PR #10305, opened 2015-02-20
Cartesian

bcoca:
adds a cartesian product lookup plugin


Sgoettschkes:
I tested this change with 3 groups containing 3, 2 and 2 items. It worked as expected!

PR #10311, opened 2015-02-21
New lookup-plugin: `dig' for performing DNS lookups

jpmens:
This makes `dnstxt.py` obsolete, and I recommend it be removed.

### DIFF
--- a/lib/ansible/runner/lookup_plugins/dig.py
+++ b/lib/ansible/runner/lookup_plugins/dig.py
@@ -1,0 +1,212 @@
+# (c) 2015, Jan-Piet Mens <jpmens(at)gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+from ansible import utils, errors
+import socket
+HAVE_DNS=False
+try:
+    import dns.resolver
+    import dns.reversename
+    from dns.rdatatype import *
+    from dns.exception import DNSException
+    HAVE_DNS=True
+except ImportError:
+    pass
+
+def make_rdata_dict(rdata):
+    ''' While the 'dig' lookup plugin supports anything which dnspython supports
+        out of the box, the following supported_types list describes which
+        DNS query types we can convert to a dict.
+
+        Note: adding support for RRSIG is hard work. :)
+    '''
+    supported_types = {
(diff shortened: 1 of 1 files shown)


lib/ansible/runner/lookup_plugins/dig.py
```diff
@@ -0,0 +1,124 @@
...
+        except dns.resolver.NoAnswer:
+            responses.append("")
+        except dns.resolver.Timeout:
+            responses.append('')
+        except dns.exception.DNSException, e:
+            raise errors.AnsibleError("dns.resolver unhandled exception", e)
+
+        for r in responses:
+            ret.append(r)
+        return ret
```
bcoca: I'm not sure ret/responses are in the correct scope, it looks like you wanted responses inside a for loop (terms one?) in which case indentation is wrong.

bcoca:
I'm tempted to rename it to just 'dns' also to make the record type a named parameter qtype='A' (default)

lib/ansible/runner/lookup_plugins/dig.py
jpmens: Tested. Works fine here.

jpmens:
You can't call it 'dns' -- that clashes with python 'dns' module requirement.

'A' is default as it is.

lib/ansible/runner/lookup_plugins/dig.py
jpmens: I've double-checked @bcoca -- this is definitely ok as it is.

bcoca:
duh! of course you cannot call it dns .... (what idiot suggested that?)

as for the qtype, yes, A is already the default, but what I wanted was it being passed as a named parameter instead of splitting the string as it allows for more flexible calling (w/o having to manipulate domain/ip data).

jpmens:
I've added a commit which adds `qtype=xxx` support. Qtype can be lower or uppercase. Old behaviour remains though, so these are all valid:

```
example.com                            A record
example.com/A
example.com qtype=a

example.net/TXT                        TXT record
example.net  qtype=txt
```

bcoca:
perfect!

jpmens:
@bcoca How's this in addition?

``` jinja
{% set data = lookup('dig', 'iis.se. qtype=a', wantlist=True)[0] -%}
{{ data }}

{% set data = lookup('dig', 'iis.se. qtype=a flat=0', wantlist=True)[0] -%}
Domain...: {{ data.owner }}
Type.....: {{ data.type }}
Address..: {{ data.address }}
TTL......: {{ data.ttl }}

{% set data = lookup('dig', 'iis.se. qtype=mx flat=0', wantlist=True) -%}
{{ data | pprint }}
```

produces

```
91.226.36.46

Domain...: iis.se.
Type.....: A
Address..: 91.226.36.46
TTL......: 60

[{'exchange': 'mx2.iis.se.',
  'owner': 'iis.se.',
  'prio': 5,
  'ttl': 60,
  'type': 'MX'},
 {'exchange': 'mx1.iis.se.',
  'owner': 'iis.se.',
  'prio': 5,
  'ttl': 60,
  'type': 'MX'}]
```

Setting `flat=0` on a per/query basis returns the data as a dict.

bcoca:
Very nice!

clickfreak:
:+1:

jpmens:
OK, as far as I'm concerned this is now feature complete. :-)

jpmens:
Thank you, @bcoca !

Issue #10281, opened 2015-02-18
Zeros are truncated from strings.


dellis23:
##### Issue Type:

Bug Report
##### Ansible Version:

`ansible 1.8.2`
##### Environment:

Mac OSX 10.9 running against Ubuntu 14.04.
##### Summary:

When defining a string, trailing `0`s get removed.
##### Steps To Reproduce:

In my inventory:

```
server apples="1.20"
```

In my playbook:

```
debug msg="{{ apples }}"
```

Output:

```
TASK: [playbook | debug msg="{{ apples }}"] ************************** 
ok: [server] => {
    "msg": "1.2"
}
```
##### Expected Results:

1.20
##### Actual Results:

1.2

sivel:
I cannot duplicate this in either 1.8.2 or current devel.  I always get:

```
ok: [server] => {
    "msg": "20"
}
```

dellis23:
Sorry, I've updated the example.  A co-worker pointed out that this is actually a problem with it (and I don't know what _it_ is) interpreting the value as a float and truncating it accordingly.  I'm not sure where along the chain of ansible -> yaml -> jinja this bug lies.  Here was my solution though:

```
apples="{{ '1.20' }}"
```

sivel:
Ok, that I can duplicate. It looks like the use of `ast.literal_eval` in `ansible.inventory.ini.InventoryParser._parse_value` doesn't treat it correctly.

sivel:
Looks like ast was introduced in https://github.com/ansible/ansible/pull/4268

abadger:
Did a quick test and it does appear that this has been fixed by #10313.  I'm going to close this.  Feel free to open a new bug report if the error continues.

# Closing This Ticket

Hi!

We believe recent commits (likely detailed above) should resolve this question or problem for you.

This will also be included in the next major release.

If you continue seeing any problems related to this issue, or if you have any further questions, please let us know by stopping by one of the two mailing lists, as appropriate:
- https://groups.google.com/forum/#!forum/ansible-project - for user questions, tips, and tricks
- https://groups.google.com/forum/#!forum/ansible-devel - for strategy, future planning, and questions about writing code

Because this project is very active, we're unlikely to see comments made on closed tickets, but the mailing list is a great way to ask questions, or post if you don't think this particular
issue is resolved.

Thank you!

PR #10313, opened 2015-02-21
Prevents trailing zeros to be truncated from strings in hosts ini inventory variables

alobbs:
Fixes bug #10281

During the string conversion process, trailing zeros were lost as
a result of the conversion of the value to a float number. That
means that, for instance, given the "1.20", the value read was
1.2, which caused issues under some circumstances.


bcoca:
re is not a good choice for a variable name as it can get confused with re the regular expression module imported in this same file.

alobbs:
Good catch. I didn't notice. It should be okay now.

Issue #10225, opened 2015-02-12
IndexError: list index out of range


amolsh:
Hi All, 
I am using ansible since 3 months. Uptill now it was well and good.But, now it is giving error on every ansible command
For Ex. running ping modue

ansible -vvvv 172.27.59.27 -m ping 

<172.27.59.27> ESTABLISH CONNECTION FOR USER: root
<172.27.59.27> REMOTE_MODULE ping
<172.27.59.27> EXEC ssh -C -tt -vvv -o ControlMaster=auto -o ControlPersist=60s -o ControlPath="/root/.ansible/cp/ansible-ssh-%h-%p-%r" -o Port=22 -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o ConnectTimeout=10 172.27.59.27 /bin/sh -c 'mkdir -p $HOME/.ansible/tmp/ansible-tmp-1423718954.13-54115388015450 && echo $HOME/.ansible/tmp/ansible-tmp-1423718954.13-54115388015450'
172.27.59.27 | FAILED => Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/ansible/runner/**init**.py", line 590, in _executor
    exec_rc = self._executor_internal(host, new_stdin)
  File "/usr/lib/python2.7/site-packages/ansible/runner/__init__.py", line 792, in _executor_internal
    return self._executor_internal_inner(host, self.module_name, self.module_args, inject, port, complex_args=complex_args)
  File "/usr/lib/python2.7/site-packages/ansible/runner/__init__.py", line 1025, in _executor_internal_inner
    result = handler.run(conn, tmp, module_name, module_args, inject, complex_args)
  File "/usr/lib/python2.7/site-packages/ansible/runner/action_plugins/normal.py", line 57, in run
    return self.runner._execute_module(conn, tmp, module_name, module_args, inject=inject, complex_args=complex_args)
  File "/usr/lib/python2.7/site-packages/ansible/runner/__init__.py", line 475, in _execute_module
    tmp = self._make_tmp_path(conn)
  File "/usr/lib/python2.7/site-packages/ansible/runner/__init__.py", line 1267, in _make_tmp_path
    result = self._low_level_exec_command(conn, cmd, None, sudoable=False)
  File "/usr/lib/python2.7/site-packages/ansible/runner/__init__.py", line 1173, in _low_level_exec_command
[...]

PR #10315, opened 2015-02-21
v2 Avoids ssh Runtime error

alobbs:
Fixes bug #10225

Prevents an exception from raising in some corner cases where SSH might
be misconfigured. It seems not to fix a specific problem, although it
makes the method a little bit more solid.


bcoca:
I believe you meant to change the lib/ansible/runner/connection_plugins/ssh.py file, but I'll do that.

PR #10322, opened 2015-02-23
Add new function to ipaddr filter to return the nth host within a given network.

mattstep:
This is very useful for calculating known hosts within a network (e.g. gateway as nthhost(1) or nthhost(-1)).


mattstep:
Totally open to a different name for the filter, I'm just not creative enough...

PR #10328, opened 2015-02-24
Respect global settings when merging hashes

jtyr:
This patch improving a condition in the `merge_hash` function to respect the global settings for merging hashes. It allows to merge hashes only if the `hash_behaviour` is set to `merge` in `ansible.cfg`.

### DIFF
--- a/lib/ansible/utils/__init__.py
+++ b/lib/ansible/utils/__init__.py
@@ -818,7 +818,7 @@ def merge_hash(a, b):
         for k, v in dicts.iteritems():
             # if there's already such key in a
             # and that key contains dict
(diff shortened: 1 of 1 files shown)


jtyr:
This patch is trying to fix the behavior described in the following thread:
https://groups.google.com/forum/#!topic/ansible-project/P0aLopALDpQ

bcoca:
actually, this is a much deeper issue, most of these using hash_merge directly should actually be using combined_vars (which already checks for the merge_hash settings).

jtyr:
This PR could be closed in favor of PR #10329.

bcoca:
closing in favor of PR linked above

Issue #10327, opened 2015-02-24
vars resolving/ group_by: special case of bad resolving in group_by where key points to a var holding another var


srgvg:
##### Issue Type:

Fancy Corner Case Bug
##### Ansible Version:

Introduced in current devel as well as 1.8.2 and up.
In the 1.8 tree, the bug was introduced by 4cd87ec93ec1f5266e5afaaa8e6734d642284afe

Note that while the bug is unrelated, this "faulty" commit is the same as for #9753
##### Environment:

N/A (tested on Ubuntu 14.04)
##### Summary:

Using a key pointing to a variable, that itself is defined as another variable construct, in combination with group_by, the resulting key (probably that subvar) is onlky calculated once for the first host, so all hosts are put in the same group.
##### Steps To Reproduce:

Create an inventory `hosts` file with some hosts where the first three chars point to a datacenter:

```
dc1-ch1-bl1                                                                     
dc2-ch1-bl2                                                                     
dc1-ch1-bl3                                                                     
dc2-ch1-bl4                                                                     
dc2-ch1-bl5   
```

and provide this variable in `group_vars/all.yml`:

```
datacenter: '{{ inventory_hostname[:3] }}'
```

Test with the commmand 
`ansible all -i hosts -c local -m group_by -a key={{datacenter}}`

Also change the _first_ host in the inventory file from `dc1-ch1-bl1` to `dc2-ch1-bl1` and observe the difference.
##### Expected Results:

Works with ansible release1.8.1 ad675235e2 last updated 2014/12/02 10:53:48 (GMT +200):

```
$ ansible all -i hosts -c local -m group_by -a key={{datacenter}}
dc1-ch1-bl1 | success >> {
    "changed": true, 
    "groups": {
        "dc1": [
            "dc1-ch1-bl1", 
            "dc1-ch1-bl3"
        ], 
        "dc2": [
            "dc2-ch1-bl2", 
            "dc2-ch1-bl4", 
            "dc2-ch1-bl5"
        ]
    }
}
```
##### Actual Results:

```
dc1-ch1-bl1 | success >> {
    "changed": true,
    "groups": {
        "dc1": [
            "dc1-ch1-bl1",
            "dc2-ch1-bl2",
            "dc1-ch1-bl3",
            "dc2-ch1-bl4",
            "dc2-ch1-bl5"
        ]
    }
}
```

or after changing the _first_ host in the inventory file from `dc1-ch1-bl1` to `dc2-ch1-bl1`:

```
[...]

PR #10329, opened 2015-02-24
changed from hash_merge to combine vars which resets default to

bcoca:
overwrite and not merge hashing

fixes #10328
but not  #10327

### DIFF
--- a/lib/ansible/playbook/play.py
+++ b/lib/ansible/playbook/play.py
@@ -115,8 +115,8 @@ class Play(object):
         _tasks    = ds.pop('tasks', [])
         _handlers = ds.pop('handlers', [])
 
-        temp_vars = utils.merge_hash(self.vars, self.vars_file_vars)
-        temp_vars = utils.merge_hash(temp_vars, self.playbook.extra_vars)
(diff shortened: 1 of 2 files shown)


bcoca:
alternate fix for #10328

jtyr:
I can confirm that this patch fixes the issue described in the PR #10328.

srgvg:
(doesn't solve #10327)

bcoca:
updated, moved group vars up and removed inventory call as it should be redundant

srgvg:
OK, with that last commit, this solves #10327

drybjed:
<del>Can't find any new issues, at least with the codebase I'm using.</del>

Edit: I was wrong, there's still an issue with one of my templates. [This code snippet](https://github.com/debops/ansible-ferm/blob/master/templates/etc/ferm/ferm.d/10_forward.conf.j2#L10-L32) is not rendered properly when firewall should have enabled forwarding.

bcoca:
And serge also saw a group_by stop working as inventory vars are not added to the 'templatable' set, re-adding inventory vars.

drybjed:
After latest change, file is templated correctly, checking the rest.

Issue #9944, opened 2015-01-06
failing lookup affects copy/template/file even when not part of the task


evoracer:
##### Issue Type:

Bug
##### Ansible Version:

ansible 1.8.2
##### Environment:

debian 7 / gentoo
##### Summary:

The copy and template modules throws an error with a wrong file.
In ansible 1.7.x everything is fine and ansible copies the jail.conf.
In ansible 1.8.2 i cannot copy any file while i have a lookup defined in group vars. Even if the defined lookup is not used in any playbook
##### Steps To Reproduce:

group_vars/all

ssh_users:
- name: evo
  
  key: "{{ lookup('file', 'evo.pub') }}"

tasks/main.yml
- name: copy fail2ban config
  
  copy: src=jail.conf dest=/etc/fail2ban/jail.conf
##### Expected Results:

ansible 1.7.x

TASK: [common-j | copy fail2ban config] ***********************************\* 
changed: [evosrv.test]
##### Actual Results:

ansible 1.8.2

TASK: [common-j | copy fail2ban config] ***********************************\* 
fatal: [evosrv.test] => could not locate file in lookup: evo.pub

evoncken:
Appears I'm having the exact same issue. Same playbook used to work, now aborts early in "ansible-playbook". CentOS 6.6 X86_64 with all updates applied.

<c3p0> ESTABLISH CONNECTION FOR USER: root
fatal: [c3p0] => could not locate file in lookup: ssh_pubkeys/root.pub

Version history:
ansible-1.8.2-1.el6.noarch << ansible-1.8.1-1.el6.noarch << ansible-1.7.2-2.el6.noarch

I'm 99% sure that the same Ansible playbooks worked with 1.8.1.
Cannot "yum downgrade", unfortunately, to be 100% sure.

rickmb:
I have a similar issue, with a completely unrelated lookup in group_vars/all suddenly blowing up whilst doing a synchronize. Downgrading to 1.8.1 fixes the issue. (This has worked fine in many earlier versions.)

evoncken:
Did some experimenting:
 commit e938f554b74c35cc3a13f72bce46ebe5fb3aab3d - FAIL
 commit 2d17d18c1b486a22727b11f6ecae8177e90c76b3 - OK

When checking out earlier commits, playbooks run just fine.

bcoca:
@evoncken so the commit just pushes up an existing error that was swallowed before, trying to avoid the lookup running at all when not used in the template

[...]

PR #10330, opened 2015-02-24
don't template hostvars on load

bcoca:
This might resovle #9944

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,7 @@
 
 NAME = ansible
 OS = $(shell uname -s)
+python = /usr/bin/python2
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -18,6 +18,7 @@
 
 NAME = ansible
 OS = $(shell uname -s)
+python = /usr/bin/python2
```
abadger, line 21: Looks like your "I have to work on arch" change snuck in here :-)

bcoca:
fixes nothing, but checked v2 and this issue is resolved there

Issue #9862, opened 2014-12-19
include tags not working with ansible 1.8.1


passie:
##### Issue Type:

Bug Report
##### Ansible Version:

1.8.1-1.el6
##### Environment:

CentOS release 6.6 (Final)
##### Summary:

Tags from includes are being ignored
Tags from within a playbook are working just fine.

After a downgrade to Ansible 1.7.2-2.el6 this is working
##### Steps To Reproduce:

include: someotherplaybook.yml tags=test_tag 
ansible-playbook someplaybook.yml -e "target=somehost" -t test_tag 
##### Expected Results:

Tags should not be ignored
##### Actual Results:

ERROR: tag(s) not found in playbook: test_tag.  possible values:

hkariti:
Still in 1.8.2 and HEAD

natefoo:
I can confirm this on 1.8.2.

kportertx:
Can confirm in 1.6.0, 1.7.2, 1.8.0, and 1.8.2.

Current behavior conflicts with docs: http://docs.ansible.com/playbooks_tags.html

The main playbook includes other playbooks which include roles

Below is an example of the structure I have
**site.yml**

```
- include: deploy-component-playbook.yml
  tags:
    - deploy-component
```

**deploy-component-playbook.yml**

```
- name: Deploy Component
  hosts: all
  roles:
    - role: deploy-subcomponent
```

**roles/deploy-subcomponent/tasks/main.yml**

```
- name: task name
  sudo: yes
  yum:
    name: ntpd
    state: present
```

kristous:
Problem still in 1.8.4

dkoppstein:
Still open in 1.9.0.1. Applies to roles as well.

kportertx:
@dkoppstein is this issue applying to roles a regression in 1.9.0.1? I'm on 1.6.0 and can use tags on roles.

dkoppstein:
Sorry, I had a stupid bug =) I confirm that tags work on roles in 1.9.0.1. Don't know about includes though.

passie:
Any news on when this will be included in the main version ?

bcoca:
so tags should work in roles and includes in 1.9.1 (currently rc 1).

passie:
Looks like it isn't working in 1.9.1.1.el6

bindlestifff:
[...]

PR #10340, opened 2015-02-25
Fixed include tags issue described by issue #9862

kportertx:
https://github.com/ansible/ansible/issues/9862#issuecomment-75912714


jimi-c:
Hi, since we cut the devel code over to the v2 base, this patch no longer applied. I have gone ahead and fixed this in a different way however, so it should now be resolved.

Thanks!

jimi-c:
That patch was 7c1d569a26b2b7a41d6b4bc9f442fbd7f8b8a188, for the record.

PR #10341, opened 2015-02-25
Release1.8.4

vigodeltoro:


bcoca:
I'm not sure what this is meant to be, but since all the relevant commits are already in devel I'm closing this ticket

PR #10346, opened 2015-02-26
Bug fix: Search only for files as candidates

lekum:
Apparently found a recently introduced bug (in 68e86de 2015-02-17 | Optimize the plugin loader). If a folder is found matching the plugin name, it is retrieved. I have experienced the issue using digital_ocean module (it is not usable now).

I have added a test for it, but I have had to add mock as a dependency. I think it is necessary in order to do good testing.

### DIFF
--- a/lib/ansible/utils/plugins.py
+++ b/lib/ansible/utils/plugins.py
@@ -176,7 +176,8 @@ class PluginLoader(object):
         found = None
         for path in [p for p in self._get_paths() if p not in self._searched_paths]:
             if os.path.isdir(path):
-                for potential_file in os.listdir(path):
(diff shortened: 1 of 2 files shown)


bcoca:
a simpler fix might be:

```
-                    if full_name not in self._plugin_path_cache
+                    if full_name not in self._plugin_path_cache and os.path.isfile(full_path):
```

abadger:
@lekum's fix allows us to break out of the loop sooner so I think that's a better place for the code.  I think we might be able to return full_path instead of potential_path as the iterated value there as well.

abadger:
Thanks for finding this and creating a fix!

# Merged

Hi!

issue is resolved.

Thank you!

PR #10350, opened 2015-02-27
Add ansible_ssh_host

taiojia:
When execute `zabbix.py --host 'hostname'` returens an empty dict eg. `{}`. Add `ansible_ssh_host` to the dict will returns the host IP that ansible dynamic inventory can do something in the host.

### DIFF
--- a/plugins/inventory/zabbix.py
+++ b/plugins/inventory/zabbix.py
@@ -72,7 +72,7 @@ class ZabbixInventory(object):
         }
 
     def get_host(self, api, name):
-        data = {}
+        data = {'ansible_ssh_host': name}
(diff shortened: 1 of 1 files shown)


amenonsen:
Can someone who uses the zabbix plugin comment on this? It seems to make some sense, but I'm a bit puzzled about why it was returning an empty dict in the first place. But I don't understand inventory plugins very well to begin with.

taiojia:
@amenonsen I use that plugin, see my commit plz.

bcoca:
manually integrated in e4022906d387a1b6e668efc23cc45698bdf3f284

PR #10358, opened 2015-02-28
git archive in ansible-galaxy fails for non-default-branch versions

ghost:
Trying to install a branch that is not the default (typically 'master') via ansible-galaxy fails because `git archive` cannot handle remote branches. Introducing a `git checkout` solves this problem by pulling the remote branch onto the local machine.

My concrete example was:

``` yaml
# galaxy.yml
- src: https://github.com/thefinn93/ansible-tlsdate.git
  version: rename-service
```

which caused:

```
- executing: git clone https://github.com/thefinn93/ansible-tlsdate.git ansible-tlsdate
- executing: git archive --prefix=ansible-tlsdate/ --output=/tmp/tmpOLH2Y4.tar rename-service
- command git archive --prefix=ansible-tlsdate/ --output=/tmp/tmpOLH2Y4.tar rename-service failed
  in directory /tmp/tmpTlAMNS
- ansible-tlsdate was NOT installed successfully.
- you can use --ignore-errors to skip failed roles.
```

### DIFF
--- a/bin/ansible-galaxy
+++ b/bin/ansible-galaxy
@@ -359,6 +359,19 @@ def scm_archive_role(scm, role_url, role_version, role_name):
         print "  in directory %s" % tempdir
         return False
 
+    if scm == 'git':
+        if role_version:
+             checkout_cmd = ['git', 'checkout', role_version]
+             with open('/dev/null', 'w') as devnull:
+                 print "- executing: %s" % " ".join(checkout_cmd)
+                 popen = subprocess.Popen(checkout_cmd, cwd=os.path.join(tempdir, role_name),
+                                          stderr=devnull, stdout=devnull)
+                 rc = popen.wait()
+             if rc != 0:
+                 print "- command %s failed" % ' '.join(checkout_cmd)
+                 print "  in directory %s" % tempdir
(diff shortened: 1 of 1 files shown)


sivel:
Btw, I believe you can also already use `remotes/origin/branch` as the version.

ghost:
That may be a useful workaround ^_^ Still, it requires the author of `galaxy.yml` to be aware of the internal workings of `ansible-galaxy` (e.g. that the repo is checked out and archived), which I find unsatisfactory. I'd say specifying repo src and branch name should work out of the box ;-)

FlorentCoppint:
Your commit fixes the issue for me.
When will it be merged to Ansible ?

ghost:
Ask the gods :-P For the time being I'm using the workaround suggested by sivel...

tjanez:
I would close this in favour of #10620.

ghost:
I haven't tried it, but looking at the diff that seems reasonable :+1:

amenonsen:
Right, this should now be closed in favour of #10620.

amenonsen:
Closing per above comment since #13203 was merged.

PR #10362, opened 2015-03-01
Clean up OpenStack inventory to use a cleaner interface

emonty:
We introduced a cleaner interface point in shade, so that group logic can live in ansible (since it's ansible specific) and metadata munging can live in and be functionally tested in shade. Consume it.

### DIFF
--- a/contrib/inventory/openstack.py
+++ b/contrib/inventory/openstack.py
@@ -2,7 +2,7 @@
 
 # Copyright (c) 2012, Marco Vito Moscaritolo <marco@agavee.com>
 # Copyright (c) 2013, Jesse Keating <jesse.keating@rackspace.com>
-# Copyright (c) 2014, Hewlett-Packard Development Company, L.P.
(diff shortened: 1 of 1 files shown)


emonty:
This depends on https://review.openstack.org/#/c/175891/ being merged and released. Will ping when it is.

bcoca:
is this not backwards compatible?

emonty:
@bcoca it should be backwards compatible - this is mostly shifting some code reuse around. However, i just found a regression, so let me address that real quick before we merge it.

emonty:
@bcoca I'm going to come back to this and do the filtering we talked about. I can't set labels in this repo, but consider this "needs_revision"

emonty:
This should work now. I know the current one is used in Tower - we should get the Tower folks to make sure this still works for them. I'd love to get this in before 2.0 releases. @bcoca @gregdek @robynbergeron

PR #10367, opened 2015-03-02
Removed subprocess import

### LINKED ISSUE #10361: subprocess is useless in ansible/inventory/__init__.py now.
Hi, 

Seem you guy can need remove this line https://github.com/ansible/ansible/blob/release1.8.4/lib/ansible/inventory/__init__.py#L23, or subprocess is useless now.

mihirvj:
fixes #10361

### DIFF
--- a/lib/ansible/inventory/__init__.py
+++ b/lib/ansible/inventory/__init__.py
@@ -20,7 +20,6 @@ import fnmatch
 import os
 import sys
 import re
-import subprocess
 
 import ansible.constants as C
 from ansible.inventory.ini import InventoryParser


sivel:
Not sure you should be changing the commit hash for the core and extras submodules in this PR.

amenonsen:
In current devel, `lib/ansible/inventory/__init__.py` still has an `import subprocess` that doesn't appear to be used. I don't know why #10361 considers it a bug. Does having the unused import actually cause any problems? Also, this PR should definitely not be updating the submodules, as @sivel pointed out.

Summary: I think this should be closed. If someone feels like it, the import can be removed too.

bcoca:
closed via 77fc3ce7596969ea0c620f7b19ce2629f37316ff

PR #10369, opened 2015-03-02
allow executable (+x) non-script inventory files.

### LINKED ISSUE #10068: Executable and non-executable inventory files should not be conflated
I ran into the following error when running ansible-playbook -i [myinventory], using Vagrant on a Windows host:

```
ERROR: The file [myinventory] is marked as executable, but failed to execute correctly. If this is not supposed to be an executable script, correct this with `chmod -x [myinventory]`
```

The problem is that vboxfs (as does many other filesystem options) will set the execute permission on all files if mounted on a filesystem that isn't Unix-like such as FAT or NTFS.

My workaround is ugly - in Vagrant, I am mounting my source files a second time so that I can refer to a version of my inventory file without +x, such as:
    `config.vm.synced_folder ".", "/vagrant-nox", :mount_options => ["fmode=666"]`

I consider the decision to execute an inventory file as a script based solely on its execute permission a bad idea. This is a non-obvious problem and a common use case - here as some examples of other users coming across the same problem:
- http://stackoverflow.com/questions/26859360/cant-use-ansible-inventory-file-because-it-is-executable
- https://groups.google.com/forum/#!topic/ansible-project/rAM6mOEjwv4
- https://github.com/phansible/phansible/issues/71

It also allows a malicious user to execute a script with elevated privileges - all they'd need is write permission on a (non-script) inventory file.

A possible solution is to use a separate command line option if an executable inventory should be used, although this does not provide backwards compatibility.

danthedeckie:
This fixes #10068 without breaking backwards compatibility.

Sometimes having an ansible directory on a network share means it mounts +x automatically, with no easy way to get around it.

### DIFF
--- a/lib/ansible/constants.py
+++ b/lib/ansible/constants.py
@@ -163,6 +163,7 @@ DEPRECATION_WARNINGS           = get_config(p, DEFAULTS, 'deprecation_warnings',
 DEFAULT_CALLABLE_WHITELIST     = get_config(p, DEFAULTS, 'callable_whitelist', 'ANSIBLE_CALLABLE_WHITELIST', [], islist=True)
 COMMAND_WARNINGS               = get_config(p, DEFAULTS, 'command_warnings', 'ANSIBLE_COMMAND_WARNINGS', False, boolean=True)
 DEFAULT_LOAD_CALLBACK_PLUGINS  = get_config(p, DEFAULTS, 'bin_ansible_callbacks', 'ANSIBLE_LOAD_CALLBACK_PLUGINS', False, boolean=True)
+ALLOW_EXECUTABLE_INVENTORY     = get_config(p, DEFAULTS, 'allow_executable_inventory', 'ALLOW_EXECUTABLE_INVENTORY', True, boolean=True)
 
 RETRY_FILES_ENABLED            = get_config(p, DEFAULTS, 'retry_files_enabled', 'ANSIBLE_RETRY_FILES_ENABLED', True, boolean=True)
 RETRY_FILES_SAVE_PATH          = get_config(p, DEFAULTS, 'retry_files_save_path', 'ANSIBLE_RETRY_FILES_SAVE_PATH', '~/')
--- a/lib/ansible/errors.py
+++ b/lib/ansible/errors.py
@@ -38,3 +38,6 @@ class AnsibleUndefinedVariable(AnsibleError):
 
 class AnsibleFilterError(AnsibleError):
     pass
+
+class AnsibleInvalidInventory(AnsibleError):
+    pass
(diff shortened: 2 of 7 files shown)


abadger:
Can we autodetect this?  If executable, try to execute the file as a script and if that fails, try to read it as an inventory file?

abadger:
Or vice-versa, try to read the file as an inventory file.  If that fails to parse and the file is executable, then try to execute it as an inventory script.

srgvg:
Actually, it depends. If the inventory points directly to a script, and  the file is executeable, it will also check for the presence of a shebang. It will fail is no shebang is present. One could change this to fallback to ini parsing. However, if the inventory points to a directory, that shebang check is not performed, and only the executeable flag is checked. The former is inventory/**init**.py, the latter is of course the evilsh inventory_dir.py, which reimplements some of the former code... One of the things for after v2, me thinks.

danthedeckie:
shebang detection does limit you to scripting languages.  I could see people who use exclusively golang (for instance) wanting their inventory program also in the same language.  Sure, you could use a wrapper script...

bcoca:
I'm almost tempted to just try reading it as the static format if the execution fails for any reason, then just give back an error that we could neither execute nor parse the file.

srgvg:
@bcoca which is what I said. Only you have two code paths to take care of this now.

bcoca:
2 places which should call the same function .... we have too many double implementations of things already.

srgvg:
Hear! Hear!

danthedeckie:
How about this?  It's a more complex change, a bit of refactoring, but I think implements the expected behaviour, without (hopefully) breaking any current systems, and still remaining reasonably flexible.

danthedeckie:
As a side-topic / question, is there a reason more custom Exceptions aren't used?  They make unit tests a heck of a lot more robust, and easier to write.  Or is this a thing for v2 / later on?

bcoca:
the custom exceptions is something we introduced much later, we are moving to that as we touch code.

danthedeckie:
_bump_ does it still needs_revision?  Is this a good solution?

cheesegrits:
_bump_ can this issue get some love? Still a nasty problem for us Winders Vagrant users.

amenonsen:
@danthedeckie @cheesegrits can you try out PR #11643 to see if it solves your problem? If so, this PR can be closed.

abadger:
Okay, we think that this was fixed by #11643 which was merged a few days ago.  I'll go ahead and close this.  If there's further code that's needed to get his working, please open a new issue/pull request and we can get it sorted out.

# Closing This Ticket

Hi!

issue is resolved.

Thank you!

PR #10389, opened 2015-03-05
Allow customization of group by tags

j1n6:
##### SUMMARY
I would like to have a cleaner cache in my EC2 dynamic inventory. This Pull Request:

Enable customization of tag prefix:
- The default is prefixed with `tag_`. Because we have multiple vendors, I would like to customize the prefix to be called `aws_`. I may generate other dynamic inventory group prefixed with `vendorX_`
- This change should keep the backwards compatibility

Enable whitelisting, only certain tags will be used for group by tags:
- There are many other tags may be created manually for other purposes (e.g. human readable record, or managed by other team). I would like to whitelist tags that are only used for managing ec2 resources. 
- You can specify a comma separated list of tag to be used for grouping

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
plugins/inventory/ec2.ini
plugins/inventory/ec2.py

##### ANSIBLE VERSION
2.3

### DIFF
--- a/plugins/inventory/ec2.ini
+++ b/plugins/inventory/ec2.ini
@@ -87,6 +87,12 @@ group_by_route53_names = True
 group_by_rds_engine = True
 group_by_rds_parameter_group = True
 
+# If you only want to group by certain tags, default is grouping by every tag
+# group_by_tag_keys_whitelist = Name,env,role,owner,id
+
+# If you want to change the group by tags prefix, the host group format is "tag_{{TAG_NAME}}_{{TAG_VALUE}}"
+# group_by_tag_keys_prefix = tag_
+
 # If you only want to include hosts that match a certain regular expression
 # pattern_include = stage-*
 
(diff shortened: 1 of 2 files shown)


j1n6:
I also added an option to group hosts by tag values. You can whitelist or filter specific tags to be used for grouping function.

bcoca:
this needs a rebase

bcoca:
I would actually wait for this https://github.com/ansible/proposals/issues/41 now, one of the things we want to add is 'use jinja2 expression as group definition'

bcoca:
#23374 should provide for endless customization on any/all fields w/o having to change code

jstoja:
@bcoca Pretty sure this can be closed with the constructed groups.

j1n6:
I understand that your team may have priority and goals. The PR was about 3 years old, only now gets reviewed - at this point, I don't think these changes would add any values for me anymore.  Thanks for reviewing this anyway.

PR #10399, opened 2015-03-06
Add prefixing and suffixing fuctionality to assemble module

underyx:
Complements ansible/ansible-modules-core#901.

I had no idea what I was doing with this action and v2 stuff and all that, but what I did in the end felt kinda right.

### DIFF
--- a/lib/ansible/plugins/action/assemble.py
+++ b/lib/ansible/plugins/action/assemble.py
@@ -34,13 +34,18 @@ class ActionModule(ActionBase):
 
     TRANSFERS_FILES = True
 
-    def _assemble_from_fragments(self, src_path, delimiter=None, compiled_regexp=None, ignore_hidden=False):
+    def _assemble_from_fragments(self, src_path, delimiter=None, compiled_regexp=None, ignore_hidden=False, header=None, footer=None):
         ''' assemble a file from a directory of fragments '''
 
         tmpfd, temp_path = tempfile.mkstemp()
         tmp = os.fdopen(tmpfd,'w')
         delimit_me = False
         add_newline = False
+
+        if header is not None:
+            if not header.endswith('\n'):
+                header += '\n'
+            tmp.write(header)
 
         for f in sorted(os.listdir(src_path)):
             if compiled_regexp and not compiled_regexp.search(f):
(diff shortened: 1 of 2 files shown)


amenonsen:
I'm not really sure why this is needed, rather than using another fragment at the beginning and end. I don't use assemble much, so I'm willing to believe someone might find this a convenience. Anyway, @underyx, if you're still interested in this, it needs to be rebased to apply to current devel; you should ignore the code under `v1/`. This PR also needs a documentation update.

underyx:
@amenonsen as for why it's needed, please compare the current

``` yaml
- name: add drush site alias prefix fragment
  copy: content="<?php" dest={{ drush_aliases_path }}/_prefix.php mode=0770

- name: add drush site alias fragment for {{ drupal_site }}
  template: src=alias.php.j2 dest={{ drush_aliases_path }}/{{ drupal_site }}.php mode=0770

- name: generate drush aliases file
  assemble: src={{ drush_aliases_path }} dest=/etc/drush/aliases.drushrc.php mode=0750
  sudo: yes
```

to

``` yaml
- name: add drush site alias fragment for {{ drupal_site }}
  template: src=alias.php.j2 dest={{ drush_aliases_path }}/{{ drupal_site }}.php mode=0770

- name: generate drush aliases file
  assemble: src={{ drush_aliases_path }} dest=/etc/drush/aliases.drushrc.php mode=0750 prefix="<?php"
  sudo: yes
```

Note that prefixing and suffixing with the current method is an alphabetical sorting hack. This, of course, is rather flaky.

I'll try and get the rebase done soon.

underyx:
@amenonsen a quick question, though: I was under the impression that documentation updates go in ansible/ansible-modules-core, as I have done [here](https://github.com/underyx/ansible-modules-core/commit/3ef327dd48db5692d93e7ef03d3c0d2f66ef1cfd#diff-204c9a1c75ed1b68284601e5efe69ac0R66). Is that not correct?

underyx:
I have rebased onto 6cdee94c3377e46c432aba8ebf86777378061843. Let me know if I messed anything up (which is quite possible, do take extra caution when reviewing!)

@bcoca, can you please remove the `needs_rebase` label? Thanks a lot!

amenonsen:
@underyx Thanks for rebasing. Yes, documentation in this case should go to `ansible-modules-core`. Sorry, I didn't realise that you had done so already.

Your rebased code looks OK. My only comment is that it seems to assume that the prefix will always be set using `prefix="some string"`, but I wouldn't be surprised if people used `prefix="{{ lookup('file', …) }}"`. So maybe you shouldn't add a `\n` unconditionally, but only if there isn't one already.

underyx:
@amenonsen that's a good point, thanks! I've updated the PRs with this change.

The Travis build failed, but it seems like it's a completely unrelated error.

Issue #10401, opened 2015-03-06
'with_sequence' should not fail with 'count=0'


maxalbert:
**Issue Type:** Bug Report

**Ansible Version:** 1.8.4

**Environment:** N/A

**Summary:**

The `with_sequence` statement should not fail when used with `count=0`. Instead, it should just skip the task.

**Steps To Reproduce:**

Here is a minimal playbook `bug.yml` to illustrate the issue:

``` yaml

---
- hosts: all
  gather_facts: False
  tasks:
    - name: "This task should be skipped because count=0 in the 'with_sequence' statement"
      command: echo "Hello world"
      with_sequence: count=0
```

It can be run with:

`ansible-playbook -c local -i localhost, bug.yml`

**Expected Results:**

The task should be skipped because `count=0`.

**Actual Results:**

The task fails with the error message "can't count backwards" (which is counterintuitive because `count` is zero, not negative).

```
TASK: [command echo "Hello world"] ******************************************** 
fatal: [localhost] => can't count backwards

FATAL: all hosts have already failed -- aborting
```

mlaccetti:
+1 this is a hell of a bug, since apprently my "when" to skip including the task doesn't prevent this from being evaluate and shitting all over the place.

changbl:
+1

Similar failure case is:

with_sequence: start=1 end={{ some_var - 1 }}

when some_var = 1 or smaller, this case should simply be skipped instead of failure.

SidBala:
Ansible version: 1.9.1

+1 this bug is preventing a crucial workflow: 

```
with_sequence: count={{numberofTimesToRun | int}}
when: numberofTimesToRun | int > 0
```

How do I exclude this task from running when numberofTimesToRun = 0 ?

PR #10402, opened 2015-03-06
Do not fail if count=0 in with_sequence statement.

maxalbert:
This is an attempt to fix issue #10401. It does indeed fix the error described there, but I'm not sure if it introduces any other problems that I'm not aware of. I would have added a unit test, but I couldn't find any for `with_sequence`.

Also, I saw that the files `lib/ansible/runner/lookup_plugins/sequence.py` and `v2/ansible/plugins/lookup/sequence.py` are almost identical. What is the difference between the two? Should I add the fix to the latter as well (once it is approved)?

### DIFF
--- a/lib/ansible/runner/lookup_plugins/sequence.py
+++ b/lib/ansible/runner/lookup_plugins/sequence.py
@@ -152,8 +152,7 @@ class LookupModule(object):
         elif self.count is not None:
             # convert count to end
             self.end = self.start + self.count * self.stride - 1
-            del self.count
-        if self.end < self.start:
+        if self.count != 0 and self.end < self.start:
             raise AnsibleError("can't count backwards")
         if self.format.count('%') != 1:
             raise AnsibleError("bad formatting string: %s" % self.format)


sgargan:
Verified that with_sequence task no longer fails if the count is zero

``` yml
- hosts: localhost
  connection: local
  gather_facts: false

  tasks:

     - debug: msg='previously failed with backward counting error'
       with_sequence: count=0
       register: count_of_zero

     - assert:
        that:
            - count_of_zero | skipped
            - not count_of_zero | failed

```

bcoca:
I went a bit overboard and added new features to the sequence, which should also solve this issue #10565

bcoca:
@sgargan updated my commit and added your test to integration tests

maxalbert:
Nice! Thanks for your work on this, guys.

amenonsen:
@bcoca, I think that means this PR can be closed now?

maxalbert:
@amenonsen Yep, looks like this can be closed. Thanks again, everyone!

Issue #10403, opened 2015-03-06
New keyword for 'with_sequence' in order to skip task if count < 0 or start > end


maxalbert:
**Issue Type:** Feature request

**Ansible Version:** 1.8.4

**Environment:** N/A

**Summary:**

When using `with_sequence` "programmatically" (i.e., in combination with variables), it would be useful to allow a negative value for the `count` argument, or a `start` value that is larger than `end`. Currently this results in an error ("can't count backwards"), but it could be useful to simply skip the task if the range is empty. This could be achieved by adding an extra keyword, e.g. `skip_with_emtpy_range` (although it should probably be less verbose), which would be `False` by default for backwards compatibility.

Example:

``` yaml

---
- hosts: all
  gather_facts: False
  vars:
    - M: 2
    - N: 1
  tasks:
    - name: This taks should be skipped because 'skip_with_empty_range' is True.
      command: echo "Hello world"
      with_sequence: start={{ M }} end={{ N }} skip_with_empty_range=True
      #with_sequence: count={{ M - N }} skip_with_empty_range=True
```

**Expected Results:**

The task should be skipped because `skip_with_empty_range` is `True`:

```
TASK: [This taks should be skipped because 'skip_with_empty_range' is True.] *** 
skipping: [localhost]
```

**Actual Results:**

Currently the task fails with the error message "can't count backwards" (of course, `skip_with_emtpy_range` needs to be omitted when running the example above because it is not supported yet).

```
TASK: [This taks should be skipped because 'skip_with_empty_range' is True.] *** 
fatal: [localhost] => can't count backwards

FATAL: all hosts have already failed -- aborting
```

bcoca:
this is fixed in 1.9.2 and current devel, a count of 0 will skip the task,  so closing this ticket.

PR #10405, opened 2015-03-06
Add new keyword `skip_with_empty_range` to `with_sequence` statement

maxalbert:
This implements the feature request in #10403. It adds a new keyword `skip_with_empty_range` (default value: `False`) to the `with_sequence` statement which allows skipping a task if the range is empty (e.g. `count < 0` or `start > end`).

I'm not sure this is the best name for the keyword, though, so I'd appreciate alternative (in particular, more concise) suggestions. Also, I don't know if it would be more intuitive to reverse the logic, e.g. have a keyword `error_on_emtpy_range` with default value `True`.

### DIFF
--- a/lib/ansible/runner/lookup_plugins/sequence.py
+++ b/lib/ansible/runner/lookup_plugins/sequence.py
@@ -84,6 +84,7 @@ class LookupModule(object):
         self.end = None
         self.stride = 1
         self.format = "%d"
+        self.skip_with_empty_range = False
 
     def parse_kv_args(self, args):
         """parse key-value style arguments"""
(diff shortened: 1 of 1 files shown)


amenonsen:
@maxalbert, how about just `skip_empty_range` or `skip_empty`? Do you feel like rebasing (filenames have changed, this must go in `lib/ansible/plugins/lookup/sequence.py` now) and push -f'ing your new commit to this PR?

bcoca:
this is already the case in devel and current stable (1.9.2), if count=0 with_sequence skips the task

PR #10416, opened 2015-03-09
Avoid the ignore_errors statement in the introduction

dagwieers:
The `ignore_errors` statement is a legacy statement that in most cases is superseded by the `failed_when` statement. To not encourage to use ignore_errors, I replaced it by the `failed_when` statement. I also provided an example of the `changed_when` statement.

### DIFF
--- a/docsite/rst/playbooks_intro.rst
+++ b/docsite/rst/playbooks_intro.rst
@@ -213,7 +213,7 @@ them work as simply as you would expect::
 
 The command and shell module care about return codes, so if you have a command
 whose successful exit code is not zero, you may wish to explicitely tell it
-that it should ignore return code:
+that it should ignore the return code:
 
    tasks:
      - name: run this command and ignore the return code
--- a/docsite/rst/playbooks_intro.rst
+++ b/docsite/rst/playbooks_intro.rst
@@ -212,7 +212,7 @@ them work as simply as you would expect::
        command: /sbin/setenforce 0
 
 The command and shell module care about return codes, so if you have a command
-whose successful exit code is not zero, you may wish to explicitely tell it
+whose successful exit code is not zero, you may wish to explicitly tell it
 that it should ignore the return code:
 
    tasks:
--- a/docsite/rst/playbooks_intro.rst
+++ b/docsite/rst/playbooks_intro.rst
@@ -212,19 +212,31 @@ them work as simply as you would expect::
        command: /sbin/setenforce 0
 
 The command and shell module care about return codes, so if you have a command
-whose successful exit code is not zero, you may wish to do this::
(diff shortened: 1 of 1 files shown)


bcoca:
I think this is a misconception, ignore_errors allows the task to error and set an failed return value, which can then be used in subsequent tasks, failed_when changes the condition on which you consider a task failed at all or not, which won't allow for subsequent tasks to check for a |failed return value. 

I always thought ignore_errors would be better named 'dont_exit_on_error'. The directives are related, but they don't do exactly the same thing.

sivel:
I agree with @bcoca, I would not consider `ignore_errors` as a legacy statement.  It still has many valid uses.

dagwieers:
Good point, maybe "legacy" was overstated. Still I see most people use `ignore_errors` to work around a failed task, that is merely a test. (Yes, in most cases this is in combination with command or shell)

And a test (e.g. if something is running) is not a failed task if the test is negative. And IMO it is wrong to indicate it failed (in red) and ignore that failure. In such cases `when_failed: False` and using the return code in a subsequent task is much more helpful.

In an environment as ours, where one team (Engineering) creates the procedures and another team (Operations) performs/monitors execution, you want to avoid ignored (Red!) errors and incorrect (Yellow!) changes if none are to be expected.

So I still think putting `ignore_errors` forward as early as the introduction, without hinting to `failed_when` or `changed_when` is unfortunate. In case you agree, what change would be acceptable ?
1. Moving `ignore_errors` from the introduction to a more advanced section, or 
2. Provide both options in the introduction
3. Something else of your own choosing :-)

PS I agree with the confusing name, but back in the days, when things were more simple and things were not as clear-cut, it somehow made sense ;-)

amenonsen:
I reviewed the proposed change, and I'm not much in favour of it.

First, "failed_when: False" is _not_ an exact replacement for "ignore_errors: True", as @bcoca said. Yes, it's true that both will cause ansible-playbook to not exit when a task fails, but if I do the former, I can't then say «when: task|failed» later,  because I already told Ansible that it didn't fail. Having to give that up and rely solely on task.rc doesn't strike me as a great recommendation. On the other hand, it's true that `ignore_errors` isn't the greatest name, but we're stuck with it now.

Apart from that, the description of `changed_when`: «There is also a `changed_when` statement that you can use to influence when a task causes a real change» is definitely incorrect, or at least misleading. It's really more like "... can use to tell ansible how to determine whether the task's changed flag should be set".

The real concern here seems to be that tasks that fail and have the failure ignored are printed in red. I think the right thing response would be to address that directly, perhaps with a PR that causes them to not be printed in red. But I think this PR isn't desirable.

dagwieers:
Maybe we should evolve to renaming `ignore_errors` to `ignore_when`

PR #10420, opened 2015-03-10
add missing AIX network facts discovery

bmanojlovic:
By default AIX store all information for network interfaces in ODM "database"
so using given tools this information is read and returned as facts

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -2180,6 +2180,34 @@ class AIXNetwork(GenericBsdIfconfigNetwork, Network):
                 else:
                     self.parse_unknown_line(words, current_if, ips)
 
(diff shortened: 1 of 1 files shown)


bcoca:
travis failed on rake???

lib/ansible/module_utils/facts.py
```diff
@@ -2180,6 +2180,34 @@ def get_interfaces_info(self, ifconfig_path):
                 else:
                     self.parse_unknown_line(words, current_if, ips)
 
+            rc, out, err = module.run_command(['/usr/bin/uname', '-W'])
```
bcoca: use the moduels' find bin function instead of hardcoding, same for 2 other instances below

bmanojlovic:
changed to use get_bin_path from hardcoded command lines

PR #10428, opened 2015-03-11
fix syntax error

srgvg:


bcoca:
already fixed, thanks though!

PR #10452, opened 2015-03-12
Autoenv script

jhermann:
Doesn't hurt anyone not using autoenv (https://github.com/kennethreitz/autoenv), and helps those that do.

### DIFF
--- a/.env
+++ b/.env
@@ -2,7 +2,10 @@
 _venv_name=ansible
 for _venv_base in .venv venv .; do
     if test -f "$_venv_base/$_venv_name/bin/activate"; then
+        deactivate 2>/dev/null || :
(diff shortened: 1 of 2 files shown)


amenonsen:
Hi @jhermann. I know absolutely nothing about `virtualenv`, but is `autoenv` similar to `virtualenvwrapper`? I ask because I submitted a rebased PR #11689 based on someone else's patch which refers to setting up the latter to make life more convenient. Is there any reason to prefer one over the other? It might be nice to ship an example script indeed, and update the wording in #11689 accordingly.

mattclay:

bcoca:
@jhermann, sorry for the delay in response. At this time we are not considering this for inclusion in Ansible.

PR #10465, opened 2015-03-14
Don't convert numbers and booleans to strings.

feanil:
Before this change if a variable was of type int or bool and the variable was referenced
by another variable, the type would change to string.  After this change, the type of the
variable will get preserved for as long as possible.

eg. defaults/main.yml

```
PORT: 4567
OTHER_CONFIG:
  secret1: "so_secret"
  secret2: "even_more_secret"

CONFIG:
  hostname: "some_hostname"
  port: "{{ PORT }}"
  secrets: "{{ OTHER_CONFIG }}"
```

If you output `CONFIG` to json or yaml, the port would get represented in the output as a
string instead of as a number, but secrets would get represented as a dictionary.  This is
a mis-match in behaviour where some "types" are retained and others are not.  This change
should fix the issue.

### DIFF
--- a/lib/ansible/utils/template.py
+++ b/lib/ansible/utils/template.py
@@ -31,6 +31,7 @@ import datetime
 import pwd
 import ast
 import traceback
+from numbers import Number
 
 from ansible.utils.string_functions import count_newlines_from_end
 from ansible.utils import to_bytes, to_unicode
@@ -81,6 +82,11 @@ class Flags:
 
 FILTER_PLUGINS = None
 _LISTRE = re.compile(r"(\w+)\[(\d+)\]")
+
+# A regex for checking to see if a variable we're trying to
+# expand is just a single variable name.
+SINGLE_VAR = re.compile(r"^{{\s*(\w*)\s*}}$")
+
 JINJA2_OVERRIDE = '#jinja2:'
 JINJA2_ALLOWED_OVERRIDES = ['trim_blocks', 'lstrip_blocks', 'newline_sequence', 'keep_trailing_newline']
 
@@ -109,7 +115,6 @@ def lookup(name, *args, **kwargs):
 def template(basedir, varname, templatevars, lookup_fatal=True, depth=0, expand_lists=True, convert_bare=False, fail_on_undefined=False, filter_fatal=True):
     ''' templates a data structure by traversing it and substituting for other data structures '''
     from ansible import utils
-
     try:
         if convert_bare and isinstance(varname, basestring):
             first_part = varname.split(".")[0].split("[")[0]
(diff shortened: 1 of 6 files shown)


feanil:
FYI @e0d

feanil:
@bcoca I've also updated the template integration test.  Would it be possible to get this merged sooner than v2?  It is the only thing holding us back at v1.5

Please let me know if there is any additional work that I can do to get this to land sooner.

abadger:
I don't think it will land sooner than v2.  We're prepping for 1.9 final release this week.  This code is unfortunately going to result in a change of behaviour for end users so it's not something we can just sneak in now.

I know it's not much of a consolation but I'm looking at this now to see if we can land it in the new tree right away or if there's more work to do.

abadger:
With this new code, what happens in this case:

```
vars:
  - version1: "5"
  - version2: "{{ version1 }}"
```

Does version2 end up being an int or a string in that case?

feanil:
It would remain a string.

lib/ansible/utils/template.py
```diff
@@ -117,11 +121,15 @@ def template(basedir, varname, templatevars, lookup_fatal=True, depth=0, expand_
...
-
+                # template_from_string may return non strings for the case where the var is just
+                # a reference to a single variable, so we should re_check before we do further evals
+                if isinstance(varname, basestring):
+                    if (varname.startswith("{") and not varname.startswith("{{")) or varname.startswith("["):
+                        eval_results = utils.safe_eval(varname, locals=templatevars, include_exceptions=True)
+                        if eval_results[1] is None:
+                            varname = eval_results[0]
+                    
+            test_eval_results = utils.safe_eval(varname, locals=templatevars, include_exceptions=True)
```
abadger: Is test_eval_results just left over from debugging this?

lib/ansible/utils/template.py
```diff
@@ -316,10 +324,18 @@ def my_finalize(thing):
...
             data = unicode(data, 'utf-8')
+   
+            # Check to see if the string we are trying to render is just referencing a single
+            # var.  In this case we don't wont to accidentally change the type of the variable
+            # to a string by using the jinja template renderer. We just want to pass it. 
+            only_one = SINGLE_VAR.match(data)
+            if only_one:
+                var_name = only_one.group(1)
+                if var_name in vars:
+                    return vars[var_name]
```
abadger, line 348: This doesn't seem to handle filters.  Then again, filters are a hard problem because they could change the type of the variable intentionally.

abadger:
Another thing @bcoca mentioned to me yesterday was since this won't hit
until v2 we should target it against the v2/ansible tree instead of the
lib/ansible tree.

lib/ansible/utils/template.py
feanil, line 348: Yea, I intentionally am only handling the use case of wanting to reference vars in a more complex structure and retaining type.  filters or other jinja operations would make it very hard to infer type correctly and seems more error prone without hooking deeper into Jinja's parsing logic.

feanil:
Taking a look at doing that now, are v2 tests different?  How can I run them locally?  It looks like I'll need to move these updates to `v2/ansible/template/__init__.py`?  Let me know if that is not the right place.

feanil:
@abadger I made the relevant changes in v2.  I changed it up a bit and made the full change in the `template` function instead of having some of that logic in the `_do_template` function.  This felt a bit cleaner since all the ansible specifc behaviour changes are now in the `template` function.  Let me know what you think.  I'm not sure how to test the v2 change so if you could point me to some docs about that, I'd be happy to test it.  I left the v1 change + tests in for now but let me know if you want me to do something different there.

feanil:
@abadger  @bcoca   Are there any more changes you would want me to make on this?  Should I removed the edits to the v1 code?

abadger:
Great!  Thanks for writing this and iterating on it with us at pycon!

# Merged

Hi!

issue is resolved.

Thank you!

Issue #10422, opened 2015-03-10
No JSON object could be decoded when not exporting AWS credentials and using ec2 inventory script


oryband:
executing `ansible-playbook` with [ec2 inventory](https://github.com/ansible/ansible/blob/devel/plugins/inventory/ec2.py) and forgetting to export credentials throws a very confusing and non-verbose Python exception. This happened to me various times, and each time I wasted 30 minutes in tracking down the source of the problem:

``` sh
# this throws an unexplainable error
$ ansible-playbook parameters...
Traceback (most recent call last):
  File "/home/ory/.virtualenvs/ubik/bin/ansible-playbook", line 324, in <module>
    sys.exit(main(sys.argv[1:]))
  File "/home/ory/.virtualenvs/ubik/bin/ansible-playbook", line 160, in main
    inventory = ansible.inventory.Inventory(options.inventory, vault_password=vault_pass)
  File "/home/ory/.virtualenvs/ubik/local/lib/python2.7/site-packages/ansible/inventory/__init__.py", line 100, in __init__
    self.parser = InventoryDirectory(filename=host_list)
  File "/home/ory/.virtualenvs/ubik/local/lib/python2.7/site-packages/ansible/inventory/dir.py", line 56, in __init__
    parser = InventoryScript(filename=fullpath)
  File "/home/ory/.virtualenvs/ubik/local/lib/python2.7/site-packages/ansible/inventory/script.py", line 49, in __init__
    self.groups = self._parse(stderr)
  File "/home/ory/.virtualenvs/ubik/local/lib/python2.7/site-packages/ansible/inventory/script.py", line 57, in _parse
    self.raw  = utils.parse_json(self.data)
  File "/home/ory/.virtualenvs/ubik/local/lib/python2.7/site-packages/ansible/utils/__init__.py", line 551, in parse_json
    results = json.loads(data)
  File "/usr/lib/python2.7/json/__init__.py", line 338, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python2.7/json/decoder.py", line 366, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python2.7/json/decoder.py", line 384, in raw_decode
    raise ValueError("No JSON object could be decoded")
ValueError: No JSON object could be decoded

# this works
$ export AWS_ACCESS_KEY_ID=...
$ export AWS_SECRET_ACCESS_KEY=...
$ ansible-playbook parameters
```

[...]

PR #10467, opened 2015-03-15
Fix for issue #10422. 

sgargan:
Issue Type:

Bugfix Pull Request

Ansible Version:

1.9

Summary:

Outputs informative error message when AWS credentials are not available.

### DIFF
--- a/plugins/inventory/ec2.py
+++ b/plugins/inventory/ec2.py
@@ -343,7 +343,7 @@ class Ec2Inventory(object):
             conn = ec2.connect_to_region(region)
         # connect_to_region will fail "silently" by returning None if the region name is wrong or not supported
         if conn is None:
-            raise Exception("region name: %s likely not supported, or AWS is down.  connection to region failed." % region)
+            self.fail_with_error("region name: %s likely not supported, or AWS is down.  connection to region failed." % region)
         return conn
 
     def get_instances_by_region(self, region):
@@ -365,12 +365,11 @@ class Ec2Inventory(object):
 
         except boto.exception.BotoServerError, e:
             if e.error_code == 'AuthFailure':
-                self.display_auth_error()
-
-            if  not self.eucalyptus:
-                print "Looks like AWS is down again:"
-            print e
-            sys.exit(1)
+                error = self.get_auth_error_message()
+            else:
+                backend = 'Eucalyptus' if self.eucalyptus else 'AWS' 
+                error = "Error connecting to %s backend.\n%s" % (backend, e.message)
+            self.fail_with_error(error)
 
     def get_rds_instances_by_region(self, region):
         ''' Makes an AWS API call to the list of RDS instances in a particular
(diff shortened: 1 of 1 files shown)


plugins/inventory/ec2.py
```diff
@@ -334,23 +334,24 @@ def do_api_calls_update_cache(self):
...
+    def connect(self, region):
+        ''' create connection to api server'''
+        if self.eucalyptus:
+            conn = boto.connect_euca(host=self.eucalyptus_host)
+            conn.APIVersion = '2010-08-31'
+        else:
+            conn = ec2.connect_to_region(region)
+        # connect_to_region will fail "silently" by returning None if the region name is wrong or not supported
+        if conn is None:
+            raise Exception("region name: %s likely not supported, or AWS is down.  connection to region failed." % region)
```
abadger: Is it intentional that htis isn't being caught anywhere?

plugins/inventory/ec2.py
```diff
@@ -379,23 +383,33 @@ def get_rds_instances_by_region(self, region):
...
-            print("region name: %s likely not supported, or AWS is down.  connection to region failed." % region)
-            sys.exit(1)
+        boto_paths = ['/etc/boto.cfg', '~/.boto', '~/.aws/credentials']
+        boto_config_found = list(p for p in boto_paths if os.path.isfile(os.path.expanduser(p)))
+        if len(boto_config_found) > 0:
+            errors.append(" - Boto configs found at '%s', but the credentials contained may not be correct" % ', '.join(boto_config_found))
+        else:
+            errors.append(" - No Boto config found at any expected location '%s'" % ', '.join(boto_paths))
+
+        raise Exception('\n'.join(errors))
```
abadger: It seems like returning this string is the expected path through the function.  So this should probably just be a return instead of raising an exception.  Then the code that's calling it can decide to print it and reraise the exception that got us here if it chooses.

sgargan: Previous to this it was printing the string and exiting with code 1. When used with ansible-playbook the string was being swallowed and not displayed in the console. In both cases raising an exception causes the error to bubble up and be displayed. It might be better named fail_due_to_auth_error as its intended to fail the script.

abadger: Okay... what about using `sys.stderr.write('%s\n' % ('\n'.join(errors)))` at the end of this function and then the calling function does sys.exit(1) afterwards?  Almost the same result but it doesn't abuse the exception mechanism to make it happen :-)

Probably should do something similar with the exception raised by the new connect method... maybe make the exception thrown there specific and then in this toplevel exception handler, print the exception message to stderr and sys.exit(1).  The general rule is exceptions should be thrown and unhandled if there are programming errors.  Error messages and sys.exit should be used when there are errors the user can correct without diving into the ansible code.

sgargan:
This more in the spirit of what you're thinking? Looking through this and the other inventory scripts, there is an opportunity to capture some of the common functionality in a support library.

abadger:
Yeah, since we hadn't really supported inventory scripts as more than examples (1.9 they'll be in the tarball for the first time but  we still don't take care of installing them) we probably haven't put as much effort into them as they deserve.

Your changes look good now.  Merged!

# Merged

Hi!

issue is resolved.

Thank you!

PR #10474, opened 2015-03-16
Including diff in Runner results return so the PlayBook, PlayBook callba...

victorgp:
1: PlayBook callbacks or plugins can deal with it.

If you run Ansible with "diff", the diff content is created in the subthreads of the multiprocess execution, then is passed back to the Runner but the runner in the _partition_results method is ignoring the "diff" so it never gets back to the PlayBook and cannot be used by the callbacks or stats modules.
That means that nothing can use the diff to deal with it (plugins, diff analysis, storing in a db, printing in an specific format, sending by email).

With this pull request that just adds a line to the code, the "diff" content is passed back to the PlayBook so the PlayBook callbacks, stats or plugins can use it to perform extra operations.

It doesn't add extra load to the network because the "diff" is already being passed to the main thread (Runner)

2: Allowing configuration of remote_cp the same as remote_tmp is allowed instead of being hardcoded in the code


victorgp:
Any update on this pull request?

bcoca:
this probably won't get integrated until after v2 is the main devel version

bcoca:
so 2.0 is out and looking at this ticket, im not sure the commits are related to the text in the PR

sivel:
@victorgp,

Hi!

sivel:
@victorgp,

Hi!

bcoca:
So we went with an alternate fix in 2.0, since we restructured runner into over a dozen classes it is very hard to match this patch to it's new equivalent. Now we just pass diffs directly into the results and the callbacks through the subprocess queues.

So I'm going to close this PR as we took an alternate fix built into the 2.0 rewrite.

victorgp:
Sorry @bcoca i didn't have time to take care of this. Thank you for the fix in the new version!

PR #10476, opened 2015-03-16
Darwin uptime_seconds fact

cblument:
Another try for darwin platform with the uptime_seconds fact.  Using the get_sysctl method returns the following for boottime:

```
derecho:~ cblument$ /usr/sbin/sysctl hw machdep kern|grep boottime
kern.boottime: { sec = 1426005997, usec = 0 } Tue Mar 10 11:46:37 2015
kern.boottime = Tue Mar 10 11:46:37 2015
```

I am guessing, depending on the version of mac os x, that the order of what is returned is different or one or the other is returned.  Maybe it changes after reboots.  This commit should handle any of those cases.

Wrapping this up in a try/except seems yucky but not sure how else to handle it.

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -1592,6 +1592,7 @@ class Darwin(Hardware):
         self.get_mac_facts()
         self.get_cpu_facts()
         self.get_memory_facts()
+        self.get_uptime_facts()
         return self.facts
 
     def get_sysctl(self):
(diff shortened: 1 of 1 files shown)


amenonsen:
@cblument there's already a `get_uptime_facts` in devel that looks at `/proc/uptime` (I don't know if Darwin has that, but I assume not, or you would probably have used it to begin with). If you're still interested in this PR, could you please update it to apply to current devel?

nitzmahone:
Closing for inactivity

PR #10478, opened 2015-03-17
ensure ansible-galaxy works for scm clones when specified version differ...

### LINKED ISSUE #10477: ansible-galaxy fails to install roles when specified version differs from repository's default branch
When we run ansible-galaxy specifying "version: some_branch_thats_not_the_repo_default_branch",

we see

```
- executing: git clone ssh://git@my.repo.com:/path/to/my/role.git rolename
- executing: git archive --prefix=rolename/ --output=/tmp/tmpaVd4UF.tar some_branch_thats_not_the_repo_default_branch
- command git archive --prefix=rolename/ --output=/tmp/tmpaVd4UF.tar some_branch_thats_not_the_repo_default_branch failed
  in directory /tmp/tmp0c9F7Y
- rolename was NOT installed successfully.
```

because the clone is using the default branch and has not pulled down the objects required to perform the export.  We have developed a fix that specifies the branch for the clone so the subsequent archive succeeds, which we will submit in a pull request.

semifocused:
...s from repository default branch

Fixes https://github.com/ansible/ansible/issues/10477

### DIFF
--- a/bin/ansible-galaxy
+++ b/bin/ansible-galaxy
@@ -345,6 +345,8 @@ def scm_archive_role(scm, role_url, role_version, role_name):
         return False
     tempdir = tempfile.mkdtemp()
     clone_cmd = [scm, 'clone', role_url, role_name]
+    if scm == 'git' and role_version is not None and role_version.strip():
+        clone_cmd.extend(['--branch', role_version])
     with open('/dev/null', 'w') as devnull:
         try:
             print "- executing: %s" % " ".join(clone_cmd)
--- a/bin/ansible-galaxy
+++ b/bin/ansible-galaxy
@@ -345,8 +345,6 @@ def scm_archive_role(scm, role_url, role_version, role_name):
         return False
     tempdir = tempfile.mkdtemp()
     clone_cmd = [scm, 'clone', role_url, role_name]
-    if scm == 'git' and role_version is not None and role_version.strip():
-        clone_cmd.extend(['--branch', role_version])
     with open('/dev/null', 'w') as devnull:
         try:
             print "- executing: %s" % " ".join(clone_cmd)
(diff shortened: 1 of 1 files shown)


willthames:
Does this work with tags? 

The primary purpose of the version flag is to use tagged versions, which relate to a specific commit, rather than a branch which can relate to different commits - it's all for repeatability. 

Can you achieve the same result through tagging? 

:-1: from me I'm afraid.

willthames:
If you do actually want this behaviour (and there's no real reason not specifying a version should only get the HEAD of the master branch), can I suggest adding a role_branch key to the role dict and specifying it that way?

semifocused:
Thanks Will for the feedback.  Full agreement that `version` needs to work for tags.  There are valid use cases for both branches and tags, and of course git fully supports both so we're going to have a look at a cleaner revision that supports both as well.  I could live with the `role_branch` approach - but git is often smart about discerning the difference between a branch vs. a tag, so we're going to look at whether there's a way to have it do so in this case as well.  We'll turn around a revision to this PR shortly, either way.

semifocused:
Will,

We've researched git behavior a bit more, and it seems that we can get the best of all worlds with an explicit `git checkout` of the specified `role_version` prior to the `git archive`.  This means that we'll always archive from `HEAD` of whatever `role_version` is checked out.  `role_version` can be a branch, a tag, or any `<tree-ish>` supported by git including specific commit IDs.

I've revised the PR to incorporate this approach, which should now work as before when `role_version` is not specified, for `role_version` specified to a git `<tree-ish>` or for `role_version` specified to a branch.

Hopefully this addresses your concerns and also addresses our (and git's) use case of operating from most recent code on a specified branch, when our devs deem that to be appropriate.

Have a look and let me know your thoughts.

Best,

- Mick

semifocused:
Will,

We've researched git behavior a bit more, and it seems that we can get the best of all worlds with an explicit `git checkout` of the specified `role_version` prior to the `git archive`.  This means that we'll always archive from `HEAD` of whatever `role_version` is checked out.  `role_version` can be a branch, a tag, or any `<tree-ish>` supported by git including specific commit IDs.

I've revised the PR to incorporate this approach, which should now work as before when `role_version` is not specified, for `role_version` specified to a git `<tree-ish>` or for `role_version` specified to a branch.

Hopefully this addresses your concerns and also addresses our (and git's) use case of operating from most recent code on a specified branch, when our devs deem that to be appropriate.

Have a look and let me know your thoughts.

Best,
- Mick

willthames:
That sounds great. I'll merge the PR into my working copy on Monday - given
how frequently I'm reinstalling roles at the moment, I'll definitely be
able to test it out for my use cases at least

willthames:
Works for my requirements. Might be worth squashing the commits but I'm not merging it so not my call.

:+1:

semifocused:
Thanks for the review Will.  @bcoca - I will squash the commits for a cleaner history, stand by for that.

semifocused:
@willthames @bcoca - squashed commits and replacing this PR with https://github.com/ansible/ansible/pull/10620

willthames:
For future reference, assuming the PRs are identical, repushing the same branch with the squashed commits would have just updated this PR (the primary benefit of which is keeping the comment history in one place).

PR #10483, opened 2015-03-17
Extending union filter for dictionaries

jtyr:
This patch allows to use the `union` filter for dictionaries. The possible use case is as follows:

```
my_var1:
  option1: val1
  option2: val2

my_var2:
  option3: val3

my_var3: "{{ my_var1 | union(my_var2) }}"
```

### DIFF
--- a/lib/ansible/runner/filter_plugins/mathstuff.py
+++ b/lib/ansible/runner/filter_plugins/mathstuff.py
@@ -53,7 +53,10 @@ def symmetric_difference(a, b):
     return c
 
 def union(a, b):
-    if isinstance(a,collections.Hashable) and isinstance(b,collections.Hashable):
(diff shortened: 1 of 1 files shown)


mingfang:
+1

amenonsen:
See also PR #7872 that provides equivalent functionality using a filter named `hash_merge`. I like the idea of using `union` more, but don't know what to do about `hash_replace` in that case. It's also possible that `hash_merge` more clearly expresses the actual behaviour (i.e. that later keys replace earlier ones).

amenonsen:
OK, this should be closed in favour of #10489, which (takes this commit and) extends it to support other set operations on dicts as well.

bcoca:
closing this in favor of a more complete solution in one of the 2 linked PRs

PR #10484, opened 2015-03-17
set_fact: merge 'dict' variables properly

ghost:
`set_fact` does not merge 'dict' variables, but replaces the old dict with the new one instead, e.g. the playbook:

``` yaml
- hosts: localhost
  gather_facts: no
  tasks:
  - set_fact:
      mydict:
        key1: value1
        key2: value2
  - set_fact:
      mydict:
        key3: value3
  - debug: var=mydict
```

will result in:

```
ok: [localhost] => {
    "mydict": {
        "key3": "value3"
    }
}
```

This pull request fixes the incorrect behavior resulting in the correct output with the tasks above:

```
ok: [localhost] => {
    "mydict": {
        "key1": "value1", 
        "key2": "value2", 
        "key3": "value3"
    }
}
```

### DIFF
--- a/lib/ansible/utils/__init__.py
+++ b/lib/ansible/utils/__init__.py
@@ -1623,10 +1623,12 @@ def _load_vars_from_folder(folder_path, results, vault_password=None):
 
 def update_hash(hash, key, new_value):
     ''' used to avoid nested .update calls on the parent '''
-
     value = hash.get(key, {})
-    value.update(new_value)
-    hash[key] = value
+    if(isinstance(value, dict)) and (isinstance(new_value, dict)):
+        hash[key] = merge_hash(value, new_value)
+    else:
+        value.update(new_value)
+        hash[key] = value
 
 def censor_unlogged_data(data):
     '''


bcoca:
this is how facts work in general, but it should be configurable.

ghost:
It could depend on 'merge_behavior' in ansible.cfg. Should I recreate a pull request with that logic?

bcoca:
as a default, i was thinking of making this settable per set_fact invocation

ghost:
That would be nice!

DavidWittman:
This configuration already exists as [hash_behaviour](http://docs.ansible.com/intro_configuration.html#hash-behaviour), so you should probably check for that here.

Here's how it's done in [combine_vars](https://github.com/ansible/ansible/blob/5853828ca51341b5f2d89ade561761e0b749bc46/v2/ansible/utils/vars.py#L25-L30):

``` python
def combine_vars(a, b):

    if C.DEFAULT_HASH_BEHAVIOUR == "merge":
        return merge_hash(a, b)
    else:
        return dict(a.items() + b.items())
```

ghost:
David, I know about this setting, in fact it's already switched on in my setup. The problem is that the 'set_fact' module does not make use of this setting, as I have written already.

DavidWittman:
@nfubh I'm suggesting that you only merge for `set_fact` calls if the default hash behavior is "merge", otherwise just leave the functionality as it currently is. @bcoca's suggestion of setting per invocation seems fine as well.

ghost:
@DavidWittman Ok, I can re-implement my pull request according to your suggestion, can you review and merge it then?

DavidWittman:
@nfubh I don't have that authority... I'm just a vocal observer.

c0rner:
I'm glad to see this is being worked on as set_facts does not currently honor the hash merge setting. However this patch will make set_facts default to merge on dicts, I'm thinking it might break things.

Having this configurable on a per set_fact invocation would be most excellent!

This playbook shows something is rotten in the land of set_fact (run with hash merge enabled).

```
---
- hosts: all
  gather_facts: no
  sudo: no
  vars:
    - test:
        msg: "message"
        extra: "filler"

  tasks:
  - set_fact:
      demo:
        msg: "{{ test.msg }} from test"

  - set_fact:
      test:
        msg: "{{ demo.msg }} from demo"

  - debug: var=demo
  - debug: var=test

  - set_fact:
      demo:
        new: "new message"

  - set_fact:
      test:
        new: "new message"

  - debug: var=demo
  - debug: var=test
```

romain-dartigues:
@bcoca Hi, would it be acceptable to allow something like:

``` yaml
---
mydict:
 key1: 1
 key2: 2
 key3: 3

---
- set_fact:
# actually not working?
    mydict.key2: value2
- debug: var=mydict
```

``` json
{
  "var": {
    "mydict": {
      "key1": 1,
      "key2": "value2",
      "key3": 3
    }
  }
}
```

I think it is indeed logical to not merge when doing:

``` yaml
- set_fact:
    mydict:
      key2: value2
```

because we excplicitely ask that "mydict" is "...".

bcoca:
@romain-dartigues, the problem there is the . notation does not work well as it is expected to be a key of the ansible_facts dictionary

amenonsen:
First, I like (and want to use) this functionality. I agree that the original patch isn't quite there yet.

I'd be all right with set_fact merging only when hash_behaviour is set to merge, or I would be even happier to have some notation like mydict.key to update specific keys. I note that jinja2's map filter can accept map(attribute='foo.bar.baz') and descend into subkeys during lookup, so there's some reasonable precedent for behaving that way when mydics.key is not a key of ansible_facts but mydict is.

@bcoca do you have anything else in mind for how to tell set_fact to merge or not per-task?

amenonsen:
Since this PR has remained uncommented for more than three months now, I think it should be closed. In the meanwhile, it's become possible to explicitly control merging/not merging by doing something like this:

```
- set_fact:
    x: "{{ x|default({})|combine({'p': 'q'}) }}"
```

amenonsen:
Closing per above comment, due to the availability of the combine filter in v2.

PR #10488, opened 2015-03-17
Become su fixes

bcoca:
- added missing inventory override for become_method
- created new pe class bsdsu due to big differences in su options (need to check aix/solaris)
- for Linux fixes issues seen with su now that checksums do use it when present


bcoca:
need to revisit of find better universal solution

PR #10489, opened 2015-03-17
Set theory dict support

bcoca:
Incorporated #10483 and added support to the rest of the set theory functions

test play:

```
- hosts: localhost
  gather_facts: False
  vars:
    dicta:
        a: 1
        b: 2
        c: 3
        d: 4
    dictb:
        c: 3
        d: 2
        e: 3
        f: 4
    union: "{{dicta|union(dictb)}}"
    inter: "{{dicta|intersect(dictb)}}"
    diff: "{{dicta|difference(dictb)}}"
    symdiff: "{{dicta|symmetric_difference(dictb)}}"
  tasks:
    - debug: var=union
    - debug: var=inter
    - debug: var=diff
    - debug: var=symdiff
```

results:

```

TASK: [debug var=union] ******************************************************* 
ok: [localhost] => {
    "var": {
        "union": {
            "a": 1,
            "b": 2,
            "c": 3,
            "d": 2,
            "e": 3,
            "f": 4
        }
    }
}

TASK: [debug var=inter] ******************************************************* 
ok: [localhost] => {
    "var": {
        "inter": {
            "c": 3,
            "d": 4
        }
    }
}

TASK: [debug var=diff] ******************************************************** 
ok: [localhost] => {
    "var": {
        "diff": {
            "a": 1,
            "b": 2
        }
    }
}

TASK: [debug var=symdiff] ***************************************************** 
ok: [localhost] => {
    "var": {
        "symdiff": {
            "a": 1,
            "b": 2,
            "e": 3,
            "f": 4
        }
    }
}

localhost                  : ok=4    changed=0    unreachable=0    failed=0   
```

### DIFF
--- a/lib/ansible/plugins/filter/mathstuff.py
+++ b/lib/ansible/plugins/filter/mathstuff.py
@@ -33,19 +33,22 @@ def unique(a):
                 c.append(x)
     return c
 
-def intersect(a, b):
+def intersect(a, b, strict=False):
     if isinstance(a, Hashable) and isinstance(b, Hashable):
         c = set(a) & set(b)
     elif isinstance(a, Mapping) and isinstance(b, Mapping):
         c = {}
         for k in intersect(a.keys(),b.keys()):
-            if a[k] == b[k]:
+            if strict:
+                if a[k] == b[k]:
+                    c[k] = a[k]
+            else:
                 c[k] = a[k]
     else:
         c = unique(filter(lambda x: x in b, a))
     return c
 
-def difference(a, b):
+def difference(a, b, strict=False):
     if isinstance(a, Hashable) and isinstance(b, Hashable):
         c = set(a) - set(b)
     elif isinstance(a, Mapping) and isinstance(b, Mapping):
@@ -53,15 +56,17 @@ def difference(a, b):
         # get keys that are different
         for k in difference(a.keys(),b.keys()):
             c[k] = a[k]
-        # get values that are different
-        for k in intersect(a.keys(), b.keys()):
-            if a[k] != b[k]:
-                c[k] = a[k]
+
+        if strict:
+            # get values that are different
+            for k in intersect(a.keys(), b.keys()):
+                if a[k] != b[k]:
+                    c[k] = a[k]
     else:
         c = unique(filter(lambda x: x not in b, a))
     return c
 
-def symmetric_difference(a, b):
+def symmetric_difference(a, b, strict=False):
     if isinstance(a, Hashable) and isinstance(b, Hashable):
         c = set(a) ^ set(b)
     elif isinstance(a, Mapping) and isinstance(b, Mapping):
@@ -72,10 +77,12 @@ def symmetric_difference(a, b):
                 c[k] = b[k]
             else:
                 c[k] = a[k]
-        # get values that are different
-        for k in intersect(a.keys(), b.keys()):
-            if a[k] != b[k]:
-                c[k] = a[k]
+
+        if strict:
+            # get values that are different
+            for k in intersect(a.keys(), b.keys()):
+                if a[k] != b[k]:
+                    c[k] = a[k]
     else:
         c = unique(filter(lambda x: x not in intersect(a,b), union(a,b)))
     return c
--- a/lib/ansible/plugins/filter/mathstuff.py
+++ b/lib/ansible/plugins/filter/mathstuff.py
@@ -39,7 +39,8 @@ def intersect(a, b):
     elif isinstance(a, Mapping) and isinstance(b, Mapping):
         c = {}
         for k in intersect(a.keys(),b.keys()):
-            c[k] = a[k]
+            if a[k] == b[k]:
+                c[k] = a[k]
     else:
         c = unique(filter(lambda x: x in b, a))
     return c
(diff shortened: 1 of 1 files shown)


abadger:
I'm not sure I like the concept because of what happens when there's collisions.  Maybe if it was more like this:

```
vars:
  dicta:
    a: 1
  dictb:
    a: 2
  union:  "{{ dicta | union(dictb) }}"  => throws an error
  union: "{{ dicta | union(dictb, dict_collision='merge') }}"   => values in dictb override values in dicta
```

Having some sort of parameter to enable overriding would force the user to read the documentation if they're overriding variables.  This would alert them to the potential caveats (including which dictionary value will win.)

If we wanted to be really advanced, we could make the default only throw an error if the value of a in dicta was different than the value of a in dictb but that could be a rabbitwarren (because of nested data structures).

bcoca:
the union is actualy from #10483, but i think we might just want to add an additional parameter to union and the rest to either match only on keys or on key+value

bcoca:
updated play to include strict/override versions:

```

TASK: [debug var=union] ******************************************************* 
ok: [localhost] => {
    "var": {                                                                                                                                                                  
        "union": {                                                                                                                                                            
            "a": 1,                                                                                                                                                           
            "b": 2,                                                                                                                                                           
            "c": 3,                                                                                                                                                           
            "d": 4,                                                                                                                                                           
            "e": 3,                                                                                                                                                           
            "f": 4                                                                                                                                                            
        }                                                                                                                                                                     
    }                                                                                                                                                                         
}                                                                                                                                                                             

TASK: [debug var=union_overwrite] ********************************************* 
ok: [localhost] => {
    "var": {                                                                                                                                                                  
        "union_overwrite": {                                                                                                                                                  
            "a": 1,                                                                                                                                                           
            "b": 2,                                                                                                                                                           
            "c": 3,                                                                                                                                                           
            "d": 2,                                                                                                                                                           
            "e": 3,                                                                                                                                                           
            "f": 4                                                                                                                                                            
        }                                                                                                                                                                     
    }
}

TASK: [debug var=inter] ******************************************************* 
ok: [localhost] => {
    "var": {
        "inter": {
            "c": 3,
            "d": 4
        }
    }
}

TASK: [debug var=inter_strict] ************************************************ 
ok: [localhost] => {
    "var": {
        "inter_strict": {
            "c": 3
        }
    }
}

TASK: [debug var=diff] ******************************************************** 
ok: [localhost] => {
    "var": {
        "diff": {
            "a": 1,
            "b": 2
        }
    }
}

TASK: [debug var=diff_strict] ************************************************* 
ok: [localhost] => {
    "var": {
        "diff_strict": {
            "a": 1,
            "b": 2,
            "d": 4
        }
    }
}

TASK: [debug var=symdiff] ***************************************************** 
ok: [localhost] => {
    "var": {
        "symdiff": {
            "a": 1,
            "b": 2,
            "e": 3,
            "f": 4
        }
    }
}

TASK: [debug var=symdiff_strict] ********************************************** 
ok: [localhost] => {
    "var": {
        "symdiff_strict": {
            "a": 1,
            "b": 2,
            "d": 4,
            "e": 3,
            "f": 4
        }
    }
}

TASK: [debug var=symdiff_strict_overwrite] ************************************ 
ok: [localhost] => {
    "var": {
        "symdiff_strict_overwrite": {
            "a": 1,
            "b": 2,
            "d": 2,
            "e": 3,
            "f": 4
        }
    }
}

localhost                  : ok=9    changed=0    unreachable=0    failed=0   

```

abadger:
Interesting....  Took me a few minutes to see what you're doing with strict.  It does allow for more flexibility.  What I'd like from overwrite is slightly different though.

If overwrite=False, default, then if (k in a) and (k in b) and a[k] != b[k]: error()
if overwrite=True then if k in b: c[k] = b[k] else: c[k] = a[k]

That way people who are just treating this as a way to add two dictionaries get a diagnostic error when they have the same value in both.  People who actually want to merge the dictionaries can add the parameter to the filter.  They can control which side overwrites by which one they specify first.

bcoca:
so i removed overwrite, kept strict, not sure about raising errors on a filter though

amenonsen:
I've pushed a rebased version of this as #11785. This PR can be closed, but please see discussion in the new PR.

abadger:
So after a long discussion on IRC, I can't see these being added to the current set theory filters as they don't implement set theory.  I can see them going into separately named filters, though (possibly as simple as dict_union, dict_intersection, etc).

So -1 as it stands but +1 if the filters are moved out of the set theory filters.

bcoca:
I don't see them as not following set theory, the limitation you point at is product of the final container (dict)  not being able to have non unique keys and that adds an additional filter, which I think users would expect.

abadger:
@bcoca Right -- and because the container format cannot do the right thing and neither of the choices of what we can make it do make sense in many circumstances this needs to go in its own filter.  I don't think the functionality is bad in its own filter but calling it a set theory function when it can't fulfill the requirements of a set is just a recipe for confusion down the line and once this enters the api of what the set filters can do, there's no good way to remove it later.

bcoca:
i don't see the confusion as it is clear that dicts have unique keys

bcoca:
also I don't see why adding new filters with just dict_<set theory name> changes how the 'violation' of the theory changes

jimi-c:
Based on the above, I believe the dictionary-specific operations should be moved to their own filters, which should make the implementation cleaner and more simple overall, even if it does expand the number of similar filters. I don't like having to add filter params for special cases of items being worked on, and it would help prevent someone from assuming `some_list|union(some_dict)` would work auto-magically.

@amenonsen / @evanccnyc if you'd like to work on reimplemmenting that go ahead, otherwise this will be something we will work on post-2.0. We can discuss further on the devel mailing list or in #ansible-devel on IRC.

Thanks!

PR #10490, opened 2015-03-18
Make it possible to access the HTTP response data in fetch_url

mtotheikle:
##### SUMMARY
Without this change you would not be able to access the http response data via e.read(). This is useful so that you can display errors to users if the API you're communicating with uses non-standard headers for error indication

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME

##### ANSIBLE VERSION
2.3


amenonsen:
Looks fine to me.

jctanner:
The fetch_url function now contains the following docstring ...

```
    :returns: A tuple of (**response**, **info**). Use ``response.body()`` to read the data.
        The **info** contains the 'status' and other meta data. When a HttpError (status > 400)
        occurred then ``info['body']`` contains the error response data::
```

The exception handling block in the function also updates the info dict with the exception data, so this PR has been deprecated.

PR #10493, opened 2015-03-18
dig lookup-plugin: add support for specifying rdclass

jpmens:
##### SUMMARY
This enhances https://github.com/ansible/ansible/pull/10311 by allowing the class to be specified, e.g. for doing `version.bind` lookups.

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
lib/ansible/plugins/lookup/dig.py

##### ANSIBLE VERSION
2.3

### DIFF
--- a/lib/ansible/plugins/lookup/dig.py
+++ b/lib/ansible/plugins/lookup/dig.py
@@ -161,7 +161,7 @@ class LookupModule(LookupBase):
                 elif opt == 'class':
                     try:
                         rdclass = dns.rdataclass.from_text(arg)
-                    except Exception, e:
+                    except Exception as e:
                         raise errors.AnsibleError("dns lookup illegal CLASS: ", str(e))
 
                 continue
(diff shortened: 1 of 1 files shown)


amenonsen:
So it looks like the `dig` lookup plugin doesn't exist in v2/devel at all.

bcoca:
not all lookup plugins have been ported yet

sivel:
@jpmens,

Hi!

jpmens:
Done

lib/ansible/plugins/lookup/dig.py
```diff
@@ -156,6 +158,11 @@ def run(self, terms, variables=None, **kwargs):
                     qtype = arg.upper()
                 elif opt == 'flat':
                     flat = int(arg)
+                elif opt == 'class':
+                    try:
+                        rdclass = dns.rdataclass.from_text(arg)
+                    except Exception, e:
```
bcoca: ```Exception as e```

lookups need to be 2.6/3.5 compatible, not 2.4 (that is for modules)

jpmens: fixed

jpmens:
I don't know what you want from me; this PR is dated TWO YEARS ago.

PR #10512, opened 2015-03-21
V2 added context to template

bcoca:
added support to be able to add an extra vars dictionary to tempate tasks


bcoca:
removing this as having vars: at task level serves the same purpose

PR #10524, opened 2015-03-24
Added installation instructions for Arch Linux.

TwoD:
Hi, thought I'd make a small contribution to the documentation.
I've built the docs and it looks good but I'd be happy to make any adjustments if deemed necessary.

Cheers,
Henrik

### DIFF
--- a/docsite/rst/intro_installation.rst
+++ b/docsite/rst/intro_installation.rst
@@ -261,6 +261,17 @@ Ansible is available for Solaris as `SysV package from OpenCSW <https://www.open
     # pkgadd -d http://get.opencsw.org/now
     # /opt/csw/bin/pkgutil -i ansible
 
+.. _from_pacman:
+
+Latest Releases Via Pacman (Arch Linux)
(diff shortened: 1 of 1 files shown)


bcoca:
We might want to add a note about ansible_python_interpreter for localhost on archlinux.

amenonsen:
The change looks good, and the URL it links to is still valid. It can be merged. I don't know where Python lives on ArchLinux, so I can't say what we should say about ansible_python_interpreter, if anything. @bcoca?

jody-frankowski:
@bcoca @amenonsen On archlinux, from the official packages, python is installed in /usr/bin as:
/usr/bin/python -> python3
/usr/bin/python2 -> python2.7

amenonsen:
So we should recommend setting `ansible_python_interpreter=/usr/bin/python2`. @TwoD, feel like adding that to the commit?

TwoD:
I added the same note Gentoo has, and included a link to the ArchWiki in case someone stumbles across something else and notes it there.

Issue #10445, opened 2015-03-12
ansible_distribution_release is NA for Debian 6.0.10 (squeeze)


tersmitten:
##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

facts.py

##### ANSIBLE VERSION

1.8.4

##### SUMMARY

I'm using Ansible 1.7.2, but the problem still exists in 1.8.4.

``` sh
vagrant@ansible-snmpd-debian-6010:~$ cat /etc/os-release
cat: /etc/os-release: No such file or directory
```

``` sh
vagrant@ansible-snmpd-debian-6010:~$ lsb_release -a
No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux 6.0.10 (squeeze)
Release:    6.0.10
Codename:   squeeze
```

tersmitten:
`ansible_lsb.codename` does contain the needed information

ebastos:
Seems related to issue #4565

ebastos:
As per Debian's official documentation [1] lsb_release is the way to go, but it may not be installed, so this is a partial patch only. Hosts 172.17.0.3 and 172.17.0.7 have lsb-core installed on the example below.

```
$ ANSIBLE_KEEP_REMOTE_FILES=1 ansible test -u root -k -m setup -a 'filter=ansible_distribution_release' -vv
SSH password: 
<172.17.0.3> REMOTE_MODULE setup filter=ansible_distribution_release
<172.17.0.5> REMOTE_MODULE setup filter=ansible_distribution_release
<172.17.0.6> REMOTE_MODULE setup filter=ansible_distribution_release
<172.17.0.7> REMOTE_MODULE setup filter=ansible_distribution_release
172.17.0.5 | success >> {
    "ansible_facts": {
        "ansible_distribution_release": "NA"
    },
    "changed": false
}

172.17.0.6 | success >> {
    "ansible_facts": {
        "ansible_distribution_release": "NA"
    },
    "changed": false
}

172.17.0.7 | success >> {
    "ansible_facts": {
        "ansible_distribution_release": "wheezy"
    },
    "changed": false
}

172.17.0.3 | success >> {
    "ansible_facts": {
        "ansible_distribution_release": "squeeze"
    },
    "changed": false
}
```

The related issue #4565 seems to show there is no easy way around it.

[1] https://www.debian.org/doc/manuals/debian-faq/ch-software.en.html#s-isitdebian

tersmitten:
[...]

PR #10526, opened 2015-03-24
Patch for issue #10445

ebastos:


ebastos:
Not a good solution. I noticed a few problems. Will work on it a little more.

abadger:
A note for your reimplementation -- you may want to leave the scanning of os-release in there as well.  Just do an additional scan of a separate file if it exists (changing the order of the two Debian entries in the list may be necessary if one file should be preferred over the other.)

ebastos:
Yes, thanks.
I submitted a new pull request.

PR #10527, opened 2015-03-24
fixes ansible_distribution release on debian 6.0.10

### LINKED ISSUE #10445: ansible_distribution_release is NA for Debian 6.0.10 (squeeze)

ebastos:
fixes #10445


lib/ansible/module_utils/facts.py
```diff
@@ -99,6 +99,7 @@ class Facts(object):
                     ('/etc/os-release', 'SuSE'),
                     ('/etc/gentoo-release', 'Gentoo'),
                     ('/etc/os-release', 'Debian'),
+                    ('/etc/debian_version', 'Debian'),
```
abadger: There doesn't seem to be any additional parsing happening for this file.  If it's not the same format as os-release then we'd need additional code below to parse it.  (Or perhaps your new code doesn't rely on debian_version so we don't need to add it here.)

ebastos: Yes, I had to add this as /etc/os-release does not exist on Squeeze, only Wheezy and higher.
So adding this path allows the get_distribution_facts() function to work properly, since it loops based on:

`for (path, name) in Facts.OSDIST_LIST:`

lib/ansible/module_utils/facts.py
```diff
@@ -412,6 +413,12 @@ def get_distribution_facts(self):
                                 if release:
                                     self.facts['distribution_release'] = release.groups()[0]
                                 break
+                            else:
+                                try:
+                                    self.get_lsb_facts()
+                                    self.facts['distribution_release'] = self.facts['lsb']['codename']
+                                except:
+                                    continue
                         elif name == 'Mandriva':
```
abadger: If I read this right I think we're relying on get_platfrm() for everything except distribution_release?  If so we probably want the first break (just above your new code) to be in the if-block.  that way we exit once facts['distribution_release'] is successfully found.

If we're only scanning one file for the distribution_release then we probably need to remove the else block and instead have a condition that checks that we're actually looking at a debian system now (from self.facts['distribution'] perhaps?)  if we are not, continue.  If we are, then try the lsb facts to get the distribution_release.

If that's successful break.

If it's unsuccessful, we may break or fail... that depends on whether we're processing one or multiple files to determine the Debian distribution_release.

lib/ansible/module_utils/facts.py
abadger: Right -- but is it the same format as os-release?  If it's not the same format you'll need to add different parsing rules for that file below.

ebastos: Not, it's not the same format. Actually it's very unhelpful:

```
root@aca8dab98b30:/# cat /etc/debian_version 
6.0.8
root@aca8dab98b30:/# cat /etc/os-release
cat: /etc/os-release: No such file or directory
```

While on Wheezy:

```
root@1db7bbc340b3:/# cat /etc/debian_version 
7.6
root@1db7bbc340b3:/# cat /etc/os-release     
PRETTY_NAME="Debian GNU/Linux 7 (wheezy)"
NAME="Debian GNU/Linux"
VERSION_ID="7"
VERSION="7 (wheezy)"
ID=debian
ANSI_COLOR="1;31"
HOME_URL="http://www.debian.org/"
SUPPORT_URL="http://www.debian.org/support/"
BUG_REPORT_URL="http://bugs.debian.org/"
root@1db7bbc340b3:/# 
```

Basically I'm using /etc/debian_version only so I have a file which tells me it's a Debian box and later down I can run the lsb commands.

Thoughts?

ebastos: I installed a newer (Jessie) and older (Lenny) servers and confirmed it works:

```
$ ANSIBLE_KEEP_REMOTE_FILES=1 ansible debian -u root -k -m setup -a 'filter=ansible_distribution_release' 
SSH password: 
172.17.0.6 | success >> {
    "ansible_facts": {
        "ansible_distribution_release": "wheezy"
    },
    "changed": false
}

172.17.0.9 | success >> {
    "ansible_facts": {
        "ansible_distribution_release": "jessie"
    },
    "changed": false
}

172.17.0.8 | success >> {
    "ansible_facts": {
        "ansible_distribution_release": "squeeze"
    },
    "changed": false
}

172.17.0.12 | success >> {
    "ansible_facts": {
        "ansible_distribution_release": "lenny"
    },
    "changed": false
}
```

abadger: If it's not making use of the file then we shouldn't add a file here.  Instead perform the check outside of the file loop altogether.

abadger: Ugh -- so this is where I look at the code and think, this needds to be refactored to make this new case sane :-).

If you don't want to refactor, here's some pseudo-code for how to do it hackily:

``` python
 # Try to handle the exceptions now ...
if self.facts['distribution_name'] == 'Debian':
    try:
        self.get_lsb_facts()
        self.facts['distribution_release'] = self.facts['lsb']['codename']
    except:
        pass
for (path, name) in Facts.OSDIST_LIST:
[...]
            if self.facts['distribution'] in ('Fedora', ) or (self.facts['distribution'] == 'Debian' and self.facts['distribution_release']  != 'NA':
```

Like I say pretty hacky.  But the code needs a refactor anyway.  And we aren't putting false values into the data if we do it this way so future programmers won't try to figure out what debian-version is doing in OSDIST_LIST.

ebastos: I have to leave for the day, but I'll work on that tomorrow.
I think I can take a stab at re-factoring it but I'll probably need some
guidance. Could you help me with that?

ebastos:
@abadger can you, please, comment on this latest commit?

abadger:
So... using a dict there is also something we can't do without more refactoring... the problem is that a dict is unordered (and in newer pythons, the ordering is not stable between runs for security).  So we can't guarantee the order in which distros and files get checked.  This could be a problem, for instance, if the RHEL check happens before the oracle check.  Since oracle provides an /etc/redhat-release the fact check would report that the distribution is RHEL instead of oracle.

If we're going to try for a major refactor, making it more robust against these types of hidden problems would be somewhere on the wishlist.  Also, only checking a specific file once for data (instead of once for every distro that has that file as a source of information).

There are a lot of little gotchas here so I'll definitely understand if you'd rather take the somewhat hacky approach rather than trying to refactor it.

ebastos:
@abadger Took a different approach here. Thoughts?

BTW. I see the built failed, but I ran here and seems to be OK:

```
Ran 231 tests in 10.082s

OK
```

amenonsen:
@ebastos any chance you could rebase this to apply to current devel and resubmit suitably-squashed commits? (It would also be nice if the commit message gave some hint of what the problem is, rather than just a bug # reference.)

ebastos:
@amenonsen @bcoca Finally had time to work on this.

Please, review and advice.

alikins:
https://github.com/ansible/ansible/pull/15221 seems to cover the same ground as this pr. 
I think 15221 is a candidate for the merging soon. Is there anything in this pr that 15221
doesn't cover?

Do the new test cases for distribution facts cover the scenarios this branch 
was fixing?

ebastos:
@alikins This is quite ancient. I think it's safe to assume #15221 is the way to go...

abadger:
Thanks @ebastos.  Closing this and if we need more fixing, we can build off of robinro's refactoring.

PR #10542, opened 2015-03-26
add list comprehensions to handle structured lists

danpilch:
useful for long structured lists and commands that can take multiple item parameters.

E.g. the ansible pip module; brought down the time to install 80 packages from 45 minutes to 2 minutes.

Example usage: 

```
list:
-
  pkg: a
  name: a
- 
  pkg: b
  name: b

"{{ list }} | list_comp_joined('pkg') }}"
```

### DIFF
--- a/lib/ansible/runner/filter_plugins/core.py
+++ b/lib/ansible/runner/filter_plugins/core.py
@@ -270,6 +270,15 @@ def get_encrypted_password(password, hashtype='sha512', salt=None):
 def to_uuid(string):
     return str(uuid.uuid5(UUID_NAMESPACE_ANSIBLE, str(string)))
 
+def list_comp(lst, elem_key):
+    """return the elem_keys of items in lst, as a list"""
+    return [unicode(i[elem_key]) for i in lst]
+
+def list_comp_joined(lst, elem_key):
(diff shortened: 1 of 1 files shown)


sivel:
I feel like this is very misleading. What you are asking to have is a filter that converts all items in a list to Unicode. Which doesn't really match the names or descriptions of this ticket verry well.

I feel like there are batter approaches here, with already existing jinja templates such as:

`"{{ list|map(attribute='pkg')|join(' ') }}"`

danpilch:
yes, this makes a lot more sense if I am honest... Map is unavailable in ansible 1.8.4 / it was introduced in jinja2 v2.7 what version does ansible currently use?

sivel:
Ansible does not dictate any specific version of jinja2.  Ansible only specifies that it is a requirement.  jinja 2.7 was released on May 20, 2013.  If you are running a version of jinja2 older than that, I would recommend seeking out an upgrade.

amenonsen:
This should be closed. As @sivel pointed out, there's already a way to do this, and the nomenclature used in the PR is misleading.

bcoca:
closing as per above

PR #10543, opened 2015-03-26
Reintroduce documented colon-split ranges

### LINKED ISSUE #9670: cannot express host pattern that is a group slice
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.7.2
##### Environment:

N/A
##### Summary:

Host pattern to do list slicing does not work. As described in http://docs.ansible.com/intro_patterns.html one should be able to express:

```
hosts: db[1:5]
```

or

```
hosts: db[1:]
```

but this will result in no hosts found.
##### Steps To Reproduce:

```
~/src/derp> cat hosts
[db]
host1
host2
~/src/derp> cat test.yaml

---
- name: test hosts
  hosts: db[1:]
  tasks:
    - name: do something
      debug: msg="do something"
~/src/derp> ansible-playbook -i hosts test.yaml --list-hosts

playbook: test.yaml

  play #1 (test hosts): host count=0
```
##### Expected Results:

host2 should have been selected. It's the one you'd get from expressing db[1:] as it's the only item in the list other than the first one.

```
~/src/derp> ansible-playbook -i hosts test.yaml --list-hosts

playbook: test.yaml

  play #1 (test hosts): host count=1
    host2
``

##### Actual Results:

No hosts are selected. This is due to a line in lib/ansible/inventory/__init__.py in def get_hosts(self, pattern="all"):
```

```
    patterns = pattern.replace(";",":").split(":")
```

```
The pattern passed in is "db[1:]" which will get split into
"db[1" and "]" which is obviously not going to work.

Searching history, seems this has been the case for quite a while, so I'm confused how this ever worked.
```

willthames:
Allow `webservers[0:25]` as documented in
http://docs.ansible.com/intro_patterns.html

That's the first 26 webservers (so behaves differently
to `webservers[0-25]` which is 25 hosts)

Works for both `--limit` and `hosts`
(the test documented in #9670 now works)

Fixes #9670

### DIFF
--- a/lib/ansible/inventory/__init__.py
+++ b/lib/ansible/inventory/__init__.py
@@ -173,8 +173,28 @@ class Inventory(object):
                 results.append(item)
         return results
 
+    def _split_pattern_groups(self, patterns):
+        """
+        takes e.g. "webservers[0:5]:dbservers:others"
+        and returns ["webservers[0:5]", "dbservers", "others"]
+        """
+        naive_split = patterns.split(':')
+        if len(naive_split) == 1:
+            return naive_split
+        results = []
+        for i in range(0,len(naive_split)):
+            # recombine "webservers[0", "5]"
+            if len(naive_split) > i and \
+               "[" in naive_split[i] and "]" not in naive_split[i] and \
+               "[" not in naive_split[i+1] and "]" in naive_split[i+1]:
+               results.append(naive_split[i] + ":" + naive_split[i+1])
+            # allow "webservers[0-5]" but not "5]" 
+            if not "]" in naive_split[i] or "[" in naive_split[i]:
+                results.append(naive_split[i])
+        return results
+
     def get_hosts(self, pattern="all"):
-        """ 
+        """
         find all host names matching a pattern string, taking into account any inventory restrictions or
         applied subsets.
         """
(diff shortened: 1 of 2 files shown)


amenonsen:
I didn't realise until just this moment that webservers[0-25] meant webserver[0],...,webserver[24], i.e. 25 hosts and not 26. The documentation (now) says only "a portion of servers in a group" and doesn't mention ":" ranges at all. I definitely thought that meant `[a-b]` was an inclusive range containing `b-a+1` elements, and would have been unpleasantly surprised to find out otherwise (the hard way).

Unfortunately, it's probably not acceptable to change the meaning of `-` ranges now. Anyone using them probably has enough hosts to not notice the off-by-one easily. I'm in favour of introducing `:` ranges with the sane behaviour (i.e. `0:25` means `0, ..., 25`), documenting precisely what they mean in `intro_patterns.rst`, and leaving the `-` form unchanged but undocumented. Or perhaps documented with a caveat so people are warned of the off-by-one when they encounter `-` ranges in the wild.

Thoughts?

amenonsen:
@willthames points out that the implementation would be considerably simpler if `:` and `-` ranges were handled the same way, but I don't find the option of enshrining the surprising non-inclusive behaviour very palatable.

willthames:
@amenonsen yep, that's the reason I went for the divergent behaviour - don't change the broken existing code/documentation, but make it work right for the new stuff!

amenonsen:
So I dug a little deeper, and `intro_patterns` mentions `webservers[0-25]`, but `intro_inventory` doesn't mention `x-y` ranges at all, only `x:y` ranges, and those behave as one would expect them to. In fact, `x-y` ranges don't seem to work at all in devel. So I think this PR can be closed, and the only change that is needed is to change the `-` to a `:` in `intro_patterns`. (I haven't looked into the situation in v1.9.)

Update: the change above is included in #11811.

jimi-c:
As the above alternate fix was merged, we'll close this now. Thanks!

PR #10545, opened 2015-03-26
cloudstack: common code used in cloudstack modules

resmo:
Common code of CloudStack modules located in ansible-modules-extras.

### DIFF
--- a/lib/ansible/module_utils/cloudstack.py
+++ b/lib/ansible/module_utils/cloudstack.py
@@ -3,19 +3,29 @@
 # (c) 2015, René Moser <mail@renemoser.net>
 #
 # This code is part of Ansible, but is an independent component.
+# This particular file snippet, and this file snippet only, is BSD licensed.
+# Modules you write using this snippet, which is embedded dynamically by Ansible
+# still belong to the author of the module, and may assign their own license
+# to the complete work.
 #
-# Ansible is free software: you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
-# the Free Software Foundation, either version 3 of the License, or
-# (at your option) any later version.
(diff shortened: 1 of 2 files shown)


lib/ansible/module_utils/cloudstack.py
```diff
@@ -0,0 +1,182 @@
...
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible. If not, see <http://www.gnu.org/licenses/>.
```
bcoca: the code in module_utils is BSD licensed, to allow for 3rd party module redistribution w/o having these to be GPL

lib/ansible/module_utils/cloudstack.py
```diff
@@ -0,0 +1,182 @@
...
+
+import sys
+
+try:
+    from cs import CloudStack, CloudStackException, read_config
+except ImportError:
+    print("failed=True " + \
+        "msg='python library cs required: pip install cs'")
+    sys.exit(1)
```
bcoca: as in modules, we do not wan't to print/exit, capture the missing library in a variable and report it missing in a fail_json call.

resmo:
@bcoca Thanks Brian for the feedback. 

BSD is fine, changed as well as the failing in case of missing cs library according your suggestion. Is the implementation okay that way?

PR #10546, opened 2015-03-26
ansible-galaxy: PEP8 compliance and add the missing parser

Lujeni:
- remove unused variables
- add the missing whitespaces
- use pythonic condition
- add a missing parser variable


bcoca:
we don't take reformatting  requests as they break attribution.

PR #10548, opened 2015-03-26
fix ansible-galaxy: add the missing parser variable

Lujeni:
https://github.com/ansible/ansible/blob/e1eca748fd0a7404145de5c591e10c0daf2e7b95/bin/ansible-galaxy#L278 using a variable was not declared.

### DIFF
--- a/bin/ansible-galaxy
+++ b/bin/ansible-galaxy
@@ -261,7 +261,7 @@ def api_get_config(api_server):
     except:
         return None
 
-def api_lookup_role_by_name(api_server, role_name, notify=True):
+def api_lookup_role_by_name(api_server, role_name, parser, notify=True):
     """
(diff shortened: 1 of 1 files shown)


bin/ansible-galaxy
```diff
@@ -703,7 +703,8 @@ def execute_info(args, options, parser):
                 del install_info['version']
             role_info.update(install_info)
 
-        remote_data = api_lookup_role_by_name(api_server, role, False)
+        remote_data = api_lookup_role_by_name(api_server=api_server, role_name=role,
```
bcoca, line 706: why change the call style?

Lujeni, line 706: Explicit is better than implicit. Using this syntax avoid an error if someone add or remove an argument. (e.g. this PR with the parser argument)

PR #10551, opened 2015-03-26
fix consul inventory issue (missing method param)

deimosfr:
Hi,

I'm encountering an issue with the latest version of consul (0.5.0). It looks like the API changed. This update resolve the issue.

Thanks

### DIFF
--- a/plugins/inventory/consul_io.py
+++ b/plugins/inventory/consul_io.py
@@ -212,7 +212,7 @@ class ConsulInventory(object):
     '''loads the data for a sinle node adding it to various groups based on
     metadata retrieved from the kv store and service availablity'''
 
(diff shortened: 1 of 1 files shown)


bcoca:
will this also work with the old api or should we institute some version detection?

deimosfr:
It looks like it changes between v0.3.7 and v0.3.8. Please look at:
http://python-consul.readthedocs.org/en/v0.3.7/#consul-catalog
http://python-consul.readthedocs.org/en/v0.3.8/#consul-catalog

bcoca:
both accept dc as a parameter, the only thing I see changing in those is the order, which should not matter for named parameters.

PR #10559, opened 2015-03-27
removed folding sudo/su to become logic from constants 

### LINKED ISSUE #10547: sudo=true in config confuses argument parsing
### Steps to reproduce

```
# touch play.yml
# ansible-playbook --ask-sudo-pass play.yml

ansible-playbook: error: Sudo arguments ('--sudo', '--sudo-user', and '--ask-sudo-pass') and su arguments ('-su', '--su-user', and '--ask-su-pass') and become arguments ('--become', '--become-user', and '--ask-become-pass') are exclusive of each other
```
### Potential cause

[This code](https://github.com/ansible/ansible/blob/8a0b8629e86efeddec7da5f8976231deee000f7f/v2/ansible/utils/cli.py#L268-L280) migrates `--ask-sudo-pass` to the new `become` options, _but does not remove the sudo setting_. Then [this code](https://github.com/ansible/ansible/blob/8a0b8629e86efeddec7da5f8976231deee000f7f/v2/ansible/utils/cli.py#L283-L301) fails because both settings are set. 
### Version

```
# ansible-playbook --version
ansible-playbook 1.9.0.1
```

bcoca:
as it is already present downstream in playbook/play/tasks, hopefully fixes #10547


fclaerho:
Hello,
quick question: how long do you think it will take to get a new ansible release with this patch? I'm having the same issue with sudo=yes in the ansible.cfg + ansible 1.9. I've applied your changes manually in my environment and it works, but as my playbook and .cfg is for production, I'm left with either giving up on the .cfg sudo= parameter, or patch constants.py in prod or wait for the next release. Of course, I prefer a new upgradable version ;-)
Best regards.

PR #3579, opened 2013-07-18
Implement sudo/sudo_user at the include and role levels

jeromew:
Hello,

After our discussion on the mailing list about 2 weeks ago, here is a pull request regarding the implementation of sudo/sudo_user behavior at the include and role level.

Implementation at the role level requires implementation at the include level because of the way roles are implemented.

There is no automated test since I am not familiar with ansible test suite and I am not sure how sudo/sudo_user behaviors can be tested. I could give it a try if you point me in the right direction.

Nevertheless I did some tests that make me confident in the implementation.

The implementation is what I call "deepest setting wins downward" : 
- any sudo or sudo_user that is set at one point in the include/task hierarchy flows downward
- any sudo or sudo_user that is set at one point in the hierachy replaces any upward setting"

best regards

PR #10566, opened 2015-03-27
readded sudo/su vars to allow role/includes to work with passed sudo/su

### LINKED ISSUE #10550: playbook sudo: yes roles are not executed as sudo.
Issue Type:

Bug

Ansible Version:  1.9.0.1 from ppa stable.

Environment:

Linux Ubuntu 12.04 host to Linux Debian wheezy target.

Summary:

playbook roles are not executed as sudo anymore.

Steps To Reproduce:

consider the following playbook working with 1.7.2 but no more working with the 1.9.0.1:

test.yml

```
- hosts: test
  roles:
    - role: addgroups
      sudo: yes
```

roles/addgroups/tasks/main.yml

```

---
- name: add toto group
  group: name=toto state=present
```

Expected Results:

add the test group without any failure.

Actual Results:

failed: [zoo] => {"failed": true, "name": "test"}
msg: groupadd: Permission denied.
groupadd: cannot lock /etc/group; try again later

digging into the verbose mode we can observe that the addgroups role is still called with sudo with older version i.e. 1.7.2 and not any more called with sudo with 1.9.0.1

1.7.2

```
u'/bin/sh -c \'sudo -k && sudo -H -S -p "[sudo via ansible, key=hgpjwxwhvkcrvjywzmwiwhxnnamxrfln] password: " -u root /bin/sh -c \'"\'"\'echo SUDO-SUCCESS-hgpjwxwhvkcrvjywzmwiwhxnnamxrfln; LANG=C LC_CTYPE=C /usr/bin/python /home/test/.ansible/tmp/ansible-tmp-1427390484.81-80330223816703/group\'"
```

1.9.0.1

```
/bin/sh -c 'LANG=C LC_CTYPE=C /usr/bin/python /home/test/.ansible/tmp/ansible-tmp-1427392790.43-74683051810820/group
```

Cheers

bcoca:
should fix #10550

### DIFF
--- a/lib/ansible/playbook/play.py
+++ b/lib/ansible/playbook/play.py
@@ -577,7 +577,7 @@ class Play(object):
 
             # evaluate privilege escalation vars for current and child tasks
             included_become_vars = {}
-            for k in ["become", "become_user", "become_method", "become_exe"]:
+            for k in ["become", "become_user", "become_method", "become_exe", "sudo", "su", "sudo_user", "su_user"]:
                 if k in x:
                     included_become_vars[k] = x[k]
                 elif k in become_vars:


graingert:
Any chance of a unittest for this? It's a regression that wasn't caught.

bcoca:
Already added to v2 tree, not a regression as it was never a bug previously.

graingert:
This is definitely a regression, as it was working in 1.8.4 and broken in 1.9.0.1...

sivel:
A regression is defined as:

> The _reappearance_ of a bug in a piece of software that had previously been fixed.

So in this case, it's just a bug, not a regression, since it was never a bug before.

graingert:
http://en.wikipedia.org/wiki/Software_regression

bcoca:
this was also not an intended feature in ansible, so by either definition....

ches:
@bcoca What part of this "was also not an intended feature in ansible"? See #3579.

Regardless of differing stances on the semantics of "regression", the 1.9.0.1 Ansible release broke my previously-working configuration runs, and I'm guessing it wasn't just @graingert and me.

Not trying to assign any blame here, just sayin' this is a pretty aggravating bug---would love to see a bugfix release bump, my team runs packaged Ansible releases and I hear an earful when I have to ask people to run from devel.

graingert:
I now have a state:

``` yaml
- name: assert ansible==1.8.4
  assert:
    that: "ansible_version.full == '1.8.4'"
```

In my ansible tasks

bcoca:
it will be in a 1.9.1 release

@ches, i had to remove some code that specifically guarded against that, my guess is that opinions on this 'feature' changed over time. In any case we are supporting it now as a feature.

PR #10587, opened 2015-04-01
dont break everything when one of the vars in inject does not template

### LINKED ISSUE #9944: failing lookup affects copy/template/file even when not part of the task

bcoca:
wait till its used  somewhere else to fail

hopefully fixes #9944

### DIFF
--- a/lib/ansible/playbook/play.py
+++ b/lib/ansible/playbook/play.py
@@ -119,7 +119,11 @@ class Play(object):
         temp_vars = utils.combine_vars(self.vars, self.vars_file_vars)
         temp_vars = utils.combine_vars(temp_vars, self.playbook.extra_vars)
 
-        ds = template(basedir, ds, temp_vars)
+        try:
+            ds = template(basedir, ds, temp_vars)
+        except errors.AnsibleError, e:
+            utils.warning("non fatal error while trying to template play variables: %s" % (str(e)))
+
         ds['tasks'] = _tasks
         ds['handlers'] = _handlers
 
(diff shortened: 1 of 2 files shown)


abadger:
+1  Is there also something that needs to be done for v2?

bcoca:
probably, hopefully we can change how we do this entirely

bcoca:
tested with:

```
- hosts: localhost
  gather_facts: false
  connection: local
  vars:
    aunknown_hosts: "{{lookup('file','/nonexistent')}}"
    known_hosts: "{{lookup('file','/home/bcoca/.ssh/known_hosts')}}"
  tasks:
        - template:
            src: templates/test.j2
            dest: /tmp/

        - template:
            src: templates/test.j2
            dest: /tmp/
          with_items: "{{aunknown_hosts}}"
```

first task should work w/o issues after patch, (before it would not even get there)
second task should always error

PR #10589, opened 2015-04-01
pep8 ec2.py plugin

graingert:
Our codebase requires that all code complies with pep8, to be able to include the ec2.py plugin we had to format it.

It could become increasingly difficult to port changes from ansible to this script back into our codebase if they diverge.

### DIFF
--- a/plugins/inventory/ec2.py
+++ b/plugins/inventory/ec2.py
@@ -22,7 +22,8 @@ you need to define:
 
     export EC2_URL=http://hostname_of_your_cc:port/services/Eucalyptus
 
-For more details, see: http://docs.pythonboto.org/en/latest/boto_config_tut.html
+For more details, see:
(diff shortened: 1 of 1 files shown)


sivel:
You may want to take a look at https://github.com/ansible/ansible/blob/devel/CODING_GUIDELINES.md 

> There is no need to submit code changes for pep8 and pyflake fixes, as these break attribution history. Project leadership will make these periodically.

graingert:
@sivel This makes a lot of sense for ansible packages, but ec2.py is a plugin that users are expected to copy into their repositories.

bcoca:
closing as per rule discussed above

adamchainz:
I would like `ec2.py` de-linted too - I've copied it into two codebases in my time and had to de-lint it each.

PR #10596, opened 2015-04-01
Switch AnsibleBaseBaseYAMLObject's position_info into a property

abadger:


abadger:
Redone and merged so as not to pick up the module commits

PR #10597, opened 2015-04-01
cloudstack: add error result handling in async job

resmo:
The async job polling did not handle error state returned by the API.


jimi-c:
The above was merged in, but for some reason github is not picking that up, so closing manually. Thanks!

PR #10603, opened 2015-04-02
Make inventory-plugin docker.py work on local system, and fix name

bond:
##### SUMMARY
Fixes the containers name, by using the name set by --name, or the first name set in names[], rather than preferring the name of links and dynamic_names. Also properly sets the containers IP from NetworkSettings.IPAddress and uses "localhost" as hostname when running on a local unix-socket, so it doesn't break by passing null-values to Host() while parsing YML/JSON.

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
plugins/inventory/docker.py

##### ANSIBLE VERSION
2.3

### DIFF
--- a/plugins/inventory/docker.py
+++ b/plugins/inventory/docker.py
@@ -263,7 +263,7 @@ def list_groups():
     for host in hosts:
         ssh_port = host.pop('private_ssh_port', None)
         default_ip = host.pop('default_ip', None)
-        hostname = host.get('base_url')
+        hostname = host.get('base_url','localhost')
 
         try:
             client = docker.Client(**host)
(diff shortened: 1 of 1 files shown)


dericcrago:
@bond - I'm not sure that I understand what this fixes. How would I reproduce the issue?

bond:
@dericcrago - Hi, is there anything in particular in description/code which is unclear? This makes the inventory plugin work for me, along with properly naming the containers if a name has been given. Before the change the dynamic-inventory plugin would not work. There is also a (imho) improvement, which adds groups for images - which makes it easy to execute a task on all docker instances of a certain image.

To reproduce it, try running it against a docker via a local socket.

abadger:
@bond Since not everyone, including me, knows what the docker dynamic inventory script is supposed to be doing I'm not sure from the description what is being changed here.  There also seem to be several different issues that are being addressed which makes it hard to simply review the code to see what's being changed.  Some ways to help us review the code for inclusion:
- Give examples of the output before your changes and after the changes have been made
- If an error is occurring before your change, give the traceback or other error message.
- Describe how each code change implements one of the fixes/features that you're describing
- Open a separate pull request for each logical change.

You don't have to do all of these things but each one that you do can help us understand what's happening more quickly and that will help us to review and apply the changes quicker.  Thanks!

bond:
@abadger - I will provide further detail, and open several logical changes where its relevant. Will try to get it done this weekend. Thanks for the feedback! Its my first pull-request on github, so I was unsure how the commits/pull-requests would be handled.

amenonsen:
Hi @bond. Did you get a chance to revise this PR as requested? The changes need to be rebased to apply to current devel (note the inventory plugins now live in `contrib/inventory`).

mckerrj:
Closed for lack of response

PR #10606, opened 2015-04-02
Default hosts file on homebrew installation

raul-gracia:
Specify on the Docs what's the default location for the hosts file where ansible will be looking if installed with homebrew.


sivel:
I believe we are actually working on removing homebrew from the docs altogether.  

Homebrew does many things that make it hard to support ansible from that installation method.

raul-gracia:
Maybe you can still add it to the docs until you remove it altogether.

bcoca:
removed it so this PR becomes obsolete

PR #10615, opened 2015-04-03
cloudstack: fix vm not found by displayname

resmo:
if a user changes the name of a VM, displayname field is going to be changed, which results in VM can not be found unless we also look for the displayname.


jimi-c:
The above was merged in, but for some reason github is not picking that up, so closing manually. Thanks!

Issue #10602, opened 2015-04-02
--force-handlers does not run handlers when a task fails


jder:
##### Issue Type:

Bug report
##### Ansible Version:

I've verified this with:

```
$ ansible --version
ansible 1.9.0.1
  configured module search path = None
```

and:

```
$ ansible --version
ansible 1.8.4
  configured module search path = None
```

and: 

```
$ ansible --version
ansible 1.9 (devel 0603b9c62c) last updated 2015/04/02 10:28:17 (GMT -400)
  lib/ansible/modules/core: (detached HEAD 613961c592) last updated 2015/04/02 10:28:46 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD eb04e45311) last updated 2015/04/02 10:28:19 (GMT -400)
  v2/ansible/modules/core: (detached HEAD 34784b7a61) last updated 2015/03/10 17:59:39 (GMT -400)
  v2/ansible/modules/extras: (detached HEAD 46e316a20a) last updated 2015/04/02 10:28:19 (GMT -400)
  configured module search path = None
```
##### Environment:

I'm running from Mac OS X 10.10, if that makes a difference.
##### Summary:

Using `--force-handlers` doesn't seem to do anything. When passing `--force-handlers`, handlers only run if there are no failures on a host after the notify occurs. This is the same behavior as without `--force-handlers`.
##### Steps To Reproduce:

Run the following playbook with `--force-handlers`:

```

---
-
  hosts: all

  tasks:
    - command: /bin/true
      notify: do ping

    - command: /bin/false

  handlers:
    - name: do ping
      ping: 
```
##### Expected Results:

I would expect the final "do ping" handler to run on all hosts, despite the fact that the /bin/false command failed, because `--force-handlers` is described to "run handlers even if a task fails."
##### Actual Results:

The "do ping" handler does not run:

```
$ ansible-playbook playbooks/test.yml --force-handlers
PLAY [all] ******************************************************************** 

GATHERING FACTS *************************************************************** 
ok: [a-20150330]
ok: [b-20150330]
ok: [c-20150330]

TASK: [command /bin/true] ***************************************************** 
changed: [a-20150330]
changed: [b-20150330]
changed: [c-20150330]

TASK: [command /bin/false] **************************************************** 
[...]

Issue #10498, opened 2015-03-18
Method to run already notified handlers on failure


jsternberg:
We have a problem with our provisioning with Ansible at the current moment related to handlers. We use handlers to restart services after modifying configuration, but we can end up in a bad situation sometimes when a failure happens after the configuration has changed, but before the notifications have run.

```
- name: change configuration
  copy:
    src: important_file.conf
    dest: /path/to/important/location.conf
  notify: restart important service

- name: failure
  fail: msg="I sometimes fail, but sometimes not."
```

In this case, the important service never gets restarted. If you run the playbook again, the configuration won't be changed and it won't trigger the restart, but the playbook run will "succeed".

I'd like a way to configure Ansible so it will run any pending notifies before it ends with a failure message.

It looks like there was some original talk about this feature in #4443, #4766, and #4777. The last was a different request then this (always run handlers). We don't want to always run them, but only when it's appropriate.

bcoca:
running the playbook with --force-handlers will run them even after task failures, currently there is no way to run them after a playbook has failed and was not run with this option.

jsternberg:
Ah, I appear to have misunderstood the conversation in those issues. I thought it just forced all handlers to run whether or not the task changed or not. It appears there's no way to turn this on by default. One of the people was very adamant he didn't want it to be configurable in ansible.cfg for some reason, but it doesn't appear like always using this flag would harm anything.

Can we get an option for this in ansible.cfg that defaults to off?

bcoca:
That is why i kept this ticket open as a feature idea, it has been on my wish list for a while.

jder:
I'm running into this problem, too, but `--force-handlers` doesn't seem to solve this, because it still won't run the handlers on the nodes which failed. (It does run them on other nodes, but this still leaves the nodes which had failures in a bad state.) Is this the expected behavior?

bcoca:
@jder it should run for nodes that fail a task and don't get to run in the rest of the playbook, but it won't work for nodes that are removed from play due to connection failures.

jder:
Thanks, @bcoca. That doesn't seem to be what's happening though. With this playbook:

```
---
-
  hosts: all

  tasks:
    - command: /bin/true
      notify: do ping

    - command: /bin/false
      when: "'a' not in group_names"

  handlers:
    - name: do ping
      ping: 
```

With `--force-handlers`, only the hosts in group 'a' run the ping task (because they skip the failing task). 

[...]

PR #10618, opened 2015-04-04
Fix --force-handlers, and allow it in plays and ansible.cfg

jder:
The `--force-handlers` command line argument was not correctly running
handlers on hosts which had tasks that later failed. This corrects that,
and also allows you to specify `force_handlers` in ansible.cfg or in a
play.

This addresses the bug here: https://github.com/ansible/ansible/issues/10602
And implements the feature here: https://github.com/ansible/ansible/issues/10498

### DIFF
--- a/bin/ansible-playbook
+++ b/bin/ansible-playbook
@@ -97,7 +97,8 @@ def main(args):
         help="one-step-at-a-time: confirm each task before running")
     parser.add_option('--start-at-task', dest='start_at',
         help="start the playbook at the task matching this name")
-    parser.add_option('--force-handlers', dest='force_handlers', action='store_true',
+    parser.add_option('--force-handlers', dest='force_handlers',
+        default=C.DEFAULT_FORCE_HANDLERS, action='store_true',
         help="run handlers even if a task fails")
     parser.add_option('--flush-cache', dest='flush_cache', action='store_true',
         help="clear the fact cache")
--- a/lib/ansible/playbook/__init__.py
+++ b/lib/ansible/playbook/__init__.py
@@ -375,17 +375,17 @@ class PlayBook(object):
 
     # *****************************************************
 
-    def _trim_unavailable_hosts(self, hostlist=[]):
+    def _trim_unavailable_hosts(self, hostlist=[], keep_failed=False):
         ''' returns a list of hosts that haven't failed and aren't dark '''
 
-        return [ h for h in hostlist if (h not in self.stats.failures) and (h not in self.stats.dark)]
+        return [ h for h in hostlist if (keep_failed or h not in self.stats.failures) and (h not in self.stats.dark)]
 
     # *****************************************************
 
-    def _run_task_internal(self, task):
+    def _run_task_internal(self, task, include_failed=False):
         ''' run a particular module step in a playbook '''
 
-        hosts = self._trim_unavailable_hosts(self.inventory.list_hosts(task.play._play_hosts))
+        hosts = self._trim_unavailable_hosts(self.inventory.list_hosts(task.play._play_hosts), keep_failed=include_failed)
         self.inventory.restrict_to(hosts)
 
         runner = ansible.runner.Runner(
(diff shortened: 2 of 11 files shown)


jder:
I should say that the docs here assume this would be part of 1.9.1, so we should modify them if it's not merged until later.

bcoca:
I need to double check, but this seems to keep hosts for subsequent tasks, not just handlers

jder:
@bcoca I believe it only affects which handlers are run because the hosts lists returned from `_trim_unavailable_hosts` is only used in the context of a single task/handler, unless there's some edge case I'm not seeing. I've added a test which checks that tasks don't run on failed hosts if `--force-handlers` is passed.

bin/ansible-playbook
```diff
@@ -147,6 +147,10 @@ def main(args):
     if options.skip_tags is not None:
         skip_tags = options.skip_tags.split(",")
 
+    force_handlers = C.DEFAULT_FORCE_HANDLERS
```
bcoca: the better way to set the default is in the options parser definition, L#100

lib/ansible/playbook/__init__.py
```diff
@@ -75,7 +75,7 @@ def __init__(self,
         diff             = False,
         any_errors_fatal = False,
         vault_password   = False,
-        force_handlers   = False,
+        force_handlers   = C.DEFAULT_FORCE_HANDLERS,
```
bcoca: no need to default x2

bin/ansible-playbook
jder: Yep, good point. I'll do this.

lib/ansible/playbook/__init__.py
jder: Sorry, I was just matching the others like the become family, forks, etc, which have the default both in the parser level and then here again. Maybe that's just for historical reasons?

bin/ansible-playbook
```diff
@@ -206,7 +210,7 @@ def main(args):
             check=options.check,
             diff=options.diff,
             vault_password=vault_pass,
-            force_handlers=options.force_handlers,
+            force_handlers=force_handlers,
```
bcoca: leave this as it was and we don't need lines 151-152

lib/ansible/playbook/__init__.py
bcoca: no, become is .... complicated ....

bcoca:
rest seems OK, if you clear those minor issues and squash commits I'll merge

jder:
@bcoca Thanks for the review. I've made those changes & squashed. 

Do you know if this will make it into 1.9.1? If not, we might want to change the version number referenced in the docs.

bcoca:
so in my testing, this does make handlers execute for more cases but still not all cases

```
- hosts: localhost
  gather_facts: False
  tasks:
    - shell: touch test0
      notify: test0

    - fail:

  handlers:
    - name: test0
      shell: touch test0_{{inventory_hostname}}

- hosts: locals
  gather_facts: False
  tasks:
    - shell: touch test
      notify: test1

    - fail:
      when: inventory_hostname == 'localhost'

  handlers:
    - name: test1
      shell: touch test1_{{inventory_hostname}}

- hosts: locals
  gather_facts: False
  tasks:
    - shell: touch test2
      notify: test2

    - fail:

  handlers:
    - name: test2
      shell: touch test2_{{inventory_hostname}}

```

inventory:

```
[locals]
21  ; this one fails on connection
localhost ansible_connection=local
localhost2 ansible_connection=local
localhost3 ansible_connection=local
```

w/o patch i don't get handler files for test2, but g4et localhost2 and 3 for test1 and localhost for test0. with the patch i also get l2&3 for test2 but not for localhost.

jder:
That looks OK to me; I wouldn't expect to get test2 for localhost; localhost already failed in an earlier play, so it never runs the task with `notify: test2` in it.

I'm "jder" in #ansible if you want to talk this over.

bcoca:
true, not sure why i was expecting it there also

jder:
Thanks, @bcoca!

PR #10620, opened 2015-04-05
make ansible-galaxy version work for branches as well as tree-ish

semifocused:
Squashes WIP merges and replaces https://github.com/ansible/ansible/pull/10478

Version can now be a branch, a tag, or any <tree-ish> supported by git including specific commit IDs.

### DIFF
--- a/lib/ansible/cli/galaxy.py
+++ b/lib/ansible/cli/galaxy.py
@@ -37,6 +37,7 @@ from ansible.galaxy import Galaxy
 from ansible.galaxy.api import GalaxyAPI
 from ansible.galaxy.role import GalaxyRole
 from ansible.playbook.role.requirement import RoleRequirement
+from ansible.playbook.role.requirement import role_spec_parse
 
 class GalaxyCLI(CLI):
 
@@ -332,6 +333,14 @@ class GalaxyCLI(CLI):
 
         self.pager(data)
 
+    def create_role_from_spec(self, role_spec):
+        rdict=role_spec_parse(role_spec)
+        if "role_name" not in rdict:
+            raise AnsibleError("unable to determine role name from specified argument '%s'" % role_spec)
+        rname=rdict["role_name"]
+        del rdict["role_name"]
+        return GalaxyRole(self.galaxy, rname, **rdict)
+
     def execute_install(self):
         """
         Executes the installation action. The args list contains the
@@ -371,18 +380,18 @@ class GalaxyCLI(CLI):
                 else:
                     # roles listed in a file, one per line
                     self.display.deprecated("Non yaml files for role requirements")
-                    for rname in f.readlines():
-                        if rname.startswith("#") or rname.strip() == '':
+                    for role_spec in f.readlines():
+                        if role_spec.startswith("#") or role_spec.strip() == '':
                             continue
-                        roles_left.append(GalaxyRole(self.galaxy, rname.strip()))
+                        roles_left.append(self.create_role_from_spec(role_spec))
                 f.close()
             except (IOError,OSError) as e:
                 raise AnsibleError("Unable to read requirements file (%s): %s" % (role_file, str(e)))
         else:
             # roles were specified directly, so we'll just go out grab them
             # (and their dependencies, unless the user doesn't want us to).
-            for rname in self.args:
-                roles_left.append(GalaxyRole(self.galaxy, rname.strip()))
+            for role_spec in self.args:
+                roles_left.append(self.create_role_from_spec(role_spec))
 
         while len(roles_left) > 0:
             # query the galaxy API for the role data
@@ -437,9 +446,10 @@ class GalaxyCLI(CLI):
                                 self.exit_without_ignore()
                                 continue
 
-                    # download the role. if --no-deps was specified, we stop here,
-                    # otherwise we recursively grab roles and all of their deps.
-                    tmp_file = role.fetch(role_data)
+                        # download the role. if --no-deps was specified, we stop here,
+                        # otherwise we recursively grab roles and all of their deps.
+                        tmp_file = role.fetch(role_data)
+
             if tmp_file:
                 installed = role.install(tmp_file)
                 # we're done with the temp file, clean it up
--- a/lib/ansible/cli/galaxy.py
+++ b/lib/ansible/cli/galaxy.py
@@ -36,8 +36,7 @@ from ansible.errors import AnsibleError, AnsibleOptionsError
 from ansible.galaxy import Galaxy
 from ansible.galaxy.api import GalaxyAPI
 from ansible.galaxy.role import GalaxyRole
-from ansible.playbook.role.requirement import RoleRequirement
-from ansible.playbook.role.requirement import role_spec_parse
+from ansible.playbook.role.requirement import RoleRequirement, role_spec_parse
 
 class GalaxyCLI(CLI):
 
--- a/lib/ansible/cli/galaxy.py
+++ b/lib/ansible/cli/galaxy.py
@@ -418,7 +418,7 @@ class GalaxyCLI(CLI):
                 if role.scm:
                     # create tar file from scm url
                     tmp_file = GalaxyRole.scm_archive_role(role.scm, role.src, role.version, role.name)
-                if role.src:
+                elif role.src:
                     if '://' not in role.src:
                         role_data = self.api.lookup_role_by_name(role.src)
                         if not role_data:
@@ -448,6 +448,10 @@ class GalaxyCLI(CLI):
                         # download the role. if --no-deps was specified, we stop here,
                         # otherwise we recursively grab roles and all of their deps.
                         tmp_file = role.fetch(role_data)
+                    else:
+                        # just download a URL - version will probably be in the URL
+                        role_data={"dummy": True} # must not be empty for role.fetch()
+                        tmp_file = role.fetch(role_data)
 
             if tmp_file:
                 installed = role.install(tmp_file)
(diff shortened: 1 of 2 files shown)


tjanez:
Looks good to me :+1:

mscherer:
For reference, this is also achieved with https://github.com/ansible/ansible/pull/9928

jimi-c:
@bcoca - for v2.

amenonsen:
@semifocused any chance you could rebase this to apply to current devel? You'll need to target `lib/ansible/cli/galaxy.py` instead of `bin/ansible-galaxy`.

semifocused:
Sure. 
- Mick

semifocused:
Sorry all for the delay in cranking through this, but taking a look at this rebase now.

I am seeing quite a few significant functional differences between ansible-galaxy behavior from 1.9 to 2.0.  In the 2.0 branch it no longer appears to be working at all for roles provided as either a named .tar.gz or as a "scm+url".  Both of these were working for me in 1.9.

@amenonsen @bcoca @jimi-c @willthames, I'll take a swing at characterizing the functional issues that I'm seeing so we can get those cases back to working at all, then I can rebase the fix to support branches etc. for the version.

Are these functional differences something that you guys expect given the state of 2.0, or am I off in the weeds?

semifocused:
[This line](https://github.com/ansible/ansible/blob/17f659a143c6917d18851bea4c248d686262e040/lib/ansible/cli/galaxy.py#L346) appears to create the  `GalaxyRole` object without the `scm` or `version` attrributes being set, and the `src` attribute still containing those unsplit data (see sample debug output below).  This is confusing downstream code which attempts to use `role.scm`, `role.version` etc.

```
role.name:  git+git://github.com/semifocused/nucleator-core-siteconfig.git,master
role.scm:  None
role.src:  git+git://github.com/semifocused/nucleator-core-siteconfig.git,master
role.version:  None
```

semifocused:
Looks like the code from this PR wants to be incorporated [here](https://github.com/ansible/ansible/blob/8153c34abf95e7cf061ad9d23e0dbe820aaba7ea/lib/ansible/galaxy/role.py#L78-L89) in lib/galaxy/role.py

semifocused:
It looks like [creation of the tar file from scm url](https://github.com/ansible/ansible/blob/17f659a143c6917d18851bea4c248d686262e040/lib/ansible/cli/galaxy.py#L369) in `lib/ansible/cli/galaxy.py` should be using [`GalaxyRole.fetch_from_scm_archive()`](https://github.com/ansible/ansible/blob/8153c34abf95e7cf061ad9d23e0dbe820aaba7ea/lib/ansible/galaxy/role.py#L57) in `lib/ansible/galaxy/role.py`.  The current code is using `scm_archive_role()` which isn't defined anywhere in v2 (it appears to be deprecated in favor of the `fetch_from_scm_archive()` behavior of `GalaxyRole`).

semifocused:
Notes:
- @bcoca resolved the mismatch between `fetch_from_scm_archive()` and `scm_archive_role()` with commit 514fa73fcdd3428cf16e11d051021ec10900206d, preferring `scm_archive_role()`, so I've rolled forward with that.
- As of 09/27/2015 [This line in galaxy.py](https://github.com/ansible/ansible/blob/2a50957ad8b966c92d260a89ae7027bc02924042/lib/ansible/cli/galaxy.py#L385)
  still creates the  `GalaxyRole` object without the `scm` or `version` attrributes being set.  This PR addresses that by adding a [new function](https://github.com/47lining/ansible/blob/68ceb27989e8c376bc572a165724d5be70f3ba05/lib/ansible/cli/galaxy.py#L336-L343) that uses the existing role utilities to perform the split, then creates and returns a GalaxyRole using the returned info.  The new function is then [invoked with the provided role_spec](https://github.com/47lining/ansible/blob/68ceb27989e8c376bc572a165724d5be70f3ba05/lib/ansible/cli/galaxy.py#L394), rather than creating the GalaxyRole directly when roles are being added for installation.
- I _suspect_ (but @willthames @bcoca @jimi-c please confirm) that we want this sort of role_spec to be respected when roles (including, potentially, scm & version) in a roles file that is specified at the ansible-galaxy command line.  As such I am also [invoking the role_spec parsing behavior when creating roles from a roles_file](https://github.com/47lining/ansible/blob/68ceb27989e8c376bc572a165724d5be70f3ba05/lib/ansible/cli/galaxy.py#L386).  Please let me know if this differs from your intent, if so it's easy enough to back it out.
- I know you all are in the push for v2 release.  I'd love for this to make it in; it's the last mod that we've contributed that currently requires us to run a forked version of ansible.

semifocused:
@amenonsen per your request I've rebased for v2 codebase.  I believe its ready to be merged.  LMK if you have any concerns.

willthames:
I am heavily in favour of this. This partly fixes  #11486 for starters. 

@semifocused note that it doesn't work for empty branch/version in the CSV (should default to master)

`make test_galaxy` in tests/integration is still useful while CSV spec continues to exist.

semifocused:
@willthames - thanks for chiming in!

You mention that

```
 it doesn't work for empty branch/version in the CSV (should default to master)
```

Can you say more words?  It seems to be working with versions specified or without, with empty versions specified, with roles specified in a roles file or with roles specified in args on the cmdline, for either/both ansible galaxy roles or scm roles.  See [sample tests and output in this gist](https://gist.github.com/semifocused/02bee933471e2a58a4f3).

Let me know what I'm missing - happy to address it if you can help me to better understand your comment and my test coverage gap.

Thanks for pointing out `make test_galaxy`.  I gave it a shot and am missing some pre-reqs (no mercurial install here), I'll see if I can sort that out as time permits.

Best,

\- Mick

willthames:
@semifocused ah, I should have been more clear. It's failing on:

```
https://bitbucket.org/willthames/http-ansible-galaxy/get/master.tar.gz,,http-role
```

where the third field is the role name but the second field is empty.

To be fair I've never used this, but it's supposed to allow for people to default to latest but specify a friendly role name.

willthames:
Looking at this further it might be a problem with remote archive files rather than missing version - it doesn't actually download anything when `role.src` is a URL.

amenonsen:
@semifocused Thanks for rebasing this.

Just a few comments based on a quick read-through of the patch: (1) I would like to see a more detailed commit message (especially as someone who doesn't use galaxy, it wasn't easy for me to follow what the patch did, originally; at least, you should reproduce the commit message from #10478, but I think an example of the problem/solution would also be useful). (2) I think both imports from `role.requirement` should be on the same line, i.e. `from ansible.playbook.role.requirement import RoleRequirement, role_spec_parse`, (3) The elaborate devnull/Popen stuff in `role.py` could be written more simply as::

```
try:
    rc = subprocess.call(checkout_cmd, cwd=tempdir)
except IOError, OSError: # no bare 'except', I'm told it's bad style
    raise AnsibleError("…")
```

Also, is the `popen.wait()`/`rc != 0` stuff actually meant to be inside the `except`? It looks like it might have been indented one step too much.

Hope this helps. (Just to be clear, the decision about whether to merge this or not is in no way up to me, but with the changes listed above, I would agree that it's ready to merge.)

amenonsen:
Or maybe the devnull stuff is actually necessary to suppress the output, but the comments about the bare except and the return-code checking still apply.

semifocused:
@amenonsen - thanks for the review and feedback.
1. Added a more useful commit message
2. Combined imports on single line as suggested
3. I was aping prior code that also appears elsewhere using the /dev/null Popen form - I do believe its necessary to suppress output.
4. I added specific exceptions as suggested to resolve the naked except.  I also propagated those to the other places in the code that use this pre-existing form, even though they're strictly outside the scope of this PR.
5. Good catch on the indentation - you were right.  Further, the mis-indentation was causing some silent masking of a failure that would have led to the default branch being installed even if a different version was specified for git scm roles.  Thank you - resolved and better tested on this end now.

Have a look and LMK if you see anything else that needs attention.

semifocused:
@willthames - with [some additional mods](https://github.com/47lining/ansible/commit/cb60a148fdcfec7463efd937c4039678b1a0aa32) to pull lost functionality back into v2, `make test_galaxy` now appears to be installing all roles as intended.  The tests blow up in ansible-playbook - it's trying to find one of the dependent roles with an incorrect role name.  I suspect there's an ansible-playbook fix needed to address that, for which we should probably open an issue.

jimi-c:
@semifocused please do.

willthames:
@semifocused I've rebased again, on top of my PR that now fixes the galaxy test (#12861). 
https://github.com/ansible/ansible/compare/ansible:devel...willthames:galaxy_10620?expand=1

I've pretty much squashed all your commits into one, as the history was starting to make less sense with all the changes to galaxy interwoven in between. 

If you want to pull in my changes and update this PR, at least we keep the PR side of history. 

Of course if you want to take a different approach, I understand!

bcoca:
closed via #12861

semifocused:
@willthames @bcoca 

Perhap's I'm missing something but: #12861 does not appear to pick up the changes from this PR, most notably changes to lib/ansible/cli/galaxy.py and lib/ansible/galaxy/role.py (https://github.com/ansible/ansible/pull/10620/files) and @bcoca has just closed this PR.

Did you really mean to do that @bcoca?

bcoca:
reopened to check this more thoroughly, i had assumed that previous changes plus #12861 had taken care of these features, in any case this PR needs a rebase and has been scheduled for post 2.0 release

semifocused:
@bcoca, based on ongoing dialogue & collaborative work with @willthames and the shift to 2.0, this PR has ended up covering significant functionalty in ansible-galaxy gaps present in 2.0 that were all working and in place prior to 2.0.  The need for #12861 was exposed through those changes made here.

I'm happy to rebase - I have been doing so for the past seven months including the rework to pull previously working functionality back into ansible galaxy.  I sure hope that can be included in 2.0.

bcoca:
what features specifically are missing (currently) from 2.0? this PR added many things way before galaxy was even implemented in 2.0 and we did many things in a different way as we wanted to restructure how things were defined.

willthames:
@bcoca, @semifocused, because of the ongoing changes to galaxy, the main change in this PR can be condensed to 1f22ed9

I also added a test for this to `make test_galaxy`, see
https://github.com/ansible/ansible/compare/ansible:devel...willthames:galaxy_10620?expand=1
for a reasonable rebase. 

Sorry for the confusion caused by my updates to #12861.

semifocused:
Thanks @willthames, that's quite helpful.

> the main change in this PR can be condensed to 1f22ed9

I will rebase this PR 10620 to include that, as well as your incremental test so that @bcoca can consider including it in 2.0.  It would be fantastic to have clean ansible galaxy behavior with 2.0.

willthames:
I've created #13203 in the absence of a rebase in this PR (still giving @semifocused appropriate credit)

It would be great if we could get #13203 merged and this one closed off.

semifocused:
+1 for merge of #13203 then close of this PR.  Thanks again @willthames!

bcoca:
closing this in favor or #13203

PR #10626, opened 2015-04-06
LDAP modules

szacirag:
ldap_entry asserts the presence or absence of entries in an LDAP
directory. ldap_attr asserts the state of an existing entry's attribute
values. Together, they can bootstrap an LDAP database.

Modules include comprehensive documentation and integration tests. The
tests currently run on Debian derivatives with slapd and python-ldap.


szacirag:
Testing LDAP

PR #10630, opened 2015-04-07
Use codecs module while reading & writing json cache file

viesti:
`utils.jsonify` might return unicode strings with non-unicode data in them (like `u'\xe4'`), in which case writing the file fails. Using codecs module fixes this.

### DIFF
--- a/lib/ansible/cache/jsonfile.py
+++ b/lib/ansible/cache/jsonfile.py
@@ -18,6 +18,7 @@
 import os
 import time
 import errno
+import codecs
 
 try:
     import simplejson as json
@@ -57,7 +58,7 @@ class CacheModule(BaseCacheModule):
 
         cachefile = "%s/%s" % (self._cache_dir, key)
         try:
-            f = open( cachefile, 'r')
+            f = codecs.open(cachefile, 'r', encoding='utf-8')
         except (OSError,IOError), e:
             utils.warning("error while trying to write to %s : %s" % (cachefile, str(e)))
         else:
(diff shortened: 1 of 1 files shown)


abadger:
Could you give us a failure case and what you see when it fails?  Some of this seems like it may be translating bytes and unicode in the wrong direction.  (for instance, in python2, unicode type strings will never contain non-unicode data.  u'\xe4'  is the code point for "ä", for instance)

viesti:
On a second read, the description wasn't too clear :) The problem actually is that writing to a file needs to be unicode aware.

```
>>> with open("/tmp/goo.txt", 'w') as f: f.write(u'\xe4')
...
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe4' in position 0: ordinal not in range(128)
```

abadger:
@bcoca: Do we ever want to handle invalid utf-8 in reading from the json cache files?  If so we may want to do that decoding via an explicit call to utils.unicode.to_unicode rather than relying on codecs.open().  Other than that, this looks good.

bcoca:
json is utf8 by definition, we are writing the files, if someone adds data manually or externally, i think it should just fail.

abadger:
Thanks @viesti !

# Merged

Hi!

issue is resolved.

Thank you!

viesti:
Thanks you! Was a bit short of time earlier, but the offending string came from facts of users with Scandinavian alphabets in their name (like 'Hämäläinen').

reinout:
Note: this fix isn't included in the latest stable 1.9.3 release. Apparently "included in next major release" means "the upcoming 2.0 release" and not the 1.9.2 and 1.9.3 bugfix releases.

reinout:
Note: this fix applies to the fact caching mechanism. Writing out the tree with `--tree some/dir` can still fail with a Unicodeencodeerror. The same fix could be applied to `utils/__init__.py`.

PR #10632, opened 2015-04-07
Add environment-handling to roles.

jhriggs:
[See also #8878.]

A new subdirectory, 'environment', is available for roles. The dictionary elements from the main.yml file within are merged with those from previous roles. (Later roles override previous ones.) Finally, the play's environment is merged into the roles' environment variables.

Thus, the (lowest-to-highest) precedence for any environment variable specified in multiple places is: roles (in role-load order) - play - task.

### DIFF
--- a/lib/ansible/playbook/play.py
+++ b/lib/ansible/playbook/play.py
@@ -425,6 +425,7 @@ class Play(object):
         new_handlers    = []
         role_vars_files = []
         defaults_files  = []
+        new_environment = {}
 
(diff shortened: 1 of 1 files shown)


amenonsen:
I like this feature, but it (a) needs to be rebased to apply to current devel, and (b) _definitely_ needs documentation updates.

bcoca:
so after discussing with the team we are going to pass on this feature, we find that the current environment: variable more than sufficient for users.

if they really need to get them from a directory a lookup() can easily supply that functionality

PR #10634, opened 2015-04-07
Don't pad AES-CTR

graingert:
AES-CTR doesn't need to be padded because the CTR mode uses AES like a stream cipher

http://www.cryptopp.com/wiki/CTR_Mode

> The mode does not require padding the plain text to the block size of the cipher.


graingert:
I'll need to fix the tests and add backwards compatibility

graingert:
@bcoca How best to make this backwards compatible with older versions of Ansible?

bcoca:
i'm thinking that this needs to up the version number on the vault type

lib/ansible/utils/vault.py
```diff
@@ -534,8 +529,8 @@ def encrypt(self, data, password):
 
         cipher = AES.new(key1, AES.MODE_CTR, counter=ctr)
 
-        # ENCRYPT PADDED DATA
-        cryptedData = cipher.encrypt(data)
```
graingert: also remove extra whitespace padding... ;)

amenonsen:
I looked at this PR, and at least the premise is perfectly correct: AES-CTR doesn't need padding, and I believe this should be merged. @graingert any chance you could rebase this to apply to current devel? It's a bit difficult to review in its present form. I'll re-review it if you can resubmit. (I'm not a cryptographer, but I've worked with a fair amount of cryptography code.)

graingert:
@amenonsen rebased

amenonsen:
@graingert sorry I wasn't more clear. The code under `v1/` is dead, and the changes need to target e.g. `lib/ansible/parsing/vault/__init__.py` and so on.

amenonsen:
@graingert any plans to make this apply to current devel? It's a desirable change, but time is running out to get it into v2.

graingert:
I'll try and get it fixed for today

amenonsen:
I'm going about fixing this in a different way (this PR's approach of adding another cipher class won't fly anyway). I'll submit a more complete set of fixes soon.

graingert:
@amenonsen It has to be a different cipher class because it's incompatible with the old padded CTR

amenonsen:
@graingert it doesn't _have to_ be a different cipher class, it's just that the code made it inconvenient in the extreme to do anything else. See PR #12130 where I've reworked things until the cipher class is responsible for  handling version differences itself. Now I can revise `encrypt()` however I like and in `decrypt()` just call `decrypt_11()` if `version=='1.1'` (though in practice I won't repeat the entire decrypt function, just the v1.1-specific parts).

So I'm going to remove padding as you suggested, and use Base64 as @mgedmin suggested in #12121 on top of this infrastructure. You can close this PR. (I would have rebased your commit to retain attribution, but as you can see the underlying code has changed too much for that to be practicable. I'll credit you in the commit message instead.)

amenonsen:
Look at the commit linked above (the last commit in https://github.com/amenonsen/ansible/tree/vault-aes, which depends on the cleanups submitted in PR #12130) for my implementation. It removes the padding and uses base64 while maintaining backwards-compatibility with v1.1 files with just a few lines of code.

amenonsen:
(This PR should be closed now.)

PR #10637, opened 2015-04-08
Print error message when no_log is True

willthames:
Without an error message, many conditions are impossible to
further diagnose.

There is a risk that the error message might contain sensitive
data but this information is more likely to be contained in
other parts of the data structure.

As an example, the current postgresql_user module errors when a DB is readonly (even if there is no change to the user - but that's a separate PR)

no_log unset: 

```
$ ansible-playbook change_db_user.yml -l db-host

PLAY [db-host] ************************************************* 

ok: [db-host]

TASK: [add postgresql user] *************************************************** 
failed: [db-host] => {"failed": true}
msg: ERROR:  cannot execute ALTER ROLE in a read-only transaction
```

no_log set:

```
$ ansible-playbook change_db_user.yml -l db-host

PLAY [db-host] ************************************************* 

ok: [db-host]

TASK: [add postgresql user] *************************************************** 
failed: [db-host] => {"censored": "results hidden due to no_log parameter", "failed": true}
```

with this change

```
$ ansible-playbook change_db_user.yml -l db-host

PLAY [db-host] ************************************************* 

ok: [db-host]

TASK: [add postgresql user] *************************************************** 
failed: [db-host] => {"censored": "results hidden due to no_log parameter", "failed": true}
msg: ERROR:  cannot execute ALTER ROLE in a read-only transaction
```

Note that most of the censorship still occurs (which is really important when using with_items, otherwise the items, including passwords, get displayed)

### DIFF
--- a/lib/ansible/utils/__init__.py
+++ b/lib/ansible/utils/__init__.py
@@ -1636,9 +1636,12 @@ def censor_unlogged_data(data):
     '''
     new_data = {}
     for (x,y) in data.iteritems():
-       if x in [ 'skipped', 'changed', 'failed', 'rc' ]:
-           new_data[x] = y
+        if x in [ 'skipped', 'changed', 'failed', 'rc', 'msg' ]:
+            new_data[x] = y
     new_data['censored'] = 'results hidden due to no_log parameter'
+    # retain msg on failure as diagnosis is difficult otherwise
+    if 'failed' in data and 'msg' in data:
+        new_data['msg'] = data['msg']
     return new_data
 
 def check_mutually_exclusive_privilege(options, parser):
--- a/test/units/TestUtils.py
+++ b/test/units/TestUtils.py
@@ -780,7 +780,7 @@ class TestUtils(unittest.TestCase):
         assert 'failed' in data
         assert 'changed' in data
         assert 'skipped' in data
-        assert 'msg' not in data
+        assert 'msg' in data
         assert data['censored'] == 'results hidden due to no_log parameter'
 
     def test_repo_url_to_role_name(self):


abadger:
I can see how msg could be seen as being immune but I think for no_log we have to err on the side of caution.  For debugging I think that the answer would be to turn off no_log if you need to debug and then turn it back on when you figure out the problem and need to push it into production.

That leaves the choice of whether to log the value of msg or not in the hands of the playbook writer whereas adding msg to that whitelist means that any msg that does contain sensitive data cannot be disabled via the playbook.

bcoca:
I'm going to close this request as msg does contain paramter values that can be secret in many modules and this request hardcodes the display, vs the existing state in which you can just turn off no_log in the play if you really need the info.

willthames:
Yeah, fair enough, I can't disagree with either of you too much.

My use case was when working with an upstream role (for which I actually contributed turning on `no_log`) - obviously you can edit it once you've downloaded it. 

There might be a case for an extra optional message for fail (e.g. sanitised_msg) that modules could choose to set and could be displayed when no_log is True - if that seems a reasonable approach I might code something up but might be worth discussing on ansible-devel or something first - it might just be not that big a deal as you say.

bcoca:
I would even be fine with doing this on msg but only after auditing all modules to remove the possible sensitive interpolated data. Doing this with the existing or a new (safe_msg, sanatized_msg) is still a lot of work and requires revising all modules.  I agree this is a good topic to bring up in ansible-devel

abadger:
Yeah, sounds reasonable in general and discussing on the mailing list for
how people would find it most useful sounds good.  Maybe a command line
switch that ignores no_log or does the sanitization as you suggest?

Issue #10577, opened 2015-03-30
Add switch to force WinRM connection type


peterloron:
Currently, if the system detects that Kerberos libraries are installed, it forces a Kerberos auth attempt. On my OS X 10.10.1 machine, this causes connections to fail since my machine is not joined to a domain nor have any Kerberos tickets been granted.

Please add a switch to force the WinRM connection type.

scentoni:
I think I'm having the same problem. Upgrading to 1.9.0.1 on my CentOS 6.6 host prevents me from connecting to Windows machines over WinRM. I get the error

```
GSSError: (('Unspecified GSS failure.  Minor code may provide more information', 851968), ("Credentials cache file '/tmp/krb5cc_0' not found", -1765328189))
```

The error goes away if I uninstall the Python `kerberos` module. Is there any other workaround for 1.9.0.1? Is there another thread I should be following? Thanks.

peterloron:
The only way I found to work around the issue (on OS X, anyway) was to use kinit to get a ticket for the correct user. There really needs to be a switch.

jhawkesworth:
I took a little look at the code earlier and I think it needs to change.

I can't find the thread /  PR right now but in the past there was a discussion somewhere regarding how to handle switching between domain and non domain login.  At the time having an explicit switch wasn't favoured, although I don't recall it being a final decision.  However, if I recall there was a suggestion that if the username looked like a domain user then it should use domain auth otherwise not.  When I say 'looks like a domain user', the  username would have to look like domuser@some.domain.name or domuser\somedomain - an ordinary user would just look like 'Administrator'.  Would this work for you?  

If I recall @ and \ are not allowed in ordinary windows usernames.

peterloron:
Switching the behavior based on the structure of the username would work for me. Thanks!

cchurch:
@bcoca and I had discussed the same thing.  A plain username would use basic auth, whereas a username with `@` would use kerberos.  The form `username@realm` would explicitly pass the kerberos realm, and `username@` would mean to use the hostname (or ansible_ssh_host) for the realm.  Thoughts?

In the future, a username of `DOMAIN\username` would use NTLM auth.

peterloron:
Yup, that would work fine!

jhawkesworth:
Created https://github.com/ansible/ansible/pull/10644 which I hope should fix this.
Please test - I've only been able to do very basic testing at the moment.

Jon

PR #10644, opened 2015-04-08
winrm.py only use kerberos if the self.user looks like a domain user

jhawkesworth:
Intended to fix
https://github.com/ansible/ansible/issues/10577

Please test!

Jon


bcoca:
@cchurch

cchurch:
Added via changes in 5675982.

PR #10652, opened 2015-04-08
Support querying systemd container information

vodik:
systemd writes a /run/systemd/container file in any container it starts to make it really easy to detect the container type.

This adds support for detecting systemd-nspawn containers (and any other container format that will write data there for compatibility). Otherwise we have to look for  `1:name=systemd:/system.slice/` in `/proc/1/cgroup`, which I'm not confident will be unique enough.

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -2394,6 +2394,12 @@ class LinuxVirtual(Virtual):
                 self.facts['virtualization_role'] = 'guest'
             return
 
+        systemd_container = get_file_content('/run/systemd/container')
+        if systemd_container:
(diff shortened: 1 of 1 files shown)


lib/ansible/module_utils/facts.py
```diff
@@ -2394,6 +2394,12 @@ def get_virtual_facts(self):
                 self.facts['virtualization_role'] = 'guest'
             return
 
+        if os.path.exists('/run/systemd/container'):
```
bcoca: use the get_file_content function instead

vodik:
Alright, I'll update this later today. Thanks

vodik:
Bump?

amenonsen:
Looks sensible to me.

PR #10668, opened 2015-04-11
Add create_lock_file to basic utilities

DavidWittman:
Prevents simultaneous execution of a module by opening an exclusive lock file. The lock file is automatically closed and removed at the end of the module run.

I needed this capability for some custom modules which do terrible things if executed simultaneously, and there are likely some other existing modules which could benefit from similar functionality (any long running tasks like filesystem, logical volume or VM creation).

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -65,6 +65,7 @@ import pwd
 import platform
 import errno
 import tempfile
+import fcntl
 
 try:
     import json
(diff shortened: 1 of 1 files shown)


DavidWittman:
Would this be better suited in its own module under module_utils? I don't want to bloat `basic.py` any more than it needs to be.

amenonsen:
I'm not a module author, so I can't comment usefully. The code certainly looks reasonable (and even has useful comments!). @DavidWittman any interest in rebasing this to apply to current devel?

PR #10678, opened 2015-04-13
Update guide_aws.rst

ninjatux:
changed `wait` value according to ec2 module documentation


amenonsen:
This PR just changes "wait: true" to "wait: yes". Since Ansible accepts y/yes/t/true/1 for boolean true, I don't really see the point. The change does no harm, but it doesn't do any good either as far as I can see.

bcoca:
agreed, closing as this does not seem to add anything useful to the docs

PR #10679, opened 2015-04-13
Add setup.py for v2 to allow for pip editable installs.

dorianpula:
setup.py for v2


abadger:
Just a setup.py so that we try to get v2 running in a virtualenv

PR #10683, opened 2015-04-13
Git integration test: remove test for ambiguous .git/branches dir

timsutton:
I have `.git/branches` in some of my cloned repos but not most of them. A bit more digging suggests that they get created under certain conditions but not always, so I don't think the test should always test they exist. I don't see any explicit references to this directory in `git.py` - the way the module lists branches is simply calling `git branch -a`.

More info: http://stackoverflow.com/questions/10398225/what-is-the-git-branches-folder-used-for


abadger:
yep, after reading through that thread I agree with you.  Thanks!

# Merged

Hi!

issue is resolved.

Thank you!

PR #10685, opened 2015-04-13
Add a generic libvirt inventory plugin

detiber:
Adds an inventory plugin for libvirt (libvirt_generic).

Current limitations:
- can only query a single provider at a time
- only queries running guests
- only queries the first interface returned for setting ansible_ssh_host
- grouping by tags requires that the domain xml doc is hardcoded to use https://github.com/ansible/ansible for the xmlns namespace as follows:

```
<domain type='kvm' id='8'>
  <name>my-host</name>
  <memory unit='GiB'>1</memory>
  <metadata xmlns:ansible="https://github.com/ansible/ansible">
    <ansible:tag>deployment-type-{{ deployment_type }}</ansible:tag>
    <ansible:tag>env-{{ cluster }}</ansible:tag>
    <ansible:tag>env-host-type-{{ cluster }}-{{ type }}</ansible:tag>
    <ansible:tag>host-type-{{ type }}</ansible:tag>
  </metadata>
...snip...
</domain>
```

### DIFF
--- a/plugins/inventory/libvirt.ini
+++ b/plugins/inventory/libvirt.ini
@@ -1,0 +1,20 @@
+# Ansible libvirt external inventory script settings
+#
+
+[libvirt]
+
+uri = qemu:///system
+
+# API calls to libvirt can be slow. For this reason, we cache the results of an API
+# call. Set this to the path you want cache files to be written to. Two files
+# will be written to this directory:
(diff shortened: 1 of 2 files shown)


bcoca:
closing as we don't accept feature commits against stable

PR #10686, opened 2015-04-13
Add a generic libvirt inventory plugin

detiber:
##### SUMMARY
Adds an inventory plugin for libvirt (libvirt_generic).

Current limitations:
- can only query a single provider at a time
- only queries running guests
- only queries the first interface returned for setting ansible_ssh_host
- grouping by tags requires that the domain xml doc is hardcoded to use https://github.com/ansible/ansible for the xmlns namespace as follows:

```
<domain type='kvm' id='8'>
  <name>my-host</name>
  <memory unit='GiB'>1</memory>
  <metadata xmlns:ansible="https://github.com/ansible/ansible">
    <ansible:tag>deployment-type-{{ deployment_type }}</ansible:tag>
    <ansible:tag>env-{{ cluster }}</ansible:tag>
    <ansible:tag>env-host-type-{{ cluster }}-{{ type }}</ansible:tag>
    <ansible:tag>host-type-{{ type }}</ansible:tag>
  </metadata>
...snip...
</domain>
```

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
plugins/inventory/libvirt.ini
plugins/inventory/libvirt_generic.py

##### ANSIBLE VERSION
2.3

### DIFF
--- a/plugins/inventory/libvirt.ini
+++ b/plugins/inventory/libvirt.ini
@@ -1,0 +1,6 @@
+# Ansible libvirt external inventory script settings
+#
+
+[libvirt]
+uri = qemu:///system
+
--- a/plugins/inventory/libvirt_generic.py
+++ b/plugins/inventory/libvirt_generic.py
@@ -1,6 +1,6 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 
-"""
+'''
 libvirt external inventory script
 =================================
 
@@ -12,7 +12,7 @@ To use this, copy this file over /etc/ansible/hosts and chmod +x the file.
 This, more or less, allows you to keep one central database containing
 info about all of your managed instances.
 
-"""
+'''
 
 # (c) 2015, Jason DeTiberus <jdetiber@redhat.com>
 #
@@ -36,9 +36,7 @@ info about all of your managed instances.
 import argparse
 import ConfigParser
 import os
-import re
 import sys
-from time import time
 import libvirt
 import xml.etree.ElementTree as ET
 
@@ -49,8 +47,11 @@ except ImportError:
 
 
 class LibvirtInventory(object):
+    ''' libvirt dynamic inventory '''
 
     def __init__(self):
+        ''' Main execution path '''
+
         self.inventory = dict()  # A list of groups and the hosts in that group
         self.cache = dict()  # Details about hosts in the inventory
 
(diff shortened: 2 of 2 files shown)


amenonsen:
I would be interested to use this to manage kvm/libvirt VMs. I notice that devel has a very rudimentary `contrib/inventory/libvirt_lxc.py` that just uses `virsh list`. I only skimmed this PR, but it looks much more complete. It does need a bit of revision, though (e.g. it uses `argparse` but should probably use `optparse` instead).

@detiber any interest in getting this updated and maybe merging with the existing plugin? Some explanations about how to use it would also be helpful.

amenonsen:
Sorry, please ignore what I said about `argparse` vs. `optparse` above, it was based on a misunderstanding. For the record: `argparse` is newer, but doesn't work with older Python (which is a problem only for modules, so using it in inventory plugins should be fine).

detiber:
@amenonsen If anyone feels strongly on preferring optparse vs argparse I can look at swapping it out.

As far as merging with the libvirt_lxc plugin, I'm not sure what you mean as far as documenting specifically how libvirt_generic could be used in place of libvirt_lxc?

bcoca:
ignore the parsing issues, you are good in that respect.

I missed the part about the merging ... not sure what/where that came up.

detiber:
@bcoca It was from @amenonsen's comment

> @detiber any interest in getting this updated and maybe merging with the existing plugin?

I do agree with the need for some documentation, I just wasn't sure of the best place to add it.  It looks like the existing inventory scripts are hit or miss on whether they provide documentation within the comments of the script themselves, or have documentation provided via docs.ansible.com.

If you point me in the correct direction for the preferred way to document the usage of a dynamic inventory plugin, I'd be happy to write something up.

bcoca:
currently most docs are either in the script or the config file it uses, some of them are better documented in the 'cloud' guides as they are part of the setup. For now just do where you think it fits best, later we will need to come back and make this more visible, probably pushing it to docs.ansible.com

mattclay:

mattclay:
@detiber Rebase needed to get tests to pass on Shippable.

jctanner:
@detiber any chance you still have interest in this? We believe it would be a worthwhile addition and also something we should be able to test within shippable.

pcahyna:
This plugin does not work for me anymore, apparently because it outputs only the "_meta" key with host variables, no group/host list. I added output of hosts in https://github.com/pcahyna/ansible/commit/bc2a2cb37f3886dea6efe7d285146703070d1589 (all running VMs appear under a "libvirt" group).
(Otherwise it is a very useful plugin, thanks.)

nitzmahone:
Closing this one, as we're not accepting new inventory scripts in the Ansible distribution at this point- see #33271 for a version of this as a true inventory plugin.

PR #10697, opened 2015-04-14
add ansible_cmdline variable so that notify plugins can obtain the actual command run

bootswithdefer:
The driver behind this change is to allow notify plugins to be able to inject the Ansible command line into their messages.  Not sure if this is the right place in the code, but it works for me.


jimi-c:
Hi @bootswithdefer, the location for "magic" variables is now in `lib/ansible/vars/__init__.py` (in the `get_vars()` method). If you're still interested in adding this, please resubmit this against the latest devel branch.

Thanks!

PR #10710, opened 2015-04-14
Updated the tox.ini file to run multiple ansible versions.

rofinn:
Purpose: so that devs can use tox to run v1 or v2 of ansible with various versions of python.
For example `tox -e py27-v2 will run python2.7 on v2. Currently, only py26 and py27 are run on v1 when
running just`tox` so that we aren't breaking builds.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,21 @@
 [tox]
-envlist = py26,py27
+envlist = {py26,py27}-v{1}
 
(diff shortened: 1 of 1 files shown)


abadger:
Thanks @Rory-Finnegan !

# Merged

Hi!

issue is resolved.

Thank you!

PR #10714, opened 2015-04-14
Fix #10698 by forcing the use C as a locale.

### LINKED ISSUE #10698: Ansible permits ssh to send local LANG and LC_ALL env vars (file misencoding, filename corruption)
Newer ssh versions like to send `LANG` and `LC_ALL` to the remote sshd, so that the remote ssh session has the same locale as the local user's session.

This makes Ansible playbook execution, including things like files created on the remote end, dependent on the environment of the user running the playbook. Files may have different language text and/or be in a different text encoding. File _names_ may have different encodings too, so that the file named `Álvaro` in `en_US.utf-8` will be listed as `??lvaro` in `en_US`, even though it'll actually match the glob `?lvaro`.

E.g.

```
$ LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8 touch "Álvaro"
$ LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8 ls
Álvaro
$ LANG=en_US LC_ALL=en_US ls
??lvaro
```

As far as I can tell Ansible doesn't support overriding environment variables for all tasks that run on a given host / hostgroup, nor does it support the `environment` keyword at the playbook level under the `hosts` dictionary. So it's not currently easy to say "Always use the xx_XX locale for this host".

These errors are often treated as minor and ignore-able, but they're anything but, especially when they result in mixing utf-8 and 1-byte encodings.

**This does not (just) affect people who use non-English languages**. I noticed the problem because I use the `en_AU.utf-8` locale on my workstation, but most of the hosts I manage only have the `C`, `C.UTF-8` and `en_US.UTF-8` locales configured. The system falls back to `C`.

[`perl` is particularly noisy about this](http://stackoverflow.com/q/2499794/398670):

```
$ perl
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
    LANGUAGE = (unset),
    LC_ALL = (unset),
    LC_PAPER = "en_AU.utf8",
    LC_MONETARY = "en_AU.utf8",
    LC_NUMERIC = "en_AU.utf8",
    LC_MEASUREMENT = "en_AU.utf8",
    LC_TIME = "en_AU.utf8",
    LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
```

The `locale` command will also complain:

```
$ locale
locale: Cannot set LC_ALL to default locale: No such file or directory
```

The best solution, IMO, is for Ansible to use the _remote end_'s default locale, by suppressing sending of `LANG` and `LC_ALL` over `ssh` by default.

This is most visible on Debian and Ubuntu systems, which default to

```
AcceptEnv LANG LC_*
```

in `/etc/ssh/sshd_config`.

As a workaround for this issue, users may wish to remove the above `AcceptEnv` directive from the `sshd_config` on the server, thus disallowing locales from being set by the `ssh` client. Alternately, on the _client_ side, users may remove:

```
SendEnv LANG LC_*
```

from their `/etc/ssh/ssh_config`. It does _not_ appear to be possible to disable this on a per-host basis, unfortunately; see:
- http://superuser.com/q/485569/167160
- https://bugzilla.mindrot.org/show_bug.cgi?id=1285
- https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=573316

So the ideal would be for Ansible to override the `LC_` variables and `LANG` before invoking `ssh`, or for Ansible to use its own global configuration file for `ssh`.

---

The simplest workaround is to create a minimal `ansible_ssh_config` in your Ansible project trees, like:

```
Host *
        ForwardAgent no
        ControlMaster=auto
        ControlPersist=60s
```

and then override `ansible`'s default ssh command in an `ansible.cfg` in your project tree, like:

```
[ssh_connection]
ssh_args=-F ansible_ssh_config
```

The downside is that if Ansible's default ssh command changes, you won't see the change.

mscherer:
This might have side effect on command line UI who will be in
english, and this might cause issue to people, but I think
that's a lesser problem that having silent breakage and corruption
of filename.

I did minimal testing with the code ( IE, verify it doesn't break regular ansible ), and i am gonna do more test, but want to push the code to be discussed as a start.

### DIFF
--- a/lib/ansible/runner/connection_plugins/ssh.py
+++ b/lib/ansible/runner/connection_plugins/ssh.py
@@ -49,6 +49,7 @@ class Connection(object):
         self.HASHED_KEY_MAGIC = "|1|"
         self.has_pipelining = True
 
+        self._env = {'LC_ALL':'C', 'LANG':'C' }
         # TODO: add pbrun, pfexec
         self.become_methods_supported=['sudo', 'su', 'pbrun']
 
@@ -107,7 +108,7 @@ class Connection(object):
     def _run(self, cmd, indata):
         if indata:
             # do not use pseudo-pty
-            p = subprocess.Popen(cmd, stdin=subprocess.PIPE,
+            p = subprocess.Popen(cmd, stdin=subprocess.PIPE, env=self._env,
                                      stdout=subprocess.PIPE, stderr=subprocess.PIPE)
             stdin = p.stdin
         else:
@@ -115,12 +116,12 @@ class Connection(object):
             try:
                 # Make sure stdin is a proper (pseudo) pty to avoid: tcgetattr errors
                 master, slave = pty.openpty()
-                p = subprocess.Popen(cmd, stdin=slave,
+                p = subprocess.Popen(cmd, stdin=slave, env=self._env,
                                      stdout=subprocess.PIPE, stderr=subprocess.PIPE)
                 stdin = os.fdopen(master, 'w', 0)
                 os.close(slave)
             except:
-                p = subprocess.Popen(cmd, stdin=subprocess.PIPE,
+                p = subprocess.Popen(cmd, stdin=subprocess.PIPE, env=self._env,
                                      stdout=subprocess.PIPE, stderr=subprocess.PIPE)
                 stdin = p.stdin
 
@@ -446,7 +447,7 @@ class Connection(object):
             cmd += ["sftp"] + self.common_args + [host]
             indata = "get %s %s\n" % (in_path, out_path)
 
-        p = subprocess.Popen(cmd, stdin=subprocess.PIPE,
+        p = subprocess.Popen(cmd, stdin=subprocess.PIPE, env=self._env,
                              stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         self._send_password()
         stdout, stderr = p.communicate(indata)
--- a/v2/ansible/plugins/connections/ssh.py
+++ b/v2/ansible/plugins/connections/ssh.py
@@ -45,6 +45,7 @@ class Connection(ConnectionBase):
         # SSH connection specific init stuff
         self.HASHED_KEY_MAGIC = "|1|"
         self._has_pipelining = True
+        self._env = {'LC_ALL':'C', 'LANG':'C' }
 
         # FIXME: move the lockfile locations to ActionBase?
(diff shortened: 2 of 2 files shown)


bcoca:
we already do this, you can even configure it from ansible.cfg with module_locale setting

abadger:
To solve #10698 we'd have to allow setting it in inventory, not in ansible.cfg.  #10698 is about the controller and managed node having different locales which causes mangled filenames.  Forcing a common locale seems to be a way to work around that.  There might be other problems though -- we hardcode utf8 in a lot of places....

ringerc:
This would be a major BC break - anyone who's been using a utf-8 `LC_CTYPE` will see different, mangled filenames and file contents after this change. So it'd fix #10698 and break _lots_ more systems than it fixed.

I don't think it's safe to do this hard-coded and by default. If you did, it'd be a little better to use `C.UTF-8` because UTF-8 locales are more common, but you'd still break lots of systems that _aren't_ using UTF-8.

I'd be better for Ansible to _clear_ all the `LC_` env vars and `LANG` before invoking ssh, so it doesn't have them in its environment to copy them to the remote end. It would thus use the remote end's default locale. This would be a safer change to make the default, because it'd only change anything in situations where the Ansible management box's locale already differs from the remote's locale. Can you do that by setting `None` as the value of:

```
LANG
LC_ALL
LC_CTYPE
LC_NUMERIC
LC_TIME
LC_COLLATE
LC_MONETARY
LC_MESSAGES
LC_PAPER
LC_NAME
LC_ADDRESS
LC_TELEPHONE
LC_MEASUREMENT
LC_IDENTIFICATION
```

?

I'd be better if OpenSSH offered `-o NoSendEnv` ... but it doesn't, unfortunately, per the docs linked in the original report.

A config option to set `ansible_locale` (or whatever) to override the locale used would be handy, but it'd have to be available on a per-host basis, because you might be managing some hosts that're using UTF-8 encodings, some that're using ISO-8859-1, and some with other single byte encodings. This is common when you have a fleet of modern Linux machines plus some other UNIXes. The ability to set a default locale in ansible.cfg would make sense too, though.

bcoca:
so I'm going to close this and I suggest we discuss this in ansible-devel, we already have a 'default' in ansible.cfg, I'm open to adding a per-host but I think we need to explore this carefully as it can create many more issues.

PR #10727, opened 2015-04-15
Fix Windows paths containing spaces

### LINKED ISSUE #9999: Windows user id (and thus the user profile path) containing a space breaks ansible windows modules
##### Issue Type:

Bug Report
##### Ansible Version:

ansible --version
ansible 1.9 (devel fe53d86a83) last updated 2015/01/07 01:33:18 (GMT -400)
  lib/ansible/modules/core: (detached HEAD 0d551d8d24) last updated 2015/01/07 01:33:50 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD d4f5b6d41a) last updated 2015/01/07 01:33:51 (GMT -400)
  v2/ansible/modules/core: (detached HEAD cb69744bce) last updated 2014/12/23 17:22:30 (GMT -400)
  v2/ansible/modules/extras: (detached HEAD 8a4f07eecd) last updated 2014/12/23 17:22:50 (GMT -400)
##### Environment:

 cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=14.10
DISTRIB_CODENAME=utopic
DISTRIB_DESCRIPTION="Ubuntu 14.10"
uname -p
i686
##### Summary:

I found out the hard way with an existing windows user account with a userid containing a space and thus also the value of the `%TEMP%` environment variable breaks many of the ansible windows modules due to treating the derived tmp path as two separate paths rather than a path containing a space. Example of this issue manifests due to the use of `shlex.split()` in `ansible/lib/ansible/runner/shell_plugins/powershell.py`
##### Steps To Reproduce:

Create a windows admin account user with a userid containing a space.
E.g. "User Name" such that the user profile path is `C:/Users/User Name/...`

Setup winrm using the `ConfigureRemotingForAnsible.ps1` script under this windows user account

Add windows host to `[windows]` group in ansible hosts file, setup host_vars/winhostname for windows host as per ansible windows documentation, etc.

Try using ansible modules such as: 
    ansible winhostname -m win_ping
    ansible winhostname -m script -a anypowershellscript.ps1
##### Expected Results:

Ansible should retrieve the shortname equivalent path from the windows host so that spaces are not included in the value for the `%TEMP%` environment variable and thus the derived "tmp" directory on the windows host used by ansible modules.

Possible workarounds:
1. The trivial case of "Don't do that!" and use a windows account without a space in the userid
2. Somehow get the 8.3 format of the `%TEMP%` directory and other paths that include  `%USERPROFILE%` as part of the directory path.  Windows CMD shell will automagically store the 8.3 format of the path in `%TEMP%` for backwards compatibility with apps that don't handle long names with spaces well, but for some reason PowerShell when run via `winrm` in ansible uses the long path format and thus includes the space in the path in `%TEMP%` as retrieved via `$env:temp` in `powershell.py` when building a path to a "tmp" directory equivalent for windows. Unfortunately getting Powershell to return the 8.3 version of a file system path is non-trivial; [See here](http://technet.microsoft.com/en-us/library/ee692751.aspx) and [here](http://blogs.technet.com/b/heyscriptingguy/archive/2013/08/01/use-powershell-to-display-short-file-and-folder-names.aspx) or possibly somehow call `win32api.GetShortPathName()` on the remote host using similar techniques as demonstrated [here](http://learn-powershell.net/tag/win32api/)
##### Actual Results:

ansible returns errors from PowerShell to the effect of `powershell script c:\Users\User.ps1 not found`

cchurch:
Wrap windows paths with quotes to handle paths containing spaces.  Updates to execution policy fix from #10664 so that parameters passed via splatting will work.  Fixes #9999.

Windows integration tests pass for raw/script/ping/slurp/fetch.

### DIFF
--- a/lib/ansible/runner/action_plugins/fetch.py
+++ b/lib/ansible/runner/action_plugins/fetch.py
@@ -94,7 +94,7 @@ class ActionModule(object):
 
         # calculate the destination name
         if os.path.sep not in conn.shell.join_path('a', ''):
-            source_local = source.replace('\\', '/')
+            source_local = source.strip('"').replace('\\', '/')
         else:
             source_local = source
 
--- a/lib/ansible/runner/connection_plugins/winrm.py
+++ b/lib/ansible/runner/connection_plugins/winrm.py
@@ -154,7 +154,7 @@ class Connection(object):
         else:
             vvv("EXEC %s" % cmd, host=self.host)
         # For script/raw support.
-        if cmd_parts and cmd_parts[0].lower().endswith('.ps1'):
+        if cmd_parts and powershell._unquote(cmd_parts[0]).lower().endswith('.ps1'):
             script = powershell._build_file_cmd(cmd_parts, quote_args=False)
             cmd_parts = powershell._encode_script(script, as_list=True)
         try:
@@ -165,9 +165,10 @@ class Connection(object):
         return (result.status_code, '', result.std_out.encode('utf-8'), result.std_err.encode('utf-8'))
 
     def put_file(self, in_path, out_path):
-        vvv("PUT %s TO %s" % (in_path, out_path), host=self.host)
+        out_path = powershell._unquote(out_path)
+        vvv('PUT "%s" TO "%s"' % (in_path, out_path), host=self.host)
         if not os.path.exists(in_path):
-            raise errors.AnsibleFileNotFound("file or module does not exist: %s" % in_path)
+            raise errors.AnsibleFileNotFound('file or module does not exist: "%s"' % in_path)
         with open(in_path) as in_file:
             in_size = os.path.getsize(in_path)
             script_template = '''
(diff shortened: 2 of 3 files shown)


jhawkesworth:
Hey thanks for this - I was surprised the other the day to get exists.false back from a win_stat on "C:\Program Files" - this is likely the explanation.  Hope to test as soon as I can. Jon

jhawkesworth:
In case anyone sees my previous comment, my problem with win_stat was actually just getting the quoting correct and is not really relevant to this change.

Using a command line like the below works - both with and without the change described in this PR.

``` bash
ansible windowshost -i windows.inventory -m win_stat -a "path='C:\\Program Files (x86)\\Windows NT'" -v
```

cchurch:
Implemented by #11719 instead.

PR #10734, opened 2015-04-15
Remove check of hardcoded AWS region list.

ralphbean:
You can extend boto to point at other regions that are defined in a
private cloud by defining `BOTO_ENDPOINTS` or `endpoints_path` in
the `~/.boto` file.

Ansible was doing a premature check against a hard-coded list of regions
that interrupted this possibility.  This commit removes that and
clarifies what the user can do if they specify a non-AWS region.

### DIFF
--- a/lib/ansible/module_utils/ec2.py
+++ b/lib/ansible/module_utils/ec2.py
@@ -32,20 +32,6 @@ try:
 except:
     HAS_LOOSE_VERSION = False
 
-AWS_REGIONS = [
-    'ap-northeast-1',
-    'ap-southeast-1',
-    'ap-southeast-2',
-    'cn-north-1',
-    'eu-central-1',
-    'eu-west-1',
-    'eu-central-1',
-    'sa-east-1',
-    'us-east-1',
-    'us-west-1',
-    'us-west-2',
-    'us-gov-west-1',
-]
 
 
 def aws_common_argument_spec():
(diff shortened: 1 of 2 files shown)


ralphbean:
Another strategy could be to ask the boto api to give us a list of possible regions, however the `aws_module` appears to be dynamic and is not necessarily available at the time `ec2_argument_spec` is invoked.

abadger:
Sent a message to the mailing list and a few select individuals about whether this seems like a good change: https://groups.google.com/forum/#!topic/ansible-devel/qGSAHZRcmLg

willthames:
:+1: from me - would have eliminated the need for #9419 for starters.

bcoca:
As long as we get all the modules on the same place, some use a 'central'
hardcoded list, others user their own. I would even add the ability to
'expand' the list in each module as it seems the one returned by boto has
not always been 'correct'.

## 

Brian Coca

willthames:
@bcoca 'some' is actually two, IIRC. Most AWS modules now use the common code to look up regions and connect to regions, there are a couple of rare outliers (which you and I discussed in ansible/ansible-modules-core#260). 

I'm not clear what you mean by 'expanding' the list, or even in what actual ways what boto returns is incorrect (unless you mean that just because a boto service module lists a region, doesn't actually mean that you can use that region - but I don't think this change will affect that)

bcoca:
In the past boto has both returned an 'incorrect list' or was not capable
of listing some regions (at least China and gov iirc), I'm just asking for
possibly adding a 'override' that would allow users to add entries they use
even if boto does not return them.

## 

Brian Coca

willthames:
@bcoca, the region is only checked if the connection to that region fails. So it really doesn't matter what's in the list as long as aws_module.connect_to_region(region) works.

abadger:
I think that this change will actually enable users to add to the default boto list (via the environment var or config file).

abadger:
Thanks for the fix!

# Merged

Hi!

issue is resolved.

Thank you!

PR #10737, opened 2015-04-15
Make some of the optional requirements optional for testing -- we'll skip the tests instead

abadger:
We don't want to pull in redis and memcached all the time.

### DIFF
--- a/v2/test/plugins/test_cache.py
+++ b/v2/test/plugins/test_cache.py
@@ -21,9 +21,25 @@ __metaclass__ = type
 
 from ansible.compat.tests import unittest
 from ansible.plugins.cache.base import BaseCacheModule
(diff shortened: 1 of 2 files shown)


v2/test/plugins/test_cache.py
```diff
@@ -21,9 +21,23 @@
...
+
+SKIP = {}
+try:
+    import memcached
+except ImportError:
+    SKIP['memcached'] = True
+else:
+    # Use an else so that the only reason we skip this is for lack of
+    # memcached, not errors importing the plugin
+    from ansible.plugins.cache.memcached import CacheModule as MemcachedCache
```
dorianpula, line 34: I would separate the try-except-else blocks to improve readability of the tests.

dorianpula:
@abadger Looks good and is the right idea to leave out the extra test requirements.

PR #10762, opened 2015-04-18
[Docs] Make shell quoting rules explanation a bit clearer

kevgathuku:
I personally got confused by the use of 'vs' in the explanation.
Maybe this explains it better?


kevgathuku:
:+1:

Issue #10764, opened 2015-04-18
AnsibleModule.fail_json() send error message to stdout make it invisible on the CLI


romain-dartigues:
##### Issue Type:

Bug Report
##### Ansible Version:

```
ansible 2.0.0 (devel f4172fb9da) last updated 2015/04/18 19:42:59 (GMT +200)
  lib/ansible/modules/core: (detached HEAD a19fa6ba48) last updated 2015/04/18 19:43:07 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD df7fcc90d9) last updated 2015/04/18 19:43:15 (GMT +200)
  v2/ansible/modules/core: (detached HEAD 34784b7a61) last updated 2015/04/18 19:43:24 (GMT +200)
  v2/ansible/modules/extras: (detached HEAD df7fcc90d9) last updated 2015/04/18 19:43:35 (GMT +200)
  configured module search path = None
```
##### Environment:

Irrevelant (Slackware GNU/Linux).
##### Summary:

Error message is not shown by [InventoryScript()](https://github.com/ansible/ansible/blob/c3d34a538613eea3d6db03a700211887f697b443/lib/ansible/inventory/script.py#L48) when generated by [AnsibleModule.fail_json()](https://github.com/ansible/ansible/blob/faadb6830899138de2dfcfca3973a898c5ace3a2/lib/ansible/module_utils/basic.py#L1242) because the latter output it to `stdout` and the former expect the error string to be found on `stderr`.
##### Steps To Reproduce:

``` sh-session
$ ln -s /path/to/ansible/modules/core/cloud/digital_ocean/digital_ocean.py
$ ansible -m ping -u root -i digital_ocean.py all
ERROR: Inventory script (digital_ocean.py) had an execution error:
```
##### Expected Results:

``` sh-session
$ ansible -m ping -u root -i digital_ocean.py all
ERROR: Inventory script (digital_ocean.py) had an execution error: {"msg": "this module requires key=value arguments (['<<INCLUDE_ANSIBLE_MODULE_ARGS>>'])", "failed": true}
```
##### Actual Results:

(see steps)

sivel:
Your issue stems from something completely different, in that `/path/to/ansible/modules/core/cloud/digital_ocean/digital_ocean.py` is not the digital ocean inventory script.

That file you reference is the digital ocean module as documented at http://docs.ansible.com/digital_ocean_module.html

[...]

PR #10765, opened 2015-04-18
Makes AnsibleModule.fail_json() write to stderr

romain-dartigues:
Suggestion to fix bug #10764.

I strongly suggest error messages to be sent to `stderr`.


romain-dartigues:
After the discussion from @sivel, I suppose what i wanted to show was more #10776.

bcoca:
closing this as a misunderstanding, the cli itself might need to write these to stderr, but the modules should not

PR #10769, opened 2015-04-19
Add a requires on python-six 1.4.0 ( for add_metaclass )

mscherer:
This also mean that this doesn't run on RHEL 7 as of today.


mscherer:
For the record, here is the rebase request for python-six in RHEL:
https://bugzilla.redhat.com/show_bug.cgi?id=1185409

PR #10774, opened 2015-04-19
Don't blow up on directory access denied when lookng for plugins.

levkowetz:
When recursing through subdirectires looking for plugins, don't blow up
if we hit a subdirectory we're not permitted to list, just move on.

This is particularly annoying when running ansible from $HOME under OS X,
as ansible will recurse into $HOME/Library/, where there are almost always
directories the regular user may not list.

My first experience with running ansible (version 1.9.0.1), having followed
the intro to the point of doing 'ansible all -m ping' , was to get a screenful of
exceptions.  Not a good start for a novice ansible user.

### DIFF
--- a/lib/ansible/utils/plugins.py
+++ b/lib/ansible/utils/plugins.py
@@ -173,7 +173,13 @@ class PluginLoader(object):
         found = None
         for path in [p for p in self._get_paths() if p not in self._searched_paths]:
             if os.path.isdir(path):
-                full_paths = (os.path.join(path, f) for f in os.listdir(path))
+                try:
+                    full_paths = (os.path.join(path, f) for f in os.listdir(path))
+                except OSError as e:
+                    if e.errno == 13:   # Permission denied; skip this directory
+                        continue
+                    else:
+                        raise
                 for full_path in (f for f in full_paths if os.path.isfile(f)):
                     for suffix in suffixes:
                         if full_path.endswith(suffix):


amenonsen:
@levkowetz any chance you could rebase this to target `lib/ansible/plugins/__init__.py` in current devel and refresh this PR with git push -f?

(Also note the typo in the commit message: subdirectires. Actually I think the commit message should just be "Ignore permission errors on subdirectories when looking for plugins")

jimi-c:
It appears another changed has already fixed this. In the current devel branch:
<pre>
            try:
                full_paths = (os.path.join(path, f) for f in os.listdir(path))
            except OSError as e:
                display.warning("Error accessing plugin paths: %s" % to_text(e))
</pre>
Seeing this, we'll close this PR now. Sorry for the late follow-up, and thanks for the submission!

levkowetz:
Hi James,

On 2017-01-06 15:04, James Cammarata wrote:
> It appears another changed has already fixed this. In the current devel branch:
> <pre>
>             try:
>                 full_paths = (os.path.join(path, f) for f in os.listdir(path))
>             except OSError as e:
>                 display.warning("Error accessing plugin paths: %s" % to_text(e))
> </pre>
> Seeing this, we'll close this PR now. Sorry for the late follow-up, and thanks for the submission!

Good to see that it's been fixed.  Thanks for the reply.


Best regards,

	Henrik

PR #10778, opened 2015-04-20
Python 3 compatible `except` statment

pahaz:


bcoca:
for python3 compatibility look at the v2 subtree, we are not taking such patches for the v1.

abadger:
V2 tree is here:

https://github.com/ansible/ansible/tree/devel/v2

Should an exceptions fix should have been merged last week.

-Toshio

PR #10794, opened 2015-04-20
Do not fail if there is 2 threads, fix #10784

### LINKED ISSUE #10784: Lookup module occansionally fails with `cannot create the path for the password lookup: <directory> (error was [Errno 17] File exists: '<directory>')`
## Issue Type

Bug Report
## Ansible Version

1.9.1 rc1
## Environment

[Ansible run from a docker container](https://github.com/generiklimited/ansible-docker) to manage Ubuntu 14.04.
## Summary:

The `lookup` module can occasionally fail when running against multiple hosts. The message received is `fatal: [<IP ADDRESS>] => cannot create the path for the password lookup: <DIRECTORY> (error was [Errno 17] File exists: '<DIRECTORY>')`.

I believe the issue is that [this line](https://github.com/ansible/ansible/blob/5dce745868720379f69c7b307a3addeafbea66e3/v2/ansible/plugins/lookup/password.py#L105) needs to check to see if the error was that the directory already exists.
## Steps To Reproduce:

Note that this issue is occurring within a sizeable Ansible code base, so the exact details might not be correct.
1. Add a variable in vars which gets its value from the `lookup` module (e.g. `mysql_password: "{{ lookup('password', 'credentials/service/mysqlpassword') }}"`)
2. Create a playbook which targets multiple hosts which uses the new variable 
3. Run a number of times, making sure to remove the created file before each run.
## Expected Results:

Password file is created at the correct location and the value stored within it is used on all of the hosts.
## Actual Results:

The `lookup` module will occasionally throw an error complaining that the directory the password file should be created in already exists.

mscherer:
If there is another ansible running who create the directory,
this can fail despites checking, due to a race condition. So
if there is a exception, we check that the file was not properly
created in the mean time before raising a error.

 fixes #10784

### DIFF
--- a/v2/ansible/plugins/lookup/password.py
+++ b/v2/ansible/plugins/lookup/password.py
@@ -102,7 +102,10 @@ class LookupModule(LookupBase):
                     try:
                         os.makedirs(pathdir, mode=0o700)
                     except OSError as e:
-                        raise AnsibleError("cannot create the path for the password lookup: %s (error was %s)" % (pathdir, str(e)))
+                        # verify if the path was not created due to some race condition
+                        # https://github.com/ansible/ansible/issues/10784
+                        if e.errno != errno.EEXIST or os.stat(pathdir).st_mode != 0o700
+                            raise AnsibleError("cannot create the path for the password lookup: %s (error was %s)" % (pathdir, str(e)))
 
                 chars = "".join([getattr(string,c,c) for c in use_chars]).replace('"','').replace("'",'')
                 password = ''.join(random.choice(chars) for _ in range(length))


wjam:
I don't think this will fully fix the issue. There will probably be an issue further down when the two threads try to write the file, with each thread generating different passwords (https://github.com/mscherer/ansible/blob/fix_10784/v2/ansible/plugins/lookup/password.py#L120).

I think the only safe approach would be to change the `open()` to `os.open()` [(source)](http://stackoverflow.com/questions/10978869/safely-create-a-file-if-and-only-if-it-does-not-exist-with-python) and to read in the file if the thread was unable to create the file.

bcoca:
in #11127 we introduced a safe way to make dirs, password lookup has not been migrated yet but should be changed to use the new function when moved.

jimi-c:
This issue has been resolved a different way, so we're going to go ahead and close this PR.

Thanks!

PR #10808, opened 2015-04-21
update documentation for lookups to mention the quoting issue using key=...

edwelker:
... syntax, and recommend key: syntax as an alternative

### DIFF
--- a/docsite/rst/playbooks_lookups.rst
+++ b/docsite/rst/playbooks_lookups.rst
@@ -198,6 +198,25 @@ template)::
 
       - debug: msg="motd value is {{ motd_value }}"
 
+Lookup contexts
+```````````````
+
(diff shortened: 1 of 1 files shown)


amenonsen:
I'm sympathetic to the purpose of this PR, having been bitten by the problem myself, but I don't think this is the right way to handle it. There's no point adding one example of "...'x'..." and another of '..."x"...' and then say "but you should use key: val instead", the example should just use key="{{ something | quote }}", and the PR should also change the example of authorized_key in playbooks_loops.rst. The warning also applies to copy's content= parameter, but I think we discourage use of that already.

@edwelker, do you feel like reworking the PR along the above lines?

amenonsen:
I guess this should be closed now, unless someone feels like rewording and submitting a new PR.

amenonsen:
Closing per above comment.

PR #10812, opened 2015-04-22
Fixes VMware module utils

jcpowermac:
Resolves syntax errors in the `wait_for_tasks`
Removes throw from `find_datacenter_by_name`


jcpowermac:
In further testing we discovered syntax errors, those have been resolved with this PR.

PR #10819, opened 2015-04-23
Fix race-condition in prepare_writeable_dir

allardhoeve:
All ansible invocations will try to create `$HOME/.ansible/cp`. When two runs are triggered at the exact same moment, they will both see `os.path.exists` return `False` and then try to create the dir. They will then collide and one run will fail with `errno.EEXIST`.

We can fix this by simplifying the code: just create the dir and accept failure because of preexistence. This is what the original code intended anyway.

### DIFF
--- a/lib/ansible/utils/__init__.py
+++ b/lib/ansible/utils/__init__.py
@@ -312,11 +312,15 @@ def prepare_writeable_dir(tree,mode=0777):
     # and normalized and free of symlinks
     tree = unfrackpath(tree)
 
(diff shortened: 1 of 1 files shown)


allardhoeve:
(it's not actually a P3 issue :wink:)

bcoca:
fixed via https://github.com/ansible/ansible/pull/11127

allardhoeve:
There is still a race condition in the new function :(

PR #10835, opened 2015-04-25
Make ec2 inventory plugin Python 3 compatible

logston:


abadger:
I hadn't remembered the inventory scripts.  They do run on the local
machine so I think it's fair game to modify them.  However we don't have a
separate copy of them for v2 yet.  So I'm not sure if we want to merge a
change before we make the v2 tree the default.

For the changes themselves, the iteritems => items changes look good.  The
exception syntax update looks good.

For configparser I think that six has a six.moves module that could handle
that a little more transparently.

For the type checking I think switching to isinstance and making use of
six.text_types would work better.

logston:
Thanks so much for your review! I have changed the import for `configparser`. 

Unfortunately, if I used `six.text_types` for type checking we would not capture all of the paths that this script is currently capturing. Currently we are checking whether or not the `type` of `value` is either `str` or `unicode`. If we were to use `six.text_types`, in Python 2, we would only capture the case where `value` was of `type` `unicode`. 

https://pythonhosted.org/six/#constants

logston:
Also, totally understand if this does not get merged in due to a v2 on the way. Would you rather I rebase these changes off of the v2 branch?

abadger:
Oops, sorry.  Wrong constant. Six.string_types.

v2 is actually just a separate directory in the devel branch so no need to
rebase.

logston:
Not sure where you would like me to place this file.

Would `ansible/v2/ansible/plugins/inventory/ec2.py` work?

abadger:
> Not sure where you would like me to place this file.

This is really what I was getting at when I said maybe we should wait to
merge it until after v2 gets moved over.  We didn't make a parallel copy of
the inventory scripts for v2 so there's not really a place to put an
alternate version right now.

Otoh, this should still be compatible with python 2.6 so we probably could
merge it now with no repercussions...

logston:
I'm pretty biased in terms of this code getting merged in so I'm not sure if I should chime in with the pros and cons of merging this PR in.

PR #10838, opened 2015-04-26
Python 3 compatible and style fixes

pahaz:


pahaz:
Is this going to be merged?

bcoca:
probably not in it's current state, the v2 branch has been cut, also most of these are pep8 changes with very little python3 fixes (the exception as)

abadger:
I just glanced quickly over this and I think it is almost all correct -- the vault.py import of `six.binary_types` probably shouldn't be there and the `stdout.strip() # TODO` is probably resolved by marking the string as `b'\r\n'`

As bcoca says we've cut a branch to stage v2 in  -- v2_final.  At some time in the near future we'll merge that back into devel.  If you'd like to rebase against the v2_final branch you could or you can -- lib/ansible is the v2 code in that branch.

As bcoca also notes, it's easier to review a large set of changes when they're all about a single topic.  So it would have been easier to review this if the style changes were in a separate patch from the python3 fixes.  And probably the unittest improvements in a third PR.

amenonsen:
@pahaz do you intend to rebase this to apply to current devel and split it up by topic for ease of review?

abadger:
I'm going to close this ticket as it's getting increasingly out of date with the current state of the code.  I'm hoping to merge #12119 soon which will allow us to start testing that all of the ansible code on the controller is correct in terms of Syntax on python3.  There's two potential next steps after that:
- Figure out how to make the module_utils code run on both python2.4 and python3.4+: We'll need that so we can continue to share code between the modules and the controller code (If this is infeasible we'll have to make copies of just those functions we need on the controller and maintain both copies).
- Start verifying the logic on both python2 and python3 and writing tests that ensure we keep things correct on both.  This will get us into the realm of bytes versus text type pretty deeply.

Issue #9280, opened 2014-10-07
Role dependencies missing version documentation for remote role support


aotd1:
##### Issue Type:

Bug Report
##### Ansible Version:

1.7.1, 1.7.2, 1.8 ( e8aa847 2014-10-06 )
##### Environment:

OS X 10.10
##### Summary:

Remote roles, described as  { role: "git+http://gitlab.local/ansible/git.git,master" } not working. Ansible-playbook parse it as local path.
##### Steps To Reproduce:

Create role with dependences `meta/main.yml`:

``` yaml

---
dependencies:
  - { role: "git+http://gitlab.local/ansible/git.git,master,git" }
  - { role: apt,    tags: environment }
  - { role: nginx,  tags: nginx }

```

Run ansible-playbook, see error:

```
ERROR: cannot find role in /Users/aotd/www/favorites.ansible/roles/git+http:/gitlab.local/ansible/git.git,master or /Users/aotd/www/favorites.ansible/git+http:/gitlab.local/ansible/git.git,master or /etc/ansible/roles/git+http:/gitlab.local/ansible/git.git,master
Ansible failed to complete successfully. Any error output should be
visible above. Please fix these errors and try again.
```
##### Expected Results:

Downloaded and used role from http://gitlab.local/ansible/git.git
As described in [documentation](http://docs.ansible.com/playbooks_roles.html#role-dependencies).
##### Actual Results:

```
ERROR: cannot find role in /Users/aotd/www/favorites.ansible/roles/git+http:/gitlab.local/ansible/git.git,master or /Users/aotd/www/favorites.ansible/git+http:/gitlab.local/ansible/git.git,master or /etc/ansible/roles/git+http:/gitlab.local/ansible/git.git,master
Ansible failed to complete successfully. Any error output should be
visible above. Please fix these errors and try again.
```

jimi-c:
Hi @aotd1, this feature is new in Ansible 1.8, so you will need to be using the devel branch right now for this feature. That documentation is missing a note regarding that, so we will get that updated.

Thanks!

aotd1:
I already tried devel version (1.8 from 2014-10-06, commit e8aa847).
I see code for "git+" syntax in role_spec_parse (`~/ansible/lib/ansible/utils/__init__.py`), but can't understand why it doesn't work (i'm not Python developer).
Looks like this issue #9173 has the same root.

``` bash
aotd$ ansible --version
ansible 1.8
  configured module search path = None
```

[...]

Issue #920, opened 2012-08-21
group_vars/ only working in ansible-playbook


jhoekx:
Variable definitions in the group_vars/ folder are only read in playbook mode. They are not used when using `ansible`.

I have multiple groups with `ansible_python_interpreter` defined as a group variable. YAML-to-INI conversion converted these groups to have their variables in group_vars/<group> and would subsequently fail when running a command on all of them through `ansible`.

mpdehaan:
This is because groups_vars content is loaded as part of vars_files code.   I'm not going to fix this, but will accept a patch.

Meanwhile, you could use the [foo:vars] method of defining group vars and it would be fine.

mpdehaan:
you'll need to do something like:

if not self.is_playbook:
    play = # somehow fake out a play object
    play.update_vars_files_for_host(...)

PR #10846, opened 2015-04-27
Fix for module role dependency declaration.

jaytaylor:
Fix for module role dependency declaration.

Module dependency declaration of git deps is claimed in the [corresponding ansible documentation](http://docs.ansible.com/playbooks_roles.html#role-dependencies).

However it's not working in real life, as [documented in issue #920](https://github.com/ansible/ansible/issues/9280).

When declaring a `git+` dep, errors like this are produced:

```
ERROR: cannot find role in ~/my-ansible-playbook/roles/https:/github.com/jaytaylor/ansible-docker-registry or ...
```

This simple 2-liner fixes the base-case.

### DIFF
--- a/lib/ansible/playbook/play.py
+++ b/lib/ansible/playbook/play.py
@@ -216,6 +216,8 @@ class Play(object):
             role_name = orig_path.get('role', None)
             if role_name is None:
                 raise errors.AnsibleError("expected a role name in dictionary: %s" % orig_path)
+            if role_name.startswith('git+'):
+                role_name = role_name.split('/')[-1]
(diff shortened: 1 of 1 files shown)


willthames:
:-1: 

This is a git specific fix for a bug that doesn't appear to be reproducible (see latest update to #9280). I'm not convinced that this fix doesn't introduce more bugs than it resolves.

bcoca:
note that v2 is imminent and this code changes drastically.

bcoca:
closing this as the uri scheme in role name will not be implemented

Issue #10850, opened 2015-04-28
Retrieving current hostname of device being acted upon, from within a Python module?


lamoni:
In the event of a custom python module requiring connection=local, where is the best injection point for making the hostname currently being acted upon accessible from within a Python module?

As of right now, it does not look like Ansible injects the current hostname into a module, thus the current hostname isn't accessible without using a contrived "host" module argument that gets filled in by a Playbook.

e.g.
Assuming my /etc/ansible/hosts file looks like:
[devices]
router1
router2

ansible all -m "mycustommodule.py" --module-path="/home/lamoni" -connection=local

Some possibilities, assuming I am not just overlooking a way that currently exists to get that hostname:
A. Inject it as a module param in _configure_module (don't like this idea, as we run the risk of collision with actual module args)
B. Resort to having a playbook only for the sake of passing the inventory_hostname through as a module arg
C. Action plugin - I don't like this idea just because it seems like a pretty generic requirement to want the current hostname from within a module

lamoni:
I've submitted PR10851 in reference to this as a possible solution.  In short, I am injecting the conn.host hostname value into module_args (we are already injecting two other values through there ('CHECKMODE', 'NO_LOG'), and I think the current device-under-execution's hostname is a general enough data item that the injection point is suitable.

bcoca:
# List Information

Hi!

Thanks very much for your interest in Ansible.  It sincerely means a lot to us. 

This appears to be a user question, and we'd like to direct these kinds of things to either the mailing list or the IRC channel.
- IRC: #ansible-devel on irc.freenode.net  
- mailing list: https://groups.google.com/forum/#!forum/ansible-devel

If you can stop by there, we'd appreciate it.  This allows us to keep the issue tracker for bugs, pull requests, RFEs and the like.

Thank you once again and we look forward to seeing you on the list or IRC.  Thanks!

PR #10851, opened 2015-04-28
In reference to Issue #10850.  This passes the current hostname of the d...

lamoni:
...evice under execution to the module by injecting it into the module_args.

Info:
When developing an ansible module that should be run as connection=local, we may need the hostname of the device that is "technically" under execution by Ansible (even if we aren't actually connecting to it with ansible).  As of right now, I don't believe there are any injection points where the hostname is actually accessible from within a Python module (other than through explicit module_args, which then means we have the requirement of a contrived 'host' argument for the module, and will have to execute the module through a Playbook to keep the benefits of concurrency), or through action_plugins.

I think the hostname of the device under execution is general enough to warrant being passed into the Ansible Modules without forcing us to rely on action plugins or playbooks.

It should be noted that this can also be useful for non connection=local modules, but I bring that up as it's easier to see why I'd need the hostname (e.g. to spin up my own subsystem SSH session to the device, if Ansible doesn't support it).

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -361,7 +361,7 @@ class AnsibleModule(object):
 
         (self.params, self.args) = self._load_params()
 
-        self._legal_inputs = ['CHECKMODE', 'NO_LOG']
+        self._legal_inputs = ['CHECKMODE', 'NO_LOG', 'HOSTNAME']
         
         self.aliases = self._handle_aliases()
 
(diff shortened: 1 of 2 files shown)


bcoca:
# Question About This One

Hi!

Unfortunately, we're not sure if we want this feature in the program, and I don't want this to seem confrontational.  We don't think this change is needed as any module that required a hostname would just get it passed as a parameter.

Issue #9851, opened 2014-12-18
to_nice_json produces invalid json


kvdveer:
##### Issue Type:

Bug Report
##### Ansible Version:

Reproduced in 1.6.0, 1.7.0, 1.8.0 and 1.8.1
##### Environment:

Ubuntu 14.10, with ansible installed in a virtual env.

$ pip freeze
Babel==1.3
Jinja2==2.7.3
MarkupSafe==0.23
PyYAML==3.11
ansible==1.8.2
argparse==1.2.1
cffi==0.8.6
cliff==1.9.0
cmd2==0.6.8
cryptography==0.7
ecdsa==0.11
enum34==1.0.4
futures==2.2.0
iso8601==0.1.10
jsonpatch==1.9
jsonpointer==1.7
jsonschema==2.4.0
netaddr==0.7.12
netifaces==0.10.4
oslo.config==1.5.0
oslo.i18n==1.1.0
oslo.serialization==1.1.0
oslo.utils==1.1.0
paramiko==1.15.1
pbr==0.10.3
prettytable==0.7.2
pyOpenSSL==0.14
pyasn1==0.1.7
pycparser==2.10
pycrypto==2.6.1
pyparsing==2.0.3
python-cinderclient==1.1.1
python-glanceclient==0.15.0
python-keystoneclient==0.11.2
python-neutronclient==2.3.10
python-novaclient==2.20.0
python-statsd==1.7.2
python-swiftclient==2.3.1
pytz==2014.10
requests==2.5.0
simplejson==3.6.5
six==1.8.0
stevedore==1.1.0
warlock==1.1.0
wsgiref==0.1.2
##### Summary:

`to_nice_json` does not properly escape strings, leading to newlines, quotes and null characters (among others) in the output.
##### Steps To Reproduce:

``` yaml
- hosts: localhost
  tasks:
  - copy:
        content: " [{{item|to_nice_json}}]"
        dest: /tmp/bug
    with_items:
    - {"k": "Quotes \"'\n"}
```
##### Expected Results:

In `/tmp/bug`:

``` js
 [{
    "k": "Quotes \"'\n"
}]
```
##### Actual Results:

In `/tmp/bug`:

``` js
 [{
    "k": "Quotes "'
"
}]
```

jeffb-stell:
I created an integration test for this issue

amenonsen:
I can confirm that this bug has been fixed in devel (and produces exactly the "expected output" above), and this can now be closed.

[...]

PR #10857, opened 2015-04-28
Add test for https://github.com/ansible/ansible/issues/9851

jeffb-stell:
A test case for https://github.com/ansible/ansible/issues/9851

I don't have a solution for the filter problem to commit


amenonsen:
Just a note to say that #9851 is fixed in devel. It may be a good idea to include the test too.

Issue #10869, opened 2015-04-29
Override of inventory_hostname stopped working


c0rner:
When doing overrides of `inventory_hostname` and `ansible_ssh_host` the `inventory_hostname` overrides stopped working resulting in a 'host not found' error.

Example inventory file:

``` ini
[test]
1.2.3.4
```

Example playbook demonstrating problem:

``` yaml

---
- hosts: all
  gather_facts: yes

  tasks:
  - set_fact: ansible_ssh_host="{{ ansible_default_ipv4.address }}"
  - set_fact: inventory_hostname="testname"
  - copy: content="test" dest=/tmp/test
```

Running the example playbook will result in the `copy` task failing with 'host not found'.

bcoca:
So this was never a feature we wanted, messing with inventory_hostname can break many things internally. If you want to add a new inventory hostname we recommend using add_host with the new name. I'm going to close this and the associated pull request.

PR #10870, opened 2015-04-29
Pullrequest for issue #10869

c0rner:
Fixes issue #10869 with override of inventory_hostname


bcoca:
closed as per comments in #10869

Issue #10871, opened 2015-04-29
Password lookup plugin deletes salt in lookup file


redrampage:
Password lookup plugin deletes salt in lookup file. This leads to constant changed status when two tasks uses same lookup, one with 'encrypt' parameter, and other without.
For example:

``` yaml
tasks:
  - name: Create user
    user:
      password: "{{ lookup('password', inventory_dir + '/creds/user/pass' ncrypt=sha512_crypt) }}"
      ...
# Lookup file 'creds/user/pass' now contain password with salt
  - name: Create htpasswd
    htpasswd:
      password: "{{ lookup('password', inventory_dir + '/creds/user/pass') }}"
      ...
# Salt gets deleted from lookup file 'creds/user/pass'
# Next run will create it again and "Create user" task will have 'changed' status
```

redrampage:
Created pull request https://github.com/ansible/ansible/pull/10872

bcoca:
closing this in favor of #10872, we don't need both open

PR #10872, opened 2015-04-29
Removed deletion of salt by 'password' lookup Issue #10871

redrampage:
Removed deletion of salt param from lookup file by 'password'
lookup_filter.
When two dependent roles use single password lookup, with encrypt parameter and without it leads to constant 'changed' status of the one task.

### DIFF
--- a/lib/ansible/runner/lookup_plugins/password.py
+++ b/lib/ansible/runner/lookup_plugins/password.py
@@ -114,11 +114,6 @@ class LookupModule(object):
                     with open(path, 'w') as f:
                         os.chmod(path, 0600)
                         f.write(content + '\n')
-                # crypt not requested, remove salt if present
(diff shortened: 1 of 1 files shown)


clickfreak:
:+1:

amenonsen:
@redrampage the merge conflict is only because the same patch needs to be targeted to lib/ansible/plugins/lookup/password.py to apply to devel. Would you care to resubmit?

amenonsen:
@redrampage the merge conflict is only because the same patch needs to be targeted to lib/ansible/plugins/lookup/password.py to apply to devel. Would you care to resubmit?

amenonsen:
I rebased and submitted this as #11734, this PR can now be closed.

redrampage:
@amenonsen Thank you.

PR #4627, opened 2013-10-21
Implement nesting lookup plugins

srgvg:
As proposed at #4060

This should allow to implement everything and more
than currently is possible in the not documented, not supported,
soon to be deprecated include: + with_items: construct

```
  - debug: msg="{{item}}"
    with_flattened:
    - lookup('first_found', ['/etc/network/interfaces', '/etc/sysconfig/network'])
    - lookup('fileglob', '/boot/*')
    - lookup('file', '/etc/hostname')
    - lookup('lines', 'ls /boot/grub/')
    - lookup('pipe', 'hostname --fqdn')
    - lookup('sequence', 'start=5 end=10 stride=5')
```

A test playbook and it's output can be found at https://gist.github.com/sergevanginderachter/7092035

[Review on Reviewable](https://reviewable.io/reviews/ansible/ansible/4627)

PR #10880, opened 2015-04-29
[subelements lookup] allow subelement property to be nested and add optional 'skip_missing' flag

### LINKED ISSUE #9827: feature request: allow subelement loops to have missing subelements
##### Issue Type:

“Feature Idea”
##### Ansible Version:

ansible 1.7.2
##### Environment:

CentOS 6.6
##### Summary:

Allow with_subelement loops to have missing subelements without error.
##### Steps To Reproduce:

Given the vars:

```

---
users:
  - name: alice
    authorized:
      - /tmp/alice/onekey.pub
      - /tmp/alice/twokey.pub
  - name: bob
    authorized:
      - /tmp/bob/id_rsa.pub
 - name: greg
```

I'd like to run the following code essentially skipping the _greg_ user because the subelement _authorized_ is deliberately missing:

```
- authorized_key: "user={{ item.0.name }} key='{{ lookup('file', item.1) }}'"
  with_subelements:
     - users
     - authorized
```

Currently I get the error:

```
could not find 'authorized' key in iterated item
```
##### Expected Results:

I expect the missing subelement to be skipped
##### Actual Results:

```
could not find 'authorized' key in iterated item
```

### LINKED ISSUE #9863: with_subelements: Allow subelement property to be nested
```
"rax": {
  "instances": [{
    "id": 0,
    "rax_addresses": {
       "public" : [ { "addr": "10.2.2.2" }, { "addr": "10.0.0.1" } ]
     }
  }]
}

- action: some attr=item.0.id ip=item.1.addr
  with_subelements:
    - rax.instances
    - rax_addresses.public 
```
##### Issue Type:

Bug Report
##### Ansible Version:

1.8.2
##### Environment:

Mac OS X 10.10.1
##### Summary:

`with_subelements` should be able to loop over nested property.
The rax_addresses.public or tax_addresses['public'] does not work with the with_subelements. Only the first argument seems to support nested property names.
##### Steps To Reproduce:

See code above
##### Expected Results:

Be able to loop over the addr properties
##### Actual Results:

Fails with can not find `tax_addresses.public`

srgvg:
Allow the subelement property to the subelements lookup plugin to be nested:

```

---
users:
  - user: bob
    password: system-password
    mysql:
      password: mysql-password
      hosts:
        - "%"
        - "{{ ansible_hostname }}"
        - "127.0.0.1"
        - "::1"
        - "localhost"
      privs:
        - "*.*:SELECT"
        - "DB1.*:ALL"
```

you can now refer to the mysql.hosts subkey 

```
- name: Setup MySQL users
  mysql_user: name={{ item.0.user }} password={{ item.0.mysql.password }} host={{ item.1 }} priv={{ item.0.mysql.privs | join('/') }}
  with_subelements:
    - users
    - mysql.hosts
```

closes #9863 

A second set of patches addresses and closes #9827 by implementing an optional third list item, holding a dictionary of flags, allowing 'skip_missing: yes' to be set, which will let the plugin just skip list items that do not contain the given subkey, instead of raising an error.

### DIFF
--- a/docsite/rst/playbooks_loops.rst
+++ b/docsite/rst/playbooks_loops.rst
@@ -189,6 +189,12 @@ Given the mysql hosts and privs subkey lists, you can also iterate over a list i
 Subelements walks a list of hashes (aka dictionaries) and then traverses a list with a given (nested sub-)key inside of those
 records.
 
+Optionally,  you can add a third element to the subelements list, that holds a
+dictionary of flags. Currently you can add the 'skip_missing' flag. If set to
+True, the lookup plugin will skip the lists items that do not contain the given
+subkey. Without this flag, or if that flag is set to False, the plugin will
+yield an error and complain about the missing subkey.
+
 The authorized_key pattern is exactly where it comes up most.
 
 .. _looping_over_integer_sequences:
--- a/docsite/rst/playbooks_loops.rst
+++ b/docsite/rst/playbooks_loops.rst
@@ -147,9 +147,26 @@ How might that be accomplished?  Let's assume you had the following defined and 
         authorized: 
           - /tmp/alice/onekey.pub
           - /tmp/alice/twokey.pub
+        mysql:
+            password: mysql-password
+            hosts:
+              - "%"
+              - "127.0.0.1"
+              - "::1"
+              - "localhost"
+            privs:
+              - "*.*:SELECT"
+              - "DB1.*:ALL"
       - name: bob
         authorized:
           - /tmp/bob/id_rsa.pub
+        mysql:
+            password: other-mysql-password
+            hosts:
+              - "db1"
+            privs:
+              - "*.*:SELECT"
+              - "DB2.*:ALL"
 
 It might happen like so::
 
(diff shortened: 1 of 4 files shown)


wolph:
Just a thought, why use a `.` syntax instead of just more subelement items?

Making it go recursive like this as well might be more flexible:

``` yaml
- name: Setup MySQL users
  mysql_user: name={{ item.0.user }} password={{ item.0.mysql.password }} host={{ item.1 }} priv={{ item.0.mysql.privs | join('/') }}
  with_subelements:
    - users
    - mysql
    - hosts
```

Just in case of a list of hashes.

srgvg:
@WoLpH for starters, it seems to me the dot notation is more in line with the jinja2 notation.
Also, I just bumped into #9827 where we propose to add a parameter to just skip list items that don't have the subkey.

wolph:
True, instead is actually the wrong wording. I was thinking of a case like this:

``` yaml
users:
  - user: bob
    password: system-password
    mysql:
      password: mysql-password
      hosts:
        production:
          -"%"
          - "{{ ansible_hostname }}"
        local:
          - "127.0.0.1"
          - "::1"
          - "localhost"
```

srgvg:
@WoLpH not sure how your last example is any different? 

```
with_subelements:
- users
- mysql.hosts.production
```

would work with my patch.

wolph:
@srvg: Would your current code work if you want to iterate through all available groups? So both local and production? It's probably a remote use-case but it might still be useful :)

srgvg:
No, it wouldn't. you could only point the key to mysql.hosts.production OR mysql.hosts.local.

There should be a more generic way to combine lookup plugins. I proposed a patch for that two years ago (#4627) but I never liked the syntax I had to implement. This should be something to look into post v2.

This being said, my patch here is on the v1 tree, and I heard v2 changes some things in lookup plugins, so I'll need to update this patch for v2.

wolph:
If you need any help with the patches, let me know :)
I'd be happy to do some work as well.

As for the syntax... how about something like this:
1. Every line works through the parent list/dict
2. Every dot is just immediate access (i.e. `__getitem__`)

Still won't cover all cases but it should allow for slightly more complex examples.

srgvg:
Closing this in favour of the one update to v2: #10995

PR #10866, opened 2015-04-29
Allow roles to use dictionary spec in dependencies

willthames:
The docs suggest that the following dependencies
specification may be used in meta/main.yml

```
dependencies:
- { role: "git+https://gitserver/rolename.git" }
```

This change makes that work

Fixes #9280

PR #10882, opened 2015-04-29
allow roles to use dictionary spec in dependencies

### LINKED ISSUE #9280: Role dependencies missing version documentation for remote role support

bitmensch:
PR based on input of previously rejected #10866, now based on v1 playbook code (fixes #9280).

### DIFF
--- a/lib/ansible/playbook/play.py
+++ b/lib/ansible/playbook/play.py
@@ -213,7 +213,7 @@ class Play(object):
         role_vars = {}
         if type(orig_path) == dict:
             # what, not a path?
-            role_name = orig_path.get('role', None)
+            role_name = utils.role_yaml_parse(orig_path)["name"]
             if role_name is None:
                 raise errors.AnsibleError("expected a role name in dictionary: %s" % orig_path)
             role_vars = orig_path


willthames:
I'm not sure what the difference is between this and #10866 - this will likely get rejected for the same reason. Any fix needs to be made in the v2 code base, afaict.

bitmensch:
@willthames From my understanding your original PR #10866 is based on ansible:devel which is currently prepared for v2. As @jimi-c mentioned in his [comment](https://github.com/ansible/ansible/pull/10866#issuecomment-97425483) your PR should only be applied to v1 as v2 may be going to handle this different. That's why I based this PR on ansible:stable-1.9 branch which looks to me like being the current v1 maintenance branch.

@jimi-c Could you please verify if I understand this correctly and probably apply the PR?

bcoca:
we are about to switch to use v2, so adding this to v1 is not a good idea

bitmensch:
@bcoca It is not a feature request I'm suggesting here but a bug-fix. Looking at the [role-dependencies documentation](http://docs.ansible.com/playbooks_roles.html#role-dependencies) it mentions that the following should work when being specified in a role's `meta/main.yml`:

``` yaml
---
dependencies:
  - { role: 'git+http://git.example.com/repos/role-foo,v1.1,foo' }
```

But it does not work as ansible tries to use this role string 1:1 when searching the roles path (which obviously fails). The suggested fix solves this by reusing existing parse functionality to avoid inconsistent behavior.

willthames:
@benjixx I think @bcoca et al. appreciate it's a bug. But with no plans for further 1.x releases, it's better that the fix goes into v2 of Ansible. And it's not a critical bug - there are _heaps_ of workarounds (e.g. just use the role spec outside of the dict

```
---
dependencies:
  - git+http://git.example.com/repos/role-foo,v1.1,foo'
```

works fine, for example
You can always use this fix in your own Ansible in the interim. I'll be testing v2 of galaxy very thoroughly once it arrives.

bitmensch:
@willthames But without using a `dict` I cannot pass any args to the role. Sometimes I want to do something like this:

``` yaml
---
dependencies:
  - role: my-role
    my_role_param1: value1
    my_role_param2: value2
```

But unfortunately I can't find a way to do the same with a role name like `git+http://git.example.com/repos/role-foo,v1.1,foo`.

If you can suggest a workaround for my problem, I would much appreciate it.

willthames:
Sure - I personally think `meta/main.yml` is the worst possible place to pass role variables.

You can pass role variables in your playbook role definition. Or given that you wish to assert a dependency on a role, you can place the role variables in the dependent role's `defaults/main.yml` or `vars/main.yml`

But also, any variables you define in inventory, playbooks, other roles will also be accepted by the role (this is a blessing and a curse and really means that you need to namespace your variables as you are doing)

However, you might have use cases that I haven't thought of, so let me know and I'll see what I can do

bitmensch:
@willthames Thanks for the suggestion. I hadn't thought of using `defaults/main.yml` or `vars/main.yml` to define variables for the as dependency included role, but I think this sounds reasonable for the most cases. I will give it a try. I already prefix all my role variables as it would be a maintenance nightmare otherwise. :)

The problematic restriction I can see is, when it comes to a role that depends multiple times on another role (the allow_duplicates case). Then I would usually want to pass different params to each of the included role. I cannot think of a valid use case right now, but I guess there might be some.
So, when working on v2, please keep that in mind.

bcoca:
closing this but keeping it in mind for v2, we are not implementing the name: uri scheme and using the galaxy format instead (separate src: field).

PR #10899, opened 2015-05-01
Tweak vault description

wenottingham:
Vault isn't specifically for source control. Make description a little more generic and descriptive.


bcoca:
I would not remove the mention of source control, maybe rephrase:
"in encrypted files (which can be source controlled and shared) "

wenottingham:
Updated.

PR #10903, opened 2015-05-01
Make "include" variable documentation consistent

gtback:
There is already a good example of a list variable ("ssh_keys")
contained above this snippet, so reuse the content here.


gtback:
Thanks, @bcoca . I just noticed this still isn't on the [public docs site](http://docs.ansible.com/playbooks_roles.html). Is the site built off of a different branch than `devel`?

bcoca:
no, we just don't update it on check in

gtback:
Ahh, ok. Thanks! I wasn't sure if you were using something like ReadTheDocs that automatically rebuilds when new commits are pushed.

PR #10904, opened 2015-05-02
Bump the TLS version to a more secure version

mscherer:
Since TLS v1 is SSL v3 and SSL v3 is deprecated and insecure,
this commit raise the negociated version to TLS v1.1

However, this requires a newer version of python ( 2.7.9 at least ), so I am not sure on how this can be pushed without breaking something.

### DIFF
--- a/lib/ansible/module_utils/urls.py
+++ b/lib/ansible/module_utils/urls.py
@@ -91,7 +91,7 @@ class CustomHTTPSConnection(httplib.HTTPSConnection):
         if self._tunnel_host:
             self.sock = sock
             self._tunnel()
-        self.sock = ssl.wrap_socket(sock, keyfile=self.key_file, certfile=self.cert_file, ssl_version=ssl.PROTOCOL_TLSv1)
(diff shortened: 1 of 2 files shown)


bcoca:
Actually they are not the same, TLS is a different protocol, SSLv3 and TLS1.0 were released together and TLS is based on SSL3, but it does not share the basic vulnerability that all SSL versions have. We patched v1 a while ago to default to TLSv1 to avoid that vulnerability.

In v2 the security version configurable and it sets default to TLSv1 in the config but it will be easy for users to set it to whatever value they need.

mscherer:
I am maybe a bit confused, but I did check on wikipedia who explain that TLS v1 permit to go back to SSL v3 ( https://en.wikipedia.org/wiki/Transport_Layer_Security#TLS_1.0 ). Now, there is also the issue of the SSL v3 used by default if we do not say what version we want. 

I think fixing this will be quite tricky, since that's for module, so it must be python 2.4 compatible.

bcoca:
downgrading was part of the spec, but it won't work with urllib anymore

bcoca:
closing this as v2 should allow users to update the default as needed and this 'future proves' us for other protocol deprecation by taking it out of the code.

PR #10905, opened 2015-05-02
Option to override playbook hosts parameter with 'all' hosts

adenot:
Sometimes when maintaining servers, I need to run playbooks against hosts there are in different groups from the one specified in the playbook's hosts definition. To avoid having to set all my playbooks to `hosts: all` and use `--limit` to decide which hosts to migrate, a better option would be to leave the hosts definition with the "normal" groups and use the proposed parameter `--all-hosts` combined with `--limit` to migrate to different groups when necessary.

It can be useful when running ad-hoc without inventory:
`ansible-playbooks --all-hosts -i rogue_webserver.hostname, configure_webservers.yml`

Or when running against dynamic groups or just created (ungrouped) servers:
`ansible-playbooks --all-hosts -i ec2.py --limit "tag_Type_webservers" configure_webservers.yml`

### DIFF
--- a/bin/ansible-playbook
+++ b/bin/ansible-playbook
@@ -102,6 +102,8 @@ def main(args):
         help="run handlers even if a task fails")
     parser.add_option('--flush-cache', dest='flush_cache', action='store_true',
         help="clear the fact cache")
+    parser.add_option('--all-hosts', dest='all_hosts', action='store_true',
+        help="override playbook hosts to use all")
 
     options, args = parser.parse_args(args)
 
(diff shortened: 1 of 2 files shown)


bcoca:
not sure we want this as an option, if you really need to you can write you playbooks like this

hosts: "{{target|default('normalgroup'}}"

and execute withe -e 'target=all' when needed.

adenot:
Thanks Brian, the solution you propose hopefully can cover most of the cases - including the ones I mention. 
In my case I will have to use the "--all-hosts" parameter since I'm running ansible playbooks written by a third-party so I don't have access to modify customer's playbooks.

bcoca:
So I understand your specific case, but this is not a feature we want for ansible in general, we really wan't plays to be the authoritative source. So I'm going to close this PR without merging.

PR #10909, opened 2015-05-03
add the dependency markupsafe

larsborn:
this dependency seems to be missing

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ setup(name='ansible',
       author_email='michael@ansible.com',
       url='http://ansible.com/',
       license='GPLv3',
-      install_requires=['paramiko', 'jinja2', "PyYAML", 'setuptools', 'pycrypto >= 2.6'],
+      install_requires=['markupsafe', 'paramiko', 'jinja2', "PyYAML", 'setuptools', 'pycrypto >= 2.6'],
       package_dir={ '': 'lib' },
       packages=find_packages('lib'),
       package_data={


sivel:
I believe markupsafe is a dependency of jinja2, we should not need to be explicit about it.

https://github.com/mitsuhiko/jinja2/blob/master/setup.py#L75

larsborn:
hm that's odd. I just installed ansible on a fresh raspbian machine with the following commands

```
aptitude install python-pip
pip install ansible
```

and got the error, that markupsafe is missing. As I understand pip, this should not happen, right? Even more odd:

```
pip install markupsafe
```

solved the problem.

abadger:
Did jinja2 come from raspbian package or did the pip install ansible install jinja2?

larsborn:
Since 

```
aptitude search jinja2
```

does not show any installed jinja2 package, I assume that

```
pip install ansible
```

installed jinja2. How can I be of more help?

amenonsen:
I guess this PR can be closed now, since there has been no further correspondence.

bcoca:
This then seemed like a pip issue of not pulling down the dependencies from jinja2. Testing on debian/ubuntu downloads markupsafe when you install jinja and again when you install ansible and the dependency resolution pulls in jinja2 and it's dependencies, including markupsafe. 

So this is either a problem with pip and raspian or the raspian version of pip not downloading the whole dependency tree.

I tried getting raspian on qemu but most of the links on the guides are dead and i'm not going to spend the time to cross compile a custom arm kernel. I suggest opening an issue with the raspian packagers about pip not pulling in dependencies correctly.

As for this issue i'm closing it for now, feel free to open a new one and point to this one if there is something that points to Ansible not stating it's dependencies correctly, but I do not think that is the case right now.

larsborn:
sounds reasonable :)

PR #10918, opened 2015-05-05
Ansible inventory

tyll:
Support for ANSIBLE_HOSTS is broken due to commit
c73254543a9fc66bf2a22f978c6e979ae361221c, therefore do not mention it in
the man pages.


tyll:
Commit c73254543a9fc66bf2a22f978c6e979ae361221c, breaks it, because the inventory section is still used even when ANSIBLE_HOSTS is set. I wrote the wrong commit id initially.

bcoca:
so i have a different fix for rst/intro_installation.rst which also mentions the deprecated ANSIBLE_HOSTS and that it happened in 1.9, remove that and i'll merge the rest for the manpages which should only care about 'current version'.

tyll:
here is the update

Issue #10878, opened 2015-04-29
winrm_https_listener_parent_path command causes failure


ngourley:
The running of this PS command:
https://github.com/ansible/ansible-modules-core/blob/devel/windows/setup.ps1#L72
causes error on JSON conversion:
https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/powershell.ps1#L68

Its a bit confusing since it doesn't appear that $winrm_https_listener_parent_path is used for anything but getting a substring and in fact, the error persists if you run the command and do not assign the output to a variable leaving $winrm_https_listener_parent_path undefined.

```
failed: [remote_host] => {"failed": true, "parsed": false}
ConvertTo-Json : An error occurred creating the configuration section handler
for system.web/httpRuntime: The type initializer for
'System.Web.Configuration.HttpRuntimeSection' threw an exception.
At C:\Temp\ansible-tmp-1430330194.46-227165649139089\setup.ps1:86 char:17
+     echo $obj | ConvertTo-Json -Depth 99
+                 ~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [ConvertTo-Json], Configuratio
   nErrorsException
    + FullyQualifiedErrorId : System.Configuration.ConfigurationErrorsExceptio
   n,Microsoft.PowerShell.Commands.ConvertToJsonCommand
```

cchurch:
Could you provide more information on your version of Windows and PowerShell?  Are you connecting to Windows over HTTPS or just HTTP?

ngourley:
Remote host is running Windows 2008 Server R2 Enterprise, SP1, Powershell 3 & Ansible prep were run on this host. Connection is using local account, and connection is over HTTPS. If I comment out that Powershell command the fact collection is successful:

```
[...]

Issue #10825, opened 2015-04-23
Process is terminated due to StackOverflowException.


basictheprogram:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 2.0.0 (devel 716a87ff36) last updated 2015/04/23 12:58:07 (GMT -500)
  lib/ansible/modules/core: (detached HEAD 022e5274e6) last updated 2015/01/11 15:27:34 (GMT -500)
  lib/ansible/modules/extras: (detached HEAD a32869d492) last updated 2015/01/11 15:27:43 (GMT -500)
  v2/ansible/modules/core: (detached HEAD 27d5e5124f) last updated 2015/01/11 15:27:48 (GMT -500)
  v2/ansible/modules/extras: (detached HEAD 8a4f07eecd) last updated 2015/01/11 15:27:53 (GMT -500)
  configured module search path = None
##### Environment:

Control: OS X 10.10.3
Client: Windows 7 Pro 64-bit
##### Summary:

Attempting to make changes to Local Administrator account and I get a StackOverflowException.
##### Steps To Reproduce:

```
- name: Clean up Local Administrator
  win_user: name=Administrator description=Local fullname=Local password_never_expires=yes
```

```
$ ansible-playbook --ask-pass -i windows-workstations windows-workstations.yml
TASK: [windows-workstations | Clean up Local Administrator] *******************
failed: [windows-test.local] => {"failed": true, "parsed": false}
```

Process is terminated due to StackOverflowException.

I did execute the play with -vvvv but I'm not sure what to post here. It's pretty long.
##### Expected Results:

Local Administrator would have the description and fullname set to Local and the password set to never expire.
##### Actual Results:

Process is terminated due to StackOverflowException.

bcoca:
@cchurch

basictheprogram:
I believe this is the relevant parts of the -vvvv. Let me know if you need more.

<windows-test.local> WINRM PUT /var/folders/d_/bm7rvz154jb_2djkqybb503h0000gn/T/tmpKfvknP to C:\Users\Administrator\AppData\Local\Temp\ansible-tmp-1429830677.31-160705540905978\arguments (offset=0 size=103)
[...]

PR #10926, opened 2015-05-06
Documentation update; Add advice about using unpatched Windows Management Framework 3.0.

jhawkesworth:
Hopefully this should reduce bug reports like #10878 and #10825


cchurch:
Merged, thanks!  Will be included whenever the docs site is updated again.

PR #10935, opened 2015-05-06
Fix for OpenBSD

stemid:
It's not a solution but it's a workaround for us who are trying to use ansible on OpenBSD. Nothing was being done about [this issue](https://github.com/ansible/ansible-modules-core/issues/278) so this workaround helped me at least.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,15 +15,20 @@ Deprecated Modules (new ones in parens):
 New Modules:
   * find
   * ec2_ami_find
+  * ec2_win_password
+  * circonus_annotation
   * consul
   * consul_acl
   * consul_kv
   * consul_session
   * cloudtrail
+  * cloudstack: cs_account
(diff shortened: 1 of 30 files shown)


rvalente:
Nice!

bcoca:
you have merge commit in your PR, you might want to fix that and push again

stemid:
@bcoca How do I do that?

bcoca:
use git rebase and/or cherry-pick to get a clean branch, then just force push to the shared branch the PR is based on

stemid:
@bcoca Ok I tried following [https://robots.thoughtbot.com/keeping-a-github-fork-updated](this guide) and have now pushed it to my fork repo.

bcoca:
so it is more messed up now than before, I just cherry picked your commit as it seems less work at this point. Closing ticket, it won't appear as 'merged' but the code has been merged.

PR #10612, opened 2015-04-03
Add ability to specify using ssh_args in synchronize

sivel:
This PR, coupled with a soon to be linked ticket for the actual module itself, adds 2 new options:
1. `use_ssh_args`
2. `ssh_args`

`use_ssh_args`, which is a boolean (defaulted to `False` to keep current behavior), is what is exposed currently via documentation, but not in the `argument_spec`.

`ssh_args`, a string, is not exposed via documentation, and does appear in `argument_spec`

`use_ssh_args` is popped from the options, and `ssh_args` is added.

PR #10943, opened 2015-05-07
Add ability to specify using ssh_args in synchronize for v2

sivel:
This PR brings the changes from #10612 into the v2_final branch.

In addition to the changes needed to get `use_ssh_args` to work, there is an additional commit to just get the synchronize action plugin to work at all.

Of important note.  Automatically delegating to localhost is _not_ fixed in this PR.  I did not have the time to get that functionality working.


jimi-c:
Merged, thanks!

PR #10946, opened 2015-05-07
Update DigitalOcean dynamic inventory to API v2

alezkv:
Update to API version 2.
In API v 2 to get inventory is done in one request, so there is no need to cache any more.

### DIFF
--- a/plugins/inventory/digital_ocean.ini
+++ b/plugins/inventory/digital_ocean.ini
@@ -3,12 +3,11 @@
 
 [digital_ocean]
 
-# The module needs your DigitalOcean Client ID and API Key.
-# These may also be specified on the command line via --client-id and --api-key
-# or via the environment variables DO_CLIENT_ID and DO_API_KEY
+# The module needs your DigitalOcean API Token.
+# It may also be specified on the command line via --api-token
+# or via the environment variables DO_API_TOKEN or DO_API_KEY
 #
-#client_id = abcdefg123456
-#api_key = 123456abcdefg
+#api_token = 123456abcdefg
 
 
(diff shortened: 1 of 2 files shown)


bcoca:
i'm not sure that removing the cache is a good thing, specially when people invoke playbooks frequently.

oryband:
digitalocean's new team account support work only with v2. Furthermore, using v1 doesn't return any error - just an OK with empty droplets list, which is very misleading.

Please update to v2, this is a very big deal.

oryband:
@ofir-petrushka @avivl @fuzzyami

bcoca:
im tempted to make this a do_v2 inventory plugin, or is there no case where someone would need to use the v1 version?

oryband:
@bcoca isn't v2 supposed to do what v1 does and more? we should just move to use v2 imo

bcoca:
I do not know, that is why I'm asking, not everyone keeps backward compatibility between versions, even when they try it is not normally perfect.

oryband:
@bcoca well, i guess the answer is here:
- https://developers.digitalocean.com/
- https://www.digitalocean.com/company/blog/apiv2-officially-leaves-beta/

alezkv:
more precise answer is here:
https://developers.digitalocean.com/documentation/changelog/api-v1/sunsetting-api-v1/

> We’re not going to pull the rug out from under you. In order to give everyone time to port their tools, version 1 will not be turned off until November 1st 2015.

oryband:
@izhukov even though they wrote this, v1 completely doesn't work **now** for users who are part of a team.

bcoca:
OK, seems that this cannot break it more than it is already but will enable people to use it.

PR #10949, opened 2015-05-08
Add compress support for paramiko

mightydok:
Hi all.
Just propose patch for compression support for paramiko transport connection. I think this change will save some network traffic and reduce module execution time because lower data size will be transfered via network.


lib/ansible/constants.py
```diff
@@ -184,6 +184,7 @@ def shell_expand_path(path):
 ANSIBLE_SSH_CONTROL_PATH       = get_config(p, 'ssh_connection', 'control_path', 'ANSIBLE_SSH_CONTROL_PATH', "%(directory)s/ansible-ssh-%%h-%%p-%%r")
 ANSIBLE_SSH_PIPELINING         = get_config(p, 'ssh_connection', 'pipelining', 'ANSIBLE_SSH_PIPELINING', False, boolean=True)
 PARAMIKO_RECORD_HOST_KEYS      = get_config(p, 'paramiko_connection', 'record_host_keys', 'ANSIBLE_PARAMIKO_RECORD_HOST_KEYS', True, boolean=True)
+PARAMIKO_COMPRESS	       = get_config(p, 'paramiko_connection', 'compress', 'ANSIBLE_PARAMIKO_COMPRESS', False, boolean=True)
```
resmo: indention not aligned :)

mightydok: Ok, fixed.

amenonsen:
So this patch adds a boolean `compress` setting in the `paramiko_connection` section, and passes its value to the `ssh.connect` call. It should be easy enough to rebase to apply to current devel (it also needs a documentation update), but first, can someone from the core team comment on whether it's wanted or not? I have no particular opinion myself, since I don't use paramiko at all.

Also, the comment about interactive sessions seems misleading in context, since ansible doesn't have 'em.

Also², the commits should be squashed into one before merging.

mightydok:
Close this pull request too.

PR #10950, opened 2015-05-08
Add compression support for ssh, scp, sftp connection.

mightydok:
Hi all.
Pull request  to add compression support for ssh, scp, sftp connection, i think it will speed up connection and file transfer.


mightydok:
Just dont fing "-C" switch for ssh command.

PR #10951, opened 2015-05-08
make compression support for ssh, scp, sftp connection configurable

mightydok:
I notice that compression used only for command execution via ssh, but not for scp or sftp file transfers. Patch will set compression=yes by default for ssh, scp and sftp. User can switch off compression via cfg file.

### DIFF
--- a/examples/ansible.cfg
+++ b/examples/ansible.cfg
@@ -208,6 +208,10 @@ fact_caching = memory
 # (default is sftp)
 #scp_if_ssh = True
 
+# if True, make ssh use compression for connection
+#compression=True
+
+
 [accelerate]
 accelerate_port = 5099
 accelerate_timeout = 30
--- a/examples/ansible.cfg
+++ b/examples/ansible.cfg
@@ -209,8 +209,7 @@ fact_caching = memory
 #scp_if_ssh = True
 
 # if True, make ssh use compression for connection
-#compression=True
-
+#compression=False
 
 [accelerate]
 accelerate_port = 5099
--- a/examples/ansible.cfg
+++ b/examples/ansible.cfg
@@ -208,7 +208,7 @@ fact_caching = memory
 # (default is sftp)
 #scp_if_ssh = True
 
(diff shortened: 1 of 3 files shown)


bcoca:
any reason we would not want compression?

mightydok:
@bcoca compression not good for binary or already compressed data transfer and interactive sessions, so ansible transfers mostly text data and i think there will be some speed up for large modules and playbooks, but iam not 100% sure. If anyone can measure playbook run timing, it will be great.

bcoca:
Ansible is a batch system so interactive is not an issue. 

Aside from unarchive & copy (and possible future support for binary modules) Ansible does not normally transfer binaries, which (aside from unarchive) are normally compressible also. Admittedly  the compression rate tends to be lower than text files.

There are a couple of callback plugins that measure playbook times and I have seen this make a difference, specially when scaling to many tasks and/or hosts

amenonsen:
OK, so this is the logical companion to #10949. It too is in need of (a) a rebase to apply to devel, (b) squashing into one commit, and (c) improved comments. Also note that the parameter in `ssh_connection` is named `compression` while the one in `paramiko_connection` is named `compress`. That won't do; both should be named `compression` IMO. And perhaps they should be two commits in one PR.

But before that, I agree with @bcoca that it's a bit overkill to make this configurable. We should just enable compression for both ssh and paramiko and be done with it.

talcloudshare:
@bcoca Please note that Windows OpenSSH implementation currently does not support compression, therefore it would be useful to be able to disable it

See comments in http://stackoverflow.com/questions/33345833/ansible-failed-to-connect-to-windows-node-via-ssh/33347112

mightydok:
@amenonsen @talcloudshare thnx for comments, tomorrow i`ll rebase that pull request and pull request for paramiko.

mightydok:
@bcoca i notice that ssh connection plugin was totally rewriten, and compression enabled by default for all ssh connections, scp and sftp too. Thanks, whait for ansible 2 =).

PR #10966, opened 2015-05-09
pybook: Python syntax to write playbooks

muravjov:
Now Ansible suggests to write playbooks in YAML and JSON.

This patch adds a possibility to write playbooks in Python language for greater flexibility. A Python playbook use a simple API to generate YAML like structures, like so:

In YAML:

```

---

- hosts: web
  vars:
    build_cbo: false

  tasks:
  - include: web-tasks.yml
```

In Python:

```
#!python

with mapping:
    append("hosts", "web")
    with mapping("vars"):
        append("build_cbo", False)

    with sequence("tasks"):
        with mapping:
            append("include", "web-tasks.py")
```

Python API consists of `<with mapping>` for starting dictionaries, `<with sequence>` for starting lists, `<append>` for appending whole Python objects and `<append_yaml>`/`<populate_yaml>` for mixing Python and YAML notation.

This patch also includes tests in test/unit/TestPybook.py .

### DIFF
--- a/lib/ansible/utils/__init__.py
+++ b/lib/ansible/utils/__init__.py
@@ -35,7 +35,9 @@ from ansible.module_utils.splitter import split_args, unquote
 from ansible.module_utils.basic import heuristic_log_sanitize
 from ansible.utils.unicode import to_bytes, to_unicode
 import ansible.constants as C
+from . import pybook
 import ast
+import pprint
 import time
 import StringIO
 import stat
@@ -774,10 +776,17 @@ def parse_yaml_from_file(path, vault_password=None):
         data = vault.decrypt(data)
         show_content = False
 
-    try:
-        return parse_yaml(data, path_hint=path)
-    except yaml.YAMLError, exc:
-        process_yaml_error(exc, data, path, show_content)
+    if re.match("#!.*python", data):
+        result = pybook.run_pybook(path)
+    else:
+        try:
+            result = parse_yaml(data, path_hint=path)
+        except yaml.YAMLError, exc:
+            process_yaml_error(exc, data, path, show_content)
+            
+    if VERBOSITY >= 3:
+        display("""Structure of file "%s":\n%s\n""" % (path, pprint.pformat(result)), color='yellow')
+    return result
 
 def parse_kv(args):
     ''' convert a string of key/value items to a dict '''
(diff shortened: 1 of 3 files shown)


bcoca:
We only suggest writing playbooks in YAML and don't explicitly support JSON, it just happens to be a subset of YAML. We do not want multiple ways to write a play, specially not as a programming language. 

The aim is to keep ansible auditable and simple, having more than one way to write playbooks takes us in the opposite direction. Most of ansible's effort is hiding code behind the plugins, this again is counter productive. 

We are going to pass on this feature.

muravjov:
Well, I proposed such an answer. And I understand your intentions, ok.

The other way, keeping playbooks auditable and simple is not that simple. For example, if you have to write many tasks with "become" like

```
- pip: virtualenv=/opt/catbo-web/env 
  become: yes
  become_user: catbo
```

then you have a lot of repetitive code, fail. 
And I suppose to use this experimental possibilty (Python playbooks) to create a better DSL above Ansible to avoid such problems.

Earlier I worked with Salt and used to think to make a better deploy system with it. But now I found Ansible, which is much more better and simpler! Thanks for your work.

bcoca:
currently you can set become options per  play, role, task include, task, soon in v2 you also get task blocks which I think will bring most of what you wanted to the standard syntax.

muravjov:
Thanks for info (task blocks).

But don't you think that you are inventing now one more programming language using YAML notation, do you? By the way, with Python syntax I can make task block myself on the fly :)

bcoca:
We try very hard to not be a programming language, while implementing enough elements of it to make it usable (fine line I admit).  The original author also wrote Func (https://fedorahosted.org/func/) which is very much what you describe, but went in a different direction with Ansible precisely because of that experience. 

I for one think it makes a much better tool (I've written plenty of CM in perl/bash/ksh/sh/python) and do sometimes miss the programming aspect. but I think the tradeoff for simplicity is a big win here.

The YAML playbooks are intended to be descriptive and not imperative, i think this is the major difference with most scripting languages. (yes i'm ignoring all lisp like things, on purpose)

muravjov:
Well, my proposed code generate the very same YAML structures, too (Python structures to be precise). 

Anyway, I'll try to keep my patch in actual state. Maybe it'll be worth it.

Issue #10937, opened 2015-05-07
ansible_distribution is wrongly reported as 'RedHat' under CentOS 7.1


vmalloc:

mightydok:
Cant confirm for last devel and stable branch:
[root@localhost ansible]# cat /etc/redhat-release
CentOS Linux release 7.1.1503 (Core)

[root@localhost ansible]# ansible -ihosts -u ans -k -m setup all -a "filter=ansible_distribution*"
SSH password:
127.0.0.1 | success >> {
    "ansible_facts": {
        "ansible_distribution": "CentOS",
        "ansible_distribution_major_version": "7",
        "ansible_distribution_release": "Core",
        "ansible_distribution_version": "7.1.1503"
    },
    "changed": false
}

For fresh vagrant box with centos 7.1.

@vmalloc can you provide more details? Ansible version, OS version.
Can u try to check problem with latest ansible version from devel branch?

Can u provide same output for setup module?

vmalloc:
Yes. My host reports this under `/etc/redhat-release`:

```
Derived from Red Hat Enterprise Linux 7.1 (Source)
```

And thus the release seems to be identified as RedHat:

```
host-003 | success >> {
    "ansible_facts": {
        "ansible_distribution": "RedHat",
        "ansible_distribution_major_version": "7",
        "ansible_distribution_release": "Core",
        "ansible_distribution_version": "7.1.1503"
    },
    "changed": false
}
```

bcoca:
so you changed your /etc/redhat-release to show redhat?

mightydok:
According to this - http://lists.centos.org/pipermail/centos-announce/2015-April/021010.html @vmalloc you just need to run yum update and problem will be fixed.
@bcoca i will try to update ansible code to prefer /etc/os-release if present.

vmalloc:
@mightydok will try it. Thanks!

vmalloc:
Seems to work for me. Thanks!

PR #10968, opened 2015-05-09
Fix ansible/ansible#10937

mightydok:
As i see python.dist() find correct distribution_name, but next code rewrite it to RedHat. I think we can add Centos to exception like Fedora.

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -293,7 +293,7 @@ class Facts(object):
             for (path, name) in Facts.OSDIST_LIST:
                 if os.path.exists(path):
                     if os.path.getsize(path) > 0:
-                        if self.facts['distribution'] in ('Fedora', ):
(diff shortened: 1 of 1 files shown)


amenonsen:
If I understand correctly, this patch says "If platform.dist() says we're running centos, don't decide differently  based on `/etc/redhat-release` (or any of the other files in `OSDIST_LIST`)". That seems fine, given the precedent for Fedora. I'm a bit fuzzy on the details of the original issue #10937, but this should be safe to merge.

bcoca:
I think this is not needed as it was just a temporary snafu on the CentOS side and it was already fixed.

dmick:
As far as I've seen, platform is usually correct in what it divines anyway; why is there even that list of /etc/ files to go read?  (I noticed the problem in #10937, and found the threads here, but my question is, why does the code that caused #10937 even exist?  Why not rely on platform?)

mightydok:
@bcoca agreed

PR #10969, opened 2015-05-10
Add filters: getgrnam and getpwnam

rrauenza:
I thought these might be useful ... and the docs encouraged submissions of useful filters.

In my particular use case, I need to create a pure-ftpd passwd file but I need the uids of the users.

(Admittedly, I didn't check if the user action returned the uid as a result...)

### DIFF
--- a/docsite/rst/playbooks_filters.rst
+++ b/docsite/rst/playbooks_filters.rst
@@ -294,6 +294,22 @@ Hash types available depend on the master system running ansible,
 'hash' depends on hashlib password_hash depends on crypt.
 
 
+Unix Group and Passwd filters
+-----------------------------
+.. versionadded:: 1.?
+
+Get the password database entry for a username::
+
+    {{ 'root'|getpwnam('root').pw_uid }}
+
+Get the group database entry for a group::
+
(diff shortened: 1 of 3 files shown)


rrauenza:
Ah... and now I've learned I can add -vvvv's and see the actual return values of user: and group: that do return the gid and uid.   That doesn't exactly help all cases because one might be looping over usernames without corresponding rules to register results from.

bcoca:
actually, filters change data, this seems more appropriate as a lookup as that is what you are really doing

rrauenza:
Hmm -- I hadn't thought of it that way.  Oh -- especially important -- where is this filter executed -- is it executed on the target system?  Because if it isn't, then this filter is not going to do what I intended!

bcoca:
nope, filters and lookups ONLY run on the master/controller, you can get most of this info from the target using the getent module.

rrauenza:
Thanks!  Then I think I will close this for many reasons -- it doesn't look it up on the target system, and a lookup would be more appropriate if it did.  And getent does it, anyway.   You could also use the registered value from user or group modules, too.

PR #10973, opened 2015-05-10
Remove trailing newline from default meta template

retr0h:
`ansible-galaxy init` does not add trailing newlines to any of the
other files it generates.  Correcting the meta template to behave.


bcoca:
so the code changed a lot in v2 and this would not apply, I moved out the template into a separate file for easier updates and applied your changes in commit 4b9a79d42bf280a742b7f759ac38a0b326ebd941

PR #10974, opened 2015-05-11
Workaround for gce.py erroring on VMs with no public_ips

jimconner:
gce.py previously assumed that all VMs have a public_ip assigned
The script was erroring when attempting to return inst.public_ips[0] from an empty list.

This workaround returns a 'null' if the list is empty.


amenonsen:
@jimconner would you mind retargeting this patch against contrib/inventory/gce.py and refreshing the PR with git push -f?

jimconner:
Hey @bcoca - I've just looked at the version of gce.py in contrib/inventory/gce.py, and that would not require the same fix as it already checks the length of inst.public_ips and uses the private IP if there are no public ones - I'll close this pull request.

PR #10978, opened 2015-05-11
Test hosts

yhekma:
Added an optional --check-first option when running ansible. When this is set and the command is run using elevevated or different privileges, the command is first run on 1 host. If that host fails, ansible considers that the only host and skips the rest.
The reason I did this was because we have a pretty rigorous security policy, so when we try to execute a command as root on 100+ servers, the account is immediately locked because of all the failed login attempts.

### DIFF
--- a/bin/ansible
+++ b/bin/ansible
@@ -161,6 +161,7 @@ class Cli(object):
             become_pass=becomepass,
             become_user=options.become_user,
             extra_vars=extra_vars,
+            check_first=options.check_first,
         )
 
         if options.seconds:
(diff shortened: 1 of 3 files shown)


sivel:
I'm really not sure how useful this is for other people.  It feels very specific to your need and I don't see anyone else really benefitting from this functionality.

yhekma:
That could very well be. In that case, please consider it as non-requested:)
Fact is, I ran into this problem on 3 seperate occasions/environments now (I am a consultant). Then again, I alone am kind of a small sample group:)

bcoca:
# Question About This One

Hi!

- This is a very specific solution to a rare situation, we have another feature planned that we feel better suits ansible and more cases (making serial support scaling lists: serial: [1, 5, 10, 50]).

PR #10984, opened 2015-05-11
added alternative syntax for tag usage h/t @ingydotcom

goozbach:
A different syntax for more compressed multiple tag entries.


amenonsen:
I don't know what h/t means, but this isn't specific to tags. Any YAML list can be written in either form, and there are numerous examples of the [] form in the documentation. I don't think it's useful to add this example.

amenonsen:

bcoca:
closed as per above

Issue #10073, opened 2015-01-23
jinja: stuck in endless loop, 100% CPU usage


christophgysin:
My ansible configuration causes ansible to lock up, using 100%CPU.

I managed to reduce to problem to the following:

```
$ echo localhost > ansible.hosts
$ cat > test.yml << EOF

---
- hosts: localhost
  vars:
    test:
      foo: "{{ lookup('file', 'files-test/baz') }}"
      bar: "{{ lookup('file', 'files-test/baz2') }}"
EOF
$ mkdir files-test
$ echo foobar > files-test/baz
$ ansible-playbook -i ansible.hosts test.yml
^C
^C
^Z
[1]+  Stopped                 ansible-playbook -i ansible.hosts test.yml -v
$ kill %1
$ 
[1]+  Terminated              ansible-playbook -i ansible.hosts test.yml -v
```

It seems that the string 'files-test/baz' is parsed by jinja, which tries to resolve "test" as a variable, ending in an endless loop. If I rename the variable or the directory, it works as expected.

So while there is a simple workaround, this seems like a bug and took me a while to isolate. I'm not sure though if the problem is in jinja itself or in the way ansible is using it.

janLo:
This issue is not limited to the file lookup. Its within [ansible.utils.listify_lookup_plugin_terms](https://github.com/ansible/ansible/blob/devel/lib/ansible/utils/__init__.py#L1450) which gives the string to the template engine to get terms defined as

```
    with_items: some_list
```

This evaluates the term as template expression. If one of the expression parts is the key for the current yaml dict part it evaluate the lookup endless recursive.

V2 seems not to have this function. I've not tested if the bug is present there.

janLo:
A simple but ugly hack is to double-quote the terms given to 

```
---
- hosts: localhost
  vars:
    test:
      foo: "{{ lookup('file', '\"files-test/baz\"') }}"
      bar: "{{ lookup('file', '\"files-test/baz2\"') }}"
```

A better solution is to do the template evaluation of the lookup term only when its called within a with_files context. This information can be provided with the `inject` variabe.

janLo:
This PR fixes the Issue for me.

janLo:
Its simple to reproduce the issue without having a endless recursion. Given the playbook:

```
- hosts: localhost
  vars:                                      
    foo: peng                                
    bar: "{{ lookup( 'file', 'foo' ) }}"
```

The upstream ansible 1.9 looks:

```
[...]

PR #10991, opened 2015-05-12
Evaluate lookup terms only in with_* contexts

janLo:
As mentioned in Issue #10073 there is a problem when evaluation lookup
terms in inline {{ lookup( ... ) }} statements. The problem is, that
the template engine dies not know that the term was a literal even if
it was quoted, because the "outer" tempalte engine have already
stripped the quotes when the templating in listify_lookup_plugin_terms
see the term.

The term will then be evaluated as template variables and if a part
of the term evaluates to the "parent" variable of the current scope it
will cause an endless recursion.

This patch fixes the issue by introducing a new inject property that
tells in whick context we are currently. So listify_lookup_plugin_terms
can skip term evaluation if we are not within a "with_*" context.

### DIFF
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -712,7 +712,9 @@ class Runner(object):
             try:
                 items_terms = self.module_vars.get('items_lookup_terms', '')
                 items_terms = template.template(basedir, items_terms, inject)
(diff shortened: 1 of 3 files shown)


abadger:
I believe this is fixed in v2.  Still evaluating whether this sohuld be applied to 1.9.x as well.  May be too big of a change.

jimi-c:
Per the comment on the other issue, since this has been applied to the devel branch we're going to go ahead and close this now.

Thanks!

PR #10994, opened 2015-05-12
Add integration test to verify #10073

janLo:
In issue #10073 a misbehaviour in literal handling for inline lookup
arguments that can cause unexpected behaviur was reported. This
integration testcase reproduce the problem.

After applying pull request #10991 the issue is fixed and the test
passes.

Signed-off-by: Jan Losinski losinski@wh2.tu-dresden.de

### DIFF
--- a/test/integration/roles/test_lookups/tasks/main.yml
+++ b/test/integration/roles/test_lookups/tasks/main.yml
@@ -129,3 +129,26 @@
   debug: msg={{item}}
   with_items: things2
 
+
+# BUG #10073 nested template handling
+
+- name: set variable that clashes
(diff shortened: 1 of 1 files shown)


abadger:
Okay, I've cherry-picked this into the v2_final branch (since it seems fixed in v2).  Haven't pushed it to the v1/devel branch as we aren't sure if we're going to take the fix in 1.9.x yet.

(Sorry about cherry-picking... I couldn't get git rebase --onto v2_final devel to work.)

jimi-c:
Since this was cherry-picked, we'll go ahead and close this now.

Thanks!

PR #10999, opened 2015-05-13
$SubjectName variable unused; clean up

mrcrilly:
Having used this script several times today, I came to notice the $SubjectName variable, being passed in via the CLI, is essentially ignored when generating the SSL certificates, rendering it useless. I believe it's a good idea to have it in place, so I've updated the script to reflect this.

I also cleaned up some random new lines throughout the file, and expanded on a comment.

It might be worth going a step further and commenting the file fully, as most people reviewing this file won't be familiar with PowerShell (like I wasn't until a few days ago). It could be helpful.

### DIFF
--- a/examples/scripts/ConfigureRemotingForAnsible.ps1
+++ b/examples/scripts/ConfigureRemotingForAnsible.ps1
@@ -1,4 +1,4 @@
-﻿# Configure a Windows host for remote management with Ansible
+# Configure a Windows host for remote management with Ansible
 # -----------------------------------------------------------
 #
 # This script checks the current WinRM/PSRemoting configuration and makes the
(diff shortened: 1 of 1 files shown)


amenonsen:
@mrcrilly cleaning up "random new lines" is usually inviting trouble (in the form of merge conflicts), but in this case I guess you're the only one changing this particular file, so you got lucky. But could I trouble you to submit the actual `$SubjectName` change as a separate commit for ease of review? Better comments are a good idea, but should be in a separate commit (along with any required cleanups).

Also note typo: listerner → listener

nitzmahone:
Merged via e9fe5f201f5830167d2ca6168c08822c930efb41

Issue #10997, opened 2015-05-12
paramiko establishes SSH connection for action plugins


sivel:
##### Issue Type:

Bug Report
##### Ansible Version:

`ansible 1.9.1 (detached HEAD b47d1d7e69) last updated 2015/05/12 16:27:04 (GMT -500)`
##### Ansible Configuration:

N/A
##### Environment:

Control: Mac OS X
Managed: N/A
##### Summary:

When using paramiko as the connection type, ansible will attempt to establish an SSH connection for an action plugin even when not necessary, such as in the example of using `set_fact`
##### Steps To Reproduce:

``` yaml

---
- hosts: all
  gather_facts: false
  tasks:
    - set_fact: foo=bar
    - set_fact: baz=qux
```

```
ansible-playbook -i some.nonexistent.host, -c paramiko -vvvv set_fact.yml
```
##### Expected Results:

```
PLAY [all] ********************************************************************

TASK: [set_fact foo=bar] ******************************************************
<some.nonexistent.host> ESTABLISH CONNECTION FOR USER: sivel
ok: [some.nonexistent.host] => {"ansible_facts": {"foo": "bar"}}

TASK: [set_fact baz=qux] ******************************************************
<some.nonexistent.host> ESTABLISH CONNECTION FOR USER: sivel
ok: [some.nonexistent.host] => {"ansible_facts": {"baz": "qux"}}

PLAY RECAP ********************************************************************
some.nonexistent.host            : ok=2    changed=0    unreachable=0    failed=0
```
##### Actual Results:

There is also a traceback, but the cause is due to a SSH connection failure.  There is no such exception in `v2_final` however paramiko still attempts to establish an SSH connection

``````
PLAY [all] ********************************************************************

TASK: [set_fact foo=bar] ******************************************************
<some.nonexistent.host> ESTABLISH CONNECTION FOR USER: sivel on PORT 22 TO some.nonexistent.host
fatal: [some.nonexistent.host] => Traceback (most recent call last):
  File "/Users/sivel/python_venvs/ansibledev/ansible/lib/ansible/runner/__init__.py", line 582, in _executor
    exec_rc = self._executor_internal(host, new_stdin)
  File "/Users/sivel/python_venvs/ansibledev/ansible/lib/ansible/runner/__init__.py", line 785, in _executor_internal
    return self._executor_internal_inner(host, self.module_name, self.module_args, inject, port, complex_args=complex_args)
[...]

PR #11002, opened 2015-05-13
Use a decorator to perform just in time connection

sivel:
This PR adds a new `ensure_connect` decorator for connection plugins.  It's function is to decorate a method requiring a connection to be open.  The replaces the explicit call to `_connect`.

WHY:  Calling `_connect` explicitly as it is, causes the connection plugin to connect to a remote host, sometimes needlessly such as when running an action plugin that does not need a connection, such as `set_fact`.

This change will ensure that a connection is not forcefully established when not needed, but still establish it just in time for the operation.

The following connection plugins have been updated due to them being ready for v2:
- ssh
- paramiko
- winrm

Additional plugins will need to have this decorator applied to `exec_command`, `put_file` and `fetch_file`.

I've tested a handful of scenarios and did not encounter any issues.

Addresses #10997

### DIFF
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -210,7 +210,6 @@ class TaskExecutor:
         # get the connection and the handler for this execution
         self._connection = self._get_connection(variables)
         self._connection.set_host_overrides(host=self._host)
-        self._connection._connect()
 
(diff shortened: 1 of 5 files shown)


sivel:
Since v2 was just merged, I'll close this and re-open against devel.

PR #11010, opened 2015-05-14
Add judgment to to fix path0 if ANSIBLE_CONFIG is set to a dir

sillydong:
I made a mistake when I set ANSIBLE_CONFIG to a dir contains ansible.cfg and hosts. After digging into the source, I found this variable should direct to ansible.cfg file itself, but I think maybe I'm not the only one who made such mistake, so I pulled this request.


amenonsen:
Seems fair enough, since if ANSIBLE_CONFIG is set to a directory it'll try to append `/ansible.cfg` and see if that exists, otherwise try the other options as it would before.

PR #11013, opened 2015-05-14
ssh_config dynamic inventory fails when config item uses identityfile

aaronbauman:
Paramiko SSHConfig returns a list for identifyfile, instead of a string.
This breaks Ansible's inventory variable handling.
Flatten the list items when assigning values.

### DIFF
--- a/plugins/inventory/ssh_config.py
+++ b/plugins/inventory/ssh_config.py
@@ -75,6 +75,8 @@ def print_list():
         tmp_dict = {}
         for ssh_opt, ans_opt in _ssh_to_ansible:
             if ssh_opt in attributes:
+                if type(attributes[ssh_opt]) is list:
+                    attributes[ssh_opt] = attributes[ssh_opt][0]
                 tmp_dict[ans_opt] = attributes[ssh_opt]
         if tmp_dict:
             meta['hostvars'][alias] = tmp_dict


abadger:
You should use isinstance instead of type().

This discards the additional identityfiles... Can we change the code to process all of them instead?  Is there any reason not to do that?

aaronbauman:
Thanks for the feedback.
I was unclear whether ansible could support multiple identify files.
If you don't know off-hand, I'll dig in and see if I can find out.

Will update the pull request as soon as I'm back at my desk.

aaronbauman:
Cross-linking to this issue from google groups, which was raised in 2013 and died:
[https://groups.google.com/forum/#!searchin/ansible-devel/multiple$20private$20keys/ansible-devel/NZm5i3ZPn_I/V5U6lNV3bIUJ]

If there were a reason _not_ to support multiple keys, it would be that [a bunch of files (24+)](https://github.com/ansible/ansible/search?q=private_key_file) need to be re-written and might not be backwards compatible.

I'm willing to put in some work to make this happen, but I don't want to waste my effort.
What's the MO for preserving backwards compatibility?

abadger:
Backwards compat probably trumps any addition of features.  However, the files that you've linked to are probably okay... We more or less would consider them to be internal changes.  Also, the v2 branch has some pretty big changes to these areas.  So private_keys becoming a list might not be problematic here.

amenonsen:
Given that (a) this PR now has merge conflicts, (b) it solves the original problem in an undesirably limited way, (c) the author has said he intends to resubmit a more complete (but significantly different and more extensive) approach, I think this can be closed now.

aaronbauman:
Haven't had time to pick this up again.

I'll open a new issue when I get around to re-rolling.

PR #11018, opened 2015-05-15
 facts: Add ed25519 ssh pubkey 

tyll:
Add missing ed25519 ssh pubkey to facts and refactor code to reduce code duplication.


bcoca:
to keep backwards compatibility, I would add = 'NA' for when keys are None.

tyll:
On Fri, May 15, 2015 at 02:01:44PM -0700, Brian Coca wrote:

> to keep backwards compatibility, I would add = 'NA' for when keys are None.

AFAICS the value "NA" is never exposed in a ssh key fact in the code
this patch changes. It is not clear to me my the local variables were
set to "NA", maybe they were also exposed as a fact before someone else
changed something here.  Currently it is only dead code - unless there
is some magic that looks at the local variables of the function.

bcoca:
@tyll sorry, had not read the whole code, I had assumed that the variables were used since they were assigned ... dead code can go away.

PR #11027, opened 2015-05-18
added module checklist docs

bcoca:


omgjlk:
This is missing my ask, the docs on what data is returned and would be accessible from a register variable.

bcoca:
updated

omgjlk:
:+1:

jhawkesworth:
Would it be possible to add some guidelines for powershell modules too.

Pretty much all of the documentation stuff ought to be common, but here's
my suggestions for powershell-specific modules.

Ansible powershell modules checklist

.ps1 file
start with
#!powershell
then
GPL header
then

# WANT_JSON

# POWERSHELL_COMMON

Make sure you haven't duplicated any code or methods that are already in
ansible/lib/ansible/module_utils/powershell.ps1

arguments:
Try and use state present and state absent like other modules.
Have you checked all your mandatory args are present? and sane
Look at existing modules for examples of argument checking.

Module logic
Favour native powershell and .net ways of doing things over calls to COM
libraries or calls
to native executables which may or may not be present in all versions of
windows.

Returning results
result object
    should allways contain an attribute called changed set to either $true
or $false
create your result object like this:

```
$result = New-Object psobject @{
          changed = $false
          other_result_attribute = $some_value
};
```

If all is well, exit with a
    Exit-Json $result

Ensure anything you return, including errors can be converted to json.
Be aware that because exception messages could contain almost anything.
ConvertTo-Json will fail if it encounters a trailing \ in a string.

If not use Fail-Json to exit.

Have you tested for powershell 3.0 and 4.0 compliance?

jhawkesworth:
Oops, missed the merge notification there.

PR #11028, opened 2015-05-18
added hmc.py and hmc.ini in plugins/inventory

katishna:
This inventory plugin pulls data from an IBM Hardware Management Console (HMC) about the physical frames attached to that HMC and the logical partitions (individual servers) on that frame. It provides a picture of what is accessible via that HMC. This version of the plugin uses ssh based commands to pull the data, which allows it to work on older HMCs. A future rewrite of the code will use the RESTful API, but will remove the backward compatibility that using ssh provides. The frames run AIX, IBMi, and Power Linux logical partitions.

I need this information regularly at work but couldn't find anything currently in Ansible to provide the data. This is my first PR and I've tried to rebase it against the 2.0 devel branch, so if I've done that wrong please let me know and I'll fix it.

### DIFF
--- a/plugins/inventory/hmc.ini
+++ b/plugins/inventory/hmc.ini
@@ -1,0 +1,10 @@
+# Ansible HMC inventory script
+#
+
+[hmc]
+
+# HMC Hostname
+hostname = hmchost.example.com
+
+# Username
+username = hscroot
--- a/plugins/inventory/hmc.py
+++ b/plugins/inventory/hmc.py
@@ -1,0 +1,73 @@
+#!/usr/bin/env python
+
+# This file requests lpar inventory information from an IBM frame via the
+# Hardware Management Console. You are not required to have ssh keys set
+# up but you will need to supply a username on the command line.
+# 
+# usage: hmc.py --host <hmc> --user <username>
+#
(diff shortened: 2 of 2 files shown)


amenonsen:
I don't know anything about HMC, but (a) how is SSH authentication handled? (b) don't you need to handle `--list` and `--host` arguments? For example, here's an example from `contrib/inventory/jail.py`:

```
if len(sys.argv) == 2 and sys.argv[1] == '--list':
    print json.dumps(result)
elif len(sys.argv) == 3 and sys.argv[1] == '--host':
    print json.dumps({'ansible_connection': 'jail'})
else:
    print "Need an argument, either --list or --host <host>"
```

Otherwise it's not really an inventory script. Also, for currentl devel, this needs to live in `contrib/inventory`.

It might also help if you pasted an example of the output from your script. At least it'll help assuage my curiosity.

katishna:
It's all ssh key-based authentication. It does grab the inventory. I have lost access to an HMC/IBM frame setup since i wrote it (new job) so I will get access and then post an example.

Thanks for the info on where it should live. The release went from 1 to 2 after I submitted this and a lot of stuff changed, I think. This is my first PR pretty much ever so I'm learning. :)

PR #11032, opened 2015-05-19
basic: fix ValueError if value of a type='int' is not an int

resmo:
With this fix, we get a friendly error message:

```
failed: [localhost] => {"failed": true}
msg: value of argument start_port is not of type int
```


bcoca:
templted to add 'and we were unable to automatically convert'

resmo:
done

PR #11040, opened 2015-05-20
doc: we need GPLv3 license headers

resmo:
GPLv2 only headers are incompatible with GPLv3


abadger:
I think our preference is GPLv3 or later headers.  Not sure if we do not take any other licenses, though.  In particular, I don't know what our stance would be on GPLv2 or later  (GPL without version is equivalent to any version of the GPL so it's just a small bit looser than GPLv2 or later).

@jimi-c could you clarify this?

resmo:
@abadger set it to v3 only

abadger:
@jimi-c clarified for me in chat that it should be GPLv3 or later.

amenonsen:
@abadger so is the patch OK as is, or does it need to be modified?

bcoca:
@amenonsen patch is fine but there is legalize invovled

PR #11058, opened 2015-05-23
limit extensions for files in group/host_vars dir

### LINKED ISSUE #11017: Text backups (xxx.yml~) in host_vars override variables
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.9.1 from ppa:ansible/ansible
##### Ansible Configuration:

default
##### Environment:

Ubuntu 15.04
##### Summary:

Text backup files ending on tilde as created e.g. by joe or gedit, in `host_vars/<hostname>/` override the actual variable data.
##### Steps To Reproduce:
- Write samle data into `host_vars/<hostname>/test.yml`
- Edit `host_vars/<hostname>/test.yml` using gedit, gedit will create `host_vars/<hostname>/test.yml~`
- Ensure `test.yml` and `test.yml~` differ

Now ansible will override a variable defined in `test.yml` using the data from `test.yml~`.
##### Expected Results:

Only `*.yml` files are recognized by ansible in `host_vars`.
##### Actual Results:

`*.yml~` files are recognized by ansible in `host_vars`.

srgvg:
inventory vars: make loading from a directory obey the same rules as
when checking the base paths, looking at the file name extensions
as defined in CONSTANTS.YAML_FILENAME_EXTENSIONS

Fixes #11017


lib/ansible/utils/__init__.py
```diff
@@ -1617,7 +1617,9 @@ def _load_vars_from_folder(folder_path, results, vault_password=None):
     names.sort()
 
     # do not parse hidden files or dirs, e.g. .svn/
-    paths = [os.path.join(folder_path, name) for name in names if not name.startswith('.')]
+    paths = [os.path.join(folder_path, name) for name in names
+            if not name.startswith('.')
+            and os.path.splitext(name)[1] in C.YAML_FILENAME_EXTENSIONS]
```
bcoca, line 1622: would -1 not be better as the array index, JIC someone names.their.files.yml this way.

srgvg, line 1622: This is `os.path.splitext` which always returns 2 items. So [1] and [-1] would always be the same.

https://docs.python.org/2/library/os.path.html#os.path.splitext
Split the pathname path into a pair (root, ext) such that root + ext == path, and ext is empty or begins with a period and contains at most one period. Leading periods on the basename are ignored; splitext('.cshrc') returns ('.cshrc', '').

```
In [1]: import os

In [2]: os.path.splitext('names.their.files.yml')
Out[2]: ('names.their.files', '.yml')
```

PR #11062, opened 2015-05-25
Factor primitives for F5 modules

Etienne-Carriere:
Factor the standard elements for all F5 Big IP modules : 
- F5 standard arguments 
- F5 login method


srgvg:
cc @mhite

mhite:
Presuming this is preparation work for moving common elements into a shared library now that Ansible supports it?

Etienne-Carriere:
My only small goal was to factor login functions for F5 . I am not aware of shared library concept in Ansible . Do you have an example that I can use as an example ? 

Thanks

Etienne-Carriere:
Yes, I plan to factor, in a first phase, the login phase and the associated parameters. Perhaps there will be more elements to factor in the future

PR #11073, opened 2015-05-26
Include more info when a task fails

msabramo:
Adds "playbook", "role", and "task" fields to the output when a task fails.

This makes it easier to pinpoint where the problem is, especially
when you have a lot of roles and playbooks.

e.g.:

```
failed: [vagrant] => {..."playbook": "/Users/marca/dev/ansible/vagrant.yml",
                      ..."role": "pythonapp",
                      ..."task": "pip install -r /opt/src/{{ sm_app_role }}/requirements.txt"...}
```

Cc: @sontek, @sudarkoff


sudarkoff:
Yes, yes, yes! :+1:

bcoca:
please note that this will be superseded in v2 with the new callbacks.

PR #11075, opened 2015-05-27
corrected the playbook name from rolling_upgrade.yml to rolling_updat…

initcron:
…e.yml as per example playbook


amenonsen:
It seems to me that a better fix would be to rename ansible-examples/lamp_haproxy/rolling_update.yml to rolling_upgrade.yml, since it's referred to as a rolling upgrade in guide_rolling_upgrade.rst

dharmabumstead:
Thanks!

PR #11076, opened 2015-05-27
module_utils.basic: set LC_ALL=C in command_run()

resmo:
Instead of passing the env into the command_run(), it should be defaulted to LC_ALL=C.
Closes https://github.com/ansible/ansible-modules-extras/pull/529


bcoca:
I"m trying to find/remember why this was removed at one point as it used to be the default for running modules.

PR #11079, opened 2015-05-28
Avoid too broad except in Inventory

amegianeg:
Capture only IOError when reading shebang from inventory file, to avoid ignoring other possible exceptions like timeouts from a task.
We are using Celery to manage different tasks and if eventually a task times out at this part of the code, it will not rollback as expected and will keep on executing the rest of the code, leading it to a hard timeout


jimi-c:
Merged, thanks!

PR #11083, opened 2015-05-28
safe_item filter plugin to safely get from deep dicts

hkariti:
It's currently very hard to use values from possibly nested dicts where each component can be undefined.

Example expression: `ansible_local.my_fact_file.my_fact`.

Using `default` or `is defined` will still fail if `ansible_local` or `my_fact_file` are undefined. My filter allows one to do this:

`ansible_local | safe_item('my_fact_file.my_fact')`

And not throw an error if any of these values are undefined or isn't a dictionary. A custom default value can be passed:

`ansible_local | safe_item('my_fact_file.my_fact', 'default')`

And the string expression can be expressed more safely as a list:

`ansible_local | safe_item(['my_fact_file', 'my_fact'], 'default')`

### DIFF
--- a/docsite/rst/playbooks_filters.rst
+++ b/docsite/rst/playbooks_filters.rst
@@ -95,6 +95,20 @@ Jinja2 provides a useful 'default' filter, that is often a better approach to fa
 In the above example, if the variable 'some_variable' is not defined, the value used will be 5, rather than an error
 being raised.
 
+Defaulting Undefined Variables in Deep Expressions
+--------------------------------------------------
+
+Ansible provides a filter for a more complex case, where one wishes to set a default value for an expression such as::
+
+    {{ some.deep.item.in.object }}
+
+In this case, the 'default' value will still fail if the parent objects ('some', 'deep', 'item', 'in') are undefined.
+To do this safely::
+
+    {{ some | safe_item('deep.item.in.object', 'default_value') }}
+    {{ some | safe_item(['deep', 'item', 'in', 'object'])}}
+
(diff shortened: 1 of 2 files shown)


tersmitten:
I like it. Way easier to write and maintain.

amenonsen:
I am in two minds about this PR.

I don't think the name `safe_item` is meaningful enough. Perhaps it should be something like `lookupattr`… which makes me think of jinja2's `selectattr` (which I've heard can't be used  in devel right now, though I don't know why, and haven't looked) and `map(attribute='…')`, both of which can take `foo.bar.baz`-style arguments for nested lookups. A quick look at `make_attrgetter` in `jinja2/filters.py` (in the current jinja2 source, not ansible) doesn't show any attempt to deal with intermediate undefined keys, but I can't help think that a flag of some sort to map/selectattr would be the right way to fix this. It can't be a good idea to have lots of different ways to lookup attributes.

On the other hand, I have no idea what/how long it would take to make a change like that to jinja2 (or even if there's some better way to do it already).

hkariti:
I dont like having custom filters for these things too. I just solved this
problem for myself and thought I should share..

sivel:
@hkariti,

Hi!

hkariti:
rebased

bcoca:
Question About This One
====================

Hi!

Unfortunately, we're not sure if we want this feature in the program, it seems to overlap with existing filters already.

luto:
@bcoca you mention that this overlaps with existing filters, which one would that be? `|default()` only applies to one layer.

PR #11092, opened 2015-05-29
Change how ExecutionPolicy is set for running remote scripts 

cchurch:
... so arguments passed via splatting work again.

Fixes this integration test that was failing:

https://github.com/ansible/ansible/blob/devel/test/integration/roles/test_win_script/tasks/main.yml#L49


jhawkesworth:
+1 

All integration tests run through for me now
Ansible controller: Ubuntu 12.04
Target hosts: 1 \* Windows Server 2012 R2 Standard

jhawkesworth:
hmm, sure this didn't contain merge conflicts when I tested it.

cchurch:
Implemented by #11719 instead.

PR #11093, opened 2015-05-29
Don't hide errors in files listed in vars_files

ghost:
Previously, files listed in 'vars_files' could have parse errors or even
not exist at all, and Ansible would not raise any error (until it found
that the expected variables were not defined).

Based on the placeholder comment, I opted to continue to ignore errors
by default, but explicitly raise them when called from the playbook
executor. The test suite passes even if the function raises the errors
in all cases, so maybe that's not necessary.

### DIFF
--- a/lib/ansible/executor/playbook_executor.py
+++ b/lib/ansible/executor/playbook_executor.py
@@ -80,7 +80,8 @@ class PlaybookExecutor:
 
                     # Create a temporary copy of the play here, so we can run post_validate
                     # on it without the templating changes affecting the original object.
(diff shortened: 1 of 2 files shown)


ghost:
Seems the tests are failing with 'ImportError: No module named selinux', but they fail in the same way on the head of the v2_final branch (12691ce109dcf1625c6c41357ce26f95da0862f0)

amenonsen:
@ssssam do you feel like rebasing this onto current devel? From a quick look at the code, it shouldn't be hard to make it apply.

jimi-c:
Closing this, as the v2_final branch is dead.

PR #11097, opened 2015-05-30
cloudstack: updates and v1 sync

resmo:


gregdek:
Sigh.  So I just merged the following PR into Extras: 

https://github.com/ansible/ansible-modules-extras/issues/560

...without realizing, until too late, that this PR is a dependency for the full functionality of that other PR.  And we haven't sorted out a clear policy for "Extras changes with dependencies in ansible/ansible" -- it's the first case I've personally come across.

So we have a few options, and I state them in order of my preference:
1. @resmo creates a new patch for the Extras modules, factoring out the bits that rely on these changes until we have more time to examine them.
2. @bcoca or @abadger or @jimi-c look at these changes and merge them (but not sure they're urgent enough right yet);
3. Revert the whole change for PR 560 in Extras. Eh, it's a big patch, and most of it is probably ok.

Help pls :)

resmo:
In this particular case, the "dependency" is not breaking anything, it is just a param `api_timeout` in the modules that does not change the `api_timeout` if you like to change it. But the modules still work as before with a defaulted api_timeout of 10 secs. 

That is why I did not separate this change. 

However in my opinion generally the policy should be that dependency commits should be sorted out in a separate PR.

bcoca:
I'm going to wait on merging this till v2 is in place, in which case the 3rd file change will actually be a problem, also v2 changes go to v2_final branch as v2/ dir in devel has not been updated

resmo:
No problem, I'll will rebase if needed. Waiting for v2 in place :)

resmo:
@bcoca rebased

PR #11111, opened 2015-06-01
new hash lookup plugin

domibarton:
Created a new lookup plugin to lookup a hash digest from a file. 

The plugin uses Python's [hashlib](https://docs.python.org/2/library/hashlib.html) to create the hash. By default md5() will be used, but it can be overwritten by adding the parameter" algorithm=..." to the filename / term.

The `try / except AttributeError` block around `getattr()` could be enhanced by checking `hashlib.algorithms`, but the `algorithms` attribute is only available in Python >= 2.7, and I wanted to keep backwards compatibility. Hope that's ok.

### DIFF
--- a/lib/ansible/runner/lookup_plugins/hash.py
+++ b/lib/ansible/runner/lookup_plugins/hash.py
@@ -1,0 +1,143 @@
+# (c) 2015, Dominique Barton <dbarton@confirm.ch>
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


jrenggli:
+1

amenonsen:
Quick review: (a) this needs to live in `lib/ansible/plugins/lookup` for current devel, (b) does «lookup a hash digest from a file» mean calculate the digest of a file? (Not a facetious question, this was really not clear to me from just the description; I had to read the code to figure it out: yes, that's what it means), (c) no documentation changes at all, and (d) I really don't like the duplicated path handling. It seems like a better idea to me to add an optional parameter to the existing `file` lookup plugin that returns the hash of the file. That way you won't be adding yet another place for path handling bugs to creep in.

sivel:
@domibarton,

Hi!

amenonsen:
Closing per above comment.

PR #11086, opened 2015-05-29
Part fix for https://github.com/ansible/ansible-modules-core/issues/1404

jhawkesworth:
Checksum function now allways returns a string.
Also had to amend some integration tests as the checksum is no
longer being returned in an array.

PR #11115, opened 2015-06-01
Part fix for https://github.com/ansible/ansible-modules-core/issues/1404 (replaces #11086)

jhawkesworth:
Replacement for previous attempt at this https://github.com/ansible/ansible/pull/11086 which accidentally included some unwanted changes.

Get-FileChecksum always returns a string now and the test_win_copy integration tests that depend on the checksum
have been updated in this change too.

Also requires https://github.com/ansible/ansible-modules-core/pull/1438


jhawkesworth:
@cchurch as you can see I wound up creating a new PR without the unwanted submodules and test_uri change.

PR #11124, opened 2015-06-02
Use a decorator to perform just in time connection

sivel:
Copy of #11002 now against the devel branch

This PR adds a new ensure_connect decorator for connection plugins. It's function is to decorate a method requiring a connection to be open. The replaces the explicit call to _connect.

WHY: Calling _connect explicitly as it is, causes the connection plugin to connect to a remote host, sometimes needlessly such as when running an action plugin that does not need a connection, such as set_fact.

This change will ensure that a connection is not forcefully established when not needed, but still establish it just in time for the operation.

The following connection plugins have been updated due to them being ready for v2:
- ssh
- paramiko
- winrm

Additional plugins will need to have this decorator applied to exec_command, put_file and fetch_file.

I've tested a handful of scenarios and did not encounter any issues.

Addresses #10997

### DIFF
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -210,7 +210,6 @@ class TaskExecutor:
         # get the connection and the handler for this execution
         self._connection = self._get_connection(variables)
         self._connection.set_host_overrides(host=self._host)
-        self._connection._connect()
 
         self._handler = self._get_action_handler(connection=self._connection, templar=templar)
 
--- a/lib/ansible/plugins/connections/__init__.py
+++ b/lib/ansible/plugins/connections/__init__.py
@@ -92,16 +92,19 @@ class ConnectionBase(with_metaclass(ABCMeta, object)):
         """Connect to the host we've been initialized with"""
         pass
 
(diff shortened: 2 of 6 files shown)


abadger:
Seems like a good direction.  We were wondering if setting the decorator on the base class might be good (or problematic).
- Could be good just for documentation
- If people call the base class's method (via super, for instance) at the top of their method then they wouldn't need to put the decorator on their own.
- Probably should put this on the local plugin as well -- I think that one was updated and it is probably one that people will cargo cult from frequently.
- We should also make sure that self._connected is used properly in these plugins (although that could be done separately).

sivel:
Totally missed this comment.  I'll look into putting in the base class soonish and the local plugin too.

sivel:
Ok, with the last commit, I have moved the decorator to the `ConnectionBase` methods, have switched to calling super from within the methods of the individual connection classes and implemented this for the local plugin too.

I haven't touched anything in regards to ensuring the plugins use `self._connected` properly, I think that should be addressed separately.

bcoca:
+1

abadger:
+1

PR #11127, opened 2015-06-02
created makedirs_safe function for use in cases of multiprocess

### LINKED ISSUE #11126: Race condition in fetch module
In ansible/lib/ansible/plugins/action/fetch.py at lines 128 and 129 there is a race condition.
128             if not os.path.isdir(os.path.dirname(dest)): 
129                 os.makedirs(os.path.dirname(dest)) 
If the destination directory does not exist, this code attempts to create it. However if multiple python threads are doing this simultaneously, it is possible for one thread to create the directory in the interval between the test on line 128 and the create on line 129. When that happens the losing thread throws:
  OSError: [Errno 17] File exists
Several solutions are discussed here: http://stackoverflow.com/questions/12468022/python-fileexists-error-when-making-directory

bcoca:
should fix #11126 and most race conditions

### DIFF
--- a/lib/ansible/utils/path.py
+++ b/lib/ansible/utils/path.py
@@ -20,6 +20,7 @@ __metaclass__ = type
 import os
 import stat
 from time import sleep
+from errno import EEXIST
 
 __all__ = ['is_executable', 'unfrackpath']
 
@@ -38,10 +39,9 @@ def unfrackpath(path):
 
 def makedirs_safe(path, mode=None):
     '''Safe way to create dirs in muliprocess/thread environments'''
-    while not os.path.exists(path):
+    if not os.path.exists(path):
         try:
             os.makedirs(path, mode)
         except OSError, e:
-            if e.errno != 17:
+            if e.errno != EEXIST:
                 raise
-            sleep(1)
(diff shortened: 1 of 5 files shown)


lib/ansible/utils/path.py
```diff
@@ -35,3 +36,12 @@ def unfrackpath(path):
     '''
     return os.path.normpath(os.path.realpath(os.path.expandvars(os.path.expanduser(path))))
 
+def makedirs_safe(path, mode=None):
+    '''Safe way to create dirs in muliprocess/thread environments'''
+    while not os.path.exists(path):
+        try:
+            os.makedirs(path, mode)
+        except OSError, e:
+            if e.errno != 17:
```
abadger: Maybe use errno.EEXIST instead of 17?

lib/ansible/utils/path.py
```diff
@@ -35,3 +37,11 @@ def unfrackpath(path):
     '''
     return os.path.normpath(os.path.realpath(os.path.expandvars(os.path.expanduser(path))))
 
+def makedirs_safe(path, mode=None):
+    '''Safe way to create dirs in muliprocess/thread environments'''
+    if not os.path.exists(path):
```
allardhoeve, line 42: This introduces a race condition. Why check for existence here? What does it add?

Issue #11130, opened 2015-06-02
ansible_ssh_host is overriden


sivel:
In `ansible.inventory.host` `ipv4_address` is set to the inventory_hostname.  https://github.com/ansible/ansible/blob/d1b43712870f5331a58abe115911725619264ca5/lib/ansible/inventory/host.py#L78

Later in `ansible.executor.task_executor._get_connection` `self._connection_info.remote_addr` is set to the `ipv4_address`.  https://github.com/ansible/ansible/blob/d1b43712870f5331a58abe115911725619264ca5/lib/ansible/executor/task_executor.py#L374

Before that re-assignment in `_get_connection`, `remote_addr` is correct, after that assignment, it is now the inventory_hostname.

I am not sure what the "correct" fix is, but removing the re-assignment in `_get_connection` seems to resolve the issue for me.

bcoca:
closing via #11131

PR #11131, opened 2015-06-02
Don't override ansible_ssh_host with inventory_hostname

sivel:
This PR addresses #11130 

It seems that this line was added before ConnectionInformation was fully functional, and does not appear to be needed any longer.


jimi-c:
Merged, thanks!

PR #11134, opened 2015-06-02
Add missing import in ansible.cli

sivel:
This PR adds a missing import in `ansible.cli`.

If `display` is `None` in the `CLI` class, display will need to be initialized with `Display` which hasn't been imported


jimi-c:
Merged, thanks!

PR #11135, opened 2015-06-02
If remote_addr isn't set, set to ipv4_address

sivel:
I wasn't able to previously determine what exactly the re-assignment to the `ipv4_address` was needed for.

After more playing, it seems there are some circumstances where `remote_addr` is not set, and should be set to the `ipv4_address`.

This PR adds the line back in, but under an if statement to first check if it is not set.


jimi-c:
Merged, thanks!

PR #11137, opened 2015-06-02
New lookup plugin: merged_dicts

telbizov:
Input:  a list of dictionaries (order in the list is important).
Output: a single dictionary containing all the unique keys from all the
input dictionaries with values from the last dictionary found in the
list.
Note:   other than the merging function it acts just like with_dict.

### DIFF
--- a/v1/ansible/runner/lookup_plugins/merged_dicts.py
+++ b/v1/ansible/runner/lookup_plugins/merged_dicts.py
@@ -1,0 +1,47 @@
+# (c) 2015, Rumen Telbizov <telbizov@gmail.com>
+#
+# This file is part of Ansible
+#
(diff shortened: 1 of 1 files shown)


telbizov:
Not sure if I was supposed to change the non-v1 (v2?) version of it. This is my first pull request and wasn't quite sure of the structure and/or if it would be found useful at all. Could revisit later.

bcoca:
v1 is meant as a 'quick switch to legacy' , nothing should be updated there.

now all plugins reside under lib/ansible/plugins/

telbizov:
OK, I'll update the pull request. Aside from that, would it be merged?

telbizov:
I'll close this one and I'll submit a new pull request for this change once I apply it in the correct place as advised.

PR #11139, opened 2015-06-03
Handle when role_dependencies is None.

wenottingham:
For some roles, role_dependencies will be returned as None instead of an empty list. Coerce it to an empty list.


bin/ansible-galaxy
```diff
@@ -845,6 +845,8 @@ def execute_install(args, options, parser):
                     role_dependencies = role_data['dependencies']
```
bcoca, line 845: another approach that has been suggested:

```
 role_dependencies = role_data.get('dependencies', [])
```

PR #11140, opened 2015-06-03
Fix for task_executor on OS X

RadishTheHut:
I get this exception during the setup task:

AttributeError: 'ConnectionInformation' object has no attribute 'remote_pass'

I believe it is supposed to be looking at the password attribute. Either that
or we should create a remote_pass attribute in ConnectionInformation.


bcoca:
the field was renamed in v2 at one point, it seems we missed one, thanks!

PR #11142, opened 2015-06-03
Add common fonctions for F5 modules (FQ Name functions)

Etienne-Carriere:
Complete the common F5 functions

### DIFF
--- a/lib/ansible/module_utils/f5.py
+++ b/lib/ansible/module_utils/f5.py
@@ -50,7 +50,7 @@ def f5_parse_arguments(module):
         module.fail_json(msg="the python bigsuds module is required")
     if not module.params['validate_certs']:
         disable_ssl_cert_validation()
-    return (module.params['server'],module.params['user'],module.params['password'],module.params['state'],module.params['partition'])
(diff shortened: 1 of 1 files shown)


Etienne-Carriere:
Begin of the code factor : https://github.com/ansible/ansible-modules-extras/pull/574

lib/ansible/module_utils/f5.py
```diff
@@ -62,3 +62,19 @@ def disable_ssl_cert_validation():
...
     ssl._create_default_https_context = ssl._create_unverified_context
 
+# Fully Qualified name (with the partition)
+def fq_name(partition,name):
+    if name is None:
+        return None
+    if name[0] is '/':
+        return name
+    else:
+        return '/%s/%s' % (partition,name)
```
bcoca, line 68: a simpler expression of same:

```
if name is not None and name[0] != '/: #  or .startswith('/')
    return '/%s/%s' % (partition,name)

return name
```

Etienne-Carriere:
@bcoca : applied the proposed simplification

PR #11147, opened 2015-06-03
Import missing MutableMapping class for module_docs

sivel:
Currently building the docs is failing right now due to a missing import:

```
  File "/home/sivel/.virtualenvs/ansible/ansible/lib/ansible/utils/module_docs.py", line 89, in get_docstring
    if isinstance(doc[key], MutableMapping):
NameError: global name 'MutableMapping' is not defined
```

This PR addresses this by importing `MutableMapping`.

### DIFF
--- a/lib/ansible/utils/module_docs.py
+++ b/lib/ansible/utils/module_docs.py
@@ -23,6 +23,7 @@ import ast
 import yaml
 import traceback
 
+from collections import MutableMapping
 from ansible.plugins import fragment_loader
 
 # modules that are ok that they do not have documentation strings


sivel:
After merging this, it may be a good idea to force another build of the web docs for docs.ansible.com.

bcoca:
no worries, i'll build locally first (i was in the middle of making this same PR ;-) ), already updated site with stable-1.9, which in the end is what most people want anyways.

bcoca:
also added the missing mutableset/sequence, etc

PR #11153, opened 2015-06-03
New lookup plugin: merged_dicts

telbizov:
Input:  a list of dictionaries (order is important)
Output: a single dictionary with all the keys from the source
dictionaries and values from the last dictionary in the input list which
contained the given key

### DIFF
--- a/lib/ansible/plugins/lookup/merged_dicts.py
+++ b/lib/ansible/plugins/lookup/merged_dicts.py
@@ -1,0 +1,36 @@
+# (c) 2015, Rumen Telbizov <telbizov@gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import (absolute_import, division, print_function)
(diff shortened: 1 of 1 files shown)


telbizov:
A rewrite of the previously closed #11137. This is my first pull request for ansible so please test thoroughly and comment.

telbizov:
@bcoca now that the feature_pull_request has been added does that mean that it will be merged or there's a process that it needs to go through ?

bcoca:
just need to review and test on our side, sorry for the delays but we are going through a very large backlog

telbizov:
No worries. I understand. Thank you for the update.

telbizov:
@bcoca any chance you've had a moment to look at merging this?

bcoca:
it is on my list, also to compare to another request that adds dict support to union/difference/intersection that might overlap with this

telbizov:
Hey @bcoca,

I was wondering if there's any development on this pull request. It has been more than a month. Any ETA?

Thank you,
Rumen Telbizov

bcoca:
sorry, have not had time to come back, its on my short list, but we do have a big backlog we are running through.

telbizov:
Thanks for the reply @bcoca. Since adding my own lookup plugin in ansible 1.9 is broken and doesn't work I can't use this functionality even if I wanted to add my own plugin bundled withing a role. I understand that the bug will be fixed in ansible 2.0 and then I wouldn't need this merged that urgently. 

Two questions are on my mind and I'd appreciate if you can comment:
- What would be a rough ETA on ansible 2.0 release?
- If you merge this pull request before 2.0 release would the plugin be available in ansible 1.9 as well?

Thank you in advance @bcoca.

Regards,
Rumen Telbizov

bcoca:
new features are only added to new releases, not backported, so the plugin is not planned for 1.9 because of this.

2.0 is expected at end of the month or possibly August, depends on critical bugs more than anything else.

telbizov:
Thank you for the update.

amenonsen:
Note that this is related to, but not the same as, the functionality proposed in #7872. I'm not sure I like the idea of this being implemented as a lookup plugin, though. Seems more natural to have a filter that allows you to merge, and then lookup as normal.

PR #11161, opened 2015-06-04
oVirt Dynamic Inventory

joshainglis:
Used gce.py as a template to create an ovirt dynamic inventory script


plugins/inventory/ovirt.ini
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/python
```
bcoca: shebang on ini file?

joshainglis:
Thanks @bcoca.
Removed the shebang

PR #11164, opened 2015-06-04
ansible_mount_<mount> facts in addition to array of mounts

shawnferry:
This change adds enumerated ansible_mount_<mountpoint> with one special case for / as "root" It does not change the existing ansible_mounts array based on the assumption that someone is already using the other facts as they are today.

This allows direct selection of a mount by name instead of looping over the list of all mounted filesystems to find the desired mount.

```
        "ansible_mount_orahome": {
        "device": "/dev/mapper/vg-lv_orahome", 
        "fstype": "ext4", 
        "mount": "/orahome", 
        "options": "rw", 
        "size_available": 200443834368, 
        "size_total": 211378954240, 
        "uuid": ""
    }, 
    "ansible_mount_root": {
        "device": "/dev/mapper/vg-lv_root", 
        "fstype": "ext4", 
        "mount": "/", 
        "options": "rw", 
        "size_available": 10591088640, 
        "size_total": 15919398912, 
        "uuid": ""
    }, 
```

Full example output before and after  https://gist.github.com/shawnferry/0c0cb623495e1f6bc715

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -889,6 +889,22 @@ class LinuxHardware(Hardware):
                          'uuid': uuid,
                          })
 
+                    if (fields[1] == '/'):
+                        mount_name = "_root"
+                    else:
+                        mount_name = fields[1].replace('/','_')
+
+                    self.facts['mount' + mount_name] = {
+                         'mount': fields[1],
+                         'device':fields[0],
+                         'fstype': fields[2],
+                         'options': fields[3],
+                         # statvfs data
+                         'size_total': size_total,
+                         'size_available': size_available,
+                         'uuid': uuid,
+                         }
+
     def get_device_facts(self):
         self.facts['devices'] = {}
         lspci = module.get_bin_path('lspci')


bcoca:
You have a commit that updates the submodule reference, which is probably why this is not 'mergable'.

I see the convenience of this feature, but I am remiss to just add this to the facts when it produces no new data and the existing copy can be easily coerced to achieve the same functionality.

shawnferry:
I'm still getting a handle on git that was also changed in the commit. I'll see if I can't back it out.

Preferentially I would remove the ansible_mounts array and/or replace it with one of the following:

Similar to ansible_interfaces:
ansible_mounts = [ /, /var, etc]
ansible_mount_root = {}, ansible_mount_var = {}

Similar to ansible_devices (this would be my preference):
ansible_mounts = { "root" = {}, "var" = {} }

The downside I see is that it would break any code that currently iterates over the current ansible_mounts array. I'm not certain that anyone else is using ansible_mounts so maybe that doesn't really matter but I don't see a way to search across galaxy

bcoca:
it is in use, that I do know for sure given that I wrote that use in some cases and plenty of people have added options or opened tickets when it has been broken.

shawnferry:
I'm not doubting that people are using ansible_mounts I'm just not sure how. Are they looping over every mount all the time or identifying individual mounts as I am. 

What I want here is an elegant way to identify the collected facts for a specific mount without executing an extra task to collect/identify it. 

The only way I see to do it today is with the extra task iterating over the mounts registering the match (or registering the index).
  register: root_mount = item
  with_items: ansible_mounts
  when: item.mount == "/"

A list comprehension (if you could use it in a play) seems slightly cleaner but it would be pulling code into the playbook.
[ m for m in ansible_mounts if m['mount'] == '/' ][0]

Alternatively creating a list of the mounts with matching ordering would work such that ansible_mounts[ansible_mounts_list.index('/')] is a viable accessor would maintain the original fact behavior.

Am I missing the elegant solution to access this data?

bcoca:
the jinja2 map filter?

shawnferry:
_sigh_ exactly that

PR #11165, opened 2015-06-04
Revert '-' replacement

hvpareja:
Please, have a look to this comment: https://github.com/ansible/ansible/commit/c7f33627950d352cdca46f71c93b0783981b8c89 for further discussion.

### DIFF
--- a/plugins/inventory/ec2.py
+++ b/plugins/inventory/ec2.py
@@ -791,7 +791,7 @@ class Ec2Inventory(object):
         ''' Converts 'bad' characters in a string to underscores so they can be
         used as Ansible groups '''
 
-        return re.sub("[^A-Za-z0-9\_]", "_", word)
+        return re.sub("[^A-Za-z0-9\-]", "_", word)
 
 
     def json_format_dict(self, data, pretty=False):


abadger:
CC: @discordianfish

bcoca:
this replacement was done to avoid many issues caused by having the '-', what is the reason to revert this?

hvpareja:
I understand, but I think this is just a work around and shouldn't be implemented. It is just my point of view, but I see more dangerous altering the data we're getting from AWS, such as the instances names (See https://github.com/ansible/ansible/commit/c7f33627950d352cdca46f71c93b0783981b8c89, please).

If AWS allows dashes in the tags, why Ansible is replacing them? At least we should keep the instances attributes intact.

bcoca:
the issue is with those tags becoming groups, but I see your point, maybe switching this to be configurable and documenting well why this option exists.

hvpareja:
Thanks for your quick responses. Could you point me to further information about such issue in the meantime?

Regards

ivanvpan:
I agree with @hvpareja. After the latest install of this plugin all of our scripts stopped working. It turned out that it was because it was replacing the dashes. We have dozens of machines with dashes in their tags. I understand that perhaps this fixes some things, but it also makes Ansible unusable for people like me (I can't go and replace all the tags on all machines). In other words just replacing the data from AWS with a '_' is not a "fix". I hope this pull request gets merged.

ivanvpan:
I agree with @hvpareja. This is not a good "fix". It broke all of our scripts as well. We could change them, but yet simply altering valid AWS tag names seems like a hack.

bennojoy:
@ivanpan i agree with @bcoca we should make this configurable in ec2.ini, i have added this to my todo list, but if someone else wants to do this before me i will be glad :)

hvpareja:
@bcoca, @bennojoy, @ivanvpan:

Make this configurable in ec2.ini means that the user should decide between two bugs: whether to assume the miss-functionality of having dashes in the ec2 instance names or the issue with those tags becoming groups. I think this approach can even cause other unexpected potential issues.

Unfortunately, I'm not familiar with the Ansible source, but I do think this problem needs to be addressed in a proper way and solve the initial problem directly, without adding more potentially confusing config options or any other work around.

Regards

bcoca:
@hvpareja sadly the issue is with python, the underlying language ansible uses, this is not an easy limitation to get around.

hvpareja:
@bcoca, yeah ... I though so. Could you please, point me to any further information about this issue, I really would like to think a bit deeper in order to suggest (if I can) any other approach.

Thanks

bcoca:
I think this is the simplest explanation:
http://www.pasteur.fr/formation/infobio/python/ch02s03.html

we do a lot of implicit transformations of groups into variables for usage in ansible and templates, moving all these to be explicit is a big undertaking.

hvpareja:
@bcoca Thank you

amenonsen:
I guess this PR should be closed now?

bcoca:
closing this as per above, we will consider a PR that makes this a option toggle

lanceblais:
Did this ever show up somewhere else? We just upgraded to 1.9.4 and are seeing this ourselves!

PR #11170, opened 2015-06-04
Fix lookup() plugin

dguerri:
lookup() plugin is currently broken because _get_file_contents() now
returns a tuple: (contents, show_data).

This patch fix that issue.


jimi-c:
Merged, thanks!

Issue #11179, opened 2015-06-05
ansible-galaxy install throws an error when the role-file option is used.


bandesz:
##### Issue Type:

Bug Report
##### Ansible Version:

```
ansible 2.0.0 (devel a4ebee4b16) last updated 2015/06/05 11:02:21 (GMT +000)
  lib/ansible/modules/core: (detached HEAD b138411671) last updated 2015/06/05 11:02:40 (GMT +000)
  lib/ansible/modules/extras: (detached HEAD 1276420a3a) last updated 2015/06/05 11:03:14 (GMT +000)
  v1/ansible/modules/core: (detached HEAD f8d8af17cd) last updated 2015/06/05 11:04:10 (GMT +000)
  v1/ansible/modules/extras: (detached HEAD 495ad450e5) last updated 2015/06/05 11:04:48 (GMT +000)
  configured module search path = None
```
##### Environment:

Debian 8.0
##### Summary:

ansible-galaxy install throws an error when the role-file option is used.
##### Steps To Reproduce:

```
$ ansible-galaxy install -r requirements-roles-3rd.yml -p ./roles-3rd/
```
##### Expected Results:

Great success. 
##### Actual Results:

```
Traceback (most recent call last):
  File "/home/vagrant/ansible/bin/ansible-galaxy", line 66, in <module>
    sys.exit(cli.run())
  File "/home/vagrant/ansible/lib/ansible/cli/galaxy.py", line 134, in run
    self.execute()
  File "/home/vagrant/ansible/lib/ansible/cli/__init__.py", line 96, in execute
    fn()
  File "/home/vagrant/ansible/lib/ansible/cli/galaxy.py", line 321, in execute_install
    role_name = self.args.pop(0).strip()
IndexError: pop from empty list
```

bcoca:
I had not finished implementing all galaxy functions, this is high on my todo list.

dkinzer:
I'm getting this same error: 

```
#ansible --version:
ansible 2.0.0 (devel 4705a79a98) last updated 2015/06/16 11:04:58 (GMT -400)
  lib/ansible/modules/core: (detached HEAD 9acc7c402f) last updated 2015/06/16 11:05:01 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 2f967a949f) last updated 2015/06/16 11:05:03 (GMT -400)
  v1/ansible/modules/core: (detached HEAD f8d8af17cd) last updated 2015/06/16 11:05:04 (GMT -400)
[...]

PR #11181, opened 2015-06-05
Fix bug report #11179 - ansible-galaxy install throws an error when the role-file option is used

bandesz:
Bug report: https://github.com/ansible/ansible/issues/11179


bandesz:
I opened this pull request too early, closing it down, the solution is not this straightforward - sadly this was never tested before the original merge :( I won't be able to fix this.

bcoca:
we merged knowing some functionality was still not done, this is one of such things, its high on my todo list

Issue #11169, opened 2015-06-04
sudo password broken after v2 modules merge


agunnerson-ibm:
##### Issue Type:

Bug Report
##### Ansible Version:

```
ansible 2.0.0 (devel f3f3fb7c49) last updated 2015/06/04 14:13:40 (GMT -400)
  lib/ansible/modules/core: (detached HEAD b138411671) last updated 2015/06/04 14:01:51 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 1276420a3a) last updated 2015/06/04 14:01:52 (GMT -400)
  v1/ansible/modules/core: (detached HEAD f8d8af17cd) last updated 2015/06/02 10:07:39 (GMT -400)
  v1/ansible/modules/extras: (detached HEAD 495ad450e5) last updated 2015/06/02 10:07:43 (GMT -400)
  configured module search path = None
```
##### Ansible Configuration:

Nothing has been changed. The inventory file is a simple file with a single IP with no host/group variables.
##### Environment:

Tested in Ubuntu 15.04 and OSX 10.10, but very likely affects all OS's.
##### Summary:

Ever since the v2 modules merge, using sudo with a password no longer works. It simply hangs until sudo times out. ~~This is likely due to code that was commented out during the v1 -> v2 transition: https://github.com/ansible/ansible/blob/f3f3fb7c491effe9e61ae5a429ac796558c2963a/lib/ansible/plugins/connections/ssh.py#L313~~
##### Steps To Reproduce:

Run any ansible command with `-K --sudo`.
##### Expected Results:

Ansible uses sudo password and runs the command successfully.
##### Actual Results:

Hangs until sudo times out. For example:

```
ansible -i inventories/docker-centos docker-centos -a 'echo hello!' -Kk --sudo -u centos -vvvv
SSH password:
SUDO password[defaults to SSH password]:
<A.B.C.D> ESTABLISH CONNECTION FOR USER: centos on PORT 22 TO A.B.C.D
<A.B.C.D> EXEC mkdir -p $HOME/.ansible/tmp/ansible-tmp-1433441710.88-220550492594694 && chmod a+rx $HOME/.ansible/tmp/ansible-tmp-1433441710.88-220550492594694 && echo $HOME/.ansible/tmp/ansible-tmp-1433441710.88-220550492594694
[...]

PR #11183, opened 2015-06-05
Don't empty out become_pass. See #11169

sivel:
This PR ensures that the `become_pass` supplied via the CLI (prompt) does not get emptied out.

become_pass is not part of `options` so we don't need to try to set it in `set_options`, and only set it from the play if it is actually set.

There is still more work to do to solve #11169, but this is a start.

### DIFF
--- a/lib/ansible/executor/connection_info.py
+++ b/lib/ansible/executor/connection_info.py
@@ -109,8 +109,6 @@ class ConnectionInformation:
             self.become_method = play.become_method
         if play.become_user:
             self.become_user   = play.become_user
-        if play.become_pass:
-            self.become_pass   = play.become_pass
 
         # non connection related
         self.no_log      = play.no_log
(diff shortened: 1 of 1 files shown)


bcoca:
.... we should not be allowing people to set passwords in play ....

bcoca:
I think this was a mistake holding over from how vars were passed in v1, become password should only be settable from 2 points, command line or inventory var

sivel:
The last commit, removes the ability to set become_pass in a play.

jsengla:
Following this issue with a need for variable-password based ssh and sudo for a while, I still wonder why not allow setting become_pass in a play?
I have a use case in a large, secure enterprise environement where we use One-Time Tokens for _some_ ssh logins and _every_ sudo call.
Every sudo password is different, even on the same host from one call to the next.  I have this kludged with a local pipe lookup to ssh-askpass set as the ansible_sudo_pass in the inventory group_vars.  While the sudo password gets looked up on _every_ play, even with "sudo: false" set (not optimal), this is the only way I've found to be able to use ansible in our environement.
I understand you may be resolving an issue with broken sudo passwords, but while you do this, please don't overlook this use scenario, where a sudo password can change not just from one play to the next, but even between tasks within a play.

PR #11188, opened 2015-06-05
Reverses delegate_to host and remote host.

### LINKED ISSUE #11067: synchronize swaps the source and target in the output
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.9.1
##### Ansible Configuration:

```
[defaults]
inventory       = ~/inventory
```
##### Environment:

Running at: Mac OS X 10.10.3
Managing: Debian 7.8
##### Summary:

synchronize works, but swaps the source and target in the output result.
##### Steps To Reproduce:

file "inventory":

```
[office-servers]
server1 ansible_ssh_host=192.168.1.1

[office-servers:vars]
ansible_ssh_user=root
ansible_ssh_private_key_file=~/.ssh/id_rsa_office
```

file "office-servers.yml':

```
- hosts: office-servers
  remote_user: root
  gather_facts: no
  roles:
    - firehol
```

file "roles/firehol/tasks/main.yml":

```
- name: Sync common files
  synchronize: archive=no recursive=yes times=yes perms=yes checksum=yes src=common/etc/firehol/ dest=/etc/firehol
```
##### Expected Results:

`changed: [127.0.0.1 -> server1]`
##### Actual Results:

ansible-playbook office-servers.yml 

```
PLAY [office-servers] ********************************************************* 

TASK: [firehol | Sync common files] ******************************************* 
changed: [server1 -> 127.0.0.1]

server1                  : ok=1    changed=1    unreachable=0    failed=0  
```

sysadmin75:
Reverses the delegate_to host and the managed host.
Fixes #11067
Patched against the 1.9 devel branch.

### DIFF
--- a/lib/ansible/callbacks.py
+++ b/lib/ansible/callbacks.py
@@ -507,7 +507,7 @@ class PlaybookRunnerCallbacks(DefaultRunnerCallbacks):
 
     def on_ok(self, host, host_result):
         if self.runner.delegate_to:
-            host = '%s -> %s' % (host, self.runner.delegate_to)
+            host = '%s -> %s' % (self.runner.delegate_to, host)
 
         item = host_result.get('item', None)
 


sivel:
I don't think this is correct.  I believe the way it is now is the correct order.

When specifying something like `delegate_to: localhost` it should display:

```
actual_host -> localhost
```

To specify that localhost is the one actually running the task.

Your change would make this:

```
localhost -> actual_host
```

Which to me would indicate that `actual_host` ran the task.

I wrote that functionality, and it was my intention to have it working the way it is currently.

sivel:
I think to solve #11067, a different approach is required.

bcoca:
I agree with @sivel, we are not going to merge this request as it breaks the expected output on delegation.

sysadmin75:
I figured this might be the case but I figured it was easier for the gurus to provide the QA than it was for me to do the testing. It was worth a try.

PR #11193, opened 2015-06-06
Don't test for play.become_pass any longer

sivel:
In #11183 we removed the ability to set `become_pass` at the play level.

This is now causing tests to fail.  This PR removes those tests.

NOTE: I believe we still allow setting `become_pass` at the task level.  That will need to be addressed in in a separate effort.


bcoca:
I removed become_pass from being inherited by any playbook object, so it should not be settable anymore at play/block/role/include nor task level

jsengla:
How is a user supposed to set the become/sudo password in a play if it changes from one play to the next?  Think One-Time-Passcodes.  A number of large, secure organizations use them, and I've found no way that Ansible supports it.

PR #11207, opened 2015-06-08
Clarify reason for failure

mackn:
Saying "will not have been written" is misleading as the file probably existed before the "sync:" timeout value elapsed.


bcoca:
actually, both might be the case

amenonsen:

Issue #11217, opened 2015-06-10
winrm stops working in latest trunk revision


rophy:
##### Issue Type:

Bug Report
##### Ansible Version:

```
# ansible --version
ansible 2.0.0 (devel 757fb39a2e) last updated 2015/06/10 12:15:29 (GMT +800)
  lib/ansible/modules/core: (detached HEAD d6ed6113a7) last updated 2015/06/10 12:15:50 (GMT +800)
  lib/ansible/modules/extras: (detached HEAD 57813a2e74) last updated 2015/06/10 12:16:11 (GMT +800)
  v1/ansible/modules/core: (detached HEAD f8d8af17cd) last updated 2015/06/10 12:16:30 (GMT +800)
  v1/ansible/modules/extras: (detached HEAD 495ad450e5) last updated 2015/06/10 12:16:51 (GMT +800)
  configured module search path = None
```
##### Ansible Configuration:

Clean check out from git
##### Environment:

Ansible Control Machine: fedora 19
Ansible Target: Windows 2008 R2 SP1
##### Summary:

Since I update to latest trunk, winrm stopped working.
I'm not sure if it's due to misconfiguration, but things were working fine before I update trunk.
- /etc/ansible/hosts contain 1 machine under [windows]
- /etc/ansible/group_vars/windows.yml contains 

```
ansible_ssh_user: Administrator
ansible_ssh_pass: _password_ 
ansible_ssh_port: 5986
ansible_connection: winrm
```
##### Steps To Reproduce:

```
$ git clone git://github.com/ansible/ansible.git --recursive
$ cd ./ansible
$ source ./hacking/env-setup
$ pip install http://github.com/diyan/pywinrm/archive/master.zip#egg=pywinrm
$ pip install kerberos
$ ansible windows -m win_ping  -vvvv
```
##### Expected Results:

ansible should run winrm module and pong the ping.
##### Actual Results:

Seems ansible is not using winrm:

```
[root@lsc102 ansible]# cat /etc/ansible/hosts
lsc102.tw.ibm.com

[windows]
lsc624.tw.ibm.com

[root@lsc102 ansible]# ansible windows -m win_ping  -vvvv
<lsc624.tw.ibm.com> ESTABLISH SSH CONNECTION FOR USER: root
[...]

Issue #11132, opened 2015-06-02
ansible_ssh_user not being picked up


jtktam:
updated the latest version and ansible is not picking up the ansible_ssh_user in my group_vars, if i specify it on the command line it works fine.

---
# development.yml

```

---
[lbservers]
dev-lb1 ansible_ssh_host=x.x.x.x

[webservers]
dev-www1 ansible_ssh_host=y.y.y.y

[dbservers]

[allservers:children]
lbservers
webservers
dbservers

[devservers:children]
allservers
```

---
# group_vars/allservers.yml

```

---
ansible_ssh_user: root

----------
# webservers.yml

---
- hosts: webservers
  roles:
    - php5
```

---

```
ansible-playbook -i development.yml lbservers.yml

PLAY: ***************************************************************************

TASK [setup] ********************************************************************
fatal: [dev-www1]: FAILED! => {"msg": "ERROR! SSH encountered an unknown error. The output was:\n", "failed": true}
```

---

```
ansible-playbook -i development.yml lbservers.yml --user=root

PLAY: ***************************************************************************

TASK [setup] ********************************************************************
ok: [dev-www1]

PLAY: ***************************************************************************
```

jtktam:
Actually looks like the entire group_vars/allservers.yml is not being picked up, there are other variables in there and they are returning as undefined in the scripts

---

ansible-playbook -i development.yml lbservers.yml --user=root

PLAY: ***************************************************************************

TASK [setup] ********************************************************************
ok: [dev-www1]

PLAY: ***************************************************************************

TASK [setup] ********************************************************************
ok: [dev-lb1]

TASK [common : Copy firewall rules] *********************************************
fatal: [dev-lb1]: FAILED! => {"msg": "AnsibleUndefinedVariable: ERROR! Unable to look up a name or access an attribute in template string. Make sure your variable name does not contain invalid characters like '-'.", "failed": true, "changed": false}

bcoca:
possibly related/overlapped with #11130

jtktam:
# group_vars/allservers.yml

```
ansible_ssh_user: root

public_ip: "{{ ansible_eth0.ipv4.address }}"
private_ip: "{{ ansible_eth1.ipv4.address }}"

public_iface: "{{ ansible_eth0.device }}"
private_iface: "{{ ansible_eth1.device }}"

public_httpd_port: 80
private_httpd_port: 80
```

---

# roles/common/templates/iptables.sav.j2

```
*filter
[...]

PR #11218, opened 2015-06-10
fixes issue where hostvars are not loaded in the Host object

bennojoy:
A proposed fix for #11217 which actually happens because the hostvars are not loaded into the host object but loaded only into varialblemanagers.host_vars_files. 
The connection object uses host.get_vars() method to set the connection variables and since the hostvars are not available there it uses the default values.

### DIFF
--- a/lib/ansible/inventory/__init__.py
+++ b/lib/ansible/inventory/__init__.py
@@ -661,10 +661,16 @@ class Inventory(object):
             if group and host is None:
                 # load vars in dir/group_vars/name_of_group
                 base_path = os.path.join(basedir, "group_vars/%s" % group.name)
+                #The variables also needs to go into the Host objects vars key
+                if self._loader.path_exists(base_path):
+                    results = self._loader.load_from_file(base_path)
                 self._variable_manager.add_group_vars_file(base_path, self._loader)
             elif host and group is None:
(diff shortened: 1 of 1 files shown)


abadger:
Is  there a way to update the connection plugins to load from the variablemanager instead?

bennojoy:
@abadger  the update of connection params via hostvars happens in the wokerprocess via 
new_connection_info = connection_info.set_task_and_host_override(task=task, host=host)

currently we only pass task object and the host object, the workerprocess also doesnt have an instance of variable manager (afais) so that we can pass to it , Also i don't see an a way to get an instance of variablemanager from the task executor. so this is the only way i could think of , also in v1 we have the variables in host.vars so went that way.

sivel:
I also have another proposed solution in #11132

I haven't had the opportunity to talk to jimi about it.

abadger:
@bennojoy One of the problems with v1 was that variables were stored all over the place.  VariableManager was intended to address some of that.  However, @jimi-c  might not want to send variablemanager to the worker processes for a reason (isolation?  Memory use?) so we do need to talk to him to see what the right thing to do is here.

jimi-c:
Closing in favor of the fix implemented for #11132.

PR #11219, opened 2015-06-10
ansible-pull: Add option to verify gpg signature of a commit

Erethon:
Add option '--verify-commit' to verify a GPG signature of the checked
out commit. As noted in the git module documentantion, this requires
git version >= 2.1.0

This needs the new git module option `verify-commit` introduced in
issue ansible/ansible-modules-core/issues/1491


resmo:
+1
ansible pull currently supports only git repos but there may be other repos like svn, hg, or even a tar.gz in the future. 

This feature would be for git only, that is why I suggest to rename it to `--git-verify-commit` and remove option `-V`

bcoca:
I would leave the --verify-commit but make note in the help that it depends on the underlying VCS system and corresponding module supporting the feature, also no -V I believe we were reserving that for something

resmo:
@Erethon looks like @bcoca would merged it in, if you would update the help message to make clear this is git only, requires git >= 2.1.0 and if you remove the `-V`

Erethon:
On it, will update the commit in a few minutes.

PR #11221, opened 2015-06-10
Update README.md

arcz:
There are over 1000 contributors now :)


bcoca:
we might need to make this 'dynamic'

PR #11231, opened 2015-06-11
Fixes #11216

### LINKED ISSUE #11216: ansible_distribution is "NA" on Arch Linux
```
$ ansible localhost -m setup -a 'filter=ansible_distribution'
127.0.0.1 | success >> {
    "ansible_facts": {
        "ansible_distribution": "NA"
    }, 
    "changed": false
}
```

remexre:


bcoca:
fixed in a more generic way in f174682e1903e246c9f7389e2e76ffcca4a04c28, this should solve the issue for any distro that has a value from the python functions already.

PR #11246, opened 2015-06-12
Update guide_gce.rst

soupdiver:
Make the docs more specific.


soupdiver:
I easily overlooked these points and I think it helps to be a bit more specific :)

PR #11252, opened 2015-06-12
Display a warning when using a deprecated module

sivel:
This PR adds a warning when using a deprecated module.  Currently a user would never know the module was deprecated unless they looked at the docs.

This should help users identify that they are using a deprecated module, to enable us to eventually remove those modules.

Example output:

```
TASK [glance_image] *************************************************************
 [WARNING]: glance_image has been deprecated, which means it is kept for
backwards compatibility but usage is discouraged.  The module documentation
details page may explain more about this rationale.
```

The text there is largely taken from the text used on docs.ansible.com.

If desired we could also get the docs and display the specific module deprecation string.

### DIFF
--- a/lib/ansible/plugins/__init__.py
+++ b/lib/ansible/plugins/__init__.py
@@ -247,6 +247,14 @@ class PluginLoader:
             for alias_name in ('_%s' % n for n in potential_names):
                 # We've already cached all the paths at this point
                 if alias_name in self._plugin_path_cache:
+                    if not os.path.islink(self._plugin_path_cache[alias_name]):
+                        d = Display()
+                        d.warning('%s has been deprecated, which means '
+                                  'it is kept for backwards compatibility '
+                                  'but usage is discouraged. The module '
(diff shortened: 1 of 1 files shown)


bcoca:
you are warning on both aliases and deprecated, this might be confusing as they are not the same in function nor in the documentation.

sivel:
We display deprecation notices in docs only based on `_`.  I was going to look at the docs for the `deprecation` but it doesn't look like we actually concern ourselves with that in core.  Not that I can't do it, just that we don't do it currently.

bcoca:
no, it has to have start with '_' and be file, if it is a symlink there is no deprecation (there is no doc either), which means it is an alias.

sivel:
So checking if it is actually a file would be the correct way to go?

bcoca:
yep, I should really make this a method somewhere

sivel:
The latest commit + squash, includes a change to ignore symlinks.  I guess I had missed that line when originally inspecting how it was handled elsewhere.

PR #11259, opened 2015-06-14
Get-FileChecksum doesn't calculate hash for files

dmitrydigi:
missing update of the checksum value before returning the result - $hash is never updated in case file exists


jhawkesworth:
Thanks for this but this is already fixed in latest devel following the merge of https://github.com/ansible/ansible/pull/11115
(which also includes updates to winrm integration tests which would otherwise fail).

bcoca:
closing this in favor of #11115

PR #11267, opened 2015-06-15
[FIX] ansible-galaxy missing an exception import

Lujeni:
``` bash
# without the fix
$ ansible-galaxy 
  File "./bin/ansible-galaxy", line 38, in <module>
    from ansible.errors import AnsibleError, AnsibleOptionsError, AnsibleParserError
ImportError: cannot import name AnsibleOptionsError
```

``` bash
# after
$ ansible-galaxy 
Usage: ansible-galaxy [init|info|install|list|remove] [--help] [options] ...

Options:
  -h, --help     show this help message and exit
  -v, --verbose  verbose mode (-vvv for more, -vvvv to enable connection
                 debugging)
  --version      show program's version number and exit
ERROR! Missing required action
```

### DIFF
--- a/lib/ansible/cli/__init__.py
+++ b/lib/ansible/cli/__init__.py
@@ -31,7 +31,7 @@ import subprocess
 
 from ansible import __version__
 from ansible import constants as C
-from ansible.errors import AnsibleError
+from ansible.errors import AnsibleError, AnsibleOptionsError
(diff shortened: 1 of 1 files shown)


bcoca:
so the v2_final branch is not used anymore, any patches would have to be against devel

PR #11269, opened 2015-06-15
local and ssh connection plugin fixes

sivel:
This PR addresses a few things:
1. On the `ssh` connection plugin, use `%` for printf formatting to replace `%s`
2. Make sure we accept kwarg `sudoable` for the `local` connection plugin and pass it when calling `exec_command` via `super`


bcoca:
hmm,, i think we were both working on the same things

sivel:
Yeah, it's no problem.  I think I didn't realize you were in process of changing things.  I'll let you close this, but can you check to see that you have fixed these 2 small things?

bcoca:
I fixed both, but only one was merged in, the other one will be soon

PR #11270, opened 2015-06-15
Add plugin that profiles playbook tasks

aioue:
[Description and console output demonstration](https://github.com/aioue/ansible-plugin-profile/blob/master/README.md#features).

Provides per-task timing, ongoing playbook elapsed time and ordered list of top 20 longest running tasks at end.


bcoca:
FYI, link does not work

aioue:
Fixed about a minute after submission :+1:, you're fast!

msabramo:
Would love to see this in ansible.

bcoca:
this seems almost the same as the already included timer.py callback

msabramo:
Oh hmm don't know if this was added since I last looked in plugins or if I never noticed.

https://github.com/ansible/ansible/blob/devel/plugins/callbacks/timer.py

I do wish for a way to use bundled plugins without copying them into my project.

aioue:
referenced `timer.py` just times the entire play, imho fairly redundant as you can just use the inbuilt Linux `time` command. The submitted version gives task times as you go, a sorted list of the longest tasks (hence 'profile'), and a timestamp of when it completed.

The link is fixed now @bcoca if you'd like to see an output example.

msabramo:
Ah right, the plugin in this PR has more features than the the `timer.py` example one we referenced.

bcoca:
I updated the timer module to work with v2. @aioue can you modify this module to do the same?

aioue:
@bcoca no problem, do you have a link to your patch please? `timer.py` isn't in `devel` branch anymore and the `v2_final` doesn't appear to have had any changes for quite a while.

bcoca:
I moved it to the 'active callbacks' as we also added a whitelist configuration that allows to add callbacks into execution w/o having to copy them:

lib/ansible/plugins/callback/timer.py

This is the new location in current devel, I'll be moving the rest of the callbacks from plugin/callbacks also, once migrated to v2 syntax.

aioue:
this is beyond my ability to update - not sure what the requirements are. Tempted to wait for v2, look at the module there and resubmit.

bcoca:
just change to the new path and I'll take care of modifications

aioue:
@bcoca resubmitted without commit noise, and to the path you requested https://github.com/ansible/ansible/pull/11625

bcoca:
closing this in favor of new pr

Issue #10960, opened 2015-05-08
UnicodeDecodeError when using ec2.py swallows real error from remote hosts


nickstenning:
##### Issue Type:

Bug Report
##### Ansible Version:

```
$ ansible --version
ansible 1.9.1
  configured module search path = None
```
##### Ansible Configuration:

This is vanilla ansible installed from the `ansible/ansible` PPA:

```
$ apt-cache show ansible
Package: ansible
Priority: optional
Section: admin
Installed-Size: 4189
Maintainer: Michael DeHaan <michael.dehaan@gmail.com>
Architecture: all
Version: 1.9.1-1ppa~trusty
Depends: python, python-support (>= 0.90), python-jinja2, python-yaml, python-paramiko, python-httplib2, python-crypto (>= 2.6), sshpass
Filename: pool/main/a/ansible/ansible_1.9.1-1ppa~trusty_all.deb
Size: 623674
MD5sum: 5fe93ff14c633238b8b11f8dab9017c1
SHA1: 7bac5506f8e453195a9efdc29dabaaf60f3b4d7e
SHA256: 006717c54311d131d0270570b8fb92624f5deba4af477d42826e1318c431b022
Description-en: A radically simple IT automation platform
 A radically simple IT automation platform that makes your applications and
 systems easier to deploy. Avoid writing scripts or custom code to deploy and
 update your applications— automate in a language that approaches plain English,
 using SSH, with no agents to install on remote systems.
Description-md5: ddfdff3530378f98f78630ce2aed5fe8
```

And I haven't changed any configuration files:

```
$ debsums -ac ansible
[NO OUTPUT]
```
##### Environment:

Ubuntu 14.04.2:

```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 14.04.2 LTS
Release:    14.04
Codename:   trusty
```

If it helps, I'm using AMI `ami-cc3b3ea4` in us-east-1.

The locale appears to be correctly configure both on the machine running ansible and the client hosts:

```
$ env | egrep 'LC|LANG'
LC_ALL=en_US.UTF-8
LANG=en_US.UTF-8
$ ssh 10.1.1.47 "env | egrep 'LC|LANG'"
LC_ALL=en_US.UTF-8
LANG=en_US.UTF-8
```
##### Summary:

[...]

PR #11276, opened 2015-06-16
support utf-8 error messages

rncry:
By no means the only solution.. but it works. And fixes bugs reporting errors.

Fixes this: https://github.com/ansible/ansible/issues/10960#issuecomment-112344083

### DIFF
--- a/v1/ansible/callbacks.py
+++ b/v1/ansible/callbacks.py
@@ -17,6 +17,8 @@
 
 import utils
 import sys
+reload(sys)
+sys.setdefaultencoding("utf-8")
 import getpass
 import os
 import subprocess


abadger:
`sys.setdefaultencoding()` should never be called outside of python's own initialization (in site.py) so this is unfortunately not the way to fix this issue.  I've been meaning to write up a blog post about sys.setdefaultencoding for a while... Decided to finish it off this morning: https://anonbadger.wordpress.com/2015/06/16/why-sys-setdefaultencoding-will-break-code/

# Question About This One

Hi!

- Even though sys.setdefaultencoding() seems like an attractive way to fix encoding issues, it really just adds the potential for even worse problems.  We'll need to look at being explicit about transforming between unicode type and str type as the real fix for this.

PR #11281, opened 2015-06-16
Bugfix: win_checksum

trondhindenes:
Win checksum had some errors causing it to output an array of strings instead of a string. This caused depending module such as win_file and win_copy to throw errors.

### DIFF
--- a/v1/ansible/module_utils/powershell.ps1
+++ b/v1/ansible/module_utils/powershell.ps1
@@ -151,7 +151,7 @@ Function Get-FileChecksum($path)
     {
         $sp = new-object -TypeName System.Security.Cryptography.SHA1CryptoServiceProvider;
         $fp = [System.IO.File]::Open($path, [System.IO.Filemode]::Open, [System.IO.FileAccess]::Read);
-        [System.BitConverter]::ToString($sp.ComputeHash($fp)).Replace("-", "").ToLower();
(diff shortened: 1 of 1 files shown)


trondhindenes:
Ref discussion arround issue 1404: https://github.com/ansible/ansible-modules-core/issues/1404

bcoca:
so this is just backporting it to the v1 in devel? we would actually need to make sure its in stable-1.9 for the 1.9.2 release

bcoca:
cc @jhawkesworth  @cchurch

trondhindenes:
Well, v1 is the only working version for WinRM usage at the moment. I plan to do a separate PR that implements feature parity between powershell.ps1 in v1 and v2 but I wanted to get this in first since @jhawkesworth has a depending change (related to win_copy)

trondhindenes:
But yes, it's essentially a backport.

jhawkesworth:
@bcoca @trondhindenes 
It sounds like v1 is now 'parked' i.e. no longer updated, for reference only. Is that right @bcoca ?

If there is going to be a 1.9.2 release then I'd hope that https://github.com/ansible/ansible/pull/11115 can be cherry-picked into the stable-1.9 branch, along with https://github.com/ansible/ansible-modules-core/pull/1438.  I think that should provide windows users with a workable copy/file/template implementation until such time as the winrm connection is ported to v2.

bcoca:
confirming those made it into stable-1.9, merging this to allow for v1/hacking/env-setup use

PR #11284, opened 2015-06-16
flake8ify plugins/inventory/ssh_config.py

msabramo:
Fix a few minor flake8 warnings.

Because I was considering vendoring this into another repo, but it causes the flake8 CI job to fail.


bcoca:
we don't take style updates as they break attribution

PR #11290, opened 2015-06-16
Refactored AIXNetwork.get_interfaces_info

### LINKED ISSUE #11289: AIX WPAR handling is inefficient and possibly buggy
The `lib/ansible/module_utils/facts.py:AIXNetwork.get_interfaces_info` method is structured like:

```
rc, out, err = module.run_command([ifconfig_path, '-a'])
for line in out.split('\n'):
    ...
    uname_path = module.get_bin_path('uname')
    if uname_path:
        rc, out, err = module.run_command([uname_path, '-W'])
```

This has two big problems:
1. It uses `subprocess.Popen` to run `uname -W` _for each line in ifconfig_. This is a performance issue because shelling out is expensive and this data isn't going to change during run on a single host. It should be collected once and then reused.
2. Note that `out` is overwritten inside the loop. It won't break the loop, which is iterating over `out.split('\n')` and not `out` directly. However, that may cause breakage in further edits that expect `out` to keep the same information throughout the loop. Consider using unique variable names like `ifconfig_out`,`entstat_out`, etc. to avoid weird, hard-to-debug errors later.

/cc @bmanojlovic who contributed this code.

kstrauser:
- Moved parse_entstat and parse_lsattr into their own methods, and
  simplified them.
- Collect get_bin_path outputs once per call to get_interfaces_info
  instead of once per line of ifconfig output.
- Execute "uname -W" once per call to get_interfaces_info.

Fixes #11289.

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -2239,6 +2239,42 @@ class AIXNetwork(GenericBsdIfconfigNetwork, Network):
     """
     platform = 'AIX'
 
+    def parse_entstat(self, entstat_path, current_if):
+        """Parse the output of the entstat command for this interface"""
+
+        if not current_if['device'].startswith('en'):
+            return
+
+        rc, out, err = module.run_command([entstat_path, current_if['device'] ])
+        if rc:
+            return
+
+        for line in out.split('\n'):
+            if not line:
+                continue
+
+            buff = re.match('^Hardware Address: (.*)', line)
+            if buff:
+                current_if['macaddress'] = buff.group(1)
+
+            if line.startswith('Device Type:') and 'Ethernet' in line:
+                current_if['type'] = 'ether'
+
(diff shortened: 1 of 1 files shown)


amenonsen:
@kstrauser any chance you could rebase this to apply to current devel?

kstrauser:
@amenonsen I just pushed the rebase.

bmanojlovic:
unfortunately neither do i have direct access to AIX machine, will look if maybe we have one still up.
But i agree that code is very inefficient. I was following code style of previous contributor (but mess with uname is mine because it was rather far inside of loop i did not look into possible speedup)

jctanner:
Deprecated by https://github.com/ansible/ansible/pull/18288

kstrauser:
I wish that #18288 was formatted more like this, instead of keeping the deeply nested loops and conditionals structure. Oh well - next person who has to deal with AIX networking can make it look pretty. 🙂

PR #11294, opened 2015-06-17
plugins/inventory/serf.py: Use SERF_RPC_* env vars

msabramo:
This makes the Serf inventory plugin use the `SERF_RPC_ADDR` and `SERF_RPC_AUTH` environment variables that the `serf` command-line tool already uses (see https://www.serfdom.io/docs/commands/members.html#_rpc_addr).

These can be used to get Serf data from a remote node instead of requiring the ansible control host to be running a serf agent and to be a member of the serf cluster. E.g.:

```
$ SERF_RPC_ADDR=serf.company.com:7373 SERF_RPC_AUTH="secret" /Users/marca/dev/git-repos/ansible/plugins/inventory/serf.py --list
{"serf": ["mt5-pyweb01", "mt3-pyweb01", "mt2-pyweb02", "mt1-pyweb01", ...
```

Cc: @bcoca, @KushalP, @sudarkoff

### DIFF
--- a/plugins/inventory/serf.py
+++ b/plugins/inventory/serf.py
@@ -20,10 +20,18 @@
 # Dynamic inventory script which lets you use nodes discovered by Serf
 # (https://serfdom.io/).
 #
-# Requires host to be a member of a Serf cluster and the `serfclient` Python
-# module from https://pypi.python.org/pypi/serfclient
+# Requires the `serfclient` Python module from
+# https://pypi.python.org/pypi/serfclient
+#
+# Environment variables
+# ---------------------
+#   - `SERF_RPC_ADDR`
+#   - `SERF_RPC_AUTH`
+#
+# These variables are described at https://www.serfdom.io/docs/commands/members.html#_rpc_addr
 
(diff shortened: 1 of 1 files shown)


plugins/inventory/serf.py
```diff
@@ -37,9 +45,22 @@
...
+
+    rpc_addr = os.getenv('SERF_RPC_ADDR')
+    if rpc_addr:
+        kwargs['host'], kwargs['port'] = rpc_addr.split(':')
+
+    rpc_auth = os.getenv('SERF_RPC_AUTH')
+    if rpc_auth:
+        kwargs['rpc_auth'] = rpc_auth
+
+    return SerfClient(**kwargs)
```
KushalP, line 59: Once https://github.com/KushalP/serfclient-py/pull/17 is cleaned up and merged, you can simplify this method to be:

``` python
def _serf_client():
    environment_config = EnvironmentConfig()
    return SerfClient(host=env.host, port=env.port, rpc_auth=env.auth_key)
```

PR #11299, opened 2015-06-17
support instances with no public ip

rncry:
default to private ip if the instance doesn't have a public ip assigned. (causes list index out of range error otherwise)


rncry:
really you probably want to allow a toggle to switch between using the private or public ip as the ansible hostname.. not got time to code that up though, this is a quick fix.

bcoca:
cc @erjohnso

erjohnso:
lgtm, thanks @rncry.

PR #11302, opened 2015-06-17
Easier use of built-in dynamic inventory scripts

msabramo:
I'm not sure this is the best way. Consider it one possible solution to a problem and a jumping off point for a discussion.

Problem: I want to make it easier to have ansible consult the dynamic inventory plugins that come with ansible -- e.g.: `ssh_config`, `nova`, `serf`, etc. I could only think of a few ways of doing this now and I don't love any of them:
1. Tell people to pass `-i /path/to/ansible/plugins/inventory/ssh_config.py` to `ansible`, `ansible-playbook`, etc. Bleh.
2. Add `/path/to/ansible/plugins/inventory/ssh_config.py` to an `ansible.cfg`; not that I know where people have installed ansible. They may all have it installed in different places. Bleh.
3. Copy the contents of the scripts I need into a directory in my project that lives alongside the `ansible.cfg`, static inventory files, playbooks, etc. -- e.g.: `hosts/nonprod/dynamic/ssh_config.py`. This is the best solution but not great, because I'm copying code, not getting the latest changes in Ansible, etc.

So here's a stab at solving this...

---

Allows passing the name of a built-in inventory script with `-i` on the command line or in `ansible.cfg` without specifying the full path. If Ansible cannot resolve the given path, it will attempt to find resolve it relative to the `ansible/plugins/inventory` directory.

This makes it easier to use these scripts without having to copy them into your project.

E.g.:

```
$ ansible gabriel -m ping -i ssh_config.py
gabriel | SUCCESS => {
    "invocation": {
        "module_name": "ping",
        "module_args": {}
    },
    "changed": false,
    "ping": "pong"
}
```

The `.py` suffix can be omitted as well:

```
$ ansible gabriel -m ping -i ssh_config
gabriel | SUCCESS => {
    "invocation": {
        "module_name": "ping",
        "module_args": {}
    },
    "changed": false,
    "ping": "pong"
}
```

and instead of using `-i` on the command-line, you could use the same syntax in an `ansible.cfg` file -- e.g.:

```
# ansible.cfg

[defaults]
inventory = ssh_config
```

and then you can simply do:

```
$ ansible gabriel -m ping
gabriel | SUCCESS => {
    "invocation": {
        "module_name": "ping",
        "module_args": {}
    },
    "changed": false,
    "ping": "pong"
}
```

Cc: @sudarkoff, @sontek


sivel:
So this implementation would not work for a release tarball as the inventory scripts are not included, so it sounds like you would require someone to copy a non release file into the release anyway.

Honestly, I think the appropriate way should be to vendor the plugin in your own project, and then it is a known path that can easily be controlled.

msabramo:
It seems that the inventory scripts **are** included in tarballs?

```
$ tar tvzf ansible-1.9.1.tar.gz | grep ssh_config
-rwxrwxr-x  0 jenkins jenkins  3198 Apr 28 11:42 ansible-1.9.1/plugins/inventory/ssh_config.py
```

although they don't seem to get installed???

```
[marca@marca-mac2 ansible]$ cd /Users/marca/python/virtualenvs/smstack/lib/python2.7/site-packages/ansible/plugins/inventory
[marca@marca-mac2 inventory]$ ls -l *.py
-rw-r--r--+ 1 marca  staff  2712 Jun 17 13:37 __init__.py
-rw-r--r--+ 1 marca  staff  2266 Jun 17 13:37 aggregate.py
-rw-r--r--+ 1 marca  staff  1991 Jun 17 13:37 directory.py
-rw-r--r--+ 1 marca  staff  2220 Jun 17 13:37 ini.py
```

It's confusing because there are multiple inventory plugins dirs and some get installed and some don't (even though they're all in the tarball):

```
[marca@marca-mac2 ansible]$ tar tvzf dist/ansible-2.0.0.tar.gz| egrep 'directory|aggregate|ssh_config'
-rw-r--r--  0 marca  staff    2266 Jun 16 09:08 ansible-2.0.0/lib/ansible/plugins/inventory/aggregate.py
-rw-r--r--  0 marca  staff    1991 Jun 16 09:08 ansible-2.0.0/lib/ansible/plugins/inventory/directory.py
-rwxr-xr-x  0 marca  staff    3198 Jun 16 23:48 ansible-2.0.0/plugins/inventory/ssh_config.py
-rw-r--r--  0 marca  staff    3088 Jun 17 11:11 ansible-2.0.0/plugins/inventory/ssh_config.pyc
```

sivel:
So perhaps they are included, but yes they are not installed.  The ones that you see installed are core inventory scripts for handling the default ini and execution of dynamic inventories and such.  The only thing installed as part of ansible, are those files that live inside of the `lib` directory.

The top level directory called `plugins` was once described to me as being "examples" and I have always thought of them and treated them as such.  Great examples, often working out of the box, but best vendored in your own repo and potentially manipulated to fit your individual needs.

As such, I still believe this patch to be insufficient and invalid.

msabramo:
Agreed, yes they are included in the tarball but not installed.

One way to solve this is cbe038e, which moves the inventory plugins into the `lib` tree so that they do get installed.

Maybe some of those plugins are non-functional examples, but I think a lot of them work perfectly out of the box, like the one that I am interested in here -- `ssh_config.py`. That one is pretty simple and works great so I don't see any reason not to include it a built-in feature of ansible.

I'm vendoring that plugin for now, but I'd rather not, as it is causing CI jobs to fail for my project because that code is not covered by tests and is not flake8 compliant. Now I could work around these issues by setting things up so that stuff is exempted, but I'd rather avoid that if I can.

That said maybe the ansible maintainers don't want to "officially" support the inventory plugins. That's their call.

sivel:
Part of vendoring and manipulating to suit your needs can involve modifying to meet your flake8 compliance.  Nothing saying you can't do that.  Or perhaps exclude that file from your CI checks to not validate that file for compliance.

msabramo:
Yeah. That's what I do now. I vendor in a plugin or two, but I'd prefer not to. I'd rather not maintain a list of exclusions or have to merge upstream changes with local mods. 

I could see where the core team might not want to support all of the inventory plugins because some might require specialized knowledge and/or infrastructure but I'm wondering if there are a few that could be made core, like ssh_config which is very simple and strikes me as a great ansible core feature.

bcoca:
so I was thinking of making these scripts and configs into python resources so they get included in the distribution packages, also to rename them to contrib/inventory_scripts as they are not really inventory plugins, that is script.py which enables inventory scripts.

amenonsen:
I am slightly surprised to hear that the inventory scripts don't get installed by the package (I've always just run from a git checkout), but I treat them as being something to include with my playbook, so it doesn't bother me. I'm not sure how the situation would change with the strategy @bcoca proposed above. In particular, is it a solution to "they don't get installed" or a solution to "I don't know where it's installed", i.e. the original problem? If I understand correctly—and I may be missing something—it's not going to make `ansible -i ssh_config` work.

bcoca:
@msabramo I move the inventory scripts to contrib/inventory and they are now referenced in setup tools and will be distributed with ansible.

On making them easier to use I was thinking of integrating the ansible-inventory cli to allow for easy install/usage of them (aside from it's current functions of displaying inventory data).

bcoca:
I'm closing this as it was 'fixed' by changes above, also looking to future inventory revamp in which we allow actual plugins (script being one and keeping current inventory script compatibility).

PR #11305, opened 2015-06-17
allow multiple vCenter/ESXi servers in single inventory

aioue:
Allow multiple vCenter/ESXi servers to be scanned from single inventory/ini combo, removing the need to have multiple sets of vmware inventory files when managing a large deployment.


aioue:
Closing. Re-submitted in #12492

PR #11315, opened 2015-06-18
Adding lookup plugin for postgresql database

gshivani:
##### SUMMARY
lookup plugin for querying postgresql

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
lib/ansible/plugins/lookup/postgresql.py

##### ANSIBLE VERSION
2.3

### DIFF
--- a/lib/ansible/plugins/lookup/postgresql.py
+++ b/lib/ansible/plugins/lookup/postgresql.py
@@ -1,0 +1,76 @@
+# 2014, Shivani Gowrishankar <s.gowrishankar@ntoggle.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import psycopg2
+
+from ansible.errors import AnsibleError
+from ansible.plugins.lookup import LookupBase
+
+class LookupModule(LookupBase):
+    """
+
+    postgresql module used to query tables in postgresql database.
+
+    Example: lookup('postgresql','database=database user=user password=password sql=SQL query')
+
+    """
+    def run(self, terms, variables, **kwargs):
+
+        ret = []
+        if not isinstance(terms, list):
+            terms = [ terms ]
+        #splitting only till 3 parameters because sql query has ' ' inside
+        for term in terms:
+            params = term.split(' ',3)
+
+        paramvals = {
+            'database' : 'database',
+            'user' : 'user',
+            'password' : 'password',
+            'sql' : 'query',
+        }
+        try:
+            for param in params:
+                name, value = param.split('=')
+                assert(name in paramvals)
+                paramvals[name] = value
+        except (ValueError, AssertionError) as e:
+            raise AnsibleError(e)
+        #assigning args for postgresql connect 
+        database = paramvals['database']
+        user = paramvals['user']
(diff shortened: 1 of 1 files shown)


msabramo:
Maybe even more useful would be a more general plugin that uses SQLAlchemy to query the database. Then you could use it with any of the database engines supported by SQLAlchemy.

http://docs.sqlalchemy.org/en/rel_1_0/dialects/index.html 

Instead of taking username, password, etc., it could take a SQLAlchemy database URL (which probably a lot of people would keep in variables and encrypt with vault).

amenonsen:
A few quick thoughts: (a) this definitely needs documentation, (b) "connection failed or something's wrong" is a complete non-starter, this is something that may well not work right the first time, and that error would drive me nuts while debugging and not help _at all_.

@msabramo's suggestion to use SQLAlchemy is… well, it's probably the Right Thing To Do™, though I personally don't want to have to install and learn SQLAlchemy and don't need anything but Postgres support. Being able to encrypt and store database connection information is certainly nice, but it's also possible to encrypt and store the `conninfo` string for Postgres with this PR already.

lib/ansible/plugins/lookup/postgresql.py
```diff
@@ -0,0 +1,76 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import psycopg2
```
bcoca, line 20: this needs to be guarded and return an ansible error in the init  (see consul_kv lookup and HAS_CONSUL)

lib/ansible/plugins/lookup/postgresql.py
```diff
@@ -0,0 +1,76 @@
...
+        try:
+            con = psycopg2.connect(database=database, user=user, password=password)
+            cur = con.cursor()
+            cur.execute(sql)
+            rows = cur.fetchall()
+            for row in rows:
+                row_string = str(row)
+                ret.append(row_string)
+        except:
+             raise AnsibleError("Connection Failed or something's wrong")
```
bcoca, line 71: Adding the exception string to the error would help the user narrow down what the issue is.

lib/ansible/plugins/lookup/postgresql.py
```diff
@@ -0,0 +1,76 @@
...
+                assert(name in paramvals)
+                paramvals[name] = value
+        except (ValueError, AssertionError) as e:
+            raise AnsibleError(e)
+        #assigning args for postgresql connect 
+        database = paramvals['database']
+        user = paramvals['user']
+        password = paramvals['password']
+        sql = paramvals['sql']
+        #Connection to postgresql
```
bcoca, line 60: lacking parameter validation, user should be notified if the minimal requirements for lookup to work are not met

bcoca (changes requested):
I know this has been open a long time, but since previous comments were not addressed we have not returned to it. Now we are trying to close these old issues and if we don't get any response to this review either we'll end up just closing the PR.

jstoja:
@bcoca No answer after +1y, you can probably close it.

ryanmerolle:
What is required here to make this shippable?  I am interested in this becoming a standard lookup plugin in ansible.

lib/ansible/plugins/lookup/postgresql.py
```diff
@@ -0,0 +1,76 @@
...
+        #splitting only till 3 parameters because sql query has ' ' inside
+        for term in terms:
+            params = term.split(' ',3)
+
+        paramvals = {
+            'database' : 'database',
+            'user' : 'user',
+            'password' : 'password',
+            'sql' : 'query',
+        }
```
ryanmerolle, line 47: It likely makes sense to add the remaining psycopg2 parameters including host and port.

gundalow:
@ryanmerolle PR needs rebasing https://docs.ansible.com/ansible/devel/dev_guide/developing_rebasing.html

It maybe worth you creating a new branch and copying across the new file into that branch

gshivani:
@ryanmerolle Glad that you are interested in this becoming a standard plugin! I'm not sure if I can get to this soon. Please feel free to create new branch and reuse the code!

Andersson007:
needs_info

PR #11323, opened 2015-06-19
Add a new "on_exception" callback method.

ralphbean:
There are two separate commits.  One adds the feature to ansible v2 and the other to v1.

This can be used to catch and log a 'KeyboardInterrupt' when a user cancels a playbook run.

We use callback plugins to log the start and stop of our playbook runs to another system ([fedmsg](http://fedmsg.com) for [Fedora Infrastructure](https://fedoraproject.org/wiki/Infrastructure)).  I sat down to try and build a dashboard showing what admins were running what playbooks at any given time.  We have logs indicating when playbooks start and stop, but _not_ when they are interrupted by control-c or some other internal exception.  Adding this callback would let us expose that kind of information.

### DIFF
--- a/lib/ansible/executor/playbook_executor.py
+++ b/lib/ansible/executor/playbook_executor.py
@@ -145,6 +145,12 @@ class PlaybookExecutor:
             if entrylist:
                 return entrylist
 
+        except Exception as e:
+            # Notify callback plugins of catastrophe
+            if self._tqm is not None:
+                self._tqm.send_callback('v2_playbook_on_exception', e)
+            # And re-raise our exception in its original context
+            raise
         finally:
             if self._tqm is not None:
                 self._cleanup()
--- a/lib/ansible/plugins/callback/__init__.py
+++ b/lib/ansible/plugins/callback/__init__.py
@@ -100,3 +100,5 @@ class CallbackBase:
     def playbook_on_stats(self, stats):
         pass
 
+    def playbook_on_exception(self, exception):
+        pass
--- a/lib/ansible/plugins/callback/default.py
+++ b/lib/ansible/plugins/callback/default.py
@@ -134,3 +134,5 @@ class CallbackModule(CallbackBase):
     def v2_playbook_on_stats(self, stats):
         pass
 
+    def v2_playbook_on_exception(self, exception):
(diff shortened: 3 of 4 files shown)


bcoca:
we really are not adding features to v1/ dirs at this point

bcoca:
I'll bring this up with the rest of core team as we were already capture, handle exceptions and have callback code that treats them differently, we still wanted to expand on that.

Also, we already capture the user interrupt in the ansible cli tools

ralphbean:
> we really are not adding features to v1/ dirs at this point

I'll rebase to remove commit  567e71b, then.

> we were already capture, handle exceptions and have callback code that treats them differently, we still wanted to expand on that.

That's cool.  If there's a way for users to hook into that with callbacks, then great.  I just didn't see a way when I looked through the code myself.

> Also, we already capture the user interrupt in the ansible cli tools

Right, but there's no way to use that user interrupt to do anything (like trigger a callback plugin).  Note that my patch here re-raises the user interrupt in its original context, so that catch in the cli tools will still work as it originally did.

amenonsen:
I have absolutely no use for this PR, but it's so cool that I want it anyway. Wanting to log interruptions in a security-conscious environment is as good an excuse as any.

I was actually hoping that the `KeyboardInterrupt` stuff was somehow related to the `pause` module being broken (see https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/action/pause.py#L96), but I guess not.

bcoca:
so an exception at that level should already call on_failed, which manages showing traceback data or not depending on verbosity, I'm not sure that we need this on top of that.

ralphbean:
I think the difference is that `runner_on_failed` (what you're referencing) is called at the task-level, whereas the `playbook_on_exception` is called at the playbook level.  You can currently write plugins to perform actions when tasks fail but not when the whole playbook exits prematurely.

abadger:
@bcoca asks why you can't notify outside of the process -- ie: ansible-playbook foo.yml || fedmsg-notify

@jimi-c thought that if there was a reason, perhaps always firing on_stats (even when we exit for a fatal execption) might be a better method than a new on_exception callback.

PR #11333, opened 2015-06-21
quotes removed from become_method

francescor:
just experimented that with quotes it does not work (ansible 1.9.1)

### DIFF
--- a/examples/ansible.cfg
+++ b/examples/ansible.cfg
@@ -161,7 +161,7 @@ fact_caching = memory
 
 [privilege_escalation]
 #become=True
-#become_method='sudo'
+#become_method=sudo
 #become_user='root'
 #become_ask_pass=False
 


bcoca:
what does not work? 

the quotes should not affect the method working or not.

francescor:
but it does, at least in my installation (fedora 21 + ansible 1.9.1)

By having the following in  /etc/ansible/ansible.cfg

``` bash
become_method='su'
become_ask_pass=True
```

I get

```
ansible -a 'whoami' example.com
'SU' password: 
example.com | FAILED => Internal Error: this module does not support running commands via 'su'
```

while everything works fine with

```
become_method=su
```

```
$ ansible -a 'whoami' example.com
SU password: 
example.com | success | rc=0 >>
root
```

(note now there are no quotes around SU)

btw: with a playbook I get an error, too: and again only if I use the quote:

```
fatal: [example.com] => Internal Error: this module does not support running commands via 'su'
```

bcoca:
updated config file to remove bad examples

abadger:
@francescor  I've just pushed the following commit to devel: https://github.com/ansible/ansible/commit/5b0b1f8da6d713410037584679ebe99a0ce099f7

That should fix unquote string values that are entered into the config file.

abadger:
# Closing This Ticket

Hi!

issue is resolved.

Thank you!

PR #11334, opened 2015-06-21
Update playbooks_tags.rst

dhoffutt:
When specifying multiple tags from the command line, one must use a space and not a comma between tags.

### DIFF
--- a/docsite/rst/playbooks_tags.rst
+++ b/docsite/rst/playbooks_tags.rst
@@ -23,7 +23,7 @@ Example::
 
 If you wanted to just run the "configuration" and "packages" part of a very long playbook, you could do this::
 
-    ansible-playbook example.yml --tags "configuration,packages"
+    ansible-playbook example.yml --tags "configuration packages"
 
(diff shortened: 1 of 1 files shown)


aioue:
This works fine:
`ansible-playbook foo.yml -i inventory/foo/ --tags="foo,bar"`

Have you tested using the '=' version?

dhoffutt:
But the version using "=" is not what is in the documentation. My goal was to fix the documentation.

sivel:
I cannot duplicate this issue on devel.  Running with `--tags "one,two"` correctly parses the tags out to:

```
set(['one', 'two'])
```

In fact, using the recommended change in this PR, causes an incorrect tag parsing that treats `one two` as a single tag.

bcoca:
I cannot reproduce the reported error and testing the fix shows it actually breaks the intended functionality.

@dhoffutt, unless you can supply more data on how to reproduce I'm inclined to close this w/o merging.

amenonsen:
As noted by others, I can't reproduce this either. This PR can be closed now.

PR #11336, opened 2015-06-21
Update Vault rekey() in vault.py to validate files passed before input

saada:
…pting for password

Update rekey() in vault.py to validate files passed exist before prompting for password.
Why enter a password when the files passed don't exist in the first place?


amenonsen:
Looks fine, though I would have preferred to see this squashed into a single commit rather than have typos fixed one by one. @saada, could you please squash and update this PR with `git push -f`?

saada:
@amenonsen, there you go, thanks!

PR #11350, opened 2015-06-22
Add example callback plugin: no_hosts_matched_fail

msabramo:
This example callback plugin makes `ansible-playbook` return a non-zero exit code if the playbook matches no hosts.

A stab at a solution to
https://groups.google.com/forum/#!topic/ansible-project/9QlUer83zA0

E.g.:

```
[marca@marca-mac2 test]$ cat foo.yml

---
- hosts: a-host-that-does-not-exist
  tasks:
    - debug: var=ansible_env

[marca@marca-mac2 test]$ ansible-playbook foo.yml
|---> plugin activated: /private/tmp/test/plugins/no_hosts_matched_fail.pyc

PLAY [a-host-that-does-not-exist] *********************************************
skipping: no hosts matched
ERROR: FATAL: no hosts matched or all hosts have already failed -- aborting


[marca@marca-mac2 test]$ echo $?
1
```

Cc: @sudarkoff, @sontek, @djeebus

### DIFF
--- a/plugins/callbacks/no_hosts_matched_fail.py
+++ b/plugins/callbacks/no_hosts_matched_fail.py
@@ -1,0 +1,11 @@
+# plugins/no_hosts_matched_fail.py
+
+from ansible.errors import AnsibleError
+
+
+class CallbackModule(object):
+    def playbook_on_start(self):
+        print('|---> plugin activated: %s' % __file__)
+
+    def playbook_on_no_hosts_matched(self):
+        raise AnsibleError("FATAL: no hosts matched or all hosts have already failed -- aborting\n")


bcoca:
First of all, NO to the idea of having exceptions in the callbacks, it is not the correct place for them, if at all, this should go in the part of executor or tqm that actually emits the callback (right after doing so). 

Second, this should be configurable to be considered at all, it will break currently expected playbook behavior, empty play gets skipped but subsequent ones should attempt to run.

msabramo:
Ok, I can update this to not use exceptions. 

How to make it configurable? Environment variable? Can you give a short example of something this would break so that I can test and make sure it's not broken when configured appropriately?

bcoca:
a normal configuration in lib/ansible/constants.py, can be covered by several options, env, ansible.cfg, etc

not sure what breakage you are looking for

amenonsen:
I'm a bit confused by this PR.

First, the use case seems a bit niche to be included by default, nor does the implementation seem especially useful as an example. Besides, I don't understand what exactly @bcoca wanted configurable here. If you don't want this behaviour, don't enable the callback plugin, right? The comment about «it will break currently expected playbook behavior, empty play gets skipped but subsequent ones should attempt to run» is especially puzzling, and makes me think it was based on a misunderstanding about what the PR seeks to do.

So some clarification is needed here from someone who isn't me.

bcoca:
the PR raises an exception from a callback, interrupting all further processing, it mimics the fail on any error functions.

callbacks are not the proper place for exceptions, i was included to allow this as a config option but it is starting to look very niche so I'm going to close this for now, if people start asking for this feature we might reconsider in the future.

PR #11357, opened 2015-06-23
Fix broken links on the "Developing Plugins" page

tersmitten:


bcoca:
this needs a rebase, also the links point to the wrong place, v1 is the version that is being deprecated and put there temporarily to give users of the git repo easy access to compare behavior, the docs should repoint to new locations under lib/ansible.

amenonsen:
This PR can be closed, the links have been fixed already.

tersmitten:
Indeed

PR #11360, opened 2015-06-23
Clarify that setting ssh_args trumps control_path

mrwacky42:


bcoca:
misspelled "ignored", though 'overridden by' might be more appropriate

mrwacky42:
Rebased, fixed embarrassing spelling.

bcoca:
he, no worries, I probably have caused the highest number of PRs to correct spelling

PR #11361, opened 2015-06-23
Update playbooks_tags.rst

wenottingham:
Clarify how tagging with roles and includes works.

### DIFF
--- a/docsite/rst/playbooks_tags.rst
+++ b/docsite/rst/playbooks_tags.rst
@@ -38,7 +38,7 @@ And you may also tag basic include statements::
 
     - include: foo.yml tags=web,foo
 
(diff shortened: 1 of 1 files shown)


bcoca:
well, they will run when the playbook is run w/o specifying tags also, maybe a better phrasing would be:

"Both of these have the function of applying the specified tags to all the tasks inside the include statement or role, so that the included tasks themselves can be selectively run when the playbook is invoked with the tags"

amenonsen:
How about "Both of these apply the specified tags to all the tasks inside the included file or role, so that they can be selectively run when the playbook is invoked with tags"?

amenonsen:
I've submitted a separate PR (#11722) with the suggested rewording. As a minor clarification, I think it's trivial enough to merge.

PR #11365, opened 2015-06-24
docs - added missing filters

halberom:
Several filters were undocumented - relpath, from_yaml, from_json, quote.  Tidied up code indentation, and standardized 'since version' comments.

### DIFF
--- a/docsite/rst/playbooks_filters.rst
+++ b/docsite/rst/playbooks_filters.rst
@@ -17,8 +17,26 @@ Filters For Formatting Data
 The following filters will take a data structure in a template and render it in a slightly different format.  These
 are occasionally useful for debugging::
 
+    {{ some_variable | to_json }}
+    {{ some_variable | to_yaml }}
+
+For human readable output, you can use::
+
     {{ some_variable | to_nice_json }}
     {{ some_variable | to_nice_yaml }}
+
+Alternatively, you may be reading in some already formatted data::
+
+    {{ some_variable | from_json }}
+    {{ some_variable | from_yaml }}
+
+for example::
+
+    tasks:
+      - shell: cat /some/path/to/file.json
+        register: result
+
+      - set_fact: myvar="{{ result.stdout | from_json }}"
 
 .. _filters_used_with_conditionals:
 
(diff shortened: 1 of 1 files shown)


docsite/rst/playbooks_filters.rst
```diff
@@ -300,13 +318,17 @@ Hash types available depend on the master system running ansible,
 Other Useful Filters
 --------------------
 
+To add quotes for shell usage::
+
+    - shell: echo={{ some_string_value | quote }}
```
bcoca: i think this was meant to be:

shell: echo {{string_value|quote}}

docsite/rst/playbooks_filters.rst
```diff
@@ -316,28 +338,32 @@ To get the directory from a path::
 
     {{ path | dirname }}
 
-To expand a path containing a tilde (`~`) character (new in version 1.5)::
+To expand a path containing a tilde (`~`) character (since 1.5)::
```
bcoca: we use 'new in version' in most other pages, I think we should keep it consistent

docsite/rst/playbooks_filters.rst
```diff
@@ -300,13 +318,17 @@ Hash types available depend on the master system running ansible,
...
 --------------------
 
+To add quotes for shell usage::
+
+    - shell: echo={{ some_string_value | quote }} 
+
 To use one value on true and another on false (since 1.9)::
 
-   {{ (name == "John") | ternary('Mr','Ms') }}
+    {{ (name == "John") | ternary('Mr','Ms') }}
```
bcoca: not sure why there are a bunch of space updates?

halberom: most of the code indents were 4 spaces, some were 3.  can remove if not necessary/desired.

docsite/rst/playbooks_filters.rst
halberom: ack'd, I wasn't sure, as there were some with 'since' and some with 'new in', so i went with the shorter option. will update.

halberom:
right, one day i'll get there, updated, hopefully better.

bcoca:
just making a note, we plan on adding inline docs to more plugins types, like what we have in task plugins/modules so we can auto-generate docs for filters/lookups/etc.

PR #11370, opened 2015-06-24
added role_path to magic var section

halberom:
Discovered that it's possible to use role_path today, it's not mentioned in the docs, so here's a brief comment.


tersmitten:
Nice to know

PR #11372, opened 2015-06-24
added six to install-from-source docs

nitzmahone:
v2 needs six installed- updated install-from-source docs to reflect that.


abadger:
Thanks for the docs fix!

Issue #9984, opened 2015-01-12
AWS credentials are not being used by ec2_tag


ISF:
##### ISSUE TYPE
Bug Report

##### COMPONENT NAME
ec2_tag module

##### ANSIBLE VERSION
```
ansible 1.8.2
configured module search path = None
ansible HEAD (as of 2015-01-12)
boto 2.35.1
```

##### OS / ENVIRONMENT
Running from a Archlinux system targeting remote AWS Linux instances.

##### SUMMARY
The AWS credentials from environment variables or boto configuration files are not  being passed to the ec2_tag module.

##### STEPS TO REPRODUCE
Minimum working example:
```
- name: Tag instances
  local_action: ec2_tag resource="{{ ec2_id }}" region=sa-east-1 state=present
  args:
    tags:
      App: "{{ App | mandatory }}"
      Environment: "{{ Environment | mandatory }}"
      Role: "{{ Role | mandatory }}"
  tags:
    - ec2_tags
```

##### EXPECTED RESULTS
Tags being updated/created on AWS EC2 instances

##### ACTUAL RESULTS
The ec_tag module is failing everytime with the following error:
```
msg: No handler was ready to authenticate. 1 handlers were checked. ['HmacAuthV4Handler'] Check your credentials
```
Which is caused when the credentials are not configured properly. However, I have a valid ${HOME}/.boto with the '[Credentials]' section correctly configured. The ec2 dynamic inventory script and connecting (and listing tags) with plain boto work fine. Passing credentials with environment variables don't work, the only workaround I found was specifying the key's values in the play, like this:
```
local_action: ec2_tag resource="{{ ec2_id }}" region=sa-east-1 state=present aws_access_key=****** aws_secret_key=******
```
IIRC this was working around May 2014, when I first made this role.
Edit: formatted the issue using the ansible's issue template (sorry for the delay).

ansibot:
Can You Help Us Out?
====================

Thanks for filing a ticket! I am the friendly GitHub Ansibot. 

It looks like you might not have filled out the issue description based on our standard issue template.  You might not have known about that, and that's ok too, we'll tell you how to do it.

We have a standard template because Ansible is a really busy project and it helps to have some standard information in each ticket, and GitHub doesn't yet provide a standard facility to do this like some other bug trackers.  We hope you understand as this is really valuable to us!. 

[...]

PR #11374, opened 2015-06-24
Get AWS access and secret keys from boto config

jason-gill:
##### SUMMARY
I noticed when I was trying to manage my AWS servers using the ec2 module I was getting the following error: 

```
msg: No handler was ready to authenticate. 1 handlers were checked. ['HmacAuthV4Handler'] Check your credentials
```

This is caused when the credentials are not configured properly. However, I have a valid ${HOME}/.boto with the '[Credentials]' section correctly configured. The ec2 dynamic inventory script and connecting  with plain boto work fine.

I googled the problem and found this issue: https://github.com/ansible/ansible/issues/9984 was having a similar problem.

I fixed the problem by adding a boto config lookup to the ec2 module utils.

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
lib/ansible/module_utils/ec2.py

##### ANSIBLE VERSION
2.3

### DIFF
--- a/lib/ansible/module_utils/ec2.py
+++ b/lib/ansible/module_utils/ec2.py
@@ -86,8 +86,9 @@ def get_aws_connection_info(module):
         elif 'EC2_ACCESS_KEY' in os.environ:
             access_key = os.environ['EC2_ACCESS_KEY']
         else:
-            # in case access_key came in as empty string
-            access_key = None
+            access_key = boto.config.get('Credentials', 'aws_access_key_id')
+            if not access_key:
+                access_key = boto.config.get('default', 'aws_access_key_id')
 
     if not secret_key:
         if 'AWS_SECRET_ACCESS_KEY' in os.environ:
@@ -97,8 +98,9 @@ def get_aws_connection_info(module):
         elif 'EC2_SECRET_KEY' in os.environ:
             secret_key = os.environ['EC2_SECRET_KEY']
         else:
-            # in case secret_key came in as empty string
-            secret_key = None
+            secret_key = boto.config.get('Credentials', 'aws_secret_access_key')
+            if not secret_key:
+                secret_key = boto.config.get('default', 'aws_secret_access_key')
 
     if not region:
         if 'AWS_REGION' in os.environ:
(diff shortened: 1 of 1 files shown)


amenonsen:
The change looks perfectly reasonable to me. I remember having a similar problem when I started using the ec2 modules, but I can't seem to reproduce it any more, so I haven't tested.

bnugent:
+1 for this. I feel dumb for assuming this was already the the case. Seems like a no-brainer.

stajkowski:
What about changing the logic a bit:

```
else:
    secret_key = boto.config.get('Credentials', 'aws_secret_access_key') if boto.config.get('Credentials', 'aws_secret_access_key') else boto.config.get('default', 'aws_secret_access_key')
    if not secret_key:
        secret_key = None
```

Pep8 the line length of course.  This would allow for setting to None if either of those options don't work.

Otherwise, looks good.

bcoca:
lenght is fine, the 80 column limit makes things less readable in my view, the trailing 'if' on the other hand can detract from readability so reformat that to multiple lines to make the reading of it clear.

stajkowski:
Always looking for a one-liner revival.  Agreed though, here's the multi-line.  I think if this were refactored for this logic, might put it in line with original intent.

```
else:
    if boto.config.get('Credentials', 'aws_secret_access_key'):
        secret_key = boto.config.get('Credentials', 'aws_secret_access_key')
    elif boto.config.get('default', 'aws_secret_access_key'):
        secret_key = boto.config.get('default', 'aws_secret_access_key')
    else:
        secret_key = None
```

preflightsiren:
+1

yoshiwaan:
I'm not sure if this is the place for it, but the standardised way to obtain aws credentials accross libraries and tools is from ~/.aws/credentials so that should be added too if it's not.

I just ran into the above handler error when calling ec2 modules from within a local_action section, but if I call the ec2 module directly (i.e. no local_action section) in a playbook that has hosts: localhost it works fine using the ~/.aws/credentials file. So I'm not sure if this is half implemented or if using hosts: localhost works because the logic of the boto library itself obtains the credentials.

windowsrefund:
Credentials are used from ~/.aws/credentials when using Ansible 2.0.1.0 and boto 2.38.0

However, strace confirms this file is never accessed when using bleeding edge

```
ansible 2.1.0 (devel 008bde91a8) last updated 2016/04/26 10:40:46 (GMT -400)
```

defionscode:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;AnzlAF27&#x2F;10-boto-credential"> boto credential</a>

mckerrj:
@s-hertel can you review and close or merge?

juliedavila:
@mckerrj needs to be rebased before it can be merged

s-hertel:
Hi @jason-gill. I've rebased this PR here: #24783 since there is no branch to push to for this pull request. I'll test it in the next couple days. Until everything has been ported to use boto3 this addition makes sense to me. It appears that after boto2 /etc/boto.cfg and ~/.boto are not used for credentials. Thanks for the PR!

ryansb:
Closing as this has been migrated to another PR, #24783

PR #11376, opened 2015-06-24
Updated fetch_url to actually use the parameters being passed in.

jason-gill:
I was trying to use the dnsmadeeasy module but the changes were not be reflected in DNS Made Easy and I noticed I was getting back some obscure error message. I tracked the issue down to the parameters being passed into the fetch_url method were not being used particularly the headers parameter.

### DIFF
--- a/lib/ansible/module_utils/urls.py
+++ b/lib/ansible/module_utils/urls.py
@@ -504,8 +504,8 @@ def fetch_url(module, url, data=None, headers=None, method=None,
     r = None
     info = dict(url=url)
     try:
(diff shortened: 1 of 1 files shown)


abadger:
Thanks!  I think I pushed pretty much an identical commit yesterday, though:

https://github.com/ansible/ansible/commit/a1a7d6c46247f313a8a9c2a1878e034324894c4b

Could you update your checkout and see if that change is available to you?

jason-gill:
Yes I can see you change and yes it is pretty much the same as mine.  It seems I had my git upstream repo set to my fork instead of the ansible repo. whoops.

PR #11389, opened 2015-06-25
Update link to github in developing plugins docs

silviot:


abadger:
Thanks for updating that link!

PR #11392, opened 2015-06-25
Store ec2 instance ids in tags and hostvars

jason-gill:
##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
ec2 dynamic inventory script

##### ANSIBLE VERSION
N/A

##### SUMMARY
I have several ec2 instances that I want to start and stop on a schedule to save money.  After some googling I came across these two conversations:

https://github.com/ansible/ansible/pull/6349
https://groups.google.com/forum/#!topic/ansible-project/OGkbqKB9sqA

Michael DeHaan suggested the ec2 inventory script would be the best place to implement this change.  Well I modified the ec2 inventory script to store the ec2 instance id of every instance such that it can be referenced in a playbook like this:

```
- hosts: itag_Environment_Development
  gather_facts: false
  tasks:
  - name: Stop EC2 instance(s)
    local_action: ec2
        state=stopped
        region="{{ hostvars[inventory_hostname]['ec2_region'] }}"
        instance_ids="{{ hostvars[inventory_hostname]['ec2_id'] }}"
```

```
- hosts: itag_Environment_Development
  gather_facts: false
  tasks:
  - name: Start EC2 instance(s)
    local_action: ec2
      state=running
      wait=yes
      wait_timeout=600
      region="{{ hostvars[inventory_hostname]['ec2_region'] }}"
      instance_ids="{{ hostvars[inventory_hostname]['ec2_id'] }}"
    register: ec2

  - name: Update DNS MadeEasy
    local_action: dnsmadeeasy
      account_key=******
      account_secret=*****
      domain={{ hostvars[item.id]['ec2_tag_Domain'] }}
      state=present
      record_name="stage"
      record_type="A"
      record_value="{{ item.public_ip }}"
    with_items: ec2.instances
    when: hostvars[item.id]['ec2_tag_Domain'] is defined
```

The "itag" is all the ec2 tags storing the ec2 instance ids instead of the destination.  I also store the host info by instance id in the hostvars so you can reference the ec2 instance by id.

### DIFF
--- a/plugins/inventory/ec2.py
+++ b/plugins/inventory/ec2.py
@@ -442,6 +442,9 @@ class Ec2Inventory(object):
             if dest is None:
                 dest = getattr(instance, 'tags').get(self.destination_variable, None)
 
+        self.group_by_itag_keys_value_is_instance_id(instance)
+        self.store_host_info_by_instance_id_in_hostvars(instance)
+
         if not dest:
             # Skip instances we cannot address (e.g. private VPC subnet)
             return
@@ -548,6 +551,17 @@ class Ec2Inventory(object):
 
         self.inventory["_meta"]["hostvars"][dest] = self.get_host_info_dict_from_instance(instance)
 
+    def store_host_info_by_instance_id_in_hostvars(self, instance):
+        self.inventory["_meta"]["hostvars"][instance.id] = self.get_host_info_dict_from_instance(instance)
+
+    def group_by_itag_keys_value_is_instance_id(self, instance):
+        if self.group_by_tag_keys:
+            for k, v in instance.tags.items():
+                if v:
+                    key = self.to_safe("itag_" + k + "=" + v)
+                else:
+                    key = self.to_safe("itag_" + k)
+                self.push(self.inventory, key, instance.id)
 
     def add_rds_instance(self, instance, region):
         ''' Adds an RDS instance to the inventory and index, as long as it is


willthames:
I don't understand the itag_ bit. 

What does itag_ solve that the existing tag_ groups don't solve?

willthames:
Also, this change contains merge commits - it should be rebased onto upstream devel, not merged into it (`git pull --rebase upstream devel` generally does the trick)

willthames:
The following seems to work for me and does what you want: 

```
- hosts: tag_Name_micro_host1
  gather_facts: False

  tasks:
  - local_action: debug msg="{{ec2_region}} {{ec2_ip_address}} {{ec2_image_id}}"
```

This is with an ec2.py in the inventory directory to gather the inventory details.
You don't need the hostvars[inventory_hostname] stuff - as you're running a task, it relates to the host you're working with.

amenonsen:
I spent some time on reading through this, but I don't understand it either.

yoshiwaan:
I think this is a fairly redundant feature as the instance id is available as a fact to each ec2 instance using ec2_facts:

```
- hosts: tag_Environment_Development
  gather_facts: false
  tasks:
  - name: get ec2 facts
    action: ec2_facts
  - name: Stop EC2 instance(s)
    local_action: ec2
        state=stopped
        region="{{ ansible_ec2_placement_region }}"
        instance_ids="{{ ansible_ec2_instance_id }}"
```

You can look at available facts like this:
`ansible  tag_Environment_Development -i ec2.py -m ec2_facts`

Now if you want to get the id of something other than an instance then I think you might have a use case

Edit: There is actually an advantage to this. You don't need to actually SSH onto the box to get the ID, you can get the ID from a host level API query. Also useful if you block the metadata API for whatever reason.

s-hertel:
@jason-gill We are not working enhancing the inventory scripts since adding the inventory plugins (new in Ansible 2.4). You can set flexible hostvars and groups using the [aws_ec2 inventory plugin](https://docs.ansible.com/ansible/latest/plugins/inventory/aws_ec2.html) with the constructed options.
Feel free to reach out if you have questions.

PR #11405, opened 2015-06-26
Remove install_requires dependency on setuptools.

Julian:
Ansible does not `install_require`  setuptools, setuptools is a build / setup-time requirement, one that if it wasn't present, would make the `setup.py` not even importable.

Putting setuptools in install_requires means that because pip has only a full-upgrade mode where all dependencies are upgraded, if someone tries to `pip install -U ansible` and is not on the latest setuptools, pip will try to upgrade it, which isn't desirable unless it's specifically intended, not to mention that there are means of installing ansible that do not require setuptools to be installed at all.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ setup(name='ansible',
       author_email='michael@ansible.com',
       url='http://ansible.com/',
       license='GPLv3',
-      install_requires=['paramiko', 'jinja2', "PyYAML", 'setuptools', 'pycrypto >= 2.6', 'six'],
+      install_requires=['paramiko', 'jinja2', "PyYAML", 'pycrypto >= 2.6', 'six'],
       package_dir={ '': 'lib' },
       packages=find_packages('lib'),
       package_data={


abadger:
The pkg_resources module is shipped as part of the setuptools python package.  pkg_resources is a runtime requirement rather than a buildtime requirement.  So that portion of the analysis in this bug is incorrect.

The lack of finer grained upgrade modes is arguably a missing feature in pip.  it certainly is a hindrance for cases like this where the user (not the upstream project) wants to pin one version of a dependency.

However, since the usage of pkg_resources started we've modified the code to make the import of pkg_resources optional.  pkg_resources let's us run in situations where multiple versions of a library exist on a system and the library version that we need is not the default.  If all the other libraries in their default versions are good for ansible (or we never hit the code [mostly vault right now] that need those specific versions of libraries) then the pkg_resources code doesn't need to be run.

The multi-version code is mostly needed for system packagers who need to ship different pieces of software that rely on different library versions.  People who do a pip install -U ansible will get the proper versions of the libraries (provided that ansible can work with the latest version of the library... I'm not sure that pip will downgrade libraries with -U... I think it will only upgrade them).  System packagers often look at the setup.py to tell what they need to add to their package requirements so if we remove setuptools from install_requires we should add a comment in the setup.py to let them know that we don't include setuptools in the install_require field but they may need to Require setuptools/pkg_resources in their packages if they have pyhton libraries where the version that we need is not the default in their distribution.

@jimi-c, @bcoca -- I think mpdehaan was the one who asked that we add setuptools to install_require.  So if either of you would like to check my thought process above and +1 removal that would make me feel better about removing something that someone else asked for ;-)

@Julian Assuming that I haven't overlooked something and I get a +1 from one of the other Core devs, if you add the comment I mentioned to setup.py near install_requires I'll merge this.

Julian:
Ah, sorry, I didn't notice that you use pkg_resources at runtime, yes definitely -- and yeah agreed, pip is mostly at fault here but it's not growing a dependency resolver anytime soon unfortunately. Apologies for being careless on the first part.

I'll read your comment a bit more carefully in a sec and add the comment you mentioned once I get a chance to digest.

Thanks for the quick turnaround.

amenonsen:
In light of the comments above, I guess this should be closed now.

Julian:
In light of which comments?

amenonsen:
In light of the comments that… I dreamed up, apparently. Sorry, I must've gotten confused somehow, but now I can't figure out if I commented on the wrong PR or what. But we're waiting for an updated patch from you that adds a comment, yes?

Julian:
:) no worries.

And yes we're waiting on an updated patch, which is waiting on the +1 that @abadger asked for for me to finish.

cdosborn:
If a user installs ansible with a `setuptools < 13.1`. They will get a runtime exception. (This is addressed in https://github.com/ansible/ansible/pull/16723). Because of pip's lack of resolution, declaring a `install_requires = [setuptools, ...]` will override any further requirements for setuptools. The error I mention is caused by cryptography having an `install_requires = [setuptools >= 13.1` that gets ignored. So, setuptools needs to a higher lower bound ( >= 15.0) or it should be removed. 

Other projects are getting burned by this:
https://github.com/erikrose/letsencrypt/commit/6d8cd2460edf415af43bc2e0134438ba2c8a22c4

jimi-c:
As this has been open for a while, and in light of the fact that it does appear to be required at some level based on @cdosborn's comment, we're going to close this and request that a new issue be opened for the setuptools version requirement.

Thank you!

PR #11415, opened 2015-06-28
Remove unnecessary imports

msabramo:


bcoca:
this adds role_apply which is also not used

msabramo:
Ok, updated to remove `role_apply`

PR #11416, opened 2015-06-28
Add ability to adhoc apply roles to hosts

msabramo:
I have often wanted the ability to adhoc apply a role to hosts. Yes, I can write a playbook to do this, but I feel like then I would accumulate a big list of playbooks that are of questionable usefulness and would clutter the repository.

This would also make it easier for people to demo and test and code review roles from galaxy as the user does not need to create a playbook just to try out the role.

I've seen people ask for this before -- e.g.:
- https://groups.google.com/forum/#!topic/ansible-project/h-SGLuPDRrs

E.g.:

```
$ ansible vagrant -r docker --sudo
-------------------------------------------------------------------------------
#!/usr/bin/env ansible-playbook

---
- hosts:
    - vagrant
  roles:
    - docker
-------------------------------------------------------------------------------
Executing: ansible-playbook --sudo /var/folders/gw/w0clrs515zx9x_55zgtpv4mm0000gp/T/tmpNimKE4.yml

PLAY: ***************************************************************************

TASK [setup] ********************************************************************
ok: [vagrant]

TASK [angstwad.docker_ubuntu : angstwad.docker_ubuntu : Fail if not a new release of Ubuntu] ***
skipping: [vagrant]
...
```

```
$ ansible vagrant --role-name=docker --sudo -vvvv --timeout=30
-------------------------------------------------------------------------------
#!/usr/bin/env ansible-playbook

---
- hosts:
    - vagrant
  roles:
    - docker
-------------------------------------------------------------------------------
Executing: ansible-playbook --sudo -vvvv --timeout=30 /var/folders/gw/w0clrs515zx9x_55zgtpv4mm0000gp/T/tmpW0f6ZX.yml
1 plays in /var/folders/gw/w0clrs515zx9x_55zgtpv4mm0000gp/T/tmpW0f6ZX.yml

PLAY: ***************************************************************************
...
```

Applying a role ad hoc seems to me to be quite similar to ad hoc applying a module in that you'd want to do these for the same reasons (e.g.: ability to quickly try things out, ability to do one-off operations, avoid accumulation of tons of trivial little playbooks, make ansible easier to demo and easier to learn). 
Ansible currently allows ad-hoc execution of modules but roles require a playbook but I see them as quite similar; hence this PR. 

Cc: @georges, @sontek, @djeebus

### DIFF
--- a/lib/ansible/cli/adhoc.py
+++ b/lib/ansible/cli/adhoc.py
@@ -25,6 +25,7 @@ from ansible.parsing.splitter import parse_kv
 from ansible.playbook.play import Play
 from ansible.cli import CLI
 from ansible.utils.display import Display
+from ansible.cli.role_apply import role_apply
 from ansible.utils.vault import read_vault_file
 from ansible.vars import VariableManager
 
@@ -53,7 +54,9 @@ class AdHocCLI(CLI):
             help="module arguments", default=C.DEFAULT_MODULE_ARGS)
         self.parser.add_option('-m', '--module-name', dest='module_name',
             help="module name to execute (default=%s)" % C.DEFAULT_MODULE_NAME,
-            default=C.DEFAULT_MODULE_NAME)
+            default=None)
+        self.parser.add_option('-r', '--role-name', dest='role_name',
+            help="role name to execute")
 
         self.options, self.args = self.parser.parse_args()
 
@@ -111,6 +114,17 @@ class AdHocCLI(CLI):
             for host in hosts:
                 self.display.display('    %s' % host)
             return 0
+
+        if self.options.role_name:
+            if self.options.module_name:
+                raise AnsibleOptionsError(
+                    "You can't pass both a role (%s) "
+                    "and a module (%s) to execute"
+                    % (self.options.role_name, self.options.module_name))
+            role_apply(hosts, [self.options.role_name], show_playbook=True)
+            return 0
+        elif self.options.module_name is None:
+            self.options.module_name = C.DEFAULT_MODULE_NAME
 
         if self.options.module_name in C.MODULE_REQUIRE_ARGS and not self.options.module_args:
             raise AnsibleOptionsError("No argument passed to %s module" % self.options.module_name)
--- a/lib/ansible/cli/role_apply.py
+++ b/lib/ansible/cli/role_apply.py
@@ -1,0 +1,94 @@
+# -*- coding: utf-8 -*-
+# (c) 2015, Marc Abramowitz <msabramo@gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+# Make coding more python3-ish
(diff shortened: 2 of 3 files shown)


ienliven:
+1

msabramo:
Now featuring...tests!

```
$ py.test test/units/cli/test_role_apply.py \
>     --cov=ansible.cli.role_apply --cov-report=term-missing
============================================================================= test session starts ==============================================================================
platform darwin -- Python 2.7.9 -- py-1.4.30 -- pytest-2.7.2
rootdir: /Users/marca/dev/git-repos/ansible, inifile:
plugins: cov
collected 3 items

test/units/cli/test_role_apply.py ...
--------------------------------------------------------------- coverage: platform darwin, python 2.7.9-final-0 ----------------------------------------------------------------
Name                         Stmts   Miss  Cover   Missing
----------------------------------------------------------
lib/ansible/cli/role_apply      42      0   100%
```

djeebus:
Nice, I like it! This makes my life easier =)

bcoca:
i'm sorry to say, we are going to pass on this feature, for several reasons.

Many roles will depend on facts, which now they will have to gather themselves and now making them adhoc only or playbook only roles or duplicating efforts on a very costly operation.

The plays and inventory groups serve as documentation of what hosts should/have what roles applied to them, this also allows for repeatable runs and for you to know which machines you should patch and upgrade.

Another reason is that we have planned adding ansible-console (https://github.com/debops/ansible-console) which will give a closer pssh/cssh like interaction for those that need it (you can currently install it independently to use today).

msabramo:
Thanks for taking the time to review!

> Many roles will depend on facts, which now they will have to gather themselves

I don't understand. My implementation puts the role in a temporary playbook so won't it gather facts automatically? I actually wondered if I should add some kind of flag that allows adding `gather_facts: no` to the generated playbook to speed things up for cases where the facts aren't needed. 

As for playbooks serving as documentation and making things repeatable, that is true of course that they do that, but I would say that is just as true for modules as it is for roles. In fact the module documentation usually includes sample playbook yaml for the module and usually doesn't mention how to use it ad hoc. But modules are allowed to be executed ad hoc if the user wishes, even though playbooks are preferred for most cases. So I don't see why roles are treated differently. Both modules and roles are used primarily as building blocks for playbooks, but only modules can be used ad hoc. Seems asymmetric.

Playbooks excel when you want to apply several roles to several hosts. But when I download a new role from Galaxy, I don't want to throw it in a playbook and deploy it to a whole bunch of hosts in my infrastructure. I want to deploy it to one host (usually localhost or a vagrant VM) and see how it works. A playbook works but feels a little overkill. 

Since this probably won't be merged, here's a tip for folks who want an easy way to deploy a single role and don't want to have a zillion different playbooks laying around. Create a playbook called `deploy_role.yml`:

``` yaml
- hosts: {{ hosts }}
  roles: {{ roles }}
```

And then you can do `ansible-playbook deploy_role.yml -e hosts=somehost -e roles=somerole`. I do this now and it kind of works but I have to be in the right directory for it to work so I'd prefer something built-in.

msabramo:
I fail to see how ansible-console addresses this. Did you mean [`ansible-shell`](https://github.com/dominis/ansible-shell) perchance?

bcoca:
yes, sorry, it was ansible-shell i meant

The part of plays being documentation is not about documenting ansible features, but documenting the relationships in your system (what machines have what roles applied). 

Adhoc was created specifically to allow for a single task play, one that does not gather facts, which would be a 2nd task (also the reason template does not work with adhoc). If you change adhoc to gather facts it looses it's 'light weight' command function.

Also your hack above has a few problems, first you are overloading internal keywords as variables, this can break things easily depending on context, use hosts: "{{target}}". Which takes me to 2nd issue, you need quotes. Another issue is that roles were not meant to be dynamically constructed and that will only work on certain versions of ansible which leak variables into directives.

msabramo:
Ok. Thanks for the further explanation!

crccheck:
ansible-shell/ansible-console does not solve my use case. Instead of a REPL, I'm looking for a way to document an ad-hoc playbook that I can share as part of a test plan for a role. It looks like this PR makes that possible.

PR #11426, opened 2015-06-29
adhoc: allow -i * to use all hosts as-is

msabramo:
Sometimes I want to do something quick against a brand new host that I
haven't yet added to the inventory. This is essentially a more flexible
version of the "comma trick". This allows passing `-i '*'`, which means
basically don't look at inventory files and just try to operate against
any host that I specify verbatim. E.g.:

```
# remotehost is not in any ansible inventory file or dynamic inventory
# but I can target it anyway
$ ansible -i '*' myuser@remotehost -a 'lsb_release -a'
myuser@remotehost | SUCCESS => {
    ...
    "stdout": "Distributor ID:\tUbuntu\nDescription:\tUbuntu 12.04.5 LTS\nRelease:\t12.04\nCodename:\tprecise",
    "cmd": [
        "lsb_release",
        "-a"
    ],
    ...
    "stdout_lines": [
        "Distributor ID:\tUbuntu",
        "Description:\tUbuntu 12.04.5 LTS",
        "Release:\t12.04",
        "Codename:\tprecise"
    ],
    "warnings": []
}
```

Note that this also makes the experience easier for new users and means that
the "Getting started" docs
(http://docs.ansible.com/intro_getting_started.html#your-first-commands) don't
need to cover inventory right away.

If one has an `ansible.cfg` with:

```
[defaults]
inventory = *
```

then it becomes even easier and `-i '*'` can be omitted -- e.g.:

```
ansible myuser@remotehost -a date
```

### DIFF
--- a/lib/ansible/cli/adhoc.py
+++ b/lib/ansible/cli/adhoc.py
@@ -100,7 +100,10 @@ class AdHocCLI(CLI):
         loader = DataLoader(vault_password=vault_pass)
         variable_manager = VariableManager()
 
-        inventory = Inventory(loader=loader, variable_manager=variable_manager, host_list=self.options.inventory)
+        if self.options.inventory == '*':
+            inventory = Inventory(loader=loader, variable_manager=variable_manager, host_list=pattern + ',')
+        else:
+            inventory = Inventory(loader=loader, variable_manager=variable_manager, host_list=self.options.inventory)
         variable_manager.set_inventory(inventory)
 
         hosts = inventory.list_hosts(pattern)


sivel:
I find this confusing.  I would think that a `*` would be the equivalent of `all` and be more inline with limiting as opposed to being the inventory itself.

This makes it feel backwards to me.

And in fact, the comma functionality is largely there to support internal cases such as what is used in `ansible-pull` as opposed to being a "production" use for the CLI.

I would actually find it more logical to type something like;

```
ansible all -i foo, -m ping
```

Which is basically what I do now when testing functionality.

sivel:
Also, just to give my description of what I understand this to do:

> Providing an inventory of `*` would ignore the "provided" inventory, and instead use the limit spec as the inventory.

I could see many places where this would also cause lots of confusion and places where it just wouldn't work.

msabramo:
My initial implementation used `-i None` to say that you are using no inventory file. That seemed weird though. I'm not really sure `'*'` is any better; I guess not if you find it confusing. I don't think `ansible all -i foo, -m ping` is particularly intuitive either.

Here's my description of what I intended this to do:

> Providing an inventory of  `*` means to not use an inventory and to try to talk to all given hosts, so that you can target actual hostnames without having to make sure that they're in an inventory.

So I don't know, maybe `--no-inventory` is more accurate.

Maybe what is should really do is allow an inventory but also allow the use of hosts that are not in the inventory.

msabramo:
I think I have a better idea...

msabramo:
Closing in favor of #11431

PR #11428, opened 2015-06-29
Display config file read when -v provided

msabramo:
Because I sometimes get confused because I don't know which config file is being picked up.

```
$ ansible foo -m ping -v
Reading config file: /Users/msabramo/smstack/ansible.cfg
...
```


bcoca:
we want to move prints out of constants, I'm thinking we should just assign it to a constant and print from the cli when verbose.

bcoca:
fixed via 5f791329ce2f452b99ee74b9cfca4de83ac37e0e

PR #11429, opened 2015-06-29
Use a list for tags, but with the assistance of sets to remove duplicates. Fixes #11424

### LINKED ISSUE #11424: 0eb1c88 causes traceback when specifying tags
Via a git bisect, I have found that a recent traceback when specifying tags on the command line, was introduced in 0eb1c88 to resolve #9380 

That commit references:

> Use itertools instead of set for tags

```
  File "/Users/matt/python_venvs/ansibledev/ansible/bin/ansible-playbook", line 66, in <module>
    sys.exit(cli.run())
  File "/Users/matt/python_venvs/ansibledev/ansible/lib/ansible/cli/playbook.py", line 161, in run
    results = pbex.run()
  File "/Users/matt/python_venvs/ansibledev/ansible/lib/ansible/executor/playbook_executor.py", line 127, in run
  File "/Users/matt/python_venvs/ansibledev/ansible/lib/ansible/executor/task_queue_manager.py", line 248, in run
    iterator = PlayIterator(inventory=self._inventory, play=new_play, connection_info=connection_info, all_vars=all_vars)
  File "/Users/matt/python_venvs/ansibledev/ansible/lib/ansible/executor/play_iterator.py", line 95, in __init__
    new_block = block.filter_tagged_tasks(connection_info, all_vars)
  File "/Users/matt/python_venvs/ansibledev/ansible/lib/ansible/playbook/block.py", line 318, in filter_tagged_tasks
    new_block.block  = evaluate_and_append_task(self.block)
  File "/Users/matt/python_venvs/ansibledev/ansible/lib/ansible/playbook/block.py", line 313, in evaluate_and_append_task
    if task.evaluate_tags(connection_info.only_tags, connection_info.skip_tags, all_vars=all_vars):
  File "/Users/matt/python_venvs/ansibledev/ansible/lib/ansible/playbook/taggable.py", line 82, in evaluate_tags
    elif tags.intersection(only_tags):
AttributeError: 'list' object has no attribute 'intersection'
```

cc @jimi-c

sivel:
This PR addresses #11424 by reverting a little of 0eb1c88 and using using `set` to remove duplicates, casting back to a list and using `set` as needed in the last mile to check for intersections.

Overall, this may not be the right course of action, maybe it would be better to handle the unhashable errors by casting to a list where needed, and not from the start.

### DIFF
--- a/lib/ansible/playbook/taggable.py
+++ b/lib/ansible/playbook/taggable.py
@@ -19,7 +19,6 @@
 from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
-import itertools
 from six import string_types
 
 from ansible.errors import AnsibleError
@@ -28,7 +27,7 @@ from ansible.template import Templar
 
 class Taggable:
 
-    untagged = set(['untagged'])
+    untagged = ['untagged']
     _tags = FieldAttribute(isa='list', default=[], listof=(string_types,int))
 
     def __init__(self):
@@ -64,11 +63,11 @@ class Taggable:
 
             if not isinstance(tags, list):
                 if tags.find(',') != -1:
-                    tags = set(tags.split(','))
+                    tags = list(set(tags.split(',')))
                 else:
-                    tags = set([tags])
+                    tags = list(set([tags]))
             else:
-                tags = [i for i,_ in itertools.groupby(tags)]
+                tags = list(set(tags))
         else:
             # this makes intersection work for untagged
             tags = self.__class__.untagged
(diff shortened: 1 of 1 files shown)


bcoca:
so my first instinct is to make sure that the incoming tags are a set already, any reason to convert all sets to lists instead?

sivel:
It was done in the original commit from @jimi-c because the set was unhashable and causing a trace back.  

If we want to keep these as lists, as you can see here, things get a bit more complicated.

jimi-c:
I don't think this is the right way to fix this. I used itertools specifically to fix another class of errors that otherwise was very difficult to catch (when someone mistakenly creates a list of lists for tags).

The simple fix is to convert the list to a set here: https://github.com/ansible/ansible/commit/0eb1c880ddac9547560040311739b5ca8291a642#diff-ea1d23f1185cc91acace441695571d0fR71, which was an oversight on my part.

jimi-c:
Fixed as noted above. Thanks!

PR #11431, opened 2015-06-29
adhoc: Allow hosts not in inventory to be targeted

msabramo:
Sometimes I want to do something quick against a brand new host that I haven't yet added to the inventory. This is essentially a more flexible version of the "comma trick".

E.g.:

```
# `somehost` is not in the ansible inventory

$ ansible somehost -m ping --allow-hosts-not-in-inventory
user@host | SUCCESS => {
    "invocation": {
        "module_name": "ping",
        "module_args": {}
    },
    "changed": false,
    "ping": "pong"
}
```

Note that this also makes the experience easier for new users and means that the "Getting started" docs
(http://docs.ansible.com/intro_getting_started.html#your-first-commands) don't need to cover inventory right away, which makes for a more shallow learning curve.

Replaces https://github.com/ansible/ansible/pull/11426

### DIFF
--- a/lib/ansible/cli/adhoc.py
+++ b/lib/ansible/cli/adhoc.py
@@ -54,6 +54,9 @@ class AdHocCLI(CLI):
         self.parser.add_option('-m', '--module-name', dest='module_name',
             help="module name to execute (default=%s)" % C.DEFAULT_MODULE_NAME,
             default=C.DEFAULT_MODULE_NAME)
+        self.parser.add_option('--allow-hosts-not-in-inventory',
+            action='store_true', dest='allow_hosts_not_in_inventory',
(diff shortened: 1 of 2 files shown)


msabramo:
Cc: @sudarkoff, @djeebus

bcoca:
Discussed this with the core team and we think a better implementation would be to add an 'adhoc' inventory plugin to enable:

`ansible -i adhoc hostnotininventory -m ping`

msabramo:
Sounds reasonable. I'll try to hack one together at some point.

amenonsen:
I think this PR should be closed now. If @msabramo has a chance to implement an adhoc inventory plugin (and I like that idea), it belongs in a separate PR anyway.

amenonsen:
Closing per above comment.

Issue #11434, opened 2015-06-29
ipwrap filter doesn't work with generators


hfinucane:
Example:

```
  - debug: msg="{{ ['fe80::a236:9fff:fe5f:6af4']|ipwrap|join(',') }}"
  - debug: msg="{{ ['fe80::a236:9fff:fe5f:6af4']|map('string')|ipwrap|join(',') }}"
```

`string` being a filter plugin (http://jinja.pocoo.org/docs/dev/templates/#string) that, confusingly, does what `unicode()` does in Python. It returns a generator.

I would expect the output to be the same, but actually:

```
TASK [debug msg=[fe80::a236:9fff:fe5f:6af4]] ************************************
ok: [xxx] => {
    "msg": "[fe80::a236:9fff:fe5f:6af4]", 
    "changed": false
}

TASK [debug msg=fe80::a236:9fff:fe5f:6af4] **************************************
ok: [xxx] => {
    "msg": "fe80::a236:9fff:fe5f:6af4", 
    "changed": false
}

```

This is true in 1.9.1 and in devel.

sivel:
Can you please fill out the [issue template](https://github.com/ansible/ansible/blob/devel/ISSUE_TEMPLATE.md) and provide it in this ticket so that we have more information including how to reproduce your error?

hfinucane:
(yeah, I hit enter before entering anything- sorry)

hfinucane:
Actually, I think I'm wrong- the root cause isn't that they're unicode, it's that the various Jinja2 filters pass _generators_ around, and there's a bunch of hardcoded tests for 'is this a list or a tuple'.

hfinucane:
Additionally, now that I've figured this out, an effective workaround is piping things through the `list` filter before sending them to `ipwrap`.

bcoca:
closing via merged #11435

PR #11435, opened 2015-06-29
Treat generators like lists and tuples

hfinucane:
Fix for #11434. Jinja2 filters use generators extensively, so anyone doing something like

```
{{ ansible_interface.ipv6|map(attribute='address')|ipwrap }}
```

is liable to get in trouble, otherwise.

### DIFF
--- a/lib/ansible/plugins/filter/ipaddr.py
+++ b/lib/ansible/plugins/filter/ipaddr.py
@@ -16,6 +16,7 @@
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 
 from functools import partial
+import types
 
 try:
     import netaddr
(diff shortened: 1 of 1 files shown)


hfinucane:
Is there a place to add integration tests for filters? I didn't see any for `ipwrap` after a cursory grep.

bcoca:
there are tests for some filters but not all, look under test/integration/test_filters.ym  in the repo and feel free to add tests for ipwrap there.

abadger:
@hfinucane I was looking at this patch the other day and wondered if there were other iterables we might like to catch (sets [probably] and dicts [on the fence], for instance).

If so, maybe we would want to do something like:

``` python
from collections import Iterable
if isinstance(value, Iterable) and not isinstance(value, basestring):
    # [....]
```

Does that seem like a good idea to you or overengineering the solution?

hfinucane:
I do think it's a good idea. I was feeling a bit shy about making larger changes, partly out of unfamiliarity, and partly out of fear of creating a regression somehow.

For the case of `ipwrap`, dicts don't really make a lot of sense-  are you iterating over keys? values? both? It's a bit awkward for the general case. Maybe if we also added `attribute=` like some of the other, more general Jinja2 filters.

abadger:
Yeah, maybe we should include dicts with basestring as things that are Iterable that we do not want to allow with this filter.

Issue #9615, opened 2014-11-25
RFE: Getting azure hosts doesn't return hosts but cloud services


jruix:
##### Summary:

In Azure, hosts are all inside a cloud service ( behind a load balancer ) and they share the same FQN so in order to access to one host or another the load balancer maps a public port from the public IP to the port 22 of the internal host's private ip.

Using Ansible Tower when adding an inventory based on Azure as SRC, it connects to azure and it gets the cloud services FQN correctly , but not the hosts. Seems that Ansible Tower think that the cloud service is in fact the host, I guess this is because the way Azure works is slightly different from AWS where all hosts has a public dns name.
##### Steps To Reproduce:

In Ansible tower Add an Inventory group using Azure as source
##### Expected Results:

I expect to get a group with all the hosts located in azure
##### Actual Results:

I got a group with all the cloud services in azure but not the hosts

mpdehaan:
Hi, for ansible Tower questions please email support@tower.com -- thank you!

jruix:
They asked me to fill a issue in the github repo :-/, seems we got an infinite loop here ;)

PR #11446, opened 2015-06-30
Updates to Windows Azure inventory from Tower.  

cchurch:
Includes earlier changes from https://github.com/ansible/ansible/pull/8403 along with:
- Configurable cache path
- Support for `--host hostname` command line syntax
- Include `['_meta']['hostvars']` key in `--list` response
- Set `ansible_ssh_host` and `ansible_ssh_port` in host variables
- Set host variable for instance status and private ID
- Return all instances in deployment to fix issue described in https://github.com/ansible/ansible/issues/9615

### DIFF
--- a/contrib/inventory/windows_azure.ini
+++ b/contrib/inventory/windows_azure.ini
@@ -13,8 +13,9 @@
 
 # API calls to Windows Azure may be slow. For this reason, we cache the results
 # of an API call. Set this to the path you want cache files to be written to.
-# One file will be written to this directory:
+# Two files will be written to this directory:
 #   - ansible-azure.cache
+#   - ansible-azure.index
 #
 cache_path = /tmp
 
--- a/contrib/inventory/windows_azure.py
+++ b/contrib/inventory/windows_azure.py
@@ -65,6 +65,14 @@ class AzureInventory(object):
         self.inventory = {}
         # Index of deployment name -> host
         self.index = {}
+        self.host_metadata = {}
+
+        # Cache setting defaults.
+        # These can be overridden in settings (see `read_settings`).
+        cache_dir = os.path.expanduser('~')
+        self.cache_path_cache = os.path.join(cache_dir, '.ansible-azure.cache')
+        self.cache_path_index = os.path.join(cache_dir, '.ansible-azure.index')
+        self.cache_max_age = 0
 
         # Read settings and parse CLI arguments
         self.read_settings()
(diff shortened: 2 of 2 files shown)


plugins/inventory/windows_azure.py
```diff
@@ -82,15 +90,39 @@ def __init__(self):
...
+                data = self.inventory
 
+            if self.args.host:
+                data_to_print = self.get_host(self.args.host)
+            else:
+                # Add the `['_meta']['hostvars']` information.
+                hostvars = {}
+                if len(data) > 0:
+                    for host in set(reduce(lambda x, y: x + y,
+                                           [i for i in data.values()])):
```
abadger: Not sure about this section.
- What's in data.values()?  If it's numbers then you could do `sum(data.values())` instead.  If it's strings, then you can do `"".join(data.values())`.
- I think that reduce returns a scalar.  So there's no reason to turn it into a set and looping over it.
- The above thoughts don't seem to jive with setting host from set(reduce([...]) as there will be only one value for host after doing this?

plugins/inventory/windows_azure.py
```diff
@@ -121,28 +153,36 @@ def read_settings(self):
 
         # Cache related
         if config.has_option('azure', 'cache_path'):
-            cache_path = config.get('azure', 'cache_path')
-            self.cache_path_cache = cache_path + "/ansible-azure.cache"
-            self.cache_path_index = cache_path + "/ansible-azure.index"
+            cache_path = os.path.expanduser(config.get('azure', 'cache_path'))
```
abadger: Do we want to do expandvars() as well?

plugins/inventory/windows_azure.py
```diff
@@ -121,28 +153,36 @@ def read_settings(self):
...
+            self.cache_path_cache = cache_path + '/ansible-azure.cache'
+            self.cache_path_index = cache_path + '/ansible-azure.index'
```
abadger: Note: idiomatically we'd do os.path.join() here instead of putting "/" in explicitly but you don't need to change it if you don't want to as it appears that it didn't trip anyone up previously.

plugins/inventory/windows_azure.py
```diff
@@ -163,11 +203,14 @@ def add_cloud_services(self):
...
+        """Makes an Azure API call to get the list of virtual machines
+        associated with a cloud service.
+        """
         try:
             for deployment in self.sms.get_hosted_service_properties(cloud_service.service_name,embed_detail=True).deployments.deployments:
-                if deployment.deployment_slot == "Production":
-                    self.add_deployment(cloud_service, deployment)
+                self.add_deployment(cloud_service, deployment)
+                #if deployment.deployment_slot == "Production":
+                #    self.add_deployment(cloud_service, deployment)
```
abadger: Why change this and comment it out?  If it was wrong we should just remove the code.  If it's right in some circumstances then maybe we need a commandline option or a config option to control it.

plugins/inventory/windows_azure.py
```diff
@@ -176,20 +219,43 @@ def add_deployments(self, cloud_service):
...
+                        break
+            except AttributeError as e:
+                pass
+            finally:
+                self.add_instance(role.instance_name, deployment, port, cloud_service, role.instance_status)
+
+    def add_instance(self, hostname, deployment, ssh_port, cloud_service, status):
+        """Adds an instance to the inventory and index"""
 
         dest = urlparse(deployment.url).hostname
```
abadger: Is this line still needed?  Looks like dest is no longer used.

abadger:
Reviewed the code from a python standpoint.  I don't use azure so I can't comment on whether the changes make sense for the user or whether the right APIs are being used. Unless someone else speaks up to object to the changes I'll trust that you all have considered those particular issues.

plugins/inventory/windows_azure.py
cchurch: `data.values()` is a list of lists, so it's just flattening that into one list.  I've updated it to use a nested list comprehension instead, which may be (slightly) more readable.

plugins/inventory/windows_azure.py
cchurch: Fixed.

plugins/inventory/windows_azure.py
cchurch: Fixed to use `os.path.join()` here and another place in the file.

plugins/inventory/windows_azure.py
cchurch: Removed for now.  It appears it was changed in Tower to return all hosts instead of just "Production".  Might be nice to have as a config option, but so far I haven't seen any requests for that feature.

plugins/inventory/windows_azure.py
cchurch: `dest` is used to create the dictionary a few lines below.

cchurch:
@abadger I've fixed and/or addressed your comments and rebased.

PR #10776, opened 2015-04-19
Error messages from a CLI should be sent to stderr

romain-dartigues:
##### Issue Type:

Feature Pull Request
##### Ansible Version:

```
ansible 2.0.0 (devel f4172fb9da) last updated 2015/04/18 19:42:59 (GMT +200)
  lib/ansible/modules/core: (detached HEAD a19fa6ba48) last updated 2015/04/18 19:43:07 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD df7fcc90d9) last updated 2015/04/18 19:43:15 (GMT +200)
  v2/ansible/modules/core: (detached HEAD 34784b7a61) last updated 2015/04/18 19:43:24 (GMT +200)
  v2/ansible/modules/extras: (detached HEAD df7fcc90d9) last updated 2015/04/18 19:43:35 (GMT +200)
  configured module search path = None
```
##### Environment:

Irrevelant (Slackware GNU/Linux).
##### Summary:

The CLI in the inventory plugins directory should write their errors to `stderr`, especially if they are meant to be called by [InventoryScript()](https://github.com/ansible/ansible/blob/c3d34a538613eea3d6db03a700211887f697b443/lib/ansible/inventory/script.py#L48) as the [docker script does](https://github.com/ansible/ansible/blob/8c4161d4a12849afd46dea7c63b540b0437e03c3/plugins/inventory/docker.py#L178).

Note this is replacing the issue #10764 i opened after my misunderstanding.
##### Steps To Reproduce:

Execute incorrectly most of the scripts from the [plugins/inventory/](https://github.com/ansible/ansible/blob/devel/plugins/inventory/) directory, ie:

``` sh-session
$ /path/to/ansible/plugins/inventory/digital_ocean.py --all >/dev/null ; echo $?
1
$ ansible -m ping -u root -i digital_ocean.py all
ERROR: Inventory script (digital_ocean.py) had an execution error:
```
##### Expected Results:

``` sh-session
$ /path/to/ansible/plugins/inventory/digital_ocean.py --all >/dev/null
Could not find values for DigitalOcean client_id and api_key.
They must be specified via either ini file, command line argument (--client-id and --api-key),
or environment variables (DO_CLIENT_ID and DO_API_KEY)
$ ansible -m ping -u root -i digital_ocean.py all
ERROR: Inventory script (digital_ocean.py) had an execution error: Could not find values for DigitalOcean client_id and api_key.
They must be specified via either ini file, command line argument (--client-id and --api-key),
[...]

PR #10491, opened 2015-03-18
Fix retrieving private address

GheRivero:
Fixes #10480

PR #11447, opened 2015-06-30
Openstack inventory updates from Tower.

cchurch:
Includes changes from https://github.com/ansible/ansible/pull/10491 and https://github.com/ansible/ansible/pull/10776 and allows specifying a config file via environment variable.

### DIFF
--- a/contrib/inventory/openstack.py
+++ b/contrib/inventory/openstack.py
@@ -51,11 +51,12 @@ import shade
 class OpenStackInventory(object):
 
     def __init__(self, private=False, refresh=False):
+        config_files = os_client_config.config.CONFIG_FILES
+        config_files.append('/etc/ansible/openstack.yml')
         self.openstack_config = os_client_config.config.OpenStackConfig(
-            os_client_config.config.CONFIG_FILES.append(
-                '/etc/ansible/openstack.yml'),
-            private)
+            config_files)
         self.clouds = shade.openstack_clouds(self.openstack_config)
+        self.private = private
         self.refresh = refresh
 
         self.cache_max_age = self.openstack_config.get_cache_max_age()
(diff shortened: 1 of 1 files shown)


plugins/inventory/openstack.py
```diff
@@ -152,8 +156,7 @@ def main():
         elif args.host:
             inventory.get_host(args.host)
     except shade.OpenStackCloudException as e:
-        print(e.message)
-        sys.exit(1)
+        sys.exit(e.message)
```
abadger: I'd rather be explicit and use:

``` python
sys.stderr.write('%s\n' % e.message)
sys.exit(1)
```

abadger:
@emonty Care to review this and +1 if it looks good to merge?

plugins/inventory/openstack.py
cchurch: Pushed a new commit to fix.  Looking at the other changes in https://github.com/ansible/ansible/pull/10776, that PR is really inconsistent in this regard.

plugins/inventory/openstack.py
```diff
@@ -51,11 +51,14 @@
 class OpenStackInventory(object):
 
     def __init__(self, private=False, refresh=False):
+        config_files = os_client_config.config.CONFIG_FILES
+        if os.environ.get('OPENSTACK_CONFIG_FILE', None):
+            config_files.insert(0, os.environ['OPENSTACK_CONFIG_FILE'])
```
emonty: Would you be amenable to changing this to "OS_CLIENT_CONFIG_FILE"? If so, I'd like to upstream support for setting this into os-client-config itself (it seems like a generally useful feature) We can land it here safely (with the name change) until we get it landed there.

emonty:
I created a commit for supporting the env-var oerride. in openstack/os-client-config here: https://review.openstack.org/198502.

emonty:
@cchurch you ok with the env var name change? We've landed the support for it generally in os-client-config and I can cut a release asap if you're good with that approach.

cchurch:
@emonty I've removed my environment variable override since it's now supported by os-client-config.

@abadger @bcoca I've rebased & squashed, should be ready to go.

PR #11459, opened 2015-07-01
Fix spacing and using name in item loop in task executor, fix using connection shell from plugins

iki:
- [x] Fix spacing and using name in item loop in task executor
- [x] Fix using connection shell from plugins


sivel:
@iki your first commit in `task_executor.py` has been fixed in another PR (https://github.com/ansible/ansible/pull/11458).

There is also another PR to address `_shell` existing as an attribute of `_connection`.  Since this PR has a merge conflict now, and requires revision, I am going to go ahead and close this PR in favor of https://github.com/ansible/ansible/pull/11461

PR #11469, opened 2015-07-02
Fix pyflakes error: use 'msg' variable

pilou-:
1. Error was: local variable 'msg' is assigned to but never used
2. avoid to display "PLAY [PLAY:]" when name of the playbook is undefined

### DIFF
--- a/lib/ansible/playbook/play.py
+++ b/lib/ansible/playbook/play.py
@@ -95,7 +95,7 @@ class Play(Base, Taggable, Become):
 
     def get_name(self):
        ''' return the name of the Play '''
-       return "PLAY: %s" % self._attributes.get('name')
+       return self._attributes.get('name')
 
     @staticmethod
     def load(data, variable_manager=None, loader=None):
--- a/lib/ansible/plugins/callback/default.py
+++ b/lib/ansible/plugins/callback/default.py
@@ -108,4 +108,4 @@ class CallbackModule(CallbackBase):
         else:
             msg = "PLAY [%s]" % name
 
-        self._display.banner(name)
+        self._display.banner(msg)
--- a/test/units/playbook/test_play.py
+++ b/test/units/playbook/test_play.py
@@ -39,7 +39,7 @@ class TestPlay(unittest.TestCase):
 
     def test_empty_play(self):
         p = Play.load(dict())
-        self.assertEqual(str(p), "PLAY: ")
+        self.assertEqual(str(p), '')
 
     def test_basic_play(self):
         p = Play.load(dict(


bcoca:
I'm pretty sure we want the PLAY: to be there for all plays, for the plays w/o names we might want to remove the redundancy there.

pilou-:
With these modifications, `PLAY` is here for all plays:
- without name:
  
  ```
  PLAY ****************************************************************************
  TASK [setup] ********************************************************************
  ```
- with name:
  
  ```
  PLAY [a test] *******************************************************************
  TASK [setup] ********************************************************************
  ```

Would you prefer the outputs below ?
- without name:
  
  ```
  PLAY: ****************************************************************************
  ```
- with name:
  
  ```
  PLAY: a test *********************************************************************
  ```

bcoca:
the "outputs below" is what we already get, not sure what you are looking to do with this PR

pilou-:
['msg' variable is defined but not used](https://github.com/ansible/ansible/blob/29f2b549c2bf4f1809be1b4011c624c16697d19c/lib/ansible/plugins/callback/default.py#L139), it should be fixed.

amenonsen:
@bcoca, this PR is correct. The code used to do:

```
name = ...
<define msg based on name>
self._display.banner(name_instead_of_msg)
```

What the patch does is to change the last `name` to `msg`. The second patch in the PR correctly changes `get_name()` to not add the "PLAY" part (since we're going to prepend it again).

I'm not sure where the merge conflict came from, but here's an updated diff for both changes:

```
diff --git a/lib/ansible/plugins/callback/default.py b/lib/ansible/plugins/callback/default.py
index 8fbb065..b6e56f9 100644
--- a/lib/ansible/plugins/callback/default.py
+++ b/lib/ansible/plugins/callback/default.py
@@ -108,4 +108,4 @@ class CallbackModule(CallbackBase):
         else:
             msg = "PLAY [%s]" % name

-        self._display.banner(name)
+        self._display.banner(msg)
diff --git a/test/units/playbook/test_play.py b/test/units/playbook/test_play.py
index 561da36..0dbac22 100644
diff --git a/lib/ansible/playbook/play.py b/lib/ansible/playbook/play.py
index ecaeac2..e8477c4 100644
--- a/lib/ansible/playbook/play.py
+++ b/lib/ansible/playbook/play.py
@@ -95,7 +95,7 @@ class Play(Base, Taggable, Become):

     def get_name(self):
        ''' return the name of the Play '''
-       return "PLAY: %s" % self._attributes.get('name')
+       return self._attributes.get('name')

     @staticmethod
     def load(data, variable_manager=None, loader=None):
--- a/test/units/playbook/test_play.py
+++ b/test/units/playbook/test_play.py
@@ -39,7 +39,7 @@ class TestPlay(unittest.TestCase):

     def test_empty_play(self):
         p = Play.load(dict())
-        self.assertEqual(str(p), "PLAY: ")
+        self.assertEqual(str(p), '')

     def test_basic_play(self):
         p = Play.load(dict(
```

This can be merged.

pilou-:
PR updated (merge conflict resolved).

PR #11475, opened 2015-07-02
Use cfacter instead of facter if possible

nibalizer:
CFacter is the facter replacement written in C++. It is available from
the puppetlabs repo.

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -2734,12 +2734,16 @@ def get_all_facts(module):
     for (k, v) in facts.items():
         setup_options["ansible_%s" % k.replace('-', '_')] = v
 
-    # Look for the path to the facter and ohai binary and set
+    # Look for the path to the facter, cfacter, and ohai binaries and set
     # the variable to that path.
 
     facter_path = module.get_bin_path('facter')
+    cfacter_path = module.get_bin_path('cfacter')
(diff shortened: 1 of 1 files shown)


emonty:
More speed in facts is more better

lib/ansible/module_utils/facts.py
```diff
@@ -2734,12 +2734,16 @@ def get_all_facts(module):
     for (k, v) in facts.items():
         setup_options["ansible_%s" % k.replace('-', '_')] = v
 
-    # Look for the path to the facter and ohai binary and set
+    # Look for the path to the facter, cfacter, and ohai binaries and set
     # the variable to that path.
 
-    facter_path = module.get_bin_path('facter')
```
bcoca: I think you mean to reverse these

lib/ansible/module_utils/facts.py
```diff
@@ -2734,12 +2734,16 @@ def get_all_facts(module):
...
+    facter_path = module.get_bin_path('cfacter')
+    cfacter_path = module.get_bin_path('facter')
     ohai_path = module.get_bin_path('ohai')
```
bcoca: I think you meant to reverse these

amenonsen:
Looks sensible, has no effect if `cfacter` is not installed.

nibalizer:
I'm unclear why the tests are failing, does anyone know?

Issue #1627, opened 2012-11-14
make "baby JSON" work again


mpdehaan:
bash modules (or even python modules) should be able to return stuff like below that is not actually JSON and it should still work for basic cases

failed=True asdf=1234 foo=bar

etc

mpdehaan:
it actually works, selinux module is just missing "import sys".

good.

PR #11482, opened 2015-07-03
facts: add aliases to ansible_all_ipv4_addresses on OpenBSD

verm666:
1. I edited only lib/ansible/module_utils/facts.py (v1/ansible/module_utils/facts.py still unchanged, is it ok?)
2. Related issue: https://github.com/ansible/ansible-modules-core/issues/1627 (#1627)


bcoca:
also backported to 1.9 tree so it should be released with a possible 1.9.3

PR #11488, opened 2015-07-04
if task is delegated, ansible_facts will be applied to delegated host

### LINKED ISSUE #10608: delegate_to not setting host variable
**_Issue Type:**_

Bug Report

**_Ansible Version:**_

1.8.1

**_Environment:**_

local : MacOsX, 10.10.2
remote: centos 6

**_Summary:**_

Im trying to identify cases when playbook host dos not exists and playbook tasks are skipped.

**_Steps To Reproduce:**_

Run playbook above and replace "192.168.0.1" with actual host you can reach.  

**_Expected Results:**_

Im expecting to set variable on localhost and change value if playbook are executed.

**_Actual Results:**_

Im getting task fail as "play_skiped" keeps = true despite "192.168.0.1" can be reach and task executed.

code: 

``` ansible

---

- hosts: localhost
  pre_tasks:
    - set_fact:
        play_skiped: true

- hosts: "192.168.0.1"
  pre_tasks:
    - set_fact:
        play_skiped: false
      delegate_to: localhost
    - debug:
        msg: "job to run"

- hosts: localhost
  pre_tasks:
    - fail: "host not found, task skiped"
      when: play_skiped
```

bcoca:
once delegate_to actually gets templated this will work for every case, right now it only works for hard-coded delegate_to

should fix #10608

### DIFF
--- a/lib/ansible/executor/process/result.py
+++ b/lib/ansible/executor/process/result.py
@@ -16,7 +16,7 @@
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 
 # Make coding more python3-ish
-from __future__ import (absolute_import, division, print_function)
+from __future__ import (absolute_import, division)
 __metaclass__ = type
 
 from six.moves import queue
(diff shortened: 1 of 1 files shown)


bcoca:
closing in favor of changes by @jimi-c

PR #11495, opened 2015-07-05
docsite/rst/intro_configuration.rst: reword Title.

xmj:
Make Configuration the first word, so that it is in line with other documents
and that system administrators/devops people don't lose the tab when having
many browser tabs open.

Discussed wording with bcoca on IRC.

### DIFF
--- a/docsite/rst/intro_configuration.rst
+++ b/docsite/rst/intro_configuration.rst
@@ -1,5 +1,5 @@
-The Ansible Configuration File
-++++++++++++++++++++++++++++++
+Configuration file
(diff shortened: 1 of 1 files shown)


amenonsen:
I don't think the rationale for this is especially compelling, but it's a trivially correct one-liner, and if @bcoca thought it was a good idea, then it's ready to merge.

amenonsen:
I don't think the rationale for this is especially compelling, but it's a trivially correct one-liner, and if @bcoca thought it was a good idea, then it's ready to merge.

PR #11502, opened 2015-07-06
Add no_syslog option in order to prevent ansible syslog its actions on the remote host

lberruti:
Add no_syslog option in order to prevent ansible syslog its actions
on the remote host.

Syslog on remote hosts can be turned off by:
- no_syslog option inside ansible.cfg.
- -n --no_syslog command line options (ansible and ansible-playbook).

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -335,7 +335,7 @@ def heuristic_log_sanitize(data):
 
 class AnsibleModule(object):
 
-    def __init__(self, argument_spec, bypass_checks=False, no_log=False,
+    def __init__(self, argument_spec, bypass_checks=False, no_log=False, no_syslog=False,
         check_invalid_arguments=True, mutually_exclusive=None, required_together=None,
         required_one_of=None, add_file_common_args=False, supports_check_mode=False,
         required_if=None):
@@ -350,6 +350,7 @@ class AnsibleModule(object):
         self.supports_check_mode = supports_check_mode
         self.check_mode = False
         self.no_log = no_log
+        self.no_syslog = no_syslog
         self.cleanup_files = []
 
         self.aliases = {}
@@ -365,7 +366,7 @@ class AnsibleModule(object):
 
         self.params = self._load_params()
 
-        self._legal_inputs = ['_ansible_check_mode', '_ansible_no_log']
+        self._legal_inputs = ['_ansible_check_mode', '_ansible_no_log', '_ansible_no_syslog']
 
         self.aliases = self._handle_aliases()
 
@@ -373,8 +374,9 @@ class AnsibleModule(object):
             self._check_invalid_arguments()
         self._check_for_check_mode()
         self._check_for_no_log()
+        self._check_for_no_syslog()
 
-        # check exclusive early 
+        # check exclusive early
         if not bypass_checks:
             self._check_mutually_exclusive(mutually_exclusive)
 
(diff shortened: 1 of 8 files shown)


bcoca:
so we already have a no_log tag that seems to overlap with this

lberruti:
The behaviours are different:
- no_log hides contents of _sensible_ data on control machine (see http://docs.ansible.com/faq.html#how-do-i-keep-secret-data-in-my-playbook), i.e.

```
# control machine:
TASK: [command sleep 1 chdir=/tmp] ********************************************
ok: [test001] => {"censored": "results hidden due to no_log parameter"}
```
- no_syslog doesn't touch information visualized on the control machine, but it removes all syslog lines on the target machine. i.e no more lines lines like

```
# target / remote machine:
Jul  6 15:37:22 test001 ansible-command: Invoked with creates=None executable=None chdir=/tmp args=sleep 1 removes=None NO_LOG=None shell=False warn=True
```

bcoca:
no_log also prevents logging to syslog of that sensitive data, if used at the task level it should also prevent all syslog entries for that task on the remote machine.

lib/ansible/module_utils/basic.py
```diff
@@ -399,7 +401,7 @@ def __init__(self, argument_spec, bypass_checks=False, no_log=False,
             self._check_required_if(required_if)
 
         self._set_defaults(pre=False)
-        if not self.no_log:
+        if not (self.no_log or self.no_syslog):
```
bcoca, line 404: here where you prevent logging with no_syslog, no_log already did the same

lberruti:
It seems that no_log has been changed from stable to devel:

```
# stable-1.9
TASK: [secret task] ***********************************************************
changed: [deb7.inter.nos] => {"censored": "results hidden due to no_log parameter", "changed": true, "rc": 0}
# devel
TASK [secret task] **************************************************************
changed: [deb7.inter.nos] => {"changed": true, "end": "2015-07-06 17:35:58.424702", "stdout": "value=secret_to_hide", "cmd": "echo value=secret_to_hide", "start": "2015-07-06 17:35:58.423016", "delta": "0:00:00.001686", "stderr": "", "rc": 0, "invocation": {"module_name": "command", "module_args": {"_raw_params": "echo value=secret_to_hide", "_ansible_no_log": true, "_uses_shell": true}}, "stdout_lines": ["value=secret_to_hide"], "warnings": []}
```

Which one is the "official" behaviour?

The purpose of this pr is to turn off all kind of "ansible syslog" on remote machines (even syslog of "setup tasks") via ansible.cfg  (or via command line switch level)

I have added a new option instead to modify no_log because:
1. My understanding of no_log is: "used to censor sensible data (also) on control machine"
2. I want to leave untouched output on controlo machine

bcoca:
1) yes but s/sensible/sensitive/
2) no_syslog then might want to turn into no_remote_syslog

lberruti:
1) oops! Sorry :)
2) ok to rename: I will do it tomorrow

bcoca:
let me bring this up with rest of core first

amenonsen:
Any verdict?

lberruti:
I can make it works again
Please let me known if the core prefer no_remote_syslog or no_syslog

amenonsen:
I think `no_syslog` is fine; we don't syslog locally anyway, so…

lberruti:
I rebased and submitted this as #12140, this PR can now be closed.

amenonsen:
@lberruti it's your own PR, you can close it yourself. :-)

lberruti:
@amenonsen ok :)

PR #11511, opened 2015-07-07
syslog_json call plugin improvements

galindro:
- Fix NameError: global name 'handler' is not defined
- Update log message format


galindro:
tks

PR #11512, opened 2015-07-07
Add complex_args to logging callback data

DazWorrall:
Callback plugins don't get given any complex module arguments on task invocation, this fixes that.

### DIFF
--- a/v1/ansible/runner/__init__.py
+++ b/v1/ansible/runner/__init__.py
@@ -1078,7 +1078,8 @@ class Runner(object):
 
             result.result['invocation'] = dict(
                 module_args=module_args,
-                module_name=module_name
+                module_name=module_name,
+                module_complex_args=complex_args,
             )
 
             changed_when = self.module_vars.get('changed_when')


jimi-c:
Hi @DazWorrall, the code in question is now out of date and is no longer used. Thanks for the submission though!

bcoca:
v1 is there for convenience to give rapid switch-over to test if a problem was introduced by the v2 engine, it will not be released so it makes no sense to update it.

DazWorrall:
Is it no longer possible to get fixes into 1.x?

jimi-c:
@DazWorrall in some rare cases, we may accept patches to the stable-1.9 branch, if they are bug fixes only. We are planning on doing a 1.9.3 release in a couple of weeks, but generally there is no guarantee that a patch there would make it into an official release once the next release is made.

Can you please describe the bug this is causing?

DazWorrall:
We've written a custom CallbackModule to log centrally and in detail what actions were performed on a host, but we use complex arguments with many tasks and so currently we're not able to capture what we need to audit changes/errors of playbook runs.

jimi-c:
@DazWorrall go ahead and resubmit this patch against the stable-1.9 branch, and we'll consider it for the 1.9.3 release.

Thanks!

DazWorrall:
Done in #11522, thanks!

PR #11518, opened 2015-07-07
test: add base path role test

resmo:
I hit an issue in v1 and tested it in v2, which **works** as expected, however I didn't find a test for it so I added one. 

v1 output:

```
make test_role_files_path 
ansible-playbook test_role_files_path.yml -i inventory

TASK: [test_role_files_path | copy a file from a task in a subdir] ************ 
fatal: [localhost] => input file not found at /home/resmo/Projects/ansible/test/integration/roles/test_role_files_path/tasks/files/test_role_files_path_file or /home/resmo/Projects/ansible/test/integration/test_role_files_path_file

           to retry, use: --limit @/home/resmo/test_role_files_path.retry

localhost                  : ok=0    changed=0    unreachable=1    failed=0 
```

Beside that I wanted to add `ignore_errors: yes` on the copy task to just ignore errors if src file was not found and fail in the assert task. I didn't find a way to skip the error in the copy task if src file was not found. Seems like another issue.


amenonsen:
Looks good to me. (I think this was fixed relatively recently.)

amenonsen:
@resmo I'm not sure it's consistent with ansible practice elsewhere for «copy src=nonexistent» to be an ignorable error. Maybe it (and possibly other similar cases) should be, but I think that's a discussion for the `ansible-devel` mailing list.

PR #11520, opened 2015-07-08
Handle 'include' tasks in handlers

pilou-:
`include` tasks in handlers must be handled : they were allowed in 1.9.


pilou-:
Pull request updated:
- handlers defined inside a playbook are also supported;
- add tests (`include` used by handler defined at playbook level and `include` used by handler defined at role level)

jimi-c:
HI @pilou-, I'm looking at fixing this a different way, to match how includes are done in the strategies. As your PR stands, the include would revert to being a "pre-processor" kind of statement (as it was in v1), which is not how includes work now.

We are interested in the integration tests for this however, so if you'd like to split those out and send them as a separate PR we'd appreciate it.

Thanks!

PR #11522, opened 2015-07-08
Add complex_args to logging callback data

DazWorrall:
Following on from #11512: Callback plugins don't get given any complex module arguments on task completion, this fixes that.

### DIFF
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -1078,7 +1078,8 @@ class Runner(object):
 
             result.result['invocation'] = dict(
                 module_args=module_args,
-                module_name=module_name
+                module_name=module_name,
+                module_complex_args=complex_args,
             )
 
             changed_when = self.module_vars.get('changed_when')


abadger:
Description of problem being solved from #11512:

> We've written a custom CallbackModule to log centrally and in detail what actions were performed on a host, but we use complex arguments with many tasks and so currently we're not able to capture what we need to audit changes/errors of playbook runs.

bcoca:
this is a feature, which we do not normally put into stable as we try to limit minor releases to bug fixes.

DazWorrall:
Adding in the extra param is maybe a feature, but I believe its the quickest and safest way to address the bug of module_args being incomplete.

bcoca:
will discuss with rest of team if this is a additional feature or a bug of 'missing feature', I can see it both ways ....

amenonsen:
No more complex_args and no more runner, ergo no more need for this PR?

DazWorrall:
This is a fix for 1.x

PR #11528, opened 2015-07-08
Added PowerPC support to facts

jasonguy:
Added support to get the vendor and model information for PowerPC processor/architecture.


amenonsen:
@jasonguy your change looks reasonable (I'm taking it for granted that it works, since I can't test), but that `if` is looking a bit unwieldy. How about defining an array with acceptable keys (each with a comment next to it) and just checking `if key in cpu_keys`?

PR #11531, opened 2015-07-08
document `jsonfile` fact cache

tamsky:


bcoca:
actually both were available at the same time, but redis was only one that was well tested

tamsky:
Then the statement in the original is false as well, and deserves correction?

> At the time of writing, Redis is the only supported fact caching engine.

Would you like my commit/PR altered to reflect this new (to me) information?

bcoca:
nah, just 'Currently Ansible ships with 2 persistent cache plugins: redis and jsonfile'

tamsky:
updated commit, reorganized it to flow more logically, and fixed the ansible.cfg example.

PR #11533, opened 2015-07-09
Cb and contrib

bcoca:
a bunch of changes to callbacks
moved inventory scripts to contrib dir and made setup.py know about em.


setup.py
```diff
@@ -25,7 +25,7 @@
       package_dir={ '': 'lib' },
       packages=find_packages('lib'),
       package_data={
-         '': ['module_utils/*.ps1', 'modules/core/windows/*.ps1', 'modules/extras/windows/*.ps1'],
+         '': ['module_utils/*.ps1', 'modules/core/windows/*.ps1', 'modules/extras/windows/*.ps1', 'contrib/README.md', 'contrib/inventory/*'],
```
abadger: I think we want these listed in MANIFEST.in, not in setup.py's package_data.    I think package_data is for thing that are going to end up in the site-packages/ansible/ directory.  We don't want the contrib/ stuff to end up there as they're more like helper scripts.

bcoca: fixed

abadger:
I've been trying to make plugins into ABCs as I go along (abstract base classes).  If I understand how callback plugins work, though, it probably doesn't gain us much.  ABCs primarily let you create abstractmethods that someone who implements the ABC must implement otherwise they'll get an error.  I think that with callback plugins we want the base class to document the potential callback methods.  A custom callback method _may_ override any of those methods but isn't obligated to override any of them if it doesn't want to.  So we really don't gain much.

If that's not the case, point me at the methods that are mandatory in the callback plugin and I'll look at making it an ABC.

bcoca:
if you look at the example callbacks ... there are no mandatory methods. All of them used to be, but it was because they did not use a base class, they were all 'def method\n pass'

abadger:
Re: abcs, yep, that was my thought too.

Re: manifest, also remember to remove the `recursive-include plugins *` since it looks like everything has moved from plugins/ to somewhere else.

PR #11537, opened 2015-07-09
assert password or ssh key provided on new image creation

chrismeyersfsu:
- grab azure credentials from env by default in example credential template
- Remove environmental variables that define azure credentials when performing undefined credentials test
- add test for no password or ssh key early fail


chrismeyersfsu:
Related PR https://github.com/ansible/ansible-modules-core/pull/1696

PR #11540, opened 2015-07-09
Change the output of ansible --list-hosts to be separated by spaces

tobiaspal:
Change the output of ansible --list-hosts to be separated by spaces instead of newlines

This way it can be readily used in `.clusterssh/config`:

```
external_cluster_command=/path/to/ansible/bin/ansible --list-hosts
```


bcoca:
so the main reason for this feature is so people can read the list of hosts, changing the format makes it less readable.

also this is trivial to change with a shell pipe for your needs:
`ansible --list-hosts all|paste -s -` or use sed or tr, etc

PR #11560, opened 2015-07-12
ability to specify any combination of EC2 instance states to return

jasonarewhy:
This addition to the ec2.py inventory script allows for specifying any valid combination of EC2 instance states to return, rather than just all or 'running.' This introduces a new, optional configuration option in ec2.ini called `instance_states`. The default state is still 'running' and if `all_instances` is True, `instance_states` will be ignored, making this backward-compatible.

As an example, we have run into scenarios where we'd like to retrieve inventory information about hosts in either 'running', 'stopped', or 'shutting-down' states, but not 'pending' or 'terminated.'

### DIFF
--- a/contrib/inventory/ec2.ini
+++ b/contrib/inventory/ec2.ini
@@ -58,6 +58,11 @@ route53 = False
 # 'all_instances' to True to return all instances regardless of state.
 all_instances = False
 
+# By default, only EC2 instances in the 'running' state are returned. Specify
+# EC2 instance states to return as a comma-separated list. This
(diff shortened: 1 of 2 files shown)


amenonsen:
This looks like a useful change. I think it should be merged.

majidaldo:
problem: in `#skip instances we cannot address`, well for instances that are not running they will not be addressable of course and the script will skip them over. ie. if i specify `instance_states=stopped` the script will not give me back that instance. obviously i want it to show up.

also even if i comment it out the groups are not populated.

PR #11561, opened 2015-07-12
Improve error message "module is missing interpreter line"

moritz:
include the module name in the error message; inspired by http://michaelheap.com/ansible-module-is-missing-interpreter-line/

### DIFF
--- a/v1/ansible/runner/__init__.py
+++ b/v1/ansible/runner/__init__.py
@@ -527,7 +527,7 @@ class Runner(object):
                 cmd = " ".join([str(x) for x in [remote_module_path, async_jid, async_limit, async_module]])
 
         if not shebang:
-            raise errors.AnsibleError("module is missing interpreter line")
+            raise errors.AnsibleError("module '%s' is missing interpreter line" % module_name)
 
         rm_tmp = None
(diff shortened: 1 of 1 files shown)


bcoca:
v1 is there for testing only, it will not be released or used aside from fast checking if a problem was caused by the v2 rewrite.

As such I'm closing this request.

MikeSpreitzer:
Huh, what?  Sorry, I am new here.  v1 of what?  I am also getting that error message, and am also mystified by it.  I could really use that clue.

jimrazmus:
I am encountering the same message using the latest devel branch.  It would be nice to get this resolved in v2.x and up.

MikeSpreitzer:
@bcoca are you saying that version 1.\* of Ansible is only for testing, only version 2 of Ansible is being maintained?  That seems a bit harsh.

amenonsen:
There was a directory named "v1" in the devel branch, which contained a copy of the v1 source for testing, and caused constant confusion because people would submit pull requests against it. It has since been removed.

MikeSpreitzer:
OK, this PR was indeed misdirected.  There is still a problem in the real code.  I will open an issue for it.

MikeSpreitzer:
I opened issue #13816 for the problem still in the code.

MikeSpreitzer:
I created PR #13817 to fix it right.

PR #11567, opened 2015-07-13
Add shelvefile lookup plugin

lekum:
Add lookup plugin to read string values of keys from a [shelve](https://docs.python.org/2/library/shelve.html#module-shelve) file.

Usage example:

```

---
- hosts: localhost
  vars:
    book: "{{ lookup('shelvefile', 'file=book.db key=current_book_name') }}"
    page: "{{ lookup('shelvefile', 'file=book.db key=current_book_page') }}"
    author: "{{ lookup('shelvefile', 'file=book.db key=current_book_author') }}"

  tasks:
    - name: Print my current reading
      debug: msg="I'm on page {{ page }} of '{{ book }}' by {{ author }}"
```

### DIFF
--- a/lib/ansible/plugins/lookup/shelvefile.py
+++ b/lib/ansible/plugins/lookup/shelvefile.py
@@ -25,8 +25,6 @@ from ansible.plugins.lookup import LookupBase
 
 class LookupModule(LookupBase):
 
-    def __init__(self, basedir=None, **kwargs):
-        self.basedir = basedir
 
     def read_shelve(self, shelve_filename, key):
         """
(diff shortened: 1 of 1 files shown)


bcoca:
lookup plugins have changed, this won't work with the current version.

we are preparing docs to show how to create plugins that will work with both 1.x and 2.x versions of ansible.

lekum:
It's based upon the code of the [file lookup](https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/lookup/file.py) of the `devel` branch. Where can I found the actual plugins that work with the `devel` version?

bcoca:
look at the current version of the file lookup, it is pretty different than it was

lekum:
Which one is the current one?

```
./v1/ansible/runner/lookup_plugins/file.py
./lib/ansible/plugins/lookup/file.py
```

bcoca:
v1/ is there only for testing, that is dead code that will be removed, so of course, the other one

lekum:
I changed it according to the new file lookup plugin

lekum:
Is there anything I could do to help with this?

PR #11576, opened 2015-07-13
Do not combine group_vars with an empty file

srgvg:
This addresses a specific case with multiple vars files
in a group_vars/${groupname}/ directory where one of those files
is empty, which returns None instead of an empty dict.

This then triggers an error in the combine_vars utility function.


srgvg:
partly discussed with @jimi-c

PR #11587, opened 2015-07-14
ansible-galaxy - add -L / --extract-symlinks option

vamme:
bin/ansible-galaxy - add -L / --extract-symlinks option that will extract symlinks (duh)

I'm not neglecting security implications, that's why this option is disabled by default. Of course it needs to be disabled in order to preserve current default behaviour. 

In my use case, where ansible-galaxy downloads files from local git repos only, it's perfectly valid to have it preserve elaborate config structures in "files" that contain symlinks.

### DIFF
--- a/bin/ansible-galaxy
+++ b/bin/ansible-galaxy
@@ -192,6 +192,9 @@ def build_option_parser(action):
         parser.add_option(
             '-r', '--role-file', dest='role_file',
             help='A file containing a list of roles to be imported')
+        parser.add_option(
+            '-L', '--extract-symlinks', dest='extract_symlinks', action='store_true', default=False,
+            help='Recreate symlinks as well as regular files')
     elif action == "remove":
         parser.set_usage("usage: %prog remove role1 role2 ...")
     elif action == "list":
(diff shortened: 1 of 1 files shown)


bcoca:
we don't take feature request against non /devel branches, please reissue this request against current devel. 

FYI this code has changed a lot compared to this version.

vamme:
Hi,

In devel (2.0) links are already unconditionally extracted (no option
needed) and you don't take prs for v1 dir in devel either (as per pr
11588).

I don't know we'll be able to migrate to 2.0 straight away and this is
breaking internal roles for me, so I guess I can stop using galaxy for this
or use a fork.

Regards,
Daniel

bcoca:
Sorry, but i'm not sure that accepting this patch would help you either as you would need to wait for the next release, if it happens, of 1.9.

v1 in devel is 'dead code', its only there as a convenient way to test if a bug is pre-existing, it will not be released and will soon be deleted.

if current devel works for you, we plan to release it at the begining of next month

vamme:
Yeah, I understand, that's fine.

I'll try devel to make sure it's doing what I need, but I won't really be
able to use it in production until 2.1 ;-). A local version of
ansible-galaxy will probably do the trick.

Thanks for taking a look anyway.

Regards,
Daniel

On 14 July 2015 at 20:31, Brian Coca notifications@github.com wrote:

> Sorry, but i'm not sure that accepting this patch would help you either as
> you would need to wait for the next release, if it happens, of 1.9.
> 
> v1 in devel is 'dead code', its only there as a convenient way to test if
> a bug is pre-existing, it will not be released and will soon be deleted.
> 
> if current devel works for you, we plan to release it at the begining of
> next month
> 
> —
> https://github.com/ansible/ansible/pull/11587#issuecomment-121349537.

## 

Regards,
Daniel Siechniewicz
nullDowntime Ltd
075 3497 5174
daniel@nulldowntime.com

PR #11588, opened 2015-07-14
* v1/bin/ansible-galaxy - add -L / --extract-symlinks option

vamme:
bin/ansible-galaxy - add -L / --extract-symlinks option that will extract symlinks (duh)

I'm not neglecting security implications, that's why this option is disabled by default. Of course it needs to be disabled in order to preserve current default behaviour.

In my use case, where ansible-galaxy downloads files from local git repos only, it's perfectly valid to have it preserve elaborate config structures in "files" that contain symlinks.

This is in line with pull request #11587.

From what I can see in lib/ansible/galaxy/role.py (2.0 stuff) it's just extracting links, it's not even optional.

### DIFF
--- a/v1/bin/ansible-galaxy
+++ b/v1/bin/ansible-galaxy
@@ -192,6 +192,9 @@ def build_option_parser(action):
         parser.add_option(
             '-r', '--role-file', dest='role_file',
             help='A file containing a list of roles to be imported')
+        parser.add_option(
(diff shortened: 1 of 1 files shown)


jimi-c:
Hi @nulldowntime, we are currently not taking any PRs for the v1 directory code, it's purpose was as an emergency fall-back only as we transitioned to the v2 code and will be deleted soon. We will consider your PR for the mainline, so closing this.

Thanks!

PR #11591, opened 2015-07-14
Connection function for boto3

juliedavila:
This is adding a simple connection function for boto3.

Inside of a module's code, it would be used like:

``` python
region, ec2_url, aws_connect_kwargs = get_aws_connection_info(module, boto3=True)

ec2_client, ec2_res = boto3_conn(module, conn_type='both', resource='ec2', region=region, endpoint=ec2_url, **aws_connect_kwargs)

all_vpcs = ec2_client.describe_vpcs()
```

There are two noticeable pieces that make it functionally different than the current connection functions:
- `conn_type`: refers to the type of session you want, whether a high level resource session or a low level client session, or both.  The boto3 docs regarding sessions are [here](http://boto3.readthedocs.org/en/latest/reference/core/session.html)
- `resource`: this is the AWS resource you want to communicate such as ec2, s3, autoscaling, etc. Full list is [here](http://boto3.readthedocs.org/en/latest/reference/services/index.html)

This should allow everyone to start using boto3 for module development/enhancement right away.

### DIFF
--- a/lib/ansible/module_utils/ec2.py
+++ b/lib/ansible/module_utils/ec2.py
@@ -46,6 +46,19 @@ AWS_REGIONS = [
     'us-gov-west-1',
 ]
 
+def boto3_conn(module, conn_type=None, resource=None, region=None, endpoint=None, **params):
+    if conn_type not in ['both', 'resource', 'client']:
+        module.fail_json(msg='There is an issue in the code of the module. You must specify either both, resource or client to the conn_type parameter in the boto3_conn function call')
+
+    resource = boto3.session.Session().resource(resource, region_name=region, endpoint_url=endpoint, **params)
+    client = resource.meta.client
+
+    if conn_type == 'resource':
+        return resource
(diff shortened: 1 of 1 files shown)


gregdek:
ping for comment: @jsmartin @scicoin-project @tombamford @garethr @lorin @jarv @jsdalton @silviud @adq @zbal @zeekin @willthames @lwade @carsongee @defionscode @tastychutney @bpennypacker @loia

bpennypacker:
Do we want to be calling sys.exit if parameter validation to boto3_conn fails, or should we handle it some other way?  It looks like if the call to boto3.session.Session fails it will just raise an exception, so perhaps raising a generic exception instead of exiting would be more appropriate.  Other than that this all looks good to me.

bcoca:
no calling exit in modules! fail_json

juliedavila:
@bpennypacker and @bcoca, done

PR #11595, opened 2015-07-15
ssh port not working as the if clause is wrong

toshywoshy:
The ssh port always results in port 22 as the if clause is never executed, this means that hosts on non-standard ssh port cannot be accessed

### DIFF
--- a/lib/ansible/plugins/connections/ssh.py
+++ b/lib/ansible/plugins/connections/ssh.py
@@ -104,7 +104,7 @@ class Connection(ConnectionBase):
         if not C.HOST_KEY_CHECKING:
             self._common_args += ("-o", "StrictHostKeyChecking=no")
 
-        if self._connection_info.port is not None:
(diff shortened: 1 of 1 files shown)


bcoca:
That sets the port to None, i believe the issue is elsewhere setting the connection port to 22 always.

toshywoshy:
I think I found it now : If the remote_port is set in ansible.cfg, this overrides all other settings in the original version and only port 22 can be used, if no remote port the code works
I my opinion and given that the documentation and backward compatibility should allow a general port to be defined in ansible.cfg and override the port if this port is 'None' on the specific host level
As per documentation it should be possible to set a general remote_port and set a specific host port, the patch allows the traditional and new options

bcoca:
The patch just does this:

```
'Port=None'
```

halberom:
probably related to #11616

bcoca:
closing this in favor of #11616

PR #11603, opened 2015-07-15
Have openssl autonegotiate tls protocol on python < 2.7.9

abadger:
This allows usage of tls-1.1 and tls-1.2 if the underlying openssl
library supports it.  Unfortunately it also allows sslv2 and sslv3 if
the server is only configured to support those.  In this day and age,
that's probably something that the server administrator should fix
anyhow.

### DIFF
--- a/lib/ansible/module_utils/urls.py
+++ b/lib/ansible/module_utils/urls.py
@@ -106,6 +106,33 @@ try:
 except ImportError:
     HAS_SSLCONTEXT = False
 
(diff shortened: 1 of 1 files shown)


bcoca:
updated #11603 which allows for user config of the default

abadger:
Discussed this on slack and decided that this was okay for now.  need to add a param to fetch_url() to allow overrides on a per-task basis.

PR #11606, opened 2015-07-16
using loops: notify didn't work when the notify value is a single string (and not a list)

pilou-:
[Task.post_validate](https://github.com/pilou-/ansible/blob/notify_with_items/lib/ansible/playbook/base.py#L242) was called for each [included task](https://github.com/pilou-/ansible/blob/notify_with_items/lib/ansible/executor/task_executor.py#L238) but [wasn't](https://github.com/pilou-/ansible/blob/notify_with_items/lib/ansible/executor/task_executor.py#L164) for the 'include' task. post_validate needs to be called because this method [modifies](https://github.com/pilou-/ansible/blob/notify_with_items/lib/ansible/playbook/base.py#L290) the task, for example the 'notify' attribute becomes a list if it's a string.

Without this patch, using an handler named 'test', [`_notified_handlers`](https://github.com/pilou-/ansible/blob/notify_with_items/lib/ansible/plugins/strategies/__init__.py#L403) looks like `{u'test': [], u's': [testhost], u'e': [testhost], u't': [testhost]}` when it should be `{u'test': [testhost]}`.

### DIFF
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -148,6 +148,9 @@ class TaskExecutor:
         # the variables and re-validate the task with the item variable
         task_vars = self._job_vars.copy()
 
+        templar = Templar(loader=self._loader, shared_loader_obj=self._shared_loader_obj, variables=task_vars)
+        self._task.post_validate(templar=templar)
+
         items = self._squash_items(items, task_vars)
         for item in items:
             task_vars['item'] = item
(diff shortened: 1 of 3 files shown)


jimi-c:
Hi @pilou- , thanks for your submission however I've decided to fix this another way (SHA above).

Thank you!

pilou-:
Hi @jimi-c, would you consider a PR which only adds the tests (c9956cc) ?

bcoca:
we welcome the tests, though it might make more sense to put in exissting integration/roles/test_handlers

PR #11609, opened 2015-07-16
Fix #11369 A result is skipped when all it's children are skipped. Th…

### LINKED ISSUE #11369: set_fact skipping
The following code works under ansible 1.9 but fails on the latest devel checkout. It runs the same rules, but skips the set_fact command.

```
- name: Set zookeeper ID facts
  gather_facts: True
  hosts: zookeeper
  user: username
  sudo: True
  tasks:
    - set_fact: zkid={{ item.0 | int + 1 }}
      when: hostvars[item.1]['ansible_hostname'] == ansible_hostname
      with_indexed_items: groups['zookeeper']
```

This assigns a unique id to each zookeeper instance dynamically.

ofpiyush:
…is makes it fundamentally different from a result that was changed/failed/unreachable

### DIFF
--- a/lib/ansible/executor/task_result.py
+++ b/lib/ansible/executor/task_result.py
@@ -40,7 +40,14 @@ class TaskResult:
         return self._check_key('changed')
 
     def is_skipped(self):
-        return self._check_key('skipped')
+        if 'results' in self._result:
+            flag = True
+            for res in self._result.get('results', []):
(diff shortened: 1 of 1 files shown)


sheldonh:
Fixes #11369.

jimi-c:
Merged, thanks!

PR #11618, opened 2015-07-16
hacking/test-module, updated to new location and non-classness of module_common

halberom:
before change

```
$ hacking/test-module -m ~/dev/ansible-modules-core/system/ping.py 
  File "hacking/test-module", line 40, in <module>
    import ansible.module_common as module_common
ImportError: No module named module_common
```

after change

```
$ hacking/test-module -m ~/dev/ansible-modules-core/system/ping.py 
* including generated source, if any, saving to: /Users/glynch/.ansible_module_generated
* this may offset any line numbers in tracebacks/debuggers!
***********************************
RAW OUTPUT
{"changed": false, "ping": "pong"}


***********************************
PARSED OUTPUT
{
    "changed": false, 
    "ping": "pong"
}
```

### DIFF
--- a/hacking/test-module
+++ b/hacking/test-module
@@ -37,7 +37,7 @@ import optparse
 import ansible.utils as utils
 from ansible.parsing.utils.jsonify import jsonify
 from ansible.parsing.splitter import parse_kv
-import ansible.module_common as module_common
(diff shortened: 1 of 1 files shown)


bcoca:
nice, i was about to tackle this

PR #11631, opened 2015-07-17
GCE tag prefix for creating ansible group.

athoune:
When you migrate your old hosting to GCE, you want to keep your beloved groups. It can be done with a simple name convention : a GCE tag beginning with group-, like "group-toto" becomes a group named "toto".


bcoca:
cc @erjohnso for review

erjohnso:
LGTM @bcoca - feel free to merge.

PR #11637, opened 2015-07-18
Moar docs

bcoca:
Added docs for new v2 features restructures modules bar and core/extras info into their own pages changed templates to refer to the new pages

### DIFF
--- a/docsite/rst/intro_configuration.rst
+++ b/docsite/rst/intro_configuration.rst
@@ -307,7 +307,7 @@ gathering
 
 New in 1.6, the 'gathering' setting controls the default policy of facts gathering (variables discovered about remote systems).
 
-The value 'implicit' is the default, meaning facts will be gathered per play unless 'gather_facts: False' is set in the play and fact caching will be ignored.
+The value 'implicit' is the default, which means that the fact cache will be ignored and facts will be gathered per play unless 'gather_facts: False' is set.
 The value 'explicit' is the inverse, facts will not be gathered unless directly requested in the play.
 The value 'smart' means each new host that has no facts discovered will be scanned, but if the same host is addressed in multiple plays it will not be contacted again in the playbook run.
 This option can be useful for those wishing to save fact gathering time. Both 'smart' and 'explicit' will use the fact cache.
--- a/docsite/rst/intro_configuration.rst
+++ b/docsite/rst/intro_configuration.rst
@@ -838,5 +838,5 @@ These are settings that control SELinux interactions.
 .. versionadded:: 1.9
 
 This is a list of file systems that require special treatment when dealing with security context.
-The normal behaviour is for operations to copyc the existing context or use the user default, this changes it to use a file system dependent context.
+The normal behaviour is for operations to copy the existing context or use the user default, this changes it to use a file system dependent context.
 The default list is: nfs,vboxsf,fuse,ramfs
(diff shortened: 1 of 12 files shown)


docsite/rst/intro_configuration.rst
```diff
@@ -144,6 +144,27 @@ different locations::
...
+==================
+
+.. versionadded:: 2.0
+
+Now ansible ships with all included callback plugins ready to use but they are disabled by default,
+this setting lets you enable a list of additional callbacks, this cannot change or override the
+default stdout callback, use :ref:`stdout_callback` for that.
+
+
 .. _command_warnings:
```
abadger, line 168: Maybe an example for both of these.  Is stdout_callback just a True|False to enable or disable it?  Is callback_whitelist just a comma separated string of plugin names?  Do the plugin names end in ".py"?

docsite/rst/intro_configuration.rst
```diff
@@ -802,3 +824,19 @@ If enabled, this setting allows multiple private keys to be uploaded to the daem
...
 
+.. _selinux_settings:
+
+Selinux Specific Settings
+-------------------------
+
+These are settings that control SELinux interactions.
+
+
+#special_context_filesystems
```
abadger: Is there supposed to be a leading hash mark here?

docsite/rst/modules_intro.rst
```diff
@@ -0,0 +1,64 @@
+Introduction
+============
+
+Modules (also reffered to as "task plugins" or "library plugins") are the ones that do
```
abadger: s/reffered/referred/

docsite/rst/playbooks_strategies.rst
```diff
@@ -0,0 +1,39 @@
+Strategies
+===========
+
+In 2.0 we added a new way to control play execution, ``strategy``, by default plays will
+still run as they used to, with what we call the ``linear`` strategy. All hosts will run each
+task befor any host starts the next task, using the number of forks (default 5) to parallelize.
+
```
abadger, line 7: s/befor/before/

docsite/rst/playbooks_strategies.rst
```diff
@@ -0,0 +1,39 @@
...
+The ``serial`` directive can 'batch' this behaviour to a subset of the hosts, which then run to
+completion of the play before the next 'batch' starts.
+
+A second ``strategy`` ships with ansible ``free``, which allows each host to run until the end of
+the play as fast as it can.::
+
```
abadger, line 13: This section seems like it is sufficient for people who are already familiar with ansible-1.x's parallelism but needs some thing for people who are new.  It seems like it would be easy for people to miss that linear is parallel within a task and that serial is related to that within-a-play parallelism.

free's parallelism could then be defined in contrast to that.  I'm also not sure what effect serial might have if the strategy was free.  Is that an error?  Does it force linear strategy behaviour?

docsite/rst/intro_configuration.rst
bcoca, line 168: i put examples commented out in examples/ansible.cfg

docsite/rst/playbooks_strategies.rst
bcoca, line 13: agreed, this was mostly a first draft, I expect james to clarify

Issue #9526, opened 2014-11-10
file operations fail on shared/synced folders that do not implement correct filesystem semantics


dopry:
##### Issue Type:

“Bug Report”, 
##### Ansible Version:

1.7.2
##### Environment:

Ubuntu 14.04.1 VM on Virtual Box 4.3.14  on Windows 8.1 
I know the issue is likely specific to vboxfs 
##### Summary:

The copy modules fails on virtual box shared folders. specifically the os.rename() fails.  I tried replacing the os.rename with shutils.copy2 and it worked fine.  

I notice there is a lot of working being done in atomic_move. There is a fallback strategy that seems to attempt to copy the source file to a temp file in the destination directory, and then rename the temp file. Why not just do the copy straight to the destination and let cleanup() handle the stuff in the temp folder?

I tested a few scenarios if I try to copy to some place besides the shared folder, the module works. So the source doesn't appear to be what is blocking the rename, but it's the destination. I tried hard coding an alternate destination at the os.rename, in case a previous operation was taking a lock, but still kept receiving error 26.

When I try opening up the repl, I get a slightly different error, but an error none the same..

> > > import os
> > > os.rename('test.txt', '/vagrant/test.txt')
> > > Traceback (most recent call last):
> > >   File "<stdin>", line 1, in <module>
> > > OSError: [Errno 18] Invalid cross-device link

It looks like os.rename doesn't work across partitions, and this is the root of the problem. (maybe this is why it tries to copy to the save folder as the destination, then renames).

https://mail.python.org/pipermail/python-list/2005-February/342893.html
https://bugs.launchpad.net/python-keystoneclient/+bug/1265190
##### Steps To Reproduce:

install virtualbox. 

put all the files from  https://gist.github.com/dopry/4bc4f2d852b8055a7b68 in a folder.

vagrant up 
vagrant ssh 

from inside the vm..

echo "test" > test.txt
ansible web -i /vagrant/ansible/inventory/vagrant/hosts.py -m copy -a "src=test.txt dest=/vagrant/test.txt"
##### Expected Results:

test.txt is copied to  /vagrant.
##### Actual Results:

vagrant.dev | FAILED >> {
    "failed": true,
    "md5sum": "bbfa1f311a5828452b953d1335cbf027",
[...]

PR #11643, opened 2015-07-18
Meta inventory and fixes

bcoca:
added meta: refresh_inventory
cleaned up AnsibleError to match general style
now dir and single file share same logic on dealing with files
if executable fails, we try ini aftewards and only then really fail
inventory ignored files are now configurable
added note to plugins/inventory that it is not currently used code

fixes  #8830,  #10068, #10369 and some cases of #9526

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ Major Changes:
  * template code now retains types for bools and numbers instead of turning them into strings.
    If you need the old behaviour, quote the value and it will get passed around as a string
  * Consolidated code from modules using urllib2 to normalize features, TLS and SNI support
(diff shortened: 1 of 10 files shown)


abadger:
Probably need to integrate new_inventory along with this.  Old inventory doesn't utilise plugins/inventory

bcoca:
only added a readme for the new plugins to note that they are not currently in use, I did not think it would be ready for 2.0 so I pushed the changes to the currently running code.

abadger:
Okay, no objections from me but you might want @jimi-c to give it a once over as well.  I'v forgotten everything I knew about how current inventory works ;-)

PR #11657, opened 2015-07-20
add sshpass support for `synchronize`

elvismacak:
add sshpass support for rsync
work together with https://github.com/ansible/ansible-modules-core/pull/1755


amenonsen:
This PR should have been closed along with #12704 and ansible/ansible-modules-core#2247 when #13099 was opened. I'm closing it now, because the problem and a suitable approach need to be discussed in #13099 before any changes are made in this area.

PR #11662, opened 2015-07-20
fix for configs getting ignored if a environment variable is supplied…

### LINKED ISSUE #11599: Synchronize module doesn't respecting ssh config
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.9.2
##### Ansible Configuration:

All ssh-related by default:

```
$ grep ssh ansible.cfg
[ssh_connection]
# ssh arguments to use
#ssh_args = -o ControlMaster=auto -o ControlPersist=60s -F inventory/ssh.config
ssh_args = -o ControlMaster=auto -o ControlPersist=60s
# "%(directory)s/ansible-ssh-%%h-%%p-%%r", however on some systems with
#control_path = %(directory)s/ansible-ssh-%%h-%%p-%%r
# if True, make ansible use scp if the connection type is ssh
scp_if_ssh = True
```
##### Environment:

Fedora 22
##### Summary:

When using the synchronize module, rsync is attempting to use direct connection to the host without configured parameters like port, user, proxy.

Similar issue was reported: https://github.com/ansible/ansible/issues/6590 but then requalified only to user setting and closed.
##### Steps To Reproduce:

Playbook tasks:

```
- file: path=files/{{type}}/ state=directory

- name: Get /opt
  synchronize: >
    mode=pull src=/opt dest="files/{{type}}/" compress=yes archive=yes checksum=yes recursive=yes set_remote_user=False delete=True >
    rsync_opts="--exclude-from=roles/common/files/{{type}}.opt.rsync.exclude,--prune-empty-dirs,--delete-excluded,--log-file=logs/rsync-opt-{{inventory_hostname}}.log,--progress,--stats,--itemize-changes"
  become: no
```

Run against host egais.app.balance. That defined in inventory:

```
egais.app.balance ansible_become=yes ansible_become_method=sudo
```

In ~/.ssh/config:

```
Host egais.app.balance
  User ssh_user
  HostName 10.0.17.34
  ProxyCommand ssh egais.app.node1 -W %h:%p
  Port 1022
```

Please note port different from 22 and ProxyCommand defined because there no direct access.
##### Expected Results:

Run playbook works, because host accessible directly with that configuration:

```
$ rsync egais.app.balance:/opt
drwxr-xr-x          4,096 2015/05/15 15:27:17 opt
```
##### Actual Results:

Fails with no access by port 22:

```
$ ap -vv Balancer:GetLocalFilesFromProd.yml

PLAY [egais.app.balance] ******************************************************

TASK: [file path=files/balancer/ state=directory] *****************************
<egais.app.balance> REMOTE_MODULE file path=files/balancer/ state=directory
ok: [egais.app.balance] => {"changed": false, "gid": 0, "group": "root", "mode": "0755", "owner": "root", "path": "files/balancer/", "secontext": "unconfined_u:object_r:user_home_t:s0", "size": 6, "state": "directory", "uid": 0}

TASK: [Get /opt] **************************************************************
failed: [egais.app.balance -> 127.0.0.1] => {"cmd": "rsync --delay-updates -F --compress --delete-after --checksum --archive --rsh 'ssh  -S none -o StrictHostKeyChecking=no' --exclude-from=roles/common/files/balancer.opt.rsync.exclude --prune-empty-dirs --delete-excluded --log-file=logs/rsync-opt-egais.app.balance.log --progress --stats --itemize-changes --out-format='<<CHANGED>>%i %n%L' \"egais.app.balance:/opt\" \"/home/pasha/Projects/Ansible.cfg/files/balancer/\"", "failed": true, "rc": 255}
msg: ssh: connect to host egais.app.balance port 22: Connection timed out
rsync: connection unexpectedly closed (0 bytes received so far) [Receiver]
rsync error: unexplained error (code 255) at io.c(226) [Receiver=3.1.1]
```

kellyp:
This should resolve #11599 which is caused by python on linux returning empty string when the environment variable is not present.  The new test passed on both OS X and Ubuntu.  I verified the fix also works on Centos6.

### DIFF
--- a/test/units/test_constants.py
+++ b/test/units/test_constants.py
@@ -1,0 +1,65 @@
+# (c) 2012-2014, Michael DeHaan <michael.dehaan@gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import pwd
+import os
+
+from ansible.compat.tests import unittest
+from ansible.compat.tests.mock import patch, MagicMock
+
+from ansible import constants as C
+from ansible.cli import CLI
+from ansible.errors import AnsibleError, AnsibleParserError
+from ansible.executor.connection_info import ConnectionInformation
+
+from units.mock.loader import DictDataLoader
+
+class TestConstants(unittest.TestCase):
+
+    def setUp(self):
+        self.config_file = open('ansible.cfg', 'w')
+        self.config_file.write(
+        '''[defaults]
+system_errors = false
+[ssh_connection]
+ssh_args = -o StrictHostKeyChecking=no -o ControlPersist=15m -F /etc/ansible/ssh.config -q
+scp_if_ssh = True
+some_false_value = False
+some_other_false_value = false
+some_empty_value =
+            ''')
+        self.config_file.close()
+
+    def tearDown(self):
+        os.remove(self.config_file.name)
+
+
+    def test_get_config(self):
+        p, path = C.load_config_file()
+        self.assertEqual(path, os.getcwd() + '/ansible.cfg')
+        self.assertEqual(C.get_config(p, 'ssh_connection', 'ssh_args', 'ANSIBLE_SSH_ARGS', None), '-o StrictHostKeyChecking=no -o ControlPersist=15m -F /etc/ansible/ssh.config -q')
+        os.environ['ANSIBLE_SSH_ARGS'] = '-o StrictHostKeyChecking=yes'
+        self.assertEqual(C.get_config(p, 'ssh_connection', 'ssh_args', 'ANSIBLE_SSH_ARGS', None), '-o StrictHostKeyChecking=yes')
+        os.environ['ANSIBLE_SSH_ARGS'] = ''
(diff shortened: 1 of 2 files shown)


bcoca:
this breaks on VAR=False

kellyp:
I added a couple more conditionals for checking.  I saw that the values are getting manually converted to booleans and coming in as strings from both the yaml and environment variables it was ok?

test-requirements.txt
```diff
@@ -8,3 +8,4 @@ passlib
 coverage
 coveralls
 unittest2
+jinja2
```
bcoca: ansible already requires jinja2 this should not be needed

bcoca:
the empty string acts as a false, which was desired in some cases, we would have to test a variety of settings before merging this

kellyp:
Cool, I added a couple more conditions.  Let me know what I can do.

test/units/test_constants.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+    def test_get_config(self):
+        p, path = C.load_config_file()
+        self.assertEqual(path, os.getcwd() + '/ansible.cfg')
+        self.assertEqual(C.get_config(p, 'ssh_connection', 'ssh_args', 'ANSIBLE_SSH_ARGS', None), '-o StrictHostKeyChecking=no -o ControlPersist=15m -F /etc/ansible/ssh.config -q')
+        os.environ['ANSIBLE_SSH_ARGS'] = '-o StrictHostKeyChecking=yes'
+        self.assertEqual(C.get_config(p, 'ssh_connection', 'ssh_args', 'ANSIBLE_SSH_ARGS', None), '-o StrictHostKeyChecking=yes')
+        os.environ['ANSIBLE_SSH_ARGS'] = ''
+        self.assertEqual(C.get_config(p, 'ssh_connection', 'ssh_args', 'ANSIBLE_SSH_ARGS', None), '-o StrictHostKeyChecking=no -o ControlPersist=15m -F /etc/ansible/ssh.config -q')
```
bcoca, line 61: so this assertion looks incorrect, i would expect and empty string as the ssh args

kellyp, line 61: that is the issue, in linux, python returns an empty string when the variable isn't set.  I would think it should then fall through to the config file.

bcoca, line 61: no, if the variable is set but empty, you should empty out the options

amenonsen:
@kellyp any plans to update this?

bcoca:
no closing this as  this is not a bug but a feature to be able to 'unconfigure' some variables.

PR #11670, opened 2015-07-21
Fix intro_inventory documented params. Update for the new 'become' system

jody-frankowski:
Replaces and updates #9588

### DIFF
--- a/docsite/rst/intro_inventory.rst
+++ b/docsite/rst/intro_inventory.rst
@@ -192,8 +192,9 @@ is an excellent way to track changes to your inventory and host variables.
 List of Behavioral Inventory Parameters
 +++++++++++++++++++++++++++++++++++++++
 
-As alluded to above, setting the following variables controls how ansible interacts with remote hosts. Some we have already
-mentioned::
+As alluded to above, setting the following variables controls how ansible interacts with remote hosts.
+
+Ssh connection::
 
     ansible_ssh_host
       The name of the host to connect to, if different from the alias you wish to give to it.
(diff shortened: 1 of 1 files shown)


amenonsen:
Looks good, nice touch to include a link to the privilege escalation documentation too.

docsite/rst/intro_inventory.rst
```diff
@@ -203,16 +204,38 @@ mentioned::
       The default ssh user name to use.
     ansible_ssh_pass
       The ssh password to use (this is insecure, we strongly recommend using --ask-pass or SSH keys)
+    ansible_connection
+      Connection type of the host. Candidates are local, ssh or paramiko.  The default is paramiko before Ansible 1.2, and 'smart' afterwards which detects whether usage of 'ssh' would be feasible based on whether ControlPersist is supported.
+    ansible_ssh_private_key_file
```
bcoca, line 212: 1.2 is really old at this point, i would not even mention the change here

docsite/rst/intro_inventory.rst
```diff
@@ -203,16 +204,38 @@ mentioned::
...
+      Private key file used by ssh.  Useful if using multiple keys and you don't want to use SSH agent.
+
+Privilege escalation (see :doc:`Ansible Privilege Escalation<become>` for further details). Since Ansible 1.9 you should use *ansible_become*::
+
```
bcoca, line 216: we try to keep the intro light, so maybe just this link with a mention that other parameters exist might be a better change

jody-frankowski, line 216: If we remove ansible_su/do from there I don't see anywhere else where they are documented.

bcoca, line 216: we can adde them to the become page under 'legacy'

amenonsen:
OK, this makes sense to me too.

PR #11674, opened 2015-07-21
Hashicorp Vault lookup Plugin

juliedavila:
This is a lookup plugin to be able to read secrets from a Hashicorp Vault endpoint.

Requires `pip install hvac`

Usage is like
`lookup('hashi_vault', 'secret=secret/hello token=c975b780-d1be-8016-866b-01d0f9b688a5 url=http://myvault:8200')`

Where url is optional and defaults to localhost and can also read in from the VAULT_ADDR environment variable.

This first iteration doesn't support other authentication mechanisms other than token and also presumes the vault is unsealed prior to executing the lookup.

### DIFF
--- a/lib/ansible/plugins/lookup/hashi_vault.py
+++ b/lib/ansible/plugins/lookup/hashi_vault.py
@@ -1,0 +1,93 @@
+# (c) 2015, Jonathan Davila <jdavila(at)ansible.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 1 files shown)


lib/ansible/plugins/lookup/hashi_vault.py
```diff
@@ -0,0 +1,93 @@
...
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import os
+
+from ansible.errors import *
+from ansible.plugins.lookup import LookupBase
+
+try:
```
bcoca: this should be moved into the run method, as we only want this to error on running, not on loading, which will always happen if this is included in ansible.

or in the hasivault class init, which gets called from run.

juliedavila:
@bcoca done

PR #11677, opened 2015-07-21
fix ipv6 for ssh connection

JuiceboxSingularity:
Fixes ipv6 when using ssh


amenonsen:
Why is tempHost needed at all? Does something break if you leave self.host as '[...]'? Because if not, you could just change that one last block in **init** and be done. Also, you should at least remove the FIXME comment just above (and maybe drop the explicit comparisons to be consistent with other code, i.e. use «if self.host» and «if self._ipv6»).

JuiceboxSingularity:
SSH breaks with the square brackets. Style and comments have been changed.

amenonsen:
Hi @JuiceboxSingularity. I've been looking into your patch, and I can't see any reason not to just _always_ wrap the hostname/IP for scp and sftp in square brackets. It works for hostnames, IPv4, and IPv6 with ssh, scp, and sftp.

amenonsen:
I submitted the simpler approach described above as #11816. Please check if that works for you?

JuiceboxSingularity:
Seems good, thanks.

Issue #8637, opened 2014-08-15
Make sftp arguments configurable


sigio:
##### Summary:

I have some stepping-stone hosts where I can't change the configuration, and they don't have sftp configured.

Sftp works just fine when running 'sftp -s /usr/lib/sftp-server'. 

It would be nice if ansible had a ansible_sftp_options configurable option in the inventory where extra sftp options can be configured.

I can currently use scp_if_ssh, but this would work nicer I guess

Currently I just manually edited connection-plugins/ssh.py to add the option hardcoded. It's only in 2 lines with 'sftp':

```
        cmd += ["sftp"] + ["-s /usr/lib/sftp-server"] + self.common_args + [host]                                    
```

Please summarize your request in this space.  You will earn bonus points for being succinct, but please add enough detail so we can understand the request.  Thanks!
##### Steps To Reproduce:

ansible -m ping all

On hosts without 'subsystem sftp' configured in ssh, ansible fails.
##### Expected Results:

ansible -m ping all works

ansibot:
Thanks for filing a ticket! I am the friendly GitHub Ansibot. I see you did not fill out your issue description based on our new issue template. Please copy the contents of this [template](https://raw2.github.com/ansible/ansible/devel/examples/issues/ISSUE_TEMPLATE.md) and **paste it into the description** of your ticket. 

Our system will automatically close tickets that do not have an issue template filled out within 7 days. Please note that due to large interest in Ansible, humans may not comment on your ticket if you ask them questions. Don't worry, you're still in the queue and the robots are looking after you.

mpdehaan:
Hi thanks for this report, I'm going to link this ticket with https://github.com/ansible/ansible/pull/6238 and they can be considered together.

Thank you!

PR #11681, opened 2015-07-21
6238 rebase: add ssh_specific_args config option

amenonsen:
This is a trivial rebase of #6238 from @jeremyherbert with an added doc patch. Note that I don't particularly like the name `ssh_specific_args`, and I've not done anything about the linked-from-#8637 request to make sftp vars configurable. But I figured it would be easier to pick up discussion after nearly a year if there were a documented, rebased PR available.

(I tested that setting the variable does actually add the arguments to ssh.)

### DIFF
--- a/docsite/rst/intro_configuration.rst
+++ b/docsite/rst/intro_configuration.rst
@@ -703,6 +703,18 @@ If set, this will pass a specific set of options to Ansible rather than Ansible'
 In particular, users may wish to raise the ControlPersist time to encourage performance.  A value of 30 minutes may
 be appropriate. If `ssh_args` is set, the default ``control_path`` setting is not used.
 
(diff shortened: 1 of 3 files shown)


amenonsen:
I'll leave this open as requested, but I no longer have the time or inclination to maintain it. If anyone's interested in this functionality, it would be best to implement it in conjunction with/using the framework provided by #11908 (see discussion there).

amenonsen:
I'm closing this PR because of the conflict with #11908 and the fact that nobody has expressed any interest in either the functionality or the commit. Unfortunately the original #6238 has not yet been closed, so this may cause some confusion in future, but I can only hope that if anyone looks into that PR, they will also follow the links and understand the whole situation.

PR #11684, opened 2015-07-21
Fix verbose output with module 'script'

j-carl:
Fix verbose output with module 'script' (same as mentioned in cf3313b).


bcoca:
so v1 is dead code, this PR fixes code that will not be released or used.

PR #5335, opened 2013-12-18
Add installation notes about using a python virtual environment

sivel:
This documentation change adds some information about running ansible from within a python virtual environment, listing out some specific caveats about the setup and configuration.

PR #11689, opened 2015-07-22
Add a note about running in a virtualenv

amenonsen:
Based on a Dec 2013 PR by @sivel, but rephrased by me.

Closes #5335

### DIFF
--- a/docsite/rst/intro_installation.rst
+++ b/docsite/rst/intro_installation.rst
@@ -296,6 +296,35 @@ Packaging Ansible or wanting to build a local package yourself, but don't want t
 
 These releases are also tagged in the `git repository <https://github.com/ansible/ansible/releases>`_ with the release version.
 
+.. _virtualenv:
+
+Using a Python virtual environment
+++++++++++++++++++++++++++++++++++
+
+You can run Ansible in a Python virtual environment either from source,
+as described above, or when installed by pip or using packages.
+
+Many Ansible modules depend on Python modules, and if these are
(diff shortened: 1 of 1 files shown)


amenonsen:
Any thoughts on the wording here? I believe this accurately reflects the present situation, but if it's at all possible to reuse the interpreter that runs ansible for local connections, it would be greatly preferable, since it eliminates the problem altogether.

abadger:
@amenonsen @sivel, I wasn't able to get these instructions to work.  I think you have to activate the virtualenv in addition to using an alternate python binary.  Something like this stackoverflow answer: http://stackoverflow.com/questions/20575084/best-way-to-always-run-ansible-inside-a-virtualenv-on-remote-machines

amenonsen:
@abadger Sorry for the delayed response. I'd never used `virtualenv` before, so I took some time to familiarise myself with it. I believe you're right, of course, it'll need to be activated in `.bashrc` or similar before it's used on the far end of an ssh connection. But I'm not sure what level of detail is appropriate for the Ansible docs in this case. Anyway, given that #5335 languished for two years with neither input nor updates, I suspect the best thing to do might be to close both PRs and accept the status quo until someone more motivated provides a complete description.

amenonsen:
I'm closing this PR now. Unfortunately, neither has #5335 been closed, nor have the outstanding questions been answered; but I never had the necessary expertise to revise the PR, and no longer have the time to try.

PR #11709, opened 2015-07-23
cobbler.py dynamic inventory - support for _meta to improve performance, adjustable group criteria

dmccue:
These commits enable caching of cobbler meta data and will return all servers when run which improves performance.
In addition it can group by owner so that you can run ansible playbooks against servers owned by a specific user, adjustable groupby_key is stored in a variable however this could be moved out to the cobbler.ini file


bcoca:
cc @jimi-c

jimi-c:
Merged, thanks!

jimi-c:
Just a note, this was merged, it just wasn't auto-detected, so I closed it manually.

PR #11731, opened 2015-07-25
Delete unused import glob

danasmera:
glob modules does not seem to be reference any where in the setup.py file. Removed it.


jimi-c:
Merged, thanks!

PR #11733, opened 2015-07-25
v2 breakage: lookupfile should lookup the given file=xxx

amenonsen:
Without this patch, if you do:

```
lookup('csvfile', 'xxx file=foo.csv')
```

you get an error like:

```
"msg": "ERROR! csvfile: [Errno 2] No such file or directory: u'/home/ams/xxx file=foo.csv'"
```

This was because we used the pre-split `term` for the `self._loader.path_dwim_relative` call.


amenonsen:
I've also added a (trivial, attributed) rebase of #10953 here because Github says that one has conflicts. I've tested that the behaviour reported there is fixed after these two commits are applied.

PR #7485, opened 2014-05-21
Expand shell vars when looking up EC2_INI_PATH in ec2.py

ahamilton55:
Currently when looking up EC2_INI_PATH in plugins/inventory/ec2.py a full and direct path must be given. Environment variables will tend to use other variables and built-ins to allow for portability and this should be expected for this variable. So, expansion of "~" and environment variables has been added to the lookup of the EC2_INI_PATH variable.

PR #11738, opened 2015-07-26
Have ec2.py expand tilde and vars when looking up the EC2_INI_PATH env variable

amenonsen:
Rebase of #7485 to target `contrib/inventory` rather than `plugins/inventory`


jimi-c:
Merged, thanks!

PR #8712, opened 2014-08-21
Pull Request for Issue #8700 - Slight Refactoring of the EC2 Inventory Script

adamhadani:
See Discussion in Issue ticket - https://github.com/ansible/ansible/issues/8700.

To recap - This removes any implicit printing, uses the standard pythonic way of making a module runnable, and as such, allows easy re-use of the functionality here by importing this module in another and deriving from EC2Inventory class.

PR #11744, opened 2015-07-26
Made ec2.py dynamic inventory module more modular and importable

adamhadani:
Trying to build off the work in ec2.py, one is tempted to try and have a new script import the stock one and then derive from Ec2Inventory. 
I've actually been using this but had to make a couple slight modifications to the ec2.py script to support this.

This PR is a clone of the original one I created a while back ( #8712 ) that was created after the various ansible modules repositories were branched out.

### DIFF
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -158,6 +158,8 @@ class Ec2Inventory(object):
         elif not self.is_cache_valid():
             self.do_api_calls_update_cache()
 
+
+    def data_to_print(self):
         # Data to print
         if self.args.host:
             data_to_print = self.get_host_info()
@@ -169,7 +171,7 @@ class Ec2Inventory(object):
             else:
                 data_to_print = self.json_format_dict(self.inventory, True)
 
-        print(data_to_print)
+        return data_to_print
 
 
     def is_cache_valid(self):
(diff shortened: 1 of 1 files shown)


contrib/inventory/ec2.py
```diff
@@ -404,7 +406,7 @@ def get_instances_by_region(self, region):
             if e.error_code == 'AuthFailure':
                 error = self.get_auth_error_message()
             else:
-                backend = 'Eucalyptus' if self.eucalyptus else 'AWS'
```
adamhadani, line 407: trailing space

contrib/inventory/ec2.py
```diff
@@ -617,7 +619,7 @@ def add_instance(self, instance, region):
                     if self.nested_groups:
                         self.push_group(self.inventory, 'security_groups', key)
             except AttributeError:
-                self.fail_with_error('\n'.join(['Package boto seems a bit older.',
```
adamhadani, line 620: trailing space

contrib/inventory/ec2.py
```diff
@@ -714,7 +716,7 @@ def add_rds_instance(self, instance, region):
                         self.push_group(self.inventory, 'security_groups', key)
 
             except AttributeError:
-                self.fail_with_error('\n'.join(['Package boto seems a bit older.',
```
adamhadani, line 717: trailing space

amenonsen:
Apart from the extraneous whitespace changes, this looks good to me. It's possible that in future, we may want to make inventory scripts more extensible anyway. I have no idea what that might look like (more plugins?), but this seems as good a place to start as any.

ryansb:
An equivalent change to this has been merged, #21637

PR #11747, opened 2015-07-27
Fix a parsing bug that prevents IPv6 addresses from being used with add_host

amenonsen:
Rebased version of #8682

### DIFF
--- a/lib/ansible/plugins/action/add_host.py
+++ b/lib/ansible/plugins/action/add_host.py
@@ -81,5 +81,5 @@ def _parse_ip_host_and_port(hostname):
         else:
             return hostname, None
     elif ':' in hostname:
-        return hostname.split(':')
+        return hostname.rsplit(':', 1)
     return hostname, None
--- a/lib/ansible/plugins/action/add_host.py
+++ b/lib/ansible/plugins/action/add_host.py
@@ -20,6 +20,8 @@
 from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
+import re
+
 from ansible.plugins.action import ActionBase
 
 class ActionModule(ActionBase):
@@ -38,8 +40,8 @@ class ActionModule(ActionBase):
         new_name = self._task.args.get('name', self._task.args.get('hostname', None))
         #vv("creating host via 'add_host': hostname=%s" % new_name)
 
-        if ":" in new_name:
-            new_name, new_port = new_name.split(":")
+        new_name, new_port = _parse_ip_host_and_port(new_name)
+        if new_port:
             self._task.args['ansible_ssh_port'] = new_port
 
         groups = self._task.args.get('groupname', self._task.args.get('groups', self._task.args.get('group', ''))) 
@@ -58,4 +60,26 @@ class ActionModule(ActionBase):
 
         return dict(changed=True, add_host=dict(host_name=new_name, groups=new_groups, host_vars=host_vars))
 
+def _parse_ip_host_and_port(hostname):
+    """
+    Attempt to parse the hostname and port from a hostname, e.g.,
 
+    some-host-name
+    some-host-name:80
+    8.8.8.8
+    8.8.8.8:80
+    2001:db8:0:1
+    [2001:db8:0:1]:80
+    """
+    if hostname.count(':') > 1:
+        match = re.match(
+            '\[(?P<ip>[^\]]+)\](:(?P<port>[0-9]+))?',
+            hostname
+        )
+        if match:
+            return match.group('ip'), match.group('port')
+        else:
+            return hostname, None
+    elif ':' in hostname:
+        return hostname.split(':')
+    return hostname, None
--- a/test/units/plugins/action/test_add_host.py
+++ b/test/units/plugins/action/test_add_host.py
@@ -1,0 +1,47 @@
+import unittest
+
+from ansible.plugins.action import add_host
+
+
+class TestAddHost(unittest.TestCase):
+
+    def test_hostname(self):
+        host, port = add_host._parse_ip_host_and_port('some-remote-host')
+        assert host == 'some-remote-host'
+        assert port is None
+
+    def test_hostname_with_port(self):
+        host, port = add_host._parse_ip_host_and_port('some-remote-host:80')
+        assert host == 'some-remote-host'
+        assert port == '80'
+
+    def test_parse_ip_host_and_port_v4(self):
+        host, port = add_host._parse_ip_host_and_port('8.8.8.8')
+        assert host == '8.8.8.8'
+        assert port is None
+
+    def test_parse_ip_host_and_port_v4_and_port(self):
+        host, port = add_host._parse_ip_host_and_port('8.8.8.8:80')
+        assert host == '8.8.8.8'
+        assert port == '80'
+
+    def test_parse_ip_host_and_port_v6(self):
+        host, port = add_host._parse_ip_host_and_port(
+            'dead:beef:dead:beef:dead:beef:dead:beef'
+        )
+        assert host == 'dead:beef:dead:beef:dead:beef:dead:beef'
+        assert port is None
+
+    def test_parse_ip_host_and_port_v6_with_brackets(self):
+        host, port = add_host._parse_ip_host_and_port(
+            '[dead:beef:dead:beef:dead:beef:dead:beef]'
+        )
+        assert host == 'dead:beef:dead:beef:dead:beef:dead:beef'
+        assert port is None
+
+    def test_parse_ip_host_and_port_v6_with_brackets_and_port(self):
+        host, port = add_host._parse_ip_host_and_port(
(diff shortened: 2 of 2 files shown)


amenonsen:
I'll leave this PR open (it was originally a bugfix by @ryanpetrello, which I rebased). The bugfix is still relevant and desirable, but I no longer have the time to maintain it.

ryanpetrello:
So...what does it take to merge this?  I originally wrote this code exactly a year ago, and it's not particularly complicated or substantial.

amenonsen:
@ryanpetrello I don't know, but while I have your attention I would be grateful if you could close #8682 (which no longer applies, that's why I rebased it and opened this PR).

ryanpetrello:
Done!

lib/ansible/plugins/action/add_host.py
```diff
@@ -58,4 +60,26 @@ def run(self, tmp=None, task_vars=dict()):
...
+        match = re.match(
+            '\[(?P<ip>[^\]]+)\](:(?P<port>[0-9]+))?',
+            hostname
+        )
+        if match:
+            return match.group('ip'), match.group('port')
+        else:
+            return hostname, None
+    elif ':' in hostname:
+        return hostname.split(':')
```
abadger: I like to give split a second parameter to control the maximum number of times it will split.

So something like: `hostname.split(':', 1)`  it may also be appropriate to use rsplit() instead of split() here (although the ipv6 address might be the only case where rsplit() vs split() would be valid and you're already taking care of that above.)

lib/ansible/plugins/action/add_host.py
```diff
@@ -58,4 +60,26 @@ def run(self, tmp=None, task_vars=dict()):
...
+    """
+    if hostname.count(':') > 1:
+        match = re.match(
+            '\[(?P<ip>[^\]]+)\](:(?P<port>[0-9]+))?',
+            hostname
+        )
+        if match:
+            return match.group('ip'), match.group('port')
+        else:
+            return hostname, None
```
abadger, line 82: What is handled by this case?  Do we still want to parse a potential port specifier off the end?

ryanpetrello, line 82: Situations where the v6 string in question doesn't actually specify a port, e.g.,

2001:db8:0:1

See:

```
def test_parse_ip_host_and_port_v6_with_brackets
```

lib/ansible/plugins/action/add_host.py
ryanpetrello: Wouldn't this be handled by the branch above?

```
if hostname.count(':') > 1:
```

amenonsen:
@abadger I agree with @ryanpetrello, the two bits of code you commented on look fine to me.

lib/ansible/plugins/action/add_host.py
abadger: Yes but changing this makes the code both self-documenting and robust against changes to the surrounding conditional logic.  So it seems best to change this call for those reasons.

abadger:
Disclaimer:  The following is a cornercase, not a best practice.  I wouldn't encourage anyone to do this but the rest of the code seems to allow it so I suppose there's a case to be made for handling it here (for now... until/unless we make it an error elsewhere).

The "name" parameter to add_hosts does not appear to be a true DNS name.  Instead it is a name that ansible uses to reference a host.  If the "name" does not also contain an ansible_ssh_hostname parameter than the name would have to be valid as an ip address or a DNS hostname.  This is the common case.  However, a cornercase exists when an inventory does specify an ansible_ssh_hostname.  In that case, the hostname does not have to be valid in DNS because ansible_ssh_hostname is used instead.  That means that the user could specify something like this:

```
In inventory:
one:two ansible_ssh_host=localhost ansible_ssh_port=22

In a playbook:
- add_host:
    name: "one:two"
    ansible_ssh_host: "localhost"
    ansible_ssh_port: 22
```

Stupid thing to do?  Yes.  Does someone out there want to do it?  Probably so.  We get some really weird bug reports sometimes.  How far should we go to support this silly cornercase?  I'm not sure I'd go too far but I do think it would be good for the code to do a little bit more checking to prevent someone doing that from masquerading as valid ipv6 addresses.  So for instance, if someone passes in "[one:two]:90" we shouldn't parse that into "one:two", 90.  We should either return "[one:two]:90", None  or "[one:two]", 90 (I lean slightly towards the former to give the user flexibility).

So maybe enhance the regex to search for four sets of `[0-9a-fA-F]{1,4}` in order to decide that the value is an ipv6 address.

Does that make sense?

alfredodeza:
@abadger it is unreasonable to be this picky about a corner case that the code that is being replaced _was also doing_:

```
    if ":" in new_name:
        new_name, new_port = new_name.split(":")
```

If better support for corner cases when parsing names and ports is needed then that sounds like a
separate issue should be created to track that and not make this PR stale more (these changes have been proposed for over a year).

Merging this PR should not mean poor support for corner cases, it means "IPV6 is now supported in `add_host`

abadger:
@alfredodeza If the fix was hard i would go along with your reasoning, merge this, and then fix the cornercases later.  However the fixes here are easy and of value to the new code by making it robust in the face of the sometimes silly things that users want to do.  I've pointed out the sections of code to be changed and what they should be changed to above... they just need to be applied to the code.  If this PR seemed stalled I might fix it myself and merge.  However, @amenonsen and @ryanpetrello seem very responsive so I'm leaving them to make the changes since that seems to be the courteous thing to do with PRs on github.

ryanpetrello:
While I still think it's redundant, the max split is fine to me for the reasons you specified (plus, it's easy to change).

I totally disagree that the regular expression edge case is worth supporting, because the example you've provided (and for that matter, _any_ hostname composed of a bracket that is _not_ an IPv6 address) _is not_ a valid hostname or IPv4 address.

According to https://tools.ietf.org/html/rfc952:

```
1. A "name" (Net, Host, Gateway, or Domain name) is a text string up
   to 24 characters drawn from the alphabet (A-Z), digits (0-9), minus
   sign (-), and period (.).  Note that periods are only allowed when
   they serve to delimit components of "domain style names".
```

...and...

```
2. Internet Addresses are 32-bit addresses [See RFC-796].  In the
   host table described herein each address is represented by four
   decimal numbers separated by a period.  Each decimal number
   represents 1 octet.
```

Additionally from http://www.ietf.org/rfc/rfc3986.txt:

```
A host identified by an Internet Protocol literal address, version 6
   [RFC3513] or later, is distinguished by enclosing the IP literal
   within square brackets ("[" and "]").  This is the only place where
   square bracket characters are allowed in the URI syntax.
```

If ansible encounters a hostname in the format of `[ ...anything ... ]` it represents, _at best_, an IPv6 address.

abadger:
@ryanpetrello Right -- But in https://github.com/ansible/ansible/pull/11747#issuecomment-133298684 I mentioned that the parameter we're operating on is not a hostname.

ryanpetrello:
Also, I'll add that it's technically @amenonsen's fork; not sure if he's actually interested in doing the work.  If not, @amenonsen, let me know and I'll try to address the comments on this review.

amenonsen:
@abadger I spent some time thinking about this, and I think this patch is a reasonable and conservative approach, and should be applied as-is.

We're not very systematic about validating inventory host names right now, and there's more than one place that would be tripped up by an inventory hostname of `foo:bar` (e.g. it would be interpreted as two host patterns joined together). I think it's better to revisit this after v2, and define strict rules about what's acceptable, and enforce them consistently. (This is something I brought up just the other day on IRC.)

But in any case, `foo:bar` is never going to work as an inventory hostname, so I don't think we need to go too far towards handling that case. I think @ryanpetrello's patch is sensible enough to just merge for v2 as a strict improvement to the status quo.

amenonsen:
In particular, the `ini.py` rewrite follows the precedent of ssh (which _requires_ `[...]` around IPv6 addresses when you specify a port number, but accepts it for IPv4 addresses and hostnames as well; the latter may also be specified with a bare ":NN" suffix, because there's no ambiguity), and I think this is a sensible position to take (and it's what this patch does too).

abadger:
@amenonsen That does sound like a good future direction -- but if we do that, we still need to enhance the regex here to detect valid and invalid ipv6 addresses.  We'll just be changing the case where we detect that it's not an ipv6 address from returning the invalid ipv6 address as a valid ansible name into a case of raising an error.  So you could see this as future proofing....

amenonsen:
The second answer in http://stackoverflow.com/questions/53497/regular-expression-that-matches-valid-ipv6-addresses shows that validating IPv6 addresses is non-trivial. I agree we should do it, but it should be done in one place and used everywhere else (rather than the three separate approaches we now have). I volunteer to do this in a separate PR.

abadger:
Merged!  Thanks for humoring my idiosyncracies, guys!

# Merged

Hi!

issue is resolved.

Thank you!

PR #11749, opened 2015-07-27
Make «copy content='…'» add a newline if there isn't one

amenonsen:
Fun fact: according to POSIX, a file should end with a newline.
http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206

Rebase of #9468

### DIFF
--- a/lib/ansible/plugins/action/copy.py
+++ b/lib/ansible/plugins/action/copy.py
@@ -274,6 +274,8 @@ class ActionModule(ActionBase):
         content = to_bytes(content)
         try:
             f.write(content)
+            if not content.endswith('\n'):
+                f.write('\n')
         except Exception as err:
             os.remove(content_tempfile)
             raise Exception(err)
--- a/lib/ansible/plugins/action/copy.py
+++ b/lib/ansible/plugins/action/copy.py
@@ -273,9 +273,9 @@ class ActionModule(ActionBase):
         f = os.fdopen(fd, 'wb')
         content = to_bytes(content)
         try:
-            f.write(content)
             if not content.endswith('\n'):
-                f.write('\n')
+                content = content + '\n'
+            f.write(content)
         except Exception as err:
             os.remove(content_tempfile)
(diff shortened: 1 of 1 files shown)


lib/ansible/plugins/action/copy.py
```diff
@@ -274,6 +274,8 @@ def _create_content_tempfile(self, content):
         content = to_bytes(content)
         try:
             f.write(content)
+            if not content.endswith('\n'):
+                f.write('\n')
```
willthames: I'd avoid the second write with:

```
if not content.endswith('\n'):
    content = content + '\n'
f.write(content)
```

amenonsen:
@willthames thanks, I've pushed another commit to the PR to make the change you suggested.

amenonsen:
I'll close this PR now, because I no longer have the time to maintain it. In any case, this commit represents a (IMHO desirable, but) backwards-incompatible change. People have undoubtedly adapted to the present behaviour, and changing this would add extra newlines. For the record, a better way to do this would be to add another module parameter with a different name and newline-adding semantics.

kostyrev:
So no one cares for POSIX?

ahammond:
Is there a best-practices work around for this? I ended up putting the newline in manually. https://stackoverflow.com/questions/24291402/how-to-store-or-read-a-literal-carriage-return-and-newline-from-yaml-in-python

```yaml
- name: No passwords for sudo
  copy:
    dest: /etc/sudoers.d/sudo
    content: "%sudo\tALL=(ALL:ALL)\tNOPASSWD: ALL\r\n"
```

That said, I think saltstack has the right idea for the semantics of this. They have a `contents_newline` option which adds a newline to the contents, if it doesn't already end with a newline. See https://docs.saltstack.com/en/latest/ref/states/all/salt.states.file.html#salt.states.file.managed Ansible could default to `False` for this initially to preserve current behavior.

PR #11754, opened 2015-07-27
vault-password: ensure other users can't read vault password

billwanjohi:


bcoca:
so the v1 tree is dead code:

we don't take PRs against it as it won't be used or released in the future.

billwanjohi:
Ha, yeah, I should have figured that when I was rebasing. I've updated the pull request to target the 2.0 library, and tested locally with a properly and improperly permissioned vault password file.

PR #11756, opened 2015-07-27
vault-password: ensure other users can't read vault password

billwanjohi:
##### SUMMARY
Take 2 on #11754, this time on live code, and after testing locally.

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
lib/ansible/cli/__init__.py

##### ANSIBLE VERSION
2.3

### DIFF
--- a/lib/ansible/cli/__init__.py
+++ b/lib/ansible/cli/__init__.py
@@ -476,6 +476,8 @@ class CLI(object):
         if not os.path.exists(this_path):
             raise AnsibleError("The vault password file %s was not found" % this_path)
 
+        CLI.check_if_file_private(this_path)
+
         if is_executable(this_path):
             try:
                 # STDERR not captured to make it easier for users to prompt for input in their scripts
@@ -494,3 +496,13 @@ class CLI(object):
 
         return vault_pass
 
+    @staticmethod
+    def check_if_file_private(path):
+        '''raises an error if the mode of the vault_pass file
+        allows access by anyone other than the owner'''
+        mode = oct(os.stat(path).st_mode)
+        if mode[-2:] != '00':
+            raise AnsibleError(
+                "Group and others should not have access to your vault password.\n" +
+                "Please run `chmod 600 {0}`".format(path)
+            )


bcoca:
I'm not sure we want this, many times a group shares a vault file, permissions should be up to the user as the context can vary.

billwanjohi:
I took inspiration for this from other common utilities like ssh-agent and postgresql that will fail when given overly permissive files. Example from [postgresql documentation](http://www.postgresql.org/docs/current/static/libpq-pgpass.html):

> On Unix systems, the permissions on .pgpass must disallow any access to world or group; achieve this by the command chmod 0600 ~/.pgpass. If the permissions are less strict than this, the file will be ignored.

I do follow you that this would be a nuisance to some subset of users (though I doubt many), but this is such an important security consideration that I'd argue for the interests of the majority of users, who will be better served by the more restrictive default.

(Ideally I'd even like to confirm that the file isn't tracked by git or other revision control systems, but that's for another pull request.)

bcoca:
I don't think it is a minority as vault is frequently used within a team setting, I would make it into a config enforce_vault_permissions=0440, for example.

amenonsen:
That makes sense to me: if `enforce_vault_permissions` is set to something and the permissions don't match, complain. If it's not set, don't do anything. (But that means this PR also needs a documentation update.)

jstoja:
I don't think this relates very well to ssh-agent or pgsql `.pgpass` file that are clear text. Ansible vault like is encrypted, which means that even if you have access to it, you cannot read its content. Limiting the permission too is nice, but if it's not done, I don't think it deserves any warning. cc @bcoca

kenyon:
> I don't think this relates very well to ssh-agent or pgsql `.pgpass` file that are clear text. Ansible vault like is encrypted, which means that even if you have access to it, you cannot read its content. Limiting the permission too is nice, but if it's not done, I don't think it deserves any warning. cc @bcoca

This is about the vault password file, not the vault data files themselves. That is, the `--vault-password-file` referred to in [ansible-vault](https://docs.ansible.com/ansible/latest/cli/ansible-vault.html).

jstoja:
Right. Sorry for the confusion.

Therefore, it probably make sense to have a warning for the others, but like suggested by @bcoca, the group is probably useful for teams.

kenyon:
IMO if a group is sharing a vault, each user should have their own password file (if they want). That way restrictive permissions could be enforced as suggested.

bcoca:
closing due to inactivity, if the proposed modifications are addressed, feel free to resubmit as a new PR.

PR #9795, opened 2014-12-12
Add LVM facts to setup module

hughsaunders:
##### Summary: Gather LVM facts in setup module

This commit adds LinuxHardware.get_device_facts() and calls that from
.populate().

LVM facts are only gathered if the setup module is running as root and
the lvm utilities are available (tested by searching for 'vgs').

If the conditions are met, facts are set for each volume group and
logical volume.

Example:

Test LVM Data:

```
$ sudo vgs
  VG   #PV #LV #SN Attr   VSize VFree
  test   1   2   0 wz--n- 5.00g 2.00g
$ sudo lvs
  LV      VG   Attr      LSize Pool Origin Data%  Move Log Copy%  Convert
  testlv  test -wi-a---- 1.00g
  testlv2 test -wi-a---- 2.00g
```

Facts Returned:

```
$ ansible localhost -i /tmp/inv -m setup -a 'filter=ansible_lvm'
localhost | success >> {
    "ansible_facts": {
        "ansible_lvm": {
            "lvs": {
                "testlv": {
                    "size_g": "1.00",
                    "vg": "test"
                },
                "testlv2": {
                    "size_g": "2.00",
                    "vg": "test"
                }
            },
            "vgs": {
                "test": {
                    "free_g": "2.00",
                    "num_lvs": "2",
                    "num_pvs": "1",
                    "size_g": "5.00"
                }
            }
        }
    },
    "changed": false
}
```

Test as non-root:

```
$ ansible localhost -i /tmp/inv-user -m setup -a 'filter=ansible_lvm'
localhost | success >> {
    "ansible_facts": {},
    "changed": false
}
```

Test without lvm utilities available

```
$ sudo mv /sbin/vgs{,.bk}
$ ansible localhost -i /tmp/inv -m setup -a 'filter=ansible_lvm'
localhost | success >> {
    "ansible_facts": {},
    "changed": false
}
```

PR #11762, opened 2015-07-28
Add LVM facts to setup module

amenonsen:
This commit adds LinuxHardware.get_device_facts() and calls that from
.populate().

LVM facts are only gathered if the setup module is running as root and
the lvm utilities are available (tested by searching for 'vgs').

If the conditions are met, facts are set for each volume group and
logical volume.

Rebase of #9795.


alxgu:
works fine for me :+1:

PR #11764, opened 2015-07-28
fixes remote code execution for su/sudo with strict remote umasks

lpirl:
This is a pull-request for v2 in analogy to #9902.

### DIFF
--- a/lib/ansible/plugins/action/__init__.py
+++ b/lib/ansible/plugins/action/__init__.py
@@ -144,7 +144,7 @@ class ActionBase:
 
         tmp_mode = None
         if self._play_context.remote_user != 'root' or self._play_context.become and self._play_context.become_user != 'root':
-            tmp_mode = 'a+rx'
+            tmp_mode = 0755
 
         cmd = self._connection._shell.mkdtemp(basefile, use_system_tmp, tmp_mode)
         self._display.debug("executing _low_level_execute_command to create the tmp path")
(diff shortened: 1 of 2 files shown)


amenonsen:
Looks reasonable.

(I love the title of this PR, "fixes remote code execution" sounds like a security fix. :-)

bcoca:
@lpirl can you walk me through the use cases here? this does not look like an improvement over the chmod command.

lpirl:
@bcoca have you read #9902? I observed that Ansible's temporary directory and the py-files on the remote machine are created using the SSH user. Ansible seems to assume that all those directories and files created in `/tmp` are also listable/readable by the su/sudo user. This assumption fails if the remote system uses a strict umask (like 077).
I think fixing this using additional `chmod`'s is complex because we don't know which directories `mkdir -p` created and thus which one to `chmod`. This is why I change the umask before the execution of the `mkdir -p`. To ensure, the umask is set back to its original value, the command is also executed in a subshell (`(…)`).
For example `mkdir -p` creates the directories `ansible/root/ansible-tmp-…` in `/tmp` then `chmod` would only affect the very last directory in the path but `ansible/root` might still be unreadable to the su/sudo user.
Hope this made it clear.

bcoca:
I see an issue with this patch making the ansible dir permissive every time, the chmod was only used when the sudo/su user was non root in which case this could not be avoided.

lpirl:
Prefixing `umask` still does only happen when the call to `mkdtemp` in `sh.py` specifies a mode. This happens only if su/sudo'ing (`_make_tmp_path` in `action/__init__.py`).
This behavior has not changed as far as I can see.

bcoca:
ah, true, had not seen that.

jimi-c:
Merged, thanks!

PR #11765, opened 2015-07-28
Use PBKDF2HMAC() from cryptography for vault keys.

ldx:
When stretching the key for vault files, use PBKDF2HMAC() from the
cryptography package instead of pycrypto. This will speed up the opening
of vault files by ~10x.

The problem is here in lib/ansible/utils/vault.py:

```
hash_function = SHA256

# make two keys and one iv
pbkdf2_prf = lambda p, s: HMAC.new(p, s, hash_function).digest()

derivedkey = PBKDF2(password, salt, dkLen=(2 * keylength) + ivlength,
                    count=10000, prf=pbkdf2_prf)
```

`PBKDF2()` calls a Python callback function (`pbkdf2_pr()`) 10000 times.
If one has several vault files, this will cause excessive start times
with `ansible` or `ansible-playbook` (we experience ~15 second startup
times).

Testing the original implementation in 1.9.2 with a vault file:

In [2]: %timeit v.decrypt(encrypted_data)
1 loops, best of 3: 265 ms per loop

Having a recent OpenSSL version and using the vault.py changes in this commit:

In [2]: %timeit v.decrypt(encrypted_data)
10 loops, best of 3: 23.2 ms per loop

### DIFF
--- a/docsite/rst/playbooks_vault.rst
+++ b/docsite/rst/playbooks_vault.rst
@@ -108,5 +108,11 @@ This is something you may wish to do if using Ansible from a continuous integrat
 
 (The `--vault-password-file` option can also be used with the :ref:`ansible-pull` command if you wish, though this would require distributing the keys to your nodes, so understand the implications -- vault is more intended for push mode).
 
+.. _speeding_up_vault:
 
+Speeding Up Vault Operations
+````````````````````````````
 
+By default, Ansible uses PyCrypto to encrypt and decrypt vault files. If you have many encrypted files, decrypting them at startup may cause a perceptible delay. To speed this up, install the cryptography package::
+
+    pip install cryptography
--- a/lib/ansible/parsing/vault/__init__.py
+++ b/lib/ansible/parsing/vault/__init__.py
@@ -81,6 +81,18 @@ try:
 except ImportError:
     HAS_AES = False
 
+# OpenSSL pbkdf2_hmac
+HAS_PBKDF2HMAC = False
+try:
+    from cryptography.hazmat.primitives.hashes import SHA256 as c_SHA256
+    from cryptography.hazmat.primitives.kdf.pbkdf2 import PBKDF2HMAC
+    from cryptography.hazmat.backends import default_backend
+    HAS_PBKDF2HMAC = True
+except ImportError:
+    pass
+
+HAS_ANY_PBKDF2HMAC = HAS_PBKDF2 or HAS_PBKDF2HMAC
+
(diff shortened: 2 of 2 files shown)


amenonsen:
Heh, I was looking into a very similar change just this morning. Please note however that the code under `v1/`  is thoroughly dead, and your PR should be targeted to `lib/ansible/parsing/vault/__init__.py` now. Also, how about (in a separate commit) abstracting out the checks so that you don't have to change n different places?

ldx:
@amenonsen thanks, I pushed a commit that implements the same in `lib/ansible/parsing/vault/__init__.py`.

Would you like me to refactor the `HAS_*` checks in v1? If v1 is outdated now, we can leave it as is. In `lib/ansible/parsing/vault/__init__.py` the check is already in a helper function.

bcoca:
please drop the changes to v1/ as it is there just to be able to quickly test if a bug was introduced in devel or was preexisting.

ldx:
I reverted the commit changing v1, and squashed the three commits in the branch into one.

amenonsen:
@ldx looks good, but `HAS_ANY_PBKDF2HMAC` already includes `HAS_PBKDF2` so you don't need to test for the latter separately in `check_prereqs`.

ldx:
@amenonsen thanks, I fixed it and squashed again.

amenonsen:
Looks good. Do you think we should mention this (i.e. the possibility to speed things up by doing `pip install cryptography`) somewhere in the documentation? I'm not sure if, and I'm not sure where.

ldx:
Mentioning this in the documentation is definitely a good idea. I guess other vault users have also been frustrated with slow startup times.

I see vault is documented in `docsite/rst/playbooks_vault.rst`. Is it okay to add a note there about speeding up vault access via installing `cryptography`?

bcoca:
i would not say it's OK, but that it is needed

ldx:
I pushed a commit with a documentation update.

amenonsen:
Sorry to nitpick, but I'd prefer more imperative wording similar to the following:

By default, Ansible uses PyCrypto to encrypt and decrypt vault files. If you have many encrypted files, decrypting them at startup may cause a perceptible delay. To speed this up, install the cryptography package::

ldx:
Amended the last commit.

amenonsen:
Cool, thanks. This looks good.

amenonsen:
Note to maintainers: this is a significant performance improvement for vault operations and should be in v2.

brainsik:
> Note to maintainers: this is a significant performance improvement for vault operations and should be in v2.

On my setup I went from waiting a painful minute for all the vaults to open — and Ansible to start doing it's thing — to waiting just 2s. This is a game changer for usability when you have a lot of vault files.

PR #11776, opened 2015-07-28
Simple automated pep8 of setup.py

slick666:
- Did text alignment
- converted over some " to ' for consistent quoting

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,9 +8,9 @@ from ansible import __version__, __author__
 try:
     from setuptools import setup, find_packages
 except ImportError:
-    print("Ansible now needs setuptools in order to build. Install it using"
-            " your package manager (usually python-setuptools) or via pip (pip"
-            " install setuptools).")
+    print('Ansible now needs setuptools in order to build. Install it using '
+          'your package manager (usually python-setuptools) or via pip (pip '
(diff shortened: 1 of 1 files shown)


amenonsen:
Style-only changes are generally not accepted because they break attribution (i.e. `git blame`). I don't think these changes should be applied.

slick666:
Isn't that what the '-w' is for in git blame? If addressing style changes directly is not an option is there a better way that we could capture this going forward then? Perhaps fold this into Travis CI somehow?

amenonsen:
@slick666 that's probably a discussion for the `ansible-devel` list. I don't mean to be discouraging at all, but this PR is unlikely to be merged as-is, and it's also unlikely that a discussion about the larger style question would progress productively here. IMHO it's better to close this PR and discuss it on-list.

jimi-c:
Hi @slick666, as @amenonsen said we generally don't accept these kinds of PRs. One of the main reasons is that it can cause merge conflicts in other outstanding PRs depending on the degree of formatting changes, meaning other authors will have to rebase their work. While that may not be the case here, as this file is generally not touched very often, it is just our general policy not to accept these kind of whitespace fixes without other work also being done in the file.

Thanks!

PR #11778, opened 2015-07-29
add credstash lookup plugin

scottcunningham:
Credstash is a small utility for managing secrets using AWS's KMS and DynamoDB: https://github.com/LuminalOSS/credstash

Example usage:

```
- name: "Test credstash lookup plugin"
  debug: msg="Credstash lookup! {{ lookup('credstash', 'my-github-password') }}"
```

All credstash options (region, table, etc) are also configurable:

```
- name: "Test credstash lookup plugin"
  debug: msg="Credstash lookup! {{ lookup('credstash', 'my-other-password', region='us-west-1') }}"
```

I have some small questions about contributing. This lookup plugin has a dependency on the `credstash` module. Is there a way for me to add this as a dependency for the plugin? Is there also a good way for me to add documentation?

### DIFF
--- a/lib/ansible/plugins/lookup/credstash.py
+++ b/lib/ansible/plugins/lookup/credstash.py
@@ -1,0 +1,42 @@
+# (c) 2015, Ensighten <infra@ensighten.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+from ansible.errors import AnsibleError
+from ansible.plugins.lookup import LookupBase
+
+import credstash
+
+
+class LookupModule(LookupBase):
+    def run(self, terms, variables, **kwargs):
+
+        if isinstance(terms, basestring):
+            terms = [terms]
+
+        ret = []
+        for term in terms:
+            try:
+                val = credstash.getSecret(term, **kwargs)
+            except credstash.ItemNotFound:
+                raise AnsibleError('Key {} not found'.format(term))
+            except Exception as e:
+                raise AnsibleError('Encountered exception while fetching {}: {}'.format(term, e.message))
+            ret.append(val)
+
+        return ret
--- a/lib/ansible/plugins/lookup/credstash.py
+++ b/lib/ansible/plugins/lookup/credstash.py
@@ -29,12 +29,11 @@ except ImportError:
     CREDSTASH_INSTALLED = False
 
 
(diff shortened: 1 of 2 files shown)


lib/ansible/plugins/lookup/credstash.py
```diff
@@ -0,0 +1,42 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+from ansible.errors import AnsibleError
+from ansible.plugins.lookup import LookupBase
+
+import credstash
```
bcoca: we should capture possible import exception and explain that this lookup requires the credstash python library when it fails

bcoca:
for documentation add it and an example to docsites/rst//playbooks_lookups.rst

scottcunningham:
Great, thanks a lot for the feedback @bcoca. I've added both now. The build is failing but it looks unrelated to me:

```
Compiling lib/ansible/module_utils/vmware.py ...
  File "lib/ansible/module_utils/vmware.py", line 125
    except vim.fault.InvalidLogin as invalid_login:
```

scottcunningham:
Fixed exception handling syntax that would've caused an error in Python 2.4.

Could I get an update on this please? Is there anything else that I can do to help move this forward?

bcoca:
no, sorry, just for us to have enough time to review and merge

lib/ansible/plugins/lookup/credstash.py
```diff
@@ -0,0 +1,52 @@
...
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+from ansible.errors import AnsibleError
+from ansible.plugins.lookup import LookupBase
+
+CREDSTASH_INSTALLED = False
+
+try:
```
bcoca, line 25: this should be in the run method, otherwise ansible itself will fail to run

lib/ansible/plugins/lookup/credstash.py
```diff
@@ -0,0 +1,52 @@
...
+            terms = [terms]
+
+        ret = []
+        for term in terms:
+            try:
+                val = credstash.getSecret(term, **kwargs)
+            except credstash.ItemNotFound:
+                raise AnsibleError('Key {} not found'.format(term))
+            except Exception, e:
+                raise AnsibleError('Encountered exception while fetching {}: {}'.format(term, e.message))
```
abadger: Lookup plugins run on the controller.  The minimum python version for the controller is python 2.6.  So str.format() can't use "{}".  The easy workaround is to use "{0}" and "{1}" instead.

lib/ansible/plugins/lookup/credstash.py
abadger, line 25: By this we mean the

```
if not CREDSTASH_INSTALLED:
    raise AnsibleError([...])
```

If importring credstash takes a long time it might be good to move that to inside the lookup module as well but that's an unrelated issue.

scottcunningham, line 25: Very good point - thanks! I've moved this to inside of the run function.

I don't think that credstash takes a particularly long time to import, but if you feel it'd be best to keep it inside the module then I can move it there.

lib/ansible/plugins/lookup/credstash.py
scottcunningham: Good point - I've updated the PR to fix this.

scottcunningham:
Update - I've fixed the issues raised above. Thanks a lot for taking the time to check this out.

scottcunningham:
Thank you! :)

PR #11784, opened 2015-07-29
tests: check that handlers are able to use 'include'

pilou-:
Following https://github.com/ansible/ansible/pull/11520#issuecomment-124164941

### DIFF
--- a/test/integration/targets/handlers/test_handlers_include.yml
+++ b/test/integration/targets/handlers/test_handlers_include.yml
@@ -1,0 +1,14 @@
+- name: verify that play can include handler
+  hosts: testhost
+  tasks:
+    - debug: msg="main task"
+      changed_when: True
+      notify: test handler
+      tags: ['playbook_include_handlers']
+  handlers:
+    - include: handlers.yml
+
+- name: verify that role can include handler
+  hosts: testhost
+  roles:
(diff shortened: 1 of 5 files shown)


pilou-:
Rebased.

test/integration/targets/handlers/test_handlers_include.yml
```diff
@@ -0,0 +1,14 @@
+- name: verify playbook handlers can includes tasks
```
mscherer: I think you got the description mixed, could you fix it ?

test/integration/targets/handlers/test_handlers_include.yml
```diff
@@ -0,0 +1,14 @@
...
+  hosts: testhost
+  tasks:
+    - debug: msg="main task"
+      changed_when: True
+      notify: test handler
+      tags: ['scenario1']
```
mscherer: I would also prefer a more explicit name for the tag

test/integration/targets/handlers/test_handlers_include.yml
pilou-: Done

test/integration/targets/handlers/test_handlers_include.yml
pilou-: Done

PR #11785, opened 2015-07-29
10489 rebase: add dict support to set theory filters

amenonsen:
This is an updated series of commits with some documentation added. It supersedes PR #10489.

### DIFF
--- a/docsite/rst/playbooks_filters.rst
+++ b/docsite/rst/playbooks_filters.rst
@@ -160,6 +160,25 @@ To get the symmetric difference of 2 lists (items exclusive to each list)::
 
     {{ list1 | symmetric_difference(list2) }}
 
+.. versionadded:: 2.0
+
+The above filters also work with two dicts. The following returns a dict
+that contains all of the keys in `dict1` and `dict2`::
+
+    {{ dict1 | union(dict2) }}
+
+If `dict2` contains any keys that are also in `dict1`, the resulting
+dict will contain the value from `dict2` for those keys. If you don't
+want this behaviour, you can set `strict` to `True`::
+
+    {{ dict1 | union(dict2, strict=True) }}
+
+In this case, the values from `dict1` will be used in the resulting hash
+for any common keys.
+
+The `intersect`, `difference`, and `symmetric_difference` filters also
+take a `strict` argument.
+
 .. _version_comparison_filters:
 
 Version Comparison Filters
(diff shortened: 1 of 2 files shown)


amenonsen:
So I rebased the earlier PR because this seemed a more promising approach than various others, including #7872 (which I liked best until now, but which has accumulated merge conflicts and the author has not responded to requests to rebase).

But I don't really like the whole `strict` thing in this PR. I think the default should be to overwrite, or merge (or whatever you want to call it) keys such that {{a|union(b)}} returns a dict where common keys in b override those in a. If you want keys in a to win, well, use {{b|union(a)}}.

I could still live with it if overwriting were opt-in behaviour, and I could live with different naming, but I unequivocally loathe and detest the idea of making it an error to have common keys or treating it as a "caveat" that keys from b override those from a or "forcing" the user to seek enlightenment in the documentation if they want this behaviour. (This is all entirely besides the technical point about raising errors in filters, which I haven't looked into.)

Anyway, given the number of times this has come up and the number of different implementations, I think it's clear we really should do _something_ for v2. The implementations are all trivial. I volunteer to write documentation for any patch that core agrees to merge. I'll even volunteer to implement the change if core will provide some guidelines on what would be considered acceptable.

evanccnyc:
Any chance this will make it into v2?

amenonsen:
I'll leave this PR open as requested, but I no longer have the time to do any further work on it. I still think this is a  decision that should be made for v2.

bcoca:
test play:

```
- hosts: localhost
  gather_facts: False
  vars:
    dicta:
        a: 1
        b: 2
        c: 3
        d: 4
    dictb:
        c: 3
        d: 2
        e: 3
        f: 4
    union: "{{dicta|union(dictb)}}"
    inter: "{{dicta|intersect(dictb)}}"
    inter_strict: "{{dicta|intersect(dictb, strict=True)}}"
    diff: "{{dicta|difference(dictb)}}"
    diff_strict: "{{dicta|difference(dictb, strict=True)}}"
    symdiff: "{{dicta|symmetric_difference(dictb)}}"
    symdiff_strict: "{{dicta|symmetric_difference(dictb, strict=True)}}"
  tasks:
    - debug: var=union
    - debug: var=inter
    - debug: var=inter_strict
    - debug: var=diff
    - debug: var=diff_strict
    - debug: var=symdiff
    - debug: var=symdiff_strict
```

jimi-c:
@amenonsen see my comments on #10489, and we can revisit the rewrite of this on the devel mailing list or #ansible-devel IRC.

Thanks!

PR #11792, opened 2015-07-30
Update dependencies for pywinrm

ghost:
Adding the python-dev and libkrb5-dev libraries avoids the error message "error: command 'x86_64-linux-gnu-gcc' failed with exit status 1" when running "pip install kerberos"

### DIFF
--- a/docsite/rst/intro_windows.rst
+++ b/docsite/rst/intro_windows.rst
@@ -30,7 +30,12 @@ On a Linux control machine::
 
 If you wish to connect to domain accounts published through Active Directory (as opposed to local accounts created on the remote host)::
 
+   # Debian-based distributions (Debian, Ubuntu, …)
    sudo apt-get install python-dev libkrb5-dev
+   pip install kerberos
+   
+   # RPM-based distributions (Fedora, CentOS, RHEL, …)
+   yum -y install python-devel krb5-devel krb5-libs krb5-workstation
    pip install kerberos
 
 Kerberos is installed and configured by default on OS X and many Linux distributions. If your control machine has not already done this for you, you will need to.
--- a/docsite/rst/intro_windows.rst
+++ b/docsite/rst/intro_windows.rst
@@ -28,16 +28,46 @@ On a Linux control machine::
 
    pip install https://github.com/diyan/pywinrm/archive/master.zip#egg=pywinrm
 
-If you wish to connect to domain accounts published through Active Directory (as opposed to local accounts created on the remote host)::
+Active Directory Support
+++++++++++++++++++++++++
 
-   # Debian-based distributions (Debian, Ubuntu, …)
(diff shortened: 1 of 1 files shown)


bcoca:
cc @cchurch @jhawkesworth

jhawkesworth:
Great idea to improve this.

This is good for apt-get users, but it would be great to include the same for other package managers.

Looking back at my nodes on Centos 6.5 machines I installed 
yum -y install python-devel krb5-devel krb5-libs krb5-workstation
I think krb5-workstation may not be strictly necessary but if I recall adds 'kinit' and 'klist' command line tools which are helpful when getting kerberos set up.

Could this be added too?

Jon

ghost:
@jhawkesworth Yep, can do!

I pulled the apt-get commands straight from the pywinrm page, so I'm not entirely sure of what is required for other package managers. Will have a look and pass in another commit.

ghost:
Let me know if there are any other package managers that you feel are worth adding.

jhawkesworth:
I guess it should really be the same package managers as are listed for the main installation of ansible here:
http://docs.ansible.com/ansible/intro_installation.html

i.e. as well as yum and apt, portage (Gentoo), pkg (FreeBSD), Mac OSX and openCSW (Solaris).

I don't have access to any boxes running the above though - I guess asking for equivalents on ansible-project mailing list might be an idea, unless you happen to be able to find out.

ghost:
Have shopped around a bit and added in what I could find. Not sure about OS X though, as there seems to be surprisingly little information out there (seeing as the module is written by Apple). 

Will put something up on the mailing list.

jhawkesworth:
Great, thanks for tackling this.

Thinking about it, I think I saw a comment a while back that kerberos comes
pre-installed in mac OSX but I don't know if that included the tools
(kinit, klist).

Jon

jhawkesworth:
sorry for delay getting back to this.  
+1 please shipit @gregdek @bcoca 
I'd like to add some more information about setting up the ansible controller as a domain client - based on @trondhindenes guide here http://hindenes.com/trondsworking/2015/07/27/ansible-and-windows-configuring-kerberosactive-directory-authentication/ , but getting all the right libs installed is good progress.

ghost:
Sorry @jhawkesworth, I've been a bit slack on my end too. I could have sworn I posted in the forum but it looks like I didn't.

Thread is up now: https://groups.google.com/forum/#!topic/ansible-project/mJY0BWCRhlw

Can weave some more of that guide information into this PR if you'd like. Or would you prefer it to be submitted via a different one?

jhawkesworth:
No worries @deliciousfudge - I thought I saw something on the mailing list a while back.
I've already started working a separate PR but good to be able to base it on this now that its been merged (thanks @bcoca)

PR #11798, opened 2015-07-30
add documentation about any_errors_fatal

ghost:
There is a setting in code named "any_errors_fatal" (so I found it there), but it isn't described on docsite (only in a few topics in the internet).
Here is my proposal to add a few doclines to docsite.

### DIFF
--- a/docsite/rst/playbooks_delegation.rst
+++ b/docsite/rst/playbooks_delegation.rst
@@ -185,6 +185,50 @@ use the default remote connection type::
     - hosts: 127.0.0.1
       connection: local
 
+.. _interrupt_execution_on_any_error:
+
+Interrupt execution on any error
+````````````````````````````````
+
+With option ''any_errors_fatal'' any failure on any host in a multi-host play will be treated as fatal and Ansible will exit immediately without waiting for the other hosts.
+
+This option is used when ''serial'' keyword is used. But sometimes ''serial'' execution is unsuitable - number of hosts is unpredictable (because of dynamic inventory), speed is crucial (simultaneous execution is required). But all tasks must be 100% successful to continue playbook execution.
+
+For example there is a service located in many datacenters, there a some load balancers to pass traffic from users to service. There is a deploy playbook to upgrade service deb-packages. Playbook stages:
+
+- disable traffic on load balancers (must be turned off simultaneously)
+- gracefully stop service
+- upgrade software (this step includes tests and starting service)
+- enable traffic on load balancers (should be turned off simultaneously)
+
+Service can't be stopped with "alive" load balancers, they must be disabled, all of them. So second stage can't be played if any server failed on "stage 1".
+
+For datacenter "A" playbook can be written this way::
+
+    ---
+    - hosts: load_balancers_dc_a
+      any_errors_fatal: True
+      tasks:
+      - name: 'shutting down datacenter [ A ]'
+        command: /usr/bin/disable-dc
+    
+    - hosts: frontends_dc_a
+      serial: '100%'
+      max_fail_percentage: 19
+      tasks:
+      - name: 'stopping service'
+        command: /usr/bin/stop-software
+      - name: 'updating software'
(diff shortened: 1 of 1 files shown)


amenonsen:
There's some misunderstanding here.

First, the description is wrong (and the tasks in your example also don't make sense to me). In a playbook with multiple plays, Ansible will continue on to a second play only if the first play succeeds _anyway_, i.e. even if you don't set `any_errors_fatal`. What the setting does is that if a _single_ host in a multi-host play fails, it's treated as a fatal error and Ansible will exit immediately without waiting for the other hosts.

Second, `any_errors_fatal` doesn't actually do anything in current devel. The field can be defined, but it doesn't cause any behaviour changes. This may be an oversight, or maybe it's just not implemented yet.

This PR should be closed.

ghost:
@amenonsen thanks for correction! I should fix the description.
But You say in 2nd version of Ansible this option would be disabled? In 1.9 it works (and is very useful).

amenonsen:
@daburmistrov I don't know what the plans are for `any_errors_fatal` in v2. Since it's possible to set it, maybe there are already some plans to implement it fully.

bcoca:
it will be implemented in v2, its on the list. 

on this PR i would note that this option is the equivalent of  `max_fail_percentage: 0`

ghost:
ok! I have a few questions:
1. Should I continue to [re]write docs about this option? (I can try to fix this in a day or two)
2. As I know, max_fail_percentage is not "OK" for my case.
3. I can give more description about my usage.
4. May be the discussion should be migrated to another platform (google groups/issues/etc)?

bcoca:
1. yes, it would be nice to document
2. the implementation of `any_errors_fatal` is to specifically to set `max_fail_percentage: 0`
3. great
4. if you want, not sure if this needs broader discussion

ghost:
Updated docs, my case is described in example there (in common steps).

And after writing this changes I had a thought: may be I can use "serial: 100%" with "max_fail_percentage: 0"?
So I am a bit confused now. :)

Comment please. Maybe my description is kind of "invalid"...

bcoca:
this is confusing, and yes serial: 100% with "max_fail_percentage: 0 amounts to the same

ghost:
> this is confusing

what is "this"?

Current diff is not appropriate? I can't find a place in current docs to add short line about "bailing out from playbook". Maybe you have an advise?

bcoca:
this == any_errors_fatal

docsite/rst/playbooks_delegation.rst
```diff
@@ -185,6 +185,50 @@ use the default remote connection type::
...
       connection: local
 
+.. _interrupt_execution_on_any_error:
+
+Interrupt execution on any error
+````````````````````````````````
+
+With option ''any_errors_fatal'' any failure on any host in a multi-host play will be treated as fatal and Ansible will exit immediately without waiting for the other hosts.
+
+This option is used when ''serial'' keyword is used. But sometimes ''serial'' execution is unsuitable - number of hosts is unpredictable (because of dynamic inventory), speed is crucial (simultaneous execution is required). But all tasks must be 100% successful to continue playbook execution.
```
bcoca: not sure this true or relevant, it seems more related to max_fail_percentage

docsite/rst/playbooks_delegation.rst
```diff
@@ -185,6 +185,50 @@ use the default remote connection type::
...
+        command: /usr/bin/upgrade-software
+    
+    - hosts: load_balancers_dc_a
+      tasks:
+      - name: 'Starting datacenter [ A ]'
+        command: /usr/bin/enable-dc
+
+
+In this example Ansible will start software upgrade on frontends only if all load balancers are successfully disabled.
+
```
bcoca, line 229: so this is a bit confusing also, as you show both examples of any_errors_fatal and serial + max_fail but you do not mention max_fail_percentage at all in the text above or below.

ghost:
Removed confusing parts. Is this okay now?

PR #11803, opened 2015-07-30
Uncomment encrypt handling for vars_prompt

amenonsen:
This makes the following work again:

```
vars_prompt:
  - name: varname
    prompt: "password"
    encrypt: "sha512_crypt"
```

It said `FIXME`, but I didn't really have to do much to `FIXIT`.


amenonsen:
I found another problem: if you set `confirm: true` and then type two different responses, it would break because `display` isn't defined.

amenonsen:
Another easy FIXME in the same code: `to_unicode` just needed to be imported, I tested that `vars_prompt` now works with both ASCII and UTF-8 input. (Without this patch, ASCII works, but I get a traceback with UTF-8.)

PR #11810, opened 2015-07-30
Allow multiple interpreters to be specified

quinot:
It may be inconvenient to set a specific ansible_<lang>_interpreter
explicitly for each host in inventory. Instead allow the specification
of a comma-separated list of interpreters, which are tried in turn until
one is found, and cache the value for future use. Interpreters can
be specified either with a full path, or with just a base name,
in which case they are located in PATH.

### DIFF
--- a/docsite/rst/intro_inventory.rst
+++ b/docsite/rst/intro_inventory.rst
@@ -251,13 +251,9 @@ ansible_shell_type
     By default commands are formatted using ``sh``-style syntax.
     Setting this to ``csh`` or ``fish`` will cause commands executed on target systems to follow those shell's syntax instead.
 ansible_python_interpreter
-    The target host python path. This is useful for systems with more
-    than one Python or not located at :command:`/usr/bin/python` such as \*BSD, or where :command:`/usr/bin/python`
-    is not a 2.X series Python.  We do not use the :command:`/usr/bin/env` mechanism as that requires the remote user's
-    path to be set right and also assumes the :program:`python` executable is named python, where the executable might
-    be named something like :program:`python2.6`.
+    The target host Python path. A colon separated list of candidate paths can be supplied.
 ansible_*_interpreter
-    Works for anything such as ruby or perl and works just like ``ansible_python_interpreter``.
+    Similar variables are supported for other interpreted languages such as Perl or Ruby.
     This replaces shebang of modules which will run on that host.
 
 .. versionadded:: 2.1
@@ -275,6 +271,10 @@ Examples from a host file::
   aws_host          ansible_ssh_private_key_file=/home/example/.ssh/aws.pem
   freebsd_host      ansible_python_interpreter=/usr/local/bin/python
   ruby_module_host  ansible_ruby_interpreter=/usr/bin/ruby.1.9.3
+
+  [all:vars]
+  # by default try some usual locations, and fall back to searching PATH
+  ansible_python_interpreter=/usr/local/bin/python2.6:/usr/bin/python:python
 
 Non-SSH connection types
 ++++++++++++++++++++++++
--- a/docsite/rst/playbooks_interpreters.rst
+++ b/docsite/rst/playbooks_interpreters.rst
@@ -44,12 +44,14 @@ command will be searched in PATH.
 Specifying multiple interpreters
 --------------------------------
 
-In the context of managing farms of heterogeneous hosts, it may be impractical
-to maintain an up-to-date record of what exact interpreter path is to
-be used on a given machine. In such a context, Ansible can be configured
-to look for the interpreter from a set of possible candidates given as a
-colon-separated list.  Each item is either a full name, or just a base name
-that will be resolved from PATH:
+In general it is best to specify the exact location of the correct
+interpreter for each machine in the inventory. However, in the
+context of managing farms of heterogeneous hosts, it may be impractical
+to maintain an up-to-date record of what exact interpreter path is
+to be used on a given machine. In such a context, Ansible can be
+configured to look for the interpreter from a set of possible
+candidates given as a colon-separated list. Each item is either a
+full name, or just a base name that will be resolved from PATH:
 
 Example::
 
@@ -62,7 +64,9 @@ This indicates that Ansible should try the following Python interpreters in orde
 - '/usr/bin/python'
 - a 'python' binary from PATH
 
-Ansible stops the search as soon as a matching executable is found.
+Ansible stops the search as soon as a matching executable is found. Note that
+this is not supported for machines using a release of the fish shell older than
+2.2.0.
 
 Interpreter caching
 -------------------
--- a/docsite/rst/playbooks_interpreters.rst
+++ b/docsite/rst/playbooks_interpreters.rst
@@ -1,0 +1,83 @@
+Specifying Interpreters
++++++++++++++++++++++++
+
+.. contents:: Topics
+
+Ansible modules are written in Python, or other interpreted languages such as Perl or Ruby.
+Following UNIX convention, the implementation language of a given module is identified
+by its shebang line, such as '#! /usr/bin/python'. This indicates the default interpreter
+to be used when running the module.
+
+However, the actual location of the appropriate interpreter may vary
+from host to host. For instance, on \*BSD systems, it is usually
+'/usr/local/bin/python'; on some machines, the default Python binary
+may be a 3.X one, whereas Ansible modules require a Python 2.X
+interpreter, which could be installed as '/usr/bin/python2'.
+
+
+Inventory variables
+-------------------
+
+The interpreter for a given language can be set using inventory variables named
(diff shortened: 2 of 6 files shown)


amenonsen:
Looks good. One quick comment: in `_remote_check_interp`, the debug statement should also say which interpreter it's checking for.

bcoca:
so I don't think we need to add this code, which would make the execution extremely slow and error prone, specially when the following play can do the same much more efficiently:

```
- hosts: all
  gather_facts: false
  tasks:
    - raw: which {{item}}
      register: mypython
      ignore_errors: True
      with_items:
        - python
        - python2
        - python2.7
        - python27
        - python2.6
        - python26

    - set_fact: ansible_python_interpreter={{(mypython.results|rejectattr('rc')|list|first).stdout.strip()|default(omit)}}

    - debug: var=ansible_python_interpreter
```

amenonsen:
That is really neat. I think it should be included in the documentation somewhere.

jhawkesworth:
Perhaps we could have a 'how to' page - a sort of 'infrequently asked
questions'.

On Fri, Jul 31, 2015 at 2:47 AM, Abhijit Menon-Sen <notifications@github.com

quinot:
@bcoca Thanks for the insightful alternative, indeed I overlooked the possibility of doing things that way. It's not entirely clear to me why this would be faster or less error-prone though: the play still needs to execute a remote command to check the interpreter; as a matter of fact your solution will always test all listed interpreters, rather than stop at the first working one found. That said, one thing to note is that the proposed change is a no-op (no cost, no behaviour change) for existing setups, since the new code will only trigger if more than one interpreter is listed, or just one but without a full path, and neither is currently valid.
So, no regression in any case, and any additional cost applies only when the feature is used. And since the determined path is cached as a fact, the check is not repeated for each play.

Also, I see the following advantages to the integrated solution I proposed over a play-based one:
- also applies to the ad-hoc command line;
- allows the interpreter to be determined prior to the implicit fact gathering phase;
- playbooks do not need to be modified to take advantage of it.

so I still see value in the proposed change.

quinot:
@amenonsen good suggestion, this is now implemented.

bcoca:
Sorry, I did not see your response till now. 

My example above is meant to be a 1 off, instead of debug use lineinfile to update the local inventory to classify the hosts or set the vars, which will avoid running the multiple pythons at each play.

The other issue is that one of those interpreters might exist but won't be the 'correct' python, it could by python 3 or another python 2 that does not have the desired libraries installed.

Still I can see the value of this (with all caveats documented) but I'm still not sure it should be included, looping in the rest of core.

amenonsen:
I think this change has value and should be included, but I agree that it needs more documentation to explain how and when it can be used. I hadn't realised that the interpreter-search would be run again for each play. Why is that the case? Surely facts gathered in one play will persist for the remaining plays?

If you're setting the list to begin with, you can take care of the found-but-incorrect problem pretty easily. Or at least, I could do so in my situation.

Another thought: Maybe the interpreter names/paths should be `:`-separated instead of comma-separated (for consistency with e.g. PATH and various other similar settings within Ansible).

quinot:
@bcoca @amenonsen thanks for your feedback!

The interpreter search is not run for each play, but only once per host for each _playbook_. The path for the interpreter that was found to be working is then recorded as a host fact. If facts caching is in use, the value is even preserved for further playbook runs, and the search is not performed again.

The use of `,` rather than `:` was deliberate: unlike `PATH`, elements in the list of candidate interpreters are not directories to search for the executable, but either base names or full paths of executables. For example in the typical case where you want to handle a mix of Linux and FreeBSD machines, you might have:

```
ansible_python_interpreter=/usr/local/bin/python,/usr/bin/python
```

and not

```
ansible_python_interpreter=/usr/local/bin:/usr/bin
```

However if you think `:` is more appropriate anyway, that's perfectly fine by me :-)

I will also make sure to add more documentation to this PR.

amenonsen:
@quinot Interpreter discovery running only once per playbook is exactly how I had thought it works. Thanks for confirming that's the case. So I think @bcoca's performance concerns are not relevant.

I do understand that `ansible_python_interpreter` is not a list of directories, but for whatever it's worth, I still think `:` is more appropriate, because it strongly suggests the familiar "first found" semantics of `PATH`.

quinot:
@amenonsen @bcoca Additional change pushed on the branch, which:
- changes the separator to ':'
- expands the documentation of the feature.

quinot:
@amenonsen @bcoca PR updated against latest devel changes.

quinot:
@amenonsen @bcoca I have just updated the pull request again against the latest devel changes. Did you have an opportunity to review the updates to code and documentation following your feedback?

abadger:
We've talked about this at a few team meetings and we're not 100% on board but we are leaning towards accepting the concept here.  However, (1) we're not going to put it into 2.0 so 2.1 would be the earliest it could be merged. and (2) we think figuring out which interpreter is correct in modify_modules() is too late... some things like remote_checksum need it at an earlier time so there's some code changes that need doing as well.

abadger:
Assigning to next milestone so that its on our radar when we start developing 2.1.

quinot:
@abadger Thanks for your feedback.
I have updated the pull request against the latest devel branch.
I see the _remote_checksum issue, I will look into how this could be addressed; presumably we'd want to do the interpreter check immediately before collecting facts.

quinot:
@abadger I have now pushed an updated change that allows the interpreter check to be made earlier for _remote_checksum.

quinot:
@abadger @bcoca Hi, I've just updated the PR against the latest devel commit. Have you had an opportunity to review the new version of the code, which takes your comments into account regarding _remote_checksum?

quinot:
@abadger @bcoca Hi, could you please give me some visibility on the status of this pull request? I believe the current version addresses all of the comments that have been made so far. Thanks!

bcoca:
we are still pushing out 2.0, as mentioned above we scheduled this for 2.1

quinot:
Thanks for the update @bcoca

quinot:
Branch updated against latest & greatest devel head. Do let me know if there is any outstanding issue with integration of this PR.
Thanks!

bcoca:
IRC: #ansible and #ansible-devel on irc.freenode.net

quinot:
I have updated the branch against the latest devel head; I see that with the integration of PR #14490, _remote_checksum does not in any case depend on a special case anymore, which is a nice cleanup!

quinot:
As discussed on IRC:
- @amenonsen volunteered to review and commit this change -- thanks! I've just updated it against the latest devel head again, and also simplified it a bit.
- @bcoca notes that it would be desirable to sanitize the facts returned by the remote host in `_parse_returned_data`, removing `ansible_*_interpreter` and all `play_context` variables, to ensure that they don't impact connection settings. Since this is an orthogonal issue, I'll submit a separate PR for that.

quinot:
A proposed implementation for the sanitization suggested by @bcoca is in PR #15925.

quinot:
New commits pushed following Public Core Meeting discussion:
- replace hard-coded || with _SHELL_OR as hinted by @abadger 
- document that the preferred usage is still to set a single interpreter per machine in inventory, per @bcoca's request
- also note that fish < 2.2.0 won't support multiple shells as it did not have "command -v".

@abadger also remarked that it was unfortunate for module_common having to call back into action. However in the current design, module_common has to know what the exact remote interpreter is, and it can be determined only once we have a working connection to the remote host. An alternative design that avoids this callback might be possible, but at the expense of a more intrusive reorganization of existing code.

mattclay:
This branch needs to be rebased in order for tests to run.

abadger:
Some more details from quinot about what it would look like to not have to callback into the action code: https://gist.github.com/abadger/bbab020cf9d87652c8432a69088fca31

We talked about this again at the last meeting for pretty much the whole time: https://meetbot.fedoraproject.org/ansible-meeting/2016-06-14/core_public_meeting.2016-06-14-19.13.log.html

People were hesitant over the complexity of adding this to the code base and unsure of what needs it was addressing.  A few alternatives were talked about (writing an action plugin that could be combined with set_fact to set the interpreter; changes to initial fact gathering so that it could perform this function.)  Agreed that we'd like to talk some more about this when @quinot can be present so that we know what the use cases are and can come to a decision about whether this PR or one of the alternatives is the best way forward.

@quinot can you make either the Thursday meeting or the Tuesday meeting next week?  https://github.com/ansible/community/blob/master/MEETINGS.md

jbergstroem:
I'd like to see this being configurable from `ansible.cfg` as well!

quinot:
It seems that no matter what efforts I invest in addressing the various issues that have been raised, an objection will remain on principle. I have explained several times, here as well as on IRC, why I believe this would be a useful feature, but it seems that Ansible leadership is not convinced by that. At this stage I think everyone's time is being wasted in this discussion, so I'll mark this pull request as abandoned.

PR #11811, opened 2015-07-31
FIXME in host range parsing

amenonsen:
1. Make the `step` argument work for alphabetic sequences: `[a:e:2]` means a,c,e (this was the `FIXME`).
2. Improved error messages for host range parsing errors: we say "host range must …something…" now
3. Mention `x:y` ranges in `intro_patterns.rst`.


jody-frankowski:
Looks good to me. I've been bitten by the wrong `[-]` range documentation already.

PR #11826, opened 2015-08-01
Another small batch of trivial FIXME removals

amenonsen:
The first two just needed to be removed, as the required changes have been made already.

The third one involves a small change suggested by @bcoca to make the raw module set in_data because `ssh.py` can't access the `module_name` any more.


jimi-c:
Merged, thanks!

PR #11827, opened 2015-08-01
Remove TODO: the play already overrides options

amenonsen:


lib/ansible/playbook/play_context.py
```diff
@@ -199,8 +199,6 @@ def __init__(self, play=None, options=None, passwords=None):
         self.password    = passwords.get('conn_pass','')
         self.become_pass = passwords.get('become_pass','')
 
-        #TODO: just pull options setup to above?
-        # set options before play to allow play to override them
```
abadger: The second line of this comment looks correct and should stay right?  options comes first so that the play can overwrite anything that the options set?

abadger:
Merged.  Thanks!

PR #11828, opened 2015-08-01
Show a better message when ansible.cfg is not found

amenonsen:
Earlier we would say «Using  as config file» if we didn't find one.


abadger:
Thanks!

Issue #11830, opened 2015-08-02
DigitalOcean Dynamic Inventory: support private_ip 


marcosgm:
When using private_networking, the API reports "ip_address": "10.132.156.108", and the Ansible Inventory gets the private IP instead of the public one, which is useless. We actually need to look for the public IP inside the networks list
            "networks": {
               "v4": [
                  {
                    "gateway": "10.132.0.1",
                    "ip_address": "10.132.156.108",
                    "netmask": "255.255.0.0",
                    "type": "private"
                  },
                  {
                    "gateway": "45.55.64.1",
                    "ip_address": "45.55.85.125",
                    "netmask": "255.255.224.0",
                    "type": "public"
                  }
                ],
We detect the feature here:
                 "features": [
                   "private_networking",
                   "virtio"],

jimi-c:
The above commit was merged, so closing this.

Thanks!

PR #11831, opened 2015-08-02
Bug #11830, keep working even if we use private_networking

marcosgm:
Just a quick hack to allow the Dynamic Inventory work even if we create instances with Private Networking. Withouth this patch, the DI reports the private IP to Ansible and we cannot connect to that.


jimi-c:
Merged, thanks!

PR #11832, opened 2015-08-02
Disallow --forks 0

amenonsen:
Without at least one worker process, things break:

  File "/home/ams/extern/ansible/ansible/lib/ansible/executor/process/result.py", line 103, in run
    result = self._read_worker_result()
  File "/home/ams/extern/ansible/ansible/lib/ansible/executor/process/result.py", line 69, in _read_worker_result
    (worker_prc, main_q, rslt_q) = self._workers[self._cur_worker]
IndexError: list index out of range

### DIFF
--- a/lib/ansible/cli/__init__.py
+++ b/lib/ansible/cli/__init__.py
@@ -186,7 +186,7 @@ class CLI(object):
             self.options.become_method = 'su'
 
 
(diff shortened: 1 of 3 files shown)


jimi-c:
Merged, thanks!

PR #11834, opened 2015-08-02
Make 'pause: prompt=…' print the prompt

amenonsen:
In stable-1.9, the prompt string is passed to raw_input(), which prints
it without an extra \n. Here we're just print()ing it, so the \n would
be doubled.


jimi-c:
Merged, thanks!

PR #11838, opened 2015-08-02
Fix for PowerShell unquote method when passed None.

cchurch:
Fix for a bug that surfaces when a template is unchanged and `src=None` is passed to the `win_file` module.


jimi-c:
Merged, thanks!

PR #11840, opened 2015-08-03
Add get_available_hash_algorithms()

muffl0n:
As described in ansible/ansible-modules-core#141 it would be handy to have the get_available_hash_algorithms() that was developed there to be included in module_utils.

This PR adds the method get_available_hash_algorithms(). Additionally it does some refactoring to ease the use of digest_from_file:
- You just need to call digest_from_file() with the name of the algorithm instead of the function itself
- digest_from_file() also does a self.fail_json() when the requested algorithm is not available

I tested the changes as good as I could. But I'm sure that I missed some corner cases, so feel free to critize. :)

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -94,27 +94,6 @@ try:
 except ImportError:
     pass
 
-HAVE_HASHLIB=False
-try:
-    from hashlib import sha1 as _sha1
-    HAVE_HASHLIB=True
-except ImportError:
-    from sha import sha as _sha1
-
-try:
-    from hashlib import md5 as _md5
-except ImportError:
-    try:
-        from md5 import md5 as _md5
-    except ImportError:
-        # MD5 unavailable.  Possibly FIPS mode
-        _md5 = None
-
-try:
-    from hashlib import sha256 as _sha256
-except ImportError:
-    pass
-
 try:
     from systemd import journal
     has_journal = True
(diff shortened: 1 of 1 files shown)


abadger:
Unfortunately this changes the signature of digest_from_file().  Maybe we can make this a small wrapper around digest_from_file() (or vice versa if that works better).

muffl0n:
Yeah, I already thought that this could be a problem. Maybe we could just do some check with `hasattr(algorithm, 'digest')`?

abadger:
@muffl0n That would work for me... Implementation-wise, use the callable() function to perform that check instead of checking for `__call__` (I think somethings (from C?) don't implement **call**).

muffl0n:
I already edited the code in my comment from checking for `__call__` to `digest` because right now digest_from_file() get's called with an object that does not have `__call__` but `digest`.
E.g.:

```
     def sha1(self, filename):
         ''' Return SHA1 hex digest of local file using digest_from_file(). '''
        return self.digest_from_file(filename, _sha1())
```

```
>>> from sha import sha as _sha1
>>> type(_sha1())
<type '_hashlib.HASH'>
>>> dir(_sha1())
['__class__', '__delattr__', '__doc__', '__format__', '__getattribute__', '__hash__', '__init__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', 'block_size', 'copy', 'digest', 'digest_size', 'digestsize', 'hexdigest', 'name', 'update']
```

abadger:
Sounds good.  Maybe check for hexdigest() since that's the function we actually use.

muffl0n:
Good hint! Thank you! I changed the code accordingly.

abadger:
Looks good except that we lose the ValueError exception in the md5 method().  Some modules are depending on that behaviour.  Since I'm going to a conference and won't be available to review this again, I'll merge and then fix that.

abadger:
@muffl0n I ended up changing things a little more than I thought I would.  Instead of keeping it a method that has to do work everytime the method is called I changed it back to storing values in a dict at the toplevel: https://github.com/ansible/ansible/commit/00b1a2c8651b7a2772cfbaec465c8d03e857402c

That should make things slightly quicker since the dict will only be populated onload, not everytime the function is called.

I don't think that changes your get_url code at all so I'll look at merging that as well.

abadger:
# Merged

Hi!

issue is resolved.

Thank you!

muffl0n:
Thanks for the merge! 
The additional changes look reasonable. I was already thinking about the possibility to just determine the algorithms once. Your solution looks much better than the one I was thinking about. :)

PR #11848, opened 2015-08-04
add 'required' to vars_prompt

simobasso:


bcoca:
I like the idea, thinking that there might be a better name, also you have several spurious whitespace changes on unrelated lines.

simobasso:
Thx @bcoca!

we have two option:
- change name, stop.
- reverse the logic (ex. name: allow_empty, default: yes)

what do you prefer?

amenonsen:
"Avoid empty prompt" is a completely misleading description for this feature. What it does is control whether empty inputs to a prompt are allowed or not. `allow_empty` would be all right, but I would prefer `required` with a default of `False`.

simobasso:
done

bcoca:
an assert/fail task at the top of the play would serve same purpose.

dharmabumstead (changes requested):
Please change to 'You can also prevent an empty prompt by including the 'required' parameter:'

dharmabumstead:
Closing due to inactivity. Please feel free to reopen.

PR #11854, opened 2015-08-04
playbooks_blocks.rst: Add syntax highlighting

msabramo:
See https://github.com/msabramo/ansible/blob/patch-2/docsite/rst/playbooks_blocks.rst


bcoca:
it seems to break the html though.

bcoca:
fixed via 479c4a4be9cbbf2576cef42de2dab26e3d524609

PR #11867, opened 2015-08-06
Allow exceptions to pass through the program

willthames:
This is useful when using `ipython --pdb -- $(which ansible-playbook) ...` for debugging.

Also show traceback when `ANSIBLE_DEBUG` is on

### DIFF
--- a/bin/ansible
+++ b/bin/ansible
@@ -36,6 +36,7 @@ import os
 import sys
 import traceback
 
+import ansible.constants as C
 from ansible.errors import AnsibleError, AnsibleOptionsError, AnsibleParserError
 from ansible.utils.display import Display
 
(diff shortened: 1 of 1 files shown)


jimi-c:
@willthames why is the last `raise` required, if the traceback is being shown and we're exiting? Simply raising there will make the program continue, which I don't think is correct.

willthames:
@jimi-c but bin/ansible will terminate on the next line anyway. I could make it `raise` and then `sys.exit`. 

The point is that the `raise` will make `ipython --pdb` immediately drop into the debugger. Which is useful in `ANSIBLE_DEBUG` mode.

jimi-c:
Ahh, I was not aware of that, regarding the debugger.

jimi-c:
Merged, thanks!

willthames:
@jimi-c thank you!

bcoca:
reverted as this breaks the clean capture of errors in constants.py

willthames:
@bcoca, can you provide some kind of bug report or tests that fail? I honestly don't understand at all what you are saying. What capturing of errors in constants.py?

PR #11875, opened 2015-08-06
Add lookup plugin with_each_found

fuhbar:
##### SUMMARY
**Expect a list of file names and filter out existing ones.**

Derived from the lookup plugin 'with_first_found' this plugin implements the feature of iterating over a list of file names returning only actual resolvable files (`./vars`, `./templates`, `./files`) while silently omitting missing ones. This is especially useful when loading var files while maintaining an order of precedence. E.g. if you want to read vars in the way that most specific ones overriding broader scoped ones while not knowing which files may be requested you can utilize this lookup.

_This is currently just a draft for review (limited to ansible version 1) in order to check whether further effort makes sense in your eyes. Please comment! If you find it worthy to hack on I am going to write code for version 2 and may add further integration tests if required._

A common use case is when you want to dynamically load vars based on host distribution facts while providing common defaults via `ansible_family_os`  and preserving the possibility to override those common defaults by more specific var files. See example below: 

```
- name: Load OS based default variables
  include_vars: "{{ item }}"
  ignore_errors: yes
  with_each_found:
    - "default.yml"
    - "{{ ansible_os_family }}.yml"
    - "{{ ansible_distribution }}.yml"
    - "{{ ansible_distribution }}-{{ ansible_distribution_release }}.yml"
```

I find this a pretty common scenario where you prepare your role for future support of even more linux distributions.

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
v1/ansible/runner/lookup_plugins/each_found.py

##### ANSIBLE VERSION
2.3

### DIFF
--- a/v1/ansible/runner/lookup_plugins/each_found.py
+++ b/v1/ansible/runner/lookup_plugins/each_found.py
@@ -1,0 +1,186 @@
+# (c) 2013, seth vidal <skvidal@fedoraproject.org> red hat, inc
+# (c) 2015, Fabian Doerk <fabian.doerk@de.clara.net> Claranet GmbH
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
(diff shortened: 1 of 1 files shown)


fuhbar:
Hi @bcoca, 

is there any chance this lookup will make into 2.0? If so, please let me know in order to let me prepare the code for 2.0 and write tests for it (etc.).

Thanks in advance.

sivel:
@fuhbar,

Hi!

samdoran:
Thank you very much for your submission to Ansible.

Unfortunately, we're not sure if we want this feature in the program. Our reasons for this are:

  * This also may be accomplished using `with_fileglob` by passing in the list of files you want discovered in the desired order. Those that are are not found are skipped.

Thank you once again for this and your interest in Ansible.

Issue #11863, opened 2015-08-05
Role dependencies only run once, even with serial enabled


MichaelDrogalis:
##### Issue Type:

Bug report
##### Ansible Version:

`ansible 2.0.0 (devel 9aa4214417) last updated 2015/07/31 08:03:38 (GMT -700)`
##### Ansible Configuration:

``` text
[defaults]                                                                                                                                                  
filter_plugins = /usr/share/ansible_plugins/filter_plugins:./filter_plugins
roles_path = roles
host_key_checking = False
```
##### Environment:

OS X 10.9
##### Summary:

I have a role A with a dependency on role B. Role B is normally executed implicitly by role A's `meta/` folder specifying role B as a dependency. Role B gets executed _against each host_ before role A. Then, I set role A to be `serial` with a value of 1. I have three hosts in the inventory. Now when I run the playbook, role B only executes against the first host in the inventory, and is skipped on subsequent hosts.
##### Steps To Reproduce:
- Make a playbook with two roles
- Make role A depend on role B
- Set the inventory size to anything greater than 1
- Make role A serial with a value of 1
- Run the playbbook
##### Expected Results:

Role B executes against every host in the inventory when role A has `serial` set to 1.
##### Actual Results:

Role B only executes against the first host in the inventory.

MichaelDrogalis:
Pull request opened at #11878.

jimi-c:
# Closing This Ticket

Hi!

We believe the above commit should resolve this problem for you. This will also be included in the next major release.

If you continue seeing any problems related to this issue, or if you have any further questions, please let us know by stopping by one of the two mailing lists, as appropriate:
- https://groups.google.com/forum/#!forum/ansible-project - for user questions, tips, and tricks
- https://groups.google.com/forum/#!forum/ansible-devel - for strategy, future planning, and questions about writing code

Because this project is very active, we're unlikely to see comments made on closed tickets, but the mailing list is a great way to ask questions, or post if you don't think this particular issue is resolved.

Thank you!

PR #11878, opened 2015-08-06
Fix for issue #11863

MichaelDrogalis:
Hello!

I've created a pull request to potentially fix issue #11863. I don't expect this patch to be merged in, so much as help the developer working on this issue to see where the problem is. When using `serial` in a playbook, role dependency execution is skipped for hosts other than those in the first batch. It looks to me that this is because a restriction is placed on the inventory, causing the boolean expression in the batch to result in `True` too early. This value should only result in `True` if all the batches in the serial execution are finished. Hope this helps, and thanks for making Ansible!

### DIFF
--- a/lib/ansible/executor/play_iterator.py
+++ b/lib/ansible/executor/play_iterator.py
@@ -93,6 +93,7 @@ class PlayIterator:
 
     def __init__(self, inventory, play, play_context, all_vars):
         self._play = play
+        self._inventory = inventory
 
         self._blocks = []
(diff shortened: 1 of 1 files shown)


jimi-c:
Hi @MichaelDrogalis, I've fixed this at a slightly deeper level per the commit above.

Thanks for the PR though, it definitely help illustrate where the problem was occuring!

MichaelDrogalis:
Sure thing, thanks for getting the fix in @jimi-c!

PR #11896, opened 2015-08-07
Add new size computation on mountpoint (used space, block used ...)

Yannig:
This pull request replace this one: https://github.com/ansible/ansible-modules-extras/pull/563

During fact gathering, the module compute new values including:
- used space (+percent) ;
- block information ;
- inode information (+percent).


Yannig:
Now, I need a way to associate a path (something like /opt/my_path) with a mountpoint (which can be /opt/my_path, /opt or / regarding the server I'm launching my ansible playbook). Where and how we can do this?

Yannig:
This code include 2 new filter:
- unit: give bytes value of a given value (1|unit('K')=> 1024, 1|unit('M')=> 1048576 ...)
- getmountfrompath: retrieve the corresponding mountpoint of a given path.

Here's an example with the playbook available in **test/integration/test_space_available.yml**:

``` yaml
$ ansible-playbook -i samples/hosts test/integration/test_space_available.yml 

PLAY [Mount check] *************************************************************

ok: [localhost]

TASK [Get corresponding mountpoint from a given path] **************************
ok: [localhost]

TASK [debug var=mount1] ********************************************************
ok: [localhost] => {
    "changed": false, 
    "mount1": {
        "block_reserved": 12164467, 
        "block_used": 30830454, 
        "device": "/dev/sda1", 
        "fstype": "ext4", 
        "inode_free_pc": 98.71936352542352, 
        "inode_reserved": 0, 
        "inode_used": 778745, 
        "inode_used_pc": 1.2806364745764853, 
        "mount": "/", 
        "options": "rw,errors=remount-ro", 
        "size_available": 804303634432, 
        "size_available_pc": 82.03740810740767, 
        "size_total": 980410830848, 
        "size_used": 176107196416, 
        "size_used_pc": 17.962591892592336, 
        "uuid": ""
    }
}

TASK [debug var=mount2] ********************************************************
ok: [localhost] => {
    "changed": false, 
    "mount2": {
        "block_reserved": 12164467, 
        "block_used": 30830454, 
        "device": "/dev/sda1", 
        "fstype": "ext4", 
        "inode_free_pc": 98.71936352542352, 
        "inode_reserved": 0, 
        "inode_used": 778745, 
        "inode_used_pc": 1.2806364745764853, 
        "mount": "/", 
        "options": "rw,errors=remount-ro", 
        "size_available": 804303634432, 
        "size_available_pc": 82.03740810740767, 
        "size_total": 980410830848, 
        "size_used": 176107196416, 
        "size_used_pc": 17.962591892592336, 
        "uuid": ""
    }
}
yaml
TASK [check 1% of free space available for first file] *************************
ok: [localhost]

TASK [1M in bytes equals =>] ***************************************************
ok: [localhost] => {
    "changed": false, 
    "msg": "1048576"
}

TASK [check 1M available for second file] **************************************
ok: [localhost]

localhost                  : ok=7    changed=0    unreachable=0
```

Yannig:
A simple example showing a crash during assertion:

``` yaml
    - name: "check 10T available for second file"
      assert:
        that: mount2.size_available > 10|unit('T')
```

This check on a drive with less than 10 Terabits result in following crash:

``` yaml
TASK [check 10T available for second file] *************************************
fatal: [localhost]: FAILED! => {"assertion": "mount2.size_available > 10|unit('T')", "changed": false, "evaluated_to": false, "failed": true}
```

bcoca:
we already have the "human_readable" filter for byte/bit sizes, also instead of making the facts always calculate the percentage, i would prefer a filter or just do the computation in jinja2 and assign to a var.

Yannig:
@bcoca: Agreed for the percent, this can be computed later

But, I do not agreed for the unit filter (the name can be changed for something else): it have a different purpose vs human_readable:
- **unit** convert human readable values into bytes (4|unit('K') => 4096)
- **humean_readable** convert bytes into a human readable string (4096 => "4K")

Maybe I can move this filter into mathstuff.py do be more consistent. What do you think about?

bcoca:
might be easier to just add an option to the existing filter to make it machine readable, very similar code.

Yannig:
@bcoca: agreed. I'll take a look

Yannig:
@bcoca: I was looking around to find a good human to bytes and bytes to human converter in python and found this implementation: http://code.activestate.com/recipes/578019-bytes-to-human-human-to-bytes-converter/

Code is under MIT. What do you think about?

Yannig:
@bcoca: Re-work done

So basically, you can now check for 10T of free space using the following code:

``` yaml
    - name: "Get corresponding mountpoint from a given path"
      set_fact: mount1="{{file1|getmountfrompath(ansible_mounts)}}"
    - name: "check 10T available for second file"
      assert:
        that: mount1.size_available > "10 To"|human_bytes
```

When available space is bellow 10 To, ansible exit with the following message:

``` yaml
TASK [check 10T available for second file] *************************************
fatal: [localhost]: FAILED! => {"assertion": "mount1.size_available > \"10 T\"|human_bytes", "changed": false, "evaluated_to": false, "failed": true}
```

lib/ansible/modules/core
```diff
@@ -1 +1 @@
-Subproject commit 85ddb1b90232dbd68798e9b2d7dafa5689a1d30e
```
bcoca, line 1: you should not be updating submodule refs, remove them from this PR

lib/ansible/plugins/filter/mathstuff.py
```diff
@@ -100,24 +100,24 @@ def inversepower(x, base=2):
...
 
-    ranges = (
-            (1<<70L, 'Z'),
-            (1<<60L, 'E'),
-            (1<<50L, 'P'),
-            (1<<40L, 'T'),
-            (1<<30L, 'G'),
-            (1<<20L, 'M'),
-            (1<<10L, 'K'),
-            (1, base)
```
bcoca, line 117: now Bytes or bits does not get appended when there is no magnitude change

lib/ansible/modules/core
Yannig, line 1: @bcoca: Thanks for your patience but have no idea how to do that ...
Maybe I can refork my repositories and reapply my work against it.
In this case, I'm also wondering if I can split this PR into 2 differents one: 1 for the filter getmountfrompath (are you OK with this name?), and another for the size converter?

lib/ansible/plugins/filter/mathstuff.py
Yannig, line 117: I'll take a look. If this is the case, I'll add a new test in test/integration playbook.

lib/ansible/modules/core
bcoca, line 1: yes, do split, makes sense as they are 2 diff features

Yannig:
First PR getmoutfrompath: https://github.com/ansible/ansible/pull/12066

lib/ansible/modules/core
Yannig, line 1: @bcoca: First PR: https://github.com/ansible/ansible/pull/12066 (getmountfrompath)

Yannig, line 1: @bcoca: Second PR: https://github.com/ansible/ansible/pull/12073 (inodes and block informations)
I'll add a third one for the human_bytes filter.

Yannig, line 1: @bcoca: last PR: https://github.com/ansible/ansible/pull/12074 (new filter human_bytes)

Yannig:
Second PR: https://github.com/ansible/ansible/pull/12073

Yannig:
And last PR: https://github.com/ansible/ansible/pull/12074

Now, closing this PR.

PR #11901, opened 2015-08-07
Linting fixes for ec2 dynamic inventory script

jbarbuto:
This applies all the suggestions from [flake8](https://flake8.readthedocs.org/en/latest/) except the one for lines > 80 characters (E501). The codes are explained at https://flake8.readthedocs.org/en/latest/warnings.html.

I'm aware the contributing guide discourages "style-only" PRs, but a script like this is likely to be checked in as part of another code base, whose linting checks may be more strict.

### DIFF
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -343,7 +343,7 @@ class Ec2Inventory(object):
 
         # Do we need to exclude hosts that match a pattern?
         try:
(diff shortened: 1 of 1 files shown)


bcoca:
As a rule we don't take style fixes PRs, closing this.

vazhnov:
> As a rule we don't take style fixes PRs, closing this.
> This is sad.

PR #11690, opened 2015-07-22
Don't disable GSSAPIAuthentication or PubkeyAuthentication when using --ask-pass

amenonsen:
Closes #6846 (earlier PR)

PR #11908, opened 2015-08-10
Add ansible_ssh_extra_args inventory variable

amenonsen:
This PR does two things: (1) adds an `ansible_ssh_extra_args` inventory variable, and (2) makes it possible to override on the command-line. When it's set, the specified arguments are added to the ssh command for a host/group/everyone. This is applied in addition to any `ssh_args` configured in `ansible.cfg`, so there's no need to repeat the default `ControlPath` settings. (Note that `ssh_args` **cannot** be set as an inventory variable, so it's all hosts or nothing.)

The primary use of this feature is to specify a per-host or per-group `ProxyCommand` in the inventory to use a "jump host". Of course, it may have other uses as well, depending on the complexity of the ssh configuration.

This is an oft-requested feature, and it's the minimum common feature to both PRs #9122 (by @sivel) and #9477 (by @nitzmahone), which seek to make it easier to use by assembling the proxy command from simpler settings. Both of those PRs are currently outdated, not being worked on actively, and complex enough that they seem unlikely to make it to v2. In contrast, this PR is very low-impact, and I've tested that it's sufficient to set a jump host. It also will not conflict with the more user-friendly change attempted in the PRs referenced above, should they be resurrected in future.

I consulted @bcoca on the implementation via IRC, and he seemed to agree that this was the least intrusive way to implement the desired feature.

### DIFF
--- a/docs/man/man1/ansible-playbook.1.asciidoc.in
+++ b/docs/man/man1/ansible-playbook.1.asciidoc.in
@@ -151,6 +151,11 @@ run operations with su as this user (default=root)
 
 Run operations with sudo (nopasswd) (deprecated, use become)
 
+*--ssh-extra-args=*''-o ProxyCommand="ssh -W %h:%p ..." ...''::
+
+Add the specified arguments to any ssh command-line. Useful to set a
+ProxyCommand to use a jump host, but any arguments may be specified.
+
 *-U*, 'SUDO_USER', *--sudo-user=*'SUDO_USER'::
 
 Desired sudo user (default=root) (deprecated, use become).
--- a/docs/man/man1/ansible-pull.1.asciidoc.in
+++ b/docs/man/man1/ansible-pull.1.asciidoc.in
@@ -105,6 +105,11 @@ Purge the checkout after the playbook is run.
 
 Sleep for random interval (between 0 and SLEEP number of seconds) before starting.  This is a useful way ot disperse git requests.
 
+*--ssh-extra-args=*''-o ProxyCommand="ssh -W %h:%p ..." ...''::
+
+Add the specified arguments to any ssh command-line. Useful to set a
+ProxyCommand to use a jump host, but any arguments may be specified.
+
 *-t* 'TAGS', *--tags=*'TAGS'::
 
 Only run plays and tasks tagged with these values.
--- a/docs/man/man1/ansible.1.asciidoc.in
+++ b/docs/man/man1/ansible.1.asciidoc.in
@@ -143,6 +143,11 @@ Run operations with su as this user (default=root)
 
 Run the command as the user given by -u and sudo to root.
 
+*--ssh-extra-args=*''-o ProxyCommand="ssh -W %h:%p ..." ...''::
+
+Add the specified arguments to any ssh command-line. Useful to set a
+ProxyCommand to use a jump host, but any arguments may be specified.
+
 *-U* 'SUDO_USERNAME', *--sudo-user=*'SUDO_USERNAME'::
 
 Sudo to 'SUDO_USERNAME' instead of root.  Implies --sudo.
--- a/docsite/rst/faq.rst
+++ b/docsite/rst/faq.rst
@@ -55,6 +55,37 @@ consider managing from a Fedora or openSUSE client even though you are managing 
 We keep paramiko as the default as if you are first installing Ansible on an EL box, it offers a better experience
 for new users.
 
+.. _use_ssh_jump_hosts:
+
+How do I configure a jump host to access servers that I have no direct access to?
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+With Ansible version 2, it's possible to set `ansible_ssh_extra_args` as
+an inventory variable. Any arguments specified this way are added to the
+ssh command line when connecting to the relevant host(s), so it's a good
+way to set a `ProxyCommand`. Consider the following inventory group:
+
+    [gatewayed]
+    foo ansible_ssh_host=192.0.2.1
+    bar ansible_ssh_host=192.0.2.2
+
+You can create `group_vars/gatewayed.yml` with the following contents:
+
+    ansible_ssh_extra_args: '-o ProxyCommand="ssh -W %h:%p -q user@gateway.example.com"'
+
+Ansible will then add these arguments when trying to connect to any host
+in the group `gatewayed`. (These arguments are added to any `ssh_args`
+that may be configured, so it isn't necessary to repeat the default
+`ControlPath` settings in `ansible_ssh_extra_args`.)
+
+Note that `ssh -W` is available only with OpenSSH 5.4 or later. With
+older versions, it's necessary to execute `nc %h:%p` or some equivalent
+command on the bastion host.
+
+With earlier versions of Ansible, it was necessary to configure a
+suitable `ProxyCommand` for one or more hosts in `~/.ssh/config`,
+or globally by setting `ssh_args` in `ansible.cfg`.
+
 .. _ec2_cloud_performance:
 
 How do I speed up management inside EC2?
--- a/docsite/rst/intro_inventory.rst
+++ b/docsite/rst/intro_inventory.rst
@@ -211,9 +211,11 @@ SSH connection::
       The ssh password to use (this is insecure, we strongly recommend using --ask-pass or SSH keys)
     ansible_ssh_private_key_file
       Private key file used by ssh.  Useful if using multiple keys and you don't want to use SSH agent.
+    ansible_ssh_args
+      This setting overrides any ``ssh_args`` configured in ``ansible.cfg``.
     ansible_ssh_extra_args
       Additional arguments for ssh. Useful to configure a ``ProxyCommand`` for a certain host (or group).
-      This is used in addition to any ``ssh_args`` configured in ``ansible.cfg``.
+      This is used in addition to any ``ssh_args`` configured in ``ansible.cfg`` or the inventory.
 
 Privilege escalation (see :doc:`Ansible Privilege Escalation<become>` for further details)::
 
--- a/docsite/rst/intro_inventory.rst
+++ b/docsite/rst/intro_inventory.rst
@@ -211,6 +211,9 @@ SSH connection::
       The ssh password to use (this is insecure, we strongly recommend using --ask-pass or SSH keys)
     ansible_ssh_private_key_file
       Private key file used by ssh.  Useful if using multiple keys and you don't want to use SSH agent.
+    ansible_ssh_extra_args
+      Additional arguments for ssh. Useful to configure a ``ProxyCommand`` for a certain host (or group).
+      This is used in addition to any ``ssh_args`` configured in ``ansible.cfg``.
 
 Privilege escalation (see :doc:`Ansible Privilege Escalation<become>` for further details)::
 
--- a/lib/ansible/cli/__init__.py
+++ b/lib/ansible/cli/__init__.py
@@ -315,6 +315,8 @@ class CLI(object):
                 help="connection type to use (default=%s)" % C.DEFAULT_TRANSPORT)
             parser.add_option('-T', '--timeout', default=C.DEFAULT_TIMEOUT, type='int', dest='timeout',
                 help="override the connection timeout in seconds (default=%s)" % C.DEFAULT_TIMEOUT)
+            parser.add_option('--ssh-extra-args', default='', dest='ssh_extra_args',
+                help="specify extra arguments to pass to ssh (e.g. ProxyCommand)")
 
         if async_opts:
             parser.add_option('-P', '--poll', default=C.DEFAULT_POLL_INTERVAL, type='int', dest='poll_interval',
--- a/lib/ansible/playbook/play_context.py
+++ b/lib/ansible/playbook/play_context.py
@@ -163,6 +163,7 @@ class PlayContext(Base):
     _private_key_file = FieldAttribute(isa='string', default=C.DEFAULT_PRIVATE_KEY_FILE)
     _timeout          = FieldAttribute(isa='int', default=C.DEFAULT_TIMEOUT)
     _shell            = FieldAttribute(isa='string')
+    _ssh_extra_args   = FieldAttribute(isa='string')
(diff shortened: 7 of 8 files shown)


amenonsen:
Here's how I use it (this is a lightly-edited excerpt from the commit message, just as an example).

`inventory`:

```
[gatewayed]
foo ansible_ssh_host=192.0.2.1
bar ansible_ssh_host=192.0.2.2
```

`group_vars/gatewayed.yml`:

```
ansible_ssh_extra_args: '-o ProxyCommand="ssh -W %h:%p -q user@gateway.example.com"'
```

Now any access to hosts `foo` and `bar` would use the configured `ProxyCommand` to go through the bounce box.

bretmartin:
This would be very useful. I know one could construct `~/.ssh/config` to manage the SSH options there, but this would keep everything contained within an Ansible tree, making it easier to manage among a team.

amenonsen:
At @bcoca's request, I have added another commit to this PR that allows `ansible_ssh_args` to be set as an inventory variable, to override the global `ssh_args` setting in `ansible.cfg`. Note that even with this change applied, the `ansible_ssh_extra_args` setting is still useful, because setting `ssh_args` (in the config file or in the inventory) means that the default `ControlPath` settings will be overriden. But if that's what is in fact desired, this commit makes it possible.

atporter:
This would be great; much better than growing the out of tree per-user hacks.

nitzmahone:
@amenonsen- thanks for picking up the ball on this one. I've just been too busy with other stuff to do the v2 port of #9477. For this one, I think the "extra" part is a problem- merging SSH args robustly in the face of user-added args is very difficult, and will only become more so as we add first-class support for other SSH features (eg, agent forwarding). We already don't do a great job of it, and if someone's using something like -F, or eclectic combinations of -o args, the likelihood of us screwing it up is pretty high. Debugging SSH connectivity with ProxyCommand is already nightmarish, and I'm afraid that this just makes it even harder. I'm all for adding ansible_ssh_args as an inventory var, but IMHO, it should stay all-or-nothing. Anything else is just a recipe for trouble down the line...

amenonsen:
@nitzmahone Thanks for taking a look.

Could you please expand on your concerns about merging ssh args? I just went through every option documented in the `ssh_config(5)` manpage. The majority are yes/no; some take an integer; some take an alphanumeric word; some take comma-separated lists of alphanumeric words; some take paths, perhaps with patterns. None of those is a problem. The port forwarding and tunnelling options are only very slightly more complex. `ProxyCommand` and `LocalCommand` are the only two that seem at all esoteric in terms of the values they accept, and `ProxyCommand` is only passed to the local ssh process, so it's not exactly a quoting hazard either.

And on the other hand, even if we were to use only an `ansible_ssh_args` inventory variable, it wouldn't be all-or-nothing, because `ssh.py` is going to merge extra `-o Blah` arguments in **anyway**. So the only practical effect of insisting on doing it that way would be to force everyone to repeat `ControlPath` settings EVERY time they want to set a ProxyCommand (or live without `ControlPath`, I suppose). I cannot see how that is desirable.

Sure, someone might screw up the quoting while setting `ansible_ssh_extra_args` in an inventory file, but that applies equally to `ansible_ssh_args`, and they would end up with either a YAML parsing error (more likely) or a local ssh error and have to look at what's being executed with `-vvvv`. Maybe not fun, but it's certainly not been nightmarish for me (and I even have this one two-level jump host setup that I hate, but can't get rid of either).

Could you outline a specific situation in which you are concerned that adding user-specified extra `-o` args would cause more problems than adding the standard `-o` args to a user-specified `ansible_ssh_args`? I confess I don't follow what you mean with `-F`, since that just uses an alternate configuration file (and that works fine for me, with or without this patch). Is it the `shlex` stuff you're worried about? Because we'd have to keep doing that even if we only set `ansible_ssh_args`.

P.S. I'm curious, what do you have in mind in terms of first-class support for agent forwarding? I mean, it's just a yes/no setting. What would Ansible have to do with it?

jody-frankowski:
This would be useful to a lot of people with complex setups I'm sure. It also looks simpler and more flexible (e.g. allows multiple hops configurations) than #9477.
The only thing that bugs me personally, is the use of a second variable. If I only had a ssh_args variable that would add to itself, I think it would fine. You could add the `ControlPath` option in ansible.cfg for all hosts, and then set bastion/jumps in groups/hosts_vars.

nitzmahone:
@amenonsen My primary concern about merging args isn't the complexity of the args themselves, more that, as we add direct support (ie, an ansible_ssh_forward_agent inventory var vs "just hack it into ssh_args") for other ssh features, the likelihood of a conflict in the arg merging increases. By conflict, I mean where we double up on an arg that's already there, or where we set an option that's mutually exclusive. If we're just talking about -o options, it's not so bad, but some of the -o options also have a shortcut arg (eg, -o ForwardAgent=yes vs -A), and once you factor an external config file in with -F (where you're parsing/matching host specs), the likelihood of us correctly parsing real-world configs without screwing something up starts to get very high. The code we already have in there that tries to figure out if ControlPersist/ControlPath/ControlMaster are present is well-intentioned, but incomplete- now imagine adding that for a bunch of other args that we have first-class support for as well. I struggled with this in #9477, and I wasn't particularly happy with the results- it starts to get hairy pretty fast. 

When I say debugging this stuff is nighmarish, I refer to the difficulty that arises when some of the ssh args are managed and some aren't- try setting up a 2-or-3 level deep proxy chain, then getting enough detail from creatively placed "-v" args at various levels to debug an authentication issue, especially in a mixed key/password env (I've dealt with this exact scenario- not a walk in the park). Having to reconcile multiple vars that are contributing to the ssh args is kinda nasty without actually attaching a debugger, which I wouldn't expect an ordinary user to be doing. Some of this could probably be addressed with better verbose logging when things are contributing to the ssh args and/or adding even more first-class inventory support for various ssh_args. It's much easier to do that in v2 with the new connection setup step @jimi-c added (so connection plugins can grab whatever inventory vars they need at setup).

@jody-frankowski An "additive" ssh_args would work in some cases, but not if you actually needed to remove/override an arg that was set in the config; plus, Ansible doesn't really have a clean way to do that right now. You could probably hack something together with set_fact and a dictionary merge filter, but then ssh_args would need to support a dictionary of args (not a bad idea, but not there today). 

While having only a single ssh_args might be inconvenient in some cases, I think the end result would be more predictable and simpler than trying to merge from two args- if it's set, we punt on contributing args. There's probably a solution for the common ControlPersist/Master/Path issue- maybe we should focus on that.

amenonsen:
> By conflict, I mean where we double up on an arg that's already there, or where we set an option that's mutually exclusive

I don't think it's Ansible's job to worry about that. If I know enough to want to set specific ssh args myself, then I can figure out that I don't want to re-enable them or disable them via other ansible settings. We can't (or shouldn't) introduce new ssh settings with defaults that might conflict; this would be unsafe anyway because the conflicting setting may well be in `.ssh/config` rather than any flavour of `ssh_args`. With safe defaults, I would have to specifically enable something to get a conflict in the first place.

> once you factor an external config file in with -F (where you're parsing/matching host specs), the likelihood of us correctly parsing real-world configs without screwing something up starts to get very high

I don't understand why Ansible should ever try to parse ssh config files. It sounds like a terrible idea to me. (Yes, I know about the `ssh_config.py` inventory plugin, but that doesn't seem to be what you're referring to.) But if I understand you correctly, this isn't your central point, so I won't pursue it further.

> There's probably a solution for the common ControlPersist/Master/Path issue- maybe we should focus on that.

I'm open to suggestions, but I don't immediately see a way to do it without breaking backwards compatibility.

I think I do understand your primary concern. What I'm proposing means that the ssh command line could be built from (a) `ssh_args` set in `ansible.cfg`; (b) `ansible_ssh_args` set in the inventory; (c) `ansible_ssh_extra_args` set in the inventory; (d) from arguments always added by `ssh.py`; (e) from `~/.ssh/config` or some alternate specified with `-F` in `*_ssh_args`; (f) from `--ssh-extra-args` on the command-line. As a result, one might well wonder where some particular thing that gets added to the command line is coming from, and I can well imagine tearing my hair out trying to figure that out when running someone else's playbook. And if `ProxyCommand`s are involved, it just gets harder.

My approach is fully backwards-compatible, but you're right that it can lead to transparency problems. But note that my patch is not responsible for (a), (d), or (e), which are going to remain anyway. I could remove (f) from my patch, it's not especially important to me. I added (b) at @bcoca's request. I think adding more verbose logging to `ssh.py` will make it easier indeed.

If you can propose a solution that solves the `ControlPersist` problem in a way that (a) doesn't involve a lot of repetition, (b) significantly reduces the complexity above, and (c) is acceptable to the maintainers, I'm willing to revise the PR along the suggested lines. But failing that, I think this approach is the least-worst one and should be merged for v2.

amenonsen:
> I'm open to suggestions, but I don't immediately see a way to do it without breaking backwards compatibility.

Well, that's not entirely true. I could in theory make `ssh_args` set in `ansible.cfg` override the default persistence settings (i.e. retain the current behaviour) BUT make the value of `ansible_ssh_args` from the inventory (new) NOT override those defaults. So if you leave `ssh_args` unset in the configuration file and set it only in the inventory, then you get the default `ControlPersist` settings and add the `ansible_ssh_args` to them. But if you set `ssh_args` in the config file AND in the inventory, then the inventory version wins (and would have to duplicate the `ControlPersist` settings). I guess that boils down to adding an inventory variable with the almost the same semantics of `ssh_extra_args` but the same name as `ssh_args`.

I just re-read the above paragraph, and I'm ashamed of myself for writing it. I think `ssh_extra_args` is **a lot** easier to explain in the documentation than those contortions.

bcoca:
so I'm currently leaning to just add ansible_ssh_args as an inventory variable, we all seem to agree on that, and then examine the extra_args issue

amenonsen:
Of course I won't object to merging the `ansible_ssh_args` change, but unfortunately by itself that doesn't even solve my _own_ use case. Would you consider the entire series if I change `ssh.py` to make it report explicitly where it's getting each setting from in `-vvvv` mode?

amenonsen:
After some discussion with @bcoca on IRC, I've pushed another commit #c5de657 that makes `ssh.py` really verbose about why it's adding a particular argument to the commandline. Here's an example:

```
<aa.bb.cc.dd> ESTABLISH SSH CONNECTION FOR USER: xyz
<aa.bb.cc.dd> SSH: ansible.cfg set ssh_args: (-o)(ControlPersist=60s)
<aa.bb.cc.dd> SSH: found only ControlPersist; added ControlPath: (-o)(ControlPath="/tmp/ansible-ssh-%h-%p-%r")
<aa.bb.cc.dd> SSH: ANSIBLE_HOST_KEY_CHECKING/host_key_checking disabled: (-o)(StrictHostKeyChecking=no)
<aa.bb.cc.dd> SSH: ANSIBLE_REMOTE_PORT/remote_port/ansible_ssh_port set: (-o)(Port=44)
<aa.bb.cc.dd> SSH: ANSIBLE_PRIVATE_KEY_FILE/private_key_file/ansible_ssh_private_key_file set: (-o)(IdentityFile="foobar")
<aa.bb.cc.dd> SSH: ansible_password/ansible_ssh_pass not set: (-o)(KbdInteractiveAuthentication=no)(-o)(PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey)(-o)(PasswordAuthentication=no)
<aa.bb.cc.dd> SSH: ANSIBLE_REMOTE_USER/remote_user/ansible_ssh_user set: (-o)(User=xyz)
<aa.bb.cc.dd> SSH: ANSIBLE_TIMEOUT/timeout set: (-o)(ConnectTimeout=10)
<aa.bb.cc.dd> SSH: inventory added ansible_ssh_extra_args: (-o)(ProxyCommand=ssh -W %h:%p ams@bounce.toroid.org)
<aa.bb.cc.dd> EXEC ssh -C -tt -vvv -o ControlPersist=60s -o ControlPath="/tmp/ansible-ssh-%h-%p-%r" -o StrictHostKeyChecking=no -o Port=44 -o IdentityFile="foobar" -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o User=xyz -o ConnectTimeout=10 -o ProxyCommand=ssh -W %h:%p ams@bounce.toroid.org aa.bb.cc.dd (umask 22 && mkdir -p "$HOME/.ansible/tmp/ansible-tmp-1439451444.72-191022745003167" && echo "$HOME/.ansible/tmp/ansible-tmp-1439451444.72-191022745003167")
```

Look at the commit message for more details on the caveats that apply to the arguments added by `ssh.py` already: we can't differentiate between settings in the environment, in the config file, in the inventory, on the command line, and defaults. Still, if you're trying to debug stuff, this should be more helpful than the status quo, with or without my `ssh_*args` changes. (If the above seems too verbose, remember it's shown only with `-vvvvv`—that's _five_ v's.)

(I have also included a revised commit based on #11690, which would otherwise conflict with these changes. That's a rebased version of a bugfix PR #6846 from @kolbyjack. The commit message explains the problem. Sorry to put it here, but it's collateral damage from the scope of this PR being widened further.)

jody-frankowski:
This PR is related to #6238 and #11681.

amenonsen:
@jody-frankowski Yes, you're right. I've actually been wondering what, if anything, to do about that PR (originally from @jeremyherbert). It would be easy to incorporate into the framework established in this PR by adding, say, an `ssh_only_args` config (not inventory) setting with suitable description. I don't see a downside per se, but (a) it's a pretty esoteric use-case, and (b) I _really_ don't want to widen the scope of this PR any more than it's already been stretched.

But if someone wants to do that, this PR would make it simpler to implement and simpler to work with.

pwnall:
This would help me specify the expected host key, by specifying a per-host `known_hosts` file (see #12159 for details).

FWIW, I was immediately turned off by `ssh_args` when I read that it fully replaces the default arguments. I wished for an `ssh_extra_args` right away. It seems like a significantly better escape hatch for adding SSH arguments that I'm pretty sure Ansible wouldn't touch.

jimi-c:
Merged in via squashed commit 6650ba7654406bcb02096bdca8d8910526d01197. Thanks!

PR #11917, opened 2015-08-10
Add listall=True param to listPublicIpAddresses

DazWorrall:
`listall=True` is [required](https://cloudstack.apache.org/api/apidocs-4.5/domain_admin/listPublicIpAddresses.html) in order to see all the IP addresses you are authorised to access, not just the ones you own.

/cc @resmo (sorry, I've mentioned you a few times today!)


DazWorrall:
Not needed anymore

PR #11923, opened 2015-08-11
#9433 rebase: Cache the results for the RackSpace dynamic inventory

amenonsen:
This is a trivial rebase (handled automatically by git's 3-way merge) of PR #9433 by @intellitronica to target `contrib/inventory/rax.py`.


grahamrhay:
this behaviour is "surprising" when you've just added a new server, and it doesn't appear in the inventory. we had to add a step to bust the cache between playbooks:

```
./inventory/rax.py --list --refresh-cache > /dev/null
```

PR #11946, opened 2015-08-12
prevent local logging of module args under -vv when no_log specified

nitzmahone:
Module args were getting logged with -vv even when no_log was set; this caused sensitive data to be logged (eg, under Tower w/ Verbose setting).


nitzmahone:
@jimi-c - the aforementioned no_log fix PR...

PR #11951, opened 2015-08-13
Fix rm call in hacking/env-setup. Since it is sourced from a user shell,

jody-frankowski:
rm can have an alias to 'rm -i', which will make sourcing hang when '-q'
is enabled.

### DIFF
--- a/hacking/env-setup
+++ b/hacking/env-setup
@@ -43,7 +43,7 @@ expr "$MANPATH" : "${PREFIX_MANPATH}.*" > /dev/null || export MANPATH="$PREFIX_M
 gen_egg_info()
 {
     if [ -e "$PREFIX_PYTHONPATH/ansible.egg-info" ] ; then
-        rm -r "$PREFIX_PYTHONPATH/ansible.egg-info"
+        /usr/bin/env rm -r "$PREFIX_PYTHONPATH/ansible.egg-info"
     fi
     python setup.py egg_info
 }


abadger:
definitely a problem.  The patch introduces a dependence on env being installed and a dependence on it being installed at that particular path.  Perhaps a better way to fix this would be to use backslash rm like this:

```
 +        \rm -r "$PREFIX_PYTHONPATH/ansible.egg-info"
```

jody-frankowski:
I was wondering which one was the most portable and hadn't found any information on how portable '\' is. Maybe if someone can find if it's part of the posix standard we can use that instead.

abadger:
I'm not sure but I think it's just forcing the shell to think of it as a different string instead of the one that alias references.  So ""rm also works.  This could be implementation dependent (although it is so widely documented on the internet that I'm tempted to think that shell author's would consider it a bug).

we could also use `/bin/rm -r` that will be portable as long as rm lives at that filesystem location.

jody-frankowski:
I'm really not sure, but it seems to methat according to http://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html section 2.3.4 `\command` is part of the standard.
According to my understanding of section 2.3.4, `'command'`, `\command` and `"command"` should call `command` without the alias, and that is what happens in my shell.
So i'm fine with '\'. Do you prefer a new PR?

abadger:
You can just update this PR.  If you want to squash the commits, though, that would keep things cleaner.

PR #11953, opened 2015-08-13
Small inventory host-range parsing fixes

amenonsen:
This removes some dead code, revises a few comments, and fixes the problem of parsing "foo[1:3]:bar:baz" into a list.


jimi-c:
Merged, but closing manually due to some oddities in the way git did things. And by git, I mean me.

Thanks!

PR #11966, opened 2015-08-14
Update the irc channel address

CorySpitzer:
The irc channel address was out of date, pointing to a blank page. Also, make it a link.

(Not sure if this is the right place to make this change)


bcoca:
the info is correct for using an irc client, we can add it as an additional line item for those that want to use the web interface.

CorySpitzer:
I think that would make sense. Since I don't use an irc client I assumed they were web browser links

CorySpitzer:
Not sure if I'm doing this process right, but something like: https://github.com/CorySpitzer/ansible/pull/1

If that change is good, then I think there might be other places in the docs this could also be added.

bcoca:
Yeah, that looks good, update this PR with that.

jimi-c:
This has been updated in the devel branch, so closing now.

Thanks!

PR #11975, opened 2015-08-16
Add become support for OpenBSD doas

### LINKED ISSUE #11972: Compatibility with Openbsd "doas"
Hello, I am currently playing with ansible on OpenBSD 5.8 (to be released later this year). I am having issues with privileges escalation, as OpenBSD 5.8 has replaced "sudo" with "doas", cf: 

http://www.tedunangst.com/flak/post/doas
http://www.openbsd.org/58.html

I have tried setting this in my config file: 

```
sudo_exe=/usr/bin/doas
sudo_flags=
```

But I think ansible still pass more parameters to sudo than just "-u", and therefore I get the following error:

```
doas: unknown option -- k
usage: doas [-ns] [-C config] [-u user] command [args]
```

Am I missing something? If not, It would be very helpful to be able to use "doas" as a replacement for "sudo" in ansible!

Thanks!

mutemule:
This adds a new become_method `doas`, to support OpenBSD's `doas`.

There are a few things about this that I'm not a big fan of, but I couldn't see a cleaner way to do this:
1. `doas` treats its first argument as the command to execute, so we can't `doas "FOO=BAR python /path/to/python/script"`; this will fail, as there is no binary called `FOO=BAR python /path/to/python/script`.
2. Similarly, environment variables aren't commands. As far as I can tell, `ansible` always starts a command by setting environment variables, so this patch makes the assumption that this is always the case. This seemed easier and less fragile than trying to parse `cmd` to determine if environment variables were, in fact, being set. Perhaps this can be made safer with `.... env ANSIBLE=TRUE %s`?

Overall, it's pretty rudimentary (and still requires modifications to [ansible-modules-core/commands/command.py](https://github.com/ansible/ansible-modules-core/blob/devel/commands/command.py#L146)), but it's at least functional on my system.

fixes #11972

### DIFF
--- a/lib/ansible/playbook/play_context.py
+++ b/lib/ansible/playbook/play_context.py
@@ -387,6 +387,20 @@ class PlayContext(Base):
                 flags = self.become_flags or ''
                 becomecmd = '%s %s /user:%s "%s"' % (exe, flags, self.become_user, success_cmd)
 
+            elif self.become_method == 'doas':
+
+                prompt = 'Password:'
+                exe = self.become_exe or 'doas'
+                flags = self.become_flags or ''
+
+                if not self.become_pass:
+                    flags += ' -n '
+
+                if self.become_user:
+                    flags += ' -u %s ' % self.become_user
+
+                becomecmd = '%s %s echo %s && %s %s env ANSIBLE=true %s' % (exe, flags, success_key, exe, flags, cmd)
+
             else:
                 raise AnsibleError("Privilege escalation method not found: %s" % self.become_method)
 
--- a/docs/man/man1/ansible-playbook.1
+++ b/docs/man/man1/ansible-playbook.1
@@ -60,7 +60,7 @@ Run operations with become (nopasswd implied)
 .PP
 \fB\-\-become\-method=BECOME_METHOD\fR
 .RS 4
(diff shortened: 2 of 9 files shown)


lib/ansible/playbook/play_context.py
```diff
@@ -387,6 +387,20 @@ def detect_su_prompt(data):
                 flags = self.become_flags or ''
                 becomecmd = '%s %s /user:%s "%s"' % (exe, flags, self.become_user, success_cmd)
 
+            elif self.become_method == 'doas':
+
+                prompt = 'Password:'
```
bcoca, line 392: is this subject to i18n?

mutemule, line 392: As far as I can tell, no.

mutemule:
I've fixed up the tests. Amongst the other problems I had, the tests don't pass environment variables, so forcing in `env` would fail. So now use `env ANSIBLE=true` instead, which makes tests happy, but is quite ugly.

mutemule:
Given the `needs_revision` tag, is there anything else I should be doing here?

While I'm not a fan of the approach taken with `env`, this seems to be the simplest implementation of `doas` support. A more proper approach would be to pass desired environment variables distinct from the command to run, but this would be a much larger re-factor.

bcoca:
need to remove the v1 changes from the PR

mutemule:
@bcoca: Removed!

PR #11990, opened 2015-08-18
Make SSH binary path a configurable variable

sahand-mozaffari:
##### SUMMARY
Add SSH binary path to the list of configuration variables and make the existing code respect its value.

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
contrib/inventory/openvz.py
examples/ansible.cfg
lib/ansible/constants.py
lib/ansible/executor/task_executor.py
lib/ansible/plugins/connections/ssh.py

##### ANSIBLE VERSION
2.3

### DIFF
--- a/contrib/inventory/openvz.py
+++ b/contrib/inventory/openvz.py
@@ -26,6 +26,7 @@
 # Groups are determined by the description field of openvz guests
 # multiple groups can be seperated by commas: webserver,dbserver
 
+from ansible import constants as C
 from subprocess import Popen,PIPE
 import sys
 import json
@@ -42,7 +43,7 @@ def get_guests():
     #Loop through vzhosts
     for h in vzhosts:
         #SSH to vzhost and get the list of guests in json
-        pipe = Popen(['ssh', h,'vzlist','-j'], stdout=PIPE, universal_newlines=True)
+        pipe = Popen([C.ANSIBLE_SSH_BINARY, h,'vzlist','-j'], stdout=PIPE, universal_newlines=True)
 
         #Load Json info of guests
         json_data = json.loads(pipe.stdout.read())
--- a/examples/ansible.cfg
+++ b/examples/ansible.cfg
@@ -226,6 +226,10 @@ fact_caching = memory
 # only be disabled if your sftp version has problems with batch mode
 #sftp_batch_mode = False
 
+# can be used to set the path to SSH binary. Can be useful if SSH is not in
+# the system path, or if a customized version of SSH is meant to be used.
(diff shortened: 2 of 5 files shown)


semekh:
Is there anything preventing this PR from getting merged? It's a bit inconvenient to symlink and/or replace binaries while using ansible.

bcoca:
can you explain the use case? Ansible expects ssh to be in the local PATH on the master machine, which is the case in every OS I can think of.

semekh:
So, we've got this pseudo-ssh thing (we call it `docker-ssh`) which basically makes it possible to ssh into docker containers without running an ssh daemon. The good thing about this approach is that any ssh-compatible provisioning tool will easily be able to support provisioning docker containers. (e.g. in particular, we're aware of the ansible docker connection plugin, and it lacks support for some of ssh functionalities like agent forwarding)
So, if you want to provision a container, you'll just specify it as a regular host (e.g. `root@containerId` instead of `root@hostIp`) and change your ssh binary path to `docker-ssh`.

jody-frankowski:
@SeMeKh how is it superior to overriding with PATH? Which I think is the way most people will expect to override binary paths with. Is it to make the setting persistent?

tsreid:
I would also like to see this feature added. It would allow us to build tooling for test where we know the explicit version of SSH being used in the testing. It would also make it easier for us to deploy newer versions of SSH for Ansible to leverage but not affect everything else.

Merovius:
I would also like to see this happen. My usecase is hooking into the ssh execution to dynamically generate an ssh-config, known_hosts file and more fitting my dynamic inventory. Configuring the ssh binary command is superior, because I can use a local path to a script which then does all the setup, generates the necessary config and exec's ssh with the right arguments. This makes the instrumentation-repo self-contained and usable without extra setup.

bmalynovytch:
My two cents: I manage multiple platforms with Ansible for multiple customers for whom I have different ssh keys and configurations, managed by an ssh wrapper, called via an alias.
The alias, by the way, also wraps "caffeinate" (Mac OS X native command to prevent the computer from going to sleep while running long deployments).
There is a real need to be able to either way configure (through ansible.cfg / env / etc ...) the "ssh" command.

thaumos:
This has been implemented.

http://docs.ansible.com/ansible/intro_configuration.html#ssh-executable

abadger:
This was implemented in Ansible-2.2 as ANSIBLE_SSH_EXECUTABLE: http://docs.ansible.com/ansible/intro_configuration.html#ssh-executable

PR #11992, opened 2015-08-18
ansible manpage: document that the 'command' module is used by default

tpo:
The change documents that when '-m module_name' is not given, then the module 'command' is used by default.

Please pull


bcoca:
the default is also configurable in ansible.cfg module_name=

PR #12001, opened 2015-08-19
double the speed of loading a 10k host inventory file (#10034 rebase)

amenonsen:
before:  104411688 function calls (104398055 primitive calls) in 46.653 seconds
after:         4431688 function calls (4418055 primitive calls) in 21.034 seconds


srgvg:
Improves parsing my inventory with about 30-35%. Yay!

PR #12002, opened 2015-08-19
Make get_group run in O(n*log(n)) instead of O(n^2) by using a cache (#9437 rebase)

amenonsen:
For `ansible all -m ping` on an inventory with 15k groups:

before:         6615582 function calls (6579393 primitive calls) in 29.742 seconds
after:         6616072 function calls (6579883 primitive calls) in 7.540 seconds


bcoca:
groups cache should also be added to the refresh_inventory call

amenonsen:
Done.

Issue #12005, opened 2015-08-19
cannot retrieve facts from hostvars


yunano:
##### Issue Type:

Bug Report
##### Ansible Version:

```
$ ansible-playbook --version
ansible-playbook 2.0.0 (devel 1b810e3101) last updated 2015/08/19 21:59:38 (GMT +900)
  lib/ansible/modules/core: (detached HEAD dea5f5d15e) last updated 2015/08/19 21:59:49 (GMT +900)
  lib/ansible/modules/extras: (detached HEAD 210ee3febc) last updated 2015/08/19 22:00:00 (GMT +900)
  v1/ansible/modules/core: (detached HEAD f8d8af17cd) last updated 2015/08/19 22:00:12 (GMT +900)
  v1/ansible/modules/extras: (detached HEAD 495ad450e5) last updated 2015/08/19 22:00:23 (GMT +900)
  config file =
  configured module search path = None
```
##### Ansible Configuration:

no change
##### Environment:

CentOS 7.1
##### Summary:

If a playbook includes hostvars[hostname], the task fails with an exception like:

```
An exception occurred during task execution. The full traceback is:
Traceback (most recent call last):
  File "/home/user/ansible/lib/ansible/executor/process/worker.py", line 117, in run
    executor_result = TaskExecutor(host, task, job_vars, new_play_context, self._new_stdin, self._loader, shared_loader_obj).run()
  File "/home/user/ansible/lib/ansible/executor/task_executor.py", line 117, in run
    res = self._execute()
  File "/home/user/ansible/lib/ansible/executor/task_executor.py", line 317, in _execute
    result = self._handler.run(task_vars=variables)
  File "/home/user/ansible/lib/ansible/plugins/action/debug.py", line 37, in run
    results = self._templar.template(self._task.args['var'], convert_bare=True)
  File "/home/user/ansible/lib/ansible/template/__init__.py", line 191, in template
    result = self._do_template(variable, preserve_trailing_newlines=preserve_trailing_newlines, fail_on_undefined=fail_on_undefined, overrides=overrides)
  File "/home/user/ansible/lib/ansible/template/__init__.py", line 316, in _do_template
    res = j2_concat(rf)
  File "<template>", line 8, in root
  File "/usr/lib/python2.7/site-packages/jinja2/environment.py", line 389, in getitem
    return obj[argument]
  File "/home/user/ansible/lib/ansible/vars/hostvars.py", line 44, in __getitem__
    host = self._inventory.get_host(host_name)
[...]

PR #12017, opened 2015-08-20
Set serialization of hostvars to pickle all the object attributes

tiangolo:
Set serialization of hostvars to pickle all the object attributes to fix issue #12005, about the raising of the exception:

```
AttributeError: 'HostVars' object has no attribute '_inventory'
```

Here I copy my last comment in that issue describing the problem and the solutions.

In this Pull Request I'm implementing what I described as "**Option to fix it B**".

---

As @Pallokala said, the issue was introduced in commit e7b2308, specifically, in the file:

```
ansible/lib/ansible/vars/hostvars.py
```

In the last section of the class `HostVars`:

``` python
    def __getstate__(self):
        return self._lookup

    def __setstate__(self, data):
        self._lookup = data
```

starting from [this line](https://github.com/ansible/ansible/blob/e7b2308b66c7930df87d77232d5e48769ab03f58/lib/ansible/vars/hostvars.py#L64).

Those methods are [meant to be used by Pickle](https://docs.python.org/2/library/pickle.html#object.__getstate__) to generate the data to pickle and then unpickle it.

**Description of the error**:

I'm not an expert in Ansible internals, but it seems that that object is being created, then pickled and then unpickled, and after being unpickled, by the `__getstate__` and `__setstate__` methods, it only preserves its `_lookup` attribute and it doesn't have the other attributes that are created in the `__init__`.

Thereafter the object is being consulted for a host, calling the `__getitem__` method, that method checks if the host is in the `_lookup` dictionary, and as it doesn't find it, it looks for it in `host = self._inventory.get_host(host_name)`, and there is when the attribute error is raised.

**Option to fix it A**:

Remove the section for pickling and unpickling (`__getstate__` and `__setstate__`).

**Option to fix it B**:

Add all the other attributes to the return of the `__getstate__` and use them in the `__setstate__` to generate the new unpiclked object, as all the other objects are used in the method `__getitem__`.

The initial commit says that the intention was to "Speed up serialization of hostvars by simply using the internal" but I'm not sure how much this would help now, as the only thing that wouldn't be pickled would be the methods, I'm not sure that the speed up would be as much as originally intended since all the other attributes would also be pickled.

I plan on submitting a PR with the option B which is the more complex, in case the maintainers don't want it, the other option will be simpler to implement.

### DIFF
--- a/lib/ansible/vars/hostvars.py
+++ b/lib/ansible/vars/hostvars.py
@@ -17,6 +17,7 @@
 
 # Make coding more python3-ish
 from __future__ import (absolute_import, division, print_function)
+
 __metaclass__ = type
 
 import collections
@@ -27,16 +28,17 @@ from ansible.template import Templar
 
 __all__ = ['HostVars']
 
+
 # Note -- this is a Mapping, not a MutableMapping
 class HostVars(collections.Mapping):
     ''' A special view of vars_cache that adds values from the inventory when needed. '''
 
     def __init__(self, vars_manager, play, inventory, loader):
         self._vars_manager = vars_manager
-        self._play         = play
-        self._inventory    = inventory
-        self._loader       = loader
-        self._lookup       = {}
+        self._play = play
+        self._inventory = inventory
+        self._loader = loader
+        self._lookup = {}
 
     def __getitem__(self, host_name):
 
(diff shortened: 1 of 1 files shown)


bcoca:
he, I'm in the middle of testing 'option A' but I think this merits testing in case we can achieve a speedup and keep the correctness. cc @ansible/core

jimi-c:
Hi @tiangolo, I fixed this in a different way via https://github.com/ansible/ansible/commit/635fa0757b06b560a7812053f8de298fc5fc2c4c. Thanks for the contribution though!

Thank you!

PR #12023, opened 2015-08-20
add get_service_mrg_facts

le9i0nx:
It adds a variable that holds the name of the system initialization

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -134,6 +134,11 @@ class Facts(object):
                  { 'path' : '/usr/bin/pkg',         'name' : 'pkg' },
     ]
 
+    # A list of dicts. Init system
+    INIT_SYSTEM_MGRS = [ { 'path': '/proc/1/comm',  'name' : 'systemd' },
+                         { 'path': '/proc/1/comm',  'name' : 'init' },
+    ]
+
     def __init__(self, load_on_init=True):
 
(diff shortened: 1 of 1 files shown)


jimi-c:
Hi @le9i0nx, could you change the name to `service_manager`? I think that having the long form is better to make it obvious what it is. Thanks!

jimi-c:
Also, the `with open` syntax is not python 2.4 compliant (which facts code needs to be), so that will need to be fixed.

lib/ansible/module_utils/facts.py
```diff
@@ -515,6 +521,15 @@ def get_pkg_mgr_facts(self):
         if self.facts['system'] == 'OpenBSD':
                 self.facts['pkg_mgr'] = 'openbsd_pkg'
 
+    def get_service_mrg_facts(self):
+        self.facts['service_mgr'] = 'auto'
```
bcoca: 'auto' is not a valid value to return from this fact

lib/ansible/module_utils/facts.py
```diff
@@ -515,6 +521,15 @@ def get_pkg_mgr_facts(self):
...
+        for service in Facts.INIT_SYSTEM_MGRS:
```
bcoca: this is not a good way to detect initd systems, upstart, openrc, rc systems will all be confused into init

le9i0nx:
decided

PR #12024, opened 2015-08-20
Add win_splitdrive filter for windows users

jhawkesworth:
Another little filter addition of use to those targeting windows.

Its more advanced usage than other similar filters as it returns a list of 2 items, but I put example usage in the docs so hopefully not too confusing.


jhawkesworth:
Thanks for super speedy merge @bcoca !

PR #11767, opened 2015-07-28
add option to ansible-vault to read new password from file for rekey

amenonsen:
The --new-vault-password-file option works the same as
--vault-password-file but applies only to rekeying (when
--vault-password-file sets the old password). Also update the manpage
to document these options more fully.

PR #12032, opened 2015-08-21
Allow vault to operate from stdin/to stdout (i.e. plaintext never hits the disk)

amenonsen:
This is something I've mentioned on IRC a few times, but not submitted as a PR before. With this, and the semi-related #11767, it's possible to automate various vault operations without ever leaking plaintext to disk, even in the form of temp files.

### DIFF
--- a/docs/man/man1/ansible-vault.1
+++ b/docs/man/man1/ansible-vault.1
@@ -1,13 +1,13 @@
 '\" t
 .\"     Title: ansible-vault
 .\"    Author: [see the "AUTHOR" section]
-.\" Generator: DocBook XSL Stylesheets v1.78.1 <http://docbook.sf.net/>
-.\"      Date: 07/28/2015
+.\" Generator: DocBook XSL Stylesheets v1.76.1 <http://docbook.sf.net/>
+.\"      Date: 08/26/2015
 .\"    Manual: System administration commands
 .\"    Source: Ansible 2.0.0
 .\"  Language: English
 .\"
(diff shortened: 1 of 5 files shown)


bcoca:
So instead of making these command line options, i would just trigger this when stdin/stdout are not a tty.  We already do this detection for vars_prompt/pause avoidance, using color on the terminal or using the pager in ansible-doc

amenonsen:
I've added a commit that sets `self.options.read_from_stdin` or `.write_to_stdout` based on whether stdin and stdout `isatty()` or not. As discussed on IRC, I'd prefer to be able to continue to explicitly specify the behaviour on the command-line in some cases. @abadger approved of the idea, so here it is.

amenonsen:
After some discussion on IRC between @abadger and @bcoca and myself, I started to implement a version that allowed `ansible-vault encrypt input.yml --output output.yml` as well as `cat input.yml|ansible-vault encrypt --output output.yml` and the equivalents for decryption. Unfortunately, this would be a very intrusive change because of the way the code is now.

So here's an implementation that adds a new `ansible-vault filter` subcommand that _always_ reads from stdin and always writes to stdout. Backwards compatibility, easy to document, no ambiguity, and a simpler change to review as well.

What do you think of this?

abadger:
@amenonsen I like the feature but I do think it belongs in encrypt/decrypt rather than a separate subcommand.  Mind if I give it a shot?

amenonsen:
See #12112 instead for the requested implementation.

PR #12036, opened 2015-08-21
get_mount_facts -- find bind mounts and add info to options field

wolfgangkarall:
Since many recent Linux distributions use a symlink /etc/mtab -> /proc/mounts the information about bind mounts is missing from the ansible_mounts fact. This fix uses the findmnt command (also available on recent distributions) to check for bind mounts and adds the info back to the options field, so that it can be used in a conditional in templates (like when monitoring file system disk space, no need to monitor the bind-mounted mount points too).


lib/ansible/module_utils/facts.py
```diff
@@ -937,6 +937,18 @@ def get_dmi_facts(self):
     @timeout(10)
     def get_mount_facts(self):
         self.facts['mounts'] = []
+        bind_mounts = []
+        findmntPath = module.get_bin_path("findmnt")
+        rc, out, err = module.run_command("%s -lnur" % ( findmntPath ), use_unsafe_shell=True)
```
bcoca: i would make this all conditional to findmnt being found or it will error out on systems that don't have it

wolfgangkarall: I tested against Debian squeeze systems (which don't have findmnt) and it worked, because run_command returns rc != 0 anyway on non-existing commands.

But I guess it's cleaner to not rely on that behavior, I'll update the code (but not today).

v1/ansible/module_utils/facts.py
```diff
@@ -894,6 +894,18 @@ def get_dmi_facts(self):
     @timeout(10)
     def get_mount_facts(self):
         self.facts['mounts'] = []
+        bind_mounts = []
```
bcoca: v1 is dead code, no updates to it are taken, it is there just for being able to quickly test if a bug was preexisting or caused by the v2 engine

wolfgangkarall:
Any chance to get this merged at some point?

PR #12054, opened 2015-08-23
Support additional options for WinRM connections via inventory variables

cchurch:
As discussed in https://groups.google.com/d/topic/ansible-devel/CpZ0c7na9cI/discussion, will allow explicitly specifying scheme (https or http) and transport (kerberos or plaintext) for additional configuration of WinRM connection. cc @trondhindenes 

Would also allow specifying custom paths to the WinRM endpoint to support multiple hosts behind an HTTP(S) proxy, as discussed in https://groups.google.com/d/topic/ansible-devel/BHxcyxamiuI/discussion cc @blakfeld

### DIFF
--- a/docsite/rst/intro_windows.rst
+++ b/docsite/rst/intro_windows.rst
@@ -160,11 +160,11 @@ In group_vars/windows.yml, define the following inventory variables::
     # ansible-vault edit group_vars/windows.yml
 
     ansible_user: Administrator
-    ansible_ssh_pass: SecretPasswordGoesHere
+    ansible_password: SecretPasswordGoesHere
     ansible_port: 5986
     ansible_connection: winrm
 
-Notice that the ssh_port is not actually for SSH, but this is a holdover variable name from how Ansible is mostly an SSH-oriented system.  Again, Windows management will not happen over SSH (`yet <http://blogs.msdn.com/b/powershell/archive/2015/06/03/looking-forward-microsoft-support-for-secure-shell-ssh.aspx>`).
+Although Ansible is mostly an SSH-oriented system, Windows management will not happen over SSH (`yet <http://blogs.msdn.com/b/powershell/archive/2015/06/03/looking-forward-microsoft-support-for-secure-shell-ssh.aspx>`).
 
 If you have installed the ``kerberos`` module and ``ansible_user`` contains ``@`` (e.g. ``username@realm``), Ansible will first attempt Kerberos authentication. *This method uses the principal you are authenticated to Kerberos with on the control machine and not ``ansible_user``*. If that fails, either because you are not signed into Kerberos on the control machine or because the corresponding domain account on the remote host is not available, then Ansible will fall back to "plain" username/password authentication.
 
@@ -184,11 +184,7 @@ You'll run this command again later though, to make sure everything is working.
 Since 2.0, the following custom inventory variables are also supported for additional configuration of WinRM connections::
 
 * ``ansible_winrm_scheme``: Specify the connection scheme (``http`` or ``https``) to use for the WinRM connection.  Ansible uses ``https`` by default unless the port is 5985.
-* ``ansible_winrm_host``: May be used as an alias for ``ansible_ssh_host`` and will take precedence if both are defined.
-* ``ansible_winrm_port``: May be used an alias for ``ansible_ssh_port`` and will take precedence if both are defined.
 * ``ansible_winrm_path``: Specify an alternate path to the WinRM endpoint.  Ansible uses ``/wsman`` by default.
-* ``ansible_winrm_user``: May be used an alias for ``ansible_ssh_user`` and will take precedence if both are defined.
-* ``ansible_winrm_pass``: May be used an alias for ``ansible_ssh_pass`` and will take precedence if both are defined.
 * ``ansible_winrm_realm``: Specify the realm to use for Kerberos authentication.  If the username contains ``@``, Ansible will use the part of the username after ``@`` by default.
 * ``ansible_winrm_transport``: Specify one or more transports as a comma-separated list.  By default, Ansible will use ``kerberos,plaintext`` if the ``kerberos`` module is installed and a realm is defined, otherwise ``plaintext``.
 * ``ansible_winrm_*``: Any additional keyword arguments supported by ``winrm.Protocol`` may be provided.
--- a/docsite/rst/intro_windows.rst
+++ b/docsite/rst/intro_windows.rst
@@ -164,7 +164,7 @@ In group_vars/windows.yml, define the following inventory variables::
     ansible_port: 5986
     ansible_connection: winrm
 
-Notice that the ssh_port is not actually for SSH, but this is a holdover variable name from how Ansible is mostly an SSH-oriented system.  Again, Windows management will not happen over SSH.
+Notice that the ssh_port is not actually for SSH, but this is a holdover variable name from how Ansible is mostly an SSH-oriented system.  Again, Windows management will not happen over SSH (`yet <http://blogs.msdn.com/b/powershell/archive/2015/06/03/looking-forward-microsoft-support-for-secure-shell-ssh.aspx>`).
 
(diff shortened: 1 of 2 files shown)


trondhindenes:
Great stuff! Maybe provide some extra logging/warning output if user specifies kerberos as scheme and the kerberos python package isnt installed?

blakfeld:
You are my hero! This is awesome!

I'm unsure if it's related, but in a separate project my team has run into an issue where if pywinrm expects Kerberos and it isn't installed it will just silently fail, so +1 to the warning.

docsite/rst/intro_windows.rst
```diff
@@ -108,6 +108,18 @@ a version that is 3 or higher.
 
 You'll run this command again later though, to make sure everything is working.
 
+Since 2.0, the following custom inventory variables are also supported for additional configuration of WinRM connections::
+
+* ``ansible_winrm_scheme``: Specify the connection scheme (``http`` or ``https``) to use for the WinRM connection.  Ansible uses ``https`` by default unless the port is 5985.
+* ``ansible_winrm_host``: May be used as an alias for ``ansible_ssh_host`` and will take precedence if both are defined.
+* ``ansible_winrm_port``: May be used an alias for ``ansible_ssh_port`` and will take precedence if both are defined.
+* ``ansible_winrm_path``: Specify an alternate path to the WinRM endpoint.  Ansible uses ``/wsman`` by default.
+* ``ansible_winrm_user``: May be used an alias for ``ansible_ssh_user`` and will take precedence if both are defined.
```
bcoca: i still would prefer to use the existing neutral: ansible_host, ansible_port, ansible_user, etc

cchurch:
I've added a new commit (b437858) to remove the winrm options that are already supported in a connection-independent way (host, port, user, password), provide a clear error message if kerberos is not installed, and to better handle exceptions from trying to connect using each transport.

blakfeld:
With this commit, can I have multiple machines behind the same proxy? I'm essentially attempting to do:

```
proxy_url.com/host1
proxy_url.com /host2
```

So from reading the commit, I can put in my inventory 

```
proxy_url.com ansible_winrm_path=host1
proxy_url.com ansible_winrm_path=host2
```

But then all I get is the access to host2, which makes sense. Is there a way I'm missing to have both entries get read?

trondhindenes:
You mean invoking commands thru a jump host? Nope, that's not supported for Windows nodes.

blakfeld:
Damn, thanks for the confirmation @trondhindenes. 

I've got an NGINX proxy set up that will route the WinRm calls to multiple backends happily, but if I cant specify the same host with different WinRm paths in an inventory I'll have to hack something together.

trondhindenes:
Sorry to be the bearer of bad news etc. The winrm_path variable is used for hosts not listening to the default http(s)://hostname/wsman endpoint. Very edgy case, in other words. You could set up your nginx proxy to do port forwarding maybe? port 9001-->5986 on node1, port 9002-->5986 on node2, etc? I think that's the closest you're gonna get with direct access to the nodes.

trondhindenes:
@blakfeld I might be wrong. I just looked at the original PR now. I haven't tested this, but as far as I can see you should be able to setup url-based forwarding. So f.ex https://nginx/wsman-node1 --> node1 and https://nginx/wsman_node2 -->node2 etc. Never tried it tho. Maybe @cchurch has?

blakfeld:
Interesting @trondhindenes I'm poking around with that now. What it looks like I'm seeing is that whatever was defined last in the inventory is what is used, like a dict is being updated, and the hostname is the key. I'll dig around and see what I come up with!

EDIT:

I get a "Can't parse 'hostname' as host[:port]" error. I bet the winrm connector code supports what I want to do, but the inventory parsing bits don't...

EDIT 2:

Boom: https://github.com/ansible/ansible/blob/fe0e4f27ad82c4b5844a3a18922fa1857242a33b/lib/ansible/inventory/ini.py#L268-L269

I'll play with that and put together a PR if I can fix it.

trondhindenes:
gotcha. I don't know what the relation is between ansible_winrm_path and the actual connection url, but since it defaults to "/wsman" its probably parsed as  `<scheme><hostname>:<port><ansible_winrm_path>`. Would be cool to make this work thru proxies, but my python skills stop just north of print('hello').

trondhindenes:
Maybe set  `ansible_host` to the fqdn of the proxy? That would hopefully end up as a parseable wsman url?

blakfeld:
BOOM! @trondhindenes That totally works!

Heres what my inventory looks like:

```
servicename ansible_host=proxy.fqdn.company.com ansible_winrm_path=servicename
```

trondhindenes:
Nice! So your forwarding "/servicename" to "/wsman" on the backend windows node? That is so cool!

blakfeld:
Exactly! I have an nginx proxy set up like:

```
server {
    listen 5986 ssl;
    server_name localhost;

    ssl_certificate ...
    ssl_certificate_key ...

    location /service_name { proxy_pass https://service_ip:5986/wsman }
}
```

blakfeld:
Exactly! I have an nginx proxy set up like:

```
server {
    listen 5986 ssl;
    server_name localhost;

    ssl_certificate ...
    ssl_certificate_key ...

    location /service_name { proxy_pass https://service_ip:5986/wsman }
}
```

one1zero1one:
Works like a charm, thank you @blakfeld . Using the same jumpbox for lnx and win is awesome.

Side node, helps to have larger timeouts in nginix to accommodate waiting times.

```
    proxy_connect_timeout 1h;
    proxy_send_timeout 1h;
    proxy_read_timeout 1h;

```

hehe2:
Well, my question might sound dumb as I'm kinda new to ansible but...

In your nginx proxy set up, where does the "service_ip" comes from ?

Let's say I have two windows servers I'd like to manage using ansible but behind an nginx proxy.

In my ansible host inventory I will need to add something like :

win_host1 ansible_host=my_proxy_host ansible_winrm_path=win_host1
win_host2 ansible_host=my_proxy_host ansible_winrm_path=win_host2

and in my nginx conf I will need something like :

location /win_host1 { proxy_pass https://hardcoded_win_host1_ip:5986/wsman }
location /win_host2 { proxy_pass https://hardcoded_win_host2_ip:5986/wsman }

?

That's not very clear for me...

Any help would be greatly appreciated...

blakfeld:
@hehe2 The service ip in this instance is just the IP of the target box I'm trying to get to. So the flow looks like this:

Ansible Master -> Proxy -> service_ip

rbq:
Is there a way to use an SSH bastion host to access a WinRM machine (without setting up a SSH tunnel manually)?

hehe2:
Well, I've achieved to get my windows nodes behind a proxy while being able to specify it directly into the Ansible inventory.

I've added a new attribute in the "pywinrm" module but it's not been accepted in the main tree so you can use it but "at your own risks". You can find it here : https://github.com/diyan/pywinrm/pull/165

The evolution I've made is trivial though, just passing by two new attributes to the "request" module, used by pywinrm to handle the WinRM connection.

Basically, those two variables are "http_proxy" and "https_proxy" which are already supported by the "request" python module.

Since Ansible pass all the variables starting with "ansible_winrm_" to the pywinrm module after removing the "ansible_winrm_" prefix, I've just hooked the two matching variables : 

ansible_winrm_http_proxy and ansible_winrm_https_proxy.

This way, since the request modules supports the SOCKS5 proxy type using the https_proxy string format

"socks5://socks_user:socks_password@socks_server:socks_port" it's very convenient.

Using a dante SOCKS5 server on my proxy I can pass WinRM connections that are relying on windows certificates (so, not using user/pw to connect to windows hosts).

I wish I also could use this SOCKS5 even for Unixes targets based on SSH without having to describe a bastion config for all of them but at the moment, I've not found a way to make this possible digging into the Ansible code (I've not spent much time on this though since creating the ssh config is not that hard).

It would be support convenient to be able to describe it the same way whatever the OS like this ;

ansible_http_proxy  (or ansible_https_proxy) and know that we could handle the socks5 kinda proxy string, like the one already supported by the request module.

You could have one single way of describing your servers groups proxies in your inventory whatever the OS type... That would be magical =)

PR #12057, opened 2015-08-23
Deprecate groupname[x-y] syntax in favour of groupname[x:y]

amenonsen:
This series of PRs cleans up and documents various aspects of the parsing of `hosts: groupname[x:y]` and deprecates the earlier `[x-y]` subscript syntax in favour of the inclusive `[x:y]` syntax. That's the only real functional change, though there are various simplifications/improvements to the code. This is based on extensive discussion on IRC.

I had originally planned to make it possible to use `nongroupname[x:y]` range syntax to match hostnames the way they can be specified when defining the inventory, but it became clear that the ambiguities introduced by the change in light of the existing slicing syntax as well as the `fnmatch` character classes made this more complicated than I wanted.

This finally supersedes #10543 by @willthames.

cc: @bcoca @abadger

### DIFF
--- a/docsite/rst/intro_patterns.rst
+++ b/docsite/rst/intro_patterns.rst
@@ -68,13 +68,19 @@ It's also ok to mix wildcard patterns and groups at the same time::
 
     one*.com:dbservers
 
-As an advanced usage, you can also select the numbered server in a group::
-   
-    webservers[0]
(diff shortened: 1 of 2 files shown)


jimi-c:
Merged, thanks!

PR #12064, opened 2015-08-24
Add regex_capture jinja2 filter.

abrefort:
##### SUMMARY
This pull request adds a Jinja2 filter to get the value of the first capturing group within a regular expression.
The use case is reusing variables and command returns directly while being more flexible and limiting the number of grep/sed/awk workarounds.

I wrote an integration test in this pull request, but integration tests for filters do not seem to be currently enabled. I think this issue belongs to another pull request.

Sincerely,

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
docsite/rst/playbooks_filters.rst
lib/ansible/plugins/filter/core.py
test/integration/roles/test_filters/files/foo.txt
test/integration/roles/test_filters/templates/foo.j2

##### ANSIBLE VERSION
2.3

### DIFF
--- a/docsite/rst/playbooks_filters.rst
+++ b/docsite/rst/playbooks_filters.rst
@@ -528,6 +528,11 @@ To escape special characters within a regex, use the "regex_escape" filter::
     # convert '^f.*o(.*)$' to '\^f\.\*o\(\.\*\)\$'
     {{ '^f.*o(.*)$' | regex_escape() }}
 
+To get the value of capturing group within a regex, use the "regex_capture" filter::
+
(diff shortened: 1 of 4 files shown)


abrefort:
Fixed merge conflict and rebased on top of devel.

abrefort:
Fixed merge conflict and rebased on top of devel.

amenonsen:
I think this could be useful in certain circumstances, but I don't like the name much. extract_match or something would be better. And I think it would be more useful if it were possible to extract multiple results.

PR #12075, opened 2015-08-24
Unicode and other fixes for vault

abadger:
The python3 work on vault broke vault's handling of unicode.  This patch fixes vault so that it only deals in byte strings inside of the Ciphers.  VaultLib and VaultEditor can take in either byte strings or unicode strings and will convert (using utf-8) before passing the data on to the ciphers.

Nonessential changes include
- prefixing variables in VaultLib containing byte strings with b_ so that it's easier to tell what the contents of a var are.  VaultLib is the only class that needs to know whether its dealing with bytes or unicode (the other classes should only be given byte strings by VaultLib or else pass their data on to VaultLib to do the conversion) so it seemed like the class most in need of this.
- Make sure that when we read in a file in parsing/**init**.py that it is read as a byte string nad then we convert it to a unicode type (either via VaultLib or inside of that method if it's paintext).
- Convert lists to frozensets since they are only being used to do containment tests.
- Correct a few typos in the to-be-used-in-the-future VaultFile code
- Fix the optimization that does not resave unchanged vault files to still resave the file if the old vault was VaultAES instead of VaultAES256.
- Made conversions from bytes to unicode throw errors instead of replacing unknown chars with "?".  In this case we're dealing with very precise formats so we don't want to mess those up here. 
- Set padding bytes to encode in the ascii encoding... this should never matter (I believe that the character can only be codepoint 0-16 here) but the code will definitely do the wrong thing if the padding is a multibyte character so doing this may catch future coding errors.

### DIFF
--- a/lib/ansible/parsing/vault/__init__.py
+++ b/lib/ansible/parsing/vault/__init__.py
@@ -22,7 +22,6 @@
 from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
-import sys
 import os
 import shlex
 import shutil
@@ -33,7 +32,7 @@ from ansible.errors import AnsibleError
 from hashlib import sha256
 from binascii import hexlify
 from binascii import unhexlify
-from six import binary_type, PY3, text_type
+from six import PY3
 
 # Note: Only used for loading obsolete VaultAES files.  All files are written
 # using the newer VaultAES256 which does not require md5
@@ -95,8 +94,9 @@ HAS_ANY_PBKDF2HMAC = HAS_PBKDF2 or HAS_PBKDF2HMAC
 
 CRYPTO_UPGRADE = "ansible-vault requires a newer version of pycrypto than the one installed on your platform. You may fix this with OS-specific commands such as: yum install python-devel; rpm -e --nodeps python-crypto; pip install pycrypto"
 
-HEADER=u'$ANSIBLE_VAULT'
-CIPHER_WHITELIST=['AES', 'AES256']
+b_HEADER = b'$ANSIBLE_VAULT'
+CIPHER_WHITELIST = frozenset((u'AES', u'AES256'))
+CIPHER_WRITE_WHITELIST=frozenset((u'AES256',))
 
 
 def check_prereqs():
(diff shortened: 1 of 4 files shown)


amenonsen:
Looks good to me, barring a couple of trivial typos that I've commented on against the commit. I think this is good to merge. I'll use this as a base for my further changes.

lib/ansible/parsing/vault/__init__.py
```diff
@@ -104,116 +104,132 @@ def check_prereqs():
     if not HAS_AES or not HAS_COUNTER or not HAS_ANY_PBKDF2HMAC or not HAS_HASH:
         raise AnsibleError(CRYPTO_UPGRADE)
 
-class VaultLib(object):
+class VaultLib:
 
     def __init__(self, password):
-        self.password = password
+        self.b_password = to_bytes(password, errors='strict', encoding='utf-8')
```
bcoca, line 110: shouldn't we do this on password input? as this is an issue with all passwrods

lib/ansible/plugins/action/copy.py
```diff
@@ -29,7 +29,6 @@
 from ansible.utils.boolean import boolean
 from ansible.utils.hashing import checksum
 from ansible.utils.unicode import to_bytes
-from ansible.parsing.vault import VaultLib
```
bcoca, line 32: this is here for when copy will transparently handle vault files

lib/ansible/parsing/vault/__init__.py
abadger, line 110: Probably not, but passwords are a little special so a case could be made:   In general we want our strategy to be unicode type internally and bytes type when we actually use it.  That way if we manipulate the string in some way (measure its length, concatenate it, strip whitespace characters, etc) there's no problems with bytes doing the wrong thing or getting mixed with unicode.  By keeping everything unicode type internally there's no confusion of what type you're dealing with and it keeps unnecessary conversions to a minimum (to_bytes and to_unicode are pretty good about unnecessary conversions but we still have the function call overhead and a few conditionals everytime we call it.)

For passwords, we do very few of these things.  So it may make sense to carry the byte type all the way through from input to output.  However, that requires tracing a lot of code, finding all the places we input vault passwords and making sure that they maintain the proper type all the way through.

Since the vault code that uses the password as a byte string is self-contained inside or behind VaultLib, I think it makes sense to do this transformation here and then check the rest of the code at our leisure later.

lib/ansible/plugins/action/copy.py
abadger, line 32: it's a simple thing to re-add when we have some code that actually makes use of it (and if we implement this via VaultFile, it won't even be the same line that we need).

abadger:
@amenonsen, @bcoca Thanks for the reviews!

PR #12078, opened 2015-08-25
Add hash_merge and hash_merge_recursive filters with documentation

amenonsen:
This PR implements the most-requested subset of the earlier (closed) PR
largely unchanged from PR #7872 by @darkk (but the filters are named
differently to address the fact that hash_behaviour=replace doesn't
actually do a shallow merge, as the earlier PR implied).

Closes #7872 by @darkk (hash_merge/hash_replace filters)
Closes #11153 by @telbizov (merged_dicts lookup plugin)

cc: @evanccnyc

### DIFF
--- a/docsite/rst/intro_configuration.rst
+++ b/docsite/rst/intro_configuration.rst
@@ -334,6 +334,11 @@ official examples repos do not use this setting::
 
 The valid values are either 'replace' (the default) or 'merge'.
 
+.. versionadded: '2.0'
+
+If you want to merge hashes without changing the global settings, use
+the `combine` filter described in :doc:`playbooks_filters`.
+
 .. _hostfile:
 
 hostfile
--- a/docsite/rst/playbooks_filters.rst
+++ b/docsite/rst/playbooks_filters.rst
@@ -316,6 +316,41 @@ To get a sha256 password hash with a specific salt::
 Hash types available depend on the master system running ansible,
 'hash' depends on hashlib password_hash depends on crypt.
 
+.. _combine_filter:
+
+Combining hashes/dictionaries
+-----------------------------
+
+.. versionadded:: 2.0
+
+The `combine` filter allows hashes to be merged. For example, the
+following would override keys in one hash:
+
+    {{ {'a':1, 'b':2}|combine({'b':3}) }}
+
+The resulting hash would be:
+
+    {'a':1, 'b':3}
+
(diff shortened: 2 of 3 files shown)


evanccnyc:
Be nice to have this but I am not sure where @bcoca and @abadger ended up.

abadger:
This looks okay to me.  I take it you couldn't get arbitrary numbers of dict params to play well with recursive being an optional keywpord arg?

bcoca:
i would have preferred something more neutral as some people know what a hash is, some people know them as dictionaries or associative arrays and many people don't know any of the above.

But adding this to the set theory filters has been declined ...

amenonsen:
@abadger that's right, I could have done `(recursive=False, *terms)`, but then it wouldn't have been optional; and `(*terms, recursive=False)` isn't allowed. I could have passed in a list argument and an optional boolean, but at that point the awkwardness seemed to outweigh the benefits.

@bcoca I stuck with `hash_merge*` in particular because of the correspondence with `hash_behaviour` in `ansible.cfg`. Given how many times this feature has been asked for and reimplemented in various different ways, I don't think the nomenclature is a major sticking point.

abadger:
Looks good.  We talked about this at the internal team meeting today and implementation seems good.  Just want to rename for @bcoca's preference for something neutral.  Either merge() and merge_recursive() or combine() and combine_recursive() were suggested.  bcoca also thought this would work for other iterables, not just dictionaries which is why he wanted the more neutral naming.  After some discussion he convinced us that hash, dict, and associative array are all jargon (not all of our users know about them) so it would be better to add support for other iterables to these filters and change the name rather than keep the name and keep the filters limited to only dicts.

I can make those changes after merge if you want or you can do it yourself and then I'll merge.  Let me know your preference.  I think I may have an idea of how to get recursive to work as a parameter too that I might try.  Something like:

``` python
def combine(*terms, **kwargs):
    recursive = kwargs.get('recursive', False)
    if len(kwargs) > 1 or ('recursive' not in kwargs and len(kwargs) == 1):
        # Error about invalid parameter
```

amenonsen:
I've pushed a revised commit along the lines explained by @abadger.

abadger:
Phew!  Thank you to everyone who worked on this problem!  And thank you to @amenonsen for seeing it through to completion!

# Merged

Hi!

issue is resolved.

Thank you!

PR #12090, opened 2015-08-25
Update cobbler.py inventory script to more fully leverage Cobbler management fields

gregmark:
##### SUMMARY
Refactored to leverage Cobbler system, profile, and management class parameters to produce a more robust and exact output; all configurable in cobbler.ini

```
    o  assigning system objects to groups based on owners and/or status and/or profiles,
       and/or management classes.
    o  assigning groups to parent groups based on the special management class parameter
       "parent". Only one parent per class is currently supported.
    o  assigning group and hostvars based on either or both ks_meta params or management
       parameters.
    o  Optional: have system objects inherit vars from their respective profiles
    o  Optional: Only include system objects with at least one interface w/ management = 1
    o  Limitations: there is no good way to assign groupvars or children to groups based on
       owners or status. Currently, parents can not be assigned in profiles and only one
       parent can be assigned per child group (this could be added easily).
    Tested with Cobbler 2.6.7.
```

Also, I refrained making two other changes before I got some maintainer feedback.  First, while the script will still support earlier Cobbler versions that lack management groups, it seems to me that using the system object's management tab as the primary driver of the Ansible inventory is the best practice moving forward. With that in mind, I only coded parent/children management to be based on "management class params", not ks_meta or mgmt_parameters from systems or profiles.

Secondly, a next good step should be to add a "parents" field to the management class dict, separate from mgmt_class[params], where multiple parents are separated by whitespace. I wasn't sure if we should support multiple parents before that change, but if we did, I also wasn't sure what the best "Cobbler way" to make this assignment. Either comma separated, whitespace, or something else?

Finally, this is the most python coding I have ever done. I am very green, so please provide suggestions if my code is messy or too complicated.

Thanks.

--Greg Chavez

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
contrib/inventory/cobbler.ini
contrib/inventory/cobbler.py

##### ANSIBLE VERSION
2.3

### DIFF
--- a/contrib/inventory/cobbler.py
+++ b/contrib/inventory/cobbler.py
@@ -35,14 +35,16 @@ Changelog:
         o  assigning system objects to groups based on owners and/or status and/or profiles,
            and/or management classes. 
         o  assigning groups to parent groups based on the special management class parameter
-           "parent". Only one parent per class is currently supported.
+           "parent". Only one parent per class is currently supported. Profile mgmt classes
+            can be inherited, optionally.
         o  assigning group and hostvars based on either or both ks_meta params or management
            parameters.
         o  Optional: have system objects inherit vars from their respective profiles  
         o  Optional: Only include system objects with at least one interface w/ management = 1
         o  Limitations: there is no good way to assign groupvars or children to groups based on
            owners or status. Currently, parents can not be assigned in profiles and only one
-           parent can be assigned per child group (this could be added easily).
+           parent can be assigned per child group (this could be added easily). inherit_profiles
+           and profile_groups are mutually exclusive, with the former preferred.
         Tested with Cobbler 2.6.7.
 
     - 2015-06-21 dmccue: Modified to support run-once _meta retrieval, results in
@@ -81,6 +83,7 @@ import os
 import re
 from time import time
 import xmlrpclib
+import pprint
 
 try:
     import json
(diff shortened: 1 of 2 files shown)


contrib/inventory/cobbler.py
```diff
@@ -159,49 +187,156 @@ def update_cache(self):
...
-
+            owners = host['owners']
             status = host['status']
             profile = host['profile']
-            classes = host[orderby_keyname]
+            classes = host['mgmt_classes']
+
+            # Prevent redundancy
+            if self.profile_groups and self.inherit_profiles:
+                self.inherit_pofiles = False
```
dcoshea, line 231: Typo "pofiles" should be "profiles"

gregmark:
Will do. I had begun to think this pull request would never happen. I'll start fixing these problems.

gregmark:
Whoa! Did not mean to close this.

thaumos:
hey @gregmark, would you be willing to do a rebase on this so it fires off another test?

dcoshea:
It would be nice if there was a way to request that all parameters be passed through to Ansible, not just `ksmeta` and `mgmt_params`.  For example, I'd like Ansible to be able to see all of the network interfaces for a machine as configured in Cobbler, and if I recall correctly (it's been a while since I looked at this) this wasn't available.

k4pu77:
@dcoshea - I was looking for the same functionality as you and can confirm that the configured network interfaces are listed when you run the script and therefore available to ansible. Even in version in the current devel branch. You can do stuff like:
```
- debug: var=cobbler['interfaces']['eth0']['ip_address']
```
There are additional parameters exported. 

Addition: I took me felt ages to realize that if you change `orderby_keyname = 'owners'` to `orderby_keyname = 'mgmt_classes'` in the inventory script, the hosts will be grouped by management class which is a pretty handy feature. No clue why this is not a configurable property.

gregmark:
OK, let's see if I can get this thing working again.

sandinak:
+1 on this .. it really needs some attention and to be integrated.

gundalow:
Looked at this during PR review day,

Looks like people would like this though it needs some work and testing.
Also the source branch (fork) no longer exists, so maybe someone wants to take over this and create a fresh PR?

PR #12093, opened 2015-08-25
Update smart transport to handle Sun_SSH_1.5 on SmartOS

travispaul:
On SmartOS the smart transport's check for the ControlPersist option doesn't behave as intended.  In the Sun_SSH implementation, the usage information is returned instead of "Bad configuration option".

### DIFF
--- a/v1/ansible/runner/__init__.py
+++ b/v1/ansible/runner/__init__.py
@@ -232,7 +232,7 @@ class Runner(object):
                 # see if SSH can support ControlPersist if not use paramiko
                 cmd = subprocess.Popen(['ssh','-o','ControlPersist'], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
                 (out, err) = cmd.communicate()
-                if "Bad configuration option" in err:
+                if "Bad configuration option" in err or "Usage:" in err:
                     self.transport = "paramiko"
 
         # save the original transport, in case it gets


travispaul:
For example the output on SmartOS:

```
[root@ansible ~]# ssh -o ControlPersist
Usage: ssh [options] host [command]
Options:
  -l user     Log in using this user name.
  -n          Redirect input from /dev/null.
  -F config   Config file (default: ~/.ssh/config).
  -A          Enable authentication agent forwarding.
  -a          Disable authentication agent forwarding (default).
  -X          Enable X11 connection forwarding.
  -x          Disable X11 connection forwarding (default).
  -i file     Identity for public key authentication (default: ~/.ssh/identity)
  -t          Tty; allocate a tty even if command is given.
  -T          Do not allocate a tty.
  -v          Verbose; display verbose debugging messages.
              Multiple -v increases verbosity.
  -V          Display version number only.
  -q          Quiet; don't display any warning messages.
  -f          Fork into background after authentication.
  -e char     Set escape character; ``none'' = disable (default: ~).
  -c cipher   Select encryption algorithm
  -m macs     Specify MAC algorithms for protocol version 2.
  -p port     Connect to this port.  Server must be on the same port.
  -L listen-port:host:port   Forward local port to remote address
  -R listen-port:host:port   Forward remote port to local address
              These cause ssh to listen for connections on a port, and
              forward them to the other side by connecting to host:port.
  -D port     Enable dynamic application-level port forwarding.
  -C          Enable compression.
  -N          Do not execute a shell or command.
  -g          Allow remote hosts to connect to forwarded ports.
  -1          Force protocol version 1.
  -2          Force protocol version 2.
  -4          Use IPv4 only.
  -6          Use IPv6 only.
  -o 'option' Process the option as if it was read from a configuration file.
  -s          Invoke command (mandatory) as SSH2 subsystem.
  -b addr     Local IP address.

```

bcoca:
so, the v1/ tree is dead code, its there only for testing when a bug was introduced and will never be released so this change is not useful.  Please resubmit against the proper code.

travispaul:
Sorry about that.  I'll open a new pull request against stable-1.9.

bcoca:
that might not help you either as 1.9.3 is already in RC and it is very unlikely to be a 1.9.4

travispaul:
Where should I commit this change?  Since there appears to be a 2.0 rewrite going on (today is my first day of attempting to use Ansible so this is news to me) should I just patch this locally and be on my way?

bcoca:
in lib/ansible/executor/task_executor.py

PR #12098, opened 2015-08-26
Don't compile .py files under .tox/

mgedmin:
'tox -e py26' would fail for me because python -m compileall would crawl under
.tox/py27 and, unsurprisingly, get SyntaxErrors on files from the Python 2.7
standard library using syntax features not supported by Python 2.6.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,7 +9,7 @@ whitelist_externals = make
 [testenv:py26]
 commands =
     python --version
-    python -m compileall -fq -x 'test|samples|contrib/inventory/vagrant.py' .
+    python -m compileall -fq -x 'test|samples|contrib/inventory/vagrant.py|.tox' .
     make tests
 deps = -r{toxinidir}/test-requirements.txt
 whitelist_externals =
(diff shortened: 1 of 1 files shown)


abadger:
@sivel could you double check this for me?  I think it's correct but I'm not sure why I have site-packages/ansible underneath the .tox/ directories.  just wanted to double check that tox isn't running the checks on the ansible directory that's in there.

sivel:
The `.tox` directories are basically virtual environments where all deps and the package itself are installed to.  Although, it is strange that this would be needed.  I haven't seen this actually failing anywhere and I cannot duplicate this issue.

Although I cannot see how adding this would hurt.

tox.ini
```diff
@@ -26,7 +26,7 @@ whitelist_externals = make
 [testenv:py34]
 commands =
     python --version
-    python -m compileall -fq -x 'lib/ansible/module_utils' lib
+    python -m compileall -fq -x 'lib/ansible/module_utils|.tox' lib
```
mgedmin, line 29: This bit is actually unnecessary, because it's crawling under `./lib/`, not directly in `.`.  I didn't notice that when I was preparing my patch.

In fact, changing `.` to `lib` in the other compileall invocations would perhaps be a better way of fixing this problem.

abadger:
@mgedmin Oops, I should have hit refresh before hitting the merge button.  Yeah, changing . to lib probably is the "more correct" fix.  I'll whip that up.  Thanks for finding and fixing this!

@sivel thanks for making sure it wouldn't have any hidden gotchas!

PR #12107, opened 2015-08-26
Update "smart" transport to handle Sun_SSH_1.5 on SmartOS

travispaul:
On SmartOS the smart transport's check for the ControlPersist option doesn't behave as intended. In the Sun_SSH implementation, the usage information is returned instead of "Bad configuration option".

My original, incorrect PR: https://github.com/ansible/ansible/pull/12093


jimi-c:
Merged, thanks!

PR #12112, opened 2015-08-26
Implement cat-like filtering behaviour for encrypt/decrypt

amenonsen:
As discussed on IRC, this includes some minor functionality-preserving cleanups for VaultEditor, and uses this to implement the stdin/stdout encryption/decryption semantics requested by @bcoca and @abadger.

### DIFF
--- a/lib/ansible/cli/vault.py
+++ b/lib/ansible/cli/vault.py
@@ -63,7 +63,19 @@ class VaultCLI(CLI):
         self.options, self.args = self.parser.parse_args()
         self.display.verbosity = self.options.verbosity
 
-        if len(self.args) == 0:
+        if self.options.output_file:
+            if self.action not in ['encrypt','decrypt']:
+                raise AnsibleOptionsError("The --output option can be used only with ansible-vault encrypt/decrypt")
+
+            # This restriction should remain in place until it's possible to
+            # load multiple YAML records from a single file, or it's too easy
+            # to create an encrypted file that can't be read back in. But in
+            # the meanwhile, "cat a b c|ansible-vault encrypt --output x" is
+            # a workaround.
+            if len(self.args) > 1:
+                raise AnsibleOptionsError("At most one input file may be used with the --output option")
+
+        elif len(self.args) == 0:
             raise AnsibleOptionsError("Vault requires at least one filename as a parameter")
 
     def run(self):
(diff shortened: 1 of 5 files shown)


abadger:
future cleanup note:  VaultLib.cipher_name is not a good attribute.  It's being used in two ways, I think:
- as  a property of the specific file/data.  Since we are able to give different filenames to VaultLib.encrypt()/decrypt() now, having a VaultLib level property specifying the cipher of the last data that went through VaultLib doesn't make sense.  Probably better to replace this with a method that returns the cipher_name when given a ciphertext.
- to tell us what cipher to use to encrypt a file.  Because this belongs to the specific file to be encrypted and that can change for each call to encrypt()/decrypt() we probably should pass this in as a parameter to encrypt() instead.

I can't think of any way that self.cipher_name will break current code so we don't have to change it in this PR but it seems conceptually wrong and might be a source of bugs with future code changes.

lib/ansible/cli/vault.py
```diff
@@ -63,7 +63,19 @@ def parse(self):
...
+
+            # This restriction should remain in place until it's possible to
+            # load multiple YAML records from a single file, or it's too easy
+            # to create an encrypted file that can't be read back in. But in
+            # the meanwhile, "cat a b c|ansible-vault encrypt --output x" is
+            # a workaround.
+            if len(self.args) > 1:
+                raise AnsibleOptionsError("At most one input file may be used with the --output option")
+
+        elif len(self.args) == 0:
```
bcoca: so even if i try to input from stdin this will error out?

abadger:
Review and everything here looks good to me.  At first I had some questions in the same area as @bcoca did but I think all of the ways of operating that I'd expect are covered.  Just not quite the same way as cat:

```
$ ansible-vault encrypt filename1   # std way we've always called the command
$ ansible-vault encrypt filename1 -o filename2    # New cmdline arg
$ cat filename1 | ansible-vault encrypt -o filename2   # Stdin to named file
$ ansible-vault encrypt filename1 -o - > filename2     # Output to stdout
$ cat filename1 | ansible-vault encrypt filename1 -o - > filename2  # stdin to stdout
```

There's other UNIX tools that operate this way (using explicit -o- to specify output to stdout) so it looks okay to me.  @bcoca does that summary of the options alleviate your concern?

abadger:
Nice work!  I like the new behaviour a lot.  thanks @amenonsen !

# Merged

Hi!

issue is resolved.

Thank you!

PR #12114, opened 2015-08-26
fixed hostvars access in conjunction with --limit usage

ilya-epifanov:
ignore_limits_and_restrictions is only checked for false value but I decided to leave it this way to emphasize that this is opt-out functionality, not opt-in.


jimi-c:
Merged, thanks!

PR #12118, opened 2015-08-27
Make sanitze_output function catch other forms of passwords

Jonnymcc:
##### SUMMARY
The motivation for this change is that we have some shell commands that contain passwords in the arguments and when running ansible with verbosity -vv in a build environment like Jenkins those passwords would then get captured in the logs. This change detects some other common password keys that one might find in a cli tool argument like those passed to the mongo shell.

Includes strings like
-p foo
password=foo
--password=foo
--password 'foo'

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
lib/ansible/module_utils/basic.py
lib/ansible/utils/__init__.py
test/units/TestUtils.py

##### ANSIBLE VERSION
2.3

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -283,6 +283,7 @@ def heuristic_log_sanitize(data):
     ''' Remove strings that look like passwords from log messages '''
     # Currently filters:
     # user:pass@foo/whatever and http://username:pass@wherever/foo
+    # and strings like -p foo and password=foo
     # This code has false positives and consumes parts of logs that are
     # not passwds
 
@@ -293,6 +294,8 @@ def heuristic_log_sanitize(data):
     #   a passwd
     # sep_search_end: where in the string to end a search for the sep
     output = []
+    PASSWORD_RE = re.compile(r"""((password| -p)[= ]?)['"]?([^ ]*)""")
+    data = PASSWORD_RE.sub(r'\1VALUE_HIDDEN', data)
     begin = len(data)
     prev_begin = begin
     sep = 1
--- a/lib/ansible/utils/__init__.py
+++ b/lib/ansible/utils/__init__.py
@@ -941,26 +941,7 @@ def getch():
 def sanitize_output(arg_string):
     ''' strips private info out of a string '''
 
-    private_keys = ('password', 'login_password')
-
-    output = []
-    for part in arg_string.split():
-        try:
-            (k, v) = part.split('=', 1)
-        except ValueError:
-            v = heuristic_log_sanitize(part)
-            output.append(v)
-            continue
-
-        if k in private_keys:
-            v = 'VALUE_HIDDEN'
-        else:
-            v = heuristic_log_sanitize(v)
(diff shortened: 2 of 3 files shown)


abadger:
heuristic_log_santize() is best effort only and has to walk a line between providing debuggable output and hiding common secret information.  Scanning for -p is definitely too broad a net so we don't want to do that here.

heuristic_log_sanitize() is also used both by modules and controller side so we'd want to make sure that we aren't ruining the module use-case when we make changes that are meant for the controller.

In most cases, even though we provide heuristic_log_sanitize(), we tell people that it's more secure to use no_log: as a task parameter to make sure that secret information doesn't leak out.

We probably would not take this as a fix to 1.9 as no_log is available to do the work and is the official answer to making sure that private information stays private.  But if you want to enhance secure_output to find --password=foo, password="foo" and such that might be acceptable for devel.  Unless we carefully check that heuristic_log_sanitize's other uses aren't negatively impacted we'd probably want that to be put into sanitize_output().

Jonnymcc:
Perhaps my approach is to general and can cause issues elsewhere. I get that heuristic_log_sanitize() is a best effort approach and the case with masking options to flags like -p could hinder debugging. no_log seems to work for most cases, the issue I raised is unrelated to no_log though. The problem being that even if no_log is used, if a playbook is run with a high level of verbosity secrets like password=something are still shown in plaintext.

calfonso:
@Jonnymcc would you trying our devel branch and see if it's still a problem? We made some changes during our 2.3 development cycle that may have already addressed this. Thanks for your interest!

abadger:
```
$ cat test.yml                                                                                                                                                    *[temp-staging-post-2.3.2]  (11:59:43)
---
- hosts: localhost
  tasks:
  - command: echo password=test
    no_log: True
```
And output of running ```ansible-playbook -vv test.yml ```
```
TASK [command] ******************************************************************************************************************************************************************************************************************************
task path: /srv/ansible/stable-2.3/test.yml:4
changed: [localhost] => {"censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result"}
```
```
$ ansible --version                                                                                                                                               *[temp-staging-post-2.3.2]  (11:59:59)
ansible 2.4.0 (devel 9a9c1c57bf) last updated 2017/07/17 16:15:50 (GMT -700)
  configured module search path = [u'/home/badger/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /srv/ansible/vanilla/lib/ansible
  executable location = /srv/ansible/vanilla/bin/ansible
  python version = 2.7.13 (default, May 10 2017, 20:04:28) [GCC 6.3.1 20161221 (Red Hat 6.3.1-1)]
```
Note that even higher levels of verbosity can still display passwords but that is on the controller-side (reporting the modules and parameters that it is sending over the wiree) and would not be fixed by updating module_utils/basic.py code.

abadger:
Leaving in needs_info for @Jonnymcc  to report whether he has a different test case that shows this addition is still required.

PR #12119, opened 2015-08-27
Some steps towards Python 3 support

mgedmin:
This makes 'tox -e py34' pass the compileall step successfully and
actually start running the tests (plus fixes some, but not all, of the
test failures).

I want to keep this PR small and as close to obviously-correct as
possible, to avoid the fate of #10838.


v1/ansible/utils/template.py
```diff
@@ -344,7 +343,7 @@ def template_from_string(basedir, data, vars, fail_on_undefined=False):
                 var_name = only_one.group(1)
                 if var_name in vars:
                     resolved_val = vars[var_name]
-                    if isinstance(resolved_val, (bool, Number, NoneType)):
+                    if resolved_val is None or isinstance(resolved_val, (bool, Number)):
```
bcoca: remove this file from the PR, v1 is dead code and will not be updated nor released, it is there just for quick testing.

mgedmin: reverted this bit and rebased.

lib/ansible/utils/unicode.py
```diff
@@ -215,7 +215,7 @@ def to_bytes(obj, encoding='utf-8', errors='replace', nonstring=None):
         return obj
     elif nonstring == 'simplerepr':
         try:
-            simple = binary_type(obj)
+            simple = str(obj)
         except UnicodeError:
```
abadger, line 219: This one is wrong as using str() would return the text type on python3 and this function should only return byte type on both python versions.

abadger:
One comment that needs changing but I think everything else is good.

abadger:
@mgedmin Oh -- If the compileall test passes with these changes you can go ahead and enable them as well in your PR.

lib/ansible/utils/unicode.py
mgedmin, line 219: There's code below this bit (lines 229-230) that converts text to bytes, if that's the case:

``` python
         if isinstance(simple, text_type):
             simple = simple.encode(encoding, 'replace')
```

mgedmin:
@abadger: Enable them how/where?  tox.ini runs compileall on py34 before running nosetests (and did that before my changes).  Is there another place where compileall should be enabled?

mgedmin:
I see merge conflicts.  I'll rebase and fix them.

abadger:
Nice work!  if we can enable the compileall tests in travis next then we hopefully won't keep regressing how much of our code compiles with python3.

# Merged

Hi!

issue is resolved.

Thank you!

PR #12122, opened 2015-08-27
Adding support for Archlinux and Slackware in fallback

maty0609:
In some cases Archlinux and Slackware is not detected by
platform.dist(). This should solve the issue.


jimi-c:
Merged, thanks!

PR #12130, opened 2015-08-27
New version of VaultAES256: no padding, base64 output (plus some cleanups)

amenonsen:
These don't alter any functionality, but they set the stage for comprehensive improvements in the AES256 cipher while maintaining backwards compatibility more easily than was possible before. These simplifications will eventually also make it simpler to complete the `VaultFile` prototype and enable transparent vault usage from playbooks.

### DIFF
--- a/lib/ansible/parsing/vault/__init__.py
+++ b/lib/ansible/parsing/vault/__init__.py
@@ -95,6 +95,7 @@ CRYPTO_UPGRADE = "ansible-vault requires a newer version of pycrypto than the on
 b_HEADER = b'$ANSIBLE_VAULT'
 CIPHER_WHITELIST = frozenset((u'AES', u'AES256'))
 CIPHER_WRITE_WHITELIST=frozenset((u'AES256',))
+DEFAULT_WRITE_CIPHER=u'AES256'
 
 
 def check_prereqs():
@@ -120,6 +121,9 @@ class VaultLib:
         return False
 
     def _cipher(self, cipher_name):
+        """
+        Takes a cipher name and returns a Vault cipher class.
+        """
 
         cipher_class_name = u'Vault{0}'.format(cipher_name)
         if cipher_class_name in globals():
@@ -129,14 +133,69 @@ class VaultLib:
 
         return Cipher
 
+    def _add_vault_header(self, cipher_name, cipher_version, b_ciphertext):
+        """
+        Takes a cipher name and version and ciphertext (formatted output from a
+        cipher class) and returns a byte string containing the vault header and
+        the unmodified ciphertext.
+
+        :arg cipher_name: cipher name as a unicode string
+        :arg version: cipher version as a unicode string
+        :arg b_data: the formatted ciphertext as a byte string
+        :returns: a byte string that can be dumped into a file.
+        """
+
+        b_vault_version = b'1.2'
+        b_cipher_name = to_bytes(cipher_name, errors='strict', encoding='utf-8')
+        b_cipher_version = to_bytes(cipher_version, errors='strict', encoding='utf-8')
+        b_data = b'%s;%s;%s;%s\n%s' % \
+            (b_HEADER, b_vault_version, b_cipher_name, b_cipher_version, b_ciphertext)
+
+        return b_data
+
+    def _parse_vault_header(self, data):
+        """
+        Takes the contents of a vault-encrypted file (i.e. the output of
+        _add_vault_header) and returns the vault metadata and ciphertext.
+
+        :arg data: vault file contents. Since vault encrypted data is an ascii
+            text format this can be either a byte str or unicode string.
+        :returns: vault version as unicode, cipher name as unicode, cipher
+            version as unicode, and ciphertext as a byte str. If any of the
+            above is None, the header is not valid.
+        """
+
+        b_data = to_bytes(data, errors='strict', encoding='utf-8')
+
+        lines = b_data.split(b'\n', 1)
+        header = lines[0].strip().split(b';')
+
+        vault_version = None
+        cipher_name = None
+        cipher_version = None
+        b_data = None
+
+        if len(header) >= 3 and header[0] == b_HEADER:
+            vault_version = to_unicode(header[1].strip())
+            cipher_name = to_unicode(header[2].strip())
+            if vault_version == u'1.2' and len(header) == 4:
+                cipher_version = to_unicode(header[3].strip())
+            elif len(header) == 3:
+                cipher_version = vault_version
+
+        if len(lines) > 1:
+            b_data = lines[1]
+
+        return (vault_version, cipher_name, cipher_version, b_data)
+
     def encrypt(self, data, cipher_name=None):
-        """Vault encrypt a piece of data.
+        """
+        Takes plaintext as a unicode string, encrypts it using the specified (or
+        default) cipher, and returns the vault header and ciphertext as a byte
+        string suitable to be written as-is to a vault file.
 
         :arg data: a utf-8 byte str or unicode string to encrypt.
-        :returns: a utf-8 encoded byte str of encrypted data.  The string
-            contains a header identifying this as vault encrypted data and
-            formatted to newline terminated lines of 80 characters.  This is
-            suitable for dumping as is to a vault file.
+        :returns: a utf-8 encoded byte str of encrypted data.
         """
         b_plaintext = to_bytes(data, errors='strict', encoding='utf-8')
 
@@ -144,28 +203,29 @@ class VaultLib:
             raise AnsibleError("input is already encrypted")
 
         if not cipher_name or cipher_name not in CIPHER_WRITE_WHITELIST:
-            cipher_name = u"AES256"
+            cipher_name = DEFAULT_WRITE_CIPHER
 
         cipher = self._cipher(cipher_name)()
 
         b_ciphertext = cipher.encrypt(b_plaintext, self.b_password)
-        b_data = self._format_output(cipher_name, cipher.version, b_ciphertext)
+        b_data = self._add_vault_header(cipher_name, cipher.version, b_ciphertext)
 
         return b_data
 
-    def decrypt(self, data):
-        """Decrypt a piece of vault encrypted data.
-
-        :arg data: a string to decrypt.  Since vault encrypted data is an
-            ascii text format this can be either a byte str or unicode string.
-        :returns: a byte string containing the decrypted data
+    def decrypt_with_metadata(self, data):
         """
-        b_data = to_bytes(data, errors='strict', encoding='utf-8')
+        Takes the contents of a vault file (i.e. header and ciphertext) and
+        returns the cipher name and version along with the plaintext.
 
+        :arg data: vault file contents. Since vault encrypted data is an ascii
+            text format this can be either a byte str or unicode string.
+        :returns: cipher name and cipher version as unicode, plaintext as
+            a byte str
+        """
         if self.b_password is None:
             raise AnsibleError("A vault password must be specified to decrypt data")
 
-        (cipher_name, version, b_ciphertext) = self._parse_vault_file(b_data)
+        _, cipher_name, cipher_version, b_ciphertext = self._parse_vault_header(data)
 
         if not cipher_name:
             raise AnsibleError("input is not encrypted")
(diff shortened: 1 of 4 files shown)


amenonsen:
The ultimate point of this cleanup is explained in the discussion on the PR and issue linked above; this PR makes it possible to solve both problems in an elegant and backwards-compatible way.

graingert:
I think this should use https://cryptography.io/en/latest/hazmat/primitives/symmetric-encryption/#cryptography.hazmat.primitives.ciphers.Cipher because cryptography is a well known and trusted interface

amenonsen:
@graingert and what we use from PyCrypto isn't well-known and trusted? (It's a real question, I honestly don't know.) I've heard people speak about using `cryptography` whenever available, but I'm not familiar with the tradeoffs/maintenance status/etc.

Unless there's some special reason to avoid PyCrypto, I don't think switching from one AES-CTR+HMAC-SHA-256 implementation to another is particularly compelling. I see that `cryptography` implements Fernet (https://github.com/fernet/spec/blob/master/Spec.md), though. I haven't reviewed the implementation (in particular, I don't know how it generates its keys), and I really prefer AES-CTR to AES-CBC, but using an existing high-level authenticated encryption implementation does have certain attractions.

graingert:
I think what's compelling is how simple the code becomes:

```
>>> import os
>>> from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
>>> from cryptography.hazmat.backends import default_backend
>>> backend = default_backend()
>>> key = os.urandom(32)
>>> iv = os.urandom(16)
>>> cipher = Cipher(algorithms.AES(key), modes.CTR(iv), backend=backend)
>>> encryptor = cipher.encryptor()
>>> ct = encryptor.update(b"a secret message") + encryptor.finalize()
>>> decryptor = cipher.decryptor()
>>> decryptor.update(ct) + decryptor.finalize()
```

Edit: Ah this doesn't do HMAC.

amenonsen:
Yeah, it doesn't do HMAC, and it doesn't use PBKDF2 for key generation. Besides, the whole `encryptor()+encryptor.update()+encryptor.finalize()` thing is actually _more_ code than we have now.

amenonsen:
I've changed the title of this PR (from "Vault cleanups, phase 2") and added the commit that introduces a new version of the `VaultAES256` cipher that doesn't use padding and Base64-encodes output. This should make it easier to understand the point of the cleanups.

test/units/parsing/vault/test_vault_editor.py
```diff
@@ -207,7 +208,7 @@ def test_rekey_migration(self):
 
         os.unlink(v10_file.name)
 
-        assert vl.cipher_name == "AES256", "wrong cipher name set after rekey: %s" % vl.cipher_name
+        assert re.search(';AES256',fdata), "wrong cipher name set after rekey"
```
mgedmin: You could do the same `';AES256' in fdata` check here as in test_vault_editor.py, and avoid the `re` module.

lib/ansible/parsing/vault/__init__.py
```diff
@@ -528,6 +533,19 @@ def create_key(self, password, salt, keylength, ivlength):
...
+        PBKDF2 into an 32-byte AES key, a 32-byte HMAC-SHA-256 key, and
+        a 16-byte IV for AES-CTR.
+
+        FIXME: There's really no reason to use PBKDF2 to generate the initial
+        counter value. It just slows things down (specifically, it adds 10000
+        iterations of the PRF, and then throws half the 32-byte output away).
+
+        CTR mode requires that the same key and the same counter are only ever
+        used once for encryption. Since we generate the key using a random salt
+        each time, it would be fine to always start the counter at 0.
```
graingert: Or just a random 16bit IV.

test/units/parsing/vault/test_vault_editor.py
amenonsen: Sorry, I meant to fix both, but forgot to commit one hunk. Now fixed.

lib/ansible/parsing/vault/__init__.py
amenonsen: The "problem" with a random 16-byte (not bit) IV is that it'll have to be stored alongside the random 32-byte salt in the output. It wouldn't hurt anything, but since there's also no point to doing it, I don't prefer this solution.

amenonsen:
@abadger IIRC you had concerns about whether the version number should belong to the cipher or the container. Did you get a chance to think about this further? I can recast the PR accordingly.

FYI: I've squashed this down to three separate commits now for easier review (it could have been two, but I left the `.version` change separate in case it needs to be reworked).

@ldx since you recently contributed changes to make the vault faster, maybe you have some thoughts about this code too?

abadger:
VaultLib:
- Perhaps we want to keep a vault file version in addition to the individual cipher versions.  It would make it easier to change the algorithm that is used to transform the binary data into ascii in the future, for instance.
- Need a method that parses the header information of a vault file and returns the cipher name and cipher version.  This is needed by VaultEditor.edit_file()

_cipher:
- Have this function return the cipher class (ie: Cipher).  Make it the responsibility of the calling code to instantiate the class.  that way the calling code can pass additional arguments to the constructor.

_format_output:
- Since the cipher_name is more important than the cipher_version, put that first.  Also make cipher_name a byte string.  Slightly less efficient but makes the interface cleaner.  New signature is: _format_output(self, b_cipher_name, b_cipher_version, b_data)
- We can remove the check for cipher_name as that is now passed in as a required parameter instead of being set on the class.
- _format_output has lost the ability to break the data into 80 column lines.  Need to restore that.

encrypt:
- Need to take a parameter for the cipher_version as well as cipher_name.  Default None is fine and then we can let the cipher choose the version it wants to use.
- Need to add cipher version to the CIPHER_WHITELIST so that we can remove individual versions of the cipher should they prove insecure.
- With the changes to _cipher(), setting the cipher variable becomes:
  cipher = self._cipher(cipher_name)(version=cipher_version)

_parse_vault_file:
- Return all byte strings.  Make it the responsibility of the calling code to transform any of the strings to unicode type if necessary.  Slightly less efficient but it makes the interface much cleaner.
- Prepend the variables in this function that are holding byte strings with b_.

decrypt: 
- Rename version to b_version so we know it holds a byte string.
- With the change to _parse_vault_file() to make callers responsible for transforming to unicode, we should be setting b_cipher_name from the return value of _parse_vault_file().  Then we'll set cipher_name = to_unicode(b_cipher_name)
- Need to add cipher version information to CIPHER_WHITELIST and check for that here.
- Might not want to use "encryption cipher" in the error message as we're decrypting here and that might confuse users.  How about u"Cipher used to encrypt this vault file is unrecognized: {0}".format(cipher_name)
- With the changes to _cipher(), setting the cipher variable becomes:
  cipher = self._cipher(cipher_name)(version=b_version)

VaultEditor:

_edit_file_helper:
- Needs to send the cipher_name and cipher_version when it calls self.vault.encrypt()
- cipher_name and cipher_version need to be passed to _edit_file_helper() so that it knows what to pass along to self.vault.encrypt()

encrypt_file:
- Needs to take an cipher_name and cipher_version.  Needs to pass those along to self.vault.encrypt()

create_file:
- Needs to take a cipher_name and cipher_version.  Needs to pass those along to _edit_file_helper()

edit_file:
- No longer a self.vault.cipher_name attribute that it can use.  Need to call a new VaultLb function to retrieve the cipher_name and cipher_version
- Needs to compare both cipher_name and cipher_version to the CIPHER_WRITE_WHITELIST.

rekey_file:
- Need to take a cipher_name and cipher_version as parameters and pass those on to new_vault.encrypt()

VaultAES:

**init**:
- Needs to take version as a parameter.  Needs to compare whether version is equal to b'1.1' and raise an error if it does not match.

decrypt:
- Should not take a version parameter since we're setting that in **init**()

VaultAES256:

**init**:
- Need to take version as a parameter.  Needs to compare whether version is allowed (b'1.1' or b'1.2')  raise an error if no match

encrypt:
- Need to be able to write 1.1 compatible encrypted data in case the user needs to maintain compatibility with older /usr/bin/ansible on other machines.
- Use base64.b64encode/b64decode instead of codecs.encode/decode('base64').  Upstream python has talked about not having encode/decode do any bytes<=>bytes or text<=>text transformations in the future so this is future-proofing.
  import base64
  base64.b64encode/decode

Other files:
- Probably need to add a command line arg to ansible-vault to allow selecting different ciphers+cipher_version

Other thoughts:
- Since I'll be out, if you think that we don't need to be able to write VaultAES256 1.1 compatible files, see if you can talk jimi-c into that viewpoint.  I'll go along with him if he thinks it's okay to only support a single format for writing.
- In the ciphers, version can either go into the **init**() and become an attribute (self.version) or it can be a parameter of both decrypt and encrypt.  I think it makes more sense in **init**() but you could try to justify the other direction instead.  If we go the other way, then the version parameter needs to be checked to be sure that it is valid wherever it is given as a parameter.

srgvg:
@abadger is that a roadmap in your comment or are you just happy to see this pull request?

amenonsen:
@abadger @jimi-c didn't need much convincing:

```
 18:00  <jimi|ansible> crab: yes that's how we did it last time
 18:00  <jimi|ansible> vault reads all but writes latest
```

I'll work on the changes from that perspective.

amenonsen:
Hi @abadger.

Thanks for taking the time to comment on this PR. I went through your
list, and here are some thoughts.

The gist is that I don't disagree with anything you're proposing, but I
think they're generalisations that should be implemented when there's an
actual use-case for them (e.g. when GPG support is added). I don't think
it makes sense to include many of them in this PR.

Also note that the comments below are based on the perspective that we
stay with the "read any, write latest" strategy.

> VaultLib:
> - Perhaps we want to keep a vault file version in addition to the
>   individual cipher versions.

Perhaps, but it can be added later (in a backwards-compatible way).

> - Need a method that parses the header information of a vault file and
>   returns the cipher name and cipher version.  This is needed by
>   VaultEditor.edit_file()

I don't understand why this is needed (or perhaps how it would be used),
but _parse_vault_file should be easily adaptable.

> _cipher:
> - Have this function return the cipher class (ie: Cipher).

OK.

> _format_output:
> - … New signature is: _format_output(self, b_cipher_name, b_cipher_version, b_data)
> - We can remove the check for cipher_name …

OK.

> - _format_output has lost the ability to break the data into 80
>   column lines.  Need to restore that.

I don't think this is needed, at least until there's an actual cipher
that needs this. Base64 encoding takes care of this already, and GPG
won't need it either.

On the other hand, I observe that base64.b64encode() doesn't wrap the
output, and the encodestring() method (which does) is described as being
a "legacy" interface. So I'll see what makes the most sense (though to
be honest, I don't see the point of wrapping at 80 chars).

> encrypt:
> - Need to take a parameter for the cipher_version as well as
>   cipher_name.

Not needed per "write latest".

> - Need to add cipher version to the CIPHER_WHITELIST so that we can
>   remove individual versions of the cipher should they prove insecure.

Can be done later.

> _parse_vault_file:
> - Return all byte strings. …
> - Prepend the variables in this function that are holding byte
>   strings with b_.

OK.

> decrypt: 
> - Rename version to b_version so we know it holds a byte string.

OK.

> - With the change to _parse_vault_file() to make callers responsible
>   for transforming to unicode, we should be setting b_cipher_name from
>   the return value of _parse_vault_file().  Then we'll set cipher_name
>   = to_unicode(b_cipher_name)

OK.

> - Need to add cipher version information to CIPHER_WHITELIST and
>   check for that here.

Again, can be done later when there's an actual need for it.

> - Might not want to use "encryption cipher" in the error message as
>   we're decrypting here and that might confuse users.  How about
>   u"Cipher used to encrypt this vault file is unrecognized:
>   {0}".format(cipher_name)

OK.

> VaultEditor:

As far as I can tell, all of these changes can be done later.

> VaultAES: […move version from decrypt to init…]
> VaultAES256: […ditto…]

OK.

> encrypt:
> - Need to be able to write 1.1 compatible encrypted data in case the
>   user needs to maintain compatibility with older /usr/bin/ansible on
>   other machines.

Not needed per "write latest".

> - Use base64.b64encode/b64decode instead of codecs.encode/decode('base64').

OK.

> Other files:
> - Probably need to add a command line arg to ansible-vault to allow
>   selecting different ciphers+cipher_version

Can be done later, e.g. when GPG support is introduced.

I would be happy to resubmit with the changes marked OK above; let me
know if you'd be willing to merge that. As for the remaining changes, I
don't think they belong with this PR (and I think it's premature to do
them before there's a concrete use-case), and I don't think it's fair
to make them dependencies for the change that I proposed.

amenonsen:
I've pushed a new version with most of the changes discussed above, except for pushing unicode conversion to the callers of `_format_output` and `_parse_vault_file` (specifically, I've left the cipher name and version as unicode strings; pushing it to the callers made for more ugly boilerplate code). IMHO that's about as far as it's reasonable to clean things up in the context of this PR.

amenonsen:
@abadger did you get a chance to look through these changes?

amenonsen:
@abadger I've pushed the changes you requested, to add a separate header version field (set to 1.2 here) in addition to the cipher version (which works as before). The test failures are due to unrelated breakage in devel caused by commit 79e364d9, but I did update the tests for the changes in this PR (and those pass):

```
kea:~/extern/ansible $ PYTHONPATH=./lib nosetests -d -w test/units/parsing/vault -q
Ran 14 tests in 0.640s

OK
```

abadger:
Marking v2 as there's a few things in here that are fixes we'd really want for v2 as well.

graingert:
Closed? WFT?

amenonsen:
Since a number of people have now written to ask why I closed the PR, the main reasons are that (a) it won't be in v2: "[we] Decided to take it off the v2 milestone because we don't have anyone in-house to review it for security problems and it's late to be adding it for v2" (per @abadger on IRC), (b) this code has significant security and compatibility considerations, and if it's not going to be in v2, I don't want people to use it (e.g. because of possible divergence with core cipher names/versions, if something else gets merged in future); I know of people who want to, at least, and we know that people have used other unmerged features from PRs in the past (e.g. with the vault+gpg changes, to name just one), (c) it's blocking @mgedmin's python 3 compatibility work because he's been holding off on touching the vault code until this PR was merged.

On a personal note, I feel it reflects poorly on me as a contributor to have long-pending PRs (and bitrotting security code in particular is the worst). I know that not everyone will think this is a sensible feeling, and it's certainly not something I mean to make a big deal about (the reasons above, to me, are sufficient to motivate the decision to close), but it's how I feel anyway.

graingert:
That's fair enough. Do you think there well be a 2.1 milestone to target?

jimi-c:
@graingert that's pretty much what the `next` milestone is. Once 2.0 is out, we'll create a 2.1 milestone and move everything from next to 2.1.

mgedmin:
> (c) it's blocking @mgedmin's python 3 compatibility work because he's been holding off on touching the vault code until this PR was merged.

FWIW that was mostly an excuse to give people an incentive to merge this quicker.

h0nIg:
could you please review this (and reopen, since 2.1 is RC and its still not included), because http://toroid.org/ansible-vault-strange-crypto points out some improvements regarding vault files and speed?

+1

amenonsen:
This is superseded by #15060, which I'll merge when I get some time to address review comments.

Issue #12016, opened 2015-08-19
Module resolution broken in 2.0


nitzmahone:
The default library path (./library) as well as command-line overrides (-M / --module-path) are broken in 2.0- modules in either of these locations will not be found as they are in 1.x. The only functioning mechanism to load custom modules in 2.0 appears to be setting the library directive in ansible.cfg.

@jimi-c - as requested

bcoca:
so i just tested in current devel and having a 'library' dir adjacent to play seems to work fine:

created snmp_facts module:

```
#!/usr/bin/python


def main():
    module = AnsibleModule(
        argument_spec={
            'test':     {'required': False,},
        },
        supports_check_mode=True
    )

    module.exit_json(msg="iran", changed=False)

# import module snippets
from ansible.module_utils.basic import *
main()

```

and play.yml:

```
- hosts: localhost
  gather_facts: false
  tasks:
    - snmp_facts:
```

i put snmp_facts into 'library':

```
ls -lR
.:
total 16
drwxr-xr-x 2 bcoca bcoca 4096 Aug 24 19:11 library
-rw-r--r-- 1 bcoca bcoca  112 Aug 24 19:11 play.yml

./library:
total 4
-rw-r--r-- 1 bcoca bcoca 304 Aug 24 19:10 snmp_facts

```

and ran:

```
ansible-playbook play.yml -v
Using /etc/ansible/ansible.cfg as config file

PLAY ***************************************************************************

TASK [snmp_facts] **************************************************************
ok: [localhost] => {"changed": false, "msg": "iran"}

PLAY RECAP *********************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=0   
```

nitzmahone:
Ah, looks like it's only broken for ansible, not ansible-playbook.

bcoca:
-M is not implemented .... it should work for ansible and ansible-{pull,doc,playbook}

PR #12137, opened 2015-08-28
restore ansible's ./library lookup, add to ansible-doc

nitzmahone:
verified that lookup order is the same as 1.9 (modules in ./library are preferred to built-in)

follow-on fix for issue #12016 

@bcoca


bcoca:
No, we don't wan't cwd as a source for these, that is what -M/--module-path is for.

nitzmahone:
OK, so long as we're breaking ansible's 1.9 behavior on purpose... :)

abadger:
We should probably document what's changed, though.

nitzmahone:
abadger: :+1:

PR #12150, opened 2015-08-28
Fix setting dns_name in cobbler inventory

opoplawski:
The 'hostname' field in cobbler is often empty for non static-IP hosts. Fall back to the previous method of searching the interfaces for a management interface to determine the DNS name. Also look for non-static interfaces.

Also, host['hostname'] will never be None, check for empty string instead.


jimi-c:
Merged, thanks!

PR #12164, opened 2015-08-31
Change hacking scripts to python3 syntax

qwazerty:
Improve compatibility with python3.

### DIFF
--- a/hacking/get_library.py
+++ b/hacking/get_library.py
@@ -22,7 +22,7 @@ import ansible.constants as C
 import sys
 
 def main():
-    print C.DEFAULT_MODULE_PATH
+    print(C.DEFAULT_MODULE_PATH)
     return 0
 
 if __name__ == '__main__':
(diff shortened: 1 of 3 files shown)


bcoca:
You are updating many files you should not in this request, including documentation and module shared code that MUST be 2.4 compatible and your changes break.

qwazerty:
Well that was indeed too much changes at once. Though the `except ... as ...` syntax would work on python2.4.

I've reduce the scope of the PR.

abadger:
`except [...] as [...]` does not work on python2.4.  That was introduced in python-2.6.

@bcoca: I've reviewed this.  as long as hacking/module_formatter.py can move compatibility up to python-2.6, this is good to merge.

qwazerty:
If this is not fine for `hacking/module_formatter.py` I'll remove it from this PR.

bcoca:
Well, the first thing is that I want to remove all those print statements and port it to use display class, so this change will be irrelevant then. Outside of contrib/ everything should be using the display class.

But since I don't know when I'll get around to that ... I don't oppose merging this in it's current state.

abadger:
Thanks @bcoca and @qwazerty !

# Merged

Hi!

issue is resolved.

Thank you!

PR #12187, opened 2015-09-01
Adds utf-8 encoding to f5 module_utils

caphrim007:
The contributor's name on line 10 (originally line 7) includes a character
that the default Python encoding (ASCII) raises an error on when interpreting
the file.

Specifying the utf-8 encoding, as is done in other modules, resolves
the error.

The error being raised is

SyntaxError: Non-ASCII character '\xc3' in file /.../lib/ansible/module_utils/f5.py
on line 7, but no encoding declared; see http://www.python.org/peps/pep-0263.html
for details

### DIFF
--- a/lib/ansible/module_utils/f5.py
+++ b/lib/ansible/module_utils/f5.py
@@ -1,3 +1,6 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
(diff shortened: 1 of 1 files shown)


jimi-c:
Merged, thanks!

PR #12188, opened 2015-09-01
Show module name and args with -vvv

amenonsen:
This information was earlier shown only with ANSIBLE_DEBUG, but it's
extremely useful in a user context, especially with module invocations
with deeply nested args like the ec2_vpc/ec2 modules.


jimi-c:
Merged, thanks!

PR #12192, opened 2015-09-01
fix arg leakage to local output on no_log

nitzmahone:
re-fixes #11946 for 2.0 after #12188 re-introduced the issue. :)

### DIFF
--- a/lib/ansible/plugins/action/__init__.py
+++ b/lib/ansible/plugins/action/__init__.py
@@ -346,8 +346,11 @@ class ActionBase:
         # set no log in the module arguments, if required
         if self._play_context.no_log:
             module_args['_ansible_no_log'] = True
+            module_display_args = "(no_log enabled, args censored)"
+        else:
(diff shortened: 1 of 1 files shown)


nitzmahone:
started writing some tests to prevent further regressions on this, and there are others that need fixing as well. I'll hit them all and include the tests in another PR

bcoca:
actually, the line should be completely removed as it ignores no_log in the module spec

nitzmahone:
There were a few issues with no_log leaking args on stdout in 1.9, and a bunch more that came in with 2.0 (not just this one)- I have a handful of small code fixes and a new integration test forthcoming, so please hold off on nuking the change until you see that. 

I'd love to come up with a way for no_log in module specs to be respected as well, but short of just adding a blanket # NO_LOG_ARGS kind of thing, I just don't see a way to do it without executing the module code locally (think non-Python modules, target deps, blech).

PR #12195, opened 2015-09-02
Make lock_host_keys a real noop, instead of half-a-noop

amenonsen:
Right now, we don't do any locking, but we still scan known_hosts files
twice per connection. That's completely unnecessary, and the proposed
solutions to the locking problem wouldn't need known_hosts scanning
anyway, so this code can go away.


jimi-c:
Merged, thanks!

PR #12202, opened 2015-09-02
Use rich comparisons for Attribute and FieldAttribute

mgedmin:
Because **cmp** is not supported on Python 3.

Fixes

  TypeError: unorderable types: FieldAttribute() < FieldAttribute()

on Python 3.


jimi-c:
Merged, thanks!

PR #12214, opened 2015-09-02
fix no_log leakages to local output, add tests

nitzmahone:
Re-fixes #11946 for 2.0 after #12188 re-introduced the issue. Also fixes a number of other leakages introduced with 2.0, and adds integration tests to check most cases fixed herein.

### DIFF
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -252,7 +252,11 @@ class TaskExecutor:
         # variable not being present which would otherwise cause validation to fail
         if not self._task.evaluate_conditional(templar, variables):
             debug("when evaulation failed, skipping this task")
-            return dict(changed=False, skipped=True, skip_reason='Conditional check failed')
+            skip_result = dict(changed=False, skipped=True, skip_reason='Conditional check failed')
+            # propagate no_log into the result if necessary
+            if self._play_context.no_log:
+                skip_result['_ansible_no_log'] = True
+            return skip_result
 
         # Now we do final validation on the task, which sets all fields to their final values.
         # In the case of debug tasks, we save any 'var' params and restore them after validating
@@ -315,6 +319,10 @@ class TaskExecutor:
             debug("running the handler")
             result = self._handler.run(task_vars=variables)
             debug("handler run complete")
+
+            # propagate no_log into the result if necessary
+            if self._play_context.no_log:
+                result["_ansible_no_log"] = True
 
             if self._task.async > 0:
                 # the async_wrapper module returns dumped JSON via its stdout
@@ -422,6 +430,9 @@ class TaskExecutor:
         if int(async_result.get('finished', 0)) != 1:
             return dict(failed=True, msg="async task did not complete within the requested time")
         else:
+            # propagate no_log into the result if necessary
+            if self._play_context.no_log:
+                async_result['_ansible_no_log'] = True
             return async_result
 
     def _get_connection(self, variables):
--- a/lib/ansible/plugins/action/__init__.py
+++ b/lib/ansible/plugins/action/__init__.py
@@ -346,8 +346,11 @@ class ActionBase:
         # set no log in the module arguments, if required
         if self._play_context.no_log:
             module_args['_ansible_no_log'] = True
+            module_display_args = "(no_log enabled, args censored)"
+        else:
+            module_display_args = module_args
 
-        self._display.vvv("Module: %s, args: %s" % (module_name, module_args))
+        self._display.vvv("Module: %s, args: %s" % (module_name, module_display_args))
 
         (module_style, shebang, module_data) = self._configure_module(module_name=module_name, module_args=module_args, task_vars=task_vars)
         if not shebang:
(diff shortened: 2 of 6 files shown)


test/units/executor/test_task_executor.py
```diff
@@ -323,5 +323,5 @@ def _get(*args, **kwargs):
         with patch.object(action_loader, 'get', _get):
             mock_templar = MagicMock()
             res = te._poll_async_result(result=dict(ansible_job_id=1), templar=mock_templar)
-            self.assertEqual(res, dict(finished=1))
+            self.assertTrue(('finished', 1) in res.items())
```
nitzmahone, line 326: the mock task has no_log set to true for some reason, so the output dict with the propagated no_log value no longer equals the static one. changed to just ensure that finished is in the results instead.

lib/ansible/plugins/callback/default.py
```diff
@@ -115,6 +115,9 @@ def v2_on_file_diff(self, result):
         if 'diff' in result._result and result._result['diff']:
             self._display.display(self._get_diff(result._result['diff']))
 
+    def _get_censored_item(self, result):
```
bcoca, line 118: default is not the correct place for this as other custom callbacks can ignore it then. 
items should be scrubbed before they even get to callbacks, or if not possible, _get_censored_item should be called automatically from the base class (which still leaves v1 callbacks able to ignore it).

lib/ansible/plugins/action/__init__.py
```diff
@@ -346,8 +346,11 @@ def _execute_module(self, module_name=None, module_args=None, tmp=None, task_var
         # set no log in the module arguments, if required
         if self._play_context.no_log:
             module_args['_ansible_no_log'] = True
+            module_display_args = "(no_log enabled, args censored)"
+        else:
+            module_display_args = module_args
 
-        self._display.vvv("Module: %s, args: %s" % (module_name, module_args))
+        self._display.vvv("Module: %s, args: %s" % (module_name, module_display_args))
```
bcoca, line 353: i removed this as even if you respect the play's no_log there is no current way to respect the module spec's no_log

nitzmahone, line 353: That's a larger problem for a lot of reasons- one of the things I was going to suggest was implementing some declarative module metadata in terms of the doc structure (or something like it) so that the runner-side can do arg validation and modify its behavior to deal with things like this. In the face of non-Python modules, it's pretty much impossible without having some sort of non-executable arg metadata.

lib/ansible/plugins/callback/default.py
nitzmahone, line 118: I considered that- depends on if we want other callback plugins to have the ability to censor at will, or keep that as a choice made by the infrastructure... IIRC, we're not always deepcopying the inputs to the callback plugins, so censoring the data in-place may have unintended consequences. If we're sure that all the different paths are deep-copied, though, I can probably move it elsewhere.

nitzmahone, line 118: Also would mean that it's impossible for callback plugins to be guaranteed access to the args/results at all- probably fine for logging plugins, but right or wrong, people _are_ doing other sinister things with callback plugins that could be broken by in-place censoring.

bcoca, line 118: I'll check, but i believe callbacks already get a copy, just need to check that it is not linked to registered vars, we should be able to censor at that point.

lib/ansible/plugins/action/__init__.py
bcoca, line 353: well, the problem only exists if we try to report on the arguments BEFORE executing the module, which we really should not need, that entry was original a debug entry that was added because we rewrote the whole engine, we should not need this anymore.

bcoca:
cc @ansible/core

lib/ansible/executor/task_executor.py
```diff
@@ -252,7 +252,11 @@ def _execute(self, variables=None):
         # variable not being present which would otherwise cause validation to fail
         if not self._task.evaluate_conditional(templar, variables):
             debug("when evaulation failed, skipping this task")
-            return dict(changed=False, skipped=True, skip_reason='Conditional check failed')
+            skip_result = dict(changed=False, skipped=True, skip_reason='Conditional check failed')
+            # propagate no_log into the result if necessary
+            if self._play_context.no_log:
+                skip_result['_ansible_no_log'] = True
+            return skip_result
```
bcoca, line 259: this one is not really needed as there is no user data passed back in skip_result

lib/ansible/executor/task_executor.py
```diff
@@ -422,6 +430,9 @@ def _poll_async_result(self, result, templar):
         if int(async_result.get('finished', 0)) != 1:
             return dict(failed=True, msg="async task did not complete within the requested time")
         else:
+            # propagate no_log into the result if necessary
+            if self._play_context.no_log:
+                async_result['_ansible_no_log'] = True
             return async_result
```
bcoca, line 436: instead of these 2 this patch should take care of it:

```
index 0e34938..3bee47a 100644
@@ -419,6 +419,9 @@ class TaskExecutor:
         if self._task.notify is not None:
             result['_ansible_notify'] = self._task.notify

+        # preserve no_log setting
+        result["_ansible_no_log"] = self._play_context.no_log
+
         # and return
         debug("attempt loop complete, returning result")
         return result
```

PR #12217, opened 2015-09-02
Delete all compiled python files when running hacking/env-setup

thyming:
Switching between branches, I got errors about modules missing error classes - I think this is because the errors module in lib/ansible was moved to its own directory. This clears our any compiled artifacts that shouldn't be there.


sivel:
This does not work for me in bash (`GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin14)`).

```
rm: **/*.pyc: No such file or directory
```

I already use the following via an alias that works well:

```
find . -type f -name "*.pyc" -delete
```

thyming:
Ah yeah, that's better.

PR #12221, opened 2015-09-02
make the way we build search paths easier to understand

isotopp:
##### SUMMARY
get_bin_path build a search path from three different sources: an opt_dirs argument, the system PATH and a fixed list of sbin directories.

The original implementation does not apply all sanity checks to all sources: Some are not checked for "None", for duplicates or for existence.

This implementation builds a total list of directories, in the same order as before and then applies all three checks to every directory: Must not be None, must be unique and must exist in the filesystem.

With the resulting clean path list the same logic as before is applied.

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
lib/ansible/module_utils/basic.py

##### ANSIBLE VERSION
2.3


isotopp:
It seems I fucked this one up by resolving a merge conflict from upstream the wrong way. You need to cherry pick https://github.com/isotopp/ansible/commit/ec568e730547c990fc79d29630cec6f31a69c51d for the intended change.

It is useful to keep this merge request alive or won't it be merged anyway?

jctanner:
Closing due to submitter timeout on rebase. If you decide to resubmit this, please rebase your branch.

PR #12226, opened 2015-09-03
Replace .iteritems() with six.iteritems()

mgedmin:
Replace .iteritems() with six.iteritems() everywhere except in module_utils
(because there's no 'six' on the remote host).  And except in
lib/ansible/galaxy/data/metadata_template.j2, because I'm not sure six is
available there.

Brings down the number of failing tests under Python 3.4 from 53 down to 34.

### DIFF
--- a/lib/ansible/plugins/cache/__init__.py
+++ b/lib/ansible/plugins/cache/__init__.py
@@ -60,7 +60,7 @@ class FactCache(MutableMapping):
 
     def copy(self):
         """ Return a primitive copy of the keys and values from the cache. """
-        return dict([(k, v) for (k, v) in self.iteritems()])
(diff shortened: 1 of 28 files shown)


lib/ansible/plugins/cache/__init__.py
```diff
@@ -60,7 +60,7 @@ def __len__(self):
 
     def copy(self):
         """ Return a primitive copy of the keys and values from the cache. """
-        return dict([(k, v) for (k, v) in self.iteritems()])
+        return dict([(k, v) for (k, v) in iteritems(self)])
```
mgedmin, line 63: Welp, I somehow missed the `from six import iteritems` here.

PR #12229, opened 2015-09-03
Refactoring: simplify and speed up RoleDefinition code

mgedmin:
Thanks to @amenonsen for suggesting that the _get_base_attributes() call 
should be hoisted out of the loop.


abadger:
Oops, sorry - I did almost the same thing you did here: 364313c01f9cb94cdabffe694ca05d6cde16a9ed   I should have asked if you were working on it.

PR #12236, opened 2015-09-03
Go back to using ~/.ansible/cp as the ControlPath

amenonsen:
This was commented out earlier because of the lack of interprocess
locking and prepare_writeable_dir in v2.

The locking was not needed: it could only protect against other siblings
of this process (since they were all locking a temporary file that was
opened in the parent), and those would be running as the same user and
with the same umask. Also, os.makedirs() tolerates intermediate paths
being created by other processes. For any other kind of error, both
locking and non-locking code paths would fail in the same way.

So all we really need to do is make sure we have write permissions.

### DIFF
--- a/lib/ansible/plugins/connections/ssh.py
+++ b/lib/ansible/plugins/connections/ssh.py
@@ -37,6 +37,7 @@ from hashlib import sha1
 from ansible import constants as C
 from ansible.errors import AnsibleError, AnsibleConnectionFailure, AnsibleFileNotFound
 from ansible.plugins.connections import ConnectionBase
+from ansible.utils.path import unfrackpath, makedirs_safe
 
 class Connection(ConnectionBase):
     ''' ssh based connections '''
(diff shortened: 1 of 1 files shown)


lib/ansible/plugins/connections/ssh.py
```diff
@@ -49,11 +50,15 @@ def __init__(self, *args, **kwargs):
...
-        #fcntl.lockf(self.runner.process_lockfile, fcntl.LOCK_UN)
+        # We set ControlPath to ~/.ansible/cp, so it must exist and be writable.
+        self._cp_dir = unfrackpath('$HOME/.ansible/cp')
+        if not os.path.exists(self._cp_dir):
+            try:
+                os.makedirs(self._cp_dir, 0700)
+            except (IOError, OSError), e:
+                raise AnsibleError("Could not make dir %s: %s" % (self._cp_dir, e))
+        if not os.access(self._cp_dir, os.W_OK):
+            raise AnsibleError("Cannot write to path %s" % self._cp_dir)
```
bcoca: makedirs_safe already does this, just use that function

jimi-c:
Merged, thanks!

Issue #11041, opened 2015-05-20
SSH connection errors are unclear


shawnferry:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.9 (devel ede94dd7f8) last updated 2015/03/29 14:04:17 (GMT -400)
  lib/ansible/modules/core: (detached HEAD de4c7f8ba6) last updated 2015/05/19 12:48:28 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD cb848fcd9e) last updated 2015/04/03 15:14:36 (GMT -400)
  v2/ansible/modules/core: (detached HEAD 34784b7a61) last updated 2015/02/26 12:26:47 (GMT -400)
  v2/ansible/modules/extras: (detached HEAD 46e316a20a) last updated 2015/04/03 15:14:36 (GMT -400)
  configured module search path = None
##### Ansible Configuration:

Nothing of note
##### Environment:

NA
##### Summary:

Hosts without name resolution result in failures with unclear ssh connection errors.
##### Steps To Reproduce:

Attempt to connect to a host in inventory which does not have name resolution. 
##### Expected Results:

Ansible with ssh errors more clearly indicates the reason for failure
###### ansible paramiko

```
fatal: [hostA] => {'msg': 'FAILED: [Errno 8] nodename nor servname provided, or not known', 'failed': True}
```
###### command line ssh

```
ssh hostA
ssh: Could not resolve hostname hostA: nodename nor servname provided, or not known
```
##### Actual Results:

Ansible with ssh errors are unclear
###### ansible ssh

```
fatal: [hostA] => SSH Error: data could not be sent to the remote host. Make sure this host can be reached over ssh
```
##### Example

If I have hundreds or thousands of hosts some of which need to be bootstrapped in a specific way.  I can't tell from the ssh errors if they need to be bootstrapped or if there is some other error which needs to be addressed separately.  In the simple case this forces me to fall back to paramiko and bootstrap them all again which is slow.

bcoca:
so in 1.9.1 iirc, we added passing more data from the ssh command line to the ssh errors, please try this with the 1.9 stable version or current devel (2.0), you seem to have an older checkout of devel.

shawnferry:
Updated I'm reporting 
     ansible 2.0.0 (devel 5658ddf02e) last updated 2015/05/31 18:48:20 (GMT -400)
[...]

PR #12243, opened 2015-09-04
Don't try to run "ssh -O stop" in Connection.close()

amenonsen:
The TaskExecutor calls connection.close() after it finishes executing
each task. Running "ssh -O stop" here would defeat ControlPersist and
force each task to create a new connection.

That this didn't actually happen was merely a happy accident: what was
in self._common_args was actually 'ControlMaster=yes', while the test
only looked for 'ControlMaster'.

### DIFF
--- a/lib/ansible/plugins/connections/ssh.py
+++ b/lib/ansible/plugins/connections/ssh.py
@@ -517,15 +517,4 @@ class Connection(ConnectionBase):
     def close(self):
         ''' not applicable since we're executing openssh binaries '''
 
-        if self._connected:
-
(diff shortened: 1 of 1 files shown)


abadger:
This looks correct to me.  @bcoca since you added this in https://github.com/ansible/ansible/issues/11041  do you want to take a look?

abadger:
Alright, since https://github.com/ansible/ansible/commit/23a22397bfdbcf9d8a4b9037fc5fb63ae8dcefc1 went in, I guess we don't need this for now.  May have to revisit when do start calling connection.close().

PR #12244, opened 2015-09-04
Python 3: use six.text_type instead of unicode

mgedmin:
Replace 'unicode' with six.text_type, everywhere but in module_utils.

### DIFF
--- a/contrib/inventory/apache-libcloud.py
+++ b/contrib/inventory/apache-libcloud.py
@@ -260,11 +260,11 @@ class LibcloudInventory(object):
             key = self.to_safe('ec2_' + key)
 
             # Handle complex types
-            if type(value) in [int, bool]:
+            if isinstance(value, (int, bool)):
                 instance_vars[key] = value
-            elif type(value) in string_types:
+            elif isinstance(value, string_types):
                 instance_vars[key] = value.strip()
-            elif type(value) == type(None):
+            elif value is None:
                 instance_vars[key] = ''
             elif key == 'ec2_region':
                 instance_vars[key] = value.name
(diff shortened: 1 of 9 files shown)


contrib/inventory/apache-libcloud.py
```diff
@@ -262,7 +262,7 @@ def get_host_info(self):
             # Handle complex types
             if type(value) in [int, bool]:
                 instance_vars[key] = value
-            elif type(value) in [str, unicode]:
+            elif type(value) in string_types:
```
abadger: Need to change this to an isinstance check.  Otherwise on python2 the conditional will be equivalent to:

```
elif str in (basetring,)
```

which will be False.

abadger:
For things inside of ansible I think it might be better to use to_unicode() rather than text_types().  We'd want to make sure that the change set the value of errors appropriately, though -- errors='strict' might be the right choice for many of the things we do inside of ansible.
I also didn't see to_unicode as one of the things changed here.  That one is a tiny bit tricky, you can take a look at the python3 port of kitchen to see what that code should look like if it were a pure python3 library. https://pypi.python.org/packages/source/k/kitchen/kitchen-1.2.1.tar.gz

contrib/inventory/apache-libcloud.py
mgedmin: Ooh, good catch!

mgedmin:
> For things inside of ansible I think it might be better to use to_unicode() rather than text_types().

Probably!

> I also didn't see to_unicode as one of the things changed here.

It was already using six.text_type.

mgedmin:
> > For things inside of ansible I think it might be better to use to_unicode() rather than text_types().
> > Probably!

(I will not be doing this any time soon. :wink:)

abadger:
Note to potential successors if I should get hit by a truck,  remember to switch six.text_type() over to ansible.utils.unicode.to_unicode()

Thanks @mgedmin!

PR #12250, opened 2015-09-04
Become clean

bcoca:
fixed a few misconceptions in the code, cleared play_context.prompt as simple test for become + password requirements.

### DIFF
--- a/lib/ansible/playbook/play_context.py
+++ b/lib/ansible/playbook/play_context.py
@@ -335,6 +335,7 @@ class PlayContext(Base):
 
         prompt      = None
         success_key = None
+        self.prompt = None
 
         if executable is None:
             executable = C.DEFAULT_EXECUTABLE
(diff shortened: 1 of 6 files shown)


lib/ansible/playbook/play_context.py
```diff
@@ -415,7 +416,8 @@ def detect_su_prompt(data):
             else:
                 raise AnsibleError("Privilege escalation method not found: %s" % self.become_method)
 
-            self.prompt      = prompt
+            if self.become_pass:
+                self.prompt = prompt
             self.success_key = success_key
```
abadger, line 422: You need to always set self.prompt here.  Since the code sets prompt  to None at the top of the file, I think self.prompt = prompt would work.  Otherwise, just an `else: self.prompt = None` would work.

test/units/playbook/test_play_context.py
```diff
@@ -134,7 +134,10 @@ def test_play_context_make_become_cmd(self):
 
         play_context.become_method = 'sudo'
         cmd = play_context.make_become_cmd(cmd=default_cmd, executable="/bin/bash")
-        self.assertEqual(cmd, """%s -c '%s %s -S -p "%s" -u %s %s -c '"'"'echo %s; %s'"'"''""" % (default_exe, sudo_exe, sudo_flags, play_context.prompt, play_context.become_user, default_exe, play_context.success_key, default_cmd))
+        if play_context.prompt:
+            self.assertEqual(cmd, """%s -c '%s %s -p "%s" -S -u %s %s -c '"'"'echo %s; %s'"'"''""" % (default_exe, sudo_exe, sudo_flags, play_context.prompt, play_context.become_user, default_exe, play_context.success_key, default_cmd))
+        else:
+            self.assertEqual(cmd, """%s -c '%s %s -n -S -u %s %s -c '"'"'echo %s; %s'"'"''""" % (default_exe, sudo_exe, sudo_flags, play_context.become_user, default_exe, play_context.success_key, default_cmd))
```
abadger: Since these are for tests we probably want to set play_context.prompt and check the output.  Then unset play_context.prompt and check the output.

bcoca: not in every case, resolved by always initializing prompt also

abadger:
+1

Issue #12166, opened 2015-08-31
Error while building the RPM


skynet:
```
RPM build errors:
    File not found by glob: /usr/local/src/ansible/rpm-build/BUILDROOT/ansible-2.0.0-0.git201508310021.6b73cac.devel.amzn1.x86_64/usr/lib/python2.7/dist-packages/ansible*
make: *** [rpm] Error 1
```

Correct location is:

```
/usr/local/src/ansible/rpm-build/BUILDROOT/ansible-2.0.0-0.git201508310021.6b73cac.devel.amzn1.x86_64/usr/lib/python2.7/site-packages/ansible>
```

jimi-c:
@skynet how are you building the RPM?

skynet:
```
$ git clone git://github.com/ansible/ansible.git --recursive
$ cd ./ansible
$ make rpm
```

skynet:
I was able to build on a different system. This error is probably related to python version and libraries.

skynet:
Worked here:

```
$python --version
Python 2.6.6
```

Failed here:

```
$ python --version
Python 2.7.9

```

toshywoshy:
Working with Python 2.7.5
https://rpmrepo1.vantosh.com/yum/el7/x86_64/srpms/ansible-2.0.0-1.el7.src.rpm

jimi-c:
Interesting, I'm on 2.7.10 and it works for me as well.

skynet:
We'll wait for others to report similar issues. Going back to stable, since 2.0 fails in other places too. Thanks.

abadger:
@skynet  It looks like something isn't configured correctly in the python or rpm you're using on that system.  The file glob in the ansible spec file that's throwing that error looks like this:

```
%{python_sitelib}/ansible*
```

The `%{python_sitelib}` is an rpm macro.  rpm itself usually defines it in /usr/lib/rpm/macros:

```
$ grep -ri python_sitelib /usr/lib/rpm
/usr/lib/rpm/macros:%python_sitelib %(%{__python} -c "from distutils.sysconfig import get_python_lib; import sys; sys.stdout.write(get_python_lib())")
$ rpm -qf /usr/lib/rpm/macros
rpm-4.12.0.1-7.fc21.x86_64
```

you can see what rpm would expand the macro to using:

```
$ rpm --eval '%{python_sitelib}'
/usr/lib/python2.7/site-packages
```

From the definition, the macro is calling python to define itself.  If you run:

```
[...]

PR #12265, opened 2015-09-06
Potential fix for amazon linux's rpm doing something strange with the python_sitelib macro #12166

abadger:


abadger:
@skynet go ahead and give this a try.  be sure to try the installed package as well -- Since I don't have Amazon Linux I don't know for sure that we need to be moving from site-packages to dist-packages rather than the other way around.  I'm just going by what I'm seeing other projects that have run into the same problem are doing.

PR #12267, opened 2015-09-07
Add the new module replacer <<INCLUDE_ANSIBLE_MODULE_JSON_ARGS>>.

### LINKED ISSUE #12266: Add "<<INCLUDE_ANSIBLE_MODULE_JSON_ARGS>>" replacer for non-Python modules.
In v2, the module-arguments file are no more passed to any modules, even if the style of the modules are `non_native_want_json` or `old`.
So, we have to embed `<<INCLUDE_ANSIBLE_MODULE_WINDOWS_ARGS>>` replacer in non-Python (nor PS1) modules to pass module-arguments.
It seems the name of the replacer is some kind of confusing, and adding the new replacer `<<INCLUDE_ANSIBLE_MODULE_JSON_ARGS>>`, the alias of `<<INCLUDE_ANSIBLE_MODULE_WINDOWS_ARGS>>`, would be helpful.

h-hirokawa:
Fixes #12266

### DIFF
--- a/lib/ansible/executor/module_common.py
+++ b/lib/ansible/executor/module_common.py
@@ -33,12 +33,13 @@ from ansible.errors import AnsibleError
 from ansible.parsing.utils.jsonify import jsonify
 from ansible.utils.unicode import to_bytes
 
-REPLACER         = "#<<INCLUDE_ANSIBLE_MODULE_COMMON>>"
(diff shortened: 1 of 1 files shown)


jimi-c:
Merged, thanks!

PR #12268, opened 2015-09-07
Fix rsync connections to IPv6 addresses

jaingaurav:
Similar to https://github.com/ansible/ansible/pull/11816 we can unconditionally
wrap the host address in square brackets. This is required by rsync for IPv6
addresses.

If this patch is acceptable, I would greatly appreciate it being back ported to the stable-1.9 branch


amenonsen:
Looks reasonable to me.

jimi-c:
Merged, thanks!

PR #12269, opened 2015-09-07
Python 3: fix TypeError: unorderable types in test

mgedmin:
The full error was

```
ERROR: test_task_executor_execute (units.executor.test_task_executor.TestTaskExecutor)
  File "/home/mg/src/ansible/test/units/executor/test_task_executor.py", line 252, in test_task_executor_execute
    mock_action.run.return_value = dict(ansible_facts=dict())
  File "/home/mg/src/ansible/lib/ansible/executor/task_executor.py", line 317, in _execute
    if self._task.async > 0:
TypeError: unorderable types: MagicMock() > int()

```

Experiments show that Python 2 MagicMock() > 0 is true, so I'm setting the
async property on mock_task to 1.  (If I set it to 0, the test fails anyway.)

### DIFF
--- a/test/units/executor/test_task_executor.py
+++ b/test/units/executor/test_task_executor.py
@@ -220,6 +220,10 @@ class TestTaskExecutor(unittest.TestCase):
         mock_task.changed_when = None
         mock_task.failed_when = None
         mock_task.post_validate.return_value = None
+        # mock_task.async cannot be left unset, because on Python 3 MagicMock()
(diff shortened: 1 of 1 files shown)


abadger:
If you add a comment about why that's there and why 1 was chosen I think
that will be fine.  (That way, if someone wants to specifically test when
async < 1 they'll know why that was chosen for this specific instance.)

mgedmin:
`git blame` not good enough? ;)

abadger:
Thanks @mgedmin!

PR #12270, opened 2015-09-07
Fix rsync connections to IPv6 addresses

jaingaurav:
Similar to https://github.com/ansible/ansible/pull/11816 we can unconditionally
wrap the host address in square brackets. This is required by rsync for IPv6
addresses.

### DIFF
--- a/lib/ansible/runner/action_plugins/synchronize.py
+++ b/lib/ansible/runner/action_plugins/synchronize.py
@@ -45,9 +45,9 @@ class ActionModule(object):
 
         if not host in ['127.0.0.1', 'localhost']:
             if user:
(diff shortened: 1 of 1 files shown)


sivel:
~~errant request for rebase~~

sivel:
@jaingaurav apologies on that last comment, I'm going to get rid of that. My inspection was flawed and didn't account for a submission against, stable-1.9

abadger:
Due to the intricacy of synchronize and the danger of regressions when we modify its code we're not going to fix this in 1.9.x.  This should already be fixed in 2.0.x so we encourage you to upgrade if you need this functionality.

PR #12272, opened 2015-09-07
Added support for generating OTP tokens using filters

rosmo:
##### SUMMARY
Adds two new filters: secret|totp_token() and secret|hotp_token(at=0). Requires PyOTP and complains if it's not installed.

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
lib/ansible/plugins/filter/otp.py

##### ANSIBLE VERSION
2.3

### DIFF
--- a/lib/ansible/plugins/filter/otp.py
+++ b/lib/ansible/plugins/filter/otp.py
@@ -1,0 +1,56 @@
+# -*- coding: utf-8 -*-
+# (c) 2015, Taneli Leppä <rosmo@rosmo.fi>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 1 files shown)


rosmo:
I think a better place this would be Galaxy, but I don't think it supports filters.

bcoca:
The would work in a role within a filter_plugins directory

rosmo:
Yeah, but no one will find it if I just stick it in some random repository and might be discouraged to use 2FA.

rosmo:
Ahh, now I understand, you mean a dummy role in Galaxy? That's a good solution!

bcoca:
still, this might merit inclusion here, just not sure what the use case is as ansible cannot really currently use 2FA

rosmo:
I haven't tried but I thought you could use "environment" to set AWS_SECURITY_TOKEN to use 2FA authentication with Amazon modules (and other modules).

bcoca:
how would you handle token expiration and reauthentication requests? <= this is the big problem that prevents us from adding this everywhere.

jctanner:
bot_status

jstoja:
@jctanner Any idea for this PR?

jctanner:
@jstoja No I don't have any plans to move this forward. The author needs to rebase and push it over the line by putting it on a community meeting agenda, or someone else needs to take over the work.

gundalow:
Given https://github.com/ansible/ansible/pull/12272#issuecomment-139379544

> how would you handle token expiration and reauthentication requests? <= this is the big problem that prevents us from adding this everywhere.

I think this should be closed

PR #12287, opened 2015-09-09
Python 3: use six.string_types instead of basestring

mgedmin:
Reduces the number of failing tests from 35 to 17.

Warning: I'm not 100% confident in the correctness of this change.  On
Python 2 nothing changes: six.string_types is basestring (wrapped in a
tuple).  On Python 3 this means byte strings won't be acceptable here,
which I _think_ is right, but I'd rather somebody double-checked.

### DIFF
--- a/lib/ansible/template/__init__.py
+++ b/lib/ansible/template/__init__.py
@@ -22,6 +22,7 @@ __metaclass__ = type
 import ast
 import re
 
+from six import string_types
 from jinja2 import Environment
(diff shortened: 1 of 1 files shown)


abadger:
These two are both good.  In the present code we're striving for those strings to always be text type, not byte type.  In python3, those will have to be text type, not byte type.

abadger:
# Merged

Hi!

issue is resolved.

Thank you!

PR #12288, opened 2015-09-09
ansible_firewalld fact

pwnall:
##### SUMMARY
This is a proof of concept implementation of a `firewalld` fact. The detection logic is lifted from the firewalld` module in extras.

My Web server deployment playbooks can run on bare-metal setups that include firewalld and on OpenStack VMs, which rely on Neutron for firewalling. The playbooks use Ansible's firewalld module, which (understandably) errors out if no firewalld support is present. In my case, I just want to skip firewalld tasks. 

With this fact, I can use a simple `when` clause, as shown below.

``` yaml
- firewalld: ...
  when: ansible_firewalld != false
```

At the moment, I'm doing exactly the same thing for SElinux, taking advantage of the existing `ansible_selinux` fact. This works quite well for me.

I think this fact should be provided by ansible, because it integrates well with the existing firewalld module.  Without this fact, I need to write my own detection logic. It's not difficult, but it makes the plays a bit less elegant.

I manually tested the code in this PR against a Fedora server installation that has firewalld, and a Fedora cloud installation that does not. I'm willing to put in the time to write automated tests, if a bit of guidance is provided.

Thank you for ansible! :heart:

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME

##### ANSIBLE VERSION
2.3

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -43,6 +43,12 @@ except ImportError:
     HAVE_SELINUX=False
 
 try:
+    import firewall.config
+    HAS_FIREWALLD = True
+except ImportError:
+    HAS_FIREWALLD = False
+
+try:
     import json
     # Detect python-json which is incompatible and fallback to simplejson in
     # that case
@@ -145,6 +151,7 @@ class Facts(object):
             self.get_cmdline()
             self.get_public_ssh_host_keys()
             self.get_selinux_facts()
+            self.get_firewalld_facts()
             self.get_fips_facts()
             self.get_pkg_mgr_facts()
             self.get_lsb_facts()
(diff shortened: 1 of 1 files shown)


bcoca:
I'm not sure this belongs with the general facts.

pwnall:
@bcoca Wouldn't supporting this be similar to the `selinux` fact? It seems to me that if Ansible includes a module, like `firewalld`, it should also provide a way to use it safely.

Are you troubled by the fact that `firewalld` is in extras? Would it make sense to have something like `setup_extras` that provides facts which integrate with the modules in extras?

Thank you for taking a look at this!

bcoca:
it is not an 'extras' issue nor do i think we should have a setup_extras module, I think this does not apply broadly enough nor are we forced to deal with it in many modules as selinux forces us with all file modification modules.

alikins:
This looks like a good candidate once we get facts finer grained and more modular.

alikins:
closing this for now, but may revisit if/when we support fact plugins.

Issue #12293, opened 2015-09-09
ansible-vault password file support broken


sheldonh:
##### Issue Type:

Bug Report
##### Ansible Version:

```
ansible 2.0.0 (devel e2e9319f06) last updated 2015/09/09 15:25:39 (GMT +200)
  lib/ansible/modules/core: (detached HEAD 34655e8e29) last updated 2015/09/09 15:27:35 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD 6a3cf63351) last updated 2015/09/09 15:28:44 (GMT +200)
  config file = /home/sheldonh/git/heisenberg-ansible/ansible.cfg
  configured module search path = None
```
##### Ansible Configuration:

None
##### Environment:

N/A
##### Summary:

The `--vault-password-file` option to `ansible-vault` was broken by ff9f5d7d.
##### Steps To Reproduce:

``` shell
echo password > vault-password
echo secret > secret.txt
ansible-vault --vault-password-file=./vault-password encrypt ./secret.txt
```
##### Expected Results:

As per the commit prior to ff9f5d7d:

```
Encryption successful
```
##### Actual Results:

```
Unexpected Exception: read_vault_password_file() takes exactly 2 arguments (1 given)
to see the full traceback, use -vvv
```

jimi-c:
# Closing This Ticket

Hi!

We believe the above commit should resolve this problem for you. This will also be included in the next major release.

If you continue seeing any problems related to this issue, or if you have any further questions, please let us know by stopping by one of the two mailing lists, as appropriate:
- https://groups.google.com/forum/#!forum/ansible-project - for user questions, tips, and tricks
- https://groups.google.com/forum/#!forum/ansible-devel - for strategy, future planning, and questions about writing code

Because this project is very active, we're unlikely to see comments made on closed tickets, but the mailing list is a great way to ask questions, or post if you don't think this particular issue is resolved.

Thank you!

PR #12300, opened 2015-09-09
Vault missing DataLoader for pwd file #12293

robusto:
Fix for #12293. --with-password-file option was broken


jimi-c:
Merged, thanks!

PR #12307, opened 2015-09-10
Doc fixes

jtyr:
This patch is fixing some minor issues in the documentation.


docsite/README.md
```diff
@@ -11,7 +11,7 @@ If you wish to verify output from the markup
 such as link references, you may install sphinx and build the documentation by running
 `make viewdocs` from the `ansible/docsite` directory.  
 
-To include module documentation you'll need to run `make webdocs` at the top level of the repository.  The generated
+To include module documentation you'll need to run `make htmldocs` at the top level of the repository.  The generated
```
bcoca: webdocs is the correct make target

jtyr:
You are right. I have removed that change.

PR #12312, opened 2015-09-10
Support etcd v2. Use this version by default

rdezavalia:
The etcd lookup plugin doesn't work with version 2 of etcd. Merge this pull in order to support etcd v2. There is a new environment variable, ANSIBLE_ETCD_VERSION, in order to choose witch version you want (default to v2) so you can still use ansible with an old etcd version.

### DIFF
--- a/lib/ansible/plugins/lookup/etcd.py
+++ b/lib/ansible/plugins/lookup/etcd.py
@@ -32,10 +32,16 @@ ANSIBLE_ETCD_URL = 'http://127.0.0.1:4001'
 if os.getenv('ANSIBLE_ETCD_URL') is not None:
     ANSIBLE_ETCD_URL = os.environ['ANSIBLE_ETCD_URL']
 
+ANSIBLE_ETCD_VERSION = 'v2'
+if os.getenv('ANSIBLE_ETCD_VERSION') is not None:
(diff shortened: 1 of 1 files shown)


bcoca:
i would default it to 'v1' to keep it compatible with current installations.

rdezavalia:
using 'v1' as defaults breaks new ansible+etcd installs. If you are upgrading ansible, you should check how to make it work with your current etcd deployment or upgrade etcd. Maybe, we could check etcd version on runtime and use it. I will look into it.

gigablah:
If we're going by defaults, then bumping `v1` to `v2` isn't enough since etcd2 canonically uses port 2379 rather than 4001. I think keeping BC would be better.

andrewmeissner:
I need this in production where I work.  We currently use v2 of etcd, so running v1 would certainly break things.  :(  I like the `ANSIBLE_ETCD_VERSION` variable as a solution.

jamesridle:
Is there any way we can get this merged? We are running v2 of etcd and can't downgrade to v1 without destroying things.

bcoca:
i was waiting for @rdezavalia  to make the change i requested, but if we wait a bit more the bulk of etcd installs will be using v2.

rdezavalia:
Sorry for the long delay, but my day work took me away from this for a while. I have just added a new commit with etcd version 1 by default.

PR #12318, opened 2015-09-11
Python 3: make ansible.template.safe_eval() work

mgedmin:
Two things changed in Python 3.4:
- 'basestring' is no longer defined, so use six.string_types
- True/False are now special AST node types (NamedConstant) rather than
  just names

(Good thing we had tests, or I wouldn't have noticed the 2nd thing!)

I found only one place where safe_eval() is called inside the ansible
codebase: in lib/template/**init**.py.  The call to safe_eval(result,
...) is protected by result.startswith('...'), which means result cannot
possibly be a byte string on Python 3 (or startswith() would raise, so
six.string_types (which excludes byte strings on Python 3) is fine here.

Also, while fixing this I noticed a bug in the existing code that tried
to support set literals, so I fixed it too.

### DIFF
--- a/lib/ansible/template/safe_eval.py
+++ b/lib/ansible/template/safe_eval.py
@@ -67,8 +67,8 @@ def safe_eval(expr, locals={}, include_exceptions=False):
     )
 
     # AST node types were expanded after 2.6
-    if not sys.version.startswith('2.6'):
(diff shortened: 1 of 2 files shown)


jimi-c:
Merged, thanks!

PR #12319, opened 2015-09-11
Special case win_slurp and win_setup as well to work with the new module finder

abadger:


abadger:
/cc @cchurch 

Please test if this works to solve the windows issues for you.

jhawkesworth:
I haven't tested anything but 'fetch' is also a module you can use to target windows hosts so might need the same consideration.

abadger:
I realize I also left off the second half of this fix.. need a pull request
in andible-modules-core that renames windows/setup.ps1 and
Windows/slurp.ps1 to be windows/win_setup.ps1 and windows/win_slurp.ps1

abadger:
Just checked and fetch is implemented as assign action plugin (python thaat
runs on the local machine) that calls out to slurp on the remote side.  So
this commit plus a commit renaming the two plugins in ansible-modules-core
should take care of fetch.

jhawkesworth:
Thanks for checking.

I don't think I have any playbooks that explictly call setup or slurp but
presumably the module lookup logic will handle translating the module names
if the are others that do so?

cchurch:
:-1: as this doesn't actually fix the problem.  Even `win_ping` doesn't work due to the changes introduced in 4b895f04e398deb9c156d4e8b6435073fc2e6f18.

cchurch:
Closing this one in lieu of fixing the plugin loader to support https://github.com/ansible/ansible/issues/12332

abadger:
Almost have a different implementation to teset.  Closing this in prep for that pull request.

PR #12323, opened 2015-09-11
New filter get_number_list

stbka:
This filter should return a list of numbers. For instance I use that list in a with_nested loop.
I have added some documentation as well.

### DIFF
--- a/docsite/rst/playbooks_filters.rst
+++ b/docsite/rst/playbooks_filters.rst
@@ -530,11 +530,11 @@ To escape special characters within a regex, use the "regex_escape" filter::
     
 To create a list of numbers (default: beginning from 0, stepping=1)::
 
-    # create list from 0 with length 4; will return [ 0, 1, 2, 3 ]
-    {{ 4 | get_number_list }}
+    # create a list from 0 with length 4; will return [ 0, 1, 2, 3 ]
+    {{ 4 | range }}
     
     # create list from 100 with length 4, stepping=2; will return [ 100, 102, 104, 106 ]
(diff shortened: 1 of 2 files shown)


sivel:
Sounds like would be better called `range`

stbka:
Yes, indeed. Although the parms are a bit different.
I have changed it.

jhawkesworth:
Erm.. isn't there already a 'range' function in jinja2
http://jinja.pocoo.org/docs/dev/templates/#list-of-global-functions
Having 2 things with the same name could get confusing.

bcoca:
this overlaps with ansible's sequence lookup

bcoca:
closing this as it overlaps both with jinja2's range and ansible's sequence

stbka:
As far as I know you cannot use jinja2's range function in tasks/playbooks and ansible's sequence is not combinable with a with_nested loop.

Here's my usecase:

solrCollection_shards_per_host is a number

``` yml
---
- name: set fact
  set_fact:
    listShardsPerHost="{{solrCollection_shards_per_host|range}}"

- name: do something
  ...
  with_nested:
    - "{{solrCollections}}"      # should be something like [ coll1, coll2, coll3 ]
    - "{{listShardsPerHost}}" # should be something like [1, 2, 3, 4]
```

bcoca:
you mean this?:

```
- name: do something
  ...
  with_nested:
    - "{{solrCollections}}"
    - "{{lookup('sequence', 'count=4'}}"
```

stbka:
Yes, this looks great! Did not know the lookup-sequence feature.
Works for me.

Thank you.

Issue #12076, opened 2015-08-25
Remove python style encoding from the beginning of custom modules not written in python. (v2)


liamoehlman:
Adding python style encoding to the beginning of custom modules not written in python prevents execution.

https://github.com/ansible/ansible/blob/devel/lib/ansible/executor/module_common.py#L45

Seems to be the cause of this in nodejs as # is not a valid comment and results in the following error:

```
# -*- coding: utf-8 -*-
^
SyntaxError: Unexpected token ILLEGAL
```

The same custom module is currently working in 1.8

abadger:
That should fix it... Out of curiosity, how does nodejs specify a shebang if # is not a valid comment?  Does it just treat the first line as special?

# Closing This Ticket

Hi!

We believe recent commits (likely detailed above) should resolve this question or problem for you.

This will also be included in the next major release.

If you continue seeing any problems related to this issue, or if you have any further questions, please let us know by stopping by one of the two mailing lists, as appropriate:
- https://groups.google.com/forum/#!forum/ansible-project - for user questions, tips, and tricks
- https://groups.google.com/forum/#!forum/ansible-devel - for strategy, future planning, and questions about writing code

Because this project is very active, we're unlikely to see comments made on closed tickets, but the mailing list is a great way to ask questions, or post if you don't think this particular
issue is resolved.

Thank you!

liamoehlman:
I can confirm that this fixes the problem. Thanks for the quick response.

Node appears to treat the #! as a special case, files with it in can be run via ./filename.js or node filename.js

quinot:
Fix looks dubious to me, interpreter is a full path, so would never start with "python". IIUC you need to:

```
diff --git a/lib/ansible/executor/module_common.py b/lib/ansible/executor/module_common.py
index 22f8c22..c66cd2e 100644
--- a/lib/ansible/executor/module_common.py
+++ b/lib/ansible/executor/module_common.py
@@ -192,7 +192,7 @@ def modify_module(module_path, module_args, task_vars=dict(), strip_comments=Fal
                 interpreter = to_bytes(task_vars[interpreter_config], errors='strict')
                 lines[0] = shebang = b"#!{0} {1}".format(interpreter, b" ".join(args[1:]))
 
-            if interpreter.startswith('python'):
+            if os.path.basename(interpreter).startswith('python'):
                 lines.insert(1, ENCODING_STRING)
         else:
             # No shebang, assume a binary module?
```

quinot:
[...]

PR #12325, opened 2015-09-11
Test for Python must be against basename of interpreter

quinot:
Followup on issue #12076


quinot:
Thanks @bcoca ! I noticed this issue by code reading while updating pull request #11810 for the latest head. I hope that that submission can be considered again once 2.0 is released.

PR #12328, opened 2015-09-11
Test state.tasks_child_state before dereferencing

amenonsen:
I hit a traceback in the FAILED_TASKS case, but it seemed prudent to add
guards in the other similar branches.


jimi-c:
Merged, thanks!

PR #12334, opened 2015-09-12
Clean up Ansible 2.0 release notes in CHANGELOG.md.

geerlingguy:
The Ansible 2.0 release notes in CHANGELOG.md might be the most important part of the Ansible 2.0 release—people need a way to see what's different, and this is the definitive feature list, so it should be concise, clear, and formatted well for GitHub-style markdown.

I was reading through the release notes and working on digging into some of the new features, and noticed that formatting is all over the place (and in many places inconsistent with previous Ansible releases, like with indentation, capitalization and punctuation), and some of the notes were written in a slightly-confusing way.

Hopefully this is a good start; there were a few more changes I would've made to tidy up the notes more, but I figured this is enough for one go.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,265 +4,272 @@ Ansible Changes By Release
 ## 2.0 "Over the Hills and Far Away" - ACTIVE DEVELOPMENT
 
 Major Changes:
- * Introducing the new block/rescue/always directives, allow for making task blocks and introducing exception like semantics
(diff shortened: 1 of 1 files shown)


geerlingguy:
Ack, I forgot that GitHub's view doesn't show per-line differences when whitespace changes. There are about 20 or so small grammar/wording cleanups in the mix, as well as GitHub markdown fixes so the code examples display inline with the list correctly. Worst case, I could split this into multiple commits...

jimi-c:
Merged, thanks!

PR #12335, opened 2015-09-12
Reword 'jump host' section, block quote fixes

skylerbunny:
An attempt to make the new 'jump host' faq section more readable, both in making the section more concise, and in fixing blockquote example sections which were not displaying correctly.

### DIFF
--- a/docsite/rst/faq.rst
+++ b/docsite/rst/faq.rst
@@ -60,31 +60,21 @@ for new users.
 How do I configure a jump host to access servers that I have no direct access to?
 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
-With Ansible version 2, it's possible to set `ansible_ssh_extra_args` as
-an inventory variable. Any arguments specified this way are added to the
-ssh command line when connecting to the relevant host(s), so it's a good
-way to set a `ProxyCommand`. Consider the following inventory group:
+Prior to Ansible version 2.0, it was necessary to configure a suitable `ProxyCommand` for individual hosts in `~/.ssh/config`, or globally for all hosts by setting `ssh_args` in `ansible.cfg`.
+
+With Ansible 2.0 and above, the reserved variable `ansible_ssh_extra_args` can be set as an inventory variable. Arguments specified in this way are appended to the ssh command line when connecting to the relevant host(s), so it's a good way to set a per-group `ProxyCommand`. Consider the following inventory group::
 
     [gatewayed]
     foo ansible_host=192.0.2.1
     bar ansible_host=192.0.2.2
 
-You can create `group_vars/gatewayed.yml` with the following contents:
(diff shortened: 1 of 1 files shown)


amenonsen:
I'm sorry, but I cannot agree that this is either more readable or more useful. The extra `:` are correct and desirable, the rest just seems like churn; in particular, I think it's a bad idea to promote information about v1 and an obsolete version of ssh to front-and-center. The removal of ControlPersist makes the "no need to repeat" part generic to the point where it doesn't need to be said.

skylerbunny:
I'd welcome any or all of the cruft being cut out, leaving what's actually relevant. Much of that I left in because I presumed with it having been approved in the first place, there was a purpose for it having been in there. Axe away.

amenonsen:
Could you please explain exactly what you consider cruft and what you think is relevant? Your patch seems to add two `:`s (which I like), remove a bunch of newlines (which I don't care about), move info about v1 up to the top (which I don't like, but could live with), put information about obsolete versions of SSH up front and remove mention of ControlXXX settings (which I object to).

skylerbunny:
I'd agree that line https://github.com/ansible/ansible/pull/12335/files#diff-19e79a3d695d284e2a43b8bee0fc7c5dR63 is basically cruft either in the current version _or_ my proposed. I moved it to the front because I assumed the comparison was in there for a reason, but personally I could see it being removed entirely.

My main confusion with the original documentation was this sentence:

'These arguments are added to any `ssh_args` that may be configured /// so it isn't necessary to repeat the default `ControlPath` settings in `ansible_ssh_extra_args`.

I did not see how sentence fragment A had any relevance to B.

Intuitively, I would expect an environment variable ansible_ssh_extra_args to be associated with ssh_args; it is in fact appended (and I documented this). I would not expect ssh_args nor ansible_ssh_extra_args to have anything to do with ControlPath; AFAICT, they don't.

As I understand it, ControlPath it is a variable set in ansible.cfg that may be overloaded in global_vars, but as above, have nothing to do with ssh_args/ansible_ssh_extra_args, so I was unclear why ControlPath was discussed at all.

amenonsen:
At 2015-09-24 21:51:46 -0700, notifications@github.com wrote:

> I did not see how sentence fragment A had any relevance to B.

Nevertheless, it is relevant.

Ansible adds "-o ControlMaster=auto -o ControlPersist=60s" to the ssh
command line by default, but does NOT do this if you set ssh_args in
ansible.cfg. So if you want to set ssh_args and also use multiplexed
connections, you must reproduce the default settings in ansible.cfg
in addition to whatever you wanted to add to the command-line.

This repetition is not necessary with ansible_ssh_extra_args, which (as
you correctly point out) is appended to the command-line.

> so I was unclear why ControlPath was discussed at all.

You're right, the explanation could be more clear. Please feel free to
suggest alternative wording.

-- Abhijit

amenonsen:
Also note that `intro_inventory.rst` documents how `ssh_args` and `ssh_extra_args` work.

amenonsen:
#12604 reworks this wording as well, so this can be closed.

PR #12350, opened 2015-09-14
docsite - playbooks_roles - workaround for conditionally included tasks

bborysenko:
Adding example of workaround for conditionally included tasks

### DIFF
--- a/docsite/rst/playbooks_roles.rst
+++ b/docsite/rst/playbooks_roles.rst
@@ -137,8 +137,13 @@ inside another.
    this, consider how you can restructure your playbook to be more
    class/role oriented.  This is to say you cannot use a 'fact' to
    decide what include file to use.  All hosts contained within the
-   play are going to get the same tasks.  ('*when*' provides some
-   ability for hosts to conditionally skip tasks).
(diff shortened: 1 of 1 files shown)


docsite/rst/playbooks_roles.rst
```diff
@@ -137,8 +137,13 @@ inside another.
    this, consider how you can restructure your playbook to be more
    class/role oriented.  This is to say you cannot use a 'fact' to
    decide what include file to use.  All hosts contained within the
-   play are going to get the same tasks.  ('*when*' provides some
-   ability for hosts to conditionally skip tasks).
+   play are going to get the same tasks.
+
+   So something like ``include: {{ ansible_os_family }}.yml`` would not work. But there is a workaround.  With the ``when`` clause you can provide some ability for hosts to conditionally skip tasks::
```
resmo, line 142: this will work with 2.0, so workaround is only for 1.x

abadger:
@OneManOpsTeam Would you care to update this to mention that the workaround is not needed with 2.0?  Once that is done we can merge.

PR #12359, opened 2015-09-14
Fix fetch to not fail for missing file when fail_if_missing=False

cchurch:
Fixes `fetch` action plugin to not fail if file is missing and `fail_if_missing=False` (the default).  Add tests to `test_fetch` role to verify it works as expected.  Also fixes an integration test failure from th `test_win_fetch` role.


jimi-c:
Merged, thanks!

PR #12363, opened 2015-09-14
allow ConfigureRemotingForAnsible.ps1 script from public zone

### LINKED ISSUE #12362: ConfigureRemotingForAnsible.ps1 fails when windows device connected via interface in "Public" zone
##### Issue Type:
- Bug Report
##### Ansible Version:

```
ansible 1.9.3
```
##### Ansible Configuration:

Default
##### Environment:

Mac OS/X control host, Windows 10 remote managed node
##### Summary:

Example script ConfigureRemotingForAnsible.ps1 requires that interfaces on device be designated as either "Domain" or "Private" zone.  Some organizations treat all interfaces as though they are connected to the equally untrusted "Public" zone regardless of where they are connected.  Script should have option to allow winrm to be initialized on public zone interfaces.
##### Steps To Reproduce:

Run ConfigureRemotingForAnsible.ps1 on a windows machine with only Public Zone interfaces

```
PS C:\Users\ansible\Desktop> .\ConfigureRemotingForAnsible.ps1
```
##### Expected Results:

Successful winrm initialization
##### Actual Results:

```
Set-WSManQuickConfig : <f:WSManFault xmlns:f="http://schemas.microsoft.com/wbem/wsman/1/wsmanfault" Code="2150859113"
Machine="localhost"><f:Message><f:ProviderFault provider="Config provider"
path="%systemroot%\system32\WsmSvc.dll"><f:WSManFault xmlns:f="http://schemas.microsoft.com/wbem/wsman/1/wsmanfault"
Code="2150859113" Machine="WIN-5HFR17R2PDT"><f:Message>WinRM firewall exception will not work since one of the network
connection types on this machine is set to Public. Change the network connection type to either Domain or Private and
try again. </f:Message></f:WSManFault></f:ProviderFault></f:Message></f:WSManFault>
At line:116 char:17
+                 Set-WSManQuickConfig -force
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [Set-WSManQuickConfig], InvalidOperationException
    + FullyQualifiedErrorId : WsManError,Microsoft.WSMan.Management.SetWSManQuickConfigCommand
```

breathe:
Fixes https://github.com/ansible/ansible/issues/12362

The current script fails on machines which have network interfaces designated as connected to "Public" networks (choices for network designation being Private, Domain, Public).  This commit adds the optional switch "SkipNetworkProfileCheck" to ConfigureRemotingForAnsible.ps1 -- when passed the switch causes the script to skip the check which enforces that only devices on Private/Domain zones can have winrm initialized.

### DIFF
--- a/docsite/rst/intro_windows.rst
+++ b/docsite/rst/intro_windows.rst
@@ -8,7 +8,7 @@ Windows Support
 Windows: How Does It Work
 `````````````````````````
 
-As you may have already read, Ansible manages Linux/Unix machines using SSH by default.  
+As you may have already read, Ansible manages Linux/Unix machines using SSH by default.
 
 Starting in version 1.7, Ansible also contains support for managing Windows machines.  This uses
 native PowerShell remoting, rather than SSH.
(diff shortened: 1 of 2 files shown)


trondhindenes:
I would probably add the parameter to the script documentation (we should probably do that with the rest of the params as well)

breathe:
I added some documentation to describe the purpose of the switch -- is this the only place that needs to be specified?

trondhindenes:
I was thinking here: https://github.com/ansible/ansible/blob/devel/docsite/rst/intro_windows.rst

breathe:
Added some documentation of two of the arguments to ConfigureRemotingForAnsible.ps1 to the docsite

trondhindenes:
Nice. +1 from me, lgtm.

jhawkesworth:
The code changes look good to me too.  There seems to be some whitespace difference in the documentation changes you have made - would be nice if you could fix and perhaps squash into a single commit though?

PR #12368, opened 2015-09-15
Contrib Inventory: NSoT

coxley:
## Network Source of Truth

Flexible IPAM/CMDB by Dropbox, I feel like this is a good feature to have upstream. Didn't notice any other similar choices, ignoring cloud inventory providers because this enables network engineers to have a better source of truth for their devices.

Full README is located over at [coxley/ansible_nsot](https://github.com/coxley/ansible_nsot) where I have some integration tests for parameters and a basic HTTP mock. I've tried to make it as easy for the user to take advantage of this.

### DIFF
--- a/contrib/inventory/nsot.py
+++ b/contrib/inventory/nsot.py
@@ -148,9 +148,6 @@ from pynsot.client import get_api_client
 from pynsot.app import HttpServerError
 from click.exceptions import UsageError
 
-# Version source of truth is in setup.py
-__version__ = pkg_resources.require('ansible_nsot')[0].version
-
 
 def warning(*objs):
         print("WARNING: ", *objs, file=sys.stderr)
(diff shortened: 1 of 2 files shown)


contrib/inventory/nsot.py
```diff
@@ -0,0 +1,345 @@
+#!/bin/env python2.7
```
bcoca: you might want to make it python vs python2.7 which will not work on all platforms

coxley: Done

coxley:
Are there any other issues here? :)

coxley:
This has been running in production for months at our place reliably. If there are any concerns, just let me know.

bcoca:
sorry for the delay, i was hoping someone from the community would review and tests this, I did not see anything that would prevent us from merging it so doing it now.

PR #12337, opened 2015-09-13
Correct default plugin search path

jan-warchol:
I have ansible 1.9.3 installed, and the configuration file in `/etc/ansible/ansible.cfg`
defines default plugin search paths as
    /usr/share/ansible_plugins/...
rather than
    /usr/share/ansible/plugins/...

PR #12378, opened 2015-09-15
normalized plugin paths and names and configs

bcoca:
now all configs, directories and code match and are singular (except vars ... which is really data loader?)

fixes #12337


jimi-c:
Merged.

PR #12379, opened 2015-09-15
Document vars

bcoca:
now specific precedence for v2 also explaining scope and connection vars a bit more.

### DIFF
--- a/docsite/rst/playbooks_variables.rst
+++ b/docsite/rst/playbooks_variables.rst
@@ -732,7 +732,20 @@ However, let's go ahead and get precedence out of the way!  It exists.  It's a r
 a use for it.
 
 If multiple variables of the same name are defined in different places, they get overwrriten in a certain order.
-The order of precedence is (last one wins):
+
+In 1.x the precedence is:
+
+ * extra vars (``-e`` in the command line) always win
+ * then come connection variables (``ansible_user``, etc)
+ * then comes "most everything else" (command line switches, vars in play, included vars, role vars, etc)
+ * then come the variables defined in inventory
+ * then come the facts discovered about a system
+ * then "role defaults", which are the most "defaulty" and lose in priority to everything.
+
+.. note:: In versions prior to 1.5.4, facts discovered about a system were in the "most everything else" category above.
+
+
+In 2.x we have made the order of precedence more specific (last one wins):
 
   * role defaults [1]_
   * inventory vars [2]_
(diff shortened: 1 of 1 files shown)


docsite/rst/playbooks_variables.rst
```diff
@@ -731,19 +731,77 @@ it complicated.
 However, let's go ahead and get precedence out of the way!  It exists.  It's a real thing, and you might have
 a use for it.
 
-If multiple variables of the same name are defined in different places, they win in a certain order, which is::
+If multiple variables of the same name are defined in different places, they get overwrriten in a certain order.
```
resmo: typo detected

docsite/rst/playbooks_variables.rst
```diff
@@ -731,19 +731,77 @@ it complicated.
...
+  * play vars_prompt
+  * play vars_files
+  * role and include vars
+  * block vars (only for tasks in block)
+  * task vars (only for the task)
+  * extra vars
+
+.. rubric:: Footnotes
+
+.. [1] Tasks in each role will see their own role's defaults tasks outside of roles will the last role's defaults
```
jimi-c, line 769: Wording here needs to be cleaned up.

docsite/rst/playbooks_variables.rst
```diff
@@ -731,19 +731,77 @@ it complicated.
...
+In 2.x we have made the order of precedence more specific (last one wins):
+
+  * role defaults [1]_
+  * inventory vars [2]_
+  * inventory group_vars
+  * inventory host_vars
+  * playbook group_vars
+  * playbook host_vars
+  * host facts
+  * registered vars
```
jimi-c, line 757: And registered vars/set_fact/include_vars are higher precedence than this, since they've been split out from facts gathered. Things which were in `vars_cache` (now `vars_cache` and `nonpersistent_fact_cache` override just about everything except extra vars.

docsite/rst/playbooks_variables.rst
```diff
@@ -731,19 +731,77 @@ it complicated.
...
+.. _variable_scopes:
+
+Variable Scopes
+```````````````
+
+Ansible has 3 main scopes:
+
+ * Global: this is set by config, environment variables and the command line
+ * Play: each play and contained structures, vars entries, include_vars, role defaults and vars.
+ * Host: variables directly associated to a host, like inventory, facts or registered task outputs
```
jimi-c, line 797: There is also parameter scope for roles and includes, in which only tasks within that role/include will see the parameter variables.

jimi-c, line 797: Probably also need to mention the new 2.0 option here which enables `role scope` (for vars only, not defaults).

docsite/rst/playbooks_variables.rst
bcoca, line 769: I take suggestions

docsite/rst/playbooks_variables.rst
bcoca, line 797: i do mention the container structures, we can be more explicit and even give examples. the 2.0 option i would leave out for now, it will create many problems with galaxy roles.

docsite/rst/playbooks_variables.rst
bcoca, line 757: not sure that should be the case, moving host and registered, also i thought we just used include_vars as play vars that were defined afterwards

Issue #12381, opened 2015-09-15
Environment not taken into account when using async


AxelDelmas:
I'm trying to set an env var DISPLAY=:99 in order to launch chromium with ansible, but when I use async I get an error message "... Gtk: cannot open display: ", similar to what I get when I connect manually in ssh and forget to do
export DISPLAY=:99
before trying to launch chromium.

If I remove the async action, chromium launches successfully but then the playbook execution hangs there.

jimi-c:
# Closing This Ticket

Hi!

We believe the above commit should resolve this problem for you. This will also be included in the next major release.

If you continue seeing any problems related to this issue, or if you have any further questions, please let us know by stopping by one of the two mailing lists, as appropriate:
- https://groups.google.com/forum/#!forum/ansible-project - for user questions, tips, and tricks
- https://groups.google.com/forum/#!forum/ansible-devel - for strategy, future planning, and questions about writing code

Because this project is very active, we're unlikely to see comments made on closed tickets, but the mailing list is a great way to ask questions, or post if you don't think this particular issue is resolved.

Thank you!

PR #12382, opened 2015-09-15
Added env_string in async_cmd

AxelDelmas:
Noticed this unused env_string in async.py, I added it to the async command.
Not completely sure what I did here (first time I look under the hood with ansible), but it did the trick for me in order to solve Issue #12381


jimi-c:
Merged, thanks!

PR #12387, opened 2015-09-15
Small change so the YAML renders as code.

jkugler:
The line above the last YAML code block in this document did not have the requisite :: characters, so the YAML did not render correctly in the docs.


jimi-c:
Merged, thanks!

PR #12389, opened 2015-09-16
Set PlayContext.timeout from the value specified by -T on the command-line

### LINKED ISSUE #12372: v2 regression: -T timeout doesn't work
With devel, `ansible -T 30` doesn't seem to set `self._play_context.timeout` at all (but setting `ANSIBLE_TIMEOUT` in the environment does). Anyone working on this already?

amenonsen:
Fixes #12372


jimi-c:
Merged, thanks!

PR #12401, opened 2015-09-16
Fix self.groups dict iteration

mcsalgado:


jimi-c:
Already fixed in 11c27078c01f8928f7135a2064f4c9983281aa87, thanks though!

PR #12402, opened 2015-09-16
Fix urlopen usage to use open_url instead

abadger:
- Change urlopen usage to use open_url
- Add a travis test for urlopen usage


abadger:
I appear to have merged this already.  Closing.

PR #12414, opened 2015-09-17
Only use become if the target user difers from the current remote user.

apollo13:
The idea behind this change is as follows: Usually I have bootstrap script which require sudo, but if someone already has access as root there is no need for a switch… While this could be implemented in the playbook itself using something like `become: "{{ ansible_ssh_user != 'root' }}"` (which is currently not possible either), it seems way nicer to to that in ansible itself. Personally I cannot see any reason to use sudo if the target user is already the current user. Thoughts?

### DIFF
--- a/docsite/rst/intro_configuration.rst
+++ b/docsite/rst/intro_configuration.rst
@@ -678,6 +678,18 @@ Ask for privilege escalation password, the default is False::
 
     become_ask_pass=True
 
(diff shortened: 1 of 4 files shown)


apollo13:
@jimi-c Added a test as requested, anything else to do?

jimi-c:
Merged, thanks!

PR #12404, opened 2015-09-16
Fix IPv6 addresses getting dropped from hosts list (devel branch)

mcsalgado:
Also see #12329

PR #12415, opened 2015-09-17
Use ',' instead of ':' or ';' to separate host patterns

amenonsen:
The earlier-recommended "pat1:pat2:pat3[x:y]" notation doesn't work well
with IPv6 addresses, so we recommend ',' as a separator instead. We know
that commas can't occur within a pattern, so we can just split on it.

We still have to accept the "foo:bar" notation because it's so commonly
used, but we issue a deprecation warning for it.

Closes #12404

### DIFF
--- a/lib/ansible/inventory/__init__.py
+++ b/lib/ansible/inventory/__init__.py
@@ -24,6 +24,7 @@ import os
 import sys
 import re
 import stat
+import itertools
 
 from ansible import constants as C
 from ansible.errors import AnsibleError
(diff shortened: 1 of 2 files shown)


jody-frankowski:
It seems to be handled in the best possible way. Mishandling ':'-separated ipv6 list isn't something we can do much about. It certainly deserves to be merged before v2.

mcsalgado:
I tested a bit before the address regex change (991d281) and it seems to be working ok.

amenonsen:
@mcsalgado thanks for testing. The address regex change is incidental and should not affect your case, I just happened to find it while writing tests for this PR.

mcsalgado:
I absolutely agree that using comma as a separator is a much more sensible choice here. But it still feels a little iffy when you pass the patterns as list. It turns the list into a string and then right after it goes back to a list again. It's kinda roundabout so I submitted a PR (amenonsen/ansible#1) addressing that.

amenonsen:
I've merged your pull request (which has updated this PR). Thanks for your help.

jimi-c:
Merged, thanks!

PR #12420, opened 2015-09-17
Fix for user defined modules not overriding modules from core.

abadger:
This fix takes into account that powershell modules are somewhat
different than regular modules and have to be kept separate.

### DIFF
--- a/lib/ansible/plugins/__init__.py
+++ b/lib/ansible/plugins/__init__.py
@@ -26,10 +26,9 @@ import inspect
 import os
 import os.path
 import sys
+from collections import defaultdict
 
 from ansible import constants as C
-from ansible.utils.unicode import to_unicode
-from ansible import errors
 
 try:
     from __main__ import display
@@ -37,6 +36,7 @@ except ImportError:
     from ansible.utils.display import Display
     display = Display()
 
+# Global so that all instances of a PluginLoader will share the caches
 MODULE_CACHE = {}
 PATH_CACHE = {}
 PLUGIN_PATH_CACHE = {}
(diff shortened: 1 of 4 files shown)


abadger:
/cc @j2sol @cchurch @jhawkesworth @trondhindenes 

Okay, here's a new potential fix for the related problems of ansible modules with no extension in custom library dirs not overriding the ones whipped with ansible and python modules being found instead of windows modules when using winrm.  Test it out to see that it keeps everything working.

cchurch:
With the following change applied to fix setup/slurp, it seems to work for me:

```
diff --git a/lib/ansible/plugins/action/__init__.py b/lib/ansible/plugins/action/__init__.py
index 13166e9..7a35feb 100644
--- a/lib/ansible/plugins/action/__init__.py
+++ b/lib/ansible/plugins/action/__init__.py
@@ -78,8 +78,12 @@ class ActionBase:
                 if module_name in ('stat', 'file', 'copy') and self._task.action != module_name:
                     module_name = 'win_%s' % module_name

+                # Explicitly use .ps1 versions of setup/slurp.
+                if module_name in ('setup', 'slurp'):
+                    module_name = '%s.ps1' % module_name
+
                 # Remove extra quotes surrounding path parameters before sending to module.
-                if module_name in ('win_stat', 'win_file', 'win_copy', 'slurp') and module_args and hasattr(self._connection._shell, '_unquote'):
+                if module_name in ('win_stat', 'win_file', 'win_copy', 'slurp.ps1') and module_args and hasattr(self._connection._shell, '_unquote'):
                     for key in ('src', 'dest', 'path'):
                         if key in module_args:
                             module_args[key] = self._connection._shell._unquote(module_args[key])
```

abadger:
The cache in `plugins/__init__.py` is supposed to take care of separating out .ps1 from .py now.  I took a look and what I was doing wrong was that when scanning a directory for plugins I needed to enter all of the files found whether they matched the present suffix type or not.  Then, when deciding whether the module is already cached, use the suffix to look at the subset of the newly cached files to determine whether the module has been found.  https://github.com/ansible/ansible/commit/f98a46896c6fab431f3383b1c763d469d1178fa0 should address that problem.

abadger:
I'll be on vacation and away from the internet until Monday (and I'll be flying home on that day.)  So if ya'll teest and this works, feel free to merge without me.

cchurch:
LGTM :+1:

PR #12422, opened 2015-09-17
Merge Host.ipv[46]_address into .address

amenonsen:
The earlier distinction was never used; .ipv6_address was always a copy
of .ipv4_address, and the latter was always used to set the remote_addr
field in the PlayContext.

Also uses the canonical ansible_host/ansible_port names when setting the
address and port from variables.


jimi-c:
Merged, thanks!

PR #12429, opened 2015-09-17
Update synchronize module plugin to work on OS X

petrmifek:
As discussed at the commit @6c107258fad96fca53bea5a74960f54ac0d7b45e Those changes broke synchronize module functionality on OS X, where the built in rsync is 2.6.9 and has some issues parsing the user@[hostname-or-ipv4-or-ipv6] format in destination.

Main issue is that it completely breaks the synchronize functionality even for IPv4 on current OS X.

The hack below seems to be working on both OS X and Linux (with newer rsync) and with IPv4/6 addresses and A and AAAA DNS names.

The rsync, especially on OS X, is quite picky about the format of SRC and DEST, so some independent testing would be good. Still, this allows for using the synchronize module again on a stock OS X without having to reinstall rsync on all the servers and clients.

Thanks for consideration!

### DIFF
--- a/lib/ansible/plugins/action/synchronize.py
+++ b/lib/ansible/plugins/action/synchronize.py
@@ -47,10 +47,13 @@ class ActionModule(ActionBase):
     def _process_origin(self, host, path, user):
 
         if host not in C.LOCALHOST:
+            userPart = ''
             if user:
-                return '%s@[%s]:%s' % (user, host, path)
+                userPart = '%s@' % (user, )
+            if ":" in host:
+                return '[%s%s]:%s' % (userPart, host, path)
             else:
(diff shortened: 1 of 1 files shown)


jaingaurav:
An unfortunate ugly hack. Sorry I didn't do much testing on OS X, but I was able to confirm that this seems to be a reasonable fix.

abadger:
@petrmifek If you'll move _needs_ipv6_brackets() up so it's not a nested function, this looks good to merge.

petrmifek:
Thanks. Have a look now please.

abadger:
Thanks @petrmifek for the code  and  @jaingaurav for testing the change!

# Merged

Hi!

issue is resolved.

Thank you!

Issue #11957, opened 2015-08-13
add_host module returns 'NoneType' object


davidvasandani:
##### ISSUE TYPE
Bug Report

##### COMPONENT NAME
add_host module

##### ANSIBLE VERSION
1.9.0.1

##### SUMMARY
Trying to use the add_host module to add fresh servers to a group for tasks later in the playbook I keep getting this error. It will fail the first run but during the second run will work fine. I've included the two previous tasks to show how the with_items works.

My tasks:

```
- name: set hostnames
  hostname: >
    name="{{ item.tags.Name }}"
  sudo: yes
  delegate_to: "{{ item.private_ip }}"
  with_items: created.tagged_instances

- debug: var=item
  with_items: created.tagged_instances

- name: add hosts to fresh group
  add_host:
    hostname="{{ item.private_ip }}"
    group=fresh
  with_items: created.tagged_instances
```

The errors.

```
TASK: [cep_aws | set hostnames] ***********************************************
<172.20.16.148> REMOTE_MODULE hostname name="awsqa-cep-primary00"

[...]

PR #12431, opened 2015-09-17
Stops using _groups_list to check for host when using 'add_host'

hslee16:
May be related to https://github.com/ansible/ansible/issues/11957

Using the 'add_host' function in current devel branch causes playbooks to return exceptions stating 

Inventory has no attribute '_groups_list'

This PR aims to use the existing attribute in the inventory and groups objects to work around this issue.

[EDIT]
For context, I am using 'add_host' in the following manner

```
- name: Add instance IP to new hosts group
  add_host: name=192.168.30.31 groupname=my_host_ip
```

### DIFF
--- a/lib/ansible/inventory/__init__.py
+++ b/lib/ansible/inventory/__init__.py
@@ -457,7 +457,10 @@ class Inventory(object):
         return matching_host
 
     def get_group(self, groupname):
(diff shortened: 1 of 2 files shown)


amenonsen:
Looks reasonable.

PR #12432, opened 2015-09-18
Python 3: there's no xrange

mgedmin:
Use six.moves.range instead (aliased to xrange on Python 2, aliased to range
on Python 3).

Also I couldn't resist replacing the elaborate chr/ord/randrange dance with
the simpler random.choice(string.ascii_lowercase) that was already used
elsewhere in the Ansible codebase.


jimi-c:
Merged, thanks!

PR #12433, opened 2015-09-18
Update intro_windows.rst

tklaus:
Arch linux contains python2 and python3. Using the standard pip command will target python3.


bcoca:
this should be a note and more generic, this is not only an Archlinux issue,

.. note: Make sure you are using pip for python2 as on some platforms it might be mapped to pip for python3

nitzmahone:
Fixed with distro-agnostic text by d699ef72d22874413c4c90d38f89a9a16c663154

PR #12442, opened 2015-09-18
add filter_plugins.mathstuff.to_bytes + tests

bmoar:
I ran into a problem where I needed to convert a memory unit string such
as "8GB" into a number of bytes to do some calculations and set vars
based on these calculations. I couldn't find a simple way to do this
without resorting to multiple variables and complicated stuff, so I
wrote this filter_plugin to do the conversion. Also wrote unit tests to ensure
bad input is handled and the logic is robust.

```
In [6]: from ansible.plugins.filter.mathstuff import human_readable, to_bytes

In [7]: human_readable(to_bytes('8 GB'))
Out[7]: '8.00 GB'
```

~~~start git commit msg

the to_bytes filter plugin allows you to convert
memory unit strings such as "8GB", "64kb", etc.

ex setting a variable:

```
postgres_shared_buffers = "8GB"
postgres_nr_huge_pages = "{{ ( postgres_shared_buffers|to_bytes / 1024 / 2048 * 1.05 ) |round |int }}"
```

the filter plugin works with any string in the format

VALUE UNIT

where valid VALUE can be any positive integer 0-max, [0-9]+

where valid UNIT can be lower or uppercase (the UNIT is parsed with .lower()) in the form:

```
basebytes = ('b', )
kilobytes = ('k', 'kb', 'kib', )
megabytes = ('m', 'mb', 'mib', )
gigabytes = ('g', 'gb', 'gib', )
terabytes = ('t', 'tb', 'tib', )
petabytes = ('p', 'pb', 'pib', )
```

the VALUE and UNIT can be seperated by any number of whitespace, including no space.

Some example formats are:

```
    "64kb"
    "64 kb"
    "64       kb"
    "64kib"
    "64KB"
    "64Kb"
    "64k b"
```

There is also an optional si flag for the filter, allowing you to
convert the input value into bytes with the SI units for memory.
ex:

```
    "8KB" | to_bytes(True) = 8 * (1000 ** 1)
    "8MB" | to_bytes(True) = 8 * (1000 ** 2)
```

### DIFF
--- a/docsite/rst/playbooks_filters.rst
+++ b/docsite/rst/playbooks_filters.rst
@@ -257,6 +257,13 @@ Square root, or the 5th::
     {{ myvar | root }}
     {{ myvar | root(5) }}
 
+Convert memory units to number of bytes, standard by default, or SI with an optional arg
+
+    {{ '64kb' | to_bytes }} => 65536
+    {{ '8 GB' | to_bytes }} => 8589934592.0
+    {{ '64kb' | to_bytes(True) }} => 64000
+    {{ '8 GB' | to_bytes(True) }} => 8000000000.0
+
 Note that jinja2 already provides some like abs() and round().
 
 .. _ipaddr_filter:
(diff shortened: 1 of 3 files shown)


bcoca:
dupe of #12074

bmoar:
closing due to dupe, thank you.

PR #12446, opened 2015-09-21
Use dict.get()

mcsalgado:


jimi-c:
Thanks, I saw that and was going to fix it later myself.

PR #12447, opened 2015-09-21
Python 3: there is no 'basestring'

mgedmin:
This fixes a failing unit test.

In actual use (which is still quite far), I'm not sure if bytes -> unicode
conversion should be done here (in which case the code will fail with an
AttributeError: 'bytes' object has no attribute 'readlines'), or inside
self._connection.exec_command() (in which case my change is correct).


jimi-c:
Merged, thanks!

PR #12451, opened 2015-09-21
added --pass-limit-to-inventory as discussed in https://groups.google.com/forum/#!topic/ansible-devel/ZGwvE7RgWTc

bigzaqui:
##### SUMMARY
the idea is simple, if  --pass-limit-to-inventory is used when running ansible-playbook, then when the inventory is called, it will have the arguments --list --limit BLABLA 

so you can use it to narrow the scope of the inventory

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
lib/ansible/cli/__init__.py
lib/ansible/cli/playbook.py
lib/ansible/inventory/__init__.py
lib/ansible/inventory/dir.py
lib/ansible/inventory/script.py

##### ANSIBLE VERSION
2.3

### DIFF
--- a/lib/ansible/cli/__init__.py
+++ b/lib/ansible/cli/__init__.py
@@ -230,6 +230,8 @@ class CLI(object):
                 help='outputs a list of matching hosts; does not execute anything else')
             parser.add_option('-l', '--limit', default=C.DEFAULT_SUBSET, dest='subset',
                 help='further limit selected hosts to an additional pattern')
+            parser.add_option('--pass-limit-to-inventory', default=False, dest='passlimit', action='store_true',
+                help='pass the --limit argument to the inventory script')
 
         if module_opts:
             parser.add_option('-M', '--module-path', dest='module_path', default=None,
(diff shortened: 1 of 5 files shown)


gundalow:
@bigzaqui We were discussion this PR in networktocode channel today. It seems like there is some interest in this still. Would you be able to rebase it please.

ogenstad:
+1

abadger:
From @bcoca on IRC:

> not good idea, several issues, not smallest one is that it disallows playbooks to access data outside limit, which most need
> having a caching inventory is normally a better solution

dbarrosop:
> not good idea, several issues, not smallest one is that it disallows playbooks to access data outside limit, which most need

That might be assuming too much. How do you know how his or my backend work? That is extremely useful for people running their own inventories as they can use that limit to determine the data to fetch from their backend. If you want to ignore the feature for people using the `hosts` file I think it's fine but custom inventories should have that information. People with their inventory scripts can also choose to ignore that limit if they don't like it.

> having a caching inventory is normally a better solution

If your operations are data driven a cache is useless.

ogenstad:
Given that this would only apply with the optional `--pass-limit-to-inventory` parameter I don't really understand the objection about accessing data outside of the limit.

What are the other issues not mentioned?

calfonso:
Hi!

We believe recent commits  #23001 should resolve this question or problem for you.

Thank you!

PR #12456, opened 2015-09-21
Replace categories with tags.

chouseknecht:
Galaxy is deprecating categories in favor of tags. Each tag is a single word that describes or categorizes a role. Role authors can supply a set of tags that best describes their role. In the next Galaxy release users will be able to search roles by tags.

This PR is intended to update the metadata template and ansible-galaxy 'search' command to reflect this change.

To aid testing added --ignore-certs option for use with install and search commands. Helpful when
using --server to point to a server with a self signed cert.

### DIFF
--- a/docsite/rst/developing_plugins.rst
+++ b/docsite/rst/developing_plugins.rst
@@ -19,7 +19,7 @@ in playbooks and with /usr/bin/ansible to decide how you want to talk to remote 
 are covered in the :doc:`intro_getting_started` section.  Should you want to extend Ansible to support other transports (SNMP? Message bus?
 Carrier Pigeon?) it's as simple as copying the format of one of the existing modules and dropping it into the connection plugins
 directory.   The value of 'smart' for a connection allows selection of paramiko or openssh based on system capabilities, and chooses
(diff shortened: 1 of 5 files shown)


bcoca:
this breaks backwards compatibility and overloads the meaning of --tags, I would prefer to keep them as categories in ansible core even if we have to translate to 'tags' in the api calls.

jimi-c:
@bcoca I had @chouseknecht change the YAML entry to `galaxy_tags` to differentiate them from regular tags.

bcoca:
it still creates ambiguity and it just seems a change for the sake of change, since we are keeping categories anyways as an equivalent.

jimi-c:
Categories will be removed eventually, and I don't think it's ambiguous with the label change.

bcoca:
You normally change names for one of 2 reasons, to disambiguate or to clarify, this seems to be the contrary. As much as galaxy_tags is not the same as tags, it will be shortened in communications and create confusion, this seems a counterproductive change with no upside.

Am I missing something? the functionality is being implemented for both names in the short term, so the name change is not required for that. If we still want to change the name it should never be to a term that is already in use and overload it.

jimi-c:
I manually cherry-picked the three relevant commits from this to the devel branch, as there are two included that have bled over into this branch. Closing manually.

PR #12457, opened 2015-09-21
Detect if core modules aren't installed and warn if that is the  case

### LINKED ISSUE #11206: anisble fails to alert when modules are missing
Is there changes in yml format in ansible dev latest? I have a simplest playbook that works fine with ansible 1.9.1 but fails with dev latest.

Playbook file:

```

---
- hosts: localhost
  connection: local
  tasks:
  - name: echo test
    shell: echo test
```

Running against ansible 1.9.1

```
[xxx]$ ansible --version
ansible 1.9.1
[xxx]$ ansible-playbook test.yml 

ok: [localhost]

TASK: [echo test] ************************************************************* 
changed: [localhost]

```

Running against dev latest

```
[yyy]$ ansible --version
ansible 2.0.0

[yyy]$ ansible-playbook test.yml  -i hosts
ERROR! no action detected in task

The error appears to have been in '/tmp/test.yml': line 5, column 5, but may
be elsewhere in the file depending on the exact syntax problem.

  tasks:
  - name: echo test
    ^ here

```

abadger:
Fixes #11206

### DIFF
--- a/lib/ansible/parsing/mod_args.py
+++ b/lib/ansible/parsing/mod_args.py
@@ -275,7 +275,14 @@ class ModuleArgsParser:
 
         # if we didn't see any module in the task at all, it's not a task really
         if action is None:
-            raise AnsibleParserError("no action detected in task", obj=self._task_ds)
+            if 'ping' not in module_loader:
+                raise AnsibleParserError("The requested action was not found in configured module paths. "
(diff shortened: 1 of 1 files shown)


abadger:
We also have code here: https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/action/__init__.py#L96  But that's not being hit in this case.

PR #12471, opened 2015-09-22
Make pipelining a connection variable

amenonsen:
SSH pipelining can be a significant performance improvement, but it will
not work if sudoers is configured to requiretty. With this change, one
could have pipelining enabled in ansible.cfg, but use sudo to turn off
requiretty in a separate play where pipelining is disabled:

```
- hosts: foo
  pipelining: no
  tasks:
    - lineinfile: dest=/etc/sudoers line='Defaults requiretty' state=absent
      sudo_user: root
```

(Note that sudoers has a complicated syntax, so the above lineinfile
invocation may be too simplistic for production use; but the point is
that a separate play can do something to disable requiretty.)

### DIFF
--- a/docsite/rst/intro_inventory.rst
+++ b/docsite/rst/intro_inventory.rst
@@ -217,6 +217,9 @@ SSH connection::
     ansible_ssh_extra_args
       Additional arguments for ssh. Useful to configure a ``ProxyCommand`` for a certain host (or group).
       This is used in addition to any ``ssh_args`` configured in ``ansible.cfg`` or the inventory.
+    ansible_ssh_pipelining
+      Determines whether or not to use SSH pipelining. This can override the
+      ``pipelining`` setting in ``ansible.cfg``.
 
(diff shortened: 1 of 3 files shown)


bcoca:
I would not make this a play variable, just another ssh connection variable

amenonsen:
@bcoca please explain?

bcoca:
This makes a lot more sense as 'ansible_ssh_pipelining' which as a connection var becomes a lot more flexible,  can now be set by host, play, task or globally (which we already have).

Since 'accelerate' we've been trying to keep connection specific vars outside of play/task directives.

amenonsen:
@bcoca OK, done the way you suggested.

bcoca:
+1

abadger:
Looks good to me, +1 as well.

abadger:
Thanks @amenonsen ! 

# Merged

Hi!

issue is resolved.

Thank you!

PR #12472, opened 2015-09-22
cloudstack: improvements and tests

resmo:
- common tags handling integration tested, account and user extended
- tags handling in cloudstack utils is now idempotent, thanks to @jeffersongirao
- add common argument spec into cloudstack utils, PR for modules implementing it, will follow.
- add general string to int handling for response values, can be used in modules where needed.


gregdek:
cc: @ansible/core team for fast tracking tests here.

PR #12481, opened 2015-09-22
Clarified that it's fine if group vars files have '.yml' extension

jeffwidman:
This seemed unintuitive to me reading the docs, because some ansible yaml files can end in .yml and some can't, so tested it and then clarified the docs.


bcoca:
actually, they should also allow for ".yaml" and ".json"

jeffwidman:
argh, messed up my git history rewriting when I tried to amend the commit... continued here: https://github.com/ansible/ansible/pull/12483

PR #12483, opened 2015-09-23
Clarified that .yml, .yaml, and .json are allowable file extensions for inventory vars files

jeffwidman:


jeffwidman:
Need anything else here?

PR #12487, opened 2015-09-23
Fix one more failing test on Python 3

mgedmin:
This required me to add some Python 3 compatibility to
module_utils/basic.py, which makes this PR larger than usual.  I tried
to split it into small, semantically coherent commits for ease of
review.

### DIFF
--- a/lib/ansible/inventory/__init__.py
+++ b/lib/ansible/inventory/__init__.py
@@ -26,6 +26,8 @@ import re
 import stat
 import itertools
 
+from six import string_types
+
 from ansible import constants as C
 from ansible.errors import AnsibleError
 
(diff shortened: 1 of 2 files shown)


abadger:
Thanks @mgedmin !

abadger:
@mgedmin we found out that six is python-2.6 and above so we can't use it here.  I'll work up an alternative for six.string_types in basic.py.  Just wanted to mention it here so that we remember this constraint for other module code.

abadger:
Oops, my bad.  six is being used in the inventory, not in basic.py.  So we're fine.

mgedmin:
I was avoiding six in module_utils/ because it might not be present on the remote host.

PR #12488, opened 2015-09-23
Fix UnicodeEncodeError when writing tree files (1.9.x)

reinout:
This is the very same fix as in #10630. The fix there applies to writing the json fact files to the cache. 

The same error occurs when you write the facts explicitly to a directory with `--tree some/directory`.

Using `--tree` is no longer really needed now that there's a json fact cache, but as long as the code is still there I thought it best to fix it.

**Note:** I based it off the stable-1.9 branch as that looked like the 1.9 bugfixes branch. The devel branch looks to be for 2.0 (and the file I fixed isn't in there anymore).

### DIFF
--- a/lib/ansible/utils/__init__.py
+++ b/lib/ansible/utils/__init__.py
@@ -22,6 +22,7 @@ import os
 import shlex
 import yaml
 import copy
+import codecs
 import optparse
 import operator
 from ansible import errors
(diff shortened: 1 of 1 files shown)


abadger:
In #10630 we knew that we would only deal with json which mush be encoded in utf-8.  Here I'm not sure that we have the same guarantee.  So the fix should be similar in that we convert buf to a byte string when we write it to the file but we should be explicit about it rather than using codecs so that we can control what happens if non-utf-8 compatible data is given.

--tree seems to be about writing the output to a file for later review by a human.  With that in mind, I think using `to_bytes(buf, encoding='utf-8', errors='replace')` is the right thing to do.

abadger:
I'll work up an alternative to this patch based on that strategy.

abadger:
Okay, and that commit should also fix this in devel.

# Closing This Ticket

Hi!

issue is resolved.

Thank you!

PR #12490, opened 2015-09-23
Send initial data before calling select whenever possible

amenonsen:
Without this, we could execute «ssh -q ...» and call select(), which
would timeout after the default 10s, and only then send initial data.
(This is a relic of the earlier change where we always ran ssh with
-vvv, so the situation where it would sit quietly never happened in
practice; but this would have been the right thing to do even then.)


jimi-c:
Merged, thanks!

PR #12492, opened 2015-09-23
allow multiple vCenter/ESXi in single inventory

aioue:
Allow multiple vCenter/ESXi server to be scanned from single inventory/ini combo.


aioue:
Closing. [Fixed in updated vmware inventory](https://github.com/ansible/ansible/pull/15967).

PR #12493, opened 2015-09-23
Fix typo in checking select results

amenonsen:
It's possible for more than one fd to be set, so 'elif' is obviously not
the right thing to use.


jimi-c:
Merged, thanks!

PR #12502, opened 2015-09-24
Python 3: make test_variable_manager_precedence pass

mgedmin:
I'm going through failing unit tests one by one and making them pass.
This PR fixes test_variable_manager_precedence from
test/units/vars/test_variable_manager.py, which was failing due to
several unrelated issues:
- shlex.split() cannot deal with bytes on Python 3
- ansible.inventory.host.Host needs to be (correctly) hashable
- one count of dict.iteritems()
- one count of 'unicode' in a module I thought I'd already
  Python-3-ified (this is why I focus on making the tests pass on Python
  3, so I can submit a PR to enable them in the default tox.ini/Travis
  run).

### DIFF
--- a/lib/ansible/inventory/host.py
+++ b/lib/ansible/inventory/host.py
@@ -41,6 +41,9 @@ class Host:
     def __ne__(self, other):
         return not self.__eq__(other)
 
+    def __hash__(self):
+        return hash(self.name)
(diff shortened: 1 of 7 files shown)


jimi-c:
Merged, thanks!

PR #12506, opened 2015-09-24
Add Weekday (0-6) as a number and add weeknumber (00-52)

hyperized:
I would love to be able to numerically use weekdays and weeknumbers to be known as facts to Ansible for backup purposes.


jimi-c:
Merged, thanks!

PR #12510, opened 2015-09-24
Initial add of logentries callback plugin

jcftang:
This callback plugin will generate json objects to be sent to the
logentries service for auditing/debugging purposes.

To use:

Add this to your ansible.cfg file in the defaults block

```
[defaults]
callback_plugins = ./callback_plugins
callback_stdout = logentries
callback_whitelist = logentries
```

Copy the callback plugin into the callback_plugings directory

Either set the environment variables

```
export LOGENTRIES_API=data.logentries.com
export LOGENTRIES_PORT=10000
export LOGENTRIES_ANSIBLE_TOKEN=dd21fc88-f00a-43ff-b977-e3a4233c53af
```

Or create a logentries.ini config file that sites next to the plugin with the following contents

```
[logentries]
api = data.logentries.com
port = 10000
tls_port = 20000
use_tls = no
token = dd21fc88-f00a-43ff-b977-e3a4233c53af
```

### DIFF
--- a/lib/ansible/plugins/callback/logentries.py
+++ b/lib/ansible/plugins/callback/logentries.py
@@ -1,6 +1,19 @@
-"""
-(c) 2015, Logentries.com
-Author: Jimmy Tang <jimmy.tang@logentries.com>
+""" (c) 2015, Logentries.com, Jimmy Tang <jimmy.tang@logentries.com>
+
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 
 This callback plugin will generate json objects to be sent to logentries
 for auditing/debugging purposes.
@@ -18,7 +31,7 @@ Add this to your ansible.cfg file in the defaults block
     callback_stdout = logentries
     callback_whitelist = logentries
 
-Copy the callback plugin into the callback_plugings directory
+Copy the callback plugin into the callback_plugins directory
 
 Either set the environment variables
 
@@ -34,17 +47,16 @@ Or create a logentries.ini config file that sites next to the plugin with the fo
     tls_port = 20000
     use_tls = no
     token = dd21fc88-f00a-43ff-b977-e3a4233c53af
+    flatten = False
 
 
 """
 
 import os
-import threading
 import socket
 import random
 import time
 import codecs
-import Queue
 import ConfigParser
 import uuid
 try:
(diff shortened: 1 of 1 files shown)


lib/ansible/modules/core
```diff
@@ -1 +1 @@
-Subproject commit d659c79db7335a7a81594f7d977cb6bb67e9c5a3
+Subproject commit a4ac171a070551f1ed17a9c09b41d2481ad779af
```
bcoca: need to redo w/o the submodule updates

lib/ansible/plugins/callback/logentries.py
```diff
@@ -0,0 +1,336 @@
+"""
+(c) 2015, Logentries.com
+Author: Jimmy Tang <jimmy.tang@logentries.com>
+
```
bcoca: needs gpl license

lib/ansible/plugins/callback/logentries.py
```diff
@@ -0,0 +1,336 @@
...
+    port = 10000
+    tls_port = 20000
+    use_tls = no
+    token = dd21fc88-f00a-43ff-b977-e3a4233c53af
+
+
+"""
+
+import os
+import threading
```
bcoca: threading is pretty useless in this context, the callback gets called in it's own sub process

lib/ansible/plugins/callback/logentries.py
```diff
@@ -0,0 +1,336 @@
...
+                        continue
+                    break
+        except KeyboardInterrupt:
+            if self.verbose:
+                print("Logentries asynchronous socket client interrupted")
+
+        self.close_connection()
+
+
+try:
```
bcoca, line 164: ~~ use the shared open_url function, it takes care of many issues with python's ssl ~~

lib/ansible/plugins/callback/logentries.py
jcftang: Is the subprocess long lived (for the duration of the play). The code was lifted from an existing library that we have and we found it's a bit more reliable to have a queue to buffer the log lines before it's sent the logentries end points

lib/ansible/plugins/callback/logentries.py
jcftang: Will update this once I check the my line manager

lib/ansible/plugins/callback/logentries.py
jcftang, line 164: I had a look at the open_url function, it appears to be for http/https based connections only, the connection that this callback opens up is a plain tcp connection.

bcoca, line 164: ignore it then,  I stopped reading at import ssl, did not realize it is an encrypted tcp socket connection and assumed http/s

lib/ansible/plugins/callback/logentries.py
bcoca: this process does live for the lentgh of the play, but it is also not thread safe, we update variables that will end up being shared in the threads.

jcftang: Ok, I will work on removing the threading then.

jcftang:
I've removed the threading and queues, added a license as required.

I've also cleaned up a bit with wrapping up the emit method.

lib/ansible/plugins/callback/logentries.py
```diff
@@ -0,0 +1,329 @@
...
+
+import os
+import socket
+import random
+import time
+import codecs
+import ConfigParser
+import uuid
+try:
+    import certifi
```
bcoca, line 66: import errors should not be fatal, look at hipchat plugin and how it handles this, the plugin will not work but will not stop play while it issues warnings about it.

lib/ansible/plugins/callback/logentries.py
```diff
@@ -0,0 +1,329 @@
...
+            self._conn = sock
+
+    SocketAppender = TLSSocketAppender
+
+
+class CallbackModule(CallbackBase):
+    CALLBACK_VERSION = 2.0
+    CALLBACK_TYPE = 'notification'
+    CALLBACK_NAME = 'logentries'
+
```
bcoca, line 196: add 
 CALLBACK_NEEDS_WHITELIST = True

this is recent addition, it keeps plugins shipped with ansible from being executed automatically unless whitelisted through config.

lib/ansible/plugins/callback/logentries.py
jcftang, line 66: Sounds reasonable, thats changed in the following commits

PR #12518, opened 2015-09-24
Windows: Added parameter validation to the get-attr function

trondhindenes:
Improved get-attr so that it also enables parameter validation (this will implement a standardizer way of validating params instead of each module having to implement it itself)

Example: If a module requires the option "state" to be either "present" or "absent" this can be expressed like:
`Get-attr -obj $params -name "State" -default "Present" -ValidateSet "Present","Absent" -resultobj $resultobj -failifempty $true`
This will fail calling `fail-json` automatically if the option is anything other than "present" or "absent" and should save module devs some coding.

### DIFF
--- a/lib/ansible/module_utils/powershell.ps1
+++ b/lib/ansible/module_utils/powershell.ps1
@@ -101,11 +101,11 @@ Function Fail-Json($obj, $message = $null)
 # Helper function to get an "attribute" from a psobject instance in powershell.
 # This is a convenience to make getting Members from an object easier and
 # slightly more pythonic
-# Example: $attr = Get-Attr $response "code" -default "1"
-#Get-attr also supports Parameter validation to save you from coding that manually:
-#Example: Get-attr -obj $params -name "State" -default "Present" -ValidateSet "Present","Absent" -resultobj $resultobj -failifempty $true
+# Example: $attr = Get-AnsibleParam $response "code" -default "1"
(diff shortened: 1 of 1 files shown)


jhawkesworth:
This is a great idea.
I'm wondering if if Get-attr is still the best name for the function, its really getting an ansible module parameter, although Get-AnsibleModuleParameter is too long.
Would be great to add a bit of info about it in the windows module checklist too: http://docs.ansible.com/ansible/developing_modules.html#windows-modules-checklist

trondhindenes:
Agree. Maybe "Get-AnsibleParam"?

jhawkesworth:
Yeah, I toyed with Get-ModuleParam (1 less character) but Module could be misunderstood as a powershell module, so I think Get-AnsibleParam is best.

trondhindenes:
Updated the function, looks like something happened in the diff. Might be my Windows-y line endings.

jhawkesworth:
diff (as displayed on github) looks good to me.

Would be good to get this in before I tackle updating my extras modules for strict as I can see it saving a few lines of code.

trondhindenes:
Yup, managed to get it cleaned up. Should be ready for merging.

bcoca:
looks good, can flatten the commits?

trondhindenes:
sorry, messed up my commits. Moved to #12578

PR #12523, opened 2015-09-25
Merge properties instead of reading only the first cfg file found

pgmillon:
Use case :

I have a personal ansible.cfg where I configure some optims (ssh_args = -o ControlPersist=30m and pipelining = True); there is an ansible.cfg in my repo that configures some structural patterns (roles_path=public/roles:private/roles).

For such setup, I would have to copy my personal configuration on the cfg of the project or create an environment variable with all the values for the cfg files.

If settings were overridden respecting the order of precedence that is describe in the doc, I can have settings splitted between files and override only what I need.

### DIFF
--- a/lib/ansible/constants.py
+++ b/lib/ansible/constants.py
@@ -62,12 +62,12 @@ def load_config_file():
 
     p = ConfigParser.ConfigParser()
 
-    path0 = os.getenv("ANSIBLE_CONFIG", None)
-    if path0 is not None:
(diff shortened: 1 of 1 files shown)


chrismeyersfsu:
Love the idea. Seem like some unit tests would get this pushed along.

abadger:
This is a new feature request so it won't be applied to stable-1.9.  If you'd like to rebase it against devel we can take a look at it.

Issue #12497, opened 2015-09-23
powershell.ps1 strictmode breaks a ton of modules


trondhindenes:
I noticed set-strictmode got turned on in a recent commit.

This breaks a ton of modules both in core and extra. Strictmode throws errors left and right and requires a very "correct" language implementation. My recommendation is that we turn this off for now. If we want to enable strictmode later we need a better testing harness for powershell modules so that we can test if each module can run with strictmode.

I'm not against turning on strictmode down the line, but now is definetely not the right time.

(ping @cchurch )

trondhindenes:
Created PR #12525

nitzmahone:
As a compromise to keep things working without destabilizing stuff too close to 2.0 release, I went through and disabled strict mode on individual modules that would clearly be broken by it (only in 2.0-stable branch). I'll be doing some more testing, and will mark any non-obvious failures the same way. Most of the fixes are trivial (modules not using Get-Attr or Get-Ansible-Param for parameter presence testing is the most common issue).

I'll try to get around to fixing the obvious ones in devel soon...

dreamcat4:
Example how to temp. disable strictmode (per-module):

https://github.com/ansible/ansible-modules-extras/blob/40acb516fd5b344bb832758fc660a1d04a6b14d3/windows/win_iis_webbinding.ps1#L24-L25

nitzmahone:
I _think_ we've got all these fixed as of 2.1.1, closing...

PR #12525, opened 2015-09-25
temporarily disable strictmode

trondhindenes:
Strictmode introdoces a number of errors in various modules, for instance modules which are testing if an argument exists outside of get-attr. We need to (possibly temporarily) disable strictmode checking.

### DIFF
--- a/lib/ansible/module_utils/powershell.ps1
+++ b/lib/ansible/module_utils/powershell.ps1
@@ -26,7 +26,8 @@
 # USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #
 
-Set-StrictMode -Version Latest
+#Temporarily removed strictmode as it breaks a lot of modules
+#Set-StrictMode -Version Latest
 
 # Ansible v2 will insert the module arguments below as a string containing
 # JSON; assign them to an environment variable and redefine $args so existing


jimi-c:
@trondhindenes if absolutely required, I would prefer this to be an option but really I would prefer to have the modules fixed.

jimi-c:
cc @cchurch

trondhindenes:
I honestly don't see the point in enabling strictmode. In my 8+ years of powershell I've never seen it used, so I wonder what the rationaly was for enabling it in the first place. If we are to enable it it will require a major refactoring of lots of modules, so just enabling it out of the blue seems a bit premature. I propose we disable it and discuss it in the groups. Until we do anything, the current devel head is very broken on Windows.

jimi-c:
The problem was that errors were silently being ignored and it would move on rather than error out, leading to weird things happening. People wrote PERL scripts for years too without using strict mode there and it was always a crap shoot as to whether you'd run across some weird bug because of it. Reading this https://technet.microsoft.com/en-us/library/hh849692.aspx and what it allows when it's disabled leads me to believe that we should absolutely have this enabled.

bcoca:
if a script really needs it, script mode can be turned off for that run in
the script itself, but we should really fix all current windows modules to
work under strict mode.

## 

Brian Coca

trondhindenes:
I absolutely agree that we should discuss this. I'm just surprised that it got turned on just like that. My vote is still: We discuss it, come to a conclusion, make the necessary changes in modules, and then implement it (if that's what's decided on).

trondhindenes:
Anyone else care to chime in on this? As far as Windows is concerned the devel branch has been broken for a week now.

jimi-c:
@trondhindenes per our discussion, please submit PRs for any modules which have seriously broken implementations, so we can disable StrictMode on a per-module basis rather than globally. I'll go ahead and close this in the meantime, as I believe we're in agreement that StrictMode is something we should generally be using, and the core set of win_ modules for which we have integration tests are passing.

trondhindenes:
Noted. For the record, I'm not in agreement.

nitzmahone:
See discussion on #12497 - we'll compromise by temporarily disabling strict mode on an individual module basis (done in 2.0-stable only). Modules in devel still need real fixes- I'll try to get those in ASAP.

dreamcat4:
Example how to temp. disable strictmode:

https://github.com/ansible/ansible-modules-extras/blob/40acb516fd5b344bb832758fc660a1d04a6b14d3/windows/win_iis_webbinding.ps1#L24-L25

PR #12140, opened 2015-08-28
 Add no_syslog option in order to prevent ansible syslog its actions on the remote host (rebase  #11502)

lberruti:
Add no_syslog option in order to prevent ansible syslog its actions
on the remote host.

Syslog on remote hosts can be turned off by:

```
no_syslog option inside ansible.cfg.
-n --no_syslog command line options (ansible and ansible-playbook).
```

Rebased version of #11502

PR #12528, opened 2015-09-26
task logging revamp

bcoca:
- allow global no_log setting, no need to set at play or task level, but can be overriden by them
- allow turning off syslog only on task execution from target host (manage_syslog), overlaps with no_log functionality
- created log function for task modules to use, now we can remove all syslog references, will use systemd journal if present
- added debug flag to modules, so they can make it call new log function conditionally
- added debug logging in module's run_command

### DIFF
--- a/examples/ansible.cfg
+++ b/examples/ansible.cfg
@@ -182,7 +182,7 @@ fact_caching = memory
 #no_log = False
 
 # prevents logging of tasks, but only on the targets, data is still logged on the master/controller
-#managed_syslog = True
+#no_target_syslog = True
 
 [privilege_escalation]
 #become=True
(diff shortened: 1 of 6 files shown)


amenonsen:
I like this change, but I find the name `manage_syslog` confusing. For consistency with `no_log`, I would prefer `no_syslog`, or perhaps `no_target_syslog`. "Manage" is a very broad scope for something that only means "turn it off".

How does the debug thing work? Sorry, it's not clear to me from a quick look.

bcoca:
currently to turn on module debug, just put debug=1 in ansible.cfg, I like no_target_syslog, changing to that

lberruti:
Hi @bcoca, what about PR #12140?
#12140 adds also command line switch and fixes syslog on async tasks.

bcoca:
@lberruti i used #12140 as a model for this PR, merged the new setting with no_log as we had discussed in  #11502 and avoided having a 2nd very similar option to no_log for the rest.

async will be added if the current PR is accepted, probably very similar to what you did in those PRs.

PR #12534, opened 2015-09-26
Updated Azure inventory script to reflect changes in azure python package

vidbina:
##### Issue Type:
- Bug Report
- Bugfix Pull Request
##### Ansible Version:

ansible 1.9.3

The [devel/contrib/inventory/windows_azure.py](https://github.com/ansible/ansible/blob/devel/contrib/inventory/windows_azure.py) file used was the latest version from the `devel` branch.
##### Ansible Configuration:

N/A
##### Environment:

The `azure` Python package has been installed through `pip install azure` and is at version 1.0.1.
##### Summary:

The `WindowsAzureError` exception has been changed to an `AzureException` child, named`AzureHttpError` (see [changelog](https://github.com/Azure/azure-sdk-for-python/blob/fa3119c497e98343c7f789e07d108d1d2d457bd2/ChangeLog.txt#L47)). This fix gets the Windows Azure inventory file up to speed on that change.
##### Steps To Reproduce:

Any call that attempts to run the python script breaks, the simplest version being `python contrib/inventory/windows_azure.py`.
##### Expected Results:

Expected result to `python contrib/inventory/windows_azure.py` is a JSON payload.
##### Actual Results:

Actual result is the following output:

```
failed=True msg='`azure` library required for this script'
```

### DIFF
--- a/contrib/inventory/windows_azure.py
+++ b/contrib/inventory/windows_azure.py
@@ -48,7 +48,7 @@ except ImportError:
 
 try:
     import azure
-    from azure import WindowsAzureError
+    from azure.common import AzureException, AzureHttpError
     from azure.servicemanagement import ServiceManagementService
 except ImportError as e:
     print("failed=True msg='`azure` library required for this script'")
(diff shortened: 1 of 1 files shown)


bcoca:
the change is not backwards compatible, what happens to those still using the older sdk ?

PR #12535, opened 2015-09-26
bugfixes (uncallable things, statement without effect and indentation issues)

pheanex:


bcoca:
these are mostly style changes, we don't normally take these PRs as they break attribution.

pheanex:
@bcoca if you like i can take out the "semicolon-changes", or do you really dont care about the bugfixes?

bcoca:
i see a few whitespace changes also, can you submit PR with any bugfixes and examples that hit the bug?

pheanex:
- The whitespace changes are just indentation issues (where code inspector mentioned there were inconsistent indentations (spaces and tabs). 
- It also complained, that close was a method not a variable ;-)
- Most importantly: The statements without effects (== vs =) <-- Those should be real bugs

I am sorry but i really don't want to have to create scenarios where this would be desastrous. These were just some things i stumbled over and thought you guys might like/appreciate it => take it or leave it ;-)

amenonsen:
The whitespace changes and semicolon changes are in their own commits, but even if they're dropped, the other three are very clearly bugfixes (`.close` → `.close()`, `==` → `=`), and should be applied.

bcoca:
cherry picked d19ac086f17eaa395d8f6deb0dbd515531de4027 and b7d40fb523168b119e50e4c1529cc6a66d8ea525

PR #12539, opened 2015-09-28
Add node private & public ips to gce groups

dohoangkhiem:
##### SUMMARY
Allow specifying gce hosts by node private or public ip addresses

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
contrib/inventory/gce.py

##### ANSIBLE VERSION
2.3

### DIFF
--- a/contrib/inventory/gce.py
+++ b/contrib/inventory/gce.py
@@ -282,6 +282,13 @@ class GceInventory(object):
             if groups.has_key(stat): groups[stat].append(name)
             else: groups[stat] = [name]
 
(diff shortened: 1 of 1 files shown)


bcoca:
it would be better to add this as an option controlled by the config file

calfonso:
@erjohnso would you mind reviewing this?

erjohnso:
Even though this is from 2015 (Yikes!), I think it's probably fine to add. It should definitely be rebased though given the age.

PR #12546, opened 2015-09-28
A couple of ssh cleanups

amenonsen:
These don't change any functionality, just move some peripheral code into its own methods.


abadger:
Thanks @amenonsen   Hopefully my moving around  of the methods won't make you cringe too much ;-)

PR #12547, opened 2015-09-28
Add key split to subtags in ec2.py autodiscover script

jnerin:
Add the option to create subgroups based on splitting the value of the
key tags by whitespace.

 AWS Tag & Value:
  "private:app" = "app01 app02 app03 app04 app05     app06"

 ec2.py groups:
  "tag_private_app_app01": [
    "172..."
  ],
  "tag_private_app_app01_app02_app03_app04_app05_____app06": [
    "172..."
  ],
  "tag_private_app_app02": [
    "172..."
  ],
  "tag_private_app_app03": [
    "172..."
  ],

The idea is to be able to have more fine grained groups like:
- private:app = "list of apps/services the host is running like dns web db cache"
- private:env = "production/staging/testing"

And be able to combine it like:
 ansible-playbook --limit 'tag_private_app_cache:&tag_private_env_staging'

### DIFF
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -90,6 +90,41 @@ In addition, if an instance has AWS Tags associated with it, each tag is a new
 variable named:
  - ec2_tag_[Key] = [Value]
 
+And if the Value has spaces it's split by whitespace in multiple tag groups
+like this:
+ AWS Tag & Value:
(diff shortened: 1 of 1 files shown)


jnerin:
A very similar feature was merged few days later at https://github.com/ansible/ansible/commit/e6a9b2cae5628aa644a0290adb9c10140a96422c

PR #10008, opened 2015-01-15
Customize indent in YAML and JSON filters

jtyr:
This commit allows to customize indent in the `to_nice_yaml` and the `to_nice_json` filters.

PR #12555, opened 2015-09-29
Created new action plug-in "config_template"

cloudnull:
Created new config_template action plug-in

This commit adds a new action plug-in named "config_template". This
action plug-ins was created to allow overrides to be passed to
to standard templates written in "ini", "yaml", or "json" which are often 
used for system configuration.

This allows a project to carry a configuration template without having to 
create specific variables for all configuration items that could potentially 
be used. This keeps configuration templates clean and simple but 
enhances their ability to be used in different ways which is great for 
developers and deployers alike as no two deployment environments are 
ever the same.

Functional Overview:
  The action plug in functions much like the regular template module
  allowing an Ansible user to write the template containing dynamic
  data through the use of jinja variables. The source template is
  rendered in transit and passed through a particular configuration
  option method should any override data be present. The override 
  data is presented to the module through a dictionary (hash). Once 
  the overrides have been applied the template it's rendered once again 
  to ensure all jinja variables are appropriately rendered which gives 
  the user the ability to use variables in the override data as well as
  the base template if they so choose. Finally, the resultant is passed
  through the copy module where it is transferred to disk.

Signed-off-by: Kevin Carter kevin.carter@rackspace.com

### DIFF
--- a/lib/ansible/plugins/action/config_template.py
+++ b/lib/ansible/plugins/action/config_template.py
@@ -1,0 +1,570 @@
+# (c) 2015, Kevin Carter <kevin.carter@rackspace.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+try:
+    import ConfigParser
+except ImportError:
+    import configparser as ConfigParser
+
+import datetime
+import io
+import json
+import os
+import pwd
+import re
+import time
+import yaml
+
+
+from ansible.plugins.action import ActionBase
+from ansible.utils.unicode import to_bytes, to_unicode
+from ansible import constants as C
+from ansible import errors
+
+
+CONFIG_TYPES = {
+    'ini': 'return_config_overrides_ini',
+    'json': 'return_config_overrides_json',
+    'yaml': 'return_config_overrides_yaml'
+}
+
+
+def _convert_2_string(item):
+    """Return byte strings for all items.
+
+    This will convert everything within a dict, list or unicode string such
+    that the values will be encode('utf-8') where applicable.
+    """
+
+    if isinstance(item, dict):
+        # Old style dict comprehension for legacy python support
+        return dict(
+            (_convert_2_string(key), _convert_2_string(value))
+            for key, value in item.iteritems()
+        )
+    elif isinstance(item, list):
+        return [_convert_2_string(i) for i in item]
+    elif isinstance(item, tuple):
+        return tuple([_convert_2_string(i) for i in item])
+    elif isinstance(item, set):
+        return item
+    else:
+        try:
+            return item.encode('utf-8')
+        except AttributeError:
+            return str(item)
+
+
+class MultiKeyDict(dict):
+    """Dictionary class which supports duplicate keys.
+    This class allows for an item to be added into a standard python dictionary
+    however if a key is created more than once the dictionary will convert the
+    singular value to a python set. This set type forces all values to be a
+    string.
+    Example Usage:
+    >>> z = MultiKeyDict()
+    >>> z['a'] = 1
+    >>> z['b'] = ['a', 'b', 'c']
+    >>> z['c'] = {'a': 1}
+    >>> print(z)
+    ... {'a': 1, 'b': ['a', 'b', 'c'], 'c': {'a': 1}}
+    >>> z['a'] = 2
+    >>> print(z)
+    ... {'a': set(['1', '2']), 'c': {'a': 1}, 'b': ['a', 'b', 'c']}
+    """
+    def __setitem__(self, key, value):
+        if key in self:
+            if isinstance(self[key], set):
+                items = self[key]
+                items.add(str(value))
+                super(MultiKeyDict, self).__setitem__(key, items)
+            else:
+                items = [str(value), str(self[key])]
+                super(MultiKeyDict, self).__setitem__(key, set(items))
+        else:
+            return dict.__setitem__(self, key, value)
+
+
+class ConfigTemplateParser(ConfigParser.RawConfigParser):
+    """ConfigParser which supports multi key value.
+    The parser will use keys with multiple variables in a set as a multiple
+    key value within a configuration file.
+    Default Configuration file:
+    [DEFAULT]
+    things =
+        url1
+        url2
+        url3
+    other = 1,2,3
+    [section1]
+    key = var1
+    key = var2
+    key = var3
+    Example Usage:
+    >>> cp = ConfigTemplateParser(dict_type=MultiKeyDict)
+    >>> cp.read('/tmp/test.ini')
+    ... ['/tmp/test.ini']
+    >>> cp.get('DEFAULT', 'things')
+    ... \nurl1\nurl2\nurl3
+    >>> cp.get('DEFAULT', 'other')
+    ... '1,2,3'
+    >>> cp.set('DEFAULT', 'key1', 'var1')
+    >>> cp.get('DEFAULT', 'key1')
+    ... 'var1'
+    >>> cp.get('section1', 'key')
+    ... {'var1', 'var2', 'var3'}
+    >>> cp.set('section1', 'key', 'var4')
+    >>> cp.get('section1', 'key')
+    ... {'var1', 'var2', 'var3', 'var4'}
+    >>> with open('/tmp/test2.ini', 'w') as f:
+    ...     cp.write(f)
+    Output file:
+    [DEFAULT]
+    things =
+        url1
+        url2
+        url3
+    key1 = var1
+    other = 1,2,3
+    [section1]
+    key = var4
+    key = var1
+    key = var3
+    key = var2
+    """
+    def _write(self, fp, section, item, entry):
+        if section:
+            if (item is not None) or (self._optcre == self.OPTCRE):
+                fp.write(entry)
+        else:
+            fp.write(entry)
+
+    def _write_check(self, fp, key, value, section=False):
+        if isinstance(value, set):
+            for item in value:
+                item = str(item).replace('\n', '\n\t')
+                entry = "%s = %s\n" % (key, item)
+                self._write(fp, section, item, entry)
+        else:
+            if isinstance(value, list):
+                _value = [str(i.replace('\n', '\n\t')) for i in value]
+                entry = '%s = %s\n' % (key, ','.join(_value))
+            else:
+                entry = '%s = %s\n' % (key, str(value).replace('\n', '\n\t'))
+            self._write(fp, section, value, entry)
+
+    def write(self, fp):
+        if self._defaults:
+            fp.write("[%s]\n" % 'DEFAULT')
+            for key, value in self._defaults.items():
+                self._write_check(fp, key=key, value=value)
+            else:
+                fp.write("\n")
+
+        for section in self._sections:
+            fp.write("[%s]\n" % section)
+            for key, value in self._sections[section].items():
+                self._write_check(fp, key=key, value=value, section=True)
+            else:
+                fp.write("\n")
+
+    def _read(self, fp, fpname):
+        cursect = None
+        optname = None
+        lineno = 0
+        e = None
+        while True:
+            line = fp.readline()
+            if not line:
+                break
+            lineno += 1
+            if line.strip() == '' or line[0] in '#;':
+                continue
+            if line.split(None, 1)[0].lower() == 'rem' and line[0] in "rR":
+                continue
+            if line[0].isspace() and cursect is not None and optname:
+                value = line.strip()
+                if value:
+                    if isinstance(cursect[optname], set):
+                        _temp_item = list(cursect[optname])
+                        del cursect[optname]
+                        cursect[optname] = _temp_item
+                    elif isinstance(cursect[optname], (str, unicode)):
+                        _temp_item = [cursect[optname]]
+                        del cursect[optname]
+                        cursect[optname] = _temp_item
+                    cursect[optname].append(value)
+            else:
+                mo = self.SECTCRE.match(line)
+                if mo:
+                    sectname = mo.group('header')
+                    if sectname in self._sections:
+                        cursect = self._sections[sectname]
+                    elif sectname == 'DEFAULT':
+                        cursect = self._defaults
+                    else:
+                        cursect = self._dict()
+                        self._sections[sectname] = cursect
+                    optname = None
+                elif cursect is None:
+                    raise ConfigParser.MissingSectionHeaderError(
+                        fpname,
+                        lineno,
+                        line
+                    )
+                else:
+                    mo = self._optcre.match(line)
+                    if mo:
+                        optname, vi, optval = mo.group('option', 'vi', 'value')
+                        optname = self.optionxform(optname.rstrip())
+                        if optval is not None:
+                            if vi in ('=', ':') and ';' in optval:
+                                pos = optval.find(';')
+                                if pos != -1 and optval[pos - 1].isspace():
+                                    optval = optval[:pos]
+                            optval = optval.strip()
+                            if optval == '""':
+                                optval = ''
+                        cursect[optname] = optval
+                    else:
+                        if not e:
+                            e = ConfigParser.ParsingError(fpname)
+                        e.append(lineno, repr(line))
+        if e:
+            raise e
+        all_sections = [self._defaults]
+        all_sections.extend(self._sections.values())
+        for options in all_sections:
+            for name, val in options.items():
+                if isinstance(val, list):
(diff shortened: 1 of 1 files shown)


odyssey4me:
:+1: This action plugin is proving useful for https://github.com/openstack/openstack-ansible as it provides a simple way to implement updates to template files without having to individually itemise every setting in the template/role.

robynbergeron:
Pinging other OpenStack module folks for their FYI and transparency and awareness and so forth.

@emonty @shrews @dguerri @juliakreger @j2sol @rcarrillocruz

ghost:
+1 for this. It is a useful module and a way to do dynamic configs in a very Ansiblely way.

For note to all reviews, this module actually has nothing to do with OpenStack at all. The team submitting it work with OpenStack, but that is where the connection to OpenStack ends. This should be evaluated on its merrit alone without connection to OpenStack since it will be useful to quite a few projects.

cloudnull:
@SamYaple  I've updated the syntax on the one line, rebased the PR off of the head of the devel branch, and retested the module. Thank you very much for the review.

cloudnull:
The latest change set was pushed to pass the boilerplate checks now being run within devel.

omgjlk:
Can you toss together a sample usage of this to help us understand how it fits together?

cloudnull:
@j2sol I've put together a quick little repo which runs a functional test for the action plugin: [ https://github.com/cloudnull/testing_config_template ]. The test play will run stand alone when Ansible 2.0 (devel) is installed. 

As for a more real world example, the OpenStack-Ansible project is using a version of this plugin with Ansible 1.9.x [ https://github.com/openstack/openstack-ansible/blob/master/playbooks/plugins/actions/config_template.py ] and all of the relevant template tasks have been updated to use the plugin; example: [ https://github.com/openstack/openstack-ansible/blob/master/playbooks/roles/os_keystone/tasks/keystone_post_install.yml#L16-L24 ].

Please let me know if you have any questions.

spotzz:
+1, code looks good

leseb:
+1 :+1: good and clean, validated on ceph-ansible as well.

jimmymccrory:
+1, great module. Very useful for avoiding having to define every possible config option in a template

cloudnull:
This was rebased to ensure compatibility with a more recent devel branch and updated the plugin to pull in recent changes that have made it into the 1.9.4 module found here: https://github.com/openstack/openstack-ansible-plugins/blob/master/action/config_template.py

stevelle:
+1 
I have been using this actively for about 7 weeks now in daily work on multiple roles destined for Galaxy, and it simplifies complex service configuration greatly. I would love to have this available in a coming Ansible release to enable easier and wider adoption.

cloudnull:
Updated this PR to work with ansible 2.1 as tested with the head of the devel branch.

cloudnull:
Bump -- @robynbergeron @gregdek -- any chance we might be able to have some additional review on this PR? we're beginning to test with Ansible 2.1.0 (head of devel) in OpenStack-Ansible and it'd be great if this can be part of core as we move that direction (assuming it's accepted that is).

gregdek:
Sorry @cloudnull but we don't really deal with the ansible/ansible queue much. Lemme ping @jimi-c for some help here.

cloudnull:
rebased from the head of devel to stay up-to-date.

cloudnull:
rebased again it would seem the fedora tests were failing for reasons outside of this PR.

jimi-c:
@cloudnull sorry we didn't get this in for 2.1. I'm going to bump the milestone to 2.2 so we can review it for the next version.

cloudnull:
Thanks @jimi-c much appreciated.

cloudnull:
@jimi-c I just saw the updated [Ansible 2.2 Road Map](https://github.com/ansible/ansible/blob/devel/docsite/rst/roadmap/ROADMAP_2_2.rst) and I'm curious if it would be possible to get this module on the road map or does being part of the milestone suffice? Sorry to pester but we have quite a few folks pulling the module from the OpenStack-Ansible project's plugins repository and I'd love for folks to be able to leverage the module without everyone having to carrying a custom plugin. Regardless, I hope you're well and thanks again for the earlier update.

ghost:
Agree with @cloudnull . I know of several projects that are using this and would love to stop carrying it in their repo.

bcoca:
-1 on this.

This is clearly a template duplicate, by just adding 'data formats' into the output, which filters normally take care of (to_json/to_yaml) and we even have a to_ini proposed ( among other things in #10008). As for the 'overrides' they overlap with the combine and set theory filters. 

I don't really want to create redundant paths for doing the same thing easily accomplished with templates and filters.

ghost:
@bcoca it is pretty clear that this is based on the template builtin, but in practice it has been _very_ useful to projects like openstack/OpenStack-Ansible and ceph/ceph-ansible. It is by far the cleanest, most true to ansible way to extend these template config files for these projects.

Perhaps this functionality can be merged into template itself since the added code doesnt impair or change the behavior of template.

I would really hate to see this rejected from Ansible in favor of lots of filters/registers. It would add alot of unneeded ugliness in my opinion.

evrardjp:
@bcoca : I agree with @SamYaple . It has been useful in the past. If you think this should be done in template module (and therefore adapt it), fine for me too.

What we are looking for is not bypass another path to get something done. What we are looking for **is** getting things done. 

For the openstack-ansible case, It's infeasable (not technically but practically) to maintain every single configuration of every configuration file that openstack provides.

bcoca:
I see it hard to justify dumping this into template when most if it can easily be done in a few lines

```
- template: src=json.j2 dest=/file
```

json.j2

```
{{ myvar|to_json }}
```

or if you press me, it can be done all in 1 line (though i don't recommend this because of other reasons):

`- copy: content={{myvar|to_json}} dest=/file`

Adding combine or union to create the desired `myvar` seems simple enough, so I'm very recalcitrant to add so much code to support what is easily supported already with existing options in a very succinct manner.

sivel:
I am in agreement with @bcoca on this as well.

Your templates are also re-usable, by doing something like the following, that way you don't have to have a template per var:

```
- template: src=json.j2 dest=/file
  vars:
    myvar: "{{ some_other_var }}"
```

Then your templates always just use `myvar`.

odyssey4me:
@bcoca If there are alternative ways to achieve this goal, then we're totally open to those options. We'd all like to carry less code as long as they meet our needs.

The OpenStack-Ansible community built the config_template based on the need to be able to let our consumers defined arbitrary INI file entries using a single var without having to template out a pre-determined format (because the INI file could have many valid sections that change over time). As a project to manage our technical debt we wanted to ensure that we only have to define the key glue bits in the template, and leave any other optional settings to be implemented at the deployers discretion.

For example, if a deployer wishes to add the following settings into an INI file called nova.conf:

```
[DEFAULT]
remove_unused_original_minimum_age_seconds = 43200

[libvirt]
cpu_mode = host-model
disk_cachemodes = file=directsync,block=none

[database]
idle_timeout = 300
max_pool_size = 10
```

The configuration is very simply implemented by adding:

```
nova_nova_conf_overrides:
  DEFAULT:
    remove_unused_original_minimum_age_seconds: 43200
  libvirt:
    cpu_mode: host-model
    disk_cachemodes: file=directsync,block=none
  database:
    idle_timeout: 300
    max_pool_size: 10
```

Each of the sections may, or may not, be in the template. Each of the settings may, or may not, be in the template. If the setting does happen to be in the template then the user's settings merge over the top of the template's content.

This has been a very powerful mechanism which has allowed us to focus less on implementing specific tunable vars in the templates and instead to focus on broader needs. It has also given our consumers a lot of freedom to tune any setting they like without requiring a wait through a development cycle.

Is there another way to achieve this?

cloudnull:
Just to echo what others have said, this module provides the ability to add/modify nested sections, keys, or values into a rendered (or otherwise static) json, yaml, or ini file; the functionality is similar to that of a deep merge as found in other configuration management and deployment automation projects. This implementation is using a rendered template as the top level element and an “override” as that of the child which will hash merge if a change is made to the in-flight template. Part of the modules hash merge capabilities includes mutlistring options for ini files when a `set` data type is provided. The inclusion of multistring functionality is incredibly powerful especially when a typical delimiter based list may be impractical or impossible in an ini file. 

**_MultiString option example**_
Input data is provided using a yaml set --

``` yaml
section:
  option:
    ? things1
    ? things2
    ? things3
```

This is the rendered output --

``` ini
[section]
option = things1
option = things2
option = things3
```

---

Like mentioned earlier we built the module so we didn't have to carry the many 1000's of configuration options and types within our supported OpenStack projects as variables in template files or as redefine json/yaml data structures. Many of the templates found within our roles are provided by the upstream OpenStack project. This community integration gives our deployers a community vetted configuration file with an option to customize based on their needs in a clean and idempotent way. If these operators went the register > filter > generator route they would have to redefine entire data structures to add or modify entries that could be a simple matter of rendering a well defined template with a user provided entry point. Additionally I'm not sure how we'd allow for multistring options without creating and increasing user facing complexity and special case options using the variable > filter > generator methods mentioned in previous comments. Maybe @leseb can speak more to how the ceph-ansible project is using the module as I believe they too have a similar use case. 

I'm currently on holiday with extremely limited connectivity but when I return state side I'd be happy to attend meetings or participate in a more real time conversation regarding  this module if need be. Regardless I appreciate the consideration and I hope all is well.

leseb:
The way we use this extremely useful module in ceph-ansible is really similar to the OpenStack Ansible project. Ceph has more a thousand configuration options. We like to offer to our users the flexibility to change any configuration option of Ceph without having to consistently editing the template file and having to PR a change to add a new option. We also have multiple INI sections in our `ceph.conf` so this module works perfectly for us. It keeps a fairly limited, simple and not too verbose configuration file.

bcoca:
I understand what it does, I'm saying we already have other ways of doing it, not complete in the ini case, but easier to add that than duplicate soo much here. This can easily be done with 2 tasks with yaml/json and filters, in the ini case we are missing the filters, but we have the to_ini already in a proposed PR, we just need the 'from_ini' to have the same capabilities as already exist with json and yaml.

evrardjp:
@bcoca What do you mean easily be done with 2 tasks? I'd be happy to know how you would do that with the current tools. I'm happy to learn how to do it differently too.

The code you gave above doesn't solve the issue we're having. This modules looks up for the appropriate sections, and throws the output where it's needed. It's like patching some static file with a variable on multiple arbitrary locations. 

When you think about it, it's like a superpowered lineinfile (where you ensure the integrity of the whole file instead of ensuring the line only) or template module (with adaptability without having to think of all the possibilities of the template sections)

This module seems (at first sight) the most elegant way of achieving the goal.

sivel:
Using theoretical `to_ini` and `from_ini` filters, this can be achieved like:

```
- set_fact:
    new_ini: "{{ old_ini_contents | from_ini | combine(nova_nova_conf_overrides, recursive=True) | to_ini }}"
```

I threw together a `to_ini` and `from_ini` and tested this out, and it works as expected.

This could also be done with `from_json`/`to_json` or `from_yaml`/`to_yaml` or any combination really.

cloudnull:
@sivel I think we're missing the fact that all of the examples on how to do this alternatively are still being done with multiple tasks for a single configuration file using incomplete and theoretical filters whereas this module is feature complete and currently working in production. The OSA project, and the few others currently using the `config_template` module, have a use-case looks that is like so [[0](https://github.com/openstack/openstack-ansible-os_neutron/blob/master/tasks/neutron_post_install.yml#L16-L51)]. We're dropping multiple, potentially very large, configuration files that would need to be sourced, parsed, loaded, munged, and copied to the target node. Based on the previous alternative proposals this would require doing a lookup of a static file, registering the stdout, munging, and copying with content. At a minimum that would be two tasks per file (if we did the filter munging within the content field of the copy module). While the single file + multiple task method may potentially solve some of a single/simple problem it would be hard, if not impossible, to use an iterator and perform the same operations on multiple target config files using different overrides. This would result in a ton of cumbersome duplicate tasks and would slow within large scale environments especially when compared to the run-time of this module.

I definitely understand the desire not to duplicate code but I believe the practicality of the module trumps purity especially when faced with adding a module to the Ansible internals which contains some duplicate code vs offloading a lot of task duplication onto a deployers roles or playbooks. 

All this said, If there's a better way to solve the [[0](https://github.com/openstack/openstack-ansible-os_neutron/blob/master/tasks/neutron_post_install.yml#L16-L51)] use-case that would be efficient at scale please let me know because I'd be happy to test pivot if there's a better way.

[0] - https://github.com/openstack/openstack-ansible-os_neutron/blob/master/tasks/neutron_post_install.yml#L16-L51

cloudnull:
As an aside, many of the configuration template files we use contain jinja logic within them. I don't know if the combination of lookup plugin + task filters would properly render the jinja content prior to applying the overrides without using temporary files rendered by the template module and re-writing the same file using the alternative multi-task method?

cloudnull:
I put together a quick test to see the difference in the methods using the tasks from within this repo: [ https://github.com/cloudnull/testing_config_template ]. In testing I found that the lookup plugin + copy module + filters was nearly 2x slower than the proposed config_template module. _Note the difference is minimal when comparing the two directly however it would make a big difference with hundreds of large configuration files across thousands of nodes_.

``` text
######################## USING CONFIG_TEMPLATE = test-ct.yml ########################
TASK: run config template ini ------------------------------------------- 0.18s
TASK: run config template yaml ------------------------------------------ 0.12s
TASK: run config template json ------------------------------------------ 0.11s

real    0m2.216s
user    0m1.890s
sys     0m0.187s
######################## USING CONFIG_TEMPLATE = test-ct.yml ########################

######################## USING LOOKUP + COPY = test-mt.yml ########################
TASK: run config template yaml ------------------------------------------ 0.32s
TASK: run config template json ------------------------------------------ 0.19s

real    0m2.305s
user    0m1.888s
sys     0m0.207s
######################## USING LOOKUP + COPY = test-mt.yml ########################
```
- Notice the two tests are different because there are currently no {to,from}_ini filters that would work to test this use-case. 
- The config_template module complete faster than the copy task method and did so with 3 operations for 3 different configuration file types whereas the copy task method only had two different  configuration files to render. 
- As an aside, setting an indentation value within the `to_nice_{json,yaml}` filters resulted in nothing when using json and a stacktrace when using yaml.

**_Error when setting the indentation level: `to_nice_yaml(indent=4)`**_

``` error
fatal: [XXXX]: FAILED! => {"failed": true, "msg": "Unexpected templating type error occurred on ({{ lookup('template', 'templates/test.yaml.j2', convert_data=False) | from_yaml | combine(config_overrides, recursive=True) | to_nice_yaml(indent=2) }}): dump() got multiple values for keyword argument 'indent'"}
```

### Rendered File results

Upon inspection of the rendered files it seems the recursive combine filter results in missing data.

**_From config_template module rendering JSON**_

``` json
{
    "DEFAULT": {
        "f": 22, 
        "x": 1, 
        "y": 2
    }, 
    "ah": "ha", 
    "default2": {
        "a": 1
    }, 
    "ensuring replacements": {
        "replacements": "this is a jina2 var"
    }, 
    "list_op": {
        "shucks": [
            8000, 
            8001, 
            9000, 
            9001, 
            9002
        ]
    }, 
    "other": {
        "blah": 9000
    }, 
    "section1": {
        "dict_item": {
            "test": 123
        }, 
        "iterator_item": [
            "thing1", 
            "thing2", 
            "thing3"
        ], 
        "key1": "s1 things", 
        "key2": "s2 this is a jina2 var replacement var"
    }, 
    "section2": {
        "key1": "s2 things", 
        "key2": "s2 this is a jina2 var replacement var"
    }
}
```

**_From copy task w/ lookup plugin + filters rendering JSON**_

``` json
{"DEFAULT": {"y": 2, "x": 1, "f": 22}, "ah": "ha", "other": {"blah": 9000}, "ensuring replacements": {"replacements": "this is a jina2 var"}, "list_op": {"shucks": [9000, 9001, 9002]}, "section2": {"key2": "s2 this is a jina2 var replacement var", "key1": "s2 things"}, "section1": {"key2": "s2 this is a jina2 var replacement var", "key1": "s1 things", "dict_item": {"test": 123}, "iterator_item": ["thing1", "thing2", "thing3"]}}
```

_Notice "to_nice_json" w/ ansible 2.1 was not rendering w/ indentations_

---

**_From config_template module rendering YAML**_

``` yaml
DEFAULT:
  f: 22
ah: ha
default:
  x: 1
  y: 2
default2:
  a: 1
ensuring replacements:
  replacements: this is a jina2 var
list_op:
  shucks:
  - 8000
  - 8001
  - 9000
  - 9001
  - 9002
other:
  blah: 9000
section1:
  dict_item:
    test: 123
  iterator_item:
  - thing1
  - thing2
  - thing3
  key1: s1 things
  key2: s2 this is a jina2 var replacement var
section2:
  key1: s2 things
  key2: s2 this is a jina2 var replacement var
```

**_From copy task w/ lookup plugin + filters rendering YAML**_

``` yaml
DEFAULT:
    f: 22
ah: ha
default:
    x: 1
    y: 2
ensuring replacements:
    replacements: this is a jina2 var
list_op:
    shucks:
    - 9000
    - 9001
    - 9002
other:
    blah: 9000
section1:
    dict_item:
        test: 123
    iterator_item:
    - thing1
    - thing2
    - thing3
    key1: s1 things
    key2: s2 this is a jina2 var replacement var
section2:
    key1: s2 things
    key2: s2 this is a jina2 var replacement var
```

If there's a way to optimize the tasks I'm using to test I'd be keen on retesting and further profiling.

cloudnull:
Closing this PR due to a lack of interest and traction. 

I'd like to say thank you to everyone who reviewed this module, I appreciate all of the feedback.  The OSA project will continue to carry the module in our plugins repository found here [ https://github.com/openstack/openstack-ansible-plugins ]. If the interest in the module changes we'd be happy to re-open this PR/update it.

leseb:
Sad to read this. ceph-ansible will keep on using this awesome module, thanks @cloudnull

PR #12561, opened 2015-09-29
Inject existing groups at InventoryDir initialization

srgvg:
This fixes a corner case where ini files live in a subdir
of the main inventory directory.

Reproducing the original error:

mkdir -p inventory/ini
cat > inventory/ini/hosts << EOF
[www]
www1
EOF

$ ansible -i inventory/ all -m ping
ERROR! 'all'

(or without the [www] group, it would complain about 'ungrouped')


jimi-c:
Merged, thanks!

PR #12568, opened 2015-09-30
Python 3: fix most tests for ansible.module_utils.basic

mgedmin:
- Fix ImportError: **builtin** during test module import
- Make most of the newly-enabled tests pass
- The remaining failure needs some deeper thought, so I'm postponing it
  for a later PR

This patchset doesn't reduce the number of failing tests (7), but it
increases the number of succeeding tests (from 260 to 282).

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -88,6 +88,15 @@ try:
 except NameError:
     bytes = str                     # Python 2
 
+try:
+    dict.iteritems
+except AttributeError:
+    def iteritems(d):               # Python 3
+        return d.items()
+else:
+    def iteritems(d):               # Python 2
+        return d.iteritems()
+
 
 try:
     import json
@@ -299,7 +308,7 @@ def json_dict_unicode_to_bytes(d, encoding='utf-8'):
     if isinstance(d, unicode):
         return d.encode(encoding)
     elif isinstance(d, dict):
-        return dict(imap(json_dict_unicode_to_bytes, d.iteritems(), repeat(encoding)))
+        return dict(imap(json_dict_unicode_to_bytes, iteritems(d), repeat(encoding)))
     elif isinstance(d, list):
         return list(imap(json_dict_unicode_to_bytes, d, repeat(encoding)))
     elif isinstance(d, tuple):
@@ -317,7 +326,7 @@ def json_dict_bytes_to_unicode(d, encoding='utf-8'):
     if isinstance(d, str):
         return unicode(d, encoding)
     elif isinstance(d, dict):
-        return dict(imap(json_dict_bytes_to_unicode, d.iteritems(), repeat(encoding)))
+        return dict(imap(json_dict_bytes_to_unicode, iteritems(d), repeat(encoding)))
     elif isinstance(d, list):
         return list(imap(json_dict_bytes_to_unicode, d, repeat(encoding)))
     elif isinstance(d, tuple):
@@ -414,7 +423,7 @@ class AnsibleModule(object):
         self.aliases = {}
 
         if add_file_common_args:
-            for k, v in FILE_COMMON_ARGUMENTS.iteritems():
+            for k, v in FILE_COMMON_ARGUMENTS.items():
                 if k not in self.argument_spec:
                     self.argument_spec[k] = v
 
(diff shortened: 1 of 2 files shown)


lib/ansible/module_utils/basic.py
```diff
@@ -397,7 +422,7 @@ def __init__(self, argument_spec, bypass_checks=False, no_log=False,
         self.aliases = {}
 
         if add_file_common_args:
-            for k, v in FILE_COMMON_ARGUMENTS.iteritems():
+            for k, v in iteritems(FILE_COMMON_ARGUMENTS):
                 if k not in self.argument_spec:
```
abadger: For all of these parameter iteritems calls it's probably better to just turn them into `dict.items()`.  So that's this call and the rest of the calls in this file.

lib/ansible/module_utils/basic.py
```diff
@@ -1255,7 +1280,7 @@ def _log_invocation(self):
             msg = 'Invoked'
 
         # 6655 - allow for accented characters
-        if isinstance(msg, unicode):
+        if not isinstance(msg, str):
             # We should never get here as msg should be type str, not unicode
             msg = msg.encode('utf-8')
```
abadger: - I wonder if we need to also catch things that are not text (for instance a number)?
- The previous code handled that further down with a format string.  We also wouldn't want to throw a traceback here on non-utf8 bytes.  Since this is logging we probably want to use errors='replace' to deal with that.
- I took a look at systemd.journal on fedora21.  (Note: these behaviours are unaffected by client locale settings.  Don't know if the systemd daemon has a settable locale setting).
  - in python2 it will take either bytes or unicode string.  If the string is not utf-8, systemd.journal registers it as a binary blob.
  - On python3 this must be a unicode string.
- Also took a look at syslog on fedora 21(these are unaffected by the client locale setting)
  - In python2 it must be a byte string.  A non-utf-8 string becomes the replacement character
  - In python3 this must be a text str type.

So this whole method needs to be reworked for python3 compatibility.

test/units/module_utils/test_basic.py
```diff
@@ -620,7 +619,7 @@ def test_module_utils_basic_ansible_module_set_context_if_different(self):
             
             with patch('selinux.lsetfilecon', return_value=0) as m:
                 self.assertEqual(am.set_context_if_different('/path/to/file', ['foo_u', 'foo_r', 'foo_t', 's0'], False), True)
-                m.assert_called_with('/path/to/file', 'sp_u:sp_r:sp_t:s0')
+                m.assert_called_with(b'/path/to/file', 'sp_u:sp_r:sp_t:s0')
 
         delattr(basic, 'selinux')
```
abadger, line 625: All the changes to  test/units/module_utils/test_basic.py look good.

lib/ansible/module_utils/basic.py
```diff
@@ -74,6 +74,31 @@
...
+try:
+    dict.iteritems
+except AttributeError:
+    def iteritems(d):               # Python 3
+        return d.items()
+else:
+    def iteritems(d):               # Python 2
+        return d.iteritems()
+
+
```
abadger: Cosmetic: I think our coding style is to comment on a separate line instead of at the end of a line.  At this indent level it also takes fewer chars which mpdehaan was conscious about in basic.py.

lib/ansible/module_utils/basic.py
mgedmin: Do you want me to revert this change, to keep the test failing as a reminder?  Full syslog/journal fixing appears to be complicated enough to warrant splitting out into a separate pull request.

abadger: Yeah, sounds like a plan for now.  If we can, it's better to fix it correctly than to do something incorrect and forget that it's still a problem.

If we get to the place where it's the only failing test left we can revisit whether a stopgap is appropriate so that other pieces of code don't regress.

mgedmin:
Rebased since there were merge conflicts.

mgedmin:
(BTW the syslog thing causes not 1 but 15 test failures, so the original PR description becomes somewhat inaccurate ;)

abadger:
Thanks @mgedmin !

lib/ansible/module_utils/basic.py
mgedmin: I filed #12586 to continue this discussion.

PR #12569, opened 2015-09-30
Bring groups handling in ini.py up-to-date (trivial fix)

amenonsen:
Since c8f2483d, ini.py expects to always be passed in a pre-created list
of groups, and can no longer deal sensibly with an empty list; this just
makes that expectation clear.


jimi-c:
Merged, thanks!

PR #12571, opened 2015-09-30
Fixes #12309

### LINKED ISSUE #12309: ansible-pull git options
Ansible Version: 1.9.2
Ansible Configuration: default
Environment: N/A

Summary:

Since 1.9 force=no is the default behaviour for git module and in case of using git from ansible playbooks this options is possible to change. However when running ansible-pull this option is impossible to change which may break the whole checkout on the machine running ansible-pull in case some file in the playbooks repo has a local change or smth else.
Could you add some option to ansible-pull that could be passed to the git module? Or make the git module read it's options from ansible.cfg.

Expected results:
ansible-pull is able to pass force=yes to the git module.

Actual results:
ansible-pull always uses git module with force=no which results in failing to run the playbook in case local modifications to the playbook repo exist.

w1r0x:
Additional options to ansible-pull git
see #12309

### DIFF
--- a/bin/ansible-pull
+++ b/bin/ansible-pull
@@ -140,6 +140,11 @@ def main(args):
                       help='adds the hostkey for the repo url if not already added')
     parser.add_option('--key-file', dest='key_file',
                       help="Pass '-i <key_file>' to the SSH arguments used by git.")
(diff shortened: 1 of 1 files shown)


jimi-c:
Merged, thanks!

PR #12572, opened 2015-09-30
Make ansible-vault view not write plaintext to a tempfile

amenonsen:
…when it can just feed the PAGER on stdin instead.

### DIFF
--- a/lib/ansible/cli/vault.py
+++ b/lib/ansible/cli/vault.py
@@ -138,7 +138,7 @@ class VaultCLI(CLI):
     def execute_view(self):
 
         for f in self.args:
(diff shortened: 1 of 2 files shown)


bcoca:
nice

bcoca:
you could also just feed self.pager() function in the cli

amenonsen:
@bcoca thanks for the suggestion, I didn't notice CLI's `pager()` method. I've pushed a revised version along those lines (and tested it with single/multiple files, with more/less/nonexistent-pager and so on).

bcoca:
he, even nicer that we get to remove soo many lines

PR #12584, opened 2015-10-01
Remove an unwanted FIXME annotation

amenonsen:
<crab> jimi|ansible: do you think it should be possible to add both
       foo:22 and foo:23 to the inventory?
<jimi|ansible> no

…so we don't want an invitation to FIXME.

### DIFF
--- a/lib/ansible/inventory/ini.py
+++ b/lib/ansible/inventory/ini.py
@@ -288,8 +288,7 @@ class InventoryParser(object):
 
         # Note that we decide whether or not to create a Host based solely on
         # the (non-)existence of its hostname in self.hosts. This means that one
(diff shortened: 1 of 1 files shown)


bcoca:
people actually use that functionality, specially when having local VMs/containers:

localhost:23
localhost:25

amenonsen:
It doesn't work.

bcoca:
which is why the FIXME seems appropriate

amenonsen:
I'm not convinced that it's very common—to make it work with VMs, you would need to have port-forwarding or redirection of some sort set up, and I've only ever seen people using private IP addresses instead. But if you want the functionality (and @jimi-c doesn't renew his objection, since he said "no" earlier) I can just fix it rather than leave the FIXME in place.

PR #12587, opened 2015-10-01
Python 3: don't pass byte strings to syslog/journal

### LINKED ISSUE #12586: Python 3 support for AnsibleModule.log()
I need a place to discuss this before I can submit a pull request, and I don't think a [discussion in the comments of a merged PR](https://github.com/ansible/ansible/pull/12568#discussion_r40802626) is a good place for that.

At the moment, AnsibleModule.log('anything') converts the argument to UTF-8 bytes, which fails on Python 3:

```
ERROR: test_module_utils_basic_ansible_module_user_and_group (units.module_utils.test_basic.TestModuleUtilsBasic)
  File "/home/mg/src/ansible/test/units/module_utils/test_basic.py", line 547, in test_module_utils_basic_ansible_module_user_and_group
    argument_spec = dict(),
  File "/home/mg/src/ansible/lib/ansible/module_utils/basic.py", line 479, in __init__
    self._log_invocation()
  File "/home/mg/src/ansible/lib/ansible/module_utils/basic.py", line 1316, in _log_invocation
    self.log(msg, log_args=log_args)
  File "/home/mg/src/ansible/lib/ansible/module_utils/basic.py", line 1277, in log
    self._log_to_syslog(msg)
  File "/home/mg/src/ansible/lib/ansible/module_utils/basic.py", line 1251, in _log_to_syslog
TypeError: [priority,] message string
```

mgedmin:
Fixes #12586.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -1260,9 +1260,16 @@ class AnsibleModule(object):
             module = 'ansible-%s' % os.path.basename(__file__)
 
             # 6655 - allow for accented characters
-            if isinstance(msg, unicode):
-                # We should never get here as msg should be type str, not unicode
-                msg = msg.encode('utf-8')
+            # On Python 2 we need to pass byte strings to syslog/journal.
+            # On Python 3 we need to pass Unicode strings to syslog/journal.
+            if not isinstance(msg, str):
+                if isinstance(msg, unicode):
+                    # We should never get here as msg should be type str, not unicode
+                    # Note: this branch can only happen on Python 2, because on
+                    # Python 3 we've aliased 'unicode' to 'str'.
+                    msg = msg.encode('utf-8')
+                else:
+                    raise TypeError("msg should be a string (got %s)" % repr(msg))
 
             if (has_journal):
                 journal_args = [("MODULE", os.path.basename(__file__))]
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -1264,7 +1264,6 @@ class AnsibleModule(object):
             # On Python 3 we need to pass Unicode strings to syslog/journal.
             if not isinstance(msg, str):
                 if isinstance(msg, unicode):
-                    # We should never get here as msg should be type str, not unicode
                     # Note: this branch can only happen on Python 2, because on
                     # Python 3 we've aliased 'unicode' to 'str'.
                     msg = msg.encode('utf-8')


lib/ansible/module_utils/basic.py
```diff
@@ -1260,9 +1260,16 @@ def log(self, msg, log_args=None):
...
 
             # 6655 - allow for accented characters
-            if isinstance(msg, unicode):
-                # We should never get here as msg should be type str, not unicode
-                msg = msg.encode('utf-8')
+            # On Python 2 we need to pass byte strings to syslog/journal.
+            # On Python 3 we need to pass Unicode strings to syslog/journal.
+            if not isinstance(msg, str):
+                if isinstance(msg, unicode):
+                    # We should never get here as msg should be type str, not unicode
```
abadger: Now that this method has been cut down I don't think this comment is necessarily true anymore.  best to omit it.

lib/ansible/module_utils/basic.py
```diff
@@ -1260,9 +1260,16 @@ def log(self, msg, log_args=None):
...
+                    # Note: this branch can only happen on Python 2, because on
+                    # Python 3 we've aliased 'unicode' to 'str'.
+                    msg = msg.encode('utf-8')
```
abadger, line 1269: Use `errors='replace'` so that we don't traceback in case of invalid utf-8.  Since this is a logging function, it's okay to display a mangled string.

abadger:
It looks like we don't do any conversion in the case that we're running on python3.  Since modules can call this code, I think we might want to be more paranoid.  Also, I think we might want to convert to text type for systemd.journal for both py2 and py3.  the reason is that by default journalctl will output more helpful text if we do the mangling:

```
>>> journal.send(b'this\xffis a \xfftest')
>>> journal.send(b'this\xffis a \xfftest'.decode('utf-8', errors='replace'))

Oct 01 08:06:27 roan.lan python2[12518]: [16B blob data]
Oct 01 07:43:48 roan.lan python2[10261]: this�is a �test
```

How about something like this:

``` python
if not isinstance(msg, (bytes, unicode)):
    msg = '%s' % msg

if sys.version_info >= (3,)
    if isinstance(msg, bytes):
        journal_msg = msg.decode('utf-8', errors='replace')
        syslog_msg = journal_msg
    else:
        journal_msg = syslog_msg = msg
else:
    if isinstance(msg, bytes):
        journal_msg = msg.decode('utf-8', errors='replace')
        syslog_msg = msg
    else:
         journal_msg = msg
         syslog_msg = msg.encode('utf-8', errors='replace')
```

There's a few places that can traceback in rare cases (an object's **str** returns unicode type on python2, a msg contains surrogateescapes in python3) but if we're going to worry about those we definitely need to_unicode and to_bytes from ansible.utils.unicode to help out.

lib/ansible/module_utils/basic.py
mgedmin, line 1269: `msg.encode('utf-8')` can never raise because UTF-8 can encode every possible Unicode character.

```
>>> len([unichr(c).encode('UTF-8') for c in range(0, 0x10FFFF+1)])
1114112
```

mgedmin:
> It looks like we don't do any conversion in the case that we're running on python3. Since modules can call this code, I think we might want to be more paranoid.

I raise TypeError if they pass anything that's not a unicode string on Python 3.  I think that's paranoid ;)

(I raise TypeError if they pass anything that's not a byte or unicode string on Python 2.)

Note: the current code will stringify non-string things via '%s' % msg if you have python-systemd, but it will pass msg straight to syslog.syslog() if you don't.

```
mg@platonas: ~/src/ansible [git:py3k $=] $ python
Python 2.7.8 (default, Oct 20 2014, 15:05:19) 
[GCC 4.9.1] on linux2
>>> import syslog
>>> syslog.syslog(42)
TypeError: must be string, not int
```

mgedmin:
> Also, I think we might want to convert to text type for systemd.journal for both py2 and py3. the reason is that by default journalctl will output more helpful text if we do the mangling

That's a good point!

Unfortunately I can't easily test systemd.journal on Python 2, because Ubuntu packagers packaged it only for Python 3.  You cannot pass bytes to journal.send() on Python 3:

```
mg@platonas: ~/src/ansible [git:py3k $=] $ python3
Python 3.4.3 (default, Mar 26 2015, 22:03:40) 
[GCC 4.9.2] on linux
>>> from systemd import journal
>>> journal.send(b"testing")
  File "/usr/lib/python3/dist-packages/systemd/journal.py", line 391, in send
    args = ['MESSAGE=' + MESSAGE]
TypeError: Can't convert 'bytes' object to str implicitly
```

mgedmin:
> How about something like this

It looks more complicated at 1st glance than I prefer, but it looks like it ought to work.

mgedmin:
I should mention that I don't intend to work on improving AnsibleModule.log('\xff') on Python 2, since I can't test the changes.

The Travis CI failure was caused by somebody committing an old-style octal literal into the codebase.

I'm not sure what I should do now.

lib/ansible/module_utils/basic.py
abadger, line 1269: That was true until surrogateescape was added:

``` python
>>> a = b'\xff'.decode('utf8', errors='surrogateescape')
>>> a
'\udcff'
>>> '\udcff'.encode('utf-8')
UnicodeEncodeError: 'utf-8' codec can't encode character '\udcff' in position 0: surrogates not allowed
>>> '\udcff'.encode('utf-8', errors='surrogateescape')
b'\xff'
>>> '\udcff'.encode('utf-8', errors='replace')
b'?'
```

abadger:
- I think we should convert byte strings to unicode strings in python3... maybe when we get serious about porting modules to python3 we can change that so that byte strings raise a TypeError... I'm on the fence about whether we should make allowances for byte strings for as long as we're heavily supporting python2 or not.  It definitely feels too early to remove the crutch now, though.
- I'd be okay with the code raising TypeError if we're passed something that's not bytes or text type.
- It sounds like we both like the idea of converting systemd.journal output to unicode strings on py2.

Taken together, I think we end up with something like the code I posted but emitting TypeError instead of trying to convert non-(byte/unicode) objects.  I'll work up an alternative based on that.

lib/ansible/module_utils/basic.py
mgedmin, line 1269: But this only happens on Python 3!  And we don't `.encode('utf-8')` on Python 3.

mgedmin:
Sounds like a plan!  I won't mind if you close this PR without merging once you've an alternative.

abadger:
Okay, alternative here: https://github.com/ansible/ansible/pull/12614

I'll close this one once you have a chance to look over the alternative and make sure it works for you.

mgedmin:
I don't see syslog-related test failures on Python 3 any more.

(I see many new failures due to other recent changes that added some Python-3-incompatible code.  It's _possible_ that they might be hiding syslog problems, but I think that's very unlikely.)

PR #12593, opened 2015-10-01
Since Connection.execute_command() returns bytes, deal with the repurcussions

abadger:
Follow up to https://github.com/ansible/ansible/pull/12447


mgedmin:
LGTM.

The same logic is duplicated for stdout and stderr, why not extract it into a tiny helper method?

abadger:
I guess because it is so tiny.  If it's used in one more place it would make sense, though.  I'll have to keep my eyes open for that.

PR #12595, opened 2015-10-01
Remove mysterious old cruft

amenonsen:
The two lines were added in 2673eb0a and modified in 60e1a1f8, but foo
is completely unused. Removing it doesn't break anything: out it goes!


jimi-c:
Merged, thanks!

PR #12604, opened 2015-10-02
Rework additional ssh argument handling

amenonsen:
Introduce `ansible_ssh_common_args` and `ansible_{sftp,scp,ssh}_extra_args` inventory variables with corresponding command-line overrides (correction: they're command-line settings, not overrides; as with the other settings handled by `set_options`, these have lower priority than task/variable settings).

### DIFF
--- a/lib/ansible/plugins/connection/ssh.py
+++ b/lib/ansible/plugins/connection/ssh.py
@@ -33,6 +33,7 @@ from ansible import constants as C
 from ansible.errors import AnsibleError, AnsibleConnectionFailure, AnsibleFileNotFound
 from ansible.plugins.connection import ConnectionBase
 from ansible.utils.path import unfrackpath, makedirs_safe
+from ansible.utils.vars import combine_vars
 
 SSHPASS_AVAILABLE = None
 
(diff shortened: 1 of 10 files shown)


amenonsen:
@jimi-c this is now ready for review, I moved all of the option-setting logic to PlayContext and made `ssh.py` just use the settings that are passed in.

lib/ansible/plugins/connection/ssh.py
```diff
@@ -221,19 +200,15 @@ def _build_command(self, binary, *other_args):
...
+        # (i.e. inventory or task settings or overrides on the command line).
 
-        if self._play_context.ssh_extra_args:
-            args = self._split_args(self._play_context.ssh_extra_args)
-            self._add_args("command-line added --ssh-extra-args", args)
-        elif self.ssh_extra_args:
-            args = self._split_args(self.ssh_extra_args)
-            self._add_args("inventory added ansible_ssh_extra_args", args)
+        for opt in ['ssh_common_args', binary + '_extra_args']:
+            if getattr(self._play_context, opt):
```
jimi-c: This should be `if getattr(self._play_context, opt, None) is not None:` just to be sure the attribute is present and set.

jimi-c: Probably best to do:

<pre>
        for opt in ['ssh_common_args', binary + '_extra_args']:
            attr = getattr(self._play_context, opt, None)
            if attr is not None:
                args = self._split_args(attr)
                self._add_args("PlayContext set %s" % opt, args)
</pre>

bcoca:
i would remove these from the command line, at this point it is too specialized to really be that useful there.

jimi-c:
Merged, thanks!

PR #12612, opened 2015-10-02
fix ansible-galaxy only installing dependent roles by name

bitmensch:
This reuses the same code that is used for reading roles from the requirements.yml, thus it ensures that a dependency specified in meta/main.yml with e.g. src attribute is installed correctly as well.
Before ansible-galaxy always tried to install dependant roles from galaxy server.

Other fixes include missing import and wrongly used conditional.


jimi-c:
# Closing This Ticket

Hi!

We believe the above commit should resolve this problem for you. This will also be included in the next major release.

Thank you!

PR #12614, opened 2015-10-02
AnsibleModules.log() fix for python3

### LINKED ISSUE #12586: Python 3 support for AnsibleModule.log()

abadger:
Also add unittests for AnsibleModules.log()

Fixes #12586

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -1262,23 +1262,41 @@ class AnsibleModule(object):
                 log_args = dict()
 
             module = 'ansible-%s' % os.path.basename(__file__)
+            if isinstance(module, bytes):
+                module = module.decode('utf-8', 'replace')
 
             # 6655 - allow for accented characters
-            if isinstance(msg, unicode):
-                # We should never get here as msg should be type str, not unicode
-                msg = msg.encode('utf-8')
+            if not isinstance(msg, (bytes, unicode)):
+                raise TypeError("msg should be a string (got %s)" % type(msg))
 
-            if (has_journal):
+            # We want journal to always take text type
+            # syslog takes bytes on py2, text type on py3
+            if sys.version_info >= (3,):
+                if isinstance(msg, bytes):
+                    journal_msg = msg.decode('utf-8', 'replace')
+                    syslog_msg = journal_msg
+                else:
+                    # TODO: surrogateescape is a danger here
+                    journal_msg = syslog_msg = msg
+            else:
+                if isinstance(msg, bytes):
+                    journal_msg = msg.decode('utf-8', 'replace')
+                    syslog_msg = msg
+                else:
+                     journal_msg = msg
+                     syslog_msg = msg.encode('utf-8', 'replace')
+
+            if has_journal:
                 journal_args = [("MODULE", os.path.basename(__file__))]
                 for arg in log_args:
                     journal_args.append((arg.upper(), str(log_args[arg])))
                 try:
-                    journal.send("%s %s" % (module, msg), **dict(journal_args))
+                    journal.send(u"%s %s" % (module, journal_msg), **dict(journal_args))
                 except IOError:
                     # fall back to syslog since logging to journal failed
-                    self._log_to_syslog(msg)
+                    self._log_to_syslog(syslog_msg)
             else:
-                self._log_to_syslog(msg)
+                self._log_to_syslog(syslog_msg)
 
     def _log_invocation(self):
         ''' log that ansible ran the module '''
--- a/test/units/module_utils/basic/test_log.py
+++ b/test/units/module_utils/basic/test_log.py
@@ -1,0 +1,237 @@
+# -*- coding: utf-8 -*-
+# (c) 2012-2014, Michael DeHaan <michael.dehaan@gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+# Make coding more python3-ish
+from __future__ import (absolute_import, division)
+__metaclass__ = type
+
+import sys
+import syslog
+
+from ansible.compat.tests import unittest
+from ansible.compat.tests.mock import patch, MagicMock
+
+from ansible.module_utils import basic
+
+try:
+    # Python 3.4+
+    from importlib import reload
+except ImportError:
+    # Python 2 has reload as a builtin
+
+    # Ignoring python3.0-3.3 (those have imp.reload if we decide we care)
+    pass
+
+
+class TestAnsibleModuleSysLogSmokeTest(unittest.TestCase):
+
+    def setUp(self):
+        self.complex_args_token = basic.MODULE_COMPLEX_ARGS
+        basic.MODULE_COMPLEX_ARGS = '{}'
+        self.am = basic.AnsibleModule(
+            argument_spec = dict(),
+        )
+
+        self.has_journal = basic.has_journal
+        if self.has_journal:
+            # Systems with journal can still test syslog
+            basic.has_journal = False
+
+    def tearDown(self):
+        basic.MODULE_COMPLEX_ARGS = self.complex_args_token
+        basic.has_journal = self.has_journal
+
+    def test_smoketest_syslog(self):
+        # These talk to the live daemons on the system.  Need to do this to
+        # show that what we send doesn't cause an issue once it gets to the
+        # daemon.  These are just smoketests to test that we don't fail.
+
+        self.am.log(u'Text string')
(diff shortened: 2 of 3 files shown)


lib/ansible/module_utils/basic.py
```diff
@@ -1264,21 +1264,37 @@ def log(self, msg, log_args=None):
...
+                     syslog_msg = msg.encode('utf-8', 'replace')
 
-            if (has_journal):
+            if has_journal:
                 journal_args = [("MODULE", os.path.basename(__file__))]
                 for arg in log_args:
                     journal_args.append((arg.upper(), str(log_args[arg])))
                 try:
-                    journal.send("%s %s" % (module, msg), **dict(journal_args))
+                    journal.send(u"%s %s" % (module, journal_msg), **dict(journal_args))
```
mgedmin, line 1294: Could `module` contain non-ASCII characters?

test/units/module_utils/basic/test_log.py
```diff
@@ -0,0 +1,227 @@
...
+            basic.has_journal = False
+
+        self.am.log(u'Text string')
+        self.am.log(u'Toshio くらとみ non-ascii test')
+
+        self.am.log(b'Byte string')
+        self.am.log(u'Toshio くらとみ non-ascii test'.encode('utf-8'))
+        self.am.log(b'non-utf8 :\xff: test')
+
+        basic.has_journal = self.has_journal
```
mgedmin: Minor: if the test fails, the value of `basic.has_journal` won't be restored.  This probably doesn't matter.

test/units/module_utils/basic/test_log.py
```diff
@@ -0,0 +1,227 @@
...
+            argument_spec = dict(),
+        )
+
+    def tearDown(self):
+        basic.MODULE_COMPLEX_ARGS = self.complex_args_token
+
+    def test_smoketest_syslog(self):
+        # These talk to the live daemons on the system.  Need to do this to
+        # show that what we send doesn't cause an issue once it gets to the
+        # daemon.  These are just smoketests to test that we don't fail.
```
mgedmin, line 62: Hm, given those 15(?) other tests that failed on Python 3 in calls to syslog.syslog(), could we monkey-patch the rest of the tests so this is the only test that actually writes to the real syslog/journal?

(EDIT: when I say "the rest of the tests", I mean in test_basic.py.)

test/units/module_utils/basic/test_log.py
```diff
@@ -0,0 +1,227 @@
...
+        self.am.log(u'Text string')
+        self.am.log(u'Toshio くらとみ non-ascii test')
+
+        self.am.log(b'Byte string')
+        self.am.log(u'Toshio くらとみ non-ascii test'.encode('utf-8'))
+        self.am.log(b'non-utf8 :\xff: test')
+
+        basic.has_journal = self.has_journal
+
+    @unittest.skipIf(basic.has_journal != True, 'python systemd bindings not installed')
```
mgedmin: `@unittest.skipUnless(basic.has_journal, ...)` might be clearer.

test/units/module_utils/basic/test_log.py
```diff
@@ -0,0 +1,227 @@
...
+        if sys.version_info >= (3,):
+            for msg, param in self.py3_output_data.items():
+                with patch('syslog.syslog', autospec=True) as mock_func:
+                    self.am.log(msg)
+                    mock_func.assert_called_once_with(syslog.LOG_INFO, param)
+        else:
+            for msg, param in self.py2_output_data.items():
+                with patch('syslog.syslog', autospec=True) as mock_func:
+                    self.am.log(msg)
+                    mock_func.assert_called_once_with(syslog.LOG_INFO, param)
```
mgedmin: The code is the same, just the data differs, so I think it would be clearer to move the loop outside the if.

``` python
output_data = self.py3_output_data if sys.version_info[0] >= 3 else self.py2_output_data
for msg, param in output_data.items():
    ...
```

test/units/module_utils/basic/test_log.py
```diff
@@ -0,0 +1,227 @@
...
+            except NameError:
+                self._fake_out_reload(basic)
+
+    def _fake_out_reload(self, module):
+        """ Something about reload being a builtin is playing havoc here.  If
+        I import it in the same method as I call it then it bashes the builtin
+        reload.  *Even if the import comes after the call*.  Since we need to
+        import it on python3, doing it in a separate method seems to make it work."""
+        from importlib import reload
+        reload(module)
```
mgedmin: Yeah, a name can either be a global or a local in a scope, and Python determines that during compilation time by looking for certain statement types (such as assignments or imports).

You could do this at the top level:

``` python
try:
    # Python 3
    from importlib import reload
except ImportError:
    # Python 2
    pass
```

or use `six.moves.reload_module`.

(Hm, [importlib.reload](https://docs.python.org/3/library/importlib.html#importlib.reload) appeared in Python 3.4, before that it was [imp.reload](https://docs.python.org/3/library/imp.html#imp.reload).  I suppose it's fine to require Python 3.4+ for Ansible -- people who don't have it can fall back to 2.7.)

test/units/module_utils/basic/test_log.py
```diff
@@ -0,0 +1,227 @@
...
+    def test_no_log(self, mock_func):
+        no_log = self.am.no_log
+
+        self.am.no_log = True
+        self.am.log('unittest no_log')
+        self.assertFalse(mock_func.called)
+
+        self.am.no_log = False
+        self.am.log('unittest no_log')
+        mock_func.assert_called_once()
```
mgedmin: There's no [Mock](https://docs.python.org/3/library/unittest.mock.html#the-mock-class).assert_called_once().  Mock will raise an exception since version 1.3 (or Python 3.5 for the stdlib version, see  http://bugs.python.org/issue21238).  Older versions of mock will assume you're calling a method of the mocked object and will silently do nothing.

The correct check is `assert mock_func.call_count == 1`, IIRC.

test/units/module_utils/basic/test_log.py
```diff
@@ -0,0 +1,227 @@
...
+        for msg, param in self.output_data.items():
+            with patch('systemd.journal.send', autospec=True) as mock_func:
+                self.am.log(msg)
+                self.assertEqual(mock_func.call_count, 1, msg='journal.send not called exactly once')
+                self.assertTrue(mock_func.call_args[0][0].endswith(param))
+
+    @patch('systemd.journal.send')
+    def test_log_args(self, mock_func):
+        self.am.log('unittest log_args', log_args=dict(TEST='log unittest'))
+        mock_func.assert_called_once()
```
mgedmin: Here too.

test/units/module_utils/basic/test_log.py
mgedmin: Yup, Python 3.5 catches this:

```
ERROR: test_log_args (units.module_utils.basic.test_log.TestAnsibleModuleLogJournal)
  File "/usr/lib/python3.5/unittest/mock.py", line 1157, in patched
    return func(*args, **keywargs)
  File "/home/mg/src/ansible/test/units/module_utils/basic/test_log.py", line 217, in test_log_args
    mock_func.assert_called_once()
  File "/usr/lib/python3.5/unittest/mock.py", line 583, in __getattr__
    raise AttributeError(name)
AttributeError: assert_called_once

ERROR: test_no_log (units.module_utils.basic.test_log.TestAnsibleModuleLogJournal)
  File "/usr/lib/python3.5/unittest/mock.py", line 1157, in patched
    return func(*args, **keywargs)
  File "/home/mg/src/ansible/test/units/module_utils/basic/test_log.py", line 196, in test_no_log
    mock_func.assert_called_once()
  File "/usr/lib/python3.5/unittest/mock.py", line 583, in __getattr__
    raise AttributeError(name)
AttributeError: assert_called_once
```

Speaking of which, I should submit a PR to add Python 3.5 to tox.ini and .travis.yml

mgedmin: > Speaking of which, I should submit a PR to add Python 3.5 to tox.ini and .travis.yml

#12627.

test/units/module_utils/basic/test_log.py
abadger, line 62: I think test_basic does need some looking at... It's big and unwieldy, though, so I didn't want to touch it in this particular change.   If the calls to syslog/journal that are presently in that file go through this method, then I think it would be fine to monkey patch them out.  We probably should break those tests out and put them in this subdirectory, though, so that people that look at it later know that log() is being tested in this file rather than simply omitted in test_basic.py.

abadger:
Think I fixed up all the comments you had.  Waiting for travis to complete and then I'll push this.

PR #12617, opened 2015-10-02
Replace pipes.quote for shlex_quote

mcsalgado:
pipes.quote is considered deprecated (https://docs.python.org/2/library/pipes.html#pipes.quote) so I think using six here is a fitting compromise.

Also this is a step for bringing python3 compatibility to the ansible codebase.

### DIFF
--- a/lib/ansible/playbook/play_context.py
+++ b/lib/ansible/playbook/play_context.py
@@ -21,12 +21,12 @@
 from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
-import pipes
 import random
 import re
 import string
 
 from ansible.compat.six import iteritems, string_types
+from ansible.compat.six.moves import shlex_quote
 from ansible import constants as C
 from ansible.errors import AnsibleError
 from ansible.playbook.attribute import Attribute, FieldAttribute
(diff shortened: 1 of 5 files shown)


lib/ansible/module_utils/basic.py
```diff
@@ -68,6 +67,8 @@
 import errno
 from itertools import repeat
 
+from six.moves import shlex_quote
```
bcoca: this now forces all targets to have six installed

lib/ansible/playbook/play_context.py
```diff
@@ -422,7 +422,7 @@ def detect_su_prompt(data):
 
             elif self.become_method == 'pbrun':
 
-                prompt='assword:'
+                prompt='Password:'
```
bcoca: this was 'assword' on purpose as diff versions did not capitalize it the same way.

bcoca:
also need to check if the quotes are any different, many combinations of systems/shells/OS and privilege escalation need to be tested

mcsalgado:
@bcoca thanks for the code review, I fixed the commits now.

About the quotes being different, currently the difference is pretty small, you can check the code for python3 and 2:

https://hg.python.org/cpython/file/5b9ffea7e7c3/Lib/shlex.py#l278
https://hg.python.org/cpython/file/45a04eadefd6/Lib/pipes.py#l267

and the corresponding unit tests:
https://hg.python.org/cpython/file/5b9ffea7e7c3/Lib/test/test_shlex.py#l176
https://hg.python.org/cpython/file/45a04eadefd6/Lib/test/test_pipes.py#l76

mcsalgado:
I updated the PR to match #12769

abadger:
I took the time to check the difference in implementation between python2.7 pipes.quote and python3.5's shlex.quote.  They are close enough that I'd be comfortable with this change.  Since this was submitted so long ago, I'll take responsibility for rebasing it to current devel and merging it.

abadger:
Merged via https://github.com/ansible/ansible/commit/ed00741a01ed4eb935797afc74cabf6ce0a0e7fa  Thanks for looking into working on this!

mcsalgado:
You're welcome!

PR #12627, opened 2015-10-04
Add Python 3.5 to tox.ini and Travis

mgedmin:


abadger:
My tox and  travis experience is mostly trial and error but this change looks logical and the output looks like it is a good change.  I'm going to merge it.

Looking hard at the output I noticed that the integration tests are using the wrong version of python (probably the system version, python2.7, each time instead of the specific version that we're trying to test).  However, that seems to be happening before your changes as well so i'll open a separate issue for that until someone has a chance to look into it.

abadger:
# Merged

Hi!

issue is resolved.

Thank you!

mgedmin:
> Looking hard at the output I noticed that the integration tests are using the wrong version of python (probably the system version, python2.7, each time instead of the specific version that we're trying to test). 

On Travis, yes.  I noticed that while I was preparing this PR, and I was going to submit a fix once this one got merged.

I think it should be pretty simple to fix, using [something like this](https://github.com/pypa/pip/blob/develop/.travis.yml#L14-L15).

mgedmin:
> i'll open a separate issue for that 

BTW did you open an issue?  I couldn't find it, but maybe by search powers are too puny.

PR #12640, opened 2015-10-06
Add an expand_csv_tags setting to ec2.ini

amenonsen:
If enabled, this will convert tags of the form "a,b,c" to a list and use
the results to create additional inventory groups.

This is based on PR #8676 by nickpeck (but not a straight rebase—both
the code and the nomenclature have been changed here).

Closes #8676

cc: @willthames @bcoca

### DIFF
--- a/contrib/inventory/ec2.ini
+++ b/contrib/inventory/ec2.ini
@@ -97,6 +97,10 @@ nested_groups = False
 # Replace - tags when creating groups to avoid issues with ansible
 replace_dash_in_groups = True
 
+# If set to true, any tag of the form "a,b,c" is expanded into a list
+# and the results are used to create additional tag_* inventory groups.
+expand_csv_tags = False
+
 # The EC2 inventory output can become very large. To manage its size,
 # configure which groups should be created.
 group_by_instance_id = True
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -141,6 +141,7 @@ except ImportError:
 
 
 class Ec2Inventory(object):
+
     def _empty_inventory(self):
         return {"_meta" : {"hostvars" : {}}}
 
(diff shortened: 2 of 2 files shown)


contrib/inventory/ec2.ini
```diff
@@ -143,3 +143,7 @@ group_by_elasticache_replication_group = True
 # A boto configuration profile may be used to separate out credentials
 # see http://boto.readthedocs.org/en/latest/boto_config_tut.html
 # boto_profile = some-boto-profile-name
+
+# If set to true, any tag of the form "a,b,c" is expanded into a list
```
halberom: I'd move this up to somewhere between lines 97 and 117, which is where the current group creation bit is.  Everything after is filtering.

contrib/inventory/ec2.py
```diff
@@ -141,6 +141,9 @@
 
 
 class Ec2Inventory(object):
+
+    expand_csv_tags = False
```
halberom: maybe move this up to line 329 as an else, matching the nested_groups if/else.

contrib/inventory/ec2.py
```diff
@@ -690,14 +696,20 @@ def add_instance(self, instance, region):
...
-                self.push(self.inventory, key, dest)
-                if self.nested_groups:
-                    self.push_group(self.inventory, 'tags', self.to_safe("tag_" + k))
-                    self.push_group(self.inventory, self.to_safe("tag_" + k), key)
+                    values = [v]
+
+                for v in values:
+                    if v:
+                        key = self.to_safe("tag_" + k + "=" + v)
+                    else:
```
halberom: may be blind, but think this will never get fired if v is empty.  you create an empty list, which will be skipped in the for loop.

amenonsen:
I've pushed a revised version that moves stuff around according to @halberom's suggestions (though I don't think the empty-tag comment applies, since the array won't be empty but contain an empty string, unless I'm missing something).

PR #12644, opened 2015-10-06
You can create private instances in SoftLayer without a primaryIpAddress

sidick:
##### SUMMARY
Currently the SoftLayer inventory script ignores any servers which don't have a public IP address which isn't always ideal

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
plugins/inventory/softlayer.py

##### ANSIBLE VERSION
2.3


calfonso:
@sidick Would you like to rebase this so we can merge?

sidick:
OK, will get that done

PR #12657, opened 2015-10-07
Explictly set the version of strict mode to use for powershell modules.

jhawkesworth:
I'd like the version of strict mode to be set explicitly in case next update to powershell comes with a new version of strict mode that.

Also updated the module development windows modules checklist to state the version of strict mode in use.


jimi-c:
Merged, thanks!

PR #12674, opened 2015-10-07
(TRIVIAL) Fail if the vault password script returns non-zero.

### LINKED ISSUE #12673: Return code is not checked when executing vault password script
...which means the error message is treated as a password and leads to potentially irreversibly encrypting the data with unknown password.

sudarkoff:
Fixes #12673

Not checking the error code could potentially lead to encrypting the data with the error message as the password.

Before:

```
$ ansible-vault --vault-password-file=~/.secrets/.vaultpw encrypt group_vars/prod
Error reading secret/ansible/vault: Get https://vault:8200/v1/secret/ansible/vault: net/http: TLS handshake timeout
Encryption successful
```

After:

```
$ ansible-vault --vault-password-file=~/.secrets/.vaultpw encrypt group_vars/prod
Error reading secret/ansible/vault: Get https://vault:8200/v1/secret/ansible/vault: net/http: TLS handshake timeout
ERROR! Vault password script /Users/georges/.secrets/.vaultpw returned non-zero (1).
```

(If this looks acceptable, I'll add tests.)

/cc @msabramo @0x9900

### DIFF
--- a/lib/ansible/cli/__init__.py
+++ b/lib/ansible/cli/__init__.py
@@ -501,6 +501,8 @@ class CLI(object):
             except OSError as e:
                 raise AnsibleError("Problem running vault password script %s (%s). If this is not a script, remove the executable bit from the file." % (' '.join(this_path), e))
             stdout, stderr = p.communicate()
+            if p.returncode != 0:
+                raise AnsibleError("Vault password script %s returned non-zero (%s)." % (this_path, p.returncode))
             vault_pass = stdout.strip('\r\n')
         else:
             try:


lib/ansible/cli/__init__.py
```diff
@@ -501,6 +501,8 @@ def read_vault_password_file(vault_password_file, loader):
             except OSError as e:
                 raise AnsibleError("Problem running vault password script %s (%s). If this is not a script, remove the executable bit from the file." % (' '.join(this_path), e))
             stdout, stderr = p.communicate()
+            if p.returncode != 0:
+                raise AnsibleError("Vault password script %s returned non-zero (%s)." % (this_path, p.returncode))
```
bcoca, line 505: I would also attach stdout/stderr

PR #12678, opened 2015-10-08
fix some warning of undefined name.

soarpenguin:


sivel:
The changes in `module_utils` are incorrect.  Those files are different from traditional Python deps/imports and have dependencies on other files.  Since they are not used as true imports, but are used for building single file modules to ship to a remote server, the dependencies and imports are mixed with the actual module as well as basic.py which includes most of these missing imports.

soarpenguin:
@sivel OK, fixed it.

soarpenguin:
@sivel,  it's means remove all of added imports?

lib/ansible/cli/__init__.py
```diff
@@ -473,13 +473,13 @@ def pager_pipe(text, cmd):
...
-        t = self._ITALIC.sub("`" + r"\1" + "'", text)    # I(word) => `word'
-        t = self._BOLD.sub("*" + r"\1" + "*", t)         # B(word) => *word*
-        t = self._MODULE.sub("[" + r"\1" + "]", t)       # M(word) => [word]
-        t = self._URL.sub(r"\1", t)                      # U(word) => word
-        t = self._CONST.sub("`" + r"\1" + "'", t)        # C(word) => `word'
+        t = cls._ITALIC.sub("`" + r"\1" + "'", text)    # I(word) => `word'
+        t = cls._BOLD.sub("*" + r"\1" + "*", t)         # B(word) => *word*
+        t = cls._MODULE.sub("[" + r"\1" + "]", t)       # M(word) => [word]
+        t = cls._URL.sub(r"\1", t)                      # U(word) => word
+        t = cls._CONST.sub("`" + r"\1" + "'", t)        # C(word) => `word'
```
abadger, line 483: This is a good hcange stylistically but it shouldn't have raised a SyntaxError... Could you post what you were seeing so I can check if something else is going on?  (Or if I'm missing something basic).

abadger:
I think your get_driver import in gce was good to leave in.

@sivel Probably for 2.1 I'm going to get a zipimporter in as an alternative to the current replacer code.  That should make module_utils more like regular python imports.  So some of these cleanups to module_utils are a good idea if they don't break existing usage.  Others may be premature though.  In particular, I agree that we shouldn't be importing between the module_utils yet.

soarpenguin:
@sivel  close this issue and new issue in #12689

PR #12687, opened 2015-10-09
Force SSL transport for pywinrm updates, get host+group vars

nitzmahone:
I PR'd a change to pywinrm to allow server certs to be ignored; but it's only on the SSL transport (which we were previously ignoring). For this to work more generally (eg, when defined as a group_var), we're also now pulling the named ansible_winrm_\* args from the merged set of host/group vars, not just host_vars.

### DIFF
--- a/lib/ansible/plugins/connection/winrm.py
+++ b/lib/ansible/plugins/connection/winrm.py
@@ -48,6 +48,7 @@ from ansible.plugins.connection import ConnectionBase
 from ansible.plugins import shell_loader
 from ansible.utils.path import makedirs_safe
 from ansible.utils.unicode import to_bytes, to_unicode, to_str
+from ansible.utils.vars import combine_vars
 
 class Connection(ConnectionBase):
     '''WinRM connections over HTTP/HTTPS.'''
@@ -76,7 +77,7 @@ class Connection(ConnectionBase):
         '''
         Override WinRM-specific options from host variables.
         '''
-        host_vars = host.get_vars()
+        host_vars = combine_vars(host.get_group_vars(), host.get_vars())
 
         self._winrm_host = self._play_context.remote_addr
         self._winrm_port = int(self._play_context.port or 5986)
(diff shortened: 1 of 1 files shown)


nitzmahone:
The relevant pywinrm PR is https://github.com/diyan/pywinrm/pull/74 

CC @cchurch

lib/ansible/plugins/connection/winrm.py
```diff
@@ -76,7 +77,7 @@ def set_host_overrides(self, host):
         '''
         Override WinRM-specific options from host variables.
         '''
-        host_vars = host.get_vars()
+        host_vars = combine_vars(host.get_group_vars(), host.get_vars())
```
bcoca, line 80: this should not be happening here, play_context should have all pertinent connection info

nitzmahone, line 80: I don't see where- remember that set_host_overrides is where we pick up connection-specific vars. _play_context.vars is empty at this point...

bcoca, line 80: hmm, forgot about that 'feature', still think that a generic function in play_context that gets ansible_<pluginname>_\* vars injected might be a better approach

@jimi-c @abadger ?

jimi-c, line 80: That was my first thought upon looking at this as well, but if I remember right the problem is we don't pass the variables down below the level of the action plugin anymore, so this was the workaround for that.

dreamcat4:
Hi guys. If we could try to come back to this problem / get some movement or any progress clarification on this issue.

As it's really hurting windows usage / uptake on v2.0. Which is the min. version needed for all of the newer windows modules are being added.

Still on anisble v1.9.4 here due to impossible-to-workaround 500 certificate verification failed error. Downgrading the system python to less than < 2.7.9 really isn't a way forwards.

nitzmahone:
The relevant pywinrm update has been released ([pywinrm 0.1.1](https://pypi.python.org/pypi/pywinrm/0.1.1))- @jimi-c @bcoca can we decide if we want to do something different here?

dreamcat4:
So if I understand correctly, the 1st PR containing most of the work has already been merged, and now pywinrm has been updated to include that code.

... but to use it we also need this PR too? Or we can hack some other local plugins / callback ?

It would be nice to know thanks, a clear and complete instruction how to avoid the 500 certificate error on ansible 2.0. Many thanks.

nitzmahone:
I'll do doc updates once all the bits to light it up are included. Short answer is that, with everything in place, you just have to set ansible_winrm_server_cert_validation=ignore on the Windows inventory hosts/groups.

This PR is required to close the loop. It does two things: 
- allows custom winrm connection vars to be sourced from group vars, not just host vars. 
- switches the pywinrm transport to https when using an https url.

Ansible was previously not using the "https" pywinrm transport, because when you didn't need to alter anything about the https context, the "plaintext" context would also work fine with an https url (basically by accident). With the new SSLContext  stuff I added in pywinrm 0.1.1 that allows cert validation to be disabled, you have to be using the "https" transport for those options to work. The second part of this PR fixes that- correctly selecting the https transport for https urls.

dreamcat4:
Sounds excellent. Well let's merge it then!

Seriously though what's the holdup with it? Some non-standard approach to setting the vars or something like that? Perhaps @bcoca you can shed some light what's missing here? It's too hackish? We need to fundamentally restructure ansible or something?

This is a serious problem and won't go away by itself :)

dreamcat4:
Hooray! This is awesome guys. And I can finally scroll up to v2.x. Many thanks.

PR #12688, opened 2015-10-09
Ansible ssh changes docswork

docschick:
Versioning note for removing "ssh" from "ansible_user" variables in 2.0, which doesn't work in older versions


bcoca:
would it not be easier to just create a cross reference and point at it from each instance?

docschick:
Yes, it would. I'd prefer it done using an "include" variable, so that it gets posted in each section but only has to be maintained in one location.  Cross-referencing could be okay, but I'd feel better about including the actual note instead of a link to it and some text telling the user why they need to view this link.   When I wrote the original notes, I had them tailored for each section. After talking with jmartin, we went with a more succinct note that would work for all. It just didn't dawn on me at the moment that repetition like that was less than ideal.   So, to take this approach, where should I put the shared content to be referenced among the various pieces? For Tower, I made a common directory at the same level as the directories for the various manuals. Is there something like this already in place for Core?

amenonsen:
depricated → deprecated everywhere.

bcoca:
We don't have a place like that, but in this case i might put it where we list the variables.

PR #12689, opened 2015-10-09
Remove some warning of undefined name and fix classmethod syntax error.

soarpenguin:
remove some warning of undefined name and fix classmethod syntax error.

### DIFF
--- a/lib/ansible/cli/__init__.py
+++ b/lib/ansible/cli/__init__.py
@@ -473,13 +473,13 @@ class CLI(object):
             pass
 
     @classmethod
-    def tty_ify(self, text):
+    def tty_ify(cls, text):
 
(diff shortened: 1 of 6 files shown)


abadger:
@soarpenguin Could you explain what was happenig with the self => cls rename?  My understanding of python methods and classmethods is that the instance or class is passed in the first argument -- it doesn't matter what the variable holding it is named.  So I like the change as it makes the code clearer but I'm not seeing how it causes a SyntaxError.  could you elaborate on what you were seeing?  (Just want to learn if there's some problem that I should look for in other code).

soarpenguin:
@abadger i never seeing it causes a SyntaxError,  just liked make the code clearer.

abadger:
Ah okay, I was confused by the PR summary.

PR #12690, opened 2015-10-09
Use AWS_DEFAULT_REGION env var if none specified

sclausson:
Since boto3 and aws cli expect AWS_DEFAULT_REGION env var, it would be good if it would be supported in get_aws_connection_info() as well.

http://boto3.readthedocs.org/en/latest/guide/configuration.html
http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html#cli-environment


jimi-c:
Merged, thanks!

PR #12698, opened 2015-10-09
bugfix: serial keyword can be specified as a percentage

klshxsh:
The serial keyword can be specified as a percentage, e.g '30%' or a number e.g. '3'; therefore it needs to be a string to allow both types

### DIFF
--- a/lib/ansible/playbook/play.py
+++ b/lib/ansible/playbook/play.py
@@ -87,7 +87,7 @@ class Play(Base, Taggable, Become):
     _any_errors_fatal    = FieldAttribute(isa='bool', default=False, always_post_validate=True)
     _force_handlers      = FieldAttribute(isa='bool', always_post_validate=True)
     _max_fail_percentage = FieldAttribute(isa='percent', always_post_validate=True)
-    _serial              = FieldAttribute(isa='string', default='0', always_post_validate=True)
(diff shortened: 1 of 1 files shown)


lib/ansible/playbook/play.py
```diff
@@ -87,7 +87,7 @@ class Play(Base, Taggable, Become):
     _any_errors_fatal    = FieldAttribute(isa='bool', default=False, always_post_validate=True)
     _force_handlers      = FieldAttribute(isa='bool', always_post_validate=True)
     _max_fail_percentage = FieldAttribute(isa='percent', always_post_validate=True)
-    _serial              = FieldAttribute(isa='int', default=0, always_post_validate=True)
+    _serial              = FieldAttribute(isa='string', default='0', always_post_validate=True)
```
jimi-c: This should be `isa='percent'` (just like _max_fail_percentage above it).

klshxsh:
Thanks - I have made the suggested change (from string to percent)

jimi-c:
Hi @klshxsh, this was actually my mistake - because of the way the code is written, changing the value to a `percentage` type broke things, so I fixed it and tweaked a few other things as well. I've also added a unit test for `_get_serialized_batches()` which should hopefully catch any future bugs.

Thanks!

klshxsh:
No worries, thanks.

PR #12699, opened 2015-10-09
Try checking other environment variables instead of accepting empty ones

mcsalgado:
##### SUMMARY
Also it's marginally faster to just use dict.get() directly this way.

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
lib/ansible/module_utils/ec2.py

##### ANSIBLE VERSION
2.3

### DIFF
--- a/lib/ansible/module_utils/ec2.py
+++ b/lib/ansible/module_utils/ec2.py
@@ -85,63 +85,33 @@ def get_aws_connection_info(module, boto3=False):
     # Check module args for credentials, then check environment vars
     # access_key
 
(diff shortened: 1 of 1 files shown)


ryansb:
I'm 👎 on this change because it changes the documented behavior, and deviates from how other tools (AWS CLI, etc) use these environment variables. I'd prefer not to start adding case insensitivity here unless you also do the docs changes so users know what to expect if an environment variable is empty.

calfonso:
Hi!

Last comment from @ryansb without response.

PR #12704, opened 2015-10-11
allow for sshpass support for `synchronize`

ritzk:
Needs: https://github.com/ansible/ansible-modules-core/pull/2244
- sshpass support
- env support for run_command

delegate_to fix - https://github.com/ansible/ansible/pull/12752
##### Security consideration
- /proc/pid/environ are only readable by owner, or root. root can read almost everything based on selinux/security policies (including /proc/kcore)
- ssh client tends to strip out env variables, SSHPASS will not be seen by client process (SendEnv ssh config bit)
##### reasons to opt for env var
- module.run_command does not support passing an fd to command module.
- ssh and sshpass do not support reading password from stdin pipe ( data arg of run_command)
- passing fd number to sshpass will not work, as fd no can be random.
- adding support for setting up env was the least invasive and most reliable change

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -1609,7 +1609,7 @@ class AnsibleModule(object):
             # rename might not preserve context
             self.set_context_if_different(dest, context, False)
 
-    def run_command(self, args, check_rc=False, close_fds=True, executable=None, data=None, binary_data=False, path_prefix=None, cwd=None, use_unsafe_shell=False, prompt_regex=None):
+    def run_command(self, args, check_rc=False, close_fds=True, executable=None, data=None, binary_data=False, path_prefix=None, cwd=None, use_unsafe_shell=False, prompt_regex=None, shell_env=None):
         '''
         Execute a command, returns rc, stdout, and stderr.
         args is the command to run
@@ -1658,10 +1658,14 @@ class AnsibleModule(object):
         msg = None
         st_in = None
 
-        # Set a temporary env path if a prefix is passed
         env=os.environ
+        # Set a temporary env path if a prefix is passed
         if path_prefix:
             env['PATH']="%s:%s" % (path_prefix, env['PATH'])
+
+        # setup shell env if passed
+        if shell_env:
+           env.update(shell_env)
 
         # create a printable version of the command for use
         # in reporting later, which strips out things like
@@ -1705,7 +1709,7 @@ class AnsibleModule(object):
             stderr=subprocess.PIPE
         )
 
-        if path_prefix:
+        if path_prefix or shell_env:
             kwargs['env'] = env
         if cwd and os.path.isdir(cwd):
             kwargs['cwd'] = cwd
--- a/lib/ansible/plugins/action/synchronize.py
+++ b/lib/ansible/plugins/action/synchronize.py
@@ -173,6 +173,10 @@ class ActionModule(ActionBase):
         if self._task.args.get('mode', 'push') == 'pull':
             (dest_host, src_host) = (src_host, dest_host)
 
+        # MUNGE SSH_PASSWD INFO
+        if self._task.args.get('login_password', None):
+            self._task.args['login_password'] = task_vars.get('ansible_ssh_pass') or self._play_context.password
+
         # MUNGE SRC AND DEST PER REMOTE_HOST INFO
         src = self._task.args.get('src', None)
         dest = self._task.args.get('dest', None)


ritzk:
Munge  is used by sshpass bit here - https://github.com/ansible/ansible-modules-core/pull/2244

tima:
@ritzk: Thanks for your submission. Please make you delegate_to fix and your ssh password implementation separate PRs. These are separate concerns complicated by the multiple competing PRs we are trying to resolve to implement ssh password support in this module. 

Also, I've seen your explanations for your implementation here and in other threads. It's not clear to me what you find broken about the delegate_to logic though. When you separate these commits, please add that to the comments for context.

ritzk:
@tima 

Single line fix. This bit ->

```
+        if delegate_to:
+            dest_host = task_vars.get('inventory_hostname')
+        else
```

Otherwise, dest_host would be the delegate itself, and not the inventory .This we end up trying to run rsync locally on the same system

```
host: serverA
- synchronize: src=/etc
               dest=/tmp
  delegate_to: serverB
  remote_user: centos

 serverB$ ssh -S none - -l centos 127.0.0.1 rsync --server -logDtprze.iLsf --log-format=%i --delay-updates . /tmp
```

Personally, I would like to keep this merged with the same commit.

tima:
What version did you use to test that delegate code with? Can you provide
the test code to reproduce the problem you are fixing?

There's a whole lot of code in the synchronize action module that follows
your change for manipulating the dest_host to determine the proper delegate
or remote host in there already. If there's a bug, short circuiting the
logic that follows would not be the way to go about addressing it. I'm
pretty sure that would just introduce other bugs for other users.

On Monday, October 12, 2015, Ritesh Khadgaray notifications@github.com
wrote:

> @tima https://github.com/tima
> 
> Single line fix. This bit ->
> if delegate_to:
> - dest_host = task_vars.get('inventory_hostname')
> 
> Otherwise, dest_host would be the delegate itself, and not the inventory
> .This we end up trying to run rsync locally on the same system
> 
> host: serverA
> - synchronize: src=/etc
>   dest=/tmp
>   delegate_to: serverB
>   remote_user: centos
> 
> serverB$ ssh -S none - -l centos 127.0.0.1 rsync --server -logDtprze.iLsf
> --log-format=%i --delay-updates . /tmp
> 
> —
> https://github.com/ansible/ansible/pull/12704#issuecomment-147501476.

## 

[image: photo]
Timothy Appnel
Solutions Architect
919.670.0720
Ansible, Inc. _•_ www.ansible.com

ritzk:
@tima devel/2.0.0.3beta1 branch.

I could not spot any place where the dest_host is set by delegate_to logic.

## 

src_host = localhost.
dest_host = ansible_ssh_host or inventory_hostname

inventory_hostname always points to the inventory host.

ansible_ssh_host points to the system, the module is running on. This could be the delegate itself. Leading to this issue, where rsync runs locally.

```
    if dest_host == delegate_to:
        # ...
        dest_host = '127.0.0.1'
```

tima:
I don't think you're reading they all of the code. Do you have test code
for that can produce a problem?

On Tuesday, October 13, 2015, Ritesh Khadgaray notifications@github.com
wrote:

> @tima https://github.com/tima devel branch.
> 
> src is always the localhost.
> 
> ansible_ssh_host points to the system, the module is running on. This
> could be the delegate itself. Leading to this issue
> 
> inventory_hostname always points to the inventory.
> 
> —
> https://github.com/ansible/ansible/pull/12704#issuecomment-147614092.

## 

[image: photo]
Timothy Appnel
Solutions Architect
919.670.0720
Ansible, Inc. _•_ www.ansible.com

ritzk:
This fails for me. I used an Exception to catch this.

    $ vim ./lib/ansible/plugins/action/synchronize.py
    125         # running on so localhost could be a non-controller machine if
    126         # delegate_to is used)
    127         src_host  = '127.0.0.1'
    128         raise Exception("ansible_ssh_host: %s \ninventory_hostname: %s"%(task_vars.get('ansible_ssh    _host'),task_vars.get('inventory_hostname')))
    129         dest_host = task_vars.get('ansible_ssh_host') or task_vars.get('inventory_hostname')
    130 
    131         dest_is_local = dest_host in C.LOCALHOST

      File "/home/ritesh/Projects/github/ansible/ansible/lib/ansible/plugins/action/synchronize.py", line 128, in run
        raise Exception("ansible_ssh_host: %s \ninventory_hostname: %s"%(task_vars.get('ansible_ssh_host'),task_vars.get('inventory_hostname')))
    Exception: ansible_ssh_host: 192.168.56.101 
    inventory_hostname: 192.168.56.102

--
    $ cat ~/inventory    
    192.168.56.102

--
    $ cat sync-test.yml
    ....
      vars:
        staging_server: 192.168.56.101

      tasks:
        - synchronize: src=/etc
                       dest=/tmp
                       login_password=true
          delegate_to: "{{ staging_server }}"

ritzk:
for delegate_to details - https://github.com/ansible/ansible/pull/12704

tima:
Can we please close this one to consolidate all efforts to resolve this thru #11657. Submitting a competing PR like this to what @elvis-macak has been leading is complicating matters and is making it harder to come to some sort of consensus.

ritzk:
@tima , I have shoot an email to @elvis-macak to see what we can do about this.

elvismacak:
yeah, I didn't know the multi submitting will lead complicating matters. should I close it myself if it's been finished?

tima:
@elvis-macak: It's only complicating matters in that the same topic is being discussed in multiple places at once and it's hard for anyone to follow along. @ritzk needs to close his PRs and work thru what you started.  Realistically, none of the current solutions are good, just 'less bad' than the others so this is not going to get resolved quickly and without the core team getting involved.

The problem with the pipe solution that @sivel suggested to you is that the synchronize module may (in the case of the delegate_to usage scenario) run remotely unlike a connection plugin he referenced that always runs local to the controller. (I overlooked this when I called it a good idea.) Piping becomes messy to impossible as I understand it over a remote connection. Sending a password via a module param is insecure and potentially dangerous. An environmental variable is also rather insecure and could mess with SSH pipelining. 

So we need to step carefully and consult with the core engineers. In the meanwhile I would like to consolidate all of this thinking and discussions in to one place -- I've designated that place to be here.

bcoca:
closing both these tickets as they are flawed implementations, opening a feature request and a discussion on how to do this securely and in a way that functions for all

tima:
Fair enough @bcoca.

elvismacak:
@bcoca, @timothy, thanks

PR #12732, opened 2015-10-13
Specify supported RHEL7's six version

mcsalgado:
As expressed by @abadger here: 4f6f2c21e877cd15300edea1a1ef3b4ab9edf6a1

Because Ansible is stuck with six 1.3.0 for now, I suggested on IRC to create a branch just for "when RHEL updates six" to put some python3 compatibility commits there. Lines like this (https://github.com/ansible/ansible/blob/e6d3c6745f6ab7f04c18dfe4fda21fcd6a98e909/lib/ansible/plugins/connection/winrm.py#L28), my PR #12617 and I believe some other stuff by @mgedmin only works for newer versions of six.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ setup(name='ansible',
       author_email='support@ansible.com',
       url='http://ansible.com/',
       license='GPLv3',
-      install_requires=['paramiko', 'jinja2', "PyYAML", 'setuptools', 'pycrypto >= 2.6', 'six'],
+      install_requires=['paramiko', 'jinja2', "PyYAML", 'setuptools', 'pycrypto >= 2.6', 'six >= 1.3.0'],
       package_dir={ '': 'lib' },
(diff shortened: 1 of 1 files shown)


mcsalgado:
d8c8ca11cfa0787bc14655439b080a9b7c4962e5 is another relevant commit

mgedmin:
If it's important for Ansble to work with six 1.3.0, it may be a good idea to add `six==1.3.0` to `test-requirements.txt` so Travis will enforce that.

abadger:
@mcsalgado I just merged bundling of python-six so that we can rely on having a more current version of the library: https://github.com/ansible/ansible/pull/12769

So I'm going to close this PR and you should be free to use more current things like six.moves.shlex_quote.  Just be sure to use ansible.compat.six  to import six functionality.

mcsalgado:
Nice, thanks for the work @abadger

PR #12733, opened 2015-10-13
added "include" for ansible_ssh_ change to another section

docschick:
changes note added to another section of intro_inventory.rst for clarity


abadger:
Thanks @docschick!

PR #12740, opened 2015-10-14
Python 3: don't convert module arguments to bytes

mgedmin:
Fixes a test failure:

```
ERROR: test_module_utils_basic_ansible_module_creation (units.module_utils.test_basic.TestModuleUtilsBasic)
  File "/home/mg/src/ansible/test/units/module_utils/test_basic.py", line 250, in test_module_utils_basic_ansible_module_creation
    supports_check_mode=True,
  File "/home/mg/src/ansible/lib/ansible/module_utils/basic.py", line 470, in __init__
    self._check_required_arguments()
  File "/home/mg/src/ansible/lib/ansible/module_utils/basic.py", line 1050, in _check_required_arguments
    self.fail_json(msg="missing required arguments: %s" % ",".join(missing))
  File "/home/mg/src/ansible/lib/ansible/module_utils/basic.py", line 1445, in fail_json
    sys.exit(1)
SystemExit: 1
-------------------- >> begin captured stdout << ---------------------
{"msg": "missing required arguments: foo", "failed": true}
```

because converting 'foo' to bytes yields b'foo' on Python 3, which doesn't match the native-unicode 'foo' argument spec.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -1240,10 +1240,7 @@ class AnsibleModule(object):
 
     def _load_params(self):
         ''' read the input and return a dictionary and the arguments string '''
-        params = json.loads(MODULE_COMPLEX_ARGS)
-        if str is bytes:
-            # Python 2
-            params = json_dict_unicode_to_bytes(params)
+        params = json_dict_unicode_to_bytes(json.loads(MODULE_COMPLEX_ARGS))
         if params is None:
             params = dict()
         return params
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -1240,7 +1240,10 @@ class AnsibleModule(object):
 
     def _load_params(self):
         ''' read the input and return a dictionary and the arguments string '''
-        params = json_dict_unicode_to_bytes(json.loads(MODULE_COMPLEX_ARGS))
+        params = json.loads(MODULE_COMPLEX_ARGS)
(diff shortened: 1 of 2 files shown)


abadger:
I think this one requires us to decide what we want to do about python3 in the modules.  I think we're too close to 2.0 to take on trying to make the modules handle python3 so I think skipping the tests is the way to go for now.

If we're only considering python2 for modules right now, then the test is what's wrong.  We're never going to instantiate an AnsibleModule on the controller so we're never going to use _load_params() under python3.  Therefore it needs to be run under python2... So skipping the test if we're running them under pyhton3 seems appropriate.

If we do that, (and we can also skip tests for anything else in AnsibleModule that requires instantiating the object.  Standalone functions we probably do still want to test, though) then we probably want to make sure we know which tests we have to revisit once we form a coherent plan for modules running under python3.  Maybe make skipping use a string that's easily grep'able and consistent so that we can find all of them later?

mgedmin:
I've no idea about the timeline for 2.0, but I strongly suspect full Python 3 support won't make it in time for either the controller or the modules -- there's just too much (untested) code that needs to be changed.

The Python 2.4 compatibility requirement for modules makes it easier to focus on Python 3 support on the controller and leave the modules for later, so we can do that.

I'm not rebasing the branch but reverting the commit, in the hopes that my original fix will be easier to find once we decide that it's time to add Python 3 support for the modules.

abadger:
Sounds like a plan.

PR #12751, opened 2015-10-14
Fix links to dynamic inventory script example.

jdelaporte:
:doc`aws_example` no longer existed.


abadger:
Thanks!  That looks great.

PR #12752, opened 2015-10-14
fix delegate_to with sync action

ritzk:
@tima moved delegate_to fix out here ( https://github.com/ansible/ansible/pull/12704 )

Single line fix. This bit ->

```
+        if delegate_to:
+            dest_host = task_vars.get('inventory_hostname')
+        else
```

Otherwise, dest_host would be the delegate itself, and not the inventory .This we end up trying to run rsync locally on the same system
## 

 I used an Exception to catch this to see why rsync was trying to use the delegate_to host as endpoint itself.

```
$ vim ./lib/ansible/plugins/action/synchronize.py
127         src_host  = '127.0.0.1'
128         raise Exception("\nansible_ssh_host: %s \ninventory_hostname: %s"%(task_vars.get('ansible_ssh    _host'),task_vars.get('inventory_hostname')))
129         dest_host = task_vars.get('ansible_ssh_host') or task_vars.get('inventory_hostname')
130         ...
144         if dest_host == delegate_to:
145             ...
148             dest_host = '127.0.0.1'
149             use_delegate = True


  File "/home/ritesh/Projects/github/ansible/ansible/lib/ansible/plugins/action/synchronize.py", line 128, in run
    raise Exception("\ninventory_hostname: %s\nansible_ssh_host: %s " %(task_vars.get('inventory_hostname'),task_vars.get('ansible_ssh_host')))
Exception: 
inventory_hostname: 192.168.56.101
ansible_ssh_host: 192.168.56.102 
```
## 

```
$ cat ~/inventory    
192.168.56.101
```
## 

```
$ cat sync-test.yml
....
  vars:
    staging_server: 192.168.56.102

  tasks:
    - synchronize: src=/etc
                   dest=/tmp
                   login_password=true
      delegate_to: "{{ staging_server }}"
```
## 

from "ps aux" on staging_server

```
 serverB$ ssh -S none - -l centos 127.0.0.1 rsync --server -logDtprze.iLsf --log-format=%i --delay-updates . /tmp
```
## 

If we are using delegate_to action. as per my understanding synchronize should on delegate_to host, and rsync to ansible_inventory host. With current code, sync tries to run this locally.
- src_host is always localhost (127.0.0.1)
- dest_host is set to ssh_hostname or inventory_hostname
- ssh_hostname is the host module is running on, ( which can be the delegate_to host )
- dest_host needs to be inventory_hostname as this is our dest in case of delegate_to action

### DIFF
--- a/lib/ansible/plugins/action/synchronize.py
+++ b/lib/ansible/plugins/action/synchronize.py
@@ -125,7 +125,10 @@ class ActionModule(ActionBase):
         # running on so localhost could be a non-controller machine if
         # delegate_to is used)
         src_host  = '127.0.0.1'
-        dest_host = task_vars.get('ansible_ssh_host') or task_vars.get('inventory_hostname')
+        if delegate_to:
+            dest_host = task_vars.get('inventory_hostname')
+        else:
+            dest_host = task_vars.get('ansible_ssh_host') or task_vars.get('inventory_hostname')
 
         dest_is_local = dest_host in C.LOCALHOST
 
@@ -200,14 +203,11 @@ class ActionModule(ActionBase):
                     user = task_vars.get('ansible_ssh_user') or self._play_context.remote_user
 
             # use the mode to define src and dest's url
-            if self._task.args.get('mode', 'push') == 'pull':
-                # src is a remote path: <user>@<host>, dest is a local path
-                src = self._process_remote(src_host, src, user)
-                dest = self._process_origin(dest_host, dest, user)
-            else:
-                # src is a local path, dest is a remote path: <user>@<host>
-                src = self._process_origin(src_host, src, user)
-                dest = self._process_remote(dest_host, dest, user)
+            set_mode = lambda host, var: \
+                          self._process_origin(host, var, user) if host == '127.0.0.1' \
+                     else self._process_remote(host, var, user)
+            src = set_mode(src_host, src)
+            dest = set_mode(dest_host, dest)
         else:
             # Still need to munge paths (to account for roles) even if we aren't
             # copying files between hosts


tima:
@jimi-c: There's some sort of bug in the conditional starting at line 181 this patch (currently) does not address.

abadger:
Not sure if the bug is in the conditional... When I wrote the conditional everything was working.  Now they aren't.  Synchronize is hard to test because we must use ssh with multiple hosts to test it well and our integration tests currently don't have that framework. Would be nice to fix that....

Here's the playbook that I'm using.  I go through and uncomment each task one at a time and see which ones do what I expect and which ones fail.

```
---
- hosts: rhel6-test:localhost:fedora21
  vars:
    top_dir: /var/tmp
    files:
      - "{{ top_dir }}/fedora"
      - "{{ top_dir }}/redhat"
      - "{{ top_dir }}/fedora21"
      - "{{ top_dir }}/redhat6"
  tasks:
    - name: cleanup
      file:
        state: absent
        path: "{{ item }}"
      with_items: files

# Localhost is a fedora box
- hosts: rhel6-test
  gather_facts: False
  vars:
    device_host: rhel6-test
    top_dir: /var/tmp
  tasks:
     # Expect file only on rhel6-test w/ roan.lan
#    - name: Push fedora file to rhel host
#      synchronize:
#        src: "{{ top_dir }}/host"
#        dest: "{{ top_dir }}/fedora"
#        mode: push
    # Expect file only on localhost w/ rhel6.lan
#    - name: Pull redhat file to fedora host
#      synchronize:
#        src: "{{ top_dir }}/host"
#        dest: "{{ top_dir }}/redhat"
#        mode: pull
    # Expect file only on rhel6-test w/ fedora21
    # (broken file on rhel6-test)
#    - name: Push fedora file to rhel host
#      synchronize:
#        src: "{{ top_dir }}/host"
#        dest: "{{ top_dir }}/fedora21"
#        mode: push
#      delegate_to: fedora21
    # Expect file only on fedora21 w/ rhel6.lan
    # (broken file on fedora21 contains fedora21)
    - name: Pull redhat file to fedora host
      synchronize:
        src: "{{ top_dir }}/host"
        dest: "{{ top_dir }}/redhat6"
        mode: pull
      delegate_to: fedora21
```

I think the failures I noted match with what @ritzk is seeing.  These failures did not exist here: e7b5cb87828a49cfc8771ab3d83fda157d42972f

abadger:
Seems like there could be multiple bugs between then and now.

This revision is the first to break the test cases: 1f5584aa5b9d1142e67fed209bbeea1ca99fc307
When I run the final task, this revision causes the file to end up on the rhel6-test host with the contents rhel6.lan.  That means that this commit causes rsync to run on the inventory_hostname host instead of the delegate_to host.  However, at that point it is contacting the correct remote server (itself) to get the file.

This revision changes the failure mode to what we're seeing at present: 3bd71d06195cdce01a85257545723e221f9c9698
Now the fedora21 host has the redhat6 file but its contents are "fedora21".  This means that it correctly ran on the fedora21 host but it pulled the file from itself instead of from the inventory_hostname host.

tima:
I know exactly what you mean about the difficulty in testing this one @abadger. ;) 

I've been looking in to this one for a bit and something seems amiss, but I'm not spotting it yet.

abadger:
Another thing odd (mentioning it because 3bd71d06195cdce01a85257545723e221f9c9698 touches that code): 

```
TASK [Pull redhat file to fedora host] *****************************************
 [WARNING]: no remote address found for delegated host fedora21, using its name by default
```

But in my inventory file I have:

```
[vms]
fedora20
rhel5
rhel6
rhel7
ubuntu12
ubuntu14
freebsd10
freebsd9
opensuse12
fedora20-fips
rhel7-test
rhel6-test
fedora21
```

lib/ansible/plugins/action/synchronize.py
```diff
@@ -125,7 +125,10 @@ def run(self, tmp=None, task_vars=dict()):
         # running on so localhost could be a non-controller machine if
         # delegate_to is used)
         src_host  = '127.0.0.1'
-        dest_host = task_vars.get('ansible_ssh_host') or task_vars.get('inventory_hostname')
+        if delegate_to:
+            dest_host = task_vars.get('inventory_hostname')
+        else:
+            dest_host = task_vars.get('ansible_ssh_host') or task_vars.get('inventory_hostname')
```
abadger, line 132: This feels wrong because if it matters that we get inventory_hostname when using delegate_to then we wouldn't want to also use inventory_hostname in the non-delegate_to case... that probably means that this happens to make some test cases work but hte real bug is elsewhere (or we shouldn't have inventory_hostname in the else?)  Would need to explore this more to understand what's actually happening here.

lib/ansible/plugins/action/synchronize.py
```diff
@@ -200,14 +203,11 @@ def run(self, tmp=None, task_vars=dict()):
...
-            else:
-                # src is a local path, dest is a remote path: <user>@<host>
-                src = self._process_origin(src_host, src, user)
-                dest = self._process_remote(dest_host, dest, user)
+            set_mode = lambda host, var: \
+                          self._process_origin(host, var, user) if host == '127.0.0.1' \
+                     else self._process_remote(host, var, user)
+            src = set_mode(src_host, src)
+            dest = set_mode(dest_host, dest)
         else:
```
abadger, line 211: This doesn't seem right... It's ignoring the push and pull mode parameter altogether so I think it will break that functionality.

IIRC, ternary also isn't available in python 2.6... just write it out as an if-then-else, it will be easier to understand as well.

abadger, line 211: Ah.. I see that push and pull are used to switch src_host and dest_host earlier in the file.  So is this portion an actual fix or just refactoring?

abadger:
alright, I see why inventory_hostname works and not ansible_ssh_host.  However, the change isn't quite right.  Discussing it with @jimi-c...  What's happening is that since the change, ansible_ssh_host contains the value for whichever host ansible is going to connect to.  In the delegate_to case, this is the delegate_to host.  So as you found, that's not the right value.

However, inventory_hostname is the name that we use in inventory.  This may not be an actual hostname that DNS can resolve on the system.  If you specify an inventory  like:

```
my-made-up-name  ansible_ssh_host=real-host.mydomain.com
```

Then inventory_hostname will contain my-made-up-name.  We need to get real-host.mydomain.com instead.

abadger:
Thanks @ritzk 2tima @jimi-c, that commit should fix this bug.

PR #12757, opened 2015-10-15
Python 3: fix AnsibleError formatting

mgedmin:
If you convert the error string to bytes and embed it inside another error
string, you get

  Prefix:

  b'Embedded\nerror\nstring'

which is not what we want.

But we also don't want Unicode in error messages causing unexpected 
UnicodeEncodeErrors when on Python 2.

So let's convert the error message into the native string type (bytes on 
Python 2, unicode on Python 3).

This fixes two failing tests on Python 3.

### DIFF
--- a/lib/ansible/errors/__init__.py
+++ b/lib/ansible/errors/__init__.py
@@ -20,15 +20,7 @@ from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
 from ansible.errors.yaml_strings import *
-from ansible.utils.unicode import to_unicode, to_bytes
-
-
-if str is bytes:
-    # Python 2
-    to_str = to_bytes
-else:
-    # Python 3
-    to_str = to_unicode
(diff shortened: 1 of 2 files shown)


lib/ansible/errors/__init__.py
```diff
@@ -19,11 +19,18 @@
...
 from ansible.utils.unicode import to_unicode, to_bytes
 
+
+if str is bytes:
+    # Python 2
+    to_str = to_bytes
+else:
+    # Python 3
+    to_str = to_unicode
+
```
abadger, line 24: We should probably put this into ansible.utils.unicode with a note about when to use it and what sort of literals to use it with for those that don't understand the native string distinction.

abadger:
Thanks @mgedmin!  This is going to be useful for getting rid of places where we're presently doing things like this as wel:

```
except Exception as e:
    raise AnsibleError('Warning: Foo raise: %s' % str(e))
```

PR #12767, opened 2015-10-15
Document using the `raw` module to install a py2.X and simplejson

ryansb:
In the installation guide, the raw module is mentioned as an option for installing Python or simplejson on managed nodes that don't have them. This change adds an example for users that may already be familiar with using ansible but are checking install docs because they don't know the requirements for managed nodes, or are using a distribution that doesn't include Python 2 by default.

### DIFF
--- a/docsite/rst/intro_installation.rst
+++ b/docsite/rst/intro_installation.rst
@@ -93,7 +93,7 @@ Also you need Python 2.4 or later, but if you are running less than Python 2.5 o
 
    If you need to bootstrap these remote systems by installing Python 2.X,
    using the 'raw' module will be able to do it remotely. For example,
-   ``ansible myhost --sudo -m raw -a "yum install python2 python-simplejson"``
+   ``ansible myhost --sudo -m raw -a "yum install -y python2 python-simplejson"``
    would install Python 2.X and the simplejson module needed to run ansible and its modules.
 
 .. _installing_the_control_machine:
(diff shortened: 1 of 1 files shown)


docsite/rst/intro_installation.rst
```diff
@@ -89,8 +89,12 @@ Also you need Python 2.4 or later, but if you are running less than Python 2.5 o
    the 'ansible_python_interpreter' variable in inventory (see :doc:`intro_inventory`) to point at your 2.X Python.  Distributions
    like Red Hat Enterprise Linux, CentOS, Fedora, and Ubuntu all have a 2.X interpreter installed
    by default and this does not apply to those distributions.  This is also true of nearly all
-   Unix systems.  If you need to bootstrap these remote systems by installing Python 2.X, 
-   using the 'raw' module will be able to do it remotely.
+   Unix systems.
+
+   If you need to bootstrap these remote systems by installing Python 2.X,
+   using the 'raw' module will be able to do it remotely. For example,
+   ``ansible myhost --sudo -m raw -a "yum install python2 python-simplejson"``
```
abadger: I think there's a -y needed for the yum command.

abadger:
Eventually we may want to create a whole page for this and link to it here as container images are going to start making this a more frequently asked question and have different variants for different OS's.

docsite/rst/intro_installation.rst
ryansb: You think correctly.

ryansb:
Yeah, it's probably worth a full page eventually. I think adding this one-liner for now would cover a lot of users, since they just have to replace "yum" with their package manager of choice.

docschick:
I agree that this could be a separate chapter or section later on, but this works for now. Thanks so much for the docs contribution! We really appreciate your time and effort, @ryansb!

PR #12775, opened 2015-10-16
Python 3: fix vault tests

mgedmin:
This fixes all four remaining unit test failures on Python 3.5 and
enables Python 3.5 tests in Travis.

Python 3.4 still has four failures in vault code due to %-formatting of
byte strings.

Note that passing tests don't mean that Ansible itself can already be
used with Python 3.5: test coverage is only 53%.


abadger:
Thanks @mgedmin!  I'm glad we have a little bit of protection against regressions in python3 compatibility now.

PR #12780, opened 2015-10-16
Python 3: avoid %-formatting of byte strings

mgedmin:
This is needed for Python 3.4 compatibility; Python 3.5 can use
`b'%s\n' bytestring` again.

Also enable unit tests on Python 3.4 in tox and Travis, because they now
pass, yay!

### DIFF
--- a/lib/ansible/parsing/vault/__init__.py
+++ b/lib/ansible/parsing/vault/__init__.py
@@ -194,10 +194,12 @@ class VaultLib:
         if not self.cipher_name:
             raise AnsibleError("the cipher must be set before adding a header")
 
-        tmpdata = [b'%s\n' % b_data[i:i+80] for i in range(0, len(b_data), 80)]
-        tmpdata.insert(0, b'%s;%s;%s\n' % (b_HEADER, self.b_version,
-            to_bytes(self.cipher_name, errors='strict', encoding='utf-8')))
-        tmpdata = b''.join(tmpdata)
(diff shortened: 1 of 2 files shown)


lib/ansible/parsing/vault/__init__.py
```diff
@@ -568,7 +570,7 @@ def encrypt(self, data, password):
 
         # COMBINE SALT, DIGEST AND DATA
         hmac = HMAC.new(key2, cryptedData, SHA256)
-        message = b'%s\n%s\n%s' % (hexlify(salt), to_bytes(hmac.hexdigest()), hexlify(cryptedData))
+        message = b''.join([hexlify(salt), b'\n', to_bytes(hmac.hexdigest()), b'\n', hexlify(cryptedData)])
```
abadger: Can this one be b'\n'.join() ?

abadger:
Personally I like to use tuples rather than lists when I can for efficiency (Many of the join()'s).  But I know it can be more clear to intersperse [] with parenthesis.... up to you if you want to change that one.

mgedmin:
I have a feeling the performance difference is measured in nanoseconds, and I do find `([])` more readable than `join(())`.

lib/ansible/parsing/vault/__init__.py
mgedmin: D'oh, certainly!

abadger:
Thanks @mgedmin !  Break open the champagne ;-)

PR #12787, opened 2015-10-16
keyword not in ubuntu 14.04

chrismeyersfsu:


abadger:
We were testing this on ubuntu 14.04 via jenkins.  is this something that went away recently?

abadger:
Hmm.. okay it's a difference between 1.9 and 2.0.  Let me see if that's an intentional change.

abadger:
Okay, the basic difference is that previously we were unconditionally adding invocation to the data and then stripping it out (also unconditionally?) if we were making any callbacks with the data.  Now we are only adding it if we're running asynchronously (in `action/__init__.py`)

```
         # store the module invocation details back into the result
+        if self._task.async != 0:
+            data['invocation'] = dict(
+               module_args = module_args,
+               module_name = module_name,
```

Not sure yet how we got from one to the other there.  Asking @jimi-c what we to have happen here.

jimi-c:
I believe invocation was always in the output, and was simply stripped out before, but this really has no bearing on the functionality of the test so it can/should be removed.

Merged, and thanks!

PR #12790, opened 2015-10-16
update checksum, make fail case fail

chrismeyersfsu:


bcoca:
@chrismeyersfsu info?

chrismeyersfsu:
@bcoca 
1. Updated dated/invalid checksum for `ansible_test_service` scripts.
2. There is an upstart broken test case that relies on a broken upstart script. From what I could tell (and from trying), the `ansible-broken.upstart` was not broken. So I broke it and updated the corresponding checksum.

chrismeyersfsu:
Thanks @abadger for making me look into this further. The `ansible-broken.upstart` invokes a failure, as it should.

This PR included a fix that has been merged in https://github.com/ansible/ansible/pull/12950

PR #12797, opened 2015-10-17
1st part of ansible config, adds ansible-config to view/manage configs

bcoca:
Start of ansible config project

moved configuration definitions to external yaml file vs hardcoded
  *  updated constants to be a data strcutures that are looped over and also return origin of settings

changed to manager/data scheme for base classes
    new cli ansible-config to view/manage ansible configuration settings
  *  prints green for default/unchanged and yellow for those that have been overriden
  *  added list action to show all configurable settings and their associated ini and env var names
  *  allows specifying config file to see what result would look like
  *  TBD update, edit and view options

removed test for functions that have been removed

right now it can only list or  dump 'current' settings


chrismeyersfsu:
:+1:  for this feature

gundalow:
Closing & reopening to trigger CI

lib/ansible/cli/__init__.py
```diff
@@ -272,20 +272,13 @@ def expand_tilde(option, opt, value, parser):
...
-        """optparse action callback to convert a PATH style string arg to a list of path strings.
-
-        For ex, cli arg of '-p /blip/foo:/foo/bar' would be split on the
-        default os.pathsep and the option value would be set to
-        the list ['/blip/foo', '/foo/bar']. Each path string in the list
-        will also have '~/' values expand via os.path.expanduser()."""
-        path_entries = value.split(os.pathsep)
-        expanded_path_entries = [os.path.expanduser(path_entry) for path_entry in path_entries]
-        setattr(parser.values, option.dest, expanded_path_entries)
+    def unfrack_path(option, opt, value, parser):
```
abadger, line 289: This seems like it would break galaxy which needed expand_paths

abadger: Never mind, I now see the change to galaxy further down.

lib/ansible/cli/config.py
```diff
@@ -0,0 +1,181 @@
...
+import traceback
+
+from ansible.cli import CLI
+from ansible.config.data import Setting
+from ansible.config.manager import ConfigManager
+from ansible.errors import AnsibleError, AnsibleOptionsError
+from ansible.parsing.yaml.dumper import AnsibleDumper
+from ansible.utils.color import stringc
+from ansible.utils.path import unfrackpath
+from ansible.utils.unicode import to_str
```
abadger: Deprecated location for this import and looks like it's unused anyway.

bcoca: wrote the initial version of this loooong time ago, will clean it up

lib/ansible/cli/config.py
```diff
@@ -0,0 +1,181 @@
...
+            '-c','local',
+            '-a','"dest=%s section=%s option=%s value=%s backup=yes"' % (self.config_file, section, option, value)
+        ])
+
+    def execute_view(self):
+        '''
+        Displays the current config file
+        '''
+        try:
+            self.pager(open(self.config_file).read())
```
abadger: This might cause problems with non-ascii but I'm not 100% certain.  On python2, the read() will return a byte string.  On python3, will use the encoding specified via locale (falling back to utf-8 I believe) to decode the bytes into text.

self.pager call display.display() which has a comment that the strings it gets must be text otherwise we'll get tracebacks (That comment seems odd as ansible-doc already calls it with byte strings on python2 I think.  But maybe those are getting coerced on python2 into text or we just don't have any non-ascii chars in the module docs.)

lib/ansible/cli/config.py
```diff
@@ -0,0 +1,179 @@
...
+        except Exception as e:
+            raise AnsibleError("Failed to open config file: %s" % str(e))
```
abadger: to_native(e) is better :-)

lib/ansible/cli/config.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def execute_edit(self):
+        '''
+        Opens ansible.cfg in the default EDITOR
+        '''
+        raise AnsibleError("Option not implemented yet")
+        try:
+            editor = shlex.split(os.environ.get('EDITOR','vi'))
+            editor.append(self.config_file)
+            subprocess.call(editor)
```
abadger, line 149: Note, anytime we use shlex.split, we have to do a song and dance of converting to byte strings on python2 and text strings on python3. (The reason is that py3 only handles text string input and python2.6 has a bug when it gets text strings as input).  For perfect resilience we should give subprocess.call() a byte string as well (probably doesn't matter as commands probably don't exist in paths with non-ascii parts).

something like:
``` python
        try:
            editor_command = to_native(os.environ.get('EDITOR', 'vi'))
            b_editor = shlex.split(editor_command)
            if PY3:
                b_editor = [to_bytes(elem, errors='surrogate_or_strict') for elem in b_editor] 
            b_editor.append(to_bytes(self.config_file, errors='surrogate_or_strict'))
            subprocess.call(b_editor)

File: lib/ansible/cli/config.py [hunk 6]
```diff
@@ -0,0 +1,179 @@
...
+        except Exception as e:
+            raise AnsibleError("Failed to open editor: %s" % str(e))
```
abadger: to_native better again  (The reason being that to_native won't traceback if str(e) yields non-ascii characters.  str(e) might traceback in that case if the locale doesn't allow those chars).

lib/ansible/cli/config.py
bcoca: yeah, this part is WIP, see 'raise not implemented' preceding it

lib/ansible/cli/config.py
bcoca: ` self.pager(open(self.config_file, 'rb').read())` ?

lib/ansible/cli/config.py
bcoca: done

lib/ansible/cli/config.py
```diff
@@ -0,0 +1,180 @@
...
+            editor = shlex.split(os.environ.get('EDITOR','vi'))
+            editor.append(self.config_file)
+            subprocess.call(editor)
+        except Exception as e:
+            raise AnsibleError("Failed to open editor: %s" % to_native(e))
+
+    def execute_list(self):
+        '''
+        list all current configs reading lib/constants.py and shows env and config file setting names
+        '''
```
abadger, line 156: Similar to earlier pager question, I'm not sure this is resistant to tracebacks when there's non-ascii characters.  Stress test with nonascii chars and a locale setting that can't handle non-ascii to see what happens.

abadger: ... On both python2 and python3

test/sanity/pep8/legacy-files.txt
```diff
@@ -1,3 +1,8 @@
+hacking/conf2yaml.py
+lib/ansible/constants.py
+lib/ansible/cli/config.py
+lib/ansible/config/data.py
+lib/ansible/config/manager.py
```
abadger, line 5: Uhm.... This isn't the point of the test blacklist file.  Cheater!

bcoca: sssh!

bcoca: most of the code was moved from constants,py which was under it (someone refactored and removed it while this PR was in flight, so that is why it looks like i re added constants.py)

lib/ansible/cli/config.py
```diff
@@ -0,0 +1,180 @@
...
+            '-c','local',
+            '-a','"dest=%s section=%s option=%s value=%s backup=yes"' % (self.config_file, section, option, value)
+        ])
+
+    def execute_view(self):
+        '''
+        Displays the current config file
+        '''
+        try:
+            self.pager(open(self.config_file, 'rb').read())
```
abadger: Did you test this?  I'm thinking we might need a to_text() around that read().  Not sure if the errors argument should then be 'surrogate_or_strict' or 'surrogate_then_replace' for that, though).

bcoca: worked w/o issues, but source file has to be valid YAML

bcoca: ah, nvmd wrong action

lib/ansible/config/manager.py
```diff
@@ -0,0 +1,242 @@
...
+
+        parser = None
+        if ftype == 'ini':
+            parser = configparser.ConfigParser()
+            try:
+                parser.read(cfile)
+            except configparser.Error as e:
+                raise AnsibleOptionsError("Error reading config file: \n{0}".format(e))
+        elif ftype == 'yaml':
+            with open(to_bytes(cfile, 'r')) as config_stream:
```
abadger: to get bytes from a file, always do this instead:
``` python
with open(cfile, 'rb') as config_stream:
```

lib/ansible/cli/config.py
bcoca: but yes, tested and mostly works (when specifying file, need to fix to also use default file)

sivel: Shouldn't we also switch to using `with open` here, to ensure the file is properly closed.  Additionally we should be more selective with what exceptions we are expecting here.  Such as `IOError`

lib/ansible/config/manager.py
bcoca: should we always read yaml files as binary?

sivel: Not sure, but looks like the `to_bytes` is for the filename, not the contents of the file.  But adding `rb` would probably be beneficial.

test/sanity/pep8/legacy-files.txt
abadger: Err.. which means that you need to pull those changes into your PR, no?  just like you'd pull a logic bug fix into your PR rather than overwriting it with the old code in different files.

lib/ansible/config/manager.py
sivel: I think the parens are wrong here anyway though.  the `'r'` shouldn't be within the `to_bytes`

hacking/conf2yaml.py
```diff
@@ -0,0 +1,124 @@
...
+
+op_map = {
+    ast.Add: '+',
+    ast.Sub: '-',
+    ast.Mult: '*',
+    ast.Div: '/',
+}
+
+
+
```
sivel, line 20: Too many spaces here

test/sanity/pep8/legacy-files.txt
sivel: I strongly recommend not doing this.  We're trying to clean it up anyway, no sense in just kicking the can down the road.

hacking/conf2yaml.py
bcoca: we can delete this file

sivel: That's fine, I didn't realize it was what I wrote until after I got further into the file.

test/sanity/pep8/legacy-files.txt
bcoca: the merge was too hard to do once the files were split, since the changes were pure style i just ignored them.

bcoca: they also had errors (now fixed in devel)

lib/ansible/cli/config.py
bcoca: i can list several possible exceptions there (OSError is also likely) but since this is very specific action i chose the generic ... if we are going to split this and add more code, i would be more specific on exception

hacking/conf2yaml.py
bcoca: hehe, yes, thanks for that, i had my own but yours was better, so i just repurposed  it, no need to keep this after the merge, was just htere to allow me to 'migrate multiple times'

lib/ansible/cli/config.py
abadger: This is synthetic code but I get an error... so I think this code can definitely error as well:
``` python
$ cat test.py

#!/usr/bin/python2 -tt

from ansible.utils.display import Display
from ansible.cli.galaxy import GalaxyCLI

c = GalaxyCLI(['test'])
#c.pager(u'caf\xe9')
c.pager(u'caf\xe9'.encode('utf-8'))

$ ./test.py >test.log
  File "./test.py", line 8, in <module>
    c.pager(u'caf\xe9'.encode('utf-8'))
  File "/usr/lib/python2.7/site-packages/ansible/cli/__init__.py", line 583, in pager
    display.display(text)
  File "/usr/lib/python2.7/site-packages/ansible/utils/display.py", line 113, in display
    if not msg.endswith(u'\n'):
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 3: ordinal not in range(128)
```

abadger: Combining the fix for that with sivel's with suggestion:
``` python
with open(self.config_file, 'rb') as f:
    self.pager(to_text(f.read(), errors='surrogate_or_strict'))
```

lib/ansible/cli/config.py
abadger: Found that the above use of pager can yield tracebacks so need to fix it here too.

test/sanity/pep8/legacy-files.txt
abadger: I agree with sivel, these files need to be fixed rather than adding them to the testing blacklist.

kumaran85:
Hi In ansible version 2.3.1.0 installed on RHEL 7. grep inventory /etc/ansible/ansible.cfg
#inventory       = /etc/ansible/hosts
inventory       = /home/ec2-user/book/inventor
my playbook:
---
- hosts: route
- become: true
- become_method: sudo
- become_user: ec2-user
  tasks:
  - name: create file
    file: touch=/root/test state=touch mode=0600
After that I added my instance IP added to (home/ec2-user/book/inventor) inventory. not created any dynamic inventory. As a OS environment using this.

If I do, ansible-playbook test.yml --syntax-check
ERROR! the field 'hosts' is required but was not set
I used playbook on VM and server level with 1.9.  But here not working for me? Please suggest me

PR #12798, opened 2015-10-17
allow invoking clis through keywords

bcoca:
allows for invoking `ansible-playbook` as `ansible playbook`, etc

### DIFF
--- a/bin/ansible
+++ b/bin/ansible
@@ -35,6 +35,7 @@ except Exception:
 import os
 import sys
 import traceback
+import glob
 
 from ansible.errors import AnsibleError, AnsibleOptionsError, AnsibleParserError
 from ansible.utils.display import Display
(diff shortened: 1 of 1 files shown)


sivel:
What happens if you have host called "playbook", "vault", "doc", etc...?

Normal running of the Ansible command allows use of a positional arg which is the hostname.

```
ansible playbook -i hosts -m ping
```

The above in the past would allow you to run against a host called "playbook".  This change introduces a reserved list of hosts/limits at this point which could be unexpected

Also the use of `.find` isn't very pythonic.

bcoca:
Yes, this will break adhoc with hosts named doc, playbook and vault (or any other new command line names), the find i can change to `'-' in ...`.

tima:
+1. More of this please.

bcoca:
# Question About This One

Hi!

PR #12800, opened 2015-10-17
remove invocation keyword check

chrismeyersfsu:
Same as https://github.com/ansible/ansible/pull/12787 but for yum instead of apt


jimi-c:
Merged, thanks!

PR #12804, opened 2015-10-17
allow env variable override of creds file

chrismeyersfsu:
This is important because there is an integration test, test_git, that
require an ssh key to clone a privileged github repo.


jimi-c:
Merged, thanks!

PR #12806, opened 2015-10-17
added common user function to get global unique SID (without include statement)

h0nIg:
added common user function to get global unique SID, since there are various parts where a global user object is needed

### DIFF
--- a/lib/ansible/module_utils/powershell.ps1
+++ b/lib/ansible/module_utils/powershell.ps1
@@ -232,54 +232,51 @@ Function Get-FileChecksum($path)
 # Example: $sid = GetAccountSID "user@domain.com"
 #          $objUser = New-Object System.Security.Principal.SecurityIdentifier($sid)
 #          $fullUsername = $objUser.Translate([System.Security.Principal.NTAccount]).Value
-Function GetAccountSID
+Function Get-AccountSid
 {
-    Param ([string]$AccountName)
-
+    Param ([string]$accountName)
     #Check if there's a realm specified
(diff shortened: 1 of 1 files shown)


marcind:
How many modules are expected to use this function? I ask because powershell.ps1 file gets included in every windows module execution. The larger it is, the slower the execution due to the increased copy time. I'm not sure how big the impact would be, I'm just trying to warn against a slippery slope where powershell.ps1 becomes a kitchen sink of any and every helper function.

h0nIg:
this function should be used on every module that may set file ownership or ACL, since the result will be a SID that can easily be converted into the wanted format.

regarding the file size, this shouldn't be a problem if https://github.com/ansible/ansible-modules-core/issues/1540#issuecomment-144771093 will fixed

trondhindenes:
I think we should generally stick to approved verbs for common functions. (`Get-AccountSid`)

daBONDi:
I think a forced way to get SIDs for the Modules would be awesome, because current Modules use differnet Methods, WMI, .Net Call and also ADSI(WTF).

**I think the only way to be future Proof is with a .Net Object like New-Object System.Security.Principal.NTAccount($accountName)**

It should also get rid of difference between Local and Domain stuff and different writing Style "user@domainfqdn, and domain\user", and it would then used "Proven" Code from MS that also get patched, and is used by Tons of .Net Programs  :-)

Can also get used to get a Group SID, so there is no thing to differentiat it.

jborean93:
Closing as this has been merged in with https://github.com/ansible/ansible/pull/27091.

PR #12815, opened 2015-10-18
Update YAMLSyntax.rst

FloorLamp:
add example for multiline yaml items


jimi-c:
There are actually two situations here, which need to be documented if we do one. Both `|` and `>` are block, however `|` is a "literal block", in which all new lines are kept while `>` is a "folded block", in which only the last newline is kept. See here for more details: http://www.yaml.org/spec/1.2/spec.html#id2795688

PR #12822, opened 2015-10-19
added repository for openSUSE ansible rpm

rensgr:
added repository and instruction how to add it to suse yast2 management tool / repo manager.
to simplify installation for opensuse users.


rensgr:
hmm, do I need to explain again ? 
I added method to install for opensuse via RPM repository @ OBS

PR #12827, opened 2015-10-19
ec2.py Add the feature of combining Autoscaling groups Tags

jordiclariana:
If there're Autoscaling groups, fetch their tags set as propagate_at_launch and combine with the tags of these Autoscaling groups.

This is specially useful for this: https://forums.aws.amazon.com/thread.jspa?messageID=588978&#588978

So you can provision EC2 instances launched from Autoscaling groups with CodeDeploy, **before** the CodeDeploy, using the tags from the Autoscaling group (as the Ansible Dynamic Inventory requires: http://docs.ansible.com/ansible/intro_dynamic_inventory.html#example-aws-ec2-external-inventory-script)

### DIFF
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -666,29 +666,30 @@ class Ec2Inventory(object):
                 reservations = []
                 autoscale_groups = conn_autoscale.get_all_groups()
                 for autoscale_group in autoscale_groups:
-                  tags = {}
-                  for tag in autoscale_group.tags:
-                    if tag.propagate_at_launch:
-                        tags[tag.key] = tag.value
+                  if autoscale_group.name:
+                    tags = {}
+                    for tag in autoscale_group.tags:
+                      if tag.propagate_at_launch:
+                          tags[tag.key] = tag.value
 
-                  for autoscale_group_instance in autoscale_group.instances:
+                    for autoscale_group_instance in autoscale_group.instances:
 
-                    if self.ec2_instance_filters:
-                        for filter_key, filter_values in self.ec2_instance_filters.items():
-                            reservations.extend(conn_ec2.get_all_instances(
-                                filters = {
-                                    filter_key : filter_values,
-                                    'instance-id': [ autoscale_group_instance.instance_id ]
-                                }
-                            ))
-                    else:
-                        reservations = conn_ec2.get_all_instances(
-                            filters = { 'instance-id': [ autoscale_group_instance.instance_id ] })
+                      if self.ec2_instance_filters:
+                          for filter_key, filter_values in self.ec2_instance_filters.items():
+                              reservations.extend(conn_ec2.get_all_instances(
+                                  filters = {
+                                      filter_key : filter_values,
+                                      'instance-id': [ autoscale_group_instance.instance_id ]
+                                  }
+                              ))
+                      else:
+                          reservations = conn_ec2.get_all_instances(
+                              filters = { 'instance-id': [ autoscale_group_instance.instance_id ] })
 
-                    for reservation in reservations:
-                      for reservation_instance in reservation.instances:
-                        reservation_instance.tags = tags
-                        self.add_instance(reservation_instance, region)
+                      for reservation in reservations:
+                        for reservation_instance in reservation.instances:
+                          reservation_instance.tags = tags
+                          self.add_instance(reservation_instance, region)
 
         except boto.exception.BotoServerError as e:
             if e.error_code == 'AuthFailure':
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -37,7 +37,6 @@ When run against a specific host, this script returns the following variables:
  - ec2_attachTime
  - ec2_attachment
  - ec2_attachmentId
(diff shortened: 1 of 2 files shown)


bcoca:
I would not do this by default, but make it conditional on a setting in the ini file (off by default)

jordiclariana:
I agree with you. Luckily I already implemented that so I just changed the default to False as required.

jordiclariana:
Hi there! Will you consider to include this change? It's been a while since I proposed it.

bcoca:
lgtm, cc @defionscode  for final look

contrib/inventory/ec2.py
```diff
@@ -592,65 +584,6 @@ def get_rds_instances_by_region(self, region):
                 error = "Looks like AWS RDS is down:\n%s" % e.message
             self.fail_with_error(error, 'getting RDS instances')
 
-    def include_rds_clusters_by_region(self, region):
```
ryansb: Looks like this merge will remove a different feature in ec2.py - can you fix this?

contrib/inventory/ec2.py
```diff
@@ -721,6 +654,51 @@ def get_elasticache_replication_groups_by_region(self, region):
...
+            conn_autoscale = self.connect_to_aws(autoscale, region)
+            conn_ec2 = self.connect_to_aws(ec2, region)
+            if conn_autoscale and conn_ec2:
+                reservations = []
+                autoscale_groups = conn_autoscale.get_all_groups()
+                for autoscale_group in autoscale_groups:
+                  if autoscale_group.name:
+                    tags = {}
+                    for tag in autoscale_group.tags:
+                      if tag.propagate_at_launch:
```
ryansb: This code mixes 2 and 4-space tabs, which isn't in line with the style guide (pretty much PEP8) that we use.

ryansb (changes requested):

jordiclariana:
@ryansb all problems fixed.

ryansb (approved):

willthames:
!needs_revision

contrib/inventory/ec2.py
```diff
@@ -286,6 +286,11 @@ def read_settings(self):
         if config.has_option('ec2', 'elasticache'):
             self.elasticache_enabled = config.getboolean('ec2', 'elasticache')
 
+        # Combine with Autoscaling groups tags?
+        self.autoscaling_enabled = True
```
willthames: should this be False by default (it's False by default in the ini file, but if someone uses an old ini file with the updated script, odd things might happen)

contrib/inventory/ec2.py
```diff
@@ -721,6 +728,51 @@ def get_elasticache_replication_groups_by_region(self, region):
         for replication_group in replication_groups:
             self.add_elasticache_replication_group(replication_group, region)
 
+    def get_autoscaling_instances_by_region(self, region):
+        ''' Makes an AWS API call to the list of Autoscale instances
+         in a particular region with tags of their Autoscale group
+         if the "propagate_at_launch" is set '''
```
willthames, line 734: This seems a roundabout way of saying 'tags of the instances in the autoscaling group' (tags without  propagate_at_launch won't be set on the instances, but that seems an unnecessary aside)

willthames:
Last comment in the linked forum thread suggests this might no longer be required? (If it is still required, then I'm happy enough for this to be included)

jordiclariana:
If it is fixed on the AWS side or not I don't know, haven't have the opportunity to test it (I don't use CodeDeploy anymore). For me to set up the environment to test this would be too much time consuming now.

I'll fix the comments you made above about the code, and it'll be up to you to merge it or not. I agree with you that if AWS fixed the original problem, this PR makes no sense anymore.

contrib/inventory/ec2.py
jordiclariana: I think is a good definition of what is does, as it filters for `propagate_at_launch` set tags:
```
if tag.propagate_at_launch:
```
I wouldn't change it.

contrib/inventory/ec2.py
```diff
@@ -721,6 +728,51 @@ def get_elasticache_replication_groups_by_region(self, region):
...
+                                        }
+                                    ))
+                            else:
+                                reservations = conn_ec2.get_all_instances(
+                                    filters = { 'instance-id': [ autoscale_group_instance.instance_id ] })
+
+                            for reservation in reservations:
+                                for reservation_instance in reservation.instances:
+                                  reservation_instance.tags = tags
+                                  self.add_instance(reservation_instance, region)
```
mattclay: CI failure due to PEP 8 indentation issue:

```
2017-02-17 09:19:55 ERROR: PEP 8: contrib/inventory/ec2.py:792:35: E111 indentation is not a multiple of four (current)
2017-02-17 09:19:55 ERROR: PEP 8: contrib/inventory/ec2.py:793:35: E111 indentation is not a multiple of four (current)
```

You can run this test locally with `make pep8`.

jordiclariana: Fixed

ryansb:
This has been fixed on AWS' side, so we don't need this workaround.

PR #12835, opened 2015-10-20
DOC: in yum examples, 'pkg' arg renamed to 'name'

zarnovican:
- alternating both pkg/name in examples might be confusing
- even if pkg= is working, it is not documented in
  module's refrence guide


bcoca:
it is documented in aliases, but name is preferred so I'm merging this

PR #12842, opened 2015-10-20
Revert back to copying self._task.args

mcsalgado:
Reverts part of 892d9b8d


bcoca:
the typo needs no explanation, but the revert does

jimi-c:
Merged, thanks!

mcsalgado:
@bcoca I was discussing this with @jimi-c before submitting this so I kind of omitted the explanation. But anyway the motivation was to fix include's with_items.

PR #12847, opened 2015-10-21
Remove mutable default arguments.

godfreyduke:


abadger:
Thanks for catching those problems!

PR #12853, opened 2015-10-21
In v2, exec_command should return bytes and the caller will take responsibility for converting to unicode

abadger:
At the moment, plugins/action/**init**.py:_low_level_exec_command(), which is the only caller of exec_command() should be able to handle returning unicode instead of bytes but we should match with the other connection plugins so that future code doesn't accidentally break this.


cchurch:
:+1:

Integration tests run with `TEST_FLAGS="-t test_win_raw,test_win_script,test_win_ping -vvv" make test_winrm` pass against my test 2008R2 instance.

PR #12855, opened 2015-10-21
make - replacement in group names optional

bcoca:
so people can control this themselves instead of toggling the hardcoded default
alternate fix of #11165


cchurch:
:+1: LGTM

Issue #12856, opened 2015-10-21
Ansible 2.0 breaks "here" documents in shell scripts


larsks:
I wanted to embed a "here" document (e.g. `cat <<EOF ...`) in a `shell:` script in my playbook, like this:

```
shell: |
  . credentials/keystonerc_admin

  cat <<-EOF
  auth_url: "$OS_AUTH_URL"
  username: "$OS_USERNAME"
  password: "$OS_PASSWORD"
  tenant_name: "$OS_TENANT_NAME"
  EOF
```

In 1.9.4, this breaks because the magic processing that looks for
module arguments in the command string resulted in the following
script:

```
. credentials/keystonerc_admin

 cat <<-EOF
 auth_url: \"$OS_AUTH_URL\"
 username: \"$OS_USERNAME\"
 password: \"$OS_PASSWORD\"
 tenant_name: \"$OS_TENANT_NAME\"
 EOF
```

Note the extra space (` `) at the beginning of every line, which
breaks shell parsing.  It was possible to fix this by passing the
script directly to the`cmd` parameter, which would bypass the magic
"look for arguments" processing:

```
shell:
  cmd: |
    . credentials/keystonerc_admin

    cat <<-EOF
    auth_url: "$OS_AUTH_URL"
    username: "$OS_USERNAME"
    password: "$OS_PASSWORD"
    tenant_name: "$OS_TENANT_NAME"
    EOF
```

This worked great.  But then Ansible 2.0 came along, and the above
fails with:

```
unsupported parameter for module: cmd
```

And without the `cmd:` parameter, it fails the same was as in 1.9.4.

larsks:
So, looking through the code, it looks a possible source of the problem is the parsing code in `lib/ansible/parsing/splitter.py`, which uses `split_args` to split up the arguments and then re-assembles them with `' '.join(...)`.  This seems like the wrong thing to do, especially when receiving a YAML literal block.  There are a few ways of fixing this:
- Add back the `cmd` parameter, for providing an argument that doesn't get processed by `splitter.py`.
- If no keyword arguments are found, return the argument string verbatim rather than splitting/re-assembling it
- I had a third idea, but it just slipped my mind.

Any preferences?

jimi-c:
Hi @larsks, in 2.0 you can actually use `_raw_params` for the `cmd` parameter, which works fine in my testing. I can add some code to the `preprocess_data()` method for Task to convert this automatically internally.

larsks:
[...]

PR #12857, opened 2015-10-21
improve handling of embedded shell scripts

larsks:
This commit addresses #12856, in which shell scripts passed to the
`shell` module are corrupted to the split-and-rejoin approach taken in
`lib/anslibe/parsing/splitter.py`.

This commit modifies the behavior of `parse_kv` such that if there are
no keyword arguments the raw arguments are returned verbatim, rather
than going through the split-and-rejoin process.


jimi-c:
Fixed alternatively in 2b3c5aa41ed135d4ba0f125670ca4bd048149679, so closing this. Thanks for the contribution!

PR #12861, opened 2015-10-22
Allow meta/main.yml to include roles through spec file

### LINKED ISSUE #11486: make test_galaxy fails
A lot of code that was at the heart of ansible-galaxy has been ripped out and not replaced. 

This means that make test_galaxy trivially fails

We need something to replace ansible.utils.role_yaml_parse and ansible.utils.role_spec_parse, and make test_galaxy should pass its tests (those tests exist for a reason!)

willthames:
ansible-playbook now works when run with a playbook
that includes a role that includes another role
specified using csv format

Updated one of the roles used in the tests to fix
broken tests - `make test_galaxy` now works

Fixes #11486. Also addresses the problem alluded to in #10620.


semifocused:
@bcoca - based on the ongoing dialogue

PR #12867, opened 2015-10-22
make callbacks declare needing whitelisting

bcoca:
this will allow for custom or v1 callbacks locally loaded to automatically be triggered as they were in 1.x

### DIFF
--- a/lib/ansible/executor/task_queue_manager.py
+++ b/lib/ansible/executor/task_queue_manager.py
@@ -150,12 +150,13 @@ class TaskQueueManager:
                 # the name of the current plugin and type to see if we need to skip
                 # loading this callback plugin
                 callback_type = getattr(callback_plugin, 'CALLBACK_TYPE', None)
+                callback_needs_whitelist  = getattr(callback_plugin, 'CALLBACK_NEEDS_WHITELIST', False)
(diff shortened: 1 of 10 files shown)


larsks:
I keep wondering if this special treatment of the "core" callback plugins isn't a little baroque.

One argument would be that maybe people want to maintain common libraries of callback plugins, so this behavior should be required regardless of where the plugins are located, and there will always be two configuration options: where to look for plugins, and which plugins to load.  This would allow people to collect callback plugins in a central location, rather than including them per-project, and enable/disable them as necessary.

But of course if a local site wants to implement something like that, they could just set `CALLBACK_NEEDS_WHITELIST` in all their locally developed plugins.  But this wouldn't work as well for someone consuming third-party plugins.

I don't know.  It just strikes me as a little hinky, but I'm not sure I can put my finger on it.

bcoca:
I'm not happy with it either, the purpose was just to be able to ship all included plugins 'ready to use' with and config change. Since users have the ability to configure the directories (change/add) that ansible uses by default for plugins it is hard to distinguish between shipped and custom plugins.

I don't want to restrict the ability of users just 'dropping' a plugin in a dir to enable them (as they can do up to now with every plugin type), just not to enable all plugins we ship by default.

PR #12880, opened 2015-10-23
Amazon cloudformation stack resource and output lookup

viper233:
Thanks to @deanwilson for the initial work and @mschurenko for the great work they did on extending this plugin.

I'm open to guidance on improvements before this can be merged.

Hopefully others will find this useful.

### DIFF
--- a/lib/ansible/plugins/lookup/cf_output.py
+++ b/lib/ansible/plugins/lookup/cf_output.py
@@ -47,9 +47,7 @@ class LookupModule(LookupBase):
     def get_regions(self):
         regions_cache = os.path.join(self.cache_dir, 'regions')
         regions = self.check_cache(regions_cache)
-        if regions:
-            pass
-        else:
+        if not regions:
             try:
                 regions = boto.ec2.regions()
                 regions = [ r.name for r in regions ]
@@ -63,9 +61,7 @@ class LookupModule(LookupBase):
     def get_stack_info(self, region, stack_name):
         stack_cache = os.path.join(self.cache_dir, region + '-' + stack_name)
         outputs = self.check_cache(stack_cache)
-        if outputs:
-          pass
-        else:
+        if not outputs:
           try:
             conn = boto.cloudformation.connect_to_region(region)
             stack = conn.describe_stacks(stack_name_or_id=stack_name)[0]
@@ -73,13 +69,13 @@ class LookupModule(LookupBase):
             outputs = stack.outputs
             pickle.dump(outputs, fh)
           except:
-            outputs = ''
+            outputs = []
 
         return outputs
 
     def run(self, terms, inject=None, **kwargs):
         if not os.path.isdir(self.cache_dir):
-            os.mkdir(self.cache_dir)
+            os.makedirs(self.cache_dir)
 
         regions = self.get_regions()
 
--- a/lib/ansible/plugins/lookup/cf_output.py
+++ b/lib/ansible/plugins/lookup/cf_output.py
@@ -1,0 +1,115 @@
+# (c) 2015, Stephen Granger <stephen(at)trinimbus.com>
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+from ansible.errors import AnsibleError
+from ansible.plugins.lookup import LookupBase
+import boto.ec2
+import boto.cloudformation
+import os
+import sys
+import time
+import pickle
+
+# region/stack/param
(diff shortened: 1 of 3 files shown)


lib/ansible/plugins/lookup/cf_output.py
```diff
@@ -0,0 +1,115 @@
...
+import os
+import sys
+import time
+import pickle
+
+# region/stack/param
+class LookupModule(LookupBase):
+    def __init__(self, basedir=None, **kwargs):
+        self.basedir = basedir
+        self.cache_dir = os.path.join(os.environ['HOME'],'.stack_outputs')
```
chuanger3: Safer to use os.getenv('HOME') because if 'HOME' doesn't exist, this will error out. With os.getenv, you can pass a default value as well if 'HOME' doesn't exist, otherwise it'll return None. In this particular case, you need to give it a default or the join will fail. So perhaps something like os.getenv('HOME', ''), which will return you an empty string if 'HOME' doesn't exist.

lib/ansible/plugins/lookup/cf_output.py
```diff
@@ -0,0 +1,115 @@
...
+        self.cache_time = 60
```
chuanger3: If you set this to a float, (self.cache_time = 60.0), you won't need to cast your time to int.

lib/ansible/plugins/lookup/cf_output.py
```diff
@@ -0,0 +1,115 @@
...
+            if ( now - int(os.path.getmtime(file)) ) < self.cache_time:
+                fh = open(file, 'r')
+                data = pickle.load(fh)
+
+        return data
+
+    def get_regions(self):
+        regions_cache = os.path.join(self.cache_dir, 'regions')
+        regions = self.check_cache(regions_cache)
+        if regions:
```
chuanger3: Rather than passing if something exist, you could check for it not existing
if not regions:
   do something

lib/ansible/plugins/lookup/cf_output.py
```diff
@@ -0,0 +1,115 @@
...
+            outputs = stack.outputs
+            pickle.dump(outputs, fh)
+          except:
+            outputs = ''
+
+        return outputs
+
+    def run(self, terms, inject=None, **kwargs):
+        if not os.path.isdir(self.cache_dir):
+            os.mkdir(self.cache_dir)
```
chuanger3: In case you want to mimic "mkdir -p", you can use os.makedirs(self.cache_dir)

lib/ansible/plugins/lookup/cf_output.py
```diff
@@ -0,0 +1,115 @@
...
+            except:
+                raise AnsibleError('Couldn\'t retrieve aws regions')
+
+        return regions
+
+    def get_stack_info(self, region, stack_name):
+        stack_cache = os.path.join(self.cache_dir, region + '-' + stack_name)
+        outputs = self.check_cache(stack_cache)
+        if outputs:
+          pass
```
chuanger3: Same as above comment for regions. Check for it not existing.
if not outputs:
   do something

lib/ansible/plugins/lookup/cf_output.py
```diff
@@ -0,0 +1,115 @@
...
+                  raise AnsibleError('%s is not a valid aws region' % region)
+              stack_name = args[0]
+              keys = args[1:]
+          else:
+              raise AnsibleError('aws region not found in argument or AWS_REGION env var')
+
+        stack_outputs = self.get_stack_info(region, stack_name)
+        outputs = []
+
+        if stack_outputs:
```
chuanger3, line 104: self.get_stack_info returns a string. I'm not sure the following for loop will work as you expect. In particular, the obj.key and obj.value

lib/ansible/plugins/lookup/cf_resource.py
```diff
@@ -0,0 +1,135 @@
...
+                keys = args[1:]
+            else:
+                raise AnsibleError('aws region not found in argument or AWS_REGION env var')
+
+        resources = self.get_stack_info(region, stack_name)
+        outputs = self.get_stack_outputs(region, stack_name)
+        returns = []
+
+        if resources:
+            resources = sorted(resources, key=lambda x: x.logical_resource_id)
```
chuanger3, line 117: self.get_stack_info returns a string. I don't think this will work as expected. You probably want it to return a list.

tbhi:
Just my two cents as an interested user; I've finding that pickle cache contains too much data, like aws credentials.  Perhaps json would be possible, conversion to a dict then json seems quite doable with something like this:

```
def boto_to_dict(boto_list_type):
    return dict([(x.key, x.value) for x in boto_list_type])
```

Also I think we should be pushing to work with boto3 (which works with python 2) since that seems to be the new stable.

tbhi:
Also it would be nice to have access to both stack.outputs and stack.parameters.

viper233:
Not sure why this is failing. I don't have a lot of experience with python though the format of stack.outputs is similar to stack.parameters to I've implemented cf_parameters in the same way.

I'll implement the pickle changes after resolving this issue.

TaiSHiNet:
Failure doesn't look like its due the plugin.
I'm testing this on a live environment and works like a dream.

Although it is missing the entire documentation, had to guess how to use it from reading the code

TaiSHiNet:
@viper233 Please add some minor usage documentation and push so tests are re-run.

viper233:
Will update this weekend, thanks for the feedback!

viper233:
This lookup has limited functionality when compared to the [ cloudformation_facts module](http://docs.ansible.com/ansible/cloudformation_facts_module.html).

Particular the ability to do cross account authenticated lookups. For this reason I'm closing this PR.

Issue #11885, opened 2015-08-06
lookup + JSON template = unexpected type error


leedm777:
##### Summary:

When attempting to use `lookup('template', 'something.json.j2')`, I'm getting an `unexpected type error`. If I change the contents of `something.json.j2` to invalid JSON, the lookup works successfully.

I attempted adding Jinja2 comments to the file as a workaround, to no avail.
##### Issue Type:

Bug Report
##### Ansible Version:

Ansible 2 (commit e16b24c).
##### Ansible Configuration:

tasks/main.yml:

``` yaml
- debug: var='{{ lookup("template", "works.json.j2") }}'
  tags: debug

- debug: var='{{ lookup("template", "breaks.json.j2") }}'
  tags: debug
```

templates/works.json.j2:

``` json
x{ "works": "{{ ansible_date_time.date }}" }
```

templates/breaks.json.j2:

``` json
{ "breaks": "{{ ansible_date_time.date }}" }
```
##### Environment:

Python 2.7; Debian. Running within Docker.
##### Steps To Reproduce:
1. Create a role with the above config
2. Add it to a playbook
3. Run the play
##### Expected Results:

Run w/o errors.
##### Actual Results:

```
TASK [docker-registry : debug var={{ lookup("template", "works.json.j2") }}] ****
ok: [54.210.133.92] => {
    "changed": false, 
    "{{ lookup(\"template\", \"works.json.j2\") }}": "x{ \"foo\": \"2015-08-06\" }\n"
}

TASK [docker-registry : debug var={{ lookup("template", "breaks.json.j2") }}] ***
fatal: [54.210.133.92]: FAILED! => {"failed": true, "msg": "ERROR! an unexpected type error occurred. Error was sequence item 0: expected string, dict found"}
```

MichaelDrogalis:
I've run into this one. Feature is unusable as in.

vzaharee:
The issue actually occurs if the template is valid **Python** data, which also happens to be JSON that doesn't use `true`, `false`, or `null`.

markosamuli:
I've run into this when using `iam_policy` with a JSON template. What seems to work for me is to just add a line break into the beginning of the JSON template and it gets cleaned up in the process.

Using ansible-playbook 2.0.0 (devel 9bb95b5235)

jensschroer:
Just stumbled across the same error.
ansible 2.0.0 (devel 404ded2b03) last updated 2015/10/22 16:34:33 (GMT +200)
[...]

PR #12881, opened 2015-10-23
fix for issue #11885 - allowing json in lookup template calls

jensschroer:
fix for issue #11885.
When a json template is loaded via lookup('template' ... ), ansible tries to understand it as python dictionary by default, as the convert_data flag is by default true.
This commit allows to pass the convert_data flag to the 'template' plugin in lookup and therefore allow to load json templates.

### DIFF
--- a/docsite/rst/playbooks_lookups.rst
+++ b/docsite/rst/playbooks_lookups.rst
@@ -271,6 +271,10 @@ Here are some examples::
 
          - debug: msg="{{ lookup('template', './some_template.j2') }} is a value from evaluation of this template"
 
+         # loading a json file from a template as a string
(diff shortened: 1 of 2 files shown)


abadger:
- Needs documentation added here: https://github.com/ansible/ansible/blob/devel/docsite/rst/playbooks_lookups.rst#more-lookups
- @jimi-c and @bcoca had some thoughts on additional heuristics to apply convert_data=False to but I don't htink either of them block this one.  (They can correct me if I'm wrong)

jensschroer:
@abadger updated documentation. would you mind to check?

abadger:
merged, thanks!

# Merged

Hi!

issue is resolved.

Thank you!

PR #12893, opened 2015-10-23
change check for GCE auth parameters to allow empty strings- "".  This…

landphil:
Currently you can't run the GCE cloud modules using instance service accounts on GCE instances.  This requires initializing the libcloud auth with empty strings for the positional arguments.  The `if not value` checks were not allowing empty strings through.  I changed the checks to `if val not None`, which still catches unset values, but lets empty strings work.

… allows service accounts on GCE instances to authenticate through libcloud by setting the email and pem files to empty strings

### DIFF
--- a/lib/ansible/module_utils/gce.py
+++ b/lib/ansible/module_utils/gce.py
@@ -41,11 +41,11 @@ def gce_connect(module, provider=None):
 
     # If any of the values are not given as parameters, check the appropriate
     # environment variables.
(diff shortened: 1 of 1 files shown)


mckerrj:
@erjohnso Trying to review or close, can you or tom take a look? @supertom ?

bcoca:
closing as code was fixed and moved to gcp.py

PR #12901, opened 2015-10-24
sanatize results

bcoca:
centralize the sanitation function, now both callback output and registered vars are properly sanitized


abadger:
+1

Issue #11266, opened 2015-06-15
Make 'ansible-galaxy install -r requirements.yml' idempotent


tjanez:
##### Issue Type:

Bug Report/Feature Idea
##### Ansible Version:

```
ansible 1.9.1
  configured module search path = None
```
##### Ansible Configuration:

``` ini
[defaults]                                                                      
log_path = ../.tmp/ansible.log                                                  
# path to the directory containing roles                                        
roles_path = roles                                                              
hash_behaviour = merge                                                          
error_on_undefined_vars = yes                                                   
host_key_checking=False
```
##### Environment:

N/A
##### Summary:

Running `ansible-galaxy install -r requirements.yml` should be idempotent (e.g. it should behave like `pip install -r requirements.txt`).
##### Steps To Reproduce:
1. Create `requirements.yml` file with the following contents:
   
   ``` yaml
   # nodejs role
   - src: https://github.com/nodesource/ansible-nodejs-role
     version: 592a239dca52e1c5ff79cfefd0072efe3831428e
     name: nodejs
   ```
2. Run `ansible-galaxy install -r requirements.yml`.
3. Re-run `ansible-galaxy install -r requirements.yml`.
##### Expected Results:

``` bash
[tadej@tlinux64 admin]$ ansible-galaxy install -r requirements.yml
- executing: git clone https://github.com/nodesource/ansible-nodejs-role nodejs
- executing: git archive --prefix=nodejs/ --output=/tmp/tmpVoJEMx.tar 592a239dca52e1c5ff79cfefd0072efe3831428e
- extracting nodejs to roles/nodejs
- nodejs was installed successfully
```

On the second run, `ansible-galaxy` should detect that the `nodejs` role is already installed and make sure the installed version matches the version specified in `requirements.yml`.
##### Actual Results:

``` bash
[tadej@tlinux64 admin]$ ansible-galaxy install -r requirements.yml
- executing: git clone https://github.com/nodesource/ansible-nodejs-role nodejs
- executing: git archive --prefix=nodejs/ --output=/tmp/tmpVoJEMx.tar 592a239dca52e1c5ff79cfefd0072efe3831428e
- extracting nodejs to roles/nodejs
- nodejs was installed successfully
[tadej@tlinux64 admin]$ ansible-galaxy install -r requirements.yml
- executing: git clone https://github.com/nodesource/ansible-nodejs-role nodejs
[...]

PR #12904, opened 2015-10-26
Improve ansible-galaxy handling of role versions

willthames:
Ensure that role versions are considered when deciding
whether or not to (re-)install a role.

Issue a warning when the version of a dependency conflicts
with the version of an already installed role

Display what version of a role is being installed
Show the versions when upgrading/downgrading a role.

Implements #11266 

```
[will@dingo tmp]$ ansible-galaxy install -r ~/src/ansible/test/integration/galaxy_roles.yml -p roles
- downloading role 'oracle_java7', owned by briancoca
- downloading role from https://github.com/bcoca/ansible-oracle_java7-role/archive/v1.0.tar.gz
- extracting oracle_java7 to roles/oracle_java7
- oracle_java7 (v1.0) was installed successfully
- extracting git-ansible-galaxy to roles/git-ansible-galaxy
- git-ansible-galaxy (v1.6) was installed successfully
- adding dependency: git-gal-dep (v0.2)
- extracting hg-ansible-galaxy to roles/hg-ansible-galaxy
- hg-ansible-galaxy was installed successfully
- downloading role from https://bitbucket.org/willthames/http-ansible-galaxy/get/master.tar.gz
- extracting http-role to roles/http-role
- http-role was installed successfully
- extracting php to roles/php
- php was installed successfully
- extracting git-gal-dep to roles/git-gal-dep
- git-gal-dep (v0.2) was installed successfully
[will@dingo tmp]$ ansible-galaxy install -r ~/src/ansible/test/integration/galaxy_roles.yml -p roles
- changing role oracle_java7 from v1.0 to unspecified
- downloading role 'oracle_java7', owned by briancoca
- downloading role from https://github.com/bcoca/ansible-oracle_java7-role/archive/v1.0.tar.gz
- extracting oracle_java7 to roles/oracle_java7
- oracle_java7 (v1.0) was installed successfully
- git-ansible-galaxy (v1.6) is already installed, skipping.
- hg-ansible-galaxy is already installed, skipping.
- http-role is already installed, skipping.
- php is already installed, skipping.
```

Note that the behaviour with galaxy roles with an unspecified version is a little odd. This is because `meta/.galaxy_install_info` gets told the version of the role that actually gets installed when installing under galaxy. However, if you're not fixing versions in your rolesfile, then it's fine to expect that the role will be reinstalled each time you run ansible-galaxy install. (This is similar to the behaviour of `yum state=latest`, for example). One fix for this would be for `.galaxy_install_info` to be told the version of the role specified, rather than the version of the role installed.

### DIFF
--- a/lib/ansible/cli/galaxy.py
+++ b/lib/ansible/cli/galaxy.py
@@ -389,8 +389,13 @@ class GalaxyCLI(CLI):
             # query the galaxy API for the role data
 
             if role.install_info is not None and not force:
-                display.display('- %s is already installed, skipping.' % role.name)
-                continue
+                if role.install_info['version'] != role.version:
+                    display.display('- changing role %s from %s to %s' %
+                            (role.name, role.install_info['version'], role.version or "unspecified"))
+                    role.remove()
+                else:
+                    display.display('- %s is already installed, skipping.' % str(role))
+                    continue
 
             try:
                 installed = role.install()
@@ -411,14 +416,18 @@ class GalaxyCLI(CLI):
                         # we know we can skip this, as it's not going to
                         # be found on galaxy.ansible.com
                         continue
-                    if dep_role.install_info is None or force:
+                    if dep_role.install_info is None:
                         if dep_role not in roles_left:
-                            display.display('- adding dependency: %s' % dep_role.name)
+                            display.display('- adding dependency: %s' % str(dep_role))
                             roles_left.append(dep_role)
                         else:
                             display.display('- dependency %s already pending installation.' % dep_role.name)
                     else:
-                        display.display('- dependency %s is already installed, skipping.' % dep_role.name)
+                        if dep_role.install_info['version'] != dep_role.version:
+                            display.warning('- dependency %s from role %s differs from already installed version (%s), skipping' %
+                                            (str(dep_role), role.name, dep_role.install_info['version']))
+                        else:
+                            display.display('- dependency %s is already installed, skipping.' % dep_role.name)
 
             if not installed:
                 display.warning("- %s was NOT installed successfully." % role.name)
--- a/lib/ansible/cli/galaxy.py
+++ b/lib/ansible/cli/galaxy.py
@@ -388,11 +388,16 @@ class GalaxyCLI(CLI):
             display.vvv('Installing role %s ' % role.name)
             # query the galaxy API for the role data
 
-            if role.install_info is not None and not force:
+            if role.install_info is not None:
                 if role.install_info['version'] != role.version:
-                    display.display('- changing role %s from %s to %s' %
-                            (role.name, role.install_info['version'], role.version or "unspecified"))
-                    role.remove()
+                    if force:
+                        display.display('- changing role %s from %s to %s' %
+                                        (role.name, role.install_info['version'], role.version or "unspecified"))
+                        role.remove()
+                    else:
+                        display.warning('- %s (%s) is already installed - use --force to change version to %s' %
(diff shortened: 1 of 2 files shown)


lib/ansible/galaxy/role.py
```diff
@@ -311,4 +311,13 @@ def spec(self):
...
-
+    @property
+    def version_string(self):
+        """
+        Returns "rolename (version)" if version is not null
+        Returns "rolename" otherwise
+        """
+        if self.version:
+            return "%s (%s)" % (self.name, self.version)
+        else:
```
Lujeni, line 92: imho, this else is not usefull ?.

``` python
if self.version:
    return xxxx
return xxxx
```

willthames, line 92: @Lujeni meh, it's really not that important. I prefer it the way it is, and you're entitled to your preference. 

Unless you can point to actual documentation backing up your preferences though (PEP-8, Ansible coding standards, etc.), I'm unlikely to change my code.

chouseknecht:
+1

geerlingguy:
This is helpful towards solving both https://github.com/ansible/ansible/issues/11266 and https://github.com/ansible/galaxy-issues/issues/49

lib/ansible/galaxy/role.py
```diff
@@ -318,3 +318,14 @@ def spec(self):
         }
         """
         return dict(scm=self.scm, src=self.src, version=self.version, name=self.name)
+
+    @property
+    def version_string(self):
```
bcoca: not really version_string .. full_name? name_and_version?

lib/ansible/cli/galaxy.py
```diff
@@ -402,8 +402,13 @@ def execute_install(self):
             # query the galaxy API for the role data
 
             if role.install_info is not None and not force:
-                display.display('- %s is already installed, skipping.' % role.name)
-                continue
+                if role.install_info['version'] != role.version:
```
bcoca: insufficient check, this can downgrade a role, which is not what most will expect

lib/ansible/cli/galaxy.py
```diff
@@ -402,8 +402,13 @@ def execute_install(self):
             # query the galaxy API for the role data
 
             if role.install_info is not None and not force:
```
bcoca: we might want to change the following to a `else force:` to keep people from changing existing roles by default

lib/ansible/cli/galaxy.py
willthames: I'll put in a LooseVersion check, but it'll have to work as long as new LooseVersion is not less (as we can't compare different branch names etc.)

lib/ansible/galaxy/role.py
willthames: I think this can be `__repr__` rather than `version_string` :)

lib/ansible/cli/galaxy.py
willthames: actually, the change to using force will stop unexpected downgrades. 

You have to pass `--force` to ansible-galaxy to get an upgrade or a downgrade. (This might not be expected but that's the problem with enforcing the use of force just to get an upgrade)

willthames:
@bcoca should be ready for re-review now, I believe I've addressed your concerns.

swalkinshaw:
👋 apologies for the bump, but this PR deserves some attention as it implements a much requested feature.

geerlingguy:
Ooh, it's merged!

PR #12556, opened 2015-09-29
Fix for SLES 11.4, which has now also an /etc/os-release file.

photoninger:
SuSE added a "/etc/os-release" file with SLES 11 SP4 and now the setup module returns "0" in "ansible_distribution_release" instead of "4"
Because "VERSION_ID" is not an the last line of this file, release was always set to "0" in the loop "for line in data.splitlines():"

Here is an example of "/etc/os-release":
NAME="SLES"
VERSION="11.4"
VERSION_ID="11.4"
PRETTY_NAME="SUSE Linux Enterprise Server 11 SP4"
ID="sles"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:11:4"

PR #12909, opened 2015-10-26
Fix for SLES 11.4, which has now also an /etc/os-release file.

photoninger:
Because I messed up my previous PR https://github.com/ansible/ansible/pull/12556 I made new clean PR.


alxgu:
+1

alxgu:
Would be nice to see this fix in ansible 2.0 :smile: :

abadger:
@bcoca and I both looked at this and think it's reasonably safe.  merging to both devel and stable-2.0.  It may make it into 2.0.1 or it may not get in untii 2.0.2 depending on whether we find other bugs that warrant a 2.0.1rc2.

PR #12932, opened 2015-10-27
allow infinite retries for until: loops

larsks:
Interpret 0 retries as "retry forever".  This permits someone to write tasks that need to wait for the completion of external events for which the runtime is unknown or inconsistent, without having to resort to such hacky tricks such as specifying "a very large number".

This was inspired by #1298.

### DIFF
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -24,8 +24,10 @@ import json
 import subprocess
 import sys
 import time
+import itertools
 
 from ansible.compat.six import iteritems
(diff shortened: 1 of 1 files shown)


jimi-c:
Hi @larsks, I believe this would be confusing and potentially bad for users as they could have a play run indefinitely. I don't think it's much of an onus to specify a very large value for the retries, and/or a very large delay.

I'm going to close this here, but if you'd like to move the discussion to the #ansible-devel group feel free:

```
* https://github.com/ansible/ansible/issues/12933
```

Thanks!

PR #12934, opened 2015-10-27
Stable 1.9

timwoo112:


abadger:
Looks to be opened by mistake as stable and devel are two separate streams.

PR #12948, opened 2015-10-28
update callback plugin documentation

larsks:
this commit updates the documentation to make explicit the fact that
plugins must be listed in `callback_whitelist` in order to run, and it
adds configuration examples for both `callback_whitelist` and
`stdout_callback`.

### DIFF
--- a/docsite/rst/intro_configuration.rst
+++ b/docsite/rst/intro_configuration.rst
@@ -160,7 +160,10 @@ stdout_callback
 
 .. versionadded:: 2.0
 
-This setting allows you to override the default stdout callback for ansible-playbook.
+This setting allows you to override the default stdout callback for
+ansible-playbook::
+
+    stdout_callback = skippy
 
 .. _callback_whitelist:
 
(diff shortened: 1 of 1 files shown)


bcoca:
this might change due to #12867

bcoca:
i initially meant to use the whitelist only for plugins shipped with ansible

larsks:
@bcoca I've modified this pr so that it has just the config examples for `intro_configuration.rst`.

docsite/rst/intro_configuration.rst
```diff
@@ -160,7 +160,10 @@ stdout_callback
 
 .. versionadded:: 2.0
 
-This setting allows you to override the default stdout callback for ansible-playbook.
+This setting allows you to override the default stdout callback for
+ansible-playbook::
```
bcoca, line 164: no need for this space change, we don't have any users stuck with an 80 column display

larsks:
And yet it's harmless, and it certainly makes it easier for me to work with.

larsks:
I'm not sufficiently motivated to deal with line length preferences.  I will let someone else submit the doc fixes.

Issue #12952, opened 2015-10-28
Bug Report: debug + with_items is not working properly


Yannig:
**Issue Type:**

Bad var substitution using with_ statement and debug.

**Ansible version:**

```
ansible 2.0.0 (devel 92dd563a75) last updated 2015/10/28 20:05:08 (GMT +200)
  lib/ansible/modules/core: (detached HEAD 211196bdf6) last updated 2015/10/26 09:59:00 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD 5d4474a2b2) last updated 2015/10/26 09:59:00 (GMT +200)
```

**Ansible Configuration:**

Nothing special

**Environment:**

Linux Mint 17.2

**Summary:**

Debug + with_items does not work properly

**Steps To Reproduce:**

Use the following playbook:

``` yaml

---
- name: "Lookup test"
  hosts: "localhost"
  connection: "local"
  vars:
    var1: "1"
    var2: "2"
  tasks:
    - debug: var={{item}}
      with_items: [ 'var1', 'var2' ]
    - debug: var=var1
    - debug: var=var2
    - debug: msg={{item}}
      with_items: [ 'var1', 'var2' ]
```

**Expected Results:**

With Ansible 1.9, we get the following result:

``` yaml
TASK: [debug var={{item}}] **************************************************** 
ok: [localhost] => (item=var1) => {
    "item": "var1", 
    "var": {
        "var1": "1"
    }
}
ok: [localhost] => (item=var2) => {
    "item": "var2", 
    "var": {
        "var2": "2"
    }
}
```

**Actual Results:**

With the last version of Ansible, I get the following output:

``` yaml
TASK [debug var={{item}}] ******************************************************
ok: [localhost] => (item=var1) => {
    "item": "var1", 
    "{{item}}": "var1"
}
ok: [localhost] => (item=var2) => {
    "item": "var2", 
    "msg": "here we are"
}
```

Yannig:
I wrote the following patch in order to try to resolve this issue:

``` diff
diff --git a/lib/ansible/executor/task_executor.py b/lib/ansible/executor/task_executor.py
index e8a785a..0d5ca69 100644
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -296,12 +296,6 @@ class TaskExecutor:
             return dict(include=include_file, include_variables=include_variables)

         # Now we do final validation on the task, which sets all fields to their final values.
[...]

PR #12954, opened 2015-10-28
Fix proposal for issue #12952 (debug+var issue)

Yannig:
This PR remove code that seems to be unnecessary and fix display issue with debug + var + with_items.

See https://github.com/ansible/ansible/issues/12952 for more details

### DIFF
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -296,12 +296,6 @@ class TaskExecutor:
             return dict(include=include_file, include_variables=include_variables)
 
         # Now we do final validation on the task, which sets all fields to their final values.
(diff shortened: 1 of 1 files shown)


jimi-c:
I'm pretty sure this isn't correct, as that code was added to deal with debug specifically, though perhaps we changed something else that is now making that code obsolete. I'll look into the original issue, however to see if this is the correct fix.

Yannig:
@jimi-c: you might be correct. This PR is only here as a proposal after a quick look around. I guess that it might have some side effect I did not notice.

jimi-c:
After digging into this, I see your PR returns the output to that shown in the stable-1.9 branch, so I've gone ahead and merged it.

Thanks!

PR #12958, opened 2015-10-29
Note previous versions usage of colon delimiter

MeanderingCode:
Doc site talks about the use of a comma as the host group delimiter for and/or/not hosts patterns.  The comma does not work in version 1.9, but I only found out about the use of colon from a closed Github issue.

PLEASE CHECK VERSION CLAIM!!

I have no idea which version introduced the change, and simply wrote that it was version 2.  Please correct before merging.

### DIFF
--- a/docsite/rst/intro_patterns.rst
+++ b/docsite/rst/intro_patterns.rst
@@ -37,6 +37,8 @@ This means the host may be in either one group or the other::
     webservers
     webservers,dbservers
 
+.. note:: Prior to version 2, the notation for listing multiple selectors was a colon, ":", instead of a comma
+
 You can exclude groups as well, for instance, all machines must be in the group webservers but not in the group phoenix::
 
     webservers,!phoenix


bcoca:
the , was introduced long ago, it was broken in 1.9 but no one noticed until we decided to phase out the : .Both were valid, but : was favored for most of previous versions, but since it creates problems with ranges and ipv6 we decided to deprecate it, w/o realizing the breakage in 1.9.

bcoca:
fixed a few things, please check 3c3aff09c9d9764f98e2824251277d45f9216963 to see if that covers everything

bcoca:
closing this as per above

tjanez:
If anyone is curious in which commits this was fixed, here they are:
- 9c09dc3ef81d4f40e25c3546c2945b4d2ad92532
- 13bf62d0e6805ecc03aee588a6e3fad407e808ec
- 3c3aff09c9d9764f98e2824251277d45f9216963

@bcoca, thanks!

PR #12962, opened 2015-10-29
fixes options for ansible pull

### LINKED ISSUE #12955: ansible-pull 2.0.0 requires an argument
with ansible 2.0.0-beta2 I get:

```
# ansible-pull --url myurl --key-file key
Usage: ansible-pull <host-pattern> [options]
....
ERROR! Missing target hosts
```

Man page still indicates that it should default to local.yml.  Seems something is messed up here.

bcoca:
- remove requirement for host patterns, use the defaults
- require destination directory (None in cwd is not a good default)
- fixed usage messages
- updated default inventory to use , and not deprecated

fixes #12955

### DIFF
--- a/lib/ansible/cli/pull.py
+++ b/lib/ansible/cli/pull.py
@@ -50,7 +50,7 @@ class PullCLI(CLI):
         ''' create an options parser for bin/ansible '''
 
         self.parser = CLI.base_parser(
-            usage='%prog  -d <destdir> -U <repository> [options]',
+            usage='%prog -U <repository> [options]',
             connect_opts=True,
             vault_opts=True,
             runtask_opts=True,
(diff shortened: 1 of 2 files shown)


abadger:
Code looks fine.  #12955 also asks for man page update.  And we should document in chaneglog if the CLI is incompatible (requires dest) as well.

bcoca:
man page is already updated

abadger:
```
*-d* 'DEST', *--directory=*'DEST'::

Directory  to checkout repository into. If not provided, a subdi-
rectory of ~/.ansible/pull/ will be used.
```

That's no longer true, correct?

lib/ansible/cli/pull.py
```diff
@@ -129,6 +126,7 @@ def run(self):
         #TODO: enable more repo modules hg/svn?
         if self.options.module_name == 'git':
             repo_opts = "name=%s dest=%s" % (self.options.url, self.options.dest)
+            print(repo_opts)
```
abadger: Is this debugging that should be removed?

bcoca:
#totallymeanttodothat

jimi-c:
Merged, thanks!

PR #12965, opened 2015-10-29
Fix ec2 infinitely nested groups

matburt:
The ec2 inventory module will incorrectly place parent groups as child
groups of themselves, this fixes that.


jimi-c:
Merged, thanks!

PR #12966, opened 2015-10-29
Expand variables

hewang93:


sivel:
Since this is just a core submodules hash change, and it doesn't match the description, I'm assuming this was made in error.

PR #12968, opened 2015-10-29
Move evaluation of conditionals to happen before templating

### LINKED ISSUE #12774: v1.9->2: templated delegate_to with undefined variable fails even if task is skipped on when: undefined
It seems in 2.0 (174ecb843c) a `delegate_to:` gets templated, even when the task would be skipped because the variable used to set the delegate is undefined. This does not happen in 1.9.

Following task snippet reproduces this
(source: https://github.com/bennojoy/mysql/blob/master/tasks/main.yml#L85)

```
139 - name: Get the current master servers replication status                       
140   mysql_replication:                                                            
141     mode:   getmaster                                                           
142   delegate_to: "{{ mysql_repl_master }}"                                        
143   register: repl_stat                                                           
144   when: slave|failed and mysql_repl_role == 'slave' and mysql_repl_master is defined
```

This works fine in 1.9, and used to work in 2.0 (at least until some months/weeks ago).

```
TASK: [mysql | Get the current master servers replication status] ****** 
skipping: [hydargos]
```

Testing with 174ecb843c, this fails:

```
TASK [mysql : Get the current master servers replication status] ********
fatal: [hydargos -> {{ mysql_repl_master }}]: FAILED! => {"failed": true, "msg": "ERROR! 'mysql_repl_master' is undefined"}
```

abadger:
This Fixes #12774 but I don't know if it's correct.  Quite a few things in this area of code depend on happening in precisely the correct sequence.  Waiting on a review by @jimi-c before merging.

### DIFF
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -259,18 +259,6 @@ class TaskExecutor:
 
         templar = Templar(loader=self._loader, shared_loader_obj=self._shared_loader_obj, variables=variables)
 
-        # apply the given task's information to the connection info,
-        # which may override some fields already set by the play or
-        # the options specified on the command line
-        self._play_context = self._play_context.set_task_and_variable_override(task=self._task, variables=variables, templar=templar)
-
-        # fields set from the play/task may be based on variables, so we have to
-        # do the same kind of post validation step on it here before we use it.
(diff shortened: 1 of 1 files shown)


jimi-c:
Closing in favor of 479cbfc63c8261577b359b7703d7e6168c1a74f5

PR #12981, opened 2015-10-30
[feature] add src_path to galaxy requirements spec

mgwilliams:
This feature makes it possible to require a role that is a subdirectory within `src`. For example, a role that is embedded in a larger project:

```
example-project/
├── extra
│   └── ansible-role
│       ├── defaults
│       │   └── main.yml
│       ├── files
│       ├── handlers
│       │   └── main.yml
│       ├── meta
│       │   └── main.yml
│       ├── README.md
│       ├── tasks
│       │   └── main.yml
│       ├── templates
│       └── vars
│           └── main.yml
├── README.md
└── src
    └── ...
├── ...
```

The following entry in `requirements.yml` would then install the above `ansible-role` as `example`:

```
- src: https://github.com/auser/example-project
  name: example
  src_path: extra/ansible-role
```

### DIFF
--- a/docsite/rst/galaxy.rst
+++ b/docsite/rst/galaxy.rst
@@ -88,6 +88,12 @@ And here's an example showing some specific version downloads from multiple sour
     - src: https://github.com/bennojoy/nginx
       version: master
       name: nginx_role
+
+    # from github, specifying a subdirectory (ansible/role) that
+    # contains the role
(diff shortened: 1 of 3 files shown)


calfonso:
Hi!

Unfortunately, we're not sure if we want this feature in the program, there is a workaround to achieve what you want by adding a meta/main.yml in the root directory.

Issue #12983, opened 2015-10-30
[v2] global name 'no_log' is not defined


stcheng:
Ansible Version: 2.0.0

Hello,

I notice that with -vvv option enabled, the action plugins are returning something that doesn't meet with expected data structure.

The error message will display 

```
[WARNING]: Error when using <bound method CallbackModule.v2_runner_on_failed of </usr/local/lib/python2.7/dist-packages/ansible-2.0.0-py2.7.egg/ansible/plugins/callback/default.CallbackModule object at 0x7f1d28c47b10>>: global name 'no_log' is not defined
```

It seems that it is because the return value of the ActionModule's function run returns a dictionary without the entry 'no_log'.

Checked in file ansible/plugins/callback/**init**.py and file out on line 74-76:

```
# remove invocation info unless its very very verbose
if 'invocation' in result and (self._display.verbosity < 3 or no_log):
    del result['invocation']
```

Not sure where this no_log comes from. I think no_log variable shall be assigned to as:

```
no_log = result.get('_ansible_no_log', False)
```

Please correct me if I did something wrong.

bcoca:
fixed via 25e3fe04a8dfb42209f3740d0236fd77b1b029f4

PR #12985, opened 2015-10-30
Fix for #12983: CallbackBase no_log variable fix

stcheng:


bcoca:
alrady fixed via 25e3fe04a8dfb42209f3740d0236fd77b1b029f4

PR #12991, opened 2015-10-31
Use shorter when statement in playbooks_conditionals documentation.

ypid:
Don’t teach bad habits.


bcoca:
the example is to show how to use parenthesis in conditionals, your change make the example not match the text

ypid:
Updated. The section is about the `when` statement so I think this change ok.

bcoca:
I think you missed the whole point, the purpose was to show how to use parenthesis to group conditionals. Closing this.

ypid:
Fair enough.

PR #12992, opened 2015-10-31
Avoid key names which are valid attributes for a dictionary in Python.

### LINKED ISSUE #12990: 'when: (item.copy)' fails (always True) when variable is called 'copy'.
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.9.4 and ansible devel (d341ba14a5f4b974d41f4b11f678b356c89f79d9)
##### Ansible Configuration:

Default configuration
##### Environment:

Debian Jessie
##### Summary:

In a `when` condition `item.copy` always evaluates to True regardless of the actual value of `item.copy`.

Has `copy` some special meaning when used as key in a list of dicts that I am unaware of?
##### Steps To Reproduce:

``` YAML
- hosts: localhost
  vars:
    test_list:
      - name: 'test'

        # Can also be completely removed thus being undefined.
        copy: False

        some_other_name: False

  tasks:
    - debug: var=test_list
    - debug:
        msg: 'False'
      when: (item.some_other_name)
      with_items: test_list
    - debug:
        msg: 'False'
      when: (item.copy)
      with_items: test_list
```
##### Expected Results:

```
TASK [debug msg=False] *********************************************************
skipping: [localhost] => (item={u'some_other_name': False, u'copy': False, u'name': u'test'})

TASK [debug msg=False] *********************************************************
skipping: [localhost] => (item={u'some_other_name': False, u'copy': False, u'name': u'test'})
```
##### Actual Results:

```
TASK [debug msg=False] *********************************************************
skipping: [localhost] => (item={u'some_other_name': False, u'copy': False, u'name': u'test'})

TASK [debug msg=False] *********************************************************
ok: [localhost] => (item={u'some_other_name': False, u'copy': False, u'name': u'test'}) => {
    "item": {
        "copy": false,
        "name": "test",
        "some_other_name": false
    },
    "msg": "False"
}
```

ypid:
Closes #12990.

https://github.com/ypid/ansible/blob/avoid-python-object-name-in-dict/docsite/rst/playbooks_variables.rst#what-makes-a-valid-variable-name

### DIFF
--- a/docsite/rst/playbooks_variables.rst
+++ b/docsite/rst/playbooks_variables.rst
@@ -36,10 +36,13 @@ Variable names should be letters, numbers, and underscores.  Variables should al
 
 ``foo-port``, ``foo port``, ``foo.port`` and ``12`` are not valid variable names.
 
-When you work with dictionaries note that you should avoid keys which are valid attributes for a dictionary in Python.
-As a rule of thump, avoid all key names that start and end with two underscores for example ``__do_not_use__``.
+When you work with more complex data types like dicts and sets note that you
+should avoid keys which are valid attributes for these data types in Python.
+As a rule of thump, avoid all key names that start and end with two underscores
+for example ``__do_not_use__``.
 Also avoid any of the following words:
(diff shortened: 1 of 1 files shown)


tersmitten:
Thanks for the clarification

bcoca:
dictionaries are not the only possible form of a var, you have lists, sets, tuples, strings, ints, floats, etc

ypid:
You are right, but I don’t expected the same kind of problems as in #12990 with ints and floats and I wanted to keep the list somewhat small. If you think we should also exclude those names then checkout #12993 instead.

bcoca:
closing in favor of #12993

PR #12999, opened 2015-11-01
Use shorter when statement in playbooks_conditionals documentation.

ypid:
Rework of #12991.


bcoca:
he, i had just changed the condition to not 'offend' you as much, this now does not match

ypid:
All right. I can life with that. :smile:

Issue #12997, opened 2015-11-01
Performance regression with Ansible v2 using include + big hash map


Yannig:
##### Issue Type:

This is a bug report about speed regression between Ansible 1.9.4 and current devel version when combining big hash map and include statement.
##### Ansible Version:

```
ansible 2.0.0 (devel f528ad1848) last updated 2015/11/01 09:23:49 (GMT +200)
  lib/ansible/modules/core: (detached HEAD a3a7cdb9aa) last updated 2015/11/01 10:03:19 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD b1db0e893c) last updated 2015/11/01 10:03:19 (GMT +200)
```
##### Ansible Configuration:

Nothing special. The playbook work with  local connection.
##### Environment:

N/A
##### Summary:

I have speed regressions against various playbooks. With Ansible 1.9.4, this playbooks was completed in about 3 to 5s. With last version of Ansible, they take 7 more times.

All my tests use a local virtual machine.
##### Steps To Reproduce:

You can retrieve playbook that reproduce this issue at the following location: https://github.com/Yannig/ansible-issue-low-speed/

This repository have 2 playbooks: one using a big hash map (issue-vars.yml) but not the other (no-vars-mapping.yml).
##### Expected Results:

Actually, using ansible 1.9.4, the playbook take about 3 seconds:

```
time ansible-playbook issue-vars/issue-vars.yml 
```

``` yaml
PLAY [Test] ******************************************************************* 

TASK: [do-include | debug var=application_name] ******************************* 
ok: [localhost] => {
    "var": {
        "application_name": "application1"
    }
}

TASK: [do-include | Répertoire des roles] ************************************ 
ok: [localhost] => (item={'key': 'field1', 'value': {'field': u'NA', 'name': 'field1'}})

TASK: [do-include | template src=test.j2 dest="/tmp/{{item.value.field}}/1-{{item.value.name}}.txt"] *** 
ok: [localhost] => (item={'key': 'field1', 'value': {'field': u'NA', 'name': 'field1'}})

TASK: [do-include | shell export FIELD={{item.item.value.field}} ; echo "$FIELD" ; cat /tmp/{{item.item.value.field}}/{{item.item.value.name}}.txt
] *** 
[...]
TASK: [do-include | shell export FIELD={{item.item.value.field}} ; echo "$FIELD" ; cat /tmp/{{item.item.value.field}}/{{item.item.value.name}}.txt
] *** 
[...]

Issue #12239, opened 2015-09-03
Low performance


lberruti:
##### ISSUE TYPE
Bug Report

##### COMPONENT NAME
core

##### ANSIBLE VERSION
```
devel
ansible 2.0.0 (detached HEAD b2cb3153f1) last updated 2015/09/03 230737 (GMT +200)
lib/ansible/modules/core (detached HEAD bbcfb1092a) last updated 2015/09/03 230739 (GMT +200)
lib/ansible/modules/extras (detached HEAD b8803306d1) last updated 2015/09/03 230739 (GMT +200)
config file =
configured module search path = /root/ansible/library
```

##### CONFIGURATION
none

##### OS / ENVIRONMENT
Debian 7

##### SUMMARY
Hi, unfortunately version 2.0 seems slow down.
With a real playbook: v2 run within 32 secs, v2, stable-1.9 19 secs.
Same conditions (target machine, control machine), no network congestions (local net), no network downloads.

##### STEPS TO REPRODUCE
time ansible-playbook -i hosts simplespeedtest.yml > /dev/null
```
---
# simplespeedtest.yml
- hosts: localhost
  tasks:
    - command: echo 1
    - command: echo 2
#  and so on
    - command: echo 30
```

##### EXPECTED RESULTS
ansible stable-1.9
ansible 1.9.3 (detached HEAD 066b7079ef) last updated 2015/09/03 23:19:28 (GMT +200)
  lib/ansible/modules/core: (detached HEAD dccef8f89a) last updated 2015/09/03 23:17:15 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD 25595b4ecc) last updated 2015/09/03 23:17:15 (GMT +200)
  configured module search path = /root/ansible/library
real    0m2.608s
user    0m1.776s
sys 0m0.464s
real    0m2.476s
user    0m1.656s
sys 0m0.468s
real    0m2.497s
user    0m1.688s
sys 0m0.444s

##### ACTUAL RESULTS
real    0m3.848s
user    0m2.484s
sys 0m0.496s
real    0m3.524s
user    0m2.208s
sys 0m0.376s
real    0m3.926s
user    0m2.396s
sys 0m0.544s

geerlingguy:
I'm seeing the same results, using the following playbook:

``` yaml
---
- hosts: localhost
  connection: local

  tasks:
    - command: echo 1
    - command: echo 2
    - command: echo 3
    - command: echo 4
    - command: echo 5
[...]

PR #13000, opened 2015-11-01
Cache for _do_template call.

Yannig:
This PR is a proposal to fix a part of the following issue https://github.com/ansible/ansible/issues/12997

May result in nice speed improvement (4-5 times faster) when used with big hash map.

Take it with a grain of salt has it may break things...

### DIFF
--- a/lib/ansible/inventory/__init__.py
+++ b/lib/ansible/inventory/__init__.py
@@ -37,6 +37,8 @@ from ansible.plugins import vars_loader
 from ansible.utils.vars import combine_vars
 from ansible.parsing.utils.addresses import parse_address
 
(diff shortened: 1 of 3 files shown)


bcoca:
judging from the failing test, the cached value is not always correct

Yannig:
Indeed :)

bcoca:
still, keeping it open for now, discussing with core team as this does show better performance even if incorrect.

Yannig:
Well, it's seems better now :)

bcoca:
worth testing as fix for #12239

jimi-c:
Merged, thanks!

Yannig:
@jimi-c nice :)

PR #13006, opened 2015-11-02
Ansible executable does not detect/report unsupported invocation

Maroloccio:
The core of this Pull Request is a plain bug fix.

The `ansible` executable (i.e. bin/ansible) is symlinked to in various forms from the bin/ directory:
- ansible-doc -> ansible
- ansible-galaxy -> ansible
- ansible-playbook -> ansible
- ansible-pull -> ansible
- ansible-vault -> ansible

Inside `ansible`, which form to take is determined by looking at sys.argv[0].

When the executable is invoked in an unsupported form, an exception "should" be raised:

```
raise AnsibleError("Program not implemented: %s" % me)
```

The code that does that is broken, as proven by creating a new symlink (e.g.: "unsupported"):

```
$ cd bin
$ ln -s ansible unsupported
$ ./unsupported
```

The resulting exception shows that the "unsupported" check is implemented incorrectly:

```
Unexpected Exception: name 'mycli' is not defined
the full traceback was:

  File "./unsupported", line 77, in <module>
    cli = mycli(sys.argv, display=display)
NameError: name 'mycli' is not defined
```

This pull request contains multiple commits, which not only fix the problem, but also
improve the executable's readability.

Examples:

The core bug fix commit is:
https://github.com/ansible/ansible/commit/3133f442ae1d32ef2ae40bdf96b2f8ad40561c90

Explicit exit/error codes:
https://github.com/Maroloccio/ansible/commit/c1d816053b6cc268eafaed28041178bacb027cd2

... for a subjectively more readable final version of:

https://github.com/Maroloccio/ansible/blob/devel/bin/ansible

### DIFF
--- a/bin/ansible
+++ b/bin/ansible
@@ -21,6 +21,7 @@ from __future__ import (absolute_import,
                         print_function)
 
 __requires__ = ['ansible']
+
 try:
     import pkg_resources
 except Exception:
@@ -35,7 +36,9 @@ import os
 import sys
 import traceback
 
-from ansible.errors import AnsibleError, AnsibleOptionsError, AnsibleParserError
+from ansible.errors import (AnsibleError,
+                            AnsibleOptionsError,
+                            AnsibleParserError)
 from ansible.utils.display import Display
 from ansible.utils.unicode import to_unicode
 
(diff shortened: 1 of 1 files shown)


bcoca:
fixed via 56a2412dc2da760d0f481b16eba88556d9b26fda

PR #13007, opened 2015-11-02
Avoid using magic numbers as exit codes

Maroloccio:
Using constants for exit codes makes it clear why some numbers are reserved.


bcoca:
this contains style changes, which we don't normally accept as they break attribution.

PR #13008, opened 2015-11-02
The else clause is the correct place to raise

Maroloccio:
The recommended place to raise seems much more logical, and idiomatic, than the confusing trap for NameError.

The code as proposed here is as explicit as it gets.

### DIFF
--- a/bin/ansible
+++ b/bin/ansible
@@ -73,12 +73,10 @@ if __name__ == '__main__':
             from ansible.cli.vault import VaultCLI as mycli
         elif me == 'ansible-galaxy':
             from ansible.cli.galaxy import GalaxyCLI as mycli
-
-        try:
-            cli = mycli(sys.argv, display=display)
-        except NameError:
+        else:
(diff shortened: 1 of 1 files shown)


bcoca:
this disallows adding custom command lines w/o editing the code

Maroloccio:
Identical to how it works now, just a better place to raise the error?

bcoca:
no, the code allows for anyone to create a new cli just by creating the right class and a symlink, this would disallow that

Maroloccio:
This seems impossible at the moment too, unless the if/elif/elif/... selection is updated too?

bcoca:
yes, thought i had merged that already, will do so soon, which makes this change counter productive

amenonsen:
The underlying changes that @bcoca mention above have been merged, and the subclass to use is no longer determined by if/elif/… blocks. So this change is no longer relevant, and doesn't apply anyway. So I'm closing this PR now. @Maroloccio thanks for the contribution, sorry it was rendered unnecessary by other changes.

PR #13011, opened 2015-11-03
Fix typo

amenonsen:


bcoca:
no typo here, just a new term in invented to both denote ownership and access control :-p

PR #13021, opened 2015-11-03
Allow PATCH to list of valid HTTP methods

erjohnso:
I'm working on the kubernetes module[1] and in order to cover the full kubernetes API, [patch operations](http://kubernetes.io/v1.0/docs/devel/api-conventions.html#patch-operations) need to be supported. For the REST calls, I'm using `ansible.module_utils.urls` and `PATCH` is not supported.

This PR adds the `PATCH` to the list of allowed HTTP methods.

[1] https://github.com/ansible/ansible-modules-extras/issues/1139


erjohnso:
Thanks @abadger!!

abadger:
My pleasure!  Thanks for adding that.

PR #13036, opened 2015-11-04
treat forks config parameter as max value...

larsla:
... instead of always creating that number of workers.
Otherwise we have a lot of workers with using resources without having any work to do.
This was exhausting resources on our Jenkins server even when the plays were on pretty small inventories.


bcoca:
+1 this was the 1.x behavior, but i would even check against serial also

PR #13064, opened 2015-11-06
v1 variable precedence order was off.

tima:
Noticed that the v1 variable precedence docs list facts discovered as having a lower precedence than inventory variables. It is in reality the other way around. The v2 section gets this right.


jimi-c:
Thanks @tima, merged.

PR #13065, opened 2015-11-06
Consistent inventory pattern delimiter in docs

tima:
The inventory pattern delimiter used in the examples switches between the comma and colon. While both are valid I've found this throws off new users so I've modified the examples to consistently use a colon, the more common of the two in my experience.

### DIFF
--- a/docsite/rst/intro_patterns.rst
+++ b/docsite/rst/intro_patterns.rst
@@ -39,7 +39,7 @@ This means the host may be in either one group or the other::
 
 You can exclude groups as well, for instance, all machines must be in the group webservers but not in the group phoenix::
 
-    webservers,!phoenix
+    webservers:!phoenix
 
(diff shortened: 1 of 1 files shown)


abadger:
Thanks @tima.  Note that sometime post 2.0 we're planning on deprecating colon and making comma the official way to delimit patterns.  But 1.9.x doesn't handle commas so we're not planning on making that transition until sometime down the road when people are mostly using 2.x.

tima:
OK but that's a bummer. What is the reasoning behind the decision to deprecate that? I find the colons easier to read and visually parse than the commas.

bcoca:
ipv6 addresses and group/host ranges already use the : and it is hard to get every case right when you overload an operator.

tima:
Oh right. Good point. Forgot about v6 IP support.

PR #13071, opened 2015-11-06
Auto inclusion of vaulted files in group_vars and host_vars

ponysmith:
Adds vaulted files to the autoloaded files in `group_vars` and `host_vars`.  Files follow the naming conventions `<hostname>.vault.yml` for vault files in `host_vars` and `<groupname>.vault.yml` for vault files in `group_vars`.

### DIFF
--- a/lib/ansible/inventory/__init__.py
+++ b/lib/ansible/inventory/__init__.py
@@ -712,10 +712,12 @@ class Inventory(object):
                 # load vars in dir/group_vars/name_of_group
                 base_path = os.path.realpath(os.path.join(basedir, "group_vars/%s" % group.name))
                 results = self._variable_manager.add_group_vars_file(base_path, self._loader)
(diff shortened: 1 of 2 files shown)


bcoca:
im not sure this is needed, you can already keep vaulted files in group/host vars and they will automatically be loaded, you can put multiple files in a <groupname or hostname>/ directory, with plain yaml and vaulted side by side.

sivel:
I agree with @bcoca the fact that inside of `host_vars` and `group_vars` you can have directories with the names of hosts or groups that can then hold multiple YAML files including vaulted files makes this already possible.

abadger:
# Question About This One

Hi!

- Using directories for your host_vars and group_vars you can already do the same thing that this Pull Request implements.

ponysmith:
Hi @abadger ... thanks for the response.  I guess I was unaware of how to accomplish this with the current usage of the host_vars and group_vars directories.  I'll post to the google group and see what solutions people have.

PR #13074, opened 2015-11-08
Use ansible_host in synchronize module

### LINKED ISSUE #13073: synchronize uses deprecated ansible_ssh_host variable
Synchronize module doesn't work for me.

http://docs.ansible.com/ansible/intro_inventory.html states that ansible_ssh_host is deprecated.
The synchronize module tries to use this deprecated variable with fallback to the inventory_hostname. Neither works for me:
- ansible_ssh_host is not defined, I use ansible_host.
- inventory_hostname is an alias that is not resolvable with dns.

I think the module should try ansible_host first, then ansible_ssh_host for legacy inventories and fall back to inventory_hostname.

Thanks!

joernheissler:
Fixes #13073

### DIFF
--- a/lib/ansible/plugins/action/synchronize.py
+++ b/lib/ansible/plugins/action/synchronize.py
@@ -131,7 +131,13 @@ class ActionModule(ActionBase):
         src_host = '127.0.0.1'
         inventory_hostname = task_vars.get('inventory_hostname')
         dest_host_inventory_vars = task_vars['hostvars'].get(inventory_hostname)
-        dest_host = dest_host_inventory_vars.get('ansible_ssh_host', inventory_hostname)
+        try:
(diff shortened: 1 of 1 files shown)


jimi-c:
@ianzd while this does fix the issue, and should probably be merged, I believe part of the underlying issue is that we're not setting all of the alternate values (if not otherwise set) when filling in missing variables (done in PlayContext). We probably want to fix that as well.

abadger:
We took a look at the code in playcontext and realized that hostvars aren't part of this merging so (at least for now) we do think that having synchronize try multiple variable names is hte way to go.

abadger:
@joernheissler Thanks for fixing this!

# Merged

Hi!

issue is resolved.

Thank you!

abadger:
Cherry-picked to the stable-2.0 branch

PR #13076, opened 2015-11-08
setup.py: do not require pycrypto

techtonik:
PyCrypto is not available for Windows as installable package


sivel:
Pycrypto is in fact a dependency, we cannot remove it like this.

Ansible does not support installing on, or running from a Windows machine. Windows is only supported as a target.

If you want to try running Ansible on Windows, you should `pip install --no-deps` and handle your deps manually.

Closing this as we don't support Windows as a controller and we require pycrypto.

techtonik:
@sivel pycrypto is optional dependency. The real dependency is paramiko, which already includes pycrypto.

PR #13078, opened 2015-11-08
Create mongo.py

prakritish:
Adds inventory script for Mongo DB.


jimi-c:
See comment on https://github.com/ansible/ansible/pull/13079, would like these combined.

prakritish:
As requested, I have included the mongo.ini in this Pull Request

PR #13080, opened 2015-11-08
add resolv and resolv_all filters

larsks:
This commit adds two filters to Ansible that permit translating a
hostname into one (or more) ip addresses:
- `resolv` resolves a hostname into a single ip address:
  
  ```
  The ip of {{hostname}} is {{hostname|resolv}}.
  ```
- `resolv_all` provides access to all the available addresses associated
  with a hostname:
  
  ```
    {{hostname}} is available as:
  
    {% for addr in hostname|resolv_all %}
    - {{addr}}
    {% endfor %}
  ```

### DIFF
--- a/docsite/rst/playbooks_filters.rst
+++ b/docsite/rst/playbooks_filters.rst
@@ -284,6 +284,27 @@ in :doc:`playbooks_filters_ipaddr`.
 
 .. _hash_filters:
 
+Hostname lookup filter
+----------------------
+.. versionadded:: 2.0
+
+To resolve a hostname to an ip address::
+
+    {{ my_var | resolv }}
+
+To resolve a hostname to a list of ip addresses (e.g., for a hostname that
+has multiple `A` records in DNS)::
+
+    {{ myvar | resolv_all }}
+
+For example, to show a list of all ``google.com`` addresses::
+
+    google.com is available at:
+
+    {% for addr in 'google.com' | resolv_all %}
+    - {{addr}}
+    {% endfor %}
+
 Hashing filters
 --------------------
 .. versionadded:: 1.9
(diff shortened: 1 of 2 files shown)


drybjed:
There is already a [dig](https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/lookup/dig.py) lookup plugin which could be used for the same purpose, with [some examples](http://jpmens.net/2015/02/20/a-new-dns-lookup-plugin-for-ansible/).

What I would be interested in is a way to get the DNS information as it is seen by the remote host instead of Ansible Controller (remember that different hosts can use different DNS servers which might return different data). But that would probably require a module. Actually, `dig` lookup plugin can use a specified DNS server to lookup the data, the way to achieve what I want would be to get the DNS servers remote host is using and ask them directly by Ansible Controller. Neat.

larsks:
This does get you the DNS information as seen by the remote host. That's one reason this is a filter plugin rather than a lookup plugin (because a lookup plugin for host lookups will often be wrong).

larsks:
This does get you the DNS information as seen by the remote host. That's one reason this is a filter plugin rather than a lookup plugin (because a lookup plugin for host lookups will often be wrong). 

Also, this filter users only core python modules and this has no dependencies other than those already required by Ansible.

larsks:
Sorry about the double comment. Apparently either the mobile interface is funky or I am an incompetent Android user.

sivel:
Both filter plugins and lookup plugins execute on the controller.  So I'm not sure why you mention that they are filters so that they resolve on the target.  Only modules have the ability to execute on the remote target.

larsks:
Because I was wrong?  Ah, well.

bcoca:
fyi, to get host resolution on target machines you can use the `getent` module with the hosts database

PR #13098, opened 2015-11-09
Fix get_file_lines on inexistent file

ghislainbourgeois:
Noticed an error when /etc/resolv.conf does not exist and tracked it down to the open call in get_file_lines. This patch fixes the error by checking for the file and read access, moving the open call inside the try block and returning an empty list.


bcoca:
there is already a function that safely returns file contents, this fixes this function but I don't think we should be using it at all

ghislainbourgeois:
You are right, I am closing this pull request and modifying the get_dns_facts in its place.

PR #13102, opened 2015-11-09
New method template_no_cache for template object.

Yannig:
This method do not compute hash and is used when you know that the result does not need to be cached (like with_items loop).
It also result in a small speed improvement.

Here's some numbers:

| playbook | before 7e04947 | last devel | after |
| --- | --- | --- | --- |
| issue-vars.yml | 3.2s | 6.1s | 2.64s |
| no-vars-mapping.yml | 2.5s | 1.8s | 1.9s |
| real-playbook.yml | 11.5s | 25.4s | 8.9s |

And of course, the yum module with item is working :)

### DIFF
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -256,7 +256,7 @@ class TaskExecutor:
                     variables['item'] = item
                     if self._task.evaluate_conditional(templar, variables):
                         if templar._contains_vars(name):
-                            new_item = templar.template(name)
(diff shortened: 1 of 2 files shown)


bcoca:
i would just add a cache=True/false parameter to template

jimi-c:
+1 to @bcoca's suggestion.

Yannig:
@bcoca, @jimi-c There's now a new cache parameter for template.

jimi-c:
Merged, thanks!

PR #13106, opened 2015-11-10
the raise syntax no longer accepts comma-separated arguments

aliceinwire:
In Python 3 exception objects are only available via the as
keyword, which was introduced in 2.6

### DIFF
--- a/lib/ansible/module_utils/ec2.py
+++ b/lib/ansible/module_utils/ec2.py
@@ -209,13 +209,13 @@ def ec2_connect(module):
     if region:
         try:
             ec2 = connect_to_aws(boto.ec2, region, **boto_params)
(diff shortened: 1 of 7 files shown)


sivel:
The files in `module_utils` support Python 2.4.  Changing this syntax will break Python 2.4 compatibility which is indicated by the failing Travis tests.

aliceinwire:
but you are breaking all python 3 by keeping as it is

sivel:
We are working on making Ansible Python 3 compatible from a controller perspective, however we are not yet targeting Python 3 support for remote targeted nodes.  The `module_utils` directory is used for execution on the remote nodes and at the moment must maintain 2.4 compatibility.

In a future release we will evaluate options for supporting Python 3 on target nodes but are not doing so quite yet.

aliceinwire:
:+1:

Issue #13095, opened 2015-11-09
Exit-Json should be less nested


davidobrien1985:
##### ISSUE TYPE
Bug Report

##### COMPONENT NAME
lib/ansible/modules/windows/setup.ps1

##### ANSIBLE VERSION
2.2

##### SUMMARY
Referencing: https://groups.google.com/forum/#!topic/ansible-devel/n8KMlvi9WJ4

Exit-Json by default converts strings up to 99 levels. I can't think of any scenario where this is needed. It actually makes my environment fail and causes PowerShell to die on me.

setup.ps1 (after my PR is eventually pulled) and win_msi are two modules that can potentially return huge strings.

IMO this is a big bug that should be fixed soon.

trondhindenes:
My suggestion is that we default this to 10 levels, that should be enough. Also, we can add a param to exit-json where the number of levels can be specified, so modules that know they potentially output a lot of nested levels can tweak the value at run-time. I can submit a PR later for this.

jhawkesworth:
Just going through old windows tickets and wondering if this one is still needed?
As mentioned on https://github.com/ansible/ansible/pull/13117 the eventual plan is to avoid ConvertTo-Json altogether.

ansibot:
@davidobrien1985 Greetings! Thanks for taking the time to open this issue. In order for the community to handle your issue effectively, we need a bit more information. 

Here are the items we could not find in your description:
- issue type
- ansible version
- component name

Please set the description of this issue with this template:
https://raw.githubusercontent.com/ansible/ansible/devel/.github/ISSUE_TEMPLATE.md

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

dagwieers:
@jhawkesworth Agreed.
@nitzmahone Is this something we want to fix ? Do we have to make a plan ? Or can we close this altogether ?

nitzmahone:
This will likely come for free with the Nano server support- I've got a couple different prototypes that have ditched the Convert-*Json cmdlets, but nothing that works across the board.

dagwieers:
@nitzmahone OK, good to know it's still being worked on !

dagwieers:
@nitzmahone This is your trimestrial update request ;-) How far are we with Nano support ? When can we expect to close this ticket ?

ansibot:
cc @jborean93
[...]

PR #13117, opened 2015-11-10
exit-json and fail-json improvements / bugfix

trondhindenes:
exit-json/fail-json now allows to module dev to specify the json depth (from 0-99), defaulting to 10. This is done to overcome a bug where PowerShell will hang indefinetely if complex objects are converted to json when depth is set to 99.

Ref #13095 and discussion at https://groups.google.com/forum/#!topic/ansible-devel/n8KMlvi9WJ4

### DIFF
--- a/lib/ansible/module_utils/powershell.ps1
+++ b/lib/ansible/module_utils/powershell.ps1
@@ -62,23 +62,35 @@ Function Set-Attr($obj, $name, $value)
 # Helper function to convert a powershell object to JSON to echo it, exiting
 # the script
 # Example: Exit-Json $result
-Function Exit-Json($obj)
+Function Exit-Json
 {
+    (
+    $obj,
+    
+    [ValidateLength(0,99)]
+    [int]$Depth=10)
     # If the provided $obj is undefined, define one to be nice
     If (-not $obj.GetType)
     {
         $obj = New-Object psobject
     }
 
-    echo $obj | ConvertTo-Json -Compress -Depth 99
+    echo $obj | ConvertTo-Json -Compress -Depth $Depth
     Exit
 }
 
 # Helper function to add the "msg" property and "failed" property, convert the
 # powershell object to JSON and echo it, exiting the script
 # Example: Fail-Json $result "This is the failure message"
-Function Fail-Json($obj, $message = $null)
+Function Fail-Json
 {
+    (
+        $obj,
+        $message = $null,
+        
+        [ValidateLength(0,99)]
+        [int]$Depth=10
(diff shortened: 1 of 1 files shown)


trondhindenes:
Any objections to getting this merged in?

cchurch:
Running `ansible -i inventory.winrm windows -m setup -vvv` using 337384d091 against 2008R2 I get the following error:

```
At C:\Users\Administrator\AppData\Local\Temp\ansible-tmp-1448375642.2-71994590113742\setup.ps1:85 char:5
+     $obj,
+     ~~~~~
ec2-xx-xxx-xxx-xxx.compute-1.amazonaws.com | FAILED! => {
    "changed": false,
    "failed": true,
    "invocation": {
        "module_args": {},
        "module_name": "setup"
    },
    "msg": "Attribute cannot be added because it would cause the variable Depth with value 0 to become invalid."
}
```

bcoca:
fix looks fine, but default might need to be increased. Since setup is probably the module with the deepest structures can we look at adjusting the level to at least what it uses?

nitzmahone:
@trondhindenes I get the same error on all modules- the problem is that the "Param" line is missing from the converted parameter sections. 

I wonder if we'd see the same hanging issue using JavaScriptSerializer against arbitrary objects? I prototyped a branch with that so we could stop using PSObject, but haven't gotten back to it in awhile...

trondhindenes:
Thanks for the feedback guys. Looking into it.
@bcoca setup currently outputs an object with 3 levels, so I thought 10 would plenty. I can look at dynamically getting the object depth and setting that as the json output depth.

jhawkesworth:
Looking back at how this ticket came about.  The original issue https://groups.google.com/forum/#!topic/ansible-devel/n8KMlvi9WJ4 and tickets for this https://github.com/ansible/ansible-modules-core/pull/2563 (merged) and https://github.com/ansible/ansible-modules-core/issues/2330 (closed)
as the issue was the contents of the pipeline (containing many large objects).

While it would be nice to make the depth overridable, as it wasn't really the cause of the issue, should we close this one off?

cc: @nitzmahone

nitzmahone:
Deep pipeline returns are definitely the cause- I'm a little nervous to just cap it off though. Shouldn't come up much in the course of normal things, and I'm hoping to switch us off ConvertTo-Json anyway (to .NET 4.0 JavascriptSerializer) with the new module API. That'll just use simple dicts for args and returns like Python API and eliminate the need for custom props on PSObjects. I'll close for now.

PR #13127, opened 2015-11-11
Elevate privileges when checking for file existence

willthames:
Certain OSs default to home directories as 0700
Without elevating privileges, checking for file existence
fails.

### DIFF
--- a/test/integration/roles/test_become/tasks/main.yml
+++ b/test/integration/roles/test_become/tasks/main.yml
@@ -24,6 +24,8 @@
     state: touch
 
 - name: check that the path in the user's home dir was created
+  become: True
+  become_user: "{{ become_test_user }}"
   stat:
     path: "~{{ become_test_user }}/foo.txt"
   register: results
@@ -41,6 +43,8 @@
     dest: "~/bar.txt"
 
 - name: check that the path in the user's home dir was created
+  become: True
+  become_user: "{{ become_test_user }}"
   stat:
     path: "~{{ become_test_user }}/bar.txt"
   register: results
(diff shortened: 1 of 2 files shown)


willthames:
I might be running tests wrongly, of course. 

But `make destructive` doesn't really work for me, even if I add a bunch of configuration.

Does `make destructive` assume connection as root user?

I tried to get the tests running using vagrant user against a vagrant VM, and with the changes here, `test_sudo` and `test_become` all worked. But later tests failed, and I realised I'd need to add `become` to all the tasks in `test_service`, so I wonder if there's an assumption that the whole thing is run as root, and if so, where that assumption is best reified (e.g. adding `remote_user: root` or `become_user: root` to `destructive.yml`)

bcoca:
b

On Wed, Nov 11, 2015 at 6:02 AM, Will Thames <notifications@github.com>
wrote:

> I might be running tests wrongly, of course.
>
> But make destructive doesn't really work for me, even if I add a bunch of
> configuration.
>
> Does make destructive assume connection as root user?
>
> I tried to get the tests running using vagrant user against a vagrant VM,
> and with the changes here, test_sudo and test_become all worked. But
> later tests failed, and I realised I'd need to add become to all the
> tasks in test_service, so I wonder if there's an assumption that the
> whole thing is run as root, and if so, where that assumption is best
> reified (e.g. adding remote_user: root or become_user: root to
> destructive.yml)
>
> —
> <https://github.com/ansible/ansible/pull/13127#issuecomment-155791194>.
>



-- 
Brian Coca

bcoca:
both make non_destructive and destructive assume you run as root for many
of the tests. I've been fixing some of the non_destructive to be able to
run as non root user but there are still plenty of tasks that fail.

## 

Brian Coca

willthames:
@bcoca, does that mean you approve of this change and would like more of the same?

bcoca:
yep, though i was just using root as the privileged user

willthames:
Sure, that would be normal, but for these two modules, becoming a different user is pretty much the point.

PR #13132, opened 2015-11-12
now we don't try to convert types if certain filters

bcoca:
specially those specifically to return a formated string

### DIFF
--- a/lib/ansible/constants.py
+++ b/lib/ansible/constants.py
@@ -261,6 +261,7 @@ GALAXY_SCMS                    = get_config(p, 'galaxy', 'scms', 'ANSIBLE_GALAXY
 
 # characters included in auto-generated passwords
 DEFAULT_PASSWORD_CHARS = ascii_letters + digits + ".,:-_"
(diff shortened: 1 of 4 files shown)


willthames:
my only concern with this is if people write their own filters, they won't be able to add them to the hardcoded list (I don't have this problem, but we do write our own filters, so I can see how this could happen). Can we make it configurable (with the list as is as default)?

bcoca:
good point, will do

lib/ansible/constants.py
```diff
@@ -261,6 +261,7 @@ def load_config_file():
 
 # characters included in auto-generated passwords
 DEFAULT_PASSWORD_CHARS = ascii_letters + digits + ".,:-_"
+DEFAULT_DONT_TYPE_FILTERS = get_config(p, 'jinja2', 'dont_type_filters', 'ANSIBLE_DONT_TYPE_FILTERS', ['to_json', 'to_nice_json', 'to_yaml', 'ppretty', 'json'], islist=True )
```
willthames: The name isn't obvious to me.

How about DEFAULT_AS_IS_FILTERS, 'as_is_filters' and 'ANSIBLE_AS_IS_FILTERS' or similar?

This should be documented too, but I can do that separately if need be.

PR #13135, opened 2015-11-12
fix 'coveralls' code coverage report

zarnovican:
Submitting code-coverage report to `coveralls.io` probably never worked. This is adding the necessary command. Integration tests were also breaking this feature. Either because they 'cd' to different directory or hack PATH/PYTHONPATH. I didn't look into it too deep. Instead, I've put them into sub-shell for protection.

While working on this change, it occurred to me that it would be prudent to separate unittest and integration tests into separate jobs.. just an idea.

_Note_: To enabled coverage reports, someone with admin rights to `ansible/ansible` would need to enable it on https://coveralls.io/repos/new.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ addons:
     packages:
       - python2.4
 install:
-  - pip install tox PyYAML Jinja2 sphinx
+  - pip install tox PyYAML Jinja2 sphinx coveralls
 script:
 # urllib2's defaults are not secure enough for us
 - ./test/code-smell/replace-urlopen.sh .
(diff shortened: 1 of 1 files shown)


gundalow:
rebase needed.

I believe one of the reasons coveralls was disabled is due to the increased execution time. Could you provide details on how longer extra the tests take with this enabled?

Thanks in advance

zarnovican:
Hi John,

this PR was _fixing_ "coveralls", which was enabled at that time.. see

```
after_success:
    - coveralls
```

Since then, Brian has disabled them for good 92f29746d97abc8f096532fa4d95f8316b72a79b, so this PR is no longer relevant. Closing.

gundalow:
Thanks for revisiting this.

PR #13142, opened 2015-11-12
Make jsonfile fact cache handle fancy paths properly

### LINKED ISSUE #13094: fact_caching_connection don't handle path expansion of ~ nor $HOME
Version: 1.9 but probably also devel.

When I configure a fact caching directory using the jsonfile cache plugin, I am specifying its path using the ~ or $HOME variable. This works with other parameters like retry_files_save_path but not with fact_caching_connection ... Ansible fails with a message that the directory does not exist. What is false.

sontags:
Basically fixes #13094

os.path.expanduser as well as s.path.expandvars is applied before before storing C.CACHE_PLUGIN_CONNECTION as instance attribute.

### DIFF
--- a/lib/ansible/plugins/cache/jsonfile.py
+++ b/lib/ansible/plugins/cache/jsonfile.py
@@ -50,7 +50,7 @@ class CacheModule(BaseCacheModule):
 
         self._timeout = float(C.CACHE_PLUGIN_TIMEOUT)
         self._cache = {}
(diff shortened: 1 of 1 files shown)


davidfischer-ch:
Fancy!

davidfischer-ch:
Sorry if it's already the case but just if its possible to cherry-pick this to v1.9 branch it will be greatly appreciated. Thanks.

PR #13145, opened 2015-11-12
lxc connection plugin

Mic92:
This plugin has been tested in production for more than 1 year.
See https://github.com/Mic92/ansible-lxc for reference.

### DIFF
--- a/lib/ansible/plugins/connection/lxc.py
+++ b/lib/ansible/plugins/connection/lxc.py
@@ -1,0 +1,215 @@
+# (c) 2015, Joerg Thalheim <joerg@higgsboson.tk>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import os
+import shutil
+import traceback
+import select
+import fcntl
+import errno
+from ansible import errors
+from ansible import constants as C
+from ansible.plugins.connection import ConnectionBase
+from ansible.utils.unicode import to_bytes
+
+HAS_LIBLXC = False
+try:
+    import lxc as _lxc
+    HAS_LIBLXC = True
+except ImportError:
+    pass
+
+class Connection(ConnectionBase):
+    ''' Local lxc based connections '''
+
+    transport = 'lxc'
+    has_pipelining = True
+    become_methods = frozenset(C.BECOME_METHODS)
+
+    def __init__(self, play_context, new_stdin, *args, **kwargs):
+        super(Connection, self).__init__(play_context, new_stdin, *args, **kwargs)
+
+        self.container_name = self._play_context.remote_addr
+        self.container = None
+
+    def _connect(self):
+        ''' connect to the lxc; nothing to do here '''
+        super(Connection, self)._connect()
+
+        if not HAS_LIBLXC:
+            msg = "lxc bindings for python2 are not installed"
+            raise errors.AnsibleError(msg)
+
+        if self.container:
+            return
+
+        self._display.vvv("THIS IS A LOCAL LXC DIR", host=self.container_name)
+        self.container = _lxc.Container(self.container_name)
+        if self.container.state == "STOPPED":
+            raise errors.AnsibleError("%s is not running" % self.container_name)
+
+    def _communicate(self, pid, in_data, stdin, stdout, stderr):
+        buf = { stdout: [], stderr: [] }
+        read_fds = [stdout, stderr]
+        if in_data:
+            write_fds = [stdin]
+        else:
+            write_fds = []
+        while len(read_fds) > 0 or len(write_fds) > 0:
+            try:
+                ready_reads, ready_writes, _ = select.select(read_fds, write_fds, [])
+            except select.error as e:
+                if e.args[0] == errno.EINTR:
+                    continue
+                raise
+            for fd in ready_writes:
+                in_data = in_data[os.write(fd, in_data):]
+                if len(in_data) == 0:
+                    write_fds.remove(fd)
+            for fd in ready_reads:
+                data = os.read(fd, 32768)
+                if not data:
+                    read_fds.remove(fd)
+                buf[fd].append(data)
+
+        (pid, returncode) = os.waitpid(pid, 0)
+
+        return returncode, b"".join(buf[stdout]), b"".join(buf[stderr])
+
+    def _set_nonblocking(self, fd):
+        flags = fcntl.fcntl(fd, fcntl.F_GETFL) | os.O_NONBLOCK
+        fcntl.fcntl(fd, fcntl.F_SETFL, flags)
+        return fd
+
+    def exec_command(self, cmd, in_data=None, sudoable=False):
+        ''' run a command on the chroot '''
+        super(Connection, self).exec_command(cmd, in_data=in_data, sudoable=sudoable)
+
+        executable = to_bytes(self._play_context.executable, errors='strict')
+        local_cmd = [executable, '-c', to_bytes(cmd, errors='strict')]
+
+        read_stdout, write_stdout = None, None
+        read_stderr, write_stderr = None, None
+        read_stdin, write_stdin   = None, None
+
+        try:
+            read_stdout, write_stdout = os.pipe()
+            read_stderr, write_stderr = os.pipe()
+
+            kwargs = {
+                'stdout': self._set_nonblocking(write_stdout),
+                'stderr': self._set_nonblocking(write_stderr),
+                'env_policy': _lxc.LXC_ATTACH_CLEAR_ENV
+            }
+
+            if in_data:
+                read_stdin, write_stdin = os.pipe()
+                kwargs['stdin'] = self._set_nonblocking(read_stdin)
+
+            self._display.vvv("EXEC %s" % (local_cmd), host=self.container_name)
+            pid = self.container.attach(_lxc.attach_run_command, local_cmd, **kwargs)
+            if pid == -1:
+                msg = "failed to attach to container %s" % self.container_name
+                raise errors.AnsibleError(msg)
+
+            write_stdout = os.close(write_stdout)
+            write_stderr = os.close(write_stderr)
+            if read_stdin:
+                read_stdin = os.close(read_stdin)
+
+            return self._communicate(pid,
+                                     in_data,
+                                     write_stdin,
+                                     read_stdout,
+                                     read_stderr)
+        finally:
+            fds = [read_stdout,
+                   write_stdout,
+                   read_stderr,
+                   write_stderr,
+                   read_stdin,
+                   write_stdin]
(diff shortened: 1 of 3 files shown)


sivel:
At first glance this plugin is not compatible with Ansible 2.0/devel.

Mic92:
Sorry. I submitted the wrong file to the wrong location.
Ok will adapt it to ansible 2.0, as soon i have time to test.

Mic92:
Update:
- Adapted and tested with current ansible master.
- Added support for pipelining (which gives a nice speedups)

Could you please review the code written so far?
It depends on python2-lxc like the existing lxc container module.

Mic92:
Something left to do?

bcoca:
We just need time to review, test and merge, this might take a bit as we are concentrating on 2.0 bugs to finalize the release. Sorry for the inconvenience.

fghaas:
@bcoca For what it's worth I've been using this plugin for a while now, and from my perspective it works very well (and runs _much_ faster than using ssh to shell into the containers). Just my two cents. :)

lib/ansible/plugins/connection/lxc.py
```diff
@@ -0,0 +1,210 @@
...
+        finally:
+            if src_file:
+                src_file.close()
+
+    def fetch_file(self, in_path, out_path):
+        ''' fetch a file from lxc to local '''
+        super(Connection, self).fetch_file(in_path, out_path)
+        display.vvv("FETCH %s TO %s" % (in_path, out_path), host=self.container_name)
+
+        if not os.path.exists(in_path):
```
abadger: How does this work?  in_path should a path relative to the root inside the container.  If we check for its existence here we end up checking whether it exists on the host, don't we?

abadger:
Note: I didn't give a comprehensive review to this (like bcoca, I'm working on bugfixing for 2.0 right now), just noted some thing that didn't seem quite right.

lib/ansible/plugins/connection/lxc.py
Mic92: This is correct. I removed this check.

fghaas:
@Mic92 I wonder if you're seeing the same issue, but for me this is currently broken on Ubuntu 15.10. Evidently something changed in the (unofficial) Python 2 bindings for lxc, so anything that actually requires interaction with a container currently breaks with a segfault:

```
IPython 2.3.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import lxc

In [2]: container = lxc.Container("mycontainer")

In [3]: container.get_ips()
Segmentation fault
```

If you run this with the officially supported Python 3 bindings, everything is fine.

This creates a silly conundrum: Ansible doesn't run in Python 3, and you can't interact with lxc containers from Python 2 anymore.

Are you seeing this too? Do you have a workaround?

Mic92:
Mhm. I rembered having issues with a broken python2 binding at some point, however in my case recompiling python2-lxc solved the problem. At the moment I don't have a ubuntu to test. Maybe you can recompile the flags with debugs flags enabled and catch coredump. Then open it in gdb and print the stacktrace. Does it happen with earlier commits of python2-lxc and the same lxc version?

fghaas:
@Mic92 See Stephane's input on lxc/python2-lxc#8, and my reply there. Yes it does appear that this is just a problem with the Ubuntu package, not an inherent problem in liblxc or the LXC Python 2 bindings. At any rate, it definitely shouldn't be a show stopper for getting this merged.

@abadger or @bcoca, now that 2.0 is out would you consider coming back to this one?

bcoca:
in bugfix mode for 2.0.1 right now, will get back to 2.1.0 features soon

fghaas:
@bcoca Any chance that this could be revisited?

fghaas:
@abadger @bcoca Well, now 2.1 is out and this PR apparently fell through the cracks again. So that was "not now, we're working on 2.0" followed by "not now, we're working on 2.0.1", then ignoring this for 2.1. What happens next? If you don't want this, then at least have the grace to tell @mic92 about it.

bcoca:
@fghaas I'm really sorry, I meant what I said, but I totally forgot about this. This is not a reflection on the interest on this PR.

I know it is no consolation,  there were several other PRs in the same state I missed for the 2.1 feature merges. I'm adding this to the next public meeting's agenda to make sure at least other eyes go over it and it gets discussed by core. https://github.com/ansible/community/issues/107

mattclay:
@Mic92: For testing this, could you add it to the integration tests? Specifically, add a section to test_connection.inventory for lxc and then make sure the tests pass by running "make test_connection".

abadger:
At today's meeting, @bcoca volunteered to test this and I brought up a potential symlink security issue.  I looked more closely afterwards with @Mic92's help and discovered that the security issue **does not** exist.  The use of shutil.copyfileobj() happens via the call to container.attach_wait() so it's happening inside of the container's more limited environment.  This is what's needed to address the symlink issue so current implementation is good.

@bcoca, this is ready for your review and then merge.

fghaas:
Thanks for picking this back up everyone.

Mic92:
I will setup something tomorrow.

Mic92:
@bcoca I have already added something to `test_connection.py`. Is there anything else left to do? There is not `test_connection` make target in the current devel branch.

mattclay:
@Mic92: See my previous comment on adding connection tests.

test/integration/test_connection.inventory
```diff
@@ -55,6 +55,13 @@ lxd-no-pipelining ansible_ssh_pipelining=false
 ansible_host=centos-7-amd64
 ansible_connection=lxd
 
+[lxc]
+lxc-pipelining ansible_ssh_pipelining=true
+lxc-pipelining ansible_ssh_pipelining=false
```
mattclay: The second entry is missing "-no", so the above should be:

```
[lxc]
lxc-pipelining    ansible_ssh_pipelining=true
lxc-no-pipelining ansible_ssh_pipelining=false
```

Mic92:
added integration tests, works on my machine (tm) - see comments for details. Tests fail for the same reason as the current HEAD.

lib/ansible/plugins/connection/lxc.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def exec_command(self, cmd, in_data=None, sudoable=False):
+        ''' run a command on the chroot '''
+        super(Connection, self).exec_command(cmd, in_data=in_data, sudoable=sudoable)
+
+        if C.DEFAULT_EXECUTABLE:
+            executable = C.DEFAULT_EXECUTABLE.split()[0]
+        else:
+            executable = '/bin/sh'
+        local_cmd = [to_bytes(executable, errors='strict'), '-c', to_bytes(cmd, errors='strict')]
```
mattclay: New connection plugins should be using the recently added `self._play_context.executable` for executable. Take a look at the lxd plugin for an example.

lib/ansible/plugins/connection/lxc.py
```diff
@@ -0,0 +1,225 @@
...
+        super(Connection, self)._connect()
+
+        if not HAS_LIBLXC:
+            msg = "lxc bindings for python2 are not installed"
+            raise errors.AnsibleError(msg)
+
+        if self.container:
+            return
+
+        display.vvv("THIS IS A LOCAL LXC DIR", host=self.container_name)
```
mattclay: Use `self._display` instead of `display`, which will also allow you to get rid of the Display import.

Mic92:
use `self._display` and `self._play_context.executable` now

mattclay:
Tested with `TEST_FLAGS='-l lxc' make test_connection`:

```
lxc-no-pipelining          : ok=10   changed=7    unreachable=0    failed=0
lxc-pipelining             : ok=10   changed=7    unreachable=0    failed=0
```

fghaas:
Sweet, thanks! Does that mean #13433 is now ready to be merged, too?

fghaas:
@mattclay I just noticed that this connection plugin is not included in the 2.1.1 builds in [the ansible PPA](https://launchpad.net/~ansible/+archive/ubuntu/ansible), built almost 2 months after this was merged. Any particular reason for that, or is that a build system oversight?

evgeni:
@fghaas as far as I can see, the merge was done into devel, but not cherry-picked into stable-2.1?

fghaas:
@evgeni — that sounds logical, but then what exactly does the `stable-2.1` milestone mean on this PR? I thought that meant it _would_ get pulled into 2.1; is that an incorrect assumption?

evgeni:
@fghaas that I don't know. let's see what people from the @ansible team will say :)

fghaas:
Roger. Thanks for looking into this!

mattclay:
@fghaas Unfortunately this plugin wasn't merged until after 2.1 stable was released, despite being marked for the 2.1 milestone. It will be included in the upcoming 2.2 release.

Mic92:
@mattclay unless you forget it again...

mattclay:
@Mic92 Since this has already been merged to devel, it will automatically become part of the 2.2 release when that branch is created.

It didn't make it into the 2.1 release because the core team hadn't reviewed and merged before the 2.1 feature freeze. As bcoca mentioned earlier, we're sorry that this was missed for the 2.1 release.

spinitron:
is there any documentation besides Mic92/ansible-lxc/README?

fghaas:
@spinitron It's really quite simple to use. Assuming you have a container named `mycontainer` running on your host, here's what you would add to your Ansible inventory:

```ini
[containers]
mycontainer ansible_connection=lxc
```

Then, try `ansible -i <path to your inventory file> -m ping`. If you have `python-lxc` installed, Ansible should directly attach to the LXC container, rather than using SSH to get in.

There is also a dynamic inventory script in `contrib/inventory/lxc_inventory.py` that will enumerate all your local containers and group them by their `lxc.group` configuration option.

One thing you do need to keep in mind is that the `lxc` connection plugin does care about whether a container is running as `root` (privileged) or non-`root` (unprivileged). If it is a privileged container, only `root` is normally allowed to attach the container, which means you must also invoke `ansible` or `ansible-playbook` as `root`. If your container is running unprivileged, then you must invoke ansible as the same user that also runs the container.

spinitron:
Thanks, Florian.

I'm still a little confused about what's happening where. I have, in a simplified model and using terminology from Ansible docs

- a local *control machine* that 
    - runs ansible-playbook
    - has an inventory file and a playbook
    - in my case cannot run lxc since it's not linux
- a remote *managed node* that
    - I can control via Ansible directly
    - runs lxc
    - has a started container called `mycontainer` that i want to control using Ansible

Should python2-lxc be on the control machine or the managed node or both?

I guess `contrib/inventory/lxc_inventory.py` runs on the managed mode, so it would be remote not local, as you described it. It suggests python2-lxc is on the managed node if not also the control machine.

Then what you say about running ansible-playbook as root surprises me. Is it not enough that the ssh user of the connection to the managed node runs as root?

mattclay:
@spinitron For managing containers on a remote host, you might want to look at using lxd instead. Ansible has a separate lxd connection plugin, which will let you manage remote lxd containers.

spinitron:
Should I understand that this lxc plugin is only for managing containers running on the control machine?

Mic92:
@spinitron in case of the lxc plugin, the control machine must be the host running the lxc container. And the managed nodes are the lxc container.

fghaas:
@spinitron, in the scenario you describe you might just as well connect to your LXC containers via the default `ssh` connection, since you're already going via the network anyway. What's nice about the `lxc` connection plugin is that it does *not* go via SSH but uses the Python equivalent of `lxc attach` directly — but, obviously, that only works for a locally running LXC, which is exactly what this was built for.

spinitron:
@fghaas none of my containers have or need public addresses or SSH servers. So I have to weigh managing that plus port forwarding that I don't really want.

So far https://github.com/chifflier/ansible-lxc-ssh seems to be working for me but I haven't tested it very much yet.

PR #13146, opened 2015-11-13
fix gce module error reporting

nitzmahone:
Error reporting was broken for GCE modules- pprint didn't work with exceptions, so you'd always get "Unexpected response: {}" instead of the real error.


sivel:
@franckcuny @erjohnso

jimi-c:
Merged, thanks!

PR #13147, opened 2015-11-13
Increase the escalation prompt timeout

### LINKED ISSUE #12916: Wrong error message when unable to connect via ssh
## Version:

ansible-playbook 2.0.0 (devel 8f77dd1cf2) last updated 2015/10/26 11:46:26 (GMT -400)
  lib/ansible/modules/core: (detached HEAD 06f301b05b) last updated 2015/10/26 11:46:28 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 405c3cb2c7) last updated 2015/10/26 11:46:30 (GMT -400)
## Issue:

When the server is not online, the error message is confusing.
## What happens now:

Get the error message:

```
fatal: [10.10.67.157]: FAILED! => {"failed": true, "msg": "ERROR! Timeout (10s) waiting for privilege escalation prompt: "}
```
## What should happen:

```
fatal: [10.10.67.157]: FAILED! => {"failed": true, "msg": "ERROR! Unable to connect to server "}
```

leedm777:
It was set to match the SSH connect timeout. Unfortunately, they would
race when we wouldn't connect, and the connect timeout usually failed.
This led to some misleading error messages.

Fixes #12916

### DIFF
--- a/lib/ansible/plugins/connection/ssh.py
+++ b/lib/ansible/plugins/connection/ssh.py
@@ -392,7 +392,10 @@ class Connection(ConnectionBase):
             become_error=False, become_nopasswd_error=False
         )
 
-        timeout = self._play_context.timeout
+        # select timeout should be longer than the connect timeout, otherwise
+        # they will race each other when we can't connect, and the connect
+        # timeout usually fails
+        timeout = 2 + self._play_context.timeout
         rpipes = [p.stdout, p.stderr]
         for fd in rpipes:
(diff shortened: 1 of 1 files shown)


jimi-c:
ping @amenonsen

bcoca:
does it really need to be double? just adding a few seconds should be enough

leedm777:
@bcoca Changed it from doubling to adding two seconds.

amenonsen:
OK, that looks fine to me. @leedm777 thanks for the analysis and fix.

I think the patch I proposed in #12916 should also be applied, so I'll submit that as a separate PR.

PR #13166, opened 2015-11-14
Patch to remove dependency on boto when only using boto3

juliedavila:
As discussed here: https://github.com/ansible/ansible-modules-extras/pull/1223

This allows the use of `get_aws_connection_info` w/o boto when using boto3 exclusively


bcoca:
no check for has_boto3 in the function? relying on caller is good enough?

juliedavila:
@bcoca I was debating on this. My thought was whatever module is using boto3/botocore should already be doing that check, it felt redundant so I didn't add an additional check here.  IIRC all modules that currently leverage boto3 are doing proper checks, but I could add it here to be ultra safe if you'd prefer.

bcoca:
normally i would think this overkill, but if we put this in the connection method it really does not need to go in anywhere else, nicer to have it centralized.

Etherdaemon:
Tested this and its working with ansible/ansible-modules-extras#1223. I'm also waiting on PR https://github.com/ansible/ansible/pull/13053 to move where the profile_name is called.

juliedavila:
Updated with an explicit check for HAS_BOTO3

Etherdaemon:
Just wondering if this will be merged soon as without it, the route53_facts and the other module where I was requested to update is broken as it relies on this patch.

Etherdaemon:
Looks like this isn't required anymore? I've checked in v2.0.0-0.8.rc3 and route53_facts, ec2_vpc_managed_ngw don't appear to fail anymore as a result of boto3

wimnat:
@defionscode @bcoca is this one ready to go? Would be useful for any new aws modules

wimnat:
Please shipit

juliedavila:
shipit

lib/ansible/module_utils/ec2.py
```diff
@@ -129,10 +130,14 @@ def get_aws_connection_info(module, boto3=False):
...
-            if not region:
-                region = boto.config.get('Boto', 'ec2_region')
+            if not boto3:
+                # boto.config.get returns None if config not found
+                region = boto.config.get('Boto', 'aws_region')
+                if not region:
+                    region = boto.config.get('Boto', 'ec2_region')
+            elif boto3 and HAS_BOTO3:
+                # here we don't need to make an additional call, will default to 'us-east-1' if the below evaluates to None.
+                region = botocore.session.get_session().get_config_variable('region')
```
bcoca, line 149: What happens when module asks for boto3 but HAS_BOTO3 is False? 

I believe we should either default to boto2 and give a warning or otherwise a fatal error.

juliedavila, line 149: If a module asks for boto3 and it's not installed then it should be an error, not fall back back to boto. I'll add the error handling

juliedavila:
updated

PR #13168, opened 2015-11-14
Do not try to load [group|host]_vars files if directory does not exist

Yannig:
If the parent directory does not exist, do not try to load various host or group file.
Result in a small speed improvement.


hloeffler:
:+1:

Yannig:
Close this PR after this PR was included: https://github.com/ansible/ansible/pull/13957

PR #13173, opened 2015-11-15
Handle unexpected database presence

willthames:
Check for database presence in a nice way, rather than
dropping the database.

Thankfully there was a syntax error in the previous
version, so no database would have been dropped.
There was no check for whether it succeeded or not.


jimi-c:
Merged, thanks!

Issue #12290, opened 2015-09-09
v2 regression: fetch/su is broken


amenonsen:
Using the latest unmodified devel (@e2e9319f), a task like this:

```
- fetch: src=/etc/motd dest=motd
  su: yes
```

produces this in the debug output:

```
<raven.toroid.org> EXEC ssh -C -tt -vvv -o ControlMaster=auto -o ControlPersist=60s -o ControlPath="/home/ams/.ansible/cp/ansible-ssh-%h-%p-%r" -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi
-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o ConnectTimeout=10 raven.toroid.org /bin/sh -c 'su  root -c "/bin/sh -c '"'"'echo BECOME-SUCCESS-nrkohdhsakbxamjlftxhmfgphtzxevls; rc=flag;
 [ -r /etc/motd ] || rc=2; [ -f /etc/motd ] || rc=1; [ -d /etc/motd ] && rc=3; python -V 2>/dev/null || rc=4; [ x"$rc" != "xflag" ] && echo "${rc}  "/etc/motd && exit 0; (python -c '"'"'"'"'"'"'"'"'import hashlib;
 BLOCKSIZE = 65536; hasher = hashlib.sha1();
afile = open("'"'"'"'"'"'"'"'"'/etc/motd'"'"'"'"'"'"'"'"'", "rb")
buf = afile.read(BLOCKSIZE)
while len(buf) > 0:
        hasher.update(buf)
        buf = afile.read(BLOCKSIZE)
afile.close()
print(hasher.hexdigest())'"'"'"'"'"'"'"'"' 2>/dev/null) || (python -c '"'"'"'"'"'"'"'"'import sha; BLOCKSIZE = 65536; hasher = sha.sha();
afile = open("'"'"'"'"'"'"'"'"'/etc/motd'"'"'"'"'"'"'"'"'", "rb")
buf = afile.read(BLOCKSIZE)
while len(buf) > 0:
        hasher.update(buf)
        buf = afile.read(BLOCKSIZE)
afile.close()
print(hasher.hexdigest())'"'"'"'"'"'"'"'"' 2>/dev/null) || (echo '"'"'"'"'"'"'"'"'0  '"'"'"'"'"'"'"'"'/etc/motd)'"'"'"'
 24251 1441789764.96724: Handling privilege escalation password prompt.
[...]

PR #13176, opened 2015-11-16
Fix su/doas support in make_become_cmd

amenonsen:


lib/ansible/playbook/play_context.py
```diff
@@ -424,7 +423,7 @@ def detect_su_prompt(data):
                     return bool(SU_PROMPT_LOCALIZATIONS_RE.match(data))
 
                 prompt = detect_su_prompt
-                becomecmd = '%s %s %s -c "%s -c %s"' % (exe, flags, self.become_user, executable, success_cmd)
+                becomecmd = '%s %s %s -c %s' % (exe, flags, self.become_user, pipes.quote("%s -c %s" % (shell, success_cmd)))
```
bcoca, line 475: I tried this before and it broke in some configurations, I need to look back to see exactly which

amenonsen:
OK, I haven't been able to reproduce any breakage, and I can't find anything in the commit logs or comments about the original problem, so I can't do anything useful to help here.

Note however that wrapping the output of `pipes.quote()` in double-quotes can't be right:

```
>>> print('sh -c "sh -c %s"' % pipes.quote('foo"bar'))
sh -c "sh -c 'foo"bar'"
      ^          ^
```

amenonsen:
See the discussion on #12290 for more details.

amenonsen:
I see that stable-2.0 has been branched. This change really should be included. I don't think P3 accurately reflects the importance of not trying to execute syntactically invalid commands in a remote root shell.

bcoca:
closed the original ticket by not issuing the incorrect request, but still want to look at this to fix the underlying problem so keeping this ticket open

lib/ansible/playbook/play_context.py
amenonsen, line 475: This may be unnecessary and obvious, but given that this patch still hasn't been merged, I wanted to point out that the change shown above from `"%s -c %s"` → `%s` doesn't mean that we stop calling `sh -c %s` (which is something I've discussed in the past). The entire shell+command invocation is quoted as one unit (which isn't clear without scrolling all the way to the right).

amenonsen:
Any news on merging this? I've been using this in production for a couple of weeks now, and haven't seen any additional problems.

amenonsen:
@bcoca I'm trying to understand why you think this needs "much more testing" before it can be merged. Could you please explain? From my perspective, it's _only_ a matter of not generating `"foo 'bar "baz" quux'"` on command lines, and quite frankly, I'm baffled that you think it might break something. Isn't quoting things properly basically what `pipes.quote()` is _for_?

amenonsen:
Now that 2.0 is out, can we please just get this merged?

PR #13185, opened 2015-11-16
Let ansible-galaxy run on py3

azazel75:
It still has some issues with https certificate verification on my machine, is it general?

### DIFF
--- a/lib/ansible/galaxy/api.py
+++ b/lib/ansible/galaxy/api.py
@@ -25,8 +25,12 @@ from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
 import json
-from urllib2 import quote as urlquote, HTTPError
-from urlparse import urlparse
+
+from ansible.compat import six
+from ansible.compat.six.moves.urllib.parse import quote as urlquote
+from ansible.compat.six.moves.urllib.error import HTTPError
+
+from ansible.compat.six.moves.urllib.parse import urlparse
 
 from ansible.errors import AnsibleError
 from ansible.module_utils.urls import open_url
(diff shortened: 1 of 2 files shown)


sivel:
`lib/ansible/module_utils/urls.py` must support python 2.4 and should not require `six` as this deviates from the only requirement of `json` as an external dependency.

sivel:
Additionally, `module_utils` cannot have a dependency on `ansible` itself, as `module_utils` is often part of the code shipped to a remote node, where `ansible` is most likely not installed.

bcoca:
As mentioned above anything in module_utils needs to be python 2.4 compatible.

azazel75:
So really the issue here is 'galaxy/apy.py' that uses open_url function that must be legacy-compatible, instead of using something that should run only on the "control" node, it's much more work to do...

azazel75:
geez, py2.4 compatible... stone age... :-(

azazel75:
but i understand if this is code sent to the node, is this documented somewhere that i missed?

abadger:
yeah, we won't be able to take this as-is so I'm going to close it.  You might want to talk with mgedmin and myself (abagder1999) on irc.freendoe.net #ansible-devel about what we can do about the module_utils directory and python3 compatibility.  Options I've thought of:
- Duplicate the code in module_utils that's needed on the controller so that the module_utils code can be python2.4 compatible and the other version can be python2.6+python3 compatible
  - Con: Duplication is a pain to maintain.  Does nothing to help with eventually porting modules to python3.
- Port the code to run on both python2.4 and python3
  - Con: Code is somewhat ugly.  Currently we'd have no way to depend upon six or other libraries that are useful in porting.  No way to write a library of generic porting utilities and include that in other module_utils code.
  - Possible help for this:
    - In 2.1 I plan on adding a way to load modules that uses actual python imports rather than preproccesing the files.  This will make it so we could potentially bundle six (or part of six) into the module_utils directory and use it.  Caveat: six does not support python2.4 anymore, we'd be forced to use an old version.

abadger:
@azazel75  yeah, this is code that is sent to the node.  It's mentioned here: http://docs.ansible.com/ansible/developing_modules.html  but it could be clearer where the module_utils directory lives and how it relates to both the controller and hte modules.  To avoid duplication of code, some of it is also used in the controller (urls.py has the complex url and tls handling stuff so it is one of the things being duplicated).

azazel75:
@abadger thanks, sorry for the noise

PR #13186, opened 2015-11-16
fix for dynamic (add_host) hosts not available in hostvars

nitzmahone:
Looks like there are two pattern caches that need to be cleared for this to work- added the second one.
Added integration tests for add_host to prevent future regressions.

If this PR runs its own tests, they may fail until #13184 is merged.


nitzmahone:
Bigger picture question- should we have two pattern caches in inventory? I haven't spent the time to figure out why we're not just using the inventory class member everywhere instead of the global (the global one appears to be hash based, the other isn't).

bcoca:
seems like one is per inventory source, to make loading each source faster, the other a global one ... probably we can do w/o the first by using the latter.

PR #13191, opened 2015-11-17
Add shared connection code for mysql modules

Jmainguy:


Jmainguy:
This needs to be merged before https://github.com/ansible/ansible-modules-core/pull/2447 can be. This allows all 3 to use the same connection method, which greatly simplifies the modules.

lib/ansible/module_utils/mysql.py
```diff
@@ -0,0 +1,56 @@
+# This code is part of Ansible, but is an independent component.
```
bcoca, line 1: module utils code does not use the GPL but a BSD license, this is so others can reuse this code when distributing independent 3rd party modules with any license they like.

bcoca:
license issue aside, looks good to me.

note: i would also add a mysql document fragment and add the connection options to it so we can have the documentation also in only 1 place and not duplicated across modules.  That is not a requirement for merging this PR in.

abadger:
Okay, merged. this refactor :-)

PR #13192, opened 2015-11-17
V2 final

ashutoshnarayan:


amenonsen:
@ashutoshnarayan I know you probably opened this by mistake (thanks for closing it too), but do you remember exactly what you did? I ask because this isn't the first time we've had a PR that's apparently the entire diff between devel and some older branch. We've been quite puzzled. Is there something in the Github UI that makes it easy to do this by mistake?

ashutoshnarayan:
@amenonsen Yes, this PR was opened by mistake, so I closed it immediately. I was actually comparing two branches. When Github UI showed there isn't anything to compare and provided sample comparisons below to check, I changed the release and hence this pull request got created.

PR #13193, opened 2015-11-17
Allow real addresses as dig lookup argument

javiplx:
This small change allows real IP addresses to be given as arguments to dig lookup. Without this change, `myres.query('127.0.0.1')` will generate "NXDOMAIN" on output. This change verifies whether the given value has inverse resolution, and in that case it returns the IP address given as input.
I know that adding '/PTR' after should solve my issue, but in that case the variable holding the addresses cannot be used anywhere else.

### DIFF
--- a/lib/ansible/plugins/lookup/dig.py
+++ b/lib/ansible/plugins/lookup/dig.py
@@ -201,7 +201,11 @@ class LookupModule(LookupBase):
                         ret.append(str(e))
 
         except dns.resolver.NXDOMAIN:
-            ret.append('NXDOMAIN')
+            try:
+                dns.reversename.from_address(domain)
+                ret.append(domain)
+            except:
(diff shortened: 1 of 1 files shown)


javiplx:
##### ISSUE TYPE
 - Feature Pull Request

calfonso:
The implementation assumes any failure on a reverse lookup should trigger the typical lookup. Is there anything you would like to submit that can evaluate the failure in more detail?

needs_info

javiplx:
Honestly, I don't remember and I'm not even using ansible right now. I've been using ansible with this patch for months, and is still in use at my previous workplace.
If you don't like my solution and/or don't think what I pretend is a desirable feature, just close this.

calfonso:
Hi!

Unfortunately, we're not sure if we want this feature in the program based upon the previous comment.

PR #13200, opened 2015-11-18
Make pipelining work with su/sudo+requiretty

amenonsen:
While writing http://toroid.org/ansible-ssh-pipelining it occurred to me that it's possible to make pipelining work with su and sudo+requiretty. Here's a working solution. I've been using this for several days now, and I can say that it's **very** pleasant to use this with pipelining and the recently-speeded-up v2.

Thanks to @mgedmin for coming up with the Python one-liner that I substituted for sed. (If the maintainers consider sed an acceptable dependency—as I do—the sed version in the commit message is shorter and easier to understand. Note that this is a change to the sh/ssh plugins, so it's not relevant to Windows.)

cc: @abadger

### DIFF
--- a/lib/ansible/plugins/connection/ssh.py
+++ b/lib/ansible/plugins/connection/ssh.py
@@ -241,7 +241,7 @@ class Connection(ConnectionBase):
 
         return self._command
 
-    def _send_initial_data(self, fh, in_data):
+    def _send_initial_data(self, fh, in_data, tty=False):
         '''
         Writes initial data to the stdin filehandle of the subprocess and closes
         it. (The handle must be closed; otherwise, for example, "sftp -b -" will
@@ -252,6 +252,8 @@ class Connection(ConnectionBase):
 
         try:
             fh.write(in_data)
+            if tty:
+                fh.write("__EOF__942d747a0772c3284ffb5920e234bd57__\n")
             fh.close()
         except (OSError, IOError):
             raise AnsibleConnectionFailure('SSH Error: data could not be sent to the remote host. Make sure this host can be reached over ssh')
@@ -314,7 +316,7 @@ class Connection(ConnectionBase):
 
         return ''.join(output), remainder
 
-    def _run(self, cmd, in_data, sudoable=True):
+    def _run(self, cmd, in_data, sudoable=True, tty=False):
         '''
         Starts the command and communicates with it until it ends.
         '''
@@ -322,25 +324,10 @@ class Connection(ConnectionBase):
         display_cmd = map(pipes.quote, cmd[:-1]) + [cmd[-1]]
         display.vvv('SSH: EXEC {0}'.format(' '.join(display_cmd)), host=self.host)
 
-        # Start the given command. If we don't need to pipeline data, we can try
-        # to use a pseudo-tty (ssh will have been invoked with -tt). If we are
-        # pipelining data, or can't create a pty, we fall back to using plain
-        # old pipes.
+        # Start the given command.
 
-        p = None
-        if not in_data:
-            try:
-                # Make sure stdin is a proper pty to avoid tcgetattr errors
-                master, slave = pty.openpty()
-                p = subprocess.Popen(cmd, stdin=slave, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-                stdin = os.fdopen(master, 'w', 0)
-                os.close(slave)
-            except (OSError, IOError):
-                p = None
-
-        if not p:
-            p = subprocess.Popen(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-            stdin = p.stdin
+        p = subprocess.Popen(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        stdin = p.stdin
 
         # If we are using SSH password authentication, write the password into
         # the pipe we opened in _build_command.
@@ -403,7 +390,7 @@ class Connection(ConnectionBase):
         # before we call select.
 
         if states[state] == 'ready_to_send' and in_data:
-            self._send_initial_data(stdin, in_data)
+            self._send_initial_data(stdin, in_data, tty)
             state += 1
 
         while True:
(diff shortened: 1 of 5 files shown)


lib/ansible/plugins/connection/ssh.py
```diff
@@ -253,6 +253,8 @@ def _send_initial_data(self, fh, in_data):
 
         try:
             fh.write(in_data)
+            if tty:
+                fh.write("__EOF__\n")
```
mgedmin: What are the chances this string occurs inside `in_data`?

amenonsen: AFAIK it's not valid Python, so I'd think zero? But if you have a better suggestion, I'm happy to change it.

amenonsen:
@abadger asked me to confirm that the `raw` and `script` modules will continue to work even if the remote machine doesn't have Python installed. Those commands are constructed differently, so this change should not affect them at all. I tested both and they do continue to work fine.

lib/ansible/plugins/connection/ssh.py
amenonsen: I've pushed another commit to add a random suffix to reduce the probability of collisions. Thanks for the suggestion.

mgedmin: I mentioned this on IRC, but for archival purposes I'll repeat myself here:

``` python
#!/usr/bin/python

... python code ...

SOME_SHELL_SCRIPT = '''
#!/bin/sh
...

cat << __EOF__
hello
world
__EOF__

... more shell code ...
'''

... more python code...
```

is valid Python code.

abadger:
Took a closer look at this, this morning --
- it looks like it changes the signature of connection plugins so we'll have to modify the other ones too?
- Does this implementation makes pipelining always happen?  Does that break the keep remote files functionality?
- I'm not sure if we ever pass things in on stdin in the current code to non-python commands but it seems like a possibility (for instance with custom non-python modules).  If so i think the shell.build_command() method would need to get its python interpreter from the caller rather than taking it out of shebang.  You can take a look at how the shell.checksum() code (and the method that calls it) does that.

amenonsen:
Thanks for taking a look.

> - it looks like it changes the signature of connection plugins so
>   we'll have to modify the other ones too?

It does not. _run and _send_initial_data are internal to ssh.py.

> - Does this implementation makes pipelining always happen?  Does that
>   break the keep remote files functionality?

It doesn't, and doesn't. It changes how pipelining works when we decide
to use it, which decision is taken in plugins/action/**init**.py based
on whether the transport supports it and so on. We change the default,
and we allow pipelining with su, but KEEP_REMOTE_FILES is tested and
taken into consideration separately.

See the diff to action/**init**.py for the full picture.

> - I'm not sure if we ever pass things in on stdin in the current code
>   to non-python commands but it seems like a possibility (for instance
>   with custom non-python modules).

The current code has no support for this, and I expect that non-Python
modules would break with pipelining turned on. (As I said earlier, raw
and script commands are built elsewhere, and are not affected by this
change at all.)

I agree that it would be nice to make this work; but the code that uses
shebang was already there, I just copied it verbatim to make sure that I
maintained compatibility above all, to give this a chance to be accepted
for 2.0 as a significant performance benefit.

But if you want me to clean it up further, I will.

(P.S. If this already has no chance of being accepted for 2.0, I would
really appreciate it if you said so, so that I can adjust my priorities
accordingly. Thanks.)

-- Abhijit

amenonsen:
(Sorry, I take that back about non-Python modules breaking with pipelining enabled. I can't say whether they would or not without trying it out, and I'm not in a position to do so easily.)

amenonsen:
I've rebased and added a commit to address @abadger's comment about the python interpreter.

abadger:
Reviewed and tested.  Looks good.  Want to squash this before I merge it to devel?

amenonsen:
Done.

abadger:
Thanks @amenonsen !

orenwolf:
@abadger Anything we can do to get this re-merged or investigated? I understand it was pulled from the build, which has lead to no end of confusion on our part since it shows merged here. :)

sourcejedi:
"Instead the pipelining patch should modify the other connection plugins to handle the new EOF marker."    https://github.com/ansible/ansible/issues/13409#issuecomment-162090736

Dmitriusan:
@amenonsen / @sourcejedi / @abadger,

Looks like this PR is merged
But latest Ansible documentation still statest that I should disable `requiretty`

http://docs.ansible.com/ansible/latest/intro_configuration.html#pipelining

What is the current status? Is documentation outdated?

tkeeler33:
@Dmitriusan - It doesn't look like it's in the current 2.4.2 release. 

Looking at the [2.5 develop](https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/connection/ssh.py) branch it does appear to be supported using the `use_tty` option in ssh settings. I have not tried this with pipelining however.

x-yuri:
From [the other issue][other-issue]:

[other-issue]: https://github.com/ansible/ansible/issues/13409#issuecomment-162091210

> note for people finding this from a web search: currently, the change introduced in PR #13200 has been reverted until we can find fixes for this and other issues. stemming from it.

I understand it's still reverted. But judging from [this conversation][requiretty-bug], I take it it's safe to disable it, like:

```yaml
    - name: sudo | disable requiretty (to enable pipelining + become)
      lineinfile:
        path: /etc/sudoers
        regexp: "^(\\s*Defaults\\s+requiretty\\s*)$"
        backrefs: yes
        line: '# \1'

```

[requiretty-bug]: https://bugzilla.redhat.com/show_bug.cgi?id=1020147

aioue:
Equivalent of @x-yuri's workaround on the official AWS CentOS AMIs:
```
- name: allow sudo without tty
  become: yes
  lineinfile:
    create: yes
    dest: /etc/sudoers.d/99-centos-cloud-init-requiretty
    line: 'Defaults:centos !requiretty'
    validate: 'visudo -cf %s'
    owner: root
    group: root
    mode: 0440
```

After running the above in it's own play with:
```
vars:
  ansible_ssh_pipelining: False
```

All subsequent runs can be actioned normally.

PR #13201, opened 2015-11-18
Fix DataLoader's docstring

ma82:
`DataLoader.__init__` doesn't take an argument named `vault_password`


jimi-c:
Merged, thanks!

PR #13203, opened 2015-11-18
Allow tree-ish versions for ansible-galaxy

willthames:
Rebase of #10620, retaining credit for @semifocused's work. For justification etc., see #10620.

### DIFF
--- a/lib/ansible/galaxy/role.py
+++ b/lib/ansible/galaxy/role.py
@@ -310,5 +310,3 @@ class GalaxyRole(object):
         }
         """
         return dict(scm=self.scm, src=self.src, version=self.version, name=self.name)
-
-
--- a/lib/ansible/playbook/role/requirement.py
+++ b/lib/ansible/playbook/role/requirement.py
@@ -190,6 +190,17 @@ class RoleRequirement(RoleDefinition):
         if rc != 0:
             raise AnsibleError ("- command %s failed in directory %s (rc=%s)" % (' '.join(clone_cmd), tempdir, rc))
 
+        if scm == 'git' and version:
+            checkout_cmd = [scm, 'checkout', version]
+            with open('/dev/null', 'w') as devnull:
+                try:
+                    popen = subprocess.Popen(checkout_cmd, cwd=os.path.join(tempdir, name), stdout=devnull, stderr=devnull)
+                except (IOError, OSError):
+                    raise AnsibleError("error executing: %s" % " ".join(checkout_cmd))
+                rc = popen.wait()
+            if rc != 0:
+                raise AnsibleError("- command %s failed in directory %s (rc=%s)" % (' '.join(checkout_cmd), tempdir, rc))
+
         temp_file = tempfile.NamedTemporaryFile(delete=False, suffix='.tar')
         if scm == 'hg':
             archive_cmd = ['hg', 'archive', '--prefix', "%s/" % name]
--- a/test/integration/galaxy_roles.yml
+++ b/test/integration/galaxy_roles.yml
@@ -3,7 +3,7 @@
   name: oracle_java7
 
 - src: git+http://bitbucket.org/willthames/git-ansible-galaxy
-  version: v1.6
+  version: pr-10620
 
 - src: http://bitbucket.org/willthames/hg-ansible-galaxy
   scm: hg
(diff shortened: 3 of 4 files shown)


semifocused:
Very helpful @willthames.  Many thanks for taking on the rebase and looks perfect.

lib/ansible/playbook/role/requirement.py
```diff
@@ -190,6 +190,17 @@ def scm_archive_role(src, scm='git', name=None, version='HEAD'):
...
 
+        if scm == 'git' and version:
+            checkout_cmd = [scm, 'checkout', version]
+            with open('/dev/null', 'w') as devnull:
+                try:
+                    popen = subprocess.Popen(checkout_cmd, cwd=os.path.join(tempdir, name), stdout=devnull, stderr=devnull)
+                except (IOError, OSError):
+                    raise AnsibleError("error executing: %s" % " ".join(checkout_cmd))
+                rc = popen.wait()
+            if rc != 0:
```
Lujeni, line 201: There is a reason do not use a simple test?
imho, it's more zen :)

``` python
if rc:
   ...
```

lib/ansible/playbook/role/requirement.py
```diff
@@ -190,6 +190,17 @@ def scm_archive_role(src, scm='git', name=None, version='HEAD'):
...
+                raise AnsibleError("- command %s failed in directory %s (rc=%s)" % (' '.join(checkout_cmd), tempdir, rc))
```
bcoca, line 202: it wold be helpful to get the error from the scm, but i would worry then that the wait call might block, it would require changing the code to account for this.

lib/ansible/galaxy/role.py
```diff
@@ -310,5 +310,3 @@ def spec(self):
         }
         """
         return dict(scm=self.scm, src=self.src, version=self.version, name=self.name)
-
-
```
bcoca, line 314: not sure why this file is changed

lib/ansible/playbook/role/requirement.py
semifocused, line 201: The approach in the modification regarding testing `rc` was "adopt-and-go" with existing coding conventions.  Shifting those conventions for enhanced zen may be desired, but the scope of doing so would be larger than this PR.  Considering occurences in the ansible code base of `rc != 0`:

```
~/git/ansible$ find . -type f | xargs -I FILE grep "rc != 0" FILE /dev/null | wc -l
353
```

Making this change may or may not be desired, if so might it best be undertaken in its own PR?

bcoca, line 201: i'm fine with the less 'zenliness' of `rc != 0` as this reflects the non pythonic shelling out to a command line utility

lib/ansible/galaxy/role.py
willthames, line 314: Not sure why there are a couple of empty lines at the bottom of the file either :). Probably don't need this though.

lib/ansible/playbook/role/requirement.py
willthames, line 202: Very similar (to the extent it might be worth abstracting it) code is in `role/requirement.py` in `scm_archive_role` twice. If it's good enough for `git clone` and `git archive`, surely it's good enough for `git checkout`.

willthames:
Please can you be very explicit about what is blocking the merge of this PR. I'm not convinced it `needs_revision` but maybe that's because we're fighting a review -> update -> review -> update -> review cycle that we've been around a dozen times and it would be nice to actually get this feature merged at last.

lib/ansible/playbook/role/requirement.py
bcoca, line 202: i've been meaning to fix that also, i finally let it go cause I did not have the time, i spent too much trying to restructure galaxy and roles. This is not a showstopper.

semifocused:
@bcoca ping on merge status?

bcoca:
i'll get back to this once we finish releasing 2.0, which should be very soon

semifocused:
Happy New Year @bcoca @willthames!!

Any further update on ability / willingness to get this merged?

bcoca:
still working on 2.0, will get to this probably next week

willthames:
@bcoca, can we finally have this included now that 2.0 has gone live?

scottcunningham:
Any updates on this? I'd love to be able to use it.

jimi-c:
Merged, thanks!

PR #1262, opened 2012-10-09
Use getopt instead of argparse

dagwieers:
On RHEL5 and RHEL6 there is not argparse, and 'make rpm' fails because of this.
With this change it works, and I can test and fix my own documentation...

PR #1261, opened 2012-10-09
playbook/tasks: allow to use a module's name instead of action:

abondis:
- allow to use a module's name instead of action: in playbook

```
- command: echo blah
- action: command echo blah
```

PR #13208, opened 2015-11-18
Add integration tests for zypper

agx:
Modeled after the yum tests but also tests local package installations
as fixed with PR#1256.

This depends on PRs #1256, #1261 and #1262 in ansible-modules-extra.


jimi-c:
Merged, thanks!

PR #13209, opened 2015-11-18
fast winrm put_file without size restrictions

nitzmahone:
passes all integration tests

@jimi-c as requested

### DIFF
--- a/lib/ansible/plugins/connection/winrm.py
+++ b/lib/ansible/plugins/connection/winrm.py
@@ -24,6 +24,8 @@ import os
 import re
 import shlex
 import traceback
+import json
+import xmltodict
 
 from ansible.compat.six.moves.urllib.parse import urlunsplit
 
@@ -44,6 +46,7 @@ except ImportError:
 
 from ansible.errors import AnsibleFileNotFound
 from ansible.plugins.connection import ConnectionBase
+from ansible.utils.hashing import secure_hash
 from ansible.utils.path import makedirs_safe
 from ansible.utils.unicode import to_bytes, to_unicode, to_str
 from ansible.utils.vars import combine_vars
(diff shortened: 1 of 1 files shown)


nitzmahone:
cc @jhawkesworth - Merry Christmas! ;)

jhawkesworth:
Awesome!  will test as soon as I can.

lib/ansible/plugins/connection/winrm.py
```diff
@@ -212,45 +241,83 @@ def exec_command(self, cmd, in_data=None, sudoable=True):
...
+            if temp_file.read(15).lower().startswith('#!powershell') and not out_path.lower().endswith('.ps1'):
+                out_path = out_path + '.ps1'
+
+        script = script_template.format(self._shell._escape(out_path))
+        cmd_parts = self._shell._encode_script(script, as_list=True, strict_mode=False)
+
+        result = self._winrm_exec(cmd_parts[0], cmd_parts[1:], stdin_iterator=self._put_file_stdin_iterator(in_path, out_path))
+        # TODO: improve error handling
+        if result.status_code != 0:
+            raise IOError(to_str(result.std_err))
```
cchurch: The convention in other connection plugins seems to be to raise an `AnsibleError` instead of `IOError`.

lib/ansible/plugins/connection/winrm.py
```diff
@@ -162,7 +179,19 @@ def _winrm_exec(self, command, args=(), from_exec=False):
...
+            command_id = self.protocol.run_command(self.shell_id, to_bytes(command), map(to_bytes, args), console_mode_stdin=(stdin_iterator == None))
+
+            # TODO: try/except around this, so we can get/return the command result on a broken pipe or other failure (probably more useful than the 500 that comes from this)
+            try:
+                if stdin_iterator:
+                    for (data, is_last) in stdin_iterator:
+                        self._winrm_send_input(self.protocol, self.shell_id, command_id, data, eof=is_last)
+            except:
+                # TODO: set/propagate an error flag, but don't throw (or include the command output in the exception)
+                pass
```
cchurch: I would suggest we at least display a message for this exception...

cchurch:
Aside from my comments above, :+1:

cchurch:
This should also fix https://github.com/ansible/ansible-modules-core/issues/1540

nitzmahone:
Updated for @cchurch feedback; should be ready to merge...

PR #13212, opened 2015-11-18
Recommend using pywinrm >= 0.1.1 from PyPI instead of GitHub version.

cchurch:
With @nitzmahone's recent updates to pywinrm (https://github.com/diyan/pywinrm/pull/74 and https://github.com/diyan/pywinrm/pull/76) included in pywinrm 0.1.1, I think we should now recommend using pywinrm from PyPI instead of GitHub.

### DIFF
--- a/docsite/rst/intro_windows.rst
+++ b/docsite/rst/intro_windows.rst
@@ -26,7 +26,7 @@ Installing on the Control Machine
 
 On a Linux control machine::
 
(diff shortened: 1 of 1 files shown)


jhawkesworth:
Good idea.  For what it's worth I have been running with pywinrm 0.1.1 on an ansible 1.9.4 instance for the last few days without any problems.

nitzmahone:
Works for me!

PR #13227, opened 2015-11-20
Allow groups by tag key values

jurgenhaas:
EC2 inventory can already be grouped by many properties like region, security group, tags keys and others. This patch introduces an additional option for getting inventory groups: tag values. This allows e.g. to create a tag "AnsibleGroups" with comma separated values like e.g. "webserver,dbserver" which will group the hosts into those tag values - and if expand_csv_tags is set to True, then you'll get a group for each item in the tag list.


sivel:
Would you mind rebasing your commits?  Your commit messages leave a bit to be desired.  They aren't very descriptive and are repetitive.

contrib/inventory/ec2.ini
```diff
@@ -112,6 +112,7 @@ group_by_key_pair = True
 group_by_vpc_id = True
 group_by_security_group = True
 group_by_tag_keys = True
+group_by_tag_values = True
```
bcoca: i would default to false as most current users won't expect a bunch of new groups

jurgenhaas:
@sivel the commit message is rewritten, I hope that's what you were looking for.

@bcoca your suggestion is good and I have modified the PR accordingly.

jurgenhaas:
I'm closing this as it is 15 months old and I wasn't in a position to wait that long so the requirement got solved differently.

PR #13239, opened 2015-11-20
We should give pipes.quote() a string every time

sysreq0:
This fixes an issue in `_execute_module` where we end up passing boolean values (i.e. `module_args['_ansible_check_mode']`, `module_args['_ansible_no_log']`, and `module_args['_ansible_debug']`) to `pipes.quote()` which expects a string.

This happens when executing an "old style" module and enabling either the "check_mode", "no_log", or "debug" options.


jimi-c:
Merged, thanks!

Issue #13240, opened 2015-11-20
Ansible 2.0 does not allow special chars in SUDO password


ThomasSteinbach:
I have a Playbook which is configured with `become: true` started it with the option `--ask-become-pass`. When it asks for the SUDO password, I entered a password with many special chars, like _k§=aI$nt_

With this (example) password I get the following error:

```
fatal: [serverxxxxx]: FAILED! => {"failed": true, "msg": "ERROR! the field 'become_pass' has an invalid value (k§=aI$nt), and could not be converted to an string. Error was: 'ascii' codec can't decode byte 0xc2 in position 1: ordinal not in range(128)"}
```

In Ansible 1.9.4 I had no problems with `sudo: yes` and `--ask-sudo-password`.

chrrrles:
Proposing this as a fix -- https://github.com/ansible/ansible/pull/13257/files
@bcoca @jimi-c @abadger  comments?

ThomasSteinbach:
I can confirm this pull request would solve the issue for me.
Also the failing password would not be printed on the command line, like it was before the fix. I see this as a critical security issue, as others could get a look at the 'rejected' password on the screen.
thanx @chrrrles

jimi-c:
@ThomasSteinbach / @chrrrles I'm ok with the patch above, with the one caveat that it should be using `to_unicode()` from `ansible.utils.unicode` instead of the try/except method (which is how we typically used to deal with this kind of issue).

ThomasSteinbach:
@chrrrles  Could we please re-open this issue, as recent changes reverted the error? Even with `to_unicode()` Ansible fails with parsing a password like _k§=aI$nt_. There is just a new error:

```
UnicodeEncodeError: 'ascii' codec can't encode character u'\xa7' in position 1: ordinal not in range(128)
```

The working implementation for me is

```
if attribute.isa == 'string':
  try:
    value = to_unicode(value)
  except UnicodeDecodeError:
    value = binary_type(value)
```

However @jimi-c , you said that the try/except way is not the best. Did you know a working alternative? What does it mean to use `binary_type` for every string?

abadger:
@ThomasSteinbach Could you give me the full traceback?

The code you're showing me as fixing this is definitely incorrect as it translates the value into either a text type (unicode) or a bytes type (str) depending on whether a UnicodeDecodeError (which should never happen with the default params to to_unicode()).

abadger:
@ThomasSteinbach reproduced... looking further.

[...]

PR #13257, opened 2015-11-23
fixing errors with utf-8 values

chrrrles:
A fix for https://github.com/ansible/ansible/issues/13240


bcoca:
lgtm

jimi-c:
As noted in chat and on the original issue, this should be using `to_unicode()` rather than the try/except method.

abadger:
Note -- we do not want to try turning the value into unicode twice.  Fixed that problem here: 1b74343

Issue #12910, opened 2015-10-26
Winrm connections are not closed, resulting in multiple winrshost.exe and conhost.exe processes on target node


Robby123:
I am running a playbook which runs many tasks on remote windows nodes, however after each playbook completes I notice that there are still many 'winrshost.exe' and 'conhost.exe' processes running on the remote nodes. This causes winrm to crash after a while as it runs out of memory

Robby123:
seems like it only happens when I use the 'raw' module

jhawkesworth:
Which ansible version are you running please?

Robby123:
2.0.0 0.0.pre

jhawkesworth:
please can you try latest devel?  I recall someone reporting this a little while ago but I believe has been addressed fairly recently.

trondhindenes:
@jhawkesworth correct, we saw this for a while, but should be resolved in devel.

jhawkesworth:
@Robby123 - any update on this?

If it is still failing in latest devel then please can you share what you are attempting using 'raw' so this can be investigated further.

gimoh:
I'm using 2.0.0-0.6.rc1 and I'm seeing this not only with raw module, but with win_chocolatey.  The playbook is just:

```
---
- hosts: default
  connection: winrm
  tasks:
    - name: install packages
      win_chocolatey:
        name: "{{ item }}"
      with_items:
        - 7zip.install
        - cmder
        - notepadplusplus.install
        - processhacker.install
```

After running it, there are multiple 'winrshost.exe' and 'conhost.exe' processes hanging about.

cchurch:
Appears to be the same as #9980.

cchurch:
@Robby123 This issue should be fixed by #13261.

PR #13261, opened 2015-11-23
Modify task executor to reuse connections inside a loop.

### LINKED ISSUE #9980: Sporadic WinRMTransportError: 500 WinRMTransport
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.8.2
##### Environment:

Ansible runner: Ubuntu 14.04
Managed machine: Windows 2008 R2 SP1
##### Summary:

From time to time we get the following Ansible error:

```
File "/usr/local/lib/python2.7/dist-packages/ansible/runner/connection_plugins/winrm.py", line 149, in exec_command
    result = self._winrm_exec(cmd_parts[0], cmd_parts[1:], from_exec=True)
   File "/usr/local/lib/python2.7/dist-packages/ansible/runner/connection_plugins/winrm.py", line 114, in _winrm_exec
     self.shell_id = self.protocol.open_shell()
   File "/usr/local/lib/python2.7/dist-packages/winrm/protocol.py", line 103, in open_shell
     rs = self.send_message(xmltodict.unparse(rq))
   File "/usr/local/lib/python2.7/dist-packages/winrm/protocol.py", line 173, in send_message
     return self.transport.send_message(message)
   File "/usr/local/lib/python2.7/dist-packages/winrm/transport.py", line 101, in send_message
     raise WinRMTransportError('http', error_message)
WinRMTransportError: 500 WinRMTransport. Bad HTTP response returned from server. Code 500
```

It does not seem that this is caused by a timeout because the previous task finished just a few seconds before the failing one (for example, 7 seconds in the case I am looking at the moment).

We use latest version of pywinrm.
##### Steps To Reproduce:

Unfortunately, none. It is a sporadic issue that we see regularly... Let's say once per day.
##### Expected Results:

The task to execute successfully.
##### Actual Results:

The described error.

cchurch:
Also fixed WinRM connection to set `_connected` attribute properly and display when remote shell is opened or closed.  Add WinRM integration test using raw + with_items.

Fixes #9980 and #12910.

### DIFF
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -67,6 +67,7 @@ class TaskExecutor:
         self._new_stdin         = new_stdin
         self._loader            = loader
         self._shared_loader_obj = shared_loader_obj
+        self._connection        = None
 
     def run(self):
         '''
@@ -361,8 +362,9 @@ class TaskExecutor:
                 self._task.args = variable_params
 
         # get the connection and the handler for this execution
-        self._connection = self._get_connection(variables=variables, templar=templar)
-        self._connection.set_host_overrides(host=self._host)
+        if not self._connection or not getattr(self._connection, '_connected', False):
+            self._connection = self._get_connection(variables=variables, templar=templar)
+            self._connection.set_host_overrides(host=self._host)
 
         self._handler = self._get_action_handler(connection=self._connection, templar=templar)
 
--- a/lib/ansible/plugins/connection/winrm.py
+++ b/lib/ansible/plugins/connection/winrm.py
@@ -169,14 +169,16 @@ class Connection(ConnectionBase):
         rs = protocol.send_message(xmltodict.unparse(rq))
 
     def _winrm_exec(self, command, args=(), from_exec=False, stdin_iterator=None):
+        if not self.protocol:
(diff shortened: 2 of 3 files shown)


amenonsen:
I think (a) it's pretty hideous to access a connection's `_connected` property from the TaskExecutor, (b) it breaks @mgedmin's #13278 because ssh sets `_connected = True` (arguably incorrectly, since it has no long-lived connection, but that didn't matter before this commit).

I think it should be up to the connection infrastructure to decide whether the connection can be reused or not, not for the TaskExecutor to poke into the connection's internal state. But at a minimum, we should set `_connected=False` [here](https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/connection/ssh.py#L63).

cchurch:
@amenonsen When @jimi-c and I were discussing fixes for this issue, his comment to me was:

> we should add an is_connected() method to each connection plugin and use that instead

I think it would _probably_ be ok to remove the `or not getattr(self._connection, '_connected', False)` check entirely in TaskExecutor.

cchurch:
Nevermind, @jimi-c is reading my mind...  65747285a4d9db63ccd8ed9ad120690860e878b9

PR #13279, opened 2015-11-24
changelog: minor formating fix

resmo:


bcoca:
the double colon you removed :: is there to indicate a code block

resmo:
it is markdown, not reStructured

bcoca:
duh! i spend way too much time in docsite/

PR #13281, opened 2015-11-24
Refactored Consul inventory script

Ari-E-S:
Behaviour remains the same.

Refactored inventory script:
- Decoupled ConsulInventory from ConsulConfig.
- Changed ConsulInventory's init so it would return the data instead of printing it.
- Parameterized the INI file's name.
- Moved "main" into a block.

These changes enable using the inventory as a python module, since I'm trying to link ansible's inventory (sourced from Consul) into Rundeck. Also allows changing the INI file.

### DIFF
--- a/contrib/inventory/consul_io.py
+++ b/contrib/inventory/consul_io.py
@@ -146,7 +146,7 @@ from six import iteritems
 
 class ConsulInventory(object):
 
-  def __init__(self):
+  def __init__(self, config):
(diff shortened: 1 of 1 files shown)


trumant:
LGTM

calfonso:
Hi!

This is a request to backport instead of devel and the refactoring doesn't buy us anything functional.

PR #13297, opened 2015-11-25
Explicitly accept become_success in awaiting_prompt state

### LINKED ISSUE #13289: become with password less sudo
Issue Type: Bug Report (2.0 regression)
Ansible Version: 2.0.0 (rc1)
Ansible Configuration: default
Environment: Ubuntu 14.04
Summary:
If become (sudo) password is provided and host is configured to allow password less sudo ansible times out.

Steps To Reproduce:
Enable password less sudo and try ansible-playbook with become: yes

Expected Results:
Playbook execution

Actual Results:
Fails in first step with timeout waiting for prompt

amenonsen:
If we request escalation with a password, we start in expecting_prompt
state. If the escalation then succeeds without the password, i.e., the
become_success response arrives, we must explicitly move into the next
state (awaiting_escalation, which immediately goes into ready_to_send),
so that we no longer try to apply the timeout.

Otherwise, if the results of the task don't arrive with become_success
(dependent on the exact timing), we leak the success notification and
eventually timeout. But if the module response did arrive before the
timeout expired, the "process has already exited" test would do the
right thing by accident (which is why it didn't fail more often).

Fixes #13289

### DIFF
--- a/lib/ansible/plugins/connection/ssh.py
+++ b/lib/ansible/plugins/connection/ssh.py
@@ -457,12 +457,17 @@ class Connection(ConnectionBase):
                 tmp_stdout = tmp_stderr = ''
 
             # If we see a privilege escalation prompt, we send the password.
+            # (If we're expecting a prompt but the escalation succeeds, we
+            # didn't need the password and can carry on regardless.)
 
-            if states[state] == 'awaiting_prompt' and self._flags['become_prompt']:
-                display.debug('Sending become_pass in response to prompt')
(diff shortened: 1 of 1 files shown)


amenonsen:
I had pushed another commit to fix #13318, but then opened a separate #13344 for that. Please merge this first.

jimi-c:
Merged, thanks!

svanschalkwyk:
This doesn't  seem to work for me. I'm deploying to Google Compute Engine, which has password-less sudo. The script keeps on telling me it has to have a password.

PR #13306, opened 2015-11-25
Potential fix for setup ansible_selinux var incorrect on Fedora 23

goozbach:
Not 100% certain that this is the verbage we want, but it makes more sense than "False".

Fixes http://github.com/ansible/ansible-modules-core/issues/2432

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -633,7 +633,7 @@ class Facts(object):
 
     def get_selinux_facts(self):
         if not HAVE_SELINUX:
-            self.facts['selinux'] = False
+            self.facts['selinux'] = 'Please install libselinux-python module to enable this fact'
             return
         self.facts['selinux'] = {}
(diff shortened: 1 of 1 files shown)


sivel:
What about people who are specifically keying off of the value being `False`?  Changing it to a populated string makes this truthy.

sourcejedi:
Hmm. The copy module seems to knows if selinux is absent (should be False) even without this. Because it only allows to proceed if selinux is absent.

`fatal: [localhost]: FAILED! => {"changed": false, "checksum": "c0d11c949e67c16d49020bc1262646d5de787835", "failed": true, "msg": "Aborting, target uses selinux but python bindings (libselinux-python) aren't installed!"}`

I think this ties into the point @sivel made. We may be running on a system where there's no package at all for `libselinux-python`, because the distribution doesn't support selinux. In that case it would be best to avoid a truthy `ansible_selinux`. Could the code from the `copy` module be re-used here, so `ansible_selinux` is only truthy when "target uses selinux"?

Issue #13221, opened 2015-11-19
Tasks with include_vars take ages :(


zentavr:
**Issue Type:** Bug Report
**Ansible Version:** 2.0.0 devel@d75e707
**Ansible Configuration:**

```
[defaults]
host_key_checking = False
forks=20
```

**Environment:** Mac OS X El Capitan 10.11.1

**Summary**
I have a playbook, which starts like this:

```
- name: "Setting up all {{ game_env }} variables"
  hosts: tag_deployment_ansible
  gather_facts: False
  connection: local
  no_log: True
  tasks:
    - name: Including the defaults
      include_vars: "variables/_defaults.yml"
      tags: [ 'always' ]

    - name: Including AWS Specific Variables
      include_vars: "variables/{{ aws_env }}.yml"
      tags: [ 'always' ]

    - name: Including AWS Specific Secret Variables
      include_vars: "variables/{{ aws_env }}-secrets.yml"
      tags: [ 'always' ]

    - name: Including Game+AWS Specific Variables
      include_vars: "{{ item }}"
      with_first_found:
        - "variables/{{ game_env }}-{{ aws_env }}.yml"
        - "variables/_defaults-{{ game }}-{{ aws_env }}.yml"
      tags: [ 'always' ]

    - name: Including Game+AWS Specific Secrets
      include_vars: "{{ item }}"
      with_first_found:
        - "variables/{{ game_env }}-{{ aws_env }}-secrets.yml"
        - "variables/_defaults-{{ game }}-{{ aws_env }}-secrets.yml"
      tags: [ 'always' ]

    - name: Including Game Variables
      include_vars: "variables/game-{{ game }}.yml"
      tags: [ 'always' ]
```

`game_env` and `aws_env` come from a command-line (`-e` parameter).
In order to execute this step i need to wait little less than an hour. Ansible 1.9.4 executes that piece of cake in ~30 seconds. I have up to 70 machines in my inventory.

**Expected Results:** ~30 seconds

jimi-c:
@zentavr can you provide any details about how many vars are in these various files?

zentavr:
@jimi-c - do you want me to attach files?

jimi-c:
@zentavr sure, if there's nothing sensitive in them. A gist is also fine.

zentavr:
@jimi-c - there are sensitive values inside, so i need to figure how to the in the best way.

jimi-c:
[...]

PR #13307, opened 2015-11-25
Fix for https://github.com/ansible/ansible/issues/13221

Yannig:
Do not call set_variable_manager each time but only at the end of _process_pending_results.

With the playbook described in https://github.com/ansible/ansible/issues/13221 applied on 20 hosts, Ansible takes 35s vs 4 minutes without this patch.


Yannig:
You can retrieve a playbook example at the following location: https://github.com/Yannig/ansible-issue-low-speed/tree/master/big-vars-files

With last version of PR, the previous playbook is running in about 9s with 20 hosts. Before, was taking about 3 mins and 50 seconds.

jimi-c:
Merged, thanks!

PR #13334, opened 2015-11-27
removes editorial

ksatirli:
I feel that Ansible is above the "my hosted Git community is better than yours" discussion and thus removed the editorial around Bitbucket


bcoca:
agreed, that was mostly the submitter griping about his experience, most hosted services have gone down at one time or another, no need to make the temporary issue a permanent comment.

ksatirli:
Thanks @bcoca!

PR #13337, opened 2015-11-27
Adds __main__.yml to possible mains

nydr:
This is to allow better overview when having many included files in a single role

### DIFF
--- a/lib/ansible/playbook/role/__init__.py
+++ b/lib/ansible/playbook/role/__init__.py
@@ -201,6 +201,7 @@ class Role(Base, Become, Conditional, Taggable):
     def _resolve_main(self, basepath):
         ''' flexibly handle variations in main filenames '''
         possible_mains = (
+            os.path.join(basepath, '__main__.yml'),
             os.path.join(basepath, 'main.yml'),
             os.path.join(basepath, 'main.yaml'),
             os.path.join(basepath, 'main.json'),


hloeffler:
i'm against this, also reason is not clear, make a subdir for your includes

nydr:
I haven't found a good solution using subdir for includes, it leads to awkward file paths when using for example the copy module.

```
# file: ./roles/primary_role/tasks/main.yml
- include: inc/sub_role.yml

# file: ./roles/primary_role/tasks/inc/sub_role.yml
- copy:
    src  : sub_role/myfile.conf
    dest : /tmp/myfile.conf
```

File is looked for in:
./roles/primary_role/tasks/files/sub_role/myfile.conf
./sub_role/myfile.conf
While I want my file to be located more something like:
./roles/primary_role/files/sub_role/myfile.conf

bcoca:
the search path is actually (since it is relative, absolute paths bypass this):

```
/pathtorole/
    tasks/sub_role/myfile.conf
    files/sub_role/myfile.conf
    sub_role/myfile.conf

/pathtoplay/
   sub_role/myfile.conf
   files/sub_role/myfile.conf
```

bcoca:
closing this as the **main**.yml really does not add anything to this issue but piles on another option we don't want to check for.

swistakm:
Today I was looking for something like this. In my opinion it would be nice to have such feature because it would simply improve developer experience while writing more complex roles.

This naming conventions is similar to `__init__.py` and `__main__.py`. In most IDEs and code editors such files will be displayed as firs in the directory. This makes navigation in the codebase simpler and quicker. Also Ansible is tool very popular in Python community and is written in Python so the origin of that conventions would be clear for many users.

Alsom when I was introduced to Ansible's Best Practices and the directory layout of role-based playbooks for the first time it wasn't clear to me at all that the `main.yml` file is that important. Someone that got used to the declarative tools like terraform or puppet could think that maybe Ansible is able to pick all of YAMLs in given role subdirectory (i.e. `rolename/tasks/*.yml`). But this is maybe a problem with documentation page. Anyway example of role layout with `__main__.yml` and other "not dunderscored" file in my opinion would more reliably suggest that the `__main__.yml` is special and serves as an "entry point" to the whole directory.

I know that what @hloeffler suggested is a solution to the problem that does not require any changes upstream. Still, it results in bloated role hierarchy and redundant include statements. It also seems like something that reduces code maintainability.

In the end, we split large files into smaller ones for humans not for programs. This helps us maintain, navigate and understand the code. The `__main__.yml` will improve playbook navigation if someone decides to use it. This small change has a strong potential to improve the experience of Ansible users, albeit in small scope and potentially unnoticeable for most of them. Simply a "little big detail".

Yes, this to list of available options, but would require only few keystrokes inside of the `_resolve_main` method and in return it saves developers time and do not require them to use bloated file hierarchy and redundant includes.

PR #13339, opened 2015-11-27
draft of common retry and rate limiting decorators for api based modules

bcoca:
this should allow for a common way to add retry and rate limiting to new and existing API calls

### DIFF
--- a/lib/ansible/module_utils/api.py
+++ b/lib/ansible/module_utils/api.py
@@ -1,0 +1,103 @@
+#
+# (c) 2015 Brian Ccoa, <bcoca@ansible.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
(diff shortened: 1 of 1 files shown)


lib/ansible/module_utils/api.py
```diff
@@ -0,0 +1,103 @@
...
+    """Retry decorator"""
+    def wrapper(f):
+        retry_count = 0
+        def retried(*args,**kargs):
+            if retries is not None:
+                ret = None
+                while True:
+                    retry_count += 1
+                    if retry_count >= retries:
+                        raise Exception("Retry limimit exceeded: %d" % retries)
```
jjshoe: s/limimit/limit/

lib/ansible/module_utils/api.py
```diff
@@ -0,0 +1,103 @@
...
+    arg_spec = (dict(
+        rate=dict(type='int'),
+        rate_limit=dict(type='int'),
+    ))
+    if spec:
+        arg_spec.update(spec)
+    return arg_spec
+
+def retry_argument_spec(spec=None):
+    """Creates an argument spec for working with rate limiting"""
```
jjshoe: "for working with retry limiting"

lib/ansible/module_utils/api.py
```diff
@@ -0,0 +1,103 @@
...
+
+In order to use this module, include it as part of a custom
+module as shown below.
+
+** Note: The order of the import statements does matter. **
+
+from ansible.module_utils.basic import *
+from ansible.module_utils.api import *
+
+The eapi module provides the following common argument specs:
```
willthames: typo: api

bw-jm:
@bcoca I have a few questions, I am running into tons of throttling issues with the AWS API as well eventual consistently problems.

- What is the state of building retries into all of the AWS modules?
- What do you think of doing this through decorators, something like this https://wiki.python.org/moin/PythonDecoratorLibrary#Retry ?

bcoca:
- that is what they are here for
- They ARE decorators

zentavr:
@bcoca - I wonder when this rate_limit feature will be available for AWS modules?

Issue #13269, opened 2015-11-24
Underscore in shell var set bombs in 2.0


nitzmahone:
Issue Type: Bug Report (2.0 regression)
Ansible Version: 2.0.0 (rc1)
Ansible Configuration: default
Environment: (breaks on CentOS 6 and OSX)
Summary:

If preceding a shell task with envvar sets, underscores in the envvar names cause ansible-playbook to bomb with the less-than-helpful error "ERROR! invalid parameter specified: '_A'" (with no contextual information as to what was happening or which file the error is in).

Steps To Reproduce: 
Run the following playbook:

```
- hosts: localhost
  tasks:
  - shell: _A=blar; echo $_A
```

Expected Results:

stdout == "blar"

Actual Results:

ERROR! invalid parameter specified: '_A' (works as expected in 1.9)

Can be worked around by using environment block to set envvars.

Filed on behalf of Callfire

Yannig:
I think, the problem lies in parsing/splitter.py. With this patch, the problem disappear:

``` patch
diff --git a/lib/ansible/parsing/splitter.py b/lib/ansible/parsing/splitter.py
index 3bbf6e7..2180c5f 100644
--- a/lib/ansible/parsing/splitter.py
+++ b/lib/ansible/parsing/splitter.py
@@ -85,7 +85,7 @@ def parse_kv(args, check_raw=False):

                 # only internal variables can start with an underscore, so
                 # we don't allow users to set them directy in arguments
-                if k.startswith('_'):
+                if k.startswith('_') and not check_raw:
                     raise AnsibleError("invalid parameter specified: '%s'" % k)

                 # FIXME: make the retrieval of this list of shell/command
```

@jimi-c what do you think about?

ThomasSteinbach:
Hi @jimi-c , what was the design desicion that 'only internal variables can start with an underscore'? You made that change in commit c83a8337406cc5149da268620a4a85be65061950 about a year ago.

I used that notation for my role internal variables, whenever I create a variable with `register` or `set_fact`. So it would be a boring work to replace that '_notation' all over my ansible roles. I just want to know what is the intention behind that code change, before I start with replacements.

jimi-c:
Yes, we didn't want internal things settable. However for a command/shell/script, it should be setting check_raw to True and bypassing that failure. That might be a problem for set_fact, however I want to make sure this doesn't accidentally allow setting internal things.

Yannig:
[...]

PR #13342, opened 2015-11-28
Devel fix bomb shell

Yannig:
Fix for https://github.com/ansible/ansible/issues/13269

Consistent behaviour between old and new style notation and allow variables starting with `_`. Now only variables starting with `_ansible_` are not allowed at all.


jimi-c:
Merged, thanks!

PR #12276, opened 2015-09-08
Improve ssh.py organisation and state machine

amenonsen:
I had to reorganise the internals of `ssh.py` to make the connection handling clearer. I've tested this with various combinations of `ANSIBLE_PIPELINING=y/n`, `ANSIBLE_SCP_IF_SSH=y/n`, `become: y/n`, `become_method: su/sudo`, etc. but more testing would be appreciated.

PR #13344, opened 2015-11-28
Use connection locking to protect against competing host key prompts

### LINKED ISSUE #13318: SSH host key prompts trample over each other
This is a known problem, for which I had submitted a patch with #12276. Although that PR was merged, the actual locking part was rejected because it (unavoidably) waited too long if ssh timed out trying to connect to unreachable hosts.

@bcoca said on IRC that he may be prepared to reconsider this for v2, because otherwise it's painful to deal with any number of newly-provisioned hosts.

I'm creating this just to keep track of the decision. If the feature is definitely wanted for v2.0, then I'll go back and rebase [https://gist.github.com/amenonsen/a2ecbea86b468e780608](the patch that I had originally submitted) (which worked fine, but by now needs rebasing, and can also be slightly improved upon).

amenonsen:
We acquire the connection lock before executing ssh, and release it as
soon as the unknown host key prompt is negotiated, or we can be sure it
won't be issued at all. This fixes the problem where many prompts pile
up and compete for input.

Unfortunately, we can only lock against other connections: there's no
output_lockfile that can prevent output from other subsystems. See the
FIXME note.

This problem was the original motivation for PR #12276, but although
the bulk of that PR was merged, the locking changes were not.

Submitted for v2 as requested in #13318

### DIFF
--- a/lib/ansible/plugins/connection/ssh.py
+++ b/lib/ansible/plugins/connection/ssh.py
@@ -23,6 +23,7 @@ import fcntl
 import os
 import pipes
 import pty
+import re
 import select
 import shlex
 import subprocess
@@ -138,6 +139,8 @@ class Connection(ConnectionBase):
             self._command += ['sshpass', '-d{0}'.format(self.sshpass_pipe[0])]
 
         self._command += [binary]
+        self._command += ['-C']
+        self._command += ['-vvv']
 
         ## Next, additional arguments based on the configuration.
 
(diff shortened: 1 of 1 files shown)


amenonsen:
Please merge #13297 first.

amenonsen:
I really think Ansible should have a "display" lock too, even if it's used _only_ for this feature. When running against 16- or 32- or 64-node clusters, you still have prompts being overrun by other output (i.e., not other prompts, but just the task output from hosts that you approved before). @jimi-c said something about it being possible to set up a shared lock in display before forking, but @bcoca had some objection/concern with doing it that way, which I unfortunately can't remember now.

bcoca:
A couple of issues, iirc. The first had to do with serialization of the locks, they did not work well inside the object so I moved it into the class.

The second has to do with certain things failing when using the class level lock, I noticed this when i initially set this up for the debug printing, I have not had time to trace this (since it only affectes debug) but it seems to be related to multiprocess, stdout cloning and the lock 'global'.

amenonsen:
The conflict noted by Github is because stable-2.0 doesn't include the pipelining change from #13200, while devel does. Since this was under consideration for inclusion in stable-2.0, I'm leaving this alone. Anyway, the conflict is trivial to resolve by hand.

amenonsen:
I'm closing this PR because @jimi-c has an alternative implementation at the strategy level that avoids all the locking problems.

PR #13345, opened 2015-11-28
Add an 'extract' filter

amenonsen:
At its most basic, this is nothing more than an array or hash lookup,
but when used in conjunction with map, it is very useful. For example,
while constructing an "ssh-keyscan …" command to update known_hosts on
all hosts in a group, one can get a list of IP addresses with:

```
groups['x']|map('extract', hostvars, 'ec2_ip_address')|list
```

This returns hostvars[a].ec2_ip_address, hostvars[b].ec2_ip_address, and
so on. You can even specify an array of keys for a recursive lookup, and
mix string and integer keys depending on what you're looking up:

```
['localhost']|map('extract', hostvars, ['vars','group_names',0])|first
    == hostvars['localhost']['vars']['group_names'][0]
        == 'ungrouped'
```

Includes documentation and tests.

### DIFF
--- a/docsite/rst/playbooks_filters.rst
+++ b/docsite/rst/playbooks_filters.rst
@@ -352,6 +352,39 @@ override those in `b`, and so on.
 This behaviour does not depend on the value of the `hash_behaviour`
 setting in `ansible.cfg`.
 
+.. _extract_filter:
+
+Extracting values from containers
+---------------------------------
+
+.. versionadded:: 2.0
+
+The `extract` filter is used to map from a list of indices to a list of
+values from a container (hash or array)::
+
+    {{ [0,2]|map('extract', ['x','y','z'])|list }}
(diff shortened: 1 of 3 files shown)


amenonsen:
This is a generalisation of the `dictmap` filter proposed in #10082, which I had objected to on the grounds that it partly reimplemented `map` in a very limited way. This more general version, which works in conjunction with `map`, is extremely useful when operating on groups of hosts (as shown in the examples).

Thanks to @halberom for a useful discussion that led to this solution. I believe he (and others, I'm sure) uses a custom filter plugin that does essentially the same thing. It would be very nice to get this merged upstream.

cc: @halberom @jamespharaoh

halberom:
+1, Like it!

It's somewhat similar to the `equalto` jinja2.8 test filter.  Unnecessary by itself, but brilliant when combined with `map`.

PR #13348, opened 2015-11-28
Put in trap for args being None

emonty:
_normalize_old_style_args can return None. If it does, the loop
"for args in args" blows up.


jimi-c:
Merged, thanks!

PR #13360, opened 2015-11-30
Make main include files format more permissive.

fossilet:
Ansible assumes that `roles/x/tasks/main.yml` is a plain list of tasks, but when the file is in the format:

```
tasks:
  - xxx
...
```

it fails with unexpected exception (AssertionError).

This patch tolerates both format. The other way would be to raise an Ansible-specifc exception for this unexpected include file format.


bcoca:
# Question About This One

Hi!

- This adds 'ignorable syntax' which is not a pattern we want to allow.

abadger:
Note, your alternative suggestion to return a more informative error
message here is a good one. If you'd like to work up a pull request with
that implemented Wed be happy to review that.

PR #13373, opened 2015-12-01
BOOLEAN should contain boolean literals

no2a:
It is natural that an argument_spec with choises=BOOLEAN accepts
boolean literal (True, False) though the current implementation
allows only string or int.


abadger:
Cool.  I don't think that True and False will help any current code but it shouldn't hurt any of it either and it does seem right from a completeness standpoint.  merged to devel.  Since it's not fixing any bugs, I'm not merging to stable-2.0 at this time.

PR #13377, opened 2015-12-01
Update intro_windows.rst

mingqiancoding:
The "ansible_ssh_pass/ansible_ssh_password" confusion costed me a little bit time/effort. So, I hope it can clearly described. 

first pull. thanks. :-)

PR #13384, opened 2015-12-01
Update ansible_ssh_changes_note.rst

mingqiancoding:
The "ansible_ssh_pass/ansible_ssh_password" confusion costed me a little bit time/effort. So, I hope it can clearly described.

### DIFF
--- a/docsite/rst/ansible_ssh_changes_note.rst
+++ b/docsite/rst/ansible_ssh_changes_note.rst
@@ -1,3 +1,4 @@
 .. note::
 
     Ansible 2.0 has deprecated the “ssh” from ``ansible_ssh_user``, ``ansible_ssh_host``, and ``ansible_ssh_port`` to become ``ansible_user``, ``ansible_host``, and ``ansible_port``. If you are using a version of Ansible prior to 2.0,  you should continue using the older style variables (``ansible_ssh_*``). These shorter variables are ignored, without warning, in older versions of Ansible. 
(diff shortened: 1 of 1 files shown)


sivel:
I'm not sure specifically what this attempts to address.  I don't think `ansible_ssh_password` is referenced anywhere.

Also we could do better with grammar here and not use "BTW" in our docs.

mingqiancoding:
Thanks for reply.
It's related to the ansible windows doc. 
https://github.com/ansible/ansible/pull/13377

docsite/rst/ansible_ssh_changes_note.rst
```diff
@@ -1,3 +1,4 @@
 .. note::
 
     Ansible 2.0 has deprecated the “ssh” from ``ansible_ssh_user``, ``ansible_ssh_host``, and ``ansible_ssh_port`` to become ``ansible_user``, ``ansible_host``, and ``ansible_port``. If you are using a version of Ansible prior to 2.0,  you should continue using the older style variables (``ansible_ssh_*``). These shorter variables are ignored, without warning, in older versions of Ansible. 
+    Please pay attention to the older style variables (``ansible_ssh_*``): ansible_ssh_password doesn't exist, should be ansible_ssh_pass.
```
docschick, line 4: I'd like to suggest a bit of editing here... maybe this instead:

For older variable styles, please also note that `ansible_ssh_password` does not exist and should be entered as `ansible_ssh_pass` instead.

PR #13405, opened 2015-12-02
Get v2_playbook_on_start working

sivel:
This PR does the following:
- Move self._tqm.load_callbacks() earlier to ensure that v2_on_playbook_start can fire
- Pass the playbook instance to v2_on_playbook_start
- Add a _file_name instance attribute to the playbook

As it stands, `v2_on_playbook_start` is never run.  This change ensures that it will, and also changes it's signature to pass in the playbook instance.

Additionally, due to how I plan on using this, it would be nice to also have the filename associated with the playbook instance available to the callback.


sivel:
This should also be cherry picked into stable-2.0 after merge to devel.

jimi-c:
Merged to devel and stable-2.0, thanks!

PR #12889, opened 2015-10-23
draft of slack callback module

bcoca:
redo of #8955 with view to v2 ... cannot get arguments passed still not ready for inclusion

PR #13408, opened 2015-12-02
V2 slack callback plugin

sivel:
This PR supersedes https://github.com/ansible/ansible/pull/8955 and https://github.com/ansible/ansible/pull/12889 

Of note here, because we don't have the ability to look into the playbook and play objects like before, I'm using `CLI.base_parser` to get some of that data.  I'm not super happy with that solution, although it does get me that data.  I'm open to suggestions.  However, this is working well for us as is.

cc @bcoca

### DIFF
--- a/lib/ansible/plugins/callback/slack.py
+++ b/lib/ansible/plugins/callback/slack.py
@@ -1,0 +1,230 @@
+# (C) 2014-2015, Matt Martz <matt@sivel.net>
+
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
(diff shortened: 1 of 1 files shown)


bcoca:
so I understand why you want the CLI stuff, but I think that does not belong inside this plugin, we might want to make it part of the base or just pass that info to callbacks in general.

sivel:
I think we are in agreement there.  I'm just not sure of what the best approach is.

bcoca:
so here i show a simpler method of getting the same info https://gist.github.com/bcoca/5c70c475b132113d82b1, now all callbacks have access to the `options` object.

bcoca:
https://github.com/ansible/ansible/pull/13412 <= now added to the 'defaults'

sivel:
I've updated this PR to use `self._options` which due to #13412 not being merged yet, is handled in this callback itself.

@bcoca let me know what you think so far.

EpiqSty:
Hello!

I have tried to use this callback module, but got :

` [WARNING]: Could not submit message to Slack: HTTP Error 404: Not Found`

error

Can you please advise, what else can be wrong here?

```
$ source .slack 
$ more .slack
SLACK_WEBHOOK_URL=https://hooks.slack.com/services/aaa/bbb/ccc
export SLACK_WEBHOOK_URL
$ grep slack ansible.cfg 
callback_whitelist = slack
```

Thanks!

EpiqSty:
I've got the reason of 404 error. Seem's that 
+        SLACK_CHANNEL     (optional): Slack room to post in. Default: #ansible
is not optional - as we have not special #ansible channel in our slack , we use another channel for testing deploys notification.
So callback start sending messages to slack only after specifying another channel via env.
It seem's strange, because  SLACK_WEBHOOK_URL at fact issued to some default channel, but if it's not #ansible - you will get such 404 error.

PR #13414, opened 2015-12-03
Don't override stdout_callback in adhoc

emonty:
The fallback condition for --oneline totally overrides the
stdout_callback setting from ansible.cfg.

### DIFF
--- a/lib/ansible/cli/adhoc.py
+++ b/lib/ansible/cli/adhoc.py
@@ -161,7 +161,7 @@ class AdHocCLI(CLI):
         if self.options.one_line:
             cb = 'oneline'
         else:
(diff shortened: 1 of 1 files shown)


bcoca:
that was on purpose for `ansible` adhoc, if you do this it will use the `default` callback to output instead of minimal.

emonty:
@bcoca that means it's not possible to configure a logger to use for adhoc at all then?

bcoca:
you can configure a logger and additional callbacks, you cannot override the stdout one, this was the case in 1.x also.

emonty:
@bcoca okie. I'll try a different approach. thanks!

Issue #13388, opened 2015-12-02
set user for docker connection


ThomasSteinbach:
I use the Docker connection plugin, introduced in Ansible 2.0, which executes Tasks with `docker exec` inside a Docker container. The command `docker exec` allows the `-u` or `--user` option, to set the user you want to execute the task within the container. It would be nice if Ansibles `remote_user` (or `become_user`?) would be applied to that option.

I already tried my luck to implement that functionality by adding the option flag and value to the Ansible connectors command list like this:

```
local_cmd = [self.docker_cmd, "exec", '-u', self._play_context.remote_user, '-i', self._play_context.remote_addr, executable, '-c', cmd]
```

However, the first task executed hangs. There is no error message, the python scrint is in process inside the container, but nothing happens. Here is the Docker connectors source file:

https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/connection/docker.py

And here are three example tasks hanging, when adding `'-u', self._play_context.remote_user,` to both exec commands in the source file:

```
- name: create jenkins container
  docker:
    name: aci_jenkins
    image: jenkins
    state: running

- name: add aci_jenkins to inventory
  add_host:
    name: aci_jenkins
    ansible_connection: docker
    ansible_user: root

- name: install git
  delegate_to: aci_jenkins
  apt:
    name: git
    state: present
```

You can execute those task on your localhost, if you have Docker installed. Any idea, why the third task is hanging?

abadger:
@ThomasSteinbach Was the hanging fixed when we reverted #13200 (at this commit: https://github.com/ansible/ansible/commit/e201a255d17a72b338be92b8db881effb79b5ece )?

ThomasSteinbach:
Yes it was fixed. The tasks via the docker connector does respond quickly as before.

ThomasSteinbach:
This issue could be closed as fixed as the corresponding PR was merged.

bcoca:
@ThomasSteinbach you as opener can close it also, but doing so anyways

ThomasSteinbach:
Oh sorry. That sounds reasonable but I haven't mentioned this. Thank you.

PR #13424, opened 2015-12-04
use remote_user or become_user in docker connection

ThomasSteinbach:
This commit makes Ansible using the `remote user` (aka `ansible_user`) when executing tasks against a Docker container with the Docker connection plugin. Alternatively the `become_user` is used when `become` is true.

This PR would solve issue #13388

**Side effect:** The `docker exec` way does not need a `become_method`, because Docker has its own user management. Thus the user is set by Docker when starting the process with `docker exec`. Using an additional become method then is not just unnescessary but also a disadvantage, because the Docker image needs to provide the become method. I don't know how to disable the `become_method` for the Docker connection plugin. Anybody with enough knowledge could to that after merging this PR, because then the `become_method` becomes obsolete like described.

### DIFF
--- a/lib/ansible/plugins/connection/docker.py
+++ b/lib/ansible/plugins/connection/docker.py
@@ -122,8 +122,9 @@ class Connection(ConnectionBase):
         super(Connection, self).exec_command(cmd, in_data=in_data, sudoable=sudoable)
 
         executable = C.DEFAULT_EXECUTABLE.split()[0] if C.DEFAULT_EXECUTABLE else '/bin/sh'
+        exec_user = self._play_context.become_user if self._play_context.become else self._play_context.remote_user
         # -i is needed to keep stdin open which allows pipelining to work
-        local_cmd = [self.docker_cmd, "exec", '-i', self._play_context.remote_addr, executable, '-c', cmd]
+        local_cmd = [self.docker_cmd, "exec", '-u', exec_user, '-i', self._play_context.remote_addr, executable, '-c', cmd]
 
         display.vvv("EXEC %s" % (local_cmd), host=self._play_context.remote_addr)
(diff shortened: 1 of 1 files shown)


abadger:
Ansible should present a consistent view to the end user so probably having the connection plugin use the remote_user and leaving become\* alone is the correct thing to do.  That way sites that want to be consistent in their use of sudo or su in their playbooks can implement that and sites that don't need that abstraction for their docker containers can just set remote_user on the task level.

ThomasSteinbach:
I fully agree with you, that the `become_method` should behave the same everywhere. That is reasonable.

Unfortunately the `remote_user` does not operate on task level ( when using `delegate_to` ?). Please have a look at Issue #13388. In the third task of the Playbook the `remote_user` would not applied to the connector, if this variable would be set there.

What if we add a new `become_method` like _docker_, which does not alter the command to execute but makes the docker connector use the `become_user` instead of the `remote_user`, as my PR currently does? And if the `become_method` is not _docker_, the docker connector uses the `remote_user` in any case.

Would that be a good approach?

ThomasSteinbach:
Hi @abadger, I changed the code of the PR according your comment, just using the `remote_user` for the docker connection and leaving the `become_user` alone. This would also simplifying the code, as just  docker run `-u self._play_context.remote_user` option would be added.

Besides I opened PR #13615 which will fix the task variables precedence (issue #13614) and make it possible to change the `become_user` for the docker connection per task level.

abadger:
Seems we got side tracked on this one with the unrelated hangs and such.  I'll rebase and merge this now.

abadger:
Rebased and added to devel as: 4ac49ed4a8535c4b4e1aab75bc0d7787be5907dd  and: 14dfad730e6356256950efbafd03e57f680499c7

thanks @ThomasSteinbach !

ThomasSteinbach:
:heart:

PR #13433, opened 2015-12-04
Add inventory script for LXC containers

fghaas:
Enables an LXC server's configuration as an inventory source for LXC containers.

In LXC, containers can be defined with an `lxc.group` configuration option that is normally used with `lxc-autostart -g`. Here, we are using the same option to build Ansible inventory groups.

In addition to being grouped according to their `lxc.group` entry (or entries, as LXC allows a single  container to be in multiple groups), we also add all containers (including those with no `lxc.group` entry) to the `all` group.

Related PR for lxc connection plugin: #13145.

### DIFF
--- a/contrib/inventory/lxc_inventory.py
+++ b/contrib/inventory/lxc_inventory.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python
+#
+# (c) 2015-16 Florian Haas, hastexo Professional Services GmbH
+# <florian@hastexo.com>
+# Based in part on:
+# libvirt_lxc.py, (c) 2013, Michael Scherer <misc@zarb.org>
+#
+# This file is part of Ansible,
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+"""
+Ansible inventory script for LXC containers. Requires Python
+bindings for LXC API.
+
+In LXC, containers can be grouped by setting the lxc.group option,
+which may be found more than once in a container's
+configuration. So, we enumerate all containers, fetch their list
+of groups, and then build the dictionary in the way Ansible expects
+it.
+"""
+from __future__ import print_function
+
+import sys
+import lxc
+import json
+
+def build_dict():
+    """Returns a dictionary keyed to the defined LXC groups. All
+    containers, including the ones not in any group, are included in the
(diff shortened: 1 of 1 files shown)


abadger:
Looks fine but I suppose we need #13145 to be merged first so that the ansible_connection setting makes sense.

contrib/inventory/lxc.py
```diff
@@ -0,0 +1,79 @@
...
+# Horrible hack: we want this script to be called lxc.py, yet need to
+# import a module named lxc. Absolute import (from __future__ import
+# absolute_import) does not work here, because this is a script, not a
+# package. The only way to tell Python that "import lxc" does *not*
+# mean "import yourself" is to pop the first entry off of sys.path.
+#
+# Since inventory scripts are always invoked directly, and never
+# imported, we should be able to get away with this.
+import sys
+sys.path.pop(0)
```
mattclay: This hack prevents calling as `./lxc.py`:

```
  File "./lxc.py", line 79, in <module>
    main(sys.argv)
  File "./lxc.py", line 70, in main
    result = build_dict()
  File "./lxc.py", line 59, in build_dict
    for c in _lxc.list_containers()])
AttributeError: 'module' object has no attribute 'list_containers'
```

Instead of this, rename the script to `lxc_inventory.py`.

fghaas:
@mattclay Import hack removed, script renamed, branch rebased on current devel.

fghaas:
Thanks!

jjaone:
Hey thanks for this LXC-inventory script, glad to report that I'm using it in git-devel based Ansible installation 2.2.0.x-devel in Xenial/Ubuntu and the script correctly discovers all local LXC:s and places them in groups if lxc.group specifier is been used. 

The 'lxc_inventory.py'  works fine when called outside playbooks but sudo must be used to get it work (probably cause I have privileged containers in /var/lib/lxc and when run as normal user the script does not connect/detect them.. 

For this reason, I had to use a shell-wrapper: "lxc_inventory.sh" that simply calls the .py script with sudo so that I can use this in `ansible-playbook -i lxc_inventory.sh ...` to discover local LXCs:

```
#!/bin/sh
[ "root" != "$USER" ] && exec sudo ./lxc_inventory.py "$@"
```

I'm not sure if this is how it should work, but that was the only way for me to get it working in plays and inventory dirs of my polaybooks.

Additionally, I would like to know is this script supposed to be used with 'lxc-connection' -plugin and if so how? Can these be used to connect and/or discover remote/target host servers that have LXCs running, or is this 'inventory' limited to local LXC-creation and utilization only..?

Cheers, and thanks for ur efforts for making Ansible better support LXC-deployments and development :)
-jukka

fghaas:
@jjaone All that the inventory script uses is the same LXC connection library that the `lxc-*` commands also use. As such, it displays the same behavior as the latter: run `lxc-ls` as a regular user, and you'll see just the unprivileged containers under your account. Run `sudo lxc-ls`, and you'll see just the privileged containers on that system.

I saw no reason to override this in the inventory script. In fact, I think doing so would violate [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

So yes, if you're managing privileged containers with this dynamic inventory, run `sudo ansible-playbook`. If you're managing your unprivileged containers, `ansible-playbook` should suffice.

On your other question, yes it's for local containers.

PR #13437, opened 2015-12-04
comment examples in default hosts file

ghost:
When a user does not specify the inventory file, ansible by default will attempt to resolve and use the hosts specified in the default hosts file.  

The examples, however should not be resolved, or attempted to be communicated with, as that is not desirable or even safe.

Linux distros package this file to /etc/ansible/hosts by default.


abadger:
I've thought the same thing myself recently.  Thanks for taking hte time to change that.

PR #13442, opened 2015-12-05
Allow ini plugin to load file using other encoding than utf8.

Yannig:
This pull request allow user to specify a alternative encoding for ini/java properties file.

### DIFF
--- a/lib/ansible/plugins/lookup/ini.py
+++ b/lib/ansible/plugins/lookup/ini.py
@@ -31,7 +31,7 @@ from ansible.plugins.lookup import LookupBase
 def _parse_params(term):
     '''Safely split parameter term to preserve spaces'''
 
-    keys = ['key', 'type', 'section', 'file', 're', 'default']
+    keys = ['key', 'type', 'section', 'file', 're', 'default', 'encoding']
     params = {}
     for k in keys:
         params[k] = ''
@@ -52,19 +52,6 @@ def _parse_params(term):
 
 class LookupModule(LookupBase):
 
-    def read_properties(self, filename, key, dflt, is_regexp):
-        config = StringIO()
-        current_cfg_file = open(to_bytes(filename, errors='surrogate_or_strict'), 'rb')
-
-        config.write(u'[java_properties]\n' + to_text(current_cfg_file.read(), errors='surrogate_or_strict'))
-        config.seek(0, os.SEEK_SET)
-        self.cp.readfp(config)
-        return self.get_value(key, 'java_properties', dflt, is_regexp)
-
-    def read_ini(self, filename, key, section, dflt, is_regexp):
-        self.cp.readfp(open(to_bytes(filename, errors='surrogate_or_strict')))
-        return self.get_value(key, section, dflt, is_regexp)
-
     def get_value(self, key, section, dflt, is_regexp):
         # Retrieve all values from a section using a regexp
         if is_regexp:
@@ -79,8 +66,6 @@ class LookupModule(LookupBase):
 
     def run(self, terms, variables=None, **kwargs):
 
-        basedir = self.get_basedir(variables)
-        self.basedir = basedir
         self.cp = configparser.ConfigParser()
 
         ret = []
@@ -94,6 +79,7 @@ class LookupModule(LookupBase):
                 'default': None,
                 'section': "global",
                 'type': "ini",
+                'encoding': 'utf-8',
             }
 
             # parameters specified?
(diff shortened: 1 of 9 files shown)


abadger:
Parameter looks good.  For implementation I would lean towards reading it in as a plain byte string and then using to_unicode/to_bytes to convert (from ansible.utils.unicode) instead of using codec.open().  Doing the conversion separately from reading lets us be explicit about what we want to happen in case the user specified encoding doesn't match with what's in the file.  (And then it's easier to change the behaviour in the future).

It looks like tests should go into this role: tests/integration/roles/test_lookups/tasks/main.yml   That would hopefully include moving the current test_lookup_properties tasks into there.  But at least,, new tests should go there (as there's no Makefile rule that's invoking test_lookup_properties, tests that are in there probably aren't getting run at all).

We should add a utf8 value to lookup.ini as well. and test that reading a non-ascii field and saving it for all three of those files work (because they hit different code paths).

Yannig:
@abadger thanks about your suggestions using ansible.utils.unicode, I'll take a look to convert my code.

For the tests, i'll take a look to add more test into test_lookups and add utf-8 strings into lookup.ini.

Yannig:
@abadger Done
Can you tell me what do you think about this commit: 4e35dcc?
I think it can help in other lookup plugins (like file or csv).

abadger:
By itself, 4e35dcc isn't something I wanted to work on enabling for 2.0
(where "something" is user specified encodings for playbooks, vars files,
and other things central to ansible's function).  There's probably still
sections of the code that can't handle utf8 let alone user specified
encodings. However, 2.1 development has begun so perhaps it is time to look
into it. We probably also should make the two changes separately as we
might have to revert or modify playbook loading changes but not want to
destroy this feature at the same time.

In the larger context of this patch, it's probably wrong to have the ini
plugin use dataloader's private method like that.  My inclination would be
if we want a general facility for loading file content it should be in a
utils module.  Not sure how much of dataloader's._get_file_content ()
belongs there and how much is specific to dataloader's needs.  A general
function for an encoding aware file loading function would need to take
many of the same arguments as to_Unicode does ( the errors parameter is
noticeably missing).

@jimi-c what do think of all this?

bcoca:
~~need to x2 check, but this might have already been fixed in #23001 with new ini  plugin~~ off topic

abadger:
@yannig I'd be willing to merge this but the tests are out of date by now.  If you're interested in going back and updating this and then pinging me on irc, I'll go ahead and merge.

test/runner/lib/util.py
```diff
@@ -170,8 +170,8 @@ def raw_command(cmd, capture=False, env=None, data=None, cwd=None, explain=False
         encoding = 'utf-8'
         data_bytes = data.encode(encoding) if data else None
         stdout_bytes, stderr_bytes = process.communicate(data_bytes)
-        stdout_text = stdout_bytes.decode(encoding) if stdout_bytes else u''
-        stderr_text = stderr_bytes.decode(encoding) if stderr_bytes else u''
+        stdout_text = stdout_bytes.decode(encoding, 'replace') if stdout_bytes else u''
+        stderr_text = stderr_bytes.decode(encoding, 'replace') if stderr_bytes else u''
```
abadger: @mattclay what do you think of this?  I think it's okay but I didn't trace through and see how stdout_text and stderr_text are used later to make sure.

mattclay: This is used by `ansible-test` to execute all external commands, so I'm not sure this is a good idea. I believe the only encoding issue is with parsing of `git diff` output, as seen from an older CI run for this PR:

https://app.shippable.com/github/ansible/ansible/runs/29803/1/console

I'm thinking that `replace` should only be used for `git diff` output, since it's not guaranteed to be `utf-8`.

@abadger Does that seem reasonable to you?

abadger: @mattclay @Yannig yep, that seems reasonable to me.

Yannig:
@abadger @mattclay what do you think about this new version. Only diff cmd is set with replace.

lib/ansible/plugins/lookup/ini.py
```diff
@@ -105,11 +91,23 @@ def run(self, terms, variables=None, **kwargs):
...
+            # Special case for java properties
             if paramvals['type'] == "properties":
-                var = self.read_properties(path, key, paramvals['default'], paramvals['re'])
-            else:
-                var = self.read_ini(path, key, paramvals['section'], paramvals['default'], paramvals['re'])
+                config.write(u'[java_properties]\n')
+                paramvals['section'] = 'java_properties'
+
+            # Open file using encoding
+            contents, show_data = self._loader._get_file_contents(path, encoding=paramvals['encoding'])
```
abadger, line 105: Do we want to use loader here or get the file straight from the filesystem?  (I don't know... I'm just bringing that up since the old code got it straight from the filesystem).

@bcoca, do you know what the general strategy for these types of things should be?

bcoca: using loader allows for the ini file to be vaulted

abadger: Cool.  So this seems like the right strategy for accessing files on the controller.

abadger:
The shippable pep8 failures need to be addressed but then this looks good to merge.   @mattclay, anything else you'd like to see changed?

mattclay:
@Yannig Please add docstring type annotations for the new `str_errors` argument. Other than that, the changes to ansible-test look good.

Yannig:
@mattclay done :)

abadger:
Merged to devel for the 2.4.0 release.

mattclay:
@Yannig For future reference, the type annotations in ansible-test should document the types, not the values of arguments. https://github.com/ansible/ansible/pull/27451 corrects the annotations from this PR, along with a few others.

PR #13450, opened 2015-12-06
bugfix for ios.py shared module argument creation

privateip:
This patch fixes a bug in module_utils/ios.py where the the wrong shared
module arguments are being generated.  This bug prevented the shared module
from operating correctly.  This patch should be generally applied.


jimi-c:
cc @privateip

Issue #13453, opened 2015-12-07
debug show VARIABLE IS NOT DEFINED on variable when content is the same as name variable


Yannig:
##### Summary:

Ansible show VARIABLE IS NOT DEFINED on variable when content is the same as name variable
##### Steps To Reproduce:

An easy way to reproduce this behaviour is to launch the following test:

```
ansible -m debug -a var=test -e test=test localhost
```
##### Expected Results:

```
localhost | SUCCESS => {
    "changed": false, 
    "test": "test"
}
```
##### Actual Results:

```
localhost | SUCCESS => {
    "changed": false, 
    "test": "VARIABLE IS NOT DEFINED!"
}
```

Yannig:
The following patch is doing the job:

``` patch
diff --git a/lib/ansible/plugins/action/debug.py b/lib/ansible/plugins/action/debug.py
index a024e28..1d8e28c 100644
--- a/lib/ansible/plugins/action/debug.py
+++ b/lib/ansible/plugins/action/debug.py
@@ -45,8 +45,12 @@ class ActionModule(ActionBase):
                 # If var is a list or dict, use the type as key to display
                 result[to_unicode(type(self._task.args['var']))] = results
             else:
+                # If var name is same as result, try to template it
                 if results == self._task.args['var']:
-                    results = "VARIABLE IS NOT DEFINED!"
+                    try:
+                        results = self._templar.template("{{" + results + "}}", convert_bare=True, fail_on_undefined=True)
+                    except:
+                        results = "VARIABLE IS NOT DEFINED!"
                 result[self._task.args['var']] = results
         else:
             result['msg'] = 'here we are'
```

Here is a simple example:
- Without test variable defined:

``` yaml
ansible -m debug -a var=test localhost
localhost | SUCCESS => {
    "changed": false, 
    "test": "VARIABLE IS NOT DEFINED!"
}
```
- Now with test variable:

``` yaml
ansible -m debug -a var=test -e test=test localhost
localhost | SUCCESS => {
    "changed": false, 
    "test": "test"
}
```

jimi-c:
Merged your patch above, so closing this.

Thanks!

Yannig:
This issue is here again in last devel. I'll publish a new PR.

Yannig:
@jimi-c the new PR is here: https://github.com/ansible/ansible/pull/13629

It contains assertions about debug behaviour but I don't known if it the right place for it. Any advice?

[...]

PR #13455, opened 2015-12-07
Fix issue when var name is the same as content.

Yannig:
See https://github.com/ansible/ansible/issues/13453 for more details.

Here is a simple example:
- Without test variable defined:

``` yaml
ansible -m debug -a var=test localhost
localhost | SUCCESS => {
    "changed": false, 
    "test": "VARIABLE IS NOT DEFINED!"
}
```
- Now with test variable:

``` yaml
ansible -m debug -a var=test -e test=test localhost
localhost | SUCCESS => {
    "changed": false, 
    "test": "test"
}
```
- Same without the patch:

``` yaml
ansible -m debug -a var=test -e test=test localhost
localhost | SUCCESS => {
    "changed": false, 
    "test": "VARIABLE IS NOT DEFINED!"
}
```

### DIFF
--- a/lib/ansible/plugins/action/debug.py
+++ b/lib/ansible/plugins/action/debug.py
@@ -45,8 +45,12 @@ class ActionModule(ActionBase):
                 # If var is a list or dict, use the type as key to display
                 result[to_unicode(type(self._task.args['var']))] = results
             else:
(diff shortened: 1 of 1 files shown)


jimi-c:
Merged, thanks!

PR #13461, opened 2015-12-07
Galaxy 2.0

chouseknecht:
Added new commands:
- login - authenticate user prior to performing CRUD operations (i.e. import and delete)
- setup - configure Travis integration
- import - import a role from a GitHub repo into Galaxy
- delete - delete a role from Galaxy

Added .travis.yml - see template in galaxy/data.

Added 'github_branch' to meta/main.yml. If set, travis notifications will be accepted from this branch, installs will come from this branch (unless the repo has tags). If not set, the repo's default branch will be used.

Reworked search to point at search/roles endpoint, provide --author option, and present more usable results.

Defined galaxy_server and galaxy_ignore_certs in constants.py.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@ Ansible Changes By Release
   by setting the `ANSIBLE_NULL_REPRESENTATION` environment variable.
 * Added `meta: refresh_inventory` to force rereading the inventory in a play.
   This re-executes inventory scripts, but does not force them to ignore any cache they might use.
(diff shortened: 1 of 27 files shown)


jimi-c:
OrderedDict is a python 2.7 feature, and can't be used as we still have to support 2.6. Alternatively you could use namedtuple(), or just sort on keys when printing the available commands out.

jimi-c:
Could you also please rebase this and put it in a feature branch? You're getting commits leaking in from other work done on devel.

Issue #4500, opened 2013-10-14
EC2 module: Using "wait" parameter as variable


bhutch:
Hello,

There may be a bug when using the "wait" parameter as a variable with the EC2 module.

$ ansible --version
ansible 1.4 (devel b8ec300d4c) last updated 2013/10/14 11:43:46 (GMT -500)

playbook "vars" section:
wait: yes

tasks section:
local_action: ec2 keypair={{ keypair }} group_id={{ group_id }} instance_type={{ instance_type }} image={{ image }} region={{ region }} wait={{ wait }} instance_tags='{"Name":"Example"}'

Error:
failed: [localhost] => {"failed": true}
msg: value of wait must be one of: yes,on,1,true,1,no,off,0,false,0, got: True

As a workaround, I've just been using "wait=yes" or "wait=true" when calling the ec2 module.

mpdehaan:
looks like booleanification needs to lowercase when it's really fed a python True

I'll look into it.

PR #13464, opened 2015-12-07
Add uppercase True/False to BOOLEANS

sontags:
Refix issue #4500 analogues to  https://github.com/ansible/ansible/commit/4833c2fdf06352835742cacac12043a1e5f0904d to solve the situation described in https://github.com/ngoyal/ansible-bool-sample


bcoca:
there should be no need for caps as we compare to the lower case of the value ( via `.lower()` ) so I'm going to close this request.

PR #13467, opened 2015-12-08
adhoc does not load plugins by default

bcoca:
reimplemented feature from 1.x which kept additional callbacks from
poluting adhoc unless specifically asked for through configuration.


abadger:
+1

PR #13471, opened 2015-12-08
Issue 13438

davidlballenger:
@abadger 

OS X El Capitan moved the /etc/ssh_\* files into /etc/ssh/*. So in the get_public_ssh_host_keys() method in lib/ansible/module_utisl/facts.py, I have added an additional check for Darwin systems. If the distribution is MacOSX and the version is >= 10.11, key_dir is set to the new location. Older versions use the old location.

I've tested this fix against systems running El Capitan and Yosemite (previous version). Both find the host keys correctly.  Also tested against Debian Jessie.

Thanks - David


jimi-c:
Merged thanks!

davidlballenger:
Thanks

PR #13460, opened 2015-12-07
Adds support for converting a MAC address to an integer

yesbox:
Adds an "int" format to the MAC address filter of ipaddr.py. Useful for doing math with MAC addresses, such as ensuring a series of MAC addresses are in a specific order when configuring VMs.

Example:

``` ---
- hosts: local
  tasks:
    - set_fact: mac_str="52:54:00:00:00:00"
    - set_fact: base_mac="{{ mac_str | macaddr('int') }}"
    - debug: msg="{{ (base_mac|int + 257) | macaddr('linux') }}" # 52:54:00:00:01:01
    - debug: msg="{{ (base_mac|int + 258) | macaddr('linux') }}" # 52:54:00:00:01:02
```

Separate from but useful with "[Adds the int_hash filter](https://github.com/ansible/ansible/pull/13473)" to generate MAC addresses on NICs in an ascending order on a host.

PR #13473, opened 2015-12-08
Adds the int_hash filter.

yesbox:
int_hash hashes a string and returns the hash in integer form, truncated to a range selected by the user.

It uses sha512 but this detail is not important to the user, it should only promise that the output is indistinguishable from random numbers given that the input changes and that the result is always the same given the same input.

This is meant as an alternative to the random integer filter, giving you a number which is for practical purposes random but allows generation of the numbers in an idempotent way. For example, the name of a host could be used as input to the int_hash to produce the same number between playbook executions in places were the random filter could have been used.

The result should not be trusted to be unique any more than randomly selected number in the same range; it's not. Basic experimentation has shown that the amount of calls to needed Python 2.7 random.randint() before a duplicate number is produced is on average the same as when using int_hash and using an input in the form of "<some_static_string>-<sequentially_increasing_number>". Based on this I conclude that for non crypto purposes it's as "good" as random numbers, as should be expected from a cryptographically secure hash such as sha512 which does all the lifting.

One real world use case for this is generating a MAC address from a hostname.
When letting libvirt select its own MAC addresses it does it from the same range the example task below does. By using int_hash with the inventory host name instead of random the task becomes idempotent and doesn't replace the MAC addresses if the playbook in rerun.

Example:

```
TASK [debug msg={{ 'hostname10-32' | int_hash(90520730730496, 90520747507711) | macaddr('linux') }}] ***
ok: [localhost] => {
    "changed": false,
    "msg": "52:54:00:26:b5:b1"
}
```

Separate from but useful with "[Adds support for converting a MAC address to an integer](https://github.com/ansible/ansible/pull/13460)" to generate MAC addresses on NICs in an ascending order on a host.

### DIFF
--- a/docsite/rst/playbooks_filters.rst
+++ b/docsite/rst/playbooks_filters.rst
@@ -309,6 +309,11 @@ Other hashes (platform dependent)::
 
     {{ 'test2'|hash('blowfish') }}
 
+Get a hash in integer form of a string. Takes two arguments where the first
+is the minimum and the second is the maximum possible integer returned.
+
+    {{ 'test3'|int_hash(0, 4294967294) }}
+
 To get a sha512 password hash (random salt)::
 
     {{ 'passwordsaresecret'|password_hash('sha512') }}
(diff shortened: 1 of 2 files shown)


dharmabumstead:
Closing due to inactivity.

PR #13475, opened 2015-12-08
Support for cowsay installed by homebrew

### LINKED ISSUE #13465: homebrew - cowsay
Support for cowsay installed by homebrew

echo "$(brew --prefix)/bin/cowsay"

aluminiumgeek:
Support for cowsay installed by homebrew.

Fixes #13465


sivel:
I'm not a fan of shelling out to brew for this.  Maybe we should just use some PATH magic that we use elsewhere for finding an executable, or add an ENV var for this.

bcoca:
agreed, i would prefer to have a var/config that adds additional paths to cowsay search.

Issue #13411, opened 2015-12-02
Transient failure caused by new pipelining patch.


abadger:
I've been testing a moderate sized playbook for fedora infrastructure.  the new requiretty + pipelining patch is causing a very strange problem.

The playbook has 136 tasks to create and configure a certain host.  A second host gets 123 tasks and skips 13.  When I run the devel branch with pipelining on the host that has 136 tasks always fails.  The one with 123 tasks never has.  However, the task that fails is not stable.  In every run I've made so far it's been a different task.  (All of the tasks that failed thus far have run on both hosts but I think that's just a percentage chance.. I don't think the tasks necessarily have to run on both).

The traceback I get back from the host is either a SyntaxError or  an IndentationError:

```
<noc02.fedoraproject.org> SSH: EXEC ssh -C -vvv -o ControlMaster=auto -o ControlPersist=60s -o Port=22 -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o User=root -o ConnectTimeout=90 -o ControlPath=/root/.ansible/cp/ansible-ssh-%h-%p-%r -tt noc02.fedoraproject.org LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8 LC_MESSAGES=en_US.UTF-8 python -uc 'import sys; [sys.stdout.write(s) for s in iter(sys.stdin.readline, "__EOF__942d747a0772c3284ffb5920e234bd57__\n")]'| LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8 LC_MESSAGES=en_US.UTF-8 /usr/bin/python
[...]

Issue #13410, opened 2015-12-02
Solaris hosts can't be pinged


fboender:
I'm running into an issue with the latest (commit 9bf1aaf7f5e86030; Wed Dec 2 12:11:16 2015) ansible code from github. When trying to ping (or really any ansible action) solaris 10 hosts, the following error is shown:

```
$ ./ansible --version | head -n1
ansible 2.1.0 (devel 9bf1aaf7f5) last updated 2015/12/02 21:28:57 (GMT +200)
$ ./ansible -m ping -i ~/Projects/ansible-solaris/hosts sol.local
sol.local | UNREACHABLE! => {
    "changed": false, 
    "msg": "ERROR! SSH Error: data could not be sent to the remote host. Make sure this host can be reached over ssh", 
    "unreachable": true
}
```

However, an older version of Ansibe (v1.9.1) works fine:

```
$ ansible --version 
ansible 1.9.1
$ ansible -m ping -i ~/Projects/ansible-solaris/hosts sol.local
sol.local | success >> {
    "changed": false,
    "ping": "pong"
}
```

The sol.local machine is: SunOS unknown 5.10 Generic_147148-26 i86pc i386 i86pc (python v 2.6.4)

Output of -vvvvv:

```
./ansible -vvvvv -m ping -i ~/Projects/ansible-solaris/hosts sol.local
Using /etc/ansible/ansible.cfg as config file
Loaded callback minimal of type stdout, v2.0
<127.0.0.1> ESTABLISH SSH CONNECTION FOR USER: fboender
<127.0.0.1> SSH: ansible.cfg set ssh_args: (-o)(ControlMaster=auto)(-o)(ControlPersist=60s)
<127.0.0.1> SSH: ANSIBLE_REMOTE_PORT/remote_port/ansible_port set: (-o)(Port=8022)
<127.0.0.1> SSH: ansible_password/ansible_ssh_pass not set: (-o)(KbdInteractiveAuthentication=no)(-o)(PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey)(-o)(PasswordAuthentication=no)
<127.0.0.1> SSH: ANSIBLE_REMOTE_USER/remote_user/ansible_user/user/-u set: (-o)(User=fboender)
<127.0.0.1> SSH: ANSIBLE_TIMEOUT/timeout set: (-o)(ConnectTimeout=10)
<127.0.0.1> SSH: PlayContext set ssh_common_args: ()
[...]

Issue #13409, opened 2015-12-02
Pull Request #13200 (pipelining + requiretty fix)  breaks other connection plugins


ThomasSteinbach:
I recently opened issue #13388 where I tried to introduce the `--user=<user>` option for Ansibles Docker connection plugin. I concluded within this issue, that my changes made the execution of Ansible tasks on Docker containers really slow.

Today I started a Playbook on another machine (at work), where I also deploy tasks to a container with the Docker connectio plugin and everything went fine, till I updated my Ansible installation. Since then every task executed on a Docker containe takes very long and Python consumes 100% CPU.

After some time of investigation it turns out that this behavior was caused by the pull request https://github.com/ansible/ansible/pull/13200 which was merged yesterday: https://github.com/ansible/ansible/commit/bebd2c5f3422b3c52ca0018d19ea4702c066c3e2.

Here is an example Playbook, which will reproduce the current behavior:

```
- name: this playbook is for testing purposes
  hosts: localhost
  tasks:

    - name: run jenkins container
      docker:
        name: jenkinstest
        image: jenkins
        state: started
        docker_user: root

    - name: add container to inventory
      add_host:
        name: jenkinstest
        ansible_connection: docker

    - name: run a task inside the container
      delegate_to: jenkinstest
      file:
        path: /root/test/me
        state: directory
```

If you reset your local repository to the mentioned pull request, this playbook will hang on the third task and Python consumes 100% CPU for a long time. If you reset your local repository to the commit before, everything works as expected and the playbook should be executed within seconds.

The changes made are too sophisticated for my humble Python knowledge. So @amenonsen please have a look, what could cause the behavior. Maybe @abadger has some good ideas too.

This issue would enable feature request #13388.

bitmensch:
Seems to be similar/same issue as #13407.

ThomasSteinbach:
Yes, this should be a duplicate. Seems I'd take me too much time for writing ;)

amenonsen:
I don't know why it would take 100% CPU, but I imagine the reason it hangs is because it's looking for the EOF marker and not finding one, because nothing writes it except the ssh plugin. So we really should be doing the pipelining command only when we're using ssh, not any other connection type. I don't see how to do that immediately, however; the command is constructed separately from the connection. I'll think about how to add an escape hatch in there somewhere. Thanks for tracking this down, sorry for the trouble.

abadger:
[...]

Issue #13407, opened 2015-12-02
ansible-playbook hangs running with "docker" connection


bitmensch:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible `devel`, commit f488de85997079f480d504f73537e3e33ff2495b
##### Ansible Configuration:

no changes
##### Environment:

ansible on Mac OS X 10.11.1
managing docker container with CentOS 6 (docker image "centos:centos6")
##### Summary:

Running ansible-playbook, a task executing on "docker" connection makes ansible-playbook process hang. The issue occurs since commit f488de85997079f480d504f73537e3e33ff2495b.
##### Steps To Reproduce:

I created a bash script to reproduce this issue:

``` bash
#!/bin/bash

# start docker container
CONTAINER_ID="$(docker run -d -t -h testhost centos:centos6)"

# create ansible inventory
echo "testhost ansible_connection=docker ansible_host=${CONTAINER_ID}" > inventory

# create ansible playbook
echo "- hosts: testhost" > playbook.yml
echo "  tasks:" >> playbook.yml
echo "    - name: host is reachable" >> playbook.yml
echo "      ping:" >> playbook.yml

# run ansible playbook
ansible-playbook -i inventory playbook.yml

# remove docker container
docker rm -f "${CONTAINER_ID}"
```

This fails since commit f488de85997079f480d504f73537e3e33ff2495b.
##### Expected Results:

Run the minimal playbook without hanging indefinitely.
##### Actual Results:

Minimal playbook hangs indefinitely.

abadger:
It seems that discsussion and more analysis are happening in #13409 so I'm close this one as a duplicate.  Note that currently the problematic commit should not be present (reverted) until this and other bugs are worked out of the code.

bitmensch:
Retested this with PR #13487. Now it is not hanging indefinitely anymore, but it fails with:

```
PLAY ***************************************************************************

TASK [setup] *******************************************************************
fatal: [testhost]: FAILED! => {"changed": false, "failed": true, "msg": "/usr/bin/python: can't find '__main__.py' in ''\n", "parsed": false}

NO MORE HOSTS LEFT *************************************************************

PLAY RECAP *********************************************************************
testhost                   : ok=0    changed=0    unreachable=0    failed=1
```

abadger:
@benjixx I'm going to copy your comment into PR #13487 so we know that it is a problem with that version of the PR.

amenonsen:
[...]

PR #13487, opened 2015-12-09
Make sudo+requiretty and ANSIBLE_PIPELINING work together

amenonsen:
Pipelining is a _significant_ performance benefit, because each task can
be completed with a single SSH connection (vs. one ssh connection at the
start to mkdir, plus one sftp and one ssh per task).

Pipelining is disabled by default in Ansible because it conflicts with
the use of sudo if 'Defaults requiretty' is set in /etc/sudoers (as it
is on Red Hat) and su (which always requires a tty).

We can (and already do) make sudo/su happy by using "ssh -t" to allocate
a tty, but then the python interpreter goes into interactive mode and is
unhappy with module source being written to its stdin, per the following
comment from connections/ssh.py:

```
    # we can only use tty when we are not pipelining the modules.
    # piping data into /usr/bin/python inside a tty automatically
    # invokes the python interactive-mode but the modules are not
    # compatible with the interactive-mode ("unexpected indent"
    # mainly because of empty lines)
```

Instead of the (current) drastic solution of turning off pipelining when
we use a tty, we can instead use a tty but suppress the behaviour of the
Python interpreter to switch to interactive mode. The easiest way to do
this is to make its stdin _not_ be a tty, e.g. with cat|python.

This works, but there's a problem: ssh will ignore -t if its input isn't
really a tty. So we could open a pseudo-tty and use that as ssh's stdin,
but if we then write Python source into it, it's all echoed back to us
(because we're a tty). So we have to use -tt to force tty allocation; in
that case, however, ssh puts the tty into "raw" mode (~ICANON), so there
is no good way for the process on the other end to detect EOF on stdin.
So if we do:

```
echo -e "print('hello world')\n"|ssh -tt someho.st "cat|python"
```

…it hangs forever, because cat keeps on reading input even after we've
closed our pipe into ssh's stdin. We can get around this by writing a
special **EOF** marker after writing in_data, and doing this:

```
echo -e "print('hello world')\n__EOF__\n"|ssh -tt someho.st "sed -ne '/__EOF__/q' -e p|python"
```

This works fine, but in fact I use a clever python one-liner by mgedmin
to achieve the same effect without depending on sed (at the expense of a
much longer command line, alas; Python really isn't one-liner-friendly).

There are two things to be careful of: that we use the new pipelining
behaviour only for _execute_module (and not, say, _remote_checksum),
and that we use it only with the ssh plugin (because the others don't
know to write the magic EOF marker). But the ssh plugin can't just do
the pipelining entirely by itself, because the build_module_command()
output may be wrapped by make_become_cmd().

We also enable pipelining by default as a consequence.

### DIFF
--- a/lib/ansible/constants.py
+++ b/lib/ansible/constants.py
@@ -245,7 +245,7 @@ MAX_FILE_SIZE_FOR_DIFF         = get_config(p, DEFAULTS, 'max_diff_size', 'ANSIB
 # CONNECTION RELATED
 ANSIBLE_SSH_ARGS               = get_config(p, 'ssh_connection', 'ssh_args', 'ANSIBLE_SSH_ARGS', '-o ControlMaster=auto -o ControlPersist=60s')
 ANSIBLE_SSH_CONTROL_PATH       = get_config(p, 'ssh_connection', 'control_path', 'ANSIBLE_SSH_CONTROL_PATH', "%(directory)s/ansible-ssh-%%h-%%p-%%r")
-ANSIBLE_SSH_PIPELINING         = get_config(p, 'ssh_connection', 'pipelining', 'ANSIBLE_SSH_PIPELINING', False, boolean=True)
+ANSIBLE_SSH_PIPELINING         = get_config(p, 'ssh_connection', 'pipelining', 'ANSIBLE_SSH_PIPELINING', True, boolean=True)
 ANSIBLE_SSH_RETRIES            = get_config(p, 'ssh_connection', 'retries', 'ANSIBLE_SSH_RETRIES', 0, integer=True)
 PARAMIKO_RECORD_HOST_KEYS      = get_config(p, 'paramiko_connection', 'record_host_keys', 'ANSIBLE_PARAMIKO_RECORD_HOST_KEYS', True, boolean=True)
 PARAMIKO_PROXY_COMMAND         = get_config(p, 'paramiko_connection', 'proxy_command', 'ANSIBLE_PARAMIKO_PROXY_COMMAND', None)
--- a/lib/ansible/plugins/action/__init__.py
+++ b/lib/ansible/plugins/action/__init__.py
@@ -183,7 +183,7 @@ class ActionBase(with_metaclass(ABCMeta, object)):
         if tmp and "tmp" in tmp:
             # tmp has already been created
             return False
-        if not self._connection.has_pipelining or not self._play_context.pipelining or C.DEFAULT_KEEP_REMOTE_FILES or self._play_context.become_method == 'su':
+        if not self._connection.has_pipelining or not self._play_context.pipelining or C.DEFAULT_KEEP_REMOTE_FILES:
             # tmp is necessary to store the module source code
             # or we want to keep the files on the target system
             return True
@@ -483,7 +483,9 @@ class ActionBase(with_metaclass(ABCMeta, object)):
                 # not sudoing or sudoing to root, so can cleanup files in the same step
                 rm_tmp = tmp
 
-        cmd = self._connection._shell.build_module_command(environment_string, shebang, cmd, arg_path=args_file_path, rm_tmp=rm_tmp)
+        python_interp = task_vars.get('ansible_python_interpreter', 'python')
+
+        cmd = self._connection._shell.build_module_command(environment_string, shebang, cmd, arg_path=args_file_path, rm_tmp=rm_tmp, python_interpreter=python_interp, connection=self._connection)
         cmd = cmd.strip()
 
         sudoable = True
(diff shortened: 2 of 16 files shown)


amenonsen:
This should fix #13411 and maybe #13401 and #13410.

@fboender @tobinlandricombe could you please test with this patch applied? Thanks.

amenonsen:
And maybe #13409 and #13407 as well. @ThomasSteinbach @benjixx could you please test?

fboender:
@amenonsen Confirmed the fix. The latest develop branch is working again with solaris ( see #13410 ). Thank you!

bitmensch:
@amenonsen Retested with #13407, but it's now failing with another error (see #13407 for details).

abadger:
From @benjixx:

ansible on Mac OS X 10.11.1
managing docker container with CentOS 6 (docker image "centos:centos6")
Summary:

Running ansible-playbook, a task executing on "docker" connection.  No longer hanging indefinitely  but it fails with:

```

fatal: [testhost]: FAILED! => {"changed": false, "failed": true, "msg": "/usr/bin/python: can't find '__main__.py' in ''\n", "parsed": false}

testhost                   : ok=0    changed=0    unreachable=0    failed=1
```

bitmensch:
@amenonsen Your new commit makes #13407 run successfully.

abadger:
Transient failure seems to still exist:

```
$ git branch
  devel
* pr/13487
```

Here's a gist with the log for the task that failed this time:
https://gist.github.com/abadger/d30214bbd14cc0a9f9ab

amenonsen:
@abadger this is pretty weird. It's executing a checksum task without -tt, so not only should python not be in interactive mode to begin with, but it shouldn't see anything on its stdin (and indeed we don't log sending any initial data), let alone `''' Find out current state '''`, which is from lib/ansible/modules/core/files/file.py's get_state() method. So at first glance, none of the pipelining code seems to even be involved in the failure. Even the checksum one-liner only imports hashlib/sha.

**Update** it's not the checksum task that's failing, just debug output from different workers confusing me. Investigating the actual failure now.

amenonsen:
This is the session that's failing:

```
<noc01.phx2.fedoraproject.org> SSH: EXEC ssh -C -vvv -o ControlMaster=auto -o ControlPersist=60s -o Port=22 -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o User=root -o ConnectTimeout=90 -o ControlPath=/root/.ansible/cp/ansible-ssh-%h-%p-%r -tt noc01.phx2.fedoraproject.org LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8 LC_MESSAGES=en_US.UTF-8 python -uc 'import sys; [sys.stdout.write(s) for s in iter(sys.stdin.readline, "__EOF__942d747a0772c3284ffb5920e234bd57__\n")]'| LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8 LC_MESSAGES=en_US.UTF-8 /usr/bin/python
```

I again notice the python vs. /usr/bin/python distinction, but it really shouldn't matter. Other than that, though, I don't see anything out of the ordinary.

amenonsen:
@fboender has said that the python vs. /usr/bin/python thing doesn't make any difference in his case. I'm inclined to believe it doesn't here either.

abadger:
Yeah, it shouldn't make a difference here:

```
[toshio@noc01 ~]$ which --all python
/usr/bin/python
[toshio@noc01 ~]$
```

amenonsen:
@abadger any more ideas?

mattclay:
@amenonsen: This change prevents the raw action from working with ssh and sudo.

It works with devel:

```
$ ansible -m raw -a whoami -i localhost, all --ask-become-pass --become --become-method=sudo
SUDO password: 
localhost | SUCCESS | rc=0 >>

root
```

But not with your changes:

```
$ ansible -m raw -a whoami -i localhost, all --ask-become-pass --become --become-method=sudo
SUDO password: 
localhost | FAILED | rc=1 >>
sudo: sorry, you must have a tty to run sudo
```

jimi-c:
Moving to 2.2 milestone, as it does not appear to be ready and missed the cutoff for 2.1.

mckerrj:
@amenonsen still looking for this? Can you review/fix or close

jimi-c:
After some further discussion, we're going to close this for now due to multiple issues with the above, and the fact that it's an issue that does not really seem to be impacting users much anymore. Pipelining has been around for over 2 years, and it's well known and documented that requiretty needs to be disabled for it to work.

Thank you!

PR #13488, opened 2015-12-09
avoid persistent containers in attribute defaults

bcoca:
moved from the field attribute declaration and created a placeholder
which then is resolved in the field attribute class.

this is to avoid unwanted persistent of the defaults across objects which introduces
stealth bugs when multiple objects of the same kind are used in succession while
not overriding the default values.


jimi-c:
Merged.

PR #13494, opened 2015-12-09
add tests for encrypted hash mysql_user

Jmainguy:


Jmainguy:
This is to support https://github.com/ansible/ansible-modules-core/pull/1581

@abadger ^^

abadger:
@jmainguy Just need to add some asserts to the test: https://gist.github.com/abadger/423fbfcc70749acd1901

The asserts check that the module returned the value we expected and fail if it isn't.  (assert is just a module and you can see it's documentation with the other module docs: http://docs.ansible.com/assert_module.html )

abadger:
Thanks @Jmainguy !

PR #13502, opened 2015-12-10
use shallow clones by default in ansible-pull

### LINKED ISSUE #8620: ansible-pull shallow option
##### Issue Type:

Fetaure Request
##### Ansible Version:

ansible 1.7
##### Environment:

Ubuntu 14.04
##### Summary:

The ansible-pull command should have a switch to enable/disable shallow clone/pull of the git repository (... -m git -a "... depth=1"). In most cases there is absolutly no need to have the complete commit history on the client.

Shallow shoud be standard.
##### Steps To Reproduce:

None
##### Expected Results:

Pull a shallow copy of the git repository
##### Actual Results:

None so far

robinro:
Allows shallow checkouts in ansible-pull by adding `--depth 1` (or higher number), fixes #8620 

Maybe it would be nicer to do this by default and add a `--full` option for a complete checkout.

The patch is work-in-progress and not thoroughly tested yest.

### DIFF
--- a/lib/ansible/cli/pull.py
+++ b/lib/ansible/cli/pull.py
@@ -156,7 +156,7 @@ class PullCLI(CLI):
 
             if self.options.verify:
                 repo_opts += ' verify_commit=yes'
-                
+
             if not self.options.fullclone:
                 repo_opts += ' depth=1'
 
--- a/lib/ansible/cli/pull.py
+++ b/lib/ansible/cli/pull.py
@@ -80,8 +80,8 @@ class PullCLI(CLI):
             help='directory to checkout repository to')
         self.parser.add_option('-U', '--url', dest='url', default=None,
             help='URL of the playbook repository')
-        self.parser.add_option('--depth', dest='depth', default=None,
-            help='Depth of checkout, shallow checkout if greater or equal 1 . Defaults to full checkout.')
+        self.parser.add_option('--full', dest='fullclone', action='store_true',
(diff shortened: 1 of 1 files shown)


bcoca:
it will probably not be the default as many users would expect resources needed for the play to be there, even if at other depths.

robinro:
@bcoca I don't really understand the concern, If you do a checkout with depth=1, the working copy of the checkout is identical to the working copy of the full checkout. Ansible-pull only looks at the working copy and thus not use any git features after the initial clone/pull.

The difference is that when you manually try to look for the history of the repo it's empty. This might affect later updates, which could be less efficient, if there are objects reused that are not present in the shallow checkout. I'm not sure git goes so far to do this.

bcoca:
nah, just me being stupid .... crossed cables and forgot what depth meant in git context.

robinro:
Today I tested shallow clones on our systems and it works beautifully. The initial clone is significantly faster (cut the transmitted size by a factor 10!) and pulling updates works without issues.

Thus I updated this pull request to set shallow clones as default and only provide a `--full` option for people wanting this. For the moment I can't think of any reason outside of development/testing, but there typically the checkout will be manually before.

This change will have direct performance benefits on most production environments with long history in the cloned repository which in my opinion is sufficient to change the default.

robinro:
The failed travis check is some other issue. I ran the version locally without issues and the failure is in some unrelated testing routine.

PR #13512, opened 2015-12-11
Clean up debug logging around _low_level_execute_command

amenonsen:
We were logging the command to be executed many times, which made debug
logs very hard to read. Now we do it only once.

Also makes the logged ssh command line cut-and-paste-able (the lack of
which has confused a number of people by now; the problem being that we
pass the command as a single argument to execve(), so it doesn't need an
extra level of quoting as it does when you try to run it by hand).


abadger:
cherry-picked to stable-2.0 today as I may need to make changes to the arguments we feed to display.*() to prevent UnicodeError tracebacks in some circumstances.

PR #13519, opened 2015-12-11
issue 2625

vmindru:
we can use this as a workaround, until we decide what is to be done with the issue itself.
https://github.com/ansible/ansible-modules-core/issues/2625

### DIFF
--- a/lib/ansible/plugins/action/async.py
+++ b/lib/ansible/plugins/action/async.py
@@ -70,7 +70,6 @@ class ActionModule(ActionBase):
         async_cmd = " ".join([str(x) for x in [env_string, async_module_path, async_jid, async_limit, remote_module_path, argsfile]])
         result.update(self._low_level_execute_command(cmd=async_cmd))
 
-        # clean up after
         time.sleep(1)
         if tmp and "tmp" in tmp and not C.DEFAULT_KEEP_REMOTE_FILES:
             self._remove_tmp_path(tmp)
--- a/lib/ansible/plugins/action/async.py
+++ b/lib/ansible/plugins/action/async.py
@@ -19,6 +19,7 @@ __metaclass__ = type
 
 import json
 import random
+import time
 
 from ansible import constants as C
 from ansible.plugins.action import ActionBase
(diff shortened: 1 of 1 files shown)


jimi-c:
I understand the problem, it looks like we should be checking the status of the async task properly there and making sure we get a started=true result before deleting the tmp dir. Doing an arbitrary sleep there will probably work 99% of the time, however I'd prefer this to be fixed properly and not covered up by that sleep.

vmindru:
surely a proper fix is better than a simple sleep, the whole poitn is rather  a workaround for someone who needs this fixed sooner then later and , if we don't have time for a proper fix , and to revel the actual problem  @jimi-c  what is the mailing list where such topics should be discussed, e.g. a collective decision on what shall be the approach to get this  fixed.  I am failry new to the community and not familiar with the habbits around.

vmindru:
Ok let's cancel this PR , this actually should be done in the async_wrapper , initial sleep here was 1 sec and seems not enough on fast connection systems, also as teh comment sugests on the initial code probably this has to be done by IPC. 

```
vmindru@vmhodesk:~/proj/ansible/ansible-devel/lib/ansible/modules/core/utilities/logic$ git diff
diff --git a/utilities/logic/async_wrapper.py b/utilities/logic/async_wrapper.py
index 55f5283..884447c 100644
--- a/utilities/logic/async_wrapper.py
+++ b/utilities/logic/async_wrapper.py
@@ -153,7 +153,7 @@ if __name__ == '__main__':
             notice("Return async_wrapper task started.")
             print json.dumps({ "started" : 1, "ansible_job_id" : jid, "results_file" : job_path })
             sys.stdout.flush()
-            time.sleep(1)
+            time.sleep(2)
             sys.exit(0)
         else:
             # The actual wrapper process
vmindru@vmhodesk:~/proj/ansible/ansible-devel/lib/ansible/modules/core/utilities/logic$ 
```

willthames:
@vmindru - here's the community page with details about where to have discussions: http://docs.ansible.com/ansible/community.html

This one would fit well in ansible-devel (mailing list or IRC!)

PR #13528, opened 2015-12-12
choose a random filename for the remote module file

### LINKED ISSUE #13527: yum module doesn't call rhnplugin, because module-file is named yum
When a RHEL/CentOS server is managed via spacewalk/rhn/satelite, yum uses `yum-rhn-plugin` to get meta-data from the spacewalk server, instead of querying the repos directly.

Example task:

``` yaml
- yum: name=curl disablerepo=* enablerepo=centos-7-x86_64-cityfan
```

(`cityfan` => http://www.city-fan.org/ftp/contrib/yum-repo/rhel7/x86_64/)

Generated output:

``` json
{
  "msg": "Error setting/accessing repos: Error getting repository data for centos-7-x86_64-cityfan-curl-only, repository not found",
  "failed": true
}
```

[`rhnplugin.py#L94-L104`](https://github.com/spacewalkproject/spacewalk/blob/4d2d833a1d5baf92a4c01883e04ac2aa9f57aa0e/client/rhel/yum-rhn-plugin/rhnplugin.py#L94-L104) checks, if the basename of the invoked command is `yum` and if there are no parameters (`cmd_args == []`).

The problem is that ansible executes the yum-module under a path similar to `/home/ansible/.ansible/tmp/ansible-tmp-144929.24-4078098/yum`. So the basename is `yum` and there are no parameters, which causes `yum-rhn-plugin` to never actually query the server. The repolist stays empty, `centos-7-x86_64-cityfan` is never found and the task fails here [yum.py#L926](https://github.com/ansible/ansible-modules-core/blob/74609ce42cb140dba528fee18422467a32acd995/packaging/os/yum.py#L926), with `enablerepo` being the repo name(s).

Spacewalk/yum isn't the only project to check, if it's executed in a certain way by checking its `argv[0]`. So, in order to fix this, I'd suggest to randomize the executable name, just like the path to the dropped ansible-module python-file: `yum-6812`.

What are your thoughts on this?

luto:
Suggested fix for #13527.
Please see the issue for additional information.

### DIFF
--- a/lib/ansible/plugins/shell/sh.py
+++ b/lib/ansible/plugins/shell/sh.py
@@ -55,7 +55,7 @@ class ShellModule(object):
 
     # some shells (eg, powershell) are snooty about filenames/extensions, this lets the shell plugin have a say
     def get_remote_filename(self, base_name):
-        return base_name.strip()
+        return '%s-%s' % (base_name.strip(), random.randint(0, 2**24))
 
     def path_has_trailing_slash(self, path):
         return path.endswith('/')


luto:
According to a quick bisect the tests seem to fail starting with commit 7f7e730dea36dbb709b47c39ca1a28cb9f6cb3f1 ("Don't mark hosts failed if they've moved to a rescue portion of a block"). Most likely due to the unmerged stash already pointed out by @benjixx.

bitmensch:
@luto The merge artifacts have already been removed in a follow-up commit by @bcoca. But, yes, it looks like tests are failing due to the commit you mentioned, anyway.

mscherer:
See #13530 for tests fixes. I just need to fix the last one, but it involve multiprocess, so a bit  tricky.

luto:
Have you had a chance to look at this?

luto:
rebased onto the currentl `devel`, which fixes the failing tests (see earlier comments).

luto:
Since this MR was ignored by the team for more than a year now, I'm going to close it.

PR #13529, opened 2015-12-12
Add tests for ansible.module_utils.known_hosts

mscherer:
I had written others tests cases (such as ipv6 addresses), but the current code fail, so I propose first incomplete tests cases, then add the fix and tests for the fix.


jimi-c:
Merged, thanks!

PR #13530, opened 2015-12-12
Complete the mock object for tests, broken by 7f7e730de

mscherer:
There is 2 others broken tests in devel, so this one only fix 1/3


jimi-c:
Thanks @mscherer, I ended up fixing this a slightly different way (and spotted a bug due to my fix as well).

PR #13533, opened 2015-12-13
debug now validates its params

### LINKED ISSUE #13532: debug action plugin silently ignores bad parameters
After looking through the code and noticing the `fail` option for the `debug` action plugin, I decided to try it out. I messed up and set `failed=yes` and was confused about why nothing happened.

The plugin should probably tell me that I'm being an idiot.

bcoca:
simplified var handling
made default message the same as in pre 2.0
fixes #13532

### DIFF
--- a/lib/ansible/plugins/action/debug.py
+++ b/lib/ansible/plugins/action/debug.py
@@ -20,40 +20,45 @@ __metaclass__ = type
 from ansible.plugins.action import ActionBase
 from ansible.utils.boolean import boolean
 from ansible.utils.unicode import to_unicode
+from ansible.errors import AnsibleUndefinedVariable
 
 class ActionModule(ActionBase):
     ''' Print statements during execution '''
 
     TRANSFERS_FILES = False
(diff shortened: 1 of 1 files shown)


lib/ansible/plugins/action/debug.py
```diff
@@ -20,40 +20,45 @@
...
+
+        elif 'var' in self._task.args:
+            try:
+                results = self._templar.template(self._task.args['var'], convert_bare=True, fail_on_undefined=True)
+                if results == self._task.args['var']:
+                    raise AnsibleUndefinedVariable
+            except AnsibleUndefinedVariable:
+                results = "VARIABLE IS NOT DEFINED!"
+
             if type(self._task.args['var']) in (list, dict):
```
ueg1990, line 55: why not use isinstance function as that is preferred over type?

bcoca, line 55: not sure, i did not change that line

PR #13535, opened 2015-12-13
add --full flag to ansible-pull man page

robinro:
add --full flag that was added in #13502


robinro:
Thanks @bcoca for the quick merges! :+1:

PR #13536, opened 2015-12-13
Add version_check callback plugin.

tjanez:
This callback module checks if the Ansible version that is used to run
the playbook matches the Ansible version specified in the requirements
file.

### DIFF
--- a/lib/ansible/plugins/callback/version_check.py
+++ b/lib/ansible/plugins/callback/version_check.py
@@ -1,0 +1,92 @@
+# (c) 2015, Tadej Janez <tadej.j@nez.si>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


lib/ansible/plugins/callback/version_check.py
```diff
@@ -0,0 +1,92 @@
...
+
+    If the version of Ansible is not specified, return None.
+
+    """
+    version = None
+    with open(req_path) as req_file:
+        for req in req_file.readlines():
+            req = req.strip()
+            if req.lower().startswith("ansible"):
+                req_split = req.split("==")
```
ueg1990, line 43: you should also include the separator '>='

lib/ansible/plugins/callback/version_check.py
```diff
@@ -0,0 +1,92 @@
...
+
+class CallbackModule(CallbackBase):
+    """
+    This callback module checks if the Ansible version that is used to run the
+    playbook matches the Ansible version specified in the requirements file.
+
+    To use it, just place 'requirements.txt' file in your playbook directory
+    with the following contents (replace '2.0.0' with your required version):
+
+    ansible=2.0.0
```
ueg1990, line 58: ansible==2.0.0

bcoca:
I'm not sure that a callback is where this belongs, we might be better off by adding a version pragma or just have a task checking the ansible_version variable.

sivel:
Question About This One
=======================

Hi!

PR #13538, opened 2015-12-14
Fix multiple backup

arthurdejong:
These commits fix a problem when multiple backup files are created within the same second. Previously the original file contents would be lost.

This changes the behaviour to not overwrite existing backup files. It is better to have only one backup file for a number of related changes than to not have a backup of the original file contents at all.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -1596,13 +1596,12 @@ class AnsibleModule(object):
             # backups named basename-YYYY-MM-DD@HH:MM:SS~
             ext = time.strftime("%Y-%m-%d@%H:%M:%S~", time.localtime(time.time()))
             backupdest = '%s.%s' % (fn, ext)
-
-            try:
-                shutil.copy2(fn, backupdest)
-            except (shutil.Error, IOError):
-                e = get_exception()
-                self.fail_json(msg='Could not make backup of %s to %s: %s' % (fn, backupdest, e))
-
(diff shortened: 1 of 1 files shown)


bcoca:
i'm not sure if we should fail here or just try to add and increment/pid/milisecs to the file name

arthurdejong:
Hi Brian,

We run into this problem very often because we use the lininfile and with_items together a lot. Failing the task would be a serious regression as most of our playbooks would fail.

Also, it is not very useful (at least not for us) to have a backup file for each item that changed something in with_items but it is useful to have a backup from before the task. The commit effectively just changes the functionality from keeping the last backup within a task to the first.

Thanks.

bcoca:
as I read the code, if the backup fails it actually exists the module in failure, that does not seem to mesh with what you are saying.

arthurdejong:
Hi Brian,

Are you referring to the original code or my change?

If I understand the code correctly (I'm new to Ansible internals), the error is set with fail_json(). I assumed that calling fail_json() results in the task to fail.

The only case that fail_json() would be called is when an exception is caught. shutil.copy2() just overwrites any existing file and would only raise an exception on a read-only file system, file system full, permission problems, etc. but not when the file already exists.

With my change the call to copy2() would be skipped altogether.

Thanks.

bcoca:
Closing as we did an alternate fix via 123d54e7365 by including the pid in the backup name to prevent the filename matching soley on time stamp. If this is still a problem for you even with the current code, we are open to revisit this decision.

PR #13550, opened 2015-12-15
Use wrapped connect_to_region everywhere in ec2.py

KrzysiekJ:
This is a bug fix which, in particular, adds respecting of `boto_profile` when fetching ElastiCache data.


bcoca:
cc @defionscode @lwade

juliedavila:
shipit

Issue #13563, opened 2015-12-15
[2.0-rc1] ansible-galaxy no longer respects path in requirements file


lae:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 2.0.0 (f2225395f96d5c309dddc66fafd984862d99b708)
##### Ansible Configuration:

no changes
##### Environment:

Arch Linux
##### Summary:

My existing requirements.yml files from using 1.9.4 and older now don't install roles into their specified directories, but instead tries to use the global roles directory.
##### Steps To Reproduce:

Just using a random role from GitHub:

```
$ cd $(mktemp -d)
$ cat >requirements.yml <<<EOL
- src: git+https://github.com/sfromm/ansible-epel.git
  name: epel
  path: ./
EOL
$ ansible-galaxy install -r requirements.yml 
```
##### Expected Results:

```
$ ansible-galaxy install -r requirements.yml 
- executing: git clone https://github.com/sfromm/ansible-epel.git epel
- executing: git archive --prefix=epel/ --output=/tmp/tmp4UEDr0.tar HEAD
- extracting epel to ./epel
- epel was installed successfully
```
##### Actual Results:

```
$ ansible-galaxy install -r requirements.yml 
- extracting epel to /etc/ansible/roles/epel
 [WARNING]: - epel was NOT installed successfully: ERROR! Could not update files in /etc/ansible/roles/epel: [Errno 13]
Permission denied: '/etc/ansible/roles'

ERROR! - you can use --ignore-errors to skip failed roles and finish processing the list.
```

chrrrles:
How are you setting roles-path for ansible-galaxy.  Have you tried this?:

```
$ ansible-galaxy install -r requirements.yml --roles-path=./
```

lae:
Using --roles-path works as described. Is the path parameter in requirements.yml being deprecated in favor of this flag, though? It's still documented here: https://github.com/ansible/ansible/blob/devel/docsite/rst/galaxy.rst#advanced-control-over-role-requirements-files

jimi-c:
cc @chouseknecht

chrrrles:
@lae Ahhh, I see it now.  I believe the referenced PR will fix your bug, may you test it also?

chrrrles:
cc @docschick

chouseknecht:
I tested with the code change found in PR 13577.  Adding 'path' back into the VALID_SPEC_KEYS list in requirements.py works. 

[...]

PR #13577, opened 2015-12-16
adding path back to galaxy role's valid spec list

chrrrles:


chrrrles:
Fixes https://github.com/ansible/ansible/issues/13563

Issue #6407, opened 2014-03-11
Incorrect warning with with_fileglob


jefferai:
When using the following with_fileglob statement:

with_fileglob: "{{ bootstrap_location }}/debs/*"

I get the following:

[WARNING]: It is unneccessary to use '{{' in loops, leave variables in loop expressions bare.

(Note: "unnecessary" is also misspelled in the warning message -- it should have one c, not two.)

However, this cannot be right; without the {{ }} there's no way it would know what to substitute. mdehann in #ansible confirmed this.

---

Much later edit:

Before anyone protests, note that I am following the instructions of the bot extremely literally...and that it has already been acknowledged/triaged/resolution suggested and intended by @mpdehaan 
##### Issue Type:

What kind of ticket is this?  You can say â€œBug Reportâ€, â€œFeature Ideaâ€, â€œFeature Pull Requestâ€, â€œNew Module Pull Requestâ€, â€œBugfix Pull Requestâ€, â€œDocumentation Reportâ€, or â€œDocs Pull Requestâ€.
##### Ansible Version:

Please supply the verbatim output from running â€œansible --versionâ€. 
##### Environment:

What OS are you running Ansible from and what OS are you managing?  Examples include RHEL 5/6, Centos 5/6, Ubuntu 12.04/13.10, *BSD, Solaris.  If this is a generic feature request or it doesnâ€™t apply, just say â€œN/Aâ€. 
##### Summary:

Please summarize your request in this space.  You will earn bonus points for being succinct, but please add enough detail so we can understand the request.
##### Steps To Reproduce:

If this is a bug ticket, please enter the steps you use to reproduce the problem in the space below.  If this is a feature request, please enter the steps you would use to use the feature.  If an example playbook is useful, please include a short reproducer inline, indented by four spaces.  If a longer one is necessary, please link one uploaded to gist.github.com. 
##### Expected Results:

Please enter your expected results in this space.  When running the steps supplied above, what would you expect to happen?  If showing example output, indent your output by four spaces so it will render correctly in GitHub. 
##### Actual Results:

Please enter your actual results in this space.  When running the steps supplied above, what actually happened?  If showing example output, indent your output by four spaces so it will render correctly in GitHub.

mpdehaan:
Basically what i've suggested is if the line doesn't end in "}}" we should not print the warning.

Should be an easy addition.

bcoca:
with_fileglob: bootstrap_location + "/debs/*"

[...]

PR #13586, opened 2015-12-17
Update with_items + variable example to remove unnecessary quotes

pimterry:
If you use the example as written it does work fine, but you're given a warning message:

```
[WARNING]: It is unnecessary to use '{{' in loops, leave variables in loop expressions bare.
```

As pointed out in https://github.com/ansible/ansible/issues/6407#issuecomment-37362386 it's preferable to drop the quotes in this case, for simplicity and elegance. It'd be good to follow that in the docs too, rather than lead people into confusing warning messages.

### DIFF
--- a/docsite/rst/playbooks_loops.rst
+++ b/docsite/rst/playbooks_loops.rst
@@ -23,7 +23,7 @@ To save some typing, repeated tasks can be written in short-hand like so::
 
 If you have defined a YAML list in a variables file, or the 'vars' section, you can also do::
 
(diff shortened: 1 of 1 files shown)


bcoca:
that warning has been removed in newer versions of ansible and the example you are showing will be deprecated and eventually stop working. So I'm not going to merge this change.

pimterry:
Ah, yes, fair enough! Turns out my Ansible install is a couple of versions behind, sorry about that.

PR #13580, opened 2015-12-16
Revert to standard port if non-standard port not open.

crlb:
If a host is to be built and use a non-standard ssh port, it is reasonable that changes to its' configuration (ssh, iptables, firewall, etc.) be accomplished through ansible. Consider the following playbook which modifies the ssh port in the common role:

> > ---
> > # file: cloud.yaml - Build an OpenStack cloud.
> > - hosts: osctl:oscmp
> >   roles:
> >   - common
> > - hosts: osctl
> >   roles:
> >   - osctl
> > - hosts: oscmp
> >   roles:
> >   - oscmp
> > - hosts: osctl
> >   tasks:
> >   - include: roles/oscmp/tasks/verify.yaml

The playbook fails after the second play with an ssh port error. The inventory has to be changed for both hosts and the playbook rerun.

This commit tests the non-standard port on the target and if it is not open sets the standard ssh port. The inventory can then be created with all non-standard ports defined, and will execute correctly even on the first execution. This is particularly useful when repetitively rebuilding Virtual Machines.

This was developed and tested  on ansible 1.9, and a second fix/pull request submitted for version 2.

PR #13589, opened 2015-12-17
Revert to standard port if non-standard port not open.

crlb:
This is ansible version 2 patch to revert to the standard ssh port if the non-standard port is not open. Please refer to the stable-1.9 pull request for the rationale for this patch. It has been tested and works as expected.


bcoca:
same as #13580

bcoca:
closing as discussed in linked tickets

PR #13593, opened 2015-12-18
Add 'filtered' stdout callback plugin

agaffney:
This plugin filters output for any task that is 'ok' or 'skipped'.
It works by subclassing the 'default' stdout callback plugin and
overriding certain functions. It will suppress display of the task
banner until there is a 'changed' or 'failed' result or an
unreachable host.

### DIFF
--- a/lib/ansible/plugins/callback/filtered.py
+++ b/lib/ansible/plugins/callback/filtered.py
@@ -1,0 +1,76 @@
+# (c) 2015, Andrew Gaffney <andrew@agaffney.org>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+# Make coding more python3-ish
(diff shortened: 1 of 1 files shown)


agaffney:
Example output:

```
$ ansible-playbook main.yml 

fatal: [shell.foo.com]: UNREACHABLE! => {"changed": false, "msg": "ERROR! SSH encountered an unknown error during the connection. We recommend you re-run the command using -vvvv, which will enable SSH debugging output to help diagnose the issue", "unreachable": true}
^C
raz1                       : ok=5    changed=0    unreachable=0    failed=0   
shell.foo.com              : ok=0    changed=0    unreachable=1    failed=0   
```

and

```
$ ansible-playbook main.yml 

TASK [ansible : inventory file] ************************************************
changed: [raz1]

raz1                       : ok=21   changed=1    unreachable=0    failed=0   
shell.foo.com              : ok=8    changed=0    unreachable=0    failed=0   
```

agaffney:
@bcoca thinks that my plugin name is boring and wants something a little...different. Suggestions are welcome :)

bcoca:
current entries: whiney, complainer, crybaby, narc, tattler,  bark

bcoca:
@agaffney going to need a GPL header

agaffney:
Perhaps that should be added to the code tests :)

bcoca:
we have it for modules (thanks @sivel) and yes, we should have it for everything else also, except modue_utils, which needs BSD test

agaffney:
I've pushed a new revision with the block copied from default.py and my own copyright. I noticed that there are other plugins in the `lib/ansible/plugins/callback` folder that do NOT have this comment block, such as `syslog_json.py` and `timer.py`.

agaffney:
The py34 and py35 tests failed, but it's not clear at all why.

bcoca:
damn, i thought i fixed those 2 files ....

ignore the tests, bad thing about travis is that once broken any new PR will appear broken, it really should be done by checking out main branch and merging/cherry-picking the change to test, otherwise we keep obscuring/introducing test errors.

PR #13606, opened 2015-12-18
Fix exeption, replace 'as' by ',' as in 31c0e523

yogo1212:


sivel:
This change unfortunately cannot be made.  The files in `module_utils` must support python 2.4 at this time.

The work around used elsewhere is to make use of the `get_exception` function provided by `basic.py`.

bcoca:
closing as mentioned above this breaks 2.4 compatibility.

PR #13609, opened 2015-12-19
Enable use of non standard hostname in inventory

### LINKED ISSUE #13608: No longer able to use non standard hostname in inventory
I am trying to work on a connection plugin using guestfish to modify a disk image directly. And while testing on devel ( 07a00593066cb ), I stumbled on a few hurdles like this:

```
$ ansible -vvv all -c guestfs -i 'Fedora-Cloud-Atomic-23-20151215.x86_64.qcow2:/dev/sda1,'  -m ping 
Using /home/misc/.ansible.cfg as config file
Unexpected Exception: 'NoneType' object has no attribute 'split'
the full traceback was:

  File "/home/misc/checkout/git/ansible/bin/ansible", line 85, in <module>
    sys.exit(cli.run())
  File "/home/misc/checkout/git/ansible/lib/ansible/cli/adhoc.py", line 187, in run
    result = self._tqm.run(play)
  File "/home/misc/checkout/git/ansible/lib/ansible/executor/task_queue_manager.py", line 237, in run
  File "/home/misc/checkout/git/ansible/lib/ansible/plugins/strategy/linear.py", line 217, in run
    task_vars = self._variable_manager.get_vars(loader=self._loader, play=iterator._play, host=host, task=task)
  File "/home/misc/checkout/git/ansible/lib/ansible/vars/__init__.py", line 245, in get_vars
    all_vars = combine_vars(all_vars, host.get_vars())
  File "/home/misc/checkout/git/ansible/lib/ansible/inventory/host.py", line 129, in get_vars
    results['inventory_hostname_short'] = self.name.split('.')[0]
AttributeError: 'NoneType' object has no attribute 'split'
```

After digging, I found that this function https://github.com/ansible/ansible/blob/devel/lib/ansible/parsing/utils/addresses.py#L167 kinda filter all non standards address, and reutnr None,None unless I am careful into what I give for argument.

This also break connection plugins like chroot, jail and potentially the docker one.

mscherer:
If the host given cannot be parsed in hostnae:port tuple,
then it might be a different format and used by a alternative
connexion plugin.

Fixes #13608

### DIFF
--- a/lib/ansible/inventory/__init__.py
+++ b/lib/ansible/inventory/__init__.py
@@ -110,7 +110,10 @@ class Inventory(object):
         elif isinstance(host_list, list):
             for h in host_list:
                 (host, port) = parse_address(h, allow_ranges=False)
-                all.add_host(Host(host, port))
+                if host is not None:
+                    all.add_host(Host(host, port))
+                else:
+                    all.add_host(Host(h))
         elif self._loader.path_exists(host_list):
             #TODO: switch this to a plugin loader and a 'condition' per plugin on which it should be tried, restoring 'inventory pllugins'
             if self.is_directory(host_list):


abadger:
Note: This behaviour may be by design.  If so, we wouldn't want this but it would probably be appropriate to throw some sort of warning or error here instead of passing None to the constructor and getting an error.

Discussing in the bug report whether what the desired behaviour is before deciding what direction this code should go in.

jimi-c:
@mscherer if the main use case here is that the hostname in question is a path, for instance when using the chroot connection plugin or something similar, the hostname should be at least validated as a path to avoid allowing garbage in. Could you please update this to at least try to rudimentary path validation?

bcoca:
closing this in favor of #13622

PR #13612, opened 2015-12-20
Add a guestfs connexion plugin

mscherer:
This permit to edit a base image in a similar fashion to packer,
thus permitting for example to customize base image before uploading
them directly to Glance or similar services for openstack, enabling
immutable server deployment.

It can be used like this:

```
$ cat hosts
fedora_22 ansible_ssh_host=Fedora-Cloud-Base-22-20150521.x86_64.qcow2 ansible_connection=guestfs

$ cat deploy.yml

---
- hosts: fedora_22
  roles:
  - somerole

$ ansible-playbook -i hosts deploy.yml
[....]
```

For now, it only support local VM with single OS, but it should be trivial
to add support for remote disks and multiple OS.

### DIFF
--- a/lib/ansible/plugins/connection/guestfs.py
+++ b/lib/ansible/plugins/connection/guestfs.py
@@ -1,0 +1,215 @@
+# Based on local.py (c) 2012, Michael DeHaan <michael.dehaan@gmail.com>
+# (c) 2013, Maykel Moya <mmoya@speedyrails.com>
+# (c) 2015, Toshio Kuratomi <tkuratomi@ansible.com>
+# (c) 2015, Michael Scherer <mscherer@redhat.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import os
+import os.path
+try:
+    import guestfs
+except ImportError:
+    raise AnsibleError("guestfs python bindings are not installed")
+
+import json
+import re
+
+from ansible.errors import AnsibleError
+from ansible.plugins.connection import ConnectionBase
+from ansible.module_utils._text import to_native
+
+try:
+    from __main__ import display
+except ImportError:
+    from ansible.utils.display import Display
+    display = Display()
+
+# keep that code python2 and python3 compliant
+# we need to set $HOME so we are not left with
+# .ansible directory on /
+# also, I kept it {} free for clarity, hence the use of
+# disct instead of the short form with brackets.
+EXECUTION_SCRIPT = """
+import sys
+import json
+import os
+from subprocess import PIPE, Popen
+
+os.environ['HOME'] = '{}'
+
+cmd_file = sys.argv[1]
+if not os.path.exists(cmd_file):
+    sys.exit(1)
+
+cmd = json.loads(open(cmd_file).read())
+if type(cmd) == type(u''):
+    cmd = cmd.encode('UTF-8')
+r = Popen(cmd, stdout=PIPE, stderr=PIPE, shell=True)
+(stdout, stderr) = r.communicate()
+
+output = dict()
+output['rc'] = r.returncode
+output['stdout'] = stdout.decode('UTF-8')
+output['stderr'] = stderr.decode('UTF-8')
+
+json.dump(output, open(cmd_file + '.out', 'w'))
+"""
+
+
+class Connection(ConnectionBase):
+    ''' Libguestfs based connections '''
+
+    transport = 'guestfs'
+    # not sure how it would react, so better disable it
+    has_pipelining = False
+
+    def __init__(self, play_context, new_stdin, *args, **kwargs):
+        super(Connection, self).__init__(play_context, new_stdin,
+                                         *args, **kwargs)
+
+        self.disk = self._play_context.remote_addr
+
+        if not os.path.isfile(self.disk):
+            raise AnsibleError("%s is not a file, aborting" % self.disk)
+
+        self.guestfs = guestfs.GuestFS(python_return_dict=True)
+
+        self.guestfs.add_drive_opts(self.disk)
+
+    def _find_python(self):
+        ''' find what python to use '''
+        self._python = None
+        for p in ('python', 'python3'):
+            try:
+                python_version = self.guestfs.sh('%s -V' % p)
+                self._python = p
+                break
+            except RuntimeError:
+                pass
+
+        if self._python is None:
+            raise AnsibleError("No python found on the image, aborting")
+
+    def _mount_tmp(self):
+        ''' mount a tmpfs over /tmp or /run to store various artefacts '''
+        self._tmp = None
+        for tmp in ('/run', '/tmp'):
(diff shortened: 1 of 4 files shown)


lib/ansible/plugins/connection/guestfs.py
```diff
@@ -0,0 +1,209 @@
...
+        self.disk = self._play_context.remote_addr
+
+        if not os.path.isfile(self.disk):
+            raise AnsibleError("%s is not a file, aborting" % self.disk)
+
+        self.guestfs = guestfs.GuestFS(python_return_dict=True)
+
+        self.guestfs.add_drive_opts(self.disk)
+
+    def _find_python(self):
```
jimi-c, line 95: Shouldn't this use the value of ansible_python_interpreter? There is a mechanism (winrm.py uses it) to pass in some host variables to connection plugins, so this should probably implement that.

mattclay:
@mscherer: Could you add this to the integration tests? Specifically, add a section to test_connection.inventory for guestfs and then make sure the tests pass by running:

```
TEST_FLAGS='-l guestfs' make test_connection
```

alikins:
Seems to work for me, I was able to point it to a qcow2 image and run ansible playbook against it. I only tested fedora images, so didn't try windows/ntfs/etc.

> For now, it only support local VM with single OS, but it should be trivial
> to add support for remote disks and multiple OS.

Would this include specifying a (local) libvirt/qemu url? ie, user vs system ?

lib/ansible/plugins/connection/guestfs.py
mscherer, line 95: I didn't knew about set_host_overrides, I will take a look.

lib/ansible/plugins/connection/guestfs.py
```diff
@@ -0,0 +1,209 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import os
+import os.path
+import guestfs
```
pilou-: You should raise `AnsibleError` when `guestfs` python module isn't available.

mscherer: Good point, fixed in the pushed branch

mscherer:
@mattclay I added a first attempt at the test, I am not sure how to run them the correct way (and I rather avoid running disruptive test on my work laptop). It requires to download a cloud image to test, so I guess I should commit that file in git :)

mscherer:
@alikins Yeah, I think so. It should be able to do all that you can do with guestfs command line.

So we can use http://libguestfs.org/guestfish.1.html#adding-remote-storage for remote disks, and or also a different libvirt url if needed. Of course, people need to make sure that remote VM are shutdown for obvious concurency issue.

I guess we could pass ansible_libvirt_uri variable or something, provided the connexion plugin can access that information (but I didn't dig that direction yet).

mscherer:
Ok so we could pass more variable using the set_host_overrides system that @jimi-c spoke earlier, but I rather add the features later.

lib/ansible/plugins/connection/guestfs.py
mscherer, line 95: So, we need autodetection for Fedora and Ubuntu (since they come with python 3 pre installed, so I need to detect them for running raw to install python2). Now, I am not sure how should the plugin react if the user give a explicit python_interpreter, and if we can assume to be installed if given (especially for something like raw module, which should work without python).

In fact, I wonder if I shouldn't also have a wrapper in shell, since you can't use it on freebsd image with the current system of using a python wrapper if python is not installed (which is the case on freebsd among others). And I might need it soon for netbsd images.

mattclay:
@mscherer No need to include the image in git, just instructions on what is necessary to run the test, as it appears you've already done. If you can run `TEST_FLAGS='-l guestfs' make test_connection` and have the tests pass, that's all that is needed for testing.

mscherer:
So I did started to look at it (after Robyn talk on ansible container), and I did it a issue, and I think that's because we have test using UTF 8 string, but the Fedora cloud image I have do not seems to support utf-8.

I will try to dig a bit, but if test requires a less featurefull image than the usual image for testing utf8, that might be a issue.

mscherer:
Ok so I fixed the first issue regarding utf8, now I hit another issue (so still trying to get it merged)

mscherer:
So I fixed the issue I had, and the tests are working (and are documented) on my laptop on RHEL 7.

mattclay:
@mscherer Now that we're using `ansible-test` to run the tests, you need to add an empty aliases file here: `test/integration/targets/connection_guestfs/aliases`

Once this is merged I can look at updating our CI process to support running the test on Shippable. In the meantime, having an empty aliases file will cause it to be skipped for CI purposes.

mattclay (changes requested):
Once the `aliases` file is added, the tests should pass.

mattclay (approved):

lib/ansible/plugins/connection/guestfs.py
bcoca: connection plugins should not have their own python logic. If distros are using different python/missing python it is up to user to add (plenty of examples with raw and ansible_python_interpreter).

lib/ansible/plugins/connection/guestfs.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def _cmd_name(self):
+        ''' return the filename for the command to execute '''
+        return '%s/cmd' % self._tmp
+
+    def _cmd_result_name(self):
+        ''' return the filename for the command output '''
+        return '%s/cmd.out' % self._tmp
+
+    def exec_command(self, cmd, in_data=None, sudoable=False):
```
bcoca, line 177: if not handling become methods update the class variable to indicate this

bcoca (changes requested):

lib/ansible/plugins/connection/guestfs.py
mscherer: The python detection is used by the connexion plugin for internal matter, ie be able to run the script whose code is in the EXECUTION_SCRIPT constant, which is a wrapper that do encode stdin/stdout as json, because guestfish do not return the 2 streams separate. 

This is separate and not related to the fact that ansible also use python. If perl was installed by default everywhere, I would use it, or if I had a good reason to believe that I could write a portable shell script doing the same without losing sanity points, I would do it.

I guess I could replace that part by a script that do execute the binary, dump stdout and stderr in 2 separate file and then get them, but this look more ugly than the current way (albeit more portable). And if I remember correctly, the python detection is needed so I can run raw or the various shell fragment in the first place. I can take a look again, because that code is not so fresh, but I am confident that if I could have avoided that, I would.

bcoca: this is why we have shell plugins .... but I do agree on the sanity part ...

lib/ansible/plugins/connection/guestfs.py
mscherer: I am not sure to understand. I suspect I need to change sudoable to True rather than False, becuase that seems to be the default everywhere else, but what class variable do I need to update ?

bcoca: not that, i'm talking about `become_methods` list class variable.

bcoca: example that supports all defaults except 'su'
```
 become_methods = frozenset(C.BECOME_METHODS).difference(('su',))
```

mscherer: I would need to test if sudo and su work I guess, as I didn't need them (but I can see why it would be required, at least 'su').

bcoca: su requires a tty, that is why most 'subprocess' plugins don't support it.

gundalow:
@mscherer 
Thank you for your PR.
As part of reviewing the backlog of PRs we are looking at PRs older PRs that haven't been updated in a while

Given that:
* PR has needed a rebase since April June 2017

Therefore I'm going to close this.

If you or anyone else wants to continue with this work then please do feel free to create a fresh PR and `@mention` the previous reviewers in here.

Issue #13614, opened 2015-12-20
Ansible 2.0 task variable precedence is lower than in documentation


ThomasSteinbach:
The task variable `remote_user` becomes overridden by the inventory variable `ansible_user`.

jimi-c:
Hi @ThomasSteinbach, this is a case of mis-documentation I think. The difference here is that these variables are considered "magic" variables in the documentation, and as such have a higher precedence than the task despite the fact that they come from inventory. I agree this is a confusing topic, which is not explained well at all in the docs, so we'll look to fix this there rather than changing code.

ThomasSteinbach:
Ah, ok @jimi-c. So these 'magic' variables are just a set of variables with higher precedence in general. But with reference to issue #13388 this leads me to the following problem: The basis is that I create a docker container, add it dynamically to the inventory and define the default user to connect to the container as follows:

```
- name: create jenkins container
  docker:
    name: aci_jenkins
    image: iteratechh/jenkins
    state: running

- name: add aci_jenkins to inventory
  add_host:
    name: aci_jenkins
    ansible_connection: docker
    ansible_user: jenkins
```

But now I have one task I want to execute within the container not as the _jenkins_ user but as _root_ user. I don't want to use become therefore, because setting the docker container user is the better way. So @abadger adviced me in a discussion in the PR #13424 to set the `remote_user` per task level, as follows:

```
- name: install git
  delegate_to: aci_jenkins
  remote_user: root
  apt:
    name: git
    state: present
```

But with the current implementation setting the `remote_user` per task level is not possible. Even if the `remote_user` inside the source code is in the array of variables meant to be changed per task level.

So what is the advantage of the 'magic' state of those variables? And how can I set the remote user for a specific task?

bcoca:
In 2 you can set vars at the task level

```
- name: install git
  delegate_to: aci_jenkins
  vars:
     ansible_user: root
  apt:
    name: git
    state: present
```

ThomasSteinbach:
Thank you @bcoca for the solution.

bcoca:
I'm going to close this ticket, as `remote_user` is NOT a variable, its a directive that gets overridden by the connection variables `ansible_user/ansible_ssh_user`. This is not a variable precedence issue, though the confusion is understandable.

PR #13615, opened 2015-12-20
fixed task variables precedence

ThomasSteinbach:
This PR would fix issue #13614 by simply moving the code block, applying the task vars, to the lines of code that otherwise would override the task vars.

### DIFF
--- a/lib/ansible/playbook/play_context.py
+++ b/lib/ansible/playbook/play_context.py
@@ -297,14 +297,6 @@ class PlayContext(Base):
 
         new_info = self.copy()
 
-        # loop through a subset of attributes on the task object and set
-        # connection fields based on their values
-        for attr in TASK_ATTRIBUTE_OVERRIDES:
-            if hasattr(task, attr):
-                attr_val = getattr(task, attr)
-                if attr_val is not None:
-                    setattr(new_info, attr, attr_val)
-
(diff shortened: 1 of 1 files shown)


ThomasSteinbach:
Ok, I see that the [test_play_context.py](https://github.com/ansible/ansible/blob/devel/test/units/playbook/test_play_context.py) check will fail, because it calls the function

```
play_context.set_task_and_variable_override(task=mock_task, variables=all_vars, templar=mock_templar)
```

and expects the variables from the _mock_task_ becomes overridden by the ones from _all_vars_. I didn't exactly know where the _all_vars_ come from in practice. But nevertheless in the Ansible documentation is stated, that [task variables have the second height precedence](http://docs.ansible.com/ansible/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable) right after the _extra-vars_.

And in practice this would be relevant for me, because as stated in issue #13614 I would override the `ansible_user` from the inventory with the `become_user` at a couple of tasks. In the inventory I say, that the connector should generally login as `ansible_user` to the remote host but within a couple of tasks I want to override that user temporary.

jimi-c:
Per my comment here, https://github.com/ansible/ansible/issues/13614#issuecomment-166338249 we're going to close this PR in favor of fixing this in the documentation.

If  you have any further questions, please let us know by stopping by one of the two mailing lists, as appropriate:

Thank you!

Issue #13515, opened 2015-12-11
Doc for GCE module refer to outdated auth strategy


rorygibson:
The docs recommend using GCE credentials in P12 format (and converting to PEM).

However, running the module, libcloud informs you that this is deprecated and you should download & use credentials in JSON format instead.
(Upon trying this, it does work!)

bcoca:
cc @erjohnso

erjohnso:
Thanks for flagging that @rorygibson and bringing it to my attention @bcoca.

Were you planning on sending a pull-request to improve them, or just wanted to file the issue? If you weren't planning on doing the work yourself, I may have a colleague that's interested in cleaning up the docs and making them current. Otherwise, I'll get to it in a few weeks.

viglesiasce:
Thanks for bringing this up @rorygibson! I noticed this as well last week and will take a pass at bringing these docs up to date.

rorygibson:
Thanks guys.
I'm very much an Ansible newbie. Enjoying the experience so far and the
docs have generally been excellent.

Issue #13594, opened 2015-12-18
GCE dynamic inventory script is not compatible with apache-libcloud (v0.20.0)


d3rzky:
Hi,

I tried generating dynamic inventory using properly configured (secrets.py) gce.py script and got next issue while running it:

> Please Go to the following URL and sign in:
> https://accounts.google.com/o/oauth2/auth?BUNCH_OF_PARAMETERS
> Enter Code: 

Following the link I got 

> Error: redirect_uri_mismatch
> ..
> The redirect URI in the request: urn:ietf:wg:oauth:2.0:oob can only be used by a Client ID for native application. It is not allowed for the &#39;WEB&#39; client type. You can create a Client ID for native application in the Credentials section of the Google Developers Console.

Everything works fine if apache-libcloud is downgraded to 0.19.0 version.

abadger:
@erjohnso Looks like one you might want to look into.

viglesiasce:
I was able to reproduce this as well when developing https://github.com/ansible/ansible/pull/13623. This also effects the GCE modules themselves. I'll take a look at this in libcloud.

viglesiasce:
This should be fixed in the next rev of libcloud:
https://issues.apache.org/jira/browse/LIBCLOUD-785

erjohnso:
Yes, this will impact users with new service accounts. Any existing service accounts will be fine, but we'll likely start to see this more and more until the next version of libcloud makes it out.

ghost:
Damn, I just stumbled upon this. I have tried libcloud 0.20.1 and 1.0.0-pre1 but no luck with neither of them :(

alex-leonhardt:
updated to latest version of apache libcloud (and ansible) and at least i'm no longer getting the same error message.. 

```
apache-libcloud==0.20.1
```

and fwiw, with correct time (local host's time) & cred settings, all seems to work fine now ...

supertom:
This works for me with current ansible and apache_libcloud==0.20.1 as well.  I recommend we close.  OK with you, @ryansb ?

ryansb:
Yup, this works for me.

PR #13623, opened 2015-12-21
Update GCE module to use JSON credentials

viglesiasce:
This fixes issue https://github.com/ansible/ansible/issues/13515 in a backwards compatible way.

GCP now recommends using JSON credentials files rather than P12 files. Using JSON files also simplifies the onboarding process for Ansible as it no longer requires openssl-foo in order to get started.

This PR updates includes the following:
- Bumps the required libcloud version to 0.17.0 which allows for JSON credentials
- Adds a parameter `credentials_file` that supplants `pem_file`
- Adds documentation for configuring via environment variables
- Adds documentation for using GCE service accounts rather than JSON files
- Updates the examples to use the new `credentials_file` parameter

This PR has a counterpart in the ansible-modules-core repository that enables this functionality in the modules themselves. That one should be merged first or at the same time.

cc @erjohnso

### DIFF
--- a/lib/ansible/module_utils/gce.py
+++ b/lib/ansible/module_utils/gce.py
@@ -39,6 +39,7 @@ USER_AGENT_VERSION="v1"
 def gce_connect(module, provider=None):
     """Return a Google Cloud Engine connection."""
     service_account_email = module.params.get('service_account_email', None)
+    credentials_file = module.params.get('credentials_file', None)
     pem_file = module.params.get('pem_file', None)
     project_id = module.params.get('project_id', None)
 
@@ -50,6 +51,8 @@ def gce_connect(module, provider=None):
         project_id = os.environ.get('GCE_PROJECT', None)
     if not pem_file:
         pem_file = os.environ.get('GCE_PEM_FILE_PATH', None)
+    if not credentials_file:
+        credentials_file = os.environ.get('GCE_CREDENTIALS_FILE_PATH', pem_file)
 
     # If we still don't have one or more of our credentials, attempt to
     # get the remaining values from the libcloud secrets file.
@@ -62,15 +65,15 @@ def gce_connect(module, provider=None):
         if hasattr(secrets, 'GCE_PARAMS'):
             if not service_account_email:
                 service_account_email = secrets.GCE_PARAMS[0]
-            if not pem_file:
-                pem_file = secrets.GCE_PARAMS[1]
+            if not credentials_file:
+                credentials_file = secrets.GCE_PARAMS[1]
         keyword_params = getattr(secrets, 'GCE_KEYWORD_PARAMS', {})
         if not project_id:
             project_id = keyword_params.get('project', None)
 
     # If we *still* don't have the credentials we need, then it's time to
     # just fail out.
-    if service_account_email is None or pem_file is None or project_id is None:
+    if service_account_email is None or credentials_file is None or project_id is None:
         module.fail_json(msg='Missing GCE connection parameters in libcloud '
                              'secrets file.')
         return None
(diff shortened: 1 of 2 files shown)


viglesiasce:
Tested on libcloud 0.19.0 as 0.20.0 is hitting the following: https://github.com/ansible/ansible/issues/13594

bcoca:
cc @erjohnso

bcoca:
the change to require libcloud 0.17 and higher is mandated by the addition of the credentials_file or is this caused by the google cloud api itself?

As in previous changes, we try to keep this backwards compatible and not force users to upgrade (many times they cannot).

viglesiasce:
Using the old way will still work. In the [other PR](https://github.com/ansible/ansible-modules-core/issues/2715) we said "apache-libcloud >= 0.13.3, >= 0.17.0 if using JSON credentials".

This PR will also bail out if you are using the new JSON creds with older libcloud:
https://github.com/ansible/ansible/pull/13623/files#diff-b6bc17b51c779b2ac4eefcfc5504a83bR91

What would you like changed? I'll patch it up ASAP.

erjohnso:
Libcloud 0.17.0 has been out for a year [[blog](https://libcloud.apache.org/blog/2015/02/18/libcloud-0-17-0-released.html)]. Is there some kind of timeline for supporting old dependency versions that we should stick to?

viglesiasce:
@bcoca bump. Anything else I need to do before this can get merged?

bcoca:
@erjohnso sadly for a lot of people in 'enterprise environments' it is extremely hard for them to update/upgrade. I would say this depends a lot on the tech and it's penetration and usage, I would prefer to keep backwards compatibility as much as possible.

viglesiasce:
Hey @bcoca,
What's left to do here? This PR is backwards compatible, we understand the importance of that quality.

bcoca:
@viglesiasce that is not clear from the 'new docs'

"The GCE modules all require the apache-libcloud module version 0.17.0 or higher,"

michaelbannister:
I just tried to use this and it appears that the docs update has been merged but the actual module hasn't?!

viglesiasce:
@bcoca please take a look. I softened the language and fixed merged conflicts. Sorry for the delay.

abadger:
schwicht from irc reports that the gce modules are currently broken in git (because the changes in ansible-modules-core have already been merged) but this one has not.  We need to decide whether to merge this or revert the module changes for 2.1rc2.

bcoca:
this has merge commit, needs to be cleaned before we can merge

lib/ansible/module_utils/gce.py
```diff
@@ -27,9 +27,13 @@
 # USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #
 
+import json
 import os
 import traceback
 
+from distutils.version import LooseVersion
+
+import libcloud
```
bcoca: this should be try/except wrapped to avoid import exceptions on modules, see ec2.py for examples

viglesiasce:
Hey @bcoca,

This should be good to go. Can you rerun the tests? Based on previous runs it looks like that Fedora 23 test is particularly flaky.

jimi-c:
@viglesiasce sorry this looks good, however it didn't make the cut for 2.1 so we'll get it in for 2.2.

jimi-c:
Scratch that, we're re-reviewing now.

abadger:
yay!  merged.  Note that bcoca's comment about the import of libcloud wasn't fixed.  I'll go ahead and fix that now that the code is in.

abadger:
This will be in devel and the 2.1 release.

abadger:
For completeness, here's the commit to implement bcoca's comments: 4d59779e0adf35b8710f20d966c527ea7efab466

viglesiasce:
Thanks so much folks!!!!!!!

rubal033:
I am using ansible-2.1 and getting ""msg": "Missing GCE connection parameters in libcloud secrets file." with "gce" module while trying to create the instance. I am running ansible from gce instance only which also have suitable Google Service Account attached. 

Could anyone you guide me with its workaround or if i am doing something wrong?

PR #13625, opened 2015-12-21
Remove new_inventory, since that's unused

mscherer:
The class have been either moved or reimplemented
elsewhere.


abadger:
We want to remove this in stable-2.0 but want it in devel.  Some people have been waiting on 2.1 development to start working on this and then have it replace the code in inventory.

bcoca:
already removed via 6a0d2116b8179399117ee58b783610e1ff6429d9

Issue #5000, opened 2013-11-21
No way to indirectly lookup vars: facts


rektide:
If I `set_fact: foo=bar var=foo` (or include_vars), I can latter lookup whatever var is named in var by doing: `{{hostvars[inventory_hostname][var]}}`. However, if foo was set in vars:, there is no way to indirectly look it up: it exists at a higher level of fact-iness than hostvars do.

Please provide some means to indirectly resolve facts set in `vars` headers of a playbook.

mpdehaan:
This seems like a more involved user question about use cases rather than a bug report, can you please stop by the mailing list?

rektide:
`{{vars[var]}}` allows access to vars. This top level variable is defined here: https://github.com/ansible/ansible/blob/512ebdb9713744d1690ea7095ff8ee80ef1c713e/lib/ansible/runner/__init__.py#L435

Issue #13584, opened 2015-12-17
with_items variable name template broken (worked in 1.9.4)


robinro:
##### Ansible Version:

ansible 2.0.0 (f41dd578b340b76aa706af9540fd8fab2fbb0eb6)
##### Ansible Configuration:

self-packaged version based on above revision
##### Environment:

openSuSE Leap 42.1
##### Summary:

Substituting variable names based on other variables does not work in ansible 2.0.0 but used to work in 1.9.4.

This should either be fixed or stated more prominently in the Changelog as it breaks existing playbooks.
##### Steps To Reproduce:

Example playbook:

```
vars:
  packages_install_42:
    - bash
    - zsh

tasks:
- name: install stuff
  zypper: name={{item}} state=present
  with_items: packages_install_{{ansible_distribution_major_version}}
```
##### Expected Results:

Variable packages_install_42 is used, the packages bash and zsh are installed.
##### Actual Results:

The task fails:
`(item=[u'packages_install_42']) => {"failed": true, "item": ["packages_install_42"], "msg": "No provider of 'packages_install_42' found.\n", "rc": 104}`

bcoca:
I don't think that is supposed to work, this is how it should be written:
`with_items: "{{ hostvars[inventory_hostname]['packages_install_' + ansible_distribution_major_version] }}"`

robinro:
**Edit**: Ignore this comment, the case there is different (not a variable)

~~It is/was suppossed to work. At least it is in the docs: 
http://docs.ansible.com/ansible/playbooks_loops.html#finding-first-matched-files~~

rektide:
I too used this old syntax. I'm fine with @bcoca's suggestion, but I can't get it working. Here's a dumb test playbook, which shows that variables in var: are not found in either hostvars[inventory_hostname] or vars;

```
---
- hosts: all
  gather_facts: False
  vars:
    LIST:
    - one
    - two
  tasks:
   - debug:
     args:
       msg: "{{ hostvars[inventory_hostname] }}"
   - debug:
     args:
       msg: "{{ vars }}"
   - debug:
     args:
       msg: "{{ hostvars[inventory_hostname]['LIST'] }}"
```

(Same result without gather_facts, it just makes output easier to read.)

I adopted this syntax because of #5000, which is where I originally found that vars: were not found in hostvars or vars. I'm fine with the syntax going away, but now as before in #5000, there's no way to indirectly reference vars from vars:, as far as I can tell.

[...]

PR #13626, opened 2015-12-22
move hostvars.vars to vars

bcoca:
this fixes duplication under hostvars and exposes all vars in the vars dict
which makes dynamic reference possible on 'non hostvars'

this fixes a regression from the fix in #5000

### DIFF
--- a/lib/ansible/vars/__init__.py
+++ b/lib/ansible/vars/__init__.py
@@ -259,8 +259,6 @@ class VariableManager:
             except KeyError:
                 pass
 
-        all_vars['vars'] = all_vars.copy()
-
         if play:
             all_vars = combine_vars(all_vars, play.get_vars())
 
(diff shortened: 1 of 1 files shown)


bcoca:
test case:

```
---
- hosts: localhost
  gather_facts: False
  vars:
    LIST:
    - one
    - two
  tasks:
   - debug:
     args:
       msg: "{{ hostvars[inventory_hostname] }}"
   - debug:
     args:
       msg: "{{ vars }}"
   - debug:
     args:
       msg: "{{ vars['LIST'] }}"
```

Before patch, the last debug fails, also if you change it to `hostvars[inventory_hostname]['LIST']` so the LIST variable was never accessible indirectly. Also hostvars.vars replicated vars in hostvars, while vars ONLY had vars from hostvars and none from play or task.

rektide:
This indeed fixes #5000, & makes it possible to work around the syntax changes mentioned in #13584. Thanks thanks thanks!

PR #13629, opened 2015-12-22
When var name is the same as var content, try to template it before reporting that var is not defined.

Yannig:
Add asserts in test_var_blending to check this special corner case.

### DIFF
--- a/lib/ansible/plugins/action/debug.py
+++ b/lib/ansible/plugins/action/debug.py
@@ -54,6 +54,9 @@ class ActionModule(ActionBase):
                 try:
                     results = self._templar.template(self._task.args['var'], convert_bare=True, fail_on_undefined=True, bare_deprecated=False)
                     if results == self._task.args['var']:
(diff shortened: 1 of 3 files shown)


Yannig:
The tests are broken due to a bug in make tests-py3 (https://github.com/ansible/ansible/issues/13638)

Yannig:
@jimi-c any comment?

jimi-c:
@Yannig I've seen this, just lower priority at the moment. It looks fine, the only thing I'd like to see cleaned up is that it's 4 commits. If you want to squash it down, that'd be helpful, otherwise I'll just do it when we merge it.

Yannig:
@jimi-c squash down into a single commit against last devel.

abadger:
Merged to devel

abadger:
And cherry-picked to stable-2.1 as well.  Thanks @Yannig !

Yannig:
@abadger thanks :)

PR #13632, opened 2015-12-22
Initial support for Windows Hosts

kpaleniu:
Ansible is an awesome tool. However, since it does not currently run on Windows without extreme pain, I decided to start working on Windows host support. This pull-request is by no means a complete support fixing every single Windows-related issue.

Module packaging is now changed to take advantage of built-in .exe generation in setuptools. To accomodate that, scripts under `bin/` have been refactored into `ansible.cli` module itself. Each module under `ansible.cli` now provides an entry point. To reduce repetition, common exception handling is gathered in `ansible.cli.common` using the so-called Lippincott function pattern.

Rest of the changes are as follows.
- Now there is an env-setup for Powershell.
- Various pieces of code was tweaked to lazy-load Unix-specific modules until they are absolutely needed.
- Implemented Windows-specific implementations for functionality, such as display, where lazy-loading did not help.

With this pull request the following can be done successfully on Windows.

```
PS > cd path\to\ansible
PS > pip install .
PS > ansible
```


kpaleniu:
The env-setup scripts broke in ways I did not anticipate. I'm working on fixing those.

bcoca:
There are major issues that prevent us from running on Windows, as an example I'll use the lack of the POSIX `fork` call.

Ansible relies on this very heavily, Python subprocess creation behaves very differently on Windows and requires a completely different design. This change also makes a huge difference in performance, we would go from `Copy on Write` and implied inheritance of resources to an initial copy and explicit passing of resources. Aside from the huge increase in memory use this requires a huge code rewrite to make sure the 'children' get all the resources they need.

There are many more things that just won't work that Ansible relies on, these are the reasons we have no inclination to even approach a full Windows port of the Ansible controller.

kpaleniu:
Okey, clearly much more deeper issue than I realized. I also realize that the core developer team is extremely busy on other, more pressing issues and features.

Just to clarify, assuming me or someone else works on Windows support on their own and manages to address issues, such as the `fork` and `COW` in decent manner, are patches still welcome? Or will it be futile effort?

No matter what decision, I think it'll be best to close this PR since this feature clearly requires much, much more thought and elbow-grease. Merging this PR would probably just open the floodgates of complaints of "why X doesn't work windows?", which I assume you're not very keen to handle.

bcoca:
We are not against a PR that would solve most issues of running on Windows, but we would be against it if it creates major issues and degradation for those on Unix/Linux.

I just posed above one of the big problems, there are many more, file system semantics being a very big one also, Ansible takes advantage of many POSIX properties that will flat out fail on Windows. Many of the Unix tools and system calls relied on would need to be ported or adjusted, in many cases (like fork) there is no good equivalent.  

The Cygwin project has done a great job of porting a quasi Unix environment on top of Windows, but even they still cannot provide a system Ansible can use reliably, which is why we don't support it. There are people have been using Ansible on it in a limited capacity, but they will hit many errors or corner cases that require enormous effort to fix.

kpaleniu:
Thank you, that clarifies enough. I'll close this PR. If I ever manage to develop a workable approach, I'll try submitting a new PR.

PR #13636, opened 2015-12-22
Force success_key value against play_context and connection play_context

### LINKED ISSUE #13624: Ansible 2.x: local connector misbehaves when using with_items
Ansible crashes when creating links with_items and using `ansible_connection=local`. Here an example playbook for reproducing the behavior:

```
- name: the last task will throw an error
  hosts: localhost
  gather_facts: no

  tasks:
    - name: create files - that will work
      copy:
        content: "test"
        dest: "/tmp/{{ item }}"
      with_items:
        - testfile1
        - testfile2

    - name: create links - that will crash
      become: yes
      file:
        src: "/tmp/{{ item }}"
        dest: "/tmp/{{ item }}link"
        state: link
      with_items:
        - testfile1
        - testfile2
```

The output is:

```
PLAY [the last task will throw an error] ***********************************

TASK [create files - that will work] *************************************************************
ok: [localhost] => (item=testfile1)
ok: [localhost] => (item=testfile2)

TASK [create links - that will crash] *************************************************************
fatal: [localhost]: FAILED! => {"failed": true, "msg": "ERROR! privilege output closed while waiting for password prompt:\nBECOME-SUCCESS-smnurvzoeqwqcapbnmmyrhlaautxdumn\n{\"src\": \"/tmp/testfile2\", \"group\": \"root\", \"uid\": 0, \"dest\": \"/tmp/testfile2link\", \"changed\": false, \"owner\": \"root\", \"state\": \"link\", \"gid\": 0, \"mode\": \"0777\", \"invocation\": {\"directory_mode\": null, \"force\": false, \"remote_src\": null, \"path\": \"/tmp/testfile2link\", \"owner\": null, \"follow\": false, \"group\": null, \"state\": \"link\", \"content\": null, \"serole\": null, \"diff_peek\": null, \"setype\": null, \"dest\": \"/tmp/testfile2link\", \"selevel\": null, \"original_basename\": null, \"regexp\": null, \"validate\": null, \"src\": \"/tmp/testfile2\", \"seuser\": null, \"recurse\": false, \"delimiter\": null, \"mode\": null, \"backup\": null}, \"size\": 14}\n"}

```

I think the local connector misbehaves with more then just the _file_ module, when using `with_items`. I have another issue open ([pip module does not work with_items in Ansible 2.0 #2552](https://github.com/ansible/ansible-modules-core/issues/2552)) where i suspected the _pip_ module having problems `with_items`. However, the general error remains the same in both issues:

```
ERROR! privilege output closed while waiting for password prompt
```

During the investigation of the other issue, I found out that the error for it was introduced with commit https://github.com/ansible/ansible/commit/53e86f3130145ce40606cec32aef2858756a20ab, which was merged with PR https://github.com/ansible/ansible/pull/13261. I haven't checked if it is the same for this issue, but it could be a first trace.

Yannig:
Do not look for sudo prompt on second item run (using with_items for exemple) for the same task.

Fix https://github.com/ansible/ansible/issues/13624

### DIFF
--- a/lib/ansible/plugins/action/__init__.py
+++ b/lib/ansible/plugins/action/__init__.py
@@ -505,6 +505,8 @@ class ActionBase(with_metaclass(ABCMeta, object)):
         if sudoable and self._play_context.become and (allow_same_user or not same_user):
             display.debug("_low_level_execute_command(): using become for this command")
             cmd = self._play_context.make_become_cmd(cmd, executable=executable)
+            # Updating connection context to share same success_key
+            self._connection._play_context.success_key = self._play_context.success_key
 
         display.debug("_low_level_execute_command(): executing: %s" % (cmd,))
         rc, stdout, stderr = self._connection.exec_command(cmd, in_data=in_data, sudoable=sudoable)
--- a/lib/ansible/plugins/connection/__init__.py
+++ b/lib/ansible/plugins/connection/__init__.py
@@ -206,7 +206,10 @@ class ConnectionBase(with_metaclass(ABCMeta, object)):
         pass
 
     def check_become_success(self, output):
-        return self._play_context.success_key == output.rstrip()
+        for line in output.splitlines(True):
+            if self._play_context.success_key == line.rstrip():
+                return True
+        return False
 
     def check_password_prompt(self, output):
         if self._play_context.prompt is None:


Yannig:
The tests are broken due to a bug in make tests-py3 (https://github.com/ansible/ansible/issues/13638)

jimi-c:
@Yannig a couple of things:

1) I don't think the flag value should be a class variable, but should be an instance variable instead.
2) It looks like you've brought in one of @abadger's patches, which fixes the py3 bug, but PRs shouldn't include patches from others, so this needs to be rebased on that.

Yannig:
@jimi-c done and happy new years from France!

jimi-c:
Looks good to me, cc'ing @amenonsen as he wrote that code originally.

jimi-c:
And Happy New Years to you too!

amenonsen:
@jimi-c no, I didn't touch local.py, only ssh.py. (But happy new year to all of you anyway! :-)

jimi-c:
Doh, thought this was about ssh.py, wasn't paying attention...

bcoca:
this should not be needed as `self.check_become_success(become_output)` should see the `echo successkey` string and not wait for the sudo prompt

Yannig:
@bcoca Agree but if you take a look at https://github.com/ansible/ansible/issues/13624, you'll see that - with items or something similar - Ansible crash saying that he fail to become the become_user (**ERROR! privilege output closed while waiting for password prompt:**).

If you launch the same playbook without sudo password (-K options), Ansible succeed in launching the playbook. The PR here fix this particular issue.  Maybe there's a better way. Maybe a bug in `self.check_become_success()` need to be fixed ...

Yannig:
Here is the playbook I'm using for doing my tests: https://github.com/Yannig/ansible-issue-low-speed/blob/master/local-connector/local-connector.yml

And here the command line I'm using:

`ansible-playbook local-connector/local-connector.yml -K`

Yannig:
Digging into play_context, I saw 2 differents keys generated for the sudo prompt (success_key) but Ansible is mixing up the keys and expect the wrong success_key.

Yannig:
@jimi-c this time, no more boolean, just using the same condition as in ssh.py to call check_become_success.

Yannig:
@bcoca what do you think about this last version?

Yannig:
@amenonsen The last commit may fix the remaining issue.

I was wondering if you are also running Ansible on a MacOS.

bcoca:
@Yannig if this fixes it for you, I think we have a deeper issue with the forks giving out incorrect play_context to the plugins that should be in lockstep.

ktaragorn:
@bcoca the problem is that the `connection` and with that the `play_context` is reused on subsequent runs, that is why it is out of sync

Yannig:
@bcoca @ktaragorn looking after this problem, Ansible have the following behaviour:
- create new connection + play_context object each time when using ssh + with_items
- create only one connection but multiple play_context when using local connector + with_items

Well, I don't know which one is wrong and I'm wondering if it maybe a better idea to have the same behaviour for every connection objects.

Yannig:
@bcoca I have a different way to get this issue fixed:

``` patch
index 4a2d30a..e59860e 100644
@@ -517,7 +517,7 @@ class TaskExecutor:
             'normal',
             task=async_task,
             connection=self._connection,
-            play_context=self._play_context,
+            play_context=self._connection._play_context,
             loader=self._loader,
             templar=templar,
             shared_loader_obj=self._shared_loader_obj,
@@ -590,7 +590,7 @@ class TaskExecutor:
                 'normal',
                 task=self._task,
                 connection=ssh_connection,
-                play_context=self._play_context,
+                play_context=ssh_connection._play_context,
                 loader=self._loader,
                 templar=templar,
                 shared_loader_obj=self._shared_loader_obj,
@@ -635,7 +635,7 @@ class TaskExecutor:
             handler_name,
             task=self._task,
             connection=connection,
-            play_context=self._play_context,
+            play_context=connection._play_context,
             loader=self._loader,
             templar=templar,
             shared_loader_obj=self._shared_loader_obj,
```

Basicaly, we use the same play context object as in connection object.

It fix local connector and do not break other things.

ktaragorn:
I think we need @cchurch  to weigh in on his motivations on making that change, and on the right way to fix this issue. The `_play_context` always used go be refreshed every iteration before that commit, so the change you suggest could look like a bug in that context @Yannig, depends on the actual utility of making it change or not.

Issue #13638, opened 2015-12-22
Bug Report: make tests-py3 is broken in devel


Yannig:
##### Ansible Version:

last devel
##### Ansible Configuration:

Nothing special
##### Environment:

Nothing special
##### Summary:

make tests-py3 is broken
##### Steps To Reproduce:

Launch `make tests-py3`
##### Expected Results:

Tests OK
##### Actual Results:

```
======================================================================
FAIL: test_sudo_only_if_user_differs (units.plugins.action.test_action.TestActionBase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/ansible/ansible/test/units/plugins/action/test_action.py", line 52, in test_sudo_only_if_user_differs
    play_context.make_become_cmd.assert_called_once_with('ECHO', executable=None)
  File "/opt/python/3.4.2/lib/python3.4/unittest/mock.py", line 782, in assert_called_once_with
    return self.assert_called_with(*args, **kwargs)
  File "/opt/python/3.4.2/lib/python3.4/unittest/mock.py", line 771, in assert_called_with
    raise AssertionError(_error_message()) from cause
nose.proxy.AssertionError: Expected call: mock('ECHO', executable=None)
Actual call: mock(b'ECHO', executable=None)
    None = (('ECHO',), {'executable': None}) if isinstance((('ECHO',), {'executable': None}), Exception) else None
>>  raise AssertionError(_error_message()) from None
```

Yannig:
The following commit broke the tests in python3: https://github.com/ansible/ansible/commit/0c013f592a31c06baac7aadf27d23598f6abe931

PR #13640, opened 2015-12-22
Fix make tests-py3 on devel.

Yannig:
Fix for https://github.com/ansible/ansible/issues/13638


abadger:
Thanks -- we probably want to pass in native strings here but expect byte strings out.  At the moment, the design has _low_level_execute_command() be the responsible party for translating between text and byte strings.

abadger:
@Yannig I'll go ahead and merge and make that change... I want to get this in so we're sure to see any other unittest failures on other commits in travis ;-)

Yannig:
@abadger Thanks :)

PR #13647, opened 2015-12-23
corrected role path search order

### LINKED ISSUE #13645: Role search path precedence change in v2
In 1.9 the path precedence appears to be:
1. playbook_dir/roles/
2. playbook_dir/

In 2.0 the path precedence has changed to what appears to be
1. playbook_dir/
2. playbook_dir/roles/

To validate this, the file tree looks like:

```
.
├── ansible.cfg
├── roles
│   └── test
│       └── tasks
│           └── main.yml
├── test
│   └── tasks
│       └── main.yml
└── test.yml
```

ansible.cfg in this case is completely empty.
#### test.yml

```

---
- hosts: all
  gather_facts: false
  roles:
    - test
```

Each `main.yml` should contain a single debug, each with different message such as:

```
- debug: msg=in_roles_dir
```

```
- debug: msg=in_playbook_dir
```

This causes a completely non-role related directory in the `playbook_dir` to potentially, override an actual role of the same name in the `roles` directory.

bcoca:
the unfraking was matching roles in current dir as it always returns a full path,
pushed to the bottom as match of last resort
fixes #13645

### DIFF
--- a/lib/ansible/playbook/role/definition.py
+++ b/lib/ansible/playbook/role/definition.py
@@ -135,46 +135,44 @@ class RoleDefinition(Base, Become, Conditional, Taggable):
         append it to the default role path
         '''
 
-        role_path = unfrackpath(role_name)
+        # we always start the search for roles in the base directory of the playbook
+        role_search_paths = [
+            os.path.join(self._loader.get_basedir(), u'roles'),
+            self._loader.get_basedir(),
+        ]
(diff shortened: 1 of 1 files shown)


sivel:
This change does appear to work as expected.  From a simple playbook run, things work the same as they did in 1.9.  :+1:

PR #13654, opened 2015-12-23
Add ProxyCommand support to the paramiko connection plugin

sivel:
This PR adds OpenSSH like `ProxyCommand` support to the paramiko connection plugin.

This is one problem that I have at work, as we have to use a bastion to access all devices, which forces us to use the `ssh` connection plugin, regardless of whether the version of OpenSSH supports `ControlPersist`, which requires us to configure `ssh_args` to an empty string, or even supply `ControlPath` and such, without `ControlPersist`.

Now you can set up your `ansible.cfg` to have your ProxyCommand such as:

```
[paramiko_connection]
proxy_command = ssh -W "%h:%p" bastion
```

or

```
[paramiko_connection]
proxy_command = ssh bastion "nc %h %p"
```

This functionality requires `paramiko>=1.9.0`, which is included in the documentation and in the code.
##### What this PR does not address
- ~~an inventory variable for supplying this configuration~~

### DIFF
--- a/docsite/rst/intro_configuration.rst
+++ b/docsite/rst/intro_configuration.rst
@@ -739,6 +739,17 @@ instead.  Setting it to False will improve performance and is recommended when h
 
     record_host_keys=True
 
+.. _paramiko_proxy_command
+
+proxy_command
+=============
+
+.. versionadded:: 2.1
+
+Use an OpenSSH like ProxyCommand for proxying all Paramiko SSH connections through a bastion or jump host. Requires a minimum of Paramiko version 1.9.0. On Enterprise Linux 6 this is provided by ``python-paramiko1.10`` in the EPEL repository::
+
+    proxy_command = ssh -W "%h:%p" bastion
+
 .. _openssh_settings:
 
 OpenSSH Specific Settings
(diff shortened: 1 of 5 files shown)


sivel:
An earlier PR I submitted for adding `ansible_ssh_proxy` was eventually superseded by a more generic solution that became `ansible_ssh_common_args`.

Based on my above "What this PR does not address", I wonder if maybe we might look at `ansible_ssh_proxy` or `ansible_ssh_proxy_command`, now that we have a specific use.

My other thoughts are to create a `ansible_paramiko_proxy_command`, or to use `ansible_ssh_common_args` and see if we can parse out the `ProxyCommand`, for use in both plugins.

bcoca:
i would just stick it in ansible_ssh_common_args

sivel:
I've updated this PR, to parse `ansible_ssh_common_args` and pull out `ProxyCommand`.

This will allow people with the same inventory, to utilize the `ProxyCommand` provided by inventory, with either `ssh` or `paramiko`.

The basic inner workings of using `SETTINGS_REGEX` and the find/replace to replace out `%h`, `%p`, and `%r` are used within `paramiko` itself.

Sample hosts file:

```
server1 ansible_ssh_common_args='-o ProxyCommand "ssh -W %h:%p bastion"'
```

The above sample hosts file with `ansible_ssh_common_args` will work for both `ssh` and `paramiko` for setting up a `ProxyCommand`.

sivel:
As of now I have still left the `proxy_command` config value.  But if deemed we should drop that, and perhaps use `ssh_args` from `ssh` we can do that as well.

bcoca:
well, its really 'all of the above' as any of them could have the setting.

sivel:
The recent commit looks at `ssh_extra_args`, `ssh_common_args`, `ssh_args` in that order, (most specific to least), and then supports an override of `proxy_command` from the config file.

dagwieers:
Coincidentally I was looking at #14205 and discovered that **ProxyCommand** is actually a shell-command (so it supports stuff like `$( command )` and variable expansion `${var%%+*}`. Not sure if we want to go that far with Paramiko, just mentioning it.

sivel:
@dagwieers largely the ProxyCommand support is handled in paramiko as of today.  It is handled via Popen, but does not have access to shell as `shell=True` is not passed.  If we want to support shell expansion, we would either need to push that upstream, or use `paramiko.proxy.ProxyCommand` as a Mixin and override it's `__init__`.

For now though, I think we can start here, and if the need arises, maybe this adds some documentation to what is required.

bcoca:
shell expansion might open the can of worms that is supporting different shells ....

PR #13662, opened 2015-12-24
Provides more complete fix for unverified ctx

caphrim007:
Noted in the code is this link

  https://www.python.org/dev/peps/pep-0476/#opting-out

Which, upon further looking provides a more complete example of how to opt-out
of SSL verification.

I noted that on a system with the following attributes
- Debian wheezy
- python 2.7.3
- python-openssl 0.13-2+deb7u1

The following error is raised if attempting to disable SSL verification

```
    ssl._create_default_https_context = ssl._create_unverified_context
    AttributeError: 'module' object has no attribute '_create_unverified_context'
```

Completing the workaround using the code at python.org dismisses this error.

### DIFF
--- a/lib/ansible/module_utils/f5.py
+++ b/lib/ansible/module_utils/f5.py
@@ -63,7 +63,15 @@ def disable_ssl_cert_validation():
     # You probably only want to do this for testing and never in production.
     # From https://www.python.org/dev/peps/pep-0476/#id29
     import ssl
-    ssl._create_default_https_context = ssl._create_unverified_context
+
+    try:
+        _create_unverified_https_context = ssl._create_unverified_context
+    except AttributeError:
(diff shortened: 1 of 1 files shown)


abadger:
@caphrim007 have a look at https://github.com/ansible/ansible/issues/13667 and https://github.com/ansible/ansible-modules-extras/issues/1424  Assuming those looks good for 2.0.0 we'll use them as they fix other problems with the validate_certs code as well.  If we decide to hold off on those for post-2.0.0  then I'll look at integrating your fix as a temporary measure.

abadger:
#13667 and https://github.com/ansible/ansible-modules-extras/pull/1424 merged to devel and stable-2.0 so closing this one.  Thanks for noting this issue.  It brought the problem to our attention otherwise would have noticed the things that needed fixing!

PR #13667, opened 2015-12-24
bigip changes as requested by bcoca and abadger:

abadger:
- Fix to error if validate_cert is True and python doesn't support it.
- Only globally disable certificate checking if really needed.  Use
  bigip verify parameter if available instead.
- Remove public disable certificate function to make it less likely
  people will attempt to reuse that


bcoca:
supersedes #13662

abadger:
Labelled P1 only so this stays on the radar until we decide whether it needs to go to 2.0.0 or can wait until a later release.

caphrim007:
@abadger I can verify that this patch similarly suppresses the errors I had mentioned in #13662 so +1 from me.

abadger:
Pushed to devel and stable-2.0

PR #13668, opened 2015-12-24
Documenting default path for hosts file - Ansible homebrew installation

gbazilio:
While following the Getting Started, I had trouble running the first command since I installed Ansible via `brew`. And by doing that, the default path for the **hosts** file is `/usr/local/etc/ansible/hosts` instead of `/etc/ansible/hosts` as the doc says. There is an issue from another repository which helped me: https://github.com/skyl/ansible-sandbox/issues/1

Thanks! : )

### DIFF
--- a/docsite/rst/intro_getting_started.rst
+++ b/docsite/rst/intro_getting_started.rst
@@ -55,6 +55,8 @@ public SSH key should be located in ``authorized_keys`` on those systems::
 
 This is an inventory file, which is also explained in greater depth here:  :doc:`intro_inventory`.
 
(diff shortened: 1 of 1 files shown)


sivel:
Homebrew is explicitly not a supported installation method for Ansible.  As such we do not wish to add documentation around homebrew installs, as it may convey that homebrew is a supported way of installing Ansible on Mac.

Instead, Ansible should be installed via `easy_install` or `pip` as indicated at http://docs.ansible.com/ansible/intro_installation.html#latest-releases-on-mac-osx

PR #13675, opened 2015-12-26
Handle unreachable hosts second-last

Deiz:
This change brings the behaviour back in line with 1.9.

Without this change, an unreachable host will always mask hosts that have had a task failure, allowing the playbook to continue running instead of aborting when all hosts have finished executing the current play.

A simple playbook illustrating this is as follows:

``` yaml

---

- name: A
  hosts: all

- name: B
  hosts: all
  tasks:
    - name: fail
      fail:

- name: C
  hosts: all
```

If all hosts are reachable, this playbook will fail as expected when each host hits `B`. At present, if there are any unreachable hosts, the failures will be masked and all hosts will also execute `C`.

### DIFF
--- a/lib/ansible/plugins/strategy/__init__.py
+++ b/lib/ansible/plugins/strategy/__init__.py
@@ -110,12 +110,12 @@ class StrategyBase:
         unreachable_hosts = set(unreachable_hosts).union(self._tqm._unreachable_hosts.keys())
 
         # return the appropriate code, depending on the status hosts after the run
-        if len(unreachable_hosts) > 0:
-            return 3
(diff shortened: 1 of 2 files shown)


jimi-c:
Hi @Deiz, looking at older issues and we believe this has been fixed. I do note there is one slight difference between 1.9 and current devel (Play C in your example executes, however there are no remaining hosts so there is no output):

https://gist.github.com/jimi-c/ee557e0850524bd416303d3e12fa0b8e

If this is still an issue, please let us know, however we will close it for now. If you have any further questions, please let us know by stopping by one of the two mailing lists, as appropriate:

Thank you!

PR #13678, opened 2015-12-27
fix for datetime exception in ecs_tasks

Etherdaemon:
Proposed fix for https://github.com/ansible/ansible-modules-extras/issues/1348 in ansible-module-extras as remove_values() function is does not have an option to match the datetime.datetime type and throws an error. Also seen in other boto3 modules - in the ones I've written I popped a date_handler function but with this, that can be removed.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -65,6 +65,7 @@ import grp
 import pwd
 import platform
 import errno
+import datetime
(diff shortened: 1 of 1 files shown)


Zephirin:
Is anyone looking to merge this in soon?  This would be super helpful for deploys that I'm trying to do to aws that require awscli for steps as well as ansible.

PR #13682, opened 2015-12-28
make ansible-pull clone only once

### LINKED ISSUE #13681: ansible pull clones double: 127.0.0.1 and localhost
##### Summary:

ansible-pull runs clone twice.
##### Steps To Reproduce:

Run ansible-pull on a system with ipv4 working and localhost resolving, i.e. 
`/usr/bin/ansible-pull -d /var/lib/ansible/local -U your_git_repo`
##### Expected Results:

The ansible git module clone the repo once and then runs inside it.
##### Actual Results:

The clone is run twice, and one of them fails.

```
Starting Ansible Pull at 2015-12-28 16:00:42
/usr/bin/ansible-pull -o -d /var/lib/ansible/local -U git@bla

localhost | FAILED! => {
    "changed": false, 
    "failed": true, 
    "msg": "Failed to download remote objects and refs:  error: cannot lock ref 'refs/remotes/origin/master': Unable to create '/var/lib/ansible/local/.git/refs/remotes/origin/master.lock': File exists.\n\nIf no other git process is currently running, this probably means a\ngit process crashed in this repository earlier. Make sure no other git\nprocess is running and remove the file manually to continue.\nFrom bla\n ! XXX..XXX  master     -> origin/master  (unable to update local ref)\n"
}
127.0.0.1 | SUCCESS => {
    "after": "XXX", 
    "before": "XXX", 
    "changed": true
}
```

robinro:
- fixes #13681,  #13925
- the initial clone should only run once on localhost (remove 127.0.0.1 there)
- the actual playbook run should be with all possible hostnames for localhost
- '-l' options to ansible-pull are overwritten. Otherwise one could execute things with
  `-l otherhost -c local` which might lead to unexpected behavior

### DIFF
--- a/lib/ansible/cli/pull.py
+++ b/lib/ansible/cli/pull.py
@@ -208,8 +208,12 @@ class PullCLI(CLI):
         if self.options.ask_sudo_pass or self.options.ask_su_pass or self.options.become_ask_pass:
             cmd += ' --ask-become-pass'
         if self.options.tags:
+            import pdb; pdb.set_trace()  # XXX BREAKPOINT
             cmd += ' -t "%s"' % self.options.tags
-        cmd += ' -l "%s"' % limit_opts_play
+        if self.options.subset:
+            cmd += ' -l "%s"' % self.options.subset
+        else:
+            cmd += ' -l "%s"' % limit_opts_play
 
         os.chdir(self.options.dest)
 
--- a/lib/ansible/cli/pull.py
+++ b/lib/ansible/cli/pull.py
@@ -130,7 +130,8 @@ class PullCLI(CLI):
         # Now construct the ansible command
         node = platform.node()
         host = socket.getfqdn()
-        limit_opts = 'localhost,%s,127.0.0.1' % ','.join(set([host, node, host.split('.')[0], node.split('.')[0]]))
+        limit_opts_clone = 'localhost'
+        limit_opts_play = 'localhost,%s,127.0.0.1,::1' % ','.join(set([host, node, host.split('.')[0], node.split('.')[0]]))
         base_opts = '-c local '
         if self.options.verbosity > 0:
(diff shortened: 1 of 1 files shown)


robinro:
@bcoca please have a look at this

bcoca:
I'm not sure this is a good solution as it still leaves you open to multiple executions of ansible-playbook.

robinro:
One need multiple hosts, because the playbook could either use hostname or fqdn or localhost

robinro:
the last version now restores 1.9 behaviour for me while not breaking #7917

bcoca:
I went a step further:

```
index 2571717..e099555 100644
--- a/lib/ansible/cli/pull.py
+++ b/lib/ansible/cli/pull.py
@@ -160,15 +160,13 @@ class PullCLI(CLI):
             if not self.options.fullclone:
                 repo_opts += ' depth=1'

-
         path = module_loader.find_plugin(self.options.module_name)
         if path is None:
             raise AnsibleOptionsError(("module '%s' not found.\n" % self.options.module_name))

         bin_path = os.path.dirname(os.path.abspath(sys.argv[0]))
-        cmd = '%s/ansible -i "%s" %s -m %s -a "%s" "%s"' % (
-            bin_path, inv_opts, base_opts, self.options.module_name, repo_opts, limit_opts
-        )
+        # hardcode local and inventory/host as this is just meant to fetch the repo
+        cmd = '%s/ansible -i "localhost," -c local %s -m %s -a "%s" all' % (bin_path, base_opts, self.options.module_name, repo_opts)

         for ev in self.options.extra_vars:
             cmd += ' -e "%s"' % ev
```

this should guarantee that only one pull is done.

bcoca:
it is basically what 1.9 did, hardcode localhost for this purpose

bcoca:
fixed via 8d4bc2003f4b157bb40b61cdcc115388e645d5a7

PR #13684, opened 2015-12-28
Add generic data structures querying

filipenf:
This plugin adds a generic query-language to ansible. This is useful when you register the results of a module that has complex data in it. The idea of writing this plugin came when I needed to tag the ec2 instance volumes just after its creation. The ec2 module outputs something like:

```
ok: [localhost] => {
    "instance": {
        "changed": false,
        "instance_ids": null,
        "instances": [],
        "tagged_instances": [
            {
                ...
                "block_device_mapping": {
                    "/dev/sda1": {
                        "delete_on_termination": false,
                        "status": "attached",
                        "volume_id": "vol-yyyyy"
                    }
                    "/dev/sdb1": {
                        "delete_on_termination": false,
                        "status": "attached",
                        "volume_id": "vol-zzzzz"
                    }
                },
                ...
            }
        ]
    }
}
```

Currently the only way to get a list of volume ids is to invoke an `ec2_vol` task with `state: list`.

So I thought that having a jq-like query plugin would be great because it would solve not only this problem but many others. 

With this plugin we can query our data structures with something like: `.tagged_instances[*].block_device_mapping..volume_id`

```
- name: Tag volumes
  ec2_tag:
    region: "us-east-1"
    resource: "{{ item }}"
    tags:
      Name: "My volume"
  with_items:
    - "{{ instance | query('.tagged_instances[*].block_device_mapping..volume_id') }}"
```

Any comments to the PR would be appreciated.


amenonsen:
I love the functionality, but I don't think it should be a loo(ku)p plugin. It will be more generally useful if it's implemented as a filter.

filipenf:
I'm glad that you liked the functionality @amenonsen, I'm going to re-implement as a filter.

filipenf:
Re-implemented as a filter plugin. Thanks @amenonsen, much better this way.

chrismeyersfsu:
@filipenf can you show an example invocation as a filter?

filipenf:
Sure thing @chrismeyersfsu. 

```
➜  cat query-filter-test.yml
- hosts: localhost
  connection: local

  tasks:
    - uri: url="https://api.github.com/orgs/ansible/repos" return_content=yes
      register: repo_list
    - debug: msg="{{ repo_list.content | from_json | query('[0:10].name') }}"
```

Will display the first 10 repositories associated with _ansible_ organization:

```
➜  ANSIBLE_NOCOWS=1 ansible-playbook -i localhost, query-filter-test.yml

ok: [localhost]

TASK [uri] *********************************************************************
ok: [localhost]

ok: [localhost] => {
    "msg": [
        "ansible",
        "ansible.github.com",
        "ansible-examples",
        "ansible-for-rubyists",
        "tower-cli",
        "revelator",
        "training-content",
        "immutablish-deploys",
        "autoscaling-blog",
        "ansible-docker-base"
    ]
}

localhost                  : ok=3    changed=0    unreachable=0    failed=0
```

chrismeyersfsu:
@filipenf I think you need to remove the dependency on the `dq` library in order to get this accepted.

amenonsen:
@chrismeyersfsu I'd rather have the extra dependency and be able to use the complete functionality than to have some some cut-down query language, or worse still, an embedded copy of dq. Manipulating complex data structures is something Ansible is particularly weak at (partly because of jinja2's weaknesses), and what few existing solutions there are to this problem are decidedly special-purpose and limited (and I'm being charitable). It would be great to have some general-purpose facility like this PR merged (though it needs some tests, and certainly additions to the documentation).

sivel:
I have a concern in that dq isn't very battle tested. Hardly any downloads per month from pypi, no issues created on github, only 10 stars and no commits for 2 years.

filipenf:
I share this concern @sivel. I've thought about using objectpath (https://github.com/adriank/ObjectPath) but not really sure if it is a better alternative.

amenonsen:
The ObjectPath README says "We are looking for a maintainer!". That's not a promising start. (Note: I'm not arguing for dq in preference to ObjectPath; in fact, I know nothing about either one. I'm just arguing that we should be able to use _some_ purpose-built data structure query language.)

filipenf:
Agreed!

So we basically have 3 options if we want this filter added:
- use dq
- look for something else (the only other query-language library written in python I could find is [python-jsonpath-rw](https://github.com/kennknowles/python-jsonpath-rw)
- write our own query-language

I think that what matters most is the language syntax used by the chosen library ( we won't want to change that in the future to avoid compatibility issues). Also, if we choose a library, it will have a much larger user base, which will contribute to its future development.

Yannig:
Well, it seems that dq is based on python-jsonpath-rw so, we'd better use this one.

my 2cts :)

filipenf:
Switched to jsonpath_rw. I did run again with the same playbook as before and the query syntax worked!

filipenf:
Accidentally pushed devel to this branch. Is there any east way to fix this?

bcoca:
force push a correct branch

filipenf:
Done, thanks @bcoca

lib/ansible/plugins/filter/query.py
```diff
@@ -0,0 +1,48 @@
+# (c) 2015, Filipe Niero Felisbino <filipenf@gmail.com>
```
bcoca: this should really be named json_query

filipenf:
@bcoca renamed it. Also, the test is failing because jsonpath-rw is not present on CI system. Should I add it to `test-requirements.txt` or maybe `./test/utils/ansible-playbook_integration_runner/roles/ansible_test_deps/test/requirements.yml` is a better place?

mattclay:
@filipenf You can add jsonpath-rw to the existing `pip install` command in test/utils/shippable/integration.sh for now.

Periodically we'll update our CI docker images to include those packages to speed up CI.

filipenf:
@mattclay I added the jsonpath-rw installation and it worked for all targets except freebsd: `TEST=remote TARGET=all PLATFORM=freebsd VERSION=10.3-STABLE` - is there anything special for this one?

mattclay:
@filipenf Add `jsonpath-rw` to the the `pip install` in `test/utils/shippable/remote-integration.sh`.

emonty:
+1

filipenf:
Guys, I'm considering to switch from jsonpath-rw to [jmespath](https://github.com/jmespath/jmespath.py). Motivations:
- More complete library with better querying expressions
- Better documented
- Already a dependency for boto/awscli so more likely to be already installed on user's computer
- Widely used

For some reason I completely missed this one when I was researching about this kind of library to implement the filter.

What do you guys think? Any concerns?

filipenf:
As nobody is using this feature yet, I believe it is worth changing from jsonpath-rw to jmespath for the reasons mentioned previously.

With jmespath we'll be able to do filtering, sorting and some cool transformations in the input data with a well-defined syntax. Things like:

`people[?age >`20`].name`

and

`reservations[].instances[].[tags[?Key=='Name'].Values[] | [0], type, state.name]`

filipenf:
Shippable is breaking with:
`2016-08-06 14:41:33 fatal: [windows]: FAILED! => {"changed": false, "display_name": "SSDP Discovery", "failed": true, "msg": "Service 'SSDP Discovery (SSDPSRV)' stop failed.", "name": "SSDPSRV", "start_mode": "auto"}`

Don't think this is related to the changes I've made. Anyone?

bcoca:
lgtm, @jimi-c @mattclay @amenonsen @chrismeyersfsu @sivel @Yannig @emonty anyone care to weigh in?

emonty:
lgtm

amenonsen:
Thanks @filipenf, this looks very good to me.

smeijer:
This one is really awesome. I can't believe I couldn't find it in the docs.

Although; I do think `query` would have been a better name than `json_query`. Because we are not querying json, we are querying plain objects.

sidewinder12s:
@filipenf Had you found a way to use ansible variables within the json_query() filter? I am not sure if it's possible currently as it would technically be a variable within a variable.

smeijer:
@sidewinder12s just don't put the argument in quotes? Should work.

sidewinder12s:
@smeijer 
This is what I had been trying but I could never get ansible to evaluate the variable since I need the git_branch variable to evaluate and still have some kind of quotes around it for the json_query filter. 

```
"{{ elasticbeanstalk_versions | json_query('versions[].VersionLabel[] | [?starts_with(@, 'git_branch')] | [2:]') }}"
```

With elasticbeanstalk_versions & git_branch being the ansible variables.

filipenf:
@sidewinder12s The parameter for json_query is a string so you can make use of ansible's templating system to concatenate the string literal with the `git_branch` variable. Something like this:

`"{{ elasticbeanstalk_versions | json_query('versions[].VersionLabel[] | [?starts_with(@, '+git_branch+')] | [2:]') }}"`

sidewinder12s:
@filipenf Do you happen to know of any docs/walkthroughs that talk about this type of ansible templating? I've seen various references around when you get deep into Ansible/PyYAML/Jinja2 templating but there doesn't seem to be much docs/walkthroughs and associates all that with how it works for Ansible. 

On topic for anyone who ends up needing to do this, I finally got this working by setting an Ansible variable _with_ the single quotes that the starts_with() jmespath filter requires:

```
  - set_fact:
      app_var_git_branch: >-
        '{{ git_branch }}'
```

And invoked with the `+variable+`:

```
    with_items: >-
      {{ elasticbeanstalk_versions | json_query('versions[].VersionLabel[] | [?starts_with(@, '+app_var_git_branch+')] | [2:]') }}
```

This was the only way I could get it working since the starts_with() filter requires single quotes around the jmespath filter variable and if those aren't in the variable Ansible always seemed to remove them if it evaluated the variable correctly.

slavekjurkowski2:
Can anyone see why this variable expansion in this jmespath expression isn't working? I've run out of ideas on how to get this working..I tried many varieties of expanding vars in this expression and even tried sidewinder12s recommendation above..but no dice. 

```
---
- hosts: all
  vars:

      list_of_dict: [
          {
              "host": "txvslctapp01.clg.local",
              "port": "53620"
          },
          {
              "host": "txvslctapp02.clg.local",
              "port": "53621"
          }
        ]


      my_host_name: txvslctapp01.clg.local


  tasks:

  - set_fact:
      my_host_name2: >-
        '{{my_host_name}}'

  - name: check value evaluating var
    debug: msg="{{ list_of_dict | json_query(\"[?host=='my_host_name'].port\") }}"
    connection: local
    run_once: true

  - name: check value evaluating var2
    debug: msg="{{ list_of_dict | json_query(\"[?host=='+my_host_name2+'].port\") }}"
    connection: local
    run_once: true

  - name: check value with string
    debug: msg="{{ list_of_dict | json_query(\"[?host=='txvslctapp01.clg.local'].port\") }}"
    connection: local
    run_once: true

```

**Current output:**

```
vagrant@precise64:~/vsl_ansible_repo$ ansible-playbook test.yml -i nonprod_inventory --limit "tc02-dev"
 [WARNING]: Optional dependency 'cryptography' raised an exception, falling back to 'Crypto'


ok: [txvslctapp01.clg.local]
ok: [txvslctapp03.clg.local]
ok: [txvslctapp02.clg.local]

TASK [set_fact] ****************************************************************
ok: [txvslctapp02.clg.local]
ok: [txvslctapp01.clg.local]
ok: [txvslctapp03.clg.local]

TASK [check value evaluating var] **********************************************
ok: [txvslctapp01.clg.local] => {
    "msg": []
}

TASK [check value evaluating var2] *********************************************
ok: [txvslctapp01.clg.local] => {
    "msg": []
}

TASK [check value with string] *************************************************
ok: [txvslctapp01.clg.local] => {
    "msg": [
        "53620"
    ]
}

txvslctapp01.clg.local     : ok=5    changed=0    unreachable=0    failed=0
txvslctapp02.clg.local     : ok=2    changed=0    unreachable=0    failed=0
txvslctapp03.clg.local     : ok=2    changed=0    unreachable=0    failed=0
```

filipenf:
@slavekjurkowski2 it seems you've got wrong quoting there. Tried this:

```
  - name: check value evaluating var2
    debug: msg="{{ list_of_dict | json_query('[?host=='+my_host_name2+'].port') }}"
    connection: local
    run_once: true
```

and it worked fine:

```
➜  sandbox ANSIBLE_NOCOWS=1 ansible-playbook -c local -i localhost, ansible-query-with-expansion.yml

ok: [localhost]

TASK [set_fact] ****************************************************************
ok: [localhost]

TASK [check value evaluating var] **********************************************
ok: [localhost] => {
    "msg": []
}

TASK [check value evaluating var2] *********************************************
ok: [localhost] => {
    "msg": [
        "53620"
    ]
}

TASK [check value with string] *************************************************
ok: [localhost] => {
    "msg": [
        "53620"
    ]
}

localhost                  : ok=5    changed=0    unreachable=0    failed=0
```

slavekjurkowski2:
awesome..thanks for the help

Yannig:
@filipenf I wrote a little documentation at the following location: https://github.com/ansible/ansible/pull/18244

PR #13689, opened 2015-12-29
use pulled ansible.cfg for inventory

### LINKED ISSUE #13688: ansible-pull does not respect ansible.cfg in pulled repo

robinro:
- restore 1.9 behavior
- only give `-i` option to ansible-playbook if explicitly given to ansible-pull
  this then allows ansible-playbook to use the pulled ansible.cfg for the inventory
- leave behavior of all other tools unchanged
- fixes #13688

### DIFF
--- a/lib/ansible/cli/__init__.py
+++ b/lib/ansible/cli/__init__.py
@@ -210,8 +210,23 @@ class CLI(object):
         setattr(parser.values, option.dest, os.path.expanduser(value))
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


robinro:
behavior useable with the patch:
- `ansible-pull -i path`: still passed on to ansible-playbook
- `ansible-pull` without `-i`: 
  - before: `-i /etc/ansible/ansible.cfg` explicitly passed to ansible-playbook
  - now: no `-i` passed to ansible-playbook
  - ansible-playbook has the same defaults, so the only situation where the used inventory value changes is then there is a ansible.cfg present in the pulled repository that differs from the `/etc/ansible/ansible.cfg` one

bcoca:
superceded by #14257

PR #13690, opened 2015-12-29
hashi_vault lookup: be more rebust, and allow fields with other name than 'value'

feliksik:
extension on a plugin by @defionscode 
- do not stacktrace if the string does not contain proper '=' chars
- do not stacktrace if token= or url= or secret= is not specified, but give a proper error
- allow user to specify the field with secret=hello/world:value2 (i.e. you can use other fields than just 'value')
- allow user to query the entire secret dict with secret=hello/world:
- still allow user to use the 'old' behavior: secret=hello/world defaults to secret=hello/world:value
- do not stacktrace if the field is not available in the vault, but give a proper error

### DIFF
--- a/lib/ansible/plugins/lookup/hashi_vault.py
+++ b/lib/ansible/plugins/lookup/hashi_vault.py
@@ -15,7 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 #
-# USAGE: {{ lookup('hashi_vault', 'secret=secret/hello token=c975b780-d1be-8016-866b-01d0f9b688a5 url=http://myvault:8200')}}
+# USAGE: {{ lookup('hashi_vault', 'secret=secret/hello:value token=c975b780-d1be-8016-866b-01d0f9b688a5 url=http://myvault:8200')}}
 #
 # You can skip setting the url if you set the VAULT_ADDR environment variable
 # or if you want it to default to localhost:8200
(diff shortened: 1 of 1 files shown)


richardzone:
+1

feliksik:
Can an Ansible maintainer comment on this PR?

lib/ansible/plugins/lookup/hashi_vault.py
```diff
@@ -47,9 +47,25 @@ def __init__(self, **kwargs):
         except ImportError:
             AnsibleError("Please pip install hvac to use this module")
 
-        self.url = kwargs.pop('url')
-        self.secret = kwargs.pop('secret')
-        self.token = kwargs.pop('token')
+        self.url = kwargs.get('url')
+        if self.url==None:
+            self.url = ANSIBLE_HASHI_VAULT_ADDR
```
bcoca: why not `self.url = kwargs.get('url', ANSIBLE_HASHI_VAULT_ADDR)` ?

lib/ansible/plugins/lookup/hashi_vault.py
```diff
@@ -62,20 +78,28 @@ def get(self):
...
         vault_args = terms[0].split(' ')
         vault_dict = {}
         ret = []
 
         for param in vault_args:
-            key, value = param.split('=')
+            try:
+                key, value = param.split('=')
+            except ValueError as e:
+                raise AnsibleError("hashi_vault plugin needs key=value pairs, but received %s" % terms)
```
bcoca, line 99: terms can be a list, not sure if this is being handled correctly

lib/ansible/plugins/lookup/hashi_vault.py
feliksik: no good reason. I will take your suggestion.

lib/ansible/plugins/lookup/hashi_vault.py
feliksik, line 99: This may be true; I have no time to test that right now. However, it is not a regression as the existing implementation does not do anything special either. So it's not related to this PR, IMHO. 

I do not currently have time to work on this, but the improvements currently implemented are self-contained. 

Can we address the list-scenario in a separate issue?

vincentvdk:
tested the field function which works for me.

juliedavila:
Sorry for the delay on that

cornfeedhobo:
Sorry to comment on such an old issue, but I see it got merged, but I don't see this functionality in stable-2.1

PR #13691, opened 2015-12-29
output color is now configurable

bcoca:
next people will want to broaden to more than ansi colors ....


lib/ansible/plugins/callback/default.py
```diff
@@ -117,9 +117,7 @@ def v2_playbook_on_task_start(self, task, is_conditional):
         if self._display.verbosity > 2:
             path = task.get_path()
             if path:
-                self._display.display("task path: %s" % path, color='dark gray')
-
-    def v2_playbook_on_cleanup_task_start(self, task):
```
agaffney: Did you intend to kill off this function?

PR #13694, opened 2015-12-30
[2.0] cloudstack: fix more case insensitivity

resmo:
cloudstack: fix has_change reports changed for case insensitivity values
For some values, the API returns capitalized values while you used lower case ones. And for most values the API handles the values case insensitive. Which means, we would handle it as a change even thought for CloudStack it is not. There are some exceptions, we will handle them later in 2.1. For now, compare the string values case insensitive produces less unwanted changes.

Needs to be backported to stable-2.0


abadger:
Merged to devel and cherrypicked to stable-2.0.1 branch.

PR #13695, opened 2015-12-30
Correction of a double negation

mgarstecki:
The sentence seemed to imply that return codes from modules are significant, while they are not. The second part of the sentence confirms this, as it advises to use standard return codes only for future proofing.


jimi-c:
Merged, thanks!

PR #13700, opened 2015-12-30
[security] shred ansible-vault tmp_file. Also when editor is interruped

feliksik:
The implementation of ansible-vault has two problems when you do `ansible-vault edit my-encrypted-secrets.yml`: 

1) if ansible-vault receives a keyboard-interrupt during editing, the file is not deleted and the secrets are left plain, unencrypted on disk (chmod 600, granted). This keyboard interrupt is not likely when you are using vim, but it is when you are using gedit or meld (you have an attached terminal then, because you gave in the password there). 
2) with or without interrupt, the unencrypted file is not shredded by ansible-vault, just deleted. 

This patch fixes both issues.

### DIFF
--- a/lib/ansible/parsing/vault/__init__.py
+++ b/lib/ansible/parsing/vault/__init__.py
@@ -221,7 +221,22 @@ class VaultEditor:
         self.vault = VaultLib(password)
         
     def _shred_file(self, tmp_path):
-        """securely destroy a decrypted file."""
+        """Securely destroy a decrypted file
+        
+        Inspired by unix `shred', try to destroy the secrets "so that they can be 
+        recovered only with great difficulty with specialised hardware, if at all".
+        
+        See https://github.com/ansible/ansible/pull/13700 .
+        
+        Note that:
+        - For flash: overwriting would have no effect (due to wear leveling). But the
+          added disk wear is considered insignificant. 
+        - For other storage systems: the filesystem lies to the vfs (kernel), the disk 
+          driver lies to the filesystem and the disk lies to the driver. But it's better
+          than nothing.
+        - most tmp dirs are now tmpfs (ramdisks), for which this is a non-issue.
+        """
+        
         def generate_data(length):
             import string, random
             chars = string.ascii_lowercase + string.ascii_uppercase + string.digits
@@ -234,7 +249,7 @@ class VaultEditor:
         ld = os.path.getsize(tmp_path)
         passes = 3
         with open(tmp_path,  "w") as fh:
-            for _ in range(int(passes)):
+            for _ in range(passes):
                 fh.seek(0,  0)
                 data = generate_data(ld)
                 fh.write(data)
--- a/lib/ansible/parsing/vault/__init__.py
+++ b/lib/ansible/parsing/vault/__init__.py
@@ -219,41 +219,67 @@ class VaultEditor:
 
     def __init__(self, password):
         self.vault = VaultLib(password)
+    
+    def _shred_file_custom(self, tmp_path):
+        """"Destroy a file, when shred (core-utils) is not available
         
-    def _shred_file(self, tmp_path):
-        """Securely destroy a decrypted file
+        Unix `shred' destroys files "so that they can be recovered only with great difficulty with 
+        specialised hardware, if at all". It is based on the method from the paper 
+        "Secure Deletion of Data from Magnetic and Solid-State Memory", 
+        Proceedings of the Sixth USENIX Security Symposium (San Jose, California, July 22-25, 1996).
         
-        Inspired by unix `shred', try to destroy the secrets "so that they can be 
-        recovered only with great difficulty with specialised hardware, if at all".
+        We do not go to that length to re-implement shred in Python; instead, overwriting with a block
+        of random data should suffice. 
         
         See https://github.com/ansible/ansible/pull/13700 .
-        
-        Note that:
-        - For flash: overwriting would have no effect (due to wear leveling). But the
-          added disk wear is considered insignificant. 
-        - For other storage systems: the filesystem lies to the vfs (kernel), the disk 
-          driver lies to the filesystem and the disk lies to the driver. But it's better
-          than nothing.
-        - most tmp dirs are now tmpfs (ramdisks), for which this is a non-issue.
         """
         
-        def generate_data(length):
-            import string, random
-            chars = string.ascii_lowercase + string.ascii_uppercase + string.digits
-            return ''.join(random.SystemRandom().choice(chars) for _ in range(length))
+        file_len = os.path.getsize(tmp_path)
+        
+        passes = 3
+        with open(tmp_path,  "wb") as fh:
+            for _ in range(passes):
+                fh.seek(0,  0)
+                # get a random chunk of data 
+                data = os.urandom(min(1024*1024*2, file_len)) 
+                bytes_todo = file_len
+                while bytes_todo > 0:
+                    chunk = data[:bytes_todo]
+                    fh.write(chunk)
+                    bytes_todo -= len(chunk)
+                    
+                assert(fh.tell() == file_len)
+                os.fsync(fh)
+        
+            
+    def _shred_file(self, tmp_path):
+        """Securely destroy a decrypted file
+
+        Note standard limitations of GNU shred apply (For flash, overwriting would have no effect 
+        due to wear leveling; for other storage systems, the async kernel->filesystem->disk calls never
+        guarantee data hits the disk; etc). Furthermore, if your tmp dirs is on tmpfs (ramdisks), 
+        it is a non-issue. 
+        
+        Nevertheless, some form of overwriting the data (instead of just removing the fs index entry) is 
+        a good idea. If shred is not available (e.g. on windows, or no core-utils installed), fall back on 
+        a custom shredding method.
+        """
         
         if not os.path.isfile(tmp_path):
             # file is already gone
             return 
+        
+        try:
+            r = call(['shred', tmp_path])
+        except OSError as e:
+            # shred is not available on this system, or some other error occured. 
+            self._shred_file_custom(tmp_path)
+            r = 0
+        
+        if r != 0:
+            # we could not successfully execute unix shred; therefore, do custom shred. 
+            self._shred_file_custom(tmp_path)
             
-        ld = os.path.getsize(tmp_path)
-        passes = 3
-        with open(tmp_path,  "w") as fh:
-            for _ in range(passes):
-                fh.seek(0,  0)
-                data = generate_data(ld)
-                fh.write(data)
-                os.fsync(fh)
         os.remove(tmp_path)
         
     def _edit_file_helper(self, filename, existing_data=None, force_save=False):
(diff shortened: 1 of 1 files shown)


chrismeyersfsu:
1. For hard disks- A sync should proceed the write to ensure the contents hit the disk.
2. For flash - overwriting would have no effect. This is a more complex problem. Maybe you could deterministically mix up the ordering of the secret written based on the secret content.

bcoca:
An `os.fsync()` might help in some rotational drives, but this is a crapshoot as the filesystem lies to the vfs (kernel), the disk driver lies to the filesystem and the disk lies to the driver. But hopefully its 'secure enough' with just adding that. Add more lies if you have an HBA/storage system, x10 for network filesystems. 

On flash drives ... you basically need to wipe the drive, so you should skip the shredding as you are only shortening the drive life.

feliksik:
> An os.fsync() might help in some rotational drives, but this is a crapshoot as the filesystem lies to the vfs (kernel), the disk driver lies to the filesystem and the disk lies to the driver. But hopefully its 'secure enough' with just adding that. Add more lies if you have an HBA/storage system, x10 for network filesystems.

Yup, it would be just as secure as unix shred, AFAIK. But it's better than nothing, and most tmp disks should be local I think. 

> On flash drives ... you basically need to wipe the drive, so you should skip the shredding as you are only shortening the drive life.

Of course you're right. But I suggest to KISS and not detect the underlying hardware; I'm downloading 600MB iso's all the time, so the 100Kb secret yml file won't break my disk ;-) 

I implemented the os.fsync() for the shred method, please review/merge.

bcoca:
I'm ok with the current implementation, i would add what we noted above to the function docs and also add that most tmp dirs are now tmpfs (ramdisks).

feliksik:
@bcoca : done!

abadger:
Some changes needed:

Algorithmic problems:
- Once unencrypted, the vault files could contain utf-8 chars.  That means that we probably need to choose all byte values when overwriting, not just ascii-chars.
- The algorithm is too slow.  Some people have vars files which are (perhaps unreasonably) large.  bcoca estimates 100MB as a reasonable upper limit.  The raw algorithm for creating the random string hits the multi-second range when we get to 100K.
- In the same vein of large files (but which will make things slower :-( we may want to chunk the data to conserve memory.  Would maximum 1MB or 10MB chunks be a good compromise between memory usage and data throughput?

Stylistic problems:
- I'd move the helper function up a level.  nested functions generally make the code less readable and prevent reuse.  I avoid them unless there's a need to create a closure or do some dynamic function creation trickery.
- Similarly, imports should be at the toplevel at the beginning of the file unless there's a need to workaround circular imports or only import an optional dependency if truly needed.  Since these imports are for things in the stdlib, that shouldn't be the case.

Possiblity:

`random.SystemRandom().getrandbits(num_bits)` is _much_ faster than getting randomness one byte at a time.  Perhaps the algorithm can be adapted to use this instead (it returns an integer instead of bytes so it needs some modifications to the code) and then we can test that the whole algorithm is a reasonable speed for up to 100MB files.  If it's still not fast enough then we might want to also add a config option to enable or disable the shredding of temp files.

abadger:
@bcoca and I talked more about this in our chat channel and we're also wondering just what we're trying to protect against.  Instead of writing out random bits could we get away with just writing out 0x00 (possibly alternating passes of 0x00 and 0xff?)  Doing that would likely be even quicker than using getrandbits().

feliksik:
@abadger thanks

> Once unencrypted, the vault files could contain utf-8 chars. That means that we probably need to choose all byte values when overwriting, not just ascii-chars.

excellent point. That can be fixed easily, just like your style suggestions. 

> In the same vein of large files (but which will make things slower :-( we may want to chunk the data to conserve memory. Would maximum 1MB or 10MB chunks be a good compromise between memory usage and data throughput?

You are right, the buffer is too big. 1MB probably doesn't even yield better performance than 8 KB, I can fix this easily if the other issues are addressed. 

Let's talk about randomness and speed. 

> The algorithm is too slow. Some people have vars files which are (perhaps unreasonably) large. bcoca estimates 100MB as a reasonable upper limit. The raw algorithm for creating the random string hits the multi-second range when we get to 100K.

You are right, I did not test this. There are various options:
- read from os.urandom(n). `time dd if=/dev/urandom of=shredme bs=1MB count=100` takes 8s on my system, fairly fast. 
- do it in standard python. My experiments show that just _generating_ 100MB of random data (not writing, not repeating) with random.SystemRandom().getrandbits() takes about a minute: 

```
import timeit
b=64
timeit.timeit("import random; x=random.SystemRandom().getrandbits(%d)" % b, number=100*1024/(b/8  ) ) * 1024
```

It's faster for b=4096 (10s), but then we need to write the binary LargeInt data to disk which is more complicated. 
- do it with numpy. Fast, but obviously an undesired dependency. 
- use unix shred. 1.8s for 100MB. This was my initial desire as we shouldn't re-invent this wheel. I think it uses some pseudo-random thing given the speed, but is doing other sophisticated things based on a  [paper](https://www.cs.auckland.ac.nz/~pgut001/pubs/secure_del.html), as referenced by the manual. But obviously this is an extra, non-portable dependency. 
- write a repeating pattern -- let's say not zeros, but 256K of random data, repeatedly. Or possibly repeatedly choosing a random-length window with random offset of that binary string. Using a One-time-pad repeatedly is bad, but remember the residual data is not an XOR (still repeated semi-random or 10101010 is better than writing zeros, as some disk encoding trickeries apparently to do some RLL encoding).
- a chain of fallback-options. E.g. use unix shred if on *nix and it is installed; otherwise write a simple pattern. 

Given the limitations with SSD's and buffer-flushing as discussed, I suggest not making this too complicated. That is: 
- detect if shred is installed; 
  - if so, do `os.system("shred ...")` 
  - otherwise (i.e. windows or shred-less *nix): write a repeated random fragment. Optionally, write to stderr that shred can optionally be installed, and then exit successfully. 

Is that a satisfying solution?

feliksik:
> @bcoca and I talked more about this in our chat channel and we're also wondering just what we're trying to protect against. Instead of writing out random bits could we get away with just writing out 0x00 (possibly alternating passes of 0x00 and 0xff?) Doing that would likely be even quicker than using getrandbits().

You are right, it need not be entirely random. The only thing is that according to https://www.cs.auckland.ac.nz/~pgut001/pubs/secure_del.html , some disk use RLL encoding and don't write repeated bits, so it would effectively not overwrite much. The paper describes a sophisticated approach with many patterns that are applied repeatedly. That's why suggest the above solution of using os.system(shred) when available and otherwise write repeated random data, to make it safe but not  overcomplicate things. 

I think 1-3 passes would do. Your call. It'll an improvement either way.

feliksik:
based on feedback from @abadger , I updated the implementation. Should be fast enough now; for a 100 MB file, it does add a few seconds shredding through. 
I modified the function `write_data(self, data, filename)` to be `write_data(self, data, filename, shred=True)`; when decrypting and editing, the encrypted data needs not be shredded it calls with shred=False.

lib/ansible/parsing/vault/__init__.py
```diff
@@ -219,22 +219,90 @@ class VaultEditor:
...
+                data = os.urandom(min(1024*1024*2, file_len)) 
+                bytes_todo = file_len
+                while bytes_todo > 0:
+                    chunk = data[:bytes_todo]
+                    fh.write(chunk)
+                    bytes_todo -= len(chunk)
+                    
+                assert(fh.tell() == file_len)
+                os.fsync(fh)
+
```
abadger, line 255: - Slicing creates a copy so it might be better to avoid doing that inside of the inner loop
- We can move some calculations outside the loop (although it probably won't make much difference with just 3 passes).
- Instead of assert, we probably want to write a unittest to check that this does the right thing.  I'm okay with leaving the assert in until we  write a unittest, though.

So it would look something like this:

``` python

chunk_len = min(1024 * 1024 * 2, file_len)
with open(tmp_path, "wb") as fh:
    for _ in range(passes):
        fh.seek(0, 0)
        # get a random chunk of data
        data = os.urandom(chunk_len)
        for _ in range(0, file_len // chunk_len):
            fh.write(data)
        fh.write(data[:file_len % chunk_len])

        assert(fh.tell() == file_len)  # Remove this assert once we have unittests to check its accuracy
        os.fsync(fh)
```

Hopefully that's right and I'm not off by one byte :-)

lib/ansible/parsing/vault/__init__.py
```diff
@@ -219,22 +219,90 @@ class VaultEditor:
...
+            r = call(['shred', tmp_path])
+        except OSError as e:
+            # shred is not available on this system, or some other error occured. 
+            self._shred_file_custom(tmp_path)
+            r = 0
+        
+        if r != 0:
+            # we could not successfully execute unix shred; therefore, do custom shred. 
+            self._shred_file_custom(tmp_path)
+
```
abadger: Could be more compactly written as:

``` python
try:
    r = call(['shred', tmp_path])
except OSError as e:
    r = 1

if r ! = 0:
    # We could not successfully execute unix shred; therefore, do custom shred.
    self._shred_file_custom(tmp_path)
```

abadger:
Added some notes on how the code can be tightened up a little bit more.  Otherwise this looks good now.

+1 to merge once those changes are made.

lib/ansible/parsing/vault/__init__.py
feliksik: Yup, much better

lib/ansible/parsing/vault/__init__.py
feliksik, line 255: > Slicing creates a copy so it might be better to avoid doing that inside of the inner loop

Are you sure? 

Your code works fine, and it is indeed more elegant. 

Only thing is that the random chunk has the same size on each pass. I'm not sure about the recovery implications, but in crypto it'd be a weakness. Not sure if the same principles apply, but I could vary the chunk length.

feliksik:
Good. 
- The random chunk has the same size on each pass. I'm not sure about the recovery implications, but in crypto it'd be a weakness. Not sure if the same principles apply, but I now vary the chunk length.
- w.r.t. unit testing, I have very little time ATM so this would take a while for me to get to. How urgent is this for you? You aim for a release? 

Test cases would be: 
- encrypt and decrypt (verify hash) with shred available in path
- encrypt and decrypt  (verify hash) without shred available in path, for: 
  - files smaller than max_chunk_len/2
  - files larger than max_chunk_len

abadger:
Merged to devel!  thanks for all the work you put into this.

feliksik:
Great,  Thanks for the help!

abadger:
I think we're a little too close to 2.0.0 final to pull this in there but I'll put it in the queue to add to 2.0.1.

abadger:
Cherry-picked to the stable-2.0.1 branch here: https://github.com/ansible/ansible/commit/a2a786ed0cb2a51c61482df04f9ff40ea08d6256  the plan is once 2.0.0 is released and tagged we'll merge stable-2.0.1 into stable-2.0 and do further development towards the 2.0.1 release there.

PR #13708, opened 2016-01-01
Replace to_string by to_unicode.

### LINKED ISSUE #13707: Bug Report: to_string is not defined when using bad formated hosts inventory
##### Ansible Version:

last devel
##### Ansible Configuration:

none
##### Environment:

none
##### Summary:

When using user@host inventory, Ansible try to call to_string and crash 
##### Steps To Reproduce:

`ansible -m debug -i test@127.0.0.1, all`
##### Expected Results:

```
test@127.0.0.1 | SUCCESS => {
    "msg": "Hello world!"
}
```
##### Actual Results:

```
Unexpected Exception: global name 'to_string' is not defined
to see the full traceback, use -vvv
```

Yannig:
Fix https://github.com/ansible/ansible/issues/13707

### DIFF
--- a/lib/ansible/inventory/__init__.py
+++ b/lib/ansible/inventory/__init__.py
@@ -112,7 +112,7 @@ class Inventory(object):
                 try:
                     (host, port) = parse_address(h, allow_ranges=False)
                 except AnsibleError as e:
-                    display.vvv("Unable to parse address from hostname, leaving unchanged: %s" % to_string(e))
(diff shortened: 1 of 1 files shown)


abadger:
Thanks @Yannig !

note for others: to_string is also defined in utils.unicode.  Yannig's switch to to_unicode is correct, because display() is expecting unicode type strings.  to_string would be used if we were re-raising the exception with a modified message, not when we are passing it to display (as in this case).

PR #13709, opened 2016-01-02
Make credstash lookup plugin support encryption contexts

michel-slm:
Previously, the lookup plugin passes all its keyword arguments to
credstash's `getSecret`; while this works for passing the standard
parameters (version, region and table), this does not allow passing
a dictionary of key-value pairs as `getSecret`'s context parameter.

Instead, pop `version`, `region` and `table` from `kwargs`, supplying
the default value if they are not defined, and pass the rest of the `kwargs`
as the `context` parameter.

[Review on Reviewable](https://reviewable.io/reviews/ansible/ansible/13709)

### DIFF
--- a/lib/ansible/plugins/lookup/credstash.py
+++ b/lib/ansible/plugins/lookup/credstash.py
@@ -38,7 +38,11 @@ class LookupModule(LookupBase):
         ret = []
         for term in terms:
             try:
-                val = credstash.getSecret(term, **kwargs)
+                version = kwargs.pop('version', '')
+                region = kwargs.pop('region', None)
+                table = kwargs.pop('table', 'credential-store')
+                val = credstash.getSecret(term, version, region, table,
+                                          context=kwargs)
             except credstash.ItemNotFound:
                 raise AnsibleError('Key {0} not found'.format(term))
             except Exception as e:


michel-slm:
Test as follows:
- credstash put hello world app=ansible env=dev team=foo

Verify that `get` works:
- `credstash get hello` should fail
- `credstash get hello app=ansible env=dev team=foo` should work
- Create the following playbook:

```
---
# test_credstash.yml

- hosts: all
  tasks:
    - debug:
        msg: >
          {{ lookup('credstash',
          'hello',
          region=YOUR_REGION_HERE, version='0000000000000000001',
          app='ansible' env='dev' team='foo') }}"
```
- run the playbook:

```
ansible-playbook -c local -i 'localhost,' ./test_credstash.yml
```

michel-slm:
@bcoca any idea when this will get merged? Thanks.

jimi-c:
Merged, thanks!

alandotcom:
@mrwacky42 I think you can use a dictionary still if you use `**` -- https://docs.python.org/3/tutorial/controlflow.html#unpacking-argument-lists

I have a PR open to update the docs

gundalow:
Hi, there is a PR inflight to improve the docs. Reviews welcome on https://github.com/ansible/ansible/pull/32475

PR #13719, opened 2016-01-04
Fixed links in README

JamshedVesuna:
Fixed from broken links in README and removed extraneous whitespace.


bcoca:
These changes are not needed, the http://docs.ansible.com/community.html is correct and the /ansible redirect temporarily was lost due to the hosting solution move, but has since then been reinstated.

PR #13729, opened 2016-01-05
Connection plugins unicode fix

abadger:
Fixes problems discovered by @drybjed when using non-ascii characters in environment, script, and raw.  Also includes integration tests.


drybjed:
Just tested, works as advertised. I hope it will get into 2.0 before release. :-)

abadger:
Okay, merged to devel and stable-2.0.  I'll look into the display calls on the other connection plugins today to see if I can find other places where we could be generating tracebacks only when we use ansible with multiple -v's.

PR #13750, opened 2016-01-07
Don't break out of plays loop when some hosts fail

ktosiek:
##### SUMMARY
It used to break out of the wrong loop - first play with failed hosts silently stopped running the playbook, instead of just skipping the current play.

##### ISSUE TYPE
Feature Pull Request

##### ANSIBLE VERSION
N/A

### DIFF
--- a/lib/ansible/executor/playbook_executor.py
+++ b/lib/ansible/executor/playbook_executor.py
@@ -165,10 +165,10 @@ class PlaybookExecutor:
                             self._unreachable_hosts.update(self._tqm._unreachable_hosts)
                             self._tqm.clear_failed_hosts()
 
-                        # if the last result wasn't zero or 3 (some hosts were unreachable),
-                        # break out of the serial batch loop
-                        if result not in (0, 3):
-                            break
+                            # if the last result wasn't zero or 3 (some hosts were unreachable),
+                            # break out of the serial batch loop
+                            if result not in (0, 3):
+                                break
 
                     i = i + 1 # per play
 


bcoca:
cc @jimi-c

ktosiek:
This was created against the stable-2.0 branch (as I consider this a bug). The milestone suggests it's not considered for 2.0, should I rebase on devel?

jimi-c:
@ktosiek As far as my testing went, I thought 1.9 and before would continue on if things were unreachable, but errors would stop them from continuing. Is there an open issue showing the old behavior vs. the new with an example playbook? If not, could you provide us one?

If this is a bug, we'll fix it in 2.0.x.

ktosiek:
I've just checked and Ansible 1.9 does continue after one host's failure. I've checked by using this playbook:

```
- hosts: all
  tasks:
    - shell: 'false'
      when: inventory_hostname == play_hosts[0]

- hosts: all
  tasks:
    - shell: echo {{ inventory_hostname }}
```

invoked with: `ansible-playbook -i 'a, b, c' -c local fail.yml`

Also, no matter which behavior is correct, the comment in the code is certainly wrong now (as we are breaking out of plays loop, not serial batches loop)

EDIT: it would also be nice to get some information that playbook is stopping early if that's really the right behavior

ktosiek:
Is there anything I can do for this to move forward?

jimi-c:
We'll look to fix this for 2.1.0.

zsais:
@jimi-c Is there any movement on this PR?

hkariti:
i believe this doesn't fix a bigger issue - ansible behaves differently if there were some hosts marked as unreachable vs if there weren't when deciding if to continue. If there were only failures, the status code is 2. If there were the same number of failures with one unreachable host, the status is 3 and a different decision will be made. This is unexpected.

jeremypruitt:
I tried the code in this PR. On one hand it does seem to fix the issue of subsequent plays not being run for any hosts if a single host fails. But in my testing it seems to run all plays against all hosts, even if a host failed in a previous play. My expectation is that if a host fails then that host will not run any subsequent plays or tasks, but all other hosts should continue on to all plays and tasks (unless/until they fail).

jimi-c:
@ktosiek / @hkariti I've created a feature branch containing @vroy's work and a few fixes of my own. Could you please test that and see if it addresses your issues?

The branch is here:
https://github.com/ansible/ansible/compare/vroy_backward-compatible-executor

hkariti:
@jimi-c , @vroy looks good. The `on_no_hosts_remaining` callback is still called at the end of the play. Don't know if it's supposed to. It's not called when there aren't failed hosts in the play (unreachable hosts also don't cause it to be called).

jimi-c:
@hkariti hrm, perhaps we need to move that up to the PlaybookExecutor level rather than the strategy.

jimi-c:
If we do move that callback source, I don't think it'd be in 2.1.0 (maybe 2.1.1). Moving that may have other ramifications, and I don't want to delay the release of 2.1 at this point.

memelet:
FYI, This seems to be fixed for me in 2.1.1.RC2

jimi-c:
@memelet yes it's fixed, this issue is being held open only for the callback issue mentioned above.

Thanks for confirming though!

jimi-c:
@ktosiek ignore the above.

jctanner:
!needs_info

ktosiek:
@jctanner The bug this is about has been fixed for a while, so I'm closing this PR. If other aspects of handling host failure mentioned here are still problematic I think they should be moved to separate issues.

PR #13755, opened 2016-01-07
added new custom search and updated footer

docschick:
Swiftype custom search code put into the searchbox.html file while other
content was commented out (would like to test this with a production build
before removing the old content). Also updated footer with date, removed a
stray period, added "a division of Red Hat, Inc." with link as well.


bcoca:
you have extra commits here, including resetting the submodules

docschick:
Yeah, I noticed something seemed off and didn't want to poke at it more
until I had a clearer idea of what I did wrong. I need to back out those
extra bits for sure.

docschick:
Trying to reset my branch and get this in better shape. Closing and will try again.

PR #13756, opened 2016-01-07
Changing MD5 checksum code to work against files in use

smadam813:
Currently when using a module such as "win_stat", it will fail when trying to stat a file that is in use.  This is due to the way the MD5 checksum is being extracted via the current powershell code.  I have changed the code to no longer require locking the file in order to properly read an MD5 checksum.  In return, you can now run "win_stat" against files that are in use.

### DIFF
--- a/lib/ansible/module_utils/powershell.ps1
+++ b/lib/ansible/module_utils/powershell.ps1
@@ -210,10 +210,9 @@ Function Get-FileChecksum($path)
     $hash = ""
     If (Test-Path -PathType Leaf $path)
     {
-        $sp = new-object -TypeName System.Security.Cryptography.SHA1CryptoServiceProvider;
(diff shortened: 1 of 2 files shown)


jhawkesworth:
Nice catch.

nitzmahone:
@smadam813 Thanks for the proposed fix, however, this change causes the entire file to be read into memory- ok for small files, but a major problem for big ones... The previous implementation fed the stream directly to the CSP, which calculates the hash in an efficient streaming fashion.

The right way to fix this is to pass the FileShare flag to File.Open so we don't expect exclusive access to the file- so long as other processes don't request exclusive access either (which will always fail for the second reader), we should be good.

I've made this change in 2.1 at ac54f66741645315eea37d36bd8350cea2e16840 - please try it out...

Issue #13234, opened 2015-11-20
Unformatted PyPI page


mgedmin:
Hello!

https://pypi.python.org/pypi/ansible looks like this, which looks a bit unprofessional:
![screenshot](https://cloud.githubusercontent.com/assets/159967/11303705/e6dbe908-8fac-11e5-9cd2-d83386890570.png)

If you convert the markup to ReStructuredText, it'll be rendered properly.  You can use pandoc for that.

Many projects use the same text for their GitHub README.rst and for their PyPI pages, by having the setup.py reuse the same text:

``` python
import setuptools

with open('README.rst') as f:
    long_description = f.read()

setup(
    ...,
    long_description=long_description,
    ...)
```

I'm not saying you should do that, but maybe it'll make your life easier?

ansibot:
@mgedmin Greetings! Thanks for taking the time to open this issue. In order for the community to handle your issue effectively, we need a bit more information. 

Here are the items we could not find in your description:
- issue type
- ansible version
- component name

Please set the description of this issue with this template:
https://raw.githubusercontent.com/ansible/ansible/devel/.github/ISSUE_TEMPLATE.md

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibotdev:
@mgedmin You have not responded to information requests in this issue so we will assume it no longer affects you. If you are still interested in this, please create a new issue with the requested information.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

PR #13758, opened 2016-01-07
moved README.md to README.rst, normalizes on rst and auto uploads description to pipy

StephenWeber:
Set a default long_description based on the first few lines of README.md

Added a script that will convert README.md into ReStructured Text for
PyPI, send project metadata to PyPI, and delete the temporary file.

Code based on
https://coderwall.com/p/qawuyq/use-markdown-readme-s-in-python-modules

Addresses #13234 when script applied to project by core dev with access to the PyPI account.


StephenWeber:
Alternatively, this script could be folded into setup.py at the risk that it could cause problems / unnecessary dependencies. The script could also live elsewhere (such as in ./hacking)

sivel:
I think I would prefer just converting the README.md to README.rst permanently and dropping the md file.  That would solve a few problems and make this easier in the process.

bcoca:
would rst appear readable in github?

bcoca:
duh! docsite/rst/ ... seems to work well enough, I'm happy to have 1 less format to deal with

StephenWeber:
Converting the MD to RST permanently would also be acceptable, and I can do that instead if that's the desire.

bcoca:
quick straw poll in core team, with over 50% votes in, its unanimous for unifying formats on RST.

gundalow:
Rebase needed
Also it's been agreed we should use these headings http://openalea.gforge.inria.fr/doc/openalea/doc/_build/html/source/sphinx/rest_syntax.html#headings

StephenWeber:
Gods I've made a mess.

PR #13764, opened 2016-01-08
Add 'sudo su -' become option

ntkoopman:
We manage a server were we do not have access to the sudoers file. The only way on this server to execute commands is to use 'sudo su -'.

This adds a 'sudosu' option to 'become_method'. It did not use the become command argument, because it is not clear if that would reference the sudo or su part. The same goes for the flags, which currently apply to the sudo part.

### DIFF
--- a/lib/ansible/constants.py
+++ b/lib/ansible/constants.py
@@ -182,9 +182,9 @@ DEFAULT_SUDO_FLAGS        = get_config(p, DEFAULTS, 'sudo_flags', 'ANSIBLE_SUDO_
 DEFAULT_ASK_SUDO_PASS     = get_config(p, DEFAULTS, 'ask_sudo_pass',    'ANSIBLE_ASK_SUDO_PASS',    False, boolean=True)
 
 # Become
-BECOME_ERROR_STRINGS      = {'sudo': 'Sorry, try again.', 'su': 'Authentication failure', 'pbrun': '', 'pfexec': '', 'runas': '', 'doas': 'Permission denied'} #FIXME: deal with i18n
-BECOME_MISSING_STRINGS    = {'sudo': 'sorry, a password is required to run sudo', 'su': '', 'pbrun': '', 'pfexec': '', 'runas': '', 'doas': 'Authorization required'} #FIXME: deal with i18n
-BECOME_METHODS            = ['sudo','su','pbrun','pfexec','runas','doas']
+BECOME_ERROR_STRINGS      = {'sudo': 'Sorry, try again.', 'su': 'Authentication failure', 'pbrun': '', 'pfexec': '', 'runas': '', 'doas': 'Permission denied', 'sudosu': ''} #FIXME: deal with i18n
+BECOME_MISSING_STRINGS    = {'sudo': 'sorry, a password is required to run sudo', 'su': '', 'pbrun': '', 'pfexec': '', 'runas': '', 'doas': 'Authorization required', 'sudosu': ''} #FIXME: deal with i18n
+BECOME_METHODS            = ['sudo','su','pbrun','pfexec','runas','doas','sudosu']
 BECOME_ALLOW_SAME_USER    = get_config(p, 'privilege_escalation', 'become_allow_same_user', 'ANSIBLE_BECOME_ALLOW_SAME_USER', False, boolean=True)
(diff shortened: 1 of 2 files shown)


bcoca:
There are many issues with this, some people are restricted to `/bin/su`, or `/bin/su - /bin/bash` or `/bin/su <username>` and many other combinations, this patch will only solve a small portion of those.

janurag:
+1

bcoca:
several issues here:
- the quoting is wrong as the options for su can be interpreted as part of sudo
- sudo and su are hardcoded, they are not always named sudo or su
- you assume both are in teh path, this is also not always true and as the above should be settable by the user.

ntkoopman:
Given those issues, what do you propose? It sounds like it would be easier and more flexible if there was some generic 'wrap' option (basically allow the user to specify becomecmd themselves).

bcoca:
The user would have to deal with hellish quoting issues and shell interactions, that is why this is not a free form config.

I know the `sudo su - user` pattern is fairly widespread even though it is totally unnecessary (`sudo -u user`). I've been trying to think for a solution for some time now and the only thing that I think can ultimately serve all users and deal with the problems mentioned (and others unmentioned) is making the commands 'stackable'. But this is far from trivial and I'm still not sure how the configuration/invocation would look.

amenonsen:
@ntkoopman can you show me what the relevant entry in sudoers looks like? (I'm also having trouble understanding why anyone would do that.)

ntkoopman:
The offending line is sudoers is
`%server_admins_grp    ALL=(root)  NOPASSWD: ALL`

We could save the user from the quoting issues by making the option a list of wrapped commands:

``` yaml
wrap:
  - /bin/su - user -G extra_group -c %s
  - /usr/bin/sudo %s
```

or used with the current options

``` yaml
become_multiple:
  - method: su
    user: user
    flags: -G extra_group
  - method: sudo
```

amenonsen:
> `%server_admins_grp    ALL=(root)  NOPASSWD: ALL`

Heh. That just seems… so pointless, giving someone root but preventing
them from using sudo -u.

> - /bin/su - user -G extra_group -c %s
> - /usr/bin/sudo %s

Having just spent ages testing various sudo/su/sh combinations, I am not
much in favour of this. For example, what you have above will break if
either user's shell is not a Bourne shell.

Sure, you could say "we're just letting the user set whatever command
they want, if they get it wrong then it's their problem", but I really
have very little confidence that it can be dealt with in a sane manner.
For example, sudo takes only an unquoted multi-word command as an
argument, but su will not (the command must be quoted), and we have no
good way to tell which one behaves which way if we're specifying a list
of arbitrary commands.

I wish I could think of some good way to help with your use-case, but I
can't. I know it's not what you want to hear, but I really think this is
fundamentally a sudo misconfiguration.

ntkoopman:
That is unfortunate, because the sudo configuration is outside our control. 

A workaround is (obvious in hindsight) `ansible_become_method=su ansible_become_exe="sudo su -"` in the inventory. Maybe that is enough until other use cases appear.

kevclarx:
This is a problem for us as well, I guess we can just maintain our own fork but it would be great to see it incorporated.  In our case we use a sudo replacement program but the issue is the same, we have to "<sudo replacement> su -" to become root, and this is not going to change anytime soon.

piercer:
Shame this never made it in :-( We are in the same boat - only allowed mechanism for escalation is 

`sudo su - <username> `

and we have no power to change that. 

Additionally we need to provide a password, and --ask-become-pass does not work with the ansible_become_method=su ansible_become_exe="sudo su -" workaround proposed above.

Real showstopper for us.

adaikal:
@ntkoopman Almost lost cool, thanks for your workaround. Cheers!.

vigyr:
Hi @ntkoopman , I have added the new sudosu become method as you suggested. But still I receive timeout error.  Not sure what the issue is. As per below output, it is doing sudo su - user1, and then timeout occurred. Can you please let me know what I am missing?

'/bin/sh -c '"'"'sudo su - user1 /bin/sh -c '"'"'"'"'"'"'"'"'echo BECOME-SUCCESS-foprluotlyzmzvcojjmnjoopsovvxibb; /usr/bin/python /var/tmp/ansible-tmp-1551377395.14-20258538643491/command.py'"'"'"'"'"'"'"'"' && sleep 0'"'"''
fatal: [va10dlvorc300]: FAILED! => {
    "msg": "Timeout (12s) waiting for privilege escalation prompt: "
}

PR #13769, opened 2016-01-08
added version information to subelements loop

xshadow:
added version information to subelements loop, since its only available in version 2.0

### DIFF
--- a/docsite/rst/playbooks_loops.rst
+++ b/docsite/rst/playbooks_loops.rst
@@ -207,6 +207,7 @@ It might happen like so::
          - users
          - authorized
 
+New in version 2.0.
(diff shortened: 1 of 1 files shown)


sivel:
This is incorrect.  subelements has existed since v1.4.0

bcoca:
closing as per above

xshadow:
May subelements existed since v1.4.0  but: 

> > Given the mysql hosts and privs subkey lists, you can also iterate over a list in a nested subkey 

and 

> > Optionally, you can add a third element to the subelements list, that holds a dictionary of flags. Currently you can add the ‘skip_missing’ flag.

is only possible in v2.0!! And that is misleading

PR #13771, opened 2016-01-08
First pass at allowing binary modules

sivel:
#6374 is a bit out of date with all of the changes for 2.0.

This PR adds some proof of concept changes that allow ansible to utilize binary modules that have no shebang and whatnot.

binary detection is done similarly to the `file` command using a concept found at http://stackoverflow.com/a/7392391

There are no doubt some things missing right now, but this is working for the standard action (~~not tested with async~~) and the `ssh` transport.

A few things that need to be looked at:
- ~~async~~ (Async seems to work fine per my tests)
- ~~windows~~ (Had to change some things around to allow `.exe` extensions for modules, but working now, assumes windows binaries are always `.exe`)
- ~~binaries other than go, and non ELF~~ (Have testing things from executables, to images, to PDFs, which all report properly)

### DIFF
--- a/docsite/rst/developing_modules.rst
+++ b/docsite/rst/developing_modules.rst
@@ -209,7 +209,7 @@ This should return something like::
 Binary Modules Input
 ~~~~~~~~~~~~~~~~~~~~
 
-Support for binary modules was added in Ansible 2.1.  When Ansible detects a binary module, it will proceed to
+Support for binary modules was added in Ansible 2.2.  When Ansible detects a binary module, it will proceed to
 supply the argument input as a file on ``argv[1]`` that is formatted as JSON.  The JSON contents of that file
 would resemble something similar to::
 
--- a/docsite/rst/developing_modules.rst
+++ b/docsite/rst/developing_modules.rst
@@ -211,10 +211,11 @@ Binary Modules Input
 
 Support for binary modules was added in Ansible 2.2.  When Ansible detects a binary module, it will proceed to
 supply the argument input as a file on ``argv[1]`` that is formatted as JSON.  The JSON contents of that file
-would resemble something similar to::
+would resemble something similar to the following payload for a module accepting the same arguments as the
+``ping`` module::
 
     {
-        "name": "Ansible",
+        "data": "pong",
         "_ansible_verbosity": 4,
         "_ansible_diff": false,
         "_ansible_debug": false,
--- a/docsite/rst/developing_modules.rst
+++ b/docsite/rst/developing_modules.rst
@@ -204,6 +204,24 @@ This should return something like::
 
     {"changed": true, "time": "2012-03-14 12:23:00.000307"}
 
+.. _binary_module_reading_input:
+
+Binary Modules Input
+~~~~~~~~~~~~~~~~~~~~
+
+Support for binary modules was added in Ansible 2.1.  When Ansible detects a binary module, it will proceed to
+supply the argument input as a file on ``argv[1]`` that is formatted as JSON.  The JSON contents of that file
+would resemble something similar to::
+
+    {
+        "name": "Ansible",
+        "_ansible_verbosity": 4,
+        "_ansible_diff": false,
+        "_ansible_debug": false,
+        "_ansible_check_mode": false,
+        "_ansible_no_log": false
+    }
+
 .. _module_provided_facts:
 
 Module Provided 'Facts'
--- a/lib/ansible/executor/module_common.py
+++ b/lib/ansible/executor/module_common.py
@@ -490,11 +490,9 @@ def recursive_finder(name, data, py_module_names, py_module_cache, zf):
         # Save memory; the file won't have to be read again for this ansible module.
         del py_module_cache[py_module_file]
 
-def _is_binary(module_path):
+def _is_binary(module_data):
     textchars = bytearray(set([7, 8, 9, 10, 12, 13, 27]) | set(range(0x20, 0x100)) - set([0x7f]))
-
-    with open(module_path, 'rb') as f:
-        start = f.read(1024)
+    start = module_data[:1024]
     return bool(start.translate(None, textchars))
 
 def _find_snippet_imports(module_name, module_data, module_path, module_args, task_vars, module_compression):
@@ -511,7 +509,9 @@ def _find_snippet_imports(module_name, module_data, module_path, module_args, ta
     # module_substyle is extra information that's useful internally.  It tells
     # us what we have to look to substitute in the module files and whether
     # we're using module replacer or ziploader to format the module itself.
-    if REPLACER in module_data:
+    if _is_binary(module_data):
+        module_substyle = module_style = 'binary'
+    elif REPLACER in module_data:
         # Do REPLACER before from ansible.module_utils because we need make sure
         # we substitute "from ansible.module_utils basic" for REPLACER
         module_style = 'new'
@@ -528,8 +528,6 @@ def _find_snippet_imports(module_name, module_data, module_path, module_args, ta
         module_substyle = 'jsonargs'
     elif b'WANT_JSON' in module_data:
         module_substyle = module_style = 'non_native_want_json'
-    elif _is_binary(module_path):
-        module_substyle = module_style = 'binary'
 
     shebang = None
     # Neither old-style, non_native_want_json nor binary modules should be modified
--- a/lib/ansible/executor/module_common.py
+++ b/lib/ansible/executor/module_common.py
@@ -490,6 +490,13 @@ def recursive_finder(name, data, py_module_names, py_module_cache, zf):
         # Save memory; the file won't have to be read again for this ansible module.
         del py_module_cache[py_module_file]
 
+def _is_binary(module_path):
+    textchars = bytearray(set([7, 8, 9, 10, 12, 13, 27]) | set(range(0x20, 0x100)) - set([0x7f]))
+
+    with open(module_path, 'rb') as f:
+        start = f.read(1024)
+    return bool(start.translate(None, textchars))
+
 def _find_snippet_imports(module_name, module_data, module_path, module_args, task_vars, module_compression):
     """
     Given the source of the module, convert it to a Jinja2 template to insert
@@ -521,11 +528,13 @@ def _find_snippet_imports(module_name, module_data, module_path, module_args, ta
         module_substyle = 'jsonargs'
     elif b'WANT_JSON' in module_data:
         module_substyle = module_style = 'non_native_want_json'
+    elif _is_binary(module_path):
+        module_substyle = module_style = 'binary'
 
     shebang = None
-    # Neither old-style nor non_native_want_json modules should be modified
+    # Neither old-style, non_native_want_json nor binary modules should be modified
     # except for the shebang line (Done by modify_module)
-    if module_style in ('old', 'non_native_want_json'):
+    if module_style in ('old', 'non_native_want_json', 'binary'):
         return module_data, module_style, shebang
 
     output = BytesIO()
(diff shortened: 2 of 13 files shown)


bcoca:
he, i was doing same last night, but ran out of gas.

bcoca:
i was actually going to leverage the stat module to get the 'binary status'

sivel:
Just as an example, this is the golang code for the module I was testing with:

```
package main

import (
    "fmt"
    "os"
)

func main() {
    fmt.Printf(`{"arg": "%v"}`, os.Args[1])
}
```

A more full featured example can be found at: https://gist.github.com/sivel/ccd81bdfb31ca0c0e05d

sivel:
So far it looks like everything is working as expected.  I've tested with ssh, paramiko, local, and winrm.  I've also mixed non binary and binary modules within the same playbook.

See this gist for details about the test playbook and example module: https://gist.github.com/sivel/ccd81bdfb31ca0c0e05d

cygnetix:
This is awesome @sivel. Thanks for taking the time to work on this. I can't wait to see it merged (fingers crossed).

tom025:
Thanks for effort @sivel can't wait to see this merged too.

sivel:
I've added integration tests for binary modules.

Executed like:

```
INVENTORY=binary_hosts make test_binary_modules
```

My `binary_hosts` looks roughly like:

```
localhost ansible_connection=local
windows ansible_host=1.1.1.1 ansible_user=Administrator ansible_password=MY_PASSWORD ansible_port=5986 ansible_connection=winrm ansible_winrm_server_cert_validation=ignore ansible_winrm_transport=ssl
paramiko ansible_host=2.2.2.2 ansible_user=root ansible_password=MY_PASSWORD ansible_connection=paramiko
ssh ansible_host=3.3.3.3 ansible_user=root ansible_password=MY_PASSWORD ansible_connection=ssh
```

In my particular case `localhost` is a Mac (darwin).

sivel:
The current build failure is due to travis not properly installing python2.4 and not related to this PR.

sivel:
Rebased and merge conflicts resolved due to https://github.com/ansible/ansible/commit/b4b24a0889ae52a4ee88e5754eef97ed258915a7

sivel:
Per the core meeting today, this will be targeted for implementation in v2.2, to land in devel after the 2.1 release (https://meetbot.fedoraproject.org/ansible-meeting/2016-04-12/ansible_core_meeting.2016-04-12-19.02.html)

I've rebased against devel, and handled merge conflicts.

I have fixed failing tests, but have not added any additional unit tests around this as of yet.  Integration tests are there, but are not currently running as part of `make`.

lib/ansible/plugins/action/__init__.py
```diff
@@ -549,7 +563,7 @@ def _execute_module(self, module_name=None, module_args=None, tmp=None, task_var
 
         # Fix permissions of the tmp path and tmp files.  This should be
         # called after all files have been transferred.
-        self._fixup_perms(tmp, remote_user, recursive=True)
+        self._fixup_perms(tmp, remote_user, recursive=True, execute=is_binary)
```
abadger: I've been thinking perhaps we should always set execute=True?  In the python module case, it is still a script intended to be executed so it seems reasonable.

lib/ansible/plugins/action/__init__.py
```diff
@@ -146,7 +153,10 @@ def _configure_module(self, module_name, module_args, task_vars=None):
         # insert shared code and arguments into the module
         (module_data, module_style, module_shebang) = modify_module(module_name, module_path, module_args, task_vars=task_vars, module_compression=self._play_context.module_compression)
 
-        return (module_style, module_shebang, module_data)
+        if self._is_binary(module_path):
+            return ('non_native_want_json', None, None, module_path, True)
+
```
abadger: This check (and _is_binary) seem like they belong in executor/module_common.py.  Make modify_module responsible for detecting that this is a binary module and returning it unchanged.  We might need to create a new module type so that the rest of the code in action/__init__.py knows it's dealing with a binary module but that shouldn't be a problem.  We already have "old" and "non_native_want_json" which do nearly the same things.

abadger:
I like this for 2.2.  I reviewed everything except the shell changes.  @nitzmahone you probably shold look at those since it looks like powershell has the most complexity to it.

About the integration tests -- it would be nice to turn it on by default.  I wonder if we could build the go module and have it in the git repo in addition to the source code?  Would that work?  What do others think about that?

docsite/rst/developing_modules.rst
```diff
@@ -200,6 +200,24 @@ This should return something like::
...
+
+Binary Modules Input
+~~~~~~~~~~~~~~~~~~~~
+
+Support for binary modules was added in Ansible 2.2.  When Ansible detects a binary module, it will proceed to
+supply the argument input as a file on ``argv[1]`` that is formatted as JSON.  The JSON contents of that file
+would resemble something similar to::
+
+    {
+        "name": "Ansible",
```
abadger: Maybe use something different than name and Ansible here to be clear that this the module parameters rather than more special variables.  Maybe something like "data" like the ping module uses?

lib/ansible/plugins/action/__init__.py
bcoca: We execute the interpreter directly by substituting ansible_<interpreter>_interpreter so not sure that this is needed (unless we want to change to always execute the file?

nitzmahone:
@abadger agreed on all counts- if the integration tests aren't running on this all the time, it _will_ get broken. I need to try with a .NET module and see, but in that case, we definitely don't want any interpreter (just direct exec the module)

sivel:
Yeah, I intend on having the tests run.  I only mentioned that they are not currently running.  I haven't had time to test that we can build the binary modules (Golang) at run time.  I think we should be fine though.

lib/ansible/plugins/action/__init__.py
sivel: I'll look into it.  It probably won't be immediately though.

lib/ansible/plugins/action/__init__.py
abadger: The two things I was thinking about were: When someone sets ANSIBLE_KEEP_REMOTE_FILES=1 and then wants to execute the wrapper and async.  async currently sets execute=True so that async_wrapper can execute the module that was copied over.  Started me wondering if there's any harm to just setting execute=True by default.

sivel:
Just as an fyi, I haven't taken the time to come back to this yet, and try to move the binary detection to `executor/module_common.py`.  Still on my list, but just haven't looked at doing so yet.

sivel:
Ok, I have some updates, and I think I have addressed some if not all of the comments:
1. I have moved `binary` detection into `executor/module_common.py`
2. I have changed `_fixup_perms` to default `execute=True`
3. I have update the docs to hopefully make things a bit more clear
4. Integration tests are running now
   1. although integration tests are running, I am grabbing golang on the fly and still building the modules during test time
   2. I am not sure how I feel about storing the binary modules in the repo.  Those binaries total about ~3MB each, due to go and how it statically links.  I would be more comfortable putting them someplace http accessible instead.
   3. if we keep them as source, it may make sense to put golang in the docker test containers, instead of grabbing the binaries for go on each build.

lib/ansible/executor/module_common.py
```diff
@@ -490,6 +490,13 @@ def recursive_finder(name, data, py_module_names, py_module_cache, zf):
         # Save memory; the file won't have to be read again for this ansible module.
         del py_module_cache[py_module_file]
 
+def _is_binary(module_path):
+    textchars = bytearray(set([7, 8, 9, 10, 12, 13, 27]) | set(range(0x20, 0x100)) - set([0x7f]))
+
+    with open(module_path, 'rb') as f:
+        start = f.read(1024)
```
abadger: Do we need to re-read this file or can we adapt this code to use module_data?

lib/ansible/executor/module_common.py
```diff
@@ -521,11 +528,13 @@ def _find_snippet_imports(module_name, module_data, module_path, module_args, ta
         module_substyle = 'jsonargs'
     elif b'WANT_JSON' in module_data:
         module_substyle = module_style = 'non_native_want_json'
+    elif _is_binary(module_path):
+        module_substyle = module_style = 'binary'
```
abadger: We might want this to be the first thing we check since `if REPLACER in module_data` is substituting something in the file.

lib/ansible/executor/module_common.py
```diff
@@ -731,7 +740,9 @@ def modify_module(module_name, module_path, module_args, task_vars=dict(), modul
 
     (module_data, module_style, shebang) = _find_snippet_imports(module_name, module_data, module_path, module_args, task_vars, module_compression)
 
-    if shebang is None:
+    if module_style == 'binary':
+        return (module_path, module_data, module_style, shebang)
```
abadger: Do we need to return module_path?  I don't see that we modify it anywhere and the calling code already has it.

lib/ansible/executor/module_common.py
```diff
@@ -753,4 +764,4 @@ def modify_module(module_name, module_path, module_args, task_vars=dict(), modul
     else:
         shebang = to_bytes(shebang, errors='strict')
 
-    return (module_data, module_style, shebang)
+    return (module_path, module_data, module_style, shebang)
```
abadger: Same question about returning module_path here.

lib/ansible/plugins/action/__init__.py
```diff
@@ -145,9 +145,9 @@ def _configure_module(self, module_name, module_args, task_vars=None):
                                    "run 'git submodule update --init --recursive' to correct this problem." % (module_name))
 
         # insert shared code and arguments into the module
-        (module_data, module_style, module_shebang) = modify_module(module_name, module_path, module_args, task_vars=task_vars, module_compression=self._play_context.module_compression)
+        (module_path, module_data, module_style, module_shebang) = modify_module(module_name, module_path, module_args, task_vars=task_vars, module_compression=self._play_context.module_compression)
```
abadger: And the last place where we'd change something if modify_module() does not need to return module_path.

lib/ansible/executor/module_common.py
sivel: that is a good point, I'll make that change

lib/ansible/executor/module_common.py
sivel: I'll do some testing, and look at how we are opening the file elsewhere.

lib/ansible/executor/module_common.py
sivel: Probably not.  I just need to ensure we have the path available to us at all places where it is needed.

abadger:
A few minor comments.  Everything else looks good.  If this is passing tests and you think it's ready to merge to devel then I'm okay with it.  Tests -- I'm not sure how long compiling the module takes but I expect it's minimal.  I think you're right, we probably should put golang into the docker images.

robinro:
I find it suboptimal to download go and and build stuff during the tests, while on the other hand we try preinstalling packages where possible to make travis faster. Maybe you can instead use precompiled binaries?

robinro:
This change broke my tests for the new images on opensuseleap and ubuntu16.04 due to missing with curl/wget (why use both?). 

There's something broken with the tests in general, because even the travis run that gave you a green light for the merge was done without curl installed:
https://travis-ci.org/ansible/ansible/jobs/129803359#L9053
No idea why it didn't fail.

robinro:
Ah, now I understand the idea, it uses either curl or wget. Still...

sivel:
We will be moving adding golang to the docker containers.  To get things in faster we didn't block this to wait on updating the images.  This is on the to do list.  We however will still do compilation at build, and will not be storing the binary files in the repo.

robinro:
Is there already a PR for golang?
There is an update to the dockerfiles by @gundalow queued up in https://github.com/ansible/ansible/pull/15692 which I depend on for https://github.com/ansible/ansible/pull/15936.

sivel:
It is on the todo list, meaning it still needs to be worked on.  There is no PR currently.

PR #13772, opened 2016-01-08
restructure vars_prompt and fix regression

### LINKED ISSUE #13770: Ansible 2.0 prompts for vars_prompt when --extra-vars is used
##### Issue Type

Bug Report
##### Ansible Version

2.0 from the "stable-2.0" branch, commit 91c7691a92e63580216e0e5c657b689361330beb

```
$ ansible --version
ansible 2.0.0
  config file =
```
##### Ansible Configuration

No changes
##### Environment

RHEL 7
##### Summary

Ansible 2.0 stops execution and requests the user to enter variables specified with `vars_prompt` even if the same variable has been specified on the comand line.

Ansible 1.9.4 does not prompt for `vars_prompt` if `--extra-vars` specifies the same variable.
###### playbook.yml

``` yaml

---
- hosts: all
  gather_facts: False
  vars_prompt:
    - name: x
      prompt: 'Enter value for x'
  tasks:
    - local_action: debug var=x
```

```
$ cat inventory
localhost

$ ansible-playbook -i inventory playbook.yml -e x=1
```
##### Expected Results:

Non-interactive run, as on 1.9.4:

```
$ ansible-playbook --version
ansible-playbook 1.9.4

$ ansible-playbook -i inventory playbook.yml -e x=foo

TASK: [debug var=x] ***********************************************************
ok: [localhost -> 127.0.0.1] => {
    "var": {
        "x": "foo"
    }
}

```
##### Actual Results:

The user is prompted:

```
$ ansible-playbook -i inventory playbook.yml -e x=foo
Enter value for x:
```

bcoca:
pushed it to use the existing propmpt from display and moved the vars prompt code there also for uniformity
changed vars_prompt to check extra vars vs the empty play.vars to restore 1.9 behaviour
sipmlified the code as it didn't need to check for syntax again (tqm is made none prior based on that)
fixes #13770

### DIFF
--- a/lib/ansible/utils/display.py
+++ b/lib/ansible/utils/display.py
@@ -267,13 +267,56 @@ class Display:
             self._errors[new_msg] = 1
 
     @staticmethod
-    def prompt(msg):
+    def prompt(msg, private=False):
         prompt_string = to_bytes(msg, encoding=Display._output_encoding())
         if sys.version_info >= (3,):
             # Convert back into text on python3.  We do this double conversion
             # to get rid of characters that are illegal in the user's locale
             prompt_string = to_unicode(prompt_string)
-        return input(prompt_string)
+
+        if private:
+            return getpass.getpass(msg)
+        else:
+            return input(prompt_string)
+
+    @classmethod
+    def do_var_prompt(self, varname, private=True, prompt=None, encrypt=None, confirm=False, salt_size=None, salt=None, default=None):
+
+        result = None
+        if sys.__stdin__.isatty():
+
+            do_prompt = self.prompt
+
+            if prompt and default is not None:
+                msg = "%s [%s]: " % (prompt, default)
+            elif prompt:
+                msg = "%s: " % prompt
+            else:
+                msg = 'input for %s: ' % varname
+
+            if confirm:
+                while True:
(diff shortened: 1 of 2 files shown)


lib/ansible/utils/display.py
```diff
@@ -267,13 +267,56 @@ def error(self, msg, wrap_text=True):
...
             prompt_string = to_unicode(prompt_string)
-        return input(prompt_string)
+
+        if private:
+            return getpass.getpass(msg)
+        else:
+            return input(prompt_string)
+
+    @classmethod
+    def do_var_prompt(self, varname, private=True, prompt=None, encrypt=None, confirm=False, salt_size=None, salt=None, default=None):
```
abadger, line 283: convention for the name of the first argument of a classmethod is "cls".  If the parameter isn't used at all, make it a @staticmethod instead and remove the parameter altogether.

bcoca, line 283: its used to reference the staticmethod 'prompt' in same class

abadger, line 283: Ah, I see it now.  yeah, so just change self to be cls to match convention when using `@classmethod` and it should be fine.

PR #13787, opened 2016-01-10
Support for playbook-level variables (new feature)

quinot:
Allow playbooks to contain vars, vars_files, and vars_prompt blocks, whose values are available to all plays in the playbook.

This allows prompted variables to be shared across plays, and also provides an alternative to defining variables in inventory (which makes playbooks less easy to read, as not all  information is kept at the same place, and does not support prompting.)

### DIFF
--- a/docsite/rst/playbooks_variables.rst
+++ b/docsite/rst/playbooks_variables.rst
@@ -75,13 +75,25 @@ See the :doc:`intro_inventory` document for multiple ways on how to define varia
 Variables Defined in a Playbook
 ```````````````````````````````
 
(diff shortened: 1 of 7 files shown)


hrstoyanov:
See http://stackoverflow.com/questions/33992153/ansible-playbook-wide-variable

quinot:
Apparently no interest from Ansible developers for the past 9 months, so closing.

skrobism:
By the number of forks, stars, & watches - you'd think Ansible might have shown interest.  This is such an obvious need.

PR #13793, opened 2016-01-11
Fixes #13763 Update connections _play_context on every iteration

### LINKED ISSUE #13763: Sudo/Become doesnt work on local connection in loops
I am having a confusing problem. It does look like a bug and hence I am reporting here, not sure if duplicate.

I am using ansible version 2.0.0 locally on OSX El Capitan, to run a series of steps which sets up my machine for development for my project. The last step here is to add a few lines to `/etc/hosts` (~15 lines)
I am using `lineinfile` to add these lines in a loop. 

I have tried `become: root` with `become_ask_pass=True` in the cfg file. This does nothing. It does not ask my password and once it hits that task, I get `The destination directory (/private/etc) is not writable by the current user.`. If I try `sudo: yes` and `ask_sudo_pass=True` then I get `fatal: [localhost]: FAILED! => {"failed": true, "msg": "ERROR! privilege output closed while waiting for password prompt:\nBECOME-SUCCESS-jpibjiakmfjdrjafspqizphqotvhjikm\n{\"msg\": \"line added\", \"changed\": true, \"backup\": \"\"}\n"}`. But infact If i check now, `/etc/hosts` file now has the first 2 lines of the list of lines I had to add. Running the playbook now doesnt help me at all.. but even more oddly, If i remove the config option to ask for password and run the play it works fully....

ktaragorn:
If this isnt updated, the _connection is reused, and thus has an outdated `_play_context`
This results in outdated `success_key` and `prompt` causing issues if sudo is run in a loop
Refer to the issue #13763 for more debugging and details

### DIFF
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -365,6 +365,9 @@ class TaskExecutor:
         if not self._connection or not getattr(self._connection, 'connected', False):
             self._connection = self._get_connection(variables=variables, templar=templar)
             self._connection.set_host_overrides(host=self._host)
+        #If connection is reused, its _play_context is no longer valid and needs to be replaced
+        #This fixes issues with tasks running sudo in a loop and having the success_key incorrect in the second iteration
+        self._connection._play_context = self._play_context
 
         self._handler = self._get_action_handler(connection=self._connection, templar=templar)
 


ktaragorn:
Thanks a lot to @drybjed for finding the commit causing this error

ktaragorn:
Connected Tickets - 
Fixes #13624 and #13763
IMHO a cleaner alternative to https://github.com/ansible/ansible/pull/13636, and displays a better understanding of the problem.

ktaragorn:
@drybjed  Isolated this problem to have been caused in commit  272778f which is where I chose to fix the issue

ktaragorn:
My ticket got a little noisy with my gradual understanding of the problem, so I would like to summarize it here

**Issue**\- If `connection: local` is used with `sudo: true` and `with_items` based looping, then task fails after the 2nd iteration of the loop with the error.

```
fatal: [localhost]: FAILED! => {"failed": true, "msg": "ERROR! privilege output closed while waiting for password prompt:\nBECOME-SUCCESS-jpibjiakmfjdrjafspqizphqotvhjikm\n{\"msg\": \"line added\", \"changed\": true, \"backup\": \"\"}\n"}
```

This occurs if `sudo` is configured to be more relaxed and only ask for password after a long timeout.
If `sudo` is setup to ask password every time, by following http://askubuntu.com/questions/636092/how-to-get-sudo-to-prompt-you-for-a-password-each-time/636095#636095, then the error becomes `ERROR! timeout waiting for privilege escalation password prompt` #13636 fails to handle this case.

**The cause** The `Action`, the `Connection` and the `TaskExecutor` objects all share a `play_context` object which stores among other things, a random prompt to verify successful `sudo` and a correct `sudo` prompt (`success_key` and `prompt` respectively). This object is recreated on every loop iteration. In the commit referenced, a change was added where the `connection` object and its `play_context` was reused for subsequent loop iterations. This meant that the `connection` now had outdated `play_context`. This results in 1 successful iteration, after which the 2nd iteration actually goes through but the check after that causes the task to fail. 

This understanding can be verified with the test play book provided in the ticket, where only 2 of the iterations of  the task happen, and to get the other ones to go through, `sudo` needs to be disabled.

Yannig:
The problem with this PR is that it breaks docker connector ...

jimi-c:
@Yannig how so?

Yannig:
If we update play_context, we get the following error:

```
EXEC ['/usr/bin/docker', 'exec', '-i', None, '/bin/sh', '-c', '( umask 22 && mkdir -p "$( echo $HOME/.ansible/tmp/ansible-tmp-1453131913.5-91802212177973 )" && echo "$( echo $HOME/.ansible/tmp/ansible-tmp-1453131913.5-91802212177973 )" )']                                                                         
```

The problem here come from '-i', None.

You can reproduce the problem using the following location: https://github.com/Yannig/ansible-issue-low-speed/tree/master/docker-example

Then run: 

```
ansible-playbook -i docker-example/hosts docker-example/create-docker.yml
```

followed by:

```
ansible-playbook -i docker-example/hosts docker-example/simple-test.yml
```

A better solution might be to remove _play_context from task executor and use only play_context from connection objects.

jimi-c:
Can't remove it, it gets templated via a post_validate() method (which is why it breaks in loops with re-used connections + delegating when the delegated info uses item data).

jimi-c:
@Yannig this patch corrects the issue with this patch and the docker connection (and I think it's more correct in general):

<pre>
index 9b7ac8b..ac4aa4f 100644
@@ -316,6 +316,11 @@ class TaskExecutor:
             # do the same kind of post validation step on it here before we use it.
             self._play_context.post_validate(templar=templar)
 
+            # now that the play context is finalized, if the remote_addr is not set
+            # default to using the host's address field as the remote address
+            if not self._play_context.remote_addr:
+                self._play_context.remote_addr = self._host.address
+
             # We also add "magic" variables back into the variables dict to make sure
             # a certain subset of variables exist.
             self._play_context.update_vars(variables)
@@ -547,9 +552,6 @@ class TaskExecutor:
         correct connection object from the list of connection plugins
         '''
 
-        if not self._play_context.remote_addr:
-            self._play_context.remote_addr = self._host.address
-
         if self._task.delegate_to is not None:
             # since we're delegating, we don't want to use interpreter values
             # which would have been set for the original target host
</pre>

I'm testing this further now, and will most likely merge it soon.

jimi-c:
Merged, along with my fix mentioned above.

Thanks!

PR #13803, opened 2016-01-11
new swiftype engine

docschick:
made new swiftype engine off http:// as suggested, replaced script, added a sitemap.xml file


bcoca:
cherry picked around recently removed searchbox.html

PR #13804, opened 2016-01-11
Fix typo in API docs

jeblair:
This typo caused a SyntaxError when attempting to run the example
in the API docs.


sivel:
Merged, thanks!

PR #13808, opened 2016-01-11
Added --ignore-certs option to ansible-galaxy init, install and info …

chouseknecht:
In the mailing list there is a thread (https://groups.google.com/forum/#!topic/ansible-project/L2H6KSAF_o0) of users having issues with ansible-galaxy behind a proxy. 

In my testing I can get it to work, but only with v2.0 using the --ignore-certs flag. The ignore_certs option gets passed to open_url() as validate_certs=False. This PR backports the same logic to stable-1.9.


abadger:
Merged to stable-1.9.  Thanks!

Issue #13419, opened 2015-12-03
ansible-galaxy install does not work in stable-2.0


rboudrie:
I have successfully installed Ansible 1.8.1 and 1.9 without difficulty.  When I attempt to install Ansible 2.0 on linux, I get the following:

```
lrmk225:~/shared-webstorage/storage/hdfssvc/playbooks/build # ansible-galaxy install -r requirements.txt -f
ERROR! Could not open readme: [Errno 2] No such file or directory: '/usr/local/lib64/python2.6/site-packages/ansible-2.0.0-py2.6.egg/ansible/galaxy/data/readme'
```

I tried creating an empty readme in this location, but that just shifts the error to:

```
lrmk225:~/shared-webstorage/storage/hdfssvc/playbooks/build # ansible-galaxy install -r requirements.txt -f
ERROR! Could not open metadata_template.j2: [Errno 2] No such file or directory: '/usr/local/lib64/python2.6/site-packages/ansible-2.0.0-py2.6.egg/ansible/galaxy/data/metadata_template.j2'
```

Is this a known issue?    This was reported to me by another individual in my organization, and I duplicated it on a different system.

rboudrie:
The "requirements.txt" file I used in my command contained one line:  
foobar_common.tar.gz,2.2.0.0.73760.718a938,foobar_common

(the product name was renamed to foobar for the sake of privacy)

bcoca:
how are you installing ansible?, those 2 files are in setup.py and should be included

rboudrie:
I suspect the problem is because I was installing Ansible 2.0 by building the package (the system I was using does not have repo access.

I used this same procedure to build Ansible 1.8.1 and 1.9 and did not have any problems.  Other users in my organization have reported the same problem.

git clone https://github.com/ansible/ansible.git
git checkout -b stable-2.0 origin/stable-2.0
git pull --rebase; git submodule update --init --recursive
cd ./ansible
source ./hacking/env-setup
make rpm
cd rpm-build/
tar -zxvf stable-2.0.tar.gz 
cd stable-2.0
python setup.py build
python setup.py install

NPrescott:
Experiencing the same using the recommended installation for OSX (via pip) on all ansible-galaxy commands:
- Python 2.7.9
- Version: 2.0.0
  
  ```
  > ansible-galaxy info
[...]

PR #13813, opened 2016-01-12
added galaxy data

sebastianneubauer:
not tested, but something like this seems to be missing in the Manifest.in. This can be a fix for #13419


sebastianneubauer:
just confirmed locally, this fixes #13419 in the ansible stable version 2.0, would be nice if it would be backported to the 2.0 and fix the released package on pypi...

PR #13817, opened 2016-01-12
Identify which module is missing the shebang

### LINKED ISSUE #13816: module not identified in "missing interpreter line" error message
lib/ansible/plugins/action/**init**.py includes this:

``` python
        if not shebang:
            raise AnsibleError("module is missing interpreter line")
```

This is not as helpful as it could be.  It would be better if this message identified the module in question.

MikeSpreitzer:
Fixes #13816


bcoca:
race condition hit, sorry, i solved already with almost exact same fix. It will be in 2.0.1 and 2.1 as they are released

MikeSpreitzer:
@bcoca: No problem, thanks for the quick work.  But I think this would be good to back-port to 1.9 also.

Issue #13110, opened 2015-11-10
[v2] TypeError: super(type, obj): obj must be an instance or subtype of type


mgedmin:
**Ansible Version**

```
ansible 2.0.0 (devel 09994c18f3) last updated 2015/11/10 08:56:11 (GMT +300)
  lib/ansible/modules/core: (detached HEAD fde6ee15d6) last updated 2015/11/10 09:01:29 (GMT +300)
  lib/ansible/modules/extras: (detached HEAD 6cc43bbd56) last updated 2015/11/10 09:01:29 (GMT +300)
```

**Summary**

My Vagrant test playbook failed with this error:

```
TASK [roundup : configure alias_database] **************************************
task path: /home/mg/src/deployments/provisioning/roles/roundup/tasks/main.yml:62
An exception occurred during task execution. The full traceback is:
Traceback (most recent call last):
  File "/home/mg/src/ansible/lib/ansible/executor/process/worker.py", line 134, in run
    shared_loader_obj,
  File "/home/mg/src/ansible/lib/ansible/executor/task_executor.py", line 118, in run
    res = self._execute()
  File "/home/mg/src/ansible/lib/ansible/executor/task_executor.py", line 381, in _execute
    result = self._handler.run(task_vars=variables)
  File "/home/mg/src/ansible/lib/ansible/plugins/action/normal.py", line 29, in run
    results = super(ActionModule, self).run(tmp, task_vars)
TypeError: super(type, obj): obj must be an instance or subtype of type

fatal: [trusty]: FAILED! => {"failed": true, "stdout": ""}
```

This playbook works with Ansible 1.9.4.

I'm using a [custom module](https://gist.github.com/mgedmin/5f8ac034df0c371444be):

```
- name: configure alias_database
  postfix: name=alias_database value="hash:/etc/aliases, hash:/etc/roundup/aliases"
  notify: postfix reload
  tags: mail
```

The module has a custom action plugin, needed to support --diff on Ansible 1.x, but this doesn't do anything on v2 -- just re-imports ActionModule from ansible.plugins.action.normal.

I'll try to reproduce with a smaller example.

mgedmin:
I'm having trouble reproducing this with a small example.

Instead I added some debugging prints

``` diff
diff --git a/lib/ansible/plugins/action/normal.py b/lib/ansible/plugins/action/normal.py
index bf93fda..2ea942f 100644
--- a/lib/ansible/plugins/action/normal.py
[...]

PR #13823, opened 2016-01-12
Fix module loading in multi-threading environments

### LINKED ISSUE #13822: Document that Ansible API is not thread-safe
When ansible loads a module from disk for a plugin, it first checks
to see if the module has already been loaded (i.e., exists in
sys.modules), and if it does, it returns that reference rather than
loading the module again (which, as described in issue #13110, must
not happen because repeating the module load may re-initialize
module global variables).    

When a module is loaded, the reference is immediately added to  
sys.modules, even before the module initialization is complete. 
Therefore, if two threads race the module loading code, the first
will begin loading the module from disk, and the second will see  
the module in sys.modules.  It might begin using that module before
it is even fully loaded, resulting in AttributeErrors on objects 
which should be defined in the module.

This can be reproduced by wrapping the Python API example at http://docs.ansible.com/ansible/developing_api.html in a function and then adding the following:

``` python
import threading
t1 = threading.Thread(target=test)
t1.start()
t2 = threading.Thread(target=test)
t2.start()
t1.join()
t2.join()
```

This results in the this traceback:

> Traceback (most recent call last):
>   File "/usr/lib/python2.7/threading.py", line 810, in **bootstrap_inner
>     self.run()
>   File "/usr/lib/python2.7/threading.py", line 763, in run
>     self.__target(_self.__args, *_self.__kwargs)
>   File "test.py", line 13, in test
>     variable_manager = VariableManager()
>   File "lib/ansible/vars/__init**.py", line 92, in **init**
>     self._fact_cache = FactCache()
>   File "lib/ansible/plugins/cache/__init__.py", line 35, in **init**
>     self._plugin = cache_loader.get(C.CACHE_PLUGIN)
>   File "lib/ansible/plugins/__init__.py", line 335, in get
>     obj = getattr(self._module_cache[path], self.class_name)(_args, *_kwargs)
> AttributeError: 'module' object has no attribute 'CacheModule'

The problem can be corrected by wrapping module loading in a mutex.

jeblair:
When ansible loads a module from disk for a plugin, it first checks
to see if the module has already been loaded (i.e., exists in
sys.modules), and if it does, it returns that reference rather than
loading the module again (which, as described in issue #13110, must
not happen because repeating the module load may re-initialize
module global variables).

When a module is loaded, the reference is immediately added to
sys.modules, even before the module initialization is complete.
Therefore, if two threads race the module loading code, the first
will begin loading the module from disk, and the second will see
the module in sys.modules.  It might begin using that module before
it is even fully loaded, resulting in AttributeErrors on objects
which should be defined in the module.

This change wraps the method which checks for existence in
sys.modules and loads modules from disk in a global mutex so that
more than one module is never loaded from disk at a time.

Fixes #13822

### DIFF
--- a/lib/ansible/plugins/__init__.py
+++ b/lib/ansible/plugins/__init__.py
@@ -26,6 +26,7 @@ import inspect
 import os
 import os.path
 import sys
+import threading
 import warnings
 
 from collections import defaultdict
@@ -42,6 +43,7 @@ except ImportError:
 MODULE_CACHE = {}
 PATH_CACHE = {}
 PLUGIN_PATH_CACHE = {}
+MODULE_LOADER_LOCK = threading.Lock()
 
 def get_all_plugin_loaders():
     return [(name, obj) for (name, obj) in inspect.getmembers(sys.modules[__name__]) if isinstance(obj, PluginLoader)]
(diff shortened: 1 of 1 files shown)


jimi-c:
Ansible's API has never been advertised as being thread-safe, as we use the multiprocessing library (which does everything in forks). This patch may fix the module loading issue, however I'm sure there are many other places in the code which may harbor similar race conditions. As such, I'm not particularly opposed to this patch, however I am against trying to make the entire API thread safe.

bcoca:
-1, adding threading and locking code we do not use adds complexity, performance degradation and possible pitfalls for no gain

emonty:
@jimi-c @bcoca would you accept a patch to the documentation then that says "the python API is known to not be thread safe?" I think that might have prevented confusion in this case.

jimi-c:
@emonty absolutely.

bcoca:
done 2374abbeda465c0aded55fabc18085a1530c9377

PR #13839, opened 2016-01-13
Include all inventory group_vars and host_vars

willthames:
Ensure results from all basedirs are included, and
not just overwritten by the results from the last
one


bcoca:
we should be using combine_vars function (to observe the hash merge setting)

bcoca:
can you show an example of this breaking?

bcoca:
@willthames can you check if my PR fixes your issue also?

willthames:
Closing in favour or #13902 which does indeed pass my tests.

Issue #7298, opened 2014-05-06
Feature: Extended Vault support: copying files, lookup plugin


kolypto:
When there's a need to keep some sensitive data close to a playbook, Vault is great for ciphering variables. However, what about encrypting files that are transferred to the target host?

Suggesting the following:
- Decrypt Vault-protected files on-the-fly with the `copy` module
- Implement a lookup plugin to read data from Vault-protected files, like `lookup('vault', 'path/to/file')`

ruudk:
This would be really great to have. A simpel `decrypt=yes` flag for the `copy` module is all I need :)

bcoca:
this could be provided by #8110

AnneTheAgile:
@bcoca , do I understand correctly that your PR #8110  Lookup vault file, aka #8533 , does not decrypt yet, but that you could add the feature? From your docs there it seems like the lookup will simply show the fact that a file is vaulted. ty!

bcoca:
no, it is supposed to decrypt it if it dectects it is encrypted and the
password is available.​

it does add the facility for other lookups to do the same but I did not
implement the decryption outside the file lookup.

rouge8:
Is there any plan to update #8533 to resolve this?

soichih:
+1 for "Decrypt Vault-protected files on-the-fly with the copy module" option.

thedodd:
+1 for a simple `decrypt=yes` flag for the `copy` module.

flmmartins:
+1 to previous comment!

dwt:
There seem to be some pull requests that provide this functionality for lookup() - still what I really want is to be able to use vault encrypted files transparently everywhere in ansible - so I think this pull request provides something unique.

See:
- https://github.com/ansible/ansible/pull/8533
- https://github.com/ansible/ansible/pull/8110

mcreenan:
Please consider adding this functionality.  Doing it other ways ends up being very convoluted.

brandonheller:
This would be great - without this, I have to write a script to copy from the vault, so that copy can do its thing.

afeinberg:
+1 for this, please

flmmartins:
This issue is making one year old! Please add this! ^^

ant8e:
+1 
I'm just starting to use Vault and was very surprised not to be able do crypt arbitrary files

mauricios:
[...]

PR #13849, opened 2016-01-13
feature/copy-vault: Add vault support to copy action plugin

cambell-prince:
This PR allows the copy task to support copying encrypted vault files as described in #7298 (at least with respect to the copying of files).  The implementation is the same as that used in the [copyv action plugin](https://github.com/saygoweb/ansible-plugin-copyv) developed for Ansible v1.9.2


bcoca:
nice, this was on my wishlist, but I was going to approach it as a general function available to all action plugins, not just copy.

cambell-prince:
@bcoca This feature seems like a 'data interpreting' kind of feature, so using the loader which is passed around all plugins seems like an ok option.  If you're ok with that approach, this PR could be improved by hiding the vault implementation within the loader and having a public method or two on the loader that would implement this functionality.  The loader is then available as a general service to all action plugins.

bcoca:
I was debating doing it that way or adding vaullib as an additional thing to pass through, this is a bit tangential to loader (which needs to be vault aware also) but  it is also very convenient there ....

cambell-prince:
I've had a look at playbook.py, and to pass an instance of VaultLib through to action plugins would seem to require that the VaultLib is constructed in playbook.py and passed through via yet another parameter in the PlayExecutor and made available in the PlayContext.  However, the PlayExecutor is already passed in a passwords dictionary.  These passwords are stored in the PlayContext which is already available to all the action plugins.  I'm thinking that it would be ok to add the vault password to the password dictionary that is already passed to the PlayExecutor and available in the PlayContext.

If that's an acceptable way to go, let me know and I'll implement that and post a new PR.

bcoca:
yes, play_context seems the 'right place' to put the vault password and i'm thinking it should also work for the vaultlib instance if we don't want to overload data loader. let me sync with other core team members before you do any work to make sure we all agree.

datamint:
Looking forward for this to be merged. Would this work also with unarchive? Vaulted files should really be transparent once running a playbook, copying and unzipping seem like the most common scenarion and will cut a good 30% of a playbook I'm writing.

namero999:
Looking forward for this to be merged. Would this work also with unarchive? Vaulted files should really be transparent once running a playbook, copying and unzipping seem like the most common scenarios and would cut a good 30% of a playbook I'm writing.

My specific use case is that I have a folder with a nested structure which contains certificates and sensible configuration files. Therefore the best thing for me is to tar/gz the whole folder, encrypt it with vault, and then unarchive it on the remote hosts. The current workaround is really cumbersome, so this feature (with unarchive support) would be amazing.

bcoca:
@cambell-prince after some discussion, we are fine on having a new method in data loader that deals with transparently decrypting files, something action/lookup plugins could use.

`get_file_path(path)` +> returns original path or path to temp file of decrypted file (i suck at namging, just a bad example)

bcoca:
closing this in favor of #14079

PR #13853, opened 2016-01-13
Add posix groups to the list of returned facts

mscherer:


bcoca:
we already have the getent module doing this, do we really need to add this to the standard facts?

mscherer:
Indeed, didn't see this module. I personally prefer when I do not need to run extra modules for getting facts, but yeah, maybe that's not needed...

hloeffler:
wasn't there the idea of having "get_facts" modular/configurable?
I don’t want it any slower than now.

bcoca:
So someone brought up that on some systems with many users/groups this can take along time.

I would either defer to the getent module or add this as a modular option to fact gathering, off by default

mscherer:
Let's do that with getent, closing the PR

PR #13352, opened 2015-11-29
add plugin for uncompress module

Jmainguy:

PR #13869, opened 2016-01-13
Unarchrefactor

pileofrogs:
This is the plugin side of a refactor to allow untar and unzip (and any future unarchivers) to have their own modules, thus allowing their own arguments.  There will be a corresponding PR for ansible-modules-core in just a moment.  Don't accept one without the other!

This PR only refactors and does not add new arguments.  

After doing it, I realized a lot of ways to make it better.  I added a _copy method to the ActionBase class, thinking this would be good because other future modules could use it.  However, it might be better to make an unarchive subclass of ActionBase and stick stuff in thier.  Also, I wound up duplicating a fair amount of code between the unarchive module and the untar and unzip modules.  I would like to separate out some of the logic.  Possibly into ansible.module_utils somewhere... sigh... 

So, basically, if you aren't picky, take this PR.  It should allow folks who want to submit PRs for new tar/zip features to be able to do so.  If you're even remotely picky, fire it back with comments and I'll unscrew whatever parts you don't like.

There will be a nearly identical sounding PR for ansible-modules-core in just a moment.  One doesn't work without the other.

### DIFF
--- a/lib/ansible/plugins/action/__init__.py
+++ b/lib/ansible/plugins/action/__init__.py
@@ -674,6 +674,4 @@ class ActionBase(with_metaclass(ABCMeta, object)):
             ),
         )
 
-    # execute the unarchive module now, with the updated args
-    result.update(self._execute_module(module_args=new_module_args, task_vars=task_vars))
     return result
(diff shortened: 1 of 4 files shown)


pileofrogs:
Hi all. Anyone out there have anything to say at all about these PRs? There are other PRs for unarchive waiting and I can't really get to them until I know if this PR is going in the right direction or not.

Jmainguy:
Looks like I did something similar trying to create the uncompress module.

https://github.com/ansible/ansible/pull/13352

I think getting most of this code into a commong module_utils file we could share between these multiple modules could be very beneficial (atleast the shared code I am duplicating). Other than that, looks good.

pileofrogs:
Lack of interest

Issue #13873, opened 2016-01-14
Ansible 2.0 hostgroup vars issues


retr0h:
I’m using a [custom vars plugin](https://github.com/blueboxgroup/ursula/blob/master/plugins/vars/default_vars.py) under ansible 1.9.  However, it looks as though the host object which [run](https://github.com/ansible/ansible/blob/4ba7158282f148c90c72f824d6ebcd1a9953b580/lib/ansible/inventory/vars_plugins/noop.py#L38) is passed is missing the group/host_vars from inventory.

I have tracked this down to what appears to be a bug with [_get_hostgroup_vars](https://github.com/ansible/ansible/blob/devel/lib/ansible/inventory/__init__.py#L702).

`basedirs = [_basedir, self._playbook_basedir]` is a list containing the path to my inventory as specified by `-i` and None (since playbook_basedir is False).  Since the loop is iterated before returning to the caller, the None directory fails to load inventory and returns an empty dict to the caller.

Ansible 1.9 was retaining the previous [result](https://github.com/ansible/ansible/blob/stable-1.9/lib/ansible/inventory/__init__.py#L651), each time through the loop.

When removing None items from the `basedirs` list, the plugin works as expected.

```
    def _get_hostgroup_vars(self, host=None, group=None, new_pb_basedir=False):
        """
        Loads variables from group_vars/<groupname> and host_vars/<hostname> in directories parallel
        to the inventory base directory or in the same directory as the playbook.  Variables in the playbook
        dir will win over the inventory dir if files are in both.
        """

        results = {}
        scan_pass = 0
        _basedir = self.basedir()

        # look in both the inventory base directory and the playbook base directory
        # unless we do an update for a new playbook base dir
        if not new_pb_basedir:
            basedirs = [_basedir, self._playbook_basedir] # <--- None at end of list here causes the problem.
        else:
            basedirs = [self._playbook_basedir]

        for basedir in basedirs:
            # this can happen from particular API usages, particularly if not run
            # from /usr/bin/ansible-playbook
            if basedir in ('', None):
                basedir = './'
```

bcoca:
# List Information

Hi!

Thanks very much for your interest in Ansible.  It sincerely means a lot to us. 

This appears to be a user question, and we'd like to direct these kinds of things to either the mailing list or the IRC channel.
- IRC: #ansible on irc.freenode.net  
- mailing list: https://groups.google.com/forum/#!forum/ansible-project

[...]

PR #13897, opened 2016-01-14
Allow _get_hostgroup_vars to properly return

retr0h:
When iterating over `basedirs`, and the playbook base directory
does not contain inventory, an empty dict is returned to the caller
instead of the inventory found.

For further details see the bug report #13873.


retr0h:
Closing.  Looks like @bcoca release a bug fix at @ #13902.

PR #13841, opened 2016-01-13
fix add_host overwrites runtime host vars

yfried:
fix ansible/ansible-modules-core#2799

PR #13901, opened 2016-01-14
fix issue with add_hosts overwriting existing vars

bcoca:
fixes http://github.com/ansible/ansible-modules-core/issues/2799 and is alternative to #13841


bcoca:
@yfried-redhat, can you verify this fixes it for you?

yfried:
works for me with latest ansible from source

PR #13902, opened 2016-01-14
overwrite/merge not remove prev inventory data

bcoca:
alternate to #13839


willthames:
@bcoca, ansible-inventory-grapher tests succeed with this PR

PR #13933, opened 2016-01-16
Add correct default options for sudo_flags

dlangille:
see https://github.com/ansible/ansible/blob/devel/lib/ansible/constants.py#L181

### DIFF
--- a/docsite/rst/intro_configuration.rst
+++ b/docsite/rst/intro_configuration.rst
@@ -587,9 +587,10 @@ the sudo implementation is matching CLI flags with the standard sudo::
 sudo_flags
 ==========
 
-Additional flags to pass to sudo when engaging sudo support.  The default is '-H' which preserves the $HOME environment variable
-of the original user.  In some situations you may wish to add or remove flags, but in general most users
-will not need to change this setting::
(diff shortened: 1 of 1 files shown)


dossy:
I would also adjust the text on line 590:

> The default is '-H' which [...]

Since the docs explain what the `-H` option does, I would probably add explanations for what `-S` an `-n` do, as well, to be consistent.

dlangille:
How does this read?

Additional flags to pass to sudo when engaging sudo support.  The default is '-H -S -n' which preserves the environment
of the original user, prompts for passwords via STDIN, and avoids prompting the user for input of any kind.  Note that '-n' will conflict with using password-less sudo auth, such as pam_ssh_agent_auth. In some
situations you may wish to add or remote flags, but in general most users
will not need to change this setting::

dossy:
@dlangille - I like the addition of the guidance around the effect `-n` will have.

However, `-H` does not preserve the environment of the original user.  That is `-E`.  The `-H` option sets the `$HOME` environment variable from passwd(5) according to the `sudo` manpage:

> The -H (HOME) option sets the HOME environment variable to the homedir of the target user (root by default) as specified in passwd(5). [...]

dlangille:
See https://github.com/ansible/ansible/pull/13934

PR #13934, opened 2016-01-16
Specify the correct default options for sudo_flags

dlangille:
The correct default options for sudo_flags can be found at: https://github.com/ansible/ansible/blob/devel/lib/ansible/constants.py#L181

Slightly alter explanation of '-H' so as not to confuse it with -E, --preserve-env (which preserves existing environment variables).

When adding the two other options, include short explanations of those options.

Add note about '-n', which did not appear in 1.x I believe, and which bit me.


bcoca:
just note that these were used in 1.x but hardcoded into the command, now you have the option to remove them (some versions of sudo did not deal with them well).

PR #13939, opened 2016-01-17
Enforce `state='file'` in `copy` module

### LINKED ISSUE #10834: copy fails when dest is hardlink
##### ISSUE TYPE
Bug Report

##### COMPONENT NAME
core

##### ANSIBLE VERSION
1.9.0.1

##### SUMMARY
copy modules fails with message
`msg: absolute paths are required`
when I try to copy file over existing hardlink

task I'm trying to run looks like this:

```
  copy:
    dest: '/path/to/existing/hardlink'
    src: '/path/to/source/file'
    owner: 'root'
    group: 'root'
    mode: '0644'
```

I'm running version 1.9.0.1

semekh:
This was causing wrong behaviour when `prev_state` was `hard`-link,
since the `file` module tried to apply the same `state` on the new
file, causing unexpected errors.

Fixes ansible/ansible#10834.
Depends on ansible/ansible-modules-core#2827.

### DIFF
--- a/lib/ansible/plugins/action/copy.py
+++ b/lib/ansible/plugins/action/copy.py
@@ -257,7 +257,8 @@ class ActionModule(ActionBase):
                     dict(
                         src=source_rel,
                         dest=dest,
-                        original_basename=source_rel
+                        original_basename=source_rel,
+                        state='file'
                     )
                 )
 
--- a/test/integration/roles/test_copy/tasks/main.yml
+++ b/test/integration/roles/test_copy/tasks/main.yml
@@ -63,6 +63,12 @@
       - "copy_result.md5sum == 'c47397529fe81ab62ba3f85e9f4c71f2'"
   when: ansible_fips != True
 
+- name: create a hardlink to the file
+  file: src={{ output_file }} dest={{ output_dir }}/hard.lnk state=hard
+
+- name: copy the same file to the same place
+  copy: src=foo.txt dest={{output_file}} mode=0444
+
 - name: check the stat results of the file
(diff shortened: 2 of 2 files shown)


sivel:
It appears as though your commit includes a change in the core submodules commit sha.  This should not be changed.

semekh:
It does. This PR depends on ansible/ansible-modules-core#2827 and that's the reason the submodule sha has changed.

semekh:
I'd like to mention that since both `overlay` and `devicemapper` storage drivers in docker use hardlinks to share files between layers, this causes most ansible playbooks to fail when working with files from layers below, rendering ansible useless for a lot of docker container use cases.

sivel:
# Rebase Request

Hi!

- You should be able to run `git rebase` against this branch to bring it up to date
- Make sure to reset your git submodules using `git submodule update`

abadger:
@SeMeKh Pardon me, I've been looking at this for a few days and I think it's a good change (I'm also asking another core dev to take a look in case ther eare some wrinkles I'm not seeing... hope to get his input by tomorrow).  I have one question, though:  You link to this change to the file module: https://github.com/ansible/ansible-modules-core/pull/2827 but i'm not sure what that fixes... It looks like a good change to me (the prev_state refers to the previous state of the src file so if we change what the source file is then it makes sense to take another look at what prev_state should be) but I can't figure out what is broken when that change is not applied.  (The test case for this bug is solved just by applying the change here, the file module change is not needed).

Can you give me a test case for that as well?

PR #13958, opened 2016-01-18
Change Solaris packaging module from pkg to pkg5

liquidat:
The Solaris package managing module in Ansible managing packages via the program '/usr/bin/pkg' is named "pkg5", not "pkg": http://docs.ansible.com/ansible/pkg5_module.html

Additionally, there is no packaging module in Ansible named 'pkg'.


bcoca:
that change does not affect only solaris

liquidat:
Ah, ok. I've searched for other package operating systems for /usr/bin/pkg and only found Solaris so I was missing the others.
But ok, than we can't do anything about it, and the tip from http://slideshare.net/bcoca/ansible-tips-tricks, slide 23 with {{ ansible_pkg_mgr }} will not be possible with Solaris machines. Thanks for checking the pull request anyway.

PR #13967, opened 2016-01-18
Return unicode string in _dump_results method + typo

### LINKED ISSUE #13964: Bug report: unicode error with include + bad args with unicode
##### Ansible Version:

Devel
##### Ansible Configuration:

You can reproduce the issue using the following playbook:

```
ansible-playbook error-parsing/error-parsing-args.yml
```

You can retrieve it at the following location: https://github.com/Yannig/ansible-issue-low-speed/tree/master/error-parsing
##### Environment:

Ubuntu
##### Summary:

Ansible v2 get a unicode error when parsing bad formated module.
##### Steps To Reproduce:

ansible-playbook error-parsing/error-parsing-args.yml
##### Expected Results:

```
fatal: [localhost]: FAILED! => {"failed": true, "reason": "ERROR! this task 'file' has extra params, which is only allowed in the following modules: command, shell, script, include, include_vars, add_host, group_by, set_fact, raw, meta\n\nThe error appears to have been in '/home/yannig/dev/yannig/ansible-issue-low-speed/error-parsing/error-parsing-args-with-include.yml': line 3, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: \"Something with unicode héhé and a bad args\"\n  ^ here\n"}
```
##### Actual Results:

```
[WARNING]: Error when using <bound method CallbackModule.v2_runner_on_failed of
</home/yannig/dev/yannig/ansible/lib/ansible/plugins/callback/default.CallbackModule object at 0x7f66ef384f90>>: 'ascii' codec can't decode byte 0xc3 in
position 510: ordinal not in range(128)
```

Yannig:
Fix https://github.com/ansible/ansible/issues/13964

### DIFF
--- a/lib/ansible/plugins/callback/__init__.py
+++ b/lib/ansible/plugins/callback/__init__.py
@@ -81,14 +81,14 @@ class CallbackBase:
         if not indent and '_ansible_verbose_always' in result and result['_ansible_verbose_always']:
             indent = 4
 
-        # All result keys stating with _ansible_ are internal, so remove them from the result before we output anything.
+        # All result keys starting with _ansible_ are internal, so remove them from the result before we output anything.
         abridged_result = strip_internal_keys(result)
 
         # remove invocation unless specifically wanting it
         if not keep_invocation and self._display.verbosity < 3 and 'invocation' in result:
             del abridged_result['invocation']
 
-        return json.dumps(abridged_result, indent=indent, ensure_ascii=False, sort_keys=sort_keys)
+        return to_unicode(json.dumps(abridged_result, indent=indent, ensure_ascii=False, sort_keys=sort_keys))
 
     def _handle_warnings(self, res):
         ''' display warnings, if enabled and any exist in the result '''


bcoca:
you should not need to do this here as display turns everything into unicode

Yannig:
Well, if you use unicode string with bad element, you get the following error:

```
[WARNING]: Error when using <bound method CallbackModule.v2_runner_on_failed of
</home/yannig/dev/yannig/ansible/lib/ansible/plugins/callback/default.CallbackModule object at 0x7f66ef384f90>>: 'ascii' codec can't decode byte 0xc3 in
position 510: ordinal not in range(128)
```

abadger:
The right thing to do with unicode issues is to fix them at the source of the data rather than the usage of the data.  I've tracked this particular one back to the strategy plugins and will fix it with this:

``` python
-                tr = TaskResult(host=host, task=included_file._task, return_data=dict(failed=True, reason=str(e)))
+                tr = TaskResult(host=host, task=included_file._task, return_data=dict(failed=True, reason=to_unicode(e)))
```

There's probably other places in the code where we're using `str()` to transform an object into a string that need to be reviewed to decide if they should be using `to_unicode()` (almost all uses) or `to_str()` (for things that are being fed _into_ an exception message.  ie:

```
except Exception, e:
    raise AnsibleError(to_str(e))
```

PR #13968, opened 2016-01-18
synchronize: First cut at not doing sudo on the control machine but on the remote machine instead.

abadger:
In 2.0.0.x become wa sreversed for synchronize.  Happening on the local machine instead of the remote machine.  This restores the ansible-1.9.x behaviour.  it does become on the remote machine.  However, there's aspects of this that are hacky (no hackier than ansible-1.9 but not using 2.0 features).  The big problem is that it does not understand any become method except sudo.  I'm willing to use a partial fix now because we don't want people to get used to the reversed semantics in their playbooks.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -1788,7 +1788,9 @@ class AnsibleModule(object):
         elif isinstance(args, basestring) and use_unsafe_shell:
             shell = True
         elif isinstance(args, basestring):
-            args = shlex.split(args.encode('utf-8'))
+            if isinstance(args, unicode):
+                args = args.encode('utf-8')
+            args = shlex.split(args)
         else:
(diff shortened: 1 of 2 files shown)


abadger:
Update: A lot of testing this week and a few updates to the code to address more cornercases.   The latest version of the code also needs this branch merged to the ansible-modules-core repository: https://github.com/ansible/ansible-modules-core/compare/synchronize-become-is-reversed?expand=.  

The playbooks I have assembled with test cases: https://gist.github.com/abadger/0fc9e36a295979210698   Note that these are not in a form suitable for automated testing yet - Need to manually check after each task if the desird/expected behaviour has been achieved (but that behaviour is mentioned in the comment of each task.)

abadger:
Gone through my battery of tests with both the ssh and paramiko connections and everything is as expected now.  i'll be merging this soon unless anyone finds a problem.

Issue #13956, opened 2016-01-18
Performance issues with over 9000 groups


towolf:
We have ~~over 9000~~ 8500 groups coming out of a dynamic inventory script and formatted as JSON. Assigned to those groups are about half that number of unique hosts. This means hosts are member of several groups at once. The inventory data coming from an internal database is basically key-value property pairs made into group identifiers:

``` json
   "host.os=Debian_6.0" : {
      "hosts" : [
         "127.0.xxx.xxx",
         "127.0.xxx.xxx"
      ]
   },
   "host.os=FreeBSD" : {
      "hosts" : [
         "127.0.xxx.xxx",
         "127.0.xxx.xxx",
       [ ... lots more ...]
         "127.0.xxx.xxx",
         "127.0.xxx.xxx"
      ]
   },
   "host.os=FreeBSD_10.1" : {
      "hosts" : [
         "127.0.xxx.xxx",
         "127.0.xxx.xxx"
      ]
   },
   "host.os=FreeBSD_7.3" : {
      "hosts" : [
         "127.0.xxx.xxx"
      ]
   },

[... and so on ...]
```

This scheme facilitates playbook host patterns like:

``` yaml
    hosts: host.os=Debian_6.0:!host.os=FreeBSD:&status=productive
```

So we have a lot of these kind of groups. 7500 groups contain only one host and
a few groups contain thousands. We also include a `_meta` object in the JSON
which supplies `hostvars` for each host with up to 30 host specific variables.

**The problem is that parsing and sorting through our inventory takes over 20 seconds.**

I have tried to work out the hot spots during parsing this large inventory and
identified a few locations.
- Looking at the file system and checking for  `group_vars/*` files for several
  relative paths and for every group and in multiple passes is very expensive.
  
  Even an `os.path.realpath` (which serves as a fancy string concatenation) has
  to stat 35000 times and takes more than 5 seconds. Not to mention what
  happens inside the `add_group_vars_file()` function.
  
  ```
  Total time: 27.3202 s
  File: /home/towolf/src/local/ansible.upstream/lib/ansible/inventory/__init__.py
  Function: _get_hostgroup_vars at line 702
  
  Line #      Hits         Time  Per Hit   % Time  Line Contents
  ==============================================================
     702                                               @profile
     703                                               def _get_hostgroup_vars(self, host=None, group=None, new_pb_basedir=False):
     704                                                   """
[...]

PR #13993, opened 2016-01-19
Avoid recursively checking JSON inventory for Unicode

towolf:
Both python-json and simplejson always return unicode strings when using
their loads() method on unicode strings. This is true at least since
2009. This makes checkign each substring unnecessary.

This commit makes parsing of large dynamic inventory at least 2 seconds
faster.

Related to issue #13956

cf: https://github.com/towolf/ansible-large-inventory-testcase

### DIFF
--- a/lib/ansible/inventory/script.py
+++ b/lib/ansible/inventory/script.py
@@ -31,7 +31,7 @@ from ansible.errors import AnsibleError
 from ansible.inventory.host import Host
 from ansible.inventory.group import Group
 from ansible.module_utils.basic import json_dict_bytes_to_unicode
-from ansible.utils.unicode import to_str
+from ansible.utils.unicode import to_str, to_unicode
 
 
 class InventoryScript:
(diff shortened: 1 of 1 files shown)


abadger:
This looks generally good but I have the feeling we should be transforming even higher up.  What about doing it at these two places instead:

https://github.com/ansible/ansible/pull/13993/files#diff-a4d695beaa410b22bfeb22749cdaa210R61
https://github.com/ansible/ansible/pull/13993/files#diff-a4d695beaa410b22bfeb22749cdaa210L64

Also use the to_unicode() function from ansible.utils.unicode.

Not sure if we want to error out if we encounter non-utf8 input.  For stdout I guess we do since it is supposed to be proper json (and json is always supposed to be encoded utf-8 AFAIK).  So you can do self.data = to_unicode(stdout, errors='strict')  Probably need to catch the exception if that fails to decode and give an error message similar to the ones in _parse()

For stderr, I think we're only using that to construct an error message so we don't need to throw an error (we're already inside of an error condition).  so probably using to_unicode(stderr)   (which does errors='replace' by default) is fine there.

towolf:
Moved the conversion to the place you suggested.

jimi-c:
This has my +1, just want @abadger to give it a review as well.

abadger:
+1 from me.  @towolf Would you be comfortable squashing the commits? https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request#squash-your-changes  If not one of us can do that before we merge.  (just trying to spread the work around ;-)

towolf:
Done.

abadger:
Merged to devel and stable-2.0.  It should be in the 2.0.1 release.  Thanks!

PR #13995, opened 2016-01-19
Prefer delegate when getting variables from inventory

### LINKED ISSUE #13994: [1.9.2] inventory lookup failure with accelerate transport when ansible_ssh_host uses IPs
The inventory lookup in the accelerate plugin relies on the correct host name for the inventory being passed to the plugin via the host attribute. However when using ansible_ssh_host and pointing to an IP address, this will result in the call at https://github.com/ansible/ansible/blob/b24daecd6e78df13f8685e61699ea0f2a15a3c77/lib/ansible/runner/connection_plugins/accelerate.py#L105 trying to perform inventory lookup with the IP address provided instead of the inventory host name.

Since the runner will set the `delegate` attribute on the connection if the host name is different to the actual_host, a simple fix is to lookup this attribute first and fallback to host if not set.

electrofelix:
Use self.delegate in preference to self.host in the accelerate
connection plugin for retrieval of variables from inventory.

Without this change setting ansible_ssh_host to an IP address, via the
inventory and configuring transport to accelerate via the configuration
will result in an exception message "host not found: <IP>" on
attempting to connect.

When ansible_ssh_host is defined for a host, the runner will pass
this value, which may be an IP address, to the connection plugin as the
value for 'host' to the connector class connect method, and eventually
it will be used for the host attribute on the requested connection
plugin. Consequently, within the Connnection plugin, using the 'host'
attribute for inventory lookup may try using an IP address instead of
the ansible inventory host name.

To handle this case, Ansible runner will check that the whether the
host is being provided (ansible_ssh_host) is the same as the inventory
host name, and when different, it will set an additional attribute
'delegate' on the connection plugin instance. Switching to use
attribute, if not None, and falling back to host, which would have the
correct value if delegate is not set, ensures inventory lookup uses the
correct host.

Fixes #13994

### DIFF
--- a/lib/ansible/runner/connection_plugins/accelerate.py
+++ b/lib/ansible/runner/connection_plugins/accelerate.py
@@ -51,6 +51,7 @@ class Connection(object):
         self.is_connected = False
         self.has_pipelining = False
         self.become_methods_supported=['sudo']
+        self.delegate = None
 
         if not self.port:
             self.port = constants.DEFAULT_REMOTE_PORT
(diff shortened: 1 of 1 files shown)


sivel:
~~errant request for rebase~~

sivel:
@electrofelix apologies on that last comment, I'm going to get rid of that. My inspection was flawed and didn't account for a submission against, stable-1.9

bcoca:
sorry but we are going to pass on this change, there are several issues with scope, setting and templating of delegate. As this is fixed in 2.x and we expect that 1.9.5 is the last release of this tree we don't expect this issue to be fixed.

bcoca:
So i believe #15024 should fix this and apply the correct vars

PR #14000, opened 2016-01-19
Conda recipe to create conda package for Ansible v1.9.4

ijstokes:


bcoca:
this should all be in the packaging/ subdir

bcoca:
sorry, we are going to pass on this, we are not adding new features to the 1.9 tree

PR #14008, opened 2016-01-20
initial add of action plugin net_template

privateip:
The net_config local action handles templating for network configuration
file.  It will also allow network device configurations to be backed up
to the control host

### DIFF
--- a/lib/ansible/plugins/action/net_template.py
+++ b/lib/ansible/plugins/action/net_template.py
@@ -1,0 +1,93 @@
+#
+# Copyright 2015 Peter Sprygada <psprygada@ansible.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


bcoca:
I'm not sure we want this here. The template module and the template lookup should be able to provide the same service w/o having to duplicate the logic. 

The backup feature also seems to create a directory wherever ansible is executed from, which will create issues with permissions, random locations (i was in /tmp/v234lkj ?), aside from the possible race condition created when multiple hosts are the target of the task.

privateip:
I disagree with using the template module  and template lookup here.  In order to do so it creates to many steps for undue reasons.  If we were to use the template module, then the config would have to pass through the template module, be written back to disk and then read again to be passed to the module.   To many places for things to get messed up there.  There are valid use cases where a playbook could have many *_template tasks in them.

WRT to backups, agreed, we should add more deterministic logic around the backup destination.

Issue #13951, opened 2016-01-18
setup module fails on OpenVZ container with Ansible 2.0.0.1


aibor:
**Issue Type:**
Bug Report

**Ansible Version:**
2.0.0.1
This issue is not present in ansible 1.9.

**Ansible Configuration:**
N/A

**Environment:**
Ansible host: Debian Jessie
Target host: CentOS6 OpenVZ container: `2.6.32-5-openvz-amd64`

**Summary:**
When running `ansible -m setup` for the container, the module fails with `MODULE FAILURE` and a python exception stack trace.

**Expected Results:**
JSON Output of facts of the target host.

**Actual Results:**

```
Traceback (most recent call last):
File "<stdin>", line 5142, in <module>
File "<stdin>", line 140, in main
File "<stdin>", line 84, in run_setup
File "<stdin>", line 5080, in ansible_facts
File "<stdin>", line 4740, in populate
File "<stdin>", line 4746, in get_virtual_facts
File "<stdin>", line 5068, in get_file_lines
File "<stdin>", line 5063, in get_file_content
UnboundLocalError: local variable 'datafile' referenced before assignment
```

**Steps To Reproduce:**
Looking at the code, it seems to result from [facts.py line 2667](https://github.com/ansible/ansible/blob/7de237c5a1742660bc240c15e97a1f5b0d1611a8/lib/ansible/module_utils/facts.py#L2667) as `/proc/1/cgroup`is available, but not readable from within the container, which results in the exception in [line 2984](https://github.com/ansible/ansible/blob/7de237c5a1742660bc240c15e97a1f5b0d1611a8/lib/ansible/module_utils/facts.py#L2984), which I can reproduce in the container with the following snippet:

```
>>> def read_cgroup():
...   try:
...     f = open('/proc/1/cgroup')
...   finally:
...     f.close()
...
>>> read_cgroup()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 5, in read_cgroup
UnboundLocalError: local variable 'f' referenced before assignment
```

Comparing the virtualization environment detection of ansible 1.9 with 2.0, it appears that in the former version the function returned before opening that file when in OpenVZ environment.

If you need any further information regarding the OpenVZ environment, please let me know.

bcoca:
```

**Comment by bcoca** [comment]
can you try running with the following patch?

```
diff --git a/lib/ansible/module_utils/facts.py b/lib/ansible/module_utils/facts.py
[...]

PR #14014, opened 2016-01-20
Update get_virtual_facts for OpenVZ 2.6.32-4-pve

mengbo:
In my old openvz virtual machine, get_virtual_facts failed, `/proc/1/cgroup` exists, but when read it, Operation not permitted. So must check openvz first.

```
43-401-panda:~# uname -a
Linux 43-401-panda 2.6.32-4-pve #1 SMP Thu Oct 21 09:35:29 CEST 2010 x86_64 GNU/Linux
43-401-panda:~# ls -l /proc/1/cgroup 
-r--r--r-- 1 root root 0 Jan 20 16:10 /proc/1/cgroup
43-401-panda:~# more /proc/1/cgroup 
/proc/1/cgroup: Operation not permitted
43-401-panda:~#
```

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -2670,6 +2670,14 @@ class LinuxVirtual(Virtual):
 
     # For more information, check: http://people.redhat.com/~rjones/virt-what/
     def get_virtual_facts(self):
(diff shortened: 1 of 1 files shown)


bcoca:
Already fixed via a773486 as documented in  #13951

PR #14021, opened 2016-01-20
Adding docs help for testing modules

gregdek:
Module testing requires dealing with git submodules. Thanks to @sivel for a quick and dirty HOWTO for pulling a particular PR from module repos for testing. We should definitely flesh this out, but this is a good start.


jimi-c:
Merged, thanks!

PR #14025, opened 2016-01-20
rework run_command's env setting to not change os.environ for the res…

abadger:
…t of the module.

New param to run_command to modify the environment for just this invocation.
Documentation and comment adjustments.


bcoca:
+1

PR #14028, opened 2016-01-20
Adds 'ansible_check_mode' boolean magic variable

### LINKED ISSUE #10696: Knowing whether a playbook is run in check mode
##### Issue Type:

Feature Idea
##### Ansible Version:

To this day, this feature does not exist. I am running `ansible 1.9.0.1`.
##### Environment:

N/A
##### Summary:

It is natively impossible to know from a task, a role, or a playbook, if we are running in [check mode](http://docs.ansible.com/playbooks_checkmode.html) or not.
##### Steps To Reproduce:

It can be very useful though. For example, some module (`ufw`, `git`, ...) require that the underlying executable must be installed. Lots of roles do that as a first task. If the task is run in check mode for the first time, the role will fail. Same applies if you try to do changes to a user that was not created, to create a symlink on a file that was not copied over yet, ...

In some cases, just knowing that we are in check mode is enough (although I agree that some cases require more work, like checking the presence of an executable which `which`, the presence of a user with `id`, ... but in the end, knowing that we are in check mode is still very valuable even in those cases).
##### Expected Results:

A variable? An extra fact? A extra directive at the same level than `always_run`, `when`, ... ? Not sure what is the best way to present this but I'd be happy to have others' opinions.
##### Actual Results:

Since it's not part of Ansible yet, it's still doable but through very [hacky solutions](http://stackoverflow.com/questions/28729567/skip-ansible-task-when-running-in-check-mode). This same page also shows some more use cases for this.

kustodian:
Adds a boolean magic variable `ansible_check_mode` which will be set to `True` if run with `--check`, otherwise it will be `False`. This allows for an easy way to skip, or ignore tasks in check mode if it doesn't make sense to run them, e.g. need to run git module, but git is not installed yet (since we are running check mode).

``` yaml
tasks:
  - name: install git
    yum: name=git

  - name: this task will be skipped in check mode
    git: repo=ssh://git@github.com/mylogin/hello.git dest=/home/mylogin/hello
    when: not ansible_check_mode

  - name: this task will ignore errors in check mode
    git: repo=ssh://git@github.com/mylogin/hello.git dest=/home/mylogin/hello
    ignore_errors: "{{ ansible_check_mode }}"
```

This also closes https://github.com/ansible/ansible/issues/10696

### DIFF
--- a/docsite/rst/playbooks_checkmode.rst
+++ b/docsite/rst/playbooks_checkmode.rst
@@ -38,7 +38,25 @@ Example::
 
 As a reminder, a task with a `when` clause evaluated to false, will
 still be skipped even if it has a `always_run` clause evaluated to
-true.
+true. 
+
+Also if you want to skip, or ignore errors on some tasks in check mode
+you can use a boolean magic variable `ansible_check_mode` (added in version 2.1)
+which will be set to `True` during check mode.
+
+Example::
+
+    tasks:
+
+      - name: this task will be skipped in check mode
+        git: repo=ssh://git@github.com/mylogin/hello.git dest=/home/mylogin/hello
+        when: not ansible_check_mode
+
+      - name: this task will ignore errors in check mode
+        git: repo=ssh://git@github.com/mylogin/hello.git dest=/home/mylogin/hello
+        ignore_errors: "{{ ansible_check_mode }}"
+
+
 
 .. _diff_mode:
 
--- a/lib/ansible/vars/__init__.py
+++ b/lib/ansible/vars/__init__.py
@@ -99,6 +99,7 @@ class VariableManager:
         self._inventory = None
         self._hostvars = None
         self._omit_token = '__omit_place_holder__%s' % sha1(os.urandom(64)).hexdigest()
+        self._options_vars = defaultdict(dict)
 
     def __getstate__(self):
         data = dict(
@@ -109,6 +110,7 @@ class VariableManager:
             host_vars_files = self._host_vars_files,
             group_vars_files = self._group_vars_files,
             omit_token = self._omit_token,
+            options_vars = self._options_vars,
             #inventory = self._inventory,
         )
         return data
@@ -122,6 +124,7 @@ class VariableManager:
         self._group_vars_files = data.get('group_vars_files', defaultdict(dict))
         self._omit_token = data.get('omit_token', '__omit_place_holder__%s' % sha1(os.urandom(64)).hexdigest())
         self._inventory = data.get('inventory', None)
+        self._options_vars = data.get('options_vars', dict())
 
     def _get_cache_entry(self, play=None, host=None, task=None):
         play_id = "NONE"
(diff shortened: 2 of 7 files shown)


docsite/rst/playbooks_checkmode.rst
```diff
@@ -38,7 +38,25 @@ Example::
...
+you can use a boolean magic variable `ansible_check_mode` (added in version 2.1)
+which will be set to `True` during check mode.
+
+Example::
+
+    tasks:
+
+      - name: this task will be skipped in check mode
+        git: repo=ssh://git@github.com/mylogin/hello.git dest=/home/mylogin/hello
+        when: "{{ not ansible_check_mode }}"
```
bcoca: conditionals NEVER need {{ }}

kustodian: I wasn't sure anymore, since from 2.0 you started pushing curly brackets intensively. I'll fix it, not a problem.

bcoca:
yes {{}} everywhere except conditionals, which always need to be fully tempalted. 

rule: if not 'when' then {{ }}

kustodian:
Haha, if explains everything :)

docsite/rst/playbooks_checkmode.rst
```diff
@@ -38,7 +38,25 @@ Example::
 
 As a reminder, a task with a `when` clause evaluated to false, will
 still be skipped even if it has a `always_run` clause evaluated to
-true.
+true. 
+
```
bcoca, line 42: ~~version added or 'starting in 2.1' is needed so people don't try using it on older versions~~

lib/ansible/vars/__init__.py
```diff
@@ -150,6 +153,9 @@ def extra_vars(self, value):
     def set_inventory(self, inventory):
         self._inventory = inventory
 
+    def set_check(self, check):
```
bcoca: I think we should have a more 'generic' option and allow several vars to be set from options, not just this one

kustodian: I can think of two ways to do this and since I'm very new to the Ansible code, what do you think would be better:
1. Create a new property like `extra_vars`, named e.g. `options_magic_vars` and then create a method in `ansible.utils.vars` which will be similar to `load_extra_vars(loader, options)`, e.g. `load_options_magic_vars(options)` (I guess I wouldn't need a loader here) which will return a dictionary of options for which we want to create magic vars. So in future we would only need to change that one method to add a new magic variable from options?
2. Create a `set_options(self, options)` which will pass all the options to VairableManager in which we can define which options need to become magic variables, so again we will define it in only one place?

Which one do you think fits better, or do you have a better suggestion?

kustodian:
@bcoca @jimi-c as requested I made the whole thing more general. To create variables from other options you would just need to edit `load_options_vars(options)` in `lib/ansible/utils/vars.py`.

If you like the changes, I will resolve conflicts when I run a rebase. Should I just force push when I do a rebase into this branch, so that you can merge it like that? Is that what it is done usually before merge?

nitzmahone:
@kustodian - yep, updating a PR that way is one of the few "sane" uses for force push, and is the way I do it all the time. ;)

bcoca:
@kustodian if you can rebase, we would like to merge this soon

kustodian:
Squashed and rebased, so it should be ready for merging. Thanks for being fast and responsive.

P.S. `CHANGELOG.md` should probably be updated.

nitzmahone:
Tested a bit under several configs- LGTM! Merging...

Issue #9723, opened 2014-12-05
make inventory_dir be available to vars_files resolution/templating


willthames:
##### Issue Type:

Bug Report
##### Ansible Version:

```
ansible 1.9 (devel 3b80f63e22) last updated 2014/12/05 14:09:09 (GMT +1000)
  lib/ansible/modules/core: (ec2_snapshot_remove b287e7c0a6) last updated 2014/12/02 11:18:53 (GMT +1000)
  lib/ansible/modules/extras: (detached HEAD 7e6fc7023d) last updated 2014/11/05 11:09:14 (GMT +1000)
  v2/ansible/modules/core: (detached HEAD cb69744bce) last updated 2014/11/05 11:09:23 (GMT +1000)
  v2/ansible/modules/extras: (detached HEAD 8a4f07eecd) last updated 2014/11/05 11:09:31 (GMT +1000)
  configured module search path = None
```
##### Environment:

Fedora 21
##### Summary:

You cannot use `inventory_dir` with `vars_files` - it's just not available to the `_update_vars_files` method. 
##### Steps To Reproduce:

```

---
- hosts: localhost

  vars_files:
    - "{{included}}"

  tasks:
    - debug: msg="{{included}}"
    - debug: msg="{{inventory_dir}}"
    - debug: msg="{{hello}}"
```

run with `ansible-playbook -i ~/hosts` and `~/files/hello` containing:

```
hello: world
```

and `~/group_vars/all.yml` containing `included: "{{inventory_dir}}/files/hello`
shows that `inventory_dir` is correctly set but `hello` which should be set by `files/hello` is never set. 
##### Expected Results:

`vars_files` should be able to expand `inventory_dir`
##### Actual Results:

```
PLAY [localhost] ************************************************************** 

GATHERING FACTS *************************************************************** 
ok: [localhost]

TASK: [debug msg="{{included}}"] ********************************************** 
ok: [localhost] => {
    "msg": "/home/will/files/hello"
}

TASK: [debug msg="/home/will"] ********************************************* 
ok: [localhost] => {
    "msg": "/home/will"
}

TASK: [debug msg="{{hello}}"] ************************************************* 
fatal: [localhost] => One or more undefined variables: 'hello' is undefined

FATAL: all hosts have already failed -- aborting

PLAY RECAP ******************************************************************** 
           to retry, use: --limit @/home/will/vars_files.retry

```

hkariti:
[...]

PR #14029, opened 2016-01-20
Allow special inventory vars to be used in vars_files

moises-silva:
I want to be able to load variables depending on the inventory file used. I know we could put the variables in the inventory file itself but feels cleaner to keep them in a separate file, so I want to be able to use `{{ inventory_file }}` in the value of vars_files. This didn't work, and I found someone else had already filed issue #9723.

I also found out this issue is already fixed in 2.0, but moving to 2.0 is not yet an option for some environments.

This commit adds _load_vars in the Play object so it can be
used when updating the vars files templated values.

The precedence of the other injected variables was kept
and _load_vars will take the lowest precedence.


bcoca:
sorry, we are going to pass on this PR, we are not adding new features to the 1.9 tree.

PR #14034, opened 2016-01-20
Use SSH askpass instead of sshpass for password logins

saltsa:
## Description

This PR makes OpenSSH to use askpass directly without the need of external sshpass program.

This sets the password to the environment variable called ANSIBLE_SSH_PASS and sets the SSH_ASKPASS enviroment to "/usr/bin/ansible-pwecho". Itse also sets the DISPLAY enviroment variable to empty string, if it's not set. When the ssh needs password, it'll execute the askpass program which simply outputs the password from the enviroment variable.
## Security considerations

The password can be read from the environment variable of the ansible-pwecho program. However, at least in Linux the permissions are limited to the current user.

On the other hand, the pwecho program is very simple.
## TODO
- Does the change to the enviroment variables affect Ansible or other programs?
- I tested this only with simple inventory file containing hostname, ansible_ssh_user and ansible_ssh_pass
- Should the feature be activated by default or with some command line switch?
- Should the sshpass-program be removed completely or left as a fallback
  - Current patch removes support for that program
- Does the fix work with sftp or scp
- Test with other systems (currently tested only in Linux)

### DIFF
--- a/bin/ansible-pwecho
+++ b/bin/ansible-pwecho
@@ -1,2 +1,1 @@
-#!/bin/sh
-echo $ANSIBLE_SSH_PASS
+ansible
--- a/bin/ansible-pwecho
+++ b/bin/ansible-pwecho
@@ -1,0 +1,2 @@
+#!/bin/sh
+echo $ANSIBLE_SSH_PASS
--- a/lib/ansible/cli/pwecho.py
+++ b/lib/ansible/cli/pwecho.py
@@ -27,12 +27,15 @@ from ansible.cli import CLI
 
 class PwechoCLI(CLI):
 
-    def __init__(self, args):
-        super(PwechoCLI, self).__init__(args)
-
     def run(self):
-        sys.stdout.write(os.getenv("ANSIBLE_SSH_PASS", ""))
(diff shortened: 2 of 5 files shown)


bcoca:
Wouldn't making it a config option in lib/ansible/constants.py work the same and be a much smaller change? Also make it conditional instead of substituting the existing working sshpass for those that do not want the envvar.

saltsa:
Sounds good! It's easy to support the sshpass too. But which one should be default? The old one or this, which has no external dependencies (only the openssh). Or should it first try with askpass and then fallback to sshpass.

bcoca:
default to current and have a toggle, if this is a good substitute we can eventually revers the default

saltsa:
I added the option, which is not enabled by default, to constants.py. It seems, that this doesn't work when using sftp but only with scp. However, if there's controlsocket, it should work.

cesarjorgemartinez:
For ask become pass exists a config option?

sivel:
Thank you very much for your submission.  We have discussed this and decided against accepting this feature.

Our recommendation is to use the paramiko connection plugin, or switch to using key based authentication.

PR #14038, opened 2016-01-21
ansible-galaxy: use HTTP for Travis Notifications

chouseknecht:
The TLS ciphers on our Azure load balancer are not compatible with Travis-ci.  This is a known issue/difficulty with travis notifications. See: https://docs.travis-ci.com/user/notifications/#Note-on-SSL-TLS-Ciphers.

Changing the default .travis.yml created by ansible-galaxy init. It should default to HTTP rather than HTTPS.

### DIFF
--- a/lib/ansible/galaxy/data/travis.j2
+++ b/lib/ansible/galaxy/data/travis.j2
@@ -26,4 +26,4 @@ script:
   - ansible-playbook tests/test.yml -i tests/inventory --syntax-check
 
 notifications:
-  webhooks: https://galaxy.ansible.com/api/v1/notifications/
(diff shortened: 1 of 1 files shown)


abadger:
We'll have to discuss the ramifications of this.

note that I couldn't find the note on the travis docs page saying that using http was the recommended change to make.  The note I see says to open a github issue to have ciphers that we might need added.

chouseknecht:
Withdrawn. Switched to using Nginx as our load balancer, which allows us to specify the exact TLS ciphers to use. So of course, I specified the same ciphers used by Travis, and it works.

PR #14049, opened 2016-01-21
Add lookup for pass - the standard unix password manager

ksmets:
Provide alternative for `password` lookup that uses [pass - the standard unix password manager](http://www.passwordstore.org/) for storing and retrieving password.

Usage:

```

---
- hosts: all
  tasks:
  - name: test generating password with default settings (length=16 no-symbols=True)
    debug: msg="{{ lookup('pass', 'test/foobar') }}"

  - name: test generating password with length 32
    debug: msg="{{ lookup('pass', 'test/foo length=32') }}"

  - name: test generating password with length 24 and allow symbols
    debug: msg="{{ lookup('pass', 'test/bar length=24 no-symbols=False') }}"
```

Code is basically a mashup of the `password` and `pipe` lookups.

### DIFF
--- a/lib/ansible/plugins/lookup/pass.py
+++ b/lib/ansible/plugins/lookup/pass.py
@@ -1,0 +1,95 @@
+# (c) 2016, Koen Smets <koen.smets@gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


bcoca:
i should add one for keepass ...

sivel:
I'm concerned about a confusion between `password` and `pass`.  I don't know what the answer is here, just wanted to "voice" that concern.

djmattyg007:
What's stopping this from being merged?

JanKoppe:
Can you explain the benefit of this lookup plugin over the existing `passwordstore`? I don't see it.

djmattyg007:
It looks like this PR was submitted over a year before #21314, yet this one was basically ignored.

bcoca:
@djmattyg007 actually that was the rebasing of an even older PR

bcoca:
closing this PR as duplicate, I suggest that any features that this one has but the existing plugin lacks be submitted as patches against the existing plugin.

PR #14051, opened 2016-01-21
Fix regression in max byte limit for --diff source files

gechr:
```
› ansible-playbook --version
ansible-playbook 2.0.0.2
  config file = /Users/george/.ansible.cfg
```

Came across an issue where running `ansible-playbook` with the `--check --diff` parameters would suggest that the dest file would be truncated, even though it wouldn't. Ran the same test with `1.9.4` and the diff produced was correct (i.e. no truncation).

I traced the error back to this line and removed the limit and all's well.

Note a similar fix was made for https://github.com/ansible/ansible/pull/8442

### DIFF
--- a/lib/ansible/plugins/action/__init__.py
+++ b/lib/ansible/plugins/action/__init__.py
@@ -613,7 +613,7 @@ class ActionBase(with_metaclass(ABCMeta, object)):
                     diff['src_larger'] = C.MAX_FILE_SIZE_FOR_DIFF
                 else:
                     diff['after_header'] = source
-                    diff['after'] = src_contents
+                    diff['after'] = src_contents + src.read()
             else:
                 display.debug("source of file passed in")
                 diff['after_header'] = 'dynamically generated'


hloeffler:
+1

bcoca:
I would make this adjustable (via config), with the default set to 8k, you really don't wan't people diffing very large files over the network.

gechr:
After taking a closer look, it would seem like this is actually a [regression in logic since v1.9.4](https://github.com/ansible/ansible/blob/v1.9.4-1/lib/ansible/runner/action_plugins/copy.py#L365-L367).

I've updated the title & pushed another commit ~~with the correct fix.~~ (superseded by https://github.com/ansible/ansible/commit/df25f765dd22eecae47518855081a318ef4a74c2)

mgedmin:
> I would make this adjustable (via config), with the default set to 8k, you really don't wan't people diffing very large files over the network.

@bcoca: this 8k limit is for reading local files on the controller machine.  This PR (or my equivalent PR #14406 which I slightly prefer over this one) doesn't affect the amount of data sent over the network.

EDIT: I hadn't realized that gechr pushed an updated fix after your comment.  Obviously I've no idea what the diff originally did.

gechr:
@mgedmin I prefer your suggestion, too. Have pushed another commit to address it.

gechr:
Oops, accidentally hit close, re-opening!

gechr:
Closing this in favour of https://github.com/ansible/ansible/pull/14416

PR #14054, opened 2016-01-21
Add Python 2.4 requirement to module checklist

gregdek:
It's in travis, but not explicitly in the docs.


sivel:
We _might_ want to tweak this a bit, so that it indicates that modules that have dependencies on a library that requires a newer python version don't need to support python2.4.  However, they should clearly indicate that dependency and the required minimum python version under the `requirements` in `DOCUMENTATION`.

sivel:
Also to be explicit, modules without such a dependency should be written to support python 2.4

gregdek:
Tweaked. :)

sivel:
Merged, Thanks!

PR #14079, opened 2016-01-22
Add vault support to copy action plugin via dataloader implementation

cambell-prince:
Added two methods to dataloader, along the lines discussed with @bcoca in #13849.
- get_real_file will decrypt vault encrypted files and return either a path to a temporary file or simply the path to the unencrypted file if its not encrypted.
- cleanup_real_file will remove a temporary file created with a previous call to get_real_file


bcoca:
closing via merged #15417

PR #14087, opened 2016-01-23
Note "ssh" deprecation for inventory variables

grahamu:
As per the [Ansible 2.0 FAQ](http://docs.ansible.com/ansible/faq.html), "Ansible 2.0 has deprecated the “ssh” from ansible_ssh_user, ansible_ssh_host, and ansible_ssh_port to become ansible_user, ansible_host, and ansible_port."

This change breaks working ansible configurations. Since Q&A sites often reference [`ansible_ssh_user`](http://serverfault.com/search?q=ansible_ssh_user) (and the others), a comment about this change should be present in the most common place people look when considering an upgrade to ansible 2.x.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ Ansible Changes By Release
   This re-executes inventory scripts, but does not force them to ignore any cache they might use.
 * New delegate_facts directive, a boolean that allows you to apply facts to the delegated host (true/yes) instead of the inventory_hostname (no/false) which is the default and previous behaviour.
 * local connections now work with 'su' as a privilege escalation method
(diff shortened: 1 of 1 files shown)


jeffwidman:
Just discovered that `ansible_ssh_user` is no longer available. 

`ansible_user` is not a complete replacement--it returns `null` when I try to use it to access the user who SSH'd into the box. I posted the issue on StackOverflow: http://stackoverflow.com/questions/34990278/in-ansible-v2-which-variable-stores-the-ssh-username

Probably the deprecation message or FAQ should also list whatever variable is the solution for this, as I can't be the only one using `ansible_ssh_user` this way.

bcoca:
@jeffwidman ansible_ssh_user is still available, but the behavior has changed, if no users is specified in now returns None, as noted in the changelog.

Issue #13848, opened 2016-01-13
--list-hosts does not work with --limit (ansible 2.0)


serialdoom:
as described [here](https://groups.google.com/forum/#!topic/ansible-project/sxuZfahMRUI), this doesn't work

```
$ ansible all -i hosts --limit ~host1 --list-hosts
hosts (10):
    host1
    host2
    host3
    host4
    host5
    host6
    host7
    host8
    host9
    host10
```

nor this

```
ansible all -i hosts --limit host4 --list-hosts
  hosts (10):
    host1
    host2
    host3
    host4
    host5
    host6
    host7
    host8
    host9
    host10
```

but this will

```
ansible-playbook test.yml -i hosts --limit ~host1

PLAY ***************************************************************************

TASK [setup] *******************************************************************
fatal: [host1]: UNREACHABLE! => {"changed": false, "msg": "ERROR! SSH encountered an unknown error during the connection. We recommend you re-run the command using -vvvv, which will enable SSH debugging output to help diagnose the issue", "unreachable": true}
fatal: [host10]: UNREACHABLE! => {"changed": false, "msg": "ERROR! SSH encountered an unknown error during the connection. We recommend you re-run the command using -vvvv, which will enable SSH debugging output to help diagnose the issue", "unreachable": true}

PLAY RECAP *********************************************************************
host1                      : ok=0    changed=0    unreachable=1    failed=0   
host10                     : ok=0    changed=0    unreachable=1    failed=0   
```

ansible version:

```
ansible 2.0.0.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = /usr/share/ansible
```

installation through `pip install --user ansible --upgrade`.
My hosts file was generate with `for f in $(seq 1 10); do echo "host$f" >> hosts; done`

PR #14090, opened 2016-01-24
Make --list-hosts honor --limit (Fix for #13848)

edwkar:
Fix for #13848. 

Makes `--list-hosts` honor `--limit` by recomputing `hosts` when `options.subset` is set.

With the fix, the output for @serialdoom's example cases is as follows:

``` bash
ansible all -i hosts --list-hosts
  hosts (10):
    host1
    host2
    host3
    host4
    host5
    host6
    host7
    host8
    host9
    host10

ansible all -i hosts --limit ~host1 --list-hosts
  hosts (2):
    host1
    host10

ansible all -i hosts --limit host4 --list-hosts
  hosts (1):
    host4
```
### Alternative solution

Recomputing `hosts` like this might be a bit smelly -- one might contend that `hosts` should hold _all_ hosts, regardless of whether a subset is selected. Another, possibly cleaner, route is to use two variables `all_hosts` and `selected_hosts`:

``` python
selected_hosts = all_hosts = inventory.list_hosts(pattern)
no_hosts = False
if len(all_hosts) == 0:
    display.warning("provided hosts list is empty, only localhost is available")
    no_hosts = True

if self.options.subset:
    inventory.subset(self.options.subset)
    selected_hosts = inventory.list_hosts(pattern)
    if len(inventory.list_hosts(pattern)) == 0 and not no_hosts:
        # Invalid limit
        raise AnsibleError("Specified --limit does not match any hosts")

if self.options.listhosts:
    display.display('  hosts (%d):' % len(selected_hosts))
    for host in selected_hosts:
        display.display('    %s' % host)
    return 0
```

If you prefer this, or some other variant, let me know, and I'll fix the PR.

### DIFF
--- a/lib/ansible/cli/adhoc.py
+++ b/lib/ansible/cli/adhoc.py
@@ -132,6 +132,7 @@ class AdHocCLI(CLI):
 
         if self.options.subset:
             inventory.subset(self.options.subset)
+            hosts = inventory.list_hosts(pattern)
             if len(inventory.list_hosts(pattern)) == 0 and not no_hosts:
                 # Invalid limit
                 raise AnsibleError("Specified --limit does not match any hosts")


bcoca:
actually i might have simpler solution that also does not recompute, some of this code was copy/pasted from ansible-playbook as adhoc will need the implicit localhost specified and would not need the current warning:

```
--- a/lib/ansible/cli/adhoc.py
+++ b/lib/ansible/cli/adhoc.py
@@ -124,17 +124,13 @@ class AdHocCLI(CLI):
         inventory = Inventory(loader=loader, variable_manager=variable_manager, host_list=self.options.inventory)
         variable_manager.set_inventory(inventory)

-        hosts = inventory.list_hosts(pattern)
-        no_hosts = False
-        if len(hosts) == 0:
-            display.warning("provided hosts list is empty, only localhost is available")
-            no_hosts = True

         if self.options.subset:
             inventory.subset(self.options.subset)
-            if len(inventory.list_hosts(pattern)) == 0 and not no_hosts:
-                # Invalid limit
-                raise AnsibleError("Specified --limit does not match any hosts")
+
+        hosts = inventory.list_hosts(pattern)
+        if len(hosts) == 0:
+            raise AnsibleError("Specified hosts options do not match any hosts")

```

^ pushing that directly, also x2 checking ansible-playbook

bcoca:
fixed via ac1d1673be1968f21f158b8a1fda6249d0fb9b1e

PR #14096, opened 2016-01-24
Added core filter 'in_container'.

marcpo:
The 'in_container' filter is similar to the 'in' operator in python.
It will return 'true' if:
- item is in a list
- item is a key in a dict (hash array)
- item (string) is in a string
  It will return false otherwise.

### DIFF
--- a/lib/ansible/plugins/filter/core.py
+++ b/lib/ansible/plugins/filter/core.py
@@ -358,6 +358,17 @@ def extract(item, container, morekeys=None):
 
     return value
 
+def in_container(value, container):
+    '''
+    Returns True is the value is in the container. False otherwise.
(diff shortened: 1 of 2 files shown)


marcpo:
Not sure what to do about the documentation. Should I take care of it ?

sivel:
Why can't you just use the `in` keyword with jinja2?  It performs the same action and is already available.

```
when: '"bar" in foo_list'
when: '"bar" in foo_dict'
when: '"bar" in foo_string'
```

All of the above is already possible without this addition.

Due to the info I've provided above, I'm going to go ahead and close this issue out.  Please feel free to swing by the ansible-project mailing list or the #ansible IRC channel.

PR #14105, opened 2016-01-25
Handle the key 'custom' in the 'diff' result from modules

towolf:
So far, when a 'diff' dict is returned with module results, it is
checked for 'before' and 'after' texts, which are processed in
_get_diff() by python difflib.  This generates the changes to display
when CLI users specify --diff.

However, some modules will generate changes that cannot easily be
expressed in a conventional diff. One example is the output of the
synchronize module, which presents changed files in a common log format
as in `rsync --itemize-changes`.

Add a check for a diff['custom'] key, which can contain custom diff text
from modules.

### DIFF
--- a/lib/ansible/plugins/callback/__init__.py
+++ b/lib/ansible/plugins/callback/__init__.py
@@ -131,6 +131,8 @@ class CallbackBase:
                         differ = difflib.unified_diff(to_unicode(diff['before']).splitlines(True), to_unicode(diff['after']).splitlines(True), before_header, after_header, '', '', 10)
                         ret.extend(list(differ))
                         ret.append('\n')
+                    if 'prepared' in diff:
+                        ret.append(to_unicode(diff['prepared']))
                     return u"".join(ret)
             except UnicodeDecodeError:
                 ret.append(">> the files are different, but the diff library cannot compare unicode strings\n\n")


sivel:
I'm not fond of naming this `custom` as it sounds too nebulous.  What about something like `combined`, `unified` or `prepared`?

towolf:
I'm not married to the name `custom`.
- `unified` is bad because that is a fixed concept for things that look like `diff -u`
- `combined` is bad because there isn't necessarily anything combined in the output
- `prepared` I could live with

I suggested `custom` because the existing way is to stuff `diff['before']` and `diff['after']` into the returned JSON to be diffed by the `_get_diff()` method. I wanted this to enable a "custom diff", and hence the name ...

towolf:
This is the output of synchronize, for example:

```
TASK [synchronize] *************************************************************
changed: [localhost]
.d..t...... ./
>f+++++++++ README.md
>f+++++++++ ansible-playbook.lprof
>f+++++++++ broken.sh
>f+++++++++ inventory.json
>f+++++++++ inventory.sh
>f+++++++++ out.stats
>f+++++++++ out.stats2
>f+++++++++ out.stats3
>f+++++++++ playbook.yml
cd+++++++++ group_vars/
```

Format is following: http://andreafrancia.it/2010/03/understanding-the-output-of-rsync-itemize-changes.html

It's a report of differences, but not a `diff` diff.

hloeffler:
Thanks for your PR.

for me custom is short for custom_diff_output
but maybe we can just use the long name

bcoca:
+1 for 'prepared' or even 'results'

chrrrles:
+1 for `prepared`

towolf:
Alright then, I put s/custom/prepared/ now.

PR #10413, opened 2015-03-09
Implement AND'd filters in ec2.py/ini

tgerla:
##### SUMMARY
This extends the EC2 inventory source filtering syntax to include "&" for ANDing filters together, as suggested by @cchurch a few days ago: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/ansible-project/CV4kOYyw6oU/-SHNI0_hSs0J

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
plugins/inventory/ec2.ini
plugins/inventory/ec2.py

##### ANSIBLE VERSION
2.3

PR #14106, opened 2016-01-25
ec2 inventory allow for AND usage when specifying multiple tags for instances

kapilt:
pretty common to want the intersection of multiple tags, ie my app tag, and the 'dev' stage in an account with lots of apps/stages.

### DIFF
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -187,7 +187,6 @@ class Ec2Inventory(object):
 
         print(data_to_print)
 
-
     def is_cache_valid(self):
         ''' Determines if the cache files have expired, or if it is still valid '''
 
@@ -199,7 +198,6 @@ class Ec2Inventory(object):
                     return True
 
         return False
-
 
     def read_settings(self):
         ''' Reads the settings from the ec2.ini file '''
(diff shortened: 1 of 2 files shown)


bcoca:
@kapilt could squash to a single commit?

kapilt:
@bcoca squashed and minor doc cleanup in config example

contrib/inventory/ec2.py
```diff
@@ -397,6 +397,14 @@ def read_settings(self):
                     continue
                 self.ec2_instance_filters[filter_key].append(filter_value)
 
+        # Instance Filters can be AND'd or OR'd based on the filters_operator
+        # parameter, default behavior is OR'd
+        self.ec2_filter_operator = 'or'
+        if config.has_option('ec2', 'filters_operator'):
+            self.ec2_filter_operator = config.get('ec2', 'filters_operator')
+            if self.ec2_filter_operator not in ('or', 'and'):
+                self.ec2_filter_operator = 'or'
```
mgedmin: Would be nice to warn the user their configuration is being ignored because the specified value wasn't recognized.

kapilt: noted, sounds good

aioue:
Related: https://github.com/ansible/ansible/pull/14901

kapilt:
@wtcross here

bcoca:
also linking #10413 as it was another attempt to same using a different approach

nicolas-g:
+1

mkrizek:

PR #14109, opened 2016-01-25
Checking node health on Consul, besides service health

raitech:
We can have a false positive of a node on the group of available nodes for a service, even if the serfHealth check of the node did not pass.

This patch/pull-request take this in to account to put a node in the available or unavailable groups.


jstoja:
I think this should be rebased and I think the documentation of the function should be edited to match the changes, but it seems like a good addition to me.

gregdek:
No response from @raitech in quite some time, so I'd recommend closure -- that said, @jstoja you might consider submitting this as a new PR if you think it would be useful.

gundalow:
Given above, closing

PR #14110, opened 2016-01-25
Change example to use pkg rather than the obsolete pkg_add

infracaninophile:
pkg_add has been obsoleted in all released versions of FreeBSD for
several years.  Change the example to use the appropriate pkg(8)
command line.

### DIFF
--- a/docsite/rst/intro_bsd.rst
+++ b/docsite/rst/intro_bsd.rst
@@ -30,7 +30,7 @@ Bootstrapping BSD
 For Ansible to effectively manage your machine, we need to install Python along with a json library, in this case we are using Python 2.7 which already has json included.
 On your control machine you can simply execute the following for most versions of FreeBSD::
 
-    ansible -m raw -a “pkg_add -r python27” mybsdhost1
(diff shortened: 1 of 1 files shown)


bcoca:
i still know of several FreeBSD 8 installed bases, I would add this as the main way of installing but leave a 'for older versions of ' with the existing example

infracaninophile:
The pkg_add command as written wouldn't work today -- since all FreeBSD 8.x versions are now well out of support, there aren't old style packages available to download from the FreeBSD FTP sites. 

You'ld have to have a custom PKGSITE setup, and AFAIK no one really advised doing that with pkg_add -- people would have been far more likely to install by compiling stuff from the ports tree,  It is only since pkg(8) has been around that using pre-compiled binary packages has achieved any traction within the FreeBSD user base.

Matthew

bcoca:
True, most of those that i know still using 8 are pointing to the archives, so i'll just updated this and add the old one back if anyone complains.

infracaninophile:
Great, thanks.

PR #14129, opened 2016-01-26
Add support for additional parameter 'pkg_opts' in package action.

indrajitr:
Current version of package action is quite limited and doesn't have much capability beyond `package: name=foo state=present` or `package: name=foo state=absent`.

Additional parameter `pkg_opts` now facilitates adding package manager specific options to be passed on to the underlying package manager module via external (platform specific) variables.

So we can now do: `package name=foo pkg_args={{my_pkg_args}}`, where `my_pkg_args`
can be configured to have actual package module specific parameter (specific to apt, pacman, homebrew) etc.

### DIFF
--- a/lib/ansible/plugins/action/package.py
+++ b/lib/ansible/plugins/action/package.py
@@ -39,6 +39,8 @@ class ActionModule(ActionBase):
 
         module = self._task.args.get('use', 'auto')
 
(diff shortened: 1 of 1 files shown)


bcoca:
closing for same reasons as in linked PR

PR #14136, opened 2016-01-26
fix cache reading code, add refresh-cache and instance_filters

kbroughton:
vmware.ini has the line
# cache_dir = ~/.cache/ansible

However, the code that reads the cache_dir does not expand the ~.  Added os.path.expanduser to fix.

To make this dynamic inventory similar to ec2.py I added
instance_filters
and a --refresh-cache flag.

To test, install jq.  Customize your vmware.ini and source your vars into the shell environment.
 note '' is two single quotes in the lines below
    $ ./vmware.py | jq ''
    Note now long it takes
    $ ./vmware.py | jq '' 
    The second time should be much faster if you uncomment cache_dir
    $ ./vmware.py --refresh-cache | jq ''
    This should return almost instantly

### DIFF
--- a/contrib/inventory/vmware.py
+++ b/contrib/inventory/vmware.py
@@ -60,8 +60,9 @@ from suds.sudsobject import Object as SudsObject
 
 class VMwareInventory(object):
 
-    def __init__(self, guests_only=None):
+    def __init__(self, guests_only=None, refresh_cache=False):
         self.config = ConfigParser.SafeConfigParser()
+        self.refresh_cache = refresh_cache
         if os.environ.get('VMWARE_INI', ''):
             config_files = [os.environ['VMWARE_INI']]
         else:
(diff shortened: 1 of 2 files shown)


jimi-c:
cc @chrrrles

contrib/inventory/vmware.py
```diff
@@ -330,6 +339,13 @@ def get_inventory(self, meta_hostvars=True):
                 if prefix_filter:
                     if vm.name.startswith( prefix_filter ):
                         continue
+                if instance_filters:
+                    included = False
+                    for instance_filter in instance_filters:
+                        if vm.name.startswith( instance_filter ):
```
chrrrles, line 345: Since we already have a `prefix_filter`, it may be better to use `instance_filters` as a general search term:

```
for instance_filter in instance_filters:
    if instance_filter in vm.name:
```

mckerrj:
Closed for out of date, and moving to plugins

PR #14153, opened 2016-01-27
Details on how to convert subnet masks into CIDR

oneswig:
Documentation addition: how to manage IP networks given with a subnet mask and how to canonicalise them into CIDR notation.

Use data reported by Ansible network facts as an example.


jimi-c:
Merged, thanks!

PR #14178, opened 2016-01-28
'bytes' object has no attribute 'encode' on ansible-vault view

jgargallo:
Encode cannot be applied to a 'bytes' string so we either assume utf-8 encoding for de original vault file or a config parameter is required for the viewer. We'll assume utf-8 for this fix

### DIFF
--- a/lib/ansible/cli/__init__.py
+++ b/lib/ansible/cli/__init__.py
@@ -459,7 +459,7 @@ class CLI(object):
             os.environ['LESS'] = CLI.LESS_OPTS
         try:
             cmd = subprocess.Popen(cmd, shell=True, stdin=subprocess.PIPE, stdout=sys.stdout)
-            cmd.communicate(input=text.encode(sys.stdout.encoding))
+            cmd.communicate(input=text.decode('utf-8').encode(sys.stdout.encoding))
         except IOError:
             pass
         except KeyboardInterrupt:


abadger:
Thanks for finding this problem!  A few notes:
- for conversion we should use to_unicode and to_bytes from ansible.utils.unicode.py.  Those functions will perform a lot of boilerplate (like checking whether the string is already a bytes or unicode object... which is the problem here, the text was already bytes).
- Although we could just do to_bytes here and get rid of the error, we really want to be cleaning up the code so that internally everything is unicode.  So we want to look around and find out where the "text" variable is entering ansible and transform it into unicode type there.  (If it's bytes it can also create problems for display.display() as that function expects unicode type).
- not sure if you're running into this because you're trying to use ansible with python3... If that's so be warned that we're in the process of porting -- ansible is not yet python3 compatible but we're definitely open to patches to fix problems.  Our goal has been to get the pieces that run on the controller (of which vault cli is one) to run under python3 and then start  working on the modules.

jgargallo:
This issue can be reproduced in python2 as well but with different exception ('ascii' codec can't encode character ...).
Since the best practice would be to use unicode but this text comes in bytes decrypted by pycrypto in [this method](https://github.com/dlitz/pycrypto/blob/master/lib/Crypto/Cipher/blockalgo.py#L765), I'm afraid we have the following options:
- assume utf-8 by default
- guess enconding using any of the character encoding detectors out there
- accept encoding as a parameter

abadger:
<nod>  We can assume utf-8 (at least for now).  vault is a bit unique in the ansible code base in that the border between bytes and text is a little deeper inside of ansible code than most things.  Haven't looked at the code yet to see where it should be.

I found a few other places that are calling the pager methods and am presently looking at the ones in doc.py (in case you want to keep looking into modifying vault).

abadger:
Fix for docs about ready -- https://gist.github.com/abadger/bea1a365252b62ea89ab  Just getting another committer to review it.

Are you working on vault or would you rather I took a look?

Once docs, galaxy, and vault are modified I'll change `cmd.communicate(input=text.encode(sys.stdout.encoding))` to `cmd.communicate(input=to_bytes(text))`.  It's good for helping me verify problems as I fix them right now so I haven't changed it yet.

jgargallo:
All yours, to be honest, a colleague from work asked me for help with a python error and ended up doing a PR here.

Nice to help in such a great project.

abadger:
Changes merged to devel and to stable-2.0.  They should be in the 2.0.1 release.  Feel free to let me know if you test that out and it doesn't solve the problem completely. (I'm abadger1999 on irc.freenode.net if you need to get ahold of me and I don't seem to have seen the bug report).

Issue #14135, opened 2016-01-26
Ansible 2.0.0.2 - Using with_items and ec2_eni is changing other variables


wholroyd:
Assuming this is a set of tasks in a role and that all variables have been validated to be defined to their correct values prior to hitting the ec2_ini task...

``` yaml
- debug: var="region"
- name: Request creation of new elastic network interfaces for OpenSwan
  ec2_eni:
    region: "{{ region }}"
    profile: "{{ cloud }}"
    state: present
    description: "{{ cloud }}-{{ platform }}-{{ component }}-openswan"
    subnet_id: "{{ item.id }}"
    private_ip_address: "{{ item.cidr | ipaddr('net') | ipaddr(4) }}"
  with_items: "{{ vpc_created.subnets }}"
  register: vpc_eni_created
- debug: var="vpc_eni_created"
```

And that this is the output from using -vvv...

```
TASK [shared_infrastructure : debug] *******************************************
task path: /home/wholroyd/source/automation/ansible/roles/aws_ec2_vpc/tasks/main.yml:52
ok: [localhost] => {
    "region": "us-west-1"
}

TASK [shared_infrastructure : Request creation of new elastic network interfaces for OpenSwan] ***
task path: /home/wholroyd/source/automation/ansible/roles/aws_ec2_vpc/tasks/main.yml:53
ESTABLISH LOCAL CONNECTION FOR USER: wholroyd
localhost EXEC /bin/sh -c '( umask 22 && mkdir -p "$( echo $HOME/.ansible/tmp/ansible-tmp-1453825989.75-66259187748651 )" && echo "$( echo $HOME/.ansible/tmp/ansible-tmp-1453825989.75-66259187748651 )" )'
localhost PUT /tmp/tmpL_mALl TO /home/wholroyd/.ansible/tmp/ansible-tmp-1453825989.75-66259187748651/ec2_eni
localhost EXEC /bin/sh -c 'LANG=C LC_ALL=C LC_MESSAGES=C /usr/bin/python /home/wholroyd/.ansible/tmp/ansible-tmp-1453825989.75-66259187748651/ec2_eni; rm -rf "/home/wholroyd/.ansible/tmp/ansible-tmp-1453825989.75-66259187748651/" > /dev/null 2>&1'
[...]

PR #14185, opened 2016-01-28
Finalize task parameter values from include actions

wholroyd:
This is a proposed fix to the issue described in https://github.com/ansible/ansible/issues/14135.

In summary, task parameters from include actions aren't finalized and will still reference `"{{ item }}"` even if you've assigned the value of it to another variable. This causes issues where you want to use the value from a parent with_\* loop inside a child with_\* loop. This does not change the behavior of `"{{ item }}"` being overwritten in child loops, only the behavior of other variables that happen to reference `"{{ item }}"` and shouldn't.

### DIFF
--- a/lib/ansible/playbook/included_file.py
+++ b/lib/ansible/playbook/included_file.py
@@ -77,12 +77,19 @@ class IncludedFile:
                     original_task = iterator.get_original_task(original_host, res._task)
 
                     task_vars = variable_manager.get_vars(loader=loader, play=iterator._play, host=original_host, task=original_task)
-                    templar = Templar(loader=loader, variables=task_vars)
 
                     include_variables = include_result.get('include_variables', dict())
                     if 'item' in include_result:
                         task_vars['item'] = include_variables['item'] = include_result['item']
 
+                        # We need to ensure that any variables coming down the included file chain are finalized before
+                        # calling into the child tasks for instance where we are nesting with_* loops between include
+                        # file levels
+                        templar = Templar(loader=loader, variables=include_variables)
+                        for k in include_variables.keys():
+                            include_variables[k] = templar.template(include_variables[k])
+
+                    templar = Templar(loader=loader, variables=task_vars)
                     if original_task:
                         if original_task._task_include:
                             # handle relative includes by walking up the list of parent include


jimi-c:
I vaguely remember a reason we specifically do not do this, though it escapes me now. During the original development of the include functionality we had originally been passing back the variables as templated during the post_validate() step of the task, which caused a problem of some kind, so I have concerns that we're going to run into that again by doing this.

wholroyd:
I started to write some unit tests for it and think I found the issue with the implementation as to why you guys didn't do this. The outer loop doesn't advance, so you repeat the same value for N items in the loop.

I wonder if backing up `job_vars`, templating the vars, executing the loop, and then restoring the vars (updating other values from the iteration of course) would help, because on the next iteration on the parent loop, any vars based on `"{{ issue }}"` should then get evaluated correctly. Somewhere in `task_executor.py:212-251 -> _run_loop(self, items)`.

I'll play with it and I'm open to suggestions.

wholroyd:
Here is the reworked code to remove the side effect you were probably noting above, and the one that I experienced. This is more of an issue when using with_\* loops than a task include, so nothing here targets a task include anymore. I've been able to run it on my playbooks without any issues and it was able to finally complete the tasks successfully.

wholroyd:
The only side effect of the current set of changes is that the value of `{{ item }}` will always represent what the value is according to the outermost with_\* loop. Once I figure out how to correct that to show the current level's value, then we should be good.

bcoca:
This might create unexpected errors for variables that do not have the data defined prior to the include:

```
- include: vlad.yml
  vars:
    later: "{{from_included_task|last}}"
```

wholroyd:
@bcoca In this particular PR, what I've changed should only finalize task parameters on task `include` actions when used with `with_*` because loops are only temporarily evaluated for their parameters and then finalized/executed later on. With the current value of `item` being held at the `play` level, it's not known until execution - but at that point it's too late to finalize other variables making references back to `item`, especially in instances where each of those variables were meant to reference entirely different `item` values.

While I see the value of having the kind of flexibility you've pointed out, that seems like misplaced logic in this scenario. Especially when the overall effort of Ansible is to supposedly to create reusable and shareable modular code via roles, playbooks, and tasks. Using a filter of a variable from within the module (in this case, a task file) to be used later on within the same module violates modularization.

Your example should still work as expected, unless you add a `with_*` to the action.

sivel:
Thank you very much for your submission.  We have discussed this and decided against accepting this feature.

We believe that this would adversely affect the ability to rely on much of the dynamic capabilities that includes provide.

PR #14191, opened 2016-01-28
Added hyphen (-) as a valid character for permissions

paulcalabro:
Here's an example of its usage:

```
fake@fake  /tmp  chmod 777 bar

fake@fake  /tmp  ls -ld bar
drwxrwxrwx  2 fake  fake  68 Jan 28 13:09 bar

fake@fake  /tmp  chmod go=- bar

fake@fake  /tmp  ls -ld bar
drwx------  2 fake  fake  68 Jan 28 13:09 bar
```


jimi-c:
Merged, thanks!

PR #14197, opened 2016-01-29
load now does not modify the incomming data

bcoca:
also removed json loader as yaml loader can do both


sivel:
Just for some historical context, the original reason to using both the `json` and `yaml` parsers was due to the fact that exceptions and errors emitted from the `yaml` module when parsing JSON data could be misleading or less precise that from the `json` module.

Not saying we shouldn't do this, but we may just want to be careful.

At this point I think we have improved many parser errors and such, so this may be moot.  Although figured it worth mentioning.

bcoca:
the JSON removal was secondary, changing to not modify the incoming data was the bigger issue here.

PR #14202, opened 2016-01-29
fixed code for v1 callback runtime compatiblity

bcoca:
old exception code obsoleted by exception avoidance
errors on v1 callback produce warnings and don't stop play
also cleaned out unused imports as per pyflakes

### DIFF
--- a/lib/ansible/executor/task_queue_manager.py
+++ b/lib/ansible/executor/task_queue_manager.py
@@ -19,7 +19,6 @@
 from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
-from multiprocessing.managers import SyncManager, DictProxy
 import multiprocessing
 import os
 import tempfile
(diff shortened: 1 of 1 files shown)


lib/ansible/executor/task_queue_manager.py
```diff
@@ -290,29 +288,27 @@ def send_callback(self, method_name, *args, **kwargs):
...
-                    except Exception as e:
-                        import traceback
-                        orig_tb = to_unicode(traceback.format_exc())
-                        try:
-                            v1_method = method.replace('v2_','')
+                        else:
+                            method()
+                    else:
+                        method(*args, **kwargs)
+                else:
```
abadger: The else clause will occur twice so the callbacks will get called twice.  This is because we're inside of a loop that's iterating through the callback methods for method_name and v2_on_any here.

Perhaps we should check whether either of the methods is None at the top and look for the v1 method there?

PR #14206, opened 2016-01-29
Remove the range part while parsing hostname and port

asouchang:
Fix the issue that the INI parser treat the hosts with range in `/etc/ansible/hosts` as a IPv6 address with port, which causes exception at runtime
For example: `server-[01:99:2].dc1`

### DIFF
--- a/lib/ansible/inventory/ini.py
+++ b/lib/ansible/inventory/ini.py
@@ -110,7 +110,7 @@ class InventoryParser(object):
                     continue
                 hostname = tokens[0]
                 port = C.DEFAULT_REMOTE_PORT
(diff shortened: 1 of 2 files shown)


abadger:
We've taken a look at this problem and the proposed solution.  We think that it is fixed in 2.0 and would recommend updating to that version if this is a problem for you.  We've thought about the risk versus benefit of taking this as a backport to 1.9.x and don't feel comfortable pushing it there.  The ini file syntax has many cornercases that could be encountered with this which could break people's current inventories.  So we'd rather not introduce the potential for breakage for 1.9.  2.x is where we hope that issues such as this can be resolved.

PR #14208, opened 2016-01-29
doc: wrong parameter "succeeded" in when statement

raittes:
In **when** statement, the documentation states the use of "_succeeded_", when it should be "_success_".

Using "succeeded" will not work and return this error:

```
Failed to template {% if mobile_status|succeeded %} True {% else %} False {% endif %}: template error while templating string: no filter named 'succeeded'
```

version: ansible==1.9.4


bcoca:
actually, that is correct on 2.0 and above (I thought I had made note of it). Closing this as the PR is incorrect, but will look into making it clearer in the docs.

Issue #14040, opened 2016-01-21
"rescue" not being executed if failed inside an "include" in development branch


dcodix:
Hi there,

I am not sure if it changed as a feature or if it is a bug .

The thing is:

main.yml

```
- name: deploy
  hosts: all
  gather_facts: True
  tasks:
  - block:
      - include: file_to_fail.yml
      #- file: path=/tmp/file1 state=directory
      - file: path=/tmp/file3 state=directory
    rescue:
      - file: path=/tmp/file2 state=directory
    #any_errors_fatal: yes
```

file_to_fail.yml

```
- file: path=/tmp/file1 state=directory
```

I am testing making the "/tmp/file1" file creation fail.

So, the thing is that in current version (2.0.0) if it fails in the include the rescue part is called and the "/tmp/file2" is created. But, with "devel" branch it does not. It works if instead of include I specify the task there (see commented line), but not with the include.

jimi-c:
# Closing This Ticket

Hi!

We believe the above commit should resolve this problem for you. This will also be included in the next release.

If you continue seeing any problems related to this issue, or if you have any further questions, please let us know by stopping by one of the two mailing lists, as appropriate:
- https://groups.google.com/forum/#!forum/ansible-project - for user questions, tips, and tricks
- https://groups.google.com/forum/#!forum/ansible-devel - for strategy, future planning, and questions about writing code

Because this project is very active, we're unlikely to see comments made on closed tickets, but the mailing list is a great way to ask questions, or post if you don't think this particular issue is resolved.

Thank you!

PR #14213, opened 2016-01-29
If a hosts is marked as failed, it seems odd to mark it as complete too.

dcodix:
Referencing the issue:  https://github.com/ansible/ansible/issues/14040

The problem is that if "run_state" is COMPLETE the host is marked as failed, and rescue is not called.
I don't know if there are other implications, but for me it has no sense mark at the same time as completed and failed. If something failed is not really complete, is it?

I only deleted the condition where in case of failed it was marked as completed and now it seems to work.

I tested it without block/rescue (a playn tasks playbook), with block/rescue with and without include and it seems to work.

Again, I am not sure if this condition was there for some special reason, and I only decided to take it off because conceptualy for me something failed is not completed.

### DIFF
--- a/lib/ansible/executor/play_iterator.py
+++ b/lib/ansible/executor/play_iterator.py
@@ -362,8 +362,6 @@ class PlayIterator:
                     state.run_state = self.ITERATING_RESCUE
                 elif state._blocks[state.cur_block].always:
                     state.run_state = self.ITERATING_ALWAYS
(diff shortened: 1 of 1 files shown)


dcodix:
This PR is not needed anymore as the issue seems to be solved in commit: 699a854bf30784cd7514f4c1b6f8bd742868f7eb

PR #14214, opened 2016-01-29
Better "conflicting action statements" message

dvarrazzo:
I had a tasks file including a snippet twice to create and set up two users.

``` yaml
# Set up the sync user
- include: user.yml
  mode: sync
  user: "{{ssh_user_sync}}"

# Set up the async user
- include: user.yml
  mode: async
  user: "{{ssh_user_async}}"
```

In ansible 2.0 it started failing with the error "conflicting action statements". It is not clear that the problem is that there cannot be an `user` variable, because `user` also happens to be a role. This worked ok in Ansible 1.9.

I think this is brittle (an include may fail in the future because a new action will be implemented...). Anyway I'm not up for a painful discussion about Ansible choices. This patch just provides a better error message:

```
ERROR! conflicting action statements (include, user)
```

which hopefully may lead an user to understand they have to rename a variable.

### DIFF
--- a/lib/ansible/parsing/mod_args.py
+++ b/lib/ansible/parsing/mod_args.py
@@ -287,7 +287,7 @@ class ModuleArgsParser:
             if item in module_loader or item == 'meta' or item == 'include':
                 # finding more than one module name is a problem
                 if action is not None:
-                    raise AnsibleParserError("conflicting action statements", obj=self._task_ds)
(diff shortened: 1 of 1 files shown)


dvarrazzo:
For people ending here googling, it is worth noting that a more robust way to use `include` is using `vars` as in:

``` yaml
- include: user.yml
  vars:
    mode: sync
    username: "{{ssh_user_sync}}"
```

this issue is only about the error message that can be improved.

jimi-c:
Merged, thanks!

PR #14218, opened 2016-01-29
Check plugin classes using isinstance()

austinhyde:
I encountered this issue while making a subclass of the SSH connection
plugin, where the plugin loader attempts to validate that a plugin has
a specific class as one of its bases. The issue is that the current
implementation only accounts for the immediate parents of the plugin.

This fix imports the actual base class at runtime and does a check
accouting for multiple levels of inheritance using isinstance().

### DIFF
--- a/lib/ansible/plugins/__init__.py
+++ b/lib/ansible/plugins/__init__.py
@@ -313,6 +313,10 @@ class PluginLoader:
                 module = imp.load_source(name, path, module_file)
         return module
 
+    def _get_base_class(self):
+        module_name, class_name = self.base_class.rsplit('.', 1)
+        return getattr(__import__(module_name, globals(), locals(), [class_name], 0), class_name)
+
     def get(self, name, *args, **kwargs):
         ''' instantiates a plugin of the given name using arguments '''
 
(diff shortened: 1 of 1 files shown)


austinhyde:
Oh, whoops, this isn't compatible with Python 2.6, I'll fix it here in a second

austinhyde:
@bcoca I hate to nag, but I was just wondering what your plans were regarding this PR. I have a plugin that depends on this for 2.0 compatibility, but if that's not going to happen, I'd like to know sooner rather than later, so I can implement a workaround.

Is there anything I can do to help move this along?

bcoca:
clone core devs ...

kidding aside, we need to discuss this as we have different opinions on the approach, I'm leaning towards checking both the expected class and not the base class (or as a fallback).

johnreed00:
FreeBSD users need this commit hopefully. Later I've tried to pull request at system modules (service, cron, user, group) to make it work with jails, using chroot or jexec calls. But it's big work to support it when branch has conflicts before commit. So I just added forks of the modules, but it's loss of new features.

bcoca:
@evlyapin how does FreeBSD factor into python class checks?

johnreed00:
@bcoca the reason is here: 
https://github.com/austinhyde/ansible-sshjail/issues/10
https://github.com/austinhyde/ansible-sshjail/pull/8

nitzmahone:
closing for inactivity/applicability

PR #14221, opened 2016-01-30
Some attributes of callbacks aren't in v2.  Port plugins to the v2 way to do that.

abadger:
Also update porting guide with info on callback porting


bcoca:
LGTM

PR #14223, opened 2016-01-30
Update intro_dynamic_inventory.rst

colynn:
vagrant@dev:~/ansible_tasks/inventory$ ./ec2.py --profile dev
usage: ec2.py [-h] [--list] [--host HOST] [--refresh-cache]
              [--boto-profile BOTO_PROFILE]
ec2.py: error: unrecognized arguments: --profile dev

vagrant@dev:~/ansible_tasks/inventory$ ./ec2.py --boto-profile dev


bcoca:
So this was broken by #9130 as it made a backwards incompatible change, I added back the --profile option via 37eac4592e2bf2814c2325b57042fee26ccdcd8a. Since the old script is still in use by most users I'm not going to merge these docs as I believe it will be more confusing going forward.

Issue #14112, opened 2016-01-25
v2.0: sluggish playbook run with role inter-dependencies


zarnovican:
##### Issue Type:
- Bug Report (performance / resource usage)
##### Ansible Version:

```
ansible 2.0.0.2
  config file = /home/zarnovic/.ansible.cfg
  configured module search path = Default w/o overrides
```
##### Ansible Configuration:

No changes
##### Environment:

N/A
##### Summary:

With more complex role dependency graph, time to execute playbook is growing non-linearly.

This became obvious after switching to Ansible 2.0 with our playbook. Time to execute jumped from 14s to 45s. After role dependencies were removed and all roles explicitly included in playbook, execution time between v2.0 vs v1.9.4 was acceptable again.
##### Steps To Reproduce:

Create a set of 7 roles, that depend on each other. Every role N, depends on all roles 1..N-1 (a triangle). In playbook, include only the last role. Here is bash script to generate roles and playbook:

``` bash
N=7
for i in `seq $N`; do
    mkdir -p roles/role$i/{tasks,meta}
    printf -- '---\n- debug: msg="task role'$i'"\n' > roles/role$i/tasks/main.yaml
    if [ $i -gt 1 ]; then
        printf -- '---\ndependencies:\n' > roles/role$i/meta/main.yaml
    fi
    for j in `seq $((i-1))`; do
        echo " - role$j" >> roles/role$i/meta/main.yaml
    done
done

cat >site.yaml <<EOF

---
- name: 'playbook'
  hosts: all
  gather_facts: no

  roles:
    - role$N
EOF
```

With higher number N, ansible will run out of memory. With number N too low, the time difference is statistically insignificant.
##### Workaround

Put all roles in playbook, in the order you would expect them to be executed. Remove inter-role dependencies.

```
rm -rf roles/role*/meta
cat >site.yaml <<EOF

---
- name: 'playbook'
  hosts: all
  gather_facts: no

  roles:
    - role1
    - role2
    - role3
    - role4
    - role5
    - role6
    - role7
EOF
```
##### Expected Results:

Ansible 1.9.4:

```
$ time ansible-playbook -i foo.ini site.yaml >/dev/null
real    0m0.350s
user    0m0.288s
sys 0m0.065s
```
##### Actual Results:

Ansible 2.0:

```
$ time ansible-playbook -i foo.ini site.yaml >/dev/null
real    0m15.303s
user    0m14.952s                            <-- notice high user CPU
sys 0m1.087s
```

[...]

PR #14224, opened 2016-01-30
Do not append parent more than once when calling serialize.

Yannig:
Can result in a 10 time speed improvement when using role with redundant dependencies (check https://github.com/ansible/ansible/issues/14112 for more details).

### DIFF
--- a/lib/ansible/playbook/role/__init__.py
+++ b/lib/ansible/playbook/role/__init__.py
@@ -354,7 +354,7 @@ class Role(Base, Become, Conditional, Taggable):
 
         return block_list
 
-    def serialize(self, include_deps=True):
+    def serialize(self, include_deps=True, already_included_roles=[]):
         res = super(Role, self).serialize()
 
         res['_role_name']    = self._role_name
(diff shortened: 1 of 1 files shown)


lib/ansible/playbook/role/__init__.py
```diff
@@ -376,7 +376,10 @@ def serialize(self, include_deps=True):
 
         parents = []
         for parent in self._parents:
-            parents.append(parent.serialize(include_deps=False))
+            # include parent only once
+            # Fix https://github.com/ansible/ansible/issues/14112
+            if parent not in self._parents:
```
jimi-c: Should this be `if parent not in parents:` instead? Otherwise, I do not see how this helps anything.

Yannig: @jimi-c Well, you're right ... This condition is always false so it's quite normal to get a huge speed improvement.

Yannig: But it's strange to see that there's no regression in test suite.

Yannig:
@jimi-c this time the commit is a little bit smarter (which is not difficult indeed ...). I'm using an array of string to keep track of already included role.

Yannig:
The improvement is still important (1s vs 14s without this patch).

thiagocotting:
Do we have an estimative when this will be merged? It's really painful to run roles with dependencies without this change

jimi-c:
@Yannig in my performance branch, we actually don't even serialize anymore: https://github.com/ansible/ansible/compare/performance_improvements

Yannig:
No more relevant in regard of Ansible performance in 2.2.

PR #14225, opened 2016-01-30
Should say ansible_ssh_user instead of ansible_user

jonatanblue:
In the inventory you should use ansible_ssh_user to specify the default SSH user for a host.

```
    myhost ansible_ssh_user=ubuntu
```

Tested with Ansible versions 1.7.1, 1.9.3 and 2.0.0.1.

### DIFF
--- a/docsite/rst/intro_inventory.rst
+++ b/docsite/rst/intro_inventory.rst
@@ -74,8 +74,8 @@ You can also select the connection type and user on a per host basis:
    [targets]
 
    localhost              ansible_connection=local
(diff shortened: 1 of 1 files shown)


sivel:
Many of the `_ssh_` values have been deprecated in favor of those without it.  This is described elsewhere in the documentation.  As such we are documenting the current value recommended with the current version of Ansible.

See https://github.com/ansible/ansible/blob/devel/docsite/rst/ansible_ssh_changes_note.rst

jonatanblue:
Sorry about that, I should've read through the whole document (and seen the note about this) before submitting a fix. Thanks for pointing me in the right direction.

PR #14234, opened 2016-02-01
Fix all permission/mode examples to use strings for octal representation

dagwieers:
This is required because yaml and ansible v2 represents this as decimal values internally. And as such it fails with an incorrect value when run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,8 +99,8 @@ allowed in future versions:
       gather_facts: false
       vars:
         my_dirs:
-          - { path: /tmp/3a, state: directory, mode: 0755 }
-          - { path: /tmp/3b, state: directory, mode: 0700 }
+          - { path: /tmp/3a, state: directory, mode: '0755' }
+          - { path: /tmp/3b, state: directory, mode: '0700' }
       tasks:
         - file:
           args: "{{item}}"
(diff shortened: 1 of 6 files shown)


mgedmin:
What?  No: YAML understands octal literals.

What breaks is when you use `mode: 755` instead of `mode: 0755` or `mode: "755"`.

dagwieers:
YAML understands octal literals, yes. But they are internally represented as scalars and Ansible v2 replaces them by decimal notation, just as yaml.dump() does.

Have you tested it with Ansible v2 lately ?

mgedmin:
Re: testing -- yes, one of my playbooks has the following task

``` yaml
- name: make sudo preserve SSH_AUTH_SOCK
  copy:
    dest: /etc/sudoers.d/pov
    content: |
      Defaults env_keep+=SSH_AUTH_SOCK
    mode: 0440
    validate: visudo -cf %s
  tags: sudo
  notify: changelog sudo ssh-agent fix
```

and ansible 2.0.0.2 in `--check` mode doesn't show any changes for the existing file, which has mode `-r--r-----`.

Can you provide an example where using an octal literal in a playbook results in incorrect behavior with Ansible 2?

dagwieers:
I'll re-open when I have the exact example from work that failed.

PR #14235, opened 2016-02-01
Groups naming convention in vagrant hosts provider

erikzaadi:
### Add Grouping by naming convention capabilities:

If the file `vagrant.ini` is present, and contains a regex:

``` ini
[vagrant]
host_match_regex = '(.*?)(?:-[0-9]+)?$'
```

Then any vagrant machine with the following naming convention will be grouped:

Machine: `<NAME>-X` => Group: `<NAME>`

This is in addition to the previously used `vagrant` group.

If the `vagrant.ini` file doesn't exist (or the needed option) then the vagrant host provider will continue as before.

Any group caught in the regex will be used as a host group.

### DIFF
--- a/contrib/inventory/vagrant.ini
+++ b/contrib/inventory/vagrant.ini
@@ -1,0 +1,2 @@
+[vagrant]
+#host_match_regex = '(.*?)(?:-[0-9]+)?$' #host name to group convention (Name: <BLA>-N => Group: <BLA>)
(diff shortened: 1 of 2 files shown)


bcoca:
this would break many an inventory that uses the - but does not follow this convention

erikzaadi:
@bcoca Perhaps if the host pattern was in an additional config file?

erikzaadi:
@bcoca updated.

gundalow:
@erikzaadi 
Thank you for your PR.
As part of reviewing the backlog of PRs we are looking at PRs older PRs that haven't been updated in a while

Given that:
* PR hasn't been updated since 2016
* Needed a rebase since end of 2016
* This is a feature rather than a bug fix

Therefore I'm going to close this.

If you or anyone else wants to continue with this work then please do feel free to create a fresh PR and `@mention` the previous reviewers in here.

PR #14241, opened 2016-02-01
cleanup unused code after #14220

robinro:
- delete unused variables: node, host, limit_opts, inv_opts
- remove duplicate `-c local` in base_opts and cmd

PR #14242, opened 2016-02-01
don't pass default repo to ansible-playbook on pull

### LINKED ISSUE #13688: ansible-pull does not respect ansible.cfg in pulled repo

robinro:
- fixes #13688
- fixes #14241 (removes the same unused things)

### DIFF
--- a/lib/ansible/cli/pull.py
+++ b/lib/ansible/cli/pull.py
@@ -30,6 +30,7 @@ import time
 
 from ansible.errors import AnsibleOptionsError
 from ansible.cli import CLI
+from ansible import constants as C
 from ansible.plugins import module_loader
 from ansible.utils.cmd_functions import run_cmd
 
@@ -130,17 +131,14 @@ class PullCLI(CLI):
         # Now construct the ansible command
         node = platform.node()
         host = socket.getfqdn()
-        limit_opts = 'localhost,%s,127.0.0.1' % ','.join(set([host, node, host.split('.')[0], node.split('.')[0]]))
+	# C.LOCALHOST is 'localhost', '127.0.0.1', '::1'
+        aliases = set(C.LOCALHOST)
+        aliases |= set([host, node, host.split('.')[0], node.split('.')[0]])
+	# limit_opts contains limit for ansible-playbook that has all possible names of local machine
+        limit_opts = ','.join(aliases)
         base_opts = '-c local '
         if self.options.verbosity > 0:
             base_opts += ' -%s' % ''.join([ "v" for x in range(0, self.options.verbosity) ])
-
-        # Attempt to use the inventory passed in as an argument
-        # It might not yet have been downloaded so use localhost if note
-        if not self.options.inventory or not os.path.exists(self.options.inventory):
-            inv_opts = 'localhost,'
-        else:
-            inv_opts = self.options.inventory
 
         #FIXME: enable more repo modules hg/svn?
         if self.options.module_name == 'git':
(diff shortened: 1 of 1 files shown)


bcoca:
i forget the reason, but I remember it was important to put fqdn between localhost and 127.0.0.1 ...

robinro:
Then at least add `::1`
The order should not matter for `--limit`.

robinro:
The old `localhost:%s:127.0.0.1` was added here: 4ae3bee41626e182bf687bee49df1b1909649a9b. I don't see a mention of the need for the special ordering.

robinro:
And we use fqdn in our playbook and it works for me if the fqdn is last.

bcoca:
that is the problem of not having good comments/commits ... i do remember it was significant for some reason ... might not be an issue anymore. 

that aside, I just realized we used to never have a default inventory .... looking at that, as it seems to cause more problems than it would fix.

robinro:
At least now we have good comments :smile:
The discussion about this got quite verbose.

bcoca:
ah, /etc/ansible/hosts ... which should be the default ... imconfused ...

robinro:
Now it's probably too late to remove the default inventory as some people might rely on it.

robinro:
The solution here still breaks if someone passes `-i /etc/ansible/hosts` explicitly, but has something else inside the ansible.cfg of the repo. The intention is obviously to overwrite the ansible.cfg setting, but the code now does not pass `-i` to ansible-playbook, such that the ansible.cfg setting would be used.

To me it looks like changing the `cli/__init__.py` is the cleaner solution.

bcoca:
i'm starting to come around to that, having no default inventory for ansible-pull does simplify things a lot (then default to 'localhost,' for the SCM pull but pass nothing to the playbook run).

bcoca:
my latest patch:

```
diff --git a/lib/ansible/cli/__init__.py b/lib/ansible/cli/__init__.py
index a91c955..1bb0d96 100644
--- a/lib/ansible/cli/__init__.py
+++ b/lib/ansible/cli/__init__.py
@@ -224,7 +224,7 @@ class CLI(object):

         if inventory_opts:
             parser.add_option('-i', '--inventory-file', dest='inventory',
-                help="specify inventory host path (default=%s) or comma separated host list" % C.DEFAULT_HOST_LIST,
+                help="specify inventory host path (default=%s) or comma separated host list. Note that for ansible-pull the default is 'localhost,'" % C.DEFAULT_HOST_LIST,
                 default=C.DEFAULT_HOST_LIST, action="callback", callback=CLI.expand_tilde, type=str)
             parser.add_option('--list-hosts', dest='listhosts', action='store_true',
                 help='outputs a list of matching hosts; does not execute anything else')
diff --git a/lib/ansible/cli/pull.py b/lib/ansible/cli/pull.py
index e099555..0f9355f 100644
--- a/lib/ansible/cli/pull.py
+++ b/lib/ansible/cli/pull.py
@@ -92,6 +92,9 @@ class PullCLI(CLI):
             help='verify GPG signature of checked out commit, if it fails abort running the playbook.'
                  ' This needs the corresponding VCS module to support such an operation')

+        # for pull we don't wan't a default
+        self.parser.set_defaults(inventory=None)
+
         self.options, self.args = self.parser.parse_args(self.args[1:])

         if not self.options.dest:
@@ -136,8 +139,8 @@ class PullCLI(CLI):
             base_opts += ' -%s' % ''.join([ "v" for x in range(0, self.options.verbosity) ])

         # Attempt to use the inventory passed in as an argument
-        # It might not yet have been downloaded so use localhost if note
-        if not self.options.inventory or not os.path.exists(self.options.inventory):
+        # It might not yet have been downloaded so use localhost as default
+        if not self.options.inventory or ( ',' not in self.options.inventory and not os.path.exists(self.options.inventory)):
             inv_opts = 'localhost,'
         else:
             inv_opts = self.options.inventory
@@ -166,7 +169,7 @@ class PullCLI(CLI):

         bin_path = os.path.dirname(os.path.abspath(sys.argv[0]))
         # hardcode local and inventory/host as this is just meant to fetch the repo
-        cmd = '%s/ansible -i "localhost," -c local %s -m %s -a "%s" all' % (bin_path, base_opts, self.options.module_name, repo_opts)
+        cmd = '%s/ansible -i "%s" %s -m %s -a "%s" all' % (bin_path, inv_opts, base_opts, self.options.module_name, repo_opts)

         for ev in self.options.extra_vars:
             cmd += ' -e "%s"' % ev
```

bcoca:
this should besuperceded by #14251

PR #14245, opened 2016-02-01
If specified non-standard ssh port is not open, revert to default ssh…

crlb:
… port.

This pull request supersedes #13580 & #13589.

### DIFF
--- a/lib/ansible/plugins/connection/ssh.py
+++ b/lib/ansible/plugins/connection/ssh.py
@@ -25,6 +25,7 @@ import pipes
 import pty
 import select
 import shlex
+import socket
 import subprocess
 import time
 
@@ -172,10 +173,17 @@ class Connection(ConnectionBase):
             )
 
         if self._play_context.port is not None:
-            self._add_args(
-                "ANSIBLE_REMOTE_PORT/remote_port/ansible_port set",
-                ("-o", "Port={0}".format(self._play_context.port))
-            )
+            test_port_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            test_port_result = test_port_socket.connect_ex( (self.host, self._play_context.port) )
+            test_port_socket.close()
+
+            if test_port_result == 0:
+                self._add_args(
+                    "ANSIBLE_REMOTE_PORT/remote_port/ansible_port set",
+                    ("-o", "Port={0}".format(self._play_context.port))
+                )
+            else:
+                display.warning('SSH unable to connect to %s:%d, reverting to default port' % (self.host, self._play_context.port))
 
         key = self._play_context.private_key_file
         if key:


sivel:
I'm not sold on this functionality.  It's not very explicit, and let's say you have 2 ssh daemons, or even a port forward on the remote box, connecting to 22 as opposed to the explicitly stated port could cause a lot of trouble.

If anything, this should be opt in behavior.

For now I am -1

crlb:
Do you have a practical alternative?

On 16-02-01 08:56 AM, Matt Martz wrote:

> I'm not sold on this functionality. It's not very explicit, and let's 
> say you have 2 ssh daemons, or even a port forward on the remote box, 
> connecting to 22 as opposed to the explicitly stated port could cause 
> a lot of trouble.
> 
> If anything, this should be opt in behavior.
> 
> For now I am -1
> 
> —
> https://github.com/ansible/ansible/pull/14245#issuecomment-178065563.

## 

Colin Leavett-Brown
Physics and Astronomy
University of Victoria
250-472-4085

sivel:
I would say that it shouldn't be done at all.  You should be explicit, and fail if your explicitly defined port is incorrect.

Another option is to handle this via tasks, such as using a play that does not gather facts, using `wait_for` to check if the explicitly defined port is open (ignoring errors), and if not, define a new correct port (using a when statement and validating failure of the `wait_for`).

I do not believe we should just be magically doing this behind the scenes.  It should be an explicit choice by the user and implemented playbooks to do this, not performing a potentially unexpected action automatically.

bcoca:
I agree with @sivel, this is not something ansible should do and you can handle it with tasks.

So I'm sorry to say we are going to pass on this feature.

crlb:
I am specifying an explicit non-standard port in the inventory which I 
want ansible to change the node to. If I put the wait_for/when logic 
that you suggest at the top of my playbook, to explicitly override the 
not open, non-standard port then the playbook will fail as soon as the 
port changes. Effectively, I have to put this logic in at least two 
places in the playbook, which makes the port specification in the 
inventory completely redundant.

I would be happy to have or add opt-in functionality. What would you 
suggest?
- "revert_to_default_ssh_port: true/false" in the config file? (port
  22 is not necessarily the default ssh port) or
- alternate port in the inventory, eg. "host:9000/2200",
  "host:9000/default", etc.?

What do you think, is there a compromise? Colin.

On 16-02-01 09:26 AM, Matt Martz wrote:

> I would say that it shouldn't be done at all. You should be explicit, 
> and fail if your explicitly defined port is incorrect.
> 
> Another option is to handle this via tasks, such as using a play that 
> does not gather facts, using |wait_for| to check if the explicitly 
> defined port is open (ignoring errors), and if not, define a new 
> correct port (using a when statement and validating failure of the 
> |wait_for|).
> 
> I do not believe we should just be magically doing this behind the 
> scenes. It should be an explicit choice by the user and implemented 
> playbooks to do this, not performing a potentially unexpected action 
> automatically.
> 
> —
> https://github.com/ansible/ansible/pull/14245#issuecomment-178080873.

## 

Colin Leavett-Brown
Physics and Astronomy
University of Victoria
250-472-4085

abadger:
Changing the port is conceptually a provisioning step (setting up the box so that it can run basic necessities like ansible) rather than ongoing configuration and deployment.  So if I was designing your playbooks I'd probably continue to have the nonstandard port in inventory and override it in a one-time provisioning play where I lay down the alternate sshd config and restart the ssh server.

I don't think that either of the two ideas you have (config file or alternative inventory) are things we want either.  Needing a second port setting for a provisioning step fits into the conceptual big picture and adding features can make it harder to maintain code and explain the niche cases of how something works to others.

Sorry to say that we aren't going to do this but didn't want to leave you hanging.

crlb:
Hello Toshio, Thank you for your update, it is much appreciated.

I have a fork and very simple demo at: 
https://github.com/crlb/ansible/tree/stable-2.0.0.1/sample_ssh_port_setting; 
you should try it. It has an inventory of two host with different 
non-standard ports (something we actually want to do). It will even 
accommodate changing a host with a non-standard port to a different port 
(also something we might want to do). I think this considerably 
simplifies (a design goal of ansible) the port changing process with no 
reasonable down sides. It makes no changes on target hosts (playbooks 
still do that) but merely enables communication between the deployment 
host and target hosts, which would otherwise fail.

I think ansible is great, but I will continue to use the patch. I think 
the evidence on various lists show that others might find it useful too. 
To re-iterate,
_it makes no changes on target hosts but merely enables communication_, 
and what is the objection to that?

Regards, Colin, a happy patched-Ansible (7 lines of code) user.

On 16-02-17 12:56 AM, Toshio Kuratomi wrote:

> Changing the port is conceptually a provisioning step (setting up the 
> box so that it can run basic necessities like ansible) rather than 
> ongoing configuration and deployment. So if I was designing your 
> playbooks I'd probably continue to have the nonstandard port in 
> inventory and override it in a one-time provisioning play where I lay 
> down the alternate sshd config and restart the ssh server.
> 
> I don't think that either of the two ideas you have (config file or 
> alternative inventory) are things we want either. Needing a second 
> port setting for a provisioning step fits into the conceptual big 
> picture and adding features can make it harder to maintain code and 
> explain the niche cases of how something works to others.
> 
> Sorry to say that we aren't going to do this but didn't want to leave 
> you hanging.
> 
> —
> https://github.com/ansible/ansible/pull/14245#issuecomment-185103153.

## 

Colin Leavett-Brown
Physics and Astronomy
University of Victoria
250-472-4085

amenonsen:
> I think this considerably simplifies (a design goal of ansible) the
> port changing process with no reasonable down sides.

Simplifying things is certainly a design goal, but there's also a strong
preference for being explicit rather than magical, and this patch feels
too magical to me. In particular, if for some reason the host _isn't_
listening on the non-default port you expect (for example, because you
confused it with a different host that is provisioned differently), this
would cause it to go happily onwards rather than fail.

Also if you're using a ProxyCommand via ansible_ssh_common_args to
connect to hosts behind a jump box, your patch would incorrectly
conclude that the port isn't open (because there's no direct
connectivity to the target host).

Maybe these aren't concerns in your own environment, but they are
possibilities that the Ansible maintainers need to take into account.

crlb:
I'm not quite sure why you guys keep bothering me, you've already 
collectively said that you are not interested in the patch. However, 
because you keep telling me how stupid I am, I will defend myself:

On 16-02-17 08:06 PM, Abhijit Menon-Sen wrote:

> > I think this considerably simplifies (a design goal of ansible) the
> > port changing process with no reasonable down sides.
> 
> Simplifying things is certainly a design goal, but there's also a strong
> preference for being explicit rather than magical, and this patch feels
> too magical to me. In particular, if for some reason the host _isn't_
> listening on the non-default port you expect (for example, because you
> confused it with a different host that is provisioned differently), this
> would cause it to go happily onwards rather than fail.
> Checking a port before using it, "too magical" -  Hmmm!
> 
> Also if you're using a ProxyCommand via ansible_ssh_common_args to
> connect to hosts behind a jump box, your patch would incorrectly
> conclude that the port isn't open (because there's no direct
> connectivity to the target host).
> So what? The patch merely tests the port and if it fails omits the "-p" 
> option from the ssh command. However, if a ProxyCommand is used via 
> ansible_ssh_common_args='-o ProxyCommand="ssh -W %h:%p 
> someuser@jumphost.example.com"', then any "-p" option is ignored:

```
Example 1:

    crlb@heplw49:~$ ssh -o ProxyCommand="ssh -p3121 -W e1:3121
    root@elephant.heprc.uvic.ca" date
    Last login: Thu Feb 18 08:25:48 2016 from mouse10.cobbler
    [crlb@mouse01 ~]$

Example 2:

    crlb@heplw49:~$ ssh -p 9999 -o ProxyCommand="ssh -p3121 -W
    e1:3121 root@elephant.heprc.uvic.ca" dateLast login: Thu Feb 18
    08:25:32 2016 from mouse10.cobbler
    [crlb@mouse01 ~]$
```

So in this case, the patch is irrelevant (it leaves out a parameter that 
would otherwise be inserted but ignored).

> Maybe these aren't concerns in your own environment, but they are
> possibilities that the Ansible maintainers need to take into account.
> 
> —
> https://github.com/ansible/ansible/pull/14245#issuecomment-185534053.
> 
> Glad to hear it.

I know you guys are really clever because ansible wouldn't be so good if 
you weren't, but to me this is an obvious flaw. And that being the case, 
I will continue to use my fork, even with my jump boxes. Colin.

By the way, Did anyone actually try the demo? I think not.

## 

Colin Leavett-Brown
Physics and Astronomy
University of Victoria
250-472-4085

bcoca:
@crlb I don't think anyone wanted to imply you were stupid, they are just trying to explain their position on why they do not think this belongs in Ansible. 

Criticism of the feature is not a personal attack, I understand why you think this should be included but I think you can see that the opinion of most maintainers is that it should not (history will decide if we are wrong).

PR #14246, opened 2016-02-01
Add option to group ec2 instances by platform.

felipecvo:
##### SUMMARY

We have instances windows and linux in our environment, and for easy configuration for access, we need to group all windows instance together.

This PR allows me to:

*inventory/base.ini*

```
[platform_windows]

[windows:children]
platform_windows
```

And *inventory/group_vars/windows.yml*

``` yaml
ansible_user: Administrator
ansible_password: SecretPasswordGoesHere
ansible_port: 5986
ansible_connection: winrm
ansible_winrm_server_cert_validation: ignore
```

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
contrib/inventory/ec2.py

##### ANSIBLE VERSION
2.3

### DIFF
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -419,6 +419,7 @@ class Ec2Inventory(object):
             'group_by_ami_id',
             'group_by_instance_type',
             'group_by_instance_state',
+            'group_by_platform',
             'group_by_key_pair',
             'group_by_vpc_id',
             'group_by_security_group',
@@ -880,6 +881,16 @@ class Ec2Inventory(object):
             self.push(self.inventory, state_name, hostname)
             if self.nested_groups:
                 self.push_group(self.inventory, 'instance_states', state_name)
+
+        # Inventory: Group by platform
+        if self.group_by_platform:
+            if instance.platform:
+                platform = self.to_safe('platform_' + instance.platform)
+            else:
+                platform = self.to_safe('platform_undefined')
+            self.push(self.inventory, platform, dest)
+            if self.nested_groups:
+                self.push_group(self.inventory, 'platforms', platform)
 
         # Inventory: Group by key pair
         if self.group_by_key_pair and instance.key_name:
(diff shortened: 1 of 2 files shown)


contrib/inventory/ec2.py
```diff
@@ -881,6 +882,16 @@ def add_instance(self, instance, region):
             if self.nested_groups:
                 self.push_group(self.inventory, 'instance_states', state_name)
 
+        # Inventory: Group by platform
+        if self.group_by_platform:
+            if instance.platform:
+                platform = self.to_safe('platform_' + instance.platform)
+            else:
+                platform = self.to_safe('platform_undefined')
+            self.push(self.inventory, platform, dest)
```
s-hertel, line 891: Why isn't this `self.push(self.inventory, platform, hostname)`?

felipecvo: @s-hertel I did this a long time ago, and honestly, I have no idea why... All that I can say is I used this in production, and it worked.

s-hertel: I'm going to add a commit to change this to hostname. hostname defaults to dest if it isn't defined, so I believe the behavior will stay the same and also then maintain consistency with the surrounding code.

s-hertel: I think this should be hostname, since hostname defaults to dest ([line 825](https://github.com/ansible/ansible/pull/14246/files#diff-c2e371537339bbab5d9d0cc5d6cb7adcL825)) if it isn't defined. That way this maintains consistency with the surrounding code. Should hostname be defined, I think it might be confusing. That was my only feedback though, this looks good.

s-hertel: I cannot add commits to this PR to fix this myself, so I'm updated this to request changes.

s-hertel (changes requested):

default50:
@felipecvo thanks for your PR, found myself in the same spot as you and been using this successfully! I've patched ec2.py with the suggestion by @s-hertel (`self.push(self.inventory, platform, hostname)`) and all OK.

Will you be able to do the modification required? I'd like this to get merged ASAP. I can create a new PR if you are not able to commit time to this.

Thanks again!

calfonso:
duplicate_of #27848

PR #14251, opened 2016-02-02
better fix for ansible-pull inventory handling

### LINKED ISSUE #13688: ansible-pull does not respect ansible.cfg in pulled repo

bcoca:
now it mirrors what it did in previous versions

alt to #14242 for:
fixes #13688
fixes #14241


robinro:
This does not work. `limit_opts` is never used, so my setup without `-l` and `-i` ends up running agains `all` (which leads to errors).

robinro:
Thanks @bcoca for the long discussion around this issue and merging this fix. The devel version now works well in our staging environment.

PR #14253, opened 2016-02-02
Allow key authentication when using `--ask-pass` (just like Ansible v2)

### LINKED ISSUE #14250: SSH publickey and password combination not allowed
If you use `--ask-pass` with ansible-playbook, Ansible appears to try and be clever by assuming that you do not want to use publickey authentication, and adds `-o PubkeyAuthentication=no` to the ssh command line.

At best, this is misleading, at worst it is wrong and prevents specific edge use cases.

It is perfectly valid for an SSH server to REQUIRE that the client provide a publickey AND a password, or more commonly, that the client provide a publickey AND a 2-factor OTP via keyboard-interactive.

Consider the sshd_config configuration:
`AuthenticationMethods publickey,password`

And ssh_config:
`PreferredAuthentications=publickey,password`

At present, the above SSH configuration is not supported by Ansible.

```
<test1> ESTABLISH CONNECTION FOR USER: testuser
<test1> REMOTE_MODULE setup
<test1> EXEC ['sshpass', '-d4', 'ssh', '-C', '-tt', '-vvv', '-o', 'PreferredAuthentications=publickey,password', '-o', 'Port=22', '-o', 'GSSAPIAuthentication=no', '-o', 'PubkeyAuthentication=no', '-o', 'ConnectTimeout=10', 'test1', "/bin/sh -c 'mkdir -p $HOME/.ansible/tmp/ansible-tmp-1454369482.89-90417182374711 && chmod a+rx $HOME/.ansible/tmp/ansible-tmp-1454369482.89-90417182374711 && echo $HOME/.ansible/tmp/ansible-tmp-1454369482.89-90417182374711'"]
```

dagwieers:
This closes #14250.

It should not have any ill-effects for existing use-cases as we would only allow additional authentication methods on top of password authentication. And since the user can authenticate in other ways already, it also has no security impact.

### DIFF
--- a/lib/ansible/runner/connection_plugins/ssh.py
+++ b/lib/ansible/runner/connection_plugins/ssh.py
@@ -91,10 +91,7 @@ class Connection(object):
             self.common_args += ["-o", "IdentityFile=\"%s\"" % os.path.expanduser(self.private_key_file)]
         elif self.runner.private_key_file is not None:
             self.common_args += ["-o", "IdentityFile=\"%s\"" % os.path.expanduser(self.runner.private_key_file)]
-        if self.password:
-            self.common_args += ["-o", "GSSAPIAuthentication=no",
-                                 "-o", "PubkeyAuthentication=no"]
-        else:
+        if not self.password:
             self.common_args += ["-o", "KbdInteractiveAuthentication=no",
                                  "-o", "PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey",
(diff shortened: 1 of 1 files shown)


dagwieers:
This fix (or feature request as you will) was tested by the user and confirmed to work.
See the original issue ticket for the details and comparison of behaviour between Ansible v1.9 and Ansible v2.

dagwieers:
Thanks !

PR #14257, opened 2016-02-02
add integration test for ansible-pull

robinro:
- this test includes the behavior described in #13688 and #13681
- it runs a minimal playbook and checks for correct use of inventory and limit

### DIFF
--- a/test/integration/Makefile
+++ b/test/integration/Makefile
@@ -21,7 +21,7 @@ VAULT_PASSWORD_FILE = vault-password
 
 CONSUL_RUNNING := $(shell python consul_running.py)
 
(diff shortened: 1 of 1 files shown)


robinro:
@bcoca 
This adds integration tests for ansible-pull, so we get some reproducibility and documentation of our discussion around #13681 and #13688 

Your pull-request in #14251 fails with this tests, while #14242 works for me.

robinro:
to run the new tests do `cd test/integration; make pull`

bcoca:
i don't think your tests are correct, the fail for me with both patches, in any case i would setup something in ansible's repo, already made a mistake with having 3rd party repos in galaxy (fixing that is on my list).

Though i might just use this PR and modify it to use an ansible/ repo, so i'm keeping it open for now.

bcoca:
adding test 'as is' but will change to use local repo (optionally with ansible/remote one if tag supplied)

PR #14260, opened 2016-02-02
No need to creat the Play instance in adhoc script

jeffrey4l:
the play instance is generated by a factory method


bcoca:
yes but inheritance of base attributes get's done in the init.

abadger:
Although this shouldn't cause any problems, I think we're agreed that really there's no reason to have factory functions here.  So instead of this we should merge the load() method into **init**().  If you'd like to submit a PR to do that we can look at merging that.

PR #14261, opened 2016-02-02
Add validate_certs param to skip SSL verification in VMware

kamsz:
This is handy when working with development environments using self-signed certificates.

### DIFF
--- a/lib/ansible/module_utils/vmware.py
+++ b/lib/ansible/module_utils/vmware.py
@@ -105,7 +105,7 @@ def vmware_argument_spec():
         hostname=dict(type='str', required=True),
         username=dict(type='str', aliases=['user', 'admin'], required=True),
         password=dict(type='str', aliases=['pass', 'pwd'], required=True, no_log=True),
-        skip_ssl=dict(type='bool', required=False, default=False),
+        validate_certs=dict(type='bool', required=False, default=True),
     )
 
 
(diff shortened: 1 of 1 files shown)


bcoca:
We use the name 'validate_certs' for this kind of option, its nice to have it consistent across plugins.

kamsz:
Changed to validate_certs.

abadger:
Looks like this requires pyvmomi and that requires python 2.7.0 or greater.  Unfortunately, SSLContext is only available on 2.7.9+ (or where the distributor of your python has backported it to an earlier version) .  So we can't use this unconditionally.  We'll have to make it conditional on whether the stdlib supports it or not. 

We've recently added validate_certs to vsphere_guest and you should be able to do something similar: https://github.com/ansible/ansible-modules-core/blob/devel/cloud/vmware/vsphere_guest.py

The portion that you have to come up with an equivalent of is right around here: https://github.com/ansible/ansible-modules-core/blob/devel/cloud/vmware/vsphere_guest.py#L1640

kamsz:
@abadger I've added verification of SSLContext in the ssl library.

abadger:
@kamsz Thanks!  Sorry, I wasn't too clear before:

In general we want to be secure by default.  that means that if `validate_certs==True` we want to be sure that we are validating certificates.  With the current PR if we're on python < 2.7.9 and `validate_certs==True` python will silently not validate certificates here.  So we need to check for SSLContext earlier in order to decide whether we need to fail because python is too old and validate_certs is True.

The code paths to handle are:
- validate_certs==True; python with SSLContext: Should work without special code
- validate_certs==True; python without SSLContext; need to fail and tell the user to either upgrade python or use validate_certs=False as this is an older version of python that does not support verifying certificates
- validate_certs==False; python with SSLContext; need to create an SSLContext with verify_mode = ssl.CERT_NONE as you are doing
- validate_certs==False; python without SSLContext; Should work without special code as older version of python do not verify certificates.

Also, noticed that this branch has conflicts due to commiting the submodule refs.  Probably easiest to git reset before the last commit, add the code to handle the above variations, commit the new change, and then force push to update the branch.

```
$ git reset fa13aa8c007ac7bb3534626d3312e05fee2c7190
$ # make new changes to the file
$ git commi vmware.py
$ git push --force
```

kamsz:
@abadger I hope the latest change reflects what you had in mind :)

abadger:
@kamsz Thanks for adding this feature!  merged to devel.  It should be in 2.1.0, the next major release.

PR #14273, opened 2016-02-02
[2.1] cloudstack: new generic code

resmo:


lib/ansible/module_utils/cloudstack.py
```diff
@@ -125,6 +137,17 @@ def get_or_fallback(self, key=None, fallback_key=None):
...
+        if not required_params:
+            return
+        missing_params = []
+        for required_param in required_params:
+            if not self.module.params.get(required_param):
+                missing_params.append(required_param)
+        if missing_params:
+            self.module.fail_json(msg="missing required arguments: %s" % ','.join(missing_params))
+
+
```
abadger: - This seems like it would be generally useful.  Probably better to put it into basic.py:AnsibleModule().
- Short docstring that explains this is for checking for a parameter when we can't check via the argspec (because we need more information than is simply given in the argspec.)

resmo:
shipit

Issue #14176, opened 2016-01-28
Non-POSIX /bin/sh is not respected since v2.0


Akint:
Problem description:
Dollar parents ( $(...) ), which aren't supported on non-POSIX shells (for example, Solaris 10's one), have been used in lib/ansible/plugins/shell/sh.py since v2.0. For example, tmp dir is being created as "/export/home/akint/$( echo /export/home/akint/.ansible/tmp/" instead of "/export/home/akint/.ansible/tmp/" because of 'mkdir -p "$( echo $HOME/.ansible/tmp/ansible-tmp-1453972288.64-179689601296487 )"' during connection initialization. 

I have found 2 ways it could be fixed:
1. Use backticks ( `...` ) instead of dollar parents:
   
   ``` diff
         diff --git a/lib/ansible/plugins/shell/sh.py b/lib/ansible/plugins/shell/sh.py
         index 8b20338..16b715a 100644
         --- a/lib/ansible/plugins/shell/sh.py
         +++ b/lib/ansible/plugins/shell/sh.py
         @@ -35,8 +35,8 @@ class ShellModule(object):
              _SHELL_REDIRECT_ALLNULL = '> /dev/null 2>&1'
              _SHELL_AND = '&&'
              _SHELL_OR = '||'
         -    _SHELL_SUB_LEFT = '"$('
         -    _SHELL_SUB_RIGHT = ')"'
         +    _SHELL_SUB_LEFT = '"`'
         +    _SHELL_SUB_RIGHT = '`"'
              _SHELL_GROUP_LEFT = '('
              _SHELL_GROUP_RIGHT = ')'
   ```
2. Set "executable = /bin/bash" in ansible.cfg. It starts complaining: "Invalid shell type specified (bash), or the plugin for that shell type is missing.", but can be fixed by copying lib/ansible/plugins/shell/sh.py to lib/ansible/plugins/shell/bash.py

Described clauses have been tested with "ping" module:

```
ansible solaris10-host -m ping -vvvv
```

abadger:
As a temporatry workaround, does setting `executable = /bin/bash` in the config file and 

```
[all:vars]
ansible_shell_type=sh
```

In inventory work?

abadger:
[...]

PR #14277, opened 2016-02-02
Establish sh as the default shell plugin.

abadger:
This is a fix for one of the problems pointed out in #14176


bcoca:
+1

nitzmahone:
Testing now on Windows...

nitzmahone:
Had to add the last Windows fix patch first, but tests fine with both implicit and explicit ansible_shell_type. :+1:

abadger:
Thanks guys!  Merged to devel and stable-2.0

PR #14290, opened 2016-02-03
Stable 2.0: suggestions for improvement

cazorla19:
Last 3 days I tested my own [roles](https://github.com/cazorla19/ansible_play) and got much more acknowledge about Ansible. Many things are great, but also this product has places to improve. Here are my ideas. Unfortunately, I'm not a developer and have no idea about possibility of solution. But if it's possible - that's would be great to get it.

1) Templates inside of tasks name.
It's going to be eaesier for other Linux-systems combination. Now what I have to do for proper Apache installation:

```
- name: avoid services collision [CentOS/Red Hat]
  yum: name=nginx state=absent
  when: ansible_distribution == "CentOS" or ansible_distribution == "Red Hat Enterprise Linux"
  tags: remove

- name: avoid services collision [Debian/Ubuntu]
  apt: name=nginx state=absent
  when: ansible_distribution == "Debian" or ansible_distribution == "Ubuntu"
  tags: remove

- name: install apache [CentOS/Red Hat]
  yum: name=httpd state=latest
  when: ansible_distribution == "CentOS" or ansible_distribution == "Red Hat Enterprise Linux"
  tags: install

- name: install apache [Debian/Ubuntu]
  apt: name=apache2 state=latest
  when: ansible_distribution == "Debian" or ansible_distribution == "Ubuntu"
  tags: install
```

How it could be:

```
- name: avoid services collision
  {{ package_manager }}: name=nginx state=absent
  tags: remove

- name: install apache
  {{ package_manager }}: name=httpd state=latest
  tags: install
```

And, of course, assign `package_manager` variable in vars file.

2) Conditionals inside of vars_file.
The same reason. I forced compared Debian/Ubuntu customization with CentOS: there is a big difference in daemons name, groups name and something like that. What if we could indicate one bunch of variables for one kind of distro and another bunch for another distro? For instance:

```

---
[block1]
   apache_daemon: apache2
   apache_root_dir: /etc/apache2
   apache_modules: /etc/apache2/mods_enabled
   apache_config: /etc/apache2/apache2.conf
when: ansible_distribution == "Ubuntu"

[block1]
   apache_daemon: apache2
   apache_root_dir: /etc/httpd
   apache_modules: /etc/httpd/conf.d
   apache_config: /etc/httpd/conf/httpd.conf
when: ansible_distribution == "CentOS"
```

3) New correct "make" module for builds.
May be I'm mistaken but I don't see favourable module for compilations like "make". I mean creation of myriad files like it's going in our host system, when we build any software from sources. With command module and creates argument I can create one file/directory, but just one. That's not enough. For instance:

```
 - name: compile ffmpeg
   command: "{{ item }}" 
   args:
     chdir: "/opt/ffmpeg" 
     creates: "/home/user/ffmpeg_build"
   with_items:
     - PKG_CONFIG_PATH="{{ ffmpeg_lib_dir }}/pkgconfig" ./configure --prefix="{{ ffmpeg_build_dir }}" --extra-cflags="-I{{ ffmpeg_build_dir }}/include" --extra-ldflags="-L{{ ffmpeg_lib_dir }}" --bindir="{{ ffmpeg_bin_dir }}" --pkg-config-flags="--static" --enable-gpl --enable-nonfree --enable-libfdk-aac --enable-libfreetype --enable-libfontconfig --enable-libfribidi --enable-libx264
     - make
     - make install
     - make distclean
   tags: ffmpeg
```

This role creates one directory during the compilation (/home/user/ffmpeg_build), but only one! The good idea would be different module that can deal with multiple files creation.

Thanks for attention.


sivel:
So, a few things.  I am going to close this, as you opened a pull request instead of an issue.

Additionally, have a look at the package module, http://docs.ansible.com/ansible/package_module.html

sivel:
As for conditionals in vars files, you could use `include_vars` and include only a var for a specific distro.

As for a `make` module, something like that may be entertained in the future.

PR #14293, opened 2016-02-03
Defined JSON booleans in global context for python eval()

### LINKED ISSUE #14291: Ansible 2.0.0.2: Booleans in dictionaries cause to_json to fail
Very weird issue that is easy to reproduce.

If a dictionary has a boolean defined, `to_json` returns a `string` rather than a real `dict`.

The following playbook displays the problem:

``` yaml

---
- hosts: localhost
  gather_facts: no
  vars:
    works:
      key1: 'string'
      key2: 1234
    fails:
      key1: 'string'
      key2: 1234
      key3: false
  tasks:
  - debug: msg={{ works | to_json }}
  - debug: msg={{ fails | to_json }}
```

If you run this, you will see that the second dict fails:

```
[dag@moria ansible.dag]$ ansible-playbook test49.yml 

ok: [localhost] => {
    "msg": {
        "key1": "string", 
        "key2": 1234
    }
}

ok: [localhost] => {
    "msg": "{\"key3\": false, \"key2\": 1234, \"key1\": \"string\"}"
}

```

We experience this issue within roles where we assign a dictionary to a role variable.

dagwieers:
We define 'false' and 'true' as variables so that python eval() recognizes them as False and True.

This fixes #14291.

### DIFF
--- a/lib/ansible/template/safe_eval.py
+++ b/lib/ansible/template/safe_eval.py
@@ -41,6 +41,13 @@ def safe_eval(expr, locals={}, include_exceptions=False):
     http://stackoverflow.com/questions/12523516/using-ast-and-whitelists-to-make-pythons-eval-safe
     '''
 
+    # define certain JSON types
+    # eg. JSON booleans are unknown to python eval()
+    JSON_TYPES = {
+        'false': False,
+        'true': True,
+    }
+
     # this is the whitelist of AST nodes we are going to
     # allow in the evaluation. Any node type other than
     # those listed here will raise an exception in our custom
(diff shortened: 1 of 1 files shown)


dagwieers:
BTW This problem also exists in Ansible v1.9.4, so I will have to patch it there as well.
And I think it should be part of the 1.9.5 release.

@abadger @jimi-c would like to have your advice on the proposed fix.
Eg.
- Is the JSON_TYPES variable what you would like to see ?
- Should we use the global context as I did or the local context ?

Same for #14304, which will help display any other failed expression evaluation or compilation error.
I am literally waiting for your feedback :-)

PR #14297, opened 2016-02-03
Add rackHd inventory script

keedya:
- Allow ansible to get hosts inventory from rackHD by node id

### DIFF
--- a/contrib/inventory/rackhd.py
+++ b/contrib/inventory/rackhd.py
@@ -5,9 +5,9 @@ import os
 import argparse
 import types
 
-MONORAIL_URL = 'http://localhost:8080'
+RACKHD_URL = 'http://localhost:8080'
 
-class OnRackInventory(object):
+class RackhdInventory(object):
     def __init__(self, nodeids):
         self._inventory = {}
         for nodeid in nodeids:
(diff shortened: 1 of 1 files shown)


contrib/inventory/rackhd.py
```diff
@@ -0,0 +1,73 @@
...
+                ipaddress = node_info[0]["ipAddress"]
+            output += '  "' + nodeid + '" : {\n'
+            output += '    "hosts": [ "' + ipaddress + '" ],\n'
+            output += '    "vars" : {\n'
+            for key,result in info.iteritems():
+                output += '      "' + key + '": ' + json.dumps(json.loads(result), sort_keys=True, indent=2) + ',\n'
+            output += '      "ansible_ssh_user": "renasar"\n'
+            output += '    }\n'
+            output += '  }\n'
+        except KeyError:
```
sivel, line 40: I'm rather confused here. It looks like we are both using the `json` module to create JSON, as well as hand crafting JSON?

I strongly recommend, dropping this functionality, and building out an actual python datastrcuture, and just using `json.dumps` on the final result.

contrib/inventory/rackhd.py
```diff
@@ -0,0 +1,73 @@
...
+try:
+    nodeids += os.environ['nodeid'].split(',')
+except KeyError:
+    url = RACKHD_URL + '/api/common/nodes'
+    r = requests.get( url, verify=False)
+    data = json.loads(r.text)
+    for entry in data:
+        if entry['type'] == 'compute':
+            nodeids.append(entry['id'])
+RackhdInventory(nodeids)
```
sivel: This also looks like this inventory script blindly ignores either `--list` or `--host` and just does the one thing no matter what?

contrib/inventory/rackhd.py
```diff
@@ -0,0 +1,73 @@
...
+    def __init__(self, nodeids):
+        self._inventory = {}
+        for nodeid in nodeids:
+            self._load_inventory_data(nodeid)
+        output = '{\n'
+        for nodeid,info in self._inventory.iteritems():
+            output += self._format_output(nodeid, info)
+            output += ',\n'
+        output = output[:-2]
+        output += '}\n'
```
sivel: Looks like instead of creating a full python data structure, this code is trying to hand craft JSON via string concatenation.  See my other comment about just using a data structure and using `json.dumps` at the very end.

keedya:
@sivel Thanks For all the valuable comments!!

keedya:
@sivel Code is updated

keedya:
Adding a link to rackHD : https://github.com/RackHD

Issue #14285, opened 2016-02-03
Wishlist: hide 'include' tasks when not in verbose mode


mgedmin:
One of the changes in Ansible 2 is that it now reports all `include` tasks.

This is a bit painful, especially when you've a big playbook and use `--tags` to select just a few of the tasks -- all the include statements for roles irrelevant to what I'm debugging now tend to hide the actual tasks being run in between the includes.

Suggestion: hide `include` tasks by default.  Show them if an error happens, or if the user asks for more verbosity (by passing at least one `-v` to `ansible-playbook`).

bcoca:
so going to close this as a no, it is simple to override the default callback to do this and the actual include issue should go away once we fix how inclusion and tags work.

mgedmin:
> the actual include issue should go away once we fix how inclusion and tags work

This is very interesting!  Do you have a bug open describing the planned changes, or something like that?

bcoca:
just the report, no pr yet #13507

PR #14300, opened 2016-02-03
Hide 'include' tasks when not in verbose mode

dagwieers:
This implements #14285 by default.

It will hide all include-tasks by default and in verbosity 1 (-v), but will show all details when verbosty 2 or higher (-vv).

It is quite annoying and confusing that Ansible emits all include tasks. Also because the color is the same color as skipping.

### DIFF
--- a/lib/ansible/plugins/callback/default.py
+++ b/lib/ansible/plugins/callback/default.py
@@ -114,23 +114,24 @@ class CallbackModule(CallbackBase):
         self._display.banner("NO MORE HOSTS LEFT")
 
     def v2_playbook_on_task_start(self, task, is_conditional):
-        args = ''
-        # args can be specified as no_log in several places: in the task or in
-        # the argument spec.  We can check whether the task is no_log but the
-        # argument spec can't be because that is only run on the target
-        # machine and we haven't run it thereyet at this time.
-        #
-        # So we give people a config option to affect display of the args so
-        # that they can secure this if they feel that their stdout is insecure
-        # (shoulder surfing, logging stdout straight to a file, etc).
-        if not task.no_log and C.DISPLAY_ARGS_TO_STDOUT:
-            args = ', '.join(('%s=%s' % a for a in task.args.items()))
-            args = ' %s' % args
-        self._display.banner("TASK [%s%s]" % (task.get_name().strip(), args))
-        if self._display.verbosity >= 2:
-            path = task.get_path()
-            if path:
-                self._display.display("task path: %s" % path, color=C.COLOR_DEBUG)
+        if self._display.verbosity > 1 or task.get_name() != 'include':
+            args = ''
+            # args can be specified as no_log in several places: in the task or in
+            # the argument spec.  We can check whether the task is no_log but the
+            # argument spec can't be because that is only run on the target
+            # machine and we haven't run it thereyet at this time.
+            #
+            # So we give people a config option to affect display of the args so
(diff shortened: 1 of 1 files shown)


lib/ansible/plugins/callback/default.py
```diff
@@ -113,11 +113,12 @@ def v2_playbook_on_no_hosts_remaining(self):
         self._display.banner("NO MORE HOSTS LEFT")
 
     def v2_playbook_on_task_start(self, task, is_conditional):
-        self._display.banner("TASK [%s]" % task.get_name().strip())
-        if self._display.verbosity > 2:
-            path = task.get_path()
-            if path:
-                self._display.display("task path: %s" % path, color=C.COLOR_DEBUG)
+        if self._display.verbosity > 1 and task.get_name() == 'include':
```
sivel: Won't this cause no task to ever be printed?  You have nested the 'TASK' banner display under a check that ensures that the task name is 'include' and that the verbosity needs to be greater than 1.

chogarcia:
great to see this

dagwieers:
@bcoca Rebased, can the flag be removed ?

bcoca:
sorry, we going to pass on this as the default, it is easy enough to substitute your own callback now that does this.

dagwieers:
I know this can be tweaked individually. The reason I proposed this pull request is because I think the default behaviour is not what most people expect, and rather confusing in specific use-cases (e.g. when using tags). So I propose we modify the default behaviour more to what it was like on Ansible v1.

Or should this be done using the new proposal process instead ?

bcoca:
includes could not be skipped in v1, they were preprocess macros, now being dynamic they are handled more as normal tasks, I believe this is a sane default.

dagwieers:
Right, but an include is not your normal task, it is not actually doing anything except including other tasks. So both the fact that it is now displayed, and in the same color as skipped is very confusing.

Not just that, I'll show you our use-case which makes it hard to impossible to get the desired output in between all the useless include statements when using tags. If we only run our playbook for a single component (role) out of 55+ components, we get hundreds of useless include task statements (and somewhere the real tasks are hidden) whereas with v1.9.4 we only see the tasks and plays that were relevant.

That's why I strongly believe that showing includes when using Ansible (when there is no issue at hand) makes absolutely no sense, whereas showing them when increasing verbosity makes sense (when debugging, if you somehow expect something to be included, but it isn't)

lberruti:
I agree with @dagwieers 
+1

dagwieers:
@lberruti @cho-is You can now properly _like_ a pull-request.
No more need to comment +1, just do a :+1: on the PR.

dagwieers:
@lberruti @cho-is Even though I do believe this is the wrong default behaviour (it makes Ansible less useful as a tool) I switched to my own stdout callback plugin named "dense", which you can find here at #14978 

Feedback welcome.

drzraf:
> sorry, we going to pass on this as the default, it is easy enough to substitute your own callback now that does this.

could you please indicate where this is documented so that I can figure out what means "easy enough".

thank you

[edit] I just read #14978 this is clearly not what I would call an "easy enough" way to reduce output garbage... And this is exactly what --verbose is for.

marcosdiez:
Hi @drzraf 
Here is an example of a 50 line (with comments) callback plugin which overrides the default one 

https://github.com/marcosdiez/ansible_human_log.py/blob/master/default.py

I got the idea from https://github.com/ansible/ansible/pull/15012

PR #14305, opened 2016-02-04
Double import sys, removed one

dagwieers:
sys was imported twice.


abadger:
Thanks @dagwieers .  Merged to devel.

PR #14306, opened 2016-02-04
Double import tempfile, remove one

dagwieers:
tempfile was imported twice.


abadger:
Merged to devel.  Thanks @dagwieers !

Issue #12620, opened 2015-10-03
Can't use variables with '{{' or '}}'


frodopwns:
##### ISSUE TYPE
Bug Report

##### COMPONENT NAME
core

##### ANSIBLE VERSION
```
1.9.3 ( configured module search path = None)
Ansible config I dont have a /etc/ansible/ansible.cfg which I assume means Im running defaults?
Env OSX Yosemite
```

##### SUMMARY

I have some randomly generated values that i have to drop into config files.
for example:
salt_value: "!@sdfdf4325gfdGHh3s{{ ewf2/f w ss498"
This type of value causes a warning at the start of the run:
```
 [WARNING]: non fatal error while trying to template play variables: Failed to
template !@sdfdf4325gfdGHh3s{{ ewf2/f w ss498: template error while templating
string: expected token 'end of print statement', got 'w'
```
And an error when attempting to use the variable:
```
TASK: [manage_site | debug msg="{{ salt_value }}"] ***
fatal: [web1] => Failed to template msg="{{ salt_value }}": Failed to template !@sdfdf4325gfdGHh3s{{ ewf2/f w ss498: template error while templating string: expected token 'end of print statement', got 'w'
FATAL: all hosts have already failed -- aborting
```
Some suggestions given to me along with their results:
salt_value: "!@sdfdf4325gfdGHh3s{% raw %}{{{% endraw %} ewf2/f w ss498"
```
TASK: [manage_site | debug msg="{{ salt_value }}"] ***
fatal: [web1] => error parsing argument string 'msg="!@sdfdf4325gfdGHh3s{{ ewf2/f w ss498"', try quoting the entire line.
FATAL: all hosts have already failed -- aborting
```
salt_value: "!@sdfdf4325gfdGHh3s{{ '{{' }} ewf2/f w ss498"
```
TASK: [manage_site | debug msg="{{ salt_value }}"] ***
fatal: [web1] => error parsing argument string 'msg="!@sdfdf4325gfdGHh3s{{ ewf2/f w ss498"', try quoting the entire line.
FATAL: all hosts have already failed -- aborting
```
Is this possible?

dagwieers:
I can reproduce this in Ansible v2.0.0.2 using the following playbook:

``` yaml
- hosts: localhost
  gather_facts: no
  vars:
    salt_value: '!@sdfdf4325gfdGHh3s{{ ewf2/f w ss498'
  tasks:
    - debug: var=salt_value                                                                                                                                 
[...]

PR #14311, opened 2016-02-04
Add new 'unsafe' YAML constructor

sivel:
This new YAML tag/constructor allows a user to mark a scalar variable value as unsafe, which prevents templating, among other things.

See https://github.com/ansible/ansible/issues/12620 for a potential use case.

Documentation is provided alongside this new functionality.

### DIFF
--- a/docsite/rst/playbooks_advanced_syntax.rst
+++ b/docsite/rst/playbooks_advanced_syntax.rst
@@ -1,0 +1,59 @@
+Advanced Syntax
+===============
+
+.. contents:: Topics
+
+This page describes advanced YAML syntax that enables you to have more control over the data placed in YAML files used by Ansible.
+
+.. _yaml_tags_and_python_types
+
+YAML tags and Python types
+``````````````````````````
+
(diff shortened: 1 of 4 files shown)


jhawkesworth:
Great idea to add this.

Just wondering if there is a better name for it than !unsafe

Probably bringing some programmer baggage here but I tend to read ! at the start of a token as meaning 'not' so !unsafe reads like 'not unsafe' and now I am guessing what that means.

I guess the ! Is kind of just intended as a warning that something is unusual here?

So I was wondering if a better name for this might be 'literal'? It avoids using the negative 'un' prefix, but still communicates how Ansible should treat the data that follow. Whether or not it needs the ! At the front I don't know but I like it when playbooks are readable without looking like a programming language.

sivel:
The PyYAML implementation requires the use of the `!` to identify YAML tags.  Built in tags use `!!` but user added tags use `!`, that is something that we don't have control over.

jhawkesworth:
Fair enough, no sense in redefining YAML tags.
I still think the word unsafe doesn't help the reader understand what Ansible is going to do with the data that follows.

sivel:
I had also debated "aliasing" `unsafe` to `raw`, but although `raw` may be more understandable, it is also not precise in what it does.  I'm hoping that my documentation is clear enough, to make things understandable.

Issue #14310, opened 2016-02-04
[2.x] connection vars are None/null


resmo:
##### Issue Type:
- Bug Report (regression)
##### Ansible Version:

2.x
##### Summary:

since 2.x all connection vars are defaulted to `None/null`, as in <=1.9.x they were undefined. This should be either documented in porting guide or "fixed" so None is handled as undefined.
##### Steps To Reproduce:

In 1.9 this is `22`, in 2.0 it is `""`

```
- debug msg={{ ansible_ssh_hosts | default(22) }}
```

Alternatively, to get the same result, the porting guide can be extended with:

```
- debug msg={{ ansible_ssh_hosts or 22 }}
```

sivel:
I have a proof of concept fix for potentially handling this at https://github.com/ansible/ansible/pull/14315

jimi-c:
# Closing This Ticket

Hi!

We believe the above commit should resolve this problem for you. This will also be included in the next release.

If you continue seeing any problems related to this issue, or if you have any further questions, please let us know by stopping by one of the two mailing lists, as appropriate:
- https://groups.google.com/forum/#!forum/ansible-project - for user questions, tips, and tricks
- https://groups.google.com/forum/#!forum/ansible-devel - for strategy, future planning, and questions about writing code

Because this project is very active, we're unlikely to see comments made on closed tickets, but the mailing list is a great way to ask questions, or post if you don't think this particular issue is resolved.

Thank you!

PR #14315, opened 2016-02-04
Revert connection vars to being undefined

sivel:
This is not necessarily ready to ship.  This is kind of a proof of concept.  There may be a different way we want to handle this.

See #14310 

Before v2, vars such as `ansible_ssh_port` were undefined, allowing the `default` filter to work on them without issue.  In v2, these values are set to `None` when not defined, preventing the basic usage of default.

This approach creates an "UNDEFINED" constant, that is then used to determine if we should define the connection var or not.

### DIFF
--- a/lib/ansible/playbook/base.py
+++ b/lib/ansible/playbook/base.py
@@ -40,13 +40,17 @@ from ansible.utils.unicode import to_unicode
 
 BASE_ATTRIBUTES = {}
 
+class Undefined:
+    def __len__(self):
+        return 0
+UNDEFINED = Undefined()
 
 class Base:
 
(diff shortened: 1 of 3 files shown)


lib/ansible/playbook/base.py
```diff
@@ -40,13 +40,17 @@
 
 BASE_ATTRIBUTES = {}
 
+class Undefined:
+    def __len__(self):
```
bcoca, line 44: so currently we have many `is None` comparisons, should this inherit from NoneType?

```
from types import NoneType

class Undefined(NoneType):
```

ignore, that does not work

PR #14319, opened 2016-02-05
moved to base class for shell plugins

bcoca:
fish now sets env vars correctly
fish checksum now works


ksmets:
Please change `pipes.quotes` to `pipes.quote`.

bcoca:
@ksmets good eye, fixed!

PR #14332, opened 2016-02-05
Introduce new 'filetree' lookup plugin

dagwieers:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

v2.0
##### SUMMARY

The new "filetree" lookup plugin makes it possible to recurse over a tree of files within the task loop. This makes it possible to e.g. template a complete tree of files to a target system with little effort while retaining permissions and ownership.

The module supports directories, files and symlinks. And also SELinux.

If you provide more than one path, it will implement a `with_first_found` logic, and will not process entries it already processed in previous paths. This enables the user to merge different trees in order of importance, or add e.g. role_vars specific paths to influence different instances of the same role.

The item dictionary consists of:
- src
- root
- path
- mode
- state
- owner
- group
- seuser
- serole
- setype
- selevel
- uid
- gid
- size
- mtime
- ctime
##### EXAMPLES

Here is an example of how we use with_filetree within a role.
The `web/` path is relative to either `roles/<role>/files/` or `files/`.

``` yaml
 - name: Create directories
   file:
     path: /web/{{ item.path }}
     state: directory
     mode: '{{ item.mode }}'
   with_filetree: web/
   when: item.state == 'directory'

 - name: Template files
   template:
     src: '{{ item.src }}'
     dest: /web/{{ item.path }}
     mode: '{{ item.mode }}'
   with_filetree: web/
   when: item.state == 'file'

 - name: Recreate symlinks
   file:
     src: '{{ item.src }}'
     dest: /web/{{ item.path }}
     state: link
     force: yes
     mode: '{{ item.mode }}'
   with_filetree: web/
   when: item.state == 'link'
```
##### SPECIAL USE

The following properties also have its special use:
- root: Makes it possible to filter by original location
- path: Is the relative path to root
- uid, gid: Makes it possible to force-create by exact id, rather than by name
- size, mtime, ctime: Makes it possible to filter out files by size, mtime or ctime

### DIFF
--- a/lib/ansible/plugins/lookup/filetree.py
+++ b/lib/ansible/plugins/lookup/filetree.py
@@ -18,7 +18,6 @@ from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
 import os
-import glob
 import pwd
 import grp
 import stat
--- a/lib/ansible/plugins/lookup/filetree.py
+++ b/lib/ansible/plugins/lookup/filetree.py
@@ -24,6 +24,8 @@ import grp
 import stat
 
 from ansible.plugins.lookup import LookupBase
+from __main__ import display
+warning = display.warning
 
 HAVE_SELINUX=False
 try:
@@ -31,12 +33,6 @@ try:
     HAVE_SELINUX=True
 except ImportError:
     pass
-
-try:
-    from __main__.display import warning
-except ImportError:
-    from ansible.utils.display import Display
-    display = Display()
 
 def _to_filesystem_str(path):
     '''Returns filesystem path as a str, if it wasn't already.
@@ -66,12 +62,12 @@ def selinux_context(path):
     return context
 
 def file_props(root, path):
-    ''' Returns dictionary with file properties, or return None on failure'''
+    ''' Returns dictionary with file properties, or return None on failure '''
     abspath = os.path.join(root, path)
 
     try:
         st = os.lstat(abspath)
-    except OSError, e:
+    except OSError as e:
         warning('filetree: Error using stat() on path %s (%s)' % (abspath, e))
         return None
 
@@ -86,7 +82,7 @@ def file_props(root, path):
         ret['state'] = 'file'
         ret['src'] = abspath
     else:
-        warning('filetree: Error file type of %s not support' % abspath)
+        warning('filetree: Error file type of %s is not supported' % abspath)
         return None
 
     ret['uid'] = st.st_uid
--- a/lib/ansible/plugins/lookup/filetree.py
+++ b/lib/ansible/plugins/lookup/filetree.py
@@ -33,7 +33,7 @@ except ImportError:
     pass
 
 try:
-    from __main__ import display
+    from __main__.display import warning
 except ImportError:
     from ansible.utils.display import Display
     display = Display()
@@ -68,25 +68,25 @@ def selinux_context(path):
 def file_props(root, path):
     ''' Returns dictionary with file properties, or return None on failure'''
     abspath = os.path.join(root, path)
-    ret = dict(root=root, path=path)
-
-    try:
-        if os.path.islink(abspath):
-            ret['state'] = 'link'
-            ret['src'] = os.readlink(abspath)
-        elif os.path.isdir(abspath):
-            ret['state'] = 'directory'
-        elif os.path.isfile(abspath):
-            ret['state'] = 'file'
-            ret['src'] = abspath
-    except OSError as e:
-        display.warning('filetree: Error querying path %s  (%s)' % (abspath, e))
-        return None
 
     try:
         st = os.lstat(abspath)
-    except OSError as e:
-        display.warning('filetree: Error using stat() on path %s  (%s)' % (abspath, e))
+    except OSError, e:
+        warning('filetree: Error using stat() on path %s (%s)' % (abspath, e))
+        return None
+
+    ret = dict(root=root, path=path)
+
+    if stat.S_ISLNK(st.st_mode):
+        ret['state'] = 'link'
+        ret['src'] = os.readlink(abspath)
+    elif stat.S_ISDIR(st.st_mode):
+        ret['state'] = 'directory'
+    elif stat.S_ISREG(st.st_mode):
+        ret['state'] = 'file'
+        ret['src'] = abspath
+    else:
+        warning('filetree: Error file type of %s not support' % abspath)
         return None
 
     ret['uid'] = st.st_uid
--- a/lib/ansible/plugins/lookup/filetree.py
+++ b/lib/ansible/plugins/lookup/filetree.py
@@ -22,7 +22,6 @@ import glob
 import pwd
 import grp
 
-from ansible.errors import AnsibleError
 from ansible.plugins.lookup import LookupBase
 
 HAVE_SELINUX=False
@@ -79,13 +78,13 @@ def file_props(root, path):
         elif os.path.isfile(abspath):
             ret['state'] = 'file'
             ret['src'] = abspath
-    except OSError, e:
+    except OSError as e:
         display.warning('filetree: Error querying path %s  (%s)' % (abspath, e))
         return None
 
     try:
         stat = os.stat(abspath)
-    except OSError, e:
+    except OSError as e:
         display.warning('filetree: Error using stat() on path %s  (%s)' % (abspath, e))
         return None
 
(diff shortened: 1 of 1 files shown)


dagwieers:
I wonder if this is also eligible for the upcoming 1.9.5 release ?
As we are not allowed to move to 2.0 or 2.1 for a considerable time due to conflicting releases.

RussellLuo:
:+1:

lib/ansible/plugins/lookup/filetree.py
```diff
@@ -0,0 +1,122 @@
...
+
+    if HAVE_SELINUX and selinux.is_selinux_enabled() == 1:
+        context = selinux_context(abspath)
+        ret['seuser'] = context[0]
+        ret['serole'] = context[1]
+        ret['setype'] = context[2]
+        ret['selevel'] = context[3]
+
+    return ret
+
```
RussellLuo, line 110: Two blank lines will be better. See [PEP 8](https://www.python.org/dev/peps/pep-0008/#blank-lines)

lib/ansible/plugins/lookup/filetree.py
```diff
@@ -0,0 +1,122 @@
...
+        ret = []
+        for term in terms:
+            term_file = os.path.basename(term)
+            dwimmed_path = self._loader.path_dwim_relative(basedir, 'files', os.path.dirname(term))
+            path = os.path.join(dwimmed_path, term_file)
+            for root, dirs, files in os.walk(path, topdown=True):
+                for entry in dirs + files:
+                    relpath = os.path.relpath(os.path.join(root, entry), path)
+                    props = file_props(path, relpath)
+                    if props != None:
```
RussellLuo: I think `if props is not None` is more conventional :smile:

lib/ansible/plugins/lookup/filetree.py
dagwieers, line 110: Fixed.

lib/ansible/plugins/lookup/filetree.py
dagwieers: Fixed.

lib/ansible/plugins/lookup/filetree.py
```diff
@@ -0,0 +1,123 @@
...
+import os
+import pwd
+import grp
+import stat
+
+from ansible.plugins.lookup import LookupBase
+from __main__ import display
+warning = display.warning
+
+HAVE_SELINUX=False
```
RussellLuo, line 29: One space around an assignment operator will be better. Sorry for my OCD of PEP 8 ;-)

lib/ansible/plugins/lookup/filetree.py
```diff
@@ -0,0 +1,123 @@
...
+try:
+    import selinux
+    HAVE_SELINUX=True
```
RussellLuo, line 32: one space around...

lib/ansible/plugins/lookup/filetree.py
```diff
@@ -0,0 +1,123 @@
...
+except ImportError:
+    pass
+
```
RussellLuo, line 35: Two blank lines between two top-level functions, sorry for not explicitly pointing this out before.

lib/ansible/plugins/lookup/filetree.py
```diff
@@ -0,0 +1,123 @@
...
+    Used in selinux interactions because it cannot accept unicode
+    instances, and specifying complex args in a playbook leaves
+    you with unicode instances.  This method currently assumes
+    that your filesystem encoding is UTF-8.
+
+    '''
+    if isinstance(path, unicode):
+        path = path.encode("utf-8")
+    return path
+
```
RussellLuo, line 48: two blank lines...

lib/ansible/plugins/lookup/filetree.py
```diff
@@ -0,0 +1,123 @@
...
+# If selinux fails to find a default, return an array of None
```
RussellLuo, line 49: And move this comment line as the function docstring of `selinux_context` will looks better ;-)

lib/ansible/plugins/lookup/filetree.py
dagwieers, line 29: This part came directly from the Ansible source code. So if the aim is to make everything PEP8 compliant, I don't mind doing my part. But I don't see a reason for being holier than the pope (dutch expression).

lib/ansible/plugins/lookup/filetree.py
dagwieers: This comes directly from the Ansible source code. Again, if this is general advice, let's fix it in Ansible first then get back to me with your PEP8 OCD.

lib/ansible/plugins/lookup/filetree.py
RussellLuo, line 29: You are definitely right! For the sake of consistency, PEP8-incompliance is never a problem.

I won't give this advice if I were more familiar with the Ansible source code. Sorry for my self-approbation ;-(

lib/ansible/plugins/lookup/filetree.py
RussellLuo, line 49: I can't agree more! Please ignore these immature suggestions :smile:

dagwieers:
shipit

dagwieers:
Or rather :shipit:

dagwieers:
@jimi-c @bcoca How about merging this one for v2.1.0 ?

jimi-c:
@dagwieers sorry this didn't make the cut, adjusted the milestone so we can grab it for 2.2.

DanyC97:
@jimi-c 2.2 is out however i don't see it documented. I've checked the stable-2.2 branch and i can see the code so i assume the doc is a bit behind ?

austinbutler:
This seems to work well for templates, except the files retain the .j2 extension. Is that intended? If so, would it be possible to have an option to strip that?

For anyone that ends up here, this seems to work:
```yml
- name: templates to files folder
  template:
    src: '{{ item.src }}'
    dest: "{{ files_dir }}/{{ item.path|regex_replace('.j2','') }}"
    force: yes
  with_filetree: templates/
  when: item.state == 'file'
```

dagwieers:
You can do that using jinja2. There is no reason to add functionality in individual modules/plugins if you can do the same using jinja2 filters. 

- http://jinja.pocoo.org/docs/2.9/templates/#builtin-filters
- http://docs.ansible.com/ansible/playbooks_filters.html

So something like: `{{ item.path | splitext | first }}` would do.
Or `{{ item.path | replace('.j2', '') }}`.
Or even: `{{ item.path | regexreplace('\.j2$', '') }}`

I haven't tested these, it's a bit theoretical. Let me know if this works, would be nice to add this to the documentation at some point. Thanks !

austinbutler:
Sorry, edited my comment on you but `regex_replace('.j2','')` worked for me. Having it in the documentation would be great, thanks!

dagwieers:
@austinbutler I would still add the `$` in that regex, just in case.

austinbutler:
You're right, that should be safer. Added and tested again, all good.

splashx:
Is this on 2.3.0.0? I'm using devel, but don't seem to be able to do either of those: `lookup('filetree', 'dir/')` or `with_filetree`

dagwieers:
Documentation added here: #26157

JPvRiel:
> You can do that using jinja2. There is no reason to add functionality in individual modules/plugins if you can do the same using jinja2 filters.

Doesn't the `template` module strip `.j2` when used with a single template file? AFAIK, the current `template` module behaviour with a normal `.j2` extension source file is to strip the `.j2` extension for you. 

So, I disagree with the argument that jinja2 filters should be used because context matters. If it was the `copy` or `fetch` module, that extra regex filter makes sense. But with `template`, why must a user expect and code around different behaviour when `with_filetree` versus normal files? That seems inconsistent to me and I bet many other users of `with_filetree` would find in non-intuitive if the `.j2` extension doesn't get stripped.

dagwieers:
@JPvRiel The **filetree** plugin returns a dictionary with information for each file, that is it. How users use it, is up to the user. I don't see what the **filetree** plugin should do different here.

PS AFAICT the **template** module does not strip anything from the provided filenames.

JPvRiel:
> PS AFAICT the template module does not strip anything from the provided filenames.

Sorry, yip, my bad - I double checked now and indeed, it doesn't strip the `.j2` extension like I thought. Your original point about using jinja2 filters is valid.

PR #14334, opened 2016-02-05
add systemd-nspawn connection driver

larsks:
This commit adds a connection driver built on top of systemd-nspawn.
This is similar to the existing `chroot` driver, except that nspawn
offers a variety of additional services. For example, it takes care of
automatically mounting `/proc` and `/sys` inside the chroot environment,
which will make a variety of tools work correctly that would otherwise
fail.

You can take advantage of other system-nspawn features to perform more
complicated tasks.  For example, on my x86_64 system I have a Raspberry
Pi disk image mounted on `/rpi`.  I can't use `chroot` with this because
the binaries contained in the image are for the wrong architecture.
However, I can use the systemd-nspawn `--bind` option to automatically
insert the appropriate qemu-arm binary into the container using an
inventory file like this:

```
pi ansible_host=/rpi ansible_nspawn_extra_args='--bind /usr/bin/qemu-arm --bind /lib64'
```

See http://www.freedesktop.org/software/systemd/man/systemd-nspawn.html
for more information about systemd-nspawn itself.

### DIFF
--- a/lib/ansible/cli/__init__.py
+++ b/lib/ansible/cli/__init__.py
@@ -286,6 +286,8 @@ class CLI(object):
                 help="specify extra arguments to pass to scp only (e.g. -l)")
             connect_group.add_option('--ssh-extra-args', default='', dest='ssh_extra_args',
                 help="specify extra arguments to pass to ssh only (e.g. -R)")
+            connect_group.add_option('--nspawn-extra-args', default='', dest='nspawn_extra_args',
+                help="specify extra arguments to pass to systemd-nspawn only (e.g. --bind)")
 
             parser.add_option_group(connect_group)
 
--- a/lib/ansible/constants.py
+++ b/lib/ansible/constants.py
@@ -244,6 +244,7 @@ ANSIBLE_SSH_CONTROL_PATH       = get_config(p, 'ssh_connection', 'control_path',
 ANSIBLE_SSH_PIPELINING         = get_config(p, 'ssh_connection', 'pipelining', 'ANSIBLE_SSH_PIPELINING', False, boolean=True)
 ANSIBLE_SSH_RETRIES            = get_config(p, 'ssh_connection', 'retries', 'ANSIBLE_SSH_RETRIES', 0, integer=True)
 PARAMIKO_RECORD_HOST_KEYS      = get_config(p, 'paramiko_connection', 'record_host_keys', 'ANSIBLE_PARAMIKO_RECORD_HOST_KEYS', True, boolean=True)
+ANSIBLE_NSPAWN_ARGS            = get_config(p, 'nspawn_connection', 'nspawn_args', 'ANSIBLE_NSPAWN_ARGS', '-q')
 
 
 # obsolete -- will be formally removed
--- a/lib/ansible/playbook/play_context.py
+++ b/lib/ansible/playbook/play_context.py
@@ -49,35 +49,36 @@ except ImportError:
 # in variable names.
 
 MAGIC_VARIABLE_MAPPING = dict(
-   connection       = ('ansible_connection',),
-   remote_addr      = ('ansible_ssh_host', 'ansible_host'),
-   remote_user      = ('ansible_ssh_user', 'ansible_user'),
-   port             = ('ansible_ssh_port', 'ansible_port'),
-   accelerate_port  = ('ansible_accelerate_port',),
-   password         = ('ansible_ssh_pass', 'ansible_password'),
-   private_key_file = ('ansible_ssh_private_key_file', 'ansible_private_key_file'),
-   pipelining       = ('ansible_ssh_pipelining', 'ansible_pipelining'),
-   shell            = ('ansible_shell_type',),
-   become           = ('ansible_become',),
-   become_method    = ('ansible_become_method',),
-   become_user      = ('ansible_become_user',),
-   become_pass      = ('ansible_become_password','ansible_become_pass'),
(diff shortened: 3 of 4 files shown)


larsks:
Made a short write-up of this driver [here](http://blog.oddbit.com/2016/02/08/a-systemd-nspawn-connection-driver-for-a/)

gdamjan:
Any future to this PR?

larsks:
@gdamjan don't know; there's been no interest from the ansible side...

bcoca:
There is a huge backlog, which means we don't always get to pay attention to every ticket. We have been reducing it but have not had time to look at this in depth.

lib/ansible/cli/__init__.py
```diff
@@ -286,6 +286,8 @@ def base_parser(usage="", output_opts=False, runas_opts=False, meta_opts=False,
                 help="specify extra arguments to pass to scp only (e.g. -l)")
             connect_group.add_option('--ssh-extra-args', default='', dest='ssh_extra_args',
                 help="specify extra arguments to pass to ssh only (e.g. -R)")
+            connect_group.add_option('--nspawn-extra-args', default='', dest='nspawn_extra_args',
+                help="specify extra arguments to pass to systemd-nspawn only (e.g. --bind)")
```
bcoca, line 290: i would prefer to not add more connection specific arguments

lib/ansible/playbook/play_context.py
```diff
@@ -49,35 +49,36 @@
...
+   become_method     = ('ansible_become_method',),
+   become_user       = ('ansible_become_user',),
+   become_pass       = ('ansible_become_password','ansible_become_pass'),
+   become_exe        = ('ansible_become_exe',),
+   become_flags      = ('ansible_become_flags',),
+   ssh_common_args   = ('ansible_ssh_common_args',),
+   sftp_extra_args   = ('ansible_sftp_extra_args',),
+   scp_extra_args    = ('ansible_scp_extra_args',),
+   ssh_extra_args    = ('ansible_ssh_extra_args',),
+   nspawn_extra_args = ('ansible_nspawn_extra_args',),
```
bcoca, line 71: this really does not need to be here, the connection plugin itself can look for it

lib/ansible/playbook/play_context.py
```diff
@@ -148,21 +149,23 @@ class PlayContext(Base):
...
-    _accelerate       = FieldAttribute(isa='bool', default=False)
-    _accelerate_ipv6  = FieldAttribute(isa='bool', default=False, always_post_validate=True)
-    _accelerate_port  = FieldAttribute(isa='int', default=C.ACCELERATE_PORT, always_post_validate=True)
+    _remote_addr       = FieldAttribute(isa='string')
+    _password          = FieldAttribute(isa='string')
+    _private_key_file  = FieldAttribute(isa='string', default=C.DEFAULT_PRIVATE_KEY_FILE)
+    _timeout           = FieldAttribute(isa='int', default=C.DEFAULT_TIMEOUT)
+    _shell             = FieldAttribute(isa='string')
+    _nspawn_args       = FieldAttribute(isa='string', default=C.ANSIBLE_NSPAWN_ARGS)
+    _nspawn_extra_args = FieldAttribute(isa='string')
```
bcoca, line 158: again, these items should be restricted to the connection plugin itself

gdamjan:
I guess any additional settings nspawn needs can be specified in the  environment? what about a section in `ansible.cfg`?

bcoca:
The section in ansible.cfg is fine, as well as inventory connection vars, they just need to be queried from the connection plugin (i believe winrm has good example of this).

lib/ansible/constants.py
```diff
@@ -244,6 +244,7 @@ def load_config_file():
 ANSIBLE_SSH_PIPELINING         = get_config(p, 'ssh_connection', 'pipelining', 'ANSIBLE_SSH_PIPELINING', False, boolean=True)
 ANSIBLE_SSH_RETRIES            = get_config(p, 'ssh_connection', 'retries', 'ANSIBLE_SSH_RETRIES', 0, integer=True)
 PARAMIKO_RECORD_HOST_KEYS      = get_config(p, 'paramiko_connection', 'record_host_keys', 'ANSIBLE_PARAMIKO_RECORD_HOST_KEYS', True, boolean=True)
+ANSIBLE_NSPAWN_ARGS            = get_config(p, 'nspawn_connection', 'nspawn_args', 'ANSIBLE_NSPAWN_ARGS', '-q')
```
bcoca, line 247: I would add (commented out) example in examples/ansible.cfg

bcoca:
so while adding a 'clean' version i realized systemd-nspawn is only useful to create a new instance and not a way to send commands to existing instances (machinectl might fit the bill but return codes seem a mess).

In it's current form this does not qualify as a connection plugin as only the 1st command sent can work, the rest will fail as the instance is 'already running', the first will also fail if that was the case.

gdamjan:
how do you mean it'll be running?

`systemd-nspawn -D /path/to/container command` will run the command and exit, nothing is left running. in the non-boot (-b) form nspawn is a glorified chroot.

bcoca:
@gdamjan which makes it impossible to actually modify a container when it has an a running service/command, which is expected to be possible via the connection plugins, also this limits access to a single instance, while no other connection does this.

gdamjan:
ok, I understand. My usecase was mainly about "containers" that are not running (yet) when configured by ansible, but I see how that's not complete.

Issue #14348, opened 2016-02-06
Self written modules in roles with become_user != root are broken


adp-knaumann:
**Issue Type:**
    Bug Report

**Ansible Version:**
kn@master:~$ ansible --version
ansible 2.0.0.2
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides

**Ansible Configuration:**
kn@master:~$ egrep -v '^$|^#' /etc/ansible/ansible.cfg
[defaults]
[privilege_escalation]
[paramiko_connection]
[ssh_connection]
[accelerate]
[selinux]

**Environment:**

Lubuntu 15.10
RHEL6.5

**Summary:**

Self written modules don't work in ansible 2.0.0.2 when the following conditions are true:
1. become: yes is used
2. become_user is not root
3. the playbook is targeting a remote host (don't use 127.0.0.1)

**Steps To Reproduce:**

Code: https://gist.github.com/adp-knaumann/8db12375329f670091d4

To recreate the issue:

kn@client:~$ sudo useradd -m testusr
kn@master:~$ ANSIBLE_KEEP_REMOTE_FILES=1 ansible-playbook -i hosts module_bug.yml

**Expected Results:**

Result with become user root is flawless (because args can be read in any case of course):

kn@master:~$ ANSIBLE_KEEP_REMOTE_FILES=1 ansible-playbook -i hosts --ask-pass --ask-become-pass module_bug.yml 
PLAY ***************************************************************************

TASK [setup] *******************************************************************
ok: [client]

TASK [module_bug : run_module] *************************************************
ok: [client]

TASK [module_bug : print] ******************************************************
ok: [client] => {
    "param1": "foobarbaz"
}

PLAY RECAP *********************************************************************
client                      : ok=3    changed=0    unreachable=0    failed=0   

**Actual Results:**
PLAY ***************************************************************************

TASK [setup] *******************************************************************
ok: [client]

TASK [module_bug : run_module] *************************************************
[...]

PR #14351, opened 2016-02-06
Fix for issue 14348

adp-knaumann:
Self written modules in roles with become_user != root are broken

Fix: Also make the args file world readable.


jimi-c:
I believe there should be an `if args_file_path:` above that, as it may be `None`.

jimi-c:
Even better, if we could chmod both of these in one shot to save the round trip time.

jimi-c:
Squashed the above commits into one and merged:

<pre>
commit 1ea21f1f8ce3b9fa74a8803c4b87af50636445c6
Author: James Cammarata <jimi@sngx.net>
Date:   Mon Feb 8 10:00:26 2016 -0500
    Also chmod the args file path when using become for old-style modules
    
    Fixes #14348
</pre>

Also cherry-picked the above to `stable-2.0`.

Thanks!

PR #14353, opened 2016-02-06
Include's tags argument allows only one tag to be specified.

felixfontein:
I have an Ansible 1.9 playbook which uses an `include` role with multiple tags:

```
tasks:
 - include: a.yml tags=b,c
```

When I try to run task `b` with Ansible 2.0 (`ansible-playbook foo.yml -t b`), no task is executed.
The problem seems to be that in `lib/ansible/plugins/strategy/__init__.py`, the `tags` argument is not split by `,` when it is a string, but simply put into a list. This PR fixes this so that `ansible-playbook foo.yml -t b` will continue to work as with Ansible 1.9.


jimi-c:
Merged, thanks!

jimi-c:
Also cherry-picked to stable-2.0

PR #14354, opened 2016-02-07
Pass boto_profile to get_route53_records

Jonnymcc:
If a boto profile is supplied on the command line it will not be used for the new route 53 connection object. This ensures that the new connection uses the profile that was specified.

### DIFF
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -1243,7 +1243,7 @@ class Ec2Inventory(object):
         ''' Get and store the map of resource records to domain names that
         point to them. '''
 
-        r53_conn = route53.Route53Connection()
+        r53_conn = route53.Route53Connection(profile_name=self.boto_profile)
         all_zones = r53_conn.get_zones()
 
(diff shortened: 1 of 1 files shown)


contrib/inventory/ec2.py
```diff
@@ -1243,7 +1243,7 @@ def get_route53_records(self):
         ''' Get and store the map of resource records to domain names that
         point to them. '''
 
-        r53_conn = route53.Route53Connection()
+        r53_conn = route53.Route53Connection(profile_name=self.boto_profile)
```
ryansb: Can you add a conditional so this only gets passed if the boto_profile is set?

ryansb (changes requested):

willthames:
shipit - I wish `boto3_conn` (or `connect_to_aws`) worked for route53 (I believe it's one of the few that doesn't, because both rely on the underlying service having a `connect_to_region` type call, and route53 is region agnostic).

ryansb (approved):

Issue #130, opened 2012-04-11
Allow hosts to access variables from other nodes


mpdehaan:
1. Implement the fetch module which stores facts per host in a specific directory
2. After setup runs, the host variables from each host are stored on the filesystem in a known directory structure, probably ~/.ansible_setup_variables/host.json or similar
3.   Add a custom Jinja2 extension is made that allows access to something like getvar(hostname, key)  and make sure our template module uses it
4.  We shift the template module to happen entirely server side (for evaluation), eliminating the need to execute templates on the host, and on the host (via Runner code) we only need to use the COPY module (not template) to move the file.   Now we may have to continue to support the local template evaluation for regular Ansible, since we've documented that you're allowed to call setup explicitly, so this is probably something only for playbooks, and we make the equivalent of template_local to do this.

RESULT -- we can do cool things like allow the webservers to know the ip address of the monitoring server and auto configure, etc.

Should be pretty simple following the above steps.

mpdehaan:
I've implemented 2 without needing to rely on the fetch module, so it's faster.

Files are saved by default in /var/lib/ansible/setup_data/hostname/setup or if run as non-root ~./ansible/setup_data/hostname/setup

We will be using this file with the template module in 4.   Runner will respond to "is_playbook=True" by running local templates versus remote templates, but this won't occur for non-playbooks.  When it uses local templates, it will create temporary files
and use the 'copy' module instead of the template module, then delete the temp file once transferred

Steps 3 and 4 are now unblocked and ready to proceed, and 4 has advantages (less-deps) even w/o 3.

mpdehaan:
#152 is opened as a prereq for the local templates

dagwieers:
Is there a caching mechanism in order to not recreate "dynamic" facts every time ? Some facts may not change without a reboot, or do not change very often, or will never change. In order to make execution faster we may want to cache facts for a certain time ? A caching mechanism would be useful both for use on the host as on nodes.

mpdehaan:
Dag -- I don't understand, are you talking about facter/ohai?  In which case they do their own thing... or the proposal to add our own?   Seems to not be about this ticket, apologies for confusion.

This is about storing the variables from the host when setup runs, so it will only happen once per node and then make those variables available to templates, so I think, basically, it does what you want... facts get evaluated once per playbook.  If you are using vars_files, actually twice.  But not three times.  Three is right out.

dagwieers:
[...]

PR #14359, opened 2016-02-07
implement folding of long lines in inventory

### LINKED ISSUE #14358: RFE: folding long lines in ansible inventory files
I would like the ability to fold long lines in ansible inventory files.  If I have an inventory entry that includes lots of per-host variables, it is hard to discuss this in email, or write about it in a blog post, or otherwise communicate with other people about it.  Something that permitted this would be lovely:

```
myhost ansible_host=a.b.c.d \
  ansible_ssh_extra_args="-F ~/.ssh/specialconfig" \
  systemowner=me@example.com \
  someotheroption=someothervalue
```

larsks:
This commit permits entries in an inventory file to be extended
over multiple lines by indenting subsequent lines, as in
https://docs.python.org/3/library/configparser.html#supported-ini-file-structure.

For example, this example from the docs:

```
other1.example.com  ansible_connection=ssh ansible_user=mpdehaan
```

Can be written as:

```
other1.example.com
  ansible_connection=ssh
  ansible_user=mpdehaan
```

This commit includes doc and test updates.

Resolves #14358.

### DIFF
--- a/docsite/rst/intro_inventory.rst
+++ b/docsite/rst/intro_inventory.rst
@@ -80,6 +80,23 @@ You can also select the connection type and user on a per host basis:
 As mentioned above, setting these in the inventory file is only a shorthand, and we'll discuss how to store them in individual files
 in the 'host_vars' directory a bit later on.
 
+Extending long lines
+~~~~~~~~~~~~~~~~~~~~
+
+Long lines can be extended using the INI-style convention of indenting
+subsequent lines, so the above could also be written like this::
+
+   [targets]
+
+   localhost
+     ansible_connection=local
+   other1.example.com
+     ansible_connection=ssh
+     ansible_user=mpdehaan
+   other2.example.com
+     ansible_connection=ssh
+     ansible_user=mdehaan
+
 .. _host_variables:
 
 Host Variables
--- a/lib/ansible/inventory/ini.py
+++ b/lib/ansible/inventory/ini.py
@@ -64,6 +64,41 @@ class InventoryParser(object):
     def _raise_error(self, message):
         raise AnsibleError("%s:%d: " % (self.filename, self.lineno) + message)
 
+    def _extendlines(self, lines):
+        '''Iterate over the physical lines in a file and return
+        logical lines by permitting lines to be extended as in
+        https://docs.python.org/3/library/configparser.html#supported-ini-file-structure'''
+
+        acc = []
+
+        for line in lines:
+            self.lineno += 1
+
+            if self.patterns['extend'].match(line):
+                if not acc:
+                    self._raise_error('Found unexpected indent')
+
+                acc.append(line.strip())
+            else:
+                if acc:
+                    yield ' '.join(acc)
+                    acc = []
+
+                # skip comments and blank lines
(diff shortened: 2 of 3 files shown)


amenonsen:
I'm not very fond of the implementation, because it mixes knowledge of comments and section headers between two different functions. I think it would be clearer if the code did the line extension further down in `_parse`, after it's established that the current line is not a section header.

There's also the problem that it'll cause the wrong line number to be reported in error messages, but maybe that's not critical (or maybe I feel that way because I don't see a really easy way to deal with it).

I also think it's much better to have -escaped newlines, but I suppose that's a moot point if these indented continuation lines are already an INI-format standard.

bcoca:
I don't really see the need for this, since we can already have these vars in host/group vars files ( a lot more readable). 

I'm also worried that this adds extra issues when trying to escape values and deal with comments.

As for "INI standard" there really isn't one, but many competing unofficial ones. Even the current inventory format departs a lot from what many people and INI parsers expect.

I'm really tempted to just add back a new YAML inventory plugin ....

amenonsen:
> I don't really see the need for this, since we can already have these
> vars in host/group vars files ( a lot more readable). 

I think it's extremely useful. It's easier to send someone a single
inventory file with required variable settings than a full tree of
hosts and variables, easier to update if you're trying to debug a
problem, easier to review on a paste site, etc.

> I'm also worried that this adds extra issues when trying to escape
> values and deal with comments.

What extra issues specifically are you worried about?

> As for "INI standard" there really isn't one, but many competing
> unofficial ones. Even the current inventory format departs a lot from
> what many people and INI parsers expect.

Then I think we should just do something like this around line #130:

```
while line.endswith('\'):
    line += lines[self.lineno].strip()
    self.lineno += 1
```

(Note: only "something like", not exactly this. The backslashes would have to be removed, for example.)

> I'm really tempted to just add back a new YAML inventory plugin ....

"I don't really see the need for this, since we can already have these
vars in host/group vars files." ;-)

abadger:
I do not like the proposed syntax (indented lines).  I've been hoping that we'd get to a place where we could make use of an ini parsing library to modify inventory from an ansible module (my need is to better automate adding vms for testing).  indented lines signifying a continuation aren't a feature of any of the ini libraries that I know of.

If we like indented format then an inventory plugin using YAML does seem like the natural choice.

larsks:
>  indented lines signifying a continuation aren't a feature of any of the ini libraries that I know of.

@abadger  as I have pointed out above, this is precisely the syntax implemented by the Python standard library.  There's even a link to the documentation.

abadger:
You link to python 3 documentation. Does this work in python 2's config
parse?

larsks:
Replace "3" with "2" in that URL: https://docs.python.org/2/library/configparser.html

The first example shows the same syntax.

bcoca:
Sadly current ini file already departs from Config parser, [group:section] naming and the lack of the mandatory [defaults] section means this needs to be 're implemented' as this PR does

abadger:
Ah, I was looking for an example in the section at the bottom.

larsks:
I'm closing this old PR.

PR #14374, opened 2016-02-08
add ismount function in module_utils

### LINKED ISSUE #2186: Mount fails if ansible user has no read permission inside mount
##### Summary:

A bug in python os.path makes the mount module fail when
- state=mounted
- filesystem: nfs (probably most others as well)
- the mount is already present
- the ansible user has no read permission inside the mount

The reason for this is that ansible uses os.path.ismount(mountpath) which fails if pointed at a mount inside which the user has no read permission, see http://bugs.python.org/issue2466.
This is quite common if you use nfs with root_squash or kerberos.

The python-bug is quite old and unfixed, so maybe it's worthwhile to have a workaround in ansible. The function in question is 10 lines of python.
##### Steps To Reproduce:

See Summary
##### Expected Results:

Should return 'ok'
##### Actual Results:

returns failed

robinro:
- needed to fix https://github.com/ansible/ansible-modules-core/issues/2186
  via pull request https://github.com/ansible/ansible-modules-core/pull/2737

### DIFF
--- a/lib/ansible/module_utils/ismount.py
+++ b/lib/ansible/module_utils/ismount.py
@@ -1,0 +1,90 @@
+# This code is part of Ansible, but is an independent component.
+# This particular file snippet, and this file snippet only, is based on
+# Lib/posixpath.py of cpython
+# It is licensed under the PYTHON SOFTWARE FOUNDATION LICENSE VERSION 2
+#
+# 1. This LICENSE AGREEMENT is between the Python Software Foundation
+# ("PSF"), and the Individual or Organization ("Licensee") accessing and
+# otherwise using this software ("Python") in source or binary form and
+# its associated documentation.
+#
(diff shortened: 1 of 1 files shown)


robinro:
I'm not completely sure the inclusion of the python license is sufficient the way I did it.

abadger:
Thanks @robinro I've merged this so that we don't have this dependency blocking the fix to the mount module.  It is in the devel branch so it will be in 2.1.0 when we release that.

abadger:
I see that https://github.com/ansible/ansible-modules-core/pull/2737 is marked stable-2.0 (meaning that it's a bugfix that we'd like to see in the 2.0.x releases at some point) so I've cherry-picked this to the stable-2.0 branch as well (so it will appear in 2.0.1 or 2.0.2).

robinro:
thanks for the quick merge @abadger

PR #14390, opened 2016-02-09
strip BECOME-SUCCESS at lower level

d3matt:
script (and probably others) still has BECOME-SUCESS in stdout

### DIFF
--- a/lib/ansible/plugins/action/__init__.py
+++ b/lib/ansible/plugins/action/__init__.py
@@ -544,6 +544,8 @@ class ActionBase(with_metaclass(ABCMeta, object)):
         if rc is None:
             rc = 0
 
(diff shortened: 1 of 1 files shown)


jimi-c:
Hi @d3matt, I fixed this per the above using the method we have in other action plugins. Not sure if it makes sense moving it to the low-level method you show above but we can discuss that (the call to the cleaning method would have to be removed from other areas so we don't call it more than once).

Thanks!

d3matt:
yea, the reason I liked it in the _low_level_execute was because it cleaned up the stdout_lines array as well

d3matt:
I'll open a separate issue to track stdout_lines and you guys can make that call

d3matt:
https://github.com/ansible/ansible/issues/14395

PR #14400, opened 2016-02-09
paramiko transport appears to hang if it gets a sudo lecture

d3matt:
- bring paramiko transport closer to ssh transport in how it deals with prompt


deggis:
+1 for this. Spent some time finding the same thing.

alikins:
Looks reasonable to me.

@jimi-c Looks like this was intended to get merge. Did it get missed?

bcoca:
@amenonsen can you give sanity check?

jimi-c:
Merged, looks sane to me (though it does not fully implement the state machine that ssh.py does).

Thanks!

PR #14402, opened 2016-02-10
Fix hiding of original value of files in diff output with no_log

abadger:
Previous fix only hid final value of file.  This hides previous value as well.


nitzmahone:
:+1:

abadger:
Thanks @nitzmahone !

PR #14417, opened 2016-02-10
Module params should default to str in most cases.

### LINKED ISSUE #2867: mysql_user passwords not quotet correctly
Ansible version: 2.0 and devel

Ansible Configuration: none

Environment: N/A

Summary:
When a password is defined as var and it is an integer the password is not quoted when set/update a user password in mysql

```
- hosts: all
  vars:
        # This works not:
        mysql_root_password: 123456
        # This works:
        #mysql_root_password: "123456"
  tasks:
        - name: Update MySQL root password
          mysql_user:
           name: root
           host: localhost
           password: "{{ mysql_root_password }}"
           check_implicit_admin: yes
```

This code results in the following sql query:

```
SET PASSWORD FOR 'root'@'localhost' = PASSWORD(123456)
```

abadger:
Fixes https://github.com/ansible/ansible-modules-core/issues/2867

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -226,7 +226,7 @@ except ImportError:
 
 FILE_COMMON_ARGUMENTS=dict(
     src = dict(),
-    mode = dict(),
+    mode = dict(type='raw'),
     owner = dict(),
     group = dict(),
     seuser = dict(),
@@ -574,6 +574,7 @@ class AnsibleModule(object):
                 'int': self._check_type_int,
                 'float': self._check_type_float,
                 'path': self._check_type_path,
+                'raw': self._check_type_raw,
             }
         if not bypass_checks:
             self._check_required_arguments()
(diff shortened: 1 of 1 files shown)


bcoca:
+1

gimoh:
This breaks modules which accept complex types.  Example that reproduces the problem:

```
---
- hosts: localhost
  connection: local
  gather_facts: False
  tasks:
    - name: test user policy
      iam_policy:
        iam_name: test-user
        iam_type: user
        policy_name: test-user-pol
        policy_json:
          Version: '2012-10-17'
          Statement:
            - Action: ['s3:GetObject']
              Effect: Allow
              Resource: 'arn:aws:s3:::test-bucket/*'
        state: present
```

here policy_json parameter can be a string or a dict which is then converted into JSON.  With this change the argument is stringified so the module will not try to convert it, but at the same time it is not a valid JSON (it's python's repr, which uses single quotes which don't work in JSON) which causes the module to fail with a confusing message that there are "Syntax errors in policy".

The bug argues that unquoted integers are treated as integers, seems obvious really, I'd argue that treating integers as strings is more confusing and it should be required that such values are quoted explicitly.

PR #14428, opened 2016-02-10
Linode Dynamic Inventory: Add public IP's as hostvars

shawnsilva:
When the linode inventory is generated the linode label is used as the inventory host. If the label isn't a FQDN ansible can't connect.

This will set the hostvars for the ansible_ssh_host to the linodes public IP.


jimi-c:
Merged, thanks!

PR #14432, opened 2016-02-11
Cleanup for networking common bits.

chouseknecht:
Move Conditional class to netcfg. Added error handling for connect and execute methods.


jimi-c:
cc @privateip

PR #14435, opened 2016-02-11
inventory.refresh_inventory does not refresh inventory._restriction

h0nIg:
If we are updating the inventory with the meta action refresh_inventory, the internal list of the inventory will be refreshed and new objects will be inserted into the new list.
but since the following code does equal check via memory address of host1 == memory address of host2, further checks will fail and the host will not be returned. see L204

refresh should search for restrition hosts on the new list and replace the old restriction host list

### DIFF
--- a/lib/ansible/inventory/__init__.py
+++ b/lib/ansible/inventory/__init__.py
@@ -759,3 +759,12 @@ class Inventory(object):
         self.groups          = {}
 
         self.parse_inventory(self.host_list)
+
+        # refresh plain host list
+        new_hosts = []
+        for host in self._restriction:
+            name = host.get_name()
+            new_host = self.get_host(name)
+            new_hosts.append(new_host)
+
+        self._restriction = new_hosts


shawnsilva:
An issue I am still having with refresh_inventory, which I think is related, is that hostvars for inventory hosts from dynamic inventory aren't refreshed. In my particular case the dynamic inventory script relies on --host for the vars, and at doesn't return any _meta information directly.

brandond:
Confirmed that https://github.com/ansible/ansible/issues/15115 is not a dupe of this.

brandond:
I have created https://github.com/h0nIg/ansible/pull/1 to fix a related issue with refresh_inventory failing to refresh external inventory provided host vars. The code does not expect to have any limits or restrictions at this point, and is further hampered by a stale restrictions list (which is fixed up shortly afterwards by @h0nIg.

brandond:
@h0nIg - Just to be clear, the issue here is not that the code is comparing memory addresses. `ansible.inventory.Host` class overrides `__eq__` with a test that confirms instance type and compares object uuids. Since the host's uuid is randomly generated at object creation, a new Host with the same hostname, facts, etc is not recognized as being the same as the existing Host - so any references need to be either cleared or updated when the inventory is reloaded.

h0nIg:
@brandond correct, but its the same result: the host will not be equal after refreshing the inventory. Therefore the reference (self._restriction) is refreshed. Are there any other references that should be refreshed?

s-hertel:
Hi, @h0nIg, it looks like this was fixed in https://github.com/ansible/ansible/commit/bc81c76f8654b77fa18c79b71d6bc46919b9e04f by switching to use host names in the list of restrictions rather than using Host objects. (Please let me know if I'm mistaken, and check on latest devel for the current behavior).

PR #14445, opened 2016-02-11
Display the string which is causing an exception to be raised.

marcpo:
In the ansible template module, when there is an error while
expanding a templated string, displaying the string causing the
exception is very useful.

I had a missing closing } in a string from the group_vars/all file. The error only happened when I was trying to access a hostvars variable in a playbook. With this error message, I found the problem right away.


jimi-c:
Merged, thanks!

PR #14474, opened 2016-02-15
document exit status for ansible-playbook

Aga303:


Aga303:
Actually, are all of the exit codes are releavant to all of the ansible-commands (ansible, ansible-pull, ansible-galaxy, etc.)?

If yes, the EXIT STATUS section could just be put in it's own file and included to all of the man-pages.

bcoca:
not all of them are common to all utilities, but there is overlap, merging this now and making note to see the most efficient way to add this info to all.

PR #14476, opened 2016-02-15
Adds callback plugin for automatically joining/leaving Windows domains

jhawkesworth:
**This pull request has been superceeded by https://github.com/ansible/ansible/pull/14972**

**Comments in the conversation below explain why.**

_Original comment text was:_

See win_domain.rst for pre-requisites, setup and usage.

This module is wrapper around the kinit and kdestroy tools which must be available to ansible for this module to operate.

### DIFF
--- a/lib/ansible/plugins/callback/win_domain.py
+++ b/lib/ansible/plugins/callback/win_domain.py
@@ -1,0 +1,96 @@
+# This plugin for ansible is intended to manage acquiring and destroying
+# kerberos / windows domain credentials.
+
+import os
+import json
+import subprocess
+from subprocess import Popen, PIPE
+from ansible import errors
+
+from ansible.plugins.callback import CallbackBase
+
+class CallbackModule(CallbackBase):
+    """
+    This callback module is intended to manage acquiring and disposing
+    of kerberos / windows domain credentials.
+    A new credential cache is created for each run and the contents
+    of the cache are destroyed when the run completes.
+    It needs to be able to find kinit, klist and kdestroy
+    Because a ticket is good for all machines on a domain,
+    a ticket is cached for the first host found from inventory.
+    
+    """
+    CALLBACK_VERSION = 2.0
+    CALLBACK_TYPE= 'aggregate'
+    CALLBACK_NAME = 'win_domain'
+    CALLBACK_NEEDS_WHITELIST = True
+
+    def __init__(self):
+
+        super(CallbackModule, self).__init__()
+
+        self._display.vv('win_domain: plugin activated')
+        self.cache_count = 0
+
+    def __del__(self):
+
+        credential_cache = os.environ['KRB5CCNAME']
+        self._display.vv("win_domain: Leaving domains and removing all tickets from credential cache: %s" % (credential_cache))
+        self._display.vv('win_domain: plugin deactivated')
+        self.destroy_all_tickets(credential_cache)
+
+    def run_kerberos_command(self, command_and_args, password):
+
+        try:
+            cmd = subprocess.Popen(command_and_args, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            if password:
+               stdout, stderr = cmd.communicate('%s\n' % password)
+            else:
+               stdout, stderr = cmd.communicate()
+            if stderr:
+                stderr_lines = stderr.decode('ascii').splitlines()
+                if stderr_lines:
+                    self._display.vv( "win_domain: STDERR Error when running %s - message was: %s" % (command_and_args[0], stderr_lines) )
+            return cmd.returncode
+        except OSError:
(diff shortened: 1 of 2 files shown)


bcoca:
cc @cchurch

Plork:
keep in mind that in Ansible 2.0 ansible_ssh_user has become ansible_user.  line 95-97 in win_domain.py.

`if ('ansible_ssh_user' in host_grp_vars):
                    if ("@" in host_grp_vars['ansible_ssh_user']):
                        self._display.vv( "win_domain: Host %s is configured for windows domain connection.  Looking for ticket for %s ..." % (host, host_grp_vars['ansible_ssh_user']))
                        self.cache_new_ticket(host_grp_vars['ansible_ssh_user'], host_grp_vars['ansible_ssh_pass'])`

other then that Windows users need functionality like this. I don't feel like doing a "kinit" every time I run a playbook. ;)

also, how would this work via Tower? because you don't have a  ansible_user in `host.get_group_vars()` when using Tower. (or even have something in `host.get_group_vars()`) I would need to know the variable tower stores the username of given credential in.

jhawkesworth:
Thanks for the review @Plork 
I will fix up the ansible_ssh_user -> ansible_user (although perhaps I should allow both)?

I don't have a Tower environment right now but I would hope that by the time Tower actually runs a playbook, the group_vars will be the same as if the playbook was run on the command line.  If not.. well as you say I guess I'll need to look wherever else it is stored.  Do you have a Tower you could test in?

Plork:
I would support both since the variable is deprecated but not invalid yet.

I got it like this now ... since apparently phyton does not have a
switch/case function :P (I am sure there is a nicer solution to avoid
duplicated logic.)

`if ('ansible_ssh_user' in host_grp_vars):
                    if ("@" in host_grp_vars['ansible_ssh_user']):
                        self._display.vv( "win_domain: Host %s is configured for windows domain connection.  Looking for ticket for %s ..." % (host, host_grp_vars['ansible_ssh_user']))
                        self.cache_new_ticket(host_grp_vars['ansible_ssh_user'], host_grp_vars['ansible_ssh_pass'])
                elif ('ansible_user' in host_grp_vars):
                    if ("@" in host_grp_vars['ansible_user']):
                        self._display.vv( "win_domain: Host %s is configured for windows domain connection.  Looking for ticket for %s ..." % (host, host_grp_vars['ansible_user']))
                        self.cache_new_ticket(host_grp_vars['ansible_user'], host_grp_vars['ansible_pass'])`

I do have a tower ... and the only things that get passed to the group_vars
are the variables you set on the group in the "inventory" itself. (probably
with inheritance when a group is member of a group etc. But have not tested
this)

So what gets passed to the variable is this. I put the following variables on the group and added: self._display.vv(% (host_grp_vars) in the script.

---

ansible_connection: winrm
ansible_port: 5986
ansible_winrm_server_cert_validation: ignore

> win_domain: plugin activated win_domain: Checking for domains to join... 
> win_domain: Considering host: server.domain.com
> {u'ansible_port': 5986, u'ansible_winrm_server_cert_validation': u'ignore', u'ansible_connection': u'winrm'} 
> win_domain: Completed windows domain join checking.
> <server.domain.com> ESTABLISH WINRM CONNECTION FOR USER: user@DOMAIN.COM on PORT 5986 TO server.domain.com 
> <server.domain.com> WINRM CONNECT: transport=kerberos endpoint=https://server.domain.com:5986/wsman

So it looks like it gets the user from somewhere else.

In tower you add credentials to organizations which can be used by your
playbooks or vcs system etc.

nitzmahone:
@jhawkesworth What do you think about putting this in contrib? I think this is a decent stopgap, but it's not the way we want to support it going forward (we'll have integrated ticket management in the winrm connection plugin, probably for 2.2)...

jhawkesworth:
@nitzmahone - Hi sorry for delay replying.  I'm delighted that the plan is to have integrated ticket management in future, so contrib indeed probably the best place for this.  
I will close this and create a new PR for contrib (with ansible_user support too ) after I have finished figuring out what is going on in https://github.com/ansible/ansible-modules-core/issues/3032

nitzmahone:
@jhawkesworth - also, probably don't want to call it win_domain to avoid (real and Google) confusion with the forthcoming domain management modules. :) Maybe win_kerb_callback or something...

jhawkesworth:
@nitzmahone Fair point about avoiding confusion with domain management modules.  How about 'active_directory_callback'?  In my experience Kerberos isn't a term that many people who work at windows shops are familiar with..

-----Original Message-----

nitzmahone:
Sure, that'll work.

Plork:
Can't imagine anybody working with Windows and Active Directory that has
not heard of Kerberos tho and if so they might need to consider a career
switch :D

Op do 10 mrt. 2016 om 21:20 schreef Matt Davis notifications@github.com:

jhawkesworth:
Closing this PR as I have now raised https://github.com/ansible/ansible/pull/14972 to add renamed module to contrib

PR #14480, opened 2016-02-15
Update playbooks_lookups.rst

ikrauchanka:


bcoca:
the `---` is optional part of YAML spec, put merging this anyways

Issue #14329, opened 2016-02-05
ansible-playbook --extra-vars does not translate 'yes' or 'no' to boolean values


marcinhlybin:
##### Issue Type:

Bug Report
##### Ansible Version:

```
ansible 2.1.0 (devel 17ae01594a) last updated 2016/02/05 10:24:05 (GMT +200)
  lib/ansible/modules/core: (detached HEAD 88e0bfd75d) last updated 2015/11/20 12:34:36 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD 7da1f8d4ca) last updated 2015/11/20 12:34:44 (GMT +200)
  config file = [...]/ansible.cfg
  configured module search path = Default w/o overrides
```
##### Ansible Configuration:

Default
##### Environment:

Mac OS X, GNU/Linux
##### Summary:

ansible-playbook playbook.yml -e 'boolean_variable=yes' command does not translate 'yes' or 'no' to boolean value.
##### Steps To Reproduce:

playbook.yml:

```

---
- hosts: localhost
  vars:
    boolean_variable: no
  tasks:
    - debug: var=boolean_variable
```

Command:
`ansible-playbook playbook.yml -e 'boolean_variable=yes'`
##### Expected Results:

```
PLAY ***************************************************************************

TASK [setup] *******************************************************************
ok: [localhost]

TASK [debug] *******************************************************************
ok: [localhost] => {
    "boolean_variable": true
}

PLAY RECAP *********************************************************************
localhost
```
##### Actual Results:

```
PLAY ***************************************************************************

TASK [setup] *******************************************************************
ok: [localhost]

TASK [debug] *******************************************************************
ok: [localhost] => {
    "boolean_variable": "yes"
}

PLAY RECAP *********************************************************************
localhost
```

jimi-c:
Hi @ahes, when using -e and the key=value syntax, the value is always interpreted as a string. If you want more complex types you should use the JSON/YAML extended variable syntax (which can be sent in as a string or file): http://docs.ansible.com/ansible/playbooks_variables.html#passing-variables-on-the-command-line (note you'd still need to use the proper JSON/YAML boolean types there).

Alternatively you can use the `|bool` filter to convert things to booleans as you use them your playbooks.

If you have any further questions, please let us know by stopping by one of the two mailing lists, as appropriate:
- https://groups.google.com/forum/#!forum/ansible-project - for user questions, tips, and tricks
[...]

PR #14484, opened 2016-02-15
Documenting that extra-vars 'key=value' is for strings

antillean:
As per jimi-c's comment [here](https://github.com/ansible/ansible/issues/14329), this documents that values passed in using the extra-vars `key=value` syntax are interpreted as strings.


jpic:
Didn't know that ! LGTM

amenonsen:
Merged, thanks.

Issue #9687, opened 2014-12-01
ansible-galaxy should not connect to API server unless its needed


hani:
as title, currently trying to init or install (from scm) a role without external connectivity results in an error that the API server cannot be contacted.

willthames:
Use with `--offline`.

hani:
To clarify, while init does work with --offline, install does not. So I cannot for example use ansible-galaxy install --offline <internal scm repo>

willthames:
I know ansible-galaxy install works fine offline, because all the modifications I made to it were at a company where http to external urls were avoided where possible. 

`ansible-galaxy install` just works - it only contacts the galaxy api if you try and install a role from galaxy itself - if you point it an scm repo (so `git+http...` or `hg+http`) or a local tar file, galaxy won't be contacted. See https://github.com/ansible/ansible/blob/devel/bin/ansible-galaxy#L798

`ansible-galaxy init` does need `--offline` although it'd be nicer if the default template could just be handled without the API calls.

voidengineer:
Are there any news on this? This is a real blocker for updating to ansible 2.x.

syndbg:
Guess, how about it mentions in the error message that you can use `--offline`? 

E.g

```
~/work/devops/deployutils   ansible_deployutils ●  ansible-galaxy init -p roles common
ERROR! The API server (https://galaxy.ansible.com/api/) is not responding.
Please try again later or run with --offline.
```

tknerr:
Same here, the `--offline` switch is not supported in Ansible 2.0.0.2:

```
vagrant@vagrant:/vagrant$ ansible-galaxy install -r requirements.yml -vvv --offline
Usage: ansible-galaxy install [options] [-r FILE | role_name(s)[,version] | scm+role_repo_url[,version] | tar_file(s)]

ansible-galaxy: error: no such option: --offline
```

In my `requirements.yml` there is only an internal repo referenced:

``` yaml
- name: foo-roles
  src: ssh://git@stash.company.com:7999/ansible-roles/foo.git
  version: master
  path: roles
```

Same if I use `src: git+ssh://git@...` as the scm protocol

compuguy:
@tknerr 

This bug is fixed in ansible 2.1.0.0 (and I can confirm this as well).

bcoca:
closing as this was implemented in 2.1

tknerr:
@compuguy true, can confirm this is fixed in 2.1.0.0

Thanks!

[...]

Issue #13991, opened 2016-01-19
Regression: ansible requires access to api server for local dependencies


compuguy:
##### Issue Type:

Bug Report - Regression
##### Ansible Version:

2.0.0.2 - from PPA
##### Ansible Configuration:

Other than upgrading to 2.0.0.2, no config change
##### Environment:

Ubuntu 15.10
##### Summary:

Ansible 2.0.0.2 via the ansible-galaxy command, will not pull from a local github enerprise dependencies, without an internet connection to https://galaxy.ansible.com/api/. Ansible version 1.9.4 had no issues pulling down local dependencies from github enterprise.
##### Steps To Reproduce:

Be on a network without internet access, but has hosted git repos on it (github-enterprise, etc). The playbook is not publicly hosted.

Run dependency script: command used: ansible galaxy install -r requirements.yml
##### Expected Results:

Command runs successfully, standard output
##### Actual Results:

ERROR! The API server (https://galaxy.ansible.com/api/) is not responding, please try again later.

compuguy:
Note as of Ansible 2.0.1.0 this is still broken. Since my workplace's network doesn't have internet, this is preventing any of our github-enterprise hosted roles from updating, restricting us to ansible versions pre 2.x.x.x.

compuguy:
After doing some searching through the issues, it looks like there is a pull request that will add an offline option for galaxy: #14486 

Without this merge, I and anyone using ansible and ansible-galaxy with an internet limited network are limitied to using pre 2.x versions of ansible.

timss:
Thanks for sharing. I guess [patching](https://patch-diff.githubusercontent.com/raw/ansible/ansible/pull/14486.patch) that PR is the way to go while we wait.

willthames:
Proposed fix in #15593

PR #14486, opened 2016-02-15
Add --offline option to galaxy install

realloc:
I have part of infrastructure without internet access. This option helps to install roles form local mirror using the ansible-galaxy tool. 
The problem is described and discussed in issue #9687

### DIFF
--- a/docs/man/man1/ansible-galaxy.1.asciidoc.in
+++ b/docs/man/man1/ansible-galaxy.1.asciidoc.in
@@ -79,6 +79,10 @@ configured in your *ansible.cfg* file (/etc/ansible/roles if not configured)
 A file containing a list of roles to be imported, as specified above. This
 option cannot be used if a rolename or .tar.gz have been specified.
 
+*--offline*::
+
+Don't query the galaxy API unless it's needed when installing roles
+
 REMOVE
 ------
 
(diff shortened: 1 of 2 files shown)


Lujeni:
:+1: 

@realloc maybe add a space after each comma of the action list ( :heart: pep8)

realloc:
@Lujeni, I tried to keep the style of original code. I think restyling should be a separate change.

kbenyous:
:+1: 
Same problem, using 

```
ansible-galaxy install -r requirements.yml 
```

without internet access, but only a local role repository.

I applied your patch and now it works file :-)

philipkirkland:
Any idea when this pull request can be merged?

tima:
I did some checking on this particular issue and #13991. Reaching out to the galaxy server when it's not needed (you're grabbing a tar ball or installing from an internal repo) is a regression. The `--offline` option was removed intentionally though. 

What that option did is not really "offline" -- it could still go out and fetch roles from other online sources. It only switched off checking galaxy. 

More importantly, the option was removed because the galaxy code was (thought to have been) written to not call out to galaxy unless it really absolutely needed to -- you are asking for a role that wasn't from a SCM or a tar.

The problem that needs to be worked out here is not adding the switch back in, but correctly the logic so the call to Galaxy doesn't get made unless it absolutely has to.

compuguy:
@tima agreed. Until either an offline flag or the ansible-galaxy logic is fixed, its a blocker for anyone working on a restricted network and ansible 2.x. The problem is that nothing has been done or discussed in 2+ months.

tima:
Understood. A patch to fix the ansible-galaxy logic would be welcome.

willthames:
Proposed fix in #15593

Issue #13179, opened 2015-11-16
ansible_shell_type and make_become_cmd are at odds


amenonsen:
##### ISSUE TYPE
Bug Report

##### COMPONENT NAME
core

##### ANSIBLE VERSION
2.2

##### SUMMARY
While investigating #12290, I found that if you set `ansible_shell_type=csh`, then we construct a csh-syntax command line (e.g. using "`env A=b python…`" and "`>& /dev/null`"), which works if you're not using privilege escalation. But `make_become_cmd` doesn't know about `ansible_shell_type`, so it ends up issuing `sh -c '<csh-syntax-command>'`, which results in "bad fd" errors from bash (which doesn't accept `>& /dev/null`).

amenonsen:
Here are some possible solutions:
1. Deprecate `ansible_shell_type` and ignore what it's set to; always generate Bourne-syntax commands and wrap everything in `sh -c '…'`. This goes a little further than what we do now in `make_become_cmd`, and will always work, but quoting in the generated commands will be completely hellish.
2. Assume that `ansible_shell_type=csh` applies not only to `remote_user` but also to any `become_user` and teach `make_become_cmd` to generate csh-syntax commands in that case. This would simplify quoting, but the approach is a bit dubious because on FreeBSD my user had `/bin/sh` as the default shell, but root had `/bin/csh` by default. But changing the user shell to csh would presumably not be much of a problem.
3. As @willthames suggested, say "csh is not supported by Ansible" and require both `remote_user` and any `become_user` to have a Bourne-style login shell. This would simplify the code and remove most of the hideous quoting. Alas, since some systems still install csh as the default root shell, I suspect this approach would make users unhappy.

amenonsen:
(1) alone won't work, because we still need to be able to `\`-escape embedded newlines if the `remote_user`'s login shell is csh. Thanks to @abadger for pointing out the `_SHELL_EMBEDDED_PY_EOL` override in `plugins/shell/csh.py`.

bcoca:
soo, ansible_shell_type is meant to be used with also setting executable=/bin/csh|/bin/tcsh etc or fish. I've been thinking about merging both settings but have not made it a priority as it already works as is w/o too much hassle.

We might want to at least give a warning when they don't match

amenonsen:
"Works as is w/o too much hassle" == "breaks if you set it to csh" (and use privilege escalation).

Here's what I propose, in some detail:
[...]

PR #14488, opened 2016-02-15
shell + become fixes

bcoca:
1 less level of shell + quoting
simplified become commands, less quote and shell

possible fix for #13179

### DIFF
--- a/lib/ansible/playbook/play_context.py
+++ b/lib/ansible/playbook/play_context.py
@@ -464,8 +464,7 @@ class PlayContext(Base):
                     return bool(SU_PROMPT_LOCALIZATIONS_RE.match(data))
                 prompt = detect_su_prompt
 
-                su_success_cmd =  '%s -c %s' % (executable, success_cmd) # this is here cause su too succeptible to overquoting
-                becomecmd = '%s %s %s -c %s' % (exe, flags, self.become_user, su_success_cmd) #works with sh
+                becomecmd = '%s %s %s -c %s' % (exe, flags, self.become_user, pipes.quote('%s -c %s' % (executable, success_cmd)))
 
             elif self.become_method == 'pbrun':
 
@@ -479,7 +478,7 @@ class PlayContext(Base):
 
             elif self.become_method == 'runas':
                 raise AnsibleError("'runas' is not yet implemented")
-                #TODO: figure out prompt
+                #FIXME: figure out prompt
                 # this is not for use with winrm plugin but if they ever get ssh native on windoez
                 becomecmd = '%s %s /user:%s "%s"' % (exe, flags, self.become_user, success_cmd)
 
@@ -494,6 +493,7 @@ class PlayContext(Base):
                 if self.become_user:
                     flags += ' -u %s ' % self.become_user
 
+                #FIXME: make shell independant
                 becomecmd = '%s %s echo %s && %s %s env ANSIBLE=true %s' % (exe, flags, success_key, exe, flags, cmd)
 
             else:
@@ -502,7 +502,7 @@ class PlayContext(Base):
             if self.become_pass:
                 self.prompt = prompt
             self.success_key = success_key
-            return ('%s -c %s' % (executable, pipes.quote(becomecmd)))
+            return becomecmd
 
         return cmd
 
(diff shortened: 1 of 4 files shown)


amenonsen:
Finally, this PR appears to get the quoting correct, by doing what PR #13176 did. But that PR wasn't merged because you had tried it and it [broke in some configurations](https://github.com/ansible/ansible/pull/13176#discussion_r44944565). I think it's probably appropriate to note in the commit message what configurations this is known to break, or otherwise somehow indicate the scope of the problem you're trying to solve, and the limitations of the proposed solution.

amenonsen:
Sorry, the second part above about "bad fd number" was me getting stuff backwards. I'm deleting it to prevent confusion.

amenonsen:
(In the above comment, I had said something about the "Bad fd number" problem, but it was confused and incorrect. Here's the correct version.)

So I'm executing a copy task against a FreeBSD host (with csh set as the user's and root's login shell, but with /bin/sh available.) First, I do it without setting `ansible_shell_type=csh` (which @bcoca says you should do only if /bin/csh is the only shell available). We execute this command:

```
26745 1455627996.68342: _low_level_execute_command(): executing: /bin/sh -c 'su  root -c '"'"'/bin/sh -c '"'"'"'"'"'"'"'"'echo BECOME-SUCCESS-uvvtpoltljclevfqqfvcvklncsvxwcla; rc=flag; [ -r /tmp ] || rc=2; [ -f /tmp ] || rc=1; [ -d /tmp ] && rc=3; /usr/local/bin/python2.7 -V 2>/dev/null || rc=4; [ x"$rc" != "xflag" ] && echo "${rc}  "/tmp && exit 0; (/usr/local/bin/python2.7 -c '"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'import hashlib; BLOCKSIZE = 65536; hasher = hashlib.sha1();
afile = open("'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'/tmp'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'", "rb")
buf = afile.read(BLOCKSIZE)
while len(buf) > 0:
    hasher.update(buf)
    buf = afile.read(BLOCKSIZE)
afile.close()
print(hasher.hexdigest())'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"' 2>/dev/null) || (/usr/local/bin/python2.7 -c '"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'import sha; BLOCKSIZE = 65536; hasher = sha.sha();
afile = open("'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'/tmp'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'", "rb")
buf = afile.read(BLOCKSIZE)
while len(buf) > 0:
    hasher.update(buf)
    buf = afile.read(BLOCKSIZE)
afile.close()
print(hasher.hexdigest())'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"' 2>/dev/null) || (echo '"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'0  '"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'/tmp)'"'"'"'"'"'"'"'"''"'"''
```

…and csh (the user's login shell) barfs on the embedded newlines:

```
 26745 1455627996.71510: _low_level_execute_command() done: rc=1, stdout=Unmatched '.
Badly placed ()'s.
Badly placed ()'s.
while: Expression Syntax.
Badly placed ()'s.
Badly placed ()'s.
Badly placed ()'s.
Badly placed ()'s.
Badly placed ()'s.
Badly placed ()'s.
while: Expression Syntax.
Badly placed ()'s.
Badly placed ()'s.
Badly placed ()'s.
Unmatched ".
, stderr=
```

But if I DO set `ansible_shell_type=csh`, then we pass a csh-style redirect to /bin/sh, which generates the "Bad fd number" error.

amenonsen:
@bcoca notes that #14490 deprecates the (horribly ugly) checksum not-quite-one-liner in favour of the stat module, and since the checksum command is the only place where we send embedded newlines anyway, it neatly sidesteps the problem. Sounds good to me.

PR #14490, opened 2016-02-15
Use stat module instead of checksum code

bcoca:
added new function for action plugins and ported copy module to it
this avoids the very fragile checksum code that is shell dependant.

### DIFF
--- a/lib/ansible/plugins/action/__init__.py
+++ b/lib/ansible/plugins/action/__init__.py
@@ -291,28 +291,54 @@ class ActionBase(with_metaclass(ABCMeta, object)):
         res = self._low_level_execute_command(cmd, sudoable=sudoable)
         return res
 
+    def _execute_remote_stat(self, path, all_vars, follow):
+        '''
+        Get information from remote file.
+        '''
+        module_args=dict(
+           path=path,
+           follow=follow,
+           get_md5=False,
+           get_checksum=True,
+           checksum_algo='sha1',
+        )
+        mystat = self._execute_module(module_name='stat', module_args=module_args, task_vars=all_vars)
+
+        if 'failed' in mystat and mystat['failed']:
+            raise AnsibleError('Failed to get information on remote file (%s): %s' % (path, mystat['msg']))
+
+        if not mystat['stat']['exists']:
+            # empty might be matched, 1 should never match, also backwards compatible
+            mystat['stat']['checksum'] = '1'
+
+        return mystat['stat']
+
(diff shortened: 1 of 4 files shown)


bcoca:
cc @abadger

lib/ansible/plugins/action/__init__.py
```diff
@@ -291,28 +291,49 @@ def _remote_chmod(self, mode, path, sudoable=False):
...
 
-        cmd = self._connection._shell.checksum(path, python_interp)
-        data = self._low_level_execute_command(cmd, sudoable=True)
+        return mystat['stat']
+
+    def _remote_checksum(self, path, all_vars):
+        '''
+        Takes a remote checksum and returns 1 if no file
+        Returns a number 0-4 for specific errors.
+        '''
```
abadger: Minor: documentation should be something like:

```
Takes a remote filename and returns 1 if no file.  Returns a number 0-4 for specific errors.  Returns the checksum (currently sha1) of the file otherwise.
```

abadger:
Minor note about docstring, otherwise looks good!  +1

abadger:
Note: unarchive.py and fetch.py still remain as unported to the new code.  Can commit those later if desired as the _remote_checksum() compat function means it's not necessary to port all at once.

lib/ansible/plugins/action/__init__.py
```diff
@@ -291,28 +291,54 @@ def _remote_chmod(self, mode, path, sudoable=False):
...
-                u"the list so it can be fixed\ncommand: %s\n----\noutput: %s\n----\n" % (to_unicode(cmd), data))
-            # this will signal that it changed and allow things to keep going
-            return "INVALIDCHECKSUM"
+                x = remote_stat['checksum'] # if 1, file is missing
+        except AnsibleError as e:
+            errormsg = to_bytes(e)
+            if errormsg.endswith('Permission denied'):
+                x = "2" # cannot read file
+            elif errormsg.endswith('MODULE FAILURE'):
+                x = "4" # python not found or module uncaught exception
```
abadger, line 338: Since we aren't sending the errormsg to an exception, we want to use to_unicode rather than to_bytes.  to_bytes won't cause any problems with the code here in python2 but it will fail in python3 (because the string literals will be of the text type and so it won't match with errormsg as bytes then.)

abadger:
One note about python3 compat.  Otherwise this looks good.  +1 with the simple fix applied.

PR #14499, opened 2016-02-16
Make --diff lines of context configurable

gechr:
...and reduce the default number of context lines to 3, as this is more consistent with [`diff`](http://man7.org/linux/man-pages/man1/diff.1.html)'s default (and subsequently what most people are accustomed to):

```
   -u -U NUM --unified[=NUM]
   Output NUM (default 3) lines of unified context.
```

I've included this option under a new `[diff]` config section in anticipation of future configurable diff improvements (colours, output format, etc.) that I plan on contributing.

I wasn't sure whether it was worth making this a command line argument as well (e.g. `--diff-context=NUM`), but I think it may be worthwhile, given that the lines of context you'd want to see are largely dependent on the type of change you're making, which varies from deployment to deployment, but I'll leave it up to you guys to advise.


gechr:
@bcoca Possibly another simple one we can include along with the "Major Changes" to `diff` in 2.1.0?

bcoca:
i'll probably rewrite the entry in changelog once we get close to release, right now i think its a bit premature to add every thing we build on top of the new diff facility (maybe a blog post/guide once it is released?)

PR #14505, opened 2016-02-16
Update stated location of common_module.py

ahartz1:
As of Ansible 2.0, `common_module.py` is located within `lib/ansible/executor`; previously was in `lib/ansible`.


bcoca:
actually, we should point people to lib/ansible/module_utils/basic.py , module_common does the replacement that enables those functions but does not present them.

ahartz1:
I've changed reference to `lib/ansible/module_utils/basic.py` per your suggestion.

PR #14512, opened 2016-02-16
Fix leftover references to sudo instead of become

scottsb:
Catching a few places that earlier PR #10788 missed.


jimi-c:
Merged, thanks!

PR #14514, opened 2016-02-16
AWS Ranges lookup

Fluro:
This is a lookup that is used to find AWS CIDR ranges for services/regions based on what you pass to it as kwargs, and returns a list of the prefixes.


ryansb:
I've rebased and updated this PR as https://github.com/ansible/ansible/pull/32749

PR #14535, opened 2016-02-17
update uptime script to use version 2.0 of the api

b4ldr:
I thought i would have a crack at converting the uptime script to use version 2.0 of the API.  I mainly got this up and running with the help of information here https://groups.google.com/forum/#!topic/ansible-devel/Vuhg_rElDjk.  

Im not sure if this is the best way of doing things but thought it would be good to have a crack at a first pass


jimi-c:
Merged, thanks!

PR #14553, opened 2016-02-18
Clarify that the requirement is about the target host

hyperized:
In reference to https://github.com/ansible/ansible-modules-core/pull/3061


bcoca:
sorry, i think i beat you to it by a hair, solved via fe09f7ee49d01cf0258852fcd7d455c3a904f507

hyperized:
That was my initial fix, but it puts the comment in the H3, which seemed out of place .. it also seems my hair is about 2 hours long ;)

PR #14557, opened 2016-02-18
Add a configuration setting that allows the user to specify printing …

### LINKED ISSUE #14554: Allow to display play arguments during run
##### Expected Results:

It would be great to get the pre-2.0 behaviour with output like "TASK [service name=httpd state=started]".
This should probably be a config option because it was disabled due to security considerations (since the argument-spec no_log isn't available at this point).
##### Actual Results:

Output like "TASK [service]".

abadger:
…of task arguments in the header.

Fixes #14554

### DIFF
--- a/docsite/rst/faq.rst
+++ b/docsite/rst/faq.rst
@@ -304,8 +304,6 @@ How do I keep secret data in my playbook?
 
 If you would like to keep secret data in your Ansible content and still share it publicly or keep things in source control, see :doc:`playbooks_vault`.
 
-.. _i_dont_see_my_question:
-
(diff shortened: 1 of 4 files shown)


nitzmahone:
Looks good to me- probably the best compromise we can do without control-side no_log metadata...

bcoca:
+1

abadger:
@skinlayers thanks, looks like I didn't include that file in my commit.  Rectified here: https://github.com/ansible/ansible/commit/f9526b2ab22e2658d0f6ce5bb93a51cba9685851

PR #14559, opened 2016-02-18
Improve efficiency of merge_hash (Ansible v1.9)

dagwieers:
This commit improves 2 things:
- It makes merging empty dicts, or equal dicts faster
- It makes merging dicts faster (backported from v2.0)

I noticed that while debugging merge_hash a lot of merges related to empty dictionaries and sometimes identical dictionaries.


dagwieers:
@bcoca As discussed. I also backported the code from v2.0 which is more efficient.

PR #14560, opened 2016-02-18
Improve efficiency of merge_hash (Ansible v2.0)

dagwieers:
This is related to PR #14559, but only the part for Ansible v2.0

This commit makes merging empty dicts, or equal dicts more efficient.

I noticed that while debugging merge_hash a lot of merges related to empty dictionaries and sometimes also identical dictionaries.


dagwieers:
@bcoca As discussed on IRC.

PR #14562, opened 2016-02-18
Backport combine_vars() logic from Ansible v2.0

dagwieers:
While debugging I noticed that _validate_both_dicts() was evaluated twice through combine_vars (when merge_hash). Looking at v2.0 the logic was improved to not do _validate_both_dicts() twice in this case.

I also backported the 'update' behaviour as it looks more pythonic.


dagwieers:
For evaluation, here is a reference to the Ansible v2.0 code:
https://github.com/ansible/ansible/blob/devel/lib/ansible/utils/vars.py#L62

jimi-c:
Merged, thanks!

Issue #14324, opened 2016-02-05
Rescue not called when in the playbook there is only the include


dcodix:
I think it might be related to this other issue: https://github.com/ansible/ansible/issues/14040

The problem there described seems to be solved now.
Now I am doing:

main2.yml

```
- name: myname
  hosts: all
  gather_facts: True
  tasks:
    - include: tasks.yml
```

tasks.yml

```
- block:
  - include: file_to_fail.yml
  #- file: path=/tmp/file1 state=directory
  - file: path=/tmp/file3 state=directory
  rescue:
  - file: path=/tmp/file2 state=directory
  any_errors_fatal: yes
```

file_to_fail.yml
`- file: path=/tmp/file1 state=directory`

And happens the same as the other issue: If I make it fail the creation of "/tmp/file1" the rescue is not called, but only if it is inside the include, if I create it directly on the block (without the include) the rescue it is called.

I did not test it with the las code, but with the PR I made some days ago it was working:
https://github.com/ansible/ansible/pull/14213

Thanks!

dcodix:
I did some more tests arround this issue. And I found a small modification that corrects it, but doing some more testing I found out some other strange things.
With this modification I got the above working:
https://github.com/dcodix/ansible/commit/cca64d2ce5585b2e14a11c2a48f0a169b6aac4cd
But it is not good because If the main.xml is:

main.yml

```
- name: myname
  hosts: all
  gather_facts: True
  tasks:
    - include: tasks.yml
    - include: tasks2.yml
```

If something fails inside tasks.yml, tasks2.yml will be executed anyway.

Doing some more testing WITHOUT my modification I discovered another thing that may be related:

With:

main.yml

```
- name: myname
  hosts: all
  gather_facts: True
  tasks:
    - block:
      - include: tasks.yml
      - include: tasks2.yml
      rescue:
      - debug: msg="failed"
      any_errors_fatal: yes
```

tasks.yml

```
- block:
  - include: file_to_fail.yml
  - file: path=/tmp/file3 state=directory
  rescue:
  - file: path=/tmp/file2 state=directory
  any_errors_fatal: yes
```

file_to_fail.yml

```
- file: path=/tmp/file1 state=directory
```

tasks2.yml

```
- block:
  - file: path=/tmp/file10 state=directory
  rescue:
  - file: path=/tmp/file20 state=directory
  any_errors_fatal: yes
```

[...]

PR #14564, opened 2016-02-18
fix rescue not executing in includes

dcodix:
Rescue not executed in case in the main playbook there is only an include.
This seems to solve it.

Reference: #14324


jimi-c:
See my comment here: https://github.com/ansible/ansible/issues/14324#issuecomment-198759916

jimi-c:
Hi @dcodix, following up on old issues and I see I believe this issue was fixed, however you never responded nor gave an example to reproduce the issue you were seeing. If you are still seeing this problem, please let us know, and we will close this issue for now.

Thank you!

PR #14571, opened 2016-02-19
Proposal for auto updating roles

willthames:
##### Issue Type:
- Docs Pull Request
##### Ansible Version:

```
ansible 2.1.0 (devel f9526b2ab2) last updated 2016/02/19 09:46:55 (GMT +100)
  lib/ansible/modules/core: (detached HEAD 8d126bd877) last updated 2016/02/19 09:49:28 (GMT +100)
  lib/ansible/modules/extras: (detached HEAD f6c5ed987f) last updated 2016/02/19 09:49:04 (GMT +100)
  config file = 
```
##### Summary:

Proposal for auto updating roles when running ansible-playbook
##### Example output:

N/A

Reflects discussion on ansible-devel with @bcoca 

First attempt at an enhancement proposal as discussed at Contributors Summit.

### DIFF
--- a/docs/proposals/auto-install-roles.md
+++ b/docs/proposals/auto-install-roles.md
@@ -1,0 +1,106 @@
+# Auto Install Ansible roles
+
+*Author*: Will Thames <@willthames>
+
+*Date*: 19/02/2016
+
+## Motivation
+
+To use the latest (or even a specific) version of a playbook with the
+appropriate roles, the following steps are typically required:
+
+```
+git pull upstream branch
+ansible-galaxy install -r path/to/rolesfile.yml -p path/to/rolesdir -f
+ansible-playbook run-the-playbook.yml
+```
+
+The most likely step in this process to be forgotten is the middle step. While
+we can improve processes and documentation to try and ensure that this step is
+not skipped, we can improve ansible-playbook so that the step is not required.
+
+## Approaches
+
+### Approach 1: Specify rolesfile and rolesdir in playbook
+
+Provide new `rolesdir` and `rolesfile` keywords:
+
+```
+- hosts: application-env
+  become: True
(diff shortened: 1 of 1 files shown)


docs/proposals/auto-install-roles.md
```diff
@@ -0,0 +1,106 @@
...
+
+The most likely step in this process to be forgotten is the middle step. While
+we can improve processes and documentation to try and ensure that this step is
+not skipped, we can improve ansible-playbook so that the step is not required.
+
+## Approaches
+
+### Approach 1: Specify rolesfile and rolesdir in playbook
+
+Provide new `rolesdir` and `rolesfile` keywords:
```
resmo, line 26: I am generally +1 but there is already an option in ansible.cfg `roles_path` and I would suggest to reuse it and also implement a `roles_requirements_file` (default: requirements.yml) as well globally.

Alternatively they can be passed in every play as suggested.

willthames, line 26: Lines 40 and 41 discuss the configuration options. 

`roles_path` is fine for this purpose as long as `roles_path` can be relative to the playbook being run. I'm not sure if that's the case right now.

bcoca:
A couple of things that I consider as requirements for this:
- role versioning:
  - handle multiple local versions
  - different version markers (PR for tree does most of this)
  - unified role declaration
- role auto download

willthames:
I can see how role versioning might need to be a thing (especially if roles_path is shared by all playbooks) but it's not an essential requirement (we can always fail if conflicting roles attempt to be installed, at least in early phases). I suggest creating a seperate proposal for multiple versions of roles that explain how it will actually work.

Role auto download is the whole point of this proposal, so it'll definitely need to work.

I'm not convinced unified role declaration is a necessary requirement, at least not of approach 1.

PR #14591, opened 2016-02-20
Allow lookups to return python data structures

### LINKED ISSUE #14541: safe_eval prevents lookups from returning JSON
`safe_eval,` due to the fact that it uses ast and eval, prevents a lookup from returning JSON or a string that evals to a python data structure other than a string.

Take the example of:

```
lookup('template', 'build_json.json.j2')
```

`safe_json` will turn that output into a dict/list, instead of allowing JSON.

Before dc48d27dd2c49e95a719d7f7596f0b68e0ddef7f this worked ok, but largely due to a bug, that caused a `NameError` for `true`, `false` and `null`, as they could not be evaled.  As a result, the non-evaled string would be returned, instead of an evaluated string.

Now, because those values are converted, the lookup returns a dict.

This is a further problem, as `Templar._lookup` will raise `TypeError: sequence item 0: expected string, dict found`.

This represents as the following error message:

```
fatal: [foo]: FAILED! => {"failed": true, "msg": "an unexpected type error occurred. Error was sequence item 0: expected string, dict found"}
```

Reproducer:
##### playbook

```

---
- hosts: all
  gather_facts: false
  tasks:
    - debug: msg="{{ lookup('template', 'json.j2') }}"
```
##### template

```
{"hostname": "{{ inventory_hostname }}"}
```

sivel:
##### Issue Type:

 Bugfix Pull Request
##### Ansible Version:

```
v2
```
##### Summary:

Fixes #14541.  This change allows a lookup to return python data structures.  As is if a lookup returns a python data structure or a string that evaluates as a python data structure, a TypeError is raised while trying to join the results with a comma.

This change catches that TypeError and will either a single item of the list if the list was only 1 element long, or return the list itself.
##### Example output:

N/A

### DIFF
--- a/lib/ansible/template/__init__.py
+++ b/lib/ansible/template/__init__.py
@@ -420,7 +420,13 @@ class Templar:
                 if wantlist:
                     ran = wrap_var(ran)
                 else:
-                    ran = UnsafeProxy(",".join(ran))
+                    try:
+                        ran = UnsafeProxy(",".join(ran))
+                    except TypeError:
+                        if isinstance(ran, list) and len(ran) == 1:
+                            ran = wrap_var(ran[0])
+                        else:
+                            ran = wrap_var(ran)
 
             return ran
         else:


sivel:
This PR should also be cherry picked into stable-2.0

PR #14612, opened 2016-02-22
Verify that the role_spec is using a comma before warning

mscherer:
##### Issue Type:
- Bugfix Pull Request
##### Ansible Version:

```
ansible 2.1.0 (fix_spurious_warning 81f7bfd014) last updated 2016/02/22 19:51:26 (GMT +200)
  lib/ansible/modules/core: (detached HEAD 8d126bd877) last updated 2016/02/13 16:22:31 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD f6c5ed987f) last updated 2016/02/13 16:22:31 (GMT +200)
  config file = /home/misc/.ansible.cfg
```
##### Summary:

Since RoleRequirement.role_spec_parse is called even with a role_spec without a comma (see ./lib/ansible/playbook/role/requirement.py), it will always show the warning.

One way to trigger this is having this snippet in meta/main.yml:

```
dependencies:
  - role: foo
    when: "use_foo == True"
```

(see https://github.com/OSAS/ansible-role-httpd for a example of why that's used)

The code in RoleRequirement.role_yaml_parse is wrongly assuming this is using the old spec format, and thus trigger the warning.

### DIFF
--- a/lib/ansible/playbook/role/requirement.py
+++ b/lib/ansible/playbook/role/requirement.py
@@ -83,8 +83,7 @@ class RoleRequirement(RoleDefinition):
         #   'version': 'v1.0',
         #   'name': 'repo'
         # }
-
-        display.deprecated("The comma separated role spec format, use the yaml/explicit format instead.")
+        display.deprecated("The comma separated role spec format, use the yaml/explicit format instead. Line that trigger this: %s" % role_spec)
 
         default_role_versions = dict(git='master', hg='tip')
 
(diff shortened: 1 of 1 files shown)


bcoca:
The depreciation message should still always fire, we are probably calling this role spec parser incorrectly.

conorsch:
So how _should_ the dependency block be written? The example OP gave is this:

```
dependencies:
  - role: foo
    when: "use_foo == True"
```

Should it instead simply be:

```
dependencies:
  - galaxyusername.foo
```

with vars overridden elsewhere in the role?

ypid:
I am asking myself the same thing. Looking at the [source code](https://github.com/ansible/ansible/blob/85f4c958437f1ed10cdca93492d8f1b9ce7f6ddf/lib/ansible/playbook/role/requirement.py#L154) it seems it should be something like:

``` YAML
dependencies:
  - src: 'galaxyusername.role_name,version,role_name'
    role_name__role_variables: [ 'Some', 'thing' ]
```

Real example:

``` YAML
dependencies:
  - src: 'debops.secret,v0.2.3,secret'
```

Can someone please clarify that?

BTW. Why not use:

``` YAML
dependencies:
  - src: 'debops.secret'
    version: 'v0.2.3'
    name: 'secret'
```

?

SebastienElet:
I have no more warning with the `src` attribute, but i got error on a syntax check

```
ERROR! role definitions must contain a role name
…

dependencies:
- src: "user.role"
  ^ here
```

I use ansible `2.1.0.0` on OSX and Ubuntu

mscherer:
I moved the conditional upper in the stack, so this should solve @bcoca remark

PR #14614, opened 2016-02-22
non-destructive says we should run test_binary at v2

gundalow:
##### Issue Type:
- Bugfix Pull Request
##### Ansible Version:

```
ansible 2.0.0.2
```
##### Summary:

When I was looking at non_destructive.yml I saw that test_binary was disabled.

This appears to run correctly on Ubuntu 14.04
##### Example output:

```

~/git/ansible-inc/test/integration (test_binary)$ ansible-playbook non_destructive.yml -i inventory -e @integration_config.yml  -v --tags=test_binary

snip

testhost                   : ok=31   changed=11   unreachable=0    failed=0   
```

We are now using Ansible v2 and these tests pass, so run them.

[Review on Reviewable](https://reviewable.io/reviews/ansible/ansible/14614)

### DIFF
--- a/test/integration/non_destructive.yml
+++ b/test/integration/non_destructive.yml
@@ -42,6 +42,5 @@
     - { role: test_embedded_module, tags: test_embedded_module }
     - { role: test_uri, tags: test_uri }
     - { role: test_add_host, tags: test_add_host }
-    # Turn on test_binary when we start testing v2
-    #- { role: test_binary, tags: test_binary }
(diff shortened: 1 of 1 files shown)


gundalow:
The comments in roles/test_binary/tasks/main.yml make imply there may be extra tests that could be added.

Running the above tests is a "quick win", which is why I added added them

abadger:
test_binary currently passes so looks like a good thing to enable.

gundalow:
Thanks!

PR #14628, opened 2016-02-23
Introduce new 'filetree' lookup plugin (Ansible v1.9)

dagwieers:
##### Issue Type:

- Feature Pull Request
##### Ansible Version:

```
ansible 1.9.4
```
##### Summary:

This is a backport of the v2 plugin in PR #14332

The new "filetree" lookup plugin makes it possible to recurse over a tree of files within the task loop. This makes it possible to e.g. template a complete tree of files to a target system with little effort while retaining permissions and ownership.

The module supports directories, files and symlinks. And also SELinux.

The item dictionary consists of:
- src
- root
- path
- mode
- state
- owner
- group
- seuser
- serole
- setype
- selevel
- uid
- gid
- size
- mtime
- ctime

EXAMPLES:
Here is an example of how we use with_filetree within a role.
The `web/` path is relative to either `roles/<role>/files/` or `files/`.

``` yaml
 - name: Create directories
   file:
     path: /web/{{ item.path }}
     state: directory
     mode: '{{ item.mode }}'
     owner: '{{ item.owner }}'
     group: '{{ item.group }}'
     force: yes
   with_filetree: web/
   when: item.state == 'directory'

 - name: Template complete tree
   file:
     src: '{{ item.src }}'
     dest: /web/{{ item.path }}
     state: 'link'
     mode: '{{ item.mode }}'
     owner: '{{ item.owner }}'
     group: '{{ item.group }}'
   with_filetree: web/
   when: item.state == 'link'

 - name: Template complete tree
   template:
     src: '{{ item.src }}'
     dest: /web/{{ item.path }}
     mode: '{{ item.mode }}'
     owner: '{{ item.owner }}'
     group: '{{ item.group }}'
     force: yes
   with_filetree: web/
   when: item.state == 'file'
```

SPECIAL USE:
The following properties also have its special use:
- root: Makes it possible to filter by original location
- path: Is the relative path to root
- uid, gid: Makes it possible to force-create by exact id, rather than by name
- size, mtime, ctime: Makes it possible to filter out files by size, mtime or ctime

### DIFF
--- a/lib/ansible/runner/lookup_plugins/filetree.py
+++ b/lib/ansible/runner/lookup_plugins/filetree.py
@@ -76,7 +76,7 @@ def file_props(root, path):
         return None
 
     try:
-        st = os.stat(abspath)
+        st = os.lstat(abspath)
     except OSError, e:
         warning('filetree: Error using stat() on path %s (%s)' % (abspath, e))
         return None
--- a/lib/ansible/runner/lookup_plugins/filetree.py
+++ b/lib/ansible/runner/lookup_plugins/filetree.py
@@ -1,0 +1,123 @@
+# (c) 2016 Dag Wieers <dag@wieers.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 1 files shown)


bcoca:
sorry, but we are going to pass on this, we are not adding new features to the 1.9 tree

dagwieers:
That's unfortunate. Is a decision made wrt. a 1.9.5 release ?

bcoca:
yes, 1.9.5 will be bugfix only, we don't expect adding any features to the 1.9 tree in the future either.

fyi, this has no influence in the submission of same plugin to 2.x which will likely be accepted.

dagwieers:
Thanks. With "unfortunate" I meant that the same functionality will not be there out-of-the-box for those not being able to jump to 2.x yet. But I do understand the need for rules, and to stick to them :-)

Thanks for the feedback !

PR #14629, opened 2016-02-23
Note that version 3.11 of PyYAML is now required.

mavit:
Solaris 11.3 still ships PyYAML 3.09, resulting in the following error
being issued by Ansible:

```
ERROR! Unexpected Exception: 'AnsibleLoader' object has no attribute 'dispose'
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -403,6 +403,7 @@ allowed in future versions:
     ```
     - debug: msg="The error message was: {{error_code |default('') }}"
     ```
+* Version 3.11 of PyYAML is now required.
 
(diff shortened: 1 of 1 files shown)


sivel:
I wonder if this is indeed required.  The `_safe_load` method that used `dispose` seems to be based off of `yaml.parser.compose`.  In pyyaml 3.09 `compose` does not use a `dispose` method, maybe we could catch the exception and just let it pass?

cc @abadger

bcoca:
fixed via cc3cb0f65e4904775ca4e3eb277a606943e78887, which now ignores the attribute error when invoking dispose.

PR #14630, opened 2016-02-23
Fix code blocks, improve english grammar in docsite/rst/test_strategies.rst

dflock:
##### Issue Type:
- Docs Pull Request
##### Ansible Version:

n/a
##### Summary:

Fix code blocks, improve english grammar in docsite/rst/test_strategies.rst
- Fixed the yaml code-blocks to be valid rst & display correctly
- Improved the english for the opening sentence of the 'Testing Lifecycle' section.
##### Example output:

n/a

### DIFF
--- a/docsite/rst/test_strategies.rst
+++ b/docsite/rst/test_strategies.rst
@@ -25,8 +25,8 @@ things in your playbooks.
 
 .. code-block:: yaml
 
-   tasks:
-     - service: name=foo state=started enabled=yes
+  tasks:
+    - service: name=foo state=started enabled=yes
 
 If you think the service may not be started, the best thing to do is request it to be started.  If the service fails to start, Ansible
 will yell appropriately. (This should not be confused with whether the service is doing something functional, which we'll show more about how to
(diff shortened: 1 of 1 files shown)


bcoca:
`::` is valid rst afaik, and the default block type is already yaml, not sure if that fix is needed.

dflock:
You're probably correct - although the rst documentation (that I could find) is very unclear about this. Making it explicit does make the editing experience _much_ better though, as your editor can then highlight/lint/etc... the contents of the code block.

Also the original document was inconsistent - the first block was explicit, with the rest not. Maybe do it one way or the other?

I'm not sure what the convention is in the Ansible documentation for this, so if you want implicit blocks everywhere, then obviously, feel free to close this pull request. I could then put the grammar fix in another PR, as I should have done originally.

bcoca:
cc @docschick 2c?

docschick:
Just using :: for the code blocks is valid rst: http://www.sphinx-doc.org/en/stable/rest.html#source-code 

I guess I don't see the need in setting the code block explicitly as yaml or anything else (like http://www.sphinx-doc.org/en/stable/markup/code.html), just using :: has been enough for me.  

My personal vote is to keep them simple and standardize on ::  but if contributors want to make them explicit for their own linting needs, etc., that's cool. I'm not against it, just don't see it as always necessary, especially for the simpler code block examples.

PR #14633, opened 2016-02-24
Fixup boto3_conn aws_session_token and verify

Etherdaemon:
Fixup boto3_conn by removing unnecessary lines since commit https://github.com/ansible/ansible/commit/6ea772931fba2151fb2fb86caab8f7be10cf5769 which overwrote commit https://github.com/ansible/ansible/commit/27398131cf31eb7ca834a30ea2d8a871a937a377

Below snippet under def get_aws_connection_info(module, boto3=False) means it is not required anymore in boto3_conn.

```
if HAS_BOTO3 and boto3:
        boto_params = dict(aws_access_key_id=access_key,
                           aws_secret_access_key=secret_key,
                           aws_session_token=security_token)
```

### DIFF
--- a/lib/ansible/module_utils/ec2.py
+++ b/lib/ansible/module_utils/ec2.py
@@ -47,8 +47,6 @@ class AnsibleAWSError(Exception):
 
 def boto3_conn(module, conn_type=None, resource=None, region=None, endpoint=None, **params):
     profile = params.pop('profile_name', None)
-    params['aws_session_token'] = params.pop('security_token', None)
-    params['verify'] = params.pop('validate_certs', None)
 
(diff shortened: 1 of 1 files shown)


bcoca:
cc @defionscode

juliedavila:
shipit.

I'll add a note here just for reference.

In prior code 

``` python
params['aws_session_token'] = params.pop('security_token', None)        
params['verify'] = params.pop('validate_certs', None)
```

was needed in order to translate boto specific keywords in the params dict to valid boto3 keywords, and as @Etherdaemon mentioned, this 'conversion' occurs line lines 163 and 164 under `def get_aws_connection_info`

Thanks!

PR #14639, opened 2016-02-24
Fixed bug in find_mount_point function

ivovangeel:
##### Issue Type:

- Bugfix Pull Request
##### Ansible Version:

```
ansible 2.1.0 (devel 2db3f290ba) last updated 2016/02/24 15:40:00 (GMT +200)

```
##### Summary:

The find_mount_point function does not resolve the mount point of paths with a soft-link correctly and returns the wrong mount-point.
##### Example output:

```
TASK [copy file to nfs-mount] **************************************************
fatal: [hostname]: FAILED! => {"changed": false, "checksum": "f34b60930a5d6d689cf49a4c16bd7f9806be608c", "cur_context": ["system_u", "object_r", "nfs_t", "s0"], "failed": true, "gid": 24170, "group": "foundation", "input_was": ["system_u", "object_r", "default_t", "s0"], "mode": "0644", "msg": "invalid selinux context: [Errno 95] Operation not supported", "new_context": ["system_u", "object_r", "default_t", "s0"], "owner": "root", "path": "/soft-link-to-directory/.ansible_tmpWCT6Z4file", "secontext": "system_u:object_r:nfs_t:s0", "size": 37, "state": "file", "uid": 0}
```

The find_mount_point function does not resolve the mount point of paths with a soft-link correctly and returns the wrong mount-point.

I have mounted an NFS filesystem on /nfs-mount. This directory contains a directory called "directory". I also created a soft-link to this last directory: /soft-link-to-directory -> /nfs-mount/directory. I created the following task to copy a file into /soft-link-to-directory:

```
- name: copy file to nfs-mount
  copy:
    src: "file"
    dest: "/soft-link-to-directory/file"
```

This throws an exception:

invalid selinux context: [Errno 95] Operation not supported

This is caused by the find_mount_point function to return '/' as the mount point for '/soft-link-to-directory/file'. This should have been /nfs-mount. Because the find_mount_point returns the wrong mount-point, the is_special_selinux_path function does not recognise the file is on an NFS mount and tries to set the default SELinux context (system_u:object_r:default_t:s0), which fails. The context should have been: system_u:object_r:nfs_t:s0

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -726,7 +726,7 @@ class AnsibleModule(object):
         return (uid, gid)
 
     def find_mount_point(self, path):
-        path = os.path.abspath(os.path.expanduser(os.path.expandvars(path)))
+        path = os.path.realpath(os.path.expanduser(os.path.expandvars(path)))
         while not os.path.ismount(path):
             path = os.path.dirname(path)
         return path


ivovangeel:
Please note that this bug is also present in older Ansible versions. I am using Ansible 1.9.

jimi-c:
Merged, and cherry-picked to stable-2.1 so this will be included in 2.1.1. Thanks!

PR #14640, opened 2016-02-24
Update consul_io.py to support HTTPS endpoints

rojaro:
##### Issue Type:
- Bugfix Pull Request
##### Ansible Version:

```
ansible 2.0.0.2
```
##### Summary:

Adds HTTPS support to the consul_io.py

### DIFF
--- a/contrib/inventory/consul_io.py
+++ b/contrib/inventory/consul_io.py
@@ -410,6 +410,7 @@ class ConsulConfig(dict):
       '''get an instance of the api based on the supplied configuration'''
       host = 'localhost'
       port =  8500
+      scheme = 'http'
       token = None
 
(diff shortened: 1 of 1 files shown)


bcoca:
this seems to change it to default to http if it cannot get the scheme from the url passed

rojaro:
The default `scheme` in the Consul library is `http`. It also supports `https`, but to use it one just has to pass the `scheme` parameter, which `consul_io.py` did not do. So this patch is 100% compatible and does not change any default behavior, but allows us to use `https` instead of `http`.

Akasurde:
Closing this PR as same patch is merged via ef7fd10b1563ff8d4465b2a9c1a4fd224e15a0b6 

@rojaro Thanks for the contribution.

PR #14648, opened 2016-02-24
Show tasks with the same name in profile output

### LINKED ISSUE #14531: Ansible callback plugin profile_tasks doesn't handle tasks with duplicate name
ansible 2.0.0.2

Run against a set of tasks like so:

```
- name: Duplicate
  debug: msg="hi"

- name: Duplicate
  command: sleep 100
```

You'll notice you only get timing on one. This is flaw in the core architecture of this plugin. It uses a dict to store values by task name it seems.

@aioue

jjshoe:
##### Issue Type:
- Bugfix Pull Request
##### Ansible Version:

```
ansible 2.0.0.2
```
##### Summary:

Fixes #14531. Tasks with the same name would only display once in profiling output. Last one in won. Makes it tough to profile if you forget to change a name playing copy/pasta. 

@aioue is a good candidate to review.


aioue:
+1 thanks for doing the heavy lifting on that @jjshoe. I appreciate you keeping the sort in. Shipit

jjshoe:
@aioue guessing you didn't test this, I've found a bug, updating shortly.

jjshoe:
Rejecting this PR for now, going to upload a newer one.

PR #14649, opened 2016-02-24
Version requirement check by callback plugin

### LINKED ISSUE #4357: Allow playbooks to require a specific version of Ansible

viatcheslavmogilevsky:
##### Issue Type:
- Docs Pull Request
##### Ansible Version:

```
ansible 2.0.0.2
```
##### Summary:

Fixes #4357 

Example of callback plugin for establish actual ansible version. It is meaningful if playbooks repository is shared across many people. Since 2.0, supporting ansible 1.9.x and >2.0 versions is tough enough. It is better to support one version and made people to upgrade ansible.

### DIFF
--- a/lib/ansible/plugins/callback/version_requirement.py
+++ b/lib/ansible/plugins/callback/version_requirement.py
@@ -1,0 +1,44 @@
+#!/usr/bin/python
+
+import sys
+from ansible import __version__
+
+if __version__ < '2.0':
+    from ansible.callbacks import display as pre2_display
+
+    def display(*args, **kwargs):
+        pre2_display(*args, **kwargs)
+else:
+    from ansible.utils.display import Display
+
+    def display(*args, **kwargs):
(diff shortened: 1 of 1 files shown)


bcoca:
this can easily be done with an assert/fail in a play. In any case, including this callback in 2.0 is not going to affect users of 1.9 so not sure why we would add this ....

viatcheslavmogilevsky:
Nevermind, it is better to consider as galaxy plugin

bcoca:
closing this then, yes, it makes much more sense to distribute with your roles or plays that need it.

martbhell:
Would it be a lot of work to make this plugin look in the ansible galaxy meta file instead?

PR #14661, opened 2016-02-25
Update the profile task callback plugin to include a fix for duplicate named tasks. Added additional features to adjust the number of tasks output and the sort order.

### LINKED ISSUE #14531: Ansible callback plugin profile_tasks doesn't handle tasks with duplicate name

jjshoe:
##### Issue Type:
- Feature Pull Request
- Bugfix Pull Request
##### Ansible Version:

```
ansible 2.0.1.0
```
##### Summary:

Bug fix: Tasks are now stored by UUID in the dict. You could have 500 tasks with the same name, the last one profiled would be the only one to print. Fixes #14531 

Feature: Control how many tasks to show at the end
Feature: Control sort of tasks at the end

Tested using: https://github.com/jjshoe/profile_tasks_test

### DIFF
--- a/lib/ansible/plugins/callback/profile_tasks.py
+++ b/lib/ansible/plugins/callback/profile_tasks.py
@@ -1,3 +1,4 @@
+# (C) 2016, Joel, http://github.com/jjshoe 
 # (C) 2015, Tom Paine, <github@aioue.net>
 # (C) 2014, Jharrod LaFon, @JharrodLaFon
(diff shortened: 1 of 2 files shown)


bcoca:
updated it to show old output by default and paths when in verbose mode (also on next line)

Issue #14686, opened 2016-02-27
Ansible makes it difficult to copy files recursively and preserve permissions


larsks:
##### Issue Type:
- Bug Report
##### Ansible Version:

ansible 2.0.1.0
  config file = 
  configured module search path = Default w/o overrides
##### Ansible Configuration:

n/a
##### Environment:

n/a (Alpine Linux)
##### Summary:

The `copy` module does not preserve file permissions, and the `synchronize` module does not respect `ansible_connection`.
##### Steps To Reproduce:

Let's say you have a role `myrole` that contains:

   $ pwd
   .../roles/myrole
   $ find . -type f
   ./tasks/main.yml
   ./files/adirectory/ascript.sh
   ./files/adirectory/afile.txt
   $ ls -l files/adirectory/ascript.sh
   -rwx-r-xr-x ... 

Try to copy the files using `copy`:
- copy:
     src: adirectory/
     dest: /tmp/adirectory/

Result:

The files are copied, but the script is mode `644`.

Try to copy the files using the `synchronize` module:

```
- synchronize:
    src: adirectory/
    dest: /tmp/adirectory/
```

Result:

```
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "Failed to find required executable ssh"}
```

Well of _course_ it failed; I'm using a `local` connection because
`ssh` isn't even installed.
##### Expected Results:

I expected one of the above mechanisms to successfully copy a directory tree while preserving permissions.
##### Actual Results:

Sadness and surprise.

larsks:
I see that 2.1.x seems to have fixed the `synchronize` module for `local` connections.  Yay!  But I still think `copy` should be able to do this.

andyceo:
Well, nice to see fix in the `larsks/ansible` repository! But guys, how about fix this in upstream?

yfshare:
```
192.168.31.110 | FAILED! => {
    "changed": false, 
    "failed": true, 
    "msg": "Failed to find required executable ssh"
}
```
This error, we need to install openssh-clients. I have solved it.
```
yum -y install openssh-clients
```
```
ansible test_hosts -m synchronize -a 'src=/tmp/helloworld.txt dest=/tmp/'
192.168.31.110 | SUCCESS => {
    "changed": true, 
[...]

PR #14687, opened 2016-02-27
allow copy action to preserve file mode

larsks:
this allows the copy action to preserve file mode by setting the value
of mode to the keyword `copy`.  For exampe:

```
- copy:
    src: mydirectory/
    dest: /tmp/mydirectory
    mode: copy
```

This resolves the problem reported in #14686, and is particularly
useful for copying directories containing scripts and other
executables.

### DIFF
--- a/lib/ansible/plugins/action/copy.py
+++ b/lib/ansible/plugins/action/copy.py
@@ -228,6 +228,17 @@ class ActionModule(ActionBase):
 
                 # Run the copy module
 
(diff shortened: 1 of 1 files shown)


larsks:
If there is interest in this PR I am happy to update it, otherwise I will plan on closing it in the near future.

larsks:
Ah, poor pull request, we hardly knew ye.

PR #14689, opened 2016-02-27
Updated versions with current released and development

jmatsushita:
##### Issue Type:
- Docs Pull Request
##### Ansible Version:

This is about the currently online ansible docs which reads `Last updated on Feb 26, 2016.`
##### Summary:

Update the version numbers after seeing [this note](http://docs.ansible.com/ansible/uri_module.html#notes) which mentioned Ansible 2.1.

To be honest, it feels like the docs should be kept updated for each major version, but I understand that it would be hard to maintain.


jmatsushita:
Or another way to "solve" this is to add the version number covered by the doc much more prominently on all pages. For instance right by the "Edit on Github" link.

bcoca:
yes a more permanent (autoupdate?) solution would be nicer.

PR #14691, opened 2016-02-27
Update Lookups documentation regarding a possible pitfall with password plugin

udesilva:
##### Issue Type:
- Docs Pull Request
##### Ansible Version:

```
ansible 2.0.0.2
  config file = /home/uds/provisioning/buildbootiso/ansible.cfg
```
##### Example output:

```

```

Added a warning regarding attempting to create a local password file when the user the playbook is running as cannot create it... causes Ansible to loop at that task.

This might be considered a bug also in that Ansible never recovers from this user error...

### DIFF
--- a/docsite/rst/playbooks_lookups.rst
+++ b/docsite/rst/playbooks_lookups.rst
@@ -69,6 +69,8 @@ This length can be changed by passing an extra parameter::
 
 .. note:: If the file already exists, no data will be written to it. If the file has contents, those contents will be read in as the password. Empty files cause the password to return as an empty string        
 
+Caution: Because this runs on the ansible host as the user running the playbook, if the file does not exist, and is not writeable by that user, the playbook will just loop.
+
 Starting in version 1.4, password accepts a "chars" parameter to allow defining a custom character set in the generated passwords. It accepts comma separated list of names that are either string module attributes (ascii_letters,digits, etc) or are used literally::
 
     ---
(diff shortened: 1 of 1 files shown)


docschick:
I think this would be better if written as an admonition, like the note above it, but as a warning. Like the following:

`.. warning:: 
    Because this runs on the ansible host as the user running the playbook, if the file does not exist, and is not writeable by that user, the playbook will just loop.`

bcoca:
soo, what does 'the playbook will just loop' mean?

the rest of the warning is more of a note: "This runs on the Ansible host as the user running the playbook, you must have access to read or, if missing, create the file."

udesilva:
Hmm. With 2.0.0.2, I was getting the ansible-playbook process sitting there apparently doing nothing at all after printing the name of the play -- but when I ran strace against the ansible-playbook process, it was forking and waiting on the forks (which were dying off immediately). Now, I can't reproduce that problem -- it just dies straight away with a (fairly obvious) message that permission was denied.

I guess Brian's comment is now appropriate, with the sole caveat that specifying "become: true" doesn't help with this. My workaround has been to generate the password files in a directory that's writeable by the local user, then copy them into the protected directory using the "copy" module, which DOES respect "becomes"; this obviously means that a new password is generated each time, and unfortunately undermines idempotence. I could copy the files back from the protected directory using the slurp module, but that's rather messy and undermines security instead...

In any event, I've rewritten the text, including the example of the use case that tripped me up.

bcoca:
not just become, no privilege escalation nor connection variables or directives apply to lookups

dharmabumstead:
Thanks for your work @udesilva!

PR #14699, opened 2016-02-29
Set fail_on_cloud_config in openstack inventory

jhesketh:
##### Issue Type:
- Feature Pull Request
##### Ansible Version:

```
ansible 2.1.0
  config file = 
```
##### Summary:

When managing across multiple openstack clouds, if any one of them is unavailable for whatever reason (the cloud may be down or the credentials misconfigured etc) then no hosts are returned from the dynamic inventory.

In some cases you may still want an inventory returned for whichever clouds are available. This is particularly useful if you're doing CD across multiple clouds for resiliency.

This relies on the recently merged upstream change in shade: https://review.openstack.org/#/c/285882/

From the commit message:

Add support for a new option to the openstack inventory. This is so
should one cloud be unavailable you can still list hosts from any
other openstack clouds you have configured.

This is exposed as an option under the extra config part of ansible
in the openstack clouds.yaml.

### DIFF
--- a/contrib/inventory/openstack.py
+++ b/contrib/inventory/openstack.py
@@ -3,6 +3,7 @@
 # Copyright (c) 2012, Marco Vito Moscaritolo <marco@agavee.com>
 # Copyright (c) 2013, Jesse Keating <jesse.keating@rackspace.com>
 # Copyright (c) 2015, Hewlett-Packard Development Company, L.P.
+# Copyright (c) 2016, Rackspace Australia
 #
 # This module is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -39,12 +40,17 @@
 # use_hostnames changes the behavior from registering every host with its UUID
 #               and making a group of its hostname to only doing this if the
 #               hostname in question has more than one server
+# fail_on_errors causes the inventory to fail and return no hosts if one cloud
+#                has failed (for example, bad credentials or being offline).
+#                When set to False, the inventory will return hosts from
+#                whichever other clouds it can contact. (Default: True)
 
 import argparse
 import collections
 import os
 import sys
 import time
+from distutils.version import StrictVersion
 
 try:
     import json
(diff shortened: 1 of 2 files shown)


bcoca:
cc @emonty @Shews

emonty:
I think this is a great option - but I think we need to make sure we can plumb it through from clouds.yaml too. (I'm sort of embarrased I never thought of this before)

jhesketh:
Yeah I had thought about exposing this in the config or elsewhere, but I'm not entirely sure how global you'd want to make it. For example, would you have a flag that says "never fail on an individual cloud - ie try them all"?

My thought process was that in most use cases you want to know when you've failed to do an operation on a cloud (in particular with shade and os-client-config), however the inventory is something where it may be reasonable to default to working partially. Especially since the --debug flag will give you the connection failures.

emonty:
Please rebase this and remove the merge commits.

However, I was thinking something like this:

http://paste.openstack.org/show/490823

which would be an ansible inventory specific option. The main problem is that if you just have (like we do in infra) a directory with inventory things and one of them is a copy of this script, there is no way to tell ansible to pass it the --fail-on-inventory-error command line arg.

jhesketh:
Thanks Monty. Done.

Because this relies on the recently merged https://review.openstack.org/#/c/285882/, do we need to release a new version of shade? I notice for the extra_config args you effectively inspect shade to see if it supports the extra config. There is no trivial way of doing that with the current shade change so we could either:
1) require a newer version of shade (not sure how since shade isn't tracked as a requirement)
2) do a try/exception for the extra argument
3) add a new commit to shade that is detectable to see if the fail_on_cloud_config parameter exists
4) use the inspect module to check if the parameter exists

Shrews:
Yes, a new shade version would need to be released.

I've been promoting using StrictVersion, as the os_ironic_node.py core module does, to verify shade version. So if this new arg is supplied, you would then make sure the shade version is recent enough.

jhesketh:
@Shrews @emonty ping. Just wondering if you have a moment to look at this please.

PR #14735, opened 2016-03-01
Lookup netrc for credentials in open_url

jsok:
##### Issue Type:
- Feature Pull Request
##### Ansible Version:

```
ansible 2.1.0 (open_url-netrc 9c4548e583) last updated 2016/03/02 09:26:26 (GMT +1100)
  lib/ansible/modules/core: (detached HEAD 45367c3d09) last updated 2016/03/02 09:12:20 (GMT +1100)
  lib/ansible/modules/extras: (detached HEAD 479f99678b) last updated 2016/03/02 09:13:50 (GMT +1100)
  config file = /Users/jsok/.ansible.cfg
```
##### Summary:

Adds support for fetching HTTP basic auth credentials from an [netrc](https://docs.python.org/2/library/netrc.html) file.

e.g. Installing ansible roles via `ansible-galaxy` which are hosted on a private HTTP server.
##### Example output:

To use a netrc not in the standard search path of the `netrc` module (`$HOME/.nertc` on unixes):

```
env NETRC=/path/to/netrc ansible-galaxy install -r requirements.yml
```

Otherwise existing behaviour and output it unmodified.

### DIFF
--- a/lib/ansible/module_utils/urls.py
+++ b/lib/ansible/module_utils/urls.py
@@ -258,6 +258,7 @@ if not HAS_MATCH_HOSTNAME:
 
 
 import httplib
+import netrc
 import os
 import re
 import sys
@@ -691,6 +692,9 @@ def open_url(url, data=None, headers=None, method=None, use_proxy=True,
     if parsed[0] != 'ftp':
         username = url_username
 
+        if headers is None:
+            headers = {}
+
         if username:
             password = url_password
             netloc = parsed[1]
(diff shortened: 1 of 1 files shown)


lib/ansible/module_utils/urls.py
```diff
@@ -716,10 +720,15 @@ def open_url(url, data=None, headers=None, method=None, use_proxy=True,
...
 
         elif username and force_basic_auth:
-            if headers is None:
-                headers = {}
+            headers["Authorization"] = basic_auth_header(username, password)
 
-            headers["Authorization"] = "Basic %s" % base64.b64encode("%s:%s" % (username, password))
+        else:
+            rc = netrc.netrc(os.environ.get('NETRC'))
+            login = rc.authenticators(parsed[1])
```
bcoca: this assumes NETRC is always present, you should make it conditional on it actually resulting in something

jsok: If `NETRC` is not in the environment, then `.get` will return `None`.

`netrc`'s behaviour is to:

> If no argument is given, the file .netrc in the user’s home directory will be read

You have however pointed out the scenario there `$HOME/.netrc` does not exist, in which case it throws `IOError: [Errno 2] No such file or directory: '/Users/jsok/.netrc'`

I'll push a fix now to ensure we don't propagate that exception!

jsok:
I've addressed the original comments and pushed a fix, ready for re-review.

jsok:
Hi @bcoca, is there anything else I can do to help get this PR reviewed and merged?

PR #14742, opened 2016-03-02
ansible-playbook: Add option to exit with an error if no plays were m…

### LINKED ISSUE #14693: "No hosts matched" should return with error code != 0
##### Issue Type:
- Feature Idea
##### Ansible Version:

2.1.0
##### Summary:

I have a "deploy" job in Jenkins but with — for example — a mistyped hostname, the call to `ansible-playbook` will (correctly) print "No hosts matched" but return with an exit code of zero. This seems to violate some kind of UNIX idea, at least in principle.
##### Expected Results:

Return with a non-zero exit code.

lamby:
##### Issue Type:
- Feature Pull Request
##### Summary:

Fixes: #14693

I have a "deploy" job in Jenkins. When I mistype a hostname in the
_group_ definitions, the call to ansible-playbook will happily print "No
hosts matched" and return with an exit code of zero. The CI system will
then believe everything was deployed successfully.

This seems to violate some kind of UNIX principle to me and now requires
parsing of the Jenkins log file for this particular string, which is
rather ugly.

The current default behaviour is correct - the no hosts matched message
occurs at the _play_ level and not at the playbook level, hence making
this an option that is disabled by default
##### Example output:

```
  $ ansible-playbook [..] --error-if-no-plays-matched
  <snip>

  PLAY [deploy] *************************************************
  skipping: no hosts matched

  PLAY RECAP ****************************************************

  [ERROR]: No plays were matched by any host.

  $ echo $?
  1
```

### DIFF
--- a/lib/ansible/cli/playbook.py
+++ b/lib/ansible/cli/playbook.py
@@ -72,6 +72,9 @@ class PlaybookCLI(CLI):
             help="one-step-at-a-time: confirm each task before running")
         parser.add_option('--start-at-task', dest='start_at_task',
             help="start the playbook at the task matching this name")
+        parser.add_option('--error-if-no-plays-matched',
+            dest='error_if_no_plays_matched', action='store_true',
+            help="exit with a non-zero status code if no plays were matched by any host")
 
         self.options, self.args = parser.parse_args(self.args[1:])
 
--- a/lib/ansible/executor/playbook_executor.py
+++ b/lib/ansible/executor/playbook_executor.py
@@ -80,7 +80,9 @@ class PlaybookExecutor:
                     self._tqm.load_callbacks()
                     self._tqm.send_callback('v2_playbook_on_start', pb)
 
-                i = 1
+                num_plays_total = 0
+                num_plays_matched = 0
+
                 plays = pb.get_plays()
                 display.vv(u'%d plays in %s' % (len(plays), to_unicode(playbook_path)))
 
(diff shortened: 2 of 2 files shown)


jjshoe:
You can achieve this now with a callback plugin, and you wouldn't need to wait on a new release of ansible. 

https://gist.github.com/jjshoe/ace3070906e5bc5cc432

bcoca:
So after discussing this with the community, we are going to pass on this feature. Part of it is not wanting to grow the command line switch list with arbitrary features, the other is that there are other ways to achieve the same w/o modifying Ansible.

For those that really need this option it is easy enough to code into a callback as done here:
https://gist.github.com/jjshoe/ace3070906e5bc5cc432 

This callback is not something we are considering to include in Ansible as it does seem too much of a niche feature.

jjshoe:
Note, callback could possibly be made simpler by writing your plugin for v2_playbook_on_no_hosts_matched

lamby:
>  it does seem too much of a niche feature

My "deploy" example doesn't convince? Ah well. Thanks for the example callback and the pointer re. `v2_playbook_on_no_hosts_matched` - will look into that.

jjshoe:
v2_playbook_on_no_hosts_matched doesn't work, as it's per play. plugin above is what you want.

eli-betito:
@bcoca @jjshoe 

Can you help me to figure out how to implement your plugin into my Ansible ?

Is there a more recent solution that can catch play/playbook's errors especially this error: "skipping: no hosts matched" ?

Thanks !

jpic:
Let's not reproduce the situation we had with the debug output logger that the community had to carry around versions in their repositories: i recommend to at least merge the callback plugin, not only doesn't it harm anything, but also in CD pipeline I just had a GREEN for a deploy job that had no host matched so that's definitely a requirement :joy:

jpic:
Having green CD jobs when "no host matched" is not possible for us, we are maintaining a version (that passes flake8) [here](https://yourlabs.io/oss/playlabs/blob/master/playlabs/callback_plugins/failer.py).

PR #14743, opened 2016-03-02
Proposal: deprecate always_run in favor of checkmode_run

resmo:
##### Issue Type:
- Proposal


bcoca:
'ignore_checkmode' ?

resmo:
@bcoca is also good for me. As long as it does not contain the word `always`

PR #14755, opened 2016-03-02
avoid issues when stdin is a closed file

bcoca:
##### Issue Type:
- Bugfix Pull Request
##### Ansible Version:

```
2.0.1
```
##### Summary:

Avoid error with pause module when using ansible under `nohup`.

This seems to happen when nohup is involved, so the check tty
does not get a chance to fail, it just works with pipes

fixes http://github.com/ansible/ansible-modules-core/issues/3166

### DIFF
--- a/lib/ansible/plugins/action/pause.py
+++ b/lib/ansible/plugins/action/pause.py
@@ -120,23 +120,29 @@ class ActionModule(ActionBase):
 
             # save the attributes on the existing (duped) stdin so
             # that we can restore them later after we set raw mode
-            fd = self._connection._new_stdin.fileno()
-            if isatty(fd):
-                old_settings = termios.tcgetattr(fd)
-                tty.setraw(fd)
+            fd = None
+            try:
+                fd = self._connection._new_stdin.fileno()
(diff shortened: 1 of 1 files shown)


lib/ansible/plugins/action/pause.py
```diff
@@ -120,20 +120,26 @@ def run(self, tmp=None, task_vars=None):
...
-                    key_pressed = self._connection._new_stdin.read(1)
-                    if key_pressed == '\x03':
-                        raise KeyboardInterrupt
+                    if fd is not None:
+                        key_pressed = self._connection._new_stdin.read(1)
+                        if key_pressed == '\x03':
+                            raise KeyboardInterrupt
 
                     if not seconds:
                         if not isatty(fd):
```
nitzmahone: Doesn't isatty bomb on none? It does for me...

nitzmahone:
Works for me as expected when run from a non-interactive shell (ie, I see the warning and it does the right thing).

nitzmahone:
cherry-picking to stable-2.0

PR #14769, opened 2016-03-03
Ensure provided mode is compared on equal grounds with existing mode

### LINKED ISSUE #14771: Using full st_mode results in difference in check-mode run and real run
##### Issue Type:

- Bug Report
##### Ansible Version:

The issue exists for both 1.9.4 and 2.0.1.0 (and likely any other Ansible version)
##### Summary:

When using (full)python permission modes, Ansible behaves differently in check-mode than it would during a real run.

Granted, this is very unusual for someone to do (except programmatically), but we found it by accident and would prefer to see this fixed.
##### Steps To Reproduce:

When using mode='0100755' which is valid for a real run because Ansible handles it correctly, it fails during check-mode because it was not converted properly.

``` yaml
- hosts: localhost
  gather_facts: no
  tasks:
  - copy:
      src: /etc/fstab
      dest: /tmp/
      mode: '0100755'
```

##### Expected Results:

```
[dag@moria ansible.testing]$ ansible-playbook test71.yml

ok: [localhost]

[dag@moria ansible.testing]$ ansible-playbook -CD test71.yml

ok: [localhost]

```
##### Actual Results:

```
[dag@moria ansible.testing]$ ansible-playbook test71.yml

ok: [localhost]

[dag@moria ansible.testing]$ ansible-playbook -CD test71.yml

changed: [localhost]

localhost                  : ok=1    changed=1    unreachable=0    failed=0   
```

dagwieers:
##### Issue Type:

- Bugfix Pull Request

This fixes #14771 for Ansible 2.x
##### Ansible Version:

```
ansible 2.0.1.0
  config file = /home/dag/home-made/ansible.testing/ansible.cfg
```
##### Summary:

We noticed that doing template/copy on files with a slightly different octal mode than expected, check-mode would result in a change, while a real run would return ok.

We traced this to the fact that Ansible converts existing permissions using sstat.S_IMODE(), however it doesn't do so with the provided mode. So when we provided '0100755', it compared '0755' with '0100755'. In a real run this would be equal, but in check-mode it would be considered a change.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -884,6 +884,7 @@ class AnsibleModule(object):
                                    details=str(e))
 
         prev_mode = stat.S_IMODE(path_stat.st_mode)
+        mode = stat.S_IMODE(mode)
 
         if prev_mode != mode:
 


bcoca:
I don't think anyone ever tried setting the 'filesystem type' part of the mode before ... I'm not sure if we should 'accept' it and deal with it or error out as it seems like bad input.

dagwieers:
I understand. What I wanted to fix was the difference between check-mode and a real run.

Currently a real run does accept the "raw" file mode and handles it correctly. So I fixed the check-mode to do the same. Again, as you say, it's unlikely a lot of people will experience this.

PS The meaning of 'mode' is in fact the complete file mode, and not just the permission bits. So if we just take it by its name we should be liberal to what we accept... :-)

bcoca:
@dagwieers there are probably a dozen people that would interpret mode that way, outside the other 10 that actually implement this feature inside the filesystem. Most understand mode as the permissions, specially since that is the only things most people have access to change, the other 'mode bits' are just descriptive of what the filesystem object is and not really meant to be changed directly by the user (would be fun changing filetype from device to symlink ....) .

tonk:
I completely missed the link between this pull and issue #14771. Sorry for that. Please, ignore my fix.

PR #14770, opened 2016-03-03
Ensure provided mode is compared on equal grounds with existing mode (Python 1.9)

### LINKED ISSUE #14771: Using full st_mode results in difference in check-mode run and real run

dagwieers:
##### Issue Type:

- Bugfix Pull Request

This fixes #14771 for Ansible 1.9.x
##### Ansible Version:

```
ansible 1.9.4
```
##### Summary:

This is a backport of #14769

We noticed that doing template/copy on files with a slightly different octal mode than expected, check-mode would result in a change, while a real run would return ok.

We traced this to the fact that Ansible converts existing permissions using sstat.S_IMODE(), however it doesn't do so with the provided mode. So when we provided '0100755', it compared '0755' with '0100755'. In a real run this would be equal, but in check-mode it would be considered a change.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -664,6 +664,7 @@ class AnsibleModule(object):
                                    details=str(e))
 
         prev_mode = stat.S_IMODE(path_stat.st_mode)
+        mode = stat.S_IMODE(mode)
 
         if prev_mode != mode:
             if self.check_mode:


bcoca:
discussion in the 2.0 version of this ticket #14769

PR #14779, opened 2016-03-03
Use custom download url if supplied by galaxy

javiplx:
##### Issue Type:
- Feature Pull Request
##### Summary:

This change enables installing roles which are not github produced tarballs. It searchs the (currently unused) url field returned by galaxy on queries like https://galaxy.ansible.com/api/v1/roles/7595/versions/, and if it is not empty for the selected role, it use that value to download the role.

### DIFF
--- a/lib/ansible/galaxy/role.py
+++ b/lib/ansible/galaxy/role.py
@@ -219,6 +219,10 @@ class GalaxyRole(object):
                     if role_versions and self.version not in [a.get('name', None) for a in role_versions]:
                         raise AnsibleError("- the specified version (%s) of %s was not found in the list of available versions (%s)." % (self.version, self.name, role_versions))
 
+                matched_version = filter( lambda v : v['name'] == self.version , role_versions )
+                if len(matched_version) == 1 :
+                    self.src = matched_version[0].get('url', self.src) or self.src
+
                 tmp_file = self.fetch(role_data)
 
         else:
--- a/lib/ansible/galaxy/role.py
+++ b/lib/ansible/galaxy/role.py
@@ -221,7 +221,7 @@ class GalaxyRole(object):
 
                 matched_version = filter( lambda v : v['name'] == self.version , role_versions )
                 if len(matched_version) == 1 :
-                    self.src = matched_version[0].get('url', self.src) or self.src
+                    self.src = matched_version[0].get('download', self.src) or self.src
 
                 tmp_file = self.fetch(role_data)
 


javiplx:
I forgot to mention that I use this on [ansible-library](https://github.com/javiplx/ansible-library/), which is a install-only galaxy implementation for private roles

bcoca:
cc @chouseknecht

bcoca:
iirc that url was to point at issue handlers for the role, in case github was not used

javiplx:
The url I reuse it at the version level, not at role level. I've reused it because it is currently empty and is on the right level.
https://galaxy.ansible.com/api/v1/roles/7595/versions/

```
{
    ...
    "results": [
        {
            "url": "", 
            "id": 10537, 
            "name": "v0.99", 
        }, 
        {
            "url": "", 
            "id": 10422, 
            "name": "v0.9", 
        },
        ...
    ]
}
```

chouseknecht:
That 'url' attribute is intended to point internally. In Django Rest
Framework parlance it is intended to be a link to the detail view of the
particular object. So in this case the url should be something like
/api/v1/role/###/versions/###. It would link to the specific version and
perhaps provide greater detail about the version. However, we do not have a
view setup to handle this, so it's blank.

Also, and more importantly, that 'url' attribute is generated by the
serializer. It's not data coming form the Galaxy database.

All that being said... There is no reason why we could not have a link to
the particular SCM version of the role. It could be a link to a branch,
commit, or whatever. The question is how do we get the correct URL imported
into Galaxy? Today we simply use GitHub tags, which makes getting new
versions into Galaxy quite simple. Add a new tag to your repo. Import the
role in Galaxy. Galaxy reads in the tags. Done.

It has been proposed here
https://github.com/ansible/ansible/blob/devel/docs/proposals/auto-install-roles.md
that
we change role versioning, allowing version to be a reference to nearly
anything in a GitHub repo. I think we likely have to add something to
meta/main.yml so that Galaxy knows how to create versions for your role.
Until we know what that is, and we figure out how to store it in Galaxy and
surface it in the API, I don't think we're ready to make the modification
you have submitted.

--Chris

On Thu, Mar 3, 2016 at 5:46 PM, Javier Palacios notifications@github.com
wrote:

> The url I reuse it at the version level, not at role level. I've reused it
> because it is currently empty and is on the right level.
> https://galaxy.ansible.com/api/v1/roles/7595/versions/
> 
> {
>     ...
>     "results": [
>         {
>             "url": "",
>             "id": 10537,
>             "name": "v0.99",
>         },
>         {
>             "url": "",
>             "id": 10422,
>             "name": "v0.9",
>         },
>         ...
>     ]
> }
> 
> —
> https://github.com/ansible/ansible/pull/14779#issuecomment-192002507.

## 

Chris Houseknecht
Principal Engineer
RedHat | Ansible http://ansible.com
919.803.9159 <9198039159> c

javiplx:
The `/api/v1/role/###/versions/###` is close to what I actually expected, as usual on galaxy itself and most other similar rest apis. I started using it but now I might favor to use a different one (download maybe?), mainly as you state that the url field is automatically generated.
The main point for this change (and ansible-library itself), is the requirement to have a single role on a single public github repo at toplevel for simple management. I have no role at toplevel, and despite of being unseful for the community, some of the roles will be private (as our software is). I know I can use requirements.yml for actually quite complex downloads, but just using "name,version" is by far more appealing.
If you like, I will submit a pull request with my own thoughts to the autoinstall roles page.

chouseknecht:
If you want to add an update to the auto install role proposal (https://github.com/ansible/ansible/blob/devel/docs/proposals/auto-install-roles.md),
go for it. All comments welcome.

--Chris

sivel:
The `ansible-galaxy` command is being phased out in favor of [mazer](https://github.com/ansible/mazer) and is no longer receiving updates to functionality.

Please review the functionality of mazer, and potentially open a feature request with that project.

PR #14780, opened 2016-03-03
Feature/configurable galaxy

javiplx:
##### Issue Type:
- Feature Pull Request
##### Summary:

This change implements the same functionality than #14779 to download roles from custom locations.
It also implements configuration of galaxy server on ansible.cfg, as current 2.x releases do.

### DIFF
--- a/bin/ansible-galaxy
+++ b/bin/ansible-galaxy
@@ -257,7 +257,7 @@ def api_get_config(api_server, ignore_certs=False):
         validate_certs = False
 
     try:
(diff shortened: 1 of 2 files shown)


javiplx:
I forgot to mention that I use this on [ansible-library](https://github.com/javiplx/ansible-library/), which is a install-only galaxy implementation for private roles

bcoca:
we don't add features to minor releases, if we eventually push out a 1.9.5 it will just have bugfixes.

PR #14793, opened 2016-03-04
remove main_q for simplicity.

xiaket:
##### Issue Type:
- Bugfix Pull Request
##### Ansible Version:

```
2.1.0
```
##### Summary:

Oh, sorry. This is not actually a Bugfix, some code in ansible are simply obsolete and should be removed, that's all.
##### Example output:

main_q is not used anywhere in the codebase.

It is created in TaskQueueManager._initialize_processes, bundled with rslt_q
into TaskQueueManger._workers, later unwrapped in StrategyBase but not used.
This queue is closed in TaskQueueManger._cleanup_processes.

Historically, main_q is passed as an init parameter into WorkerProcess.
It is introduced in 62d7956, but this init behavior is changed in 120b9a7.

### DIFF
--- a/lib/ansible/executor/process/result.py
+++ b/lib/ansible/executor/process/result.py
@@ -65,7 +65,7 @@ class ResultProcess(multiprocessing.Process):
         result = None
         starting_point = self._cur_worker
         while True:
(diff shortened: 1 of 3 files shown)


xiaket:
Oh, I believe the failing check has nothing to do with my commit.

PR #14796, opened 2016-03-04
prop: ignore_checkmode is a better name

resmo:
##### Issue Type:
- Docs Pull Request

as suggested by @bcoca


bcoca:
i did misspell ignore in IRC, but you might want to fix it in the proposal ....

resmo:
ooobs, my fault :)

PR #14797, opened 2016-03-04
Use to_bytes on filenames in filesystem calls.

mattclay:
##### Issue Type:

Bugfix Pull Request
##### Ansible Version:

```
ansible 2.1.0 (devel fa630872d9) last updated 2016/03/04 01:11:06 (GMT -700)
  config file = 
```
##### Summary:

Added `to_bytes(path, errors='strict')` for paths being passed to filesystem calls to prevent unicode paths from throwing encoding errors.
##### Example output:

Without these fixes unicode filenames used by modules such as copy and fetch can result in a UnicodeEncodeError when LANG=C on the control host. For example:

```
```

### DIFF
--- a/lib/ansible/parsing/dataloader.py
+++ b/lib/ansible/parsing/dataloader.py
@@ -118,7 +118,7 @@ class DataLoader():
 
     def path_exists(self, path):
         path = self.path_dwim(path)
-        return os.path.exists(to_bytes(path))
+        return os.path.exists(to_bytes(path, errors='strict'))
 
     def is_file(self, path):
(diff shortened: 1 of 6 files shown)


mattclay:
@abadger: Here's the first round of to_bytes fixes we discussed on IRC.

I found the areas requiring changes and then tested the fixes using connection plugin tests I created, which are available here: https://github.com/mattclay/ansible/tree/connection-test

If you think there's value in those tests I can submit a PR for those as well.

abadger:
Merged.  Thanks @mattclay for cleaning those up.

PR #14805, opened 2016-03-04
Accept both " import *" and " import AnsibleModule"

dbarrosop:
##### Issue Type:

- Bugfix Pull Request
##### Ansible Version:

```
ansible 2.0.0.2
  config file =
```
##### Summary:

When developing an ansible module ansible forces you to import the AnsibleModule with the line:

`from ansible.module_utils.<lib name> import *`

which breaks PEP8. But there is no real reason to do not this instead:

`from ansible.module_utils.<lib name> import AnsibleModule`

which doesn't break anything.
##### Example output:

N/A

### DIFF
--- a/lib/ansible/executor/module_common.py
+++ b/lib/ansible/executor/module_common.py
@@ -94,7 +94,7 @@ def _find_snippet_imports(module_data, module_path, strip_comments):
             import_error = False
             if len(tokens) != 3:
                 import_error = True
-            if b" import *" not in line:
+            if (b" import *" not in line) and not (b" import AnsibleModule" in line):
                 import_error = True
             if import_error:
                 raise AnsibleError("error importing module in %s, expecting format like 'from ansible.module_utils.<lib name> import *'" % module_path)


sivel:
# Possible Misunderstanding

Hi!

We believe the ticket you have filed is being somewhat misunderstood, as one thing works a little differently than stated.

That "import" is not a real import, and is actually used as replacement text.  That "import" line, gets replaced with the full contents of the actual file.

In the future, this might be a topic more well suited for the user list, which you can also post here if you'd like some more help with the above.
- https://groups.google.com/forum/#!forum/ansible-project

dbarrosop:
I understand it's not a real import so I don't understand why you enforce it by breaking PEP8. Specially when the solution in this PR works without breaking any stuff.

sivel:
Technically it does, as it would lead to incorrect assumptions.

`import *` and `import AnsibleModule` are different, and, in the end, with your change the result is always the same as `import *`.  So by specifying `import AnsibleModule` you are creating an incorrect assumption for anyone who sees it that you only have access to `AnsibleModule` as opposed to everything within that file.

Additionally, this does not break pep8.  F403 comes from pyflakes, and it is just a warning.

To get technical, the fact that we require them at the bottom is a pep8 validation issue, but this is done to keep line numbers correct in error messages due to the code insertion, and is not something we plan on changing as of yet.

You may be interested in the planned work around Ziploader which is outlined at https://github.com/ansible/ansible/blob/devel/ROADMAP.md

- IRC: #ansible on irc.freenode.net
- mailing list: https://groups.google.com/forum/#!forum/ansible-project

dbarrosop:
I am sorry but I completely disagree. I am not breaking any assumption. I am not the one using something that the whole python community treats as an import to actually replace it with code coming from somewhere else. So if someone is breaking assumptions here it is actually ansible. You made a very bad choice by reusing python keywords to do something that _only_ ansible does as that confuses people, linters, IDEs and everybody. Ask non-core developers from ansible about that line and tell me how many people knows that you are actually replacing that line with code with something coming from outside. Tests passed and my whole environment is working fine with that patch so I still don't understand what things and assumptions this PR breaks.

Regarding the Ziploader, that looks really awesome. Is there an ETA? I wouldn't mind using the devel branch if it has that particular feature.

abadger:
For those coming across this later: Discussion has moved to the mailing list: https://groups.google.com/d/msg/ansible-project/0bwHEFfKOro/meU337AxCQAJ

abadger:
I'm going to lock this issue just so we don't have new discussion happening here that we probably won't see.  Discussion on the mailing list may lead to a different Pull Request, though, that introduces different syntax for triggering module replacer (and ziploader may also make the underlying issue moot).

PR #14812, opened 2016-03-04
Add tests for connection plugins.

mattclay:
##### Issue Type:

Feature Pull Request
##### Ansible Version:

```
ansible 2.1.0 (test-connection ca62bc5db3) last updated 2016/03/04 13:16:31 (GMT -700)
  config file = 
```
##### Summary:

Add connection plugin integration tests for:
- local
- chroot
- ssh
- paramiko_ssh

The same tests run for each connection plugin, with both the default locale and the C locale, with and without pipelining enabled.

These tests were used to find the bugs fixed by ansible/ansible#14797 and then test those fixes.

The local and chroot plugin tests will run as part of the build. If these tests are merged I can work on adding support for testing ssh and paramiko_ssh in docker containers with a local SSH server.
##### Example output:

Assuming root and a local SSH server, full tests can be run manually with:

```
TEST_FLAGS='-l "" -e ansible_user=some-local-user' make test_connection
```

Resulting in a play recap like this (twice, once for the default locale and once for LANG=C):

```
chroot-no-pipelining       : ok=9    changed=6    unreachable=0    failed=0   
chroot-pipelining          : ok=9    changed=6    unreachable=0    failed=0   
local-no-pipelining        : ok=9    changed=6    unreachable=0    failed=0   
local-pipelining           : ok=9    changed=6    unreachable=0    failed=0   
paramiko_ssh-no-pipelining : ok=9    changed=6    unreachable=0    failed=0   
paramiko_ssh-pipelining    : ok=9    changed=6    unreachable=0    failed=0   
ssh-no-pipelining          : ok=9    changed=6    unreachable=0    failed=0   
ssh-pipelining             : ok=9    changed=6    unreachable=0    failed=0   
```

### DIFF
--- a/test/integration/Makefile
+++ b/test/integration/Makefile
@@ -73,18 +73,17 @@ environment: setup
 non_destructive: setup
 	ansible-playbook non_destructive.yml -i $(INVENTORY) -e outputdir=$(TEST_DIR) -e @$(VARS_FILE) $(CREDENTIALS_ARG) -v $(TEST_FLAGS)
 
-test_connection: setup
-ifeq ($(EUID),0)
-	# Test connection plugins when running as root (lang unspecified).
-	                ansible-playbook test_connection.yml -i test_connection.inventory -l '!skip-during-build' $(TEST_FLAGS)
-	# Test connection plugins when running as root (lang=C).
-	LC_ALL=C LANG=C ansible-playbook test_connection.yml -i test_connection.inventory -l '!skip-during-build' $(TEST_FLAGS)
-else
-	# Test connection plugins when not running as root (lang unspecified).
-	                ansible-playbook test_connection.yml -i test_connection.inventory -l '!skip-during-build !chroot' $(TEST_FLAGS)
-	# Test connection plugins when not running as root (lang=C).
-	LC_ALL=C LANG=C ansible-playbook test_connection.yml -i test_connection.inventory -l '!skip-during-build !chroot' $(TEST_FLAGS)
+# Skip connection plugins which require root when not running as root.
+ifneq ($(EUID),0)
+TEST_CONNECTION_FILTER := !chroot
 endif
+
+# Connection plugin test command to repeat with each locale setting.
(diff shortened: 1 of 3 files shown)


test/integration/Makefile
```diff
@@ -72,6 +73,19 @@ environment: setup
...
+	# Test connection plugins when running as root (lang unspecified).
+	                ansible-playbook test_connection.yml -i test_connection.inventory -l '!skip-during-build' $(TEST_FLAGS)
+	# Test connection plugins when running as root (lang=C).
+	LC_ALL=C LANG=C ansible-playbook test_connection.yml -i test_connection.inventory -l '!skip-during-build' $(TEST_FLAGS)
+else
+	# Test connection plugins when not running as root (lang unspecified).
+	                ansible-playbook test_connection.yml -i test_connection.inventory -l '!skip-during-build !chroot' $(TEST_FLAGS)
+	# Test connection plugins when not running as root (lang=C).
+	LC_ALL=C LANG=C ansible-playbook test_connection.yml -i test_connection.inventory -l '!skip-during-build !chroot' $(TEST_FLAGS)
+endif
```
bcoca, line 79: instead of duplicating the commands, use the if to set a var and use that in the command

mattclay, line 79: I've eliminated the command duplication using variables.

test/integration/test_connection.yml
```diff
@@ -0,0 +1,35 @@
+- hosts: all
+  gather_facts: no
+  serial: 1
+  tasks:
+
```
bcoca, line 5: we might want to avoid ssh hosts by default as that setup is not guaranteed for the tests.

mattclay, line 5: The inventory group "skip-during-build" lists ssh and paramiko_ssh, so those won't run automatically when running `make test_connection`, but you can override that with TEST_FLAGS.

If we can get a local ssh server running in the docker containers during build in the future we could allow those to run in that scenario.

PR #14813, opened 2016-03-04
Add lxd connection plugin.

mattclay:
##### Issue Type:

Feature Pull Request
##### Ansible Version:

```
ansible 2.1.0 (lxd-plugin 43929197f1) last updated 2016/03/04 13:37:19 (GMT -700)
  config file = 
```
##### Summary:

Add a connection plugin for [lxd](https://linuxcontainers.org/lxd/). Connects to a local or remote lxd server using the lxc command line client. Works with and without pipelining.

Tested on a Ubuntu 15.10 control host with a Ubuntu 15.10 lxd host, using the official lxd install as well as the official PPA version. Tested manually as well as with the tests in PR ansible/ansible#14812.
##### Example output:

Connecting to a local container:

```
$ ansible -m ping -i 'centos-7-amd64,' all -e 'ansible_connection=lxd'
centos-7-amd64 | SUCCESS => {
    "changed": false, 
    "ping": "pong"
}
```

Connecting to a remote container:

```
$ ansible -m ping -i 'iron:ubuntu-wily-amd64-os,' all -e 'ansible_connection=lxd'
iron:ubuntu-wily-amd64-os | SUCCESS => {
    "changed": false, 
    "ping": "pong"
}
```

If a container is not running:

```
$ ansible -m ping -i 'iron:ubuntu-wily-amd64-pip,' all -e 'ansible_connection=lxd'
iron:ubuntu-wily-amd64-pip | FAILED! => {
    "failed": true, 
    "msg": "container not running: iron:ubuntu-wily-amd64-pip"
}
```

If a container is not found:

```
$ ansible -m ping -i 'iron:does-not-exist,' all -e 'ansible_connection=lxd'
iron:does-not-exist | FAILED! => {
    "failed": true, 
    "msg": "container not found: iron:does-not-exist"
}
```

If the lxc command line client is not found:

```
# ansible -m ping -i 'iron:does-not-exist,' all -e 'ansible_connection=lxd'
iron:does-not-exist | FAILED! => {
    "failed": true, 
    "msg": "lxc command not found in PATH"
}
```

### DIFF
--- a/lib/ansible/plugins/connection/lxd.py
+++ b/lib/ansible/plugins/connection/lxd.py
@@ -1,0 +1,112 @@
+# (c) 2016 Matt Clay <matt@mystile.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import os
+from distutils.spawn import find_executable
+from subprocess import call, Popen, PIPE
+
+from ansible.errors import AnsibleError, AnsibleConnectionFailure, AnsibleFileNotFound
+from ansible.plugins.connection import ConnectionBase
+from ansible.utils.unicode import to_bytes, to_unicode
+
+
+class Connection(ConnectionBase):
+    """ lxd based connections """
+
+    transport = "lxd"
+    has_pipelining = True
+
+    def __init__(self, play_context, new_stdin, *args, **kwargs):
+        super(Connection, self).__init__(play_context, new_stdin, *args, **kwargs)
+
+        self._host = self._play_context.remote_addr
+        self._lxc_cmd = find_executable("lxc")
+
+        if not self._lxc_cmd:
+            raise AnsibleError("lxc command not found in PATH")
+
+        if self._play_context.remote_user is not None and self._play_context.remote_user != 'root':
+            self._display.warning('lxd does not support remote_user, using container default: root')
+
+    def _connect(self):
+        """connect to lxd (nothing to do here) """
+        super(Connection, self)._connect()
+
+        if not self._connected:
+            self._display.vvv(u"ESTABLISH LXD CONNECTION FOR USER: root", host=self._host)
+            self._connected = True
+
+    def exec_command(self, cmd, in_data=None, sudoable=True):
+        """ execute a command on the lxd host """
+        super(Connection, self).exec_command(cmd, in_data=in_data, sudoable=sudoable)
+
+        self._display.vvv(u"EXEC {0}".format(cmd), host=self._host)
+
+        local_cmd = [self._lxc_cmd, "exec", self._host, "--", self._play_context.executable, "-c", cmd]
+
+        local_cmd = [to_bytes(i, errors='strict') for i in local_cmd]
+        in_data = to_bytes(in_data, errors='strict', nonstring='passthru')
+
+        process = Popen(local_cmd, stdin=PIPE, stdout=PIPE, stderr=PIPE)
+        stdout, stderr = process.communicate(in_data)
+
+        stdout = to_unicode(stdout)
+        stderr = to_unicode(stderr)
+
+        if stderr == "error: Container is not running.\n":
+            raise AnsibleConnectionFailure("container not running: %s" % self._host)
+
+        if stderr == "error: not found\n":
+            raise AnsibleConnectionFailure("container not found: %s" % self._host)
+
+        return process.returncode, stdout, stderr
+
+    def put_file(self, in_path, out_path):
+        """ put a file from local to lxd """
+        super(Connection, self).put_file(in_path, out_path)
+
+        self._display.vvv(u"PUT {0} TO {1}".format(in_path, out_path), host=self._host)
+
+        if not os.path.isfile(to_bytes(in_path, errors='strict')):
+            raise AnsibleFileNotFound("input path is not a file: %s" % in_path)
+
+        local_cmd = [self._lxc_cmd, "file", "push", in_path, self._host + "/" + out_path]
+
+        local_cmd = [to_bytes(i, errors='strict') for i in local_cmd]
+
+        call(local_cmd)
+
+    def fetch_file(self, in_path, out_path):
+        """ fetch a file from lxd to local """
+        super(Connection, self).fetch_file(in_path, out_path)
+
+        self._display.vvv(u"FETCH {0} TO {1}".format(in_path, out_path), host=self._host)
+
+        local_cmd = [self._lxc_cmd, "file", "pull", self._host + "/" + in_path, out_path]
+
+        local_cmd = [to_bytes(i, errors='strict') for i in local_cmd]
+
+        call(local_cmd)
+
+    def close(self):
+        """ close the connection (nothing to do here) """
+        super(Connection, self).close()
+
+        self._connected = False
--- a/test/integration/test_connection.inventory
+++ b/test/integration/test_connection.inventory
@@ -48,6 +48,13 @@ paramiko_ssh-no-pipelining ansible_ssh_pipelining=false
 ansible_host=localhost
 ansible_connection=paramiko_ssh
 
+[lxd]
+lxd-pipelining    ansible_ssh_pipelining=true
+lxd-no-pipelining ansible_ssh_pipelining=false
+[lxd:vars]
+ansible_host=centos-7-amd64
+ansible_connection=lxd
+
 [test_default:children]
 local
 chroot


lib/ansible/plugins/connection/lxd.py
```diff
@@ -0,0 +1,116 @@
...
+import ansible.constants as C
+from ansible.errors import AnsibleError, AnsibleFileNotFound
+from ansible.plugins.connection import ConnectionBase
+from ansible.utils.unicode import to_str, to_bytes, to_unicode
+
+try:
+    from __main__ import display
+except ImportError:
+    from ansible.utils.display import Display
+    display = Display()
```
bcoca: you should not need this as connectionbase already has self._display

lib/ansible/plugins/connection/lxd.py
```diff
@@ -0,0 +1,116 @@
...
+
+        self._connected = True
+
+    def exec_command(self, cmd, in_data=None, sudoable=True):
+        """ execute a command on the lxd host """
+        super(Connection, self).exec_command(cmd, in_data=in_data, sudoable=sudoable)
+
+        display.vvv(u"EXEC {0}".format(cmd), host=self._host)
+
+        executable = C.DEFAULT_EXECUTABLE.split()[0] if C.DEFAULT_EXECUTABLE else "/bin/sh"
```
bcoca: action plugin should have already taken care of this

lib/ansible/plugins/connection/lxd.py
```diff
@@ -0,0 +1,116 @@
...
+
+        local_cmd = [self._lxc_cmd, "exec", self._host, "--", executable, "-c", cmd]
+
+        local_cmd = map(to_bytes, local_cmd)
+        in_data = to_bytes(in_data)
```
bcoca: Not sure you want to do this if in_data is None

lib/ansible/plugins/connection/lxd.py
mattclay: Thanks for pointing out the extra display. I've fixed that.

lib/ansible/plugins/connection/lxd.py
mattclay: I've fixed the to_bytes call on in_data to return None if that's what's given. This is the same as what the ssh plugin does.

lib/ansible/plugins/connection/lxd.py
nitzmahone: Yes and no- the action plugin sticks the right executable wrapper on there, but it returns it as a single string (which absolutely won't work when passed directly to lxc- I checked). This second level of wrap lets the shell on the remote-side parse that command- the alternative would be shlex.split(), but that would make the behavior of "raw" inconsistent in this plugin compared to all others. eg:
```
raw: echo hi && echo there
```
would be run under a shell and return "hi there" on every other connection plugin type, but for this one would return "hi && echo there" because a shell didn't parse it first. 

Seems like we have to do the remote shell if we want raw's behavior to remain consistent...

lib/ansible/plugins/connection/lxd.py
```diff
@@ -0,0 +1,110 @@
...
+
+        self._connected = True
+
+    def exec_command(self, cmd, in_data=None, sudoable=True):
+        """ execute a command on the lxd host """
+        super(Connection, self).exec_command(cmd, in_data=in_data, sudoable=sudoable)
+
+        self._display.vvv(u"EXEC {0}".format(cmd), host=self._host)
+
+        executable = C.DEFAULT_EXECUTABLE.split()[0] if C.DEFAULT_EXECUTABLE else "/bin/sh"
```
nitzmahone: @bcoca - your comment about the action_plugin already doing this is half-correct- the action plugin sticks the right executable wrapper on there, but it returns the command as a single string (which absolutely won't work when passed directly to lxc- I checked). This second level of wrap lets the shell on the remote-side parse that command- the alternative would be shlex.split(), but that would make the behavior of "raw" inconsistent in this plugin compared to all others. eg:

```
raw: echo hi && echo there
```

would be run under a shell and return "hi there" on every other connection plugin type, but for this one would return "hi && echo there" because a shell didn't parse it first.

Seems like we have to do the second remote shell if we want raw's behavior to remain consistent...

bcoca: we can probably change it to a list, still this would be the wrong executable in some cases as it does not allow per host config, its also redundant to fallback to /bin/sh as it always is set (default is /bin/sh already).

In other cases it will still be a problem, specially when privilege escalation is involved, due to the quoting.

mattclay: Looking over all the connection plugins, I'm not sure changing cmd to a list actually helps the situation. Every connection plugin passes cmd through a shell, either explicitly within the plugin or by way of delegating that responsibility to the remote host.

These plugins use subprocess with `shell=true` as the shell to run cmd:
- local

These plugins use `C.DEFAULT_EXECUTABLE` as the shell to run cmd:
- chroot
- docker
- jail
- libvirt_lxc
- lxd

These plugins let the remote system select the shell to run cmd:
- funcd
- paramiko_ssh
- ssh
- zone

So it seems all the plugins (including this one) are about about as consistent as they can be, although the local plugin could be updated to not use `shell=true`.

Additionally, using `C.DEFAULT_EXECUTABLE` for the shell may not be the best choice, but if we're going to change that, we should do so for all of the plugins to keep things consistent.

bcoca:
it is a bit more subtle than  that, the reason we specify a shell on the action plugin/privilege escalation side is to control the quoting. 

The connection should not specify the shell as it does not have all the info that it needs to select the correct one, this happens in the command construction so there is no need of adding it again, the plugins that do so now need to be corrected.

mattclay:
Since `_low_level_execute_command` passes `cmd` to all connections as a string it's always being parsed by a shell specified by the connection.

The only difference between connections is how that shell is chosen:
- ssh - The shell of the ssh user on the remote host is used.
- lxd - The shell chosen by the connection plugin is used.

Either way the shell parsing `cmd` is selected without any awareness of quoting requirements for the command.

That's why `_low_level_execute_command` wraps `cmd` in an additional shell before passing it to `self._connection.exec_command`.

If there are scenarios where this doesn't work, changing some plugins to use a list isn't going to help, because others, like ssh, don't have that option, they can only pass a string to the remote host.

Is there a specific scenario where the current behavior with ssh breaks?

bcoca:
no, as ssh does not set it's own shell or quoting, it works in all scenarios, the ones that will break are the ones constructing the cmd with their own shell, like local and lxd

mattclay:
Correct, the ssh connection plugin does not set its own shell or quoting, but the remote host does.

Here's a demo showing the ssh and lxd connection plugins behave the same:

Give "tester" a custom shell so we can see the args it receives:

```
$ grep tester /etc/passwd
tester:x:1001:1001:,,,:/home/tester:/home/tester/shell
```

Here's the custom shell:

```
$ cat /home/tester/shell
#!/usr/bin/python
import sys
print('remote host shell: {0}'.format(sys.argv))
```

Here's a manual connection using ssh:

```
$ ssh tester@localhost 'command goes here'
remote host shell: ['/home/tester/shell', '-c', 'command goes here']
```

Here's a manual connection using lxd:

```
$ lxc exec centos-7-amd64 -- /home/tester/shell '-c' 'command goes here'
remote host shell: ['/home/tester/shell', '-c', 'command goes here']
```

Here's a connection using ansible ssh:

```
$ ansible -m ping -i 'localhost,' -e 'ansible_user=tester ansible_connection=ssh ansible_ssh_pipelining=true' all
localhost | FAILED! => {
    "changed": false, 
    "failed": true, 
    "module_stderr": "", 
    "module_stdout": "remote host shell: ['/home/tester/shell', '-c', \"/bin/sh -c 'LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8 LC_MESSAGES=en_US.UTF-8 /usr/bin/python'\"]\n", 
    "msg": "MODULE FAILURE", 
    "parsed": false
}
```

Here's a connection using ansible lxd (with '/home/tester/shell' instead of DEFAULT_EXECUTABLE):

```
$ ansible -m ping -i 'centos-7-amd64,' -e 'ansible_connection=lxd ansible_ssh_pipelining=true' all
centos-7-amd64 | FAILED! => {
    "changed": false, 
    "failed": true, 
    "module_stderr": "", 
    "module_stdout": "remote host shell: ['/home/tester/shell', '-c', \"/bin/sh -c 'LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8 LC_MESSAGES=en_US.UTF-8 /usr/bin/python'\"]\n", 
    "msg": "MODULE FAILURE", 
    "parsed": false
}
```

There is no difference in how ssh or lxd execute `cmd`. They both use a shell chosen by the connection to execute `cmd`.

What am I missing?

mattclay:
@bcoca: Once PR #14845 is merged I'll update this PR to use the new cmd list. I already have the changes locally, since I tested PR #14845 against this plugin.

mattclay:
I've rebased and squashed the requested fixes back into the original commit. At this point I think I'm just waiting on PR #14845 so I can update the lxd plugin to use cmd as a list.

mattclay:
Rebased and switched from `C.DEFAULT_EXECUTABLE` to the new `self._play_context.executable`.

There's still a double shell, as with other plugins, but the lxd plugin no longer chooses the shell.

mattclay:
Rebased to resolve merge conflicts. I'll squash when finished.

mattclay:
@bcoca: Now that this plugin is using `self._play_context.executable`, is there anything else you'd like changed (aside from squashing commits), before this can be merged?

When https://github.com/ansible/ansible/pull/14845 is completed I can update this plugin for the new cmd list, along with the other plugins.

PR #14814, opened 2016-03-04
Allow executing ansible like 'python -m ansible playbook'

sivel:
##### Issue Type:

Feature Pull Request
##### Ansible Version:

```
v2.1
```
##### Summary:

I saw a user asking if ansible could be executed via `python -m` due to `/usr/bin/env python` returning the default python2.4 for a system.  

As an experiment, I created this.  If it is deemed to not have much value, feel free to close.
##### Example output:

```
$ python -m ansible playbook | head -2
ERROR! You must specify a playbook file to run
$ python -m ansible | head -2
Usage: ansible <host-pattern> [options]
ERROR! Missing target hosts
```


sivel:
Just as a mention, using setuptools entry points could eliminate this, as setuptools writes it's own shebang pointing to the python binary that was used during install.  However this makes it a bit harder for running from source, requiring aliases for ansible/ansible-playbook, which isn't as portable.

bcoca:
if you do a `make install` or install through packages it does use setuptools to rewrite the shebang, not sure his is very useful change.

sivel:
> not sure his is very useful change

You may indeed be right, like I said, kind of an experiment, and figured I wouldn't just throw it away, if it seemed like something we may want.

sivel:
Looks like this has some unintended side effects.  I don't know that it is worth really looking into them.  Going to close this for now.

ssbarnea:
Please reopen this, the ability to call ansible with a specific version of python is essential, and is the only safe way to control how Python applications are loaded. Running ansible as a module -- `python -m ansible` must be a documented way of calling it, if not the even the recommended way.

All major python module developers that do provide command line tools for calling the module will be able to tell you that this is the recommended way to call it. The command line tools are shortcuts, but these shortcuts do have their own limitations.

bcoca:
@ssbarnea, you are reversing the purpose of the project, the 'command line tools' ARE the main focus, it 'happens' to use python.  The API is not the project focus and can change w/o warning or compatibility in service of the CLI. As such, there is little interest in changes that benefit the API in detriment of the CLI.

ssbarnea:
Let me be clear, I do like command line tools and there is nothing wrong with them, when they work. Still due to the nature of ansible (dependency on a specific version of Python, which may not be the default python version on that platform), we need to be able to start it using the module approach. 

I do not expect the end-users to use this approach, but **I do expect to be a very popular way of starting it by any Ansible developer**, as that's the way to debug it properly, to test it with different interpreters and/or virtual environments.

Is not something that is supposed be hard to implement because the command line parsing logic would still be in a single place. If you are not aware about how this is supposed to be implemented you can check [py.test](https://pytest.org/latest/usage.html) or [flake8](https://gitlab.com/pycqa/flake8/tree/master) which are quite a good examples of tools that allow to be called in both ways.

Think that you may have 4-5 different python interpreters installed (on Mac almost everyone has both system one and the homebrew one) and you want to start ansible using a specific one.

PR #14823, opened 2016-03-05
add find_vm_by_name function to vmware utils

jjahns:
##### Issue Type:
- Feature Pull Request
##### Ansible Version:

```
ansible 2.0.1.0
```
##### Summary:

This is an addition to the vmware module utilities.  It is a simple function that can do a search for a virtual machine, similar to how the other functions are defined.

In many cases, such as when utilizing automation to deploy virtual machines, this is a useful function because it aids in making automation related to virtual machines idempotent.
##### Example output:

```

```

### DIFF
--- a/lib/ansible/module_utils/vmware.py
+++ b/lib/ansible/module_utils/vmware.py
@@ -99,6 +99,15 @@ def find_hostsystem_by_name(content, hostname):
     return None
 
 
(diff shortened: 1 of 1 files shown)


jjahns:
There looks to be a failure during build that doesn't appear to be related to this, unless I'm mistaken.

PR #14824, opened 2016-03-05
draft 1st release of ansible shell

bcoca:
##### Issue Type:
- Feature Pull Request
##### Ansible Version:

2.1
##### Summary:

porting @dominis 's tool from 1.9 and integrating it into ansible


ghjm:
I kicked the tires on this and noticed the following that work in my pull request at dominis/ansible-shell/pull/51 but do not work here:
- Running 'ansible-console -i inventory' gives an arguments error because it is expecting a host pattern. You should be able to just launch this with an inventory. If a host pattern is specified it should be used as the initial cwd.
- Tab completion of hosts does not work.
- Cd to a wildcard host pattern does not work.
- The 'become' command should print a confirmation of the new state after the toggle.
- A syntax error in module parameters (e.g., 'user foo') crashes out of ansible-console - it should print an error and return to the ansible-console prompt.

Also, I think the following are needed improvements to both ansible-shell and ansible-console:
- Ctrl+C currently exits to bash, which feels wrong - it should return to the ansible-console prompt.
- It is hard to find the ansible-console specific commands (cd, list, remote_user, become) in the output of 'help' because they are mixed in with the list of modules. It would be ideal to separate these out. And maybe throw in some pagination.

Thanks,

-Graham

bcoca:
pushed version that fixes most of the issues above, still todo:
- Tab completion of hosts on 'cd'
- color code builtin commands to differentiate from modules
- color code deprecated modules

sivel:
@bcoca, looks like there might be a rebase issue here.  The only commit is showing `proposal: publish/subscribe for handlers` and includes the changes from the commit with that actual name.

thaumos:
I have tested so far, and tab complete is working for me on hosts while in a group.

bcoca:
pushed fix, amended commit history, also fixed tab completion

ghjm:
+100

On Wed, Mar 9, 2016 at 1:21 PM, Brian Coca notifications@github.com wrote:

> Merged #14824 https://github.com/ansible/ansible/pull/14824.
> 
> —
> https://github.com/ansible/ansible/pull/14824#event-583708508.

## 

Graham Mainwaring
Senior Software Engineer
Ansible by Red Hat

http://rc-1.com/asb/email/+1 919 332 6966 |  http://rc-1.com/asb/email/
ansible.com https://www.ansible.com/

dominis:
:+1: very good, thanks for this @bcoca

bcoca:
@dominis thank you for creating this in the first place!

rolandjohann:
awesome work!
where do one find some documentation about that tool?

PR #14833, opened 2016-03-06
make all conditionals lists

### LINKED ISSUE #13905: failed_when directive does not accept lists
##### Issue Type:

Bug Report
##### Ansible Version:

2.0.0.1
##### Environment:

ubuntu 14.04
##### Summary:

The `failed_when` module does not support lists anymore
##### Steps To Reproduce:

```
- hosts: localhost
  tasks:
    - name: hello
      command: "/dev/null"
      register: myresult
      failed_when:
        - "myresult == 1"
        - "myresult == 2"
```
##### Expected Results:

pass
##### Actual Results:

> ERROR! The field 'failed_when' is supposed to be a string type, however the incoming data structure is a <class 'ansible.parsing.yaml.objects.AnsibleSequence'>

bcoca:
##### Issue Type:
- Bugfix Pull Request
##### Ansible Version:

2.0.x
##### Summary:

this brings them to equivalence with `when:`, making all conditionals behave the same and take the same input (lists are implicit 'and's)

fixes #13905

### DIFF
--- a/lib/ansible/playbook/task.py
+++ b/lib/ansible/playbook/task.py
@@ -70,11 +70,11 @@ class Task(Base, Conditional, Taggable, Become):
 
     _any_errors_fatal     = FieldAttribute(isa='bool')
     _async                = FieldAttribute(isa='int', default=0)
-    _changed_when         = FieldAttribute(isa='string')
+    _changed_when         = FieldAttribute(isa='list', default=[])
     _delay                = FieldAttribute(isa='int', default=5)
     _delegate_to          = FieldAttribute(isa='string')
     _delegate_facts       = FieldAttribute(isa='bool', default=False)
-    _failed_when          = FieldAttribute(isa='string')
+    _failed_when          = FieldAttribute(isa='list', default=[])
     _first_available_file = FieldAttribute(isa='list')
     _loop                 = FieldAttribute(isa='string', private=True)
     _loop_args            = FieldAttribute(isa='list', private=True)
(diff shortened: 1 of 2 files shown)


willthames:
In #13905,  the `failed_when` suggests an implicit or rather than an implicit and.

Should there be tests for this?

bcoca:
the lists are an 'and' in `when`, so keeping it the same, will point that out if this is accepted.

missing both tests and docs

willthames:
I think that this is a serious break to backward compatibility - it's probably better that lists remain broken than change behaviour.

This definitely does not fix #13905.

lib/ansible/playbook/task.py
```diff
@@ -70,11 +70,11 @@ class Task(Base, Conditional, Taggable, Become):
 
     _any_errors_fatal     = FieldAttribute(isa='bool')
     _async                = FieldAttribute(isa='int', default=0)
-    _changed_when         = FieldAttribute(isa='string')
+    _changed_when         = FieldAttribute(isa='list', default=[])
     _delay                = FieldAttribute(isa='int', default=5)
```
nitzmahone, line 74: doesn't this style of initializer on FieldAttribute cause the instance aliasing issue (eg, where all instances end up sharing the same empty list)?

nitzmahone, line 74: nevermind- I see we added the isa check to deepcopy those for specific cases

bcoca:
@willthames this brings backwards compatibility back, the example has a logic error but the ticket is about a syntax error. In previous versions these worked with lists and they were also implicit ANDs.

lib/ansible/playbook/task.py
bcoca, line 74: @nitzmahone yes, we fixed that in the processing code to allow definitions to stay the same.

willthames:
@bcoca, I see - I've tested the behaviour under 1.9 and you're quite right.

willthames:
@bcoca, the following playbook fails when run against ansible containing this PR.

```
- hosts: localhost
  connection: local
```

(Or just run `ansible -m setup localhost`)

At the end of the setup output I see

```
    "changed": true, 
    "failed": true, 
    "failed_when_result": true
```

bcoca:
@willthames good catch, made sure empty/false conditional defaults to false for all conditionals except when:, which needs to be true

PR #14845, opened 2016-03-07
Add support for cmd list to connection plugins.

### LINKED ISSUE #10450: libvirt_lxc connection won't work on non-SElinux enabled platforms (like Ubuntu)
There is a long-standing bug in virsh that will affect users of Ubuntu or any platform that does not ship SElinux:

https://www.redhat.com/archives/libvirt-users/2013-July/msg00146.html

On Ubuntu platforms, the only advertised security model for LXC, per `virsh -c lxc:/// capabilities`, is `none`:

``` xml
    <secmodel>
      <model>none</model>
      <doi>0</doi>
    </secmodel>
```

When executing `virsh -c lxc:/// lxc-enter-namespace` like the libvirt_lxc connection plugin does, this results in the following error:

```
$ virsh -c lxc:/// lxc-enter-namespace test /bin/bash
libvirt:  error : argument unsupported: Security model none cannot be entered
```

The libvirt_lxc connection plugin would see this on every `exec_command`, `put_file` and `fetch_file`, making the plugin rather unusable under the circumstances. 

The workaround suggested in the ML thread is to invoke `lxc-enter-namespace` with the `--noseclabel` option:

```
$ virsh -c lxc:/// lxc-enter-namespace --noseclabel test /bin/bash
root@test:/#
```

I'm hesitant to send a patch for `libvirt_lxc.py` to that effect though, because I have no way to test on an SElinux-enabled LXC system.

mattclay:
##### Issue Type:

Feature Pull Request
##### Ansible Version:

```
ansible 2.1.0 (cmd-list 6c2e32b046) last updated 2016/03/10 16:27:09 (GMT -700)
  lib/ansible/modules/core: (detached HEAD c86a0ef84a) last updated 2016/03/10 14:54:36 (GMT -700)
  lib/ansible/modules/extras: (detached HEAD 33a557cc59) last updated 2016/03/10 14:54:36 (GMT -700)
  config file = /root/.ansible.cfg
```
##### Summary:

Some connection plugins require cmd as a list instead of a string.

Previously these plugins converted the cmd string provided to exec_command
to a list by wrapping cmd in another shell. Now they receive cmd as a list
and no longer specify their own shell.

Connection plugins which were changed:
- chroot
- docker
- jail
- libvirt_lxc
- local

These changes were made after discussions with @bcoca over PR #14813 which resulted in the decision to allow passing cmd as a list to plugins which require it.
##### Example output:

Tested as root on Ubuntu 15.10 with:

```
LIBVIRT_LXC_NOSECLABEL=true TEST_FLAGS='-l "!jail" -e ansible_user=local-test-user' make test_connection
```

Resulting a play recap of:

```
chroot-no-pipelining       : ok=9    changed=6    unreachable=0    failed=0   
chroot-pipelining          : ok=9    changed=6    unreachable=0    failed=0   
docker-no-pipelining       : ok=9    changed=6    unreachable=0    failed=0   
docker-pipelining          : ok=9    changed=6    unreachable=0    failed=0   
libvirt_lxc-no-pipelining  : ok=9    changed=6    unreachable=0    failed=0   
libvirt_lxc-pipelining     : ok=9    changed=6    unreachable=0    failed=0   
local-no-pipelining        : ok=9    changed=6    unreachable=0    failed=0   
local-pipelining           : ok=9    changed=6    unreachable=0    failed=0   
paramiko_ssh-no-pipelining : ok=9    changed=6    unreachable=0    failed=0   
paramiko_ssh-pipelining    : ok=9    changed=6    unreachable=0    failed=0   
ssh-no-pipelining          : ok=9    changed=6    unreachable=0    failed=0   
ssh-pipelining             : ok=9    changed=6    unreachable=0    failed=0 
```

Tested as root on FreeBSD 10.2 with:

```
TEST_FLAGS='-l jail' gmake test_connection
```

Resulting a play recap of:

```
jail-no-pipelining         : ok=9    changed=6    unreachable=0    failed=0   
jail-pipelining            : ok=9    changed=6    unreachable=0    failed=0   
```

### DIFF
--- a/lib/ansible/plugins/action/__init__.py
+++ b/lib/ansible/plugins/action/__init__.py
@@ -611,10 +611,10 @@ class ActionBase(with_metaclass(ABCMeta, object)):
             display.debug("_low_level_execute_command(): using become for this command")
             cmd = self._play_context.make_become_cmd(cmd, executable=executable)
 
-        if self._connection.allow_executable:
-            if executable is None:
-                executable = self._play_context.executable
-            cmd = executable + ' -c ' + pipes.quote(cmd)
+        if executable is None:
+            executable = self._play_context.executable
+
+        cmd = self._connection._shell.create_cmd_list(executable, cmd)
 
         display.debug("_low_level_execute_command(): executing: %s" % (cmd,))
         rc, stdout, stderr = self._connection.exec_command(cmd, in_data=in_data, sudoable=sudoable)
--- a/lib/ansible/plugins/connection/winrm.py
+++ b/lib/ansible/plugins/connection/winrm.py
@@ -64,7 +64,6 @@ class Connection(ConnectionBase):
     transport = 'winrm'
     module_implementation_preferences = ('.ps1', '')
     become_methods = []
-    allow_executable = False
 
     def __init__(self,  *args, **kwargs):
 
@@ -215,6 +214,7 @@ class Connection(ConnectionBase):
 
     def exec_command(self, cmd, in_data=None, sudoable=True):
         super(Connection, self).exec_command(cmd, in_data=in_data, sudoable=sudoable)
+        cmd = self._shell.join_cmd_list(cmd)
         cmd_parts = shlex.split(to_bytes(cmd), posix=False)
         cmd_parts = map(to_unicode, cmd_parts)
         script = None
--- a/lib/ansible/plugins/connection/__init__.py
+++ b/lib/ansible/plugins/connection/__init__.py
@@ -64,7 +64,6 @@ class ConnectionBase(with_metaclass(ABCMeta, object)):
     # as discovered by the specified file extension.  An empty string as the
     # language means any language.
     module_implementation_preferences = ('',)
-    allow_executable = True
 
     def __init__(self, play_context, new_stdin, *args, **kwargs):
         # All these hasattrs allow subclasses to override these parameters
@@ -159,7 +158,7 @@ class ConnectionBase(with_metaclass(ABCMeta, object)):
     def exec_command(self, cmd, in_data=None, sudoable=True):
         """Run a command on the remote host.
 
-        :arg cmd: byte string containing the command
+        :arg cmd: list of byte strings containing the command
         :kwarg in_data: If set, this data is passed to the command's stdin.
             This is used to implement pipelining.  Currently not all
             connection plugins implement pipelining.
--- a/lib/ansible/plugins/connection/accelerate.py
+++ b/lib/ansible/plugins/connection/accelerate.py
@@ -30,6 +30,7 @@ from ansible.errors import AnsibleError, AnsibleFileNotFound, AnsibleConnectionF
 from ansible.parsing.utils.jsonify import jsonify
 from ansible.plugins.connection import ConnectionBase
 from ansible.utils.encrypt import key_for_hostname, keyczar_encrypt, keyczar_decrypt
+from ansible.utils.unicode import to_bytes
 
 try:
     from __main__ import display
@@ -179,7 +180,6 @@ class Connection(ConnectionBase):
         data = dict(
             mode='command',
             cmd=cmd,
-            executable=C.DEFAULT_EXECUTABLE,
         )
         data = jsonify(data)
         data = keyczar_encrypt(self.key, data)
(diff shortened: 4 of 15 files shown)


mattclay:
I split out the unicode fixes and tests into PR #14875 in case this takes a while to get merged. I'll rebase this PR if those fixes are merged first.

mattclay:
Several of the fixes that were included in the original version of this PR have been merged separately, so I've rebased and updated the description to reflect the changes.
- The fixes for issue #10450 in PR #14910 were merged.
- The unicode fixes in PR #14875 were merged.
- The additional connection plugin tests in #14875 were merged.

This leaves only the actual cmd as list changes remaining in this PR.

mattclay:
I've rebased again to keep up with other changes to the connection plugins.

mattclay:
Re-based again and re-worked this based on discussions in IRC. The major changes are:
1. No more cmd_as_list flag, all connection plugins must accept cmd as a list.
2. Shell plugins generate the cmd list from executable and cmd instead of the action.
3. Shell plugins generate the cmd string from a list instead of the connection plugins.

The change to using shell plugins for the cmd list/string handling provide a few benefits:
- No more need for allow_executable just for the winrm connection plugin.
- The action no longer needs to deal with shell arguments: executable + ' -c ' + cmd
- In most cases connection plugins don't need to know about shell quoting

In addition to the testing in the original description, I've also run the connection tests against the accelerate connection plugin. At the moment that leaves the following untested:
- funcd
- winrm
- zone

mattclay:
I'm particularly interested in feedback on the changes involving the shell plugins. If the current approach is acceptable, there are a few other places in the connection plugins (ssh, zone) which could take advantage of this. If not, I can remove the shell changes and just inline the shell quoting in each connection plugin.

abadger:
I like this approach with shell knowing how to construct the list better than what we had before.  It feels like there's one rough edge here -- It seems like shell should know what executable is rather than having to pass it in (otherwise buggy code could mismatch the executable with the shell plugin).  You'd have to check where we instantiate the shell plugin to see if that's doable or not, though.  It might require more effort than necessary for the present PR to clean that up.  If so it can be changed at a later date.

I think there might be a bug in the powershell shell plugin.  It looks like create_cmd_list() returns a string.  then join_cmd_list looks like it might have expected a list of one element.  And then join_cmd_list returns whatever it was given instead of making sure it was a string.  This might just be more closely tied to the winrm connection plugin than the other shells and other connections but I thought I should mention it in case that wasn't quite what it was supposed to do.

mattclay:
@abadger: Good catch on the powershell plugin. I now have a working winrm test setup, so I've fixed the issue and verified it works with the win_ping module.

mattclay:
@abadger: I agree it would be better to have the shell know what executable is, rather than passing it in. However, I'm not sure how we should handle the case where the raw action overrides executable.

mattclay:
Rebased to resolve merge conflicts. I'll squash when finished.

lib/ansible/plugins/connection/winrm.py
```diff
@@ -215,6 +214,7 @@ def _connect(self):
 
     def exec_command(self, cmd, in_data=None, sudoable=True):
         super(Connection, self).exec_command(cmd, in_data=in_data, sudoable=sudoable)
+        cmd = self._shell.join_cmd_list(cmd)
         cmd_parts = shlex.split(to_bytes(cmd), posix=False)
         cmd_parts = map(to_unicode, cmd_parts)
```
nitzmahone, line 219: Just at a glance, shouldn't we be able to to_unicode the list directly instead of joining + split?

mattclay, line 219: The added join here is a bit misleading. It's there to maintain consistency with other the plugins. We always pass cmd as a list to exec_command now, but in the case of powershell that list always contains a single item -- the cmd that was previously passed as a string.

Take a look at the changes to [powershell.py](https://github.com/ansible/ansible/pull/14845/files#diff-9b6915232549a7bcd00b8b8f142b7232) and I think you'll see what I mean.

lib/ansible/plugins/action/__init__.py
```diff
@@ -611,10 +611,10 @@ def _low_level_execute_command(self, cmd, sudoable=True, in_data=None, executabl
             display.debug("_low_level_execute_command(): using become for this command")
             cmd = self._play_context.make_become_cmd(cmd, executable=executable)
 
-        if self._connection.allow_executable:
-            if executable is None:
-                executable = self._play_context.executable
-            cmd = executable + ' -c ' + pipes.quote(cmd)
+        if executable is None:
```
bcoca, line 614: this breaks `raw` and `script` which should not have a shell unless they pass it, it also breaks winrm which requires never having 'executable' prepended

mattclay:
Closing this as there are better ways to solve this that I may look at in the future.

PR #14851, opened 2016-03-07
Add Homebrew installation option

robynlarsen:


bcoca:
Sorry, but we don't recommend using homebrew to install ansible and have removed it from our docs to minimize its use. 

As such I'm going to pass on this request.

robynlarsen:
@bcoca is there a reason homebrew isn't recommended?

bcoca:
yes,  do to the homebrew package mangling ansible internals, creating a slew of user issues.

After trying to fix the issues with the packages we got nowhere, so we 'unlisted' homebrew as a way to minimize the number of issues from users (worked!).

robynlarsen:
Thanks @bcoca

Issue #14860, opened 2016-03-08
Proposal: Controlling when diff output should be produced


dagwieers:
##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
core

##### SUMMARY
Control when a diff is shown on a per-task level.

##### SOLUTIONS

Posibilities:
1. **Create a new special control `diff_when` using a condition**
   
   If there would be a need to control it independently from `changed_when`, an extra `diff_when` could be created that could have a different condition to control when a diff should be displayed.
2. **Create a new special control `show_diff` or `hide_diff` using a boolean only**
   
   Or we could decide that a special flag is sufficient that one has to explicitly flag as false or true in order to hide the diff output.

##### USE-CASE
One use-case is in fact what's already implemented in the puppet module. Since the diff may be leaking sensitive information (ansible/ansible-modules-extras#1187), they have added the opion `show_diff`, which at some point defaulted to `False`. If this was controllable from the task-level, individual modules wouldn't need their own switch.

##### NOTES
We may also want to look at providing flags to the diff routine. There is an open feature request (#14674) to ignore whitespace changes in the diff output. Since this depends on the file type/content this would make most sense to have on the task-level.

bcoca:
even simpler `diff: no|yes` flag

boedah:
This would pair nicely with [`no_log`](http://docs.ansible.com/ansible/faq.html#how-do-i-keep-secret-data-in-my-playbook).

My use case is that some templated files contain sensitive information that I would not want to be shown on the console even with `diff` mode on.

MrMMorris:
would love to see this. I use the citadel lookup plugin for secrets, but they are exposed during playbook runs.

This was a useful feature in Chef, called `sensitive`: https://docs.chef.io/resource_common.html#properties

sirkubax:
Any expected delivery time known yet? :)

ansibot:
@dagwieers Greetings! Thanks for taking the time to open this issue. In order for the community to handle your issue effectively, we need a bit more information. 

Here are the items we could not find in your description:
- component name

Please set the description of this issue with this template:
https://raw.githubusercontent.com/ansible/ansible/devel/.github/ISSUE_TEMPLATE.md

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

dagwieers:
@sirkubax Good news: https://github.com/ansible/ansible/pull/28581

bcoca:
[...]

PR #14864, opened 2016-03-08
Only show diff when the task actually induced a change

dagwieers:
##### Issue Type:

- Feature Pull Request 
##### Summary:

This implements solution 1 in the proposal #14860.

It only shows the diff if the task induced a change, which means that if the changed_when control overrides the task, no diff will be displayed.

See #14860 for a rationale and the use-case.


bcoca:
i'm not sure about this one, i might not want to reflect 'changed' but still see what would happen when people implement their own 'check mode' in scripts and other commands.

jimi-c:
@bcoca check mode should still return `changed = true` when appropriate, so that shouldn't impact the above. I agree this should be the way diffs are displayed, because if there isn't a change there shouldn't be a diff.

PR #14872, opened 2016-03-09
page rds instances

### LINKED ISSUE #14861: ec2.py return a maximum of 100 rds instances from AWS
##### Issue Type:

- Bug Report
##### Ansible Version:

```
ansible 2.0.1.0
```
##### Ansible Configuration:

None
##### Environment:

Debian
##### Summary:

When using the script `ec2.py` to pull information from AWS, your list of RDS instances would be limited to 100 results only.
This is due to module RDS in boto:
http://boto.cloudhackers.com/en/latest/ref/rds.html#module-boto.rds
Method name: `get_all_dbinstances`
##### Steps To Reproduce:

```
ec2.py --list | grep rds
```

Inside the above script, you'll find the following:

```
    def get_rds_instances_by_region(self, region):
        ''' Makes an AWS API call to the list of RDS instances in a particular
        region '''

        try:
            conn = self.connect_to_aws(rds, region)
            if conn:
                instances = conn.get_all_dbinstances()
                for instance in instances:
                    self.add_rds_instance(instance, region)
```

##### Expected Results:

Receive a list of all RDS instances present on AWS
##### Actual Results:

Limitation of 100 RDS instances

bcoca:
##### Issue Type:
- Bugfix Pull Request
##### Ansible Version:

2.0.1
##### Summary:

Attempt to page through all results of rds instances
hopefully fixes #14861

### DIFF
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -498,9 +498,14 @@ class Ec2Inventory(object):
         try:
             conn = self.connect_to_aws(rds, region)
             if conn:
-                instances = conn.get_all_dbinstances()
-                for instance in instances:
-                    self.add_rds_instance(instance, region)
+                marker = None
+                while True:
+                    instances = conn.get_all_dbinstances(marker=marker)
+                    marker = instances.marker
+                    for instance in instances:
+                        self.add_rds_instance(instance, region)
+                    if not marker:
+                        break
         except boto.exception.BotoServerError as e:
             error = e.reason
 


contrib/inventory/ec2.py
```diff
@@ -491,9 +491,14 @@ def get_rds_instances_by_region(self, region):
         try:
             conn = self.connect_to_aws(rds, region)
             if conn:
-                instances = conn.get_all_dbinstances()
-                for instance in instances:
-                    self.add_rds_instance(instance, region)
+                marker = None
+                while True:
+                    instances = conn.get_all_dbinstances(marker=marker)
+                    marker = instances.next_marker
```
Sodki: Should be `marker = instances.marker`.

bcoca: @Sodki , from docs it is supposed to be 'next_marker' as marker returns what i passed in, but this might be the issue https://github.com/boto/boto/issues/3109

as for marker, i even looked at the code:

`:param marker: The marker provided by a previous request.`

Sodki: @bcoca I see no reference to `next_marker` in the [official documentation](http://boto.cloudhackers.com/en/latest/ref/rds.html), only `marker`.

Your code's logic is correct: you do a first request, then you get a marker, then you do another request with the marker you obtained previously. But `next_marker` is always empty, while `marker` gives you the correct information. I've tested it and it works.

I think that boto/boto#3109 is invalid because we do get a marker in the `marker` attribute of the returned object.

bcoca: The other boto APIs use next_marker and the code docs themselves seem ambiguous so ... will fix to use marker in this case but I suspect it might change in the future as they align the APIs.

Sodki: Thanks, I think it's a sound middle ground.

PR #14882, opened 2016-03-09
Detect empty cache files in openstack inventory

emonty:
There are cases where it makes more sense to zero out a cache file as a
form of invalidation instead of removing it. Detect those approrpriately


sivel:
LGTM, simple enough of a change.

PR #14890, opened 2016-03-10
Check for `_data_source` and `_line_number` properties on `task._ds` before accessing them

joshuaconner:
##### Issue Type:
- Bugfix Pull Request
##### Ansible Version:

```
ansible 2.0.0.2
  config file =
```

(but present on `devel` also)
##### Summary:

I'm attempting to use ansible programmatically and am passing in a non-default Callback Module in order to get a higher verbosity level. In `task.py`, it sees that the verbosity level is higher and tries to get the file path and line number for each task, but since I'm passing in tasks via the API they don't (and IMO shouldn't?) have these properties. 

This PR adds more check to ensure `task._ds` has `_data_source` and `_line_number` properties before attempting to access them.
##### Example output:

```
 [WARNING]: Failure when attempting to use callback plugin
(<ansible.plugins.callback.default.CallbackModule object at 0x10ae2c050>): 'dict' object
has no attribute '_data_source'
```

Note that the problem isn't actually in the CallbackModule here – it's the task that's throwing, but it's getting caught and logged as a callback plugin error in `task_queue_manager.py`.

### DIFF
--- a/lib/ansible/playbook/task.py
+++ b/lib/ansible/playbook/task.py
@@ -97,7 +97,7 @@ class Task(Base, Conditional, Taggable, Become):
     def get_path(self):
         ''' return the absolute path of the task with its line number '''
 
-        if hasattr(self, '_ds'):
+        if hasattr(self, '_ds') and hasattr(self._ds, '_data_source') and hasattr(self._ds, '_line_number'):
             return "%s:%s" % (self._ds._data_source, self._ds._line_number)
 
     def get_name(self):


bcoca:
i disagree that this is a bug, also you should still point out a data source and maybe not line number but task number would be relevant to errors.

joshuaconner:
Having some meta-info with tasks sounds totally reasonable to me.

As it stands, changing the verbosity of the example [http://docs.ansible.com/ansible/developing_api.html#python-api-2-0](here) would cause this warning to be printed also, and I'd like to help others who would like to use Ansible programmatically have a better time here. Would you be open to a docs PR? What can I do to help make this better?

bcoca:
yes, though there are others already writing up some docs, i would wait a bit till those are posted (or deemed vaporware).

joshuaconner:
great, thanks! i'll close this for now.

PR #14901, opened 2016-03-10
Change ec2 instance filtering to allow for AND

rmasters:
##### Issue Type:
- Feature Pull Request
##### Ansible Version:

```
ansible 2.0.0.2
  config file = /Users/ross/.ansible.cfg
```
##### Summary:

This modifies the ec2 inventory script to hand-off the instance-filter combination to AWS's API, as described under ["Combining Search Filters"](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Filtering.html).

Currently a request is made for each filter and the results combined - this change combines the filters into a single request. These filters are combined with AND for differing key names and OR for similar keys.

I appreciate this is something that we would (and have) customised in our own setups, but I spent a bit of time searching for this and thought it might help save some time for others, even if it isn't merged.
##### Example output:

This allows for filters like:

```
tag:app=MyApp,tag:env=stage,tag:role=web,tag:role=job
becomes (tag:app==MyApp AND tag:env==stage AND (tag:role==web OR tag:role=job))
```

This will change the result of one of the examples:

```
instance_filters = instance-type=t1.micro,tag:env=staging
before: (instance-type==t1.micro OR tag:env==staging)
after: (instance-type==t1.micro AND tag:env==staging)
```

### DIFF
--- a/contrib/inventory/ec2.ini
+++ b/contrib/inventory/ec2.ini
@@ -131,7 +131,11 @@ group_by_elasticache_replication_group = True
 # inventory. For the full list of possible filters, please read the EC2 API
 # docs: http://docs.aws.amazon.com/AWSEC2/latest/APIReference/ApiReference-query-DescribeInstances.html#query-DescribeInstances-filters
 # Filters are key/value pairs separated by '=', to list multiple filters use
-# a list separated by commas. See examples below.
+# a list separated by commas.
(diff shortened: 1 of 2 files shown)


aioue:
4-5 PRs open with the same functionality.

mkrizek:

PR #13443, opened 2015-12-05
Add a way to restrict gathered facts in Ansible.

Yannig:
Needed by the following feature idea: https://github.com/ansible/ansible/issues/12088

PR #14902, opened 2016-03-10
Three cleanups to facts.py

abadger:
- Use a passed in modules parameter rather than global modules.  This is needed for ziploader
- Remove unneeded **init**()
- Remove uneeded multiple inheritance from a base class


abadger:
Merging these changes in with the work on #13443

PR #14903, opened 2016-03-10
test_test_infra fixes

nitzmahone:
##### Issue Type:
- Feature Pull Request
##### Ansible Version:

```
ansible 2.1.0 (test_test_infra_fixes a8fe0a2042) last updated 2016/03/10 11:31:39 (GMT -700)
  lib/ansible/modules/core: (detached HEAD 816bd42b07) last updated 2016/03/09 15:27:12 (GMT -700)
  lib/ansible/modules/extras: (detached HEAD e0be11da08) last updated 2016/03/09 15:27:12 (GMT -700)
  config file =
```
##### Summary:

Ensure test_test_infra is run before any target
target only localhost
ignore arg overrides
##### Example output:

```

```

### DIFF
--- a/test/integration/Makefile
+++ b/test/integration/Makefile
@@ -26,9 +26,9 @@ EUID := $(shell id -u -r)
 all: setup test_test_infra parsing test_var_precedence unicode test_templating_settings environment non_destructive destructive includes blocks pull check_mode test_hash test_handlers test_group_by test_vault test_tags test_lookup_paths no_log test_connection
 
 test_test_infra:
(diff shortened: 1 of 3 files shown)


bcoca:
i still think this should NOT be a dependency on setup

nitzmahone:
@jimi-c @abadger ? My contention is that our significant reliance on the fail and assert modules working for tests (and the way they failed yesterday) needs to be asserted before _any_ other integration test runs. This is a way to do that- I'm open to other suggestions, but yesterday's failure should _never_ be allowed to happen again.

PR #14915, opened 2016-03-11
HTTPError can also function as a non-exceptional file-like return value

job-so:
##### Issue Type:

- Bugfix Pull Request
##### Ansible Version:

```
ansible 2.1.0 (devel 5946fe8a6c) last updated 2016/03/11 07:09:57 (GMT +200)
  lib/ansible/modules/core: (detached HEAD c86a0ef84a) last updated 2016/03/11 07:03:51 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD 33a557cc59) last updated 2016/03/11 07:03:54 (GMT +200)
  configured module search path = /root/ansible.CaaS/library
```
##### Summary:

ansible/lib/ansible/module_utils/urls.py
fetch_url

Fix : Return file content on HTTPError exception.

[Quote from https://docs.python.org/2/library/urllib2.html]
exception urllib2.HTTPError

```
Though being an exception (a subclass of URLError), an HTTPError can also function as a non-exceptional file-like return value (the same thing that urlopen() returns). This is useful when handling exotic HTTP errors, such as requests for authentication.
```

[/Quote]
##### Example output:

```
```

### DIFF
--- a/lib/ansible/module_utils/urls.py
+++ b/lib/ansible/module_utils/urls.py
@@ -849,7 +849,11 @@ def fetch_url(module, url, data=None, headers=None, method=None,
     except (ConnectionError, ValueError), e:
         module.fail_json(msg=str(e))
     except urllib2.HTTPError, e:
-        info.update(dict(msg=str(e), status=e.code, response=e.read(), **e.info()))
+        try:
+            body = e.read()
+        except AttributeError:
+            body = ''
+        info.update(dict(msg=str(e), status=e.code, body=body, **e.info()))
     except urllib2.URLError, e:
         code = int(getattr(e, 'code', -1))
         info.update(dict(msg="Request failed: %s" % str(e), status=code))
--- a/lib/ansible/module_utils/urls.py
+++ b/lib/ansible/module_utils/urls.py
@@ -849,6 +849,7 @@ def fetch_url(module, url, data=None, headers=None, method=None,
     except (ConnectionError, ValueError), e:
         module.fail_json(msg=str(e))
     except urllib2.HTTPError, e:
+        r = e #useful when handling exotic HTTP errors, such as requests for authentication
         info.update(dict(msg=str(e), status=e.code, **e.info()))
     except urllib2.URLError, e:
         code = int(getattr(e, 'code', -1))
--- a/lib/ansible/module_utils/urls.py
+++ b/lib/ansible/module_utils/urls.py
@@ -849,7 +849,7 @@ def fetch_url(module, url, data=None, headers=None, method=None,
     except (ConnectionError, ValueError), e:
         module.fail_json(msg=str(e))
     except urllib2.HTTPError, e:
-        info.update(dict(msg=str(e), status=e.code, response=e, **e.info()))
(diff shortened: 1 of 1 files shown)


abadger:
For your case, perhaps utilizing open_url() gives you the flexibility you need and would be a better way to do things than to add this to fetch_url().  I worry that preexisting code might be doing something like this:

```
r, info = fetch_url([..])
if r is None:
    module.fail_json(msg=info['msg'])
```

that kind of code would be broken for any HTTPError that doesn't need to be processed by the calling code.

mgruener:
I would also like to see this implemented. Some REST APIs (for example Cloudflares) return actual meaningfull error details in the response of an HTTP error. Using open_url could be an option, but that would mean that for example modules needing the response returned with an HTTP error would need to reimplement lots of fetch_urls code.

Code that checks for an empty response to see if an HTTP error happened does it wrong anyway I think. To check for an HTTP error, the info object should be used.

job-so:
@abadger : Thanks for pointing that side effect. Indeed, i checked some existing modules, and there are codes like that...
I aggree with @mgruener, reimplementing a fetch module each time we need to get the web page content for an error is not efficient.
So my new proposition is to keep `r` to none, but adding the response in `info`.
`info.update(dict(msg=str(e), status=e.code, response=e.read(), **e.info()))`

mgruener:
@job-so Your solution just adds the response body but the response contains more than just that. You might also need the response headers. I have to deal with a situation where I have to access the Location header returned with a HTTP 409 status (https://tools.ietf.org/html/draft-ietf-acme-acme-02#section-6.3). 

And just doing `resp.read()` can throw an AttributeError exception if the response body is empty.

While talking casually about this issue to @resmo he suggested that a new parameter to fetch_url might be a way to keep backwards compatibility:

```
def fetch_url(module, url, data=None, headers=None, method=None,
              use_proxy=True, force=False, last_mod_time=None, timeout=10, response_on_error=False):
...
    except urllib2.HTTPError, e:
        info.update(dict(msg=str(e), status=e.code, **e.info()))
        if response_on_error:
            r = e
...
```

sivel:
> You might also need the response headers

I am pretty sure that `**e.info()` actually includes the headers and such.  I added it when adding the RedirectHandler and migrating the `uri` module to use `fetch_url`

mgruener:
Thanks @sivel for the pointer. **e.info() does indeed include the headers and, according to the urllib2 documentation, all other meta data of the response. In this case adding the response body to `info` should work as long as the `AttributeError` is caught in case of an empty body. @job-so could you please update the code to something like this:

```
try:
    body = e.read()
except AttributeError:
    body = ''
info.update(dict(msg=str(e), status=e.code, body=body, **e.info()))
```

I suggest using `body` as the name for the new field as it would really just contain the body and not the whole response.

If this gets accepted it should still be marked as a workaround or something like this as this is only necessary for compatibility reasons. I am convinced the proper way to do this would be to return the whole response object like @job-so suggested initially, if it wouldn't risk breaking a lot of modules.

job-so:
Done, thanks.

jimi-c:
Merged, thanks!

PR #14928, opened 2016-03-11
Merge properties instead of reading only the first cfg file found

pgmillon:
Use case :

I have a personal ansible.cfg where I configure some optims (ssh_args = -o ControlPersist=30m and pipelining = True); there is an ansible.cfg in my repo that configures some structural patterns (roles_path=public/roles:private/roles).

For such setup, I would have to copy my personal configuration on the cfg of the project or create an environment variable with all the values for the cfg files.

If settings were overridden respecting the order of precedence that is describe in the doc, I can have settings splitted between files and override only what I need.

Replaces #12523 with rebased on devel, with tests and doc updated.

### DIFF
--- a/docsite/rst/intro_configuration.rst
+++ b/docsite/rst/intro_configuration.rst
@@ -10,6 +10,15 @@ for most users, but there may be reasons you would want to change them.
 
 Changes can be made and used in a configuration file which will be processed in the following order::
 
+    * /etc/ansible/ansible.cfg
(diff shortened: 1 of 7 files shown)


pgmillon:
I updated the tests but I fail to mock the open call in ConfigParser with Python3.
Help welcome, that's the only thing missing IMHO to merge that.

dharmabumstead:
Closing due to inactivity.

PR #14930, opened 2016-03-11
Modified files to use single quotes rather than double for file path.…

mholiv:
##### Issue Type:

- Bugfix Pull Request
##### Ansible Version:

ansible 2.0.1.0
  config file = 
##### Summary:

Modified files to use single quotes rather than double for file path. Powershell does not process $ variables in strings that are single quoted. Powershell DOES process $ variables that are in double quoted strings. 

Using single quotes enables ansible to  handle file paths that contain folders that start with $. (i.e. C:/Users/$admin/...)

### DIFF
--- a/lib/ansible/plugins/connection/winrm.py
+++ b/lib/ansible/plugins/connection/winrm.py
@@ -274,7 +274,7 @@ class Connection(ConnectionBase):
 
         script_template = u'''
             begin {{
-                $path = "{0}"
+                $path = '{0}'
 
                 $DebugPreference = "Continue"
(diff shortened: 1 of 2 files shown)


bcoca:
cc @nitzmahone  @jhawkesworth @trondhindenes

nitzmahone:
Passes integration tests using a "special" user that starts with a $ (:rage:). Suspect there will be other issues to be solved (user module and others, perhaps), but this is a clean fix to get the core connection stuff working. Thanks!

nitzmahone:
cherry-picked to stable-2.0

PR #14931, opened 2016-03-11
skipping unused modules to avoid false positive

progfou:
##### Issue Type:
- Bugfix Pull Request: #14938
##### Ansible Version:

```
ansible 1.9.4
```
##### Summary:

Some OS preload the 'kvm' module even if it's not actually used.
Skipping unused modules allows to avoid false positive.
Note that the third column contains the module usage count.


bcoca:
what bug does this fix? loaded modules, even if unused, are useful information

progfou:
Sorry for that. I created the bug report #14938.

bcoca:
instead of ignoring the module being loaded we could just check if it in use, in any case I think it is still correctly marked as a 'kvm host' as the module's presence makes this possible.

jimi-c:
As the original issue was closed, we'll also close this.

Thanks!

PR #14936, opened 2016-03-12
add ansible_executable inventory var

bcoca:
##### Issue Type:
- Feature Pull Request
##### Ansible Version:

```
2.1
```
##### Summary:

Add `ansible_executable` inventory var to control the 'remote shell' used per group/host and override the `executable` setting in `ansible.cfg`

### DIFF
--- a/docsite/rst/intro_inventory.rst
+++ b/docsite/rst/intro_inventory.rst
@@ -231,8 +231,7 @@ SSH connection::
     ansible_ssh_extra_args
       This setting is always appended to the default ssh command line.
     ansible_ssh_pipelining
-      Determines whether or not to use SSH pipelining. This can override the
-      ``pipelining`` setting in ``ansible.cfg``.
+      Determines whether or not to use SSH pipelining. This can override the ``pipelining`` setting in ``ansible.cfg``.
 
 Privilege escalation (see :doc:`Ansible Privilege Escalation<become>` for further details)::
 
@@ -260,6 +259,14 @@ Remote host environment parameters::
     ansible_*_interpreter
       Works for anything such as ruby or perl and works just like ansible_python_interpreter.
       This replaces shebang of modules which will run on that host.
+
+.. versionadded:: 2.1
+
+::
+
+    ansible_executable
+      This sets the shell the ansible controller will use on the target machine, overrides ``executable`` in ``ansible.cfg`` which defaults to '/bin/sh'.
+      You should really only change it if is not possible to use '/bin/sh' (i.e. it is not in the list of allowed shells for your users).
 
 Examples from a host file::
 
--- a/test/units/plugins/action/test_action.py
+++ b/test/units/plugins/action/test_action.py
@@ -617,8 +617,8 @@ class TestActionBase(unittest.TestCase):
         play_context.make_become_cmd.assert_not_called()
 
         play_context.remote_user = 'apo'
-        action_base._low_level_execute_command('ECHO', sudoable=True)
-        play_context.make_become_cmd.assert_called_once_with("ECHO", executable='/bin/sh')
+        action_base._low_level_execute_command('ECHO', sudoable=True, executable='/bin/csh')
+        play_context.make_become_cmd.assert_called_once_with("ECHO", executable='/bin/csh')
 
         play_context.make_become_cmd.reset_mock()
 
@@ -627,6 +627,6 @@ class TestActionBase(unittest.TestCase):
         try:
             play_context.remote_user = 'root'
             action_base._low_level_execute_command('ECHO SAME', sudoable=True)
-            play_context.make_become_cmd.assert_called_once_with("ECHO SAME", executable='/bin/sh')
+            play_context.make_become_cmd.assert_called_once_with("ECHO SAME", executable=None)
         finally:
             C.BECOME_ALLOW_SAME_USER = become_allow_same_user
--- a/docsite/rst/intro_configuration.rst
+++ b/docsite/rst/intro_configuration.rst
@@ -289,6 +289,8 @@ This indicates the command to use to spawn a shell under a sudo environment.  Us
 
     executable = /bin/bash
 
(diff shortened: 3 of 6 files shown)


docsite/rst/intro_inventory.rst
```diff
@@ -231,8 +231,7 @@ SSH connection::
     ansible_ssh_extra_args
       This setting is always appended to the default ssh command line.
     ansible_ssh_pipelining
-      Determines whether or not to use SSH pipelining. This can override the
-      ``pipelining`` setting in ``ansible.cfg``.
+      Determines whether or not to use SSH pipelining. This can override the ``pipelining`` setting in ``ansible.cfg``.
```
jpic, line 234: Shouldn't this line be word-wrapped like the others ?

Your editor probably has a feature, in vim it's `gqq`

test/units/plugins/action/test_action.py
```diff
@@ -617,8 +617,8 @@ def test_action_base_sudo_only_if_user_differs(self):
         play_context.make_become_cmd.assert_not_called()
 
         play_context.remote_user = 'apo'
-        action_base._low_level_execute_command('ECHO', sudoable=True)
-        play_context.make_become_cmd.assert_called_once_with("ECHO", executable='/bin/sh')
+        action_base._low_level_execute_command('ECHO', sudoable=True, executable='/bin/csh')
+        play_context.make_become_cmd.assert_called_once_with("ECHO", executable='/bin/csh')
```
jpic: Note that we test the override here, but not the default anymore.

test/units/plugins/action/test_action.py
```diff
@@ -627,6 +627,6 @@ def test_action_base_sudo_only_if_user_differs(self):
         try:
             play_context.remote_user = 'root'
             action_base._low_level_execute_command('ECHO SAME', sudoable=True)
-            play_context.make_become_cmd.assert_called_once_with("ECHO SAME", executable='/bin/sh')
+            play_context.make_become_cmd.assert_called_once_with("ECHO SAME", executable=None)
```
jpic, line 630: Isn't the default /bin/sh anymore ?

What happens when make_become_cmd is executed with executable=None ?

bcoca, line 630: defaults to self.executable, which gets set from inventory or if all else fails the old DEFAULT_EXECUTABLE, which is `/bin/sh` by default

bcoca, line 630: the reason i left as `None` in the test is that it is called with `None` but does not result in it (diff test needed for that)

jpic, line 630: I understand, pretty cool, I guess that's the kind of things that makes ansible work all the time, thanks a heap B)

docsite/rst/intro_inventory.rst
bcoca, line 234: it does, but i purposefully avoid the 80 col nonsense since most monitors exceeded that limit in 1992.

abadger, line 234: Unfortunately, you need to word wrap this and your other entries here as well.  The reason is that we'the wrong markup for these areas.  We're using `::` which means a code block or preformatted text.  So long lines are not going to wrap in the html display either.

We shold probably be using definition lists for all of these instead of code blocks.  As a short example that would look like:

```
:ansible_ssh_extra_args:
       This setting is always appended to the default ssh command line.
:ansible_ssh_pipelining: Determines whether or not to use SSH pipelining. This can override the ``pipelining`` setting in ``ansible.cfg``.
```

( The definition can start on the same line as the definition term or the next line.  If you have line breaks (I set mine at 160 chars in a compromise between you and sanity ;-) subsequent lines just need to be indented.

I'll leave it up to you whether you want to add line breaks or reformat the lists to be definition lists instead.

bcoca, line 234: i'm fine with change to definition lists, but would make that a subsequent PR and not part of this one

docsite/rst/intro_inventory.rst
```diff
@@ -261,6 +260,14 @@ Remote host environment parameters::
...
       This replaces shebang of modules which will run on that host.
 
+.. versionadded:: 2.1
+
+::
+
+    ansible_executable
+      This sets the shell the ansible controller will use on the target machine, overrides ``executable`` in ``ansible.cfg`` which defaults to '/bin/sh'.
+      You should really only change it if is not possible to use '/bin/sh' (i.e. it is not in the list of allowed shells for your users).
+
```
abadger, line 270: I don't think allowed shells matters because we're invoking it as a command, not as a login shell.  So this probably should be if /bin/sh is not available or not present on the machine.

abadger, line 270: We probably also want to say something about ansible_shell_type needing to agree with this setting.  maybe in the ansible_shell_type documentation (since autodetection means that you might not need to set shell_type when you set executable but you always need to set executable if you set shell_type).

bcoca, line 270: that is already there

abadger, line 270:
```
-       The shell type of the target system. You should not use this setting unless you have set the 'executable' to a non sh compatible shell.
+       The shell type of the target system. You should not use this setting unless you have set the 'ansible_executable' inventory variable to a non sh compatible shell.
```

And then for here we do need to change the list of allowed shells to something like installed on the system.

abadger:
Hmm... why is travis failing....

abadger:
I also realized as I was reading the documentation that ansible_executable isn't a great choice for an external name.  "Support: What's ansible_executable set to?" "Newbie: Uhhh... /usr/bin/ansible I think?"

Maybe use ansible_shell_executable as the inventory varname so people can tell it can relate both to exutable in ansible.cfg and ansible_shell_type in inventory vars.

abadger:
Note:  I don't care if it stays executable internally... just think the external name should change.

abadger:
Merged and remaining changes made in a second  commit.

PR #14942, opened 2016-03-12
Flexible dependency specification

javiplx:
##### Issue Type:
- Feature Pull Request
##### Summary:

Although this code works, it's actually closer to a proposal. It implements two similar methods based on python's pkg_resources to allow specify role versions in a manner similar to what requirements.txt file does with pip. This means that `geerlingguy.php,<2` will install the greatest 1.x release of role.

### DIFF
--- a/lib/ansible/galaxy/role.py
+++ b/lib/ansible/galaxy/role.py
@@ -222,6 +222,8 @@ class GalaxyRole(object):
                     if not pkgs.specs :
                         pkgs, = parse_requirements("%s==%s" % (self.name, self.version))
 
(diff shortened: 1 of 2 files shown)


bcoca:
Hi!

Unfortunately, we are going to decline this PR because we have pushed this kind of functionality into the new 'collections' content management feature.

However, we're absolutely always up for discussion.  
Since this is a really busy project, we don't always see comments on closed tickets, but want to encourageopen dialog.  

You can stop by the development list, and we'd be glad to talk about it - and we might even be persuaded otherwise!

PR #14967, opened 2016-03-14
Facts subset and cleanup

abadger:
##### Issue Type:
- Feature Pull Request
##### Summary:

This builds on @Yannig's PR #13443 to implement a way to restrict fact gathering.  After discussion between Yannig, @bcoca and myself, the changes needed to also include a fix for FieldAttributes to transofmr comma separated strings into lists.  Further discussion with bcoca lead to merging the ability to exclude facter and ohai into the same configuration value/play parameter.  Documentation, code, and tests updated accordingly.  There will need to be an ansible-module-utils-core update to go along with this.  I'll submit that next.


abadger:
@Yannig, @bcoca If you'll be so kind as to review this and https://github.com/ansible/ansible-modules-core/pull/3247 to see if it's suitable that would be great!

nitzmahone:
fails from ad-hoc (eg ```ansible localhost -m setup```)- gather_subset should default to all...

nitzmahone:
Oh yeah- :+1:

abadger:
Latest commit makes the integration tests assert on failure.

Yannig:
@abadger: well done :)
The way to exclude facts using '!' is a good idea.
Using ohai or facter as a subset option is also a good idea.
So, you can go ahead and merge this !

abadger:
Thanks  @nitzmahone and @Yannig  for reviewing!  Thanks @Yannig for the first 90% of the work to get this feature started :-)

Yannig:
@abadger You're welcome :)

PR #14973, opened 2016-03-15
Allow timeout to be configurable

ogenstad:
##### Issue Type:
- Feature Pull Request
##### Ansible Version:

```
ansible 2.0.1.0
```
##### Summary:

Some times devices takes a long time to process a command. For example when you save the configuration it could complete within a few seconds some times, while other times it might take 15 seconds. This causes the commands to fail as the timeout is hardcoded to 10 seconds.


ogenstad:
The tests failed, but I plead not guilty..

Qalthos:
This looks good to me. Unfortunately I modified a few things so this won't merge, but it should be a simple rebase for you to fix.

Sorry for the inconvenience, and hopefully we can get this moving soon.

ogenstad:
The conflicts should be resolved now.

Qalthos:
Thanks for that, could you just squash the two commits so that we don't have a merge conflict sitting in the commit history, and then I think we might be good to go.

ogenstad:
Ah sorry. It should look better now.

PR #14976, opened 2016-03-15
use __mro__ for plugin loading when we search for its base class.

xiaket:
##### Issue Type:
- Feature Pull Request
##### Ansible Version:

2.1.0
##### Summary:

This would relax the constraint a little bit, allowing directly subclassing existing plugins.
##### Example output:

I tried to sublcass a strategy plugin(linear), and wrote the following code:

```
class StrategyModule(Linear):

    def run(self, iterator, play_context):
        print("running our own strategy here.")

        return Linear.run(self, iterator, play_context)
```

and got the following traceback:

```
  File "run_ansible.py", line 39, in <module>
    result = tqm.run(play)
  File "/usr/local/lib/python2.7/dist-packages/ansible/executor/task_queue_manager.py", line 218, in run
    raise AnsibleError("Invalid play strategy specified: %s" % new_play.strategy, obj=play._ds)
ansible.errors.AnsibleError: ERROR! Invalid play strategy specified: ownstrategy.
```

After some digging, I found that after we've found a specified plugin, the PluginLoader will do a check to see if the obj has the right base class. This is done by:

```
if self.base_class and self.base_class not in [base.__name__ for base in obj.__class__.__bases__]:
   return None
```

which, forbid direct subclass of the plugin.

This commit changed this behavior.

Signed-off-by: 夏恺(Xia Kai) xiaket@gmail.com

### DIFF
--- a/lib/ansible/plugins/__init__.py
+++ b/lib/ansible/plugins/__init__.py
@@ -329,8 +329,13 @@ class PluginLoader:
             obj = getattr(self._module_cache[path], self.class_name)
         else:
             obj = getattr(self._module_cache[path], self.class_name)(*args, **kwargs)
-            if self.base_class and self.base_class not in [base.__name__ for base in obj.__class__.__mro__]:
-                return None
+            if self.base_class:
+                # The import path is hardcoded and should be the right place,
+                # so we are not expecting an ImportError.
+                module = __import__(self.package, fromlist=[self.base_class])
+                # Check whether this obj has the required base class.
+                if not issubclass(obj.__class__, getattr(module, self.base_class, None)):
+                    return None
 
         return obj
 
--- a/lib/ansible/plugins/__init__.py
+++ b/lib/ansible/plugins/__init__.py
@@ -329,7 +329,7 @@ class PluginLoader:
             obj = getattr(self._module_cache[path], self.class_name)
         else:
             obj = getattr(self._module_cache[path], self.class_name)(*args, **kwargs)
-            if self.base_class and self.base_class not in [base.__name__ for base in obj.__class__.__bases__]:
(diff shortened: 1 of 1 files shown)


xiaket:
There is a way to walk around this in my case, just inherit both StrategyModule from linear.py and StrategyBase from `__init__.py`, but this fix is clearly not optimal.

jpic:
lgtm

abadger:
If we're going to allow this (and I can't think of why we wouldn't... but want to check with jimi-c/bcoca about whether there's any corener cases I'm missing) shouldn't we just switch to using issubclass()?

bcoca:
i would actually argue that looking at base class is wrong (or not totally right), looking at the 'current class' should be enough. The base classes are provided as a convenience, not as a mandate.

abadger:
We discussed on IRC and agreed that for now we're willing to relax this to any ancestry.  We'll revisit the idea of making base classes non-mandatory at some point in the future when people are interested in rewriting PluginLoader. 

So this PR just needs to be switched to use `issubclass()` instead of comparing to `__mro__` and it can be merged.

xiaket:
Thank you all for the comment.

I'm more than willing to revise this PR, but it is to be noted that the `self.base_class` here is a string, if we use `issubclass()`, we'd have to import the base class first. Is that really necessary?

abadger:
Yeah, I think that's fine. We can load on demand if we think it's a source
of slowness but I'm willing to bet it doesn't add anything to startup
time.  (We already have to import the `__init__` s that the base
classes live in in order to load any of the plugins of that type)

xiaket:
Sorry for the lateness, busy weekend here.

abadger:
@xiaket Thanks for the PR!  No worries about taking a weekend to do other things ;-)

PR #14978, opened 2016-03-15
New "dense" callback plugin

dagwieers:
##### Issue Type:

- Feature Pull Request
##### Ansible Version:

v2
##### Summary:

The goal for the "dense" output is to only show changes and failures on-screen (the Unix-way).
However, since we still want to have a sense of progress, we use terminal capabilities to display progress.
##### Design goals
- Only relevant stuff on-screen
  - How far are we ? (during run, last line)
  - What issues occurred
  - What changes occurred
  - Diff output (in diff-mode)
- If verbosity increases, act as default output
  So that users can easily switch to default for troubleshooting
- Rewrite the output during processing                                                                                               
  - We use the cursor to indicate where in the task we are.
    Output after the prompt is the output of the previous task.
  - If we would clear the line at the start of a task, there would often
    be no information at all, so we leave it until it gets updated
- Use the same color-conventions of Ansible
- Ensure the verbose output (-v) is also dense.
  Remove information that is not essential (eg. timestamps, status)
##### Installation

You have to copy `dense.py` into your callback plugins directory (e.g. _lib/ansible/plugins/callback/_) and then configure your `ansible.cfg` as follows:

``` ini
[defaults]
stdout_callback = dense
```

### DIFF
--- a/lib/ansible/plugins/callback/dense.py
+++ b/lib/ansible/plugins/callback/dense.py
@@ -19,7 +19,6 @@
 from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
-from ansible import constants as C
 from ansible.plugins.callback.default import CallbackModule as CallbackModule_default
 from ansible.utils.color import colorize, hostcolor
 from collections import OrderedDict
@@ -43,25 +42,41 @@ import sys
 #  + If verbosity increases, act as default output
 #    So that users can easily switch to default for troubleshooting
 #
-#  + Leave previous task output on screen
-#    - If we would clear the line at the start of a task, there would often
-#      be no information at all
-#
+#  + Rewrite the output during processing
 #    - We use the cursor to indicate where in the task we are.
-#      Output after the prompt is the output of the previous task
+#      Output after the prompt is the output of the previous task.
+#    - If we would clear the line at the start of a task, there would often
+#      be no information at all, so we leave it until it gets updated
 #
 #  + Use the same color-conventions of Ansible
+#
+#  + Ensure the verbose output (-v) is also dense.
+#    Remove information that is not essential (eg. timestamps, status)
 
 
 # TODO:
 #
-#  + Ensure all other output is properly displayed
 #  + Properly test for terminal capabilities, and fall back to default
 #  + Modify Ansible mechanism so we don't need to use sys.stdout directly
-#  + Remove items from result to compact the json output when using -v
-#  + Make colored output nicer
 #  + Find an elegant solution for line wrapping
-#  + Support notification handler
+#  + Support notification handlers
+#  + Better support for item-loops (in -v mode)
+
+# When using -vv or higher, simply do the default action
+
+# FIXME: Importing constants as C simply does not work, beats me :-/
+#from ansible import constants as C
+class C:
+    COLOR_HIGHLIGHT   = 'white'
+    COLOR_VERBOSE     = 'blue'
+    COLOR_WARN        = 'bright purple'
+    COLOR_ERROR       = 'red'
+    COLOR_DEBUG       = 'dark gray'
+    COLOR_DEPRECATE   = 'purple'
+    COLOR_SKIP        = 'cyan'
+    COLOR_UNREACHABLE = 'bright red'
+    COLOR_OK          = 'green'
+    COLOR_CHANGED     = 'yellow'
 
 
 # Taken from Dstat
@@ -115,16 +130,17 @@ class ansi:
 
 
 colors = dict(
-    ok=ansi.darkgreen,
-    changed=ansi.darkyellow,
-    skipped=ansi.darkcyan,
-    failed=ansi.darkred,
-    unreachable=ansi.redbg+ansi.white
+    ok = ansi.darkgreen,
+    changed = ansi.darkyellow,
+    skipped = ansi.darkcyan,
+    ignored = ansi.redbg + ansi.cyan,
+    failed = ansi.redbg + ansi.darkred,
+    unreachable = ansi.red,
 )
 
 states = ( 'skipped', 'ok', 'changed', 'failed', 'unreachable' )
 
-class CallbackModule(CallbackModule_default):
+class CallbackModule_dense(CallbackModule_default):
 
     '''
     This is the dense callback interface, where screen estate is still valued.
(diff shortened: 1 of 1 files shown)


towolf:
You are not handling `v2_playbook_item_on_ok`. This will lead to changed results not being printed when a task is looped using `with_items` or similar.

(Same for `v2_playbook_item_on_failed`)

I just added this [in my PR](https://github.com/towolf/ansible/commit/e51c234dde592cf77729ecbe752d82cafb7f236e) for the related `actionable` callback plugin.

Your plugin sounds like actionable with more bells and whistles. Gonna try it out now.

towolf:
I'm also getting a warning:

`[WARNING]: Failure when attempting to use callback plugin (<ansible.plugins.callback.dense.CallbackModule object at 0x7fd290435bd0>): 'CallbackModule' object has no attribute
'_display'`

And --diff doesn't work.

dagwieers:
Thanks for the feedback. I know it has a lot of issues right now, that's why it is tagged Work In Progress. To ensure errors are properly printed I think it requires modifying the callback/display mechanism.

dagwieers:
@towolf Thanks for the pointer to _actionable_, I have reused `super_ref` in order to make _dense_ use verbosity. Now normal operation is very dense, '-v' will show changes/failures and everything _higher_ will work as default. Feel free to test again and report back ;-)

The warning you reported is now fixed as well.

dagwieers:
@towolf With the latest commit I think it is now functional and almost feature complete.

There is one issue where line-wrapping causes issues when your host-list is longer than the terminal-width. (And this is confusing at best) One possibility is to rotate the host-list so it never wraps. But it requires a `v2_playbook_on_play_end()` hook to print the final line (if there are changes/failures), which doesn't seem to exist.

Feedback welcome.

towolf:
Just checked it out. It's pretty cool, but seems to need some fixes still.

With --check it deletes the third task line, when the shell prompt comes up:

```
towolf@zentaur:~/src/local/ansible-diff$ cat playbook.yml 
---
- name: reproduce --diff bug
  hosts: localhost

  tasks:
    - name: template non-item
      template: src=/tmp/bar dest=/tmp/baz.foo
    - name: template itemized
      template: src=/tmp/{{ item }} dest=/tmp/{{ item }}.foo
      with_items:
        -  bar
        -  baz
    - name: template itemized
      template: src=/tmp/{{ item.0 }} dest=/tmp/{{ item.1 }}.foo
      with_together:
        -  [foo, baz]
        -  [bar, quux]
towolf@zentaur:~/src/local/ansible-diff$ ansible-playbook  -i hosts playbook.yml 
PLAY 1: REPRODUCE --DIFF BUG
task 1: localhost
task 2: localhost
task 3: localhost
towolf@zentaur:~/src/local/ansible-diff$ ansible-playbook  -i hosts playbook.yml --check
PLAY 1: REPRODUCE --DIFF BUG
task 1: localhost
task 2: localhost
towolf@zentaur:~/src/local/ansible-diff$ 
```

With --diff it needs a newline somewhere:

```
towolf@zentaur:~/src/local/ansible-diff$ ansible-playbook  -i hosts playbook.yml --diff
PLAY 1: REPRODUCE --DIFF BUG
task 1: localhost--- before: /tmp/baz.foo
+++ after: dynamically generated
@@ -1 +1 @@
-foo bar
+foo


task 2: localhost--- before: /tmp/bar.foo
+++ after: dynamically generated
@@ -1 +1 @@
-foo bar
+foo

task 2: localhost--- before: /tmp/baz.foo
+++ after: dynamically generated
@@ -1 +1 @@
-foo
+foo bar

task 2: localhost--- before: /tmp/bar.foo
+++ after: dynamically generated
@@ -1 +1 @@
-foo bar
+foo

--- before: /tmp/baz.foo
+++ after: dynamically generated
@@ -1 +1 @@
-foo
+foo bar


task 3: localhost--- before: /tmp/bar.foo
+++ after: dynamically generated
@@ -1 +1 @@
-foo
+foo bar

task 3: localhost--- before: /tmp/bar.foo
+++ after: dynamically generated
@@ -1 +1 @@
-foo
+foo bar
```

dagwieers:
@towolf Thanks for the feedback.

Currently the plugin mechanism doesn't have any exit-hooks. So the closing new-line in check-mode is hard to produce correctly. A similar issue with diff-mode. Since I cannot predict when diff-output is going to be produced, I cannot provide it with the needed newline.

So the plugin mechanism may need to be adapted if this is considered useful.

jhawkesworth:
You might be able to use class destructor since there is no  on_play_exit event.

dagwieers:
@jhawkesworth That's an interesting idea.

Another problem is the fact that Ansible also outputs other stuff like warnings, prompts or error output. And none of that is exposed to the callback plugin. So we cannot "theme" it, which means if it spurs while we are updating the host-list (in normal mode) the error starts wherever the cursor was.

The solution here is to have hooks for all output.

Despite all that, the plugin is functional for everyday use (at least I am doing that :-))

jhawkesworth:
Yeah I appreciate it only gets you so far, you don't have fine control over when the destructor runs - I used it in my windows domain callback plugin here: https://github.com/jhawkesworth/ansible/blob/active_directory_callback/contrib/callback/active_directory_callback.py

hooks for all output (and an on_play_ended event) would be cool though.

DanyC97:
is this work still going to be merged to master? is been tag as 2.2 milestone ... 
what is it blocking ?

dagwieers:
I have to be honest that I did not test it with Ansible v2.2 yet. But I will look at it tomorrow.

vazhnov:
In Ansible 2.2.2, with fresh `dense.py`:
* user should set `deprecation_warnings=False` in `ansible.cfg` (if tasks has something deprecated), otherwise output will be corrupted.
* There are only task numbers in output, no task names:
```
ansible-playbook -i inventory/testing role_user_access.yml --limit=test-01.example.com
PLAY 1: ALL
task 14: test-01.example.com
```

tima:
This is a merged/closed ticket and is highly unlikely to get the attention of core engineering team. You should file a new issue instead. Thanks.

ssbarnea:
If anyone is interested in comparing the different stdout callbacks have a look at https://github.com/ssbarnea/ansible-output -- my favourite one is "debug" even if its name is confusing because is not really so verbose.

I think this bug does not affect ansible 2.3.

PR #14985, opened 2016-03-16
Never save modules with world readable permissions.

abadger:
##### Issue Type:
- Bugfix Pull Request
##### Ansible Version:

```
ansible 2.1.0 (module-permissions-fix 2523c8d467) last updated 2016/03/15 20:35:56 (GMT -700)
  lib/ansible/modules/core: (devel a98cd86f88) last updated 2016/03/15 17:45:18 (GMT -700)
  lib/ansible/modules/extras: (devel f9b96b9a8a) last updated 2016/03/15 17:44:59 (GMT -700)
```
##### Summary:

When using become to an unprivileged user, ansible saves modules files as world readable on the remote machine.  This was done because the sudo'd user was unable to read the modules written by the connection user (as was done with ssh connection's put_file() -- which uses scp and sftp).  This change sends the module over stdin to dd running under the become command.  That allows dd to write the file as the sudo'd user, not as the ssh connection user and therefore we're able to save the module file with restrictive permissions. 

Also documents the current and previous behaviour.

### DIFF
--- a/docsite/rst/become.rst
+++ b/docsite/rst/become.rst
@@ -85,7 +85,7 @@ on how it works.  Users should be aware of these to avoid surprises.
 Becoming an Unprivileged User
 =============================
 
-Ansible has a limitation with regards to becoming an
+Ansible 2.0.x and below has a limitation with regards to becoming an
 unprivileged user that can be a security risk if users are not aware of it.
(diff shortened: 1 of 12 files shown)


abadger:
Closing and creating a new PR with a different implementation

PR #14988, opened 2016-03-16
VMware self signed certs

kamsz:
That's just a small fix for VMware self-signed certificates. Tested on Python 2.7.11 and Ubuntu 15.10/14.04.

### DIFF
--- a/lib/ansible/module_utils/vmware.py
+++ b/lib/ansible/module_utils/vmware.py
@@ -133,7 +133,7 @@ def connect_to_api(module, disconnect_atexit=True):
     except vim.fault.InvalidLogin, invalid_login:
         module.fail_json(msg=invalid_login.msg, apierror=str(invalid_login))
     except requests.ConnectionError, connection_error:
-        if '[SSL: CERTIFICATE_VERIFY_FAILED]' in str(connection_error) and not validate_certs:
+        if 'certificate verify failed' in str(connection_error) and not validate_certs:
             context = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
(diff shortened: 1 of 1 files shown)


lib/ansible/module_utils/vmware.py
```diff
@@ -133,7 +133,7 @@ def connect_to_api(module, disconnect_atexit=True):
     except vim.fault.InvalidLogin, invalid_login:
         module.fail_json(msg=invalid_login.msg, apierror=str(invalid_login))
     except requests.ConnectionError, connection_error:
-        if '[SSL: CERTIFICATE_VERIFY_FAILED]' in str(connection_error) and not validate_certs:
+        if 'certificate verify failed' in str(connection_error) and not validate_certs:
```
bcoca, line 136: I'm going to guess this string has changed across versions of the product, I would try both just to keep it backwards compatible

kamsz, line 136: I'll verify in next few days

jctanner:
@kamsz what version of vcenter are you using? The current code works for me against 5 and 6.

lib/ansible/module_utils/vmware.py
nerzhul: @kamsz did you verified ? I think it's not needed anymore, my certificates are invalid at work and this is working properly

dagwieers:
Ok, this PR has so many issues now and the original reporter does not seem to care (anymore?).
I would propose to close it, and let it be reopened if it really is needed.

jctanner:
closing due to submitter timeout.

PR #14993, opened 2016-03-16
Fixes ovirt inventory to only override credentials when keys exist

karmab:
##### Issue Type:

- Bugfix Pull Request
##### Ansible Version:

ansible 2.1.0 (devel 86080fbaa9) last updated 2016/03/15 17:52:46 (GMT +200)
##### Summary:

Fixes ovirt inventory to only override credentials when keys exist as environment variables
instead of current behaviour when they dont get populated.

### DIFF
--- a/contrib/inventory/ovirt.py
+++ b/contrib/inventory/ovirt.py
@@ -172,9 +172,9 @@ class OVirtInventory(object):
 
         # If the appropriate environment variables are set, they override
         # other configuration; process those into our args and kwargs.
-        kwargs['url'] = os.environ.get('OVIRT_URL')
-        kwargs['username'] = os.environ.get('OVIRT_EMAIL')
-        kwargs['password'] = os.environ.get('OVIRT_PASS')
+	 kwargs['url'] = os.environ.get('OVIRT_URL', kwargs['url'])
+	 kwargs['username'] = next(val for val in [os.environ.get('OVIRT_EMAIL'), os.environ.get('OVIRT_USERNAME'), kwargs['username']] if val is not None)
+	 kwargs['password'] = next(val for val in [os.environ.get('OVIRT_PASS'), os.environ.get('OVIRT_PASSWORD'), kwargs['password']] if val is not None)
 
         # Retrieve and return the ovirt driver.
         return API(insecure=True, **kwargs)


contrib/inventory/ovirt.py
```diff
@@ -172,9 +172,12 @@ def get_ovirt_driver():
...
         # other configuration; process those into our args and kwargs.
-        kwargs['url'] = os.environ.get('OVIRT_URL')
-        kwargs['username'] = os.environ.get('OVIRT_EMAIL')
-        kwargs['password'] = os.environ.get('OVIRT_PASS')
+        if 'OVIRT_URL' in os.environ.keys():
+            kwargs['url'] = os.environ.get('OVIRT_URL')
+        if 'OVIRT_USERNAME' in os.environ.keys():
+            kwargs['username'] = os.environ.get('OVIRT_USERNAME')
+        if 'OVIRT_PASSWORD' in os.environ.keys():
+            kwargs['password'] = os.environ.get('OVIRT_PASSWORD')
```
jpic: Why not leverage .setdefault here ?

```
         kwargs.setdefault('url', os.environ.get('OVIRT_URL'))
         kwargs.setdefault('username', os.environ.get('OVIRT_USERNAME'))
         kwargs.setdefault('password', os.environ.get('OVIRT_PASSWORD'))
```

￼

amenonsen:
I understand the description of this PR, and it sounds sensible.

However, I don't understand the code:

```
          # If the appropriate environment variables are set, they override
          # other configuration; process those into our args and kwargs.
 -         kwargs['url'] = os.environ.get('OVIRT_URL')
 -         kwargs['username'] = os.environ.get('OVIRT_EMAIL')
 -         kwargs['password'] = os.environ.get('OVIRT_PASS')
 +         kwargs.setdefault('url', os.environ.get('OVIRT_URL'))
 +         kwargs.setdefault('username', os.environ.get('OVIRT_USERNAME'))
 +         kwargs.setdefault('password', os.environ.get('OVIRT_PASSWORD'))
```

What we want, per the comment above, is something like this: "If OVIRT_URL is set, set kwargs['url'] to it". Right? Is that what this clever `.setdefault()` stuff is actually doing? My impression from a quick bit of searching is that it's doing "If kwargs['url'] is not set, set it to OVIRT_URL", but I could be wrong.

In any case, please squash this down to one commit and get rid of the "X" merge commit.

karmab:
You are right, order needs to be reversed if we want env variables to have
precedence.
I will resquash

contrib/inventory/ovirt.py
```diff
@@ -172,9 +172,9 @@ def get_ovirt_driver():
 
         # If the appropriate environment variables are set, they override
         # other configuration; process those into our args and kwargs.
-        kwargs['url'] = os.environ.get('OVIRT_URL')
-        kwargs['username'] = os.environ.get('OVIRT_EMAIL')
-        kwargs['password'] = os.environ.get('OVIRT_PASS')
+         kwargs.setdefault('url', os.environ.get('OVIRT_URL'))
+         kwargs.setdefault('username', os.environ.get('OVIRT_USERNAME'))
+         kwargs.setdefault('password', os.environ.get('OVIRT_PASSWORD'))
```
jpic: Looks like you've got an extra space prefixed here !

contrib/inventory/ovirt.py
jpic: Ignore my last comment, turns out it's not correct, my bad.

karmab:
resquashed and came for the cleanest syntax possible taking into account that env variables take precedence

contrib/inventory/ovirt.py
```diff
@@ -172,9 +172,9 @@ def get_ovirt_driver():
 
         # If the appropriate environment variables are set, they override
         # other configuration; process those into our args and kwargs.
-        kwargs['url'] = os.environ.get('OVIRT_URL')
-        kwargs['username'] = os.environ.get('OVIRT_EMAIL')
-        kwargs['password'] = os.environ.get('OVIRT_PASS')
+	kwargs['url'] = environ.get('OVIRT_URL') if 'OVIRT_URL' in os.environ.keys() else kwargs['url']
```
jpic: There a missing space here.

Are you sure `environ` is defined here ? I don't see it.

Perhaps this would work too, if you **do** want `OVIRT_URL=` to override `kwargs['url']`:

```
kwargs['url'] = environ.get('OVIRT_URL', kwargs['url'])
```

Example:

```
$ A= python -c 'import os; print os.environ.get("A", "X")'

```

(we get an empty string)

If you **do not** want `OVIRT_URL=` to override `kwargs['url']`, then you can use `or`:

```
kwargs['url'] = os.environ.get('OVIRT_URL') or kwargs['url']
```

Example:

```
$ A= python -c 'import os; print os.environ.get("A") or "X"'
X
```

(we get the value that's at the right of `or`)

Hope this helps

contrib/inventory/ovirt.py
```diff
@@ -172,9 +172,9 @@ def get_ovirt_driver():
 
         # If the appropriate environment variables are set, they override
         # other configuration; process those into our args and kwargs.
-        kwargs['url'] = os.environ.get('OVIRT_URL')
-        kwargs['username'] = os.environ.get('OVIRT_EMAIL')
-        kwargs['password'] = os.environ.get('OVIRT_PASS')
+	kwargs['url'] = environ.get('OVIRT_URL', kwargs['url'])
+	kwargs['username'] = environ.get('OVIRT_USERNAME', kwargs['username'])
+	kwargs['password'] = environ.get('OVIRT_PASSWORD', kwargs['password'])
```
jpic: Isn't it missing a whitespace ?

Are you sure environ is defined here ? I don't see it.

Perhaps this would work too, if you do want OVIRT_URL= to override kwargs['url']:

kwargs['url'] = environ.get('OVIRT_URL', kwargs['url'])
Example:

$ A= python -c 'import os; print os.environ.get("A", "X")'

(we get an empty string)

If you do not want OVIRT_URL= to override kwargs['url'], then you can use or:

kwargs['url'] = os.environ.get('OVIRT_URL') or kwargs['url']
Example:

$ A= python -c 'import os; print os.environ.get("A") or "X"'
X
(we get the value that's at the right of or)

Hope this helps

karmab: yeah sure, typo. my bad,
put the correct line and rebased.

amenonsen:
So I was going to merge this when I noticed that OVIRT_EMAIL/OVIRT_PASS changed to OVIRT_USERNAME/OVIRT_PASSWORD. Is that intended? If so, shouldn't you test the older names to maintain backwards compatibility? (I don't even know what ovirt is, this is just a general concern based on the diff. I don't know if there's some convention, as there are for AWS_\* variables.)

karmab:
There s no convention around those variable names , i think it s rather a
typo from pasting google compute engine.
I believe those new env variables to be more intuitive

abadger:
I agree with @amenonsen ,  We shouldn't remove the old names for no real reason.  If you want to add handling of new, more intuitive names in addition to the old ones, that's fine.

abadger:
Thanks @karmab for the PR and @amenonsen for reviewing!  merged to the devel branch for inclusion in 2.1.0

PR #14998, opened 2016-03-16
[Doc fix] Fixed ansible 2 python API example

jpic:
##### Issue Type:
- Docs Pull Request
##### Ansible Version:

```
ansible 2.0.1.0
  config file = /home/jpic/.ansible.cfg
  configured module search path = /home/jpic/ansible/library:/usr/share/ansible
```
##### Summary:

The example Python API for Ansible 2.0 crashes.
##### Example output:

```
$ python test.py
  File "test.py", line 44, in <module>
    result = tqm.run(play)
  File "/home/jpic/env/lib/python2.7/site-packages/ansible/executor/task_queue_manager.py", line 206, in run
    play_context = PlayContext(new_play, self._options, self.passwords, self._connection_lockfile.fileno())
  File "/home/jpic/env/lib/python2.7/site-packages/ansible/playbook/play_context.py", line 212, in __init__
    self.set_options(options)
  File "/home/jpic/env/lib/python2.7/site-packages/ansible/playbook/play_context.py", line 258, in set_options
    self.remote_user = options.remote_user
AttributeError: 'Options' object has no attribute 'remote_user'
```

Turns out there are a lot of default options we need and getting them from the base parser is the shortest way to do it for now ...

### DIFF
--- a/docsite/rst/developing_api.rst
+++ b/docsite/rst/developing_api.rst
@@ -41,16 +41,25 @@ In 2.0 things get a bit more complicated to start, but you end up with much more
 
     from collections import namedtuple
     from ansible.parsing.dataloader import DataLoader
+    from ansible.cli import CLI
     from ansible.vars import VariableManager
     from ansible.inventory import Inventory
     from ansible.playbook.play import Play
     from ansible.executor.task_queue_manager import TaskQueueManager
 
-    Options = namedtuple('Options', ['connection', 'module_path', 'forks', 'become', 'become_method', 'become_user', 'check'])
     # initialize needed objects
     variable_manager = VariableManager()
     loader = DataLoader()
-    options = Options(connection='local', module_path='/path/to/mymodules', forks=100, become=None, become_method=None, become_user=None, check=False)
+    parser = CLI.base_parser(
+        connect_opts=True,
+        module_opts=True,
+        fork_opts=True,
+        runas_opts=True,
+        check_opts=True,
+    )
+    # ignore sys args for the sake of this example
+    options, args = parser.parse_args([])
+    options.connection = 'local'
     passwords = dict(vault_pass='secret')
 
     # create inventory and pass to var manager
--- a/docsite/rst/developing_api.rst
+++ b/docsite/rst/developing_api.rst
@@ -41,25 +41,16 @@ In 2.0 things get a bit more complicated to start, but you end up with much more
 
     from collections import namedtuple
     from ansible.parsing.dataloader import DataLoader
-    from ansible.cli import CLI
     from ansible.vars import VariableManager
     from ansible.inventory import Inventory
     from ansible.playbook.play import Play
+    from ansible.playbook.play_context import PlayContext
     from ansible.executor.task_queue_manager import TaskQueueManager
 
     # initialize needed objects
     variable_manager = VariableManager()
     loader = DataLoader()
-    parser = CLI.base_parser(
-        connect_opts=True,
-        module_opts=True,
-        fork_opts=True,
-        runas_opts=True,
-        check_opts=True,
-    )
-    # ignore sys args for the sake of this example
-    options, args = parser.parse_args([])
-    options.connection = 'local'
+    options = PlayContext.options_factory(connection='local')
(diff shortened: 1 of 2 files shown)


amenonsen:
@jpic Seems `Options` a few lines above is unused now? And I'd prefer to see `parser = CLI.base_parser(…); options = parser.parse_args(…)` the way that things in lib/ansible/cli actually do.

jpic:
Fixed, thanks !

bcoca:
so the example is for use from python directly, not for CLIs as most people will use it from a webservice and not have arguments on the command line.

to create a new CLI, just use an existing one as an example.

jpic:
Thanks @bcoca, the only problem with the example is that it doesn't work, instead it crashes (see traceback in the issue description) because there are too many undefined variables (perhaps you could try it and see for yourself if you don't believe me !). The only way I've found to get the default options is through the CLI class.

Is it ok to leave the example that crashes as is ?

Otherwise, could you suggest a better fix for the example ?

Thanks a lot for your help.

amenonsen:
I agree that it's the "wrong" fix in the sense that CLI.base_parser isn't fundamentally needed, but it's the right fix in the sense that the example is broken and there's no easier way to make it work. The "really right" fix would perhaps be to update the Options namedtuple with all the fields and values that the rest of the code expects to be set, but that seems a verbose and fragile approach. Maybe the "really really right" fix would be to provide some constructor that returns a default-configuration object, and have the parser update that.

But I think we can merge this as-is as a short-term bandaid for the problem. @bcoca?

jpic:
Don't get me wrong, I completely agree that this fix is wrong !

I'm making another PR to provide a default configuration object, but it will be a bit heavy as you can imagine.

jpic:
I pushed the "better fix".

It works for me on devel _if_ I revert 4547ac7f first (see comment on [another commit](https://github.com/ansible/ansible/commit/4547ac7f#commitcomment-16743136))

What do you think ?

amenonsen:
I'm hoping there's a better way to build the default list (I don't know what it would look like, it's just speculation), and maybe options_factory should be named default_options, but otherwise this looks like a promising and quite self-contained cleanup. @bcoca?

jpic:
Thanks ! It did take a handful of tries to find a self-contained way I recon ;)

I tried to provide an Options namedtuple with defaults, but didn't find a way to update a namedtuple instance:

```
In [1]: from collections import namedtuple

In [2]: a=namedtuple('foo', ['bar'])

In [3]: b=a(bar='test')

In [4]: b.bar = 'x'
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-4-8b341975cb35> in <module>()
----> 1 b.bar = 'x'

AttributeError: can't set attribute
```

So I didn't find a way to provide defaults as a named tuple. This wouldn't work:

```
# Ansible code
def default_options():
    return Options(**defaults)

# User code:
options = default_options()
options.connection = 'local'  # raises "Can't set attribute"
```

So I had to go for options_factory(**options) instead of a default_options() method.

Isn't [PlayContext.DEFAULT_OPTIONS](https://github.com/ansible/ansible/pull/14998/commits/f769d1ae9515f872262d805c8fee0d5d214a29e3#diff-efa3a13dd7cbec87c75b359049e1c7d3R196) the default option list you were hoping for ?

In this patch, defaults are provided as a dict.

It's possible to reuse this dict in CLI's base_parser if you see fit.

Another way would be to introduce an Options class, but that would be a bit heavier a patch.

bcoca:
@jpic the current version should have been correct, but it seems some update to play_context broke it, i'm going to look into fixing that.

bcoca:
@jpic  it seems you are using an older version of the API, using it against the current version works just fine.

bcoca:
you might want this version of the document https://github.com/ansible/ansible/blob/7037f793b9a90f07e51c15b2bca6be862f159ff4/docsite/rst/developing_api.rst

jpic:
Awesome, thanks @bcoca !

So the current online docs are for the _devel_ version, not the latest stable, that's really good to know !

PR #15001, opened 2016-03-16
Update submodule refs for docs fix

abadger:
##### Issue Type:

- Docs Pull Request
##### Summary:

Ansible has long had a security caveat around becoming/sudoing to an unprivileged user.  Modules are made world readable on the remote host in this case.  Add documentation about this issue and the ways that users can avoid it.


abadger:
+1'd by jimi-c

PR #15009, opened 2016-03-16
Py3 compat fixes

bcoca:
##### Issue Type:
- Bugfix Pull Request
##### Ansible Version:

```
2.1
```
##### Summary:

several fixes, left as their own commit for easy cherrypicking

### DIFF
--- a/lib/ansible/executor/playbook_executor.py
+++ b/lib/ansible/executor/playbook_executor.py
@@ -168,8 +168,9 @@ class PlaybookExecutor:
                 # send the stats callback for this playbook
                 if self._tqm is not None:
                     if C.RETRY_FILES_ENABLED:
-                        retries = list(set(self._tqm._failed_hosts.keys() + self._tqm._unreachable_hosts.keys()))
-                        retries.sort()
+                        retries = set(self._tqm._failed_hosts.keys())
+                        retries.update(self._tqm._unreachable_hosts.keys())
+                        retries = sorted(retries)
                         if len(retries) > 0:
                             if C.RETRY_FILES_SAVE_PATH:
                                 basedir = C.shell_expand(C.RETRY_FILES_SAVE_PATH)
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -607,7 +607,8 @@ class TaskExecutor:
                 try:
                     cmd = subprocess.Popen(['ssh','-o','ControlPersist'], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
                     (out, err) = cmd.communicate()
-                    if "Bad configuration option" in err or "Usage:" in err:
+                    err = to_unicode(err)
+                    if u"Bad configuration option" in err or u"Usage:" in err:
                         conn_type = "paramiko"
                 except OSError:
                     conn_type = "paramiko"
@@ -645,7 +646,9 @@ class TaskExecutor:
             try:
                 connection._connect()
             except AnsibleConnectionFailure:
+                display.debug('connection failed, fallback to accelerate')
                 res = handler._execute_module(module_name='accelerate', module_args=accelerate_args, task_vars=variables, delete_remote_tmp=False)
+                display.debug(res)
                 connection._connect()
 
         return connection
--- a/lib/ansible/plugins/callback/logentries.py
+++ b/lib/ansible/plugins/callback/logentries.py
@@ -60,8 +60,9 @@ import socket
 import random
 import time
 import codecs
-import ConfigParser
 import uuid
+from ansible.compat.six.moves import configparser
+
 try:
     import certifi
     HAS_CERTIFI = True
@@ -212,7 +213,7 @@ class CallbackModule(CallbackBase):
                                  'Disabling the Logentries callback plugin.')
 
         config_path = os.path.abspath(os.path.dirname(__file__))
-        config = ConfigParser.ConfigParser()
+        config = configparser.ConfigParser()
         try:
             config.readfp(open(os.path.join(config_path, 'logentries.ini')))
             if config.has_option('logentries', 'api'):
--- a/lib/ansible/plugins/connection/local.py
+++ b/lib/ansible/plugins/connection/local.py
@@ -79,7 +79,7 @@ class Connection(ConnectionBase):
 
         p = subprocess.Popen(
             cmd,
-            shell=isinstance(cmd, basestring),
+            shell=isinstance(cmd, (text_type, binary_type)),
             executable=executable, #cwd=...
             stdin=subprocess.PIPE,
             stdout=subprocess.PIPE,
--- a/lib/ansible/executor/task_queue_manager.py
+++ b/lib/ansible/executor/task_queue_manager.py
@@ -34,6 +34,7 @@ from ansible.template import Templar
 from ansible.vars.hostvars import HostVars
 from ansible.plugins.callback import CallbackBase
 from ansible.utils.unicode import to_unicode
+from ansible.compat.six import string_types
 
 try:
     from __main__ import display
@@ -143,7 +144,7 @@ class TaskQueueManager:
 
         if isinstance(self._stdout_callback, CallbackBase):
             stdout_callback_loaded = True
-        elif isinstance(self._stdout_callback, basestring):
+        elif isinstance(self._stdout_callback, string_types):
             if self._stdout_callback not in callback_loader:
                 raise AnsibleError("Invalid callback for stdout specified: %s" % self._stdout_callback)
             else:


bcoca:
cc @abadger @mscherer

lib/ansible/module_utils/urls.py
```diff
@@ -82,10 +82,14 @@
 # Agreement.
 
 try:
-    import urllib2
+    from urllib2 import urlopen, Request, HTTPError, URLError, BaseHandler, HTTPPasswordMgrWithDefaultRealm, HTTPSHandler, AbstractHTTPHandler, HTTPRedirectHandler, buil_opener as url_buil_opener, install_opener as url_install_opener, HTTPBasicAuthHandler
     HAS_URLLIB2 = True
-except:
+except ImportError:
     HAS_URLLIB2 = False
```
abadger: HAS_URLLIB2 = False needs to move further down

lib/ansible/module_utils/urls.py
```diff
@@ -82,10 +82,14 @@
 # Agreement.
 
 try:
```
abadger: Add a comment to this that we may be able to use six.moves for this once we have ziploader.

lib/ansible/module_utils/urls.py
```diff
@@ -256,8 +260,11 @@ def match_hostname(cert, hostname):
 
     HAS_MATCH_HOSTNAME = True
 
+try:
```
abadger: Comment here about revisiting with ziploader and six here as well.

lib/ansible/module_utils/urls.py
```diff
@@ -313,7 +320,7 @@ class NoSSLError(SSLValidationError):
 # Some environments (Google Compute Engine's CoreOS deploys) do not compile
 # against openssl and thus do not have any HTTPS support.
 CustomHTTPSConnection = CustomHTTPSHandler = None
-if hasattr(httplib, 'HTTPSConnection') and hasattr(urllib2, 'HTTPSHandler'):
+if hasattr(httplib, 'HTTPSConnection') and 'HTTPSHandler' in sys.modules:
```
abadger: This won't work because HTTPSHandler isn't a module, it's just a class.

lib/ansible/module_utils/urls.py
```diff
@@ -622,7 +629,8 @@ def http_request(self, req):
             # close the ssl connection
             #ssl_s.unwrap()
             s.close()
-        except (ssl.SSLError, socket.error), e:
+        except (ssl.SSLError, socket.error):
+            e = sys.exc_info()[1]
```
abadger: mark this with a comment to revisit with ziploader too -- we'll be able to write some functions to make this more mnemonic and import it into any file we want.

lib/ansible/module_utils/urls.py
```diff
@@ -840,22 +848,28 @@ def fetch_url(module, url, data=None, headers=None, method=None,
         info.update(r.info())
         info['url'] = r.geturl()  # The URL goes in too, because of redirects.
         info.update(dict(msg="OK (%s bytes)" % r.headers.get('Content-Length', 'unknown'), status=200))
```
abadger: Add one ziploader comment around here so that we go through all of these and convert them over when the time comes.

lib/ansible/plugins/callback/logentries.py
```diff
@@ -60,9 +60,13 @@
 import random
 import time
 import codecs
-import ConfigParser
 import uuid
 try:
+    import ConfigParser
+except ImportError:
+    import configparser as ConfigParser
+
```
abadger: Is this all we have to do for this?  I think in py3, SafeConfigParser has taken over the name that ConfigParser used to inhabit.

lib/ansible/module_utils/urls.py
abadger: typo url_buil_opener... typo is in multiple places.

abadger: buil_opener as well

lib/ansible/module_utils/urls.py
```diff
@@ -444,12 +451,12 @@ class RedirectHandler(urllib2.HTTPRedirectHandler):
         def redirect_request(self, req, fp, code, msg, hdrs, newurl):
             handler = maybe_add_ssl_handler(newurl, validate_certs)
             if handler:
-                urllib2._opener.add_handler(handler)
+                http_opener.add_handler(handler)
```
abadger: where does http_opener come from I don't see it imported in the changes?

lib/ansible/executor/playbook_executor.py
```diff
@@ -168,7 +168,7 @@ def run(self):
                 # send the stats callback for this playbook
                 if self._tqm is not None:
                     if C.RETRY_FILES_ENABLED:
-                        retries = list(set(self._tqm._failed_hosts.keys() + self._tqm._unreachable_hosts.keys()))
+                        retries = list(set(list(self._tqm._failed_hosts.keys()) + list(self._tqm._unreachable_hosts.keys())))
```
abadger: Better to write this like this:

``` python
import itertools
[...]
list(frozenset(itertools.chain(a.keys(), b.keys())))

```

Or if you want to write it out:

``` python
retries = set(a.keys())
retries.update(b.keys())
retries = list(retries)
```

Either of those avoid instantiating two extra lists.

abadger:
Comments made.  You might also want to run pyflakes on the files and check for undefined variables.

lib/ansible/executor/playbook_executor.py
bcoca: i was going for least intrusive, but your 2nd one seems to be the least bad

lib/ansible/module_utils/urls.py
bcoca: `locals().has_key('HTTPSHandler')`?

lib/ansible/plugins/callback/logentries.py
bcoca: will change to `from ansible.compat.six.moves import configparser`

lib/ansible/module_utils/urls.py
bcoca: fixed typos, not sure about urllib2 var, it really does not work as things fail waaay before fetch_url can check for it, as it is used in class defs.

lib/ansible/module_utils/urls.py
abadger: Common python idiom would be at the top:

``` python
HTTPSHandler = None
try:
    from urllib2 import HTTPSHandler
[...]

if hasattr(httplib, 'HTTPSConnection') and HTTPSHandler:
```

lib/ansible/module_utils/urls.py
abadger: Could remove it entirely if it doesn't serve any purpose

lib/ansible/executor/playbook_executor.py
```diff
@@ -168,8 +168,9 @@ def run(self):
                 # send the stats callback for this playbook
                 if self._tqm is not None:
                     if C.RETRY_FILES_ENABLED:
-                        retries = list(set(self._tqm._failed_hosts.keys() + self._tqm._unreachable_hosts.keys()))
-                        retries.sort()
+                        retries = set(self._tqm._failed_hosts.keys())
+                        retries.update(self._tqm._unreachable_hosts.keys())
+                        retries = list(retries).sort()
```
abadger: You'll need:

``` python
retries = sorted(list(retries))
```

sort() is an inplace sort and returns None.

lib/ansible/module_utils/urls.py
```diff
@@ -814,6 +824,7 @@ def fetch_url(module, url, data=None, headers=None, method=None,
     Fetches a file from an HTTP/FTP server using urllib2.  Requires the module environment
     '''
 
+    #FIXME: not really working here as class defs fail when using it for their inheritance
```
abadger: urllib2 comes from the stdlib so my recommendation is to remove this check and the HAS_URLLIB2 attribute.

lib/ansible/plugins/callback/logentries.py
```diff
@@ -212,7 +213,7 @@ def __init__(self):
                                  'Disabling the Logentries callback plugin.')
 
         config_path = os.path.abspath(os.path.dirname(__file__))
-        config = ConfigParser.ConfigParser()
+        config = configparser.ConfigParser()
```
abadger, line 216: <nod>  What I was wondering before is that I believe in python2 this is the equivalent of RawConfigParser and i python3 it's the equivalent of SafeConfigParser.  Just wondered if there's anything else that needed to be adjusted for that change.

bcoca:
updated, now i can execute:

`ansible -m debug -a 'var=vars' localhost`

 w/o issues, anything using a connection plugin still fails though

bcoca:
cows are all messed up :-(

```
ansible-playbook play.yml 
b' __________________ \n< PLAY [localhost] >\n ------------------ \n        \\   ^__^\n         \\  (oo)\\_______\n            (__)\\       )\\/\\\n                ||----w |\n                ||     ||\n'
b' ______________ \n< TASK [debug] >\n -------------- \n        \\   ^__^\n         \\  (oo)\\_______\n            (__)\\       )\\/\\\n                ||----w |\n                ||     ||\n'
ok: [localhost] => {
    "msg": "py3 works"
}
b' ______________ \n< TASK [pause] >\n -------------- \n        \\   ^__^\n         \\  (oo)\\_______\n            (__)\\       )\\/\\\n                ||----w |\n                ||     ||\n'
[pause]
Press enter to continue:
Press 'C' to continue the play or 'A' to abort 
ok: [localhost]
b' ____________ \n< PLAY RECAP >\n ------------ \n        \\   ^__^\n         \\  (oo)\\_______\n            (__)\\       )\\/\\\n                ||----w |\n                ||     ||\n'
```

lib/ansible/plugins/connection/local.py
```diff
@@ -80,7 +80,7 @@ def exec_command(self, cmd, in_data=None, sudoable=True):
 
         p = subprocess.Popen(
             cmd,
-            shell=isinstance(cmd, basestring),
+            shell=isinstance(cmd, text_type),
```
abadger: This one is wrong... needs to account for both text strings and byte strings.  So probably have to compose it from individual constants in six.

``` python
shell=isinstance(cmd,(text_type, binary_type))
```

lib/ansible/executor/task_executor.py
```diff
@@ -605,6 +605,7 @@ def _get_connection(self, variables, templar):
                 try:
                     cmd = subprocess.Popen(['ssh','-o','ControlPersist'], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
                     (out, err) = cmd.communicate()
+                    err = to_str(err)
```
abadger: to_str will work here but probably for cleanness sake, want to use to_unicode here, like this:

``` python
err = to_unicode(err)
if u'Bad configuration option' in err or u'Usage:' in err:
```

lib/ansible/plugins/callback/logentries.py
bcoca, line 216: I just used the same one we are using for constants.py

lib/ansible/executor/task_executor.py
bcoca: tried that, it did not like it

lib/ansible/playbook/base.py
```diff
@@ -106,7 +106,8 @@ def _generic_g(prop_name, self):
         if hasattr(self, method):
             return getattr(self, method)()
 
-        value = self._attributes[prop_name]
+        value = self._attributes.get(prop_name)
```
abadger: This feels like there's something very wrong.  Was the former code throwing a KeyError?  If so, then .get() is just masking the problem (returning None instead of throwing a KeyError).  I think this is a bad change but would have to look at the error to know what the correct change is.

bcoca: yes, it was throwing keyerror on 'become_pass'

lib/ansible/executor/task_executor.py
abadger: Odd -- on python3, to_str should just be an alias for to_unicode so there shouldn't be any difference.  On python2, they'd be different but it should work because of automatic coercion of byte strings to unicode strings in python2.

abadger:
``` python
if PY3:
    to_str = to_unicode
else:
    to_str = to_bytes
```

lib/ansible/executor/playbook_executor.py
```diff
@@ -168,8 +168,9 @@ def run(self):
                 # send the stats callback for this playbook
                 if self._tqm is not None:
                     if C.RETRY_FILES_ENABLED:
-                        retries = list(set(self._tqm._failed_hosts.keys() + self._tqm._unreachable_hosts.keys()))
-                        retries.sort()
+                        retries = set(self._tqm._failed_hosts.keys())
+                        retries.update(self._tqm._unreachable_hosts.keys())
+                        retries = sorted(list(retries))
```
abadger: Just discovered, don't need the intermediate list here, either.  sorted takes an iterable so `new_list = sorted(set((5,4,3,2,1)))` is perfectly valid.

lib/ansible/module_utils/urls.py
```diff
@@ -81,11 +81,18 @@
 # agrees to be bound by the terms and conditions of this License
 # Agreement.
 
+# FIXME: try to use six.moves once we exclusively use ziploader
+HTTPSHandler = None
 try:
-    import urllib2
+    from urllib2 import urlopen, Request, HTTPError, URLError, BaseHandler, HTTPPasswordMgrWithDefaultRealm, HTTPSHandler, AbstractHTTPHandler, HTTPRedirectHandler, build_opener as url_build_opener, install_opener as url_install_opener, HTTPBasicAuthHandler, OpenerDirector
```
jpic: Would it be preferable to split the import over multiple lines for readability ? ie. 

```
from urllib2 import (
    urlopen, 
    Request, 
    HTTPError, 
    URLError, 
    BaseHandler, 
    HTTPPasswordMgrWithDefaultRealm, 
    HTTPSHandler, 
    AbstractHTTPHandler, 
    HTTPRedirectHandler, 
    build_opener as url_build_opener, 
    install_opener as url_install_opener, 
    HTTPBasicAuthHandler, 
    OpenerDirector,
)
```

lib/ansible/module_utils/urls.py
```diff
@@ -81,11 +81,18 @@
...
     HAS_URLLIB2 = True
-except:
+except ImportError:
     HAS_URLLIB2 = False
+    try: # try python3
+        from urllib.request import urlopen, Request, HTTPError, URLError, BaseHandler, HTTPPasswordMgrWithDefaultRealm, HTTPSHandler, AbstractHTTPHandler, HTTPRedirectHandler, build_opener as url_build_opener, install_opener as url_install_opener, HTTPBasicAuthHandler, OpenerDirector
```
jpic: Same here, would this be more readable ?

```
        from urllib.request import (
            urlopen, 
            Request, 
            HTTPError, 
            URLError, 
            BaseHandler, 
            HTTPPasswordMgrWithDefaultRealm, 
            HTTPSHandler, 
            AbstractHTTPHandler, 
            HTTPRedirectHandler, 
            build_opener as url_build_opener, 
            install_opener as url_install_opener,
            HTTPBasicAuthHandler,
            OpenerDirector,
        )
```

lib/ansible/module_utils/urls.py
```diff
@@ -832,30 +844,34 @@ def fetch_url(module, url, data=None, headers=None, method=None,
...
         if distribution is not None and distribution.lower() == 'redhat':
             module.fail_json(msg='%s. You can also install python-ssl from EPEL' % str(e))
         else:
             module.fail_json(msg='%s' % str(e))
-    except (ConnectionError, ValueError), e:
+    except (ConnectionError, ValueError):
         module.fail_json(msg=str(e))
-    except urllib2.HTTPError, e:
+    except HTTPError:
         info.update(dict(msg=str(e), status=e.code, **e.info()))
```
jpic: e is not defined here anymore, shouldn't we have this instead ?

```
except urllib2.HTTPError as e:
```

lib/ansible/executor/task_executor.py
jpic: Shouldn't we just use `err = six.text_type(err)` ?

lib/ansible/plugins/connection/local.py
```diff
@@ -80,7 +80,7 @@ def exec_command(self, cmd, in_data=None, sudoable=True):
 
         p = subprocess.Popen(
             cmd,
-            shell=isinstance(cmd, basestring),
+            shell=isinstance(cmd, (text_type, binary_type)),
```
jpic, line 82: Why binary_type ? Isn't text_type enough ?

lib/ansible/module_utils/urls.py
```diff
@@ -81,11 +81,18 @@
...
+    from urllib2 import urlopen, Request, HTTPError, URLError, BaseHandler, HTTPPasswordMgrWithDefaultRealm, HTTPSHandler, AbstractHTTPHandler, HTTPRedirectHandler, build_opener as url_build_opener, install_opener as url_install_opener, HTTPBasicAuthHandler, OpenerDirector
     HAS_URLLIB2 = True
-except:
+except ImportError:
     HAS_URLLIB2 = False
+    try: # try python3
+        from urllib.request import urlopen, Request, HTTPError, URLError, BaseHandler, HTTPPasswordMgrWithDefaultRealm, HTTPSHandler, AbstractHTTPHandler, HTTPRedirectHandler, build_opener as url_build_opener, install_opener as url_install_opener, HTTPBasicAuthHandler, OpenerDirector
+        HAS_URLLIB_PY3 = True
+    except:
+        HAS_URLLIB_PY3 = False
```
jpic: Perhaps this is a bit more clean:

```
if six.PY2:
    from urllib2 import (
        urlopen,
        Request,
        HTTPError,
        URLError,
        BaseHandler,
        HTTPPasswordMgrWithDefaultRealm,
        HTTPSHandler,
        AbstractHTTPHandler,
        HTTPRedirectHandler,
        build_opener as url_build_opener,
        install_opener as url_install_opener,
        HTTPBasicAuthHandler,
        OpenerDirector,
    )   
else:
    from urllib.request import (
        urlopen, 
        Request, 
        HTTPError,
        URLError,  
        BaseHandler,
        HTTPPasswordMgrWithDefaultRealm,
        HTTPSHandler, 
        AbstractHTTPHandler,
        HTTPRedirectHandler, 
        build_opener as url_build_opener,
        install_opener as url_install_opener,
        HTTPBasicAuthHandler,
        OpenerDirector,
    )   

# Then, do if six.PY2 or if six.PY3
```

lib/ansible/module_utils/urls.py
```diff
@@ -256,8 +263,11 @@ def match_hostname(cert, hostname):
 
     HAS_MATCH_HOSTNAME = True
 
+try:
+    import httplib
+except ImportError:
+    import http.client as httplib
```
jpic: Agreed, is is perhaps more explicit:

```
if six.PY2:
    import httplib
else:
    import http.client as httplib

```

lib/ansible/module_utils/urls.py
```diff
@@ -832,30 +844,34 @@ def fetch_url(module, url, data=None, headers=None, method=None,
...
-    except urllib2.URLError, e:
+    except URLError:
         code = int(getattr(e, 'code', -1))
         info.update(dict(msg="Request failed: %s" % str(e), status=code))
-    except socket.error, e:
+    except socket.error:
         info.update(dict(msg="Connection failure: %s" % str(e), status=-1))
-    except Exception, e:
+    except Exception:
         info.update(dict(msg="An unknown error occurred: %s" % str(e), status=-1))
```
jpic: ~~Same as above, where is `e` defined here ?~~

lib/ansible/executor/task_executor.py
abadger: That can throw tracebacks without a lot of control possibilities from us.  Using to_unicode and to_bytes is better.  Since we're dealing with it as text in this case, to_unicode is what we should be using.

lib/ansible/module_utils/urls.py
abadger: if I remember python scoping rules correctly, e from the inner try: except will be present here.   I thought this was a bit tricky/magical when I saw it but I have to admit that for this particular structure it is pretty convenient.

We can't use `Exception as e` style code in this file because the file also has to be usable on python2.4.  So if this does get rid of the magicalness of the inner try: except, it will have to use  `e = sys.exc_info()[1]`

lib/ansible/plugins/connection/local.py
abadger, line 82: We could remove text_type here if we wanted.... binary_type is the important one to keep.
subprocess deals in byte strings internally.  Passing in text strings can yield tracebacks in both python2 and python3 in some circumstances (more often in pyhon2 than python3 but try this: so we try always to give bytes to subprocess.  Not sure if we want to remove text_type or not... it could work for many customers who are using just ascii or have good locales set if we leave it in.  OTOH, if we take it out, it could be an early warning for us that we have to adjust our  code somewhere.

Example of subprocess failure on python3 with text type:

```
LC_ALL=C python3  
>>> import subprocess
>>> subprocess.Popen('printf \u304f\u3089\u3068\u307f', shell=True)
  File "/usr/lib64/python3.4/subprocess.py", line 859, in __init__
    restore_signals, start_new_session)
  File "/usr/lib64/python3.4/subprocess.py", line 1401, in _execute_child
    restore_signals, start_new_session, preexec_fn)
UnicodeEncodeError: 'ascii' codec can't encode characters in position 7-10: ordinal not in range(128)
```

lib/ansible/module_utils/urls.py
abadger: Even better: `from six.moves import http_client` ;-)

lib/ansible/executor/task_executor.py
jpic: I understand, theses are really nice functions you got there, thanks a heap for your reply !

lib/ansible/module_utils/urls.py
jpic: Missed that on line 857, thanks, really tight code, keep up the great work.

abadger:
-1 on the urls.py changes.  +1 on the rest.

bcoca:
why the -1?

abadger:
Mainly we should stop using the HAVE_URLLIB2 conditional.  It's been broken for, what 2 years now and no one's complained.  Making the code this ugly for something that no one uses isn't a good thing.

I also saw that you switched to importing six but we can't do that until (1) ziploader lands and (2) we bundle the subset of six that will work on python2.4 (maybe an older version?)  There might be other things that need fixing in the urls code as well but it's hard to tell from the diff because the change to accomodate HAVE_URLLIB2  obscures much of the code.

bcoca:
totallymissedthat

abadger:
Yay!  One step closer.

PR #15018, opened 2016-03-17
Updated facts.py to prevent Samba mounts from appearing in ansible_mounts

ivovangeel:
##### Issue Type:
- Bugfix Pull Request
##### Ansible Version:

```
ansible 2.1.0 (devel 921e04185a) last updated 2016/03/17 12:36:27 (GMT +200)
  lib/ansible/modules/core: (detached HEAD 345d9cbca8) last updated 2016/03/17 13:39:25 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD f47b499bb9) last updated 2016/03/17 13:39:33 (GMT +200)
```
##### Summary:

By default the facts module generates a list of filesystems, where the device starts with a '/' Samba mounts start with '//', but should probably not be included.
##### Example output:

```
        "ansible_mounts": [
            {
                "device": "/dev/sda2",
                "fstype": "ext4",
                "mount": "/",
                "options": "rw",
                "size_available": 6786015232,
                "size_total": 11490541568,
                "uuid": "6ddc8945-1679-407c-bb31-f43e28054201"
            },
            {
                "device": "//centos6/share",
                "fstype": "cifs",
                "mount": "/mnt",
                "options": "rw",
                "size_available": 7209725952,
                "size_total": 11490541568,
                "uuid": "NA"
            }
        ],
```

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -1086,7 +1086,9 @@ class LinuxHardware(Hardware):
         for line in mtab.split('\n'):
             if line.startswith('/'):
                 fields = line.rstrip('\n').split()
-                if(fields[2] != 'none'):
+                samba_filesystems = ['cifs', 'smbfs']
+                # Samba filesystems start with '//' but should not be included in mounts
+                if(fields[2] != 'none' and not fields[2] in samba_filesystems):
                     size_total, size_available = self._get_mount_size_facts(fields[1])
                     if fields[0] in uuids:
                         uuid = uuids[fields[0]]


bcoca:
I'm not sure why we would want 'less' information, other network mounts already appear here.

ivovangeel:
I was thinking of creating a new fact for remote filesystems, since NFS mounts do not show up in the ansible_mounts fact. I did not realise that other remote filesystems are already included in the mounts list.

I would like NFS mounts to appear as well, but this may not be the way to go. Any ideas ?

amenonsen:
I didn't realise NFS mounts weren't included in ansible_mounts. How come they aren't?

bcoca:
nfs mounts don't start with / which is how the setup tries to avoid proc/swap and other virtual mounts

bcoca:
now NFS mounts are show (via d43fc631dd22372407f55265982d60632bad73a2)

amenonsen:
@ivovangeel so this should probably be closed, right?

ivovangeel:
Yes, I'll try and close it myself.

PR #14947, opened 2016-03-13
Add doc on ansible_version, fix #13204

mscherer:
##### Issue Type:
- Docs Pull Request

PR #15021, opened 2016-03-17
Fix versionadded for ansible_version

jpic:
@bcoca wrote:

> ansible_version was added in 1.8, not 2.0

https://github.com/ansible/ansible/pull/14947#issuecomment-197900946


amenonsen:
@bcoca fixed this already in 84bbd2b4, so closing.

PR #15023, opened 2016-03-17
Clarify developing modules documentation

bpsuntrup:

### DIFF
--- a/docsite/rst/developing_modules.rst
+++ b/docsite/rst/developing_modules.rst
@@ -71,6 +71,9 @@ There's a useful test script in the source checkout for ansible::
     source ansible/hacking/env-setup
     chmod +x ansible/hacking/test-module
 
+For instructions on setting up ansible from source, please see
+:doc:`intro_installation`.
+
 Let's run the script you just wrote with that::
 
     ansible/hacking/test-module -m ./timetest.py


bcoca:
as mentioned in the bug ticket, this should not be needed as it is a requirement of ansible itself

bpsuntrup:
Hi bcoca,

The reason this particular addition seems necessary is because the module development page doesn't assume the user has ansible installed to begin with. If ansible's being installed were a prerequisite to the tutorial, there would be no reason to perform the Git clone.

It seems to me that if installing Jinja2 would be off topic, then so would the Git clone. They seem to me to go together. Putting them together simply clarifies the documentation, especially for new users.

Thanks,
bpsuntrup

bcoca:
There are installation instructions for ansible from git, those already state the dependencies.

bpsuntrup:
Hi bcoca,

Here is a permalink to the particular section I'm talking about: http://docs.ansible.com/ansible/developing_modules.html#testing-modules. You can see how in context, asking a user to ensure Jinja2 is installed seems to flow better, especially for a new user just trying to follow the tutorial. For such users, a hiccup in the flow could cause them to question whether the documentation is up to date, especially so with wording such as "You should see output that looks something like this: ... If you did not, you might have a typo in your module, so recheck it and try again."

Good, clear, flowing documentation is essential for the success of ansible, as it not only serves as a faithful reference, but is one of the first impressions new potential users interact with.

Thank you,
bpsuntrup

bcoca:
a link to how to use ansible from check out (instructions already in the install tutorial) seems more relevant here.

bpsuntrup:
I agree. Perhaps a link to http://docs.ansible.com/ansible/intro_installation.html#running-from-source would be best here.

bcoca:
agreed

Issue #14684, opened 2016-02-26
ansible_ssh_hostname bashes delegate_to


abadger:
##### Issue Type:
- Bug Report
##### Ansible Version:

```
ansible 2.1.0 (detached HEAD devel) last updated 2016/02/26 10:29:37 (GMT -700)
  lib/ansible/modules/core: (devel e9454fa44f) last updated 2016/02/24 08:52:37 (GMT -700)
  lib/ansible/modules/extras: (devel fade5b7936) last updated 2016/02/24 08:53:09 (GMT -700)
  config file = /srv/ansible/troubleshooting/ansible.cfg
  configured module search path = Default w/o overrides

Also affects:
ansible 2.0.1.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = /srv/web/infra/ansible/library:/usr/share/ansible

```
##### Environment:

Fedora 23 and RHEL7 hosts are affected.  Probably platform agnostic.
##### Summary:

If an inventory host has ansible_ssh_host set then that host is used instead of the host specified in delegate_to.
##### Steps To Reproduce:

```
$ cat inventory
rhel7 ansible_ssh_host=rhel7
fedora22
$ cat inventory-bug.yml

---
- hosts: rhel7
  tasks:
  - command: hostname
    delegate_to: fedora22
```
##### Expected Results:

```
$ ansible-playbook -i inventory-bug inventory-bug.yml -v         (13:21:30)
Using /srv/ansible/troubleshooting/ansible.cfg as config file

PLAY [rhel7] *******************************************************************

TASK [setup] *******************************************************************
ok: [rhel7]

TASK [command] *****************************************************************
changed: [rhel7 -> fedora22] => {"changed": true, "cmd": ["hostname"], "delta": "0:00:00.001760", "end": "2016-02-26 13:21:34.300978", "rc": 0, "start": "2016-02-26 13:21:34.299218", "stderr": "", "stdout": "fedora22.lan", "stdout_lines": ["rhel7.lan"], "warnings": []}

PLAY RECAP *********************************************************************
rhel7                      : ok=2    changed=1    unreachable=0    failed=0
```

stdout should be fedora22.lan
##### Actual Results:

```
$ ansible-playbook -i inventory-bug inventory-bug.yml -v         (13:21:30)
Using /srv/ansible/troubleshooting/ansible.cfg as config file

PLAY [rhel7] *******************************************************************

TASK [setup] *******************************************************************
ok: [rhel7]

[...]

PR #15024, opened 2016-03-17
ensure we use delegated vars on delegation

bcoca:
##### Issue Type:
- Bugfix Pull Request
##### Ansible Version:

```
2.x+
```
##### Summary:

This prevents falling back to connection vars from the inventory_hostname
when matching connection var is not in delegated host.

Probably fixes many issues listed here: https://github.com/ansible/ansible/issues/14684#issuecomment-190423485

### DIFF
--- a/lib/ansible/playbook/play_context.py
+++ b/lib/ansible/playbook/play_context.py
@@ -359,12 +359,15 @@ class PlayContext(Base):
             for variable_name in variable_names:
                 if attr in attrs_considered:
                     continue
-                if isinstance(delegated_vars, dict) and variable_name in delegated_vars:
-                    setattr(new_info, attr, delegated_vars[variable_name])
-                    attrs_considered.append(attr)
+                # if delegation task ONLY use delegated host vars, avoid delegated FOR host vars
+                if task.delegate_to is not None:
+                    if isinstance(delegated_vars, dict) and variable_name in delegated_vars:
+                        setattr(new_info, attr, delegated_vars[variable_name])
+                        attrs_considered.append(attr)
                 elif variable_name in variables:
                     setattr(new_info, attr, variables[variable_name])
                     attrs_considered.append(attr)
+                # no else, as no other vars should be considered
 
         # make sure we get port defaults if needed
         if new_info.port is None and C.DEFAULT_REMOTE_PORT is not None:


bcoca:
please confirm that #15024 resolves this issue for you

lib/ansible/playbook/play_context.py
```diff
@@ -359,9 +359,10 @@ def set_task_and_variable_override(self, task, variables, templar):
             for variable_name in variable_names:
                 if attr in attrs_considered:
                     continue
-                if isinstance(delegated_vars, dict) and variable_name in delegated_vars:
-                    setattr(new_info, attr, delegated_vars[variable_name])
-                    attrs_considered.append(attr)
+                if delegated_vars:
```
abadger: Took another look at the logic setting delegated_vars -- I think this check needs to be

```
 if task.delegate_to is not None
```

Otherwise if the delegated_to host has no vars, I think it would still end up with an empty dict on Line 318 and therefore this condition won't be triggered.

abadger:
Found a way to make the implementation better I think.  The idea that delegated_to vars should never be overridden by vars from the inventory host sounds right to me.  If @jimi-c has time to confirm that, it's great but I think it makes sense.

abadger:
+1 from me.

dflock:
Any idea which releases will this end up in? `delegate_to` is fairly broken for me in 2.0.0.2 & 2.0.1. Any chance the fixes for the `deletate_to` stuff will get backported, or will we have to wait for 2.1?

bcoca:
already added to stable-2.0 branch so any >2.0.1 release will have it, as well as 2.1 and above

aperigault:
I Installed ansible 2.1 and bug still here :/
If my delegated host not match limit delegated host_vars are not loaded...

bcoca:
@aperigault that would be an unrelated bug, if delegated_to host is in inventory host vars should be loaded independent of anything else.

deepix:
I'm on Ansible 2.1 (stable) and I'm having this problem as well.  With dynamic inventory and "delegate_to" in the playbook, it fails to connect.  Can give more details - let me know.

deepix:
I figured it out: if I have a bare host (=not belonging to any group), then I should have something like this in the (generated) json:

```
{
  "bare_hostname": [ "bare_hostname" ],
  ...
  "_meta": {
    "hostvars": {
      "bare_hostname": {
        "ansible_host": "10.217.216.195"
        ...
      },
      ...
  }
}
```

Then, I can do something like:

```
- name: ...
  ...
  delegate_to: bare_hostname
```

Otherwise it just doesn't include bare_host in the hostvars.  I confirmed with a `debug: var=hostvars`

gusai:
Hello,
ansible_user is not taken into account for synchronize -> delegate_to.
Ansible version 2.0.2.0

Snippet of task:
synchronize:
    src: "{{ vertica_remote_package_location }}"
    dest: "{{ vertica_package_location }}"
    mode: pull
  delegate_to: vertica-repo
  when: not vertica_use_apt and ansible_os_family == 'RedHat' and vertica_remote_rpm

Hosts:
[vertica]

[vertica-repo]
remotehost.net  ansible_user=userA ansible_ssh_pass=userAPW

PR #15031, opened 2016-03-17
prepare packaging for newer Debian/Ubuntu releases

evgeni:
##### Issue Type:
- Feature Pull Request
##### Summary:

Ubuntu removes support for `python-support` in 16.04 LTS (Xenial), Debian does so in 9 (Stetch). The desired migration path is to `dh-python` (https://wiki.debian.org/Python/TransitionToDHPython2).

While at it, document new `apt` command in Xenial/Stetch.
##### Warning:

This breaks compat with Debian 7 and Ubuntu 12.04 LTS. If you want to keep compat with those we'd have to come up with a bit of fallback code/deps

### DIFF
--- a/packaging/debian/README.md
+++ b/packaging/debian/README.md
@@ -4,7 +4,7 @@ Ansible Debian Package
 To create an Ansible DEB package:
 
     sudo apt-get install python-paramiko python-yaml python-jinja2 python-httplib2 python-setuptools python-six sshpass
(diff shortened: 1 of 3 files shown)


bcoca:
we probably want to keep backwards compat

evgeni:
@bcoca added some compat code, tested on 12.04 and 16.04 only :)

PR #15043, opened 2016-03-18
Use docker exec -u only when supported.

mattclay:
##### Issue Type:

Bugfix Pull Request
##### Ansible Version:

```
ansible 2.1.0 (docker-user a4182ee8c6) last updated 2016/03/18 12:58:52 (GMT -700)
  lib/ansible/modules/core: (detached HEAD 22a9a15a5b) last updated 2016/03/18 10:12:33 (GMT -700)
  lib/ansible/modules/extras: (detached HEAD 33e1d9d1cb) last updated 2016/03/18 10:12:33 (GMT -700)
  config file = 
```
##### Summary:

Commit 4ac49ed4a8535c4b4e1aab75bc0d7787be5907dd added the -u option to docker exec. There are two issues with this:
1. It requires docker 1.7 or later: [CHANGELOG.md](https://github.com/docker/docker/blob/master/CHANGELOG.md#170-2015-06-16)
2. If remote_user is None, the command will fail on any docker version unless None is a valid user.

The corrected behaviour only uses exec with -u when remote_user is provided and supported. If provided and unsupported a warning will be displayed.

As part of this fix I also removed the unused can_copy_bothways code in a separate commit, since docker cp has no support for -u in any docker version.
##### Example output:

Tested on Ubuntu 15.10 with Docker version 1.6.2, build 7c8fca2

Before the fix with no remote_user:

```
$ ansible -m raw all -i 'ubuntu-latest,' -a id -e 'ansible_connection=docker'
ubuntu-latest | FAILED | rc=2 >>
flag provided but not defined: -u
See '/usr/bin/docker exec --help'.
```

After the fix with no remote user:

```
$ ansible -m raw all -i 'ubuntu-latest,' -a id -e 'ansible_connection=docker'
ubuntu-latest | SUCCESS | rc=0 >>
uid=0(root) gid=0(root) groups=0(root)
```

Before the fix with a remote_user:

```
$ ansible -m raw all -i 'ubuntu-latest,' -a id -e 'ansible_connection=docker ansible_user=nobody'
ubuntu-latest | FAILED | rc=2 >>
flag provided but not defined: -u
See '/usr/bin/docker exec --help'.
```

After the fix with a remote_user:

```
$ ansible -m raw all -i 'ubuntu-latest,' -a id -e 'ansible_connection=docker ansible_user=nobody'
 [WARNING]: docker 1.6.2 does not support remote_user, using container default

ubuntu-latest | SUCCESS | rc=0 >>
uid=0(root) gid=0(root) groups=0(root)
```

Also tested on Ubuntu 15.10 using the latest Docker version 1.10.3, build 20f81dd

Before the fix with no remote_user:

```
$ ansible -m raw all -i 'ubuntu-latest,' -a id -e 'ansible_connection=docker'
ubuntu-latest | FAILED | rc=255 >>
Unable to find user None
```

After the fix with no remote user:

```
$ ansible -m raw all -i 'ubuntu-latest,' -a id -e 'ansible_connection=docker'
ubuntu-latest | SUCCESS | rc=0 >>
uid=0(root) gid=0(root) groups=0(root)
```

Before the fix with a remote_user:

```
$ ansible -m raw all -i 'ubuntu-latest,' -a id -e 'ansible_connection=docker ansible_user=nobody'
ubuntu-latest | SUCCESS | rc=0 >>
uid=65534(nobody) gid=65534(nogroup) groups=65534(nogroup)
```

After the fix with a remote_user:

```
$ ansible -m raw all -i 'ubuntu-latest,' -a id -e 'ansible_connection=docker ansible_user=nobody'
ubuntu-latest | SUCCESS | rc=0 >>
uid=65534(nobody) gid=65534(nogroup) groups=65534(nogroup)
```

### DIFF
--- a/lib/ansible/plugins/connection/docker.py
+++ b/lib/ansible/plugins/connection/docker.py
@@ -74,15 +74,31 @@ class Connection(ConnectionBase):
             if not self.docker_cmd:
                 raise AnsibleError("docker command not found in PATH")
 
-        self.can_copy_bothways = False
-
         docker_version = self._get_docker_version()
         if LooseVersion(docker_version) < LooseVersion('1.3'):
             raise AnsibleError('docker connection type requires docker 1.3 or higher')
-        # Docker cp in 1.8.0 sets the owner and group to root rather than the
-        # user that the docker container is set to use by default.
-        #if LooseVersion(docker_version) >= LooseVersion('1.8.0'):
-        #    self.can_copy_bothways = True
+
+        # The remote user we will request from docker (if supported)
+        self.remote_user = None
+        # The actual user which will execute commands in docker (if known)
+        self.actual_user = None
+
+        if self._play_context.remote_user is not None:
+            if LooseVersion(docker_version) >= LooseVersion('1.7'):
+                # Support for specifying the exec user was added in docker 1.7
+                self.remote_user = self._play_context.remote_user
+                self.actual_user = self.remote_user
+            else:
+                self.actual_user = self._get_docker_remote_user()
+
+                if self.actual_user != self._play_context.remote_user:
+                    display.warning('docker {0} does not support remote_user, using container default: {1}'
+                                    .format(docker_version, self.actual_user or '?'))
+        elif self._display.verbosity > 2:
+            # Since we're not setting the actual_user, look it up so we have it for logging later
+            # Only do this if display verbosity is high enough that we'll need the value
+            # This saves overhead from calling into docker when we don't need to
+            self.actual_user = self._get_docker_remote_user()
 
     @staticmethod
     def _sanitize_version(version):
@@ -107,12 +123,43 @@ class Connection(ConnectionBase):
 
         return self._sanitize_version(cmd_output)
 
+    def _get_docker_remote_user(self):
+        """ Get the default user configured in the docker container """
+        p = subprocess.Popen([self.docker_cmd, 'inspect', '--format', '{{.Config.User}}', self._play_context.remote_addr],
+                             stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+
+        out, err = p.communicate()
+
+        if p.returncode != 0:
+            display.warning('unable to retrieve default user from docker container: %s' % out + err)
+            return None
+
+        # The default exec user is root, unless it was changed in the Dockerfile with USER
+        return out.strip() or 'root'
+
+    def _build_exec_cmd(self, cmd):
+        """ Build the local docker exec command to run cmd on remote_host
+
+            If remote_user is available and is supported by the docker
+            version we are using, it will be provided to docker exec.
+        """
+
+        local_cmd = [self.docker_cmd, 'exec']
+
+        if self.remote_user is not None:
+            local_cmd += ['-u', self.remote_user]
+
+        # -i is needed to keep stdin open which allows pipelining to work
+        local_cmd += ['-i', self._play_context.remote_addr] + cmd
+
+        return local_cmd
+
     def _connect(self, port=None):
         """ Connect to the container. Nothing to do """
         super(Connection, self)._connect()
         if not self._connected:
             display.vvv(u"ESTABLISH DOCKER CONNECTION FOR USER: {0}".format(
-                self._play_context.remote_user, host=self._play_context.remote_addr)
+                self.actual_user or '?', host=self._play_context.remote_addr)
             )
             self._connected = True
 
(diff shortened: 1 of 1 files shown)


lib/ansible/plugins/connection/docker.py
```diff
@@ -50,6 +50,7 @@ class Connection(ConnectionBase):
 
     transport = 'docker'
     has_pipelining = True
+    remote_user = None
```
abadger: We wan this to be an instance attribute instead of a class attribute.  A class attribute will only have one value that's shared among all classes.

lib/ansible/plugins/connection/docker.py
```diff
@@ -74,15 +75,17 @@ def __init__(self, play_context, new_stdin, *args, **kwargs):
...
-        self.can_copy_bothways = False
-
         docker_version = self._get_docker_version()
         if LooseVersion(docker_version) < LooseVersion('1.3'):
             raise AnsibleError('docker connection type requires docker 1.3 or higher')
-        # Docker cp in 1.8.0 sets the owner and group to root rather than the
-        # user that the docker container is set to use by default.
-        #if LooseVersion(docker_version) >= LooseVersion('1.8.0'):
-        #    self.can_copy_bothways = True
+
```
abadger, line 80: So set `self.remote_user = None` here (in `__init__()`)  and then your current code will override it if necessary.

lib/ansible/plugins/connection/docker.py
```diff
@@ -74,15 +74,19 @@ def __init__(self, play_context, new_stdin, *args, **kwargs):
...
+        self.remote_user = None
+
+        if self._play_context.remote_user is not None:
+            if LooseVersion(docker_version) >= LooseVersion('1.7'):
+                # Support for specifying the exec user was added in docker 1.7
+                self.remote_user = self._play_context.remote_user
+            else:
+                # The default exec user is root, unless it was changed in the Dockerfile with USER
+                display.warning('docker {0} does not support remote_user, using container default'.format(docker_version))
```
abadger: So we weren't going to issue a warning if _play_context.remote_user matches with what docker is going to use anyways.  So I suppose we probably want to do something like this in the else:

``` python
p = subprocess.Popen([self.docker_cmd, 'inspect', '--format', '{{.Config.Image}}', self._play_context.remote_addr], shell=False, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
out, err = p.communicate()
if p.returncode != 0:
    display.warning() # Warning message about being unable to get user to login as from  docker + out + err
docker_remote_user = out.strip() or 'root'
if docker_remote_user != self._play_context.remote_user:
  display.warning('docker {0} does not support remote_user, using container default'.format(docker_version))
```

abadger: Ah -- I see you want to use a user name for the ESTABLISH DOCKER CONNECTION line later... we can save docker_remote_user for that purpose as well.

lib/ansible/plugins/connection/docker.py
```diff
@@ -132,6 +135,23 @@ def exec_command(self, cmd, in_data=None, sudoable=False):
         stdout, stderr = p.communicate(in_data)
         return (p.returncode, stdout, stderr)
 
+    def _build_exec_cmd(self, cmd):
```
abadger: Ansible convention has methods and functions that are called internally are above the methods that call them.

lib/ansible/plugins/connection/docker.py
mattclay: I believe we want '{{.Config.User}}' not '{{.Config.Image}}'.

lib/ansible/plugins/connection/docker.py
```diff
@@ -156,32 +176,24 @@ def put_file(self, in_path, out_path):
...
-            # only docker >= 1.8.1 can do this natively
-            args = [ self.docker_cmd, "cp", in_path, "%s:%s" % (self._play_context.remote_addr, out_path) ]
-            args = [to_bytes(i, errors='strict') for i in args]
-            p = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        out_path = pipes.quote(out_path)
+        # Older docker doesn't have native support for copying files into
+        # running containers, so we use docker exec to implement this
+        # Although docker version 1.8 and later provide support, the
+        # owner and group of the files are always set to root
+        executable = C.DEFAULT_EXECUTABLE.split()[0] if C.DEFAULT_EXECUTABLE else '/bin/sh'
```
abadger: bcoca's patch has been merged so we can use `play_context.executable` instead of C.DEFAULT_EXECUTABLE.  And the default value should be built in so no need for the final `else /bin/sh`

abadger:
Looks good.  Squash and I'll merge!

abadger:
Thanks @mattclay!

PR #15046, opened 2016-03-19
removes python requirement to script

bcoca:
##### Issue Type:
- Bugfix Pull Request
##### Ansible Version:

```
2.x
```
##### Summary:

dependency on python mistakenly added when checksum was made to use stat module


abadger:
If /bin/stat is portable across *nixes then this looks like a good change.  If it's not, maybe we only need to check whether the file exists for script's purposes(which could be done with test -e).

bcoca:
they both should be 'as portable' and part of posix

abadger:
Looks good.  +1

PR #15050, opened 2016-03-19
added '--raw' option to ansible cli, which outputs raw task results in json format

clime:
##### Issue Type:
- Feature Pull Request
##### Ansible Version:

ansible 2.1.0 (devel 2258dce262) last updated 2016/03/19 19:42:42 (GMT +200)
##### Summary:

I would like to programmatically invoke the 'ansible' command-line tool and have machine-parsable results on stdout mainly for error-checking purposes (rc code is quite essential and occassionally we also want to extract something from stdout of remotely run commands and store it to db). For that I made new --raw option, which outputs task results in json. Also I have added the 'raw' stdout callback for that purpose. We would like to use the ansible command line tool with this option in our project to programmatically execute commands on remote machines.
##### Example output:

clime@coprbox ~/ansible $ PYTHONPATH=~/ansible/lib ~/ansible/bin/ansible -r localhost -m shell -a 'echo foo'
{"changed": true, "cmd": "echo foo", "delta": "0:00:00.001782", "end": "2016-03-19 22:24:58.670540", "rc": 0, "start": "2016-03-19 22:24:58.668758", "stderr": "", "stdout": "foo", "stdout_lines": ["foo"], "warnings": []}clime@coprbox ~/ansible $


sivel:
Ansible already has built in support for a `json` stdout, however it currently only works with `ansible-playbook`, but I imagine it could be made to work fairly similarly for `ansible` as well.

As it stands, the way this is implemented, would cause a bunch of JSON data structures to be printed instead of 1 at the end.

Consolidating the output, into a standard datastructure (that of the already existing `json` callback), would allow for more consistency, and easier parsing of return data.

I'm not too in favor of adding many more command line args, and to keep it consistent with `ansible-playbook`, it would be nice to see it implemented so that the following works:

```
$ ANSIBLE_STDOUT_CALLBACK=json ANSIBLE_LOAD_CALLBACK_PLUGINS=1 ansible all -i localhost, -c local -a 'echo foo'
```

The following implementation allows for this to work:

``` diff
diff --git a/lib/ansible/cli/adhoc.py b/lib/ansible/cli/adhoc.py
index dcb65ad..773e25b 100644
--- a/lib/ansible/cli/adhoc.py
+++ b/lib/ansible/cli/adhoc.py
@@ -163,7 +163,7 @@ class AdHocCLI(CLI):
         elif self.options.one_line:
             cb = 'oneline'
         else:
-            cb = 'minimal'
+            cb = C.DEFAULT_STDOUT_CALLBACK

         run_tree=False
         if self.options.tree:
@@ -186,6 +186,8 @@ class AdHocCLI(CLI):
                 )

             result = self._tqm.run(play)
             self._tqm.send_callback('v2_playbook_on_stats', self._tqm._stats)
         finally:
             if self._tqm:
                 self._tqm.cleanup()
diff --git a/lib/ansible/constants.py b/lib/ansible/constants.py
index 365bda0..3b194df 100644
--- a/lib/ansible/constants.py
+++ b/lib/ansible/constants.py
@@ -20,6 +20,7 @@ from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type

 import os
+import sys
 from string import ascii_letters, digits

 from ansible.compat.six import string_types
@@ -28,6 +29,7 @@ from ansible.compat.six.moves import configparser
 from ansible.parsing.quoting import unquote
 from ansible.errors import AnsibleOptionsError

 # copied from utils, avoid circular reference fun :)
 def mk_boolean(value):
     if value is None:
@@ -215,7 +217,7 @@ DEFAULT_VARS_PLUGIN_PATH       = get_config(p, DEFAULTS, 'vars_plugins',       '
 DEFAULT_FILTER_PLUGIN_PATH     = get_config(p, DEFAULTS, 'filter_plugins',     'ANSIBLE_FILTER_PLUGINS', '~/.ansible/plugins/filter:/usr/share/ansible/plugins/filter', ispath=True)
 DEFAULT_TEST_PLUGIN_PATH       = get_config(p, DEFAULTS, 'test_plugins',       'ANSIBLE_TEST_PLUGINS', '~/.ansible/plugins/test:/usr/share/ansible/plugins/test', ispath=True)
 DEFAULT_STRATEGY_PLUGIN_PATH   = get_config(p, DEFAULTS, 'strategy_plugins',   'ANSIBLE_STRATEGY_PLUGINS', '~/.ansible/plugins/strategy:/usr/share/ansible/plugins/strategy', ispath=True)
-DEFAULT_STDOUT_CALLBACK        = get_config(p, DEFAULTS, 'stdout_callback',    'ANSIBLE_STDOUT_CALLBACK', 'default')
+DEFAULT_STDOUT_CALLBACK        = get_config(p, DEFAULTS, 'stdout_callback',    'ANSIBLE_STDOUT_CALLBACK', 'minimal' if os.path.basename(sys.argv[0]) == 'ansible' else 'default')
 # cache
 CACHE_PLUGIN                   = get_config(p, DEFAULTS, 'fact_caching', 'ANSIBLE_CACHE_PLUGIN', 'memory')
 CACHE_PLUGIN_CONNECTION        = get_config(p, DEFAULTS, 'fact_caching_connection', 'ANSIBLE_CACHE_PLUGIN_CONNECTION', None)
```

clime:
I like the idea of unified format with ansible-playbook. It makes sense because running an ad-hoc task should be a special case of running a one-task playbook. And the output of the json callback is better for sure (mainly it is grouped by hostname).

I have some notes to that example impl.: 
- sending of that event 'v2_playbook_on_stats' should not be conditioned by value of stdout_callback
- that sys.argv[0] condition in constants.py seems to be hackish so not sure if a new command-line option is not better in the end

sivel:
> I have some notes to that example impl.:
> 
> sending of that event 'v2_playbook_on_stats' should not be conditioned by value of stdout_callback

Agreed, and I have updated it.  `minimal` and `oneline` do not respond to that callback method.  I've updated my diff to reflect that.

> that sys.argv[0] condition in constants.py seems to be hackish so not sure if a new command-line option is not better in the end

Effectively we are already using this in https://github.com/ansible/ansible/blob/49f8376051edc40577d766492050b910217edbd2/bin/ansible#L60 for determining which CLI we are calling.  Unfortunately due to how the imports are processed in `bin/ansible` we cannot use `from __main__ import me` in `constants.py` as it hasn't been defined by the time `contants.py` is first imported.  And since we already have the capability to achieve this functionality in `ansible-playbook`, I'm not a fan of diverging and adding more command line options.

Basically if can get the `ansible` command to respect setting the `stdout` callback, then this already works.

bcoca:
i prefer to add a stdout override config like playbook already has, this is easily implemented in callbacks and adding new command line options for specific callbacks is not the direction we want to go, bad enough --tree and --oneline are already there.

clime:
```
--- a/lib/ansible/cli/adhoc.py
+++ b/lib/ansible/cli/adhoc.py
@@ -162,10 +162,8 @@ class AdHocCLI(CLI):
             cb = self.callback
         elif self.options.one_line:
             cb = 'oneline'
         else:
-            cb = 'minimal'
+            cb = C.get_config(C.p, C.DEFAULTS, 'stdout_callback', 'ANSIBLE_STDOUT_CALLBACK', 'minimal')

         run_tree=False
         if self.options.tree:
@@ -188,6 +186,7 @@ class AdHocCLI(CLI):
                 )

             result = self._tqm.run(play)
+            self._tqm.send_callback('v2_playbook_on_stats', self._tqm._stats)
         finally:
             if self._tqm:
                 self._tqm.cleanup()
```

What about this then? No sys.argv[0] in constants.py if that matters.

---

Personally, I would still go for the command-line option just because --tree and --one-line are already there (and there are no such options for ansible-playbook), so adding this one doesn't really seem like diverging from anything, rather the opposite. ANSIBLE_STDOUT_CALLBACK would be then playbook-specific (it could be put into its name). Like this, there would be nicely formatted output for humans (minimal) on one side and comprehensive machine-parsable output on the other (raw/json) available directly from ansible command-line interface.

But, yeah, I will be happy if any of these is put to work.

sivel:
I believe the goal is to create a new configuration for the default callback used by Ansible adhoc, as opposed to reusing the same one.

As far as `--raw`, i believe @bcoca and I are in agreement that we don't want additional formatting flags.

clime:
> I believe the goal is to create a new configuration for the default callback used by Ansible adhoc, as opposed to reusing the same one.

Different from ANSIBLE_STDOUT_CALLBACK? I am sorry but I don't quite understand the idea. 

Anyway, that's a different proposal from the original so you can close this one.

calfonso:
Hi!

- The recommendation is to not add an additional output formatting flag
- You can set the ANSIBLE_LOAD_CALLBACK_PLUGINS=true

PR #15054, opened 2016-03-20
Add shared functionality to return list of security group IDs from list of names

wimnat:
##### Issue Type:

- Feature Pull Request
##### Ansible Version:

2.1.0
##### Summary:

This change adds code to return a list of security group IDs when passed a list of names or combination list of names and IDs.

This functionality is already duplicated across some ec2 modules and has been requested in others (see https://github.com/ansible/ansible-modules-extras/pull/953).  Rather than continue to duplicate this code, I have added to ec2 utils.

### DIFF
--- a/lib/ansible/module_utils/ec2.py
+++ b/lib/ansible/module_utils/ec2.py
@@ -375,3 +375,66 @@ def ansible_dict_to_boto3_tag_list(tags_dict):
         tags_list.append({'Key': k, 'Value': v})
 
     return tags_list
+
+
+def get_ec2_security_group_ids_from_names(sec_group_list, ec2_connection, vpc_id=None, boto3=True):
+
+    """ Return list of security group IDs from security group names """
+
+    def get_sg_name(sg, boto3):
+
+        if boto3:
+            return sg['GroupName']
+        else:
+            return sg.name
+
+
+    def get_sg_id(sg, boto3):
+
+        if boto3:
+            return sg['GroupId']
+        else:
+            return sg.id
+
+
+    sec_group_id_list = []
+
+    if isinstance(sec_group_list, basestring):
+        sec_group_list = [sec_group_list]
+
+    # Get all security groups
+    if boto3:
+        if vpc_id:
+            filters = [
+                {
+                    'Name': 'vpc-id',
+                    'Values': [
+                        vpc_id,
+                    ]
+                }
+            ]
+            all_sec_groups = ec2_connection.describe_security_groups(Filters=filters)['SecurityGroups']
+        else:
+            all_sec_groups = ec2_connection.describe_security_groups()['SecurityGroups']
+    else:
+        if vpc_id:
+            filters = { 'vpc-id': vpc_id }
+            all_sec_groups = ec2_connection.get_all_security_groups(filters=filters)
+        else:
+            all_sec_groups = ec2_connection.get_all_security_groups()
+
+    unmatched = set(sec_group_list).difference(str(get_sg_name(all_sg, boto3)) for all_sg in all_sec_groups)
+    sec_group_name_list = list(set(sec_group_list) - set(unmatched))
+
+    if len(unmatched) > 0:
+        # If we have unmatched names that look like an ID, assume they are
+        import re
+        sec_group_id_list[:] = [sg for sg in unmatched if re.match('sg-[a-fA-F0-9]+$', sg)]
+        still_unmatched = [sg for sg in unmatched if not re.match('sg-[a-fA-F0-9]+$', sg)]
+        if len(still_unmatched) > 0:
+            raise ValueError("The following group names are not valid: %s" % ', '.join(still_unmatched))
+
+    sec_group_id_list += [ str(get_sg_id(all_sg, boto3)) for all_sg in all_sec_groups if str(get_sg_name(all_sg, boto3)) in sec_group_name_list ]
+
+    return sec_group_id_list
+
--- a/lib/ansible/module_utils/ec2.py
+++ b/lib/ansible/module_utils/ec2.py
@@ -379,7 +379,11 @@ def ansible_dict_to_boto3_tag_list(tags_dict):
 
 def get_ec2_security_group_ids_from_names(sec_group_list, ec2_connection, vpc_id=None, boto3=True):
 
-    """ Return list of security group IDs from security group names """
+    """ Return list of security group IDs from security group names. Note that security group names are not unique
+     across VPCs.  If a name exists across multiple VPCs and no VPC ID is supplied, all matching IDs will be returned. This
+     will probably lead to a boto exception if you attempt to assign both IDs to a resource so ensure you wrap the call in
+     a try block
+     """
 
     def get_sg_name(sg, boto3):
 


wimnat:
@defionscode @willthames @lwade - interested in feedback.  It's borrowing heavily from how the same functionality is implemented in ec2 module.  I'd like to push for a quick merge on this one as i've got an outstanding request to implement this in my ec2_eni module.

willthames:
There is a lot of common code. Only two lines are boto2/boto3 specific (get all the security groups, and create the unmatched set) - might be better to have one function with boto2/boto3 specific sections.

wimnat:
@willthames refactored to try and remove the duplication

andynelson:
When VPCs are being used the securitygroups are VPC specific, and names can be reused across VPCs... I'd suggest adding an optional vpc_id parameter to the subroutine to avoid the risk of returning an ID for a securitygroup from a different VPC than the required one.

(also, a securitygroup name starting 'sg-' is valid - looking for the group by name before dropping back to assuming it's an ID would seem prudent)

wimnat:
@andynelson thanks for the feedback.   I've added support for a vpc id filter and moved the sg- ID lookup until after the name lookup.

linuxdynasty:
In order to make Ansible more dynamic in nature and not rely on tools like CloudFormation and Terraform, I have written quite a few plugins that would do sort of what you are doing here. My question to everyone in this conversation, is this the correct approach? Meaning writing functions in ec2.py for making calls that are repeatable in aws in this one file?
- Would this be better as filter? (I'm not saying it will be, I am just asking)
- Maybe instead of just having 1 file for ec2, maybe we can break it up into multiple files? For instance ec2_query.py, for all modules that query the ec2 class of api calls? Another example would be for kinesis.. kinesis_query.py.

Maybe I am trying to decouple a little to much, but I think this would be a nice fit as well, I think when doing the import, this would be cleaner and easier to grok.

Also some accompanying tests would be nice as well.

wimnat:
Well this file shouldn't be called ec2.py at all as it is aws specific not ec2 but that's another story.  I would like this merged to help out with a number of aws modules that use security groups.  As for the greater discussion around breaking things out, maybe that's a conversation for the google group.

ready_for_review

wimnat:
@willthames @linuxdynasty @andynelson @defionscode friendly nudge for re-review

willthames:
meets_guidelines but I haven't tested it. I'd be happy for this to be merged into devel so that we can use some modules with it (as that's where the testing will really happen)

linuxdynasty:
I personally think the vpc_id should be mandatory as @andynelson pointed out. You can have duplicate names across vpc's and I think this will cause confusion. Also shouldn't boto be the default until the Ansible core team makes Boto3 the default?

wimnat:
@linuxdynasty thanks for your feedback.  boto3 is the default since Ansible 2.0.  You can't make vpc_id mandatory because the user may be using ec2 classic security groups in which case there is no vpc id.  Hope that answers your questions.

andynelson:
looks good to me

linuxdynasty:
@wimnat, I completely forgot about non VPC Security Groups. Looks good to me.

juliedavila:
I'll test this today. Assuming no issues I'll merge it in

juliedavila:
Ok so here are some comments, interested in hearing what you all think.

For @linuxdynasty:
- I think there is a use case for having similar (even duplicate at times) functionality via filters. However, I do see a valid reason to have these types of functions in module_utils as it allows devs to leverage that same type of logic within a module. It comes down to the use case.
- I completely agree with you and @wimnat that `ec2.py` is a really a crappy name for it. `aws.py` would be more proper. I would be 100% open to a PR that implements this, but it would also need to refactor all current aws modules from `from ansible.module_utils.ec2 import *` to `from ansible.module_utils.aws import *`. Right now there are no subdirs in module_utils so I would avoid going down the path of multiple libraries; I see the name change as a bigger thing.

For @wimnat:
- Tested and works nicely. 
- UX question. You implemented the vpc_id piece which is cool, and made it non-mandatory, which is also reasonable. However, can you, or anyone else, think of a use case where you'd want SG ids across multiple VPCs when their names match? I cannot, atm. If you agree, then whenever vpc id is omitted then it should only return sg ids from ec2 classic, ie sg ids w/o a relationship to a vpc. Currently, if you supply a name and no id, then it returns all matches.

andynelson:
Re: the UX question from @defionscode

I largely agree that in a classic account you'd likely only search without a vpc_id if you were working in classic, and with a vpc_id if you were working in a vpc.

_however_ if you're using a modern aws account, then operations such as securitygroup manipulation or instance launch use the 'default' VPC, which leads to many people treating it like classic (e.g. not specifying the vpc_id on operations, despite there being a vpc in place) if a search without a vpc_id wasn't returning the securitygroups in the default vpc, then it would likely appear broken due to not doing what people expect.

juliedavila:
I get what your saying here and would agree if we are discussing a specific ansible module. However, since this is about module_utils functionality should we not be more strict with module developers? In other words, would it not be better to force module developers (myself included) to be very explicit? I suppose adding an option for supporting `all` as a vpc id (or smth along those lines, and not as a default) might be a better approach. Thoughts?

linuxdynasty:
:+1:  for being explicit. Or maybe keyword argument vpc_type with these being the possible values ['classic', 'private', 'all']

andynelson:
so code gets added here to handle 'all' then every module that consumes it has to default 'all' if no vpc_id is required - I'm not seeing the value add?

juliedavila:
No, to clarify. I don't think the default behavior should be to gather all security group ids where the name matches across VPCs. When no VPC id is supplied I think it should do a 'search' and if more than 1 sg matches that name, it should then error out, as that would mean there are multiple sgs across VPCs or across VPC(s) and ec2 classic,  unless something like an `all` was supplied as the id which indicates teh developer explicitly wants all SG ids independent of duplication across vpcs

wimnat:
This main (and only?) purpose for this functionality is so that modules can take a list of sg IDs or sg names and use that in a subsequent call.  The subsequent call is almost certain to fail if the list returned contains sg IDs across multiple VPCs.

I'd prefer to leave edge case error handling to boto/boto3.  I don't think overly complex error handling here is necessary as boto will error out later with appropriate messages if we are returning IDs across two or more VPCs.
1. Let's make sure developers of modules explicitly specify a vpc id if they can (we can make a note in aws guidelines)
2. If they don't, we search all VPCs and return what we find.  Boto will error out anyway later on if there are sg IDs across multiple VPCs.  The boto error message should make it clear the sec group id isn't valid.

juliedavila:
@wimnat I did not see an error when I specified a SG name that (`default`) that exists across vpcs. The call gave me back all SG IDs with that name across a few VPCs.

wimnat:
Yeh but then if you use the list returned to you in a subsequent call, for
example, try to create an ec2 instance with that list, boto will error out.

I think it's OK to expect that exception be caught as any boto call should
be wrapped in a try statement.

wimnat:
@defionscode do you agree with my comment above or is more work required?

wimnat:
@defionscode sorry to hassle you on this but i just have a few other pending PRs for modules that want to use this code.  Any changes required?

juliedavila:
@wimnat can you just add a disclaimer in the function docstring that it if the name exists across the vpcs, and no vpc id is supplied, then all ids will be returned?

Let me know once you add that and I'll merge this in, unless of course you have an objection to that.

wimnat:
@defionscode done

juliedavila:
Thanks @wimnat

Issue #9823, opened 2014-12-15
ansible_sudo_pass override from inventory affects all subsequent hosts in a run


spacerodent:
##### Summary:

It appears that inventory variables are not limited to the host they are for
##### Steps To Reproduce:

Create a inventory file with three hosts listed.  The second one should specify the sudo password using ansible_sudo_pass.

[host one]
127.0.0.1
[host two]
127.0.0.1 ansible_sudo_pass=some_password
[host three]
127.0.0.1

Run the ping command using one fork

ansible -m ping -K -f 1

Assume host one and host two using a sudo password other than "some_password"
##### Expected Results:

The ansible_sudo_pass option is only applied to the host it is listed for
##### Actual Results:

Using my example inventory host three fails with a failed sudo password error

Original issue text:

In my hosts file I have a couple of hosts that specify the sudo password using the "ansible_sudo_pass" option.  Example:

[alpha]
192.168.10.1
[localhost]
127.0.0.1 ansible_sudo_pass=vagrant
[beta]
192.178.10.2

I found when doing a command on all (Example: ansible all -m ping -K) that some of my hosts would fail due to invalid sudo password.  It appeared to be semi-random until I set the forks value to 1.  Now every host that comes after localhost complains about an invalid sudo password.  

It appears to me that the ansible_sudo_pass value is being used for hosts other than the ones I specified it for.

I am running on OSX 10.10.1.  I am using the system python interpreter which is (2.7.6) and am using Ansible (1.8.2) installed via Brew (0.9.5).

mscherer:
I think that by design, variable are not reset to 0 after a play, in order to be able to reuse them. For example, that's how you can use facts from 1 server on the others. I would be tempted to say "this is working on purpose" but i indeed see how this is a problem in the case you gave. I wonder if that's the same with ansible_ssh_port and others variables, maybe that's a bug.

redbaron:
that is by design. Variables are kept per host name. if you don't want them to mixup, use different hostname pointing to the same IP address

spacerodent:
I don't think I did a good enough job of setting up the error condition so let me use a better example.

inventory file:
[alpha]
alpha.example.com
[beta]
beta.example.com
[vagrant]
localhost ansible_sudo_pass=vagrant
[gamma]
gamma.example.com
[delta]
delta.example.com

now if you do  "ansible -K -o -m ping all" you will see something like

beta.example.com | success >> { "changed": false, "ping": "pong" }
[...]

PR #15055, opened 2016-03-20
Fix ansible_sudo_pass inventory variable so that it overrides setting of sudo password from the command line

abadger:
##### Issue Type:
- Bugfix Pull Request
##### Ansible Version:

```
ansible 2.1.0 (ansible_sudo_pass-should-override-cli 70b031b3ad) last updated 2016/03/20 10:34:29 (GMT -700)
  lib/ansible/modules/core: (devel d71b9bae89) last updated 2016/03/20 07:36:06 (GMT -700)
  lib/ansible/modules/extras: (devel d5030ae555) last updated 2016/03/20 08:03:45 (GMT -700)
  config file = /srv/ansible/troubleshooting/ansible.cfg
```

Probably all ansible-2.0.x versions as well.
##### Summary:

When looking at whether #9823 was fixed, I found that although that bug is fixed, a new bug has cropped up.  Currently, if you have an inventory like this:

```
localhost ansible_sudo_pass=saving_passwords_to_disk_not_recommended
```

and run this command line:

```
ansible localhost -i inven -m command -a 'whoami' -K --sudo
```

the command line sudo password will override the ansible_sudo_pass in the inventory file.  Instead we want the setting in the inventory file to override the command line (the command line sets the default, the inventory sets host-specific overrides).  If you use ansible_become_pass in hte inventory, then things work as expected and the inventory host-var overrides the CLI.

This PR fixes that behaviour by adding logic to play_context to check for the deprecated CLI names separately from the deprecated inventory variable names.    We could instead fix it in VariableManager by having VariableManager know that become has replaced sudo.  However, that would mean that knowing about the deprecation of sudo and su is in both play_context and variablemanager.  I picked play_context so that the hacks are all in one place.

### DIFF
--- a/lib/ansible/playbook/play_context.py
+++ b/lib/ansible/playbook/play_context.py
@@ -361,7 +361,6 @@ class PlayContext(Base):
                 if exe_var in variables:
                     setattr(new_info, 'executable', variables.get(exe_var))
 
-
         attrs_considered = []
         for (attr, variable_names) in iteritems(MAGIC_VARIABLE_MAPPING):
             for variable_name in variable_names:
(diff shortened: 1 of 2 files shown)


lib/ansible/playbook/play_context.py
```diff
@@ -377,17 +376,34 @@ def set_task_and_variable_override(self, task, variables, templar):
...
             elif new_info.become_method == 'su' and new_info.su_pass:
                setattr(new_info, 'become_pass', new_info.su_pass)
 
+        # become legacy updates -- from inventory file (inventory overrides
+        # commandline)
+        for become_pass_name in MAGIC_VARIABLE_MAPPING.get('become_pass'):
+            if become_pass_name in variables:
+                break
+        else:  # This is a for-else
+            if new_info.become_method == 'sudo':
```
bcoca, line 392: A simpler version might be setting the become sudo_pass magic var and letting the lines above (385) deal with it

jimi-c:
@abadger since we have `test/units/playbook/test_play_context.py`, I'd like to see some unit tests added to verify this (and to make sure we don't break anything).

lib/ansible/playbook/play_context.py
abadger, line 392: Lines 380-85 doesn't deal with this because it won't override become_pass.  We need to know that we're dealing with vars that came from the inventory and that they're only overriding vars that came from the command line.

abadger:
Works in travis except for an unrelated problem downloading repodata in the yum module on fedora rawhide.  Merging.

PR #15060, opened 2016-03-21
Vault improvements: base64 encoding, no AES-CTR padding, code cleanups

### LINKED ISSUE #12121: Vault files are 3 times bigger than they need to be
##### ISSUE TYPE
 - Feature Idea

##### COMPONENT NAME
vault

##### ANSIBLE VERSION
```
ansible 2.2.1.0
  config file = 
```

##### CONFIGURATION
n/a

##### OS / ENVIRONMENT
n/a

##### SUMMARY
The current Vault formats encode binary data as hex, add some newlines, encode it as hex again, then wrap at 80 columns.   This results in data files that are 4 times bigger than the encoded binary data.

I suggest using base64 (once) instead, which would result in data files that are only 1.35 times bigger.

My primary motivation is shorter `git diff` output for changes in encrypted vault files.

This could perhaps be done at the same time as dropping the unneeded PKCS#7 padding (PR #10634), since it would require introducing a new format version.

amenonsen:
##### Issue Type:
- Feature Pull Request
##### Summary:

Introduce an improved AES256 cipher version

Now we don't add spurious padding for AES-CTR (which is used as a stream cipher, and doesn't need padding) and we use base64 to encode the output message instead of blowing up its size by hexlify()ing everything twice. We also don't unnecessarily use PBKDF2 to generate the initial value for the encryption counter.

We always use the new format for output, but maintain read-compatibility with a few lines of 1.1-specific decoding in decrypt(). (Note that we do not force v1.1 files to be rewritten by ansible-vault edit unless their
contents have actually changed, in which case the new cipher version is used automatically).

Closes #10634 from graingert (which removed the padding, but added yet another cipher class for backwards compatibility)

Fixes #12121 from mgedmin (which reported the double-hexlify() problem)

### DIFF
--- a/lib/ansible/parsing/vault/__init__.py
+++ b/lib/ansible/parsing/vault/__init__.py
@@ -24,6 +24,7 @@ import shutil
 import sys
 import tempfile
 import random
+import base64
 from io import BytesIO
 from subprocess import call
 from ansible.errors import AnsibleError
@@ -615,7 +616,7 @@ class VaultAES256:
 
         check_prereqs()
 
-        self.version = b'1.1'
+        self.version = b'1.2'
 
     def create_key(self, password, salt, keylength, ivlength):
         hash_function = SHA256
@@ -627,12 +628,28 @@ class VaultAES256:
                             count=10000, prf=pbkdf2_prf)
         return derivedkey
 
-    def gen_key_initctr(self, password, salt):
+    def gen_key_initctr(self, password, salt, generate_iv=False):
+        '''
+        Takes a password and random 32-byte salt and stretches it using
+        PBKDF2 into an 32-byte AES key, a 32-byte HMAC-SHA-256 key, and
+        (optionally) a 16-byte IV for AES-CTR.
+
+        The initial value for the counter is zero by default, but we support
+        generating a value using PBKDF2 for backwards compatibility.
+
+        CTR mode requires that the same key and the same counter are only ever
+        used once to encrypt data. Since we generate the key using a random salt
+        each time, the initial counter value can be 0. Using PBKDF2 to generate
+        the value just slows things down (specifically, it adds 10000 iterations
+        of the PRF, and then throws half of the 32-byte output away).
+        '''
         # 16 for AES 128, 32 for AES256
         keylength = 32
 
         # match the size used for counter.new to avoid extra work
-        ivlength = 16
+        ivlength = 0
+        if generate_iv:
+            ivlength = 16
 
         if HAS_PBKDF2HMAC:
             backend = default_backend()
@@ -648,10 +665,13 @@ class VaultAES256:
 
         key1 = derivedkey[:keylength]
         key2 = derivedkey[keylength:(keylength * 2)]
-        iv = derivedkey[(keylength * 2):(keylength * 2) + ivlength]
 
-        return key1, key2, hexlify(iv)
+        iv = 0
+        if generate_iv:
+            iv = derivedkey[(keylength * 2):(keylength * 2) + ivlength]
+            iv = int(hexlify(iv), 16)
 
+        return key1, key2, iv
 
 
     def encrypt(self, plaintext, password):
@@ -659,16 +679,11 @@ class VaultAES256:
         salt = os.urandom(32)
         key1, key2, iv = self.gen_key_initctr(password, salt)
 
-        # PKCS#7 PAD DATA http://tools.ietf.org/html/rfc5652#section-6.3
-        bs = AES.block_size
-        padding_length = (bs - len(plaintext) % bs) or bs
-        plaintext += to_bytes(padding_length * chr(padding_length), encoding='ascii', errors='strict')
-
         # COUNTER.new PARAMETERS
         # 1) nbits (integer) - Length of the counter, in bits.
         # 2) initial_value (integer) - initial value of the counter. "iv" from gen_key_initctr
 
-        ctr = Counter.new(128, initial_value=int(iv, 16))
+        ctr = Counter.new(128, initial_value=iv)
 
         # AES.new PARAMETERS
         # 1) AES key, must be either 16, 24, or 32 bytes long -- "key" from gen_key_initctr
@@ -680,8 +695,7 @@ class VaultAES256:
         ciphertext = cipher.encrypt(plaintext)
         hmac = HMAC.new(key2, ciphertext, SHA256)
 
-        message = b'\n'.join([hexlify(salt), to_bytes(hmac.hexdigest()), hexlify(ciphertext)])
-        message = hexlify(message)
+        message = base64.b64encode(salt+hmac.digest()+ciphertext)
         lines = [b'%s\n' % message[i:i+80] for i in range(0, len(message), 80)]
         data = ''.join(lines)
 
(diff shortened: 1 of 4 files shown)


tomato42:
I don't see unit test for the new file format, there's only the old one (1.1) in https://github.com/ansible/ansible/blob/devel/test/units/parsing/vault/test_vault.py

the number of iterations for PKBDF2 should be stored with the encrypted data, and the number itself almost certainly should be increased to something that takes at least 0.1s to compute on modern CPU

lib/ansible/parsing/vault/__init__.py
```diff
@@ -690,32 +704,39 @@ def encrypt(self, plaintext, password):
...
 
         # EXIT EARLY IF MAC DOESN'T VALIDATE
         hmac = HMAC.new(key2, ciphertext, SHA256)
-        if not self.is_equal(mac, to_bytes(hmac.hexdigest())):
+        if not self.is_equal(mac, to_bytes(hmac.digest())):
             return None
 
         # SET THE COUNTER AND THE CIPHER
-        ctr = Counter.new(128, initial_value=int(iv, 16))
+        ctr = Counter.new(128, initial_value=iv)
```
tomato42, line 727: I'm not familiar with this API, but if you set the "initial value" of a counter, where do you set the nonce?

In line 669 you're setting it to zero, so is the whole counter 0?

amenonsen, line 727: The whole counter is 0.

amenonsen:
> I don't see unit test for the new file format, there's only the old one (1.1) in https://github.com/ansible/ansible/blob/devel/test/units/parsing/vault/test_vault.py

`test_encrypt_decrypt_aes256`

amenonsen:
> the number of iterations for PKBDF2 should be stored with the encrypted data

Good idea.

> the number itself almost certainly should be increased to something that takes at least 0.1s to compute on modern CPU

Such as?

lib/ansible/parsing/vault/__init__.py
tomato42, line 727: Then it is not secure. CTR mode requires an unpredictable and unique value as part of the counter. Previous code was correct in taking the full IV from PBKDF2 output.

tomato42:
> `test_encrypt_decrypt_aes256`

you do not have test with an explicit `'$ANSIBLE_VAULT;1.2` to decrypt

> >   the number itself almost certainly should be increased to something that takes at least 0.1s to compute on modern CPU
> 
> Such as?

Probably around a 10^5 or 10^6 as a baseline, more if the CPU the system is running on is fast

lib/ansible/parsing/vault/__init__.py
amenonsen, line 727: CTR mode _requires_ no such thing. Its security depends on a given counter value never being reused with the same key, which we ensure by using a new random seed for PBKDF2 every time.

tomato42, line 727: yes, and the best way to make sure that the key-IV pair does not repeat when you can't have a full list of all IVs used is to use random values in both

random seed for PBKDF2 is not enough, it just makes the key random, not unique

amenonsen, line 727: And what makes a random nonce unique?

tomato42, line 727: nothing. The point is to make collision between nonce and key as unlikely as possible. Using static nonce doesn't give you that.

Cryptography is hard enough to get right _without_ going against recommended best practice.

amenonsen, line 727: > The point is to make collision between nonce and key as unlikely as
> possible. Using static nonce doesn't give you that.

We generate a random seed for PBKDF2 using os.urandom(32).

So you're concerned that we may accidentally reuse the same AES-256 key
by reading the same 32 bytes from /dev/urandom on multiple occasions,
and want to further reduce the likelihood of key reuse.

Since you say "as unlikely as possible", I infer that you would prefer
to use a 32-byte random value instead of using a 16-byte random value
and a 16-byte counter as the initial counter value.

Would you be satisfied if I called os.urandom(64), used the first 32
bytes as the PBKDF2 seed, and the rest as the initial counter value?

tomato42, line 727: If you're willing to modify the code further, then yes
1. using salt the size of SHA-256 output for PBKDF2 
2. using the initial counter value (nonce) directly from os.urandom

Would be the best possible approach. (The first is a _bit_ of overkill, but since we're using AES-256, it's probably safer this way)

Ah, and by the way. I think it would be good to add comments like "We're using SHA-256 because we're using AES with 256-bit key size", "There's no need to use bigger salt size than 32 bytes as we're using PBKDF2 with SHA-256, which has 32 byte output" and "The maximum size of the nonce is the block size of cipher, 16 bytes in case of AES" to explain to people reading the code later why we're using such and such magic values and not others (and as a hint when they may need to be changed).

lib/ansible/parsing/vault/__init__.py
```diff
@@ -631,26 +665,25 @@ def gen_key_initctr(self, password, salt):
 
         key1 = derivedkey[:keylength]
         key2 = derivedkey[keylength:(keylength * 2)]
-        iv = derivedkey[(keylength * 2):(keylength * 2) + ivlength]
 
-        return key1, key2, hexlify(iv)
+        iv = 0
+        if generate_iv:
```
graingert, line 670: why does it need an iv for backwards compatibility? Surely old clients will just read the 0 IV and work fine?

robinro:
@amenonsen any update on this? I hoped this would make 2.2, which it didn't, but maybe it can be in 2.3 at least?

amenonsen:
Yes, I'm working on this.

alikins:
I mentioned in IRC some multiple vault identity support as something to be aware of. I rebased the WIP branch and fixed it up to partially working, mostly just for reference since there doesnt seem to be any interest in the feature. It is at https://github.com/alikins/ansible/tree/vaults_secrets_more. Mostly it replaces passing a password string around with passing around a VaultSecrets object. VaultSecrets itself can hold multiple identities as well as a 'default' identity for backwards compat.

Lekensteyn:
Hey, the idea of this change (use base64 instead of double hex; remove unnecessary padding) looks good to me. (Maybe the cleanups could be split a bit in separate commits, I skipped over them because it contains a lot of unrelated changes which made reviewing harder.)

(I have done a small audit of the existing Ansible Vault code, and while the old "AES" variant is broken due to a weak KDF, the "AES256" seems "okay" other than wasting space (double hex) and unnecessary complexity (incl. padding.)

Some other comments:

- For forwards compat, shouldn't "b_version" be checked (and error out or warn)? A quick look in the `parsing/vault/__init__.py`  file did not show a user of the `b_version` field. Older clients will presumably crash on data generated by new programs because `unhexlify(b_vaulttext)` chokes on base64 data. Throwin an error would clearly signal that the user that the program is too old to understand it. Warning (or silently ignoring as is done now) would allow for decrypting containers if the format has not drastically been changed.
- Ideally the PBKDF2 iteration count is also stored in the file such that it can be increased in the future if needed.
- (Meta-comment: I wonder how much prior research has been done on encrypted container formats, some of these issues were avoidable...)

calfonso:
@alikins would you mind discussing this with our security person to ensure you're doing the right thing here. Also, if you could update us on the details/timing of resolving this issue that would be very helpful. We'd like to ensure anything we do is backwards compatible. Thanks!

abeluck:
Has this cleanup stalled out?

amenonsen:
> Has this cleanup stalled out?

The code has changed (for the worse) so much in the meantime that I
couldn't finish it in the time I hoped. But I'm still working on it,
because I need the feature.

bcoca:
closing due to inactivity, feel free to resubmit as a new PR if you find the time.

PR #15062, opened 2016-03-21
Add `issubset` and `issuperset` tests

gechr:
##### Issue Type:
- Feature Pull Request
##### Ansible Version:

```
2.1.0
```
##### Summary:

I needed to use `isssubet` to validate a manually-configured list against a pre-configured list of valid items and noticed it was missing!

Besides, Ansible implements all other set operations from https://docs.python.org/2/library/sets.html#set-objects, so I can't see a reason not to implement these as well.


bcoca:
these might be better as 'tests', which would make these usable with map/select type filters

gechr:
@bcoca Done + rebased.

Issue #15067, opened 2016-03-21
EC2 dynamic inventory script should vary cache file when using different AWS profiles


OwenTuz:
PR below, just creating an issue to tie it to. This is based on the latest version in the 'devel' branch.

The EC2 inventory script currently caches AWS API lookups in "<configurable cache_path>/ansible-ec2.cache". 

For those of us who manage multiple AWS accounts, this is a problem as if I run the inventory script in account A, then in account B within the cache expiry time (5 minutes by default), I receive the cached results from account A.

At best, this is inconvenient, but at worst could lead to playbooks being executed against hosts in the wrong account (...no, I haven't, but it is the sort of thing I would manage to do).

This can already be worked around by keeping multiple ec2.ini config files, each with a different cache path. However, for a more sensible default and simpler behaviour out of the box, I would prefer to simply vary the cache filename depending on the Boto profile in use.

The attached PR uses the profile (if present) to form part of the cache filename.

amenonsen:
Closed via #15068.

PR #15068, opened 2016-03-21
Issue 15067

OwenTuz:
https://github.com/ansible/ansible/issues/15067

If a Boto profile is in use (either because it has been passed to the script as an argument or in the config file, or because it is being passed directly to Boto via the AWS_PROFILE environment variable), the name of the profile will be used to form the cache file that stores API lookups. 

This PR should prevent the situation where lookups for one account return cached data from another account. It should reflect the order of precedence:

arguments to CLI script > AWS_PROFILE environment variable > AWS_ACCESS_KEY_ID environment variable

### DIFF
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -331,8 +331,14 @@ class Ec2Inventory(object):
         if not os.path.exists(cache_dir):
             os.makedirs(cache_dir)
 
-        self.cache_path_cache = cache_dir + "/ansible-ec2.cache"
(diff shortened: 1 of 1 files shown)


amenonsen:
This looks good to me. Would you please squash the commits and remove the whitespace changes? That would make it easier to apply to back branches.

OwenTuz:
Done, thanks

OwenTuz:
Done, thanks.

amenonsen:
Merged, thanks.

PR #15069, opened 2016-03-21
Use docker cp instead of docker volumes to allow for testing with remote docker servers

sivel:
##### Issue Type:

Feature Pull Request
##### Ansible Version:

```
v2
```
##### Summary:

Instead of using docker volumes, which is really only capable of doing bind mounts with the host docker server, use `docker cp` instead.  This makes it easier to use a remote docker server instead of having to run docker locally.

cc @jimi-c 
##### Example output:

```
N/A
```

### DIFF
--- a/test/utils/run_tests.sh
+++ b/test/utils/run_tests.sh
@@ -10,7 +10,8 @@ if [ "${TARGET}" = "sanity" ]; then
 else
     set -e
     export C_NAME="testAbull_$$_$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 8 | head -n 1)"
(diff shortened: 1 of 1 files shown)


mattclay:
Keep in mind that this will only work with newer versions of docker ([1.8.0](https://github.com/docker/docker/blob/master/CHANGELOG.md#180-2015-08-11) and later).

jimi-c:
@mattclay I believe we're already making that assumption due to our use of the `:Z` option for selinux contexts.

jimi-c:
This was merged, but some rebasing/cleanup is making github not auto-close it.

Thanks!

PR #15074, opened 2016-03-21
Cleanup some minor issues in PuginLoader:

abadger:
##### Issue Type:
- Bugfix Pull Request
##### Ansible Version:

```
ansible 2.1.0 (ansible_sudo_pass-should-override-cli 70b031b3ad) last updated 2016/03/20 10:34:29 (GMT -700)
  lib/ansible/modules/core: (devel d71b9bae89) last updated 2016/03/20 07:36:06 (GMT -700)
  lib/ansible/modules/extras: (devel d5030ae555) last updated 2016/03/20 08:03:45 (GMT -700)

```
##### Summary:

This cleanup fixes the following latent bugs in PluginLoader.  We may have run across these and not known it, ignored the limitation at the time (class_only I think falls into that case), or just not used the API in the way that would trigger it at the moment.
- class_only was a keyword arg of get() and all() that was mistakenly
  passed on to Plugins.  Be sure to strip it from the keyword args
  before instantiating Plugins.  (Reworked API probably should either
  not instantiate Plugins or take the args for the Plugin as a separate
  list and a dict.)
- Checking required base_classes was only done in get() and only if
  class_only was False (ie: that Plugin was instantiated).  This meant
  that different plugins could be found depending on whether the call
  was to .get() or to all() and whether it was for classes or instances.
  Fixed so that required base_classes are always checked.

### DIFF
--- a/lib/ansible/plugins/__init__.py
+++ b/lib/ansible/plugins/__init__.py
@@ -316,6 +316,7 @@ class PluginLoader:
     def get(self, name, *args, **kwargs):
         ''' instantiates a plugin of the given name using arguments '''
 
+        class_only = kwargs.pop('class_only', False)
         if name in self.aliases:
             name = self.aliases[name]
         path = self.find_plugin(name)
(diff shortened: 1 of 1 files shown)


abadger:
Passed in travis except for an unrelated transient error fetching yum repo data on one builder.  merging.

PR #15078, opened 2016-03-21
Don't create world-readable module and tempfiles without explicit user permission

abadger:
##### Issue Type:
- Bugfix Pull Request
##### Ansible Version:

```
ansible 2.1.0 (module-permissions-fix 7dbe671775) last updated 2016/03/21 14:18:23 (GMT -700)
  lib/ansible/modules/core: (devel 22a9a15a5b) last updated 2016/03/18 09:09:18 (GMT -700)
  lib/ansible/modules/extras: (devel 33e1d9d1cb) last updated 2016/03/18 09:05:36 (GMT -700)
  config file = /srv/ansible/troubleshooting/ansible.cfg
```
##### Summary:

When Ansible uses become to switch users and the new user is unprivileged it creates temporary files on the remote system as one user and then makes them world readable so that the second user can read them.  This can be a securitry hazard if all users on the remote machine are not trusted and any sensitive data is given as a parameter to the module.

This PR mitigates that behaviour in several ways:
- If the connection to the remote machine is made as the root user, we chown the files to the second unprivileged user rather than setting them world readable.
- If the remote machine has filesystem acls and the setfacl command in the path then we use fs acls to grant read permission just to the second unprivileged user.  We no longer have to trust everyone on the remote host.
- If Ansible has no other way to give the second unprivileged user permissions on the module except to make them world readable, ansible will fail unless the user has set an ansible.cfg setting to allow world readable tmpfiles.  This makes the user aware of the insecurity but allows them to do it anyway if they trust their environment (or as a stopgap as they port their playbooks to more secure styles).

The security implications are documented in become.rst.

### DIFF
--- a/docsite/rst/become.rst
+++ b/docsite/rst/become.rst
@@ -85,25 +85,33 @@ on how it works.  Users should be aware of these to avoid surprises.
 Becoming an Unprivileged User
 =============================
 
-Ansible has a limitation with regards to becoming an
+Ansible 2.0.x and below has a limitation with regards to becoming an
 unprivileged user that can be a security risk if users are not aware of it.
 Ansible modules are executed on the remote machine by first substituting the
 parameters into the module file, then copying the file to the remote machine,
-and finally executing it there.  If the module file is executed without using
-become, when the become user is root, or when the connection to the remote
-machine is made as root then the module file is created with permissions that
-only allow reading by the user and root.
+and finally executing it there.
 
-If the become user is an unprivileged user and then Ansible has no choice but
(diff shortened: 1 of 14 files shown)


abadger:
Travis is now passing -- ready for review.

abadger:
Travis says failing on centos7 and fedora but the tests all completed successfully.  It's the test framework that's having troubles shutting down the containers cleanly, not the tests themselves that are failing.

abadger:
Updated with some changes to documentation.  No changes to code.

PR #15106, opened 2016-03-23
check required and default in py code

robberphex:
check required and default in py code
##### Issue Type:
- fix test code
##### Ansible Version:

```
devel
```
##### Summary:

In test code `test/code-smell/required-and-default-attributes.sh` just check `FieldAttribute`, `default` and `required` appear in same line.

We can check this in run-time.


sivel:
I'm not sure I am a fan of this implementation.  I'd like to see us get away from using `assert` in the codebase at all.  If python is run with optimizations, `assert` statements will be removed from the code.

mattclay:

   * We already have many sanity checks such as this as part of our automated tests. We would like to keep them there, instead of making them run-time checks.

PR #15144, opened 2016-03-24
Add winrm tests and fix exposed unicode errors.

mattclay:
##### ISSUE TYPE

Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0 (winrm-test 35c2b45fbe) last updated 2016/03/24 10:46:43 (GMT -700)
  lib/ansible/modules/core: (detached HEAD 7efc09ef08) last updated 2016/03/24 08:50:40 (GMT -700)
  lib/ansible/modules/extras: (detached HEAD 7f9cdc0350) last updated 2016/03/24 08:50:40 (GMT -700)
  config file = 
```
##### SUMMARY

Add winrm version of connection tests and add missing to_bytes on file paths exposed by the tests.

Tests were run on a Ubuntu 15.10 control host against a Windows Server 2012 R2 install.

The test command used was: `TEST_FLAGS='-vvvvv' make test_connection_winrm`

Before adding the missing to_bytes calls, the tests failed. The first failure was:

```
  File "/home/matt/mattclay/ansible/lib/ansible/executor/task_executor.py", line 124, in run
    res = self._execute()
  File "/home/matt/mattclay/ansible/lib/ansible/executor/task_executor.py", line 426, in _execute
  File "/home/matt/mattclay/ansible/lib/ansible/plugins/action/copy.py", line 215, in run
    self._transfer_file(source_full, tmp_src)
  File "/home/matt/mattclay/ansible/lib/ansible/plugins/action/__init__.py", line 257, in _transfer_file
    self._connection.put_file(local_path, remote_path)
  File "/home/matt/mattclay/ansible/lib/ansible/plugins/connection/winrm.py", line 268, in put_file
    if not os.path.exists(in_path):
  File "/usr/lib/python2.7/genericpath.py", line 26, in exists
    os.stat(path)
```

After the to_bytes calls were added, the tests passed, with a play recap of:

```
winrm-no-pipelining        : ok=10   changed=7    unreachable=0    failed=0   
winrm-pipelining           : ok=10   changed=7    unreachable=0    failed=0 
```

Each change made to winrm was the result of encountering an error while running the tests.

NOTE: When porting the connection tests to winrm, the pipelining test was kept in case winrm gains pipelining support in the future.

### DIFF
--- a/test/integration/test_connection_winrm.yml
+++ b/test/integration/test_connection_winrm.yml
@@ -1,0 +1,37 @@
+- hosts: winrm
+  gather_facts: no
+  serial: 1
+  tasks:
+
+  ### raw with unicode arg and output
+
+  - name: raw with unicode arg and output
+    raw: echo 汉语
+    register: command
+  - name: check output of raw with unicode arg and output
+    assert: { that: "'汉语' in command.stdout" }
+
+  ### copy local file with unicode filename and content
+
+  - name: create local file with unicode filename and content
+    local_action: lineinfile dest=/tmp/ansible-local-汉语/汉语.txt create=true line=汉语
+  - name: remove remote file with unicode filename and content
+    win_file: path=c:/windows/temp/ansible-remote-汉语/汉语.txt state=absent
+  - name: create remote directory with unicode name
+    win_file: path=c:/windows/temp/ansible-remote-汉语 state=directory
+  - name: copy local file with unicode filename and content
+    win_copy: src=/tmp/ansible-local-汉语/汉语.txt dest=c:/windows/temp/ansible-remote-汉语/汉语.txt
+
+  ### fetch remote file with unicode filename and content
+
+  - name: remove local file with unicode filename and content
+    local_action: file path=/tmp/ansible-local-汉语/汉语.txt state=absent
+  - name: fetch remote file with unicode filename and content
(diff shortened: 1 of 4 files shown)


test/integration/test_connection_winrm.inventory
```diff
@@ -0,0 +1,10 @@
+[winrm]
```
nitzmahone: This should probably just be combined with inventory.winrm.template (with the vagrant-ish details scrubbed out), since anyone wanting to run this will need to adjust accordingly (just as with standard winrm test inventory).

test/integration/test_connection_winrm.yml
```diff
@@ -0,0 +1,37 @@
+- hosts: all
```
nitzmahone: probably better to target the specific hosts/groups that you're trying to test (eg, `hosts: winrm-pipelining:winrm-no-pipelining` or `hosts: winrm`) in case someone's running with a broader inventory.

nitzmahone:
Looks good, once we figure out what shape the inventory should be. I'd probably lean towards combining this with the standard winrm inventory template, though you'll probably want to isolate the aliased hosts to their own group(s) and target those groups explicitly, or use add_host in the playbook to do the aliased copy with pipelining turned on instead (should be a noop at this point since winrm doesn't yet support it).

PR #15145, opened 2016-03-24
[WIP] fix for incorrect template search paths

bcoca:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
2.x
```
##### SUMMARY

jinja2 loader now uses dependency chain vs role path var, which could also be `:` separated path list
also added role dep path to primary template search


bcoca:
closing in favor of #16387

PR #15146, opened 2016-03-24
Override params from environment variables, if set.

Qalthos:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
2.1
```
##### SUMMARY

Adds support for `ANSIBLE_NET_USERNAME`, `ANSIBLE_NET_PASSWORD`, `ANSIBLE_NET_AUTHORIZE`, `ANSIBLE_NET_AUTH_PASS`, and `ANSIBLE_NET_SSH_KEYFILE` to the EOS module.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -1387,6 +1387,8 @@ class AnsibleModule(object):
                 wanted = 'str'
 
             value = self.params[k]
+            if value is None:
+                continue
 
             try:
                 type_checker = self._CHECK_ARGUMENT_TYPES_DISPATCHER[wanted]
--- a/lib/ansible/module_utils/eos.py
+++ b/lib/ansible/module_utils/eos.py
@@ -16,6 +16,7 @@
 # You should have received a copy of the GNU General Public License
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 #
+import os
 
 NET_PASSWD_RE = re.compile(r"[\r\n]?password: $", re.I)
 
@@ -24,11 +25,20 @@ NET_COMMON_ARGS = dict(
     port=dict(type='int'),
     username=dict(required=True),
     password=dict(no_log=True),
+    ssh_keyfile=dict(type='path'),
     authorize=dict(default=False, type='bool'),
     auth_pass=dict(no_log=True),
     transport=dict(default='cli', choices=['cli', 'eapi']),
     use_ssl=dict(default=True, type='bool'),
     provider=dict(type='dict')
+)
+
+NET_ENV_ARGS = dict(
+    username='ANSIBLE_NET_USERNAME',
+    password='ANSIBLE_NET_PASSWORD',
+    ssh_keyfile='ANSIBLE_NET_SSH_KEYFILE',
+    authorize='ANSIBLE_NET_AUTHORIZE',
+    auth_pass='ANSIBLE_NET_AUTH_PASS',
 )
 
 CLI_PROMPTS_RE = [
@@ -139,12 +149,13 @@ class Cli(object):
 
         username = self.module.params['username']
         password = self.module.params['password']
+        key_filename = self.module.params['ssh_keyfile']
 
         try:
             self.shell = Shell(CLI_PROMPTS_RE, CLI_ERRORS_RE)
-            self.shell.open(host, port=port, username=username, password=password)
+            self.shell.open(host, port=port, username=username, password=password, key_filename=key_filename)
         except Exception, exc:
-            msg = 'failed to connecto to %s:%s - %s' % (host, port, str(exc))
+            msg = 'failed to connect to %s:%s - %s' % (host, port, str(exc))
             self.module.fail_json(msg=msg)
 
     def authorize(self):
@@ -177,9 +188,12 @@ class NetworkModule(AnsibleModule):
         params = super(NetworkModule, self)._load_params()
         provider = params.get('provider') or dict()
         for key, value in provider.items():
-            if key in NET_COMMON_ARGS.keys():
-                if not params.get(key) and value is not None:
+            if key in NET_COMMON_ARGS:
+                if params.get(key) is None and value is not None:
                     params[key] = value
+        for key, env_var in NET_ENV_ARGS.items():
+            if params.get(key) is None and env_var in os.environ:
+                params[key] = os.environ[env_var]
         return params
 
     def connect(self):
--- a/lib/ansible/utils/module_docs_fragments/eos.py
+++ b/lib/ansible/utils/module_docs_fragments/eos.py
@@ -45,9 +45,15 @@ options:
     required: true
   password:
     description:
-      - Specifies the password to use when authentication the connection to
+      - Specifies the password to use to authenticate the connection to
         the remote device.  This is a common argument used for either I(cli)
         or I(eapi) transports.
+    required: false
+    default: null
+  ssh_keyfile:
+    description:
+      - Specifies the SSH keyfile to use to authenticate the connection to
+        the remote device.  This argument is only used for I(cli) transports.
     required: false
     default: null
   authorize:


Qalthos:
This _is_ missing `ANSIBLE_NET_SSH_KEYFILE` as there didn't seem to be an obvious way of achieving that in my exploration, but I wanted to get visibility for my implementation in the meantime.

Qalthos:
A trip through EOS' documentation later, and this handles `ANSIBLE_NET_SSH_KEYFILE` now, albeit in a very different way.

lib/ansible/module_utils/eos.py
```diff
@@ -180,6 +186,17 @@ def _load_params(self):
...
+        env_overrides = dict(
+            username='ANSIBLE_NET_USERNAME',
+            password='ANSIBLE_NET_PASSWORD',
+            ssh_keyfile='ANSIBLE_NET_SSH_KEYFILE',
+            authorize='ANSIBLE_NET_AUTHORIZE',
+            auth_pass='ANSIBLE_NET_AUTH_PASS',
+        )
+        for key, env_var in env_overrides.items():
+            if env_var in os.environ:
+                params[key] = os.environ[env_var]
```
abadger: I don't think we want to have environment variables overriding parameters.  What is okay is for environment variables to set defaults if the parameters are not given.

To do that you probably want the parameters to set a default value of None in the argument_spec.  Then in here you can check whether they are None.  If so, then try to set from the environment.  If it's still None and you have a sensible default (doesn't look like we would here except maybe for ssh_keyfile) then set it to that value here.

Also, could shorten this code slightly like this:

``` python
for key, env_var in env_overrides.items():
    params[key] = os.environ.get(env_var, None)
```

lib/ansible/module_utils/eos.py
```diff
@@ -139,12 +141,16 @@ def connect(self, **kwargs):
...
         password = self.module.params['password']
+        key_filename = self.module.params['ssh_keyfile']
 
         try:
             self.shell = Shell(CLI_PROMPTS_RE, CLI_ERRORS_RE)
-            self.shell.open(host, port=port, username=username, password=password)
+            if key_filename:
+                self.shell.open(host, port=port, username=username, key_filename=key_filename)
+            else:
+                self.shell.open(host, port=port, username=username, password=password)
```
abadger: Why not implement this logic (whether to use key_filename or password) in the Shell.open method?

Qalthos: Thank you, I thought something still looked suspicious here. Really, there's no need for any check at all, paramiko uses key_filename first, and fails to password anyway.

lib/ansible/module_utils/eos.py
Qalthos: I thought this way too at first, and that's how my very first attempt went, but upon rereading @privateip's request, I think this is the intended order. Specifically, the precedence order given to me was:
1. Env
2. Task arg
3. Provider arg

This code should apply the values in that order; the task arguments are initially in place, then the provider arguments fill in any gaps, and finally the environment variables clobber anything if a value is set. This is also specifically why I don't use `environ.get()` here, I don't want to replace existing parameters with `None`.

abadger: For people following at home, we had a talk with privateip, qalthos, bcoca, jimi-c, and myself about this on slack, decided for now to have parameters overriding env vars.  We'll discuss this more next week and see if we can push the values somewhere else (perhaps into inventory variables that are set on the connection plugin).

Qalthos:
Well, I should have expected that--we do indeed need _some_ logic to avoid sending `'None'` to paramiko (instead of `None`) :sweat_smile:

lib/ansible/module_utils/eos.py
```diff
@@ -24,6 +25,7 @@
     port=dict(type='int'),
     username=dict(required=True),
```
abadger, line 26: This should now be required=False right?

lib/ansible/module_utils/eos.py
```diff
@@ -24,6 +25,7 @@
...
     password=dict(no_log=True),
+    ssh_keyfile=dict(type='path'),
     authorize=dict(default=False, type='bool'),
```
abadger, line 29: And this needs to default None now, I think.

lib/ansible/module_utils/eos.py
```diff
@@ -180,6 +191,9 @@ def _load_params(self):
             if key in NET_COMMON_ARGS.keys():
                 if not params.get(key) and value is not None:
                     params[key] = value
+        for key, env_var in NET_ENV_ARGS.items():
+            if not params.get(key) and os.environ.get(env_var):
```
abadger: This should ask if params.get(key) is None.  That way you can tell that the user didn't specify a falsey value (like authorize=False).

privateip:
Agreed with @abadger on comments for removing required and default values.  For required arguments (those that have required=True today) though, we should introduce a function to verify that all required values have been set by some form or fashion otherwise return fail_json in a meaningful way

lib/ansible/module_utils/eos.py
```diff
@@ -180,6 +191,9 @@ def _load_params(self):
...
+                params[key] = os.environ[env_var]
```
abadger: One more thing needs to go in this section:  defaults and required.
- Since we have to set the argspec to default to a sentinel value (None) so that we know the difference between "user didn't specify anything" and "user specified something that's falsey" we also need to add any default values here.  Judging from the previous argspec, this is just if authorize is still None then set authorize equal to False.
- A similar thing happens with required items.  We have to set the argspec to required=False so that we can also look for the values in these envvars.  Once we've looked added any envvars to the parameters we have to manually check that any of those that we require are now set.  Judging from the argspec changes, we just need to check if username is still None then call module.fail_json() to let the user know that a username has to be set.

abadger:
After some IRC conversation and looking at the code, @Qalthos pointed out that `AnsibleModule.__init__` is calling `_load_params()`, then `_set_defaults()`, and then `_check_required_arguments()`.  So you can disregard everything from my last review... it should all work this way.  So PR approved.  Feel free to merge!

Future thought:

Depending on ordering in AnsibleModule is somewhat fragile (`AnsibleModule. __init__()` does too much so refactoring can sometimes confuse people and they end up breaking things.)  We probably should think about a way to do this in the future where either people will know that changing the ordering will break things or that doesn't depend on the ordering.

For the first option, maybe we could integrate environment variable defaults into a future version of the argspec.  This would be helpful for both networking and cloud modules.  They could then specify something like:

```
argument_spec=dict(
   username=dict(required=False, default='admin', env_default='OPENSTACK_ADMIN', type='str'),
   password=dict(required=True, env_default='OPENSTACK_ADMIN_PASS', type='str')
)
```

and the default _load_params in basic.py would take care of loading the environment variable if no parameter was passed in the playbook.

We can then add this to the unittests so that if someone breaks things in refactoring we quickly know it.

lib/ansible/module_utils/basic.py
```diff
@@ -1211,7 +1211,7 @@ def _check_required_arguments(self):
         missing = []
         for (k,v) in self.argument_spec.items():
             required = v.get('required', False)
-            if required and k not in self.params:
+            if required and self.params.get(k) is None:
```
Qalthos: This should prevent a missing argument from skipping the required argument check

lib/ansible/module_utils/eos.py
```diff
@@ -24,6 +25,7 @@
     port=dict(type='int'),
     username=dict(required=True),
     password=dict(no_log=True),
+    ssh_keyfile=dict(type='path'),
```
privateip, line 28: need to add this argument to the doc fragment as well at utils/module_docs_fragments/eos.py

lib/ansible/module_utils/basic.py
abadger: This one is probably going to cause a (extremely cornercase) bug.  A user could specify `!!null` in the playbook to set the parameter to None.  That should satisfy the required parameter (since the user set the param). We're only going to run into this if the module in question also thinks that None is a valid value for the parameter and specifies required=True.  Taken all together, I think this will be very unlikely to happen but I'm not sure we want to close out the possibility.

abadger:
Looks good except for the change to `_check_required_arguments()`.  take that one out and this looks good to merge.

lib/ansible/module_utils/basic.py
Qalthos: Thanks for the explanation, that clears it up a bit

PR #15152, opened 2016-03-25
Adding ip address to ansible-playbook RECAP

arxcruz:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
"ansible 2.1.0"
```
##### SUMMARY

**Before:**

```
localhost                   : ok=2    changed=1    unreachable=0    failed=0
```

**After:**

```
localhost - 127.0.0.1      : ok=2    changed=1    unreachable=0    failed=0
```

### DIFF
--- a/lib/ansible/executor/stats.py
+++ b/lib/ansible/executor/stats.py
@@ -22,7 +22,7 @@ __metaclass__ = type
 class AggregateStats:
     ''' holds stats about per-host activity during playbook runs '''
 
-    def __init__(self):
+    def __init__(self, inventory=None):
 
         self.processed = {}
         self.failures  = {}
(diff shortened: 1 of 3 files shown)


bcoca:
inventory can end up being a huge dataset, it is not something we should pass as a whole.

sivel:
I'm not convinced that this should be the default.

arxcruz:
@sivel So, do you think that it's better as a plugin?
@bcoca Yeah, you right, I can get set a dictionary on tqm only with the hosts.

What do you guys think? Let me know and I'll update it.

calfonso:
Hi!

- We don't want to pass the entire inventory as the default
- You can create a custom callback plugin to do what you want

PR #15163, opened 2016-03-25
Use _connection instead of _play_context for information about the connection

### LINKED ISSUE #3136: syncronize regression between 2.0.0.2 and 2.0.1.0
##### Issue Type:

- Bug Report
##### Plugin Name:

synchronize
##### Ansible Version:

```
ansible 2.0.1.0
  config file = 
```
##### Ansible Configuration:

  ANSIBLE_CONFIG = {toxinidir}/playbooks/ansible.cfg
##### Environment:

Both ubuntu-trusty and centos-7
##### Summary:

After upgrading to ansible 2.0.1.0 a regression was found with an existing role. 

```
TASK [windmill.nginx : Rsync nginx configuration.] *****************************
fatal: [jenkins01]: FAILED! => {"changed": false, "failed": true, "msg": "Could not determine controller hostname for rsync to send to"}
```
##### Steps To Reproduce:

Setup ubuntu-trusty or centos-7 VM

```
git clone https://github.com/openstack/windmill
cd windmill
tox -edeploy
```

That will provision your localhost and demo the breakage.

Here is the role being used: https://github.com/openstack/ansible-role-nginx/blob/master/tasks/config.yaml#L33
##### Expected Results:

role to work as expected with 2.0.0.2

```
TASK [windmill.nginx : Rsync nginx configuration.] *****************************
changed: [jenkins01]
```

http://logs.openstack.org/30/280430/7/check/gate-windmill-deploy-centos7-nv/184cc1f/console.html#_2016-02-18_16_38_49_774
##### Actual Results:

```
TASK [windmill.nginx : Rsync nginx configuration.] *****************************
fatal: [jenkins01]: FAILED! => {"changed": false, "failed": true, "msg": "Could not determine controller hostname for rsync to send to"}
```

http://logs.openstack.org/81/285081/2/check/gate-windmill-deploy-centos7-nv/1e84964/console.html#_2016-02-28_21_01_55_143

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0 (devel f4eb9aac24) last updated 2016/03/25 09:24:38 (GMT -700)
  lib/ansible/modules/core: (devel 0268864211) last updated 2016/03/25 07:41:37 (GMT -700)
  lib/ansible/modules/extras: (devel 6978984244) last updated 2016/03/25 07:41:46 (GMT -700)
```
##### SUMMARY

If we're not delegating then we change _connection into a local
connection midway through the file but we don't change
_play_context.connection (no need to alter that).  When we later check
it in process_remote() we need to know the actual connection, not the
connection that we thought we were going to use at the start of run().
So we have to use _connection.transport in process_remote().  The rest
of the places could use either one (because we have not yet changed to
a local connection) but we go ahead and switch those to
_connection.transport as well to avoid confusion in the future.

Fixes https://github.com/ansible/ansible-modules-core/issues/3136

With inventory:

```
zm01 ansible_host=localhost
```

Before:

```
$ ansible -m synchronize -a 'src=/etc/passwd dest=/var/tmp/te' -i playbooks/inventory zm01
zm01 | FAILED! => {
    "changed": false, 
    "failed": true, 
    "msg": "Could not determine controller hostname for rsync to send to"
```

After:

```
$ ansible -m synchronize -a 'src=/etc/passwd dest=/var/tmp/te' -i playbooks/inventory zm01
zm01 | SUCCESS => {
    "changed": false, 
    "cmd": "/usr/bin/rsync --delay-updates -F --compress --archive --rsh '/usr/bin/ssh  -S none -o StrictHostKeyChecking=no' --out-format='<<CHANGED>>%i %n%L' \"/etc/passwd\" \"/var/tmp/te\"", 
    "msg": "", 
    "rc": 0, 
    "stdout_lines": []
```

### DIFF
--- a/lib/ansible/plugins/action/synchronize.py
+++ b/lib/ansible/plugins/action/synchronize.py
@@ -81,7 +81,7 @@ class ActionModule(ActionBase):
             is a different host (for instance, an ssh tunnelled port or an
             alternative ssh port to a vagrant host.)
         """
-        transport = self._play_context.connection
+        transport = self._connection.transport
         if host not in C.LOCALHOST or transport != "local":
             if port_matches_localhost_port and host in C.LOCALHOST:
                 self._task.args['_substitute_controller'] = True
@@ -144,13 +144,13 @@ class ActionModule(ActionBase):
 
         result = super(ActionModule, self).run(tmp, task_vars)
 
-        # self._play_context.connection accounts for delegate_to so
+        # self._connection accounts for delegate_to so
         # remote_transport is the transport ansible thought it would need
         # between the controller and the delegate_to host or the controller
         # and the remote_host if delegate_to isn't set.
 
         remote_transport = False
-        if self._play_context.connection != 'local':
+        if self._connection.transport != 'local':
             remote_transport = True
 
         try:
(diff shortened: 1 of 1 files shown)


abadger:
Discussed with @jimi-c yesterday and it seems conceptually like this is fine.  However, jimi-c thought he remembered us doing things this way before and it had bad consequences.  I'm going to merge this to devel for now (what will be 2.1.x) but not merge it to 2.0.x.  That way it can see a lot more testing.

A workaround does exist in https://github.com/ansible/ansible-modules-core/issues/3136 so people can use that if they're dealing with it in 2.0.x.  If there are a large number of other people suffering from this we can look at backporting it to 2.0.x at a later time.

bcoca:
in this case it should not matter as you already have resolved them and loaded the connection based on the play_context value, iirc the issue James remembers was using this in action plugins in general.

abadger:
Sounds good.  based on that, I've pushed this to stable-2.0 as well.

PR #15164, opened 2016-03-25
Updates the community code of conduct

### LINKED ISSUE #40: Code of Conduct update
Ours is somewhat outdated. Should consider the Fedora CoC as a base:

https://getfedora.org/code-of-conduct

ghost:
##### ISSUE TYPE
- Docs Pull Request
##### ANSIBLE VERSION

```
ansible 2.0.1.0
  config file = /Users/abowles/.ansible.cfg
```
##### SUMMARY

Fixes https://github.com/ansible/community/issues/40 - this is a documentation update, so command output is not applicable.

### DIFF
--- a/docsite/rst/community.rst
+++ b/docsite/rst/community.rst
@@ -319,17 +319,16 @@ participation in all Ansible-related forums and mailing lists, code and document
 contributions, public IRC channels, private correspondence, and public meetings.
 
 Ansible community members are...
-================================
 
 Considerate
------------
-Contributions of every kind have far-ranging consequences. Just as your work depends on
+...........
+Contributions of every kind have far.ranging consequences. Just as your work depends on
 the work of others, decisions you make surrounding your contributions to the Ansible
 community will affect your fellow community members. You are strongly encouraged to take
 those consequences into account while making decisions.
 
 Patient
--------
+.......
 Asynchronous communication can come with its own frustrations, even in the most responsive
 of communities. Please remember that our community is largely built on volunteered time,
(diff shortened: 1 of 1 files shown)


bcoca:
cc @gregdek @robynbergeron @docschick

alikins:
Seems like this could be a good time to move the code of conduct to it's own top level page in the docs (instead of as a subsection of the general community page). 

That would also provide a small degree of separation between the slightly informal community page and the slightly more formal code of conduct. That would also make it a little easier to follow the history of the code of conduct since it wouldn't be intertwingled with the changes to the general community doc. 

[This pr comment is in no way a yay/nay on this pr or it's content, just a related suggestion.]

docschick:
sounds like a good idea to me

amenonsen:
I also think this should be in its own file.

gregdek:
Top notch changes @bwlz. Thanks so much for taking this on.

First order of business before we can merge is to set up the email address for codeofconduct@ansible.com; I'll get that done asap. Once that's done, we'll merge this, so we've got something in the immediate short term.

After that, we'll figure out where its permanent home is. That will take a little bit of discussion, but should be done in the next couple of weeks.

ghost:
Thanks @gregdek et al, appreciate the feedback. I'm happy to help with this when folks decide where the CoC should live permanently (since it sounds like we're all in general agreement that it should really have its own page).

gregdek:
@bwlz highly likely that we'll hang it off of this page directly:

https://www.ansible.com/community

cutecycle:
Awesome!

PR #15168, opened 2016-03-25
Script exists fix

bcoca:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
2.x
```

### DIFF
--- a/lib/ansible/plugins/action/__init__.py
+++ b/lib/ansible/plugins/action/__init__.py
@@ -92,6 +92,13 @@ class ActionBase(with_metaclass(ABCMeta, object)):
             )
         return results
 
+    def _remote_file_exists(self, path):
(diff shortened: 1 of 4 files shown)


lib/ansible/plugins/action/__init__.py
```diff
@@ -92,6 +92,14 @@ def run(self, tmp=None, task_vars=None):
             )
         return results
 
+    def _remote_file_exists(self, path):
+        cmd = self._connection._shell.exists(path)
+        result = self._low_level_execute_command(cmd=cmd, sudoable=True)
+        print(result)
```
abadger: debug statement that should be removed?

bcoca: thatwasonpurpose ....

PR #15175, opened 2016-03-26
Add function to convert CamelCased key names to snake_names

wimnat:
##### ISSUE TYPE
- Feature Pull Request
##### SUMMARY

Boto3 will always return its results with CamelCased dictionary key names.  In order to make it easy for new Ansible boto3 modules to deal with this, i've added a module_utils function that will recursivly substitute all CamelCase key names with snake_names

See https://groups.google.com/forum/#!searchin/ansible-devel/_snake/ansible-devel/saxf-cWt1VU/zwLScjpJAAAJ for discussion

### DIFF
--- a/lib/ansible/module_utils/ec2.py
+++ b/lib/ansible/module_utils/ec2.py
@@ -262,3 +262,24 @@ def paging(pause=0):
         return page
     return wrapper
 
+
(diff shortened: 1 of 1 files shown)


bcoca:
might need to put this in basic.py at one point, but not sure anyone needs it outside of aws yet

PR #15176, opened 2016-03-26
Split "all" into three subsections so that Travis can split tests

gundalow:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.0.1.0
```
##### SUMMARY

Experimentation with parallel builds with Travis

Will no doubt need some work before it can go live

This should allow Travis to run tests via the "Matrix" function

---

This change is [Reviewable](https://reviewable.io/reviews/ansible/ansible/15176)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,12 +5,32 @@ services:
 language: python
 matrix:
   include:
-    - env: TARGET=centos6
-    - env: TARGET=centos7 TARGET_OPTIONS="--volume=/sys/fs/cgroup:/sys/fs/cgroup:ro"
-    - env: TARGET=fedora23 TARGET_OPTIONS="--volume=/sys/fs/cgroup:/sys/fs/cgroup:ro"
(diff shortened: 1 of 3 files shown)


gundalow:
The statistics from Travis really don't make sense the elapsed time is sometimes less than the longest test time.

jimi-c:
Travis reports the total time used, which is obviously much greater than the actual time as things generally run in parallel.

gundalow:
I need to look at this more before before we merge this. Raised so people can look.

jimi-c:
@bob-smith sure, I'm also not sure it's speeding things up and may actually be slowing them down somewhat. The reason being that there is some overhead to Travis starting up the worker and configuring things (installing packages and pulling/starting up the docker container). However when run as one big blob as we do currently, that overhead is limited to just once as opposed to 4 times as this PR does.

gundalow:
As per agreement with @jimi-c this was useful investigation though not the right solution for the moment.

https://github.com/ansible/community/issues/47 has other details about Travis Performance

PR #15183, opened 2016-03-28
Some helper functions to work with boto3

wimnat:
##### ISSUE TYPE

- Feature Pull Request
##### SUMMARY

Some helper functions for working with boto3. See docstring for full detail.

func1 - convert boto3 list of dicts representing tags to flat dict of key:value pairs
func2 - inverse of above
func3 - convert ansible dict of filters to list of filters usable by boto3 for describe_xxx() requests

The first two should be useful for pretty much every boto3 module and the third will be useful for any boto3 _facts module.

Discussion: https://groups.google.com/forum/#!topic/ansible-devel/1gc9UG5yDZo

### DIFF
--- a/lib/ansible/module_utils/ec2.py
+++ b/lib/ansible/module_utils/ec2.py
@@ -283,3 +283,95 @@ def camel_dict_to_snake_dict(camel_dict):
         snake_dict[camel_to_snake(k)] = v
 
     return snake_dict
+
+
+def ansible_dict_to_boto3_filter_list(filters_dict):
+
+    """ Convert an Ansible dict of filters to list of dicts that boto3 can use
+    Args:
+        filters_dict (dict): Dict of AWS filters.
+    Basic Usage:
+        >>> filters = {'some-aws-id', 'i-01234567'}
+        >>> ansible_dict_to_boto3_filter_list(filters)
+        {
+            'some-aws-id': 'i-01234567'
+        }
+    Returns:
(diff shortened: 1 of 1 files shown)


linuxdynasty:
@wimnat, This would save some duplication in my code that I do practically for each module I write. Also loving the documentation on this.  One thing I would like to see in this PR, is a test for each function. I've tested your code and it does indeed work, but tests will make sure in the future that it always works.

You can take a look at the tests I wrote for the [kinesis_stream module](https://github.com/linuxdynasty/ld-ansible-modules/blob/master/test/cloud/amazon/test_kinesis_stream.py)

In my opinion this is ready to :ship: once it has some tests with it.
Example below..

```
    def test_make_tags_in_aws_format(self):
        example = {
            'env': 'development'
        }
        should_return = [
            {
                'Key': 'env',
                'Value': 'development'
            }
        ]
        aws_tags = kinesis_stream.make_tags_in_aws_format(example)
        self.assertEqual(aws_tags, should_return)

    def test_make_tags_in_proper_format(self):
        example = [
            {
                'Key': 'env',
                'Value': 'development'
            },
            {
                'Key': 'service',
                'Value': 'web'
            }
        ]
        should_return = {
            'env': 'development',
            'service': 'web'
        }
        proper_tags = kinesis_stream.make_tags_in_proper_format(example)
        self.assertEqual(proper_tags, should_return)
```

PR #15189, opened 2016-03-29
fix Mac chown/chmod -R issue, add error checks

nitzmahone:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0 (fix_mac_chown_chmod 124e4a22f8) last updated 2016/03/28 22:13:16 (GMT -700)
  lib/ansible/modules/core: (detached HEAD dfbf180be0) last updated 2016/03/28 22:08:24 (GMT -700)
  lib/ansible/modules/extras: (detached HEAD 33a557cc59) last updated 2016/03/28 22:08:24 (GMT -700)
```
##### SUMMARY

The changes to chown/chmod were broken on Mac (-R was being appended to the end of the command- OSX requires it before the file list).
A number of module setup commands were also blindly proceeding without checking for success. Added error raises for unrecoverable failure cases.

### DIFF
--- a/lib/ansible/plugins/action/__init__.py
+++ b/lib/ansible/plugins/action/__init__.py
@@ -308,11 +308,15 @@ class ActionBase(with_metaclass(ABCMeta, object)):
             # to work!
             if remote_user == 'root':
                 # SSh'ing as root, therefore we can chown
-                self._remote_chown(remote_path, self._play_context.become_user, recursive=recursive)
+                res = self._remote_chown(remote_path, self._play_context.become_user, recursive=recursive)
+                if res['rc'] != 0:
+                    raise AnsibleError('Failed to set owner on remote files (rc: {0}, err: {1})'.format(res['rc'], res['stderr']))
                 if execute:
                     # root can read things that don't have read bit but can't
                     # execute them.
-                    self._remote_chmod('u+x', remote_path, recursive=recursive)
+                    res = self._remote_chmod('u+x', remote_path, recursive=recursive)
+                    if res['rc'] != 0:
+                        raise AnsibleError('Failed to set file mode on remote files (rc: {0}, err: {1})'.format(res['rc'], res['stderr']))
             else:
                 if execute:
                     mode = 'rx'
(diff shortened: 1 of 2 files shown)


lib/ansible/plugins/shell/__init__.py
```diff
@@ -55,8 +55,10 @@ def path_has_trailing_slash(self, path):
     def chmod(self, mode, path, recursive=True):
         path = pipes.quote(path)
         cmd = ['chmod', mode, path]
+
         if recursive:
-            cmd.append('-R')
+            cmd.insert(1, '-R') # Mac chmod requires -R before file list
+
```
abadger: inserts are expensive.  Just add -R earlier than adding mode and path.

lib/ansible/plugins/shell/__init__.py
```diff
@@ -70,7 +72,7 @@ def chown(self, path, user, group=None, recursive=True):
             cmd = ['chown', '%s:%s' % (user, group), path]
 
         if recursive:
-            cmd.append('-R')
+            cmd.insert(1, '-R') # Mac chown requires -R before file list
```
abadger: Same thing here about adding -R earlier

lib/ansible/plugins/action/__init__.py
```diff
@@ -325,31 +325,35 @@ def _fixup_perms(self, remote_path, remote_user, execute=False, recursive=True):
...
             # Can't depend on the file being transferred with execute
             # permissions.  Only need user perms because no become was
             # used here
-            self._remote_chmod('u+x', remote_path, recursive=recursive)
+            self._remote_chmod('u+x', remote_path, recursive=recursive, raise_on_error=True)
 
         return remote_path
 
-    def _remote_chmod(self, mode, path, recursive=True, sudoable=False):
+    def _remote_chmod(self, mode, path, recursive=True, sudoable=False, raise_on_error=False):
```
abadger: Sorry, I think we had a misunderstanding when we talked earlier.  I don't think we should be raising unless we add more exceptions so that we can tell why an exception occurred (right now, we'd have to catch the AnsibleError and then parse the message to tell why we failed.)

raise_on_error is bad API.  The API should either raise whenever there is an error or let the caller discriminate.  Passing in a flag to tell the function to raise isn't meaningful.

If we start raising an error, then we have to audit the code and decide what the failure case means in the present code.  If the code doesn't depend on it (or works in some scenarios) then we probably have to replicate that behaviour instead of changing to always failing.

abadger:
I think we misunderstood each other on IRC.  The fix for recursive chown/chmod can be improved just a small amount and then is mergable.

The error raising shouldn't be done this way.  I was proposing auditing to make sure the _caller_ does the right thing for 2.1.x.  If we instead want to change the API so that the method that calls the command raises the error then we need to add an exception and audit the code to see whether it should handle the exception or let it bubble up (depending on whether the present code ever works when the command silently fails or if it is always an error that only becomes apparent later.)

Adding a raise_on_error argument to the methods is.... kinda silly.  We should just decide to do one of these or the other.  Adding the parameter is just making us do some of the work now and then do all of the work again later because we no longer trust that we really examined the code thoroughly the first time when we started using the parameter.

tonk:
This problem also applies to FreeBSD and therefore probably to more OS'es. So I think it would be best to have the options always in front of the arguments. This is the (maybe old-school) standard way for UNIX/Linux.

As it would require multiple `append`s to add the mode and path, will that be less expensive as an `insert`? I'm not sure about that.

nitzmahone:
@abadger Heh, I originally wrote it with append/extend (which I went back to here), but switched it to insert for minimal diff. Microperf, regardless, for such small lists, but happy to go back to the original impl, as it reads better that way too (the command is built in the proper order now). 

I know we said we didn't want to change  _low_level_execute_command to do the default error handling, but since these were new, I thought I'd try it that way. I'm a big fan of "don't repeat yourself", so it feels wrong to me to copy-paste the same check/raise to several places, but I also agree that raise_on_error is a lame API pattern. 

Anyway, the callers are all checking their own RCs now.

abadger:
Merged.  Thanks putting up with me ;-)

PR #15196, opened 2016-03-29
IAM role support for EC2 dynamic inventory

areian:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0
```

Tested on both `2.1.0` and `1.9.4`
##### SUMMARY

This change allows the EC2 dynamic inventory to assume an IAM role prior to connecting to the AWS modules.
We needed this functionality at my company due to how we have structured our AWS setup, and I thought others might find it useful.

### DIFF
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -132,6 +132,7 @@ from boto import ec2
 from boto import rds
 from boto import elasticache
 from boto import route53
+from boto import sts
 import six
 
 from ansible.module_utils import ec2 as ec2_utils
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -421,6 +421,10 @@ class Ec2Inventory(object):
         else:
             self.replace_dash_in_groups = True
 
+        # IAM role to assume for connection
+        if config.has_option('ec2', 'iam_role'):
+            self.iam_role = config.get('ec2', 'iam_role')
+
         # Configure which groups should be created.
         group_by_options = [
             'group_by_instance_id',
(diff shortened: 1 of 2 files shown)


defionscode:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;UPYq68tv&#x2F;6-iam-role-support-for-ec2-dynamic-inventory"> IAM role support for EC2 dynamic inventory</a>

MichaelBaydoun:
It would be great if we could do this for the entire playbook, and not just for ec2 dynamic inventory.

ryansb:
Looks good to me. 

@MichaelBaydoun do you mean for all the other AWS modules? IMO ec2.py is at least a little separate so I wouldn't block this pending IAM support for all other modules.

MichaelBaydoun:
@ryansb Agreed, I'm not suggesting holding this merge up, it's a good improvement.  Just floating the idea that down the road, it would be great if we had iam role assumption capability in the core, for all the modules.

contrib/inventory/ec2.py
```diff
@@ -475,6 +479,13 @@ def connect_to_aws(self, module, region):
             connect_args['profile_name'] = self.boto_profile
             self.boto_fix_security_token_in_profile(connect_args)
 
+        if self.iam_role:
```
ryansb, line 558: This needs a guard, since `self.iam_role` will be unset if the option isn't specified.

ryansb (changes requested):
Tweak on the `if` check, other than that looks good.

contrib/inventory/ec2.py
areian: It seemed more in line with the rest of the code to define the variable with a default value in `read_settings`.

ryansb (approved):

ryansb:
shipit

willthames:
shipit

nethershaw:
-1

This merge would cause regression of https://github.com/ansible/ansible/pull/15890.

willthames:
@nethershaw it doesn't seem to relate to the change in #15890, so might need some more detail as to how you identified that it would cause such a regression.

HontoNoRoger:
I tried your changes with current Ansible 2.2.0 with the dynamic inventory from the docs. Works as intended!

The only thing missing was
```python
from boto import sts
```
at the top of `ec2.py` script. Did I overlook that or is it still missing in the PR?

areian:
@HontoNoRoger You are absolutely correct. An oversight on my part when creating the PR. It has been fixed.

Will fix merge conflict later today

ryansb (approved):

PR #15217, opened 2016-03-31
extend env plugin to accept a default value as parameter

gmauro:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.0.1.0

```
##### SUMMARY

I have extended the env plugin to accept a parameter "default" as a default value if the environment variable is not set.

In this way,  an ansible variable can be defined as:
ANSIBLEVAR: "{{ lookup('env','ENVVAR', default='DEFAULTVALUE') }}"

### DIFF
--- a/lib/ansible/plugins/lookup/env.py
+++ b/lib/ansible/plugins/lookup/env.py
@@ -25,7 +25,7 @@ class LookupModule(LookupBase):
 
     def run(self, terms, variables, **kwargs):
 
-        default_v = kwargs.get('default', None)
+        default_v = kwargs.get('default', '')
         ret = []
         for term in terms:
             var = term.split()[0]
--- a/lib/ansible/plugins/lookup/env.py
+++ b/lib/ansible/plugins/lookup/env.py
@@ -25,9 +25,10 @@ class LookupModule(LookupBase):
 
     def run(self, terms, variables, **kwargs):
 
+        default_v = kwargs.get('default', None)
         ret = []
         for term in terms:
             var = term.split()[0]
-            ret.append(os.getenv(var, ''))
+            ret.append(os.getenv(var, default_v))
 
         return ret


bcoca:
I think this is redundant with the `default` filter.

Also, I do think the default should be `None`, but the fact is that this is not currently the case so this change is backwards incompatible.

gmauro:
I agree regards the backwards compatibility but I don't think is redundant with the default filter.
With default filter , I have to set default value every time I use the variable. Instead, extended lookup('env') permit to set variable, and his default value, just one time (for example in group_vars) and use it several.

group_vars/v.yml:

ANSIBLEVAR: "{{ lookup('env','ENVVAR', default='DEFAULTVALUE') }}"

roles/task/tasks/main.yml:
- name: set env variable 
  lineinfile: dest=/home/user/.profile line="export USER_ENV_VAR={{ ANSIBLEVAR }}"
- name: download a file
  get_url: url={{ ANSIBLEVAR }}/thing dest=/tmp

bcoca:
No you don't:

```
ANSIBLEVAR: "{{ lookup('env','ENVVAR')|default('defaultvalue', boolean=True) }}"
```

or use `ternary` as the '' current default can be confusing with other false values:

```
ANSIBLEVAR: "{{ lookup('env','ENVVAR') == '' |ternary('defaultvalue', lookup('env','ENVVAR')) }}"
```

gmauro:
I understood. 
Thank for your time

gimoh:
Some lookups already support a default value, I think it would be more intuitive if they all supported that, it mimics python's `dict.get()` which makes sense to me.  I mean, just looking at it, which is more intuitive?

```
vars:
  foo: "{{ lookup('env', 'FOO') | default('fff', true) }}"
  bar: "{{ lookup('env', 'BAR', default='bbb') }}"
```

To me it's `bar` hands down, it's the first thing I tried, it actually took a bit of hunting around the docs and a couple of WTFs to figure out how to correctly use the default filter with lookup.

jctanner:
Hi!

We are interested in this idea and would like to see a wider discussion on it
on one of our lists.  Reasons for this include:

   * It would be nice to make the lookup plugins consistent in their use of defaults, but we should come to a community consensus. 

Can you please post on ansible-development list so we can talk about this idea with the wider group?

You can also submit a proposal for it using this process.

   * https://github.com/ansible/proposals/blob/master/proposals_process_proposal.md

PR #15221, opened 2016-03-31
refactor distribution detection in facts.py

### LINKED ISSUE #15087: Refactor facts.py os version detection
##### Issue Type:
- Bug Report
##### Ansible Version:

devel
##### Summary:

facts.py at the moment has up to 12 levels of indentation in the code that determines the os release. This is barly readble and should be cleaned up. see
https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/facts.py#L486

Also one could rewrite it a little to separate readline files in /etc/*release from the parsing such that one can have unit tests for the parsing.

I'd be happy to contribute if such a refactor is of interest to the core team.

robinro:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

devel
##### SUMMARY

This commit refactors the distribution version parsing in `module_utils/facts.py`. By doing so it fixes #15087. 

The old version had many duplications and got up to 12 levels of indentation. The refactor splits off methods for each distribution, reduces duplication and aims to improve readability.

I tried to keep the original logic as far as possible. The logic is rather weird in some places, which I now marked 'TODO', but don't want to change yet without being able to test them.
The overall idea is that a list of 'path, name' is iterated and `path` is read to get a distribution_version for `name` out of it. Whether or not the parsing is successful is not always obvious and sometimes the code assumes that even after parsing one file other files are still read. 

I also added a test_distribution_version.py that should be extended to test the correct parsing of `/etc/*release` files of different distributions. The tests are non-trivial as also calls to the module `platform` have to be mocked.

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -31,16 +31,42 @@ import struct
 import datetime
 import getpass
 import pwd
-import ConfigParser
+
+try:
+    # python2
+    import ConfigParser as configparser
+except ImportError:
+    # python3
+    import configparser
 from ansible.module_utils.basic import get_all_subclasses
 
 # py2 vs py3; replace with six via ziploader
 try:
+    # python2
     from StringIO import StringIO
 except ImportError:
+    # python3
     from io import StringIO
 
-from string import maketrans
+
+try:
+    # python2
+    from string import maketrans
+except ImportError:
+    # python3
+    maketrans = str.maketrans # TODO: is this really identical?
+
+try:
+    dict.iteritems
+except AttributeError:
+    # Python 3
+    def iteritems(d):
+        return d.items()
+else:
+    # Python 2
+    def iteritems(d):
+        return d.iteritems()
+
 
 try:
     import selinux
@@ -118,23 +144,6 @@ class Facts(object):
     _I386RE = re.compile(r'i([3456]86|86pc)')
     # For the most part, we assume that platform.dist() will tell the truth.
     # This is the fallback to handle unknowns or exceptions
-    OSDIST_LIST = ( ('/etc/oracle-release', 'OracleLinux'),
-                    ('/etc/slackware-version', 'Slackware'),
-                    ('/etc/redhat-release', 'RedHat'),
-                    ('/etc/vmware-release', 'VMwareESX'),
-                    ('/etc/openwrt_release', 'OpenWrt'),
-                    ('/etc/system-release', 'OtherLinux'),
-                    ('/etc/alpine-release', 'Alpine'),
-                    ('/etc/release', 'Solaris'),
-                    ('/etc/arch-release', 'Archlinux'),
-                    ('/etc/os-release', 'SuSE'),
-                    ('/etc/SuSE-release', 'SuSE'),
-                    ('/etc/gentoo-release', 'Gentoo'),
-                    ('/etc/os-release', 'Debian'),
-                    ('/etc/lsb-release', 'Mandriva'),
-                    ('/etc/altlinux-release', 'Altlinux'),
-                    ('/etc/os-release', 'NA'),
-                )
     SELINUX_MODE_DICT = { 1: 'enforcing', 0: 'permissive', -1: 'disabled' }
 
     # A list of dicts.  If there is a platform with more than one
@@ -158,7 +167,7 @@ class Facts(object):
                  { 'path' : '/usr/bin/pkg',         'name' : 'pkg' },
                  { 'path' : '/usr/bin/xbps-install','name' : 'xbps' },
                  { 'path' : '/usr/local/sbin/pkg',  'name' : 'pkgng' },
-    ]
+                ]
 
     def __init__(self, module, load_on_init=True):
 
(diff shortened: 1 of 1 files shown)


robinro:
The overall setup should probably be adjusted. At the moment I moved everything related to the distribution-guessing to a class `Distribution`. I don't yet see how to integrate this in the bigger picture behind the `FACT_SUBSETS` that @bcoca introduced recently, but I guess factoring out the distribution code as I did is a first step to any way of making `facts` more modular.

lib/ansible/module_utils/facts.py
```diff
@@ -157,32 +143,32 @@ class Facts(object):
...
 
         self.module = module
         self.facts = {}
+
         ### TODO: Eventually, these should all get moved to populate().  But
         # some of the values are currently being used by other subclasses (for
         # instance, os_family and distribution).  Have to sort out what to do
         # about those first.
-        if load_on_init:
-            self.get_platform_facts()
```
bcoca: don't remove this `if`

bcoca:
not sure that the change makes this better, specially the restructuring of the family distro lookup is now obscured and generated from the hardcoded lists when before it was just hardcoded itself.

lib/ansible/module_utils/facts.py
robinro: readded it. I didn't see it used when I touched that part, maybe because the branchpoint was too old before the rebase.

robinro:
The OS_FAMILY change is not really core to the refactor, so I reverted it. I found the old way of writing it very hard to read, but that might be just taste.

The main point is splitting up the `get_distribution_facts` method and making the individual parts more consistent.

@bcoca can you have a look at the `test_distribution_version.py`?
This should in principle also work with the old `get_distribution_version` implementation, but I'm unsure whether my use of mocking and calling the whole Facts class is still an appropriate use of the tests. If it is I would try to add more examples to get this as close to comprehensive as possible.

The main motivation for the refactor is the rather often occuring bugs with seldomly used distros (SuSE in my case), where some corner-case is lost due to the non-obvious code.

robinro:
The changes are now rebased/squashed to clean up the mess I went through to make Travis happy.

The test set is already quite long: Ubuntu 12.04, 14.04, Debian wheezy, jessie, stretch, openSuSE, SLES, CoreOS, ...
Independent of the refactor the test should really make it into devel. The test and facts.py is already python3 ready, but depends on `basic.py` which is not.

I had a look at the code to see whether more intrusive refactoring might be an option. I believe for the moment without more test coverage it's not. The distribution information is used in Facts(), so this has to be cleaned up first. Also the current setup depends on the order in which things are read. So I tried to split it up in small pieces and put some `TODO`s on the obscure bits. When we have test cases for them one could try a furthergoing refactor.

alikins:
The Distribution() class here looks good to me. A couple of thoughts:

(for discussion, not blocking...)
    \* It might be useful to pull all the mapping data out into a config/data file (OS_FAMILY, OSDIST_LIST, etc). 
        \* I wonder if there is a good upstream source for that info? Seems like it could be useful for many tools. If not, it may be useful to start one   
- A tiny potential general case optimization would be to sort OSDIST by popularity. Only suggested since the current list doesn't appear to have any ordering. I'd prefer that list be either sorted alphabetically for readability, or in rough order of commonness as mentioned above.

robinro:
@alikins Thanks for your comments
1. I'm not sure about having an external config; if doing so adds more lines than the data is long it's not worth the effort; it's probably more straight forward to make the whole thing a separate python module
2. There is some upstream I know of doing this, the cpython stdlib:
   https://github.com/python/cpython/blob/master/Lib/platform.py
   But it's a mess. The code in ansible is (in theory) only those parts that they get wrong. Their update cycle is quite slow and the coverage of distribution also not ideal.
   I also thought about making the whole thing a standalone python module. Maybe sometime...
3. Sorting OSDIST_LIST is dangerous, as some parts depend on the ordering (i.e. SuSE). If all the open TODOs in the code are fixed it should be ordering-independent. But at least the SuSE code needs some work for this.

robinro:
As discussed in https://meetbot.fedoraproject.org/teams/ansible_core_meeting/ansible_core_meeting.2016-04-14-15.50.log.html this is moved to 2.2.0. It should be included eary in the cycle.

abadger:
Okay, stable-2.1 has branched so we're ready to merge things to devel (2.2) now!  Feel free to rebase, cleanup merge commits, squash if you like, and then ping me on irc to merge this!

robinro:
rebased and squashed and ready to be merged

Sometime in the future one should extend the refactor to clean out all TODOs and make sure each file under /etc is read/parsed only once, but without more testcases there is a high probability to break some rare case. The general problem is that there is often redundant but conflicting information (both etc/suse-release and /etc/os-release exist with different content). I'm happy to help with that though if someone feels adventurous :roller_coaster:

Generating testcases is easy, just run https://github.com/ansible/ansible/blob/692bf51fde67d8aa29c9928bd5719cf53a05154b/hacking/tests/gen_distribution_version_testcase.py on the test machine (assumes python and ansible working). Or collect the necessary data manually.

abadger:
Thanks @robinro!  Yeah -- I think it makes sense to go forth and make possibly breaking changes during the first part of 2.2 development.  Gives users the maximum amount of time to encounter the problem in devel and us to fix it.

Maybe send a message out to the ansible mailing list letting people know what we're doing here and to test, to run the script and contribute the generated test cases, etc.

PR #15226, opened 2016-03-31
Add followup to argument_spec

Qalthos:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
2.1
```
##### SUMMARY

Allows parameters in argument_spec to pull fallback values from the environment, in addition to a fixed default value.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -516,6 +516,18 @@ def is_executable(path):
             or stat.S_IXOTH & os.stat(path)[stat.ST_MODE])
 
 
+class AnsibleFallbackNotFound(Exception):
+    pass
+
+def env_fallback(*args, **kwargs):
+    ''' Load value from environment '''
+    for arg in args:
+        if arg in os.environ:
+            return os.environ[arg]
+    else:
+        raise AnsibleFallbackNotFound
+
+
 class AnsibleModule(object):
     def __init__(self, argument_spec, bypass_checks=False, no_log=False,
         check_invalid_arguments=True, mutually_exclusive=None, required_together=None,
@@ -550,6 +562,7 @@ class AnsibleModule(object):
 
         self._load_constants()
         self._load_params()
+        self._set_fallbacks()
 
         # append to legal_inputs and then possibly check against them
         try:
@@ -1420,6 +1433,23 @@ class AnsibleModule(object):
                 # make sure things without a default still get set None
                 if k not in self.params:
                     self.params[k] = default
+
+    def _set_fallbacks(self):
+        for k,v in self.argument_spec.items():
+            fallback = v.get('fallback', (None,))
+            fallback_strategy = fallback[0]
+            fallback_args = []
+            fallback_kwargs = {}
+            if k not in self.params and fallback_strategy is not None:
+                for item in fallback[1:]:
+                    if isinstance(item, dict):
+                        fallback_kwargs = item
+                    else:
+                        fallback_args = item
+                try:
+                    self.params[k] = fallback_strategy(*fallback_args, **fallback_kwargs)
+                except AnsibleFallbackNotFound:
+                    continue
 
     def _load_params(self):
         ''' read the input and set the params attribute'''
--- a/lib/ansible/utils/module_docs_fragments/eos.py
+++ b/lib/ansible/utils/module_docs_fragments/eos.py
@@ -41,26 +41,31 @@ options:
       - Configures the usename to use to authenticate the connection to
         the remote device.  The value of I(username) is used to authenticate
         either the CLI login or the eAPI authentication depending on which
-        transport is used.
-    required: true
+        transport is used. If the value is not specified in the task, the
+        value of environment variable ANSIBLE_NET_USERNAME will be used instead.
+    required: false
   password:
     description:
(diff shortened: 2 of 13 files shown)


abadger:
Looks good.  I think that the documentation formatter won't pick up on the env_default: field so we'll have to figure out how to work that into the code.  I tested and the presence of env_default doesn't break docs building so I'm okay to merge this now and we can add env_default in a separate PR.

abadger:
The documentation formatter script is here: https://github.com/ansible/ansible/blob/devel/hacking/module_formatter.py  with supporting code inside of https://github.com/ansible/ansible/blob/devel/lib/ansible/utils/module_docs.py

abadger:
Link to discussion we had on Friday: https://gist.github.com/abadger/cb920b2baf9de25b743506121b3cf980

Working on getting a final decision on this in irc right now.

abadger:
Here's what we decided:

> abadger1999: Okay, so I think we're settling on new parameter named fallback.  takes a callable.  We provide an env var function that can be used to get the fallback from the environment.  required=True means that the default can come from either the task param or from the fallback.
> jimi|ansible: +1
> abadger1999: +1
> nitzmahone: abadger1999: sorry, I was in a sync-up meeting with some of the Windows community (trond/jhawkesworth)- yes, I'm still +1 to what we discussed last week

abadger:
So on the arg_spec side it should look like this:

```
    username=dict(required=True, fallback=(env_fallback, ['ANSIBLE_NET_USERNAME'])), 
```

Documentation we can probably update for now to spell things out in words like "username must be set either in the playbok task or in the ANSIBLE_NET_USERNAME environment variable on the target" and "ssh_auth_keyfile may be set in the playbook task or in the ANSIBLE_NET_SSH_KEYFILE environment variable on the target if it wasn't set in the playbook"

basic.py implementation will need to split  `_set_environment_defaults(self)` into two methods, one like `env_fallback()` that does the work of retrieving the value from the environment or raising a new, specific exception if the value doesn't exist and the other like `_set_fallback(self)` that does the generic portion of setting the value so that it does requires handling properly and doesn't override settings from the task params.

lib/ansible/module_utils/basic.py
```diff
@@ -516,6 +516,18 @@ def is_executable(path):
             or stat.S_IXOTH & os.stat(path)[stat.ST_MODE])
 
 
+class AnsibleFallbackNotFound(Exception):
```
Qalthos, line 519: I wasn't entirely sure where to put this. If someone has a better idea, I'd love to hear it.

lib/ansible/module_utils/basic.py
```diff
@@ -1421,6 +1434,15 @@ def _set_defaults(self, pre=True):
                 if k not in self.params:
                     self.params[k] = default
 
+    def _set_fallbacks(self):
+        for k,v in self.argument_spec.items():
+            fallback_strategy, fallback_parameters = v.get('fallback', (None,[]))
```
abadger: We probably want a third parameter for keyword args.  Some fallback methods might want the ability to use optional parameters.

lib/ansible/module_utils/basic.py
```diff
@@ -516,6 +516,18 @@ def is_executable(path):
             or stat.S_IXOTH & os.stat(path)[stat.ST_MODE])
 
 
+class AnsibleFallbackNotFound(Exception):
+    pass
+
+def env_fallback(*args):
```
abadger: Should add **kwargs (that aren't used by this fallback method but could be used by others) as well.

lib/ansible/module_utils/basic.py
abadger, line 519: I'm not sure.  Putting it in basic.py isn't bad for now.  basic is such a mess that in the future we'll probably refactor it into different modules anyway at some indeterminate future date.

lib/ansible/module_utils/basic.py
abadger: I suppose it's nice to people to using this in their modules if they can specify only args or only kwargs as well... Makes the code more complex, but we can do that with something like this:

``` python
fallback_entry = v.get('fallback',(None,))
fallback_strategy = fallback_entry[0]
fallback_args = []
fallback_kwargs = {}
for item in fallback_entry[1:]:
    if isinstance(item, dict):
        fallback_kwargs = item
    else:
        fallback_args = item
[...]
               self.params[k] = fallback_strategy(*fallback_args, **fallback_kwargs)
```

lib/ansible/utils/module_docs_fragments/eos.py
```diff
@@ -43,25 +43,29 @@ class ModuleDocFragment(object):
         either the CLI login or the eAPI authentication depending on which
         transport is used.
     required: true
+    env_default: ANSIBLE_NET_USERNAME
```
abadger: We'll need to change all this documentation before release of 2.1.0 (feature freeze on the 18th of April).  We can move it into the description or adjust the module formatter (but the table format with a short word might not be good for it since fallback is generic now... I guess description probably is the best place for it for now.)

abadger:
Looks generally good.  Had some comments on adding kwarg support to fallbacks and comment on the docs.

abadger:
Looks good.  +1 to merge.

PR #15241, opened 2016-04-01
Get rid of old v2 reference

dagwieers:
##### ISSUE TYPE

- Docs Pull Request
##### ANSIBLE VERSION

```
v2.0.2.0
```
##### SUMMARY

Old left-over reference to v2 sub-directory.


bcoca:
leave in as we still checkout older versions that have this in their tree and might push a v2/ structure by mistake as switching branches does not remove it.

PR #15246, opened 2016-04-01
Ziploader

abadger:
##### ISSUE TYPE
- Feature Pull Request
##### SUMMARY

Ziploader is a new way to construct modules for shipping over the wire from the controller to the managed machine.  It works for new-style python modules (all non-windows modules that are shipped with ansible).  All other module types continue to be processed by module replacer.

This pull request should be feature complete with regards to replacing module replacer.  It additionally has some slightly better error reporting than module replacer as module_utils tracebacks will display proper line number and file name with this code.  If ZIP_DEFLATED is selected as the compression type (the default) then the over-the-wire size of modules is also reduced at the cost of spending CPU compressing the modules.

ziploader is mostly backwards compatible with module replacer code.  These are the things that I've identified that would be backwards incompatible:
# Technically backwards incompatible changes
## ziploader changes
- module_utils code no longer has access to globals from another file.  This is
  due to not substituting the code into one big python module.  Modules
  themselves still have access to module_utils globals until explicitly changed
  because they use "import *"
  - signed off at meeting in early March.  jimi-c, bcoca.
## Changes made at the same time as ziploader as general module_common cleanup
- No longer substitute all REPLACER strings
  - Removed remnants of REPLACER_ARGS (not substituted in v2 code so not backwards incompatible)
  - In old-style and non_native_want_json modules, do no REPLACER substitution.
    - jimi-c says this is how it was supposed to be but the code didn't reflect
      it before, March 29
  - In new-style JSONARGS modules do not substitute
    `from ansible.module_utils.FOO import *`.  The new code decides that if
    we are using `from ansible.module_utils.FOO` then there's no reason to
    use JSONARGS (Note: JSONARGS modules never substituted REPLACER in the
    past and continues not to substitute it now).
  - Removed REPLACER_WINARGS (use REPLACER_JSONARGS instead)
    [x] signed off by nitzmahone, Mar 30
  - In new-style windows modules, only substitute REPLACER_WINDOWS
    (include powershell.ps1) and REPLACER_JSONARGS (arguments as a json
    string).  All others deemed to be python or *nix centric and therefore not
    needed for windows powershell modules.
  - In new-style python modules (determined by presence of REPLACER or
    "[..]import[..]ansible.module_utils[...]" only substitute REPLACER (with
    import \* from basic).  Some are deemed not needed and others were, added
    as constants:
    - REPLACER_COMPLEX: added to os.environ['ANSIBLE_MODULE_ARGS']
    - REPLACER_WINDOWS: windows only (substitution is powershell)
    - REPLACER_JSONARGS: not needed for new-style python scripts as they get
      args via ANSIBLE_MODULE_ARGS
    - REPLACER_VERSION: `from ansible import __version__` is the ziploader way to
      get this.
    - REPLACER_SELINUX: Added to `os.environ['ANSIBLE_MODULE_CONSTANTS']` as
      SELINUX_SPECIAL_FS.  changed from a string into a list.  See the
      ziploader section for more info.
    - syslog.LOG_USER: Added to `os.environ['ANSIBLE_MODULE_CONSTANTS']` as
      SYSLOG_FACILITY.  Changed from the syslog variable into a string.  See
      the ziploader section for more details.

The ziploader section mentioned above is here: https://gist.github.com/abadger/d3d1917cb9cb7b1cf31454f0c726e41b#file-gistfile1-txt-L278


abadger:
Also -- I'll be working on adding more features on top of this.  This PR is to get us a checkpoint at basic functionality so that we can revert to it if the new features don't pan out in time for 2.1 for whatever reasons.

lib/ansible/executor/module_common.py
```diff
@@ -50,6 +55,69 @@
...
+                raise Exception('Something wrong with this module zip file: should not contain absolute paths')
+            dest_filename = os.path.join(basedir, filename)
+            if dest_filename.endswith(os.path.sep) and not os.path.exists(dest_filename):
+                os.makedirs(dest_filename)
+            else:
+                directory = os.path.dirname(dest_filename)
+                if not os.path.exists(directory):
+                    os.makedirs(directory)
+                f = open(dest_filename, 'w')
+                f.write(z.read(filename))
```
nitzmahone, line 92: need to close the fd here (will using a 'with' on the fd break in 2.4 even if we don't execute this function?)

lib/ansible/executor/module_common.py
```diff
@@ -50,6 +55,69 @@
...
+import sys
+import base64
+import tempfile
+
+ZIPDATA = """%(zipdata)s"""
+
+def debug(command, zipped_mod):
+    # Okay to use __file__ here because we're running from a kept file
+    basedir = os.path.dirname(__file__)
+    if command == 'expand':
```
nitzmahone: It works fine to collapse the expand and execute into one shot- makes interactive debugging easier. We could still support them independently too, but the pycharm thing works a lot easier if it's all in one.

nitzmahone:
If the couple things I mentioned get fixed, module debuggability on KEEP_REMOTE_FILES is still decent. I'll try out debugging modules natively w/ module_utils next week.

lib/ansible/executor/module_common.py
abadger, line 92: Now closed on the next line.

abadger, line 92: And yeah, with is a keyword being used in a way that couldn't be mistaken for a variable name so it will cause a syntaxerror on python-2.4.

lib/ansible/executor/module_common.py
abadger: We talked about htis on slack and decided that since expand/explode overwrites the files, we would either have to leave it as two commands or make the single command check for file existence and not overwrite.  Doing file existence checks seems a bit complex for what's supposed to be a short simple wrapper.  So we'll leave this as two commands.

lib/ansible/executor/module_common.py
```diff
@@ -50,6 +55,61 @@
...
+
+def debug(command, zipped_mod):
+    # Okay to use __file__ here because we're running from a kept file
+    basedir = os.path.dirname(__file__)
+    if command == 'explode':
+        import zipfile
+        z = zipfile.ZipFile(zipped_mod)
+        for filename in z.namelist():
+            if filename.startswith('/'):
+                raise Exception('Something wrong with this module zip file: should not contain absolute paths')
```
jimi-c, line 83: Should we be raising an Exception here, or returning some JSON/sys.exit(1) as a nice exit response?

abadger, line 83: We should be raising an Exception because this won't be sent back to ansible.

The debug() functionality is something I worked on with @nitzmahone last night.  We realized that it would be harder to debug modules via ANSIBLE_KEEP_REMOTE_FILES using ziploader because what you would end up keeping is this wrapper with the zipped payload in ZIPDATA.  So the debug() function contains two things to ameliorate that.  First: If you run the wrapper file with `explode` as a command line arg it will take the payload and extract it into the ansible temporary directory (where the module file was uploaded).  Second: If you run the wrapper file with `execute` as a command line arg it will run the module files that were exploded to disk rather than the zipped payload that's contained in the wrapper.

These two commands enable you to debug a module using the following steps:

```
$ ANSIBLE_KEEP_REMOTE_FILES=1 ansible localhost -m ping -a data=crash -vvv
<127.0.0.1> ESTABLISH LOCAL CONNECTION FOR USER: badger
<127.0.0.1> EXEC /bin/sh -c '( umask 77 && mkdir -p "` echo $HOME/.ansible/tmp/ansible-tmp-1459605113.37-203813258173
480 `" && echo "` echo $HOME/.ansible/tmp/ansible-tmp-1459605113.37-203813258173480 `" )'
<127.0.0.1> PUT /var/tmp/tmpvdVWjY TO /home/badger/.ansible/tmp/ansible-tmp-1459605113.37-203813258173480/ping
<127.0.0.1> EXEC /bin/sh -c 'LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8 LC_MESSAGES=en_US.UTF-8 /usr/bin/python /home/badger
/.ansible/tmp/ansible-tmp-1459605113.37-203813258173480/ping'
[...]
# Log into remote machine
$ /usr/bin/python /home/badger
/.ansible/tmp/ansible-tmp-1459605113.37-203813258173480/ping explode
Module expanded into: /home/badger/.ansible/tmp/ansible-tmp-1459605113.37-203813258173480/ansible
# Edit the files in /home/badger/.ansible/tmp/ansible-tmp-1459605113.37-203813258173480/ansible to introspect what is happening as the module runs.   Change the code to see if it will still run, etc.
$ /usr/bin/python /home/badger/.ansible/tmp/ansible-tmp-1459605113.37-203813258173480/ping execute
# The execute command runs the exploded files instead of the zipped payload.  So it runs the module with all of the changes you made in the step above.
```

abadger:
Talked to @jimi-c today and added more comments to the wrapper and implemented comment stripping of it.  This lets us make the wrapper more readable in source code but still maintain a small size for over the wire transfer.

abadger:
Okay, All found problems addressed.  Merging.  New abadger/ziploader branch will have recursive imports.

PR #15248, opened 2016-04-02
Fix async for old style modules and don't send an args file for new-style modules

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### SUMMARY

Discovered that async wasn't properly formatting arguments for old-style modules.  They need key=value but async was trying to send json.

Also made it so we don't send an argument file for new-style modules.  New-style modules have the arguments embedded inside of them so there's no need to send arguments as a separate file.

### DIFF
--- a/lib/ansible/plugins/action/async.py
+++ b/lib/ansible/plugins/action/async.py
@@ -22,7 +22,7 @@ import random
 
 from ansible import constants as C
 from ansible.plugins.action import ActionBase
-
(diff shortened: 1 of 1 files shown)


abadger:
This needs to have https://github.com/ansible/ansible-modules-core/pull/3358 in order to work.

nitzmahone:
:+1:

PR #15275, opened 2016-04-04
Update ConfigureRemotingForAnsible.ps1

Cryptophobia:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
Any Version
```
##### SUMMARY

Set $ForceNewSSLCert if the system has been syspreped and a new SSL Cert must be forced on the winRM listener when re-running this script. This is necessary when a CN name changes and the self-signed cert is no longer valid and winRM is not allowing a connection because of winRM SSL validation errors.

Added the ForceNewSSLCert option for changing SID and CN machines/images.

### DIFF
--- a/docsite/rst/intro_windows.rst
+++ b/docsite/rst/intro_windows.rst
@@ -217,6 +217,9 @@ Pass the -CertValidityDays option to customize the expiration date of the genera
 Pass the -SkipNetworkProfileCheck switch to configure winrm to listen on PUBLIC zone interfaces.  (Without this option, the script will fail if any network interface on device is in PUBLIC zone)
   powershell.exe -File ConfigureRemotingForAnsible.ps1 -SkipNetworkProfileCheck
 
+Pass the -ForceNewSSLCert switch to force a new SSL certificate to be attached to an already existing winrm listener. (Avoids SSL winrm errors on syspreped Windows images after the CN changes)
+  powershell.exe -File ConfigureRemotingForAnsible.ps1 -ForceNewSSLCert
+
 .. note::
    On Windows 7 and Server 2008 R2 machines, due to a bug in Windows
    Management Framework 3.0, it may be necessary to install this
--- a/examples/scripts/ConfigureRemotingForAnsible.ps1
+++ b/examples/scripts/ConfigureRemotingForAnsible.ps1
@@ -12,19 +12,26 @@
 # DOMAIN or PRIVATE zones.  Provide this switch if you want to enable winrm on
 # a device with an interface in PUBLIC zone.
 #
+# Set $ForceNewSSLCert if the system has been syspreped and a new SSL Cert
+# must be forced on the WinRM Listener when re-running this script. This
+# is necessary when a new SID and CN name is created.
+#
 # Written by Trond Hindenes <trond@hindenes.com>
 # Updated by Chris Church <cchurch@ansible.com>
 # Updated by Michael Crilly <mike@autologic.cm>
+# Updated by Anton Ouzounov <Anton.Ouzounov@careerbuilder.com>
 #
 # Version 1.0 - July 6th, 2014
 # Version 1.1 - November 11th, 2014
 # Version 1.2 - May 15th, 2015
+# Version 1.3 - April 4th, 2016
 
 Param (
     [string]$SubjectName = $env:COMPUTERNAME,
     [int]$CertValidityDays = 365,
     [switch]$SkipNetworkProfileCheck,
-    $CreateSelfSignedCert = $true
+    $CreateSelfSignedCert = $true,
+    [switch]$ForceNewSSLCert
 )
 
 Function New-LegacySelfSignedCert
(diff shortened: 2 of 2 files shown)


jhawkesworth:
Hi,

This looks like a useful addition, but I just tried this as follows:
    .\ConfigureRemotingForAnsible.ps1 -ForceNewSSLCert true
against an existing Server 2012 R2 vm running WMF 4.0 (which still had a valid self-signed cert) but got the following error and now winrm is no longer listening:

New-WSManInstance : A parameter cannot be found that matches parameter name 'Force'.
At C:\Users\Administrator\ConfigureRemotingForAnsible.ps1:176 char:110
- ... eSet $valueset -Force
-                    ~~~~~~
  - CategoryInfo          : InvalidArgument: (:) [New-WSManInstance], ParameterBindingException
  - FullyQualifiedErrorId : NamedParameterNotFound,Microsoft.WSMan.Management.NewWSManInstanceCommand

If this could be fixed I'd be very interested in this as the ability to force create new SSL Certs would be well worth having.

Many thanks,

Jon

Cryptophobia:
Hi Jon,

Thank you for testing this. I forgot to remove the `-Force` parameter which does not exist on the New-WSManInstance cmdlet which was left over from previous code that I had written. Can you test it again?

Cryptophobia:
I just tested it and it works. It forces a new cert on the winRM listener successfully.

jhawkesworth:
Hi @Cryptophobia  - sorry for delay - Just tried this I can confirm running

.\ConfigureRemotingForAnsible.ps1 -ForceNewSSLCert true

works against Server 2012 R2 and Server 2016 TP4

Cryptophobia:
@jhawkesworth Thanks for testing it! Hope it is useful for everyone!

jhawkesworth:
Was thinking about this a bit more last night...

ConfigureRemotingForAnsible.ps1 now has several options so it would be good to document them, although not sure if 'in line' with the rest of the ansible windows introduction is necessarily the best place.

Also the change looks fairly well isolated from the rest of the existing script but I'd still like it to be tested against other versions of windows before merging.

jhawkesworth:
Ah.  I just noticed the other options are documented in line (here: http://docs.ansible.com/ansible/intro_windows.html) so makes sense to document this in the same place. 

@Cryptophobia are you ok to add this to the documentation page?

Cryptophobia:
@jhawkesworth @nitzmahone Just added a commit to add this switch option to the documentation too. Thanks for pointing me to it!

docsite/rst/intro_windows.rst
```diff
@@ -217,6 +217,9 @@ Pass the -CertValidityDays option to customize the expiration date of the genera
 Pass the -SkipNetworkProfileCheck switch to configure winrm to listen on PUBLIC zone interfaces.  (Without this option, the script will fail if any network interface on device is in PUBLIC zone)
   powershell.exe -File ConfigureRemotingForAnsible.ps1 -SkipNetworkProfileCheck
 
+Pass the -ForceNewSSLCert switch to force a new SSL certificate to be attached to and already existing winrm listener. (Avoids SSL winrm errors on syspreped Windows images after the CN changes)
```
jhawkesworth: Thanks for tackling the documentation too.
I know I'm being super-fussy here but I think you mean 'to an already existing' not 'to and already existing'.?

Cryptophobia: Fixed. Thanks for taking a look!

jhawkesworth:
Just to say i would like a few others to try this on versions of windows that haven't been tested yet before this is merged.  Since the docs point straight to the latest development version of this, as soon as its been merged, the latest version could be being used, so want to be extra sure this is good to go.

Cryptophobia:
I just did a few more tests on a couple of additional vagrant boxes:
1. Windows 7 Professional x64 with Service Pack 1 (BoxName: ferventcoder/win7pro-x64-nocm-lite)
2. Windows 10 with Internet Explorer 11 (BoxName: modernIE/w10-edge)

I found one problem and pushed a commit to fix it. When creating certs on older Windows systems, we need to use `New-LegacySelfSignedCert` instead of `New-SelfSignedCertificate`. I added a commit that contains the logic for this.

Now it works with older systems like Windows 7. Glad I did these extra tests on these vagrant boxes.

I think since this is an extra switch we added, users will not know about it until they look at the documentation or the actual code, so I do not think it will break any existing systems that call this script. The switch has to be passed in order for it to change the behavior of the script. Otherwise when the switch is not passed, the script behavior and code executed remains exactly the same as before adding this switch.

jhawkesworth:
Thanks @Cryptophobia for additional testing and fixing for older windows versions.
I agree the changes you have made are fairly isolated from the rest of the script so should not affect existing users unless they start using the new command line option.

jhawkesworth:
Thanks @Cryptophobia for additional testing and fixing for older windows versions.
I agree the changes you have made are fairly isolated from the rest of the script so should not affect existing users unless they start using the new command line option.  

@nitzmahone - do you want any more testing done before this gets merged?

jhawkesworth:
Yay, thanks for the contribution and the merge.

PR #15277, opened 2016-04-05
Jinja2 Filter: xmlescape, for escaping special characters in xml file…

jmalacho:
##### ISSUE TYPE

- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0
  config file = 
```
##### SUMMARY

 Jinja2 Filter: xmlescape, for escaping special characters in xml files (common problem when templating passwords)

```
N/A (unless you want the entire out put of make non_destructive)
```

### DIFF
--- a/lib/ansible/plugins/filter/xmlescape.py
+++ b/lib/ansible/plugins/filter/xmlescape.py
@@ -1,0 +1,20 @@
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+from jinja2 import Environment, FileSystemLoader
+from ansible import errors
+from xml.sax.saxutils import escape
+ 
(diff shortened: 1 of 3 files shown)


sivel:
Can you explain how this is different than the built in jinja2 `escape` filter?

As far as I can tell they perform the same job.

http://jinja.pocoo.org/docs/dev/templates/#escape

jmalacho:
Hah, Embarrassing. I'm new to ansible/jinja2 and I've never noticed that full list of filters (I just did a ctrl-f on http://docs.ansible.com/ansible/playbooks_filters.html for xml, and when I didn't find it I added my own).  I'm assuming it is the same.

At least now I know how to build and test ansible

PR #15281, opened 2016-04-05
Regression fix: proper plugin configuration was reset

jpic:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

  lib/ansible/modules/core: (detached HEAD cf01087a30) last updated 2016/04/05 11:48:52 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD 204b4bab56) last updated 2016/04/05 11:48:52 (GMT +200)
  config file = /home/jpic/.ansible.cfg
##### SUMMARY

Before this patch, if config was ['/some/path'] then it would enter the
else block and config would be set to [].

The regression this patch fixes was introduced by 700db154.

### DIFF
--- a/lib/ansible/plugins/__init__.py
+++ b/lib/ansible/plugins/__init__.py
@@ -66,7 +66,7 @@ class PluginLoader:
 
         if config and not isinstance(config, list):
             config = [config]
-        else:
+        elif not config:
             config = []
 
         self.config = config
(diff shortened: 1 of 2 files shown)


jpic:
Added tests for other use cases that the code was handling in a separate commit.

Please let me know if you want them squashed !

bitmensch:
@jpic This fixes exactly the problem I currently have with my callback plugin not being loaded.

nitzmahone:
cc @bcoca

nitzmahone:
Yep- I slammed in a quickie fix to bcoca's original fix to keep working and broke the case the original fix was trying to fix. Ugh. Thanks for the tests- that oughtta keep this from happening again!

jpic:
Cool, thanks @benjixx for your feedback and @nitzmahone for merging :tada:

PR #15289, opened 2016-04-05
Optional Use of urllib3 for SNI verification

sivel:
##### ISSUE TYPE

Feature Pull Request
##### ANSIBLE VERSION

```
v2.1
```
##### SUMMARY

This change adds optional support to `urls.py` to utilize `urllib3.contrib.pyopenssl` for SNI verification.

If the deps are not installed, things work as they do now, except the error message mentions that if installed, you can do SNI verification on older python versions.

To activate this functionality, the following python modules would need to be installed:
1. urllib3 or requests with urllib3 vendored (although urllib3 is listed as the dep)
2. pyopenssl
3. ndg-httpsclient
4. pyasn1

### DIFF
--- a/lib/ansible/module_utils/urls.py
+++ b/lib/ansible/module_utils/urls.py
@@ -487,6 +487,33 @@ def RedirectHandlerFactory(follow_redirects=None, validate_certs=True):
     return RedirectHandler
 
 
+def build_ssl_validation_error(hostname, port, paths):
+    '''Inteligently build out the SSLValidationError based on what support
+    you have installed
+    '''
+
+    msg = [
+        ('Failed to validate the SSL certificate for %s:%s.'
+         ' Make sure your managed systems have a valid CA'
+         ' certificate installed.')
+    ]
+    if not HAS_SSLCONTEXT:
+        msg.append('If the website serving the url uses SNI you need'
+                   ' python >= 2.7.9 on your managed machine')
+        if not HAS_URLLIB3_SNI_SUPPORT:
+            msg.append('or you can install the `urllib3`, `pyopenssl`,'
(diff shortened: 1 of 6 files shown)


sivel:
This addresses the concern from https://github.com/ansible/ansible-modules-core/issues/3355 where SNI would not be achievable on RHEL6 or any OS with Python 2.6 or an older version of Python 2.7 such as Ubuntu.

sivel:
Adhoc command for testing:

```
ansible all -i localhost, -c local -m uri -a 'url=https://sni.velox.ch return_content=True'
```

In the return on a success you should also see:

```
Great! Your client </strong>[ansible-httpget] <strong> \nsent the following TLS server name indication extension\n
```

sivel:
This has received a +1 from both @jimi-c and @abadger on IRC.

abadger:
+1 -- I know you were hesitating to add more tests because it would add deps while testing.  There's no problem with that.  We have lots of deps for testing only (database modules, for instance).  Might want to install and uninstall the deps as part of testing since the deps are optional, though.

sivel:
> Might want to install and uninstall the deps as part of testing since the deps are optional, though.

Cool, I'll handle that.

sivel:
After hours of trying to figure out how to handle the deps across all of our docker images correctly, I have integration tests as part of the `uri` module, that test the variances in SNI verification.

Of note, I moved `test_uri` into destructive tests, as it is now installing and uninstalling packages.

I also am not running the "urllib3" tests on ubuntu1204 as there is no `python-urllib3` and that made things too messy trying to handle it.  We still get full tests on other older pythons with other OSes/versions such as CentOS6 and Ubuntu14.04.

abadger:
Looks good to merge!

sivel:
Ok, merging.

PR #15301, opened 2016-04-06
get ssh_keys without using distribution info

robinro:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

devel
##### SUMMARY

In `facts.py` the Facts class is very complicated and entangles many things. One aspect I propose is to factor out the parsing of linux distribution_version info in #15221

This factoring out is only complete, when the rest of _Facts_ does not use distribution information.

This change fills the _ssh_host_key_ variables without using distribution info, but by trying out the possible directories in the most common order. This could fail if someone has keys in one of the other directories, but I don't see a reason why this should happen.


robinro:
thanks

PR #15304, opened 2016-04-06
WIP: Travis to defend "make deb" and "make rpm"

gundalow:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.0.1.0

```
##### SUMMARY

As part of my ongoing effort to improve testing.

I recently noticed that `make deb` (or rpm) were broken even when Travis was green.

I've opted to only install the extra dependencies within the script (rather than in `.travis.yml`) so the installation doesn't slow down the other 11 Travis subjobs

This adds in approx 90 second test to ensure that we catch these issues as soon as they occur. Placed between functional and sanity tests so longest tests are run first (based on @abadger recent investigation) to get a performance boost.

Also taken the opportunity to get `test/utils/run_tests.sh` shellcheck (bash linter) clean (apart from one line)

An example of a recent issue that I needed to fix as the build was failing:
https://github.com/ansible/ansible-modules-extras/pull/1935

---

This change is [Reviewable](https://reviewable.io/reviews/ansible/ansible/15304)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,12 +5,12 @@ services:
 language: python
 matrix:
   include:
-    - env: TARGET=centos6
-    - env: TARGET=centos7 TARGET_OPTIONS="--volume=/sys/fs/cgroup:/sys/fs/cgroup:ro"
-    - env: TARGET=fedora23 TARGET_OPTIONS="--volume=/sys/fs/cgroup:/sys/fs/cgroup:ro"
-    - env: TARGET=fedora-rawhide TARGET_OPTIONS="--volume=/sys/fs/cgroup:/sys/fs/cgroup:ro"
-    - env: TARGET=ubuntu1204
-    - env: TARGET=ubuntu1404
+    - env: TARGET=centos6 MAKETARGETS="rpm"
+    - env: TARGET=centos7 TARGET_OPTIONS="--volume=/sys/fs/cgroup:/sys/fs/cgroup:ro" MAKETARGETS="rpm"
+    - env: TARGET=fedora23 TARGET_OPTIONS="--volume=/sys/fs/cgroup:/sys/fs/cgroup:ro" MAKETARGETS="rpm"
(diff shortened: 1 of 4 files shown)


gundalow:
@mckerrj Here are the tests I mentioned on IRC earlier in the week

gundalow:
Feedback from @sivel @bcoca and @abadger is to do the check inside the docker instance.

abadger:
+1 to the concept.  Saw one thing that we could improve:   Instead of installing the packages via the run_tests.sh script maybe we can add it to the docker images.  If so, that would speed up the tests a bit and not make us more prone to failures (if the distro's repositories weren't reachable).

/cc @jimi-c to find out how to change teh docker images.

gundalow:
https://github.com/ansible/ansible/pull/15692/files updates the docker images. Once that's been merged and live I'll update this PR to remove the calls to apt & yum.

Also based on IRC in ansible-devel comments I'll make the script throw a meaningful error if TARGET isn't set

abadger:
https://github.com/ansible/ansible/pull/15692 merged but the images haven't been rebuilt yet.  So this is close :-)

I think this PR is probably closer to being merged than the PR to enable suse and ubuntu16 tests so we probably want to remove the suse and ubuntu16 enabling lines from here and let those be done via https://github.com/ansible/ansible/pull/15936/

mattclay:
Agreed that we should wait for #15936 to enable the new distros.

Also, rather than having to hack `DOCKER_IMAGE` for local testing, how about something like:

```
DOCKER_IMAGE="${PREFIX-ansible/ansible:}${TARGET}"
```

gundalow:
I agree with the above two comments and will update accordingly. 

This should have been marked as WIP (due to the dependency on the Docker changes), have done so now.

mattclay:
The docker images have been updated.

gundalow:
I need to update this to use Shippable now that https://github.com/ansible/ansible/pull/16000 has been integrated.

PR #15311, opened 2016-04-06
Fix network modules for ziploader changes

Qalthos:
##### ISSUE TYPE
- Bugfix Pull Request
##### SUMMARY

ziploader (#15246) changed the behavior of _load_params such that it does not return the parameter dictionary, but rather sets it directly. This fixes the network modules so that they use this preset dictionary rather than the output of `super(...)._load_params()` (as there is now no output).

Additionally, some modules check for task-set parameters first, and others do not. This change standardizes this behavior to prefer task-set parameters.

### DIFF
--- a/lib/ansible/module_utils/eos.py
+++ b/lib/ansible/module_utils/eos.py
@@ -194,16 +194,15 @@ class NetworkModule(AnsibleModule):
         return self._config
 
     def _load_params(self):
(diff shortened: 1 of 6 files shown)


privateip:
:+1:

abadger:
Looks good.  Thanks @Qalthos and @privateip !

PR #15314, opened 2016-04-06
add win_reboot action

nitzmahone:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0 (detached HEAD 6c007afcfa) last updated 2016/04/06 12:05:21 (GMT -700)
  lib/ansible/modules/core: (win_reboot 8a3e2ec20f) last updated 2016/04/06 12:07:22 (GMT -700)
  lib/ansible/modules/extras: (detached HEAD 7c3999a92a) last updated 2016/04/04 08:31:53 (GMT -700)
  config file =
```
##### SUMMARY

Adds win_reboot action to wrap the several necessary tasks to reboot a windows box and wait for it to come back up into a single task.

Docs are at https://github.com/ansible/ansible-modules-core/pull/3376

### DIFF
--- a/lib/ansible/plugins/action/win_reboot.py
+++ b/lib/ansible/plugins/action/win_reboot.py
@@ -1,0 +1,140 @@
+# (c) 2016, Matt Davis <mdavis@ansible.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+# CI-required python3 boilerplate
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+from ansible.plugins.action import ActionBase
+from ansible.utils.boolean import boolean
+from ansible.utils.unicode import to_unicode
+from ansible.errors import AnsibleUndefinedVariable
+
+import socket
+import time
+import traceback
+
+from datetime import datetime, timedelta
+
+try:
+    from __main__ import display
+except ImportError:
+    from ansible.utils.display import Display
+    display = Display()
+
+class TimedOutException(Exception):
+    pass
+
+class ActionModule(ActionBase):
+    TRANSFERS_FILES = False
+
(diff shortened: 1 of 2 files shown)


nitzmahone:
@trondhindenes @jhawkesworth as promised - review/feedback appreciated!

lib/ansible/plugins/action/win_reboot.py
```diff
@@ -0,0 +1,134 @@
...
+                    display.debug("win_reboot: %s fail (expected), sleeping before retry..." % what_desc)
+                time.sleep(fail_sleep_sec)
+
+        raise TimedOutException("timed out waiting for %s" % what_desc)
+
+    def run(self, tmp=None, task_vars=None):
+        if task_vars is None:
+            task_vars = dict()
+
+        # TODO: allow override via task args
```
jhawkesworth: Looks like you have already done this TODO

nitzmahone: Heh, oh yeah- I should probably remove the TODOs as I do them...

albertux:
@nitzmahone 
In your file: lib/ansible/plugins/action/win_reboot.py add:

``` python
from __future__ import (absolute_import, division, print_function)
__metaclass__ = type
```

msheiny:
I have a kind of weird edge case that I'm not sure is worth addressing. Running Virtualbox `5.0.16 r105871` under Linux. I'm testing against a VM under NAT networking and connecting through a port forward. Upon reboot, Virtualbox keeps the forwarded port open even though it is closed on the guest VM. This leaves this task stuck checking for the port to close which it never will. I did a quick hack around this by adding an extra variable for me to test locally, `skip_shutdown_check = False` which just skips the raise_if_port_open check and then waits `pre_reboot_delay_sec+5`  seconds before trying the connection again. Yeah, kind of hacky but it fixed the problem for me and possible others testing against virtualbox NAT networking. 

``` python
        try:
            if not skip_reboot_check:
                self.do_until_success_or_timeout(raise_if_port_open, shutdown_timeout_sec, what_desc="winrm port down")
            else: time.sleep(pre_reboot_delay_sec+5)

            def connect_winrm_port():
                sock = socket.create_connection((winrm_host, winrm_port), connect_timeout_sec)
                sock.close()
```

nitzmahone:
@msheiny - one of the methods that I've considered for this (reflected in one of the FUTURE comments in the code) would probably address your situation nicely: sample the last boot time, then poll for it to change, ignoring failures, for up to `reboot_timeout_sec`. That optimizes it so you're not waiting any longer than you have to, as well as only making one "loop" in the code instead of three, but still ensuring that the machine did in fact reboot. You've pointed out another great reason for that behavior (I had a couple in mind already), so I'm thinking I'll just change it over to that.

jhawkesworth:
Just to say I have run this successfully against a Server 2012 R2 test vm.

nitzmahone:
May re-address later, but there are a couple of issues with switching to the high-level-only reboot, and I'd rather get it out there for the masses...

jhawkesworth:
Thanks for merging (I've started using this a bit already)!

PR #15327, opened 2016-04-07
Add "excommunicate" debug option

abadger:
##### ISSUE TYPE
- Feature Pull Request
##### SUMMARY

Some debuggers are easier to work with when we do everything in a single
process.  This debug option caters to that at the expense of being
different from what Ansible will actually do to invoke a module.

When we document this we should be clear that this shouldn't be used for
general purpose debugging and that some modules may show strange
"errors" when used with this.  Those won't be considered real bugs as
it's not how ansible really invokes the modules.


nitzmahone:
Works great!

Issue #14503, opened 2016-02-16
ERROR! failed to transfer file to root@centos67:/root\u0007/root/.ansible/tmp/ansible-tmp-1455628791.1-102858065322850/setup


velmuruganvelayutham:
I am able to ssh to the remote machine. Also I am able to sftp successfully..
I have even tried with Openssh specific settings as mentioned in the website. [http://docs.ansible.com/ansible/intro_configuration.html#id127](http://docs.ansible.com/ansible/intro_configuration.html#id127)

TASK [setup] *******************************************************************
<166.45.16.96> ESTABLISH SSH CONNECTION FOR USER: root
<166.45.16.96> SSH: EXEC ssh -C -vvv -o ControlMaster=auto -o ControlPersist=60s -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o User=root -o ConnectTimeout=10 -o ControlPath=/home/martha/.ansible/cp/ansible-ssh-%h-%p-%r -tt 166.45.16.96 'mkdir -p "$( echo $HOME/.ansible/tmp/ansible-tmp-1455628975.54-121125587875243 )" && echo "$( echo $HOME/.ansible/tmp/ansible-tmp-1455628975.54-121125587875243 )"'
<166.45.16.96> PUT /tmp/tmp_0NsCj TO /root/.ansible/tmp/ansible-tmp-1455628975.54-121125587875243/setup
<166.45.16.96> SSH: EXEC sftp -b - -C -vvv -o ControlMaster=auto -o ControlPersist=60s -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o User=root -o ConnectTimeout=10 -o ControlPath=/home/martha/.ansible/cp/ansible-ssh-%h-%p-%r '[166.45.16.96]'
[...]

PR #15337, opened 2016-04-08
Guard against a shell profile printing extraneous data

sivel:
##### ISSUE TYPE

Bugfix Pull Request
##### ANSIBLE VERSION

```
v2
```
##### SUMMARY

This change guards against extraneous data that a shell profile may output.  Specifically ansi escape sequences.  Often people have custom `PS1` or `PROMPT_COMMAND` or something related to `npm` in their `.bashrc` that prints escape sequences causing invalid paths such as:

```
\u001b[?25h\u001b[0G\u001b[K\u001b[?25h\u001b[0G\u001b[K/home/sivel/.ansible/tmp/ansible-tmp-1460128827.54-257813177082508/ping
```

This above is the result of having the following in `.bashrc` (taken from https://github.com/ansible/ansible/issues/14503#issuecomment-188814463):

```
which npm &> /dev/null && . <(npm completion)
```

The changes made here are to use the "calculated" `basefile` as a string we can split the stdout on.  Additionally, since we alerady have a function to strip leading non JSON lines from stdout, I have added `\n` to `print()` statements in `basic.py` so that we ensure that the stuff printed before the module output is removed.

Let me know what you think.  We could instead of adding the `\n` to the print statements, could adjust `_filter_leading_non_json_lines` to try and filter on the same line where the JSON starts midway, however since escape sequences may contain `[` it would make this difficult.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -136,10 +136,10 @@ except ImportError:
     try:
         import simplejson as json
     except ImportError:
-        print('{"msg": "Error: ansible requires the stdlib json or simplejson module, neither was found!", "failed": true}')
+        print('\n{"msg": "Error: ansible requires the stdlib json or simplejson module, neither was found!", "failed": true}')
         sys.exit(1)
     except SyntaxError:
-        print('{"msg": "SyntaxError: probably due to installed simplejson being for a different python version", "failed": true}')
+        print('\n{"msg": "SyntaxError: probably due to installed simplejson being for a different python version", "failed": true}')
         sys.exit(1)
 
 HAVE_SELINUX=False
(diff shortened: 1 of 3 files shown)


sivel:
Copying in comments from IRC for context and so that I remember to cherry pick:

```
12:16:41 <@abadger1999> sivel: Your change looks good.  If it passes tests, +1 to merge from me.
12:17:44 <@abadger1999> sivel: Since I cherry-picked the non-JSON stripping code, I think it's appropriate to cherry-pick for 2.1 as well.
```

topagae:
I'm running into this on older versions of Ansible, is there any work around because I CANNOT upgrade.

Fruglemonkey:
> I'm running into this on older versions of Ansible, is there any work around because I CANNOT upgrade.

Wrap the offending code in your bashrc (or even your entire bashrc) with this:

```
if [[ $- == *i* ]]; then
fi
```

Basically checks if your bashrc is being run interactively or not. I.E When ansible runs, it's doing so via a non-interactive shell on the remote machine, so the problematic code in the .bashrc won't execute.

PR #15340, opened 2016-04-08
Remove workaround for fixed bug.

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### SUMMAR

The bug where PluginLoader required objects to directly inherit from
base_classes has been fixed.  Remove workaround from this strategy
plugin   Also switched to using super so that we don't have to modify
all of hte code anytime something like that happens.

@ks888 @bcoca

### DIFF
--- a/lib/ansible/plugins/__init__.py
+++ b/lib/ansible/plugins/__init__.py
@@ -31,7 +31,7 @@ import warnings
 from collections import defaultdict
 
 from ansible import constants as C
-from ansible.utils.unicode import to_str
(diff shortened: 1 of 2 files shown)


abadger:
The commit  which fixed PluginLoader so we didn't need to directly inherit anymore was: 7ce130212fe34b1fa5c0753f0ca25869081f83fb

ks888:
Nice! LGTM :+1:

abadger:
@ks888 thanks for taking a look!

PR #15346, opened 2016-04-09
Add requirements to run the tests

rajatguptarg:
##### ISSUE TYPE
- Bugfix Pull Request
##### SUMMARY

This requirements are needed to run the tests.

### DIFF
--- a/test/utils/tox/requirements.txt
+++ b/test/utils/tox/requirements.txt
@@ -12,3 +12,6 @@ redis
 python-memcached
 python-systemd
 pycrypto
+pyyaml
+six
+requests


bcoca:
pyyaml is already required by ansible, six is not as it is vendored. 

requrests should be a requirement of winrm and only installed when that is used

rajatguptarg:
pyYAML is needed when you clone the repo and want to run the tests and you need to install at that time. So it is a requirement to run the test.

bcoca:
already documented in the install docs, which include 'running from source'

mattclay:
The recommended method of running Ansible tests is now via `ansible-test`, which uses the requirements files found here: https://github.com/ansible/ansible/tree/devel/test/runner/requirements

When running `ansible-test` with the `--tox`, `--docker` or `--requirements` options, the appropriate requirements will be installed for the test command being executed.

We no longer maintain `tox.ini` or the `test/utils/tox/` requirements files.

PR #15356, opened 2016-04-09
Fixes to the documentation build

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
- Docs Pull Request
##### SUMMARY

Various fixes to the docs build script:
- Could only have one alias before.  Subsequent aliases overrode the
  previous ones.  Now multiple aliases work.
- Fix BLACKLISTED_MODULES.   Previously, modules were listed in the
  generated documentation despite being blacklisted
- Deprecated modules form extras were showing the (E) tag and not the
  (D) tag. Reversed that now (Probably not necessary to also show the
  E tag).
- Sort the deprecated modules alphabetically in the Category docs as
  well as the list of all modules
- Optimization: Previously rendered the modules to rst twice once in all
  group and once in individual categories.  Fixed to only render them
  once.

**EDIT**: Forgot the main reason I started this: Fixed the generation of deprecated modules having a leading underscore where they weren't supposed to.  That cleans up the large number of warning messages during docs builds like this:

```
/srv/ansible/devel/docsite/rst/list_of_cloud_modules.rst:192: WARNING: toctree contains reference to nonexisting document u'_nova_keypair_module'
```

### DIFF
--- a/hacking/module_formatter.py
+++ b/hacking/module_formatter.py
@@ -19,6 +19,7 @@
 #
 
 from __future__ import print_function
+
 import os
 import glob
 import sys
@@ -28,6 +29,8 @@ import optparse
 import datetime
 import cgi
 import warnings
+from collections import defaultdict
+
 from jinja2 import Environment, FileSystemLoader
 from six import iteritems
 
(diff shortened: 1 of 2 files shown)


abadger:
Note: To test, I ran the build script in devel and this branch and then compared the rst.  Only changes between the two were the ones fixing bugs.

abadger:
bcoca reviewed over slack and suggested that that we balcklist everything in utils/helpers and async_wrapper so I've made that change.  He also suggested moving async_wrapper into the helpers directory.  I'll do that as a separate PR so that I can check that it doesn't cause any test failures.

PR #15359, opened 2016-04-10
Adding common utils module for new Azure modules.

chouseknecht:
##### ISSUE TYPE
- New Module Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0 (azure_common 6d5cfcc0de) last updated 2016/04/10 10:39:40 (GMT -400)
  lib/ansible/modules/core: (detached HEAD c52f475c64) last updated 2016/04/10 08:32:14 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 5abb914315) last updated 2016/04/10 08:23:55 (GMT -400)
  config file =
```
##### SUMMARY

Submitting the common utility module needed for the new Azure modules.

### DIFF
--- a/lib/ansible/module_utils/azure_rm_common.py
+++ b/lib/ansible/module_utils/azure_rm_common.py
@@ -1,0 +1,619 @@
+#
+# Copyright (c) 2016 Matt Davis, <mdavis@ansible.com>
+#                    Chris Houseknecht, <house@redhat.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+import ConfigParser
+import json
+import os
+import re
+import sys
+import copy
+
+from os.path import expanduser
+from ansible.module_utils.basic import *
+
+AZURE_COMMON_ARGS = dict(
+    profile=dict(type='str'),
+    subscription_id=dict(type='str', no_log=True),
+    client_id=dict(type='str', no_log=True),
+    secret=dict(type='str', no_log=True),
+    tenant=dict(type='str', no_log=True),
+)
+
+AZURE_CREDENTIAL_ENV_MAPPING = dict(
+    profile='AZURE_PROFILE',
+    subscription_id='AZURE_SUBSCRIPTION_ID',
+    client_id='AZURE_CLIENT_ID',
+    secret='AZURE_SECRET',
+    tenant='AZURE_TENANT',
+    ad_user='AZURE_AD_USER',
+    password='AZURE_PASSWORD'
+)
+
+AZURE_TAG_ARGS = dict(
+    tags=dict(type='dict'),
+    purge_tags=dict(type='bool', default=False),
+)
+
+AZURE_COMMON_REQUIRED_IF = [
+    ('log_mode', 'file', ['log_path'])
+]
+
+CIDR_PATTERN = re.compile("(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1"
+                          "[0-9]{2}|2[0-4][0-9]|25[0-5])(\/([0-9]|[1-2][0-9]|3[0-2]))")
+
+AZURE_SUCCESS_STATE = "Succeeded"
+AZURE_FAILED_STATE = "Failed"
+
+AZURE_MIN_VERSION = "2016-03-30"
+
+HAS_AZURE = True
+HAS_AZURE_EXC = None
+
+try:
+    from enum import Enum
+    from msrest.serialization import Serializer
+    from msrestazure.azure_exceptions import CloudError
+    from azure.mgmt.compute import __version__ as azure_compute_version
+    from azure.mgmt.network.models import PublicIPAddress, NetworkSecurityGroup, SecurityRule, NetworkInterface, \
+        NetworkInterfaceIPConfiguration, Subnet
+    from azure.common.credentials import ServicePrincipalCredentials, UserPassCredentials
+    from azure.mgmt.network.network_management_client import NetworkManagementClient,\
+                                                             NetworkManagementClientConfiguration
+    from azure.mgmt.resource.resources.resource_management_client import ResourceManagementClient,\
+                                                                         ResourceManagementClientConfiguration
+    from azure.mgmt.storage.storage_management_client import StorageManagementClient,\
+                                                             StorageManagementClientConfiguration
+    from azure.mgmt.compute.compute_management_client import ComputeManagementClient,\
+                                                             ComputeManagementClientConfiguration
+    from azure.storage.cloudstorageaccount import CloudStorageAccount
+except ImportError, exc:
+    HAS_AZURE_EXC = exc
+    HAS_AZURE = False
+
+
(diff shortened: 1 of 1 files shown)


lib/ansible/module_utils/azure_rm_common.py
```diff
@@ -0,0 +1,635 @@
+#!/usr/bin/python
```
bcoca: no shebang, this is for import not for execution

lib/ansible/module_utils/azure_rm_common.py
```diff
@@ -0,0 +1,635 @@
...
+    return result
+
+
+class AzureRMModuleBase(object):
+
+    def __init__(self, derived_arg_spec, bypass_checks=False, no_log=False,
+                 check_invalid_arguments=True, mutually_exclusive=None, required_together=None,
+                 required_one_of=None, add_file_common_args=False, supports_check_mode=False,
+                 required_if=None, supports_tags=True):
+
```
abadger: bcoca and I came to an agreement that **kwargs would be well used here to blackbox the arguments that are forwarded onto AnsibleModule.  So this signature should look like:

``` python
def __init__(self, derived_arg_spec, supports_tags=True, required_if=None, **kwargs):
```

lib/ansible/module_utils/azure_rm_common.py
```diff
@@ -0,0 +1,635 @@
...
+        self.module = AnsibleModule(argument_spec=merged_arg_spec,
+                                    bypass_checks=bypass_checks,
+                                    no_log=no_log,
+                                    check_invalid_arguments=check_invalid_arguments,
+                                    mutually_exclusive=mutually_exclusive,
+                                    required_together=required_together,
+                                    required_one_of=required_one_of,
+                                    add_file_common_args=add_file_common_args,
+                                    supports_check_mode=supports_check_mode,
+                                    required_if=merged_required_if)
```
abadger, line 129: ... and then the call to instantiate AnsibleModule can look like this:

``` python
self.module = AnsibleModule(argument_spec=merged_arg_spec, required_if=merged_required_if, **kwargs)
```

lib/ansible/module_utils/azure_rm_common.py
```diff
@@ -0,0 +1,635 @@
...
+        return changed, new_tags
+
+    def exec_module(self):
+        '''
+        Execute module implementation.
+
+        :return: None
+        '''
+        res = self.exec_module_impl(**self.module.params)
+        self.module.exit_json(**res)
```
abadger: A better pattern for this is to make exec_module an abstract method (ie: one that the subclasses must implement).  Then have the subclass call self.module.exit_json() directly.

In python-2.6+ you can use the abc library: https://docs.python.org/2/library/abc.html or for something quick you can just do:

``` python
def exec_module(self):
    raise NotImplementedError()
```

The subclasses then implement exec_module and do **not** call the base class's exec_module method.

lib/ansible/module_utils/azure_rm_common.py
```diff
@@ -0,0 +1,635 @@
...
+        try:
+            return self.rm_client.resource_groups.get(resource_group)
+        except CloudError:
+            self.fail("Parameter error: resource group {0} not found".format(resource_group))
+        except Exception, exc:
+            self.fail("Error retrieving resource group {0} - {1}".format(resource_group, str(exc)))
+
+    def _get_profile(self, profile="default"):
+        path = expanduser("~")
+        path += "/.azure/credentials"
```
abadger: trivial but you can just do:

``` python
path = expanduser("~/.azure/credentials")
```

so that you only allocate a string once instead of three times.  (+= creates a new string)

lib/ansible/module_utils/azure_rm_common.py
```diff
@@ -0,0 +1,635 @@
...
+        try:
+            config = ConfigParser.ConfigParser()
+            config.read(path)
+        except Exception, exc:
```
abadger: Not necessary now but to bear in mind for the future we're going to need to go through and change these to `except Exception as exc:`... that syntax will be compatible with python2.6+ (I saw that this module needs   python-2.7) and python3 which will be the big push starting with ansible 2.2.

(Tangential information -- for code that has to also be compatible with python < 2.6 we have a helper function: `module_utils.basic.get_exception` that can be used to get exception info.)

lib/ansible/module_utils/azure_rm_common.py
```diff
@@ -0,0 +1,635 @@
...
+                pass
+
+        if credentials.get('client_id') is not None or credentials.get('ad_user') is not None:
+            return credentials
+
+        return None
+
+    def _get_env_credentials(self):
+        env_credentials = dict()
+        for attribute, env_variable in AZURE_CREDENTIAL_ENV_MAPPING.iteritems():
```
abadger: Another py3 porting note for 2.2 and beyond: if you don't need to use `.iteritems()` (because the number of entries would be small) just use `.items()`.  If you do need to use iteritems because the items take up a lot of memory, in 2.2 we'll be adding ansible.module_utils.six which has an iteritems function to replace the dict method.

lib/ansible/module_utils/azure_rm_common.py
```diff
@@ -0,0 +1,635 @@
...
+        '''
+        try:
+            delay = 20
+            while not poller.done():
+                self.log("Waiting for {0} sec".format(delay))
+                poller.wait(timeout=delay)
+            return poller.result()
+        except Exception, exc:
+            self.log(str(exc))
+            raise Exception(str(exc))
```
abadger: It's almost always better to do this instead:

``` python
except Exception as exc:
    self.log(str(exc))
    raise
```

By using raise all by itself in the exception handler, the exaclt exception you caught will be continued up the stack.  raising the exception again like this:

```
raise exc
```

Will change the stack trace and doing it as you have it here will additionally change the exception type (which can be good in some circumstances but probably isn't what you want here since you'd just be changing it to the Exception base class.)

abadger:
Noticed that `log()` doesn't go through `ansible.module_utils.basic.AnsibleModule`s log methods.  So it's not using any of the methods that are supposed to help remove passwords and other secrets from log messages.  seems like we don't want that in production.  Looking at the module's PR to see what is being logged, it seems to mostly be static strings and some exception messages.  The static strings are fine but I don't know what's in the exception messages and I don't know what other people might end up using it for.  Maybe just have the AzureModule's log method call  AnsibleModule's log method.

abadger:
chouseknecht pointed out on IRC that the syslog log_mode does go through AnsibleModule.debug() (via the logging.Handler at the bottom of the file) so that's being sanitized.  However then we saw the log_mode parameter and such and that became a problem.  Current majority view is that this should migrate to into logging provided for all modules in module_utils.basic and that the params and non-syslog settings should be disabled for now.

lib/ansible/module_utils/azure_rm_common.py
chouseknecht: There's no need for kwargs.

lib/ansible/module_utils/azure_rm_common.py
chouseknecht, line 129: All references to kwargs have been removed.

lib/ansible/module_utils/azure_rm_common.py
chouseknecht: changed to items()

lib/ansible/module_utils/azure_rm_common.py
chouseknecht: Fixed

chouseknecht:
Removed logging.

lib/ansible/module_utils/azure_rm_common.py
chouseknecht: Changed

lib/ansible/module_utils/azure_rm_common.py
chouseknecht: Fixed

lib/ansible/module_utils/azure_rm_common.py
chouseknecht: Fixed

lib/ansible/module_utils/azure_rm_common.py
chouseknecht: Fixed.

PR #15372, opened 2016-04-11
Create remote temp directory with search permission for all users

n-st:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

Change added on top of https://github.com/ansible/ansible/commit/005dc84.
`ansible --version` after fix:

```
ansible 2.1.0 (remote-file-permission-fix 51dfd282d2) last updated 2016/04/11 22:37:18 (GMT +200)
  lib/ansible/modules/core: (detached HEAD c52f475c64) last updated 2016/04/11 22:28:26 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD 5abb914315) last updated 2016/04/11 22:28:26 (GMT +200)
  config file = 
```
##### SUMMARY

When a playbook uses the "become" feature to switch to a user other than the one that was used to connect, the temporary files required for playbook execution are made accessible to that user. However, the directory that holds the files was previously created with mode 0o700, so the "become" user couldn't access the files within regardless of their permissions.
Changing the directory mode to 0o711 allows all users to traverse the directory, which mitigates the missing access problem. Since the files contained in the directory will have access restrictions of their own, this doesn't compromise security.

### DIFF
--- a/lib/ansible/plugins/action/__init__.py
+++ b/lib/ansible/plugins/action/__init__.py
@@ -210,7 +210,7 @@ class ActionBase(with_metaclass(ABCMeta, object)):
         if self._play_context.become and self._play_context.become_user not in ('root', remote_user):
             use_system_tmp = True
 
-        tmp_mode = 0o700
+        tmp_mode = 0o711
 
         cmd = self._connection._shell.mkdtemp(basefile, use_system_tmp, tmp_mode)
         result = self._low_level_execute_command(cmd, sudoable=False)


abadger:
@bcoca I think I've traced this down to a different root cause:

When creating a temp dir for an unprivileged user to an unprivileged user we're supposed to create it as "system".  This ensures that the path it's created in will be accessible to the second unprivileged user.  When we don't do that we get into the situation we have here where the second unprivileged user is unable to access the directory in which the temporary files have been placed.  Right now this is failing to happen in all cases because we have this code in the shell module's mkdtemp:

``` python
        basetmp = self.join_path(C.DEFAULT_REMOTE_TMP, basefile)
        if system and (basetmp.startswith('$HOME') or basetmp.startswith('~/')):
            basetmp = self.join_path('/tmp', basefile)
```

if DEFAULT_REMOTE_TMP does not begin with $HOME or ~/ then we ignore that the calling code requested to use system.  I can't think of a reason that that would matter so I think we should just change the conditional to:

``` python
        if system:
            basetmp = self.join_path('/tmp', basefile)
```

Can you guys think of a reason that will make a difference?

abadger:
I talked with jimi-c on slack and he concurred that the check seems incorrect.  Went ahead and pushed the change and tested with the role that was given.  the directory seems to work with that.

@n-st Let me know on IRC if this piece is still broken after that commit.

bcoca:
actually i believe that was a way to try to track using the user's home which meant either there was no change in user permissions or the escalation was to a privileged user, I do not think removing this is correct.

abadger:
@bcoca  the default is system=False.  Which means it's the marker for the two things you're talking about -- that there's no change in user permissions or that escalation is to a privileged user.  A system tmp dir must be used in the other case, when we're escalating to an unprivileged user.  Thus system=True is the tracking mechanism for us escalating to an unprivileged user.  We have no reason to track whether a user's home dir is being used in this code to attempt track those states -- system itself is the mechanism to look at that.

Since the previous code would decide not to install into the system temp even though system=True was given it was definitely wrong.  The new code may have deficiencies as well -- if so we'll either have to address those (last night I added code to address the selection of /var/tmp that you thought was the reason code existed) or we'll have to throw an error that tells the user they can't set local_tmp (except to "/tmp" or "/var/tmp") and use become to an unprivileged user.

bcoca:
I was actually involved with the changes to this code, the /var/tmp was one of several cases this 'kindof' solved, I agree the original code was incorrect, but 'mostly worked'. IIRC system got added later on and there were issues with removing the 'home' handling. 

The 'right way'TM is to actually look at actual permissions (t), but this will require a remote check, looking for a '/tmp' string or w/o /home,~/,$HOME was the compromise.

abadger:
If we're looking at permissions, then the right way involves many remote_checks... We would have to figure out that all of the directories leading up to where we're creating the tempdir allow the second user to search the directory.

Also of note, for this code, merely checking for 'tmp' in the directory string is not enough.  a remote_tmp setting to a directory like `badger:badger -rwx------ /tmp/mytmp/` will break the nonprivileged=>nonprivileged use case.  (We create our temp directory inside of /tmp/mytmp with proper permissions but /tmp/mytmp itself blocks the second user).

PR #15373, opened 2016-04-11
Remote file chown: Don't rely on username to check for root privileges

n-st:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

Change added on top of https://github.com/ansible/ansible/commit/005dc84.
`ansible --version` after fix:

```
ansible 2.1.0 (remote-chown-fix 7b1373480e) last updated 2016/04/11 22:57:10 (GMT +200)
  lib/ansible/modules/core: (detached HEAD c52f475c64) last updated 2016/04/11 22:28:26 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD 5abb914315) last updated 2016/04/11 22:28:26 (GMT +200)
  config file = 
```
##### SUMMARY

The SSH username isn't a reliable way to check if we've got root privileges on the remote system (think "toor" on FreeBSD). Because of this check, Ansible previously tried to use the fallback solutions for granting file access (ACLs, world-readable files) even on systems where it had root privileges when the remote username didn't match the literal string "root".

Instead of running checks on the username, just try using `chmod` in any case and fall back to the previous "non-root" solution when that fails.

### DIFF
--- a/lib/ansible/plugins/action/__init__.py
+++ b/lib/ansible/plugins/action/__init__.py
@@ -311,18 +311,23 @@ class ActionBase(with_metaclass(ABCMeta, object)):
         if self._play_context.become and self._play_context.become_user not in ('root', remote_user):
             # Unprivileged user that's different than the ssh user.  Let's get
             # to work!
-            if remote_user == 'root':
-                # SSh'ing as root, therefore we can chown
-                res = self._remote_chown(remote_path, self._play_context.become_user, recursive=recursive)
-                if res['rc'] != 0:
-                    raise AnsibleError('Failed to set owner on remote files (rc: {0}, err: {1})'.format(res['rc'], res['stderr']))
+
+            # Try chown'ing the file. This will only work if our SSH user has
+            # root privileges, but since we can't reliably determine that from
+            # the username (think "toor" on FreeBSD), let's just try first and
+            # apologize later:
+            res = self._remote_chown(remote_path, self._play_context.become_user, recursive=recursive)
+            if res['rc'] == 0:
+                # Only continue with chmod if chown worked
                 if execute:
                     # root can read things that don't have read bit but can't
                     # execute them.
                     res = self._remote_chmod('u+x', remote_path, recursive=recursive)
                     if res['rc'] != 0:
                         raise AnsibleError('Failed to set file mode on remote files (rc: {0}, err: {1})'.format(res['rc'], res['stderr']))
+
             else:
+                # Chown'ing failed. We're probably lacking root privileges; let's try something else.
                 if execute:
                     mode = 'rx'
                 else:


abadger:
This is a good idea but falling back if we are privileged is a big change.  We don't want to do that if we can help it.  So we probably need to try the chown.  If it fails, then check if the username is root and fail if so.  If username is not root, then perform the fallbacks.

lib/ansible/plugins/action/__init__.py
```diff
@@ -311,18 +311,23 @@ def _fixup_perms(self, remote_path, remote_user, execute=False, recursive=True):
...
-                if res['rc'] != 0:
-                    raise AnsibleError('Failed to set owner on remote files (rc: {0}, err: {1})'.format(res['rc'], res['stderr']))
+
+            # Try chown'ing the file. This will only work if our SSH user has
+            # root privileges, but since we can't reliably determine that from
+            # the username (think "toor" on FreeBSD), let's just try first and
+            # apologize later:
+            res = self._remote_chown(remote_path, self._play_context.become_user, recursive=recursive)
+            if res['rc'] == 0:
+                # Only continue with chmod if chown worked
```
abadger, line 321: This comment is misleading -- If we're privileged, then we set the file to be executable.  If we're unprivileged we set the file to be both readable and executable (ie: there's a chmod in both pathways).

bcoca, line 321: @abadger read his comment, the remote user might have 'root' privileges but not be named 'root', 'toor' is a good example as it has been in FreeBSD for a while.

n-st, line 321: @abadger The behaviour of only setting `chmod u+x` if we're privileged is preserved, since it now depends on the successful `chown`, which will only succeed if we're privileged.

abadger:
The docstring for this function could also be more explicit about what is going on.  Maybe something like:

```
"""
We need the files we upload to be readable (and sometimes executable) by the user being sudo'd to but we want to limit other people's access (because the files could contain passwords or other
private information.  We achieve this in one of these ways:

* If no sudo is performed or the remote_user is sudo'ing to themselves, we don't have to change permisions.
* If the remote_user sudo's to a privileged user (for instance, root), we don't have to change permissions
* If the remote_user is a privileged user and sudo's to an unprivileged user then we change the owner of the file to the unprivileged user so they can read it.
* If the remote_user is an unprivieged user and we're sudo'ing to a second unprivileged user then we attempt to grant the second unprivileged user access via file system acls.
  * If granting file system acls fails we can set the file to be world readable so that the second unprivileged user can read the file.  Since this could allow other users to get access to private information we only do this ansible is configured with "allow_world_readable_tmpfiles" in the ansible.cfg
```

n-st:
@abadger I realise this might change the overall behaviour if the `chown` fails for reasons other than missing privileges, but the only accurate way to check for root privileges would be to run `id -u` on the remote system. Not sure if the check is worth this extra step…

We could use `ansible_user_uid` from the fact list, but if I'm not mistaken that's not guaranteed to be present either…

lib/ansible/plugins/action/__init__.py
abadger, line 321: @n-st right -- but that's obvious from the code (if rc == 0: chmod).  So it's not necessary to state that in a comment unless it means something more... like "If and only if chown succeeds do we want to set this file to be executable").  And that is not what we're doing here so that's why the comment is misleading.

abadger, line 321: @bcoca -- wrong comment... The multiline first comment is good.  It's the second comment that should be taken out.

n-st:
@abadger Should the doctring really reiterate the inner workings of the function? Normally, I would write documentation to explain only the prerequisites and intended outcome of a function, since a detailed functional description would be redundant and prone to becoming outdated by future code changes.

bcoca:
@n-st it is a tricky part of the code, i believe it should explain WHY we are doing it in such a way in case someone thinks of a 'clever' way to do otherwise after misinterpreting the code and it's intentions.

abadger:
@n-st I made a proposal that kept your fix but I think also preserved the correct behaviour...... Could you comment on that?

lib/ansible/plugins/action/__init__.py
n-st, line 321: @abadger So you'd prefer something like

```
-                # Only continue with chmod if chown worked
-                if execute:
-                     # root can read things that don't have read bit but can't
-                     # execute them.
+                # root can read things that don't have read bit but can't
+                # execute them, so we might need to set the execute bit if
+                # we're root. We just ran chown successfully, so apparently
+                # we are root.
+                if execute:
```

?
That does sound more useful than my original version.

abadger:
@n-st yeah, the docstring for this function needs to tell us (1) the reason that we have to do this (balancing security with the need for the sudo'd user to be able to read the file) (2) All the potential permutations between remote_user and become_user that we can encounter and what the final state of the files at the end of the function will be.  Otherwise we can end up with a file with secrets being world readable when it's not supposed to or alternately, ansible not able to read the module files when a way to make them readable exists.

I think my proposed docstring covers those and avoids delving into implementation as much as it can given those constraints.

n-st:
@abadger I agree that the docstring should state the reason for using this function, which your first paragraph explains nicely.
However the list of case distinctions shouldn't usually concern a user of the function, and in case it is ever of interest, it is already observable in the code itself.
Hence I'd suggest just replacing the current docstring with the first paragraph of your suggestion ("We need the files we upload to be readable (and sometimes executable) by the user being sudo'd to but we want to limit other people's access (because the files could contain passwords or other
private information).").

abadger:
@n-st so to wrap this up... We need to change the code.  As mentioned on IRC I think it's fine if we first try to chown and if that fails we check if the username is root and fail if so.  Otherwise go to the fallbacks.  It's not perfect but @bcoca mentioned that even id -u isn't perfect since some systems grant privileges to uid 1. :-(

It's simple to implement and it takes care of the failure case on freebsd that you're experiencing while not opening any new ways for files to become world readable so I think it's the best thing to merge for 2.1.x.  We can revisit and look at doing it better if there's more people who experience problems related to this.

abadger:
The docstring for the module -- It's my fault that it doesn't capture all the nuances of what the documentation says about when one of these temp file can become world readable so I'll just update it once we've merged your fix.  no need to worry overly much about that.

abadger:
Merged https://github.com/ansible/ansible/pull/15482 which should fix this.  Going to go ahead and close this bug.  @n-st feel free to contact me on IRC if the problem is still there.

PR #15377, opened 2016-04-12
[not a generic fix] W/R for locale independent wide-characters.

asmca:
##### ISSUE TYPE

- Bugfix Pull Request
##### ANSIBLE VERSION

```
- 2.0.1.0
```
##### SUMMARY

Force the encoding to utf8 if we have any non-utf8 characters even in the utf8 ps session.

```

This is not a generic fix, just w/r for those windows' wide-character users who hit similar problems

the orig. codes can resolve the problems of wide-characters when powershell active code is set to 65001(UTF8).

but we may still come into some scenario that we may use some wide-chars(CJK for example), for NIC name, file/folder names.. etc.

if we hit this problem, we can just modify/hardcode on the response.std_out/std_err,  to change encode to utf8 as the result.
```

### DIFF
--- a/lib/ansible/plugins/connection/winrm.py
+++ b/lib/ansible/plugins/connection/winrm.py
@@ -171,6 +171,9 @@ class Connection(ConnectionBase):
             self._connected = True
         if not self.shell_id:
             self.shell_id = self.protocol.open_shell(codepage=65001) # UTF-8
+            # self.shell_id = self.protocol.open_shell(codepage=936) # Chinese Simplified.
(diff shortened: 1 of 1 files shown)


bcoca:
cc @abadger @nitzmahone @jhawkesworth

nitzmahone:
Not sure there's really anything we can do with this- all codepoints can be represented properly in UTF8, so if a module is returning improperly-formatted characters, let's deal with it in the modules...

Issue #15215, opened 2016-03-31
Inventory script does not work with assumed roles from the command line


ayqazi:
##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
ec2 dynamic inventory script

##### ANSIBLE VERSION
ansible 2.0.1.0

##### OS / ENVIRONMENT
On Ubuntu Linux, and Amazon Linux on AWS

##### SUMMARY
I cannot assume a role on the command line by exporting the new temporary credentials as environment variables and run the inventory script.

I have my 'master' AWS account's credentials in ~/.aws/credentials, then I assume a role on another account, and export the AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_SECRET_TOKEN and AWS_SESSION_TOKEN values it gives me. When I run the inventory script (which I have updated from latest devel, as of now) I get:

```
ERROR: "Authentication error retrieving ec2 inventory.
 - AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY environment vars found but may not be correct
 - Boto configs found at '~/.aws/credentials', but the credentials contained may not be correct", while: getting EC2 instances
```

I am using the latest inv/ec2.py as of now.

Thanks

SydOps:
we need add `boto.sts` and use `boto.sts.STSConnection` to connect aws if `ROLE_ARN` is exist. 

let me see if I can add this function.

ozbillwang:
So several ways for dynamic ec2 host inventory.

#### Way 1, use setting in system `ENV`

```
AWS_ACCESS_KEY_ID
AWS_SECRET_ACCESS_KEY
AWS_SESSION_TOKEN  (option)
```

#### Way 2, read credentials from `.aws/credentials` or `.boto`

```
[default]
aws_access_key_id = XXXXXXXXXX
aws_secret_access_key = XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX

[profile marketingadmin]
aws_access_key_id = XXXXXXXXXX
aws_secret_access_key = XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
aws_security_token=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX (option)
```

[...]

PR #15378, opened 2016-04-12
contrib/inventory/ec2.py can read AWS credentials from ec2.ini

pwnall:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.0.1.0
  config file =
```
##### SUMMARY

The EC2 external inventory uses boto, which reads AWS credentials from a few hard-coded paths relative to the user's home directory.

This PR adds the ability to read AWS credentials from the `ec2.ini` file that accomplishes the EC2 inventory script. The advantage of this approach is that AWS credentials can be specified inside the Ansible project.

For example, I use a playbook to perform a MFA (multi-factor authentication) and generate an ec2.ini that contains the obtained AWS credentials, as well as a few other settings derived from the playbook. I don't want to have to modify any files outside my Ansible project.

I think this capability would be useful to others, and I hope you will consider merging it. Thank you very much for all your work on Ansible!

### DIFF
--- a/contrib/inventory/ec2.ini
+++ b/contrib/inventory/ec2.ini
@@ -165,3 +165,22 @@ group_by_elasticache_replication_group = True
 # A boto configuration profile may be used to separate out credentials
 # see http://boto.readthedocs.org/en/latest/boto_config_tut.html
 # boto_profile = some-boto-profile-name
+
+
+[credentials]
+
+# The AWS credentials can optionally be specified here. Credentials specified
+# here are ignored if the environment variable AWS_ACCESS_KEY_ID or
+# AWS_PROFILE is set, or if the boto_profile property above is set.
+#
+# Supplying AWS credentials here is not recommended, as it introduces
+# non-trivial security concerns. When going down this route, please make sure
+# to set access permissions for this file correctly, e.g. handle it the same
+# way as you would a private SSH key.
+#
+# Unlike the boto and AWS configure files, this section does not support
+# profiles.
+#
+# aws_access_key_id = AXXXXXXXXXXXXXX
+# aws_secret_access_key = XXXXXXXXXXXXXXXXXXX
+# aws_security_token = XXXXXXXXXXXXXXXXXXXXXXXXXXXX
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -158,6 +158,9 @@ class Ec2Inventory(object):
         # Boto profile to use (if any)
         self.boto_profile = None
 
+        # AWS credentials.
+        self.credentials = {}
+
         # Read settings and parse CLI arguments
         self.parse_cli_args()
         self.read_settings()
@@ -225,7 +228,7 @@ class Ec2Inventory(object):
         configRegions_exclude = config.get('ec2', 'regions_exclude')
         if (configRegions == 'all'):
             if self.eucalyptus_host:
-                self.regions.append(boto.connect_euca(host=self.eucalyptus_host).region.name)
+                self.regions.append(boto.connect_euca(host=self.eucalyptus_host).region.name, **self.credentials)
             else:
                 for regionInfo in ec2.regions():
                     if regionInfo.name not in configRegions_exclude:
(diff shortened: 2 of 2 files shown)


bcoca:
@defionscode

juliedavila:
I'd like to see a disclaimer that supplying credentials in the ini is not recommended. Also, it would be good to show a note that profiles are not supported with ini credentials as they are in boto or aws config files.

pwnall:
@defionscode I think I addressed your concerns. Can you please take another look?

In case you're wondering, I'm using this feature in conjunction with [sts_session_token](https://github.com/ansible/ansible-modules-extras/pull/1988) to generate an inventory file that contains MFA (multi-factor-auth) credentials. The generated file is `.gitignore`d, and protected by the same permission bits as SSH keys used to log into machines.

amenonsen:
I don't like this feature at all. Letting credentials proliferate outside the places they're conventionally expected to be is just asking for trouble, and having a warning that says "don't do the exact thing I just added code to support" is a pretty weak protection against mistakes. But I guess I was a bit too late in expressing my reservations.

juliedavila:
@amenonsen I don't either. Internally we decided to okay this, given a disclaimer, as other cloud inventory scripts allow for the storage of credentials within their configs.

ozbillwang:
@amegianeg  & @defionscode & @bcoca 

What happen for this commit, seems it has been merged to ansible coding, but I can see at least two people don't like it, and now plus me. 

We should never use ec2.ini for real AWS key, which is really bad idea. Remember file`ec2.ini` has to be saved to version control (git/svn) when using with `ec2.py`. 

For MFA, the reason by @pwnall is wrong as well.  `~/.aws/config` supports to enable MFA directly. 

Second, because of this commit, it conflicts my commits (#15451) which can fix the issue #15215.

We should vote to rollback @pwnall's commit.

pwnall:
@SydOps I think you missed "I don't want to have to modify any files outside my Ansible project." in my PR description. From a single user's perspective, `~/.aws/config` is global state. In some cases, changing this global state is undesirable. My project was designed so that it can be `git cloned`ed into a directory, and then its commands can be executed without changing anything outside that directory.

Also, in my setup, `ec2.ini` is generated dynamically when a MFA token is obtained. The file is in `.gitignore` so it will not end up in the repository. Asides from the MFA aspect, dynamically generating the file lets me restrict the inventory to regions that I care about, even though the set of regions I care about changes every few days.

Last, there is nothing in my PR preventing you from following the golden path of statically generating an `ec2.ini` and checking it in. This makes me think that your real issue with my PR is that it conflicts with your changes. I think that a more constructive approach would be to ask for help rebasing your commit. If you squash your commits and clean up your code (at the very least, I'd revisit `Not get profile properly`), and then ask nicely, I can try to help you rebase.

ozbillwang:
@pwnall 

First `~/.aws/config and ~/.aws/credentials` are standard for aws cli, `~/.boto` is standard for python/boto. Your way is creating new non-standard way which no one supports in other place. Is it worth to be added here to increase the script complexity and maintainability?

Second, `~/.aws/config` is flexible to support different projects with different profiles, your idea is only for your own project as shortcut, it is not for other, again it is wrong way to put key/password in ec2.ini, not in `~/.aws/config and ~/.aws/credentials`. 

Third, `ec2.ini` should be checked in, can not be ignored. It is very important to customise dynamics inventory result. Do you have to manually share your own `ec2.ini` to other developers in your team, and keep updating to your team separately, when this file need be changed?

So my recommends, you should spend time to study the usage about `~/.aws/config and ~/.aws/credentials` on how to manage your projects in proper way, and play with `ec2.ini` a lot that how many different results you can get with changing it.

ozbillwang:
@defionscode , @bcoca & @amegianeg 

We need review the change in `ec2.py` again. The feature @pwnall added didn't give big improvement for ansible which is only for his own project, but my change for assume_role is higher priority which gives a big help for `enterprise level` requirement and it is `must-have` feature for ec2 dynamic inventory.

PR #15380, opened 2016-04-12
Apache2 module integration test

robinro:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

devel
##### SUMMARY

Add integration test for apache2_module.

Tests (so far):
- whether the module works at all
- idempotency
- implicit: test it works on the distributions used for integration tests


test/integration/roles/test_apache2_module/tasks/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+
+- name: include only on supported systems
+  include: actualtest.yml
+  when: ansible_os_family in ['Debian', 'Suse']
```
gundalow, line 5: Why doesn't this work on all RHEL based OS?

robinro, line 5: RHEL (at least centos7, which I tested) does not have a2enmod/a2dismod

As far as I understand debian came up with it and opensuse also implemented it, but independently and with other arguments.

robinro:
@gundalow as you already had a look, can this be merged?

robinro:
thanks @resmo for the merge

PR #15385, opened 2016-04-12
mysql_*: enable mysql connection via unix socket

bjne:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION
- 2.0.1.0
- master
- any?
##### SUMMARY

module_utils/mysql.py sets host by default, which makes unix socket connections unavailable.
This patch fixes that issue, and is backwards compatible, as unix socket connection has never
worked with ansible, seems.

### DIFF
--- a/lib/ansible/module_utils/mysql.py
+++ b/lib/ansible/module_utils/mysql.py
@@ -36,9 +36,7 @@ except ImportError:
     mysqldb_found = False
 
 def mysql_connect(module, login_user=None, login_password=None, config_file='', ssl_cert=None, ssl_key=None, ssl_ca=None, db=None, cursor_class=None, connect_timeout=30):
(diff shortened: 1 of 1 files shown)


jjshoe:
looks good to me

bjne:
@Jmainguy can you look at getting this merged/backported?

Jmainguy:
Tested, works, fixes mysql_variables. 

:shipit:

abadger:
Thanks @bjne @jjshoe  and @Jmainguy  for the fix and testing.  Merged to devel.  Cherry-picked to stable-2.0 but it's likely to miss 2.0.2 (I believe we're going to get final out today.)  It will be in a 2.0.3 if we end up shipping one.

PR #15058, opened 2016-03-20
Move destructive git test to test_git_destructive.

mattclay:
##### Issue Type:

Bugfix Pull Request
##### Ansible Version:

```
ansible 2.1.0 (git-destructive-tests 95da1ea5fd) last updated 2016/03/20 15:08:51 (GMT -700)
  lib/ansible/modules/core: (detached HEAD d71b9bae89) last updated 2016/03/20 15:10:53 (GMT -700)
  lib/ansible/modules/extras: (detached HEAD d5030ae555) last updated 2016/03/20 15:10:54 (GMT -700)
  config file = 
  configured module search path = Default w/o overrides
```
##### Summary:

The test_git role is run from the non_destructive integration tests. However, the task "remove known_host files" in that role is destructive.

Correcting this would be easy if the git module supported SSH options for GlobalKnownHostsFile and UserKnownHostsFile instead of hard-coded paths: ansible/ansible-modules-core#3281

In the meantime, I've moved the tests depending on the destructive behaviour from test_git to test_git_destructive. I've also made the tests less destructive by backing up and restoring the known_hosts files. 
##### Example output:

Tested using the ubuntu1404 docker image.

`TEST_FLAGS='-t test_git_destructive' make destructive` without credentials has a play recap of:

```
PLAY RECAP *********************************************************************
testhost                   : ok=71   changed=3    unreachable=0    failed=0   
```

`TEST_FLAGS='-t test_git_destructive' make destructive` with credentials has a play recap of:

```
PLAY RECAP *********************************************************************
testhost                   : ok=75   changed=6    unreachable=0    failed=0   
```

PR #15400, opened 2016-04-13
move test_git to destructive

robinro:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

devel
##### SUMMARY

The integration test test_git remove the ssh_known_hosts for the user and globally, see:
https://github.com/ansible/ansible/blob/08b3dbcda3f28815af3adff168f67506395d3c17/test/integration/roles/test_git/tasks/main.yml#L89-L91

That is destructive!
It should thus be in destructive and not non_destructive.

### DIFF
--- a/test/integration/destructive.yml
+++ b/test/integration/destructive.yml
@@ -21,3 +21,4 @@
     - { role: test_zypper, tags: test_zypper}
     - { role: test_zypper_repository, tags: test_zypper_repository}
     - { role: test_uri, tags: test_uri }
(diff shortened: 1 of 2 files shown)


sivel:
I don't have the reference right now, but I saw a similar issue recently, and the decision was to fix the reason for removing ssh_known_hosts instead of moving these tests.

robinro:
This is a duplicate of #15058

PR #15410, opened 2016-04-13
adds support for netconf to junos shared module

privateip:
Netconf support is provided using the ncclient library and the shared
module depends on ncclient to be installed on the local Ansible host.  This
commit also adds changes to the netcfg library to handle receiving messages
over netconf.

To use netconf, specify transport=netconf for junos module.  Be sure that
netconf has been enabled on the remote device

### DIFF
--- a/lib/ansible/module_utils/junos.py
+++ b/lib/ansible/module_utils/junos.py
@@ -16,19 +16,48 @@
 # You should have received a copy of the GNU General Public License
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 #
+from distutils.version import LooseVersion
 
 from ansible.module_utils.basic import AnsibleModule, env_fallback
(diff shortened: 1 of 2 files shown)


lib/ansible/module_utils/junos.py
```diff
@@ -60,16 +69,106 @@ def connect(self, **kwargs):
             msg = 'failed to connecto to %s:%s - %s' % (host, port, str(exc))
             self.module.fail_json(msg=msg)
 
-    def send(self, commands):
+        if self.shell._matched_prompt.strip().endswith('%'):
+            self.send('cli')
+        self.send('set cli screen-length 0')
+
+    def send(self, commands, **kwargs):
         return self.shell.send(commands)
```
Qalthos, line 108: Should this inherit the

```
except ShellError:
    module.fail_json(msg='failed to run commands')
```

from junos_command, instead of losing the try block entirely?

PR #15417, opened 2016-04-14
Copy vaulted

bcoca:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
2.1
```
##### SUMMARY

This is building on top of #14079 that adds the ability of transparently using vaulted files in the copy module.


MikeCoutts:
This PR appears to automatically try and decrypt any vaulted files it finds, which might not be the desired behavior. I have a playbook that is protected by one password that sets up a Docker container by copying a series of files (playbook.yaml, vault.yaml, Dockerfile etc) into a working directory that the container then accesses as it is built and where the internal playbook.yaml run in that container uses a different password. The outer play now fails with fatal: [localhost]: FAILED! => {"failed": true, "msg": "Decryption failed"} as it is trying to decrypt vault.yaml using the other password when the desired behavior was to simply copy the file across as is. The ability to toggle the decrypt behavior of the copy module would have been nice to have here.

donovan:
I too would like the ability to use the copy module to copy vault encrypted files to a remote host without the transparent decryption kicking in. Would be great if this feature could be toggled.

PR #15419, opened 2016-04-14
add _load_params debug overrides for module args/file passed on cmdline

nitzmahone:
##### ISSUE TYPE
- Feature Pull Request
##### SUMMARY

Adds overrides to module_utils basic.py _load_params to allow module args to be passed via a filepath or JSON on cmdline (makes for easier debugging in many cases).

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -1435,11 +1435,23 @@ class AnsibleModule(object):
 
     def _load_params(self):
         ''' read the input and set the params attribute.  Sets the constants as well.'''
+        # debug overrides to read args from file or cmdline
+
         # Avoid tracebacks when locale is non-utf8
-        if sys.version_info < (3,):
-            buffer = sys.stdin.read()
+        if len(sys.argv) > 1:
+            if os.path.isfile(sys.argv[1]):
+                fd = open(sys.argv[1], 'rb')
+                buffer = fd.read()
+                fd.close()
+            else:
+                buffer = sys.argv[1]
+        # default case, read from stdin
         else:
-            buffer = sys.stdin.buffer.read()
+            if sys.version_info < (3,):
+                buffer = sys.stdin.read()
+            else:
+                buffer = sys.stdin.buffer.read()
+
         try:
             params = json.loads(buffer.decode('utf-8'))
         except ValueError:
--- a/test/units/module_utils/basic/test__log_invocation.py
+++ b/test/units/module_utils/basic/test__log_invocation.py
@@ -22,75 +22,62 @@ __metaclass__ = type
 
 import sys
 import json
-from io import BytesIO, StringIO
 
-from ansible.compat.six import PY3
-from ansible.utils.unicode import to_bytes
+from units.mock.procenv import swap_stdin_and_argv
 
 from ansible.compat.tests import unittest
 from ansible.compat.tests.mock import MagicMock
 
+
 class TestModuleUtilsBasic(unittest.TestCase):
-    def setUp(self):
-        self.real_stdin = sys.stdin
-        args = json.dumps(
+    @unittest.skipIf(sys.version_info[0] >= 3, "Python 3 is not supported on targets (yet)")
+    def test_module_utils_basic__log_invocation(self):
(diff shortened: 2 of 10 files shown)


lib/ansible/module_utils/basic.py
```diff
@@ -1435,11 +1435,22 @@ def _set_fallbacks(self):
...
+            if os.path.isfile(sys.argv[1]):
+                fd = open(sys.argv[1], 'rb')
+                buffer = fd.read()
+                fd.close()
+            else:
+                buffer = sys.argv[1]
+        # default case, read from stdin
         else:
-            buffer = sys.stdin.buffer.read()
+            if sys.version_info < (3,):
```
abadger, line 1450: We lost the comment about non-utf8 locales here.

nitzmahone, line 1450: Still there, just moved it down around the actual try/except

abadger, line 1450: Ah -- but the comment isn't about the try except.  It's about reading via the bytes interfaces here so that we get raw bytes in and then transform them to unicode ourselves later.  (ie: the comment is to explain why we use open([..] 'rb'), and stdin.buffer on PY3.

test/units/module_utils/basic/test__log_invocation.py
```diff
@@ -22,75 +22,62 @@
 
 import sys
 import json
-from io import BytesIO, StringIO
 
-from ansible.compat.six import PY3
-from ansible.utils.unicode import to_bytes
+from units.mock.procenv import swap_stdin_and_argv
```
abadger, line 26: Not sure if this is the best way to refer to the module... what happens if we don't run nosetests form the right directory?  Maybe relative imports are what we want to use here (since the unittest code is python2.6+).
I think it's `from ... import swap_stdin_and_argv`)

lib/ansible/module_utils/basic.py
nitzmahone, line 1450: got it, moved that back up

test/units/module_utils/basic/test__log_invocation.py
nitzmahone, line 26: I actually tried that first, but it gripes about relative imports in non-packages when I run the tests- damned if you do, damned if you don't... We're also doing this in a _bunch_ of places for DictDataLoader (not that that excuses it). I'd love to make the relative import work for all those- seemed like the right thing to do.

test/units/module_utils/basic/test_exit_json.py
```diff
@@ -23,39 +23,34 @@
...
+        self.fake_stream = self.stdout_swap_ctx.__enter__()
 
         self.module = basic.AnsibleModule(argument_spec=dict())
 
     def tearDown(self):
-        sys.stdout = self.old_stdout
-        sys.stdin = self.old_stdin
+        # since we can't use context managers and "with" without overriding run(), call them directly to clean up
+        self.stdin_swap_ctx.__exit__(None, None, None)
+        self.stdout_swap_ctx.__exit__(None, None, None)
```
nitzmahone, line 54: It's the exits() that are nasty, but I don't think I can call the generators directly anymore (with next()) as it appears they're munged by @contextmanager.

test/units/module_utils/basic/test_exit_json.py
```diff
@@ -23,39 +23,34 @@
...
     def setUp(self):
-        self.old_stdin = sys.stdin
         args = json.dumps(dict(ANSIBLE_MODULE_ARGS={}, ANSIBLE_MODULE_CONSTANTS={}))
-        if PY3:
-            sys.stdin = StringIO(args)
-            sys.stdin.buffer = BytesIO(to_bytes(args))
-        else:
-            sys.stdin = BytesIO(to_bytes(args))
+        self.stdin_swap_ctx = swap_stdin_and_argv(stdin_data=args)
+        self.stdin_swap_ctx.__enter__()
```
nitzmahone, line 42: This isn't _horrible_, but again, non-obvious- not sure if this is better or worse than overriding run().

abadger:
Alright!  Thanks for guiding this all the way to the finish line! ;-)

PR #15391, opened 2016-04-12
Expand git integration test

robinro:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

devel
##### SUMMARY

This contains additional tests to cover
- ansible/ansible-modules-core#3254 (i.e. the bug ansible/ansible#14954)
- ansible/ansible-modules-core#527

PR #15428, opened 2016-04-15
fix broken build, test git version before running test on depth

robinro:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

devel, 2.1.0
##### SUMMARY
#15391 added new integration tests for git in combination with the option `depth`. There is a bug in git versions older than `1.8.2` regarding this feature, which broke the tests on older distributions.

This PR checks the git version and only runs the critical tests on sufficiently new git versions.

### DIFF
--- a/test/integration/roles/test_git/tasks/main.yml
+++ b/test/integration/roles/test_git/tasks/main.yml
@@ -40,6 +40,11 @@
 - name: verify that git is installed so this test can continue
   shell: which git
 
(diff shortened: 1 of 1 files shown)


robinro:
@abadger this fixes the broken tests from yesterday on centos6/ubuntu12.04

abadger:
Thanks robinro!

PR #15434, opened 2016-04-15
Add port argument for bigsuds

caphrim007:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
>= ansible 1.9.2
```
##### SUMMARY

This patch adds the port argument as a valid parameter to the f5_spec.

This argument is supported in bigsuds version 1.0.4 and greater, so
this patch uses the **version** variable of the bigsuds module to
determine when the port value should be honored by the module.

The rationale for this patch is that on some platforms that VEs are deployed to, a port-forwarding configuration may be set up where you would want to connect to the forwarded port because you are unable to connect to the true port.

For example, a Vagrant instance of BIG-IP may forward a port from 10443 to 443, and your ansible playbooks are on the host, so it would need to access the BIG-IP from https://localhost:10443

### DIFF
--- a/lib/ansible/module_utils/f5.py
+++ b/lib/ansible/module_utils/f5.py
@@ -43,6 +43,7 @@ def f5_argument_spec():
         user=dict(type='str', required=True),
         password=dict(type='str', aliases=['pass', 'pwd'], required=True, no_log=True),
         validate_certs = dict(default='yes', type='bool'),
+        server_port = dict(type='int', default=443, required=False),
         state = dict(type='str', default='present', choices=['present', 'absent']),
         partition = dict(type='str', default='Common')
     )
@@ -57,12 +58,16 @@ def f5_parse_arguments(module):
         if not hasattr(ssl, 'SSLContext'):
             module.fail_json(msg='bigsuds does not support verifying certificates with python < 2.7.9.  Either update python or set validate_certs=False on the task')
 
-    return (module.params['server'],module.params['user'],module.params['password'],module.params['state'],module.params['partition'],module.params['validate_certs'])
+    return (module.params['server'],module.params['user'],module.params['password'],module.params['state'],module.params['partition'],module.params['validate_certs'],module.params['server_port'])
 
-def bigip_api(bigip, user, password, validate_certs):
+def bigip_api(bigip, user, password, validate_certs, port=443):
     try:
-        # bigsuds >= 1.0.3
-        api = bigsuds.BIGIP(hostname=bigip, username=user, password=password, verify=validate_certs)
+        if bigsuds.__version__ >= '1.0.4':
+            api = bigsuds.BIGIP(hostname=bigip, username=user, password=password, verify=validate_certs, port=port)
+        elif bigsuds.__version__ == '1.0.3':
+            api = bigsuds.BIGIP(hostname=bigip, username=user, password=password, verify=validate_certs)
+        else:
+            api = bigsuds.BIGIP(hostname=bigip, username=user, password=password)
     except TypeError:
         # bigsuds < 1.0.3, no verify param
         if validate_certs:
@@ -92,5 +97,3 @@ def fq_list_names(partition,list_names):
     if list_names is None:
         return None
     return map(lambda x: fq_name(partition,x),list_names)
-
-


caphrim007:
@mite can you look at this?

caphrim007:
@mhite can you look at this?

mhite:
I'm a little worried that you are doing exact string matches on version numbers. Would be nice if we presumed anything below 1.0.3 doesn't support port, and anything 1.0.4 and greater does. What do you think?

caphrim007:
good point. lemme fix that

caphrim007:
how about now. does >= 1.0.4 make more sense?

mhite:
Yes, better! Hopefully using ASCII sort order string comparison won't present any gotchas. (I don't think it will.)

mhite:
Actually, you'll need a different parameter name other than `port` as this is already used in various modules which leverage f5_argument_spec(). IE. bigip_pool_member, etc. Additionally, each module will need their `DOCUMENTATION =` section updated to reflect the new option. Messy :) but necessary once this is merged.

caphrim007:
Suggestions on what it should be called?

caphrim007:
bigip_port? I can update the documentation strings if this is merged

mhite:
Since the BIG-IP host parameter is `server`, I'm thinking `server_port` is more appropriate -- and not taken by any of the BIG-IP modules themselves. (You can always have an alias, too, in the argument spec to allow for `bigip_port` also.)

Sound good?

caphrim007:
Works for me! Lemme update the pr

caphrim007:
@mhite updated PR. also, I forgot to add it to the values returned by f5_argument_spec so I added that too.

mhite:
I think I uncovered a bigger issue, unfortunately. There was an old PR that move some of the common setup logic (well, mostly) into `ansible/lib/ansible/module_utils/f5.py`. When you changed `f5_parse_arguments` in that file, it basically impacted all F5 modules:

```
return (module.params['server'],module.params['user'],module.params['password'],module.params['state'],module.params['partition'],module.params['validate_certs'],module.params['server_port'])
```

Notice all the modules this unpack them:

```
$ grep f5_parse_arguments *
bigip_monitor_http.py:    (server,user,password,state,partition,validate_certs) = f5_parse_arguments(module)
bigip_monitor_tcp.py:    (server,user,password,state,partition,validate_certs) = f5_parse_arguments(module)
bigip_node.py:    (server,user,password,state,partition,validate_certs) = f5_parse_arguments(module)
bigip_pool.py:    (server,user,password,state,partition,validate_certs) = f5_parse_arguments(module)
bigip_pool_member.py:    (server,user,password,state,partition,validate_certs) = f5_parse_arguments(module)
bigip_virtual_server.py:    (server,user,password,state,partition,validate_certs) = f5_parse_arguments(module)
```

So basically all those have to also be updated in unison or else all the modules with throw `ValueError` when you try to run them.

`ValueError: too many values to unpack`

This brings up something I don't like which is the helper utility file is not in -extras, but in the normal ansible repo, so rev'ing them together is not possible AFAIK. I kind of hate that. Also, the bigip_facts module doesn't use this helper, so it's an outlier.

So sorry you kicked over a rock, but this brings up the bigger issue of how to best deal with fixing this issue long term. ;( Any thoughts?

caphrim007:
touche`

I think this is more of a fundamental problem with the way that the different modules receive the parameters given. By unpacking those arguments as they do, that function blocks the future expansion of common variables that can be added....sorta.

We cannot change the underlying data without affecting all of the modules that use it. A more appropriate return value may be something like a dict or plain object that itself is populated by the module.param.get method.

I see that we also fail in the parse method if bigsuds is not found which, as mentioned offline, is going to not work when only a REST approach is used

So lemme add a workaround that takes both your comments into account and I'll post an updated PR! Thanks for the fantastic testing!

mhite:
I have a few thoughts on addressing this long term:
- Can the common library live in extras, and not the main repo? [I don't think so, but I'm not sure.]
- Can we just coordinate this update along with all the necessary module updates?
- If we are going to do that, perhaps we should revisit the data structure returned to be more 'update' friendly. IE. Have it return a dictionary or object container.

caphrim007:
1. I dont think it can live in extras because it is in module_utils, and that is not an extras "thing"
2. We could, as long as all the changes land in a single release. If we include an interim set of conditional code to check, then we can have separate PRs that address each change without needing to get all the changes in a single PR.
3. Lemme update this PR with a more flexible data structure for you to review.

mhite:
It would actually be great to NOT have the included library outside extras. It seems to make things unnecessarily hard. @bcoca is this possible?

caphrim007:
actually, the bigip_facts module is close enough to how I was going to suggest it be done that why dont i just make PR's for the other modules to have them look like bigip_facts?

f5_parse_arguments cannot be used for REST related work, so I would need to use a different check instead of bigsuds_found. So if I change the existing modules, I would extract that and the ssl context check is already handled by bigip_api

bcoca:
@mhite, not yet, we are looking at options to 'decouple' this better, but there are issues as most shared code is used across core and extras and would create duplication.

bcoca:
for now, just create 2 PRs and make sure they are linked and we'll do the merges as close to each other as possible (make sure link and warning about dependency is on the top of each one).

caphrim007:
thanks for the info @bcoca

mhite:
@caphrim007 -- Now that the changes have been merged for all the various modules in ansible-modules-extras/#2054, it should be safe to merge this. Agree?

 https://github.com/ansible/ansible-modules-extras/pull/2054

mhite:
@caphrim007 -- Now that the unpacking behavior has been changed in https://github.com/ansible/ansible-modules-extras/pull/2102, we should be good to merge this. Agree?

caphrim007:
i suspect yes, but we should run the modules through a test run first, "just in case"

mhite:
Good point. I'll run it through the test suite.

mhite:
I ran it through the test suite; it didn't introduce any regression. However, I didn't do any specific testing of the new feature itself since all my F5's run on standard 443. I did notice this when peaking at the code:

```
server_port = dict(default='443', required=False),
```

You probably actually want:

```
server_port = dict(type='int', default=443, required=False),
```

caphrim007:
@mhite addressed your comment

mhite:
Looks good. Gave it another test to check for regression. No issue. 

Not really sure what's up with the Travis CI build failure. Presuming that's something else unrelated to this change, I'd say...

shipit

mhite:
Also, once this gets merged, we need to update the doc string on all the modules impacted to mention the new server_port parameter.

caphrim007:
@mhite the build failed because of a url timeout. re-triggering travis to try again.

caphrim007:
@gregdek tried to re-run travis several times, but the build doesnt seem to pass (not due to this patch). Will this prevent the merge from happening?

gregdek:
@caphrim007 unclear to me; I'm not involved in the PR process for ansible/ansible repo. @bcoca?

jctanner:
+1

PR #15437, opened 2016-04-15
SSH fails with '"parsed": false' error message

### LINKED ISSUE #15436: ssh connection module could fail with "parsed": false
##### ISSUE TYPE
- Bug Report
##### ANSIBLE VERSION

```
1.9.4
2.0.1.0
```
##### COMPONENT NAME
ssh connection plugin

##### CONFIGURATION

``` ini
[defaults]
timeout = 1

[ssh_connection]
pipelining = True
```
##### OS / ENVIRONMENT

N/A
##### SUMMARY

Ansible ssh connection plugin could fail with "parsed: false" error message in very rare circumstances.
The _run() method of Connection class (in ansible.plugins.connection.ssh module) contains an infinite loop using select to read data from the child process. At the end of the loop the return value of the child process is checked (p.poll() is not None) and if all output has read (not rpipes or <b>not rfd</b>) we break out of the loop. In an extreme situation it could happen that the select does not return any file descriptors because of timeout but when we check the the exit status of the process we already have a result. This means that we exit the loop without fully reading the output of the process resulting a failed task run.

``` python
406         while True:
407             rfd, wfd, efd = select.select(rpipes, [], [], timeout) # select returns with [], [], [] because of timeout
...             # the result of child process could arrive during execution of lines 408-506
507             if p.poll() is not None:
508                 if not rpipes or not rfd: # we break out the loop without reading the output
509                     break
```
##### STEPS TO REPRODUCE

``` yaml
- hosts: all
  gather_facts: False
  tasks:
   # sleep parameter should be ~ timeout + 2
   # timeout is configured in the [defaults] section of ansible.cfg
    - command: sleep 2.967
      become: True
      with_sequence: start=1 end=100000
```
##### EXPECTED RESULTS

In very rare cases (running the above playbook thousands of times) the above playbook fails.

```
$ ansible-playbook test.yml 

changed: [vagrant] => (item=1)
changed: [vagrant] => (item=2)
changed: [vagrant] => (item=3)
...

vagrant                    : ok=1    changed=1    unreachable=0    failed=0   
```
##### ACTUAL RESULTS

```
$ ansible-playbook test.yml 

changed: [vagrant] => (item=1)
changed: [vagrant] => (item=2)
changed: [vagrant] => (item=3)
failed: [vagrant] => (item=4) => {"failed": true, "item": "4", "module_stderr": "", "module_stdout": "", "msg": "MODULE FAILURE", "parsed": false}
changed: [vagrant] => (item=5)
changed: [vagrant] => (item=6)
changed: [vagrant] => (item=7)
...

        to retry, use: --limit @test.retry

vagrant                    : ok=0    changed=0    unreachable=0    failed=1   
```

leki75:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
$ ansible --version
ansible 2.4.0 (ssh f0296f2266) last updated 2017/05/27 09:31:19 (GMT +200)
  config file =
  configured module search path = [u'/Users/leki/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/leki/src/ansible-ssh/ansible/lib/ansible
  executable location = /Users/leki/src/ansible-ssh/ansible/bin/ansible
  python version = 2.7.13 (default, Jan  7 2017, 08:23:07) [GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)]
```
##### SUMMARY

Fixes: #15436

##### ADDON
The error message changed from 2.0 to 2.4:
```
failed: [test] (item=29) => {"failed": true, "item": "29", "module_stderr": "", "module_stdout": "", "msg": "MODULE FAILURE", "rc": 0}
```

### DIFF
--- a/lib/ansible/plugins/connection/ssh.py
+++ b/lib/ansible/plugins/connection/ssh.py
@@ -610,6 +610,7 @@ class Connection(ConnectionBase):
 
         try:
             while True:
+                poll = p.poll()
                 events = selector.select(timeout)
 
                 # We pay attention to timeouts only while negotiating a prompt.
@@ -619,7 +620,7 @@ class Connection(ConnectionBase):
                     if state <= states.index('awaiting_escalation'):
                         # If the process has already exited, then it's not really a
                         # timeout; we'll let the normal error handling deal with it.
-                        if p.poll() is not None:
+                        if poll is not None:
                             break
                         self._terminate_process(p)
                         raise AnsibleError('Timeout (%ds) waiting for privilege escalation prompt: %s' % (timeout, to_native(b_stdout)))
@@ -721,7 +722,7 @@ class Connection(ConnectionBase):
                 # Now we're awaiting_exit: has the child process exited? If it has,
                 # and we've read all available output from it, we're done.
 
-                if p.poll() is not None:
+                if poll is not None:
                     if not selector.get_map() or not events:
                         break
                     # We should not see further writes to the stdout/stderr file
--- a/test/units/plugins/connection/test_ssh.py
+++ b/test/units/plugins/connection/test_ssh.py
@@ -415,7 +415,7 @@ class TestSSHConnectionRun(object):
             self.conn._flags['become_success'] = True
         return (b'', b'')
 
-    def test_pasword_with_prompt(self):
+    def test_password_with_prompt(self):
         # test with password prompting enabled
         self.pc.password = None
         self.pc.prompt = b'Password:'
@@ -440,7 +440,7 @@ class TestSSHConnectionRun(object):
         assert self.conn._send_initial_data.call_count == 1
         assert self.conn._send_initial_data.call_args[0][1] == 'this is input data'
 
-    def test_pasword_with_become(self):
+    def test_password_with_become(self):
         # test with some become settings
         self.pc.prompt = b'Password:'
         self.pc.become = True
@@ -550,8 +550,8 @@ class TestSSHConnectionRetries(object):
 
         monkeypatch.setattr('time.sleep', lambda x: None)
 
-        self.mock_popen_res.stdout.read.side_effect = [b""] * 11
-        self.mock_popen_res.stderr.read.side_effect = [b""] * 11
+        self.mock_popen_res.stdout.read.side_effect = [b""] * 10
+        self.mock_popen_res.stderr.read.side_effect = [b""] * 10
         type(self.mock_popen_res).returncode = PropertyMock(side_effect=[255] * 30)
 
         self.mock_selector.select.side_effect = [


lib/ansible/plugins/connection/ssh.py
```diff
@@ -505,7 +505,7 @@ def _run(self, cmd, in_data, sudoable=True):
             # and we've read all available output from it, we're done.
 
             if p.poll() is not None:
-                if not rpipes or not rfd:
+                if not rpipes or timeout == 0:
```
bcoca: I'm not sure we need to remove `not rfd`, also existing when timeout is 0 prevents our 'last ditch' effort to read anything left in the buffer.

leki75: This problem appears less frequently on Ansible 2.0 as the timeout value is much higher in select (ssh connection timeout + 2), but try with the following modification:

``` python
            # Now we're awaiting_exit: has the child process exited? If it has,
            # and we've read all available output from it, we're done.
            time.sleep(.5)
            if p.poll() is not None:
                if not rpipes or not rfd:
                    break
```

Then execute the following playbook with ssh connection timeout set to 1:

``` yaml
---
- hosts: all
  gather_facts: False
  tasks:
    - command: sleep {{ item | regex_replace('(\d)(\d)', '\1.\2') }}
      with_sequence: start=28 end=38
```

The output on my computer with Ansible 2.0.0.2:

```
$ ANSIBLE_TIMEOUT=1 ansible-playbook test.yml

changed: [vagrant] => (item=28)
changed: [vagrant] => (item=29)
failed: [vagrant] (item=30) => {"failed": true, "item": "30", "module_stderr": "", "module_stdout": "", "msg": "MODULE FAILURE", "parsed": false}
failed: [vagrant] (item=31) => {"failed": true, "item": "31", "module_stderr": "", "module_stdout": "", "msg": "MODULE FAILURE", "parsed": false}
failed: [vagrant] (item=32) => {"failed": true, "item": "32", "module_stderr": "", "module_stdout": "", "msg": "MODULE FAILURE", "parsed": false}
failed: [vagrant] (item=33) => {"failed": true, "item": "33", "module_stderr": "", "module_stdout": "", "msg": "MODULE FAILURE", "parsed": false}
failed: [vagrant] (item=34) => {"failed": true, "item": "34", "module_stderr": "", "module_stdout": "", "msg": "MODULE FAILURE", "parsed": false}
changed: [vagrant] => (item=35)
changed: [vagrant] => (item=36)
changed: [vagrant] => (item=37)
changed: [vagrant] => (item=38)

    to retry, use: --limit @test.retry

vagrant                    : ok=0    changed=0    unreachable=0    failed=1
```

Changing **not rfd** to **timeout == 0** guarantees that **p.stdout** is fully read and the error disappears.

calfonso:
@leki75 would you mind testing this on the latest to see if it's still an issue? 

needs_info

leki75:
This issue still exists. The original `"parsed": False` message has changed to `"rc": 0`.

jctanner:
@leki75 what is the purpose of removing `not events` rather than doing something like ...
```
if not selector.get_map() or not events or timeout == 0:
```

Was the check for events incorrect?

needs_info

abadger:
From @electrofelix in https://github.com/ansible/ansible/issues/15436  
> Tbh, the logic looks good to me anyway. The other alternative, if there is a concern with changing multiple areas in the code is to read the value of p.poll() in to a variable at the start of the loop before performing the select and replace existing checks of it with the variable so that the status was always got before output was checked for, but that's just a kludge to reduce the size of the change.

Reading the code a few times, that alternative seems like it makes more sense.  What do you think of implementing the fix that way?

Evaluation of the two approaches:
* Checking on timeout == 0 means: Once the process returns, always run the loop one more time to process any more data that may have arrived since we last asked for events.
* Moving where we call p.poll() means: First check whether we have exited, then check whether there's any events left.   The drawback here is that we end up checking for whether we've exited more frequently (every time through the loop) instead of only as needed.

abadger:
And feel free to disagree with me, I'm close to being on the fence as to which is better.  Honestly want a some more reasoning to push me one way or the other.

leki75:
@jctanner: the events variable is set at the beginning of the loop and p.poll() is called many lines later. In rare circumstances using `not events` the ssh process could exit after a timeout without checking any new data has arrived. `timeout == 0` guarantees that we always run the loop one more time.

@abadger: putting the result of p.poll() into a variable at the beginning of the loop is another solution. If it is more readable and understandable than I will refactor the code.

abadger:
Merged to devel for the 2.4.0 release.  Thanks for the work on this and patience in getting it merged!

PR #15414, opened 2016-04-14
Improve galaxy handling of tgz roles

willthames:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0 (devel 2163a24a12) last updated 2016/04/14 12:28:50 (GMT +1000)
  lib/ansible/modules/core: (detached HEAD 99cd31140d) last updated 2016/04/14 12:29:48 (GMT +1000)
  lib/ansible/modules/extras: (detached HEAD ab2f4c4002) last updated 2016/04/14 12:29:51 (GMT +1000)
  config file = 
  configured module search path = Default w/o overrides
```
##### SUMMARY

Gerrithub presents tgz downloads that do not have a containing
directory. This causes a stack trace.

As all roles must contain a meta/main.yml, change ansible-galaxy
to use that to determine the enclosing directory (in the case
of multiple meta/main.yml files, use the one with the shortest parent)

Fixes #15413

PR #15444, opened 2016-04-16
Auto install roles

willthames:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0 (devel 9e6988c7b5) last updated 2016/04/16 23:37:26 (GMT +1000)
  lib/ansible/modules/core: (detached HEAD 5409ed1b28) last updated 2016/04/16 23:37:29 (GMT +1000)
  lib/ansible/modules/extras: (detached HEAD 3afe117730) last updated 2016/04/16 23:37:29 (GMT +1000)
  config file = 

```
##### SUMMARY

Implementation of auto roles install ansible/proposals#7

There are some enhancements that I'd like to make
- It would be nice if `roles_path` and `roles_file` could be obtained from ansible.cfg and also be relative to e.g `playbook_dir`. I'm not sure if ansible.cfg can manage that
##### NOTES:
- Incorporates #15414 and #12904 because they're both useful to have (#12904 is the basis of this work anyway - I could split out #15414 if need be)
- Fixes to failing tests are coming along nicely. The test suite has found some odd edge cases, which is great.

### DIFF
--- a/docsite/rst/galaxy.rst
+++ b/docsite/rst/galaxy.rst
@@ -68,6 +68,25 @@ To request specific versions (tags) of a role, use this syntax in the roles file
 
 Available versions will be listed on the Ansible Galaxy webpage for that role.
 
+Automatic Role Installation
+===========================
+
+Since Ansible 2.2, you can install roles using just ansible-playbook by setting auto_install_roles to True in the ansible configuration file and specifying your roles requirements file in the playbook itself::
+
+   - hosts: localhost
+     roles_file: "{{ playbook_dir }}/rolesfile.yml"
+     roles_path: "{{ playbook_dir }}/roles"
+
+Running ansible-playbook with that playbook will install the roles under a directory called roles alongside the playbook.
+
+Each role in the ``roles_file`` will be tested against the ``auto_install_roles_whitelist`` in the ansible configuration file. Typically this will be a list containing only your internal version control service or your public space (e.g. github.com/yourcompany)
+
+e.g.::
+
+   [default]
+   auto_install_roles = True
+   auto_install_roles_whitelist = git.yourcompany.com,github.com/yourcompany
+
 Advanced Control over Role Requirements Files
 =============================================
 
--- a/lib/ansible/cli/galaxy.py
+++ b/lib/ansible/cli/galaxy.py
@@ -402,8 +402,13 @@ class GalaxyCLI(CLI):
             # query the galaxy API for the role data
 
             if role.install_info is not None and not force:
-                display.display('- %s is already installed, skipping.' % role.name)
-                continue
+                if role.install_info['version'] != role.version:
+                    display.display('- changing role %s from %s to %s' %
+                            (role.name, role.install_info['version'], role.version or "unspecified"))
+                    role.remove()
+                else:
+                    display.display('- %s is already installed, skipping.' % role.version_string)
+                    continue
 
             try:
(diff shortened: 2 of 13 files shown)


willthames:
@chouseknecht @bcoca this is the draft implementation of my auto install roles proposal

I know there is more to your proposals, but I think that the implementation of your proposals wouldn't be much different on top of this implementation (other than that this implementation should simplify how modules interact)

bcoca:
A few of things just from the head of this:
- I don't want to use another requirements entry/file but have this part of normal playbook execution, there should be no need to specify roles N times. 
- This feature should be configurable and needs the ability of specifying whitelisted sources.
- if we need to install roles manually we have ansible-galaxy for that, i don't think we should overload playbook with the same options, it is already a stretch to have the 'autoinstall' in playbook, lets not push the manual also.

I'm more partial to start with galaxy and handover to playbook (like pull does) than this intermingling of both.

 Also, a lot of this code seems to swap places between galaxy and playbook classes and not sure why this is needed.

willthames:
To address your points in turn:
- No new requirements file exists. This just uses the same requirements file. Specifying the sources of the roles should be independent of specifying which roles are used (similar to how you use a pom.xml or a requirements.txt to specify versions of libraries rather than importing a versioned module in source code). I have zero intention of changing this bit.
- I can add a couple of configuration pieces for whitelist and auto_install 
- I can eliminate the command line options, and it will simplify the change, definitely.

Some of the code swaps make sense. I can probably move some back, I was struggling to avoid circular dependencies, but there are other ways to do that.

willthames:
@bcoca, what do you think should be the behaviour of the whitelist by default, when auto role install is not enabled? 

Presumably default whitelist should not prevent someone already using galaxy to install roles from then installing roles in the future. But neither do I think that the whitelist should only be used when auto installing roles. Does that mean the default whitelist should be everything?

bcoca:
No, i proposed it ONLY for the autoinstall, if we want to use it for all galaxy installs i would say we need a 2nd setting to toggle that behavior (galaxy_use_whitelist=False|true).

The default whitelist is always empty.

willthames:
@bcoca, the reason for the code moves is that playbooks using ansible.cli.galaxy to install roles seems like a missing abstraction (compared to ansible.cli.galaxy and ansible.cli.playbook using some common code). A lot of role installation stuff happens in ansible.playbook.role.requirement so it kind of made sense to move the shared code (reading roles file, installing roles from that file) into the same place. That refactor doesn't have to happen, but makes the code a little easier to understand, but if there's a better way of doing the abstractions, I'd be happy to hear it.

What I don't want to do is shell out to ansible-galaxy if I can avoid it, as that seems like a nasty hack.

willthames:
Ok, vastly simplified this (although some of the refactor is still present because it's role installation shouldn't really be a concern of the CLI but of the role code (whether that should be somewhere in ansible.galaxy.role or ansible.playbook.role.requirement is debatable, but the latter is currently where some role installation occurs))

Added `auto_install_roles` and `auto_install_roles_whitelist` plus a test case (the docs and ansible.cfg under test/integration shows how this works)

Command line options removed so that roles will be installed by ansible-playbook only if roles_file is present in the playbook and auto_install_roles is on in configuration.

bcoca:
Since we are still discussing the proposal, I"m going to wait on that part is final.

PR #9790, opened 2014-12-12
Update digital_ocean.py to provide "_meta" header

FlorinAndrei:
Dynamic inventory plugins are supposed to provide a "_meta" header to speed things up. This plugin doesn't have that header. I've added the header for the --list option. The speed increase is pretty big even when this plugin is used alone.

If you have multiple dynamic inventory plugins, and one of them doesn't do "_meta", things get EXTREMELY slow, especially when you have lots of hosts. See issue #9774

https://github.com/ansible/ansible/issues/9774

PR #15462, opened 2016-04-18
use _meta; make cache in more expected way

haz-mat:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

ansible 2.0.1.0
##### SUMMARY

digitalocean.py dynamic inventory library makes an API call to DigitalOcean for every host defined in hosts file. Adding _meta toplevel element makes it possible to retrieve all hostvars in a single call. This update includes a behavior change/bug fix, wherein the script used to ignore cached data for listing droplets.


oryband:
Amazing! I was waiting for someone to patch the missing `do_*` variables so they would be available on host discovery. I found #9790 but it was closed without being merged for some reason.

**NOTE** that in order to get the `do_region` for example, you have to fetch `region.slug`. I still don't understand why the variables aren't available in `do_*` form, but this is good enough.

Please merge!

calfonso:
Hi!

We are closing this as this has already been added in e4c1eff52e4

PR #15468, opened 2016-04-19
Handle case of both Key and key

wimnat:
##### ISSUE TYPE
- Bugfix Pull Request
##### SUMMARY

If someone uses the ec2 helper function to convert to snake_case for boto3 results before handling their tags, they may end up passing a dict with keys in lower case rather than upper case.

This small PR handles both 'Key' and 'key'

### DIFF
--- a/lib/ansible/module_utils/ec2.py
+++ b/lib/ansible/module_utils/ec2.py
@@ -364,7 +364,10 @@ def boto3_tag_list_to_ansible_dict(tags_list):
 
     tags_dict = {}
     for tag in tags_list:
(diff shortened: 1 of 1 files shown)


wimnat:
The newly merged ec2_snapshot_facts module is broken until this PR is applied.

wimnat:
@bcoca @defionscode hey guys can we please merge to solve the ec2_snapshot_facts issue shown above?  Or at least change the label to bugfix if this helps with prioritisation?

Thanks

Constantin07:
Any ETA when this will be merged?

PR #15469, opened 2016-04-19
Make ziploader handle python packages as well as python modules

abadger:
##### ISSUE TYPE
- Feature Pull Request
##### SUMMARY

Previous versions of ziploader did not handle importing python packages from ansible.module_utils, only python modules.  This update adds the missing python package support.


abadger:
No objections so merged.

PR #15482, opened 2016-04-19
Fixup perms dont rely on privileged user named root

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### SUMMARY

On some systems, the privileged administrative user is not named "root".  For these, fixup_perms fails to try chowning the file even though chowning would work. Changed fixup_perms to try chown'ing before checking for root privileges.  If chown'ing fails, we check whether we're "root" and fail if so.  If we're not root we try our other methods of making the file readable to the unprivileged user.

Thanks to @n-st for most of the work towards this fix.


abadger:
This has passed my tests so I'm going to merge this. @n-st feel free to ping me on IRC if this doesn't fix things for you.

PR #15509, opened 2016-04-20
Adding docker_common util module.

chouseknecht:
##### ISSUE TYPE
- New Module Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0 (docker_testing 9bb069f873) last updated 2016/04/18 12:36:03 (GMT -400)
  lib/ansible/modules/core: (detached HEAD cf01087a30) last updated 2016/04/05 16:31:53 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 204b4bab56) last updated 2016/04/05 16:31:56 (GMT -400)
  config file =
```
##### SUMMARY

Adding docker_common, the shared utility library for the new docker modules.

### DIFF
--- a/lib/ansible/module_utils/docker_common.py
+++ b/lib/ansible/module_utils/docker_common.py
@@ -38,7 +38,7 @@ try:
     from docker.tls import TLSConfig
     from docker.constants import DEFAULT_TIMEOUT_SECONDS, DEFAULT_DOCKER_API_VERSION
     from docker.utils.types import Ulimit, LogConfig
-except ImportError as exc:
+except ImportError, exc:
     HAS_DOCKER_ERROR = str(exc)
     HAS_DOCKER_PY = False
 
--- a/lib/ansible/module_utils/docker_common.py
+++ b/lib/ansible/module_utils/docker_common.py
@@ -17,21 +17,19 @@
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 
 import os
-import logging
 import re
 import json
 import sys
 import copy
 
-from requests.exceptions import SSLError
 from urlparse import urlparse
-from logging import Handler, NOTSET
 from ansible.module_utils.basic import *
 
 HAS_DOCKER_PY = True
 HAS_DOCKER_ERROR = None
 
 try:
+    from requests.exceptions import SSLError
     from docker import Client
     from docker import __version__ as docker_version
     from docker.errors import APIError, TLSParameterError, NotFound
@@ -60,8 +58,6 @@ DOCKER_COMMON_ARGS = dict(
     tls_verify=dict(type='bool'),
     debug=dict(type='bool', default=False),
     filter_logger=dict(type='bool', default=False),
-    log_path=dict(type='str', default='docker.log'),
-    log_mode=dict(type='str', choices=['stderr', 'file', 'syslog'], default='syslog'),
 )
 
 DOCKER_MUTUALLY_EXCLUSIVE = [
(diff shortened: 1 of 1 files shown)


lib/ansible/module_utils/docker_common.py
```diff
@@ -0,0 +1,498 @@
...
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import logging
+import re
+import json
+import sys
+import copy
+
+from requests.exceptions import SSLError
```
bcoca: requests is required by docker, but not normally installed, put in same try/except as docker import or in it's own to give 'graceful' import error messages, this should be the norm for all non-core libraries

lib/ansible/module_utils/docker_common.py
```diff
@@ -0,0 +1,498 @@
...
+
+if not HAS_DOCKER_PY:
+    # No docker-py. Create a place holder client to allow
+    # instantiation of AnsibleModule and proper error handing
+    class Client(object):
+        def __init__(self, **kwargs):
+            pass
+
+
+def human_to_bytes(number):
```
bcoca, line 84: this should be a common function in basic.py (i think there is this one or the reverse there already)

lib/ansible/module_utils/docker_common.py
```diff
@@ -0,0 +1,498 @@
...
+    i = 1
+    for each in BYTE_SUFFIXES[1:]:
+        if number[-len(each):] == BYTE_SUFFIXES[i]:
+            return int(number[:-len(each)]) * (1024 ** i)
+        i += 1
+
+    raise ValueError("Failed to convert %s. The suffix must be one of %s" % (number, ','.join(BYTE_SUFFIXES)))
+
+
+def log_msg(data, pretty_print=False):
```
bcoca: already have this in basic.py, no need to create your own, it also includes sanitation and no_log

lib/ansible/module_utils/docker_common.py
```diff
@@ -0,0 +1,498 @@
...
+        if pretty_print:
+            # Attempt formatting JSON. The sanitizing may have broken things.
+            try:
+                out = json.dumps(sanitized, sort_keys=True, indent=4, separators=(',', ': '))
+            except:
+                out = sanitized
+    return out
+
+
+def sanitize_dict(data):
```
bcoca: same for these 2, though in diff from

lib/ansible/module_utils/docker_common.py
```diff
@@ -0,0 +1,498 @@
...
+            self.fail("Error connecting: %s" % exc)
+
+    def log(self, msg, pretty_print=False):
+        self.logger.debug(log_msg(msg, pretty_print=pretty_print))
+    
+    def fail(self, msg):
+        self.module.fail_json(msg=msg)
+
+    @staticmethod
+    def _get_value(param_name, param_value, env_variable, default_value):
```
bcoca, line 181: this is already done by argspec when param is defined as boolean, all redundant

bcoca:
so only showstopper is import error the other is just a lot of code that should not live here but in basic.py

abadger:
Are we no longer going to check for the docker server version?  Or is that something that only the modules themselves will know if they should do?

bcoca:
IIRC the 'new docker modules' are starting at much higher version of docker.py which implies higher version of the server. But we really need to make this very clear to users.

chouseknecht:
We follow the pattern in the current docker module:
- We require min version of dockery-py, which is 1.7.0 (the current module has a min of 0.3.0).
- We default the api_version parameter to the default value taken from docker-py.
- We do not check for a min api_version.

Maybe we should require a min api version? We have only tested against the latest API.

lib/ansible/module_utils/docker_common.py
chouseknecht, line 84: Maybe it should be, but i don't see anything right off in basic.py that turns a character string into an int. I see something called pretty_bytes() in basic.py which does the opposite of what I need. It turns a number into a string with this format: '%.2f %s' % (float(size)/ limit, suffix)

lib/ansible/module_utils/docker_common.py
chouseknecht, line 181: pretty sure this is specific to docker_common, especially where we drop in specific values for cert_path, cacert_path and key_path.

chouseknecht:
@bcoca @abadger 

Fixed the import issue.

Removed the logging bits. By the way, what is in basic.by does not work when debugging a module within the context of a playbook.
    a) it writes to syslog, which is not helpful for debugging during dev.
    b) it does not handle a string containing JSON (in fact, the sanitize routine blows up)
    c) even if it did handle a JSON string, it offers no pretty print option for JSON.

So, what I had was actually quite handy. But, I'm tired of fighting the logging battle.

abadger:
> Maybe we should require a min api version? We have only tested against the latest API.

Yeah, from experience with bug reports from the previous docker module I think we're going to need that in the future.  We should be able to add new min api checks in the future once we start getting bug reports, though, so I don't think it should block this PR.  We can add it once it becomes an issue that our users are telling us about.

Issue #14464, opened 2016-02-13
The ec2.py dynamic inventory script does not group RDS instances by tag


karlmdavis:
This makes it unpossible to delete (or otherwise poke at) RDS instances by tag. This would be useful in all the same ways that grouping EC2 instances by tag, e.g. for maintaining separate dev/production instances via an "Environment" tag.

karlmdavis:
As an ugly workaround, you can append the environment name to the RDS instance's `instance_name` property, instead.

For example:

```
- name: RDS - Provision 'myinstancename'
  rds:
    command: create
    region: "{{ region }}"
    instance_name: myinstancename-{{ env }}
    db_engine: postgres
    size: 10
    instance_type: db.t2.medium
    publicly_accessible: false
    username: "{{ postgres_dev_master_username }}"
    password: "{{ postgres_dev_master_password }}"
    tags:
      # Note: it will create the tag, but that tag won't be used for any grouping when the inventory is built via ec2.py.
      Environment: "{{ env }}"
    wait: true
    wait_timeout: 900
```

aioue:
https://github.com/ansible/ansible/pull/15609

nethershaw:
#15609 does not solve this issue. It provides tags for RDS clusters, but not RDS instances.

Why would clusters have received this fix before instances? What obstacles remain to expose the resource tags of RDS instances via dynamic inventory?

ansibot:
@karlmdavis Greetings! Thanks for taking the time to open this issue. In order for the community to handle your issue effectively, we need a bit more information. 

Here are the items we could not find in your description:
- issue type
- ansible version
- component name

Please set the description of this issue with this template:
https://raw.githubusercontent.com/ansible/ansible/devel/.github/ISSUE_TEMPLATE.md

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ghost:
Any movement on this issue since May? Seems to not be supported still and would be useful for our use case.

ansibotdev:
@karlmdavis You have not responded to information requests in this issue so we will assume it no longer affects you. If you are still interested in this, please create a new issue with the requested information.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

Issue #11569, opened 2015-07-13
Too few tags (none) from ec2.py inventory script for RDS instances


zentavr:
##### ISSUE TYPE
Bug Report

##### COMPONENT NAME
contrib/inventory/ec2.py

##### ANSIBLE VERSION
N/A

##### SUMMARY
In fact it was cloned from [here](https://github.com/ansible/ansible-modules-extras/issues/668)

When using [ec2.py](https://github.com/ansible/ansible/blob/devel/contrib/inventory/ec2.py) inventory script, it returns no tags for RDS instances.
Saying, when it is a regular EC2 node, we have this:

```
"172.29.5.134": {
    "ec2__in_monitoring_element": false, 
    "ec2_ami_launch_index": "0", 
    "ec2_architecture": "x86_64", 
    "ec2_client_token": "", 
    "ec2_dns_name": "", 
    "ec2_ebs_optimized": false, 
    "ec2_eventsSet": "", 
    "ec2_group_name": "", 
    "ec2_hypervisor": "xen", 
    "ec2_id": "i-3ace31c4", 
    "ec2_image_id": "ami-d85e75b0", 
    "ec2_instance_profile": "", 
    "ec2_instance_type": "m3.medium", 
    "ec2_ip_address": "", 
    "ec2_item": "", 
    "ec2_kernel": "aki-919dcaf8", 
    "ec2_key_name": "devkey", 
    "ec2_launch_time": "2015-07-01T00:44:19.000Z", 
    "ec2_monitored": false, 
    "ec2_monitoring": "", 
    "ec2_monitoring_state": "disabled", 
    "ec2_persistent": false, 
    "ec2_placement": "us-east-1b", 
    "ec2_platform": "", 
    "ec2_previous_state": "", 
    "ec2_previous_state_code": 0, 
    "ec2_private_dns_name": "ip-172-29-5-134.ec2.internal", 
    "ec2_private_ip_address": "172.29.5.134", 
    "ec2_public_dns_name": "", 
    "ec2_ramdisk": "", 
    "ec2_reason": "", 
    "ec2_region": "us-east-1", 
    "ec2_requester_id": "", 
    "ec2_root_device_name": "/dev/sda1", 
    "ec2_root_device_type": "ebs", 
    "ec2_security_group_ids": "sg-dd42aeba", 
    "ec2_security_group_names": "public", 
    "ec2_sourceDestCheck": "true", 
    "ec2_spot_instance_request_id": "", 
[...]

PR #15526, opened 2016-04-21
Do not set RDS instance hostname if hostname_variable is a tag

Jonnymcc:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.0.1.0
  config file = 
```
##### SUMMARY

#14464 and #11569 are related. The problem is that boto.rds does not return RDS instance tags.

Tags can be retrieved by using boto.rds2, although there are challenges identified by @deyvsh https://github.com/ansible/ansible/issues/11569#issuecomment-175740957. This is a workaround that would at least allowing using a hostname_variable like "tag_Name" and still generate inventory. The alternative is to configure ec2.py to exclude RDS instances from inventory.

Example error before change...

```
  File "./ec2.py", line 1385, in <module>
    Ec2Inventory()
  File "./ec2.py", line 173, in __init__
    self.do_api_calls_update_cache()
  File "./ec2.py", line 446, in do_api_calls_update_cache
    self.get_rds_instances_by_region(region)
  File "./ec2.py", line 521, in get_rds_instances_by_region
    self.add_rds_instance(instance, region)
  File "./ec2.py", line 804, in add_rds_instance
    hostname = instance.tags.get(self.hostname_variable[4:], None)
AttributeError: 'DBInstance' object has no attribute 'tags'
```

boto.rds does not return RDS instance tags. Since it does not there
is no need to check for RDS tags here. Also, if the hostname_variable
is not available the default hostname (the destination address)
should be returned.

### DIFF
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -921,10 +921,7 @@ class Ec2Inventory(object):
         # Set the inventory name
         hostname = None
         if self.hostname_variable:
-            if self.hostname_variable.startswith('tag_'):
-                hostname = instance.tags.get(self.hostname_variable[4:], None)
-            else:
-                hostname = getattr(instance, self.hostname_variable)
+            hostname = getattr(instance, self.hostname_variable, None)
 
         # If we can't get a nice hostname, use the destination address
         if not hostname:


FlorinAndrei:
Can we get this merged, please? I'm hitting this exact same bug.

rhartkopf:
👍 This fixed the issue for me as well.

aairey (changes requested):
I think you meant to make this change on line 924, on the rds level.

Jonnymcc:
ready_for_review

bcoca:
cc @ryansb @s-hertel

s-hertel:
Thanks @Jonnymcc for fixing that. I can reproduce that failure, but not since #23989 was merged. Tagging with rds instances is not finished (can't group by tags yet) but now specifying hostname_variable in ec2.ini works. Closing this because I don't think we should remove that functionality.

PR #15540, opened 2016-04-22
[test_apache2_module] use userdir module as example instead of alias

robinro:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

2.1.0, devel
##### SUMMARY
- alias module is very basic and removing it leads to the suse default config failing
- future improvements might test different modules and the effect of them being removed


gundalow:
Looks like a sensible improvement. Can you please rebase, which should give us green tests. Thanks in advance.

resmo:
fixes tests of apache_module on ubuntu 16.04 without need to set force=true. waiting for the travis build before I merge.

PR #15550, opened 2016-04-22
Adding route53_included_zones parameter to limit zones queried when building ec2.py inventory

jerm:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0 (devel 6758561b82) last updated 2016/04/22 09:34:59 (GMT -700)
```
##### SUMMARY

Rationale: Having many zones can severely slow down the inventory script and it's
conceivable if not likely that one may only want/need a few zones in their inventory.
This is a simple enough addition that it was a no-brainer and is completely opt-in.

Also, now stripping spaces from route53_included_zones and route53_excluded_zones
because spaces around commas would break the list and the example had spaces.
Figured it made more sense to make the parameter more flexible than just fix the
example.

command output is just ec2.py json output and would be _huge_ as I have thousands of zones, but suffice to say there would just be less of it if we only include 1 or 2


bcoca:
commit history is all merges, please clean up

calfonso:
Hi!

- Due to no response to previous comments.

PR #15556, opened 2016-04-22
support become_user with Docker's native user management

bitmensch:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

ansible@devel
##### SUMMARY

This is basically an implementation of @ThomasSteinbach's suggestion in [#13424](https://github.com/ansible/ansible/pull/13424#issuecomment-163461619), where the use of `become_user` with `docker` connection has been considered, but eventually not implemented.
This PR introduces a new `become_method` named `dockerexec` that calls into the docker container with the configured `become_user` using Docker's own user (switching) management.

The main advantage is that Docker containers don't have to be prepared for another `become_method` like `sudo`. One could argue that using `remote_user` might be sufficient for the Docker container use case, but unfortunately if you want to use the very same Ansible playbook/role with other non-docker hosts then it is not.

Please let me know if something is missing, or if there's even an easier way to get this implemented.

### DIFF
--- a/docs/man/man1/ansible.1.asciidoc.in
+++ b/docs/man/man1/ansible.1.asciidoc.in
@@ -70,7 +70,7 @@ Run commands in the background, killing the task after 'NUM' seconds.
 *--become-method=*'BECOME_METHOD'::
 
 Privilege escalation method to use (default=sudo),
-valid choices: [ sudo | su | pbrun | pfexec | runas | doas | dzdo ]
+valid choices: [ sudo | su | pbrun | pfexec | runas | doas | dzdo | dockerexec ]
 
 *--become-user=*'BECOME_USER'::
 
--- a/docsite/rst/become.rst
+++ b/docsite/rst/become.rst
@@ -29,7 +29,7 @@ become_user
     set to user with desired privileges, the user you 'become', NOT the user you login as. Does NOT imply `become: yes`, to allow it to be set at host level.
 
 become_method
-    at play or task level overrides the default method set in ansible.cfg, set to 'sudo'/'su'/'pbrun'/'pfexec'/'doas'/'dzdo'
+    at play or task level overrides the default method set in ansible.cfg, set to 'sudo'/'su'/'pbrun'/'pfexec'/'doas'/'dzdo'/'dockerexec'
 
 
 Connection variables
@@ -60,7 +60,7 @@ New command line options
 
 --become-method=BECOME_METHOD
     privilege escalation method to use (default=sudo),
-    valid choices: [ sudo | su | pbrun | pfexec | doas | dzdo ]
+    valid choices: [ sudo | su | pbrun | pfexec | doas | dzdo | dockerexec ]
 
 --become-user=BECOME_USER
     run operations as this user (default=root), does not imply --become/-b
--- a/docsite/rst/intro_configuration.rst
+++ b/docsite/rst/intro_configuration.rst
@@ -804,7 +804,7 @@ The equivalent of adding sudo: or su: to a play or task, set to true/yes to acti
 become_method
 =============
 
-Set the privilege escalation method. The default is ``sudo``, other options are ``su``, ``pbrun``, ``pfexec``, ``doas``::
+Set the privilege escalation method. The default is ``sudo``, other options are ``su``, ``pbrun``, ``pfexec``, ``doas``, ``dockerexec``::
 
     become_method=su
 
--- a/lib/ansible/constants.py
+++ b/lib/ansible/constants.py
@@ -201,9 +201,9 @@ DEFAULT_SUDO_FLAGS        = get_config(p, DEFAULTS, 'sudo_flags', 'ANSIBLE_SUDO_
 DEFAULT_ASK_SUDO_PASS     = get_config(p, DEFAULTS, 'ask_sudo_pass',    'ANSIBLE_ASK_SUDO_PASS',    False, boolean=True)
 
 # Become
-BECOME_ERROR_STRINGS      = {'sudo': 'Sorry, try again.', 'su': 'Authentication failure', 'pbrun': '', 'pfexec': '', 'runas': '', 'doas': 'Permission denied', 'dzdo': ''} #FIXME: deal with i18n
-BECOME_MISSING_STRINGS    = {'sudo': 'sorry, a password is required to run sudo', 'su': '', 'pbrun': '', 'pfexec': '', 'runas': '', 'doas': 'Authorization required', 'dzdo': ''} #FIXME: deal with i18n
-BECOME_METHODS            = ['sudo','su','pbrun','pfexec','runas','doas','dzdo']
+BECOME_ERROR_STRINGS      = {'sudo': 'Sorry, try again.', 'su': 'Authentication failure', 'pbrun': '', 'pfexec': '', 'runas': '', 'doas': 'Permission denied', 'dzdo': '', 'dockerexec': ''} #FIXME: deal with i18n
+BECOME_MISSING_STRINGS    = {'sudo': 'sorry, a password is required to run sudo', 'su': '', 'pbrun': '', 'pfexec': '', 'runas': '', 'doas': 'Authorization required', 'dzdo': '', 'dockerexec': ''} #FIXME: deal with i18n
+BECOME_METHODS            = ['sudo','su','pbrun','pfexec','runas','doas','dzdo','dockerexec']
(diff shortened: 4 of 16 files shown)


bcoca:
i'm not sure what this brings us except making `remote_user` and `become_user` interchangeable, just setting `ansible_become=False` for docker containers should be enough to make plays compatible.

ThomasSteinbach:
Of course from the technical perspective the PR enables just a replacement of `remote_user` and `become_user`. But I think you should also consider how people percieve the usage of both options. The `remote_user` is in my perception the default user I want to _connect_ with the Docker container. The `become_user` is in my perception the one I use when I want to _switch_ to instead of using the default user. That both options have technically the same implementation should be indifferent for the Ansible user.

Following a little example that changing the `remote_user` per task feels queer:

```
- name: create jenkins container
  docker:
    name: my_jenkins
    image: jenkins

- name: add docker container my_jenkins to ansible inventory
  add_host:
    name: my_jenkins
    ansible_connection: docker
    ansible_user: jenkins # this is optional as the default user is already 'jenkins'
  changed_when: false

- name: example of how switching to root currently works
  delegate_to: my_jenkins
  vars:
    ansible_user: root
  command: date

- name: example of what would be a better approach to switch to root
  delegate_to: my_jenkins
  become: yes
  command: date
```

Additionally reading the docs of _sudo_ and the _-u_ option would underline the meaning of the `become_user`:

man sudo:

```
sudo, sudoedit — execute a command as another user
```

Docker run referene:

```
-u="", --user="": Sets the username or UID used and optionally the groupname or GID for the specified command.
```

imho - further discussion welcomed

bitmensch:
I just tried using `ansible_become=False` but this didn't work as expected either. I will prepare a minimal example that shows that `remote_user` and `become_user` are not the same.

bitmensch:
I hope this minimal example helps to understand the difference I see between `remote_user` and `become_user`.

In my playbooks/roles I tend to follow this pattern to install a service:

```
## inventory
host1 ansible_user=root

## playbook.yml
- hosts: host1

  tasks:
    - name: create myservice system group
      group:
        name: myservice
        system: yes
        state: present

    - name: create myservice system user
      user:
        name: myservice
        group: myservice
        shell: /sbin/nologin
        system: yes
        state: present

    - name: create myservice install directory
      file:
        path: /opt/myservice
        owner: myservice
        group: myservice
        mode: 0755
        state: directory

    - name: copy myservice package
      copy:
        src: myservice.tar.gz
        dest: /tmp/myservice.tar.gz
        owner: myservice
        group: myservice
        mode: 0644

    - name: install myservice
      command: "tar -xzf /tmp/myservice.tar.gz -o -C /opt/myservice --strip-components 1"
      become: yes
      become_user: myservice
```

#### Expectations
1. The playbook/role should be reusable independently of the connection type.
2. The playbook should result in the same outcome when used with ssh connection and e.g. sudo become method as well as docker connection. Generally speaking the `become_method` should not make any difference to the result.
3. In this specific example, files under the `/opt/myservice` directory are expected to be owned by the system user `myservice`.

#### Thoughts

When running this with `docker` connection, `remote_user` and `become_user` are actually interchangeable as I can _connect_ to the docker container even with the `myservice` system user.

But when running this with an `ssh` connection, `remote_user` and `become_user` are not interchangeable as the `myservice` system user would not be able to _connect_ to the host via SSH due to the lack of login shell (which is restricted by design). Therefor in this case I would need to switch users via sudo/su, etc.

bitmensch:
@bcoca @ThomasSteinbach Do you see another alternative to make my example working? If not what else needs to be done (besides rebasing) to get this PR accepted? I would be happy to make any necessary changes to get this accepted at a faster pace.

bcoca:
Yes, setting `ansible_become=False` at the 'docker container hosts', this is one of the reasons host connection vars override play directives.

Consider that not all connection plugins support all become methods, this is very host specific (winrm does not support sudo/su, for example). 

Making this connection plugin behave differently will just add to confusion.

bitmensch:
@bcoca As already said in one of my previous comments: setting `ansible_become=False` on a docker container host does not work. This still results in the `remote_user` being used to connect to the docker container, and not the `become_user`.

That's most likely because the docker connection plugin explicitly runs the `docker exec` command with `-u self._play_context.remote_user`. So, to fix this, I would need to provide the correct `become_user` to the `docker exec` command, but as `self._play_context.become` would be always `False` when set for a docker container host in the inventory, I would have no chance to distinguish between 'when I actually have to become another user' and 'when I have to stick with the remote user'.

@bcoca Did I miss something?

bcoca:
setting ansible_remote_user also?

bitmensch:
@jimi-c What's your opinion on this one? It seems we're stuck here.

ThomasSteinbach:
As far as I have read the converation, I am with @benjixx that the behavior of Ansible should remain the same and also the way people would use Ansible, disregarding the connection type.

bcoca:
I would like that too, but its already not possible when considering most connection types.

bitmensch:
Last night I rebased the PR and explicitly removed `dockerexec` from the set of supported become_methods on non-docker connections, to make sure `dockerexec` cannot be used in the wild. In the long term I think it's better to explicitly define the set of supported become_methods on a connection instead of calculating it, but this should be handled in another PR (from someone who is more experienced in what all these become_methods are and where they might potentially be available).

bcoca:
We already have that, each module has a class variable 'become_methods' that explicitly states the ones they support.

bitmensch:
@bcoca Yes, I know. I updated the variable 'become_methods' on all connections in one of the commits of this PR, 036c633. I just was trying to say that the current calculation of the 'become_methods' should be probably replaced with an explicit listing of become_methods per connection type (for the sense of being more explicit here about what is supported per connection type). Anyway, nothing to be handled in this PR. Sorry for the fuss.

bitmensch:
@jimi-c Did you have a chance to look at this? Would be interested in your opinion on this one. AFAIK Red Hat has partnered with Docker, so Ansible should probably have first-class support for Docker as well!? This PR would be a step in this direction.

bitmensch:
@bcoca @jimi-c Do you think the right way to get something like this into Ansible is by also making become methods pluggable in the same way a user can already plugin its own connection plugin? Or how can we proceed here?

bitmensch:
@bcoca @jimi-c Could someone please answer the questions from my previous two comments to move things forward?

ghost:
I am currently running ansible on a very minimal container and would like to use another user. Does this PR accomplish this?

bitmensch:
@matuschewski-schulze Yes, that's exactly what it is for. As docker already supports becoming another user via its own user management (the `-u` argument on the `exec`command) there would be no need to separately make something like `sudo` available in the docker image anymore.

bitmensch:
@bcoca @jimi-c As this is something I cannot easily workaround, I really would like to get this PR merged. I understand that the overall Ansible issue list is currently getting bigger and bigger and that you're probably absorbed in getting things fixed, but I would much appreciate it if you can spend some minutes providing guidance how to proceed here. Or is there someone else from the Ansible team that could look into this?

bcoca:
i would not put dockerexec in the common become methods (I also need to remove 'runas'), it makes the update much smaller.

bitmensch:
@bcoca If we don't put `dockerexec` and `runas` in the common become methods then how can we still make custom become methods available to Ansible? What's your way forward with `runas`? Are you planning drop support for it, or are you thinking about some `become` plugin support and make it separately available?

bcoca:
- we never added runas support
- you don't need to do plugins to just add it as supported in the docker connection plugin

bitmensch:
@chouseknecht I just saw on the [Ansible 2.2 roadmap](https://github.com/ansible/ansible/blob/devel/docsite/rst/roadmap/ROADMAP_2_2.rst) that you're responsible for the Docker parts of the next release and are also working on [ansible-container](https://github.com/ansible/ansible-container). As ansible-container "connects to your other containers via the Docker engine instead of SSH" you probably have a need to support become_user with Docker's user management as well.

Could you please have a look at this PR?
Do you have a need for this? If so, how can we move this forward?

j00bar:
IMNSHO, the ways to fix `become` with the Docker connection plugin are to either actually figure out why the `su` method doesn't work so that we can include it, or to investigate running `docker exec` with `-t` so that the `sudo` method will work. Until then, I'm using `remote_user` and `ansible_become=False`. While this approach is very creative, and while I share (very literally) the frustrations over `become` with that connection plugin, I'm -1 on the approach in this PR.

bitmensch:
@j00bar Thanks for your feedback. I also started investigating into the "su problem" and just created PR #16226 to enable it for docker connection plugin. Probably you can also comment on #16226.

calfonso:
Hi!

- We don't agree with the implementation and we are already using docker exec for remote_user.

PR #15559, opened 2016-04-23
Make template's relative path available to ansible_managed in regards to #8693

chn42:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0 (template-relpath ddede08b52) last updated 2016/04/23 13:27:26 (GMT +200)
  lib/ansible/modules/core: (detached HEAD 804a8e6378) last updated 2016/04/23 13:44:51 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD 3031105e78) last updated 2016/04/23 13:44:54 (GMT +200)
  config file = 
```
##### SUMMARY

In PR #8693 @mmoya requested to include relative path to template variables and substitute it to ansible_managed. I was searching for exactly this feature but unfortunately it was not merged for some reason.

Due to the time passed by and reorganised directory structure I was not able to rebase the original request, so I decided to at least mention it inhere with the full original PR message.

> Given that {file} in ansible_managed gets substituted to an absolute path, it can change in situations where an username is part of that path no matter if the template is the same file with respect to the playbook using it. For example /home/foo/ansible-site/path/to/template.j2 vs. /home/bar/work/tests/ansible-site/path/to/template.j2)
> 
> This changeset exports the template's relative path as template_relpath var and also substitutes relfile in ansible_managed. 

With this PR it will be possible to use:
- `$template_relpath` as a variable
- `{realpath}` as an alternative to `{file}` in `ansible_managed`


bcoca:
Hi

`ansible_managed` is not a feature we are looking to expand, but to remove instead. Setting a `my_managed` var yourself in any other part of the play is more configurable and less limtied and would still be available in template.

PR #15561, opened 2016-04-23
tests: make tests work for Debian 8/Ubuntu 16.04

resmo:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
2.2
```

tested on Debian 7 / 8 and Ubuntu 12.04 / 14.04 / 16.04


robinro:
for the apache2_module test instead of force, I would recommend to use userdir instead: https://github.com/ansible/ansible/pull/15540
I didn't run it yet, but it should be enough

resmo:
@robinro agree, testing your suggestion on all debian based systems now

PR #15578, opened 2016-04-25
Fix substituting dots for underscores when using ips

jnerin:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.0.1.0
```
##### SUMMARY

ec2.py was substituting the dots on ip addresses inside the groups when not using hostnames like:
  "ec2": [
    "10_10_1_1", 
...
now it's:
  "ec2": [
    "10.10.1.1",
...

### DIFF
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -668,8 +668,8 @@ class Ec2Inventory(object):
         # If we can't get a nice hostname, use the destination address
         if not hostname:
             hostname = dest
-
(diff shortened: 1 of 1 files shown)


alikins:
Diff looks good to me, but haven't tested.

ping @bcoca (for review and to_safe usage review)

bcoca:
need to add string checking to to_safe, but not the job of this PR.

PR #15587, opened 2016-04-25
Fix non-200 responses with json content.

### LINKED ISSUE #15555: uri module: non-200 application/json responses are missing the `json` attribute.
##### ISSUE TYPE

- Bug Report
##### ANSIBLE VERSION

``` shell
$ ansible --version
ansible 2.0.2.0
```
##### SUMMARY

If the uri module returns with a non-200 application/json response, the `json` attribute is missing.
##### STEPS TO REPRODUCE

I whipped up a quick Docker container for this. Create this file locally:

`nginx.conf`:

```
server {
    listen       80;

    location / {
        default_type application/json;
        return 400 '{"msg": "Heyo."}';
    }
}
```

Start the container:

```
$ docker run -v ~/path/to/nginx.conf:/etc/nginx/conf.d/default.conf -P -d nginx
```

Other findings:

As I was looking through `urls.py`, I noticed that all http errors are rescued here: https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/urls.py#L899-L900. It looks like the exception text is added to the `info` dictionary, but not the actual http response body.
##### EXPECTED RESULTS

Under `2.0.1`:

```
$ ansible -m uri -c local -a "url=http://dockerhost:CONTAINER_PORT/" localhost
localhost | FAILED! => {
    "changed": false,
    "connection": "keep-alive",
    "content": "{\"test\": \"Heyo.\"}",
    "content_length": "17",
    "content_type": "application/json",
    "date": "Fri, 22 Apr 2016 20:13:09 GMT",
    "failed": true,
    "json": {
        "test": "Heyo."
    },
    "msg": "Status code was not [200]",
    "redirected": false,
    "server": "nginx/1.9.14",
    "status": 400
}
```

Under `2.0.2`:
##### ACTUAL RESULTS

```
$ ansible -m uri -c local -a "url=http://dockerhost:CONTAINER_PORT/" localhost

localhost | FAILED! => {
    "changed": false,
    "connection": "close",
    "content": "",
    "content_length": "17",
    "content_type": "application/json",
    "date": "Fri, 22 Apr 2016 20:14:03 GMT",
    "failed": true,
    "msg": "Status code was not [200]: HTTP Error 400: Bad Request",
    "redirected": false,
    "server": "nginx/1.9.14",
    "status": 400,
    "url": "http://192.168.99.100:32791/"
}
```

alikins:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

ansible 2.1.0 (http_error_non_200_json 33a2564d03) last updated 2016/04/25 14:57:32 (GMT -400)
  lib/ansible/modules/core: (detached HEAD 804a8e6378) last updated 2016/04/25 15:02:08 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 3031105e78) last updated 2016/04/25 14:57:42 (GMT -400)
##### SUMMARY

Further fix for https://github.com/ansible/ansible/issues/15555

With the first fix for that, the 'resp' from fetch_url would be None, which would
cause some of the response handling to be skipped. This included things like setting the 
'json' attribute.

On HTTPError in fetch_url, return the HTTPError
itself as the 'resp', as well as updating the
'info' member of the response tuple.

fixes #15555

### DIFF
--- a/lib/ansible/module_utils/urls.py
+++ b/lib/ansible/module_utils/urls.py
@@ -897,11 +897,10 @@ def fetch_url(module, url, data=None, headers=None, method=None,
     except (ConnectionError, ValueError), e:
         module.fail_json(msg=str(e))
     except urllib2.HTTPError, e:
-        try:
-            body = e.read()
-        except AttributeError:
-            body = ''
-        info.update(dict(msg=str(e), status=e.code, body=body, **e.info()))
+        info.update(dict(msg=str(e), status=e.code, **e.info()))
+        # urllib.HTTPError can be treated as a Response, so
+        # use it for the response object as well.
+        r = e
     except urllib2.URLError, e:
         code = int(getattr(e, 'code', -1))
         info.update(dict(msg="Request failed: %s" % str(e), status=code))


sivel:
The decision in #14915 was to not set `r = e` due to modules looking to see if `r` is None as a way to determine if the request failed.  Changing this would require that we go back and ensure that all modules are and can handle this correctly.

jimi-c:
@sivel Modules shouldn't be relying on that, they should be looking at the status code. So if this is a problem we should fix the modules.

alikins:
https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/nxos.py#L146  and https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/eos.py#L122  look like they could both be broken with current devel code (if resp is None). 

https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/openswitch.py#L132  would be incorrect, but not crashy. 

https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/a10.py#L64 could change behavior, but only for non-200 statuse codes  < 400. I'd argue that module shouldn't be checking for 'resp' not being None.

alikins:
https://github.com/ansible/ansible-modules-core/blob/devel/packaging/os/apt.py#L626 could start to fail, it doesn't actually check the status code, and only fails if the response.read() throws an exception

alikins:
ditto https://github.com/ansible/ansible-modules-core/blob/devel/packaging/os/yum.py#L584

sivel:
> @sivel Modules shouldn't be relying on that, they should be looking at the status code. So if this is a problem we should fix the modules

I don't disagree.  I also wasn't involved in making that decision.  I am +1 on fixing modules that handle this incorrectly and doing `r = e`

abadger:
We're still talking about this in slack as well -- the main problem isn't our modules... it's third party modules.  We don't have a defined policy on backwards incompatibility in module_utils... but it's kinda been asumed on my part that we don't break compat.  We need to figure that out.  /me talks to people about moving from slack to irc for this conversation.

alikins:
apt_key looks ok

rpm_key looks ok

https://github.com/ansible/ansible-modules-core/blob/devel/packaging

/os/apt_repository.py#L372  looks ok, though it shouldn't need to json.load(response)

network/basics/get_url looks ok

authorized_keys looks ok (checks status itself)

_ec2_ami_search looks ok as well

ec2_facts _fetch() is a little harder to say, but it looks ok. It does check for resp=None, but only to set the response data to none. If something checks return of _fetch() for none, that could change behavior

alikins:
https://github.com/ansible/ansible-modules-core/blob/devel/files/unarchive.py#L652 looks like it could be wrong now. That only checks for resp is None, and not the status code.

alikins:
https://github.com/ansible/ansible-modules-extras/blob/devel/packaging/language/maven_artifact.py#L226 looks to be ok (checks status code
looks okay as well

https://github.com/ansible/ansible-modules-extras/blob/devel/packaging/os/layman.py#L116 looks ok (checks status)

https://github.com/ansible/ansible-modules-extras/blob/devel/network/dnsmadeeasy.py#L167   will likely break. It checks for '200'-ish error codes, but doesn't exclude others. Further up the stack, it checks for resp/content being None to indicate failure in a few places

https://github.com/ansible/ansible-modules-extras/blob/devel/network/citrix/netscaler.py#L124: hard to say, it doesn't seem to check for existince of resp, but it does eventually include it in the module return

https://github.com/ansible/ansible-modules-extras/blob/devel/network/ipify_facts.py#L78: Likely will break . checks for existence of resp and doesn't check status code  (and also, json.loads() the content body itself)

alikins:
https://github.com/ansible/ansible-modules-extras/blob/devel/network/cloudflare_dns.py#L319 : Likely OK. Checks for status codes explicitly,  but there is a 
chunk at https://github.com/ansible/ansible-modules-extras/blob/devel/network/cloudflare_dns.py#L358  that tries to handle error status code _and_ no resp to throw a module fail. So that will be different behavior for any status codes it doesn't handle explicitly [200,304,400,401,403,429,405,415].

https://github.com/ansible/ansible-modules-extras/blob/devel/web_infrastructure/jira.py#L186  looks OK to me.

https://github.com/ansible/ansible-modules-extras/blob/devel/cloud/cloudstack/cs_facts.py#L155  Likely will change, checks for existence of response, and returns None in that case.

https://github.com/ansible/ansible-modules-extras/blob/devel/monitoring/datadog_event.py#L144   Looks like it would be OK. It checks for the status code.

https://github.com/ansible/ansible-modules-extras/blob/devel/monitoring/librato_annotation.py#L134  looks OK, checks status code.

https://github.com/ansible/ansible-modules-extras/blob/devel/monitoring/pagerduty.py#L173 (etc) Looks OK. Checks status code. This module also json.loads() the response content on 200 instead of resp.json.

https://github.com/ansible/ansible-modules-extras/blob/devel/monitoring/stackdriver.py#L143  Should be OK with change. Checks status code.

https://github.com/ansible/ansible-modules-extras/blob/devel/monitoring/boundary_meter.py#L126  (and calls to http_request at https://github.com/ansible/ansible-modules-extras/blob/devel/monitoring/boundary_meter.py#L167 https://github.com/ansible/ansible-modules-extras/blob/devel/monitoring/boundary_meter.py#L191https://github.com/ansible/ansible-modules-extras/blob/devel/monitoring/boundary_meter.py#L212  Looks OK to me. It always checks the status code for is !200 for failure.

alikins:
https://github.com/ansible/ansible-modules-extras/blob/devel/monitoring/airbrake_deployment.py#L120  Looks OK, checks status code.

https://github.com/ansible/ansible-modules-extras/blob/devel/monitoring/rollbar_deployment.py#L122  Looks OK, checks status code is 200.

https://github.com/ansible/ansible-modules-extras/blob/devel/monitoring/newrelic_deployment.py#L136  Looks OK, checks status codes.

https://github.com/ansible/ansible-modules-extras/blob/devel/monitoring/uptimerobot.py#L96   Hard to say, it doesn't look like that module handles http errors in any way. It also returns part of the response in exit_json()

https://github.com/ansible/ansible-modules-extras/blob/devel/monitoring/pagerduty_alert.py   Looks OK. Checks status code. Another module that does it's own json.loads(response).

https://github.com/ansible/ansible-modules-extras/blob/devel/monitoring/bigpanda.py#L176  Looks OK. checks status code for 200.

alikins:
https://github.com/ansible/ansible-modules-extras/blob/devel/clustering/kubernetes.py#L237   Hard to say. The use of fetch_url itself checks for 'resp is None',
but... other parts of the code seem to assume they will get a resp on non-200. For
ex, https://github.com/ansible/ansible-modules-extras/blob/devel/clustering/kubernetes.py#L247 expects a 409 to have a json body and for it to be in the response, which eventually ends up in the exit_json() args

https://github.com/ansible/ansible-modules-extras/blob/devel/database/misc/riak.py#L157  looks OK. Checks status code.

https://github.com/ansible/ansible-modules-extras/blob/devel/notification/twilio.py#L130
https://github.com/ansible/ansible-modules-extras/blob/devel/notification/twilio.py#L162 Looks OK, checks status code.

alikins:
https://github.com/ansible/ansible-modules-extras/blob/devel/notification/hipchat.py#L157  looks OK,checks status code.

https://github.com/ansible/ansible-modules-extras/blob/devel/notification/flowdock.py#L182  Looks OK, checks status code and ignores resp.

https://github.com/ansible/ansible-modules-extras/blob/devel/notification/pushover.py#L86  Looks OK, checks status code

https://github.com/ansible/ansible-modules-extras/blob/devel/notification/nexmo.py#L93
Looks OK, checks status code. This module also does it's own json.loads(resp)

https://github.com/ansible/ansible-modules-extras/blob/devel/notification/hall.py#L72
Looks OK, checks status code, ignores resp.

https://github.com/ansible/ansible-modules-extras/blob/devel/notification/grove.py#L84
Looks OK, checks status code, ignores resp

https://github.com/ansible/ansible-modules-extras/blob/devel/notification/slack.py#L213
Looks OK, checks status code.

https://github.com/ansible/ansible-modules-extras/blob/devel/notification/campfire.py#L119  Looks OK, checks status code.

https://github.com/ansible/ansible-modules-extras/blob/devel/notification/sendgrid.py#L134  Looks OK, checks status code, ignores resp

https://github.com/ansible/ansible-modules-extras/blob/devel/source_control/github_hooks.py#L122  mixed bag here. Most of fetch_url checks status code
and ignores response on non-200.  https://github.com/ansible/ansible-modules-extras/blob/devel/source_control/github_hooks.py#L143 however, doesn't check status code and does return the response body, however, nothing in the module checks return of _delete. So likely OK.

PR #15588, opened 2016-04-25
Add a .mailmap for 'shortlog'

alikins:
##### ISSUE TYPE
- Docs Pull Request
##### SUMMARY

Add a .mailmap file to git, so that tools like 'git shortlog' do a better job of handling names/emails that change over time.


bcoca:
i'm just happy i'm not only one with more than 3 addresses ...

PR #15586, opened 2016-04-25
Fix bugs from API version 1.20 testing.

chouseknecht:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0 (docker_testing 0f373c1767) last updated 2016/04/23 12:34:33 (GMT -400)
  lib/ansible/modules/core: (detached HEAD cf01087a30) last updated 2016/04/05 16:31:53 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 204b4bab56) last updated 2016/04/05 16:31:56 (GMT -400)
  config file =
  configured module search path = Default w/o overrides
```
##### SUMMARY

Improvements and bug fixes from testing against API version 1.20.

PR #15595, opened 2016-04-26
Fix python 2.4 checks for modules

willthames:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0 (devel 826190e02e) last updated 2016/04/26 17:29:36 (GMT +1000)
  lib/ansible/modules/core: (detached HEAD 4549ea5e85) last updated 2016/04/26 13:27:28 (GMT +1000)
  lib/ansible/modules/extras: (detached HEAD 86f08bfcda) last updated 2016/04/26 13:27:28 (GMT +1000)
  config file = 
```
##### SUMMARY
#15586 introduced a python 2.4 error causing all subsequent builds to fail.

Using `except Exception as e` fails with python 2.4, which
all modules (and thus common module code) is expected to support.

This should fix broken builds

### DIFF
--- a/lib/ansible/module_utils/docker_common.py
+++ b/lib/ansible/module_utils/docker_common.py
@@ -426,7 +426,7 @@ class AnsibleDockerClient(Client):
         '''
         try:
             response = self.images(name=name)
-        except Exception as exc:
+        except Exception, exc:
             self.fail("Error searching for image %s - %s" % (name, str(exc)))
         images = response
         if tag: 


d0ugal:
LGTM.

bcoca:
actually, not the right fix, this file already requires +2.6 due to docker-py dependency, it should be added to those excluded from this check.

bcoca:
so discussed and closing in favor of excluding this file from 2.4 checks.

abadger:
This is the commit that should fix this: https://github.com/ansible/ansible/commit/e4a1622f538e68415758f51b22af6bc03be0fdfb

PR #15600, opened 2016-04-26
Ignore failures in Python 2.4

d0ugal:
I feel like this could be removed as it always fails. As I
understand 2.4 isn't supported, so, ignoring it makes the
feedback to GitHub clearer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,8 @@ matrix:
     - env: TARGET=sanity TOXENV=py35
       python: 3.5
     - env: TARGET=sanity TOXENV=py24
+  allow_failures:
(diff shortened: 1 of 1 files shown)


bcoca:
2.4 is supported for modules, though we make exceptions for when they already depend on libraries that require higher versions. 

Having an error on this check means that an action must be taken, either correct the error by making the code compatible or add the file to the exclusion list.

That is why we cannot accept this PR.

d0ugal:
I see, makes sense. Sorry for the misunderstanding. I see somebody has already submitted a fix. https://github.com/ansible/ansible/pull/15595

PR #15609, opened 2016-04-26
Provide full and filtered RDS Aurora details in ec2 inventory

aioue:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.0.2.0
  config file =
```
##### SUMMARY

Add db_clusters to the ec2 inventory. Show tags. Only show clusters matching tags in the .ini.

Use boto3 to get the clusters.

```
  "db_clusters": {
    "test-eu-west-1": {
      "AllocatedStorage": 1,
      "AvailabilityZones": [
        "eu-west-1a",
        "eu-west-1b",
        "eu-west-1c"
      ],
      "BackupRetentionPeriod": 3,
      "DBClusterIdentifier": "test-eu-west-1",
      "DBClusterMembers": [
        {
          "DBClusterParameterGroupStatus": "in-sync",
          "DBInstanceIdentifier": "test-eu-west-1-0",
          "IsClusterWriter": true,
          "PromotionTier": 1
        },
        {
          "DBClusterParameterGroupStatus": "in-sync",
          "DBInstanceIdentifier": "test-eu-west-1-1",
          "IsClusterWriter": false,
          "PromotionTier": 1
        }
      ],
      "DBClusterParameterGroup": "aurora-cluster-test",
      "DBSubnetGroup": "dev_test_db_group",
      "Endpoint": "test-eu-west-1.cluster-sdfsdfsd.eu-west-1.rds.amazonaws.com",
      "Engine": "aurora",
      "EngineVersion": "5.6.10a",
      "KmsKeyId": "arn:aws:kms:eu-west-1:345345345345:key/sfsdfsdfsdfsdf",
      "MasterUsername": "sdfsdfsdfs",
      "Port": 33306,
      "PreferredBackupWindow": "05:21-05:51",
      "PreferredMaintenanceWindow": "sat:03:18-sat:03:48",
      "Status": "available",
      "StorageEncrypted": true,
      "Tags": [
        {
          "Key": "env",
          "Value": "test"
        }
      ],
      "VpcSecurityGroups": [
        {
          "Status": "active",
          "VpcSecurityGroupId": "sg-234234234"
        }
      ]
    }
  }
```


dangerousben:
+1

edinhodzic:
👍

jasonpaige:
+1

argyakrivos:
👍

adamdyer:
:+1:

contrib/inventory/ec2.py
```diff
@@ -129,6 +129,7 @@
 from boto import rds
 from boto import elasticache
 from boto import route53
+import boto3
```
Jonnymcc, line 132: Would this work with boto.rds2 instead of adding a dependency?

aioue, line 132: @Jonnymcc rds2 ony allows for instance describe, not cluster. http://boto.cloudhackers.com/en/latest/ref/rds2.html#boto.rds2.layer1.RDSConnection.describe_db_instances

contrib/inventory/ec2.py
```diff
@@ -530,6 +539,55 @@ def get_rds_instances_by_region(self, region):
...
+                if e.message.find('DBInstanceNotFound') >= 0:
+                    # AWS RDS bug (2016-01-06) means deletion does not fully complete and leave an 'empty' cluster.
+                    # Ignore errors when trying to find tags for these
+                    pass
+
+            # ignore empty clusters caused by AWS bug
+            if len(c['DBClusterMembers']) == 0:
+                continue
+            elif matches_filter:
+                c_dict[c['DBClusterIdentifier']] = c
```
ryansb, line 587: It looks like if there are no instance filters defined, then this will always be false, causing clusters to never get added. The default behavior should be to add all the clusters if there is no filter.

contrib/inventory/ec2.py
```diff
@@ -530,6 +539,55 @@ def get_rds_instances_by_region(self, region):
...
+                if self.ec2_instance_filters:
+                    for filter_key, filter_values in self.ec2_instance_filters.items():
+                        # get AWS tag key e.g. tag:env will be 'env'
+                        tag_name = filter_key.split(":", 1)[1]
+                        # Filter values is a list (if you put commas into the value it will be multiple items) - ignore the rest of the items for now
+                        matches_filter = any(d['Key'] == tag_name and d['Value'] == filter_values[0] for d in c['Tags'])
+
+                        if matches_filter:
+                            # print "Add DB to list"
+                            # it matches a filter, so stop looking for further matches
```
ryansb, line 574: Looks like these debug statements should be removed.

ryansb:
Close, but change the `matches_filter` to default to true so the module works when no instance filters are given. 

needs_revision

aioue:
@ryansb this matches the existing rds behaviour, which is completely unfiltered

ryansb:
I don't believe it does. When I provide no instance tag filter, I still get all the RDS instances in my account, but no clusters show up in the inventory. If you make `matches_filter` true, then the functionality will match.

contrib/inventory/ec2.py
```diff
@@ -530,6 +539,55 @@ def get_rds_instances_by_region(self, region):
...
+                            break
```
Jonnymcc, line 575: This would break at the first matching filter right? If there was more than one filter, wouldn't we want to match all filters?

contrib/inventory/ec2.py
Jonnymcc: Relating to my previous comment, instance filters are iterated through here https://github.com/ansible/ansible/pull/15609/files#diff-c2e371537339bbab5d9d0cc5d6cb7adcR564 and matches_filter is the result of testing each filter. Doesn't this mean only the last filter needs to match for the cluster to be included?

contrib/inventory/ec2.py
ryansb, line 575: No, because all we're looking for is whether it matches any filters so we can choose whether or not to include it in the inventory.

contrib/inventory/ec2.py
```diff
@@ -530,6 +539,55 @@ def get_rds_instances_by_region(self, region):
...
+                # arn:aws:rds:<region>:<account number>:<resourcetype>:<name>
+                tags = client.list_tags_for_resource(
+                    ResourceName='arn:aws:rds:' + region + ':' + account_id + ':cluster:' + c['DBClusterIdentifier'])
+                c['Tags'] = tags['TagList']
+
+                if self.ec2_instance_filters:
+                    for filter_key, filter_values in self.ec2_instance_filters.items():
+                        # get AWS tag key e.g. tag:env will be 'env'
+                        tag_name = filter_key.split(":", 1)[1]
+                        # Filter values is a list (if you put commas into the value it will be multiple items) - ignore the rest of the items for now
```
Jonnymcc: Are filter values allowed to be lists of items? Is this expected behavior? From ec2.ini...

```
# Filters are key/value pairs separated by '=', to list multiple filters use
# a list separated by commas. See examples below.
```

contrib/inventory/ec2.py
Jonnymcc, line 575: Ok, you're right, that is consistent with the description of filters in ec2.ini. :+1:

contrib/inventory/ec2.py
ryansb: They're split before this code

Jonnymcc: What I'm getting at is that 
`for filter_key, filter_values` should be `for filter_key, filter_value`
and 
`matches_filter = any(d['Key'] == tag_name and d['Value'] == filter_values[0] for d in c['Tags'])`
should be
`matches_filter = any(d['Key'] == tag_name and d['Value'] == filter_value for d in c['Tags'])`

ryansb: Ahhh, I see what you're saying. That'll need to get fixed either in this PR or I can submit a followup.

aioue:
@ryansb have modified it to produced cluster information for all clusters when no filters specified. Happy for this to be picked up by Ansible staff to match the coding style.

ryansb:
@aioue Looks good, but can you apply this quick patch to just get rid of the extra comments and support any tag filter (instead of just one value per tag)?

```
-                        # Filter values is a list (if you put commas into the value it will be multiple items) - ignore the rest of the items for
-                        matches_filter = any(d['Key'] == tag_name and d['Value'] == filter_values[0] for d in c['Tags'])
+                        # Filter values is a list (if you put multiple values for the same tag name)
+                        print filter_values
+                        matches_filter = any(d['Key'] == tag_name and d['Value'] in filter_values for d in c['Tags'])

                         if matches_filter:
-                            # print "Add DB to list"
                             # it matches a filter, so stop looking for further matches
                             break
-                            # else:
-                            # print "ignore this DB, filter out"
```

That way users can specify filters like `tag:env=staging,tag:env=development` and have the proper clusters.

aioue:
@ryansb done. Sans the print statement which appeared redundant.

ryansb:
Ah, oops, was doing a bit of testing myself. Thanks!

Looks good to me: ship_it

aioue:
It's not clear why the build is failing. Would help if the travis output was colorized :)

ryansb:
It looks like an unrelated failure:

```
TASK [test_uri : test https fetch to a site with mismatched hostname and certificate and validate_certs=no] ***

fatal: [testhost]: FAILED! => {"accept_ranges": "bytes", "changed": false, "connection": "close", "content": "\n<?xml version=\"1.0\" encoding=\"utf-8\"?>\n<!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.0 Strict//EN\"\n \"http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd\">\n<html>\n  <head>\n    <title>503 Connection timed out</title>\n  </head>\n  <body>\n    <h1>Error 503 Connection timed out</h1>\n    <p>Connection timed out</p>\n    <h3>Guru Mediation:</h3>\n    <p>Details: cache-ord1739-ORD 1462822853 4288211105</p>\n    <hr>\n    <p>Varnish cache server</p>\n  </body>\n</html>\n", "content_length": "454", "content_type": "text/html; charset=utf-8", "date": "Mon, 09 May 2016 19:40:53 GMT", "failed": true, "gid": 0, "group": "root", "mode": "0644", "msg": "Status code was not [200]: HTTP Error 503: Connection timed out", "owner": "root", "path": "/root/ansible_testing/kreitz.html", "redirected": false, "retry_after": "0", "server": "Varnish", "size": 454, "state": "file", "status": 503, "uid": 0, "url": "https://www.kennethreitz.org/", "via": "1.1 varnish", "x_cache": "MISS", "x_cache_hits": "0", "x_served_by": "cache-ord1723-ORD"}
```

I think the change to fix this one is in, can you rebase your change to current devel and let the build re-run?

aioue:
@ryansb what do I need to do get this into shape?

ryansb:
You need to rebase your commits on top of the current devel - merging devel in to your branch leads to a messy git history. 

Other than that, just a response to this comment: https://github.com/ansible/ansible/pull/15609/files#r62228605

aioue:
@ryansb ec2.py got changed underneath while we were refining. Resubmitted as https://github.com/ansible/ansible/pull/16203

aioue:
Closing as https://github.com/ansible/ansible/pull/16203 against fresh devel is passing. Please review latest PR.

PR #15630, opened 2016-04-27
Make -q flag totally quiet for env-setup.fish

brettinman:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.0.2.0 (stable-2.0 24d9e5e0b4) last updated 2016/04/19 15:40:15 (GMT -700)
  lib/ansible/modules/core: (detached HEAD 87b2094bbd) last updated 2016/04/26 17:04:17 (GMT -700)
  lib/ansible/modules/extras: (detached HEAD 4eb177e545) last updated 2016/04/26 17:03:29 (GMT -700)
```
##### SUMMARY

The first "echo" and the setup.py step are not quiet even when using the '-q' flag. This results in output that is annoying if your Fish is configured to source the file - every time you open a terminal, you'll see a lot of output. This fixes that. There is no effect on output if you're not using the "-q" flag.

### DIFF
--- a/hacking/env-setup.fish
+++ b/hacking/env-setup.fish
@@ -8,6 +8,15 @@ set PREFIX_PYTHONPATH $ANSIBLE_HOME/lib
 set PREFIX_PATH $ANSIBLE_HOME/bin 
 set PREFIX_MANPATH $ANSIBLE_HOME/docs/man
 
+# set quiet flag
+if set -q argv
+    switch $argv
+    case '-q' '--quiet'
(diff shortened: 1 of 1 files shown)


abadger:
I don't use fish so I'll trust that this works as advertised :-)  Thanks!

abadger:
merged to devel and stable-2.1 (so it should be in the ansible-2.1.0 release).

Issue #12086, opened 2015-08-25
make with_ loops configurable


bcoca:
##### ISSUE TYPE
Feature Idea

##### COMPONENT NAME
core

##### ANSIBLE VERSION
2.1

##### SUMMARY
```
how: 
    forks: 1
    pause: 0
    squash: name
    label: "{{item.name}}"
    end: on_fail
with_items: ...
```
- forks: forks within the loop to do items in parallel, default 1, this needs warnings
- ~~pause: between loop executions, useful in throttled api scenario~~ _Done in 2.2_
- ~~squash: join all items into list and pass to provided option, works like current hardcoded opts for apt, yum, etc, by default it should be None~~ _abandon_: reversed opinion, we should remove this feature
- end: when to interrupt the loop, default is 'last item', options? on_fail, on_success (first one)?
- ~~label: (#13710) what to display when outputting the item loop~~ _Done in 2.2_

docs to current state at:

http://docs.ansible.com/ansible/playbooks_loops.html#loop-control

amenonsen:
Please let's not call it `how`. That's even worse to read than `become: true`. But the functionality under it looks great.

bcoca:
includes fix for #10695

mahemoff:
Excellent. In the interests of bikeshedding, maybe call it `looping:`.

realcnbs:
:+1:

yikaus:
+1

jyennaco:
+1 especially for within host parallelization!

rbarabas:
:+1:

hloeffler:
:+1: 
but lets not call it "how"

bcoca:
so here is a workaround for breaking a loop task after the first failure

```
- hosts: localhost
  vars:
    myvar:
        - 1
        - 2
        - 3
        - 4
        - 5
  tasks:
    - name: break loop after 3
      debug: msg={{item}}
      failed_when: item == 3
      register: myresults
      when: not (myresults|default({}))|failed
      with_items: "{{myvar}}"
```

yikaus:
@bcoca not working from end (ansible 1.9.3 ubuntu ) 

TASK: [break loop after 3] ***************************************************\* 
ok: [localhost] => (item=1) => {
    "failed": false, 
    "failed_when_result": false, 
    "item": 1, 
    "msg": "1"
}
ok: [localhost] => (item=2) => {
    "failed": false, 
    "failed_when_result": false, 
    "item": 2, 
[...]

PR #15657, opened 2016-04-29
Simplify and improve package name squashing

### LINKED ISSUE #15649: yum module fails with 'dict object' has no attribute '\\x00$' when used in loop and with dict lookup
##### ISSUE TYPE
- Bug Report
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel f2980fc565) last updated 2016/04/28 14:44:57 (GMT +000)
  lib/ansible/modules/core: (detached HEAD e78ee3b128) last updated 2016/04/28 14:46:09 (GMT +000)
  lib/ansible/modules/extras: (detached HEAD ca310f3d15) last updated 2016/04/28 14:46:13 (GMT +000)
  config file = 
```
##### CONFIGURATION

No configuration.
##### OS / ENVIRONMENT

Reproduced on Fedora 23. I'm connecting to localhost, however the error happens before connection attempt is made.
##### SUMMARY

Running the attached playbook produces the rather cryptic error `'dict object' has no attribute '\\x00$'`. The bug appears to be related to:
- yum module (dnf, apt too),
- loops,
- dictionary lookup in the name argument to the module,

as taking any out of the equation does not reproduce the error.
##### STEPS TO REPRODUCE

Run following playbook:

```

---
- hosts: all
  gather_facts: false
  vars:
    package_map:
      foo: bar
    packages:
    - foo
  tasks:
  - yum: name={{ package_map[item] }}
    with_items: "{{ packages }}"
```

`ansible-playbook -i hosts reproduce.yml -vvvv`
##### EXPECTED RESULTS

Package `bar` to be installed using `yum`.
##### ACTUAL RESULTS

```
Loaded callback default of type stdout, v2.0

PLAYBOOK: reproduce.yml ********************************************************
1 plays in reproduce.yml

TASK [yum] *********************************************************************
task path: /home/fedora/ansible-reproducer/reproduce.yml:12
fatal: [localhost]: FAILED! => {"failed": true, "msg": "'dict object' has no attribute '\\x00$'"}

        to retry, use: --limit @reproduce.retry

```

willthames:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel 7af47a3886) last updated 2016/04/29 11:28:56 (GMT +1000)
  lib/ansible/modules/core: (detached HEAD e78ee3b128) last updated 2016/04/29 11:28:58 (GMT +1000)
  lib/ansible/modules/extras: (detached HEAD ca310f3d15) last updated 2016/04/29 11:28:58 (GMT +1000)
  config file = 
```
##### SUMMARY
- Allow variables that contain a dict lookup
- Squash names built from dictionary values  e.g.
  
  ```
  package: name="{{item.name}}-{{item.version}}"
  with_items:
  - name: hello
    version: 1.2.3
  - name: goodbye
    version: 2.3.4
  ```
  
  now works the same as
  
  ```
  package: name="{{item}}"
  with_items:
  - hello-1.2.3
  - goodbye-2.3.4
  ```
- Fix some erroneous test assumptions  e.g.
  
  ```
  package: name="{{pkg_mgr}}"
  with_items:
  - hello-1.2.3
  - goodbye-2.3.4
  ```
  
  should just install a single package with  the name `pkg_mgr` (in effect `item` is not a bound variable)

Fixes #15649

### DIFF
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -280,48 +280,33 @@ class TaskExecutor:
             task_action = templar.template(task_action, fail_on_undefined=False)
 
         if len(items) > 0 and task_action in self.SQUASH_ACTIONS:
-            if all(isinstance(o, string_types) for o in items):
-                final_items = []
+            # use list rather than set to preserve ordering
+            final_items = list()
 
-                name = None
-                for allowed in ['name', 'pkg', 'package']:
-                    name = self._task.args.pop(allowed, None)
-                    if name is not None:
-                        break
+            name = None
+            for allowed in ['name', 'pkg', 'package']:
+                name = self._task.args.pop(allowed, None)
+                if name is not None:
+                    break
 
-                # This gets the information to check whether the name field
-                # contains a template that we can squash for
-                template_no_item = template_with_item = None
-                if name:
-                    if templar._contains_vars(name):
-                        variables[loop_var] = '\0$'
-                        template_no_item = templar.template(name, variables, cache=False)
-                        variables[loop_var] = '\0@'
-                        template_with_item = templar.template(name, variables, cache=False)
-                        del variables[loop_var]
-
-                    # Check if the user is doing some operation that doesn't take
-                    # name/pkg or the name/pkg field doesn't have any variables
-                    # and thus the items can't be squashed
-                    if template_no_item != template_with_item:
-                        for item in items:
-                            variables[loop_var] = item
-                            if self._task.evaluate_conditional(templar, variables):
-                                new_item = templar.template(name, cache=False)
-                                final_items.append(new_item)
-                        self._task.args['name'] = final_items
-                        # Wrap this in a list so that the calling function loop
-                        # executes exactly once
-                        return [final_items]
-                    else:
-                        # Restore the name parameter
-                        self._task.args['name'] = name
-            #elif:
-                # Right now we only optimize single entries.  In the future we
-                # could optimize more types:
-                # * lists can be squashed together
-                # * dicts could squash entries that match in all cases except the
-                #   name or pkg field.
+            for item in items:
+                variables[loop_var] = item
+                if self._task.evaluate_conditional(templar, variables):
+                    new_item = templar.template(name, cache=False)
+                    # Do not add templated name if other task arguments
+                    # e.g. state are also templated - this could be
+                    # optimised if results of all other templates are
+                    # the same
+                    if new_item != name and \
+                            not any([v != templar.template(v, cache=False)
+                                for (k, v) in self._task.args.items()]) and \
+                            not new_item in final_items:
+                        final_items.append(new_item)
+            if final_items:
+                self._task.args['name'] = final_items
+                return [final_items]
+            else:
+                self._task.args['name'] = name
         return items
 
     def _execute(self, variables=None):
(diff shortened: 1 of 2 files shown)


bcoca:
I wanted to go one step further and make it configurable as shown in #12086 

ignore `how` we went with `loop_control`

willthames:
@bcoca, the tests broke on `loop_control` with my first push, but it's now working with `loop_control: { 'loop_var': 'another_name_for_item' }` too. 

It would be good to get the fix for #15649 backported to 2.0 and 2.1. Looks like it can be trivially cherry-picked onto stable-2.1 and reasonably easily cherry-picked onto stable-2.0.

willthames:
Also, this is mislabelled feature_pull_request.

kustodian:
Could we please merge this into devel, and not wait for 2.2, since this is a bugfix pull request, not a feature request. Currently in devel modules which use squashing of `with` loops + dict lookup are broken as it was reported in #15649 (which was closed even though it wasn't fixed).

I tested this pull request on the latest devel and it worked like a charm.

abadger:
Okay, I cudgeled my brain and finally recalled the reason that we template the name field twice in the old code.  Things like this are caught by the old code but not by the proposed code:

```
---
- hosts: localhost
  vars:
    list_of_pkgs:
      - python-q
      - python-epdb
  tasks:
  # Uninstall package and then make sure it is the latest to work around
  # improper deps inside old package
  - dnf:
      name: "{{ list_of_pkgs }}"
      state: "{{ items }}"
    with_items:
      - absent
      - latest
```

Templating the name field twice catches the fact that list_of_pkgs is templated to the same value in both instances of the loop.  Simply comparing the templated name field to the original name field does not catch this case.

abadger:
I don't think that's a deal breaker, though.  I don't think the problem being solved was in the double templating so we should be able to drop that portion back into the new code without issues (with a better comment as to why we're doing things that way).

abadger:
How does this new code handle things that aren't strings?  For instance,

```
dnf:
  name: "{{ item }}"
  state: present
with_items:
  - [ "python", "python-q"]
  - [ "python3", "python3-q" ]
```

It looks like name will end up being

```
name: [ ["python", "python-q"], ["python3", "python3-q"] ]
```

willthames:
The old code only coped with strings - is that last use case actually a regression?

abadger:
if the new code puts the items together like that, it is a regression... The old code would have skipped squashing in that case.

willthames:
@abadger squashing now squashes lists of lists. I've "fixed" a bunch of tests where the test was broken for the results returned (usually because the squash happened where it wasn't before, and the conditional evaluation removed items)

abadger:
Okay, update -- will and I talked on IRC last night.  I didn't think some of the test cases were right so I worked on those for a while and merged a new set today.  The cleanups here don't pass the new tests so also decided to push a fix jsut for #15649.  Those changes are in and cherry-picked to stable-2.1 and stable-2.0 branches.

Aside from the tests, I think we still need to template the name field twice in order to catch whether it contains a loop_var or not.  Checking against the original string wouldn't catch things like a non-loop-var being templated into name.

when in doubt about the validity of squashing we should err on the side of safety and not squash.  Squashing is really a dead-end feature now that every module that squashes has to take a list type (as users can just pass a list directly instead of having to use with_items to create their list).

PR #15663, opened 2016-04-29
Dist version fix for Red Hat and more tests

robinro:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

2.2, devel
##### SUMMARY

Add test cases for centos6, redhat 6 and 7.
Fix redhat parsing (to make the testcases work and restore 2.0 behavior).
Potential fix for vmware and arch version parsing (testcases needed).
- fixes #15671

PR #15675, opened 2016-04-29
Streamline Arch Linux detection following systemd standard (via /etc/os-release)

indrajitr:
##### ISSUE TYPE

- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (archlinux-os-release b22a84a996) last updated 2016/04/29 18:30:16 (GMT -500)
  lib/ansible/modules/core: (devel b6ad3b6773) last updated 2016/04/29 16:41:32 (GMT -500)
  lib/ansible/modules/extras: (devel 3a7e4b5834) last updated 2016/04/29 16:41:54 (GMT -500)
  config file = 
```
##### SUMMARY

systemd is a first class citizen in Arch Linux variants and thus `/etc/os-release` will always have the right info [1]. We don't need to rely on `/etc/arch-release` and `allowempty` workaround.

[1] https://www.freedesktop.org/software/systemd/man/os-release.html

This is a followup to #15663

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -613,7 +613,7 @@ class Distribution(object):
         {'path': '/etc/system-release', 'name': 'Amazon'},
         {'path': '/etc/alpine-release', 'name': 'Alpine'},
         {'path': '/etc/release', 'name': 'Solaris'},
-        {'path': '/etc/arch-release', 'name': 'Archlinux', 'allowempty': True},
+        {'path': '/etc/os-release', 'name': 'Archlinux'},
         {'path': '/etc/os-release', 'name': 'SuSE'},
         {'path': '/etc/SuSE-release', 'name': 'SuSE'},
         {'path': '/etc/gentoo-release', 'name': 'Gentoo'},
@@ -628,6 +628,7 @@ class Distribution(object):
         'OracleLinux': 'Oracle Linux',
         'RedHat': 'Red Hat',
         'Altlinux': 'ALT Linux',
+        'Archlinux': 'Arch Linux',
     }
 
     # A list with OS Family members


indrajitr:
Not sure if these Travis failures are related. @abadger, any advice?

bcoca:
what about pre-systemd versions?

robinro:
Since when is arch on systemd? I guess it's long enough to ignore the pre-systemd versions since the arch support timescale is rather short.

Can you provide a testcase for arch, i.e. the output of https://github.com/ansible/ansible/blob/1d6608e84f145907667c6c264949332b6aa33356/hacking/tests/gen_distribution_version_testcase.py

Your patch doesn't yet fix arch and breaks most other distributions currently. We would need a proper implementation of parsing /etc/os-release and check for which distribution to run this on.This would be a good idea to reunite the different distribution codes. I'd be happy to have a look at it if you want to try a patch. Be sure to run it against the existing tests.

robinro:
At least the dock0/arch docker image does not have a `/etc/os-release` file, but they do have `/etc/arch-release` (but empty).

One could try `os-release` first and parse it if it exists and otherwise fall back to `arch-release` (or suse-release or whatever).

That's the same issue as with lsb-release, which does not have to be there and is not installed in many minimal setups.

djmattyg007:
There's no real concept of a "version" of Arch Linux because it's a rolling release distribution. All Arch users are expected to keep their systems up to date on a regular basis. Arch adopted systemd more than four years ago, and therefore it is more than reasonable to assume that any Arch Linux installation is using systemd by default.

That doesn't mean it should be assumed that systemd is always in use on an Arch Linux system, just that it should certainly be the default assumption.

alikins:
@indrajitr I rebase this to devel in a new pr at https://github.com/ansible/ansible/pull/22949

PR #15677, opened 2016-04-30
Fix the mapping of module_name to Locks

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

stable-2.1.0 and devel
##### SUMMARY

The mapping was reinitialized every time we forked before so we weren't sharing
the same Locks.  It also was not accounting for modules which were
directly invoked by an action plugin instead of going through the
strategy plguins.


abadger:
This was reported by @kamsz  as causing tracbacks like this: https://gist.github.com/kamsz/279350a6edfdac78d65a4afc346553d0

PR #15689, opened 2016-05-02
Restore Ansible 2.0 compatibility for include:

### LINKED ISSUE #15687: include: Ansible 2.1 breaks 2.0 compatibility
##### ISSUE TYPE
- Bug Report
##### ANSIBLE VERSION

ansible 2.1.0.0
##### OS / ENVIRONMENT

N/A
##### SUMMARY

Changes to include code break compatibility with Ansible 2.0
In Ansible 2.0 all includes have been treated as dynamic, so it was possible to skip large code paths by not processing any sub-tasks and their descendands.

Since commit 2c20579a06925a16abb95798237b97ce4987173d if the include filename string does not contain variables or the include command does not contain loops, all includes are FORCIBLY treated as static. This breaks compatibility with Ansible 2.0 playbooks and users have no way to workaround - the "static: no" keyword does not work as it is ignored.
##### EXPECTED RESULTS

Skipped includes should not be processed at all.
##### ACTUAL RESULTS

Almost all includes are treated as static.

mmatuska:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

ansible 2.2.0 (devel 1fc44e4103) last updated 2016/05/02 12:54:17 (GMT +200)
##### SUMMARY

Since commit 2c20579 compatibility for `include:` with 2.0 is broken. Ff the include filename string does not contain variables or the include command does not contain loops, all includes are FORCIBLY treated as static. Users have no way to workaround - the "static: no" keyword does not work as it is ignored.

I consider this a merge candidate for stable-2.1. Please consider at least commit 1.)

This fix consists of two commits:
1.) 438ed70a4373f1035ae90a042bc745791b40fa76 restores 2.0 compatibility by not forcing includes to be static - all are dynamic by default
2.) 434031a0ce73c3e5b630a22a6a8c9dbaf3f241e3 treat `static: no` and `static: yes` with higher priority than `task_includes_static` in ansible.cfg. This allows using `static: no` even if  `task_includes_static = yes` is defined.

Fixes: #15687

### DIFF
--- a/lib/ansible/playbook/helpers.py
+++ b/lib/ansible/playbook/helpers.py
@@ -108,13 +108,14 @@ def load_list_of_tasks(ds, play, block=None, role=None, task_include=None, use_h
                 all_vars = variable_manager.get_vars(loader=loader, play=play, task=t)
                 templar = Templar(loader=loader, variables=all_vars)
 
-                # check to see if this include is static, which can be true if:
-                # 1. the user set the 'static' option to true
+                # check to see if this include is dynamic or static:
+                # 1. the user has set the 'static' option to false or true
                 # 2. one of the appropriate config options was set
-                # 3. the included file name contains no variables, and has no loop
-                is_static = t.static or \
-                            C.DEFAULT_TASK_INCLUDES_STATIC or \
-                            (use_handlers and C.DEFAULT_HANDLER_INCLUDES_STATIC)
+                if t.static is not None:
+                    is_static = t.static
+                else:
+                    is_static = C.DEFAULT_TASK_INCLUDES_STATIC or \
+                                (use_handlers and C.DEFAULT_HANDLER_INCLUDES_STATIC)
 
                 if is_static:
                     if t.loop is not None:
--- a/lib/ansible/playbook/helpers.py
+++ b/lib/ansible/playbook/helpers.py
@@ -114,8 +114,7 @@ def load_list_of_tasks(ds, play, block=None, role=None, task_include=None, use_h
                 # 3. the included file name contains no variables, and has no loop
                 is_static = t.static or \
                             C.DEFAULT_TASK_INCLUDES_STATIC or \
-                            (use_handlers and C.DEFAULT_HANDLER_INCLUDES_STATIC) or \
-                            not templar._contains_vars(t.args.get('_raw_params')) and t.loop is None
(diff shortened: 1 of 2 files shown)


lib/ansible/playbook/helpers.py
```diff
@@ -112,10 +112,11 @@ def load_list_of_tasks(ds, play, block=None, role=None, task_include=None, use_h
                 # 1. the user set the 'static' option to true
                 # 2. one of the appropriate config options was set
                 # 3. the included file name contains no variables, and has no loop
-                is_static = t.static or \
-                            C.DEFAULT_TASK_INCLUDES_STATIC or \
-                            (use_handlers and C.DEFAULT_HANDLER_INCLUDES_STATIC) or \
-                            not templar._contains_vars(t.args.get('_raw_params')) and t.loop is None
+                if t.static is not None:
+                    is_static = t.static
```
jimi-c, line 115: What about the case where the include filename is based on a var? If there is a var in the name, we either need to check to see if the var is available now (which might still cause problems for plays where the var is changed later before the include happens), raise an error if is_static == True when there's a var in the name, or fall back to is_static = False when there's a var.

mmatuska: If an user requests `static: yes` his request should be honored, no fallbacks whatorever. So I would opt for raising an error. I am going to update my patch tomorrow.

mmatuska, line 115: You are actually doing that already a couple of lines below:

``` python
                        try:
                            include_target = templar.template(t.args['_raw_params'])
                        except AnsibleUndefinedVariable as e:
                            raise AnsibleParserError(
                                      "Error when evaluating variable in include name: %s.\n\n" \
                                      "When using static includes, ensure that any variables used in their names are defined in vars/vars_files\n" \
                                      "or extra-vars passed in from the command line. Static includes cannot use variables from inventory\n" \
                                      "sources like group or host vars." % t.args['_raw_params'],
                                      obj=task_ds,
                                      suppress_extended_error=True,
                                  )
```

jimi-c, line 115: Ahh ok, fair enough :) I knew we still had the loop check below as well, forgot we errored on templating there too.

jimi-c:
Merged, thanks!

jimi-c:
Also cherry-picked to stable-2.1.

PR #15696, opened 2016-05-03
Add get_distribution_Archlinux

jpic:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel c42501cfe7) last updated 2016/05/03 11:02:47 (GMT +200)
  lib/ansible/modules/core: (detached HEAD e78ee3b128) last updated 2016/04/29 10:01:41 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD ca310f3d15) last updated 2016/04/29 10:01:41 (GMT +200)
  config file = /home/jpic/.ansible.cfg
  configured module search path = ['/home/jpic/ansible/library', '/usr/share/ansible']
```
##### SUMMARY

Ansible would raise an exception on Archlinux hosts:

```
<!-- Paste verbatim command output here, e.g. before and after your change -->
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: AttributeError: 'Distribution' object has no attribute 'get_distribution_Archlinux'
fatal: [localhost]: FAILED! => {
    "changed": false, 
    "failed": true, 
    "module_stderr": "Traceback (most recent call last):\n  File \"/tmp/ansible_vJU6O6/ansible_module_setup.py\", line 127, in <module>\n    main()\n  File \"/tmp/ansible_vJU6O6/ansible_module_setup.py\", line 119, in main\n    data = get_all_facts(module)\n  File \"/home/jpic/env/src/ansible/lib/ansible/module_utils/facts.py\", line 3230, in get_all_facts\n    facts = ansible_facts(module, additional_subsets)\n  File \"/home/jpic/env/src/ansible/lib/ansible/module_utils/facts.py\", line 3174, in ansible_facts\n    facts.update(Facts(module).populate())\n  File \"/home/jpic/env/src/ansible/lib/ansible/module_utils/facts.py\", line 182, in __init__\n    self.facts.update(Distribution().populate())\n  File \"/home/jpic/env/src/ansible/lib/ansible/module_utils/facts.py\", line 652, in populate\n    self.get_distribution_facts()\n  File \"/home/jpic/env/src/ansible/lib/ansible/module_utils/facts.py\", line 710, in get_distribution_facts\n    distfunc = getattr(self, 'get_distribution_' + name)\nAttributeError: 'Distribution' object has no attribute 'get_distribution_Archlinux'\n", 
    "module_stdout": "", 
    "parsed": false
}

```

This patch fixes it.

Issue #15695, opened 2016-05-03
refactor distribution version detection and base it on lsb_release or os-release


robinro:
##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
lib/ansible/module_utils/facts.py

##### ANSIBLE VERSION

2.2, devel
##### CONFIGURATION

N/A
##### OS / ENVIRONMENT

Linux
##### SUMMARY

This is a follow-up to: https://github.com/ansible/ansible/pull/15221

The `distribution_*` variables get filled very differently for different distributions. We could try to unite this more. Best sources for this: `lsb_release` or `/etc/os-release`. 

Problem: Those work only if some packages are installed, like lsb, systemd, ... 

If we use them, we might need other methods as fallback when either the distribution version is too old or the package is not installed (which is likely on some minimal container setups). Also we might break backwards-compatibility, because the info in `lsb_release` often differs from the formulation in the dist-specific files.

robinro:
@bcoca in https://github.com/ansible/ansible/issues/10445#issuecomment-216382541 you mentioned you like the lsb first approach. What do you think about:
- breaking backwards compatibility when the lsb output differs from what we now output
- dealing with lsb not being installed because we're in a container or too old

bcoca:
@robinro I don't disagree, I just wanted to hold on to the illusion of a standard being followed for once ...

abadger:
bcoca clarified in slack that we do need to favor backwards compatibility over setup module cleanliness.  @robinro, I don't know if you have some ideas for making the two co-exist or if it means you want to abandon this part of making things more standardized.  (Feel free to close this ticket if it no longer appleis).

robinro:
Thanks for the input. With backwards compatibility one is restricted a lot more, but I would leave this open and maybe do more cleanup in the future. One approach would be to at least change the parsing order to try the sane stuff (lsb/os-release) first and fall back later (now the order is historic, but relevant to get the correct output). This is not a priority for me, so it might take some time until I get back to it.

slack is not open to non-core I guess?

tbenst:
Would love to see this. In Debian Stretch (Testing), I get the same problem as #10445. Any workarounds for now?

robinro:
@tbenst use the ansible_lsb_\* variables if you have a working lsb setup.

Other than that: consider to submit a PR. I'd be glad to help.

ansibot:
[...]

PR #15705, opened 2016-05-03
better fix for arch version detection

robinro:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

devel
##### SUMMARY

for Arch we only set `distribution`, so it does not need it's own function but it's enough to check the existance of the arch-release file.
fixes  #15696

This restores the 2.1 case. A more complete fix might be good, see discussion at  #15695

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -604,6 +604,10 @@ class Distribution(object):
     This is unit tested. Please extend the tests to cover all distributions if you have them available.
     """
 
+    # every distribution name mentioned here, must have one of
+    #  - allowempty == True
+    #  - be listed in SEARCH_STRING
+    #  - have a function get_distribution_DISTNAME implemented
     OSDIST_LIST = (
         {'path': '/etc/oracle-release', 'name': 'OracleLinux'},
         {'path': '/etc/slackware-version', 'name': 'Slackware'},
@@ -628,7 +632,6 @@ class Distribution(object):
         'OracleLinux': 'Oracle Linux',
         'RedHat': 'Red Hat',
         'Altlinux': 'ALT Linux',
-        'Archlinux': '',
     }
 
     # A list with OS Family members
@@ -688,12 +691,12 @@ class Distribution(object):
 
                 if not os.path.exists(path):
                     continue
+                # if allowempty is set, we only check for file existance but not content
+                if 'allowempty' in ddict and ddict['allowempty']:
+                    self.facts['distribution'] = name
+                    break
                 if os.path.getsize(path) == 0:
-                    if 'allowempty' in ddict and ddict['allowempty']:
-                        self.facts['distribution'] = name
-                        break
-                    else:
-                        continue
+                    continue
 
                 data = get_file_content(path)
                 if name in self.SEARCH_STRING:
@@ -708,13 +711,19 @@ class Distribution(object):
                     break
                 else:
                     # call a dedicated function for parsing the file content
-                    distfunc = getattr(self, 'get_distribution_' + name)
-                    parsed = distfunc(name, data, path)
-                    if parsed is None or parsed:
-                        # distfunc return False if parsing failed
-                        # break only if parsing was succesful
-                        # otherwise continue with other distributions
-                        break
+                    try:
+                        distfunc = getattr(self, 'get_distribution_' + name)
+                        parsed = distfunc(name, data, path)
+                        if parsed is None or parsed:
+                            # distfunc return False if parsing failed
+                            # break only if parsing was succesful
+                            # otherwise continue with other distributions
+                            break
+                    except AttributeError:
+                        # this should never happen, but if it does fail quitely and not with a traceback
+                        pass
+
+
 
                     # to debug multiple matching release files, one can use:
                     # self.facts['distribution_debug'].append({path + ' ' + name:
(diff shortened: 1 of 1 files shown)


lib/ansible/module_utils/facts.py
```diff
@@ -780,10 +781,6 @@ def get_distribution_OpenWrt(self, name, data, path):
         if release:
             self.facts['distribution_release'] = release.groups()[0]
 
-    def get_distribution_Archlinux(self, name, data, path):
-        self.facts['distribution'] = 'Archlinux'
-        self.facts['distribution_version'] = data
-
     def get_distribution_Alpine(self, name, data, path):
         self.facts['distribution'] = 'Alpine'
         self.facts['distribution_version'] = data
```
bcoca, line 795: wouldn't Alpine merit the same?

lib/ansible/module_utils/facts.py
```diff
@@ -707,13 +711,19 @@ def get_distribution_facts(self):
...
+                        distfunc = getattr(self, 'get_distribution_' + name)
+                        parsed = distfunc(name, data, path)
+                        if parsed is None or parsed:
+                            # distfunc return False if parsing failed
+                            # break only if parsing was succesful
+                            # otherwise continue with other distributions
+                            break
+                    except AttributeError:
+                        # this should never happen, but if it does fail quitely and not with a traceback
+                        pass
```
jpic, line 724: Would it be possible to clarify what would raise an AttributeError instead of "this should never happen" ?

Wouldn't this be perhaps a bit more precise ?

```
distfunc = getattr(self, 'get_distribution_' + name, None)
if distfunc is not None:
    parsed = distfunc(name, data, path)
    if parsed is None or parsed:
        break
```

robinro, line 724: I tried to cover this in line 607-610

lib/ansible/module_utils/facts.py
```diff
@@ -604,6 +604,10 @@ class Distribution(object):
     This is unit tested. Please extend the tests to cover all distributions if you have them available.
```
jpic, line 604: Is that still true ?

lib/ansible/module_utils/facts.py
jpic, line 724: So AttributeError here would be raised by the getattr() call, isn't the other suggested form better ?

Why do we need all this code in a try block ?

lib/ansible/module_utils/facts.py
```diff
@@ -604,6 +604,10 @@ class Distribution(object):
...
     """
 
+    # every distribution name mentioned here, must have one of
+    #  - allowempty == True
+    #  - be listed in SEARCH_STRING
+    #  - have a function get_distribution_DISTNAME implemented
```
jpic, line 610: Shouldn't we have a test for that ?

It could have caught the error caused with Arch linux if we did couldn't it ?

robinro, line 610: there is no arch test, please add one

lib/ansible/module_utils/facts.py
robinro, line 724: The try catches if the function name that got constructed does not exist.
This will NEVER happen with the existing code, because all distribution names are caught before in SEARCH_STRING or have a function defined. The try is the conservative approach to catch bugs that will be introduced in the future is some change leads to a function call that does not exist.
The failure should then not be a traceback, but a wrong setting of the distribution name/version.

lib/ansible/module_utils/facts.py
robinro, line 795: As discussed in IRC: I guess Alpine has content in it's release-file that is relevant to have in distribution_version. Arch does not (as far as I know).

robinro:
@bcoca please clarify the `needs_info`

abadger:
Talked with @bcoca on slack.  The needs_info was for the Alpine Linux questoin which was answered here.  Going to go ahead and merge.

abadger:
Thanks @robinro!

lib/ansible/module_utils/facts.py
jpic, line 724: > The try catches if the function name that got constructed does not exist.

Yes, also it would catch any AttributeError raised in `distfunc(name, data, path)`, is there any reason for this ?

Won't this harden debugging in such cases ?

lib/ansible/module_utils/facts.py
jpic, line 795: Confirmed here on lxc:

```
    "ansible_distribution_major_version": "NA", 
    "ansible_distribution_release": "NA", 
    "ansible_distribution_version": "3.3.3", 
```

Also, pretty impressive distribution to use !

Keep up the great work B)

lib/ansible/module_utils/facts.py
jpic, line 604: Anybody knows if we could make the existing test case to prevent this on other distros like arch linux or Alpine ?

lib/ansible/module_utils/facts.py
jpic, line 610: Thanks, do you mean a test case for arch linux or do you mean running the existing test suite on arch linux and perhaps more distros would cover this ?

lib/ansible/module_utils/facts.py
robinro, line 724: At least the existing function can't raise AttributeError. Feel free to open a PR for a better implementation.

lib/ansible/module_utils/facts.py
robinro, line 604: please have a look at the existing tests, It's straight-forward to extend them to other distributions. See also the discussion on the previous issues for how to do this. Let me know if you have specific questions.

PR #15716, opened 2016-05-04
Avoid duplicate system related fact calls.

jctanner:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
$ ansible --version
ansible 2.2.0 (FACTS_TUNE cd1a660bad) last updated 2016/05/04 00:06:50 (GMT -400)
  lib/ansible/modules/core: (devel 67de0675c3) last updated 2016/05/03 22:14:46 (GMT -400)
  lib/ansible/modules/extras: (devel e8391d6985) last updated 2016/05/03 22:14:54 (GMT -400)
  config file =
```
##### SUMMARY

When getting each "subset" of facts, some will load/run all the generic fact calls again. This patch attempts to eliminate that behavior and reuse previously collected facts.

Addresses https://github.com/ansible/ansible-modules-core/issues/1461

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -169,10 +169,13 @@ class Facts(object):
                  { 'path' : '/usr/local/sbin/pkg',  'name' : 'pkgng' },
                 ]
 
-    def __init__(self, module, load_on_init=True):
+    def __init__(self, module, load_on_init=True, cached_facts={}):
 
         self.module = module
-        self.facts = {}
+        if not cached_facts:
+            self.facts = {}
+        else:
+            self.facts = cached_facts    
         ### TODO: Eventually, these should all get moved to populate().  But
         # some of the values are currently being used by other subclasses (for
         # instance, os_family and distribution).  Have to sort out what to do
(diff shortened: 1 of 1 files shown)


lib/ansible/module_utils/facts.py
```diff
@@ -169,10 +169,13 @@ class Facts(object):
                  { 'path' : '/usr/local/sbin/pkg',  'name' : 'pkgng' },
                 ]
 
-    def __init__(self, module, load_on_init=True):
+    def __init__(self, module, load_on_init=True, cached_facts={}):
```
bcoca: pass None, you really want to avoid [] and {} as initializers in python functions

jctanner:
I did an out of band patch to make this patch toggled on/off with an environment variable. Here are the results with the patch enabled + disabled ...

```
$ ./stats_patches.py 
-------------------------------------
FILENAME: results_unpatched.txt
        samples: 1000
           mean: 0.267
         median: 0.189
           mean: 0.267
-------------------------------------
FILENAME: results_patched.txt
        samples: 1000
           mean: 0.199
         median: 0.141
           mean: 0.199
```

abadger:
Discussed on irc.  This looks good to me (as good as facts can look without major changes).  +1 to merge.  Be sure to add it to the changelog as a speedup of facts is sure to make some people very happy :-)

Issue #15685, opened 2016-05-02
tools that paginate show spurious `less --version` output


Habbie:
##### ISSUE TYPE
- Bug Report
##### ANSIBLE VERSION

```
ansible 2.0.2.0
  config file =
  configured module search path = Default w/o overrides
```
##### CONFIGURATION

no configuration present/necessary
##### OS / ENVIRONMENT

`Darwin marconi.local 14.5.0 Darwin Kernel Version 14.5.0: Mon Jan 11 18:48:35 PST 2016; root:xnu-2782.50.2~1/RELEASE_X86_64 x86_64` with ansible installed via brew
##### SUMMARY

in `ansible/cli/__init__.py` there is this line:

```
        elif subprocess.call('(less --version) 2> /dev/null', shell = True) == 0:
```

but on every system I have tested, the output of `--version` goes to standard out, not standard error. It looks like the `2>` should just be `>`.
##### STEPS TO REPRODUCE

```
$ ansible-galaxy search json
less 418
Copyright (C) 1984-2007 Mark Nudelman

less comes with NO WARRANTY, to the extent permitted by law.
For information about the terms of redistribution,
see the file named README in the less distribution.
Homepage: http://www.greenwoodsoftware.com/less

Found 2 roles matching your search:

 Name                            Description
 ----                            -----------
 dereulenspiegel.alfred-json     Installs alfred-json
 minderaswcraft.chef_solo_runner Runs chef-solo giving the cookbooks archive, the json attributes (usually called node.json) file and optionaly
```

```
$ ansible-doc json
 [WARNING]: module json not found in /usr/local/Cellar/ansible/2.0.2.0/libexec/lib/python2.7/site-
packages/ansible/modules:[TRUNCATED LIST OF PATHS]

less 418
Copyright (C) 1984-2007 Mark Nudelman

less comes with NO WARRANTY, to the extent permitted by law.
For information about the terms of redistribution,
see the file named README in the less distribution.
Homepage: http://www.greenwoodsoftware.com/less
```
##### EXPECTED RESULTS

Exactly what happened, minus the output of `less --version`
##### ACTUAL RESULTS

The output includes the output of `less --version`

selivan:
Bug reproduces on my system too, ansible 2.1.1 on Ubuntu 14.04 Trusty. Like in #17137, if you do `export PAGER=less`, bug disappears.

bcoca:
closing via merged #17449

PR #15720, opened 2016-05-04
fixes-#15685-tools-that-paginate-show-spurious-less-output: less --ve…

kamigerami:
fixes-#15685-tools-that-paginate-show-spurious-less-output: 

As mentioned in the issue #15685 : 
less --version outputs to standard out not to standard error so this changes the redirect from 2> to >
which does not show the less version anymore when running commands.

Before fix : 

```
$ ansible-galaxy search json
less 458 (POSIX regular expressions)
Copyright (C) 1984-2012 Mark Nudelman

less comes with NO WARRANTY, to the extent permitted by law.
For information about the terms of redistribution,
see the file named README in the less distribution.
Homepage: http://www.greenwoodsoftware.com/less

Found 2 roles matching your search:

 Name                            Description
 ----                            -----------
 dereulenspiegel.alfred-json     Installs alfred-json
 minderaswcraft.chef_solo_runner Runs chef-solo giving the cookbooks archive, the json attributes (usually called node.json) file and optionaly the location of databags, environment
(END)
```

After fix:  

```
$ ./bin/ansible-galaxy search json

Found 2 roles matching your search:

 Name                            Description
 ----                            -----------
 dereulenspiegel.alfred-json     Installs alfred-json
 minderaswcraft.chef_solo_runner Runs chef-solo giving the cookbooks archive, the json attributes (usually called node.json) file and optionaly the location of databags, environment
(END)

```

### DIFF
--- a/lib/ansible/cli/__init__.py
+++ b/lib/ansible/cli/__init__.py
@@ -476,7 +476,7 @@ class CLI(object):
                 display.display(text)
             else:
                 self.pager_pipe(text, os.environ['PAGER'])
-        elif subprocess.call('(less --version) 2> /dev/null', shell = True) == 0:
+        elif subprocess.call('(less --version) &> /dev/null', shell = True) == 0:
             self.pager_pipe(text, 'less')
         else:
             display.display(text)


tommymalmqvist:
LGTM

lib/ansible/cli/__init__.py
```diff
@@ -476,7 +476,7 @@ def pager(self, text):
                 display.display(text)
             else:
                 self.pager_pipe(text, os.environ['PAGER'])
-        elif subprocess.call('(less --version) 2> /dev/null', shell = True) == 0:
+        elif subprocess.call('(less --version) > /dev/null', shell = True) == 0:
```
bcoca: this would show any errors on screen, we probably want:

`2> /dev/null 1>&2`
or
`&> /dev/null`

kamigerami: of course! nice @bcoca, updated!

kamigerami:
whats going on with all of the failed builds all over the place @bcoca

selivan:
`&>` is bash-spesific. On systems with other default shell, `/bin/sh` for example, this won't work.
You should better use `> /dev/null 2>&1`. Created PR: https://github.com/ansible/ansible/pull/17444

Issue #15719, opened 2016-05-04
local_tmp is ignored in ansible.cfg


aholbreich:
##### ISSUE TYPE
- Bug Report
##### ANSIBLE VERSION

```
ansible-playbook 2.1.0 (devel e003ef93fc) last updated 2016/03/31 19:38:45 (GMT +200)
  lib/ansible/modules/core: (detached HEAD 18367243e4) last updated 2016/03/31 19:38:51 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD 3498cd2eae) last updated 2016/03/31 19:38:58 (GMT +200)
  config file = /home/alho/workspace_eclipse/ansible/ansible.cfg
  configured module search path = Default w/o overrides

```
##### CONFIGURATION

project directory ansible.cfg

```
[defaults]
inventory       = hosts-dev
nocows = 1
local_tmp = /tmp/.ansible/tmp
```
##### OS / ENVIRONMENT

Linux LDME (Debian)
3.16.0-4-amd64 #1 SMP Debian 3.16.7-ckt7-1 (2015-03-01) x86_64 GNU/Linux
##### SUMMARY

While running Playbook

```
ansible-playbook monitoring.yml -K -b --become-method=su
```

local_tmp seem to be ignored and files are created in $USER_HOME$.ansible/tmp
##### EXPECTED RESULTS

 if local_tmp  is specified as  /tmp/ then temp files are created in that folder.
##### ACTUAL RESULTS

If local_tmp is set to new location in the project ansible.cfg, then it's ignored and standard tmp location is used.

alikins:
> ansible-playbook 2.1.0 (devel e003ef93fc) last updated 2016/03/31 19:38:45 (GMT +200)

The 'local_tmp' config option was added after 2.1.0 in the commit 
dcc5dfdf811f24c3adbb879e77a1bf8789bf1738. Version 2.1.0 (devel e003ef93fc) 
does not support 'local_tmp'.

(It is referenced at http://docs.ansible.com/ansible/intro_configuration.html#local-tmp which is confusing since it those docs are generated from the latest versions. )

alikins:
docsite/rst/intro_configuration.rst needs a 'added_in' note about local_tmp, that would help avoid confusion.

aholbreich:
Ok, yes confusing... But thank you. Now i know...

aholbreich:
> docsite/rst/intro_configuration.rst needs a 'added_in' note about local_tmp

I can try...

[...]

PR #15730, opened 2016-05-04
Add a doc note that local_tmp was added in 2.1

alikins:
##### ISSUE TYPE
- Docs Pull Request
##### SUMMARY

Add a docs notes about the version local_tmp was added in.
As per https://github.com/ansible/ansible/issues/15719

### DIFF
--- a/docsite/rst/intro_configuration.rst
+++ b/docsite/rst/intro_configuration.rst
@@ -457,6 +457,8 @@ Ansible knows how to look in multiple locations if you feed it a colon separated
 local_tmp
 =========
 
+.. versionadded:: 2.1
+
 When Ansible gets ready to send a module to a remote machine it usually has to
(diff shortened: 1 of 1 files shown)


bcoca:
its 2.2 and probably better to use version_added construct

alikins:
It's been a while, but we discussed this patch and decided that the option was added in 2.1.

docsite/rst/intro_configuration.rst
```diff
@@ -468,6 +470,10 @@ If you'd like to change that, you can do so by altering this setting::
 
 Ansible will then choose a random directory name inside this location.
 
+.. note::
```
bcoca: the note seems to be overkill as we already have version_added

alikins:
updated and rebased

abadger:
Thanks @alikins!

Issue #6941, opened 2014-04-10
Pycrypto error PowmInsecureWarning: Not using mpz_powm_sec.  You should rebuild using libgmp >= 5 to avoid timing attack vulnerability.


goozbach:
##### Issue Type:

Bug Report
##### Ansible Version:

```
[root@vagrant-centos-6_4 ansible]# ansible --version
ansible 1.5.4
```
##### Environment:

Ansible-host: CentOS 6.5
Ansible-client: CentOS 6.5
##### Summary:

This is most likely a bug in pycrypto  but it seems to be related to or the same as #276 or #6190 
##### Steps To Reproduce:
1. fresh install of centos 6.5
2. rpm -ivh http://url/to/epel-release
3. yum -y install ansible --enablerepo='epel*'
4. create localhost ssh key and authorized keys
5. install goozbach.EPEL galaxy role
6. create following playbook:
   
   ```
   ---
   - name: master playbook
     hosts: all
     user: root
     roles:
       - { role: goozbach.EPEL, epel_enabled: 1, epel_testing_enabled: 1 }
   ```
7. run your ansible playbook
##### Expected Results:

Pycrypto shouldn't show warning message.
##### Actual Results:

``````
```
[root@vagrant-centos-6_4 ansible]# ansible-playbook site.yml
/usr/lib64/python2.6/site-packages/pycrypto-2.6.1-py2.6-linux-x86_64.egg/Crypto/Util/number.py:57: PowmInsecureWarning: Not using mpz_powm_sec.  You should rebuild using libgmp >= 5 to avoid timing attack vulnerability.
  _warn("Not using mpz_powm_sec.  You should rebuild using libgmp >= 5 to avoid timing attack vulnerability.", PowmInsecureWarning)
<SNIP all good output>
```
``````

**Comment by goozbach** [comment]
@ansibot done!

**Comment by jctanner** [comment]
Reproduced

```
# ansible-playbook -i inventory site.yml 1>/dev/null
/usr/lib64/python2.6/site-packages/pycrypto-2.6.1-py2.6-linux-x86_64.egg/Crypto/Util/number.py:57: PowmInsecureWarning: Not using mpz_powm_sec.  You should rebuild using libgmp >= 5 to avoid timing attack vulnerability.
  _warn("Not using mpz_powm_sec.  You should rebuild using libgmp >= 5 to avoid timing attack vulnerability.", PowmInsecureWarning)
```

jctanner:
[...]

PR #15731, opened 2016-05-04
Catch DistributionNotFound when pycrypto is absent

mavit:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.0.1.0
```
##### SUMMARY

On Solaris 11, module `pkg_resources` throws `DistributionNotFound` on import if `cryptography` is installed but `pycrypto` is not.  This change causes that situation to be handled gracefully.

I'm not using Paramiko or Vault, so I my understanding is that I don't need `pycrpto`.  I could install `pycrypto` to make the error go away, but:
- The latest released version of `pycrypto` doesn't build cleanly on Solaris (https://github.com/dlitz/pycrypto/issues/184).
- Solaris includes an old version of GMP that triggers warnings every time Ansible runs (https://github.com/ansible/ansible/issues/6941).  I notice that I can silence these warnings with `system_warnings` in `ansible.cfg`, but not installing `pycrypto` seems like a safer solution.

### DIFF
--- a/lib/ansible/parsing/vault/__init__.py
+++ b/lib/ansible/parsing/vault/__init__.py
@@ -68,7 +68,7 @@ try:
     from cryptography.hazmat.primitives.kdf.pbkdf2 import PBKDF2HMAC
     from cryptography.hazmat.backends import default_backend
     HAS_PBKDF2HMAC = True
-except ImportError:
+except Exception:
     pass
 
 from ansible.compat.six import PY3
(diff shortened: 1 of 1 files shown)


nitzmahone:
Amenable to this, but would prefer that you check for both ImportError and DistributionNotFound rather than swallowing anything typed Exception, eg:

```
except (ImportError, DistributionNotFound):
```

mavit:
The snag there is that `DistributionNotFound` is provided by `pkg_resources`, but if you try to import `pkg_resources` it throws `DistributionNotFound`.  I feel like there aught to be a way around that, but I don't know what it would be.

alikins:
maybe something like

```
try:
    <cryptography import stuff here>
except ImportError:
   pass
except Exception as e:
     if e.__class__.__name__ != 'pkg_resources.DistributionNotFound':
         raise
```

Not very pretty, but alas.

Issue #15702, opened 2016-05-03
SSH keys aren't checked against known_hosts when using networking modules


ogenstad:
##### ISSUE TYPE
- Feature Idea
- Security issue
##### ANSIBLE VERSION

```
2.1
```
##### CONFIGURATION

N/A
##### OS / ENVIRONMENT

N/A
##### SUMMARY

The shared module_util [shell.py](https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/shell.py) which is used by the different network modules doesn't validate the ssh key on the target device.

``` Python
    def open(self, host, port=22, username=None, password=None,
            timeout=10, key_filename=None, pkey=None, look_for_keys=None,
            allow_agent=False):

        self.ssh = paramiko.SSHClient()
        self.ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
```

This works in a trusted environment but is a security issue when connecting through a public network.

By default it would be good if system known hosts and user known hosts are loaded. Then it's up for debate regarding what should happen by default if it should use autoadd or autoreject.

Another question to ask is how to handle the situation when a user needs to confirm a new or changed ssh key. Probably the module should fail and insert information about the keys (new or changed) to a variable which is available through Ansible.

Perhaps a new parameter common among modules called something like key_policy with these options
- ignore - Ignore all ssh keys (like the current module)
- strict - Reject unknown or changed keys
- loose - Auto add unknown keys fail on changed keys

Qalthos:
This should be complete with #15736.

PR #15736, opened 2016-05-05
Provide options for SSH key checking

Qalthos:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
2.x
```
##### SUMMARY

Addresses #15702 by adding key_policy parameter to Shell.open

A value of "ignore" preserves the current behavior of automatically adding unknown host keys with no known hosts defined
A value of "strict" loads known host keys from the host and rejects hosts it does not recognize.
Any other value loads the system host keys, but does not reject unknown hosts, only changed hosts.

### DIFF
--- a/lib/ansible/module_utils/shell.py
+++ b/lib/ansible/module_utils/shell.py
@@ -16,6 +16,7 @@
 # You should have received a copy of the GNU General Public License
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 #
+import os
 import re
 import socket
 import time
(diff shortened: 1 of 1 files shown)


ogenstad:
``` Python
if key_policy != "ignore":
    self.ssh.load_system_host_keys()
```

Somewhere under the above lines it would be good to have something like:

``` Python
try:
    self.ssh.load_host_keys('~/.ssh/known_hosts')
except:
    pass
```

Most users won't be able to write to the system known_hosts file and will need to add new hosts to their own file.

privateip:
+1

gundalow:
This was discussed during Network meeting Wednesday 24th August:

```
[17:44] < privateip> +1 for loose as the default for 2.2
[17:45] <@gundalow> caphrim007: mhite stianvi What are your thoughts?
[17:45] <@gundalow> + anyone else lurking in here :)
[17:45] < mhite> I'm ok with loose
[17:46] < skg_net> +1 loose as default as well
[17:46] < caphrim007> i concur
```

PR #15756, opened 2016-05-06
Fix unititests with py.test

robinro:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

devel, 2.2
##### SUMMARY

This makes the unittests compatible with py.test, by using absolute imports in combination with reload. py.test is rather restrictive about the use of reload.

Also for the async test I reduce the sleep time (from 3sec to 0.1sec). For me this saves more than 50% of the total runtime of the unittests and makes them even more attractive to run.

### DIFF
--- a/test/units/module_utils/basic/test_run_command.py
+++ b/test/units/module_utils/basic/test_run_command.py
@@ -30,8 +30,7 @@ from ansible.compat.tests.mock import call, MagicMock, Mock, patch, sentinel
 
 from units.mock.procenv import swap_stdin_and_argv
 
-from ansible.module_utils import basic
-from ansible.module_utils.basic import AnsibleModule
+import ansible.module_utils.basic
 
 class OpenBytesIO(BytesIO):
     """BytesIO with dummy close() method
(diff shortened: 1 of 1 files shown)


robinro:
@abadger does this fix work for you? I saw an old comment in IRC that you tried it.

robinro:
After the work in  #15782 the changes to get py.test working are now rather short. Only one place needs absolute imports (don't ask me why py.test fails with from foo import bar).

@abadger This version now runs py.test for me, please have a look

```
 403 passed, 3 skipped, 2 pytest-warnings in 6.63 seconds
```

abadger:
btw, jimi-c signed off on the async changes so you can add those in too.

yes, this works.  Not sure what the difference is here but it looks fine.

Since you'll add the async changes back as well, one thing to change is the commit message -- both the old and new imports are absolute imports as they start with a module in sys.path.  Maybe a commit message like "Don't use 'from ansible.module_utils import foo' style here as it breaks py.test".   If you know why it breaks py.test, that's even better information... but if not, I think that's sufficient.

robinro:
Updated the commit message to your suggestion. I don't know why py.test fails with the other import style.

For the async part I opened a separate PR earlier: https://github.com/ansible/ansible/pull/15797 
It's unrelated and I expected the py.test cleanup to take longer.

PR #15758, opened 2016-05-06
Make it clear that there is something wrong with the "inventory file".

nhooey:
##### ISSUE TYPE
- Docs Pull Request
##### ANSIBLE VERSION

```
ansible 2.0.2.0
```
##### SUMMARY

This fixes the error message in two ways:
1. It quotes the inventory filename so it doesn't get confused with other words in the sentence
2. It specifies that the file in question is an "inventory file" to give the error message some context

### DIFF
--- a/lib/ansible/inventory/dir.py
+++ b/lib/ansible/inventory/dir.py
@@ -59,8 +59,11 @@ def get_file_parser(hostsfile, groups, loader):
             parser = InventoryScript(loader=loader, groups=groups, filename=hostsfile)
             processed = True
         except Exception as e:
-            myerr.append("The file %s is marked as executable, but failed to execute correctly. " % hostsfile + \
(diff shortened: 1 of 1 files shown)


alikins:
Seems reasonable to me.

For ref:

Before

```
ERROR! The file inventory.sh looks like it should be an executable inventory script, but is not marked executable. Perhaps you want to correct this with `chmod +x inventory.sh`?
```

After

```
ERROR! The inventory file 'inventory.sh' looks like it should be an executable inventory script, but is not marked executable. Perhaps you want to correct this with `chmod +x 'inventory.sh'`?
```

alikins:
@nhooey A variation of this got applied in https://github.com/ansible/ansible/pull/17589
(2c54a8471fbc9614ec59bebfb4207ce392605a5b) so I'm going to close this.

PR #15774, opened 2016-05-09
Enable logging for development mode.

chouseknecht:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (docker e50dd6fb7b) last updated 2016/05/09 04:45:51 (GMT -400)
  lib/ansible/modules/core: (detached HEAD 1f5cf669dd) last updated 2016/05/08 18:16:21 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 431591c2b4) last updated 2016/05/08 18:16:24 (GMT -400)
  config file =
```
##### SUMMARY

Turn debug logging on/off via constant DEVELOPMENT_MODE

### DIFF
--- a/lib/ansible/constants.py
+++ b/lib/ansible/constants.py
@@ -138,7 +138,8 @@ DEFAULT_PATTERN           = get_config(p, DEFAULTS, 'pattern', None, None)
 
 #### GENERALLY CONFIGURABLE THINGS ####
 DEFAULT_DEBUG             = get_config(p, DEFAULTS, 'debug',            'ANSIBLE_DEBUG',            False, boolean=True)
-DEFAULT_HOST_LIST         = get_config(p, DEFAULTS,'inventory', 'ANSIBLE_INVENTORY', DEPRECATED_HOST_LIST, ispath=True)
+DEFAULT_REMOTE_LOG_TARGET = get_config(p, DEFAULTS, 'remote_log_target', 'ANSIBLE_REMOTE_LOG_TARGET', None, ispath=True)
+DEFAULT_HOST_LIST         = get_config(p, DEFAULTS, 'inventory',        'ANSIBLE_INVENTORY', DEPRECATED_HOST_LIST, ispath=True)
 DEFAULT_MODULE_PATH       = get_config(p, DEFAULTS, 'library',          'ANSIBLE_LIBRARY',          None, ispathlist=True)
 DEFAULT_ROLES_PATH        = get_config(p, DEFAULTS, 'roles_path',       'ANSIBLE_ROLES_PATH',       '/etc/ansible/roles', ispathlist=True)
 DEFAULT_REMOTE_TMP        = get_config(p, DEFAULTS, 'remote_tmp',       'ANSIBLE_REMOTE_TEMP',      '$HOME/.ansible/tmp')
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -54,6 +54,8 @@ import pwd
 import platform
 import errno
 import datetime
+import logging
+
 from itertools import repeat, chain
 
 try:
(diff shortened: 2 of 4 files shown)


bcoca:
this is something we should enable for all modules, not just the docker ones

chouseknecht:
As suggested, extended this PR to provide a logging framework for all modules. The logging happens in basic.py using the Python logging module. Modules will call AnsibleModule.debug(msg, pretty_print=<bool>). If logging is enabled, the message will be sent to a file.

Enable logging by setting constants: debug=true and remote_log_target=<path to file>. When both debug and remote_log_target are truthy, log messages will be sent to file specified by remote_log_target.

Messages sent to AnsibleModule.debug() are sanitized by passing the message text through remove_values. When the pretty_print option is true, the message text is passed through json.dumps(), allowing formatting of structured data, and then passed through remov_values().

lib/ansible/module_utils/basic.py
```diff
@@ -1518,9 +1530,16 @@ def _log_to_syslog(self, msg):
             syslog.openlog(str(module), 0, facility)
             syslog.syslog(syslog.LOG_INFO, msg)
 
-    def debug(self, msg):
+    def debug(self, msg, pretty_print=False):
         if self._debug:
-            self.log(msg)
+            if self.debug_logger:
+                if pretty_print:
+                    msg_txt = json.dumps(msg, sort_keys=True, indent=4, separators=(',', ': '))
```
alikins, line 1537: This could be done in the logging Formatter (and/or a logging.Filter). And then the modules
could just pass objects to self.debug_logger.debug(), and the handler/formatter decides how to pretty print.

lib/ansible/module_utils/basic.py
```diff
@@ -624,6 +629,10 @@ def __init__(self, argument_spec, bypass_checks=False, no_log=False,
         if not self.no_log and self._verbosity >= 3:
             self._log_invocation()
 
+        if not self.no_log and self._debug and self._remote_log_target:
+            self.debug_logger = logging.getLogger(self.__class__.__name__)
```
alikins, line 633: I'd like a shorter name[1] here, but it looks like self.log and self.debug are taken.
self.dl is likely too far.

[1] since there will be a lot of `self.debug_logger.debug('foo %s %s', obj1, otherobj)`
and `self.debug_logger.exception(e)`

alikins, line 633: And for new style ziploader modules, a module scope 'log' object could work.

```
log = logging.getLogger(__name__)      # <-- does __name__ get weird on remote modules?
```

Then instead of `self.debug_logger.debug(foo)`, it would be `log.debug(foo)`

That would remove the need for the module introspect in AnsibleModule.log(). Then potentially of the rest of AnsibleModule.log() becomes a matter of which logging Handler is setup.

lib/ansible/constants.py
```diff
@@ -138,7 +138,8 @@ def load_config_file():
 
 #### GENERALLY CONFIGURABLE THINGS ####
 DEFAULT_DEBUG             = get_config(p, DEFAULTS, 'debug',            'ANSIBLE_DEBUG',            False, boolean=True)
-DEFAULT_HOST_LIST         = get_config(p, DEFAULTS,'inventory', 'ANSIBLE_INVENTORY', DEPRECATED_HOST_LIST, ispath=True)
+DEFAULT_REMOTE_LOG_TARGET = get_config(p, DEFAULTS, 'remote_log_target', 'ANSIBLE_REMOTE_LOG_TARGET', None, ispath=True)
```
alikins, line 141: Add to examples/ansible.cfg

chouseknecht:
Working on a revision to this. Adrian walked me through his comments and thoughts on making this better. Should have something later today or tomorrow morning.

PR #15784, opened 2016-05-09
When walking through module directories, always follow symlinks.

### LINKED ISSUE #15783: Module loader doesn't allow for symlinks under modules/
##### ISSUE TYPE
- Bug Report
##### ANSIBLE VERSION

```
$ ansible --version
ansible 2.2.0 (devel 30e5999812) last updated 2016/05/09 10:56:55 (GMT -400)
  lib/ansible/modules/core: (devel 1f5cf669dd) last updated 2016/05/09 10:54:28 (GMT -400)
  lib/ansible/modules/extras: (devel bb965eebee) last updated 2016/05/09 10:43:13 (GMT -400)
  config file = 
```
##### CONFIGURATION

N/A
##### OS / ENVIRONMENT

CentOS 7
##### SUMMARY

Under lib/ansible/modules/, directories are traversed with os.walk. If the directories are not symlinks, they will be returned as possible module locations. If the directories are symlinks, os.walk ignores them and prevents usage of the modules found below.
##### STEPS TO REPRODUCE
1. Delete lib/ansible/modules/core
2. Symlink an ansible-modules-core checkout to lib/ansible/modules/core
3. ansible-doc -l and check for any core modules
4. None will be listed
##### EXPECTED RESULTS

All modules are listed in ansible-doc -l and are also available to ansible+ansible-playbook.
##### ACTUAL RESULTS

The modules are not found.

jctanner:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
$ ansible --version
ansible 2.2.0 (devel 30e5999812) last updated 2016/05/09 10:56:55 (GMT -400)
  lib/ansible/modules/core: (devel 1f5cf669dd) last updated 2016/05/09 10:54:28 (GMT -400)
  lib/ansible/modules/extras: (devel bb965eebee) last updated 2016/05/09 10:43:13 (GMT -400)
  config file = 
```
##### SUMMARY

Module loader can traverse multiple subdirectories for module files. Currently, it is limited by the default behavior of os.walk and can not traverse through symlinks. This makes it difficult to swap out various module checkouts under ansible without managing submodules. If I want to quickly compare a single ansible checkout with two different checkouts of modules-core, I really have to have 2 separate checkouts of ansible. 

Fixes #15783

### DIFF
--- a/lib/ansible/plugins/__init__.py
+++ b/lib/ansible/plugins/__init__.py
@@ -136,7 +136,7 @@ class PluginLoader:
     def _all_directories(self, dir):
         results = []
         results.append(dir)
-        for root, subdirs, files in os.walk(dir):
+        for root, subdirs, files in os.walk(dir, followlinks=True):
            if '__init__.py' in files:
                for x in subdirs:
                    results.append(os.path.join(root,x))


abadger:
Looks good.  Shouldn't break anything and makes it slightly more robust.  +1 to merge.

alikins:
Looking through the history, I don't see any reason that needed to or intended to ignore symlinks.

Only caveats is that 'followsymlinks' is python 2.6+, but thats fine afaik.

Maybe worth adding a unit/integration test for recursive symlinks. Based
on note at https://docs.python.org/2/library/os.html#os.walk

```
Note

Be aware that setting followlinks to True can lead to infinite recursion if a link points to a parent directory of itself. walk() does not keep track of the directories it visited already. 
```

But 'shipit' from me.

PR #15797, opened 2016-05-10
reduce async timeout in unittests

robinro:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

devel
##### SUMMARY

reduce from 3 sec to 0.1 sec; the 3 sec sleep was about half of the total unittest time on my development machine...


robinro:
@abadger this was the second change I had included with the py.test fixes, which is actually unrelated
According to IRC you and jimi-c already looked at that part so it should be merge-ready.

nitzmahone:
faster tests FTW- thanks!

PR #15798, opened 2016-05-10
Switch to a different url for testing SNI right now.

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### SUMMARY

Fix SNI tests for uri by using the same url as get_url tests.


gundalow:
I've already got a note in my Testing Plan around test stability and removing external dependencies, which includes this specific case

abadger:
<nod> Yeah -- once we get rid of external dependencies we'll be in control of things here :-)

PR #15811, opened 2016-05-10
Use httptester docker image for http tests

sivel:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
v2
```
##### SUMMARY

As recently evidenced, when we rely on external services for http tests, we often run into connection time outs, rate limits, etc...

This PR will spin up a docker container (currently using sivel/httptester) and links it into the test containers.

The httptester image creates a CA, and certificates needed for the tests.  It also contains httpbin, and nginx so we get a full http/https/sni/httpbin/badssl type of testing capabilities.

The names such as `ansible.http.tests` is made available through docker linking, and reference the container running nginx/httpbin.

If desired we can pull the docker image in from https://github.com/sivel/docker-images/tree/master/httptester and make it available from `ansible/ansible:httptester` or similar.

It will probably take a while for the builds to complete.  Prior build to me squashing commits: https://travis-ci.org/sivel/ansible/builds/129248935

### DIFF
--- a/test/integration/destructive.yml
+++ b/test/integration/destructive.yml
@@ -21,4 +21,5 @@
     - { role: test_zypper, tags: test_zypper}
     - { role: test_zypper_repository, tags: test_zypper_repository}
     - { role: test_uri, tags: test_uri }
+    - { role: test_get_url, tags: test_get_url }
     - { role: test_apache2_module, tags: test_apache2_module }
--- a/test/integration/non_destructive.yml
+++ b/test/integration/non_destructive.yml
@@ -38,7 +38,6 @@
     - { role: test_command_shell, tags: test_command_shell }
     - { role: test_script, tags: test_script }
     - { role: test_authorized_key, tags: test_authorized_key }
-    - { role: test_get_url, tags: test_get_url }
     - { role: test_embedded_module, tags: test_embedded_module }
     - { role: test_add_host, tags: test_add_host }
     - { role: test_binary, tags: test_binary }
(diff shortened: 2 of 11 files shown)


nitzmahone:
This is awesome- my only concern with this is that it appears to break running these integration tests outside of our travis docker env. We should probably variable-ize these things and have the prepare_http_tests role override the URLs appropriately, but only run that role when container_http_tests=true or something. (so that the default behavior when running outside the container env is still to hit the external endpoints)

sivel:
It's not really tied to travis, but more requires the use of docker, such as running tests via `test/utils/run_tests.sh`.

These are destructive tests anyway, or at least the `test_uri` tests are as it installs and uninstalls packages, so I don't really recommend running outside of a container anyway.

However, I can utilize some combination of `defaults` and potentially `set_fact` to try and achieve something that is more flexible.

I'll look into that tomorrow.

Otherwise we could always merge now and do that as another enhancement so we can at least get tests passing sooner.

nitzmahone:
I'm good with "merge now" (well, as soon as travis gets around to running) and figure out the other stuff later...

sivel:
@nitzmahone I had an idea of how to get what you were asking for, so it has been implemented in https://github.com/ansible/ansible/pull/15811/commits/0033b1133397064686e61227bee2508931017e0f

It uses env vars created by docker when linking a container to determine if we have access to httptester or not.  If not, then the defaults (public test sites) are used, otherwise, we override with the httptester hostnames.

jctanner:
@sivel thank you very much! I've gotten another successful run via run_tests.sh in your latest patch and then hit the expected SNI failure when running in a vagrant box. Thanks again for getting this done!!

PR #15820, opened 2016-05-11
Key names in the ansible_local variable are converted into lower case

jinesh-choksi:
##### ISSUE TYPE
- Docs Pull Request
##### ANSIBLE VERSION

```
ansible 2.0.2.0
```
##### SUMMARY

Added a note to the Ansible documentation to highlight that key names for local facts will be stored in lower case inside the ansible_local variable due to Ansible's use of the Python ConfigParser module which converts option names into lower case by default.


dharmabumstead (approved):
Looks good. Thanks.

PR #15825, opened 2016-05-12
If we can't squash for any reason, then simply do not optimize the items loop.

### LINKED ISSUE #15649: yum module fails with 'dict object' has no attribute '\\x00$' when used in loop and with dict lookup

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

2.x
##### SUMMARY

Catch any traceback in the squashing code.

Also add more squashing testcases

Fixes #15649

### DIFF
--- a/test/units/executor/test_task_executor.py
+++ b/test/units/executor/test_task_executor.py
@@ -180,8 +180,10 @@ class TestTaskExecutor(unittest.TestCase):
 
         mock_host = MagicMock()
 
+        loop_var = 'item'
+
         def _evaluate_conditional(templar, variables):
-            item = variables.get('item')
+            item = variables.get(loop_var)
             if item == 'b':
                 return False
             return True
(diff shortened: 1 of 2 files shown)


willthames:
Can we backport this to 2.0 also?

test/units/executor/test_task_executor.py
```diff
@@ -230,9 +232,31 @@ def _evaluate_conditional(templar, variables):
...
+        self.assertEqual(new_items, ['a', 'b', 'c'])
+
+        # Maybe should raise an error in this case.  The user would have to specify:
+        # - yum: name="{{ packages[item] }}"
+        #   with_items:
+        #     - ['a', 'b']
+        #     - ['foo', 'bar']
+        # you can't use a list as a dict key so that would probably throw
+        # an error later.  If so, we can throw it now instead.
+        # Squashing in this case would not be intuitive as the user is being
```
willthames, line 247: Squashing is fine in this case, I think, if a list of lists is effectively just another way of writing a list

e.g. sometimes you might want to do:

```
yum: 
  name: "{{item}}"
  state: present
with_items:
- "{{list_of_default_rpms}}"
- "{{list_of_custom_rpms}}"
```

You can use union to join the lists of course, but I'm not sure what other use case drives this test that isn't amenable to squashing

abadger:
I'll try cherry-picking after merging.  I'm not sure that we'll release another 2.0.x, though.

abadger:
jimi-c found one issue at the meeting today (removed the line that called _squash_items by mistake).  I fixed.  PR is passing travis tests on my private repo.  Going to merge and cherry-pick so it's ready for tomorrow's rc2.

abadger:
Okay, had to resolve some conflicts but this is now merged to stable-2.1 and stable-2.0 as well.

PR #15832, opened 2016-05-12
Adding Config Encoder filters

jtyr:
##### ISSUE TYPE

Feature Pull Request
##### ANSIBLE VERSION

For Ansible v2.2
##### SUMMARY

This is a set of Jinja2 filters which allows to create configuration files in various formats from a variable. The benefit of using these filters is that the variable can be dynamically built (merging dicts, lists) and therefore there is no need to update the template file, only to override the variable. This allows to create Ansible roles with truly universal configuration that helps to reduce fragmentation of Ansible roles.

More details about the [motivation](https://github.com/jtyr/ansible/blob/jtyr-config_encoders/docsite/rst/playbooks_filters_config_encoders.rst#motivation) together with [examples](https://github.com/jtyr/ansible/blob/jtyr-config_encoders/docsite/rst/playbooks_filters_config_encoders.rst#example) and full [documentation](https://github.com/jtyr/ansible/blob/jtyr-config_encoders/docsite/rst/playbooks_filters_config_encoders.rst#supported-encoders) of all the filters is part of the PR.

### DIFF
--- a/docsite/rst/playbooks_filters.rst
+++ b/docsite/rst/playbooks_filters.rst
@@ -457,6 +457,46 @@ which will produce this output::
     # host: myhost
     #
 
+.. _config_encoders:
+
+Config Encoder Filters
+----------------------
+
+.. versionadded:: 2.2
+
+Config Encoder filters were developed to facilitate creation of configuration
+files from a YAML data structure. Imagine the following INI file::
+
+    [section1]
+    option11=value11
+    option12=value12
+
+Such configuration file can be described as a YAML data structure::
+
+    myapp_config:
+      section1:
+        option11: value11
+        option12: value12
+
+The variable is then used together with the ``encode_ini`` Config Encoder
+filter in the template file ``myapp.cfg.j2`` like this::
+
+    {{ myapp_config | encode_ini }}
+
+And finally, the template file is used in a task like this::
+
+    - name: Create config file
+      template:
+        src: myapp.cfg.j2
+        dest: /etc/myapp/myapp.cfg
+
+When the task is executed, it creates exactly the same file as the
+original INI file.
+
+More information about Config Encoder filters including all supported
+configuration file formats and complete usage guide can be found in
+:doc:`playbooks_filters_config_encoders`.
+
 .. _other_useful_filters:
 
 Other Useful Filters
--- a/docsite/rst/playbooks_filters_config_encoders.rst
+++ b/docsite/rst/playbooks_filters_config_encoders.rst
@@ -1,0 +1,1154 @@
+Config Encoder Filters
+======================
+
+.. versionadded:: 2.2
+
+
+Motivation
+----------
+
+Ansible Galaxy contains a lot of useful roles. Some of them exist in
+many variations which differ only by their parameterization. The
+parameterization is often used mainly in templates which generate the
+configuration file. A good example such issues are roles for Nginx of
+which you can find almost 200 in the Ansible Galaxy.
+
+Nginx is possible to configure in infinite number of ways and therefore
+is almost impossible to create an Ansible template file which would
+capture all possible variations of the configuration. Even if a suitable
+roles is found, users often want to customize even more. This is where
+people normally clone the role and add parameters they are missing. Some
+people try to get the change back into the original role by creating a
+pull request (PR) but sometimes such change is not accepted by the
+maintainer of the original role and the user ends up maintaining his/her
+own clone forever.
+
+This is why the Config Encoder filters were developed to facilitate the
+creation of Ansible roles with universal configuration. The structure of
+the configuration file is described as a YAML data structure stored in a
+variable. The variable together with he Config Encoder filter is then
+used in the template file which is used to generate the final
+configuration file. This approach allows to shift the paradigm of
+thinking about configuration files as templates to thinking about them as
+data structures. The data structure can be dynamically generated which
+allows to create truly universal configuration.
+
+
+Example
+-------
+
+Imagine the following INI file::
+
+    [section1]
+    option11=value11
+    option12=value12
+
+Such configuration file can be described as a YAML data structure::
+
+    myapp_config:
+      section1:
+        option11: value11
+        option12: value12
+
+The variable is then used together with the ``encode_ini`` Config Encoder
+filter in the template file ``myapp.cfg.j2`` like this::
+
+    {{ myapp_config | encode_ini }}
+
+And finally, the template file is used in a task like this::
+
+    - name: Create config file
+      template:
+        src: myapp.cfg.j2
+        dest: /etc/myapp/myapp.cfg
+
+When the task is executed, it creates exactly the same file as the
+original INI file.
+
+So we can describe the configuration as a data structure which is then
+converted into the final configuration file format with the Config
+Encoder filter.
+
+In order to change the above configuration, we would have to overwrite
+the ``myapp_config`` which is not very practical. Therefore we break the
+monolithic variable into a set of variables which will allow us to change
+any part of the configuration without the need to overwrite the whole
+data structure::
+
+    myapp_config_section1_option11: value11
+    myapp_config_section1_option12: value12
+
+    myapp_config_section1__default:
+      option11: "{{ myapp_config_section1_option11 }}"
+      option12: "{{ myapp_config_section1_option12 }}"
+
+    myapp_config_section1__custom: {}
+
+    myapp_config_default:
+      section1: "{{
+        myapp_config_section1__default.update(myapp_config_section1__custom) }}{{
+        myapp_config_section1__default }}"
+
+    myapp_config__custom: {}
+
+    myapp_config: "{{
+      myapp_config__default.update(myapp_config__custom) }}{{
+      myapp_config__default }}"
+
+Like this, if we want to change the value of the ``option11``, we only
+override the variable ``myapp_config_section1_option11``::
+
+    myapp_config_section1_option11: My new value
+
+If we want to add a new option into the ``section1``, we add it into the
+variable ``myapp_config_section1__custom`` which is then merged with the
+default list of options::
+
+    myapp_config_section1__custom:
+      section13: value13
+
+And if we want to add a new section, we add it into the variable
+``myapp_config__custom`` which is then merged with the default list of
+sections::
+
+    myapp_config__custom:
+      section2:
+        option21: value21
+
+The above is showing an example for INI configuration files only but the
+same principle is possible to use for all the supported Config Encoders
+listed bellow.
+
+
+Supported encoders
+------------------
+
+The following is the list of supported Config Encoder filters. Each
+filter requires special data structure as its input. Each filter also has
+a set of parameters which can modify the behaviour of the filter.
+
+
+encode_apache
+^^^^^^^^^^^^^
+
+This filter helps to create configuration in the format used by Apache
+web server. The expected data structure is the following::
+
+    my_apache_vhost:
+      content:
+        - sections:
+          - name: VirtualHost
+            param: "*:80"
+            content:
+              - options:
+                - DocumentRoot: /www/example1
+                - ServerName: www.example.com
+                - ErrorLog: /var/log/httpd/www.example.com-error_log
+                - CustomLog:
+                  - /var/log/httpd/www.example.com-access_log
+                  - common
+                - "#": Other directives here ...
+
+The variable starts with ``content`` which can contain list of
+``sections`` or ``options``. ``sections`` then contain list of individual
+sections which has the ``name``, ``param`` and ``content`` parameter. The
+``content`` can again contain a list of `sections`` or ``options``.
+
+The above variable can be used in the template file like this::
+
+    {{ my_apache_vhost | encode_apache }}
+
+The output of such template would be::
+
+    <VirtualHost *:80>
+      DocumentRoot /www/example1
+      ServerName www.example.com
+      ErrorLog /var/log/httpd/www.example.com-error_log
+      CustomLog /var/log/httpd/www.example.com-access_log common
+      # "Other directives here ..."
+    </VirtualHost>
+
+The filter can have the following parameters:
+
+- ``convert_bools=false``
+
+  Indicates whether Boolean values presented as a string should be
+  converted to a real Boolean value. For example ``var1: 'True'`` would
+  be represented as a string but by using the ``convert_bools=true`` it
+  will be converted into Boolean like it would be defined like ``var1:
+  true``.
+
+- ``convert_nums=false``
+
+  Indicates whether number presented as a string should be converted to
+  number. For example ``var1: '123'`` would be represented as a string
+  but by using the ``convert_nums=true`` it will be converted it to a
+  number like it would be defined like ``var1: 123``. It's also possible
+  to use the YAML type casting to convert string to number (e.g. ``!!int
+  "1234"``, ``!!float "3.14"``).
+
+- ``indent="  "``
+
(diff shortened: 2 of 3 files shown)


sivel:
I'm not really sure how I feel about this.  My first feeling is -1

Here are some of my first reactions:
1. This looks hugely complicated
2. it duplicates some existing functionality such as `to_json` and `to_yaml`
3. It doesn't use any native encoders, and relies instead on building encoders itself
4. I feel like some of these things are too specific
5. it makes things too magical
6. goes against my personal feelings of using templates as best practice
7. We would be forcing opinion on how we believe these things should be built

bcoca:
also -1, I agree with each of @sivel's points

jtyr:
Thank you for your feedback. Please let me comment on your points:

> This looks hugely complicated

It might look complicated but it's actually very simple to use once such universal role is created. I have created [tens of roles](https://github.com/jtyr?tab=repositories) using these filters during the last 1+ year and find it really great for creating reusable roles. The complexity required to achieve the same in the template file would result in even more complicated thing.

> It duplicates some existing functionality such as to_json and to_yaml

Yes, it duplicates some of the functionality of the `to_json` and `to_yaml` filters but with more advanced result. I could enhance the existing filters to serve the same purse but I thought it would be kind of nice to have all the filters named the same way.

> It doesn't use any native encoders, and relies instead on building encoders itself

That's correct. The problem with native encoders (if they exist) is that they don't easily allow to use features like convert a number as string to real number which is required by the final configuration format. For example some of the configuration formats require to have numbers unquoted to be represented as a real number and not as a string. The same applies for boolean values. On top of that, there are no filters build for certain configuration formats and even if there would, they would have to be installed as a dependency of the filter which is always more difficult than just use the Ansible filter itself.

> I feel like some of these things are too specific

Yes and no. Some formats are really specific (e.g. `encode_apache`, `encode_xml`) but some are quite universal. For example `encode_ini` can be used to configure sysconfig files or in general as any `key = value` configuration. Other example is the `encode_nginx` filter which can be used for [Bacula](https://github.com/jtyr/ansible-bacula_director) configuration. Each filter provides a set of parameters which can modify the behaviour of the filter so it's possible to reuse already existing filter instead of creating a new one.

> It makes things too magical

I don't think this is a bad thing after all. Ansible is beautiful tool which allows to achieve difficult thing in a simple manner. If this set of filters helps the users simplify their daily work with Ansible (it does for me and many users of my roles), then it should be provided as an alternative to the classical Jinja2 templates.

> Goes against my personal feelings of using templates as best practice

I understand that not everybody must like these filters but on the other hand, they are not forced to use them if they don't want to.

> We would be forcing opinion on how we believe these things should be built

By merging these filters, nobody would be forced to use them if they don't want to. They would be only available to those who are interested to create truly universal roles and who would like to contribute to the reduction of Ansible roles fragmentation (see the [motivation](https://github.com/jtyr/ansible/blob/jtyr-config_encoders/docsite/rst/playbooks_filters_config_encoders.rst#motivation)). Go and search for `nginx` in the [Ansible Galaxy](https://galaxy.ansible.com/list#/roles?page=1&page_size=10&autocomplete=nginx). You will find hundreds of roles doing the same thing - install nginx package, configure the web server, make sure the service is enabled and running. The only difference between the majority of the roles is the configuration. This is, in the first place, why I have created these filters - to be able to provide single role with universal configuration satisfying any scenario. Without these filters you can never create a [Nginx role](https://github.com/jtyr/ansible-nginx) which allows to create any possible configuration.

---

I have spent significant amount of time to bring the functionality, the code and the documentation to a good level to be useful to many users. I would appreciate if you would spend some time by exploring some of [my roles](https://github.com/jtyr?tab=repositories) and possibly test the filters by yourself before closing this PR as inappropriate.

sivel:
# Question About This One

Hi!

Unfortunately, we're not sure if we want this feature in the program, and I don't want this to seem confrontational.  Our reasons for this are listed above in a prior comment.

bcoca:
So the 'data format' filters make some sense (renaming encode_, to existing to_ convention), to_xml (_cringe_) and to_ini , but i draw the line in templating application specific configs, this is what the template modules is for and it is trivially easy to create roles that do this specifically.

this is not what filters are for, they manipulate data, not config files.

jtyr:
Thank you for taking a closer look, @bcoca, I really appreciate it. I have nothing against renaming the filters to `to_*` and name the set of filters as "Data formatting filters".

I just don't see your point of drawing the line between [INI, JSON, TOML, XML, YAML] and [Apache, Erlang, Haproxy, Logstash, Nginx, PAM] file formats. They are all configuration file formats although some of them are suitable only for one application. Having a filter which helps to generate the configuration file in a very simple way would definitely be appreciated by many users.

You are right that the filters can be implemented as a Jinja2 templates (and [I have done it](https://github.com/jtyr/config-encoder-macros)!) but it's not that convenient like filters. If people implement the same as standard Jinja2 template, they end up copying the algorithm from template to template which is a big nightmare for maintenance after a while. And if you were thinking of using the application-specific template as an Ansible role, then that's even less convenient because you either have to use it as a dependency of your new role (executed before the new role - config file can be overwritten by the package installation) OR you have to use it after the new role in the playbook (executed after the application is installed - the service is probably already running, so the formatting role would have to restart the service if it would change the config file).

As you can see, I went through several solutions and have found out that only the filters are really useful solution. Please explore some of [my roles](https://github.com/jtyr?tab=repositories) to see how useful these filters are. I deeply believe that these filters could be a great benefit to advanced users and I'm willing to do whatever modification you propose to bring it to a mergeable state.

bcoca:
The distinction is simple, data vs config formats, filters are for data transformation.  

Using a template in the role has the same effect and does not duplicate code. So you can rework your roles to do that instead of using the filters.

If you really wanted, you could do the same thing in lookup/callback/connection or any plugin, but those are not the proper places to do this work, the same goes for filters.

jtyr:
> The distinction is simple, data vs config formats,

I see your point now, @bcoca. By your measures, I think, only the Erlang, JSON, YAML and XML can be considered as the real data formats. TOML and INI are pure configuration formats, with the difference that they are not designed only for one application, and still you think they would _make sense_ as the `to_*` filters.

> filters are for data transformation.

I absolutely agree with you. And this is exactly what all these filters are doing. The filters read a **data structure** (set of lists/dicts) and transform it into a text. The text is then placed into the output of the standard Ansible template. The filters do no remove the best practice of using templates to deliver configuration. It just allows the users to use another level of abstraction which has certain advantages:
- Unification of the produced output (the same text formatting)
- Simple configuration file creation (it's simpler to write a YAML data structure than a XML file)
- Unlimited possibilities of extending the role configuration without the need to modify the role (only override a variable to do a change)

> Using a template in the role has the same effect and does not duplicate code. So you can rework your roles to do that instead of using the filters.

You are right. I could use the application-specific filters (Apache, Haproxy, Logstash, Nginx, PAM) directly in the application roles by placing the filter into the `filter_plugins` directory. Unfortunately there are two problem with this concept:
1. The application-specific filters can be used also for other applications. For example the Nginx filter can be used for [Bacula configuration](https://github.com/jtyr/ansible-bacula_director/blob/master/templates/bacula-dir.conf.j2) which would result into a code redundancy in multiple roles. This is why it would be really great to have these filters distributed as a standard part of Ansible to prevent this kind of code redundancy.
2. The application-specific filters (Apache, Haproxy, Nginx, PAM) can be used in other roles to create additional configuration files in the `*.d` directories (e.g. Apache - `/etc/httpd/conf.d/`, Nginx - `/etc/nginx/conf.d/`, PAM - `/etc/pam.d/`). I'm sure that a lot of users would welcome an unified way how to create these config files in a simple manner.

> If you really wanted, you could do the same thing in lookup/callback/connection or any plugin, but those are not the proper places to do this work, the same goes for filters.

Other types of plugins are not very useful for data manipulation. I believe that the only plugin type suitable for this job is filter.

bcoca:
No, I meant specifically, application specific templates, not filters. I agree ini and TOML are on the boundary and I'm not sure if we even want those, but I am at least willing to discuss them.

jtyr:
My feeling from your reactions is that you don't feel comfortable with the fundamental idea of these filters - to be able to create configuration file based on a data structure. Is that correct? Or what else is behind your hesitation, @bcoca?

The filters are nothing else than a representation of patters which I observed during the creation of tens of Ansible roles. Most of the roles are publicly available on my [GitHub site](https://github.com/jtyr?tab=repositories) and are used by many users. I have also presented it several times in front of many professionals, including the audience of the [Ansible London meetup](http://www.slideshare.net/jtyr/automation-and-ansible), and got a positive feedback. The only negative feedback was that it's not a standard part of Ansible and therefore it won't be widely used.

I was hoping that these filters will help the users to create more universal and reusable roles. But that won't happen without the support of the Ansible project.

bcoca:
i think the objectives are orthogonal, reusable roles do NOT require these filters, you can make them the same using tempalte.

jtyr:
You are absolutely right, @bcoca. You can do exactly the same directly in the template. But doing it so, you have to invest much more work or you have to copy the code from an other role. Please let me illustrate that on the following example.

Please imagine this simple configuration file:

```
host = localhost
ip = 192.168.1.13

[fw]
open = 22
```

It's a simple INI file which can be described with this data structure:

```
my_config:
  host: localhost
  ip: 192.68.1.13
  fw:
    open: 22
```

In order to encode this data structure back into the INI format, you would have to write a template like this:

```
{#- First process properties without a section #}
{%- for property, value in my_config.iteritems() | sort %}
  {%- if value is not mapping -%}
    {{ property ~ " = " ~ value ~ "\n" }}
  {%- endif %}
{%- endfor %}

{#- Then process sections #}
{%- for section, options in my_config.iteritems() | sort %}
  {%- if options is mapping -%}
    {{ "\n[" ~ section ~ "]\n" }}
    {%- for property, value in options.iteritems() | sort -%}
      {{ property ~ " =  " ~ value ~ "\n" }}
    {%- endfor %}
  {%- endif %}
{%- endfor %}
```

This implementation brings several issues:
1. If you have another role where you want to create an INI file from a variable, you would have to copy the template to the other role.
2. If you would like to add something into the algorithm while using the template across multiple roles, you would have to change multiple roles.
3. The variable `my_config` is hardcoded in the template. If you would like to use the same template in another role, you would have to change the variable every time. This issue can be eliminated by wrapping the code into a Jinja2 [macro](https://github.com/jtyr/config-encoder-macros/blob/master/macros/ini_encode_macro.j2). But even then the usage is [not that nice](https://github.com/jtyr/config-encoder-macros/blob/master/templates/test.ini.j2) as in the case of using filters.

For comparison, here is the same as above but with the INI Config Encoder:

```
{{ my_config | encode_ini }}
```

Isn't that simpler and more convenient than using templates?

bcoca:
both encode_ini and the template can be reused, you are comparing a function to it's implementation, that is a flawed comparison. Comparing the template to the code of the encode_ini filter would be a apples to apples.

`- template: src=my_ini.j2 dest=/etc/myconfig.ini`

jtyr:
I was just trying to compare the amount of work the user must put in to receive the same result using each of the approaches. Both approaches are the same from the Playbook perspective as both are using the `template` module. The only difference is in the content of the template file.

@bcoca, please have a look at the above example in [this GitHub repo](https://github.com/jtyr/ansible-test_filters).

bcoca:
The same role can be made reusable to the user using a filter or a template, i favor the latter for config formats, also it is easier to extend as templating languages are normally a much lower barrier than programming ones.

jtyr:
You are right, @bcoca, both can be used. But if you use filters, you don't have to copy (and possibly modify) the template code between roles.

Please have a look at an example [here](https://github.com/jtyr/ansible-test_filters/tree/sharing). In order to share the algorithm between two roles, I had to copy the template file from the `role_template1` into the `role_template2` and change the name of the variable `my_config1` to `my_config2` in two places in the template code. If the template would be more complex, I would probably had to change it even on more places. And imagine if I would find a bug or would like to do an enhancement in the template. I would have to change it in all the roles where the template is used. And what if somebody else started to use the template? How would I distribute the fix/enhancement?

In the contrast, having the logic implemented as a filter, you change the code on one place and everybody who is using it will get the fix/enhancement automatically. On top of that, it allows the user to use the functionality straight away without fiddling with the code every time (s)he wants to use it elsewhere.

IMHO, at the moment when you have to change the code, you cannot talk about real reusability.

bcoca:
encapsulating the template in a role achieves the exact same result.

jtyr:
@bcoca, I have already commented on this approach in one of the comments [above](https://github.com/ansible/ansible/pull/15832#issuecomment-218906134):

> And if you were thinking of using the application-specific template as an Ansible role, then that's even less convenient because you either have to use it as a dependency of your new role (executed before the new role - config file can be overwritten by the package installation) OR you have to use it after the new role in the playbook (executed after the application is installed - the service is probably already running, so the formatting role would have to restart the service if it would change the config file).

Using a role as a replacement of a simple filter is not a very good workaround.

jtyr:
Is there anything else I can clarify about these filters, @bcoca? I think I have clearly shown that the role of these filters is the same as the role of Ansible modules where most of the Ansible modules could be replaced by an Ansible role but for the user it's much more convenient to use the module because it can be used from within the role as a task. Absolutely the same applies to the "data formatting" filters which are possible to replace by roles but for the user it's much more convenient to use it as a filter because it can be used from within the role as a task using the `template` module.

These filters just allow to use another level of abstraction from within the template file which allows the user to create configuration files in a simple and an unified way. Together with the Jinja2 data manipulation practices, like merging lists and dicts, it's possible to create truly universal roles that can help to reduce the fragmentation of Ansible roles.

bcoca:
I disagree, you keep conflating issues to prove your point, most of these filters can be implemented and reused in template form, I really don't see it as a good idea to make configuration file specific filters.

jtyr:
The reusability of the template files is the issue the filters are trying to address. I already commented on why [here](https://github.com/ansible/ansible/pull/15832#issuecomment-219557156), @bcoca.

Please can you describe how would you prevent copying the code in the template file from one role to another one? Once you start copying the code, it becomes hard to maintain it as I described [here](https://github.com/ansible/ansible/pull/15832#issuecomment-219572526). Wrapping the filter functionality into a role is also not convenient as I described [here](https://github.com/ansible/ansible/pull/15832#issuecomment-219829376).

Please read the responses I'm linking to from this post and think about it. The only logical conclusion is that there is no more convenient solution for this problem than using filters.

bcoca:
you don't need to copy the template, just setup the role to reuse it instead of the lookup

jtyr:
How would you share a role that is pointing to a template file which is not part of the role? People would have to download the template file manually and place it somewhere the role can find it in order the role to work. That's not a very user-friendly way how to share roles.

Apart of that, implementing algorithms is much simpler and more powerful in Python than in Jinja2 language. Remember, I have [done both](https://github.com/jtyr/config-encoder-macros) so I know what I'm talking about.

jtyr:
I think I have found a better way how to share the functionality of the filters, @bcoca. I have created an Ansible [role](https://github.com/jtyr/ansible-config_encoder_filters) which contains the Config Encoder Filters in the `filter_plugins` directory. If this role is set as a dependency of another role, the filters are available to the role (and any other role used by that playbook).

I still believe that these filters would make more sense if they would be as a standard part of Ansible. Do you think that there is still a possibility that you would change your opinion if more users would request this functionality? For example, would you consider to merge this PR (with whatever modifications you would suggest) if let's say 1000 people would vote with a positive GitHub reaction (:+1: or :smile:) on my first comment?

PR #15836, opened 2016-05-12
docs: {{ net_mask | ipaddr('net') }}  example is wrong

BuildTheRobots:
##### ISSUE TYPE

- Docs Pull Request
##### ANSIBLE VERSION

```
ansible 2.0.2.0
```
##### SUMMARY

Example in documentation is wrong for ipaddr('net')

To get "ipaddress/cidr" output the documentation recommends `ipaddr('net').
On my systems this gives me blank output.
After trial and error using`ipaddr('host')` seems to give the correct transformation
##### SYSINFO

centos 6.x amd64, latest updates as of today's date.
python-netaddr.noarch  0.7.5-4.el6

### DIFF
--- a/docsite/rst/playbooks_filters_ipaddr.rst
+++ b/docsite/rst/playbooks_filters_ipaddr.rst
@@ -311,7 +311,7 @@ This result can be canonicalised with ``ipaddr()`` to produce a subnet in CIDR f
     # {{ net_mask | ipaddr('prefix') }}
     '24'
 
-    # {{ net_mask | ipaddr('net') }}
+    # {{ net_mask | ipaddr('host') }}
     '192.168.0.0/24'
 
 


BuildTheRobots:
should this not be tagged as "docs_pull_request"?

bcoca:
in my testing this is reversed, 'net' gets me the expected result and 'host' gives me a blank

```
bcoca@erebos:~ $ ansible -m debug -a 'msg={{"192.168.0.0/255.255.255.0"|ipaddr("net")}}' localhost
localhost | SUCCESS => {
    "msg": "192.168.0.0/24"
}
bcoca@erebos:~ $ ansible -m debug -a 'msg={{"192.168.0.0/255.255.255.0"|ipaddr("host")}}' localhost
localhost | SUCCESS => {
    "msg": ""
}
```

BuildTheRobots:
This behaviour seems to change depending on if you feed it a network address or a host IP address. I don't know if this is supposed to be different or not, my feeling is it shouldn't be, but I might well be misunderstanding what's going on (this is my second week with ansible).

`[root@bsc ansible]# ansible -m debug -a 'msg={{"192.168.0.0/255.255.255.0"|ipaddr("net")}}' localhost
localhost | SUCCESS => {
    "msg": "192.168.0.0/24"
}`

`[root@bsc ansible]# ansible -m debug -a 'msg={{"192.168.0.0/255.255.255.0"|ipaddr("host")}}' localhost
localhost | SUCCESS => {
    "msg": ""
}`

`[root@bsc ansible]# ansible -m debug -a 'msg={{"192.168.0.1/255.255.255.0"|ipaddr("net")}}' localhost
localhost | SUCCESS => {
    "msg": ""
}`

`[root@bsc ansible]# ansible -m debug -a 'msg={{"192.168.0.1/255.255.255.0"|ipaddr("host")}}' localhost
localhost | SUCCESS => {
    "msg": "192.168.0.1/24"
}`

edit: can't work out how to format code block whist keeping newlines intact.

bcoca:
i can see it both ways, as the input can be considered valid/invalid depending on exact context.

BuildTheRobots:
Yep, I agree. All about context.
I would suggest cancelling my pull request as it's obviously wrong but possibly updating the documentation to explain that `net` will only return if it's a valid network address (192.168.0.128/25 returns correctly) and that `host` should be used if it's an actual (not network) IP. I can see the value of being able to _validate_ the difference between network and host IP, though it's slightly infuriating there isn't a generic "ip-without-mask" filter :)

PR #15840, opened 2016-05-12
community.rst - Formatting and readability

gundalow:
##### ISSUE TYPE
- Docs Pull Request
##### ANSIBLE VERSION

N/A
##### SUMMARY

By using the correct formatting we can avoid the double dash/em dash symbol.

I also took the opportunity to make this a bit more readable. Adding :code:`somecode` makes it a lot easier when someone scans this page to spot things, such as `make tests` is a thing that they should do.

Add in some extra links to the source so people can find more information

Note to reviews: In github select "Display the rich diff" button, which is next to "view" at the beginning of the diff, at the same section as the file name

---

This change is [Reviewable](https://reviewable.io/reviews/ansible/ansible/15840)

### DIFF
--- a/docsite/rst/community.rst
+++ b/docsite/rst/community.rst
@@ -27,7 +27,7 @@ Where needed, link to gists or github repos to show examples, rather than sendin
 
 We recommend using Google search to see if a topic has been answered recently, but comments found in older threads may no longer apply, depending on the topic.
 
-Before you post, be sure you are running the latest stable version of Ansible.  You can check this by comparing the output of :code:`ansible --version` with the version indicated on `PyPi <https://pypi.python.org/pypi/ansible>`_.
+Before you post, be sure you are running the latest stable version of Ansible.  You can check this by comparing the output of ``ansible --version`` with the version indicated on `PyPi <https://pypi.python.org/pypi/ansible>`_.
 
 Alternatively, you can also join our IRC channel - #ansible on irc.freenode.net.  It's a very high traffic channel as well, if you don't get an answer you like, please stop by our mailing list, which is more likely
 to get attention of core developers since it's asynchronous.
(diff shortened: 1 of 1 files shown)


sivel:
I cancelled the travis build here, since this is just docs changes.  Ignore the "failure"

docschick:
I haven't seen this use of Sphinx/RST markup before, where you use ":code:`ansible --version`" instead of backticks, like `ansible --version`.  (And perhaps I just have a lot to learn still, which may very well be the case.)  Wouldn't the backticks be easier and more like what is used in other places in the docs?

gundalow:
Hi @docschick Thanks for looking at this.

This URL shows the change a bit clearer https://github.com/ansible/ansible/pull/15840/files?short_path=39d1134#diff-39d113434fb0cf469014d915725c2c60

If you search for "version" on http://docs.ansible.com/ansible/community.html#i-ve-got-a-question you can see the type of formatting issue we have at the moment

docschick:
I do see the problem in the HTML output, and thank goodness you saw it too, but I still think the backquotes/backticks are the easiest way to solve this and make them stand out (and seem more commonly used, to me, in Sphinx/RST as inline markup).  But the backquotes need to be double backquotes for the markup to render the way you need it to. (http://www.sphinx-doc.org/en/stable/rest.html#inline-markup)

As an example of the output I think you'd prefer to see, maybe this works: http://docs.ansible.com/ansible-tower/latest/html/administration/tower-manage.htmll  --> like how the `tower-manage inventory_import [--help]` command is shown. I did this using the double backquotes for markup.  And, as you can see, the double dash is preserved here as it should be.  

Sorry for the Tower docs example, but it was the easiest link for me to find quickly for comparison of the markup. I'm also used to seeing the backquotes treatment in the Ansible docs:  
http://docs.ansible.com/ansible/developing_modules.html  -->  like is used for markup around the "`--module-path` command line option" text in the 3rd sentence. 

I guess it boils down to consistency for me, in how markup is already used in the docs and the concern for introducing something a bit new that doesn't match up in the display/rendering with other places in the docs.

gundalow:
@docschick You are correct, let me change this to double backticks.

I didn't see that else where, so I looked up the formatting and so the example I used

gundalow:
@docschick Done :)

https://github.com/ansible/ansible/pull/15840/files?short_path=39d1134#diff-39d113434fb0cf469014d915725c2c60

docschick:
Thanks so much, @gundalow! Looks good to me. I'll confirm it's cool to merge and then I'll happily push the button. :)

gundalow:
@docschick Thanks :)

PR #15847, opened 2016-05-13
Improve the reporting related to module errors

dagwieers:
##### ISSUE TYPE

- Feature Pull Request
##### ANSIBLE VERSION

All versions
##### SUMMARY

When you get reports that a specific module does not allow check-mode, or was provided an unsupported parameter, it helps to know what module we are talking about in the same message.


bcoca:
it should be apparent already from the task run, also this needs to be in sync with the same message in action plugins.

dagwieers:
The task run does not show the module name. Indirectly, you have to look for the role/playbook to find out the module name based on the task name.

dagwieers:
Superseded by #16087

PR #15869, opened 2016-05-16
Add doc on multiple exceptions and python 3

mscherer:
##### ISSUE TYPE
- Docs Pull Request


mscherer:
Ok that was wrong, not sure what I had in mind

PR #15875, opened 2016-05-16
Port rax.py to python3 compatible syntax

mscherer:
##### ISSUE TYPE
- Feature Pull Request
##### SUMMARY

Since the pyrax website say that only python 2.7 is tested,
I do not think it is worth to aim for python 2.4 compatibility
for the various rackspace modules.


abadger:
@sivel what do you think?  do you want the rax module_utils to maintain python-2.4 compatiblity or to update?

sivel:
LGTM

abadger:
Thanks guys!  Merged.

PR #15896, opened 2016-05-17
Fix OS version check for RHEL 4 to prevent execution of ipv6 ip route…

gbevan:
… command that causes kernel panics on RedHat/CentOS 4.8 (kernel 2.6.9-89)
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.0.2.0
and
ansible 1.9.4
```
##### COMPONENT NAME
##### SUMMARY

RedHat OS version check for version 4 was checking that distribution_version startswith '4.' <- note the '.'.  This works fine with CentOS 4.8, but RHEL 4.8 returns just '4' - so the check fails to prevent the ipv6 ip route command from being run, which triggers a kernel panic.

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -2565,8 +2565,8 @@ class LinuxNetwork(Network):
         )
         interface = dict(v4 = {}, v6 = {})
         for v in 'v4', 'v6':
-            if (v == 'v6' and self.facts['os_family'] == 'RedHat' and
-                    self.facts['distribution_version'].startswith('4.')):
+            if v == 'v6' and self.facts['os_family'] == 'RedHat' \
+                and self.facts['distribution_version'].startswith('4'):
                 continue
             if v == 'v6' and not socket.has_ipv6:
                 continue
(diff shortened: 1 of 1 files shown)


mscherer:
So what about distribution that could be in the version 40 ? (ie, Fedora in some year, or opensuse 42.1 right now).

gbevan:
Agreed.  I guess the underlying problem (aside from an old kernel that panics with ipvs6 packets) is that python's platform.dist(), which ansible is using to get the distribution_version from doesn't report in the standard that all the other RedHat family OSes do.  e.g.:

python -c "import platform; print platform.dist()"
('redhat', '4', '')

Looks like it doesnt parse the /etc/redhat-release file correctly (or that /etc/redhat-release is formatted differently).

We could lock it down further to a kernel release, e.g.:

python -c "import platform; print platform.release()"
2.6.9-89.EL
?

mscherer:
Also, I hate to be the bearer of bad news, but I think we kinda consider RHEL 4 to not be supported anymore.

gbevan:
Well, I'm still encountering them out there in the real world (scary as that sounds)... Plus the original code in ansible is intended to prevent the ipv6 ip route command from running on any RedHat family with a distribution_version of 4 - presumably to prevent these kernel panics...

gbevan:
What if we kept the "4." test but or'd with (startswith "4" AND kernel startswith "2.6.9-") ?  Would that be acceptable?

mattclay:
@gbevan Please rebase your branch and update this PR so tests can be re-run.

mattclay:
The branch this PR is based on is too old to work with CI. Once it is rebased, tests will re-run automatically.

calfonso:
RHEL 4 was never supported by Ansible so we are going to close this PR.

Thank you!

PR #15913, opened 2016-05-18
Make load_params into a function that custom modules can access

abadger:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

2.1 and devel
##### SUMMARY

mtnbikenc was asking how his dynamic module should introspect the module args to create an argspec prior parsing the arguments for real.  With ansiballz reading from stdin there wasn't a way to do this.  Moving _load_params() to the top level gives users a way to read this that is hopefully more stable than them parsing it themselves.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -222,10 +222,10 @@ _literal_eval = literal_eval
 # Backwards compat.  There were present in basic.py before
 from ansible.module_utils.pycompat import get_exception
 
-# Internal global holding passed in params and constants.  This is consulted
-# in case multiple AnsibleModules are created.  Otherwise each AnsibleModule
-# would attempt to read from stdin.  Other code should not use this directly
-# as it is an internal implementation detail
+# Internal global holding passed in params.  This is consulted in case
+# multiple AnsibleModules are created.  Otherwise each AnsibleModule would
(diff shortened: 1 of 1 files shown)


lib/ansible/module_utils/basic.py
```diff
@@ -508,9 +508,59 @@ def is_executable(path):
...
+                buffer = sys.stdin.read()
+            else:
+                buffer = sys.stdin.buffer.read()
+        _ANSIBLE_ARGS = buffer
+
+    try:
+        params = json.loads(buffer.decode('utf-8'))
+    except ValueError:
+        # This helper used too early for fail_json to work.
+        print('\n{"msg": "Error: Module unable to decode valid JSON on stdin.  Unable to figure out what parameters were passed", "failed": true}')
```
bcoca, line 551: the error should go to stderr, since sys.exit does this by default (and still returns rc=1) just use that.

abadger:
@mtnbikenc This is for you.  Hope to get it merged today.

mtnbikenc:
@abadger Cool.  I'll look into testing this.

lib/ansible/module_utils/basic.py
abadger, line 551: We talked about this in slack and realized that nothing else is returning json on stderr.  We'll look at it separately and if we decide it's appropriate we'll implement that in 2.2.

PR #15918, opened 2016-05-18
Forked gce utility module to support Google Cloud DNS

walbert947:
##### ISSUE TYPE
- New Module Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (google_dns d2b1e9a6d8) last updated 2016/05/18 11:20:27 (GMT -700)
  lib/ansible/modules/core: (detached HEAD 92bf802cb8) last updated 2016/05/18 11:19:41 (GMT -700)
  lib/ansible/modules/extras: (detached HEAD e710dc47fe) last updated 2016/05/18 11:19:51 (GMT -700)
  config file = /home/william/.ansible.cfg
```
##### SUMMARY

This is a quick and dirty fork of the gce.py utility module to enable support for libcloud DNS, which I need for an upcoming Google Cloud DNS module PR.

The original gce.py module is hard-coded to use libcloud's compute provider, whereas the Google Cloud DNS module requires libcloud's dns provider (which has a different name).

I also changed the function name to avoid colliding with the existing gce.py module.

I'll link to the Google Cloud DNS PR as soon as it's submitted.

### DIFF
--- a/lib/ansible/module_utils/gcdns.py
+++ b/lib/ansible/module_utils/gcdns.py
@@ -1,0 +1,126 @@
+# This code is part of Ansible, but is an independent component.
+# This particular file snippet, and this file snippet only, is BSD licensed.
+# Modules you write using this snippet, which is embedded dynamically by Ansible
+# still belong to the author of the module, and may assign their own license
+# to the complete work.
+#
+# Copyright (c), Franck Cuny <franck.cuny@gmail.com>, 2014
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification,
+# are permitted provided that the following conditions are met:
+#
+#    * Redistributions of source code must retain the above copyright
+#      notice, this list of conditions and the following disclaimer.
+#    * Redistributions in binary form must reproduce the above copyright notice,
+#      this list of conditions and the following disclaimer in the documentation
+#      and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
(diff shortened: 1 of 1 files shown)


walbert947:
Google Cloud DNS PR is at https://github.com/ansible/ansible-modules-extras/pull/2252.

sivel:
Generally speaking, we try to keep all related code in a single file in `module_utils`.  So all gce code should just go in `gce.py`

walbert947:
I definitely consider this more of a proof of concept than a permanent solution.

> all gce code should just go in gce.py

GCE is one of several projects in Google Cloud. My modules cover Google Cloud DNS, which is another project and is handled internally in libcloud by a set of providers in the `libcloud.dns` package. 

With respect to `gce.py`, the following imports caused my Google Cloud DNS modules to fail:

``` python
try:
    from libcloud.compute.types import Provider
    import libcloud
    from libcloud.compute.providers import get_driver
    HAS_LIBCLOUD_BASE = True
except ImportError:
    HAS_LIBCLOUD_BASE = False
```

These are in pulling from the `libcloud.compute` package, so they're _specifically_ for GCE, and would fail when used with any other product in Google Cloud Platform.

A cleaner way of fixing this would be to remove the hard-coded libcloud imports from `gce.py` and make providing the right set of types and drivers the responsibility of whatever client modules are calling it.

So instead of a function like this:

``` python
def gce_connect(module, provider=None):
```

You'd have something like this:

``` python
def gce_connect(module, provider, get_driver):
```

This can be taken a step further by renaming the main connection module to `gcp.py` (to represent all of Google Cloud Platform), and then making wrapper modules for `gce.py`, `gcdns.py`, etc. which subsequently call the main `gcp.py` with the appropriate libcloud components for their respective projects. This should minimize code duplication, while still maintaining compatibility with existing GCE modules.

However, all that would require a substantial refactoring work. Since Ansible (at this point) only has modules related specifically to GCE, it would be difficult for me to make a case as to why refactoring is needed. That's why I opted for a quick and minimal fork, so that I could demonstrate code that uses Google Cloud but doesn't work with Ansible's `gce.py` module as-is.

walbert947:
@sivel I've created another PR with a more permanent (and much less hacky) solution to my problem, so I'm going to close this one. I'd appreciate it if you could review the new PR: https://github.com/ansible/ansible/pull/15924

PR #15704, opened 2016-05-03
Add Ubuntu xenial 16.04 image to travis

robinro:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

2.2, devel
##### SUMMARY

Adds ubuntu xenial 16.04 to the list of images run by travis.
Also fix broken tests.

Before merging the image should appear in dockerhub ansible/ansible and the IMAGE=robinro/\* hack be removed.

PR #15650, opened 2016-04-28
opensuse image for travis tests

robinro:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

2.1.0, devel
##### SUMMARY

Run the test suite also on openSUSE leap 42.1

Before merging, someone from ansible/core needs to:
- build the docker image over at ansible/ansible
- correct the docker image used in run_tests (i.e. remove the `if` line (L14) that points to my dockerhub repo)

Future ToDos:
- mysql, postgresql: disabled for now, need some fixes to use the suse-specific start/... scripts, I would leave this for some future time or someone who can do this more easily than me

PR #15936, opened 2016-05-20
Run tests on ubuntu1604 and opensuseleap

robinro:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

2.2
##### SUMMARY

This change enables the tests to run on ubuntu1604 and opensuseleap.
The tests are updated to pass on those distributions.

Also there in test_uri there is an assumption failing with 1604, which I skip. Since that test includes 12.04 workarounds already it should probably be a separate task to clean it up and make it work with all versions.

This PR replaces https://github.com/ansible/ansible/pull/15650 and https://github.com/ansible/ansible/pull/15704 (1604 and leap separately).

### DIFF
--- a/test/integration/roles/setup_postgresql_db/tasks/main.yml
+++ b/test/integration/roles/setup_postgresql_db/tasks/main.yml
@@ -8,24 +8,10 @@
       paths: '../vars'
 
 # Make sure we start fresh
-- name: remove rpm dependencies for postgresql test
-  package: name={{ postgresql_package_item }} state=absent
-  with_items: "{{postgresql_packages}}"
-  loop_control:
-    loop_var: postgresql_package_item
-  when: ansible_os_family == "RedHat"
-
-- name: remove dpkg dependencies for postgresql test
-  apt: name={{ postgresql_package_item }} state=absent
-  with_items: "{{postgresql_packages}}"
-  loop_control:
-    loop_var: postgresql_package_item
-  when: ansible_pkg_mgr  ==  'apt'
-
 - name: remove old db (red hat)
   command: rm -rf "{{ pg_dir }}"
   ignore_errors: True
-  when: ansible_os_family == "RedHat"
+  when: ansible_os_family == "RedHat" or ansible_os_family == "Suse"
 
 # Theoretically, pg_dropcluster should work but it doesn't so rm files
 - name: remove old db config (debian)
@@ -43,7 +29,7 @@
   with_items: "{{postgresql_packages}}"
   loop_control:
     loop_var: postgresql_package_item
-  when: ansible_os_family == "RedHat"
+  when: ansible_os_family == "RedHat" or ansible_os_family == "Suse"
 
 - name: install dpkg dependencies for postgresql test
   apt: name={{ postgresql_package_item }} state=latest
@@ -66,6 +52,10 @@
   ignore_errors: True
   when: ansible_os_family == 'Debian'
 
+- name: Initialize postgres (Suse)
+  service: name=postgresql state=restarted
+  when: ansible_os_family == 'Suse'
+
 - name: Copy pg_hba into place
   copy: src=pg_hba.conf dest="{{ pg_hba_location }}" owner="postgres" group="root" mode="0644"
 
@@ -77,13 +67,17 @@
   command: locale-gen es_MX
   when: ansible_os_family == 'Debian'
 
+- name: install i18ndata
+  zypper: name=glibc-i18ndata state=present
+  when: ansible_os_family == 'Suse'
+
 - name: Generate pt_BR locale (Red Hat)
   command: localedef -f ISO-8859-1 -i pt_BR pt_BR
-  when: ansible_os_family == 'RedHat'
+  when: ansible_os_family == "RedHat" or ansible_os_family == "Suse"
 
 - name: Generate es_MX locale (Red Hat)
   command: localedef -f ISO-8859-1 -i es_MX es_MX
-  when: ansible_os_family == 'RedHat'
+  when: ansible_os_family == "RedHat" or ansible_os_family == "Suse"
 
 - name: restart postgresql service
   service: name={{ postgresql_service }} state=restarted
--- a/test/integration/roles/setup_postgresql_db/tasks/main.yml
+++ b/test/integration/roles/setup_postgresql_db/tasks/main.yml
@@ -8,7 +8,7 @@
       paths: '../vars'
 
 # Make sure we start fresh
-- name: remove old db (red hat)
+- name: remove old db (RedHat or Suse)
   command: rm -rf "{{ pg_dir }}"
   ignore_errors: True
   when: ansible_os_family == "RedHat" or ansible_os_family == "Suse"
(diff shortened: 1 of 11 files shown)


robinro:
I'm aware of the discussion about travis performance, but hope that once travis is quicker this can be merged.
There is no way around ubuntu1604 soon and ansible is also quite popular in the Suse community and opensuseleap is failry close to SLES12.

robinro:
I fixed the postgres/mysql issues now and run them on both new distros.

There is now only one test skipped: `test_uri` on ubuntu1604. The output from a httptester request has a lot more content then the comparison string. Maybe someone (@sivel @gundalow) can take a look at that?

This still depends on some additional packages (curl and locales) installed, see https://github.com/ansible/ansible/pull/15979

With this one assertion skipped and the packages added to the docker images, both distros pass the full integration test suite on my machines.

test/integration/roles/setup_postgresql_db/tasks/main.yml
```diff
@@ -8,24 +8,10 @@
...
-  when: ansible_os_family == "RedHat"
-
-- name: remove dpkg dependencies for postgresql test
-  apt: name={{ postgresql_package_item }} state=absent
-  with_items: "{{postgresql_packages}}"
-  loop_control:
-    loop_var: postgresql_package_item
-  when: ansible_pkg_mgr  ==  'apt'
-
 - name: remove old db (red hat)
```
gundalow: Redhat OR SLES

robinro: fixed

btw: removing the lines 11-24 in this change saves quite a bit on the test runtime

gundalow: I've previously spoken with abadger regarding this. Without this runs of tests can be effected by the existing state of the system under tests.  

Although this isn't an issue for Travis runs (as it's a fresh docker everytime) it does effect people testing in there own systems outside of docker.

robinro: I know the motivation, but those tasks don't help there. Removing/Reinstalling the package does not really provide a clean start since at least on Ubuntu/Suse this won't remove the databases. Removing the databases/config/... is far less trivial and depends a lot on the distribution.

sivel:
@robinro it looks like something is somehow different with the ubuntu 1604 image.

The tests are failing due to not using httptester, they are hitting the public testing servers, instead of the local docker container.  All prior containers are working fine, it seems it is just the ubu16 container.

If you look at `prepare_http_tests`, we swap out public hosts for the httptester hosts if a special docker env var is set, and this seems to work for all of the other images:

https://github.com/ansible/ansible/blob/devel/test/integration/roles/prepare_http_tests/tasks/main.yml#L1-L4

On the ubu16 container we see:

```
TASK [prepare_http_tests : set_fact] *******************************************
ok: [testhost] => {"ansible_facts": {"has_httptester": false}, "changed": false}
```

This indicates it doesn't have access to the httptester container.  I'm unsure why this env var wouldn't be available as it is with all other images.

robinro:
I see, it seems that /bin/sh on 16.04 doesn't get the envs; if I exec `/bin/bash -c env` it has the link envs, while `/bin/sh -c env` does not. No idea what's going on.

Shouldn't the tests also work without the httptester container?

sivel:
Some of the sites we use for testing specifically block certain providers.  The SNI test site seems to block the provider that runs our travis tests, which was a primary reason for introducing the httptester.

However there is a need to not enforce that everyone run containerized tests, or need to run httptester.  So the default is to use the public testing sites.

We could potentially set our own, `--env` for docker to indicate we want httptester, but its nice that you don't need to think about it, and use the functionality provided by docker to indicate whether we have the capability.

robinro:
I still don't know why /bin/sh doesn't get the envs.
If I make /bin/sh point to /bin/bash instead of /bin/dash, the tests pass. This seems to be some problem between 16.04, dash and docker.

robinro:
found it; envs with a dot don't show up in dash

possible solutions: 
- use bash: somewhat misses the point of an integration test if ubuntu uses dash by default
- rename the linked containers to something without dots: such that the envs don't have dots, but this needs new certificates for the sni checking

Also according to http://unix.stackexchange.com/questions/93532/exporting-a-variable-with-dot-in-it bashs manpage forbirds dots in envs.

I couldn't find a reference for the allowed docker naming scheme, but the deployments I've seen so far don't use dots in container linking.

sivel:
I think we should just add another explicit env var.

Something like:

``` diff
diff --git a/test/integration/roles/prepare_http_tests/tasks/main.yml b/test/integration/roles/prepare_http_tests/tasks/main.yml
index c3678aa..5cc1463 100644
--- a/test/integration/roles/prepare_http_tests/tasks/main.yml
+++ b/test/integration/roles/prepare_http_tests/tasks/main.yml
@@ -1,7 +1,7 @@
 # The docker --link functionality gives us an ENV var we can key off of to see if we have access to
 # the httptester container
 - set_fact:
-    has_httptester: "{{ lookup('env', 'ANSIBLE.HTTP.TESTS_PORT_80_TCP_ADDR') != '' }}"
+    has_httptester: "{{ lookup('env', 'ANSIBLE.HTTP.TESTS_PORT_80_TCP_ADDR') ~ lookup('env', 'HTTPTESTER') != '' }}"

 # If we are running with access to a httptester container, grab it's cacert and install it
 - block:
diff --git a/test/utils/run_tests.sh b/test/utils/run_tests.sh
index a63cf24..6a6f2cc 100755
--- a/test/utils/run_tests.sh
+++ b/test/utils/run_tests.sh
@@ -20,7 +20,7 @@ else
     fi
     export C_NAME="testAbull_$$_$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 8 | head -n 1)"
     docker pull ansible/ansible:${TARGET}
-    docker run -d --volume="${PWD}:/root/ansible:Z" $LINKS --name "${C_NAME}" ${TARGET_OPTIONS:=''} ansible/ansible:${TARGET} > /tmp/cid_${TARGET}
+    docker run -d --volume="${PWD}:/root/ansible:Z" $LINKS --name "${C_NAME}" --env HTTPTESTER=1 ${TARGET_OPTIONS:=''} ansible/ansible:${TARGET} > /tmp/cid_${TARGET}
     docker exec -ti $(cat /tmp/cid_${TARGET}) /bin/sh -c "export TEST_FLAGS='${TEST_FLAGS:-''}'; cd /root/ansible; . hacking/env-setup; (cd test/integration; LC_ALL=en_US.utf-8 make ${MAKE_TARGET:-})"
     docker kill $(cat /tmp/cid_${TARGET})
```

The above should allow us to key off of the explicitly set env var, or the one from docker linking.

robinro:
ok, I added the HTTPTESTER flag (and dropped the other test)

robinro:
It's green, so: ready_to_review

mattclay:
These changes look good. However, we should wait until https://github.com/ansible/ansible/pull/16000 is merged to enable these tests, so we can run them on Shippable instead of Travis.

mattclay:
@robinro: These changes have been working well with the tests on Shippable. Go ahead and remove the changes to .travis.yml and I'll merge this. We'll use Shippable to run the tests on the new images once #16000 is merged.

robinro:
@mattclay I removed the travis changes. Shippable is looking quite good already. Thanks for including the new images.

mattclay:
@robinro: Thanks. Squashed and merged. I updated the commit message to reflect not running the tests on Travis.

PR #15942, opened 2016-05-20
Sort groups by depth then by name

aperigault:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
2.2.0
```
##### SUMMARY

Actually group vars load order is random.
The goal is to predict vars precedence between two groups.

### DIFF
--- a/lib/ansible/inventory/host.py
+++ b/lib/ansible/inventory/host.py
@@ -138,6 +138,6 @@ class Host:
     def get_group_vars(self):
         results = {}
         groups = self.get_groups()
-        for group in sorted(groups, key=lambda g: g.depth):
+        for group in sorted(groups, key=lambda g: (g.depth, g.priority, g.name)):
             results = combine_vars(results, group.get_vars())
         return results
--- a/lib/ansible/vars/__init__.py
+++ b/lib/ansible/vars/__init__.py
@@ -252,7 +252,7 @@ class VariableManager:
             # we merge in vars from groups specified in the inventory (INI or script)
             all_vars = combine_vars(all_vars, host.get_group_vars())
 
-            for group in sorted(host.get_groups(), key=lambda g: g.depth):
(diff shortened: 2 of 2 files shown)


bcoca:
actually, i added a 'priority' property to groups that should be between depth and names

aperigault:
New property added !

aperigault:
I added a bug fix related to vars precedence.
Before ansible 2.0 inventory group vars are load after script group vars.
For people that use ansible tower to manage inventories, ansible 2.0 behavior is a big problem. We can't anymore overwrite vars define by group_by module for example.

h0nIg:
@bcoca why there is a need to sort groups by depth? it should be sorted by name and/or priority but the technical group structure should not matter

aperigault:
Depth is use lo load subgroup vars after group vars.
If you define a same var in group and in his subgroup. Subgroup var must win.

h0nIg:
i repeat: the technical structure of the groups should not matter. there should be an order by priority and then by name to make this understandable!

e.g. i create a lot of groups to structure my variables (and use variable merge instead of replace) and therefore the technical structure should not matter!

h0nIg:
by the way, this should be modified too: https://github.com/ansible/ansible/blob/e5462194261c7b55ccdf41adc4525dc86a1a34c1/lib/ansible/inventory/host.py#L136

aperigault:
You do not matter about depth because of merge use. With default vars behaviour use depth should be usefull.
If you define vars in subgroup the goal is to overwrite group vars no ?

I will modify host.py

aperigault:
Hello,

This PR is not supposed to be merge in 2.2.0 ?

h0nIg:
ping @bcoca

bcoca:
so i did not include as the name sorting would change existing behavior depending on filesystem sort.

I would make it an optional component depending on config. Since priority will be a noop unless specified I think that can stay as is.

aperigault:
Fixed here #17243

PR #15943, opened 2016-05-20
Add a few python3 fix, mostly found by trial and errors

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
- Docs Pull Request
##### SUMMARY

So far, with this, I am able to run command remotely on a
python3 only system.

```
 $ ansible  -u root -i 'fedora23,' all -a id -e 'ansible_python_interpreter=/usr/bin/python3'  
fedora23 | SUCCESS | rc=0 >>
uid=0(root) gid=0(root) groupes=0(root) contexte=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023
```

However, as that's mostly trial and errors, I am not sure if I am doing the right fixes.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -764,6 +764,8 @@ class AnsibleModule(object):
         '''
         if isinstance(path, unicode):
             path = path.encode("utf-8")
+        if isinstance(path, bytes):
+            path = path.decode("utf-8")
         return path
 
     # If selinux fails to find a default, return an array of None
@@ -1972,6 +1974,8 @@ class AnsibleModule(object):
         elif isinstance(args, basestring):
             if isinstance(args, unicode):
                 args = args.encode('utf-8')
+            if isinstance(args, bytes):
+                args = args.decode('utf-8')
             args = shlex.split(args)
         else:
             msg = "Argument 'args' to run_command must be list or string"
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -422,7 +422,7 @@ def remove_values(value, no_log_strings):
             value = unicode(bytes_value, 'utf-8', errors='replace')
         else:
             value = bytes_value
-    elif isinstance(value, Sequence):
+    elif isinstance(value, (Sequence, type({}.keys()))):
         return [remove_values(elem, no_log_strings) for elem in value]
     elif isinstance(value, Mapping):
         return dict((k, remove_values(v, no_log_strings)) for k, v in value.items())
@@ -2018,7 +2018,7 @@ class AnsibleModule(object):
         # passwords from the args list
         if isinstance(args, basestring):
             if isinstance(args, unicode):
-                b_args = args.encode('utf-8')
+                b_args = str(args.encode('utf-8'))
             else:
                 b_args = args
             to_clean_args = shlex.split(b_args)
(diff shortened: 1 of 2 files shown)


lib/ansible/module_utils/basic.py
```diff
@@ -1599,7 +1601,7 @@ def _log_invocation(self):
                 if not isinstance(param_val, basestring):
                     param_val = str(param_val)
                 elif isinstance(param_val, unicode):
-                    param_val = param_val.encode('utf-8')
+                    param_val = str(param_val.encode('utf-8'))
```
bcoca, line 1612: don't you want to_str?

lib/ansible/module_utils/basic.py
```diff
@@ -422,7 +422,7 @@ def remove_values(value, no_log_strings):
             value = unicode(bytes_value, 'utf-8', errors='replace')
         else:
             value = bytes_value
-    elif isinstance(value, Sequence):
+    elif isinstance(value, (Sequence, type({}.keys()))):
```
bcoca: is this just trying to match the iterable type returned by dict.keys()?

mscherer: Yes, I was pondering on finding a better way, but didn't look further.

mscherer:
So it fail for now on "test_utf8_output (units.module_utils.basic.test_run_command.TestAnsibleModuleRunCommand) ... ERROR"

lib/ansible/module_utils/basic.py
```diff
@@ -2095,11 +2097,15 @@ def run_command(self, args, check_rc=False, close_fds=True, executable=None, dat
                 rfd, wfd, efd = select.select(rpipes, [], rpipes, 1)
                 if cmd.stdout in rfd:
                     dat = os.read(cmd.stdout.fileno(), 9000)
+                    if isinstance(dat, bytes):
+                        dat = dat.decode('utf-8', 'replace')
```
mscherer, line 2111: So this part is likely responsible of the test failure.

lib/ansible/module_utils/basic.py
sivel: In py3.3 this is accessible from `collections.abc.KeysView`.  I would move this into the `try/except` logic farther up, where `Sequence` is built for older python.  Maybe create `SEQUENCE_TYPES` or such.

mscherer: Seems sensible, let's do that

mscherer:
So for now, I do push on the same branch so people can see the evolution, but I will squash the commits for merge (or split them in smaller chunk, if that's better)

lib/ansible/module_utils/basic.py
mscherer, line 1612: Do not seems to be present. But yeah, because this is likely doing something wrong (as I found out on others part of the code)

mscherer, line 1612: In fact, the problem is that on py3, this creates a bytes object, and so I need to convert it to string again. So i guess the right fix is to not encode in the first place for python3 or something like this.

abadger:
Just took a look at this to see what might have been fixed through other commits.  Looks like facts.py is now fixed.  I think all but these three functions are now fixed in basic.py:
- _check_arguments
- _log_invocation
- run_command

mscherer:
Yeah, that's a old PR, not sure if anything is needed anymore, let's close it. If there is problems, they will surface again sooner or later.

PR #15949, opened 2016-05-21
Fix the error handling for loop squashing to restore the name argument into the task args

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
stable-2.1 rc4:
a0ff49194e38f0ee944b42a53c1ac536cc287056
```
##### SUMMARY

If we bail out of loop squashing because of an error we need to make sure that we restore the name argument to its original value.  Otherwise modules will error (dnf, yum) or silently not perform their action (apt if update_cache is also given).

### DIFF
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -269,6 +269,7 @@ class TaskExecutor:
         Squash items down to a comma-separated list for certain modules which support it
         (typically package management modules).
         '''
+        name = None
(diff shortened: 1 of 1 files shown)


abadger:
Labelling this P1 as we need to make a decision as to whether this goes into 2.1.0 or gets deferred to 2.1.1.  Also need to decide whether to make an rc4 or not.

PR #15952, opened 2016-05-21
Add a alias for 'long' since python3 do not have it

mscherer:
##### ISSUE TYPE
- Feature Pull Request
##### SUMMARY

That's a commit I cherry picked from my fix_py3 branch. That's the easiest and most uncontroversial code so I hope it wouldn't be hard to review and merge.

Long is used for get_memory_facts on various platforms.


mscherer:
In fact, looking more closely, unarchive also use 'long', so maybe this should be moved to basics.py instead ?

PR #15955, opened 2016-05-21
Simplify the meta class in facts.py

mscherer:
##### ISSUE TYPE
- Feature Pull Request
##### SUMMARY

Since we are never using any arguments as part of **new** in the
subclass and in the only call to each class, we can get ride
of arguments and keywords dictionnary and make the code portable
to python 3

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -942,12 +942,12 @@ class Hardware(Facts):
     """
     platform = 'Generic'
 
-    def __new__(cls, *arguments, **keyword):
+    def __new__(cls):
         subclass = cls
         for sc in get_all_subclasses(Hardware):
             if sc.platform == platform.system():
                 subclass = sc
-        return super(cls, subclass).__new__(subclass, *arguments, **keyword)
+        return super(cls, subclass).__new__(subclass)
 
     def populate(self):
         return self.facts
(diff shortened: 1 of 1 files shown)


chris-l-weber:
At pycon2016 sprints, Looking at the object documentation (https://docs.python.org/3/reference/datamodel.html) **new** only uses the one parameter, Facts is based off object and defines no new **new**, So there is no reason to pass the additional arguments when inheriting from Facts. These **new** methods do not use _arguments and *_keyword so there is no reason to accept them. 
So all good.

chris-l-weber:
`fatal: [testhost]: FAILED! => {"changed": false, "failed": true, "module_stderr": "Traceback (most recent call last):\n  File \"/tmp/ansible_kIUItK/ansible_module_setup.py\", line 127, in <module>\n    main()\n  File \"/tmp/ansible_kIUItK/ansible_module_setup.py\", line 119, in main\n    data = get_all_facts(module)\n  File \"/tmp/ansible_kIUItK/ansible_modlib.zip/ansible/module_utils/facts.py\", line 3242, in get_all_facts\n  File \"/tmp/ansible_kIUItK/ansible_modlib.zip/ansible/module_utils/facts.py\", line 3188, in ansible_facts\nTypeError: __new__() takes exactly 1 non-keyword argument (2 given)\n", "module_stdout": "", "msg": "MODULE FAILURE", "parsed": false}`

The travis-ci error looks related to the code change, when it is initializing this the Facts class is running new with more then one argument. 

`````` python
facts.update(FACT_SUBSETS[subset](module,
                                         load_on_init=False,
                                         cached_facts=facts).populate()) ```
``````

abadger:
Note: This change is contained in https://github.com/ansible/ansible/pull/15955/files (with conditional so that it is only applied on python3).  If that gets merged we can close this.  Alternatively if this updates to have the conditional and we merge this first, we can take that out of #15955

abadger:
Copy and pasted wrong issue: https://github.com/ansible/ansible/pull/17036

abadger:
#17036 merged to devel.  closing this one.

PR #15960, opened 2016-05-22
get rid of temporary hack for v2_playbook_on_start

### LINKED ISSUE #15959: Hacky introspection in task_queue_manager.py results in hard to diagnose breakage
##### ISSUE TYPE
Bug Report

##### COMPONENT NAME
core

##### ANSIBLE VERSION
```
ansible 2.0.1.0
config file =
```

##### CONFIGURATION
N/A

##### OS / ENVIRONMENT
N/A

##### SUMMARY
In `task_queue_manager.send_callback` there is some weird introspection going on to try to decide whether or not to pass arguments to `v2_playbook_on_start`.  It looks like this:
```
for method in methods:
    try:
        # temporary hack, required due to a change in the callback API, so
        # we don't break backwards compatibility with callbacks which were
        # designed to use the original API
        # FIXME: target for removal and revert to the original code here after
        if method_name == 'v2_playbook_on_start':
            import inspect
            (f_args, f_varargs, f_keywords, f_defaults) = inspect.getargspec(me
            if 'playbook' in f_args:
                method(*args, **kwargs)
            else:
                method()
```
This introspection is not documented, which means that if your
callback module has:
```
def v2_playbook_on_start(self, pb):
  ...
```
It will fail, or if it has:
```
def v2_playbook_on_start(self, *args):
   ...
```
It will fail differently, and if it happens to be wrapped by a
decorator that doesn't exactly reproduce function signatures...
```
@commit_after
def v2_playbook_on_start(self, playbook):
    ...
```
It will _also_ fail.

##### EXPECTED RESULTS
I expect my callback function to get called with the expected
parameters, regardless of what I happen to name my arguments.

##### ACTUAL RESULTS
```
```

larsks:
This removes the "temporary hack" for v2_playbook_on_start, because
the callback API has been stable for quite a while now and the
introspection used by this hack can cause unexpected and hard to
diagnose failures in callback code.

Closes #15959

### DIFF
--- a/lib/ansible/executor/task_queue_manager.py
+++ b/lib/ansible/executor/task_queue_manager.py
@@ -319,19 +319,7 @@ class TaskQueueManager:
 
             for method in methods:
                 try:
-                    # temporary hack, required due to a change in the callback API, so
-                    # we don't break backwards compatibility with callbacks which were
-                    # designed to use the original API
-                    # FIXME: target for removal and revert to the original code here after a year (2017-01-14)
-                    if method_name == 'v2_playbook_on_start':
-                        import inspect
-                        (f_args, f_varargs, f_keywords, f_defaults) = inspect.getargspec(method)
-                        if 'playbook' in f_args:
-                            method(*args, **kwargs)
-                        else:
-                            method()
-                    else:
-                        method(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


larsks:
Closing this old PR.  I thought it was helpful.

PR #15966, opened 2016-05-23
Fix vars precedence bug introduced in ansible 2

aperigault:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
2.1
```
##### SUMMARY

Before ansible 2.0 inventory group vars are load after script group vars.
For people that use ansible tower to manage inventories, ansible 2.0 behavior is a big problem. We can't anymore overwrite vars define by group_by module for example.

### DIFF
--- a/docsite/rst/playbooks_variables.rst
+++ b/docsite/rst/playbooks_variables.rst
@@ -817,7 +817,12 @@ In 2.x, we have made the order of precedence more specific (with the last listed
   * role defaults [1]_
   * inventory vars [2]_
   * inventory group_vars
+     * ``group_vars/all`` file
+     * ``group_vars/<group name>`` file
+     * inventory file or script
   * inventory host_vars
+     * ``host_vars/<hostname>`` file
+     * inventory file or script
   * playbook group_vars
   * playbook host_vars
   * host facts
--- a/lib/ansible/vars/__init__.py
+++ b/lib/ansible/vars/__init__.py
@@ -250,9 +250,7 @@ class VariableManager:
                 for item in data:
                     all_vars = combine_vars(all_vars, item)
 
-            # we merge in vars from groups specified in the inventory (INI or script)
-            all_vars = combine_vars(all_vars, host.get_group_vars())
-
+            # then we merge in the group_vars/<group name> file, if it exists
             for group in sorted(host.get_groups(), key=lambda g: (g.depth, g.name)):
                 if group.name in self._group_vars_files and group.name != 'all':
                     for data in self._group_vars_files[group.name]:
@@ -260,8 +258,8 @@ class VariableManager:
                         for item in data:
                             all_vars = combine_vars(all_vars, item)
 
-            # then we merge in vars from the host specified in the inventory (INI or script)
-            all_vars = combine_vars(all_vars, host.get_vars())
+            # then we merge in vars from groups specified in the inventory (INI or script)
+            all_vars = combine_vars(all_vars, host.get_group_vars())
 
             # then we merge in the host_vars/<hostname> file, if it exists
             host_name = host.get_name()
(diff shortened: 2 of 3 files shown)


jimi-c:
Hi @aperigault, the unit tests for this would need to be adjusted as well, plus I would like to see the documentation on variable precedence updated to fully document this grey area of the variable precedence.

aperigault:
I updated the documentation but I think I need help to adjust unit tests.

jimi-c:
@aperigault right, the unit tests were written with the expectation of the variable precedence as it exists in 2.0 now, so any asserts will need to be adjusted accordingly (and may even require some small changes in steps).

aperigault:
@jimi-c I updated unit tests. Tell me if I missed something.

jimi-c:
@aperigault I'm not seeing the unit tests updated, and our tests show them still failing.

aperigault:
@jimi-c I rebase my branch on master and changes break mine. For now I don't understand how can I load inventory group vars after playbook group_vars. Vars seems to be loaded together with https://github.com/aperigault/ansible/blob/74aa516794563e2825cecd7417fd6eaff1630108/lib/ansible/vars/__init__.py#L254-L259 no ?

jimi-c:
Hi @aperigault, sorry for the long delay but just going through old issues and was wondering if this was still a problem? I recall we did address several issues along these lines, but without an example reproducer here I can't test it myself.

Thanks!

aperigault:
@jimi-c This was still a problem. I attempt to fix unit tests but I think I have a problem in my PR. I don't understand the load order between inventory groups/hosts files and playbook groups/hosts files. It seems to be loaded together ? I'm right ? Do you think it's a good thing ?
For me playbook group/host files are like default.yml files (it was the case before 2.0).

jimi-c:
@aperigault we're currently looking into some variable precedence issues (partially because of this issue), however looking at your code you're moving the INI/Script sources to be HIGHER in precedence than they already are, which doesn't seem correct to me. We've always stated those vars are a lower priority, and we've been working on a script to improve our testing of the variable precedence levels here: https://github.com/ansible/ansible/issues/21845

In devel currently, the INI/script source is higher than we have documented, but in a feature branch I've created, the order of variables is as follows:
<pre>
$ ./ansible-var-precedence-check 
TOTAL: 21
1. extra_vars - var passed via the cli
2. set_fact - var made by set_fact
3. include_vars - var in included file
4. task_vars - vars defined on the task
5. role_var - var in ???
6. vars_file - var in file added by vars_file
7. play_var - var set in playbook header
8. pb_host_vars_file - var in playbook/host_vars/host
9. ini_host_vars_file - var in inventory/host_vars/host
10. ini_host - host var inside the ini
11. pb_group_vars_file_child - var in playbook/group_vars/child
12. ini_group_vars_file_child - var in inventory/group_vars/child
13. pb_group_vars_file_parent - var in playbook/group_vars/parent
14. ini_group_vars_file_parent - var in inventory/group_vars/parent
15. pb_group_vars_file_all - var in playbook/group_vars/all
16. ini_group_vars_file_all - var in inventory/group_vars/all
17. ini_child - child group var inside the ini
18. ini_parent - parent group var inside the ini
19. ini_all - all group var inside the ini
20. role_parent_default - var in roles/role_parent/defaults/main.yml
21. role_default - var in roles/role/defaults/main.yml
</pre>
We're still working on adding test cases to this script, but as far as we can tell this order is correct. The general rule of thumb for inventory is:

1) INI/script < inventory < playbook
2) group_vars < host_vars

So combining those two, you get the order above as highlighted by lines 8 through 19.

calfonso:
@aperigault This should now be fixed so closing the issue. We are also working on 2.4, in which you'll be able to configure precedence (you can check it on the devel branch as it was just merged last night).

Thank you!

PR #15934, opened 2016-05-20
If an inventory script throws an error, don't try to parse it as an ini

brandond:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0
  config file = /home/ec2-user/repos/infosec-ansible/ansible.cfg
  configured module search path = Default w/o overrides
```
##### SUMMARY

It's a script. We ran it and and threw and error; no need to be naive and hope against hope it was really an ini file. Most of the time this is just going to result in another indecipherable 'Error parsing host definition' message that obscures the initial script failure.
##### Before:

```
(ansible-env)[ec2-user@l257387-vm infosec-ansible]$ ansible localhost -m ping
ERROR! Inventory script (/home/ec2-user/repos/infosec-ansible/inventory/ec2.py) had an execution error: ERROR: "Error connecting to AWS backend.
Request has expired", while: getting EC2 instances
/home/ec2-user/repos/infosec-ansible/inventory/ec2.py:3: Error parsing host definition ''''': No closing quotation
(ansible-env)[ec2-user@l257387-vm infosec-ansible]$ head -n 3 /home/ec2-user/repos/infosec-ansible/inventory/ec2.py
#!/usr/bin/env python

'''
```
##### After:

```
(ansible-env)[ec2-user@l257387-vm infosec-ansible]$ ansible localhost -m ping
ERROR! Inventory script (/home/ec2-user/repos/infosec-ansible/inventory/ec2.py) had an execution error: ERROR: "Error connecting to AWS backend.
Request has expired", while: getting EC2 instances
```

PR #15969, opened 2016-05-23
skip non executable inventory when no shebang

bcoca:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
2.2
```
##### SUMMARY

alternate fix #15934


abadger:
Looks like this doesn't offer an improvement for binary modules but it does solve the issue with script modules which are going to be far and away the common case.  Failure here is not a traceback but just an error message that isn't as obvious.  The combination of those two makes me feel that it's a cornercase we don't have to address until someone notices it.

+1 to merge as this is an improvment over what we have now.

PR #15980, opened 2016-05-24
cleanup test_binary_modules

robinro:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

devel
##### SUMMARY

Cleans up the current use of test_binary_modules
- remove overly verbose output (ls)
- run only once and not via different connections, the other inventory entries are mostly for testing connection plugins

besides making the code and it's output more readable this reduces the runtime (with hot cache) of 

```
TARGET=centos7 TARGET_OPTIONS="--volume=/sys/fs/cgroup:/sys/fs/cgroup:ro" MAKE_TARGET="test_binary_modules" TEST_FLAGS="" test/utils/run_tests.sh
```

from 41 to 23 seconds.

### DIFF
--- a/test/integration/Makefile
+++ b/test/integration/Makefile
@@ -289,14 +289,12 @@ no_log: setup
 
 test_binary_modules:
 	mytmpdir=$(MYTMPDIR); \
-	ls -al $$mytmpdir; \
 	curl https://storage.googleapis.com/golang/go1.6.2.$(UNAME)-amd64.tar.gz | tar -xz -C $$mytmpdir; \
 	[ $$? != 0 ] && wget -qO- https://storage.googleapis.com/golang/go1.6.2.$(UNAME)-amd64.tar.gz | tar -xz -C $$mytmpdir; \
-	ls -al $$mytmpdir; \
 	cd library; \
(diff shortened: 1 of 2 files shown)


robinro:
@sivel please comment, the changed code was from you

sivel:
I'm ok with removing the `ls` command, as that was debugging that was apparently missed. 

The change to hosts is a different problem.  I am using an alternative inventory for these tests, and we may be using an alternative one for performing these tests on Windows as well, so restricting to `testhost` instead of `all` will cause issues here.

Additionally, due to some differences in connection plugins, it is our intention that it runs for multiple connection types.

robinro:
Can you specify which cases should be covered? At the moment it runs on 11 inventory entries, which only cover `connection=local` and default ssh to localhost as far as I see.

sivel:
Is there some specific problem with running it on multiple hosts?  Excessive, maybe, but I don't think it really degrades performance by much.  It will run parallel, and the tasks are pretty fast.

robinro:
As mentioned before for me it takes 41 seconds instead of 23.

Also (my original problem) it makes the output a lot less readable and debugging problems harder.

robinro:
Since with shippable test-performance is no longer important, I'll close this.

PR #3661, opened 2013-07-25
fixes #3609 Add max_fail_pct to playbook

kavink:
This implements an additional option along with serial option, So if total number of failures for a batch execeed max_fail_pct \* total number of hosts in the batch, do not go to the next serial batch.

Default behavior remains same as before.

PR #15995, opened 2016-05-25
Fixes #3661 - Return the connection type so we can better handle ftp connections…

ronnietaylor:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel d08fda9320) last updated 2016/05/24 21:03:58 (GMT -700)
  lib/ansible/modules/core: (detached HEAD aa995806b9) last updated 2016/05/24 21:04:04 (GMT -700)
  lib/ansible/modules/extras: (detached HEAD d1a4f703ce) last updated 2016/05/24 21:04:09 (GMT -700)
  config file = /Users/ronnietaylor/Documents/Development/ansible-experiments/vangrants/ubuntu1404/ansible.cfg
```
##### SUMMARY

Fixes #3661

Added a return variable to open_url and fetch_url that contains the output from `urlparse.urlparse(url)` (i.e. 'ftp', 'http' or 'https').

This ultimately lets get_url check against something other than just HTTP status codes since successful FTP connections don't return status codes in the response headers like HTTP does. By passing along the connection type we can prevent incorrectly throwing errors for successful FTP connections.

Since the parsing of the url already happens at this level then it seemed best to just capture it once and pass it through to get_url instead of parsing the url again later.

This new variable will contain a string like the following:

```
>>> conn_type
'ftp'
```

… in the get_url module.

### DIFF
--- a/lib/ansible/module_utils/urls.py
+++ b/lib/ansible/module_utils/urls.py
@@ -837,7 +837,8 @@ def open_url(url, data=None, headers=None, method=None, use_proxy=True,
         urlopen_args.append(timeout)
 
     r = urllib2.urlopen(*urlopen_args)
-    return r
+    conn_type = parsed[0]
+    return r, conn_type
 
 #
 # Module-related functions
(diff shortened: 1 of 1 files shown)


ronnietaylor:
I noticed some of the integration tests failed here at the same time I noticed a regression in my local code base after introducing this change.

I'd like to propose a simpler solution which I've noted in this related pull request:

https://github.com/ansible/ansible-modules-core/pull/3743

I'd love to get some feedback so I can finish this up.

Thanks!

ghost:
Ran in to this bug today when I experimented with get_url and found this bugfix.
Tested to parse the url and exclude "ftp:/" in the if-statement. Works in my simple test.

The commit
https://github.com/tecarlsson/ansible-modules-core/commit/12171de4e8a7094b9bac72b2e89bc3135ea92557

But I think it would be better to change the if-statement to just include http/https instead of excluding file and ftp.

ronnietaylor:
Closing this and creating a single, streamlined pull request.

PR #15998, opened 2016-05-25
added more descriptive error when missing shebang caused by BOM

iagogarrido:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0
  config file = 
```
##### SUMMARY

I add a additional check for a BOM mask in the first line of the module, if the shebang is not found.

Yesterday, I spent 1 hour guessing why my custom powershell module runs correctly in the host machine and crashes when running in a playbook. I think it's important to outline this error (it seems Powershell ISE saves powershell scripts with BOM by default)

### DIFF
--- a/lib/ansible/executor/module_common.py
+++ b/lib/ansible/executor/module_common.py
@@ -22,6 +22,7 @@ __metaclass__ = type
 
 import ast
 import base64
+import codecs
 import imp
 import json
 import os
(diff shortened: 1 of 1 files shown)


lib/ansible/executor/module_common.py
```diff
@@ -740,6 +749,8 @@ def modify_module(module_name, module_path, module_args, task_vars=dict(), modul
 
             if os.path.basename(interpreter).startswith(b'python'):
                 lines.insert(1, to_bytes(ENCODING_STRING))
+        elif _check_bom(lines[0]):
+            raise AnsibleError("BOM detected in module (%s). Module scripts encoded with BOM are not supported." % module_name)
```
bcoca: for users, a nicer message that explains what BOM is might be nicer:

'Unsupported byte order mark detected, this type of file encoding is not supported for modules'

iagogarrido:
I agree. I change the message by a more descriptive one.

nitzmahone:
See commit reference above: we went about a different way of fixing this since we already know the module is powershell (the shebang isn't actually needed). 

Thanks!

PR #15999, opened 2016-05-26
Fix unarchive failures when the destination is a symlink to a directory

### LINKED ISSUE #3725: add a regex Jinja2 filter
Folks would like to do

```
foo | search(some_regex)
```

in expressions from Jinja2 and re.search does not work.

Need to test if foo.search is already avail in regular templates and then it may just be a matter of importing in playbooks to enable foo.search

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
[pts/26@roan /srv/ansible/abadger/test/integration]$ ansible --version          *[fix-unarchive-symlink]  (18:49:46)
ansible 2.2.0 (fix-unarchive-symlink 67be03521e) last updated 2016/05/25 18:43:58 (GMT -700)
  lib/ansible/modules/core: (detached HEAD 92bf802cb8) last updated 2016/05/18 05:56:45 (GMT -700)
  lib/ansible/modules/extras: (detached HEAD e710dc47fe) last updated 2016/05/18 05:56:45 (GMT -700)

```
##### SUMMARY

When the destination of unarchive is a symlink to a directory instead of the directory itself, unarchive would fail.  This was because it was seeing that the symlink was not a directory and thus did not think that it was a valid target.  This PR modifies the way unarchive examines the destination so that it follows the symlink.  That way if the symlink points at a directory unarchive will proceed.

Also add integration tests for this problem and unicode filenames inside
a tarball.

Fixes #3725


abadger:
Tests passing, Travis failures are unrelated.  merging.

abadger:
merged to devel and stable-2.1

PR #16012, opened 2016-05-26
Modules: check for list-like choices in arg spec

stpierre:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
% ansible --version                                            
ansible 2.2.0 (list-like-choices a27804eed6) last updated 2016/05/26 12:07:33 (GMT -500)
  lib/ansible/modules/core: (detached HEAD fb77ab49ab) last updated 2016/05/26 11:56:22 (GMT -500)
  lib/ansible/modules/extras: (detached HEAD 0e4a023a7e) last updated 2016/05/26 11:56:26 (GMT -500)
```

And also:

```
% ansible --version
ansible 1.9.4
  configured module search path = ./library:$ANSIBLE_HOME/library
```
##### SUMMARY

This makes it possible to use anything other than a list (e.g., a
tuple, or dict.keys() in py3k) for argument_spec choices. It also
improves the error messages if you don't use a list type.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -1322,14 +1322,14 @@ class AnsibleModule(object):
             choices = v.get('choices',None)
             if choices is None:
                 continue
-            if type(choices) == list:
+            if isinstance(choices, SEQUENCETYPE):
                 if k in self.params:
                     if self.params[k] not in choices:
                         choices_str=",".join([str(c) for c in choices])
                         msg="value of %s must be one of: %s, got: %s" % (k, choices_str, self.params[k])
                         self.fail_json(msg=msg)
             else:
-                self.fail_json(msg="internal error: do not know how to interpret argument_spec")
+                self.fail_json(msg="internal error: choices for argument %s are not iterable: %s" % (k, choices))
 
     def safe_eval(self, str, locals=None, include_exceptions=False):
 


nitzmahone:
@stpierre I'm not seeing how this addresses the issue for most of the types you mention- `_check_type_list()` doesn't return bools (it raises TypeError), and `isinstance(x,list)` (its primary applicable check) returns False for tuples and most other listish types I tried...

stpierre:
I pretty clearly failed at reading code. Wow. I'll resubmit something that isn't egregiously terrible.

stpierre:
This might actually work.

The other approach would be:

```
if hasattr(choices, "__iter__") and hasattr(choices, "__contains__"):
    ...
```

But that looks horrid. The TypeError approach runs the risk of overzealously catching some other TypeError, but that seems unlikely.

sivel:
@stpierre we have handled some py3 things around the dict view types like you can see in https://github.com/ansible/ansible/commit/cf44db58e051e9c18f15c5ef698cca065711dd9e

Just as potential insight into other places doing similar things.

stpierre:
@sivel Looks reasonable. I've updated my PR to use that approach -- consistency FTW.

nitzmahone:
WFM, thanks!

nitzmahone:
Since Travis seems to have taken off early for the weekend, and this is using an approach we're already using elsewhere, I'll merge before the test run passes... 🔥

PR #16034, opened 2016-05-27
raise an exception on undefined environment variables

### LINKED ISSUE #16032: Ansible will create a literal $HOME directory if the variable is not set
##### ISSUE TYPE

Bug Report
##### ANSIBLE VERSION

ansible 2.1.0.0
##### CONFIGURATION

Nothing relevant
##### OS / ENVIRONMENT

Not platform specific
##### SUMMARY

When the $HOME variable is not set, Ansible creates a literal $HOME directory in the current working directory:

```
(venv) ┬─[dmsimard@hostname:~/dev/ara/$HOME/.ansible]─[10:53:57 AM]
╰─> (static-generation) $ tree
.
└── tmp
    ├── ansible-local-tmp2Gdd9B
    ├── ansible-local-tmp3DDjOt
    ├── ansible-local-tmp3zLvr4
    ├── ansible-local-tmp3ZRISO
    ...
```

Alt text
##### STEPS TO REPRODUCE

Inventory:

```
```

Unset $HOME, run Ansible.
I came across this problem when running Ansible from tox without "HOME" being specified in "passenv" so the $HOME variable wasn't passed to the tox environment.
##### EXPECTED RESULTS

I don't expect Ansible to create a literal $HOME directory in the current working directory. At worst, create a ".ansible" folder if $HOME is not set.
##### ACTUAL RESULTS

A $HOME directory is created in the current working directory.

larsks:
This commit replaces `os.path.expandvars` with code that will detect
and error out on undefined environment variables.  This prevents the
current behavior described in #16032, in which undefined environment
variables will be used literally.

Closes #16032

### DIFF
--- a/lib/ansible/constants.py
+++ b/lib/ansible/constants.py
@@ -20,6 +20,7 @@ from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
 import os
+import re
 import tempfile
 from string import ascii_letters, digits
 
@@ -28,6 +29,10 @@ from ansible.compat.six.moves import configparser
 
 from ansible.parsing.quoting import unquote
 from ansible.errors import AnsibleOptionsError
+
+
+re_shell_var_bare = re.compile('\$([a-zA-Z]\w*)')
+re_shell_var_braces = re.compile('\${([a-zA-Z]\w*)}')
 
 # copied from utils, avoid circular reference fun :)
 def mk_boolean(value):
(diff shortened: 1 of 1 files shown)


bcoca:
Actually it will not, the expandvars there does not execute on the remote machine, the remote does not use lib/constants.py, it is passed directly to the shell for it to interpret it `mkdir <path>`. 

It doesn't make sense to 'translate it' on the controller except when both machines are the same `~bcoca` on controller can be `/home/bcoca`, but on target `/usr/home/bcoca`.

This will affect local paths created on the controller itself.

larsks:
@bcoca, that is exactly the problem this is attempting to solve (anisble on the controller creating paths containing literal `$HOME` or other undefined environment variables).

westsouthnight:
@bcoca I very need that, thats reproduce with consul_kv

Link on stackoverflow - http://stackoverflow.com/questions/36043393/set-environment-in-ansible-doesnt-seem-to-work/37950548#37950548

larsks:
Closing an old PR.

PR #16038, opened 2016-05-27
docs: add some examples to the documentation for become

larsks:
there were some questions on `#ansible` about the `become` suite of
directives, and I thought it would be helpful if this document had an
example or two.


bcoca:
looks good to me, assigning to @docschick for final approval

docschick:
I'm good with the changes. :) Thanks @larsks for the PR and docs update!

PR #16056, opened 2016-05-30
add check_mode option for tasks

### LINKED ISSUE #12498: Allow running task or play in check_mode from playbook
Ansible check_mode allows determining if a task _would_ have changed something, but only on the level of an entire play.

But there are obvious cases where idempotency requires being able to change the order of events - i.e. check if a file would be changed, if so, stop a daemon and change the file - otherwise do nothing.

This can be replicated by comparing mtimes locally and remotely, but leads to complicated syntax for something the existing feature can already provide.

Ideally something like this would be possible:

``` yaml
- hosts: webservers
  tasks:
  - copy: src=monitoredfile dest=monitoredfile
    check_mode: true
    register: copy_check

  - service: name=webserver state=stopped
    when: copy_check.changed

  - copy: src=monitoredfile dest=monitoredfile
    when: copy_check.changed

  - service: name=webserver state=started
    when: copy_check.changed
```

This is simple and readable, and helps in dealing with services which may be monitoring config directories for new files but which would might like to try and avert for changes like that (exim with a perl interpreter can be a problem in this regard).

robinro:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

devel, 2.2
##### SUMMARY

Introduce an option `check_mode` that can be added to tasks. Forces or prevents the module to run under check_mode.

Replaces the `always_run` option (now: `check_mode: no`) but extends it by allowing to run under check mode even if check mode is not specified on the command line (`check_mode: yes`).

Fixes #12498. See that issue and links therein also for discussion why this is nice.

Includes example testcases for the template module. One of the benefits of including per-task check_mode is that testing, both of ansible itself and external projects using ansible, becomes easier.

### DIFF
--- a/docsite/rst/playbooks_checkmode.rst
+++ b/docsite/rst/playbooks_checkmode.rst
@@ -30,7 +30,7 @@ be added to tasks.
 
 If the playbook is run with checkmode on the command line, you can force tasks to still run, by specifying `check_mode: no`.
 On the other hand you can setup a test section which, independetly of the command line arguments, always runs a tasks in check mode with `check_mode: yes`.
-note:: Prior to version 2.2 only the ability to turn this on per-task existed.  The notation for that was `always_run: yes`.
+.. note:: Prior to version 2.2 only the ability to turn this on per-task existed.  The notation for that was `always_run: yes`.
 
 Instead of `yes`/`no` you can use a Jinja2 expression, just like the `when` clause.
 
--- a/docsite/rst/playbooks_checkmode.rst
+++ b/docsite/rst/playbooks_checkmode.rst
@@ -25,26 +25,45 @@ Enabling or disabling check mode for tasks
 
 .. versionadded:: 2.2
 
-Sometimes you may want to modify the check mode behavior of individual tasks. This is done via the `check_mode` option, which can
-be added to tasks.
+Sometimes you may want to modify the check mode behavior of individual tasks. This is done via the ``check_mode`` option, which can
+be added to tasks. 
 
-If the playbook is run with checkmode on the command line, you can force tasks to still run, by specifying `check_mode: no`.
-On the other hand you can setup a test section which, independetly of the command line arguments, always runs a tasks in check mode with `check_mode: yes`.
-.. note:: Prior to version 2.2 only the ability to turn this on per-task existed.  The notation for that was `always_run: yes`.
+There are two options:
 
-Instead of `yes`/`no` you can use a Jinja2 expression, just like the `when` clause.
+1. Force a task to **run in check mode**, even when the playbook is called **without** ``--check``. This is called ``check_mode: yes``.
+2. Force a task to **run in normal mode** and make changes to the system, even when the playbook is called **with** ``--check``. This is called ``check_mode: no``.
+
+.. note:: Prior to version 2.2 only the the equivalent of ``check_mode: no`` existed. The notation for that was ``always_run: yes``.
+
+Instead of ``yes``/``no`` you can use a Jinja2 expression, just like the ``when`` clause.
 
 Example::
 
     tasks:
 
-      - name: this task is run even in check mode
+      - name: this task will make changes to the system even in check mode
         command: /something/to/run --even-in-check-mode
         check_mode: no
 
-Also if you want to skip, or ignore errors on some tasks in check mode
-you can use a boolean magic variable `ansible_check_mode` (added in version 2.1)
-which will be set to `True` during check mode.
+      - name: this task will always run under checkmode and not change the system
+        lineinfile: line="important config" dest=/path/to/myconfig.conf state=present
+        check_mode: yes
+
+
+Running single tasks with ``check_mode: yes`` can be useful to write tests for
+ansible modules, either to test the module itself or to the the conditions under
+which a module would make changes. 
+With ``register`` (see :doc:`playbooks_conditionals`) you can check the
+potential changes.
+
+Information about check mode in variables
+`````````````````````````````````````````
+
+.. versionadded:: 2.1
+
+If you want to skip, or ignore errors on some tasks in check mode
+you can use a boolean magic variable ``ansible_check_mode``
+which will be set to ``True`` during check mode.
 
 Example::
 
--- a/docsite/rst/playbooks_checkmode.rst
+++ b/docsite/rst/playbooks_checkmode.rst
@@ -30,6 +30,7 @@ be added to tasks.
 
 If the playbook is run with checkmode on the command line, you can force tasks to still run, by specifying `check_mode: no`.
 On the other hand you can setup a test section which, independetly of the command line arguments, always runs a tasks in check mode with `check_mode: yes`.
+note:: Prior to version 2.2 only the ability to turn this on per-task existed.  The notation for that was `always_run: yes`.
 
 Instead of `yes`/`no` you can use a Jinja2 expression, just like the `when` clause.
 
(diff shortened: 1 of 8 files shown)


robinro:
btw: I was quite impressed when I realized that this is actually a 3-line patch. Ansible is really flexible!

bcoca:
the patch does not handle correctly when check mode is specified in the command line but the task has it set to false.

robinro:
The case where check mode is given on the command line but we still want to run the task was so far covered by `always_run`. I extended the implementation to replace `always_run` by `check_mode: no` and have `check_mode: yes` to enforce check mode even if it is not given on the command line.

This implements/extends the _rename_always_run_ proposal by @resmo: https://github.com/ansible/proposals/blob/master/rename_always_run.md.

lib/ansible/playbook/play_context.py
```diff
@@ -434,8 +434,12 @@ def set_task_and_variable_override(self, task, variables, templar):
         # set become defaults if not previouslly set
         task.set_become_defaults(new_info.become, new_info.become_method, new_info.become_user)
 
+        if task.check_mode is not None:
```
bcoca, line 442: ~~you still don't deal with `check_mode: False` and~~ now always_run will override check_mode setting, I would reverse it in view of deprecating always_run

robinro:
@bcoca makes sense, I switched the two to have `check_mode` overwrite `always_run`

bcoca:
nice, now we just need enough committers to agree.

docsite/rst/playbooks_checkmode.rst
```diff
@@ -20,27 +20,27 @@ Example::
...
-Sometimes you may want to have a task to be executed even in check
-mode. To achieve this, use the `always_run` clause on the task. Its
-value is a Jinja2 expression, just like the `when` clause. In simple
-cases a boolean YAML value would be sufficient as a value.
+Sometimes you may want to modify the check mode behavior of individual tasks. This is done via the `check_mode` option, which can
+be added to tasks.
+
+If the playbook is run with checkmode on the command line, you can force tasks to still run, by specifying `check_mode: no`.
+On the other hand you can setup a test section which, independetly of the command line arguments, always runs a tasks in check mode with `check_mode: yes`.
+note:: Prior to version 2.2 only the ability to turn this on per-task existed.  The notation for that was `always_run: yes`.
```
abadger: For proper rst syntax, this should be:

```
.. note:: Prior to[...]
```

abadger:
+1 from me.

robinro:
Based on the discussion in the core meeting I updated the docs (still mention always_run) and changed the version for deprecating always_run to 2.4.

Thanks for the review!

alxgu:
+1

robinro:
@abadger @bcoca Can this be merged? In the core-meeting discussion 3 weeks ago you approved it: https://meetbot.fedoraproject.org/ansible-meeting/2016-06-09/ansible_core_team_public_meeting.2016-06-09-15.11.log.html

The shippable failure is unrelated afaik.

lib/ansible/playbook/play_context.py
```diff
@@ -434,10 +434,15 @@ def set_task_and_variable_override(self, task, variables, templar):
         # set become defaults if not previouslly set
         task.set_become_defaults(new_info.become, new_info.become_method, new_info.become_user)
 
-        # have always_run override check mode
         if task.always_run:
+            display.deprecated("always_run is deprecated. Use check_mode = no instead.", version="2.4", removed=False)
```
resmo, line 438: I would change the message to just: `always_run is deprecated. Use check_mode instead`

robinro, line 438: In 2.1 one used `always_run: yes` and with the short message one might try `check_mode: yes`, which is the opposite.

amenonsen:
As documented and implemented, I find this feature extremely confusing and unintuitive. I understand the motivation, and think the underlying feature is desirable (in fact, I submitted a PR for it years ago), but I don't think it should be merged without revisions.

robinro:
@amenonsen can you be more specific and suggest improvements?
I thought about this change a lot and think that check_mode: yes/no is the most straight-forward way to call it and also does not conflict with the previous implementation.

bcoca:
@amenonsen can you expand on your comments? otherwise i'll be merging this soon. 

I do think we need to have very clear and explicit docs on how this is different than 'always_run' and how it makes it obsolete.

mattclay:
The expanded documentation on this helps. I think the only thing that would make it more clear is if the options for check_mode where "always" or "never", instead of a boolean. However, I'm willing to accept this as is.

robinro:
Thanks for merging @bcoca

dagwieers:
According to the documentation:

> There are two options:
> - Force a task to run in check mode, even when the playbook is called without --check. This is called check_mode: yes.
> - Force a task to run in normal mode and make changes to the system, even when the playbook is called with --check. This is called check_mode: no.

I understand where we are coming from, and see how you now have more possibilities to get something run despite being in check-mode. But having to add `check_mode: no` to a task that needs to run in check-mode as well as normal run-mode (which is to most common use-case) is so counter-intuitive that it becomes a big usability problem IMO.

Compare this to: `always_run: yes` and you can see how this is very clear to what it means, whereas `check_mode: no` does the opposite of what you would assume it would do. The boolean does not properly indicate what it is switching on/off.

This is far from the idea that a playbook is human-readable and easy to understand. And I would propose to revert the deprecation of `always_run` in v2.0.1 **and** replace the check_mode option with something that is far more readable. (before people start using this...)

robinro:
@dagwieers Can you propose a better notation? Or is your proposal to go back to `always_run` and not unify the `check_mode: yes/no` options in one variable?

I like the `check_mode: yes/no` notation. The tasks `run` always anyway, but sometimes they run in `check_mode` and sometimes they don't. There is a flag on the command line to enable/disable checkmode for all tasks and if you set `check_mode: yes/no` on a task it overwrites the global flag.

If you are a `always_run` user it might be countintuitive, but if you start unbiased I feel this is the right setup.

dagwieers:
The problem is we are using the term _check_mode_ for two distinct things. One such thing is the mode we are running in and the other thing is how we want the task to behave when being run. And that is causing the confusion.

And as such we have 4 different cases:
- Execute when in run-mode, run task as if in run-mode (normal behaviour)
- Execute when in check-mode, run task as if in check-mode (normal behaviour when --check)
- Execute when in run-mode, run task as if in check-mode (check_mode: yes)
- Execute when in check-mode, run task as in run-mode (check_mode: no)

The option `always_run` was straightforward because it only applied to the mode we were running in. And simply meant run it in both run-mode as well as in check-mode. But the `check_mode` option is diffusing all this by adding the potential to modify _how_ a task is being run, not just _when_ it ought to run.

So if we would want to allow common combinations, an option could be

``` yaml
    run_mode: [ check, normal ]
    behave: check/normal
```

And the default behaviour is that a task behaves as the run-mode it runs in. Another way to write it out could be specify the behaviour for each running mode (even more options):

``` yaml
    check_mode: check/run
    run_mode: run/check
```

and while that is very explicit and even gives more freedom (you could make it do the exact opposite), I would prefer we introduce new terminology for the task-behavior (operate, behave, act, ...). Why ? Because run-mode/check-mode/diff-mode are already known/loaded terms. Having new terminology will make it easier to communicate with one another, especially for giving support.

Another issue I have with the current design (but also with _always_run_) is the fact that it limits to the two run-modes we have today, but we may have other run-modes in the future (in fact, diff-mode + check-mode is already close to a report-mode I envisioned once) and the current design. So I would prefer a solution that doesn't forces us in the existing model. (E.g. some users may want to always enable diff-mode for specific tasks.)

So unfortunately I think we jumped the gun on this one.

PR #16062, opened 2016-05-31
Fix sleep intervals (14219)

### LINKED ISSUE #14219: High CPU Usage
**Issue Type:** Bug/Annoyance
**Ansible Version:**

```
ansible 2.0.0.2

```

**Ansible Configuration:** 
Default configuration supplied from Ubuntu PPA except uncommented:

```
[privilege_escalation]
become=True
become_method=sudo
```

**Environment:**
Ubuntu 15.10, Python 2.7.10 running the playbook.  The target host in group is ArchLinux with python2 version Python 2.7.11

**Summary:**
When I run a playbook that runs a long task (say backup target machine, or do virus scanning), the machine running the playbook spawns 2 "python /usr/bin/ansible-playbook stress.yml" processes that consume a high amount of CPU (15-20% each on my machine which is not that powerful).  Since the machine running the playbook is pretty much waiting for remote task to complete it shouldn't use up this much CPU.

It doesn't seem to matter whether the host group contains 1 machine or more.  This happens when playbook is executed in serial mode as well.

**Steps To Reproduce:**
For some host group, let's say "stress" (where the machine running ansible command is not part of that group), execute something that takes a while on the remote machine.  Eg:
`ansible stress -m shell -a "dd if\=/dev/random of\=/dev/null bs\=1M count\=2"`
or
`ansible stress -m shell -a "stress --cpu 8 --timeout 60s"`
Monitor cpu usage with top on the machine that ran ansible command.

**Expected Results:**
CPU usage closer to idle.  I did not have this problem with Ansible 1.9.2.

**Actual Results:**
Fairly high CPU consumption.

andriishestakov:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0
```
##### SUMMARY

Increase sleep intervals because too often polling causes high CPU usage.
Fixes #14219

### DIFF
--- a/lib/ansible/executor/process/result.py
+++ b/lib/ansible/executor/process/result.py
@@ -106,7 +106,7 @@ class ResultProcess(multiprocessing.Process):
             try:
                 result = self._read_worker_result()
                 if result is None:
-                    time.sleep(0.0001)
+                    time.sleep(0.01)
                     continue
 
                 # send callbacks for 'non final' results
--- a/lib/ansible/plugins/strategy/__init__.py
+++ b/lib/ansible/plugins/strategy/__init__.py
@@ -196,7 +196,7 @@ class StrategyBase:
                 self._cur_worker += 1
                 if self._cur_worker >= len(self._workers):
                     self._cur_worker = 0
-                    time.sleep(0.0001)
+                    time.sleep(0.01)
                 if queued:
                     break
 
@@ -389,7 +389,7 @@ class StrategyBase:
                     raise AnsibleError("unknown result message received: %s" % result[0])
 
             except Queue.Empty:
-                time.sleep(0.0001)
+                time.sleep(0.01)
 
             if one_pass:
                 break
@@ -408,7 +408,7 @@ class StrategyBase:
         while self._pending_results > 0 and not self._tqm._terminated:
             results = self._process_pending_results(iterator)
             ret_results.extend(results)
-            time.sleep(0.0001)
+            time.sleep(0.01)
         display.debug("no more pending results, returning what we have")
 
         return ret_results


jimi-c:
@andreyshestakov but too low causes performance issues. Could you please create a test doing a simple 1000 pings and report the total CPU time as well as the total wall clock time to run with and without your patch? In my experience, since we're just doing poll and sleep, the CPU time is really not that much higher despite the increased polling frequency.

andriishestakov:
@jimi-c I not sure what means simple 1000 pings, maybe ping localhost 1000 times?
I have following playbook which just do pause:

```
---
- hosts: localhost
  tasks:
    - pause: minutes=5
```

Ansible from devel branch:

```
# ansible-playbook --version
ansible-playbook 2.2.0
  config file = 
```

Running on VM with Ubuntu 3.13.0-86-generic and 2 VCPU Intel(R) Xeon(R) CPU E5-2620 v3 @ 2.40GHz:

```
# time ansible-playbook pp

ok: [localhost]

TASK [pause] *******************************************************************
Pausing for 300 seconds
(ctrl+C then 'C' = continue early, ctrl+C then 'A' = abort)
ok: [localhost]

real    5m5.048s
user    0m35.873s
sys 0m20.565s
```

With fix:

```
# time ansible-playbook pp

ok: [localhost]

TASK [pause] *******************************************************************
Pausing for 300 seconds
(ctrl+C then 'C' = continue early, ctrl+C then 'A' = abort)
ok: [localhost]

real    5m4.025s
user    0m4.536s
sys 0m2.555s
```

Real time is same, but with 0.01 sleep intervals CPU time is much less then with 0.0001.

My real case is use 'wait_for' module for wait until remote server appears online.
I think issue with high CPU usage is same in both cases.

jimi-c:
@andreyshestakov I mean a playbook with 1000 pings in it, to localhost. I find that to be a good timing test as the module execution is negligible and highlights how small compounded time differences for each task can add up in large playbooks.

jimi-c:
I went ahead and did it. With your patch:

```
real    2m23.537s
user    1m31.774s
sys 0m36.252s
```

Without (devel):

```
real    2m5.186s
user    1m40.636s
sys 0m38.731s
```

So with your patch there's about 18s more execution time while saving about 9s in CPU time. I do see how in a long sleep as you demonstrate how the CPU time is quite a bit less, so the trade off may be worth it.

jimi-c:
Splitting the difference (kind of) and using 0.005:
1000 pings:

```
real    2m13.686s
user    1m33.099s
sys 0m35.889s
```

5 minute sleep:

```
real    5m0.380s
user    0m4.132s
sys 0m1.241s
```

So this seems a bit better while maintaining the much lower CPU usage during long polls.

jimi-c:
Merged and tweaked the poll times to be 0.005 as above.

Thanks!

PR #15924, opened 2016-05-19
Refactored gce util module to support other GCP services

walbert947:
##### ISSUE TYPE
- New Module Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (gcp_module_util 9cb2ab4e2c) last updated 2016/05/19 02:57:46 (GMT -700)
  lib/ansible/modules/core: (detached HEAD 92bf802cb8) last updated 2016/05/18 11:19:41 (GMT -700)
  lib/ansible/modules/extras: (detached HEAD e710dc47fe) last updated 2016/05/18 11:19:51 (GMT -700)
  config file = /home/william/.ansible.cfg
  configured module search path = ['/usr/share/ansible']
```
##### SUMMARY

This is a refactoring of the existing GCE utility module to support other projects on Google Cloud Platform.

The current `gce.py` module is hard-coded specifically for GCE, and attempting to use it with other projects in GCP fails. [This post](https://github.com/ansible/ansible/pull/15918#issuecomment-220165913) goes into more detail about the problem. This has also been an issue for others in the past, although they've handled it by simply duplicating some of the logic of `gce.py` in their own modules ([example](https://github.com/ansible/ansible-modules-core/blob/devel/cloud/google/gce_lb.py#L221-226)).

OTOH, I didn't want to duplicate code for my own modules, since we have a perfectly good utility module available. :stuck_out_tongue:

Here's what I changed:
- The existing `gce.py` module was renamed to `gcp.py`, and modified to remove any imports or other code that refers to `libcloud.compute` or GCE (the GCE_\* params were retained for compatibility). I also renamed the `gce_connect` function to `gcp_connect`, and modified the function signature to make supplying a provider, driver, and agent information mandatory.
- A new `gce.py` module was created to handle connectivity to GCE. It imports the appropriate `libcloud.compute` providers and drivers, and then passes them on to `gcp_connect` in `gcp.py`. The constants and function signatures are the same as the old `gce.py`, so compatibility with existing modules is retained.
- A new `gcdns.py` module was created to support my PR https://github.com/ansible/ansible-modules-extras/pull/2252 for two new Google Cloud DNS modules, and to demonstrate support for a non-GCE Google Cloud service. It follows the same basic structure as the new `gce.py` module, but imports from `libcloud.dns` instead.

PR #16064, opened 2016-05-31
Allow the use of paths like ~/.project.json

jayme-github:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file = /home/janis/.ansible.cfg
```
##### SUMMARY

This makes it easy to manage credentials files outside of the repo (and/or use user specific credentials). For example:

``` yaml
      vars:
         service_account_email: unique-id@developer.gserviceaccount.com
         credentials_file: ~/.project.json
         project_id: project-id
         machine_type: n1-standard-1
         image: debian-7
```


jayme-github:
Updated PR to reflect changes from #15924 .

tback:
What is blocking this? It seems simple to me and it's unlikely that it'll break anything.

lib/ansible/module_utils/gcp.py
```diff
@@ -231,7 +231,7 @@ def _validate_credentials_file(module, credentials_file, require_valid_json=True
     """
     try:
         # Try to read credentials as JSON
-        with open(credentials_file) as credentials:
+        with open(os.path.expanduser(credentials_file)) as credentials:
```
bcoca: this is the wrong fix, set the type=path in argspec and it will also resolve issus like $HOME/file.json

tback: Tried that, didn't work! How is this supposed to work? I've never heard of python doing magic on argspecs.

bcoca: no magic, type 'path' is run via expanduser AND expandvars, it is part of the AnsibleModule class. Can you define 'didn't work' as that would affect many modules?

tback: You don't want to know what I thought. Fixed it now.

tback:
Everything should be fine now. Please let me know if you need anything else for this to pass.

supertom:
Thanks for the PR!  (Sorry I missed this, it looks like it didn't have the 'gce' label.)  LGTM!

tback:
Do you mind to merge it then?

PR #16076, opened 2016-05-31
Pin paramiko to < 2

### LINKED ISSUE #15665: Backwards incompatible change in paramiko breaking install
##### ISSUE TYPE
Bug Report

##### COMPONENT NAME
lib/ansible/connection/paramiko_ssh.py

##### ANSIBLE VERSION
```
All versions using unspecified version of paramiko (https//github.com/ansible/ansible/blob/devel/setup.pyL25)
```

##### SUMMARY
A backwards incompatible change was introduced with the release of paramiko 2, see http://www.paramiko.org/changelog.html
Since paramiko now uses the cryptography library, systems that don't have the requirements for that library installed will suddenly not be able to install Ansible through pip anymore.

##### STEPS TO REPRODUCE
Install Ansible through pip, on a system that does not have all required libs to install cryptography through pip.

2rs2ts:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.0.2.0
  config file = /home/admin/.ansible.cfg
```
##### SUMMARY

Pins the paramiko dependency to < 2.0.0, since the 2.0.0 release introduced breaking changes and prevents successful installs of ansible unless paramiko is already installed at 1.x or the necessary crypto libs are already present and compiled on the system.

Fixes #15665

Signed-off-by: Jesse Szwedko jesse.szwedko@getbraintree.com

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ setup(name='ansible',
       license='GPLv3',
       # Ansible will also make use of a system copy of python-six if installed but use a
       # Bundled copy if it's not.
-      install_requires=['paramiko', 'jinja2', "PyYAML", 'setuptools', 'pycrypto >= 2.6'],
+      install_requires=['paramiko < 2.0.0', 'jinja2', "PyYAML", 'setuptools', 'pycrypto >= 2.6'],
       package_dir={ '': 'lib' },
       packages=find_packages('lib'),
       package_data={


sivel:
I'm not really sure I agree with this.  I do understand that due to paramiko changing its dependencies, it has caused an install to potentially fail, although this is pretty easily remediated, by installing the necessary dependencies.

It should be the job of OS packagers to pin packages to versions available on a specific OS and handle dependencies appropriately.

However there are benefits to the upgrade and dependency change that paramiko made such as improved performance.

To the best of my knowledge, other than a change in dependency there are no backwards incompatible changes as it affects Ansible.  Pinning to a lower version to prevent needing to update dependencies at the price of not getting better performance through the updated crypto lobs does not make sense to me.

Perhaps documentation should be updated, instead of this change.

bcoca:
I agree with everything @sivel says above, this is not something we should 'fix' this way as it prevents Ansible from benefiting from fixes and bugs that get added to the paramiko codebase.

2rs2ts:
Then you should consider pinning to exact versions of dependencies so that a user can downgrade ansible to avoid an issue like this one. I understand the desire to benefit from bugfixes but you also suffer from regressions if you don't pin versions. Food for thought.

PR #16084, opened 2016-06-01
Revert "Add prefixing and suffixing fuctionality to assemble module"

bcoca:
Reverts ansible/ansible#10399

this is missing docs and should deal with conflicting delimiter settings

### DIFF
--- a/lib/ansible/plugins/action/assemble.py
+++ b/lib/ansible/plugins/action/assemble.py
@@ -32,18 +32,13 @@ class ActionModule(ActionBase):
 
     TRANSFERS_FILES = True
 
-    def _assemble_from_fragments(self, src_path, delimiter=None, compiled_regexp=None, ignore_hidden=False, header=None, footer=None):
+    def _assemble_from_fragments(self, src_path, delimiter=None, compiled_regexp=None, ignore_hidden=False):
         ''' assemble a file from a directory of fragments '''
 
         tmpfd, temp_path = tempfile.mkstemp()
         tmp = os.fdopen(tmpfd,'w')
         delimit_me = False
         add_newline = False
-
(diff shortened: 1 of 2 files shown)


underyx:
😿

underyx:
@bcoca I'm not sure I understand the what further docs are required other than https://github.com/ansible/ansible-modules-core/pull/901/files#diff-204c9a1c75ed1b68284601e5efe69ac0R65

I also am not sure why the header and footer would need to be concerned with delimiters. Let's assume you're trying to build a file with nice little comments to guide the reader, like in the current docs examples:

``` yml
- assemble: src={{ fragments_path }} delimiter="# --- moving on to the next fragment! ---" prefix="# FYI these fragments are assembled from {{ fragment_path }}/* by Ansible"
```

run this and you get:

```
# FYI these fragments are assembled from /foo/bar/* by Ansible
fragment 1
# --- moving on to the next fragment! ---
fragment 2
# --- moving on to the next fragment! ---
fragment 3
```

This I think makes way more sense than having the delimiter between the header and the first fragment as well, and
1. I'm pretty sure this use case would be more popular.
2. If we have no delimiter there by default, it can still be added manually to the `header` string, making this feature useful for everyone, while it would be impossible to remove an unneeded delimiter from between the header and first fragment if we went with that as the default behavior.

underyx:
@chouseknecht heeelp :D

bcoca:
well, part of the missing docs is exactly what you explained above

other issues i still have with this are mixing inline strings with files and opening another `copy: content={{sttuff}}` source of headaches. The same functionality can be achieved by using assemble or even template again.

underyx:
Alright, I'll be coming to a core meeting then.

PR #16085, opened 2016-06-01
raw and script should not use default executable

### LINKED ISSUE #16054: Changes in raw module
##### ISSUE TYPE
- Bug Report
##### ANSIBLE VERSION

```
ansible --version
ansible 2.1.0.0
  config file = /home/pujan14/.ansible.cfg
```
##### OS / ENVIRONMENT

"N/A"
##### SUMMARY

ssh commands used with raw modules are changed and they assume that we have sh running/accessible on remote host.
On some devices that is not the case (switches, printers etc.)

This issue occurs with ansible 2.1 but not with ansible 2.0.*
##### STEPS TO REPRODUCE

```
  tasks:
    - raw: racadm getconfig -g idracinfo -o idRacType
```
##### EXPECTED RESULTS

This is what I get with ansible 2.0.0.1

```
EXEC sshpass -d15 ssh -C -vvv -o ControlMaster=auto -o ControlPersist=60s -o Port=22 -o User=abcd -o ConnectTimeout=60 -o 'ControlPath=~/.ssh/ansible-ssh-%h-%p-%r' -tt 1.2.3.4 'racadm getconfig -g idracinfo -o idRacType'
```
##### ACTUAL RESULTS

This is what I get with ansible 2.1

```
EXEC sshpass -d16 ssh -C -vvv -o ControlMaster=auto -o ControlPersist=60s -o Port=22 -o User=abcd -o ConnectTimeout=60 -o 'ControlPath=~/.ssh/ansible-ssh-%h-%p-%r' -tt 1.2.3.4 '/bin/sh -c '"'"'racadm getconfig -g idracinfo -o idRacType && sleep 0'"'"''
```

### LINKED ISSUE #16146: raw module not working for Cisco Routers
##### ISSUE TYPE

- Bug Report
##### ANSIBLE VERSION

```
ansible 2.1.0.0
```
##### CONFIGURATION

None relating to SSH.
##### OS / ENVIRONMENT

A fresh virtual env.
##### SUMMARY

The `raw` Module fails to connect via SSH under unknown circumstances. Executing the debug messages of Ansible directly works though. It works on routers with more modern SSH versions.
##### STEPS TO REPRODUCE

Execute command with the `raw` Module, even consider the `executable` advice from  #16054:

```
naegele@librenms0:~$ ansible all -i some-router.company.net, -m raw -a 'executable="" show int desc' -vvv
<some-router.company.net> ESTABLISH SSH CONNECTION FOR USER: None
<some-router.company.net> SSH: EXEC ssh -C -q -o ControlMaster=auto -o ControlPersist=60s -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o ConnectTimeout=10 -o ControlPath=/home/naegele/.ansible/cp/ansible-ssh-%h-%p-%r -tt some-router.company.net 'show int desc'
some-router.company.net | UNREACHABLE! => {
    "changed": false,
    "unreachable": true
}
```

Copying the command from above into the shell directly:

```
naegele@librenms0:~$ ssh -C -q -o ControlMaster=auto -o ControlPersist=60s -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o ConnectTimeout=10 -o ControlPath=/home/neuner/.ansible/cp/ansible-ssh-%h-%p-%r -tt some-router.company.net 'show int desc'


Mon Jun  6 12:55:17.543 CEST

Interface          Status      Protocol    Description
--------------------------------------------------------------------------------
Lo0                up          up          ***
<snip>
Te0/1/0/2          admin-down  admin-down
Te0/1/0/3          up          up          ***
```

##### EXPECTED RESULTS

It should work in either case.
##### ACTUAL RESULTS

It only works when working around ansible by pasting the command directly. However, both attemps look the same in the routers log:

```
RP/0/RSP0/CPU0:some-router#RP/0/RSP0/CPU0:Jun  6 13:43:30 CEST:
SSHD_[65888]: %SECURITY-SSHD-4-WARNING_LOGIN : The requested term-type
'screen' is not supported
RP/0/RSP0/CPU0:Jun  6 13:43:31 CEST: SSHD_[65888]:
%SECURITY-SSHD-6-INFO_SUCCESS_COMMAND : Successfully authenticated user
'naegele' from '2001::ffff' on 'vty1'(cipher 'aes128-cbc', mac
'hmac-md5', command 'show int desc')



RP/0/RSP0/CPU0:some-router#RP/0/RSP0/CPU0:Jun  6 13:43:40 CEST:
SSHD_[65888]: %SECURITY-SSHD-4-WARNING_LOGIN : The requested term-type
'screen' is not supported
RP/0/RSP0/CPU0:Jun  6 13:43:40 CEST: SSHD_[65888]:
%SECURITY-SSHD-6-INFO_SUCCESS_COMMAND : Successfully authenticated user
'naegele' from '2001::ffff' on 'vty1'(cipher 'aes128-cbc', mac
'hmac-md5', command 'show int desc')
```

Setting `TERM=vt100` clears the warning from the Cisco's log, but does not help otherwise.

Again, note that it works on a router with a more recent version of sshd.

bcoca:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

2.1
##### SUMMARY

fixes #16054
fixes #16146

also removed unused cruft in script

### DIFF
--- a/lib/ansible/playbook/play_context.py
+++ b/lib/ansible/playbook/play_context.py
@@ -29,11 +29,9 @@ import string
 from ansible.compat.six import iteritems, string_types
 from ansible import constants as C
 from ansible.errors import AnsibleError
-from ansible.playbook.attribute import Attribute, FieldAttribute
+from ansible.playbook.attribute import FieldAttribute
 from ansible.playbook.base import Base
-from ansible.template import Templar
 from ansible.utils.boolean import boolean
-from ansible.utils.unicode import to_unicode
 
 __all__ = ['PlayContext']
 
@@ -447,15 +445,20 @@ class PlayContext(Base):
         success_key = None
         self.prompt = None
 
-        if executable is None:
-            executable = self.executable
-
         if self.become:
+
+            if executable is None:
+                executable = self.executable
 
             becomecmd   = None
             randbits    = ''.join(random.choice(string.ascii_lowercase) for x in range(32))
             success_key = 'BECOME-SUCCESS-%s' % randbits
             success_cmd = pipes.quote('echo %s; %s' % (success_key, cmd))
+
+            if executable:
+                command = '%s -c %s' % (executable, success_cmd)
+            else:
+                command = success_cmd
 
             # set executable to use for the privilege escalation method, with various overrides
             exe = self.become_exe or \
@@ -485,9 +488,9 @@ class PlayContext(Base):
                 # force quick error if password is required but not supplied, should prevent sudo hangs.
                 if self.become_pass:
                     prompt = '[sudo via ansible, key=%s] password: ' % randbits
-                    becomecmd = '%s %s -p "%s" -u %s %s -c %s' % (exe,  flags.replace('-n',''), prompt, self.become_user, executable, success_cmd)
+                    becomecmd = '%s %s -p "%s" -u %s %s' % (exe,  flags.replace('-n',''), prompt, self.become_user, command)
                 else:
-                    becomecmd = '%s %s -u %s %s -c %s' % (exe, flags, self.become_user, executable, success_cmd)
+                    becomecmd = '%s %s -u %s %s' % (exe, flags, self.become_user, command)
 
 
             elif self.become_method == 'su':
@@ -498,7 +501,7 @@ class PlayContext(Base):
                     return bool(SU_PROMPT_LOCALIZATIONS_RE.match(data))
                 prompt = detect_su_prompt
 
-                becomecmd = '%s %s %s -c %s' % (exe, flags, self.become_user, pipes.quote('%s -c %s' % (executable, success_cmd)))
+                becomecmd = '%s %s %s -c %s' % (exe, flags, self.become_user, pipes.quote(command))
 
             elif self.become_method == 'pbrun':
 
@@ -534,7 +537,7 @@ class PlayContext(Base):
 
                 exe = self.become_exe or 'dzdo'
 
-                becomecmd = '%s -u %s %s -c %s' % (exe, self.become_user, executable, success_cmd)
+                becomecmd = '%s -u %s %s' % (exe, self.become_user, command)
 
             else:
                 raise AnsibleError("Privilege escalation method not found: %s" % self.become_method)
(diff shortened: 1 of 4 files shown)


mattclay:
Why the change in behavior for script? The script action in 2.0 added a shell.

bcoca:
script and raw have very similar purposes, we added the shell by mistake to ALL modules in 2.0 but those 2 should not have had it.

abadger:
Using empty string feels a bit fragile to me.  I'd be inclined to use False instead of empty string so that code would fail sooner.  It's something of a bikeshed, though.

Jake, Shane, @mattclay, and I also found a bug in the fix.  It's if we use sudo with raw.  Both empty string and False get sent to play_context.make_become_cmd().  In there, they are used in string formats.  If the string is format is False then "False" gets injected into the string.  If empty string is used then a "-c" which is a part of the format string ends up hanging out there without the shell command that it was expecting to be used with.  This one needs to be fixed (I think by changing the become output to only output executable + "-c" if executable is truthy.

abadger:
We should also update the docs in several places:
- for the raw and script modules to show that we do not have a shell there unless one is specified via executable.
- In the environment: task attribute docs, a warning box to say that raw and script do not honor environment.

abadger:
Looks good now.  @mattclay had an idea that we should either have a parameter to raw or have a second raw module so that we can do things like environment on boxes that have a shell but no python but still preserve the ability for raw to work when there's no shell installed on the remote system like this case.  I like that idea but it can be discussed separately from this fix.

+1 to merge.

mattclay:
The change to script prevents it from working with become. I don't think we should change the behavior of script now anyways, as it's unrelated to the issue with raw. The existing script behavior of adding a shell has been present since at least 1.9.6.

mattclay:
Using become with this modification results in a shell command like this being passed to sudo:

`'echo BECOME-SUCCESS...; LANG=... some_script.sh'`

Instead of this:

`/bin/sh -c 'echo BECOME-SUCCESS...; LANG=... some_script.sh'`

mattclay:
Both raw and script have the same problem with become, but at least with raw you can pass executable to correct the issue.

PR #16087, opened 2016-06-01
Give a module the possibility to known its own name

### LINKED ISSUE #15830: Modules don't know their own name
##### ISSUE TYPE

- Feature Idea
##### ANSIBLE VERSION

All
##### SUMMARY

In the old days when modules were run, they lived on disk and one way to get the module name was to simply run `os.path.basename(__file__)`. This is still used at various places.

However with pipelining the module name is no longer retrievable this way (it returns `<stdin>` instead), however various errors would be much better if it would explicitly mention the module name when throwing the error (e.g. remote module _cron_ does not support check mode).

Also logging now logs using logname "<stdin>" and you get syslog messages like:

```
Jan 01 00:00:00 system01 ansible-<stdin>: Invoked with parm1=foo parm2=bar
```

It would be quite easy to retain the module name so the ansible core in every context knows the module name, but also so that the module itself can report its own name in a consistent way.

dagwieers:
##### ISSUE TYPE

- Bugfix Pull Request
##### ANSIBLE VERSION

v2.1.0.0 and older
##### SUMMARY

This is useful for logging and reporting and fixes the longstanding problem with syslog-messages:

```
May 30 15:50:11 moria ansible-<stdin>: Invoked with ...
```

now becomes:

```
Jun  1 17:32:03 moria ansible-copy: Invoked with ...
```

This fixes #15830 and #16087

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -1253,7 +1253,7 @@ class AnsibleModule(object):
                 self.ansible_version = v
 
             elif k == '_ansible_module_name':
-                self.ansible_module_name = v
+                self._name = v
 
             elif check_invalid_arguments and k not in self._legal_inputs:
                 self.fail_json(msg="unsupported parameter for module: %s" % k)
@@ -1263,7 +1263,7 @@ class AnsibleModule(object):
                 del self.params[k]
 
         if self.check_mode and not self.supports_check_mode:
-                self.exit_json(skipped=True, msg="remote module (%s) does not support check mode" % self.ansible_module_name)
+                self.exit_json(skipped=True, msg="remote module (%s) does not support check mode" % self._name)
 
     def _count_terms(self, check):
         count = 0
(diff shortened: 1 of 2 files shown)


bcoca:
this seems a lot of work for `os.path.splitext(os.path.basename(__file__))[0]`

dagwieers:
Yes, because what you propose simply does not work. It returns either `<stdin>` when pipelining is enabled, or `basic.py` with the ziploader...

bcoca:
ah, forgot that ziploader changes it ... need to fix some custom modules then ...

dagwieers:
Yes, if this is merged I will updated the handful of modules.

But more importantly (for me) this fixes the "remote module does not support check mode" message as well as the syslog output. And I hope to improve some of the error messages to include the module name too.

sivel:
It looks like this PR has broken unit tests.  The CI for this PR shows that it is failing due to tests that were not updated to reflect the changes.

https://app.shippable.com/runs/575ade6db4d2260c0016fc7e

dagwieers:
Ouch, I'll see to it.

I wasn't expecting unit tests to create fake AnsibleModule objects (in a way they won't exist in the _real_ world). I thought Travis CI was acting up again...

PR #16089, opened 2016-06-01
Add junit callback plugin.

mattclay:
##### ISSUE TYPE

Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (junit-callback db645cb212) last updated 2016/06/01 10:47:34 (GMT -700)
  lib/ansible/modules/core: (detached HEAD ca4365b644) last updated 2016/06/01 10:36:50 (GMT -700)
  lib/ansible/modules/extras: (detached HEAD b0aec50b9a) last updated 2016/05/31 17:41:12 (GMT -700)
```
##### SUMMARY

Add junit callback plugin which records playbook output in junit xml format. This is useful for running playbooks as tests and collecting the results for analysis.

This is intended to be used with the upcoming integration of https://shippable.com/ for CI.

### DIFF
--- a/lib/ansible/plugins/callback/junit.py
+++ b/lib/ansible/plugins/callback/junit.py
@@ -45,7 +45,7 @@ class CallbackModule(CallbackBase):
 
     This plugin makes use of the following environment variables:
         JUNIT_OUTPUT_DIR (optional): Directory to write XML files to.
-                                     Default: Current working directory.
+                                     Default: ~/.ansible.log
 
     Requires:
         junit_xml
@@ -60,7 +60,7 @@ class CallbackModule(CallbackBase):
     def __init__(self):
         super(CallbackModule, self).__init__()
 
-        self._output_dir = os.getenv('JUNIT_OUTPUT_DIR', os.getcwd())
+        self._output_dir = os.getenv('JUNIT_OUTPUT_DIR', os.path.expanduser('~/.ansible.log'))
         self._playbook_path = None
         self._playbook_name = None
         self._play_name = None
@@ -72,6 +72,9 @@ class CallbackModule(CallbackBase):
             self.disabled = True
             self._display.warning('The `junit_xml` python module is not installed. '
                                   'Disabling the `junit` callback plugin.')
+
+        if not os.path.exists(self._output_dir):
+            os.mkdir(self._output_dir)
 
     def _start_task(self, task):
         """ record the start of a task for one or more hosts """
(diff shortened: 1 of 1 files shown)


lib/ansible/plugins/callback/junit.py
```diff
@@ -0,0 +1,237 @@
...
+
+        name = '[%s] %s: %s' % (host_data.name, task_data.play, task_data.name)
+        duration = host_data.finish - task_data.start
+
+        if host_data.status == 'included':
+            return TestCase(name, task_data.path, duration, host_data.result)
+
+        res = host_data.result._result
+        rc = res.get('rc', 0)
+        dump = self._dump_results(res, indent=0)
```
alikins, line 128: Aside from self._dump_results here, this could almost be a static method or module scope method.

I sort of want to suggest always just passing the entire host_data.result to TestCase. Would seem more useful for the intended use case  (viewing results of our test cases).  But, I could also see this being useful with real 'live' data 
and the results being sensitive  (something like someone using ansible via jenkins as part of a test setup, and getting the task results as jenkins would be handy). Maybe include an AbridgedTestCase(TestCase) that can do the _dump_results style cleanup itself?

Something that has a TestCase() [or perhaps as a classmethod alternative constructor] could also handle figuring out the duration, name, and add_*_info info.

Then this method is more or less

```
def _build_test_case(self, task_data, host_data):
    return AnsibleTestCase(task_data, host_data)
```

Similarly for TestSuite.
test_suite = AnsibleTestSuite(playbook_name, task_data)

(And maybe TaskData would get playbook_name passed to it's init?)

lib/ansible/plugins/callback/junit.py
```diff
@@ -0,0 +1,237 @@
...
+            test_case.add_skipped_info(message)
+
+        return test_case
+
+    def _generate_report(self):
+        """ generate a TestSuite report from the collected TaskData and HostData """
+
+        test_cases = []
+
+        for task_uuid, task_data in self._task_data.items():
```
alikins, line 159: task_uuid seems unused

lib/ansible/plugins/callback/junit.py
```diff
@@ -0,0 +1,237 @@
...
+            for host_uuid, host_data in task_data.host_data.items():
```
alikins, line 160: ditto host_uuid

(Or is need task_date.items() necessary with OrderedDict?)

alikins:
All my comments are all minor design stuff though and likely a touch overengineering, and nothing that would prevent this from being used as intended (or more importantly, not needed to start using this via shippable...)

The idea is good, and the code looks good to me. shipit or whatever we say now.

PR #16091, opened 2016-06-01
Allow names to include the dot character

chrisgilmerproj:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
$  ansible --version
ansible 2.0.2.0
  config file = 
```
##### SUMMARY

If I want to put the fqdn in the tag `Name` then the current script will convert it to have underscores. So `www.example.com` would become `www_example_com`, which is undesirable if I have a `.ssh/config` file already configured for `www.example.com`.  Instead of having to add additional entries I can now set `replace_dash_in_groups = False`.

```
ansible -i ec2.py dotted.tag.name -m ping
```

This change should be backwards compatible.

### DIFF
--- a/contrib/inventory/ec2.ini
+++ b/contrib/inventory/ec2.ini
@@ -110,6 +110,7 @@ cache_max_age = 300
 nested_groups = False
 
 # Replace - tags when creating groups to avoid issues with ansible
+replace_dot_in_groups = True
 replace_dash_in_groups = True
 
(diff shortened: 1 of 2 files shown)


chrisgilmerproj:
I'd like to get a review on this, anything I need to do to spruce it up?

chrisgilmerproj:
This has been open a long time with no comments so I'm closing it out.

PR #16124, opened 2016-06-03
Port urls.py to python3 and other byte vs text fixes

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

devel (2.2)
##### SUMMARY

Fixes (largely normalizing byte vs text strings) for python3

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -148,6 +148,8 @@ except ImportError:
         print('\n{"msg": "SyntaxError: probably due to installed simplejson being for a different python version", "failed": true}')
         sys.exit(1)
 
+from ansible.module_utils.six import PY2, PY3, b, binary_type, text_type, string_types
+
 HAVE_SELINUX=False
 try:
     import selinux
@@ -540,11 +542,11 @@ def _load_params():
                 fd.close()
             else:
                 buffer = sys.argv[1]
-                if sys.version_info >= (3,):
+                if PY3:
                     buffer = buffer.encode('utf-8', errors='surrogateescape')
         # default case, read from stdin
         else:
-            if sys.version_info < (3,):
+            if PY2:
                 buffer = sys.stdin.read()
             else:
                 buffer = sys.stdin.buffer.read()
@@ -557,7 +559,7 @@ def _load_params():
         print('\n{"msg": "Error: Module unable to decode valid JSON on stdin.  Unable to figure out what parameters were passed", "failed": true}')
         sys.exit(1)
 
-    if sys.version_info < (3,):
+    if PY2:
         params = json_dict_unicode_to_bytes(params)
 
     try:
@@ -1567,7 +1569,7 @@ class AnsibleModule(object):
                 # TODO: surrogateescape is a danger here on Py3
                 journal_msg = remove_values(msg, self.no_log_values)
 
-            if sys.version_info >= (3,):
+            if PY3:
                 syslog_msg = journal_msg
             else:
                 syslog_msg = journal_msg.encode('utf-8', 'replace')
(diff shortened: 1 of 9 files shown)


lib/ansible/module_utils/basic.py
```diff
@@ -2020,15 +2031,15 @@ def run_command(self, args, check_rc=False, close_fds=True, executable=None, dat
...
+        to_clean_args = args
+        if PY2:
+            if isinstance(args, text_type):
+                to_clean_args = args.encode('utf-8')
         else:
-            to_clean_args = args
+            if isinstance(args, binary_type):
+                to_clean_args = args.decode('utf-8', errors='replace')
+        if isinstance(args, (text_type, binary_type)):
+            to_clean_args = shlex.split(to_clean_args)
```
alikins, line 2042: This seems like it could be a util function. 

Could module_utils.splitter.split_args() be used here?

lib/ansible/module_utils/basic.py
```diff
@@ -2087,13 +2098,19 @@ def run_command(self, args, check_rc=False, close_fds=True, executable=None, dat
             # the communication logic here is essentially taken from that
             # of the _communicate() function in ssh.py
 
-            stdout = ''
-            stderr = ''
+            stdout = b('')
+            stderr = b('')
```
alikins, line 2102: [comment on code that was pre-existing to this pr and doesn't need to be revised now]

My preference would be to rename these to something like stdout_data/stderr_data. Having vars named 'stdout' and
'stderr' floating around code dealing with subprocess/popen/select seem error prone, especially since the 'stdout' var here is not a pipe or file object.

lib/ansible/playbook/base.py
```diff
@@ -112,6 +112,9 @@ def _generic_g(prop_name, self):
         if hasattr(self, method):
             return getattr(self, method)()
 
+        if prop_name not in self._attributes:
+            return None
+
```
alikins: I'd like to see a comment either here or on the commit explaining this. 

It seems like a general bug fix and unrelated to py2/py3. If it is related to py2/py3 I'd like to see an explanation.

lib/ansible/module_utils/basic.py
alikins, line 2102: This changes the return type of this method doesnt it? Do modules need to be changed?

lib/ansible/module_utils/urls.py
```diff
@@ -98,14 +98,24 @@
     # Python 3
     import http.client as httplib
 
+import ansible.module_utils.six.moves.urllib.request as six_request
+from ansible.module_utils.six.moves.urllib.request import (build_opener,
+        install_opener, urlopen, BaseHandler, HTTPBasicAuthHandler,
+        HTTPPasswordMgrWithDefaultRealm, HTTPRedirectHandler,
+        ProxyHandler, Request)
```
alikins: My preference here would be to just import request as six_request and remove the object imports.
Mostly because of:
- It's a lot of imports, and a lot of imports on a single line
- In this case, it would be useful to refer to this objects via module scope, since they are not (always) the regular urlib/urllib2 objects. Using the six_request module namespace makes that more clear.

But, thats a style thing, so does not require revision now.

lib/ansible/playbook/base.py
abadger: @mattclay ^  It seemed like you understood this thoroughly -- want to let me know what the comment should say?

lib/ansible/module_utils/basic.py
abadger, line 2102: It shouldn't change the return type.  Remember, this is python2 code being ported to run on python3 in addition.

lib/ansible/module_utils/basic.py
abadger, line 2042: splitter is much more complex than this code.  Since shlex has always worked here I'd rather not.  We should attempt to move as much as possible away from splitter rather than to it.

abadger, line 2042: Thinking about whether this should be a utility function... the problem I see is that if we attempt to hide the complexity here we come out with an unclean API.  If we don't hide the complexity inside of hte function, then the function really doesn't do anything.

What I mean by unclean API is that the function would say "Takes either text or utf-8 encoded byte strings on input.  On output gives text type on python3 and bytes type on python2".  The unrelated matrix elements make it unintuitive how the function behaves.  It would be better if the return type was linked to the input type or the input type had to be appropriate for the version of python (or if both values were constant).

lib/ansible/playbook/base.py
mattclay: @abadger The behavior of hasattr changed in Python 3.2 (http://stackoverflow.com/questions/903130/hasattr-vs-try-except-block-to-deal-with-non-existent-attributes/16186050#16186050)  so it only catches AttributeError.

We may want to solve this a different way, but without this an exception will be raised when prop_name doesn't exist under Python 3.

abadger:
Merged to devel.

Issue #16125, opened 2016-06-03
2.1 regression: status fields like "changed" can return non-boolean values


dmsimard:
##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
core

##### ANSIBLE VERSION

ansible 2.1.0.0
##### CONFIGURATION

Not relevant
##### OS / ENVIRONMENT

Not relevant
##### SUMMARY

An Ansible task can return a non-boolean value in status fields like "changed".
This violates the contract of these fields and breaks plugin implementations.

This was noticed and reproduced around fact gathering and hostvars.
##### STEPS TO REPRODUCE

```
- name: Reproducer
  hosts: localhost
  gather_facts: yes
  tasks:
    - name: Output hostvars to file
      copy:
        content: "{{ hostvars[inventory_hostname] }}"
        dest: "/tmp/ansible_hostvars.txt"
      ignore_errors: True
```
#2.1 with gather_facts: yes

```
PLAYBOOK: reproducer.yml *******************************************************
1 plays in playbooks/reproducer.yml

PLAY [Reproducer] **************************************************************

TASK [setup] *******************************************************************
ok: [localhost]

TASK [Output hostvars to file] *************************************************
task path: /home/dmsimard/dev/ansible/playbooks/reproducer.yml:5
changed: [localhost] => {"changed": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER", "checksum": "b9ccd3a********3********e********c********************************b9********************************9********e********acea********9936************************", "dest": "********tmp********ansible_********vars.txt", "gid": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER", "group": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER", "md5sum": "****************9b************************3cfc3f6********ebad********c********3********************************b********d", "mode": "********6****************", "owner": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER", "secontext": "unconfined_u:object_r:user_tmp_t:s********", "size": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER", "src": "****************.ansible********tmp********ansible-tmp-****************6********96********3****************.********-************************3************************************************3********************************source", "state": "file", "uid": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER"}

PLAY RECAP *********************************************************************
localhost                  : ok=2    changed=1    unreachable=0    failed=0 
```
#2.1 with gather_facts: no

```
[...]

PR #16127, opened 2016-06-03
Fix the copy action to not send the content as a param.

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
2.1.0 and devel
```
##### SUMMARY

content has already been transferred as a tempfile.

This fixes the error in https://github.com/ansible/ansible/issues/16125
but there may be higher level issues that should be fixed as well (other
modules might be able to cause status fields like failed and changed to
return a censored string instead of a bool).  So leaving 16125 open for
now.

### DIFF
--- a/lib/ansible/plugins/action/copy.py
+++ b/lib/ansible/plugins/action/copy.py
@@ -245,6 +245,8 @@ class ActionModule(ActionBase):
                         original_basename=source_rel,
                     )
                 )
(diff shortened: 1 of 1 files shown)


larsks:
:+1: 

That certainly solves the immediate problem.

abadger:
Excellent.  Merged to devel and stbale-2.1 trees.

PR #16129, opened 2016-06-03
docker_common - Fix image_lookup Tag Matching

amcrn:
**ISSUE TYPE**
- Bugfix Pull Request

**ANSIBLE VERSION**

```
ansible 2.1.0.0 (detached HEAD b599477242) last updated 2016/05/31 22:32:16 (GMT +000)
  lib/ansible/modules/core: (detached HEAD 04a871d007) last updated 2016/05/31 22:32:28 (GMT +000)
  lib/ansible/modules/extras: (detached HEAD b5fa2b3416) last updated 2016/05/31 22:32:34 (GMT +000)
  config file = /ansible/ansible.cfg
  configured module search path = ['/opt/ansible/ansible/library']
```

**SUMMARY**

`_image_lookup` incorrectly defaults the return value to a list of all
images matching a name (aka all tags). if the tag does not match any
of the images, the entire list is returned, resulting in the following
error:

```
Registry returned more than one result for...
```

example: if using docker_container with state=present and image=foo:3,
if the destination host has foo:2 and foo:1 on disk, then
[foo:1, foo:2] will be returned from `_image_lookup` and ansible will
present the error: "Registry returned more than one result for foo:3".
if an empty list was returned, then find_image would correctly return
None, resulting in the pull of the missing image.

example: if using docker_container with state=present and image=foo:3,
if the destination host has foo:1 on disk, then [foo:1] will be
returned from `_image_lookup` and ansible will inspect the _wrong_
image and return the inspection!

solution: if the tag is not found in the image list, return an empty
list as expected.

### DIFF
--- a/lib/ansible/module_utils/docker_common.py
+++ b/lib/ansible/module_utils/docker_common.py
@@ -430,6 +430,7 @@ class AnsibleDockerClient(Client):
             self.fail("Error searching for image %s - %s" % (name, str(exc)))
         images = response
         if tag: 
+            images = list()
             lookup = "%s:%s" % (name, tag)
             for image in response:
                 self.log(image, pretty_print=True)


amcrn:
cc @chouseknecht

amcrn:
@chouseknecht mind taking a look at this? without this fix, the docker_container module is unusable if prior versions of an image are present on a host. thanks!

cpatry-poly:
This fix work like a charm.
Thanks !

And I confirm, docker_container and docker_image_facts are useless without this.

amcrn:
looks like @chouseknecht wrote his own version a week or so ago: https://github.com/ansible/ansible/commit/16a3f28f6efa7ac3ab59970863bcc12434585350

Closing.

davidham:
Sorry if this is the wrong place to post this, but I'm running ansible installed via pip locally, and as part of Jenkins on my CI server. I ran `sudo pip install --upgrade ansible` locally, but I'm still getting the error, and I don't know how to update the Jenkins one. How do I get the `devel` fixes?

PR #16143, opened 2016-06-06
Filter tags

jordiclariana:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0.0
```
##### SUMMARY

 Add filter_tags option to filter tags after retrieve them.

This is complement to instance_filters which can only do OR when it have more than one rule.
This filter will be executed after retrieving all tags (instance_filters is send with the
AWS API call so the instances are filtered server-side). only if all tags specified exist
and the value of the tag match they will be included  in the final list.

### DIFF
--- a/contrib/inventory/ec2.ini
+++ b/contrib/inventory/ec2.ini
@@ -165,6 +165,13 @@ group_by_elasticache_replication_group = True
 # (ex. webservers15, webservers1a, webservers123 etc)
 # instance_filters = tag:Name=webservers1*
 
+# Filter instances exclusively by tags
+# Comma separated values that will be ANDed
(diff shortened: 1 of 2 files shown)


s-hertel (approved):

s-hertel:
Hi, @jordiclariana. We are not working on adding features to the inventory scripts since adding the inventory plugins (new in Ansible 2.4). I think this is something you can already do with the [aws_ec2 inventory plugin](https://docs.ansible.com/ansible/latest/plugins/inventory/aws_ec2.html) using the constructed option `keyed_groups`. Feel free to reach out if you have questions.

Issue #15143, opened 2016-03-24
Allow Ansible to read inventory created via bash process substitution


agaffney:
##### ISSUE TYPE
- ~Bug  Report~
- Feature Idea
##### COMPONENT NAME
core

##### ANSIBLE VERSION

```
ansible 2.0.1.0
  config file = /home/agaffney/.ansible.cfg
  configured module search path = Default w/o overrides
```
##### CONFIGURATION

N/A
##### OS / ENVIRONMENT

N/A
##### SUMMARY

When using bash process substitution to create an inventory dynamically, ansible throws an error claiming that it cannot read the file.
##### STEPS TO REPRODUCE

```
$ ansible all -i <(for i in 1 2; do echo "local_${i}"; done) -c local -m shell -a 'uptime'
ERROR! the file_name '/dev/fd/63' does not exist, or is not readable
```
##### EXPECTED RESULTS

I used a modified command below to achieve the same results.

```
$ ansible all -i $(for i in 1 2; do echo "local_${i}"; done | tr '\n' ',') -c local -m shell -a 'uptime'
local_1 | SUCCESS | rc=0 >>
 10:56:23 up 72 days, 20:55, 14 users,  load average: 0.17, 0.30, 0.38

local_2 | SUCCESS | rc=0 >>
 10:56:23 up 72 days, 20:55, 14 users,  load average: 0.17, 0.30, 0.38
```
##### ACTUAL RESULTS

```
$ ansible all -i <(for i in 1 2; do echo "local_${i}"; done) -c local -m shell -a 'uptime'
ERROR! the file_name '/dev/fd/63' does not exist, or is not readable
```

Additionally, strace shows that ansible is successfully opening and reading the file before throwing the error.

```
stat("/dev/fd/63", {st_mode=S_IFIFO|0600, st_size=0, ...}) = 0
stat("/dev/fd/63", {st_mode=S_IFIFO|0600, st_size=0, ...}) = 0
open("/dev/fd/63", O_RDONLY)            = 3
fstat(3, {st_mode=S_IFIFO|0600, st_size=0, ...}) = 0
fstat(3, {st_mode=S_IFIFO|0600, st_size=0, ...}) = 0
mmap(NULL, 4096, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f8e13282000
read(3, "local_1\nlocal_2\n", 8192)     = 16
read(3, "", 4096)                       = 0
[...]

PR #16155, opened 2016-06-06
Fix for bash process substitution usage in inventory/extra-vars

LarsFronius:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
~ $ ansible --version
ansible 2.2.0
```
##### SUMMARY

The bug occurs when using an inventory from a process substitution as
described in #15143 or when using extra-vars e.g. `-e @<(echo 'foo: "bar"')`
on ansible-playbook.

### DIFF
--- a/lib/ansible/parsing/dataloader.py
+++ b/lib/ansible/parsing/dataloader.py
@@ -125,7 +125,7 @@ class DataLoader():
 
     def is_file(self, path):
         path = self.path_dwim(path)
-        return os.path.isfile(to_bytes(path, errors='strict')) or path == os.devnull
+        return (os.path.exists(to_bytes(path, errors='strict')) and not os.path.isdir(to_bytes(path, errors='strict'))) or path == os.devnull
 
     def is_directory(self, path):
         path = self.path_dwim(path)


LarsFronius:
About the version - I did the bugfix based on working with the upstream git codebase.

alikins:
Minor revisions to commit message:
1) I like the summary, but a shorter first line in the commit message is recommended.
2) References bug fixes with 'Fixes #15143"

mattclay:
This fix appears incomplete. Running `ansible -m ping all -i <(echo 'localhost')` fails for me:

```
ERROR! Unexpected Exception: 'NoneType' object has no attribute 'startswith'
```

Without this fix it fails as originally noted:

```
ERROR! the file_name '/dev/fd/63' does not exist, or is not readable
the file_name '/dev/fd/63' does not exist, or is not readable
```

Also, it would be good to add an integration test to verify the fix and prevent regressions in the future.

bcoca:
don't change the is_file test it has deeper repercussions than this error, a better fix is not using ONLY this test.

LarsFronius:
Do we have any integration tests already testing how ansible works together with certain bash/unix/posix features like pipes etc. that I could check beforehand?
I have to admit I didn't try the inventory way and it seems it generally works different than the extra-vars use-case as one prefixes with an @ when a file is used and one not. So generally, there should be two different code paths imho?
Anyway, with a bit of guidance I think I can address your issues:
- where to add integration tests for specific posix/bash behaviour
- shall we split behaviour of is_file for inventory/extra-vars

mattclay:
I don't believe we have any existing integration tests that are appropriate, so you'll probably want to add another target to `test/integration/Makefile` to cover this case.

You should take a look at issue #15143 for my original notes on fixing this.

To deal with the is_file issue, rather than changing the implementation of is_file, you should change the if statement which calls it.

PR #16158, opened 2016-06-06
Network Module: EOS

Qalthos:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
2.2
```
##### SUMMARY

Introduces `module_utils/network.py` and modifies `module_utils/eos.py` to use it. Takes advantage of Ziploader to have reusable structure that network modules can share.


lib/ansible/module_utils/eos.py
```diff
@@ -17,78 +17,219 @@
...
+            if save_config:
+                module.config.save_config()
+
+        result['changed'] = True
+        result['updates'] = commands
+
+    return result
+
+def expand_intf_range(interfaces):
+    match = INTF_NAME_RE.match(interfaces)
```
Qalthos: @privateip, this function seems incomplete... `INTF_NAME_RE` is never defined, nor are `first_port` or `last_port` below. Sorry I didn't catch it earlier on the PR, but now that everything is finally together, this might be an issue.

I also notice it's not used anywhere presently, but that only turns it from an existing issue to an exciting future issue

privateip: check here for the regex fix https://github.com/Qalthos/ansible/pull/5/commits/89f2a106db4f7296822e118ad24f66cea84f3be9#diff-e72efaa6b72beb5339ad847e21a31220L87

Good catch on the first_port / last_port ... will add a fix for those.

This is a common pattern to be used by eos modules.    I have held off on pushing the modules upstream until we get the initial push of the refactored network shared modules

PR #16161, opened 2016-06-06
moved pending reboot to powershell.ps1

trondhindenes:
##### ISSUE TYPE
- Feature Pull Request
  ##### ANSIBLE VERSION
  2.2.0
##### SUMMARY

Discussed here: https://github.com/ansible/ansible-modules-core/pull/3795
Decided to move the code change into powershell.ps1, there will be another PR for setup.ps1 (modules-core) referencing this function to output pending reboot as a fact.

### DIFF
--- a/lib/ansible/module_utils/powershell.ps1
+++ b/lib/ansible/module_utils/powershell.ps1
@@ -225,3 +225,20 @@ Function Get-FileChecksum($path)
     }
     return $hash
 }
+
+Function Get-PendingRebootStatus
+{
+    # Check if reboot is required, if so notify CA. The MSFT_ServerManagerTasks provider is missing on client SKUs
+    #Function returns true if computer has a pending reboot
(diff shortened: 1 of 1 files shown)


trondhindenes:
ping @nitzmahone for review

nitzmahone:
Looks good- glad to see the `-EA Ignore` in there to keep things happy on workstation OSs...

lib/ansible/module_utils/powershell.ps1
```diff
@@ -225,3 +225,20 @@ Function Get-FileChecksum($path)
...
+    if(($featureData -and $featureData.RequiresReboot) -or $regData)
+    {
+        return $True
+    }
+    else 
+    {
+        return $False
+    }
+
+}
```
nitzmahone, line 244: @trondhindenes After a couple rounds of "WTH?!" with this on my Win10 box (it took 3 reboots to get to "false"- Chrome kept putting things in the PendingFileRenameOperations key), I'm thinking we might want an optional ref arg where we'll return a list of the reasons. I also have a couple more locations to look (win updates, CBS, etc), so with all those places, it'd probably be good to know why a reboot is pending.

PR #16172, opened 2016-06-07
univention: add common code for univention corporate server modules

keachi:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0
```
##### SUMMARY

This is common code for univention corporate server modules:
actually there are some PR on the ansible-modules-extras repository:
- [udm_dns_zone](https://github.com/ansible/ansible-modules-extras/pull/2382)
- [udm_dns_record](https://github.com/ansible/ansible-modules-extras/pull/2394)
- [udm_group](https://github.com/ansible/ansible-modules-extras/pull/2396)
- [udm_user](https://github.com/ansible/ansible-modules-extras/pull/2406)
- [udm_share](https://github.com/ansible/ansible-modules-extras/pull/2439)

Additional modules will follow with reference to this PR.


lib/ansible/module_utils/univention.py
```diff
@@ -0,0 +1,214 @@
+#!/usr/bin/python
```
bcoca: this file is an import, no shebang needed

lib/ansible/module_utils/univention.py
```diff
@@ -0,0 +1,214 @@
...
+
+"""univention corporate server access module"""
+
+try:
+    import univention.uldap
+    import univention.config_registry
+    import univention.admin.uldap
+    import univention.admin.objects
+    import univention.admin.config
+    import re
```
bcoca: not all these imports belong here, they should have specific messages if dependencies other than univention are missing

2-B:
We changed the whole import process, because the import should be executed on the target machine.
I removed the shebang.

keachi:
I removed the shebang, thanks for reviewing.
The whole import should be executed on the target machine, so we changed the import process.

resmo:
You used GPL, which is fine for module but for module_utils, a BSD style license is preferred . The reason is that closed source modules can be distributed using the code in module_utils.

keachi:
I/we prefer GPL, because we want, that other modules for univention are also open source.

resmo:
I understand, I am ok with it but would like to get the ok from ansible core. /cc @bcoca

resmo:
I brought this up to the Public Core Meeting (see log https://meetbot.fedoraproject.org/ansible-meeting/2016-08-09/ansible_core_public_meeting.2016-08-09-19.07.html)

I am afraid, Ansible insists on all items below module_utils/ not being GPL.

resmo:
As an option you could keep GPL but move your shared code to `ansible-modules-extras/univention/__init__.py` and adjust your include to it in your modules e.g. `from ansible.modules.extras.univention import ...`

resmo:
ignore my last comment, this won't work. Never mind.

keachi:
I got the rights to relicense to BSD-2-clause.

resmo:
kicking the build again

resmo:
LGTM

PR #16176, opened 2016-06-07
Fix distro detection for Solaris and derivatives

natefoo:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (solaris-facts 41ad001e61) last updated 2016/06/07 15:13:35 (GMT -400)
```

(devel)
##### SUMMARY

The excellent refactoring of distribution facts by @robinro in #15221 included a bug for Solaris and derivatives resulting in the loss of distribution detection in `ansible_distribution_*` and the changing of `ansible_os_family` from `Solaris` to `SunOS` for all distros (including Solaris proper).

In the code, Solaris was previously lumped in with Linux in running the list of release files. In the refactor, the "catchall" that included Linux was made explicit for Linux and Solaris was left out. Because Solaris detection only reads `/etc/release` and is already separated out from Linux by `platform.distribution()`, I added it to `systems_implemented` and made minor changes accordingly.

With current devel:

```
% ansible smartos -m setup | egrep 'distribution|os_family'
        "ansible_distribution": "SunOS", 
        "ansible_distribution_release": "5.11", 
        "ansible_distribution_version": "joyent_20160330T234717Z", 
        "ansible_os_family": "SunOS", 
```

With this PR:

```
% ansible smartos -m setup | egrep 'distribution|os_family'
        "ansible_distribution": "SmartOS", 
        "ansible_distribution_release": "SmartOS x86_64", 
        "ansible_distribution_version": "16.1.0", 
        "ansible_os_family": "Solaris", 
```

With 2.1.0.0:

```
% ansible smartos -m setup | egrep 'distribution|os_family'
        "ansible_distribution": "SmartOS", 
        "ansible_distribution_major_version": "NA", 
        "ansible_distribution_release": "SmartOS x86_64", 
        "ansible_distribution_version": "16.1.0", 
        "ansible_os_family": "Solaris", 
```

The loss of `ansible_distribution_major_version` should not really be a problem - it was always "NA" on Solaris (this was an artifact of falling through to the "default" detection mechanism with Linux).

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -812,7 +812,7 @@ class Distribution(object):
             self.facts['distribution_release'] = ora_prefix + data
             return
 
-        uname_rc, uname_out, uname_err = self.module.run_command(['uname', '-v'])
+        uname_v = get_uname_version(self.module)
         distribution_version = None
         if 'SmartOS' in data:
             self.facts['distribution'] = 'SmartOS'
@@ -825,7 +825,7 @@ class Distribution(object):
         elif 'OmniOS' in data:
             self.facts['distribution'] = 'OmniOS'
             distribution_version = data.split()[-1]
-        elif uname_rc == 0 and 'NexentaOS_' in uname_out:
+        elif uname_v is not None and 'NexentaOS_' in uname_v:
             self.facts['distribution'] = 'Nexenta'
             distribution_version = data.split()[-1].lstrip('v')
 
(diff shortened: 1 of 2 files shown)


robinro:
@natefoo Thanks for the PR and sorry for introducing the bug. To make sure we don't break it again in some future refactor (which I expect in the not too far future), please also provide a testcase. See 
[gen_distribution_version_testcase.py](https://github.com/ansible/ansible/blob/devel/hacking/tests/gen_distribution_version_testcase.py)

natefoo:
@robinro Unfortunately I don't have all of these various illumos distros to actually run the detection code on to generate the tests, but I can manually generate the `TESTSETS` entries based on the data I collected to do illumos distro detection in Ansible in the first place:

https://gist.github.com/natefoo/7af6f3d47bb008669467

natefoo:
Ok, I pushed the test cases. I did not commit the relatively simple script (and copies of the files in /etc) I used to generate it since I think it's a lot of clutter, but I can do that if necessary.

robinro:
Thanks for the testcases. They look reasonable to me. And the fix seems good as well.

I don't know much about Solaris though but due to the given test cases would still suggest to shipit.

alikins:
LGTM, rebased/squashed and merged to devel at https://github.com/ansible/ansible/commit/2275a08b9a51072fb53aca2bae370d6201dd5c58

PR #16182, opened 2016-06-08
Refresh TREE_DIR

andy-zhangtao:
If run ad-hoc in multiple process, the TREE_DIR will keep the same
value. So refresh this value in **init** function.

### DIFF
--- a/lib/ansible/plugins/callback/tree.py
+++ b/lib/ansible/plugins/callback/tree.py
@@ -23,7 +23,7 @@ import os
 from ansible.plugins.callback import CallbackBase
 from ansible.utils.path import makedirs_safe
 from ansible.utils.unicode import to_bytes
-from ansible.constants import TREE_DIR
+
 
 
 class CallbackModule(CallbackBase):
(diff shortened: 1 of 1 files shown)


bcoca:
can you show an example of the problem?

andy-zhangtao:
I need a webserver to run ad-hoc(like ansible tower). In my webserver, I have 5 or 6 processes to run. If I run a module (with --tree=xxx to gather logs), ansible will output all the log in the same folder except reboot the server.

So I modify this file to fix this issue.

bcoca:
Internal Ansible API
====================

Hi!

Thanks very much for your submission to Ansible.  It sincerely means a lot to us
to have an active community that is willing to submit feedback so that we may
continue to make Ansible better.

The internal Python API is considered an unsupported aspect of Ansible and this
is not considered a bug unless there is an issue with an Ansible binary
(`ansible`, `ansible-playbook`, `ansible-doc`, etc) or an issue with an external
API such as are provided for the development of plugins (modules, dynamic
inventories, callbacks, strategies, etc).

As such, we're going to close this ticket.  However, we're open to being corrected, should you wish to discuss.  You can stop by one of our two mailing lists or IRC to talk about this and we might be persuaded otherwise.

PR #16196, opened 2016-06-08
Drop unused imports from docker_common.py

stefanha:
##### ISSUE TYPE
- Bugfix Pull Request
##### SUMMARY

There are several unused imports which can be removed to keep the
imports lean and up-to-date.

Signed-off-by: Stefan Hajnoczi stefanha@gmail.com

### DIFF
--- a/lib/ansible/module_utils/docker_common.py
+++ b/lib/ansible/module_utils/docker_common.py
@@ -32,10 +32,9 @@ try:
     from requests.exceptions import SSLError
     from docker import Client
     from docker import __version__ as docker_version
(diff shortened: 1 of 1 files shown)


sivel:
This may not work.  Some of these imports are utilized by the actual modules, such as in docker_container, LogConfig comes from importing this docker_common

stefanha:
Thanks for explaining.  I realize my error now: I didn't init submodules so grepping didn't turn up the users in modules/core/cloud/docker/.  Closing since this patch is broken.

PR #16205, opened 2016-06-09
Remove a useless section, only act on 'shell'

dagwieers:
##### ISSUE TYPE

- Feature Pull Request
##### ANSIBLE VERSION

v2.1
##### SUMMARY

Simplify the logic, only act when the module is 'shell' as it is in fact 'command'.

### DIFF
--- a/lib/ansible/parsing/mod_args.py
+++ b/lib/ansible/parsing/mod_args.py
@@ -114,11 +114,6 @@ class ModuleArgsParser:
         to shell/command being treated special and nothing else
         '''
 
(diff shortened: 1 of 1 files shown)


dagwieers:
@nitzmahone This is such an easy, non-intrusive, cleanup. Should be a simple decision to merge or don't bother at all.

nitzmahone:
Yeah, that's fine- it's not like we've been sitting around mulling it over. Little stuff like this usually just gets merged during triage (when we look for "nasty" issues and add the tags, etc), but this one got lost in the shuffle.

Issue #16191, opened 2016-06-08
get_url module fails to download url: AttributeError: 'function' object has no attribute 'urlsplit'


jladdjr:
##### ISSUE TYPE
- Bug Report
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel 5f1bbb4fcd) last updated 2016/06/08 14:05:44 (GMT -400)
  lib/ansible/modules/core: (detached HEAD a8072f9ef0) last updated 2016/06/08 13:35:54 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 6ee1cfc5cf) last updated 2016/06/08 13:35:54 (GMT -400)
  config file = /Users/jladd/.ansible.cfg
  configured module search path = Default w/o overrides
```
##### CONFIGURATION

```
[ssh_connection]
control_path = %(directory)s/%%h-%%r
```
##### OS / ENVIRONMENT

N/A
##### SUMMARY

In the `stable-2.1` and `stable-1.9` versions of ansible, I am able to download `http://www.google.com` using `ansible -vvv localhost -i inventory -m get_url -a 'url=http://www.google.com dest=/tmp'`. On the devel branch, though, I'm seeing the following error: 

`AttributeError: 'function' object has no attribute 'urlsplit'`
##### STEPS TO REPRODUCE
1. Download the Ansible devel repo. Run `git submodule update --init --recursive` in the repo, then `source hacking/env-setup`.
2. Create new file, `inventory`, with contents `localhost ansible_connection=local`.
3. Run `ansible -vvvv localhost -i inventory -m get_url -a 'url=http://www.google.com dest=/tmp'`
##### EXPECTED RESULTS

On ansible 2.1.0.0, command is successful ([results](https://gist.github.com/jladdjr/23c535228a61150303ef5b6cde6b0a72)). Confirmed that google homepage appears in /tmp/index.html.

On ansible 1.9.6, command is successful ([results](https://gist.github.com/jladdjr/7e3d00013e33fd5823e7692f516bbbce)). Confirmed that google homepage appears in /tmp/index.html.
##### ACTUAL RESULTS

On `devel` branch, seeing:

```
ansible -vvvv localhost -i inventory -m get_url -a 'url=http://www.google.com dest=/tmp'
Using /Users/jladd/.ansible.cfg as config file
Loaded callback minimal of type stdout, v2.0
Using module file /Users/jladd/ansible/devel/lib/ansible/modules/core/network/basics/get_url.py
[...]

PR #16212, opened 2016-06-09
Add a get_url test for www.google.com

jctanner:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
$ ansible --version
ansible 2.2.0 (TEST_GETURL_GOOGLE d9448980a1) last updated 2016/06/09 14:24:23 (GMT -400)
  lib/ansible/modules/core: (devel 4a3916bd46) last updated 2016/06/09 13:27:16 (GMT -400)
  lib/ansible/modules/extras: (devel 93b59ba852) last updated 2016/06/09 13:27:21 (GMT -400)
  config file =
```
##### SUMMARY

Addresses #16191

### DIFF
--- a/test/integration/roles/test_get_url/tasks/main.yml
+++ b/test/integration/roles/test_get_url/tasks/main.yml
@@ -205,3 +205,8 @@
     that:
       - "result.changed == true"
       - "stat_result.stat.mode == '0070'"
+
+- name: Test url split with no filename #https://github.com/ansible/ansible/issues/16191
+  get_url:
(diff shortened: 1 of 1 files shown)


mattclay:
That test doesn't fail without the fix. Try putting the module args on one line, like this:

`get_url: url=http://www.google.com/ dest=/tmp`

Be careful what you use for dest, as some values prevent it from failing without the fix.

Also, I'd put the comment on it's own line, instead of at the end of the name line.

PR #16228, opened 2016-06-10
fixed dzdo command to work like sudo and properly detect a prompt by …

rybal06:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0
```
##### SUMMARY

```
Command being ran: 
ansible -i hosts all -m "ping"  -bK --become-user=root --become-method=dzdo -u username --ask-pass

Before changes:
Timeout (12s) waiting for privilege escalation prompt: 

After changes: 
hostname.domain.com | SUCCESS => {
    "changed": false,
    "ping": "pong"
}
```

…setting the -p option to a known string

 Committer: Ryan Ballanger  rybal06@gmail.com

### DIFF
--- a/lib/ansible/playbook/play_context.py
+++ b/lib/ansible/playbook/play_context.py
@@ -531,10 +531,12 @@ class PlayContext(Base):
                 becomecmd = '%s %s echo %s && %s %s env ANSIBLE=true %s' % (exe, flags, success_key, exe, flags, cmd)
 
             elif self.become_method == 'dzdo':
-
-                exe = self.become_exe or 'dzdo'
-
-                becomecmd = '%s -u %s %s -c %s' % (exe, self.become_user, executable, success_cmd)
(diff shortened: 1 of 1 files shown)


bcoca:
earlier versions of dzdo did not have -p option, we might need to handle this based on version

rybal06:
I cannot say for certain when that was the case, but I have confirmed that version 4.3.0-170 includes dzdo and supports the -p option.  4.3 is a release from 2009.  I also checked and 4.1 did not have dzdo included.

It is possible that 4.2 included dzdo without the -p option, but anything 4.3.0-170 (2009) and newer should support using the -p option.

rybal06:
This seems to have been fixed in the latest dev master branch. It looks like this PR solved the issue. https://github.com/ansible/ansible/pull/17840

PR #16234, opened 2016-06-10
Make destination_format_tags accept all attrs.

awong-dev:
##### ISSUE TYPE

- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.0.2.0
  config file = 
```
##### SUMMARY

The ec2.py script has 3 variables used to define the output of the dynamic host list.  The `destination_variable` and `vpc_destination_variable` draw from the same domain of values
whereas `destination_format_tags` draws from a subdomain of those values.  This restricted
the utility of the `destination_format` unnecessarily.

This change extracts the variable resolution logic into a function and calls it for all three variables to unify the domains.

### DIFF
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -680,17 +680,19 @@ class Ec2Inventory(object):
         if instance.state not in self.ec2_instance_states:
             return
 
+        def get_attr_or_tag(instance, tag, default):
+            dest = getattr(instance, tag, default)
+            if dest is None:
+                dest = getattr(instance, 'tags').get(tag, default)
+            return dest
+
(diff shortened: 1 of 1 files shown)


lathama:
@awong-dev is this still of interest to you? If not please close.

s-hertel:
Hi @awong-dev. You can modify the `ansible_host` (or any hostvar) flexibly using the constructed inventory plugin options. We are not working on adding features to the inventory scripts since adding the inventory plugins (new in Ansible 2.4). This looks like something you can do with the [aws_ec2](https://docs.ansible.com/ansible/latest/plugins/inventory/aws_ec2.html).

PR #16240, opened 2016-06-11
Make sure we don't end up with an empty PYTHONPATH

### LINKED ISSUE #16195: Running python scripts with command or shell modules ends up adding the cwd into sys.path
##### ISSUE TYPE

- Bug Report
##### ANSIBLE VERSION

```
ansible 2.1.0.0
```
##### CONFIGURATION

Issue was observed when the `-i` flag was used with `sudo_flags`, but only because this caused the current working directory of the remote script to become the user's home directory, which was dirty and contained random files like `subprocess.py`
##### OS / ENVIRONMENT

rhel6.6
rhel7.2
##### SUMMARY

When running a python script remotely via ansible's _command_ or _shell_ modules, the script's current working directory gets added to its `sys.path`. If the current working directory contained python files that name-conflicted with modules the script was trying to import, a crash inevitably happens. In our case, a rogue, out-of-date `subprocess.py` located in the running user's home directory caused the script to crash upon attempting to `import multiprocessing`.

We tracked the issue to commit https://github.com/ansible/ansible/commit/16c1f10e18f6b4701c5114c51c8f7f6f0ceb96d4, where it looks like ansible is removing paths from `$PYTHONPATH` that was previously added by ziploader. If this step results in an empty list of paths, `$PYTHONPATH` will still exist in `os.environ` but will be set to an empty string.

If python encounters `$PYTHONPATH` set in the environment but with an empty value, it will add an empty string to `sys.path`, and as a result the running python script will look in its current working directory for modules to import. Note that when `site.py` is invoked, the paths will be canonicalized, so it will show within the script as a full path to the current working directory instead of merely `''`, unless python is invoked with the `-Ss` flags. This behavior is true for at least python 2.7.5.

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel a42ddf981f) last updated 2016/06/11 13:15:25 (GMT -700)
  lib/ansible/modules/core: (devel d6f01d0b4f) last updated 2016/06/11 13:14:11 (GMT -700)
  lib/ansible/modules/extras: (devel 43760b2c4a) last updated 2016/06/11 13:14:22 (GMT -700)
```
##### SUMMARY

When the PYTHONPATH is an empty string python will treat it as though
the cwd is in the PYTHONPATH.  This can be undesirable.  So make sure we
delete PYTHONPATH from the environment altgether in this case.

Fixes #16195

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -2031,6 +2031,8 @@ class AnsibleModule(object):
                         if not x.endswith('/ansible_modlib.zip') \
                         and not x.endswith('/debug_dir')]
             os.environ['PYTHONPATH'] = ':'.join(pypaths)
+            if not os.environ['PYTHONPATH']:
+                del os.environ['PYTHONPATH']
 
         # create a printable version of the command for use
         # in reporting later, which strips out things like


abadger:
We're getting ready to release 2.1.1-rc1 so I'm going ahead and merging this to devel and stable-2.1 branches.  @zhaojack if you do test this, please let us know in a new ticket if it does not fix your issue.  Thanks!

zhaojack:
@abadger Sorry for not getting back to you sooner. I won't have cycles to revisit this for another few days. The fix looks good though, and I'll file a new ticket if it turns out to not fix our issue. Thank you very much!

PR #16247, opened 2016-06-13
Bugfix: (re)allow ansible_python_interpreter to contain more than 1 arg

lpirl:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (ansible_python_interpreter-with-args 8fb75fb7e1) last updated 2016/06/13 09:53:28 (GMT +200)
  lib/ansible/modules/core: (detached HEAD 138430f116) last updated 2016/06/13 09:50:39 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD 883ccbefe5) last updated 2016/06/13 09:50:46 (GMT +200)
  config file = /home/lukas/.ansible.cfg
```
##### SUMMARY

Allows `ansible_python_interpreter` to contain args again (e.g. `nice -n10 python2 -OO`).
This used to work and stopped working somewhere between 2.0 and 2.1.
This patch uses `shlex` to split the value of `ansible_python_interpreter` accordingly.

The output is unchanged, except that Ansible does fail upon the execution of the first module.

### DIFF
--- a/lib/ansible/executor/module_common.py
+++ b/lib/ansible/executor/module_common.py
@@ -639,15 +639,10 @@ def _find_snippet_imports(module_name, module_data, module_path, module_args, ta
         if shebang is None:
             shebang = u'#!/usr/bin/python'
 
-        executable = interpreter.split(u' ', 1)
-        if len(executable) == 2 and executable[0].endswith(u'env'):
-            # Handle /usr/bin/env python style interpreter settings
-            interpreter = u"'{0}', '{1}'".format(*executable)
-        else:
-            # Still have to enclose the parts of the interpreter in quotes
-            # because we're substituting it into the template as a python
-            # string
-            interpreter = u"'{0}'".format(interpreter)
+        # Enclose the parts of the interpreter in quotes because we're
+        # substituting it into the template as a Python string
+        interpreter_parts = interpreter.split(u' ')
+        interpreter = u"'{0}'".format(u"', '".join(interpreter_parts))
 
         output.write(to_bytes(ACTIVE_ZIPLOADER_TEMPLATE % dict(
             zipdata=zipdata,


abadger:
Changes to handle more than a command and single argument should be done here: https://github.com/ansible/ansible/blob/devel/lib/ansible/executor/module_common.py#L631

rather than inside of the Ansiballz wrapper script.

Notes:
- What operating system is your target?  http://www.in-ulm.de/~mascheck/various/shebang/#results seems to show there aren't many OS's where multiple arguments will work correctly.
- If you move from split to shlex, you'll have to convert to byte strings if on python2 and use text strings for python3.  I believe that continuing to use split will work fine as ansible_python_interpreter is also used in the shebang lines and the shebang won't handle args with whitespace or shell quotes.  You'll just have to be looser about how many args split() is allowing.

abadger:
Also, feel free to ping me on irc (irc.freenode.net, #ansible-devel abadger1999) if you update this and I don't seem to have noticed.

abadger:
@lpirl do you need any other information to update this?

lpirl:
Maybe – thanks for asking. I'll come back to you if needed when I find time to take care of this.

lpirl:
> Changes to handle more than a command and single argument should be done here: https://github.com/ansible/ansible/blob/devel/lib/ansible/executor/module_common.py#L631
> rather than inside of the Ansiballz wrapper script.

[your link to the file at the time of your comment.](https://github.com/ansible/ansible/blob/5a3493be5f1f9e6652464b927bbccd01ce98a53a/lib/ansible/executor/module_common.py#L631)

Thanks for the pointer – I'll move my changes there.

> What operating system is your target? http://www.in-ulm.de/~mascheck/various/shebang/#results seems to show there aren't many OS's where multiple arguments will work correctly.

Debian, where it seems to work fine:

```
$ cat ./shebang.py
#!/usr/bin/python -O
print __debug__
$ ./shebang.py
False
```

Same on FreeBSD 8.4 and Cygwin 6.1. However, it is correct that users of Ansible should be aware of this and configure it accordingly – but it should be supported.

> If you move from split to shlex, you'll have to convert to byte strings if on python2 and use text strings for python3. I believe that continuing to use split will work fine as ansible_python_interpreter is also used in the shebang lines and the shebang won't handle args with whitespace or shell quotes. You'll just have to be looser about how many args split() is allowing.

I see, fair enough. :)

lpirl:
@abadger something like e1dc4a8?

abadger:
@lpirl Thanks!  Testing now.  Regarding how you tested shebangs... Your test case was only a single argument.  Try something like this:

```
$ cat ./shebang.py
#!/usr/bin/python -tt -O
print __debug__
$ ./shebang.py
Unknown option: - 
usage: /usr/bin/python [option] ... [-c cmd | -m mod | file | -] [arg] ...
Try `python -h' for more information.
```

lpirl:
Thanks @abadger for the information. IMHO, however, the `ansible_python_interpreter` setting is very specific to the targeted system anyway, so that administrators should be aware of this (i.e. responsible for choosing an compatible setting).

abadger:
Thanks @lpirl!  I've merged this to devel and tmp-holding-branch-post-2.1.1.  That means it will be in 2.2.0 and also in 2.1.2 when we release it (not in 2.1.1 which is in [hopefully] the final release candidate stage now).

lpirl:
Nice, thanks.

PR #16257, opened 2016-06-13
Add support for merging lists when merging hashes

notnownikki:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0
  config file = 
```
##### SUMMARY

Problem: a 'common' role might do things such as creating users
or installing ssh keys. Group level vars might specify a list
of users to create on the target systems. When doing test
deployments to vagrant or openstack, it's common to supply
extra_vars which have a list of users to create on all instances
for access purposes. Because of how the merging currently
works, the extra_vars override those lists and the users
in them are not created.

Solution: allow the user to specify a list of variables which
are lists to be merged along with the hashes.

Example:

ANSIBLE_MERGE_LISTS=common.users

Would result in the following two lists:

```
common:
  users:
    - admin

common:
  users:
    - service_user
```

combining into

```
common:
  users:
    - admin
    - service_user
```

but all other lists would keep the original overwriting behaviour.


raben2:
+1

calfonso:
Hi!

There is another PR under consideration to do this https://github.com/ansible/ansible/pull/25438/ however, we aren't convinced white listing lists is the route to go.

notnownikki:
Hey no problem! I did this as a way of learning Ansible internals, and figured I'd see if it was actually wanted or not.

PR #16270, opened 2016-06-14
Fix string coercion problem in error handling code                   …

art-at-sky:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.0.1.0
  config file =
```
##### SUMMARY

Bug was introduced in commit b122374

Manifests as the following stack trace

```
  File "/usr/local/Cellar/ansible/2.0.1.0/libexec/bin/ansible-playbook", line 86, in <module>
    sys.exit(cli.run())
  File "/usr/local/Cellar/ansible/2.0.1.0/libexec/lib/python2.7/site-packages/ansible/cli/playbook.py", line 150, in run
    results = pbex.run()
  File "/usr/local/Cellar/ansible/2.0.1.0/libexec/lib/python2.7/site-packages/ansible/executor/playbook_executor.py", line 140, in run
  File "/usr/local/Cellar/ansible/2.0.1.0/libexec/lib/python2.7/site-packages/ansible/executor/task_queue_manager.py", line 238, in run
  File "/usr/local/Cellar/ansible/2.0.1.0/libexec/lib/python2.7/site-packages/ansible/plugins/strategy/linear.py", line 334, in run
    display.error(e, wrap_text=False)
  File "/usr/local/Cellar/ansible/2.0.1.0/libexec/lib/python2.7/site-packages/ansible/utils/display.py", line 259, in error
    new_msg = u"ERROR! " + msg
TypeError: coercing to Unicode: need string or buffer, AnsibleParserError found
```

### DIFF
--- a/lib/ansible/utils/display.py
+++ b/lib/ansible/utils/display.py
@@ -278,7 +278,7 @@ class Display:
             wrapped = textwrap.wrap(new_msg, self.columns)
             new_msg = u"\n".join(wrapped) + u"\n"
         else:
-            new_msg = u"ERROR! " + msg
+            new_msg = u"ERROR! %s" % msg
         if new_msg not in self._errors:
             self.display(new_msg, color=C.COLOR_ERROR, stderr=True)
             self._errors[new_msg] = 1


abadger:
This looks like an okay change but we also should find what's sending a
byte string msg to the function and fix it to convert into text (py2
Unicode type) as early as possible.  We're working on making the controller
python 3 compatible and mixing bytes with text will still fail on python 3.

art-at-sky:
I know very little Python, but it seems to be passing an AnsibleParserError object rather than a byte string

abadger:
@art-at-sky Can you post the full traceback?  Then I can track down where the AnsibleParserError is being passed in.

art-at-sky:
OK I've updated the issue with the full trace

abadger:
Thanks @art-at-sky.  I've taken a look at the traceback and the source of the error is fixed in the git repo:
- devel a7f89cac63d22c2b33f7ce7c6f0846c47e77d8a8
- stable-2.1: a7f89cac63d22c2b33f7ce7c6f0846c47e77d8a8

I'll go ahead and merge this to devel as well since it isn't a bad thing to do.

Issue #13287, opened 2015-11-24
ansible inline vault


h0nIg:
Sometimes you want to define a complex variable dictionary and do not want to create a separate vault file for passwords. It would be cool for history purpose to use a plain yml file for variable definition and inline vault variable content like the following. This feature is described by @jpmens on http://jpmens.net/2014/02/22/my-thoughts-on-ansible-s-vault/

```
nameserver: 10.0.0.1
admin: Jane Jolie
dbpassword: @vault@"AHchx0a+G8mejs84tGxCNKxMFP7tM7Y7kl"
webservertype: nginx
```

bcoca:
closing as dupe of #12087

h0nIg:
this is not a duplicate, since the other guy wants to get this implemented by filters and my intention is that this is handled by the ansible yaml parser itself, please reopen!

bcoca:
i doubt that will happen as the performance penalty of examining every possible value for vault content will slow down Ansible considerably.

h0nIg:
therefore the initial idea was to scan only the first letters and check for a special content like @vault@

bcoca:
yes, that is what will be prohibitive for most users, only those with tiny inventories and few variables would not notice a huge delay.

We have been doing a lot of performance optimization prior to the 2.0 release, most speedups came from avoiding copying and examining the variables as much as we did.

h0nIg:
hmm ok.

but to check this statement: do you really think another loop on VariableManager.get_vars for replacing variable content is expensive too? (in comparision to copying on combine_vars). just scrolled the source code and it seems that it would work

bcoca:
try it out, but look at the commit history and you'll see we removed lots of code that scanned and/or copied the variables. What you see now should be the minimal needed for ansible's functionality and it is still slower than what we would want when large inventories are invovled

bcoca:
@h0nIg i do think your idea makes for a better user interface, though no need to add @vault@ as vault already puts in a specific recognizable header.

If you can implement w/o it impacting performance too much, go ahead, i suspect it will be hard to do but will be very happy if you prove me wrong.

bcoca:
i've been thinking about this more, a much more efficient way to do this would be to use a YAML type handler instead of parsing the values.

```
key: !vault vaultedvalue
```

would just require that we create a function that handles vaulted strings and register it with the module parser as the one to handle the `!vault` type.

h0nIg:
it seems this is pretty easy, first implementation (not tested)

[...]

PR #16274, opened 2016-06-14
embedded vault (encrypted yaml values) 

alikins:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

2.2. devel
##### SUMMARY

Part of 2.2 roadmap vault changes "Add ‘per variable’ vault support (!vault YAML directive, existing PR already)"

This is based on https://github.com/h0nIg/ansible/commit/e680f504aa0f6826ff47263d79dfb19d8a74b0e7 mentioned in the thread for https://github.com/ansible/ansible/issues/13287


sivel:
We had talked about the potential of marking the decrypted data as having come from a vault, and effectively set it as `no_log`.  I'm not sure that is a requirement, but would be a super nice to have.

Something like that may require us to create some form of class that we wrap the data in, similar to Unsafe, so that it is easy to tag that data.

Just some thoughts and ideas.

bcoca:
yes, make it something similar to 'unsafe' and always mark as no_log

h0nIg:
nice to see that my code is used as a base :)

the current code looks like the solution, no need to decrypt to a complex type like a map/list. but it may be necessary to include the vault version, to prevent problems with the encryption on future releases

alikins:
I think the version as of a11742ab0a14d90ada6f94898df27c36e3769388 seems to work for me, if folks wanted to test.

No docs yet, so the best example is probably the new integration test test/integration/test_embedded_vault.yml

Not integrated with 'ansible-vault edit' yet, so getting the vault blobs into a playbook is a cut and paste affair. And a caveat there... most editors will add a new line to the plaintext on save, which the normal 'ansible-vault encrypt' will include and may not be what you mean if you try to encrypt something like

``` yaml
vars:
      my_password: 'hunter42'
```

If you aren't careful with the plaintext file, it will get the newline and end up with the value with a newline like:

``` yaml
vars:
     my_password: 'hunter42\n'
```

alikins:
Ready for review.

I'll leave the commits unsquashed for now, but can start squashing them into logical chunks if needed.

lib/ansible/parsing/vault/__init__.py
```diff
@@ -473,6 +545,7 @@ def _editor_shell_command(self, filename):
 
         return editor
 
+# TODO: does anything use this?
 class VaultFile(object):
```
bcoca: no, this was a 'prototype' i was working on, feel free to remove

bcoca:
so most of the changes in the PR seem to be reformating/renaming of vars and code, I'm not against the changes as they clarify what a lot of the vars are, but I would like to keep these changes separate from the actual feature. It is hard to review the feature it self when it is lost in a sea of unrelated changes.

PR #16275, opened 2016-06-14
Unprivileged become on HP-UX/UNIX (Fixes #16249)

### LINKED ISSUE #16249: Unprivileged become on HP-UX/UNIX does not work
##### ISSUE TYPE
- Bug Report
##### ANSIBLE VERSION

```
ansible 2.1.0.0
```
##### CONFIGURATION

remote_user = ansible
##### OS / ENVIRONMENT

Ansible Host: SUSE Linux Enterprise Server 12 SP1
Managed Node: HP-UX B.11.31 U ia64
##### SUMMARY

Executing a play task with `become`on HP-UX fails with following error message:

```
[WARNING]: Using world-readable permissions for temporary files Ansible needs to create when becoming an unprivileged user which may be insecure. For information on securing this, see
https://docs.ansible.com/ansible/become.html#becoming-an-unprivileged-user

fatal: [host]: FAILED! => {"failed": true, "msg": "Failed to set file mode on remote files (rc: 1, err: chmod: can't traverse /tmp/ansible-tmp-1465816244.81-217081167531834/\n)"
```

This is due to the fact that a `chown` on HP-UX (probably all UNIX systems) behaves different as on Linux:
- Linux: Executing `chown user2 file` as unprivileged `user1` will fail
- HP-UX: Executing `chown user2 file` as unprivileged `user1` will **not** fail

In https://github.com/ansible/ansible/blob/2bf2e0fdfdb7293e28e3bb0543b2d0d14a65169f/lib/ansible/plugins/action/__init__.py#L296 you can see that ansible expects that a `chown` will fail in case it is executed by a unprivileged user.
##### STEPS TO REPRODUCE

Example Playbook

```
- name: Privilege Escalation on HP-UX
  hosts: all

  tasks:
    - name: test
      command: whoami
      become_user: user1
```
##### POSSIBLE SOLUTION
1. Try `_remote_set_user_facl` first
2. If it fails: try `_remote_chmod`
3. If it fails: raise AnsibeError
4. If it works: try `_remote_chown`
   ...

jokurz:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0.0
```
##### SUMMARY

Problem: When setting the file permissions on the remote server for
unprivileged users ansible expects that a chown will fail for unprivileged
users. For some systems (e.g. HP-UX) this is not the case.

Solution: Change the order how ansible sets the remote permissions.
- If the remote_user sudo's to an unprivileged user then we attempt to
  grant the unprivileged user access via file system acls.
- If granting file system acls fails we try to change the owner of the
  file with chown which only works in case the remote_user is privileged
  or the remote systems allows chown calls by unprivileged users (e.g.
  HP-UX)
- If the chown fails we can set the file to be world readable so that
  the second unprivileged user can read the file. Since this could allow
  other users to get access to private information we only do this
  ansible is configured with "allow_world_readable_tmpfiles" in the
  ansible.cfg

### DIFF
--- a/lib/ansible/plugins/action/__init__.py
+++ b/lib/ansible/plugins/action/__init__.py
@@ -301,17 +301,17 @@ class ActionBase(with_metaclass(ABCMeta, object)):
         information.  We achieve this in one of these ways:
 
         * If no sudo is performed or the remote_user is sudo'ing to
-          themselves, we don't have to change permisions.
+          themselves, we don't have to change permissions.
         * If the remote_user sudo's to a privileged user (for instance, root),
           we don't have to change permissions
-        * If the remote_user is a privileged user and sudo's to an
-          unprivileged user then we change the owner of the file to the
-          unprivileged user so they can read it.
-        * If the remote_user is an unprivieged user and we're sudo'ing to
-          a second unprivileged user then we attempt to grant the second
-          unprivileged user access via file system acls.
-        * If granting file system acls fails we can set the file to be world
-          readable so that the second unprivileged user can read the file.
+        * If the remote_user sudo's to an unprivileged user then we attempt to
+          grant the unprivileged user access via file system acls.
+        * If granting file system acls fails we try to change the owner of the
+          file with chown which only works in case the remote_user is
+          privileged or the remote systems allows chown calls by unprivileged
+          users (e.g. HP-UX)
+        * If the chown fails we can set the file to be world readable so that
+          the second unprivileged user can read the file.
           Since this could allow other users to get access to private
           information we only do this ansible is configured with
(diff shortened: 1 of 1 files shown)


abadger:
Merged this to devel.  Currently debating whether to push this into stable-2.1 (and whether to do that before or after 2.1.1 goes out).  It's a bug that we want to fix but it touches code that is used by pretty much any ansible task so we're hesitant to put it in 2.1.1 without more real-world testing.

abadger:
Decided that we'll put this into stable-2.1 after 2.1.1 is out the door so that it can get more real-world testing.

PR #16277, opened 2016-06-14
Copy the config functions from ansible into rax.py inventory script

jctanner:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
2.2
```
##### SUMMARY

Remove the imports from ansible and copy the functions directly to the script.

### DIFF
--- a/contrib/inventory/rax.py
+++ b/contrib/inventory/rax.py
@@ -155,8 +155,6 @@ import ConfigParser
 
 from six import iteritems
 
-from ansible.constants import get_config, mk_boolean
-
 try:
     import json
 except ImportError:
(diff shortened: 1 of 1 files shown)


jctanner:
I think this patch is going down the wrong path in terms of code deduplication and how we eventually want to refactor inventory scripts, but i'll leave that to the review process.

sivel:
I assume this is related to tower, and since I don't have access to that issue, it might be helpful to know what problem this is attempting to solve.

abadger:
Talked with @matburt -- for the next tower release, they'll probably use this patch on their vendored dynamic inventory scripts.  We won't include it in mainline ansible.

@matburt, note, the following dynamic inventory scripts will also not work unless you modify them as well (open_url is used because of lack of security with the stdlib's urllib2... be aware that we've had to update open_url when new means of bypassing certificate checking have been discovered in the past or to enhance what sites it can handle (for instance, SNI support in python  < 2.7.9)):

```
$ grep -ri ansible . |grep import              *[devel]  (08:32:48)
./openshift.py:from ansible.module_utils.urls import open_url
./rax.py:from ansible.constants import get_config, mk_boolean
./collins.py:from ansible.module_utils.urls import open_url
./abiquo.py:from ansible.module_utils.urls import open_url
./proxmox.py:from ansible.module_utils.urls import open_url
```

cc: @simfarm

matburt:
This has to do with the way Tower invokes the inventory script.   The normal method is to provide it as a parameter to Ansible.   We instead call it directly and import it into Tower's own inventory... in this case Ansible itself isn't available for various reasons, so importing it fails.

jctanner:
Not merging into devel.

PR #16281, opened 2016-06-14
Extract a censor_args method to module_utils

alikins:
##### ISSUE TYPE

- Feature Pull Request
- Bugfix Pull Request
##### SUMMARY

Some refactoring of module_utils.basic.py and add some tests.

This was existing code, just moved out of the
innards of AnsibleModule to a module level
method. Mostly for easier testing.

Add test cases for censor_args.

Split module_utils. import tests into new test module:
test_basic.py is now test_basic.py and test_basic_imports.py

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -86,6 +86,7 @@ import tempfile
 import traceback
 import grp
 import pwd
+import pipes
 import platform
 import errno
 import datetime
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -651,6 +651,45 @@ def human_to_bytes(number, default_unit=None, isbits=False):
     return int(round(num * limit))
 
 
+def censor_args(args, no_log_values):
+    """create a printable version of the command for use
+    in reporting later, which strips out things like
+    passwords from the args list"""
+
+    to_clean_args = args
+
+    if PY2:
+        if isinstance(args, text_type):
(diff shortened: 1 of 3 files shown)


abadger:
Perhaps we're ready to have a separate file in module_utils for things related to removing private data from strings?  I hate to add another toplevel function to basic.py since we'll need to carry backwards compat for that around.  (OTOH, since the code for this lived in basic before, we aren't really adding more code to basic... So if you think that we are not ready to break these functions into their own python module I'm not firmly against doing that now.)

abadger:
We can also clean up this code by switch .encode() and .decode() to use module_utils._text.to_bytes, to_native, and to_text instead.

bcoca:
i'm thinking `module_utils/basic/__init__.py` and other files might be the 'final' division of these

abadger:
`basic/__init__.py` with the other pieces that make it up in the basic subdirectory won't work if we want both backwards compatibility and to save people from having to download all of the things that make up basic.py now.  That's because for `from ansible.module_utils.basic import XXXX` to continue to import the various pieces that it does now the file `basic/__init__.py` would need to itself import all the pieces that were in the present basic.py.  For people to use an individual piece that lives in (for example) `basic/clean_strings.py` the code would also have to load `basic/__init__.py` which in turn, needs all of those pieces.

It's better to break these pieces up into one or modules with a different name than "basic" and then keep basic.py with an increasing amount of its code implemented by code in different python modules.

alikins:
meh

PR #16285, opened 2016-06-14
Lastpass lookup plugin

azenk:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (lpass_lookup 106ec7db0b) last updated 2016/06/16 19:35:29 (GMT -500)
  lib/ansible/modules/core: (detached HEAD d4eaeac8c9) last updated 2016/06/16 14:39:31 (GMT -500)
  lib/ansible/modules/extras: (detached HEAD eed9d601b5) last updated 2016/06/16 14:39:32 (GMT -500)
  config file = 
```
##### SUMMARY

This plugin can be used with the lpass cli interface for lastpass.
[lastpass-cli](https://github.com/lastpass/lastpass-cli)

Example:
Add a lookup to your playbooks/variables somewhere:

```
some_variable: "{{ lookup('lpass','Some Lastpass entry name or ID', field='username') }}"
```

Start a lpass session prior to using ansible, run ansible, logout when finished:

```
lpass login user@domain.com
ansible-playbook foo.yml
lpass logout
```

### DIFF
--- a/lib/ansible/plugins/lookup/lastpass.py
+++ b/lib/ansible/plugins/lookup/lastpass.py
@@ -67,7 +67,7 @@ class LPass(object):
 class LookupModule(LookupBase):
 
     def run(self, terms, variables=None, **kwargs):
-        lp = kwargs.get('__mock_lpass_obj', LPass())
+        lp = LPass()
 
         if not lp.logged_in:
             raise AnsibleError("Not logged into lastpass: please run 'lpass login' first")
--- a/lib/ansible/plugins/lookup/lastpass.py
+++ b/lib/ansible/plugins/lookup/lastpass.py
@@ -1,0 +1,79 @@
+# (c) 2016, Andrew Zenk <azenk@umn.edu>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+from subprocess import Popen, PIPE
+
+from ansible.errors import AnsibleError
+from ansible.plugins.lookup import LookupBase
+
+
+class LPassException(AnsibleError):
+    pass
+
+
+class LPass(object):
+
+    def __init__(self, path='lpass'):
+        self._cli_path = path
+
+    @property
+    def cli_path(self):
+        return self._cli_path
+
+    @property
+    def logged_in(self):
+        out, err = self._run(self._build_args("logout"), stdin="n\n", expected_rc=1)
+        return err.startswith("Are you sure you would like to log out?")
+
+    def _run(self, args, stdin=None, expected_rc=0):
+        p = Popen([self.cli_path] + args, stdout=PIPE, stderr=PIPE, stdin=PIPE)
+        out, err = p.communicate(stdin)
+        rc = p.wait()
+        if rc != expected_rc:
+            raise LPassException(err)
+        return out, err
+
+    def _build_args(self, command, args=None):
+        if args is None:
+            args = []
+        args = [command] + args
+        args += ["--color=never"]
+        return args
+
+    def get_field(self, key, field):
+        if field in ['username', 'password', 'url', 'notes', 'id', 'name']:
+            out, err = self._run(self._build_args("show", ["--{0}".format(field), key]))
+        else:
+            out, err = self._run(self._build_args("show", ["--field={0}".format(field), key]))
+        return out.strip()
+
+
+class LookupModule(LookupBase):
+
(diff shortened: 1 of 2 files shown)


jhawkesworth:
Interesting idea. I wonder if it would be possible/uaeful to implement as a callback plugin.

azenk:
I'm not quite sure what the potential benefits to implementing this as a callback plugin rather than a lookup plugin would be.  Can you expand on the thought?

jhawkesworth:
Probably not many, on reflection, it was just an idle though.  

I created a callback plugin to acquire kerberos credentials for passwordless connection to windows hosts and just thought maybe the same thing could be done for lastpass logins.  But I guess the difference is you might well have lots of lastpass logins for different hosts/services, whereas on windows you typically have a domain login that lets you do whatever you are allowed to do on lots of machines.

bcoca:
a lookup makes more sense as you can do: 

```
ansible_password: "{{lookup('lastpass' ....
```

the kerberos case is 'special' as it was not really a password but a ticket/token you maintain for the whole communication AFTER having authenticated

jhawkesworth:
Yeah, lookup makes sense. Callbacks really meant for reporting anyway, so it's bending things to use for Kerberos tickets. Forget that I commented!

azenk:
While reading the comment from @bcoca it occurred to me that lastpass would be a better name for this plugin, so I'll make that change and clean up some white space issues.

azenk:
I added some tests and refactored things a little.  It looks like all the tests are passing now.  I think it's ready to go.  I'm happy to make further improvements if requested though.

DebbieGillespie:
We would really like this feature. Could this pull request be merged?

sivel:
My 2 cents here.  I would much rather see this use the lastpass API, potentially through something like `lastpass-python` or even just direct API calls instead of shelling out to lastpass-cli.

azenk:
Are there docs available on that?  I looked for API docs prior to writing this and did not find any mention of anything other than the provisioning API.  This plugin has been working fine for us.  One nice benefit of using the cli, is that the lastpass vault session management is handled outside of ansible.   

Provided the session management could be dealt with properly (not using ansible vault to store my lastpass credentials) I'd be happy to modify this to use the API, if you can point me towards docs.

azenk:
As far as I can tell, lastpass-python cannot be used without providing your username and password at runtime, which would make the plugin useless for me.  So, I won't be using that.  Apart from shelling out to the cli, are there any other objections to merging this?

georgepsarakis:
@azenk there is a function for executing shell commands named [run_cmd](https://github.com/ansible/ansible/blob/devel/lib/ansible/utils/cmd_functions.py#L27).

Perhaps it would be preferrable to use this function, instead of using `subprocess.Popen` directly?

georgepsarakis:
@azenk regarding the API, if I understand correctly from the code, both the CLI and the
`lastpass-python` library make certain HTTP requests that return XML.

I think it should be also noted that `lastpass-python` is not officially supported whereas the CLI is.

azenk:
@georgepsarakis I could swap run_cmd in, but I'm not sure I see an advantage other than maintaining consistency with other ansible code.

georgepsarakis:
@azenk yes that is the main reason. Reusing well-tested code in large projects is beneficial, don't you think?

azenk:
@georgepsarakis Yes, absolutely.  In this case it's slightly problematic since I'd have to build an escaped command string just to have shlex.split return the same list of args that I started with (is there a good way to do this in python 2.x?  3.3 has shlex.quote).  While this is possible, I think the advantages of skipping all that and using Popen directly outweigh the negatives.  It's certainly doable though.

azenk:
What's the status of this?  I'm not planning to make any revisions to this unless someone offers concrete feedback about changes that are required prior to merging.

azenk:
Is there any chance this could be merged for 2.3?

Dispader:
Very excited to have this new feature— anyone have any more thoughts, here?

lib/ansible/plugins/lookup/lastpass.py
```diff
@@ -0,0 +1,79 @@
...
+            out, err = self._run(self._build_args("show", ["--{0}".format(field), key]))
+        else:
+            out, err = self._run(self._build_args("show", ["--field={0}".format(field), key]))
+        return out.strip()
+
+
+class LookupModule(LookupBase):
+
+    def run(self, terms, variables=None, **kwargs):
+        lp = kwargs.get('__mock_lpass_obj', LPass())
```
mattclay: Instead of making the plugin aware of the tests, use `mock.patch` to mock the `LPass` class in the tests.

mattclay (changes requested):
@azenk Looks good overall. Just one change to how unit tests are implemented should be made.

azenk:
@mattclay I think I've addressed the issue.  I'm a bit of a novice with mock.patch, so please forgive any rookie mistakes, etc.

test/units/plugins/lookup/test_lastpass.py
```diff
@@ -26,6 +26,8 @@
 from argparse import ArgumentParser
 import six
 
+from mock import patch
```
mattclay: Sorry, I should have mentioned this in my original feedback. Import `patch` using:

```python
from ansible.compat.tests.mock import patch
```

mattclay (changes requested):
Could you amend your last commit message to remove the @ mention of me. That should help cut down on spurious notifications on GitHub.

Other than that, and the one requested change, it looks good. Thanks for the quick response to feedback!

azenk:
@mattclay  So amended.  I also rolled the import change into the amended commit.

mattclay (approved):

mattclay:
@azenk Thanks for the new lookup plugin!

PR #16286, opened 2016-06-14
Adding instance_states to gce inventory to align with other inventories.

kwoodson:


kwoodson:
The idea behind this pull request is to be able to filter instances in gce based on instance_states.  I modeled these states after the current ec2.py and ec2.ini instance_states.  The default behavior surrounding ec2.py is to only return instances in the RUNNING state.  This pull request now returns states in RUNNING only.  If you want all_instances or specific instance_states, you would specify them in the gce.ini file.

`[gce]
instance_states = RUNNING,STOPPING`

or if you would prefer to get them all back

`[gce]
all_instances = True`

Ansible relies on instances to be up and running for the majority of management and orchestration tasks.  I believe making these providers behave similarly would make things simpler.

erjohnso:
Thanks for this @kwoodson!

I like the idea of more common ground across inventory scripts.  But this doesn't seem to be backwards compatible and could unintentionally surprise users when their 'stopped' (`TERMINATED`) instances are filtered out by default.

What do you think about a more general approach across all inventory scripts that could be done as part of a larger ansible release?  For instance, I'd think you could cover all of this in a single directive in the ini file,

```
[gce]
  # Filter inventory based on on state. Leave undefined to return instances regardless of state.
  # example: Uncomment to only return inventory in the running or provisioning state
  #instance_states = RUNNING,PROVISIONING
```

/cc @supertom, @ryansb

kwoodson:
@erjohnso,

Thanks for the reply.  
- Updated the code to default to all instances for backwards compatibility.
- Removed the need for all_instances flag as you mentioned.
- instance_states introduced into gce.ini
- Tested with RUNNING and TERMINATED states and appears to be working.

I'd love to have some default behavior for all inventories but as long as they have the options we need that should satisfy our needs.  Thanks.

kwoodson:
Added comments in gce.ini as per @erjohnso's comment.

supertom:
@kwoodson thanks for the contribution.  Chatted with @ryansb today and we agree this is a neat feature to have.  To @erjohnso 's point, could probably be larger change, but I think we're OK with piloting it here.
1. read_settings(self): Could we roll it into get_config?
2. default value of instance_states: rather than having pre-populated list, could we make self.instance_states default to empty (show all), and then have users populate the ini with the states they would like?  This would maintain backwards compatibility, keep us from maintaining a separate states list and allow for new states to be added without requiring a change for common users.

kwoodson:
@supertom,

I rolled these changes into get_config and defaulted the self.instance_states to an empty list.  If its empty, all instances are returned.  If its defined, then only instance states in the list will be returned.

contrib/inventory/gce.py
```diff
@@ -283,6 +295,11 @@ def group_instances(self):
         meta["hostvars"] = {}
 
         for node in self.driver.list_nodes():
+
+            # check desired states from config file
```
supertom: Maybe a more detailed comment about what it's doing here?  The ultimate behavior is what the user expects, but the code to implement might not be.  Worth explaining that we only check for the state in the list IF the list is actually populated.

contrib/inventory/gce.py
```diff
@@ -159,6 +159,18 @@ def get_config(self):
...
+
+        #########
+        # Section added for processing ini settings
+        #########
+
+        # filter on instance states
+        self.instance_states = []
+
+        if config.has_option('gce', 'instance_states'):
+            for inst_state in config.get('gce', 'instance_states').split(','):
```
supertom: split already returns a list, so do we need do the append?  If you split a string without a comma, it would just return the (single) item in a list, so I think you can avoid the loop.

supertom:
Hey @kwoodson, thanks for this.  Left you a couple of commas, but after that, I'm good.

kwoodson:
@supertom,

Thanks for the suggestions.
- Removed the for-loop as split returns a list.
- Added documentation about the instance_states if statement in an attempt to clarify code
- Added another check on the config value returning ''

The last check added was discovered during my testing.  If I left instance_states empty in the config file it returned an empty string.  This empty string was then split into a list with a single element of an empty string ['']. This then ignored all instances because instance states never matched an empty string.  This was undesired.  I now check if the instance_states is set before splitting and skip if its empty.

Tested with RUNNING only, TERMINATED only, RUNNING and TERMINATED, no setting in ini file, and empty setting in ini file.  Results as expected.

ryansb:
Looks good to me - tested on my GCP account.

PR #16287, opened 2016-06-14
try/except xmltodict import, misc cleanups

alikins:
##### ISSUE TYPE
- Bugfix Pull Request
##### SUMMARY

The sphinx api docs branch would fail to import this module. This fixes imports and a few style things.

The 'import xmltodict' was causing import
errors when generating documentation. Since
xmltodict is a required but not stdlib module,
throw AnsibleError if unable to import.

Add a faux-constant HALF_MB to represent the 2**19
magic number used as a buffer size.

Remove unused combine_vars.

Replace a use of 'stdin_iterator == None' with
idiomatic 'stdin_iterat is None'

Misc pep8 cleanups.

### DIFF
--- a/lib/ansible/plugins/connection/winrm.py
+++ b/lib/ansible/plugins/connection/winrm.py
@@ -25,17 +25,22 @@ import re
 import shlex
 import traceback
 import json
-import xmltodict
+
 
 from ansible.compat.six.moves.urllib.parse import urlunsplit
-
 from ansible.errors import AnsibleError, AnsibleConnectionFailure
+
 try:
     import winrm
     from winrm import Response
     from winrm.protocol import Protocol
 except ImportError:
     raise AnsibleError("winrm is not installed")
+
+try:
+    import xmltodict
+except ImportError:
+    raise AnsibleError("xmltodict is not installed")
 
 HAVE_KERBEROS = False
 try:
@@ -49,14 +54,12 @@ from ansible.plugins.connection import ConnectionBase
 from ansible.utils.hashing import secure_hash
 from ansible.utils.path import makedirs_safe
 from ansible.utils.unicode import to_bytes, to_unicode, to_str
-from ansible.utils.vars import combine_vars
 
 try:
     from __main__ import display
 except ImportError:
     from ansible.utils.display import Display
     display = Display()
-
 
 class Connection(ConnectionBase):
     '''WinRM connections over HTTP/HTTPS.'''
@@ -148,7 +151,7 @@ class Connection(ConnectionBase):
 
                 # open the shell from connect so we know we're able to talk to the server
                 if not self.shell_id:
-                    self.shell_id = protocol.open_shell(codepage=65001) # UTF-8
+                    self.shell_id = protocol.open_shell(codepage=65001)  # UTF-8
                     display.vvvvv('WINRM OPEN SHELL: %s' % self.shell_id, host=self._winrm_host)
 
                 return protocol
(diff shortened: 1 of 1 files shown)


lib/ansible/plugins/connection/winrm.py
```diff
@@ -25,18 +25,23 @@
...
     from winrm import Response
     from winrm.protocol import Protocol
 except ImportError:
     raise AnsibleError("winrm is not installed")
 
+try:
+    import xmltodict
+except ImportError:
+    raise AnsibleError("winrm is not installed")
+
```
abadger, line 44: Misleading error message?  Should it say "xmltodict is not installed.  Needed to use winrm connection type"?

lib/ansible/plugins/connection/winrm.py
```diff
@@ -49,14 +54,14 @@
...
 from ansible.utils.unicode import to_bytes, to_unicode, to_str
-from ansible.utils.vars import combine_vars
 
 try:
     from __main__ import display
 except ImportError:
     from ansible.utils.display import Display
     display = Display()
 
+HALF_MB = 2**19  # 0.5MB chunks
```
abadger: Better to name this after what it is used for.  BUFSIZE is used in other connection plugins.

abadger:
@nitzmahone assigned you to see if these cleanups seem okay.

lib/ansible/plugins/connection/winrm.py
alikins, line 44: doh, indeed.

alikins, line 44: Now that I think about it, I definitely cargo cult'ed that import error idiom from the existing code. It doesn't seem to be common in the code base. 

There are a couple of places with similar requirements (try to import something absolutely required, but for an optional plugin...). Is there a preferred way to handle those cases? 

The try/except and AnsibleError seems reasonable (at least with the correct error message ;->). But there are a few approaches in the code:

plugins/cache/redis.py  also raises an AnsibleError
plugins/cache/memcached.py prints() an error and calls sys.exit()   (eek!)

[module_utils/basic.py does variations of sys.exit, but that's a special case...]

lib/ansible/plugins/connection/winrm.py
alikins: I think I may just undo that change and revisit. There are two different hardcoded buffersizes in that module

``` python

      # FUTURE: determine buffer size at runtime via remote winrm config?
      def _put_file_stdin_iterator(self, in_path, out_path, buffer_size=250000):
          in_size = os.path.getsize(to_bytes(in_path, errors='strict'))
          offset = 0
```

and the one changed in this patch

``` python
    def fetch_file(self, in_path, out_path):
          super(Connection, self).fetch_file(in_path, out_path)
          in_path = self._shell._unquote(in_path)
          out_path = out_path.replace('\\', '/')
          display.vvv('FETCH "%s" TO "%s"' % (in_path, out_path), host=self._winrm_host)
          buffer_size = HALF_MB
```

Maybe 

``` python
FETCHFILE_BUFSIZE = 2**19  # in bytes, 0.5MB
PUTFILE_BUFSIZE = 250000
```

(PUTFILE_STDIN_BUFSIZE? )

lib/ansible/plugins/connection/winrm.py
nitzmahone, line 44: Since we don't actually need those present and functional to build docs, could just set HAS_PYWINRM/HAS_XMLTODICT and check them in CP init instead of raising immediately (though I suspect we may run into similar dependency issues in other plugins)

alikins:
Fixed the error message and removed the HALF_MB change.

PR #16289, opened 2016-06-14
zone plugin updates for newer api

alikins:
##### ISSUE TYPE
- Bugfix Pull Request
##### SUMMARY

This module would fail to import when sphinx-apidoc finds it. Updated to use more recent api.

Fixes errors on import.


abadger:
merged to devel ans stable-2.1 branches.  Thanks alikins!

PR #16294, opened 2016-06-15
Output stderr_lines

pgporada:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0
```
##### SUMMARY

Stdout_lines exists and shows pretty output. Stderr did not previously have the same feature. It makes it easier to see certain command output from the shell.

Before

```
ok: [test] => {
    "msg": {
        "changed": true, 
        "cmd": "apachectl -t", 
        "delta": "0:00:00.041603", 
        "end": "2016-06-15 00:59:42.204139", 
        "rc": 0, 
        "start": "2016-06-15 00:59:42.162536", 
        "stderr": "AH00112: Warning: DocumentRoot [/var/www/phil/public_html] does not exist\nAH00558: httpd: Could not reliably determine the server's fully qualified domain name, using 127.0.0.1. Set the 'ServerName' directive globally to suppress this message\nSyntax OK", 
        "stdout": "", 
        "stdout_lines": [], 
        "warnings": []
    }
}
```

After

```
ok: [test] => {
    "msg": {
        "changed": true, 
        "cmd": "apachectl -t", 
        "delta": "0:00:00.039094", 
        "end": "2016-06-15 00:35:41.612449", 
        "rc": 0, 
        "start": "2016-06-15 00:35:41.573355", 
        "stderr": AH00112: Warning: DocumentRoot [/var/www/phil/public_html] does not exist\nAH00558: httpd: Could not reliably determine the server's fully qualified domain name, using 127.0.0.1. Set the 'ServerName' directive globally to suppress this message\nSyntax OK", 
        "stderr_lines": [
            "AH00112: Warning: DocumentRoot [/var/www/phil/public_html] does not exist", 
            "AH00558: httpd: Could not reliably determine the server's fully qualified domain name, using 127.0.0.1. Set the 'ServerName' directive globally to suppress this message", 
            "Syntax OK"
        ], 
        "stdout": "", 
        "stdout_lines": [], 
        "warnings": []
    }
}
```

### DIFF
--- a/lib/ansible/plugins/action/__init__.py
+++ b/lib/ansible/plugins/action/__init__.py
@@ -663,10 +663,12 @@ class ActionBase(with_metaclass(ABCMeta, object)):
         # parse the main result
         data = self._parse_returned_data(res)
 
-        # pre-split stdout into lines, if stdout is in the data and there
-        # isn't already a stdout_lines value there
+        # pre-split stdout and stderr into lines
         if 'stdout' in data and 'stdout_lines' not in data:
             data['stdout_lines'] = data.get('stdout', u'').splitlines()
+
+        if 'stderr' in data and 'stderr_lines' not in data:
+            data['stderr_lines'] = data.get('stderr', u'').splitlines()
 
         display.debug("done with _execute_module (%s, %s)" % (module_name, module_args))
         return data
@@ -747,7 +749,7 @@ class ActionBase(with_metaclass(ABCMeta, object)):
         out = self._strip_success_message(out)
 
         display.debug("_low_level_execute_command() done: rc=%d, stdout=%s, stderr=%s" % (rc, stdout, stderr))
-        return dict(rc=rc, stdout=out, stdout_lines=out.splitlines(), stderr=err)
+        return dict(rc=rc, stdout=out, stdout_lines=out.splitlines(), stderr=err, stderr_lines=err.splitlines())
 
     def _get_first_available_file(self, faf, of=None, searchdir='files'):
 


nitzmahone:
@pgporada Like the change, happy to merge, but causing test failures at the moment.

lib/ansible/plugins/action/__init__.py
```diff
@@ -747,7 +749,7 @@ def _low_level_execute_command(self, cmd, sudoable=True, in_data=None, executabl
         out = self._strip_success_message(out)
 
         display.debug("_low_level_execute_command() done: rc=%d, stdout=%s, stderr=%s" % (rc, stdout, stderr))
-        return dict(rc=rc, stdout=out, stdout_lines=out.splitlines(), stderr=err)
+        return dict(rc=rc, stdout=out, stdout_lines=out.splitlines(), stderr=err, stderr_lines=err.splitlines())
```
nitzmahone, line 752: We aren't guaranteed that the stderr coming back from a connection plugin is any particular type- in this case, it's coming back from a test as a bool. Recommend type-checking and supplying a single-element repr list if it's not a type that supports splitlines()

pporada-gl, line 752: @nitzmahone Thanks for the comment. This is the first actual python I've written. I'll have to do some research and figure out what your comment means.

nitzmahone, line 752: Sorry, that was a bit "dense". :) The test that's failing is because the stderr object that comes back is not guaranteed to be something that has splitlines() defined (in this test, for whatever reason, it comes back as a bool). It isn't for stdout either, but all the connection plugins have been normalized to fix that (otherwise this would've blown up before on stdout). 

In Python it's actually cheaper to just call splitlines() and catch the exception if it's not there- I'd suggest setting the stderr_lines value to `[ str(stderr) ]` in that case.

tedder:
I think this was done as part of the async work:
https://github.com/ansible/ansible/commit/c86a17b7a003ac22c173cfa832ed2368e15015e6#diff-69a29e19a76e98587550dab380483594R740

calfonso:
Hi!

We believe recent commits should resolve this question or problem for you. Please re-test this with a current version of Ansible, thanks!

Thank you!

Issue #16100, opened 2016-06-02
Digital ocean inventory script does not correctly group servers


adamscybot:
##### ISSUE TYPE

Bug Report
##### ANSIBLE VERSION

N/A
##### CONFIGURATION

N/A
##### OS / ENVIRONMENT

Mac OSx
##### SUMMARY

The digital ocean inventory script only ever returns a maximum of one server per group.
##### STEPS TO REPRODUCE
- Grab https://github.com/ansible/ansible/blob/devel/contrib/inventory/digital_ocean.py
- Grab and setup https://github.com/ansible/ansible/blob/devel/contrib/inventory/digital_ocean.ini
- Run `python digital_ocean.py  -p`
##### EXPECTED RESULTS

A list of groups with their applicable hosts is shown in json format.
##### ACTUAL RESULTS

Each group only ever has a maximum of one host in it. E.g. I have 20 servers in the lon1 digital ocean region, but the script only outputs

```
 "region_lon1": {
    "hosts": [
      "**ip was here**"
    ], 
    "vars": {}
  }, 
```

bcoca:
@adamscybot please test the patch above

adamscybot:
@bcoca I'm getting:

```
Traceback (most recent call last):
  File "digital_ocean.py", line 471, in <module>
    DigitalOceanInventory()
  File "digital_ocean.py", line 230, in __init__
    self.build_inventory()
  File "digital_ocean.py", line 384, in build_inventory
    self.inventory[group].add(dest)
TypeError: unhashable type: 'dict'
```

Probably unrelated side point: I have to use dopy 0.3.6 and not the latest 0.3.7 as the latest causes this script to complain dopy is not installed.

bcoca:
updated to avoid that issue, i did not realize we were appending dictionaries to the groups instead of strings.

the other issue seems unrelated, do you have multiple python versions installed?

adamscybot:
@bcoca 

Ok, now the output is nearly there but not quite right, e.g:

```
  "size_8gb": [
    {
      "hosts": [
        "*ip was here*"
      ],
      "vars": {}
    },
    {
      "hosts": [
        "*ip was here*"
      ],
      "vars": {}
    }
  ],
```

The value of the group key is now an array not an object. Presumably it should be appending them all to the one hosts array inside of a single object?

Python issue: No just the one version. Quite strange, i'll see if I can reproduce on another machine and open another issue.

bcoca:
[...]

PR #16297, opened 2016-06-15
Fixed digital ocean inventory script "bad data for host list"

adamscybot:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.0.1.0
```
##### SUMMARY

Since the changes introduced in #16102, using the digital ocean inventory script causes an error when called via ansible:

```
You defined a group "8457573" with bad data for the host list:
 {'hosts': u'IP WAS HERE'}
```

This is because the modifications to the script output the droplet name and droplet id groups as 

``` json
{"id_or_name": "IP"}
```

Instead of the correct format:

``` json
{"id_or_name": ["IP"]}
```

This fix builds on #16100.

### DIFF
--- a/contrib/inventory/digital_ocean.py
+++ b/contrib/inventory/digital_ocean.py
@@ -373,8 +373,8 @@ or environment variables (DO_API_TOKEN)''')
 
             self.inventory['all']['hosts'].append(dest)
 
-            self.inventory[droplet['id']] = dest
-            self.inventory[droplet['name']] = dest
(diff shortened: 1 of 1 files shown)


bcoca:
this change does not 'look right' as those are single item properties, not lists, the 'hosts' entry in a group should be a list but that would not be related to the individual entries

bcoca:
nvmd, forgot the JSON format kind of conflates groups/hosts

PR #16302, opened 2016-06-15
This adds a recursive option for the assemble module

techhazard:
##### ISSUE TYPE

- Feature Pull Request

This new option is False by default (so it doesn't mess with the old behaviour)
When this option is True assemble will look into the subdirectories for files and add them, recursively

see https://github.com/ansible/ansible-modules-core/pull/3889 for the module changes

### DIFF
--- a/lib/ansible/plugins/action/assemble.py
+++ b/lib/ansible/plugins/action/assemble.py
@@ -32,11 +32,15 @@ class ActionModule(ActionBase):
 
     TRANSFERS_FILES = True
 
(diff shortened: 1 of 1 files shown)


techhazard:
I don't know how I can test if this works. Could anyone help me with that (or test it)?

Also: the "Details" link in the "Shippable" check does not link to the details but to a list of builds, and I'm not sure which  one is about this PR

techhazard:
The only failing test is about yum being deprecated, something that I don't think is related to this pull-request

techhazard:
I would really like to be able to see this merged, and for that I need to be able to test this.

PR #16328, opened 2016-06-16
Handle color on multiple line in a cleaner way

mscherer:
##### ISSUE TYPE
- Feature Pull Request
##### SUMMARY

When using ansible deployment on git push, git insert "remote:"
at the start of ansible output. If you force the color on ansible,
the "remote:" also get colored if the string to display is on
more than 1 line.

This change make sure that each end of line reset the color, instead
of reseting only at the end of the string.


mattclay:
@mscherer Thanks for this fix. It makes the color output much nicer on Shippable.

PR #16335, opened 2016-06-16
Add mention of the behavior of transport on OSX

dekimsey:
##### ISSUE TYPE
- Docs Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file = /Users/dkimsey/ansible/ansible.cfg
  configured module search path = ['library/']
```
##### SUMMARY

Currently there is no indication anywhere that smart actually means paramiko on OSX machines. This change adds a much needed blurb to the documentation.

Aside: Might be good if the connection type ('ssh', 'paramiko', 'winrm') was emitted in vvvvv logging. Tracing this was unexpectedly fun.

### DIFF
--- a/docsite/rst/intro_configuration.rst
+++ b/docsite/rst/intro_configuration.rst
@@ -758,6 +758,12 @@ Users should usually leave this setting as 'smart' and let their playbooks choos
 
     transport = paramiko
 
+On OSX, sshpass v1.05 is known to cause cause kernel panics. When executing on this system 'smart' will fall-back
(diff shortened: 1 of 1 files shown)


sivel:
This is not fully correct.  If the system requires password authentication and you are on Mac, then it will switch to paramiko.  It doesn't always use paramiko on Mac when set to `smart`.

If you are using ssh key auth instead of password auth, then it will continue to use ssh.

dekimsey:
You are correct, I'll update this.

Unrelated note: it appears my mail app is filtering github correspondence.

dharmabumstead:
Closing. Please feel free to rebase and reopen. Thanks.

PR #16336, opened 2016-06-17
add support for pmrun by dell

juliedavila:
##### ISSUE TYPE

Feature Pull Request
##### ANSIBLE VERSION

ansible 2.2.0 (pmrun c708c21e61) last updated 2016/06/16 17:49:57 (GMT -700)
  lib/ansible/modules/core: (detached HEAD 1d0f408897) last updated 2016/06/14 10:10:53 (GMT -700)
  lib/ansible/modules/extras: (detached HEAD 146c969c07) last updated 2016/06/14 10:10:53 (GMT -700)
  config file =
##### SUMMARY

Adds support for no-password pmrun priv escalation

### DIFF
--- a/lib/ansible/playbook/play_context.py
+++ b/lib/ansible/playbook/play_context.py
@@ -539,6 +539,17 @@ class PlayContext(Base):
 
                 becomecmd = '%s -u %s %s' % (exe, self.become_user, command)
 
+            elif self.become_method == 'pmrun':
+                #TODO: Add support for prompt
+                exe = self.become_exe or 'pmrun'
+
+                prompt = 'Password:'
+
+                if self.become_user:
+                    flags += ' -u %s ' % self.become_user
+
+                becomecmd = '%s %s %s' % (exe, flags, cmd)
+
             else:
                 raise AnsibleError("Privilege escalation method not found: %s" % self.become_method)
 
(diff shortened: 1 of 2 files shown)


lib/ansible/playbook/play_context.py
```diff
@@ -539,6 +539,17 @@ def detect_su_prompt(data):
 
                 becomecmd = '%s -u %s %s' % (exe, self.become_user, command)
 
+            elif self.become_method == 'pmrun':
+                #TODO: Add support for prompt
```
bcoca, line 543: so 'Password:' is not a valid prompt? you should leave unset in that case

bcoca:
To use this it also requires whitelisting at the connection plugin level.

wtcross:
@bcoca @defionscode Do one of you have a way to test this code? I think that adding password support is a great idea, but we'll need to test that it's working before merging.

wtcross:
I'm happy to help out with this PR, btw.

lib/ansible/playbook/play_context.py
juliedavila: it is a valid prompt

wtcross: If this is the valid prompt, then isn't it going to work when pmrun prompts for a password? My understanding of this bit of code isn't very good, but it seems like you just set the `prompt` variable and then password input is handled for you.

@bcoca is this a correct assumption?

wtcross:
The `pmrun` priv escalation method has been whitelisted at the connection plugin level.

juliedavila:
@wtcross looks like your pr to my fork caused some merge conflicts

PR #16339, opened 2016-06-17
Allowing hostcolor colours to be set

ProjectRice:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
2.2.0 0.0.devel
```
##### SUMMARY

Previously hostcolor would default to red, green or yellow rather than colours set in ansible.cfg etc.

### DIFF
--- a/lib/ansible/utils/color.py
+++ b/lib/ansible/utils/color.py
@@ -86,10 +86,10 @@ def colorize(lead, num, color):
 def hostcolor(host, stats, color=True):
     if ANSIBLE_COLOR and color:
         if stats['failures'] != 0 or stats['unreachable'] != 0:
-            return u"%-37s" % stringc(host, 'red')
+            return u"%-37s" % stringc(host, C.COLOR_ERROR)
         elif stats['changed'] != 0:
-            return u"%-37s" % stringc(host, 'yellow')
+            return u"%-37s" % stringc(host, C.COLOR_CHANGED)
         else:
-            return u"%-37s" % stringc(host, 'green')
+            return u"%-37s" % stringc(host, C.COLOR_OK)
     return u"%-26s" % host
 


lib/ansible/utils/color.py
```diff
@@ -86,10 +86,10 @@ def colorize(lead, num, color):
...
     if ANSIBLE_COLOR and color:
         if stats['failures'] != 0 or stats['unreachable'] != 0:
-            return u"%-37s" % stringc(host, 'red')
+            return u"%-37s" % stringc(host, COLOR_ERROR)
         elif stats['changed'] != 0:
-            return u"%-37s" % stringc(host, 'yellow')
+            return u"%-37s" % stringc(host, COLOR_CHANGED)
         else:
-            return u"%-37s" % stringc(host, 'green')
+            return u"%-37s" % stringc(host, COLOR_OK)
```
bcoca: where do these vars come from?

ProjectRice: Apologies.  They come from constants, however I forgot to refer to this.
The pull request should now be updated.

Without this change, the hostname will always be displayed in 1 of the 3 hard coded colours.  This should allows them to match whatever has been specified in the config.

Thanks.

lib/ansible/utils/color.py
```diff
@@ -86,10 +86,10 @@ def colorize(lead, num, color):
...
     if ANSIBLE_COLOR and color:
         if stats['failures'] != 0 or stats['unreachable'] != 0:
-            return u"%-37s" % stringc(host, 'red')
+            return u"%-37s" % stringc(host, c.COLOR_ERROR)
         elif stats['changed'] != 0:
-            return u"%-37s" % stringc(host, 'yellow')
+            return u"%-37s" % stringc(host, c.COLOR_CHANGED)
         else:
-            return u"%-37s" % stringc(host, 'green')
+            return u"%-37s" % stringc(host, c.COLOR_OK)
```
bcoca: you need `C`, not `c` to actually pick them up from constants

ProjectRice: You are correct..   I should have paid more attention.   Sorry.
Fixed.  

Also not sure if this should be a feature request instead of "bug".  I think it probably sits on the fence.

willthames:
ready_for_review

Issue #16189, opened 2016-06-08
Lookup password returns password including the salt - regression from 1.9


stravid:
##### ISSUE TYPE

- Bug Report
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file = 
  configured module search path = Default w/o overrides
```
##### OS / ENVIRONMENT

N/A
##### SUMMARY

(Disclaimer: I already [asked on the mailing list](https://groups.google.com/forum/#!msg/ansible-project/TRttzH0jJ54/7P9dKVmXAQAJ) but nobody seemed to take interest in the problem.)

The password produced by a create user task cannot be used somewhere else since `lookup` returns the password including the salt as suffix. Following workflow to create a user with a random password and using the password later as `ansible_sudo_password` worked in 1.9 but is broken in 2.X.

Task for creating the user:

```
- name: Create user
  user: name=deployer group=deployer password={{item}} state=present
  with_password: "passwords/users/deployer encrypt=sha256_crypt"
```

Using the password later on:

```
- hosts: all 
  remote_user: deployer
  vars: 
  - ansible_sudo_pass: "{{ lookup('password', 'passwords/users/deployer') }}"
```

pilou-:
@stravid in order to workaround the problem, you can use:
`- ansible_sudo_pass: "{{ lookup('password', 'passwords/users/deployer encrypt=') }}"`

PR #16361, opened 2016-06-19
Lookup password omit salt

pilou-:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (lookup_password_omit_salt 75cd580e6b) last updated 2016/06/20 00:42:57 (GMT +200)
  lib/ansible/modules/core: (detached HEAD f0f5272f90) last updated 2016/06/20 00:45:51 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD 6cb6829384) last updated 2016/06/20 00:45:51 (GMT +200)
```
##### SUMMARY

Unencrypted password lookup (password lookup with _encrypt_ parameter unset) must not return salt value.

Add two tests: one related to this issue, another about the behavior.

### DIFF
--- a/lib/ansible/plugins/lookup/password.py
+++ b/lib/ansible/plugins/lookup/password.py
@@ -137,25 +137,23 @@ class LookupModule(LookupBase):
 
                 password = content
                 salt = None
-                if params['encrypt'] is not None:
-                    try:
-                        sep = content.rindex(' ')
-                    except ValueError:
-                        # No salt
-                        pass
-                    else:
-                        salt_field = content[sep + 1:]
-                        if salt_field.startswith('salt='):
-                            password = content[:sep]
-                            salt = salt_field[len('salt='):]
 
+                try:
+                    sep = content.rindex(' salt=')
+                except ValueError:
+                    # No salt
+                    pass
+                else:
+                    salt = password[sep + len(' salt='):]
+                    password = content[:sep]
+
+                if params['encrypt'] is not None and salt is None:
                     # crypt requested, add salt if missing
-                    if not salt:
-                        salt = self.random_salt()
-                        content = '%s salt=%s' % (password, salt)
-                        with open(path, 'w') as f:
-                            os.chmod(path, 0o600)
-                            f.write(content + '\n')
+                    salt = self.random_salt()
+                    content = '%s salt=%s' % (password, salt)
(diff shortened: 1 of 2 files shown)


pilou-:
@mattclay how can I update the docker images used by shippable ? I would like to add `python-passlib` packages in order to add some non-destructive integration tests related to password lookup (see ad80f9a03c7a0b6f70e0087baca5994e541e3581).

mattclay:
@pilou- Create a separate PR with the Dockerfile changes and reference this PR. Once that PR is merged I can rebuild the docker images so the tests on this PR can pass.

mattclay:
This PR restores the behavior found in the 1.9 release and would resolve issue #16189.

lib/ansible/plugins/lookup/password.py
```diff
@@ -137,26 +137,23 @@ def run(self, terms, variables, **kwargs):
...
+                try:
+                    sep = content.rindex(' salt=')
+                except ValueError:
+                    # No salt
+                    pass
+                else:
+                    salt = password[sep + len(' salt='):]
+                    password = content[:sep]
+
+                # crypt requested, add salt if missing
```
mattclay: Restore the conditional before this block:

``` python
if params['encrypt'] is not None:
```

Otherwise the salt will be generated and saved to the file even if crypt is not requested.

lib/ansible/plugins/lookup/password.py
```diff
@@ -137,26 +137,23 @@ def run(self, terms, variables, **kwargs):
...
+                if not salt:
+                    salt = self.random_salt()
+                    content = '%s salt=%s' % (password, salt)
+                    with open(path, 'w') as f:
+                        os.chmod(path, 0o600)
+                        f.write(content + '\n')
             if params['encrypt']:
```
mattclay: Keep a blank line before this if statement.

pilou-: Done

lib/ansible/plugins/lookup/password.py
pilou-: Done (and another test added)

mattclay:
@pilou- Merged. Thanks for the fix.

PR #16362, opened 2016-06-20
[docker_container] image lookup should default to returning an empty list before searching

cornfeedhobo:
##### ISSUE TYPE

- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0.0
```
##### SUMMARY

When pulling multiple tags of the same image, `_image_lookup()` was returning all existing images, without respect to the passed tag

**Example playbook:**

```

---

- docker_container:
    name: one
    image: alpine:3.1
    command: 'true'

- docker_container:
    name: two
    image: alpine:3.2
    command: 'true'
```

Expected outcome:
  Both images should be pulled and started

Current outcome:
  The second image is erroneously matched and attempted to be started

### DIFF
--- a/lib/ansible/module_utils/docker_common.py
+++ b/lib/ansible/module_utils/docker_common.py
@@ -156,7 +156,7 @@ class AnsibleDockerClient(Client):
                                                                                            MIN_DOCKER_VERSION))
 
         self.debug = self.module.params.get('debug')
-        self.check_mode = self.module.check_mode
+        self.check_mode = self.module.check_mode  
         self._connect_params = self._get_connect_params()
 
         try:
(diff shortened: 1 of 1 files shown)


lib/ansible/module_utils/docker_common.py
```diff
@@ -418,24 +418,26 @@ def find_image(self, name, tag):
...
 
-    def _image_lookup(self, name, tag): 
+    def _image_lookup(self, name, tag):
         '''
-        Including a tag in the name parameter sent to the docker-py images method does not 
+        Including a tag in the name parameter sent to the docker-py images method does not
         work consistently. Instead, get the result set for name and manually check if the tag
         exists.
         '''
+        images = list()
```
mgedmin: An empty list is more commonly spelled `[]` in Python.

PR #16367, opened 2016-06-20
Create unique name when two hosts have the same hostname

Jonnymcc:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel 1db02dfb71) last updated 2016/06/20 10:03:16 (GMT -400)
  lib/ansible/modules/core: (detached HEAD f0f5272f90) last updated 2016/06/17 14:30:25 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 6cb6829384) last updated 2016/06/17 14:26:51 (GMT -400)
  config file = 
```
##### SUMMARY

ec2.py allows specifying a different source for the inventory_name by setting hostname_variable in ec2.ini. An example is setting hostname_variable = tag_Name. In the event where there are multiple hosts with the same tag_Name value (hostname_variable) the subsequent instances overwrite the hostvars for the instance in the inventory file.

So if I had two instances named "app" and one had tag_env_dev and the other had tag_env_production, when running ansible/ansible-playbook with --limit tag_env_dev, the play would actually run against the instance tagged production.

Before

```
{
  "_meta": {
    "hostvars": {
      "app": {
...
  }, 
  "ami_f0091d91": [
    "app", 
    "app"
  ], 
```

After

```
{
  "_meta": {
    "hostvars": {
      "app": {
...
      }, 
      "app(i-d9f43776)": {
        "ansible_ssh_host": "10.95.101.221",
...
  }, 
  "ami_f0091d91": [
    "app", 
    "app(i-d9f43776)"
  ], 
```

### DIFF
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -723,6 +723,9 @@ class Ec2Inventory(object):
         if self.pattern_exclude and self.pattern_exclude.match(hostname):
             return
 
+        if hostname in self.index:
+            hostname = hostname + '(%s)' % instance.id
+
         # Add to index
         self.index[hostname] = [region, instance.id]
 
@@ -854,6 +857,9 @@ class Ec2Inventory(object):
             hostname = dest
 
         hostname = self.to_safe(hostname).lower()
+
+        if hostname in self.index:
+            hostname = hostname + '(%s)' % instance.id
 
         # Add to index
         self.index[hostname] = [region, instance.id]


Jonnymcc:
I believe the problem here was fixed and tests need to be re-run.

Jonnymcc:
ready_for_review

calfonso:
This scenario would only occur if you choose to use hostname in the index. You can choose to use the uuid for the ec2 instance and you wouldn't hit this edge case. Closing this and if you disagree, let us know.

Thank you!

Jonnymcc:
Although I do not use a value other than the default these days, it would be nice to have fixed collisions like described here. While I stumbled over this issue it was somewhat harmless (deployed my app to my jenkins server instead...) for others being caught by surprise like this could have worse consequences. It may be an edge case, but in programming we should try to account for as many edge cases as we can, no?

PR #16377, opened 2016-06-21
fixed plugin directories

steenzout:
##### ISSUE TYPE
- Docs Pull Request
##### ANSIBLE VERSION

All
##### SUMMARY

The paths written in the docs for the plugins is inconsistent with [lib/ansible/constants.py](https://github.com/ansible/ansible/blob/devel/lib/ansible/constants.py).

### DIFF
--- a/docsite/rst/developing_plugins.rst
+++ b/docsite/rst/developing_plugins.rst
@@ -106,13 +106,12 @@ Distributing Plugins
 Plugins are loaded from both Python's site_packages (those that ship with ansible) and a configured plugins directory, which defaults
 to /usr/share/ansible/plugins, in a subfolder for each plugin type::
 
-    * action_plugins
-    * lookup_plugins
-    * callback_plugins
-    * connection_plugins
-    * filter_plugins
-    * vars_plugins
-    * strategy_plugins
+    * action
+    * lookup
+    * callback
+    * connection
+    * filter
+    * strategy
 
 To change this path, edit the ansible configuration file.
 


docsite/rst/intro_configuration.rst
```diff
@@ -70,7 +70,7 @@ Actions are pieces of code in ansible that enable things like module execution,
 This is a developer-centric feature that allows low-level extensions around Ansible to be loaded from
 different locations::
 
-   action_plugins = ~/.ansible/plugins/action_plugins/:/usr/share/ansible_plugins/action_plugins
+   action_plugins = ~/.ansible/plugins/action:/usr/share/ansible/plugins/action
```
bcoca: these changes are not needed, names are actually arbitrary here.

the one thing it does reinforce is the usage of `x_plugins` which is still the way playbook/role plugin directories are named

docsite/rst/developing_plugins.rst
```diff
@@ -106,13 +106,13 @@ Distributing Plugins
...
-    * connection_plugins
-    * filter_plugins
-    * vars_plugins
-    * strategy_plugins
+    * action
+    * lookup
+    * callback
+    * connection
+    * filter
+    * vars
```
bcoca: vars plugins are not loaded from here (we do plan to do so in future)

steenzout: @bcoca well that's what was on the docs.
so I should just remove it then?

docsite/rst/intro_configuration.rst
steenzout: in this particular case/example/context, it has nothing to do with the directories you would have if you were implementing a role which is emphasized by this statement: `low-level extensions around Ansible to be loaded from different locations`.

I understand they are arbitrary but, in my opinion, the best example case here is to use what you have as default which makes it straightforward step for any developer to know what typically is set and how to modify it to their own taste.

sivel: I am in alignment with @bcoca.  The typical naming patterns are to use `x_plugins` and we should not diverge from that in our examples..

steenzout: I'll remove it then but out of curiosity, @sivel why do you follow a different convention in `constants.py`?

steenzout:
@bcoca @sivel I've updated the PR and removed the elements you didn't like.

let me know if more changes are needed.

PR #16400, opened 2016-06-22
Remove break word for docsite tables

jlmitch5:
This still allows the lines to wrap with lots of text, like comments cells.


jimi-c:
Merged, thanks!

PR #16404, opened 2016-06-22
Bug fix for GalaxyCLI.execute_info FIXME

s-hertel:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file =
```
##### SUMMARY

The initial code was broken for both 1.9 and 2.0 because another method (GalaxyCLI._display_role_info) was changed (to always return something). The intention of the old code was trying to identify whether or not the role existed (by the returned value from the aforementioned method). My fix for the code looks at the role's path and checks that the directory for the role actually exists.

### DIFF
--- a/lib/ansible/cli/galaxy.py
+++ b/lib/ansible/cli/galaxy.py
@@ -144,7 +144,7 @@ class GalaxyCLI(CLI):
 
         text = [u"", u"Role: %s" % to_unicode(role_info['name'])]
         text.append(u"\tdescription: %s" % role_info.get('description', ''))
-
+        
         for k in sorted(role_info.keys()):
 
             if k in self.SKIP_INFO_KEYS:
@@ -158,7 +158,7 @@ class GalaxyCLI(CLI):
                     text.append(u"\t\t%s: %s" % (key, role_info[k][key]))
             else:
                 text.append(u"\t%s: %s" % (k, role_info[k]))
-
+        
         return u'\n'.join(text)
 
 ############################
@@ -174,7 +174,7 @@ class GalaxyCLI(CLI):
         init_path  = self.get_opt('init_path', './')
         force      = self.get_opt('force', False)
         offline    = self.get_opt('offline', False)
-
+        
         role_name = self.args.pop(0).strip() if self.args else None
         if not role_name:
             raise AnsibleOptionsError("- no role name specified for init")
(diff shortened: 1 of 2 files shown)


lib/ansible/cli/galaxy.py
```diff
@@ -305,9 +305,9 @@ def execute_info(self):
                 role_info.update(role_spec)
 
             data = self._display_role_info(role_info)
-            ### FIXME: This is broken in both 1.9 and 2.0 as
-            # _display_role_info() always returns something
-            if not data:
+            
+            role_location = os.path.join(roles_path[0], role)
```
bcoca: this seems incorrect, why would you assume the role is only in the first possible path?

lib/ansible/cli/galaxy.py
```diff
@@ -174,7 +174,7 @@ def execute_init(self):
         init_path  = self.get_opt('init_path', './')
         force      = self.get_opt('force', False)
         offline    = self.get_opt('offline', False)
-
+
```
bcoca, line 177: please don't add this extra whitespace

bcoca:
a title of 'bugfix' is not very informative, please state what bug you are fixing and then give details in the description.

lib/ansible/cli/galaxy.py
s-hertel: Does roles_path ever store more than one string? Hadn't noticed that with any of the roles I've installed and my interaction with the code.

lib/ansible/cli/galaxy.py
s-hertel, line 177: Sorry. Still figuring this out. The lines were unintentional, I didn't know how to get rid of them.

lib/ansible/cli/galaxy.py
bcoca: roles_path is not from roles, it is an Ansible configuration that defines the directories in which Ansible will look for roles (aside from roles/ adjacent to play). Not to be confused with `role_path` which returns the path to the current role.

alikins:
I have a related work-in-progress branch at https://github.com/ansible/ansible/pull/16283

It runs into a similar issue when finding roles by name: The same role can exist in multiple places if there are multiple dirs in roles_paths.  Not sure what the best way to resolve which is 'best' in that case, but I ended up just picking the first one as well (see https://github.com/ansible/ansible/pull/16283/commits/7e03007e244a238a7814e8fb67342e94ef239ba5#diff-8547ca680c18ff8ac6cbba85802d5c1eR85). Haven't messed with branch in a while, but I believe the order should respect the order of paths in the configured roles paths. 

The other assumption I had to make was when assigning a path to a role that was created anew. Currently my branch will assign the first path in roles path in that case.
Not sure if that is correct however.

bcoca:
Also you need to consider that a roles/ dir adjacent to play should be included in the lookup

s-hertel:
Alright. Thanks for the patience! It's helpful to see your branch, alikins (I'm understanding the issue better now with your elaboration). bcoca: can I close this pull request?

Issue #16445, opened 2016-06-26
Holders for partitions are not being captured


larsf:
##### ISSUE TYPE
Bug Report

##### COMPONENT NAME
setup module

##### ANSIBLE VERSION
```
Covers all Ansible versions
```

##### CONFIGURATION
default

##### OS / ENVIRONMENT
Linux

##### SUMMARY
Ansible facts does not collect Holders for partitions on linux

##### STEPS TO REPRODUCE
ansible -m setup

##### EXPECTED RESULTS
```
"sda": {
                "holders": [],
                "host": "SCSI storage controller: LSI Logic / Symbios Logic 53c1030 PCI-X Fusion-MPT Dual Ultra320 SCSI (rev 01)",
                "model": "Virtual disk",
                "partitions": {
                    "sda1": {
                        "holders": [],
                        "sectors": "2097152",
                        "sectorsize": 512,
                        "size": "1.00 GB",
                        "start": "2048"
                    },
                    "sda2": {
                        "holders": [
                            "vg_root-lv_root",
                            "vg_root-lv_swap"
                        ],
                        "sectors": "207616000",
                        "sectorsize": 512,
                        "size": "99.00 GB",
                        "start": "2099200"
                    }
                },
                "removable": "0",
                "rotational": "1",
                "scheduler_mode": "deadline",
                "sectors": "209715200",
                "sectorsize": "512",
                "size": "100.00 GB",
                "support_discard": "0",
                "vendor": "VMware"
            },
...
 "sdd": {
                "holders": [
                    "luks-sdd"
                ],
                "host": "SCSI storage controller: LSI Logic / Symbios Logic 53c1030 PCI-X Fusion-MPT Dual Ultra320 SCSI (rev 01)",
                "model": "Virtual disk",
                "partitions": {},
                "removable": "0",
                "rotational": "1",
                "scheduler_mode": "deadline",
                "sectors": "209715200",
                "sectorsize": "512",
                "size": "100.00 GB",
                "support_discard": "0",
                "vendor": "VMware"
            },
```
##### ACTUAL RESULTS
```
"sda": {
                "holders": [],
                "host": "SCSI storage controller: LSI Logic / Symbios Logic 53c1030 PCI-X Fusion-MPT Dual Ultra320 SCSI (rev 01)",
                "model": "Virtual disk",
                "partitions": {
                    "sda1": {
                        "sectors": "2097152",
                        "sectorsize": 512,
                        "size": "1.00 GB",
[...]

PR #16446, opened 2016-06-26
add holders to partitions

larsf:
Fix for Issue #16445

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -1249,6 +1249,18 @@ class LinuxHardware(Hardware):
                          'uuid': uuid,
                          })
 
+    def get_holders(self, block_dev_dict, sysdir):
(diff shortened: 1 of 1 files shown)


lib/ansible/module_utils/facts.py
```diff
@@ -1308,6 +1308,17 @@ def get_device_facts(self):
                     if not part['sectorsize']:
                         part['sectorsize'] = get_file_content(part_sysdir + "/queue/hw_sector_size",512)
                     part['size'] = self.module.pretty_bytes((float(part['sectors']) * float(part['sectorsize'])))
+                    part['holders'] = []
```
bcoca: since this is basically a copy of the existing code for devices, i would make it into a function and call it from both spots

larsf: will do

larsf:
@bcoca created the function

PR #16458, opened 2016-06-27
Don't treat parsing problems as async task timeout

emonty:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file = 
```
##### SUMMARY

If there is a problem reading/writing the status file that manifests as
not being able to parse the data, that doesn't mean the task timed out,
it means there was what was likely a tempoarary problem. Move on and
keep polling for success. The only things that should cause the async
status to not be parseable are bugs in the async_runner.

### DIFF
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -596,7 +596,11 @@ class TaskExecutor:
             time_left -= self._task.poll
 
         if int(async_result.get('finished', 0)) != 1:
-            return dict(failed=True, msg="async task did not complete within the requested time")
+            if async_result.get('parsed'):
+                return dict(failed=True, msg="async task did not complete within the requested time")
+            else:
+                return dict(failed=True, msg="async task produced unparseable results", async_result=async_result)
+            else:
         else:
             return async_result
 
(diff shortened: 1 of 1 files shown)


pabelanger:
+1

abadger:
Tentative +1 to this.  We talked about it in IRC.  The drawback of doing this is that parsing errors will no longer fail the task... we'll end up waiting the full timeout period instead.

@emonty we should add a comment explaining the condtions that we exit here, though, and your explanation of why we are ignoring parsing errors.   (otherwise some later refactoring may remove it accidentally.)

bcoca:
looks good, but you got submodule ref updates that should not be here

lib/ansible/executor/task_executor.py
```diff
@@ -585,13 +585,22 @@ def _poll_async_result(self, result, templar):
...
 
             time_left -= self._task.poll
 
         if int(async_result.get('finished', 0)) != 1:
-            return dict(failed=True, msg="async task did not complete within the requested time")
+            if async_result.get('parsed'):
+                return dict(failed=True, msg="async task did not complete within the requested time")
+            else:
+                return dict(failed=True, msg="async task produced unparseable results", async_result=async_result)
+            else:
```
abadger: Spurious `else:`
Breaking the tests.

PR #16459, opened 2016-06-27
Make '/usr/share/ansible' default modules path

machacekondra:
##### ISSUE TYPE

- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  configured module search path = ['', '/home/ondra/workspace/ovirt-engine-ansible/ansible']
```
##### SUMMARY

I can see everywhere in documentaion that '/usr/share/ansible' directory is default directory where modules are searched, but it's not true at all. This patch adds '/usr/share/ansible' as DEFAULT_MODULE_PATH.

```
ansible 2.1.0.0
  configured module search path = ['', '/home/ondra/workspace/ovirt-engine-ansible/ansible']
```


bcoca:
That is incorrect, /usr/share/ansible is the default in most distributions but not in all, this depends on the install. 
Ansible has some autodetect magic when run, so there is no need to have a hardcoded default.

PR #16487, opened 2016-06-29
In some cornercases, sys.executable is empty.  Work around that by choosing a sensible default path to python and emitting a warning

### LINKED ISSUE #13585: No PATH in Environment results in "Permission denied" for tasks on localhost
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 2.1.0
  configured module search path = /usr/share/ansible
##### Environment:

Debian Jessie, but should not matter
##### Summary:

If $PATH is not set in the environment, ansible uses an empty string for `ansible_python_interpreter` for localhost. This leads to a task failure with "Permission denied" when running tasks on localhost.

(The problem did not appear with ansible 1.7.2 as shipped with Debian and only appeared when I switched to the development version)
##### Steps To Reproduce:

```
root@foo:/ # env -i HOME=/root ansible -m shell -a : localhost
```
##### Expected Results:

```
localhost | SUCCESS | rc=0 >>
```
##### Actual Results:

```
localhost | FAILED | rc=0 >>
/bin/sh: 1: /root/.ansible/tmp/ansible-tmp-1450293879.27-221034150901375/command: Permission denied
```

Here's the same command with PATH:

```
root@foo:/ # env -i HOME=/root PATH="/usr/sbin:/usr/bin:/sbin:/bin" ansible -m shell -a : localhost
localhost | SUCCESS | rc=0 >>
```

The reason for this is that in https://github.com/ansible/ansible/blob/devel/lib/ansible/inventory/__init__.py#L453 the python interpreter for localhost is set to `sys.executable`, which is empty, if PATH is empty:

```
root@foo:/ # python -c 'import sys; print(sys.executable)'
/usr/bin/python
root@foo:/ # env -i python -c 'import sys; print(sys.executable)'
```

The shebang for the shell module is then set to `#!` while preparing the module for execution, which leads to the `Permission denied` message. 

I spotted this, when running an ansible playbook from a remote host through `remctl`, which does not set PATH, so there is an actual use case where this problem might occur.

I could not find a better way to reliably determine the python interpreter than using `sys.executable`. So I would suggest to print a warning message and ask the user to set `ansible_python_interpreter` in the inventory if `sys.executable` is empty, because it is really not obvious that no PATH results in "Permission denied".

abadger:
##### ISSUE TYPE

- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.x
```
##### SUMMARY

Fixes #13585

### DIFF
--- a/lib/ansible/inventory/__init__.py
+++ b/lib/ansible/inventory/__init__.py
@@ -479,7 +479,12 @@ class Inventory(object):
         new_host.vars = self.get_host_vars(new_host)
         new_host.set_variable("ansible_connection", "local")
         if "ansible_python_interpreter" not in new_host.vars:
-            new_host.set_variable("ansible_python_interpreter", sys.executable)
+            py_interp = sys.executable
+            if not py_interp:
+                # sys.executable is not set in some cornercases.  #13585
+                display.warning('Unable to determine python interpreter from sys.executable. Using /usr/bin/python default. You can correct this by setting ansible_python_interpreter for localhost')
+                py_interp = '/usr/bin/python'
+            new_host.set_variable("ansible_python_interpreter", py_interp)
         self.get_group("ungrouped").add_host(new_host)
         return new_host
 


abadger:
Merged to devel and stable-2.1 branches.

PR #16492, opened 2016-06-29
basic: omit args if default value is set

resmo:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
2.2
```
##### SUMMARY

Currently, for `required_one_of`, `mutually_exclusive`, `required_together`, etc. the checks only complain if the param was passed or not passed.

With this change, the checks `required_one_of`, `mutually_exclusive`, `required_together` will only complain if the param passed has a different value than the default value.

As the param has the default value set anyway, passing the default value again, should not be handled as an error.

This change makes it possible to omit or "reset" a param so it is not going to be handled in error checking if passed.

Let's make a simple example:

Given is a inventory having a group of web servers

``` ini
[webservers]
web01
web02
web03
```

A simple playbook:

``` yaml
# xy.yaml

---
- hosts: webservers
  tasks: 
  - local_action:
      module: test_defaults
      name: hello
      template: "{{ template | default(omit) }}"
      iso: "{{ iso | default(omit) }}"
```

We want to use a template for all webs, lets make a group_var

``` yaml
# group_vars/webservers

---
template: mytemplate
```

Works ok!

Now we realize one webserver would need an iso instead the template. This would **not** work as template is still defined,

``` yaml
# host_vars/web03

---
iso: iso
```

With the patch, you can "reset" the template with the default value, this would work:

``` yaml
# host_vars/web03

---
iso: iso
template: null
```
###### Test setup:

Dummy module:

``` python
#!/usr/bin/python

def main():
    argument_spec = dict(
        name = dict(default=None),
        display_name = dict(default=None),
        template = dict(default=None),
        iso = dict(default=None),
        domain = dict(default=None),
        account = dict(default=None),
    )

    module = AnsibleModule(
        argument_spec=argument_spec,
        required_together=(
            ['domain', 'account'],
        ),
        required_one_of = (
            ['display_name', 'name'],
        ),
        mutually_exclusive = (
            ['template', 'iso'],
        ),
        supports_check_mode=True
    )

    module.exit_json(changed=False)

# import module snippets
if __name__ == '__main__':
    main()
```

Test playbook:

``` yaml

---
- hosts: localhost
  tasks:

# This task would fail in ansible 2.1 as mutually exclusive params are set.
# It does not after this change as one param value is identical to the default value.
  - local_action:
      module: test_defaults
      name: hello
      # mutually_exclusive
      template: null
      iso: foo
    register: test1
    ignore_errors: true

# This task would fail in ansible 2.1 as required_together param account is not given.
# It does not after this change as the one param value is identical to the default value.

  - local_action:
      module: test_defaults
      # requried_one_of
      name: hello
      # required_together
      domain: null
    register: test2
    ignore_errors: true

# this would not fail in 2.1 but it should as the default values are set but we expect a non-default one.

  - local_action:
      module: test_defaults
      # requried_one_of
      name: null
      display_name: null
    register: test3
    ignore_errors: true

# this fails in 2.1 and with this change, as different values then the defaults are used

  - local_action:
      module: test_defaults
      name: hello
      # mutually_exclusive
      template: my_template
      iso: my_iso
    register: test4
    ignore_errors: true

# this would not fail in 2.1 but it should.
# For the required_together the default values are set but we expect a non-default one.

  - local_action:
      module: test_defaults
      name: hello
      # required_together
      domain: root
      account: null
    register: test5
    ignore_errors: true

  - assert:
      that:
      - test1|success

      - test2|success

      - not test3|success
      - '"one of the following is required" in test3.msg'

      - not test4|success
      - '"parameters are mutually exclusive" in test4.msg'

      - not test5|success
      - '"parameters are required together" in test5.msg'
```

Test results:

```
 $ ansible-playbook test.yml -vv

1 plays in test.yml

ok: [localhost]

TASK [test_defaults] ***********************************************************
task path: /home/resmo/tmp/ansible-defaults-revamp/test.yml:7
ok: [localhost -> localhost] => {"changed": false}

TASK [test_defaults] ***********************************************************
task path: /home/resmo/tmp/ansible-defaults-revamp/test.yml:19
ok: [localhost -> localhost] => {"changed": false}

TASK [test_defaults] ***********************************************************
task path: /home/resmo/tmp/ansible-defaults-revamp/test.yml:30
fatal: [localhost -> localhost]: FAILED! => {"changed": false, "failed": true, "msg": "one of the following is required: display_name,name"}
...ignoring

TASK [test_defaults] ***********************************************************
task path: /home/resmo/tmp/ansible-defaults-revamp/test.yml:40
fatal: [localhost -> localhost]: FAILED! => {"changed": false, "failed": true, "msg": "parameters are mutually exclusive: ['template', 'iso']"}
...ignoring

TASK [test_defaults] ***********************************************************
task path: /home/resmo/tmp/ansible-defaults-revamp/test.yml:52
fatal: [localhost -> localhost]: FAILED! => {"changed": false, "failed": true, "msg": "parameters are required together: ['domain', 'account']"}
...ignoring

TASK [assert] ******************************************************************
task path: /home/resmo/tmp/ansible-defaults-revamp/test.yml:61
ok: [localhost] => {"changed": false, "msg": "all assertions passed"}

localhost                  : ok=7    changed=0    unreachable=0    failed=0 
```

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -1276,7 +1276,9 @@ class AnsibleModule(object):
         count = 0
         for term in check:
             if term in self.params:
-                count += 1
+                v = self.argument_spec[term]
+                if self.params[term] != v.get('default'):
+                    count += 1
         return count
 
     def _check_mutually_exclusive(self, spec):


abadger:
The benefit of this patch is that default values specified in roles and overridden in a task won't trip over mutually_exclusive.

The drawback of doing this is that mutually_exclusive (and the other checks that depend on the same knowledge of what was specified) will still pass if the user specifies two mutually exclusive values in the parameter list and one of the values is the default value.

I'm slightly against doing this because of the latter problem but could be convinced that the first problem is more severe than the second if someone cares to make a case for it.

resmo:
Thanks @abadger for the review.

The drawback you mentioned doesn't look like a real problem to me as the default value of the param is set in either way, by passing the it with an argument or by not passing it using the default arg spec value. The only drawback I see is that users may not understand why no error happens (this can easily be solved by giving some output in -v).

But the real benefit I see with this patch is simplicity, especially for roles with defaults:

Imaging we have a role xy and using the module mentioned in the description having mutually exclusive arguments  `iso` and `template`:

We'd like to use defaults to make some defaults for our role:

``` yaml
# file: xy/defaults/main.yml
iso: Debian 8
template: null
```

In 2.1 you must use 2 tasks if you want to use defaults for the value:

``` yaml
# file xy/tasks/main.yml
  - local_action:
      module: test_defaults
      name: hello
      template: "{{ template }}"
    when: template is defined and template

  - local_action:
      module: test_defaults
      name: hello
      iso: "{{ iso }}"
    when: iso is defined and iso
```

if you have an extensive module with many mutually_exclusive values, it even gets a lot more to do.

With the patch you can simply make one task and you don't even need a condition as long as the value is the default one (used anyway if set or not).

``` yaml
# file xy/tasks/main.yml
  - local_action:
      module: test_defaults
      name: hello
      template: "{{ template }}"
      iso: "{{ iso }}"
```

bcoca:
wouldn't this already work by using`|default(omit)`?

resmo:
`|default()` handles undefined vars, since we default the vars in the role defaults, `|default(omit)` would not be used.

When we would not set`template: null` in role defaults, we could use `|default(omit)` as `template` is undefined. But this would only work in case we want to use `iso`. When we want to use `template` we would have to make the `iso` "undefined" so `|default(omit)` would handle it but AFAIK there is no way to undefine a variable once defined.

Even if we say, we do not define any role defaults, once defined e.g. `iso` in e.g group vars/all there is no way to take it back and use `template` later in the group vars hierarchy as there is no way to undefine a variable again.

But with this patch allows the variable in the group vars hierarchy to be omitted if the default value is set and role defaults are handled properly.

bcoca:
that is a bit of tail chasing .... if you want to omit the option when it is not defined ... don't define it!

you don't NEED to define it in defaults if you use `|default(omit)`.

resmo:
This advice is not practicable. You always have a default as there is variable precedence!

bcoca:
only because you are doing it explicitly

resmo:
**NOTE: I put the following example to the summary**

Not sure if I understand you correctly. It is not a question of role defaults, it was just an example. 

Let me give another example, the most obvious one:

given is a inventory having a group of web servers

``` ini
[webservers]
web01
web02
web03
```

A simple playbook:

``` yaml
# xy.yaml
---
- hosts: webservers
  tasks: 
  - local_action:
      module: test_defaults
      name: hello
      template: "{{ template | default(omit) }}"
      iso: "{{ iso | default(omit) }}"
```

We want to use a template for all webs, lets make a group_var

``` yaml
# group_vars/webservers
---
template: mytemplate
```

Works ok!

Now we realize one webserver would need an iso instead the template. This would **not** work as template is still defined,

``` yaml
# host_vars/web03
---
iso: iso
```

With the patch, you can "reset" the template with the default value, this would work:

``` yaml
# host_vars/web03
---
iso: iso
template: null
```

resmo:
PR is obsolete as it can be solved with

```
{{ some_var | default(omit, True ) }}
```

Docs http://docs.ansible.com/ansible/playbooks_filters.html#defaulting-undefined-variables need an update.

resmo:
Unfortunately `"{{ <falsy> | default(omit, true) }}"`will fail if the param is a bool and set to like to set the value `False`. In this case it will be omitted.

sample module:

``` python
    module = AnsibleModule(
        argument_spec=dict(
            name = dict(default=None),
            enabled = dict(type='bool', default=True),
            force = dict(type='bool', default=False),
        ),
        mutually_exclusive = (
            ['enabled', 'force'],
        ),
    )
```

sample playbook, set enabled=False

``` yaml
- hosts: localhost
  connection: local
  vars:
    enabled: False
    force: ''
  tasks:
  - local_action:
      module: test_defaults
      name: hello
      # mutually_exclusive
      enabled: "{{ enabled | default(omit, true) }}"
      force: "{{ force | default(omit, true) }}"
```

## expected:

enabled=False

```
ok: [localhost -> localhost] => {
    "changed": false, 
    "invocation": {
        "module_args": {
            "enabled": false, 
            "force": false, 
            "name": "hello"
        }, 
        "module_name": "test_defaults"
    }
}
```

## actually

enabled=True

```
ok: [localhost -> localhost] => {
    "changed": false, 
    "invocation": {
        "module_args": {
            "enabled": true, 
            "force": false, 
            "name": "hello"
        }, 
        "module_name": "test_defaults"
    }
}
```

with patch and playbook:

``` yaml
---
- hosts: localhost
  connection: local
  vars:
    enabled: False
    force: False
  tasks:
  - local_action:
      module: test_defaults
      name: hello
      # mutually_exclusive
      enabled: "{{ enabled }}"
      force: "{{ force }}"
```

I get what I expected:

```
ok: [localhost -> localhost] => {"changed": false, "invocation": {"module_args": {"enabled": false, "force": false, "name": "hello"}, "module_name": "test_defaults"}}
```

IMHO this PR is still valid.

bcoca:
Well, you are telling it to assume false is undefined, just omit the `true` or write `"{{ <falsy> | default(omit, false) }}"` when you are using a 'falsifiable' default.

bcoca:
Hi!

PR #16493, opened 2016-06-29
version_compare  argument fix

ytetsuro:
##### ISSUE TYPE

- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel 216a845244)
```
##### SUMMARY

```
-def version_compare(value, version, operator='eq', strict=False):
 +def version_compare(value, operator='eq', version='0.0', strict=False):
```

It is difficult to understand the order
## before

{{ version | version_compare('2.5', '<=') }}

Do not know whether the "version <= '2.5'" and whether the "'2.5' <= version"

it proposes a this for difficult to understand that it is "version <= '2.5'" .

### DIFF
--- a/lib/ansible/plugins/filter/core.py
+++ b/lib/ansible/plugins/filter/core.py
@@ -186,7 +186,7 @@ def ternary(value, true_val, false_val):
         return false_val
 
 
-def version_compare(value, version, operator='eq', strict=False):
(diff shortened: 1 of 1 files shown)


jctanner:
@ytetsuro I don't completely understand what you are trying to fix, but your patch breaks the version compare integration tests. Please revise.

PR #16500, opened 2016-06-29
Add some test cases for parsing/data_loader

alikins:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME
core

##### ANSIBLE VERSION

```
ansible 2.2.0 (dataloader_testcases 9bc5573b4b) last updated 2016/06/29 11:32:54 (GMT -400)
  lib/ansible/modules/core: (detached HEAD 2f0f04437b) last updated 2016/06/29 10:48:47 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD d25c487ac8) last updated 2016/06/29 10:48:48 (GMT -400)

```
##### SUMMARY

Some potential unit test cases related to the dataload path_dwim changes.

Note some of these tests currently fail, either
because of unhandled exceptions on bad input (which
may be expected) or because the returned values
need to be verify correct.

So, this may just need tests updated, or it could
mean changes in the methods.


test/units/parsing/test_data_loader.py
```diff
@@ -62,6 +63,119 @@ def test_parse_fail_from_file(self, mock_def):
...
+        print(ret)
+
+    def test_empty_strings(self):
+        ret = self._loader.path_dwim_relative('', '', '')
+        assert ret is not None, "Returned path should not be None"
+        # FIXME: Not sure what this is expected to do outside of a playbook
+        self.assertEquals(os.path.curdir, ret)
+
+    def test_all_slash(self):
+        ret = self._loader.path_dwim_relative('/', '/', '/')
```
bcoca: dirname should never be a path, just a string

test/units/parsing/test_data_loader.py
```diff
@@ -62,6 +63,119 @@ def test_parse_fail_from_file(self, mock_def):
...
+        self.assertEquals(u'/', ret)
+
+    def test_path_endswith_role_from_slash(self):
+        ret = self._loader.path_dwim_relative(path='foo/bar/tasks/', dirname='/', source='/')
+        assert ret is not None, "Returned path should not be None"
+        self.assertEquals(u'/', ret)
```
bcoca: this would not be the expected return, it should be a list with one item, `foo/bar/tasks`

test/units/parsing/test_data_loader.py
bcoca: source is expected to be a file or path to a file, we should ALSO test directories, but the main focus should be files

test/units/parsing/test_data_loader.py
```diff
@@ -62,6 +63,119 @@ def test_parse_fail_from_file(self, mock_def):
...
+        # FIXME: trailing / expected? if not, wrap ret in a os.path.join(ret, '')
+        assert ret is not None, "Returned path should not be None"
+        self.assertEquals(os.path.expanduser(u'~/'), ret)
+
+    def test_all_slash(self):
+        ret = self._loader.path_dwim_relative_stack('/', '/', '/')
+        assert ret is not None, "Returned path should not be None"
+        self.assertEquals(u'/', ret)
+
+    def test_path_endswith_role(self):
```
bcoca: a missing file should return None, but we should also test 'finding a file'

alikins:
rebased and updated with more tests.

This also adds a few test data files for testing with real fs calls. Usually try to avoid that, but I think the extra test coverage is worth it here (and avoids lots of fragile/tedious mocking...)

test/units/parsing/data.yml
```diff
@@ -0,0 +1,6 @@
+---
+- item1
```
alikins, line 2: Currently, the local data files needed for the tests is just added to the same directly as the test module.
A specific test data dir may be more appropriate (a fixtures/ subdir, or a test/units/data/ dir, etc) and can
be changed if this simpler approach isnt sufficient.

mattclay:
CI failure due to:

```
2017-02-01 23:26:28 =================================== FAILURES ===================================
2017-02-01 23:26:28 _______________________ TestDataLoader.test_is_directory _______________________
2017-02-01 23:26:28 
2017-02-01 23:26:28 self = <units.parsing.test_dataloader.TestDataLoader testMethod=test_is_directory>
2017-02-01 23:26:28 
2017-02-01 23:26:28     def test_is_directory(self):
2017-02-01 23:26:28         ret = self._loader.is_directory('parsing/test_data')
2017-02-01 23:26:28 >       self.assertTrue(ret)
2017-02-01 23:26:28 E       AssertionError: False is not true
2017-02-01 23:26:28 
2017-02-01 23:26:28 test/units/parsing/test_dataloader.py:88: AssertionError
```

mattclay:
Full logs for CI failure here: https://app.shippable.com/runs/58926c20d3cc6a1000c37e01/2/console

mattclay:
@alikins I see the tests passed on the most recent commit. Is there anything else that needs to be done here, other than rebase the PR?

bcoca:
@alikins since i merged #23001 some of these tests might be obsolete or need to change

Akasurde:
superseded by #50942

PR #16515, opened 2016-06-30
Ignore broken pipe errors if the sshpass process has exited

graywulf:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (oserror32-broken-pipe ff273fe4f0) last updated 2016/06/30 04:18:58 (GMT +000)
  lib/ansible/modules/core: (detached HEAD 2f0f04437b) last updated 2016/06/30 02:51:49 (GMT +000)
  lib/ansible/modules/extras: (detached HEAD d25c487ac8) last updated 2016/06/30 02:51:59 (GMT +000)
  config file = 
```
##### SUMMARY

When password-less SSH is configured and ansible_ssh_pass is provided, the following OSError exception will occur if the sshpass process exits before the password can be written to the sshpass pipe. The situation can be reproduced easily by slowing down the parent process with a time.sleep() inserted just before the os.write().

There are two possible solutions:
1. Create a dummy SSH connection to determine if password-less SSH is configured. If configured, ignore self._play_context.password and do not use sshpass.
2. Capture the OSError exception and ignore it if it is a broken pipe and if the sshpass process has exited.

I chose the latter solution to avoid creating extra connections (performance) and to limit the amount of code change (invasiveness).

```
File ".../site-packages/ansible/executor/task_executor.py", line 120, in run
res = self.execute()
File ".../site-packages/ansible/executor/task_executor.py", line 418, in _execute
File ".../site-packages/ansible/plugins/action/normal.py", line 33, in run
results = merge_hash(results, self._execute_module(tmp=tmp, task_vars=task_vars))
File ".../site-packages/ansible/plugins/action/__init.py", line 404, in execute_module
tmp = self.make_tmp_path()
File ".../site-packages/ansible/plugins/action/__init.py", line 211, in make_tmp_path
result = self.low_level_execute_command(cmd, sudoable=False)
File ".../site-packages/ansible/plugins/action/__init.py", line 520, in low_level_execute_command
rc, stdout, stderr = self._connection.exec_command(cmd, in_data=in_data, sudoable=sudoable)
File ".../site-packages/ansible/plugins/connection/ssh.py", line 609, in exec_command
#0 = success
File ".../site-packages/ansible/plugins/connection/ssh.py", line 586, in _exec_command
File ".../site-packages/ansible/plugins/connection/ssh.py", line 361, in _run
os.write(self.sshpass_pipe[1], "{0}\n".format(to_bytes(self._play_context.password)))
OSError: [Errno 32] Broken pipe
```

This fix prevents a broken pipe exception from occurring when password-less
SSH is configured and the sshpass process exits and closes the pipe before
the password is written to the pipe.

### DIFF
--- a/lib/ansible/plugins/connection/ssh.py
+++ b/lib/ansible/plugins/connection/ssh.py
@@ -19,6 +19,7 @@
 from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
+import errno
 import fcntl
 import os
 import pipes
@@ -346,7 +347,12 @@ class Connection(ConnectionBase):
 
         if self._play_context.password:
             os.close(self.sshpass_pipe[0])
-            os.write(self.sshpass_pipe[1], "{0}\n".format(to_bytes(self._play_context.password)))
+            try:
+                os.write(self.sshpass_pipe[1], "{0}\n".format(to_bytes(self._play_context.password)))
+            except OSError as e:
+                # Ignore broken pipe errors if the sshpass process has exited.
+                if e.errno != errno.EPIPE or p.poll() is None:
+                    raise
             os.close(self.sshpass_pipe[1])
 
         ## SSH state machine


nitzmahone:
Thanks @graywulf!

arielman:
Can we use it also in ansbile version 2.1.2?

PR #16524, opened 2016-06-30
ipaddr('net') does not work: ipaddr('subnet') does

mbarcia:
##### ISSUE TYPE
- Docs Pull Request
##### ANSIBLE VERSION

```
ansible 2.0.2.0
  config file = /home/vagrant/ansible.cfg
```
##### SUMMARY

I develop and mantain several roles, and found that ipaddr('net') gave an empty string. ipaddr('subnet') instead, renders the expected result. So, changing the docs to reflect that.

### DIFF
--- a/docsite/rst/playbooks_filters_ipaddr.rst
+++ b/docsite/rst/playbooks_filters_ipaddr.rst
@@ -178,12 +178,12 @@ Internet, or if they are in private networks::
 
 You can check which values are specifically network ranges::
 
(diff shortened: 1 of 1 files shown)


mbarcia:
Didn't see that subnet was another option and thought that the documenation was just outdated. I was trying to filter a host, so ipaddr('subnet') did the job. Thanks @drybjed for clarifying it! Closing it.

Issue #16156, opened 2016-06-06
warning: setlocale: LC_CTYPE: cannot change locale (UTF-8): No such file or directory


alikins:
This is a new issue split off from the issue https://github.com/ansible/ansible/issues/16133
##### ISSUE TYPE
- Bug Report
##### ANSIBLE VERSION

```
ansible --version
ansible 2.0.2.0
  config file =
  configured module search path = Default w/o overrides
```
##### OS / ENVIRONMENT

MacOS 10.9
##### SUMMARY

As originally mentioned in https://github.com/ansible/ansible/issues/16133 there can be problems using osx terminal + ansible + linux/openssh/glibc remote systems related to the locale env vars. 

The article [OS X Terminal: -bash: warning: setlocale: LC_CTYPE: cannot change locale (UTF-8): No such file or directory Fix](http://www.cyberciti.biz/faq/os-x-terminal-bash-warning-setlocale-lc_ctype-cannot-change-locale/) references this.
##### STEPS TO REPRODUCE

Note original report was against 2.0.2, so it's possible this is already resolved in 2.1/2.1/devel.
##### EXPECTED RESULTS

The locale warnings are either avoided, or filtered out, so they don't cause module failures.
##### ACTUAL RESULTS

See https://github.com/ansible/ansible/issues/16133 for an example.

The stdout from the module can end up like:

```
['stdout': u'/etc/profile.d/lang.sh: line 19: warning: setlocale: LC_CTYPE: cannot change locale (UTF-8): No such file or directory\\r\\n{\"started\": 1, \"results_file\": \"/root/.ansible_async/177959462548.25553\", \"ansible_job_id\": \"177959462548.25553\"}\\r\\n'](url)
```

alikins:
Mentioning original submitter @Jack47

Issue #16133, opened 2016-06-04
ansible task executor should print the result information when json.load rasie an ValueError


Jack47:
##### ISSUE TYPE

- Feature Idea

Could ansible task executor print the result information when `json.load` rasie an ValueError?
It will save programer's time and make life happier.

I spent a whole day and finally solved the problem I met as illustrated below. I could solved it in minutes if ansible task executor has printed the result information when `json.load` rasie an ValueError.
##### ANSIBLE VERSION

```
ansible --version
ansible 2.0.2.0
  config file =
  configured module search path = Default w/o overrides
```
##### OS / ENVIRONMENT

MacOS 10.9
##### SUMMARY

I want to launch shell command to build docker image asynchronously, the build is success but this error occured:
`fatal: [192.168.0.1]: FAILED! => {"changed": false, "failed": true, "msg": "The async task did not return valid JSON: No JSON object could be decoded"}`

Although it explained that the async task return a invalid json, but **_I don't know the returned result**_.

I modified the `executor/task_executor.py` to print the result, and I found there is something wrong with my term's locale setting

`['stdout': u'/etc/profile.d/lang.sh: line 19: warning: setlocale: LC_CTYPE: cannot change locale (UTF-8): No such file or directory\\r\\n{\"started\": 1, \"results_file\": \"/root/.ansible_async/177959462548.25553\", \"ansible_job_id\": \"177959462548.25553\"}\\r\\n'](url)`

I solved the locale problem with the help of this [link](http://www.cyberciti.biz/faq/os-x-terminal-bash-warning-setlocale-lc_ctype-cannot-change-locale/), and this problem didn't occur.

alikins:
I think this should probably be split into two issues:

1) bug report/behavior change request for showing more useful info on the failures. 
    The needs fixing/improving, but it is not unique to (2)

2) bug report referencing http://www.cyberciti.biz/faq/os-x-terminal-bash-warning-setlocale-lc_ctype-cannot-change-locale/ 

  Not sure yet how ansible could workaround this, but it's obtuse enough to be it's own issue. If nothing else it's a good candidate for info to include in the documentation.

I'm going to leave this issue open and consider it (1) above. I'll open a new issue for (2) and reference here.

mattclay:
PR #16534 been merged and may fix your issue. The problem may go away entirely with this fix. If not, the error message "The async task did not return valid JSON" you reported should be replaced with a more informative error, allowing further troubleshooting.

[...]

Issue #15162, opened 2016-03-25
The async task did not return valid JSON    RHEL5 only


binbashroot:
##### ISSUE TYPE

- Bug Report
##### ANSIBLE VERSION

```
“ansible 2.0.1.0
  config file = /home/binbashroot/ansible-prod.cfg
  configured module search path = /usr/lib/python2.6/site-packages/ansible/modules
” 

```
##### OS / ENVIRONMENT

RHEL5/6
##### SUMMARY

Tasks using async/poll with RH5 servers fails 
##### STEPS TO REPRODUCE

```
<!--- Paste example playbooks or commands between quotes -->
```
## "
- hosts: alpha
  gather_facts: True
  become: True
  tasks:
  - debug: msg="{{ansible_distribution_version}}"
    when: ansible_os_family == "RedHat"
  - name: Test Async with rhel5 and rhel6
    shell: /bin/sleep 30
    async: 45
    poll: 7
    when: ansible_os_family == "RedHat"
    "
    
    
##### ACTUAL RESULTS

"
PLAY ***************************************************************************

TASK [setup] *******************************************************************
ok: [p1rhrep]
ok: [dgitapp01]

TASK [debug] *******************************************************************
ok: [p1rhrep] => {
    "msg": "5.11"
}
ok: [dgitapp01] => {
    "msg": "6.7"
}

TASK [Test Async with rhel5 and rhel6] *****************************************
fatal: [p1rhrep]: FAILED! => {"changed": false, "failed": true, "msg": "The async task did not return valid JSON: Extra data: line 2 column 1 - line 5 column 1 (char 115 - 244)"}
changed: [dgitapp01]

NO MORE HOSTS LEFT *************************************************************

PLAY RECAP *********************************************************************
dgitapp01                  : ok=3    changed=1    unreachable=0    failed=0
p1rhrep                    : ok=2    changed=0    unreachable=0    failed=1

"

```
<!--- Paste verbatim command output between quotes -->
```

jonathannaguin:
I have had a similar issue: host is RHEL5 and Ansible 2.0.2.0 complained about the JSON returned by Async.
I fixed upgrading my Python in the RHEL5 to Python 2.6, which contains bundled the JSON module.

mattclay:
PR #16534 been merged and may fix your issue. The problem may go away entirely with this fix. If not, the error message "The async task did not return valid JSON" you reported should be replaced with a more informative error, allowing further troubleshooting.

[...]

Issue #14568, opened 2016-02-18
async conflict with ansible_user and --become flag


bbyhuy:
##### Issue Type:
- Bug Report
##### Ansible Version:

```
ansible 2.0.0.2
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides
```
##### Ansible Configuration:

host_key_checking = false
##### Environment:

Ubuntu 14.04
##### Summary:

I think it is related to:
[https://github.com/ansible/ansible/issues/13965]

My host file specifies the `ansible_user=ubuntu`
My command: `ansible-playbook site.yml -i myhost -b -v`

My playbook essentially runs a shell script that executions a compiled program and writes out a file using `async`.

What I think is happening, as mentioned in the linked bug above, is that, async_task is being written to `~root/.ansible_async/*` and not `~ubuntu/.ansible_async/*`

So while I can see the program compiled correctly by accessing the node and seeing the output file. Ansible will throw up the `"msg": "The async task did not return valid JSON: No JSON object could be decoded"` 

I tried adding `become: yes become_user: ubuntu` but that gave me some permission errors for that particular task. Which i worked around by throwing `sudo` in front of the shell command. 
##### Steps To Reproduce:
1. Simple playbook that requires sudo, apt-get or something while also using async to writing to a protected directory (dd?).
2. Set ansible_user in the host file (anything but root)
3. execute with `ansible-playbook site.yml -i myhost -b -v`
##### Expected Results:

```
2016-02-18 13:48:34,356 p=32700 u=root |  PLAY RECAP *********************************************************************
2016-02-18 13:48:34,357 p=32700 u=root |  10.115.23.218              : ok=42   changed=25   unreachable=0    failed=0
```
##### Actual Results:

```
2016-02-18 13:38:04,760 p=29753 u=root |  ^[[0;31mfatal: [10.115.23.218]: FAILED! => {"changed": false, "failed": true, "msg": "The async task did not return valid JSON: No JSON object could be decoded"}^[[0m
2016-02-18 13:38:04,762 p=29753 u=root |  PLAY RECAP *********************************************************************
2016-02-18 13:38:04,762 p=29753 u=root |  10.115.23.218              : ok=18   changed=11   unreachable=0    failed=1
```

pshchelo:
I was experiencing the same issue and found following reason in my particular case (may be someone find it useful):

[...]

PR #16534, opened 2016-06-30
Parse async response in async action.

mattclay:
##### ISSUE TYPE

Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (async-fix c59944272d) last updated 2016/06/30 14:09:12 (GMT -700)
  lib/ansible/modules/core: (detached HEAD 2f0f04437b) last updated 2016/06/29 12:25:17 (GMT -700)
  lib/ansible/modules/extras: (detached HEAD f5b7b3b2f2) last updated 2016/06/29 12:25:20 (GMT -700)
```
##### SUMMARY

Parse async output in the async action instead of task_executor. This allows use of the same parsing code used for non-async tasks.

This eliminates problems with non-JSON data before module output causing async tasks to fail, when a non-async version of the task would succeed. One test has been added to verify this.

This should resolve issue #16156.

### DIFF
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -458,15 +458,6 @@ class TaskExecutor:
                 vars_copy[self._task.register] = wrap_var(result.copy())
 
             if self._task.async > 0:
-                # the async_wrapper module returns dumped JSON via its stdout
-                # response, so we parse it here and replace the result
-                try:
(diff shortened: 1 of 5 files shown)


mattclay:
This may resolve several other issues related to the "The async task did not return valid JSON" exception. Issues likely affected by this include:

#14568
#15162
#16133

PR #16535, opened 2016-06-30
Fix cpu facts for hosts missing fields in cpuinfo (16533)

### LINKED ISSUE #16533: "ansible_processor_count" and "ansible_processor_vcpus" reported as 0
##### ISSUE TYPE

- Bug Report
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file =
```
##### CONFIGURATION

N/A
##### OS / ENVIRONMENT

Running OS: Ubuntu 16.04 LTS x86_64
Managing OS: Ubuntu 16.04 LTS ppc64le
##### SUMMARY

On systems without vendor_id/model_name in /proc/cpuinfo, the "ansible_processor_count" and "ansible_processor_vcpus" facts are reported as 0.
##### STEPS TO REPRODUCE

```
When running Ansible and managing a host OS running on POWER/ppc64le, run:

ansible -i hosts -m setup all | grep proc

```

##### EXPECTED RESULTS

Upon fact gathering, I expected Ansible to report a number of processors/vcpus equal to what is returned by /proc/cpuinfo.

$ sudo ansible -i /home/openstack/hosts -m setup all | grep proc
        "ansible_processor": [
        "ansible_processor_cores": 1,
        "ansible_processor_count": 32,
        "ansible_processor_threads_per_core": 1,
        "ansible_processor_vcpus": 32,

$ cat /proc/cpuinfo
processor       : 0
cpu             : POWER8 (architected), altivec supported
clock           : 3359.000000MHz
revision        : 2.1 (pvr 004b 0201)

processor       : 1
cpu             : POWER8 (architected), altivec supported
clock           : 3359.000000MHz
revision        : 2.1 (pvr 004b 0201)

...

processor       : 31
cpu             : POWER8 (architected), altivec supported
clock           : 3359.000000MHz
revision        : 2.1 (pvr 004b 0201)
##### ACTUAL RESULTS

Upon fact gathering, Ansible reported zero (0) for both processors/vcpus, which is inaccurate and different what what's returned by /proc/cpuinfo. This is due to the fact that in /proc/cpuinfo on POWER there are no vendor_id or model_name flags returned unlike on x86. This means that when the current flow hits:

```
        if vendor_id_occurrence == model_name_occurrence:
            i = vendor_id_occurrence
```

at https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/facts.py#L1109-L1110

it sets i = 0, causing processor_count and processor_vcpus to be 0.

$ sudo ansible -i /home/openstack/hosts -m setup all | grep proc
        "ansible_processor": [
        "ansible_processor_cores": 1,
        "ansible_processor_count": 0,
        "ansible_processor_threads_per_core": 1,
        "ansible_processor_vcpus": 0,

cat /proc/cpuinfo
processor       : 0
cpu             : POWER8 (architected), altivec supported
clock           : 3359.000000MHz
revision        : 2.1 (pvr 004b 0201)

processor       : 1
cpu             : POWER8 (architected), altivec supported
clock           : 3359.000000MHz
revision        : 2.1 (pvr 004b 0201)

...

processor       : 31
cpu             : POWER8 (architected), altivec supported
clock           : 3359.000000MHz
revision        : 2.1 (pvr 004b 0201)

```

```

adreznec:
##### ISSUE TYPE

- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file =
```
##### SUMMARY

Adds a conditional check on vendor_id_occurence being greater than 0,
allowing linux fact gathering to correctly gather ansible_processor_count
and ansible_processor_vcpus on systems without vendor_id/model_name in
/proc/cpuinfo. This is required due to the fact that on systems without those
fields in cpuinfo, the processor_count is being set to 0 due to vendor_id_occurence
and model_name_occurence technically being the same (0). This overwrites the
correctly calculated value for processor_count from cpuinfo, causing misreported
facts for ansible_processor_count and ansible_processor_cpus.

Fixes  [#16533](https://github.com/ansible/ansible/issues/16533)

```

```

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -1106,8 +1106,10 @@ class LinuxHardware(Hardware):
             elif key == '# processors':
                 self.facts['processor_cores'] = int(data[1].strip())
 
-        if vendor_id_occurrence == model_name_occurrence:
-            i = vendor_id_occurrence
+        # Skip for platforms without vendor_id/model_name in cpuinfo (e.g ppc64le)
+        if vendor_id_occurrence > 0:
+            if vendor_id_occurrence == model_name_occurrence:
+                i = vendor_id_occurrence
 
         if self.facts['architecture'] != 's390x':
             if xen_paravirt:


bcoca:
I'm not sure this is a bugfix but a change of behavior, can you explain how/why this is a bug?

adreznec:
This does change the behavior, but the behavior itself was flawed before. For architectures like ppc64el that don't expose the vendor_id/model_name fields in /proc/cpuinfo output, the current behavior will always cause the loop value to get set to 0 because there will always be 0 occurrences of both vendor_id and model_name. This means that all values dependent on that count below (processor_count, processor_vcpus, etc) will be reported as 0 regardless of the amount of processors counted in /proc/cpuinfo before the check.

alikins:
Looks good to me.

Tested this against a couple of x86_64 and ppc64/POWER cpuinfo files from https://github.com/alikins/socket-count-test-data and it does the right thing. (and a wip unittest branch at https://github.com/alikins/ansible/tree/test_facts_cpuinfo).

PR #16538, opened 2016-07-01
Use loop_control.loop_var directly

amenonsen:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
devel@9b7d782abbfd10cd7a147b9396e3129e3ed5f2d4
```
##### SUMMARY

6eefc11c converted task.loop_control into an object, but while the other
callers were updated to use .loop_var instead of .get('loop_var'), this
site was overlooked.


amenonsen:
I went ahead and merged this because it's a trivial bugfix with a simple testcase.

PR #16539, opened 2016-07-01
Add given ip under proxy exception.

kvinod1010:
##### ISSUE TYPE

- Feature Pull Request
##### ANSIBLE VERSION

```

```
##### SUMMARY

```

```

Execution:
./proxy_override.ps1 <ipaddress>


bcoca:
please fill in the issue template as I'm not sure what this does/fix/break aside from adding an EOF character which I'm not even sure is needed.

nitzmahone:
This doesn't appear to be broadly applicable to Ansible Windows users, so it's probably not a good candidate for inclusion. 

Feel free to bring up in discussion on our Google Groups if you feel strongly otherwise- if there's significant support, we might consider it.

Thanks!

PR #16575, opened 2016-07-05
better fix for stubbing AnsibleModule

privateip:
This is a better fix for stubbing AnsibleModule to make use of fetch_url
from module_utils/urls.py


Qalthos:
Also pulled into #16636

PR #16589, opened 2016-07-06
User reports that adding PID to the name prevents parallel runs of /usr/bin/ansible-playbook from removing the tmp directory prematurely.

### LINKED ISSUE #16489: OSError: [Errno 2] No such file or directory: xxx/ziploader_cache
##### ISSUE TYPE
- Bug Report
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file = /home/ansible/infra-ansible-configuration/ansible.cfg
```
##### CONFIGURATION

```
[defaults]
host_key_checking = False
lookup_plugins = plugins/lookup/
gathering = smart
fact_caching = redis
fact_caching_timeout = 3600

[ssh_connection]
control_path = %(directory)s/%%h-%%r
ssh_args = -o ControlMaster=yes
```
##### OS / ENVIRONMENT

```
Ubuntu 14.04.4
Linux xxx 3.19.0-61-generic #69~14.04.1-Ubuntu SMP Thu Jun 9 09:09:13 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
```
##### SUMMARY

When using any module we sometimes - about half the time in fact, get the error shown below. Sometimes this will happen many times in a row until finally the run works.
##### STEPS TO REPRODUCE

Just a simple copy of a local file to remote will trigger this on my systems about half the time.

##### EXPECTED RESULTS

I expect the file to be coped over.
##### ACTUAL RESULTS

```
fatal: [REDACTED]: FAILED! => {"failed": true, "msg": "Unexpected failure during module execution.", "stdout": ""}
  File "/usr/lib/python2.7/dist-packages/ansible/executor/task_executor.py", line 124, in run
    res = self._execute()
  File "/usr/lib/python2.7/dist-packages/ansible/executor/task_executor.py", line 446, in _execute
  File "/usr/lib/python2.7/dist-packages/ansible/plugins/action/normal.py", line 33, in run
    results = merge_hash(results, self._execute_module(tmp=tmp, task_vars=task_vars))
  File "/usr/lib/python2.7/dist-packages/ansible/plugins/action/__init__.py", line 583, in _execute_module
    (module_style, shebang, module_data) = self._configure_module(module_name=module_name, module_args=module_args, task_vars=task_vars)
  File "/usr/lib/python2.7/dist-packages/ansible/plugins/action/__init__.py", line 149, in _configure_module
    (module_data, module_style, module_shebang) = modify_module(module_name, module_path, module_args, task_vars=task_vars, module_compression=self._play_context.module_compression)
  File "/usr/lib/python2.7/dist-packages/ansible/executor/module_common.py", line 718, in modify_module
    (module_data, module_style, shebang) = _find_snippet_imports(module_name, module_data, module_path, module_args, task_vars, module_compression)
  File "/usr/lib/python2.7/dist-packages/ansible/executor/module_common.py", line 586, in _find_snippet_imports
    os.mkdir(lookup_path)
OSError: [Errno 2] No such file or directory: '/home/ansible/.ansible/tmp/ansible-local-tmpvqsFYh/ziploader_cache'
```

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
2.1.0 and devel
```
##### SUMMARY

From the code I'm not sure why that would help but it doesn't do any
harm.  Proposing this pull request to make that change inside of
ansible.

Possible Fixes #16489

### DIFF
--- a/lib/ansible/constants.py
+++ b/lib/ansible/constants.py
@@ -76,7 +76,8 @@ def get_config(p, section, key, env_var, default, boolean=False, integer=False, 
             value = shell_expand(value)
             if not os.path.exists(value):
                 os.makedirs(value, 0o700)
-            value = tempfile.mkdtemp(prefix='ansible-local-tmp', dir=value)
+            prefix = 'ansible-local-%s' % os.getpid()
+            value = tempfile.mkdtemp(prefix=prefix, dir=value)
         elif ispathlist:
             if isinstance(value, string_types):
                 value = [shell_expand(x, expand_relative_paths=expand_relative_paths) \


abadger:
@MerlinDMC reports that this reduces but does not eliminate the problem.  Based on that, I do not believe this is a necessary fix.  This probably only reduces the problem due to to changing the timing of operations so as not to tickle the race condtion as frequently.  We may want to merge this in to try to help out MerlinDMC's specific use case under 2.1.1 but it should in no way be considered a fix.

bcoca:
it should remove this particular problem, PIDs are unique, even if the system was spawning procs fast enough to try to issue the same PID again it would detect it is already in use and skip it.

bcoca:
That said .. not sure what is specifically deleting these dirs, the problem with the python function is that it returns a 'random' string for the dir, which is normally good enough, but not guaranteed to be unique. 

Safe 'temp name' functions add pid/time/mac/etc to the random string to ensure uniqueness, for our use either of the first 2 should be more than enough. 

In this case the original ticket seems to be about ziploader, yet this fix seems to apply to 'any and all' tempdirs defined in configuration, I suspect there might be a race condition outside of running parallel ansible-playbooks.

PR #16604, opened 2016-07-06
With os.path.realpath it is impossible to have ec2.py as symlink in i…

ask0n:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.0.2.0
```
##### SUMMARY

With os.path.realpath it is impossible to have ec2.py as symlink in inventories folder and separate ec2.ini per inventory.
os.path.abspath instead give an ability to have common ec2.py symlinked to every inventory with its own ec2.ini config.

```
cd inventories/TEST
ln -s ../ec2.py ec2.py
touch ec2.ini
 ./ec2.py --list
  File "./ec2.py", line 1326, in <module>
    Ec2Inventory()
  File "./ec2.py", line 163, in __init__
    self.read_settings()
  File "./ec2.py", line 224, in read_settings
    configRegions = config.get('ec2', 'regions')
  File "/usr/lib/python2.7/ConfigParser.py", line 607, in get
    raise NoSectionError(section)
ConfigParser.NoSectionError: No section: 'ec2'

After changes all work as expected because ec2.ini is located near symlink to ec2.py
```

### DIFF
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -210,7 +210,7 @@ class Ec2Inventory(object):
             config = configparser.ConfigParser()
         else:
             config = configparser.SafeConfigParser()
-        ec2_default_ini_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'ec2.ini')
+        ec2_default_ini_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'ec2.ini')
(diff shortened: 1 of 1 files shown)


s-hertel:
@ask0n We are not working on adding features to the inventory scripts since adding the inventory plugins (new in Ansible 2.4). Please check out the [aws_ec2 inventory plugin](https://docs.ansible.com/ansible/latest/plugins/inventory/aws_ec2.html) which is designed to handle separate config files per inventory.

PR #163, opened 2012-04-18
Template the source file of the copy module.

jhoekx:
In a playbook, given that this works:

```

---
- hosts: all
  vars:
    port: 5150
  tasks:
  - name: test basic shell, plus two ways to dereference a variable
    action: shell echo $HOME $port {{ port }}
```

I expect this to work too:

```

---
- hosts: all
  vars:
    template: "../templates"
  tasks:
  - name: DNS configuration
    action: copy src=$template/etc/resolv.conf dest=/etc/resolv.conf
```

It currently fails because we template the arguments after we try to copy the file.

This is a quick fix, but we really need to rethink/refactor our templating strategy. We template far too often. With this fix 14 times per copy play for two nodes!

PR #16608, opened 2016-07-06
Adding to tests for GalaxyCLI and setting up testing class to allow a…

s-hertel:
##### ISSUE TYPE
- Feature Pull Request 
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file =
```
##### SUMMARY

Wrote tests for following GalaxyCLI methods: parse, run, exit_without_ignore, execute_init, execute_info, execute_install, execute_remove, execute_list, execute_search, execute_login, and execute_import.
Since some of the GalaxyCLI methods require authentication, I added a setUpClass to ask the user once for credentials to allow thorough testing. The user is able to decline this or if accepted, has the ability to decline any or all of the four following authentication prompts. The effected tests will be avoided but coverage will be affected. If the tester inputs bad credentials an error reflecting that is raised.
Fixes #163 of galaxy-issues by adding a missing decorator in GalaxyAPI. Without this the GalaxyAPI baseurl is None, and attempting to open None/Imports is unsuccessful (as is identifying what kind of url it is, resulting in a ValueError).
I added try/except statements to try to prevent failed offline tests (instead allowing them to use SkipTest, which will be clearly visible to the tester upon test completion). My thought process behind this: if testing with wifi turned off, the errors from the methods requiring internet contain the messages "Failed to get data from the API server" or "Failed to validate the SSL certificate”. So I try to run the tests, but if they fail with those specific messages we can use except and raise SkipTest (with an informative message for the tester). I originally tried to create a method to verify internet connection, but there seem to be problems verifying that depending on the platform of the tester. It seemed like an iffy hack and I am not familiar enough with a large variety of platforms, so I opted for the Pythonic strategy of asking for forgiveness rather than permission (i.e. the try/except strategy depicted above). I identified the errors raised by running the tests offline. There may be a better solution.

…uthentication for thorough testing. Also fixed a line missing in GalaxyAPI that is needed to set the baser. Fixes #163 of galaxy-issues.

### DIFF
--- a/lib/ansible/galaxy/api.py
+++ b/lib/ansible/galaxy/api.py
@@ -153,7 +153,7 @@ class GalaxyAPI(object):
             return data['results']
         return data
 
-    @g_connect   # Added this to fix a bug. Without this, self.baseurl == None.
+    @g_connect
     def get_import_task(self, task_id=None, github_user=None, github_repo=None):
         """
         Check the status of an import task.
--- a/lib/ansible/galaxy/api.py
+++ b/lib/ansible/galaxy/api.py
@@ -137,6 +137,7 @@ class GalaxyAPI(object):
         data = json.load(resp)
         return data
 
+    @g_connect
     def create_import_task(self, github_user, github_repo, reference=None):
         """
         Post an import request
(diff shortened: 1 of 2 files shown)


s-hertel:
Accidentally removed an added a line in GalaxyAPI.

s-hertel:
Closing because of size.

PR #16611, opened 2016-07-06
adds restconf transport to ios shared module

privateip:
- add restconf transport option to ios shared module
- refactor Cli to use common NetCli 
- remove NetworkModule from ios.py
- add common argument spec ios_argument_spec
- add get_config() function to module


Qalthos:
I took your changes and put them on top of the NetworkModule in #16636. Let me know if anything is missing or out of place.

PR #16613, opened 2016-07-06
Don't check that we import basic.py in new-style modules

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

Ansible 2.2
##### SUMMARY

Since Ansiballz, we no longer need to import basic directly into
a new-style module.  Some modules, like the Networking modules, may
import basic in their own module_utils files and the module will import
that specialized module_util file rather than basic.

### DIFF
--- a/lib/ansible/executor/module_common.py
+++ b/lib/ansible/executor/module_common.py
@@ -574,8 +574,6 @@ def _find_snippet_imports(module_name, module_data, module_path, module_args, ta
         if os.path.exists(cached_module_filename):
             display.debug('ZIPLOADER: using cached module: %s' % cached_module_filename)
             zipdata = open(cached_module_filename, 'rb').read()
-            # Fool the check later... I think we should just remove the check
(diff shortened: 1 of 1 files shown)


gundalow:
What do I need to do to progress this?

abadger:
@gundalow just take a look and see if it looks good to you.  I'll merge it if so.

gundalow:
Change looks sensible, however I don't know the executor code. Is Shippable enough of a test? 

Could there be modules that aren't currently tested by Shippable that may fail, due to strange imports? Maybe not given the checks by ansible-validate-module before my most recent PR. 

If you are happy with the above, then I'm happy for you to merge.

abadger:
Thanks @gundalow - yeah, I think that none of the current modules will have any problems with this code being removed.  This _should_ just allow new code to be created that doesn't need to make use of basic.py.

PR #16618, opened 2016-07-07
Refactor network and eos module_utils to use a subclass instead of factory function to create the NetworkModule

abadger:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
devel
```
##### SUMMARY

Just seems like a more OO API this way.

### DIFF
--- a/lib/ansible/module_utils/eos.py
+++ b/lib/ansible/module_utils/eos.py
@@ -19,8 +19,10 @@
 
 import re
 
-from ansible.module_utils.basic import json, get_exception, AnsibleModule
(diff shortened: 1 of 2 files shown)


privateip:
Overall these changes look fine to me.  Lets go ahead and merge them but I think we should keep a small factory function get_module() short term to return NetworkModule only to ease the transition of the modules otherwise everything breaks immediately.  We can migrate the module code (all of them are going to be touched for 2.2 anyway) and then drop the get_module() function at the end before 2.2 release.

Issue #14857, opened 2016-03-08
with_first_found will always find a task before vars in a role


JamesBelchamber:
##### Issue Type:
- Bug Report
##### Ansible Version:

```
ansible 2.0.1.0
```
##### Ansible Configuration:

None
##### Environment:

OS X
##### Summary:

When `with_first_found` is used to look up a file in `vars/` or `files/` and a file with the same name exists in `tasks/`, the file in `tasks/` will be returned instead.
##### Steps To Reproduce:

The galaxy role DavidWittman.filebeat exhibits this problem, as described in  DavidWittman/ansible-filebeat#1. This is the relevant part of the role layout:

```
├── tasks
│   ├── debian.yml
│   ├── main.yml
│   └── redhat.yml
└── vars
    ├── Debian.yml
    ├── RedHat.yml
    └── main.yml
```

And here is the play:

```
- include_vars: "{{ item }}"
  with_first_found:
    - "{{ ansible_os_family }}.yml"
    - default.yml
```

Please bear in mind that on OS X, the default filesystem is HFS+ which is case-insensitive.
##### Expected Results:

`vars/RedHat.yml` should be returned on a EL server.
##### Actual Results:

`tasks/redhat.yml` is returned and an error is reported (because it is not a vars file).

```
TASK [DavidWittman.filebeat : include_vars] ************************************
fatal: [x.x.x.x]: FAILED! => {"failed": true, "msg": "/usr/local/etc/ansible/roles/DavidWittman.filebeat/tasks/RedHat.yml must be stored as a dictionary/hash"}
```

JamesBelchamber:
This is caused by the `path_dwim_relative` function. When `with_first_found` is used within a role, the role's path is passed to `path_dwim_relative`, which (correctly) identifies it as a role and appends the role's `tasks/` directory to the search list. This means that `tasks/` is essentially searched in every iteration. The code is [here](https://github.com/ansible/ansible/blob/4657be4eabc4d3882f407624d54715d2af197142/lib/ansible/parsing/dataloader.py#L246).

I don't really know why this behaviour is desirable so I don't know how I would go about patching this bug. It's possible that the 'patch' is really to let `path_dwim_relative` handle the list of places to search (instead of the list that's iterated through [here](https://github.com/ansible/ansible/blob/26cd4650684ce4e82fb9c6c507a23dd40718198d/lib/ansible/plugins/lookup/first_found.py#L188)), and document the order of places that `with_first_found` will search by default.

[...]

PR #16630, opened 2016-07-07
fixed lookup search path

bcoca:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
2.2
```
##### SUMMARY

added ansible_search_path var that contains the proper list and in order
removed roledir var which was only used by first_found, rest used role_path
added needle function for lookups that mirrors the action plugin one, now
both types of plugins use same pathing.

Also fixes first_found in particular, to be 'aware' of task subdir requirements, this info is now available to all `with_<lookup>`.

fixes #16630

### DIFF
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -188,7 +188,18 @@ class TaskExecutor:
                     except AnsibleUndefinedVariable as e:
                         display.deprecated("Skipping task due to undefined Error, in the future this will be a fatal error.: %s" % to_bytes(e))
                         return None
-                items = self._shared_loader_obj.lookup_loader.get(self._task.loop, loader=self._loader, templar=templar).run(terms=loop_terms, variables=self._job_vars, wantlist=True)
+
+                # get lookup
+                mylookup = self._shared_loader_obj.lookup_loader.get(self._task.loop, loader=self._loader, templar=templar)
+
+                # give lookup task 'context' for subdir (mostly needed for first_found)
+                for subdir in ['tempalte', 'var', 'file']: #TODO: move this to constants?
+                    if subdir in self._task.name:
+                        break
+                setattr(mylookup,'_subdir', subdir + 's')
+
+                # run lookup
+                items = mylookup.run(terms=loop_terms, variables=self._job_vars, wantlist=True)
             else:
                 raise AnsibleError("Unexpected failure in finding the lookup named '%s' in the available lookup plugins" % self._task.loop)
 
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -83,12 +83,8 @@ class TaskExecutor:
         display.debug("in run()")
 
         try:
-            # lookup plugins need to know if this task is executing from
-            # a role, so that it can properly find files/templates/etc.
-            roledir = None
-            if self._task._role:
-                roledir = self._task._role._role_path
-            self._job_vars['roledir'] = roledir
+            # get search path for this task to pass to lookup plugins
+            self._job_vars['ansible_search_path'] = self._task.get_search_path()
 
             items = self._get_loop_items()
             if items is not None:
--- a/lib/ansible/plugins/lookup/__init__.py
+++ b/lib/ansible/plugins/lookup/__init__.py
@@ -115,6 +115,6 @@ class LookupBase(with_metaclass(ABCMeta, object)):
 
         result = self._loader.path_dwim_relative_stack(paths, subdir, needle)
         if result is None:
-            raise AnsibleError("Unable to find '%s' in expected paths." % needle)
+            raise AnsibleFileNotFound("Unable to find '%s' in expected paths." % needle)
 
         return result
--- a/lib/ansible/plugins/lookup/__init__.py
+++ b/lib/ansible/plugins/lookup/__init__.py
@@ -103,7 +103,7 @@ class LookupBase(with_metaclass(ABCMeta, object)):
         """
         pass
 
-    def find_needle(self, myvars, subdir, needle):
+    def find_file_in_search_path(self, myvars, subdir, needle):
         '''
         Return a file (needle) in the task's expected search path.
         '''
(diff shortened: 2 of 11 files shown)


lib/ansible/plugins/lookup/__init__.py
```diff
@@ -101,3 +102,19 @@ def run(self, terms, variables=None, **kwargs):
             result_string = to_unicode(result_string)
         """
         pass
+
+    def find_needle(self, myvars, subdir, needle):
```
alikins: Maybe 'def find_file_in_expected_search_path()'  or 'find_file_in_search_path()' ?

lib/ansible/plugins/lookup/first_found.py
```diff
@@ -173,22 +171,21 @@ def run(self, terms, variables, **kwargs):
...
             else:
-                if roledir is not None:
-                    # check the templates and vars directories too,if they exist
-                    for subdir in ('templates', 'vars', 'files'):
-                        path = self._loader.path_dwim_relative(roledir, subdir, fn)
-                        if os.path.exists(path):
-                            return [path]
+                # check the templates and vars directories too,if they exist
+                #TODO: This should match 'task module' (include_vars/vars, template/templates, etc)
+                for subdir in ('templates', 'vars', 'files'):
```
alikins: Could maybe use self.find_file_in_search_paths(variables, ['templates', 'vars', 'files'], fn) here.

lib/ansible/plugins/lookup/template.py
```diff
@@ -36,26 +35,25 @@ class LookupModule(LookupBase):
...
-                        searchpath.insert(1, variables['role_path'])
-
+                    # set jinja2 internal search path for includes
+                    if 'ansible_search_path' in variables:
+                        searchpath = variables['ansible_search_path']
+                    else:
+                        searchpath = [self._loader._basedir, os.path.dirname(lookupfile)]
                     self._templar.environment.loader.searchpath = searchpath
+
+                    # do the emplating
```
alikins: s/emplating/templating

lib/ansible/plugins/lookup/__init__.py
alikins: And maybe a

``` python
def find_file_in_search_paths(self, myvars, subdirs, target_name):
    results = []
    for subdir in subdirs:
          try:
               result.append(self.find_file_in_search_path(myvars, subdir, target_name))
          # could use a more specific exception here. Or let find_needle/find_file_in_search_path() return results as []
         except AnsibleError:
              continue
   return results
```

lib/ansible/plugins/lookup/fileglob.py
```diff
@@ -26,12 +26,10 @@ class LookupModule(LookupBase):
 
     def run(self, terms, variables=None, **kwargs):
 
-        basedir = self.get_basedir(variables)
-
         ret = []
         for term in terms:
             term_file = os.path.basename(term)
-            dwimmed_path = self._loader.path_dwim_relative(basedir, 'files', os.path.dirname(term))
+            dwimmed_path = self.find_needle(variables, 'files', os.path.dirname(term))
```
alikins: Could 'files' be replaced with a @property (self.file_search_path?).  For these cases it would return ['files']
or  below it would be ['templates', 'vars', 'files']

lib/ansible/plugins/lookup/__init__.py
alikins: And/or, throw some variety of IGotTooManyAnswersException if find_file_in_search_paths end up with len(result) > 1

something like:

``` python
def find_file_in_search_paths(self, myvars, subdirs, target_name):
    results = []
    for subdir in subdirs:
          try:
               result.append(self.find_file_in_search_path(myvars, subdir, target_name))
          # could use a more specific exception here. Or let find_needle/find_file_in_search_path() return results as []
         except AnsibleError:
              continue
   if  not results:
       raise NoFilesFoundInSearchPathException(target_file=target_name, subdir=subdir)

  if len(results) > 1:
      raise TooManyFilesException(target_file=target_name, subdir=subdir, results=results)
   return results
```

lib/ansible/plugins/lookup/fileglob.py
bcoca: no, this is very wrong as is, we need to select which subdir depending on task action, otherwise include_vars will try to read an template with the same name

lib/ansible/plugins/lookup/first_found.py
bcoca: no, this is very wrong as is, we need to select which subdir depending on task action, otherwise include_vars will try to read in a template with the same name

lib/ansible/plugins/lookup/fileglob.py
bcoca: could be, i just kept it as a value as we also do in action plugins

lib/ansible/plugins/lookup/__init__.py
bcoca: it cannot, as it always returns the first one found

bcoca: to rephrase, we don't care that multiple files can match, it MUST return the first one it finds, that is why search path order matters.

lib/ansible/plugins/lookup/first_found.py
bcoca: https://github.com/ansible/ansible/issues/14857, my patch does not change this issue, but is a first step. Next i need to update task_executor to pass the 'subdir' to first_found which then passes it to this function.

lib/ansible/executor/task_executor.py
```diff
@@ -188,7 +188,18 @@ def _get_loop_items(self):
                     except AnsibleUndefinedVariable as e:
                         display.deprecated("Skipping task due to undefined Error, in the future this will be a fatal error.: %s" % to_bytes(e))
                         return None
-                items = self._shared_loader_obj.lookup_loader.get(self._task.loop, loader=self._loader, templar=templar).run(terms=loop_terms, variables=self._job_vars, wantlist=True)
+
+                # get lookup
+                mylookup = self._shared_loader_obj.lookup_loader.get(self._task.loop, loader=self._loader, templar=templar)
+
+                # give lookup task 'context' for subdir (mostly needed for first_found)
+                for subdir in ['tempalte', 'var', 'file']: #TODO: move this to constants?
```
eb4x, line 196: template is misspelled.

Nosmoht:
Hi,

with this change no files in role/vars are looked up anymore. I've to rename the **vars** directory inside each role to **files**. Is this expected or a bug?

PR #16644, opened 2016-07-08
ansible su root prompt

lvshaco:
case: 
ansible all -m shell -a "whoami" --become-method=su -b -K 

su root prompt, in chinese 密码： (to match the：)


amenonsen:
This would be better written as `[:：]` rather than `(:|：)`.

bcoca:
they all look like squares to me, not sure what i would be adding, also if the change should be there or in the definition of the SU_PROMPT_LOCALIZATIONS

lvshaco:
@amenonsen [:：] match any byte in it， but：is multi byte ?
@bcoca yes, SU_PROMPT_LOCALIZATIONS is a better place, but it will change all in it，or it can be fix by change " ?: ?" to " ?:? ?"

bcoca:
this branch seems to be full of unrelated commits to the initial patch

lathama:
This pull request appears to be abandoned. Good to close.

mattclay:
Submodules are no longer used in the ansible/ansible repository.

mattclay:
@lvshaco If you're still having issues with su, please open a new issue or submit a new PR, now that the submodules have been merged into the ansible/ansible respository.

PR #16645, opened 2016-07-08
Adding a test for GalaxyCLI. Tests execute_init method.

s-hertel:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file =
```
##### SUMMARY

The execute_init method creates a directory and files (a framework) for a role. My test checks the initial creation of the framework, tests what happens if the directory already exists with and without the force flag, and then cleans up the files. This adds a few seconds more to the testing time. Because of this, I'm not sure if it would be better to mock out the method's dir/file creation and simply test that the display message was called with the expected successful message, or keep as-is to enable more thorough testing.
Since the execute_init method relies on internet connection to run, I made the design choice to use a try/except to try to prevent a test failure if the tester is running tests without internet. Without this, if the internet is not working an error is raised that doesn't make note of this possible problem and I feel it could lead to misconstruing the reason for the failed test (there might be nothing wrong with the code at all). Instead of resulting in a failure, if the error raised contains the message 'Failed to get data from the API server' (which is true for the AnsibleError raised when I try to run this test without internet) then a SkipTest will be raised instead with an informative message to the user, explaining why the test was skipped. This is a compromise since this will also occur when there isn't a valid CA certificate installed. This is noted in the SkipTest error message for the tester.

```
before:

Ran 3 tests in 0.001s

OK

after:

verifies that execute_init created a skeleton framework of a role that complies with the galaxy metadata format ... ok

Ran 4 tests in 2.745s
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ Ansible Changes By Release
 - netconf_config
 - openstack
   * os_keystone_service
+  * os_stack
 - rocketchat
 - sensu_subscription
 - smartos
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,11 @@ Ansible Changes By Release
   - ansible_totalmem (renamed to ansible_memtotal_mb, units changed to MB instead of bytes)
 
 ####New Modules:
+- asa
+  * asa_acl
+  * asa_command
+  * asa_config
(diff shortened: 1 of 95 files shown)


s-hertel:
Failed to rebase correctly.

PR #16649, opened 2016-07-08
[WIP] Ssh connection stderr parsing for better ssh errors

alikins:
##### ISSUE TYPE

- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (ssh_connection_stderr_parse_for_error 9c4b6957c9) last updated 2016/07/08 13:33:05 (GMT -400)
  lib/ansible/modules/core: (detached HEAD 799159b8ee) last updated 2016/07/07 18:09:27 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 8502d2595a) last updated 2016/07/07 18:09:27 (GMT -400)
  config file = ssh_error_tests/ansible.cfg


```
##### SUMMARY

[This is a work in progress, but looking for feedback]

This pr is a change to the ssh connection plugin that tries to improve the quality of error message when ansible fails because of underlying ssh failures.

Currently, when using 'ssh' transport, if the 'ssh' command itself fails in any fashion, the only indicator is a return code of 255. That causes all failures related to ssh config and setup to fail with the same ansible error. Examples include ssh key permission errors, connection errors, name resolution and DNS errors, and any ssh authentication related errors.

However, most of those errors do show some info about the cause of the error on stderr. This pr captures that stderr, and checks it for common errors via regex matches, and then raises exceptions that include additional info.

Changes include:
- Add an assortment of ssh specific AnsibleError subclasses
- Changing ssh plugin to raise those as needed
- Removing '-q' from the default list of args specified to the ssh command. With the -q enabled, ssh will suppress warnings and most errors. We need those errors to figure out what broke.
- misc exception plumbing to take advantage of addition info in the new exceptions.
- some tweaks to display formatting so that debug output is 'unescaped' so that embedded newlines
  and such are expanded when displaying to the console. This isn't critical, but it improves the output a lot.

Things Missing yet:

Any change in the completely unregular, unstructured, undocument, non-standard output of stderr could cause the detailed errors to miss, and potentially, to trigger in cases where they shouldnt.
- The parser is really specific to openssh output.
- The parser is really specific to openssh version.
  (That said, looking at openssh history, those messages haves haven't changed in a long time. They are also not translated or localized).
- Some support for figuring out what version of ssh is in play, and which set of parsing rules to use.
- Add support for multiple parsing rules.
- The stderr message to exception handling isn't as slick as it could be. Basically, needs a clever exception mapper.
- More info could be given to the exceptions on creation, so they can be displayed later.

Other options
- Arguably, looking for error events in the stderr output should be tied into the ssh state machine in ssh.py. The code already parses stdout/stderr for prompts to use to determine the state. The additional failure modes could be part of the state machine.

There is also a ad hoc test framework in ssh_error_tests currently (will eventually move to an integration test). This includes a ssh_config and inventory script that and test play that will connect
to testhost with a variety of ssh options and keys.

### DIFF
--- a/lib/ansible/plugins/connection/paramiko_ssh.py
+++ b/lib/ansible/plugins/connection/paramiko_ssh.py
@@ -72,6 +72,10 @@ with warnings.catch_warnings():
         pass
 
 
+class AnsibleParamikoConnectionFailure(AnsibleConnectionFailure):
+    ''' the Paramiko transport connection plugin had a fatal error '''
+
+
 class MyAddPolicy(object):
     """
     Based on AutoAddPolicy in paramiko so we can determine when keys are added
@@ -239,9 +243,9 @@ class Connection(ConnectionBase):
             elif "Private key file is encrypted" in msg:
                 msg = 'ssh %s@%s:%s : %s\nTo connect as a different user, use -u <username>.' % (
                     self._play_context.remote_user, self._play_context.remote_addr, port, msg)
-                raise AnsibleConnectionFailure(msg)
+                raise AnsibleParamikoConnectionFailure(msg)
             else:
-                raise AnsibleConnectionFailure(msg)
+                raise AnsibleParamikoConnectionFailure(msg)
 
         return ssh
 
@@ -262,7 +266,7 @@ class Connection(ConnectionBase):
             msg = "Failed to open session"
             if len(str(e)) > 0:
                 msg += ": %s" % str(e)
-            raise AnsibleConnectionFailure(msg)
+            raise AnsibleParamikoConnectionFailure(msg)
 
         # sudo usually requires a PTY (cf. requiretty option), therefore
         # we give it one by default (pty=True in ansble.cfg), and we try
--- a/lib/ansible/plugins/connection/ssh.py
+++ b/lib/ansible/plugins/connection/ssh.py
@@ -95,7 +95,8 @@ class OpenSshErrorParser():
 
     error_regexes = {r"""Permissions (.*) for '(.*)' are too open.""": 'too_open',
                      r"""Load key (\".*\"): bad permissions""": 'bad_permssions',
-                     r"""could not open key file '(.*)': (.*)""": 'could_not_open_key_file'}
+                     r"""could not open key file '(.*)': (.*)""": 'could_not_open_key_file',
+                     r"""ssh: Could not resolve hostname (.*): No address associated with hostname""": 'could_not_resolve'}
 
     def __init__(self, stderr=None):
         self.stderr = stderr
@@ -104,16 +105,15 @@ class OpenSshErrorParser():
 
     def _match_errors(self):
         for error_regex in self.error_regexes:
-            print(error_regex)
-            #print(self.stderr)
+            # print(error_regex)
+            # print(self.stderr)
             match = re.search(error_regex, self.stderr)
-            print(match)
+            # print(match)
             if not match:
                 continue
 
-
             for match_group in match.groups():
-                print('error_match %s' % match_group)
+                # print('error_match %s' % match_group)
                 self._matches[self.error_regexes[error_regex]].append(match_group)
 
     def __str__(self):
--- a/lib/ansible/plugins/connection/ssh.py
+++ b/lib/ansible/plugins/connection/ssh.py
@@ -80,23 +80,13 @@ class AnsibleSshCommandErrorConnectionFailure(AnsibleSshConnectionFailure):
 class AnsibleSshCommandErrorPermissionDenied(AnsibleSshCommandErrorConnectionFailure):
     pass
 
-# Example errors
-_example_ssh_errors = """
-@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
-@         WARNING: UNPROTECTED PRIVATE KEY FILE!          @
-@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
-Permissions 0666 for '/home/adrian/.ssh/id_rsa' are too open.
-It is required that your private key files are NOT accessible by others.
-This private key will be ignored.
-Load key "/home/adrian/.ssh/id_rsa": bad permissions
-Permission denied (publickey,gssapi-keyex,gssapi-with-mic,password).
-"""
 
 # FIXME: better enum-like
 class SshErrors:
     permission_denied = 'permission_denied'
     no_route_to_host = 'no_route_to_host'
     could_not_resolve = 'could_not_resolve'
+    connection_refused = 'connection_refused'
 
 class OpenSshStderrErrorMapper:
     # Attempt to make sense of openssh errors and warnings in hopes of generating
@@ -108,6 +98,7 @@ class OpenSshStderrErrorMapper:
                      r"""Load key (\".*\"): bad permissions""": 'bad_permssions',
                      r"""could not open key file '(.*)': (.*)""": 'could_not_open_key_file',}
     error_regexes = {r"""connect to host (.*) port (.*): No route to host""": SshErrors.no_route_to_host,
+                     r"""connect to address (.*) port (\d*): Connection refused""": SshErrors.connection_refused,
                      r"""Permission denied \((.*)\)""": SshErrors.permission_denied,
                      r"""ssh: Could not resolve hostname (.*): No address associated with hostname""":
                     SshErrors.could_not_resolve}
--- a/lib/ansible/plugins/connection/ssh.py
+++ b/lib/ansible/plugins/connection/ssh.py
@@ -19,6 +19,7 @@
 from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
+import codecs
 import errno
 import fcntl
 import os
@@ -74,7 +75,7 @@ class AnsibleSshCommandErrorConnectionFailure(AnsibleSshConnectionFailure):
         self.ssh_cmd = ssh_cmd
         self.ssh_stderr = ssh_stderr
 
-        self.message = '%s\nSTDERR: %s' % (self.message, self.ssh_stderr)
+        self.message = '%s\nSTDERR: %s' % (self.message, codecs.escape_decode(self.ssh_stderr))
 
 
 # Example errors
@@ -431,9 +432,13 @@ class Connection(ConnectionBase):
         display.v('ssh stderr: %s' % stderr)
         display.v('ssh error state: %s' % state)
         error_mapper = OpenSshStderrErrorMapper(stderr=stderr)
-        for error in error_mapper.matched_errors.values():
-            suppress_output = False
+        for error in error_mapper.errors:
+            # FIXME: pass in useful stuff to the exception constructor
+            raise AnsibleSshConnectionFailure(error)
 
+        for warning in error_mapper.warnings:
+            display.v(msg='ssh warning: %s' % warning)
+            raise AnsibleSshConnectionFailure(warning)
 
 
     def _run(self, cmd, in_data, sudoable=True):
@@ -765,8 +770,17 @@ class Connection(ConnectionBase):
         if return_code != 255:
             return False
 
+        display.v('ssh stderr: %s' % stderr)
+        error_mapper = OpenSshStderrErrorMapper(stderr=stderr)
+        for error in error_mapper.errors:
+            # FIXME: pass in useful stuff to the exception constructor
+            raise AnsibleSshConnectionFailure(error)
 
-        raise AnsibleSshConnectionFailure("Failed to connect to the host via ssh.")
+        for warning in error_mapper.warnings:
+            display.v(msg='ssh warning: %s' % warning)
+            raise AnsibleSshConnectionFailure(warning)
+
+        return False
 
     def put_file(self, in_path, out_path):
         ''' transfer a file from local to remote '''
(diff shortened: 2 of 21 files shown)


lib/ansible/plugins/connection/paramiko_ssh.py
```diff
@@ -72,6 +72,10 @@
         pass
 
 
+class AnsibleParamikoConnectionFailure(AnsibleConnectionFailure):
```
bcoca, line 75: I dont see the need for a 'connection module specific' error class, the base connection failure works well enough and if we ever need the connection type, we should already have it in the caller.

lib/ansible/plugins/connection/ssh.py
```diff
@@ -44,6 +48,119 @@
 SSHPASS_AVAILABLE = None
 
 
+class AnsibleSshConnectionFailure(AnsibleConnectionFailure):
+    ''' the ssh connection plugin had a fatal error '''
+    pass
+
+
+# Could track retry attempts here. A little odd.
+class AnsibleAttemptedSshConnectionFailure(AnsibleConnectionFailure):
```
bcoca, line 57: same as above, this seems uneccessary

lib/ansible/plugins/connection/ssh.py
```diff
@@ -44,6 +48,119 @@
...
+    ''' the ssh connection plugin had a fatal error but may attempt to retry '''
+    pass
+
+class AnsibleSshInitialWriteConnectionFailure(AnsibleSshConnectionFailure):
+    ''' the ssh connection plugin had a fatal error during the initial data write '''
+    pass
+
+
+class AnsibleSshCommandErrorConnectionFailure(AnsibleSshConnectionFailure):
```
bcoca, line 66: these seem more valid, i would just drop the Ssh part and make them specific 'connection failure causes' but not per connection plugin

lib/ansible/plugins/connection/ssh.py
```diff
@@ -44,6 +48,119 @@
...
+    permission_denied = 'permission_denied'
+    no_route_to_host = 'no_route_to_host'
+    could_not_resolve = 'could_not_resolve'
+
+class OpenSshStderrErrorMapper:
+    # Attempt to make sense of openssh errors and warnings in hopes of generating
+    # more useful ansible error messages
+
+    # FIXME: The names on the rhs are just placeholders for now, to be replaced with exceptions
+    warning_regexes = {r"""Permissions (.*) for '(.*)' are too open.""":
```
bcoca, line 96: i18n?

alikins:
some examples:

devel with a ssh key that can't be read:

```
[fedora-23:ansible (devel % u=)]$ SSH_AUTH_SOCK= ansible 'cantread' -m ping
cantread | UNREACHABLE! => {
    "changed": false, 
    "unreachable": true
}
```

this pr:

```
[fedora-23:ansible (ssh_connection_stderr_parse_for_error %)]$ SSH_AUTH_SOCK= ansible 'cantread' -m ping
cantread | UNREACHABLE! => {
    "_connection_failure_excection_message": "Permission denied (publickey,gssapi-keyex,gssapi-with-mic,password).
", 
    "_connection_failure_exception_type": "<class 'ansible.plugins.connection.ssh.AnsibleSshCommandErrorPermissionDenied'>", 
    "_connection_failure_exception_value": "Permission denied (publickey,gssapi-keyex,gssapi-with-mic,password).
", 
    "changed": false, 
    "msg": "Permission denied (publickey,gssapi-keyex,gssapi-with-mic,password).
", 
    "unreachable": true
}
```

A ssh private key has perms that are too open:

```
[fedora-23:ansible (devel % u=)]$ SSH_AUTH_SOCK= ansible 'tooopen' -m ping
tooopen | UNREACHABLE! => {
    "changed": false, 
    "unreachable": true
}
```

this pr  (with no '-v's):

```
[fedora-23:ansible (ssh_connection_stderr_parse_for_error %)]$ SSH_AUTH_SOCK= ansible 'tooopen' -m ping
tooopen | UNREACHABLE! => {
    "_connection_failure_excection_message": "Permission denied (publickey,gssapi-keyex,gssapi-with-mic,password).
", 
    "_connection_failure_exception_type": "<class 'ansible.plugins.connection.ssh.AnsibleSshCommandErrorPermissionDenied'>", 
    "_connection_failure_exception_value": "Permission denied (publickey,gssapi-keyex,gssapi-with-mic,password).
", 
    "changed": false, 
    "msg": "Permission denied (publickey,gssapi-keyex,gssapi-with-mic,password).
", 
    "unreachable": true
}
```

this pr with one '-v':

```
[fedora-23:ansible (ssh_connection_stderr_parse_for_error %)]$ SSH_AUTH_SOCK= ansible -v 'tooopen' -m pingUsing ssh_error_tests/ansible.cfg as config file
return_code=255
stderr=@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
@         WARNING: UNPROTECTED PRIVATE KEY FILE!          @
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
Permissions 0664 for '/home/adrian/src/ansible/ssh_error_tests/id_rsa_testhost_too_open' are too open.
It is required that your private key files are NOT accessible by others.
This private key will be ignored.
Load key "/home/adrian/src/ansible/ssh_error_tests/id_rsa_testhost_too_open": bad permissions
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
@         WARNING: UNPROTECTED PRIVATE KEY FILE!          @
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
Permissions 0664 for '/home/adrian/src/ansible/ssh_error_tests/id_rsa_testhost_too_open' are too open.
It is required that your private key files are NOT accessible by others.
This private key will be ignored.
Load key "/home/adrian/src/ansible/ssh_error_tests/id_rsa_testhost_too_open": bad permissions
Permission denied (publickey,gssapi-keyex,gssapi-with-mic,password).

ssh stderr: @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
@         WARNING: UNPROTECTED PRIVATE KEY FILE!          @
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
Permissions 0664 for '/home/adrian/src/ansible/ssh_error_tests/id_rsa_testhost_too_open' are too open.
It is required that your private key files are NOT accessible by others.
This private key will be ignored.
Load key "/home/adrian/src/ansible/ssh_error_tests/id_rsa_testhost_too_open": bad permissions
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
@         WARNING: UNPROTECTED PRIVATE KEY FILE!          @
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
Permissions 0664 for '/home/adrian/src/ansible/ssh_error_tests/id_rsa_testhost_too_open' are too open.
It is required that your private key files are NOT accessible by others.
This private key will be ignored.
Load key "/home/adrian/src/ansible/ssh_error_tests/id_rsa_testhost_too_open": bad permissions
Permission denied (publickey,gssapi-keyex,gssapi-with-mic,password).

tooopen | UNREACHABLE! => {
    "_connection_failure_excection_message": "Permission denied (publickey,gssapi-keyex,gssapi-with-mic,password).
", 
    "_connection_failure_exception_type": "<class 'ansible.plugins.connection.ssh.AnsibleSshCommandErrorPermissionDenied'>", 
    "_connection_failure_exception_value": "Permission denied (publickey,gssapi-keyex,gssapi-with-mic,password).
", 
    "changed": false, 
    "msg": "Permission denied (publickey,gssapi-keyex,gssapi-with-mic,password).
", 
    "unreachable": true
}
```

(The dupes are from debug code still in there, goal would be to show ssh stderr once)

Name resolution error in devel:

```
[fedora-23:ansible (devel % u=)]$ SSH_AUTH_SOCK= ansible 'namefailure' -m ping
namefailure | UNREACHABLE! => {
    "changed": false, 
    "unreachable": true
}
```

in this pr:

```
[fedora-23:ansible (ssh_connection_stderr_parse_for_error %)]$ SSH_AUTH_SOCK= ansible 'namefailure' -m ping
namefailure | UNREACHABLE! => {
    "_connection_failure_excection_message": "ssh: Could not resolve hostname namefailure: No address associated with hostname
", 
    "_connection_failure_exception_type": "<class 'ansible.plugins.connection.ssh.AnsibleSshCommandErrorConnectionFailure'>", 
    "_connection_failure_exception_value": "ssh: Could not resolve hostname namefailure: No address associated with hostname
", 
    "changed": false, 
    "msg": "ssh: Could not resolve hostname namefailure: No address associated with hostname
", 
    "unreachable": true
}
```

lib/ansible/plugins/connection/ssh.py
```diff
@@ -624,6 +747,33 @@ def exec_command(self, *args, **kwargs):
 
         return return_tuple
 
+    def _check_for_errors_in_return_results(self, return_code, stdout, stderr):
```
bcoca, line 741: so this function seems to be most of the 'clarification' sadly it uses message scraping, which is already very fragile, but also subject to change due to language locales.

lib/ansible/plugins/connection/paramiko_ssh.py
alikins, line 75: Initial motivation for the more specific exceptions in general was time spent attempting to track down a connection error that could have been raised from multiple places. [AnsibleConnectionFailure('SSH Error: data could not be sent to the remote host. Make sure this host can be reached over ssh')] 

Generally my preference is to raise the most specific exception I can, especially if there is a reasonable expectation that the exception could be handled differently than it's base exception. But even if it's only to convey additional context its useful.

Granted, at the moment, this module or the app doesn't make any use of this specific exception (aside from the places the exception name is exposed to the user). 

Conceptually, the app could make use of it (for example, a smart connection type could allow a failed paramiko connection to be retried as a ssh connection or vice versa). Long term the connection type specific exceptions will be useful in nested or stacked connection scenarios.

lib/ansible/plugins/connection/ssh.py
alikins, line 66: I kind of debated if it made sense to share 'ssh connection exceptions' across paramiko/ssh.  In a lot of ways, there are some potential failures modes that are more conceptually 'ssh protocol' related than specific implementations (a failure in the crypto handshake setting up the transport layer for example).

The reasons I went for connection plugin specific exceptions were:
- A shared exception would need to live in either:
  - base ansible (lib/ansible/errors/)
  - connection base (plugin/connection/**init** for ex)   - either of these seems like it's pushing ssh specific concepts when it seems like there has been significant effort to remove ssh'isms from the app/lib
  - something new (plugin/connection/base_ssh.py ?)  - maybe makes more sense but that approach doesnt seem common for plugins  [plugins/strategy and plugins/shell/fish.py and plugins/action/net_template do however]
- While something like an auth or transport failure is a general ssh concept, in practice the causes of the failures (and the potential remedies) tend to be implementation specific. Or tied to the implementation specific configuration or setup. 

Given those constraints, I pushed the more specific exceptions into the connection plugins.

It may make sense to push some slightly more specific base connection exceptions into lib/ansible/errors.py.
Other connection failure modes that exist across multiple connection types.

(psuedo-code sketch)

``` python
class AnsibleConnectionTransportFailure(AnsibleConnectionFailure):
    pass

class AnsibleConnectionAuthenticationFailure(AnsibleConnectionFailure):
   pass

# not a great name, but ssh channels, pipes/sockets (to dockerd  or popen comm for local for ex), 
class AnsibleConnectionChannelFailure(AnsibleConnectionFailure):
  pass

# Also not great names here, but something to separate failures for:
#  'the local command I run to make the connection failed'                  #  ssh, docker
#     and 
# 'the command I have to to run on the remote end as part of the connection failed'    #  "jail connection requires dd command in the jail" ,  no python, no powershell
class AnsibleConnectionCommandFailure(AnsibleConnectionFailure):
  pass

class AnsibleCommandConnectionFailure(AnsibleConnectionFailure):
  pass

etc
```

lib/ansible/plugins/connection/ssh.py
alikins, line 96: The openssh messages searched for here are not translated. (afaict, no openssh messages are translated and no message catalogs are included). 

If a different implementation did translate the message (or just used different messages), the error mapper would fail to find a more specific cause and would fall back to the existing errors. rc=255 will always be a failure, it's just a matter if we can detect a more specific reason. Ditto for the messages that may be libc/os
specific (errno, 'No route to host' etc).

No doubt the regexes will need tuning and updating over time.

lib/ansible/plugins/connection/ssh.py
alikins, line 741: Yup, very fragile. But given the myriad ways the ssh command can fail, we need to try to provide as much info to users as possible. If we can enhance the context provided with error messages in this scenario (where rc=255 and the ssh connection has failed without a doubt[1]) it should help cut down on user frustration. Especially given that current versions hide all of the ssh 

[1] Or more accurately, that the exec of the ssh client failed somehow, but it's about all we've got.

alikins:
shrug

PR #16647, opened 2016-07-08
Adding a test for GalaxyCLI and two methods on which it depends. Test…

s-hertel:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file =
  configured module search path = Default w/o overrides
```
##### SUMMARY

This tests the execute_info method which provides the user with information associated with a role. The execute_info method relies on having an installed role (otherwise there is no info to display). As the tester may not have a role installed this requires a setup and cleanup to test. Installing a role can be done with a local tar.gz file and is more simple for testing than trying to verify internet connection, determine what platform the tester is on, and download an acceptable role. I created two other methods to make this tar.gz file with the bare minimum needed to install a role successfully. I then test that the information provided during installation is made available when execute_info is called.

```
before:

shertel-OSX:ansible shertel$ PYTHONPATH=./lib nosetests -d -w test/units -v -s cli.test_galaxy
test_display_galaxy_info (cli.test_galaxy.TestGalaxy) ... ok
test_display_min (cli.test_galaxy.TestGalaxy) ... ok
test_init (cli.test_galaxy.TestGalaxy) ... ok

----------------------------------------------------------------------
Ran 3 tests in 0.001s

OK

after:

shertel-OSX:ansible shertel$ PYTHONPATH=./lib nosetests -d -w test/units -v -s cli.test_galaxy
test_display_galaxy_info (cli.test_galaxy.TestGalaxy) ... ok
test_display_min (cli.test_galaxy.TestGalaxy) ... ok
testing that execute_info displays information associated with a role ... [DEPRECATION WARNING]: The comma separated role spec format, use the
yaml/explicit format instead..
This feature will be removed in a future
release. Deprecation warnings can be disabled by setting
deprecation_warnings=False in ansible.cfg.
ok
test_init (cli.test_galaxy.TestGalaxy) ... ok

----------------------------------------------------------------------
Ran 4 tests in 0.035s

OK
```

…s execute_info method.

PR #16650, opened 2016-07-08
 Adding a test for GalaxyCLI and two methods on which it depends. Tests execute_install method.

s-hertel:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file =
```
##### SUMMARY

This tests the execute_install method which installs a role given a local tar.gz file containing necessary data. My test would be more thorough if I tested installing a role from galaxy.ansible.com, but I am not sure if there is a good solution for that yet (the platform of the tester needs to be determined and an role searched for). I included a # TODO comment in my code so the possibility of improving this test isn't overlooked in the future.
Like my test for execute_info (see #16647) the execute_install method needs to setup a local tar.gz file for installing a role. The two methods this test relies on are the same as the two that my test for execute_info require.

```
before:

Ran 3 tests in 0.001s

OK

after:

test_execute_install (cli.test_galaxy.TestGalaxy) ... ok

Ran 4 tests in 0.024s

OK
```

### DIFF
--- a/test/units/cli/test_galaxy.py
+++ b/test/units/cli/test_galaxy.py
@@ -23,6 +23,14 @@ from ansible.compat.six import PY3
 from ansible.compat.tests import unittest
 
 from nose.plugins.skip import SkipTest
+import ansible
+import os
+import shutil
+import tarfile
+
+from mock import patch, call
+
+from ansible.errors import AnsibleError
 
 if PY3:
     raise SkipTest('galaxy is not ported to be py3 compatible yet')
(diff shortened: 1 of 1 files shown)


s-hertel:
Will revise: instead of a tarball, will store files as artifacts in a fixtures directory, or that setup() generates a role based on the code underlying "ansible-galaxy init".

s-hertel:
Closed in favor of PR #16679

PR #16653, opened 2016-07-08
[WIP] Adding a test for GalaxyCLI. Tests execute_search method.

s-hertel:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file =
```
##### SUMMARY

The execute_search method uses internet; to always verify its behavior regardless of the tester's connection, I used mock.patch to ensure the appropriate methods are called (and with the right parameter where applicable). 
In addition, I made the design choice to use a try/except to allow an additional possible test using internet. If try/except is not used and the internet is not working an error is raised that doesn't make note of this possible problem and I feel it could lead to misconstruing the reason for the failed test (there might be nothing wrong with the code at all). Instead of resulting in a failure, if the error raised contains the message 'Failed to get data from the API server' (which is true for the AnsibleError raised when I try to run this test without internet) then a SkipTest will be raised instead with an informative message to the user, explaining why the test was skipped. This is a compromise since this will also occur when there isn't a valid CA certificate installed. This is noted in the helpful SkipTest error message for the tester. 

```
before:

Ran 3 tests in 0.001s

OK

after:

test_execute_search (cli.test_galaxy.TestGalaxy) ... ok

Ran 4 tests in 5.383s

OK
```

### DIFF
--- a/test/units/cli/test_galaxy.py
+++ b/test/units/cli/test_galaxy.py
@@ -23,6 +23,12 @@ from ansible.compat.six import PY3
 from ansible.compat.tests import unittest
 
 from nose.plugins.skip import SkipTest
+import ansible
+
+from mock import patch
+
+from ansible.errors import AnsibleError
+from ansible.module_utils.urls import SSLValidationError
 
 if PY3:
     raise SkipTest('galaxy is not ported to be py3 compatible yet')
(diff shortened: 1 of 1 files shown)


s-hertel:
Not a good thing to unit test.

PR #16660, opened 2016-07-09
Enable color output from integration tests in Docker containers

gtrafimenkov:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.0.2.0
  config file = 
```
##### SUMMARY

Enable color output from integration tests in Docker containers when then output is attached to the a terminal.  It is much easier to see results of the test this way.

### DIFF
--- a/test/utils/shippable/integration.sh
+++ b/test/utils/shippable/integration.sh
@@ -67,8 +67,12 @@ if [ "${controller_shared_dir}" ]; then
     cp -a "${SHIPPABLE_BUILD_DIR}" "${controller_shared_dir}"
 fi
 
+# Enabling colored output in Shippable console
+color_settings="--env ANSIBLE_FORCE_COLOR=1"
+
 httptester_id=$(docker run -d "${http_image}")
 container_id=$(docker run -d \
+    ${color_settings} \
     -v "/sys/fs/cgroup:/sys/fs/cgroup:ro" \
     -v "${host_shared_dir}:${test_shared_dir}" \
     --link="${httptester_id}:ansible.http.tests" \
--- a/test/utils/shippable/integration.sh
+++ b/test/utils/shippable/integration.sh
@@ -71,12 +71,9 @@ if [ "${controller_shared_dir}" ]; then
     cp -a "${SHIPPABLE_BUILD_DIR}" "${controller_shared_dir}"
 fi
 
-# Enabling colored output in Shippable console
-color_settings="--env ANSIBLE_FORCE_COLOR=$force_color"
-
 httptester_id=$(docker run -d "${http_image}")
 container_id=$(docker run -d \
-    ${color_settings} \
+    --env "ANSIBLE_FORCE_COLOR=${force_color}" \
     -v "/sys/fs/cgroup:/sys/fs/cgroup:ro" \
     -v "${host_shared_dir}:${test_shared_dir}" \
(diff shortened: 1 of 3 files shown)


jctanner:
@mattclay does this look okay to you? I am in favor of merging.

gundalow:
@mattclay Would a similar patch also need applying to https://github.com/ansible/ansible/blob/devel/test/utils/shippable/integration.sh assuming Shippable’s test console handle termcolor?

mattclay:
@gtrafimenkov Sorry, I saw this earlier then forgot to comment. As @gundalow mentioned, it would be good to include this for shippable as well, since we're using it instead of Travis.

Would you mind trying to implement it there as well?

gtrafimenkov:
Yes, Shippable's test console can handle colors.  I've made necessary changes in the `integration.sh` script.

It looks awesome: https://db.tt/sw4Pg8RA

test/utils/shippable/integration.sh
```diff
@@ -67,8 +67,12 @@ if [ "${controller_shared_dir}" ]; then
     cp -a "${SHIPPABLE_BUILD_DIR}" "${controller_shared_dir}"
 fi
 
+# Enabling colored output in Shippable console
+color_settings="--env ANSIBLE_FORCE_COLOR=1"
```
mattclay: Forcing color output causes causes the `blocks` integration test to fail. You can test this with:

`IMAGE=ansible/ansible:ubuntu1604 TARGET=blocks test/utils/shippable/integration.sh`

This is due to the test using grep on the playbook output. It probably wouldn't hurt to modify the test to work with or without color output.

I do like having the option to force color output, as it makes it possible to have color when running tests locally and saving them to a log file. I think the ideal solution would be to have force color the default, but being able to disable it when running integration.sh locally. You'll have to fix the tests to work with color though.

gtrafimenkov:
@mattclay thank you for your feedback.  I've made the necessary changes.  All tests are passed now.

test/utils/shippable/integration.sh
```diff
@@ -67,8 +71,12 @@ if [ "${controller_shared_dir}" ]; then
     cp -a "${SHIPPABLE_BUILD_DIR}" "${controller_shared_dir}"
 fi
 
+# Enabling colored output in Shippable console
+color_settings="--env ANSIBLE_FORCE_COLOR=$force_color"
+
 httptester_id=$(docker run -d "${http_image}")
 container_id=$(docker run -d \
+    ${color_settings} \
```
mattclay: Inline `--env ANSIBLE_FORCE_COLOR=$force_color` here. Doing so avoids an extra variable and allows this script to pass shellcheck.

mattclay:
@gtrafimenkov This makes the build output so much easier to read, both locally and on shippable. Thanks for taking the time to work on this. This will be ready to merge once you make the change to allow shellcheck to pass on integration.sh (see my previous comment).

test/utils/shippable/integration.sh
mattclay: Correction, use `--env "ANSIBLE_FORCE_COLOR=${force_color}"`.

gtrafimenkov:
@mattclay Thank you.  I am glad to help.

gundalow:
Excellent, thanks for this

PR #16678, opened 2016-07-11
cache the check for controlpersist on the controlhost

### LINKED ISSUE #15844: default transport (smart) performance hit
This is probably a tiny part of Issue #12239
##### ISSUE TYPE
- Bug Report
##### ANSIBLE VERSION

```
ansible 2.0.2.0
```

also latest dev

```
2.2.0-0.git201605121902.ae09648.devel
```
##### CONFIGURATION

transport left to default, or set explicitly transport=smart
##### OS / ENVIRONMENT

N/A
##### SUMMARY

When transport is "smart", ansible checks to see if it can use ControlPersist at the start of every single task before setting the type to "ssh". This causes a noticeable performance hit.

If you explicitly set it to "ssh" it does not do this.

It should probably test this once, at the start of the play, or during setup.
##### STEPS TO REPRODUCE

```
# ansible.cfg
[defaults]
# transport=smart
```

vs

```
#ansible.cfg
[defaults]
transport=ssh
```

(everything else default, no pipelining, etc)

simple playbook with 10 minimal tasks  "command: echo"
##### EXPECTED RESULTS

Ansible be really fast.
(with transport=ssh)

```
real    0m13.252s
user    0m1.646s
sys 0m1.606s
```
##### ACTUAL RESULTS

Ansible be really not fast.
(with transport=smart)

```
real    0m15.548s
user    0m1.992s
sys 0m1.878s
```

Ok, so it's only a few % increase, but it does help!

jctanner:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
2.2
```
##### SUMMARY

Fixes #15844

### DIFF
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -625,10 +625,8 @@ class TaskExecutor:
                 conn_type = "paramiko"
             else:
                 # see if SSH can support ControlPersist if not use paramiko
-                has_cp = True
-                if C.DEFAULT_CACHE_CP_CHECK:
-                    has_cp = variables['hostvars']['localhost'].get('ansible_controlpersist')
-                else:
+                has_cp = variables['hostvars']['localhost'].get('ansible_controlpersist')
+                if has_cp is None:
                     has_cp = check_for_controlpersist()
                 if not has_cp:
                     conn_type = "paramiko"
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -37,6 +37,7 @@ from ansible.template import Templar
 from ansible.utils.encrypt import key_for_hostname
 from ansible.utils.listify import listify_lookup_plugin_terms
 from ansible.utils.unicode import to_unicode, to_bytes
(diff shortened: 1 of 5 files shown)


bcoca:
for bikeshedding sake, a better name might be `ansible_ssh_enable_controlpersist`

lib/ansible/executor/task_executor.py
```diff
@@ -624,13 +625,10 @@ def _get_connection(self, variables, templar):
...
                 # see if SSH can support ControlPersist if not use paramiko
-                try:
-                    cmd = subprocess.Popen(['ssh','-o','ControlPersist'], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-                    (out, err) = cmd.communicate()
-                    err = to_unicode(err)
-                    if u"Bad configuration option" in err or u"Usage:" in err:
-                        conn_type = "paramiko"
-                except OSError:
+                has_cp = variables['hostvars']['localhost'].get('ansible_controlpersist')
+                if has_cp is None:
```
bcoca, line 629: so true/false or any value disables check?

PR #16684, opened 2016-07-12
Add a function to check for killed processes

jctanner:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
2.2
```
##### SUMMARY

The "terminated" property doesn't catch sigkills, so this adds a function to check for any killed threads and exit early if exists.


chrismeyersfsu:
Verified, this PR works. Ansible parent process is aware of children killed with `SIGKILL` and exits.

Please extend this PR to handle the case where children are killed via `SIGTERM`. The Ansible parent process should exit in the same way.

Issue #4177, opened 2013-09-19
Ansible 1.3 - Destination directory doesn't exist, but actually does.


maxamillion:
For some reason when I have a fresh install of Fedora 19 (fully updated) and I run my playbook https://github.com/maxamillion/ansible-openshift_origin it gets to the step in the output below and claims the diretory doesn't exist, I ssh into the machine and verify it does exist and then run the playbook again and it works as expected. It's strange, but I can reproduce it (I launch fresh machines often for testing).

```
TASK: [Configure mcollective broker plugin] *********************************** 
<192.168.122.10> ESTABLISH CONNECTION FOR USER: root
<192.168.122.10> EXEC ['ssh', '-tt', '-q', '-o', 'ControlMaster=auto', '-o', 'ControlPersist=60s', '-o', 'ControlPath=/home/admiller/.ansible/cp/ansible-ssh-%h-%p-%r', '-o', 'Port=22', '-o', 'KbdInteractiveAuthentication=no', '-o', 'PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey', '-o', 'PasswordAuthentication=no', '-o', 'User=root', '-o', 'ConnectTimeout=10', '192.168.122.10', "/bin/sh -c 'mkdir -p $HOME/.ansible/tmp/ansible-1379600720.02-161234373739939 && echo $HOME/.ansible/tmp/ansible-1379600720.02-161234373739939'"]
[...]

PR #16692, opened 2016-07-13
Don't clobber backup, fixes: #4177 ( https://github.com/ansible/ansib…

sigio:
Fix for issue #4177 (ansible-modules-core)
https://github.com/ansible/ansible-modules-core/issues/4177


bcoca:
i would instead increment a sequence and create another backup

sigio:
Fixed (differently) in https://github.com/ansible/ansible-modules-core/issues/4177

PR #16701, opened 2016-07-13
GalaxyCLI unit tests - test execute_install no role and using a tar file as a role

s-hertel:
##### ISSUE TYPE
- Test Pull Request
##### ANSIBLE VERSION

```
2.4.0
```
##### SUMMARY

Tests installing a role using a tar file and tests a failure is raised if a role isn't specified

### DIFF
--- a/test/units/cli/test_galaxy.py
+++ b/test/units/cli/test_galaxy.py
@@ -89,7 +89,6 @@ class TestGalaxy(unittest.TestCase):
         if os.path.isdir(cls.role_path):
             shutil.rmtree(cls.role_path)
 
-
     def setUp(self):
         self.default_args = []
 
--- a/test/units/cli/test_galaxy.py
+++ b/test/units/cli/test_galaxy.py
@@ -89,6 +89,7 @@ class TestGalaxy(unittest.TestCase):
         if os.path.isdir(cls.role_path):
             shutil.rmtree(cls.role_path)
 
+
     def setUp(self):
         self.default_args = []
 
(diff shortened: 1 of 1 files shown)


test/units/cli/test_galaxy.py
```diff
@@ -115,6 +117,32 @@ def test_display_galaxy_info(self):
         if display_result.find('\t\tgalaxy_tags:') > -1:
             self.fail('Expected galaxy_tags to be indented twice')
 
+    def test_execute_install(self):
+        # testing installing with insufficient information
+        gc = GalaxyCLI(args=["install"])
+        with patch('sys.argv', ["-c"]):
```
alikins: I think this indicates GalaxyCLI is doing something odd with it's args. I would expect passing args to GalaxyCLI to pass the value of args to GalaxyCLI's self.parser.parse_args instead of of calling it as self.parser.parse_args() which defaults to looking at sys.argv.

ie, change GalaxyCLI.parse to take args

``` python
def parse(self, args=None):
```

and then have GalaxyCLI.parse call:

``` python
self.options, self.args =self.parser.parse_args(args)
```

Then the test would just pass ['install', '-c'] as args instead of having to patch sys.argv so the testing
is easier and less likely to cause odd side effects, especially if anything else is looking at sys.argv (like nosetest
itself for example).

If the top level app also passes in a copy of sys.argv instead of a reference (ie, GalaxyCLI(args=sys.argv[:])), it avoids problems potentially caused by app code manipulating sys.argv.

That said, it's been doing that a long time, so doesn't need to be changed for this pr, but it is an example of a case where changing the code to be more testable is useful.

test/units/cli/test_galaxy.py
```diff
@@ -115,6 +117,32 @@ def test_display_galaxy_info(self):
...
+            galaxy_parser = gc.parse()
+        self.assertRaises(AnsibleError, gc.run)
+
+        # installing role
```
alikins: I would split these ('install' and 'install --offline -p -r etc') into two separate tests so any failures are more specific and granular.

test/units/cli/test_galaxy.py
```diff
@@ -115,6 +117,32 @@ def test_display_galaxy_info(self):
...
+        gc.api = ansible.galaxy.api.GalaxyAPI(gc.galaxy)
+        completed_task = gc.execute_install()
+
+        # testing correct installation
+        role_file = os.path.join(self.role_path, self.role_name)
+        self.assertTrue(os.path.exists(role_file))
+        self.assertTrue(completed_task == 0)
+
+        # cleaning up
+        gc.args = ["remove"]
```
alikins: Preference would be for a unit test to be mocked sufficiently that a cleanup like this is not required.

However, in this case if that involves a large amount of fragile mocking, I wouldn't worry about it now
as having test coverage is better than not having it, even if it is a little messy.

test/units/cli/test_galaxy.py
```diff
@@ -115,6 +117,32 @@ def test_display_galaxy_info(self):
...
+        with patch('sys.argv', ["-c"]):
+            galaxy_parser = gc.parse()
+        self.assertRaises(AnsibleError, gc.run)
+
+        # installing role
+        gc = GalaxyCLI(args=["install"])
+        with patch('sys.argv', ["--offline", "-p", self.role_path, "-r", self.role_req]):
+            galaxy_parser = gc.parse()
+        super(GalaxyCLI, gc).run()
+        gc.api = ansible.galaxy.api.GalaxyAPI(gc.galaxy)
```
alikins: Not sure I understand this. I would expect gc.run() here (maybe with gc.action set)

PR #16704, opened 2016-07-14
A little unittest refactoring

abadger:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
devel
```
##### SUMMARY
- Add a class decorator to generate tests when using a unittest.TestCase base class
- Add a TestCase subclass with setUp() and tearDown() that sets up
  module parameter parsing
- Move test_safe_eval to use the class decorator and ModuleTestCase base
  class
- Move testing of set_mode_if_different into its own file.  Someday we
  will split this into several test methods to follow best practices.

### DIFF
--- a/test/units/module_utils/basic/test_safe_eval.py
+++ b/test/units/module_utils/basic/test_safe_eval.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# (c) 2015, Toshio Kuratomi <tkuratomi@ansible.com>
+# (c) 2015-2016, Toshio Kuratomi <tkuratomi@ansible.com>
 #
 # This file is part of Ansible
 #
--- a/test/units/module_utils/basic/test_safe_eval.py
+++ b/test/units/module_utils/basic/test_safe_eval.py
@@ -24,47 +24,72 @@ import sys
 import json
 
 from ansible.compat.tests import unittest
-from units.mock.procenv import swap_stdin_and_argv
+from units.mock.procenv import ModuleTestCase
+from units.mock.generator import add_method
 
-class TestAnsibleModuleExitJson(unittest.TestCase):
 
-    def test_module_utils_basic_safe_eval(self):
+# Strings that should be converted into a typed value
+VALID_STRINGS = (
+        [("'a'", 'a')],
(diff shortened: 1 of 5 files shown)


test/units/module_utils/basic/test_safe_eval.py
```diff
@@ -24,47 +24,72 @@
...
+    self.assertEqual(res[0], "a=1")
+    self.assertEqual(type(res[1]), SyntaxError)
+
+@add_method(_check_simple_types, *VALID_STRINGS)
+@add_method(_check_simple_types, *NONSTRINGS)
+@add_method(_check_simple_types_with_exceptions, *VALID_STRINGS)
+@add_method(_check_simple_types_with_exceptions, *NONSTRINGS)
+@add_method(_check_invalid_strings, *[[i[0][0:-1]] for i in INVALID_STRINGS])
+@add_method(_check_invalid_strings_with_exceptions, *INVALID_STRINGS)
+class TestSafeEval(ModuleTestCase):
```
alikins, line 78: For the cases where I've wanted to use a nose test generator, I've ended up just making the test class inherit from object. You lose the TestCase methods (various self.assert*) but usually test generators methods have pretty simple assertions so not too much of a issue. To me it is simpler than injecting the support into a TestCase subclass.

(an example is https://github.com/alikins/ansible/blob/6e8a3d0c25fdeaa349351c6636b062376c30b0a6/test/units/plugins/lookup/test_ini.py)

test/units/module_utils/basic/test_set_mode_if_different.py
```diff
@@ -0,0 +1,145 @@
...
+        [dict(previous_changes=True)],
+        [dict(previous_changes=False)],
+        )
+@add_method(_check_mode_changed_to_0660,
+        *SYNONYMS_0660
+        )
+@add_method(_check_mode_unchanged_when_already_0660,
+        *SYNONYMS_0660
+        )
+class TestSetModeIfDifferent(TestSetModeIfDifferentBase):
```
alikins, line 97: Bravo on tackling one of the gnarlier test setups ;-> :+1:

PR #16705, opened 2016-07-14
ensure localhost is available when using dynamic ec2 inventory

philsch:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0.0
```
##### SUMMARY

My cookbook consists of roles that should run locally (AWS API calls) and roles that run on created EC2 instances. Our use case doesn't allow a strict separation of those roles in two cookbooks. If I'm executing 

```
pre_tasks:
  - meta: refresh_inventory
```

in my cookbook, all following commands do not recognize `hosts: localhost` anymore (output `no hosts matched`).

To fix this problem, localhost should always be present in the dynamic created inventory.

### DIFF
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -475,6 +475,10 @@ class Ec2Inventory(object):
                 self.get_elasticache_clusters_by_region(region)
                 self.get_elasticache_replication_groups_by_region(region)
 
+        # ensure localhost always reachable
+        self.push(self.inventory, 'localhost', 'localhost')
+        self.push(self.inventory, 'localhost', '127.0.0.1')
+
         self.write_to_cache(self.inventory, self.cache_path_cache)
         self.write_to_cache(self.index, self.cache_path_index)
 


bcoca:
A localhost is ALWAYS available as it is built into Ansible, it will be used if not offered by inventory. 
The other thing the 'built in' version does is set connection to 'local'. 

With this change the default would be trying to ssh to localhost when action/connection/delegate are set to localhost.

philsch:
Thanks for your response. Can this maybe then a bug in `meta: refresh_inventory`? The cookbook looks like this:

```
- name: Create AWS resources
  hosts: localhost
  connection: local
  gather_facts: false
  roles:
    - ...

- name: Provision stuff
  hosts: "tag_Name_xyz:&tag_Stage_prod"
  remote_user: ec2-user
  pre_tasks:
    - meta: refresh_inventory
  roles:
    - ...

- name: Again API calls
  hosts: localhost
  connection: local
  gather_facts: false
  roles:
    - ...
```

However if I run the playbook with `-i ec2.py` the first step is executed (even localhost is not included in the inventory). Step 2 find the new created instance, so the inventory is refreshed correctly. But Step 3 prints an _no hosts matched_.. So the built in localhost entry is somehow lost here?

bcoca:
@philsch can you file a bug with the refresh_inventory issue? I'm going to close this PR as it is not a fix but a workaround

PR #16664, opened 2016-07-09
Test service enabled during check run (from bug report)

gtrafimenkov:
##### ISSUE TYPE
- Bug Demonstration Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (bug-service-enabled-during-check-run f6859a426e) last updated 2016/07/09 20:58:44 (GMT +000)
  lib/ansible/modules/core: (detached HEAD db8af4c5af) last updated 2016/07/09 15:54:05 (GMT +000)
  lib/ansible/modules/extras: (detached HEAD 618e740d52) last updated 2016/07/09 15:54:05 (GMT +000)
  config file = 
  configured module search path = Default w/o overrides
```
##### SUMMARY

This pull request demonstrates a bug in the service module.  Integration tests are expected to fail on Debian 7, Ubuntu 12.04, Ubuntu 14.04 until the bug is fixed.

PR #16739, opened 2016-07-17
Testing that service not enabled during check run

gtrafimenkov:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (detached HEAD 134d70e7b9) last updated 2016/07/17 12:09:48 (GMT +300)
  lib/ansible/modules/core: (detached HEAD 7de287237f) last updated 2016/07/17 01:09:14 (GMT +300)
  lib/ansible/modules/extras: (detached HEAD 68ca157f3b) last updated 2016/07/16 20:29:35 (GMT +300)
  config file = 
```
##### SUMMARY

This pull request adds a complex destructive test that checks that a system service is not enabled when a playbook is run in the check (dry) mode.  This tests that issue https://github.com/ansible/ansible-modules-core/issues/4146 is fixed.

This is improved version of the previous pull request https://github.com/ansible/ansible/pull/16664

### DIFF
--- a/test/integration/roles/test_service/files/ansible.systemd
+++ b/test/integration/roles/test_service/files/ansible.systemd
@@ -5,3 +5,6 @@ Description=Ansible Test Service
 ExecStart=/usr/sbin/ansible_test_service "Test\nthat newlines in scripts\nwork"
 ExecReload=/bin/true
 Type=forking
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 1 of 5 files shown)


robinro:
It might be interesting for you to use the syntax introduced by #16056. You can turn on/off checkmode on individual tasks, so you don't need to call ansible twice to get this done.

mattclay:
@gtrafimenkov I agree with @robinro that this would be much better implemented using per-task check mode. You should be able to add these tests to the existing test_service role once PR #16056 is merged.

gtrafimenkov:
@mattclay @robinro Yes, it is a good idea to use new syntax.  It will make the test much simpler easier to understand.  So, let's put this PR on hold until #16056 is merged.

mattclay:
@gtrafimenkov PR #16056 has been merged.

gtrafimenkov:
I rewrote the check using new syntax.  It looks much more clear now.

PR #16741, opened 2016-07-17
Add a test for implicit int to float conversion

### LINKED ISSUE #16529: expires expects float epoch time
##### ISSUE TYPE

- Bug Report
##### ANSIBLE VERSION

```
ansible 2.1.0.0
```
##### CONFIGURATION

inventory      = /var/local/ansible/hosts
roles_path      = /var/local/ansible/roles
log_path = /home/ansible/ansible.log
##### OS / ENVIRONMENT

CentOS Linux release 7.0.1406 (Core) 
##### SUMMARY

When adding a user, with the `expires` option, specifying epoch time as an int throws an error.

```
expires: 3268684800
```

results in:
`"argument expires is of type <type 'int'> and we were unable to convert to float"`

It works when changed to:

```
expires: 3268684800.0
```
##### STEPS TO REPRODUCE

```
  tasks:
  - name: add exampleuser
    user:
      name: exampleuser
      groups: wheel
      append: yes
      password: $1$X65joYEp$WiZRYzEF25NxYjBvmQK/C/
      expires: 3268684800
```

##### EXPECTED RESULTS

```
changed: [96.118.61.180]
```
##### ACTUAL RESULTS

```
fatal: [192.168.61.180]: FAILED! => {"changed": false, "failed": true, "msg": "argument expires is of type <type 'int'> and we were unable to convert to float"}
```

Christoph-D:
edit: See my comment below.  This pull request now adds only a test because 71ef25d already fixes the issue.
##### ISSUE TYPE
- Bugfix Pull Request
##### SUMMARY

This used to be about converting int to float:

This fixes #16529 by implicitly converting module parameters from int to float.

Converting implicitly should be OK here because the code already implicitly converts from strings.

Checking if the conversion is lossless would be possible, but I think such a check would likely surprise the user more than implicit truncation of non-significant digits.

I added a test verifying that the type checks for int/float parameters work as expected.
I also needed to reset the arguments in an unrelated test or this test would fail.

After introducing 71ef25d,

### DIFF
--- a/test/units/module_utils/test_basic.py
+++ b/test/units/module_utils/test_basic.py
@@ -343,6 +343,51 @@ class TestModuleUtilsBasic(ModuleTestCase):
                 supports_check_mode=True,
             )
 
+    def test_module_utils_basic_ansible_module_type_check(self):
+        from ansible.module_utils import basic
+
+        arg_spec = dict(
+            foo = dict(type='float'),
+            foo2 = dict(type='float'),
+            foo3 = dict(type='float'),
+            bar = dict(type='int'),
+            bar2 = dict(type='int'),
+        )
+
+        # should test ok
+        args = json.dumps(dict(ANSIBLE_MODULE_ARGS={
+            "foo": 123.0, # float
+            "foo2": 123, # int
+            "foo3": "123", # string
+            "bar": 123, # int
+            "bar2": "123", # string
+        }))
+
+        with swap_stdin_and_argv(stdin_data=args):
+            basic._ANSIBLE_ARGS = None
+            am = basic.AnsibleModule(
+                argument_spec = arg_spec,
+                no_log=True,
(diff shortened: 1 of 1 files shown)


grastogi23:
Hi
I just hit this issue. I was glad to see there is a pull request. I would not have to do a workaround in my code for this reason.

cheers!
Gaurav

Christoph-D:
Someone must have overlooked this pull request because 71ef25d5 is already in ansible/devel and fixes the issue.

The result is that this pull request is now reduced to adding a test for the int/float conversion.

abadger:
@Christoph-D Oops, sorry about that.  Thanks for updating the tests, though!  Gladly merged.

abadger:
Cherrypicked this test change to devel and stable-2.1 (since the fix also went into stable-2.1).  At least the fix will be in 2.1.2.

PR #16766, opened 2016-07-20
task_result _check_key should handle empty results

### LINKED ISSUE #4214: yum repo not showing failed on a package not found
##### ISSUE TYPE

- Bug Report
##### COMPONENT NAME

yum module
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file = 

```
##### CONFIGURATION

standard, nothing off the wall
##### OS / ENVIRONMENT

CentOS 7
##### SUMMARY

Output shows `ok` but the task actually failed, very confusing
##### STEPS TO REPRODUCE

```
- name: install some repos
  become: yes
  yum:
    name: "{{ item }}"
    state: latest
  with_items:
    - epel-release
    - docker-io

- name: Install pip, python-dev package and docker with yum
  become: yes
  yum:
    pkg: "{{ item }}"
    state: latest
  with_items:
    - python-dev
    - python-pip
    - docker

- name: Install pip modules
  become: yes
  pip:
    name: "{{ item.name }}"
    version: "{{ item.version | default(omit) }}"
  with_items:
    - name: docker-py
      version: 1.8.1
    - name: docker-compose

- name: enable docker
  become: yes
  service:
    name: docker
    state: started
    enabled: yes

```
##### EXPECTED RESULTS

I would expect the task to fail, python-dev doesn't exist in centos 7
##### ACTUAL RESULTS

```
ok: [xxx.xxx.xxx.xxx] => (item=[u'python-dev', u'python-pip', u'docker']) => {"changed": false, "failed": true, "invocation": {"module_args": {"conf_file": null, "disable_gpg_check": false, "disablerepo": null, "enablerepo": null, "exclude": null, "install_repoquery": true, "list": null, "name": ["python-dev", "python-pip", "docker"], "state": "latest", "update_cache": false, "validate_certs": true}, "module_name": "yum"}, "item": ["python-dev", "python-pip", "docker"], "msg": "No Package matching 'python-dev' found available, installed or updated", "rc": 0, "results": []}
```

willthames:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel 73d29c5dea) last updated 2016/07/20 14:19:09 (GMT +1000)
  lib/ansible/modules/core: (detached HEAD 7de287237f) last updated 2016/07/20 13:59:30 (GMT +1000)
  lib/ansible/modules/extras: (detached HEAD 68ca157f3b) last updated 2016/07/20 13:59:30 (GMT +1000)
  config file = 
```
##### SUMMARY

When a task result has an empty results list, the
list should be ignored when determining the results
of `_check_key`. Here the empty list is treated the
same as a non-existent list.

This fixes a bug that manifests itself with squashed
items - namely the task result contains the correct
value for the key, but an empty results list. The
empty results list was treated as zero failures
when deciding which handler to call - so the task
show as a success in the output, but is deemed to
have failed when deciding whether to continue.

This also demonstrates a mismatch between task
result processing and play iteration.

A test is also added for this case, but it would not
have caught the bug - because the bug is really in
the display, and not the success/failure of the
task (visually the test is more accurate).

Fixes ansible/ansible-modules-core#4214

### DIFF
--- a/lib/ansible/executor/task_result.py
+++ b/lib/ansible/executor/task_result.py
@@ -62,7 +62,7 @@ class TaskResult:
         return self._check_key('unreachable')
 
     def _check_key(self, key):
-        if 'results' in self._result and self._task.loop:
+        if self._result.get('results', []) and self._task.loop:
             flag = False
             for res in self._result.get('results', []):
                 if isinstance(res, dict):
--- a/test/integration/roles/test_yum/tasks/yum.yml
+++ b/test/integration/roles/test_yum/tasks/yum.yml
@@ -185,3 +185,15 @@
 
 - name: uninstall sos and sharutils
   yum: name=sos,sharutils state=removed
+
+- name: install non-existent rpm 
+  yum: name="{{ item }}"
+  with_items:
+  - does-not-exist
+  register: non_existent_rpm
+  ignore_errors: True
+
+- name: check non-existent rpm install failed
+  assert:
+    that:
+    - non_existent_rpm|failed


jimi-c:
Merged, thanks!

PR #16779, opened 2016-07-21
Make returning results after loading vars optional

mirskifa:
##### ISSUE TYPE

- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0
  config file = 
```
##### SUMMARY

Commit 0ba9a6a removed an expensive merge operation that is normally not required.  However, there is no public method of VariableManager to read the set of variables an inventory defines that apply just to that group or host.  This use case is required for a plugin used by our team.

As calculating the results is an expensive operation and is normally not required, I have defaulted the behaviour to not generate the result.  This makes the default behaviour the same as after 0ba9a6a, but allows for the previous behaviour to be used if required.

### DIFF
--- a/lib/ansible/inventory/__init__.py
+++ b/lib/ansible/inventory/__init__.py
@@ -718,13 +718,13 @@ class Inventory(object):
             self._vars_per_host = {}
             self._vars_per_group = {}
 
-    def get_host_vars(self, host, new_pb_basedir=False):
+    def get_host_vars(self, host, new_pb_basedir=False, return_results=False):
         """ Read host_vars/ files """
-        return self._get_hostgroup_vars(host=host, group=None, new_pb_basedir=new_pb_basedir)
+        return self._get_hostgroup_vars(host=host, group=None, new_pb_basedir=new_pb_basedir, return_results=return_results)
 
-    def get_group_vars(self, group, new_pb_basedir=False):
+    def get_group_vars(self, group, new_pb_basedir=False, return_results=False):
         """ Read group_vars/ files """
-        return self._get_hostgroup_vars(host=None, group=group, new_pb_basedir=new_pb_basedir)
+        return self._get_hostgroup_vars(host=None, group=group, new_pb_basedir=new_pb_basedir, return_results=return_results)
 
     def _find_group_vars_files(self, basedir):
         """ Find group_vars/ files """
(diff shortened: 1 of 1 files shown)


willthames:
@jctanner, @jimi-c I think this is a great fix, I think both 0ba9a6a and #16779 should be removed from 2.1 and moved to 2.2 - this shouldn't go into 2.1.1 as to me it's a breaking API change (I don't have huge standards around breaking undocumented APIs but don't really expect a point release to do so).

jhawkesworth:
For what its worth, I have a plugin that needs access to host vars, so I've been hit by this with 2.1.1 testing.

Following on from this suggestion on the devel group: https://groups.google.com/forum/#!topic/ansible-devel/3WKVYhAPKJU 
I have a less than pretty way of getting vars for a host like this 

```
def v2_playbook_on_task_start(self, *args, **kwargs):
   # get the loader
   _loader = args[0]._loader
   # get all vars using the loader
   all_vars = args[0].get_vars(_loader, task=args[0])
   # get hosts for the current inventory
   hosts = args[0].get_variable_manager()._inventory.get_hosts()
   # pick the hostvars out of all_vars
   inventory_host_vars = all_vars['hostvars']
   # loop through inventory hosts
   for host in hosts:
      # pick the group vars for the host
      host_group_vars = inventory_host_vars[str(host)]
      # do plugin-specific stuff if a host_group_var exists
      if ('some_interesting_thing' in host_group_vars):
         # do useful plugin stuff...
```

I don't really mind if its got to be this way for my plugin but I thought I'd mention it in case it can unblock anyone.

willthames:
@jhawkesworth I'll only be blocked if the change makes it live without this fix.

My need is to work out where in inventory a variable comes from - so I need to be able to isolate host specific vars from group specific vars and work out at which level the variable actually applies (so host vars beat child group vars beat their parent group vars)

mirskifa:
Similarly to @willthames, I'd be happiest if any releases have either neither 0ba9a6a and #16779 or both of them.  No strong opinions over whether that's by putting #16779 into 2.1 or by pulling 0ba9a6a out and into 2.2, but having 0ba9a6a without #16779 means having to run from a feature branch which I'd obviously like to avoid.

jimi-c:
Merged and cherry-picked to stable-2.1, and cut 2.1.1 rc5 for this.

willthames:
Thanks @jimi-c

mirskifa:
@jimi-c: Thanks for merging that into 2.1

PR #16787, opened 2016-07-21
Ignore ControlPersist broken pipe errors. Fixes #16731

### LINKED ISSUE #16731: Task fails when ControlPersist is about to expire
##### ISSUE TYPE
Bug Report

##### COMPONENT NAME
ssh controlpersist

##### ANSIBLE VERSION
```
ansible 2.1.1.0
Tried it on stable-2.1 and some other branches.
```

##### CONFIGURATION
forks = 200
ssh_args = -o ControlMaster=auto -o ControlPersist=60s
ssh_args above is the default but I put it here since its important for this bug.

##### OS / ENVIRONMENT
CentOS 6/Ubuntu 16.04

##### SUMMARY
If you run a play with a lot of hosts (our play has 130 hosts) and if there is a task which will be executed just when the `ControlPersist` time out is about to expire, random hosts can fail with the following SSH error:
```
Failed to connect to the host via ssh. (255, '', 'OpenSSH_5.3p1, OpenSSL 1.0.1e-fips 11 Feb 2013
debug1: Reading configuration data /home/user/.ssh/config
debug1: Reading configuration data /etc/ssh/ssh_config
debug1: Applying options for *
debug1: auto-mux: Trying existing master
debug2: fd 3 setting O_NONBLOCK
mux_client_hello_exchange: write packet: Broken pipe)
```
My guess is that Ansible is just about to reuse the SSH socket to connect to the remote host, but at that moment ControlMaster closes the socket to the server. The playbook which can reproduce this will make it more clear what I mean here.
Also a big thanks to @jctanner for providing a patch so that we could see this SSH error, since Ansible hides them. I reported a bug about it in https://github.com/ansible/ansible/issues/16732.

##### STEPS TO REPRODUCE
To reproduce this bug you will need as many hosts in the playbook as you can come up with and they all have to have different IP addresses, so that SSH's ControlMaster (CM) creates a new socket for each of hosts. We have 130 hosts in our play and this play manages to fail every time, the more hosts you have in a play the easier is to catch this bug. It should fail with less, but it will be harder capture it.
```
---
- hosts: large_group_of_hosts
  gather_facts: no
  tasks:
    - ping:
    - ping:
    - pause: seconds=55
    - ping:
      with_items: [ 1, 2, 3]
```
The idea behind this play is that we run `ping` module two times in a row so that the first time `ping` connects to all the host, while the second time it will run faster since CM will hold an SSH connection already to each host, which means all hosts will execute `ping` very fast and in a similar time frame. Then we pause for 55 seconds, so that when we can run ping after that because within those 5 seconds `ControlPersist=60s` time out will be expiring on all hosts while ping is being executed and random hosts Ansible will report as Unreachable because of the SSH error above.
I used `with_items` so that ping is executed on each hosts 3 times, because this made it even easier for the error to occur. I guess lowering the number of forks to less than the number of hosts (e.g. 50 in this situation) would also help, since for me it failed with `forks = 50` as well.
You might need to tweak number of seconds of pause depending on how fast ping executes on the hosts, but you should be able to reproduce this issue if you have enough hosts in the play.
If we disable `ControlMaster` or if we set `ControlPersist` to a higher number (e.g. `120s`) all hosts finish without a problem.

##### EXPECTED RESULTS
Playbook should finish execution.

##### ACTUAL RESULTS
Random hosts return as unreachable.

kustodian:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0
```
##### SUMMARY

Automatically retires an ssh connection if it broke because of `mux_client_hello_exchange: write packet: Broken pipe` which happens when Ansible attempts to reuse an SSH ControlMaster socket just when the ControlPersist time out expires. More about the issue this pull request fixes in #16731.

### DIFF
--- a/lib/ansible/plugins/connection/ssh.py
+++ b/lib/ansible/plugins/connection/ssh.py
@@ -131,6 +131,10 @@ except ImportError:
 SSHPASS_AVAILABLE = None
 
 
+class AnsibleControlPersistBrokenPipeError(AnsibleError):
+    ''' ControlPersist broken pipe '''
+    pass
+
 def _ssh_retry(func):
     """
     Decorator to retry ssh/scp/sftp in the case of a connection failure
@@ -148,11 +152,17 @@ def _ssh_retry(func):
         cmd_summary = "%s..." % args[0]
         for attempt in range(remaining_tries):
             try:
-                return_tuple = func(self, *args, **kwargs)
-                display.vvv(return_tuple, host=self.host)
-                # 0 = success
-                # 1-254 = remote command return code
-                # 255 = failure from the ssh command itself
+                try:
+                    return_tuple = func(self, *args, **kwargs)
+                    display.vvv(return_tuple, host=self.host)
+                    # 0 = success
+                    # 1-254 = remote command return code
+                    # 255 = failure from the ssh command itself
+                except (AnsibleControlPersistBrokenPipeError) as e:
+                    # Retry one more time because of the ControlPersist broken pipe (see #16731)
+                    display.vvv(u"RETRYING BECAUSE OF CONTROLPERSIST BROKEN PIPE")
+                    return_tuple = func(self, *args, **kwargs)
+
                 if return_tuple[0] != 255:
                     break
                 else:
@@ -743,6 +753,12 @@ class Connection(ConnectionBase):
         if p.returncode != 0 and controlpersisterror:
             raise AnsibleError('using -c ssh on certain older ssh versions may not support ControlPersist, set ANSIBLE_SSH_ARGS="" '
                                '(or ssh_args in [ssh_connection] section of the config file) before running again')
+
+        # If we find a broken pipe because of ControlPersist timeout expiring (see #16731),
+        # we raise a special exception so that we can retry a connection.
+        controlpersist_broken_pipe = b'mux_client_hello_exchange: write packet: Broken pipe' in b_stderr
+        if p.returncode == 255 and controlpersist_broken_pipe:
+            raise AnsibleControlPersistBrokenPipeError('SSH Error: data could not be sent because of ControlPersist broken pipe.')
 
         if p.returncode == 255 and in_data and checkrc:
             raise AnsibleConnectionFailure('SSH Error: data could not be sent to remote host "%s". Make sure this host can be reached over ssh' % self.host)
--- a/test/units/plugins/connection/test_ssh.py
+++ b/test/units/plugins/connection/test_ssh.py
@@ -590,7 +590,7 @@ class TestSSHConnectionRetries(object):
 
         self.mock_popen_res.stdout.read.side_effect = [b"", b"my_stdout\n", b"second_line"]
         self.mock_popen_res.stderr.read.side_effect = [b"", b"my_stderr"]
-        type(self.mock_popen_res).returncode = PropertyMock(side_effect=[255] * 3 + [0] * 4)
+        type(self.mock_popen_res).returncode = PropertyMock(side_effect=[255] * 4 + [0] * 4)
 
         self.mock_selector.select.side_effect = [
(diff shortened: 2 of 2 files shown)


jctanner:
@kustodian hey, thanks for writing this up. I would like for it to also provide the full errors for other problems too.

kustodian:
@jctanner this won't display actual SSH errors which you want to add into
Ansible, I left that part to you, since you said you will fix that :-)

This only fixes the bug with Ansible disconnecting because of
ControlPersist timeout expiring when Ansible is just about to connect,
which causes random host disconnects and is explained in detail in #16731.

abadger:
@bcoca has given this a +1 on slack.

With kustodian's explanation: https://github.com/ansible/ansible/issues/16731#issuecomment-291443094 I think this looks okay as well.

Waiting for @nitzmahone to have a chance to look at this and then can decide whether it's ready for merge.

abadger:
nitzamahone also gives +1 with the explanation here: https://github.com/ansible/ansible/issues/16731#issuecomment-291443094

lib/ansible/plugins/connection/ssh.py
```diff
@@ -550,8 +550,14 @@ def _run(self, cmd, in_data, sudoable=True):
         if p.returncode != 0 and controlpersisterror:
             raise AnsibleError('using -c ssh on certain older ssh versions may not support ControlPersist, set ANSIBLE_SSH_ARGS="" (or ssh_args in [ssh_connection] section of the config file) before running again')
 
+        # If we find a broken pipe because of ControlPersist timeout expiring (see #16731),
+        # we raise an internal error so that we can retry a connection.
+        controlpersist_broken_pipe = 'mux_client_hello_exchange: write packet: Broken pipe' in stderr
+        if p.returncode == 255 and controlpersist_broken_pipe:
+            raise AnsibleInternalError('SSH Error: data could not be sent because of ControlPersist broken pipe.')
```
nitzmahone: Since this is a generic exception, I think we need a specific error code or something to ensure that if someone else tosses this down the line for another reason, that the associated retry doesn't get triggered.

lib/ansible/plugins/connection/ssh.py
```diff
@@ -596,10 +602,15 @@ def exec_command(self, *args, **kwargs):
...
-                # 1-254 = remote command return code
-                # 255 = failure from the ssh command itself
+                try:
+                    return_tuple = self._exec_command(*args, **kwargs)
+                    # 0 = success
+                    # 1-254 = remote command return code
+                    # 255 = failure from the ssh command itself
+                except (AnsibleInternalError) as e:
+                    # Retry one more time because of the ControlPersist broken pipe
+                    display.vvv(u"RETRYING BECAUSE OF CONTROLPERSIST BROKEN PIPE")
```
nitzmahone: Ensure that it's the exact right exception (by some mechanism, eg, error code as mentioned above) before we trigger the retry.

nitzmahone (changes requested):

nitzmahone:
I'm OK with the overall approach, just want to make sure we don't trigger this inadvertently in the future if someone else throws `AnsibleInternalError` for some other reason. Also perhaps some more caution tape around the actual retry, saying that it's only safe to use in this exact scenario.

kustodian:
Because in the mean time `ssh.py` was changed a lot, it was very hard for me to do a proper merge. It was a lot easier to apply my changes to the latest version of the file.

@nitzmahone I created a separate class for this exception, so this type of exception won't get raised by a mistake by something else, so I guess this should solve your concern.

kustodian:
@nitzmahone I would really appreciate if you could help me with the two unit tests which are failing, I don't even know where to start.

abadger:
Merged to devel.

PR #16795, opened 2016-07-22
Only show the traceback for importing cryptography when in Ansible Debug

abadger:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
devel
```
##### SUMMARY

Most users won't care whether we use cryptography or pycrypto in vault.  Some may feel that cryptography is more secure or faster than pycrypto, though.  Having the traceback be available but only in debug mode is a good compromise.


abadger:
@cdosborn just letting you know I'm modifying the warning handling in your patch slightly.  If it passes tests I'll merge this to devel and stable-2.1

PR #16824, opened 2016-07-25
Fixing remote docker issues

gianperrone:
##### ISSUE TYPE

- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel b435ab17ed) last updated 2016/07/26 00:59:22 (GMT +200)
  lib/ansible/modules/core: (detached HEAD d64060ace0) last updated 2016/07/25 22:27:29 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD bef9a1c14f) last updated 2016/07/25 22:27:32 (GMT +200)
  config file = /cygdrive/d/ownCloud/ownCloud/Ansible/ansible.cfg
```
##### SUMMARY

This fixes two things in the docker connection plugin.
1. docker_extra_args were previously ignored by some commands including the version check for newer versions which made connecting to newer remote hosts impossible.
2. When connecting to a remote host with a newer client version, docker return an error. This is easily fixed by setting the DOCKER_API_VERSION environment variable to the server version.

It is debatable whether b) is a bugfix, a feature or a workaround (why doesn't docker do this itself?) but it is definitely enabling the use case of a current docker client with a "stable" (as in Debian) docker host.

I'd like to further enhance the remote feature enabling simpler syntax than docker_extra_args for it, but that will be a different pull request.

### DIFF
--- a/lib/ansible/plugins/connection/docker.py
+++ b/lib/ansible/plugins/connection/docker.py
@@ -4,6 +4,7 @@
 # (c) 2014, Lorin Hochstein
 # (c) 2015, Leendert Brouwer
 # (c) 2015, Toshio Kuratomi <tkuratomi@ansible.com>
+# (c) 2016, Gian Perrone
 #
 # Maintainer: Leendert Brouwer (https://github.com/objectified)
 #
--- a/lib/ansible/plugins/connection/docker.py
+++ b/lib/ansible/plugins/connection/docker.py
@@ -114,15 +114,16 @@ class Connection(ConnectionBase):
         cmd = list(self.docker_cmd)
         cmd += ['version']
 
-        try:
-            cmd_output = subprocess.check_output(cmd, stderr=subprocess.STDOUT)
-        except subprocess.CalledProcessError as err:
+        p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        cmd_output, cmd_error = p.communicate()
+
+        if p.returncode != 0:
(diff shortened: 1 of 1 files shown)


lib/ansible/plugins/connection/docker.py
```diff
@@ -106,33 +111,35 @@ def _sanitize_version(version):
...
-
-        if self._play_context.docker_extra_args:
-            cmd += self._play_context.docker_extra_args.split(' ')
-
+        cmd = list(self.docker_cmd)
         cmd += ['version']
 
-        cmd_output = subprocess.check_output(cmd)
+        try:
+            cmd_output = subprocess.check_output(cmd, stderr=subprocess.STDOUT)
```
alikins: subprocess.check_output is python 2.7+ and controller side code like connection plugins target python2.6

This module was using subprocess.check_output before this pr, but that needs to be fixed as well
(see https://github.com/ansible/ansible/pull/16841). This pr doesn't need to fix the existing cases, but should avoid adding new use of subprocess.check_output()

gianperrone:
Both instances of subprocess.check_output have been substituted with subprocess.Popen. Additionally fixes #16841 collaterally.

calfonso:
@gianperrone this requires a rebase, please do so and we'll finish the review.

gianperrone:
There is a merge conflict and neither do I have a suitable test environment anymore nor do I have any interest in redoing the code after more than one year.

lathama:
@gianperrone if this is not of interest any longer can you close the PR.

PR #16838, opened 2016-07-27
Indent JSON output on verbose

jpic:
As discussed on the [Improving default output thread](https://groups.google.com/forum/#!topic/ansible-devel/pyJGj1dXZ-c), this PR enables output indentation with only -v.

### DIFF
--- a/lib/ansible/plugins/callback/__init__.py
+++ b/lib/ansible/plugins/callback/__init__.py
@@ -89,7 +89,7 @@ class CallbackBase:
         if result.get('_ansible_no_log', False):
             return json.dumps(dict(censored="the output has been hidden due to the fact that 'no_log: true' was specified for this result"))
 
-        if not indent and (result.get('_ansible_verbose_always') or self._display.verbosity > 2):
+        if not indent and (result.get('_ansible_verbose_always') or self._display.verbosity > 0):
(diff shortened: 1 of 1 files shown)


bcoca:
I would not make this change for all users, many prefer the condensed output and will use higher verbosity if they need to really check the large json outputs.

It would be better to make it configurable and keep the current default.

jpic:
Perhaps we could make a new option, such as '--readable' ?

Should I group both PRs ?

bcoca:
separate is fine, i would not want another command line option, using config should be enough, either through ansible.cfg or environment variable.

jpic:
The environment variable is a very good idea. ANSIBLE_READABLE ?

bcoca:
bikeshedding now: ANSIBLE_INDENT_VERBOSITY

jpic:
Well, dunno if it's bikeshedding since we're going to have to live with this for the rest of our lifes or something :p What I don't like with INDENT_VERBOSITY is that I don't understand what it means as a non native speaker, perhaps it's possible to find a more representative name in one word if possible ?

I'm updating the branch for config/env var support

jpic:
Closing in favor of #16839 which adds a debug callback that does the same.

PR #16843, opened 2016-07-27
Strip group names of whitespaces, so we can use space after comma in openstack's groups metadata field.

cantorek:
##### ISSUE TYPE

- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.0.2.0
  config file =
```
##### SUMMARY

strip() is non destructive on normal strings, and in this case it will allow to separate group names in openstack metadata by comma and space ", ". This is more elegant and easier to read. Without this, openstack.py will create a group names with leading space.


nitzmahone:
lgtm, thanks!

PR #16854, opened 2016-07-28
Add default path and override for azure_rm.ini

dol3y:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.0.0.2
  config file = 
```
##### SUMMARY

The `azure_rm.ini` is not loaded when placed alongside the `azure_rm.py` script and the script is called from a higher directory e.g. `inventory/azure_rm.py --list`.

This change expects the .ini file to be placed alongside the script but can be overridden by the environment variable `AZURE_INI_PATH`.


nitzmahone:
looks great- thanks for the fix/enhancement!

PR #16905, opened 2016-08-01
introduce fine-grained control of yaml parser warnings

### LINKED ISSUE #13603: Fine grained warning disabling
Issue Type:
Bug Report / Feature Idea

Ansible Version:
latest and v2.0.0-0.7.rc2

Ansible Configuration:
Nothing relevant

Environment:
OSX

Summary:
The warning about duplicate dict keys when parsing yaml is annoying for legitimate use cases and there should be a feature to disable it.

Steps To Reproduce:
I use YAML anchors/aliases extensively to make my giant var files much easier to read.  The following gist demonstrates what I'm doing and what I'm seeing:
https://gist.github.com/mspiegle/ef4630cbdf8b18f93613

The Ansible code responsible for this is:
https://github.com/ansible/ansible/blob/v2.0.0-0.7.rc2/lib/ansible/parsing/yaml/constructor.py#L67

Expected Results:
I either expect Ansible to not bother me about this, or provide some way for me to disable this particular warning.

Actual Results:
I get lots of warnings about duplicate dict keys.  The duplicate dict keys are completely normal and expected in my use case.

### LINKED ISSUE #16903: Ansible should fail in the presence of duplicate keys
##### ISSUE TYPE
Bug Report

##### COMPONENT NAME
core

##### ANSIBLE VERSION
```
ansible 2.1.1.0
config file =
```

##### CONFIGURATION
N/A

##### OS / ENVIRONMENT
N/A

##### SUMMARY
We recently had a change submitted to the triple-quickstart project
that contained the following code:
```
- name: Rebuild inventory
  hosts: localhost
  roles:
    - rebuild-inventory
  tags:
    - provision
# check the target virthost distro for RHEL or CentOS
  hosts: virthost
  roles:
    - provision/support_check
  tags:
    - provision
```
Note that the second task list is missing the YAML list marker; it
should look like this:
```
- hosts: virthost
  roles:
    - provision/support_check
  tags:
    - provision
```
This is a clear error, but it passes through the YAML parser with only
the following warning:
```
[WARNING]: While constructing a mapping from /home/lars/src/tripleo-
quickstart/playbooks/provision.yml, line 21, column 3, found a
duplicate dict key (roles).  Using last defined value only.
```

##### STEPS TO REPRODUCE
(see above playbook samples)

##### EXPECTED RESULTS
Since duplicate keys are in fact not possible, I would expect Ansible
to abort with an error at this point.

##### ACTUAL RESULTS
Ansible ignores the error and continues on its merry way.
Because this warning comes from deep within the parser, it's not even
possible for something like `ansible-lint` to catch the problem,
either.
##### ADDITIONAL COMMENTARY
This is, oddly enough, the exact opposite of #13603.

larsks:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (bug/16903 1155eb740c) last updated 2016/08/01 14:25:00 (GMT -400)
  lib/ansible/modules/core: (detached HEAD 24db4de245) last updated 2016/08/01 12:00:49 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD db979dde74) last updated 2016/08/01 12:00:50 (GMT -400)
  config file = 
```
##### SUMMARY

This patch addresses both #16903 and #13603 by introducing a mechanism
for fine-grained control over warnings in
`ansible.parsing.yaml.constructor.AnsibleConstructor`.  We introduce
two new parameters:
- `fatal_warnings`: named warnings in this list will raise an
  exception instead of emitting a warning message.
- `ignore_warnings`: named warnings in this list will be ignored.

The constructor code currently recognizes one named warning,
`duplicate-key`, which is generated when the parser encounters a
duplicate key in a YAML mapping.

This patches _does not_ change the default behavior, because while the
following is clearly an error...

```
- host: myhost
  tasks:
    - debug:
        msg: this is a task

  host: yourhost
  tasks:
    - debug:
        msg: oops
```

...there are other situations in which people intentionally use
duplicate keys to override values set via YAML anchors, as in #13603.

This patch explicitly does not include a user interface for setting
these options; it is expected that initially they will be used by
third-party tools that call the ansible modules directly.

Fixes: #16903
Fixes: #13603

### DIFF
--- a/lib/ansible/parsing/dataloader.py
+++ b/lib/ansible/parsing/dataloader.py
@@ -63,10 +63,12 @@ class DataLoader():
         ds = dl.load_from_file('/path/to/file')
     '''
 
-    def __init__(self):
+    def __init__(self, fatal_warnings=None, ignore_warnings=None):
         self._basedir = '.'
         self._FILE_CACHE = dict()
         self._tempfiles = set()
+        self._fatal_warnings=fatal_warnings
+        self._ignore_warnings=ignore_warnings
 
         # initialize the vault stuff with an empty password
         self.set_vault_password(None)
@@ -149,7 +151,9 @@ class DataLoader():
     def _safe_load(self, stream, file_name=None):
         ''' Implements yaml.safe_load(), except using our custom loader class. '''
 
-        loader = AnsibleLoader(stream, file_name)
+        loader = AnsibleLoader(stream, file_name,
+                               fatal_warnings=self._fatal_warnings,
+                               ignore_warnings=self._ignore_warnings)
         try:
             return loader.get_single_data()
         finally:
--- a/lib/ansible/parsing/yaml/constructor.py
+++ b/lib/ansible/parsing/yaml/constructor.py
@@ -24,6 +24,8 @@ from yaml.nodes import MappingNode
 from ansible.parsing.yaml.objects import AnsibleMapping, AnsibleSequence, AnsibleUnicode
 from ansible.vars.unsafe_proxy import wrap_var
 
+named_warnings = set(['duplicate-key'])
+
 try:
     from __main__ import display
 except ImportError:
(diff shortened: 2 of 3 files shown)


bcoca:
After thinking about this (run for cover!), I'm not against making this configurable but I do wan't to change the method for squelching warnings or toggling them as errors or not. 

This is something that we have encountered in other places and doing in a piecemeal way, its clear that we need a generic 'error toggle' and 'warning filter'. cc @alikins @abadger @jimi-c @jctanner @mattclay @ryansb

larsks:
Closing an old PR.  Would love to see something like this in some other form.

PR #16920, opened 2016-08-02
How do nested loops work? Update documentation...

alexlenail:
##### SUMMARY

This PR may not be correct but that's exactly the point -- I'm confused about how this method works and I'd like for the docs to be clarified a little... If this PR's update is wrong, then I think it proves my point that this PR is needed...


alexlenail:
@jctanner looks like all the tests pass -- could you give this a look over?

alexlenail:
@jctanner any news?

alexlenail:
Okay guess this isn't important.

PR #16924, opened 2016-08-02
Use file list, not recursion, in _fixup_perms.

mattclay:
##### ISSUE TYPE

Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (setfacl-pipe f20bea6c5f) last updated 2016/08/02 15:15:18 (GMT -700)
  lib/ansible/modules/core: (detached HEAD 24db4de245) last updated 2016/07/28 11:12:27 (GMT -700)
  lib/ansible/modules/extras: (detached HEAD db979dde74) last updated 2016/07/28 11:12:27 (GMT -700)
  config file = 
```
##### SUMMARY

Calling setfacl on FreeBSD using find -exec always returns 0, even on failure.

Instead of recursively updating permissions and file ownership in _fixup_perms, we now update them explicitly for each uploaded file.

This means _fixup_perms now passes a list of paths to the following commands instead of a single path on which to operate recursively:
- setfacl
- chown
- chmod

Tests for FreeBSD which pass as a result of this fix have been enabled as well.

### DIFF
--- a/lib/ansible/plugins/action/__init__.py
+++ b/lib/ansible/plugins/action/__init__.py
@@ -293,7 +293,7 @@ class ActionBase(with_metaclass(ABCMeta, object)):
 
         return remote_path
 
-    def _fixup_perms(self, remote_path, remote_user, execute=True, recursive=True):
+    def _fixup_perms(self, remote_paths, remote_user, execute=True):
         """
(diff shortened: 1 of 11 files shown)


abadger:
+1 to merge!

PR #16954, opened 2016-08-04
Adding docker_service plugin

chouseknecht:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel 8e956675f7) last updated 2016/08/04 12:46:14 (GMT -400)
  lib/ansible/modules/core: (detached HEAD 24db4de245) last updated 2016/08/02 04:22:23 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD db979dde74) last updated 2016/08/02 04:22:23 (GMT -400)
```
##### SUMMARY

Adding docker_service action plugin to copy project_src to dest on target node. Enables the following playbook task:

```
- hosts: centosv
  tasks:
    - docker_service:
        project_name: django_flask
        project_src: flask
        dest: ~/flask
```

Uses the copy module to recursively copy the contents of _project_src_ to the _dest_ on the target node. It then executes the docker_service module with the value  of _project_src_ set to the value of _dest_.

See [PR 4332](https://github.com/ansible/ansible-modules-core/pull/4332) for docker_service module changes.

### DIFF
--- a/lib/ansible/plugins/action/docker_service.py
+++ b/lib/ansible/plugins/action/docker_service.py
@@ -1,0 +1,182 @@
+# Copyright 2016 Red Hat | Ansible
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+# Make coding more python3-ish
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import os
+
+from ansible.errors import AnsibleError
+from ansible.plugins.action import ActionBase
+from ansible.utils.hashing import checksum
+from ansible.utils.unicode import to_bytes, to_str
+
+
+class ActionModule(ActionBase):
+
+    def run(self, tmp=None, task_vars=None):
+        ''' handler for file transfer operations '''
+        if task_vars is None:
+            task_vars = dict()
+
+        result = super(ActionModule, self).run(tmp, task_vars)
+
+        source  = self._task.args.get('project_src', None)
+        dest    = self._task.args.get('dest', None)
+        raw     = False
+        force   = True
+        follow  = True
+
+        if source and dest:
+
+            # Check if the source ends with a "/"
+            source_trailing_slash = False
+            if source:
+                source_trailing_slash = self._connection._shell.path_has_trailing_slash(source)
+
+            try:
+                source = self._find_needle('files', source)
+            except AnsibleError as e:
+                result['failed'] = True
+                result['msg'] = to_str(e)
+                return result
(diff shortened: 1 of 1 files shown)


lib/ansible/plugins/action/docker_service.py
```diff
@@ -0,0 +1,182 @@
...
+                if not self._connection._shell.path_has_trailing_slash(dest):
+                    dest = self._connection._shell.join_path(dest, '')
+            else:
+                source_files.append((source, os.path.basename(source)))
+
+            changed = False
+            module_return = dict(changed=False)
+
+            # If this is a recursive action create a tmp path that we can share as the _exec_module create is too late.
+            remote_user = task_vars.get('ansible_ssh_user') or self._play_context.remote_user
```
bcoca: you should NOT use ansible_ssh_user directly, it might be the wrong var, play_context.remote_user should always be the correct one.

lib/ansible/plugins/action/docker_service.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def run(self, tmp=None, task_vars=None):
+        ''' handler for file transfer operations '''
+        if task_vars is None:
+            task_vars = dict()
+
+        result = super(ActionModule, self).run(tmp, task_vars)
+
+        source  = self._task.args.get('project_src', None)
+        dest    = self._task.args.get('remote_src', None)
```
bcoca: remote_src is a boolean, but you are using it as a path here, I think you have confused how this feature works elsewhere.

it is just a flag indicating that project_src is on the target (True) or on the controller (False)

look at copy or assemble to get this working right.

chouseknecht: Yeah, I get what you mean. So if the thing to do is keep the parameters meanings similar to the copy module, then the remote path parameter should actually be _dest_.  I've gone ahead and made that change. 

I don't think docker_service needs to support _remote_src_. So I'm removing any reference to remote_src for now. If it turns out to be a feature we need, we can bring it back and use it exactly as the copy module does. 

Thanks for point this out.

lib/ansible/plugins/action/docker_service.py
chouseknecht: Roger that. FWIW the copy module still uses ansible_ssh_user.

lib/ansible/plugins/action/docker_service.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def run(self, tmp=None, task_vars=None):
+        ''' handler for file transfer operations '''
+        if task_vars is None:
+            task_vars = dict()
+
+        result = super(ActionModule, self).run(tmp, task_vars)
+
+        source  = self._task.args.get('project_src', None)
+        dest    = self._task.args.get('dest', None)
```
bcoca, line 40: at this point you are just making a copy of the copy action plugin, this is something we want to avoid, a role that uses both copy and docker_service is more appropriate in this case.

chouseknecht, line 40: Except users don't want a role. They want this. And, I'm not clear on how changing the name of the parameter from 'remote_src' to 'dest' caused you to arrive at this conclusion.

kiview:
This feature would be really appreciated, do you see any way to implement this in a different fashion? Since the `docker_service_plugin` allows the usage of docker-compose syntax inside the playbook, would it be possible to use an external file instead of specifying the configuration in place? Having a dedicated `docker-compose.yml` file allows for a much better portability.

kiview:
I've just published an Ansible role which kind of fulfills the need for this feature:
https://galaxy.ansible.com/kiview/docker-service-deploy/

PR #16959, opened 2016-08-04
Allow multiple newline delimited passwords in the vault password

jctanner:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
2.2
```
##### SUMMARY

If the vault password contains newlines, split it and iterate through the list to attempt decryption.

### DIFF
--- a/lib/ansible/parsing/vault/__init__.py
+++ b/lib/ansible/parsing/vault/__init__.py
@@ -113,7 +113,6 @@ class VaultLib:
             for _password in password.split('\n'):
                 b_password = to_bytes(_password, errors='strict', encoding='utf-8')
                 passwords.append(b_password)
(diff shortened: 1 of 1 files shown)


lib/ansible/parsing/vault/__init__.py
```diff
@@ -103,10 +103,19 @@ def check_prereqs():
...
         self.cipher_name = None
         self.b_version = b'1.1'
 
+    def split_password(self, password):
+        passwords = []
+        if password:
+            for _password in password.split('\n'):
+                b_password = to_bytes(_password, errors='strict', encoding='utf-8')
+                passwords.append(b_password)
+        passwords = [x.strip() for x in passwords if x.strip()]
```
alikins: This seems like it would break passwords with leading/trailing whitespace.

PR #16961, opened 2016-08-04
YAML treats some unquoted strings as booleans. For instance,

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
devel and 2.1
```
##### SUMMARY

```
  uri:
    follow_redirects: no
```

Will lead yaml to set follow_redirects=False.  This is problematic when
the module parameter is not a boolean value but a string.  For instance:

```
  follow_redirects = dict(required=False, default='safe', choices=['all', 'safe', 'none', 'yes', 'no']),
```

Our parameter validation code ends up getting follow_redirects="False"
instead of "no".  The 100% fix is for the user to quote their strings in
playbooks like:

```
  uri:
    follow_redirects: "no"
```

But we can fix quite a few common cases by trying to switch "False" back
into the string that it was specified as.  We only do this if there is
only one correct choices value that could have been specified.  In the
follow_redirects example, a value of "True" only maps back to "yes" and
a value of "False" only maps back to "no" so we can do this.  If choices
also contained "on" and "off" then we couldn't map back safely and would
need to force the module author to change the module to handle this
case.

Fixes parts of the following PRs:
- https://github.com/ansible/ansible-modules-core/pull/4220
- https://github.com/ansible/ansible-modules-extras/pull/2593

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -585,6 +585,19 @@ def env_fallback(*args, **kwargs):
     else:
         raise AnsibleFallbackNotFound
 
+def _lenient_lowercase(lst):
+    """Lowercase elements of a list.
+
+    If an element is not a string, pass it through untouched.
+    """
+    lowered = []
+    for value in lst:
+        try:
+            lowered.append(value.lower())
+        except AttributeError:
+            lowered.append(value)
+    return lowered
+
 
 class AnsibleFallbackNotFound(Exception):
     pass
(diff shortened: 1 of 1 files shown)


lib/ansible/module_utils/basic.py
```diff
@@ -1338,6 +1351,23 @@ def _check_argument_values(self):
                 continue
             if isinstance(choices, SEQUENCETYPE):
                 if k in self.params:
+                    # PyYaml converts certain strings to bools.  If we can unambiguously convert back, do so before checking the value.  If we can't figure this out, module author is responsible.
+                    lowered_choices = None
+                    if self.params[k] == 'False':
```
bcoca: shouldn't you only check if it does not already match the choices? Just in case 'False' or 'True' are valid choices

abadger:
Small change so we don't have to try to convert back to string if the parameter is already in choices.

abadger:
@tmshn Thanks for reviewing.  Merged to devel and cherry-picked to stable-2.1

PR #11658, opened 2015-07-20
Integrate EUI64 SLAAC into ipaddr() Jinja2 filter

vctgross:
##### SUMMARY
usage: {{ 2db8::/64 | ipaddr(l2addr) }}
- Enforce RFC 4862 5.5.3 and ignore networks with a prefixlen other than 64
- does not require a separate keyword, acts as if l2addr was an index in the
  subnet
- does not work with bare hwaddr (such as "001122334455") to keep templates
  like {{ 2db8::/32 | ipaddr(32) }} working. Better force hwaddr format as in
  {{ 2db8::/64 | ipaddr(l2addr|hwaddr('eui48')) }} when unsure.

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
lib/ansible/plugins/filter/ipaddr.py

##### ANSIBLE VERSION
2.3

PR #11394, opened 2015-06-25
IPv6 SLAAC address computation filter

dlundquist:
Jinja2 filter to compute SLAAC address.

Usage:
    {{ '2db8::/64' | slaac(ansible_eth0.macaddress) }}

PR #16976, opened 2016-08-05
Refactor and fix the slaac filter.

chrko:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (slaac_fix_and_include_prefix 707622ccf2) last updated 2016/08/05 16:44:53 (GMT +200)
  lib/ansible/modules/core: (detached HEAD 24db4de245) last updated 2016/08/05 16:41:56 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD db979dde74) last updated 2016/08/05 16:41:56 (GMT +200)
  config file = /home/christian/.ansible.cfg
```
##### SUMMARY

The implementation of the slaac filter doesn't work (and may have never worked; introduced in cf7744f2f131708acd67c1312f622a3d4e639455 #11394). It calls a method on string instead of the corresponding `netaddr`-Class, see [lib/ansible/plugins/filter/ipaddr.py @ cf7744f](https://github.com/ansible/ansible/commit/cf7744f2f131708acd67c1312f622a3d4e639455#diff-52b5d47bc5ffb759787765a1aae56df7R603) L603 vs L606.

This implementation tries to do be more a wrapper to the existing functions instead of reinventing the wheel. Only respecting the RFC requires the code around the lines L635-L644.

While developing a fix for the slaac filter I also tried another approach. This will be a separate pull request.

Due to this pull request #11658 will be obsolete.

### DIFF
--- a/lib/ansible/plugins/filter/ipaddr.py
+++ b/lib/ansible/plugins/filter/ipaddr.py
@@ -255,6 +255,18 @@ def _win_query(v):
     v.dialect = netaddr.mac_eui48
     return str(v)
 
+def _eui64_query(v):
+    return str(v.eui64())
+
+def _eui64_int_query(v):
+    return int(v.eui64())
+
+def _modified_eui64_query(v):
+    return str(v.modified_eui64())
+
+def _modified_eui64_int_query(v):
+    return int(v.modified_eui64())
+
 
 # ---- IP address and network filters ----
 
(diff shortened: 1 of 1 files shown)


chrko:
I have rebased the pull request. Now there is no more randomly failed build.

calfonso:
@chrko We are catching up on PRs and we're interested in getting this one merged, if you could rebase it first then we'll get it reviewed. Thanks!

chrko:
Ok, I will so this the following week. Thank you for the notification!

chrko:
Yeah. Checked it again. The main bug got fixed by #17551.
I'll probably create a new pull request to extend the behaviour of the slaac filter to meet the requirements of the RFC. (Assume always a 64bit prefix or treat everthing else than an 64-bit ipv6 network as an error).
#16979 is also obsolete now.

Issue #16888, opened 2016-07-31
document module_utils.*


kontrafiktion:
##### ISSUE TYPE
- Documentation Report
##### COMPONENT NAME

internal: module_utils.*
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file =
  configured module search path = Default w/o overrides
```
##### SUMMARY

Module developers/maintainers have a hard time using module_utils.\* because there is nearly no documentation. At least I couldn't find any.

Proposal:
1. Create documentation for module_utils.*
2. Publish it on the Ansible web site
###### Details

For example, the fetch_url method does not document its return values. You have to debug and know the internally used urllib2 to understand what the "info" return value contains. 

When creating a new module, using the "requests" library, the build says: 

```
8.60s$ ansible-validate-modules --exclude 'utilities/' .
============================================================================
./cloud/digital_ocean/digital_ocean_tag.py
============================================================================
ERROR: requests import found, should use ansible.module_utils.urls instead
```

Therefore: when asking (demanding) to use a library it should be documented (as thoroughly as possible). Otherwise the entry hurdle becomes rather high

bcoca:
We had just discussed this at the contributor summit last week.

kontrafiktion:
See for a very early version: https://github.com/ansible/ansible/pull/16890 (superceded by #16987)

kbreit:
Is there still effort around this? I'm new to Ansible and developing modules and I've found the lack of documentation around module_utils to be challenging. I'd be happy to help on this, but probably don't have time to take it all on myself.

danihodovic:
I wish the `ansible.module_utils.urls` provided a  simple way to log requests / responses at HTTP level.

acozine:
@kontrafiktion thanks for opening this issue and #16987!

Since then we've expanded and improved the documentation on `module_utils` - the latest changes came from PR #55931. 

@kontrafiktion @kbreit could you take a look at the [latest version of this page](https://docs.ansible.com/ansible/devel/dev_guide/developing_module_utilities.html) on the `devel` docs and let me know if we can close this issue? The content will be published to the `latest` docs when Ansible 2.8 is released later this month.

samccann:
Documentation has been updated (see prior comment). If more is missing, please open individual issues/PRs to address what is lacking. Thanks!

PR #16890, opened 2016-07-31
first documentation of module_utils

kontrafiktion:
##### ISSUE TYPE
- Docs Pull Request
##### ANSIBLE VERSION

ansible 2.1.0.0
  config file =
  configured module search path = Default w/o overrides

```

##### SUMMARY
#16888
```

PR #16987, opened 2016-08-07
16888: comment module_utils/urls.py

kontrafiktion:
##### ISSUE TYPE
- Docs Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0.0
config file =
```
##### SUMMARY

one step for #16888

### DIFF
--- a/lib/ansible/module_utils/urls.py
+++ b/lib/ansible/module_utils/urls.py
@@ -899,36 +899,35 @@ def url_argument_spec():
 
 def fetch_url(module, url, data=None, headers=None, method=None,
               use_proxy=True, force=False, last_mod_time=None, timeout=10):
-    '''
-    Sends a request via HTTP(S) or FTP (needs the module as parameter)
+    '''Sends a request via HTTP(S) or FTP (needs the module as parameter)
 
-    Args:
-        :module (AnsibleModule): The AnsibleModule (used to get username, password etc. (s.b.).
-        :url:                    The url to use.
-        :data:                   The data to be sent (in case of POST/PUT).
-        :headers (dict):         A dict with the request headers.
-        :method (String):        "POST", "PUT", etc.
-        :use_proxy (Boolean):    Default: True
-        :force (Boolean):        If True: Do not get a cached copy (Default: False)
-        :last_mod_time:          Default: None
-        :timeout (int):          Default: 10
+    :arg module: The AnsibleModule (used to get username, password etc. (s.b.).
+    :arg url:             The url to use.
 
-    Returns:
-        :response: the response (body). Use `resp.body()` to read the data.
-        :info:     status code and other meta data. When a HttpError (status > 400) occurred
-                   then info['body'] contains the error response data::
+    :kwarg data:          The data to be sent (in case of POST/PUT).
+    :kwarg headers:       A dict with the request headers.
+    :kwarg method:        "POST", "PUT", etc.
+    :kwarg boolean use_proxy:     Default: True
+    :kwarg boolean force: If True: Do not get a cached copy (Default: False)
+    :kwarg last_mod_time: Default: None
+    :kwarg int timeout:   Default: 10
 
-    Example:
-        >>> data={...}
-            resp, info = fetch_url("http://example.com", 
-                                   data=module.jsonify(data)
-                                   header={Content-type': 'application/json'},
-                                   method="POST")
-            status_code = info["body"]
-            body = resp.read()
-            if status_code >= 400 :
-                body = info['body']
-    '''
+    :returns: A tuple of (**response**, **info**). Use ``response.body()`` to read the data.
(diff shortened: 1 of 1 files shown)


jctanner:
@kontrafiktion I like this, but I believe you added tabs on lines 741-743

kontrafiktion:
@jctanner no

kontrafiktion:
Sorry, but I will not invest more time in this documentation change

lib/ansible/module_utils/urls.py
```diff
@@ -81,6 +81,16 @@
...
 
+'''
+The **urls** utils module offers a replacement for the `requests`_ library for most cases. 
+
+Sometimes you really need the requests library, but you should always check, whether 
+this utils module is sufficient for your use case.
+
+.. _requests:
+   http://docs.python-requests.org/en/master/
+'''
```
abadger, line 97: This module is really a replacement for `urllib.openurl()` which should never be used because it suffers from security limitation.  requests has similar security features as this module_util but is an external dependency and therefore would be an additional requirement on the remote machine.

lib/ansible/module_utils/urls.py
```diff
@@ -728,11 +738,11 @@ def maybe_add_ssl_handler(url, validate_certs):
...
 def open_url(url, data=None, headers=None, method=None, use_proxy=True,
-        force=False, last_mod_time=None, timeout=10, validate_certs=True,
-        url_username=None, url_password=None, http_agent=None,
-        force_basic_auth=False, follow_redirects='urllib2'):
+             force=False, last_mod_time=None, timeout=10, validate_certs=True,
+             url_username=None, url_password=None, http_agent=None,
+             force_basic_auth=False, follow_redirects='urllib2'):
     '''
-    Fetches a file from an HTTP/FTP server using urllib2
+    Sends a request via HTTP(S) or FTP using urllib2 (Python2) or urllib (Python3)
```
abadger, line 750: As long as you're clarifying the documentation, we could drop urllib2/urllib mention here.  It's not important to users of the function.

lib/ansible/module_utils/urls.py
```diff
@@ -870,22 +880,49 @@ def url_argument_spec():
...
+    Example:
+        >>> data={...}
+            resp, info = fetch_url("http://example.com", 
+                                   data=module.jsonify(data)
+                                   header={Content-type': 'application/json'},
+                                   method="POST")
+            status_code = info["body"]
+            body = resp.read()
+            if status_code >= 400 :
+                body = info['body']
```
abadger: Please use Sphinx's resturctured text format for API documentation.  There's an example of a method using that style of documentation here:  https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/connection/__init__.py#L159

We'll be building the html version of the documentation using sphinx so following that format will let these docstrings be used.  Sphinx's documentation is here: http://www.sphinx-doc.org/en/stable/tutorial.html

lib/ansible/module_utils/urls.py
abadger, line 97: Hmm.... I see in the original bug that ansible-modules-validate is throwing an error for using requests and that was part of what prompted this doc update.  That does make it a good idea to mention requests here.... What about something like:

```
The **urls** utils module offers a replacement for the urllib2 python library.

urllib2 is the python stdlib way to retrieve files from the Internet but it lacks some security features (around verifying SSL certificates) that users should care about in most situations.  using the functions in this module corrects deficiencies in the urllib2 module wherever possible.

There are also third-party libraries (for instance, requests) which can be used to replace urllib2 with a more secure library.  However, all third party libraries require that the library be installed on the managed machine.  That is an extra step for users making use of a module.  If possible, avoid third party libraries by using this code instead.
```

lib/ansible/module_utils/urls.py
kontrafiktion: @abadger what of the above is not restructured text? In #16890 I had the documentation generated and it looked fine to me. The only additions compared to ".../**init**.py#L159" I can see, are:
- The documentation in this PR explicitly states the type of the parameter in some cases
- "Args:" and "Returns:" are added

The documentation at http://www.sphinx-doc.org/en/stable/domains.html#signatures defines a different format. But when I tried that, it rendered the "param" string literally.

abadger: It does the right thing for me.  I whipped up a quick container page:

```
$ cat out/index.rst
.. automodule:: ansible.plugins.connection
    :members:
```

and when I generated it with sphinx-build I got this output for the exec_command sample: https://toshio.fedorapeople.org/ansible/test-autodoc/#ansible.plugins.connection.ConnectionBase.exec_command

abadger:
Thanks, @kontrafiktion Looks good!  I've merged this to devel.

PR #16997, opened 2016-08-08
doas update

### LINKED ISSUE #15597: Become with doas not working
##### ISSUE TYPE
- Bug Report
##### ANSIBLE VERSION

```
ansible 2.0.0.2
```
##### CONFIGURATION

```
[privilege_escalation]
become=True
become_method=doas
become_user=root
become_ask_pass=True
```
##### OS / ENVIRONMENT

From: Mac OS X 10.11
To: OpenBSD 5.9 AMD64
##### SUMMARY

The doas prompt definition in play_context.py is incorrect. Currently it is set to 'Password:' but the doas command uses the following layout 'doas (user@host.domain.com) password: ' with user, host and domain replaced with the actual values. When I replace the prompt with the values for my server it will work, however, since it will only work with that particular server this is not an option. I guess that using wildcards would be an option.
##### STEPS TO REPRODUCE

By setting become to doas and sending a ping to the host.

```
ansible host.domain.com -m ping -K
```
##### EXPECTED RESULTS

```
host.domain.com | SUCCESS => {
    "changed": false,
    "ping": "pong"
}
```
##### ACTUAL RESULTS

```
host.domain.com | FAILED! => {
    "failed": true,
    "msg": "ERROR! Timeout (12s) waiting for privilege escalation prompt: \r"
}
```

bcoca:
##### ISSUE TYPE

- Bugfix Pull Request
##### ANSIBLE VERSION

```
2.2
```
##### SUMMARY

Adapts to changes to doas prompts, this is backwards incompatible with older version of doas, which should be fine as it is a very new tool and there should not be large base depending on it.

hopefully fixes #15597

### DIFF
--- a/lib/ansible/playbook/play_context.py
+++ b/lib/ansible/playbook/play_context.py
@@ -534,7 +534,7 @@ class PlayContext(Base):
 
             elif self.become_method == 'doas':
 
-                prompt = 'Password:'
+                prompt = 'doas (%s@' % self.remote_user
                 exe = self.become_exe or 'doas'
 
                 if not self.become_pass:


alikins:
Looks good to me.

PR #17000, opened 2016-08-08
Allow to make the jsonfile cache files pretty (indented and sorted)

dagwieers:
##### ISSUE TYPE

- Feature Pull Request
- Docs Pull Request
##### ANSIBLE VERSION

v2.1.1.0
##### SUMMARY

Since the json cache files are condensed, it is not very practical to look for something in them. Having indented/sorted cache files makes debugging and playbook/inventory development a lot easier to do.

I made it configurable (using `fact_caching_format`) in case people would object to the performance hit this would have, but to be honest, then they probably should be looking at other cache plugins instead.

### DIFF
--- a/docsite/rst/playbooks_variables.rst
+++ b/docsite/rst/playbooks_variables.rst
@@ -547,6 +547,7 @@ To configure fact caching using redis, enable it in ``ansible.cfg`` as follows::
     gathering = smart
     fact_caching = redis
     fact_caching_timeout = 86400
+    # seconds
 
 To get redis up and running, perform the equivalent OS commands::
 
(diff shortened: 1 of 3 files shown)


bcoca:
seems a lot of scaffolding for something you can do in a one liner: `python -m json.tool < cache.json`

dagwieers:
Well, not if you have more than 4000 files and regularly compare files or grep through all files. I'd been using **json_reformat**, but it's a hassle (a lot of scaffolding, as you will) compared to this easy fix.

And that's why I modified my own code first, but I consider it useful to others (and do not want to maintain modified code). Maybe the option is overkill and we should have pretty json files by default.

Voila, turned it into a simple improvement, not a configurable option.

dagwieers:
Voila, turned it into a simple improvement, not a configurable option.

PR #17003, opened 2016-08-08
Fix from @jimi-c to use task_vars when polling async tasks.

### LINKED ISSUE #14101: async_status file not using ansible_python_interpreter (was: Ansible 2.0.0.2 async polling is not working)
This is with PCBSD 10.2 system running Ansible to configure a digital ocean FreeBSD 10.2 droplet.

This is the task:

```
- name: Just testing
  command: sleep 120
  async: 7200
  poll: 3
```

The problems are:
- With the poll present; the process stops after only 10 seconds with the error: "async task did not complete within the requested time".
- With poll line removed ;then no error and the playbook continues after approximately 4 seconds and completes without error.

In both cases logging in with ssh to the droplet shows the sleep 120 is still running, and once complete the a zero size file in ".ansible_async" gets the results added to it (the results file looks ok).

Notes:
- The intsaledl Ansible 2.002 has `plugins/action/__init__.py` from devel branch to allow csh to work.
- I also added user with zsh which gave exactly the same results.

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
devel and other 2.x
```
##### SUMMARY

This is needed so that async_status can use ansible_python_interpreter
for the host being contacted.

Fixes #14101

### DIFF
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -468,7 +468,7 @@ class TaskExecutor:
 
             if self._task.async > 0:
                 if self._task.poll > 0:
-                    result = self._poll_async_result(result=result, templar=templar)
+                    result = self._poll_async_result(result=result, templar=templar, task_vars=vars_copy)
 
                 # ensure no log is preserved
                 result["_ansible_no_log"] = self._play_context.no_log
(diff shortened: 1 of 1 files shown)


abadger:
Merged to devel and stable-2.1

PR #17034, opened 2016-08-10
Fix classes that select a subclass based on platform

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

ansible/module_utils/facts.py
##### ANSIBLE VERSION

```
devel
```
##### SUMMARY

When unittesting this we found that the platform selecting class
hierarchies weren't working in all cases.  If the subclass was directly
created (ie: LinuxHardware()), then it would use its inherited **new**()
to try to create itself.  The inherited **new** would look for
subclasses and end up calling its own **new**() again.  This would
recurse endlessly.  The new code detects when we want to find a subclass
to create (when the base class is used, ie: Hardware()) vs when to
create the class itself (when the subclass is used, ie:
LinuxHardware()).

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -961,6 +961,12 @@ class Hardware(Facts):
     platform = 'Generic'
 
     def __new__(cls, *arguments, **keyword):
+        # When Hardware is created, it chooses a subclass to create instead.
+        # This check prevents the subclass from then trying to find a subclass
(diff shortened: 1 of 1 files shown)


alikins:
Works for me if  (cherry-picked into https://github.com/ansible/ansible/pull/17036 that has test cases for LinuxHardware and change tests to create LinuxHardware directly).

Issue #10779, opened 2015-04-20
get_mount_facts timeout is very small and not ignored as expected


rosowiecki:
I've analysed #10746 more deeply and found that my first suspition was wrong and timeout error was not really caused by lack of lsblk. Moreover, it was not ingored as intened.

The reason is that UUID getting added by #10055 broke timeout handling because of Python's subprocess module code. In _execute_child method there's an except code catching all exceptions: that's why TimeoutError is not caught in  Facts.populate and ignored.

```
failed: [SOME_HOST] => {"cmd": "/bin/lsblk -ln --output UUID /dev/sdi2", "failed": true, "rc": 257}
msg: Traceback (most recent call last):
  File "<stdin>", line 1650, in run_command
  File "/usr/lib64/python2.4/subprocess.py", line 550, in __init__
    errread, errwrite)
  File "/usr/lib64/python2.4/subprocess.py", line 988, in _execute_child
    data = os.read(errpipe_read, 1048576) # Exceptions limited to 1 MB
  File "<stdin>", line 1796, in _handle_timeout
TimeoutError: Timer expired
```

Suggested solution: get UUID's without calling lsblk with subprocess.

Margin note: 10s timeout was enough before, but with lsblk get_mount_facts on my system with no network filesystems mounted lasted even 15 seconds.

bcoca:
I disagree that 10s is too little for listing UUIDs and in the many systems i've tested lsblk takes less than 1s. Keeping this open to investigate the issue about the timeouts not being handled properly.

bitmensch:
I ran into the same issue some days ago. In my particular case I have a install-once-and-throw-away test environment where I'm using docker containers for the underlying system.

It turns out that in this case the problem was already known (see https://github.com/docker/docker/issues/12192) and that upgrading the kernel of my docker host system resolved the issue.

epyonavenger:
In our production environment, we see this come up fairly often since the servers are pretty busy and not particularly high-performance. When RAID is rebuilding it's especially common. Even if the default stays the same, it would be nice if we could override or otherwise ignore when it isn't pertinent.

infernix:
We are seeing this happen a lot on CEPH nodes with LSI SAS controllers. It's extremely random.

`failed: [obj-005] => {"cmd": "/bin/lsblk -ln --output UUID /dev/sdb1", "failed": true, "rc": 257}`
`msg: Traceback (most recent call last):`
`File "<stdin>", line 1671, in run_command`
[...]

PR #17036, opened 2016-08-10
Collect block device uuids in one lsblk call so facts collection doesn't timeout 

alikins:
##### ISSUE TYPE

```
ansible 2.2.0 (stop_collecting_lsblk_uuid_10779 55210edb54) last updated 2016/08/10 14:48:26 (GMT -400)
  lib/ansible/modules/core: (detached HEAD a216ef210b) last updated 2016/08/10 09:47:29 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 39153ea154) last updated 2016/08/10 09:47:29 (GMT -400)


```
##### SUMMARY

Work in progress fix for https://github.com/ansible/ansible/issues/10779#issuecomment-238731228

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -38,7 +38,9 @@ try:
 except ImportError:
     # python3
     import configparser
+
 from ansible.module_utils.basic import get_all_subclasses
+from ansible.module_utils.six import PY3
 
 # py2 vs py3; replace with six via ansiballz
 try:
@@ -971,7 +973,10 @@ class Hardware(Facts):
         for sc in get_all_subclasses(Hardware):
             if sc.platform == platform.system():
                 subclass = sc
-        return super(cls, subclass).__new__(subclass, *arguments, **keyword)
+        if PY3:
+            return super(cls, subclass).__new__(subclass)
+        else:
+            return super(cls, subclass).__new__(subclass, *arguments, **keyword)
 
     def populate(self):
         return self.facts
@@ -996,6 +1001,12 @@ class LinuxHardware(Hardware):
     ORIGINAL_MEMORY_FACTS = frozenset(('MemTotal', 'SwapTotal', 'MemFree', 'SwapFree'))
     # Now we have all of these in a dict structure
     MEMORY_FACTS = ORIGINAL_MEMORY_FACTS.union(('Buffers', 'Cached', 'SwapCached'))
+
+    # regex used against findmnt output to detect bind mounts
+    BIND_MOUNT_RE = re.compile(r'.*\]')
+
+    # regex used against mtab content to find entries that are bind mounts
+    MTAB_BIND_MOUNT_RE = re.compile(r'.*bind.*"')
 
     def populate(self):
         self.get_cpu_facts()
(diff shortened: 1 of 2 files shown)


lib/ansible/module_utils/basic.py
```diff
@@ -311,6 +311,10 @@ def get_all_subclasses(cls):
     subclasses = cls.__subclasses__()
     to_visit = list(subclasses)
     # Then visit all subclasses
+    print('to_visit: %s' % to_visit)
```
alikins: I'll clean up the debug/logging stuff shortly...

test/units/module_utils/test_facts.py
```diff
@@ -0,0 +1,463 @@
...
+
+
+class TestFactsHardwareNew:
+    """Verify that the automagic in Hardware.__new__ selects the right subclass."""
+    @patch('platform.system')
+    def _hardware_new(self, platform_id, hardware_class, mock_platform):
+        mock_platform.return_value = platform_id
+        hardware = facts.Hardware(module=Mock(), load_on_init=False)
+        assert_msg = '%s should have been created for platform=%s' % (hardware_class, platform_id)
+        assert isinstance(hardware, hardware_class), assert_msg
```
abadger: We get good messages for free if we use unittest.assertIsinstance(), assertEq(), etc.  Need to inherit from unittest.TestCase to have access to those.

alikins:
The tests for the facts.Hardware() factory changed to subclass TestCase and replaced with test classes for each facts.Facts() subclasses.

test/units/module_utils/test_facts.py
```diff
@@ -0,0 +1,464 @@
...
+        assert '/not/a/real/bind_mount' in bind_mounts, 'expected /not/a/real/bind_mount in bind_mounts but it is missing'
+
+    @patch('ansible.module_utils.facts.LinuxHardware._run_lsblk', return_value=(0, LSBLK_OUTPUT,''))
+    def test_lsblk_uuid(self, mock_run_lsblk):
+        module = Mock()
+        lh = facts.LinuxHardware(module=module, load_on_init=False)
+        lsblk_uuids = lh._lsblk_uuid()
+        self.assertIsInstance(lsblk_uuids, dict)
+        assert '/dev/loop9' in lsblk_uuids
+        assert '/dev/sda1' in lsblk_uuids
```
abadger: add a message or use

``` python
self.assertIn('/dev/loop9', lsblk_uuids)
self.assertIn('/dev/sda1', lsblk_uuids)
```

lib/ansible/module_utils/facts.py
```diff
@@ -1208,58 +1213,122 @@ def get_dmi_facts(self):
...
+        return uuids
+
+    def _run_findmnt(self, findmnt_path):
+        args = ['--list', '--noheadings', '--notruncate']
+        cmd = [findmnt_path] + args
+        rc, out, err = self.module.run_command(cmd)
+        return rc, out, err
+
+    def _find_bind_mounts(self):
         bind_mounts = []
```
abadger: make this a set to be a bit more efficient.  (We're only using it for containment checks.)

lib/ansible/module_utils/facts.py
```diff
@@ -1208,58 +1213,122 @@ def get_dmi_facts(self):
...
-                # find bind mounts, in case /etc/mtab is a symlink to /proc/mounts
-                for line in out.split('\n'):
-                    fields = line.rstrip('\n').split()
-                    if(len(fields) < 2):
-                        continue
-                    if(re.match(".*\]",fields[1])):
-                        bind_mounts.append(fields[0])
+        findmnt_path = self.module.get_bin_path("findmnt")
+        if not findmnt_path:
+            return
```
abadger: Need to `return bind_mounts` here

lib/ansible/module_utils/facts.py
```diff
@@ -1208,58 +1213,122 @@ def get_dmi_facts(self):
...
+        # find bind mounts, in case /etc/mtab is a symlink to /proc/mounts
+        lines = out.splitlines()
+        for line in lines:
+            fields = line.split()
+            # fields[0] is the TARGET, fields[1] is the SOURCE
+            if len(fields) < 2:
+                continue
+
+            # bind mounts will have a [/directory_name] in the SOURCE column
+            if re.match(".*\]",fields[1]):
```
abadger: Since we're optimizing to solve this bug, would be good to move regex compilation up to be a one-time thing.  Maybe a class attribute or global for each compiled regex.

``` python

class LinuxHardware(Hardware): 
    BIND_MOUNT_RE = re.compile(r".*\]")
[...]
if re.match(self.BIND_MOUNT_RE, fields[1]):
```

lib/ansible/module_utils/facts.py
```diff
@@ -1208,58 +1213,122 @@ def get_dmi_facts(self):
...
+                bind_mounts.append(fields[0])
```
abadger: If we make bind_mounts a set, then this becomes:

``` python
bind_mounts.add(fields[0])
```

lib/ansible/module_utils/facts.py
```diff
@@ -1208,58 +1213,122 @@ def get_dmi_facts(self):
...
+
+        # each line will be in format:
+        # <devicename><some whitespace><uuid>
+        # /dev/sda1  32caaec3-ef40-4691-a3b6-438c3f9bc1c0
+        for lsblk_line in lsblk_lines:
+            if not lsblk_line:
+                continue
+
+            line = lsblk_line.strip()
+            fields = line.split()
```
abadger: The comment on line 1235 says that we'll always have [filepath][whitespace][uuid].  Is that true?  If so we should calculate fields like this:

``` python
fields = line.rsplit(1)
```

That way we'll handle systems which might have spaces in the filepath portion.

lib/ansible/module_utils/facts.py
```diff
@@ -1208,58 +1213,122 @@ def get_dmi_facts(self):
...
 
+        return bind_mounts
+
+    def _mtab_entries(self):
         mtab = get_file_content('/etc/mtab', '')
-        for line in mtab.split('\n'):
+        lines = mtab.splitlines()
+        mtab_entries = []
+        for line in lines:
             fields = line.rstrip('\n').split()
```
abadger: I think you can get rid of the rstrip('\n') here for the same reason as you got rid of it in _find_bind_mounts() (or alternatively, if rstrip is necessary here, then it's probably still needed in _find_bind_mounts() as well).

lib/ansible/module_utils/facts.py
```diff
@@ -1208,58 +1213,122 @@ def get_dmi_facts(self):
...
+    def get_mount_facts(self):
+        self.facts['mounts'] = []
+
+        bind_mounts = self._find_bind_mounts()
+        uuids = self._lsblk_uuid()
+        mtab_entries = self._mtab_entries()
+
+        mounts = []
+        for fields in mtab_entries:
+            device, mount, fstype, options = fields[0], fields[1], fields[2], fields[3]
```
abadger, line 1307: More pythonic to combine this unpacking with the for loop:

``` python
for device, mount, fstype, options in mtab_entries:
```

lib/ansible/module_utils/facts.py
```diff
@@ -1208,58 +1213,122 @@ def get_dmi_facts(self):
...
+        if not findmnt_path:
+            return
+
+        rc, out, err = self._run_findmnt(findmnt_path)
+        if rc != 0:
+            return bind_mounts
+
+        # find bind mounts, in case /etc/mtab is a symlink to /proc/mounts
+        lines = out.splitlines()
+        for line in lines:
```
abadger: style note on all of these.. Unless you need lines later, it's more idiomatic python not to allocate a named temporary variable here.  Instead use out.splitlines() directly:

``` python
for line in out.splitlines():
```

lib/ansible/module_utils/facts.py
```diff
@@ -1208,58 +1213,122 @@ def get_dmi_facts(self):
...
+                continue
+
+            if fstype == 'none':
+                continue
+
+            size_total, size_available = self._get_mount_size_facts(mount)
+
+            if mount in bind_mounts:
+                # only add if not already there, we might have a plain /etc/mtab
+                if not re.match(".*bind.*", options):
```
abadger: Another regex that can be precompiled.

abadger:
Okay, reviewed all the changes.  Found a couple bugs, a few readable optimizations, and a few pure-style changes to make.  Once those changes are all made, this is good to merge.

lib/ansible/module_utils/facts.py
alikins: Wasn't sure if we ever cared about the ordering, but we don't seem to so a set makes sense.

lib/ansible/module_utils/facts.py
alikins: I like giving the regex a name, the pre compile is a bonus.

(Not sure there would be any noticeable performance changes, even with the >1000 block devices some users mentioned. Given the simple regex and the regex caching. That regex will get more complicated eventually though...)

(note while it's in my head, generally don't like logic in class def scope, but this is straightforward and given the unusual **init**'s here...)

abadger: <nod>  Before working on ansible, I'd do regex compilation at module scope.  mpdehaan and jimi-c liked things at class level so I've put more things there but it still doesn't feel natural in python...  Global scope and in the `__init__` should both do the right thing for this piece of code (the classes are only instantiated once so there's no performance disadvantage to compiling in `__init__`; I think that **init** should function normally as all of our `__new__` methods are returning a class.)

lib/ansible/module_utils/facts.py
alikins, line 1307: This is to guard against the size of fields changing and causing tuple unpacking errors.
 _mtab_entries() will potentially return entries with > 4 items.
Though the code to parse the entries  is already a bit split across _mtab_entries and get_mount_facts so that could be enforced there.

lib/ansible/module_utils/facts.py
alikins: The rstrip() usage that was used for parsing findmnt output (now in _find_bind_mounts()) was in a section of code that was never executed due to wrong args to findmnt, so it wasn't worth trusting. The mtab splitting code at least appears to work in the wild, so I left it as is. afaict, it isn't needed...

alikins:
added some test cases that would have caught the missing 'return bind_mounts'

lib/ansible/module_utils/facts.py
abadger: Right -- it shouldn't be needed because splitlines() will remove all "\n".

abadger:
- The shippable test failure looks to be a timeout creating the freebsd container so that's fine.
- The rstrip() after splitlines is extraneous and can be removed but isn't a blocker.
- All other notes taken care of.

+1 to merge.

PR #17037, opened 2016-08-10
add cloudforms inventory script

mrjoshuap:
[cloudforms--list.json.txt](https://github.com/ansible/ansible/files/412082/cloudforms--list.json.txt)
[cloudforms--host.json.txt](https://github.com/ansible/ansible/files/412086/cloudforms--host.json.txt)
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

Contributed inventory for Red Hat CloudForms / ManageIQ
##### ANSIBLE VERSION

```
ansible 2.1.0.0
```
##### SUMMARY

This is a new contributed inventory script for Red Hat CloudForms / ManageIQ.  I tested with Red Hat CloudForms 4.1.  See attached output examples.

```
usage: cloudforms.py [-h] [--list] [--host HOST] [--refresh-cache] [--debug]

Produce an Ansible Inventory file based on CloudForms managed VMs

optional arguments:
  -h, --help       show this help message and exit
  --list           List instances (default: True)
  --host HOST      Get all the variables about a specific instance
  --refresh-cache  Force refresh of cache by making API requests to cloudforms
                   (default: False - use cache files)
  --debug          Show debug output while running (default: False)
```

based on the foreman inventory script, features:
- cached results (default 600 seconds)
- paginated host results (default 100 hosts)
- ssl verification (default True)
- arguments to flush cache and run in debug mode

### DIFF
--- a/contrib/inventory/cloudforms.py
+++ b/contrib/inventory/cloudforms.py
@@ -161,7 +161,7 @@ class CloudFormsInventory(object):
         if config.has_option('cloudforms', 'purge_actions'):
             self.cloudforms_purge_actions = config.getboolean('cloudforms', 'purge_actions')
         else:
-            self.cloudforms_purge_actions = False
+            self.cloudforms_purge_actions = True
 
         if config.has_option('cloudforms', 'clean_group_keys'):
             self.cloudforms_clean_group_keys = config.getboolean('cloudforms', 'clean_group_keys')
--- a/contrib/inventory/cloudforms.py
+++ b/contrib/inventory/cloudforms.py
@@ -303,7 +303,7 @@ class CloudFormsInventory(object):
                         safe_key = self.to_safe(group['name'])
                         if safe_key:
                             if self.args.debug:
-                                print "Adding sub-group '%s' to parent 'tags'" % safe_tag_name
+                                print "Adding sub-group '%s' to parent 'tags'" % safe_key
 
                             if safe_key not in self.inventory['tags']['children']:
                                 self.push(self.inventory['tags'], 'children', safe_key)
--- a/contrib/inventory/cloudforms.py
+++ b/contrib/inventory/cloudforms.py
@@ -228,7 +228,7 @@ class CloudFormsInventory(object):
         try:
             results = json.loads(ret.text)
         except ValueError:
-            warnings.warn("Unexpected response from {0} ({1}): {2}".format(self.cloudforms_url, r.status_code, r.reason))
+            warnings.warn("Unexpected response from {0} ({1}): {2}".format(self.cloudforms_url, ret.status_code, ret.reason))
             results = {}
 
         if self.args.debug:
--- a/contrib/inventory/cloudforms.py
+++ b/contrib/inventory/cloudforms.py
@@ -34,6 +34,7 @@ try:
 except ImportError:
     import simplejson as json
 
+
 class CloudFormsInventory(object):
     def __init__(self):
         """
--- a/contrib/inventory/cloudforms.py
+++ b/contrib/inventory/cloudforms.py
@@ -168,6 +168,11 @@ class CloudFormsInventory(object):
         else:
             self.cloudforms_clean_group_keys = True
 
+        if config.has_option('cloudforms', 'nest_tags'):
+            self.cloudforms_nest_tags = config.getboolean('cloudforms', 'nest_tags')
+        else:
+            self.cloudforms_nest_tags = False
+
         # Ansible related
         try:
             group_patterns = config.get('ansible', 'group_patterns')
@@ -282,12 +287,62 @@ class CloudFormsInventory(object):
 
             # Create ansible groups for tags
             if 'tags' in host:
-                for group in host['tags']:
-                    safe_key = self.to_safe(group['name'])
-                    self.push(self.inventory, safe_key, host['name'])
 
-                    if self.args.debug:
-                        print "Found tag [%s] for host which will be mapped to [%s]" % (group['name'], safe_key)
+                # Create top-level group
+                if 'tags' not in self.inventory:
+                    self.inventory['tags'] = dict(children=[], vars={}, hosts=[])
+
+                if not self.cloudforms_nest_tags:
+                    # don't expand tags, just use them in a safe way
+                    for group in host['tags']:
+                        # Add sub-group, as a child of top-level
+                        safe_key = self.to_safe(group['name'])
+                        if safe_key:
+                            if self.args.debug:
+                                print "Adding sub-group '%s' to parent 'tags'" % safe_tag_name
+
+                            if safe_key not in self.inventory['tags']['children']:
+                                self.push(self.inventory['tags'], 'children', safe_key)
+
+                            self.push(self.inventory, safe_key, host['name'])
+
+                            if self.args.debug:
+                                print "Found tag [%s] for host which will be mapped to [%s]" % (group['name'], safe_key)
+                else:
+                    # expand the tags into nested groups / sub-groups
+                    # Create nested groups for tags
+                    safe_parent_tag_name = 'tags'
+                    for tag in host['tags']:
+                        tag_hierarchy = tag['name'][1:].split('/')
+
+                        if self.args.debug:
+                            print "Working on list %s" % tag_hierarchy
+
+                        for tag_name in tag_hierarchy:
+                            if self.args.debug:
+                                print "Working on tag_name = %s" % tag_name
+
+                            safe_tag_name = self.to_safe(tag_name)
+                            if self.args.debug:
+                                print "Using sanitized name %s" % safe_tag_name
+
+                            # Create sub-group
+                            if safe_tag_name not in self.inventory:
+                                self.inventory[safe_tag_name] = dict(children=[], vars={}, hosts=[])
+
+                            # Add sub-group, as a child of top-level
+                            if safe_parent_tag_name:
+                                if self.args.debug:
+                                    print "Adding sub-group '%s' to parent '%s'" % (safe_tag_name, safe_parent_tag_name)
+
+                                if safe_tag_name not in self.inventory[safe_parent_tag_name]['children']:
+                                    self.push(self.inventory[safe_parent_tag_name], 'children', safe_tag_name)
+
+                            # Make sure the next one uses this one as it's parent
+                            safe_parent_tag_name = safe_tag_name
+
+                        # Add the host to the last tag
+                        self.push(self.inventory[safe_parent_tag_name], 'hosts', host['name'])
 
             # Set ansible_ssh_host to the first available ip address
             if 'ipaddresses' in host and host['ipaddresses'] and isinstance(host['ipaddresses'], list):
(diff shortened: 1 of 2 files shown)


contrib/inventory/cloudforms.py
```diff
@@ -0,0 +1,363 @@
...
+        if self.args.host:
+            if self.debug:
+                print "Fetching host [%s]" % self.args.host
+            data_to_print += self.get_host_info(self.args.host)
+        else:
+            self.inventory['_meta'] = {'hostvars': {}}
+            for hostname in self.cache:
+                self.inventory['_meta']['hostvars'][hostname] = {
+                    'cloudforms': self.cache[hostname],
+                    'cloudforms_params': self.params[hostname],
```
nitzmahone: Why the two dicts? Seems like some stuff is duplicated between cloudforms and cloudforms_params- would it be more user-friendly to squash to a single "cloudforms_facts" dict or something?

nitzmahone:
Sweet- just the one question about the two sub-dicts, otherwise LGTM!

mrjoshuap:
There is no technical reason they can't be squashed into a single dict; this was based on the foreman.py inventory script that shipped with Ansible Tower 3.0.1, which used the second for host group parameters.  Let me rework it with the suggestion and update this PR.

mrjoshuap:
Reworked to squash into a single dict.  Additionally:
- removed second cache / dict with duplicate info
- added purge_actions configuration option to remove the actions from a host (defaults to False)
- added prefer_ip_address configuration option so give the option of using ip address instead of name (defaults to True)
- removed self variables — just use the arguments directly
- added --pretty command line option to pretty print results
- renamed _resolve_params to _resolve_host

New output attached.

[cloudforms--host.json.txt](https://github.com/ansible/ansible/files/415494/cloudforms--host.json.txt)
[cloudforms--list.json.txt](https://github.com/ansible/ansible/files/415495/cloudforms--list.json.txt)

mrjoshuap:
Also, --help updated:

```
usage: cloudforms.py [-h] [--list] [--host HOST] [--pretty] [--refresh-cache]
                     [--debug]

Produce an Ansible Inventory file based on CloudForms managed VMs

optional arguments:
  -h, --help       show this help message and exit
  --list           List instances (default: True)
  --host HOST      Get all the variables about a specific instance
  --pretty         Pretty print JSON output (default: False)
  --refresh-cache  Force refresh of cache by making API requests to CloudForms
                   (default: False - use cache files)
  --debug          Show debug output while running (default: False)
```

contrib/inventory/cloudforms.py
```diff
@@ -0,0 +1,351 @@
...
+            if host.has_key('tags'):
+                for group in host['tags']:
+                    safe_key = self.to_safe(group['name'])
+                    self.push(self.inventory, safe_key, dns_name)
+
+                    if self.args.debug:
+                        print "Found tag [%s] for host which will be mapped to [%s]" % (group['name'], safe_key)
+
+            full_host = self._resolve_host(host)
+            if host.has_key('ipaddresses'):
```
jlaska: Can you use `if 'ipaddresses' in host` instead?  We should use this where ever you have `has_key()`

contrib/inventory/cloudforms.py
```diff
@@ -0,0 +1,351 @@
...
+            if self.cloudforms_prefer_ip_address and host.has_key('ipaddresses') and host['ipaddresses'][0]:
+                dns_name = host['ipaddresses'][0]
+            else:
+                dns_name = host['name']
+
+            if self.args.debug:
+                print "Host named [%s] will use [%s] as dns_name" % (host['name'], dns_name)
+
+            # Create ansible groups for hostgroup, location and organization
+            if host.has_key('tags'):
```
jlaska: Suggest changing to `if 'tags' in host:`

contrib/inventory/cloudforms.py
mrjoshuap: I'll rework both of them as suggested and catch any others.

contrib/inventory/cloudforms.py
```diff
@@ -0,0 +1,351 @@
...
+        if self.args.debug:
+            print "Updating cache..."
+
+        for host in self._get_hosts():
+            if host['power_state'] != 'on':
+                if self.args.debug:
+                    print "Skipping %s because power_state = %s" % (host['name'], host['power_state'])
+                continue
+
+            if self.cloudforms_prefer_ip_address and host.has_key('ipaddresses') and host['ipaddresses'][0]:
```
jlaska: Instead of changing the `name`, can we set `ansible_ssh_host` to the discovered `ipaddress`?

jlaska: Also, this results in a traceback for me.  I think you want to do ...

``` python
if 'ipaddresses' in host and isinstance(host['ipaddresses'], list) and host['ipaddresses']:
```

mrjoshuap:
I'm addressing all of those concerns and adapting it to catch more exceptions.

mrjoshuap:
All issues and suggestions addressed, needs reviewed and tested by others :-)

```
usage: cloudforms.py [-h] [--list] [--host HOST] [--pretty] [--refresh-cache]
                     [--debug]

Produce an Ansible Inventory file based on CloudForms managed VMs

optional arguments:
  -h, --help       show this help message and exit
  --list           List instances (default: True)
  --host HOST      Get all the variables about a specific instance
  --pretty         Pretty print JSON output (default: False)
  --refresh-cache  Force refresh of cache by making API requests to CloudForms
                   (default: False - use cache files)
  --debug          Show debug output while running (default: False)
```

Also updated outputs:

[cloudforms_py--host.json.txt](https://github.com/ansible/ansible/files/423716/cloudforms_py--host.json.txt)
[cloudforms_py--list.json.txt](https://github.com/ansible/ansible/files/423717/cloudforms_py--list.json.txt)

contrib/inventory/cloudforms.py
mrjoshuap: done 👍

contrib/inventory/cloudforms.py
mrjoshuap: this is actually not needed since it will keep the name and specify `ansible_ssh_host` as the first IP.

mrjoshuap:
Sidebar:

Can anyone confirm if group names need to be sanitized?  I have checked and searched but not found anything definitive.  For example, one of the tags returned is named **/managed/folder_path_yellow/datacenters** and a type is returned as **ManageIQ::Providers::Redhat::InfraManager::Vm** ... I believe slashes and colons will prevent Ansible from using them as groups or worse?

I added a configuration option to make it optional and defaulted it to True because I wasn't sure about the character validity in Ansible...

They are now added (as groups) as:
- _managed_folder_path_yellow_datacenters
- ManageIQ__Providers__Redhat__InfraManager__Vm

I could argue that at least for tags that I might want to make group -> subgroup split on the slashes... others are not as straight forward.

mrjoshuap:
It is not a good idea to split tags on the slashes (/) ... the last sub-group key can be duplicated in multiple parents squashing all hosts into a single group and wreaking havoc upon unsuspecting innocents.

mrjoshuap:
I don't have enough data to confirm my previous comment re: splitting tags on /... I created a new branch where this could be tested but not act as a blocker for this PR.

https://github.com/mrjoshuap/ansible/tree/cloudforms_inventory_nested_tags

jlaska:
@mrjoshuap in your sample JSON output, you incude the `actions` key for each host.  Not a show stopper, but I wonder if that data is useful to ansible?  Should it be excluded in the hostvars?

contrib/inventory/cloudforms.py
```diff
@@ -0,0 +1,400 @@
...
+        ret = requests.get(url,
+                           auth=HTTPBasicAuth(self.cloudforms_username, self.cloudforms_pw),
+                           verify=self.cloudforms_ssl_verify)
+
+        ret.raise_for_status()
+
+        try:
+            results = json.loads(ret.text)
+        except ValueError:
+            warnings.warn("Unexpected response from {0} ({1}): {2}".format(self.cloudforms_url, r.status_code, r.reason))
```
jlaska: heads up, `r` is undefined.

jlaska:
Nit pick, but running this through `flake8` will highlight a few other minor gotchas.

contrib/inventory/cloudforms.py
mrjoshuap: Whoops!  silly me.  On that now.

mrjoshuap:
I agree with regards to actions; I made it configurable and default it to True; should I default it to False?  Also, I will utilize the flake8 and see what it says.

contrib/inventory/cloudforms.py
jlaska: Oh, I see you've made the `actions` configurable.  Disregard that comment.

mrjoshuap:
flake8 returns all clear with exception of long lines, some of which I'm not sure the best way to resolve.  Are we extremely picky on that front?

mrjoshuap:
For some reason, I removed the purge_actions method and forgot to add it back in.  I fixed that.  Also, added a nest_tags option so that CloudForms tags would show up as a nested list.  I split them based on the / in the tag name.  It is disabled by default because my sample size is vary small and there might be problems when a node with the same name appears in multiple tags.

mrjoshuap:
Here's some sample outputs....

[cloudforms_py--host.json.txt](https://github.com/ansible/ansible/files/425403/cloudforms_py--host.json.txt)
[cloudforms_py--list--nested-tags.json.txt](https://github.com/ansible/ansible/files/425404/cloudforms_py--list--nested-tags.json.txt)
[cloudforms_py--list.json.txt](https://github.com/ansible/ansible/files/425405/cloudforms_py--list.json.txt)

jlaska:
> flake8 returns all clear with exception of long lines, some of which I'm not sure the best way to resolve. Are we extremely picky on that front?

I'm not, I don't believe @nitzmahone is either.

contrib/inventory/cloudforms.py
```diff
@@ -0,0 +1,460 @@
...
+                    self.inventory['tags'] = dict(children=[], vars={}, hosts=[])
+
+                if not self.cloudforms_nest_tags:
+                    # don't expand tags, just use them in a safe way
+                    for group in host['tags']:
+                        # Add sub-group, as a child of top-level
+                        safe_key = self.to_safe(group['name'])
+                        if safe_key:
+                            if self.args.debug:
+                                print "Adding sub-group '%s' to parent 'tags'" % safe_tag_name
```
jlaska: Heads  up, I think `safe_tag_name` is undefined here

mrjoshuap: you are correct sir, my bad, I missed the output from flake8!  /sigh

nitzmahone:
Nope- we're trying to tighten up pep/flake kinds of things in the core codebase going forward, but this being contrib, we probably don't need to be too militant. :)

jlaska:
@nitzmahone looks good to me, thanks @mrjoshuap for all the work!

PR #17039, opened 2016-08-10
CloudRetry/AWSRetry backoff decorator with unit tests

linuxdynasty:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

CloudRetry Base class inside of module_utils/cloud.py
AWSRetry decorator inside of module_utils/ec2.py
##### ANSIBLE VERSION

```
ansible 2.1.1.0
```
##### SUMMARY

The CloudRetry class can be implemented by any other cloud provider, that wants to implement a basic backoff algorithm in a decorator.

AWSRetry class overwrites 2 methods.
- status_code_from_exception (Parses the exception for the exact string)
- found (Iterates over a list of exceptions to match against.

AWSRetry.backoff() decorator can be applied to any function that is making an aws boto/boto3 call.
It will only retry on the following Exceptions.
This list of failures is based on this [API Reference](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html).
- RequestLimitExceeded
- Unavailable
- ServiceUnavailable
- InternalFailure
- InternalError
- "\w+.NotFound" (Eventual Consistency) 

The CloudRetry.backoff decorator takes on the following kwargs.
- tries (number of times to try) default=10
- delay (initial delay between retries in seconds) default=3
- backoff (This is the multiplier, that will double on each retry) default=2

---

This change is [Reviewable](https://reviewable.io/reviews/ansible/ansible/17039)

### DIFF
--- a/lib/ansible/module_utils/cloud.py
+++ b/lib/ansible/module_utils/cloud.py
@@ -1,0 +1,105 @@
+#
+# (c) 2016 Allen Sanabria, <asanabria@linuxdynasty.org>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+This module adds shared support for generic cloud modules
+
+In order to use this module, include it as part of a custom
+module as shown below.
+
+from ansible.module_utils.cloud import *
+
+The 'cloud' module provides the following common argument specs:
+
+    * CloudRetry Class
+        - The base class to be used by other cloud providers, in order to
+          provide a backoff/retry decorator based on status codes.
+
+        - Example using the AWSRetry class which inherits from CloudRetry.
+          @AWSRetry.retry(tries=20, delay=2, backoff=2)
+          get_ec2_security_group_ids_from_names()
+
+"""
+from functools import wraps
+import syslog
+import time
+
+
+class CloudRetry(object):
+    """ CloudRetry can be used by any cloud provider, in order to implement a
+        backoff algorithm/retry effect based on Status Code from Exceptions.
+    """
+    # This is the base class of the exception.
+    # AWS Example botocore.exceptions.ClientError
+    base_class = None
+
+    @staticmethod
+    def status_code_from_exception(error):
+        """ Return the status code from the exception object
+        Args:
+            error (object): The exception itself.
+        """
+        pass
+
+    @staticmethod
+    def found(response_code):
+        """ Return True if the Response Code to retry on was found.
+        Args:
+            response_code (str): This is the Response Code that is being matched against.
+        """
+        pass
+
+    @classmethod
+    def retry(cls, tries=10, delay=3, backoff=2):
+        """ Retry calling the Cloud decorated function using an exponential backoff.
+        Kwargs:
+            tries (int): Number of times to try (not retry) before giving up
+                default=10
+            delay (int): Initial delay between retries in seconds
+                default=3
+            backoff (int): backoff multiplier e.g. value of 2 will double the delay each retry
+                default=2
+
+        """
+        def deco(f):
+            @wraps(f)
+            def retry(*args, **kwargs):
+                max_tries, max_delay = tries, delay
+                while max_tries > 1:
+                    try:
+                        return f(*args, **kwargs)
+                    except Exception as e:
+                        if isinstance(e, cls.base_class):
+                            response_code = cls.status_code_from_exception(e)
+                            if cls.found(response_code):
+                                msg = "{0}: Retrying in {1} seconds...".format(str(e), max_delay)
+                                syslog.syslog(syslog.LOG_INFO, msg)
+                                time.sleep(max_delay)
+                                max_tries -= 1
+                                max_delay *= backoff
+                            else:
+                                # Return original exception if exception is not a ClientError
+                                raise e
+                        else:
+                            # Return original exception if exception is not a ClientError
+                            raise e
+                return f(*args, **kwargs)
+
+            return retry  # true decorator
+
+        return deco
--- a/lib/ansible/module_utils/cloud.py
+++ b/lib/ansible/module_utils/cloud.py
@@ -39,7 +39,7 @@ from functools import wraps
 import syslog
 import time
 
-from pycompat24 import get_exception
+from ansible.module_utils.pycompat24 import get_exception
 
 
 class CloudRetry(object):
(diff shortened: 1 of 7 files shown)


bcoca:
so instead of re implementing from scratch, would it not be better to expand the existing retry functions to allow for the custom conditions?

https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/api.py

linuxdynasty:
@bcoca I read api.py before writing my own. This one is purely focused on AWS and not just on the RateLimiting issue. This one also gets around the NotFound" (Eventual Consistency) issue as well. Could we add an argument to the decorator in API to take in Exception Strings. Sure, but I think this way the writer of the modules will not have to think of the multiple use cases beyond RequestLimitExceeded. Also if an exception is not part of the list of retries, it will just raise it.

bcoca:
@linuxdynasty i'm not saying modules should wrap the existing retries, i was suggesting this function could be re-implemented as a simple wrapper to an expanded 'generic retry'. 

This would make 'cloud specific' retry functions much easier in general.

linuxdynasty:
I refactored it @bcoca. Now there is a base class called CloudRetry in module_utils/cloud.py. AWSRetry class in module_utils/ec2.py  inherits from CloudRetry. Now only two staticmethods need to be overridden when another cloud provider wants to use this decorator. Test also updated.

rodrickbrown:
+1 this looks good well needed feature request thanks @linuxdynasty

linuxdynasty:
Can we add botocore, boto3 to the ansible core testing framework. As one of the failures, is that botocore does not exist.

mattclay:
@linuxdynasty You can add `botocore` and `boto3` to `test/utils/shippable/sanity-requirements.txt`.

lib/ansible/module_utils/cloud.py
```diff
@@ -0,0 +1,105 @@
...
+
+        """
+        def deco(f):
+            @wraps(f)
+            def retry_func(*args, **kwargs):
+                max_tries, max_delay = tries, delay
+                while max_tries > 1:
+                    try:
+                        return f(*args, **kwargs)
+                    except Exception as e:
```
mattclay: This isn't compatible with Python 2.4, which is causing it to fail the py24 tests:

```
2016-08-11 20:25:51 + python2.4 -m compileall -fq -x 'module_utils/(a10|rax|openstack|ec2|gce|docker_common|azure_rm_common|vca|vmware|gcp|gcdns).py' lib/ansible/module_utils
2016-08-11 20:25:51 Compiling lib/ansible/module_utils/cloud.py ...
2016-08-11 20:25:51   File "lib/ansible/module_utils/cloud.py", line 86
2016-08-11 20:25:51     except Exception as e:
2016-08-11 20:25:51                       ^
2016-08-11 20:25:51 SyntaxError: invalid syntax
```

You can use this instead:

``` python
except Exception:
    e = get_exception()
```

mattclay:
@linuxdynasty My apologies, I had you put the dependencies in the wrong requirements file. You should add them to the two requirements files in the `test/utils/tox` directory instead.

mattclay:
@linuxdynasty If you want to run the unit tests locally with an environment similar to Shippable, use the following command: `TOXENV=py27 test/utils/shippable/sanity.sh`

That will run the unit tests on Python 2.7. You can use other Python versions as well, if you have them installed.

test/utils/shippable/sanity-requirements.txt
```diff
@@ -2,3 +2,5 @@ tox
 pyyaml
 jinja2
 setuptools
+botocore
+boto3
```
mattclay: Changes to this file aren't needed since the dependencies are in the `test/utils/tox` requirements files.

lib/ansible/module_utils/cloud.py
```diff
@@ -0,0 +1,108 @@
...
+
+"""
+from functools import wraps
+import syslog
+import time
+
+from ansible.module_utils.pycompat24 import get_exception
+
+
+class CloudRetry(object):
```
bcoca, line 45: not exactly what i was asking, but good enough

linuxdynasty:
@mattclay thank you for all your help. Test are finally passing :D

linuxdynasty:
tests are failing, but not do to this PR.

mattclay:
I've restarted the build. The failure was due to a temporary timeout.

test/units/module_utils/ec2/test_aws.py
```diff
@@ -0,0 +1,87 @@
...
+        self.assertEqual(self.counter, 1)
+
+    def test_retry_once(self):
+        self.counter = 0
+        err_msg = {'Error': {'Code': 'InstanceId.NotFound'}}
+
+        @AWSRetry.backoff(tries=2, delay=0.1)
+        def retry_once():
+            self.counter += 1
+            if self.counter < 2:
```
fearphage, line 44: These tests would be a lot cleaner [using `Mock`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.side_effect).

lib/ansible/module_utils/cloud.py
senorsmile, line 45: I do think this could be useful for any module (not just cloud modules).

test/units/module_utils/ec2/test_aws.py
bcoca, line 44: Which is why the original 'retry' is in `api.py`, I might migrate this after merge

linuxdynasty:
@bcoca any updates on if you want me to move CloudRetry from cloud.py to api.py and still leave AWSRetry in ec2.py?

lib/ansible/module_utils/cloud.py
```diff
@@ -0,0 +1,108 @@
...
+    @staticmethod
+    def found(response_code):
+        """ Return True if the Response Code to retry on was found.
+        Args:
+            response_code (str): This is the Response Code that is being matched against.
+        """
+        pass
+
+    @classmethod
+    def backoff(cls, tries=10, delay=3, backoff=2):
```
ryansb: I'm not sure about these defaults - a delay of 3 and backoff of 2 for 10 tries would mean that, to fail, this retry decorator would wait for 3069 seconds (`3 + 3*2 + 3*2*2 ....`, or `sum([3 * 2**i for i in range(10)])`) or about 50 minutes. That seems like a really long time, especially since most modules make several calls. 

A better default might be 4 tries, for a total default wait time of 45 seconds and having a max of, say, a minute between tries. That way, if someone wanted 10 tries it would only take about 7.5 minutes to fail.

linuxdynasty: Makes sense and I will update it shortly.

linuxdynasty: I'm thinking of changing the  default backoff of 2 seconds to 1.1 seconds. I think this is more reasonable. `sum([3 * 1.1**i for i in range(10)])`
About a total of 48 seconds in 10 tries. @ryansb thoughts?

jjshoe: This should be configurable, because this isn't enough for the work I do with auto scale groups.

ryansb: @jjshoe It is configurable, these values are just the defaults that will be used if the developer calls the decorator without arguments. 

@linuxdynasty That seems reasonable to me - go for it. Should be plenty to wait out most rate limits.

linuxdynasty:
@ryansb I pushed the changes. The build failed cause it could not spin a container.

linuxdynasty:
@ryansb does everything else look good to you?

Issue #17059, opened 2016-08-12
retries documentation and behaviour for 1.9 is inconsistent


electrofelix:
##### ISSUE TYPE
- Bug Report
##### COMPONENT NAME

runner
##### ANSIBLE VERSION

```
ansible 1.9.6
  configured module search path = None
```
##### CONFIGURATION

none
##### OS / ENVIRONMENT

N/A
##### SUMMARY

Current documentation for `retries` under playbook_loops contains the following text:

> The task returns the results returned by the last task run. The results of individual retries can be viewed by -vv option. The registered variable will also have a new key "attempts" which will have the number of the retries for the task.

This suggests that you would see the same message appearing on failure from the last attempt as you would if retries was not applied to the task. Unfortunately the code replaces the contents of 'msg' in the results returned, thus hiding any information from many executed module as to what the error was that caused the retry to fail. Some modules such as the `shell` and `command` modules will include `stdout` separately, however as the common module boilerplate https://github.com/ansible/ansible/blob/stable-1.9/docsite/rst/developing_modules.rst#id5 encourages many modules to use `msg` for any messages back to the user when calling `exit_json`, those that include useful information via other attributes upon error with retries/until to be displayed are the exception rather than the rule.

While it is possible to print the result from each attempt using '-vv', it's not practical to enable this for long playbooks.
##### STEPS TO REPRODUCE

As the command module uses msg for errors for cases where it detects a problem with the arguments given, it provides a simple example

```
- hosts: localhost
  connection: local
  gather_facts: False
  tasks:
    - command:
      register: _test_retries_shell_result
      until: _test_retries_shell_result.rc == 0
      retries: 3
      delay: 
```

Alternatively, can see better with the pip module

```
- hosts: localhost
  connection: local
  gather_facts: False
  tasks:
    - pip:
        name: pip wheel virtualenv
        virtualenv: ~/tmp/test-virtualenv
        state: present
        extra_args: >
          --upgrade
          --timeout 30
          --invalid-arg
      register: _test_retries_virtualenv_result
      until: _test_retries_virtualenv_result|success
      retries: 3
      delay: 1
```
##### EXPECTED RESULTS

Expected to see the same output from the last iteration of the retry task as would normally appear if retries was not applied to the task, and the additional message about failing due to exceeding the specified number of retries.

command example

```
PLAY [localhost] ************************************************************** 

TASK: [command ] ************************************************************** 
[...]

PR #17060, opened 2016-08-12
Include last message from module on retry failure

electrofelix:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

runner
##### ANSIBLE VERSION

```
ansible 1.9.6
```
##### SUMMARY

Preserve the last "msg" from the module executed when failure results
in the retry limit exceeded. This fixes the code to work as documented
where it is currently stated 'The task returns the results returned by
the last task run', but in reality this is incorrect as 'msg' is
overridden, thus hiding part of the result from the last run.

When using `retry` and `until`, upon max attempts being reached,
ansible will replace the 'msg' value returned from the executed module
with a simple message of:

  Task failed as maximum retries was encountered

This hides any messages returned from the module leaving the user non
the wiser as to the underlying cause. While it may be possible to
increase the verbosity to see each execution, for larger playbooks this
causes so much info to be printed, that it becomes problematic viewing
results from such output.

Inline with expectations based on the documentation, this change
preserves the last message returned by the module to be included in
what is displayed to the user using 'last_msg'.

It uses 'last_msg' to avoid unintentional regressions where users are expecting to only see the message:

> msg: Task failed as maximum retries was encountered

Fixes: #17059

Example playbook

```
- hosts: localhost
  connection: local
  gather_facts: False
  tasks:
    - pip:
        name: pip wheel virtualenv
        virtualenv: ~/tmp/test-virtualenv
        state: present
        extra_args: >
          --upgrade
          --timeout 30
          --invalid-arg
      register: _test_retries_virtualenv_result
      until: _test_retries_virtualenv_result|success
      retries: 3
      delay: 1
```

Before:

```

TASK: [pip ] ****************************************************************** 
failed: [localhost] => {"attempts": 3, "cmd": "/home/<user>/tmp/test-virtualenv/bin/pip install --upgrade 
--timeout 30 --invalid-arg\n pip wheel virtualenv", "failed": true}
msg: Task failed as maximum retries was encountered

pip  -------------------------------------------------------------------- 4.21s
-------------------------------------------------------------------------------
Total: ------------------------------------------------------------------ 4.24s
           to retry, use: --limit @/home/<user>/test-virtualenv.retry

```

After:

```

TASK: [pip ] ****************************************************************** 
failed: [localhost] => {"attempts": 3, "cmd": "/home/<user>/tmp/test-virtualenv/bin/pip install --upgrade 
--timeout 30 --invalid-arg\n pip wheel virtualenv", "failed": true, "last_msg": "\n:stderr: \nUsage:   \n  
pip install [options] <requirement specifier> [package-index-options] ...\n  pip install [options] -r 
<requirements file> [package-index-options] ...\n  pip install [options] [-e] <vcs project url> ...\n  pip 
install [options] [-e] <local project path> ...\n  pip install [options] <archive url/path> ...\n\nno such 
option: --invalid-arg\n"}
msg: Task failed as maximum retries was encountered

pip  -------------------------------------------------------------------- 4.22s
-------------------------------------------------------------------------------
Total: ------------------------------------------------------------------ 4.28s
           to retry, use: --limit @/home/<user>/test-virtualenv.retry

```

### DIFF
--- a/docsite/rst/playbooks_loops.rst
+++ b/docsite/rst/playbooks_loops.rst
@@ -238,7 +238,8 @@ Sometimes you would want to retry a task until a certain condition is met.  Here
 The above example run the shell module recursively till the module's result has "all systems go" in its stdout or the task has
 been retried for 5 times with a delay of 10 seconds. The default value for "retries" is 3 and "delay" is 5.
 
-The task returns the results returned by the last task run. The results of individual retries can be viewed by -vv option.
+The task returns the results returned by the last task run, including the last 'msg' saved as 'last_msg'.
+The results of individual retries can be viewed by -vv option.
 The registered variable will also have a new key "attempts" which will have the number of the retries for the task.
 
 .. _with_first_found:
--- a/lib/ansible/runner/__init__.py
+++ b/lib/ansible/runner/__init__.py
@@ -1063,6 +1063,7 @@ class Runner(object):
                     if utils.check_conditional(cond, self.basedir, inject, fail_on_undefined=self.error_on_undefined_vars):
                         break
                 if result.result['attempts'] == retries and not utils.check_conditional(cond, self.basedir, inject, fail_on_undefined=self.error_on_undefined_vars):
+                    result.result['last_msg'] = result.result['msg']
                     result.result['failed'] = True
                     result.result['msg'] = "Task failed as maximum retries was encountered"
             else:


bcoca:
This was not a bug in 1.9, just how it worked. 

We  do not normally backport features and it is very unlikely we release another 1.9 at this point, so I'm going to close this issue.

electrofelix:
The documentation for 1.9 says the following:

> The task returns the results returned by the last task run. The results of individual retries can be viewed by -vv option. The registered variable will also have a new key "attempts" which will have the number of the retries for the task.

Which sounds like, that's not the way it's supposed to work for 1.9, since if it exceeds the number of attempts it will **NOT** contain the results from the last task run.

The existing behaviour makes "retries" almost impossible to work with for 1.9, unless you're prepared for large amount of console spamming in order to see what broke.

PR #17067, opened 2016-08-12
Revert to the default behavior of combining/returning group/hostvars

angstwad:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

`ansible.inventory.Inventory` group and host var combining
##### ANSIBLE VERSION

```
ansible 2.2.0 (var_combine_fix 50b789b950) last updated 2016/08/12 11:26:36 (GMT -500)
  lib/ansible/modules/core: (detached HEAD a216ef210b) last updated 2016/08/11 19:17:25 (GMT -500)
  lib/ansible/modules/extras: (detached HEAD 39153ea154) last updated 2016/08/11 19:17:53 (GMT -500)
  config file =
```
##### SUMMARY
#16779 changes the internal API to prevent host and group vars being combined and returned by the `Inventory` class.   This greatly affects `vars_plugins` functionality by not setting combined host and group vars on their respective `Host` and `Group` objects.   Because of this, vars plugins no longer have access to these vars and can no longer change them from the context of a plugin.  Additionally, the inventory object passed to the vars plugin has itself not been changed due to order of operations, there is no access to current vars, compounding the problem presented by this #16779.

This new behavior has broken some of my team's prod vars plugins. I feel that this kind of internal API change perhaps is better suited for a major (or even minor) release.  

This PR reverts the default behavior to its previous default of combining vars, with the option of changing the behavior to prevent the "expensive" combine operation via configuration.

Please consider cherry-picking this change to the 2.x stable release to remediate the changed behavior.

### DIFF
--- a/lib/ansible/constants.py
+++ b/lib/ansible/constants.py
@@ -185,6 +185,8 @@ DEFAULT_LOG_PATH          = get_config(p, DEFAULTS, 'log_path',           'ANSIB
 DEFAULT_FORCE_HANDLERS    = get_config(p, DEFAULTS, 'force_handlers', 'ANSIBLE_FORCE_HANDLERS', False, boolean=True)
 DEFAULT_INVENTORY_IGNORE  = get_config(p, DEFAULTS, 'inventory_ignore_extensions', 'ANSIBLE_INVENTORY_IGNORE', ["~", ".orig", ".bak", ".ini", ".cfg", ".retry", ".pyc", ".pyo"], islist=True)
 DEFAULT_VAR_COMPRESSION_LEVEL = get_config(p, DEFAULTS, 'var_compression_level', 'ANSIBLE_VAR_COMPRESSION_LEVEL', 0, integer=True)
+DEFAULT_GROUP_VAR_COMBINE = get_config(p, DEFAULTS, 'group_var_combine', 'ANSIBLE_GROUP_VAR_COMBINE', True, boolean=True)
+DEFAULT_HOST_VAR_COMBINE  = get_config(p, DEFAULTS, 'host_var_combine', 'ANSIBLE_HOST_VAR_COMBINE', True, boolean=True)
 
 # static includes
 DEFAULT_TASK_INCLUDES_STATIC    = get_config(p, DEFAULTS, 'task_includes_static', 'ANSIBLE_TASK_INCLUDES_STATIC', False, boolean=True)
--- a/lib/ansible/inventory/__init__.py
+++ b/lib/ansible/inventory/__init__.py
@@ -740,11 +740,11 @@ class Inventory(object):
             self._vars_per_host = {}
             self._vars_per_group = {}
 
-    def get_host_vars(self, host, new_pb_basedir=False, return_results=False):
+    def get_host_vars(self, host, new_pb_basedir=False, return_results=C.DEFAULT_HOST_VAR_COMBINE):
         """ Read host_vars/ files """
         return self._get_hostgroup_vars(host=host, group=None, new_pb_basedir=new_pb_basedir, return_results=return_results)
 
-    def get_group_vars(self, group, new_pb_basedir=False, return_results=False):
+    def get_group_vars(self, group, new_pb_basedir=False, return_results=C.DEFAULT_GROUP_VAR_COMBINE):
         """ Read group_vars/ files """
         return self._get_hostgroup_vars(host=None, group=group, new_pb_basedir=new_pb_basedir, return_results=return_results)
 
@@ -768,7 +768,7 @@ class Inventory(object):
             found_vars = set(os.listdir(to_unicode(path)))
         return found_vars
 
-    def _get_hostgroup_vars(self, host=None, group=None, new_pb_basedir=False, return_results=False):
+    def _get_hostgroup_vars(self, host=None, group=None, new_pb_basedir=False, return_results=C.DEFAULT_HOST_VAR_COMBINE):
         """
         Loads variables from group_vars/<groupname> and host_vars/<hostname> in directories parallel
         to the inventory base directory or in the same directory as the playbook.  Variables in the playbook


jimi-c:
@angstwad we did follow that up with an option to re-enable the old behavior.

angstwad:
@jimi-c Is that in another PR?  Looks like what's in the code today has no option to re-enable the old behavior.  When I searched I didn't find anything.

jimi-c:
@angstwad it's mainly for accessing the API directly. Commit 7dc09adaf4d07444c09229e58bddac762f2e818e

angstwad:
@jimi-c That's exactly the commit I'm referring to - this change is in the API, and changes fundamental behavior WRT how vars are combined.  This directly impacts vars plugins.  The change implemented in this PR allows this functionality to be tweaked at runtime.  

Personally, I don't see the benefit of this change at all since the "expensive" operation is run once and cached.  But since you guys liked it, that's fine, but let's make it configurable.  Otherwise, vars plugins are broken for us (and presumably others).

jimi-c:
@angstwad it did not seem to be run just once, and with a large number of hosts + vars, it was a significant performance hit. This is all going to be changing in the future, inventory is a mess and it should not be dealing with vars at all. The proper way to do this via the API should be `VariableManager.get_vars(host=some_host)`. Adding the above flag in was a stop-gap solution until we rewrite this. I'm unclear as to why this impacts vars plugins, as that code doesn't appear to be changed by this at all (https://github.com/ansible/ansible/blob/devel/lib/ansible/inventory/__init__.py#L552). Your patch doesn't add any code to that section.

angstwad:
@jimi-c 
See https://github.com/ansible/ansible/blob/devel/lib/ansible/inventory/__init__.py#L567, in the function you reference.

WRT how this impacts vars plugins, since this operation now does not return a value, these variables are not returned out of the call stack (several functions deep).  The call stack looks like this:

[`parse_inventory:L159`](https://github.com/ansible/ansible/blob/devel/lib/ansible/inventory/__init__.py#L159) 
[`get_group_variables:L549`](https://github.com/ansible/ansible/blob/devel/lib/ansible/inventory/__init__.py#L549)
[`_get_group_variables:L567`](https://github.com/ansible/ansible/blob/devel/lib/ansible/inventory/__init__.py#L567)
[`get_group_vars:L755`](https://github.com/ansible/ansible/blob/devel/lib/ansible/inventory/__init__.py#L755)
[`_get_hostgroup_vars:L823-827`](https://github.com/ansible/ansible/blob/devel/lib/ansible/inventory/__init__.py#L823-L827)

At [`parse_inventory:L159`](https://github.com/ansible/ansible/blob/devel/lib/ansible/inventory/__init__.py#L159), you'll see that if the results of the call stack aren't returned all the way back, then group vars are not set on the group object, and are lost when vars plugins run for hosts, which will be called down in the call stack that starts in the next for loop, [`parse_inventory:L164`](https://github.com/ansible/ansible/blob/devel/lib/ansible/inventory/__init__.py#L164) .

I won't deny that this is a complicated bit of code going on here and I can see why you'd want to refactor a bit here.  But that said, debugging this down to this point in the code is where I've traced the issue.  

Haven't tried the VariableManager trick yet, but I was using pydev and debugging the plugin, and didn't see any object with the state I needed.  I'll see what comes of another test.

omgjlk:
Hey, could we get some attention here? Another release is going to come out and so far there hasn't been a solution for fixing our vars plugin.

jimi-c:
@j2sol I'm still unclear why this change in inventory would be affecting a method defined in a vars plugin. Those methods should be returning results themselves shouldn't they? Unless it's calling code from the Inventory object, which seems wrong to me. Can you share the vars plugin code?

sivel:
Just to add to this, I was contacted recently about this same problem (external to this report) at work, where the vars plugin needed to use inventory data to perform work.

The info I was given was:

> the inventory/host information that is being passed to my vars plugin is missing all of the host vars from the dynamic inventory, which is kind of perplexing.  the inventory hosts are being passed in correctly, but only name, short name, group membership, etc.  Any ideas?

As a result, we have an internal team as well that is now maintaining a fork of ansible for their use with this patch.

omgjlk:
@jimi-c I can provide a PR where I changed our vars plugin code to act differently and "resolve" the issue (but it's ugly).

Our plugin would first need to get a current read of all the variable data that exists for hosts/groups. It would then read through a list of "default" variables for a site and add in any variable that wasn't already defined at the host/group level. The entire structure is then returned. In this way we can have some site "defaults" that can be overridden at the group or host level.

https://github.com/blueboxgroup/ursula/pull/2230/files

bcoca:
FYI, i was planning on killing vars_plugins in their current form, can I get use cases?

We haven't really gotten any contributions to that plugin type.

omgjlk:
@bcoca See my last comment. We use one Ansible playbook set with a lot of different inventories. We have some variable data that we like to express as a "default" for all the inventories, and then allow overriding those values in the inventory itself (via group or host vars). These defaults can be or often are different from the role defaults within the playbooks.

Without a vars plugin, we'd have to implement this as a dynamic inventory plugin, that re-uses as much of Ansible code as possible to parse the inventory files, and then we would inject our variable data as needed. (in fact, I have a work in progress that does just that: https://github.com/blueboxgroup/ursula-cli/pull/54/files#diff-e0f80483921e90e63127db61d3ccd68dR1 )

angstwad:
@bcoca delete vars_plugins completely?  That sounds like a bad idea. I can appreciate that you don't see community contributions, but how are you measuring private usage of vars_plugins?  There are undoubtedly many teams relying on this functionality, with two use cases described right here.

bcoca:
that is why I'm asking, i have no way to evaluate usage that i don't know about.

We have been wanting to redo inventory since 2.0, but have been pushing it back, we want to make 'true plugins' instead of hardcoded ones. We also want to integrate caching and other features and transition from inventory scripts (script plugin will stay for backwards compatibility). 

In that design, current vars_plugins seemed not have a place, I was thinking of either eliminating them or changing them to 'vars format plugins' for dataloader to be able to read ini, yaml, json, xml, etc other data formats.

Do your current vars_plugins make sense as inventory_plugins? 
do we still need a 'vars traversal special plugin' type?

omgjlk:
Mine can work as an inventory plugin, but it requires carrying a lot more code than I would like, code to convince Ansible to parse my inventory per-normal then allow me to inject variable data.

If the inventory objects were easier to work with, maybe this would be less of a concern.

sivel:
In our use case, we are either using a more general purpose inventory, or are using something such as the default tower inventory.  In which case, we supplement the inventory with vars plugins to fetch additional data about the servers from another source.

I have also seen vars plugins written to pull data from other cfg mgmt systems such as chef.

Although none of have been contributed directly to ansible, I think vars plugins are useful.

bcoca:
I might be misreading this, but they all sound as 'complimentary inventory scripts'

omgjlk:
Well, correct me if I'm wrong, but vars plugins are only processed at the time inventory is parsed, right? Doesn't that by nature make _any_ vars plugin essentially a 'complimentary inventory script' ?

bcoca:
Yes, which makes me lean into having them be inventory plugins instead of a
separate type.

---

Brian Coca

jimi-c:
@j2sol per our discussion at AnsibleFest, is this still an issue?

omgjlk:
@jimi-c I could use some guidance if I'm using the python API correctly to achieve my goal: https://github.com/blueboxgroup/ursula/pull/2230/files

(ignore the bad return of `group_vars`, that was fixed in a follow up)

jimi-c:
@j2sol sorry for the long delay in follow-up, but yeah that looks fine. Based on the fact that your PR above was merged, I'll go ahead and close this issue now. @angstwad if you had any further questions/concerns, please let me know via IRC or here.

Thanks!

PR #17073, opened 2016-08-12
Fix issue with proxy setting leaving temp files

swimlappy:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

module_utils/urls.py
##### ANSIBLE VERSION

```
ansible 2.1.1.0
```
##### SUMMARY

For sites that use a proxy server, often a no_proxy environment variable is used in Linux. This can be used to exclude local hosts from having to use the proxy. In our case, we have a local proxy and often connect via HTTPS to a Splunk server for Ansible playbooks. We found that Ansible was leaving a bunch of temp files around when connecting to this server. We discovered that urls.py was not removing the temp files when it discovered that the no_proxy variable was set to the current URL.

This small patch fixes that issue by removing the temp file.

### DIFF
--- a/lib/ansible/module_utils/urls.py
+++ b/lib/ansible/module_utils/urls.py
@@ -631,6 +631,13 @@ class SSLValidationHandler(urllib2.BaseHandler):
         use_proxy = self.detect_no_proxy(req.get_full_url())
 
         if not use_proxy:
+            try:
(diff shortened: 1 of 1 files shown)


alikins:
Looks good to me.

PR #17077, opened 2016-08-13
Update junos docs fragment to note the default port for transport=netconf

agaffney:
##### ISSUE TYPE

- Docs Pull Request
##### COMPONENT NAME

junos_command (and related modules)
##### ANSIBLE VERSION

```
ansible 2.1.0.0
```
##### SUMMARY

The `junos` docs snippet shows that the default value for the `port` parameter is `22`. However, this is only true with `transport=cli`, which is only valid for the `junos_netconf` module. Otherwise, the default port is `830` with `transport=netconf`.

### DIFF
--- a/lib/ansible/utils/module_docs_fragments/junos.py
+++ b/lib/ansible/utils/module_docs_fragments/junos.py
@@ -32,7 +32,7 @@ options:
     description:
       - Specifies the port to use when buiding the connection to the remote
         device.  The port value will default to the well known SSH port
-        of 22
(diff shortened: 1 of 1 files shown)


agaffney:
@gundalow

agaffney:
The `transport` parameter should probably be documented, but at the same time, there's not really a choice for the parameter in any of the modules.

gundalow:
@agaffney Thanks for taking the time to improve the documentation. 

I agree with your comments regarding `transport`.

shipit

Assigning to privateip for final review and merge

Issue #14736, opened 2016-03-02
galaxy-init: support externally defined skeletons


josephholsten:
##### Issue Type:
- Feature Idea

##### Component Name:
`ansible-galaxy init`

##### Summary:

It would be lovely to support a configurable role skeleton, similar to [`chef generate cookbook`](https://docs.chef.io/ctl_chef.html#chef-generate-cookbook) or [`puppet module generate`](https://docs.puppetlabs.com/puppet/latest/reference/modules_fundamentals.html#writing-modules)

This provides a way for different groups to automate their style guides, allowing users to scaffold things like:
- commit hooks
- static analysis and other dev tools (eg [ansible-lint](https://github.com/willthames/ansible-lint))
- testing systems (eg [test-kitchen](http://kitchen.ci/), [circleci](https://circleci.com/))
- [License](http://choosealicense.com/), [changelog](http://keepachangelog.com/), [contributing](https://github.com/blog/1184-contributing-guidelines) files
- alternatives to the built-in templates
##### Desired Behaviour:

Ideally, the `ansible-galaxy init` command would take an additional parameter to a skeleton repo:

```
ansible-galaxy init myface --skeleton https://github.com/josephholsten/ansible-skeleton
```
##### Implementation Considerations

Chef implements their generation using chef resources, so I though it might be interesting to attempt doing the same. I've converted the `ansible-galaxy init` skeleton to a role, and embedded it into https://github.com/josephholsten/ansible-generate. If this seems like a worthwhile direction to continue, I will convert my meager shell script into an ansible CLI subclass.

azenk:
I was just about to suggest something similar.  I'd also like to be able to override the metadata that is injected into the templates.  For example, much of the template data would be consistent across our org, however author will change based on who creates/maintains a role.  With your approach it seems like this would be easy to implement though.  :+1:

ansibot:
@josephholsten Greetings! Thanks for taking the time to open this issue. In order for the community to handle your issue effectively, we need a bit more information. 

Here are the items we could not find in your description:
- component name

Please set the description of this issue with this template:
https://raw.githubusercontent.com/ansible/ansible/devel/.github/ISSUE_TEMPLATE.md

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

Akasurde:
Closing this issue as PR #17079 is merged. If issue persists, please feel free to re-open.

PR #17079, opened 2016-08-13
Add role skeleton support

azenk:
##### ISSUE TYPE
- Feature Pull Request (Related feature idea #14736)
##### ANSIBLE VERSION

```
ansible 2.2.0 (role_templating acca2b5bc5) last updated 2016/08/13 12:56:44 (GMT -500)
  lib/ansible/modules/core: (detached HEAD d4eaeac8c9) last updated 2016/06/16 14:31:06 (GMT -500)
  lib/ansible/modules/extras: (detached HEAD eed9d601b5) last updated 2016/06/16 14:31:06 (GMT -500)
  config file = 
```
##### SUMMARY

The default template used by ansible-galaxy is good, but it doesn't
allow organizations to customize it to suit their needs.  This patch
addresses that by adding the option to point ansible-galaxy at a
template directory.  The contents of this directory are then
copied (or rendered) into the output role directory.  Several command
line options and configuration entries are added to allow for further
customization from the template.


azenk:
It seems like I should skip rendering for .j2 files in the templates folder.  Thoughts?

bcoca:
I'm fine with the template dir option (this was on my wishlist) but we want to limit the number of command line options and the rest seem like overkill and easily handled by the template itself.

azenk:
I'm happy to trim down the options.  We'll need some way to inject author info, etc.  How about leaving a description and adding an option/config that points to a vars.yml file that can be used for populating the remaining variables in the templates?

bcoca:
I need to think about the options, lets split it and make this just the template dir, we can deal with the rest in subsequent PRs

azenk:
I removed the command line options, and tweaked the variables that are read from the config file.  Is that along the lines of what you were thinking?  Without the ability to set template variables somehow, a lot of the functionality is lost for me, so I'm hesitant to remove the ability to set those variables entirely.  I think the approach of allowing a var file to be read in provides maximum flexibility with a minimum of configuration overhead.  The other changes that were rolled into that commit are documented in the commit message.

bcoca:
https://github.com/ansible/proposals/issues/22 <= proposal that outlined the 'configurable' skeleton option

azenk:
Except for the command line option and adding some tests and documentation, I think the existing code satisfies the skeleton functionality.  Unless there are any suggestions/objections, I think this is how I will proceed:
- [x] replace existing references to a template path with skeleton, I like that better
- [x] add a '-s/--skeleton' command line option
- [x] add tests for both the default and an external skeleton 
- [x] skip template rendering for .j2 files in the templates subfolder
- [x] use something like shutil.copyfile instead of reading/writing regular files
- [x] document, document, document
- [x] split the init_vars_file/author out into a separate PR
- [x] fix the ignore pattern matching to use the full relative path within the role for matching files/folders

chouseknecht:
@azenk - I added your new data directory structure, template expansion and file copy to #18157, which was merged. 

It should be pretty easy to rebase and fix the conflicts. Ping me when you have it, and I'll test and merge this PR.

Thanks for the great work!

azenk:
@chouseknecht - I think this is ready to go.  I know there are a couple failed tests in the current shippable run, but it looks like it got a cached/stale copy of the head of this PR.  Those two failing tests are fixed in this rebased version.  They pass locally.  I assume you can re-run?

Changes since you last saw the code:
- Light cleanup, and squashing the commits down to make rebasing easier.  
- I added a couple of empty folders with .git_keep files to the default and container skeletons.  Without these, the `files` and `templates` directories aren't created.  I considered some other options, but decided this was simplest to implement and offered the greatest flexibility.  Will setuptools play nicely with this?  
- I added a set of tests for your container enabled skeleton.  It's far from full coverage, but it should at least make sure that the basics are right.

Let me know if there's anything else that needs to change.  Thanks!

azenk:
Sorry about that, realized that I had not committed the code that I was _sure_ I had.  Now the build is failing, but seemingly through no fault of mine.

azenk:
I see that the needs revision label is still attached here.   Is that accurate?  I think all the issues have been resolved.

chouseknecht:
@azenk, just tested this.

I don't understand the need for the `.git_keep` files. If I remove them from `lib/ansible/galaxy/data/default/templates` and `lib/ansible/galaxy/data/default/files`,  the `templates` and `files` directories still get created. Testing with python 2.7.12 on OSX. Not sure if that makes a difference.

If we can remove the `.git_keep` files, then we no longer need to have `.git_keep` referenced in `role_skeleton_ignore`. That feels like a better user experience.

Also, can you modify the galaxy man pages with this change. See [docs/man/man1/ansible-galaxy.1.asciidoc.in](https://github.com/ansible/ansible/blob/devel/docs/man/man1/ansible-galaxy.1.asciidoc.in).

Other than those 2 minor things, this looks good to me.

azenk:
If they are removed, I don't think the now empty folder will be tracked by git.  My thought was using a keep file of some sort would make maintaining the default skeleton and galaxy code simpler.  I could certainly add some code to deal with a sparse default and create the missing folders, but I don't think we should force those folders to be created for user provided skeletons, just the default.  Let me know what you think.

I'll update the man pages within the next day or so.

chouseknecht:
OK. That makes sense. I didn't realize the need for .git_keep, and so it hasn't been creating `templates` and `files` since I added the `--container-enabled` option.

Looks like you need to rebase. Once you do that, I'll merge. If you don't get to the man page changes, no worries. I can add that.

azenk:
This has been rebased.  I updated my tests to reflect the removal of platform data as well.  If you don't mind taking care of the man page updates that would be awesome.  I can dig into it if needed, but at this point I probably won't have a chance for a few days.

chouseknecht:
Merging. Test failures don't seem to have anything to do with the Galaxy client.

sean-abbott:
...Does this only work in 2.3?  Because the docs don't reflect that, and I just spent an hour with 2.2.1 trying to get this work and it ignoring me.

PR #17090, opened 2016-08-15
implement module params deprecation warnings

resmo:
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

module_utils
##### ANSIBLE VERSION

```
2.3
```
##### SUMMARY

implemented a simple way to deprecate a module param by adding `deprecated_version=2.4` 

Test module
~~~python
#!/usr/bin/python

def main():
    module = AnsibleModule(dict(
        foo=dict(deprecated_version=2.4),
        )
    )
    module.exit_json(msg="ok")

if __name__ == '__main__':
    main()
~~~

output:
~~~
 $ ansible-playbook test.yml

PLAY [localhost] ******************************************************************************************************************************************************************************************************************************

TASK [foobar] *********************************************************************************************************************************************************************************************************************************
ok: [localhost]
[DEPRECATION WARNING]: Param 'foo' is deprecated. See the module docs for more information.
This feature will be removed in version 2.4. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.

PLAY RECAP ************************************************************************************************************************************************************************************************************************************
localhost                  : ok=3    changed=0    unreachable=0    failed=0   
~~~

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -681,6 +681,7 @@ class AnsibleModule(object):
         self._debug = False
         self._diff = False
         self._verbosity = 0
+        self._deprecated_param_warnings = []
         # May be used to set modifications to the environment for any
         # run_command invocation
         self.run_command_environ_update = {}
@@ -709,6 +710,9 @@ class AnsibleModule(object):
         self.no_log_values = set()
         # Use the argspec to determine which args are no_log
         for arg_name, arg_opts in self.argument_spec.items():
+            if arg_opts.get('deprecated', False) and arg_name in self.params:
+                self._deprecated_param_warnings.append("Param '%s' is deprecated. See the module docs for more information" % arg_name)
+
             if arg_opts.get('no_log', False):
                 # Find the value for the no_log'd param
                 no_log_object = self.params.get(arg_name, None)
@@ -1885,6 +1889,15 @@ class AnsibleModule(object):
         for path in self.cleanup_files:
             self.cleanup(path)
 
+    def _handle_deprecation_warnings(self, kwargs):
+        if self._deprecated_param_warnings:
+            if 'deprecation_warnings' not in kwargs:
+                kwargs['deprecation_warnings'] = list()
+            elif not isinstance(kwargs['deprecation_warnings'], list):
+                kwargs['deprecation_warnings'] = [kwargs['deprecation_warnings']]
+            kwargs['deprecation_warnings'].extend(self._deprecated_param_warnings)
+        return kwargs
+
     def exit_json(self, **kwargs):
         ''' return from the module, without error '''
         self.add_path_info(kwargs)
@@ -1892,6 +1905,7 @@ class AnsibleModule(object):
             kwargs['changed'] = False
         if 'invocation' not in kwargs:
             kwargs['invocation'] = {'module_args': self.params}
+        kwargs = self._handle_deprecation_warnings(kwargs)
         kwargs = remove_values(kwargs, self.no_log_values)
         self.do_cleanup_files()
         print('\n%s' % self.jsonify(kwargs))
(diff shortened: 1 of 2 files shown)


docsite/rst/developing_modules.rst
```diff
@@ -793,6 +793,27 @@ This example allows the stat module to be called with fileinfo, making the follo
...
+Deprecating module arguments
+````````````````````````````
+
+Starting in 2.2, you can deprecate module arguments by adding a deprecated='yes' to the arg_spec, e.g. for deprecation is_enabled::
+
+    module = AnsibleModule(
+        argument_spec = dict(
+            name=dict(required=True, aliases=['domain']),
+            force=dict(type='bool', default='no'),
+            is_enabled=dict(type='bool', default='no', deprecated='yes'),
```
bcoca: i would make this an actual boolean, not 'string' yes, this is python code, so True/False, also `default=False` should be the example for booleans.

lib/ansible/module_utils/basic.py
```diff
@@ -652,6 +653,9 @@ def __init__(self, argument_spec, bypass_checks=False, no_log=False,
         self.no_log_values = set()
         # Use the argspec to determine which args are no_log
         for arg_name, arg_opts in self.argument_spec.items():
+            if arg_opts.get('deprecated', False):
+                self._deprecated_params.append(arg_name)
```
bcoca: why not just create deprecated warnings here and append directly in exit/fail_jsons?

resmo: my intention was to keep the code as generic as possible and allow module developers to add own custom deprecation_warnings as it is just a list of msgs.

bcoca: still works if you just append

docsite/rst/developing_modules.rst
```diff
@@ -806,6 +806,26 @@ This example allows the stat module to be called with fileinfo, making the follo
...
+            name=dict(required=True, aliases=['domain']),
+            force=dict(type='bool', default=False),
+            is_enabled=dict(type='bool', default=False, deprecated=True),
+            state=dict(choices=['present', 'absent'], default='present'),
+        ),
+    )
+
+The user will see a deprecation warning::
+
+    [DEPRECATION WARNING]: Param 'is_enabled' is deprecated. See the module docs for more infomation.
```
mscherer: missing 'r' (information, not infomation)

lib/ansible/module_utils/basic.py
```diff
@@ -675,6 +676,9 @@ def __init__(self, argument_spec, bypass_checks=False, no_log=False,
         self.no_log_values = set()
         # Use the argspec to determine which args are no_log
         for arg_name, arg_opts in self.argument_spec.items():
+            if arg_opts.get('deprecated', False) and arg_name in self.params:
+                self._deprecated_param_warnings.append("Param '%s' is deprecated. See the module docs for more infomation" % arg_name)
```
mscherer: Same as in the doc

docsite/rst/developing_modules.rst
resmo: oh...fixing

bcoca:
i'm thinking (sorry) that deprecated_version implies deprecated already, simpler interface?

resmo:
It does actually imply already.  see `foo=dict(deprecated_version=2.4),` 

```
TASK [foobar] *********************************************************************************************************************************************************************************************************************************
ok: [localhost]
[DEPRECATION WARNING]: Param 'foo' is deprecated. See the module docs for more information.
This feature will be removed in version 2.4. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
```

or do you mean we should only provide `deprecation_version`?

lib/ansible/module_utils/basic.py
```diff
@@ -708,6 +709,12 @@ def __init__(self, argument_spec, bypass_checks=False, no_log=False,
         self.no_log_values = set()
         # Use the argspec to determine which args are no_log
         for arg_name, arg_opts in self.argument_spec.items():
+            if (arg_opts.get('deprecated', False) or arg_opts.get('deprecated_version')) and arg_name in self.params:
```
bcoca: here you should not need to check for deprecated if we are going to mandate deprecated_version

bcoca:
i liked your method so much i implemented similar for warnings #18517 (but had to be backwards compat)

resmo:
rebased

bcoca:
@resmo, I merged the PR that deals with warnings/deprecation keys, can you revise this one to generate the warnings based on the argument's deprecation field?

bcoca:
closed by #20884

PR #17102, opened 2016-08-16
added bits and bytes arg types

bcoca:
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

module_utils/basic
##### ANSIBLE VERSION

```
2.2
```
##### SUMMARY

2 new helper types for arg spec bits and bytes which take both int values or 'human readable' 10Mb and always reduce to bit/byte value.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -2277,12 +2277,10 @@ class AnsibleModule(object):
         elif full in number:
             result = int(number.replace(full,''))
         else:
-            i = 0
-            for suffix in suffixes:
+            for i, suffix in enumerate(suffixes):
                 if suffix in number:
                     result = int(number.replace(suffix ,'')) * (1024 ** i)
                     break
-                i += 1
 
         if result is None:
             raise ValueError("Failed to convert %s. The suffix must be one of %s or %s" % (number, full, ', '.join(suffixes)))
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -2239,9 +2239,7 @@ class AnsibleModule(object):
         fh.close()
 
     def bytes_to_human(self, size):
-        return self.pretty_bytes(size)
 
-    def pretty_bytes(self, size):
         ranges = (
                 (1<<70, 'ZB'),
                 (1<<60, 'EB'),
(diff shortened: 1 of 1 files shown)


lib/ansible/module_utils/basic.py
```diff
@@ -2223,7 +2238,10 @@ def append_to_file(self, filename, str):
         fh.write(str)
         fh.close()
 
-    def pretty_bytes(self,size):
+    def bytes_to_human(self, size):
+        return self.pretty_bytes(size)
+
+    def pretty_bytes(self, size):
```
abadger: Instead of this, I'd rename the function to the name we like and use assignment for the backwards-compat name.  Like this:

``` python
def bytes_to_human(self, size):
    ranges = (
                   (1<<70, 'ZB'),
    [...]
    return '%.2f %s' % (float(size)/ limit, suffix)

pretty_bytes = bytes_to_human
```

lib/ansible/module_utils/basic.py
```diff
@@ -2239,6 +2255,40 @@ def pretty_bytes(self,size):
...
+            result = 0
+        elif isinstance(number, int):
+            result = number
+        elif number.isdigit():
+            result = int(number)
+        elif full in number:
+            result = int(number.replace(full,''))
+        else:
+            i = 0
+            for suffix in suffixes:
```
abadger: pythonic style, use:

``` python
for i, suffix in enumerate(suffixes):
```

rather than tracking i outside of the loop control.

abadger:
We had a discussion about whether to allow case insensitivity and allowing floats.

For case insensitivity, we know from the argument spec that we're dealing with either bytes or bits so case doesn't really matter but in common parlance, MB is megabytes and Mb is megabits.  For now decided that the latter argument would win.

For floats, decided that we can enhance it later to handle floats.  The hard part is that to avoid floating point precision problems we should deal with this as integers.  The basic algorithm for that would be:

``` python
i = 1 # We've determined that this is KiloBytes already
num = "12.76"
whole, decimal = num.split(".")
result = int(whole) * 1024**i + (int(decimal) * 1024**i)/10**len(decimal)
```

The question of i18n came up, though.  The decimal separator isn't "." in all languages.  I think at the moment "." is the only valid separator in other places as well (the float arg spec type and yaml for instance) so i don't think it's a problem to demand that "." is the separator here as well but could use some more exploration before deciding for sure.

+1 to merge.

Issue #15286, opened 2016-04-05
hostvars contains ansible_ssh_host key instead of ansible_host key


joe1chen:
##### ISSUE TYPE
- Bug Report
##### COMPONENT NAME
core

##### ANSIBLE VERSION

```
ansible 2.0.1.0
```
##### CONFIGURATION

[defaults]
host_key_checking = no
record_host_keys = no
jinja2_extensions = jinja2.ext.do
timeout = 15
gathering = smart
roles_path = roles
forks = 100

[ssh_connection]
ssh_args = -o ControlMaster=auto -o ControlPersist=30m
scp_if_ssh = True
pipelining = True
##### OS / ENVIRONMENT

OS X 10.11.4 El Capitan
##### SUMMARY

I'm upgrading from Ansible 1.x to 2.0.1.0.  The documentation mentions that `ansible_ssh_host` should become `ansible_host`.

However, in `host_vars[host]`, I am seeing that the key is still called `ansible_ssh_host` instead of `ansible_host`.
##### STEPS TO REPRODUCE

Debug output hostvars:

```
- name: Debug hostvars
  debug: msg="{{ hostvars['my_server'] }}"
```
##### EXPECTED RESULTS

```
ok: [my_server] => {
    "msg": "{...
u'ansible_host': u'54.208.99.210'
...}"
}
```
##### ACTUAL RESULTS

```
ok: [my_server] => {
    "msg": "{...
u'ansible_ssh_host': u'54.208.99.210'
...}"
}
```

jbscare:
I see where to fix this in ec2.py, and will put in a PR for that. I'm not sure if it needs a wider fix.

ansibot:
@joe1chen Greetings! Thanks for taking the time to open this issue. In order for the community to handle your issue effectively, we need a bit more information. 

Here are the items we could not find in your description:
- component name

Please set the description of this issue with this template:
https://raw.githubusercontent.com/ansible/ansible/devel/.github/ISSUE_TEMPLATE.md

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

biomassives:
This message was created automatically by mail delivery software.

A message that you sent could not be delivered to one or more of its
recipients. This is a temporary error. The following address(es) deferred:

  acmeideal@gmail.com
    Domain biomassiv.es has exceeded the max emails per hour (122/100 (122%)) allowed.  Message will be reattempted later

[...]

PR #17113, opened 2016-08-17
Update the ec2.py inventory script to set ansible_host as well as ansible_ssh_host

jbscare:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

contrib/inventory/ec2.py
##### ANSIBLE VERSION

```
ansible 2.2.0 (inventory-ec2-py-ansible-host 118d7957b4) last updated 2016/08/16 21:17:54 (GMT -400)
  config file = 
```
##### SUMMARY

contrib/inventory/ec2.py sets ansible_ssh_host, but http://docs.ansible.com/ansible/intro_inventory.html#hosts-and-groups says

> Ansible 2.0 has deprecated the "ssh" from ansible_ssh_user, ansible_ssh_host, and ansible_ssh_port to become ansible_user, ansible_host, and ansible_port.

Since the ansible_ssh_\* variables aren't removed, just deprecated, this commit simply adds ansible_host to the variables set by ec2.py.

May address some (but probably not all) of #15286.

### DIFF
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -999,7 +999,7 @@ class Ec2Inventory(object):
         self.push(self.inventory, 'ec2', hostname)
 
         self.inventory["_meta"]["hostvars"][hostname] = self.get_host_info_dict_from_instance(instance)
-        self.inventory["_meta"]["hostvars"][hostname]['ansible_ssh_host'] = dest
+        self.inventory["_meta"]["hostvars"][hostname]['ansible_host'] = dest
 
     def add_rds_instance(self, instance, region):
         ''' Adds an RDS instance to the inventory and index, as long as it is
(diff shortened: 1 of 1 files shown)


nitzmahone:
@jlaska can you think of any issues this would cause on the Tower side?

willthames:
At this point in time I wonder if we should just replace ansible_ssh_host with ansible_host rather than add it alongside.

jbscare:
@willthames : Sure, I've updated the PR to reflect that.

s-hertel:
@jbscare I can't push commits to your branch to rebase this. Would you mind doing so? This looks good to me.

jbscare:
Sure thing; done!

ryansb (approved):

PR #17115, opened 2016-08-17
Relative ansible cfg lookup

### LINKED ISSUE #11175: ansible.cfg lookup paths for ansible-playbook
##### ISSUE TYPE
Feature Idea

##### COMPONENT NAME
ansible.cfg

##### ANSIBLE VERSION
N/A

##### SUMMARY
Would ya'll be open to a PR that adds a new location for finding `ansible.cfg`?
I would like for `ansible-playbook` to search sibling to the playbook being used. I think it should be possible to run `ansible-playbook` from any directory     _without modifying global or per-user configuration files_ and have customized `ansible.cfg` settings respected.
I'm proposing a lookup precedence of:
- ANSIBLE_CONFIG (an environment variable)
- ansible.cfg (in the current directory)
- **ansible.cfg (sibling to the playbook when ansible-playbook is used)**
- .ansible.cfg (in the home directory)\n- /etc/ansible/ansible.cfg
I understand this wouldn't make sense for the `ansible` command, but I'm hopeful that isn't considered an issue."

akamensky:
##### ISSUE TYPE
- Feature Pull Request #11175 
##### COMPONENT NAME

constants.py / locating config
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel 30268f6bd0) last updated 2016/08/16 09:26:05 (GMT -400)
  lib/ansible/modules/core: (detached HEAD 45c1ae0ac1) last updated 2016/08/16 09:27:13 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD a6b34973a8) last updated 2016/08/16 09:27:30 (GMT -400)
  config file =
```
##### SUMMARY

Fixes #11175 allowing to traverse from playbook file location towards the root in order to find ansible.cfg

### DIFF
--- a/lib/ansible/constants.py
+++ b/lib/ansible/constants.py
@@ -20,6 +20,7 @@ from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
 import os
+import sys
 import tempfile
 from string import ascii_letters, digits
 
@@ -103,7 +104,7 @@ def _get_config(p, section, key, env_var, default):
     return default
 
 def load_config_file():
-    ''' Load Config File order(first found is used): ENV, CWD, HOME, /etc/ansible '''
+    ''' Load Config File order(first found is used): ENV, CWD, playbook sibling traversed, HOME, /etc/ansible '''
 
     p = configparser.ConfigParser()
 
@@ -121,6 +122,21 @@ def load_config_file():
     # FIXME: Needs deprecation? See: https://github.com/ansible/ansible/issues/11175#issuecomment-109386699
     path1 = os.getcwd() + "/ansible.cfg"
     path_list.append(path1)
+
+    # It only makes sense to search for playbook location when we use ansible-playbook and we have argv[1]
+    if len(sys.argv[0]) >= 2 and sys.argv[0].endswith("ansible-playbook"):
+        # Only check for playbook reference in position 1
+        # See https://github.com/ansible/ansible/issues/11175#issuecomment-240129682
+        first_argument = sys.argv[1]
+        if os.path.isfile(first_argument):
+            adjacent_path = os.path.dirname(os.path.abspath(first_argument))
+            while True:
+                if adjacent_path == "/":
+                    path_list.append("/ansible.cfg")
+                    break
+                else:
+                    path_list.append(adjacent_path + "/ansible.cfg")
+                adjacent_path = os.path.dirname(adjacent_path)
 
     # Get config for HOME
     path2 = os.path.expanduser("~/.ansible.cfg")
--- a/lib/ansible/constants.py
+++ b/lib/ansible/constants.py
@@ -107,16 +107,30 @@ def load_config_file():
 
     p = configparser.ConfigParser()
 
+    path_list = []
+
+    # Get config file location from ENV
     path0 = os.getenv("ANSIBLE_CONFIG", None)
     if path0 is not None:
         path0 = os.path.expanduser(path0)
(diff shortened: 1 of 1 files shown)


lib/ansible/constants.py
```diff
@@ -103,20 +104,49 @@ def _get_config(p, section, key, env_var, default):
...
+    # Get config file location from ENV
     path0 = os.getenv("ANSIBLE_CONFIG", None)
     if path0 is not None:
         path0 = os.path.expanduser(path0)
         if os.path.isdir(path0):
             path0 += "/ansible.cfg"
+    path_list.append(path0)
+
+    # Get config file location from CWD
+    # FIXME: Needs deprecation? See: https://github.com/ansible/ansible/issues/11175#issuecomment-109386699
```
abadger, line 122: From the conversation further down the issue thread, I think this is staying.

lib/ansible/constants.py
```diff
@@ -103,20 +104,49 @@ def _get_config(p, section, key, env_var, default):
...
+
+    # Get config from /etc/ansible
     path3 = "/etc/ansible/ansible.cfg"
+    path_list.append(path3)
 
-    for path in [path0, path1, path2, path3]:
+    for path in path_list:
         if path is not None and os.path.exists(path):
             try:
                 p.read(path)
```
abadger, line 152: Reading the linked feature report, I think this the spec we're trying to implement: https://github.com/ansible/ansible/issues/11175#issuecomment-110485820

So we're missing a warning here if the directory is world writable.   I think that looks something like:

``` python
import stat
[...]
cfg_directory = os.path.dirname(path)
cfg_directory_stat = os.stat(cfg_directory)
if cfg_directory_stat[stat.ST_MODE] & stat.S_IWOTH:
    display.warning('Reading ansible config file from a world readable directory (%s).  This can be a security risk.  If this is the intended location, please change permissions on the directory.' % cfg_directory)
```

abadger:
Hi @akamensky sorry it's taken a while to look at this feature request.  Right now is the prime time to get new features into Ansible (for the 2.3.x release).  I've commented on this and if you're interested in fixing it up I'll get it in.  If not, let me know and I'll try to get it the rest of the way.

If you update and I don't appear to have seen it, please feel free to get in touch with me on IRC.  I'm abadger1999 in #ansible-devel on irc.freenode.net.

abadger:
I've taken a quick look at the cli code... I think I agree with you about there being a need for a large reworking of the code to make that a clean change.  I think it could be done uncleanly with just a few things changed but it would be fragile.  Future devs would have to know that use of constants and parsing args were being done in a particular order and a particular style out of necessity but that wouldn't be obvious from just reading the code.  What's being worked on here seems like it is good enough for now and we can work on other fixes if there's a need later.

ssbarnea:
Do we have any updates on this? I would file really useful if ansible would detect the location of the ansible.cfg by looking for the parent folders. 

This would allow people to keep playbooks inside subfolders of their repos without having to be extremely careful from which CWD they are running a playbook.

bcoca:
https://github.com/ansible/proposals/issues/35 will provide this 'feature' in a saner way

lassizci:
> This would allow people to keep playbooks inside subfolders of their repos without having to be extremely careful from which CWD they are running a playbook.

That and having e.g `forks` and other settings that are rather local and not project or repo specific but which used to get overwritten by repo config.

calfonso:
Hi!

We are closing this in favor of a future implementation, noted in the comment regarding proposal 35.

PR #17126, opened 2016-08-17
actions/unarchive: fix unarchive from remote url

dustymabe:
Currently unarchive from remote url does not work because the core
unarchive module was updated to support 'remote_src' [1], but the
unarchive action plugin was not updated for this. This causes failures
because the action plugin assumes it needs to copy a file to the
remote server, but in the case of downloading a file from a remote
url a local file does not exist, so an error occurs when the file is
not found.

[1] https://github.com/ansible/ansible-modules-core/commit/467516e

### DIFF
--- a/lib/ansible/plugins/action/unarchive.py
+++ b/lib/ansible/plugins/action/unarchive.py
@@ -38,8 +38,19 @@ class ActionModule(ActionBase):
 
         source  = self._task.args.get('src', None)
         dest    = self._task.args.get('dest', None)
-        copy    = boolean(self._task.args.get('copy', True))
+        remote_src = boolean(self._task.args.get('remote_src', False))
         creates = self._task.args.get('creates', None)
+
+        # "copy" is deprecated in favor of "remote_src".
+        if 'copy' in self._task.args:
+            # They are mutually exclusive.
+            if 'remote_src' in self._task.args:
+                result['failed'] = True
+                result['msg'] = "parameters are mutually exclusive: ('copy', 'remote_src')"
+                return result
+            # We will take the information from copy and store it in
+            # the remote_src var to use later in this file.
+            remote_src = not boolean(self._task.args.get('copy'))
 
         if source is None or dest is None:
             result['failed'] = True
@@ -66,7 +77,7 @@ class ActionModule(ActionBase):
         dest = self._remote_expand_user(dest) # CCTODO: Fix path for Windows hosts.
         source = os.path.expanduser(source)
 
-        if copy:
+        if not remote_src:
             try:
                 source = self._loader.get_real_file(self._find_needle('files', source))
             except AnsibleError as e:
@@ -87,7 +98,7 @@ class ActionModule(ActionBase):
             self._remove_tmp_path(tmp)
             return result
 
-        if copy:
+        if not remote_src:
             # transfer the file to a remote tmp location
             tmp_src = self._connection._shell.join_path(tmp, 'source')
             self._transfer_file(source, tmp_src)
@@ -95,7 +106,7 @@ class ActionModule(ActionBase):
         # handle diff mode client side
         # handle check mode client side
 
-        if copy:
+        if not remote_src:
             # fix file permissions when the copy is done as a different user
             self._fixup_perms((tmp, tmp_src), remote_user)
             # Build temporary module_args.
--- a/test/integration/roles/test_unarchive/tasks/main.yml
+++ b/test/integration/roles/test_unarchive/tasks/main.yml
@@ -299,7 +299,7 @@
     src: "test-unarchive-nonascii-くらとみ.tar.gz"
     dest: "{{ output_dir }}/test-unarchive-nonascii-くらとみ-tar-gz"
     mode: "u+rwX,go+rX"
-    remote_src: yes
+    remote_src: no
   register: nonascii_result0
 
 - name: Check that file is really there


dagwieers:
@dustymabe Yes, the specific error can have multiple root causes. Unfortunately the maintainers have not been involved with the copy/remote_src change. I would take this up with the author (@jimi-c) of that commit (https://github.com/ansible/ansible-modules-core/commit/467516e)

dagwieers:
BTW I think the goal is to have both `copy` (deprecated) and `remote_src` to work. So I think the logic should support that.

dustymabe:
> BTW I think the goal is to have both copy (deprecated) and remote_src to work. So I think the logic should support that.

hey @dagwieers - the change I made should support both `copy` (deprecated) and `remote_src` just as you describe.

dagwieers:
@dustymabe Then I am most likely missing something, I don't see the case where copy is set and remote_src is not (which is what most people do right now). Or do we handle that elsewhere ? If so, the mutual-exclusiveness of both options should be handled there as well I presume.

dustymabe:
```
-        copy    = boolean(self._task.args.get('copy', True))
+        remote_src = boolean(self._task.args.get('remote_src', False))
         creates = self._task.args.get('creates', None)

+        # "copy" is deprecated in favor of "remote_src".
+        if 'copy' in self._task.args:
+            # They are mutually exclusive.
+            if 'remote_src' in self._task.args:
+                result['failed'] = True
+                result['msg'] = "parameters are mutually exclusive: ('copy', 'remote_src')"
+                return result
+            # We will take the information from copy and store it in
+            # the remote_src var to use later in the playbook
+            remote_src = not boolean(self._task.args.get('copy'))
```

you can see from the above code that we check to see if `'copy' in self._task.args` and if it is then we set `remote_src = not boolean(self._task.args.get('copy'))`. So if a user provides `copy` then it gets translated  to `remote_src` internally for later use within the code. This is so that we don't have to handle both `copy` and `remote_src` logic all throughout the code within the file. 

also in the code above is a check to make sure the user didn't provide both `copy` and `remote_src`, as they are mutually exclusive

lib/ansible/plugins/action/unarchive.py
```diff
@@ -38,9 +38,20 @@ def run(self, tmp=None, task_vars=None):
...
 
+        # "copy" is deprecated in favor of "remote_src".
+        if 'copy' in self._task.args:
+            # They are mutually exclusive.
+            if 'remote_src' in self._task.args:
+                result['failed'] = True
+                result['msg'] = "parameters are mutually exclusive: ('copy', 'remote_src')"
+                return result
+            # We will take the information from copy and store it in
+            # the remote_src var to use later in the playbook
```
dustymabe: this should read `to use later in this file` I didn't mean to say `later in this playbook`. I will update this soon.

dagwieers: Right, and the same should logic should be used in the unarchive module IMO.

I was wondering whether we should fail for mutual exclusive options (when deprecated), or simply take the non-deprecated option and throw a warning to the user. Would be nice if Ansible does this consistently everywhere options are being deprecated.

dagwieers:
@dustymabe Thanks for clarifying, I missed that somehow.

lib/ansible/plugins/action/unarchive.py
dustymabe: > Right, and the same should logic should be used in the unarchive module IMO.

what logic specifically? I tried to match what the unarchive module is doing. From a users perspective they can provide either `copy` or `remote_src` or neither of them and it should work as desired. The logic is consistent between both. 

> I was wondering whether we should fail for mutual exclusive options (when deprecated), or simply take the non-deprecated option and throw a warning to the user. Would be nice if Ansible does this consistently everywhere options are being deprecated.

I decided to fail because that is what the module does already: https://github.com/ansible/ansible-modules-core/blob/91a839f1e3de58be8b981d3278aa5dc8ff59c508/files/unarchive.py#L721

dagwieers: Hmmm, I don't think they are mutually exclusive. One deprecates the other and the error message should reflect that IMO. Anyhow, fine by me.

dagwieers: BTW That does not change the fact that the logic in unarchive still has to check both copy and remote_src in various places, which is overly complex for a deprecated option IMO. That's why I favored your implementation.

jimi-c:
@dustymabe this is failing integration tests (example here: https://app.shippable.com/runs/57b522fdf635a90f00519a27/13/console).

lib/ansible/plugins/action/unarchive.py
dustymabe: > BTW That does not change the fact that the logic in unarchive still has to check both copy and remote_src in various places, which is overly complex for a deprecated option IMO. That's why I favored your implementation.

Yeah - the logic is implemented twice. Once in actions/unarchive.py and once in the unarchive.py in the module. I've been told modules will be merged into the main repo in the future so this code can all live in the same repo and perhaps share the logic for parsing the task's arguments. That would be a better solution.

dustymabe:
@jimi-c - can you link me to the source of that failing test so I can try to run it myself?

jimi-c:
@dustymabe those all live in the `test/integration` directory in ansible/ansible. The failure in the link I posted above specifically comes from this file: https://github.com/ansible/ansible/blob/devel/test/integration/roles/test_unarchive/tasks/main.yml#L295

You can test things locally yourself if you have docker/a few other things installed as follows:

`TEST=integration TARGET=all IMAGE=ansible/ansible:<distro> test/utils/shippable/ci.sh` (where the distro is any of our docker images here: https://hub.docker.com/r/ansible/ansible/tags/)

dustymabe:
I fixed the failing test. please evaluate my 2nd commit

dagwieers:
shipit

jimi-c:
Merged, thanks!

jimi-c:
Was this impacting devel only? I looked at cherry-picking it to stable-2.1 but it didn't apply cleanly, and looking at why leads me to believe the breaking changes for unarchive were not in stable-2.1.

dagwieers:
@jimi-c Correct the `remote_src` was a recent change (that as a maintainer surprised me ;-)). This is Ansible v2.2 material AFAICT.

dustymabe:
I originally started looking at unarchive because it was failing for me when using ansible-container - I think it is ansible 2.1. Turns out I was reading the online docs (which mention `remote_src`) so I was using `remote_src` vs `copy` which was causing a failure. 

I started to dig in and tested against latest devel, which also failed for me (in a different way) so I just fixed it there. 

Turns out if I just used `copy` vs `remote_src` in 2.1 it would have worked.

PR #17131, opened 2016-08-18
Fix facts.py for python3

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

module_utils/facts.py
##### ANSIBLE VERSION

```
devel
```
##### SUMMARY

Previous to this commit, running setup under python3 would traceback:

```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: 'str' doe
s not support the buffer interface
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "module_stderr": "Traceback (most recent call last)
:\n  File \"/var/tmp/ansible_mv8a3v4a/ansible_module_setup.py\", line 134, in <module>\n    main()\n  File \"/var/tmp
/ansible_mv8a3v4a/ansible_module_setup.py\", line 126, in main\n    data = get_all_facts(module)\n  File \"/var/tmp/a
nsible_mv8a3v4a/ansible_modlib.zip/ansible/module_utils/facts.py\", line 3314, in get_all_facts\n  File \"/var/tmp/an
sible_mv8a3v4a/ansible_modlib.zip/ansible/module_utils/facts.py\", line 3253, in ansible_facts\n  File \"/var/tmp/ans
ible_mv8a3v4a/ansible_modlib.zip/ansible/module_utils/facts.py\", line 202, in __init__\n  File \"/var/tmp/ansible_mv
8a3v4a/ansible_modlib.zip/ansible/module_utils/facts.py\", line 479, in get_caps_facts\nTypeError: 'str' does not sup
port the buffer interface\n", "module_stdout": "", "msg": "MODULE FAILURE", "parsed": false}
```

With these fixes applied, facts can be successfully gathered.  Test by running:

```
python3 bin/ansible localhost -m setup
```

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -32,6 +32,10 @@ import datetime
 import getpass
 import pwd
 
+from ansible.module_utils.basic import get_all_subclasses
+from ansible.module_utils.six import PY3, iteritems
+
+# py2 vs py3; replace with six via ansiballz
 try:
     # python2
     import ConfigParser as configparser
@@ -39,10 +43,6 @@ except ImportError:
     # python3
     import configparser
 
-from ansible.module_utils.basic import get_all_subclasses
-from ansible.module_utils.six import PY3
-
-# py2 vs py3; replace with six via ansiballz
 try:
     # python2
     from StringIO import StringIO
@@ -50,24 +50,12 @@ except ImportError:
     # python3
     from io import StringIO
 
-
 try:
     # python2
     from string import maketrans
 except ImportError:
     # python3
     maketrans = str.maketrans # TODO: is this really identical?
-
-try:
-    dict.iteritems
-except AttributeError:
-    # Python 3
-    def iteritems(d):
-        return d.items()
-else:
-    # Python 2
-    def iteritems(d):
-        return d.iteritems()
 
 try:
     # Python 2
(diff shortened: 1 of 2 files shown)


abadger:
@bcoca and I discussed whether run_command() should contain the text decoding functionality itself.  We decided not to block this PR but there are some pros and cons to doing so and some right and wrong ways to add it:
- If we modify run_command, run_command should always return text.  Returning either text or bytes would be bad API.
- We would want the run_command API to take an encoding and an errors parameter so that the calling code can control what information they get back.  These could default to 'utf-8' and 'strict'.
- If the calling code expects undecodable bytes and wants to just pass those through to some other interface, they would need to give it an encoding that covers all possible bytes and re-encode to bytes themselves:

``` python
rc, temp_out, temp_err = module.run_command(cmd, encoding='latin-1', errors='strict')
bytes_real_out = temp_out.encode('latin-1')
output = open(filename, 'wb')
output.write(bytes_real_out)
output.close()
```
- There is extra work involved in many cases as a good run_command() API would decode both stdout and stderr while many of the places we use it only care about stdout.

anmolnagpal:
Hello @abadger 
Getting the same error with python3 

```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: 'str' does not support the buffer interface
```

abadger:
Please open a new ticket with details of how to reproduce (ansible version, python version, playbook that's executing and output when run with -vvv so I can see the traceback.  Thanks!

PR #17141, opened 2016-08-18
Add foreman callback plugin

agx:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

plugins/callback/foreman.py
##### ANSIBLE VERSION

```
ansible 2.2.0 (foreman-callback 6f2cfacca8) last updated 2016/08/18 20:03:16 (GMT +200)
  lib/ansible/modules/core: (detached HEAD 91a839f1e3) last updated 2016/08/18 20:06:19 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD 1aeb9f8a8c) last updated 2016/08/18 20:06:19 (GMT +200)
```
##### SUMMARY

Add a callback to report facts and reports to the foreman:

```
https://theforeman.org/
```

### DIFF
--- a/lib/ansible/plugins/callback/foreman.py
+++ b/lib/ansible/plugins/callback/foreman.py
@@ -1,0 +1,200 @@
+# -*- coding: utf-8 -*-
+# (C) 2015, 2016 Daniel Lobato <elobatocs@gmail.com>
+#     2016 Guido Günther <agx@sigxcpu.org>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import os
+from datetime import datetime
+from collections import defaultdict
+import json
+import time
+
+try:
+    import requests
+    HAS_REQUESTS = True
+except ImportError:
+    HAS_REQUESTS = False
+
+from ansible.plugins.callback import CallbackBase
+
+
+class CallbackModule(CallbackBase):
+    """
+    This callback will report facts and reports to Foreman https://theforeman.org/
+
+    It makes use of the following environment variables:
+
+    FOREMAN_URL: URL to the Foreman server
+    FOREMAN_SSL_CERT: X509 certificate to authenticate to Foreman if
+      https is used
+    FOREMAN_SSL_KEY: the corresponding private key
+    FOREMAN_SSL_VERIFY: wether to verify the Foreman certificate
+      It can be set to '1' to verify SSL certificates using the
+      installed CAs or to a path pointing to a CA bundle. Set to '0'
+      to disable certificate checking.
+    """
+
+    CALLBACK_VERSION = 2.0
+    CALLBACK_TYPE = 'notification'
+    CALLBACK_NAME = 'foreman'
+    CALLBACK_NEEDS_WHITELIST = True
+
+    FOREMAN_URL = os.getenv('FOREMAN_URL', "http://localhost:3000")
+    FOREMAN_SSL_CERT = (os.getenv('FOREMAN_SSL_CERT',
+                                  "/etc/foreman/client_cert.pem"),
+                        os.getenv('FOREMAN_SSL_KEY',
+                                  "/etc/foreman/client_key.pem"))
+    FOREMAN_SSL_VERIFY = os.getenv('FOREMAN_SSL_VERIFY', "1")
+    FOREMAN_HEADERS = {
+      "Content-Type": "application/json",
+      "Accept": "application/json"
+    }
+    TIME_FORMAT = "%Y-%m-%d %H:%M:%S %f"
+
+    def __init__(self):
+        super(CallbackModule, self).__init__()
+        self.items = defaultdict(list)
+        self.start_time = int(time.time())
+
+        if HAS_REQUESTS:
+            requests_major = int(requests.__version__.split('.')[0])
+            if requests_major >= 2:
+                self.ssl_verify = self._ssl_verify()
+            else:
+                self._disable_plugin('The `requests` python module is too old.')
+        else:
+            self._disable_plugin('The `requests` python module is not installed.')
+
+    def _disable_plugin(self, msg):
+        self.disabled = True
+        self._display.warning(msg + ' Disabling the Foreman callback plugin.')
+
+    def _ssl_verify(self):
+        if self.FOREMAN_SSL_VERIFY.lower() in ["1", "true", "on"]:
+            verify = True
+        elif self.FOREMAN_SSL_VERIFY.lower() in ["0", "false", "off"]:
+            requests.packages.urllib3.disable_warnings()
+            self._display.warning("SSL verification of %s disabled" %
+                                  self.FOREMAN_URL)
+            verify = False
+        else:  # Set ta a CA bundle:
+            verify = self.FOREMAN_SSL_VERIFY
+        return verify
+
+    def send_facts(self, host, data):
+        """
+        Sends facts to Foreman, to be parsed by foreman_ansible fact
+        parser.  The default fact importer should import these facts
+        properly.
+        """
+        data["_type"] = "ansible"
+        data["_timestamp"] = datetime.now().strftime(self.TIME_FORMAT)
+        facts = {"name": host,
+                 "facts": data,
+                 }
+        requests.post(url=self.FOREMAN_URL + '/api/v2/hosts/facts',
+                      data=json.dumps(facts),
+                      headers=self.FOREMAN_HEADERS,
+                      cert=self.FOREMAN_SSL_CERT,
+                      verify=self.ssl_verify)
+
+    def _build_log(self, data):
+        logs = []
+        for entry in data:
+            source, msg = entry
+            if 'failed' in msg:
+                level = 'err'
+            else:
+                level = 'notice' if 'changed' in msg and msg['changed'] else 'info'
+            logs.append({"log": {
+                'sources': {'source': source},
+                'messages': {'message': json.dumps(msg)},
+                'level':     level
+                }})
+        return logs
+
+    def send_reports(self, stats):
+        """
+        Send reports to Foreman to be parsed by its config report
+        importer. THe data is in a format that Foreman can handle
+        without writing another report importer.
+        """
+        status = defaultdict(lambda: 0)
+        metrics = {}
+
+        for host in stats.processed.keys():
+            sum = stats.summarize(host)
+            status["applied"] = sum['changed']
+            status["failed"] = sum['failures'] + sum['unreachable']
+            status["skipped"] = sum['skipped']
+            log = self._build_log(self.items[host])
+            metrics["time"] = {"total": int(time.time()) - self.start_time}
+            now = datetime.now().strftime(self.TIME_FORMAT)
+            report = {
+                "report": {
+                    "host": host,
+                    "reported_at": now,
+                    "metrics": metrics,
+                    "status": status,
+                    "logs": log,
+                    }
+                 }
+            # To be changed to /api/v2/config_reports in 1.11.  Maybe we
+            # could make a GET request to get the Foreman version & do
+            # this automatically.
+            requests.post(url=self.FOREMAN_URL + '/api/v2/reports',
+                          data=json.dumps(report),
+                          headers=self.FOREMAN_HEADERS,
+                          cert=self.FOREMAN_SSL_CERT,
+                          verify=self.ssl_verify)
+            self.items[host] = []
+
+    def append_result(self, result):
+        name = result._task.get_name()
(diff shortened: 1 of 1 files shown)


lib/ansible/plugins/callback/foreman.py
```diff
@@ -0,0 +1,246 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+from datetime import datetime
+from collections import defaultdict
+import json
+import requests
```
bcoca: We try to avoid the requests library as it has many issues across older versions and even major incompatibilities across it's own versions, look at using the ansible open_url/fetch_url functions instead.

lib/ansible/plugins/callback/foreman.py
```diff
@@ -0,0 +1,246 @@
...
+import time
+
+try:
+    from ansible.plugins.callback import CallbackBase
+    parent_class = CallbackBase
+except ImportError:
+    parent_class = object
```
bcoca: unsure of why this is happening, since you are adding to ship with ansible, callbackbase should always be available

lib/ansible/plugins/callback/foreman.py
```diff
@@ -0,0 +1,246 @@
...
+            # this automatically.
+            requests.post(url=FOREMAN_URL + '/api/v2/reports',
+                          data=report_json,
+                          headers=FOREMAN_HEADERS,
+                          cert=FOREMAN_SSL_CERT,
+                          verify=self.ssl_verify)
+
+    def on_any(self, *args, **kwargs):
+        pass
+
```
bcoca: there should be no need to implement these when using callback base, also you should use v2_ events as the old ones are not guaranteed to work, in some cases they get slightly different data or lack the data they had in pre 2.0 Ansible.

lib/ansible/plugins/callback/foreman.py
```diff
@@ -0,0 +1,246 @@
...
+    def runner_on_failed(self, host, res, ignore_errors=False):
+        self.items[host].append(res)
+
+    def runner_on_ok(self, host, res):
```
bcoca: this is never invoked as the v2_ method that would forward to this in the callbackbase is overridden below

agx:
@bcoca thanks for having a look at this PR. I think I addressed your concerns except for the python-requests removal. I'm happy to switch to open_url / fetch_url but I can't see how to get the current behaviour (allow to specify X509 certificate, key and CA) without reimplementing parts of requests myself or did I miss somthing in the *_url functions?

agx:
Hi, anything I can do to move this further? I'd like to stick to requests since we need client certificates for authentication against foreman but if it's a must have to use the *_url functions I can copy over the necessary code.

bcoca:
if you insist on requests i would specify version and handle either it being missing or of insufficient/incompatible version. The API of that library has changed in backwards incompatible ways.

agx:
@bcoca thanks for your patience! I've added the check for requests. We're using very few parts of the requests API so I did not add any checks yet. I've tried several versions 2.10.0, 2.7.0, 2.4 and it worked there.

mattclay:
Tests on shippable failed because this PR branch was created before shippable was enabled. If you rebase the branch the tests will pass, although in this case the tests will just no-op since there are no tests for this plugin.

agx:
@mattclay I rebased but the tests still fail.

mattclay:
@agx Look at the console log here for the test failure:

https://app.shippable.com/runs/57c7b815236da20f00dc7638/35/console

The error is:

2016-09-01 05:12:52 == Missing from **future** import (absolute_import, division, print_function) ==

agx:
On Thu, Sep 01, 2016 at 12:10:31AM -0700, Matt Clay wrote:

> @agx Look at the console log here for the test failure:
> 
> https://app.shippable.com/runs/57c7b815236da20f00dc7638/35/console
> 
> The error is:
> 
> 2016-09-01 05:12:52 == Missing from **future** import (absolute_import, division, print_function) ==

Thanks, I also fixed another failure but the current one does not look
related:

```
https://app.shippable.com/runs/57c86e08640b890e00210509/1/console
```

Just repin?

agx:
@bcoca @mattclay Tests pass now, any chance this can be applied?

bcoca:
@agx, still, you need to specify requests versions, many machines come with <2.0 installed and I know it does not work because the API changed in incompatible ways.

agx:
On Mon, Sep 05, 2016 at 04:20:02PM -0700, Brian Coca wrote:

> @agx, still, you need to specify requests versions, many machines come with <2.0 installed and I know it does not work because the API changed in incompatible ways.

Fixed now. Cheking for >= 2.0.

agx:
@mattclay Tests are failing again but I don't think this is caused by the change:

2016-09-06 05:52:30 ERROR! Unexpected Exception: 'ascii' codec can't decode byte 0xc3 in position 95: ordinal not in range(128)

@bcoca  Check for requests is now there (I noticed that no other code that uses requests checks for the version. Should we move this to common code and add checks as well in a follow up commit?)

mattclay:
@agx The error is due to the non-ascii character in your name. Removing it eliminates the error.

Here's the full traceback:

```
  File "/root/ansible/bin/ansible-playbook", line 97, in <module>
    exit_code = cli.run()
  File "/root/ansible/lib/ansible/cli/playbook.py", line 154, in run
    results = pbex.run()
  File "/root/ansible/lib/ansible/executor/playbook_executor.py", line 82, in run
    self._tqm.load_callbacks()
  File "/root/ansible/lib/ansible/executor/task_queue_manager.py", line 277, in load_callbacks
    for callback_plugin in callback_loader.all(class_only=True):
  File "/root/ansible/lib/ansible/plugins/__init__.py", line 386, in all
    self._module_cache[path] = self._load_module_source(name, path)
  File "/root/ansible/lib/ansible/plugins/__init__.py", line 321, in _load_module_source
    module = imp.load_source(name, path, module_file)
  File "/usr/lib/python3.5/imp.py", line 172, in load_source
    module = _load(spec)
  File "<frozen importlib._bootstrap>", line 693, in _load
  File "<frozen importlib._bootstrap>", line 673, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 661, in exec_module
  File "<frozen importlib._bootstrap_external>", line 766, in get_code
  File "/usr/lib/python3.5/imp.py", line 156, in get_data
    return file.read()
  File "/usr/lib/python3.5/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 94: ordinal not in range(128)
```

The `imp` module is deprecated as of python 3.4 in favor of `importlib`, which is probably why this error only shows up when we test on python 3.

@abadger: This looks like a python 3 compatibility issue we'll need to address.

lib/ansible/plugins/callback/foreman.py
```diff
@@ -0,0 +1,211 @@
...
+  FOREMAN_URL: URL to the Foreman server
+  FOREMAN_SSL_CERT: X509 certificate to authenticate to Foreman if
+    https is used
+  FOREMAN_SSL_KEY: the corresponding private key
+  FOREMAN_SSL_VERIFY: wether to verify the Foreman certificate
+    It can be set to '1' to verify SSL certificates using the
+    installed CAs or to a path pointing to a CA bundle. Set to '0'
+    to disable certificate checking.
+
+"""
```
mattclay: This docstring should be used for the `CallbackModule` class instead of the module.

lib/ansible/plugins/callback/foreman.py
```diff
@@ -0,0 +1,211 @@
...
+
+try:
+    import requests
+    HAS_REQUESTS = True
+except ImportError:
+    HAS_REQUESTS = False
+
+from ansible.plugins.callback import CallbackBase
+
+FOREMAN_URL = os.getenv('FOREMAN_URL', "http://localhost:3000")
```
mattclay: Move `FOREMAN_URL` and the other variables that follow inside the `CallbackModule` class.

lib/ansible/plugins/callback/foreman.py
```diff
@@ -0,0 +1,211 @@
...
+    "metrics": %(metrics)s,
+    "status": %(status)s,
+    "logs" : %(log)s
+  }
+}
+"""
+
+
+class CallbackModule(CallbackBase):
+    """
```
mattclay: The various `CALLBACK_` prefixed class attributes are missing. Take a look at the other callbacks for an example. In particular, this is needed:

``` python
CALLBACK_NEEDS_WHITELIST = True
```

abadger:
imp on python3 issue should be fixed.  Reading the file as bytes instead of text and passing it to imp allows imp to properly find the encoding.  You'll need to rebase to pick up that change.

mattclay:
@agx I reran the build after the python 3 fix was made by @abadger and all tests are now passing.

agx:
@mattclay Thanks for looking into the testsuite! I've addressed your comments. One thing I noticed:

As soon as I add any of the CALLBACK_\* variables to the callback, the callback won't be invoked anymore whith:

```
 ansible -i localhost, -msetup -clocal localhost
```

while it still gets invoked fine with a playbook:

```
 ansible-playbook -i localhost, -l localhost play.yml -c local
```

Is this intentional?

mattclay:
@agx Yes. Callbacks included with Ansible must be enabled by setting the `ANSIBLE_CALLBACK_WHITELIST` environment variable.

For your plugin, you would use: `ANSIBLE_CALLBACK_WHITELIST=foreman`

Without this whitelisting, all callbacks shipped with Ansible would be used when running ansible-playbook.

agx:
@mattclay  I know I have to whitelist the plugin (and I do so via ansible.cfg) but it only gets run with ansible-playbook not with ansible -msetup ...

This isn't a bug in the plugin itself but looks rather like a behaviour change between plugins using CALLBACK_ and those that dont.

mattclay:
@agx You need to set `ANSIBLE_LOAD_CALLBACK_PLUGINS=true` to use additional callbacks with `ansible`.

agx:
@mattclay ahh...thanks for the hint. Yes, with `ANSIBLE_LOAD_CALLBACK_PLUGINS=true` this works as expected.

agx:
@mattclay  @bcoca  just doublechecked that the plugin works with Foreman 1.12.x , would be great if it could be merged now.

lib/ansible/plugins/callback/foreman.py
```diff
@@ -0,0 +1,210 @@
...
+    def send_facts(self, host, data):
+        """
+        Sends facts to Foreman, to be parsed by foreman_ansible fact
+        parser.  The default fact importer should import these facts
+        properly.
+        """
+        data["_type"] = "ansible"
+        data["_timestamp"] = datetime.now().strftime(self.TIME_FORMAT)
+        data = json.dumps(data)
+        facts_json = self.FACTS_FORMAT % dict(host=host, data=data)
```
mattclay: Why not use `json.dumps` to create `facts_json`?

lib/ansible/plugins/callback/foreman.py
```diff
@@ -0,0 +1,210 @@
...
+        for host in stats.processed.keys():
+            sum = stats.summarize(host)
+            status["applied"] = sum['changed']
+            status["failed"] = sum['failures'] + sum['unreachable']
+            status["skipped"] = sum['skipped']
+            log = self._build_log(self.items[host])
+            metrics["time"] = {"total": int(time.time()) - self.start_time}
+            self.items[host] = []
+
+            report_json = self.REPORT_FORMAT % dict(host=host,
```
mattclay: Why not use `json.dumps` to create `report_json`?

agx:
On Fri, Sep 16, 2016 at 02:21:11PM -0700, Matt Clay wrote:

> mattclay commented on this pull request.
> 
> > -                                  self.FOREMAN_URL)
> > -            verify = False
> > -        else:  # Set ta a CA bundle:
> > -            verify = self.FOREMAN_SSL_VERIFY
> > -        return verify
> >   +
> > -    def send_facts(self, host, data):
> > -        """
> > -        Sends facts to Foreman, to be parsed by foreman_ansible fact
> > -        parser.  The default fact importer should import these facts
> > -        properly.
> > -        """
> > -        data["_type"] = "ansible"
> > -        data["_timestamp"] = datetime.now().strftime(self.TIME_FORMAT)
> > -        data = json.dumps(data)
> > -        facts_json = self.FACTS_FORMAT % dict(host=host, data=data)
> 
> Why not use `json.dumps` to create `facts_json`?

I think the idea here was to have the format facts should be posted
(FACTS_FORMAT) separated from the (json encoded) facts data. This way
once ca see at one glance what the format should look like. I can change
this to s.th. like:

```
 json.dumps({
    "name": ...
    "data": {
            "_type": "ansible"
            ...
    }       
```

but I think the current form is better readable.

> > -        get a report json that Foreman can handle without writing
> > -        another report importer.
> > -        """
> > -        status = defaultdict(lambda: 0)
> > -        metrics = {}
> >   +
> > -        for host in stats.processed.keys():
> > -            sum = stats.summarize(host)
> > -            status["applied"] = sum['changed']
> > -            status["failed"] = sum['failures'] + sum['unreachable']
> > -            status["skipped"] = sum['skipped']
> > -            log = self._build_log(self.items[host])
> > -            metrics["time"] = {"total": int(time.time()) - self.start_time}
> > -            self.items[host] = []
> >   +
> > -            report_json = self.REPORT_FORMAT % dict(host=host,
> 
> Why not use `json.dumps` to create `report_json`?

Same here.

agx:
@mattclay I removed the hand built json making the code more compact. Anything else I can do to get this merged?

agx:
On Sun, Oct 02, 2016 at 09:13:04PM -0700, Brian Coca wrote:

> Merged #17141.

Thanks!
 -- Guido

PR #17143, opened 2016-08-18
become_method: Add support for ksu

dsommers:
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

playbook/play_context.py
##### ANSIBLE VERSION

git devel commit 677a34a191b7e5cde815ce36737fcada222f20c4
##### SUMMARY

On kerberized systems using ksu will simplify privilege escalation.

The requirements for ksu to work is that the calling user have
a valid kerberos ticket, that the remote system supports GSSAPI
authentication where the kerberos ticket is forwarded and that
the target user have a .k5login file with the kerberos principal
of the user calling ksu listed.  With this in place, no password
will be needed.

### DIFF
--- a/lib/ansible/playbook/play_context.py
+++ b/lib/ansible/playbook/play_context.py
@@ -552,6 +552,10 @@ class PlayContext(Base):
 
                 becomecmd = '%s -u %s %s' % (exe, self.become_user, command)
 
+            elif self.become_method == 'ksu':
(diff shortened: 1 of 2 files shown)


dsommers:
I see that a similar solution have been merged already from a different contributor in commit d54f5277d50748abb41.  Closing this one.

PR #17155, opened 2016-08-19
Make _to_filesystem_str return a string on py3

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

ansible
##### SUMMARY

On python 3, this return a bytes object, so this fail
with the following traceback:

```
  File \"/tmp/ansible_rs2o3x04/ansible_module_authorized_key.py\", line 499, in <module>
    main()
  File \"/tmp/ansible_rs2o3x04/ansible_module_authorized_key.py\", line 493, in main
    results = enforce_state(module, module.params)
  File \"/tmp/ansible_rs2o3x04/ansible_module_authorized_key.py\", line 468, in enforce_state
    writekeys(module, keyfile(module, user, do_write, path, manage_dir), existing_keys)
  File \"/tmp/ansible_rs2o3x04/ansible_module_authorized_key.py\", line 371, in writekeys
    module.atomic_move(tmp_path, filename)
  File \"/tmp/ansible_rs2o3x04/ansible_modlib.zip/ansible/module_utils/basic.py\", line 1905, in atomic_move
  File \"/tmp/ansible_rs2o3x04/ansible_modlib.zip/ansible/module_utils/basic.py\", line 816, in selinux_context
TypeError: in method 'lgetfilecon_raw', argument 1 of type 'char const *'
```

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -790,6 +790,8 @@ class AnsibleModule(object):
         '''
         if isinstance(path, unicode):
             path = path.encode("utf-8")
+        if isinstance(path, bytes):
+            path = path.decode("utf-8")
         return path
 
     # If selinux fails to find a default, return an array of None
--- a/test/units/module_utils/test_basic.py
+++ b/test/units/module_utils/test_basic.py
@@ -588,8 +588,8 @@ class TestModuleUtilsBasic(ModuleTestCase):
             argument_spec = dict(),
         )
 
-        self.assertEqual(am._to_filesystem_str(u'foo'), b'foo')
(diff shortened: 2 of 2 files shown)


mscherer:
To reproduce, you just need to use ansible against a freshly installed F24 with ansible_python_interpreter pointing to python3. I am not 100% of the validity of the fix I did on the test suite.

mscherer:
Ok so this change seems to break copy on python 2, despites test suite coverage. I guess the fix is not right, so I will defer for now.

abadger:
This may be right and copy just needs updating as well.  I'll look into copy soon (I'm trying to get to the place where the stat integration tests can be run and copy is part of that).

PR #17170, opened 2016-08-20
Update command for generating password

thomaslorentsen:
##### ISSUE TYPE
- Docs Pull Request
##### COMPONENT NAME

/docsite/rst/faq.rst
##### SUMMARY

When using `mkpasswd`, the option to generate a password is `--method=sha-512` instead of `--method=SHA-512` which produces a hash not found error


nitzmahone:
thanks!

PR #17177, opened 2016-08-22
Add RHEV host detection support

### LINKED ISSUE #17058: RHEV virtualization is not detected
##### ISSUE TYPE

- Bug Report
##### COMPONENT NAME

setup module_util
##### SUMMARY

The default facts do not cover RHEV virtualization, and while RHEV is using KVM underneath it would be useful for users to make a clear distinction between a RHEV host and a normal KVM host.

I don't know if there is a way to determine if a VM is a RHEV guest, but it probably doesn't matter that much as it does for hosts.

dagwieers:
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

facts
##### SUMMARY

This adds RHEV host detection support based on a running 'vdsm' process and the existence of _/rhev/_ (which are both part of the vdsm RPM package in a RHEV installation). Without this change, a RHEV host would be reported as a kvm host (which is also true, but often not specific enough).

This closes #17058

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -3119,7 +3119,20 @@ class LinuxVirtual(Virtual):
                 modules.append(data[0])
 
             if 'kvm' in modules:
-                self.facts['virtualization_type'] = 'kvm'
+                
+                # Build a list of process-names (to detect vdsm running on RHEV)
+                proc_list = []
+                for f in glob.glob('/proc/[0-9]*/comm'):
+                    try:
+                        proc_list.append(open(f).read().rstrip())
+                    except:
+                        pass
+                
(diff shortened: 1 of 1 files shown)


dagwieers:
Shippable test failed because of "Error: Error downloading packages:\n  Curl error (52): Server returned nothing (no headers, no data) for https://mirrors.fedoraproject.org/metalink?repo=updates-released-f23&arch=x86_64 [Empty reply from server]" which has nothing to do with this patch :-(

gundalow:
Does this need updating to check for `rhv` as well as (or instead of) `rhev`?

Also what should `virtualization_type` be now `RHV` or `RHEV`?

dagwieers:
@gundalow Probably yes, if that is the new consensus :-)

PR #17178, opened 2016-08-22
New filter: crc32

nerzhul:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

filter: add crc32 filtering
##### ANSIBLE VERSION

```
Commit ID 321d2e8ceea50bba386376c5f81e6ed8b60511cb
```
##### SUMMARY

This charge add a new useful filter crc32 which parse strings to generate crc32 output. This is useful with databases id or whenever you need to do a very lightweight comparison

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,10 @@ Ansible Changes By Release
   * All functionality from *_template network modules have been combined into *_config module
   * Network *_command modules not longer allow configuration mode statements
 
+
+####New Filters:
+* crc32
+
 ####New Modules:
 - apache2_mod_proxy
 - asa
--- a/lib/ansible/plugins/filter/core.py
+++ b/lib/ansible/plugins/filter/core.py
@@ -50,7 +50,7 @@ from ansible.compat.six import iteritems, string_types
 from ansible.compat.six.moves import reduce
 from ansible.module_utils._text import to_text
 from ansible.parsing.yaml.dumper import AnsibleDumper
-from ansible.utils.hashing import md5s, checksum_s
+from ansible.utils.hashing import md5s, checksum_s, crc32_s
 from ansible.utils.unicode import unicode_wrap
 from ansible.utils.vars import merge_hash
 from ansible.vars.hostvars import HostVars
@@ -470,9 +470,11 @@ class FilterModule(object):
             'quote': quote,
 
             # hash filters
+            # crc32 digest of string
+            'crc32': crc32_s,
             # md5 hex digest of string
             'md5': md5s,
-            # sha1 hex digeset of string
+            # sha1 hex digest of string
             'sha1': checksum_s,
             # checksum of string as used by ansible for checksuming files
             'checksum': checksum_s,
--- a/lib/ansible/utils/hashing.py
+++ b/lib/ansible/utils/hashing.py
@@ -20,6 +20,7 @@ from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
 import os
+import binascii
 
 # Note, sha1 is the only hash algorithm compatible with python2.4 and with
 # FIPS-140 mode (as of 11-2014)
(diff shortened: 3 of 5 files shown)


lib/ansible/utils/hashing.py
```diff
@@ -93,3 +93,6 @@ def md5(filename):
         raise ValueError('MD5 not available.  Possibly running in FIPS mode')
     return secure_hash(filename, _md5)
 
+from binascii import crc32 as crc32
```
alikins: Only need to import binascii module here and it should be near the top of the file.

alikins: 'import crc32 as crc32' is redundant and unneeded.

lib/ansible/utils/hashing.py
```diff
@@ -93,3 +93,6 @@ def md5(filename):
...
+def crc32(data):
```
alikins: the imported 'crc32' name and this method name collide.

filter/core.py expects this method to be called 'crc32_s' which doesn't exist.
I think you meant to name this crc32_s  (or better, crc32s to match existing md5s).

At top of file:

``` python
import binascii
```

(before the 'import os' since those are python std lib packages and 'binascii' is alphabetically first'

Then at end of module:

``` python
def crc32s(data):
     return binascii.crc32(data) & 0xffffffff
```

lib/ansible/plugins/filter/core.py
```diff
@@ -42,7 +42,7 @@
 
 from ansible import errors
 from ansible.parsing.yaml.dumper import AnsibleDumper
-from ansible.utils.hashing import md5s, checksum_s
+from ansible.utils.hashing import md5s, checksum_s, crc32_s
```
alikins: This will need to be match whatever is in hashing.py

lib/ansible/utils/hashing.py
alikins: Looking at hashing.py, I take back the comment about the naming convention, since there doesn't really seem to be one there. So 'crc32_s' is fine.

amenonsen:
No docs, no tests, and an unconvincing description of the use case. Isn't an IP address _already_ a unique integer?

lib/ansible/plugins/filter/core.py
nerzhul: Fixed

nerzhul:
Thanks for the review, i drafted it fast before leaving work to have a trace, but i will improve it a little bit.
I solved @alikins issues.

@amenonsen i should look at documentation issue, it's not clear in the link provided for contribution maybe could be improved on your side too :)

nerzhul:
I added a unit test. Now documentation is missing, but i don't know where document this, as there is no real index for filters in docs.ansible.com just a page with some filters missing

nerzhul:
Okay found, sorry :) thanks to grep. Documentation added , unit test added, no new issue, waiting for Shippable

nerzhul:
The freebsd test has a timeout, else everything is good :)

nerzhul:
@amenonsen @alikins i rebased the change to last devel version. Can someone do a review and tell me if the feature can be integrated ?

Thanks

CHANGELOG.md
```diff
@@ -62,6 +62,10 @@ Ansible Changes By Release
   * All functionality from *_template network modules have been combined into *_config module
   * Network *_command modules not longer allow configuration mode statements
 
+
+####New Filters:
```
gundalow, line 66: Space needed between `#` and title

nerzhul: wow a such old PR i did :D i need to rebase, remove changelog etc :)

gundalow: I assume this is something that would still add value

calfonso:
Hi!

This isn't something we feel is needed in the project due to the fact that crc32 is extremely outdated and not recommended for use.

Issue #15962, opened 2016-05-23
ansible galaxy command error: ERROR! Unexpected Exception: execv() arg 2 must contain only strings


ozbillwang:
##### ISSUE TYPE
- Bug Report
##### ANSIBLE VERSION

```
$ ansible --version
ansible 2.0.2.0
```
##### CONFIGURATION

[defaults]
inventory = ./inventory/hosts
host_key_checking = False
roles_path = roles.galaxy:roles
##### OS / ENVIRONMENT

Mac OS X EI Capitan
version 10.11.14
##### SUMMARY

get error when run ansible-galaxy command with version number `1.0`, but fine with commit number `65c8a4e35f7e59ff250c5cd5526ff752a2905cac`

```
ERROR! Unexpected Exception: execv() arg 2 must contain only strings
```
##### STEPS TO REPRODUCE

```
$ cat requirements.yml
# logrotate role
- src: https://github.com/retr0h/ansible-logrotate
  scm: git
  version: 1.0
  name: logrotate

$ ansible-galaxy install -r requirements.yml
ERROR! Unexpected Exception: execv() arg 2 must contain only strings
```

But I replace with commit number:

```
# logrotate role
- src: https://github.com/retr0h/ansible-logrotate
  scm: git
  version: 65c8a4e35f7e59ff250c5cd5526ff752a2905cac
  name: logrotate
```

The command runs fine.
##### EXPECTED RESULTS

download this galaxy role from github.
##### ACTUAL RESULTS

```
ERROR! Unexpected Exception: execv() arg 2 must contain only strings
```

bcoca:
I run same with 2.1 file and get:

```
- extracting logrotate to /etc/ansible/roles/logrotate
- logrotate was installed successfully

```

So I'm going to close this as fixed

jswinarton:
I'm also having the same issue, on a later version of Ansible:

```
$ ansible-galaxy install -r provisioning/requirements.yml -vvv
[...]
found role {'scm': 'git', 'src': 'https://github.com/jswinarton/hackbox-user', 'version': 0.2, 'name': 'hackbox-user'} in yaml file
Installing role hackbox-user
ERROR! Unexpected Exception: execv() arg 2 must contain only strings
the full traceback was:

Traceback (most recent call last):
  File "/usr/local/Cellar/ansible/2.1.1.0/libexec/bin/ansible-galaxy", line 92, in <module>
    exit_code = cli.run()
[...]

PR #17182, opened 2016-08-22
Add documentation on YAML floating-point values

jswinarton:
##### ISSUE TYPE
- Docs Pull Request
##### COMPONENT NAME
- docsite/rst/YAMLSyntax.rst
- docsite/rst/galaxy.rst
##### SUMMARY

When specifying ansible roles to be downloaded by ansible-galaxy in a `requirements.yml` file, you can specify a tag, if the ansible role is a git repository:

```
- src: git@gitlab.company.com:mygroup/ansible-base.git
   scm: git
   version: 0.1.0
```

Certain values of `version` (e.g. `1.0`) get parsed into floating-point values by YAML, which causes an error. To help people diagnose this issue, I have added some documentation to the YAML gotchas section and to the galaxy documentation itself.

### DIFF
--- a/docsite/rst/YAMLSyntax.rst
+++ b/docsite/rst/YAMLSyntax.rst
@@ -143,6 +143,14 @@ In these cases just use quotes::
     other_string: "False"
 
 
+YAML converts certain strings into floating-point values, such as the string
+`1.0`. If you need to specify a version number (in a requirements.yml file, for
+example), you will need to quote the value if it looks like a floating-point
+value::
+
+  version: "1.0"
+
+
 .. seealso::
 
    :doc:`playbooks`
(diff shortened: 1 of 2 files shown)


dharmabumstead:
Thanks @jswinarton!

scottsb:
Should this be considered a bug and fixed instead of documented?

bcoca:
@scottsb I don't see how this is a bug, it adheres to the YAML spec, so pyyaml is working as designed and documented

scottsb:
I was thinking of changing it at the point of the Git call, not the YAML parsing. But had an exchange in #15962 on why that's not a good solution either and now agree.

Never mind. 😄

PR #17186, opened 2016-08-22
Fix errors when using -vvvv with python 3

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

setup
##### SUMMARY

```
  File "/tmp/ansible_tpehdgt7/ansible_module_setup.py", line 134, in <module>
    main()
  File "/tmp/ansible_tpehdgt7/ansible_module_setup.py", line 124, in main
    supports_check_mode = True,
  File "/tmp/ansible_tpehdgt7/ansible_modlib.zip/ansible/module_utils/basic.py", line 696, in __init__
  File "/tmp/ansible_tpehdgt7/ansible_modlib.zip/ansible/module_utils/basic.py", line 1670, in _log_invocation
  File "/tmp/ansible_tpehdgt7/ansible_modlib.zip/ansible/module_utils/basic.py", line 469, in heuristic_log_sanitize
TypeError: 'str' does not support the buffer interface
```

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -457,6 +457,8 @@ def heuristic_log_sanitize(data, no_log_values=None):
     # prev_begin: where in the overall string to start a search for
     #   a passwd
     # sep_search_end: where in the string to end a search for the sep
+    data = to_native(data)
+
     output = []
     begin = len(data)
     prev_begin = begin


abadger:
Thanks.  Two issues with this.  We're aiming to do "native strings" for things in module_utils.  That means that on python2 we'll use bytes (or either bytes or unicode if it maks sense) as the input and the output from functions and on python3 we'll use text.  Internal to any given function might require text or bytes depending on whether we have to deal with other public data.  By using data.decode("utf-8", "replace") all the time at the top I think we'll end up returning text on python2 instead of 

The other is that we should use ansible.module_utils.six instead of the hacky bytes and unicode types that we added to basic.py at first.  So the isinstance should be checking binary_type from there.

I think that both of those can actually be fixed by using 

``` python
data = to_native(data)
```

mscherer:
Ok, will take a look.

abadger:
Thanks @misc!

PR #17188, opened 2016-08-22
Update module_utils/asa.py with NetworkModule

Qalthos:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

module_utils/asa.py
##### ANSIBLE VERSION

```
2.2
```
##### SUMMARY

Going along with EOS, IOS, et al., this module follows the same general pattern, so has been updated accordingly.

### DIFF
--- a/lib/ansible/module_utils/asa.py
+++ b/lib/ansible/module_utils/asa.py
@@ -19,176 +19,85 @@
 #
 
 import re
-from ansible.module_utils.basic import AnsibleModule, env_fallback, get_exception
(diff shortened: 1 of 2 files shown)


privateip:
looks good.  one minor update, remove abort_config and commit_config methods.  Those were removed from netcfg.Config and are implemented on a device by device basis.  Should be good to merge after that

Qalthos:
I was also thinking of moving all the `raise NotImplementedError` into CliBase so that it wouldn't have to be specified if missing (at least for Cli-based implementations)

PR #17204, opened 2016-08-23
Fix "Text file busy" exception in atomic_move (#9526)

### LINKED ISSUE #9526: file operations fail on shared/synced folders that do not implement correct filesystem semantics

stockalexander:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

module_utils
##### ANSIBLE VERSION

```
ansible 2.2.0 (fix_textfilebusy f659e54a35) last updated 2016/08/23 22:59:01 (GMT +000)
  lib/ansible/modules/core: (detached HEAD 368ca738fa) last updated 2016/08/23 21:57:28 (GMT +000)
  lib/ansible/modules/extras: (detached HEAD 0749ce6faa) last updated 2016/08/23 21:57:36 (GMT +000)
```
##### SUMMARY

tempfile.NamedTemporaryFile keeps a file handle causing os.rename() to fail with windows based vboxfs: [Errno 26] Text file busy. Changed NamedTemporaryFile to mkstemp() and added a finally block to unlink the temp file in each and every case.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -1876,55 +1876,59 @@ class AnsibleModule(object):
                 dest_dir = os.path.dirname(dest)
                 dest_file = os.path.basename(dest)
                 try:
-                    tmp_dest = tempfile.NamedTemporaryFile(
+                    tmp_dest_fd, tmp_dest_name = tempfile.mkstemp(
                         prefix=".ansible_tmp", dir=dest_dir, suffix=dest_file)
                 except (OSError, IOError):
                     e = get_exception()
                     self.fail_json(msg='The destination directory (%s) is not writable by the current user. Error was: %s' % (dest_dir, e))
 
-                try: # leaves tmp file behind when sudo and  not root
-                    if switched_user and os.getuid() != 0:
-                        # cleanup will happen by 'rm' of tempdir
-                        # copy2 will preserve some metadata
-                        shutil.copy2(src, tmp_dest.name)
-                    else:
-                        shutil.move(src, tmp_dest.name)
-                    if self.selinux_enabled():
-                        self.set_context_if_different(
-                            tmp_dest.name, context, False)
+                try:
                     try:
-                        tmp_stat = os.stat(tmp_dest.name)
-                        if dest_stat and (tmp_stat.st_uid != dest_stat.st_uid or tmp_stat.st_gid != dest_stat.st_gid):
-                            os.chown(tmp_dest.name, dest_stat.st_uid, dest_stat.st_gid)
-                    except OSError:
-                        e = get_exception()
-                        if e.errno != errno.EPERM:
-                            raise
-                    os.rename(tmp_dest.name, dest)
-                except (shutil.Error, OSError, IOError):
-                    e = get_exception()
-                    # sadly there are some situations where we cannot ensure atomicity, but only if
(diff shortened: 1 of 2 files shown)


stockalexander:
Added a simple vagrant/ansible example to reproduce the issue (tested with windows 10 host and virtualbox 5.1.2): [textfilebusy.tar.gz](https://github.com/ansible/ansible/files/433766/textfilebusy.tar.gz)

abadger:
I've reviewed this and think that any problems that might exist here also existed in the former code.  So I'm merging into devel.  For backporting to stable-2.1, we're already in the release candidate period for the 2.1.2 release and bcoca would like to review the code to see if holding open the filehandle was a feature (perhaps keeping the data from being removed by an external process).  So I'm not merging to stable-2.1 yet but I've opened a ticket: https://github.com/ansible/ansible/issues/17260  to evaluate backporting once 2.1.2 is released.

PR #17211, opened 2016-08-24
Workaround for Ubuntu Python3 looking for modules in wrong places

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

module_common.py
##### ANSIBLE VERSION

```
devel
```

### DIFF
--- a/lib/ansible/executor/module_common.py
+++ b/lib/ansible/executor/module_common.py
@@ -92,7 +92,28 @@ ANSIBALLZ_WRAPPER = True # For test-module script to tell this is a ANSIBALLZ_WR
 # LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 # USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 import os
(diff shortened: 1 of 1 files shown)


lib/ansible/executor/module_common.py
```diff
@@ -92,7 +92,22 @@
 # LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 # USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 import os
+import os.path
 import sys
+import __main__
+
+# With Python3 on Ubuntu, we pick up this script in the temporary directory.
```
gundalow: All Ubuntu versions or only certain ones? That will make it clearer if someone needs to update this in the future to know what needs testing

gundalow:
minor docs suggestion

lib/ansible/executor/module_common.py
abadger: We don't know.  We do know that it happens with both Ubuntu15 (python-3.4) and Ubuntu16 (python-3.5) and that it does not happen with Fedora 23 (python-3.4).

PR #17225, opened 2016-08-24
The former match for a section header included hosts that began with a range

### LINKED ISSUE #15331: Hostnames can't start with a range in inventory as of 2.0.1.0
##### ISSUE TYPE

- Bug Report
##### ANSIBLE VERSION

```
ansible 2.0.1.0
```
##### CONFIGURATION

```
% cat ansible.cfg
[defaults]
hostfile = inventory

% cat inventory
[test]
[1:2].fail
fail.[1:2]
fail.[1:2].fail
```
##### OS / ENVIRONMENT

N/A
##### SUMMARY

The fix for this issue breaks hosts that start with a range:
https://github.com/ansible/ansible/issues/13917
This commit:
https://github.com/ansible/ansible/commit/a7dd425620e0a40911908a41e1b378183ae690f4
##### STEPS TO REPRODUCE

setup config and inventory as shown above, then run:

```
ansible test --list-hosts

```

##### EXPECTED RESULTS

```
ansible test --list-hosts
  hosts (6):
    1.fail
    2.fail
    fail.1
    fail.2
    fail.1.fail
    fail.2.fail
```
##### ACTUAL RESULTS

```
ansible test --list-hosts
ERROR! inventory:2: Invalid section entry: '[1:2].fail'. Please make sure that there are no spacesin the section entry, and that there are no other invalid characters

```

Changing line 127 in ansible/inventory/ini.py to the following might fix both issues, since ranges have to have something either before or after or both for them to be useful. In other words, if it starts and ends with square brackets and doesn't match the section regex, then it is an actual error.

```
            elif line.startswith('[') and line.endswith(']'):
```

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

 lib/ansible/inventory/ini.py
##### ANSIBLE VERSION

```

```
##### SUMMARY

Checking that the line ends with "]" narrows that window somewhat,

Fixes #15331

### DIFF
--- a/lib/ansible/inventory/ini.py
+++ b/lib/ansible/inventory/ini.py
@@ -124,7 +124,7 @@ class InventoryParser(object):
                     del pending_declarations[groupname]
 
                 continue
-            elif line.startswith('['):
+            elif line.startswith('[') and line.endswith(']'):
                 self._raise_error("Invalid section entry: '%s'. Please make sure that there are no spaces" % line + \
                                   "in the section entry, and that there are no other invalid characters")
 


lib/ansible/inventory/ini.py
```diff
@@ -124,7 +124,7 @@ def _parse(self, lines):
                     del pending_declarations[groupname]
 
                 continue
-            elif line.startswith('['):
+            elif line.startswith('[') and line.endswith(']'):
```
bcoca, line 127: ~~strip line before matching?~~ nvmd, already in code higher up

PR #17227, opened 2016-08-24
Add OpenBSD virtualization facts.

eest:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

facts
##### ANSIBLE VERSION

```
ansible 2.2.0 (openbsd_vm_facts df66dc4acc) last updated 2016/08/24 18:15:04 (GMT +200)
  lib/ansible/modules/core: (detached HEAD 2f26352e49) last updated 2016/08/24 18:02:24 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD 65eba72ee6) last updated 2016/08/24 18:02:24 (GMT +200)
  config file = 
```
##### SUMMARY

Add OpenBSD virtualization facts.

Before:

```
$ ansible -i hosts -u root 192.168.1.60 -m setup | grep ansible_virtual 
        "ansible_virtualization_role": "", 
        "ansible_virtualization_type": "", 

```

After:

```
$ ansible -i hosts -u root 192.168.1.60 -m setup | grep ansible_virtual
        "ansible_virtualization_role": "guest", 
        "ansible_virtualization_type": "virtualbox", 
```

Patch written by @jasperla.

Tested by various people on:
- virtualbox
- vmware esx(i) + fusion
- kvm (smartos + plain linux + a random cloud provider)

This patch is already present in the OpenBSD port of ansible.

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -3169,7 +3169,9 @@ class OpenBSDVirtual(Virtual):
         return self.facts
 
     def get_virtual_facts(self):
-        rc, out, err = self.module.run_command("/usr/sbin/sysctl -n hw.product")
+        sysctl_path = self.module.get_bin_path('sysctl')
+
+        rc, out, err = self.module.run_command("%s -n hw.product" % sysctl_path)
         if rc != 0:
             self.facts['virtualization_type'] = ''
             self.facts['virtualization_role'] = ''
@@ -3193,7 +3195,7 @@ class OpenBSDVirtual(Virtual):
             self.facts['virtualization_role'] = 'guest'
         else:
             # Try harder and see if hw.vendor has anything we could use.
-            rc, out, err = self.module.run_command("/usr/sbin/sysctl -n hw.vendor")
+            rc, out, err = self.module.run_command("%s -n hw.vendor" % sysctl_path)
             if rc != 0:
                 self.facts['virtualization_type'] = ''
                 self.facts['virtualization_role'] = ''
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -3169,43 +3169,48 @@ class OpenBSDVirtual(Virtual):
         return self.facts
 
     def get_virtual_facts(self):
-        sysctl_path = self.module.get_bin_path('sysctl', required=True)
+        sysctl_path = self.module.get_bin_path('sysctl')
 
-        rc, out, err = self.module.run_command("%s -n hw.product" % sysctl_path)
-        if rc != 0:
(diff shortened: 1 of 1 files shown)


lib/ansible/module_utils/facts.py
```diff
@@ -3169,8 +3169,47 @@ def populate(self):
...
+          return
+        elif out.rstrip() == 'Parallels':
+          self.facts['virtualization_type'] = 'parallels'
+          self.facts['virtualization_role'] = 'guest'
+          return
+        elif out.rstrip() == 'RHEV Hypervisor':
+          self.facts['virtualization_type'] = 'RHEV'
+          self.facts['virtualization_role'] = 'guest'
+          return
+
```
bcoca: use an else here and you can have only 1 return at the end (even that is not needed)

lib/ansible/module_utils/facts.py
```diff
@@ -3169,8 +3169,42 @@ def populate(self):
         return self.facts
 
     def get_virtual_facts(self):
-        self.facts['virtualization_type'] = ''
-        self.facts['virtualization_role'] = ''
+        rc, out, err = self.module.run_command("/usr/sbin/sysctl -n hw.product")
```
bcoca: I would use module.get_bin_path function instead of hardcoding path (future proof), same below

eest:
@bcoca: Updated.

lib/ansible/module_utils/facts.py
```diff
@@ -3169,8 +3169,44 @@ def populate(self):
         return self.facts
 
     def get_virtual_facts(self):
-        self.facts['virtualization_type'] = ''
-        self.facts['virtualization_role'] = ''
+        sysctl_path = self.module.get_bin_path('sysctl', required=True)
```
bcoca: This fails the whole fact gathering if missing, i would just add an `if sysctl_path:` instead

eest: @bcoca: I tested what would happen if i removed the sysctl binary on a testbox, and noticed the setup-module will already fail, because there are some places where the tool path is hardcoded and this leads to the module erroring out. This means my required=True addition does not make anything worse than it already is.

If the real solution is to also replace other instances of hardcoded sysctl paths (which would have to be a separate PR imo), then my question is what the desired behaviour is when the tool is not found: setting the effected fact variables to the empty string, or not setting them at all.

bcoca: It's on my list to fix those, not asking you to do that, just this case so i have less work later (im selfish like that!).

I would say an empty string would be less error prone for uses so they don't have to check if the variable is defined.

eest:
@bcoca: How about now? :)

PR #17234, opened 2016-08-25
Get the ssh plugin working with python3

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

lib/ansible/plugins/connection/ssh.py
##### ANSIBLE VERSION

```
devel
```
##### SUMMARY

This is enough to get ssh connection plugin working to run the ping module.  Code is probably doing too much work (transforming to bytes multiple times) and probably leaves things out in other code paths (fetch, for example, is definitely not tested).  Making sure that it doesn't break any python2 tests and possibly merging this so people can play around with it -- I'm going to be spreading my attention around to a few other connection plugins and some modules so that we can get the connection integration tests running.  If someone else picks up this code and improves it, that would be stellar. :-)

### DIFF
--- a/lib/ansible/plugins/connection/__init__.py
+++ b/lib/ansible/plugins/connection/__init__.py
@@ -240,27 +240,29 @@ class ConnectionBase(with_metaclass(ABCMeta, object)):
         """Terminate the connection"""
         pass
 
-    def check_become_success(self, output):
-        for line in output.splitlines(True):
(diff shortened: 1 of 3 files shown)


abadger:
Merged to devel.  PR's to clean up the code welcomed or I'll try to get to it after I get minimal functionality in some of the other, higher priority areas.

Issue #17205, opened 2016-08-23
Jinja2 math filter: implement `add()` function


muxator:
##### ISSUE TYPE
- Feature Idea
##### COMPONENT NAME

Jinja2 math filter: add `increment()` function
##### ANSIBLE VERSION

```
ansible 2.1.1.0
```
##### CONFIGURATION

independent on configuration
##### OS / ENVIRONMENT

N/A
##### SUMMARY

There are situations in which, given a variable containing a list of ints:

``` yaml
list_of_ints: [ 1, 7, 8 ]
```

Someone may want to have the same list, but incremented by a fixed quantity, let's say 100:

``` yaml
incremented_list: [ 101, 107, 108 ]
```

Imagine for example having to run different instances of the same service on different ports, and having to pass a subset of the port list as a parameter to the service. In this case you really need to keep incremented_list as a separate (but correlated) list.

My proposal is enabling this:

``` yaml
# generates [ 101, 107, 108 ]
incremented_list: "{{ list_of_ints | map('increment', 100) | list }}"
```
##### STEPS TO REPRODUCE

The proposal is adding an `add()` function in https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/filter/mathstuff.py and write the relevant documentation in order to enable this use case.

If you see some utility in this functionality I can prepare a patch for you to review.

nitzmahone:
Sure, why not? I'd probably just call the filter `add` (might as well do `subtract` `multiply` and `divide` for completeness sake)- go ahead and file a PR, should be nice and simple...

muxator:
Hi, I have created PR #17251.

The implemented functions are:
- `add`
- `subtract`
- `multiply`
- `divide`
- `modulo`

I have also updated the documentation, trying to mention that these filters won't be available in older Ansible versions.

muxator:
In the spirit of this change, I am also proposing #17267 (`string_prefix` and `string_suffix`).

That functionality can be achieved by advanced users using `regex_replace` anyway, but maybe it makes sense to use simple functions for simple use cases.

bcoca:
Hi!

Thanks very much for your submission to Ansible.  It sincerely means a lot to us that you've taken time to contribute.

Unfortunately, we are going to decline this because you can already use operators to achieve the same thing (+ , - , / ...)

However, we're absolutely always up for discussion.  
Since this is a really busy project, we don't always see comments on closed tickets, but want to encourage open dialog.  

[...]

PR #17251, opened 2016-08-25
Implement Jinja2 filters for basic arithmetic (add, subtract, multiply, divide, modulo)

muxator:
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

Jinja2 math filter
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel 35a3653dfc) last updated 2016/08/25 23:10:15 (GMT +200)
```
##### SUMMARY

Implement `add`, `subtract`, `multiply`, `divide` and `modulo` operations as Jinja filters and updated the relevant documentation.

The original feature request for this PR is #17205

### DIFF
--- a/docsite/rst/playbooks_filters.rst
+++ b/docsite/rst/playbooks_filters.rst
@@ -212,31 +212,31 @@ Square root, or the 5th::
 
 Note that jinja2 already provides some like abs() and round().
 
-Since version 2.2 Ansible also includes some basic arithmetic filters. They are useful only in advanced cases (like a map() call). Normally, you can use the standard arithmetic operators (+, -, *, /, %).
+Starting from version 2.2 Ansible also includes some basic arithmetic filters. They are useful only in advanced cases (like a map() call). Normally, you can use the standard arithmetic operators (+, -, *, /, %).
 
 Add a number::
 
-    # only useful in a map() call, otehrwise use '+'
+    # only useful in a map() call, otherwise use '+'
     {{ list_of_numbers | map('add', 2) | list }}
 
 Subtract a number::
 
-    # only useful in a map() call, otehrwise use '-'
+    # only useful in a map() call, otherwise use '-'
     {{ list_of_numbers | map('subtract', 2) | list }}
 
 Multiply by a number::
 
-    # only useful in a map() call, otehrwise use '*'
-    {{ list_of_numbers | map('multiply' ,2) | list }}
+    # only useful in a map() call, otherwise use '*'
+    {{ list_of_numbers | map('multiply', 2) | list }}
 
 Divide by a number::
 
-    # only useful in a map() call, otehrwise use '/'
+    # only useful in a map() call, otherwise use '/'
     {{ list_of_numbers | map('divide', 2) | list }}
 
 Modulo operation::
 
-    # only useful in a map() call, otehrwise use '%'
+    # only useful in a map() call, otherwise use '%'
     {{ list_of_numbers | map('modulo', 7) | list }}
 
 .. _ipaddr_filter:
--- a/docsite/rst/playbooks_filters.rst
+++ b/docsite/rst/playbooks_filters.rst
@@ -217,27 +217,27 @@ Since version 2.2 Ansible also includes some basic arithmetic filters. They are 
 Add a number::
 
     # only useful in a map() call, otehrwise use '+'
-    {{ myvar | map('add', 2) }}
+    {{ mylist | map('add', 2) | list }}
 
 Subtract a number::
 
     # only useful in a map() call, otehrwise use '-'
-    {{ myvar | map('subtract', 2) }}
+    {{ mylist | map('subtract', 2) | list }}
 
 Multiply by a number::
 
     # only useful in a map() call, otehrwise use '*'
-    {{ myvar | map('multiply' ,2) }}
+    {{ mylist | map('multiply' ,2) | list }}
 
 Divide by a number::
 
     # only useful in a map() call, otehrwise use '/'
-    {{ myvar | map('divide', 2) }}
+    {{ mylist | map('divide', 2) | list }}
 
 Modulo operation::
 
     # only useful in a map() call, otehrwise use '%'
-    {{ myvar | map('modulo', 7) }}
+    {{ mylist | map('modulo', 7) | list }}
 
 .. _ipaddr_filter:
 
--- a/docsite/rst/playbooks_filters.rst
+++ b/docsite/rst/playbooks_filters.rst
@@ -210,6 +210,26 @@ Square root, or the 5th::
     {{ myvar | root }}
     {{ myvar | root(5) }}
 
+Add a number (since version 2.X)::
+
+    {{ myvar | add(2) }}
+
+Subtract a number (since version 2.X)::
+
+    {{ myvar | subtract(2) }}
+
+Multiply by a number (since version 2.X)::
+
+    {{ myvar | multiply(2) }}
+
+Divide by a number (since version 2.X)::
+
+    {{ myvar | divide(2) }}
+
+Modulo operation (from version 2.X)::
+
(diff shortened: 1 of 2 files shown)


sivel:
I'd like to see docs examples of using this in real world examples.  Jinja2 supports standard math operations already, and the examples only duplicate that.

Based on the example in the initial issue, they would be useful really only in a more complex example.

Otherwise it's just the same as:

```
{{ var * 2 }}
```

georgepsarakis:
Nice addition to the standard filters. One suggestion though if you don't mind, wouldn't using the `operator` built-in library be simpler than redefining these simple functions?

**Example**

``` python
import operator

 def filters(self):
         return {
             ...
              'add' : operator.add,
              'multiply' : operator.mul,
             ...
```

muxator:
@sivel: you are right, the standard math operators are totally fine as long as you just need to work on simple values:

``` yaml
var: 1
incremented_var: "{{ var + 9200 }}"
```

In my use case, though, I wanted to generalize this on a list:

``` yaml
list_of_ints:
  - 1
  - 2
```

The natural way of doing this would be a `map()` filter. But `*` is not a function you can actually pass to `map()`:

``` yaml
# this is not possible in Ansible 2.1 (no support in Jinja)
list_of_incremented_ints: "{{ list_of_ints | map('*', 2)        | list }}" 

# proposed solution
list_of_incremented_ints: "{{ list_of_ints | map('multiply', 2) | list }}"
```

You can call it a sort of "functional" approach to writing playbooks.

The broader context of my effort was:
- set independent variables (for example, at inventory level) that characterize the dimensions of my system. These variables define the environment I am going to deploy in (e.g., how many instances of a service I want to run on a physical box, how many boxes do I have)
- I can write a different inventory for each different environment (e.g.: just 2 instances in QA, but 4 in production)
- based on those parameters, compute dependent variables (for example, at group level) that are needed to perform the actual operation (e.g.: instance _i_ will bind on port _9200 + i_)
- **limit task usage to operations that actually perform side effects on the system** (e.g., given that set of variables, actually launch the correct number of instances with the consequent configuration)

Maybe my approach is too complicated, or not idiomatic. Trying to do this was quite natural, but I found no direct way to do it.

Anyways, together with the folks at stack overflow, we came up with a [solution](http://stackoverflow.com/questions/39111787/ansible-given-a-list-of-ints-in-a-variable-define-a-second-list-in-which-each) that uses a task for computing that list.

``` yaml
# this solution works, but uses a task just for computing a variable
tasks:
  - set_fact:
      list_of_ints_doubled: "{{ list_of_ints_doubled | default([]) + [ item * 2 ]  }}"
    with_items: "{{ varlist }}"
```

Let me know your thoughts on this.

sivel:
@muxator I agree, my comment was that your current examples that you added to the documentation are lacking.  What I am looking for is your explanation about use with `map`.

A user looking at the current examples may be lead to think that is the proper way to just do math, whereas the real reason they were added is for your `map` scenario.

muxator:
I agree on this, @sivel: it was difficult for me to decide where to describe this use case in the documentation. Probably the [math section](http://docs.ansible.com/ansible/playbooks_filters.html#math) is not its only place.

Maybe whe can try this:
- continue documenting the existence of the filters in the "math" section
- add a note specifiyng that these functions are really useful only for advanced use cases
- add a section at the end of http://docs.ansible.com/ansible/playbooks_filters.html that describes the use case. It might be "transforming lists", or "advanced uses", or "functional approach".

I am no expert with RST, though, so any help is appreciated.

muxator:
I modified the documentation for this pull request.
This is the current form:

> Starting from version 2.2 Ansible also includes some basic arithmetic filters. They are useful only in advanced cases (like a map() call). Normally, you can use the standard arithmetic operators (+, -, *, /, %).
> 
> Add a number:
> 
> ``` yaml
> # only useful in a map() call, otherwise use '+'
> {{ list_of_numbers | map('add', 2) | list }}
> ```
> 
> [...]

muxator:
@sivel, @nitzmahone, is it ok to merge this, or do you feel I can do something to improve it?

sivel:
I do like the idea proposed here to use the built in methods from `operator`, which would likely reduce the amount of code required for this functionality.

muxator:
It can be done for sure.

If we get rid of the wrapping function, anyway, we will not explicitly handle TypeError, and the user will get a slightly different error message compared to the one that he will receive on `inversepower()`, or `logarithm()`.

Now that I look better at the code, `min()` or `max()`, are wrapped in a function but with no `try..catch`, so in this case we are a bit inconsistent even now.

What do you think it is better to do to [astonish our users as little as possible](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)? I can change my code or wrap the other functions, anything is ok.

muxator:
Hi @sivel, @nitzmahone.

Just to have some data, this is an example of failure for the code in this PR:

``` bash
# code proposed in this PR
fatal: [localhost]: FAILED! => {"failed": true, "msg": "add() can only be used on numbers: cannot concatenate 'str' and 'int' objects"}
```

If we get rid of the wrapper function, the failure will become this:

``` bash
# shorter code, as per @georgepsarakis proposal
fatal: [localhost]: FAILED! => {"failed": true, "msg": "Unexpected templating type error occurred on ({{ 1 | addwithoperator('qw') }}): unsupported operand type(s) for +: 'int' and 'str'"}
```

By the way, the current way `log()` and other functions currently fail is this (similar to the proposed code):

``` bash
fatal: [localhost]: FAILED! => {"failed": true, "msg": "log() can only be used on numbers: a float is required"}
```

Let me know what is your preference, I am ok with anything.

georgepsarakis:
@muxator I think there is a fundamental difference between the `log` and the `add` functions.
The error message that you are suggesting is not really accurate for all operations.

The `+` operator (and `operator.add`) will work with other object types as well and not strictly with `int` or `float` instances:
- `[1,2] + [3]`
- `'808' + '2'`

The same applies for multiplication too:
- `'1' * 4`
- `[1] * 4`

If you wanted the filters to work strictly with numbers then a strict type checking should probably be applied on the arguments before adding:

``` python
def _is_number(x):
    return isinstance(x, (int, long, float))

def add(a, b):
    # The following check could be applied easily
    # with a decorator to multiple functions.
    if _is_number(a) and _is_number(b):
        return a + b
    else:
        raise Exception
```

However I do not see the reason to reduce the built-in functionality.

muxator:
Hi @georgepsarakis, thanks for your comment.

The use case I would like to enable is doing simple `map()` transformations on list of numbers (#17205) and list of strings (#17267).
In this context, I really would like to incorporate the code you proposed and be strict with types (and would _gladly_ accept some help with using decorators...)

Do you think it makes sense?

georgepsarakis:
@muxator I do not think that strict type checking is a very good idea here. As I mentioned before,
`add` can have a much wider range of use cases without restricting it to numbers only.

After some thought, I came up with the following proposal, which uses the built-in `operator` functions and produces a custom error message for `TypeError` cases:

``` python
from functools import partial
import operator

def _binary_operation(operator_function, a, b):
    try:
        return operator_function(a, b)
    except TypeError as e:
        message = '%s %s' % (operator_function.__doc__, str(e))
        raise errors.AnsibleFilterError(message)

def div(a, b):
    '''div(a, b) -- Same as a / b.'''
    return operator.truediv(a, b)

class FilterModule(object):
     ''' Ansible math jinja2 filters '''

     def filters(self):
         return {
             # ...
             'add': partial(_binary_operation, operator.add),
             'sub': partial(_binary_operation, operator.sub),
             'mul': partial(_binary_operation, operator.mul),
             'div': partial(_binary_operation, div),
             'mod': partial(_binary_operation, operator.mod),
             # ...
         }
```

Looking forward to hearing your thoughts on this :smiley: .

dharmabumstead:
Closing this due to inactivity.

PR #17252, opened 2016-08-26
Correctly infer task results

### LINKED ISSUE #17208: wrong handler is called when squashed items fail
##### ISSUE TYPE
- Bug Report
##### COMPONENT NAME

handler
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel a9322e8b8b) last updated 2016/08/24 14:46:19 (GMT +1000)
  lib/ansible/modules/core: (detached HEAD 368ca738fa) last updated 2016/08/24 15:12:35 (GMT +1000)
  lib/ansible/modules/extras: (detached HEAD 0749ce6faa) last updated 2016/08/24 15:12:35 (GMT +1000)
  config file = 
```
##### CONFIGURATION

N/A
##### OS / ENVIRONMENT

N/A
##### SUMMARY

Results of modules with squashed items appear to be misinterpreted.
##### STEPS TO REPRODUCE

```
- hosts: localhost
  connection: local
  become: true

  tasks:
  - name: install stuff
    yum:
      name: "{{ item }}"
      state: present
    with_items:
    - bash
    - does-not-exist
```

Note that if you swap bash and does-not-exist around, it works as expected. Basically, because results array is populated by each successful installation, the behaviour is different when it fails on the first item vs the second item.

This is similar to #16766 in that the result of is_failed is incorrect, but for a different reason. I raise this as an issue rather than a PR as I'm not sure what the results should be.

One might argue that this should be an against yum rather than core - but the results of the yum task look reasonable in the absence of documentation that explicitly sets out conditions for module failues, and it seems to be the parsing of the results from yum that is at fault rather than yum module itself.
##### EXPECTED RESULTS

failure message is displayed in output as when does-not-exist is first item in list:

```
$ ansible-playbook yum_fail.yml -K -vv
Using /home/wthames/tmp/ansible/ansible.cfg as config file
SUDO password: 

PLAYBOOK: yum_fail.yml *********************************************************
1 plays in yum_fail.yml

Wednesday 24 August 2016  15:22:02 +1000 (0:00:00.030)       0:00:00.030 ****** 
ok: [localhost]

TASK [install stuff] ***********************************************************
task path: /home/wthames/tmp/ansible/yum_fail.yml:8
Wednesday 24 August 2016  15:22:05 +1000 (0:00:03.175)       0:00:03.205 ****** 
failed: [localhost] (item=[u'does-not-exist', u'bash']) => {"changed": false, "failed": true, "item": ["does-not-exist", "bash"], "msg": "No Package matching 'does-not-exist' found available, installed or updated", "rc": 0, "results": []}
    to retry, use: --limit @yum_fail.retry

Wednesday 24 August 2016  15:22:06 +1000 (0:00:00.910)       0:00:04.116 ****** 
=============================================================================== 
setup ------------------------------------------------------------------- 3.18s
 ------------------------------------------------------------------------------
install stuff ----------------------------------------------------------- 0.91s
/home/wthames/tmp/ansible/yum_fail.yml:8 --------------------------------------
Playbook run took 0 days, 0 hours, 0 minutes, 4 seconds
```
##### ACTUAL RESULTS

ok message is displayed in output.

```
ansible-playbook yum_fail.yml -K -vv
Using /home/wthames/tmp/ansible/ansible.cfg as config file
SUDO password: 

PLAYBOOK: yum_fail.yml *********************************************************
1 plays in yum_fail.yml

Wednesday 24 August 2016  15:25:05 +1000 (0:00:00.033)       0:00:00.033 ****** 
ok: [localhost]

TASK [install stuff] ***********************************************************
task path: /home/wthames/tmp/ansible/yum_fail.yml:6
Wednesday 24 August 2016  15:25:08 +1000 (0:00:03.253)       0:00:03.287 ****** 
ok: [localhost] => (item=[u'bash', u'does-not-exist']) => {"changed": false, "failed": true, "item": ["bash", "does-not-exist"], "msg": "No Package matching 'does-not-exist' found available, installed or updated", "rc": 0, "results": ["bash-4.3.42-3.fc23.x86_64 providing bash is already installed"]}
    to retry, use: --limit @yum_fail.retry

Wednesday 24 August 2016  15:25:09 +1000 (0:00:00.935)       0:00:04.222 ****** 
=============================================================================== 
setup ------------------------------------------------------------------- 3.25s
 ------------------------------------------------------------------------------
install stuff ----------------------------------------------------------- 0.94s
/home/wthames/tmp/ansible/yum_fail.yml:6 --------------------------------------
Playbook run took 0 days, 0 hours, 0 minutes, 4 seconds
```

willthames:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

task_result
##### ANSIBLE VERSION

```

```
##### SUMMARY

The results field in a task result is not necessarily a list of dicts containing `failed` and `changed` keys, but can just be a list of strings (e.g. the results from the yum module). In that case, the results list is not useful for calculating `changed` or `failed`, at which point `_check_key` should fail back to the `changed` or `failed` key in the task result itself 

e.g. the result from yum looks a little like:

```
self._result = { 'failed': true,
                 'results': ['a already installed',
                             'b already installed',
                             'could not install c'] }
```

and the result should be the `failed` key, and not the fact that none of the results contained a `failed` key.

The side effect prior to this change is that tasks are displayed as if they succeed, and then the playbook fails (i.e. the runner_v2_on_ok handler is called, and then the playbook stops)

Fixes #17208

### DIFF
--- a/lib/ansible/executor/task_result.py
+++ b/lib/ansible/executor/task_result.py
@@ -62,11 +62,8 @@ class TaskResult:
         return self._check_key('unreachable')
 
     def _check_key(self, key):
-        if self._result.get('results', []):
+        if 'results' in self._result:
             flag = False
-            for res in self._result.get('results', []):
-                if isinstance(res, dict):
-                    flag |= res.get(key, False)
-            return flag
-        else:
-            return self._result.get(key, False)
+            if any([isinstance(res, dict) and key in res and res[key] for res in self._result['results']]):
+                return True
+        return self._result.get(key, False)


bcoca:
I would say this is a bug in the yum module and that this code should warn about a failure of the module to return proper results in stead of 'overlooking' the bug.

willthames:
@bcoca, I disagree. As I say in #17208:

"One might argue that this should be an against yum rather than core - but the results of the yum task look reasonable in the absence of documentation that explicitly sets out conditions for module failues, and it seems to be the parsing of the results from yum that is at fault rather than yum module itself."

This code passes all tests and improves the logic of the check.

willthames:
The current check is ridiculous "If none of the contents of results have a `failed: True` key, return False"

bcoca:
I'm not saying we keep the check 'as is', but also ignoring incorrect 'results' returns is wrong.

I would prefer this code errors out our at the very least warns that the module output was incorrect, if a module is going to give a 'simulated' items 'results', it should do it correctly. 
(this seems to be another consequence of the 'squash' feature ...)

So there are 2 bugs here, one in the yum module's reporting, another in how we handled incorrect 'results' output.

bcoca:
_rambling:_
Since the 'results' key is 'internal' for when using `with_` loops I'm almost tempted to not allow modules to return it or only allow if there was a 'squash' happening, in which case the module should return in the same format produces on the Ansible side result processing.

Another thing we should do when squashing, is ensure there is a 'results' key populated per item passed, if the module does not provide one.
- note: the format of the 'results' key should always be a list of dicts, each dict being a 'normal' module response.

willthames:
You're talking about changing a behaviour introduced in the yum module 4 years ago. I'm talking about fixing the code that handles those results, introduced as a result of the Ansible 2.0 rewrite. I'd really prefer not to change yum's behaviour in that it's barely changed in four years, so we might end up breaking people's assumptions about yum's results, rather than fixing broken check code in core.

The code in task_result.py seems so buggy and poorly tested that almost every recent commit to that module is fixing some problem with `_check_key` or code that calls `_check_key`. I wonder if @nitzmahone's change to `is_skipped` in 133395d is basically handling the same problem.

dagwieers:
My proposed change (ansible/ansible-modules-core#4617) to yum in this case is to ensure that the module does not return rc=0 when there is a failure. At least this is picked up correctly as a failure of the task in a with-loop. I know that the real fix should look different, but this is the least invasive way for me to have correct output awaiting a decision, or even consensus :-)

jimi-c:
Closing, as an alternate fix was merged in.

PR #17255, opened 2016-08-26
Fix fetch idempotence

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

lib/ansible/plugins/action/fetch.py
##### ANSIBLE VERSION

```
devel and stable-2.1
```
##### SUMMARY

Fetch always follows symlinks when downloading so it needs to always
follow symlinks when getting the checksum of the file as well.


abadger:
Merged to devel and cherry-picked to stable-2.1

PR #17259, opened 2016-08-26
Fix wrong error class

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### SUMMARY

AnsibleError is not imported in that file, and since that's
a parsing time issue, better raise AnsibleParserError like the
rest of the file.

Issue signaled on irc by gordon`


mscherer:
Issue also affect stable release, so it can be a good idea to back port fix. It will likely puzzle people with incorrect playbook

gundalow: clarified that this means back port fix

PR #17261, opened 2016-08-26
Do not convert Nonetype to "None"

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### SUMMARY

If someone use a task with a empty name like this:

```
   - name:
     command: true
```

This will result in showing 'None' as a task name instead of
command.


abadger:
Merged to devel and cherry-picked to stable-2.1  Looks like we're going to do a 2.1.2rc2 so this change should make it in there.

PR #17274, opened 2016-08-28
make module argument type conversion more flexible

### LINKED ISSUE #4084: user: expires expects float epoch time
##### ISSUE TYPE

- Bug Report
##### COMPONENT NAME

user module
##### ANSIBLE VERSION

```
ansible 2.1.0.0
```
##### CONFIGURATION

inventory      = /var/local/ansible/hosts
roles_path      = /var/local/ansible/roles
log_path = /home/ansible/ansible.log
##### OS / ENVIRONMENT

CentOS Linux release 7.0.1406 (Core) 
##### SUMMARY

When adding a user, with the `expires` option, specifying epoch time as an int throws an error.

```
expires: 3268684800
```

results in:
`"argument expires is of type <type 'int'> and we were unable to convert to float"`

It works when changed to:

```
expires: 3268684800.0
```
##### STEPS TO REPRODUCE

```
  tasks:
  - name: add exampleuser
    user:
      name: exampleuser
      groups: wheel
      append: yes
      password: $1$X65joYEp$WiZRYzEF25NxYjBvmQK/C/
      expires: 3268684800
```

##### EXPECTED RESULTS

```
changed: [192.168.61.180]
```
##### ACTUAL RESULTS

```
fatal: [192.168.61.180]: FAILED! => {"changed": false, "failed": true, "msg": "argument expires is of type <type 'int'> and we were unable to convert to float"}
```

bcoca:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

module_utils/basic
##### ANSIBLE VERSION

```
2.2
```
##### SUMMARY

do not assume which types are convertible, try and deal with any errors

fixes https://github.com/ansible/ansible-modules-core/issues/4084

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -1483,31 +1483,28 @@ class AnsibleModule(object):
         raise TypeError('%s cannot be converted to a dict' % type(value))
 
     def _check_type_bool(self, value):
-        if isinstance(value, bool):
-            return value
-
-        if isinstance(value, string_types) or isinstance(value, int):
-            return self.boolean(value)
-
-        raise TypeError('%s cannot be converted to a bool' % type(value))
+        if not isinstance(value, bool):
+            try:
+                return self.boolean(value)
+            except:
+                raise TypeError('%s cannot be converted to a bool' % type(value))
+        return value
 
     def _check_type_int(self, value):
-        if isinstance(value, int):
-            return value
-
-        if isinstance(value, string_types):
-            return int(value)
-
-        raise TypeError('%s cannot be converted to an int' % type(value))
+        if not isinstance(value, int):
+            try:
+                return int(value)
+            except:
+                raise TypeError('%s cannot be converted to an int' % type(value))
+        return value
 
     def _check_type_float(self, value):
-        if isinstance(value, float):
-            return value
-
-        if isinstance(value, string_types):
-            return float(value)
-
-        raise TypeError('%s cannot be converted to a float' % type(value))
+        if not isinstance(value, float):
+            try:
+                return float(value)
+            except:
+                raise TypeError('%s cannot be converted to a float' % type(value))
+        return value
 
     def _check_type_path(self, value):
         value = self._check_type_str(value)


lib/ansible/module_utils/basic.py
```diff
@@ -1483,31 +1483,28 @@ def _check_type_dict(self, value):
...
-        if isinstance(value, bool):
-            return value
-
-        if isinstance(value, string_types) or isinstance(value, int):
-            return self.boolean(value)
-
-        raise TypeError('%s cannot be converted to a bool' % type(value))
+        if not isinstance(value, bool):
+            try:
+                return self.boolean(value)
```
abadger, line 1488: Would love to also modify self.boolean to raise a TypeError but unfortunately that's a public method that modules sometimes use :-(  I guess we can live with that doing something different and making a note to change it when we refactor basic.py.

lib/ansible/module_utils/basic.py
```diff
@@ -1483,31 +1483,28 @@ def _check_type_dict(self, value):
...
-        if isinstance(value, int):
-            return value
-
-        if isinstance(value, string_types):
-            return int(value)
-
-        raise TypeError('%s cannot be converted to an int' % type(value))
+        if not isinstance(value, int):
+            try:
+                return int(value)
```
abadger, line 1496: Do we want floats like int(1.9999999999)  to convert to 1?  We might want to throw an error on float still.  OTOH, if we allow floats, we probably also need to handle a string value of "1.0".

abadger, line 1496: For clarity -- in this PR, currently int(1.9) ==> 1   and int("1.9") ==> TypeError

abadger, line 1496: Also, things like int(True) == 1 and int(False) == 0, in python3 int(b'characters'[0]) == 99....

lib/ansible/module_utils/basic.py
```diff
@@ -1483,31 +1483,28 @@ def _check_type_dict(self, value):
...
-        if isinstance(value, float):
-            return value
-
-        if isinstance(value, string_types):
-            return float(value)
-
-        raise TypeError('%s cannot be converted to a float' % type(value))
+        if not isinstance(value, float):
+            try:
+                return float(value)
```
abadger, line 1504: int(True) == 1 and int(False) == 0, in python3 int(b'characters'[0]) == 99....

abadger:
After looking at the ways I could break this, I think a whitelist of allowed types is still the way to go.  We should just add other types to the whitelist that we think convert to the correct type in a sane manner.

abadger:
So because the original bug is that ints are not floats, we should just make this change in _check-type_float:

``` python
-        if isinstance(value, string_types):
+       if isinstance(value, (text_type, binary_type, int)):
```

string_types is a tuple so we can't use it if we want to add more types to check against.

abadger:
Merged the simple alternative: https://github.com/ansible/ansible/issues/17325  Closing this one.

PR #17289, opened 2016-08-29
implemented loop pausing

bcoca:
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

loops
##### ANSIBLE VERSION

```
2.2
```
##### SUMMARY

Add a configurable pause to each iteration of a loop.

### DIFF
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -255,7 +255,8 @@ class TaskExecutor:
             # pause between loop iterations
             if loop_pause and ran_once:
                 time.sleep(loop_pause)
-            ran_once = True
+            else:
+                ran_once = True
 
             try:
                 tmp_task = self._task.copy(exclude_parent=True, exclude_tasks=True)
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -237,17 +237,25 @@ class TaskExecutor:
 
         loop_var = 'item'
         label = None
+        loop_pause = 0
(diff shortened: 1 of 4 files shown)


lib/ansible/executor/task_executor.py
```diff
@@ -269,6 +270,8 @@ def _run_loop(self, items):
             self._rslt_q.put(TaskResult(self._host.name, self._task._uuid, res), block=False)
             results.append(res)
             del task_vars[loop_var]
+            if loop_pause:
```
mscherer: Loop_pause would be undefined if self._task.loop_control: is not defined on line 237

lib/ansible/executor/task_executor.py
```diff
@@ -269,6 +270,8 @@ def _run_loop(self, items):
...
+                time.sleep(loop_pause)
```
mscherer: Also, if there is a error in the task (like a AnsibleParserError), there would be no pause for that iteration

mscherer:
Tests and docs are missing.

lib/ansible/executor/task_executor.py
bcoca: ~~it is inside the if from 237, which ensures that loop_control is defined and has a value~~ missed the other if ... so fixed with declaration above.

lib/ansible/executor/task_executor.py
bcoca: I'm fine with that as the action did not happen

PR #17291, opened 2016-08-29
Fix the lines lookup to work on python3

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### SUMMARY

Since stdout is (on python3) of type 'bytes', callbacks
plugins fail in the test suite, because calls backs expect
a string.

### DIFF
--- a/lib/ansible/plugins/lookup/lines.py
+++ b/lib/ansible/plugins/lookup/lines.py
@@ -21,6 +21,7 @@ __metaclass__ = type
 import subprocess
 from ansible.errors import AnsibleError
 from ansible.plugins.lookup import LookupBase
+from ansible.module_utils._text import to_text
 
 class LookupModule(LookupBase):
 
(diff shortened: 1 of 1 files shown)


lib/ansible/plugins/lookup/lines.py
```diff
@@ -31,7 +32,7 @@ def run(self, terms, variables, **kwargs):
             p = subprocess.Popen(term, cwd=self._loader.get_basedir(), shell=True, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
             (stdout, stderr) = p.communicate()
             if p.returncode == 0:
-                ret.extend(stdout.splitlines())
+                ret.extend(list([to_native(l) for l in stdout.splitlines()]))
```
abadger: For controller code, we're using a text strategy for strings.  It's only for modules that we want to use the native string strategy.  You also shouldn't need the list() constructor.  A list comprehension will already create a list.  So the code can look like:

``` python
from ansible.module_utils._text import to_text
[...]

ret.extend([to_text(l) for l in stdout.splitline()])
```

abadger:
Thanks @mscherer !

PR #17292, opened 2016-08-29
Fix indexed_items.py to run on python3

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### SUMMARY

On python3, zip is a iterator so we need
to explicitly create the list from that.


abadger:
Thanks misc!

PR #17293, opened 2016-08-30
Make password lookup run on python3

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

password lookup
##### SUMMARY

os.urandom return 'bytes', and chars is a list of 'str',
thus the tests were failling with

```
"msg": "Unexpected templating type error occurred on
({{ lookup('password', output_dir + '/lookup/password_with_salt encrypt=sha256_crypt') }}):
'in <string>' requires string as left operand, not bytes"
```

### DIFF
--- a/lib/ansible/plugins/lookup/password.py
+++ b/lib/ansible/plugins/lookup/password.py
@@ -31,6 +31,7 @@ from ansible.plugins.lookup import LookupBase
 from ansible.parsing.splitter import parse_kv
 from ansible.utils.encrypt import do_encrypt
 from ansible.utils.path import makedirs_safe
+from ansible.module_utils._text import to_native
 
 DEFAULT_LENGTH = 20
 VALID_PARAMS = frozenset(('length', 'encrypt', 'chars'))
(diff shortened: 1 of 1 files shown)


lib/ansible/plugins/lookup/password.py
```diff
@@ -95,7 +96,7 @@ def random_password(self, length=DEFAULT_LENGTH, chars=C.DEFAULT_PASSWORD_CHARS)
 
         password = []
         while len(password) < length:
-            new_char = os.urandom(1)
+            new_char = to_native(os.urandom(1))
```
abadger, line 99: We should use to_text here instead of to_native since it's controller code.  Probably need to change the default in constants.DEFAULT_PASSWORD_CHARS to be a text literal string as well.

Not new with your code but `os.urandom(1)` won't work correctly with non-ascii characters.  Previous code may have been worse (I think it could have generated mangled, invalid utf-8.  The new code will only omit any non-ascii chars.)

We could use `random.SystemRandom().choice(chars)` here.  The upstream documentation is unclear but I think that is available everywhere that os.urandom() is so we don't lose any portability by using it.

abadger:
I've made the suggested changes (and a few other bytes vs text changes) here: https://github.com/ansible/ansible/pull/17475   alikins might do some more refactoring as he's trying to add unittests and having everything in a singe run() function is not conducive to that.

abadger:
Fix merged via https://github.com/ansible/ansible/pull/17626

PR #17294, opened 2016-08-30
Support for specifying item label in a loop

agaffney:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

Loop control
##### ANSIBLE VERSION

```
ansible 2.2.0 (loop_control_item_label faf35aa209) last updated 2016/08/29 18:41:15 (GMT -600)
  lib/ansible/modules/core: (detached HEAD 5310bab12f) last updated 2016/08/29 18:32:12 (GMT -600)
  lib/ansible/modules/extras: (detached HEAD 2ef4a34eee) last updated 2016/08/29 18:32:16 (GMT -600)
```
##### SUMMARY

This commit adds support for specifying the label for each item in a loop. This allows you to override the default of displaying the entire content of `item`, which is...verbose when looping over a complex structure.

Test playbook:

```

---
- hosts: localhost
  gather_facts: no
  vars:
    my_var:
      - name: foo
      - name: bar
      - name: baz
  tasks:
    - debug: var=item
      with_items: "{{ my_var }}"
      loop_control:
        label: "{{ item.name }}"
```

Result:

```
ok: [localhost] => (item=foo) => {
    "item": {
        "name": "foo"
    }
}
ok: [localhost] => (item=bar) => {
    "item": {
        "name": "bar"
    }
}
ok: [localhost] => (item=baz) => {
    "item": {
        "name": "baz"
    }
}
```

### DIFF
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -234,9 +234,11 @@ class TaskExecutor:
         task_vars = self._job_vars
 
         loop_var = 'item'
+        label = None
         if self._task.loop_control:
             # the value may be 'None', so we still need to default it back to 'item' 
             loop_var = self._task.loop_control.loop_var or 'item'
+            label = self._task.loop_control.label or ('{{' + loop_var + '}}')
 
         if loop_var in task_vars:
             display.warning("The loop variable '%s' is already in use. You should set the `loop_var` value in the `loop_control` option for the task to something else to avoid variable collisions and unexpected behavior." % loop_var)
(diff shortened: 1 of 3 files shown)


bcoca:
we are trying to keep new internal keys in the `_ansible` namespace, which will also get automatically removed from the final output. Other than that, looks good.

agaffney:
Updated to use `_ansible_item_label`

agaffney:
As part of testing this, I noticed a bug with the existing `loop_var` support. It seem the callback plugin base is hard-coded to expect an attribute named `item`, but when `loop_var` is in use, the loop_var value is used as the key when adding to the results dict. The result is:

```
ok: [localhost] => (item=None) => {
    "foo": {
        "name": "bar"
    }
}
```

The current version of this commit makes it behave properly.

```
ok: [localhost] => (item={u'name': u'foo'}) => {
    "foo": {
        "name": "foo"
    }
}
```

agaffney:
The latest commit only does the templating if `loop_control` is in use. This prevents adding extra overhead when it's not needed.

agaffney:
This touches some of the same lines as https://github.com/ansible/ansible/pull/17289, so one of them will probably need to be rebased to resolve merge conflicts after the other is merged.

amarao:
I'd like to open this issue because 'label' is not noted in documentation in http://docs.ansible.com/ansible/latest/playbooks_loops.html.

agaffney:
Yes, it is. It's under the first "Note" in this section:

https://docs.ansible.com/ansible/latest/playbooks_loops.html#loop-control

PR #17295, opened 2016-08-30
Fix url lookup for python 3

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

url lookup
##### SUMMARY

Several small fixes to make test test suite run with python 3, mostly around the url lookup.

### DIFF
--- a/lib/ansible/module_utils/urls.py
+++ b/lib/ansible/module_utils/urls.py
@@ -115,6 +115,7 @@ except ImportError:
 
 import ansible.module_utils.six.moves.urllib.request as urllib_request
 import ansible.module_utils.six.moves.urllib.error as urllib_error
+from ansible.module_utils.six import b
 
 try:
     # python3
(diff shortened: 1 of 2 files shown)


lib/ansible/module_utils/urls.py
```diff
@@ -606,11 +606,11 @@ def get_ca_certs(self):
                     full_path = os.path.join(path, f)
                     if os.path.isfile(full_path) and os.path.splitext(f)[1] in ('.crt','.pem'):
                         try:
-                            cert_file = open(full_path, 'r')
+                            cert_file = open(full_path, 'br')
                             os.write(tmp_fd, cert_file.read())
-                            os.write(tmp_fd, '\n')
+                            os.write(tmp_fd, b'\n')
```
abadger: Because this is in a module_utils file that could be run on python2.4 we need to use the helper functions from six for this:

``` python

from ansible.module_utils.six import b
[...]
os.write(tmp-fd, b('\n'))
```

lib/ansible/module_utils/urls.py
```diff
@@ -606,11 +606,11 @@ def get_ca_certs(self):
                     full_path = os.path.join(path, f)
                     if os.path.isfile(full_path) and os.path.splitext(f)[1] in ('.crt','.pem'):
                         try:
-                            cert_file = open(full_path, 'r')
+                            cert_file = open(full_path, 'br')
```
abadger: It seems that standard style in Ansible is "rb" rather than "br" (primary mode first, modifiers afterwards) but it's just style, not a blocker.

abadger:
One change needed and one style thing that you can change or ignore at your option.  This is good to merge after that.

abadger:
Thanks @mscherer !

PR #17302, opened 2016-08-30
Ignore difference on whitespace

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

test
##### SUMMARY

While trying to fix the test suite on python3, I noticed
this test fail due to to_json adding more whitespace in
python3 than in python2. So -w should ignored those
differences.


abadger:
Thanks @mscherer !

PR #17303, opened 2016-08-30
file() function in python3 no longer exist, so use open()

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

assemble


abadger:
A lot of times with file operations we also want to make sure that we're reading in byte strings and then do the conversion to text strings ourselves.  But that's something that can be done separately from getting rid of file().

Thanks!

PR #17304, opened 2016-08-30
Reduce no longer exist in python3, so we have to use six

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

filter


abadger:
Thanks.  merged.

PR #17305, opened 2016-08-30
On python3, 'str' no longer have a decode method

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

assemble
##### SUMMARY

So since delimiter is a string, I will assume that unicode
escaping is already handled natively and skip that part.

However, I am not 100% sure this is the right fix, maybe we should make sure that delimiter is a 'bytes' object, but that would requires a lot more change.

### DIFF
--- a/lib/ansible/plugins/action/assemble.py
+++ b/lib/ansible/plugins/action/assemble.py
@@ -28,6 +28,7 @@ from ansible.plugins.action import ActionBase
 from ansible.utils.boolean import boolean
 from ansible.utils.hashing import checksum_s
 from ansible.utils.unicode import to_str, to_unicode
+from ansible.module_utils.six import PY3
 
 
 class ActionModule(ActionBase):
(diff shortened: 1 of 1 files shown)


abadger:
This is a hard problem.  The reason for the "unicode-decode" is to unescape things like "\n" in the string.  There's no simple and correct way to do this in python3.  However, there do appear to be two ways to do this, one simple and "incorrect" and one complex.

Simple and "incorrect"... I lean towards this way of doing it:  http://stackoverflow.com/a/37059682

That would look something like this:

``` python
from ansible.module_utils._text import to_bytes, to_text
from codecs import escape_decode
[...]
b_delimiter = to_bytes(delimiter)
b_delimiter = escape_decode(b_delimiter)[0]
delimiter = to_text(b_delimiter)
```

The more complex answer involves a regex: http://stackoverflow.com/a/24519338
I won't go into that here as the example is pretty self-explanatory.

Right now we have some code (in unarchive) using the simple solution and some code in parsing/splitter.py using a modified regex (we decided not to handle octals in the parsing/splitter code).  I think as long as the simple answer continues to work, that's the one we should standardize on.

abadger:
Fixed via escape_decode in https://github.com/ansible/ansible/pull/17866

PR #17306, opened 2016-08-30
Running the test with a locale other than English make that test fail

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

tests
##### SUMMARY

While tests are supposed to be run in docker, people might still
want to use them on their own boxes.

In my case, I found out because git started to be translated in french on Fedora 24.

### DIFF
--- a/test/integration/roles/test_git/tasks/main.yml
+++ b/test/integration/roles/test_git/tasks/main.yml
@@ -421,6 +421,8 @@
   register: remote_url
   args:
     chdir: '{{ checkout_dir }}'
+  environment:
+    LC_ALL: C
 
 - assert:
     that: 


abadger:
This is probably a bug in the git module.  It's good that you found it!

We probably are screenscraping output from git and need to set the module to use the C locale before doing that.  The yum and apt modules show the two different ways of doing that

apt does it globally for any invocation of run_command: https://github.com/ansible/ansible-modules-core/blob/devel/packaging/os/apt.py#L698

yum does it for each individual command: https://github.com/ansible/ansible-modules-core/blob/devel/packaging/os/yum.py#L282

In most cases either strategy will work.  Globally setting it is less work.  If there's some commands that shouldn't set the locale related environment variables then you have to use the individual command method.

abadger:
So -- I think that this is a genuine bug in the git module.  If you think it shouldn't be solved there, let me know what the output you're getting is and we can revisit changing the test instead.

mscherer:
Nope, taht's a bug in the test, since it use the shell module to scrap the output of git command.

abadger:
Roger that.  Merging.  Thanks for straightening me out!

PR #17309, opened 2016-08-30
First release of the callback selective.py

dbarrosop:
##### ISSUE TYPE
- New callback
##### COMPONENT NAME

selective.py
##### ANSIBLE VERSION

```
$ ansible --version
ansible 2.1.0.0
  config file = /Users/dbarroso/workspace/ansibly/ansible.cfg
  configured module search path = ['library']
```
##### SUMMARY

This is a new callback that let's you tag the tasks you want to show on the screen. This is useful when you have a playbook with hundreds of tasks and you don't want to clutter the output with tasks that provide no operational value like "creating folder to hold configuration files" or "setting facts" or "dropping temporary files" that you will assemble later on, etc.

By default, tasks are replaced on screen by a dot. Tasks that fail are always printed on screen. If you want to show a particular task you have to tag it with the label "print_action".

For example:

The playbook:

```

---
- name: Test playbook
  hosts: all

  tasks:
    - debug: msg="This will not be printed"
    - debug: msg="This will not be printed"
    - debug: msg="This will not be printed"
    - debug: msg="This will not be printed"
    - debug: msg="This will not be printed"
    - debug: msg="But this will"
      tags: [print_action]
    - assert: that=False
```

Will yield the following output:

```
$ ansible-playbook test.yml --limit "edge-lab*"
............
# debug *******************************************************************************************************************
  * edge-lab-100002         - changed=False -- But this will -----------------------------------
  * edge-lab-100001         - changed=False -- But this will -----------------------------------
# assert ******************************************************************************************************************
  * edge-lab-100002         - FAILED!!! --  ----------------------------------------------------
  * edge-lab-100001         - FAILED!!! --  ----------------------------------------------------

# STATS *******************************************************************************************************************
edge-lab-100001    : ok=7       changed=0       failed=1        unreachable=0
edge-lab-100002    : ok=7       changed=0       failed=1        unreachable=0
```


mattclay:
Tests on Shippable failed because this PR branch was created before Shippable was enabled. If you rebase the branch the tests will run correctly.

dbarrosop:
Well, apparently some tests in the branch are broken because shippable is still failing :)

mattclay:
@dbarrosop Shippable is failing due to issues with the callback plugin, such as this one:

https://app.shippable.com/runs/57d26668da95e70f00417d3a/34/console

```
SyntaxError: ('invalid syntax', ('lib/ansible/plugins/callback/selective.py', 61, 43, '    return {k: (prv.get(k), nxt.get(k)) for k in keys if prv.get(k) != nxt.get(k)}\n'))
```

You'll need to look at the failures on each tab with errors to see if there are any other issues.

dbarrosop:
@mattclay tests seem to be fine now.

dbarrosop:
So, new errors seem unrelated to the PR, looks like some windows stuff is failing for some other reason.

mattclay:
@dbarrosop I restarted the build and it's now passing.

dbarrosop:
@mattclay so, what is the next step for this? Do I have to do something here?

lib/ansible/plugins/callback/selective.py
```diff
@@ -0,0 +1,257 @@
...
+    def _print_host_or_item(self, host_or_item, changed, msg, diff, is_host, error, stdout, stderr):
+        if is_host:
+            indent_level = 0
+            name = colorize(host_or_item.name, 'not_so_bold')
+        else:
+            indent_level = 4
+            if isinstance(host_or_item, dict):
+                if 'key' in host_or_item.keys():
+                    host_or_item = host_or_item['key']
+            name = colorize(unicode(host_or_item), 'bold')
```
mattclay: Python 3.5 does not have `unicode`. Take a look at the [ansible text helper methods](https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/_text.py) for alternatives.

Only noting the first occurrence.

mattclay (changes requested):
There's at least one Python 3.5 compatibility issue as noted. Testing on Python 2.6, 2.7 and 3.5 is important. A good way to do this would be to write a simple integration test to use the callback plugin.

dbarrosop:
Is there any interest at all here? This has been open since August and I am not in the mood of having to amend the PR every 6 months because you keep changing things and it's never merged.

mattclay:
@dbarrosop As noted in my review feedback, `unicode` isn't supported on python 3.5. Please correct this so the callback will work on python 2.6, 2.7 and 3.5.

mattclay (approved):

mattclay:
@dbarrosop Thanks for the new callback.

PR #17310, opened 2016-08-30
Add support for become_flags on the task level

mscherer:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

become
##### SUMMARY

Working on the test suite, I tried to replace a call to sudo to a
call to su, and found out that I can't change user to 'nobody'
without changing the option become_flags in ansible.cfg

As this would be dependent on the user and the task, it make more sense
to push the setting there.


abadger:
jimi-c reviewed and gave a +1.  merged to devel.  Thanks!

PR #17312, opened 2016-08-30
Improved caching for urls

### LINKED ISSUE #2966: get_url prevents caching of data
Issue Type:
Bug Report

Component Name:
get_url module

Ansible Version:
ansible 2.0.0.2
Ansible Configuration:

Stock.

Environment:

Ubuntu 14.04

Summary:

get_url seems to add headers that prevent efficient use of a caching proxy.

Steps To Reproduce:

I am trying to download some ISOs to multiple machines (via a proxy to 
conserve bandwidth).

The ISO is being stored by the proxy, and the machine is using the proxy, but 
it is downloading from the upstream source every time. Squid is showing in its 
logs:

1454462392.008 532579 192.168.122.10 TCP_CLIENT_REFRESH_MISS/200 632291702 GET 
http://mirrors.kernel.org/centos/7.2.1511/isos/x86_64/CentOS-7-x86_64-Minimal-1511.iso - HIER_DIRECT/198.145.20.143 application/octet-stream

According to the squid docs:

TCP_CLIENT_REFRESH_MISS
The client issued a "no-cache" pragma, or some analogous cache control command 
along with the request. Thus, the cache has to refetch the object.

Using a standard wget (or, say, yum to retrieve packages) does not cause 
CLIENT_REFRESH_MISSes.

Is there something in the get_url code that is causing the sending of a no 
cache pragma? Or maybe it's not turning off some default option in the 
underlying urllib (or whatever it uses under the hood)?

Expected Results:

File will be downloaded much faster, and from the cache.

Actual Results:

Squid is actually downloading the file again.

MichaelRiss:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

##### ANSIBLE VERSION

```
ansible 2.2.0 (get_url-cache-improvement f7d0698bda) last updated 2016/08/30 11:54:22 (GMT +200)
  lib/ansible/modules/core: (detached HEAD 5310bab12f) last updated 2016/08/30 11:14:57 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD 2ef4a34eee) last updated 2016/08/30 11:15:02 (GMT +200)
```
##### SUMMARY

With this change `get_url` can now make use of a proxy even when the file does not yet exist on the target system. This is important especially when you want to mass deploy and use a cache to relieve your internet connection.

Fixes ansible/ansible-modules-core#2966

```

```
- When there is no file at the destination yet, we have no modification time for the `If-Modified-Since`-Header. In this case trust the cache to make the right decision to either serve a cached version or to refresh from origin. This should help with mass-deployment scenarios where you want to use a local cache to relieve your uplink.
- If you don't trust the cache to make the right decision you can still force it to refresh by providing the `force: yes` option.

### DIFF
--- a/lib/ansible/module_utils/urls.py
+++ b/lib/ansible/module_utils/urls.py
@@ -846,13 +846,14 @@ def open_url(url, data=None, headers=None, method=None, use_proxy=True,
     # with sites that block the default urllib agent string
     request.add_header('User-agent', http_agent)
 
-    # if we're ok with getting a 304, set the timestamp in the
-    # header, otherwise make sure we don't get a cached copy
-    if last_mod_time and not force:
+    # Cache control
+    # Either we directly force a cache refresh
+    if force:
+        request.add_header('cache-control', 'no-cache')
+    # or we do it if the original is more recent than our copy
+    elif last_mod_time:
         tstamp = last_mod_time.strftime('%a, %d %b %Y %H:%M:%S +0000')
         request.add_header('If-Modified-Since', tstamp)
-    else:
-        request.add_header('cache-control', 'no-cache')
 
     # user defined headers now, which may override things we've set above
     if headers:


MichaelRiss:
I'm a bit confused about the failing checks. Yesterday after submitting the PR, the checks passed. I didn't change anything in the mean time and now they fail? Also, the failing test doesn't look like it has anything to do with URL handling.

MichaelRiss:
Figured out how to restart the shippable test - now they pass again.
Also I changed the type to `Bugfix Pull Request` because after going back to the vanilla ansible and having every file squeezed through my internet connection, not using the cache really feels like a bug.

jpmens:
shipit

MichaelRiss:
Ping? Or what's the roadmap for this PR now? Can I do something?

abadger:
This has been merged to devel andcherry-picked to the temp-post-2.2.0-staging branch.  That means it won't make the 2.2.0 release but will be in 2.2.1.  @MichaelRiss Thanks for writing the code and @jpmens thanks for reviewing it!

PR #17320, opened 2016-08-31
fixes issue with converting commands to an iterable was not being done

privateip:
This fix will now force the commands arg in **call** to be a list of
objects which otherwise would cause netcli not process the stack

This should fix the issue reported in ansible/ansible-modules-core#4419


gundalow:
@ipspace has confirmed this fix resolves the issue

PR #17326, opened 2016-08-31
Enable the test of 'su', since it seems to work fine now

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

tests
##### SUMMARY

While there is a comment about disabling it, a test show that it now work, so let's bring the test back.

### DIFF
--- a/test/integration/destructive.yml
+++ b/test/integration/destructive.yml
@@ -3,7 +3,7 @@
   roles:
     # In destructive because it creates and removes a user
     - { role: test_sudo, tags: test_sudo}
-    #- { role: test_su, tags: test_su} # wait till su support is added to local connection, needs tty
+    - { role: test_su, tags: test_su}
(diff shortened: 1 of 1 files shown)


mscherer:
So the error message is surprising:

```
2016-08-31 16:05:51 fatal: [testhost]: FAILED! => {"changed": true, "failed": true, "msg": "Could not replace file: /tmp/ansible-tmp-1472659550.9-14443715784950/source to /home/ansibletest1/bar.txt: [Errno 13] Permission denied: '/tmp/ansible-tmp-1472659550.9-14443715784950/source'"}
```

mattclay:
@mscherer The tests for `su` have been removed, as the feature is deprecated. The feature was to be removed in `stable-2.2` but that has been delayed until the 2.3 release.

PR #17327, opened 2016-08-31
Disable yum test on python 3

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

test
##### SUMMARY

Since yum is not gonna be running on python3 ever, we have to filter it.


abadger:
Yep.  Seems like the only choice we can make.

Thanks.

PR #17336, opened 2016-08-31
Add test for the pause in loop feature

mscherer:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

loops
##### SUMMARY

See https://github.com/ansible/ansible/pull/17289


abadger:
Looks good to me.

@mattclay care to take a quick look and merge if you like it too?

bcoca:
this is subject to 'test lag', i would instead make the test run a `date` command and make sure that the difference is >= than the pause.

mattclay:
@mscherer Check only for `>= 4` seconds and leave off the `<= 6` second check. That way slow CI won't cause the test to fail.

PR #17337, opened 2016-08-31
Add test of the selinux module (as a start)

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

tests
##### SUMMARY

the test should be skipped if there is no selinux, but as asked on irc, i am not sure selinux is enabled anywhere in our CI.

### DIFF
--- a/test/integration/roles/setup_selinux/tasks/main.yml
+++ b/test/integration/roles/setup_selinux/tasks/main.yml
@@ -1,0 +1,15 @@
+- include_vars: '{{ item }}'
+  with_first_found:
+  - files:
+    - '{{ ansible_distribution }}-{{ ansible_distribution_major_version }}.yml'
+    - '{{ ansible_os_family }}-{{ ansible_distribution_major_version }}.yml'
+    - '{{ ansible_distribution }}.yml'
+    - '{{ ansible_os_family }}.yml'
+    - 'default.yml'
+    paths: '../vars'
+
+- name: Install packages for selinux
+  package:
+    name: "{{ item }}"
+    state: present
+  with_items: "{{ selinux_packages[ansible_python['version']['major']] | default([]) }}"
--- a/test/integration/destructive.yml
+++ b/test/integration/destructive.yml
@@ -23,3 +23,4 @@
     - { role: test_uri, tags: test_uri }
     - { role: test_get_url, tags: test_get_url }
     - { role: test_apache2_module, tags: test_apache2_module }
+    - { role: test_selinux, tags: test_selinux, when: ansible_selinux is defined and ansible_selinux.status == 'enabled' }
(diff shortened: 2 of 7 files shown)


mattclay:
@mscherer Looks good. All the test failures appear to be unrelated to the changes. As discussed in IRC the tests are being skipped on Shippable since selinux isn't available.

mscherer:
Yup, so do we decide to merge that,  and risking bitrot, or we merge in the hope that's useful for the internal jenkins ?

mattclay:
I think they're worth merging, even though we can't run them on Shippable yet.

test/integration/roles/setup_selinux/tasks/main.yml
```diff
@@ -0,0 +1,4 @@
+- name: Install packages for selinux
+  package:
+    name: libselinux-python
+    state: present
```
abadger: Do we need to install the libselinux-python3 package if ansible is running on python3?

abadger:
+1 to adding the tests and risking bitrot.  Made a note that we may need to install the python3 version of the package as well.

test/integration/roles/setup_selinux/tasks/main.yml
mscherer: That's likely. But I do not know what is the plan for testing on python3, since we do have the same problem for postgresql and mysql and anything that need a python3 library.

mattclay:
The `libselinux-python` package is distro specific. The setup should work more like `setup_postgresql_db`, which accommodates different distros. If a distro isn't supported, the tests can be skipped, just as if selinux isn't available.

That would also be the correct way to handle python3. For example, on Fedora the python3 package is `libselinux-python3` instead of `libselinux-python`.

mscherer:
But how do we know that the tests are running on python3 ? Last time I checked, we just had PYTHON3=1 as a env var, so should we care now, or just change it later ?

mattclay:
@mscherer You can check the `ansible_python_version` fact to determine the python version.

mattclay:
@mscherer Do you want to update these tests to work with `ansible-test`?

mattclay:
CI now runs integration tests on RHEL 7.3, so we're finally able to run selinux tests.

gundalow:
@mscherer `test/integration/destructive.yml` has been removed so you don't need to update this anymore

gundalow:
If these files get moved to `test/integration/targets` they will be able to run via `ansible-test` and therefore defended in Shippable

mattclay:
@mscherer Are you interested in updating this PR?

Akasurde:
Closing this PR as selinux testcases are added via 00df1fda1072223580022a9e42d25fba393c494a

@mscherer Thanks for the contribution.

PR #17351, opened 2016-09-01
1st draft really documenting common return vals

bcoca:
##### ISSUE TYPE

- Docs Pull Request
##### COMPONENT NAME

module return vals
##### ANSIBLE VERSION

```
2.2
```
##### SUMMARY

Document common return values.

### DIFF
--- a/docsite/rst/common_return_values.rst
+++ b/docsite/rst/common_return_values.rst
@@ -45,7 +45,7 @@ Some modules execute command line utilities or are geared for executing commands
 
 stderr_lines
 ````````````
-When c(stderr) is returned we also always provide this field wich is a list of strings, one item per line from the original.
+When c(stderr) is returned we also always provide this field which is a list of strings, one item per line from the original.
 
 stdout
 ``````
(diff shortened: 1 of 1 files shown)


docsite/rst/common_return_values.rst
```diff
@@ -8,35 +8,68 @@ Common Return Values
...
+A boolean that indicates if the task was skipped or not
+
+stderr
+``````
+Some modules execute command line utilities or are geared for executing commands directly (raw, shell, command, etc), this field contains the error output of these utilities.
 
-.. _status:
+stderr_lines
+````````````
+When c(stderr) is returned we also always provide this field wich is a list of strings, one item per line from the original.
```
gundalow: s/wich/which/

docsite/rst/common_return_values.rst
```diff
@@ -8,35 +8,68 @@ Common Return Values
...
+stdout
 ``````
+Some modules execute command line utilities or are geared for executing commands directly (raw, shell, command, etc), this field contains the normal output of these utilities.
 
-Every module must return a status, saying if the module was successful, if anything changed or not. Ansible itself
-will return a status if it skips the module due to a user condition (when: ) or running in check mode when the module
-does not support it.
+stdout_lines
+````````````
+When c(stdout) is returned we also always provide this field wich is a list of strings, one item per line from the original.
```
gundalow: s/wich/which/

gundalow:
Is it worth adding example output?

bcoca:
fixed typo and merging this

@gundalow 
examples would be great, we can add in subsequent PRs

gundalow:
I'll pool the Testing Working Group members to get some good examples

Issue #16878, opened 2016-07-29
group_vars ignored with inventory script in ansible 2.1.1.0


andrenarchy:
##### ISSUE TYPE
- Bug Report
##### ANSIBLE VERSION

```
ansible 2.1.1.0
```
##### OS / ENVIRONMENT

Ubuntu 16.04
##### SUMMARY

group_vars are ignored if the inventory is the output of a script that is stored in a subdirectory. This includes the popular ec2 inventory script. This bug was introduced in 2.1.1.0 (2.1.0.0 works fine).
###### `inventory/script.sh`

```
#!/bin/sh

echo '{"webservers": ["paperberry"]}'
```
###### `group_vars/all.yml`:

```

---
frontend_path: /var/www/html
```
###### `playbooks/webservers.yml`:

```

---
- hosts: all
  connection: local
  tasks:
  - debug: var=frontend_path
```

Run the following code:

```
ansible-playbook -i inventory/script.sh playbooks/webservers.yml 
```
##### EXPECTED RESULTS

ansible 2.1.0.0 shows

```
ok: [paperberry] => {
    "frontend_path": "/var/www/html"
}
```
##### ACTUAL RESULTS

ansible 2.1.1.0 shows

```
ok: [paperberry] => {
    "frontend_path": "VARIABLE IS NOT DEFINED!"
}
```

bcoca:
# Possible Misunderstanding

Hi!

Thanks very much for your submission to Ansible.  It sincerely means a lot to us.

We believe the ticket you have filed is being somewhat misunderstood, as one thing works a little differently than stated.

It has always been that group_vars are adjacent to inventory or play, not to the directory that contains them.

What you might have run into was a bug in 2.1 in which group_vars were also loaded from the 'current working directory', which was fixed in 2.1.1.

In the future, this might be a topic more well suited for the user list, which you can also post here if you'd like some more help with the above.
- https://groups.google.com/forum/#!forum/ansible-project

Thank you once again for this and your interest in Ansible!

andrenarchy:
Thanks for pointing this out. The docs are also clear about this.

haad:
Would it be possible to bring this behavior back, maybe disabled by default?

RRAlex:
Strange, this has always worked for me (tm), even before 2.1.0 and all the way to 1.9.x.
Am I missing something or is it now quite hard to organize playbooks without doing a bunch of "ln -s" to the group_vars / host_vars from within all the playbooks folders?

[...]

PR #17354, opened 2016-09-01
Improves Inventory Best Practices Documentation

kustodian:
##### ISSUE TYPE
- Docs Pull Request
##### COMPONENT NAME

Documentation
##### ANSIBLE VERSION

```
ansible 2.2.0
```
##### SUMMARY

Splits the production and staging inventories into separate directories, so that they can have their own separate `group_vars`/`host_vars`. Having the same group/host_vars doesn't make sense, because you usually want to configure different variables in production and staging.

### DIFF
--- a/docsite/rst/playbooks_best_practices.rst
+++ b/docsite/rst/playbooks_best_practices.rst
@@ -68,6 +68,47 @@ The top level of the directory would contain files and directories like so::
 
 .. note: If you find yourself having too many top level playbooks (for instance you have a playbook you wrote for a specific hotfix, etc), it may make sense to have a playbooks/ directory instead.  This can be a good idea as you get larger.  If you do this, configure your roles_path in ansible.cfg to find your roles location.
 
+.. _alternative_directory_layout:
+
+Alternative Directory Layout
+````````````````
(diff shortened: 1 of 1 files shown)


bcoca:
I would post this as an 'alternate' way of setting up inventory, not necessarily an improvement for all environments.

kustodian:
If just changing hosts in the inventory file and keeping all group/host_vars the same works for you, the current best practices is ok, but in practice it usually isn't like that. Having separate directories for inventories gives a lot more flexibility, but it does make things more complex.

If you think it would be best to make this an alternate way, I will make the change like that and explain the reasoning for this approach.

kustodian:
@bcoca I did what you suggested. I added it as an alternative layout. Could you please check if you are ok with these changes, or would you like to make some further changes.

viper233:
How will playbook group_vars and host_vars work?

Causing some issues for others.

https://github.com/ansible/ansible/issues/16878

Sodki:
@viper233 You should be able to have them both in the "root level" directory of Ansible, to affect all hosts, AFAIK.

PR #17357, opened 2016-09-01
Add detection of python3 to hacking/env-setup

mscherer:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

hacking/env-setup
##### SUMMARY

This permit to at least run the tests on python 3 with TARGET=ubuntu1604py3 ./test/utils/run_tests.sh

People can still override the detection using PYTHON_BIN


abadger:
Thanks @mscherer!  I don't know if this is the best way but it will certainly work.  Merging and if anyone thinks up something more fancy we can always change it.

PR #17364, opened 2016-09-02
Add oVirt utility module

machacekondra:
##### ISSUE TYPE

- New Module Pull Request
##### COMPONENT NAME

utils/module_docs_fragments/ovirt.py
module_utils/ovirt.py
##### ANSIBLE VERSION

```
2.2.0
```
##### SUMMARY

This commit consist of two new files added to ansible.
1. _module_utils/ovirt.py_. This file contains helper class and functions, for oVirt ansible modules.
2. _utils/module_docs_fragments/ovirt.py_. This file contains common documentation for oVirt ansible modules.

The oVirt modules PR is located [here](https://github.com/ansible/ansible-modules-extras/pull/2836)

This patch add oVirt utility module, which contains helper functions,
for oVirt modules and also shared documentation fragment for [oVirt](http://ovirt.org/).

### DIFF
--- a/lib/ansible/module_utils/ovirt.py
+++ b/lib/ansible/module_utils/ovirt.py
@@ -123,7 +123,7 @@ def convert_to_bytes(param):
     elif param.isdigit():
         return int(param) * 2**10
     else:
-        raise Exception(
+        raise ValueError(
             "Unsupported value(IEC supported): '{value}'".format(value=param)
         )
 
@@ -225,12 +225,13 @@ def search_by_name(service, name, **kwargs):
     return res[0]
 
 
-def wait(service, condition, timeout=180, wait=True):
+def wait(service, condition, fail_condition=lambda e: False, timeout=180, wait=True):
     """
     Wait until entity fulfill expected condition.
 
     :param service: service of the entity
     :param condition: condition to be fulfilled
+    :param fail_condition: if this condition is true, raise Exception
     :param timeout: max time to wait in seconds
     :param wait: if True wait for condition, if False don't wait
     """
@@ -239,8 +240,11 @@ def wait(service, condition, timeout=180, wait=True):
         start = time.time()
         while time.time() < start + timeout:
             # Exit if the condition of entity is valid:
-            if condition(service.get()):
+            entity = service.get()
+            if condition(entity):
                 return
+            elif fail_condition(entity):
+                raise Exception("Error while waiting on result state of the entity.")
             else:
                 time.sleep(float(wait))
 
@@ -296,8 +300,8 @@ class BaseModule(object):
     @abstractmethod
     def build_entity(self):
         """
-        This method should return oVirt Python SDK type, which we want want
-        to create or update, initialized by values passed by Ansible module.
+        This method should return oVirt Python SDK type, which we want to
+        create or update, initialized by values passed by Ansible module.
 
         For example if we want to create VM, we will return following:
           types.Vm(name=self._module.params['vm_name'])
@@ -340,7 +344,7 @@ class BaseModule(object):
         """
         pass
 
-    def create(self, entity=None, result_state=None, search_params=None, **kwargs):
+    def create(self, entity=None, result_state=None, fail_condition=lambda e: False, search_params=None, **kwargs):
         """
         Method which is called when state of the entity is 'present'. If user
         don't provide `entity` parameter the entity is searched using
@@ -348,12 +352,13 @@ class BaseModule(object):
         the entity should be updated is checked by `update_check` method.
         The corresponding updated entity is build by `build_entity` method.
 
-        Additional task can be executed after entity is created by overriding
-        `post_create` method. Another task can be executed after update, by
-        overriding `post_update` method.
+        Function executed after entity is created can optionally be specified
+        in `post_create` parameter. Function executed after entity is updated
+        can optionally be specified in `post_update` parameter.
 
         :param entity: Entity we want to update, if exists.
         :param result_state: State which should entity has in order to finish task.
+        :param fail_condition: Function which checks incorrect state of entity, if it returns `True` Exception is raised.
         :param search_params: Dictionary of parameters to be used for search.
         :param kwargs: Additional parameters passed when creating entity.
         :return: Dictionary with values returned by Ansible module.
@@ -386,10 +391,11 @@ class BaseModule(object):
 
         state_condition = lambda entity: entity
         if result_state:
-            state_condition = lambda entity: entity and entity.status == result_state,
+            state_condition = lambda entity: entity and entity.status == result_state
         wait(
             service=entity_service,
             condition=state_condition,
+            fail_condition=fail_condition,
             wait=self._module.params['wait'],
             timeout=self._module.params['timeout'],
         )
@@ -414,8 +420,8 @@ class BaseModule(object):
         don't provide `entity` parameter the entity is searched using
         `search_params` parameter. If entity is found it's removed.
 
-        Additional task can be executed after entity is searched and before
-        remove action is executed by overriding `pre_remove` method.
+        Function executed before remove is executed can optionally be specified
+        in `pre_remove` parameter.
 
         :param entity: Entity we want to remove.
         :param search_params: Dictionary of parameters to be used for search.
@@ -457,6 +463,7 @@ class BaseModule(object):
         entity=None,
         action_condition=lambda e: e,
         wait_condition=lambda e: e,
+        fail_condition=lambda e: False,
         pre_action=lambda e: e,
         post_action=lambda e: None,
         search_params=None,
@@ -466,18 +473,18 @@ class BaseModule(object):
         This method is executed when we want to change the state of some oVirt
         entity. The action to be executed on oVirt service is specified by
         `action` parameter. Whether the action should be executed can be
-        specified by passing `action_condition` parameter. Which state the entity
-        should have after executing the action can be specified by passing
-        `wait_condition` parameter.
+        specified by passing `action_condition` parameter. State which the
+        entity should be in after execution of the action can be specified
+        by `wait_condition` parameter.
 
-        Additional task can be executed before action on entity is executed
-        by passing `pre_action` parameter to method. Another task can be
-        executed after action is executed, by passing `post_action` parameter
-        to method.
+        Function executed before an action on entity can optionally be specified
+        in `pre_action` parameter. Function executed after an action on entity can
+        optionally be specified in `post_action` parameter.
 
-        :param action: Action which should be executed on service.
+        :param action: Action which should be executed by service on entity.
         :param entity: Entity we want to run action on.
         :param action_condition: Function which is executed when checking if action should be executed.
+        :param fail_condition: Function which checks incorrect state of entity, if it returns `True` Exception is raised.
         :param wait_condition: Function which is executed when waiting on result state.
         :param pre_action: Function which is executed before running the action.
         :param post_action: Function which is executed after running the action.
@@ -507,8 +514,9 @@ class BaseModule(object):
         post_action(entity)
 
         wait(
-            service=entity_service,
+            service=self._service.service(entity.id),
             condition=wait_condition,
+            fail_condition=fail_condition,
             wait=self._module.params['wait'],
             timeout=self._module.params['timeout'],
         )
--- a/lib/ansible/module_utils/ovirt.py
+++ b/lib/ansible/module_utils/ovirt.py
@@ -1,18 +1,22 @@
+#!/usr/bin/python
 # -*- coding: utf-8 -*-
 #
 # Copyright (c) 2016 Red Hat, Inc.
 #
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
+# This file is part of Ansible
 #
-#   http://www.apache.org/licenses/LICENSE-2.0
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
 #
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


machacekondra:
CC @ryansb @detiber

gundalow:
needs_revision

Shippable is failing with:

```
2016-09-02 15:07:12 + python2.4 -m compileall -fq -x 'module_utils/(a10|rax|openstack|ec2|gce|lxd|docker_common|azure_rm_common|vca|vmware|gcp|gcdns).py' lib/ansible/module_utils
2016-09-02 15:07:12 Compiling lib/ansible/module_utils/ovirt.py ...
2016-09-02 15:07:12   File "lib/ansible/module_utils/ovirt.py", line 236
2016-09-02 15:07:12     yield
```

detiber:
@machacekondra I don't see any issues with a quick review. With this being in module_utils, I suspect there are more strict requirements on python versions that need to be supported, but one of the ansible devs would have to provide more info.

t-v:
is there a specific reason this is in module_utils, and not made as a normal module? I can personally see no reason why it should be running on the ansible host itself, especially since it creates 2 dependencies on that host:
- it should have the python oVirt SDK.
- it should have a web access to the manager.
  These 2 reasons are actually why I would prefer to see it act as a normal module.

And a personal 3rd reason, why would we need a 3rd module to manage VMs on the same platform?
We have the original oVirt module: https://github.com/ansible/ansible-modules-extras/blob/devel/cloud/misc/ovirt.py
We have a rhevm module:
https://github.com/ansible/ansible-modules-extras/blob/devel/cloud/misc/rhevm.py

I created the rhevm module, because the original one was lacking in some features like multiple disks and interfaces and since all of the changes were too big, about a year ago we decided to make it a totally different module. That way, all of the people using the old module didn't have any issue since it was already available in v1.4, but more features were available in the new one. So out of curiousity, what does this one add to the table so there might be a way to simplify and remove the other 2?
I would personally see no issue in it replacing my module, but ofcourse that means you should implement the features that are already in the other ones.

And last but not least, since you are using the same name as the other ovirt module, what about conflicts?

@vincentvdk your view? since you made the original one.

machacekondra:
> is there a specific reason this is in module_utils, and not made as a normal module?

Yes, there is. This _module_utils_ is shared across all modules we will add in future, we already have modules for vms,disks,hosts,storage_domains,cluster,datanceters,user/groups,permissions. But we addded PR with _vms_ and _auth_ only, for now. So this util is there so we have can reuse that common code.

> And a personal 3rd reason, why would we need a 3rd module to manage VMs on the same platform?
>  So out of curiousity, what does this one add to the table so there might be a way to simplify and remove the other 2?

As said in one of your PR. Your modules supports only REST API v 3 of the oVirt. New modules will support version 4 and it will support much more entities of oVirt.

> And last but not least, since you are using the same name as the other ovirt module, what about conflicts?

There are no conficts I use _ovirt_ directory in _cloud_ directory, you are using _misc_ directory and completely different names of modules.

> I would personally see no issue in it replacing my module, but ofcourse that means you should implement the features that are already in the other ones.

Those module would not replace yours, people using v3 API needs to use those, but people using v4 API will have to use new modules.

I hope it answers most of your questions, please ask if not, or you have more. :)

machacekondra:
CC @mwperina @lukas-bednar

lib/ansible/utils/module_docs_fragments/ovirt.py
```diff
@@ -0,0 +1,56 @@
...
+        description:
+            - "The amount of time the module should wait for the instance to
+               get into desired state."
+        default: 180
+requirements:
+  - python >= 2.7
+  - ovirt-engine-sdk-python >= 4.0.0
+notes:
+  - "In order to use this module you have to install oVirt Python SDK.
+     To ensure it's installed with correct version you can create following task:
```
mykaul: following task -> the following task

lib/ansible/utils/module_docs_fragments/ovirt.py
```diff
@@ -0,0 +1,56 @@
...
+            certificate and host name should be checked."
+            - "C(ca_file) - A PEM file containing the trusted CA certificates. The
+            certificate presented by the server will be verified using these CA
+            certificates. If `C(ca_file)` parameter is not set, system wide
+            CA certificate store is used."
+            - "C(kerberos) - A boolean flag indicating if Kerberos authentication
+            should be used instead of the default basic authentication."
+    timeout:
+        description:
+            - "The amount of time the module should wait for the instance to
```
mykaul: the amount of time in seconds

lib/ansible/utils/module_docs_fragments/ovirt.py
```diff
@@ -0,0 +1,56 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+class ModuleDocFragment(object):
+
+    # Standard oVirt documentation fragment
+    DOCUMENTATION = '''
+options:
+    wait:
```
mykaul, line 27: Is the order of the options important? I think it should be alphabetically or by priority - wait is not the 1st one in either cases, I reckon.

lib/ansible/module_utils/ovirt.py
```diff
@@ -0,0 +1,536 @@
...
+
+    def search_entity(self, search_params=None):
+        """
+        Always first try to search by `ID`, if ID isn't specified,
+        check if user constructed special search in `search_params`,
+        if not search by `name`.
+        """
+        entity = None
+
+        if 'id' in self._module.params and self._module.params['id'] is not None:
```
mykaul, line 558: I would actually assume most of the users will not be using ID but name. Don't think it matters that much though.

lib/ansible/module_utils/ovirt.py
```diff
@@ -0,0 +1,536 @@
...
+        specified by passing `action_condition` parameter. Which state the entity
+        should have after executing the action can be specified by passing
+        `wait_condition` parameter.
+
+        Additional task can be executed before action on entity is executed
+        by passing `pre_action` parameter to method. Another task can be
+        executed after action is executed, by passing `post_action` parameter
+        to method.
+
+        :param action: Action which should be executed on service.
```
mykaul: On service or on entity?

lib/ansible/module_utils/ovirt.py
```diff
@@ -0,0 +1,536 @@
...
+        :param entity: Entity we want to run action on.
+        :param action_condition: Function which is executed when checking if action should be executed.
+        :param wait_condition: Function which is executed when waiting on result state.
```
mykaul: I wondering if we should add an optional 'fail' condition as well. For example, suppose you install a host, and wish to wait to 'up'. If it is in 'install_failed' you don't need to wait to the timeout. This is quite common in many scenarios I found. (Or is the expectations this is going to be part of the wait_condition?)

lib/ansible/module_utils/ovirt.py
machacekondra: It's actually by service on entity, thanks.

lib/ansible/module_utils/ovirt.py
machacekondra: That's something I didn't think about, it's great idea, the current flow wouldn't work well enough if it would be part of the _wait_condition_. I will definitelly add it, thanks.

lib/ansible/module_utils/ovirt.py
machacekondra: The modules always return the ID, so they can use IDs which are returned from modules, I believe someone will use it.

lib/ansible/utils/module_docs_fragments/ovirt.py
machacekondra: Well, the generated [documentation](http://docs.ansible.com/ansible/modules_by_category.html) sorts the parameters alphabetically and it doesn't matter how they are sorted in this yaml documention. We can use some sorting rule, but in the  result documentation it doesn't matter.

lib/ansible/module_utils/ovirt.py
```diff
@@ -0,0 +1,545 @@
...
+
+    @changed.setter
+    def changed(self, changed):
+        if not self._changed:
+            self._changed = changed
+
+    @abstractmethod
+    def build_entity(self):
+        """
+        This method should return oVirt Python SDK type, which we want want
```
mwperina: we want want -> we want

lib/ansible/module_utils/ovirt.py
```diff
@@ -0,0 +1,545 @@
...
+        pre_action=lambda e: e,
+        post_action=lambda e: None,
+        search_params=None,
+        **kwargs
+    ):
+        """
+        This method is executed when we want to change the state of some oVirt
+        entity. The action to be executed on oVirt service is specified by
+        `action` parameter. Whether the action should be executed can be
+        specified by passing `action_condition` parameter. Which state the entity
```
mwperina: Which state the entity should have after executing the action can be specified by passing `wait_condition` parameter.

-> 

State which the entity should be in after execution of the action can be specified by `wait_condition` parameter.

lib/ansible/module_utils/ovirt.py
```diff
@@ -0,0 +1,545 @@
...
+        should have after executing the action can be specified by passing
+        `wait_condition` parameter.
+
+        Additional task can be executed before action on entity is executed
+        by passing `pre_action` parameter to method. Another task can be
+        executed after action is executed, by passing `post_action` parameter
+        to method.
```
mwperina: Additional task can be executed before action on entity is executed by passing `pre_action` parameter to method. Another task can be  executed after action is executed, by passing `post_action` parameter  to method.

->

Task executed before an action on entity can optionally be specified in `pre_action` parameter. Task executed after an action on entity can optionally be specified in `post_action` parameter.

mwperina:
Looks good to me, thanks. :+1:

lib/ansible/module_utils/ovirt.py
```diff
@@ -0,0 +1,544 @@
...
+    # Get rid of whitespaces:
+    param = ''.join(param.split())
+
+    # Convert to bytes:
+    if param[-3].lower() in ['k', 'm', 'g', 't', 'p']:
+        return int(param[:-3]) * BYTES_MAP.get(param[-3:].lower(), 1)
+    elif param.isdigit():
+        return int(param) * 2**10
+    else:
+        raise Exception(
```
lukas-bednar: I would  rather see ValueError instead of general exception

mwperina (commented):
:+1:

machacekondra:
Shippable failed due to some infra issue and I don't see any button to re-run.

gundalow:
The Shippable failure was in Windows (which this PR doesn't touch) I've clicked rerun..

I believe anyone can rerun jobs that are triggered by their own PRs. You need to sign into Travis then click "rerun" on the top of the job page. This will rerun all jobs started by this PR, not just the individual failing job. This is a limitation of Shippable.

machacekondra:
ready_for_review

lib/ansible/module_utils/ovirt.py
```diff
@@ -0,0 +1,544 @@
...
+    'gib': 2**30,
+    'tib': 2**40,
+    'pib': 2**50,
+}
+
+
+def check_sdk(module):
+    if not HAS_SDK:
+        module.fail_json(
+            msg='ovirtsdk4 version 4.0.0 or higher is required for this module'
```
ryansb, line 51: Great, thanks for putting this check in here.

lib/ansible/module_utils/ovirt.py
```diff
@@ -0,0 +1,544 @@
...
+    """
+    def remove_underscore(val):
+        if val.startswith('_'):
+            val = val[1:]
+            remove_underscore(val)
+        return val
+
+    res = {}
+    if struct is not None:
+        for key, value in struct.__dict__.iteritems():
```
ryansb: For compatibility we ask folks to use `.items()` since it works in Python 2.4 up to Python 3.x. If you _must_ use iteritems please use `six.iteritems()`.

lib/ansible/module_utils/ovirt.py
```diff
@@ -0,0 +1,544 @@
...
+                        res[key].append(i)
+            else:
+                res[key] = value
+
+    return res
+
+
+def create_connection(auth):
+    """
+    Create a connection to Python SDK, from task `auth` parameter.
```
ryansb, line 92: In the pydoc for this function can you please add info about what keys in the dictionary are required?

lib/ansible/module_utils/ovirt.py
```diff
@@ -0,0 +1,544 @@
...
+    :param service: service of the entity
+    :param condition: condition to be fulfilled
+    :param fail_condition: if this condition is true, raise Exception
+    :param timeout: max time to wait in seconds
+    :param wait: if True wait for condition, if False don't wait
+    """
+    # Wait until the desired state of the entity:
+    if wait:
+        start = time.time()
+        while time.time() < start + timeout:
```
ryansb, line 257: It'd be good to include a default sleep here so this loop doesn't continuously hit the oVirt endpoint. Even just a one or two second sleep would be fine between tries.

lib/ansible/module_utils/ovirt.py
```diff
@@ -0,0 +1,544 @@
...
+    )
+    spec.update(kwargs)
+    return spec
+
+
+def check_params(module):
+    """
+    Most modules must have either `name` or `id` specified.
+    """
+    if module.params['name'] is None and module.params['id'] is None:
```
ryansb: Just in case it's probably a good idea to use `.get()` instead of dict access syntax in case the key is unset.

lib/ansible/module_utils/ovirt.py
machacekondra: Done

lib/ansible/module_utils/ovirt.py
machacekondra: Done

lib/ansible/module_utils/ovirt.py
machacekondra: Done

lib/ansible/module_utils/ovirt.py
machacekondra: Done

PR #17371, opened 2016-09-02
Update intro_installation.rst

mbigras:
##### ISSUE TYPE

- Docs Pull Request
##### COMPONENT NAME

Installation documentation
##### ANSIBLE VERSION

```
Ansible’s “raw” module (for executing commands in a quick and dirty way) and the script module don’t even need that. 
```
##### SUMMARY

Minor language change 😄 don't -> doesn't

```
Ansible’s “raw” module (for executing commands in a quick and dirty way) and the script module doesn't even need that. 
```

### DIFF
--- a/docsite/rst/intro_installation.rst
+++ b/docsite/rst/intro_installation.rst
@@ -72,7 +72,7 @@ You also need Python 2.4 or later. If you are running less than Python 2.5 on th
 .. note::
 
    Ansible's "raw" module (for executing commands in a quick and dirty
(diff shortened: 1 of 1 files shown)


bcoca:
I believe "don't" is correct here as it applies to both raw and script modules.

docschick:
Agreed.   Since this refers to both the raw and script modules, "don't" is correct.

dharmabumstead:
Thanks, but the wording is correct as is.

PR #17381, opened 2016-09-03
Fix pipelining with copy module.

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

lib/ansible/plugins/action/copy.py
##### ANSIBLE VERSION

```
stbale-2.1
```
##### SUMMARY

The fix to always clean temp directories broke pipelining with
the copy module.  This reverts part of the change in
a5d79a39d51a256937d6e5742befa169629b7caf

### DIFF
--- a/lib/ansible/plugins/action/copy.py
+++ b/lib/ansible/plugins/action/copy.py
@@ -247,7 +247,7 @@ class ActionModule(ActionBase):
                 if 'content' in new_module_args:
                     del new_module_args['content']
 
-                module_return = self._execute_module(module_name='copy', module_args=new_module_args, task_vars=task_vars, tmp=tmp, delete_remote_tmp=delete_remote_tmp)
+                module_return = self._execute_module(module_name='copy', module_args=new_module_args, task_vars=task_vars, delete_remote_tmp=delete_remote_tmp)
                 module_executed = True
 
             else:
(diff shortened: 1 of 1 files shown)


abadger:
Playbook to see the problem:

```
- hosts: all
  tasks:
  - copy:
      src: /etc/os-release
      dest: "/var/tmp/r"
```

And invocation:

```
ANSIBLE_SSH_PIPELINING=1 ansible-playbook -i '127.0.0.2' -c ssh test.yml -vvvv
```

When broken, you'll see that the copy module is sftp'd to the remote machine despite pipelining.  After copying, the module is still invoked remotely by being piped to the stdin of python on the remote machine.

abadger:
Talked with @bcoca, and _execute_module() does need to get the tmp directories so we need to change _execute_module() to figure out whether to send the module via the pipelining flag, not whether tmpdir exists or not.

Also, the tmp cleanup change is unoptimizing the removal of the directory in copy (requiring an additional connection to the remote machine rather than doing it as  part of the last call to execute_module).  We should look into fixing that as well, perhaps for 2.3.  (bcoca and I agree that the copy code is very arcane in this area... Could do with some simplification as we work on fixing the unoptimization).

PR #17390, opened 2016-09-05
Add public_ip option to linode dynamic inventory

Jared-Sprague:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

linode dynamic inventory
##### ANSIBLE VERSION

```
ansible 2.1.1.0
```
##### SUMMARY

Currently the linode dynamic inventory in contrib/inventory/linode.py assumes that your linode labels match a resolvable hostname.  This patch adds the option to use the linodes public IP instead of the linode label.   This is convenient for linode users who have many linodes that don't have labels set that match hostnames, but public IPs will always be resolvable no matter what the linode label is set to. 

This patch does not change default behavior but just adds an additional config parameter in linode.ini called `use_public_ip` that is set to false by default.  Setting it to true changes the output to public IP instead of label.  See the below output. 

### Before

```
{
  "2196623": [
    "node-dallas-1"
  ], 
  "dallas": [
    "node-dallas-1", 
  ], 
  "nodes-dallas": [
    "node-dallas-1"
  ]
}
```
### After with use_public_ip = true

```
{
  "2196623": [
    "45.33.11.123"
  ], 
  "dallas": [
    "45.33.11.123", 
  ], 
  "nodes-dallas": [
    "45.33.11.123"
  ]
}
```

### DIFF
--- a/contrib/inventory/linode.ini
+++ b/contrib/inventory/linode.ini
@@ -15,4 +15,4 @@ cache_path = /tmp
 cache_max_age = 300
 
 # If set to true use the hosts public ip in the dictionary instead of the label
-use_public_ip = true
+use_public_ip = false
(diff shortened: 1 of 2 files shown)


Jared-Sprague:
@bcoca check it out

PR #17404, opened 2016-09-06
Fix list separator typo

asolano:
##### ISSUE TYPE

 Docs Pull Request
##### COMPONENT NAME

http://docs.ansible.com/ansible/intro_patterns.html
##### ANSIBLE VERSION

Not relevant.
##### SUMMARY

The ':' (colon) had turned into a ';' (semicolon).


bcoca:
no typo, ':' is still valid, ';' is deprecated

asolano:
I see. There is no other mention of  ';' in the page, so maybe an additional comment about it would help newcomers.

bcoca:
it was an undocumented separator we are going to remove, that is why it is a deprecation notice. I don't think we are going to document it at this point.

asolano:
Good to know, thank you for the explanation.

Issue #13981, opened 2016-01-19
atomic_move inside docker fails for /etc/hosts


pgrange:
When running ansible inside a docker container, ansible will fail if we try to change some systems file, such as /etc/hosts.

This issue is related to this docker issue : https://github.com/docker/docker/issues/9295

You can reproduce with the following code : https://github.com/pgrange/ansible_docker_resource_busy

I wrote an uggly, untested, patch just to validate an alternative that would consist of overwriting the content of the target file instead of renaming or copying. This patch is available here : https://github.com/pgrange/ansible/commit/844eb0bfea04473719e195fcad73a515516f1df6

bcoca:
Basically docker says to not do this, to use their command line options.

bcoca:
@pgrange I'm not liking the solution ... but it has nothing to do with your code, its the artificial limitation added at the source ...

I would consider the patch, but if we add an update_atomically=yes|no option to trigger it.

pgrange:
@bcoca sorry for the delay and thank you for your answer.

I forgot to explain my use case. First of all, I do not use ansible to configure a docker container. But I use a docker container to validate an ansible role before applying it on actual servers.

The idea here is to write ansible roles in a kind of TDD way. Running the role on the container, then asserting some states on this container to ensure role behaves as expected.

Is docker appropriate to do that is an open question but the fact is, exept for this issue, it is working quite well.

I make this precision so that you can understand that I do not want to tweak my ansible role so that it can run inside a docker container. Since the container is just a mean to test the role that will be executed on actual production servers.

I need some clarification on your suggestion. Do you suggest to introduce a global configuration option "update_atomically=yes|no" that one can set in a ansible.cfg file for instance and that will apply to every module that relies on update_atomically ? Or do your suggest to add such an option directly to the lineinfile module for instance ?

The first option could solve my issue. The second, not so much. Let me know which option you suggest.

Just so you know, I took inspiration in the patch that has been added to handle errno 26 (text file busy) which happens on vagrant synced folders:

https://github.com/ansible/ansible/commit/013c4631e3a65035471d85aabd9227c0fa701e10

And I think both issues compare quite well. Here, we introduce a patch to deal with errno.EBUSY for docker instead of errno 26 for vagrant. OK, the way we work around the issue is different (and the patch is crap ;)).

Thanks,
Pascal.

bcoca:
[...]

PR #981, opened 2012-08-31
Fix supervisorctl module after merge issue

mattupstate:

PR #17405, opened 2016-09-06
fixes issue #13981: unsafe_writes block appeared too late in the atom…

pgrange:
##### ISSUE TYPE

- Feature Pull Request
- New Module Pull Request
- Bugfix Pull Request
- Docs Pull Request
##### ANSIBLE VERSION

```

```
##### SUMMARY

```

```

…ic_move

workflow. This led to errno.EBUSY to not be managed in the context of
issue #!#981

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -1941,24 +1941,7 @@ class AnsibleModule(object):
             os.rename(b_src, b_dest)
         except (IOError, OSError):
             e = get_exception()
-            # sadly there are some situations where we cannot ensure atomicity, but only if
-            # the user insists and we get the appropriate error we update the file unsafely
-            if unsafe_writes and e.errno == errno.EBUSY:
-                #TODO: issue warning that this is an unsafe operation, but doing it cause user insists
-                try:
-                    try:
-                        out_dest = open(b_dest, 'wb')
-                        in_src = open(b_src, 'rb')
-                        shutil.copyfileobj(in_src, out_dest)
-                    finally: # assuring closed files in 2.4 compatible way
-                        if out_dest:
-                            out_dest.close()
-                        if in_src:
-                            in_src.close()
-                except (shutil.Error, OSError, IOError):
-                    e = get_exception()
-                    self.fail_json(msg='Could not write data to file (%s) from (%s): %s' % (dest, src, e))
-            elif e.errno not in [errno.EPERM, errno.EXDEV, errno.EACCES, errno.ETXTBSY]:
+            if e.errno not in [errno.EPERM, errno.EXDEV, errno.EACCES, errno.ETXTBSY, errno.EBUSY]:
                 # only try workarounds for errno 18 (cross device), 1 (not permitted),  13 (permission denied)
                 # and 26 (text file busy) which happens on vagrant synced folders and other 'exotic' non posix file systems
                 self.fail_json(msg='Could not replace file: %s to %s: %s' % (src, dest, e))
@@ -2003,7 +1986,26 @@ class AnsibleModule(object):
                         os.rename(b_tmp_dest_name, b_dest)
                     except (shutil.Error, OSError, IOError):
                         e = get_exception()
-			self.fail_json(msg='Could not replace file: %s to %s: %s' % (src, dest, e))
+                        # sadly there are some situations where we cannot ensure atomicity, but only if
+                        # the user insists and we get the appropriate error we update the file unsafely
+                        if unsafe_writes and e.errno == errno.EBUSY:
+                            #TODO: issue warning that this is an unsafe operation, but doing it cause user insists
+                            try:
+                                try:
+                                    out_dest = open(b_dest, 'wb')
+                                    in_src = open(b_tmp_dest_name, 'rb')
+                                    shutil.copyfileobj(in_src, out_dest)
+                                finally: # assuring closed files in 2.4 compatible way
+                                    if out_dest:
+                                        out_dest.close()
+                                    if in_src:
+                                        in_src.close()
+                            except (shutil.Error, OSError, IOError):
+                                e = get_exception()
+                                self.fail_json(msg='Could not write data to file (%s) from (%s): %s' % (dest, src, e))
+
+                        else:
+                            self.fail_json(msg='Could not replace file: %s to %s: %s' % (src, dest, e))
                 finally:
                     self.cleanup(b_tmp_dest_name)
 
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -1983,29 +1983,12 @@ class AnsibleModule(object):
                             e = get_exception()
                             if e.errno != errno.EPERM:
                                 raise
-                        os.rename(b_tmp_dest_name, b_dest)
+                        try:
+                            os.rename(b_tmp_dest_name, b_dest)
+                        except (shutil.Error, OSError, IOError):
+                            self._unsafe_writes(unsafe_writes, b_tmp_dest_name, b_dest, get_exception())
                     except (shutil.Error, OSError, IOError):
-                        e = get_exception()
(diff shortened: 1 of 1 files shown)


bcoca:
not sure what you are trying to fix here, unsafe writes should be 'last resort', not a top option

pgrange:
Sorry, I did not format my commit comment right for github.

This is related to issue #13981

You are right, unsafe_write should be last resort. However with current code the unsafe_write block is never executed in the context of issue #13981.

Maybe you might prefer another version of the patch but with current code, when we encounter a problem with atomic move, if it is not one of [errno.EPERM, errno.EXDEV, errno.EACCES, errno.ETXTBSY] then we fail.

What I need is to not fail if it is a errno.EBUSY so that unsafe_writes is managed.

pgrange:
I updated this repo to reproduce the corresponding issue : 
https://github.com/pgrange/ansible_docker_resource_busy

pgrange:
Hi @bcoca,

I updated this pull request to minimize changes. I think it clarifies the issue by changing just two lines.

Let me know if you prefer closing this pull request so that I open a new cleaner one.

Regards,

lib/ansible/module_utils/basic.py
```diff
@@ -1941,7 +1941,7 @@ def atomic_move(self, src, dest, unsafe_writes=False):
             os.rename(b_src, b_dest)
         except (IOError, OSError):
             e = get_exception()
-            if e.errno not in [errno.EPERM, errno.EXDEV, errno.EACCES, errno.ETXTBSY]:
+            if e.errno not in [errno.EPERM, errno.EXDEV, errno.EACCES, errno.ETXTBSY, errno.EBUSY]:
```
bcoca, line 1944: so this makes more sense now

lib/ansible/module_utils/basic.py
```diff
@@ -1993,7 +1993,7 @@ def atomic_move(self, src, dest, unsafe_writes=False):
                             try:
                                 try:
                                     out_dest = open(b_dest, 'wb')
-                                    in_src = open(b_src, 'rb')
+                                    in_src = open(b_tmp_dest_name, 'rb')
```
bcoca: are you sure this is what we want? b_src should be what we 'tried to copy to b_tmp_dest_name'

pgrange: Not sure why but at this point in code, b_src refers to a non existing file when b_tmp_dest_name refers to a file that exists and has correct data inside.

When running with b_src we have this error:

```
TASK [insert 127.0.0.2 line in /etc/hosts] *************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: UnboundLocalError: local variable 'in_src' referenced before assignment
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "module_stderr": "Traceback (most recent call last):\n  File \"/tmp/ansible_WvGe4f/ansible_module_lineinfile.py\", line 425, in <module>\n    main()\n  File \"/tmp/ansible_WvGe4f/ansible_module_lineinfile.py\", line 417, in main\n    ins_aft, ins_bef, create, backup, backrefs)\n  File \"/tmp/ansible_WvGe4f/ansible_module_lineinfile.py\", line 298, in present\n    write_changes(module, b_lines, dest)\n  File \"/tmp/ansible_WvGe4f/ansible_module_lineinfile.py\", line 175, in write_changes\n    unsafe_writes=module.params['unsafe_writes'])\n  File \"/tmp/ansible_WvGe4f/ansible_modlib.zip/ansible/module_utils/basic.py\", line 2002, in atomic_move\nUnboundLocalError: local variable 'in_src' referenced before assignment\n", "module_stdout": "", "msg": "MODULE FAILURE"}
        to retry, use: --limit @/tmp/ansible/etc_host.retry
```

This is an error raised by the incorrect handling of file not found exception. Removing this line handler (line 2002) we can get to the actual issue:

```
TASK [insert 127.0.0.2 line in /etc/hosts] *************************************
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "Could not write data to file (/etc/hosts) from (/tmp/tmpll43Kj): [Errno 2] No such file or directory: '/tmp/tmpll43Kj'"}
        to retry, use: --limit @/tmp/ansible/etc_host.retry
```

pgrange: OK, now it is clearer. I am sure I want b_tmp_dest_name and not b_src because the exception is raised by os.rename(b_tmp_dest_name, b_dest) line 1987.

So when we catch the exception and try to unsafe_write, we just failed to os-rename b_tmp_dest_name so we must unsafe_write b_tmp_dest_name

bcoca: no, the exception is around this code:
`os.rename(b_src, b_dest)`

as the original exception on this 'downward path', checking if we are improperly resetting b_src between that and this line

bcoca: so this looks like a problem `shutil.move(b_src, b_tmp_dest_name)`

bcoca: but it can also fail here `shutil.copy2(b_src, b_tmp_dest_name)`, so we might just need a flag on 'which to use'

abadger: You could either do an os.path.exists check here to decide based on whether b_src exists or you could abstract the unsafe_writes fallback into a helper method, and then write two try: excepts that target each section of code.  If the code fails in  shutil.copy2 pr shutil.move then call the helper function with b_src.  If it fails in os.rename then call the helper with b_tmp_dest_name.  atomic_move is getting rather long so the latter may be the better option.

abadger (changes requested):

pgrange:
@abadger I extracted unsafe_writes code as suggested.
I focused the try/except code that relies on _unsafe_writes() around os.rename code.
However I did not try/except around shutil.copy2 and shutil.move since they are writing to b_tmp_dest_name and no problem can occur when writing to b_tmp_dest_name... no problem that _unsafe_writes can solve, actually.
See #13981 for details about the problem that _unsafe_writes tries to solve.

Regards,

lib/ansible/module_utils/basic.py
```diff
@@ -2022,6 +2005,28 @@ def atomic_move(self, src, dest, unsafe_writes=False):
             # rename might not preserve context
             self.set_context_if_different(dest, context, False)
 
+    def _unsafe_writes(self, unsafe_writes, src, dest, exception):
+      # sadly there are some situations where we cannot ensure atomicity, but only if
+      # the user insists and we get the appropriate error we update the file unsafely
+      if unsafe_writes and exception.errno == errno.EBUSY:
```
abadger: I'd do this checking in the caller.  That way the function call reads as "Do an unsafe write".  As it's written now it's more of a "I think we should do an unsafe_write.  Do you agree?"

abadger:
Only one comment.  Rest looks good to me.  I'll give @bcoca a chance to reply in case he sees something else.

pgrange:
@abadger agree. New commit checks if unsafe_writes in the caller.

abadger (approved):

lib/ansible/module_utils/basic.py
```diff
@@ -1983,29 +1983,15 @@ def atomic_move(self, src, dest, unsafe_writes=False):
                             e = get_exception()
                             if e.errno != errno.EPERM:
                                 raise
-                        os.rename(b_tmp_dest_name, b_dest)
+                        try:
```
bcoca, line 1986: so this changes the point of coverage of unsafe_writes, it will not be used as fallback except for the narrower error on temp file renaming.

bcoca (changes requested):

lib/ansible/module_utils/basic.py
abadger: The answer to that is in the commit message: "If the code fails in shutil.copy2 or shutil.move this should not be related to issue #13981 since they write to b_tmp_dest_name."

bcoca (approved):

pgrange:
Great. Thanks !

PR #17421, opened 2016-09-06
Allow for no prefix in the comment filter

drybjed:
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

The `comment` filter.
##### ANSIBLE VERSION

```
ansible 2.2.0
```
##### SUMMARY

At the moment, when the `comment` filter is used with `prefix=''` (empty `prefix`), it creates an empty line above the comment. In most cases it's a good behaviour to leave an empty line between sections in the configuration files with comments, however in certain corner cases, like a top of the file, it is annoying and impossible to override either directly in the `comment` parameters, or through Jinja.

This PR changes the behaviour of the

```
{{ "string" | comment(prefix='') }}
```

parameter. Instead of leaving an empty line, nothing  will be prepended to the output. To use the old behaviour, set the parameter as

```
{{ "string" | comment(prefix='\n') }}
```

This will add an empty line before the comment.

This change will affect existing templates that use the `prefix=''` parameter, however an alternative is provided. The change brings the `prefix=` parameter in line with the `beginning=` parameter which works in a similar way, therefore I think it's the correct course of action.

### DIFF
--- a/lib/ansible/plugins/filter/core.py
+++ b/lib/ansible/plugins/filter/core.py
@@ -336,8 +336,14 @@ def comment(text, style='plain', **kw):
     str_beginning = ''
     if p['beginning']:
         str_beginning = "%s%s" % (p['beginning'], p['newline'])
-    str_prefix = str(
-        "%s%s" % (p['prefix'], p['newline'])) * int(p['prefix_count'])
+    str_prefix = ''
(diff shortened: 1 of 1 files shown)


drybjed:
Awesome, thanks!

PR #17429, opened 2016-09-07
Clean up unarchive action plugin wrt. creates=

dagwieers:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

unarchive
##### SUMMARY

This ensures the unarchive action-plugin is using the existing framework to test for remote files. We have some issues (ansible/ansible-modules-core#913 and ansible/ansible-modules-core#4164) reported related to **creates=** and maybe this fixes those.

### DIFF
--- a/lib/ansible/plugins/action/unarchive.py
+++ b/lib/ansible/plugins/action/unarchive.py
@@ -66,9 +66,7 @@ class ActionModule(ActionBase):
             # do not run the command if the line contains creates=filename
             # and the filename already exists. This allows idempotence
             # of command executions.
-            result = self._execute_module(module_name='stat', module_args=dict(path=creates), task_vars=task_vars)
-            stat = result.get('stat', None)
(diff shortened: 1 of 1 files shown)


lib/ansible/plugins/action/unarchive.py
```diff
@@ -66,9 +66,7 @@ def run(self, tmp=None, task_vars=None):
             # do not run the command if the line contains creates=filename
             # and the filename already exists. This allows idempotence
             # of command executions.
-            result = self._execute_module(module_name='stat', module_args=dict(path=creates), task_vars=task_vars)
-            stat = result.get('stat', None)
-            if stat and stat.get('exists', False):
+            if self._remote_file_exists(creates):
```
bcoca, line 69: This code should not change it working or not, but it is more light weight and efficient.

PR #17433, opened 2016-09-07
Show only reference title in modules

asolano:
##### ISSUE TYPE

Docs Pull Request
##### COMPONENT NAME

http://docs.ansible.com/ansible/intro_adhoc.html
##### ANSIBLE VERSION

Not relevant
##### SUMMARY

The :ref: in the page is being expanded to include the description of the module. It makes the sentence hard to read.

Not exactly familiar with rst syntax, but according to http://www.sphinx-doc.org/en/stable/markup/inline.html#ref-role a label can be specified.


dharmabumstead:
Stale - closing.

PR #17457, opened 2016-09-08
Pass the absolute path to dirname when assigning basedir

### LINKED ISSUE #17456: Ansible fails to create retry files with [Errno 2] No such file or directory: ''
##### ISSUE TYPE
- Bug Report
##### COMPONENT NAME

retry files
##### ANSIBLE VERSION

```
ansible 2.1.1.0
  config file = 
```
##### CONFIGURATION

none
##### OS / ENVIRONMENT

Ubuntu 16.04
##### SUMMARY

When a playbook fails, Ansible tries to create a retry file and then fails.
##### STEPS TO REPRODUCE

```
# test.yml

---
- hosts: localhost
  gather_facts: no
  tasks:
    - command: "false"
```

Run `ansible-playbook -i localhost, -c local test.yml`
##### EXPECTED RESULTS

Playbook fails, ansible doesn't complain about failing to create `test.retry`.
##### ACTUAL RESULTS

```

fatal: [localhost]: FAILED! => {"changed": true, "cmd": ["false"], "delta": "0:00:00.001666", "end": "2016-09-08 11:42:55.135782", "failed": true, "rc": 1, "start": "2016-09-08 11:42:55.134116", "stderr": "", "stdout": "", "stdout_lines": [], "warnings": []}

 [WARNING]: Could not create retry file 'test.retry'.         [Errno 2] No such file or directory: ''


```

Shaps:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

`PlaybookExecutor`
##### ANSIBLE VERSION

```
ansible 2.1.1.0
  config file =
```
##### SUMMARY

As reported in #17456 ansible-playbook fails to create a retry file. This happens when a playbook is called without passing a path ( `ansible-playbook test.yml` ). 
The basedir created with `os.path.dirname(playbook_path` will return `''` which will, cause `os.path.basename` to return `''` ending with an empty string as filename. 
This will cause  the error later in `_generate_retry_inventory` to write the file as the `retry_path` filename will be empty. 

```

ok: [localhost]

fatal: [localhost]: FAILED! => {"changed": true, "cmd": ["false"], "delta": "0:00:00.001884", "end": "2016-09-08 10:20:40.023480", "failed": true, "rc": 1, "start": "2016-09-08 10:20:40.021596", "stderr": "", "stdout": "", "stdout_lines": [], "warnings": []}

 [WARNING]: Could not create retry file 'test.retry'.         [Errno 2] No such file or directory: ''


```

```

ok: [localhost]

fatal: [localhost]: FAILED! => {"changed": true, "cmd": ["false"], "delta": "0:00:00.002157", "end": "2016-09-08 10:21:38.494251", "failed": true, "rc": 1, "start": "2016-09-08 10:21:38.492094", "stderr": "", "stdout": "", "stdout_lines": [], "warnings": []}

/root/test.retry
        to retry, use: --limit @/root/test.retry

```

Fixes #17456

### DIFF
--- a/lib/ansible/executor/playbook_executor.py
+++ b/lib/ansible/executor/playbook_executor.py
@@ -186,7 +186,7 @@ class PlaybookExecutor:
                             if C.RETRY_FILES_SAVE_PATH:
                                 basedir = C.shell_expand(C.RETRY_FILES_SAVE_PATH)
                             elif playbook_path:
-                                basedir = os.path.dirname(playbook_path)
+                                basedir = os.path.dirname(os.path.realpath(playbook_path))
                             else:
                                 basedir = '~/'
 
--- a/lib/ansible/executor/playbook_executor.py
+++ b/lib/ansible/executor/playbook_executor.py
@@ -186,7 +186,7 @@ class PlaybookExecutor:
                             if C.RETRY_FILES_SAVE_PATH:
                                 basedir = C.shell_expand(C.RETRY_FILES_SAVE_PATH)
                             elif playbook_path:
-                                basedir = os.path.dirname(os.path.realpath(playbook_path))
+                                basedir = os.path.dirname(os.path.abspath(playbook_path))
                             else:
                                 basedir = '~/'
 


lib/ansible/executor/playbook_executor.py
```diff
@@ -186,7 +186,7 @@ def run(self):
                             if C.RETRY_FILES_SAVE_PATH:
                                 basedir = C.shell_expand(C.RETRY_FILES_SAVE_PATH)
                             elif playbook_path:
-                                basedir = os.path.dirname(playbook_path)
+                                basedir = os.path.dirname(os.path.realpath(playbook_path))
```
mgedmin: Hm.  If the playbook is a symlink (playbooks/mybook.yml -> ../elsewhere/mybook.yml), then this will create the retry files in elsewhere/ rather than in playbooks/.  Is that desirable?

mgedmin: I suggest `os.path.normpath(os.path.dirname(playbook_path))` since `os.path.normpath('') == '.'`.

abadger: I suggested using `os.path.abspath` on IRC.  shaps thought maybe: `os.path.dirname(os.path.abspath(playbook_path))`

Shaps: Also, @bcoca was suggesting to replace that with `unfrackpath`

abadger: `unfrakpath()` won't work because it calls `os.path.realpath()` on the passed in path which is what we need to avoid.  At least in the cases that are failing here, tilde and vars are already substituted by the shell before it is given to ansible-playbook on the commandline so we don't need os.path.expandvars or os.path.expanduser (but they don't hurt either).

bcoca: shell would not be involved if set in ansible.cfg

abadger: Playbook path is set in ansible.cfg?

abadger:
Merged to devel and stable-2.1.  Whether it gets into 2.1.2 depends on whether we have an rc4 or go to 2.1.2 final.  Thanks for the bug report and the fix!

PR #17458, opened 2016-09-08
Re-fix the per-item diff output

towolf:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (refix-diff-for-task-items 534ae73265) last updated 2016/09/08 12:33:33 (GMT +200)
  lib/ansible/modules/core: (devel 1d48b47cad) last updated 2016/09/01 09:55:33 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD 2ef4a34eee) last updated 2016/08/31 16:07:04 (GMT +200)
  config file = /home/towolf/src/ansible/users/tobias.wolf/ansible.cfg
  configured module search path = Default w/o overrides
```
##### SUMMARY

**Re-fix the per-item diff output**

which got lost in recent big 'performance improvements' merge by @jimi-c.

I had made a previous PR to fix this, then @bcoca had committed an
improved fix. Now it's lost again.

cf: d2b3b2c03e2934b126f5701e5f6e25821e2dbe35 (lost here)
cf:  25e9b5788bb2c43743a06f877f34c2336d19338b (previous fix)

Earlier PR #14849
Earlier issue #14843

Please note that jimi-c broke this last time as well ... seeing a
pattern here.

PR #17459, opened 2016-09-08
Re-fix diff display in the `actionable` stdout callback

towolf:
##### ISSUE TYPE

- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.2.0 (temp e6deae7abf) last updated 2016/09/08 15:05:15 (GMT +200)
  lib/ansible/modules/core: (devel 1d48b47cad) last updated 2016/09/01 09:55:33 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD 2ef4a34eee) last updated 2016/08/31 16:07:04 (GMT +200)
  config file = /home/towolf/src/ansible/users/tobias.wolf/ansible.cfg

```
##### SUMMARY

**Regarding the actionable plugin**

The diff was displayed before the TASK banner for for itemized and
unitemized tasks:
1. diff
2. banner
3. item_on_ok

I removed the loopwise handling of diff in the default callback and
handle the on_file_diff callback in the actionable callback plugin itself.

Now all diff display correctly.

This PR requires the previous PR #17458

Please review the logic for this.

### DIFF
--- a/lib/ansible/plugins/callback/actionable.py
+++ b/lib/ansible/plugins/callback/actionable.py
@@ -32,6 +32,7 @@ class CallbackModule(CallbackModule_default):
         self.super_ref.__init__()
         self.last_task = None
         self.shown_title = False
+        self.diff = False
 
     def v2_playbook_on_handler_task_start(self, task):
         self.super_ref.v2_playbook_on_handler_task_start(task)
@@ -40,6 +41,7 @@ class CallbackModule(CallbackModule_default):
     def v2_playbook_on_task_start(self, task, is_conditional):
         self.last_task = task
         self.shown_title = False
+        self.diff = False
 
     def display_task_banner(self):
         if not self.shown_title:
@@ -53,6 +55,8 @@ class CallbackModule(CallbackModule_default):
     def v2_runner_on_ok(self, result):
         if result._result.get('changed', False):
             self.display_task_banner()
+            if self.diff:
+                self.super_ref.v2_on_file_diff(result)
             self.super_ref.v2_runner_on_ok(result)
 
     def v2_runner_on_unreachable(self, result):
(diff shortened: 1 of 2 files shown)


towolf:
Would be great if this could be looked at before 2.2. @bcoca?

We tested this PR with the following stdout callbacks and it all looks proper:
- skippy
- minimal
- json
- debug
- oneline

lib/ansible/plugins/callback/default.py
```diff
@@ -145,13 +145,7 @@ def v2_playbook_on_play_start(self, play):
...
 
     def v2_on_file_diff(self, result):
-        if result._task.loop and 'results' in result._result:
-            for res in result._result['results']:
-                if 'diff' in res and res['diff'] and res.get('changed', False):
-                    diff = self._get_diff(res['diff'])
-                    if diff:
-                        self._display.display(diff)
-        elif 'diff' in result._result and result._result['diff'] and result._result.get('changed', False):
+        if 'diff' in result._result and result._result['diff'] and result._result.get('changed', False):
```
bcoca, line 148: this breaks showing diffs from loop items

lib/ansible/plugins/callback/actionable.py
```diff
@@ -32,6 +32,7 @@ def __init__(self):
         self.super_ref.__init__()
         self.last_task = None
         self.shown_title = False
+        self.diff = False
```
bcoca, line 35: the correct fix here is changing super to `CallbackModule_default` instead of `CallbackModule`, then the `self.diff` and any future changes in the init will be inherited

bcoca (changes requested):

lib/ansible/plugins/callback/default.py
towolf: Cannot confirm that. It's working as intended, with my other PR that was merged.

image

image

lib/ansible/plugins/callback/actionable.py
towolf: Not entirely sure, what you mean by that :confused: 

I added the `self.diff` locally to override handling of diffs in this particular callback.

Are you proposing an entirely different and better fix?

towolf:
@bcoca, my colleague tested this at length and he said it works correctly. He tested the other callbacks as well.

bcoca:
The actionable has now been deprecated (in favor of options in the default callback) so the issues with self.diff should be solved there, can you confirm that works for you?

needs_info

towolf:
Yes, closing

PR #17465, opened 2016-09-08
Improve unit testing of 'password' lookup

### LINKED ISSUE #17461: password lookup doesn't always respect chars setting
##### ISSUE TYPE
- Bug Report
##### COMPONENT NAME

password lookup
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  configured module search path = ['/usr/share/ansible/']
```
##### CONFIGURATION

Pretty much default configuration
##### OS / ENVIRONMENT

Ansible server : Ubuntu 14.04.3 LTS
Managed Host : Debian 8.5
##### SUMMARY

Password don't respect specified characters and use punctuation when it shouldn't
##### STEPS TO REPRODUCE

It's a bit random. I have that issue frequently, but not each time.
When I try to reproduce the problem with a test playbook, I couldn't reproduce the issue.
May be I am crazy or something

```
- name: Create replication role
  postgresql_user:
    name: replication
    encrypted: yes
    password: "{{ lookup('password', '/etc/psql_credentials/' + ansible_hostname + ':replication chars=ascii_letters,digits') }}"
    role_attr_flags: LOGIN,REPLICATION
  become_user: postgres
```
##### EXPECTED RESULTS

The newly created file should contain a password with only ascii_letters and digits.
Something like that :

```
root@ans01:# cat /etc/psql_credentials/namindpsql01:replication
yn0P43VUTYsP4Q0kuZHz
root@ans01:# cat /etc/psql_credentials/alllndpsql01:replication
fgZpUPQtypsaCXkPmMb7
```
##### ACTUAL RESULTS

Instead I got this :

```
root@ans01:# cat /etc/psql_credentials/namindpsql01:replication
yn0P.3VUTYsP4Q0k.ZHz
root@ans01:# cat /etc/psql_credentials/alllndpsql01:replication
fgZpUPQ:-psaCXkPmMb7
```

Maybe it's not a bug and my password line is bad, but I can't figure out what I do wrong.
A little help will be much appreciated.

Best regards,

alikins:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

lib/ansible/plugins/lookup/password.py
##### ANSIBLE VERSION

```
ansible 2.2.0 (password_lookup ab5bd4a9eb) last updated 2016/09/08 14:30:44 (GMT -400)
  lib/ansible/modules/core: (detached HEAD db38f0c876) last updated 2016/09/07 12:50:14 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 8bfdcfcab2) last updated 2016/09/07 12:50:14 (GMT -400)

```
##### SUMMARY

Was attempting to reproduce https://github.com/ansible/ansible/issues/17461 with
unit tests, and these were changes to make the module a little easier to test. ( Note
this doesn't fix https://github.com/ansible/ansible/issues/17461)

The tests showed some UnicodeErrors for the
cases where the 'chars' param include unicode,
causing the 'getattr(string, c, c)' to fail.
So the candidate char generation code try/excepts
UnicodeErrors there now.

Some refactoring of the password.py module to make
it easier to test, and some new tests that cover more
of the password and salt generation.

### DIFF
--- a/lib/ansible/plugins/lookup/password.py
+++ b/lib/ansible/plugins/lookup/password.py
@@ -138,7 +138,7 @@ def _gen_candidate_chars(characters):
     chars = u''.join(chars).replace(u'"', u'').replace(u"'", u'')
     return chars
 
-def _create_password_file(b_path):
+def _create_password_file_dir(b_path):
     b_pathdir = os.path.dirname(b_path)
 
     try:
@@ -177,7 +177,7 @@ def _format_content(password, salt, encrypt):
 
     return u'%s salt=%s' % (password, salt)
 
-def _read_or_create_password_file(b_path):
+def _read_or_create_password_file_dir(b_path):
     salt = None
     plaintext_password = None
     content = None
@@ -186,7 +186,7 @@ def _read_or_create_password_file(b_path):
         b_file_content = _read_password_file(b_path)
         content, plaintext_password, salt = _parse_password(b_file_content)
     else:
-        _create_password_file(b_path)
+        _create_password_file_dir(b_path)
 
     return content, plaintext_password, salt
 
@@ -216,7 +216,7 @@ class LookupModule(LookupBase):
             b_path = to_bytes(path, errors='surrogate_or_strict')
 
             # get password or create it if file doesn't exist
-            content, plaintext_password, salt = _read_or_create_password_file(b_path)
+            content, plaintext_password, salt = _read_or_create_password_file_dir(b_path)
             content, plaintext_password, salt = _update_content(plaintext_password, salt, params)
 
             # save it
--- a/lib/ansible/plugins/lookup/password.py
+++ b/lib/ansible/plugins/lookup/password.py
@@ -105,6 +105,35 @@ class LookupModule(LookupBase):
         salt_chars = ascii_letters + digits + './'
         return self.random_password(length=8, chars=salt_chars)
 
+    def gen_candidate_chars(self, chars):
+        candidate_chars_list = []
+        for c in chars:
+            try:
+                defaults = getattr(string, c, c)
+            except UnicodeError:
+                defaults = c
+            candidate_chars_list.append(defaults)
+        candidate_chars = "".join(candidate_chars_list)
+        candidate_chars.replace('"', '')
+        candidate_chars.replace("'", '')
+        return candidate_chars
+
+    def gen_password(self, length, chars):
+        chars = self.gen_candidate_chars(chars)
+        password = ''.join(random.choice(chars) for _ in range(length))
+        return password
+
+    def _build_content_string(self, password, salt):
+        return '%s salt=%s' % (password, salt)
+
+    def gen_content(self, length, chars, encrypt):
+        password = self.gen_password(length, chars)
+        if not encrypt:
+            return password
+
+        salt = self.random_salt()
+        return self._build_content_string(password, salt)
+
     def run(self, terms, variables, **kwargs):
 
         ret = []
@@ -121,14 +150,10 @@ class LookupModule(LookupBase):
                 except OSError as e:
                     raise AnsibleError("cannot create the path for the password lookup: %s (error was %s)" % (pathdir, str(e)))
 
-                chars = "".join(getattr(string, c, c) for c in params['chars']).replace('"', '').replace("'", '')
-                password = ''.join(random.choice(chars) for _ in range(params['length']))
+                content = self.gen_content(length=params['length'],
+                                           chars=params['chars'],
+                                           encrypt=params['encrypt'])
 
-                if params['encrypt'] is not None:
-                    salt = self.random_salt()
-                    content = '%s salt=%s' % (password, salt)
-                else:
-                    content = password
                 with open(path, 'w') as f:
                     os.chmod(path, 0o600)
                     f.write(content + '\n')
@@ -150,7 +175,7 @@ class LookupModule(LookupBase):
                 if params['encrypt'] is not None and salt is None:
                     # crypt requested, add salt if missing
                     salt = self.random_salt()
-                    content = '%s salt=%s' % (password, salt)
+                    content = self._build_content_string(password, salt)
                     with open(path, 'w') as f:
                         os.chmod(path, 0o600)
                         f.write(content + '\n')
--- a/lib/ansible/plugins/lookup/password.py
+++ b/lib/ansible/plugins/lookup/password.py
@@ -32,9 +32,13 @@ from ansible.parsing.splitter import parse_kv
 from ansible.utils.encrypt import do_encrypt
 from ansible.utils.path import makedirs_safe
 
+from ansible.module_utils._text import to_bytes, to_native, to_text
+
 DEFAULT_LENGTH = 20
 VALID_PARAMS = frozenset(('length', 'encrypt', 'chars'))
 
+
+# ALIKINS
 
 def _parse_parameters(term):
     # Hacky parsing of params
@@ -50,7 +54,7 @@ def _parse_parameters(term):
         params = parse_kv(first_split[1])
         if '_raw_params' in params:
             # Spaces in the path?
-            relpath = ' '.join((relpath, params['_raw_params']))
+            relpath = u' '.join((relpath, params['_raw_params']))
             del params['_raw_params']
 
             # Check that we parsed the params correctly
@@ -73,37 +77,60 @@ def _parse_parameters(term):
     params['chars'] = params.get('chars', None)
     if params['chars']:
         tmp_chars = []
-        if ',,' in params['chars']:
+        if u',,' in params['chars']:
             tmp_chars.append(u',')
-        tmp_chars.extend(c for c in params['chars'].replace(',,', ',').split(',') if c)
+        tmp_chars.extend(c for c in params['chars'].replace(u',,', u',').split(u',') if c)
         params['chars'] = tmp_chars
     else:
         # Default chars for password
-        params['chars'] = ['ascii_letters', 'digits', ".,:-_"]
+        params['chars'] = [u'ascii_letters', u'digits', u".,:-_"]
 
     return relpath, params
+
+def _random_password(length=DEFAULT_LENGTH, chars=C.DEFAULT_PASSWORD_CHARS):
+    '''
+    Return a random password string of length containing only chars.
+    NOTE: this was moved from the old ansible utils code, as nothing
+          else appeared to use it.
+    '''
+
+    # hate how python implements asserts. Our text strategy says that we
+    # should never be passing byte strings around unless the function
+    # explicitly takes a byte string as its argument.  Therefore if we get
+    # a byte string here it means that the problem is in the calling code.
+    # Asserts would be perfect for this if they were disabled by default.
+    # Alas, in python, they're active by default; only turned off if
+    # python is run with optimization.
+    # assert isinstance(chars, text_type)
+    random_generator = random.SystemRandom()
+
+    password = []
+    while len(password) < length:
+        new_char = random_generator.choice(chars)
+        if new_char in chars:
+            password.append(new_char)
+
+    return u''.join(password)
+
+def _random_salt():
+    salt_chars = to_text(ascii_letters + digits + './', errors='surrogate_or_strict')
+    return _random_password(length=8, chars=salt_chars)
 
 
 class LookupModule(LookupBase):
 
     def random_password(self, length=DEFAULT_LENGTH, chars=C.DEFAULT_PASSWORD_CHARS):
-        '''
-        Return a random password string of length containing only chars.
-        NOTE: this was moved from the old ansible utils code, as nothing
-              else appeared to use it.
-        '''
-
-        password = []
-        while len(password) < length:
-            new_char = os.urandom(1)
-            if new_char in chars:
-                password.append(new_char)
-
-        return ''.join(password)
+        return _random_password(length=length,
+                                chars=chars)
 
     def random_salt(self):
-        salt_chars = ascii_letters + digits + './'
-        return self.random_password(length=8, chars=salt_chars)
+        return _random_salt()
+
+    def new_gen_candidate_chars(self, characters):
+        chars = []
+        chars.append(to_text(getattr(string, to_native(characters))), errors='surrogate_or_strict')
+        chars = u''.join(to_text(c) for c in chars).replace(u'"', u'').replace(u"'", u'')
+        return chars
 
     def gen_candidate_chars(self, chars):
         candidate_chars_list = []
@@ -129,16 +156,16 @@ class LookupModule(LookupBase):
 
         return '%s salt=%s' % (password, salt)
 
-    def _create_password_file(self, path):
-        pathdir = os.path.dirname(path)
+    def _create_password_file(self, b_path):
+        b_pathdir = os.path.dirname(b_path)
 
         try:
-            makedirs_safe(pathdir, mode=0o700)
+            makedirs_safe(b_pathdir, mode=0o700)
         except OSError as e:
-            raise AnsibleError("cannot create the path for the password lookup: %s (error was %s)" % (pathdir, str(e)))
+            raise AnsibleError("cannot create the path for the password lookup: %s (error was %s)" % (b_pathdir, str(e)))
 
-    def _read_password_file(self, path):
-        content = open(path).read().rstrip()
+    def _read_password_file(self, b_path):
+        content = open(b_path).read().rstrip()
 
         password = content
         salt = None
@@ -154,22 +181,22 @@ class LookupModule(LookupBase):
 
         return content, password, salt
 
-    def _write_password_file(self, path, content):
-        with open(path, 'w') as f:
-            os.chmod(path, 0o600)
+    def _write_password_file(self, b_path, content):
+        with open(b_path, 'w') as f:
+            os.chmod(b_path, 0o600)
             f.write(content + '\n')
 
-    def _create_password(self, path, params):
+    def _create_password(self, b_path, params):
         salt = None
 
-        if not os.path.exists(path):
-            self._create_password_file(path)
+        if not os.path.exists(b_path):
+            self._create_password_file(b_path)
 
             password = self.gen_password(length=params['length'],
                                          chars=params['chars'])
 
         else:
-            content, password, salt = self._read_password_file(path)
+            content, password, salt = self._read_password_file(b_path)
 
         if not salt:
             salt = self.random_salt()
@@ -178,7 +205,7 @@ class LookupModule(LookupBase):
                                       salt=salt,
                                       encrypt=params['encrypt'])
 
-        self._write_password_file(path, content)
+        self._write_password_file(b_path, content)
 
         if params['encrypt']:
             password = do_encrypt(password, params['encrypt'], salt=salt)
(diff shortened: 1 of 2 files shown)


alikins:
assigned to @abadger to verify the text encoding issues handling

abadger:
- I do like breaking the run() function up.  Nice job there.
- I'd probably move those all out to toplevel private functions.... None of them use the class except to call other methods which themselves should be at the toplevel.  (Too many methods vs functions is an artifact of how Ansible was originally coded.)
- The  `test_gen_password()` unittest shouldn't call `parse_parameters()` as that is already being tested on that data in `test_parse_parameters()`.  Instead directly use old_style_params_data[FOO]['filename'] and old_style_params_data[FOO]['params'].
  - Similarly, you probably want a unittest for `get_candidate_chars()`.  Once you have that, you don't have to call it in `test_gen_password()` anymore as you'll have the chars you expect as part of the test data.
- I don't think the text encoding issues are being handled right... I actually have some code started for this because mscherer was looking at porting this to python3 and I noticed that the way characters were being selected for the password would fail horribly if non-ascii was given: https://github.com/ansible/ansible/pull/17293/files#r76843964  (which might be the same problem as  #17461 ). Let me push my branch up and you can build off of that.

abadger:
Here's my changes: https://github.com/ansible/ansible/pull/17475  I haven't tested it yet so there may well be bugs in there but I'm way too tired to have another looks at it tonight... will definitely be willing to go over it in IRC/slack with you tomorrow, though.

alikins:
Took me longer to get reasonable test coverage here than expected, but should be close now.

lib/ansible/plugins/lookup/password.py
```diff
@@ -32,10 +32,14 @@
...
 from ansible.utils.path import makedirs_safe
 
+from ansible.module_utils._text import to_bytes, to_native, to_text
+
 DEFAULT_LENGTH = 20
 VALID_PARAMS = frozenset(('length', 'encrypt', 'chars'))
 
 
+# ALIKINS
+
```
abadger: Debugging stuff left in?

test/units/plugins/lookup/test_password.py
```diff
@@ -20,117 +20,349 @@
...
+from units.mock.loader import DictDataLoader
 from ansible.compat.tests import unittest
+from ansible.compat.tests.mock import patch
 
-from ansible.plugins.lookup.password import LookupModule, _parse_parameters, DEFAULT_LENGTH
+from ansible.compat.six import text_type
+
+from ansible.errors import AnsibleError
+from ansible.plugins import PluginLoader
+from ansible.plugins.lookup import password
```
abadger: Style-wise, I like to make the thing being tested the last import and visually separate from the other imports (a blank line between them and the import for the thing(s) being tested.

lib/ansible/plugins/lookup/password.py
```diff
@@ -73,91 +77,149 @@ def _parse_parameters(term):
...
+    # Asserts would be perfect for this if they were disabled by default.
+    # Alas, in python, they're active by default; only turned off if
+    # python is run with optimization.
+    # assert isinstance(chars, text_type)
+    random_generator = random.SystemRandom()
+
+    password = []
+    while len(password) < length:
+        new_char = random_generator.choice(chars)
+        if new_char in chars:
```
abadger: Ah, this conditional is no longer needed as the new code only chooses from the valid characters.

lib/ansible/plugins/lookup/password.py
```diff
@@ -73,91 +77,149 @@ def _parse_parameters(term):
...
+    '''
+
+    # hate how python implements asserts. Our text strategy says that we
+    # should never be passing byte strings around unless the function
+    # explicitly takes a byte string as its argument.  Therefore if we get
+    # a byte string here it means that the problem is in the calling code.
+    # Asserts would be perfect for this if they were disabled by default.
+    # Alas, in python, they're active by default; only turned off if
+    # python is run with optimization.
+    # assert isinstance(chars, text_type)
```
abadger: I've been thinking about this, I think we do need to either error or convert if the chars are not text type.  If bytes type creeps in by mistake, then random_generator.choice(chars) will pick mangled byte values.  So I suppose uncomment the assert and I can track down any bugs that it finds.

lib/ansible/plugins/lookup/password.py
```diff
@@ -73,91 +77,149 @@ def _parse_parameters(term):
...
+def _gen_candidate_chars(characters):
+    chars = []
+    for chars_spec in characters:
+        chars.append(to_text(getattr(string, to_native(chars_spec), chars_spec),
+                            errors='surrogate_or_strict'))
+    chars = u''.join(to_text(c) for c in chars).replace(u'"', u'').replace(u"'", u'')
+    return chars
+
+
+def _gen_password(length, chars):
```
abadger: We don't need both `_gen_password()` and `_random_password`.  Remove one of them.

lib/ansible/plugins/lookup/password.py
```diff
@@ -73,91 +77,149 @@ def _parse_parameters(term):
...
+    return u''.join(password)
+
+def _random_salt():
+    salt_chars = to_text(ascii_letters + digits + './', errors='surrogate_or_strict')
+    return _random_password(length=8, chars=salt_chars)
+
+def _gen_candidate_chars(characters):
+    chars = []
+    for chars_spec in characters:
+        chars.append(to_text(getattr(string, to_native(chars_spec), chars_spec),
```
abadger: Add a comment here that we're looking for things like "digits", "ascii_letters", etc.  Everytime I see this I spend a  little while figuring out what it does all over again.

lib/ansible/plugins/lookup/password.py
```diff
@@ -73,91 +77,149 @@ def _parse_parameters(term):
...
+                            errors='surrogate_or_strict'))
+    chars = u''.join(to_text(c) for c in chars).replace(u'"', u'').replace(u"'", u'')
```
abadger: The to_text() and generator expression aren't needed here as you ran to_text on everything in chars above.

lib/ansible/plugins/lookup/password.py
```diff
@@ -73,91 +77,149 @@ def _parse_parameters(term):
...
+    return chars
+
+
+def _gen_password(length, chars):
+    chars = _gen_candidate_chars(chars)
+    password = u''.join(random.choice(chars) for _ in range(length))
```
abadger: This uses the pseudo random number generator in the stdlib.  That's documented as being unsuited for cryptographic work.  If you keep _gen_password() instead of _random_password(), you'll need to use SystemRandom() as we do there.

lib/ansible/plugins/lookup/password.py
```diff
@@ -73,91 +77,149 @@ def _parse_parameters(term):
...
+    return password
+
+def _create_password_file(b_path):
+    b_pathdir = os.path.dirname(b_path)
+
+    try:
+        makedirs_safe(b_pathdir, mode=0o700)
+    except OSError as e:
+        raise AnsibleError("cannot create the path for the password lookup: %s (error was %s)" % (b_pathdir, str(e)))
```
abadger: Exceptions take native strings.  So you should do this like:

``` python
raise AnsibleError("cannot create the path for the password lookup: %s (error was %s)" % (to_native(b_pathdir), to_native(e)))
```

Byte strings could cause problems on python3 and str(e) could traceback.

lib/ansible/plugins/lookup/password.py
```diff
@@ -73,91 +77,149 @@ def _parse_parameters(term):
...
+    chars = u''.join(to_text(c) for c in chars).replace(u'"', u'').replace(u"'", u'')
+    return chars
+
+
+def _gen_password(length, chars):
+    chars = _gen_candidate_chars(chars)
+    password = u''.join(random.choice(chars) for _ in range(length))
+    return password
+
+def _create_password_file(b_path):
```
abadger: I notice this doesn't actually create the password file.  Perhaps "_create_password_file_dir"

Also, this is just a very light wrapper around makedirs_safe() and is called from a single location.... Perhaps it doesn't deserve its own function?

lib/ansible/plugins/lookup/password.py
```diff
@@ -73,91 +77,149 @@ def _parse_parameters(term):
...
+    b_pathdir = os.path.dirname(b_path)
+
+    try:
+        makedirs_safe(b_pathdir, mode=0o700)
+    except OSError as e:
+        raise AnsibleError("cannot create the path for the password lookup: %s (error was %s)" % (b_pathdir, str(e)))
+
+def _read_password_file(b_path):
```
abadger: Also a light wrapper around file.read()... Could remerge it with create_or_read_password_file.  Maybe this one has some justification because you can unittest with monkey patching this instead of mocking open if this is here... I'm not a fan of writing more complex code to make unittests easier but I can see where that might be desirable....

lib/ansible/plugins/lookup/password.py
```diff
@@ -73,91 +77,149 @@ def _parse_parameters(term):
...
+    content = open(b_path, 'rb').read().rstrip()
+    return content
```
abadger: In general, if we're going to treat something as text we want to transform it to text at the border. Since we never use password as a byte string, it's just when reading and writing the file that it is bytes, we should transform it here and return text:

``` python
    b_content = open(b_path, 'rb').read().rstrip()
    return to_text(b_content)
```

lib/ansible/plugins/lookup/password.py
```diff
@@ -73,91 +77,149 @@ def _parse_parameters(term):
...
+
+def _parse_password(b_content):
+    content = to_text(b_content, errors='surrogate_or_strict')
```
abadger: made a note earlier about transforming this to text when we read it from the file.  So we can pass in content as text and not do a transformation here.

lib/ansible/plugins/lookup/password.py
```diff
@@ -73,91 +77,149 @@ def _parse_parameters(term):
...
-    def random_password(self, length=DEFAULT_LENGTH, chars=C.DEFAULT_PASSWORD_CHARS):
-        '''
-        Return a random password string of length containing only chars.
-        NOTE: this was moved from the old ansible utils code, as nothing
-              else appeared to use it.
-        '''
+def _write_password_file(b_path, content):
+    with open(b_path, 'wb') as f:
+        os.chmod(b_path, 0o600)
+        f.write(content)
```
abadger: Just like transforming into text as soon as possible, we also want to transform into bytes as late as possible.  So we want to do the to_bytes() call here rather than before _write_password_file() is called:

``` python
with open(b_path, 'wb') as f:
    os.chmod(b_path, 0o600)
    f.write(to_bytes(content))
```

lib/ansible/plugins/lookup/password.py
```diff
@@ -73,91 +77,149 @@ def _parse_parameters(term):
...
-                        os.chmod(path, 0o600)
-                        f.write(content + '\n')
+            content, plaintext_password, salt = _read_or_create_password_file(b_path)
+            content, plaintext_password, salt = _update_content(plaintext_password, salt, params)
 
-            if params['encrypt']:
-                password = do_encrypt(password, params['encrypt'], salt=salt)
+            # save it
+            b_content = to_bytes(content, errors='surrogate_or_strict') + b'\n'
+            _write_password_file(b_path, b_content)
```
abadger: As noted earlier, pass the content as a text string and let _write_password_file transform it into bytes.  So:

``` python
content = content + u'\n'
_write_password_file(b_path, content)
```

lib/ansible/plugins/lookup/password.py
```diff
@@ -73,91 +77,149 @@ def _parse_parameters(term):
...
-        return self.random_password(length=8, chars=salt_chars)
+def _read_or_create_password_file(b_path):
+    salt = None
+    plaintext_password = None
+    content = None
 
-    def run(self, terms, variables, **kwargs):
+    if os.path.exists(b_path):
+        b_file_content = _read_password_file(b_path)
+        content, plaintext_password, salt = _parse_password(b_file_content)
```
abadger: As noted earlier, we'd be passing this around as file_content (text), rather than b_file_content (bytes).

lib/ansible/plugins/lookup/password.py
```diff
@@ -73,91 +77,149 @@ def _parse_parameters(term):
...
+    else:
+        _create_password_file(b_path)
+
+    return content, plaintext_password, salt
```
abadger: The split in responsibilities between this and _update_content feels wrong.  Here we're responsible for reading the information from the file and returning it along with the parsed form.  But if we don't have a file, we return data that flags an this step as incomplete.  In _update_content, we're responsible for producing a new combined content from the parsed form.  But if we get the flag, we have to create the exploded form of the data from scratch.

Instead we should make it this function's responsibility to return the parsed data.  It will either read the data in or create the data.  _update_content() can then go away.... We may want to call _format_content() in its place or (since we're returning the content here) we may want to call _format_content() from this function as well.

lib/ansible/plugins/lookup/password.py
```diff
@@ -73,91 +77,149 @@ def _parse_parameters(term):
...
+
+def _update_content(plaintext_password, salt, params):
```
abadger: As noted above, some or all of this should be merged with _read_or_create_password_file().

lib/ansible/plugins/lookup/password.py
```diff
@@ -73,91 +77,149 @@ def _parse_parameters(term):
...
-
-                if params['encrypt'] is not None and salt is None:
-                    # crypt requested, add salt if missing
-                    salt = self.random_salt()
-                    content = '%s salt=%s' % (password, salt)
-                    with open(path, 'w') as f:
-                        os.chmod(path, 0o600)
-                        f.write(content + '\n')
+            content, plaintext_password, salt = _read_or_create_password_file(b_path)
+            content, plaintext_password, salt = _update_content(plaintext_password, salt, params)
```
abadger: As noted earlier, _update_content should be merged with _read_or_create_password_file().  You may end up calling _format_content here in its place or that might be merged into _read_or_create_password_file().

test/units/plugins/lookup/test_password.py
```diff
@@ -20,117 +20,349 @@
...
+    @patch('ansible.plugins.lookup.password._create_password_file')
+    @patch('ansible.plugins.lookup.password._write_password_file')
+    def test_encrypt(self, mock_get_paths, mock_create_pf, mock_write_file):
+        mock_get_paths.return_value = ['/path/one', '/path/two', '/path/three']
+        fake_loader = DictDataLoader({'/path/to/somewhere':'sdfsdf'})
+        password_lookup = password.LookupModule(loader=fake_loader)
+        results = password_lookup.run([u'/path/to/somewhere encrypt=pbkdf2_sha256'],
+                                      None)
+
+        # FIXME: assert something useful
```
abadger: For encrypted you should be able to assert something about the format:

```
>>> crypt = getattr(passlib.hash, 'pbkdf2_sha256')
>>> crypt.encrypt(u'asda')
'$pbkdf2-sha256$20000$PUeoVar1/h8jxNjb./8fgw$UzKj5pfmKEDtguOh.XEvHCqQr2b4AHPgwL/M8BEzez8'
```

Initial string, number of "$", length of the salt and length of the hashed value.

test/units/plugins/lookup/test_password.py
```diff
@@ -20,117 +20,349 @@
...
+    @patch('ansible.plugins.lookup.password._write_password_file')
+    def test_only_a(self, mock_get_paths, mock_create_pf, mock_write_file):
+        mock_get_paths.return_value = ['/path/one', '/path/two', '/path/three']
+        fake_loader = DictDataLoader({'/path/to/somewhere':'sdfsdf'})
+        password_lookup = password.LookupModule(loader=fake_loader)
+        results = password_lookup.run([u'/path/to/somewhere chars=a'],
+                                  None)
+        for result in results:
+            for char in result:
+                self.assertEqual(char, u'a')
```
abadger: You should be able to know the length too (either the default or specify it) and then you don't have to loop through each character in result.

test/units/plugins/lookup/test_password.py
```diff
@@ -20,117 +20,349 @@
...
+        self.assertEqual(content, u'12345')
+
+    def test_no_password(self):
+        content, pw, salt = password._update_content(None, u'12345678', self.default_params)
+        self._assert_type(content, pw, salt)
+
+
+class TestParseParamaters(unittest.TestCase):
+    def test(self):
+        for testcase in old_style_params_data:
```
abadger, line 282: Could make this into a generated testcase if you were so inclined.

abadger (changes requested):

lib/ansible/plugins/lookup/password.py
alikins: Thought that seemed unneeded. Removing.

lib/ansible/plugins/lookup/password.py
alikins: Hmm, strings.digits (and ansible.constants.DEFAULT_PASSWORD_CHARS) are py2 strs on py2 and fail this assert. 

Moving the default for chars into the method and a None default arg

alikins: Though, I'm a little surprised a value from constants ended up as a regular string type...

lib/ansible/plugins/lookup/password.py
alikins: The name is bad.

It is split into it's own method so that it can be easily mocked for the unit tests.

alikins: I haven't figured out a good way to split _read_or_create_file better... The logic of the module is a little odd in that regards.

read_or_create_password_file() is sort of get_to_the_state_where_we_are_ready_to_write_the_password_file()

It's crammed into one method so that later code can avoid having to deal with the two potential scenarios. And, or, avoid  abstractions like

``` python

pw_store =  load_password_store()
pw_data = pw_store.read()
pw_data = update_pw_data(pw_data)
pw_store.update(pw_data)
pw_store.save()
```

lib/ansible/plugins/lookup/password.py
alikins: Yup, primarily for better testing.

And to keep os/io level stuff slightly encapsulated in lieu of a more complicated abstraction (pw_store, etc).

lib/ansible/plugins/lookup/password.py
alikins: I can add a wrapper function that will call both, but I'd prefer to keep the impls as separate methods.

alikins:
Think i have all the changed done.

Sans the flow around _read_or_create_password_file_dir().

lib/ansible/plugins/lookup/password.py
abadger: You can just monkeypatch makedirs_safe instead.

lib/ansible/plugins/lookup/password.py
```diff
@@ -73,91 +75,163 @@ def _parse_parameters(term):
...
+    '''
+
+    # hate how python implements asserts. Our text strategy says that we
+    # should never be passing byte strings around unless the function
+    # explicitly takes a byte string as its argument.  Therefore if we get
+    # a byte string here it means that the problem is in the calling code.
+    # Asserts would be perfect for this if they were disabled by default.
+    # Alas, in python, they're active by default; only turned off if
+    # python is run with optimization.
+    chars = chars or to_text(C.DEFAULT_PASSWORD_CHARS)
```
abadger, line 102: Convert C.DEFAULT_PASSWORD_CHARS into text in lib/ansible/constants,py instead.

abadger:
These commits and additional changes merged via: https://github.com/ansible/ansible/pull/17626

PR #17475, opened 2016-09-09
Fix the text model of the password lookup.

### LINKED ISSUE #17461: password lookup doesn't always respect chars setting

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

lookup/password.py
##### ANSIBLE VERSION

```
2.2
```
##### SUMMARY

The password lookup was using bytes when it **needed** to use text.  Using bytes where it did would lead to mangled passwords when working with non-ascii characters.
- This didn't work on python3.
- It didn't work with non-ascii characters.

This might fix #17461 although I didn't see that issue until after
I started working on this.

### DIFF
--- a/lib/ansible/constants.py
+++ b/lib/ansible/constants.py
@@ -25,9 +25,9 @@ from string import ascii_letters, digits
 
 from ansible.compat.six import string_types
 from ansible.compat.six.moves import configparser
-
-from ansible.parsing.quoting import unquote
 from ansible.errors import AnsibleOptionsError
+from ansible.module_utils._text import to_text
+from ansible.parsing.quoting import unquote
 from ansible.utils.path import makedirs_safe
 
 BOOL_TRUE = frozenset([ "true", "t", "y", "1", "yes", "on" ])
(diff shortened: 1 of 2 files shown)


abadger:
If/when this is added we can close this: https://github.com/ansible/ansible/pull/17293

alikins:
I may see if I can merge this into https://github.com/ansible/ansible/pull/17465  (and it's proposed changes). That would get some test coverage in here and better handle the text encoding oddities I saw.

alikins:
https://github.com/ansible/ansible/pull/17465 should have these changes, my changes, and more test coverage included now.

Issue #17269, opened 2016-08-27
with_fileglob not matching *just* directories


charlienewey:
##### ISSUE TYPE

 Bug Report
##### COMPONENT NAME

with_fileglob
##### ANSIBLE VERSION

```
ansible 2.0.0.2
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides
```
##### CONFIGURATION

nil/defaults
##### OS / ENVIRONMENT

Ubuntu 16.04
##### SUMMARY

Directories not being matched under file glob. This should be possible; some testing in various UNIX shells (bash, sh, zsh) and in the Python `glob` module itself seems to show that the following glob pattern: 

```
some/path/*/
```

will _just_ match directories under `{{variable_holding_some_path}}/`.
##### STEPS TO REPRODUCE

This task:

``` yml
- debug: "{{item}}"
  with_fileglob:
    - "{{variable_holding_some_path}}/*/"
```
##### EXPECTED RESULTS

A list of all directories under `some/path`.
##### ACTUAL RESULTS

No output, but task is supposedly run.

charlienewey:
Update: I can get this to work as expected by using the following:

``` yml
- debug: "{{item}}"
  vars:
    glob_: "{{variable_holding_some_path}}/*/"
  with_items:
    - "{{glob_ | fileglob}}"
```

dagwieers:
@charlienewey The implementation of the **fileglob** lookup plugin _only_ returns files, not directories. That's how it was implemented as the aim is to iterate over files. If the lookup plugin would return a mix of files and directories, it would not be obvious what would be a file or a directory. So this behaviour is on purpose. And that is why it is called **fileglob**, and not just **glob**.

If you are interested in listing directories, you can make use of the **filetree** lookup plugin, it does return a list of files/directories/symlinks using a recursive scan, and it does provide you with additional information (like whether a _file_ is a directory, symlink or regular file).

That said, there is an obvious mismatch between the **fileglob** lookup plugin and the **fileglob** filter plugin (jinja2). A fix is in the making :-)

bcoca:
closing as dupe of #17136

Issue #17136, opened 2016-08-18
with_fileglob does not handle a `*` to match a directory


rbauduin:
##### ISSUE TYPE

- Bug Report
##### COMPONENT NAME

with_fileglob
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel bd31cc096a) last updated 2016/08/18 15:46:34 (GMT +200)
  lib/ansible/modules/core: (detached HEAD ff2eae0729) last updated 2016/08/18 15:46:38 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD 7ba5f340c0) last updated 2016/08/18 15:46:38 (GMT +200)
  config file = /home/rbauduin/gits/ansible-playbooks/ansible.cfg
  configured module search path = Default w/o overrides
```
##### CONFIGURATION

ansible.cfg:

```ini
[default]
forks=20

[ssh_connection]
ssh_args = -F ssh-config 
scp_if_ssh = True
pipelining = False

[privilege_escalation]
become=True

```
##### OS / ENVIRONMENT

Ubuntu 14.04 to Centos 7.2
##### SUMMARY

with_fileglob does not seem to handle a `*` to match a directory in the glob.
I think this worked with a 1.X version of ansible.
##### STEPS TO REPRODUCE

This debug task finds the file located at ./downloaded_files/client/root/.ssh/id_rsa.pub

```yaml
  tasks:
  - name: "locate base dir"
    debug: msg="file is {{item}}"
    with_fileglob:
            - ../downloaded_files/client/root/.ssh/id_rsa.pub
```

This one doesn't:

```yaml
  tasks:
  - name: "locate base dir"
    debug: msg="file is {{item}}"
    with_fileglob:
            - ../downloaded_files/*/root/.ssh/id_rsa.pub
```

But these do:

```yaml
  tasks:
  - name: "locate base dir"
    debug: msg="file is {{item}}"
    with_fileglob:
            - ../downloaded_files/client/root/.ssh/*
```

```yaml
  tasks:
  - name: "locate base dir"
    debug: msg="file is {{item}}"
    with_fileglob:
            - ../downloaded_files/client/root/.ssh/*.pub*
```
##### EXPECTED RESULTS

Debug output with the file path.
##### ACTUAL RESULTS

No file is found by the fileglob with a `*` matching directories in the path glob.

```
TASK [locate base dir] *********************************************************

PLAY RECAP *********************************************************************
```

nitzmahone:
[...]

PR #17480, opened 2016-09-09
Fix fileglob filter to work just like fileglob lookup plugin

dagwieers:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

fileglob filter plugin
##### ANSIBLE VERSION

v2.2 and older
##### SUMMARY

The fileglob lookup plugin only returns files, not directories.
This is to be expected, as a mixed list would not be very useful in with_fileglob.
However the fileglob filter does return anything glob.glob() returns.

This change fixes this, so that fileglob returns files (as the name indicates).

PS We could also offer a glob filter for thos that would need it ?

This relates to comments in issue #17136 and fixes confusion in #17269.

### DIFF
--- a/lib/ansible/plugins/filter/core.py
+++ b/lib/ansible/plugins/filter/core.py
@@ -126,8 +126,8 @@ def quote(a):
     return pipes.quote(a)
 
 def fileglob(pathname):
-    ''' return list of matched files for glob '''
-    return glob.glob(pathname)
+    ''' return list of matched regular files for glob '''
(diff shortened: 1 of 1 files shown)


bcoca:
this is a bad filter ... as it looks at the 'controller' filesystem ... this should have stayed as a lookup ... but your change is congruent.

dagwieers:
It would not be a bad idea to also improve the documentation and refer to the **find** module for remote listings where **fileglob** is mentioned.

PR #9047, opened 2014-09-17
Introduce 'play_hosts_all' next to 'play_hosts'

srgvg:
`play_hosts` are all (active) hosts in the current serialized play.
`play_hosts_all` (_now_) are all (active) hosts in the full play = all serial batches.

use case: in context of deploying clusters of 2 nodes with `serial: 1`, performing a pre-deploy healtcheck, only if the peer node is part of the (full) play: `when: peer_hostname in play_hosts_all`

`play_hosts` didn't suit, as it only contains the serialized batch in a play.

[Review on Reviewable](https://reviewable.io/reviews/ansible/ansible/9047)

PR #17498, opened 2016-09-10
Feature/add ansible play hosts all

j0hnsmith:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

Varible `ansible_play_hosts`
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file = 
```
##### SUMMARY

When using serial, `ansible_play_hosts` includes only the hosts in the current batch, currently there's no way to get 'all' the hosts regardless of serial.

This PR adds `ansible_play_hosts_all` which does just that. There an [old PR](https://github.com/ansible/ansible/pull/9047) against ansible v1.x that also has some relevant discussion.

This is my first foray into the ansible source, I hope the approach I've taken is reasonable, if it's not please suggest an alternative.

### DIFF
--- a/lib/ansible/vars/__init__.py
+++ b/lib/ansible/vars/__init__.py
@@ -413,8 +413,7 @@ class VariableManager:
                 # DEPRECATED: play_hosts should be deprecated in favor of ansible_play_hosts,
                 #             however this would take work in the templating engine, so for now
                 #             we'll add both so we can give users something transitional to use
-                host_list = [x.name for x in self._inventory.get_hosts()]
-                variables['play_hosts'] = host_list
+                variables['play_hosts'] = [x.name for x in self._inventory.get_hosts()]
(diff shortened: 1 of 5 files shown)


bcoca:
actually, this is what ansible_play_hosts is supposed to be, it was made planning to deprecate play_hosts in favor of ansible_batch_hosts or similar.

lib/ansible/vars/__init__.py
```diff
@@ -416,6 +416,7 @@ def _get_magic_variables(self, loader, play, host, task, include_hostvars, inclu
                 host_list = [x.name for x in self._inventory.get_hosts()]
                 variables['play_hosts'] = host_list
                 variables['ansible_play_hosts'] = host_list
+                variables['ansible_play_hosts_all'] = [x.name for x in self._inventory.get_hosts(pattern=play.hosts or 'all', ignore_restrictions=True)]
```
bcoca: just remove the previous line and drop '_all' from the variable

j0hnsmith: done

Issue #15787, opened 2016-05-09
Command line parsing of key=value arguments causes invalid Windows paths to be generated


jhawkesworth:
##### ISSUE TYPE
- Bug Report
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel 848b015ce6) last updated 2016/05/09 08:31:01 (GMT +100)
  lib/ansible/modules/core: (detached HEAD 1f5cf669dd) last updated 2016/05/09 08:31:10 (GMT +100)
  lib/ansible/modules/extras: (detached HEAD 7b7af3bcf9) last updated 2016/05/09 08:31:06 (GMT +100)
  config file = 
  configured module search path = Default w/o overrides

```
##### CONFIGURATION

Defaut configuration
##### OS / ENVIRONMENT

Ansible on: Lubuntu 14.04 
3.16.0-70-generic #90~14.04.1-Ubuntu SMP Wed Apr 6 22:56:34 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
LANG=en_GB.UTF-8
GDM_LANG=en_GB

Targeting: Windows Server 2012 R2
##### SUMMARY

When valid windows paths are supplied as command line arguments to either ansible or ansible-playbook backslash followed by certain characters is turned into a unicode escape sequence, which has unintended side effects as the windows paths are then either incorrect, or invalid.

Attempting to modify the windows path in this way can create an invalid path, causing windows to loose powershell from the path, thereby stopping ansible from running any remote commands (by only following a reboot of the windows host as system environment variables (typically) are considered immutable following process creation.
##### STEPS TO REPRODUCE

for ad-hoc ansible commands, the following will reproduce the problem:

```
ansible WIN-TEST -m win_environment -a 'level=machine name=DONTDOTHIS value=C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\WindowsPowerShell\v1.0\' 
```

As similar issue arises when passing -e extra vars to ansible-playbook.

```
cat promote-component.yml 
- hosts: localhost
  tasks:
    - name: show the parameters supplied
      debug:
        var: installer

ansible-playbook promote-component.yml -e 'installer=B:\artifacts\components\testComponent\1.2.3.4\testComponent.1.2.3.4.zip' -c local
```
##### EXPECTED RESULTS

ansible example - expected

```
WIN-TEST | SUCCESS => {
    "before_value": null,
    "changed": true,
    "level": "machine",
[...]

PR #17527, opened 2016-09-12
Fixes to the controller text model

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

Throughout ansible cli.
##### ANSIBLE VERSION

```
devel
```
##### SUMMARY
- Change command line args to text type
- Make display replace undecodable bytes with replacement chars.  This
  is only a problem on pyhton3 where surrogates can enter into the msg
  but sys.stdout doesn't know how to handle them.
- Remove a deprecated playbook syntax in unicode.yml

### DIFF
--- a/bin/ansible
+++ b/bin/ansible
@@ -91,9 +91,16 @@ if __name__ == '__main__':
             else:
                 raise
 
-        cli = mycli(sys.argv)
-        cli.parse()
-        exit_code = cli.run()
+        try:
+            args = [to_text(a, errors='surrogate_or_strict') for a in sys.argv]
+        except UnicodeError:
+            display.error('Command line args are not in utf-8, unable to continue.  Ansible currently only understands utf-8')
+            display.display(u"The full traceback was:\n\n%s" % to_text(traceback.format_exc()))
+            exit_code = 6
+        else:
+            cli = mycli(args)
+            cli.parse()
+            exit_code = cli.run()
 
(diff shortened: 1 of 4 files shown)


jhawkesworth:
Not sure yet if this is relevant but I'm interested to see if this fixes the issues described in https://github.com/ansible/ansible/issues/15787 - I will do some testing now.

abadger:
It will be great if it fixes some things... I'm anticipating that this may break some other things.  It follows the unicode sandwich strategy that we've been working on since 2.0 (and to some extent in 1.x) in core so we're going to try to put this in and see if we can fix any problems reported before 2.2.

jhawkesworth:
Unfortunately it doesn't seem to fix the thing I'm particular interested in (passing windows paths on the command line).  From what I've just read its a feature of python though, so not sure how I could get round that.

jhawkesworth:
I just read about 'raw strings' which might help a bit if there was some way to pass that a -e parameter was to be treated as a raw string, but even then that would only help if passing in a full path to a file, rather than any windows path - windows directories are often written with a trailing \  which would not be helped by raw strings anyway.  Oh well, I'll think about it some more.

abadger:
raw strings are only for string literals.  In the case of things passed in via the command line, it's already a variable that's been interpreted by bash and I believe the parse_kv() function inside of ansible.

PR #17533, opened 2016-09-13
Added ability to add multiple hosts from custom ActionModule by returning 'add_hosts' in result

nickmy9729:
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

ansible.plugins.stategy.StrategyBase class
##### ANSIBLE VERSION

```
ansible 2.2.0
  configured module search path = ['/Users/nmyers002/custom_modules']
```
##### SUMMARY

Added ability for ActionModules to return 'add_hosts', with a list of dictionaries of the same data format as 'add_host'.  By doing this I created a loop to call the _add_host method to allow for adding multiple files in one task call instead of creating a task loop.  

The rational behind this choice was the need for a custom ActionModule that would receive a list of params, query a CMDB for the host data, and then add all hosts in a single call.  This simplified the add_host call greatly, and reduced the yaml needed to to accomplish the task with the default add_host call.

```

```


nickmy9729:
Added ability to add multiple hosts in a custom ActionModule by returning 'add_hosts' in result key with a list of dict's.

calfonso:
Hi!

We don't want to add the ability to modules to add arbitrary hosts. This should be done using add_host module in a loop.

Issue #17350, opened 2016-09-01
Revisiting realtime stdout/stderr


PeterMartini:
##### ISSUE TYPE

- Feature Idea
##### COMPONENT NAME

Realtime stdout/stderr capture
##### ANSIBLE VERSION

devel
##### OS / ENVIRONMENT

Various *nix
##### SUMMARY

I'd like to revisit #3887, collecting realtime stdout/stderr.  The last comment there pointed out the difficulties of feeding the output back to the ansible command layer, which I fully understand, so I have a counterproposal (and a prototype at https://github.com/PeterMartini/ansible/tree/ansible-tmp) - if stdout/stderr are buffered to disk alongside a copy of the python wrapper script ansible uses to actually execute its tasks (on *nix), then that data becomes available for an out-of-band process to pull back and examine the data.

With the branch referenced, a trivial callback plugin can be created to capture the task UUID for each host at the time the task is started, and use that UUID to derive the path where stdout/stderr will be buffered on the remote system.  With that information, stdout and stderr can be checked in realtime either manually or via a separate set of scripts.

If this makes sense, I'd like to work with whoever's interested to get that branch sufficiently polished to create a PR, review, and merge.

ChristianTacke:
2.1 added this: "Added facility for modules to send back 'diff' for display when ansible is called with --diff, updated several modules to return this info"
So returning some data seems to be possible.

PeterMartini:
If I run ansible with -vvv (offhand, I don't know which verbosity level is needed, I just like that one) its clear stdout/stderr are returned, they're just returned too late (meaning only after the process has finished).  The diff flag similarly only returns when the process is finished, which doesn't help where I need it (checking whether a long running process is stuck).

When I get back to a computer, I'll post more details of where/how this is useful (including a sample callback)

PeterMartini:
Bah, I did some more testing, and discovered there are at least two issues:
1. The callback is only called once at the start of each task, not once per (task, host)
2. The uuid similarly is for the task, so any kind of `local_action`would end up sharing the same directory

bcoca:
The issue is much deeper than that, the task ONLY returns info to the Ansible controller AFTER it has ended. 

I have a `update_json` proposal that would allow for tasks to return data DURING execution but it has many hurdles to get through before being viable.  

https://github.com/bcoca/ansible/tree/update_json

Once this facility exists, then we can have each module implement it.

PeterMartini:
[...]

PR #13620, opened 2015-12-21
update_json for module intermediate comms

bcoca:
- new 'live' keyword to indicate you want intermediate (live) updates from module execution
- added 'live' intelligence to 'run_command' for those modules executing commands using it

TODO:
- displays when -vvvvv ... really should be using callback now that sivel added queuing

implements https://github.com/ansible/proposals/issues/92

PR #17536, opened 2016-09-13
Add config option to tee stdout and stderr to temp files

PeterMartini:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

executor / actions plugin
##### ANSIBLE VERSION

based off of devel at 036650cb097352a9810c6acfe2bd3354859a5149
##### SUMMARY

This is a WIP PR to address #17350.

The basic idea:

With the 3 commits here, and a plugin installed that looks something like:

```
from ansible.plugins.callback import CallbackBase

class CallbackModule(CallbackBase):
    CALLBACK_VERSION = 2.0
    CALLBACK_TYPE = 'logging'
    CALLBACK_NAME = 'logger'

    def __init__(self, display=None):
        super(CallbackModule, self).__init__(display)

    def v2_runner_on_start(self, host, task):
        print "You can tail the logs at ~/.ansible/tmp/%s-%s/std{out,err}" % (host, task._uuid)
```

The logs are now available on disk on the remote system and available for inspection through whatever means are handy.

The 3 commits are to:
1. Add a new callback for when the runner starts, since both host and task information is necessary
2. Add a config variable to control the `tee` behavior, buffering to `stdout` and `stderr`
3. Change the temporary directory formula, which currently includes a random number, to use host and the task uuid instead to make the full path predictable.  Passing the tmp dir in or out would seem like a better solution, but it seems like that would be more intrusive.

*\* Opening as a WIP for discussion **

### DIFF
--- a/lib/ansible/constants.py
+++ b/lib/ansible/constants.py
@@ -177,6 +177,7 @@ DEFAULT_SFTP_BATCH_MODE   = get_config(p, 'ssh_connection', 'sftp_batch_mode', '
 DEFAULT_MANAGED_STR       = get_config(p, DEFAULTS, 'ansible_managed',  None,           'Ansible managed: {file} modified on %Y-%m-%d %H:%M:%S by {uid} on {host}')
 DEFAULT_SYSLOG_FACILITY   = get_config(p, DEFAULTS, 'syslog_facility',  'ANSIBLE_SYSLOG_FACILITY', 'LOG_USER')
 DEFAULT_KEEP_REMOTE_FILES = get_config(p, DEFAULTS, 'keep_remote_files', 'ANSIBLE_KEEP_REMOTE_FILES', False, boolean=True)
+DEFAULT_TEE_REMOTE_OUTPUT = get_config(p, DEFAULTS, 'tee_remote_output', 'ANSIBLE_TEE_REMOTE_OUTPUT', False, boolean=True)
 DEFAULT_HASH_BEHAVIOUR    = get_config(p, DEFAULTS, 'hash_behaviour', 'ANSIBLE_HASH_BEHAVIOUR', 'replace')
 DEFAULT_PRIVATE_ROLE_VARS = get_config(p, DEFAULTS, 'private_role_vars', 'ANSIBLE_PRIVATE_ROLE_VARS', False, boolean=True)
 DEFAULT_JINJA2_EXTENSIONS = get_config(p, DEFAULTS, 'jinja2_extensions', 'ANSIBLE_JINJA2_EXTENSIONS', None)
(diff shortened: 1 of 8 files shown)


bcoca:
Why do you want to copy the same data you already get back at the controller?

Also, consider that this won't work or will break pipelining (which avoids writing to disk on remote hosts).

PeterMartini:
I don't get the data back at the controller until the command finishes.  If the whatever I'm running has printed "Waiting for <foo>" to continue, or has just printed "97% complete" to stdout, there's no way I'll know that until I kill off that particular task.  Being able to tee the files to disk automatically means that if I suspect it's either waiting for some signal or streams a progress indicator to stdout, I can see that by just knowing where to look.

Point taken about pipelining though.  I'll look into what it would take to add that, under the assumption that using the two options in conjunction would imply the extra writing to disk is fine, it's the performance boost of not needing separate copies that motivates the grouping.

bcoca:
this won't give  you the data any sooner, the module only outputs when it is finished.

bcoca:
I was creating a facility for modules to be able to send updates before finishing, https://github.com/bcoca/ansible/tree/update_json is my branch (outdated). 

This mostly implements the facility on the controller, modules would still be required to send the data.

PeterMartini:
I'll be following the update_json branch with interest, but I do want to make a bit more explicit why I took this approach:

I don't expect to check the status from the controller.  Having the callback means I can log enough details to construct where the output files live, and then I expect either another tool - like Tower - to pull the data back out of band, or to open a session to the problematic machine directly and check the logs.

bcoca:
I don't think i made myself clear, those files would not populated until the task ends.  The modules only output when they finish, even locally. 

The only case in which those files would get you the output and not at the controller is if the network connection fails after the task ends and all the data isn't transmitted back. They would not give you intermediate updates on the task.

PeterMartini:
My apologies, I believe I understand what you mean now, but I'm not sure if it applies to the code as written here.  Here's my test setup so you can see what I see (and hopefully advise how I can build this into a proper test).

My playbook, `test.yml`:

```
---
- hosts: local
  gather_facts: false
  tasks:
    - name: test 1
      local_action: shell i=0; while [ ! -f /tmp/test1 ]; do echo $i; i=$((i+1)); sleep 1; done

    - name: test 2
      shell: i=0; while [ ! -f /tmp/test2 ]; do echo $i; i=$((i+1)); sleep 1; done
```

`inventory`:

```
[local]
localhost
```

`ansible.cfg`:

```
[defaults]
tee_remote_output = True
callback_plugins  = /home/pmartini/plugins/callback
```

Using this setup, and the plugin at the top of this PR, when I run `ansible-playbook -i inventory test.yml` I can tail the filename it prints out while the command is running and observe it moving along.

The key piece of the PR is everytime select returns that a filehandle is ready, `stdout` the buffer gets extended, and `stdoutfile` the filehandle gets written to, so it does stream updates in real time.  And the `0` in the open call ensures there's no output buffering.

bcoca:
closing in favor of   #13620

PR #17551, opened 2016-09-13
Fix slaac filter

jre:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

slaac filter module
##### ANSIBLE VERSION

```
ansible 2.1.1.0

```
##### SUMMARY

The code which attempts to test an IP address version attribute doesn't work, the address is a string. I've followed the style of the existing code and used the ipaddr() function to do so.

Using the template data `{{ 'fe80::' | slaac('00:11:22:33:44:55') }}` the slaac filter output changes from False to a SLAAC address:

```
TASK [template] ****************************************************************
--- before: /tmp/testfile
+++ after: dynamically generated
@@ -1 +1 @@
-False
+fe80::211:22ff:fe33:4455
```

### DIFF
--- a/lib/ansible/plugins/filter/ipaddr.py
+++ b/lib/ansible/plugins/filter/ipaddr.py
@@ -608,7 +608,7 @@ def slaac(value, query = ''):
         elif vtype == 'network':
             v = ipaddr(value, 'subnet')
 
-        if v.version != 6:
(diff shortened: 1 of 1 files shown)


mscherer:
We would need to add that to #17628

Issue #2750, opened 2013-04-23
ec2.py module shabang


devdave:
https://github.com/ansible/ansible/commit/9dda148679c32299bc7cbed2c19874a3e3e1db72#commitcomment-3064604

ec2.py requires boto to be installed to the system's python library and not to a virtualenv which is why the last iteration had #!/usr/bin/env python versus current explicit one.  I am using ec2.py via a local_action and it fails without editing the shabang.

mpdehaan:
So modules aren't meant to run within a virutalenv.  Never were.

If you want to change the interpreter line, you set ansible_python_interpreter via a variable to what you need it to be.

The previous usage of /usr/bin/env breaks the ansible_python_interpreter sytem of replacing the shebang on a per host basis.

If you have further questions please stop by the list.

PR #17571, opened 2016-09-14
lxd_container module fails if certificate already in trust store #2750

june07:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

module_utils/lxd.py
affected lxd_container module
##### ANSIBLE VERSION

```
ansible 2.2.0

```
##### SUMMARY

When the client certificate is already stored, lxd returns a JSON error with message "Certificate already in trust store".  This "error" will occur on every task run after the initial run.  The cert should be in the trust store after the first run and this error message should really only be viewed as informational as it does not indicate a real problem.  I certainly don't think an exception should be being thrown?!

Fixes:
https://github.com/ansible/ansible-modules-extras/issues/2750

```

```


lib/ansible/module_utils/lxd.py
```diff
@@ -121,6 +121,8 @@ def _send_request(self, method, url, body_json=None, ok_error_codes=None, timeou
             if resp_type == 'error':
                 if ok_error_codes is not None and resp_json['error_code'] in ok_error_codes:
                     return resp_json
+                if resp_json['error'] == "Certificate already in trust store":
```
hnakamur, line 124: @june07 Thanks for your pull request.

At first, I thought I do not like depending on the particular error message and it would be better to call the `GET /1.0/certificates` or `GET /1.0/certificates/<fingerprint>` beforehand to check the certificate is in trust store.

However, since the LXD REST API does not have concept of transaction, after checking with one of the two APIs and before calling `POST /1.0/certificates`, there may be other requests to register the same certificate.
So I think checking the error of `POST /1.0/certificates` is the way to go.

I looked at the LXD code and confirmed the error message is the only way to check the error type.
https://github.com/lxc/lxd/blob/26858d333f13ac5e97bbc40a542374c7ec9156c3/lxd/certificates.go#L147

So after all, I agree to your pull request.

hnakamur (approved):

lib/ansible/module_utils/lxd.py
june07: Thanks.

alikins:
Not sure what happened to the branch, but it needs some attention to the git branching.

june07:
Closing.  See https://github.com/ansible/ansible/pull/18115 instead.
The commit history got messed up so I had to resubmit the PR as above.

Issue #4840, opened 2013-11-07
Allow the user to set the disk size of the root partition partition on ec2


evanccnyc:
In boto, you can set the size of the root, before you launch the instance.  This is really helpful if you need more disk size then whatever the ami has.

The example boto code is here:

 https://gist.github.com/anonymous/df0ddf374b3dd982c670

lwade:
Would be good to support block device mappings.  Keep in mind that this example is for EBS-backed instances only, not Instance-store backed.

willthames:
This would be a minor tweak to https://github.com/ansible/ansible/pull/4534
to allow the special case of /dev/sda1 with a custom size.

mpdehaan:
I'm closing this ticket as we have a pull request open now that tracks this issue, just to keep github easier to organize.   This ticket can still be found from the other.

PR #17572, opened 2016-09-14
nxos_config, nxos_template fixes per issue #4840

dgjustice:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

nxos_config
nxos_template
##### ANSIBLE VERSION

```
ansible 2.2.0
  config file =
```
##### SUMMARY

nxos_config and nxos_template modules run fine in standalone playbooks when using the "src" option.  When imported from a role, they both fail with "_task has no _block attribute."  I added checks to see if the attribute exists.  This caused an error to be thrown depending on the transport type (nxapi or cli).  I also added checks in nxos.py in the LoadConfig class to call self.execute with the correct parameters depending on the transport class it is being called from.

```

```


dgjustice:
I removed the old lookup plugin from my repo and pushed again.  Didn't catch that earlier.  It was old work, and should have been left out.  ./lib/ansible/plugins/lookup/link_from_graph.py

privateip:
Please split the lookup plugin into a separate PR

gundalow:
Just spoken to dgjustice, a fresh PR will follow.

PR #17577, opened 2016-09-14
Update to Get-AnsibleParams to allow for multiple names

adbertram:
##### ISSUE TYPE

- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.1.0
```
##### SUMMARY

```

```

As-is, Get-AnsibleParams only supports a single $name parameter. This means in order to check for multiple names in a module, a user would have to do this:

```
$requiredParams = 'param1','param2'
foreach ($n in $requiredParams)
{
    Get-AnsibleParams -obj $params -name $n
}
```

Now the user just needs to do this: `Get-Ansible-Params -obj $params -name $requiredParams`


trondhindenes:
@adbertram we'd still need to call it multiple times to separate required from non-required params, right? Or am I understanding it wrongly?

trondhindenes:
Also, we need a way to control the variable where the result of "Get-AnsibleParam" is stored. How does your mod deal with that?

adbertram:
Good point. I didn't think of that use case. I'm being pulled in another direction for awhile so won't be able to do much Ansible work for a few weeks. I'll come back around to this.

PR #17580, opened 2016-09-14
Add state machine capabilities for use by modules

relaxdiego:
##### ISSUE TYPE

Feature Pull Request
##### COMPONENT NAME

N/A
##### ANSIBLE VERSION

```
ansible 2.2.0 (detached HEAD 6787fc70a6) last updated 2016/09/14 14:52:45 (GMT -700)
  lib/ansible/modules/core: (detached HEAD db5cb54b23) last updated 2016/09/08 14:25:22 (GMT -700)
  lib/ansible/modules/extras: (load-balancer-rewrite 0f3df9a68f) last updated 2016/09/13 15:52:35 (GMT -700)
  config file = 
```
##### SUMMARY

This change adds a new class called AnsibleModuleFSM which is built on top of AnsibleModule via composition (not inheritance). The main advantage of using this class is that a module developer can now write modules as state machines which can be easier to manage especially for those that have a complex decision tree.
##### RATIONALE

Non-interactive code with decision trees that range from mid to high complexity becomes hard to
manage very quickly if implemented using procedural methods (if conditions). Take for example this part of the `ec2_customer_gateway` module:

``` python
if module.params['state'] == 'present':
    if existing['CustomerGateways']:
        results['gateway']=existing
        if existing['CustomerGateways'][0]['Tags']:
            tag_array = existing['CustomerGateways'][0]['Tags']
            for key, value in enumerate(tag_array):
                if value['Key'] == 'Name':
                    current_name = value['Value']
                    if current_name != name:
                        results['name'] = gw_mgr.tag_cgw_name(
                            results['gateway']['CustomerGateways'][0]['CustomerGatewayId'],
                            module.params['name'],
                        )
                        results['changed'] = True
    else:
        if not module.check_mode:
            results['gateway'] = gw_mgr.ensure_cgw_present(
                module.params['bgp_asn'],
                module.params['ip_address'],
            )
            results['name'] = gw_mgr.tag_cgw_name(
                results['gateway']['CustomerGateway']['CustomerGatewayId'],
                module.params['name'],
            )
        results['changed'] = True

elif module.params['state'] == 'absent':
    if existing['CustomerGateways']:
        results['gateway']=existing
        if not module.check_mode:
            results['gateway'] = gw_mgr.ensure_cgw_absent(
                existing['CustomerGateways'][0]['CustomerGatewayId']
            )
        results['changed'] = True
```

Note the 5 layers of if conditions with two of them belonging inside a for loop. If one or more parts of this behavior were to change, the developer would need to be very careful to ensure that a change in one or more conditions/variables doesn't affect any other part unintentionally since there's no way to scope variables within a single if condition.

Other modules get around this problem by moving some parts of the code to their own function. While that solves the deeply-nested-if problem, it introduces another complexity in that a developer will have to scroll up and down the module's code to understand its workflow.

This change addresses both problems by providing a means for the module developer to implement a module using state machines. Using state machines allows the developer to avoid deeply-nested if conditions while still keeping the workflow information in a single place.
##### SAMPLE USAGE

Using this feature is a matter of using the AnsibleModuleFSM class instead of AnsibleModule. The former accepts the same arguments as the latter plus two additional keyword args, `state_machine` and `render_path`.

``` python
#!/usr/bin/env python
# Note the additional imports, create_events and create_states.
# These are optional helper functions that allow you to define
# more descriptive events and states without too much boilerplate.
from ansible.module_utils.basic import AnsibleModuleFSM, create_events, create_states

# Use create_states to define descriptive state names/values.
# The following allows us to, for example, use S.initializing
# in our code below.
S = create_states('choosing_action',
                  'creating',
                  'deleting',
                  'exiting',
                  'initializing',
                  'getting',
                  'waiting')

# This allows us to define descriptive event names/values
E = create_events('create_started',
                  'delete_started',
                  'initialized',
                  'exited',
                  'found',
                  'not_found',
                  'must_wait',
                  'must_create',
                  'must_delete',
                  'must_exit')
def main():
    # Define the module's transitions. Note how we are using the
    # states and events we defined above.
    #
    # Note how we pass in functions such as get_obj, choose_action,
    # create_obj, etc. You will find the definition of these
    # functions further down this example.
    #
    # The dict's structure is:
    #    (CURRENT_STATE, EVENT): (NEW_STATE, ENTRY_ACTION)
    #
    # The state machine purposely doesn't support exit actions to
    # keep its implementation simple.
    transitions = {
        (S.initializing, E.initialized):    (S.getting, get_obj),
        (S.getting, E.found):               (S.choosing_action, choose_action),
        (S.getting, E.not_found):           (S.choosing_action, choose_action),
        (S.choosing_action, E.must_create): (S.creating, create_obj),
        (S.choosing_action, E.must_delete): (S.deleting, delete_obj),
        (S.choosing_action, E.must_exit):   (S.exiting, exit_now),
        (S.creating, E.create_started):     (S.waiting, wait),
        (S.deleting, E.delete_started):     (S.waiting, wait),
        (S.waiting, E.must_wait):           (S.waiting, wait),
        (S.waiting, E.must_exit):           (S.exiting, exit_now)
    }

    # Instantiate the object.
    AnsibleModuleFSM(
        # Uncomment this argument to make the module render
        # the state diagram and exit immediately. The file
        # extension will be appended automatically.
        # render_path="/tmp/mymodule",

        state_machine=dict(
            starting_state=S.initializing,
            starting_action=initialize,
            exit_event=E.exited,
            transitions=transitions
        ),

        argument_spec=dict(
            module_arg1=dict(required=True, type='str'),
            module_arg2=dict(required=False, default=100, type='int')
        )
    )

    # No further code is needed in `main()`. The AnsibleModuleFSM object
    # will call the functions/actions below in proper order until the exit_event
    # is returned or `fail_json` or `exit_json` is called.

# =======
# ACTIONS
# =======

def create_obj(module, data):
    #
    # Do stuff here
    #
    # The `module` argument is a standard AnsibleModule object
    # while `data` is a standard dict that gets passed from one function
    # to another as the state machine progresses through the transitions.
    #
    return E.create_started, data

def delete_obj(module, data):
    #
    # Do stuff here
    #
    return E.delete_started, data

def choose_action(module, data):
    if data['some_key'] == 'value1':
        event = E.must_create
    elif data['some_key'] == 'value2':
        event = E.must_delete
    else:
        event = E.must_exit

    return event, data

def exit_now(module, data):
    #
    # Do stuff here
    #
    module.exit_json(msg="Done with the thing.")

def get_obj(module, data):
    data['some_key'] = 'value1'
    return E.found, data

def initialize(module, data):
    #
    # Do stuff here
    #
    return E.initialized, data

def wait(module, data):
    #
    # Do stuff here
    #
    return E.must_exit, data

# ======

if __name__ == '__main__':
    main()
```

For more information on AnsibleModuleFSM, please see its docstring documentation.
##### AUTOGENERATED STATE CHART

By uncommenting the `render_path="/tmp/mymodule"` keyword argument in the example above and executing the module (via `$ hacking/test-module -m path/to/mymodule.py -a "module_arg1='something'"` or via the normal playbook way), a chart is automatically rendered (provided that graphviz is installed). The following image is an actual rendering generated by the above code:

Note that if the `render_path` argument is set, the module will always fail immediately after rendering the chart. This is so that the developer will not accidentally release a module with the rendering enabled since this can negatively impact module performance.
###### ADVANTAGES OF CHANGE
- State machines are excellent for modules that have complex decision trees that would otherwise have to be implemented with multiple nested if conditions;
- Separates workflow from action implementation. The workflow is clearly stated in the transition table whereas the action implementations are defined in their own functions;
- Auto-generation of state diagram which is also very useful for debugging/understanding the internals of complex modules.
- Encapsulates all state-machine-related boilerplate code so that modules that do want to use a state machine are as DRY as possible.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -114,6 +114,12 @@ except ImportError:
         print('\n{"msg": "SyntaxError: probably due to installed simplejson being for a different python version", "failed": true}')
         sys.exit(1)
 
+from ansible.module_utils.simple_state_machine import StateMachine, TransitionError
+# This is so that developers who want to use this with AnsibleModuleFSM don't
+# have to know about the simple_state_machine package. The NOQA comment is
+# added to avoid linters from tripping on it.
+from ansible.module_utils.simple_state_machine import create_events, create_states  # NOQA
+
 AVAILABLE_HASH_ALGORITHMS = dict()
 try:
     import hashlib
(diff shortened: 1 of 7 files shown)


mattclay:
CI failure due to multiple unit test failures, as seen in the log here: https://app.shippable.com/runs/58926c3ff2e87f0f00908e6d/2/console

bcoca:
Hi!

relaxdiego:
No worries. Thanks for the feedback!

PR #17590, opened 2016-09-15
Vault fixes

abadger:
##### ISSUE TYPE
- Feature Pull Request
- Bugfix Pull Request
##### COMPONENT NAME

vault
##### ANSIBLE VERSION

```
devel
```
##### SUMMARY

- Fix is_encrypted_file() not resetting the file position
- Make is_encrypted_file() handle files opened in both text and binary mode
- Fix use of format string without format operator
- Remove unnecessary use of six.b()
- Change is_encrypted to take either text or byte string
- Have both is_encrypted and is_encrypted_file validate that all the data they are given is ascii (vault data is ascii-only at this time)
- Mark vault cipher methods other than encrypt() and decrypt() as private.
- Rename all variables and parameters that are for byte strings with the b_\* prefix.
- Normalize on plaintext/ciphertext/vaulttext for variable and parameter names rather than the generic data (and slightly less generic enc_data/dec_data).
- VaultAES._is_equal() now throws a TypeError if given a non-byte string.
- Mark VaultAES methods that don't use self as either classmethod or staticmethod.
- Get rid of unused VaultFile
- Re-combine VaultLib.encrypt and VaultLib.encrypt_bytestring as the bug that caused their split was actually caused by something else.
- Print a deprecation method when VaultAES is used.
- Test changes:
  - Remove redundant tests( both checking the same byte string)
  - Import mock from ansible.compat so they work with python3 's builtin mock.  No need to install it from pypi.
  - Enable vault editor tests on python3
  - Initialize the vault_cipher for testing in setUp()
  - Initialize the VaultLib instance for testing in setUp()
  - Give asserttrue and assertFalse the actual method calls for better error messages
  - Convert assert into a unittest assert method call for better error messages
  - test that _is_equal compares non-ascii byte strings correctly
  - test that _is_equal throws TypeError on text strings and int.
  - Remove test_methods_exist(): This isn't really needed as we only have one VaultLib implementation
  - Add tests for both text and byte string input where the API allows that.

### DIFF
--- a/lib/ansible/parsing/vault/__init__.py
+++ b/lib/ansible/parsing/vault/__init__.py
@@ -1,4 +1,6 @@
 # (c) 2014, James Tanner <tanner.jc@gmail.com>
+# (c) 2016, Adrian Likins <alikins@redhat.com>
+# (c) 2016, Toshio Kuratomi <tkuratomi@ansible.com>
 #
 # Ansible is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -25,7 +27,6 @@ import tempfile
 import random
 from io import BytesIO
 from subprocess import call
-from ansible.errors import AnsibleError
 from hashlib import sha256
 from binascii import hexlify
 from binascii import unhexlify
@@ -61,7 +62,9 @@ try:
 except ImportError:
     HAS_AES = False
 
-from ansible.compat.six import PY3
+from ansible.compat.six import PY3, binary_type
+from ansible.compat.six.moves import zip
+from ansible.errors import AnsibleError
 from ansible.module_utils._text import to_bytes, to_text
 
 try:
@@ -84,14 +87,13 @@ except Exception as e:
     import traceback
     display.debug("Traceback from import of cryptography was {0}".format(traceback.format_exc()))
 
-
 HAS_ANY_PBKDF2HMAC = HAS_PBKDF2 or HAS_PBKDF2HMAC
+
 
 CRYPTO_UPGRADE = "ansible-vault requires a newer version of pycrypto than the one installed on your platform." \
     " You may fix this with OS-specific commands such as: yum install python-devel; rpm -e --nodeps python-crypto; pip install pycrypto"
 
 b_HEADER = b'$ANSIBLE_VAULT'
-HEADER = '$ANSIBLE_VAULT'
 CIPHER_WHITELIST = frozenset((u'AES', u'AES256'))
 CIPHER_WRITE_WHITELIST = frozenset((u'AES256',))
 # See also CIPHER_MAPPING at the bottom of the file which maps cipher strings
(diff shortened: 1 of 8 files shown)


lib/ansible/parsing/vault/__init__.py
```diff
@@ -123,16 +135,25 @@ def is_encrypted(b_data):
...
-
     :arg file_obj: A file object that will be read from.
     :returns: True if the file is a vault file. Otherwise, False.
     """
     # read the header and reset the file stream to where it started
     current_position = file_obj.tell()
-    b_header_part = file_obj.read(len(b_HEADER))
-    file_obj.seek(current_position)
+    try:
+        file_obj.seek(0)
```
alikins: Not sure if it matters, but the existing behavior was to read from the current position instead of resetting
to the beginning. I think everything that actually called it did so with file pointer at 0, so not sure it matters.

lib/ansible/parsing/vault/__init__.py
```diff
@@ -143,48 +164,41 @@ def __init__(self, password):
         self.cipher_name = None
         self.b_version = b'1.1'
 
-    # really b_data, but for compat
-    def is_encrypted(self, data):
+    @staticmethod
+    def is_encrypted(data):
```
alikins, line 173: I think these should stay regular methods. That way if VaultLib.is_encrypted() could be more
specific than the module scope version. The primary case I'm thinking of is checking if the data
is encrypted by the key that the VaultLib instance knows about. ie, 'is this file something _I_ can decrypt'

(granted at the moment, nothing uses that so it doesn't matter...)

test/units/cli/test_galaxy.py
```diff
@@ -24,11 +24,9 @@
 import tarfile
 import tempfile
 
-from mock import patch
-
 from ansible.compat.six import PY3
 from ansible.compat.tests import unittest
-from mock import patch, call
+from ansible.compat.tests.mock import call, patch
```
alikins, line 29: These unit tests changes seem unrelated?

alikins (changes requested):
Changes look okay to me. The thing that needs addressing is the unrelated unit test changes.

lib/ansible/parsing/vault/__init__.py
alikins: They could likely also go away in favor of using the module versions everywhere. They were left in for compat, but that may not be needed.

lib/ansible/parsing/vault/__init__.py
abadger: Hmm... That's an interesting point.  If we don't seek(0) then we depend on the calling code to do the right thing when setting up the filehandle.  OTOH, it allows calling code to do things like "read this vault data from the middle of this email message".  What if we add a new parameter?

``` python
def  is_encrypted_file(file_obj, file_pos=0):
    [...]
    file_obj.seek(file_pos)
```

lib/ansible/parsing/vault/__init__.py
abadger: <nod>  I already put deprecation messages in with the idea that they would go away in 2.4.  If we do code up a context-aware is_encrypted function, we'll then be able to bring them back with the new behaviour if we want.

test/units/cli/test_galaxy.py
abadger: They are but the change to use ansible.compat.tests.mock is needed in all of them... If you're ready for it to be committed, I can leave them in a separate commit when I squash.  (I'll jsut have to remember to do a merge commit in the github UI rather than a squash).

lib/ansible/parsing/vault/__init__.py
alikins: Interesting. Does this help the scenario where we needed to read the entire file to get the encoding?

test/units/cli/test_galaxy.py
alikins: Yup, thats fine with me.

alikins (approved):

lib/ansible/parsing/vault/__init__.py
abadger: It gives the caller the ability to optimize if they want but it will always be a tradeoff between accuracy and speed.  The caller could specify a count of 2048 for instance so that the header and a few bytes of the vaulttext are read.  That would be faster then reading the whole file.  But it could pass but when the rest of the vaulttext is later read, we could find out that there was non-ascii data after the first 2048 bytes and then decryption would fail.

PR #17598, opened 2016-09-15
More verbose ssh failure error messages.

alikins:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

lib/ansible/plugins/connection/ssh.py
##### ANSIBLE VERSION

```
ansible 2.2.0 (minimal_better_ssh_failures 6b5627ad93) last updated 2016/09/15 17:05:00 (GMT -400)
  lib/ansible/modules/core: (detached HEAD 48d932643b) last updated 2016/09/15 17:01:15 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD aa45bd8a94) last updated 2016/09/15 17:01:15 (GMT -400)
  config file = /home/adrian/ansible/ansible.cfg
  configured module search path = ['/home/adrian/src/ansible-modules-core', '/home/adrian/src/ansible-modules-extras']

```
##### SUMMARY

Don't force 'ssh -q' for verbosity < 'vvv'.

Include ssh stderr in exception **str** for
ssh connection errors.

Fixes #15706

PR #17602, opened 2016-09-15
Unescape cli output and error messages

alikins:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

lib/ansible/utils/display.py
##### ANSIBLE VERSION

```
ansible 2.2.0 (minimal_better_ssh_failures 456a6bc9ad) last updated 2016/09/15 18:55:37 (GMT -400)
  lib/ansible/modules/core: (detached HEAD 48d932643b) last updated 2016/09/15 17:01:15 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD aa45bd8a94) last updated 2016/09/15 17:01:15 (GMT -400)
  configured module search path = ['/home/adrian/src/ansible-modules-core', '/home/adrian/src/ansible-modules-extras']

```
##### SUMMARY

Use escape_decode() to unescape things like '\n'
from error messages and other strings.

Especially useful when used with https://github.com/ansible/ansible/pull/17598

### DIFF
--- a/lib/ansible/utils/display.py
+++ b/lib/ansible/utils/display.py
@@ -18,6 +18,7 @@
 from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
+import codecs
 import fcntl
 import textwrap
 import os
@@ -105,6 +106,10 @@ class Display:
 
         # FIXME: this needs to be implemented
         #msg = utils.sanitize_output(msg)
+
+        msg, unused = codecs.escape_decode(msg.encode('utf8'))
+        msg = msg.decode()
+
         nocolor = msg
         if color:
             msg = stringc(msg, color)


abadger:
Why are those ending up in the strings in the first place?

abadger:
Thought about this a little more -- I don't think we can do this at the display level.  Doing it in display means that things that actually want to send strings with those sequences in them will have to double escape the sequences.  Instead, do the unescaping before we send the string into display.

-1 on this PR.

abadger:
We talked about this quite a bit and came to the conclusion that the callbacks are where we probably needed to do this.  Some poking around found that in _dump_results() we call `json.dumps(abridged_result)`.  That's what is escaping the newline characters.

We can't unescape all of the results because then short variables will be unreadable and people who are looking at the results to know how to register a var and then get at the data inside will be confused.  So a tentative idea is to somehow mark (or whitelist) some result fields that should be displayed in a different manner.  Here's some proof of concept code for doing that for the output of ping data=crash

```
diff --git a/lib/ansible/plugins/callback/__init__.py b/lib/ansible/plugins/callback/__init__.py
index d02e5ac..a8aca86 100644
--- a/lib/ansible/plugins/callback/__init__.py
+++ b/lib/ansible/plugins/callback/__init__.py
@@ -105,8 +105,19 @@ class CallbackBase:
         # remove exception from screen output
         if 'exception' in abridged_result:
             del abridged_result['exception']
+        import codecs

-        return json.dumps(abridged_result, indent=indent, ensure_ascii=False, sort_keys=sort_keys)
+        stdout = abridged_result.pop('module_stdout', None)
+        stderr = abridged_result.pop('module_stderr', None)
+
+        output = json.dumps(abridged_result, indent=indent, ensure_ascii=False, sort_keys=sort_keys)
+
+        if stdout is not None:
+            output = '%s\nstdout:\n%s\n' % (output, stdout)
+        if stderr is not None:
+            output = '%s\nstderr:\n%s\n' % (output, stderr)
+
+        return output
```

```
 def _handle_warnings(self, res):
     ''' display warnings, if enabled and any exist in the result '''
```

```


This doesn't handle the output from ssh.  For that we have to figure out what field the results data is being put into and similarly handle it differently.
```

alikins:
I think I found a variation of https://github.com/ansible/ansible/pull/17598 that will let us avoid
the need for unescape.

The branch:

https://github.com/ansible/ansible/compare/devel...alikins:minimal_better_ssh_failures_no_escape?expand=1

The gist of it is that we include the exception in the info in a TaskResult when there is a connection failure.  The AnsibleSshConnectionFailure added in #17598 has an attribute for 'stderr' that is populated from the ssh connection plugin. Then we move the code that formats the stderr for display into the default callback (default.py).  The callback can then look at that exception (including its 'stderr' attr) and display the results.

That keeps the default display logic in default.py, but also allows other callback plugins to display info about the exception if they want to.

alikins:
closing for now. We can revisit post 2.2.

PR #17604, opened 2016-09-16
module_utils/dimensiondata

relaxdiego:
##### ISSUE TYPE

New Module Pull Request
##### COMPONENT NAME

dimensiondata
##### ANSIBLE VERSION

```
ansible 2.2.0 (for-upstream-pr bef03042b8) last updated 2016/09/15 12:53:04 (GMT -700)
  lib/ansible/modules/core: (detached HEAD db5cb54b23) last updated 2016/09/08 14:25:22 (GMT -700)
  lib/ansible/modules/extras: (for_devel 0b5e7136aa) last updated 2016/09/15 16:39:58 (GMT -700)
  config file = 
```
##### SUMMARY

This change contains common code required by the dimensiondata modules. 
##### SPECIAL NOTE
- Requires `apache-libcloud`
##### REQUIRED BY
- TBD

### DIFF
--- a/lib/ansible/module_utils/dimensiondata.py
+++ b/lib/ansible/module_utils/dimensiondata.py
@@ -68,21 +68,21 @@ def get_credentials():
 
     To set in environment:
 
-        export DIDATA_USER='myusername'
-        export DIDATA_PASSWORD='mypassword'
+        export MCP_USER='myusername'
+        export MCP_PASSWORD='mypassword'
 
     To set in dot file place a file at ~/.dimensiondata with
     the following contents:
 
         [dimensiondatacloud]
-        DIDATA_USER: myusername
-        DIDATA_PASSWORD: mypassword
+        MCP_USER: myusername
+        MCP_PASSWORD: mypassword
     """
     check_libcloud_or_fail()
 
     # Attempt to grab from environment
-    user_id = os.environ.get('DIDATA_USER', None)
-    key = os.environ.get('DIDATA_PASSWORD', None)
+    user_id = os.environ.get('MCP_USER', None)
+    key = os.environ.get('MCP_PASSWORD', None)
 
     # Environment failed try dot file
     if not user_id or not key:
@@ -90,8 +90,8 @@ def get_credentials():
         config = ConfigParser.RawConfigParser()
         config.read("%s/.dimensiondata" % home)
         try:
-            user_id = config.get("dimensiondatacloud", "DIDATA_USER")
-            key = config.get("dimensiondatacloud", "DIDATA_PASSWORD")
+            user_id = config.get("dimensiondatacloud", "MCP_USER")
+            key = config.get("dimensiondatacloud", "MCP_PASSWORD")
         except (ConfigParser.NoSectionError, ConfigParser.NoOptionError):
             pass
 
--- a/lib/ansible/module_utils/dimensiondata.py
+++ b/lib/ansible/module_utils/dimensiondata.py
@@ -1,0 +1,367 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2016 Dimension Data
+#
+# This module is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This software is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this software.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Authors:
+#   - Aimon Bustardo <aimon.bustardo@dimensiondata.com>
+#
+# Common methods to be used by versious module components
+import os
+import sys
+import ConfigParser
+from os.path import expanduser
+from libcloud.common.dimensiondata import API_ENDPOINTS
+from libcloud.common.dimensiondata import DimensionDataAPIException
+from uuid import UUID
+
+
+# -------------------------
+# This method will get user_id and key from environemnt or dot file.
+# Environment takes priority.
+#
+# To set in environment:
+# export DIDATA_USER='myusername'
+# export DIDATA_PASSWORD='mypassword'
+#
+# To set in dot file place a file at ~/.dimensiondata with
+# the following contents:
+# [dimensiondatacloud]
+# DIDATA_USER: myusername
+# DIDATA_PASSWORD: mypassword
+# -------------------------
+def get_credentials():
+    user_id = None
+    key = None
+
+    # Attempt to grab from environment
+    if 'DIDATA_USER' in os.environ and 'DIDATA_PASSWORD' in os.environ:
+        user_id = os.environ['DIDATA_USER']
+        key = os.environ['DIDATA_PASSWORD']
+
+    # Environment failed try dot file
+    if user_id is None or key is None:
+        home = expanduser('~')
+        config = ConfigParser.RawConfigParser()
+        config.read("%s/.dimensiondata" % home)
+        try:
+            user_id = config.get("dimensiondatacloud", "DIDATA_USER")
+            key = config.get("dimensiondatacloud", "DIDATA_PASSWORD")
+        except:
+            pass
+
+    # Return False if either are not found
+    if user_id is None or key is None:
+        return False
+
+    # Both found, return data
+    return dict(user_id=user_id, key=key)
+
+
+# -------------------------
+# Get the list of available regions
+# whos vendor is Dimension Data.
+# -------------------------
+def get_dd_regions():
+    # Get endpoints
+    all_regions = API_ENDPOINTS.keys()
+
+    # filter to only DimensionData endpoints
+    region_list = filter(lambda i: i.startswith('dd-'), all_regions)
+
+    # Strip prefix
+    regions = [region[3:] for region in region_list]
+
+    return regions
+
+
+# ----------------------------------------
+# Get a network domain object by its name
+# ----------------------------------------
+def get_network_domain_by_name(driver, name, location):
+    networks = driver.ex_list_network_domains(location=location)
+    network = filter(lambda x: x.name == name, networks)
+    if len(network) > 0:
+        return network[0]
+    else:
+        return None
+
+
+# ---------------------------------------------
+# Get a network domain object by its name or id
+# ---------------------------------------------
+def get_network_domain(driver, locator, location):
+    if is_uuid(locator):
+        net_id = locator
+    else:
+        networks = driver.ex_list_network_domains(location=location)
+        found_networks = filter(lambda x: x.name == locator, networks)
+        if len(found_networks) > 0:
+            net_id = found_networks[0].id
+        else:
+            return False
+    try:
+        return driver.ex_get_network_domain(net_id)
+    except DimensionDataAPIException:
+        return False
+
+
+# ---------------------------------------------
+# Get a VLAN object by its name or id
+# ---------------------------------------------
+def get_vlan(driver, locator, location, network_domain):
+    if is_uuid(locator):
+        vlan_id = locator
+    else:
+        vlans = driver.ex_list_vlans(location=location,
+                                     network_domain=network_domain)
+        found_vlans = filter(lambda x: x.name == locator, vlans)
+        if len(found_vlans) > 0:
+            vlan_id = found_vlans[0].id
+        else:
+            return False
+    try:
+        return driver.ex_get_vlan(vlan_id)
+    except:
+        return False
+
+
+# ----------------------------------------
+# Get a locations MCP version
+# ----------------------------------------
+def get_mcp_version(driver, location):
+    # Get location to determine if MCP 1.0 or 2.0
+    location = driver.ex_get_location_by_id(location)
+    if 'MCP 2.0' in location.name:
+        return '2.0'
+    return '1.0'
+
+
+# ---------------------
+# Test if valid v4 UUID
+# ---------------------
+def is_uuid(u, version=4):
+    try:
+        uuid_obj = UUID(u, version=version)
+        return str(uuid_obj) == u
+    except:
+        return False
+
+
+# --------------------------------------------
+# Expand public IP block to show all addresses
+# --------------------------------------------
+def expand_ip_block(block):
+    addresses = []
+    ip_r = block.base_ip.split('.')
+    last_quad = int(ip_r[3])
+    address_root = "%s.%s.%s." % (ip_r[0], ip_r[1], ip_r[2])
+    for i in range(int(block.size)):
+        addresses.append(address_root + str(last_quad + i))
+    return addresses
+
+
+# ---------------------------
+# Get public IP block details
+# ---------------------------
+def get_public_ip_block(module, driver, network_domain, block_id=False,
+                        base_ip=False):
+    # Block ID given, try to use it.
+    if block_id is not 'False':
+        try:
+            block = driver.ex_get_public_ip_block(block_id)
+        except DimensionDataAPIException:
+            e = sys.exc_info()[1]
+            # 'UNEXPECTED_ERROR' should be removed once upstream bug is fixed.
+            # Currently any call to ex_get_public_ip_block where the block does
+            # not exist will return UNEXPECTED_ERROR rather than
+            # 'RESOURCE_NOT_FOUND'.
+            if e.code == "RESOURCE_NOT_FOUND" or e.code == 'UNEXPECTED_ERROR':
+                module.exit_json(changed=False, msg="Public IP Block does "
+                                 "not exist")
+            else:
+                module.fail_json(msg="Unexpected error while retrieving "
+                                 "block: %s" % e.code)
+            module.fail_json(msg="Error retreving Public IP Block " +
+                                 "'%s': %s" % (block.id, e.message))
+    # Block ID not given, try to use base_ip.
+    else:
+        blocks = list_public_ip_blocks(module, driver, network_domain)
+        if blocks is not False:
+            block = filter(lambda x: x.base_ip == base_ip, blocks)[0]
+        else:
+            module.exit_json(changed=False, msg="IP block starting with "
+                             "'%s' does not exist." % base_ip)
+    return block
+
+
+# --------------------------------
+# Get list of NAT rules for domain
+# --------------------------------
+def list_nat_rules(module, driver, network_domain):
+    try:
+        return driver.ex_list_nat_rules(network_domain)
+    except DimensionDataAPIException:
+        e = sys.exc_info()[1]
+        module.fail_json(msg="Failed to list NAT rules: %s" % e.message)
+
+
+# -----------------------------------------
+# Get list of public IP blocks for a domain
+# -----------------------------------------
+def list_public_ip_blocks(module, driver, network_domain):
+    try:
+        blocks = driver.ex_list_public_ip_blocks(network_domain)
+        if len(blocks) == 0:
+            return []
+        else:
+            return blocks
+    except DimensionDataAPIException:
+        e = sys.exc_info()[1]
+        module.fail_json(msg="Error retreving Public IP Blocks: %s" % e)
+
+
+# --------------------------------------
+# Get public IP block allocation details
+# --------------------------------------
+def get_block_allocation(module, cp_driver, lb_driver, network_domain, block):
+    #  Shows all ips in block and if they are allocated
+    # ex: {'id': 'eb8b16ca-3c91-45fb-b04b-5d7d387a9f4a',
+    #       'addresses': [{'address': '162.2.100.100',
+    #                      'allocated': True
+    #                      },
+    #                      {'address': '162.2.100.101',
+    #                       'allocated': False
+    #                      }
+    #                     ]
+    #     }
+    nat_rules = list_nat_rules(module, cp_driver, network_domain)
+    balancers = list_balancers(module, lb_driver)
+    pub_ip_block = get_public_ip_block(module, cp_driver, network_domain,
+                                       block.id, False)
+    pub_ips = expand_ip_block(pub_ip_block)
+    block_detailed = {'id': block.id, 'addresses': []}
+    for ip in pub_ips:
+        allocated = False
+        nat_match = filter(lambda x: x.external_ip == ip, nat_rules)
+        lb_match = filter(lambda x: x.ip == ip, balancers)
+        if len(nat_match) > 0 or len(lb_match) > 0:
+            allocated = True
+        else:
+            allocated = False
+        block_detailed['addresses'].append({'address': ip,
+                                            'allocated': allocated})
+    return block_detailed
+
+
+def list_balancers(module, lb_driver):
+    try:
+        return lb_driver.list_balancers()
+    except DimensionDataAPIException:
+        e = sys.exc_info()[1]
+        module.fail_json(msg="Failed to list Load Balancers: %s" % e.message)
+
+
+# --------------------------------------
+# Get list of public IP blocks with one
+# or more unallocated IPs
+# --------------------------------------
+def get_blocks_with_unallocated(module, cp_driver, lb_driver, network_domain):
+    # Gets ip blocks with one or more unallocated IPs.
+    # ex:
+    #   {'unallocated_count': <total count of unallocated ips>,
+    #    'ip_blocks': [<list of expanded blocks with details
+    #                  (see get_block_allocation())>],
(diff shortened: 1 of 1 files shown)


lib/ansible/module_utils/dimensiondata.py
```diff
@@ -0,0 +1,367 @@
...
+#
+# Authors:
+#   - Aimon Bustardo <aimon.bustardo@dimensiondata.com>
+#
+# Common methods to be used by versious module components
+import os
+import sys
+import ConfigParser
+from os.path import expanduser
+from libcloud.common.dimensiondata import API_ENDPOINTS
```
bcoca: cannot assume non 'core' modules are in the system, this should attempt to import and then allow for the AnsibleModule class to report about the missing module in a 'friendly' way, see ec2.py for examples.

lib/ansible/module_utils/dimensiondata.py
```diff
@@ -0,0 +1,367 @@
...
+
+
+# ---------------------------------------------
+# Get a network domain object by its name or id
+# ---------------------------------------------
+def get_network_domain(driver, locator, location):
+    if is_uuid(locator):
+        net_id = locator
+    else:
+        networks = driver.ex_list_network_domains(location=location)
```
bcoca: don't all API calls need to capture exceptions?

lib/ansible/module_utils/dimensiondata.py
```diff
@@ -0,0 +1,367 @@
...
+# ---------------------------
+# Get public IP block details
+# ---------------------------
+def get_public_ip_block(module, driver, network_domain, block_id=False,
+                        base_ip=False):
+    # Block ID given, try to use it.
+    if block_id is not 'False':
+        try:
+            block = driver.ex_get_public_ip_block(block_id)
+        except DimensionDataAPIException:
```
bcoca: Note: I'm assuming libcloud requires python >2.6, as such you can use `Exception as e`, the exc_info is only required for modules that are 2.4 compatible, which most cloud modules cannot be due to library dependencies.

bcoca (changes requested):
Just from cursory review, a couple of issues, I only made one comment of each but both are repeated in several spots.

lib/ansible/module_utils/dimensiondata.py
aimonb: Exception handling is done in module code. This way we can keep the handling of the error specific to the module using the shared code.

lib/ansible/module_utils/dimensiondata.py
aimonb: Typically we always use "except DimensionDataAPIException as e". This will be updated to reflect out standard process.

relaxdiego: @bcoca @aimonb using `Exception as e` causes some builds in shippit to fail though.

lib/ansible/module_utils/dimensiondata.py
bcoca: it seems inconsistent as you do capture it in some cases

lib/ansible/module_utils/dimensiondata.py
bcoca: a) was just note, not required to pass review
b) if libcloud already requries 2.6 we can make an exception in the check as we do for other cloud modules.

lib/ansible/module_utils/dimensiondata.py
aimonb: Yes in some cases the error is always the same so we return False to ease module development. In other cases there are multiple error possibilities from API, some of which we ignore in one case and not another.

aimonb: If the above is an issue we can make all return the exception and consistently handle them in the module.

lib/ansible/module_utils/dimensiondata.py
aimonb: I vote for consistency of "except DimensionDataAPIException as e"

lib/ansible/module_utils/dimensiondata.py
relaxdiego: +1 on raising a custom exception instead of failing silently.

relaxdiego:
Rebased to ansible/devel

gundalow:
@relaxdiego @mikeoutland This PR was added to https://github.com/ansible/community/issues/124 we have discussed this and as we are now in feature freeze for 2.2 this (and related PRs) will not be reviewed/merged till _after_ Ansible 2.2 has reached rc1 (release candidate 1). At that point we will have branched and allow new PRs to be merged.

lib/ansible/module_utils/dimensiondata.py
```diff
@@ -0,0 +1,393 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this software.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Authors:
+#   - Aimon Bustardo <aimon.bustardo@dimensiondata.com>
+#   - Mark Maglana   <mmaglana@gmail.com>
+#
+# Common methods to be used by versious module components
+import os
+import ConfigParser
```
abadger: This formulation of importing configparser is not compatible with python3.  You can use

``` python
from ansible.module_utis.six.moves import configparser
```

lib/ansible/module_utils/dimensiondata.py
```diff
@@ -0,0 +1,393 @@
...
+    check_libcloud_or_fail()
+
+    # Get endpoints
+    all_regions = API_ENDPOINTS.keys()
+
+    # filter to only DimensionData endpoints
+    region_list = filter(lambda i: i.startswith('dd-'), all_regions)
+
+    # Strip prefix
+    regions = [region[3:] for region in region_list]
```
abadger: List comprehensions and generator expressions are more pythonic than using filter.  This one can be rewritten as a single listcomprehension or a combination of as two steps:

``` python
# One combined list comprehension
regions = [r[3:] for r in all_regions if r.startswith('dd-')]

# two steps, using a generator expression
region_list = (r for r in all_regions if r.startswith('dd-'))
regions = [r[3:] for r in region_list]
```

lib/ansible/module_utils/dimensiondata.py
```diff
@@ -0,0 +1,393 @@
...
+
+    return regions
+
+
+def get_network_domain_by_name(driver, name, location):
+    """
+    Get a network domain object by its name
+    """
+    networks = driver.ex_list_network_domains(location=location)
+    found_networks = filter(lambda x: x.name == name, networks)
```
abadger: This usage of filter won't work on python3.  filter on python3 will return a generator which will evaluate to true in the conditional below.  Use a list comprehension so that found_networks ends up with an empty list on python3 which will evaluate to false.

abadger: Same change needs to be made to many other uses of filter.

lib/ansible/module_utils/dimensiondata.py
```diff
@@ -0,0 +1,393 @@
...
+    return addresses
+
+
+def get_public_ip_block(module, driver, network_domain, block_id=False,
+                        base_ip=False):
+    """
+    Get public IP block details
+    """
+    # Block ID given, try to use it.
+    if block_id is not 'False':
```
abadger: I don't think this works?  block_id defaults to the boolean False in the parameters to this function.  So you probably need to check:

``` python
if block_id is not False:
```

rather than checking against a string.

lib/ansible/module_utils/dimensiondata.py
```diff
@@ -0,0 +1,393 @@
...
+            else:
+                module.fail_json(msg="Unexpected error while retrieving "
+                                 "block: %s" % e.code)
+            module.fail_json(msg="Error retreving Public IP Block " +
+                                 "'%s': %s" % (block.id, e.message))
+    # Block ID not given, try to use base_ip.
+    else:
+        blocks = list_public_ip_blocks(module, driver, network_domain)
+        if blocks is not False:
+            block = filter(lambda x: x.base_ip == base_ip, blocks)[0]
```
abadger: This will fail on python3 for a similar reason as the conditional check noted above.  filter returns a generator.  The generator can't be indexed.

Use a list comprehension instead:

``` python
block = [b for b in blocks if b.base_ip == base_ip]
```

lib/ansible/module_utils/dimensiondata.py
```diff
@@ -0,0 +1,393 @@
...
+        else:
+            module.exit_json(changed=False, msg="IP block starting with "
+                             "'%s' does not exist." % base_ip)
```
abadger, line 262: the message here looks like a failure case rather than an exit case... Perhaps it should use module.fail_json or the message should be changed?

Stylistically, exit_json() cases should retun back to the main function in the module and let the exit_json() occur there.  It's more flexible to changing needs inside of the module than calling exit_json() here.

lib/ansible/module_utils/dimensiondata.py
```diff
@@ -0,0 +1,393 @@
...
+                                                              cp_driver,
+                                                              lb_driver,
+                                                              network_domain)
+        free_ips = blocks_with_unallocated['unallocated_addresses']
+    if len(free_ips) < count:
+        num_needed = count - len(free_ips)
+        for i in range(num_needed):
+            block = cp_driver.ex_add_public_ip_block_to_network_domain(
+                network_domain)
+            b_dict = get_block_allocation(module, cp_driver, lb_driver,
```
abadger: Just a style note... If you could use block_dict instead of b_dict that would be good.  We use the "b_" prefix as a convention to tell when we're dealing with byte strings as opposed to text strings.

lib/ansible/module_utils/dimensiondata.py
```diff
@@ -0,0 +1,393 @@
...
+
+    # Environment failed try dot file
+    if not user_id or not key:
+        home = expanduser('~')
+        config = ConfigParser.RawConfigParser()
+        config.read("%s/.dimensiondata" % home)
+        try:
+            user_id = config.get("dimensiondatacloud", "DIDATA_USER")
+            key = config.get("dimensiondatacloud", "DIDATA_PASSWORD")
+        except ConfigParser.NoSectionError, ConfigParser.NoOptionError:
```
abadger: I don't think this will work correctly on python2.  Pretty sure you need parenthesis there:

``` python
except (ConfigParser.NoSectionError, ConfigParser.NoOptionError):
```

lib/ansible/module_utils/dimensiondata.py
```diff
@@ -0,0 +1,393 @@
...
+            pass
+
+    # One or more credentials not found. Function can't recover from this
+    # so it has to raise an error instead of fail silently.
+    if not user_id:
+        raise MissingCredentialsError("Dimension Data user id not found")
+    elif not key:
+        raise MissingCredentialsError("Dimension Data key not found")
```
abadger, line 125: Note: Our modules usually take credentials as parameters and fall back to files and environment variables if the credentials aren't present in the parameters.  There's multiple ways to implement that, though, with processing the parameters first or the library function processing the parameters in addition to the config files and env vars.  Not sure which way you'd want to fix this.

lib/ansible/module_utils/dimensiondata.py
```diff
@@ -0,0 +1,393 @@
...
+
+    # Both found, return data
+    return dict(user_id=user_id, key=key)
+
+
+def get_dd_regions():
+    """
+    Get the list of available regions whose vendor is Dimension Data.
+    """
+    check_libcloud_or_fail()
```
abadger, line 135: Note that this won't work in the current forumlation of the modules (because the exception that check_libcloud_or_fail() raises goes uncaught in the module).  I'll write more about that in the module, though, as that's probably where it will have to be fixed.

abadger:
needs_revision

mattclay:
@relaxdiego Tests are failing due to a python 2.4 syntax error:

```
2016-12-13 01:48:55 Compiling ./lib/ansible/module_utils/dimensiondata.py ...
2016-12-13 01:48:55   File "./lib/ansible/module_utils/dimensiondata.py", line 221
2016-12-13 01:48:55     except DimensionDataAPIException as e:
2016-12-13 01:48:55                                       ^
2016-12-13 01:48:55 SyntaxError: invalid syntax
```

Here's an example of how to handle exceptions so it will work in python 2.4:

https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/system/user.py#L317

relaxdiego:
@mattclay I'm no longer associated with Dimension Data. Hopefully somebody in the company can pick up the slack. :-)

gundalow:
Do you know the GitHub account of anyone else there we could ping? 
Enjoy your new adventures.

relaxdiego:
Thanks @gundalow! Maybe @tonybaloney has an idea.

tonybaloney:
@tintoy is the guy

tintoy:
Cheers - I'll take a proper look at this tomorrow :)

From first impressions all the requested fixes and other changes look pretty minor. Give me an hour or 2 tomorrow to get familiar with the code and then hopefully I should be able to make the requested changes by the end of the day (AU EST).

gundalow:
Since we have done the repo merge now, you may wish to add the empty file `lib/ansible/modules/cloud/dimensiondata/__init__.py` into this PR, which will make it easier to add the modules

tintoy:
> Since we have done the repo merge now, you may wish to add the empty file `lib/ansible/modules/cloud/dimensiondata/__init__.py` into this PR, which will make it easier to add the modules

Hi, just want to confirm before I proceed (sorry, I'm new here 😊) - it's definitely `lib/ansible/modules/cloud/dimensiondata/__init__.py`?
It's just that there are only 2 folders currently under `lib/ansible/modules`, `core` and `extras`, but there is a `cloud` under `extras`.

lib/ansible/module_utils/dimensiondata.py
tintoy: Sorry, just checking - do you mean:

```python
from ansible.module_utis.six.moves import configparser
```
or
```python
from ansible.module_utis.six.moves import ConfigParser
```

(I've never seen this code before so it's a lot to take in, but it looks like the latter rather than the former?)

tintoy: BTW, apologies for asking potentially stupid questions, but I'm jumping into this blind - it's been a while since I've written anything fancier than a simple script in Python and I'm a little rusty:

I can find `ansible/module_utils/six.py` but I'm not sure where `moves` comes from?

tintoy:
Ok, sorry, it looks like this may take me a while - I'm having a little difficulty understanding how the branching and sub-modules have been set up (hooray for multiple remotes). I'll get back to you once I've sorted that all out.

tintoy:
Ah, ok - repomerge. I think I get it now. Sorry for the confusion.

Just doing a quick rebase, then I can address the issues raised above.

lib/ansible/module_utils/dimensiondata.py
tintoy: BTW, if I'm importing it as `configparser` I assume references to the `ConfigParser` class are still valid (again, sorry if this is a stupid question)?

lib/ansible/module_utils/dimensiondata.py
tintoy: Thanks - I've replaced all uses of `filter` with comprehensions.

lib/ansible/module_utils/dimensiondata.py
tintoy: I'll come back to this after we've sorted out the other issues you've noted :)

tintoy:
Ok, I've hopefully addressed most of the issues raised but there are one or 2 that I think may need a little discussion (e.g. `check_libcloud_or_fail`) before I understand them well enough to proceed.

tintoy:
Ok - the remaining failing tests seem to be unrelated to these changes; are you guys happy with the fixes made so far?

gundalow:
I should have been clearer and said a rebate was needed, glad you've got that sorted. 

I've hit the retry button on Shippable,  I'm fairly confident they'll turn green. 

What's needed to progress `check_libcloud_or_fail`? 

Was there another issue as well?

tintoy:
> I should have been clearer and said a rebate was needed, glad you've got that sorted.

No worries, I should have done my homework - won't make that mistake again :)

> What's needed to progress `check_libcloud_or_fail`?

@abadger wrote:

> Note that this won't work in the current forumlation of the modules (because the exception that check_libcloud_or_fail() raises goes uncaught in the module). I'll write more about that in the module, though, as that's probably where it will have to be fixed.

I was just wondering if this meant that they hadn't written their additional notes yet, or if this referred to something else in the review notes. Either way, I (unfortunately) don't really know what this means. Sorry, I appreciate that my ignorance is kinda wasting your time, but I lack a fair bit of domain knowledge here - I use Ansible fairly frequently (especially with Terraform) for prototyping but have only a shallow understanding of the internals. If you can point me at any docs that describe the area they're referring to I'm happy to go off and do my own research :)

As for the other thing:

@abadger wrote:

> Note: Our modules usually take credentials as parameters and fall back to files and environment variables if the credentials aren't present in the parameters. There's multiple ways to implement that, though, with processing the parameters first or the library function processing the parameters in addition to the config files and env vars. Not sure which way you'd want to fix this.

This seems like more of a design change than fixing things up. I'm happy to do it if that's best-practice, but it'll take a little longer to get familiar enough with the codebase to make that sort of change (I can probably work it out, but it might take a couple of days).

tintoy:
@gundalow - I wonder if I am perhaps still being tripped up by the repo-merge concept; is what I'm missing the fact that this stuff in the core needs to be merged first, _then_ the code from the old repo (what used to be a sub-module) that uses it can be merged?

I'm thinking this because I can see the `dimensiondata` module in `lib/ansible/module_utils/dimensiondata.py`, but nowhere it's used in this repository.

tintoy:
BTW, if I'm right about the repo-merge [in my comment above](https://github.com/ansible/ansible/pull/17604#issuecomment-266895283), then I think we're good to merge this PR (from my end at least). Then I can make the other changes as part of merging the `ansible-modules-extras` work.

mattclay:
I see two `dimensiondata` PRs in `ansible/ansible` (this one and another with tests).

There are also 9 module PRs in the old `ansible/ansible-modules-extras` repository which depend on this PR:

https://github.com/ansible/ansible-modules-extras/pulls?utf8=%E2%9C%93&q=is%3Apr%20is%3Aopen%20dimensiondata

The module_utils PR will need to be either merged before, or at the same time as one of the modules that uses it. There's not much point in merging the module_utils without at least one module, since nothing will be using them.

Ideally the tests would be merged at the same time as the module they test so we can be sure they're passing.

The module PRs will need to be migrated using the PR mover, as noted in the comments on those PRs.

@gundalow may have a different opinion, but it may be best to merge the module_utils code along with one module and the tests for that module as one PR. Then migrate the remaining modules (hopefully along with tests), each as their own PR.

tintoy:
Thanks @mattclay, that makes sense. I'll look into the related PRs and see about getting the simplest one migrated.

tintoy:
@mattclay as for the other PR, that's a little tricker since the PR is from a personal fork that I don't have write access to. I'll talk to the owner (@relaxdiego) and see how he wants to proceed here.

tintoy:
Ugh - the PR mover only works for personal forks. I'm trying to avoid a repeat of the existing situation, so perhaps I'm better off manually moving the other PRs?

mattclay:
@tintoy You should be able to migrate a PR even if you aren't the owner, as long as you have a fork of the `ansible/ansible` repository to migrate it to. What error are you seeing when you try to migrate the PR?

tintoy:
Ah, sorry I was talking about 2 different things :)

The [PR that I don't own](https://github.com/ansible/ansible/pull/17802) is _already_ in the `ansible/ansible` repository. I can't migrate those using the tool, can I? I thought it's only for PRs from the `ansible/ansible-modules-core` or `ansible/ansible-modules-extras` repositories?

As for the other one I've created a personal fork and [migrated the PR to there](https://github.com/ansible/ansible/pull/19325) but it's pretty ugly; we've been trying to keep all of our work in the DimensionDataResearch organisation's fork (rather than personal ones) so we can all work on them. Now anyone from DD who wants to work on that has to be made a contributor in my fork.

I get that this is only due to the repo-merge though, so we'll just do it this way for the remaining PRs. 

All new work after that will be done in the DimensionDataResearch fork (we just need to get these sorted out first).

mattclay:
@tintoy You're correct, the PR mover is only for PRs coming from core and extras.

Feel free to migrate the PRs manually if the PR mover isn't helpful for your workflow. It was intended to make things easier for most users, but in this case it seems it's not helping much.

Sorry for the inconvenience. In the end having everything in one repo is going to be much easier to work with than split across three, especially when it comes to dealing with module_utils and module tests.

tintoy:
@mattclay, no worries - very much agree about single repository. Fortunately that PR is for tests that relate to a feature that depends on _most_ of our other PRs being merged first so it'll keep for now :)

Thanks again for your help!

tintoy:
BTW, that PR mover _is_ a nifty piece of magic and a real time-saver. It'll certainly help for all our other PRs.

tintoy:
Hi @bcoca - are you satisfied with the changes that @relaxdiego made?

bcoca:
still needs to address @abadger's review

lib/ansible/module_utils/dimensiondata.py
tintoy: @abadger - could you elaborate on this a little? For example, [this](https://github.com/ansible/ansible/pull/19325) is the first PR we'll try to merge for the module itself. Do you mean that the module should be using `except LibcloudNotFound` where it calls `get_credentials`? And, if so, that it should call `module.fail_json` when it does? If you wouldn't mind making a note about what you'd like me to do there I'm happy to have a go. Thanks!

lib/ansible/module_utils/dimensiondata.py
tintoy: @abadger - is this the kind of thing you're talking about?

* [change to module_utils](https://github.com/ansible/ansible/pull/17604/commits/bb8817ac1f7818c895abd51f6aff66d670766250)
* [change to module](https://github.com/ansible/ansible/pull/19325/commits/f82888a62a89a42180babb44173d6eb3aab2978e)

tintoy:
> still needs to address @abadger's review

Thanks, dealing with it now.

lib/ansible/module_utils/dimensiondata.py
abadger: @tintoy, yes that's about what I'm talking about.  I'd probably put all that logic into a single function, though.  So you'd call get_credentials(module) and it would first look in the module parameters, then the environment, and finally in the dot files (returning the first set that it found).

lib/ansible/module_utils/dimensiondata.py
abadger: Yes, the revisions you'v emade to this look good to me.

tintoy: Thanks!

abadger: I took a closer look at the module and the concern here is still valid but I have (for real!) left the information in the module PR now :-)  So we can decide what to do there.  It doesn't block this one from going in.

lib/ansible/module_utils/dimensiondata.py
tintoy: Ok, how does this look?

85164a272274514877a6369877e851461a5acf0b

lib/ansible/module_utils/dimensiondata.py
```diff
@@ -61,43 +61,10 @@ def check_libcloud_or_fail():
...
-        key=module.params['mcp_password']
-    )
-
-
-def get_credentials():
-    """
-    This method will get user_id and key from module configuration, environment, or dotfile.
-    Environment takes priority.
+    Get user_id and key from module configuration, environment, or dotfile.
+    Order of priority is environment, dotfile, module.
```
abadger: Near as I can tell the order of priority here is module, environment, dotfile.  Module should be first, so the coded order is fine but the documentation here should agree with the code.

abadger:
@tintoy.  Looks good.  Just a disagreement between the documentation and the code to straighten out.  Once that's fixed, this is ready to merge.

lib/ansible/module_utils/dimensiondata.py
tintoy: Sorry, I think the current logic will only fall back to module if environment variable has not been specified: https://github.com/ansible/ansible/pull/17604/files/6c412fbeefa07e4296b33ec7544c5b1a69a6dcd8..85164a272274514877a6369877e851461a5acf0b#diff-0b1c579a3ccbeed5e582e7693e046cb6R112

I'll change the order, though, to put module first. Give me 5 mins.

tintoy: Done.

tintoy:
Thanks again for your patience, BTW.

abadger:
Merged to devel!  thanks for your patience as well :-)

tintoy:
Great! I'll start looking at the first PR that depends on this :)

PR #17623, opened 2016-09-17
Fixed transport issues when calling self.execute from Cli

dgjustice:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

nxos
##### ANSIBLE VERSION

```
ansible 2.2.0
  config file =
```
##### SUMMARY

PR #17622 introduces another error in nxos.py when a module is using Cli as transport.  The Cli class has no "output" variable in the method definition.  This PR simply adds a try/except block to call the correct version of self.execute from nxos.py.  This fix uncovers other bugs in ansible-modules-core.  I will work on those under a different PR to that repo.

```

```

### DIFF
--- a/lib/ansible/module_utils/nxos.py
+++ b/lib/ansible/module_utils/nxos.py
@@ -201,14 +201,20 @@ class Nxapi(NxapiConfigMixin):
 
         for cmd in commands:
             if output and output != cmd.output:
-                responses.extend(self.execute(cmds, output=output))
+                try:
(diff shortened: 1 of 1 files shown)


dgjustice:
@privateip @gundalow I rebased this off of a clean pull from devel.  I don't know if the try/except thing is kosher, or if you would prefer to check the calling class.  Let me know, and I can whittle it to whatever is best.

PR #17638, opened 2016-09-19
fixes issue where netcli would cause exception with an invalid conditional

privateip:
The Conditional instance will cause a stack trace if the provided conditional
does not map properly to the response.  This fixes that issue so that the
Conditional instance will now raise a FailedConditionalError with the
conditional that caused the failure.

Modules *_command modules (and any other modules that create an instance
of Conditional) should be updated to catch the FailedConditionalError
exception.

### DIFF
--- a/lib/ansible/module_utils/netcli.py
+++ b/lib/ansible/module_utils/netcli.py
@@ -47,6 +47,11 @@ class FailedConditionsError(Exception):
         super(FailedConditionsError, self).__init__(msg)
         self.failed_conditions = failed_conditions
 
+class FailedConditionalError(Exception):
+    def __init__(self, msg, failed_conditional):
+        super(FailedConditionalError, self).__init__(msg)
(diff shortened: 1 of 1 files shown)


lib/ansible/module_utils/netcli.py
```diff
@@ -47,6 +47,11 @@ def __init__(self, msg, failed_conditions):
         super(FailedConditionsError, self).__init__(msg)
         self.failed_conditions = failed_conditions
 
+class FailedConditionalError(Exception):
+    def __init__(self, msg, conditional):
```
alikins: I would expect the arg name here to be 'failed_conditional'  based on the names in FailedConditionsError and the
'self.failed_conditional = conditional' line. But I don't really understand how it's used, so that may not make sense.

privateip: Agreed

PR #17642, opened 2016-09-19
make ec2.py importable

serialdoom:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

contrib/inventory/ec2.py
##### ANSIBLE VERSION

```
all
```
##### SUMMARY

Update `ec2.py` script to make it importable by other scripts.

This will allow people to capture and manipulate the output further than the default.

### DIFF
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -187,6 +187,8 @@ class Ec2Inventory(object):
         elif not self.is_cache_valid():
             self.do_api_calls_update_cache()
 
+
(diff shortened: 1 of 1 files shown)


mattclay (changes requested):
The call to `parse_cli_args` should also be moved, with the parsed args being passed to `Ec2Inventory` when it is instantiated. Otherwise using the class will require accepting the same command line arguments.

ryansb:
An equivalent change to this has been merged, #21637

PR #17644, opened 2016-09-19
Fix http_request for IPv6-only machines

### LINKED ISSUE #7218: Error validating launchpad.net SSL certificate

ei-grad:
ansible.module_utils.urls creates socket with AF_INET family, which
doesn't work for IPv6-only machine.

### DIFF
--- a/lib/ansible/module_utils/urls.py
+++ b/lib/ansible/module_utils/urls.py
@@ -304,6 +304,16 @@ if not HAS_MATCH_HOSTNAME:
 
     HAS_MATCH_HOSTNAME = True
 
+try:
+    from socket import create_connection
+except ImportError:
+    # Just a graceful fallback for py < 2.6, implementing old behaviour.
+    # Wouldn't work with IPv6, but it is unlikely that python 2.4 could be met
+    # on IPv6-only machine.
+    def create_connection(address):
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        sock.connect(address)
+        return sock
 
 # This is a dummy cacert provided for Mac OS since you need at least 1
 # ca cert, regardless of validity, for Python on Mac OS to use the
@@ -682,11 +692,10 @@ class SSLValidationHandler(urllib_request.BaseHandler):
             return req
 
         try:
-            s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
             if https_proxy:
                 proxy_parts = generic_urlparse(urlparse(https_proxy))
                 port = proxy_parts.get('port') or 443
-                s.connect((proxy_parts.get('hostname'), port))
+                s = create_connection((proxy_parts.get('hostname'), port))
                 if proxy_parts.get('scheme') == 'http':
                     s.sendall(self.CONNECT_COMMAND % (self.hostname, self.port))
                     if proxy_parts.get('username'):
@@ -710,7 +719,7 @@ class SSLValidationHandler(urllib_request.BaseHandler):
                 else:
                     raise ProxyError('Unsupported proxy scheme: %s. Currently ansible only supports HTTP proxies.' % proxy_parts.get('scheme'))
             else:
-                s.connect((self.hostname, self.port))
+                s = create_connection((self.hostname, self.port))
                 if context:
                     ssl_s = context.wrap_socket(s, server_hostname=self.hostname)
                 elif HAS_URLLIB3_SNI_SUPPORT:


bcoca:
the issue you point at in the title seems to have been closed a long time

ei-grad:
If you look down throught the comments you will find that there are another (unreported) issues messing with that one. Also, see #17645.

ei-grad:
But you are right, I should probably remove its mention from the commit message.

lib/ansible/module_utils/urls.py
```diff
@@ -686,7 +685,7 @@ def http_request(self, req):
                 else:
                     raise ProxyError('Unsupported proxy scheme: %s. Currently ansible only supports HTTP proxies.' % proxy_parts.get('scheme'))
             else:
-                s.connect((self.hostname, self.port))
+                s = socket.create_connection((self.hostname, self.port))
```
bcoca: this breaks 2.4 compatibility as it was added in 2.6

bcoca (changes requested):

lib/ansible/module_utils/urls.py
ei-grad: @bcoca is there a common practice to backport such things?
Should I just copy it to `ansible.module_utils.urls`?
https://github.com/python/cpython/blob/master/Lib/socket.py#L688

ei-grad: Should I just copy it to `ansible.module_utils.urls`?

ei-grad: https://github.com/python/cpython/blob/master/Lib/socket.py#L688

abadger: @jimi-c made the following suggestion:

Since we're going to be stopping support of python-2.4 in new versions of Ansible in April of next year, perhaps it makes sense to use create_connection() if  it's available, otherwise using socket.connect().  Would also need a note that the code won't work with ipv6 on python-2.4.

ei-grad: Fixed in 4e19451878779023bba0679baa35baba1d9e96d7

lib/ansible/module_utils/urls.py
```diff
@@ -302,6 +302,16 @@ def match_hostname(cert, hostname):
...
+if hasattr(socket, 'create_connection'):
+    create_connection = socket.create_connection
+else:
+    # Just a graceful fallback for py < 2.6, implementing old behaviour.
+    # Wouldn't work with IPv6, but it is unlikely that python 2.4 could be met
+    # on IPv6-only machine.
+    def create_connection(address):
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        sock.connect(address)
+        return sock
```
sivel, line 316: I'm not really a fan of this approach, and generally speaking, we have tried to stay away from it.

I am more in favor of just doing:

```
try:
    sock = socket.create_connection(address)
except AttributeError:
    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    sock.connect(address)
```

ei-grad: And I'm definitely is not fan of abusing try/except for non-exceptional situations. Your code would raise and handle exception every time on python 2.4. This is not something that could be called "python 2.4" support...

abadger: I don't have a huge preference here but I think @sivel's code is slightly better. We do use try: except (Ask forgiveness) quite a lot for compatibility between python versions.  Asking forgiveness seems to be the accepted idiomatice way to write python versus look before you leap.

Since this is for python-2.4 only it makes even more sense.  In the large picture there's only a few users running python-2.4 and we're going to be getting rid of support for it in less than a year.

Probably the major problem with the approach currently in the code is that it creates an ambiguity in what happens when create_connection() is called that isn't apparent when you are looking at the caller.  This can be worked around by adding the new function under a different name and then putting the conditional into the calling code to select which of the two functions to call or putting the condition and the implementing code into the caller (The latter repeats itself once but for this specific case is probably okay -- it adds less lines of code which often indicates that the function can be considered too trivial to warrant being moved outside the caller).  That will be slower for the common case of operating on non-python-2.4 (exceptions are faster than conditionals when following the happy path) but may be a good compromise.  @sivel is the exception vs conditional the major thing you're looking at or the obfuscation of create_connection's meaning and the relative burden between caller and called?

sivel: Sorry for the slow response, I am concerned both with pythonic implementations, speed and obfuscation combined.

Creating a function named the same as the original method, by which it is hard to tell in the calling location which is happening will often lead to confusion.

hasattr can also sometimes work in strange ways such as what is described in https://hynek.me/articles/hasattr/

Additionally, with hasattr you are performing the lookup multiple times, instead of just trying to use it and handle the exception.  In the majority of cases, just trying to use it and handling the exception will be the faster way.

ei-grad: I don't create a function named the same as original method, I just use it if it is available, and define a fallback if it is not available (in which case there is no original method). For python>=2.6 this function would be the original method. There is no confusion.

The problem with `hasattr` is not a problem since we use it not on a custom class instance but on a module.

Also, it doesn't perform lookup multiple times, since it is called on module import.

ei-grad: Handling exception each time on python2.4 definitely would not be the faster way. The try/except block also has its own little overhead, so it would be slower even on python>=2.6.

ei-grad:
@bcoca @sivel I rebased this on current devel, and replaced `hasattr` by try/except block in import scope. Does it look better for you now?

lib/ansible/module_utils/urls.py
sivel: I'm going to leave this in needs_revision.  My requirements for considering this change have been outlined above.

abadger:
Sidestepping the whole question of how to make compatibility code... We've just branched stable-2.3 which means that devel is now open for developing Ansible-2.4.  Ansible-2.4 drops support for python-2.4 and python-2.5 so we can go back to the code that simply assumes that socket.create_connection is available.

mattclay:
https://github.com/ansible/ansible/pull/17999 is a simplified fix expected to be in 2.4.2, which does not require Python 2.4 support.

Issue #15679, opened 2016-04-30
get_url: SSL cert verification failed for redirect


zeldigas:
##### ISSUE TYPE

- Bug Report
##### ANSIBLE VERSION

```
ansible 2.0.1.0
  config file = /cygdrive/c/Users/me/.ansible.cfg
  configured module search path = Default w/o overrides
```
##### OS / ENVIRONMENT

Running ansible from windows. Ansible is installed via `pip install` command in cygwin environment. Managed machine is ubuntu 14.04 in aws.
##### SUMMARY

File download failed if site redirects to cdn on another domain.
##### STEPS TO REPRODUCE

Having a instance with running ubuntu 14.04 execute

```
ansible ubuntu_box -m get_url -a "url=https://packages.chef.io/stable/ubuntu/12.04/chefdk_0.13.21-1_amd64.deb dest=/tmp/my.deb"
```
##### EXPECTED RESULTS

File downloads just fine
##### ACTUAL RESULTS

```
ubuntu_box | FAILED! => {
    "changed": false,
    "failed": true,
    "invocation": {
        "module_args": {
            "backup": null,
            "checksum": "",
            "content": null,
            "delimiter": null,
            "dest": "/tmp/my.deb",
            "directory_mode": null,
            "follow": false,
            "force": false,
            "force_basic_auth": false,
            "group": null,
            "headers": null,
            "http_agent": "ansible-httpget",
            "mode": null,
            "owner": null,
            "regexp": null,
            "remote_src": null,
            "selevel": null,
            "serole": null,
            "setype": null,
            "seuser": null,
            "sha256sum": "",
            "src": null,
            "timeout": 10,
            "url": "https://packages.chef.io/stable/ubuntu/12.04/chefdk_0.13.21-1_amd64.deb",
            "url_password": null,
            "url_username": null,
            "use_proxy": true,
            "validate_certs": true
        },
        "module_name": "get_url"
    },
    "msg": "SSL Certificate does not belong to packages.chef.io.  Make sure the url has a certificate that belongs to it or use validate_certs=False (insecure)"
}

```

sivel:
Have you tested in 2.0.2?  I believe this may have been resolved already.

Deepakkothandan:
Can confirm the same issue on ansible version 2.0.2.0, running from arch linux and ubuntu 14.04 as target

sivel:
Your initial report indicated that this happened for redirects.  It does not appear that this URL redirects in any manner.

[...]

Issue #12161, opened 2015-08-30
Getting SSL issue again


tolmasky:
msg: Failed to validate the SSL certificate for get.docker.com:443. Use validate_certs=False or make sure your managed systems have a valid CA certificate installed. Paths checked for this platform: /etc/ssl/certs, /etc/pki/ca-trust/extracted/pem, /etc/pki/tls/certs, /usr/share/ca-certificates/cacert.org, /etc/ansible

This is with ansible version 1.9.2. I read on another bug that this is fixed in development branch. Does this mean my only options are to do this insecurely (=False) or run the alpha 2.0 branch?

abadger:
I don not believe that any fixes for that issue have made it into devel that have not made it into 1.9.2.  Could you link the bug and I can verify for you?

Also, what is the playbook that you are using and the version of python on the box that is actually running the task that connects to get.docker.com?

tolmasky:
This is what I have:

 # Config Docker
- name: Add Docker repository key
  sudo: yes
  apt_key: url="https://get.docker.com/gpg" id="A88D21E9"

(We used to have docker.io, but that 301's and also breaks for that reason). I'll report back the python version in a bit.

jbaramidze:
same issue here

abadger:
@tolmasky, I did a little bit of looking into this.  Looks like docker is using SNI to serve get.docker.com.  Unfortunately, SNI isn't supported by python until python-2.7.9.  If the host doing the downloading is using python-2.7.9 or greater then things should be working.  If that host is using python less than 2.7.9 then you unfortunately cannot use get_url securely as the python module that implements the url fetching and certificate verification did not gain the ability to do SNI until 2.7.9. Two workarounds exist: You can tell get_url to download this without verifying the certificate belongs to the server using validate_certs=False. Since this means the connection is subject to man-in-the-middle attacks I would not recommend this as, with that URL, it's likely you're hoping to establish a chain of trust.  If the SSL certificate can't be verified then the chain of trust is broken.  Instead, you can use curl or wget with the command module. They have had support for SNI for longer than the python standard library functions.

[...]

Issue #9966, opened 2015-01-09
apt_repository: Failed to validate the SSL certificate for launchpad.net:443


maedox:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.8.2
  configured module search path = /usr/share/ansible
##### Environment:

Running from: Linux Mint 17.1 (based on Ubuntu 14.04)
Managing: Ubuntu 10.04, 12.04
##### Summary:

Adding a PPA with the apt_repository module fails with certificate validation problems.
Manually running add-apt-repository on the host works.
##### Steps To Reproduce:

``` yaml
- name: "PHP: Add PHP 5.4 PPA"
  apt_repository:
    repo: ppa:ondrej/php5-oldstable
```
##### Expected Results:

PPA added under /etc/apt/sources.list.d/
##### Actual Results:

```
TASK: [php | PHP: Add PHP 5.4 PPA] ******************************************** 
failed: [host.example.com] => {"failed": true}
msg: Failed to validate the SSL certificate for launchpad.net:443. Use validate_certs=no or make sure your managed systems have a valid CA certificate installed. Paths checked for this platform: /etc/ssl/certs, /etc/pki/ca-trust/extracted/pem, /etc/pki/tls/certs, /usr/share/ca-certificates/cacert.org, /etc/ansible
```

jimi-c:
Hi, this situation is no longer present in the devel branch (which will be ansible 2.0). Here is the output using a version of your example above:

<pre>
root@ubuntu1404:~# ansible-playbook -vv 9966.yml -i localhost, -c local
1 plays in 9966.yml
PLAY: ***************************************************************************
TASK [PHP: Add PHP 5.4 PPA] *****************************************************
changed: [localhost] => {"repo": "ppa:ondrej/php5-oldstable", "invocation": {"module_name": "apt_repository", "module_args": {"repo": "ppa:ondrej/php5-oldstable"}}, "state": "present", "changed": true}
PLAY RECAP **********************************************************************
localhost                  : ok=1    changed=1    unreachable=0    failed=0   
</pre>

If you continue seeing any problems related to this issue, or if you have any further questions, please let us know by stopping by one of the two mailing lists, as appropriate:
- https://groups.google.com/forum/#!forum/ansible-project - for user questions, tips, and tricks
- https://groups.google.com/forum/#!forum/ansible-devel - for strategy, future planning, and questions about writing code

[...]

Issue #8221, opened 2014-07-21
apt_repository fails to validate SSL cert for launchpad.net


sw00:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.6.6
##### Environment:

Ubuntu 14.04 on hosts to be provisioned.
OSX with boot2docker-vm on master host.
##### Summary:

apt_repository fails to add repo, fails to validate SSL cert for launchpad.net
##### Steps To Reproduce:

Run this command to use apt_repository to add any launchpad repo.

`ansible all -i devops/docker.inventory -m apt_repository -a "repo=ppa:juju/pkgs" -vvvv`
##### Expected Results:

apt_repository should add the repository.
##### Actual Results:

```
$ ansible all -i devops/docker.inventory -m apt_repository -a "repo=ppa:chris-lea/node.js" -vvvv
<192.168.59.103> ESTABLISH CONNECTION FOR USER: root
<192.168.59.103> REMOTE_MODULE apt_repository repo=ppa:chris-lea/node.js
<192.168.59.103> EXEC ['ssh', '-C', '-tt', '-vvv', '-o', 'ControlMaster=auto', '-o', 'ControlPersist=60s', '-o', 'ControlPath=/Users/sett/.ansible/cp/ansible-ssh-%h-%p-%r', '-o', 'StrictHostKeyChecking=no', '-o', 'Port=50061', '-o', 'IdentityFile=devops/docker/common/docker_rsa', '-o', 'KbdInteractiveAuthentication=no', '-o', 'PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey', '-o', 'PasswordAuthentication=no', '-o', 'User=root', '-o', 'ConnectTimeout=10', '192.168.59.103', "/bin/sh -c 'mkdir -p $HOME/.ansible/tmp/ansible-tmp-1405949653.07-188801175664130 && chmod a+rx $HOME/.ansible/tmp/ansible-tmp-1405949653.07-188801175664130 && echo $HOME/.ansible/tmp/ansible-tmp-1405949653.07-188801175664130'"]
<192.168.59.103> PUT /var/folders/y7/zy_gg2bx2_d5lx_wyl5wm7lc0000gn/T/tmpLCs86T TO /root/.ansible/tmp/ansible-tmp-1405949653.07-188801175664130/apt_repository
[...]

PR #17645, opened 2016-09-19
Fix misleading SSL error message

ei-grad:
The `except` block with exception matching throught
`if 'connection refused' in str(e).lower():` is funny,
but is not user-friendly.

Probably related issues:
- #15679
- #12161
- #9966
- #8221
- #7218

... and more

### DIFF
--- a/lib/ansible/module_utils/urls.py
+++ b/lib/ansible/module_utils/urls.py
@@ -721,15 +721,11 @@ class SSLValidationHandler(urllib_request.BaseHandler):
             # close the ssl connection
             #ssl_s.unwrap()
             s.close()
-        except (ssl.SSLError, socket.error):
-            e = get_exception()
-            # fail if we tried all of the certs but none worked
-            if 'connection refused' in str(e).lower():
-                raise ConnectionError('Failed to connect to %s:%s.' % (self.hostname, self.port))
-            else:
-                build_ssl_validation_error(self.hostname, self.port, paths_checked)
-        except CertificateError:
+        except (ssl.SSLError, CertificateError):
             build_ssl_validation_error(self.hostname, self.port, paths_checked)
+        except socket.error:
+            e = get_exception()
+            raise ConnectionError('Failed to connect to %s at port %s: %s' % (self.hostname, self.port, str(e)))
 
         try:
             # cleanup the temp file created, don't worry


ei-grad:
#17644 is also related)

ei-grad:
@bcoca updated

ei-grad:
@abadger updated

lib/ansible/module_utils/urls.py
```diff
@@ -697,15 +697,11 @@ def http_request(self, req):
...
-            if 'connection refused' in str(e).lower():
-                raise ConnectionError('Failed to connect to %s:%s.' % (self.hostname, self.port))
-            else:
-                build_ssl_validation_error(self.hostname, self.port, paths_checked)
-        except CertificateError:
+        except (ssl.SSLError, CertificateError):
             build_ssl_validation_error(self.hostname, self.port, paths_checked)
+        except socket.error:
+            e = get_exception()
+            raise ConnectionError('Failed to connect to %s at port %s: %s' % (self.hostname, self.port, e))
```
bcoca: use `str(e)` (this is not the optimal, but that is true for most of this file)

cc @abadger

bcoca (changes requested):

lib/ansible/module_utils/urls.py
ei-grad: @bcoca using `%s` in format string uses str() internally, why is it needed to call it twice?..

bcoca: I cannot find the original reason, now it is mostly so I can do a regex to subsitute all of them for `to_native` which avoids unicode errors, or you can just use that and i'll skip this instance.

ei-grad: Imho, there is no need in such hack here, the only case I can see - it would fail if someone use non-english locale, but then this place would probably be last thing to worry about :-). Maybe we shall keep it without str() and to_native here?

ei-grad: @bcoca @abadger friendly ping

abadger: yeah, to_native is the way to do this correctly.  Otherwise sooner or later we will end up with UnicodeErrors being thrown.  Gotta start somewhere so might as well start with new code.

ei-grad:
@bcoca @abadger I rebased this on current devel, and added `str()` as you asked.

abadger:
Thanks for your patience @ei-grad.  Merged.

PR #17669, opened 2016-09-20
removed 'bare var handling' from conditinals

### LINKED ISSUE #17650: Strange behaviour of when: conditionals with string values
##### ISSUE TYPE
- Bug Report
##### COMPONENT NAME

task conditional handling (`when: ...`)
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel 1d2e1a5648) last updated 2016/09/19 20:42:37 (GMT +100)
  lib/ansible/modules/core: (detached HEAD e3b6f52bf5) last updated 2016/09/19 20:46:40 (GMT +100)
  lib/ansible/modules/extras: (detached HEAD 0b778350ec) last updated 2016/09/19 20:46:40 (GMT +100)
  config file =
```

Same behaviour also observed with 2.1.0.0
##### CONFIGURATION

No special configuration
##### OS / ENVIRONMENT

N/A (tested with both OSX 10.11.6 and Ubuntu 16.04)
##### SUMMARY

I expected a conditional like `when: myvar` to be considered true if the variable is set to any non-false value (including a non-empty string)

However it appears to be doing a _second_ level of expansion. For example if myvar expands to the string "foo" then ansible tries to expand the value of expression "foo", which fails if there is no variable called "foo" in the runtime environment. If myvar is set to the string "foo.bar" then it tries to expand foo; presumably it would then apply the "bar" method to the value if one was found.

Note: `when: not myvar` works as expected: the action is taken if myvar is empty string, and not taken if myvar is a non-empty string; there is apparently no second level of expansion.

So a workaround is to write `when: not not myvar`
##### STEPS TO REPRODUCE

Run the following playbook. You can use either variant of the "when:" statements.

```
- hosts: localhost
  gather_facts: false
  vars:
    myserv: www.example.com
    #myserv: ''
  tasks:
    - debug: msg="wibble"
      when: not myserv
      #when: '{{ not myserv }}'

    - debug: msg="bibble"
      when: myserv
      #when: '{{ myserv }}'
```

It's the same whether you use the braces or not.

##### EXPECTED RESULTS

1. If myserv is any 'false-like' value (e.g. null, False, empty string) then the first debug runs
2. If myserv is any 'true-like' value (e.g. non-empty string) then the second debug runs

The [documentation](http://docs.ansible.com/ansible/playbooks_conditionals.html#the-when-statement) gives these examples:

```
tasks:
    - shell: echo "This certainly is epic!"
      when: epic

tasks:
    - shell: echo "This certainly isn't epic!"
      when: not epic
```

which seems consistent with what I'm trying (except that in the documentation, 'epic' is set only to the values True and False, rather than other values like an empty string or a non-empty string)
##### ACTUAL RESULTS

When myserv is empty string, it works fine. The first debug runs and the second does not.

When myserv is the string "foobar" then you get this error from the second debug:

> error while evaluating conditional ({{ myserv }}): 'foobar' is undefined

When myserv is the string "www.example.com" then you get the error: `'www' is undefined`

> error while evaluating conditional ({{ myserv }}): 'www' is undefined

```

skipping: [localhost]

fatal: [localhost]: FAILED! => {"failed": true, "msg": "The conditional check '{{ myserv }}' failed. The error was: error while evaluating conditional ({{ myserv }}): 'www' is undefined\n\nThe error appears to have been in '/Users/brian/git/ansible/ert.yml': line 11, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n    - debug: msg=\"bibble\"\n      ^ here\n"}
    to retry, use: --limit @/Users/brian/git/ansible/ert.retry

```

bcoca:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

when
##### ANSIBLE VERSION

```
2.2
```
##### SUMMARY

fixes #17650

### DIFF
--- a/lib/ansible/playbook/conditional.py
+++ b/lib/ansible/playbook/conditional.py
@@ -88,9 +88,6 @@ class Conditional:
         if conditional is None or conditional == '':
             return True
 
-        if conditional in all_vars and '-' not in text_type(all_vars[conditional]):
-            conditional = all_vars[conditional]
-
         # make sure the templar is using the variables specified with this method
         templar.set_available_variables(variables=all_vars)
 
--- a/test/integration/group_vars/all
+++ b/test/integration/group_vars/all
@@ -13,5 +13,5 @@ test_hash:
 # variables used for conditional testing
 test_bare: true
 test_bare_var: 123
-test_bare_nested_good: "{{test_bare_var}} == 123"
-test_bare_nested_bad: "{{test_bare_var}} == 321"
+test_bare_nested_good: "test_bare_var == 123"
+test_bare_nested_bad: "test_bare_var == 321"
--- a/test/integration/roles/test_conditionals/tasks/main.yml
+++ b/test/integration/roles/test_conditionals/tasks/main.yml
@@ -184,7 +184,7 @@
 
 - name: test good conditional based on nested variables
   shell: echo 'testing'
-  when: test_bare_nested_good
+  when: "{{test_bare_nested_good}}"
   register: result
 
 - name: assert good conditional based on nested var ran
@@ -196,13 +196,13 @@
 
 - name: test bad conditional based on nested variables
   shell: echo 'testing'
-  when: test_bare_nested_bad
+  when: "{{test_bare_nested_bad}}"
   register: result
 
 - name: assert that the bad nested conditional did NOT run
   assert:
     that:
-    - "result.skipped == true"
+    - result|skipped
 
 #-----------------------------------------------------------------------
 # proper booleanification tests (issue #8629)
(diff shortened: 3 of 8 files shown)


mattclay:
Is this supposed to remove support for nested conditionals?

The [nested conditional integration test](https://github.com/ansible/ansible/blob/devel/test/integration/roles/test_conditionals/tasks/main.yml#L202) is now failing:

https://app.shippable.com/runs/57e3077dd95a2b10004d4ba7/23/tests

bcoca:
that test looks all kinds of wrong to me

mattclay:
I don't know if they're correct or not, @jimi-c added them: https://github.com/ansible/ansible/commit/3baf2dc0956d84daa7e2c6475b511ae7cab67ff4

PR #17676, opened 2016-09-20
Add more options to magic vars

sivel:
##### ISSUE TYPE

Feature Pull Request
##### COMPONENT NAME

N/A
##### ANSIBLE VERSION

```
devel
```
##### SUMMARY

This PR adds the CLI options of `tags`, `subset` and `skip_tags` to magic vars.

The mapping is currently:
- `options.tags` -> `ansible_tags`
- `options.skip_tags` -> `ansible_skip_tags`
- `options.subset` -> `ansible_limit`

My concern with this is naming confusion.  I think `skip_tags` is fine.  I am concerned that `ansible_tags` could be confused with the tags you have placed on a task, rather than supplied on the CLI.  This goes for `ansible_limit` as well, since it could be confused with `hosts:`.

Open to suggestions.

Also, to support `ansible_limit` for adhoc, I am combining `self.args` and `self.options.subset` for adhoc.  This is effectively what adhoc does anyway.

### DIFF
--- a/lib/ansible/cli/adhoc.py
+++ b/lib/ansible/cli/adhoc.py
@@ -79,6 +79,10 @@ class AdHocCLI(CLI):
         elif len(self.args) > 1:
             raise AnsibleOptionsError("Extranous options or arguments")
 
+        if self.options.subset:
+            self.options.subset += ',%s' % self.args
+        else:
+            self.options.subset = self.args
         display.verbosity = self.options.verbosity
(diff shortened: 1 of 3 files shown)


sivel:
`main` may not always have a `cli`, such as the example where someone is using the python API in their own app.  Need to add a `try/except` when importing.

calfonso:
@sivel are you still interested in getting this merged?

needs_info

bcoca:
implemented as part of #30379

PR #17677, opened 2016-09-20
update intro to networking section

privateip:
removes out of date device list and adds note about supported transports


dharmabumstead:
Thanks Peter!

PR #17718, opened 2016-09-23
Add tests for the mount module

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

integration tests for mount module
##### ANSIBLE VERSION

```
devel, 2.1
```
##### SUMMARY

@jtyr contributed a fix for bind mounts.  This integration test makes sure that the module now behaves as expected.  This also needs to have a working bind unmount.  So the module PR that needs to be applied for this to work is: https://github.com/ansible/ansible-modules-core/pull/4980

### DIFF
--- a/test/integration/roles/test_mount/tasks/main.yml
+++ b/test/integration/roles/test_mount/tasks/main.yml
@@ -1,5 +1,4 @@
-# test code for the ping module
-# (c) 2014, James Cammarata <jcammarata@ansible.com>
+# (c) 2016, Toshio Kuratomi <tkuratomi@ansible.com>
 
 # This file is part of Ansible
 #
--- a/test/integration/roles/test_mount/tasks/main.yml
+++ b/test/integration/roles/test_mount/tasks/main.yml
@@ -96,7 +96,7 @@
 - name: Unmount the bind mount
   mount:
     name: "{{ outputdir }}/mount_dest"
-    state: "unmounted"
+    state: "absent"
   when: ansible_system in ('Linux', 'FreeBSD')
   register: unmount_result
 
--- a/test/integration/roles/test_mount/tasks/main.yml
+++ b/test/integration/roles/test_mount/tasks/main.yml
@@ -1,0 +1,114 @@
+# test code for the ping module
+# (c) 2014, James Cammarata <jcammarata@ansible.com>
+
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+- name: Create the mount point
+  file:
+    state: "directory"
+    path: "{{ outputdir }}/mount_dest"
+
+- name: Create a directory to bind mount
+  file:
+    state: "directory"
+    path: "{{ outputdir }}/mount_source"
+
+- name: Put something in the directory so we see that it worked
+  copy:
+    content: "Testing\n"
+    dest: "{{ outputdir }}/mount_source/test_file"
+  register: orig_info
+
+# The opts type of bind mount only works on Linux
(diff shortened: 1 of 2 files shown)


test/integration/roles/test_mount/tasks/main.yml
```diff
@@ -0,0 +1,114 @@
+# test code for the ping module
```
mattclay: Not a test for the ping module... :)

alikins:
Tested this with the old module and it fails as expected. Tested with new module and it passes as expected.

But it did modify /etc/fstab and didn't clean it up afterwards.
('TEST_FLAGS='-vvv --tags test_mount' make non_destructive' output at https://gist.github.com/alikins/c63e09a3a3fab8556eae8429ff28bd07)

Before:

```
[root@dhcp129-186 integration]# cat /etc/fstab

#
# /etc/fstab
# Created by anaconda on Thu Mar 20 17:56:50 2014
#
# Accessible filesystems, by reference, are maintained under '/dev/disk'
# See man pages fstab(5), findfs(8), mount(8) and/or blkid(8) for more info
#
/dev/mapper/fedora_dhcp129--186-root /                       ext4    defaults        1 1
UUID=32caaec3-ef40-4691-a3b6-438c3f9bc1c0 /boot                   ext4    defaults        1 2
/dev/mapper/fedora_dhcp129--186-home /home                   ext4    defaults        1 2
/dev/mapper/fedora_dhcp129--186-swap swap                    swap    defaults        0 0
```

After:

```
[root@dhcp129-186 integration]# cat /etc/fstab

#
# /etc/fstab
# Created by anaconda on Thu Mar 20 17:56:50 2014
#
# Accessible filesystems, by reference, are maintained under '/dev/disk'
# See man pages fstab(5), findfs(8), mount(8) and/or blkid(8) for more info
#
/dev/mapper/fedora_dhcp129--186-root /                       ext4    defaults        1 1
UUID=32caaec3-ef40-4691-a3b6-438c3f9bc1c0 /boot                   ext4    defaults        1 2
/dev/mapper/fedora_dhcp129--186-home /home                   ext4    defaults        1 2
/dev/mapper/fedora_dhcp129--186-swap swap                    swap    defaults        0 0
/root/ansible_testing/mount_source /root/ansible_testing/mount_dest None bind 0 0
```

abadger:
Regarding fstab... yes.  Here's my analysis of that:

> Looks like it also has the problem that it always creates an fstab entry.  But my reading of the module docs is that should only be done for state=present and state=absent.  (state=mounted and state=unmounted should just mount/unmount -- not persist)
> I have not fixed that.
> Fixing it would take two steps: (1) not creating the fstab entry on mounted/unmounted and (2) making use of the args like opts, etc when calling mount instead of relying on mount using the fstab entry that contains them.
>  Oddly enough, the args are created in the mounted() and unmounted() code path.. they just aren't passed on to the command. (passed on to the mount command)

jtyr:
I agree with @abadger. The `mount` state should only mount but not create record in the fstab. Changing it would break the logic for users as that behaviour was there since long time. The best would be to have the `mounted`, `present` and `mounted_present` states but that still doesn't solve the problem with the backward compatibility as the current `mounted` would be named differently (`mounted_present`).

abadger:
Okay -- reading the documentation, it looks like fstab is always documented to be modified.  Mounting is the part that's optional.  I'll have to re-read the code with that in mind to see if mounting is really optional and probably also move this to the destructive tests.

abadger:
@jtyr yeah -- Two new states is probably the best thing.  And for backwards compat the new states would need to be the ones that only did mount and unmount (without modifying fstab).   Let's do that portion as a separate pull request so that we can get the bind mount changes finished today.

abadger:
Ouch... according to comments inside of the main function, present/unmounted and mounted/absent are the opposing states:

```
    # absent:
    #   Remove from fstab and unmounted.
    # unmounted:
    #   Do not change fstab state, but unmount.
    # present:
    #   Add to fstab, do not change mount state.
    # mounted:
    #   Add to fstab if not there and make sure it is mounted. If it has
    #   changed in fstab then remount it.
```

abadger:
Read the code comments and docs... they all disagree about what each state does:

Comment:
- absent: remove from ftab and unmount
- unmounted: unmount
- present: add to fstab
- mounted: add to fstab and remount
  
  Docs: 
- absent: remove from fstab
- unmounted: remove from fstab and unmount
- present: Add to fstab
- mounted: Add to fstab and unmount

Code:
- absent: Remove from fstab and unmount
- unmounted: unmount
- mounted: Add to fstab and mount
- present: Add to fstab

Also found out that check mode is broken

abadger:
Addressed the comments issue raised by @mattclay and changed from unmounted to absent to clean up fstab.

PR #17731, opened 2016-09-23
Provide functions for IP address or port range validating (fixes #16256)

### LINKED ISSUE #16256: Provide functions for IP address or port range validating
##### ISSUE TYPE
Feature Idea

##### COMPONENT NAME
module_utils

##### ANSIBLE VERSION
```
ansible 2.2.0.0
```

##### OS / ENVIRONMENT
any

##### SUMMARY
ansible.module_utils.basic should provide functions for validating IP addresses/ranges, port ranges etc.

xen0l:
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

module_utils/basic.py
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file = 
```
##### SUMMARY

The change adds new types and validations for ports, L3/L4/L7 protocols, IPv4 and IPv6 addresses. Fixes issue #16256.

```

```

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -163,7 +163,7 @@ from ansible.module_utils.six import (
 )
 from ansible.module_utils.six.moves import map, reduce, shlex_quote
 from ansible.module_utils._text import to_native, to_bytes, to_text
-from ansible.module_utils.parsing.convert_bool import BOOLEANS, BOOLEANS_FALSE, BOOLEANS_TRUE, boolean
+from ansible.module_utils.parsing.convert_bool import BOOLEANS, BOOLEANS_FALSE, BOOLEANS_TRUE, boolean as parse_boolean
 from ansible.module_utils.validators.core import (
     check_type_str,
     check_type_list,
@@ -817,6 +817,17 @@ def safe_eval(value, locals=None, include_exceptions=False):
         if include_exceptions:
             return (value, e)
         return value
+
+
+def boolean(arg):
+        ''' return a bool for the arg '''
+        if arg is None:
+            return arg
+
+        try:
+            return parse_boolean(arg)
+        except TypeError:
+            raise TypeError('%s cannot be converted to a bool' % type(arg))
 
 
 class AnsibleFallbackNotFound(Exception):
@@ -2170,10 +2181,6 @@ class AnsibleModule(object):
         return bin_path
 
     def boolean(self, arg):
-        ''' return a bool for the arg '''
-        if arg is None:
-            return arg
-
         try:
             return boolean(arg)
         except TypeError as e:
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -164,7 +164,19 @@ from ansible.module_utils.six import (
 from ansible.module_utils.six.moves import map, reduce, shlex_quote
 from ansible.module_utils._text import to_native, to_bytes, to_text
 from ansible.module_utils.parsing.convert_bool import BOOLEANS, BOOLEANS_FALSE, BOOLEANS_TRUE, boolean
-from ansible.module_utils.validators.core import check_type_str
+from ansible.module_utils.validators.core import (
+    check_type_str,
+    check_type_list,
+    check_type_dict,
+    check_type_bool,
+    check_type_int,
+    check_type_float,
+    check_type_path,
+    check_type_jsonarg,
+    check_type_raw,
+    check_type_bytes,
+    check_type_bits,
+)
 
 # Note: When getting Sequence from collections, it matches with strings.  If
 # this matters, make sure to check for strings before checking for sequencetype
@@ -778,6 +790,35 @@ def jsonify(data, **kwargs):
     raise UnicodeError('Invalid unicode encoding encountered')
 
 
+def safe_eval(value, locals=None, include_exceptions=False):
+
+    # do not allow method calls to modules
+    if not isinstance(value, string_types):
+        # already templated to a datavaluestructure, perhaps?
+        if include_exceptions:
+            return (value, None)
+        return value
+    if re.search(r'\w\.\w+\(', value):
+        if include_exceptions:
+            return (value, None)
+        return value
+    # do not allow imports
+    if re.search(r'import \w+', value):
+        if include_exceptions:
+            return (value, None)
+        return value
+    try:
+        result = literal_eval(value)
+        if include_exceptions:
+            return (result, None)
+        else:
+            return result
+    except Exception as e:
+        if include_exceptions:
+            return (value, e)
+        return value
+
+
 class AnsibleFallbackNotFound(Exception):
     pass
 
@@ -865,17 +906,17 @@ class AnsibleModule(object):
 
         self._CHECK_ARGUMENT_TYPES_DISPATCHER = {
             'str': check_type_str,
-            'list': self._check_type_list,
-            'dict': self._check_type_dict,
-            'bool': self._check_type_bool,
-            'int': self._check_type_int,
-            'float': self._check_type_float,
-            'path': self._check_type_path,
-            'raw': self._check_type_raw,
-            'jsonarg': self._check_type_jsonarg,
-            'json': self._check_type_jsonarg,
-            'bytes': self._check_type_bytes,
-            'bits': self._check_type_bits,
+            'list': check_type_list,
+            'dict': check_type_dict,
+            'bool': check_type_bool,
+            'int': check_type_int,
+            'float': check_type_float,
+            'path': check_type_path,
+            'raw': check_type_raw,
+            'jsonarg': check_type_jsonarg,
+            'json': check_type_jsonarg,
+            'bytes': check_type_bytes,
+            'bits': check_type_bits,
         }
         if not bypass_checks:
             self._check_required_arguments()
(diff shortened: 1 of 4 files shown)


xen0l:
Anyone?

abadger:
I hate to add this to basic.py since only a few things will need it but every module needs basic.py.  At some point (sooner for some things like parameter parsing since network modules want to have access to those routines from ansible/ansible code) we want to pull basic.py apart into smaller pieces.  Is there some way we could make parameter parsing more modular when we do this and then make this one optional piece of that?

xen0l:
Perhaps, we could create a file, which would contain additional module parsing types on it's own. This file would be then imported in network or regular modules? I am thinking about this:

parameter_parsing_additional.py -> network_basic.py -> network_module

The first file could be also imported into regular modules.

abadger:
Yeah, something like that sounds good.  Maybe using a subdirectory to hold all parameter related modules.

What I'm still thinking about is how we tell the argument_spec handling code about additional types.

Maybe something like:

``` python
# classes that implement new parameter validation types
from ansible.module_utils.parameters.ipaddress import IPValidator
from ansible.module_utils.parameters.basic import BasicValidator

AnsibleModule(parameter_validators=(BasicValidators, IPValidators), [...])
```

The Validators are really just mappings of type names to functions.  So maybe it should look like:

``` python
# dictionaries mapping string type names to parameter validator/conversion functions
from ansible.module_utils.parameters.ipaddress import IPVALIDATORS
from ansible.module_utils.parameters.basic import BASICVALIDATORS

AnsibleModule(parameter_validators=(BASICVALIDATORS, IPVALIDATORS), [...])
```

AnsibleModule parameter_validators can default to  [BASICVALIDATORS].

Does this start to sound like a good plan?

xen0l:
Yeah, this seems reasonable. However, I think that we should do refactor in another PR.

abadger:
@xen0l Once it's added to basic.py it has to stay in basic.py for backwards compatibility.  (look at the imports near the top of basic.py that simply bring old names into the basic namespace).  This means that the size of downloads for modules which utilize basic.py just continues to grow and grow.

Since this is new functionality that is infrequently needed I think it is okay to gate acceptance of the feature on making it optional.  We can bring it up with other core committers for a second opinion, though.

nitzmahone:
I like where @abadger is going- a slightly simpler approach might be to extend the `type` value in the arg dicts to accept a validator type and/or function arg in addition to the string value it accepts today. Still meets the goal of not needing to predeclare what types we're using in the AnsibleModule ctor to extend the mapping while still allowing it be infinitely extended without further changes to basic.py.

eg:

```
from ansible.module_utils.parameters.ipaddress import subnet_mask_validator

...

module = AnsibleModule(arg_spec=dict(
  subnet=dict(type=subnet_mask_validator)
))
```

privateip:
@xen0l @abadger we have a network_common.py in module_utils which might be a good place to add this functionality.

gundalow:
I like this, added `network` and milestone 2.4 so we will track it

xen0l:
I will revive it and move to network_common.py.

calfonso:
@xen0l just checking in on this, are you still reviving this PR?

xen0l:
@calfonso sorry for late response, I had a brief look, but was busy during this time. I will have some time over the weekend to work on this.

xen0l:
@privateip if I add these to network_common.py, then I won't be able to include them in _CHECK_ARGUMENT_TYPES_DISPATCHER. From what I understood, shall I just migrate the basic function and then import then in basic.py?

abadger:
@xen0l Going along with what @nitzmahone has suggested we think that it should look more like this:

basic.py only needs a minimal change around here: https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L1743 .  probably this code will do:
``` python
            if callable(wanted):
                type_checker = wanted
            else:
                try:
                    type_checker = self._CHECK_ARGUMENT_TYPES_DISPATCHER[wanted]
                except KeyError:
                    self.fail_json(msg="implementation error: unknown type %s requested for %s" % (wanted, k))
```

Once that's in place, the new functions can go into new python file.  Then the module is responsible for putting these two together. The module code would look something like this:
``` python
from ansible.module_utils.validators.network import check_type_port
[...]
AnsibleModule(argument_spec={'foo': {'type': check_type_port}},[..])
```

If you want to keep implementation detail of the new functions making use of the existing validators, we could move them to the validators as well.  So something like this:

```
module_utils
├── validators
│   ├── core.py        ==> Contains the current validators
│   ├── __init__.py   ==> An empty file to make this a python module
│   └── network.py   ==> Contains the ip and port validators
```

AnsibleModule.safe_eval will probably need to move out to the toplevel of basic.py as well so that _check_type_dict() can make use of it.  (The safe_eval method will need to become a backwards compat shim that calls the toplevel function then).

abadger:
Note that it is okay for the validators to import basic.py  but it isn't okay for basic.py to import any validators except for core.  That way we aren't increasing the things that have to be shipped across the wire.

calfonso:
@xen0l just a quick re-query on this, are you up for adding to this PR so we can get it reviewed/merged?

gundalow:
@xen0l Rebase needed

gundalow:
@xen0l Hi, wondering if you had time to work on this? I think this would be really useful for many modules, and not just the Networking ones.

gundalow:
Along with rebase this needs moving to `network_common.py`

xen0l:
@gundalow sorry for late response, but I am more less in summer mode, so I am not always reacting so fast. Yes, I plan to pick this up, but not sooner than this weekend. Regarding your comment of moving this to network_common.py, what approach shall I follow then? Move these functions only or also work on validator thing?

abadger:
@xen0l Yeah, the validator thing is needed in order to make the split up approach work.  Just moving the functions there won't help.

gundalow:
Hi,
This hasn't made the cut for 2.4.0

For 2.5 we can make this live and update a few modules to use it, as well as adding tests and documentation.

gundalow:
Now that `stable-2.4` has been branched this can be put into `network_common.py` and once fixed merged into `devel`

abadger:
@xen0l nitzmahone added the framework for this to go in here:  https://github.com/ansible/ansible/pull/27183/  (It's in 2.4.x) so if you want to rework this PR to use that framework it can get merged.

gundalow:
@xen0l You able to look at this?

gundalow:
@xen0l nitzmahone added the framework for this to go in here: #27183 (It's in 2.4.x) so if you want to rework this PR to use that framework it can get merged.

gundalow:
@xen0l Gentle poke :)

gundalow:
For the record I really like https://github.com/ansible/ansible/pull/17731#issuecomment-303870640

xen0l:
@gundalow is #17731 (comment) the way how to do it?

gundalow:
@xen0l Correct

xen0l:
@gundalow @abadger so I updated the PR as requested, but there are few issues, on which I will need your guidance:

1) I had to move `boolean` outside of `AnsibleModule` class (https://github.com/ansible/ansible/pull/17731/files#diff-90085fdcec6ed8b273ba885eaee60328R822) in order to be able to use it inside validators. However, I am not sure if this is the best approach how to tackle the problem. Given the poor naming choice of `boolean` from `ansible.module_utils.parsing.convert_bools` I can't create the function with same name, so I think this was the easiest option. If you happen to have a better idea, please let me know.
2) I moved `safe_eval` outside of `AnsibleModule` as requested
3) The code is currently not working, because in `ansible.module_utils.validators.core` we are trying to import `jsonify` (among other utils) from `ansible.module_utils.basic`, but it fails. I think this is because it results in circular imports. Would be the solution to move `jsonify`, `safe_eval` etc to different file?
4) What is the best path for placing validator tests?

Feedback/reviews are welcome.

xen0l:
@gundalow / @abadger kind reminder.

mkrizek:
bot_skip

mkrizek:
^^^ `LabelWafflingError: "needs_rebase" label is waffling on https://github.com/ansible/ansible/pull/17731`

gundalow:
@xen0l Could you rebase, I'd like to get this in  Ansible 2.8

xen0l:
@gundalow yes, I will rebase. However, there are issues with circular imports, which need to be solved and I might need to get input from Ansible developers to know, which solution is acceptable.

gundalow:
```
Command "ansible-doc -t cache jsonfile memcached memory mongodb pickle redis yaml" returned exit status 250.
>>> Standard Error
/root/ansible/bin/ansible-doc:103: DeprecationWarning: BaseException.message has been deprecated as of Python 2.6
  msg = e.message
ERROR! Unexpected Exception, this is probably a bug: cannot import name jsonify
```

```
Command "ansible-doc -t cache jsonfile memcached memory mongodb pickle redis yaml" returned exit status 250.
>>> Standard Error
ERROR! Unexpected Exception, this is probably a bug: cannot import name 'jsonify' from 'ansible.module_utils.basic' (/root/ansible/lib/ansible/module_utils/basic.py)
```

gundalow:
@xen0l If you rebase we can then take a look at the circular imports

gundalow:
I'd love to see this get into Ansible 2.9, it would be useful for many modules

Issue #17750, opened 2016-09-26
Support 256-color terminals


yurikhan:
##### ISSUE TYPE
- Feature Idea
##### COMPONENT NAME

lib/ansible/utils/color.py
##### ANSIBLE VERSION

```
ansible 2.1.1.0
```
##### OS / ENVIRONMENT

Running on Ubuntu 14.04; OS being managed is not relevant
##### SUMMARY

Ansible 2.1 added configurable colors (see commit d3deb24). However, the set of available colors is limited to 8 standard terminal colors + bold, due to a hardcoded map `codeCodes` in `color.py`.

I propose adding support for `color0` through `color255`, with aliases `rgb000` through `rgb555` for 16–231 and `gray0` through `gray23` for 232–255. This color naming scheme is used in several existing programs and libraries.

I will prepare a pull request if the general idea gets approved.

This could further be extended to allow text attributes (bold, italic, underline…) and backgrounds (e.g. `on gray4`) if desired.
##### STEPS TO REPRODUCE

```
[colors]
verbose = rgb025
debug = gray14
```
##### EXPECTED RESULTS

Verbose output uses color 33 (= 16 + 0 \* 36 + 2 \* 6 + 5) of the 256-color palette; debug uses color 246 (= 16 + 216 + 14).

ansibot:
Files identified in the description:
* [lib/ansible/utils/color.py](https://github.com/ansible/ansible/blob/devel/lib/ansible/utils/color.py)

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

samdoran:
Closing as this was resolved in #17752

PR #17752, opened 2016-09-26
Support 256-color terminals (#17750)

yurikhan:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

lib/ansible/utils/color.py
##### ANSIBLE VERSION

devel
##### SUMMARY

Extend the set of supported colors to include `color0` through `color255`, `rgb000` through `rgb555`, and `gray0` through `gray23`.

This color naming scheme is used by Midnight Commander, Taskwarrior, and the Term::ANSIColor Perl library; the `color0`–`color255` subset is also supported by Emacs.

An alternative naming scheme is provided by the Python [colored](https://pypi.python.org/pypi/colored) package, but I find it less intuitive.

The regular expression could be compiled, if performance is a concern.

### DIFF
--- a/lib/ansible/utils/color.py
+++ b/lib/ansible/utils/color.py
@@ -17,6 +17,7 @@
 from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
+import re
 import sys
 
 from ansible import constants as C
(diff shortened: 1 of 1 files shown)


bcoca:
@yurikhan please rebase, i tried to fix using github tools and it just messed it up more

yurikhan:
Rebased to current devel.

jhawkesworth:
Ooh, now I know this is I thing I want it. Found this link for anyone wanting to try out 256 color terminals: http://www.robmeerman.co.uk/unix/256colours

PR #17753, opened 2016-09-26
Ignore error when setting permissions fail due to EPERM

dbarrosop:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

`AnsibleModule.atomic_move`
##### ANSIBLE VERSION

```
ansible 2.1.0.0
```
##### SUMMARY

Some network equipment uses `vfat` as their filesystem of choice. As vfat doesn't support `chmod` and `chown` operations, the method `atomic_move` will fail when trying to set the permissions.

This PR tries to address that issue. If we get to the point where the file is already in place and we just have to fix permissions, the most likely reason to get an `EPERM` error is due to the fact that the underlying filesystem doesn't support those operations in, which case, is safe to ignore the error.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -1998,11 +1998,18 @@ class AnsibleModule(object):
         if creating:
             # make sure the file has the correct permissions
             # based on the current value of umask
-            umask = os.umask(0)
-            os.umask(umask)
-            os.chmod(b_dest, DEFAULT_PERM & ~umask)
-            if switched_user:
-                os.chown(b_dest, os.getuid(), os.getgid())
+            try:
+                umask = os.umask(0)
+                os.umask(umask)
+                os.chmod(dest, DEFAULT_PERM & ~umask)
+                if switched_user:
+                    os.chown(dest, os.getuid(), os.getgid())
+            except OSError:
+                # Some filesystems like vfat will fail with EPERM if
+                # you try to chmod or chown a file.
+                e = get_exception()
+                if e.errno != errno.EPERM:
+                    raise
 
         if self.selinux_enabled():
             # rename might not preserve context


dbarrosop:
Are all those errors in `shippable` expected? I can't make sense of them and they seem unrelated to the changes in this PR.

bcoca:
This obscures errors that might happen for other reasons, not just `vfat`, this should be an explicit option or just avoid calling atomic_move when dealing with `vfat` devices, as it is irrelevant in those contexts.

bcoca:
FYI, Ansible is designed to deal with POSIX systems, vfat is not even close to being POSIX. The atomic_move function is to ensure consistency for POSIX filesytems, if using vfat, you should avoid calling this function at all.

lib/ansible/module_utils/basic.py
```diff
@@ -1998,11 +1998,18 @@ def atomic_move(self, src, dest, unsafe_writes=False):
...
             # make sure the file has the correct permissions
             # based on the current value of umask
-            umask = os.umask(0)
-            os.umask(umask)
-            os.chmod(b_dest, DEFAULT_PERM & ~umask)
-            if switched_user:
-                os.chown(b_dest, os.getuid(), os.getgid())
+            try:
+                umask = os.umask(0)
+                os.umask(umask)
```
alikins, line 2003: I think the umask lines should be outside/before the try

alikins:
Is there an example of when/where/how this fails? (I suspect in various uses of core/files/*.py...)

I tend to agree with @bcoca about preferring to not intertwingle support for vfat into this method. However, there is some existing code for ignoring errors from non unix file semantics. There is also existing code to handle non-posix selinux support.

 If there is a reasonable way to detect the dest file is on a vfat fs, it may make sense to mimic the selinux support and make the owner/perm checks conditional on the fs type (ie, that block for setting owner/perm on vfat would be a noop...). I don't really like having to have that conditionals in the method, but there is some precedent.

alikins:
https://github.com/ansible/ansible/pull/17753#issuecomment-250526423 is also based on the assumption that we don't want to just fix the places that call atomic_move() but could live without it.

That would also mean code using atomic_move() may need to also (optionally) specific unsafe_writes=True. It would be kind of unfortunate to not be able to enforce atomic moves though...

dbarrosop:
> This obscures errors that might happen for other reasons, not just vfat

If you made it that far into the function, what else it could be? You already copied the file and did everything, you are just brute forcing the permissions.

> Is there an example of when/where/how this fails?

As far as I can tell it affects `template` and `copy` modules. It might be more but those are the two I spotted.

> this should be an explicit option

I am fine if you prefer that, I just thought it would be safe enough to do it directly like that for the same reason I mentioned before, if you made it that far it should be fairly safe to assume the FS doesn't support permissions

>  just avoid calling atomic_move when dealing with vfat devices

What would you use instead.

Regarding the rest you mention (SELINUX, detecting non POSIX, etc.), you got me there, I am afraid I don't know the internals of ansible too much and I don't care that much about a solution or another as long as we fix them. I hate that we have to fix something like this in 2016 but some major networking vendors seem to think `vfat` is a FS worth using in modern equipment ;)

bcoca:
So copy and template are normally not usable on appliances as they require python installed on the targets, I'm assuming you are using this from some network module. Instead of using atomic move just use `shtutil.copy` or `os.rename` directly, it will save a lot of time that atomic_move spends on trying to do things 'right' but it expects filesystems/OSs that support this workflow.

We do have 'some' non POSIX support, mostly because of flawed network/virtual mounts of some filesystems that don't support the full standard but do support most operations. I would draw the line on vfat as it doesn't support any of the things we do in atomic_move.

dbarrosop:
@bcoca I am using network appliances and I have linux access to the machine, root access, python and I am using just regular modules that I could literally point to a linux server and it would work. There are plenty of operating systems for network appliances based on linux that give you full access as if it was a linux server; Arista EOS, Openswitch, Cumulus... to name just a few.

dbarrosop:
And as I mentioned before, I don't have a particular opinion on how to fix this problem, I just did the simplest change because I assumed it was safe. You are brute forcing setting permissions, if we weren't to set them when they were already correct or if that part of the code could be disabled with an argument it would be acceptable as well. But discarding all the network operating systems that behave like just a regular linux server seems like a bad idea to me.

bcoca:
My networkfu is a bit out of date, i'm aware that most were linux/bsd under the hood but aside from Cumulus and Openswitch I as not aware that new appliances were giving access to the 'actual OS' and not just their interface on top.

I need to think on the best approach here, i'm leaning to a blacklisting of filesystem to just bypass the whole thing (both fat and ntfs derivatives lack atomicity in copy so the whole function is mute when those are involved).

abadger:
@dbarrosop how are we getting to this point in the code?  The initial os.rename() has to fail in order to get here but I'm not sure under what conditions that is happening.  Can you tell me what that is?

dbarrosop:
If I move the rename method outside the try..except I get the following:

```
      File "/tmp/ansible_xWxZv_/ansible_module_copy.py", line 348, in <module>
        main()
      File "/tmp/ansible_xWxZv_/ansible_module_copy.py", line 327, in main
        module.atomic_move(src, dest)
      File "/tmp/ansible_xWxZv_/ansible_modlib.zip/ansible/module_utils/basic.py", line 1868, in atomic_move
    OSError: [Errno 18] Invalid cross-device link
```

abadger:
@dbarrosop nevermind, I see that we're doing that whenever we create a file.

abadger:
@dbarrosop So we've been having a few discussions on this to try to figure out a way to fix this.  bcoca made a proposal which seems to be the only workable one:
- Add a new parameter to atomic_move() which allows us the code to ignore errors when setting owner and permissions.
- In every module that calls atomic move, add a new module parameter to say ignoring the error is okay.  Pass that on to atomic_move()
- Users in this situation need to specify the new parameter to modules which call atomic move in their playbooks.

Looking through the code I wonder if the set_*_if_different() methods are also affected by this.  set_owner_if_different, set_group_if_different, set_mode_if_different (and the two helper functions that call those, set_fs_attributes_if_different, and set_directory_attributes_if_different).  Can you test some modules (the file module is probably a  good candidate to test although you likely won't use that module in production.. instead you'd hit it from invoking something like unarchive or copy with certain parameters.

Also, does selinux need the same treatment or does that API already ignore problems if the filesystem doesn't have support for selinux?

abadger:
@dbarrosop  Also -- we're thinking that this is going to affect quite a few modules and it's rather late in the 2.2 release cycle so we're probably going to target the fix for this at 2.3 instead of 2.2.

dbarrosop:
@abadger your comments sound good to me. Let me know how you want to proceed with this (should I contribute to `devel` or to different branch?) and what you want me to help with. I can certainly help with the `atomic_move` method itself and the with the `copy`, `template` and `assemble` modules.

Regarding SELINUX, I have no idea, my experience with it is very limited.

dbarrosop:
Ok, interestingly enough, looks like the `devel` branch includes a patch similar to mine.

```
<<<<<<< dbarrosop/fix-vfat-permissions
            try:
                umask = os.umask(0)
                os.umask(umask)
                os.chmod(dest, DEFAULT_PERM & ~umask)
                if switched_user:
                    os.chown(dest, os.getuid(), os.getgid())
            except OSError:
                # Some filesystems like vfat will fail with EPERM if
                # you try to chmod or chown a file.
                e = get_exception()
                if e.errno != errno.EPERM:
                    raise
=======
            umask = os.umask(0)
            os.umask(umask)
            os.chmod(b_dest, DEFAULT_PERM & ~umask)
            try:
                os.chown(b_dest, os.geteuid(), os.getegid())
            except OSError:
                # We're okay with trying our best here.  If the user is not
                # root (or old Unices) they won't be able to chown.
                pass
>>>>>>> devel
```

It solves my problem so I am closing this PR.

gundalow:
Thanks for confirming.

PR #17760, opened 2016-09-26
swallow errors during async polling

nitzmahone:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

task_executor
##### SUMMARY

Fixes async polling failures due to transient connection issues. Also use connection reset on exceptions if available (eg, prevent reuse of "stuck" WinRM connections due to post-module-run-reboot/NIC bounce/etc)

### DIFF
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -619,14 +619,24 @@ class TaskExecutor:
         while time_left > 0:
             time.sleep(self._task.poll)
 
-            async_result = normal_handler.run(task_vars=task_vars)
-            # We do not bail out of the loop in cases where the failure
-            # is associated with a parsing error. The async_runner can
-            # have issues which result in a half-written/unparseable result
-            # file on disk, which manifests to the user as a timeout happening
-            # before it's time to timeout.
-            if int(async_result.get('finished', 0)) == 1 or ('failed' in async_result and async_result.get('_ansible_parsed', False)) or 'skipped' in async_result:
-                break
+            try:
+                async_result = normal_handler.run(task_vars=task_vars)
+                # We do not bail out of the loop in cases where the failure
+                # is associated with a parsing error. The async_runner can
+                # have issues which result in a half-written/unparseable result
+                # file on disk, which manifests to the user as a timeout happening
+                # before it's time to timeout.
+                if int(async_result.get('finished', 0)) == 1 or ('failed' in async_result and async_result.get('_ansible_parsed', False)) or 'skipped' in async_result:
+                    break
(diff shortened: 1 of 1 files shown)


lib/ansible/executor/task_executor.py
```diff
@@ -619,14 +619,24 @@ def _poll_async_result(self, result, templar, task_vars=None):
...
+                # before it's time to timeout.
+                if int(async_result.get('finished', 0)) == 1 or ('failed' in async_result and async_result.get('_ansible_parsed', False)) or 'skipped' in async_result:
+                    break
+            except Exception as e:
+                # Connections can raise exceptions during polling (eg, network bounce, reboot); these should be non-fatal.
+                # On an exception, call the connection's reset method if it has one (eg, drop/recreate WinRM connection; some reused connections are in a broken state)
+                display.vvvv("Exception during async poll, retrying... (%s)" % to_text(e))
+                display.debug("Async poll exception was:\n%s" % to_text(traceback.format_exc()))
+                try:
+                    normal_handler._connection.reset()
```
alikins: Is winrm the only connection plugin that supports reset()?

And I only see a _reset() method on ansible.plugins.connection.winrm.Connection. Does this always raise a AttributeError?

alikins:
What kind of exceptions were getting raised that were causing problems?

lib/ansible/executor/task_executor.py
nitzmahone: Yes (currently), and yes. I could see other CPs adding support for it at some point (eg, SSH zap ControlPersist state for the given host). I'd originally planned to implement this on ConnectionBase last year when I added it to winrm.py, but there was some objection (don't recall what now).

nitzmahone:
As there's a very large chunk of code that executes to run a module, it was failing all over the place with varying exceptions. Non-WinRM connections can fail the same way (if, for example, the sshd is bounced by an update)- depending on where it fails, it may or may not come back as an "Unreachable" or any other particular kind of failure. I don't think we want to say that all CPs need to wrap any Exception they could raise anywhere in some kind of `AnsibleItBrokeException`, so I think we're left with treating any exception during polling as a transient failure until the async timeout has elapsed.

alikins:
> I don't think we want to say that all CPs need to wrap any Exception they could raise anywhere in some kind of AnsibleItBrokeException

I would totally vote for 'connection plugins will only raise AnsibleConnectionError' or similar on the roadmap. Especially since ssh and docker are using subprocesses for most of the error prone parts and already have to[1]. But not for this pr...

[1] though ssh can raise/reraise at least OSError and AnsibleError isn't much better

PR #17766, opened 2016-09-27
Moved the _inventory.clear_group_dict_cache() from creating a group w…

jamessewell:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

group_by
##### ANSIBLE VERSION

HEAD
##### SUMMARY

See https://github.com/ansible/ansible-modules-core/issues/5046

### DIFF
--- a/lib/ansible/plugins/strategy/__init__.py
+++ b/lib/ansible/plugins/strategy/__init__.py
@@ -612,10 +612,12 @@ class StrategyBase:
             changed = True
 
         if group_name not in host.get_groups():
-            # clear cache of group dict, which is used in magic host variables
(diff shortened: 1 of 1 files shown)


jimi-c:
@jamessewell rather than move it to where you have it in this PR, would it make more sense instead to move it to a new if block that does:

<pre>
if changed:
    self._inventory.clear_group_dict_cache()
</pre>

I'm concerned that removing it from the other section might introduce other weird bugs.

jamessewell:
Changed as suggested

jimi-c:
Merged, thanks!

PR #17780, opened 2016-09-27
copy: Use the local file's mode for the argument if not explicitly given.

### LINKED ISSUE #1124: copy module should preserve mode of src file when mode is not explicitly set
##### Summary:

When using the copy module to copy multiple files to their destination, it would be really useful if the mode of the source was preserved. This would allow copying several files in a list without explicitly setting modes for each one.

In any case, this is the expected behavior when going by the documentation (which does not specify a default for mode -- leading one to assume the mode would be preserved unless explicitly specified).
##### Steps To Reproduce:

```
- name: copy test
  copy: src={{ item }} dest=/destination owner=root group=root
  with_items:
    - test1.sh
    - test1.conf
```

with test1.sh locally with 0755 permissions and test1.conf locally with 0644 permissions
##### Expected Results:

test1.sh on target with 0755 permissions and test1.conf on target with 0644 permissions
##### Actual Results:

test1.sh and test1.conf on target with 0644 permissions.

If you were to want to maintain the current behavior for backwards compatibility, the default mode should at least be documented (with possibly a recommendation to look at the synchronize module instead). Alternatively this could be done with something like 

```
- name: copy test
  copy: src={{ item }} dest=/destination owner=root group=root mode=preserve
  with_items:
    - test1.sh
    - test1.conf
```

jctanner:
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

copy action plugin
##### ANSIBLE VERSION

```
2.2
```
##### SUMMARY

Fixes https://github.com/ansible/ansible-modules-core/issues/1124

### DIFF
--- a/lib/ansible/plugins/action/copy.py
+++ b/lib/ansible/plugins/action/copy.py
@@ -151,7 +151,6 @@ class ActionModule(ActionBase):
         mode_value = self._task.args.get('mode', None)
 
         diffs = []
-        mode_value = self._task.args.get('mode', None)
         for source_full, source_rel in source_files:
 
             source_full = self._loader.get_real_file(source_full)
--- a/lib/ansible/plugins/action/copy.py
+++ b/lib/ansible/plugins/action/copy.py
@@ -155,6 +155,12 @@ class ActionModule(ActionBase):
             # Generate a hash of the local file.
             local_checksum = checksum(source_full)
 
+            # Get the local mode and set if not user defined
+            # https://github.com/ansible/ansible-modules-core/issues/1124
+            if not self._task.args.get('mode', None):
+                lmode = oct(os.stat(source_full).st_mode & 0777)
+                self._task.args['mode'] = lmode
+
             # If local_checksum is not defined we can't find the file so we should fail out.
             if local_checksum is None:
                 result['failed'] = True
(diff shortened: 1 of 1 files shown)


lib/ansible/plugins/action/copy.py
```diff
@@ -155,6 +155,12 @@ def run(self, tmp=None, task_vars=None):
             # Generate a hash of the local file.
             local_checksum = checksum(source_full)
 
+            # Get the local mode and set if not user defined
+            # https://github.com/ansible/ansible-modules-core/issues/1124
+            if not self._task.args.get('mode', None):
```
abadger: make the conditional check is None since someone could specify mode=0.

lib/ansible/plugins/action/copy.py
```diff
@@ -155,6 +155,12 @@ def run(self, tmp=None, task_vars=None):
...
+                lmode = oct(os.stat(source_full).st_mode & 0o777)
```
abadger: Instead of oct, use '0%03o' % os.stat(source_full).st_mode & 0o777 for python3 compat.

I just remembered there's also a function in the python stat module which may do the & 0o777 a bit prettier:

```
import stat
lmode = '0%03o' % stat.S_IMODE(os.stat(source_full).st_mode)
```

PR #17782, opened 2016-09-27
Make password lookup treat /dev/null as a special case

mscherer:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

password lookup
##### SUMMARY

I have from time to time a need of random password without
wanting to write them down (one example is mailman list creation,
that requires a password to be given to be sent to the list owner).

But using /dev/null do not return null, but the empty string, which
doesn't generate a password at all and so do not achieve my use case.

### DIFF
--- a/docs/docsite/rst/playbooks_lookups.rst
+++ b/docs/docsite/rst/playbooks_lookups.rst
@@ -51,6 +51,9 @@ a file at a given filepath.
 If the file exists previously, it will retrieve its contents, behaving just like with_file. Usage of variables like "{{ inventory_hostname }}" in the filepath can be used to set
 up random passwords per host (which simplifies password management in 'host_vars' variables).
 
+A special case is using ``/dev/null`` as a path. The password lookup will generate a new random password each time, but will not write it to ``/dev/null``. This can be used when you need a password
+without storing it on the controller.
+
 Generated passwords contain a random mix of upper and lowercase ASCII letters, the
 numbers 0-9 and punctuation (". , : - _"). The default length of a generated password is 20 characters.
 This length can be changed by passing an extra parameter::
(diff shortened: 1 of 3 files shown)


docsite/rst/playbooks_lookups.rst
```diff
@@ -51,6 +51,9 @@ a file at a given filepath.
 If the file exists previously, it will retrieve its contents, behaving just like with_file. Usage of variables like "{{ inventory_hostname }}" in the filepath can be used to set
 up random passwords per host (which simplifies password management in 'host_vars' variables).
 
+A special case is using /dev/null as a path. The password lookup will generate a new random password each time, but will not write it to /dev/null. This can be used when you need password
```
mattclay: Wrap `/dev/null` as:

```
``/dev/null``
```

The last sentence should read:

```
This can be used when you need a password without storing it on the controller.
```

mattclay (changes requested):
Aside from the minor doc changes, it would be good to add an integration test for this new feature:

https://github.com/ansible/ansible/blob/devel/test/integration/targets/lookups/tasks/main.yml#L36

mscherer:
Good idea, I will add it.

mattclay (approved):

docsite/rst/playbooks_lookups.rst
```diff
@@ -51,6 +51,9 @@ a file at a given filepath.
 If the file exists previously, it will retrieve its contents, behaving just like with_file. Usage of variables like "{{ inventory_hostname }}" in the filepath can be used to set
 up random passwords per host (which simplifies password management in 'host_vars' variables).
 
+A special case is using ``/dev/null`` as a path. The password lookup will generate a new random password each time, but will not write it to /dev/null. This can be used when you need a password
```
mattclay: The second use of `/dev/null` here should be quoted the same as the first use.

mattclay (changes requested):

mattclay (approved):

PR #17794, opened 2016-09-28
The mode that we fake from stat is checked more closely under python3.5 on El Capitan

abadger:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

unittests for the symbolic mode conversion in basic.py
##### ANSIBLE VERSION

```
devel
```
##### SUMMARY

Python2 seems to allow any integer.  Python3.5 on Linux seems to allow
a 32 bit unsigned int.  Python3.5 on El Capitan seems to limit it to
a smaller size...  perhaps a 16 bit int.


abadger:
This is an alternative fix for #17658

Issue #12213, opened 2015-09-02
Fact cache being ignored on 2.0?


mattwillsher:
```
ansible 2.0.0 (devel 843138e7d9) last updated 2015/09/02 18:58:01 (GMT +100)
  lib/ansible/modules/core: (detached HEAD ef1f31c652) last updated 2015/08/31 21:19:36 (GMT +100)
  lib/ansible/modules/extras: (detached HEAD 5b39e19387) last updated 2015/08/31 21:19:36 (GMT +100)
  config file = /Users/matt/Development/ansible/ansible.cfg
  configured module search path = None
```

ansible.cfg

```
[default]
gathering = smart
fact_caching = jsonfile
fact_caching_connection = factcache
fact_caching_timeout = 86400
```

In 1.9.2 setup doesn't show up when the cache is used. 
2.0.0 calls setup each time and setup appears slower that in 1.9.2/

jimi-c:
I see the problem. If `gather_facts: smart` is specified in the play, things work as expected, however the setting is not being picked up from the ansible.cfg correctly. I'll work on a patch shortly.

jimi-c:
Well actually, it appears that `smart` in the play may simply make it skip fact gathering as well, so continuing to investigate.

jimi-c:
Scratch what I said above, I thought it could be specified at the play level, but was incorrect. Still looking into why it's not picking that setting up.

jimi-c:
# Closing This Ticket

Hi!

We believe the above commit should resolve this problem for you. This will also be included in the next major release.

If you continue seeing any problems related to this issue, or if you have any further questions, please let us know by stopping by one of the two mailing lists, as appropriate:
- https://groups.google.com/forum/#!forum/ansible-project - for user questions, tips, and tricks
- https://groups.google.com/forum/#!forum/ansible-devel - for strategy, future planning, and questions about writing code

Because this project is very active, we're unlikely to see comments made on closed tickets, but the mailing list is a great way to ask questions, or post if you don't think this particular issue is resolved.

Thank you!

mattwillsher:
That's fixed if for me. Thanks, @jimi-c

PR #17799, opened 2016-09-28
Fix gathering=smart and custom module facts

### LINKED ISSUE #17213: "gather_facts: yes" ignored in combination with gathering = smart and a custom module
##### ISSUE TYPE
- Bug Report
##### COMPONENT NAME

ansible facts
##### ANSIBLE VERSION

```
ansible 2.1.1.0
```
##### CONFIGURATION

In ansible.cfg:
gathering = smart

Environment:
PYTHONPATH=/srv/ansible/2.1.1/lib:/srv/python-libs
PATH=/srv/ansible/2.1.1/bin:/usr/local/bin:/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/sbin
##### OS / ENVIRONMENT

RHEL7
##### SUMMARY

Since Ansible v2 I've noticed the setup module is under certain circumstances skipped regardless of whether "gather_facts: yes" is set. I've written a minimal playbook and module to illustrate the issue. It seems related to the "gathering = smart" setting. Setting it to implicit solves the problem. I also noticed the issue is gone when I don't return "ansible_facts" in the module but custom variables (to be accessed via the "register" construct) instead.
##### STEPS TO REPRODUCE

```
$ cat test.yml
- name: play1
  hosts: all
  gather_facts: no
  tasks:
  - test_bug:
- name: play2
  hosts: all
  gather_facts: yes
  tasks:
  - debug: msg='{{ ansible_virtualization_type }}'


$ cat library/test_bug
#!/usr/bin/python

def main():
   module = AnsibleModule(
      argument_spec = dict()
   )

   dictio={ 'test': 'success' }
   module.exit_json(changed=False, ansible_facts=dictio)

if __name__ == '__main__':
    main()


$ ansible-playbook -l hostname -vv test.yml
```

##### EXPECTED RESULTS

The "gather_facts: yes" should instruct Ansible to retrieve the system specific facts.
##### ACTUAL RESULTS

The setup module is never launched and as a result the playbook fails when trying to use the 'ansible_virtualization_type' variable

```
$ ansible-playbook -l hostname -vv test.yml

2 plays in test.yml

PLAY [play1] *******************************************************************

TASK [test_bug] ****************************************************************
task path: /home/user/git/ansible/test.yml:5
ok: [el2582] => {"ansible_facts": {"test": "success"}, "changed": false}

PLAY [play2] *******************************************************************

task path: /home/user/git/ansible/test.yml:10
fatal: [hostname]: FAILED! => {"failed": true, "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'ansible_virtualization_type' is undefined\n\nThe error appears to have been in '/home/user/git/ansible/test.yml': line 10, column 5, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n  tasks:\n  - debug: msg='{{ ansible_virtualization_type }}'\n    ^ here\nWe could be wrong, but this one looks like it might be an issue with\nmissing quotes.  Always quote template expression brackets when they\nstart a value. For instance:\n\n    with_items:\n      - {{ foo }}\n\nShould be written as:\n\n    with_items:\n      - \"{{ foo }}\"\n"}

 [WARNING]: Could not create retry file 'test.retry'.         [Errno 2] No such file or directory: ''


el2582                     : ok=1    changed=0    unreachable=0    failed=1

```

alikins:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

fact gathering policy
##### ANSIBLE VERSION

```
ansible 2.2.0 (gather_smart_facts_17213 4649e70eec) last updated 2016/09/28 15:56:26 (GMT -400)
  lib/ansible/modules/core: (detached HEAD 7808f42aff) last updated 2016/09/28 15:39:58 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 881d2f95a8) last updated 2016/09/28 15:39:58 (GMT -400)

```
##### SUMMARY

If the 'gathering' facts collection config was set
to 'smart', any module that returned ansible_facts would
mark a host as having had _gathered_facts=True, by way
of adding the hostname to the fact cache for module facts,
even if 'setup' itself was never invoked and no other facts
were cached.

Instead of allowing existense of the hostname key in fact
cache to determine initial _gathered_facts, only set if
if the 'setup' action was invoked.

Fixes #17213

### DIFF
--- a/lib/ansible/plugins/strategy/__init__.py
+++ b/lib/ansible/plugins/strategy/__init__.py
@@ -481,6 +481,8 @@ class StrategyBase:
                                 if original_task.action == 'set_fact':
                                     self._variable_manager.set_nonpersistent_facts(target_host, result_item['ansible_facts'].copy())
                                 else:
+                                    if original_task.action == 'setup':
+                                        target_host._gathered_facts = True
                                     self._variable_manager.set_host_facts(target_host, result_item['ansible_facts'].copy())
 
                 if 'diff' in task_result._result:
--- a/lib/ansible/executor/play_iterator.py
+++ b/lib/ansible/executor/play_iterator.py
@@ -196,10 +196,7 @@ class PlayIterator:
         start_at_matched = False
         for host in inventory.get_hosts(self._play.hosts):
             self._host_states[host.name] = HostState(blocks=self._blocks)
-            # if the host's name is in the variable manager's fact cache, then set
-            # its _gathered_facts flag to true for smart gathering tests later
-            if host.name in variable_manager._fact_cache:
-                host._gathered_facts = True
+
             # if we're looking to start at a specific task, iterate through
             # the tasks for this host until we find the specified task
             if play_context.start_at_task is not None and not start_at_done:


lib/ansible/plugins/strategy/__init__.py
```diff
@@ -481,6 +481,8 @@ def parent_handler_match(target_handler, handler_name):
                                 if original_task.action == 'set_fact':
                                     self._variable_manager.set_nonpersistent_facts(target_host, result_item['ansible_facts'].copy())
                                 else:
+                                    if original_task.action == 'setup':
+                                        target_host._gathered_facts = True
```
bcoca, line 485: I'm not sure why/who changed this, but originally we used this fact `module_setup=True` to figure out if 'facts ran', this also allows for users to substitute custom fact modules as 'main fact gathering'.

alikins: Running it without this code is enough to get the reproduce case to pass, but that more or less reverts
https://github.com/ansible/ansible/commit/4ac2bafc4baa59d06c937eb8cb9ade0c6ae45a92 

Not really sure how https://github.com/ansible/ansible/commit/4ac2bafc4baa59d06c937eb8cb9ade0c6ae45a92 fixed https://github.com/ansible/ansible/issues/12213 , so just removing it may be a regression. pinging @jimi-c

alikins: A 'module_setup' fact sounds like a good indicator. Would it make sense to set _gathered_facts here based on that fact, or try to check 'module_setup' fact directly in play_interator?

bcoca: that is how it worked before, not sure why/when it changed

bcoca: 'before' it was checked directly in play_iterator (before that in runner), if the host property is only used there, that would remove the need for it. 
The other advantage is that it also prevents gathering if the 'cache' is fresh.

jimi-c:
We're going to fix this another way, per discussion today.

Issue #13401, opened 2015-12-02
ERROR! SSH Error: data could not be sent to the remote host. Make sure this host can be reached over ssh


tobinlandricombe:
Issue Type: 

Bug Report - related to #12594

Ansible Version & Configuration:

```
$ ansible --version
ansible 2.0.0
  config file = 
  configured module search path = Default w/o overrides
```

Environment: OSX 10.9.5

Summary:
Upgraded to 2.0.0 and started getting error above

Steps To Reproduce:

```
$ cat .ssh/config
# remote.site requires a cert; test requires a password
Host test
        ProxyCommand ssh -q -A -W %h:22 me@remote.site -p 28194
$ ansible -m ping -c ssh -u root -k -i test test
```

Expected Results:

```
No config file found; using defaults
SSH password: 
test | SUCCESS => {
    "changed": false, 
    "ping": "pong"
}
```

Actual Results:

```
No config file found; using defaults
SSH password: 
test | UNREACHABLE! => {
    "changed": false, 
    "msg": "ERROR! SSH Error: data could not be sent to the remote host. Make sure this host can be reached over ssh", 
    "unreachable": true
}
```

tobinlandricombe:
```
$ ansible -vvvvv -m ping -c ssh -u root -k -i test test
No config file found; using defaults
SSH password: 
Loaded callback minimal of type stdout, v2.0
<test> ESTABLISH SSH CONNECTION FOR USER: root
<test> SSH: ansible.cfg set ssh_args: (-o)(ControlMaster=auto)(-o)(ControlPersist=60s)
<test> SSH: ANSIBLE_REMOTE_USER/remote_user/ansible_user/user/-u set: (-o)(User=root)
<test> SSH: ANSIBLE_TIMEOUT/timeout set: (-o)(ConnectTimeout=10)
<test> SSH: PlayContext set ssh_common_args: ()
<test> SSH: PlayContext set ssh_extra_args: ()
<test> SSH: found only ControlPersist; added ControlPath: (-o)(ControlPath=/Users/tobin/.ansible/cp/ansible-ssh-%h-%p-%r)
[...]

PR #17754, opened 2016-09-26
Fix for failed ssh connection and connection timeouts

pasoroki:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

plugins/connection/ssh.py
##### ANSIBLE VERSION

ansible 2.2.0
##### SUMMARY

Fixes #14426 
Fixes #13278 

Solution for SSH module issues:
1) Timeouts on privilege escalation (e.g: https://github.com/ansible/ansible/issues/14426)
2) Failed SSH connection

Ansible v2 already have patch for failed SSH connections (https://github.com/ansible/ansible/pull/9927). Unfortunately that is only works for ssh commands. File copy isn't covered, so it is still can fall down during "copy" or "template" execution.

In the meantime Timeouts caused by ssh misbehaving. Sometimes it don't respond.

I run simple test: 100 times execute command:

```
ssh -C -vvv -o StrictHostKeyChecking=no -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o User=ubuntu -o ConnectTimeout=30 $HOSTNAME '/bin/sh -c '"'"'( umask 77 && mkdir -p "` echo $HOME/.ansible/tmp/ansible-tmp-1474730393.59-109335297653381 `" && echo ansible-tmp-1474730393.59-109335297653381="` echo $HOME/.ansible/tmp/ansible-tmp-1474730393.59-109335297653381 `" ) && sleep 0'"'"''
```

In result : 5 timeouts out of 100

If I introduce retries:
1 retry worked for 90% of my playbook actions. With 3 retries I had no issues left.

This patch uses code from https://github.com/ansible/ansible/pull/9927 as base, I just replaced _run with that code and introduced additional error for Timeouts.

PR #17813, opened 2016-09-29
Smart mode for sftp+scp

jctanner:
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

ssh connection plugin
##### ANSIBLE VERSION

```
2.2
```
##### SUMMARY

If the first method fails, roll to the next method. This saves users
from having to know about the scp_if_ssh method when sftp is broken
on the remote host.

Addresses #13401


lib/ansible/plugins/connection/ssh.py
```diff
@@ -641,18 +643,45 @@ def put_file(self, in_path, out_path):
...
-            cmd = self._build_command('sftp', to_bytes(host))
-            in_data = u"put {0} {1}\n".format(pipes.quote(in_path), pipes.quote(out_path))
-
-        in_data = to_bytes(in_data, nonstring='passthru')
-        (returncode, stdout, stderr) = self._run(cmd, in_data)
-
-        if returncode != 0:
-            raise AnsibleError("failed to transfer file to {0}:\n{1}\n{2}".format(to_native(out_path), to_native(stdout), to_native(stderr)))
+        # since this can be a non-bool now, we need to handle it correctly
+        scp_if_ssh = C.DEFAULT_SCP_IF_SSH
```
bcoca, line 649: i would move this into the 'else' below when it does not match 'smart', also probably should be .lower()

bcoca: also need to return AnsibleOptionsError if the value of the config is incorrect

alikins:
Does fetch_file() need similar changes?  

There is a pr (https://github.com/ansible/ansible/pull/17754/files) that combines some of the logic for exec_command/put_file/get_file (mostly for handling timeouts) and that makes me think the error handling should apply to both.

lib/ansible/plugins/connection/ssh.py
```diff
@@ -641,18 +645,49 @@ def put_file(self, in_path, out_path):
...
+        if not isinstance(scp_if_ssh, bool):
```
alikins, line 650: I'd move this block into it's own method since fetch_file will likely need it as well.

lib/ansible/plugins/connection/ssh.py
```diff
@@ -641,18 +645,49 @@ def put_file(self, in_path, out_path):
...
+        success = False
+        res = None
+        for method in methods:
+            if method == 'sftp':
+                cmd = self._build_command('sftp', to_bytes(host))
+                in_data = u"put {0} {1}\n".format(pipes.quote(in_path), pipes.quote(out_path))
+            elif method == 'scp':
+                cmd = self._build_command('scp', in_path, u'{0}:{1}'.format(host, pipes.quote(out_path)))
+                in_data = None
+
```
alikins, line 673: may be cleaner to build the list of methods, iterate on that list to build a list of (cmd,in_data,checkrc) for each method, then iterate over that list when invoking it. something like:

``` python
method_names =['sftp']
if scp_if_ssh == 'smart':
     method_names.append('scp')
elif scp_if_ssh:
     method_names = ['scp']

methods = []
for method_name in method_names:
    if method_name == 'sftp':
          methods.append(self._build_sftp_put_command(host, in_path, out_path))
   if method_name == 'scp':
         methods.append(self._build_scp_put_command(host, in_path, out_path))

for cmd, in_data, checkrc, method in methods:
    (returncode, stdout, stderr) = self._run(cmd, in_data, checkrc)
    if returncode == 0:
         return
   if scp_if_ssh == 'smart':
         <the display warning/debug code here>

# nothing succeeded and returned
raise AnsibleError...
```

alikins (commented):
Like the idea of 'smart' scp usage, but would like to see the impl be a bit more general.

PR #17817, opened 2016-09-29
Refactor parsing of CLI args so that we can modify them in the base class

abadger:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

cli
##### ANSIBLE VERSION

```
devel
```
##### SUMMARY

Implement tag and skip_tag handling in the CLI() class.  Change tag and
skip_tag command line options to be accepted multiple times on the CLI
and add them together rather than overwrite.

### DIFF
--- a/docsite/rst/intro_configuration.rst
+++ b/docsite/rst/intro_configuration.rst
@@ -179,17 +179,6 @@ different locations::
 
 Most users will not need to use this feature.  See :doc:`developing_plugins` for more details
 
-.. _stdout_callback:
-
-stdout_callback
-===============
-
-.. versionadded:: 2.0
-
-This setting allows you to override the default stdout callback for ansible-playbook::
-
-    stdout_callback = skippy
-
 .. _callback_whitelist:
 
 callback_whitelist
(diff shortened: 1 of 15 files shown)


abadger:
Although a feature pull request, this will fix the scripts we have for running integration tests on shippable for when they use tags to limit which tests are to be run.  (Right now, changes to the mount module will fail their python3 tests because the tags from mount will override the tags from python3.)  Being able to specify the --tags parameter multiple times allows us to leave the code in the bash script simple.

This change also has the nice effect of cleaning up the ansible code.  The cli can worry about parsing and transforming the data received from the command line args into a usage format.  The playbook code only has to worry about making use of that.

abadger:
mattclay, bcoca, and I discussed this on IRC.  decided that some may have scripts which depend on the overwrite behaviour so we're going to target for 2.3 and go through the deprecation process for it.

abadger:
Okay, bcoca gave an earlier version of this a +1.   Will wait until tomorrow to see if either @bcoca or @mattclay want to take a look at this final revision.
ready_for_review

PR #17823, opened 2016-09-29
Filetree deindent

alikins:
##### ISSUE TYPE

- Cleanup
##### COMPONENT NAME

lib/ansible/plugins/lookup/filetree.py
##### ANSIBLE VERSION

```
ansible 2.2.0 (Conditional_classmethod_constructor ded30ac93e) last updated 2016/09/29 19:23:58 (GMT -400)
  lib/ansible/modules/core: (detached HEAD 537a7eb924) last updated 2016/09/29 16:48:01 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD a58e1d59c0) last updated 2016/09/29 16:48:01 (GMT -400)

```
##### SUMMARY

Reduce code nesting in filetree.py

```
Move the use of os.walk() into it's own
method to lessen the indent level (7)
```

```

```

### DIFF
--- a/lib/ansible/plugins/lookup/filetree.py
+++ b/lib/ansible/plugins/lookup/filetree.py
@@ -109,10 +109,12 @@ def walk(path, topdown=None):
             relpath = os.path.relpath(os.path.join(root, entry), path)
 
             # Skip if relpath was already processed (from another root)
-            if relpath not in [entry['path'] for entry in ret]:
-                props = file_props(path, relpath)
-                if props is not None:
-                    ret.append(props)
+            if relpath in [entry['path'] for entry in ret]:
+                continue
+
+            props = file_props(path, relpath)
+            if props is not None:
+                ret.append(props)
     return ret
 
 class LookupModule(LookupBase):
--- a/lib/ansible/plugins/lookup/filetree.py
+++ b/lib/ansible/plugins/lookup/filetree.py
@@ -22,10 +22,10 @@ import pwd
 import grp
 import stat
 
-HAVE_SELINUX=False
+HAVE_SELINUX = False
 try:
     import selinux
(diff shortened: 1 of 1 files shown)


lib/ansible/plugins/lookup/filetree.py
```diff
@@ -100,6 +100,22 @@ def file_props(root, path):
 
     return ret
 
+def walk(path, topdown=None):
+    topdown = topdown or False
```
robinro, line 104: `topdown` is always `True`, so don't make it an argument here.

lib/ansible/plugins/lookup/filetree.py
```diff
@@ -100,6 +100,22 @@ def file_props(root, path):
...
+
+    ret = []
+    for root, dirs, files in os.walk(path, topdown=True):
+        for entry in dirs + files:
+            relpath = os.path.relpath(os.path.join(root, entry), path)
+
+            # Skip if relpath was already processed (from another root)
+            if relpath in [entry['path'] for entry in ret]:
```
robinro, line 112: This is really inefficient, it creates a O(n) list at each step. Either have a list of paths in parallel to ret or fine a more efficient way to check the path is already included.

alikins: I think my version is buggy as well, since the `entry['path'] for entry in ret` in walk() will only check the local 'ret', which will be per term, where the original check against ret would see all results.

I could pass the run() level 'ret' to walk() and let walk modify it accumulating results. And something
similar for a 'visited_paths' collection. 

I sketched out a generator based version that could work, but is currently unfinished.

alikins:
closing for now, this is buggy. I'll try to revisit[1] with a generator based walk().

[1] pun slightly intended

PR #4386, opened 2013-10-07
Enable do-until to fail on max retries completed scenario

bennojoy:
This pr would make the task fail if the max retries is exhausted in do until loop.

PR #17831, opened 2016-09-30
add path arg to _write_backup method

robphoenix:
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

add functionality to be able to specify backup path
##### ANSIBLE VERSION

```
ansible 2.1.1.0
```
##### SUMMARY

From conversations in [issue #4386 in ansibles-module-core](https://github.com/ansible/ansible-modules-core/issues/4386)
This adds the ability to specify the backup path where device configurations can be written, instead of just defaulting to a directory called backups in the cwd. This is especially useful when running Ansible in a Vagrant environment on windows hosts when using shared directories, as it is often not possible for ansible to write to the shared directory. I realise this reason is fairly niche, but I expect this functionality may still be useful

I considered adding a `try/except` block around the file write operation:

```
try:
    with open(filename, 'w') as f:
                f.write(contents)
except EnvironmentError as exc:
    return dict(failed=True, msg=exc.message)
```

to catch any exceptions due to the specified path not being writable but didn't want to return a different type to the `filepath` variable in the `run` method. Is this an exception that needs catching in another manner?

This is a first code contribution to Ansible, and I'm not certain I've done everything I should do, so any comments, suggestions, criticisms are welcome.

### DIFF
--- a/lib/ansible/plugins/action/net_config.py
+++ b/lib/ansible/plugins/action/net_config.py
@@ -36,8 +36,8 @@ class ActionModule(ActionBase):
 
     TRANSFERS_FILES = False
 
-    def run(self, tmp=None, task_vars=None):
-        result = super(ActionModule, self).run(tmp, task_vars)
+    def run(self, tmp=None, task_vars=None, path=None):
+        result = super(ActionModule, self).run(tmp, task_vars, path)
         result['changed'] = False
 
         if self._task.args.get('src'):
@@ -55,7 +55,7 @@ class ActionModule(ActionBase):
             # User requested backup and no error occurred in module.
             # NOTE: If there is a parameter error, _backup key may not be in results.
             filepath = self._write_backup(task_vars['inventory_hostname'],
-                                          result['__backup__'])
+                                          result['__backup__'], path)
             result['backup_path'] = filepath
 
         # strip out any keys that have two leading and two trailing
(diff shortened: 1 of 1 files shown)


lib/ansible/plugins/action/net_config.py
```diff
@@ -36,8 +36,8 @@ class ActionModule(ActionBase):
 
     TRANSFERS_FILES = False
 
-    def run(self, tmp=None, task_vars=None):
-        result = super(ActionModule, self).run(tmp, task_vars)
+    def run(self, tmp=None, task_vars=None, path=None):
```
privateip, line 39: This isn't where you are going to get the path information.  Most likely the backup path will be passed from the module in which case you will find that information in `_task_args`.

lib/ansible/plugins/action/net_config.py
```diff
@@ -72,10 +72,13 @@ def _get_working_path(self):
             cwd = self._task._role._role_path
         return cwd
 
-    def _write_backup(self, host, contents):
-        backup_path = self._get_working_path() + '/backup'
+    def _write_backup(self, host, contents, path=None):
+        if path is None:
+            backup_path = os.path.join(self._get_working_path(), 'backup')
+        else:
+            backup_path = os.path.abspath(path)
```
privateip, line 79: I'm not sure this is how this should be tackled as it sets the wrong precedence.  If the module arg is a relative path it should be relative to the playbook or role root dir.   If the back path is absolute then this is okay.

lib/ansible/plugins/action/net_config.py
```diff
@@ -72,10 +72,13 @@ def _get_working_path(self):
             cwd = self._task._role._role_path
         return cwd
 
-    def _write_backup(self, host, contents):
-        backup_path = self._get_working_path() + '/backup'
+    def _write_backup(self, host, contents, path=None):
```
privateip, line 75: Assuming path comes from the module args, you can access the path information from the instance variable `_task_args` so no need to pass it here

privateip (changes requested):
Couple of overall notes knowing this is your first contribution to Ansible.   Spend some time reviewing how modules execute and where values are stored.  In addition, there will need to be the module side code before we can accept the PR into Ansible so we understand how the whole feature works.  We can continue to review this code but it won't get merged until the entire feature is complete.

privateip:
needs_revision

robphoenix:
Thankyou for the feedback on this @privateip 
I'll take some time to digest it all and gain a better understanding of module execution.
This is all really useful in understanding Ansible, so thanks again for taking the time to comment.
I'll close this PR, and open a new one when I've got a better idea of how to implement this.

PR #17834, opened 2016-09-30
bugfixes to JSON junk filter, added unit/integration tests to exercise

nitzmahone:
##### ISSUE TYPE
- Bugfix Pull Request

https://github.com/ansible/ansible-modules-core/pull/5107 needs to be merged first before these tests will pass


abadger:
Looks good but let's move `_filter_non_json_lines(data)` into module_utils now so that we aren't tempted to think of it as just an action function anymore.

I wonder if we should remove the leading underscore at the same time?  Maybe think on that for the future but don't do it yet. [probably should document it's primary purpose is for filtering return values from modules.  Other uses cannot be guaranteed to keep working.

abadger:
Oh -- we should also add a comment that says we have a copy in async_wrapper.  That way there's a chance someone who needs to modify this will realize they need to modify the copy in async_wrapper as well.

nitzmahone:
Yep- I almost did that one too but was on the fence. Decided to go for minimal change due to late date, but seems like a pretty harmless change.

Issue #17846, opened 2016-10-01
handlers listen: duplicate named handlers break handlers listen


resmo:
##### ISSUE TYPE

- Bug Report
##### COMPONENT NAME

handlers listen
##### ANSIBLE VERSION

```
2.2
```
##### CONFIGURATION

N/A
##### OS / ENVIRONMENT

N/A
##### SUMMARY

While writing test for handlers listen, I found inconsistencies in how handlers are marked for execution:

If handlers listen, it is not necessarily a must that they require a name (as we notify the listen param), if the task notifies two or more handlers having the same module (here set_fact) only one handler will be executed:

``` yaml
  handlers:
    - set_fact:
      notify_listen_ran_1: True
      listen: notify_listen

    - set_fact:
        notify_listen_ran_2: True
      listen: notify_listen
```

Because ansible uses the "module/plugin" name (https://github.com/ansible/ansible/blob/devel/lib/ansible/playbook/task.py#L120 and https://github.com/ansible/ansible/blob/devel/lib/ansible/executor/task_queue_manager.py#L150), only the first handler will be executed.
##### STEPS TO REPRODUCE

``` yaml
- name: test handlers listen
  hosts: localhost
  gather_facts: false
  connection: local
  pre_tasks:
    - name: notify some handlers
      command: echo foo
      notify:
        - notify_listen
  tasks:
    - meta: flush_handlers

    - name: assert all defined handlers ran without error
      assert:
        that:
        - "notify_listen_ran_1 is defined"
        - "notify_listen_ran_2 is defined"

  handlers:
    - set_fact:
        notify_listen_ran_1: True
      listen: notify_listen

    - set_fact:
        notify_listen_ran_2: True
      listen: notify_listen
```
##### EXPECTED RESULTS

My suggestion is, that we enforce to have a handler name specified. So Ansible should fail for handlers without a name.
##### ACTUAL RESULTS

```
No config file found; using defaults
 [WARNING]: Host file not found: /etc/ansible/hosts

 [WARNING]: provided hosts list is empty, only localhost is available


PLAYBOOK: test_handlers_listen2.yml ********************************************
1 plays in test_handlers_listen2.yml

PLAY [test handlers listen] ****************************************************

TASK [notify some handlers] ****************************************************
task path: /home/resmo/Projects/resmo/ansible/test/integration/test_handlers_listen2.yml:6
NOTIFIED HANDLER set_fact
[...]

PR #17847, opened 2016-10-01
handler: force handlers to be named

resmo:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

handler
##### ANSIBLE VERSION

```
2.2
```
##### SUMMARY

As described in #17846, unnamed handlers tend to have an unexpected behavior with handler listen. Enforcing a handler to be named.


bcoca:
As discussed in meeting, this is not the appropriate fix, even if we decide that handlers are allowed non unique names when using `listen`.

PR #17866, opened 2016-10-03
Remove unicode escape

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

plugins/action/assemble.py
##### ANSIBLE VERSION

```
devel
```
##### SUMMARY

The unicode-escape codec is not present in python3 as it is a text=>text transformation.  Replace it with the codecs.escape_decode() function.


abadger:
This is an alternate fix to #17305

PR #17870, opened 2016-10-03
New modules for CHANGELOG.md

gundalow:
##### ISSUE TYPE
- Docs Pull Request
##### COMPONENT NAME

docs
##### SUMMARY

Full review of new modules


gundalow:
~~Should the following two be added

async_status
async_wrapper

I can see a `ps1` added since 2.1, but not corresponding `.py`, I'm not sure if we list these are modules or not.~~

After speak to with Matt the above don't need documenting, they aren't intended for human consumption. He's already added Windows async support to the changelog

PR #17879, opened 2016-10-03
out.split('\n') -> out.splitlines()

alikins:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

##### ANSIBLE VERSION

```
ansible 2.2.0 (password_lookup b6a2f3fe28) last updated 2016/09/09 19:00:26 (GMT -400)
  lib/ansible/modules/core: (detached HEAD bf5b3de83e) last updated 2016/09/09 14:31:05 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD f83aa9fff3) last updated 2016/09/09 14:31:05 (GMT -400)

```
##### SUMMARY

This is related to module_utils.basic.run_command() native type changes, but shouldn't
actually require it (bytes has 'splitlines' as well as str/unicode)

foo.split('\n') is picky about the type of 'foo'.
if 'foo' is a bytes type, then foo.split('\n')
will fail on py3 with:

   TypeError: a bytes-like object is required, not 'str'

The foo.split('\n') change isn't strictly required
when run_command returns native str types, but it
is more idiomatic and conceptually also supports other
line endings.

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -232,18 +232,18 @@ class Facts(object):
             getconf_bin = self.module.get_bin_path('getconf')
             if getconf_bin:
                 rc, out, err = self.module.run_command([getconf_bin, 'MACHINE_ARCHITECTURE'])
-                data = out.split('\n')
+                data = out.splitlines()
                 self.facts['architecture'] = data[0]
             else:
                 bootinfo_bin = self.module.get_bin_path('bootinfo')
                 rc, out, err = self.module.run_command([bootinfo_bin, '-p'])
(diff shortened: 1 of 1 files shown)


abadger:
This looks good to merge to devel.  If we want it in stable-2.2 as well, we need to audit the usages to make sure none of the places actually require split('\n') rather than splitlines()  (the two differences I know of are splitlines hits additional eol characters and has different behaviour about newlines at the end of the string.

alikins:
>  If we want it in stable-2.2 as well, we need to audit the usages

Agreed. I'd like to see some reports from folks on weird platforms have success before merging to stable-2.2.

PR #17881, opened 2016-10-03
Make run_command return native strings

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

module_utils/badic.py
##### ANSIBLE VERSION

```
devel and 2.2
```
##### SUMMARY

basic.py is using a native string strategy for its API.  Make run_command() comply with that.


alikins:
Commit message should make it clear this will change default module api behavior and could break existing modules.

alikins:
+1 on the code, but would like to see more detailed commit message

abadger:
merged and cherry-picked to stable-2.2

PR #17885, opened 2016-10-03
Allow for namespaced galaxy module calls

jctanner:
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

plugin loader
##### ANSIBLE VERSION

```
2.3
```
##### SUMMARY

Allow for a playbook such as ...

```
- hosts: el6host
  connection: local
  gather_facts: False
  tasks:
    - jctanner.pinger.pinger:
```

"jctanner.pinger" is the galaxy role reference and "pinger" is a module file in the role's library path. Once the playbook is run, a few things will happen:
1. Check if the module has two periods in it
2. Split the name into the role name and the module name
3. Check if the role is installed, and install if not
4. Put the module's FQN into the plugin loader

The task will then operate like any other module was called.

### DIFF
--- a/lib/ansible/parsing/mod_args.py
+++ b/lib/ansible/parsing/mod_args.py
@@ -27,9 +27,6 @@ from ansible.plugins import module_loader
 from ansible.parsing.splitter import parse_kv, split_args
 from ansible.template import Templar
 
-# dynamically installed roles ...
-from ansible.galaxy.helpers import get_role_by_module_fqn
-
 # For filtering out modules correctly below
 RAW_PARAM_MODULES = ([
     'command',
@@ -287,11 +284,6 @@ class ModuleArgsParser:
 
         # walk the input dictionary to see we recognize a module name
         for (item, value) in iteritems(self._task_ds):
-            if item.count('.') == 2:
-                # This is a module from a galaxy role
-                gr = get_role_by_module_fqn(item)
-                if gr.module_file not in module_loader._module_cache:
-                    module_loader.add_module_by_file(item, gr.module_file)
 
             if item in module_loader or item in ['meta', 'include', 'include_role']:
                 # finding more than one module name is a problem
--- a/lib/ansible/parsing/mod_args.py
+++ b/lib/ansible/parsing/mod_args.py
@@ -27,6 +27,9 @@ from ansible.plugins import module_loader
 from ansible.parsing.splitter import parse_kv, split_args
 from ansible.template import Templar
 
+# dynamically installed roles ...
+from ansible.galaxy.helpers import get_role_by_module_fqn
+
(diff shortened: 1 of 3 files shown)


lib/ansible/plugins/__init__.py
```diff
@@ -416,6 +416,16 @@ def all(self, *args, **kwargs):
...
             yield obj
 
+    def add_module_by_file(self, name, filename):
+        self._module_cache[name] = self._load_module_source(name, filename)
+        # self._plugin_path_cache[suffix]
+        #self._plugin_path_cache[name] = filename
+        self._plugin_path_cache[''][name] = filename
+        #self.add_directory(os.path.dirname(filename))
+        if not self.find_plugin(name):
+            import sys; sys.exit(1)
```
alikins: Is this debugging that got left in?  Needs to be removed.

alikins (changes requested):
remove sys.exit

lib/ansible/parsing/mod_args.py
```diff
@@ -27,6 +27,9 @@
 from ansible.parsing.splitter import parse_kv, split_args
 from ansible.template import Templar
 
+# dynamically installed roles ...
+from ansible.galaxy.helpers import get_role_by_module_fqn
```
alikins: Seems wrong to have mod_args import galaxy code just to look a module file name for a galaxy role nickname.

Maybe remove the galaxy bits from here, and instead populate the 'jctanner.pinger.pinger' style name as a name or alias in plugins.PluginLoader()? ie, add 'jctanner.pinger.pinger' to the names in the cache in https://github.com/ansible/ansible/blob/unified_ansible/lib/ansible/plugins/__init__.py#L273-L285

Then this module doesn't need to know it's a galaxy specific name.  

(the name could be URI like with a scheme indicating it's a galaxy name 'galaxy_role_module://jctanner/pinger/module/pinger' or whatever. Someway for mod args to tell that is referring to a module in a galaxy role.)

lib/ansible/parsing/mod_args.py
```diff
@@ -284,6 +287,12 @@ def parse(self):
 
         # walk the input dictionary to see we recognize a module name
         for (item, value) in iteritems(self._task_ds):
+            if item.count('.') == 2:
+                # This is a module from a galaxy role
+                gr = get_role_by_module_fqn(item)
+                if gr.module_file not in module_loader._module_cache:
+                    module_loader.add_module_by_file(item, gr.module_file)
```
alikins: This could go away since the 'item' would be enough for plugin loader to get the right module

alikins (changes requested):
avoid importing galaxy into mod_args

PR #17886, opened 2016-10-03
Specify run_command decode error style as arg

alikins:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

##### ANSIBLE VERSION

```ansible 2.3.0 (facts_no_decode 973b65cb53) last updated 2016/10/03 16:07:31 (GMT -400)
  lib/ansible/modules/core: (detached HEAD 6c4d71a7fa) last updated 2016/10/03 12:09:32 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD df35d324d6) last updated 2016/10/03 08:42:50 (GMT -400)

```

##### SUMMARY

```

```

Instead of getting the stdout/stderr text from
run_command, and then decoding to utf-8 with a
particular error scheme, use the 'errors' arg
to run_command so it does that itself.
```

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -2422,7 +2422,7 @@ class LinuxNetwork(Network):
             rc, primary_data, stderr = self.module.run_command(args, errors='surrogate_or_replace')
 
             args = [ip_path, 'addr', 'show', 'secondary', device]
-            rc, secondaty_data, stderr = self.module.run_command(args, errors='surrogate_or_replace')
+            rc, secondary_data, stderr = self.module.run_command(args, errors='surrogate_or_replace')
 
             parse_ip_output(primary_data)
             parse_ip_output(secondary_data, secondary=True)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lib/ansible/module_utils/facts.py [hunk 1]
```diff
@@ -261,9 +261,8 @@ def get_local_facts(self):
                 # try to read it as json first
                 # if that fails read it with ConfigParser
                 # if that fails, skip it
-                rc, out, err = self.module.run_command(fn)
                 try:
-                    out = out.decode('utf-8', 'strict')
+                    rc, out, err = self.module.run_command(fn, errors='strict')
```
alikins: I could see a scenario where stdout is valid utf-8 that can be decoded with strict, but stderr could have garbage that would cause 'strict' to fail. Not sure if it's worth being able to specifying different error schemes for both (or different encodings for that matter)

abadger: i don't think the run_command API should do that bu the calling code can like this:

``` python
rc, out, err = self.module.run_command(fn, encoding=None)
out = to_native(errors='surrogate_or_strict')
err = to_native(errors='surrogate_or_replace')
```

abadger:
Committed to devel and cherry-picked to stable-2.2

lib/ansible/module_utils/facts.py
alikins: Oooh, I like that. +1

PR #17889, opened 2016-10-03
Added iocage tag lookup to jail connector

slohse:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

jail connector
##### ANSIBLE VERSION

```
devel
```
##### SUMMARY

added a lookup for iocage tags in the jail connector. If the given jail name is not found and if iocage is installed, the jail connector now also checks the tags of iocage jails and if it finds the given name, it changes the jail name to the one associated with the iocage tag (`ioc-<UUID>`). Everything after that is unchanged from the normal behaviour of the jail connector.

### DIFF
--- a/lib/ansible/plugins/connection/jail.py
+++ b/lib/ansible/plugins/connection/jail.py
@@ -60,7 +60,13 @@ class Connection(ConnectionBase):
 
         self.jls_cmd = self._search_executable('jls')
         self.jexec_cmd = self._search_executable('jexec')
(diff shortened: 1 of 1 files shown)


slohse:
This solution however does not work if you have a "normal" jail with a name that is the same as the tag of an iocage jail - in that case the normal jail will be the one that is used.
Could it be a better idea to build an iocage connector of its own, that wraps the jail connector?

slohse:
Working on a better solution

Issue #17848, opened 2016-10-01
handler listen: notify a undefined handler must not be an error


resmo:
##### ISSUE TYPE

- Bug Report
##### COMPONENT NAME

handler listen
##### ANSIBLE VERSION

```
2.2
```
##### CONFIGURATION

N/A
##### OS / ENVIRONMENT

N/A
##### SUMMARY

With the new feature of handler listen, a common use case is to notify an inexistent handler. However, currently Ansible expects the handler to be there and shows an error.

The error should be replaced with a warning so potential issues with typos in handler names and listen can be identified.
##### STEPS TO REPRODUCE

``` yaml

---
- name: test handlers listen
  hosts: localhost
  gather_facts: false
  connection: local
  tasks:
    - name: notify some handlers
      command: echo foo
      notify:
        - notify_listen_not_there
```
##### EXPECTED RESULTS

playbook gets executed with a warning.
##### ACTUAL RESULTS

```
 $ ansible-playbook test_handlers_listen2.yml -vv
No config file found; using defaults
 [WARNING]: Host file not found: /etc/ansible/hosts

 [WARNING]: provided hosts list is empty, only localhost is available


PLAYBOOK: test_handlers_listen2.yml ********************************************
1 plays in test_handlers_listen2.yml

PLAY [test handlers listen] ****************************************************

TASK [notify some handlers] ****************************************************
task path: /home/resmo/Projects/resmo/ansible/test/integration/test_handlers_listen2.yml:6
ERROR! The requested handler 'notify_listen_not_there' was found in neither the main handlers list nor the listening handlers list

```

bcoca:
So after discusisng in IRC we all agreed that a missing handler should by default be an error, as such this is not a bug.

In any case I added PR #17902 to allow a user to toggle into a warning for those that have that preference.

PR #17902, opened 2016-10-04
toggle missing handler errors/warnings via config

bcoca:
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

handlers
##### ANSIBLE VERSION

```
2.2
```
##### SUMMARY

Allows user to decide if missing handler is fatal or warning

### DIFF
--- a/examples/ansible.cfg
+++ b/examples/ansible.cfg
@@ -71,6 +71,9 @@
 #task_includes_static = True
 #handler_includes_static = True
 
+# Controls if a missing handler for a notification event is an error or a warning
+#error_on_missing_handler = True
+
 # change this for alternative sudo implementations
 #sudo_exe = sudo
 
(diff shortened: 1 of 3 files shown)


alikins:
Does this fix https://github.com/ansible/ansible/issues/17848 ? if so, commit should reference it.

alikins:
This seems like an odd thing to add a config option for, especially since it defaults to being fatal.
Whats the use case for it being fatal?

bcoca:
@alikins left comments on the other issues as this does not really solve them but give a 'middle ground'. 

As for why it is an error, it mostly because that was the behavior since handlers were introduced, this 'broke' in 2.0 and was a bug for a while but eventually fixed. 

One of the referenced PRs was changing it into a warning but this is not a backwards compatible change, hence this PR.

alikins:
Could there be a default handler that gets invoked if there are no specifically attached handlers?
Then the default handler could warn in this case. Or if it makes sense to be fatal, the
default handler could raise the exception.

bcoca:
that could be a feature but beyond the scope of this PR, this implements what was discussed in the proposal but never made it into the code, as such it was an oversight that this PR remedies.

resmo:
tested
+1

PR #17924, opened 2016-10-06
Add SSL support to Consul inventory script

ghost:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

contrib/inventory/consul_io.py
##### ANSIBLE VERSION

```
ansible 2.1.2.0
  config file = 
```
##### SUMMARY

Add SSL support to the Consul inventory script. If a Consul cluster is behind an SSL load balancer or proxy, the current inventory script will always use `http` as the scheme. This change will derive the scheme from the cluster URL. The default scheme remains `http`.

### DIFF
--- a/contrib/inventory/consul_io.py
+++ b/contrib/inventory/consul_io.py
@@ -465,6 +465,7 @@ class ConsulConfig(dict):
       host = 'localhost'
       port =  8500
       token = None
+      scheme = 'http'
 
       if hasattr(self, 'url'):
           from urlparse import urlparse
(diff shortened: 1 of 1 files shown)


ghost:
Pinging @sgargan for a quick look.

sgargan (approved):
looks good thanks for adding this. Looking at the docs a description for the url is missing (my bad), any chance you could write a line to describe it?

ship it!

ghost:
Thanks for the quick feedback! Just pushed documentation for url option.

sgargan:
@johnnycarlin that's great thanks for adding.

PR #17931, opened 2016-10-07
Adding self.log() on IOErrors and OSErrros

bechampion:
##### ISSUE TYPE
- Feature request

##### COMPONENT NAME

Basic
##### ANSIBLE VERSION

```
ansible 2.2.0.0 (stable-2.2-logging d24341a359) last updated 2016/10/07 10:25:03 (GMT +100)

```
##### SUMMARY
Add syslog logging when commands fail due to IOError or OSError

These two lines just add self.log() to IOErrors and OSErrors , so users will be able to see what went wrong on syslog and be able to send it to kibana and others.

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -2266,11 +2266,11 @@ class AnsibleModule(object):
             rc = cmd.returncode
         except (OSError, IOError):
             e = get_exception()
-            self.log("%s: Error Executing CMD:%s PERM:%s Exception:%s" % (args[0] , args[1] , str(e)))
+            self.log("Error Executing CMD:%s Exception:%s" % (clean_args, to_native(e)))
             self.fail_json(rc=e.errno, msg=to_native(e), cmd=clean_args)
         except Exception:
             e = get_exception()
-            self.log("%s: Error Executing CMD:%s PERM:%s Exception:%s" % (args[0] , args[1] , str(traceback.format_exc())))
+            self.log("Error Executing CMD:%s Exception:%s" % (clean_args,to_native(traceback.format_exc())))
             self.fail_json(rc=257, msg=to_native(e), exception=traceback.format_exc(), cmd=clean_args)
 
         # Restore env settings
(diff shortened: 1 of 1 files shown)


lib/ansible/module_utils/basic.py
```diff
@@ -2266,9 +2266,11 @@ def run_command(self, args, check_rc=False, close_fds=True, executable=None, dat
             rc = cmd.returncode
         except (OSError, IOError):
             e = get_exception()
+            self.log("%s: Error Executing CMD:%s PERM:%s Exception:%s" % (args[0] , args[1] , str(e)))
```
bcoca: I would not add these here as the log itself can also fail due to OSError,IOError

lib/ansible/module_utils/basic.py
```diff
@@ -2266,9 +2266,11 @@ def run_command(self, args, check_rc=False, close_fds=True, executable=None, dat
...
             self.fail_json(rc=e.errno, msg=to_native(e), cmd=clean_args)
         except Exception:
             e = get_exception()
+            self.log("%s: Error Executing CMD:%s PERM:%s Exception:%s" % (args[0] , args[1] , str(traceback.format_exc())))
```
abadger: If we do add something like this, three thoughts on the formatting the string:
- Use clean_args since that will have had private parameters stripped out.
- What's PERM?  Should this be ARGS instead?
- Use to_native(e) and to_native(traceback.format_exc()) instead of str().  (clean_args should be a native string so it shouldn't need to be coerced here).

alikins:
@bechampion I did some small edits to the origin comment (the 'issue type' field) since the bot (@ansibot) was being confused by the previous value  (add_syslog_logging_when_commands_fail_due_to_ioerror_or_oserror) and was making that  a label

lib/ansible/module_utils/basic.py
bechampion: @abadger i will fix this and commit back.
Thanks!

bechampion:
This spits out this:
"Error Executing CMD:iddd Exception:[Errno 2] No such file or directory: 'iddd'"

Thanks!

PR #17948, opened 2016-10-08
Openstack dynamic inventory: respect OS_CLOUD env var

alvaroaleman:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

Openstack dynamic inventory
##### ANSIBLE VERSION

```
devel
```
##### SUMMARY

RIght now the Openstack inventory script always requests -all- defined clouds. This is an extremely expensive operation. This change makes it possible to use the OS_CLOUD environment variable to specify a cloud the same way the openstack command uses it:

``` shell
openstack --help|grep -B1 OS_CLOUD
  --os-cloud <cloud-config-name>
                        Cloud name in clouds.yaml (Env: OS_CLOUD)
```

### DIFF
--- a/contrib/inventory/openstack.py
+++ b/contrib/inventory/openstack.py
@@ -218,6 +218,7 @@ def main():
             refresh=args.refresh,
             config_files=config_files,
             private=args.private,
+            cloud=os.environ.get('OS_CLOUD')
         )
         if hasattr(shade.inventory.OpenStackInventory, 'extra_config'):
             inventory_args.update(dict(
--- a/contrib/inventory/openstack.py
+++ b/contrib/inventory/openstack.py
@@ -29,8 +29,11 @@
 #  - /etc/openstack/clouds.yaml
 #  - /etc/ansible/openstack.yml
 # The clouds.yaml file can contain entries for multiple clouds and multiple
-# regions of those clouds. If it does, this inventory module will connect to
-# all of them and present them as one contiguous inventory.
+# regions of those clouds. If it does, this inventory module will by default
+# connect to all of them and present them as one contiguous inventory.  You
+# can limit to one cloud by passing the `--cloud` parameter, or use the
+# OS_CLOUD environment variable.  If caching is enabled, and a cloud is
+# selected, then per-cloud cache folders will be used.
 #
 # See the adjacent openstack.yml file for an example config file
 # There are two ansible inventory specific options that can be set in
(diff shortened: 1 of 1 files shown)


alvaroaleman:
CC @emonty

rcarrillocruz (commented):
I like this, actually asked Monty about something similar a while ago.

Do you think you could add an option to read a cloud from CLI args too?
If both CLI and envvars are present, CLI prevails.

halberom:
Hello,  I've had a quick play with something very similar to this.  Several points
- adding cli option can be as easy as adding the following diff
```
197,198d196
<     parser.add_argument('--cloud', default=os.environ.get('OS_CLOUD'),
<                         help='Cloud name (default: None')
223d220
<             cloud=args.cloud,
```
- my main concern however is that if cache expiration is defined, you must _always_ refresh the cache when changing the cloud at present.  What would be nicer is if the cache was per cloud (similar to how the aws ec2.py does it - https://github.com/ansible/ansible/blob/devel/contrib/inventory/ec2.py#L394-L395).  Something like
```
111,112c111,112
< def get_host_groups(inventory, refresh=False, cloud=None):
<     (cache_file, cache_expiration_time) = get_cache_settings(cloud)
---
> def get_host_groups(inventory, refresh=False):
>     (cache_file, cache_expiration_time) = get_cache_settings()
179c179
< def get_cache_settings(cloud=None):
---
> def get_cache_settings():
185,186d184
<     if cloud:
<         cache_path = '{0}_{1}'.format(cache_path, cloud)
199,200d196
<     parser.add_argument('--cloud', default=os.environ.get('OS_CLOUD'),
<                         help='Cloud name (default: None')
225d220
<             cloud=args.cloud,
240c235
<             output = get_host_groups(inventory, refresh=args.refresh, cloud=args.cloud)
---
>             output = get_host_groups(inventory, refresh=args.refresh)
```

Which is really messy to read so here's a gist - https://gist.github.com/halberom/2b1659c65cf9af61ebf449a238b0bea2

alvaroaleman:
@halberom can you push that change somewhere so I can cherry-pick it into this PR?

halberom:
here you go https://github.com/halberom/ansible/blob/feature/openstack_os_cloud/contrib/inventory/openstack.py

alvaroaleman:
Thanks @halberom!

samdoran:
@rcarrillocruz Can you look at this again and see if it's ok to merge?

emonty (approved):
shipit

abadger:
Merged to devel for 2.4.0 release

Issue #1210, opened 2012-10-03
Add -q/--quiet option for running Ansible from Cron


dagwieers:
Would this be accepted ?

jpmens:
That certainly would be appreciated here. :)

On the other hand, running `ansible-playbook xx.yml > /dev/null` already works ...

dagwieers:
Not really, fatal errors go to stdout as well. So cron will not send errors out when failed.

jpmens:
Fatal errors to stdout and not stderr is a bug.

dagwieers:
Right, but the general issue here is that if you have a fatal error, you also want the context which at that point has been sent to stdout (already). In the below case, you want to send (at least) the play, the task and the failure to stderr, which is harder to do if you are just piping stdout to /dev/null. Ansible has no knowledge about the fact you are doing that. If you would have -q/--quiet, Ansible could modify its error-output to provide the context because we know it has not been printed before.

Just to indicate that only sending failures to stderr is not going to cut it IMO.

```
PLAY [This will fail on task 2] ********************* 

TASK: [command pwd] ********************* 
changed: [x220] => {"changed": true, "cmd": ["pwd"], "delta": "0:00:00.002359", "end": "2012-10-03 12:28:13.419084", "rc": 0, "start": "2012-10-03 12:28:13.416725", "stderr": "", "stdout": "/root"}
changed: [localhost] => {"changed": true, "cmd": ["pwd"], "delta": "0:00:00.002065", "end": "2012-10-03 12:28:13.418790", "rc": 0, "start": "2012-10-03 12:28:13.416725", "stderr": "", "stdout": "/root"}

TASK: [fail msg="Told you so"] ********************* 
skipping: [x220]
failed: [localhost] => {"failed": true, "msg": "Told you so", "rc": 1}
```

dagwieers:
In fact, now that I think of it, when doing -q/--quiet we probably also want to see the statistics at the end _only if there were errors_ so you can at least see what succeeded despite the failures.

jpmens:
Yes, both things sound sensible to me.

mpdehaan:
Here's what I'll accept.   They must all go together when submitted to
ensure completeness.

(A) Callback errors go to stderr.   (you won't get the task headers in cron
reports but ah well).

(B) We _ALWAYS_ output the summary, but depending on whether there are
failures, it goes to stdout or stderr as appropriate.

(C) There shall not be a quiet yet.  We can talk about this later, but I
[...]

PR #17962, opened 2016-10-10
Fixes #17958: use stderr instead of the default stdout for fatal errors

### LINKED ISSUE #17958: fatal errors are sent to stdout instead of stderr
##### ISSUE TYPE

- Bug Report
##### ANSIBLE VERSION

```
ansible 2.0.2.0
```
##### SUMMARY

Ansible reporting output sends fatal errors to stdout instead of stderr.
##### STEPS TO REPRODUCE

Just run a playbook or a command for which you expect a fatal error (e.g. host is down) and redirect stderr to a file. The error output will be visible on the console, but the file will be empty.

```
~# ansible unreachablehost.example.com -m command -a 'uptime' 2>errout
unreachablehost.example.com | UNREACHABLE! => {
    "changed": false, 
    "unreachable": true
}
~# cat errout
~#
```

##### EXPECTED RESULTS

I expected no output on the console and the file `errout` to contain the output instead.
##### ACTUAL RESULTS

Errors were output to the console and the `errout` file was empty

```

```

fredericve:
##### ISSUE TYPE

- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.0.2.0
```
##### SUMMARY

Fixes #17958 
The changes in this PR send any error output to stderr instead of the default stdout. This is useful for deployments where ansible is running e.g. from cron and you only want error output to be sent in an e-mail (redirect stdout to /dev/null)

Before change:

```
~# ansible-playbook /usr/local/share/ansible/playbook/site.yml --limit unreachablehost.example.com >/dev/null

PLAY [sops] ********************************************************************

TASK [pubkeys : include] *******************************************************
included: /usr/local/share/ansible/playbook/roles/pubkeys/tasks/pubkeys.yml for unreachablehost.example.com

TASK [pubkeys : Get user information] ***********************************
fatal: [unreachablehost.example.com]: UNREACHABLE! => {"changed": false, "msg": "Failed to connect to the host via ssh.", "unreachable": true}
    to retry, use: --limit @/usr/local/share/ansible/playbook/site.retry

unreachablehost.example.com : ok=1    changed=0    unreachable=1    failed=0   

```

After change:

```
~# ansible-playbook /usr/local/share/ansible/playbook/site.yml --limit unreachablehost.example.com >/dev/null
fatal: [unreachablehost.example.com]: UNREACHABLE! => {"changed": false, "msg": "Failed to connect to the host via ssh.", "unreachable": true}

```

### DIFF
--- a/lib/ansible/plugins/callback/stderr.py
+++ b/lib/ansible/plugins/callback/stderr.py
@@ -1,0 +1,81 @@
+# (c) 2017, Frederic Van Espen <github@freh.be>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+# Make coding more python3-ish
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+from ansible import constants as C
+from ansible.plugins.callback.default import CallbackModule as CallbackModule_default
+
+
+class CallbackModule(CallbackModule_default):
+
+    '''
+    This is the stderr callback plugin, which reuses the default
+    callback plugin but sends error output to stderr.
+    '''
+
+    CALLBACK_VERSION = 2.0
+    CALLBACK_TYPE = 'stdout'
+    CALLBACK_NAME = 'stderr'
+
+    def __init__(self):
+
+        self.super_ref = super(CallbackModule, self)
+        self.super_ref.__init__()
+
+    def v2_runner_on_failed(self, result, ignore_errors=False):
+
+        delegated_vars = result._result.get('_ansible_delegated_vars', None)
+        self._clean_results(result._result, result._task.action)
+
+        if self._play.strategy == 'free' and self._last_task_banner != result._task._uuid:
+            self._print_task_banner(result._task)
+
+        self._handle_exception(result._result, errors_to_stderr=True)
+        self._handle_warnings(result._result)
+
+        if result._task.loop and 'results' in result._result:
+            self._process_items(result)
+
+        else:
+            if delegated_vars:
+                self._display.display("fatal: [%s -> %s]: FAILED! => %s" % (result._host.get_name(), delegated_vars['ansible_host'],
+                                                                            self._dump_results(result._result)), color=C.COLOR_ERROR,
+                                                                            stderr=True)
+            else:
+                self._display.display("fatal: [%s]: FAILED! => %s" % (result._host.get_name(), self._dump_results(result._result)),
+                                      color=C.COLOR_ERROR, stderr=True)
+
+        if ignore_errors:
+            self._display.display("...ignoring", color=C.COLOR_SKIP)
+
+    def _handle_exception(self, result, errors_to_stderr=False):
+
+        if 'exception' in result:
+            msg = "An exception occurred during task execution. "
+            if self._display.verbosity < 3:
+                # extract just the actual error message from the exception text
+                error = result['exception'].strip().split('\n')[-1]
+                msg += "To see the full traceback, use -vvv. The error was: %s" % error
+            else:
+                msg = "The full traceback is:\n" + result['exception']
+                del result['exception']
+
+            self._display.display(msg, color=C.COLOR_ERROR, stderr=errors_to_stderr)
+
--- a/lib/ansible/plugins/callback/stderr.py
+++ b/lib/ansible/plugins/callback/stderr.py
@@ -57,7 +57,7 @@ class CallbackModule(CallbackModule_default):
             if delegated_vars:
                 self._display.display("fatal: [%s -> %s]: FAILED! => %s" % (result._host.get_name(), delegated_vars['ansible_host'],
                                                                             self._dump_results(result._result)), color=C.COLOR_ERROR,
-                                                                            stderr=True)
+                                      stderr=True)
             else:
                 self._display.display("fatal: [%s]: FAILED! => %s" % (result._host.get_name(), self._dump_results(result._result)),
                                       color=C.COLOR_ERROR, stderr=True)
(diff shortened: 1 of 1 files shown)


bcoca:
these are not fatal errors for the ansible-playbook, though they might be for individual tasks for that host.

fredericve:
Indeed, it is a fatal error for that host and playbook execution for that host will stop at that point. That still makes it a fatal error. The existing code also uses the term _fatal_.

bcoca:
So what i meant to say, is that ansible outputs to stderr when there is a fatal error with ansible itself, not when it captures an error with a task or a host.

Most people don't want it to otuput to stderr for each intermediate error that might happen with a task or host, so I'm not partial to update the default callback this way. 

If you want to change this and submit it as an alternate stdout plugin we can easily add that to Ansible.

fredericve:
@bcoca OK, got that. I pushed the changes as an alternate stdout plugin. Is this an acceptable change?

bcoca:
wouldn't you need to output the task for which teh error happened to stderr also? the output will be confusing as you can have errors across several tasks.

fredericve:
I would consider that an improvement indeed, but isn't really required for us. Playbook execution stops anyway when it fails.

I can have a look into adding it, but I'm not really familiar with ansible, nor python. Is it a requirement to get it merged?

bcoca:
If anyone is using ignore_errors or not failing at single host in play, this won't be very useful. 

We merge features when we see them as useful for more than just one case/party.

fredericve:
My last comment wasn't accurate. I'm not stopping the full playbook execution when one host fails. What I meant to say was that by default, the rest of the playbook is not executed for that specific host when a task fails. This can benefit anyone who only wants to see error messages.

For ignore_errors indeed this won't be very useful indeed but I imagine such a person would then not be using this callback plugin in the first place.

Note that others will likely find this useful as well. I also found issue #1210 which suggested implementing the same.

bcoca:
you still would not have any indication on which task the host failed on

fredericve:
Yes indeed. That was my question. Is that a requirement to get it merged?

bcoca:
I think its a requirement for it being usable for anyone other than you

fredericve:
I don't agree. I'm sure I'm not the only one running ansible from cron (see issue #1210) where you only want a notification email when errors occurred. At least you can see which host wasn't updated properly, manually run the playbook and act on it.

People are using workarounds in order to hide the output when everything went well.

bcoca:
most people will also want to know WHICH task failed, not just host

jtognazzi:
We are using ansible for deployment via atlassian Bamboo, and sending fatal error on stderr would make Bamboo writes these error in red. I should test though to see how usefull it is...

wyaeld:
the change discussed would be useful for me too, same use-case as @fredericve 

essentially I'm using ansible in this case to run scheduled operation tasks (backups, cleanups, reports etc) from cron, and don't want the output hidden if there are failures

bcoca:
This needs a rebase, other than that I'm fine with the stderr plugin, but we don't want the changes to the default one.

alikins:
@bcoca 

> This needs a rebase, other than that I'm fine with the stderr plugin, but we don't want the changes to the default one.

What are the issues with the changes to the default callback? It only adds an optional arg and doesnt change the default behavior.

bcoca:
that parameter would have to be hardcoded in tqm, makes no sense to me to add it.

alikins:
@bcoca 
> that parameter would have to be hardcoded in tqm, makes no sense to me to add it.

Not sure I understand. What part of the pr requires a tqm change? AFAICT, only the stderr callback plugin would call the v2_runner_on_failed() with the error_to_stderr parameter.

Does the current pr not work?

bcoca:
@alikins that is a standard event, we should not change the signature as it is part of the public API

fredericve:
@bcoca 

I'm happy to rebase and fix the pull request, but how do you suggest I fix it? Initially I had extended the v2_runner_on_failed() function to avoid code duplication and to maintain standard callback behaviour on all other levels.

Should I just duplicate code now in the stderr plugin?

bcoca:
or you can use default as your base object (see skippy callback).

fredericve:
@bcoca 

That's what the pull request already does. It uses default as the base object. The issue is that in order to send to stderr, the ```stderr``` parameter needs to be set to ```True``` when calling _self._display.display(). That can only be done in the object calling the routine. That's why I added an additional parameter and maintained default behaviour.

Apart from duplicating chunks of code inside the stderr callback plugin I don't see how I can make this work, but then again, duplicating code likely means broken code if it is not maintained.

I'll already rebase so it can fit the current state of the devel branch and we'll see where we go from there.

fredericve:
@bcoca 

Jesus, messed up the rebase. Should I create a new, clean PR?

bcoca:
just force push a clean version, the PR will reset

fredericve:
@bcoca the PR is clean again.

Let me know if this works for you, or if I should reimplement by diverging from the default callback plugin.

bcoca:
@fredericve i normally want to avoid new callbacks having to change the base class or the default one. I see that the changes are minimal here and would require a lot of duplicate code to do otherwise, but i'm still on the fence on this one. I'll wait for more committers to weigh in on it as I don't want to be the sole decider on this one..

abadger:
We talked about this in a small group of core committers today and we think that we need to have it go through the proposal process and be discussed at the public IRC meetings.  (Open a proposal here: https://github.com/ansible/proposals/issues/  Put it on the meeting agenda here: https://github.com/ansible/community/issues/162 ).  Here's the thoughts we had in our smaller group:

* In the past we've discussed and approved an idea of having callback plugins "own" stdout and the rest of ansible "own" stderr.  The aim there was for callbacks like the json callback plugin would know that it wouldn't have conflicts on stdout with random messages from ansible proper.  That doesn't conflict necessarily with this (as that use case was more about guaranteeing a space for the callback plugin than guaranteeing a space for the other parts of ansible) but it does seem a little odd conceptually.

* Another way to achieve the goal of being able to cron ansible and only getting email on an error would be to combine the above with a callback plugin that only wrote to stdout on errors.  Probably want to combine this with a --quiet flag so that ansible (noncallback) only displays something in case of errors.

* Doing things the way this proposaes could lead to other uses by different types of callback plugins but without those use cases we'd want to make this more generic:
  * We'd probably wantto make it applicable to all functions, not just the error callback.
  * We'd probably want to be able to use other streams, not just a boolean to switch between stdout and stderr.

All of these ideas together need to be sorted out and made into a coherent whole picture which is why a proposal and discussion in the irc meetings is appropriate for this feature.

abadger:
@fredericve  At today's public IRC meeting we decided that a callback using stderr is fine but we would like it to be done entirely in the callback instead of changing the callback API.  If you rework the code that way, we'll be able to merge this.

fredericve:
@abadger Thank you for considering. As requested, I reworked the code to have the changes fully inside the callback plugin. Basically this duplicates some of the code in the default callback plugin and in the base object, with only changes to it to send to stderr in the display() calls.

abadger:
Merged to devel.  Thanks!

fredericve:
Cool. Thanks!

PR #17965, opened 2016-10-11
Ensure vars_files doesn't bomb out in delegated tasks when inventory_hostname is used in the variable

willthames:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

vars_files
##### ANSIBLE VERSION

```
ansible 2.3.0 (devel 20aedf90e7) last updated 2016/10/11 12:19:58 (GMT +1000)
  lib/ansible/modules/core: (detached HEAD b4f6a25195) last updated 2016/10/10 13:34:50 (GMT +1000)
  lib/ansible/modules/extras: (detached HEAD 734aa8f8e9) last updated 2016/10/10 13:34:54 (GMT +1000)
  config file = 
```
##### SUMMARY

If `vars_files` depends on `inventory_hostname`, the `vars_files` can fail to load when delegating to another host (this is during the call to `_get_delegated_vars`)

For example, if secrets/testhost/vault.yml exists but secrets/localhost/vault.yml does not, the second debug task fails:

```
- hosts: testhost
  vars_files: secrets/{{inventory_hostname}}/vault.yml
  tasks:
  - debug:
     msg: this works
  - debug:
     msg: this fails
    delegate_to: localhost
```

Additionally, there are a number of reasons why `vars_files` might not be loadable at the current point in time.

Attempt to print out the undefined variables and the paths attempted.

### DIFF
--- a/lib/ansible/vars/__init__.py
+++ b/lib/ansible/vars/__init__.py
@@ -298,6 +298,7 @@ class VariableManager:
                 # as soon as we read one from the list. If none are found, we
                 # raise an error, which is silently ignored at this point.
                 try:
+                    files_not_found = list()
                     for vars_file in vars_file_list:
                         vars_file = templar.template(vars_file)
                         try:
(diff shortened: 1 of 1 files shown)


willthames:
I still believe this is a bug, but there is a workaround - rather than use `delegate_to: localhost`, you can use `connection: local`. (This is only useful if the delegation is to localhost, rather than another target)

bcoca:
I'm not sure what behavior we want here, have the original host's vars files loaded seems the correct one to me, but that might not always be the case.

Adding needs_info to make sure more committers weigh in, I'm assuming @willthames just wants to turn this into a warning (when delegated) as per the code submitted

jimi-c:
# Closing This Ticket

Hi!

We believe the above commit should resolve this problem for you. This will also be included in the next release.

Thank you!

PR #17970, opened 2016-10-11
Fix unbound method call for JSONEncoder

rochacbruno:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

JSONEncoder
##### SUMMARY

The way it is currently it will lead to unbound method error

``` python
In [1]: import json

In [2]: json.JSONEncoder.default('object_here')
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
----> 1 json.JSONEncoder.default('object_here')

TypeError: unbound method default() must be called with JSONEncoder instance as first argument (got str instance instead)

```

But what is really wanted is to let the json module to raise the "is not serializable error" which demands a bounded instance of `JSONEncoder()`

``` python
In [3]: json.JSONEncoder().default('object_here')
---------------------------------------------------------------------------
TypeError: 'object_here' is not JSON serializable 

```

BTW: I think it would try to call `.to_json` of object before raising as it is a common pattern.

### DIFF
--- a/lib/ansible/plugins/filter/core.py
+++ b/lib/ansible/plugins/filter/core.py
@@ -67,7 +67,7 @@ class AnsibleJSONEncoder(json.JSONEncoder):
         if isinstance(o, HostVars):
             return dict(o)
         else:
-            return json.JSONEncoder.default(o)
+            return json.JSONEncoder().default(o)
 
 def to_yaml(a, *args, **kw):
     '''Make verbose, human readable yaml'''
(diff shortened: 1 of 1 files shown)


abadger:
Thanks.  This is a method call to the superclass's method.  So it should probably either call super() to get the bound version of the method or pass self to the method explicitly.  I prefer super so that in the unlikely case we use multi-inheritance with this it will just work.  So the change should look like this:

``` python
   return super(AnsibleJSONEncoder, self).default(o)
```

rochacbruno:
@abadger done!

abadger:
@rochacbruno Thanks!  Merged to devel, stable-2.2, and stable-2.1.  The next release we're probably making will be 2.2.0 and the fix will be present there.  If we make another 2.1.x release, the fix will also be in that.

PR #17974, opened 2016-10-11
cloudstack: docs: use local_action, not connection=local

resmo:
##### ISSUE TYPE

- Docs Pull Request
##### COMPONENT NAME

cloudstack guide
##### ANSIBLE VERSION

```
2.3
```
##### SUMMARY

use local_action in examples as connection=local does not always work (e.g. if python3 is the default python)

### DIFF
--- a/docsite/rst/guide_cloudstack.rst
+++ b/docsite/rst/guide_cloudstack.rst
@@ -78,7 +78,8 @@ By passing the argument ``api_region`` with the CloudStack modules, the region w
 .. code-block:: yaml
 
     - name: ensure my ssh public key exists on Exoscale
-      local_action: cs_sshkeypair
+      local_action:
+        module: cs_sshkeypair
         name: my-ssh-key
         public_key: "{{ lookup('file', '~/.ssh/id_rsa.pub') }}"
         api_region: exoscale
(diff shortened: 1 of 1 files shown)


bcoca:
docs should not be fixed, connection=local should ALWAYS work, if not, it is a bug.

resmo:
I should have extended it with ".. as expected in some situations"

Correct me if I am wrong, but one difference of connection=local vs local_action is that connection=local uses /usr/bin/python where local_action uses the python in your PATH. This means you get slightly more control about what python is going to be used on your local machine for local tasks.

Having a team with different OS and setups (macOS, Ubuntu, Debian) where /usr/bin/python may or may not be python3 or python2, this can be much easier to be handled by the team member individually and for all ansible projects globally than using ansible_python_interpreter. 

Further, I steamlined all my playbooks for that reason to local_action and would also cover this in the docs.

bcoca:
No, both should use 'sys.executable' which should reflect the python chosen on install for bin/ansible. 

As they rely on the implicit localhost, if localhost is defined in the inventory that will change the behavior.

resmo:
hmm, it seems I was confused with sourcing ./hacking/env-setup which uses the python in my PATH (which is ~/bin/python) where ansible pip installed uses /usr/bin/python, but for connection=local and local_action. 

Never mind.

PR #17975, opened 2016-10-11
Making the ANSI escape regexes configurable in the Shell class in the…

woody77:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

Shell and CliBase
##### ANSIBLE VERSION

```
ansible 2.3.0 (devel 0641225e74) last updated 2016/10/11 14:58:00 (GMT -700)
  lib/ansible/modules/core: (detached HEAD b4f6a25195) last updated 2016/10/11 11:18:35 (GMT -700)
  lib/ansible/modules/extras: (detached HEAD 734aa8f8e9) last updated 2016/10/11 11:18:40 (GMT -700)
  config file = 
```
##### SUMMARY

In trying to use ansible with a network appliance similar to a cisco switch, I ran into an issue where the ios_command "show version" would hang.  After debugging, I determined that the cause was unusual ANSI escape sequences in the appliance's shell.  These weren't being filtered out by the ANSI_RE regexes in module_utils/shell.py.

To facilitate these sorts of devices, this change lets the Cli override the ANSI_RE regexes using the same code pattern that it uses to provide the prompts and errors regexes.

### DIFF
--- a/lib/ansible/module_utils/shell.py
+++ b/lib/ansible/module_utils/shell.py
@@ -37,6 +37,7 @@ except ImportError:
 from ansible.module_utils.basic import get_exception
 from ansible.module_utils.network import NetworkError
 
+# Default ANSI and terminal escape sequence regexes
 ANSI_RE = [
     re.compile(r'(\x1b\[\?1h\x1b=)'),
     re.compile(r'\x08.')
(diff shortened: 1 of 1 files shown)


woody77:
Created the pull request from the wrong branch, ignore this.

PR #17986, opened 2016-10-12
iocage connector that wraps around jail connector.

slohse:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

iocage connection plugin
##### ANSIBLE VERSION

```
devel
```
##### SUMMARY

This change is a new connector for FreeBSD jails that are managed by iocage. It wraps around the existing jail connector and only translates the tag or UUID of an iocage-managed jail to the underlying actual name of the jail. It uses the `iocage` command to do the translation, so besides tags one can also use a partial UUID like with iocage itself.

_What is iocage and why does it need its own connector_? Iocage is a popular tool that simplifies creating and managing FreeBSD jails. But iocage uses UUIDs for the name property of a jail and provides its own "tag" property for legible names. So while the jails created by iocage can be managed in ansible with the existing jail connector, one would have to use the jail _name_ as remote host with the result that inventories and ansible run outputs would only contain hostnames such as `ioc-b39796cc-8f0d-11e6-8d6b-001d9263409c`.

### DIFF
--- a/lib/ansible/plugins/connection/iocage.py
+++ b/lib/ansible/plugins/connection/iocage.py
@@ -19,13 +19,11 @@
 from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
-import distutils.spawn
 import os
 import subprocess
-from jail import Connection as jail
+from ansible.plugins.connection.jail import Connection as Jail
 
 from ansible.errors import AnsibleError
-from ansible.plugins.connection import ConnectionBase, BUFSIZE
 
 try:
     from __main__ import display
@@ -34,30 +32,24 @@ except ImportError:
     display = Display()
 
 
-class Connection(ConnectionBase):
+class Connection(Jail):
     ''' Local iocage based connections '''
 
     transport = 'iocage'
-    has_pipelining = jail.has_pipelining
-    become_methods = jail.become_methods
 
     def __init__(self, play_context, new_stdin, *args, **kwargs):
-        super(Connection, self).__init__(play_context, new_stdin, *args, **kwargs)
+        self.ioc_jail = play_context.remote_addr
 
-        self.ioc_jail = self._play_context.remote_addr
-
-        if os.geteuid() != 0:
-            raise AnsibleError("iocage/jail connection requires running as root")
-
-        self.iocage_cmd = jail._search_executable('iocage')
+        self.iocage_cmd = Jail._search_executable('iocage')
 
         jail_uuid = self.get_jail_uuid()
 
-        play_context.remote_addr = 'ioc-{}'.format(jail_uuid)
+        kwargs[Jail.modified_jailname_key] = 'ioc-{}'.format(jail_uuid)
 
-        display.vvv(u"Jail {iocjail} has been translated to {rawjail}".format(iocjail = self.ioc_jail, rawjail = play_context.remote_addr))
-        
-        self.jail_connector = jail(play_context, new_stdin, *args, **kwargs)
+        display.vvv(u"Jail {iocjail} has been translated to {rawjail}".format(
+            iocjail=self.ioc_jail, rawjail=kwargs[Jail.modified_jailname_key]))
+
+        super(Connection, self).__init__(play_context, new_stdin, *args, **kwargs)
 
     def get_jail_uuid(self):
         p = subprocess.Popen([self.iocage_cmd, 'get', 'host_hostuuid', self.ioc_jail],
@@ -73,21 +65,4 @@ class Connection(ConnectionBase):
             raise AnsibleError(u"iocage returned an error: {}".format(stdout))
 
         return stdout.strip('\n')
-    
-    def _connect(self):
-        self.jail_connector.connect()
-        self._connected = self.jail_connector._connected
-
-    def exec_command(self, cmd, in_data=None, sudoable=True):
-        return self.jail_connector.exec_command(cmd, in_data, sudoable)
-
-    def put_file(self, in_path, out_path):
-        return self.jail_connector.put_file(in_path, out_path)
-
-    def fetch_file(self, in_path, out_path):
-        return self.jail_connector.fetch_file(in_path, out_path)
-
-    def close(self):
-        self.jail_connector.close()
-        self._connected = False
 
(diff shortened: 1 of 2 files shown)


mattclay:
@slohse The [tests](https://app.shippable.com/runs/584f22919559830f0009086c/1/console) are failing due to missing boilerplate in `lib/ansible/plugins/connection/iocage.py`:

```
2016-12-12 23:03:35 Run command: test/sanity/code-smell/boilerplate.sh
2016-12-12 23:03:35 
2016-12-12 23:03:35 == Missing __metaclass__ = type ==
2016-12-12 23:03:35 ./lib/ansible/plugins/connection/iocage.py
2016-12-12 23:03:35 
2016-12-12 23:03:35 == Missing from __future__ import (absolute_import, division, print_function) ==
2016-12-12 23:03:35 ./lib/ansible/plugins/connection/iocage.py
2016-12-12 23:03:35 ERROR: Command "test/sanity/code-smell/boilerplate.sh" returned exit status 2.
```

Here's an example of the needed boilerplate:

https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/__init__.py#L20

slohse:
Thanks for the hint @mattclay :)
Still failing though :/

mattclay:
@slohse The failure was due to an unrelated CI issue which has been fixed. Tests are now passing.

slohse:
Excellent, thank you!

mattclay:
@slohse This definitely looks useful. I'm particularly interested in this as it should make it easier to add integration test coverage for `jail` to our existing CI process.

I discussed this briefly with @abadger and we think it would be more appropriate to inherit from the existing `jail` connection plugin instead of wrapping it. You'll need to make a few minor changes to the existing plugin to permit setting the jail name without changing `self._play_context.remote_addr`.

slohse:
Okay, I modified `jail` now in a way that it accepts a custom jailname via `kwargs`. I'm not entirely convinced I found the most elegant solution, though.

lib/ansible/plugins/connection/jail.py
```diff
@@ -54,6 +56,8 @@ def __init__(self, play_context, new_stdin, *args, **kwargs):
         super(Connection, self).__init__(play_context, new_stdin, *args, **kwargs)
 
         self.jail = self._play_context.remote_addr
+        if(self.modified_jailname_key in kwargs):
```
mattclay: Remove the parenthesis from the conditional, as they're not needed.

lib/ansible/plugins/connection/iocage.py
```diff
@@ -0,0 +1,68 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import os
```
mattclay: Unused import.

lib/ansible/plugins/connection/iocage.py
```diff
@@ -0,0 +1,68 @@
...
+    def __init__(self, play_context, new_stdin, *args, **kwargs):
+        self.ioc_jail = play_context.remote_addr
+
+        self.iocage_cmd = Jail._search_executable('iocage')
+
+        jail_uuid = self.get_jail_uuid()
+
+        kwargs[Jail.modified_jailname_key] = 'ioc-{}'.format(jail_uuid)
+
+        display.vvv(u"Jail {iocjail} has been translated to {rawjail}".format(
```
mattclay, line 48: Include a `host=kwargs[Jail.modified_jailname_key]` kwarg when calling `display.vvv`.

This will maintain consistency with other calls to `display.vvv` made by the `Jail` connection plugin.

mattclay (changes requested):
This looks really good, except for a few minor issues.

Thanks for the prompt response to feedback.

lib/ansible/plugins/connection/iocage.py
slohse: Wouldn't it be helpful in terms of error tracing to use the initial iocage tag (i.e. `self.ioc_jail`) instead of the already converted jail uuid for `host`?

lib/ansible/plugins/connection/jail.py
slohse: Whoops, my non-pythonic background shows :D

lib/ansible/plugins/connection/iocage.py
slohse: Right you are

lib/ansible/plugins/connection/iocage.py
mattclay: That would be ideal, as long as you maintain the log entry you already have to indicate the mapping between the two.

You'll need to update the `Jail` plugin so there's a separate `self.jail_friendly_name` or something similar which is used for the `host` passed to `display.vvv`, so you're able to show the iocage jail name on all the messages from the underlying `Jail` plugin.

slohse: Hm, I meant in that particular place in `iocage.py` only. I think trying to pass that down to `jail.py` could get confusing quickly and kind of blurs the edges of the two connectors.

mattclay: It's important to have the same `host` kwarg value used for all calls to `display.vvv` for the same connection, whether that be the actual jail name or the iocage jail name.

As written currently, the `display.vvv` call above is simply lacking the `host` kwarg. It can be added without making any changes to the message displayed.

slohse: Oh, okay. If it has to be the same throughout the connection, then I guess using the actual jail name is the most sensible approach

mattclay (approved):

mattclay:
@slohse Thanks for the connection plugin. I'll take a look at using this to add iocage and jail to CI.

slohse:
Cool, thank you :)

PR #17987, opened 2016-10-12
Change `v2_playbook_on_start` logic to positively detect legacy plugins

stevekuznetsov:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

`executor.TaskQueueManager`
##### ANSIBLE VERSION

```
$ ansible --version
ansible 2.2.0.0
$ git log --oneline -n 2
b2efbd1 Change `v2_playbook_on_start` logic to positively detect legacy plugins
6be09ee Merge pull request #17973 from cgwalters/docs-playbook-loop
```
##### SUMMARY

In order to support legacy plugins, the following two method signatures
are allowed for `CallbackBase.v2_playbook_on_start`:

```
def v2_playbook_on_start(self):
def v2_playbook_on_start(self, playbook):
```

Previously, the logic to handle this divergence checked to see if the
callback plugin being called supported an argument named `playbook`
in its `v2_playbook_on_start` method. This was fragile in a few ways:
- if a plugin author did not use the literal `playbook` to name their
  method argument, their plugin would not be called correctly
- if a plugin author wrapped their `v2_playbook_on_start` method and
  by doing so changed the argspec to no longer expose an argument
  with that literal name, their plugin would not be called correctly

In order to continue to support both types of callback for backwards
compatibility while making the call more robust for plugin authors,
the logic can be reversed in order to have a positive check for the old
method signature instead of a positive check for the new one.

Signed-off-by: Steve Kuznetsov skuznets@redhat.com

### DIFF
--- a/lib/ansible/executor/task_queue_manager.py
+++ b/lib/ansible/executor/task_queue_manager.py
@@ -353,17 +353,20 @@ class TaskQueueManager:
 
             for method in methods:
                 try:
-                    # temporary hack, required due to a change in the callback API, so
-                    # we don't break backwards compatibility with callbacks which were
-                    # designed to use the original API
+                    # Previously, the `v2_playbook_on_start` callback API did not accept
+                    # any arguments. In recent versions of the v2 callback API, the play-
+                    # book that started execution is given. In order to support both of
+                    # these method signatures, we need to use this `inspect` hack to send
+                    # no arguments to the methods that don't accept them. This way, we can
+                    # not break backwards compatibility until that API is deprecated.
                     # FIXME: target for removal and revert to the original code here after a year (2017-01-14)
                     if method_name == 'v2_playbook_on_start':
                         import inspect
-                        (f_args, f_varargs, f_keywords, f_defaults) = inspect.getargspec(method)
-                        if 'playbook' in f_args:
-                            method(*args, **kwargs)
-                        else:
+                        argspec = inspect.getargspec(method)
+                        if argspec.args == ['self']:
                             method()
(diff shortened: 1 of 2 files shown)


stevekuznetsov:
Had some `virtualenv` weirdness with the tests, working on fixing that now.

stevekuznetsov:
ping @abadger @jimi-c @sivel

stevekuznetsov:
ping @abadger @jimi-c @sivel

 The test failure was due to a Fedora repository flake: `Failed to synchronize cache for repo 'updates'`:

```
2016-10-25 00:58:54 TASK [yum : install sos] *******************************************************
2016-10-25 00:59:09 fatal: [testhost]: FAILED! => {"changed": true, "failed": true, "msg": "Yum command has been deprecated, redirecting to '/usr/bin/dnf -d 2 -y install sos'.\nSee 'man dnf' and 'man yum2dnf' for more information.\nTo transfer transaction metadata from yum to DNF, run:\n'dnf install python-dnf-plugins-extras-migrate && dnf-2 migrate'\n\nError: Failed to synchronize cache for repo 'updates'\n", "rc": 1, "results": [""]}
```

abadger:
This looks good to me.  I'm going to merge to devel shortly.  Trying to attract another set of eye to see if it should be cherry-picked for 2.2.1 as well.

abadger:
Merged to devel and cherry-picked to temp-post-2.2.0-staging.  If all goes well it will be in 2.2.1.

PR #17989, opened 2016-10-12
Making the ANSI escape regexes configurable in the Shell class

woody77:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

Shell
##### ANSIBLE VERSION

```
ansible 2.3.0 (shell_ansi_escape_fix a616dcbe1a) last updated 2016/10/12 11:54:17 (GMT -700)
  lib/ansible/modules/core: (detached HEAD b4f6a25195) last updated 2016/10/11 11:18:35 (GMT -700)
  lib/ansible/modules/extras: (detached HEAD 734aa8f8e9) last updated 2016/10/11 11:18:40 (GMT -700)
  config file = 
```
##### SUMMARY

The module_utils Shell class uses a hardcoded set of regexes to filter out ansi color sequences and terminal escape codes.  For using the Shell with some network appliances, this set of regexes needs to be overridden (or expanded).

This change allows a CliBase subclass to override the ansi escape regexes in the same manner that it uses to provide the prompt and errors with CLI_PROMPTS_RE and CLI_ERRORS_RE.  This checks for a CLI_ANSI_RE, and if provided, uses that instead of the default ANSI_RE for sanitizing lines read from the remote shell.

### DIFF
--- a/lib/ansible/module_utils/shell.py
+++ b/lib/ansible/module_utils/shell.py
@@ -37,6 +37,7 @@ except ImportError:
 from ansible.module_utils.basic import get_exception
 from ansible.module_utils.network import NetworkError
 
+# Default ANSI and terminal escape sequence regexes
 ANSI_RE = [
     re.compile(r'(\x1b\[\?1h\x1b=)'),
     re.compile(r'\x08.')
@@ -61,7 +62,7 @@ class ShellError(Exception):
 
 class Shell(object):
 
-    def __init__(self, prompts_re=None, errors_re=None, kickstart=True):
+    def __init__(self, prompts_re=None, errors_re=None, ansi_re=None, kickstart=True):
         self.ssh = None
         self.shell = None
 
@@ -70,6 +71,7 @@ class Shell(object):
 
         self.prompts = prompts_re or list()
         self.errors = errors_re or list()
+        self.ansi = ansi_re or ANSI_RE
 
     def open(self, host, port=22, username=None, password=None, timeout=10,
              key_filename=None, pkey=None, look_for_keys=None,
@@ -118,7 +120,7 @@ class Shell(object):
         self.receive()
 
     def strip(self, data):
-        for regex in ANSI_RE:
+        for regex in self.ansi:
             data = regex.sub('', data)
         return data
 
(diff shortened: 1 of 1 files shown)


woody77:
The test failures seem to be a problem with the CI system:

```
2016-10-12 19:04:32 Pulling repository docker.io/ansible/ansible
2016-10-12 19:04:32 Tag opensuseleap not found in repository docker.io/ansible/ansible
```

```
2016-10-12 19:03:31 Pulling repository docker.io/ansible/ansible
2016-10-12 19:03:31 Tag centos7 not found in repository docker.io/ansible/ansible
```

abadger:
We took a look at this among a small group in today's meeting and weren't sure if this was the right API for doing ths.  We thought that perhaps overriding the split method or passing a method into the constructor rather than a regex might make more sense.  But we were limited in evaluating what would be appropriate by not knowing how this is intended to be used.  We have here the ability to override the Shell() constructor but nothing that's calling the constructor with overridden values.  So we're not seeing the complete picture and knowing what the API should look like to accommodate that.

Could you give us some more information about how this would be called?

needs_info

woody77:
The background on this was that I was using ansible with a network appliance which used a series of escape codes that were subtly different from those used by say a Cisco router.  So I was calling the Shell() constructor with a different list of regexes (one was the same, but I had to add about 5-6 more to get Shell() to properly parse the output of the particular appliance I was working with).

The simplest method that I could find was to add the ability to provide a different set of regexes when instantiating the Shell class, so that it could be customized for communicating with different appliances.

IIRC, overriding split() looked like it was workable, but heavy-handed since the method performed all the right steps, it just needed a larger set of regexes to deal with some quirks in how the appliance returned prompts and did pagination.

OTOH, passing in a function that returned a list of appropriate regexes could also work.  That, to me, feels like a stylistic choice, but retains the overall same functionality.

I don't have access to the full source that I was using at the time (this was for a project at a previous employer), so I'm going somewhat off of memory on this.

gundalow:
bot_status

bcoca:
Thank you very much for your interest in Ansible. 

This plugin is no longer maintained in this repository and has been migrated to  https://github.com/ansible-collections/community.network

Please re-submit this issue in the above repository. If you have further questions please stop by IRC or the mailing list: 

IRC: #ansible on irc.freenode.net 
mailing list: https://groups.google.com/forum/#!forum/ansible-project

PR #18016, opened 2016-10-13
Minor cleanup/speedup to hacking/env-setup

bdowling:
##### ISSUE TYPE

- Feature Pull Request
- Docs Pull Request
##### COMPONENT NAME

hacking/env-setup
##### ANSIBLE VERSION

```
ansible 2.3.0 (fix_env-setup 26c4d6065c) last updated 2016/10/13 16:50:09 (GMT -400)
  lib/ansible/modules/core: (detached HEAD 275fa3f055) last updated 2016/10/13 14:12:11 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 3e1ea76a75) last updated 2016/10/13 14:12:11 (GMT -400)
```
##### SUMMARY

Made it faster to source the hacking/env-setup by allowing gen_egg_info and removal of .pyc files to be run in the background.   I went down a path to make gen_egg dependent on file-timestamps as well, but backed out of that option.

```
Nothing to see here.
```

Allow egg-info creation (and .pyc removal) to happen in background when -q is
used to speedup sourcing hacking/env-setup.  Couple other minor fixes:
chdir back to parent was unnecessary due to already in subshell,
.\* on expr matches was extraneous.

Added doc info to make env-setup a bit more static by adding a source command
to your profile

### DIFF
--- a/docsite/rst/intro_installation.rst
+++ b/docsite/rst/intro_installation.rst
@@ -144,6 +144,10 @@ If you want to suppress spurious warnings/errors, use::
 
     $ source ./hacking/env-setup -q
 
+If you want to make ansible always avaialble in your environment, add it to your shell's profile (assumes you already source'd this above to set $ANSIBLE_HOME)::
+
+    echo source $ANSIBLE_HOME/hacking/env-setup -q >> ~/.bash_profile
+
 If you don't have pip installed in your version of Python, install pip::
 
     $ sudo easy_install pip
--- a/docsite/rst/intro_installation.rst
+++ b/docsite/rst/intro_installation.rst
@@ -140,13 +140,13 @@ Using Fish::
 
     $ . ./hacking/env-setup.fish
 
-If you want to suppress spurious warnings/errors, use::
+You can use *-q* if you want to suppress spurious warnings/errors, and *-b* to background a few longer-running tasks so it returns faster for interactive or .profile use.::
 
     $ source ./hacking/env-setup -q
 
 If you want to make ansible always avaialble in your environment, add it to your shell's profile (assumes you already source'd this above to set $ANSIBLE_HOME)::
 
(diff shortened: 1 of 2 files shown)


robinro:
@bdowling I like the idea, but this might break a setup that uses env-setup in some other script, since the files are not delted at the end of the command.

Better put `wait` at the end. At least things run in parallel then.

bdowling:
I had thought that might be a response to this, so we're thinking alike, that's good. ;)

What if I were to add another flag to optionally background?  Then you could pass -b -q for example?

This is also comes back to user's comfort level -- e.g. sourcing a file updated from github in .profile's vs just add the *PATH vars to their profile statically, the only thing they might be missing is the updated egg_info at times..

robinro:
`-b` sounds good to me

docsite/rst/intro_installation.rst
```diff
@@ -140,10 +140,14 @@ Using Fish::
 
     $ . ./hacking/env-setup.fish
 
-If you want to suppress spurious warnings/errors, use::
+You can use *-q* if you want to suppress spurious warnings/errors, and *-b* to background a few longer-running tasks so it returns faster for interactive or .profile use.::
 
     $ source ./hacking/env-setup -q
 
+If you want to make ansible always avaialble in your environment, add it to your shell's profile (assumes you already source'd this above to set $ANSIBLE_HOME)::
```
mscherer: 'avaialble' is a typo.

samdoran:
Thank you very much for your submission to Ansible.

Unfortunately, we're not sure if we want this feature in the program. Our reasons for this are:

   * This adds a lot of complexity without a lot of benefit
   * Optimization of this script isn't _really_ needed.

Thank you once again for this and your interest in Ansible.

PR #18034, opened 2016-10-14
Remove pbrun's '-b' option.

qbit:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
ansible 2.3.0
```
##### COMPONENT NAME

```
ansible
```
##### SUMMARY

This PR removes the '-b' option. It is still possible to pass it by setting `become_flags`.

I am not 100% sure why but `assword:` stopped working (Timeout waiting for esc prompt) with the rest of the changes.. but `Password:` works. Would anyone be able to shed some light on that? I left it as an extra commit so it can be cherry picked out if needed.

### DIFF
--- a/lib/ansible/playbook/play_context.py
+++ b/lib/ansible/playbook/play_context.py
@@ -517,8 +517,8 @@ class PlayContext(Base):
 
             elif self.become_method == 'pbrun':
 
-                prompt='assword:'
-                becomecmd = '%s -b %s -u %s %s' % (exe, flags, self.become_user, success_cmd)
+                prompt='Password:'
+                becomecmd = '%s %s -u %s %s' % (exe, flags, self.become_user, success_cmd)
 
             elif self.become_method == 'ksu':
                 def detect_ksu_prompt(data):


lib/ansible/constants.py
```diff
@@ -235,6 +235,7 @@ def load_config_file():
 DEFAULT_SUDO_EXE          = get_config(p, DEFAULTS, 'sudo_exe', 'ANSIBLE_SUDO_EXE', None)
 DEFAULT_SUDO_FLAGS        = get_config(p, DEFAULTS, 'sudo_flags', 'ANSIBLE_SUDO_FLAGS', '-H -S -n')
 DEFAULT_ASK_SUDO_PASS     = get_config(p, DEFAULTS, 'ask_sudo_pass',    'ANSIBLE_ASK_SUDO_PASS',    False, value_type='boolean')
+DEFAULT_PBRUN_FLAGS       = get_config(p, DEFAULTS, 'pbrun_flags', 'ANSIBLE_PBRUN_FLAGS', '-b')
```
bcoca: we are moving away from 'per method flags' these should be in 'become flags'

lib/ansible/playbook/play_context.py
```diff
@@ -517,8 +518,8 @@ def detect_su_prompt(data):
 
             elif self.become_method == 'pbrun':
 
-                prompt='assword:'
-                becomecmd = '%s -b %s -u %s %s' % (exe, flags, self.become_user, success_cmd)
+                prompt='Password:'
```
bcoca, line 520: -b should appear if no password is supplied, we can make conditional like with sudo, but it is there to avoid prompts, when ansible is not expecting them

bcoca:
not sure about the differences with the password prompt, since i don't have access to pbrun to test I have to rely on user and contributor feedback.

lib/ansible/constants.py
qbit: OK, I will remove this.

lib/ansible/playbook/play_context.py
qbit: Sounds good.

qbit:
I am getting a test instance together so I can test -b changes.. I will also update the diff here shortly.

qbit:
@bcoca updated.

PR #18037, opened 2016-10-14
Fix uninitialized variable in deserialize method

sengaya:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

playbook/block.py
##### ANSIBLE VERSION

git devel (probably since 2016-08-01)
##### SUMMARY

Fix uninitialized variable in deserialize method

The bug was introduced with commit 06d4f4ad0e8f62c6b02b01833b94a0e44be008d0.
Added a simple test.


sengaya:
The tests failed because of some (temporary?) connection problem in "yum : install sos" in build_ci. Can the test somehow restarted?

PR #3617, opened 2013-07-22
svr4pkg doesn't work on Sol10

dermute:
--- This has already been written in issue ansible/ansible#3547

While removing a package from Solaris 10, I always receive the following error:

```
rc, out, err = module.run_command(' '.join(cmd), shell=False)
TypeError: run_command() got an unexpected keyword argument 'shell'
```

I fixed it with modifying the svr4pkg module (old code is commented out)

```
    #rc, out, err = module.run_command(' '.join(cmd), shell=False)
    rc, out, err = module.run_command(' '.join(cmd))
```

PR #18038, opened 2016-10-14
Add a SIGTERM handler that just exits.

alikins:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

```
ansible 2.3.0 (devel 578170a908) last updated 2016/10/14 19:16:23 (GMT -400)
  lib/ansible/modules/core: (detached HEAD 275fa3f055) last updated 2016/10/12 14:06:23 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 3e1ea76a75) last updated 2016/10/13 11:34:41 (GMT -400)

```
##### SUMMARY

```

```

Prevents the main process from ending up
being orphaned if it gets a SIGTERM.

Fixes ansible-tower issue #3617

### DIFF
--- a/bin/ansible
+++ b/bin/ansible
@@ -45,6 +45,12 @@ from ansible.errors import AnsibleError, AnsibleOptionsError, AnsibleParserError
 from ansible.utils.display import Display
 from ansible.module_utils._text import to_text
 
+import signal
+def signal_handler(signum, frame):
+	sys.exit(0)
+
+signal.signal(signal.SIGTERM, signal_handler)
+
 
 ########################################
 ### OUTPUT OF LAST RESORT ###


mscherer:
So, isn't it supposed to be the default already ? 

SIGTERM should result in the process to stop, and I do not have access to the bug on ansible-tower to be able to see what is the exact issue

alikins:
For the tower case, tower ends up launching the ansible-playbook subprocess, and attempting to kill it with SIGTERM. And it was ending up leaving one of the ansible-playbook processes reattached to pid 1 (I believe there is an exec or two via pexpect involved and tower is sigterm'ing one  of those instead of the ansible-playbook MainProcess, but I forget the details now).

The sigterm handler seemed to prevent tower from leaving the dangling process as well as letting various atexit handlers run which they weren't before. 

I've been told we've tried this before though and eventually revert'ed for some reason or another but dont know the details.

PR #18044, opened 2016-10-15
Fix become password using non-ascii for local connection

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

plugins/connection/local.py
##### ANSIBLE VERSION

```
devel 2.2 and 2.1
```
##### SUMMARY

```

```


abadger:
Failure is a timeout starting the windows target host.  This is for the local connection plugin.  So we're good to go.

abadger:
Merged to devel and cherry-picked to stable-2.2 and stable-2.1

abadger:
This should be in 2.2.0 and 2.1.3

PR #18049, opened 2016-10-16
Enable test on mercurial for python 3

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

test for hg


mscherer:
See #18051

mattclay:
Closing this since the method for enabling/disabling tests on python 3 has changed. Also, the test currently won't pass in CI because we only have python 3 installed when testing python 3.

We'll either need to wait until mercurial supports python 3, or we'll need to update our python 3 test infrastructure to provide python 2.

PR #18052, opened 2016-10-16
Activate service test for python 3

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

test of service

### DIFF
--- a/test/utils/shippable/python3-test-tag-blacklist.txt
+++ b/test/utils/shippable/python3-test-tag-blacklist.txt
@@ -4,5 +4,4 @@ test_git
 test_hg
 test_lookups
 test_pip
-test_service
 test_uri


mattclay:
@mscherer This test shouldn't be removed from the blacklist. It doesn't actually pass on Shippable. The only reason tests are passing is because the test is flagged as needs_privileged and all privileged tests are currently being skipped on Shippable.

mscherer:
Do we have a list of errors of service on python 3 ? I just took a checkout of HEAD on a fresh ubuntu 16.04, and after fixing the shebang on ansible-playbook, and doing some change to make sure this run with python3, "make destructive" run fine for the service tests, but I only have that with a systemd enabled service.

Since the test cover upstart, systemd and sysv. Systemd is working as far as I know. Upstart is not gonna be used on a system with python 3.5 afaik. So we just need to verify sysv I guess, or do we consider it would have the same fate as upstart as far as compatibility with python 3 is concerned ?

frenzymadness:
Could somebody please rerun the tests? I want to take a look at this issue but logs are not available in shippable.

Maybe @mscherer could solve the simple conflict in blacklist file which may run tests again.

Thanks.

mscherer:
So the problem is not related to python 3, it is related to the fact that this is blacklisted in both python 2 and python 3 for CI related reason. I can rebase, but since the service plugin is not tested in CI, this will not trigger any errors :)
And as far as I know, service is working on python 3 (and I did tried it on various BSD and Linux).

mattclay:
@mscherer This change is obsolete. I'm looking at re-enabling the service tests in PR #19841.

PR #18056, opened 2016-10-16
Enable python3 test for git

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

test of git


mscherer:
So the CI fail on unrelated issues (async, windows, freebsd), but not on git and python 3, so I would suspect it to be mergable.

abadger:
Merged via https://github.com/ansible/ansible/commit/2ac12432ef666d43b39b5fa60087f839cc64f586

PR #18060, opened 2016-10-16
Enable tests on python 3 for uri

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

test of uri

### DIFF
--- a/test/utils/shippable/python3-test-tag-blacklist.txt
+++ b/test/utils/shippable/python3-test-tag-blacklist.txt
@@ -5,4 +5,3 @@ test_hg
 test_lookups
 test_pip
 test_service
-test_uri


mscherer:
So the failure is in 'check pass_json' ( https://github.com/ansible/ansible/blob/devel/test/integration/targets/uri/tasks/main.yml#L62 ), with a error about invalid json (see the failure in shippable).

frenzymadness:
I spent almost whole day investigating this issue and I learned a lot of new stuff about Ansible and its core code.

I've found that mentioned test fails only with file pass0.json. Tests with files pass[1-4].json are correct. I think that reason for this issue is that during registration of variable `pass` the output of `fetch pass_json` is not recognized as JSON and output of `fetch pass_json` is stored in `pass` variable as a string. In Python 2, the output of `fetch pass_json` is correctly recognized and stored as JSON in variable `pass`.

I want to solve this problem, but I cannot find source code which takes care about registering variables. Could you please point me to the certain file?

Thank you.

mscherer:
In https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/strategy/__init__.py , in the _process_pending_results function.

frenzymadness:
Hi.

I was on vacation last week. But now, I am debugging this problem from different points of view. I've checked this problem from `TaskExecutor`/`Worker`/`TaskResult` side, then from `assert`, `uri`, `together` and `debug` modules side and then I've finally found problematic part of code.

The problem is in `save_eval` function [0] which in Python 3 raises `invalid expression` exception and then result of this function isn't converted from string to dictionary.

As I said before, this happens only if the result of `fetch pass_json`[1] (with pass0.json item) is registered as variable and then used in some next task as in the test of the uri module.

I am using this simple playbook (with some hard coded values) to reproduce this issue:
```
---
- hosts: localhost
  gather_facts: False
  tasks:
    - name: checksum pass_json
      stat: path=./original/{{ item }}.json get_checksum=yes
      register: pass_checksum
      with_sequence: start=0 end=0 format=pass%d

    - name: fetch pass_json
      uri: return_content=yes url=http://localhost:8000/{{ item }}.json
      register: pass
      with_sequence: start=0 end=0 format=pass%d

    - name: 'print registered variable pass (from http)'
      debug: var=pass
```

The difference is visible in the output of the last task (If you run this playbook in Python2/3).

I am not currently familiar with AST module, but I'll continue with the investigation and I'll try to fix this issue.

[0] https://github.com/ansible/ansible/blob/devel/lib/ansible/template/safe_eval.py#L29
[1] https://github.com/ansible/ansible/blob/devel/test/integration/targets/uri/tasks/main.yml#L57

frenzymadness:
Ok, I have a solution. The solution is to add `ast.USub` to `SAFE_NODES` set in `safe_eval.py`.

The ast.USub is the class to represent unary operators and it looks like that JSON file used for testing contains something which needs this for parsing this JSON as part of Python object.

@mscherer: Should I create new PR for this change or could you please give me somehow push privileges to source branch of this PR?

mscherer:
Wow, good job. I am ok to let you commit on the branch, but I have no idea on how to do it. Better take my 1 line patch in a new PR.

abadger:
Merged https://github.com/ansible/ansible/issues/18597 which contains this.  Closing.

PR #18063, opened 2016-10-16
Enable the pip test for python 3

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

test of pip

### DIFF
--- a/test/utils/shippable/python3-test-tag-blacklist.txt
+++ b/test/utils/shippable/python3-test-tag-blacklist.txt
@@ -3,6 +3,5 @@ test_filters
 test_git
 test_hg
 test_lookups
-test_pip
 test_service
 test_uri


mscherer:
It fail because pip install the version for python2 of epdb. I am not sure why it does that, and if that's a bug in the test suite of the module.

mscherer:
So https://github.com/ansible/ansible-modules-core/pull/5281 should hopefully fix that. But I didn't test it, and I am not sure if that's the right default behavior ( it is logical for me, but maybe not for the vast majority of user)

mscherer:
So i am now slightly puzzled. Now that the PR is merged, we can see:

```
2016-10-17 21:30:05 changed: [testhost] => {"changed": true, "cmd": "/usr/bin/pip3 install epdb", "name": ["epdb"], "requirements": null, "state": "present", "stderr": "You are using pip version 8.1.1, however version 8.1.2 is available.\nYou should consider upgrading via the 'pip install --upgrade pip' command.\n", "stdout": "Collecting epdb\n  Downloading epdb-0.12.tar.bz2\nBuilding wheels for collected packages: epdb\n  Running setup.py bdist_wheel for epdb: started\n  Running setup.py bdist_wheel for epdb: finished with status 'done'\n  Stored in directory: /root/.cache/pip/wheels/98/a9/08/2b68b7aa2a3cbaae687b3f2c625ba3436041929e7956a820fb\nSuccessfully built epdb\nInstalling collected packages: epdb\nSuccessfully installed epdb-0.12\n", "stdout_lines": ["Collecting epdb", "  Downloading epdb-0.12.tar.bz2", "Building wheels for collected packages: epdb", "  Running setup.py bdist_wheel for epdb: started", "  Running setup.py bdist_wheel for epdb: finished with status 'done'", "  Stored in directory: /root/.cache/pip/wheels/98/a9/08/2b68b7aa2a3cbaae687b3f2c625ba3436041929e7956a820fb", "Successfully built epdb", "Installing collected packages: epdb", "Successfully installed epdb-0.12"], "version": null, "virtualenv": null}
```

But it still install a non python3 compatible version of epdb and even not the latest one (since that's 0.12 being installed, and pypi say that latest is 0.15: https://pypi.python.org/pypi/epdb/0.15 ).

abadger:
I think that was because the latest doesn't have a source tarball, it only had a wheel.  I've switched the package we test from epdb to isort.  That seems to have gotten past that error but has arrived at a different one.

mscherer:
For the record, the error we see now:

```
TASK [pip : install the requirement file in a virtualenv] **********************
fatal: [testhost]: FAILED! => {"changed": false, "cmd": "/usr/bin/virtualenv --no-site-packages /root/ansible_testing/pipenv", "failed": true, "msg": "stdout: The executable python2 (from --python=python2) does not exist\n"}
```

mscherer:
So that kind sound like a bug in pip and/or virtualenv, because we do not specify --python anywhere in the module.

And python 3.3 do have its own venv module built-in to replace virtualenv it seems (cf https://docs.python.org/3/library/venv.html#module-venv ), so maybe we need to take that in account ?

abadger:
Merged via 67ac37518830f92db8d4f453922abf328d48ddf4 to  devel and stable-2.2

PR #18064, opened 2016-10-16
Enable test for lookups on python 3

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

test of lookups on python 3

### DIFF
--- a/lib/ansible/utils/encrypt.py
+++ b/lib/ansible/utils/encrypt.py
@@ -67,7 +67,7 @@ except ImportError:
 
 from ansible import constants as C
 from ansible.errors import AnsibleError
-from ansible.module_utils._text import to_text
+from ansible.module_utils._text import to_text, to_native
 
 __all__ = ['do_encrypt']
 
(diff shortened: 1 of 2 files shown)


mscherer:
The failure is the following:

```
Unexpected templating type error occurred on ({{ lookup('password', output_dir + '/lookup/password_with_salt encrypt=sha256_crypt') }}): salt must be unicode, not bytes
```

On "create a password (with salt) file" ( https://github.com/ansible/ansible/blob/devel/test/integration/targets/lookups/tasks/main.yml#L104 )

mscherer:
So the issue is linked to https://bitbucket.org/ecollins/passlib/src/530cffc97945a940d15af42e226543ae3a018035/passlib/utils/handlers.py?at=default#handlers.py-1382 

Using to_native instead of to_bytes just seems to break something else in the test suit.

mscherer:
So it seems that depending on the salt, it can or cannot be unicode (or bytes), since some salt have the HasSalt and some have the HasRawSalt mixin. For example, fhsp, scram, pbkdf2 would take a 'bytes' for salt, while other would take 'unicode', unless if running on python 2. 

So my understanding is that we would need to convert depending on the type of algo being used, and that's slightly annoying.

abadger:
Does seem a little hacky but we're just working around the limitations of the library so it seems like a good fix.  Thanks!

Issue #17308, opened 2016-08-30
Test suite fail on python3 for test_filters.yml


mscherer:
##### ISSUE TYPE
- Bug Report
##### COMPONENT NAME

test
##### OS / ENVIRONMENT

Fedora 24
##### SUMMARY

test_filters.yml fail on python3 due to incorrect test.
##### STEPS TO REPRODUCE

trying to fix the test suite for python3 (or rather, the code), I stumbled on this failure:

```
TASK [test_filters : compare templated file to known good] *********************
task path: /home/misc/checkout/git/ansible/test/integration/roles/test_filters/tasks/main.yml:54
Using module file /home/misc/checkout/git/ansible/lib/ansible/modules/core/commands/command.py
<127.0.0.1> ESTABLISH LOCAL CONNECTION FOR USER: root
<127.0.0.1> EXEC /bin/sh -c '( umask 77 && mkdir -p "` echo $HOME/.ansible/tmp/ansible-tmp-1472563893.2081912-240123655525728 `" && echo ansible-tmp-1472563893.2081912-240123655525728="` echo $HOME/.ansible/tmp/ansible-tmp-1472563893.2081912-240123655525728 `" ) && sleep 0'
<127.0.0.1> PUT /tmp/tmpm3vcpqpg TO /root/.ansible/tmp/ansible-tmp-1472563893.2081912-240123655525728/command.py
<127.0.0.1> EXEC /bin/sh -c 'chmod u+x /root/.ansible/tmp/ansible-tmp-1472563893.2081912-240123655525728/ /root/.ansible/tmp/ansible-tmp-1472563893.2081912-240123655525728/command.py && sleep 0'
<127.0.0.1> EXEC /bin/sh -c '/usr/bin/python /root/.ansible/tmp/ansible-tmp-1472563893.2081912-240123655525728/command.py; rm -rf "/root/.ansible/tmp/ansible-tmp-1472563893.2081912-240123655525728/" > /dev/null 2>&1 && sleep 0'
fatal: [testhost]: FAILED! => {
    "changed": true,
    "cmd": "diff -w ~/ansible_testing/foo.templated ~/ansible_testing/foo.txt",
    "delta": "0:00:00.003852",
    "end": "2016-08-30 15:31:33.386117",
    "failed": true,
    "invocation": {
        "module_args": {
            "_raw_params": "diff -w ~/ansible_testing/foo.templated ~/ansible_testing/foo.txt",
            "_uses_shell": true,
[...]

PR #18065, opened 2016-10-16
Enable filters test for python 3

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

tests

### DIFF
--- a/test/utils/shippable/python3-test-tag-blacklist.txt
+++ b/test/utils/shippable/python3-test-tag-blacklist.txt
@@ -1,4 +1,3 @@
-test_filters
 test_git
 test_hg
 test_lookups


mscherer:
This fail on:
```
TASK [lookups : create a password (with salt) file] ****************************
fatal: [testhost]: FAILED! => {"failed": true, "msg": "Unexpected templating type error occurred on ({{ lookup('password', output_dir + '/lookup/password_with_salt encrypt=sha256_crypt') }}): salt must be unicode, not bytes"}
```

Which is different from https://github.com/ansible/ansible/issues/17308

mscherer:
So this PR is linked to https://github.com/ansible/ansible/issues/17308 (but at the time I comment, shippable do not react)

mscherer:
So now that #17308 is fixed, and since CI should be green (it was red due to yum timeout), I guess this one can be merged ?

abadger:
Merged to devel and cherry-picked to stable-2.2

PR #18073, opened 2016-10-17
fix cwd for ssh connections

### LINKED ISSUE #17896: [2.1.2] template and copy still uses old CWD
##### ISSUE TYPE

- Bug Report
##### ANSIBLE VERSION

```
2.1.2 
```
##### CONFIGURATION

ansible.cfg with an special config for a basion jump host.

``` ini
[ssh_connection]
ssh_args = -F ssh.config
```

playbooks and ssh.config are located in a subdir `playbooks/`
##### OS / ENVIRONMENT

N/A
##### SUMMARY

With e2f17f8d9b212b1e8d65c12a778d036753831ab6 the CWD changed to the location where the playbooks are located, in this case ./playbooks. ssh.config has been moved there and this worked. 

However for the plugins setup, copy and template, ansible still expects the ssh.config to be located in the location as in 2.1.1. 
##### STEPS TO REPRODUCE

```
echo www.example.com > hosts
mkdir -p playbooks/files
echo "foo" > playbooks/files/test.txt
touch playbooks/ssh.config # can be empty!

cat ansible.cfg
[ssh_connection]
ssh_args = -F ssh.config 

cat playbooks/test.yml 

---
- hosts: all
  tasks:
  - copy: src=files/test.txt dest=/tmp

ansible-playbook playbooks/test.yml -i hosts --user root
```

after cp the ssh.config to the old "ansible 2.0" location, it works again!

```
cp playbooks/ssh.config .   
```

##### EXPECTED RESULTS

```

ok: [build-01.stxt.media.int]

changed: [build-01.stxt.media.int]

build-01.stxt.media.int    : ok=2    changed=1    unreachable=0    failed=0  
```
##### ACTUAL RESULTS

```
 $ ansible-playbook playbooks/test.yml -i hosts --user root

fatal: [build-01.stxt.media.int]: UNREACHABLE! => {"changed": false, "msg": "Failed to connect to the host via ssh.", "unreachable": true}
    to retry, use: --limit @/home/resmo/Projects/swisstxt/tmp/playbooks/test.retry

build-01.stxt.media.int    : ok=0    changed=0    unreachable=1    failed=0   

```

With `gather_facts: no`:

```

fatal: [build-01.stxt.media.int]: UNREACHABLE! => {"changed": false, "msg": "SSH Error: data could not be sent to the remote host. Make sure this host can be reached over ssh", "unreachable": true}
    to retry, use: --limit @/home/resmo/Projects/swisstxt/tmp/playbooks/test.retry

build-01.stxt.media.int    : ok=0    changed=0    unreachable=1    failed=0
```

resmo:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

n/a
##### ANSIBLE VERSION

```
devel 
2.2 
2.1.2
```
##### SUMMARY

Before ansible still used the cwd as in 2.1.1 when etablishing ssh connection
resulting in a ssh config e.g. '-F ssh.config' is not found.

Fixes #17896

### DIFF
--- a/lib/ansible/plugins/action/__init__.py
+++ b/lib/ansible/plugins/action/__init__.py
@@ -272,7 +272,12 @@ class ActionBase(with_metaclass(ABCMeta, object)):
             self._low_level_execute_command(cmd, sudoable=False)
 
     def _transfer_file(self, local_path, remote_path):
-        self._connection.put_file(local_path, remote_path)
+        cwd = os.getcwd()
+        os.chdir(self._loader.get_basedir())
+        try:
+            self._connection.put_file(local_path, remote_path)
+        finally:
+            os.chdir(cwd)
         return remote_path
 
     def _transfer_data(self, remote_path, data):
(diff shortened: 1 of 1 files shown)


bcoca:
This will use diff paths for file transfers than for setting up connections, if the problem is that -f ./config is getting ignored ... this won't help. This, however, sets the relative paths for file transfers to be based on the play.

PR #18077, opened 2016-10-17
Fix ansible-pull on python3

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

ansible-pull via utils/cmd_functions.py
##### ANSIBLE VERSION

```
devel, 2.2
```
##### SUMMARY

On python3, we can't write bytes directly to sys.stdout.

Write to sys.stdout.buffer instead.


abadger:
Test failure is network related rather than ansible related.

PR #18078, opened 2016-10-17
epdb doesn't work on python3 so we need a different package to test pip

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

test for pip
##### ANSIBLE VERSION

```
devel 2.2
```
##### SUMMARY

epdb doesn't work on python3 so pip is unable to install it for python3.  Use a different package to test pip.


abadger:
Merged to devel and stable-2.2

PR #15551, opened 2016-04-22
Add tag verification test

pixelrebel:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.0.1.0
```
##### SUMMARY

This tests the PR https://github.com/ansible/ansible-modules-core/pull/2654

PR #18089, opened 2016-10-18
Pixelrebel amc pr2654

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

git tests
##### ANSIBLE VERSION

```
devel 2.2
```
##### SUMMARY

Rebased the tests from here: https://github.com/ansible/ansible/pull/15551

### DIFF
--- a/test/integration/targets/git/tasks/main.yml
+++ b/test/integration/targets/git/tasks/main.yml
@@ -674,7 +674,7 @@
   command: gpg --import-ownertrust /tmp/ownertrust-git.txt
 
 - name: Clone signed repo and verify tag
-  git: repo={{ repo_signed }} dest={{ checkout_dir }} version=v1.9.1 verify_commit=yes
+  git: repo={{ repo_verify }} dest={{ checkout_dir }} version=v1.9.1 verify_commit=yes
 
 - name: Remove Junio C Hamano GPG key
   command: gpg --batch --yes --delete-key B0B5E88696AFE6CB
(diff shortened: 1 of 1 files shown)


abadger:
gpg tests are being skipped on Fedora hosts because gnupg2 (/usr/bin/gpg2) is installed by default but not gnupg1 (/usr/bin/gpg).

@pixelrebel @robinro do we want to support gnupg2 in the git module?  If not, I'll ask mattclay to install the gnupg1 package in the fedora docker images.

abadger:
Other than that question (which doesn't need any code changes here), this is ready to be merged... will do so tomorrow.

abadger:
Note: this feature will probably miss 2.2.0 but will be in 2.2.1 (and it may make 2.2.0 if we have more RCs).

robinro:
@abadger @pixelrebel I think we should support gpg2. 

This pretty much works already. Changes needed are:
- [ ] gpg2 instead of gpg in the `command` calls of the tests
- [ ] `git config --global gpg.program gpg2` before the git command
- [ ] install `dirmngr` on ubuntu. We can do this in the test for the moment and at some point move it to the docker image.

With this it runs on ubuntu1604, opensuseleap and fedora-rawhide.

I would leave the git module as is and leave it to the user/OS to install gpg and set the git config to gpg or gpg2 depending on their use. Some people might have both.

abadger:
Merged this to devel and stable-2.2.

@robinro sounds like a good plan.  if you need me to merge something, feel free to ping me on irc.freenode.net (I'm abadger1999 on #ansible-devel)

PR #18093, opened 2016-10-18
[GCE] Caching support for inventory script.

supertom:
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

GCE inventory script
##### ANSIBLE VERSION

```
ansible 2.3.0 (ans-inventory 13d9dd37b6) last updated 2016/10/18 19:51:14 (GMT +000)
  lib/ansible/modules/core: (detached HEAD 500736d301) last updated 2016/10/17 16:09:30 (GMT +000)
  lib/ansible/modules/extras: (detached HEAD 9897c502a6) last updated 2016/10/17 16:09:32 (GMT +000)
  config file = /home/supertom/.ansible.cfg
```
##### SUMMARY

The GCE inventory script now supports reading from a cache rather than making the request each time.  The format of the list and host output have not changed.

On script execution, the cache is checked to see if it older than 'cache_max_age', and if so, it is rebuilt (it can also be explicity rebuilt).

To support this functionality, the following have been added/modified:
- Config file (gce.ini) changes: A new 'cache' section has been added to the config file, with 'cache_path' and 'cache_max_age' options to allow for configuration.  There are intelligent defaults in place if that section and options are not found in the configuration file.
- Command line argument: A new --refresh-cache argument has been added to force the cache to be rebuild.
- A CloudInventoryCache class, contained in the same file has been added.  As a seperate class, it allowed for testing (unit tests not included in this PR) and hopefully could be re-used in the future (it contains borrowed code from other inventory scripts)
- load_inventory_from_cache and do_api_calls_and_update_cache methods (, which were largely lifted from other inventory scripts, in a hope to promote consistency in the future) to determine if the cache is fresh and rebuild if necessary.
- A 'main' check, to support the script being imported and testable.
- A new dictionary has been added to the list output, located at ['_meta']['stats'] that informs if the cache was used and how long it took to load the inventory (in 'cache_used' and 'inventory_load_time', respectively).

/cc @ryansb

### DIFF
--- a/contrib/inventory/gce.py
+++ b/contrib/inventory/gce.py
@@ -220,7 +220,7 @@ class GceInventory(object):
             'libcloud_secrets': '',
             'inventory_ip_type': '',
             'cache_path': '~/.ansible/tmp',
-            'cache_max_age': 60
+            'cache_max_age': '300'
         })
         if 'gce' not in config.sections():
             config.add_section('gce')
--- a/contrib/inventory/gce.py
+++ b/contrib/inventory/gce.py
@@ -69,8 +69,8 @@ Examples:
   $ contrib/inventory/gce.py --host my_instance
 
 Author: Eric Johnson <erjohnso@google.com>
-Contributors: Matt Hite <mhite@hotmail.com>
-Version: 0.0.2
+Contributors: Matt Hite <mhite@hotmail.com>, Tom Melendez <supertom@google.com>
+Version: 0.0.3
 '''
 
 __requires__ = ['pycrypto>=2.6']
@@ -89,6 +89,9 @@ USER_AGENT_VERSION="v2"
 import sys
 import os
 import argparse
+
+from time import time
+
 import ConfigParser
 
 import logging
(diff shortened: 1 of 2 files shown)


sivel:
This PR seems to make a number of stylistic changes to the code which is not permitted.

Please remove all style changes and resubmit with only the changes targeting this new feature.

supertom:
ready_for_review

OK, backed out style-related changes.

supertom:
Windows error:

```
2016-10-18 21:24:32 PLAY RECAP *********************************************************************
2016-10-18 21:24:32 windows                    : ok=15   changed=1    unreachable=0    failed=1 
```

contrib/inventory/gce.py
```diff
@@ -117,22 +169,36 @@ def __init__(self):
...
-                    self.get_instance(self.args.host)),
-                    pretty=self.args.pretty))
-            sys.exit(0)
-
-        zones = self.parse_env_zones()
-
-        # Otherwise, assume user wants all instances grouped
-        print(self.json_format_dict(self.group_instances(zones),
-            pretty=self.args.pretty))
+            print self.json_format_dict(
```
ryansb: Ah, you'll need to use the `print('foo')` function instead of the statement for Python 3.x compatibility.

contrib/inventory/gce.py
```diff
@@ -117,22 +169,36 @@ def __init__(self):
...
+                self.inventory['_meta']['hostvars'][self.args.host],
+                pretty=self.args.pretty)
+        else:
+            # Otherwise, assume user wants all instances grouped
+            zones = self.parse_env_zones()
+            print self.json_format_dict(self.inventory,
```
ryansb: Same here

supertom: Done

contrib/inventory/gce.py
supertom: Done

supertom:
@ryansb py3.5 error seems resolved, but now I see this:
 TASK [yum : install sos]  fatal: (snipped) No more mirrors to try.\nhttp://download-i2.fedoraproject.org/pub/fedora/linux/updates/23/x86_64/repodata/repomd.xml: [Errno -1] repomd.xml does not match metalink for updates\

supertom:
@ryansb PTAL.  Had a user verify it as well (and report a minor error, which I fixed).

ryansb:
Looks good to me - thanks for the contribution :smile:

PR #18098, opened 2016-10-19
Update the ansible_managed documentation.

### LINKED ISSUE #17947: Default ansible_managed to a static string
##### ISSUE TYPE
- Feature Idea
##### COMPONENT NAME

Various templates
##### ANSIBLE VERSION

```
devel, 2.3, 2.2
```
##### SUMMARY

Using the default values of "{{ ansible_managed }}" in a template file leads to the file being changed.  This is due to the variables used in the default definition of ansible_managed:
- The date replacement leads to it being changed whenever the date of the template's source is different.
  - This can happen when playbooks are stored in git as git does not store file timestamps.  Each checkout creates files from that date.  In a multi-admin setup, it is common for the playbooks to be shared with other admins via git.
- The uid replacement leads to it being changed anytime the playbook is run from a different account.
  - Running ansible does not require a single admin account because of its integration with sudo and su.  But this portion of the string hampers that because template will report changed each time a different account is used.
- The host replacement leads it to being changed whenever the playbook is run from a different controller.
  - Ansible is useful to run playbooks from each admin's laptop or workstation rather than having a single central master.  This setting makes it so templates cannot be used idempotently in that manner.
- The file replacement leads to the file being changed whenever the file is substituted from a different path.
  - When ansible gets its playbooks from a git checkout in each admin's home directory the paths to the template file will be different.  Thus each admin will get a change in the file.
##### STEPS TO REPRODUCE

Run the following playbook twice to show the first problem with date.  Note that in the real-world, the change in timestamp would come from using a fresh git checkout or a user on another machine, not from state='touch' in the playbook:

```

---
- hosts: localhost
  gather_facts: False
  tasks:
  - file:
      path: "/var/tmp/not-idempotent.j2"
      state: "touch"
  - copy:
      content: "{{'{{ ansible_managed }}\nSome content\n'}}"
      dest: "/var/tmp/not-idempotent.j2"
  - template:
      src: "/var/tmp/not-idempotent.j2"
      dest: "/var/tmp/not-idempotent"
```

You can take out the `file:` task and run the playbook as another user or on a different control host to see the problems with uid and host respectively.
##### EXPECTED RESULTS

The second run of the playbook should show no changes for the template task.
##### ACTUAL RESULTS

The template task shows changes every single run:

```
TASK [template] ****************************************************************
changed: [localhost]
```

abadger:
##### ISSUE TYPE
- Docs Pull Request
##### COMPONENT NAME

docsitre/rst/intro_configuration.rst
##### ANSIBLE VERSION

```
devel
```
##### SUMMARY
- New default (a static string)
- Explanation of all the fields and how they impact idempotence

Fixes #17947

### DIFF
--- a/docsite/rst/intro_configuration.rst
+++ b/docsite/rst/intro_configuration.rst
@@ -105,18 +105,50 @@ It is useful when becoming an unprivileged user::
 ansible_managed
 ===============
 
-Ansible-managed is a string that can be inserted into files written by Ansible's config templating system, if you use
-a string like::
+``ansible_managed`` is a string that can be inserted into files written by
+Ansible's config templating system when you use a string like::
 
    {{ ansible_managed }}
 
-The default configuration shows who modified a file and when::
+The default value only says that Ansible is managing this file::
 
-    ansible_managed = Ansible managed: {file} modified on %Y-%m-%d %H:%M:%S by {uid} on {host}
+    ansible_managed = Ansible managed
 
-This is useful to tell users that a file has been placed by Ansible and manual changes are likely to be overwritten.
+This can be helpful to inform people looking at the file that they should not
+edit it in-place; they need to edit it in Ansible otherwise their changes will
+be overwritten.
 
-Note that if using this feature, and there is a date in the string, the template will be reported changed each time as the date is updated.
+There are several special values that, when placed in the string, will be
+replaced with useful information.  These are not in the default
+``ansible_managed`` string because they can cause Ansible to think that the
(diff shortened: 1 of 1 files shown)


gundalow:
Idempotence information is good.

I think this should go into the CHANGELOG as well since it's a change in behaviour.

abadger:
Thanks guys.
@gundalow changelog entry went in here: https://github.com/ansible/ansible/commit/a0a4d0e3f4835c282212799a99834fc017c4a450

PR #18100, opened 2016-10-19
Remove module_utils/splitter

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

modules_utils
##### SUMMARY

This code is not used anywhere, and that's a duplicate from
lib/ansible/parsing/splitter.py. However, the code in ansible.parsing
is tested for python3, and this one is not and fail.

So to avoid code duplication, and license headaches, since
nothing in the repository use the code, better remove it.

blockinfile import (but do not use) the module, and this is tracked in https://github.com/ansible/ansible-modules-extras/pull/3235 

ansible-lint use it too, and this is discussed on https://github.com/willthames/ansible-lint/pull/209


mscherer:
So after a quick search on github, I found a ton of people who bundle the whole source code of ansible, and only a handful of modules using it:
https://github.com/BROADSoftware/ansible_roles/blob/b605ff98702ae5b56e8401b9787dd692689960da/hadoop-user/hdfs_modules/library/hdfs_cmd.py

https://github.com/osielczakd/ansible/blob/master/modules/iptables.py

https://github.com/dispansible/dispansible/blob/af9868b3b59d4125207d13b73306b11810c93521/ansible/library/disp_command.py

https://github.com/openshift/openshift-ansible/blob/master/roles/etcd_common/library/delegated_serial_command.py

https://github.com/tcr/tusk/blob/master/template/library/run/run

Out of all of this, only disp_command seems to rely on split_args, rest is likely just cut and paste from command.py without the needed cleanup.

mscherer:
And shippable failure is on yum tests, due to timeout.

abadger:
@jimi-c wants to keep module_utils/splitter.py and get rid of parsing /splitter.py because of the third-party code.

bcoca:
closing as per above, the intention is to migrate the code that references the other files to this one.

Issue #202, opened 2012-04-23
python2.5 support for ansible and ansible-playbook


egon1024:
Downloaded ansible for the first time this morning and am trying to run through it.  I installed the multiprocessing module (I'm running on Python 2.5) via pip.  I got the following:

```
ctuininga@erbium$ ansible all -m ping --ask-pass
/home/ctuininga/ansible/ansible/lib/ansible/connection.py:24: Warning: 'with' will become a reserved keyword in Python 2.6
Traceback (most recent call last):
  File "/home/ctuininga/ansible/ansible/bin/ansible", line 27, in <module>
    import ansible.runner
  File "/home/ctuininga/ansible/ansible/lib/ansible/runner.py", line 34, in <module>
    import ansible.connection
  File "/home/ctuininga/ansible/ansible/lib/ansible/connection.py", line 24
    with warnings.catch_warnings():
                ^
SyntaxError: invalid syntax
```

Using the "with" statement in Python 2.5 requires a "from **future** import with_statement"

In case it somehow turns out to be relevant, my platform is Python 2.5.6 running on FreeBSD 7.2

egon1024:
I added the "from future" statement locally which led to my next 2.5 error ... the warnings module does not have a catch_warnings attribute:

```
ctuininga@erbium$ ansible all -m ping --ask-pass
Traceback (most recent call last):
  File "/home/ctuininga/ansible/ansible/bin/ansible", line 27, in <module>
    import ansible.runner
  File "/home/ctuininga/ansible/ansible/lib/ansible/runner.py", line 34, in <module>
    import ansible.connection
  File "/home/ctuininga/ansible/ansible/lib/ansible/connection.py", line 25, in <module>
    with warnings.catch_warnings():
AttributeError: 'module' object has no attribute 'catch_warnings'
```

mpdehaan:
Yeah, I'm not sure we should claim support for 2.5 on the ansible client.   We support all the way down to 2.4 on the nodes, but there are nice things in 2.6 I'd like to use, and RHEL 6 already has 2.6.

This also means someone doesn't have to go out of their way to get the multiprocessing module to work.

egon1024:
Ansible looks pretty damned cool, so I'm wondering if you'll be willing to take patches for 2.5 support if I'm willing to put them together.  If it's just not something you're interested in, I certainly understand.  :)

mpdehaan:
[...]

PR #18101, opened 2016-10-19
Refresh the Galaxy docs

chouseknecht:
##### ISSUE TYPE
- Docs Pull Request
##### COMPONENT NAME

galaxy.rst
##### ANSIBLE VERSION

It was pointed out to me on [Galaxy issue #202](https://github.com/ansible/galaxy-issues/issues/202) that this doc was quite outdated and needed some attention.

### DIFF
--- a/docsite/rst/galaxy.rst
+++ b/docsite/rst/galaxy.rst
@@ -1,105 +1,110 @@
 Ansible Galaxy
 ++++++++++++++
 
-"Ansible Galaxy" can either refer to a website for sharing and downloading Ansible roles, or a command line tool for managing and creating roles.
+*Ansible Galaxy* refers to the `Galaxy <https://galaxy.ansible.com>`_  website where users can share roles, and to a command line tool for installing, 
+creating and managing roles.
 
 .. contents:: Topics
 
 The Website
 ```````````
 
-The website `Ansible Galaxy <https://galaxy.ansible.com>`_, is a free site for finding, downloading, and sharing community developed Ansible roles. Downloading roles from Galaxy is a great way to jumpstart your automation projects.
+`Galaxy <https://galaxy.ansible.com>`_, is a free site for finding, downloading, and sharing community developed roles. Downloading roles from Galaxy is 
+a great way to jumpstart your automation projects.
 
-Access the Galaxy web site using GitHub OAuth, and to install roles use the 'ansible-galaxy' command line tool included in Ansible 1.4.2 and later.
+You can also use the site to share roles that you create. By authenticating with the site using your GitHub account, you're able to *import* roles, making 
+them available to the Ansible community. Imported roles become available in the Galaxy search index and visible on the site, allowing users to 
+discover and download them.
 
-Read the "About" page on the Galaxy site for more information.
+Learn more by viewing `the About page <https://galaxy.ansible.com/intro>`_.
 
-The ansible-galaxy command line tool
-````````````````````````````````````
+The command line tool
+`````````````````````
 
-The ansible-galaxy command has many different sub-commands for managing roles both locally and at `galaxy.ansible.com <https://galaxy.ansible.com>`_.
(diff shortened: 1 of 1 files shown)


chouseknecht:
@bcoca can you please give this a once-over when you have a minute. I want to make sure my re-write is technically accurate. I'm sure I missed something.

Thanks!

docsite/rst/galaxy.rst
```diff
@@ -1,105 +1,104 @@
...
+::
+
+    $ ansible-galaxy install username.role_name
 
 roles_path
 ==========
 
-You can specify a particular directory where you want the downloaded roles to be placed::
+Be aware that Ansible downloads roles to *ANSIBLE_ROLES_PATH*, which defaults to */etc/ansible/roles*. You can override this by setting the environment 
+variable in your session, defining *roles_path* in an *ansible.cfg* file, or by using the *--roles-path* option.
```
bcoca: roles path can be a series of directories i.e. `/etc/ansible/roles:~/.ansible/roles`, the default is correct but it should install to the 'first writable path configured'

docsite/rst/galaxy.rst
```diff
@@ -1,105 +1,104 @@
...
 
-To install multiple roles, the ansible-galaxy CLI can be fed a requirements file.  All versions of ansible allow the following syntax for installing roles from the Ansible Galaxy website::
+    $ ansible-galaxy install --roles-path . geerlingguy.apache
 
-   $ ansible-galaxy install -r requirements.txt
+.. seealso::
 
-Where the requirements.txt looks like::
+   :doc:`intro_configuration`
+      All about configuration files
```
bcoca, line 58: This file is deprecated in favor of the YAML format and should be removed in 2.3

docsite/rst/galaxy.rst
```diff
@@ -129,48 +128,100 @@ And here's an example showing some specific version downloads from multiple sour
       scm: git
       version: 0.1.0
 
-As you can see in the above, there are a large amount of controls available
-to customize where roles can be pulled from, and what to save roles as. 
+Installing multiple roles from multiple files
+=============================================
+
```
bcoca: this was never implemented, i thought i removed these docs already

docsite/rst/galaxy.rst
```diff
@@ -129,48 +128,100 @@ And here's an example showing some specific version downloads from multiple sour
...
+    - include: /etc/roles/web_server_roles.yml
+    - include: /var/roles/security.yml
+
+    # Additional roles
+    - src: geerlingguy.apache
+    - src: https://github.com/some_org/install_me.git
+      version: 1.2.0
+
+Muliple roles from a single repo 
+================================
```
bcoca: do we really want to support this?

bcoca:
We should also mention that the 'galaxy server' is configurable in case they want to host their own.

docsite/rst/galaxy.rst
chouseknecht: Good point. Updated.

docsite/rst/galaxy.rst
chouseknecht: Yes. References to _requirements.txt_ have been removed.

docsite/rst/galaxy.rst
chouseknecht: Nope. Removed.

docsite/rst/galaxy.rst
chouseknecht: It looks to me like it is implemented: https://github.com/ansible/ansible/blob/devel/lib/ansible/cli/galaxy.py#L358

I have not actually tested, so maybe it's not fully implemented?

chouseknecht: Removed references to _includes_ for now. We can add it later, when/if it's fully formed and tested.

PR #18112, opened 2016-10-19
Use six.move for module in module_utils/facts.py

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

setup


abadger:
Merged to devel and stable-2.2.  This will be in either 2.2.0 or 2.2.1 depending on whether we have a 2.2.0rc3.

PR #18115, opened 2016-10-20
Resubmitting Approved Pull Request #17571

june07:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

module_utils/lxd.py
affected lxd_container module
##### ANSIBLE VERSION

```
ansible 2.2.0

```
##### SUMMARY

When the client certificate is already stored, lxd returns a JSON error with message "Certificate already in trust store".  This "error" will occur on every task run after the initial run.  The cert should be in the trust store after the first run and this error message should really only be viewed as informational as it does not indicate a real problem.  I certainly don't think an exception should be being thrown?!

Fixes:
https://github.com/ansible/ansible-modules-extras/issues/2750

```

```


abadger:
This got maintainer approval in https://github.com/ansible/ansible/pull/17571  I'm merging it to devel and cherry-picking to stable-2.2.  We're in the late stages of the rc process for 2.2.0 so this might not make it in until 2.2.1 depending on whether we have another rc for 2.2.0 or go from here to final.

PR #18118, opened 2016-10-20
Add tunnel linux

mscherer:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

setup

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -2260,6 +2260,11 @@ class LinuxNetwork(Network):
     - ipv4_address and ipv6_address: the first non-local address for each family.
     """
     platform = 'Linux'
+    INTERFACE_TYPE = {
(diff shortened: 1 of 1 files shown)


lib/ansible/module_utils/facts.py
```diff
@@ -2221,6 +2221,12 @@ class LinuxNetwork(Network):
     - ipv4_address and ipv6_address: the first non-local address for each family.
     """
     platform = 'Linux'
+    _interface_type = {
```
alikins: looks like other similar maps in facts tends to use constant style naming
(ie, INTERFACE_TYPE).

alikins (changes requested):
aside from the var name style, LGTM

mscherer:
@alikins  I did the change. Not sure if it would be polite to dismiss the review

PR #18122, opened 2016-10-20
Refactor code for VirtualFacts

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

setup
##### SUMMARY

The populate method is cut and paste on every subclass,
so we should push it up, and add a default method that is overloaded

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -2999,7 +2999,12 @@ class Virtual(Facts):
             return super(cls, subclass).__new__(subclass, *arguments, **keyword)
 
     def populate(self):
+        self.get_virtual_facts()
         return self.facts
+
+    def get_virtual_facts(self):
+        self.facts['virtualization_type'] = ''
+        self.facts['virtualization_role'] = ''
 
 class LinuxVirtual(Virtual):
     """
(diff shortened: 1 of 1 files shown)


alikins:
Wondering if the default should be to not set those keys, set them to '', set them to 'NA' like LinuxVirtual, or other...

lib/ansible/module_utils/facts.py
```diff
@@ -2999,8 +2999,13 @@ def __new__(cls, *arguments, **keyword):
             return super(cls, subclass).__new__(subclass, *arguments, **keyword)
 
     def populate(self):
+        self.get_virtual_facts()
         return self.facts
 
+    def get_virtual_facts(self):
+        self.facts['virtualization_type'] = ''
+        self.facts['virtualization_role'] = ''
```
alikins, line 3007: > bcoca> alikins: in most cases we don't add keys when they are not detected, not just virtualization (which we also need to change into lists instead of fixed values)
> <bcoca> ^ we either change the behaviour to always populate all keys or keep as is, we should not change just for one set of keys
> <bcoca> it would be nice to be consistent one way or the other

Going by that, I'd suggest the base get_virtual_facts is

``` python
def get_virtual_facts(self):
    pass
```

But as you mention, that would change behavior. So pr as is LGTM to me, we can change the behavior later.

PR #18128, opened 2016-10-20
group_vars: host_vars: warn for non-yaml files in dirs

### LINKED ISSUE #17968: Extensions of variable files are not limited as they should be per documentation. Backup copies thus may take precedence over expected files
##### ISSUE TYPE
- Bug Report
##### COMPONENT NAME

initialization
##### ANSIBLE VERSION

```
ansible 2.1.2.0
```
##### OS / ENVIRONMENT

Host: Ubuntu 14.01
Guest: Scientific Linux release 7.2 (Nitrogen)
##### SUMMARY

Variables loaded from ./group_vars/all/config.yml~ wins over ./group_vars/all/config.yml
According to [docs](http://docs.ansible.com/ansible/intro_inventory.html#splitting-out-host-and-group-specific-data) this should not be allowed: 

> ....Valid file extensions include ‘.yml’, ‘.yaml’, ‘.json’, or no file extension
> And by "include" it means included in a list and not should include a sub-string in an extension.
##### STEPS TO REPRODUCE
1. Create ./group_vars/all/config.yml file with simple content

```

---
varrr: proper_value
```
1. Copy file and concat with tilde at the end of the filename ./group_vars/all/config.yml~ 
   . Change content

```

---
varrr: incorrect_value
```
1. Run simple task to print variable

```
- name: Print varrr
  debug: var=varrr
```
##### EXPECTED RESULTS

Expected content from ./group_vars/all/config.yml
##### ACTUAL RESULTS

Got content from ./group_vars/all/config.yml~

resmo:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

host_vars, group_vars
##### ANSIBLE VERSION

```
devel, 2.2, 2.1
```
##### SUMMARY

fixes #17968 
~~ignore~~ warn for files without a known yaml extension in group_/host_vars if directories like group_vars/group/... are used.

### DIFF
--- a/lib/ansible/vars/__init__.py
+++ b/lib/ansible/vars/__init__.py
@@ -557,13 +557,17 @@ class VariableManager:
         else:
             file_name, ext = os.path.splitext(path)
             data = None
-            if not ext:
+            if not ext or ext not in C.YAML_FILENAME_EXTENSIONS:
+                # Warn for non yaml extensions or backup files
+                if ext or file_name.endswith('~'):
+                    display.warning("Loaded vars in %s although the vars file has not proper name or valid extension" % (path,))
+
                 for test_ext in C.YAML_FILENAME_EXTENSIONS:
                     new_path = path + test_ext
                     if loader.path_exists(new_path):
                         data = loader.load_from_file(new_path)
                         break
-            elif ext in C.YAML_FILENAME_EXTENSIONS:
+            else:
                 if loader.path_exists(path):
                     data = loader.load_from_file(path)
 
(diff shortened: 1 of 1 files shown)


jhawkesworth:
I don't think the documentation is consistent about using extensions for host_vars / group_vars.
For example here on the best practices page:
http://docs.ansible.com/ansible/playbooks_best_practices.html#variables-and-vaults

This change would mean I would have to change all the files in my group_vars folders.

Perhaps a warning if backup file names, say those containing ~ are found in group vars / host vars would be a practical compromise?

resmo:
I don't see any conflicts with the best practice guide,"...create two files named vars and vault...", that still works as these files have no extension.

But I would say this commit has some potential to break certain setups. A compromise of adding a warning would be an option of course.

bjolivot:
bot_status

bcoca:
this does not seem like the right fix, it seems to compensate for the extension filter being broken, we shoudl fix that instead.

PR #18129, opened 2016-10-20
Now that we convert salt inside of do_encryptas needed, keep salt as text_type until then

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

do_encrypt (called in lookup/password.py implemented in utils/encrypt.py)
##### ANSIBLE VERSION

```
devel 2.2
```


abadger:
Merged t odevel and cherry-picked to stable-2.2

PR #18134, opened 2016-10-21
fix hosts.yml filename for inventory

gmella:
##### ISSUE TYPE

- Docs Pull Request
##### ANSIBLE VERSION

```

```
##### SUMMARY

```
Fix typo?
and add comment trying to explain that directory can be used as -i argument values.
```

### DIFF
--- a/docsite/rst/playbooks_best_practices.rst
+++ b/docsite/rst/playbooks_best_practices.rst
@@ -27,8 +27,8 @@ Directory Layout
 
 The top level of the directory would contain files and directories like so::
 
-    production                # inventory file for production servers
-    staging                   # inventory file for staging environment 
+    production                # inventory file for production servers (or directory that contains hosts.yml)
+    staging                   # inventory file for staging environment (or directory that contains hosts.yml)
 
     group_vars/
        group1                 # here we assign variables to particular groups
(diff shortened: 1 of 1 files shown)


docsite/rst/playbooks_best_practices.rst
```diff
@@ -27,8 +27,8 @@ Directory Layout
 
 The top level of the directory would contain files and directories like so::
 
-    production                # inventory file for production servers
-    staging                   # inventory file for staging environment 
+    production                # inventory file for production servers (or directory that contains hosts.yml)
+    staging                   # inventory file for staging environment (or directory that contains hosts.yml)
```
bcoca: if it is a directory it won't pick up group_vars, we use a 2nd example with directories to reflect this

gmella: I'm still unclear about switching from production to staging easily. I just propose it because it seems coded in this way for hosts.yml file. I don't understand yet the rules for group_vars...
Anyway, that is fine to cancel this part of my PR. Maybe a playbook command line sample could make it more clear for everybody ?
Cheers,

dharmabumstead:
Thanks @gmella!

PR #18135, opened 2016-10-21
Restructure .gitignore for readability

Damgaard:
##### ISSUE TYPE

- Docs Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.2.0
  config file = /home/andreas/.ansible.cfg
```
##### SUMMARY

Restructure the `.gitignore` file to be readable.

It has been rebuilt into a 3 level hirachical structure. With
the area highest, then grouping then final tool.

All entries within a grouping were sorted alphabetically.

During the restructure ignores that were already in a caregory and
could fit that category were kept in that category. Even if I
disagreed with their placements. Ignores that I was uncertain on
were also kept in their only original category, only files that
were clearly misplaced or where their original category were split
was moved.

As an example credentials.yml is not a vim swap file and so it was
moved.

*.retry I believe to be a ansible retry file and misplaced under
testing, but since I couldn't be sure that it wasn't also a tox
ignore file, it's been kept under tox.

Category names have not been changed either, except for
capitalization of the first word.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@ ansible.egg-info/
 AUTHORS.TXT
 build
 
-# RPM stuff... #
+# RPM #
 dist
 MANIFEST
 rpm-build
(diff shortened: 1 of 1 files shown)


abadger:
A hierarchical organization to the .gitignore is too complex for my tastes.  There's currently 51 entries and 15 header lines.  The new version is 117 lines (some blank, though) for the same number of entries.

If you want to move some files around or add new categories I'd be fine with that.

I'm not the only core dev... we can also leave this as is if you want a second opinion.

bcoca:
I also think the current is readable 'as is' and the proposed one is less so.

abadger:
# Question About This One

Hi!

   *We think that the .gitignore file is more readable with a single level of categorization than a deeply nested hierarchy.  We would be open to adding/removing categories and moving files around amongst them if you'd like to submit a PR with a more limited scope.

PR #18147, opened 2016-10-22
Configurable fact path

alvaroaleman:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

`play_iterator.py/play.py/constants.py`
##### ANSIBLE VERSION

```
devel
```
##### SUMMARY

This change allows to configure a fact_path via playbook paramter, ansible.cfg or environment variable.

This is necessary to allow using local facts on systems without root access.

### DIFF
--- a/lib/ansible/executor/play_iterator.py
+++ b/lib/ansible/executor/play_iterator.py
@@ -163,9 +163,9 @@ class PlayIterator:
         self._task_uuid_cache = dict()
 
         # Default options to gather
-        gather_subset = C.DEFAULT_GATHER_SUBSET
-        gather_timeout = C.DEFAULT_GATHER_TIMEOUT
-        fact_path = C.DEFAULT_FACT_PATH
+        gather_subset = play_context.gather_subset
+        gather_timeout = play_context.gather_timeout
+        fact_path = play_context.fact_path
 
         # Retrieve subset to gather
         if self._play.gather_subset is not None:
--- a/lib/ansible/executor/play_iterator.py
+++ b/lib/ansible/executor/play_iterator.py
@@ -165,6 +165,7 @@ class PlayIterator:
         # Default options to gather
         gather_subset = C.DEFAULT_GATHER_SUBSET
         gather_timeout = C.DEFAULT_GATHER_TIMEOUT
+        fact_path = C.DEFAULT_FACT_PATH
 
         # Retrieve subset to gather
         if self._play.gather_subset is not None:
@@ -172,6 +173,9 @@ class PlayIterator:
         # Retrieve timeout for gather
         if self._play.gather_timeout is not None:
             gather_timeout = self._play.gather_timeout
+        # Retrieve fact_path
+        if self._play.fact_path is not None:
+            fact_path = self._play.fact_path
 
         setup_block = Block(play=self._play)
         setup_task = Task(block=setup_block)
@@ -183,6 +187,8 @@ class PlayIterator:
         }
         if gather_timeout:
             setup_task.args['gather_timeout'] = gather_timeout
+        if fact_path:
+            setup_task.args['fact_path'] = fact_path
         setup_task.set_loader(self._play._loader)
         setup_block.block = [setup_task]
 
--- a/test/integration/targets/gathering_facts/test_gathering_facts.yml
+++ b/test/integration/targets/gathering_facts/test_gathering_facts.yml
@@ -129,3 +129,48 @@
           - 'ansible_interfaces|default("UNDEF_NET") != "UNDEF_NET"'
           - 'ansible_mounts|default("UNDEF_HW") == "UNDEF_HW"'
           - 'ansible_virtualization_role|default("UNDEF_VIRT") == "UNDEF_VIRT"'
+
+- hosts: facthost9
+  tags: [ 'fact_local']
+  connection: local
+  gather_facts: no
+  tasks:
+    - name: Create fact directories
+      become: true
+      with_items:
+        - /etc/ansible/facts.d
+        - /tmp/custom_facts.d
+      file:
+        state: directory
+        path: "{{ item }}"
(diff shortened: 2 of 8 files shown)


alvaroaleman:
Rebased PR to reflect current devel.

docsite/rst/playbooks_variables.rst
```diff
@@ -450,6 +450,7 @@ For instance, what if you want users to be able to control some aspect about how
 
 If a remotely managed system has an ``/etc/ansible/facts.d`` directory, any files in this directory
 ending in ``.fact``, can be JSON, INI, or executable files returning JSON, and these can supply local facts in Ansible.
+An altenate directory can be specified using the ``fact_path`` play variable.
```
mattclay: Typo in `alternate`.

docsite/rst/intro_configuration.rst
```diff
@@ -386,6 +386,17 @@ is very very conservative::
 
     forks = 5
 
+.. _fact_path:
+
+fact_path
+=========
+
+This option allows to configure a custom path for local facts::
```
mattclay: This should read `allows you to configure` (the word `you` is missing).

mattclay (changes requested):

test/integration/targets/gathering_facts/test_gathering_facts.yml
```diff
@@ -129,3 +129,48 @@
           - 'ansible_interfaces|default("UNDEF_NET") != "UNDEF_NET"'
           - 'ansible_mounts|default("UNDEF_HW") == "UNDEF_HW"'
           - 'ansible_virtualization_role|default("UNDEF_VIRT") == "UNDEF_VIRT"'
+
+- hosts: facthost9
```
mattclay, line 133: There is no `facthost9`, so these tests are not actually running, as seen in the CI logs here:

https://app.shippable.com/runs/585c33c6666b4b1000d0575d/33/console

alvaroaleman: Added a facthost9 to fix that :)

alvaroaleman:
I don't think the needs_revision label fits, since I've fixed everything mentioned by @mattclay

mattclay (approved):

docsite/rst/intro_configuration.rst
```diff
@@ -386,6 +386,17 @@ is very very conservative::
 
     forks = 5
 
+.. _fact_path:
+
+fact_path
+=========
+
+This option allows you to configure a custom path for local facts::
```
bcoca: i would make note that this is 'globally'

bcoca: also link to 'local fact' docs

docsite/rst/playbooks_variables.rst
```diff
@@ -450,6 +450,7 @@ For instance, what if you want users to be able to control some aspect about how
 
 If a remotely managed system has an ``/etc/ansible/facts.d`` directory, any files in this directory
 ending in ``.fact``, can be JSON, INI, or executable files returning JSON, and these can supply local facts in Ansible.
+An alternate directory can be specified using the ``fact_path`` play variable.
```
bcoca: if implemented in play its a directive (or keyword), not a variable

lib/ansible/executor/play_iterator.py
```diff
@@ -181,6 +185,8 @@ def __init__(self, inventory, play, play_context, variable_manager, all_vars, st
         }
         if gather_timeout:
             setup_task.args['gather_timeout'] = gather_timeout
+        if fact_path:
+            setup_task.args['fact_path'] = fact_path
```
bcoca, line 191: a better way is to just have play default to the config setting

lib/ansible/executor/play_iterator.py
```diff
@@ -181,6 +181,8 @@ def __init__(self, inventory, play, play_context, variable_manager, all_vars, st
         }
         if gather_timeout:
             setup_task.args['gather_timeout'] = gather_timeout
+        if self._play.fact_path:
```
alvaroaleman: @bcoca: Regarding your comment 
> a better way is to just have play default to the config setting

Do you mean the setting from `constants.py`?

If yes: `fact_path` is None by default, because the actual default comes from the setup module and I didn't want to duplicate that.

lib/ansible/executor/play_iterator.py
alvaroaleman: Do you mean the setting from constants.py?

If yes: fact_path is None by default, because the actual default comes from the setup module and I didn't want to duplicate that.

bcoca: sorry, my comment might be misleadingly placed, i'm not against the 'if/task.args' update, but on how you inject the default here, just use normal play object inheritance upto this point.

Possibly the best place to 'resolve' the default is the task after it checks parents for inheritance. We don't have a clear rule, for example have strategy set this way, but most other defaults are populated in play_context. @jimi-c any opinions on this?

alvaroaleman: Ah okay, I get your point now. That totally makes sense, but would make it inconsistent with the non-inherited defaults for `gather_subset` and `gather_timeout`

bcoca: my reasoning applies to those as well, this is not specific to your change.

alvaroaleman: Can I go ahead and move the defaulting to play_context or should I wait on an opinion on Play vs PlayContext by @jimi-c?

bcoca: go ahead, we might even merge as is and revamp later to have a consistent path for these types of configs

alvaroaleman: done

PR #18153, opened 2016-10-22
Fix 18151, by converting float to int

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

facts


abadger:
Better way to fix this is to use python's floor division operator:

``` python
int(val) // 1024
```

We might want to make 3/4 == 0.75 in all of our code by using:

```
from __future__ import division
```

in module code as well as controller code.  I hadn't thought about that before because the other `from __future__` imports that we use require python-2.6+ but `__future__.division` was introduced in python-2.2 so it will be available everywhere we care about.

abadger:
The only failure was a timeout on opensuseleap.  This looks good so merging to devel and stable-2.2.  It will be in 2.2.0.

PR #18165, opened 2016-10-24
Build debs with pbuilder

shanemcd:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

Debian packaging
##### ANSIBLE VERSION

N/A
##### SUMMARY

We started seeing failures related to build dependencies when building 16.04 debs on our 14.04 Jenkins node. `pbuilder` ensures a clean-room environment inside a chroot.

### DIFF
--- a/packaging/debian/Dockerfile
+++ b/packaging/debian/Dockerfile
@@ -1,0 +1,16 @@
+FROM ubuntu:xenial
+
+RUN apt-get update && apt-get install -y \
+    asciidoc \
+    cdbs \
+    debootstrap \
+    devscripts \
+    make \
+    pbuilder \
+    python-setuptools
+
+VOLUME /ansible
+WORKDIR /ansible
+
+ENTRYPOINT ["/bin/bash", "-c"]
+CMD ["make deb"]
--- a/packaging/debian/README.md
+++ b/packaging/debian/README.md
@@ -13,6 +13,16 @@ git submodule update --init
 DEB_DIST='xenial trusty precise' make deb
 ```
 
+Building in Docker:
+
+```
+git clone git://github.com/ansible/ansible.git
+cd ansible
+git submodule update --init
+docker build -t ansible-deb-builder -f packaging/debian/Dockerfile .
+docker run --privileged -e DEB_DIST='trusty' -v $(pwd):/ansible ansible-deb-builder
+```
+
 The debian package file will be placed in the `deb-build` directory. This can then be added to an APT repository or installed with `dpkg -i <package-file>`.
 
 Note that `dpkg -i` does not resolve dependencies.
--- a/packaging/debian/README.md
+++ b/packaging/debian/README.md
@@ -3,26 +3,29 @@ Ansible Debian Package
 
 To create an Ansible DEB package:
 
-    sudo apt-get install python-paramiko python-yaml python-jinja2 python-httplib2 python-setuptools python-six sshpass
-    sudo apt-get install cdbs debhelper git-core reprepro fakeroot asciidoc devscripts docbook-xml xsltproc libxml2-utils
-    sudo apt-get install dh-python build-essential
-    git clone git://github.com/ansible/ansible.git
-    cd ansible
-    make deb
+__Note__: You must run this target as root or set `PBUILDER_BIN='sudo pbuilder'`
 
-On older releases that do not have `dh-python` (like Ubuntu 12.04), install `python-support` instead:
+```
+apt-get install asciidoc cdbs debootstrap devscripts make pbuilder python-setuptools
+git clone git://github.com/ansible/ansible.git
+cd ansible
+git submodule update --init
+DEB_DIST='xenial trusty precise' make deb
+```
 
-    sudo apt-get install python-support
-
-The debian package file will be placed in the `../` directory. This can then be added to an APT repository or installed with `dpkg -i <package-file>`.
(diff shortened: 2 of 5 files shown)


jlaska:
Thanks for updating the README's too, this is a welcome change to use `pbuilder`!

bcoca:
So i would not eliminate the old way of building it, just add a new one. some users are stuck in older versions of debian and/or do not use pbuilder. It is easy enough to have this 'in addition to'

jlaska:
@bcoca normally I would agree, however the old way of building the DEBs produces DEBs _only_ for the platform you are building on.  In addition, there was logic in there that resulted in failed builds depending on the state of packages installed on the system.  Chroot builds is much preferred to provide consistent builds and accurately test `BuildRequires`.

ghjm:
`DEB_DIST='precise trusty xenial' make deb` worked on my Ubuntu 14.04 box with pbuilder and build deps installed. Looks good to me.

shanemcd:
@bcoca Great suggestion. I've added a `local_deb` target to allow for building outside of a chroot.

packaging/debian/README.md
```diff
@@ -3,26 +3,35 @@ Ansible Debian Package
...
-    sudo apt-get install python-paramiko python-yaml python-jinja2 python-httplib2 python-setuptools python-six sshpass
-    sudo apt-get install cdbs debhelper git-core reprepro fakeroot asciidoc devscripts docbook-xml xsltproc libxml2-utils
-    sudo apt-get install dh-python build-essential
-    git clone git://github.com/ansible/ansible.git
-    cd ansible
-    make deb
+```
+apt-get install cdbs debootstrap devscripts make pbuilder python-setuptools
+git clone git://github.com/ansible/ansible.git
+cd ansible
```
mattclay: Add `git submodule update --init` here.

packaging/debian/README.md
```diff
@@ -3,26 +3,35 @@ Ansible Debian Package
...
+DEB_DIST='xenial trusty precise' make deb
+```
 
-On older releases that do not have `dh-python` (like Ubuntu 12.04), install `python-support` instead:
+Building in Docker:
 
-    sudo apt-get install python-support
+```
+git clone git://github.com/ansible/ansible.git
+cd ansible
```
mattclay: Add `git submodule update --init` here.

packaging/debian/README.md
```diff
@@ -3,26 +3,35 @@ Ansible Debian Package
...
 To create an Ansible DEB package:
 
-    sudo apt-get install python-paramiko python-yaml python-jinja2 python-httplib2 python-setuptools python-six sshpass
-    sudo apt-get install cdbs debhelper git-core reprepro fakeroot asciidoc devscripts docbook-xml xsltproc libxml2-utils
-    sudo apt-get install dh-python build-essential
-    git clone git://github.com/ansible/ansible.git
-    cd ansible
-    make deb
+```
+apt-get install cdbs debootstrap devscripts make pbuilder python-setuptools
```
mattclay: Additional packages needed on a fresh install of Ubuntu 16.04:
- git
- asciidoc

packaging/debian/README.md
```diff
@@ -3,26 +3,35 @@ Ansible Debian Package
 
 To create an Ansible DEB package:
```
mattclay, line 4: It should be noted that running as root is required.

mattclay (changes requested):
I've successfully tested this using the provided instructions for docker.

However, I'm unable to produce a deb package running in a fresh Ubuntu 16.04 VM. The build process finishes without errors, but no deb package is produced.

packaging/debian/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM ubuntu:xenial
+
+RUN apt-get update && apt-get install -y \
+    cdbs \
```
mattclay: The `asciidoc` package is needed.

shanemcd:
@mattclay Thanks for the review. I've implemented your feedback, and am currently debugging on a fresh 16.04 VM to diagnose:

> The build process finishes without errors, but no deb package is produced.

For some reason, it is placing the deb in a top-level directory at `/deb-build` when running with sudo, but works when running directly as root. It works fine when running on a 14.04 box. ¯_(ツ)_/¯ Will report back when I find something.

shanemcd:
@mattclay I've pushed a fix that works on a fresh 16.04 box. It works when using GNU Make's `CURDIR` variable instead of `PWD`. Also tested with the Docker based approach.

mattclay (approved):

packaging/debian/README.md
```diff
@@ -3,26 +3,39 @@ Ansible Debian Package
...
 
-    sudo apt-get install python-paramiko python-yaml python-jinja2 python-httplib2 python-setuptools python-six sshpass
-    sudo apt-get install cdbs debhelper git-core reprepro fakeroot asciidoc devscripts docbook-xml xsltproc libxml2-utils
-    sudo apt-get install dh-python build-essential
-    git clone git://github.com/ansible/ansible.git
-    cd ansible
-    make deb
+__Note__: You must either run as root or have write access to `/var/cache/pbuilder`
```
mattclay: @shanemcd Write access to `/var/cache/pbuilder` isn't enough. The build process fails with this error when not running as root:

```
E: debootstrap can only run as root
```

bcoca:
IIRC pbuilder should be doing fakeroot by default

PR #18168, opened 2016-10-24
Fix fact gathering intg test asserts

alikins:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

test/integration/targets/gathering_facts
##### ANSIBLE VERSION

ansible 2.3.0 (fix_gathering_facts_test_quoting 7d5ee87df2) last updated 2016/10/24 15:57:31 (GMT -400)
  lib/ansible/modules/core: (detached HEAD 124bb92416) last updated 2016/10/24 11:26:36 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 8ffe314ea5) last updated 2016/10/24 11:26:36 (GMT -400)

```

```
##### SUMMARY

If the facts returned by setup included strings that
had double quotes in them, the asserts in test_gathering_facts.yml
would fail with errors like:

```
"The conditional check '\"[{u'mounts': {u'options':
u'rw,context=\"system_u:\"'}}]\" != \"UNDEF_HW\"' failed. The error was:
template error while templating string: expected token 'end of statement
block', got 'system_u'. String: {% if \"[{u'mounts': {u'options':
u'rw,context=\"system_u:\"'}}]\" != \"UNDEF_HW\" %} True {% else %}
False {% endif %}"
```

For one example, if mount facts returned an 'options' field that
included double quoated selinux context ids, the test would fail.

Fix is removing the double quoting in the assert 'that:' lines.

### DIFF
--- a/test/integration/targets/gathering_facts/test_gathering_facts.yml
+++ b/test/integration/targets/gathering_facts/test_gathering_facts.yml
@@ -8,14 +8,13 @@
   tasks:
     - setup:
       register: facts
-    - debug: var=facts
     - name: Test that retrieving all facts works
       assert:
         that:
-          - '"{{ ansible_user_id|default("UNDEF_MIN") }}" != "UNDEF_MIN"'
-          - '"{{ ansible_interfaces|default("UNDEF_NET") }}" != "UNDEF_NET"'
-          - '"{{ ansible_mounts|default("UNDEF_HW") }}" != "UNDEF_HW"'
-          - '"{{ ansible_virtualization_role|default("UNDEF_VIRT") }}" != "UNDEF_VIRT"'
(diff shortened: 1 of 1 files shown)


abadger:
Interesting.  I experimented with this a bit and then looked at the code for assert.  Found out that the curly braces aren't needed at all because the assert module runs each item from the list given to that through templating itself.

```
---
- hosts: localhost
  gather_facts: False
  vars:
    test: 'this is a" test case'
    test2: this is a" test case
    test3: this is a" test case2
  tasks:
  - assert:
      that:
        - "test == test2"
  - assert:
      that:
        - "test == test3"
```

First assert succeeds, second one fails.

No need to take advantage of that now, though.  +1 to merge this.

bcoca:
conditionals don't need mustaches!

PR #18171, opened 2016-10-24
Make executable setup.py with execution rights

ghost:
##### ISSUE TYPE
- Feature Pull Request
##### SUMMARY

setup.py is an executable, so it is supposed to be executable. Simple as that !


abadger:
Since there's a shebang this makes sense but in the modern era, the recommendation from distutils upstream is that people should invoke setup.py like: /usr/bin/python2 setup.py. That way, if they want to install for python3, they can slo do /usr/bin/python3 setup.py.

So perhaps it makes more sense to remove the shebang from setup.py?

ghost:
I understand your point, which I find that make a lot of sense. I prepare the commit over that pull request at the moment.

abadger:
Thanks @dcotet The new commit includes a commit of the ansible/modules/core submodule, though.  Can you remove that?  Something like:
``
git reset --mixed 6dccd4b9ff471067bdfa4520b984e2a525010083
git commit setup.py
git push -f

```
```

ghost:
It is done !
Thank you very much and have a great day !

abadger:
Resolved conflicts and merged to devel here: 221cafe769190f5a47950b02e54b491ae84de4cc

PR #18174, opened 2016-10-25
Update winrm to allow fetch_file to work on files in use by processes

cj-dimaggio:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

connection.winrm
##### ANSIBLE VERSION

```
Current 2.3.0 but goes back several versions as well.
```
##### SUMMARY

WinRM's fetch_file method currently calls "[System.IO.File]::OpenRead" which opens up the file not only in read mode but also [attempts to set the FileShare mode to "Read" only](https://msdn.microsoft.com/en-us/library/system.io.file.openread%28v=vs.110%29.aspx#Anchor_2) which raises the error "The process cannot access the file [path] because it is being used by another process." if the file is already open for write access by another program (a log file for instance). This patch simply calls the IO.FileStream constructor explicitly so we can set the ShareMode to ReadWrite and thus allows us to transfer files even if they're open by another program. Further information about this kind of CLR error can be found at http://coding.infoconex.com/post/2009/04/21/How-do-I-open-a-file-that-is-in-use-in-C and http://stackoverflow.com/questions/12942717/read-log-file-being-used-by-another-process 

```

```

### DIFF
--- a/lib/ansible/plugins/connection/winrm.py
+++ b/lib/ansible/plugins/connection/winrm.py
@@ -385,7 +385,7 @@ class Connection(ConnectionBase):
                     script = '''
                         If (Test-Path -PathType Leaf "%(path)s")
                         {
-                            $stream = [System.IO.File]::OpenRead("%(path)s");
+                            $stream = New-Object IO.FileStream("%(path)s", [System.IO.FileMode]::Open, [System.IO.FileAccess]::Read, [IO.FileShare]::ReadWrite);
                             $stream.Seek(%(offset)d, [System.IO.SeekOrigin]::Begin) | Out-Null;
                             $buffer = New-Object Byte[] %(buffer_size)d;
                             $bytesRead = $stream.Read($buffer, 0, %(buffer_size)d);


jhawkesworth (commented):
It seems like a good idea to add this.  Off the top of my head I cannot think of any existing functionality that would need to work as this does now. I know there is a windows wait_for module in the works but I haven't seen a pr for it get, so no idea if it might use fetch_file logic when waiting for files.  I guess it wouldn't be much work to support fetching with or without locking in future if needed.

Do you have a test playbook that others could use to try this out?

cj-dimaggio:
Ah, I hadn't thought about locking the file potentially being a feature. In any case, here's an example playbook of a scenario that doesn't work with the current implementation but does work with this change. (The playbook spins up a scheduled task to simulate an application that opens up a file with ReadWrite share and tries to clean it up afterwords but you still might want to scrub it yourself)

```
- hosts: all
  name: Test winrm connection's fetch_file with open files
  tasks:
    - set_fact:
         temp_file: C:\Users\{{ ansible_user }}\AppData\Local\Temp\foo.log

    - name: Create test file
      win_file:
        path: "{{ temp_file }}"
        state: touch

    - name: put some content in the test file for validation purposes
      win_lineinfile:
        dest: "{{ temp_file }}"
        line: "Hello, World"

    # The win_scheduled_task module doesn't work too good
    - name: Remove existing scheduled task if it exists
      raw: if (Get-ScheduledTask ansible_winrm_test -ErrorAction SilentlyContinue) { Unregister-ScheduledTask ansible_winrm_test -confirm:$false}

    - name: Create scheduled task so we can create a job that exists after this connection exits
      raw: Register-ScheduledTask -TaskName ansible_winrm_test -Action (New-ScheduledTaskAction -Execute powershell.exe -Argument "`$stream = New-Object IO.FileStream('{{ temp_file }}', [System.IO.FileMode]::Open, [System.IO.FileAccess]::Write, [IO.FileShare]::ReadWrite); Start-Sleep -s 100");

    - name: Start our scheduled task
      raw: Start-ScheduledTask ansible_winrm_test

    - name: Try to fetch the open file
      fetch:
        src: "{{ temp_file }}"
        dest: /tmp/ansible_winrm_test.log
        flat: yes
      ignore_errors: yes


    - name: Cleanup scheduled task
      raw: Stop-ScheduledTask ansible_winrm_test; Unregister-ScheduledTask ansible_winrm_test -confirm:$false
```

rogerza:
+1 for this update

nitzmahone (approved):

nitzmahone:
Thanks!

PR #18175, opened 2016-10-25
A few fixes for python3

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

module_utils/urls.py
##### ANSIBLE VERSION

```
devel, 2.2.1
```
##### SUMMARY
- socket interfaces take bytes so convert text strings to bytes when
  using them.
- Use b64encode() instead of str.encode('base64')

### DIFF
--- a/lib/ansible/module_utils/urls.py
+++ b/lib/ansible/module_utils/urls.py
@@ -105,8 +105,6 @@ import platform
 import tempfile
 import base64
 
-from ansible.module_utils.basic import get_distribution, get_exception
-
 try:
     import httplib
 except ImportError:
@@ -115,7 +113,9 @@ except ImportError:
 
 import ansible.module_utils.six.moves.urllib.request as urllib_request
 import ansible.module_utils.six.moves.urllib.error as urllib_error
+from ansible.module_utils.basic import get_distribution, get_exception
 from ansible.module_utils.six import b
+from ansible.module_utils._text import to_bytes
 
 try:
     # python3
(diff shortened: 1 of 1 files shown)


abadger:
@Foxlik here's the changes to make this code python3 compatible (at least potentially... could definitely be other bugs hiding in there).  If you'd be willing to test that it works with your https proxy that would be great.  I don't need to have it tested with python3 unless you have time for that.  I mainly want to make sure that python2 support doesn't regress as I try to fix the code for python3.  Thanks!

abadger:
Note to who merges this: it depends on https://github.com/ansible/ansible/pull/18083 which is presently only merged to devel (which will be 2.3.x).  If we want this in the 2.2.x series we'll need to cherry-pick #18083 as well.

mscherer:
Would it have a effect on test_uri failure ?
https://github.com/ansible/ansible/pull/18060 
(I guess that's also why we may not have detect the issue sooner )

mscherer:
Ok so answering to myself,  no, unlikely to impact the test, but the lack of those test is why we didn't see the issue.

abadger:
@Foxlik Looks like we're going to have a 2.2.0rc4 on Thursday.  If you can test this before then I can push your change and this one to 2.2.0.  Otherwise it will probably need to wait until either 2.2.1 or 2.3.0.

nitzmahone:
Tests fine for me against an an https endpoint via an authenticating HTTPS proxy on both Python 2.7.10 and 3.4.3. Fails with 407 when auth is not provided to proxy, succeeds when auth is provided. Anything else you want to see?

Foxlik:
@abadger unfortunately I've been fixing this for my colleague on an isolated test case only. For that your commit works fine. But to test it in the real world myself, I need our lab, which is currently broken. 😭 It should be fixed this week, but I can give no guarantees. 

It's easy to test with two tinyproxy instances though. Using one as the upstream for another. That's how we found out about the issue in the first place. (using one squid as an upstream for another squid was fine even before the patch as it sends headers in one packet/fast enough)

abadger:
Alright.  I think I've got enough testing to feel confident about this. @Foxlik 's patch is doing the fixing for the problem cases.  My changes should either fail or work.  So with both nitzmahone and foxlik confirming the combination works I'll push this to stable-2.2

abadger:
Merged to devel and cherry-picked to stable-2.2.  Should be in 2.2.0 when it is released.

abadger:
Thanks @Foxlik For the previous fix and for testing this!

Foxlik:
Thanks for getting it in so fast it makes it into 2.2. My colleagues are going to be happy ;-).

PR #18194, opened 2016-10-25
task_queue_manager: fix calc of forks when serial is %

### LINKED ISSUE #18191: serial with % groups task per host
##### ISSUE TYPE

- Bug Report
##### COMPONENT NAME

serial
##### ANSIBLE VERSION

```
devel 
ansible 2.2.0.0 (detached HEAD eafb4043c9) last updated 2016/10/25 13:47:30 (GMT +200)
```
##### CONFIGURATION

n/a
##### OS / ENVIRONMENT

n/a
##### SUMMARY

When using serial with `%`, there is a suspicious grouping per host for every task, while I don't see such a grouping when using serial with a number:
##### STEPS TO REPRODUCE

See gist https://gist.github.com/resmo/c650dc1846c14cdccbc41d509c92f4c0

##### EXPECTED RESULTS

```
changed: [three] => (item=1)
changed: [one] => (item=1)
changed: [two] => (item=1)
changed: [three] => (item=1)
changed: [one] => (item=1)
changed: [two] => (item=1)
changed: [three] => (item=1)
changed: [one] => (item=1)
changed: [two] => (item=1)
changed: [three] => (item=1)
changed: [one] => (item=1)
changed: [two] => (item=1)

```
##### ACTUAL RESULTS

```
changed: [two] => (item=1)
changed: [two] => (item=1)
changed: [two] => (item=1)
changed: [two] => (item=1)
changed: [one] => (item=1)
changed: [one] => (item=1)
changed: [one] => (item=1)
changed: [one] => (item=1)
changed: [three] => (item=1)
changed: [three] => (item=1)
changed: [three] => (item=1)
changed: [three] => (item=1)
```

resmo:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

task_queue_manager
##### ANSIBLE VERSION

```
2.2, 2.3, devel
```
##### SUMMARY

if serial in % was used, the calculation of forks was incorrect. 

~~Serial can be ignored at this point since len(self._inventory.get_hosts(new_play.hosts)) is identical to amount of hosts with serial applied.~~ len(self._inventory.get_hosts(new_play.hosts, ignore_restrictions=True)) is the correct fix. Thanks @bcoca 

Fixes #18191

### DIFF
--- a/lib/ansible/executor/task_queue_manager.py
+++ b/lib/ansible/executor/task_queue_manager.py
@@ -222,7 +222,7 @@ class TaskQueueManager:
         )
 
         # Fork # of forks, # of hosts or serial, whichever is lowest
-        num_hosts = len(self._inventory.get_hosts(new_play.hosts))
+        num_hosts = len(self._inventory.get_hosts(new_play.hosts, ignore_restrictions=True))
 
         max_serial = 0
         if new_play.serial:


bcoca:
I'm not sure this is correct, get_hosts does not seem to factor serial at all (it shouldn't afaik), the playbook_iterator should but not the inventory itself.

PR #18204, opened 2016-10-26
Port rax dynamic inventory script to Ansible 2.3 get_config()

### LINKED ISSUE #18203: [rax.py] TypeError: get_config() got an unexpected keyword argument 'islist'
### Summary

ansible 2.3 has modified the parameters for `ansible.constants.get_config()`: https://github.com/ansible/ansible/commit/9f6bbf8c2f02cdc70df149095a6623ee0c2c7ba7 so `contrib/inventory/rax.py` is no longer compatible:

```
    File "/var/lib/awx/venv/tower/lib/python2.7/site-packages/awx/plugins/inventory/rax.py", line 459, in <module>
      main()
    File "/var/lib/awx/venv/tower/lib/python2.7/site-packages/awx/plugins/inventory/rax.py", line 450, in main
      regions = setup()
    File "/var/lib/awx/venv/tower/lib/python2.7/site-packages/awx/plugins/inventory/rax.py", line 433, in setup
      islist=True)
  TypeError: get_config() got an unexpected keyword argument 'islist'
```
### Environment
- Ansible 2.3 (`devel`)
### Steps To Reproduce:

Run an inventory update against rackspace using ansible 2.3.

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

contrib/inventory/rax.py
##### ANSIBLE VERSION

```
devel (2.3)
```
##### SUMMARY

Fixes #18203

### DIFF
--- a/contrib/inventory/rax.py
+++ b/contrib/inventory/rax.py
@@ -148,7 +148,6 @@ Examples:
 import os
 import re
 import sys
-import inspect
 import argparse
 import warnings
 import collections
@@ -171,10 +170,6 @@ from time import time
 
 from ansible.constants import get_config, mk_boolean
 
-get_config_args = inspect.getargspec(get_config)
-NEW_GET_CONFIG = False
-if 'value_type' in get_config_args:
-    NEW_GET_CONFIG = True
 
 NON_CALLABLES = (basestring, bool, dict, int, list, type(None))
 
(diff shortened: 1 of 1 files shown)


sivel:
I'm not a huge fan of using inspect, especially in "production" code.  I would much rather seeing this caught as a `TypeError` exception instead.  Or just being opinionated and straight update it to just work with 2.3.

abadger:
@sivel, The other alternative I thought of was to use an ansible.release.**version** check.  Would that be better?

I started off thinking about using TypeError but it seemed more hacky than using inspect if done at a global level to set the flag.  Using TypeError at all of the call sites seemed worse than using a flag (mostly because of nested exceptions.)

Being opinionated is problematic since tower takes a copy of rax.py and bundles it.  So they'll need a version of the script that can work on both.

jlaska:
@sivel @abadger tested and works great here.  Thanks!

abadger:
Merged to devel.  get_config() API changes are only in devel (2.3) so no need to backport to older branches.

PR #18205, opened 2016-10-26
Ansible requires tar

sdodson:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

packaging/rpm
##### ANSIBLE VERSION

devel
##### SUMMARY

We should add tar to the specfile `Requires:` so that the archive module local actions work as expected on hosts that don't previously have tar installed. Usually the case when run inside a container or limited environment. This obviously won't fix problems when tar is running on the remote system.

### DIFF
--- a/packaging/rpm/ansible.spec
+++ b/packaging/rpm/ansible.spec
@@ -76,6 +76,7 @@ Requires: python-six
 %endif
 
 Requires: sshpass
+Requires: tar
 
 %description
 


abadger:

Unfortunately, we're not sure if we want this feature in the program.  Our reasons for this are:
- We don't generally add requirements for modules to the controller-side (in setup.py or in the packaging build files).  The main reason is, as you point out, that the main purpose of modules is to run on remote machines and the packaging deps don't affect that.  If someone is writing a playbook to tar some files on all of their machines, they'll need to make sure tar is installed via their playbooks, not via rpm packaging deps.  Additionally, the module deps are only required if the module is run on the controller.  Many people will never use the archive module on the controller so forcing them to install tar there isn't nice for them.

PR #18208, opened 2016-10-26
Permit to have a action plugin without a empty module file

mscherer:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

ansible
##### SUMMARY

Currently, someone writing a action plugin will also need
to have a empty file in the module path to avoid triggering
the error "no action detected in task.".

This PR verify that either a module or a action plugin correspond to a task, permitting to action plugin without a empty file (in case that's purely a action plugin controller side)

( cc @dmsimard )

### DIFF
--- a/lib/ansible/cli/doc.py
+++ b/lib/ansible/cli/doc.py
@@ -124,6 +124,8 @@ class DocCLI(CLI):
             try:
                 # if the plugin lives in a non-python file (eg, win_X.ps1), require the corresponding python file for docs
                 filename = loader.find_plugin(plugin, mod_type='.py', ignore_deprecated=True)
+                if filename is None and loader == module_loader:
+                    filename = action_loader.find_plugin(module, mod_type='.py')
                 if filename is None:
                     display.warning("%s %s not found in %s\n" % (plugin_type, plugin, DocCLI.print_paths(loader)))
                     continue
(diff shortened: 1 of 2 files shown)


bcoca:
they should really have module file as it is only way to document the plugin

dmsimard:
@bcoca so we need to have an empty module file with just the DOCUMENTATION and EXAMPLE strings?

mscherer:
For a external plugin, yeah. For a private one, that seems rather counter intuitive.

bcoca:
ansible-doc still works with private plugins, its a nice way for them to give docs to users that didn't write them or forgot what they did 3 months ago and what the options mean.

dmsimard:
It requires an extra configuration step, though. That means you do not only need to specify an action_plugins path but also a library path. A bit awkward IMO.

Would it not be possible to embed the doc directly in the action module ? Is there a limitation ?

Sorry if I sound naive, I'm not familiar with this bit of code. Genuine question.

mscherer:
It is possible, here is a fix, so now, if we have a action plugin, ansible-doc will also be able to get doc from the plugin. (code could surely be refactored a bit, but just to see if we agree on the basics)

mattclay:
CI is failing nearly all test jobs with the error:

```
2017-02-01 23:47:13 ERROR! conflicting action statements
```

Full logs for one of the failed jobs here: https://app.shippable.com/runs/58926c7bf2e87f0f00908e8e/7/console

calfonso:
Hi!

We believe recent commits should resolve this question or problem for you. Please test again with the latest and let us know if you still see an issue.

Thank you!

jimi-c:
Rebased this myself and merged in 4811c7f117c4acaf06d295d738cb4c7ed4414500.

Thanks!

PR #18210, opened 2016-10-26
Move PluginLoader from __init__.py to loader.py

alikins:
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

lib/ansible/plugins/
##### ANSIBLE VERSION

```
ansible 2.3.0 (plugin_loader_loader_create 1297271bc1) last updated 2016/10/26 12:37:13 (GMT -400)
  lib/ansible/modules/core: (detached HEAD c51ced56cc) last updated 2016/10/26 10:59:51 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 8ffe314ea5) last updated 2016/10/24 11:26:36 (GMT -400)


```
##### SUMMARY

Move the PluginLoader class and it's global caches from
plugins/**init**.py to plugins/loader.py

One reason is so PluginLoader gets it's own module namespace
so that **name** is more accurate so that standard log setup
will get a better logger name ('ansible.plugins.loader' vs
'ansible.plugins'). This will let logging be enabled for the
plugin loader without enabling it for all plugins.

It also makes it more clear that plugins/**init**.py is where all
the loader objects are created instead of being at the end of the
file after a large class.

Third reason is it is a incremental step to making plugins/**init**.py be a empty package file with less side effects on import.

### DIFF
--- a/lib/ansible/plugins/__init__.py
+++ b/lib/ansible/plugins/__init__.py
@@ -20,19 +20,14 @@
 from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
-import glob
-import imp
 import inspect
 import os
 import os.path
 import sys
-import warnings
-
-from collections import defaultdict
 
 from ansible import constants as C
-from ansible.module_utils._text import to_text
 
+from ansible.plugins import loader
 
 try:
     from __main__ import display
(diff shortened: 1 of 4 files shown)


alikins:
This seemed like a reasonable small logical step towards https://github.com/ansible/ansible/blob/devel/docsite/rst/roadmap/ROADMAP_2_3.rst (Plugin Loader section)

alikins:
shrug

PR #18211, opened 2016-10-26
Simplify surrogate check in to_text()

sijis:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

`to_text()` in `module_utils/_text.py`
##### ANSIBLE VERSION

```
devel branch
```
##### SUMMARY

I notice an odd/redundant conditional check while poking through the code.
This should have the same behaviour but simplified.

Since `error` is being checked if it has a `None` value, you can simply move the extra (specific) condition up a few lines since the fallback is the same.

### DIFF
--- a/lib/ansible/module_utils/_text.py
+++ b/lib/ansible/module_utils/_text.py
@@ -163,7 +163,7 @@ def to_text(obj, encoding='utf-8', errors=None, nonstring='simplerepr'):
         return obj
 
     if errors in (None, 'surrogate_or_replace'):
-        if HAS_SURROGATEESCAPE:
+        if HAS_SURROGATEESCAPE or PY3:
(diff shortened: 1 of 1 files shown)


abadger:
Great!  You don't have to add the PY3 to the upper conditional, though.  The HAS_SURROGATE check is more exact.

sijis:
@abadger are you suggesting the extra PY3 condition isn't needed at all? If that's the case, I can easily update the PR accordingly.

sijis:
I see what you mean now. i've updated the PR. Thanks!

abadger:
Merged to devel and the stable-2.2 branch.  The change will likely show up first in 2.2.1.  Thanks @sijis!

PR #18221, opened 2016-10-27
Exception.message gone in 3.x

Qalthos:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

network, shell
##### ANSIBLE VERSION

```
2.3
```
##### SUMMARY

As far as I understand, this is the proper way to get around the lack of Exception.message in Python 3.x. `ShellError` had a self-defined `message` attribute, but not `NetworkError`, and I don't think it should be necessary anyway.

### DIFF
--- a/lib/ansible/module_utils/eos.py
+++ b/lib/ansible/module_utils/eos.py
@@ -88,7 +88,7 @@ class EosConfigMixin(object):
                 self.execute(['no configure session %s' % session])
         except NetworkError:
             exc = get_exception()
-            if 'timeout trying to send command' in exc.message:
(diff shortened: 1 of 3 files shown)


abadger:
Use to_native (from ansible.module_utils._text) instead of str().  Otherwise this looks good.

Qalthos:
@abadger thanks for the tip!

abadger:
+1 to merge!

abadger:
Also cherry-picked to stable-2.2.  This will be in 2.2.1

PR #18225, opened 2016-10-27
Fix OpenSSH-related ssh process exit race

jasdeep-hundal:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

connection/ssh.py
##### ANSIBLE VERSION

```
ansible 2.3.0
  config file = /Users/jasdeephundal/.ansible.cfg
  configured module search path = ["'./'"]
```
##### SUMMARY

This PR aims to solve the issue observed in this post: https://groups.google.com/forum/#!topic/ansible-project/oUmPgMN5HpU

When the called ssh process exists, it will first close its stdout file descriptor before exiting, writing an EOF to the descriptor. This leaves a tiny window of possibility that we see an EOF, but polling the process indicates that it has not returned. This window is small enough that any system call between the EOF write and the exit will ensure that we don’t hit this case, but for some reason our deployments hit this regularly.

This only causes in issue because OpenSSH does not close stderr when using ControlPersist. This has been reported as a bug here, but the fix merged still does not close stderr if using any ‘-v’ option: https://bugzilla.mindrot.org/show_bug.cgi?id=1988

What results is a select call on the stderr fd with the standard timeout of 90+ seconds, since we only adjust the timeout if we see that the process has exited. We do eventually see an EOF when the ControlPersist time has elapsed, resulting in pauses that correspond exactly with the ControlPersist timeout.

This fix changes the timeout whenever stdout sees an EOF, as well as setting the timeout to 0 in all cases when the process is closed since we should never see any more writes to the fds at that point. This has solved the issues we have seen.

### DIFF
--- a/lib/ansible/plugins/connection/ssh.py
+++ b/lib/ansible/plugins/connection/ssh.py
@@ -451,6 +451,14 @@ class Connection(ConnectionBase):
                 b_chunk = p.stdout.read()
                 if b_chunk == b'':
                     rpipes.remove(p.stdout)
+                    # When ssh has ControlMaster (+ControlPath/Persist) enabled, the
+                    # first connection goes into the background and we never see EOF
+                    # on stderr. If we see EOF on stdout, lower the select timeout
+                    # to reduce the time wasted selecting on stderr if we observe
+                    # that the process has not yet existed after this EOF. Otherwise
+                    # we may spend a long timeout period waiting for an EOF that is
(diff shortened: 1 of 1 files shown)


abadger:
Merged to devel and the temp-post-2.2.0-staging branch.  This will be in 2.3.0 and if testing goes okay, 2.2.1 as well.

jasdeep-hundal:
Thanks @abadger !

PR #18226, opened 2016-10-27
Add tests for dnf modelled after the yum tests

abadger:
##### ISSUE TYPE
- Test Pull Request
##### COMPONENT NAME

test/integration/target/dnf
##### ANSIBLE VERSION

```
devel
```
##### SUMMARY

Integration tests for dnf.

Note that this requires https://github.com/ansible/ansible-modules-extras/pull/3309   a feature to install the python bindings for dnf if they are not installed on the box. That's a new feature so if we backport the tests we need to comment out a few tests at the very beginning of this new role.

### DIFF
--- a/test/integration/destructive.yml
+++ b/test/integration/destructive.yml
@@ -7,6 +7,7 @@
     - { role: pip, tags: test_pip }
     - { role: gem, tags: test_gem }
     - { role: yum, tags: test_yum }
+    - { role: dnf, tags: test_dnf }
     - { role: apt, tags: test_apt }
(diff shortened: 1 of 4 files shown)


abadger:
dnf currently broken on Fedora RawHide due to: https://github.com/ansible/ansible-modules-extras/issues/3310  Going to have to wait to merge this until that is fixed.

abadger:
Current failures are unrelated to the dnf module.  Merged to devel.

abadger:
Backported to stable-2.2 with some changes because we don't have the feature which ensures that the python bindings are installed on the host in 2.2.x.  The tests should be running against 2.2.1

PR #18228, opened 2016-10-28
Add environment variables parsing in ec2.ini

redstyr:
ISSUE TYPE

Feature Pull Request
COMPONENT NAME

ec2.py

ANSIBLE VERSION

```
ansible 2.2.0
  config file = /home/aws-ansible/ansible.cfg
SUMMARY
```

ec2.py was unable to pick ec2.ini with environment variables. This will enable it to pick environment variables where needed.

One use case is if we need to filter the returned host list with a particular tag name which is defined in the environment.

```
pattern_include = %(AWS_PROFILE)s_
```

AWS_PROFILE is already defined and being used in other parts of the system.

This enables to use the same AWS Account for multiple environments like sandbox, test, staging etc without mixing up the nodes.

### DIFF
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -219,7 +219,7 @@ class Ec2Inventory(object):
         if six.PY3:
             config = configparser.ConfigParser()
         else:
-            config = configparser.SafeConfigParser()
+            config = configparser.SafeConfigParser(os.environ)
         ec2_default_ini_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'ec2.ini')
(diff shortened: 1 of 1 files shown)


willthames:
Does this work with python3?

redstyr:
I don't have a setup now to test but it should work. The pydoc doesn't mention os.environ in both versions.

s-hertel:
Hi, @redstyr. We aren't working on adding new features to the contrib scripts since adding inventory plugins in 2.4. The [aws_ec2 inventory plugin](https://docs.ansible.com/ansible/latest/plugins/inventory/aws_ec2.html) supports environment variables (including AWS_PROFILE), as well as improving a lot of other things, like using STS credentials defined in ~/.aws/config and falling back to attached instance roles. Feel free to reach out if you have questions.

PR #18230, opened 2016-10-28
Cleanup StringIO import for module_utils/shell.py

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

modules_utils


abadger:
Merged to devel and cherry-picked to stable-2.2... Probably too late for 2.2.0 but it will be in 2.2.1.  Will be in 2.2.0 if there's some other last minute major 2.2.0 fixes that require another rc.

PR #18237, opened 2016-10-28
attempt to ensure the correct python is used

### LINKED ISSUE #18235: Ansible should automatically set ansible_python_interpreter to sys.executable with 'connection: local'
##### ISSUE TYPE
- Feature Idea
##### COMPONENT NAME

`local` connection plugin
##### ANSIBLE VERSION

```
ansible 2.2.0.0
```
##### CONFIGURATION

N/A
##### OS / ENVIRONMENT

N/A
##### SUMMARY

There is currently logic to set `ansible_python_interpreter` to the value of `sys.executable` (the python interpreter that is running ansible) in the code that creates the implicit `localhost` entry in the inventory. This works for `local_action` (as it implicitly uses `localhost`), but it does not work when using `connection: local`.
##### STEPS TO REPRODUCE

N/A
##### EXPECTED RESULTS

N/A
##### ACTUAL RESULTS

N/A

bcoca:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

connection
##### ANSIBLE VERSION

```
2.3
```
##### SUMMARY

when using ansible with 'connection: local' and not using the implicit localhost
the python version can be incorrect, this should solve most of those cases.

Fixes #18235
TODO: this might make the implicit localhost setting irrelevant

### DIFF
--- a/lib/ansible/playbook/play_context.py
+++ b/lib/ansible/playbook/play_context.py
@@ -426,17 +426,18 @@ class PlayContext(Base):
                 elif getattr(new_info, 'connection', None) == 'local' and (not remote_addr_local or not inv_hostname_local):
                     setattr(new_info, 'connection', C.DEFAULT_TRANSPORT)
 
-        # if the final connection type is local, reset the remote_user value
-        # to that of the currently logged in user, to ensure any become settings
-        # are obeyed correctly
-        # additionally, we need to do this check after final connection has been
-        # correctly set above ...
+        # this needs to happen after final connection is set (above)
+        # if the connection type is local, reset the remote_user value to that of the currently logged in user,
+        # this ensures any become settings are obeyed correctly
+        # also set the python executable as this can/will vary from the initial host
         if new_info.connection == 'local':
             new_info.remote_user = pwd.getpwuid(os.getuid()).pw_name
-            if not new_info.get(ansible_python_interpreter):
-                mypython = sys.executable()
-                if mypython: # in some corner cases sys.executable is blank so we avoid rewriting modules
-                    new_info['ansible_python_interpreter'] = mypython
+            mypython = sys.executable()
+            if mypython: # in some corner cases sys.executable is blank so we avoid rewriting modules
+                new_info['ansible_python_interpreter'] = mypython
+            else:
+                display.warning('Unable to determine python interpreter from sys.executablem for local execution, using defaults.'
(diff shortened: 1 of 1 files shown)


abadger:
-1.  ansible_python_interpreter is independent of connection.  We should not be conflating the two.  People who want ansible python interpreter should just set ansible_python_interpreter:

```
- hosts: rhel7
  connection: local
  vars:
    ansible_python_interpreter: /my/other/python
```

Side note: this code is both broken and doesn't seem to work....  I fixed the breakage like this:

``` python
-            mypython = sys.executable()
+            mypython = sys.executable
             if mypython: # in some corner cases sys.executable is blank so we avoid rewriting modules
-                new_info['ansible_python_interpreter'] = mypython
+                setattr(new_info, 'ansible_python_interpreter', mypython)
```

But testing with those changes showed that it didn't change which python interpreter Ansible used to execute modules.

agaffney:
After thinking about this a bit, it may make sense to leave the `connection: local` behavior as-is and just tell people to use `local_action` or `delegate_to: localhost` to get this behavior.

willthames:
Great, so if you have a playbook with 100 tasks, you have to delegate_to: localhost on all of them?

context: aws playbooks

jhawkesworth:
@willthames could you not put the 100 tasks in a play that starts:
`
hosts: localhost`

**edit**: just read https://github.com/ansible/ansible/issues/33052 - I get it now.

Perhaps not applicable as I have static inventory, but looking at my playbooks I have an inventory group called 'controllers' which just contains the ansible controller machine, so any playbooks that just need to run locally are in a play
`hosts: controllers`
I think I did it this way so I can make use of some inventory-specific group vars but might be worth a try.

willthames:
@jhawkesworth I know all about [providing inventory specific variables for local hosts](http://willthames.github.io/2017/10/31/making-the-most-of-inventory.html)

But that doesn't help - the modules are run 'remotely' on the localhost, and no matter what connection you use, the default python interpreter is used, and there's no way of setting a better python interpreter *dynamically*

So you can do `ansible_python_interpreter: /usr/local/bin/python` in inventory/host_vars/localhost.yml or inventory/group_vars/controllers.yml but if you're running under multiple different virtualenvs,  or if multiple users with different pythons are using the same playbooks and inventory, you're out of luck.

PR #5428, opened 2013-12-26
ec2 module: Ephemeral devices can now be specified at instance creation time.

Geryon:
For EC2 instances that use EBS for the root device there is not instance store available at boot time.  These instance store volumes can't be added after the instance has been launched.  This adds the ability to specify the number of instance stores to attach to the instance during provisioning.  This is the equivalent to specifying the command line option(s) '-b "/dev/xvdb=ephemeral0" -b "/dev/xvdc=ephemeral1"'.

Example provisioning play:

```
- name: Provision and launch a new AWS EC2 Instance
  local_action: >
    ec2
    keypair='MyKeyPair'
    group='default'
    instance_type='m1.large'
    image='ami-a73264ce'
    ephemeral=2
    wait=true
    region='us-east-1'
    zone='us-east-1b'
  register: ec2
```

PR #18241, opened 2016-10-28
Handle the socket.timeout that paramiko can return

bdowling:
##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

Network modules, shell.py

##### ANSIBLE VERSION

```
ansible 2.3.0 (fix_command_timeout 886f146de9) last updated 2016/10/28 15:10:13 (GMT -400)
  lib/ansible/modules/core: (fix_command_timeout 93051cf470) last updated 2016/10/28 15:02:12 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 8ffe314ea5) last updated 2016/10/25 15:34:40 (GMT -400)
```
##### SUMMARY

```
   "msg": "timeout waiting for next prompt from remote device",
    "**stdout**": "copy running-config flash:test\r\nDestination filename [test]? ",
    "**stdout_lines**": [
        "copy running-config flash:test",
        "Destination filename [test]? "
```

The socket.timeout was not handled and causes fatal errors.
This is particularly noticeable when the device is doing alternative
prompts.  This PR also adds the output of the device to the 'stdout'
attribute and returns it to help see the hidden device interaction(s).

This PR will depend on a separate PR posted to ansible_code_modules.

Note there is a change here to stdout_lines that is set in lib/ansible/plugins/action/**init**.py as well that handles both cases of 'stdout' as a string or list.  Since the the network modules do, all of the to_list() functions duplicated across network modules could thus be removed and this will handle stdout_lines for them.  One added side effect here is that it uses splitlines() vs split("\n"), so the extra "\r" that sometimes showed in stdout will be gone.

This, in-part, is related to https://github.com/ansible/ansible-modules-core/issues/5402

Sorry to intermingle these changes somewhat, they are inter-related.

### DIFF
--- a/lib/ansible/module_utils/shell.py
+++ b/lib/ansible/module_utils/shell.py
@@ -50,9 +50,10 @@ def to_list(val):
 
 class ShellError(Exception):
 
-    def __init__(self, msg, command=None):
+    def __init__(self, msg, command=None, stdout=None):
         super(ShellError, self).__init__(msg)
         self.command = command
+        self.stdout = stdout
 
 
 class Shell(object):
@@ -125,7 +126,14 @@ class Shell(object):
         handled = False
 
         while True:
-            data = self.shell.recv(200)
+            try:
+                data = self.shell.recv(200)
+            except socket.timeout:
+                exc = get_exception()
+                exc.command = cmd
+
+                raise ShellError("timeout waiting for next prompt from remote device",
+                                 stdout = recv.getvalue())
 
             recv.write(data)
             recv.seek(recv.tell() - 200)
@@ -247,7 +255,7 @@ class CliBase(object):
             return self.shell.send(commands)
         except ShellError:
             exc = get_exception()
-            raise NetworkError(to_native(exc), commands=commands)
+            raise NetworkError(to_native(exc), command=exc.command, stdout=exc.stdout)
 
     def run_commands(self, commands):
         return self.execute(to_list(commands))
--- a/lib/ansible/plugins/action/__init__.py
+++ b/lib/ansible/plugins/action/__init__.py
@@ -659,7 +659,11 @@ class ActionBase(with_metaclass(ABCMeta, object)):
         # pre-split stdout into lines, if stdout is in the data and there
         # isn't already a stdout_lines value there
         if 'stdout' in data and 'stdout_lines' not in data:
-            data['stdout_lines'] = data.get('stdout', u'').splitlines()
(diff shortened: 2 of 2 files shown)


bdowling:
@privateip I noticed that the https://github.com/ansible/ansible-modules-core/pull/5428 was merged, but this piece was not.  Without these changes merged in, I'm getting the error on the splitlines that is addressed by this pr:

```
  File "/home/bdowling/src/ansible/lib/ansible/plugins/action/__init__.py", line 662, in _execute_module
    data['stdout_lines'] = data.get('stdout', u'').splitlines()
AttributeError: 'list' object has no attribute 'splitlines'
```

privateip:
agh i just caught this as well.   please don't intermingle unrelated changes.  i dont think making the change to action/**init**.py is the correct path forward here.  This PR will need to be rebased as well.  I would recommend updating this one to pull out the change to ActionBase.  I'm going to revert #5428 so it will need to be updated to

bdowling:
@privateip  I'm not sure I understand your reasoning here.  Can you please help me understand why the action/init change is not the correct path?   I agree there are two changes here, but they are rather small and they are so closely related that it felt natural.  

An exception was not handled properly and returned an error that offered no understanding to the user what happened.  I fixed the exception and the result still offered no value to understand what the device was doing at the time, so I added stdout to the return so that the error could display what was the last thing received from the network device when it failed.   When I realized that stdout_lines was not going to be handled by the to_lines in the network code, I found that there was an essentially equivalent function within action/init.py, but it didn't handle the array data.  

When I realized this, I also saw it was doing the same function as the 23 copies of the to_lines() function definition across the various network modules.  It seemed natural to instead rely on the functionality provided by the higher-level framework would be better and improve the overall situation?

Thanks...

privateip:
The stdout_lines is a hack around the standard implementation due to the constraints about how the current network modules are implemented.   We are going to fix this in 2.3 so that network modules are more tightly integrated and use the standard directives.   

For instance, this

```
ios_command:
  commands:
    - show version
    - show running-config
```

should really be implemented as 

```
ios_command: "{{ item }}"
with_items:
  - show version
  - show running-config
```

Of course with the current implementation, the latter while possible, is a performance killer.    So my reasoning for not making the change in action base is to confine the hack to the modules for now at which point we will get the underlying connection fixed and the hack can just go away with out legacy code in action base.

calfonso:
We believe this can be closed.

Thank you!

PR #18266, opened 2016-10-31
Fix auth in collins.py inventory

michaeljs1990:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
collins.py

##### ANSIBLE VERSION
Issue exists in devel however I encountered it fist in the version below
```
ansible --version
ansible 2.0.0.2
  config file = 
```

##### SUMMARY

This forces basic auth to be used. Using the normal HTTPPasswordMgrWithDefaultRealm
password manager from urllib2 fails since collins doesn't send a 401 retry on failure.
More about this can be seen here http://stackoverflow.com/questions/2407126/python-urllib2-basic-auth-problem.
I added a small comment about the format of the host so others don't waste time like i did.

```
# Before change
python collins.py
{}
# After change
python collins.py
{ ... bunch of top secret stuff}
```

cc: @jbrockett

### DIFF
--- a/contrib/inventory/collins.ini
+++ b/contrib/inventory/collins.ini
@@ -3,6 +3,8 @@
 
 [collins]
 
+# You should not have a trailing slash or collins
+# will not properly match the URI
 host = http://localhost:9000
 
 username = blake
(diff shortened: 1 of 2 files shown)


michaeljs1990:
sorry to bother about this one again but as it's a relatively minor change would someone be able to take a quick look? Thanks!

PR #18274, opened 2016-10-31
Implement basic DMI facts for OpenBSD

jasperla:
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
facts (setup module)

##### ANSIBLE VERSION

```
ansible 2.2.0.0

```

##### SUMMARY

This PR adds support for retrieving basic DMI facts on OpenBSD. A lot of information can be retrieved through sysctl(8). As described in the code I explicitly chose not to depend on dmidecode(1) which would get all the information for two reasons:
- it would rely on superuser privileges
- an additional package would have to be installed on the managed node

As an added bonus of using sysctl(8) (when the fields are actually populated by the kernel) is that for example the `ansible_system_vendor` fact works on non-x86 platforms such as OpenBSD/mips64.

For example on OpenBSD/amd64 the new facts result in the following (obfuscated for obvious reasons):
```
    "ansible_facts": {
        [...]
        "ansible_product_name": "20CLS0QX00", 
        "ansible_product_serial": "product-serial-goes here", 
        "ansible_product_uuid": "random-uuid-goes-here", 
        "ansible_product_version": "ThinkPad X250",
       [...]
        "ansible_system_vendor": "LENOVO", 
     }
```

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -1585,7 +1585,8 @@ class OpenBSDHardware(Hardware):
     - processor_cores
     - processor_count
     - processor_speed
-    - devices
+
+    In addition, it also defines number of DMI facts and device facts.
     """
     platform = 'OpenBSD'
     DMESG_BOOT = '/var/run/dmesg.boot'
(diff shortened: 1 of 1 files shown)


jasperla:

jasperla:
Thanks @bcoca !

PR #18277, opened 2016-11-01
On OpenBSD, swap the values for distribution_version and distribution…

jasperla:
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME

facts (setup module)

##### ANSIBLE VERSION

```
ansible 2.2.0.0
```

##### SUMMARY

This PR swap the values for the `distribution_release` and `distribution_version` facts on OpenBSD. This was done to be more inline with other platforms Ansible runs on, and to have less "surprising" values for OpenBSD users.
OpenBSD ships regular releases with a numbered version (e.g. 6.0). There are several "flavors" so to say, `release`, `current`, `beta` and `stable` which are output into the `sysctl kern.version` string.

This PR brings the values inline with for example RH7 where the `_version` is a numerical value (7.2) and the `_release` a string (Maipo).

Before:
```
localhost | SUCCESS => {
    "ansible_facts": {
        "ansible_distribution": "OpenBSD", 
        "ansible_distribution_release": "6.0", 
        "ansible_distribution_version": "current"
    }, 
    "changed": false
}
```
after:
```
localhost | SUCCESS => {
    "ansible_facts": {
        "ansible_distribution": "OpenBSD", 
        "ansible_distribution_release": "current", 
        "ansible_distribution_version": "6.0"
    }, 
    "changed": false
}
```

Which is closer to:
```
        "ansible_distribution": "RedHat", 
        "ansible_distribution_major_version": "7", 
        "ansible_distribution_release": "Maipo", 
        "ansible_distribution_version": "7.2", 
```

This change was discussed and agreed with several other OpenBSD developers.

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -764,12 +764,13 @@ class Distribution(object):
             self.facts['distribution_version'] = '%s.%s' % (data.group(1), data.group(2))
 
     def get_distribution_OpenBSD(self):
+        self.facts['distribution_version'] = platform.release()
         rc, out, err = self.module.run_command("/sbin/sysctl -n kern.version")
         match = re.match('OpenBSD\s[0-9]+.[0-9]+-(\S+)\s.*', out)
         if match:
-            self.facts['distribution_version'] = match.groups()[0]
+            self.facts['distribution_release'] = match.groups()[0]
         else:
-            self.facts['distribution_version'] = 'release'
+            self.facts['distribution_release'] = 'release'
 
     def get_distribution_DragonFly(self):
         pass


jasperla:

abadger:
Reviewed the code change.  From the code standpoint this looks good to me.  It is backwards incompatible, though.  I've mentioned this on IRC to get feedback.

jasperla:
@abadger Thanks; I've talked to a fellow OpenBSD developers who use ansible (@r-p-e and @ajacoutot) and they agreed with the change at least.

ajacoutot:
Yes, I confirm this is something we want in OpenBSD, even if it's backward incompatible.

abadger:
Merged to devel and added a 2.3.0 changelog entry:
- The version and release facts for OpenBSD hosts were reversed.  This has been
  changed so that version has the numeric portion and release has the name of
  the release.

jasperla:
Thanks!

PR #18278, opened 2016-11-01
Rework how OpenBSD processor facts are resolved

jasperla:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME
facts (setup module)

##### ANSIBLE VERSION

```
ansible 2.2.0.0
```

##### SUMMARY

This PR simplifies the resolution of the processor facts on OpenBSD by using the sysctl `hw.model` and `hw.ncpu` values Previously` /var/run/dmesg.boot` or dmesg(8) would be parsed however that could lead to bogus values (e.g. there would be `Enhanced SpeedStep 2195 MHz: speeds: 2301, 2300, 2200, 2000, 1900, 1800, 1700, 1500, 1400, 1300, 1100, 1000, 900, 800, 600, 500 MHz` entry in `ansible_processor`) if the buffer contained multiple (partial) dmesgs.

This also brings the output inline with Linux for example where the trailing `, 1234.56 MHz` isn't present either.

Before:
```
        "ansible_processor": [
            "Intel(R) Core(TM) i5-5300U CPU @ 2.30GHz, 2195.12 MHz",
            "Intel(R) Core(TM) i5-5300U CPU @ 2.30GHz, 2194.92 MHz",
            "Intel(R) Core(TM) i5-5300U CPU @ 2.30GHz, 2194.92 MHz",
            "Intel(R) Core(TM) i5-5300U CPU @ 2.30GHz, 2194.92 MHz"
        ],
        "ansible_processor_cores": "NA",
        "ansible_processor_count": 4,
```

After:
```
        "ansible_processor": [
            "Intel(R) Core(TM) i5-5300U CPU @ 2.30GHz",
            "Intel(R) Core(TM) i5-5300U CPU @ 2.30GHz",
            "Intel(R) Core(TM) i5-5300U CPU @ 2.30GHz",
            "Intel(R) Core(TM) i5-5300U CPU @ 2.30GHz"
        ],
        "ansible_processor_count": 4,
```

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -1589,7 +1589,6 @@ class OpenBSDHardware(Hardware):
     In addition, it also defines number of DMI facts and device facts.
     """
     platform = 'OpenBSD'
-    DMESG_BOOT = '/var/run/dmesg.boot'
 
     def populate(self):
         self.sysctl = self.get_sysctl()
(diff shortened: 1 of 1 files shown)


jasperla:

abadger:
Backwards compatibility again... it's probably better to list both processor_cores and processor_count with the same value than to get rid of processor_count.  It's not too much of a lie.. If we can't tell the difference between cores and count it's possible that we have X processors with 1 core each.  Other than that it looks good.  sysctl seems like a less error prone method to use than searching through dmesg.

If someone would like to test this, that would be appreciated but I'm willing to merge it once the change to keep processor_cores is made.  Ping me in #ansible-devel on irc.freenode.net once the PR is updated.

abadger:
Thanks @jasperla !  Merged to devel for 2.3.0.

abadger:
Also merged into stable-2.2.  It will be in the next 2.2 release (probably 2.2.1)

Issue #14430, opened 2016-02-11
"changed_when: True" does not display the task as changed


J-Griffin:
Ansible version 2.0.0.2
Expected "Test changed_when" to be "changed" and yellow. Was "ok" and green instead.

**changed_when_test.yml:**

``` yaml

---
- name: Test Play
  hosts: localhost
  gather_facts: no
  tasks:
    - name: Test changed_when
      debug: msg="This should be changed"
      changed_when: True
```

**Result:**

```
PLAY [Test Play] *********************************************

TASK [Test changed_when] *******************************************************
ok: [localhost] => {
    "msg": "This should be changed"
}

PLAY RECAP *********************************************************************
localhost        : ok=1    changed=1    unreachable=0    failed=0
```

bcoca:
so the status IS set to changed, though the debug output is 'hardcoded' as OK as it cannot change the acutal machines (play summary does show 1 changed task). You can see the status using the following:

```
---
- name: Test Play
  hosts: localhost
  gather_facts: no
  tasks:
    - debug: msg="test"
      changed_when: True
      register: mystatus

   - debug: var=mystatus
```

output:

```
PLAY ***************************************************************************

TASK [debug] *******************************************************************
ok: [localhost] => {
    "msg": "test"
}

TASK [debug] *******************************************************************
ok: [localhost] => {
    "mystatus": {
        "changed": true, 
        "msg": "test"
    }
}

PLAY RECAP *********************************************************************
localhost                  : ok=1   changed=1    unreachable=0    failed=0   
```

J-Griffin:
That's confusing. It counts as changed in the recap, but not the task? In Ansible 1.9, both recap count and task display showed changed.

bcoca:
That is why i didn't close this, still an issue and can be used by subsequent tasks. Not sure why you would want this 'feature' to display changed though.

J-Griffin:
A fair point. My use case is a bit goofy. If important tasks in my playbook are changed, I notify a "you should know this" handler. I set changed_when on the handler so that it's yellow and stands out. Something like:

``` yaml
handlers:
  - name: elasticsearch config changed
    debug: msg="ES config changed! Rolling restart needed"
    changed_when: True
```

The reason I do this is that some services are painful or dangerous to restart, so it's best to just know that it needs to happen.

rvben:
[...]

PR #18279, opened 2016-11-01
Fix for #14430: 'changed_when: True' does not display the task as cha…

rvben:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

callback

##### ANSIBLE VERSION

```
ansible 2.3.0 (devel e4115e0b98) last updated 2016/11/01 12:23:33 (GMT +200)
  lib/ansible/modules/core: (detached HEAD 679a0ae5e9) last updated 2016/11/01 12:14:14 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD cf524673e1) last updated 2016/11/01 12:14:14 (GMT +200)
```

###### The Bug
When forcing a 'debug'-task to be flagged as 'changed' with 'changed_when': True, the task itself prints as 'ok', while it is counted as 'changed' in the end-results. This is caused by the '_clean_results'-function, which removes the 'changed'-entry from the results-variable. However, this happens after the task is counted as 'changed', so an inconsistency is born.

###### The Fix
It's good that by default the 'changed'-entry is removed. However, if 'changed_when: True' is it's clear that the intention is that it should be marked as 'changed', both in task-result and in play-result. That's why it seemed best to prevent the 'changed'-result from being removed when it's value is True.

This is the result. The first is a normal debug task, and the second one is marked with 'changed_when: True'.
```
TASK [Debug Test Ok] **************************************************************
ok: [localhost] => {
    "msg": "Test message that should be green, and ok.
}

TASK [Debug Test Changed] **************************************************************
changed: [localhost] => {
    "changed": true, 
    "msg": "Test message that should be yellow, and changed."
}
```

### DIFF
--- a/lib/ansible/executor/task_executor.py
+++ b/lib/ansible/executor/task_executor.py
@@ -534,7 +534,7 @@ class TaskExecutor:
 
             # if we didn't skip this task, use the helpers to evaluate the changed/
             # failed_when properties
-            if 'skipped' not in result:
+            if 'skipped' not in result and 'debug' not in self._task.action:
                 _evaluate_changed_when_result(result)
                 _evaluate_failed_when_result(result)
 


bcoca:
`debug` is by design always 'ok', your change basically invalidates the whole condition and the `del` below.

rvben:
@bcoca, Thanks for your quick response. I really want to get this right, so I investigated it further. The design choice to have it always 'ok' is reasonable. That's why it seems best to skip evaluation of the changed_when and failed_when variables. Both these variables should not have any effect on the outcome of a debug-task, if I understand you correctly.
Can you please look at my revision, and see if this is a better solution for the issue?

mattclay:
CI is failing on the `handlers` integration test:

```
2017-02-01 23:56:38 ++ ansible-playbook test_handlers.yml -i inventory.handlers -v -vvvvvv --tags scenario2 -l A
2017-02-01 23:56:38 ++ egrep -o 'RUNNING HANDLER \[test_handlers : .*?]'
2017-02-01 23:56:39 + '[' '' = 'RUNNING HANDLER [test_handlers : test handler]' ']'
2017-02-01 23:56:39 Command exited with status 1 after 5.20362496376 seconds.
2017-02-01 23:56:39 NOTICE: To resume at this test target, use the option: --start-at handlers
```

Full logs for one of the failures is here: https://app.shippable.com/runs/58926c8fd3cc6a1000c37e63/37/console

This one is a bit tricky to troubleshoot, since the test involves running `grep` on the playbook output.

The tests can be run locally with: `test/runner/ansible-test integration handlers -v`

bcoca:
The fix is wrong, for example, it breaks people using debug to execute handlers.

jctanner:
Hi!

As of April of 2016, we have started using the Ansible Proposal process for large feature ideas or changes in current functionality, such as this. If you are still interested in seeing this new feature get into Ansible, please submit a proposal for it using this process.

https://github.com/ansible/proposals/blob/master/proposals_process_proposal.md

If you have any further questions, please let us know by stopping by our devel mailing list, or our devel IRC channel:

   * #ansible-devel on Freenode

Thank you!

PR #18286, opened 2016-11-01
Fix role dependency syntax in galaxy.rst

danrue:
##### ISSUE TYPE
 - Docs Pull Request

##### COMPONENT NAME
galaxy
##### ANSIBLE VERSION
Verified in 1.9.4 and 2.1.2.0.

##### SUMMARY

The correct dictionary key is 'role', not 'name', in a role dependency. Using 'name' results in "expected a role name in dictionary" error.


bcoca:
This was changed in 2.2, going forward `name` is the preferred way to specify a role in all circumstances. We had too many different ways to reference roles and we are trying to reduce them. The docs should not simply revert but should make clear when this change happened.

PR #18288, opened 2016-11-01
aix network facts: Separate out the uname call to reduce total calls

### LINKED ISSUE #11289: AIX WPAR handling is inefficient and possibly buggy

jctanner:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME
facts.py module util

##### ANSIBLE VERSION

```
2.3
```

##### SUMMARY

Fixes #11289

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -2837,6 +2837,14 @@ class AIXNetwork(GenericBsdIfconfigNetwork):
             all_ipv4_addresses = [],
             all_ipv6_addresses = [],
         )
+
+        uname_rc = None
+        uname_out = None
+        uname_err = None
+        uname_path = self.module.get_bin_path('uname')
+        if uname_path:
+            uname_rc, uname_out, uname_err = self.module.run_command([uname_path, '-W'])
+
         rc, out, err = self.module.run_command([ifconfig_path, ifconfig_options])
 
         for line in out.splitlines():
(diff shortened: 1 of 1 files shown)


lib/ansible/module_utils/facts.py
```diff
@@ -2866,12 +2874,12 @@ def get_interfaces_info(self, ifconfig_path, ifconfig_options='-a'):
                     self.parse_inet6_line(words, current_if, ips)
                 else:
                     self.parse_unknown_line(words, current_if, ips)
-            uname_path = self.module.get_bin_path('uname')
+
             if uname_path:
```
bcoca: why not have the execution here and avoid the repeat of `if uname_path`

jctanner: @bcoca i'm not sure what you mean

PR #18296, opened 2016-11-02
Fix adding the same trusted certificates multiple times

puiterwijk:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
module_utils/urls.py

##### ANSIBLE VERSION
```
ansible 2.2.0.0
  configured module search path = ['/srv/web/infra/ansible/library', '/usr/share/ansible']
```

##### SUMMARY
If there is an intermittent network failure, we might be trying to reach
an URL multiple times. Without this patch, we would be re-adding the same
certificate to the OpenSSL default context multiple times.
Normally, this is no big issue, as OpenSSL will just silently ignore them,
after registering the error in its own error stack.
However, when python-cryptography initializes, it verifies that the current
error stack of the default OpenSSL context is empty, which it no longer is
due to us adding the certificates multiple times.
This results in cryptography throwing an Unknown OpenSSL Error with details:

OpenSSLErrorWithText(code=185057381L, lib=11, func=124, reason=101,
reason_text='error:0B07C065:x509 certificate routines:X509_STORE_add_cert:cert already in hash table'),

### DIFF
--- a/lib/ansible/module_utils/urls.py
+++ b/lib/ansible/module_utils/urls.py
@@ -182,6 +182,8 @@ if not HAS_SSLCONTEXT and HAS_SSL:
         del libssl
 
 
+LOADED_VERIFY_LOCATIONS = set()
+
 HAS_MATCH_HOSTNAME = True
 try:
     from ssl import match_hostname, CertificateError
@@ -590,6 +592,8 @@ class SSLValidationHandler(urllib_request.BaseHandler):
         paths_checked.append('/etc/ansible')
 
         tmp_fd, tmp_path = tempfile.mkstemp()
+        to_add_fd, to_add_path = tempfile.mkstemp()
+        to_add = False
 
         # Write the dummy ca cert if we are running on Mac OS X
         if system == 'Darwin':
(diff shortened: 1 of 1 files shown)


jctanner:
This was found in koji's fedora builds where networking is disabled in the mock environment.

lib/ansible/module_utils/urls.py
```diff
@@ -590,6 +592,8 @@ def get_ca_certs(self):
         paths_checked.append('/etc/ansible')
 
         tmp_fd, tmp_path = tempfile.mkstemp()
+        to_add_fd, to_add_path = tempfile.mkstemp()
```
abadger, line 595: Is there a reason that we have to keep this separate from tmp_fd, tmp_path?

abadger: If we do have to keep it separate are we able to remove it when we remove tmp_path around line 728

puiterwijk: Yes. The reason for that is that the tmp_ca_cert_path is used in more places, and is expected to contain everything for further underlying libraries. This way, I can limit the specific list that we add to OpenSSL to non-previously-loaded without limiting the others.

puiterwijk: And yes, we are able to remove it, and I'll add the code to do so.

abadger: Excellent.  :-)  I'll merge once that code is applied.

abadger:
Errors in shippable appear to be unrelated (Looks like a problem in the gpg keyserver that the git integration tests are using).  I'll hit rebuild on it once and then merge if the pattern continues to look unrelated.

puiterwijk:
Oh, I'd just added a single more if to don't go try and delete a None file (which makes no actual difference, but just makes me feel more sane :) )

abadger:
Merged to devel and stable-2.2.  Fix will be in 2.2.1  Thank @puiterwijk !

PR #18320, opened 2016-11-02
Invoke vault password scripts via original path specified

jmehnle:
##### ISSUE TYPE
 - Bug Report

##### COMPONENT NAME
`--vault-password-file`

##### ANSIBLE VERSION
```
ansible 1.9.3
```
(but also occurs in Ansible 2)

##### DESCRIPTION
When specifying an executable vault password file, Ansible invokes it by its "real" path, not by the path as specified. This matters if what's passed is actually a *symlink* to an executable, and the executable uses the path by which it is invoked to determine the password to produce.

For example:

```
$ ls -l
total 16
-rwxr-xr-x  1 julian  staff  290 Nov  2 23:03 get-password
-rw-------  1 julian  staff  355 Nov  2 23:10 secrets.yml
```

`secrets.yml` is a dummy variables vault file created by `ansible-vault create` with the password `foo-bar-ansible`.

```
$ cat get-password
#!/bin/sh

basename="$(basename $0)"
basename_prefix="get-password"
default_password="foo-bar"

case "${basename}" in
  "${basename_prefix}"-*)
    password="${default_password}-${basename#${basename_prefix}-}"
    ;;
  *)
    password="${default_password}"
    ;;
esac

echo "${password}"
```

This script produces the password `foo-bar` if invoked directly, but produces a different password when invoked through a symlink that begins with `get-password-`, e.g., `foo-bar-quux` when invoked as `get-password-quux`. This script is a synthetic example, but I actually use this mechanism with a more complicated password retrieval script that installs various symlinks for different applications, Ansible being one of them.

Now lets create a symlink with a suitable name:

```
$ ln -s get-password get-password-ansible
$ ls -l get-password*
-rwxr-xr-x  1 julian  staff  290 Nov  2 23:03 get-password
lrwxr-xr-x  1 julian  staff   12 Nov  2 23:24 get-password-ansible -> get-password
$ ./get-password
foo-bar
$ ./get-password-ansible
foo-bar-ansible
$ ansible all -i localhost, -c local -a uptime -e @secrets.yml --vault-password-file get-password-ansible
ERROR: Decryption failed
```

Oops. We're invoking `get-password-ansible`, which as we just saw, produces the correct password, `foo-bar-ansible`. What gives?

Well, now let's try replacing the symlink with a straight-up copy of the real file:

```
$ rm -f get-password-ansible
$ cp get-password get-password-ansible
$ ls -l get-password*
-rwxr-xr-x  1 julian  staff  290 Nov  2 23:03 get-password
-rwxr-xr-x  1 julian  staff  290 Nov  2 23:12 get-password-ansible
$ ./get-password
foo-bar
$ ./get-password-ansible
foo-bar-ansible
$ ansible all -i localhost, -c local -a uptime -e @secrets.yml --vault-password-file get-password-ansible
localhost | success | rc=0 >>
23:12  up 29 days,  2:46, 24 users, load averages: 1.22 1.50 1.58
```

Oh, hey, this time it worked!

And this is consistent with how the `real_vault_password_file` function is defined in https://github.com/ansible/ansible/blob/devel/lib/ansible/cli/__init__.py#L613. It calls `os.path.realpath()` on the file name provided and then invokes the script by that name.

It would be more useful if the vault password file's executability was determined by its real path, etc., but then invoked by its *original* path.

### DIFF
--- a/changelogs/fragments/ensure_bytes_run_command.yml
+++ b/changelogs/fragments/ensure_bytes_run_command.yml
@@ -1,0 +1,2 @@
+bugfixes:
+    - ensure run_command passes bytes to Popen, which is what it expects.
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -2452,13 +2452,16 @@ class AnsibleModule(object):
 
             # stringify args for unsafe/direct shell usage
             if isinstance(args, list):
-                args = " ".join([shlex_quote(x) for x in args])
+                args = b" ".join([to_bytes(shlex_quote(x), errors='surrogate_or_strict') for x in args])
+            else:
+                args = to_bytes(args, errors='surrogate_or_strict')
 
             # not set explicitly, check if set by controller
             if executable:
-                args = [executable, '-c', args]
+                executable = to_bytes(executable, errors='surrogate_or_strict')
+                args = [executable, b'-c', args]
             elif self._shell not in (None, '/bin/sh'):
-                args = [self._shell, '-c', args]
+                args = [to_bytes(self._shell, errors='surrogate_or_strict'), b'-c', args]
             else:
                 shell = True
         else:
@@ -2474,9 +2477,9 @@ class AnsibleModule(object):
 
             # expand ``~`` in paths, and all environment vars
             if expand_user_and_vars:
-                args = [os.path.expanduser(os.path.expandvars(x)) for x in args if x is not None]
+                args = [to_bytes(os.path.expanduser(os.path.expandvars(x)), errors='surrogate_or_strict') for x in args if x is not None]
             else:
-                args = [x for x in args if x is not None]
+                args = [to_bytes(x, errors='surrogate_or_strict') for x in args if x is not None]
 
         prompt_re = None
         if prompt_regex:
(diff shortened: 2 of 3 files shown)


calfonso:
@jmehnle Would you mind rebasing and testing, we believe this has been fixed.

jmehnle:
@calfonso, how would this have been fixed? It seems the code still suffers from the same confusion of what my patch calls the "full path" with the "real path". Unless you separate the two out, I don't believe the problem can be fixed.

mattclay:
Rebase needed.

jmehnle:
I just rebased it the other day. What gives??

mattclay:
@jmehnle The latest rebase was needed because the file changed by this PR had been modified by other PR(s) which were already merged, causing conflicts with your changes.

sivel:
This issue still exists, but the code has change quite a lot.

I think this would be a good change, as I understand the reasons behind it.

The new code is located at:

https://github.com/ansible/ansible/blob/9773a1f2896a914d237cb9926e3b5cdc0f004d1a/lib/ansible/parsing/vault/__init__.py#L372-L373

I think we just need to change `realpath` to `abspath` there.

bcoca:
@jmehnle can you confirm that PR above resolves your issue?

PR #18322, opened 2016-11-02
Fix indentation of multiline descriptions

abadger:
##### ISSUE TYPE

 - Bugfix Pull Request
 - Docs Pull Request

##### COMPONENT NAME
hacking/templates/rst.j2

##### ANSIBLE VERSION

```
devel
```

##### SUMMARY
When generating html docs, the options table is created as raw html rather than an rst html table.  Because the raw html is inside of an rst html block we need to keep a proper indentation level.  The description of options element wasn't doing that previously if there were embedded newlines inside of them.  This fix changes embedded newlines into a newline plus indentation so that it does not end the raw block.

### DIFF
--- a/hacking/templates/rst.j2
+++ b/hacking/templates/rst.j2
@@ -79,7 +79,7 @@ Options
     {% else %}
     <td><ul>{% for choice in v.get('choices',[]) -%}<li>@{ choice }@</li>{% endfor -%}</ul></td>
     {% endif %}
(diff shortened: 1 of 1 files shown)


abadger:
Before this commit, the efs_facts module documentation is malformed.  The raw html table does not get closed so the rest of the page ends up in one of the columns.  With this change, the efs_facts module docs, although not yet pretty, do properly structure the table.

PR #18327, opened 2016-11-03
Exclude floppy disks from lsblk call for uuids

### LINKED ISSUE #18326: Setup fails with "TimeoutError" on machines with floppy drive
##### ISSUE TYPE
 - Bug Report

##### COMPONENT NAME

setup module

##### ANSIBLE VERSION
```
ansible 2.2.0.0
```
Also devel

##### CONFIGURATION
default

##### OS / ENVIRONMENT
OpenSuSE 42.1 Leap

##### SUMMARY
Running the setup module of ansible 2.2.0.0 on a machine with a floppy drive fails with a timeout.

##### STEPS TO REPRODUCE

```
ansible oldmachine -m setup
```

##### EXPECTED RESULTS
Normal setup result

##### ACTUAL RESULTS
```
oldmachine | FAILED! => {
    "changed": false, 
    "cmd": "/usr/bin/lsblk --list --noheadings --paths --output NAME,UUID", 
    "failed": true, 
    "msg": "Timer expired", 
    "rc": 257
}
```

The issue was introduced by #17036.
`lsblk` is called once to collect all devices. For machines that still have floppy drives attached (don't laugh, those still exist), this includes floppy drives, which take very long to answer.

In our case it takes 12 seconds for the floppy drive, 0.1 seconds for everything else.

robinro:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
setup

##### ANSIBLE VERSION
```
ansible 2.2.0.0
```

##### SUMMARY
Fixes #18326

Since #17036 `lsblk` first collects uuids from all devices and then connects those to the active mounts. This also includes floppy drives, which are a lot slower to answer than HDD/SSD/network.

This patch adds the option `--exclude 2`, that excludes floppy drives from the lsblk command.

Consequences:
1. runtime of lsblk goes from 12 seconds to 0.1 seconds in our use case
2. floppy mounts are listed with `uuid: N/A`

Does anyone care for the uuid of floppy discs? If not I would favor 1. over 2. and ask you to merge this in also in 2.2 stable.

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -1228,7 +1228,11 @@ class LinuxHardware(Hardware):
                     self.facts[k] = 'NA'
 
     def _run_lsblk(self, lsblk_path):
-        args = ['--list', '--noheadings', '--paths',  '--output', 'NAME,UUID']
+        # call lsblk and collect all uuids
+        # --exclude 2 makes lsblk ignore floppy disks, which are slower to answer than typical timeouts
+        # this uses the linux major device number
+        # for details see https://www.kernel.org/doc/Documentation/devices.txt
+        args = ['--list', '--noheadings', '--paths',  '--output', 'NAME,UUID', '--exclude', '2']
         cmd = [lsblk_path] + args
         rc, out, err = self.module.run_command(cmd)
         return rc, out, err


robinro:
@alikins @abadger 
You wrote/merged #17036, maybe you can comment on or merge this one?

ryansb:
Looks good to me, but I don't have a floppy to test.

robinro:
The question for me was whether ansible wants to support UUIDs for floppy discs. If not, then normal shippable runs should be enough of a test.

abadger:
I don't care about floppy UUIDs and the fact that they're so slow makes it pragmatic to exclude them.

Reading the lsblk man page, it says that --exclude defaults to excluding RAM disk devices.  Do we need to include their major number in the --exclude or is it additive?

Code looks good otherwise.  It's in the LinuxHardware class so it should only affect Linux so it's okay to hardcode the Linux floppy major number.

robinro:
@abadger Thanks for taking a look.

About ramdisks: According to https://www.kernel.org/doc/Documentation/devices.txt this is for "older kernels". Modern ramdisks like tmpfs won't show up that way. Anyhow including fast disks doesn't hurt, since they just won't be used in the later steps of the code.

abadger:
Merged to devel and cherry-picked to stable-2.2 (should show up in 2.2.1).

abadger:
Thanks @robinro for the fix!

robinro:
Thanks for merging!

PR #18332, opened 2016-11-03
Add support for AWS profiles and IAM ARN for AssumeRole.

mkubenka:
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
Credstash Lookup

##### ANSIBLE VERSION
```
ansible 2.2.0.0
  config file = ~/.ansible.cfg
```

##### SUMMARY
Add AWS profiles and IAM ARN for AssumeRole support to Credstash lookup plugin.

### DIFF
--- a/lib/ansible/plugins/lookup/credstash.py
+++ b/lib/ansible/plugins/lookup/credstash.py
@@ -41,8 +41,14 @@ class LookupModule(LookupBase):
                 version = kwargs.pop('version', '')
                 region = kwargs.pop('region', None)
                 table = kwargs.pop('table', 'credential-store')
+
+                profile = kwargs.pop('profile', None)
+                arn = kwargs.pop('arn', None)
+
+                session_params = credstash.get_session_params(profile, arn)
+
                 val = credstash.getSecret(term, version, region, table,
-                                          context=kwargs)
+                                          context=kwargs, **session_params)
             except credstash.ItemNotFound:
                 raise AnsibleError('Key {0} not found'.format(term))
             except Exception as e:


dlethin:
This pull request would be very helpful.  Would it be possible to support dynamically resolving variable references in the provided arn.  for example:

```
arn_credstash_reader: "arn:aws:iam::[MYAWSACCOUNT}:role/credstash_reader"
my_secret: "{{ lookup('credstash', 'my.secret.key', arn='{{ arn_credstash_reader }}') }}"
```
The benefit to this is everywhere we use this lookup plugin we wouldn't have to embed our arn -- instead just provide a variable reference to it.  However, this doesn't seem to work -- the variable reference is not resolved and the call to credstash fails.

Note that if I have a variable reference in then name of the key I'm looking up, it works fine, such as:

```
my_secret: "{{ lookup('credstash', 'my.secret.key.{{ env }}', arn='arn:aws:iam::XXXXXXXXXX:role/credstash-reader') }}"
```
I tried to lookup in the ansible python code how to do the string interpolation inside this custom lookup plugin but couldn't figure it out to offer a suggestion.

Thanks.

mkubenka:
@dlethin 

Looks like there are currently 2 options used by Ansible:

1. Variables are parsed from lookup argument ([password lookup](http://docs.ansible.com/ansible/playbooks_lookups.html#the-password-lookup), [csvfile lookup](http://docs.ansible.com/ansible/playbooks_lookups.html#the-csv-file-lookup), ...):

```yaml
password: "{{ lookup('password', '/tmp/passwordfile chars=ascii_letters,digits,hexdigits,punctuation') }}"
```

```yaml
user: "{{ lookup('ini', 'user section=integration file=users.ini') }}"
```

However, this option in not implemented by credstash lookup plugin.

2. Arguments to the lookup become the arguments to [run](https://github.com/ansible/ansible/blob/v2.2.1.0-1/lib/ansible/plugins/lookup/__init__.py#L74) method.

```yaml
my_secret: "{{ lookup('credstash', 'my.secret.key', profile='production') }}"
```

So for your use case you can use:

```yaml
aws_account_id: 123456789
arn_credstash_reader: "arn:aws:iam::{{ aws_account_id }}:role/credstash_reader"
my_secret: "{{ lookup('credstash', 'my.secret.key', arn=arn_credstash_reader) }}"
```

ryansb:
I believe this feature has been provided by #23077 (now merged). Please reopen if there's something not covered by #23077.

PR #18340, opened 2016-11-03
Add fix for s3_bucket in extras subproject

martin-paulus:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
s3_bucket.py

##### ANSIBLE VERSION
```
ansible 2.3.0
```
(current devel branch)

##### SUMMARY
The versioning attribute was not set for new S3 buckets even if you explicitly set the versioning boolean to true ('yes').

This is a regression from an earlier incarnation of the module (in ansible-2.1.1.0). However, that version of the module was unusable due to broken json parsing of the bucket policy.

The current version of the module on the devel branch was broken because the response from Amazon resulted in an empty dictionary. While this module assumed the 'Versioning' attribute was at least present (it checks whether its value != "Enabled").

My suggested fix is to initialize the 'Versioning' attribute to "Disabled" in case the dictionary is empty. This way the rest of the module functions as it should.

```
Relevant output is absent because the module did not fail. Instead versioning was simply not set on newly created S3 buckets.
```


martin-paulus:
Excuse me, I should have made this PR against the submodule in question.
Corrected in: https://github.com/ansible/ansible-modules-extras/pull/3356

PR #18343, opened 2016-11-03
Playbook parsing add error info if tabs are found in the yaml

alikins:
##### ISSUE TYPE

 - Feature Pull Request
 - Bugfix Pull Request

##### COMPONENT NAME

lib/ansible/error
uxd

##### ANSIBLE VERSION

```
ansible 2.3.0 (show_tab_errors_in_yaml 119dd0b0f3) last updated 2016/11/03 13:31:14 (GMT -400)
  lib/ansible/modules/core: (detached HEAD 7cc4d3fe04) last updated 2016/11/03 12:00:46 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD e4bc618956) last updated 2016/11/03 12:00:46 (GMT -400)

```

##### SUMMARY
If a yaml file fails to load because of tabs being used
for formatting, detect that and show a error message
with more details.

Before, a tab would cause a syntax error, and the extended yaml error message would show you
where, but it wouldn't indicate a tab was at fault. So it just pointed at an invisible error.

for ex, before:
```
[fedora-23:ansible (devel % u=)]$ cat bad_tab.yml 

---
# There is a tab in the '- name: ping' line
- name: The first Play
  hosts: localhost
  gather_facts: False
  tasks:
        - name: ping
          ping:
[fedora-23:ansible (devel % u=)]$ ansible-playbook bad_tab.yml 

ERROR! Syntax Error while loading YAML.


The error appears to have been in '/home/adrian/src/ansible/bad_tab.yml': line 7, column 1, but may
be elsewhere in the file depending on the exact syntax problem.

  tasks:
        - name: ping
^ here
```

and after:

```
[fedora-23:ansible (show_tab_errors_in_yaml %)]$ cat bad_tab.yml 

---
# There is a tab in the '- name: ping' line
- name: The first Play
  hosts: localhost
  gather_facts: False
  tasks:
        - name: ping
          ping:
[fedora-23:ansible (show_tab_errors_in_yaml %)]$ ansible-playbook bad_tab.yml 

ERROR! Syntax Error while loading YAML.


The error appears to have been in '/home/adrian/src/ansible/bad_tab.yml': line 7, column 1, but may
be elsewhere in the file depending on the exact syntax problem.

  tasks:
        - name: ping
^ here
There appears to be a tab character at the start of the line.

YAML does not use tabs for formatting. Tabs should be replaced with spaces.

For example:
    - name: update tooling
      vars:
        version: 1.2.3
#    ^--- there is a tab there.

Should be written as:
    - name: update tooling
      vars:
        version: 1.2.3
# ^--- all spaces here.
```

### DIFF
--- a/lib/ansible/errors/__init__.py
+++ b/lib/ansible/errors/__init__.py
@@ -24,7 +24,8 @@ from ansible.errors.yaml_strings import ( YAML_POSITION_DETAILS,
         YAML_COMMON_DICT_ERROR,
         YAML_COMMON_UNQUOTED_COLON_ERROR,
         YAML_COMMON_PARTIALLY_QUOTED_LINE_ERROR,
-        YAML_COMMON_UNBALANCED_QUOTES_ERROR )
+        YAML_COMMON_UNBALANCED_QUOTES_ERROR,
+        YAML_COMMON_LEADING_TAB_ERROR)
 from ansible.module_utils._text import to_native, to_text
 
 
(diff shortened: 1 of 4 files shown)


abadger:
Looks good.  +1 to merge once it passes in shippable.

PR #18344, opened 2016-11-03
Fix setup.py install.

### LINKED ISSUE #18231: ansible install via setup.py fails with 'can't copy 'lib/ansible/galaxy/data/default'' depending on setuptools version
##### ISSUE TYPE

- Bug Report
##### COMPONENT NAME

installing via "python setup.py install"
##### ANSIBLE VERSION

```
git HEAD
```
##### OS / ENVIRONMENT

ubuntu 14.04
##### SUMMARY

installing via "python setup.py install" fails with 'can't copy 'lib/ansible/galaxy/data/default': doesn't exist or not a regular file'

unable to install
##### STEPS TO REPRODUCE

just attempt to install

```

```

##### EXPECTED RESULTS

install going all the way to the end
##### ACTUAL RESULTS

install crash with message in subject

```

```

abadger:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME
setup.py

##### ANSIBLE VERSION

```
devel
```

##### SUMMARY

In setuptools prior to 21.2.1, package_data can't directly reference
a directory.  Modify the entry for galaxy data so that it includes the
files but not the directories

Fixes #18231

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ setup(name='ansible',
       package_dir={ '': 'lib' },
       packages=find_packages('lib'),
       package_data={
-         '': ['module_utils/*.ps1', 'modules/core/windows/*.ps1', 'modules/extras/windows/*.ps1', 'galaxy/data/*'],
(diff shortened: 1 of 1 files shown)


shanemcd:
👍 This works with the version of python-setuptools provided in the repos for all distros we currently build debs for. (xenial, trusty, precise, vivid, and wily)

abadger:
Excellent.  Thanks for testing @shanemcd !

PR #16392, opened 2016-06-22
fixed memoryerror when coping huge file

nyasukun:
##### ISSUE TYPE

- Bugfix Pull Request
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides
```
##### SUMMARY

- vault.is_encrypted accept file object
- dataloader call vault.is_encrypted with file object in order to avoid memory error

Fixes #16391

```
$ ansible-playbook -i hosts huge_copy.yml 

PLAY [all] *********************************************************************

TASK [setup] *******************************************************************
ok: [192.168.80.129]

TASK [Copy huge file] **********************************************************
changed: [192.168.80.129]

PLAY RECAP *********************************************************************
192.168.80.129             : ok=2    changed=1    unreachable=0    failed=0   
```

PR #16329, opened 2016-06-16
If decryption of a vaulted file failed, include the filename in the error

jctanner:
##### ISSUE TYPE
- Bugfix Pull Request
##### ANSIBLE VERSION

```
$ ansible --version
ansible 2.2.0 (FIX_16327 6f38c34112) last updated 2016/06/16 13:40:15 (GMT -400)
  lib/ansible/modules/core: (devel 1c41b6fdb2) last updated 2016/06/16 13:28:28 (GMT -400)
  lib/ansible/modules/extras: (devel 127ddc1f05) last updated 2016/06/16 13:28:32 (GMT -400)
  config file =
  configured module search path = Default w/o overrides
```
##### SUMMARY

Fixes #16327

Issue #16391, opened 2016-06-22
Got MemoryError when copy huge local file to remote


nyasukun:
##### ISSUE TYPE

- Bug Report
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides
```
##### OS / ENVIRONMENT

Ubuntu 14.04.4 LTS
##### SUMMARY

Got MemoryError when copy huge local file to remote.
##### STEPS TO REPRODUCE

```
$ cat huge_copy.yml 

---
- hosts: all
  become: yes

  tasks:
    - name: Copy huge file
      copy: src=5gb dest=dst

$ dd if=/dev/zero of=5gb bs=1G count=5
5+0 records in
5+0 records out
5368709120 bytes (5.4 GB) copied, 206.341 s, 26.0 MB/s

$ ansible-playbook -i hosts huge_copy.yml
```

##### EXPECTED RESULTS

Copied successfully.
##### ACTUAL RESULTS

```
$ ansible-playbook -i hosts huge_copy.yml -vvv
Using /etc/ansible/ansible.cfg as config file

PLAYBOOK: huge_copy.yml ********************************************************
1 plays in huge_copy.yml

PLAY [all] *********************************************************************

TASK [setup] *******************************************************************
<192.168.80.129> ESTABLISH SSH CONNECTION FOR USER: ansible
<192.168.80.129> SSH: EXEC ssh -C -q -o ControlMaster=auto -o ControlPersist=60s -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o User=ansible -o ConnectTimeout=10 -o ControlPath=/home/y/.ansible/cp/ansible-ssh-%h-%p-%r 192.168.80.129 '/bin/sh -c '"'"'( umask 77 && mkdir -p "` echo $HOME/.ansible/tmp/ansible-tmp-1466567672.43-71840509019686 `" && echo ansible-tmp-1466567672.43-71840509019686="` echo $HOME/.ansible/tmp/ansible-tmp-1466567672.43-71840509019686 `" ) && sleep 0'"'"''
<192.168.80.129> PUT /tmp/tmp3Bk56k TO /home/ansible/.ansible/tmp/ansible-tmp-1466567672.43-71840509019686/setup
[...]

PR #18365, opened 2016-11-04
Check encrypted before loading file

ferrants:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
copy

##### ANSIBLE VERSION
```
$ ansible --version
ansible 2.2.0 (devel 15ee97d) last updated 2016/10/03 13:05:00 (GMT +000)
lib/ansible/modules/core: (detached HEAD 7de2872) last updated 2016/07/17 11:35:06 (GMT +000)
lib/ansible/modules/extras: (detached HEAD 68ca157) last updated 2016/07/17 11:35:10 (GMT +000)
```

##### SUMMARY
Checks if the file is vault-encrypted before loading it into memory. Memory issues are hit with copying large files. This fix was put in place https://github.com/ansible/ansible/pull/16392/, looks like this broke again because of https://github.com/ansible/ansible/pull/16329

### DIFF
--- a/lib/ansible/parsing/dataloader.py
+++ b/lib/ansible/parsing/dataloader.py
@@ -173,8 +173,8 @@ class DataLoader():
         show_content = True
         try:
             with open(b_file_name, 'rb') as f:
-                data = f.read()
-                if is_encrypted(data):
+                if is_encrypted(f):
+                    data = f.read()
                     data = self._vault.decrypt(data, filename=b_file_name)
                     show_content = False
 


ferrants:
There was some discussion on https://github.com/ansible/ansible/issues/16391#issuecomment-258497462 about this being an issue now again. I haven't run into it though, just being a good citizen

abadger:
A quick grep and I think the issue is in get_real_file() rather than _get_file_contents().  Solution there is to limit (via the count parameter) how many bytes are read from the file.

abadger:
Okay -- devel and 2.2 was fixed with https://github.com/ansible/ansible/issues/18368

Found out that 2.1 didn't have this fix backported to it.  So I cherry-picked adea1f2b80d806d94ca1bdb2d06f2df077feb948 which is mostly the same as this.  Should be all fixed now.  thanks for opening this PR... We wouldn't have seen that there was more activity on the closed bug without it.

ferrants:
Thanks for the quick responses, @abadger!

PR #18368, opened 2016-11-04
Limit how much of the file we read to test if it's an encrypted vault file

abadger:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME
vault

##### ANSIBLE VERSION

```
devel 2.2
```

##### SUMMARY

Fixes memory errors reported in #16391


abadger:
Confirmed that this fixes the MemoryError reported in  #16391

bdowling:
@abadger Would it not be better to make the required count the default instead of changing this from a caller's perspective?  e.g.

```
+++ b/lib/ansible/parsing/vault/__init__.py
@@ -132,7 +132,7 @@ def is_encrypted(data):
     return False


-def is_encrypted_file(file_obj, start_pos=0, count=-1):
+def is_encrypted_file(file_obj, start_pos=0, count=len(b_HEADER)):
     """Test if the contents of a file obj are a vault encrypted data blob.
```

PR #5454, opened 2013-12-30
Add rax_identity module

claco:
For some tasks, I need to drop the username/api_key into configuration files.
Rather than rely on how I'm calling the rax modules, it seemed more
appropriate to authenticate against Rackspace and return the wealth of
information contained in the pyrax identity payload. Sharing that module for
others.

PR #18371, opened 2016-11-04
Validate apache2 mpm module switching for bug #5328

### LINKED ISSUE #5328: htpasswd module: traceback with badly-formatted (invalid?) htpasswd files
When trying to use the `htpasswd` module with a htpasswd file formatted as below, ansible returns the message "msg: need more than 1 value to unpack". I found this confusing and it took me a long time to work out as I thought the error was with the playbook or variables. The file doesn't seem to cause any problems for nginx basic auth, which increased my confusion.

```

user:cryptpass

```

I'd be happy if ansible just said that there was some error passing the current htpasswd file.

michaelgugino:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
web_infrastructure/apache2_module

##### ANSIBLE VERSION

```
  lib/ansible/modules/core: (detached HEAD 55b1813ebd) last updated 2016/11/01 14:02:20 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD cf524673e1) last updated 2016/11/01 12:39:51 (GMT -400)
  config file = 
```

##### SUMMARY

This patchset updates module integration testing to
validate the changes in PR #5454

This patchset instructs apache2_module to disable all
mpm modules, enable mpm_event, then disable mpm_event
and enable mpm_worker.

Fixes: #5328

### DIFF
--- a/test/integration/targets/apache2_module/tasks/actualtest.yml
+++ b/test/integration/targets/apache2_module/tasks/actualtest.yml
@@ -89,9 +89,88 @@
     force: True
   when: "ansible_os_family == 'Debian'"
 
-
 - name: enable evasive module, test https://github.com/ansible/ansible/issues/22635
   apache2_module:
     name: evasive
     state: present
+  when: "ansible_os_family == 'Debian'"
+
+- name: disable mpm modules
+  apache2_module:
+    name: "{{ item }}"
+    state: absent
+    ignore_configcheck: True
+  with_items:
+    - mpm_worker
+    - mpm_event
+    - mpm_prefork
+  when: "ansible_os_family == 'Debian'"
+
+- name: enabled mpm_event
+  apache2_module:
+    name: mpm_event
+    state: present
+    ignore_configcheck: True
+  register: enabledmpmevent
+  when: "ansible_os_family == 'Debian'"
+
+- name: ensure changed mpm_event
+  assert:
+    that:
+      - 'enabledmpmevent.changed'
+  when: "ansible_os_family == 'Debian'"
+
+- name: switch between mpm_event and mpm_worker
+  apache2_module:
+    name: "{{ item.name }}"
+    state: "{{ item.state }}"
+    ignore_configcheck: True
+  with_items:
+    - name: mpm_event
+      state: absent
+    - name: mpm_worker
+      state: present
+  when: "ansible_os_family == 'Debian'"
+
+- name: ensure mpm_worker is already enabled
+  apache2_module:
+    name: mpm_worker
+    state: present
+  register: enabledmpmworker
+  when: "ansible_os_family == 'Debian'"
+
+- name: ensure mpm_worker unchanged
+  assert:
+    that:
+      - 'not enabledmpmworker.changed'
+  when: "ansible_os_family == 'Debian'"
+
+- name: try to disable all mpm modules with configcheck
+  apache2_module:
+    name: "{{item}}"
+    state: absent
+  with_items:
+    - mpm_worker
+    - mpm_event
+    - mpm_prefork
+  ignore_errors: yes
+  register: remove_with_configcheck
+  when: "ansible_os_family == 'Debian'"
+
+- name: ensure configcheck fails task with when run without mpm modules
+  assert:
+    that:
+      - "{{ item.failed }}"
+  with_items: "{{ remove_with_configcheck.results }}"
+  when: "ansible_os_family == 'Debian'"
+
+- name: try to disable all mpm modules without configcheck
+  apache2_module:
+    name: "{{item}}"
+    state: absent
+    ignore_configcheck: True
+  with_items:
+    - mpm_worker
+    - mpm_event
+    - mpm_prefork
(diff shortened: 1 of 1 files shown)


test/integration/targets/apache2_module/tasks/actualtest.yml
```diff
@@ -59,3 +59,36 @@
   assert:
     that:
       - 'disablefinal.changed'
+
+- name: disable mpm modules
+  apache2_module: name={{ item }} state=absent
```
robinro: Ansible now recommends yaml format:

```
apache2_module:
  name: "{{item}}"
  state: absent
```

test/integration/targets/apache2_module/tasks/actualtest.yml
```diff
@@ -59,3 +59,36 @@
...
+  with_items:
+    - mpm_prefork
+    - mpm_event
+    - mpm_worker
+
+- name: enabled mpm_event
+  apache2_module: name=mpm_event state=present check_config=1
```
robinro: Use `True` or `False` for boolean values.

test/integration/targets/apache2_module/tasks/actualtest.yml
robinro: Needs `check_config: False`

robinro (changes requested):
Please also add a test that tries to remove all mpm modules with `check_mode: True` and then fails.

Use `failed_when: False` and then check `result|failed`.

robinro:
I'd suggest to add a test for the current behavior that fails to remove all modules. See also https://github.com/ansible/ansible-modules-core/pull/5454

```
- name: try to disable all mpm modules with configcheck
  apache2_module:
    name: "{{item}}"
    state: absent
  with_items:
    - mpm_event
    - mpm_worker
    - mpm_prefork
  failed_when: False
  register: remove_with_configcheck

- name: ensure configcheck fails task
  assert:
    that:
      - "remove_with_configcheck.failed"

- name: try to disable all mpm modules without configcheck
  apache2_module:
    name: "{{item}}"
    state: absent
    ignore_configcheck: True
  with_items:
    - mpm_event
    - mpm_worker
    - mpm_prefork
  failed_when: False
  register: remove_without_configcheck

- name: ensure without configcheck task succeeds
  assert:
    that:
      - "not remove_with_configcheck.failed"
```

robinro:
Also test `switch between mpm_event and mpm_worker` both ways, once first remove then add, once first add then remove. a2{en/dis}mod fails for different reasons in those cases.

jctanner:
@michaelgugino does this test work on redhat/centos/fedora?

needs_info

michaelgugino:
Yes, I believe it should work against all supported distros.  I cannot recall if I tested it against RH.

robinro:
@jctanner The apache2_module module only supports debian and suse type distros.

@michaelgugino  I had some some more suggestions for tests, can you implement them or comment on them?

michaelgugino:
@robinro I think the suggestions are good.  I will get an updated patch out later this week.

gundalow:
Closing & reopening to trigger CI

gundalow:
Module needs updating to support opensuse, see failing integration tests

michaelgugino:
I cannot recall if the mpm specific portions are actually applicable to OpenSUSE.  I seem to remember apache only shipping with one mpm module in that distro.  Thoughts?

gundalow:
1) Update module so it works on OpenSUSE - preferred
2) Update tests so they skip on OpenSUSE (conditional on OS)

gundalow:
Failing on OpenSuse still
https://app.shippable.com/github/ansible/ansible/runs/24887/17/tests

michaelgugino:
@robinro 
I have confirmed locally with OpenSuSe 42.2 that apache2 does not ship with mpm_event or mpm_worker.  The only available module is mpm_prefork.  Thus, the checks are failing as expected because the desired module mpm_event is never present in the stdout of apache2ctl -M.

Please let me know your thoughts.

mattclay:
@michaelgugino Are you planning on making any additional changes? I see that @robinro requested testing with check mode, which would be good to have, but I don't see that included in the tests.

michaelgugino:
@mattclay I'm not sure that's what was suggested.  A new feature was added to this module called 'ignore_configcheck' to allow disabling validation of the apache configs during a run.  This feature was developed as a fix to bug-5328, and this bug seems to really only impact Debian based distros.

I'm happy to add more to the tests, but I personally don't feel that check-mode testing is 100% necessary here.

mattclay:
@michaelgugino The first [comment](https://github.com/ansible/ansible/pull/18371#pullrequestreview-7320347) from @robinro requested testing check mode.

Having check mode tests is important for making sure it actually works. There have been issues where check mode is reported by a module as supported, yet it either fails or actually makes changes.

With that being said, I don't think we should block this PR due to lack of testing check mode. Those tests can be added in a follow-up PR if anyone is interested.

mattclay:
@michaelgugino Thanks for the additional tests.

PR #18373, opened 2016-11-04
Loader get file contents bytes 18355

### LINKED ISSUE #18355: Ansible Vault depcryption failing with ansible 2.2
- Bug Report
##### COMPONENT NAME

Ansible Version: ansible 2.2.0.0
Cofiguration: ansible.cfg(in present directory)

OS: Ubuntu 14.04 LTS

ansible-playbook not able to decrypt the files which encrypted with ansible 2.1.0.0 & below.

Steps to reproduce:

1. Install ansible 2.1.0.0 or below
2. Encrypt an inventory file
3. Add vault_password_file in ansible.cfg
4. Run, ansible-playbook -i inventory/qa site.yml which works fine.
5. Now install ansible 2.2.0.0
6. Run, ansible-playbook -i inventory/qa site.yml which gives below error.

fatal: [$ANSIBLE_VAULT;1.1;AES256]: UNREACHABLE! => {
    "changed": false,
    "msg": "Failed to connect to the host via ssh: ssh: Could not resolve hostname $ansible_vault;1.1;aes256: Name or service not known\r\n",
    "unreachable": true
}

##### EXPECTED RESULTS
Plays has to start executing.

##### ACTUAL RESULTS
Get below error:
fatal: [$ANSIBLE_VAULT;1.1;AES256]: UNREACHABLE! => {
    "changed": false,
    "msg": "Failed to connect to the host via ssh: ssh: Could not resolve hostname $ansible_vault;1.1;aes256: Name or service not known\r\n",
    "unreachable": true
}

alikins:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

lib/ansible/parsing/dataloader.py
lib/ansible/inventory

##### ANSIBLE VERSION

```
ansible 2.3.0 (loader_get_file_contents_bytes_18355 f3b9b5cfae) last updated 2016/11/04 17:35:53 (GMT -400)
  lib/ansible/modules/core: (detached HEAD 2584fca0ae) last updated 2016/11/04 13:15:56 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD a1dcbf9ce5) last updated 2016/11/04 13:15:56 (GMT -400)

```

##### SUMMARY
This is a fix for devel. Will need to be backported to stable-2.2

   Make DataLoader._get_file_contents return bytes
    
    This is first part of fix for #18355
    
    The issue #18355 is caused by a change to inventory to
    stop using _get_file_contents so that it can handle text
    encoding itself to better protect against harmless text
    encoding errors in ini files (invalid unicode text in
    comment fields).
    
    So this makes _get_file_contents return bytes so it and other
    callers can handle the to_text().
    
    The data returned by _get_file_contents() is now a bytes object
    instead of a text object. The callers of _get_file_contents() have
    been updated to call to_text() themselves on the results.
    
    Fixes #18355 (partially)

    Fix bug (#18355) where encrypted inventories fail
    
    Previously, the ini parser attempted to work around
    ini files that potentially include non-vailid unicode
    in comment lines. To do this, it stopped using
    DataLoader._get_file_contents() which does the decryption of
    files if vault encrypted. It didn't use that because _get_file_contents
    previously did to_text() on the read data itself.
    
    _get_file_contents() returns a bytestring now, so ini.py
    can call it and still special case ini file comments when
    converting to_text(). That also means encrypted inventory files
    are decrypted first.
    
    Fixes #18355

### DIFF
--- a/lib/ansible/inventory/ini.py
+++ b/lib/ansible/inventory/ini.py
@@ -51,18 +51,21 @@ class InventoryParser(object):
 
         # Read in the hosts, groups, and variables defined in the
         # inventory file.
+        if loader:
+            (b_data, private) = loader._get_file_contents(filename)
+        else:
+            with open(filename, 'rb') as fh:
+                b_data = fh.read()
 
-        with open(filename, 'rb') as fh:
-            data = fh.read()
-            try:
-                # Faster to do to_text once on a long string than many
-                # times on smaller strings
-                data = to_text(data, errors='surrogate_or_strict')
-                data = [line for line in data.splitlines() if not (line.startswith(u';') or line.startswith(u'#'))]
-            except UnicodeError:
-                # Skip comment lines here to avoid potential undecodable
-                # errors in comments: https://github.com/ansible/ansible/issues/17593
-                data = [to_text(line, errors='surrogate_or_strict') for line in data.splitlines() if not (line.startswith(b';') or line.startswith(b'#'))]
+        try:
+            # Faster to do to_text once on a long string than many
+            # times on smaller strings
+            data = to_text(b_data, errors='surrogate_or_strict')
+            data = [line for line in data.splitlines() if not (line.startswith(u';') or line.startswith(u'#'))]
+        except UnicodeError:
+            # Skip comment lines here to avoid potential undecodable
+            # errors in comments: https://github.com/ansible/ansible/issues/17593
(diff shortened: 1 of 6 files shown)


abadger:
Had one question: Whether vault.decrypt returns bytes now.  If it does, then this looks good to merge.  If it doesn't then we need to convert from text to bytes in _get_file_contents() Line 181.

PR #18376, opened 2016-11-05
Fix media_type detection on NetBSD

mscherer:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
setup
##### SUMMARY

On NetBSD 7.0.1, ifconfig return this:

    $ ifconfig
    ne0: flags=8863<UP,BROADCAST,NOTRAILERS,RUNNING,SIMPLEX,MULTICAST> mtu 1500
            ec_capabilities=1<VLAN_MTU>
            ec_enabled=0
            address: 00:20:91:45:00:78
            media: Ethernet 10baseT full-duplex
            inet 192.168.156.29 netmask 0xffffff00 broadcast 192.168.156.255

Which result into setup returning:

    "media_type": "ull-duplex",

I couldn't find a good reason to skip the first char on the string looking
at the output of ifconfig on FreeBSD and NetBSD, nor imagine why it was done
like this.

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -2950,6 +2950,22 @@ class NetBSDNetwork(GenericBsdIfconfigNetwork):
     """
     platform = 'NetBSD'
 
+    def parse_media_line(self, words, current_if, ips): 
+        # example of line:
+        # $ ifconfig
+        # ne0: flags=8863<UP,BROADCAST,NOTRAILERS,RUNNING,SIMPLEX,MULTICAST> mtu 1500
+        #    ec_capabilities=1<VLAN_MTU>
+        #    ec_enabled=0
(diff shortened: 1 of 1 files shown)


mscherer:
So I found the reason for that weird trick.

On FreeBSD 11.0:

```
    media: Ethernet autoselect (100baseTX <full-duplex>)
```

On NetBSD 7.0.1:

```
    media: Ethernet 10baseT full-duplex
```

So I guess the right fix is to split that method on a per variant basis.

hccuneyt:
ansible 2.2.2.0,  FreeBSD 11, media type is detected as "full-duplex>", there is an extra ">" . this might be related issues.

mscherer:
@hccuneyt Can you test on 2.3.0, and if that's still a issue, fill a new issue ?

PR #18379, opened 2016-11-05
Feature/lookup keyring

scboucher:
##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

- lookup/keyring.py

##### ANSIBLE VERSION

```
ansible 2.2.0.0
  config file =
```

##### SUMMARY

Add a lookup keyring.py
I believe it would be useful to access the OS keychain directly to  get password  or token.
Hardcoding them in the playbook or leaving them on the filesystem is not always the best solution.
It uses python keyring module. 
This lookup access the keyring  to fetch the secret.

```
N/A
```

I tried the lookup on both OS X  and Fedora with Gnome

### DIFF
--- a/lib/ansible/plugins/lookup/keyring.py
+++ b/lib/ansible/plugins/lookup/keyring.py
@@ -1,0 +1,43 @@
+# (c) 2016, Samuel Boucher <boucher.samuel.c@gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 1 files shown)


scboucher:
From my understanding, there is a problem with the fedora 23  test setup it tries to use yum instead of dnf

```
"msg": "Yum command has been deprecated, redirecting to '/usr/bin/dnf -d 2 -y install sos'.\nSee 'man dnf' and 'man yum2dnf' for more information.\nTo transfer transaction metadata from yum to DNF, run:\n'dnf install python-dnf-plugins-extras-migrate && dnf-2 migrate'\n\nError: Failed to synchronize cache for repo 'updates'\n", 
"rc": 1, 
```

PR #18387, opened 2016-11-06
Add DMI facts on NetBSD

mscherer:
##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME
setup
##### SUMMARY

The PR refactor part of the code used by OpenBSD (and use it), make it portable on NetBSD to support DMI facts. I think I could refactor it a bit more, but I am not sure if that's worth, given that OpenBSD also use sysctl for others things, and this did make the code slightly more complex when I attempt, without (IMHO) much gain.

Tested on OpenBSD 6.0 and NetBSD 7.0.1

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -1831,7 +1831,7 @@ class DragonFlyHardware(FreeBSDHardware):
     platform = 'DragonFly'
 
 
-class NetBSDHardware(Hardware):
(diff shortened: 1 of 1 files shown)


mscherer:
Further refactoring could be done with the class used on Darwin, who have also a improved version of get_sysctl()

mscherer:
So I will close this one in favor of https://github.com/ansible/ansible/pull/18388

PR #18388, opened 2016-11-06
Add DMI facts on NetBSD and refactor code

mscherer:
##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME
setup
##### SUMMARY

So after sending https://github.com/ansible/ansible/pull/18387 , I took a look if the code could be refactored, and it seems that Darwin also use sysctl to detect various things. So I refactored the method to run on OpenBSD and remove the function. And added NetBSD while on it.


bcoca:
I would just make it part of the base class as most systems use it to get info

mscherer:
Ok, done.

mscherer:
Unless you wanted in another class than Hardware maybe ?

PR #18399, opened 2016-11-07
Text's .translate() is easier to use than bytes

abadger:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME
facts.py

##### ANSIBLE VERSION

```
devel 2.2
```

##### SUMMARY

Text strings and byte strings both have a translate method but the byte
string version is harder to use.  It requires a mapping of all 256 bytes
to a translation value.  Text strings only require a mapping from the
characters that are changing to the new string.  Switching to text
strings on both py2 and py3 allows us to state what we're getting rid of
simply without having to rely on the maketrans() helper function.


abadger:
Merged to devel and stable-2.2

PR #18408, opened 2016-11-08
openstack: iterate through nova addresses with six

bencord0:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
`ansible.module_utils.openstack`
part of the `os_server` module

##### ANSIBLE VERSION
```
ansible 2.2.0.0
  config file = 

```

##### SUMMARY
Allow the `os_server` module to work under python3.

### DIFF
--- a/lib/ansible/module_utils/openstack.py
+++ b/lib/ansible/module_utils/openstack.py
@@ -28,6 +28,7 @@
 
 import os
 
+from ansible.module_utils.six import iteritems
 
 def openstack_argument_spec():
     # DEPRECATED: This argument spec is only used for the deprecated old
@@ -61,7 +62,7 @@ def openstack_argument_spec():
 def openstack_find_nova_addresses(addresses, ext_tag, key_name=None):
 
     ret = []
-    for (k, v) in addresses.iteritems():
+    for (k, v) in iteritems(addresses):
         if key_name and k == key_name:
             ret.extend([addrs['addr'] for addrs in v])
         else:


lib/ansible/module_utils/openstack.py
```diff
@@ -28,6 +28,7 @@
 
 import os
 
+from ansible.compat.six import iteritems
```
resmo: this should import from because module_utils will be transfered to remote target
`from ansible.module_utils.six import iteritems`

lib/ansible/module_utils/openstack.py
```diff
@@ -61,7 +62,7 @@ def openstack_argument_spec():
 def openstack_find_nova_addresses(addresses, ext_tag, key_name=None):
 
     ret = []
-    for (k, v) in addresses.iteritems():
+    for (k, v) in iteritems(addresses):
```
resmo, line 65: this should not be changed IMHO

abadger: @resmo Is there a reason you don't like this?  To work on python3 we either need to do this or:

``` python
for (k, v) in addresses.items():
```

Either way is probably okay.

resmo: Oh, sorry, my fault. I mixed up items() and iteritems(). All fine!

abadger:
Merged to devel and cherry-picked to stable-2.2 for the 2.2.1 release.

PR #18439, opened 2016-11-09
Add a check for type() instead of isinstance()

### LINKED ISSUE #18310: write code smell test for use of type(foo) ==
##### ISSUE TYPE

 - Feature Idea

##### COMPONENT NAME

code smell tests

##### ANSIBLE VERSION

```
2.3
```

##### SUMMARY

per https://github.com/ansible/ansible-modules-extras/pull/2977 modules should use isinstance instead of "type(foo) ==" for comparison.

Need to write a script to check for the string and place in https://github.com/ansible/ansible/tree/devel/test/sanity/code-smell

jctanner:
##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME
valide-modules script

##### ANSIBLE VERSION

```
2.3
```

##### SUMMARY
Fixes #18310


test/sanity/validate-modules/fix_isinstance.py
```diff
@@ -0,0 +1,141 @@
...
+    (so, se) = p.communicate()
+    return (p.returncode, so, se)
+
+def find_modules(basepath):
+    cmd = 'find %s -type f -name "*.py"' % basepath
+    (rc, so, se) = run_command(cmd)
+    files = [x.strip() for x in so.split('\n') if x.strip()]
+    files = [x for x in files if not x.endswith('__init__.py')]
+    files = [x for x in files if not '.git' in x]
+    files = [x for x in files if not '/test/' in x]
```
abadger: Just a neat python thing that you might be unaware of.  Python has generator expressions which are a quick way to create and chain iterators.  They can act similar to shell pipelines if you don't need actual lists but can just act on iterators the whole way through.  It saves on memory usage sometimes.  Rewritten as generator expressions, this becomes:

``` python
+    files = (x.strip() for x in so.split('\n') if x.strip())
+    files = (x for x in files if not x.endswith('__init__.py'))
+    files = (x for x in files if not '.git' in x)
+    files = [x for x in files if not '/test/' in x]
```

The last one remains a list comprehension so that you can return a list rather than an iterator.
(This is just informational, not a blocker.)

test/sanity/validate-modules/fix_isinstance.py
```diff
@@ -0,0 +1,141 @@
...
+import sys
+import subprocess
+import tempfile
+
+TYPE_REGEX = re.compile(r'.*\stype\(.*')
+r1 = re.compile(r'.*\stype\(.*\)\s+is\s+(int|bool|str|dict|list|tuple|MethodType|vim\..*):')
+r2 = re.compile(r'.*\stype\(.*\)\s+is\snot\s+(int|bool|str|dict|list|tuple|MethodType|vim\..*):')
+r3 = re.compile(r'.*\stype\(.*\)\s+==\s+(int|bool|str|dict|list|tuple|MethodType|vim\..*):')
+r4 = re.compile(r'.*\stype\(.*\)\s+!=\s+(int|bool|str|dict|list|tuple|MethodType|vim\..*):')
+r5 = re.compile(r'.*\stype\(.*\)\s+is\s+not\s+(int|bool|str|dict|list|tuple|MethodType|vim\..*):')
```
abadger: Maybe some comments with an example of what each of these matches.

abadger:
Really minor things not required for merge.  This looks good otherwise.  +1 from me.

abadger:
Ah, found somethingto add:  List the new test in the README file: https://github.com/ansible/ansible/blob/devel/test/sanity/validate-modules/README.rst

mattclay:
@jctanner What is the purpose of having `fix_isinstance.py` in the repo? We won't need it for modules in core or extras, since we'll need to fix those before this can be merged.

jctanner:
@mattclay it's just a helper that I used to fix all the modules in core+extras already.

mattclay:
@jctanner Ah, I hadn't noticed the fixes were already in core/extras because the submodule refs hadn't been updated yet. I've done that. Now that they've all been updated, is there any reason to keep `fix_isinstance.py` around?

jctanner:
@mattclay that's up to you. If not in ansible/ansible, I'll just move it to jctanner/ansible-tools

Issue #35, opened 2012-03-14
refactor runner, playbooks python code (+ runner script)


mpdehaan:
Lots of room for cleanup, some stuff belongs in utils

mpdehaan:
runner is cleaned up, playbooks have some work to do yet

mpdehaan:
done

PR #18445, opened 2016-11-09
namespace facts

bcoca:
##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

facts

##### ANSIBLE VERSION

```
2.3
```

### DIFF
--- a/docs/docsite/rst/intro_configuration.rst
+++ b/docs/docsite/rst/intro_configuration.rst
@@ -768,7 +768,8 @@ always default to the current user if this is not defined::
 restrict_facts_namespace
 ========================
 
-.. versionadded:: 2.3
+.. versionadded:: 2.4
+
 This allows restricting facts in their own namespace (under ansible_facts) instead of pushing them into the main.
 False by default, can also be set via the environment variable `ANSIBLE_RESTRICT_FACTS`.Using `ansible_system` as an example.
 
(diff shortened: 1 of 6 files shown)


mscherer:
So, I think I understand the feature, but I am curious on the use case.

abadger:
I like it.  We'd need to decide whether to deprecate or just leave it as an option forwever.

bcoca:
@mscherer it avoids collisions with existing vars by making sure all facts are restricted to `ansible_facts`.

mscherer:
What is annoying is that suddenly, it reduce the reuse of roles. We had the discussion with my teammate on hash_behaviour, because if he turn it on for ansible, then I can't reuse some of his role. the same go for that feature, so I wonder if we can do something to prevent and/or avoid that.

For example, could we activate/desactivate it on a per role basis, etc ?

bcoca:
That issue is solved via https://github.com/ansible/proposals/issues/35

mscherer:
The proposal #35 may not solve the issue. If I have a role that expect facts to be in their own namespace and another that don't, I can't use both of them, because the setting applie only for setup.

bcoca:
it allows each role to merge it's own config params

dharmabumstead (approved):
Doc update approved with minor edit. Thanks.

PR #18454, opened 2016-11-10
Adds `follow` parameter to `template` action

### LINKED ISSUE #14469: Add 'follow' option to 'template'
##### Issue Type:

Bug report in Core Module
##### Ansible Version:

ansible --version:

```
ansible 2.1.0 (devel 4b953c4b16) last updated 2016/02/14 15:32:35 (GMT +200)
  lib/ansible/modules/core: (detached HEAD 8d126bd877) last updated 2016/02/14 15:32:42 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD f6c5ed987f) last updated 2016/02/14 15:32:48 (GMT +200)
```
##### Ansible Configuration:

Summary of changes:
- forks increased to 50
- vault_password_file defined
- gathering set to smart
- timeout set to 30
- plugins/actions added to action_plugins
- fact_caching set to jsonfile
- `-o BatchMode=yes -o UserKnownHostsFile=/dev/null` added to ssh_args
- pipelining set to true
##### Environment:

Running Ansible from CentoS 7, managing Ubuntu 14.04 LTS.
##### Summary:

https://docs.ansible.com/ansible/template_module.html suggests the existence of a `follow` option to disable following symlinks when creating files from templates.

Yet, this does not seem to be working as intended.
##### Steps To Reproduce:

e.g.

``` yaml
- name: set up resolv.conf
  template:
    src: resolv.conf.j2
    dest: /etc/resolv.conf
    follow: no
```

Then run this on an environment where /etc/resolv.conf is a symlink (e.g. to /run/resolvconf/resolv.conf).
##### Expected Results:

/etc/resolv.conf is replaced by a file containing the evaluated template contents.
##### Actual Results:

e.g.

```
fatal: [example-host]: FAILED! => {"changed": true, "failed": true, "msg": "Destination /run/resolvconf not writable"}
```
##### Further Info:

I can work around this issue like so:

``` yaml
- name: check if resolv.conf is a symlink
  stat: path=/etc/resolv.conf
  register: resolvconf

- name: delete resolv.conf if it's a symlink
  file:
    path: /etc/resolv.conf
    state: absent
  when: resolvconf.stat.islnk

- name: set up resolv.conf
  template:
    src: resolv.conf.j2
    dest: /etc/resolv.conf
```

But this is actually rather bothersome because without a proper resolv.conf set up, it's possible for every task to take a long time to execute on the remote end.

fuhbar:
##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

Action: `template`

##### ANSIBLE VERSION

Tested against current devel branch.

```
ansible-playbook --version
ansible-playbook 2.3.0 (feature/add-follow-to-template-action 36b7db8fc5) last updated 2016/11/10 17:08:11 (GMT +200)
  lib/ansible/modules/core: (detached HEAD 2584fca0ae) last updated 2016/11/10 12:07:13 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD a1dcbf9ce5) last updated 2016/11/10 12:07:19 (GMT +200)
```

##### SUMMARY

This PR introduces the `follow` option to the `template` action plugin. The desired behaviour as reported in #14469 is the same as of the `copy` and `file` module, since this action plugin just relies on these modules. This means, if this template is applied to a link with `follow` set to `False` then the link will be converted to a regular file containing the template content. 

Fixes #14469 

Using this playbook:
```
- hosts: all
  vars:
    greetings: "Hello f*ck*ng hell!"
  tasks: 
    - name: test
      template: 
        src: "test"
        dest: "/tmp/link"
        follow: False
```

... and this template file:
```
{{ greetings }}
```

The following output describes the new behaviour:
```
$ echo "Hello World" > /tmp/source
$ ln -fs /tmp/source /tmp/link

$ ls -la /tmp/link 
lrwxrwxrwx 1 foo foo 11 Nov 10 17:18 /tmp/link -> /tmp/source

$ ansible-playbook -i inventory test.yml  -v

PLAY [all] *******************************************************************************************

TASK [Gathering Facts] *******************************************************************************
ok: [localhost]

TASK [test] ******************************************************************************************
changed: [localhost] => {"changed": true, "checksum": "7bc2660f74cfd11b442505e28b54be611cc013cd", "dest": "/tmp/link", "gid": 1000, "group": "fd", "md5sum": "6ecc0909e828485c6363d203942f580e", "mode": "0664", "owner": "fd", "size": 26, "src": "/home/fd/.ansible/tmp/ansible-tmp-1478794956.02-5427137697491/source", "state": "file", "uid": 1000}

PLAY RECAP *******************************************************************************************

$ ls -la /tmp/link
-rw-rw-r-- 1 foo foo 26 Nov 10 17:22 /tmp/link

$ cat /tmp/link
Hello f*ck*ng hell!
```

Whereas the default of this parameter is `True` to retain backward compatibility.

### DIFF
--- a/lib/ansible/plugins/action/template.py
+++ b/lib/ansible/plugins/action/template.py
@@ -36,14 +36,14 @@ class ActionModule(ActionBase):
 
     TRANSFERS_FILES = True
 
-    def get_checksum(self, dest, all_vars, try_directory=False, source=None, tmp=None):
+    def get_checksum(self, dest, all_vars, try_directory=False, source=None, tmp=None, follow=False):
         try:
             dest_stat = self._execute_remote_stat(dest, all_vars=all_vars, follow=False, tmp=tmp)
 
             if dest_stat['exists'] and dest_stat['isdir'] and try_directory and source:
                 base = os.path.basename(source)
                 dest = os.path.join(dest, base)
-                dest_stat = self._execute_remote_stat(dest, all_vars=all_vars, follow=False, tmp=tmp)
+                dest_stat = self._execute_remote_stat(dest, all_vars=all_vars, follow=follow, tmp=tmp)
 
         except AnsibleError:
             return dict(failed=True, msg=to_native(get_exception()))
@@ -62,6 +62,7 @@ class ActionModule(ActionBase):
         dest   = self._task.args.get('dest', None)
         force  = boolean(self._task.args.get('force', True))
         state  = self._task.args.get('state', None)
+        follow = self._task.args.get('follow', True)
 
         if state is not None:
             result['failed'] = True
@@ -178,7 +179,7 @@ class ActionModule(ActionBase):
                         src=xfered,
                         dest=dest,
                         original_basename=os.path.basename(source),
-                        follow=True,
+                        follow=follow,
                         ),
                 )
                 result.update(self._execute_module(module_name='copy', module_args=new_module_args, task_vars=task_vars, tmp=tmp, delete_remote_tmp=False))
@@ -197,7 +198,7 @@ class ActionModule(ActionBase):
                 dict(
                     src=None,
                     original_basename=os.path.basename(source),
-                    follow=True,
+                    follow=follow,
                 ),
             )
             result.update(self._execute_module(module_name='file', module_args=new_module_args, task_vars=task_vars, tmp=tmp, delete_remote_tmp=False))


fuhbar:
Keep in mind that the [`action` module](https://github.com/ansible/ansible-modules-core/blob/devel/files/template.py) in [ansible-modules-core](https://github.com/ansible/ansible-modules-core) needs to be updated as well, since it serves as some kind of a placeholder for the documentation.

jannooo:
Are there any plans to merge this PR into devel?

fuhbar:
@elsobrino Dunno, we should ping any of the maintainers.

fuhbar:
But in order to get this into devel branch I've rebased the commit onto the most current devel branch. So whoever wants to merge this in, this commit seems to be prepared. :)

jhawkesworth:
Would this work with windows (windows equivalents of file and copy not quite as advanced as the linux equivalents)?  (win_template and template action plugins share all the same logic by the way).

fuhbar:
@jhawkesworth good question! I haven't dig into the `win_template` code yet. But would it be a dealbreaker if this `follow` parameter will be not available to `win_template`? Since both actions/modules do not share feature parity anyway I don't consider this as a problem.

But generally spoken, Windows - resp. NTFS - does support symbolic links and it should somehow be possible to implement this.

jhawkesworth:
I think as long as its clear in the docs where this functionality is available, that's ok (although it would be nice if it was actually implemented for windows too I suppose - feature parity is always something we should aim at).  Actually it would be good to add something to the template documentation page saying that 'follow' was added in this version of ansible, so I'd want to see that in this PR.

Since the docs should be updated, I'm flagging this as 

needs_revision

calfonso:
@fuhbar could you rebase this and we'll review. Thanks!

maxamillion:
Thank you for the PR, but this was implemented in Ansible 2.4. Closing

PR #18474, opened 2016-11-12
Assume template file contains variables when override header present

### LINKED ISSUE #18192: Not using `{{` or `}}` in template file skips Jinja2 even when markers are overriden
Although this is an issue exhibited in a module, I suspect the code that causes this exists in this repository
##### ISSUE TYPE

- Bug Report
##### COMPONENT NAME

Jinja2 templating
##### ANSIBLE VERSION

```
ansible 2.1.2.0
  config file = 
```
##### CONFIGURATION

```
[defaults]
inventory = ./hosts
```
##### OS / ENVIRONMENT

**N/A** but Arch Linux

```
Linux [...] 4.8.4-1-ARCH #1 SMP PREEMPT Sat Oct 22 18:26:57 CEST 2016 x86_64 GNU/Linux
```
##### SUMMARY

Not using `{{` or `}}` at any point in a Jinja2 template causes the file not to be run through Jinja2, even when `variable_start_string` and `variable_end_string` is provided (and used throughout the file)
##### STEPS TO REPRODUCE

- Create a Jinja2 file with the following content
- Use the `template` directive

```
#jinja2: variable_start_string: '[%', variable_end_string: '%]'
" [% ansible_managed %]
```

Since neither `{{` nor `}}` occurs in the file, it is not run through Jinja2. Adding either will cause the expected results to appear (with the `{{` or `}}` in the rendered output)

##### EXPECTED RESULTS

```
" Ansible managed: [...]
```
##### ACTUAL RESULTS

```
#jinja2: variable_start_string: '[%', variable_end_string: '%]'
" [% ansible_managed %]
```

saleemrashid:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

Ansible (`lib/ansible/template`)

##### ANSIBLE VERSION

```
ansible 2.3.0 (jinja2-override-no-skip fea57a43bd) last updated 2016/11/12 10:59:59 (GMT +100)
  lib/ansible/modules/core: (detached HEAD 8d1fe3b444) last updated 2016/11/12 10:39:31 (GMT +100)
  lib/ansible/modules/extras: (detached HEAD 88e58df86b) last updated 2016/11/12 10:39:31 (GMT +100)
```

##### SUMMARY

When checking if the file should be run through the templating engine, Ansible checks whether it contains any of the the Jinja2 markers (`{{`, `}}` or `{#`) otherwise it won't call the template engine

If you use the Jinja2 override header to override any of these markers, Ansible won't take these changes into consideration at this stage.

However, rather than reading the header and checking _those_ markers instead, I chose to use the presence of the header as an indicator of whether to use the templating engine.

The use of the header shows that the user obviously wants the file passed through the templating engine and, at the very least, it should be stripped. However, it's more than likely that the file will contain variables using that marker. 

Fixes #18192

### DIFF
--- a/lib/ansible/template/__init__.py
+++ b/lib/ansible/template/__init__.py
@@ -385,6 +385,10 @@ class Templar:
         returns True if the data contains a variable pattern
         '''
         if isinstance(data, string_types):
+            # Assume that it contains variables since the user obviously wants it passed through the templating engine
+            if data.startswith(JINJA2_OVERRIDE):
+                return True
+
             for marker in (self.environment.block_start_string, self.environment.variable_start_string, self.environment.comment_start_string):
                 if marker in data:
                     return True
--- a/test/integration/targets/template/files/override.txt
+++ b/test/integration/targets/template/files/override.txt
@@ -1,0 +1,11 @@
+templated_var_loaded
+
+{
+    "bool": true,
+    "multi_part": "1Foo",
+    "null_type": null,
+    "number": 5,
+    "string_num": "5"
+}
+
+{{ templated_var }}
--- a/test/integration/targets/template/tasks/main.yml
+++ b/test/integration/targets/template/tasks/main.yml
@@ -20,6 +20,10 @@
   template: src=foo.j2 dest={{output_dir}}/foo.templated mode=0644
   register: template_result
 
+- name: fill in a basic template with an override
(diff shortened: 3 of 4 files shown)


saleemrashid:
I have noticed that the integration tests I added would have passed without this patch 😭

PR #18488, opened 2016-11-14
Add a tool to work with module metadata

abadger:
##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME
hacking script to modify metadata 

##### ANSIBLE VERSION

```
2.3
```

##### SUMMARY
This commit supplies a script that can add and modify module metadata.

### DIFF
--- a/hacking/metadata-tool.py
+++ b/hacking/metadata-tool.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python2 -tt
+#!/usr/bin/env python
 
 import ast
 import csv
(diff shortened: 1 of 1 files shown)


hacking/metadata-tool.py
```diff
@@ -0,0 +1,510 @@
...
+
+    Print out all the modules that have metadata and all the ones that do not.
+
+    :kwarg version: If given, the metadata must be at least this version.
+        Otherwise return it as not having metadata
+    """
+    # List of all plugins
+    plugins = module_loader.all(path_only=True)
+    plugins = ((os.path.splitext((os.path.basename(p)))[0], p) for p in plugins)
+    plugins = (p for p in plugins if p[0] != 'async_wrapper')
```
mscherer: Is there a reason to hardcode it and not reuse NONMODULE_PY_FILES ?

abadger: Yeha, I probably shouldn't hardcode it.... Have to store an extensionless version for this, though.  (async_wrapper.py vs async_wrapper)

abadger: Okay, changed it so that it isn't hardcoded within the code.

PR #18493, opened 2016-11-15
Change old recommendation about insecure variables

silviot:
##### ISSUE TYPE
 - Docs Pull Request

##### COMPONENT NAME
Introduction to inventory

##### ANSIBLE VERSION
```
ansible 2.3.0
```

##### SUMMARY
Change warnings about intrinsically insecure variables `ansible_ssh_pass` and `ansible_become_pass`: they were relevant before ansible had the secure vault feature, since any secret put into a variable used to be a bad idea.

With the vault feature (available since ansible 1.5) it's no longer a bad idea to use these variables, as long as they aren't stored in plain text.

### DIFF
--- a/docsite/rst/intro_inventory.rst
+++ b/docsite/rst/intro_inventory.rst
@@ -218,7 +218,7 @@ ansible_port
 ansible_user
     The default ssh user name to use.
 ansible_ssh_pass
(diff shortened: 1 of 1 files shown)


dharmabumstead:
Thanks @silviot!

PR #18517, opened 2016-11-17
added warnings list to module and autoadd

bcoca:
##### ISSUE TYPE

 - Feature Pull Request
 
##### COMPONENT NAME

module_utils

##### ANSIBLE VERSION

```
2.3
```

##### SUMMARY

added better way of adding warnings to return data
backwards compatible if warnings key already exists

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -684,6 +684,9 @@ class AnsibleModule(object):
         # May be used to set modifications to the environment for any
         # run_command invocation
         self.run_command_environ_update = {}
+        self._warnings = []
+        self._deprecations = []
+        self._passthrough = ['warnings', 'deprecations']
 
         self.aliases = {}
         self._legal_inputs = ['_ansible_check_mode', '_ansible_no_log', '_ansible_debug', '_ansible_diff', '_ansible_verbosity', '_ansible_selinux_special_fs', '_ansible_module_name', '_ansible_version', '_ansible_syslog_facility']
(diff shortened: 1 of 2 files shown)


lib/ansible/module_utils/basic.py
```diff
@@ -1885,28 +1886,37 @@ def do_cleanup_files(self):
...
+
         self.add_path_info(kwargs)
-        if not 'changed' in kwargs:
-            kwargs['changed'] = False
+
         if 'invocation' not in kwargs:
             kwargs['invocation'] = {'module_args': self.params}
+
+        if 'warnings' not in kwargs and self.warnings:
+            kwargs['warnings'] = self.warnings
```
nitzmahone: Shouldn't this also assert the "shape" of both the `warnings` arg and/or the warnings that were passed in (ie, that it's a list)? Maybe make self.warnings a getter-only property instead so it can't be overwritten accidentally (instead just appended to)?

nitzmahone (dismissed):

lib/ansible/module_utils/basic.py
bcoca: so changed it to use the setter to deal with format

abadger:
I think that @nitzmahone 's idea in slack to use a @property makes more sense for this.  You can test with code like this:
``` python
class Foo(object):
    def __init__(self):
        self._warnings = []
    @property
    def warnings(self):
        return self._warnings

if __name__ == '__main__':
    a = Foo()
    a.warnings.append('Warning one')
    try:
        a.warnings = 'Bad'
    except Exception as e:
        print(e)
    print(a.warnings)
```

bcoca:
that will break modules using `self.warnings = x` ?

Issue #18513, opened 2016-11-16
fetch_url fails with empty exception


EvanK:
##### ISSUE TYPE
 - Bug Report

##### COMPONENT NAME
`fetch_url` helper in module utils

##### ANSIBLE VERSION
```
ansible 2.2.0.0
  config file = /Users/ekaufman/GIT/evolution/wordpress/temp/ansible.cfg
  configured module search path = Default w/o overrides
```

##### CONFIGURATION
```
[ssh_connection]
control_path = %(directory)s/%%h-%%p-%%r
```

##### OS / ENVIRONMENT

Control is OSX 10.11.6, running python 2.7.10

Remote is a vagrant vm of Ubuntu 14.04.5, running python 2.7.6, with pip installs of urllib3, pyopenssl, ndg-httpsclient, pyasn1 (for SNI support)

##### SUMMARY

A `get_url` task is failing with the vague response: `An unknown error occurred: []`

I traced it back to [a caught exception](https://github.com/ansible/ansible/blob/fcca6245d175bc2577e68478f69330020d70f330/lib/ansible/module_utils/urls.py#L1012-L1014) in the `fetch_url` helper call.

The exception itself has no message or args, but I gleaned a class type from it: `<class 'OpenSSL.SSL.Error'>`

##### STEPS TO REPRODUCE

Run the below playbook against a vagrant machine, via: `ansible-playbook --user=vagrant -k -vvv test.yml`

```yml
---
- hosts: "local.example.com"
  gather_facts: yes

  tasks:
  - get_url:
      dest: /tmp/purge-kernels.py
      url: https://raw.githubusercontent.com/EvanK/ubuntu-purge-kernels/master/purge-kernels.py
      force: yes
      mode: '0755'
    sudo: yes
```

##### EXPECTED RESULTS

A file to be fetched from given url and stored to the filesystem

##### ACTUAL RESULTS

```
TASK [get_url] *****************************************************************
task path: /Users/ekaufman/GIT/evolution/wordpress/temp/test.yml:7
Using module file /Library/Python/2.7/site-packages/ansible/modules/core/network/basics/get_url.py
<local.example.com> ESTABLISH CONNECTION FOR USER: vagrant on PORT 22 TO local.example.com
[...]

PR #18533, opened 2016-11-17
Return a traceback for empty exceptions in `get_exception`util

EvanK:
##### ISSUE TYPE
 - Feature Pull Request
 - Bugfix Pull Request

##### COMPONENT NAME
`get_exception` module util helper

##### ANSIBLE VERSION
```
ansible 2.2.0.0
  config file = /Users/ekaufman/.ansible.cfg
```

##### SUMMARY

Related to (but not a complete fix for) #18513, the `get_exception` helper blindly returns any exception.

This would iterate the exception first, testing for any truthy values, and if none are found, instead return a ~~stringified traceback~~ list-formatted traceback wrapped in a generic exception.

### DIFF
--- a/lib/ansible/module_utils/pycompat24.py
+++ b/lib/ansible/module_utils/pycompat24.py
@@ -28,6 +28,7 @@
 #
 
 import sys
+import traceback
 
 def get_exception():
     """Get the current exception.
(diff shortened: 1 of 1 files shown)


EvanK:
Turns out, checking for an empty exception is...not so straightforward :facepalm:

abadger:
I don't think this is correct... I think get_exception() currently returns the triggering exception which is good.  It's the caller's responsibility to decide how to use that exception (to either handle it or produce a good error message from it.)

Doing this here would break idioms like

``` python
try:
   [...]
except OpenSSL.SSL.Error:
   Do something good
```

if OpenSSL.SSL.Error has an empty string representation.

PR #3485, opened 2013-07-09
Let rabbitmq plugin module use nonstandard path

frankshearar:
Sometimes you need to install a Rabbit in a non-standard location. With this commit you can enable/disable plugins for your special Rabbit with

``` yaml
- name: Enable the management plugin
  action: rabbitmq_plugin names=rabbitmq_management path=/custom/path/to/rabbit/ state=enabled
```

PR #18542, opened 2016-11-18
FreeIPA module polymorphic restructuring, Python 3 support, and small…

mkrupcale:
##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
FreeIPA modules

##### ANSIBLE VERSION
```
ansible 2.3.0 (devel 2d2bb626d4) last updated 2016/11/16 17:08:46 (GMT -400)
  lib/ansible/modules/core: (detached HEAD a4cddac368) last updated 2016/11/16 17:08:52 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 19c7d5b31b) last updated 2016/11/16 17:08:56 (GMT -400)
```

##### SUMMARY
This is a change coincident with pull request [#3485](https://github.com/ansible/ansible-modules-extras/pull/3485) in the `ansible/ansible-modules-extras` repository. It adds the ipa module which contains the IPAClient class to ansible.module_utils, as well as some improvements for Python 3 support and fixes (properly URL encoding form data).

* Moved JSON-RPC client IPAClient class to ansible.module_utils.ipa, which is extended by all ipa modules
* IPAClient: Changed to 2-clause BSD license
* IPAClient (lines 37-39): Added some additional imports for use with Python 3
* IPAClient (line 41): Explicitly extend Python base object
* IPAClient (line 57): Properly URL quoted the username/password form data as per https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4.1
* IPAClient (line 62): Data should be bytes or bytearray in Python 3 (still str in Python 2)
* IPAClient (line 65): Print error message, not returned body
* IPAClient (line 70): getheader() is not present in Python 3 version of HTTPMessage; get() is present in both Python 2/3
* IPAClient (line 88): Convert form data to bytes for Python 3 again
* IPAClient (line 91): Print error message, not returned body
* IPAClient (line 96-104): json.loads() requires a string; HTTPResponse.read() returns bytes in Python 3 and str in Python 2, so decode the bytes/string using the HTTPResponse returned charset (default to 'latin-1')
* Add author/copyright notice

### DIFF
--- a/lib/ansible/module_utils/ipa.py
+++ b/lib/ansible/module_utils/ipa.py
@@ -1,0 +1,119 @@
+# -*- coding: utf-8 -*-
+# This code is part of Ansible, but is an independent component.
+# This particular file snippet, and this file snippet only, is BSD licensed.
+# Modules you write using this snippet, which is embedded dynamically by Ansible
+# still belong to the author of the module, and may assign their own license
+# to the complete work.
+#
+# Copyright (c) 2016 Thomas Krahn (@Nosmoht)
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification,
+# are permitted provided that the following conditions are met:
+#
+#    * Redistributions of source code must retain the above copyright
+#      notice, this list of conditions and the following disclaimer.
+#    * Redistributions in binary form must reproduce the above copyright notice,
+#      this list of conditions and the following disclaimer in the documentation
+#      and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+# IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
+# USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+try:
+    import json
+except ImportError:
+    import simplejson as json
+
+from ansible.module_utils.pycompat24 import get_exception
+from ansible.module_utils.urls import fetch_url
+from ansible.module_utils.six.moves.urllib.parse import quote
+from ansible.module_utils.six import PY3
+from ansible.module_utils._text import to_bytes, to_text
+
+class IPAClient(object):
+    def __init__(self, module, host, port, protocol):
+        self.host = host
+        self.port = port
+        self.protocol = protocol
+        self.module = module
+        self.headers = None
+
+    def get_base_url(self):
+        return '%s://%s/ipa' % (self.protocol, self.host)
+
+    def get_json_url(self):
+        return '%s/session/json' % self.get_base_url()
+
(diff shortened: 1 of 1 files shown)


lib/ansible/module_utils/ipa.py
```diff
@@ -0,0 +1,102 @@
...
+        url = '%s/session/login_password' % self.get_base_url()
+        data = 'user=%s&password=%s' % (quote(username, safe=''), quote(password, safe=''))
+        headers = {'referer': self.get_base_url(),
+                   'Content-Type': 'application/x-www-form-urlencoded',
+                   'Accept': 'text/plain'}
+        try:
+            resp, info = fetch_url(module=self.module, url=url, data=b(data), headers=headers)
+            status_code = info['status']
+            if status_code not in [200, 201, 204]:
+                self._fail('login', info['body'])
```
Nosmoht: I made i mistake in the original code here already. 
Line must be

``` python
self._fail('login', info['msg'])
```

Nosmoht:
Thanks @mkrupcale for the optimization done. 

Onle two remarks from my side.
- Please change line 51 like mentioned above, otherwise people want see any error. ( This error was introduced in my original code already. )
- It might make sense to add a note about the author so people know whom to contact in case of issues.

I've tested the change together with https://github.com/ansible/ansible-modules-extras/pull/3485. Both are working well.

``` shell
ansible 2.3.0 (ipa daffdc041a) last updated 2016/11/20 17:44:17 (GMT +200)
  lib/ansible/modules/core: (detached HEAD a4cddac368) last updated 2016/11/18 09:10:52 (GMT +200)
  lib/ansible/modules/extras: (ipa 8da2bcc7c3) last updated 2016/11/20 17:45:56 (GMT +200)
  config file = /home/tkrah/workspace/vagrant/ipa/ansible.cfg
```

shipit

mkrupcale:
No problem @Nosmoht I'm glad I could contribute to your good work. I've pushed the change on line 51 you mentioned, and I also noticed that the same issue was present in the _post_json() method on line 77 and went ahead and updated it as well. Regarding adding a note about the author, I'll allow you to make that change so that you can specify your desired email address, unless you want to provide it here for me to add.

Nosmoht:
@mkrupcale: I would do it the same like i see it in most modules
```
# Author: Thomas Krahn (@Nosmoht)
```
Feel free to add yourself as well and thanks for your contribution.

lib/ansible/module_utils/ipa.py
```diff
@@ -0,0 +1,102 @@
...
+                self._fail(method, info['msg'])
+        except Exception:
+            e = get_exception()
+            self._fail('post %s' % method, str(e))
+
+        if PY3:
+            charset = resp.headers.get_content_charset('latin-1')
+        else:
+            response_charset = resp.headers.getparam('charset')
+            charset = response_charset if response_charset else 'latin-1'
```
abadger: ternary isn't available in python-2.4  If the freeipa module works on python2.4 then this will need to be expanded into a regular if-else.

abadger:
With a quick glance, I didn't see any problems with the usages of str(), .decode(), and .encode() in this patch but in general it's less error prone to use to_text(), to_bytes(), and to_native() from ansible.module_utils._text.  The reason is that str(), .decode(), and .encode() can throw exceptions if you end up with the wrong initial value (trying to run .decode() on a text string for instance).  Remembering what's text and what's bytes gets even harder when dealing with py2+py3 code (where some libraries may return bytes on py2 and text on py3 whereas other libraries will always return the same type).

mkrupcale:
I've pushed a new revision which incorporates the suggestions you made, @abadger, as well as the author/copyright notice for @Nosmoht (also in some of the files in [ansible/ansible-modules-extras#3485](https://github.com/ansible/ansible-modules-extras/pull/3485)). In particular, I've changed the ternary to a if-else structure for potential use in Python 2.4 (not sure how well the rest of the module will work with Python 2.4 though), changed the use of the b() function from ansible.module_utils.six (which calls encode() directly in Python 3) to to_bytes() from ansible.module_utils._text, and similarly used to_text() rather than decode() directly.

resmo:
@abadger this part is under GPL, is this still an issue?

abadger:
Good catch resmo.... @mkrupcale are you able to license this with two clause BSD?  module_utils are licensed that way instead of GPLv3+ so that users can use the module_utils code to write their own custom modules.  If you want to copy text from another module_utils file you can look at the header here and change the copyright lines: https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py

mkrupcale:
@abadger I don't personally have a problem with licensing it as 2-clause BSD, but since this code is mostly taken from the IPA modules in the extras modules repository written by @Nosmoht (which were GPLv3+ licensed), I used GPL here. As long as @Nosmoht is fine with 2-clause BSD, I'm fine with it as well.

Nosmoht:
I'm fine with BSD

gundalow:
If people could take another look at this and give comments I'll merge this and the corresponding module changes

resmo:
LGTM

gundalow (approved):

PR #18546, opened 2016-11-18
fixes issue with setting the terminal length (pager) in vyos

privateip:
`set terminal length 0` actually sets `VYATTA_PAGER=cat`
`set terminal length [some number]` actually sets `stty length [some number]`

### DIFF
--- a/lib/ansible/module_utils/shell.py
+++ b/lib/ansible/module_utils/shell.py
@@ -36,7 +36,8 @@ from ansible.module_utils._text import to_native
 
 ANSI_RE = [
     re.compile(r'(\x1b\[\?1h\x1b=)'),
-    re.compile(r'\x08.')
+    re.compile(r'\x08'),
+    re.compile(r'\x1b[^m]*m')
 ]
 
 def to_list(val):
--- a/lib/ansible/module_utils/vyos.py
+++ b/lib/ansible/module_utils/vyos.py
@@ -27,6 +27,7 @@
 #
 
 import re
+import os
 
 from ansible.module_utils.network import NetworkModule, NetworkError
 from ansible.module_utils.network import register_transport, to_list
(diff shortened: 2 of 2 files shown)


gundalow (approved):
shipit 
With these changes https://github.com/ansible/ansible/pull/18553 now passes

mikemccllstr:
I am trying to use the `vyos_config` module in Ansible 2.2.0.0 against an Ubiquiti ERLite‑3 device running the latest firmware (1.9.1), and I'm running into a timeout as Ansible tries to execute the `show configuration commands` statement. The timeout is caused by the router feeding the command's output through the pager. In debugging the problem, I found this issue in the tracker. However, the code referenced by this issue does not address the problem on Ubiquiti EdgeRouter devices, as they do not support the `set terminal length` command.

This thread describes the same problem when it was encountered with the Rancid project: http://www.gossamer-threads.com/lists/rancid/users/8293#8293

There are two alternate fixes that seem to work through manual testing on my ERLite-3:

1. Pipe the output of specific commands through the `no-more` command, like this: `show configuration commands |no-more`.
2. Explicitly unset the `VYATTA_PAGER` environment variable. This could be done in the `connect()` method, around line 55 of `lib/ansible/module_utils/vyos.py`.

As the second option is durable across all the commands that follow it in the session, it seems preferred.

mikemccllstr:
Another thing I discovered... The `configure` command gets a new environment, so it's necessary to unset the `VYATTA_PAGER` again before calling the `show` command in `load_config()`, near line 81 of same file:

```
         if not self.execute('compare')[0].startswith('No changes'):
+            self.execute(['unset VYATTA_PAGER'])
             diff = self.execute(['show'])[0]
         else:
             diff = None
```

samdoran:
@mikemccllstr I discovered this a while ago doing the same thing (using the VyOS modules on EdgeOS). I'm working on separate modules for EdgeOS to address the differences between the two network operation systems.

mikemccllstr:
Hi @samdoran, I've got a limited installation of them, but I'd be happy to try your modules when they are ready.

samdoran:
@mikemccllstr I just dusted off and got working (in my limited testing) the [EdgeOS modules](https://github.com/samdoran/ansible/tree/edgos-modules/lib/ansible/modules/network/edgeos) I started before The Greate Merge.

Since there is a `module_utils` file and docs fragments, the best way to test these out is by checking out the `edgeos-modules` branch from my [fork](https://github.com/samdoran/ansible/tree/edgos-modules) and running `env-setup` script as described in [Testing Modules](http://docs.ansible.com/ansible/dev_guide/developing_modules.html#testing-modules).

Let me know how things go. I need to do more extensive testing, but the basics seem to work.

PR #18547, opened 2016-11-18
Update async.py

devdkerr:
##### ISSUE TYPE

 - Feature Pull Request
 - New Module Pull Request
 - Bugfix Pull Request
 - Docs Pull Request

##### ANSIBLE VERSION

```

```

##### SUMMARY

```

```

fix to address deprecation of iteritems builtin


mattclay:
The `iteritems` used here is from `ansible.compat.six` and is valid for python 2 and 3.

The proposed change will not work with python 3, as the `dict.iteritems` method has been removed.

PR #18551, opened 2016-11-18
change to ~ instead of $HOME to avoid undefined

### LINKED ISSUE #16032: Ansible will create a literal $HOME directory if the variable is not set

bcoca:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

home

##### ANSIBLE VERSION

```
2.3
```

##### SUMMARY

fixes #16032


stevekuznetsov:
If this is going to be interpreted by a shell, the idiomatic way is to use `"${var:?}"` -- see [the doc](http://www.tldp.org/LDP/abs/html/parameter-substitution.html) or this commit for an example: https://github.com/openshift/origin/pull/9197/commits/0c804382069f5592727a853015729d34a9f5bf8c

abadger:
It's not interpreted by a shell.  We use ```os.path.expanduser(os.path.expandvars(VARIABLE))``` which does its own substitutions of environment variables into the VARIABLE.  I just checked the python stdlib source and it can handle ```{}``` but it won't handle any of the ```:``` syntax.

abadger:
merged to devel and cherry-picked to stable-2.2

PR #18560, opened 2016-11-20
cloudstack: add helpers to distinguish VMs not in VPC

resmo:
##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME
cloudstack

##### ANSIBLE VERSION

```
2.3
```

##### SUMMARY

```

```

VMs in VPC and not in VPC can have an identical name. This means VMs in a VPC must be sorted out if no VPC is given. Due the API limitation, the only way is to check if the network of the VM is in a VPC.

/cc @sbrueseke


resmo:
shipit

PR #18561, opened 2016-11-20
import reduce from six label:python3

kkoralsky:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 
module_utils.facts
plugins.callback.profile_tasks
plugins.strategy.debug

##### ANSIBLE VERSION
```
ansible 2.3.0
```

##### SUMMARY

import reduce from six for python3 comaptibility

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -34,9 +34,8 @@ import pwd
 
 from ansible.module_utils.basic import get_all_subclasses
 from ansible.module_utils.six import PY3, iteritems
-from ansible.module_utils.six.moves import configparser, StringIO
+from ansible.module_utils.six.moves import configparser, StringIO, reduce
 from ansible.module_utils._text import to_native, to_text
-from ansible.compat.six.moves import reduce
 
 try:
     import selinux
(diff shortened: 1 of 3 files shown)


lib/ansible/module_utils/facts.py
```diff
@@ -36,6 +36,7 @@
 from ansible.module_utils.six import PY3, iteritems
 from ansible.module_utils.six.moves import configparser, StringIO
 from ansible.module_utils._text import to_native, to_text
+from ansible.compat.six.moves import reduce
```
abadger: We don't have access to ansible.compat.six in modules (and thus we shouldn't use it in module_utils either.)  Use this instead:
``` python
from ansible.module_utils.six moves import reduce
````

We plan on merging the two together eventually, once we can remove module-side support for python-2.4 (Right now the python-2.4 requirement means that module_utils has an older version of six than in ansible.compat.)

abadger:
One place needs to import six from a different location due to being used module-side.  Other than that this is a good change.  Once it's fixed this can be merged to both devel and stable-2.2 branches.

abadger:
needs_revision

abadger:
Thanks @xkoralsky !  Merged to devel and stable-2.2.  Will be in 2.3.0 and will probably make it into 2.2.1 (if not, then it will make it into 2.2.2)

PR #18567, opened 2016-11-21
Elasticcache primaryEndpoint can be unassigned

DrHofman:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

ec2.py

##### ANSIBLE VERSION

```
ansible 2.2.0.0
  config file = /Users/danielzitzman/projects/ops-tools-infrastructure/ansible.cfg
  configured module search path = ['library/']
```

##### SUMMARY

If the elasticcache cluster dose not have a primary endpoint the inventory script fails

### DIFF
--- a/contrib/inventory/ec2.py
+++ b/contrib/inventory/ec2.py
@@ -1195,7 +1195,7 @@ class Ec2Inventory(object):
             return
 
         # Select the best destination address (PrimaryEndpoint)
-        dest = replication_group['NodeGroups'][0]['PrimaryEndpoint']['Address']
+        dest = replication_group['NodeGroups'][0]['PrimaryEndpoint']
 
         if not dest:
             # Skip clusters we cannot address (e.g. private VPC subnet)


contrib/inventory/ec2.py
```diff
@@ -1195,7 +1195,7 @@ def add_elasticache_replication_group(self, replication_group, region):
             return
 
         # Select the best destination address (PrimaryEndpoint)
-        dest = replication_group['NodeGroups'][0]['PrimaryEndpoint']['Address']
+        dest = replication_group['NodeGroups'][0]['PrimaryEndpoint']
```
s-hertel, line 1198: If dest is a dict now, this will result in failure on line 1205

s-hertel (changes requested):

s-hertel:
this bug will be fixed by #20190

s-hertel:
Fixed in commit [4ca7726](https://github.com/ansible/ansible/commit/4ca7726e7590972e775408ee99a62cd51f9598a6).

Issue #18569, opened 2016-11-21
Target host syslog logging accidentally tied to runtime verbosity


omgjlk:
##### ISSUE TYPE

 - Bug Report

##### COMPONENT NAME

Module execution logging on the target host

##### ANSIBLE VERSION

```
ansible-playbook 2.2.0.0
  config file = /Users/jkeating/src/hoist/ansible.cfg
  configured module search path = Default w/o overrides
```

##### OS / ENVIRONMENT

N/A
##### SUMMARY

Prior to https://github.com/ansible/ansible/commit/a56c0bc27b07356c3639c0ac39805d4e5dd91171 when ansible executed a module on a target host, the invocation of that module was logged to syslog, which was useful for auditing who's doing what to a particular host with Ansible. After that commit, the logging only happens if the execution of ansible was done with verbosity level 3 or above.

##### STEPS TO REPRODUCE

Execute ansible-playbook with less than verbosity 3, no indication on remote host that anything was done.

##### EXPECTED RESULTS

I would expect to see things like:

```
Nov 21 19:06:08 datadog-demo ansible-user[31027]: Invoked with comment=None ssh_key_bits=0 update_password=always non_unique=False force=False ssh_key_type=rsa ssh_key_passphrase=NOT_LOGGING_PARAMETER createhome=True uid=None home=None append=False skeleton=None ssh_key_comment=ansible-generated on datadog-demo group=nodepool system=False state=present shell=None expires=None ssh_key_file=None groups=None move_home=False password=NOT_LOGGING_PARAMETER name=nodepool seuser=None remove=False login_class=None generate_ssh_key=None
```

##### ACTUAL RESULTS

There is no logging of the module execution.

jctanner:
@bcoca ping

jylenhofgfi:
I confirm the patch works also on the 2.2 branch. 
Could you please push it to the 2.2 stable branch ?

Thanks

bcoca:
backported to 2.2, so closing this issue.

PR #18570, opened 2016-11-21
log on target based on nolog, not verbosity

bcoca:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

module logs

##### ANSIBLE VERSION

```
2.2,2.3
```

##### SUMMARY

fies #18569

### DIFF
--- a/lib/ansible/module_utils/basic.py
+++ b/lib/ansible/module_utils/basic.py
@@ -672,6 +672,7 @@ class AnsibleModule(object):
         see library/* for examples
         '''
 
+        self._name = os.path.basename(__file__) #initialize name until we can parse from options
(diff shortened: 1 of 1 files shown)


omgjlk:
This improves the situation, but I still think it would be valuable to log on the target that ansible did _something_, even if all the input data was masked by NOLOG.

bcoca:
no_log is specifically there to turn this logging off.

jylenhofgfi:
Is it possible to backport this fix to the 2.2 branch ? We are affected by this bug.

jylenhofgfi:
I confirm this patch applied correctly to the 2.2 branch too.

PR #18577, opened 2016-11-22
Get windows support to work with Python 3

moriyoshi:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
winrm

##### ANSIBLE VERSION
```
ansible 2.3.0 (moriyoshi/python3-fix 3dc4aa2530) last updated 2016/11/22 11:00:53 (GMT +900)
  lib/ansible/modules/core: (detached HEAD 5c986306be) last updated 2016/11/22 08:12:39 (GMT +900)
  lib/ansible/modules/extras: (detached HEAD 9511de1e3d) last updated 2016/11/22 08:12:39 (GMT +900)
  config file =
```

##### SUMMARY

* This PR contains a number of fixes for the insufficient py3 support of winrm connector.
* As for the hunk containing `format_exc()`, this simply seems to be a bug.

### DIFF
--- a/lib/ansible/plugins/connection/winrm.py
+++ b/lib/ansible/plugins/connection/winrm.py
@@ -208,7 +208,7 @@ class Connection(ConnectionBase):
 
             except Exception as ex:
                 from traceback import format_exc
-                display.warning("FATAL ERROR DURING FILE TRANSFER: %s" % format_exc(ex))
+                display.warning("FATAL ERROR DURING FILE TRANSFER: %s" % format_exc())
                 stdin_push_failed = True
 
             if stdin_push_failed:
@@ -249,8 +249,8 @@ class Connection(ConnectionBase):
 
     def exec_command(self, cmd, in_data=None, sudoable=True):
         super(Connection, self).exec_command(cmd, in_data=in_data, sudoable=sudoable)
-        cmd_parts = shlex.split(to_bytes(cmd), posix=False)
-        cmd_parts = map(to_text, cmd_parts)
+        cmd_parts = shlex.split(to_text(cmd), posix=False)
+        cmd_parts = list(map(to_text, cmd_parts))
         script = None
         cmd_ext = cmd_parts and self._shell._unquote(cmd_parts[0]).lower()[-4:] or ''
         # Support running .ps1 files (via script/raw).
@@ -275,8 +275,8 @@ class Connection(ConnectionBase):
         except Exception:
             traceback.print_exc()
             raise AnsibleConnectionFailure("failed to exec cmd %s" % cmd)
-        result.std_out = to_bytes(result.std_out)
-        result.std_err = to_bytes(result.std_err)
+        result.std_out = result.std_out
+        result.std_err = result.std_err
 
         # parse just stderr from CLIXML output
         if self.is_clixml(result.std_err):
@@ -289,7 +289,7 @@ class Connection(ConnectionBase):
         return (result.status_code, result.std_out, result.std_err)
 
     def is_clixml(self, value):
-        return value.startswith("#< CLIXML")
+        return value.startswith(b"#< CLIXML")
 
     # hacky way to get just stdout- not always sure of doc framing here, so use with care
     def parse_clixml_stream(self, clixml_doc, stream_name='Error'):
@@ -303,11 +303,11 @@ class Connection(ConnectionBase):
         in_size = os.path.getsize(to_bytes(in_path, errors='surrogate_or_strict'))
         offset = 0
         with open(to_bytes(in_path, errors='surrogate_or_strict'), 'rb') as in_file:
-            for out_data in iter((lambda:in_file.read(buffer_size)), ''):
+            for out_data in iter((lambda:in_file.read(buffer_size)), b''):
                 offset += len(out_data)
                 self._display.vvvvv('WINRM PUT "%s" to "%s" (offset=%d size=%d)' % (in_path, out_path, offset, len(out_data)), host=self._winrm_host)
                 # yes, we're double-encoding over the wire in this case- we want to ensure that the data shipped to the end PS pipeline is still b64-encoded
-                b64_data = base64.b64encode(out_data) + '\r\n'
+                b64_data = base64.b64encode(out_data) + b'\r\n'
                 # cough up the data, as well as an indicator if this is the last chunk so winrm_send knows to set the End signal
                 yield b64_data, (in_file.tell() == in_size)
 
@@ -359,7 +359,7 @@ class Connection(ConnectionBase):
         if result.status_code != 0:
             raise AnsibleError(to_native(result.std_err))
 
-        put_output = json.loads(result.std_out)
+        put_output = json.loads(to_text(result.std_out))
         remote_sha1 = put_output.get("sha1")
 
         if not remote_sha1:
--- a/lib/ansible/plugins/shell/powershell.py
+++ b/lib/ansible/plugins/shell/powershell.py
@@ -164,8 +164,8 @@ class ShellModule(object):
         return self._encode_script(script)
 
     def build_module_command(self, env_string, shebang, cmd, arg_path=None, rm_tmp=None):
-        cmd_parts = shlex.split(to_bytes(cmd), posix=False)
-        cmd_parts = map(to_text, cmd_parts)
+        cmd_parts = shlex.split(cmd, posix=False)
+        cmd_parts = list(map(to_text, cmd_parts))
         if shebang and shebang.lower() == '#!powershell':
             if not self._unquote(cmd_parts[0]).lower().endswith('.ps1'):
                 cmd_parts[0] = '"%s.ps1"' % self._unquote(cmd_parts[0])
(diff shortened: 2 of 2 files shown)


lib/ansible/plugins/connection/winrm.py
```diff
@@ -249,8 +249,8 @@ def _reset(self):  # used by win_reboot (and any other action that might need to
 
     def exec_command(self, cmd, in_data=None, sudoable=True):
         super(Connection, self).exec_command(cmd, in_data=in_data, sudoable=sudoable)
-        cmd_parts = shlex.split(to_bytes(cmd), posix=False)
-        cmd_parts = map(to_text, cmd_parts)
+        cmd_parts = shlex.split(to_text(cmd), posix=False)
+        cmd_parts = list(map(to_text, cmd_parts))
```
abadger, line 253: shlex.split needs to use bytes on python2 (more specifically, on python-2.6.  It's a bug that wasn't fixed until python-2.7).  So you can use a conditional on six.PY3 to change this.

In this (and indeed, in many cases), map is better done as a list comprehension.  So:

``` python
cmd_parts = [to_text(p) for p in cmd_parts]
```

lib/ansible/plugins/connection/winrm.py
```diff
@@ -275,8 +275,8 @@ def exec_command(self, cmd, in_data=None, sudoable=True):
         except Exception:
             traceback.print_exc()
             raise AnsibleConnectionFailure("failed to exec cmd %s" % cmd)
-        result.std_out = to_bytes(result.std_out)
-        result.std_err = to_bytes(result.std_err)
+        result.std_out = result.std_out
+        result.std_err = result.std_err
```
abadger, line 279: Connection plugins return bytes for stdout and stderr (the callers are responsible for transforming to text or not).  So this needs to remain with to_bytes().

lib/ansible/plugins/connection/winrm.py
```diff
@@ -359,7 +359,7 @@ def put_file(self, in_path, out_path):
         if result.status_code != 0:
             raise AnsibleError(to_native(result.std_err))
 
-        put_output = json.loads(result.std_out)
+        put_output = json.loads(to_text(result.std_out))
```
abadger, line 362: You might want to specify a different error handler here.  The default handler is surrogate_or_replace.  When non-utf8 byte sequences are encountered, it either uses surrogateescape if available or replace.  This can munge output on python2.  Since we're sending this to json.loads (rather than displaying it to the user) it might be better to use surrogate_or_strict as the error handler here.

lib/ansible/plugins/shell/powershell.py
```diff
@@ -164,8 +164,8 @@ def checksum(self, path, *args, **kwargs):
         return self._encode_script(script)
 
     def build_module_command(self, env_string, shebang, cmd, arg_path=None, rm_tmp=None):
-        cmd_parts = shlex.split(to_bytes(cmd), posix=False)
-        cmd_parts = map(to_text, cmd_parts)
+        cmd_parts = shlex.split(cmd, posix=False)
+        cmd_parts = list(map(to_text, cmd_parts))
```
abadger, line 168: Same shlex comments here as for the connection plugin.

abadger:
Made some comments that need changing.. I'm alerting @nitzmahone to this as well as I'm familiar with the python2+python3 issues but may not be familiar with the widows issues.  I'll also let him decide whether to cherry-pick this to 2.2.x when we do merge it.

nitzmahone:
Yeah, this change is currently breaking the Unicode WinRM connection tests on py2- without digging in, I'm guessing it's either the stdout/stderr conversion or the inbound arg conversion (possibly both).

lib/ansible/plugins/connection/winrm.py
moriyoshi: Hmm, AFAIK py2.6 isn't supported by ansible anymore?  I tested the patch both on 2.7 and 3.5 and it works well.

+1 for a list comprehension.  I'm going to rewrite it so.

lib/ansible/plugins/connection/winrm.py
moriyoshi: In what case is the JSON output encoded in an encoding other than UTF-8 here?  `to_text()` looks fine to me.

lib/ansible/plugins/connection/winrm.py
moriyoshi: result.std_out and result.std_err are already byte at this point s and I don't think these need to be converted again.

moriyoshi:
I'll take a look into the failing tests.

lib/ansible/plugins/connection/winrm.py
abadger: We support python-2.6 and python2.7 (and python3.5 and above) in controller code (which this is).  For modules we support python-2.4 and above unless there's a dependency that requires a newer version of python.

lib/ansible/plugins/connection/winrm.py
abadger: I remember having to add the to_bytes()... it could be that an old version of winrm returned text here.  I do not remember for sure if that was fixed in the lowest winrm version that we support.

Two other notes:
1) This needs to return bytes on python3 as well as python2
2) to_bytes() is a no-op (overhead of the function call and a couple conditionals but no encoding is done) if the string is already a byte string.

lib/ansible/plugins/connection/winrm.py
abadger: If something on the managed machine messes up, this could be in some other encoding.  Also, there's always the chance that the way the client machine handles non-utf-8 bytes will lead to those bytes coming through unescaped.  Basically, the connection plugin is a boundary between Ansible and another system.  So we need to be a little more paranoid in sanitizing our data in this area.

mattclay:
CI failure due to:

```
2017-02-02 00:10:39 TASK [raw with unicode arg and output] *****************************************
2017-02-02 00:10:39 task path: /root/src/github.com/ansible/ansible/test/integration/targets/connection/test_connection.yml:8
2017-02-02 00:10:39 <ec2-54-236-64-151.compute-1.amazonaws.com> ESTABLISH WINRM CONNECTION FOR USER: administrator on PORT 5986 TO ec2-54-236-64-151.compute-1.amazonaws.com
2017-02-02 00:10:39 <ec2-54-236-64-151.compute-1.amazonaws.com> WINRM CONNECT: transport=ssl endpoint=https://ec2-54-236-64-151.compute-1.amazonaws.com:5986/wsman
2017-02-02 00:10:39 <ec2-54-236-64-151.compute-1.amazonaws.com> WINRM OPEN SHELL: 597D0145-710E-4FAE-802F-8B8AB59D8026
2017-02-02 00:10:39 <ec2-54-236-64-151.compute-1.amazonaws.com> WINRM CLOSE SHELL: 597D0145-710E-4FAE-802F-8B8AB59D8026
2017-02-02 00:10:39 An exception occurred during task execution. The full traceback is:
2017-02-02 00:10:39 Traceback (most recent call last):
2017-02-02 00:10:39   File "/root/src/github.com/ansible/ansible/lib/ansible/executor/task_executor.py", line 126, in run
2017-02-02 00:10:39     res = self._execute()
2017-02-02 00:10:39   File "/root/src/github.com/ansible/ansible/lib/ansible/executor/task_executor.py", line 504, in _execute
2017-02-02 00:10:39     result = self._handler.run(task_vars=variables)
2017-02-02 00:10:39   File "/root/src/github.com/ansible/ansible/lib/ansible/plugins/action/raw.py", line 41, in run
2017-02-02 00:10:39     result.update(self._low_level_execute_command(self._task.args.get('_raw_params'), executable=executable))
2017-02-02 00:10:39   File "/root/src/github.com/ansible/ansible/lib/ansible/plugins/action/__init__.py", line 802, in _low_level_execute_command
2017-02-02 00:10:39     rc, stdout, stderr = self._connection.exec_command(cmd, in_data=in_data, sudoable=sudoable)
2017-02-02 00:10:39   File "/root/src/github.com/ansible/ansible/lib/ansible/plugins/connection/winrm.py", line 293, in exec_command
2017-02-02 00:10:39     cmd_parts = shlex.split(to_text(cmd), posix=False)
2017-02-02 00:10:39   File "/usr/lib/python2.7/shlex.py", line 275, in split
2017-02-02 00:10:39     lex = shlex(s, posix=posix)
2017-02-02 00:10:39   File "/usr/lib/python2.7/shlex.py", line 25, in __init__
2017-02-02 00:10:39     instream = StringIO(instream)
2017-02-02 00:10:39 UnicodeEncodeError: 'ascii' codec can't encode characters in position 5-6: ordinal not in range(128)
2017-02-02 00:10:39 
2017-02-02 00:10:39 fatal: [windows-2008-SP2]: FAILED! => {
2017-02-02 00:10:39     "failed": true, 
2017-02-02 00:10:39     "msg": "Unexpected failure during module execution.", 
2017-02-02 00:10:39     "stdout": ""
2017-02-02 00:10:39 }
```

Full logs for the failure here: https://app.shippable.com/runs/58926d1ed3cc6a1000c37edd/5/console

jborean93:
Is this PR still required now that https://github.com/ansible/ansible/pull/24744 has been merged in?

jborean93:
I am going to close this PR, there have been numerous changes to the core engine to update Python compatibility with Windows. There may be a few niche cases but the basics work in 2.4.

PR #18581, opened 2016-11-22
Allow the extension ".ini" in inventory file names

clonedagain:
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
Global change (constants)

##### ANSIBLE VERSION
```
ansible 2.3.0 (inventory-ini-files a1c5cb55ce) last updated 2016/11/22 15:56:11 (GMT +200)
  lib/ansible/modules/core: (detached HEAD 5c986306be) last updated 2016/11/22 15:58:33 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD 9511de1e3d) last updated 2016/11/22 15:58:38 (GMT +200)
  config file = 
```

##### SUMMARY
Since inventory files are "ini-like", allowing the corresponding extension makes it easier for users to edit them with their favourite software.

The extension ".ini" is currently blacklisted, can we allow it ?
If not, can you please explain the reason why ?


bcoca:
Question About This One
=======================

Hi!

The 'ini' extension was left out on purpose as most dynamic inventory scripts expect a ini config file beside the script, if ansible did not ignore it it would constantly issue errors as it is an invalid inventory file.

If your installation does not have this issue, just set the config option in ansible.cfg, but we cannot make this the default for the reasons I gave above.

PR #18590, opened 2016-11-22
Fix the Solaris POSIX acl fix

abadger:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME
lib/ansible/plugins/action/__init__.py

##### ANSIBLE VERSION

```
devel 2.2
```

##### SUMMARY

For setfacl on Solaris we need to specify permissions like r-x.
For chmod, we need to specify them as rx (r-x means to make the file
readable and *not* executable)

This fixes the problem introduced with e326188784ae0b008079aec1fc2be962b1dd9c72


abadger:
Merged to devel and stable-2.2 since the solaris fix was already merged there.

PR #18598, opened 2016-11-23
adds config option to auto add keys when using paramiko

privateip:
* updates paramiko_ssh to auto add keys
* updates constants with new config options

This commit adds a new feature that will allow paramiko to automatically
accept and save a host ssh key.  This feature is controlled by the
`host_key_auto_add` config setting in the paramiko section.  The default
is False to maintain current functionality.  It also includes a new
setting `look_for_keys` with the default to False for maintaining current the
current setting.


gundalow:
I've made a note in Trello that we need to document this

PR #18617, opened 2016-11-24
Fix regression in jinja2 include search path

### LINKED ISSUE #18526: Regression in template {% include %} search path in v2.2
##### ISSUE TYPE
 
Bug Report

##### COMPONENT NAME

template lookup

##### ANSIBLE VERSION

`2.2.0.0`

##### CONFIGURATION

Nothing

##### OS / ENVIRONMENT

Gentoo control machine

##### SUMMARY

When using `{% include %}` in a template, templates in the `templates/` subfolder (relative to playbook) are not found anymore (they were in ansible v2.1).

##### STEPS TO REPRODUCE

I've created a minimal repo to reproduce the issue: https://github.com/savoirfairelinux/ansible-template-test

##### EXPECTED RESULTS

The contents of `foo.txt` and `bar.txt` in the debug message

##### ACTUAL RESULTS

This error:

```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TemplateNotFound: bar.txt
fatal: [localhost]: FAILED! => {"failed": true, "msg": "Unexpected failure during module execution.", "stdout": ""}
```

ghost:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME

template lookup

##### ANSIBLE VERSION
```
ansible-playbook 2.3.0 (devel b606078d69) last updated 2016/11/24 11:46:57 (GMT -400)
  lib/ansible/modules/core: (detached HEAD 351dd9344e) last updated 2016/11/24 11:47:50 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 43bb97bc37) last updated 2016/11/24 11:47:52 (GMT -400)
```

##### SUMMARY

Fixes #18526.

Since commit 3c39bb5, the 'ansible_search_path' variable is used to set
jinja2's search path for {% include %} directives. However, this path is
not the proper one because our templates live in 'templates' subdirs in
our search path.

This is a regression because previously, our include search path would
include the dirname of the currently interpreted file, which worked most
of the time.

Results after PR:

```
TASK [debug] ********************************************************************************************************************************************************************************************************************************
ok: [localhost] => {
    "msg": "Hello, I'm foo!\nHello, I'm bar!\n"
}
```


gundalow:
cc @bcoca Please can you take a look

bcoca:
this is incomplete as not all templates are inside a templates/ dir, https://github.com/bcoca/ansible/commit/edd5518afd3ba88c4ab712b01471f83a510d0db2 solves the issue by including both lists.

bcoca:
pushed #18670 as alternative

ghost:
@bcoca If I understand correctly, your solution *adds* template-suffixed paths to the search path rather than replace the searchpath?

If it's only this, I find the code a bit verbose for what it does. We could simply replace `=` with `+=` in my one liner to achieve a similar effect.

ghost:
To add a bit to my previous comment. My naive understanding of `ansible_search_path` (I can't find documentation for it anywhere. Where is it documented?) is that it's not s search path for templates, but for "root items" (playbooks, group_vars/*, host_vars/*, etc). Which is why I assume that `ansible_search_path` is never supposed to contain a path ending with `templates/`, which would make my `+=` solution adequate and minimal.

bcoca:
no, it is the basic search path for lookup and action plugins (what template is), but for each plugin it should pre-pend additional entries terminating with one of files/, vars/ or templates/ depending on the plugin type (files being default). 

The reason adding `+=` to your code does not work is that order is important, the terminating directories should have higher precedence.

ghost:
Oh, I see. Ok.

In that case, I find the `endswith()` checks to be spurious. I've pushed a new version of this PR which, I think, achieves the same results in a more minimal manner.

ghost:
As a bonus, I even throw in a... complimentary integration test!

ghost:
Oops, sorry, I didn't think about running `code_smell.sh` before pushing (it's my first time playing with this part of the code). Fixing right away.

PR #18625, opened 2016-11-25
WIP: Add Get-SIDFromAccountName to powershell utils

jswetzen:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME
module_utils

##### ANSIBLE VERSION

```
ansible 2.3.0 (detached HEAD 43714f859a) last updated 2016/11/25 14:43:56 (GMT +200)
```

##### SUMMARY

Part of fixing ansible/ansible-modules-extras#2933
Add a function that's used in three windows modules for getting the user SID from an account. Proposed by @jhawkesworth as part of pull request ansible/ansible-modules-extras#3546

```
ansible 2.3.0 (devel e87275ee4b) last updated 2016/11/25 14:24:30 (GMT +200)
```

### DIFF
--- a/lib/ansible/module_utils/powershell.ps1
+++ b/lib/ansible/module_utils/powershell.ps1
@@ -247,3 +247,13 @@ Function Get-PendingRebootStatus
     }
 
 }
+
+Function Get-SIDFromAccountName($accountName)
+{
+    # Helper function to get the SID of an account.
+    # $accountName can be either
+    # - a User Principal Name, on the form user@domain, or
+    # - an account with domain, on the form domain\user
+    $objUser = New-Object System.Security.Principal.NTAccount($accountName)
+    return $objUser.Translate([System.Security.Principal.SecurityIdentifier]).Value
+}


jswetzen:
Actually integrating this with the module code didn't work for my case, so something has to be changed. I have '$' in the username, so it can be hard to handle.

gundalow:
@jswetzen If this PR needs work can you please update the title to start with `WIP:` (work in progress). DO you have an idea on how to make this so it does work with users containing a `$`?

cc @nitzmahone

gundalow:
needs_info

jswetzen:
I haven't started investing yet, but I'll probably find a solution when I continue on Monday.

jswetzen:
Stupid error, I had just missed an `Else` line in the module using the code. However, this is actually part of a 64-line function called `UserSearch` that's duplicated in all three places where this function would be used. I'm not sure that it makes sense to move this part if the rest is duplicated anyway. @jhawkesworth, any thoughts?

jhawkesworth:
Sorry my bad, I just looked at the diff.  Since its 64 lines I think its better to keep it out of powershell.ps1.
At some point we might be able to include different things other than #WANT_POWERSHELL include at which point it would be sensible to make this stuff shared into a different file (or some other mechanism).

nitzmahone:
Yep, that's the "modularize Windows module_utils" line in the [2.3 roadmap](https://github.com/ansible/ansible/blob/devel/docsite/rst/roadmap/ROADMAP_2_3.rst). :) Only thing that keeps me from doing it now is that I'm playing with a Windows version of Ansiballz that would allow those to be actual PS modules instead of getting patched in as scripts. Would hate to move over to a non-Module module system, only to change it again in the next release...

PR #18627, opened 2016-11-25
Make sure include_role inherit variables from parent role

### LINKED ISSUE #18540: include_role does not import variables from outer role
##### ISSUE TYPE
 - Bug Report

##### COMPONENT NAME
include_role

##### ANSIBLE VERSION
```
ansible 2.2.0.0
```

##### CONFIGURATION
Irrelevant

##### OS / ENVIRONMENT
Ubuntu 16.04 and OS X 10.11 (same behavior)

##### SUMMARY

Using `include_role` does not define variables from the outer role in the inner role, which makes it very cumbersome to combine roles in larger ones with common configuration.

##### STEPS TO REPRODUCE

test.yml
```yaml
- hosts: localhost
  tasks:
    - include_role: name=a
```

roles/a/defaults/main.yml
```yaml
outer_variable: I_SHOULD_BE_DEFINED
```

roles/a/tasks/main.yml
```yaml
- include_role: name=b
```

roles/b/tasks/main.yml
```yaml
- debug: msg={{ outer_variable }}
```

##### EXPECTED RESULTS
Correct value of `outer_variable` is printed when running `ansible-playbook -vv test.yml`

##### ACTUAL RESULTS
```
fatal: [localhost]: FAILED! => {"failed": true, "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'outer_variable' is undefined\n\nThe error appears to have been in '/Users/danielkza/ansible/bug_test/roles/b/tasks/main.yml': line 1, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- debug: msg={{ outer_variable }}\n  ^ here\nWe could be wrong, but this one looks like it might be an issue with\nmissing quotes.  Always quote template expression brackets when they\nstart a value. For instance:\n\n    with_items:\n      - {{ foo }}\n\nShould be written as:\n\n    with_items:\n      - \"{{ foo }}\"\n"}
```

danielkza:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
include_role

##### ANSIBLE VERSION
Tested with Ansible 2.2.0.0, 2.2.1.0-0.1.rc1 and 2.3.0.0 (devel @ 43714f859aaf2282f78a2c99b05718c89797cc53)

```
ansible 2.3.0 (fix-include_role-variable-inheritance 519074a128) last updated 2016/11/25 15:45:06 (GMT -200)
  lib/ansible/modules/core: (detached HEAD dedfe2becf) last updated 2016/11/25 15:50:21 (GMT -200)
  lib/ansible/modules/extras: (detached HEAD 43bb97bc37) last updated 2016/11/25 15:50:22 (GMT -200)
  config file =
```

##### SUMMARY
Setting the parent of task blocks generated by include_role after they
have been produced is not sufficient - it means the tasks don't have the
correct dependency chain set. Instead, pass the parent role as part of the
dep_chain *while* compiling the tasks.

Fixes #18540.

### DIFF
--- a/lib/ansible/playbook/role_include.py
+++ b/lib/ansible/playbook/role_include.py
@@ -78,7 +78,13 @@ class IncludeRole(Task):
 
         # compile role with parent roles as dependencies to ensure they inherit
         # variables
-        dep_chain = [] if not self._parent_role else [self._parent_role]
+        if not self._parent_role:
+            dep_chain = []
+        else:
+            dep_chain = list(self._parent_role._parents)
+            dep_chain.extend(self._parent_role.get_all_dependencies())
+            dep_chain.append(self._parent_role)
+
         blocks = actual_role.compile(play=myplay, dep_chain=dep_chain)
         for b in blocks:
             b._parent = self
--- a/lib/ansible/playbook/role_include.py
+++ b/lib/ansible/playbook/role_include.py
@@ -76,10 +76,10 @@ class IncludeRole(Task):
         actual_role = Role.load(ri, myplay, parent_role=self._parent_role, from_files=self._from_files)
         actual_role._metadata.allow_duplicates = self.allow_duplicates
 
-        # compile role
-        blocks = actual_role.compile(play=myplay)
-
-        # set parent to ensure proper inheritance
+        # compile role with parent roles as dependencies to ensure they inherit
+        # variables
+        dep_chain = [] if not self._parent_role else [self._parent_role]
+        blocks = actual_role.compile(play=myplay, dep_chain=dep_chain)
         for b in blocks:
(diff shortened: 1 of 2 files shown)


danielkza:
I wrote some tests that caught that only variables from one level up were being inherited. Fixed it now.

danielkza:
@bcoca Thank you! Can you please cherry-pick this to 2.2 as well?

kimamula:
An included role still does not inherit variables defined in `defaults` of the parent.
Is this an expected behavior?

I am using Ansible 2.2.1.0.

PR #18632, opened 2016-11-26
expose timeout param to socket

privateip:
The timeout param was exposed to the socket connection but was not
enforced for commands.  This update will now cause a command to timeout
based on the module parameter.


gundalow (approved):
shipit

PR #18633, opened 2016-11-26
pass module timeout value to fetch_url in eapi connection

privateip:
This change causes the eapi connection to honor the module timeout
value when calling fetch_url


gundalow (approved):
shipit

PR #18634, opened 2016-11-26
fixes timeout param in netconf provider for junos

privateip:
This change will now cause the netconf provider to honor the module
timeout value when making calls to pyez.


gundalow (approved):

gundalow (approved):
shipit

PR #18635, opened 2016-11-26
updates the network shared docs to include the timeout parameter

privateip:
The network shared doc fragment was missing the timeout parameter from
the docstring.  This commit adds the timeout parameter back to the
doc fragment


lib/ansible/utils/module_docs_fragments/eos.py
```diff
@@ -52,6 +52,13 @@ class ModuleDocFragment(object):
         value of environment variable C(ANSIBLE_NET_PASSWORD) will be used instead.
     required: false
     default: null
+  timeout:
+    description:
+      - Specifies the timeout for communicating with the network device
+        for either connecting or sending commands.  If the timeout is
+        exceeded before the operation is completed, the module will error
+    require: false
```
gundalow: `version_added: 2.3`

lib/ansible/utils/module_docs_fragments/eos.py
```diff
@@ -52,6 +52,13 @@ class ModuleDocFragment(object):
         value of environment variable C(ANSIBLE_NET_PASSWORD) will be used instead.
     required: false
     default: null
+  timeout:
+    description:
+      - Specifies the timeout for communicating with the network device
```
gundalow: Specifies the timeout (in seconds) for...

gundalow (changes requested):
needs_revision

lib/ansible/utils/module_docs_fragments/eos.py
privateip: This was actually added in 2.2 it was just missing the docstring

gundalow (approved):
shipit

PR #18651, opened 2016-11-28
fixes exception raised when extracting timeout value for eapi url

privateip:


gundalow (approved):

gundalow:
Tests now get further, so this fix is correct

PR #18660, opened 2016-11-29
Fix for AnsiballZ when the remote clock is behind

### LINKED ISSUE #18640: Ziploader (ansiballz) fails to work if time predates 1980
##### ISSUE TYPE

 - Bug Report

##### COMPONENT NAME

ziploader

##### ANSIBLE VERSION

v2.1 and newer

##### CONFIGURATION

Raspberry Pi 3 running aarch64

##### OS / ENVIRONMENT

CentOS 7

##### SUMMARY

Raspberry Pi's don't have a hardware clock and depends heavily on NTP. Which means that systems won't have a proper date set (they go to epoch by default) until NTP is installed, configured and working.

Now there is a silly problem with zip that it fails to work if the time predates 1980.

```
[dag@moria berries]$ ansible-playbook config.yml -l berry4 -u root

PLAY [berries] *****************************************************************

An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ValueError: ZIP does not support timestamps before 1980
fatal: [berry4]: FAILED! => {"changed": false, "failed": true, "module_stderr": "Traceback (most recent call last):\n  File \"<stdin>\", line 128, in <module>\n  File \"/usr/lib64/python2.7/zipfile.py\", line 1202, in writestr\n    date_time=time.localtime(time.time())[:6])\n  File \"/usr/lib64/python2.7/zipfile.py\", line 305, in __init__\n    raise ValueError('ZIP does not support timestamps before 1980')\nValueError: ZIP does not support timestamps before 1980\n", "module_stdout": "", "msg": "MODULE FAILURE"}
```

Not sure if we can easily do anything about this, or even if this is something we want to fix, as workarounds are possible using raw/command. It's still a burden that ziploader has this dependency to the realtime clock especially since there's a lot to do to set up NTP (network, packages, configuration)...

abadger:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME
Ansiballz  (in executor/module_common.py )

##### ANSIBLE VERSION

```
devel, 2.2, 2.1
```

##### SUMMARY

Some machines have system clocks which can fall behind (for instance,
a host without a CMOS battery like Raspberry Pi).  When managing those
machines we have to workaround the fact that the zip format does not
handle file timestamps before 1980.  The workaround is to substitute in
the timestamp from the controller instead of from the managed machine.

Fixes #18640

### DIFF
--- a/lib/ansible/executor/module_common.py
+++ b/lib/ansible/executor/module_common.py
@@ -22,6 +22,7 @@ __metaclass__ = type
 
 import ast
 import base64
+import datetime
 import imp
 import json
 import os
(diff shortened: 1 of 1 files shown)


abadger:
Merged to devel and stable-2.2.  This fix will end up in 2.2.1 or 2.2.2 depending on whether we need a 2.2.1rc2.

dagwieers:
Nice ! I wasn't expecting a solution, and certainly not at this speed :-)

PR #18661, opened 2016-11-29
Support external templates directory for role init

realloc:
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
ansible-galaxy cli

##### ANSIBLE VERSION

```
ansible 2.3.0 (feature/role_templates 520a12f6f0) last updated 2016/11/28 19:31:06 (GMT +900)
```

##### SUMMARY
As a role developer I want to use custom role templates for each of my projects
and still use `ansible-galaxy init` for initial role skeleton generation. The
additional `-t` argument sets the path to search for such templates instead of
standard one. It may be useful in corporate environment where each team has it's
own role templates.

Usage example: 
```
ansible-galaxy init  -p ~/Project1/roles -t ~/Project1/role_templates role1
```

### DIFF
--- a/docs/man/man1/ansible-galaxy.1.asciidoc.in
+++ b/docs/man/man1/ansible-galaxy.1.asciidoc.in
@@ -121,6 +121,11 @@ Force overwriting an existing role.
 The path in which the skeleton role will be created.The default is the current 
 working directory.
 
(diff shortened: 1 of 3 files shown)


abadger:
An equivalent feature named role skeletons was added via this PR: https://github.com/ansible/ansible/pull/17079  The feature is in the current stable release (ansible-2.3.x).

PR #18662, opened 2016-11-29
Allow variables to be passed in to the template lookup plugin

amenonsen:
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
plugins/lookup/template.py

##### ANSIBLE VERSION
```
devel
```

##### SUMMARY

This allows template evaluation to be controlled with inline variable
definitions that augment whatever variables are defined already:

    {{ lookup('template','x.j2',set_variables=dict(x=42)) }}

### DIFF
--- a/docs/docsite/rst/playbooks_lookups.rst
+++ b/docs/docsite/rst/playbooks_lookups.rst
@@ -556,6 +556,9 @@ Here are some examples::
 
          - debug: msg="{{ lookup('template', './some_template.j2') }} is a value from evaluation of this template"
 
+         # Since 2.4, you can pass in variables during evaluation
+         - debug: msg="{{ lookup('template', './some_template.j2', template_vars=dict(x=42)) }} is evaluated with x=42"
+
          - name: loading a json file from a template as a string
            debug: msg="{{ lookup('template', './some_json.json.j2', convert_data=False) }} is a value from evaluation of this template"
 
--- a/lib/ansible/plugins/lookup/template.py
+++ b/lib/ansible/plugins/lookup/template.py
@@ -36,6 +36,7 @@ class LookupModule(LookupBase):
     def run(self, terms, variables, **kwargs):
 
         convert_data_p = kwargs.get('convert_data', True)
+        lookup_template_vars = kwargs.get('template_vars', {})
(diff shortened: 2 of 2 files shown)


calfonso:
Hi!

We understand you want to get around declaring the vars in the task with this PR; however, we don't understand why. If you're interested in talking through it and convincing us, we can re-open this PR and review again.

amenonsen:
I've pushed an updated version of this PR now. When I originally submitted it, ``templar.set_available_variables()`` did not exist, now there's no need for me to change ``lib/template/__init__.py`` any more.

As explained in the commit message and in a lengthy discussion on IRC, this allows a template to be evaluated different ways within the same task, which is not possible by setting ``vars:`` on a task (in particular, if you want to generate the strings ``"x:1"`` and ``"x:2"`` from the same template ``"x:{{y}}"``, then you can do it by passing in different values of ``y`` to the lookup, but with ``vars:`` you would need to have two different vars ``y1:1`` and ``y2:2`` and two different templates, ``"x:{{y1}}"`` and ``"x:{{y2}}"``.)

I've left the name of the argument as ``vars=`` here, because I can't think of anything better. Maybe ``extra_vars`` could work, but anything else seems like a gratuitous change, e.g., ``variables=``. Suggestions?

docs/docsite/rst/playbooks_lookups.rst
```diff
@@ -556,6 +556,9 @@ Here are some examples::
 
          - debug: msg="{{ lookup('template', './some_template.j2') }} is a value from evaluation of this template"
 
+         # passing in variables during evaluation
```
bcoca: add 'since version 2.4'

lib/ansible/plugins/lookup/template.py
```diff
@@ -36,6 +36,7 @@ class LookupModule(LookupBase):
     def run(self, terms, variables, **kwargs):
 
         convert_data_p = kwargs.get('convert_data', True)
+        extra_vars = kwargs.get('vars', {})
```
bcoca: from core meeting, the vote is in and we would like the new option to be named `template_vars`

PR #18671, opened 2016-11-29
Fix overwrite parameter in module_utils.junos

ikelos:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

ansible.module_utils.junos

##### ANSIBLE VERSION

```
ansible 2.2.0.0
  config file = 
```

##### SUMMARY

The overwrite parameter is forcibly set to false, meaning a module
passing that parameter will have no effect.  The overwrite facility
is necessary to ensure that conflicting options can be written the
configuration (which, in replace mode, they cannot).

This change ensures that if overwrite is set, it will not be changed
to False in the logic.

At the moment this will have no effect, because it requires a change
to junos_config in order to take effect.

### DIFF
--- a/lib/ansible/module_utils/junos.py
+++ b/lib/ansible/module_utils/junos.py
@@ -184,8 +184,8 @@ class Netconf(object):
             merge = False
             overwrite = False
         elif overwrite:
-            merge = True
-            overwrite = False
+            merge = False
+            overwrite = True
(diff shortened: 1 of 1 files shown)


privateip:
Thanks for the fix... merged into devel and stable-2.2

PR #18677, opened 2016-11-30
with_sequence: pass AnsibleError through

benallard:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
with_sequence

##### ANSIBLE VERSION
```
ansible 1.7.2
```

##### SUMMARY

The parsing methods try as hard as possible to generate meaningful error messages that are all ignored and immediately overwritten by a new AnsibleError instance. Better use the original one instead.


abadger:
This looks good to me.  Shippable errors appear to be timeouts that are occuring on different hosts every run.

Merged to devel but not to stable-2.2 in case of unintended side-effects.

PR #18681, opened 2016-11-30
Handle tuples in `hash_params` when creating a role

### LINKED ISSUE #18680: Exception when looping include_role using with_subelements
##### ISSUE TYPE

 - Bug Report
 
##### COMPONENT NAME
include_role module

##### ANSIBLE VERSION

```
ansible 2.2.0.0
```

##### CONFIGURATION
N/A

##### OS / ENVIRONMENT
N/A

##### SUMMARY
`with_subelements` creates an `item` variable which is a tuple, where the first item (at least) is a dict.

Parameters passed through the `include_role` module are hashed when creating the role. The `hash_params` function only recurses over dicts and lists, which means that the `item` tuple is inserted into the set of params unmodified, causing a `TypeError: unhashable type: 'dict'` to be thrown.

##### STEPS TO REPRODUCE

```
- hosts: localhost
  tasks:
    - include_role: name=test
      with_subelements:
        - - name: foo
            values: ['bar']
        - values
```

##### EXPECTED RESULTS

The `test` role is included once and is passed an `item` tuple equal to `({'name': 'foo'}, 'bar')`

##### ACTUAL RESULTS

The play fails with an unexpected exception.

```
TASK [include_role] ************************************************************
task path: /Users/ahaines/work/iot/infrastructure/test.yml:3
ERROR! Unexpected Exception: unhashable type: 'dict'
the full traceback was:

  File "/usr/local/bin/ansible-playbook", line 103, in <module>
    exit_code = cli.run()
  File "/usr/local/Cellar/ansible/2.2.0.0_1/libexec/lib/python2.7/site-packages/ansible/cli/playbook.py", line 159, in run
    results = pbex.run()
  File "/usr/local/Cellar/ansible/2.2.0.0_1/libexec/lib/python2.7/site-packages/ansible/executor/playbook_executor.py", line 154, in run
  File "/usr/local/Cellar/ansible/2.2.0.0_1/libexec/lib/python2.7/site-packages/ansible/executor/task_queue_manager.py", line 282, in run
  File "/usr/local/Cellar/ansible/2.2.0.0_1/libexec/lib/python2.7/site-packages/ansible/plugins/strategy/linear.py", line 304, in run
    all_role_blocks.extend(new_ir.get_block_list(play=iterator._play, variable_manager=self._variable_manager, loader=self._loader))
  File "/usr/local/Cellar/ansible/2.2.0.0_1/libexec/lib/python2.7/site-packages/ansible/playbook/role_include.py", line 76, in get_block_list
    actual_role = Role.load(ri, myplay, parent_role=self._parent_role, from_files=self._from_files)
  File "/usr/local/Cellar/ansible/2.2.0.0_1/libexec/lib/python2.7/site-packages/ansible/playbook/role/__init__.py", line 118, in load
    hashed_params = hash_params(params)
  File "/usr/local/Cellar/ansible/2.2.0.0_1/libexec/lib/python2.7/site-packages/ansible/playbook/role/__init__.py", line 54, in hash_params
    s.update((k, hash_params(v)))
  File "/usr/local/Cellar/ansible/2.2.0.0_1/libexec/lib/python2.7/site-packages/ansible/playbook/role/__init__.py", line 61, in hash_params
    s.update((k, v))
TypeError: unhashable type: 'dict'
```

haines:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME
include_role module

##### ANSIBLE VERSION

```
ansible 2.2.0.0
```

##### SUMMARY
When using `with_subelements` with `include_role`, the `item` variable is a tuple containing at least one dict. This cannot be inserted into the role's set of hashed params without being passed through `hash_params` itself.

```
- hosts: localhost
  tasks:
    - include_role: name=test
      with_subelements:
        - - name: foo
            values: ['bar']
        - values
```

Before
```
TASK [include_role] ************************************************************
ERROR! Unexpected Exception: unhashable type: 'dict'
```

After
```
TASK [include_role] ************************************************************

TASK [test : debug] ************************************************************
ok: [localhost] => {
    "item": [
        {
            "name": "foo"
        }, 
        "bar"
    ]
}
```

Fixes #18680

### DIFF
--- a/lib/ansible/playbook/role/__init__.py
+++ b/lib/ansible/playbook/role/__init__.py
@@ -52,7 +52,7 @@ def hash_params(params):
         for k,v in iteritems(params):
             if isinstance(v, dict):
                 s.update((k, hash_params(v)))
-            elif isinstance(v, list):
+            elif isinstance(v, (list, set, tuple)):
                 things = []
                 for item in v:
                     things.append(hash_params(item))


lib/ansible/playbook/role/__init__.py
```diff
@@ -52,7 +52,7 @@ def hash_params(params):
         for k,v in iteritems(params):
             if isinstance(v, dict):
                 s.update((k, hash_params(v)))
-            elif isinstance(v, list):
+            elif isinstance(v, list) or isinstance(v, tuple):
```
abadger: This can be written slightly better as 
``` python
elif isinstance(v, (list, tuple)):
```

abadger:
@bcoca this looks right to me (aside from the minor change I proposed which doesn't change the logic any.)  Does this look good to you as well?

lib/ansible/playbook/role/__init__.py
haines: Nice, thanks

bcoca:
looks good, do we need sets also?

haines:
Yeah, it would probably make sense to add sets, will do.

abadger:
Note: If we don't have a lot of control over what type of data structures we might see here (I thought it was standard stuff that we were creating but I didn't read beyond the PR) then it's better to use the collections, abstract base classes: https://docs.python.org/2/library/collections.html#collections-abstract-base-classes

So the isinstance check would become:
``` python
from collections import Sequence, Set
[...]
if isinstance(v, (Sequence, Set)):
```

bcoca:
~~actually, i think that we want a 'hashable' test instead~~, need container test to see if all elements are hashable, probably need to rethink the whole thing

abadger:
heh, I htink we're going down a rabbit hole here... Let me think on this for 30 minutes.

abadger:
@haines Okay, see if https://github.com/ansible/ansible/pull/18701 works for you.  It worked in my testing.  Sorry I thought this would be simple at first but then kept finding cornercases that expanded the scope.

haines:
@abadger #18701 works for me! Thanks for looking into it.

PR #18685, opened 2016-11-30
tell subprocess.popen to use universal_newlines

tedder:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
Ansible CLI vaultpass handling

##### ANSIBLE VERSION

```
ansible 2.3.0
  config file = /opt/app/ansible.cfg
  configured module search path = ['./library']
```

##### SUMMARY
Fixes ##18684. Very straightforward unless I missed something- universal_newlines ensures that the CLI won't barf on byte/string issues.

### DIFF
--- a/lib/ansible/cli/__init__.py
+++ b/lib/ansible/cli/__init__.py
@@ -624,7 +624,7 @@ class CLI(with_metaclass(ABCMeta, object)):
         if loader.is_executable(this_path):
             try:
                 # STDERR not captured to make it easier for users to prompt for input in their scripts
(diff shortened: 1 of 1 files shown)


abadger:
If https://github.com/ansible/ansible/pull/18699 works as well, then we should use that as it is a more complete fix to make sure that we only use bytes to operate on stdout and then convert to text when we store it into vault_pass.

abadger:
#18699 has been merged so closing this one.

PR #18689, opened 2016-11-30
set_artifact action plugin with tests

chrismeyersfsu:
##### ISSUE TYPE

 - New Module Pull Request

##### COMPONENT NAME

`set_artifact`

ansible/ansible-modules-core#5776

### DIFF
--- a/test/integration/targets/set_artifact/tasks/main.yml
+++ b/test/integration/targets/set_artifact/tasks/main.yml
@@ -1,3 +1,20 @@
+# (c) 2016, Chris Meyers <cmeyers@ansible.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+#
 ---
 - name: Init artifact structure
   set_artifact:
--- a/test/integration/targets/set_artifact/tasks/main.yml
+++ b/test/integration/targets/set_artifact/tasks/main.yml
@@ -1,0 +1,101 @@
+---
+- name: Init artifact structure
+  set_artifact:
+    data:
+      a: 7
+  register: result_init
+
+- name: Verify module returns artifact_data
+  assert:
+    that:
+      - result_init.artifact_data is defined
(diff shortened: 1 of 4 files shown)


abadger:
The new tests are failing.
needs_revision

test/integration/targets/set_artifact/tasks/main.yml
```diff
@@ -0,0 +1,118 @@
...
+  set_artifact:
+    data:
+      b: 8
+  register: result_merge
+
+- name: Ensure set_artifacts does not set facts on hosts other than the first
+  assert:
+    that:
+      - artifact_data is undefined
+  when: "inventory_hostname != ansible_play_hosts[0]"
```
bcoca, line 49: This test is irrelevant as the play only runs against 'testhost'

test/integration/targets/set_artifact/tasks/main.yml
```diff
@@ -0,0 +1,118 @@
...
+
+- name: Verify first-host artifact_data fact set
+  assert:
+    that:
+      - artifact_data is defined
+      - artifact_data.a == 7
+      - artifact_data.b == 8
+      - artifact_data|length == 2
+  run_once: true
```
bcoca, line 58: run_once does not matter ... as play only uses single host

amenonsen:
An interesting-looking new module PR… with *not even one line* of explanation about what it does?

mattclay:
For tests that have special requirements, such as needing custom inventory, use a script based test instead of a role based test.

Take a look at the [unicode test](https://github.com/ansible/ansible/blob/devel/test/integration/targets/unicode/runme.sh) for an example.

chrismeyersfsu:
@amenonsen since this is a action plugin there is a corresponding ansible module that contains the docs, see https://github.com/ansible/ansible-modules-core/pull/5776

@mattclay It's been a while since I've delved into the Ansible test structure. Previously, adding the role invocation to `non_destructive.yml` and forming your tests around the respective `inventory` was the way to go. Is `non_destructive.yml` not a thing anymore? Thanks for the `unicode test` reference.

bcoca:
@chrismeyersfsu it is but non_destructive does not test 'multi host issues', which your tests seem to require.

chrismeyersfsu:
yep, your right. I'll use the custom inventory methodology.

bcoca:
-1, this seems like an alternate implementation of a fact/non_fact cache with aggregation of dicts, which can also be achieved via template/vars_fiiles|include_vars and combine filter.

amenonsen:
As @bcoca says, I've found this functionality useful enough to implement it using include_vars (though I used copy with ``content: {{ variables|to_json }}`` instead of template; minor difference). So I like the idea, but I don't much like ``set_artifact``. It just feels a bit arbitrary and confusing. I know what an artifact is, but I don't know why I would be setting one, and the module documentation doesn't really help me. The term brings to mind build artifacts (e.g., ``*.pyc`` files) rather than anything else.

I wouldn't mind a ``set_variable`` that made it convenient to persistently set variables in the local inventory (and maybe even the remote fact cache), so maybe it's a matter of repackaging this functionality in a more general way.

PR #18690, opened 2016-11-30
Added tests for sefcontext module

dagwieers:
##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

sefcontext

##### ANSIBLE VERSION

v2.2

##### SUMMARY

Added tests for sefcontext module

This is my first set of integration tests, bear with me.

### DIFF
--- a/test/integration/non_destructive.yml
+++ b/test/integration/non_destructive.yml
@@ -1,5 +1,5 @@
 - hosts: testhost
-  gather_facts: True
+  gather_facts: yes
   roles: 
     - { role: ping, tags: test_ping }
     - { role: special_vars, tags: test_special_vars }
@@ -34,3 +34,4 @@
     - { role: loops, tags: test_loops }
     - { role: mount, tags: [test_mount, needs_root, needs_privileged]}
     - { role: include_vars, tags: test_include_vars }
+    - { role: sefcontext, tags: [test_sefcontext, needs_root]}
--- a/test/integration/targets/sefcontext/aliases
+++ b/test/integration/targets/sefcontext/aliases
@@ -1,0 +1,2 @@
+needs/root
+posix/ci/group2
--- a/test/integration/targets/sefcontext/tasks/main.yml
+++ b/test/integration/targets/sefcontext/tasks/main.yml
@@ -1,0 +1,28 @@
+# (c) 2016, Dag Wieers <dag@wieers.com>
+
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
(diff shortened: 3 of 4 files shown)


abadger:
shippable tests were fialing due to timeouts.  I've triggered a rebuild to the tests.

test/integration/non_destructive.yml
```diff
@@ -34,3 +34,4 @@
     - { role: loops, tags: test_loops }
     - { role: mount, tags: [test_mount, needs_root, needs_privileged]}
     - { role: include_vars, tags: test_include_vars }
+    - { role: sefcontext, tags: [test_sefcontext, needs root]}
```
mattclay: This should be `needs_root` with an underscore, not a space.

mattclay:
@dagwieers You should run [reorganize-tests.sh](https://github.com/ansible/ansible/blob/devel/test/runner/reorganize-tests.sh) to generate the necessary `aliases` file for the new test so it can be included in the PR.

This is a temporary measure to keep the old test playbooks in sync with the new `aliases` files used by `ansible-test`. Once we've removed the old playbooks we'll just update the `aliases` files manually.

dagwieers:
Is this normal:

```
[dag@moria ansible.dag]$ ./test/runner/reorganize-tests.sh
./test/runner/reorganize-tests.sh: line 16: lib/ansible/modules/core/test/utils/shippable/sanity-skip-python24.txt: No such file or directory
```

dagwieers:
Alright, it isn't working. It seems the test infrastructure does not have SELinux enabled.

2016-12-01 21:34:13 TASK [sefcontext : debug] ******************************************************
2016-12-01 21:34:13 ok: [testhost] => {
2016-12-01 21:34:13     "msg": "SELinux is disabled"
2016-12-01 21:34:13 }

So the sefcontext tests are useless :-/
Please merge :-)

test/integration/targets/sefcontext/aliases
```diff
@@ -0,0 +1,3 @@
+needs\/privileged
+needs\/root
```
mattclay: These aliases aren't right. They should be `needs/root` and `needs/privileged`.

mattclay: Also, have you verified that `needs/privileged` is necessary? You'd have to run the tests in docker on a host with selinux support to see if docker privileged mode is necessary. Otherwise, leave out `needs/privileged` until we know it's needed.

dagwieers:
I tried both.

The incorrected aliases were created by that script, which I had to modify to run.

mattclay:
@dagwieers What version of bash did you use to run the script? On what OS? I ran it locally and it produced the correct results.

mattclay:
@dagwieers I've verified that selinux will always be reported as disabled inside a docker container, even when it is supported by the host. So we won't be able to run these as part of CI until we have a Linux VM to test with. 

Go ahead and remove the `needs\/privileged` alias and correct the other to be `needs/root`.

dagwieers:
Done.
shipit

PR #18698, opened 2016-12-01
ssh-keyscan can fail to find keys for a host.

### LINKED ISSUE #18676: Git module fails to add hostkey to known_hosts and doesn't notice it
##### ISSUE TYPE
 - Bug Report

##### COMPONENT NAME

ansible.module_utils.known_hosts

##### ANSIBLE VERSION
```
ansible 2.1.3.0
  config file = 
```

##### OS / ENVIRONMENT

Ubuntu 16.04 

##### SUMMARY

Function add_host_key which gets called from  add_git_host_key, which gets called from the git module when accept_hostkey=yes, does not check that the output from ssh-keyscan is not empty. But ssh-keyscan has a timeout (5 seconds by default), and if the server could not answer in a timely manner, ssh-keyscan prints nothing to stdout and returns with return code 0. Ansible assumes that this operation is successful and proceeds to call git, which then hangs on a host key confirmation dialog.

The demonstration is as follows (I used -T1 to reproduce more easily, with default timeout of 5 seconds this also happens)
```
~$ ssh-keyscan -t rsa -T 1  bitbucket.org // this is succesful
# bitbucket.org:22 SSH-2.0-conker_1.0.263-aeae7ee app-134
bitbucket.org ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAubiN81eDcafrgMeLzaFPsw2kNvEcqTKl/VqLat/MaB33pZy0y3rJZtnqwR2qOOvbwKZYKiEO1O6VqNEBxKvJJelCq0dTXWT5pbO2gDXC6h6QDXCaHo6pOHGPUy+YBaGQRGuSusMEASYiWunYN0vCAI8QaXnWMXNMdFP3jHAJH0eDsoiGnLPBlBp4TNm6rYI74nMzgz3B9IikW4WVK+dc8KZJZWYjAuORU3jc1c/NPskD2ASinf8v3xnfXeukU0sJ5N6m5E8VLjObPEO+mN2t/FZTMZLiFqPWc/ALSqnMnnhwrNi2rbfg/rd/IpL8Le3pSBne8+seeFVBoGqzHM9yXw==
~$ echo $? // ret code zero
0
~$ ssh-keyscan -t rsa -T 1  bitbucket.org // this one failed
~$ echo $? // ret code still zero
0
```

##### STEPS TO REPRODUCE

Execute a git task with a slow git server with accept_hostkey=yes

##### EXPECTED RESULTS

Fail with a meaningful error specifying that the keyscan was unsuccessful. Also, a way to increase the default timeout would be nice.

##### ACTUAL RESULTS

The module hangs indefinitely

related issue: https://github.com/ansible/ansible-modules-core/issues/1343

abadger:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME
lib/ansible/module_utils/known_hosts.py

##### ANSIBLE VERSION

```
devel, 2.2
```

##### SUMMARY

When it does, we need to fail otherwise other code will fail later.
This bug can cause the git module to hang indefinitely (because it's waiting for user input to add the host key since we did not properly set it here).

Fixes #18676

### DIFF
--- a/lib/ansible/module_utils/known_hosts.py
+++ b/lib/ansible/module_utils/known_hosts.py
@@ -187,6 +187,9 @@ def add_host_key(module, fqdn, key_type="rsa", create_dir=False):
     this_cmd = "%s -t %s %s" % (keyscan_cmd, key_type, fqdn)
 
     rc, out, err = module.run_command(this_cmd)
+    # ssh-keyscan gives a 0 exit code and prints nothins on timeout
+    if rc != 0 or not out:
+        module.fail_json(msg='failed to get the hostkey for %s' % fqdn)
     module.append_to_file(user_host_file, out)
 
     return rc, out, err


tannerjcghtestuser4:
shipit

jctanner:
shipit

ngo:
Possibly a better fix would be to check if the key is in known_hosts after keyscan, just to be 300% sure. But this works well enough too.

abadger:
I've merged this fix into devel and cherry-picked to temp-post-2.2.1-staging for potential inclusion in 2.2.2... We're in late release candidate stage for 2.2.1 so I'm trying to slow down the rate of change there a little bit.

Did not merge to stable-2.1 as it's not a regression there and we'll probably only do major security releases form that branch now.

abadger:
Talked about it on our slack channel and decided that we could go ahead and merge this for 2.2.1 as the change only affects the git module.  It's now in the stable-2.2 branch

PR #18734, opened 2016-12-03
Fix to correctly add hostkey for non-standard git ports

nev3rm0re:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
ansible.module_utils.basic.known_hosts

##### ANSIBLE VERSION

```
ansible 2.3.0 (devel ea5c3df043) last updated 2016/12/02 23:56:21 (GMT +300)
  lib/ansible/modules/core: (detached HEAD 6890003c4f) last updated 2016/12/02 23:59:15 (GMT +300)
  lib/ansible/modules/extras: (detached HEAD 20ea46642b) last updated 2016/12/02 23:59:21 (GMT +300)
```

##### SUMMARY

This change fixes `git` module's `accept_hostkey` behavior for non-standard ports. 

Currently, when git url uses non-standard port (my repo sits on 7999, for example), incorrect hostkey is added to `~/.ssh/known_hosts`. 

After the fix, non-standard ports will be explicitly passed as an argument to `ssh-keyscan` which will then add proper hostkey.

__not sure what to write here__
```

```

### DIFF
--- a/lib/ansible/module_utils/known_hosts.py
+++ b/lib/ansible/module_utils/known_hosts.py
@@ -190,7 +190,10 @@ def add_host_key(module, fqdn, key_type="rsa", create_dir=False, port=SSH_KEYSCA
     elif not os.path.isdir(user_ssh_dir):
         module.fail_json(msg="%s is not a directory" % user_ssh_dir)
 
-    this_cmd = "%s -t %s %s" % (keyscan_cmd, key_type, fqdn)
+    if port == SSH_KEYSCAN_DEFAULT_PORT:
+        this_cmd = "%s -t %s %s" % (keyscan_cmd, key_type, fqdn)
+    else:
+        this_cmd = "%s -t %s %s -p %s" % (keyscan_cmd, key_type, fqdn, port)
 
     rc, out, err = module.run_command(this_cmd)
     # ssh-keyscan gives a 0 exit code and prints nothins on timeout
--- a/lib/ansible/module_utils/known_hosts.py
+++ b/lib/ansible/module_utils/known_hosts.py
@@ -40,6 +40,7 @@ except ImportError:
     import sha as sha1
 
 HASHED_KEY_MAGIC = "|1|"
+SSH_KEYSCAN_DEFAULT_PORT = "22"
 
 def add_git_host_key(module, url, accept_hostkey=True, create_dir=True):
 
@@ -71,10 +72,15 @@ def is_ssh_url(url):
     return False
 
 def get_fqdn(repo_url):
+    return get_fqdn_and_port(repo_url)[0]
+
+def get_fqdn_and_port(repo_url):
 
     """ chop the hostname out of a url """
 
     result = None
+    port = SSH_KEYSCAN_DEFAULT_PORT
+
     if "@" in repo_url and "://" not in repo_url:
         # most likely an user@host:path or user@host/path type URL
         repo_url = repo_url.split("@", 1)[1]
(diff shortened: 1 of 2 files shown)


nev3rm0re:
Wanted to mention - this fixes several issues with the git module: one example is https://github.com/ansible/ansible-modules-core/issues/1886

kostaatsk:
This bug also exists in version ansible 2.2.1.0 (stable). The previous version 2.2.0.0 has no issues. Can you please fix it for both versions 2.2.x ?

abadger:
Is there more to this change than just what's in this PR or perhaps some of it is unnecessary?  It looks like there's changes in two parts of the file but they aren't hooked together.

It seems like add_host_key() needs to have a caller changed to start sending the ssh port to it.  If you can inform me of how this is supposed to work I can review it.

kostaatsk:
@abadger you are right, add_host_key() is called around line 56  https://github.com/ansible/ansible/blob/0d9afa84d50e02c912e90ff6bbe5a89cea30b8be/lib/ansible/module_utils/known_hosts.py and there are no changes in this pull request. ssh port comes from get_fqdn which is passed along to add_host_key as the second argument and port is expected on 5th argument. A bit confusing. @nev3rm0re can you please revise that ?

lib/ansible/module_utils/known_hosts.py
```diff
@@ -71,10 +72,15 @@ def is_ssh_url(url):
...
 
 def get_fqdn(repo_url):
+    return get_fqdn_and_port(repo_url)[0]
+
+def get_fqdn_and_port(repo_url):
 
     """ chop the hostname out of a url """
 
     result = None
+    port = SSH_KEYSCAN_DEFAULT_PORT
```
abadger, line 82: We should not assume SSH port inside of this function.  (that way we'll be able to move it tp a more generic location later).  But we should be able to return a sentinel (like None) and then the caller can set the port to SSH_KEYSCAN_DEFAULT_PORT when that's the case.

nev3rm0re:
Hello!

This is my first submission, so I was a bit nervous not to break anything. Looks like I didn't push the last commit that actually enables the new logic. Will update pull request with required changes.

lib/ansible/module_utils/known_hosts.py
```diff
@@ -190,7 +190,10 @@ def add_host_key(module, fqdn, key_type="rsa", create_dir=False, port=SSH_KEYSCA
     elif not os.path.isdir(user_ssh_dir):
         module.fail_json(msg="%s is not a directory" % user_ssh_dir)
 
-    this_cmd = "%s -t %s %s" % (keyscan_cmd, key_type, fqdn)
```
jimi-c, line 187: I think simply doing this would be better:
<pre>
if port is None:
    port = SSH_KEYSCAN_DEFAULT_PORT
this_cmd = "%s -t %s %s -p %s" % (keyscan_cmd, key_type, fqdn, port)
</pre>
There's no reason to conditionally exclude the `-p` option, it should always be set.

abadger:
As willthames was working on something similar, I asked him to merge the relevant portions of your fix into his PR here: https://github.com/ansible/ansible/pull/20840.  merged into devel today so it should be fixed in 2.3.0.

lib/ansible/module_utils/known_hosts.py
nev3rm0re: Yeah, I totally agree, I just didn't want to break "backwards-compatibility". What if right now Ansible is happily working on some system where `-p` option is not supported? For default port this can be omitted - so I omitted it.

PR #18741, opened 2016-12-05
Fixes Problem 5520 - timeout: ios_config: save

dancwilliams:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME

ansible/lib/ansible/module_utils/ios.py

##### ANSIBLE VERSION

```
ansible 2.2.0.0
```

##### SUMMARY

Fixes Issue 5520

I know that the `write memory` command was once looked at to be deprecated, but now no longer has that designation in the latest Cisco command references.

I adjusted my local Ansible install to use the `write memory` command instead of the `copy running-config startup-config` command and the save option now works correctly within ios_config on both IOS and IOS-XE devices.

Before:
```
The command timed out with the message:
timeout trying to send command: copy running-config startup-config
```
After:  No error.  Save is successful on both IOS and IOS-XE.

### DIFF
--- a/lib/ansible/module_utils/ios.py
+++ b/lib/ansible/module_utils/ios.py
@@ -204,6 +204,6 @@ class Cli(CliBase):
         return self.configure(commands)
 
     def save_config(self):
-        self.execute(['copy running-config startup-config'])
+        self.execute(['write memory'])
 
 Cli = register_transport('cli', default=True)(Cli)


dancwilliams:
Accidentally Closed...

dancwilliams:
Not sure why it is failing this time.  Will have to look.

dancwilliams:
Looks like it is failing on the units test with the error "paramiko is not installed".  Not sure how to overcome that.

PR #18785, opened 2016-12-06
Remove nat_destination check from openstack FIP

emonty:
There is no internal_network attribute on a floating ip and adding one
would be very expensive. Also, nat_destination is a way to find the
right address on a server, it's super helpful in and of itself as an
attribute to use for up-to-date checks.

Finally, the equality checking on the internal network is backwards.


emonty:
Migrated from https://github.com/ansible/ansible-modules-core/pull/5845

gundalow:

PR #18806, opened 2016-12-06
win_file: fix error when creating an existing dir

ilpianista:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

win_file
##### ANSIBLE VERSION

```
ansible 2.1.1.0
```
##### SUMMARY

Fix error when creating an existing dir.

If something else created the dir New-Item will throw an exception and the task will fail. Check the existing file and as long as it's a dir, we don't need to error out.


ilpianista:
Migrated from https://github.com/ansible/ansible-modules-core/pull/4823

gundalow:

PR #18816, opened 2016-12-07
Fixed another map to list

caphrim007:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

bigip_virtual_server.py

##### ANSIBLE VERSION

```
ansible 2.2.0.0
  config file = 
```

##### SUMMARY

For the comparisions that need to be done, this map call needs
to convert to a list because the six import in ansible changes
the behavior of map to return an iterator instead of a list


caphrim007:
@gundalow

PR #18820, opened 2016-12-07
Remove the hint to report module issue elsewhere

dagwieers:
##### ISSUE TYPE

 - Docs Pull Request

##### COMPONENT NAME

GitHub template

##### SUMMARY

Remove the hint to report module issue elsewhere.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,7 +1,7 @@
 <!---
-Please verify first that your issue/request is not already reported in GitHub
+Verify first that your issue/request is not already reported on GitHub.
+Also test if the latest release, and master branch are affected too.
 -->
-
 
 ##### ISSUE TYPE
 <!--- Pick one below and delete the rest: -->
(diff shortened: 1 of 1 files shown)


abadger:
@dagwieers I think that a similar change has already been merged.  Could you either close or rebase this depending on whether you think the ISSUE_TEMPLATE could use more changes?

dagwieers:
@abadger Thanks for bringing ansibot back ! :-)

It's true that in the meantime someone else made the most important change. So I took the time to see what else could be improved. Let me know what you think.

PS We removed the various "Please ..." statements a year ago and made the instructions more direct. So that's what I did now as well.

bcoca:
i would not use regression, since we don't use the type in our workflow at all I don't see it being useful (its still a bug).

dagwieers:
Sure, I thought it would be useful to prepare a release.
Regressions may be useful to list when doing the next RC.
Also, with regressions you can indicate when it last worked.

I will remove it.

dagwieers:
Please re-review !

PR #18822, opened 2016-12-07
Replace module Python 3 fix

mkrupcale:
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
replace module

##### ANSIBLE VERSION
```
ansible 2.3.0 (replace-python3 d3e372fc37) last updated 2016/12/08 15:27:46 (GMT -400)
```

##### SUMMARY
Fixes error in Python 3 where `regexp` and `replace` parameters are strings which can't be searched in a byte object. In particular, an error

`TypeError: cannot use a string pattern on a bytes-like object`

was raised in the `re.subn()` function.

replace (line 104): format `write_changes` function definition with spaces between parameters.
replace (lines 163-164): convert `params['regexp']` and `params['replace']` to bytes for searching `contents`
replace (line 192): import to_bytes

### DIFF
--- a/lib/ansible/modules/files/replace.py
+++ b/lib/ansible/modules/files/replace.py
@@ -101,7 +101,7 @@ EXAMPLES = r"""
     validate: '/usr/sbin/apache2ctl -f %s -t'
 """
 
-def write_changes(module,contents,dest):
+def write_changes(module, contents, dest):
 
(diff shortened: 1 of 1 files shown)


EvanK:
@ansibot This should have been resolved_by_pr #19188

PR #18827, opened 2016-12-07
Remove unneeded quotes - windows/win_unzip.py

Fale:
##### ISSUE TYPE
 - Docs Pull Request

##### COMPONENT NAME
windows/win_unzip.py

##### ANSIBLE VERSION
```
2.2
```

##### SUMMARY
Remove unneeded quotes.

@gundalow


gundalow (commented):
Looks good to me, though needs an extra set of eyes before merging.

jborean93:
I believe you need to double the `\` when using string in the py file. It will only come up as 1 `\` when viewing it in the webdocs format but you will have to double check that.

nitzmahone:
@jborean93 The thing that makes it work is that the herestring is raw (prefaced with `r`) to turn off all Python escaping - that's the way we should be doing any docs with naked Windows paths going forward.

PR #18829, opened 2016-12-07
Make the example in proper yaml format - windows/win_iis_virtualdirectory.py

Fale:
##### ISSUE TYPE
 - Docs Pull Request

##### COMPONENT NAME
windows/win_iis_virtualdirectory.py

##### ANSIBLE VERSION
```
2.2
```

##### SUMMARY
Make examples parsable YAML code (as for guidelines), fix quotation

@gundalow

### DIFF
--- a/lib/ansible/modules/windows/win_iis_virtualdirectory.py
+++ b/lib/ansible/modules/windows/win_iis_virtualdirectory.py
@@ -60,12 +60,24 @@ author: Henrik Wallström
 '''
 
 EXAMPLES = '''
-# This creates a virtual directory if it doesn't exist.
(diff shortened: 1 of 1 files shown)


gundalow (approved):
Looks good to me, though needs an extra set of eyes before merging.

lib/ansible/modules/windows/win_iis_virtualdirectory.py
```diff
@@ -60,12 +60,24 @@
 '''
 
 EXAMPLES = '''
```
nitzmahone: This needs to be a raw string for these to display properly in ansible-doc / HTML doc (preface with `r`). Actually it might be ok in this case, since `\v` isn't a valid escape sequence, but making it a raw string will prevent any future paths from having that issue...

Fale: Thanks. Just pushed the fix

PR #18830, opened 2016-12-07
Make examples in YAML format - windows/win_feature.py

Fale:
##### ISSUE TYPE
 - Docs Pull Request

##### COMPONENT NAME
windows/win_feature.py

##### ANSIBLE VERSION
```
2.2
```

##### SUMMARY
Make examples parsable YAML code (as for guidelines) and improve the examples

@gundalow

### DIFF
--- a/lib/ansible/modules/windows/win_feature.py
+++ b/lib/ansible/modules/windows/win_feature.py
@@ -82,25 +82,27 @@ author:
 '''
 
 EXAMPLES = r'''
-# This installs IIS.
-# The names of features available for install can be run by running the following Powershell Command:
-# PS C:\Users\Administrator> Import-Module ServerManager; Get-WindowsFeature
-$ ansible -i hosts -m win_feature -a "name=Web-Server" all
-$ ansible -i hosts -m win_feature -a "name=Web-Server,Web-Common-Http" all
-ansible -m "win_feature" -a "name=NET-Framework-Core source=C:/Temp/iso/sources/sxs" windows
+- name: Install IIS (Web-Server only)
+  win_feature:
+    name: Web-Server
+    state: present
 
+- name: Install IIS (Web-Server and Web-Common-Http)
+  win_feature:
+    name: Web-Server,Web-Common-Http
+    state: present
 
-# Playbook example
----
-- name: Install IIS
-  hosts: all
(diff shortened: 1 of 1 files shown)


lib/ansible/modules/windows/win_feature.py
```diff
@@ -82,25 +82,27 @@
 '''
 
 EXAMPLES = r'''
-# This installs IIS.
-# The names of features available for install can be run by running the following Powershell Command:
-# PS C:\Users\Administrator> Import-Module ServerManager; Get-WindowsFeature
-$ ansible -i hosts -m win_feature -a "name=Web-Server" all
-$ ansible -i hosts -m win_feature -a "name=Web-Server,Web-Common-Http" all
-ansible -m "win_feature" -a "name=NET-Framework-Core source=C:/Temp/iso/sources/sxs" windows
```
gundalow, line 90: I wonder if this is would be a useful ad-hoc command to document
@nitzmahone what do you think?

gundalow (changes requested):
maybe add ad-hoc command

lib/ansible/modules/windows/win_feature.py
Fale: This is actually against the "Examples should be written in YAML format in plain text in an EXAMPLES string within the module" present in http://docs.ansible.com/ansible/dev_guide/developing_modules.html#example

gundalow: @fale, I can't remember if you were around when we discussed this however for the handful of modules that people *may* execute using ad hoc we said we'd document them in both formats.

Fale: I see it for some, but sincerly I don't really see it for this one. Maybe @nitzmahone can say his opinion on this usage of this module.

Also, we should fix the contributing doc (I'll try to do it today/tomorrow)

nitzmahone: Dunno, I must not have been around for that discussion, because I probably would have objected. Pretty much any module *can* work adhoc, but dealing with all the arg transformation and shell escaping mess to do so (especially with Windows paths) probably doesn't belong in the individual module docs IMHO.

PR #18833, opened 2016-12-07
Add infiniband as a network interface type - facts

martbhell:
##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

facts.py

##### ANSIBLE VERSION

```
ansible 2.3.0 (feature/infiniband_type 51e665d8d2) last updated 2016/12/07 12:36:27 (GMT +300)
```

##### SUMMARY

Add type: "infiniband" for Infiniband interfaces. 

Checked on DDR and FDR Infiniband cards from Mellanox in both HP and Bull servers and it was always /sys/class/net/ib0/type was always "32". Linux kernel code [1] also says that infiniband should have type 32.

```
$ ansible -m setup server|grep -A 80 ansible_ib0|grep type
            "type": "unknown"
```

```
$ ansible -m setup server|grep -A 80 ansible_ib0|grep type
            "type": "infiniband"

```

[1]: https://github.com/torvalds/linux/blob/master/include/uapi/linux/if_arp.h

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -2293,6 +2293,7 @@ class LinuxNetwork(Network):
     platform = 'Linux'
     INTERFACE_TYPE = {
         '1': 'ether',
+        '32': 'infiniband',
         '512': 'ppp',
         '772': 'loopback',
         '65534': 'tunnel',


martbhell:
Better now?

PR #18841, opened 2016-12-07
Fix direct use of fact gathering

zigarn:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

`setup`/fact gathering

##### ANSIBLE VERSION

```
ansible 2.2.0.0
  config file = /tmp/ansible.cfg
  configured module search path = ['library']
```

##### SUMMARY

When programmatically gather facts, we got an error:
>   File \"/tmp/ansible_ogBqT7/ansible_modlib.zip/ansible/module_utils/facts.py\", line 102, in wrapper
> UnboundLocalError: local variable 'seconds' referenced before assignment

See openshift/openshift-ansible#2247

### DIFF
--- a/lib/ansible/module_utils/facts.py
+++ b/lib/ansible/module_utils/facts.py
@@ -83,14 +83,15 @@ class TimeoutError(Exception):
 
 def timeout(seconds=10, error_message="Timer expired"):
 
+    if 'GATHER_TIMEOUT' in globals():
+        if GATHER_TIMEOUT:
+            seconds = GATHER_TIMEOUT
+
     def decorator(func):
         def _handle_timeout(signum, frame):
             raise TimeoutError(error_message)
 
         def wrapper(*args, **kwargs):
-            if 'GATHER_TIMEOUT' in globals():
-                if GATHER_TIMEOUT:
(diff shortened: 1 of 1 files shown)


abadger:
This patch looks obviously wrong but I'm not seeing how the original error would be happening from just what's given here so I'm not sure what's actually going on.... Could you explain how you think the new code is fixing the issue?  And perhaps, a simple reproducer for the issue?

bcoca:
either way the code is wrong, the variable should not override specific values passed to the function, I'll start diff PR the resolve that part (should make this all obsolete).

zigarn:
When fact gathering is used directly in a custom module (like in [openshift/openshift-ansible/openshift_facts](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_facts/library/openshift_facts.py#L1821)], this error is raised.
The patch may be wrong but it works like a charm. (In both case
Looks like the @bcoca PR #19462 may fix the problem.

bcoca:
This patch might avoid the error, but both original code and patch 'break' the actual function, so I'm going to close this in favor of #19462 which fixes the functionality and avoids incorrectly raising the error.

PR #18847, opened 2016-12-07
Fix mkstemp() usage in jenkins_plugin.py

nehaljwani:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

jenkins_plugin
##### ANSIBLE VERSION

```
2.2.0.0
```

##### SUMMARY

Commit ac0e6d2c tried to make the creation of temporary file more
secure. However, the return values of mkstemp() were not handled
correctly, which resulted in the following error:

```
  File "/tmp/ansible_2VBlrF/ansible_module_jenkins_plugin.py", line 829, in <module>
    main()
  File "/tmp/ansible_2VBlrF/ansible_module_jenkins_plugin.py", line 812, in main
    changed = jp.install()
File "/tmp/ansible_2VBlrF/ansible_module_jenkins_plugin.py", line 495, in install
    self._write_file(plugin_file, data)
  File "/tmp/ansible_2VBlrF/ansible_module_jenkins_plugin.py", line 646, in _write_file
    fd = open(tmp_f, 'wb')
TypeError: coercing to Unicode: need string or buffer, tuple found
```
This patch attempts to fix this bug by properly consuming the
return values for `mkstemp()` and fixing some open FDs.


nehaljwani:
A similar PR is https://github.com/ansible/ansible/pull/18849 , but it leaks FDs.

PR #18848, opened 2016-12-07
Add support to the `make` module for targets with options

stevekuznetsov:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

`make`
##### ANSIBLE VERSION

```
$ ansible --version
ansible 2.2.0.0
```
##### SUMMARY

In order to run a `make` target with options, we need to not
assume that the target that is passed in to the module is to
be interpreted as one command-line argument. This allows us
to support targets with options like:
  $ make some-target -o some-file

Where the `some-target` target skips rebuilding `some-file`
even if the file is old. Currently, the implementation would
attempt to run the `some-target -o some-file` target, which
does not exist.

`make` does not allow for targets with spaces in their names,
so we should be safe to break up the command-line arguments
like this and not risk mutilating a valid target name.

Signed-off-by: Steve Kuznetsov skuznets@redhat.com
###### BEFORE

```
TASK [make : execute make target(s) "test -o build" in the origin-web-console repository] ***
failed: [openshiftdevel] (item=test -o build) => {
    "cmd": "/bin/make 'test -o build'", 
    "failed": true, 
    "invocation": {
        "module_args": {
            "chdir": "/data/src/github.com/openshift/origin-web-console", 
            "params": null, 
            "target": "test -o build"
        }, 
        "module_name": "make"
    }, 
    "item": "test -o build", 
    "msg": "make: *** No rule to make target 'test -o build'.  Stop.", 
    "rc": 2, 
    "stderr": "make: *** No rule to make target 'test -o build'.  Stop.\n", 
    "stdout": "", 
    "stdout_lines": []
}
```
###### AFTER

```
TASK [make : execute make target(s) "test -o build" in the origin-web-console repository] ***
failed: [openshiftdevel] (item=test -o build) => {
    "cmd": "/bin/make test -o build", 
    "failed": true, 
    "invocation": {
        "module_args": {
            "chdir": "/data/src/github.com/openshift/origin-web-console", 
            "params": null, 
            "target": "test -o build"
        }, 
        "module_name": "make"
    }, 
    "item": "test -o build", 
    "msg": "<snip>make: *** [test] Error 3", 
    "rc": 2, 
    "stderr": "<snip>make: *** [test] Error 3\n", 
    "stdout": "<snip>Makefile:45: recipe for target 'test' failed\n", 
    "stdout_lines": [
        <snip>
        "Makefile:45: recipe for target 'test' failed"
    ]
}
```

(example after has a failed `make` target, but it runs correctly, just that example failed in the middle of the run)

@LinusU @gregdek  I've added the version comment in the documentation.

### DIFF
--- a/lib/ansible/modules/system/make.py
+++ b/lib/ansible/modules/system/make.py
@@ -34,7 +34,7 @@ description:
 options:
   target:
     description:
-      - The target to run
+      - The target to run. As of Ansible 2.3, options can be specified as well.
     required: false
     default: none
   params:
@@ -66,14 +66,16 @@ EXAMPLES = '''
     params:
       NUM_THREADS: 4
       BACKEND: lapack
+
+# Skip building a file as part of a target
+- make:
+    chdir: /home/ubuntu/cool-project
+    target: 'all -o somefile'
 '''
 
 # TODO: Disabled the RETURN as it was breaking docs building. Someone needs to
 # fix this
 RETURN = '''# '''
-
-from ansible.module_utils.six import iteritems
-from ansible.module_utils.basic import AnsibleModule
 
 
 def run_command(command, module, check_rc=True):
@@ -100,9 +102,9 @@ def sanitize_output(output):
     :return: sanitized output
     """
     if output is None:
-        return ''
+        return b('')
     else:
-        return output.rstrip("\r\n")
+        return output.rstrip(b("\r\n"))
 
 
 def main():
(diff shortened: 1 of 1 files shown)


lib/ansible/modules/system/make.py
```diff
@@ -34,7 +34,7 @@
 options:
   target:
     description:
-      - The target to run
+      - The target to run. As of Ansible 3.2, options can be specified as well.
```
resmo: Ansible 3.2? we are not there yet 😂

stevekuznetsov: I'll blame that one on dyslexia :)

stevekuznetsov:
@jimi-c do you know why this `needs_revision`?

stevekuznetsov:
Closing for inactivity ...

PR #18858, opened 2016-12-07
Docs (jenkins_plugin): Specify version as string

nehaljwani:
##### ISSUE TYPE

 - Docs Pull Request

##### COMPONENT NAME

jenkins_plugins
##### ANSIBLE VERSION

```
2.2.0.0
```

##### SUMMARY

If the plugin version expected is, say '1.20', then specifying it
as...
```
    version: 1.20
```
... will make the YAML parser interpret it as a float, and the
value obtained by the module will be 1.2 instead of 1.20, which
will cause downloading of wrong version of the module.

This patch updates the docs so that users don't face this issue.

### DIFF
--- a/lib/ansible/modules/web_infrastructure/jenkins_plugin.py
+++ b/lib/ansible/modules/web_infrastructure/jenkins_plugin.py
@@ -120,6 +120,8 @@ options:
         manually.
       - It might take longer to verify that the correct version is installed.
         This is especially true if a specific version number is specified.
+      - Quote the version to prevent the value to be interpreted as float. For
+        example if C(1.20) would be unquoted, it would become C(1.2).
   with_dependencies:
     required: false
     choices: ['yes', 'no']
@@ -159,7 +161,7 @@ EXAMPLES = '''
 - name: Install specific version of the plugin
   jenkins_plugin:
     name: token-macro
-    version: 1.15
+    version: "1.15"
 
 - name: Pin the plugin
   jenkins_plugin:
@@ -212,7 +214,7 @@ EXAMPLES = '''
       token-macro:
         enabled: yes
       build-pipeline-plugin:
-        version: 1.4.9
+        version: "1.4.9"
         pinned: no
         enabled: yes
   tasks:


lib/ansible/modules/web_infrastructure/jenkins_plugin.py
```diff
@@ -120,6 +120,8 @@
         manually.
       - It might take longer to verify that the correct version is installed.
         This is especially true if a specific version number is specified.
+      - Put the value under quotes so that values like '1.2.30' are not parsed
+        incorrectly as 1.2.3
```
jtyr: The example here is incorrect as the `1.2.30` will not be interpreted as `1.2.3`. Please use the example from the description of the issue where `1.20` is interpreted as `1.2`.

Please change the text to this:

```
      - Quote the version to prevent the value to be interpreted as float. For
        example if C(1.20) would be unquoted, it would become C(1.2).
```

lib/ansible/modules/web_infrastructure/jenkins_plugin.py
```diff
@@ -159,7 +161,7 @@
 - name: Install specific version of the plugin
   jenkins_plugin:
     name: token-macro
-    version: 1.15
+    version: '1.15'
```
jtyr: Please use double quotes.

lib/ansible/modules/web_infrastructure/jenkins_plugin.py
```diff
@@ -212,7 +214,7 @@
       token-macro:
         enabled: yes
       build-pipeline-plugin:
-        version: 1.4.9
+        version: '1.4.9'
```
jtyr: Please use double quotes.

jtyr (changes requested):

nehaljwani:
@jtyr Updated patch as per your review.

jtyr:
Thanks for your contribution, @nehaljwani. @ansibot, please `shipit`.

PR #18877, opened 2016-12-07
Add 'update' parameter in junos_config module

ganeshrn:
##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME
junos_config

##### ANSIBLE VERSION

```
ansible --version
ansible 2.3.0 
```

##### SUMMARY

```
Add update parameter in junos_config module which supports
configuration action like merge, replace and overwrite.

```
Related PR https://github.com/ansible/ansible/pull/18886
Fixes issue: https://github.com/ansible/ansible-modules-core/issues/5839


hlantz (commented):
Hi @ganeshnalawade - do you also need to include the replace variable on the self.config.load() statement on (new) lines 186-187 here, or is that implicit if overwrite and merge are both False?

ganeshrn:
@hlantz Yes by default the action for 'self.config.load()' is 'replace' hence not required to pass it.
Did the patch work for you?

hlantz:
@ganeshnalawade I didn't have a chance to try it out until now (because timezones), but it appears to work.  I've tested the following cases:
- **replace** full config with no changes = OK (idempotent)
- **replace** full config with bits added = OK
- **replace** full config with bits removed = OK (does not remove the missing bits, same as **load replace**)
- **replace** full config changes to exclusive values such as VLANs on access ports = OK (well, the actual config fails, but this is the same behaviour as for **load replace** in my test cases, so it's technically a pass)
- **overwrite** full configuration = OK (idempotent)
- **merge** individual snippets = OK (idempotent)

So yes - I would say that the patch works perfectly.  Thank you!

ganeshrn:
@hlantz Thank you for reviewing and testing exhaustively

ganeshrn:
Seems to be affected by recent migration. Will close this PR and raise a new one.

PR #18890, opened 2016-12-07
Fixes #3539 "win_robocopy does not return changed properly"

### LINKED ISSUE #3539: Bug in Conditional Include
Hi,

I know that when using conditionals on an include, 'All the tasks get evaluated, but the conditional is applied to each and every task'.  However this breaks when some of that tasks register variables and other tasks in the group use those variable.

Example:

main.yml:

```
- include: extra.yml
  when: do_extra is defined
```

extra.yml:

```
- name: check if we can do task A
  shell: check_if_task_A_possible
  register: A_possible
  ignore_errors: yes

- name: task A
  shell: run_task_A
  when: A_possible.rc == 0
```

Now if you run main.yml and 'do_extra' is not defined, the run will fail on 'task A'  because when the 'when' condition is evaluated, the variable A_possible will not exist.

It is not sufficient to just add the top-level include conditional above the other because right now it looks like the two conditions are compounded and tested together which will still fail because A_possible is not defined.  I think you would have to run the file level conditional before the task level ones to keep this from happening.

sherlockmatt:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

win_robocopy

##### ANSIBLE VERSION

```
ansible 2.2.1.0
  config file = /ansible/ansible_config
```

##### SUMMARY

Fixes #3539 "win_robocopy does not return changed properly":
Remove the win_robocopy object from $result. This brings the code in line with the documentation, and correctly sets the changed attribute.

BEFORE:
```
ok: [x.x.x.x] => (item=path/to/file.ext) => {
    "changed": false,
    ...
    "win_robocopy": {
        "changed": true
    },
    ...
```

AFTER:
```
changed: [x.x.x.x] => (item=path/to/file.ext) => {
    "changed": true,
    ...
```

I can't copy and paste from my ansible machine so I've manually typed a shortened output. It should still be enough to get the point though.

Also worth noting that this doesn't fix #3498, that will still need to be merged.


sherlockmatt:
Migrated from https://github.com/ansible/ansible-modules-extras/pull/3540

gundalow:

twistyWater:
@sherlockmatt The PR is closed but not merged, what is happening with it?

gundalow:
@twistyWater Is this it https://github.com/ansible/ansible-modules-extras/issues/3539

twistyWater:
@gundalow ansible/ansible-modules-extras#3539 is the original, this PR is the moved one I think, but there was a problem as you commented on Dec 8. it doesn't seem to be added to ansible/community#147

gundalow:
Hum, strange
Doesn't look like @sherlockmatt re did the prmove

https://github.com/ansible/ansible/pulls?utf8=%E2%9C%93&q=is%3Apr%20author%3Asherlockmatt doesn't show any other PRs (open nor closed)

sherlockmatt:
Ah yes I forgot to redo it. Thanks for reminding me :) It was a hellishly busy time at work so I never got round to it.

EDIT: Looks like 147 is closed (I think I'd be a bit late for the meeting anyway now :P). Should I add it somewhere else instead?

twistyWater:
@sherlockmatt Try https://github.com/ansible/community/issues/148

PR #18905, opened 2016-12-07
Removing a non-existing directory complains

dagwieers:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

file

##### ANSIBLE VERSION
v2.2

##### SUMMARY

The following playbook:

```yaml
- hosts: localhost
  tasks:
  - file:
      path: /tmp/non-existing-foo-bar
      state: absent
      recurse: yes
```

causes this error:

```
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "recurse option requires state to be 'directory'", "path": "/tmp/non-existing-foo-bar", "state": "absent"}
```

The included fix ensures that when recurse is added, we no longer assume
it is a file, but accept that it is a directory.


dagwieers:
Migrated from https://github.com/ansible/ansible-modules-core/pull/5667

gundalow:

PR #18920, opened 2016-12-07
fixes #4441 passing subnet_name parameter

### LINKED ISSUE #4441: Add a "move_home" parameter to the user module
The following playbook fails with the message `msg: usermod: directory /tmp exists`:

```

---
- hosts: server
  tasks:
  - user: name=test
          system=yes
          home=/
          state=present
  - user: name=test
          system=yes
          home=/tmp
          state=present
```

This is because the user module passes the `-m` option for modifying a user when the home directory option is passed (`user` module, line 366), which would move the contents of the previous home directory.

Would it be better to accept an argument to control this behavior?

But ideally, I'd think that the user module should simply not pass -m implicitly, instead gaining a move_home= boolean option (default false) and pass that flag only if this is true. An admin should know if they expect the home directory to move, and accidentally doing so can cause all sorts of havoc.

pmarques:
As far as I could check this is related to this [function](https://github.com/Azure/azure-sdk-for-python/blob/master/azure-mgmt-network/azure/mgmt/network/operations/subnets_operations.py#L128) on azure python SDK

### DIFF
--- a/lib/ansible/modules/cloud/azure/azure_rm_virtualmachine.py
+++ b/lib/ansible/modules/cloud/azure/azure_rm_virtualmachine.py
@@ -1243,7 +1243,7 @@ class AzureRMVirtualMachine(AzureRMModuleBase):
 
         if self.subnet_name:
             try:
-                subnet = self.network_client.subnets.get(self.resource_group, virtual_network_name)
+                subnet = self.network_client.subnets.get(self.resource_group, virtual_network_name, self.subnet_name)
                 subnet_id = subnet.id
             except Exception as exc:
                 self.fail("Error: fetching subnet {0} - {1}".format(self.subnet_name, str(exc)))


pmarques:
Migrated from https://github.com/ansible/ansible-modules-core/pull/4442

pmarques:
@nitzmahone , @chouseknecht this was closed by mistake but I think the issue still exist, can you please check it.

paprickar:
Hey,

for me this merge would be really important is there any timeframe when you are able to merge this?

(I got the exact same issue)

paprickar (commented):
With these changes i was able to deploy a Azure VM in a existing vnet and subnet. If i can help please let me know.

pmarques:
Hi @paprickar, can you confirm that the issue is present on the last version of Ansible? I will reopen the issue if you tell me so.

paprickar:
Right now i am using ansible 2.2.2.0 and its still affected from this issue.

I checked the source code from stable-2.3 and the entry is still missing. But i did not tested it with stable-2.3 because i cant install it via pip. I will now go with the manual patched version and get my tasks done. I will then try to run it from source.

pmarques:
@paprickar thanks for the details! The PR is open and I had to rebase due to merge conflicts.
If the authors can have a look will be great @chouseknecht @nitzmahone 

Mean while, what I'm doing to keep the patched copy in my local library in the playbook

abadger:
Fix for this bug: https://github.com/ansible/ansible-modules-core/issues/4441

nitzmahone (approved):

PR #18922, opened 2016-12-07
Created digital_ocean_floating_ip module

pmarques:
##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

cloud/digital_ocean/digital_ocean_floating_ip

##### SUMMARY

Add support for create delete and assign Digital Ocean Floating IPs to a droplet.


pmarques:
Migrated from https://github.com/ansible/ansible-modules-core/pull/5221

lib/ansible/modules/cloud/digital_ocean/do_floating_ip.py
```diff
@@ -0,0 +1,330 @@
...
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
++ANSIBLE_METADATA = {'status': ['preview'],
```
abadger: Note -- this PR is failing tests because this line has an extra leading plus sign.

pmarques:
Moved facts module to new PR  https://github.com/ansible/ansible/pull/26372

lib/ansible/modules/cloud/digital_ocean/do_floating_ip.py
```diff
@@ -0,0 +1,329 @@
...
+
+notes:
+  - Version 2 of DigitalOcean API is used.
+requirements:
+  - "python >= 2.6"
+'''
+
+
+EXAMPLES = '''
+- name: "Create a Floating IP in regigin lon1"
```
BondAnthony: Spelling.. "region"

lib/ansible/modules/cloud/digital_ocean/do_floating_ip.py
```diff
@@ -0,0 +1,329 @@
...
+    region: lon1
+
+- name: "Create a Floating IP assigned to Droplet ID 123456"
+  do_floating_ip:
+    state: present
+    droplet_id: 123456
+
+- name: "Delete a Floating IP with ip 1.2.3.4"
+  do_floating_ip:
+    state: present
```
BondAnthony: Shouldn't this be absent instead of present?

lib/ansible/modules/cloud/digital_ocean/do_floating_ip.py
```diff
@@ -0,0 +1,329 @@
...
+    required: true
+
+notes:
+  - Version 2 of DigitalOcean API is used.
+requirements:
+  - "python >= 2.6"
+'''
+
+
+EXAMPLES = '''
```
BondAnthony: None of your examples below actually list the required parameter of "oauth_token". Secondly, adding oauth_token as the token parameter would create three different "api token" parameter names within the DO modules. Do you think we should pick a standard and stick with that one as we migrate the modules to the v2 api? Currently the DO modules use "api_token" and "api_key". I believe the api_key is only used in the v1 ssh keys module. Since you have written a new v2 implementation of ssh_keys should we stick with "api_token" or "oauth_token"?

lib/ansible/modules/cloud/digital_ocean/do_floating_ip.py
```diff
@@ -0,0 +1,329 @@
...
+    module = AnsibleModule(
+        argument_spec=dict(
+            state=dict(choices=['present', 'absent'], default='present'),
+            ip=dict(aliases=['id'], required=False),
+            region=dict(required=False),
+            droplet_id=dict(required=False),
+            oauth_token=dict(
+                no_log=True,
+                # Support environment variable for DigitalOcean OAuth Token
+                fallback=(env_fallback, ['DO_OAUTH_TOKEN']),
```
BondAnthony: Other DO modules are already looking at "DO_API_TOKEN", "DO_API_KEY". Should we keep this standard as we move forward so users don't have to setup more than one environment variable?

Please see above comments about the use of oauth_token.

BondAnthony (commented):
Couple questions and some documentation changes

BondAnthony:
needs_revision

lib/ansible/modules/cloud/digital_ocean/do_floating_ip.py
pmarques: Good points, the first I completely agree (I was using the env var and I forgot to add that one).
About the name of the **_token*. If you are asking my opinion, I called it *ouath* because is what it is although, I'm fine with keeping the old name or at least make an alias to keep backward compatibility (if this makes sense).

lib/ansible/modules/cloud/digital_ocean/do_floating_ip.py
pmarques: nice catch!

lib/ansible/modules/cloud/digital_ocean/do_floating_ip.py
BondAnthony: Since the newer modules coming in will be named do_* I think now would be a good time to start pushing a standard. I second your point about making the parameter oauth_token since it provides some meaning to the end users. Since your adding a number of new modules and we only have three working DO modules today it will be easier to update the existing three to oauth_token.

Just wanting to know what you were thinking since you have a number of PRs. Thank you for contributing!

lib/ansible/modules/cloud/digital_ocean/do_floating_ip.py
BondAnthony: Probably best to include the older DO_API_TOKEN env as an additional failback.

lib/ansible/modules/cloud/digital_ocean/do_floating_ip.py
pmarques: Sure, I can create a new PR to add the alias to **oauth_token** to the old modules.

BondAnthony (approved):

BondAnthony:
shipit

@abadger up for you to review.

lib/ansible/modules/cloud/digital_ocean/do_floating_ip.py
pmarques: I just noticed that old module use **DO_API_KEY** and **DO_API_TOKEN**, so I added **DO_API_KEY** to keep the previous ones.

BondAnthony (approved):

BondAnthony:
@pmarques I would assume you will be changing the name of this module based on the changes that occurred on #18930?

pmarques:
@BondAnthony yes, actually I need to change other PRs too (#26365  #26372), just waiting for @ryansb to approve.

pmarques:
btw, this modules still tagged with **affects_2.3** should it be updated to **affects_2.4**

pmarques:
@BondAnthony change all my PRs to match **digital_ocean_***

BondAnthony:
-label affects_2.3

BondAnthony:
+label affects_2.4

lib/ansible/modules/cloud/digital_ocean/digital_ocean_floating_ip.py
```diff
@@ -0,0 +1,336 @@
...
+
+notes:
+  - Version 2 of DigitalOcean API is used.
+requirements:
+  - "python >= 2.6"
+'''
+
+
+EXAMPLES = '''
+- name: "Create a Floating IP in regigin lon1"
```
BondAnthony: Mispelling of "region"

BondAnthony:
shipit

lib/ansible/modules/cloud/digital_ocean/digital_ocean_floating_ip.py
```diff
@@ -0,0 +1,336 @@
...
+        ]),
+        mutually_exclusive=(
+            ['region', 'droplet_id']
+        ),
+    )
+
+    try:
+        core(module)
+    except Exception as e:
+        module.fail_json(msg=str(e))
```
abadger: catchall exception handling isn't very useful as it hides whatever might have gone wrong from sight.  I'd probably just call core(module) here.   If there is a specific class of errors that you are expecting, then catching those and returning a more informative error message could be appropriate.  I often see the following pattern used which isn't so bad:
``` python
import traceback

from ansible.module_utils._text import to_native

try:
    core(module)
except SomeException as e:
    module.fail_json(msg="Helpful error message: %s" % to_native(e), exception=traceback.format_exc())
```

lib/ansible/modules/cloud/digital_ocean/digital_ocean_floating_ip.py
```diff
@@ -0,0 +1,336 @@
...
+        return self.send('POST', path, data, headers)
+
+    def delete(self, path, data=None, headers=None):
+        return self.send('DELETE', path, data, headers)
+
+
+def wait_action(module, rest, ip, action_id, timeout=10):
+    end_time = time.time() + 10
+    while time.time() < end_time:
+        response = rest.get('floating_ips/{}/actions/{}'.format(ip, action_id))
```
abadger: I see that the requirements lists python-2.6 as the minimum python version.  Unfortunately, format strings in python-2.6 are more limited than in python-2.7 and later.  So you need to be explicit about the position in the format args list you are looking at like this:
``` python
response = rest.get('floating_ips/{0}/actions/{1}'.format(ip, action_id))
```

There's a few other places with format strings that have to be fixed as well.

abadger:
Added a few comments of things to fix.  Once those are done this can be merged.  Feel free to ping me to merge this via IRC if you want.  I'm abadger1999 in #ansible-devel on irc.freenode.net.

lib/ansible/modules/cloud/digital_ocean/digital_ocean_floating_ip.py
pmarques: Ups, my bad. I forgot to change this in some modules. Should be fixed.

lib/ansible/modules/cloud/digital_ocean/digital_ocean_floating_ip.py
pmarques: True, I should be using the same behaviour on PR #18930, just calling core(module)

abadger:
merged to devel for 2.4.0  Thanks @pmarques and @BondAnthony!

PR #18946, opened 2016-12-08
Set stats

bcoca:
##### ISSUE TYPE

 - Feature Pull Request

##### ANSIBLE VERSION

```
2.3
```

### DIFF
--- a/lib/ansible/executor/stats.py
+++ b/lib/ansible/executor/stats.py
@@ -19,6 +19,8 @@
 from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
+from ansible.utils.vars import merge_hash
+
 class AggregateStats:
     ''' holds stats about per-host activity during playbook runs '''
 
@@ -30,6 +32,9 @@ class AggregateStats:
         self.dark      = {}
         self.changed   = {}
         self.skipped   = {}
+
+        # user defined stats, which can be per host or global
+        self.custom = {}
 
     def increment(self, what, host):
         ''' helper function to bump a statistic '''
@@ -48,4 +53,32 @@ class AggregateStats:
             changed     = self.changed.get(host, 0),
             skipped     = self.skipped.get(host, 0)
         )
+
+    def set_custom_stats(self, which, what, host=None):
+        ''' allow setting of a custom stat'''
+
+        if host is None:
+            host = '_run'
+        if host not in  self.custom:
+            self.custom[host] = {which: what}
+        else:
+            self.custom[host][which] = what
+
+    def update_custom_stats(self, which, what, host=None):
+        ''' allow aggregation of a custom stat'''
+
+        if host is None:
+            host = '_run'
+        if host not in self.custom or which not in self.custom[host]:
+            return self.set_custom_stats(which, what, host)
+
+        # mismatching types
+        if type(what) != type(self.custom[host][which]):
+            return None
+
+        if isinstance(what, dict):
+            self.custom[host][which] =  merge_hash(self.custom[host][which], what)
+        else:
+            # let overloaded + take care of other types
+            self.custom[host][which] += what
 
--- a/lib/ansible/modules/utilities/logic/set_stats.py
+++ b/lib/ansible/modules/utilities/logic/set_stats.py
@@ -1,0 +1,69 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# Copyright 2016 Ansible RedHat, Inc
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+ANSIBLE_METADATA = {'status': ['preview'],
+                    'supported_by': 'community',
+                    'version': '1.0'}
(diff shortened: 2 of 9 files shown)


jlaska:
@bcoca sounds like this might need a rebase?

alikins:
Needs tests.

lib/ansible/executor/stats.py
```diff
@@ -49,3 +54,31 @@ def summarize(self, host):
             skipped     = self.skipped.get(host, 0)
         )
 
+    def set_custom_stats(self, which, what, host=None):
+        ''' allow setting of a custom fact '''
```
alikins: s/fact/stat

lib/ansible/executor/stats.py
```diff
@@ -49,3 +54,31 @@ def summarize(self, host):
...
+
+        if host is None:
+            host = '_run'
+        if host not in  self.custom:
+            self.custom[host] = {which: what}
+        else:
+            self.custom[host][which] = what
+
+    def update_custom_stats(self, which, what, host=None):
+        ''' allow aggregation of a custom fact '''
```
alikins: s/fact/stat

alikins (commented):
Some cut&paste errors

lib/ansible/plugins/callback/default.py
```diff
@@ -268,6 +268,22 @@ def v2_playbook_on_stats(self, stats):
 
         self._display.display("", screen_only=True)
 
+        # print custom stats
+        if C.SHOW_CUSTOM_STATS and stats.custom:
+            self._display.banner("CUSTOM STATS: ")
+            # per host
+            #TODO: come up with 'pretty format'
+            for k in sorted(stats.custom.keys()):
```
alikins, line 276: Could use some docs/comments explaining what the stats.custom data structure looks like.

bcoca: its up to the user, the top level is just the 'var name of the stat', after that it can be anything

lib/ansible/executor/stats.py
```diff
@@ -49,3 +54,31 @@ def summarize(self, host):
             skipped     = self.skipped.get(host, 0)
         )
 
+    def set_custom_stats(self, which, what, host=None):
+        ''' allow setting of a custom stat'''
+
+        if host is None:
+            host = '_run'
+        if host not in  self.custom:
```
alikins, line 62: Could we use ```defaultdict(dict)``` here?

alikins (commented):
Could we use defaultdict(dict) here?

lib/ansible/plugins/strategy/__init__.py
```diff
@@ -491,30 +495,37 @@ def parent_handler_match(target_handler, handler_name):
...
-                            else:
-                                host_list = [actual_host]
-
                             for target_host in host_list:
                                 if original_task.action == 'set_fact':
                                     self._variable_manager.set_nonpersistent_facts(target_host, result_item['ansible_facts'].copy())
                                 else:
                                     self._variable_manager.set_host_facts(target_host, result_item['ansible_facts'].copy())
 
+                    if 'ansible_stats' in result_item and 'data' in result_item['ansible_stats'] and result_item['ansible_stats']['data']:
```
alikins, line 513: May be worthwhile to make lines 513-528 a method on the class, so strategy sub classes could override it easier. May also make it easier to write tests for.

lib/ansible/modules/utilities/logic/set_stats.py
```diff
@@ -0,0 +1,69 @@
...
+options:
+  data:
+    description:
+      - A dictionary of which each key represents a stat (or variable) you want to keep track of
+    required: true
+  per_host:
+    description:
+        - boolean that indicates if the stats is per host or for all hosts in the run.
+    required: no
+    default: yes
```
chrismeyersfsu: Would prefer the `per_host` default to be `no`

alikins (changes requested):
If the custom stat aggregation is something that a different strategy plugin may need to extend, may be worthwhile moving this code into it's own method.

lib/ansible/executor/stats.py
bcoca: possibly, this does not need to be 2.4, i'm fine either way

lib/ansible/plugins/strategy/__init__.py
```diff
@@ -491,30 +495,37 @@ def parent_handler_match(target_handler, handler_name):
...
+
+                        if 'per_host' not in result_item['ansible_stats'] or result_item['ansible_stats']['per_host']:
+                            host_list = self.get_task_hosts(iterator, original_host, original_task)
+                        else:
+                            host_list = [None]
+
+                        data = result_item['ansible_stats']['data']
+                        aggregate = 'aggregate' in result_item['ansible_stats'] and result_item['ansible_stats']['aggregate']
+                        for myhost in host_list:
```
alikins, line 522: Could this block (520-528) be a method on the AggregateStats class? something like AggregateStats.aggregate_custom_stats(data=result_item['ansible_stats'], host_list=host_list)

lib/ansible/modules/utilities/logic/set_stats.py
bcoca: done

lib/ansible/plugins/strategy/__init__.py
bcoca: we can refactor that easily if we ever get the the point of needing it, i really dont see a use case so I consider that a premature optimization at this point

lib/ansible/plugins/callback/default.py
```diff
@@ -268,6 +268,22 @@ def v2_playbook_on_stats(self, stats):
 
         self._display.display("", screen_only=True)
 
+        # print custom stats
+        if C.SHOW_CUSTOM_STATS and stats.custom:
+            self._display.banner("CUSTOM STATS: ")
```
alikins, line 273: Splitting this block (273-286) into it's own method would make it easier for sub classes of the default CallbackModule (like debug, skippy, actionable, dense) to extend this behavior.

chrismeyersfsu:
LGTM 👍

Issue #11347, opened 2015-06-22
authorized_key: exclusive=yes with multiple key files


crane-denny:
If I `exclusive=yes` is set it is not possible to add multiple keys by looping through a list of files. They has to be in one file (like mentioned in the docs ;)). Otherwise the last keys wins and every update would trigger a change. 

Something like this would be nice if possible:

``` yaml
- name: root public key authentication
  authorized_key:
    user: root
    key: "{{ item }}"
    exclusive: yes
  with_file:
    - key1.pub
    - key2.pub
    - key3.pub
```

bcoca:
wrong repo, this should be in http://github.com/ansible/ansible-modules-core, in any case fixed via  
https://github.com/ansible/ansible-modules-core/commit/6f6d7f5c18296e3fe84fa5aef674948753ff52ae

Issue #1253, opened 2012-10-08
Ansible should parse the playbook before asking for sudo or ssh passwords


ghost:
For example, I have a playbook named remotesetup.yml and I run it from the command line like that:

ansible-playbook remotesetupp.yml -K 

It will ask for the sudo password even if the file doesn't exist (I made a typo, two p at the end).
In the same way, if there's a yaml syntax error in the playbook it will still ask me for the sudo pass even though it can't run the playbook.

Thus, the parsing of the playbook should be done prior to asking for sudo or ssh pass.

mpdehaan:
duplicate of #1227

PR #18951, opened 2016-12-08
Add example for multiple authorized_keys

svenstaro:
##### ISSUE TYPE
- Docs Pull Request
##### ANSIBLE VERSION

```
devel
```

##### COMPONENT NAME
authorized_key

##### SUMMARY

This adds a rather complicated but important example to the authorized_keys module. It shows how to use `exclusive=yes` with multiple ssh keys. I feel that this is a common scenario for many users with no obvious solution. I've taken quite some time to come up with this solution so I think it's a good idea to share this. It looks like other people have also had trouble coming up with this: ansible/ansible#11347, #1253, http://serverfault.com/questions/665410/using-ansible-to-add-users-public-keys-to-one-users-authorized-keys-file, http://stackoverflow.com/questions/33032747/ansible-how-to-concatenate-files-contents-into-a-variable

### DIFF
--- a/lib/ansible/modules/system/authorized_key.py
+++ b/lib/ansible/modules/system/authorized_key.py
@@ -153,6 +153,16 @@ EXAMPLES = '''
     user: ubuntu
     state: present
     key: "{{ lookup('file', lookup('env','HOME') + '/.ssh/id_rsa.pub') }}"
+
+- block:
+  - name: Set up authorized_keys exclusively with many keys
+    set_fact: pubkey_list="{{ lookup('file', 'pubkeys/' + item) }}"
+    register: pubkeys
+    with_items:
+      - key1.pub
+      - key2.pub
+  - set_fact: pubkey_string={{ pubkeys.results | map(attribute='ansible_facts.pubkey_list') | join('\n') }}
+  - authorized_key: user=root key="{{ pubkey_string }}" exclusive=yes
 '''
 
 RETURN = '''


svenstaro:
Original PR here: https://github.com/ansible/ansible-modules-core/pull/4167

gundalow:
@svenstaro Rebase needed, you've pulled in a lot of other peoples changes

svenstaro:
@gundalow Done.

lib/ansible/modules/system/authorized_key.py
```diff
@@ -152,6 +152,17 @@
     user: ubuntu
     state: present
     key: "{{ lookup('file', lookup('env','HOME') + '/.ssh/id_rsa.pub') }}"
+
+# Set up authorized_keys exclusively with many keys
```
gundalow: Please use multiline yaml and set a  `- name:` rather than a comment which is more readable and matches the other examples in this file

svenstaro:
@gundalow I believe I've found a tidy solution by using a `block` to group all of my statements. This makes it easier to see that all commands belong together.

svenstaro:
I do not understand the YAML validation problem. I believe that I contributed valid YAML. @gundalow could you please point out to me what is incorrect here?

lib/ansible/modules/system/authorized_key.py
```diff
@@ -153,6 +153,16 @@
...
     key: "{{ lookup('file', lookup('env','HOME') + '/.ssh/id_rsa.pub') }}"
+
+- block:
+  - name: Set up authorized_keys exclusively with many keys
+    set_fact: pubkey_list="{{ lookup('file', 'pubkeys/' + item) }}"
+    register: pubkeys
+    with_items:
+      - key1.pub
+      - key2.pub
+  - set_fact: pubkey_string={{ pubkeys.results | map(attribute='ansible_facts.pubkey_list') | join('\n') }}
```
gundalow, line 164: Missing `"..."`
`set_fact: pubkey_string="{{ pubkeys.results | map(attribute='ansible_facts.pubkey_list') | join('\n') }}"`

sivel:
I'm not really sold on this solution.  I would claim that using `register` on `set_fact` is an anti-pattern, and should be avoided.

Some other options:

```
    - slurp:
        path: "{{ item }}"
      with_fileglob: "./pubkeys/*"
      delegate_to: localhost
      register: pubkeys

    - authorized_keys:
        user: root
        key: "{{ pubkeys.results|map(attribute='content')|map('b64decode')|join('\n') }}"
        exclusive: yes
```

```
    - set_fact:
        pubkeys: "{% for key in keys %}{{ lookup('file', key) ~ '\n' }}{% endfor %}"
      vars:
        keys:
          - ./pubkeys/one
          - ./pubkeys/two
```

yvbeek:
Perhaps useful for people looking for a solution, I use:

```yaml
- name: Combine SSH keys
  local_action: shell cat files/ssh-keys/*
  register: ssh_keys
  become: no
  changed_when: no
  check_mode: no

- name: Set SSH keys
  authorized_key: key={{ssh_keys.stdout}} user=myuser exclusive=yes
```

PR #19007, opened 2016-12-08
set_extra_var module to set extra variable from play

PieterVoet:
##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

set_extra_var
##### ANSIBLE VERSION

```
ansible 2.3.0
  config file = /home/pieter/PycharmProjects/myansible/ansible.cfg
  configured module search path = ['/home/pieter/ansible/modules']

```

##### SUMMARY

Sets an extra variable from a playbook. If the 'extra variable' is already been set from
the command line, it will NOT be overruled by the module.

Example inventory :

group_all = { host_1, host_2, host_3 }
group_one = { host_1 }
group_two = { host_2, host3 }

Example playbook 'my_playbook.yml' :
---
- hosts: group_one
  tasks:
  - name: list /tmp
    shell: ls /tmp
    register: out

  - name: set extra variable
    set_extra_var: name=myvar value="{{out}}"

- hosts: group_two
  tasks:
    debug: msg="{{myvar}}"

Example command

ansible-playbook my_playbook.yml --limit group_all

```

```


jhawkesworth:
-e extra vars have long, if not always, had highest precedence, so this could break playbooks.

You could put the vars you want to include in a separate file and use include_vars although they wouldn't have quite the same precedence as -e extra vars

PieterVoet:
Hi jhawkesworth

to prevent breaking playbooks, I only allow setting an 'extra var' from the playbook if it doesn't already exists !  I think that's safe enough to not break playbooks right ?

Next, this plugin specificly can be used when dealing with dynamic variables, i.e. variables that are discovered runtime, and hence cannot be set upfront using files or vars.
In addition to that, my sample playbook illustrates a playbook where there's no way to expose the variable to other  hosts in the play, since a first play discovers the variable value for a different subset of hosts than the variable is used at.  Therefore also 'set_fact' cannot be used to deal with this.
For me, I really required this in order to do proper playbook orchestration.
Hope this clafifies the need for this plugin.

jhawkesworth:
looks like this has lots of unrelated commits

needs_revision

PieterVoet:
will re-submit in a new pullrequest

PR #19009, opened 2016-12-08
Cloudformation - allow re-using an existing template

adamchainz:
* Allow re-using an existing template when updating a stack by not passing 'template' or 'template_url'. This is a big one for me as our deploy process creates a new stack and then modifies the old one; to avoid changing the resources inside the old one, we have had to avoid using the Ansible module and use the AWS CLI instead in order to pass `--use-previous-template`.
* Split create and update logic into separate functions
* Remove dead `update` variable


adamchainz:
Migrated from https://github.com/ansible/ansible-modules-core/pull/5665

adamchainz:
This was tagged `shipit` if you see the original PR

tedder:
yep. `shipit`.

PR #19015, opened 2016-12-08
unarchive: Various changes/improvements, add debugging output

### LINKED ISSUE #5149: unarchive does not recognize size differs
##### ISSUE TYPE

- Feature Idea
##### COMPONENT NAME

unarchive
##### ANSIBLE VERSION

```
ansible 2.1.1.0
```
##### CONFIGURATION

Nothing unusual or special about configuration, simply using defaults.
##### OS / ENVIRONMENT

CentOS 6.6 32bit
##### SUMMARY

The unarchive task does not consider `Size differs` in the output of gtar in it's is_unarchived method, which perhaps was by initial design or just not thought of. 

As an enhancement it would be nice to add control over if `Size differs` is considered. 

The situation that this would be used for is when you have a custom built archive that needs to replace existing files, for example: configuration or python scripts or static website files/resources.

I did see a ticket on adding a checksum option so perhaps the above situation could be resolved using that feature instead, in which case this is a duplicate or won't fix.

Finally perhaps using the force parameter as a way to always unarchive would also be a solution. This doesn't seem to be the current case, if it is then it should be documented.

Here is a snippet from the unarchive.py is_unarchived method and the output of gtar --diff when the file size differs.

```
[root@niu01 v-ui-crs]# gtar --diff -vf /opt/xpressworkx/app-manager/apps/web/application-files.tar application

application/
application/__init__.py: Size differs
application/configure.py: Mod time differs
application/configure.py: Size differs
application/override.py: Size differs
```

From /usr/lib/python2.6/site-packages/ansible/modules/core/files/unarchive.py 

```
    def is_unarchived(self):
        cmd = '%s -C "%s" -d%s' % (self.cmd_path, self.dest, self.zipflag)
        if self.opts:
            cmd += ' ' + ' '.join(self.opts)
        if self.file_args['owner']:
            cmd += ' --owner="%s"' % self.file_args['owner']
        if self.file_args['group']:
            cmd += ' --group="%s"' % self.file_args['group']
        if self.file_args['mode']:
            cmd += ' --mode="%s"' % self.file_args['mode']
        if self.module.params['keep_newer']:
            cmd += ' --keep-newer-files'
        if self.excludes:
            cmd += ' --exclude="' + '" --exclude="'.join(self.excludes) + '"'
        cmd += ' -f "%s"' % self.src
        rc, out, err = self.module.run_command(cmd)

        ....
        ....

```
##### STEPS TO REPRODUCE

Simply use an unarchive task in a situation where you need to replace existing files of the same name, owner, group, and mode.

```
- name: Unpack Static Files
  unarchive:
    src: "{{ role_path }}/../../web/sdist/static-files.tar"
    dest: "{{ web_config.data['file_root'] }}"
    owner: "{{ web_config.data['web_user'] }}"

```

##### EXPECTED RESULTS

Unarchive the tar into the destination even though files of the same name, owner, group and mode exist.
##### ACTUAL RESULTS

Task returned ok as in already run.

```
TASK [web-ui-common : Unpack Application Static Files] *************************
ok: [192.168.124.152]
```

### LINKED ISSUE #5744: unarchive with zip file: 'owner' and 'group' set files but not directories
##### ISSUE TYPE

 - Bug Report
 
##### COMPONENT NAME

unarchive

##### ANSIBLE VERSION

```
ansible 2.2.0.0
  config file = /home/ubuntu/vtp/ansible/nmm/ansible.cfg
```

##### CONFIGURATION

Nothing significant

```
$ cat /home/ubuntu/vtp/ansible/nmm/ansible.cfg
[defaults]
inventory = ./hosts
roles_path = ../shared/roles
```

##### OS / ENVIRONMENT

Running ansible from Ubuntu 16.04, managing Ubuntu 16.04

##### SUMMARY

unarchive module 'owner' and 'group' set the *file* ownership but not the *directory* ownership

##### STEPS TO REPRODUCE

```
- name: unzip website content
  unarchive:
    src: nmm.zip
    dest: /var/www/html
    owner: www-data
    group: www-data
```

##### EXPECTED RESULTS

That all files and directories in the unpacked filetree would be owned by www-data:www-data

##### ACTUAL RESULTS

Files are owned by www-data:www-data, but directories are owned by root:root

```
root@www:~# ls -l /var/www/html
total 0
drwxr-xr-x 1 root root 76 Nov 27 16:51 NMM 5 days
root@www:~# ls -l /var/www/html/NMM\ 5\ days/
total 8
-rw-rw-rw- 1 www-data www-data 6811 Nov 27 15:37 agenda.md.html
drwxr-xr-x 1 root     root       24 Nov 27 16:51 css
drwxr-xr-x 1 root     root      164 Nov 27 16:51 images
drwxr-xr-x 1 root     root        4 Nov 27 16:44 netmgmt
drwxr-xr-x 1 root     root       30 Nov 27 16:51 software
```

dagwieers:
##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

unarchive

##### ANSIBLE VERSION

v2.2

##### SUMMARY

Various changes and improvements:

- Fixed a few issues with the existing unzip support which were related to Zipfile (which is fundamentally incompatible with unzip excludes/opts).  And issues related to forced ownership.

- Improvements to the UnarchiveError Exception, so it can report more detailed information to why an archive cannot be handled.

- A bugfix relatd to unicode filenames and crc32 handling.

- Renamed a few things so the code is easier to read.

- Ensure gtar considers size changes to unarchive its content. This fixes ansible/ansible-modules-core#5149

- Added a note about ZIP archives without directories. This fixes ansible/ansible-modules-core#5744.

The unarchive module is a hard beast to tame, corner-cases and platform-specific issues make it hard to troubleshoot individual issues.

- Improves the default output (e.g. unzip no longer shows standard output, just like gtar) and adds the ability to get debug-output when using -vvvv (verbosity >= 4) or -vvvvv. So that when issues are i reported we have the means to look what is going on internally.

- Because the output by default is reduced, proper warnings/errors are now more obvious to end-users, e.g.

```
TASK [unarchive]
***************************************************************
task path: /home/dag/home-made/ansible.testing/test126.yml:11
changed: [localhost] => {"changed": true, "dest": "/tmp/test126a", "extract_results": {"cmd":
["/usr/bin/unzip", "-q", "-o", "/home/dag/.ansible/tmp/ansible-tmp-1479989967.66-3171393713112/source",
"-x", "some.directory/", "another.directory/simple.test.file.txt", "-d", "/tmp/test126a"], "err":
"caution: excluded filename not matched: some.directory/\ncaution: excluded filename not matched:
another.directory/simple.test.file.txt\n", "out": "", "rc": 0}, "gid": 500, "group": "dag", "handler":
"ZipArchive", "mode": "0775", "owner": "dag", "secontext": "unconfined_u:object_r:user_tmp_t:s0",
"size": 4096, "src": "/home/dag/.ansible/tmp/ansible-tmp-1479989967.66-3171393713112/source",
"state": "directory", "uid": 500}
```
Whereas before the complete list of deflated files was being shown (when using -v) which was much more than anyone was willing to look at.

*Migrated from https://github.com/ansible/ansible-modules-core/pull/5721*

### DIFF
--- a/lib/ansible/modules/files/unarchive.py
+++ b/lib/ansible/modules/files/unarchive.py
@@ -29,21 +29,21 @@ description:
 options:
   src:
     description:
-      - If C(remote_src=no) (default), local path to archive file to copy to the target server; can be absolute or relative. If C(remote_src=yes), path on the
-        target server to existing archive file to unpack.
-      - If C(remote_src=yes) and C(src) contains C(://), the remote machine will download the file from the URL first. (version_added 2.0). This is only for
-        simple cases, for full download support use the M(get_url) module.
-    required: true
+    - If C(remote_src=no) (default), local path to archive file to copy to the target server; can be absolute or relative.
+    - If C(remote_src=yes), path on the target server to existing archive file to unpack.
+    - If C(remote_src=yes) and C(src) contains C(://), the remote machine will download the file from the URL first. (version_added 2.0).
+      This is only for simple cases, use the M(get_url) module for full download support.
+    required: yes
   dest:
     description:
       - Remote absolute path where the archive should be unpacked.
-    required: true
+    required: yes
   copy:
     description:
-      - If true, the file is copied from local 'master' to the target machine, otherwise, the plugin will look for src archive at the target machine.
+      - If C(yes), the file is copied from local 'master' to the target machine, otherwise, the plugin will look for src archive at the target machine.
       - This option has been deprecated in favor of C(remote_src).
       - This option is mutually exclusive with C(remote_src).
-    type: 'bool'
+    type: bool
     default: 'yes'
   creates:
     description:
@@ -52,29 +52,30 @@ options:
   list_files:
     description:
       - If set to True, return the list of files that are contained in the tarball.
-    type: 'bool'
+    type: bool
     default: 'no'
     version_added: "2.0"
   exclude:
     description:
       - List the directory and file entries that you would like to exclude from the unarchive action.
+    default: []
     version_added: "2.1"
   keep_newer:
     description:
       - Do not replace existing files that are newer than files from the archive.
-    type: 'bool'
+    type: bool
     default: 'no'
     version_added: "2.1"
   extra_opts:
     description:
       - Specify additional options by passing in an array.
-    default: ""
+    default: ''
     version_added: "2.1"
   remote_src:
     description:
       - Set to C(yes) to indicate the archived file is already on the remote system and not local to the Ansible controller.
       - This option is mutually exclusive with C(copy).
-    type: 'bool'
+    type: bool
     default: 'no'
     version_added: "2.2"
   validate_certs:
@@ -82,22 +83,26 @@ options:
       - This only applies if using a https URL as the source of the file.
       - This should only set to C(no) used on personally controlled sites using self-signed certificate.
       - Prior to 2.2 the code worked as if this was set to C(yes).
-    type: 'bool'
+    type: bool
     default: 'yes'
     version_added: "2.2"
-author: Dag Wieers (@dagwieers)
-todo:
-    - Re-implement tar support using native tarfile module.
-    - Re-implement zip support using native zipfile module.
+author:
+- Dag Wieers (@dagwieers)
 notes:
     - Requires C(gtar)/C(unzip) command on target host.
+      (Free)BSD users should verify they use recent versions with current capabilities.
     - Can handle I(.zip) files using C(unzip) as well as I(.tar), I(.tar.gz), I(.tar.bz2) and I(.tar.xz) files using C(gtar).
+    - It does not support compressed files, only compressed archives.
     - Uses gtar's C(--diff) arg to calculate if changed or not. If this C(arg) is not
       supported, it will always unpack the archive.
     - Existing files/directories in the destination which are not in the archive
       are not touched. This is the same behavior as a normal archive extraction.
     - Existing files/directories in the destination which are not in the archive
       are ignored for purposes of deciding if the archive should be unpacked or not.
+    - ZIP files consisting of files, but not directories may find that ownership or
+      permissions of parent directories may not be set as expected when created.
+    - For troubleshooting you can add C(-vvvv) or C(-vvvvv) to the ansible
+      command-line. This will provide more verbose output and logic to the user.
     - For Windows targets, use the M(win_unzip) module instead.
 '''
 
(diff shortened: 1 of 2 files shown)


dagwieers:
Migrated from https://github.com/ansible/ansible-modules-core/pull/5721

mattclay:
CI failure on FreeBSD due to:

```
2017-02-14 12:02:54 TASK [unarchive : unarchive a zip file] ****************************************
2017-02-14 12:02:54 fatal: [testhost]: FAILED! => {"changed": false, "dest": "/root/ansible_testing/test-unarchive-zip", "details": "[Errno 2] No such file or directory: '/root/ansible_testing/test-unarchive-zip/foo-unarchive.txt'", "failed": true, "gid": 0, "group": "wheel", "handler": "ZipArchive", "mode": "0755", "msg": "Unexpected error when accessing exploded file", "owner": "root", "size": 512, "src": "/root/ansible_testing/test-unarchive.zip", "state": "directory", "uid": 0}
```

dagwieers:
No CI, close and re-open.

mattclay:
CI failure on FreeBSD. The error is fairly lengthy, so I won't copy it here.

dagwieers:
The problem is known here, FreeBSD unzip is quite a different beast than all the other OSes. Much like bsdtar vs gtar :-/ The solution is not yet known as it is harder to debug.

calfonso:
@dagwieers could you rebase this and fix up the BSD zip case?

dagwieers:
@calfonso Right, that was the plan. But fixing BSD was much harder than expected. Feel free to create a PR for that one against my branch.

mattclay:
CI failure in integration tests due to traceback:

```
  File "/tmp/ansible_OCWhAc/ansible_module_unarchive.py", line 867, in <module>
    main()
  File "/tmp/ansible_OCWhAc/ansible_module_unarchive.py", line 817, in main
    handler = pick_handler(src, dest, file_args, module)
  File "/tmp/ansible_OCWhAc/ansible_module_unarchive.py", line 739, in pick_handler
    (can_handle, info) = obj.can_handle_archive()
  File "/tmp/ansible_OCWhAc/ansible_module_unarchive.py", line 558, in can_handle_archive
    e = get_exception()
NameError: global name 'get_exception' is not defined
```

mattclay:
CI failure in FreeBSD integration tests:

https://app.shippable.com/github/ansible/ansible/runs/32038/8/tests

dagwieers:
Yup, we know. Still the same thing: https://github.com/ansible/ansible/pull/19015#issuecomment-281199523

dagwieers:
So I am very happy to drop FeeBSD support altogether. If they want to have a freaky unzip with different behavior, this module is not for them.

Maybe it helps to get the pure-python implementation finally going.

jborean93:
Closing due to the age of the PR.

PR #19018, opened 2016-12-08
Performance improvement using in-operator on dicts

dagwieers:
##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

facts

##### ANSIBLE VERSION

v2.2

##### SUMMARY

Just a small cleanup for the existing occurences.

Using the in-operator for hash lookups is faster than using .keys()
http://stackoverflow.com/questions/29314269/why-do-key-in-dict-and-key-in-dict-keys-have-the-same-output


dagwieers:
This was part of a series #18525, ansible/ansible-modules-core#5650 and ansible/ansible-modules-extras#3496

dagwieers:
The Shippable error is related to problems with the docker test infrastructure.

PR #19019, opened 2016-12-08
Initial value causing exception when vlan_range param is present

ebeahan:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME
modules/network/nxos/nxos_vlan

##### ANSIBLE VERSION

```
ansible 2.2.0.0
  config file =
```

##### SUMMARY

This PR addresses an issue where a variable is initially set to `None`, but the value of the variable is never updated if a `vlan_range` parameter is specified. This results in a `AttributeError` exception when the `.get()` method is called against a `NoneType` object.


privateip:
Thanks, merged into devel

PR #19021, opened 2016-12-08
new AWS module for ec2 VPC vgw facts

naslanidis:
##### ISSUE TYPE
- New Module Pull Request
##### COMPONENT NAME

**Recreated PR due to old one being deleted
https://groups.google.com/forum/#!topic/ansible-devel/0eRSpiGkios

ec2_vpc_vgw_facts
##### ANSIBLE VERSION

```
ansible 2.3.0
```
##### SUMMARY

This module returns virtual gateway facts for an AWS account.

```
TASK [print output] ************************************************************
ok: [localhost] => {
    "msg": {
        "changed": false, 
        "virtual_gateways": [
            {
                "state": "available", 
                "tags": [
                    {
                        "key": "Name", 
                        "value": "TEST-VGW"
                    }
                ], 
                "type": "ipsec.1", 
                "vpc_attachments": [
                    {
                        "state": "attached", 
                        "vpc_id": "vpc-22a93c74"
                    }
                ], 
                "vpn_gateway_id": "vgw-42e3d64e"
            }
        ]
    }
}
```

### DIFF
--- a/lib/ansible/modules/cloud/amazon/ec2_vpc_vgw_facts.py
+++ b/lib/ansible/modules/cloud/amazon/ec2_vpc_vgw_facts.py
@@ -1,0 +1,169 @@
+#!/usr/bin/python
+#
+# This is a free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This Ansible library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this library.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: ec2_vpc_vgw_facts
+short_description: Gather facts about virtual gateways in AWS
+description:
+    - Gather facts about virtual gateways in AWS
+version_added: "2.3"
+requirements: [ boto3 ]
+options:
+  filters:
+    description:
+      - A dict of filters to apply. Each dict item consists of a filter key and a filter value. See U(http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeRouteTables.html) for possible filters.
+    required: false
+    default: None
+  VpnGatewayIds:
+    description:
+      - Get details of specific Virtual Gateway ID
+      - Provide this value as a list
+    required: false
+    default: None
+author: "Nick Aslanidis (@naslanidis)"
+extends_documentation_fragment:
+    - aws
+    - ec2
+'''
+
+EXAMPLES = '''
+# # Note: These examples do not set authentication details, see the AWS Guide for details.
+
+- name: Gather facts about all virtual gateways for an account or profile
+  ec2_vpc_vgw_facts:
+    region: ap-southeast-2
+    profile: production
+  register: vgw_facts
+
+- name: Gather facts about a filtered list of Virtual Gateways
+  ec2_vpc_vgw_facts:
+    region: ap-southeast-2
+    profile: production
+    filters:
+        "tag:Name": "vgw-123"
+  register: vgw_facts
+
+- name: Gather facts about a specific virtual gateway by VpnGatewayIds
+  ec2_vpc_vgw_facts:
+    region: ap-southeast-2
+    profile: production
+    VpnGatewayIds: vgw-c432f6a7 
+  register: vgw_facts
+'''
+
+RETURN = '''
+virtual_gateways:
+    description: The virtual gateways for the account
+    returned: always
+    type: list
+
+changed:
+    description: True if listing the virtual gateways succeeds
+    type: bool
+    returned: always
+'''
+
+import json
+
+try:
+    import botocore
+    import boto3   
+    HAS_BOTO3 = True
+except ImportError:
+    HAS_BOTO3 = False
+
+
+def get_virtual_gateway_info(virtual_gateway):
+    virtual_gateway_info = {'VpnGatewayId': virtual_gateway['VpnGatewayId'],
+                             'State': virtual_gateway['State'],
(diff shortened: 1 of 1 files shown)


lib/ansible/modules/cloud/amazon/ec2_vpc_vgw_facts.py
```diff
@@ -0,0 +1,169 @@
...
+    profile: production
+    VpnGatewayIds: vgw-c432f6a7 
+  register: vgw_facts
+'''
+
+RETURN = '''
+virtual_gateways:
+    description: The virtual gateways for the account
+    returned: always
+    type: list
```
ryansb, line 77: Can you please include an example of the return values of this module? Is it a list of the names?

lib/ansible/modules/cloud/amazon/ec2_vpc_vgw_facts.py
```diff
@@ -0,0 +1,169 @@
...
+            params['Filters'].append(temp_dict)
+
+    if module.params.get("DryRun"):
+        params['DryRun'] = module.params.get("DryRun")
+
+    if module.params.get("VpnGatewayIds"):
+        params['VpnGatewayIds'] = module.params.get("VpnGatewayIds")
+
+    try:
+        all_virtual_gateways = client.describe_vpn_gateways(**params)
```
ryansb, line 130: Would it be possible for you to use the AWSRetry decorator here, so that instead of failing there will be a little bit of tolerance to the API being throttled/unavailable?

lib/ansible/modules/cloud/amazon/ec2_vpc_vgw_facts.py
```diff
@@ -0,0 +1,169 @@
...
+            DryRun = dict(type='bool', default=False),
+            VpnGatewayIds = dict(type='list', default=None)
+        )
+    )
+
+    module = AnsibleModule(argument_spec=argument_spec)
+
+     # Validate Requirements
+    if not HAS_BOTO3:
+        module.fail_json(msg='json and botocore/boto3 is required.')
```
ryansb: Just say `boto3` here, since botocore is a transitive dependency of boto3.

ryansb (changes requested):

lib/ansible/modules/cloud/amazon/ec2_vpc_vgw_facts.py
```diff
@@ -0,0 +1,169 @@
...
+requirements: [ boto3 ]
+options:
+  filters:
+    description:
+      - A dict of filters to apply. Each dict item consists of a filter key and a filter value. See U(http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeRouteTables.html) for possible filters.
+    required: false
+    default: None
+  VpnGatewayIds:
+    description:
+      - Get details of specific Virtual Gateway ID
```
gundalow: `descriptions:` should be sentences, eg capital letters and full stops

gundalow (changes requested):
needs_revision

abadger:
needs_revision

abadger:
@naslanidis any update to this to take care of the requested changes?

lib/ansible/modules/cloud/amazon/ec2_vpc_vgw_facts.py
naslanidis: Hi @ryansb 

I'd love to implement this but are there any other modules or examples I can use to assist with this? If there's other AWS modules that you know of with this implemented that would be an excellent reference.

naslanidis:
@abadger I've made the requested changes with the exception of the AWSRetry decorator as I'm not sure how to implement it. I will look at that in more detail though.

ryansb (dismissed):

lib/ansible/modules/cloud/amazon/ec2_vpc_vgw_facts.py
ryansb: Yes, the CloudFormation module uses AWSRetry, see [here](https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/cloud/amazon/cloudformation.py#L364-L366) for an example. Just import the retry, and then use that decorator to do the backoff/retry.

gundalow (approved):
shipit

lib/ansible/modules/cloud/amazon/ec2_vpc_vgw_facts.py
```diff
@@ -0,0 +1,168 @@
...
+    region: ap-southeast-2
+    profile: production
+  register: vgw_facts
+
+- name: Gather facts about a filtered list of Virtual Gateways
+  ec2_vpc_vgw_facts:
+    region: ap-southeast-2
+    profile: production
+    filters:
+        "tag:Name": "vgw-123"
```
ryansb: Can you use a different example here? It's easy to confuse this with a dummy VGW ID. Maybe "main-virt-gateway" would be a better docs slug.

lib/ansible/modules/cloud/amazon/ec2_vpc_vgw_facts.py
```diff
@@ -0,0 +1,168 @@
...
+    params['Filters'] = ansible_dict_to_boto3_filter_list(module.params.get('filters'))
+    params['DryRun'] = module.check_mode
+
+    if module.params.get("vpn_gateway_ids"):
+        params['VpnGatewayIds'] = module.params.get("vpn_gateway_ids")
+
+    try:
+        all_virtual_gateways = client.describe_vpn_gateways(**params)
+    except botocore.exceptions.ClientError as e:
+        module.fail_json(msg=str(e))
```
ryansb: Can you add the exception path here, with `fail_json(msg=...., exception=traceback.format_ext())`

ryansb (changes requested):

lib/ansible/modules/cloud/amazon/ec2_vpc_vgw_facts.py
abadger: notet: ```exception=traceback.format_exc()```

ryansb (approved):

ryansb:
Thanks for your contribution, merged!

PR #19024, opened 2016-12-08
boto3-powered s3 synchronization module

tedder:
##### ISSUE TYPE

 - New Module Pull Request

##### COMPONENT NAME
`s3_sync`

##### ANSIBLE VERSION

```
ansible 2.2.0 (devel 1f03801fcc) last updated 2016/09/29 14:29:35 (GMT -700)
  lib/ansible/modules/core: (devel b16e2597a4) last updated 2016/10/22 19:33:36 (GMT -700)
  lib/ansible/modules/extras: (devel 6c31d91fa5) last updated 2016/10/11 16:10:16 (GMT -700)
  config file = 
```

##### SUMMARY

rsync for s3 files, basically. Designed to push largeish numbers of files to s3 in an efficient manner. The `s3` module does a poor job of this, due to Ansible design decisions.

```

```

### DIFF
--- a/lib/ansible/modules/cloud/amazon/s3_sync.py
+++ b/lib/ansible/modules/cloud/amazon/s3_sync.py
@@ -1,0 +1,492 @@
+#!/usr/bin/python
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: s3_sync
+short_description: Efficiently upload multiple files to S3
+description:
+     - The S3 module is great, but it is very slow for a large volume of files- even a dozen will be noticeable. In addition to speed, it handles globbing, inclusions/exclusions, mime types, expiration mapping, recursion, and smart directory mapping.
+version_added: "2.3"
+options:
+  mode:
+    description:
+    - sync direction.
+    required: true
+    default: 'push'
+    choices: [ push ]
+  file_change_strategy:
+    description:
+    - Difference determination method to allow changes-only syncing. Unlike rsync, files are not patched- they are fully skipped or fully uploaded.
+    - date_size will upload if file sizes don't match or if local file modified date is newer than s3's version
+    - checksum will compare etag values based on s3's implementation of chunked md5s.
+    - force will always upload all files.
+    required: false
+    default: 'date_size'
+    choices: [ force, checksum, date_size ]
+  bucket:
+    description:
+    - Bucket name.
+    required: true
+  key_prefix:
+    description:
+    - In addition to file path, prepend s3 path with this prefix. Module will add slash at end of prefix if necessary.
+    required: false
+  file_root:
+    description:
+    - File/directory path for synchronization. This is a local path.
+    - This root path is scrubbed from the key name, so subdirectories will remain as keys.
+    required: true
+  permission:
+    description:
+    - Canned ACL to apply to synced files.
+    - Changing this ACL only changes newly synced files, it does not trigger a full reupload.
+    required: false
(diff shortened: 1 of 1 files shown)


tedder:
Migrated from https://github.com/ansible/ansible-modules-extras/pull/3326

lib/ansible/modules/cloud/amazon/s3_sync.py
```diff
@@ -0,0 +1,492 @@
...
+        error = err.error_message
+    elif hasattr(err, 'message'):
+        error = str(err.message) + ' ' + str(err) + ' - ' + str(type(err))
+    else:
+        error = '%s: %s' % (Exception, err)
+
+    return error
+
+# the following function, calculate_multipart_etag, is from tlastowka
+# on github and is used under its (compatible) MIT license. So this
```
ryansb: This looks like a GPL license declaration, are you sure it's MIT?

lib/ansible/modules/cloud/amazon/s3_sync.py
```diff
@@ -0,0 +1,492 @@
...
+    keeplist = list(s3filelist)
+
+    for e in keeplist:
+      e['_strategy'] = strategy
+
+    # fetch info from S3 if we're going to use it for comparisons
+    if not strategy == 'force':
+        keeplist = head_s3(s3, bucket, s3filelist)
+
+    if strategy == 'checksum':
```
ryansb, line 382: Can you make this an `elif` for consistency?

lib/ansible/modules/cloud/amazon/s3_sync.py
```diff
@@ -0,0 +1,492 @@
...
+    # else: probably 'force'. Basically we don't skip with any with other strategies.
+    else:
+        pass
+
+    # prune 'please skip' entries, if any.
+    return [x for x in keeplist if not x.get('skip_flag')]
+
+def upload_files(s3, bucket, filelist, params):
+    ret = []
+    for entry in filelist:
```
ryansb, line 425: It looks like this is just a serial upload, how is this faster than the current S3 module? I definitely see the benefit of the glob & sync strategies, but it seems like this would be just as fast.

lib/ansible/modules/cloud/amazon/s3_sync.py
tedder: my bad. fixing.

lib/ansible/modules/cloud/amazon/s3_sync.py
tedder: it's not meant to be an `elif`; I'm updating comments, though. The `not` is basically a common section so it isn't duplicated in the `if/elif/elif` strategy section.

lib/ansible/modules/cloud/amazon/s3_sync.py
tedder: The thing is that s3.py is really slow when using `with_` globbing. I'll give a performance comparison.

tedder: local test: 14sec vs 35sec.

abadger:
Per today's meeting, ryansb to do some testing.

ryansb:
In testing this I think I found an error with the date/size check type. Using this task:

```
- s3_sync: 
    profile: testing
    bucket: ryansb-ansible-scratch
    file_change_strategy: date_size
    file_root: /tmp/testdir
```

If I add lines to one of the files, the task continues to come back as `state=ok` and the updated version of the file is never uploaded. If I change to `file_change_strategy: checksum` the task changes and the new version of the file gets uploaded. It seems like a modified mtime and a different size should mean the new version is uploaded by the `date_size` strategy. Is there something I'm missing?

lib/ansible/modules/cloud/amazon/s3_sync.py
```diff
@@ -0,0 +1,493 @@
...
+                remote_modified_datetime = entry['s3_head']['LastModified']
+                delta = (remote_modified_datetime - datetime.datetime(1970, 1, 1, tzinfo=tz.tzutc()))
+                remote_modified_epoch = delta.seconds + (delta.days*86400)
+
+                remote_size = entry['s3_head']['ContentLength']
+
+                entry['whytime'] = '{} / {}'.format(local_modified_epoch, remote_modified_epoch)
+                entry['whysize'] = '{} / {}'.format(local_size, remote_size)
+
+                if local_modified_epoch <= remote_modified_epoch or local_size != remote_size:
```
ryansb: Ah, I think this ought to be `local_size == remote_size`, since this conditional causes the file to be skipped for this strategy. That explains the odd behavior I saw with `date_size`

ryansb (changes requested):

tedder:
Thanks @ryansb - good call. The logic was flipped, it was correct for 'include' but not for 'skip'. Duh.

ryansb (approved):

ryansb:
Squashed as be7a568, and I added the ANSIBLE_METADATA to denote community support.

PR #19025, opened 2016-12-08
systemd: Add boolean option to enable --no-block

towolf:
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME

systemd

##### SUMMARY

In our environment we have custom services that need to be stopped and
restarted very gracefully to not interrupt active sessions.

A stop job, depending on the state, can take up to 20 minutes until the
process exits. It simply reacts to SIGTERM with a graceful shutdown.

By default, systemctl blocks until the job has completed, which leads to
Ansible hanging on this task for up to 20 minutes.

Thankfully systemctl supports the `--no-block` flag which lets the job
continue in the background.

This PR adds support for that flag as the `no_block` boolean option.

From the man page:

     --no-block
       Do not synchronously wait for the requested operation to
       finish. If this is not specified, the job will be
       verified, enqueued and systemctl will wait until the
       unit's start-up is completed. By passing this argument,
       it is only verified and enqueued. This option may not be
       combined with --wait.

### DIFF
--- a/lib/ansible/modules/system/systemd.py
+++ b/lib/ansible/modules/system/systemd.py
@@ -68,6 +68,14 @@ options:
         description:
             - run systemctl talking to the service manager of the calling user, rather than the service manager
               of the system.
+    no_block:
+        required: false
+        default: no
+        choices: [ "yes", "no" ]
+        description:
+            - Do not synchronously wait for the requested operation to finish.
+              Enqueued job will continue without Ansible blocking on its completion.
+        version_added: "2.3"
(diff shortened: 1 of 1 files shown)


towolf:
Migrated from https://github.com/ansible/ansible-modules-core/pull/5507

bcoca:
why not just use `async: true` on the task?

towolf:
Because I want the exit code of the systemctl call immediately, not after 20 minutes.

PR #19026, opened 2016-12-09
Add Yarn module

verkaufer:
##### ISSUE TYPE

- New Module Pull Request
##### COMPONENT NAME

Yarn
##### ANSIBLE VERSION

```
ansible 2.0.1.0
  config file = 
```
##### SUMMARY

Adds [Yarn](https://yarnpkg.com/) Node.js package manager support to Ansible. Because Yarn is very similar to NPM, the module draws heavily from Chris Hoffman's `npm` Ansible module. The output of Yarn was omitted because it makes heavy use of Emojis which do not translate into a clear JSON format.


verkaufer:
Migrated from https://github.com/ansible/ansible-modules-extras/pull/3199

rightaway:
Any update on whether this can be merged to master yet? I've tried it and it works great.

thoiberg:
shipit

lib/ansible/modules/packaging/language/yarn.py
```diff
@@ -0,0 +1,316 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: yarn
+short_description: Manage node.js packages with Yarn
+description:
+  - Manage node.js packages with Yarn (https://yarnpkg.com/)
+version_added: "2.3"
```
maxamillion: If you're able to get this change in by today, please bump this to `2.5`. Otherwise please bump to `2.6` and I'll merge it in the next release window.

lib/ansible/modules/packaging/language/yarn.py
```diff
@@ -0,0 +1,316 @@
...
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
```
maxamillion: Please add an `ANSIBLE_METADATA` as per https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html

lib/ansible/modules/packaging/language/yarn.py
```diff
@@ -0,0 +1,316 @@
...
+            out = yarn.update()
+    else: #absent
+        installed, missing = yarn.list()
+        if name in installed:
+            changed = True
+            out = yarn.uninstall()
+
+    module.exit_json(changed=changed, out=out)
+
+# import module snippets
```
maxamillion: We recommend not `import *` and stylistically prefer `from ansible.module_utils.basic import AnsibleModule` to come right after the global docstrings.

maxamillion (changes requested):

maxamillion:
needs_revision

verkaufer:
I will make this changes this week and we can target it for `2.6`!

verkaufer:
@maxamillion Ready to roll.

maxamillion:
This is great, thank you! It would be great if you would please add some [integration tests](https://docs.ansible.com/ansible/latest/dev_guide/testing_integration.html). We're attempting to continue to increase our test coverage (which will be a big focus in the Ansible 2.6 development cycle), so if you could add them for your new module that would be greatly appreciated.

maxamillion (approved):

verkaufer:
Sure thing! I will look into adding a few. If they don't make it into this PR, I'll open a new one with the tests.

lib/ansible/modules/packaging/language/yarn.py
```diff
@@ -0,0 +1,335 @@
...
+import os
+import re
+
+from ansible.module_utils.basic import AnsibleModule
+
+try:
+    import json
+except ImportError:
+    try:
+        import simplejson as json
```
webknjaz: json module is in stdlib since Python 2.6, which is lowest version supported: https://github.com/ansible/ansible/blob/devel/setup.py#L206

So there's no need to do fallback.

verkaufer: Is this a blocker for PR approval?

astorije:
Hey there! What are the chances this would make it into `master`? 🙏 Thanks!

lib/ansible/modules/packaging/language/yarn.py
webknjaz: no, but it's nice to avoid having "dead code" :)

maxamillion:
The code looks good. If you add integration tests much like the [npm module](https://github.com/ansible/ansible/tree/devel/test/integration/targets/npm), I'm happy to merge.

Thanks!

verkaufer:
I'll get to adding integration tests this weekend :)

verkaufer:
Update: Still working on integration tests. Hit a snag with some changes coming from Yarn itself and need to update some handling. Will be a few more days :)

maxamillion:
@verkaufer no worries, we still have time for the 2.6 development cycle. Thanks for the update!

verkaufer:
@maxamillion Should be all set now! Give it a look and we can square this away for 2.6

test/integration/targets/yarn/tasks/run.yml
```diff
@@ -0,0 +1,114 @@
...
+  block:
+
+    # Get the version of Yarn and register to a variable
+    - shell: '{{ yarn_bin_path }}/yarn --version'
+      environment:
+        PATH: '{{ node_bin_path }}:{{ ansible_env.PATH }}'
+      register: yarn_version
+
+    - name: 'Create dummy package.json'
+      template:
```
webknjaz: @verkaufer you need copy module, not a template as you don't interpolate any vars there.

test/integration/targets/yarn/tasks/run.yml
```diff
@@ -43,8 +43,8 @@
       register: yarn_version
 
     - name: 'Create dummy package.json'
-      template:
-        src: package.j2
+      copy:
+        src: templates/package.j2
```
webknjaz: It's better to put static files into files folder instead of templates and use normal file extension instead of `.j2`, since it's not a template.

maxamillion: @verkaufer if you don't mind filing a follow-up PR to fix that, I'd appreciate it. Thanks :)

PR #19027, opened 2016-12-09
Add module "launchd" to manage Mac OS X services.

bjoernalbers:
##### ISSUE TYPE

New Module Pull Request
##### COMPONENT NAME

osx_service
##### ANSIBLE VERSION

```
ansible 2.1.1.0
  config file = 
```
##### SUMMARY

This module allows to start, stop and restart Mac OS X launchd daemons
via `launchctl`.
It supports system services (stuff that has a plist under
/Library/LaunchDaemons) and must be run with root permissions.

About the implementation...

Service management on Mac OS X is a different horse and deserves its own module.
I build the first draft in Ruby: [osx_service.rb](https://gist.github.com/bjoernalbers/f40b3fc17067959b9e4663be5021c1be).
But with the help of `AnsibleModule` I was able to port it to Python so that others could use this as well.

Here is some example output...

```
TASK [stop crashplan] **********************************************************
changed: [localhost] => {"changed": true, "name": "com.crashplan.engine", "state": "stopped"}
```

Thanks in advance,
Björn

### DIFF
--- a/lib/ansible/modules/system/launchd.py
+++ b/lib/ansible/modules/system/launchd.py
@@ -27,7 +27,7 @@ author:
     - 'Ansible Core Team'
     - 'Theo Crevon (@tcr-ableton)'
     - 'Björn Albers <bjoernalbers@gmail.com>'
-version_added: '2.2'
+version_added: '2.3'
 short_description: Manage Mac OS X launchd services.
 description:
     - Manage (start, stop and restart) Mac OS X launchd services via launchctl.
--- a/lib/ansible/modules/system/launchd.py
+++ b/lib/ansible/modules/system/launchd.py
@@ -30,7 +30,7 @@ author:
 version_added: '2.2'
 short_description: Manage Mac OS X launchd services.
 description:
-    - Manage (start, stop & restart) Mac OS X launchd services via launchctl.
+    - Manage (start, stop and restart) Mac OS X launchd services via launchctl.
 options:
     name:
         required: true
@@ -39,7 +39,7 @@ options:
         aliases: [ 'service', 'label' ]
     state:
         required: true
-        choices: [ 'started', 'loaded', 'stopped', 'unloaded', 'restarted', 'reloaded ]
+        choices: [ 'started', 'loaded', 'stopped', 'unloaded', 'restarted', 'reloaded' ]
         description:
             - C(started), C(loaded), C(stopped) and C(unloaded) are idempotent
               actions that will not run commands unless necessary.
--- a/lib/ansible/modules/system/launchd.py
+++ b/lib/ansible/modules/system/launchd.py
@@ -160,4 +160,5 @@ def main():
 
 # import module snippets
 from ansible.module_utils.basic import *
-main()
+if __name__ == '__main__':
+    main()
--- a/lib/ansible/modules/system/launchd.py
+++ b/lib/ansible/modules/system/launchd.py
@@ -1,6 +1,8 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
-
+#
+# (c) 2016, Brian Coca <bcoca@ansible.com>
+# (c) 2106, Theo Crevon (https://github.com/tcr-ableton)
 # (c) 2016, Björn Albers <bjoernalbers@gmail.com>
 #
 # This file is part of Ansible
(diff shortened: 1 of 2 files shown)


bjoernalbers:
Migrated from https://github.com/ansible/ansible-modules-extras/pull/3073

lib/ansible/modules/system/launchd.py
```diff
@@ -0,0 +1,163 @@
...
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+
+DOCUMENTATION = '''
+module: launchd
+author:
+    - 'Ansible Core Team'
+    - 'Theo Crevon (@tcr-ableton)'
+    - 'Björn Albers <bjoernalbers@gmail.com>'
+version_added: '2.2'
```
gundalow: 2.3

lib/ansible/modules/system/launchd.py
```diff
@@ -0,0 +1,163 @@
...
+        changed = service.stop()
+    elif state in [ 'restarted', 'reloaded' ]:
+        changed = service.restart()
+
+    module.exit_json(name=name, state=state, changed=changed) 
+
+
+# import module snippets
+from ansible.module_utils.basic import *
+main()
```
gundalow: To enable the debugger to work, please rewrite as
```
if __name__ == '__main__':
    main()
```

gundalow (changes requested):
needs_revision just a few minor comments.

gundalow:
@bcoca from the original PR could you please review,  thanks.

abadger:
At the meeting today, bcoca raised the concern that this might not work with all versions of MacOSX.  So we settled on testing and documentation as the ways to address that.  So if there's a known limit to how recent OSX must be you should put that into the module docs.  bcoca also mentioned his PR: https://github.com/ansible/ansible-modules-core/pull/3694/files as using interfaces that were documented to work on older OSX.

Is there a separate PR for integration tests?  Now that we have modules and tests in the same repo, putting the tests in this PR will let us see that they pass before merging.  If there's not tests, could some be written?  We do have MacOSX in our continuous integration framework: https://app.shippable.com/projects/573f79d02a8192902e20e34b/status/  Take a look at one of the devel runs and see one of hte jobs in environment TEST=osx/10.11 

(If you need help writing an integration test, take a look at how we test some of the other modules..  Probably best to start with something simple like the ping test: https://github.com/ansible/ansible/blob/devel/test/integration/targets/ping/tasks/main.yml
Then add a conditional so the launchd tests are only run on macOSX.)

lib/ansible/modules/system/launchd.py
```diff
@@ -0,0 +1,164 @@
...
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+
```
abadger, line 23: We'll need to add ANSIBLE_METADATA now.  (there will soon be a test that this exists for every module we ship).  The metadata is prety minimal right now so this should be fine:
```
ANSIBLE_METADATA = {'status': ['preview'],
                    'supported_by': 'community',
                    'version': '1.0'}
```

lib/ansible/modules/system/launchd.py
```diff
@@ -0,0 +1,164 @@
...
+    state:
+        required: true
+        choices: [ 'started', 'loaded', 'stopped', 'unloaded', 'restarted', 'reloaded' ]
+        description:
+            - C(started), C(loaded), C(stopped) and C(unloaded) are idempotent
+              actions that will not run commands unless necessary.
+              Launchd does not support C(restarted) / C(reloaded) natively, so
+              these will both trigger a stop and start as needed.
+requirements:
+    - target host OS must be Mac OS X
```
abadger, line 49: Add minimum version if needed.

lib/ansible/modules/system/launchd.py
```diff
@@ -0,0 +1,164 @@
...
+    - you have to run the module as root via sudo
```
abadger, line 50: Does it really require sudo?  Or just root?  (for instance ssh'ing in as root or using a different program to escalate privileges)

abadger: ...Or being root and running ansible locally.  (bcoca suggests the wording: "root privileges are required"

lib/ansible/modules/system/launchd.py
```diff
@@ -0,0 +1,164 @@
...
+      name: com.docker.vmnetd
+      state: started
+      become: yes
+      become_user: root
+'''
+
+
+import os
+
+class OSXService:
```
abadger, line 83: As long as we're supporting python2 we should make this a new-style class like this:
```
class OSXService(object):
```

lib/ansible/modules/system/launchd.py
```diff
@@ -0,0 +1,164 @@
...
+        if self.__started():
+            return self.__launchctl(['unload', '-w', self.__plist()])
+        else:
+            return False
+
+    def restart(self):
+        self.stop()
+        return self.start()
+
+    def __started(self):
```
abadger, line 107: Only use a single leading underscore for private methods.  Typical python style is about convention rather than forcing (and the name obfuscation is just obfuscation anyway.... it's still possible to call a private method if you convert the name yourself).

lib/ansible/modules/system/launchd.py
```diff
@@ -0,0 +1,164 @@
...
+        (rc, _, _) = self.module.run_command(
```
abadger, line 108: We try not to use "_" as an identifier just in case we want to internationalize the modules someday (the convention for gettext is to mark strings using the function _().  So we don't want to inadvertantly override _ with another identifier).

lib/ansible/modules/system/launchd.py
```diff
@@ -0,0 +1,164 @@
...
+    elif state in [ 'stopped', 'unloaded' ]:
+        changed = service.stop()
+    elif state in [ 'restarted', 'reloaded' ]:
+        changed = service.restart()
+
+    module.exit_json(name=name, state=state, changed=changed) 
+
+
+# import module snippets
+from ansible.module_utils.basic import *
```
abadger, line 162: From Ansible 2.1 onwards (since this is a new module, it will only be in 2.3 and above) we can put this near the top, with the other imports.  We can also stop using the wildcard import.  So this would be between "import os" and "class OSXService" and read like this:

``` python
```

abadger:
Left comments -- they're largely style and convention for our project.

Like bcoca I also have a question about why you're omitting enabled/disabled and just what "-w" means.

Start and stop mean to make the service run or stop it if it is running during the current session (for system services, that probably means until the next reboot).

Enabled and disabled mean to set the service to start when a new session starts (meaning that when we reboot the machine, the service will be started).

From the documentation you quoted, it seems like -w is needed to override whatever is specified persistently to implement start and stop.  if that's correct, that part seems fine.

But this PR seems to ignore enabled and disabled entirely.  It's not giving the user the ability to set the service to start on boot or not.  This is a feature that needs to be added to have parity with our other service-style modules.  Looks like bcoca's PR had to resort to parsing .plists in order to do that since he wasn't able to find a launchd command that modified the RunAtLoad attribute.

abadger:
needs_revision

abadger:
@bjoernalber any updates/further questions/discussion on this?

bjoernalbers:
@abadger Hello Toshio,

thanks for having a look at this and your feedback.

I'm totally fine with your suggested style changes and the integration tests.
But I won't introduce hacks just to enforce parity with the other service-modules.
macOS *is* different from linux in many ways and in my opinion Ansible should honour this - after all that is Ansible's selling point / advantage over tools like Chef and Puppet, right?
So, enable / start will always persist the running state in my module and thus survive a reboot.

Please let me now if you can live with this so I can incorporate the other changes.
And if not that just close this PR.
I won't be angry, seriously :-)


Björn

abadger:
@bjoernalbers   Hey Björn,

I've been talking with bcoca in slack.  He's of the opinion that launchd supports these concepts even if our implementation would have to be a bit of a hack.  So I think that we'd need those changes to merge this.

You can submit this to galaxy inside of a role if you'd still like to put it out there for other people to use, though. Take a look at:  http://docs.ansible.com/ansible/playbooks_roles.html#embedding-modules-and-plugins-in-roles  for the basic concepts of putting modules inside of roles and https://galaxy.ansible.com/intro#share   for information on how to get the role onto galaxy itself.

Thanks for working on this and making the ecosystem as a whole better :-)

bjoernalbers:
@abadger Thanks for the tip!

PR #19032, opened 2016-12-09
Allow kerberos_auth in URL requests

willthames:
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
module_utils/urls

##### ANSIBLE VERSION

```
ansible 2.3.0 (kerberos_auth_uri a6c870d62c) last updated 2016/12/09 13:22:30 (GMT +1000)
  config file = 
```

##### SUMMARY

Enable kerberos authentication for URLs.

Makes use of very recent urllib_kerberos module available on pypi.

### DIFF
--- a/lib/ansible/module_utils/urls.py
+++ b/lib/ansible/module_utils/urls.py
@@ -82,6 +82,13 @@ try:
 except ImportError:
     HAS_SSLCONTEXT = False
 
+try:
+    import urllib_kerberos
+    import kerberos
+    HAS_KERBEROS_AUTH = True
+except ImportError:
+    HAS_KERBEROS_AUTH = False
+
 # SNI Handling for python < 2.7.9 with urllib3 support
 try:
     # urllib3>=1.15
@@ -756,7 +763,7 @@ def open_url(url, data=None, headers=None, method=None, use_proxy=True,
              force=False, last_mod_time=None, timeout=10, validate_certs=True,
              url_username=None, url_password=None, http_agent=None,
              force_basic_auth=False, follow_redirects='urllib2',
-             client_cert=None, client_key=None, cookies=None):
+             client_cert=None, client_key=None, cookies=None, kerberos_auth=False):
     '''
     Sends a request via HTTP(S) or FTP using urllib2 (Python2) or urllib (Python3)
 
@@ -811,6 +818,9 @@ def open_url(url, data=None, headers=None, method=None, use_proxy=True,
 
         elif username and force_basic_auth:
             headers["Authorization"] = basic_auth_header(username, password)
+
+        elif kerberos_auth:
+            handlers.append(urllib_kerberos.HTTPKerberosAuthHandler())
 
         else:
             try:
@@ -923,6 +933,7 @@ def url_argument_spec():
         force_basic_auth=dict(required=False, type='bool', default='no'),
         client_cert=dict(required=False, type='path', default=None),
         client_key=dict(required=False, type='path', default=None),
+        kerberos_auth=dict(type='bool', default='no'),
     )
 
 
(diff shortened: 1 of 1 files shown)


calfonso:
@willthames could you rebase and resolve any conflicts? Then we'll finish up the review, thanks!

bgsanthosh:
Any update on this PR ? I am looking for similar feature for one of my task -- is there a workaround for this  other than developing new module ?

calfonso:
@bgsanthosh waiting on the contributor to fixed the conflicts.

willthames:
Rebased, looks good but untested.

lib/ansible/module_utils/urls.py
```diff
@@ -812,6 +819,9 @@ def open_url(url, data=None, headers=None, method=None, use_proxy=True,
         elif username and force_basic_auth:
             headers["Authorization"] = basic_auth_header(username, password)
 
+        elif kerberos_auth:
+            handlers.append(urllib_kerberos.HTTPKerberosAuthHandler())
```
puiterwijk, line 823: This looks really opinionated to me, not allowing specifying the Service Principal Name (neither service, hardcoded to "HTTP" or hostname), keytab or ccache location.
Especially given that this runs on the ansible target, I highly doubt most servers would have a current credential in their credential cache at any point in time.
So basically, this code would require someone to have a manual `kinit` play before *any* play that uses `kerberos_auth=yes`.

lib/ansible/module_utils/urls.py
```diff
@@ -82,6 +82,13 @@
 except ImportError:
     HAS_SSLCONTEXT = False
 
+try:
+    import urllib_kerberos
```
puiterwijk, line 86: I see a few problems with using this library:
- If you have `urllib_kerberos` installed but not `kerberos` (which is allowed since `setup.py` does not require it), any run of `ansible` with this patch in that situation will return SystemExit (https://github.com/willthames/urllib_kerberos/blob/fdd0d2954905aa82ff9940d350c0ee8787d379b3/lib/urllib_kerberos/__init__.py#L39)
- It does not actually have a way to enforce mutual authentication, which is marked as TODO and should be configurable, defaulting to enforced (https://github.com/willthames/urllib_kerberos/blob/fdd0d2954905aa82ff9940d350c0ee8787d379b3/lib/urllib_kerberos/__init__.py#L127)

lib/ansible/module_utils/urls.py
```diff
@@ -971,6 +982,9 @@ def fetch_url(module, url, data=None, headers=None, method=None,
     force_basic_auth = module.params.get('force_basic_auth', '')
 
     follow_redirects = module.params.get('follow_redirects', 'urllib2')
+    kerberos_auth = module.params.get('kerberos_auth')
+    if kerberos_auth and not HAS_KERBEROS_AUTH:
+        module.fail_json(msg="kerberos or urllib_kerberos is not installed. Please install from pip")
```
puiterwijk, line 987: I am not sure we should advice people to install this from pip.

puiterwijk (changes requested):
I think there's some fixes that should be made to the underlying library, and there should probably be documentation or clear notes that the kerberos negotiation is initiated from the *ansible target*, meaning that the credentials (keytab or active credential cache) needs to be available there.

willthames:
Thanks for the review @puiterwijk - this fix was enough to get kerberos working for URL requests for my needs but probably not suitable for production release.

I'll just close this off - I have no need for this fix any longer, and better fixes are likely possible - using a better implementation of urllib combined with kerberos would probably be a good goal, but not one that I'm interested in implementing or maintaining.

PR #19039, opened 2016-12-09
Fix ignore_image option of docker_container module not work correctly.

tntC4stl3:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

cloud/docker/docker_container.py
##### ANSIBLE VERSION

```
ansible 2.3.0 (devel 680cade77a) last updated 2016/10/27 14:55:40 (GMT +800)
  lib/ansible/modules/core: (detached HEAD c51ced56cc) last updated 2016/10/27 14:58:15 (GMT +800)
  lib/ansible/modules/extras: (detached HEAD 8ffe314ea5) last updated 2016/10/27 14:58:15 (GMT +800)
```
##### SUMMARY

I have two docker images with same image ID but different image tag:

```
➜  ~ docker images | grep nginx 
nginx                                  1.10.2              54d9f6a22b41        7 days ago          180.7 MB
nginx                                  1.10.3              54d9f6a22b41        7 days ago          180.7 MB
```

And I have a running docker container belongs to nginx:1.10.2

```
➜  ~ docker ps
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS               NAMES
136914b076a7        nginx:1.10.2        "nginx -g 'daemon off"   5 hours ago         Up 5 hours          80/tcp, 443/tcp     nginx
```

Playbook:

```

---
- hosts: localhost
  tasks:
  - name: Test ignore_image
    docker_container: 
      name: nginx
      image: nginx:1.10.3
      ignore_image: true
      state: started
```

After execute the playbook, the running docker container becomes to nginx:1.10.3

```
➜  ~ ansible-playbook test.yml

ok: [localhost]

TASK [Test ignore_image] *******************************************************
changed: [localhost]

➜  ~ docker ps
CONTAINER ID        IMAGE               COMMAND                  CREATED              STATUS              PORTS               NAMES
696327d6e4d2        nginx:1.10.3        "nginx -g 'daemon off"   About a minute ago   Up About a minute   80/tcp, 443/tcp     nginx
```

Currently the ignore_image option can be set, but can not work as it is descripted in document. 
The reason is the code will check the difference of configurations between current container and target image, and it will mark the `different` to `True` when the image is different even we set `ignore_image=true`, that will cause the container being re-create.

So my solutions is ignore the difference of image when comparing the configurations of current docker container and target docker image.

### DIFF
--- a/lib/ansible/modules/cloud/docker/docker_container.py
+++ b/lib/ansible/modules/cloud/docker/docker_container.py
@@ -1251,7 +1251,6 @@ class Container(DockerBaseClass):
         # Map parameters to container inspect results
         config_mapping = dict(
             auto_remove=host_config.get('AutoRemove'),
-            image=config.get('Image'),
             expected_cmd=config.get('Cmd'),
             hostname=config.get('Hostname'),
             user=config.get('User'),


tntC4stl3:
Migrated from https://github.com/ansible/ansible-modules-core/pull/5416

lib/ansible/modules/cloud/docker/docker_container.py
```diff
@@ -1257,6 +1257,9 @@ def has_different_configuration(self, image):
             volume_driver=host_config.get('VolumeDriver')
         )
 
+        if kwargs.get('ignore_image', False):
```
dorianpula: Why not declare a keyword argument with a default of False, up at the method signature?  That would be more explicit than a kwargs.

dorianpula (changes requested):

lib/ansible/modules/cloud/docker/docker_container.py
tntC4stl3: Thank you for pointing this out, actually we can get `ignore_image` from `self.parameters`.

I will update this soon.

tntC4stl3:
@dorianpula Could you help to review again?
Thanks, :)

dorianpula (approved):
Looks good @tntC4stl3!

abadger:
It looks like has_different_configuration is only called from here: https://github.com/tntC4stl3/ansible/blob/cfa7f031bbfef890fd6e800bde751c8c55c604ad/lib/ansible/modules/cloud/docker/docker_container.py#L1740

And right after the call to different_configuration, there's a call to _image_is_different: https://github.com/tntC4stl3/ansible/blob/cfa7f031bbfef890fd6e800bde751c8c55c604ad/lib/ansible/modules/cloud/docker/docker_container.py#L1743

which is properly shielded inside of an ```if not self.parameters.ignore_image:```.  So perhaps we should remove checking of the image inside of has_different_configuration altogether.  (ie: simply remove this line: https://github.com/ansible/ansible/pull/19039/files#diff-ff29a0f2b632e64441519bf4d9c578d3R1254 ).

@tntC4stl3 It sounds like you'd be all setup to test that removing that line fixes ignore_image=True and does not break ignore_image=False.  Would you care to make just that change, test, and then update the PR if that works?

needs_info

tntC4stl3:
@abadger Let me make it more clear.
So your suggestion is remove the check on image in `has_different_configuration` and let the check remain in `_image_is_different`?

Hmm... probably a good idea and let me test for it.

abadger:
@tntC4stl3 Correct.  From just looking at the code, I think that will work.  If it works, I would favour it as it yields less code in the end.

tntC4stl3:
@abadger 

After removing the image check in `has_different_configuration`, I did two kinds of tests.

I have three nginx images on my laptop, and two of them have the same image id.
```
➜  /tmp docker images | grep nginx
nginx                                                1.10.2              e809eede2c4e        4 months ago        181MB
nginx                                                1.10.3              e809eede2c4e        4 months ago        181MB
nginx                                                1.10.1              bf2b4c2d7bf5        8 months ago        181MB
```

And before testing, the docker container is running under `nginx:1.10.2`:
```
➜  /tmp docker ps
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS               NAMES
6da08d4d181f        nginx:1.10.2        "nginx -g 'daemon ..."   13 minutes ago      Up 13 minutes       80/tcp, 443/tcp     nginx
```

## Test 1
After executing the playbook, the container will remain the same:
```
➜  /tmp cat test.yaml
---
- hosts: localhost
  tasks:
  - name: Test ignore_image
    docker_container: 
      name: nginx
      image: nginx:1.10.3
      ignore_image: true
      state: started
➜  /tmp ansible-playbook test.yaml 
 [WARNING]: No inventory was parsed, only implicit localhost is available

PLAY [localhost] ***********************************************************************************************************************************************************************************************************************

TASK [Gathering Facts] *****************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [Test ignore_image] ***************************************************************************************************************************************************************************************************************
ok: [localhost]

PLAY RECAP *****************************************************************************************************************************************************************************************************************************

➜  /tmp docker ps
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS               NAMES
6da08d4d181f        nginx:1.10.2        "nginx -g 'daemon ..."   15 minutes ago      Up 15 minutes       80/tcp, 443/tcp     nginx
```

## Test 2
The container will be recreated due to the difference of `expected_env` in `has_different_configuration` checking.
```
➜  /tmp cat test.yaml                              
---
- hosts: localhost
  tasks:
  - name: Test ignore_image
    docker_container: 
      name: nginx
      image: nginx:1.10.1
      ignore_image: true
      state: started
➜  /tmp ansible-playbook test.yaml
 [WARNING]: No inventory was parsed, only implicit localhost is available

PLAY [localhost] ***********************************************************************************************************************************************************************************************************************

TASK [Gathering Facts] *****************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [Test ignore_image] ***************************************************************************************************************************************************************************************************************
changed: [localhost]

PLAY RECAP *****************************************************************************************************************************************************************************************************************************

➜  /tmp docker ps
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS               NAMES
f8cf3c8af105        nginx:1.10.1        "nginx -g 'daemon ..."   3 seconds ago       Up 2 seconds        80/tcp, 443/tcp     nginx
```
Configuration difference:
```
[{'expected_env': {'container': [u'PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin', u'NGINX_VERSION=1.10.2-1~jessie'], 'parameter': [u'PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin', u'NGINX_VE
RSION=1.10.1-1~jessie']}}]
```

abadger:
@tntC4stl3 Cool.  Can you also do one test that ```ignore_image: false``` works as expected?  If that works too, then this looks good to merge.

tntC4stl3:
hmm... Got problem now.

```
➜  /tmp docker ps
CONTAINER ID        IMAGE               COMMAND                  CREATED              STATUS              PORTS               NAMES
4ff9b4913208        nginx:1.10.2        "nginx -g 'daemon ..."   About a minute ago   Up About a minute   80/tcp, 443/tcp     nginx
➜  /tmp cat test.yaml 
---
- hosts: localhost
  tasks:
  - name: Test ignore_image
    docker_container: 
      name: nginx
      image: nginx:1.10.3
      ignore_image: false
      state: started
➜  /tmp ansible-playbook test.yaml
 [WARNING]: No inventory was parsed, only implicit localhost is available

PLAY [localhost] ***********************************************************************************************************************************************************************************************************************

TASK [Gathering Facts] *****************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [Test ignore_image] ***************************************************************************************************************************************************************************************************************
ok: [localhost]

PLAY RECAP *****************************************************************************************************************************************************************************************************************************

➜  /tmp docker ps
CONTAINER ID        IMAGE               COMMAND                  CREATED              STATUS              PORTS               NAMES
4ff9b4913208        nginx:1.10.2        "nginx -g 'daemon ..."   About a minute ago   Up About a minute   80/tcp, 443/tcp     nginx
```
The playbook doesn't re-create the container because the `_image_is_different` check the image id and treat nginx:1.10.2 and nginx:1.10.3 as the same image.

So I think we should come to an agreement for those cases (suppose no other difference in has_different_configuration checking):

|                                 | Image A and B have same ID | Image A and B have different ID |
| -------------------------- | --------------------------------------- | ------------------------------------------- |
| ignore_image=true  | no re-create                            | no re-create                                 |
| ignore_image=false | re-create                                 | re-create                                      |

What's your opinion @abadger ?
I'm gonna to find other solution if you agree with me.

abadger:
I'm not a docker user so I don't know what people would expect... @dorianpula @ryansb @chouseknecht ideas for what should be the expected outcomes in the last comment?

ryansb:
I think that for your table of examples, I'd prefer "A and B having same ID and no other differences" to not re-create. 

|                                 | Image A and B have same ID | Image A and B have different ID |
| -------------------------- | --------------------------------------- | ------------------------------------------- |
| ignore_image=true  | no re-create                            | no re-create                                 |
| ignore_image=false | no re-create                                 | re-create                                      |

Because I'd want to be able to run the module with no changes, and have it not change the state of my running containers.

tntC4stl3:
@ryansb If so, the current patch in this PR already fit your requirement now and no need for further modification.

tntC4stl3:
@abadger @ryansb do you guys have other concern on this PR?

ryansb (approved):

PR #19042, opened 2016-12-09
[PostgreSQL] Support forced SSL-verified connection with given CA certificate

veger:
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
postgresql modules

##### ANSIBLE VERSION
```
ansible 2.2.0.0
```

##### SUMMARY
Add support for a forced SSL connection to a PostgreSQL databases using a given certificate.
When the certificate does not match (or is incorrect at any way), the connection is refused.

### DIFF
--- a/lib/ansible/modules/database/postgresql/postgresql_privs.py
+++ b/lib/ansible/modules/database/postgresql/postgresql_privs.py
@@ -120,6 +120,20 @@ options:
       - 'Alias: I(login_password))'
     default: null
     required: no
+  ssl_mode:
+    description:
+      - Determines whether or with what priority a secure SSL TCP/IP connection will be negotiated with the server.
+      - See https://www.postgresql.org/docs/current/static/libpq-ssl.html for more information on the modes.
+    required: false
+    default: disable
+    choices: [disable, allow, prefer, require, verify-ca, verify-full]
+    version_added: '2.3'
+  ssl_rootcert:
+    description:
+      - Specifies the name of a file containing SSL certificate authority (CA) certificate(s). If the file exists, the server's certificate will be verified to be signed by one of these authorities.
+    required: false
+    default: null
+    version_added: '2.3'
 notes:
   - Default authentication assumes that postgresql_privs is run by the
     C(postgres) user on the remote host. (Ansible's C(user) or C(sudo-user)).
@@ -139,6 +153,7 @@ notes:
     specified via I(login). If R has been granted the same privileges by
     another user also, R can still access database objects via these privileges.
   - When revoking privileges, C(RESTRICT) is assumed (see PostgreSQL docs).
+  - The ssl_rootcert parameter requires at least Postgres version 8.4 and I(psycopg2) version 2.4.3.
 requirements: [psycopg2]
 author: "Bernhard Weitzhofer (@b6d)"
 """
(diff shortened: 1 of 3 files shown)


veger:
Migrated from https://github.com/ansible/ansible-modules-core/pull/5698

lib/ansible/modules/database/postgresql/postgresql_privs.py
```diff
@@ -120,6 +120,18 @@
       - 'Alias: I(login_password))'
     default: null
     required: no
+  ssl_mode:
+    description:
+      - Determines whether or with what priority a secure SSL TCP/IP connection will be negotiated with the server
```
gundalow: Missing full stop at the end of this sentence, and the one later on

gundalow (changes requested):

lib/ansible/modules/database/postgresql/postgresql_privs.py
```diff
@@ -120,6 +120,18 @@
       - 'Alias: I(login_password))'
     default: null
     required: no
+  ssl_mode:
```
gundalow, line 123: Missing example or detail of what the valid options are.

gundalow (changes requested):
needs_revision 

Docs/examples need expanding to explain in more details how to use the options.

veger:
Implemented feedback of @gundalow.

Additionally, updated the description of `ssl_mode` with the possible choices and added 'forced' default instead of relying on the PostgreSQL default implementation (which might change in the future...)

gundalow (approved):

gundalow:
Thanks for the updates.

I'll let someone else from the Core Team look at this then they can merge into `devel`

alikins:
Should the new modules options be sslmode/sslrootcert (to match the postgresql names?)

veger:
I matched the modules parameter style/names, which seemed most user-friendly to me.

If the (generic) preference is to support the postgresql names, let me know and I'll make the changes.

lib/ansible/modules/database/postgresql/postgresql_privs.py
```diff
@@ -541,7 +557,9 @@ def main():
             port=dict(type='int', default=5432),
             unix_socket=dict(default='', aliases=['login_unix_socket']),
             login=dict(default='postgres', aliases=['login_user']),
-            password=dict(default='', aliases=['login_password'], no_log=True)
+            password=dict(default='', aliases=['login_password'], no_log=True),
+            ssl_mode=dict(default='disable'),
+            ssl_rootcert=dict(default='')
```
abadger: default here is empty string but both in the documentation and in the other file, postgresql_user.py , it's None.  Should standardize on either None or empty string in all places.

abadger:
Left some comments.  On the question of how to spell the module parameters, I can see the argument either way.  I think matching the other module params is probably the slightly better way to go.

Once the comments are addressed this can be merged.

abadger:
needs_revision

lib/ansible/modules/database/postgresql/postgresql_privs.py
veger: I tried to keep the same format per module. Each module uses its own format (empty string/ None/ etc). I agree that standardizing this is the way to go! But I also think that this is outside the scope of this PR..?

abadger: Not sure I understand what you're saying?  This PR adds ssl_rootcert to both modules.  Seems like they should both default to either empty string or both default to None?

veger: postgres_db.py uses `""` for its defaults, postgres_privs.py uses `''` and postgres_user.py uses `None`. This PR tries to adhere to the format/style of each file (which differs between each other...).

Postgres itself does not mind either way.

abadger: If postgres doesn't care, let's standardize on None in all three modules.

abadger:
Only one thing to change.  Once that's done, this is good to merge.

abadger:
@veger do you have everything you need to update this or do you need more information from me?

veger:
I have everything (but time :P ), hopefully I'll get to it tomorrow

veger:
Changed the defaults to `None`

abadger:
/me trying to figure out why the postgres integration tests failed for this PR.

abadger:
Ahh... you misunderstood me. Don't mess with the defaults for the other options.  Only change the defaults for the ssl options you're adding.  Changing the other options breaks the code.

veger:
Oops... I have reverted those changes. Hopefully the tests will pass now!

veger:
Hm... looks like the PostgreSQL installed on CentOS 6.1 does not support the `sslrootcert` parameter.

Can we ignore this, or is this a serious issue?

abadger:
@veger it is an issue but we can conditionalize use of the parameter on the version of the server.  In the documentation, update it to say that the sslrootcert parameter is only valid for postgresql server >= X.Y

In the code it'll be trickier because we need the parameter in order to connect.  I suggest that we omit the parameter if it's None (since that's the default and means the user hasn't set it).  If it's not None and the connection to the server fails, it will be nice to look at the error message and if it tells us that it fails because of sslrootcert, then we return a nice error via module.fail_json() telling the user that postgresql needs to be at least version X.Y in order to use sslrootcert.

abadger:
@vger okay, I haven't tested but I think this is about what you need to do.  I only changed postgres_db but the other files should be similar: https://toshio.fedorapeople.org/ansible/postgres.patch

veger:
@abadger Thanks for the patch, I used it as a basis to modify the 3 modules.

Centos 6.1 tests now passing as well

abadger:
Thanks @veger!  Merged to the devel branch for inclusion in 2.3.0

PR #19043, opened 2016-12-09
Leave current ElastiCache security groups as-is if parameter is not provided

veger:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME
elasticache

##### ANSIBLE VERSION
```
ansible 2.2.0.0
```

##### SUMMARY
When no security groups are provided to this module, the default (empty list) is used. This is not allowed by Amazon, and undesired in my opinion.

This PR modifies this behavior, by leaving the existing security groups when the default/empty list is provided. Which makes it easier to modify the elasicache instance whith the need to grabbing the current security groups first (in order to keep them)

### DIFF
--- a/lib/ansible/modules/cloud/amazon/elasticache.py
+++ b/lib/ansible/modules/cloud/amazon/elasticache.py
@@ -397,12 +397,13 @@ class ElastiCacheManager(object):
             return True
 
         # check vpc security groups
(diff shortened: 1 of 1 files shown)


veger:
Migrated from https://github.com/ansible/ansible-modules-core/pull/5747

PR #19047, opened 2016-12-09
swdepot.py : filter swlist result header

melankh:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
packaging/os/swdepot.py 

##### ANSIBLE VERSION
```
ansible 2.2.0.0
```

##### SUMMARY
swlist add a header that could match package name.
Real example : "tar" that match "target"

```
$swlist -a revision -l product tar | grep tar
# Contacting target "yrvm08"...
  tar   1.29
```

So filtering it via "grep -v \#"
```
$ swlist -a revision -l product tar | grep -v \# |  grep tar
  tar   1.29
```


melankh:
Migrated from https://github.com/melankh/ansible-modules-extras/pull/1

lib/ansible/modules/packaging/os/swdepot.py
```diff
@@ -93,9 +93,9 @@ def query_package(module, name, depot=None):
 
     cmd_list = '/usr/sbin/swlist -a revision -l product'
     if depot:
-        rc, stdout, stderr = module.run_command("%s -s %s %s | grep %s" % (cmd_list, pipes.quote(depot), pipes.quote(name), pipes.quote(name)), use_unsafe_shell=True)
+        rc, stdout, stderr = module.run_command("%s -s %s %s | grep -v \# | grep %s" % (cmd_list, pipes.quote(depot), pipes.quote(name), pipes.quote(name)), use_unsafe_shell=True)
     else:
-        rc, stdout, stderr = module.run_command("%s %s | grep %s" % (cmd_list, pipes.quote(name), pipes.quote(name)), use_unsafe_shell=True)
+        rc, stdout, stderr = module.run_command("%s %s | grep -v \# | grep %s" % (cmd_list, pipes.quote(name), pipes.quote(name)), use_unsafe_shell=True)
```
bcoca: the grep seems too generic and will remove any line with a #, you probably only want lines beginning with #

jctanner:
@melankh do you intend to rebase this?

needs_info

bcoca (changes requested):

lib/ansible/modules/packaging/os/swdepot.py
melankh: There's no # in the results of swinstall (or swcopy/swverify/...) except in the beginning of the lines.

But, if you prefer, the grep command can be replaced by :   **`egrep -v \^\#`**

bcoca: no need, if the tools enforce that already, you still need to fix the conflicts

PR #19049, opened 2016-12-09
vmware_guest: Add multiple nics management + Bug fixes

aperigault:
##### ISSUE TYPE
 - Feature Pull Request
 - Bugfix Pull Request

##### COMPONENT NAME
vmware_guest

##### ANSIBLE VERSION
```
ansible 2.2.0.0
```

##### SUMMARY
- Fix var name:
function getfolder return folder_map instead of foldermap.

- Fix bug in deploy_template function:
If folder not start by /vm there is a "no folder matched the path" error

- Power on VM if state is set to poweredon

New feature : Move nic creation out of customizations + Add multiple nics management

### DIFF
--- a/lib/ansible/modules/cloud/vmware/vmware_guest.py
+++ b/lib/ansible/modules/cloud/vmware/vmware_guest.py
@@ -821,6 +821,12 @@ class PyVmomiHelper(object):
                 nic.device.wakeOnLanEnabled = True
                 nic.device.addressType = 'assigned'
 
+                if 'type' in device:
+                    if device['type'].lower() == "e1000":
+                        nic.device = vim.vm.device.VirtualE1000()
+                    elif device['type'].lower() == "vmxnet2":
+                        nic.device = vim.vm.device.VirtualVmxnet2()
+
                 if hasattr(get_obj(self.content, [vim.Network], device['network']), 'portKeys'):
                     # VDS switch
                     pg_obj = get_obj(self.content, [vim.dvs.DistributedVirtualPortgroup], device['network'])
--- a/lib/ansible/modules/cloud/vmware/vmware_guest.py
+++ b/lib/ansible/modules/cloud/vmware/vmware_guest.py
@@ -404,8 +404,8 @@ class PyVmomiHelper(object):
         if not self.datacenter:
             self.get_datacenter()
         self.folders = self._build_folder_tree(self.datacenter.vmFolder)
-        self.folder_map = self._build_folder_map(self.folders)
-        return (self.folders, self.folder_map)
+        self.foldermap = self._build_folder_map(self.folders)
+        return (self.folders, self.foldermap)
 
     def compile_folder_path_for_object(self, vobj):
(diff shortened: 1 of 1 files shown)


aperigault:
Migrated from https://github.com/ansible/ansible-modules-extras/pull/3481

gundalow:
I believe the current maintainers are @jctanner, @jcpowermac, @mtnbikenc, @chrrrles could you please review, thanks.

nerzhul:
Please note i did a similar but more complete version here:

https://github.com/ansible/ansible/pull/19228

I didn't saw this PR, and i changed some bad design, but noticed some identical bugs and added more features

aperigault:
@nerzhul I read your PR. There is good things but I'm not agree with your nic management. For my use case I need to add nics only with a vlan id args. I think we need to manage nics and customizations separately.

nerzhul:
vlan id is not used everytime, but could be an improvement, i'm working on idempotency and creating VM from scratch whereas the original version you modify only used templates

aperigault:
The vlan_id feature is an improvement already include in my PR. The goal is to add nics (with vlan_id for example or with network_name) with a minimum informations. Your PR need too many informations to add a nic.

nerzhul:
@aperigault i agree it's not useful, i only kept the origin module compatibility, i also think there is too many PR, only have the switch name or the vlan id is sufficient

jctanner:
@dav1x can you review this please?

dav1x:
Tested and works great @jctanner

gundalow:
Looks like this needs a rebase.

jctanner (approved):

jctanner:
@aperigault please rebase your branch.

aperigault:
@jctanner Rebased !

angystardust:
@aperigault can you please add a configuration example with 2 nics? thanks

aperigault:
@angystardust I added an example with two nics. One with network_name and one with network_vlan. You need to define one of these.

angystardust:
Thanks a lot @aperigault.
How about the ip customization instead? Can I do that with multiple ips? How can I distinguish between the first nic and the second one?

dav1x:
@angystardust the differentiation is the VM Network name (or port group). I have code that will set multiple IP addresses once this PR is merged. It's based on this code.

nerzhul:
@aperigault why not merge your change into #19228 instead of doing the same thing twice.

#19228 has many interesting change the only missing change which is in this PR is the vlan ID support

this would permit a better feature integration as i pass many times to optimize the module and add more features, and you too on network.

The add mode on disk is non idempotent it's not good for ansible.

aperigault:
@nerzhul I will retest your branch and do a PR in your branch.

nerzhul:
@aperigault thanks :)

aperigault:
@dav1x @jctanner @nerzhul I merged my changes in nerzhul PR #19228 
I think we need to merge this quickly because there is a lot of good changes that break all others PR

PR #19057, opened 2016-12-09
Fix UnboundLocalError remote_head in git

robinro:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
git

##### ANSIBLE VERSION
```devel```

##### SUMMARY

Fixes ansible/ansible-modules-core#5505

The use of remote_head was a leftover of ansible/ansible-modules-core#4562.
remote_head is not necessary, since the repo is unchanged anyway and
after is set correctly.

Further changes:
* Set changed=True and msg once local_mods are detected and reset.
* Remove need_fetch that is always True (due to previous if) to improve
clarity.
* Don't exit early for local_mods but run submodules update and
switch_version.

### DIFF
--- a/lib/ansible/modules/source_control/git.py
+++ b/lib/ansible/modules/source_control/git.py
@@ -888,7 +888,7 @@ def main():
     ssh_opts  = module.params['ssh_opts']
     umask  = module.params['umask']
 
-    result = dict( warnings=list() )
+    result = dict(changed = False, warnings=list())
 
     # evaluate and set the umask before doing anything else
     if umask is not None:
@@ -968,7 +968,7 @@ def main():
         # this does no checking that the repo is the actual repo
         # requested.
         result['before'] = get_version(module, git_path, dest)
-        result.update(changed=False, after=result['before'])
+        result.update(after=result['before'])
         module.exit_json(**result)
     else:
         # else do a pull
@@ -981,6 +981,7 @@ def main():
             # if force and in non-check mode, do a reset
             if not module.check_mode:
                 reset(git_path, module, dest)
+                result.update(changed=True, msg='Local modifications exist.')
 
         # exit if already at desired sha version
         if module.check_mode:
@@ -990,26 +991,19 @@ def main():
             remote_url_changed = set_remote_url(git_path, module, repo, dest, remote)
         result.update(remote_url_changed=remote_url_changed)
 
-        if need_fetch:
-            if module.check_mode:
-                remote_head = get_remote_head(git_path, module, dest, version, remote, bare)
-                result.update(changed=(result['before'] != remote_head), after=remote_head)
-                # FIXME: This diff should fail since the new remote_head is not fetched yet?!
-                if module._diff:
-                    diff = get_diff(module, git_path, dest, repo, remote, depth, bare, result['before'], result['after'])
-                    if diff:
-                        result['diff'] = diff
-                module.exit_json(**result)
-            else:
-                fetch(git_path, module, repo, dest, version, remote, depth, bare, refspec, git_version_used)
+        if module.check_mode:
+            remote_head = get_remote_head(git_path, module, dest, version, remote, bare)
+            result.update(changed=(result['before'] != remote_head), after=remote_head)
+            # FIXME: This diff should fail since the new remote_head is not fetched yet?!
+            if module._diff:
+                diff = get_diff(module, git_path, dest, repo, remote, depth, bare, result['before'], result['after'])
+                if diff:
+                    result['diff'] = diff
+            module.exit_json(**result)
+        else:
+            fetch(git_path, module, repo, dest, version, remote, depth, bare, refspec, git_version_used)
 
         result['after'] = get_version(module, git_path, dest)
-
-        if result['before'] == result['after']:
-            if local_mods:
-                result.update(changed=True, after=remote_head, msg='Local modifications exist')
-                # no diff, since the repo didn't change
-                module.exit_json(**result)
 
     # switch to version specified regardless of whether
     # we got new revisions from the repository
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by robinro** [comment]
Migrated from https://github.com/ansible/ansible-modules-core/pull/5720

**Comment by gundalow** [comment]
From the original PR
> Since this affects quite a few people I would suggest to also have @jurgenhaas or one of the other commenters make sure this covers their use case before merging it.

**Comment by jurgenhaas** [comment]
I've tested this approach and can confirm that my particular use case is working with this one as well.

**Comment by gundalow** [comment]
Thanks for the review. 
I wonder if this should go into `stable-2.2`, will discuss in Core Meeting on Tuesday (it's already on the agenda)

**Comment by abadger** [comment]
@robinro I just started looking at migrating this: https://github.com/ansible/ansible/pull/19257  since @ryansb had given that one a +1.  (original PR: https://github.com/ansible/ansible-modules-core/pull/5260/files ).  Do these two approaches need to be merged?  Is part of https://github.com/ansible/ansible/pull/19257  needed but hte other part unrelated?

**Comment by robinro** [comment]
From a quick look I can't think of a usecase that #19257 improves over this one, but I might be missing something. Maybe @sbussetti or @ryansb have more to add?

In general we should have integration tests for these codepaths.

**Comment by sbussetti** [comment]
@robinro -- yeah looks like ships passing in the night.

I ran into this branch of code with the bad scoping for `remote_head` as a result of doing shallow forced checkouts.  Honestly I had a tough time figuring a way to write a test for that, (any examples of a similar test would be warmly welcomed), however, I'm gonna live test #19057 now and report back.

**Comment by sbussetti** [comment]
So, this pull fixes the scoping issue, but there's a latent problem here that my pull also addressed.  I'm using a directive like the following to deploy my source:

```
    - name: Deploy Source
      become: true
      become_user: "{{deploy_user}}"
      git:
        accept_hostkey=True
        depth=1
        force=True
        repo="{{repo_url}}"
        dest="{{www_root}}"
        remote="{{repo_remote}}"
        version="{{repo_sha|default(repo_branch, true)}}"
        refspec="{{repo_ref|default('refs/heads/' ~ repo_branch, true)}}"
        key_file="/home/{{deploy_user}}/.ssh/id_rsa-deploy"
      tags:
        - source

```

And the git module skips one critical step (on the devel and on robinro's) branch -- that being actually fetching the requested branch / spec / version.   Since it never fetches, it just hard resets to whatever was available the first time it checked out the repo.  Subsequent runs get "stuck" on that earlier commit.

My adjustments to reset, here: https://github.com/ansible/ansible/pull/19257/files#diff-423dabf0f8d841053684543485e386c9L435

Which were related to this but also smoothed over the issue because without fetching the appropriate head, you're force resetting to the wrong thing.

It seems like I need to submit a new PR to just address that?  Please advise.

**Comment by robinro** [comment]
For tests I would suggest to use git repos that are setup within the test role itself. So far we rely mostly on external repositories, which make it harder to understand how the testcase was constructed and adds a dependency on the network/github/some-private-repo being available.

An example I like is #18728. You write a shell one-liner that sets up your repository/... and then you run the git module to test it. One can also clone a repo and then run some shell commands to add local modifications. I'd be happy to review and assist with those tests.

**Comment by abadger** [comment]
@robinro So it sounds like we want this change to fix remote_head and then part of https://github.com/ansible/ansible-modules-core/pull/5260 to fix the fetching/reset problem?

@sbussetti Feel free to migrate/redo https://github.com/ansible/ansible-modules-core/pull/5260 to do that and either you or robinro can ping me in irc to make sure I see it and merge it.  And yes, tets are very much appreciated.  The git module code has many cornercases that we run the risk of re-introducing without tests :-)

**Comment by robinro** [comment]
I added a testcase for local modifications and overwriting them with force while fetching a new commit from a remote repo. Both for full-history and shallow clones. They both pass with the changes made here.

@sbussetti I don't see the "getting stuck on an old commit" situation in that test. Can you give a (minimal) working example that fails? Maybe you can construct a test based on the one I added here?

**Comment by sbussetti** [comment]
@robinro  On it -- thank you for the guidance so far.

**Comment by sbussetti** [comment]
Okay, 3d52ae53f17928cde7d9d14e101a83c0e33b0384 provides an exhaustive test for my issue.

What I was experiencing (basically that on subsequent deploys, HEAD was not being fetched from the remote), is no longer occurring against mainline devel.  So either it was an artifact of my pull, or otherwise resolved in devel, so I think case closed on this one.  Would like to get this test added to the PR though if possible, and drop my changes in #19257

**Comment by robinro** [comment]
It seems all issues are fixed now and this can be merged.
For the test another PR is probably best.

**Comment by sbussetti** [comment]
@robinro Done.  Cheers!

**Comment by abadger** [comment]
Merged to devel and backported to stable-2.2.

**Comment by meabed** [comment]
I still have the error : 
```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: UnboundLocalError: local variable 'remote_head' referenced before assignment
fatal: [10.100.90.54]: FAILED! => {"changed": false, "failed": true, "module_stderr": "Traceback (most recent call last):\n  File \"/tmp/ansible_gRy6vs/ansible_module_git.py\", line 1040, in <module>\n    main()\n  File \"/tmp/ansible_gRy6vs/ansible_module_git.py\", line 994, in main\n    result.update(changed=True, after=remote_head, msg='Local modifications exist')\nUnboundLocalError: local variable 'remote_head' referenced before assignment\n", "module_stdout": "", "msg": "MODULE FAILURE"}
```

robinro:
@Meabed Which version of the git module are you using? Did you clone the devel branch within the last 7 days? The error message is exactly as for the old version. So far the fix is not part of any public release.

PR #19060, opened 2016-12-09
nmcli: Use dbus only if it is present

Fale:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME
network/nmcli

##### ANSIBLE VERSION

```
2.2
```

##### SUMMARY
Call dbus subfunction only if the dbus library is present, otherwise the following error appears:

```
NameError: name 'dbus' is not defined
```

### DIFF
--- a/lib/ansible/modules/network/nmcli.py
+++ b/lib/ansible/modules/network/nmcli.py
@@ -524,7 +524,8 @@ class Nmcli(object):
 
     platform='Generic'
     distribution=None
-    bus=dbus.SystemBus()
+    if HAVE_DBUS:
+        bus=dbus.SystemBus()
     # The following is going to be used in dbus code
     DEVTYPES={1: "Ethernet",
                    2: "Wi-Fi",


Fale:
Migrated from https://github.com/ansible/ansible-modules-extras/pull/3664

lib/ansible/modules/network/nmcli.py
```diff
@@ -524,7 +524,8 @@ class Nmcli(object):
 
     platform='Generic'
     distribution=None
-    bus=dbus.SystemBus()
+    if HAVE_DBUS:
```
gundalow, line 527: This should be dealt with at
https://github.com/ansible/ansible/pull/19060/files#diff-fc8edeab30941b14afd491f7f42fb881R1139

Not sure why this is been run before main, that feels like an implementation issue

gundalow (changes requested):

lib/ansible/modules/network/nmcli.py
Fale: Because is in the class (outside methods), so it gets run at class init phase

gundalow (approved):

PR #19070, opened 2016-12-09
win_file: fix error when creating an existing dir

ilpianista:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

win_file
##### ANSIBLE VERSION

```
ansible 2.1.1.0
```
##### SUMMARY

Fix error when creating an existing dir.

If something else created the dir New-Item will throw an exception and the task will fail. Check the existing file and as long as it's a dir, we don't need to error out.

### DIFF
--- a/lib/ansible/modules/windows/win_file.ps1
+++ b/lib/ansible/modules/windows/win_file.ps1
@@ -66,7 +66,7 @@ function Remove-File($file, $checkmode) {
             Remove-Item -Path $file.FullName -Force -WhatIf:$checkmode
         }
     } catch [Exception] {
-        Fail-Json (New-Object psobject) "Failed to delete $($file.FullName): $($_.Exception.Message)"
+        Fail-Json $result "Failed to delete $($file.FullName): $($_.Exception.Message)"
     }
 }
 
@@ -116,7 +116,18 @@ if (Test-Path $path) {
     }
 
     if ($state -eq "directory") {
-        New-Item -Path $path -ItemType Directory -WhatIf:$check_mode | Out-Null
+        try {
+            New-Item -Path $path -ItemType Directory -WhatIf:$check_mode | Out-Null
+        } catch {
+            if ($_.CategoryInfo.Category -eq "ResourceExists") {
+                $fileinfo = Get-Item $_.CategoryInfo.TargetName
+                if ($state -eq "directory" -and -not $fileinfo.PsIsContainer) {
+                    Fail-Json $result "path $path is not a directory"
+                }
+            } else {
+                Fail-Json $result $_.Exception.Message
+            }
+        }
         $result.changed = $true
     } elseif ($state -eq "file") {
         Fail-Json $result "path $path will not be created"


ilpianista:
Migrated from https://github.com/ansible/ansible-modules-core/pull/4823

gundalow:
Following on from @jhawkesworth comment in the original PR, as the modules have now moved into the main repo it would be great if a test  was added to defend this fix. Tests are in `test/integration/targets/`

abadger:
\cc @nitzmahone

ilpianista:
@gundalow looking at the tests folder, I cannot find an example of a tests that runs the same task **in parallel**. Any help?

gundalow:
I'm not aware of any, @mattclay any ideas?

mattclay:
We don't have integration tests for race conditions, as there's typically no way to guarantee that the race condition is actually encountered during the test. That kind of testing is usually more suited for unit testing, where we can use mocking to trigger the necessary code paths. However, we only have unit testing for python code currently.

gundalow:
Also I think we are on diminishing returns for that. If you have some spare cycles reviewing the existing `win_file` tests in https://github.com/ansible/ansible/tree/devel/test/integration/targets/win_file would be cool, though that would be a separate PR

gundalow:
Rebase needed

ilpianista:
Thank you @gundalow I missed the notification.

jhawkesworth:
@ilpianista if you need help rebasing please see this
https://docs.ansible.com/ansible/dev_guide/developing_rebasing.html
if you need further help please ask in #ansible-devel irc channel.

ilpianista:
@jhawkesworth what's wrong with my rebase?

jhawkesworth:
Sorry @ilpianista I must have been looking at an outdated version of the page - there is nothing wrong with your rebase.

jhawkesworth:
The integration tests are still passing and this does add better exception handling so I see no reason not to include this

shipit

jhawkesworth:
bot_status

gundalow:
Rebase needed

lib/ansible/modules/windows/win_file.ps1
```diff
@@ -116,7 +116,20 @@ if (Test-Path $path) {
...
     if ($state -eq "directory") {
-        New-Item -Path $path -ItemType Directory -WhatIf:$check_mode | Out-Null
+        if (-not $check_mode) {
+            try {
+                New-Item -Path $path -ItemType Directory -WhatIf:$check_mode | Out-Null
+            } catch {
+                if ($_.CategoryInfo.Category -eq "ResourceExists") {
+                    $fileinfo = Get-Item $_.CategoryInfo.TargetName
+                    if ($state -eq "directory" -and -not $fileinfo.PsIsContainer) {
+                        Fail-Json (New-Object psobject) "path is not a directory"
```
dagwieers: Make this:
```powershell
Fail-Json $result "Existing path $path is not a directory"
```

lib/ansible/modules/windows/win_file.ps1
```diff
@@ -116,7 +116,20 @@ if (Test-Path $path) {
...
+                    }
+                } else {
+                    Fail-Json (New-Object psobject) $_.Exception.Message
```
dagwieers: Make this:
```powershell
Fail-Json $result $_.Exception.Message
```

dagwieers (changes requested):

ilpianista:
@dagwieers thanks, done

lib/ansible/modules/windows/win_file.ps1
```diff
@@ -66,7 +66,7 @@ function Remove-File($file, $checkmode) {
             Remove-Item -Path $file.FullName -Force -WhatIf:$checkmode
         }
     } catch [Exception] {
-        Fail-Json (New-Object psobject) "Failed to delete $($file.FullName): $($_.Exception.Message)"
+        Fail-Json $result "Failed to delete $file.FullName: $_.Exception.Message"
```
dagwieers: Does this work ? I though it had to be "$($file.FullName)" and "$($_.Exception.Message)".

lib/ansible/modules/windows/win_file.ps1
```diff
@@ -116,7 +116,20 @@ if (Test-Path $path) {
     }
 
     if ($state -eq "directory") {
-        New-Item -Path $path -ItemType Directory -WhatIf:$check_mode | Out-Null
+        if (-not $check_mode) {
```
dagwieers: This is not needed because we already do `-WhatIf:$check_mode`.
And the failure paths should be tested in check-mode to.

dagwieers (changes requested):

lib/ansible/modules/windows/win_file.ps1
ilpianista: This is how it's done at line 66, so I thought this is the same case

dagwieers: Please test it, I am petty sure it does not work in a string.

ilpianista: I can't test it at the moment, so I reverted it.

nitzmahone (approved):

PR #19072, opened 2016-12-09
Source Mage GNU/Linux package manager support

vaygr:
I'm glad to present the 'sorcery' module for Ansible.

[Sorcery](http://sourcemage.org/Sorcery) is a package management system for some super-flexible source-based Linux distro called [Source Mage GNU/Linux](http://sourcemage.org). It is one of the [oldest](http://sourcemage.org/linux-timeline.png) distros available, and even while being not so popular, it still takes [its niche](http://sourcemage.org/About) quite well.

The module provides wide-range functionality for controlling SMGL (so we call them) systems the way you want them to be (pretty minimal and up-to-date), including dependencies management, full/partial system rebuild, updating of the components (package manager, collections of software, packages), so we could provide a universal alternative to our outdated [Cabal command](http://sourcemage.org/Sorcery/Commands/Cabal), used to be a multi-system control tool.

[Intro page](http://sourcemage.org/Intro) is there if you need some more information on how our system works.
You can also find us lurking on #sourcemage at Freenode IRC network and ask a question!

Thanks!

### DIFF
--- a/lib/ansible/modules/packaging/os/sorcery.py
+++ b/lib/ansible/modules/packaging/os/sorcery.py
@@ -360,14 +360,12 @@ def match_depends(module):
     if rc != 0:
         module.fail_json(msg="wrong dependencies for spell '%s'" % spell)
 
-    rex_spell = re.compile("^%s:" % spell)
-
     fi = fileinput.input(sorcery_depends, inplace=True)
 
     try:
         try:
             for line in fi:
-                if rex_spell.match(line):
+                if line.startswith(spell + ':'):
                     match = None
 
                     for d in depends:
--- a/lib/ansible/modules/packaging/os/sorcery.py
+++ b/lib/ansible/modules/packaging/os/sorcery.py
@@ -142,7 +142,6 @@ import fileinput
 import os
 import re
 import shutil
-import sys
 
 
 # auto-filled at module init
@@ -384,10 +383,7 @@ def match_depends(module):
                             if match.group('lstatus') == depends[d]:
                                 depends[d] = None
 
-                                if sys.version_info[0] > 2:
-                                    print(line, end='')
-                                else:
-                                    print line,
+                                print line,
 
                             # status is not that we need, so keep this dependency
                             # in the list for further reverse switching;
--- a/lib/ansible/modules/packaging/os/sorcery.py
+++ b/lib/ansible/modules/packaging/os/sorcery.py
@@ -63,7 +63,7 @@ options:
               to turn it on/off ('+' is optional though)
             - this option is ignored if C(name) parameter is equal to '*' or
               contains more than one spell
-            - providers must be supplied in the form recognized by Sorcery, i.e.
+            - providers must be supplied in the form recognized by Sorcery, e.g.
               'openssl(SSL)'
         required: false
         default: null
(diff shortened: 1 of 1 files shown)


vaygr:
Migrated from https://github.com/ansible/ansible-modules-extras/pull/1040

lib/ansible/modules/packaging/os/sorcery.py
```diff
@@ -0,0 +1,632 @@
...
+        description:
+            - Time in seconds to invalidate grimoire collection on update
+            - especially useful for SCM and rsync grimoires
+            - makes sense only in pair with C(update_cache)
+        required: false
+        default: null
+'''
+
+
+EXAMPLES = '''
```
gundalow, line 96: Can you please rewrite the examples with a 
`name:` and multi line statements using `key: value`
We are in the process of standardising all the modules to use that format.

lib/ansible/modules/packaging/os/sorcery.py
```diff
@@ -0,0 +1,632 @@
...
+        update_codex(module)
+
+    if params['name']:
+        manage_spells(module)
+
+
+# import module snippets
+from ansible.module_utils.basic import *
+
+main()
```
gundalow: To enable the module to work with debuggers can you please rewrite this as:
```
if __name__ == '__main__':
    main()
```

lib/ansible/modules/packaging/os/sorcery.py
```diff
@@ -0,0 +1,632 @@
...
+
+    if params['update_cache'] or params['state'] == 'latest':
+        update_codex(module)
+
+    if params['name']:
+        manage_spells(module)
+
+
+# import module snippets
+from ansible.module_utils.basic import *
```
gundalow: Could you just import the specific things you need?

gundalow (changes requested):
Just a few minor tweaks needed. Nice to see a module with decent documentation. 

I see @bcoca has a few comments on the previous PR which may be outstanding, so we should get his mark of approval before merging as well.

lib/ansible/modules/packaging/os/sorcery.py
vaygr: absolutely, was thinking about it.

lib/ansible/modules/packaging/os/sorcery.py
vaygr: Yeah, I see, done.

lib/ansible/modules/packaging/os/sorcery.py
vaygr: done,

btw, is there any value in making this module more portable so it'd work with older branches of Ansible, e.g. `1.9`?

gundalow (approved):

gundalow:
Ansible 1.9 will be going out of support very soon. Ansible 2.2 was released recently and I'd strongly suggest upgrading. 

As a general rule we don't backport new modules. However if you include this file in `./library/` directory along side `roles` if will be available to you.

gundalow:
@vaygr if @bcoca has any feedback please include that in a follow up PR.

Thanks for the new module.

vaygr:
will do, thank you.

PR #19073, opened 2016-12-09
Add full change check for rabbitmq policy

### LINKED ISSUE #1566: Ansible barfs if ANSIBLE_HOSTS file is +x
If your ANSIBLE_HOSTS file ends up being +x for some reason, it appears that ansible... attempts to run it as a program?

### LINKED ISSUE #29264: rabbitmq_policy doesn't update a policy that already exists
##### ISSUE TYPE
 - Bug Report

##### COMPONENT NAME
rabbitmq_policy

##### ANSIBLE VERSION
`2.3`

##### SUMMARY
From @tedgin on 2016-01-28T18:29:35Z
Assume we have a policy named `user_queues` that applies to all queues matching the pattern `^prefix_`.  Requirements change and `user_queues` needs to be applied to queues matching the new pattern `/`.  If the play using `rabbitmq_policy` to set policy is modified to set the new pattern and run, it will run successfully, but the `user_queues` policy remains unchanged.

It would be better if `rabbitmq_policy` would change policies that already exists or at least generate a warning or error.

Copied from original issue: ansible/ansible-modules-extras#1566

puzan:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

rabbitmq_policy
##### ANSIBLE VERSION

```
ansible 2.1.0 (devel 7e3963420d) last updated 2016/05/15 11:11:28 (GMT +300)
  lib/ansible/modules/core: (detached HEAD 25414bbd1b) last updated 2016/05/15 12:07:46 (GMT +300)
  lib/ansible/modules/extras: (detached HEAD e13942f9c5) last updated 2016/05/15 12:07:46 (GMT +300)
  config file =
```
##### SUMMARY

Fixes #29264.

Add check for all policy fields (name, apply_to, pattern, tags, priority) to have correct changed state. Previosly changed state was based on policy name only.

Test playbook (pattern change):

``` yaml
- name: Update policy
  hosts: localhost
  tasks:
    - name: add policy
      rabbitmq_policy: >
        name=HA
        pattern='.*'
        tags="ha-mode=all"
        node=rabbit@localhost
    - name: update policy
      rabbitmq_policy: >
        name=HA
        pattern='test'
        tags="ha-mode=all"
        node=rabbit@localhost
```

Before fix:

```
PLAY [Update policy] ***********************************************************

ok: [localhost]

TASK [add policy] **************************************************************
changed: [localhost]

TASK [update policy] ***********************************************************
ok: [localhost]

localhost                  : ok=3    changed=1    unreachable=0    failed=0
```

After fix:

```
PLAY [Update policy] ***********************************************************

ok: [localhost]

TASK [add policy] **************************************************************
changed: [localhost]

TASK [update policy] ***********************************************************
changed: [localhost]

localhost                  : ok=3    changed=2    unreachable=0    failed=0
```

### DIFF
--- a/lib/ansible/modules/messaging/rabbitmq/rabbitmq_policy.py
+++ b/lib/ansible/modules/messaging/rabbitmq/rabbitmq_policy.py
@@ -38,12 +38,17 @@ options:
     version_added: "2.1"
   pattern:
     description:
-      - A regex of queues to apply the policy to.
-    required: true
+      - A regex of queues to apply the policy to. Required when
+        C(state=present). 'Required' flag was changed to 'false' in 2.8.
+        version.
+    required: false
+    default: null
   tags:
     description:
-      - A dict or string describing the policy.
-    required: true
+      - A dict or string describing the policy. Required when
+        C(state=present). 'Required' flag was changed to 'false' in 2.8.
+    required: false
+    default: null
   priority:
     description:
       - The priority of the policy.
@@ -102,16 +107,19 @@ class RabbitMqPolicy(object):
             return out.splitlines()
         return list()
 
-    def list(self):
+    def has_modifications(self):
+        if self._pattern is None or self._tags is None:
+            self._module.fail_json(msg=('pattern and tags are required for '
+                                        'state=present'))
+
         policies = self._exec(['list_policies'], True)
 
-        for policy in policies:
-            if not policy:
-                continue
-            policy_name = policy.split('\t')[1]
-            if policy_name == self._name:
-                return True
-        return False
+        return not any(self._policy_check(policy) for policy in policies)
+
+    def should_be_deleted(self):
+        policies = self._exec(['list_policies'], True)
+
+        return any(self._policy_check_by_name(policy) for policy in policies)
 
     def set(self):
         args = ['set_policy']
@@ -128,14 +136,40 @@ class RabbitMqPolicy(object):
     def clear(self):
         return self._exec(['clear_policy', self._name])
 
+    def _policy_check(self, policy):
+        if not policy:
+            return False
+
+        policy_data = policy.split('\t')
+
+        policy_name = policy_data[1]
+        apply_to = policy_data[2]
+        pattern = policy_data[3].replace('\\\\', '\\')
+        tags = json.loads(policy_data[4])
+        priority = policy_data[5]
+
+        return (policy_name == self._name and
+                apply_to == self._apply_to and
+                tags == self._tags and
+                priority == self._priority and
+                pattern == self._pattern)
+
+    def _policy_check_by_name(self, policy):
+        if not policy:
+            return False
+
+        policy_name = policy.split('\t')[1]
+
+        return policy_name == self._name
+
 
 def main():
     arg_spec = dict(
         name=dict(required=True),
         vhost=dict(default='/'),
-        pattern=dict(required=True),
+        pattern=dict(required=False, default=None),
         apply_to=dict(default='all', choices=['all', 'exchanges', 'queues']),
-        tags=dict(type='dict', required=True),
+        tags=dict(type='dict', required=False, default=None),
         priority=dict(default='0'),
         node=dict(default='rabbit'),
         state=dict(default='present', choices=['present', 'absent']),
@@ -151,14 +185,12 @@ def main():
     rabbitmq_policy = RabbitMqPolicy(module, name)
 
     result = dict(changed=False, name=name, state=state)
-    if rabbitmq_policy.list():
-        if state == 'absent':
-            rabbitmq_policy.clear()
-            result['changed'] = True
-        else:
-            result['changed'] = False
-    elif state == 'present':
+
+    if state == 'present' and rabbitmq_policy.has_modifications():
         rabbitmq_policy.set()
+        result['changed'] = True
+    elif state == 'absent' and rabbitmq_policy.should_be_deleted():
+        rabbitmq_policy.clear()
         result['changed'] = True
 
     module.exit_json(**result)
--- a/lib/ansible/modules/messaging/rabbitmq/rabbitmq_policy.py
+++ b/lib/ansible/modules/messaging/rabbitmq/rabbitmq_policy.py
@@ -39,14 +39,13 @@ options:
   pattern:
     description:
       - A regex of queues to apply the policy to. Required when
-        C(state=present). 'Required' flag was changed to 'false' in 2.8.
-        version.
+        C(state=present). This option is no longer required as of Ansible 2.9.
     required: false
     default: null
   tags:
     description:
       - A dict or string describing the policy. Required when
-        C(state=present). 'Required' flag was changed to 'false' in 2.8.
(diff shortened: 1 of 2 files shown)


puzan:
Migrated from https://github.com/ansible/ansible-modules-extras/pull/2236

puzan:
It fixes https://github.com/ansible/ansible-modules-extras/issues/1566

puzan:
Hello @retr0h, @bcoca, @jimi-c!

Who can review this old PR?

lib/ansible/modules/messaging/rabbitmq_policy.py
```diff
@@ -51,13 +51,15 @@
     version_added: "2.1"
   pattern:
     description:
-      - A regex of queues to apply the policy to.
-    required: true
+      - A regex of queues to apply the policy to. Required when
+        C(state=present).
+    required: false
```
bcoca: this is changing the required/default from how the module used to work, you need to update the description and note that this changes in 2.4 vs the previous 'required' behaviour

bcoca (changes requested):

bcoca:
cc @chrishoffman @manuel-sousa @romanek-adam for comments/review

lib/ansible/modules/messaging/rabbitmq_policy.py
puzan: @bcoca I have updated description. Is there 'version_changed' tag like 'version_added'?

puzan:
ready_for_review

puzan:
bot_status

puzan:
ready_for_review

puzan:
ready_for_review

lib/ansible/modules/messaging/rabbitmq/rabbitmq_policy.py
```diff
@@ -93,25 +101,55 @@ def __init__(self, module, name):
         self._node = module.params['node']
         self._rabbitmqctl = module.get_bin_path('rabbitmqctl', True)
 
-    def _exec(self, args, run_in_check_mode=False):
+        self._version = self._rabbit_version()
+
+    def _exec(self, args, run_in_check_mode=False, split_lines=True):
         if not self._module.check_mode or (self._module.check_mode and run_in_check_mode):
             cmd = [self._rabbitmqctl, '-q', '-n', self._node]
             args.insert(1, '-p')
```
barryib: @puzan `rabbitmqctl status` doesn't support `vhost`. I run RabbitMQ 3.6.16.

As quick fix, I added `add_vhost=True` as default argument to `_exec` to have something like this :

```python
def _exec(self, args, run_in_check_mode=False, split_lines=True, add_vhost=True):
    .... some code here ....
    if add_vhost:
        args.insert(1, '-p')
        args.insert(2, self._vhost)
```

barryib (changes requested):

samdoran:
!waffling needs_revision

barryib:
@puzan @samdoran It would be nice to do some kind of sanitization for policy tags. In fact, for ansible, almost everything is a string. So it is difficult to use variable to define integer tags.

```yml
# Variable
my_tags:
  expires: 604800000
  ha-mode: exactly
  ha-params: "{{ (( rab_nodes_list | length ) / 2) | int + 1 }}"
  ha-sync-mode: manual

# Policy definition
rabbitmq_policy:
  name: ha
  pattern: 'queue.*'
  priority: 10
  apply_to: queues
  state: present
  tags: "{{ my_tags }}"
```

`ha-params` must be an integer. But with ansible, `{{ (( rab_nodes_list | length ) / 2) | int + 1 }}` return a string.

lib/ansible/modules/messaging/rabbitmq/rabbitmq_policy.py
puzan: I have added your solution.

barryib:
@puzan great.

@puzan @samdoran @bcoca What about my second suggestion for sanitization ? Should I open a different PR for that ?

Actually, here is how I fixed the issue with integer in rabbitmq policy 👍 

```python
# Sanitize function
def _sanitize_tags(self, tags):
    sanitezed_tags = tags
    for key, value in sanitezed_tags.items():
        if isinstance(value, str) and value.isdigit():
            sanitezed_tags[key] = int(value)
    return sanitezed_tags

# Module init
def __init__(self, module, name):
    ....
    self._tags = self._sanitize_tags(module.params['tags'])
    ....
```

BTW, any chance to see this PR merged ? It's almost 3 years old.

samdoran:
@barrylib The best way to validate a field like that is to create a custom validation function. You can use your existing function, just change the argument_spec of tags.

```python
...
    tags={'type': _sanitize_tags, 'required': True}
...
```

This should be in a separate PR.

samdoran:
@barryib Since this is a community module, we rely on votes from community members to gauge if it's appropriate to merge. Please [add your vote](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md#when-will-your-pull-request-be-merged).

lib/ansible/modules/messaging/rabbitmq/rabbitmq_policy.py
```diff
@@ -38,12 +38,17 @@
     version_added: "2.1"
   pattern:
     description:
-      - A regex of queues to apply the policy to.
-    required: true
+      - A regex of queues to apply the policy to. Required when
+        C(state=present). 'Required' flag was changed to 'false' in 2.8.
```
samdoran:
```suggestion
        C(state=present). This option is no longer required as of Ansible 2.9.
```

lib/ansible/modules/messaging/rabbitmq/rabbitmq_policy.py
```diff
@@ -38,12 +38,17 @@
...
+        version.
+    required: false
+    default: null
   tags:
     description:
-      - A dict or string describing the policy.
-    required: true
+      - A dict or string describing the policy. Required when
+        C(state=present). 'Required' flag was changed to 'false' in 2.8.
```
samdoran:
```suggestion
        C(state=present). This option is no longer required as of Ansible 2.9.
```

samdoran (changes requested):
Please create a changelog fragment and we can merge this. See [this fragment](https://github.com/ansible/ansible/blob/stable-2.7/changelogs/fragments/reboot-change-default-boot-command.yaml) as an example.

lib/ansible/modules/messaging/rabbitmq/rabbitmq_policy.py
puzan: Fixed

lib/ansible/modules/messaging/rabbitmq/rabbitmq_policy.py
puzan: Fixed

puzan:
@samdoran I have added changelog and apply your comments.

puzan:
ready_for_review

